PR #525, opened 2016-07-05
Prevent massive amount of MSVC C4100 warnings

neusdan:
[C4100 warning](https://msdn.microsoft.com/de-de/library/26kb9fy0.aspx)

> C4100 can also be issued when code calls a destructor on a otherwise unreferenced parameter of primitive type. This is a limitation of the Visual C++ compiler.

Without this "fix" there are > 150 of this warnings.

### DIFF
--- a/src/lib/base/secmem.h
+++ b/src/lib/base/secmem.h
@@ -82,7 +82,12 @@ class secure_allocator
          ::new(static_cast<void*>(p)) U(std::forward<Args>(args)...);
          }
 
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable: 4100)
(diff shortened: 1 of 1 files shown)


src/lib/base/secmem.h
```diff
@@ -82,7 +82,12 @@ class secure_allocator
          ::new(static_cast<void*>(p)) U(std::forward<Args>(args)...);
          }
 
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable: 4100)
       template<typename U> void destroy(U* p) { p->~U(); }
```
cdesjardins: I fix this warning locally with this:
template<typename U> void destroy(U* p) { p = p; p->~U(); }

Not that it really matters.

PR #529, opened 2016-07-11
Add test vectors for block cipher padding modes

cordney:
Exports get_bc_pad() to be used from tests. Adds separate handcrafted
tests for block cipher padding modes. They were previously only tested
implicitly during the block cipher modes of operation tests, though not
all padding modes were covered. And in case a mode of operation is
not part of the enabled modules, the previously tested padding modes
are not covered at all. Fixes an off-by-one bug in the previously
untested ANSI X9.23 padding mode, where the number of zero bytes
in the pad was one more than allowed by the standard.


cordney:
FYI: This is the first in a series of merge request to push the code coverage with BSI module policy to 80%. Currently, it's at 76,2%. Though, coverage with all modules will certainly also profit.

randombit:
Nice. This could also be extended later on to better test handling of invalid paddings.

PR #530, opened 2016-07-11
Remove egd entropy source from BSI module policy

cordney:
Little is known about the quality of randomness provided by egd, whereas /dev/random has been analyzed quite deeply, e.g., in https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Publikationen/Studien/LinuxRNG/LinuxRNG_Studie.pdf?__blob=publicationFile (German only). Also most platforms provide /dev/random or a similar strong entropy source nowadays.

### DIFF
--- a/src/build-data/policy/bsi.txt
+++ b/src/build-data/policy/bsi.txt
@@ -59,7 +59,6 @@ beos_stats
 cryptoapi_rng
 darwin_secrandom
 dev_random
-egd
 hres_timer
 proc_walk
 rdrand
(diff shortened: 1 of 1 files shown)


randombit:
At this point EGD support has not been tested in a long time, and is probably not needed anymore; either the OS has a /dev/random or else it is an obsolete/unpatched system. It dates back to supporting (then modern, now very old) versions of Solaris and IRIX which did not have a random device.

I'll merge this for now and add a deprecation warning for EGD in the next release, so if anyone is still using EGD support downstream they have time to object. I would be surprised if anyone is, though.

cordney:
Agree. Thx for merging.

PR #533, opened 2016-07-13
DLIES rework

neusdan:
- With these fixes the implementation is now compatible with bouncycastle and it should operate
as it is specified in "DHIES: An encryption scheme based on Diffie-Hellman Problem" or in BSI
technical guideline TR-02102-1.

- In addition to the already present XOR-encrypion/decryption mode it's now possible to use DLIES with a block cipher.

- Previously the input to the KDF was the concatenation of the (ephemeral) public key
and the secret value derived by the key agreement operation:

```
secure_vector<byte> vz(m_my_key.begin(), m_my_key.end());
vz += m_ka.derive_key(0, m_other_key).bits_of();

const size_t K_LENGTH = length + m_mac_keylen;
secure_vector<byte> K = m_kdf->derive_key(K_LENGTH, vz);
```

I don't know why this was implemented like this. But now the input to the KDF is only the secret value obtained by the key agreement operation.

- Furthermore the order of the output was changed from {public key, tag, ciphertext} to {public key, ciphertext, tag}.

- Multiple test vectors added that were generated with bouncycastle and some with botan itself.


neusdan:
Unfortunately i did not find any official test vectors.

randombit:
> I don't know why this was implemented like this.

I wrote this code a long long time ago based on a IEEE 1363 draft, and do not remember the details or if I intentionally did something 'extra'. Perhaps the spec itself changed, before the final version. Anyway it is unfortunate to break compatability with older versions with this, but being able to interop with another version is probably worth the pain.

Issue #11, opened 2014-02-01
MSVC x86 Unresolved External Symbol Error on __mm256_set_epi64x


LightBender:
However it compiles just fine on x64. My research indicates that this intrinsic is present in both archs (http://msdn.microsoft.com/en-us/library/hh977023.aspx), so I am wondering if a header is missing from the x86 amalgamation build or some other misconfiguration?

PR #540, opened 2016-07-19
Fix introduced incompatibility in use of HMAC_DRBG in PKCS#11 test

mouse07410:
The change in HMAC_DRBG introduced incompatibility with `test_pkcs11_high_level.cpp`. This PR remedies it.

### DIFF
--- a/src/tests/test_pkcs11_high_level.cpp
+++ b/src/tests/test_pkcs11_high_level.cpp
@@ -1342,16 +1342,21 @@ Test::Result test_pkcs11_hmac_drbg()
    Test::Result result("PKCS11 HMAC_DRBG using PKCS11_RNG");
    TestSession test_session(true);
 
-   HMAC_DRBG drbg("SHA-512", 0);
-   result.test_success("HMAC_DRBG(HMAC(SHA512)) instantiated with PKCS11_RNG");
+   PKCS11_RNG *rng = new PKCS11_RNG(test_session.session());
+   MessageAuthenticationCode *hmac = MessageAuthenticationCode::create("HMAC(SHA-512)").release();
+   HMAC_DRBG drbg(hmac) ;
+   result.test_success("HMAC_DRBG(\"SHA-512\") instantiated with PKCS11_RNG");
 
    result.test_eq("HMAC_DRBG is not seeded yet.", drbg.is_seeded(), false);
 
+   secure_vector<byte> input = rng->random_vec(hmac->output_length());
+   drbg.initialize_with(input.data(), input.size());
+   
+   result.test_eq("HMAC_DRBG is seeded now", drbg.is_seeded(), true);
+
    std::string personalization_string = "Botan PKCS#11 Tests";
    std::vector<byte> personalization_data(personalization_string.begin(), personalization_string.end());
-   drbg.initialize_with(personalization_data.data(), personalization_data.size());
-
-   result.test_eq("HMAC_DRBG is seeded now", drbg.is_seeded(), true);
+   drbg.add_entropy(personalization_data.data(), personalization_data.size());
 
    auto rnd_vec = drbg.random_vec(256);
    result.test_ne("HMAC_DRBG generated a random vector", rnd_vec, std::vector<byte>(256));
--- a/src/tests/test_pkcs11_high_level.cpp
+++ b/src/tests/test_pkcs11_high_level.cpp
@@ -1342,14 +1342,14 @@ Test::Result test_pkcs11_hmac_drbg()
    Test::Result result("PKCS11 HMAC_DRBG using PKCS11_RNG");
    TestSession test_session(true);
 
-   HMAC_DRBG drbg(MessageAuthenticationCode::create("HMAC(SHA-512)").release(), new PKCS11_RNG(test_session.session()));
+   HMAC_DRBG drbg("SHA-512", 0);
    result.test_success("HMAC_DRBG(HMAC(SHA512)) instantiated with PKCS11_RNG");
 
    result.test_eq("HMAC_DRBG is not seeded yet.", drbg.is_seeded(), false);
 
    std::string personalization_string = "Botan PKCS#11 Tests";
    std::vector<byte> personalization_data(personalization_string.begin(), personalization_string.end());
-   drbg.add_entropy(personalization_data.data(), personalization_data.size());
+   drbg.initialize_with(personalization_data.data(), personalization_data.size());
 
    result.test_eq("HMAC_DRBG is seeded now", drbg.is_seeded(), true);
 
--- a/src/tests/test_pkcs11_high_level.cpp
+++ b/src/tests/test_pkcs11_high_level.cpp
@@ -1345,7 +1345,7 @@ Test::Result test_pkcs11_hmac_drbg()
    PKCS11_RNG *rng = new PKCS11_RNG(test_session.session());
    MessageAuthenticationCode *hmac = MessageAuthenticationCode::create("HMAC(SHA-512)").release();
    HMAC_DRBG drbg(hmac) ;
-   result.test_success("HMAC_DRBG(\"SHA-512\") instantiated with PKCS11_RNG");
+   result.test_success("HMAC_DRBG(HMAC(SHA512)) instantiated with PKCS11_RNG");
 
    result.test_eq("HMAC_DRBG is not seeded yet.", drbg.is_seeded(), false);
 


src/tests/test_pkcs11_high_level.cpp
```diff
@@ -1342,14 +1342,14 @@ Test::Result test_pkcs11_hmac_drbg()
    Test::Result result("PKCS11 HMAC_DRBG using PKCS11_RNG");
    TestSession test_session(true);
 
-   HMAC_DRBG drbg(MessageAuthenticationCode::create("HMAC(SHA-512)").release(), new PKCS11_RNG(test_session.session()));
+   HMAC_DRBG drbg("SHA-512", 0);
```
randombit: Is this really the right fix here? With this change PKCS11_RNG is not tested...

neusdan: exactly, the HMAC DRBG has to be instantiated with the PKCS11 RNG here.

mouse07410:
> the HMAC DRBG has to be instantiated with the PKCS11 RNG here

Good point - but how would you do that, given that the current constructors of HMAC_DRBG class (as listed in `src/lib/rng/hmac_drbg/hmac_drbg.h`) do not allow for another parameter, let alone another RNG?

neusdan:
that's indeed a problem

mouse07410:
> Is this really the right fix here? With this change PKCS11_RNG is not tested...

It is the right fix if you want to be able to build `botan-test` executable at all, and run the other tests.

>> how would you do that, given that the current constructors of HMAC_DRBG class (as listed in `src/lib/rng/hmac_drbg/hmac_drbg.h`) do not allow for another RNG?
> 
> that's indeed a problem

And I'm afraid this problem can only be addressed by modifications to `src/lib/rng/hmac_drbg/hmac_drbg.*`, unless you're willing to test PKCS11_RNG on some other generator...?

In any case, keeping the master un-compilable is not good, and probably should be addressed rather sooner than later.

src/tests/test_pkcs11_high_level.cpp
mouse07410: No argument, but how are you going to do that? HMAC_DRBG does not have the appropriate constructors any more.

HMAC_DRBG needs to be restored to what it was last September to provide that capability again.

mouse07410:
@randombit, @neusdan - what do you think? Should we scrap `hmac-drbg-pkcs11` test? Should @randombit restore the old constructors and behavior of `HMAC_DRBG`? Since the current master is broken, a quick decision seems necessary.

mouse07410:
If you would rather not modify the current `HMAC_DRBG` (e.g. because the current implementation is NIST-conformant, and you'd rather keep it that way), would the following make sense?

```
diff --git a/src/tests/test_pkcs11_high_level.cpp b/src/tests/test_pkcs11_high_level.cpp
index 877c44f..f3bc04e 100644
--- a/src/tests/test_pkcs11_high_level.cpp
+++ b/src/tests/test_pkcs11_high_level.cpp
@@ -1342,15 +1342,20 @@ Test::Result test_pkcs11_hmac_drbg()
    Test::Result result("PKCS11 HMAC_DRBG using PKCS11_RNG");
    TestSession test_session(true);
 
-   HMAC_DRBG drbg("SHA-512", 0);
-   result.test_success("HMAC_DRBG(HMAC(SHA512)) instantiated with PKCS11_RNG");
+   PKCS11_RNG *rng = new PKCS11_RNG(test_session.session());
+   MessageAuthenticationCode *hmac = MessageAuthenticationCode::create("HMAC(SHA-512)").release();
+   HMAC_DRBG drbg(hmac) ;
+   result.test_success("HMAC_DRBG(\"SHA-512\") instantiated with PKCS11_RNG");
 
    result.test_eq("HMAC_DRBG is not seeded yet.", drbg.is_seeded(), false);
 
-   std::string personalization_string = "Botan PKCS#11 Tests";
-   std::vector<byte> personalization_data(personalization_string.begin(), personalization_string.end());

-   drbg.initialize_with(personalization_data.data(), personalization_data.size());
+   secure_vector<byte> input = rng->random_vec(hmac->output_length());
+   drbg.initialize_with(input.data(), input.size());
+   
    result.test_eq("HMAC_DRBG is seeded now", drbg.is_seeded(), true);
 
    auto rnd_vec = drbg.random_vec(256);
```

If so, would you like me to update my PR?

neusdan:
For me this wouldn't be sufficient.  I need a way to use the HMAC_DRBG which only gets seeded from the token/PKCS11_RNG and not from other entropy sources.

mouse07410:
> I need a way to use the HMAC_DRBG which only gets seeded from the token/PKCS11_RNG and not from other entropy sources.

But... I thought that's what the proposed code does - seeding the HMAC_DRBG with PKCS11_RNG output...?

neusdan:
Ok, yes, the HMAC_DRBG is initially seeded with the PKCS11_RNG that way, but it's never reseeded again automatically.

So if for example 512 byte output are produced the HMAC_DRBG should reseed itself with token entropy.

That's how it was before the RNG changes i think.

mouse07410:
> Ok, yes, the HMAC_DRBG is initially seeded with the PKCS11_RNG that way, but it's never reseeded again automatically.

I see. But unless HMAC_DRBG functionality is changed (and I'm not sure whether it would still be NIST-conformant in that case) that's how it is from now on. 

And it satisfies your requirement:
> HMAC_DRBG ... only gets seeded from the token/PKCS11_RNG and not from other entropy sources

by the virtue of not having any other entropy sources available to HMAC_DRBG.
   
> So if for example 512 byte output are produced the HMAC_DRBG should reseed itself with token entropy. That's how it was before the RNG changes i think.

Yes that's how it was - but it isn't any more. So with the current code I don't think it is possible. If you insist on (automatic continuous) re-seeding, I see two options:

1. Leave it as is (with my changes - because otherwise it won't even compile, let alone build or run). After all, all we should need here is HMAC_DRBG being able to pull randomness from the token/PKCS11 RNG, right?
2. Try to build `Entropy_Sources` object and add `PKCS11_RNG` to it, then try `reseed_with_sources()`. I have no idea (a) whether this would work at all, (b) what's the `poll_bits` should be for a token (AFAIK, 64 bits should be reasonable), and (c) what the `poll_timeout` should be (in milliseconds :). Do you know how to do that?

neusdan:
Of course this is much better than having a build that does not compile (we really need to integrate the PKCS#11 tests into the CI builds).

But for the automatic reseed i need to find a solution. I like to hear @randombit opinion on this.

mouse07410:
> Of course this is much better than having a build that does not compile (we really need to integrate the PKCS#11 tests into the CI builds)

:-)  Yes and yes. :-)

> But for the automatic reseed i need to find a solution. I like to hear @randombit opinion on this.

I'd like to hear his opinion too. But in the meanwhile - why are you not satisfied with the "normal" NIST-approved HMAC-DRBG functionality, and require it to automatically reseed? Rather than just seed it form the token, and once in a while (if/when the application thinks it consumed "enough" randomness) reseed it? Or subclass HMAC-DRBG to add a constructor that takes RNG as an extra parameter, and automatic `reseed_from_rng()` that would be called by re-defined `update()`?

randombit:
The CI build enables most of the third party deps but both PKCS #11 (and TPM) support are missing completely. As with the experience with the CVC code which bitrotted and was eventually just removed entirely rather than spend the effort fixing, having no coverage quickly causes bitrot and bugs.

For the RNG changes. @neusdan your use case of HMAC_DRBG combined with the system RNG or a hardware RNG (token, RDRAND, TPM) without any other form of entropy collection is a totally reasonable one in my view and worth supporting. My changes were driven by a desire to make HMAC_DRBG a standalone RNG type, without requiring HMAC_RNG, but now instead it requires/calls the entropy collecting code which is an even larger dependency and has effects on applications.

Here is my proposal for a resolution. I will partially revert the changes to HMAC_DRBG. which gets back a non-owned RandomNumberGenerator* member variable which is used for all seeding operations, if unset then reseeding is disabled and it is is purely a DRBG (for example as required for deterministic signatures). Then for AutoSeeded_RNG, HMAC_DRBG reads from an HMAC_RNG as before, and HMAC_RNG is the sole user of the entropy API. I would also I think revert the change merging auto_rng into the base module; for example if you only wanted to use the system or token RNG, then allowing auto_rng to be disabled in the build (and thus allowing also removing HMAC_RNG and the entropy collection code)

Another RNG change that I need to make to complete this RNG work: HMAC_RNG is based on the HKDF Extract-then-Expand paper, and uses an early/nonstandard version of HKDF. I am planning on changing it to use normal HKDF, and remove the non-determistic elements of the PRF (which are now included as the additional input by the AutoSeeded_RNG). That would allow writing tests for HMAC_RNG, which currently has none :(

The current reseeding logic seems unsatisfactory in a number of ways, and it is basically impossible to (for example) prevent the library from trying to collect entropy by doing things like walking /proc (or executing arbitrary processes...). I do still think hashing in (say) /proc/interrupts along with the OS RNG outputs to generate the DRBG seed is a good hedge, but not everyone agrees and it would be good to give applications better control over what happens and when. I have an idea for addressing this which I'll propose in a later PR.

mouse07410:
> I will partially revert the changes to HMAC_DRBG. which gets back a non-owned RandomNumberGenerator* member variable which is used for all seeding operations, if unset then reseeding is disabled and it is is purely a DRBG (for example as required for deterministic signatures).

I think this would be the best solution.

Any timeframe on this change? ;)

mouse07410:
Looks like it's been fixed now:
```
. . . . .
PKCS11 import ECDH public key ran 1 tests all ok
PKCS11 HMAC_DRBG using PKCS11_RNG ran 4 tests all ok
PKCS11 RNG add entropy random ran 1 tests all ok
PKCS11 RNG generate random ran 1 tests all ok
PKCS11 X509 cert import ran 2 tests all ok
Tests complete ran 78 tests all tests ok
```

Issue #465, opened 2016-04-05
Trouble with static initializers for Algo_Registry


steipete:
Hi there - first of all, thanks for this wonderful library and the opportunity to ditch OpenSSL. We're using Botan in our project (PSPDFKit) and it mostly works great. One design problem is Botan's use of static initializers for Algorithms.

Botan's static constructor runs very early:

```
(lldb) bt
* thread #1: tid = 0x2d67f9, 0x000000010d15b104 PSPDFKitTests`Botan::Algo_Registry<Botan::BlockCipher>::global_registry() + 4 at botan_all_internal.h:70, queue = 'com.apple.main-thread', stop reason = breakpoint 6.2
    frame #0: 0x000000010d15b104 PSPDFKitTests`Botan::Algo_Registry<Botan::BlockCipher>::global_registry() + 4 at botan_all_internal.h:70
    frame #1: 0x000000010d168244 PSPDFKitTests`Botan::Algo_Registry<Botan::BlockCipher>::Add::Add(this=0x000000010e0c16d8, basename="AES-128", fn=Botan::Algo_Registry<Botan::BlockCipher>::maker_fn @ 0x00007fff5e299470, provider="base", pref='d')>, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, unsigned char) + 52 at botan_all_internal.h:123
    frame #2: 0x000000010d107bd2 PSPDFKitTests`Botan::Algo_Registry<Botan::BlockCipher>::Add::Add(this=0x000000010e0c16d8, basename="AES-128", fn=<unavailable>, provider="base", pref='d')>, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, unsigned char) + 50 at botan_all_internal.h:122
  * frame #3: 0x000000010d25e51d PSPDFKitTests`::__cxx_global_var_init.145() + 397 at botan_all.cpp:4695
    frame #4: 0x000000010d261523 PSPDFKitTests`_GLOBAL__sub_I_botan_all.cpp + 19 at botan_all.cpp:0
    frame #5: 0x000000010198e9cf dyld_sim`ImageLoaderMachO::doModInitFunctions(ImageLoader::LinkContext const&) + 265
    frame #6: 0x000000010198eb48 dyld_sim`ImageLoaderMachO::doInitialization(ImageLoader::LinkContext const&) + 40
[...]

PR #542, opened 2016-07-19
Mark LibraryInitializer as deprecated

neusdan:
first tried to deprecated the ctor, but for some reason this had no effect. 

With this patch the warning is shown when the `init.h` header is included.


randombit:
I don't want to do this quite yet - it may turn out we need an initialization function to resolve the current mess with static initialization (#465). Until a fix for that is designed and landed (I hope to work on this soon, it is certainly a blocker to 2.0 in my mind), I don't want to limit our options in advance.

PR #373, opened 2015-12-20
RdRand and RdSeed logic changes

neusdan:
I like to hear your opinion on the following changes.

For RdRand it is [guaranteed](https://software.intel.com/en-us/articles/intel-digital-random-number-generator-drng-library-implementation-and-uses) to generate a random number within 10 retries. So i changed the logic to perform at max 10 retries. Additionally i want to be able to configure how often RdRand and RdSeed is polled.

Unfortunately for RdSeed there are no guarantees as to how often a thread should retry the instruction. So i made the number of retries also configurable.

PR #543, opened 2016-07-19
Add RDRAND_RNG

randombit:
An RNG type that calls RDRAND directly.

Adds static functions `RDRAND_RNG::rdrand_status` which retries up to the limit set in build.h, and `RDRAND_RNG::rdrand` which retries indefinitely.

There was some discussion of need for these interfaces in #373 

The RDRAND entropy source just generates output using `RDRAND_RNG` now.


neusdan:
Besides the configure.py error, LGTM.

PR #544, opened 2016-07-20
Make Stream_Cipher::set_iv() pure virtual

cordney:
It provided a default implementation that only checked that the length was correct, but ignored the actual data and did not notify the caller, which seemed like a rather odd behaviour.

The only implementation that used this default implementation, RC4, now throws an exception.


randombit:
The old version had much the same effect since the default implementation of StreamCipher::valid_iv_length only accepts empty IVs. But this version is clearer, and it is very unlikely any new stream cipher that did not support IVs would be introduced.

PR #545, opened 2016-07-20
Add OctetString unit tests

neusdan:
...and add the possibility to compare OctetStrings with the botan test-framework


randombit:
Thanks

PR #549, opened 2016-07-21
Refactor of MP and SIMD headers

randombit:
This is somewhat large but mechanical change. It merges the different versions of mp_madd.h and mp_asmi.h into a single file with preprocessor checks. Previously these relied on some configure.py magic to choose the right version to be inserted into the build header directory.

It also merges the SSE2, AltiVec and scalar-emulation SIMD into a single class, instead of having them split into different headers.


randombit:
In addition to CI coverage I've tested this on PowerPC for checking AltiVec, everything looks good. I think this can be merged, any comments?

webmaster128:
good job, thanks

PR #551, opened 2016-07-24
ANSSI elliptic curve cryptography frp256v1

scogliani:
- Parameters availables here: https://www.legifrance.gouv.fr/affichTexte.do?cidTexte=JORFTEXT000024668816
- DER format according to the ANS1 syntax defined in ANSI X9.62 standard
  available here: http://www.ssi.gouv.fr/agence/publication/publication-dun-parametrage-de-courbe-elliptique-visant-des-applications-de-passeport-electronique-et-de-ladministration-electronique-francaise/

I created a simple program for validating the implementation. Same as these from the NIST CAVP.

Testing program available on my github: https://github.com/scogliani/ecc-test-vectors


randombit:
As far as the actual changes to the source -  looks fine and thank you for the test cases also.

I am somewhat unsure if adding this curve is really desirable, though - my understanding is unlike the NIST curves this curve was generated without even the pretense of being random and with the parameters completely unexplained. And its twist has only about 2**80 security versus about 2**120 for P-256 (https://safecurves.cr.yp.to/twist.html). Can you explain why anyone would want to use this curve instead of P-256 or X25519?

scogliani:
The ANSSI (Agence Nationale de la Sécurité des Systèmes d'Information) is the French equivalent of the German BSI (Bundesamt für Sicherheit in der Informationstechnik). According to their General Security Recommendation, this curve is used with ECDSA/ECKCDSA: https://www.keylength.com/en/5/. Also recommanded.

Furthermore, this curve is implemented in the French electronic passport and French administrative representatives.

Eventually, it would be interesting having the possibility of selecting a specific ECC standard even if it is not fully known how it was generated and secure in a specific context, instead of implementing a patch.

randombit:
OK, seems a sufficient justification for inclusion to me. I think there is a need for an ability to control at build time which curves are included but that is not a problem specific to this curve.

PR #552, opened 2016-07-25
AEAD test improvements

neusdan:
- test clear()
- test update()

set tag length in GCM_Mode::name()

clear only the state in AEAD modes not the values that were used during construction

### DIFF
--- a/src/lib/modes/aead/ccm/ccm.cpp
+++ b/src/lib/modes/aead/ccm/ccm.cpp
@@ -1,6 +1,7 @@
 /*
 * CCM Mode Encryption
 * (C) 2013 Jack Lloyd
+* (C) 2016 Daniel Neus, Rohde & Schwarz Cybersecurity
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
@@ -33,7 +34,13 @@ CCM_Mode::CCM_Mode(BlockCipher* cipher, size_t tag_size, size_t L) :
 
 void CCM_Mode::clear()
    {
-   m_cipher.reset();
+   m_cipher->clear();
+   reset();
+   }
+
+void CCM_Mode::reset()
+   {
+   m_nonce.clear();
    m_msg_buf.clear();
    m_ad_buf.clear();
    }
--- a/src/lib/modes/aead/ccm/ccm.h
+++ b/src/lib/modes/aead/ccm/ccm.h
@@ -1,6 +1,7 @@
 /*
 * CCM Mode
 * (C) 2013 Jack Lloyd
+* (C) 2016 Daniel Neus, Rohde & Schwarz Cybersecurity
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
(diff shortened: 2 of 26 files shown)


randombit:
Worth noting that Cipher_Mode (base class of AEAD_Mode) clear (ex CBC, CFB) also clears the cipher key, so there is some inconsistency here. Though `clear` behaving in the way this PR has it (resetting for a new message under the same key) is arguably more useful.

Also this commit does not seem to change GCM_Mode::clear which has `m_ctr->clear();`; is there a test missing for the update+clear+update sequence?

randombit:
I'm merging this, there is still improvements possible here (see previous comment) but more tests are good.

randombit:
The AEAD docs should be updated clearly state that the correct way to un-key a cipher is to either rekey it (eg with zeros or a random key) or to delete the Cipher_Mode object entirely, rather than calling `clear`.

In thinking about this more there are actually two operations here, both of which are useful to applications. The existing `clear` - zeroise all state and the needed function `reset` which resets just the message specific state but allows encrypting again under the existing key. So, I am not going change what `clear` means, that's confusing and nasty for users. In version X, `clear` zeros the key. In version X+1, it does not. This is bad.

But, I would accept adding new `virtual void Cipher_Mode::reset()` (or some other name, whatever), which resets the message specific state, and `clear` calling `reset` and then also zeroing the keys. So, cancel my previous comment about merging this now - I think a silent API change like this is a bad idea.

randombit:
@neusdan Did you have any interest in making the above change (wrt meaning of `clear` vs a new `AEAD_Mode::reset_msg`)? I would like the tests you added here, and being able to reset message state without also clearing keying state is a useful addition, sometimes an application would need to do that, say to help handle an IO error during processing. I just don't think changing the meaning of `clear` is the way to access it. I would make the change myself eventually (ie before feature freeze in a few weeks) but if you can do it, I can instead work on the many other things that also need to happen before that time.

To simplify things the API can just be on AEAD_Mode for now, not Cipher_Mode, so no need to update CBC and company.

neusdan:
Sorry for not responding earlier. I will make your proposed  changes within one week.

randombit:
Perfect thanks

neusdan:
Updated the PR

- Add `Cipher_Mode::reset()` which resets just the message specific state and allows encrypting again under the existing key
- In `Cipher_Mode::clear()` (at some planes) use `cipher->clear()` instead of resetting the pointer which would make the cipher object unusable
- `EAX_Decryption::output_length()` bugfix?! Now its possible to decrypt an empty ciphertext (just a tag)
- Bugfix for `GCM_Decryption::finish()`
- set tag length in `GCM_Mode::name()`
- `Cipher_Mode` tests: add tests for `reset()` and `process()`
- `AEAD_Mode` tests: add tests for `reset()`, `clear()`, `update()` and `process()`

Some notes:
- Unfortunately there is a lot of duplication in the tests. But i have no more time to work on this, at least in this week.
- For `Cipher_Mode::process` the documentation states that the input size must be a multiple of update_granularity(). I think this not necessary everywhere. But if so then there should be 
```BOTAN_ARG_CHECK(sz % update_granularity() == 0);```
- In most places the `Cipher_Mode::reset()` call is not necessary because for example `set_ad()` and `start_msg()` already clear the relevant members.

randombit:
Thanks!

I'll look into the EAX size restriction, looks like a bug.

PR #553, opened 2016-07-25
EME / pubkey test improvements

neusdan:
- add test for EME::maximum_input_size()
- additionally use maximum_input_size() before pad() in OAEP and PKCS1 (remove code duplication)
- prevent C4800 MSVC warnings


neusdan:
Don't know why the AppVeyor build failed, could you restart it @randombit ?

randombit:
@neusdan I don't know why it failed either, it seems to have hung (60 minute runtime with no output), but the web interface will not let me restart the job.

randombit:
LGTM

PR #556, opened 2016-07-26
DH negative tests and bugfix

neusdan:
- fix: load_check() was called instead of gen_check() during DH privatekey generation
- Add negative tests with invalid inputs for the DH key agreement operation


neusdan:
One Travis CI build failure in the TLS v1.2 tests. I don't think this is caused by the changes in this PR. Maybe restart the job?

PR #559, opened 2016-07-26
Add PKCS#11 module in AppVeyor CI build

neusdan:
This PR:

- adds the external PKCS#11 header files into Botan
- adds a temporary fix for the PKCS#11 build and tests failures
- enables the PKCS#11 module in the AppVeyor CI build

Next steps will be to enable the PKCS#11 module in the other CI services and to add SoftHSMv2 and enable the PKCS#11 tests

### DIFF
--- a/src/scripts/ci/appveyor.yml
+++ b/src/scripts/ci/appveyor.yml
@@ -23,9 +23,12 @@ install:
        call "%ProgramFiles(x86)%\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" %PLATFORM%
     )
   - cl # check compiler version
+  
+  - appveyor DownloadFile http://download.qt.io/official_releases/jom/jom.zip -FileName jom.zip
+  - 7z e jom.zip
 
 build_script:
-  - python configure.py --cc=msvc --cpu=%PLATFORM% %MODE%
-  - nmake
+  - python configure.py --cc=msvc --cpu=%PLATFORM% %MODE% --with-pkcs11
+  - jom
   - botan-test
   - nmake install
--- a/src/scripts/ci/appveyor.yml
+++ b/src/scripts/ci/appveyor.yml
@@ -29,6 +29,6 @@ install:
 
 build_script:
   - python configure.py --cc=msvc --cpu=%PLATFORM% %MODE% --with-pkcs11
-  - jom
+  - jom -j2
   - botan-test
   - nmake install
--- a/configure.py
+++ b/configure.py
@@ -384,7 +384,7 @@ def process_command_line(args):
                           help='minimize build')
 
     # Should be derived from info.txt but this runs too early
-    third_party  = ['boost', 'bzip2', 'lzma', 'openssl', 'sqlite3', 'zlib', 'tpm']
+    third_party  = ['boost', 'bzip2', 'lzma', 'openssl', 'sqlite3', 'zlib', 'tpm', 'pkcs11']
 
     for mod in third_party:
         mods_group.add_option('--with-%s' % (mod),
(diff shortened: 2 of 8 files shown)


mouse07410:
As a temporary fix, why not instantiate PKCS11_RNG and seed HMAC_DRBG from it? I thought the main idea was to demonstrate that it can get randomness from PKCS11_RNG, not that HMAC_DRBG can be instantiated?

neusdan:
Yeah we can do this, but as already mentioned the main idea is that the HMAC_DRBG seeds itself from   PKCS11_RNG automatically :-) Either way both "solutions" are temporary and I hope this gets fixed as soon as possible.

mouse07410:
> Yeah we can do this, but as already mentioned the main idea is that the HMAC_DRBG seeds itself from PKCS11_RNG automatically :-)

I would much rather that it seeds itself from PKCS11_RNG manually when asked to, than to have a test in the PKCS11 branch that does nothing with PKCS11_<anything> at all. In that case might just drop HMAC_DRBG test altogether - as it is now it has nothing to do with PKCS11.

> Either way both "solutions" are temporary and I hope this gets fixed as soon as possible.

I concur. Though the "manual seeding" fix can be used as a workaround because it does get DRBG seeded from the token, and allows for reseeding at will. So I'd at least keep it as a sample for users who want to seed from the token *now*. While the fix is being contemplated (hopefully).

neusdan:
done

mouse07410:
Thank you!

src/scripts/ci/appveyor.yml
```diff
@@ -23,9 +23,12 @@ install:
...
   - cl # check compiler version
+  
+  - appveyor DownloadFile http://download.qt.io/official_releases/jom/jom.zip -FileName jom.zip
+  - 7z e jom.zip
 
 build_script:
-  - python configure.py --cc=msvc --cpu=%PLATFORM% %MODE%
-  - nmake
+  - python configure.py --cc=msvc --cpu=%PLATFORM% %MODE% --with-pkcs11
+  - jom
```
randombit: Why this change?

neusdan: jom is a replacement for nmake which supports multi core  compilation and hopefully reduces compilation time.

webmaster128: > reduces compilation time

Not without an explicit jobs parameter `/J` or `-j` ;)

neusdan: no need to

> It automatically detects the number of processor cores in your PC by calling QThread::idealThreadCount()

Source: http://blog.qt.io/blog/2009/03/27/speeding-up-visual-c-qt-builds/

webmaster128: Oh, yeah, idd. You are right. The information is still up to date in the current jom source. This is a strange default given that GNU make uses 1 job if not specified otherwise and jom should be a drop-in replacement for GNU make/nmake.

However, we experienced that in virtual environments like Travis, core detection functions like Linux `nproc` can detect some 32 cores while the power of the virtual machine is limited to 1.5 cores. Do we know what idealThreadCount() returns on AppVeyor? [We get 2 cores there](https://www.appveyor.com/docs/build-configuration#build-environment), so maybe we should fix that number.

neusdan: Yeah agree, probably better to set it fix to 2.

mouse07410:
Don't mean to be a pest, but when would this PR be merged?  And what's the timeframe for a more permanent fix (aka adding constructors to and changing behavior of HMAC_DRBG)?

PR #560, opened 2016-07-28
Add X509_Time::to_std_timepoint()

neusdan:
I need a way to calculate the validity days left of a X.509 certificate. For this i've added 

`X509_Time::to_std_timepoint()`

now i can do the follwing:

```c++
Botan::X509_Time currentTime( std::chrono::system_clock::now() );
Botan::X509_Time endTime( cert.end_time(), Botan::ASN1_Tag::UTC_OR_GENERALIZED_TIME );
auto secondsLeft = std::chrono::duration_cast<std::chrono::seconds>( endTime.to_std_timepoint() - currentTime.to_std_timepoint() );
auto certificateDaysValid = static_cast<int32_t>( secondsLeft.count() / 86400U );
```

Or is there an easier way?

### DIFF
--- a/src/lib/asn1/asn1_time.cpp
+++ b/src/lib/asn1/asn1_time.cpp
@@ -250,6 +250,11 @@ bool X509_Time::passes_sanity_check() const
    return true;
    }
 
+std::chrono::system_clock::time_point X509_Time::to_std_timepoint() const
+   {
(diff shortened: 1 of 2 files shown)


randombit:
The ASN.1/DER code is very limited (this has pluses and minuses). I looked and don't see an easier way, and in any case this is a good addition to X509_Time itself. Thanks.

PR #561, opened 2016-07-28
X509_Certificate::allowed_usage fix

neusdan:
in my understanding it should not return true if ex_constraints() returns an empty list

### DIFF
--- a/src/lib/cert/x509/x509cert.cpp
+++ b/src/lib/cert/x509/x509cert.cpp
@@ -265,7 +265,7 @@ bool X509_Certificate::allowed_usage(const std::string& usage) const
    const std::vector<std::string> ex = ex_constraints();
 
    if(ex.empty())
-      return true;
+      return false;
 
    if(std::find(ex.begin(), ex.end(), usage) != ex.end())
       return true;


neusdan:
I have to lookup the test failures, but from the Doxygen documentation:

> Returns true if and only if name (referring to an extended key constraint, eg "PKIX.ServerAuth") is included in the extended key extension.

i do not unterstand why this should return true, if there are no extended key constraints.

randombit:
I think the idea is that if no extended constraint extension is in the cert then there is no particular implication on which usages are allowed/disallowed.

neusdan:
But I want to check if a specific extended key usage is in the certificate. Is there an other way to check this?

randombit:
Examine return value of `ex_constraints` for the desired constraint

neusdan:
yeah right, thx

neusdan:
Can we add a `has_constraint( Key_Constraints usage )` and `has_ex_constraint( const std::string& ex_constraint )`
or something similar?

randombit:
Yeah that seems fine, feel free to submit a PR with those.

PR #566, opened 2016-08-03
Added the possibility to enable or disable ECC

scogliani:
As asked in #551, I added folders for each ECC with an info.txt and some preprocessings checking, allowing the behave of disabling or enabling specific curves. For the moment I just changed **test_ecc_pointmul**. If this approach seems relevant, I will continue for the other test programs.


randombit:
I went back and forth on your approach for some time, but overall it offers excellent control to the builder of the library as to which curves are included without too much extra noise. So if you update the tests to handle missing curves, this change works for me.

randombit:
Closing this due to inactivity and lots of conflicts.

Still would be useful to have this functionality but given the massive refactorings that have happened in EC_Group since 2.4.0 I expect we'll want to go with a different approach.

PR #567, opened 2016-08-04
WIP: SonarQube integration

neusdan:
Closes #524

PR #568, opened 2016-08-04
SonarQube integration

randombit:
Retry of #567

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,9 @@ compiler:
   - gcc
 
 env:
+  global:
+    - secure: IOnsRvoIXYp2Wc+S9TtKNC3qlnDr3bBO9O68TWfmaelOvzdACPYqeBNbUgSEhqHNexujf2lYHT31y8zqvrZacNSWsWYwp6RoBHd9Z5fMG0U6Z50rGwBPMVZTsWJlHGjsXk1EvovCKXOuQq1GvdPIdTUHJsY/Ctg4KEDClEyw9X0=
+  
   matrix:
     - TARGETOS="native" MODULES="all" BOOST="n" BUILD_MODE="shared"
     - TARGETOS="native" MODULES="all" BOOST="n" BUILD_MODE="static"
@@ -97,6 +100,7 @@ install:
 
 script:
   - if [ "${COVERITY_SCAN_BRANCH}" != 1 ]; then ./src/scripts/ci/travis/build.sh ; fi
+  - sonar-scanner -v
 
 after_success:
   - ./src/scripts/ci/travis/after_success.sh
@@ -105,6 +109,8 @@ notifications:
   email: botan-commits@lists.randombit.net
 
 addons:
+  sonarqube: true
+
   apt:
     sources:
     - ubuntu-toolchain-r-test
--- a/.travis.yml
+++ b/.travis.yml
@@ -110,10 +110,6 @@ cache:
 install:
   - ./src/scripts/ci/travis/install.sh
 
-before_script:
-  # for sonarqube:
-  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then jdk_switcher use oraclejdk8 ; fi
-
 script:
   - if [ "${COVERITY_SCAN_BRANCH}" != 1 ]; then ./src/scripts/ci/travis/build.sh ; fi
 
@@ -125,17 +121,6 @@ notifications:
 
 addons:
   sonarqube: true
-
-  apt:
-    packages:
-    - oracle-java8-installer
-    - libssl-dev
-    - libsqlite3-dev
-    - zlib1g-dev
-    - libbz2-dev
-    - liblzma-dev
-    - python
-    - python3
 
   coverity_scan:
     project:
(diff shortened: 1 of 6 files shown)


neusdan:
Interesting, that worked. The addon was activated and the `sonar-scanner` executable was found. Unfortunately the actual scan execution is missing. I've added this in #567. Could you update this PR with the latest commit from my PR? Thanks!

randombit:
Done

randombit:
Need anything from me on this?

neusdan:
It doesn't work because now the java version is too old. But i don't want to disturb you further with this. 
I've requested an own invite from Sonarqube and will try to set everything up on an other project first. Nevertheless thanks so far ;)

neusdan:
Can you update this PR with the latest commits from #567 once again? Thanks!!

randombit:
Done

neusdan:
Finally the analysis is running but there is a problem uploading the results to the SonarQube server:

```
ERROR: Error during SonarQube Scanner execution
ERROR: Insufficient privileges
```

I've reencrypted the SONAR_TOKEN in #567 that you have sent me . Can you update the PR? Thanks!

neusdan:
Damn, same error. Can you check the token or create a new one? Thanks a lot.

randombit:
Just mailed you a new token

neusdan:
Thanks Jack. Updated #567 with the new token. Hopefully this works now.

neusdan:
same error, no idea what's going on there :-(

neusdan:
Changed `sonar.projectKey=Botan` to `sonar.projectKey=botan`. That could be problem. If that doesn't help i contact the support.

neusdan:
It's works! :-)

https://sonarqube.com/dashboard/index?id=botan

I've removed the test execution in #567. With this commit this PR can be reviewed/integrated. 

The analysis will then run for new commits on master and for all pull requests.

.travis.yml
```diff
@@ -24,12 +28,20 @@ compiler:
   - clang
   - gcc
 
+git:
+  depth: 5
+
 env:
+  global:
+    - secure: "VezRbHFg6kllV5WG06M3tG3aHJaC3xrMylJ6RCVbL+uz2JeralVCqV7eIk4fVb9cu83Li+weEa0AJj0wkxpIUJ+vUh5F65L6gSWSbgHP7muOSVsmnEc6KvX4n3av/ZGe4geSmsxqh2pd/2xI1h7KioGRhKeqaZIdjVgWgGJW2iQ="
```
randombit: Are both of these needed? It would be good for there to be a comment re what these tokens are for.

src/scripts/ci/travis/build.sh
```diff
@@ -55,19 +49,50 @@ fi
...
+fi
+
+if [ "$MODULES" != "min" ] && [ "${TARGETOS:0:3}" != "ios" ] && [ "$BUILD_MODE" != "sonarqube" ]; then
     ./botan-test
 fi
 
 if [ "$MODULES" != "min" ] && [ "$BUILD_MODE" = "shared" ] && [ "$TARGETOS" = "native" ]
 then
-    python2 --version
+    python --version
```
randombit: Does Ubuntu 14.04 not have a `python2` alias? If so please put a comment to that effect just above these lines, so someone doesn't come along and 'fix' it to python2 later. Using `python` for Python2 in this script is fine since we know exactly what systems Travis uses.

randombit:
@neusdan Nice! I had a couple minor nits, after fixing those can you squash this to a single commit, and update the commit message to also mention this upgrades Travis CI to Ubuntu 14.04 with the standard system toolchain, and lcov to 1.12 for coverage.

Just open a new PR with the squashed commit.

src/scripts/ci/travis/build.sh
webmaster128: Regular Ubuntu 14.04 installations provide a `/usr/bin/python2` -> `/usr/bin/python2.7` symbolic link in the package python-minimal.

.travis.yml
neusdan: The first is the `SONAR_TOKEN` which is needed to access the SonarQube server. The second is the encrypted `GITHUB_TOKEN`. This is the token of a GitHub user i have created that (in principal, hopefully this works) posts found issues as comments in the PR. This user can also update the PR status if the user has commits rights on the repo, which my user obviously not has. So if you find this feature useful you maybe create an own user and change the `GITHUB_TOKEN` variable.

I'll add a comment which line is which token.

src/scripts/ci/travis/build.sh
randombit: According to the docs https://docs.travis-ci.com/user/trusty-ci-environment/#Python the Travis Trusty image comes preinstalled with CPython 2.6.9, 2.7.10, 3.2.6, 3.3.6, 3.4.3, and 3.50, along with two (old?) versions of PyPy.

It would be great to just run the Python tests under all of them... I'm not sure if there is a way to get at the list of installed Python versions from a script, or if we are doomed to hardcoding `python-3.3.6` and such, and having CI break every time  Travis updates their machine image. Run `dpkg`?

neusdan: I can't remember exactly why i have changed this. I'll change it back to `python2` and we can try again if it works.

randombit:
Closing in favor of #592

PR #572, opened 2016-08-10
Fix x86 32-bit building for clang on Mac OS

Bi11:
-m32 must be set for 32-bit building. -m64 is the default to clang (Xcode) that released in recent a couple of years.

### DIFF
--- a/src/build-data/cc/clang.txt
+++ b/src/build-data/cc/clang.txt
@@ -75,5 +75,6 @@ x86_32  -> "-m32"
 x86_64  -> "-m64"
 ppc64   -> "-m64"
 
+darwin  -> "-stdlib=libc++"
 netbsd  -> "-D_NETBSD_SOURCE"
 </mach_abi_linking>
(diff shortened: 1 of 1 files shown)


Bi11:
With Apple clang, if MACOSX_DEPLOYMENT_TARGET is set to 10.8 or 10.7, -stdlib is default to libstdc++. It will cause a error says: 'cstdint' file not found.

webmaster128:
Shouldn't `-stdlib=` be in the `lang_flags` along with `-std=c++11`?

Bi11:
I don't know whether it should be applied to all Linux.
I can only confirm with Mac OS at this time, and I don't want to break anything else.
We should move `-stdlib` into `lang_flags` if this issue can be confirmed by someone familiar with Linux.

webmaster128:
You're right. I did not see that this affects Linux. mach_abi_linking end up in CXX, so it is fine there.

randombit:
Looks good to me. Thank you for the patch.

PR #573, opened 2016-08-10
Reseed test for HMAC_DRBG

flanfly:
This test verifies that a reseeding takes place after `max_output_before_reseed` bytes have been returned.

/cc @neusdan, @cordney


randombit:
The test failure was AppVeyor having a build get stuck and killed after 60 minute runtime. The other builds from that commit between 6 and 11 minutes, though the stalled build did seem to still be making progress on the build at the point it was terminated. Perhaps it ended up on a very overloaded VM.

Anyway, new test LGTM thank you.

PR #574, opened 2016-08-12
Fix amalgamation build if pkcs11 module is enabled

neusdan:
```
ERROR: Error processing file src\lib\prov\pkcs11\pkcs11.h for amalgamation: No header guard found in src\lib\prov\pkcs11\pkcs11.h
```

Tried first to add the Botan include guard to all three external PKCS11 files, but it's not possible for `pkcs11f.h`. This file must be included multiple times from `pkcs11.h`

Now i've solved it by adding the possibility to mark headers as external: `<header:external>`
These headers are copied/linked into build_dir/include/external
This has the advantage that external includes can be taken as they are, they haven't to be modified.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -123,10 +123,12 @@ class BuildConfigurationInformation(object):
         self.include_dir = os.path.join(self.build_dir, 'include')
         self.botan_include_dir = os.path.join(self.include_dir, 'botan')
         self.internal_include_dir = os.path.join(self.botan_include_dir, 'internal')
+        self.external_include_dir = os.path.join(self.include_dir, 'external')
 
         self.modules = modules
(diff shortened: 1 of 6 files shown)


randombit:
LGTM. IMO this approach makes sense for any other imported system headers we end up using also. I could imagine this being useful also for the TPM (so Trousers is not required at build time but on a system with TPM+Trousers everything just works with a stock build) or the Linux/BSD crypto devices (haven't investigated this much so don't know if it would be required). Any not having to maintain any patches against such external headers would make maintenance easier.

Some repetition in configure.py bugged me so I have tacked one extra commit on your changes in #577, can you take a look?

neusdan:
Closed in favor of #577

PR #577, opened 2016-08-12
Update on PR #574

randombit:
Repetition in configure.py bugged me so tacking on 2e485c3


neusdan:
yeah thats better, thanks :-)

PR #492, opened 2016-05-11
Encrypt then mac with policy

jurajsomorovsky:
Encrypt-then-MAC extension (RFC 7366) 
Extended TLS policy configuration

PR #457, opened 2016-03-23
Reduction of code complexity.

mgierlings:
-reduced number of parameters in various methods
-reduced cyclomatic complexity (McCabe-Metric)
-removed "TLSEXT_HEARTBEAT_SUPPORT" from tls_extensions.h (leftover from heartbeat extension removal?)

# Complexity of Functions, Summary of changes.

## Reduction of argument count.
During code analysis 16 methods requiring too many arguments have been identified. In order to improve the code structure related arguments have been grouped within in objects and passed into the functions as complex type, rather than a list of base types. The following classes have been newly introduced for this purpose:

- TLS::Channel::Callbacks
- TLS::Client::Properties
- TLS::Client_Hello::Settings
- TLS::Client_Properties
- TLS::Server_Hello::Settings
- TLS::Record
- TLS::Record_Raw_Input
- TLS::Record_Message
- X509_CA::Certificate_Properties

The new types improve the readability by replacing abstract types like „string“ with more verbose type names and reducing the overall length of method signatures. 

## Reduction of complexity
3 Methods showed critical code complexity during the evaluation. These methods have been handling multiple tasks, resulting in too long method bodies and a high cyclomatic complexity. To resolve the issues sub tasks have been identified and swapped out to dedicated methods. In detail the following changes have been applied:

### Method „TLS::Channel::received_data
The received_data method handles incoming TLS traffic. Originally all record types have been processed inside this method. The following new methods have been introduced to capsule the processing of different record types:
process_handshake_ccs (TLS-Handshake and -Change Cipher Spec records)
process_application_data
process_alert

### Constructor „TLS::Client“
Prior to refactoring the TLS-Client class constructor performed two crucial operations, the creation of a handshake state object and sending an initial client_hello message. Both actions were capsuled and swapped out into dedicated methods.

### Method ___Globals.tls_client(int, char**)
This method has been removed.

PR #578, opened 2016-08-13
Updates for PR #492

randombit:
@CheariX @jurajsomorovsky I've given a final review to #492 and with some changes things are ready to merge, can you please review my additional commit e290246? I would like your +1s on my changes, in particular my adding copyright statements to the files you changed in your commits.

If possible I would like to merge this to master within the next week or so. The actual record layer changes may end up conflicting with #457 depending on which lands first, but I will handle the merge.


jurajsomorovsky:
+1
I rechecked the compatibility with OpenSSL and the RSA key handling, and executed some fuzzing. Everything seems to work well...

Thank you!

CheariX:
+1

Thank you!

Issue #167, opened 2015-07-03
Should any algorithms be removed prior to 2.0?


randombit:
Some of the algorithms in botan are obscure, obsolete, or outright broken and we should consider removing them (as has been done in past releases for various also-ran algorithms of the 90s like SHARK and HAVAL). Doing so prior to the 2.0 release would be the logical choice.

The maintenance burden of having these algorithms available is not high (really, almost zero), but it creates additional confusion for the user about what they might want to use, and by the very fact it is implemented and available at all leaves it possible for an application to potentially be tricked into using an insecure algorithm. Botan may be a kitchen sink, but it ideally would be a highly focused kitchen sink containing only the best available crypto primitives for any purpose.

Looking things over, the short list for removal would be Nyberg-Rueppel signatures, MARS, RC2, RC5, RC6. HAS-160, TEA, XTEA, SAFER, Lion, MD2, MD4, CBC-MAC, and X9.19-MAC. MD5 would be on the list too, but is still just too commonly used, and I'm on the fence about TEA/XTEA since they are somewhat common also and not (AFAIK) broken.

Are there good reasons to keep any of these algorithms? For example are there protocols or formats that would be useful to support which make use of any of them? And are there additional algorithms that we should consider for removal?

mouse07410:
I say NO - we do not remove anything unless there is a good reason to. I see no “additional confusion for the user” at all.


On Jul 3, 2015, at 13:52 , Jack Lloyd <notifications@github.com<mailto:notifications@github.com>> wrote:


Some of the algorithms in botan are obscure, obsolete, or outright broken and we should consider removing them (as has been done in past releases for various also-ran algorithms of the 90s like SHARK and HAVAL). Doing so prior to the 2.0 release would be the logical choice.

The maintenance burden of having these algorithms available is not high (really, almost zero), but it creates additional confusion for the user about what they might want to use, and by the very fact it is implemented and available at all leaves it possible for an application to potentially be tricked into using an insecure algorithm. Botan may be a kitchen sink, but it ideally would be a highly focused kitchen sink containing only the best available crypto primitives for any purpose.

[...]

PR #580, opened 2016-08-14
Remove deprecated algorithms

randombit:
Removes MARS, RC2, RC5, RC6, SAFER-SK, TEA, MD2, HAS-160, RIPEMD-128, Nyberg-Rueppel and Rabin-Williams. Some of these are marginally broken, and all are rarely if ever used.

XTEA was deprecated but I decided it is common enough to keep for now. The XTEA SIMD impl was tested to be no faster than the scalar code (perhaps due to autovectorization, I didn't check) so that was removed.

Removes IF_Scheme_PublicKey/IF_Scheme_PrivateKey which was a common base class of RSA and Rabin-Williams. All members move to RSA.


mouse07410:
I don't think this is a good idea, because old data may be encrypted by these "deprecated" algorithms. 

It is perfectly OK to remove them from the *protocols* (like TLS and SSH), so no communicator would be able to negotiate their use *now* with a remote peer over the network. But it is not OK to risk making an old backup suddenly unusable because the algorithm it was encrypted with is deemed "rarely used".

randombit:
> But it is not OK to risk making an old backup suddenly unusable

And if there was any demonstrated usage of such, that would be a consideration. But these algos were publicly deprecated 5 months ago, and nobody has mentioned any actual usage.

> Rabin-Williams signature has advantages over RSA

It does have advantages over RSA, but IMO those advantages are not worth the additional code and complexity it brings to the library as a whole. And at this point integer factorization based schemes are dead - nobody is going to start adopting RW signatures now.

mouse07410:
> And if there was any demonstrated usage of such, that would be a consideration. But these algos were publicly deprecated 5 months ago, and nobody has mentioned any actual usage.

This is the problem with Data At Rest vs. actual online protocols. I'm sure few if any are using those deprecated algorithms *now*. I know I don't (and haven't been using them for several years at least). But I can't be sure about *all* of my backup files and old pieces of weirdly-encrypted emails stored somewhere. And IMHO nobody can be expected to go over *all* of his storage media to check every single file.

randombit:
> But I can't be sure about all of my backup files and old pieces of weirdly-encrypted emails stored somewhere.

But the problem surely is not as bad as this. If we can decide that no piece of software using botan actually used (say) MARS, then it is safe to say that no user could be affected by MARS being removed. They might well happen to have some old ciphertext around encrypted with MARS, but if they never in the first place had a tool using botan that actually used botan's MARS implementation to process it then... well they probably would not notice MARS being removed from the library. They should just continue to use whatever software they used before, to process these odd MARS ciphertexts.

In some senses I'm very sympathetic to your argument. Removing an algorithm entirely breaks compatibility (for users of those algorithms) in a very serious way. And to be clear, if it turned out there was some actively maintained/used software using botan which used MARS (or any of the others), I would reverse course. But, no such user has been found after repeated advertisements in the release notes and mailing list.

Also I'll note that from a project maintenance standpoint, your proposal draws an impossibly hard line that no algorithm can ever be removed, since we could not prove it was not being used somewhere. But, that is death for project agility - if it is impossible to remove an algorithm the bar for entry of a new algo also becomes impossibly high.

mouse07410:
> If we can decide that no piece of software using botan actually used (say) MARS, then it is safe to say that no user could be affected by MARS being removed.

The problem is - how can we know? Archived backup files may be decade(s) old and not online, unaccessible until really needed. Software changes all the time - possibly the original software cannot run (probably does not exist) any more. New software came to take its place, and we hope it would handle the task inherited from the retired one. 

> ...they probably would not notice MARS being removed from the library. 

Correct!! The only case they'd notice it is if they have to dig out a 15-years old TAR file that was encrypted by some code that does not exist any more (and it's very likely that the OS it ran on cannot run either!). In that case, Botan and Crypto++ would probably be the only alternatives (I don't recall how wide the algorithm coverage of OpenSSL is).

> They should just continue to use whatever software they used before, to process these odd MARS cipher texts.

You probably know better than I do that while archives tend to live forever (thankfully, most disappear without ever being needed :), but the processing software recycles rapidly. As I said above - the most probable case is that not only the original software that produced those archives, but even the OS it ran on are no longer available. 

> ...from a project maintenance standpoint, your proposal draws an impossibly hard line that no algorithm can ever be removed... 

That's the problem with implementing a successful cryptographic library. You cannot arbitrarily drop "least used" algorithms (Microsoft used to every year fire those in the bottom 5% of performance evaluation - until found that it caused undesirable results :).  So the library would grow. The rate that new algorithms are added thankfully is very low - we don't have to worry about it taking too much space any time soon.

Partially this is your own fault :), as your library moved from barely existing on the radar to an excellent, popular, widely used, *and usable* cryptographic product applicable to many areas and tasks. As this transformation happened - more people became dependent on it. Probably still less than on OpenSSL, but still enough to give a pause to a thought of reducing the scope.

> But, that is death for project agility - if it is impossible to remove an algorithm the bar for entry of a new algo also becomes impossibly high.

You don't have a hard limit on the number of algorithms used. A company might think that in order to hire new experts they have to fire those no longer useful, in order to maintain the total headcount cap. But a library? I'm sure you're contemplating addition of Argon2. Does it mean that you have to remove one of the existing PBKDF to make room?

I'm sure there would be PQ algorithms other than McEliece. Does it mean that RSA would have to go to make room for them? RSA won't have any online use, but there would still be plenty of CMS left around secured by it...

I'm sympathetic to your argument too, but I think you're overestimating the cost of maintaining old algorithms. Once they're done (code, test-vectors, etc.) - they're done and require almost nothing any more, especially if you drop support for their *optimized* versions.

As I said before, It makes sense to reduce the options for the online communications, lest the peers are tricked into using marginally secure crypto (so the ultimate reason for pruning is security, not maintenance or library size). But this case is different.

P.S. My apologies for being too verbose - but I've had a tiring day, so can't be as concise as I'd like to.

cordney:
Aside from the fact that we already had the same discussion in #167, some notes from me:

> Correct!! The only case they'd notice it is if they have to dig out a 15-years old TAR file that was
> encrypted by some code that does not exist any more (and it's very likely that the OS it ran on
> cannot run either!). In that case, Botan and Crypto++ would probably be the only alternatives (I don't 
> recall how wide the algorithm coverage of OpenSSL is).

What about using an older version of Botan or Crypto++ then? Compilers almost never break existing code, so Botan 1.11.30 can very likely also sucessfully be compiled with a GCC 15.

> I'm sympathetic to your argument too, but I think you're overestimating the cost of maintaining old
> algorithms. Once they're done (code, test-vectors, etc.) - they're done and require almost nothing
> any more, especially if you drop support for their optimized versions.

It's not as simple as that in the software world, unfortunately. A good example is in this PR even: You have to keep base class interfaces that otherwise could be removed and thus code can not be simplified, as is the case with IF_Scheme_PublicKey/IF_Scheme_PrivateKey. Another aspect: When refactoring code, you have to carry the deprecated code with you, that is, you will very likely have to make changes in this code, too. And this extra effort needed may even keep you from refactoring code, e.g., because you are not willing to go that extra mile. This can lead to more code clutter over time.

Unfortunately so to say, I know what I'm talking of. In-house, we manage about 6-8 C++ libraries for almost 10 years now that all teams use in their client products and we constantly discuss the topic of API/ABI compatibility.

randombit:
> When refactoring code, you have to carry the deprecated code with you, that is, you will very likely have to make changes in this code, too. And this extra effort needed may even keep you from refactoring code, e.g., because you are not willing to go that extra mile.

Indeed. One reason for submitting this PR now is to start clearing the decks for addressing some wide-ranging issues around static initialization. Removing as many algos beforehand will reduce the amount of work required to perform these changes. Reducing the amount of time such changes take dramatically increases the chances of them being successfully performed, because I have only very limited time available for working on this project.

mouse07410:
> What about using an older version of Botan or Crypto++ then? Compilers almost never break existing code, so Botan 1.11.30 can very likely also sucessfully be compiled with a GCC 15.

In my experience, compilers consistently break the existing code if it is old "enough". System include files change, supporting libraries change, language dialects change, etc. While I'm sure that simple programs would be compilable forever (after all, we still use Ritchie's "Hello World" :), more complex programs (like Botan library, for example) are IMHO less likely to stay future-proof.

> A good example is in this PR even: You have to keep base class interfaces that otherwise could be removed and thus code can not be simplified, as is the case with IF_Scheme_PublicKey/IF_Scheme_PrivateKey. 

True. But I'd personally prefer that it is left alone rather than simplified. Especially since I use Rabin-Williams signatures.

> Another aspect: When refactoring code, you have to carry the deprecated code with you, that is, you will very likely have to make changes in this code, too.

True. But usually (IMHO and experience, at least) we try to preserve functionality when refactoring. Also, in crypto - what's "deprecated"? 

And when something has been done - usually I don't tinker with it without a reason, and then I evaluate how much work would the refactoring cost, and what benefit it would bring. In one of my current project, I decided *not* to refactor - because the complexity of the change would be too big for the comfort, and the possible improvement was not necessary - the existing code is good "enough" as is.

randombit:
> RSA won't have any online use, but there would still be plenty of CMS left around secured by it...

Right, and being able to verify all those old RSA signatures will be a plenty good reason to keep RSA implemented for a long long time. Point me to a single Rabin-Williams signature anywhere on the Internet.

> Also, in crypto - what's "deprecated"?

I'm sure answering that in the case of crypto standards in general is an involved and nuanced conversation. One which fortunately can happen elsewhere.

In the case of whats deprecated in the botan c++ library, it is probably the same as deprecated code in any other software library: code that the maintainers plan to (or at least hope to) remove, because the the cost the code incurs in ongoing maintenance exceeds its usefulness.

> Especially since I use Rabin-Williams signatures.

You may want to expand on this.

mouse07410:
> Point me to a single Rabin-Williams signature anywhere on the Internet.

Not aware of any. I didn't have in mind online usage.

>> Especially since I use Rabin-Williams signatures.
>
> You may want to expand on this.

Well, one of my projects uses Rabin-Williams (faster than RSA, avoids ECC) for integrity and non-repudiation in crypto filesystem... Not much to say... I'd probably just used ECC, but there were non-technical reasons against ECC in that particular case.

cordney:
On the content side: Don't forget to update the module policy files (bsi/modern).

randombit:
@cordney Thanks for the reminder, I edited them for the block cipher removals but not the hashes or signature schemes.

randombit:
One potential problem with removing MD2, is that MD2 self-signed trust roots will no longer verify as (IIRC) even the trust root's self-signature is verified during path validation. I'm not sure how common a MD2 CA cert is anymore, one would hope CAs have created new SHA-1 or SHA-256 certs for the same keys by now.

mouse07410:
Also, (I did not pay attention before) I noticed that [Macports](http://www.macports.org) uses RIPEMD160 for all of their ports. That's large.

randombit:
@mouse07410 Only RIPEMD-128 is being removed, RIPEMD-160 is relatively common and IIRC not even broken yet.

randombit:
I'm at the point of starting to work on some of the changes to address #465 for 1.11.32 - changes which are going to involve touching nearly every file in the tree. So I am going to be merging this PR now, to reduce the amount of work I have to do. No point updating code just to rip it out again later. ;)

Issue #477, opened 2016-04-18
RFC: Using AES_128 type directly is insecure


randombit:
A reasonable programmer might read the public headers aes.h and gcm.h, write

```
// Don't need the flexibility of runtime lookup as our protocol is fixed to aes-128-gcm
GCM_Encryption gcm(new AES_128);
```

and reasonably expect it to be secure. However because AES_128 specifically refers to the lame (slow, side channel ridden) T-box implementation, rather than the (very fast, constant time) AES-NI, or (still constant time, and faster than lookups) SSSE3 version, or calling the OpenSSL version, or any future version that might have similar side channel secured implementations such as for NEON or ARM v8. Instead they must use `BlockCipher::create("AES-128")`, or perform the equivalent build and CPUID check themselves.

I recall mentioning this in an Issue and at least one person who had some experience using the library expressed surprise at this state of affairs. And in fact it is surprising, in that no other crypto library (AFAIK) exposes more than one implementation in this way, with the safe version hidden away - so this design seems likely to break user expectations. Perhaps if their application was not processing large amounts of data, they might not even notice the poor performance and investigate if their application was using AES-NI properly.

Maybe it would be better to merge all implementations into just one public type (AES_128) which does the best thing available?

This same problem occurs for the ciphers with SIMD versions (Threefish, Serpent, Noekeon, IDEA) and also the SHA-1 SSE2 code; though there the problem is 'just' a reduction in performance since the baseline versions are already constant time. The base versions should just do a ifdef check followed by a cpuid check, and directly call the SIMD implementation if possible. Some code, such as GCM's use of clmul, already follows this pattern.

The current way of having completely distinct object types for each implementation does ease testing and performance comparisons. But this convenience does not make up for a poor public API which encourages use which will prevent use of optimizations, or open up side channels.

In the near term testing can be addressed by adding a CI instance that just runs the baseline algorithms (perhaps by faking CPUID?) seems sufficient. Right now we have cross compiling ARM builds in CI, but running the tests in qemu or iOS simulator is not set up.

This merger would seem to obviate almost all uses of the `provider` argument in the lookup code. When the choices are "slower/vuln to side channel" vs "faster/constant time" the choice seems obvious.

cordney:
[...]

PR #582, opened 2016-08-15
Add sha1_sse2 to the TLS module dependencies

neusdan:
When Botan is configured with the BSI module policy and additionally the TLS module is enabled the `sha1_sse2` module stays disabled.

The `sha1` and `sha1_sse2` module are both not mentioned in the BSI policy file, so they are allowed to load if a dependency needs them.

### DIFF
--- a/src/lib/tls/info.txt
+++ b/src/lib/tls/info.txt
@@ -48,8 +48,7 @@ par_hash
 prf_tls
 rng
 rsa
-sha1
-sha1_sse2
+sha1_sse2|sha1
 sha2_32
 x509
 </requires>
--- a/src/lib/tls/info.txt
+++ b/src/lib/tls/info.txt
@@ -49,6 +49,7 @@ prf_tls
 rng
 rsa
 sha1
+sha1_sse2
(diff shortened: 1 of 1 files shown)


randombit:
I see what you are trying to do but this is not quite right - it would prevent enabling TLS at all on non-x86, because the required module would not be available.

neusdan:
Ok right. So is there any other solution?

Does the following work?

```
<requires>
sha1_sse2|sha1
</requires>
```

randombit:
> Does the following work?

It should. But with #549 landed no other info files requires this alternative module selection logic, and I'd planned on removing it from configure.py entirely.

I think as part of fixing issues #477 and #465 this distinction is going to go away entirely and everything will just work; if you enabled SHA-1, and if your CPU has/might have SSE2, then you get SHA-1 with SSE2. In the short term I guess the `|` logic can stay in configure.py and the dep can change as you suggest.

neusdan:
Updated the PR and look forward to the fixes for #477 and #465 :-)

PR #583, opened 2016-08-15
Clean up TLS ciphersuite handling

randombit:
Been meaning to do this for a while.

### DIFF
--- a/src/lib/tls/tls_ciphersuite.cpp
+++ b/src/lib/tls/tls_ciphersuite.cpp
@@ -11,42 +11,11 @@
 #include <botan/stream_cipher.h>
 #include <botan/hash.h>
 #include <botan/mac.h>
-#include <sstream>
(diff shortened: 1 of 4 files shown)


randombit:
Some kind of problem with MSVC and static libraries, `botan-test` seems to be crashing at startup.

randombit:
Making the `std::vector<Ciphersuite>` a function-level static instead of a global static const seems to be sufficient to fix the problem I saw with MSVC. Commits rebased to clean up history, I think this is ready to check in once CI completes.

randombit:
For the record I still don't understand why having the Ciphersuite list as a global static cons didn't work on Windows in static mode, considering it is struct considering only of `const char*`s to literal strings, some ints, and a bool. Maybe an ordering issue between tls_ciphersuite.o and tls_suite_info.o

Issue #432, opened 2016-02-21
visual studio botan lib pdb is missing  after compilation.


agrandville:
The makefile generate it but it's immediatly replaced when compiling the botan.exe. 
I've found a workaround changing the app output file from botan.exe to botan-cli.exe in the Makefile (CLI = .\botan-cli.exe).

randombit:
Oh, unfortunate.

I wonder if there is some command line flag to tell MSVC to output the PDB for the CLI file to some other location to avoid the overwrite, or to skip it entirely. The PDB for the command line is probably not particularly useful for most people.

neusdan:
There is a compiler option `/Fd` 

https://msdn.microsoft.com/en-us/library/9wst99a9.aspx

and a linker option `/PDB`

https://msdn.microsoft.com/en-us/library/kwx19e36.aspx

not sure which is the right one.

agrandville:
I think /Fd flag is used to specify the output folder during compilation, the /PDB file locate the final pdb file.  
I do some test and obtain the best result with this change in my Makefile  
LIB_LINK_CMD   = $(LINKER) /DLL /DEBUG /PDB:botanlib.pdb

neusdan:
Can be closed now.

randombit:
Thanks

PR #584, opened 2016-08-16
No PDB output for CLI and tests

neusdan:
Fixes GH #432 : Previously the Botan CLI PDB file has overwritten the Botan lib PDB file.

Furthermore the output filename of the lib is changed to `botand` in debug mode.

### DIFF
--- a/src/build-data/cc/msvc.txt
+++ b/src/build-data/cc/msvc.txt
@@ -48,7 +48,7 @@ default-debug -> "$(LINKER) /DLL /DEBUG"
 
 <binary_link_commands>
 default       -> "$(LINKER)"
-default-debug -> "$(LINKER) /DEBUG"
+default-debug -> "$(LINKER)"
 </binary_link_commands>
 
 <mach_abi_linking>
--- a/src/build-data/cc/msvc.txt
+++ b/src/build-data/cc/msvc.txt
@@ -48,7 +48,7 @@ default-debug -> "$(LINKER) /DLL /DEBUG"
 
 <binary_link_commands>
 default       -> "$(LINKER)"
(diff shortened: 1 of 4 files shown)


src/build-data/cc/msvc.txt
```diff
@@ -48,7 +48,7 @@ default-debug -> "$(LINKER) /DLL /DEBUG"
 
 <binary_link_commands>
 default       -> "$(LINKER)"
-default-debug -> "$(LINKER) /DEBUG"
```
webmaster128: Don't we need that to debug tests and cli as well as benchmark stuff?

neusdan: Hmm, i don't know. Most people probably not. I'm debugging Botan most of the time directly in Visual Studio with a VS-Solution generated by Bakefile so i don't need the other PDB files.

An other solution to this problem is the one suggested by @agrandville in #432 by changing the filename of the CLI tool to botan-cli.exe. I'm happy with both ways.

webmaster128: I added `/DEBUG` in order to be able to run the CLI and tests in the Qt Creator debugger. Having different filenames sounds more solid to me.

randombit:
I have no opinion here

@neusdan + @webmaster128 are you both good with this PR being merged now?

webmaster128:
I did not need the feature for some time, but I think having debug symbols for CLI and tests is a useful thing. So I'd rename botan.exe to botan-cli.exe. This is more consistent, explicit and solid. The cli file name could then be renamed in a custom deployment step.

neusdan:
Yeah fine. Changed the filename instead of no PDB output ;)

PR #586, opened 2016-08-16
Remove configure extra information from build.h

webmaster128:
Removes user and hostname: as mentioned by a downstream maintainer, the information should not be included in packages. In our projects, the information is contained in the amalgamation check-in commit message.

Removes timestamp: this ensures that the header content is equal in two equal configure runs.

Both changes together make it possible to effectively use compiler caches for Botan (ccache on Unix, in direct mode; clcache on Windows).

Closes #587.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -4,7 +4,7 @@
 Configuration program for botan
 
 (C) 2009,2010,2011,2012,2013,2014,2015 Jack Lloyd
-(C) 2015 Simon Warta (Kullo GmbH)
+(C) 2015,2016 Simon Warta (Kullo GmbH)
 
 Botan is released under the Simplified BSD License (see license.txt)
 
@@ -29,7 +29,6 @@ import shutil
 import string
 import subprocess
 import logging
-import getpass
 import time
 import errno
 import optparse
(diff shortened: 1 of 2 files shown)


lamby:
:+1: for removing the infomation entirely.

webmaster128:
From a caching point of view, I think stable header files should be the default setting. That would also help Botan's CI, where ccache is in place but does not hit a lot.

randombit:
Yeah this would be one improvement towards having a fully deterministic build which is a worthy goal for sure.

The associated variables are probably not used anywhere else, so the logic to export them can be removed from `configure.py`. (And doing so would assure that no uses remained)

webmaster128:
Thanks for positive feedback.

I am going to delete the data sources in this PR.

randombit:
LGTM

randombit:
BTW on my desktop machine this improved the time of a clean reconfigure+`make -j12` with a fully populated ccache from ~4 seconds to under 1 second, due to ccache direct vs preprocessed mode.

webmaster128:
:) I'm glad we talked about it. Thanks all

lamby:
Thanks :)

PR #587, opened 2016-08-16
Make the build reproducible.

lamby:
If SOURCE_DATE_EPOCH[0] is set, either use the value or use "fixed" values
for username/hostname/timestamp.

 [0] https://reproducible-builds.org/specs/source-date-epoch/

Signed-off-by: Chris Lamb <chris@chris-lamb.co.uk>

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -202,13 +202,17 @@ class BuildConfigurationInformation(object):
         return 'botan-%d.%d.pc' % (self.version_major, self.version_minor)
 
     def username(self):
+        if 'SOURCE_DATE_EPOCH' in os.environ:
+            return 'generic-username'
         return getpass.getuser()
 
     def hostname(self):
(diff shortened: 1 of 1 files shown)


lamby:
Filed in Debian as https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=834549

webmaster128:
See also #586

lamby:
:+1: for removing the infomation entirely; it's sometimes easier to suggest "fixing" it rather than removing :)

randombit:
The fields should just be removed as in #586 so deterministic build is the default behavior, so closing in favor of that PR.

Thanks for raising the issue, and for working on the deterministic builds for Debian.

lamby:
> Thanks for raising the issue, and for working on the deterministic builds for Debian.

No problem - thank you :)

PR #590, opened 2016-08-17
[Ready] Cilk/OpenMP parallelism

randombit:
Adds flags to enable Cilk+ (GCC) or OpenMP (GCC or Clang), `BOTAN_PARALLEL_FOR` and some other (currently unused) Cilk and OpenMP constructs. I think the OpenMP constructs I'm using are not supported by Visual C++. They probably are supported by Intel C++, but I haven't tested this.

Rewrites some block cipher loops to be fully parallel, avoiding loop dependencies like updating a shared variable. This is a good idea anyway, because it makes the loop easier to reason about. For plain ECB the speedups are great with either OpenMP or Cilk+, for XTS/OCB modes the performance is less clear. I'll have real numbers before merging this.

Changes RSA to use a parallel loop instead of `std::async`. However this means RSA performance takes a substantial hit on 'plain' builds, which is not great. I'm not sure about this change. But, in general starting a new thread per simple operation like a signature is not a good idea, especially on platforms where thread creation is expensive.

I would not really expect someone would build Botan especially to use Cilk+ or OpenMP for the increased performance, especially as it requires the application be linked with special flags/libraries. It seems more likely that some application already using OpenMP/Cilk for parallelism would enable this in their amalgamation build.

TODO: Add CI builds (Clang/OpenMP and GCC/Cilk at least)
TODO: Can other uses of std::async be replaced?
TODO: Karatsuba using fork-join (I think a larger workspace is required to do this, which changes the calling convention for this function)
TODO: Parallel ECC scalar multiplication

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -323,6 +323,11 @@ def process_command_line(args):
     build_group.add_option('--with-external-includedir', metavar='DIR', default='',
                            help='use DIR for external includes')
 
+    build_group.add_option('--with-openmp', default=False, action='store_true',
+                           help='enable use of OpenMP')
+    build_group.add_option('--with-cilkplus', default=False, action='store_true',
+                           help='enable use of Cilk Plus')
+
     link_methods = ['symlink', 'hardlink', 'copy']
     build_group.add_option('--link-method', default=None, metavar='METHOD',
                            choices=link_methods,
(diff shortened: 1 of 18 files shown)


randombit:
I've reverted the RSA changes (not great results there) and am leaving the rest I mentioned (ECC, Karatsuba, etc) to the future. What is left is good to merge now I think, and can be used as a base for further work later.

Issue #550, opened 2016-07-22
EMSA1 verification


neusdan:
I'm trying to understand what's happening inside the `EMSA1::verify` function:
https://github.com/randombit/botan/blob/master/src/lib/pk_pad/emsa1/emsa1.cpp#L67-L96

``` c++
bool EMSA1::verify(const secure_vector<byte>& coded,
                   const secure_vector<byte>& raw, size_t key_bits)
   {
   try {
      if(raw.size() != m_hash->output_length())
         throw Encoding_Error("EMSA1::encoding_of: Invalid size for input");

      secure_vector<byte> our_coding = emsa1_encoding(raw, key_bits);

      if(our_coding == coded) return true;
      if(our_coding.empty() || our_coding[0] != 0) return false;
      if(our_coding.size() <= coded.size()) return false;

      size_t offset = 0;
      while(offset < our_coding.size() && our_coding[offset] == 0)
         ++offset;
      if(our_coding.size() - offset != coded.size())
         return false;

      for(size_t j = 0; j != coded.size(); ++j)
         if(coded[j] != our_coding[j+offset])
            return false;

      return true;
      }
   catch(Invalid_Argument)
      {
      return false;
      }
   }
```

In IEEE 1363-2000 the verification operation is described as:
1. Use the Encoding Operation of EMSA1 to compute a message representative g, a non-negative
   integer. If the encoding operation outputs “error,” output “invalid” and stop.
2. If f = g, output “valid”; otherwise, output “invalid.”

So in my understanding following code should be sufficient:

``` c++
bool EMSA1::verify(const secure_vector<byte>& coded,
                   const secure_vector<byte>& raw, size_t key_bits)
   {
   try
      {
      return coded == emsa1_encoding(raw, key_bits);
      }
   catch(Encoding_Error)
      {
      return false;
      }
   }
```

I could only find one test case that's not working with my changed code. It's the first testcase from Nyberg-Rueppel where the message is `FCABA2CF434AC8DCFCEC`.

The SHA-1 hash of this message is `0000000069113490a6b0793f802c7810bb16e943`. These leading zeros are not returned from NR_Verification_Operation::verify_mr() and i think that's the bug here.

randombit:
[...]

PR #425, opened 2016-02-17
Fix decoding of DECIPHER_ONLY Key_Constraints bit

Mironenko:
Key usage decipherOnly bit used to be decoded into DIGITAL_SIGNATURE
Key_Constraints value.

PR #591, opened 2016-08-17
Improve X.509 tests

cordney:
This PR originally started with trying to improve the X.509 unit tests, but along the way we needed to fix a few bugs, so this PR now contains:

* A fix for #425. Adds a test to `test_x509_cert()`. As the code was moved to the Extensions code, it also removes the now unused code from `key_constraint.cpp`.
* When adding the test for #425, we discovered an issue in `X509Cert::allowed_usage(Key_Constraints)`:
The last line currently reads `return ((constraints() & usage) != 0);` where we think it should be 
`return ((constraints() & usage) == usage);` instead. We wrote unit tests for this function and could thereby confirm that the current version would return `true` if the cert has only constraint A and a user called `allowed_usage(A | B)`, whereas the latter version would return `false` in this case.
* We also refactored the X.509 unit tests to run with RSA, DSA, ECDSA, ECGDSA and ECKCDSA signature algorithms, previously they would only run with ECDSA, if available, or RSA if available, or DSA, if available (in this order). Occasionally, we saw faillures with the
X.509 unit tests and ECKCDSA ("strange signature size"), for which #550 seems to be the cause, so a fix is included in this PR for DSA, ECDSA, ECGDSA and ECKCDSA.

When debugging code we stumbled upon `find_constraints()` in `key_constraint.cpp`. This one is called during certificate creation, it's purpose is to match the key type with suitable constraints. I find it odd though that it removes all constraints that do not match the key type silently, without letting the caller know. For example, when trying to create a certificate with an ECDSA key and the `KEY_ENCIPHERMENT` key usage, the caller would get an exception about "encoding of an empty key constraints extension not allowed". But when trying to create a certificate with an ECDSA key and the `KEY_ENCIPHERMENT` and `DIGITAL_SIGNATURE` key usage, `find_constraints()` would only return `DIGITAL_SIGNATURE` and the caller would silently get a certificate with only the `DIGITAL_SIGNATURE` usage set. Shouldn't it throw an exception in this case?

### DIFF
--- a/src/lib/cert/x509/key_constraint.cpp
+++ b/src/lib/cert/x509/key_constraint.cpp
@@ -29,7 +29,7 @@ Key_Constraints find_constraints(const Public_Key& pub_key,
 
    if(name == "RSA" || name == "RW" || name == "NR" ||
       name == "DSA" || name == "ECDSA" || name == "ECGDSA" || name == "ECKCDSA")
-      constraints |= DIGITAL_SIGNATURE | NON_REPUDIATION;
+      constraints |= DIGITAL_SIGNATURE | NON_REPUDIATION | KEY_CERT_SIGN | CRL_SIGN;
 
    if(limits)
       constraints &= limits;
--- a/src/lib/cert/x509/key_constraint.cpp
+++ b/src/lib/cert/x509/key_constraint.cpp
@@ -12,29 +12,35 @@
 namespace Botan {
 
 /*
-* Find the allowable key constraints
+* Make sure the given key constraints are permitted for the given key type
 */
-Key_Constraints find_constraints(const Public_Key& pub_key,
-                                 Key_Constraints limits)
+void verify_cert_constraints_valid_for_key_type(const Public_Key& pub_key,
+                                                      Key_Constraints constraints)
    {
    const std::string name = pub_key.algo_name();
 
-   size_t constraints = 0;
+   size_t permitted = 0;
 
    if(name == "DH" || name == "ECDH")
-      constraints |= KEY_AGREEMENT;
+      {
+      permitted |= KEY_AGREEMENT | ENCIPHER_ONLY | DECIPHER_ONLY;
+      }
 
    if(name == "RSA" || name == "ElGamal")
-      constraints |= KEY_ENCIPHERMENT | DATA_ENCIPHERMENT | ENCIPHER_ONLY | DECIPHER_ONLY;
+      {
+      permitted |= KEY_ENCIPHERMENT | DATA_ENCIPHERMENT;
+      }
 
    if(name == "RSA" || name == "RW" || name == "NR" ||
       name == "DSA" || name == "ECDSA" || name == "ECGDSA" || name == "ECKCDSA")
-      constraints |= DIGITAL_SIGNATURE | NON_REPUDIATION | KEY_CERT_SIGN | CRL_SIGN;
+      {
+      permitted |= DIGITAL_SIGNATURE | NON_REPUDIATION | KEY_CERT_SIGN | CRL_SIGN;
+      }
 
-   if(limits)
-      constraints &= limits;
-
-   return Key_Constraints(constraints);
+   if ( ( constraints & permitted ) != constraints )
+      {
+      throw Exception("Constraint not permitted for key type " + name);
+      }
    }
 
 }
--- a/src/lib/cert/x509/key_constraint.cpp
+++ b/src/lib/cert/x509/key_constraint.cpp
@@ -1,44 +1,15 @@
 /*
 * KeyUsage
 * (C) 1999-2007 Jack Lloyd
+* (C) 2016 René Korthaus, Rohde & Schwarz Cybersecurity
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
 
 #include <botan/key_constraint.h>
 #include <botan/x509_key.h>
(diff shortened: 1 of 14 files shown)


randombit:
Thanks, reviewing now.

I mailed MITRE on Sunday for a CVE for the `allowed_usage` bug, have not heard back from them... it is not super critical IMO but certainly security relevant.

> I find it odd though that it removes all constraints that do not match the key type silently, without letting the caller know. 

This is bogus, when creating a new cert the user should either get exactly what they requested or an exception explaining why the library cannot or will not create such a cert, rather than something with subtle modifications.

src/lib/cert/x509/key_constraint.cpp
```diff
@@ -54,11 +25,11 @@ Key_Constraints find_constraints(const Public_Key& pub_key,
...
 
    if(name == "RSA" || name == "ElGamal")
-      constraints |= KEY_ENCIPHERMENT | DATA_ENCIPHERMENT;
+      constraints |= KEY_ENCIPHERMENT | DATA_ENCIPHERMENT | ENCIPHER_ONLY | DECIPHER_ONLY;
 
    if(name == "RSA" || name == "RW" || name == "NR" ||
-      name == "DSA" || name == "ECDSA")
-      constraints |= DIGITAL_SIGNATURE | NON_REPUDIATION;
+      name == "DSA" || name == "ECDSA" || name == "ECGDSA" || name == "ECKCDSA")
+      constraints |= DIGITAL_SIGNATURE | NON_REPUDIATION | KEY_CERT_SIGN | CRL_SIGN;
```
randombit: This causes all end entity certificates to have these constraints set:

./botan keygen > rsa.priv  
$ ./botan gen_self_signed rsa.priv localhost | openssl x509 -text | grep -A1 "Key Usage"
            X509v3 Key Usage: critical
                Digital Signature, Non Repudiation, Key Encipherment, Data Encipherment, **Certificate Sign, CRL Sign, Encipher Only, Decipher Only**

vs on master:
            X509v3 Key Usage: critical
                Digital Signature, Non Repudiation, Key Encipherment, Data Encipherment

Also `find_constraints` is only called for non-CA keys, CA keys are forced to exactly KEY_CERT_SIGN | CRL_SIGN.

`find_constraints` is a public function but it is bogus and only has 3 callers internally (in x509_ca.cpp and x509self.cpp), perhaps there is some other approach where user can better control how this field is set.

cordney: I would rename it to `match_constraints()` and have it throw an exception if limits contains a constraint that it not allowed for the key type, and return `void` otherwise, so lines 34ff. would become:

```
   if ( ( limits & constraints ) != limits )
      {
      throw Exception("Constraint not permitted for this key type");
      }
```

That would change the behaviour of commands like `gen_self_signed` such that all end entity certificates would contain no constraints, if not explicitly given, which seems correct to me. Would you support such an implementation?

randombit: @cordney That change works for me. If no constraints are set, the Key Usage extension should be omitted entirely during encoding. Can I suggest to make the operation more explicitly named (eg `verify_cert_constraints_valid_for_key_type`). Both `find_constraints` + `match_constraints` are not really clear about what they do.

cordney: Change applied in 6cbff45. I took the time to extract some typical key usage scenarios from RFC 5280 and add unit tests for `verify_cert_constraints_valid_for_key_type()`. During which I corrected that ENCIPHER_ONLY and DECIPHER_ONLY were allowed for RSA and ElGamal, where RFC 5280 says that both are undefined if not used together with KEY_AGREEMENT. This change was only introduced in this PR though, so no worries.

`$ ./botan gen_self_signed rsa.priv localhost` now creates a certificate with no key usage extension, which is what we wanted.

randombit:
The CircleCI test failure was unrelated to this change:

mceliece ran 1 tests 1 FAILED
Failure 1: mce_decrypt failed: False assertion 'Valid degree' (expression v_deg > 0) in goppa_decode @botan_all.cpp:26417

DRBG seed was 146C350A378F4A04. I cannot replicate, maybe some difference in the build config. In any case that test is randomized so more runs should find it again.

It occurs to me that using the fuzzers to test McEliece would be a really good idea given how complex that code is.

src/lib/cert/x509/x509self.cpp
```diff
@@ -99,16 +107,27 @@ PKCS10_Request create_cert_req(const X509_Cert_Options& opts,
...
    extensions.add(
       new Cert_Extension::Basic_Constraints(opts.is_CA, opts.path_limit));
-   extensions.add(
-      new Cert_Extension::Key_Usage(
-         opts.is_CA ? Key_Constraints(KEY_CERT_SIGN | CRL_SIGN) :
-                      find_constraints(key, opts.constraints)
-         )
-      );
+
+   if(constraints != NO_CONSTRAINTS)
```
randombit: This check isn't required since Key_Usage::should_encode returns false if its value is set to NO_CONSTRAINTS. But checking here is harmless and saves a few mallocs for what is otherwise a no-op.

cordney: I see now. But `Key_Usage::should_encode()` is not called from `Extensions::add()`, instead `encode_inner()` is called directly to place the extension value in `m_extensions_raw`.

PR #597, opened 2016-08-19
Fix minimized build when pkcs11 module is enabled

neusdan:
Failures can be seen here:

https://travis-ci.org/randombit/botan/jobs/153653338

### DIFF
--- a/src/tests/test_pkcs11_high_level.cpp
+++ b/src/tests/test_pkcs11_high_level.cpp
@@ -415,6 +415,7 @@ Test::Result test_attribute_container()
    return result;
    }
 
+#if defined(BOTAN_HAS_ASN1)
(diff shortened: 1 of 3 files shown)


randombit:
One build still failing https://travis-ci.org/randombit/botan/jobs/153853993

randombit:
I just looked at this build error a bit and don't understand what is going on. I can reproduce the same error using the configure flags `--with-pkcs11 --minimized --via-amalgamation`. The build errors out because PKCS11::Session is not defined at a certain point, but Session does seem to appear before that line in `botan_all.h`! I don't see any problems with the PKCS11 headers (eg duplicated header guard names, implicit header depenedencies, ...) that would bug the amalgamation but clearly something is off.

PR #598, opened 2016-08-21
Add more HMAC_DRBG unit tests

cordney:
Adds some more unit tests for HMAC_DRBG, namely:

* Test that RNG fails to output if entropy source(s) fail
* Test that the size limit on nonces is enforced
* Test that setting `max_output_before_reseed = 1` makes the RNG reseed on every request
* Test that RNG reseeds after a fork

The already previously present generic reseed tests currently fail, but I changed nothing on these. Also the nonce length tests fail, I think the current implementation is either broken or does not what I expect (test is wrong). The fork safety test failed and I could quickly spot the error. `get_process_id()` was using the wrong define for OS type Unix (BOTAN_TARGET_OS_IS_UNIX vs. BOTAN_TARGET_OS_TYPE_IS_UNIX), so it would always return a zero pid on Unixes and the RNG would never reseed after fork. I fixed it and made `get_process_id()` throw an exception on unsupported platforms. IMHO it's a too important helper function to just let it fail silently.

### DIFF
--- a/src/lib/utils/os_utils.h
+++ b/src/lib/utils/os_utils.h
@@ -15,7 +15,7 @@ namespace Botan {
 namespace OS {
 
 /**
-* Returns the OS assigned process ID, if available. Otherwise returns 0.
+* Returns the OS assigned process ID, if available. Otherwise throws.
 */
 uint32_t get_process_id();
 


src/lib/utils/os_utils.cpp
```diff
@@ -29,12 +29,12 @@ namespace OS {
...
 uint32_t get_process_id()
    {
-#if defined(BOTAN_TARGET_OS_IS_UNIX)
+#if defined(BOTAN_TARGET_OS_TYPE_IS_UNIX)
    return ::getpid();
 #elif defined(BOTAN_TARGET_OS_IS_WINDOWS)
    return ::GetCurrentProcessId();
 #else
-   return 0;
+   throw Exception("get_process_id not supported");
```
randombit: OK, but update the function comment in os_utils.h

cordney:
Oh and one of the entropy input broken test fails, but I think it's only because the HMAC_DRBG constructor that takes both an RNG and entropy sources does not pass the entropy sources to the Stateful_RNG base class constructor.

src/lib/utils/os_utils.cpp
cordney: Fixed in 8a92285.

randombit:
Ha, oops. Fixed in fe83c73 thanks

randombit:
I think for the invalid nonce size tests, the problem is `initialize_with` is basically silently failing; it accepts the input and incorporates it but does not consider the DRBG seeded. But then, because the HMAC_DRBG references the Test::rng, it just automatically reseeds itself anyway. Take a look at e95938a

PR #600, opened 2016-08-22
Add HMAC_DRBG reseed logic tests

cordney:
I was asking myself what happens when one requests more than `max_output_before_reseed bytes` and wrote a test, which fails for me currently. NIST SP800-90A says if too many bytes requested, return ERROR. I see that this is inconvenient for the caller, so currently a reseed occurs once and then the caller gets his requested number of bytes. But if this number exceeds `max_output_before_reseed`, then he is effectively circumventing the `max_output_before_reseed` limit. @neusdan and me think that in this case a reseed has to occur bytes_requested/max_output_before_reseed times, e.g., if `max_output_before_reseed` is 16 bytes and the caller requests 32 bytes, then reseed a first time, get `max_output_before_reseed bytes`, reseed a second time, get `max_output_before_reseed` bytes, concatenate and return 32 bytes.

I also wrote tests to check if HMAC_DRBG really reseeds after `max_output_before_reseed` bytes requested and there seems to be a bug in the counting of `m_bytes_since_reseed`. The reason seems to be that the reseed_from_rng()/reseed() call sets `m_bytes_since_reseed = 0`, not counting the `bytes_requested` that will be taken from the entropy pool and given to the caller just after that the reseed.

The tests are annotated, let me know if you have questions. I didn't fix the RNG right away, wanted to hear your opinion first.

### DIFF
--- a/src/tests/test_rng.cpp
+++ b/src/tests/test_rng.cpp
@@ -161,9 +161,9 @@ class HMAC_DRBG_Unit_Tests : public Test
          };
 
    public:
-      Test::Result test_reseed()
+      Test::Result test_reseed_kat()
          {
-         Test::Result result("HMAC_DRBG Reseed");
+         Test::Result result("HMAC_DRBG Reseed KAT");
 
          auto mac = Botan::MessageAuthenticationCode::create("HMAC(SHA-256)");
          if(!mac)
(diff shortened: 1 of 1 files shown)


randombit:
I did it that way because reseeding multiple times in the middle of a long request seemed kind of pointless, and additionally expensive. Some earlier versions had logic like that and it caused key generation in particular to be very slow. Though moving the RNG just reading the OS RNG during reseeds should at least make this much less costly.

Looking at SP800-90A to see what the options are, the only output limit in terms of bytes I can see is in section 8 which uses `max_number_of_bits_per_request`, which I think represents a  different concept than botan's `max_output_before_reseed`. SP variable doesn't affect reseeding, it is an implementation defined limit on the maximum size of any request. For HMAC_DRBG for all hash functions this value is required to be <= 2**19 bits, or 64 KiB. One way to handle long requests would be to treat them as several (contiguous) maximum length requests.

Instead HMAC_DRBG seems to use a counter of the number of calls to generate (`Number of requests between reseeds (reseed_interval)`), ignoring their size (aside from the maximum size limit). Section 10.1.2.5

I could change the reseed logic to a request counter instead of a byte counter, which seems to be NISTs intent. In that case, there is no need to reseed in the middle of a single request, except in the case of requests which were split up due to size limits in which case a reseed check would be performed for each sub-request.

cordney:
>
I could change the reseed logic to a request counter instead of a byte counter, which seems to be NISTs intent. In that case, there is no need to reseed in the middle of a single request, except in the case of requests which were split up due to size limits in which case a reseed check would be performed for each sub-request.
>

Sounds good for me.

PR #605, opened 2016-08-25
Add unit test for HMAC_DRBG::randomize_with_ts_input()

cordney:
This is the last function in HMAC_DRBG and Stateful_RNG that was not already covered with tests.


randombit:
I cancelled the Travis build which seems to be stalled due to an outage Travis is having currently with OS X builds.

randombit:
Merged to master, thanks!

PR #608, opened 2016-08-26
Travis CI cleanup and additions

randombit:
Use a single variable to specify the build target instead of multiplexing over 4. Only test configurations removed are Boost-free OS X builds, as we were already testing same config but with Boost. Travis OS X is slow so I'd prefer to minimize usage.

For comparison, current master: https://travis-ci.org/randombit/botan/builds/155134214
this PR: https://travis-ci.org/randombit/botan/builds/155385851

Adds Linux cross compile targets for ARM32, ARM64, PPC64, and MinGW x86. Currently all of these build but the tests are failing to run for misc reasons, so the Linux cross compile targets are marked as allowed to fail (for now). For ARM32/ARM64, problem seems to be related to starting threads (possibly due to the use of static binaries?). On PPC64, I simply cannot get qemu to execute the binary. And on Wine the entropy poll fails which subsequently causes a crash due to a bug in the FFI tests.

Adds sanitizer targets for both LLVM Clang and Apple Clang.

Adds a valgrind target (in addition to a general memcheck run, this uses the memcheck API to test const time annotations)

Adds several small hacks in order to allow MinGW GCC cross compile. I am not sure if any of these are 'correct', but they got things to build.

I'd prefer to get the cross tests all running before merging, but I think as it stands it's already a nice improvement.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,113 +1,86 @@
-dist: trusty
-
-sudo: required
-
 language: cpp
 
-# BEGIN BUILD MATRIX
-#
-# Build matrix must be smaller on branch coverity_scan because quota allows
-# only 5 Coverity Scan jobs at once. Thus we have different versions of this
-# file on branch master and coverity_scan.
-#
-# Please keep this file as much in sync as possible to allow easy merging
-# from master into coverity_scan and only have differences in the build
-# matrix block.
-#
-# Check the difference as follows:
-#   $ git checkout master
-#   $ git checkout coverity_scan
-#   $ git difftool master coverity_scan .travis.yml
 os:
   - linux
   - osx
 
+dist: trusty
+sudo: required
 osx_image: xcode7
 
 compiler:
   - clang
   - gcc
 
-git:
-  depth: 5
-
 env:
   global:
+    - CCACHE_SIZE=50M
+    - CCACHE_COMPRESS=1
+    - BUILD_JOBS=2
     # SONAR_TOKEN for accessing the SonarQube server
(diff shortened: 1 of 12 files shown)


randombit:
Many commits + several rebases later ...

Tests for ARM32 and ARM64 are running (and passing!) under qemu. They've been moved to must-pass. I added a PPC32 build because Trusty has the cross compiler and libs. PPC32 tests start under qemu, but crashes. And I still cannot get PPC64 qemu to run the generated botan-test. I think for more progress on both of these I need an Ubuntu VM set up.

I removed the Apple Clang sanitizer build. It didn't seem to recognize UbSan flags, so I'm assuming there is no special Apple Clang sanitizer magic we are missing here. LLVM Clang still has a sanitizer build. When running the tests is very slow (valgrind/qemu) I limited the build to a subset (policy 'modern').

Tests basically pass under MinGW/Wine, except that there is no filesystem access enabled so the block, hash, kdf etc tests do no run and are marked as failures. So it is still in the allow-fail group for now, but it should be an easy upgrade. Oddly, we may be in the position of having to write Win32 code to support GCC when Microsoft lets us use the TR2 filesystem interface. (Well, GCC 6 has filesystem, but Trusty's 4.8 does not...)

I'm out of steam on this and assuming CI is green think it is good to go. I will revisit the PPC issue later, for now just having PPC cross compile configure/build is a great start.

mouse07410:
@randombit to use sanitizers with Clang on Apple you need a Macports-installed Clang (and correspondingly, Macports-installed LLVM). As you noticed, the native (Xcode) Clang unfortunately does not support sanitizers (I wonder why, but that's besides the point).

PR #610, opened 2016-08-28
Fix coverage

neusdan:


cordney:
Seems this change didn't help, though for others it worked. What else could be the reason?

neusdan:
should be fixed now :-)

cordney:
Nice, thanks! So this was because the new lcov version required an absolute path instead of a relative one?

randombit:
Thanks for digging into this

Issue #412, opened 2016-01-21
Keccak_1600 output does not match NIST SHA3 test vectors.


BouncyCCPP:
Hi,
Is Keccak_1600 implementation in Botan v 1.11.26 based on the final approved SHA-3 algorithm?  I wrote a little test program to get  a 256 bit digest.  The program is run on a Window 7.  My code is as   following:
secure_vector<byte> sha3v2(vector<byte> input, size_t output_bits) {
  Keccak_1600 sha3(output_bits);
  sha3.update(input);
  return (sha3.final());
}
...
cout << "SHA3v2-256: " + Botan::hex_encode(sha3v2(data, 256)) << "\n";

Where the data is "abc".

Botan returns: 
SHA3v2-256: 4E03657AEA45A94FC7D47BA826C8D667C0D1E6E33A64A036EC44F58FA12D6C45
while NIST vector (http://www.di-mgt.com.au/sha_testvectors.html) for the same input is:
SHA-3-256    3a985da74fe225b2045c172d6bd390bd855f086e3e9d525b46bfe24511431532

Is it my test routine or is it that Keccak_16000 is not based on the approved SHA-3 algorithm?  If it is the second case, then will Botan be updated to support the final version?  If yes, is there some time frame?

Thanks

randombit:
The Keccak version implemented is the round 3 version from the NIST hash competition. I believe NIST made some changes to the algorithm for the final SHA-3, as well as defining various alternative modes like SHAKE.

The final SHA-3 will doubtless be supported eventually (likely with a rename of Keccak, since it will be an incompatible change of the algorithm), but the timeline would be either when someone does the research and submits the appropriate patch or whenever I decide SHA-3 support is relevant to some other goal and do it. Since SHA-3 is slower than SHA-512 and not widely deployed yet, it has not been a priority for me personally.

mouse07410:
My preference would be to keep both the original Keccak, *and* the modified algorithm renamed as SHA-3.

randombit:
That's true it should be easy to continue supporting the current version of Keccak which has been available in the library now for some time (since 2010 or so) in addition to adding SHA-3 support. No idea if anyone is using Keccak but there is no point in breaking their usage, as the currently implemented version remains secure.

mouse07410:
> No idea if anyone is using Keccak...

[...]

PR #613, opened 2016-08-30
[For Review] NewHope Ring-LWE scheme

randombit:
Based on the public domain ref implementation. An unauthenticated KEM scheme, the initiator sends 1824 bytes, the recipient sends 2048 bytes back, after which both parties can compute a shared 256 bit key. Provides conjectured ~200 bits security against a quantum attacker. Also pretty fast.

This is not compatible with the version in BoringSSL which IIRC uses different hashes/XOFs, and may have other differences - I'm not sure about this. I would eventually like to implement both versions (ref + boringssl) since that would allow interop with Google's public NEWHOPE experiment https://security.googleblog.com/2016/07/experimenting-with-post-quantum.html

Test vector file is the first 1000 outputs generated by ref's testvectors.c. I've tested the first 100,000 outputs locally.

Currently has some of its own logic for SHA-3 and SHAKE-128 baked in, these can be pulled out whenever SHAKE is properly added.

The API is raw, basically just ref's newhope.h. The currently is no formalism in pubkey.h for a 2-phase KEM scheme of this kind, I am not sure how to approach that currently.


randombit:
As best I can tell the only difference between ref and BoringSSL is the hashes - AES128-CTR instead of SHAKE-128 for the expansion of `a` and SHA-256 instead of SHA3-256 for the final key derivation.

randombit:
I think stage 1 is just this bare-bones API with the standard (SHA-3 based) parameters, leave it at that for 1.11.32 and expand with BoringSSL variant, TLS ciphersuites, and pubkey.h wrapper later on. Since there is no long term key (my understanding is that you can reuse (just) the `a` param, but doing even that is somewhat unwise) there may not ever be `NewHope_{Public,Private}Key` types. I will probably merge this at 6b2e3aa in a day or two unless there are objections. The release notes for 1.11.32 will emphasize that NewHope is basically a technology preview.

Merging this will add some umph to finally supporting proper SHA-3 (#412)

PR #616, opened 2016-09-01
[For review] ChaCha SSE2 optimizations

randombit:
Keeps the same 16-word state but each time ChaCha runs to expand an output it generates 4 blocks at once instead of just 1. With unrolling/interleaving allows a lot of ILP in the SSE2 version which seems important. My original single block SSE2 version in 858e3be was only ~5-10% faster than the scalar version, I suspect because ILP there was very low (almost every single result depended on the result of some earlier instruction). This same principle can be expanded to an 8-way AVX2 ChaCha later on.

This version is ~60% faster than scalar on Skylake, seeing about 830 MB/s here. Probably performance is not great on x86-32 due to only 8 architectural SSE registers.

Main driver of this work is having a very fast RNG for use by NewHope (#613) but faster ChaCha is useful in lots of places.

Also worth noting this takes a different tact from the traditional providers model used elsewhere - if SSE2 is available we jump to it or otherwise use the scalar version, but it's all with just `ChaCha`, no `ChaCha_SSE2` and company. I expect adopting this model across the library will be the final resolution to #477


randombit:
Checking on an older processor, 2.4 GHz Westmere, best of 5 runs for 1 second, scalar was 240.5 MB/s SSE2 402.4 MB/s so again roughly 60% speedup from the current work.

It would be easy to translate this method to AVX2, but I will keep that for a future PR.

randombit:
I don't plan further commits on this branch

Issue #562, opened 2016-07-29
Problem with amalgamation


MaxChu719:
I tried to built the amalgamation in Mac (Version 10.11.6) using the command: 
`./configure.py --gen-amalgamation --single-amalgamation-file --via-amalgamation`

Then I type `make` to compile the code and i get the following error:

```
clang++  -m64 -pthread -fPIC -fvisibility=hidden -std=c++11 -D_REENTRANT -fstack-protector -O3 -Wall -Wextra -Wpedantic -Wshadow -Wstrict-aliasing -Wstrict-overflow=5 -Wcast-align -Wmissing-declarations -Wpointer-arith -Wcast-qual -Wunreachable-code -Ibuild/include -c botan_all.cpp -o build/obj/lib/botan_all.o
botan_all.cpp:778:7: error: always_inline function '_mm_aesenc_si128' requires target feature 'aes', but would be inlined into function 'encrypt_n' that is compiled without
      support for 'aes'
      AES_ENC_4_ROUNDS(K1);
      ^
botan_all.cpp:711:12: note: expanded from macro 'AES_ENC_4_ROUNDS'
      B0 = _mm_aesenc_si128(B0, K);             \
           ^
botan_all.cpp:778:7: error: always_inline function '_mm_aesenc_si128' requires target feature 'aes', but would be inlined into function 'encrypt_n' that is compiled without
      support for 'aes'
botan_all.cpp:712:12: note: expanded from macro 'AES_ENC_4_ROUNDS'
      B1 = _mm_aesenc_si128(B1, K);             \
           ^
botan_all.cpp:778:7: error: always_inline function '_mm_aesenc_si128' requires target feature 'aes', but would be inlined into function 'encrypt_n' that is compiled without
      support for 'aes'
botan_all.cpp:713:12: note: expanded from macro 'AES_ENC_4_ROUNDS'
      B2 = _mm_aesenc_si128(B2, K);             \
           ^
botan_all.cpp:778:7: error: always_inline function '_mm_aesenc_si128' requires target feature 'aes', but would be inlined into function 'encrypt_n' that is compiled without
      support for 'aes'
botan_all.cpp:714:12: note: expanded from macro 'AES_ENC_4_ROUNDS'
      B3 = _mm_aesenc_si128(B3, K);             \
           ^
botan_all.cpp:779:7: error: always_inline function '_mm_aesenc_si128' requires target feature 'aes', but would be inlined into function 'encrypt_n' that is compiled without
      support for 'aes'
[...]

PR #621, opened 2016-09-03
Simplify amalgamation generation

webmaster128:
As discussed somewhere months ago, this is an attempt to simplify the amalgamation generation.

Previously, there were two moves:

* `--gen-amalgamation`: configure the library as usual and by the way, drop some amalgamations into the folder
* `--via-amalgamation`: build amalgamation files and use them to build the library

I guess the first one is there for historic reasons and has no real advantage. If a user really wants to generate amalgamation files and build the library without amalgamation, he can easily and explicitly configure twice.

This PR does two things:

1. Remove `--gen-amalgamation`
2. Rename `--via-amalgamation` to `--amalgamation`

This should avoid confusion when people start using the library (e.g. #562).

I would keep the old options for some releases to help people switching their build system and then drop them entirely.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -289,11 +289,15 @@ def process_command_line(args):
 
     build_group.add_option('--gen-amalgamation', dest='gen_amalgamation',
                            default=False, action='store_true',
-                           help='generate amalgamation files')
+                           help='generate amalgamation files and build without amalgamation (removed)')
 
     build_group.add_option('--via-amalgamation', dest='via_amalgamation',
                            default=False, action='store_true',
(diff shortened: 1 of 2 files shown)


randombit:
You are correct, generate amalgamation vs build-with were distinct projects thus the strange split in the option. This seems fine to me.

webmaster128:
Okay, thanks. I'll add an changelog entry for this.

webmaster128:
Added a changelog entry for this. This is ready from my point of view.

randombit:
Thanks for adding the relnote

PR #623, opened 2016-09-03
Address issue #477

randombit:
See #477 for rationale.

This turned out to be easier than I expected. I do not think this change breaks many applications, though it is possible someone is using `AES_128_NI` or the like directly (rather than via `BlockCipher::create`).


webmaster128:
Is there still a way to debug the implementations? I.e.

1. Get a list of all compiled implementations of a given algorithm
2. Get runtime information about which implementation is used

Both work fine in the current library.

randombit:
If an application was especially curious it could still detect (say AES-SSSE3) by checking for the BOTAN_HAS_AES_SSSE3 macro plus the runtime CPUID bit, but the actual entry point is not exposed and outside of CPUID bit hackery there is no way to force a particular implementation to be used.

I'm not happy about losing the visibility on testing/perf that results from this change, and truly this PR is only safe when assisted by a very large CI matrix (but we have that now). I do not make tradeoffs of such qualities lightly, but in this case it allows a safer to use API that works more in line with what a developer will expect.

For an application developer or end user, I think they will always want to use the fastest/safest version. So this change seems like a strict improvement for them, AFAICT.

webmaster128:
Couldn't aes.h provide those information?

```
static std::vector<std::string> AES_128::implementations()
static std::string AES_128::default_implementation()
```

From there it should not be too hard to explicitly choose an implementation in the AES_128 constructor.

The implementation type might be string as right now, or change to an enum.

randombit:
It could, but afaict such functionality is literally only useful for testing. In which case, it seems better to instead be able to annotate a set of tests to indicate that they should be run multiple times. For example, putting something like this in aes.vec

```
#cpuid aesni ssse3
```

would indicate to the tests that these cpuid flags are special to this implementation, and run each test 3 times, once with AES-NI CPUID bit set, once with SSSE3 CPUID bit set, and once with neither bit set.

This would also allow testing code like GCM which already does not provide any way to access both versions (if CLMUL bit is set, that implementation is always used).

Is there some reason I am not thinking of (other than tests) that any code would want to pick a particular implementation, versus always using whatever is fastest on that platform? It seems like this otherwise just complicates the API; we are asking the developer a question, but there is only one right answer, which we already know... so why ask?

webmaster128:
> Is there some reason I am not thinking of (other than tests) that any code would want to pick a particular implementation, versus always using whatever is fastest on that platform? It seems like this otherwise just complicates the API; we are asking the developer a question, but there is only one right answer, which we already know... so why ask?

I am only interested in asking, which implementation is used. I don't want to implement it via cpuid because I don't want to duplicate the decision logic (which might change). We recently added AES-NI to the build and it was just good to be able to double check, which machines did pick up which implementation.

randombit:
> I am only interested in asking, which implementation is used. I don't want to implement it via cpuid because I don't want to duplicate the decision logic (which might change).

That makes sense. That is easier to solve IMO, add `std::string T::provider() const` which returns "base" most of the time but for implementations with special handling, returns an appropriate string "sse2", "avx2" or the like. This would effectively mirror the `provider` argument to `T::create` as it is used now (in master).

randombit:
Rebased to pick up ChaCha SSE2 from master, added T::provider in 04bf8dc

webmaster128:
:+1: Could we make the return type std::string?

randombit:
I think this branch is ready to merge. Already I am happy with adding the CPUID overloading to tests, which allows testing both paths of the GCM multiply (clmul vs normal) which was not possible. It will also be useful when testing future AVX2 or ADX based BigInt optimizations.

randombit:
This is ready to go except for a persistent test failure only on the ARM32 cross compile/qemu build:

```
...
Curve25519 scalarmult ran 18 tests in 73.07 msec all ok
terminate called after throwing an instance of 'Botan::Provider_Not_Found'
  what():  Could not find provider 'openssl' for DH
qemu: uncaught target signal 6 (Aborted) - core dumped
```

I have been unable to replicate the problem so far with a native build. Added 4266b51 to help debug, that try/catch should have been there anyway.

PR #624, opened 2016-09-04
Call base class assignment operator in X509_Certificate

cordney:
`X509_Certificate::operator=()` did only copy private members, but not that of the `X509_Object` base class. That resulted in zombie certificates in the following use case:

```cpp
X509_Certificate cert1( "cert1.crt" );
X509_Certificate cert2( "cert2.crt" );

cert1 = cert2;
// congrats, cert1 now has subject, issuer and extensions from cert2, but signature still from cert1!
```

Fixed by calling the base class, that is `X509_Object` assignment operator from within `X509_Certificate`'s assignment operator. Turns out we recently added unit tests for this assignment operator, but the test was wrong. Fixed it too and got a test failure, with the production code fix it's green now.

### DIFF
--- a/src/lib/cert/x509/x509cert.cpp
+++ b/src/lib/cert/x509/x509cert.cpp
@@ -89,6 +89,7 @@ X509_Certificate& X509_Certificate::operator=(const X509_Certificate& other)
       }
    else
       {
+      X509_Object::operator=(other);
       m_subject = other.m_subject;
       m_issuer = other.m_issuer;
       m_self_signed = other.m_self_signed;
--- a/src/lib/cert/x509/x509cert.cpp
+++ b/src/lib/cert/x509/x509cert.cpp
@@ -72,33 +72,6 @@ X509_Certificate::X509_Certificate(const std::vector<byte>& in) :
    do_decode();
    }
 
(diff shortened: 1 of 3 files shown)


randombit:
Nice find, thank you.

`X509_Certificate::operator=` seems to just member-wise copy the elements - could it instead a `= default` compiled generated version?

cordney:
Good idea, I tested successfully with both a `= default` assignment op and copy constructor and with not declaring them at all. Which solution do you prefer?

neusdan:
AFAIK best praticse is to declare all or none of the special member functions

cordney:
@neusdan That's what I had in mind for the second solution.

cordney:
We could also remove the `private` declaration and definition of the default constructor, which is not auto-generated by the compiler anyway.

randombit:
@cordney I prefer explicit  `default` declarations because it is then clear to the reviewer that the operator= (or whatever) is intentionally defined, rather than leaving them to wonder if perhaps the operator simply happens to exist by accident due to the original programmer forgetting to make the type explicitly non-copyable.

I think the private default constructor might be used by friends like `X509_CA`, not sure.

cordney:
@randombit Updated the merge request accordingly.

PR #625, opened 2016-09-05
Remove Algo_Registry/static initialization from pubkey code

randombit:
WIP but posting for comments. Instead of going through the algo_registry, you get an object suitable for a particular operation (encryption, signing, ...) by asking the key directly. There isn't any user visible change as a result of this.

A major change here is it is no longer possible to eg register a new RSA operation at runtime. Instead the base type will have to know about it; there is a basic sketch in the RSA code for openssl. I don't view this as a huge problem in practice because the only thing it really effects is someone writing an out of tree custom operation for an existing key type.

This code also adds a rng to each function creating a PK operation. In the current iteration this parameter is universally ignored but it will be used for setting up blinding or other per-operation randomizations, and to avoid having to break API right now pubkey.h just passes `system_rng` all the time. This will change before merge, and instead the caller will be forced to provide an RNG here (+ adding deprecated compatibility constructors which pass `system_rng`, for systems with `system_rng`)

This is step 2/N towards removing use of static initialization in the library.


randombit:
TODO: ECIES is disabled in the build. It does funny things with the key type.
TODO: PKCS 11 ECDH cofactor mode is missing, needs to take the mode from `params`
TODO: fix the RNG hack in pubkey.cpp, update pubkey.h API, update tests
TODO: use app-provided RNG in `Blinder`

randombit:
With a40ea20 this should be ready to go (or at least for review). There is more work to do along these lines, notice for instance when encrypting we require an RNG now both during setup of the PK_Encryptor and for each encryption operation. I will approach that issue in a later PR, since this one is already large and that change would have a lot of implications for the tests.

I will probably rebase before merging, otherwise not expecting any more changes.

randombit:
Despite my previous comment, some further changes in 31506d8:

- In most cases `provider` was being ignored, now instead check for explicit values we know about (using empty to indicate 'choose a default') or else throw an exception. This better matches the current behavior.

- Changed (back) PK_Verifier to not consume an RNG. Since the operation is public, it does not require blinding or the like, and at least all existing algos are deterministic (AFAICT no signature scheme with a probabilistic verification algorithm has been defined, or at least is not in common use). In all cases the existing code ignored the value provided.

randombit:
This is ready to go except for a persistent test failure only on the ARM32 cross compile/qemu build:

```
...
Curve25519 scalarmult ran 18 tests in 73.07 msec all ok
terminate called after throwing an instance of 'Botan::Provider_Not_Found'
  what():  Could not find provider 'openssl' for DH
qemu: uncaught target signal 6 (Aborted) - core dumped
```

I have been unable to replicate the problem so far with a native build. Added 4266b51 to help debug, that try/catch should have been there anyway.

randombit:
Modulo the strange (but consistent) test failure on the arm32 Linux cross build, this is ready to go IMO. I'm hoping 4266b51c helps debug the test failure.

@neusdan @cordney any comments on this branch? Does my change to PKCS11 ECDH cofactor handling seem OK to you?

cordney:
+1, especially for forwarding the user-defined RNG& to the blinder. Apart from that, do the new `create_xxx_op()` functions really need to be `public` on `PrivateKey` and `PublicKey`?  Is there any use case for a typical Botan user?

cordney:
On PKCS11 I cannot really judge, I'd need to hear @neusdan's opinion.

randombit:
I don't forsee any typical user needing these functions (continuing to use pubkey.h as the documented pubkey interface), but I didn't see any way for them to be `private` without forcing *each* public/private key type to `friend` all appropriate PK_xxx classes which IMO is a lot of work/maint just to prevent someone from doing something mostly harmless.

It was not obvious from the header that these functions were not for typical app use, added a warning in 7830a77

Another option which would completely prevent use of these functions is to make pk_ops.h an internal only header - then applications could call `create_encryption_op` but could not do anything with the resulting pointer because the declaration of the type would not be available.

Thinking about our public API as part of our technical debt (by making an API public, we promise not to break it, or at least do our best - and each promise like that we make incrementally adds more work for maintenance over time), minimizing the size of the public API where possible is always sound and so internaling pk_ops.h now seems like a good idea anyway - it is new in 1.11, not documented, not used directly by either cli or tests, and doesn't expose any new functionality not available through pubkey.h. Why promise to maintain that interface, if we don't have to?

So, I'll update this PR to also hide pk_ops.h from the public API

Doing a quick scan of all the exported headers only `gf2m_small_m.h` and `polyn_gf2m.h` from McEliece, and maybe `pow_mod.h` seemed likely candidates for being made internal only in the same way. I will deal with those in a later PR.

randombit:
Also I have no idea what's up with the persistent failure on ARM32 - std::terminate is being called due to an uncaught exception, despite multiple try/catch blocks that should match.

As best I can tell the throw is originating from the DH KAT tests. test_pubkey.cpp attempts to explicitly create an "openssl" instance (along with "base") so all versions can be tested, but in these builds OpenSSL was not built into the library. So the library throws an exception, which the test runner expects since openssl is not always available and it creates the PK_xxx object in a try/catch block. But on ARM32 the exception is not caught and the process terminates completely. I tried changing the exception to be exactly of type `Lookup_Error` instead of a subclass, as I have in the past seen issues like this regarding catching exceptions thrown across a shared library boundary. But that was a long time ago and I thought it was a resolved issue, in any case that did not fix it and not even the last ditch `catch(...)` in main seems to trigger.

The crash is entirely consistent and I'm out of ideas for debugging it via CI, which means I'll need to get an Ubuntu VM setup up to figure out what the ### is going on and merging this PR is blocked until then.

randombit:
After upgrading qemu in CI, the inexplicable crash seems to have gone away. I am going to chalk it up to a qemu bug. I still need to internal pk_ops.h

randombit:
Good to merge here I think, both CI failures for 97339a6 were the still undiagnosed occasional FFI test failure

Issue #619, opened 2016-09-01
TLS server does not honor TLS signature_algorithms extension


jurajsomorovsky:
According to the TLS 1.2 spec, the server should reject a connection if the SignatureAndHashAlgorithm extension does not contain appropriate algorithms (https://tools.ietf.org/html/rfc5246#section-7.4.3):

```
   If the client has offered the "signature_algorithms" extension, the
   signature algorithm and hash algorithm MUST be a pair listed in that
   extension.  Note that there is a possibility for inconsistencies
   here.  For instance, the client might offer DHE_DSS key exchange but
   omit any DSA pairs from its "signature_algorithms" extension.  In
   order to negotiate correctly, the server MUST check any candidate
   cipher suites against the "signature_algorithms" extension before
   selecting them.  This is somewhat inelegant but is a compromise
   designed to minimize changes to the original cipher suite design. 
```

Botan TLS server does not honor the list of algorithms offered in this extension. If the extension contains, for example, SignatureAndHashAlgorithm `ANONYMOUS-SHA512`, it is still possible to execute a correct TLS handshake with the `TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA` cipher suite.

To reproduce this issue, generate an RSA key pair and start the Botan server with:

```
openssl` genpkey -algorithm RSA -out rsa2048.pem -pkeyopt rsa_keygen_bits:2048
openssl req -key rsa2048.pem -new -x509 -days 365 -out rsa2048.crt
./botan tls_server rsa2048.crt rsa2048.pem --port=4433
```

Download TLS-Attacker 1.1 (https://github.com/RUB-NDS/TLS-Attacker/releases/tag/v1.1) and start it with the following command:

```
java -jar TLS-Attacker-1.1.jar client -workflow_input invalid-ssh-algorithm.txt -connect localhost:4433
```

with the following handshake configuration file:
[invalid-ssh-algorithm.txt](https://github.com/randombit/botan/files/450870/invalid-ssh-algorithm.txt)

As you will see, the Botan server ignores the SignatureAndHashAlgorithm extension ( with `ANONYMOUS-SHA512`) and uses `RSA-SHA1` for ServerKeyExchange signing.

According to the specification the server should reject the connection.
Please note that this is rather a compatibility issue. Note also that with TLS1.1 and 1.0, the server should behave differently.

See also the wireshark trace, where the ClientHello and ServerKeyExchange message contents are displayed:
[...]

PR #630, opened 2016-09-21
TLS Server should respect client signature_algorithms

randombit:
Should address GH #619 but needs more testing.

### DIFF
--- a/src/lib/tls/msg_client_hello.cpp
+++ b/src/lib/tls/msg_client_hello.cpp
@@ -80,24 +80,26 @@ Client_Hello::Client_Hello(Handshake_IO& io,
                                     client_settings.srp_identifier() != "")),
    m_comp_methods(policy.compression())
    {
+   BOTAN_ASSERT(policy.acceptable_protocol_version(client_settings.protocol_version()),
+                "Our policy accepts the version we are offering");
+
+   /*
+   * Place all empty extensions in front to avoid a bug in some sytems
+   * which reject hellos when the last extension in the list is empty.
+   */
    m_extensions.add(new Extended_Master_Secret);
+   m_extensions.add(new Session_Ticket());
+   if(policy.negotiate_encrypt_then_mac())
+      m_extensions.add(new Encrypt_then_MAC);
+
    m_extensions.add(new Renegotiation_Extension(reneg_info));
    m_extensions.add(new Server_Name_Indicator(client_settings.hostname()));
-   m_extensions.add(new Session_Ticket());
-   m_extensions.add(new Supported_Elliptic_Curves(policy.allowed_ecc_curves()));
-
-   if(m_version.supports_negotiable_signature_algorithms())
-      m_extensions.add(new Signature_Algorithms(policy.allowed_signature_hashes(),
-                                                policy.allowed_signature_methods()));
-
-   if(m_version.is_datagram_protocol())
-     m_extensions.add(new SRTP_Protection_Profiles(policy.srtp_profiles()));
 
    if(reneg_info.empty() && !next_protocols.empty())
       m_extensions.add(new Application_Layer_Protocol_Notification(next_protocols));
-   
-   if(policy.negotiate_encrypt_then_mac())
-      m_extensions.add(new Encrypt_then_MAC);
+
+   if(m_version.is_datagram_protocol())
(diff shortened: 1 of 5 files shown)


cordney:
Tested again with TLS Attacker and two things are definitely fixed: ANON is rejected in SignatureAndHashAlgorithms and it respects the signature algorithm with regard to both ciphersuite AND signature algorithm. Two problems remain though:

1) the requested hash algorithm is still ignored (only signature algos from the extension are honored)
2) server chains with mixed signature algos are not handled

For 1), I could test this with @jurajsomorovsky's TLS Attacker again as he mentioned in #619. Regardless of which hash algorithm TLS Attacker sends, let it SHA-1 or SHA-512, the tls_server always sends the rsaWithSha256 certificate.

For 2), TLS 1.2 (in contrast to TLS 1.1) explicitly allows mixed signature+hash algorithm chains:

```
RFC 5246, 7.4.2
   If the client provided a "signature_algorithms" extension, then all
   certificates provided by the server MUST be signed by a
   hash/signature algorithm pair that appears in that extension.  Note
   that this implies that a certificate containing a key for one
   signature algorithm MAY be signed using a different signature
   algorithm (for instance, an RSA key signed with a DSA key).  This is
   a departure from TLS 1.1, which required that the algorithms be the
   same.  Note that this also implies that the DH_DSS, DH_RSA,
   ECDH_ECDSA, and ECDH_RSA key exchange algorithms do not restrict the
   algorithm used to sign the certificate.  Fixed DH certificates MAY be
   signed with any hash/signature algorithm pair appearing in the
   extension.  The names DH_DSS, DH_RSA, ECDH_ECDSA, and ECDH_RSA are
   historical.
```

randombit:
> Regardless of which hash algorithm TLS Attacker sends, let it SHA-1 or SHA-512, the tls_server always sends the rsaWithSha256 certificate.

Wow, I had completely missed this in the TLS spec, what a dumb idea to conflate the issue of which algorithms to use during the handshake happening now with those which were used (by several different parties at points in the past) to sign the cert chain. I don't think that's a good idea. It seems like many other implementations also ignores this RFC requirement - if I set `signature_hashes=SHA-512` many servers reply with certs signed with SHA-256 or even SHA-1, but correctly authenticating using a RSA/SHA-512 signature. (Though a few reply authenticating with SHA-1 or SHA-256 signatures instead)

Anyway, given the server can send either the RSA/SHA-256 cert chain it has or just give up and send an alert, it might as well send the certs and let the client decide - no point in any change there. On the client side, verifying which hashes were seen during certificate verification can be performed by an application overriding `Credentials_Manager::verify_certificate_chain` and examining the result of `Path_Validation_Result::trusted_hashes()` after verification.

I'm skeptical of making that the default check though, because it effectively forces us to allow SHA-1 *during the handshake* simply because many certs are historically signed with SHA-1, which means a MITM who can break SHA-1 can probably attack our handshake. Even if SHA-1 remained disabled by default, in practice most applications would be forced to allow it to deal with existing SHA-1 cert chains. [this is bad]

cordney:
Ha, good point. @jurajsomorovsky what do YOU think?

jurajsomorovsky:
I think if you want to enforce usage of, for example, RSA-SHA256 you have to do this in all algorithms used to authenticate the handshake. Otherwise, your security breaks to the least secure crypto algorithm.

Especially when we are considering usage of signature algorithms in certificates and TLS handshakes, breaking a signature algorithm in a certificate is easier, because:
- it is static and you have to do this only once (vs. a TLS handshake dynamic signature generation where you have to perform active MitM attacks) 
- in a TLS authentication (especially in 1.2 server authentication) you have not that much freedom to construct collisions, in comparison to certificates.

Given an attacker can break SHA-1, he would probably rather attack the server certificate than attack one of the TLS handshakes. This is easier. Therefore, if you want to achieve a specific security level with the used signature and hash algorithms, you have to check the algorithms in certificates as well. Allowing insecure certificate algorithms but forcing secure ServerKeyExchange signatures makes not that much sense to me.

I think the most relevant reference would be the recent SLOTH attack: https://www.internetsociety.org/sites/default/files/blogs-media/transcript-collision-attacks-breaking-authentication-tls-ike-ssh.pdf

randombit:
Good points. I'm going to merge this PR as it stands as an improvement to master, but leave #619 open to track enforcing the restrictions on the cert chain by the client.

Issue #471, opened 2016-04-07
Certificate_Store find_cert method returns pointer to store-owned X509_Certificate object


al45tair:
The `find_cert` method on `Certificate_Store` is currently defined as follows:

```c++
virtual const X509_Certificate*
         find_cert(const X509_DN& subject_dn, const std::vector<byte>& key_id) const = 0;
```

The documentation seems to imply that the original intention was to return `std::vector<X509_Certificate>` from these kinds of methods, but this method is returning a pointer. Unfortunately, looking at the implementation of `Certificate_Store_In_Memory`, it becomes clear that the code currently expects the *store* to own the pointer. This is problematic if you want to implement an external certificate store, e.g. based on the OS X keychain or on some external server, because in those cases the `X509_Certificate` object would have to be allocated on the fly and there is no way for the store to know when it should deallocate such an object.

randombit:
This area of the code is in need of some help. As can be seen by there not being any implementation for system-level cert stores, or even just a basic sqlite3-backed certstore, it has not been driven by application requirements overmuch but is just kind of the minimum interface that is necessary for testing and simple applications. And, as you rightly point out, it's too simple to even be extended properly for such real world uses. :/

Part of the problem is Certificate_Store represents two different things - a way of doing lookups on certs, generally and also being something to pass around with the idea of "my trust roots can be looked up via this particular object". This confusion also affects the API - does any real application actually need more than one Certificate_Store? Certainly not when they are all in memory anyway, but still `x509_path_validate` consumes a `vector` of them.

Returning a `shared_ptr<X509_Certificate>` from `find_cert` would be a good start since that would accommodate either in-memory or DB-backed stores.

Amusingly? the main user of this class, path validation, makes a full copy of the `X509_Certificate` object immediately, while building the path. It could instead keep the `shared_ptr`s, and avoid the copy.

al45tair:
OK, in that case I might take a look at this area and see if I can come up with a sensible patch (including new documentation). I’m looking to use Botan’s TLS implementation in a cross-platform app, and I’d like it to use the system certificate stores (whether the System keychain on OS X or the Certificate Stores on Windows) rather than shipping around its own trust roots.

flanfly:
Any update on this?

randombit:
I have not touched this issue. But I'm open to API changes to better interop with system/token PKI stores.

al45tair:
[...]

PR #631, opened 2016-09-22
Certificate store in SQL

flanfly:
This adds an SQLite3 backed certificate store. See #471 for the initial discussion. The PR also modifies parts of the X509 path validation code and the certificate store interface itself. Now, they all use `shared_ptr<X509_Certificate>` instead of plain pointers.


src/lib/cert/x509/certstor_sql/certstor_sql.cpp
```diff
@@ -0,0 +1,316 @@
...
+#include <botan/auto_rng.h>
+#include <botan/hash.h>
+#include <botan/hex.h>
+
+namespace Botan {
+
+Certificate_Store_In_SQL::Certificate_Store_In_SQL(std::shared_ptr<SQL_Database> db,
+                                                   const std::string& passwd,
+                                                   const std::string& table_prefix)
+: m_database(db), m_prefix(table_prefix), m_password(passwd)
```
randombit: Table prefix is being ignored

src/lib/cert/x509/certstor_sql/certstor_sql.cpp
```diff
@@ -0,0 +1,316 @@
...
+         "UPDATE certificates SET priv_fingerprint = ?1 WHERE fingerprint == ?2");
+
+   stmt2->bind(1,fpr);
+   stmt2->bind(2,cert.fingerprint("SHA-256"));
+   stmt2->spin();
+
+   return true;
+   }
+
+std::string Certificate_Store_In_SQL::fingerprint_key(const Botan::Private_Key& key) const
```
randombit: This function seems to be a dup of X509_Certificate::fingerprint. It could instead be on `Private_Key` and `Public_Key` directly.

src/tests/test_certstor.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+#if defined(BOTAN_HAS_CERTSTOR_SQLITE3)
+   #include <botan/certstor_sqlite.h>
+   #include <botan/sqlite3.h>
+   #include <botan/internal/filesystem.h>
+   #include <botan/pkcs8.h>
+   #include <botan/auto_rng.h>
+   #include <sstream>
+   extern "C" {
+   #include <unistd.h> // unlink()
```
randombit: This include for `unlink` will be trouble, not sure what to do here past adding an appropriate wrapper in filesystem.h and hoping C++17 gets here eventually.

(Also how is this even compiling on Windows/MinGW?)

randombit:
Looking at the code again I think supporting a table prefix would be a pain with the current SQL interface. The DB wrapper could/should be updated to better support this usage, but for now feel free to just drop the prefix constructor argument and leave it at that, unless you feel motivated to support this feature.

flanfly:
I fixed the problem with the table prefix and rebase'd everything. I don't really know what made that one ARM test fail.

randombit:
Merged. Thanks!

PR #634, opened 2016-09-23
Self-issued vs. self-signed certificates

cordney:
We came across a chain with two certificates during debugging, which was generated more by accident then on purpose. I could reproduce the behaviour by generating fresh certificates the same way:

```
#0:
        Issuer: CN=Test CA
        Validity
            Not Before: Sep 23 13:43:24 2016 GMT
            Not After : Sep 23 13:43:24 2017 GMT
        Subject: CN=Test CA
#1:
        Issuer: CN=Test CA
        Validity
            Not Before: Sep 23 13:43:37 2016 GMT
            Not After : Sep 23 13:43:37 2017 GMT
        Subject: CN=Test CA
```

Note that both certs have the same DN and cert#1 therefore has subject_dn == issuer_dn, whereas cert#0 is self-signed and cert#0 also signed cert#1. cert#1 is what is called a self-issued certificate. When trying to verify this chain, `path_validate()` will output a signature error. We found that this is because the self-issued cert (1) would be taken for a self-signed certificate and (2) `find_issuing_cert()` would find the same self-issued certificate that we want to verify, generating the signature error during signature verification.
    
To fix, we now first identify a certificate as self-signed only if subject_dn == issuer_dn AND if we can verify the cert signature with it's own key. Verification will bring some extra costs, but we only do it once, in `X509_Certificate`'s constructor. Second, we make sure `find_issuing_cert()` does not return the very same certificate we want to verify. This should be no problem, since path validation currently does not seem to support validating a self-signed certificate.

This PR marked as WIP, because I am not really sure if these changes break any existing code, though tests run fine.

Also in this PR: Make CLI's `sign_cert` key password parameter optional.

### DIFF
--- a/src/cli/x509.cpp
+++ b/src/cli/x509.cpp
@@ -30,11 +30,19 @@ class Sign_Cert final : public Command
       void go() override
          {
          Botan::X509_Certificate ca_cert(get_arg("ca_cert"));
+         std::unique_ptr<Botan::PKCS8_PrivateKey> key;
 
-         std::unique_ptr<Botan::PKCS8_PrivateKey> key(
-            Botan::PKCS8::load_key(get_arg("ca_key"),
-                                   rng(),
-                                   get_arg("ca_key_pass")));
+         if(flag_set("ca_key_pass"))
+            {
+            key.reset(Botan::PKCS8::load_key(get_arg("ca_key"),
+                     rng(),
+                     get_arg("ca_key_pass")));
+            }
+         else
+            {
+            key.reset(Botan::PKCS8::load_key(get_arg("ca_key"),
+                     rng()));
+            }
 
          if(!key)
             throw CLI_Error("Failed to load key from " + get_arg("ca_key"));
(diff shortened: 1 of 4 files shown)


randombit:
This change makes sense to me. An additional check could be made on IssuerKeyId/SubjectKeyId extensions, but these are not always set.

One interesting interaction - it *will* break all MD2 self-signed trust roots, since the hash will be unavailable to perform validation. Of the 171 trust roots in my systems `/etc/ssl/certs`, just 1 uses MD2, a 1024 bit RSA key published in early 1996 and valid until 2028.

```
    Signature Algorithm: md2WithRSAEncryption
        Subject: C=US, O=VeriSign, Inc., OU=Class 3 Public Primary Certification Authority
        Issuer: C=US, O=VeriSign, Inc., OU=Class 3 Public Primary Certification Authority
        Validity
            Not Before: Jan 29 00:00:00 1996 GMT
            Not After : Aug  1 23:59:59 2028 GMT
```

I would guess the NSA finished factoring this key some years ago. ;)

randombit:
FWIW I'm fine with merging this if you are. The extra overhead during parsing seems ok when limited to issuer==subject case. It will slow down loading a large trust root a bit, but RSA verify is fast.

cordney:
Thx

PR #637, opened 2016-09-29
make RC4 optional: comply with openssl w/o RC4 cipher

t0b3:
Help us fade out RC4. We shouldn't support rely on it, it just builds an attack vector due to it's weak stream ciphers. 
With this patch accepted, openssl could be installed with ./configure --no-rc4, dropping all RC4 attack surface at once.

### DIFF
--- a/src/lib/stream/stream_cipher.cpp
+++ b/src/lib/stream/stream_cipher.cpp
@@ -24,7 +24,7 @@
   #include <botan/ofb.h>
 #endif
 
-#if defined(BOTAN_HAS_RC4)
+#if defined(BOTAN_HAS_RC4) && !defined(OPENSSL_NO_RC4)
   #include <botan/rc4.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


randombit:
Nice, thanks

src/lib/stream/stream_cipher.cpp
```diff
@@ -24,7 +24,7 @@
   #include <botan/ofb.h>
 #endif
 
-#if defined(BOTAN_HAS_RC4)
+#if defined(BOTAN_HAS_RC4) && !defined(OPENSSL_NO_RC4)
```
randombit: This is wrong - just building against an OpenSSL with RC4 disabled should not implicitly disable RC4 in the library itself.

t0b3: I agree on giving package maintainers explicit control would be nice.
May I close this PR and request [641](https://github.com/randombit/botan/pull/641) ?

randombit:
Closing in favor of #641

PR #639, opened 2016-09-29
2.3.1 tls tests publish

jurajsomorovsky:
Added some new positive and negative TLS tests for parsing and processing protocol messages.

For this purpose, I needed to insert `BOTAN_DLL` in a few of the protocol messages (a better way would be to specify a new export macro for botan tests as we discussed a few months ago...but for that I would need to have more experience with the botan build system).

My negative tests explicitly check exception message contents thrown while parsing invalid protocol messages. I hope this is a valid test approach.


jurajsomorovsky:
Sorry, too many commits included (scheduled for future pull requests)

PR #640, opened 2016-09-29
Extended tls tests

jurajsomorovsky:
Added some new positive and negative TLS tests for parsing and processing protocol messages.

For this purpose, I needed to insert `BOTAN_DLL` in a few of the protocol messages (a better way would be to specify a new export macro for botan tests as we discussed a few months ago...but for that I would need to have more experience with the botan build system).

My negative tests explicitly check exception message contents thrown while parsing invalid protocol messages. I hope this is a valid test approach.


randombit:
Exposing the TLS protocol messages to applications is something I have already been considering, so annotating them as `BOTAN_DLL` for tests is not really a problem. Currently, the application receives a callback with a `Handshake_Message&`, but has no way to dynamic cast the reference to examine or log the actual message, all it can do is get the type and the serialized bits. The only reason `tls_messages.h` is not exposed for this right now is due to ABI concerns, as those struct change size somewhat frequently. But things are slowly seeming more stable in this regard.

Thank you for the new tests. Re the exception messages, got to admit it kind of smells, but there isn't a cleaner way to do this as the test framework currently stands. So, good enough for now. And, tests. ;)

PR #641, opened 2016-09-30
fix compiler error: openssl w/o rc4

t0b3:
Compiling botan with disabled rc4 module fails in case of openssl w/o rc4...
Error:
./src/lib/prov/openssl/openssl_rc4.cpp:15:25: fatal error: openssl/rc4.h: No such file or directory
 #include <openssl/rc4.h>


t0b3:
Travis failure seems not be reproducible
Travis passed after dummy change https://travis-ci.org/randombit/botan/jobs/164227416

randombit:
Merged, thank you

PR #642, opened 2016-10-01
[For Review] Move TLS CBC+HMAC construction to AEAD

randombit:
Adds an (internal only) AEAD_Mode which implements the TLS CBC+HMAC construction.

Currently the old version of the CBC code in tls_record.cpp remains, but is guarded by aborts. This is just for ease of review and backchecking, obviously before final merge all of this now-dead code will be removed and the record layer simplified accordingly; tls_record.cpp should be cut in about half and will only interact with an `AEAD_Mode`.

This split will also make it easier to test/fuzz the TLS CBC decryptor, though I'm not planning on adding that here.

With the addition of logic for handling CBC mode (in both 1.0 implicit and 1.1/1.2 explicit IV modes), the nonce handling code in the TLS record layer is getting contorted. I have a plan to address that in a later work.

### DIFF
--- a/src/lib/tls/info.txt
+++ b/src/lib/tls/info.txt
@@ -22,6 +22,7 @@ tls_version.h
 </header:public>
 
 <header:internal>
+tls_cbc.h
 tls_extensions.h
 tls_handshake_hash.h
 tls_handshake_io.h
--- a/src/lib/tls/tls_cbc.cpp
+++ b/src/lib/tls/tls_cbc.cpp
@@ -1,0 +1,403 @@
+/*
+* TLS CBC Record Handling
+* (C) 2012,2013,2014,2015,2016 Jack Lloyd
+*     2016 Juraj Somorovsky
+*     2016 Matthias Gierlings
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/internal/tls_cbc.h>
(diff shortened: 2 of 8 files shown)


randombit:
It seems that very rarely during tests a packet is being rejected with a MAC validation failure, seen in https://travis-ci.org/randombit/botan/jobs/164263858 and https://travis-ci.org/randombit/botan/jobs/164263865. In the first instance it is a DTLS failure-to-handshake, I am guessing this was due to the finished message being rejected. I have been unable to replicate this so far, so it's also not clear if this is the fault of the encryption or decryption code here. I will switch back to using the previous decryption code, and see what happens.

randombit:
Issue fixed I think (though we'll see on CI) - the CBC code was assuming the IV was part of the message length, which it is not in the changed context (AEAD_Mode vs part of the TLS record layer) and so falsely rejecting very short records as being too short. Very rarely the tests would randomly generate such a short message and fail - changed the test to always split the sent records to make such bugs easier to repro.

randombit:
Modulo rebasing to clean up history a bit I think this change is good to go (assuming CI comes back green)

randombit:
TLS EtM missing the check on the padding is also present in master.

randombit:
Rebased to clean up history. Last call for comments, plan on merging this in the next few days to allow continued TLS work.

PR #643, opened 2016-10-01
Remove support for DSA authentication from TLS

randombit:
Long promised...

### DIFF
--- a/src/lib/tls/msg_cert_req.cpp
+++ b/src/lib/tls/msg_cert_req.cpp
@@ -25,8 +25,7 @@ std::string cert_type_code_to_name(byte code)
       {
       case 1:
          return "RSA";
-      case 2:
-         return "DSA";
+      // case 2 is DSA
       case 64:
          return "ECDSA";
       default:
@@ -38,8 +37,6 @@ byte cert_type_name_to_code(const std::string& name)
    {
    if(name == "RSA")
       return 1;
-   if(name == "DSA")
-      return 2;
    if(name == "ECDSA")
       return 64;
 
@@ -57,7 +54,7 @@ Certificate_Req::Certificate_Req(Handshake_IO& io,
                                  const std::vector<X509_DN>& ca_certs,
                                  Protocol_Version version) :
    m_names(ca_certs),
-   m_cert_key_types({ "RSA", "DSA", "ECDSA" })
+   m_cert_key_types({ "RSA", "ECDSA" })
    {
    if(version.supports_negotiable_signature_algorithms())
       {
--- a/src/lib/tls/tls_ciphersuite.cpp
+++ b/src/lib/tls/tls_ciphersuite.cpp
@@ -139,13 +139,7 @@ bool Ciphersuite::is_usable() const
 #endif
       }
 
-   if(sig_algo() == "DSA")
-      {
-#if !defined(BOTAN_HAS_DSA)
-      return false;
-#endif
-      }
-   else if(sig_algo() == "ECDSA")
+   if(sig_algo() == "ECDSA")
       {
 #if !defined(BOTAN_HAS_ECDSA)
       return false;
--- a/src/lib/tls/tls_extensions.cpp
+++ b/src/lib/tls/tls_extensions.cpp
@@ -398,8 +398,7 @@ std::string Signature_Algorithms::sig_algo_name(byte code)
       {
       case 1:
          return "RSA";
-      case 2:
-         return "DSA";
+      // case 2 is DSA
       case 3:
          return "ECDSA";
       default:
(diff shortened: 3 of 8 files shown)


cordney:
Can you give a short reasoning why you are removing this now? Just curious.

randombit:
It seems that DSA certificates are all but dead in practice, thus continuing to support them just increases our attack surface with no corresponding gain in functionality.

randombit:
The actual code savings/simplification here turned out to be pretty minimal so I'm fine with holding off on this for now and initially just disabling DSA from the default policy (what it is still doing allowed by default I don't even know), assuming someone can point me to a server somewhere with a DSA certificate, ideally one which roots in a trusted CA but self signed is fine, so I can test the handshake actually works. ;) There are no (0) DSA CA certificates in my `/etc/ssl/certs` and from a quick search I could not find a CA that issues DSA certs anymore. I can believe there are some long-lived server DSA certs still floating around, but probably not very many.

A restrictive feature set and restrictive default policy saved us a couple of times from cross-stack TLS attacks; for examples 1024 bit DH primes have been required by default since 1.11.0 (2012) and 3DES has been disabled by default since 1.11.2  (2013),  neatly avoiding Logjam (2015) and Sweet32 (2016). I was late in removing RC4 support from TLS last year, after RC4 was already publicly broken. I would like to stay ahead of the breaks coming out next year (or the attack which was developed last year, but wasn't published), and the easiest way to do that is to disable everything possible, or even better remove it outright so no combination of policy misconfiguration and protocol downgrade bugs can result in an exploit.

For DSA certs it seems easy to just remove, rather than disable, because once the certs are no longer being issued and the active certs expire, no further use of the ciphersuites is even possible.

cordney:
Thanks for the detailed explanation. I'll check with a few people here and come back to you this week.

jurajsomorovsky:
I am not sure I can contribute here, but maybe these statistics from censys could help you. There are:
- more than 3000 DSA certificates: https://censys.io/certificates?q=parsed.subject_key_info.key_algorithm.name%3A+DSA
- 23 out of them are nss trusted certificates: https://censys.io/certificates?q=current_valid_nss%3A+true+AND+parsed.subject_key_info.key_algorithm.name%3A+DSA

According to the newest research results, you should only accept at least 2048-bit parameters: http://eprint.iacr.org/2016/961

cordney:
People at BSI I talked with are asking if there are any new weaknesses or attacks or problems with the implementation that justify removal. You spoke of breaks coming out next year and an attack that wasn't published last year, can you point us to more information on that?

randombit:
Apologies, what I meant by that was epistemicly speaking we cannot really say if current best-practice TLS does or does not a cross protocol/cross ciphersuite attack involving DSA such as the DH/ECDH cross ciphersuite attack from "A Cross-Protocol Attack on the TLS Protocol" http://homes.esat.kuleuven.be/~fvercaut/papers/ACM2012.pdf

> ABSTRACT
> This paper describes a cross-protocol attack on all versions
> of TLS; it can be seen as an extension of the Wagner and
> Schneier attack on SSL 3.0. The attack presents valid explicit
> elliptic curve Diffie-Hellman parameters signed by a
> server to a client that incorrectly interprets these parameters
> as valid plain Diffie-Hellman parameters. Our attack
> enables an adversary to successfully impersonate a server to
> a random client after obtaining 240 signed elliptic curve keys
> from the original server. While attacking a specific client is
> improbable due to the high number of signed keys required
> during the lifetime of one TLS handshake, it is not completely
> unrealistic for a setting where the server has high
> computational power and the attacker contents itself with
> recovering one out of many session keys. We remark that
> popular open-source server implementations are not susceptible
> to this attack, since they typically do not support the
> explicit curve option.

But we know that we can reduce the probability of cross protocol attacks by minimizing the number of possible interactions - for example if the client and server do not support DH but only ECDH for key exchange, then the attack referenced above is not possible. There doesn't seem any particular reason to me that DSA auth is intriniscally less secure than RSA auth in either the TLS v1.2 protocol or in TLS as implemented by Botan, *but* it is unlike RSA an auth method we can get away with removing, and in doing so reduce the combinatoric possibilities.

Additionally according to https://en.wikipedia.org/wiki/Comparison_of_TLS_implementations [*] DSA is not supported in modern SecureTransport, and in even the most recent version of Microsoft's SChannel it is limited to 1024 bits, which is not an acceptable security level anymore.

[*] Which admittedly seems to have many errors for Botan's info, so who knows if the rest is actually correct.

randombit:
In 4c972845183 I changed the default policy to disable DSA (as well as static RSA), for now DSA can remain supported (if deprecated), and be removed in the next dev cycle as part of adding TLS 1.3

neusdan:
That's really helpful for us that it's not removed completely for now.  thanks.

Issue #636, opened 2016-09-26
Negotiating custom EC curves in TLS


cordney:
We must be able to use custom (prime) curves in the ECDHE key exchange in TLS. Additionally, the points must be transferred in compressed format.

From looking into [RFC 4492](https://tools.ietf.org/html/rfc4492), it seems we must implement two things:

1) `arbitrary_explicit_prime_curves` in the [Supported Elliptic Curves Extension](https://tools.ietf.org/html/rfc4492#page-12)
2) [Supported Point Formats Extension](https://tools.ietf.org/html/rfc4492#page-13)

After looking into the code, I would propose the following changes:

On (1), I would change `TLS_Policy::allowed_ecc_curves()` to return a `std::vector<EC_Group>` instead of a `std::vector<std::string>`. `EC_Group` has a `std::string` constructor, so it would have the same semantics as before. `TLS::Supported_Elliptic_Curves` class would then match `EC_Group` to the named curve id if the OID is known, otherwise it would send the `arbitrary_explicit_prime_curves` id. On the server side, in `Server_Key_Exchange::Server_Key_Exchange()`, `TLS_Policy::choose_curve()` would also return an `EC_Group` instead of an `std::string`. If the `EC_Group` can not be looked up using the OID, additional code would encode an `ECParameters` struct for `case explicit_prime:`.

On (2), the Supported Point Formats Extension must be implemented. `PointGFp` already supports compressed and uncompressed points. `EC_Group` by default always encodes points uncompressed, so maybe there should be an overload specifying the point compression type. I am not sure how to get the uncompressed/compressed decision to the user. Maybe `TLS_Policy::allowed_ecc_curves()` should return an `std::vector<std::pair<PointGFp::Compression_Type, EC_Group>>`?

randombit:
> We must be able to use custom (prime) curves in the ECDHE key exchange in TLS.

I'll admit I can see a certain logic, for some very specialized applications, in using a randomly generated 'in-house' curve. But:

> 1) arbitrary_explicit_prime_curves in the Supported Elliptic Curves Extension

Strong NAK on this. Arbitrary curves are not supported by any other implementation that I know of, and there is a reason for that. Smart protocol stacks don't accept arbitrary ECC curves from strangers. The situation with TLS 1.2 DH in that regard is bad enough.

[...]

PR #645, opened 2016-10-03
Tls compressed points

cordney:
Adds support for the TLS Supported Point Format Extension per RFC 4492. Users can control compressed/uncompressed via a new `TLS_Policy::use_ecc_point_compression()` toggle. If compression is preferred by either party, client kex and server kex send the ECDH public point as compressed, respectively. Note that per RFC 4492 also client and server certificate messages have to respect the preference of the other party, but we leave this to the user and currently don't check the certificate.

First part to implement #636.


src/lib/tls/msg_server_kex.cpp
```diff
@@ -85,7 +85,15 @@ Server_Key_Exchange::Server_Key_Exchange(Handshake_IO& io,
...
 
-      append_tls_length_value(m_params, ecdh->public_value(), 1);
+      // follow client's preference for point compression
+      if(state.client_hello()->prefers_compressed_ec_points())
+         {
+         append_tls_length_value(m_params, ecdh->public_value(PointGFp::COMPRESSED), 1);
+         }
+      else
+         {
+         append_tls_length_value(m_params, ecdh->public_value(PointGFp::UNCOMPRESSED), 1);
```
randombit: Same ?:

src/lib/tls/tls_extensions.cpp
```diff
@@ -353,6 +356,51 @@ Supported_Elliptic_Curves::Supported_Elliptic_Curves(TLS_Data_Reader& reader,
...
+
+   // if we send this extension, we prefer compressed points,
+   // otherwise we don't send it (which is equal to supporting only uncompressed)
+   buf.push_back(ANSIX962_COMPRESSED_PRIME);
+
+   // if this extension is sent, it MUST include uncompressed (RFC 4492, section 5.1)
+   buf.push_back(UNCOMPRESSED);
+
+   buf[0] = static_cast<byte>(buf.size()-1);
+
```
randombit: Why not `return std::vector<byte>{2, ANSI_X962_COMPRESSED_PRIME, UNCOMPRESSED}` ?

cordney: Nice one, thx.

src/lib/tls/msg_server_kex.cpp
cordney: Fixed

src/lib/tls/msg_client_kex.cpp
```diff
@@ -172,7 +172,15 @@ Client_Key_Exchange::Client_Key_Exchange(Handshake_IO& io,
...
 
-         append_tls_length_value(m_key_material, priv_key.public_value(), 1);
+         // follow server's preference for point compression
+         if(state.server_hello()->prefers_compressed_ec_points())
+            {
+            append_tls_length_value(m_key_material, priv_key.public_value(PointGFp::COMPRESSED), 1);
+            }
+         else
+            {
+            append_tls_length_value(m_key_material, priv_key.public_value(PointGFp::UNCOMPRESSED), 1);
```
cordney: Fixed

cordney:
Also fixed the tls_messages tests. Somehow this passed unnoticed locally.

cordney:
Strange test failure on cross-arm32:

```
FFI ran 226 tests 1 FAILED
Failure 1: FFI botan_privkey_export_encrypted unexpectedly failed with error code -10
```

randombit:
I've been chasing that FFI test error for a while and thought I had fixed it in 4e42b0186b but it still recurs. It is not ARM specific but does seem to be related to how slow the machine is during the test run - it is much more commonly seen in QEMU or AppVeyor builds. I have never been able to repro it locally.

PR #646, opened 2016-10-03
MSVC static_assert fail in debug mode

webmaster128:
I recently found that this statement fails in MSVS 2015, debug:

```
static_assert(std::is_integral<T>::value, "secure_allocator supports only integer types");
```

Let's see if we can reproduce this on AppVeyor.

### DIFF
--- a/src/lib/base/secmem.h
+++ b/src/lib/base/secmem.h
@@ -24,7 +24,16 @@ template<typename T>
 class secure_allocator
    {
    public:
+      /*
+      * Assert exists to prevent someone from doing something that will
+      * probably crash anyway (like secure_vector<non_POD_t> where ~non_POD_t
+      * deletes a member pointer which was zeroed before it ran).
+      * MSVC in debug mode uses non-integral proxy types in container types
+      * like std::vector, thus we disable the check there.
+      */
+#if !defined(_ITERATOR_DEBUG_LEVEL) || _ITERATOR_DEBUG_LEVEL == 0
       static_assert(std::is_integral<T>::value, "secure_allocator supports only integer types");
+#endif
 
       typedef T          value_type;
 
--- a/src/scripts/ci/appveyor.yml
+++ b/src/scripts/ci/appveyor.yml
@@ -4,6 +4,10 @@ platform:
   - x86
   - x86_amd64
 
+configuration:
+  - Release
+  - Debug
+
 environment:
   matrix:
     - COMPILER: msvc-12.0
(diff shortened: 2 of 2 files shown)


webmaster128:
Here we go. MSVC fails at

```
static_assert(std::is_integral<T>::value, "secure_allocator supports only integer types");
```

for `T = Botan::byte = std::uint8_t`. Locally I had the same issue with `T = Botan::u64bit`.

Does someone have a clue what the issue is here?

randombit:
What the heck?!?

Good news is this `static_assert` is just to prevent someone from doing something that will probably crash anyway (like `secure_vector<non_POD_t>` where ~non_POD_t deletes a member pointer which was zeroed before it ran) so we can just skip it on MSVC in debug config if we must. But why wouldn't it work ...

webmaster128:
@cordney @neusdan, you do a lot of MSVC development, right? Any idea what is going on here?

webmaster128:
`std::_Container_proxy` seems to be an internal type used when `_ITERATOR_DEBUG_LEVEL` is in place (confirm http://stackoverflow.com/a/31805705) and `std::is_integral<std::_Container_proxy>::value` is false.

I just found the following statement about allocators, but I have no idea if this is correct:

>  Note that an allocator must be capable of allocating instances of any type - that's the whole point of rebind. It's illegal for an allocator to rely on T having any particular methods.

https://social.msdn.microsoft.com/Forums/vstudio/en-US/dda45867-fc61-4fd2-8fe3-deec1f4c6e02/problems-with-stdcontainerproxy-in-debug-version?forum=vclanguage

Can it be true, that allocators must not restrict the type that is allocated?

randombit:
Thanks for digging into this. I think what you have here is for now the best option; if allocators really truly can't restrict their Ts we'd probably have to just revert back to a hand-rolled vector type (as was used in v1.10 and older versions)

webmaster128:
:+1: I tried to find a reliable source for how allocators are supposed to work but I did not find anything I am capable of understanding. Thanks for the clarification about the purpose of the assert. I copied your description from here and hope you don't mind.

cordney:
Sorry for not having the chance to look into this further. On MSVC I can not help, as personally I do Linux development the whole day. @neusdan is working on Windows stuff. :)

From taking a look into the [standard](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3337.pdf), I understand that restricting to certain types is ok:

```
§ 17.6.3.5

5 An allocator may constrain the types on which it can be instantiated and the arguments for which its
construct member may be called. If a type cannot be used with a particular allocator, the allocator class
or the call to construct may fail to instantiate.
```

PR #648, opened 2016-10-03
Resolve #647

### LINKED ISSUE #647: Botan attempts to open /dev/urandom for writing
See: https://github.com/randombit/botan/blob/master/src/lib/rng/system_rng/system_rng.cpp#L77

Botan is attempting to open /dev/urandom as read-write, which (correct me if I'm wrong) is not correct -- I know of no reason to write to /dev/urandom (what would that even do?).

I note that Ubuntu's new package format, Snappy, sandboxes apps by default, and the AppArmor profile it uses forbids opening /dev/{,u}random in write mode. This prevents apps which use Botan from being packaged with Snappy.

nathanhourt:
Implement a backoff approach to opening the system RNG: if opening read-write fails, try to open read-only. This will allow the RNG to be used, but attempts to add entropy will fail. If opening as read-only also fails, only then throw an exception.


randombit:
Merged, thank you for the bug report and patch.

PR #653, opened 2016-10-06
[Review needed] OCSP and certificate validation refactor

randombit:
Addresses some issues raised in #471 with how TLS accesses the network in order to make OCSP requests.

Moves the verify callback from the credentials manager to TLS::Callbacks. Given that (unlike every other function on Credentials_Manager) it is called to process session specific and attacker controlled inputs, instead of just returning the appropriate cred, it makes more sense to have it there.

Change PKIX path validation to take a function which makes HTTP calls for OCSP, returning a `future` representing the body. Not currently used for CRL access but could be.

Makes some changes to OCSP, including splitting up the operations of parsing the response and checking the signature.

Adds support for basic BSD sockets in addition to Boost.Asio for HTTP access. So, OCSP will now be working out of the box on Linux/BSD/OS X builds. And once I actually had OCSP enabled by default, running `tls_client` quickly showed that there were problems handling responses from common responders which was addressed in 865b9fc.

@al45tair I'd love your take on if this new API would have worked for you with libuv


randombit:
I thought I was being clever in e64d709 by exposing a single HTTP callback (for OCSP+ CRLs requests) instead of having 2 functions returning future<OCSP::Response> or future<CRL>. The problem is, the second approach allows an application to maintain effective CRL and OCSP caches, which is a very useful thing. If all the application sees is the HTTP request it is forced to actually make the request, with no control over how it was formed. In addition due to limitations of util's HTTP client the request is always an HTTP/1.0, preventing any effective connection pooling for multiple requests to the same responder (a very common scenario I would imagine). I will make this change before merging.

Caching OCSP responses and CRLs would be a good addition to the SQL cert store too. I should probably add this also, to ensure everything works as expected.

randombit:
This branch ended up going all over the place, but I think what is here (in c1cfb82) will work for most users including fixing IncludeOS builds by avoiding any use of `future/async` outside of certain functions that are ifdef'ed out entirely for non-threaded platforms. It touches a lot of the OCSP and path validation code, so even a basic review at this point would be appreciated.

I'm going to rebase one more time to split up the commit into 2 or 3 (tomorrow), and may try to add at least some basic parsing/signature checking OCSP tests, but the major work is done. I am going to try to write OCSP stapling soon on top of this branch, but it will be in a different PR.

OCSP is enabled by default for path validation currently. It might be better to be disabled by default and explicitly opt-in by passing a non-zero timeout, since the extra network call can result in unexpected latency problems which tends to make people unhappy.

randombit:
This is holding up OCSP stapling which I want to also have merged before the next release, so I plan to merge this PR on Friday.

cordney:
I'll try to take a look at this tomorrow.

src/lib/x509/x509path.cpp
```diff
@@ -154,136 +115,486 @@ check_chain(const std::vector<std::shared_ptr<const X509_Certificate>>& cert_pat
...
+      }
+
+   std::vector<std::set<Certificate_Status_Code>> status =
+      PKIX::check_chain(cert_path, ref_time,
+                        hostname, usage,
+                        restrictions.minimum_key_strength(),
+                        restrictions.trusted_hashes());
+
+   if(path_building_result != Certificate_Status_Code::OK)
+      status[0].insert(path_building_result);
```
cordney: You checked for `if(path_building_result != Certificate_Status_Code::OK)` already in line 549 above and `return` from there, so this line here will never be triggered.

src/lib/x509/x509path.h
```diff
@@ -39,7 +46,8 @@ class BOTAN_DLL Path_Validation_Restrictions
       * @param minimum_key_strength is the minimum strength (in terms of
       *        operations, eg 80 means 2^80) of a signature. Signatures
       *        weaker than this are rejected.
-      * @param ocsp_all_intermediates
+      * @param ocsp_all_intermediates Make OCSP requests for all CAs as
```
cordney: Now we can also fix the `FIXME add doc` for `ocsp_all_intermediates()`

src/lib/x509/x509path.cpp
cordney: Thinking about it, returning early makes more sense. That's also what the old implementation does.

src/lib/x509/x509path.h
```diff
@@ -151,52 +160,121 @@ class BOTAN_DLL Path_Validation_Result
...
+*/
+std::vector<std::set<Certificate_Status_Code>>
+BOTAN_DLL check_chain(const std::vector<std::shared_ptr<const X509_Certificate>>& cert_path,
+                      std::chrono::system_clock::time_point ref_time,
+                      const std::string& hostname,
+                      Usage_Type usage,
+                      size_t min_signature_algo_strength,
+                      const std::set<std::string>& trusted_hashes);
+
+std::vector<std::set<Certificate_Status_Code>>
```
cordney: API documentation should be added for this and the other check functions, too. I also think for ordinary library users, the documentation should be more expressive on which functions to use for path validation. Previously, there was only `x509_path_validate()` and all the others were wrapped inside the .cpp, now there are some more which also "perform certificate validation". Library users will wonder which ones to use. +1 for making these available for tests.

src/lib/base/buf_comp.h
```diff
@@ -101,6 +101,13 @@ class BOTAN_DLL Buffered_Computation
          return output;
          }
 
+      std::vector<byte> final_stdvec()
```
randombit: Mostly because it's very unfortunate at runtime - many times a hash/MAC output is not secret and `unlock(final())` bounces an allocation+deallocation in the pool allocator for no reason. Many places in the code instead use something like

```
std::vector<byte> out(h->output_length())
h->final(out.data());
```

to avoid this, they could instead use the (definitely unfortunately named) `final_stdvec` instead. I'm fine with painting this bikeshed a different color.

src/lib/x509/certstor.h
```diff
@@ -31,6 +31,15 @@ class BOTAN_DLL Certificate_Store
          find_cert(const X509_DN& subject_dn, const std::vector<byte>& key_id) const = 0;
 
       /**
+      * Find a certificate by searching for one with a matching SHA-1 hash of
+      * public key. Used for OCSP.
+      * @param key_hash SHA-1 hash of the subject's public key
+      * @return a matching certificate or nullptr otherwise
+      */
+      virtual std::shared_ptr<const X509_Certificate>
+         find_cert_by_pubkey_sha1(const std::vector<byte>& key_hash) const = 0;
```
randombit: Unfortunately SHA-1 seems to be baked into OCSP especially in the key hash which is used to identify the response signer. Even in the latest OCSP RFC this is SHA-1 only (no algo id possible) and many responders on the net prefer providing a key hash instead of a DN. I don't see any way of implementing (a useful amount of) OCSP without SHA-1.

src/lib/x509/x509path.cpp
randombit: You're right, the early return is the right one to have. I think originally the first `if` only returned early on certain errors, then I realized that was a bad idea - if building the chain failed forget anything else - but forgot the remove the now redundant logic.

cordney: OK

src/lib/x509/certstor.h
randombit: It definitely is not pretty. We could make SHA-1 optional and have requests be issued with SHA-256 (causing some responders to reject our requests) and ignore/fail any replies that are identified with a key hash. That's a substantial hit to functionality, but maybe some closed systems who control their OCSP responders would still find it usable.

Alternately we can split out OCSP from X509, but I would really rather not because it would complicate the API (having to have 2 different functions where there is currently 1: one consuming OCSP the other not).

cordney: The first option seems the way to go in the light of real-word responders.

randombit: There are some things we can do to improve things in the near term, even if we can't remove SHA-1 entirely from the build.

- [easy] Change default path validation parameters to ~110 expected strength instead of 80. This disables SHA-1 as a trusted hash during certificate validation, and increases RSA minimum to (just under) 2048 bit keys, instead of 1024 allowed now.

- [easy] Disable SHA-1 HMAC in TLS by default. Not sure if it's a real issue

- [medium] Restructure TLS code to allow disabling MD5/SHA-1 (and coorespondingly disabling TLS v1.0/v1.1 support, since these versions require MD5+SHA-1)

cordney: Setting the default strength to 110 sounds good. Effectively disabling TLS 1.0/1.1 would be a really great addition.

randombit:
Thanks for the review. Addressed all the doc problems (missing docs + Doxygen errors), and exposed a couple of additional small functions to callers in PKIX. PKIX namespace now warns that it is intended for advanced users who need to build their own validation logic.

randombit:
I've exposed the parameter controlling the minimum accepted signature strength for PKIX validation during TLS handshakes as `TLS::Policy::minimum_signature_strength`, default is 110 to allow ~2048 bit RSA.

cordney:
Everything looks good here now, thank you!

PR #660, opened 2016-10-10
Add doxygen mainpage

cordney:
Adds a Crypto++-like doxygen mainpage that covers the most-important classes. Replaces the formerly empty mainpage. The `namespace Botan {` is for doxygen to autolink the classes, otherwise all would have to be prefixed with `Botan::`.

### DIFF
--- a/src/build-data/botan.doxy.in
+++ b/src/build-data/botan.doxy.in
@@ -2,6 +2,7 @@
 
 PROJECT_NAME           = Botan
 PROJECT_NUMBER         = %{version}
+PROJECT_BRIEF          = Crypto&nbsp;and&nbsp;TLS&nbsp;for&nbsp;C&#43;&#43;11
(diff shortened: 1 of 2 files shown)


randombit:
This is great to have, thank you.

Back when the sources were maintained in Monotone, I had a script which automatically rebuilt the Doxygen and website on every push. But since the git move, this refresh only happens manually when I log into the web server over ssh and run multiple commands (so I only ever do it when I must - for a new release). I need to figure out how to re-automate this with git, so any update to master rebuilds the live website again. Anyway, FYI until that is resolved, any Doxygen changes will not be reflected on the website until the next release.

randombit:
BTW for comment-only changes you can include `[ci skip]` in the commit message to skip the build step. Sometimes our CIs can get quite backlogged, by even as much as 12+ hours at times of heavy commit activity.

cordney:
Yeah, I know. Just forget this sometimes.

Issue #486, opened 2016-04-29
Proposed changes for use in regulated applications: request for comments


DanielTBrown:
﻿Botan is a wonderful tool for encrypted communications. I am integrating botan as a TLS library into a new product. The software is subject to DHHS HIPAA privacy regulations, and in some uses may be subject to FDA regulations. The software is embedded, there is no user interface. I have run into several issues, and I would appreciate any comments about my proposed solutions:

**Compilation warnings**

The software must compile without warnings. The existing botan build suppresses Microsoft Visual C++ compiler warnings C4250, C4251, C4267, and C4275. I compiled with all warnings enabled:

Regarding C4267, I submitted pull request #484 to fix about 70 places the source code used implicit casting to narrow a value (e.g. assigning a size_t to a uint16). I made the casting explicit, eliminating the compiler warning messages. This should not change the generated code.

Regarding C4250, I have not fixed this yet. The warning occurs because of specific inheritance in a few C++ classes:

DL_Scheme_PrivateKey
DH_PrivateKey
EC_PrivateKey
ECDH_PrivateKey
ECDSA_PrivateKey
IF_Scheme_PrivateKey
RSA_PrivateKey

I assume that all of these classes have the same problem, and that with some investigation, I can find a simple fix. Any guidance would be appreciated.

Regarding C4251 and C4275, these warnings appear to affect only DLLs, and do not occur when compiling for static linking. So, if configure.py is called with --disable-shared, the warnings can be left enabled. I do not see how to control the /wd4251 and /wd4275 compiler options in configure.py, should I use a '#ifdef' and '#pragma warning' in the C++ source code?

**Some background**

The software uses only the TLS component of botan. The targets are: Microsoft Windows 64-bit (a vestigial target, only for equipment for which no Linux drivers exist), Linux 64-bit (the primary target), and if possible, an embedded ARM microcontroller (for future battery-operated equipment).

I invoke botan configure.py with  '--gen-amalgamation', and '--disable-shared'. Then I compile and link botan as part of the project. The software is statically linked, in order to control the runtime configuration.

The software must conform to USA DHHS HIPAA regulations, which govern personal identifying information. The concept of 'personal identifying information' is interpreted broadly. Even an IP address is considered personal identifying information. A typical penalty for an unintentional first-time violation is apparently US$ 50,000 followed by several years of mandatory monitoring.

[...]

PR #665, opened 2016-10-13
Add IncludeOS target, make fs+threads optional

randombit:
This allows building the library as part of the IncludeOS unikernel (http://www.includeos.org/ https://github.com/hioa-cs/IncludeOS). It adds toggles completely disabling threads and filesystem support since neither exist in that environment. The lock guard and mutex types are abstracted in `<botan/mutex.h>`. These changes will also be useful for other RTOS like environments in the future (#486).

It removes `HTTP::GET_async` entirely instead of `#ifdef`ing it since it was not used by the library at all, and easily implemented by the application.

These changes aren't quite enough to build for IncludeOS, due to use of `std::future` to handle OCSP requests during certificate validation. For now in testing on IncludeOS I'm just disabling OCSP checking. Supporting this on IncludeOS will affect the final shape of #653.

The CLI doesn't build with filesystem support disabled, would be nice to fix but not necessary for IncludeOS (which has no shell and just needs the amalgamation files).

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1019,7 +1019,6 @@ class OsInfo(object):
                         'bin_dir': 'bin',
                         'lib_dir': 'lib',
                         'doc_dir': 'share/doc',
-                        'with_filesystem': 'yes',
                         'building_shared_supported': 'yes',
                         'install_cmd_data': 'install -m 644',
                         'install_cmd_exec': 'install -m 755'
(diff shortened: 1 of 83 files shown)


randombit:
Of broader interest is 8b3bda4 which adds ISA annotations so that the ISA specific amalgamation files can be compiled without special compiler flags. In fact at least with GCC 6.2 and Clang 3.8 it's possible to compile the amalgamation as a single source file and without any ISA specific flags. (This was already possible under MSVC, since it doesn't require any flags for intrinsics)

PR #666, opened 2016-10-15
Reduce AppVeyor build matrix

randombit:
After #646 AppVeyor builds are taking 7 hours which is just not OK - the continuous bit stops applying when builds come days behind the commits. The problem is the new debug builds, which take 20-30 minutes each to run (slower than running PPC binaries under qemu on Travis!). 2 hours is about the upper bound on how long a CI build I'm willing to tolerate (before #646 AppVeyor took about 1h45m) so I may add back targets if this turns out to be more of a pruning than was required.

Before:
https://ci.appveyor.com/project/randombit/botan/build/1.0.1861
With this commit:
https://ci.appveyor.com/project/randombit/botan/build/1.0.1868

Just enough MSVC 12 to make sure everything still compiles, and just 2 debug builds both under MSVC 14, on the assumption its checks are better and will catch anything MSVC 12 would have caught.

@webmaster128 @neusdan as people actually developing on Windows, does this give enough coverage / cover the right cases? TBH I don't much care which configurations AppVeyor runs, as long as it runs them all in under 2h so feel free to suggest alternatives.

### DIFF
--- a/src/scripts/ci/appveyor.yml
+++ b/src/scripts/ci/appveyor.yml
@@ -4,24 +4,16 @@ platform:
   - x86
   - x86_amd64
 
-configuration:
-  - Release
-  - Debug
-
 environment:
   matrix:
     - COMPILER: msvc-12.0
-      MODE: --enable-shared
-    - COMPILER: msvc-12.0
       MODE: --disable-shared
-    - COMPILER: msvc-12.0
-      MODE: --amalgamation
     - COMPILER: msvc-14.0
       MODE: --enable-shared
     - COMPILER: msvc-14.0
-      MODE: --disable-shared
+      MODE: --enable-shared --enable-debug
     - COMPILER: msvc-14.0
-      MODE: --amalgamation
+      MODE: --disable-shared --amalgamation
 
 install:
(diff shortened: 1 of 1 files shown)


webmaster128:
I totally support the idea and was already a bit concerned about the matrix I did build up by doubling the job count.

Let's try to do something clever with the build time we have. We started with 12 jobs, went to 24 in #646 and have 8 with this PR. So the sweet spot is somewhere at ~10 jobs.

Let's call MSVS 2015 the default compiler, 64 bit the default architecture and --enable-shared the default mode.

1. I'd start with the four basic builds: 32/64bit on MSVS2013/2015
2. Then I'd add static lib and amalgamation
3. Then add a debug build on MSVS2013/2015. Use amalgamation here because it's fast

This is 8 jobs. My experience for number 3 is that the compiler matters more than the architecture because Debug-only fails happened especially with checked iterators and containers.

I would keep the AppVeyor Debug/Release flag because it gives a convenient switch that is displayed separatly and can be used to exclude configurations.

If this is okay, I will open a follow-up PR implementing this.

randombit:
@webmaster128 Your approach sounds good to me

PR #667, opened 2016-10-16
Update Appveyor matrix

webmaster128:
Closes #666

### DIFF
--- a/src/scripts/ci/appveyor.yml
+++ b/src/scripts/ci/appveyor.yml
@@ -1,33 +1,69 @@
 os: Visual Studio 2015
 
-platform:
-  - x86
-  - x86_amd64
-
-configuration:
-  - Release
-  - Debug
+# Let's call MSVS 2015 the default compiler, 64 bit the default architecture,
+# release the default configuration and --enable-shared the default mode.
+#
+# Build jobs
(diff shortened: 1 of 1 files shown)


webmaster128:
Let's see if this works out

neusdan:
Can we add one build with BSI policy enabled? If this fits into the 2 hour time frame...

And i wanted to add the PKCS#11 tests to the CI builds when I find time for this. This will require compiling SoftHSM and some additional time for the test execution. 

Otherwise LGTM, thanks !

randombit:
For the Linux SoftHSM builds, it seems softhsm2 is only in 16.04 so I was planning on having the CI just `wget` a prebuilt softhsm library from somewhere, instead of rebuilding it on each checkin. I'm not sure how what the easiest thing is for Windows is.

randombit:
@neusdan It may be easier to fit the BSI policy into the Travis builds. Already for some where running the tests is very slow (qemu/valgrind) CI is set to build with a minimized policy, but currently always uses `modern`. I will change some of these to use BSI policy instead, I think right now we do not check within CI that BSI policy actually builds/tests at all.

neusdan:
it would be nice to have BSI policy builds on all OSes but Travis is a good start. thanks!

randombit:
1h43m for 588ab9e, lgtm

PR #638, opened 2016-09-29
ossl_arc4.cpp: make RC4 from OpenSSL optional

t0b3:
Help us fade out RC4. We shouldn't support rely on it, it just builds an attack vector due to it's weak stream ciphers.
With this patch accepted, openssl could be installed with ./configure --no-rc4, dropping all RC4 attack surface at once.

PR #668, opened 2016-10-17
Rewrite lookup code to avoid global constructors

randombit:
Needing to deal with this was painfully obvious for a long time; the problems around static initialization were known even in the release that introduced them. And for running TLS on IncludeOS they are a total blocker (resulting in fatal system crashes, when we are initialized before the stack is set up - oops).

In a lot of ways this replacement code feels very primitive: you can't change preferences at runtime, add algos at runtime, etc, etc. But likely no applications ever used these things so the application visible change here is roughly nil, and with fewer complicated templates, fewer macros, no use of global constructors, and no global mutexes when looking up an object (this added contention may help explain why the TLS stack didn't seem to scale much past ~4 cores when I tested on a large system some time ago).


randombit:
I should probably mention the full downsides of this change:

- No longer possible to set algorithm preferences, if provider is "" then we choose our favorite. If OpenSSL is enabled, it is used if available - for everything I've bothered to check OpenSSL is either as fast or faster on most platforms, no surprise considering the amount of effort very skilled people put just at making OpenSSL fast. Similar logic would apply to future providers such as CommonCrypto - just a static preference defined by the order of the checks in the `create` function. This is slightly unfortunate but if applications are picky for some reason they can specify `provider` and in cases where the library calls `create` without allowing specifying a provider, maybe API should be changed to either accept a provider arg or take a pointer/ref to T instead.

- Not possible to add algo implementations from outside the library, but nobody actually copied out the internal-only header they needed to do this.

- No longer possible to add algo implementations at runtime (eg from a shared object). This used to actually work (in 1.10) and is theoretically an interesting feature but really is not a very good idea in practice for a crypto library.

- No longer possible to add an extra alias at runtime, so that `BlockCipher::create("MyAppCipher3")` would work. I'm guessing effectively nobody ever did this.

In retrospect I spent way too much time on this lookup code, and definitely in 1.11.14 second system effected myself in designing an overly general mush. But actually nothing clever or complicated is required here at all. I don't think applications will notice any change from this PR.

An interesting post script. I had started on this when hitting undebuggable system corruptions/crashes on IncludeOS. On realizing the global constructors were seemingly being called before the kernel ran its startup routine (!) I figured there was no hope and started rewriting. It turns out per Alfred Bratterud of IncludeOS that the crashes are due to a limitation in IncludeOS's loader, which assumed that all strings in the binary could fit into 640k. This turns out to be false with Botan included; the workaround is to use `make stripped` and a full fix is expected shortly. But even with the original motivator gone, looking at the results I'm pretty convinced; the extra flexibility `Algo_Registry` (and likewise the `Engine` API from 1.10) offered is just not enough extra functionality to be worth the price of contending on a global lock (much less the complicated code, global constructors, etc).

randombit:
CI is green and I think this is ok to land. Any reviews, comments, feelings that any of the lost functionality might be useful to someone somewhere, ...?

cordney:
For us the limitations are acceptable.

randombit:
I should reference here #465 where some of the problems of current system are discussed and where I laid out more or less what is implemented in this PR (comment from April 18)

cordney:
One thing should be mentioned that seems like a bug to me on master, but is fixed in this branch: If a module is disabled, but an external provider is enabled that provides it, it will be in the library. For example, `./configure.py --disable-modules=sha1` will disable SHA-1 as expected, but `./configure.py --disable-modules=sha1 --with-openssl` will result in SHA-1 being offered by the library through the OpenSSL provider's implementation. One could argue that technically speaking, the module *is* disabled, so it does what the name suggests, but most library users will expect the whole algorithm not being offered by the library, even not by providers. I'm glad this is fixed here.

randombit:
That was proposed for RC4 in #638 and seemed worth following across the board.

PR #669, opened 2016-10-19
Add SHA-3 (and SHAKE-128, and BoringSSL compatible NewHope mode)

randombit:
Really this is just a stopping point on the way to CECPQ1 NewHope TLS ciphersuites, but a useful one. I think all of the remaining changes are in TLS itself.

Adds SHAKE-128 just as a stream cipher for 16, 24, 32, ..., 160 byte keys, because that is both sufficient to handle abstracting SHAKE-128 out of NewHope and enough to be potentially useful in other contexts. It is not a fully satisfactory XOF interface but I'm not really sure what that looks like yet anyway.


randombit:
@cordney I notice the BSI policy allows `keccak` which is not technically approved (at least by NIST, no idea about BSI), should this change to just include `sha3` now?

neusdan:
yeah, please remove keccak in favor of sha3. thanks!

cordney:
Yes, thanks!

Issue #652, opened 2016-10-05
Improving the Doxygen API doc


cordney:
We want to improve on the coverage of the Doxygen API throughout the library going through the it and adding Doxygen API doc for all public functions. However, quite often in derived classes we currently declare functions from the base class interface as `override`, but with no documentation. This sometimes leads to headers having no Doxygen API doc, except for a class description, e.g., https://github.com/randombit/botan/blob/master/src/lib/pubkey/dsa/dsa.h. Our proposal for derived class overridden member functions would be to add a one-line description of the function, presumable taken from the base class function, together with a reference to the base class function:

```cpp
/**
* DSA Public Key
*/
class BOTAN_DLL DSA_PublicKey : public virtual DL_Scheme_PublicKey
   {
   public:
      /**
      * Get the name of the underlying public key scheme.
      * See @ref Public_Key::algo_name()
      */
      std::string algo_name() const override { return "DSA"; }
```

What do you think? Any other ideas for improving the API doc coverage?

randombit:
Sounds great to me.

WRT other Doxygen improvements, there are many warnings right now when Doxygen runs due to undocumented parameters - sometimes due to additional params being added to functions, other times due to renaming of variables.

cordney:
OK, we'll check that, too.

cordney:
Looking at the html generated by doxygen, I see that it (1) automatically takes the description from the base member function for the derived member function and (2) automatically links to the base member function, so the apidoc for `DSA_PublicKey::algo_name()` already looks like this (https://botan.randombit.net/doxygen/classBotan_1_1DSA__PublicKey.html#ae888d86bda27e83e2e2282174ac10405):

```
std::string Botan::DSA_PublicKey::algo_name() const

Get the name of the underlying public key scheme.

Returns
   name of the public key scheme

Implements Botan::Public_Key.
```

Of course you only get that if you look at the doxygen generated output, not by simply browsing the header files in your IDE. Do we still want to duplicate apidoc in derived classes now? My take now would be to keep the status quo and only annotate derived member functions when it adds more (derived-specific) information. Though I must admit that I dislike that an IDE like Eclipse is not able to show me the same information doxygen does (or I simply can not find the right setting).

randombit:
I agree, no reason to duplicate this information in each subclass if Doxygen handles it correctly and there is nothing additional to mention about the derived version.

randombit:
[...]

PR #670, opened 2016-10-19
Improving doxygen coverage

cordney:
This is a follow-up of #652, #661, #662 and #663 and contains improvements in all other modules, except math and util, which I had to drop because of time constraints. It's basically one commit per src folder (block, hash, kdf, ...), with 99d71e5 at the the end fixing small things across multiple src folders and da2e26b fixing the doxygen warnings.

Some notable things:

* doxygen cannot handle forward declarations in parameter lists, such as `func(const class UnknownType& p)`. It generates a warning about definition not matching function declaration, so I changed these 3 occurrences to forward declare at the header scope.
* Even with da2e26b, for me some doxygen warnings remain. These are all for functions in modules that are `#ifdef`ed now, such as `aes_ni`'s `AES_128::aesni_encrypt_n()`, which has a doxygen comment in the definition, but the declaration is not processed by doxygen, because it skips the `#if defined(BOTAN_HAS_AES_NI)` block in `aes.h`. The solution would either be moving the doxygen comment into the header or adding these defines to the `PREDEFINED = ` doxygen variable. Either is fine with me, just let me know and I'll make the change.


cordney:
Also I stumbled upon https://github.com/randombit/botan/blob/master/src/lib/pubkey/pk_keys.h#L174, which is a bug when called with no arguments.

randombit:
Wow, thanks. On the second issue you mentioned, either approach is fine with me. Just setting PREDEFINED seems easier, maybe.

Nice catch on `fingerprint`.

cordney:
Updated the PR accordingly. I additionally had to reorganize some code in anonymous namespaces in d2fd41d, for which doxygen could not match function definitions and declarations.

PR #673, opened 2016-10-21
X25519 key exchange for TLS

randombit:
Internet draft for using x25519 in TLS 1.2 is expired, but this interops with whatever is running on google.com


randombit:
Got a little trigger happy in cleanup up branches ...

PR #674, opened 2016-10-22
Check Doxygen and Sphinx build in CI

randombit:
Adds a Travis `docs` build that runs Doxygen+Sphinx and errors out if there are any problems. With all of @cordney's great work fixing all (!) of the Doxygen warnings it would be sad to slip back as I immediately did in 6aa855bba613c7b.

Doxygen WARN_ON_ERROR option was only added quite recently (https://github.com/doxygen/doxygen/pull/412) so right now the CI build won't actually error out on a warning while building the Doxygen docs. :( We will probably have to wait for Travis to support 16.04 (which has a new enough Doxygen) for that to work.


cordney:
Thanks!

PR #675, opened 2016-10-22
Lucky13 attack countermeasure

jurajsomorovsky:
These changes attempt to implement a countermeasure against the Lucky 13 attack. 

The implemented countermeasure is not equal to the countermeasure presented in the original Lucky 13 paper. After applying the proposed countermeasure from the Lucky 13 paper, we could still see clear measurable timing differences. We assume the reason is the non-constant timing code in hmac.cpp and other hash functions.

The implemented countermeasure attempts to compute an HMAC over a part of the decrypted data. Depending on the padding validity and padding number count, the HMAC is verified over a specific number of bytes so that a computed number of hash compressions are performed. The number of compressions needed to be performed is computed according to the Lucky 13 paper.

The countermeasure is not ideal. To make it ideal, we would need to do something like Adam Langley and make everything constant-time: https://www.imperialviolet.org/2013/02/04/luckythirteen.html 
This is a huge overhead. 

The implemented countermeasure, however, makes the timing difference at least 10x smaller and attacker's life much harder. Here are some figures that present clock cycle differences (between the 40th and 50th percentile) between record decryptions resulting in 4 and 5 compressions:
SHA-256 before:
sha-256-before

SHA-256 after:
sha-256-after

SHA-384 before:
sha-384-before

SHA-384 after:
sha-384-after

The code is available here: https://github.com/Hackmanit/botan-extended-tests

### DIFF
--- a/src/lib/tls/tls_cbc/tls_cbc.cpp
+++ b/src/lib/tls/tls_cbc/tls_cbc.cpp
@@ -298,20 +298,19 @@ size_t TLS_CBC_HMAC_AEAD_Decryption::output_length(size_t) const
 *   - 56+64 - 55+2*64 bytes: 3 compressions
 * 
 * The implemented countermeasure:
-* 1) computes max_comp: number of maximum compressions performed on the 
+* 1) computes max_compressions: number of maximum compressions performed on the 
 *    decrypted data
-* 2) computes current_comp: number of compressions performed on the decrypted
-*    data, without padding
-* 3) if current_comp != max_comp: It computes HMAC over dummy data so that 
-*    max_comp compressions are performed. Otherwise, (max_comp-1).
+* 2) computes current_compressions: number of compressions performed on the 
+*    decrypted data, without padding
+* 3) if current_compressions != max_compressions: It computes HMAC over dummy 
+*    data so that max_comp compressions are performed. Otherwise, (max_comp-1).
 * 
 * Note that the padding validation in Botan is always performed over
 * min(plen,256) bytes, see the function check_tls_padding. This differs
 * from the countermeasure described in the paper.
 * 
 * Note that the padding length padlen does also count the last byte
-* of the decrypted plaintext. This is different from the typical 
-* padding computation and different from the Lucky 13 paper.
+* of the decrypted plaintext. This is different from the Lucky 13 paper.
 * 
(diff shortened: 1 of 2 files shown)


randombit:
Thank you for the patch and excellent writeup.

One question, instead of creating a new MAC object (`dmac`) can the ciphersuite MAC object be used instead? This seems ok, since we are about to fail anyway and the state of the mac object is not needed anymore. Additionally that avoids calling `update` on an unkeyed HMAC object; this is basically undefined behavior in Botan. It happens to do the right thing with the base version in `HMAC` but may not with some other implementation (eg calling OpenSSL), and worse since this function is only ever called in an error situation so may not be tested well or at all, so a bad interaction could go unnoticed for some time, then cause a crash at runtime for some poor user.

randombit:
CircleCI build error is due to test certs having recently expired (just fixed in 6aa855bba61), no problem with your changes.

jurajsomorovsky:
Thank you for the feedback. You are right, re-using the given MAC object would be much cleaner. I decided to initialize a new MAC because during my timing tests I observed better results with a new object. I could not explain that. 
You can see the same SHA-256 result after a patch with the given MAC object here:
sha-256-after-no-new-hmac
As you can see, there is a difference of a few ticks, which is still very small.

However, now I tried to re-analyze the measurements with my dmac object. I initialized dmac with a set_key method and now I get nearly the same results as with the given MAC object. It seems that the missing key initialization made the difference in timing behaviour, and accidentally caused better Lucky 13 countermeasure (by a few ticks).

This means it would be better to use the cipher suite MAC object as you proposed.

jurajsomorovsky:
I executed further tests with the code reusing the MAC and added some dummy data MAC updates to records without compressions. In addition, I re-wrote the patch so that it works (nearly) as proposed in the Lucky 13 paper. The code is now much simpler and easier to understand. I observed timing differences of at most 120 clock cycles (in one SHA-384 measurement), which is still fine.
(for the record, note that my measurements were executed just by invoking the decrypt function, no network was involved)

Final results:
SHA-1:
sha-1-final-valid-invalid

SHA-256:
sha-256-final-valid-invalid

SHA-384:
sha-384-final-valid-invali

Thank you for your feedback on the HMAC initialization and key scheduling. I think this was not clean and could have caused some troubles in specific cases or environments.

PR #679, opened 2016-10-23
Reunify registered test names

cordney:
This change may be a bit esotheric, but with this year's many additions tests were registered with different naming schemes, e.g., `ecies-iso` vs. `ecgdsa_keygen`. There seems to be a naming scheme already in Botan, and I'd like to see things clarified (and preferably added to the docs). From what I understand:

* use `_` instead of `-` to separate words (fixed in this PR)
* a unit test, that is, a test that doesn't require external resources such as sockets, filesystem (that includes .vec test vectors), etc., goes in a `unit_xxx.cpp` file, all other tests go in a `test_xxx.cpp` file
* a unit test registers as `xxx_unit`, all other tests register as `xxx[_yyy_zzz...]`
* some known answer tests currently register as `xxx_kat`, such as `dh_kat` and `dsa_kat`, others simply as `xxx`, such as `ecies` and `ecdsa`; what's our choice here?

### DIFF
--- a/src/tests/test_dlies.cpp
+++ b/src/tests/test_dlies.cpp
@@ -191,7 +191,7 @@ class DLIES_Unit_Tests : public Test
          }
    };
 
-BOTAN_REGISTER_TEST("dlies-unit", DLIES_Unit_Tests);
+BOTAN_REGISTER_TEST("dlies_unit", DLIES_Unit_Tests);
 
 #endif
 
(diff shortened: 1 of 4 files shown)


randombit:
I don't think the system of test naming and especially unit test vs other tests is so well thought out. Especially unit_tls is a terrible name considering it is really an integration test. But anyway, your changes here seem a definite improvement. Probably all known answer tests should be named `X_kat` for consistency.

PR #684, opened 2016-10-25
Add PKCS #11 and TPM to coverage build

randombit:
Should be a nice bump in the coverage report, and TPM was not being built at all before.

It would be nicer to have a proper apt repository with SoftHSMv2 built for Trusty, but a tarball of something I compiled on a VM is good enough for now.


neusdan:
wow great, thanks!

PR #686, opened 2016-10-25
Try to correct behavior of EMSA1::verify

neusdan:
I've checked again if we can close now #550.

Nyberg-Rueppel was removed, so this is no longer a problem. But the RSA tests still fail.

The problem with RSA is that `RSA_Verify_Operation::verify_mr` does not return leading zeros. But i think this can't be fixed here. The number of possible leading zeros depends on the encoding method and the hash function used.

I need some help here. I've found a solution where all tests pass but it doesn't look right to me.
Maybe the best is if the EMSA class would have a method that returns the output bytes. This could then be used to calculate the needed padding bytes.

### DIFF
--- a/src/lib/pk_pad/emsa1/emsa1.cpp
+++ b/src/lib/pk_pad/emsa1/emsa1.cpp
@@ -67,29 +67,11 @@ secure_vector<byte> EMSA1::encoding_of(const secure_vector<byte>& msg,
 bool EMSA1::verify(const secure_vector<byte>& coded,
                    const secure_vector<byte>& raw, size_t key_bits)
    {
-   try {
-      if(raw.size() != m_hash->output_length())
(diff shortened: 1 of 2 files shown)


randombit:
Yeah something feels not quite right with your patch, though your change to `EMSA1::verify` is certainly a step in the right direction. I'll try to look at this soon so we can get it nailed down before RC1.

randombit:
Looked at this again, definitely not right - the extra complication in the high level PK signature code to handle what is actually an interaction between only RSA and EMSA1 is bad news. The fix should be in either RSA (or EMSA1 (or both)), but not by having PK_Verifier mediate their disagreement by translating between them.

At this point RSA is the only signature scheme with message recovery still in the library, so more or less we can do whatever we want quite easily. Eg, if necessary change `verify_mr` signature to return some additional necessary information.

Part of the problem is the verify code is overly complicated to begin with, #705 for that.

PR #687, opened 2016-10-26
Fixes randombit/botan#677

### LINKED ISSUE #677: Botan fails to build when configured in "maintainer mode".
*Steps to reproduce*:
1. Clone Botan repository
2. In the repo base directory run: `./configure.py --maintainer-mode`
3. Compile the code using `make`
4. Observe the following errors:
```
./src/tests/test_rng.cpp: In member function ‘virtual Botan_Tests::Test::Result Botan_Tests::{anonymous}::X931_RNG_Tests::run_one_test(const string&, const VarMap&)’:
./src/tests/test_rng.cpp:55:74: warning: ‘Botan::ANSI_X931_RNG::ANSI_X931_RNG(Botan::BlockCipher*, Botan::RandomNumberGenerator*)’ is deprecated: X9.31 RNG is deprecated and will be removed soon [-Wdeprecated-declarations]
          Botan::ANSI_X931_RNG rng(bc.release(), new Fixed_Output_RNG(ikm));
                                                                          ^
In file included from ./src/tests/test_rng.cpp:16:0:
build/include/botan/x931_rng.h:39:7: note: declared here
       ANSI_X931_RNG(BlockCipher* cipher,
       ^
./src/tests/test_rng.cpp: In member function ‘Botan_Tests::Test::Result Botan_Tests::{anonymous}::HMAC_DRBG_Unit_Tests::test_fork_safety()’:
./src/tests/test_rng.cpp:413:47: error: ignoring return value of ‘ssize_t read(int, void*, size_t)’, declared with attribute warn_unused_result [-Werror=unused-result]
             read(fd[0], &count, sizeof(count));
                                               ^
./src/tests/test_rng.cpp:421:61: error: ignoring return value of ‘ssize_t read(int, void*, size_t)’, declared with attribute warn_unused_result [-Werror=unused-result]
             read(fd[0], &child_bytes[0], child_bytes.size());
                                                             ^
./src/tests/test_rng.cpp:434:48: error: ignoring return value of ‘ssize_t write(int, const void*, size_t)’, declared with attribute warn_unused_result [-Werror=unused-result]
             write(fd[1], &count, sizeof(count));
                                                ^
./src/tests/test_rng.cpp:436:62: error: ignoring return value of ‘ssize_t write(int, const void*, size_t)’, declared with attribute warn_unused_result [-Werror=unused-result]
             write(fd[1], &child_bytes[0], child_bytes.size());
                                                              ^
In file included from ./src/tests/test_rng.cpp:9:0:
./src/tests/test_rng.h: In member function ‘virtual void Botan_Tests::Fixed_Output_RNG::add_entropy(const uint8_t*, size_t)’:
./src/tests/test_rng.h:49:12: warning: assuming signed overflow does not occur when changing X +- C1 cmp C2 to X cmp C2 -+ C1 [-Wstrict-overflow]
       void add_entropy(const uint8_t b[], size_t s) override
            ^
./src/tests/test_rng.h:49:12: warning: assuming signed overflow does not occur when changing X +- C1 cmp C2 to X cmp C2 -+ C1 [-Wstrict-overflow]
```

mgierlings:
gcc was still not happy but this fixes #677 on my machine.

### DIFF
--- a/src/tests/test_rng.cpp
+++ b/src/tests/test_rng.cpp
@@ -447,7 +447,7 @@ class HMAC_DRBG_Unit_Tests : public Test
             ::close(fd[0]); // close read end in child
             rng.randomize(&child_bytes[0], child_bytes.size());
             count = counting_rng.randomize_count();
-            ::write(fd[1], &count, sizeof(count));
+            ssize_t written = ::write(fd[1], &count, sizeof(count));
             try {
                rng.randomize(&child_bytes[0], child_bytes.size());
             }
@@ -455,7 +455,8 @@ class HMAC_DRBG_Unit_Tests : public Test
                {
                fprintf(stderr, "%s", e.what());
                }
-            ::write(fd[1], &child_bytes[0], child_bytes.size());
+            written = ::write(fd[1], &child_bytes[0], child_bytes.size());
+            BOTAN_UNUSED(written);
             ::close(fd[1]); // close write end in child
             ::_exit(0);
             }


webmaster128:
Does this also work without an explicit variable?

`BOTAN_UNUSED(::write(fd[1], &child_bytes[0], child_bytes.size()));`

mgierlings:
@webmaster128 Unfortunately it does not work. Also just assing the return value to the variable "written" will not work. I had to use the combination of both, assigning the return value to a variable and applying the BOTAN_UNUSED macro to that variable, to make the warning disappear.

webmaster128:
Okay. Could that be wrapped in a new BOTAN_DISCARD_RESULT macro, which does something like the following?

```
{
    auto result = expression
    BOTAN_UNUSED(result);
}

mgierlings:
@webmaster128, @randombit  Yes that should work, however we might want to consider to change BOTAN_UNUSED to utilize [this](http://git.savannah.gnu.org/cgit/gnulib.git/tree/lib/ignore-value.h) little trick instead.

randombit:
@mgierlings Actually the two proposals seem to have the same effect, just one using C++11 auto the other GCC's `__typeof__` extension. But given the option I would prefer `BOTAN_UNUSED` argument always be just a variable name rather than an expression, so we can paste the variable name into a `_Pragma(unused)`, `[[gnu::unused]]` C++11 attribute, or `__attribute__((unused))` - http://stackoverflow.com/questions/31909631/c11-style-unused-attribute-in-gcc has some examples.

PR #688, opened 2016-10-27
Fix TLS resumption bugs

randombit:
Surprise surprise, try writing some tests to add coverage and find some bugs (details in the commit). No security issue I can see, just resumption failures that shouldn't happen.


randombit:
It would be better if the session manager interface allowed specifying which version we'd like to use, so it could manage multiple session versions correctly (especially TLS vs DTLS). Right now if a client has a shared session cache and opens both TLS and DTLS handshakes for a single host/service, the sessions will stomp on each other in the cache in unfortunate ways. Not sure I want to take on another API change here at this point, though - maybe better to just mention that issue in the session manager docs and move on.

PR #689, opened 2016-10-27
Add more tests for Pipe/Filter

randombit:
Turns out `Threaded_Fork` is completely broken - not really sure the best course of action with this.


neusdan:
Not related to the tests, but you planned to remove the pipe and filters API, right?

randombit:
Not for 2.0 anyway - we're at least a year too late to deprecate and remove such a major API. I think the status for Pipe in 2.0 is effectively 'supported, but a dead end discouraged for new use and with no further improvements expected' (the docs need to be updated to make this clear)

PR #690, opened 2016-10-28
HMAC_DRBG: Make max_number_of_bytes_per_request configurable

### LINKED ISSUE #672: HMAC_DRBG: configuration of max_number_of_bytes_per_request
Default the Botan `HMAC_DRBG` uses 64KB for `max_number_of_bytes_per_request`. I need to use a custom value here. 

So it would be nice if we can make this configurable by adding a new constructor parameter or a new value in `build.h`. Not sure which is the best. 

For a test run i already changed the hardcoded value to 64 Bytes and got a lot of test errors. Which i think is no surprise because this low value enforces a lot of reseeds which the test vectors are not expecting.

neusdan:
Closes #672

Before this change the `max_number_of_bytes_per_request` were hardcoded to 64 * 1024 Bytes. Now you can pass the value as an argument to the constructor.

@cordney and me think that it makes only sense to set a custom `max_number_of_bytes_per_request` in combination with a custom `reseed_interval` (especially `reseed_interval = 1`). So the parameter is only added to constructors where the `reseed_interval` can be set.

### DIFF
--- a/src/lib/rng/hmac_drbg/hmac_drbg.cpp
+++ b/src/lib/rng/hmac_drbg/hmac_drbg.cpp
@@ -12,38 +12,63 @@ namespace Botan {
 
 HMAC_DRBG::HMAC_DRBG(std::unique_ptr<MessageAuthenticationCode> prf,
                      RandomNumberGenerator& underlying_rng,
-                     size_t reseed_interval) :
+                     size_t reseed_interval,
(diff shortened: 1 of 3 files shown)


randombit:
LGTM, thanks. I think you are right, if there is no actual source to reseed from then there does not seem to be any useful reason to set max_output to below the default.

PR #698, opened 2016-10-30
Use OPENSSL_cleanse for zero_mem if available.

randombit:
This is a small change but I'd like to get some comment on it. OpenSSL has asm specific versions of `OPENSSL_cleanse` for most platforms, which are guaranteed to do the correct thing without interference from those pesky optimizers. On platforms without asm, OpenSSL uses memset through volatile pointer just as we do.

This does introduce another dependency point on OpenSSL, and one that is hard to test meaningfully. But I'm sure `OPENSSL_cleanse` has more eyes on it than whatever we are doing.

### DIFF
--- a/src/lib/utils/mem_ops.h
+++ b/src/lib/utils/mem_ops.h
@@ -15,14 +15,21 @@
 namespace Botan {
 
 /**
-* Zeroize memory
+* Zeroize memory in a way that a compiler should not remove
+* Use this to zero memory before deallocating it.
+*
 * @param ptr a pointer to memory to zero out
 * @param n the number of bytes pointed to by ptr
 */
 BOTAN_DLL void zero_mem(void* ptr, size_t n);
 
 /**
-* Zeroize memory
+* Zero memory before use. This simply calls memset and should not be
+* used in cases where the compiler cannot see the call as a
+* side-effecting operation (for example, if calling clear_mem before
+* deallocating memory, the compiler would be allowed to omit the call
+* to memset entirely under the as-if rule.)
+*
 * @param ptr a pointer to an array
 * @param n the number of Ts pointed to by ptr
 */
--- a/src/lib/utils/zero_mem.cpp
+++ b/src/lib/utils/zero_mem.cpp
@@ -1,4 +1,4 @@
- /*
+/*
 * Zero Memory
 * (C) 2012,2015 Jack Lloyd
 *
@@ -9,6 +9,8 @@
 
 #if defined(BOTAN_TARGET_OS_HAS_RTLSECUREZEROMEMORY)
   #include <windows.h>
+#elif defined(BOTAN_HAS_OPENSSL)
+  #include <openssl/crypto.h>
 #endif
 
 namespace Botan {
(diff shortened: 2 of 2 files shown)


randombit:
I looked at this more, `OPENSSL_cleanse` is actually not what we want since in many versions it overwrites the memory with (effectively) random values which is certainly not what a caller to a function named `zero_mem` will have expected. And pairing it with a memset to zero the memory seems to introduce the same potential issues all over again.

Dropping this change, I'm just going to update the Doxygen comments on master

randombit:
And on a few minutes further reflection, no reason to not rename `zero_mem` to `secure_scrub_memory` to be more explicit about its actual intended usage.

neusdan:
Is it somehow proven that OpenSSLs technique is safer than Botans? It's probably not that easy but i would rather learn what we are missing here and then implement this our self :-)

randombit:
Assuming the compiler is 'nice' and respects our volatile pointer (to either data bytes or a function pointer, depending on the ifdef) everything works great. What we do is exactly what OpenSSL 1.1.0 does (when OpenSSL has no asm support). But in theory a sufficiently nasty (ie, well optimizing) compiler could still optimize out the code in undesirable ways: it seems within the bounds of the C/C++ rules that the compiler could dereference the volatile pointer at runtime, check to see if it still pointed at memset, and if so once again decide to skip the memset (since the memset in fact has no visible side effects, we just use the volatile pointer to try to force the compiler to evaluate it). This was argued in http://www.daemonology.net/blog/2014-09-05-erratum.html and it makes sense to me. I am not feeling motivated right now to dig through the C99 and C++11 specs to see exactly what they say on the topic, and anyway it seems `volatile` semantics are very much compiler specific and in GCC seem primarily driven by supporting hardware drivers.

Is any current compiler optimizer that good? No. But that is a difficult game to play, especially when we have no effective way to test the resulting binary to assure that no memsets were elided. (Though see this Crypto++ bug which was found with an instrumented Clang https://github.com/weidai11/cryptopp/issues/331). Whereas, hand written asm to clear the buffer is assured to work, now and in the future. But, this is an additional maintenance hassle, especially with MSVC not supporting inline asm on x64, lots of platforms to test, etc. But the OpenSSL developers have already written (and hopefully tested...) the code for us for most major platforms, all we have to do is call it. Thus my motivation on this patch.

I was recommended to look into the technique of OpenSSL 1.0.2 by Falko Strenzke (who has contributed major pieces to the library in the past including McEliece and the original ECC implementation). It uses a kind of strange RNG which the compiler is thought to be unable to optimize out effectively (http://stackoverflow.com/questions/26433772/why-does-openssl-cleanse-look-so-complex-and-thread-unsafe). But, it also seems to introduce data races which while harmless may trigger under dynamic analyzers like valgrind's helgrind checker. I really hate how OpenSSL causes valgrind false positives within applications calling OpenSSL, and would prefer to avoid inflicting that on our users.

The actual ultimate fix here is to get a `__builtin_scrub_memory` added to Clang and GCC. Probably neither would be a difficult patch to write though I have no idea if upstream would accept them. It does seem like a widely acknowledged issue with current compilers. Or for the ultimate challenge, fixing it in the ISO standard, even if just by some special case rule explicitly requiring that a volatile function pointer initialized with memset be always handled as an unconditional jump.

For now on master I've renamed `zero_mem` to `secure_scrub_memory`, it still zeros out but is documented as maybe not doing that all of the time to leave us an opening for `OPENSSL_cleanse` or some RNG-like technique in the future.

neusdan:
Thanks for taking the time to write this detailed explanation. 

> especially with MSVC not supporting inline asm on x64

I think we can ignore Windows here because as far as i know it's guaranteed that the compiler does not optimze away `RtlSecureZeroMemory`. 

For the other OS i have found a summary of techniques here:

https://programfan.github.io/contents/2015/04/27/prevent-gcc-optimize-away-code/

Maybe we can use some / all of them additionally.

randombit:
> I think we can ignore Windows here because as far as i know it's guaranteed that the compiler does not optimze away RtlSecureZeroMemory.

That's a good point and excellent news since we can use GCC-style inline asm with effectively every other major compiler, which simplifies that side of the problem at least.

Thanks for the link, the volatile empty asm barrier may do what we want and without having to actually write CPU specific asm for all the loops.

PR #709, opened 2016-11-03
Fix DSA parameter generation to use the correct loop bound.

randombit:
I should probably reread FIPS 186-3 before I merge this; is this the only change I missed between 186-2 and 186-3?

Removes 512 and 768 bit DSA param generation. I really hope nobody was using that.

There are no tests for DSA param generation right now, it would be nice to fix that to ensure we are compatible with other implementations.

### DIFF
--- a/src/lib/math/numbertheory/dsa_gen.cpp
+++ b/src/lib/math/numbertheory/dsa_gen.cpp
@@ -20,7 +20,7 @@ namespace {
 bool fips186_3_valid_size(size_t pbits, size_t qbits)
    {
    if(qbits == 160)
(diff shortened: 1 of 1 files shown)


randombit:
The good news, when I added DSA parameter verification tests from FIPS 186-3 they all passed. The bad news, they took almost 8 minutes to finish. So it is limited to just 1024 bit params by default (~3 seconds), even just adding 224,2048 params bumped it to almost a minute.

randombit:
I tried to diff FIPS 186-2 Appendix 2 and 186-3 A.1.1.2, hard as the notation changed quite a bit. Didn't spot any differences besides this issue and the obvious extension to larger hashes.

PR #710, opened 2016-11-03
Add one in-house EC curve for TLS at compile-time

cordney:
One additional, application-specific curve can be added at compile time, using the new
`configure.py --house-curve=curve.pem,funky311,1.2.3.4,FEFF`.

I couldn't find any group in configure.py and build.h that fit, so I created a new group "misc".

After #706 was merged, I added an ifdef'd OID to `oids.cpp`. Since the file is auto-generated now, oids.py needs to be updated as well, but I'd like to hear your opinion first. Excuse my poor python skills.


randombit:
Works for me, just update oids.py script and regenerate oids.cpp

Leaving a reference to #636 here so I remember to list both issue numbers in the release notes.

cordney:
Updated

cordney:
Thanks

PR #716, opened 2016-11-05
Implement barriers for Threaded Fork

### LINKED ISSUE #695: Threaded_Fork does not behave correctly
Threaded_Fork starts a pool of work threads (one per filter) which have the following main loop:

```
   while(true)
      {
      /*
      * This is plain wrong: a single thread can get the semaphore
      * more than one time, meaning it will process the input twice
      * and some other thread/filter will not see this input.
      */
      m_thread_data->m_input_ready_semaphore.acquire();

      if(!m_thread_data->m_input)
         break;

      filter->write(m_thread_data->m_input, m_thread_data->m_input_length);
      m_thread_data->m_input_complete_semaphore.release();
      }
```

A semaphore does not guarantee that each of N thread will acquire it exactly once, so for example if you had a `Threaded_Fork` that had two hash functions, and you sent the string 'A' to the pipe, it's possible one filter would see 'AA' and other other nothing at all, causing incorrect computations in both. Of course sometimes the semaphore acquisition happens to behave as the code expects and everything is fine - the odds decrease as the number of filters (and thus competing threads) increases.

There is a disabled test in `test_filters.cpp` that fails with some regularity on a multicore machine.

Opening an issue for this in the hopes someone else wants to take this on. If not I'll probably disable or remove Threaded_Fork before 2.0 so as not to ship something silently broken. I think the fix is just to replace the semaphore with a barrier? (http://stackoverflow.com/questions/24465533/implementing-boostbarrier-in-c11)

lowjoel:
This commit introduces a concept of a barrier, where all threads must
synchronise before continuing. Threaded Fork uses this to ensure that all
input is consumed by each sink exactly once.

Fixes #695.

### DIFF
--- a/src/lib/filters/threaded_fork.cpp
+++ b/src/lib/filters/threaded_fork.cpp
@@ -11,6 +11,7 @@
 #if defined(BOTAN_TARGET_OS_HAS_THREADS)
 
 #include <botan/internal/semaphore.h>
+#include <botan/internal/barrier.h>
 
 namespace Botan {
 
@@ -23,14 +24,14 @@ struct Threaded_Fork_Data
    Semaphore m_input_ready_semaphore;
 
    /*
-   * Ensures that all threads have completed processing data.
+   * Synchronises all threads to complete processing data in lock-step.
    */
-   Semaphore m_input_complete_semaphore;
+   Barrier m_input_complete_barrier;
 
    /*
    * The work that needs to be done. This should be only when the threads
    * are NOT running (i.e. before notifying the work condition, after
-   * the input_complete_semaphore is completely reset.)
+   * the input_complete_barrier has reset.)
    */
    const byte* m_input = nullptr;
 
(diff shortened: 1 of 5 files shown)


lowjoel:
Oh, the threaded fork test was broken, let me enable it.

lowjoel:
```
SecureQueue ran 6 tests all ok
Threaded_Fork ran 4 tests all ok
FPE_FE1 ran 4 tests in 25.11 msec all ok
```

randombit:
Hi Joel, thanks for looking at this.

You can run the filter tests alone with `botan-test filter`. I ran the test in a loop 40000 times on my 4-core w/HT desktop, and with your change the original test now passes fine each time. Then, I confidently wrote (in b03b5c3f23, on master now) another, more complicated test for `Threaded_Fork` that used several filters and multiple writes operations. With the old version, the results it gives are incorrect. With your barrier approach, this new test promptly deadlocks in almost all runs. :( I'm not really sure why that is happening.

lowjoel:
Let me have a look at it later today.

lowjoel:
Sorry, accidentally deleted the branch 😓 let me open another PR.

lowjoel:
I got it fixed -- calling sync() and wait() in a tight loop caused some of the threads to sleep across syncs. I've modified the barrier so that the previously sleeping threads can awaken to ensure progress.

This is similar to the generational or a generalised version of the up/down counting barrier in the Stack Overflow thread.

randombit:
Merged to master, thank you!

lowjoel:
Glad to be of help :)

PR #717, opened 2016-11-06
Added Extended Hash-Based Signatures (XMSS)

mgierlings:
[1] XMSS: Extended Hash-Based Signatures,
    draft-itrf-cfrg-xmss-hash-based-signatures-06
    Release: July 2016.
    https://datatracker.ietf.org/doc/
    draft-irtf-cfrg-xmss-hash-based-signatures/?include_text=1

Provides XMSS_PublicKey and XMSS_PrivateKey classes as well as implementations
for the Botan interfaces PK_Ops::Signature and PK_Ops::Verification. XMSS has
been integrated into the Botan test bench, signature generation and verification
can be tested independently by invoking "botan-test xmss_sign" and
"botan-test xmss_verify"

- Some headers that are not required to be exposed to users of the library have
  to be declared as public in `info.txt`. Declaring those headers private will
  cause the amalgamation build to fail. The following headers have been
  declared public inside `info.txt`, even though they are only intended for
  internal use:
    * atomic.h
    * xmss_hash.h
    * xmss_index_registry.h
    * xmss_address.h
    * xmss_common_ops.h
    * xmss_tools.h
    * xmss_wots_parameters.h
    * xmss_wots_privatekey.h
    * xmss_wots_publickey.h

- XMSS_Verification_Operation Requires the "randomness" parameter out of the
  XMSS signature. "Randomness" is part of the prefix that is hashed *before*
  the message. Since the signature is unknown till sign() is called, all
  message content has to be buffered. For large messages this can be
  inconvenient or impossible.

  **Possible solution**: Change PK_Ops::Verification interface to take
  the signature as constructor argument, and provide a setter method to be able
  to update reuse the instance on multiple signatures. Make sign a parameterless
  member call. This solution requires interface changes in botan.

  **Suggested workaround** for signing large messages is to not sign the message
  itself, but to precompute the message hash manually using Botan::HashFunctio
  and sign the message hash instead of the message itself.

- Some of the available test vectors for the XMSS signature verification have
  been commented out in order to reduce testbench runtime.


randombit:
It's going to take me a while to work through this fully but one quick comment on the OID: there should not be direct edits to this file, instead values are set in `src/build-data/oids.txt` and then the `oids.cpp` file regenerated. Also, IANA has assigned me an arc 1.3.6.1.4.1.25258 to which I have assigned 1.3.6.1.4.1.25258.1 arc for public key algorithms that don't have some OID already assigned (McEliece, etc). The next unassigned OID in that arc is 1.3.6.1.4.1.25258.1.5, I just assigned that value to XMSS in master. This can be used until the IETF or whoever defines an official format and OID. You should drop the all-zeros OID and rebase.

mgierlings:
Okay thanks you, OIDs should be fixed now.

mgierlings:
@randombit Any idea why that [FFI test error](https://travis-ci.org/randombit/botan/jobs/173760724#L1455) occurs on arm? I'm afraid I can't reproduce this on my platform.

randombit:
I think it has something to do with very large timing variability introduced by running CPU hog qemu on CI VMs. It seems to show up most often under qemu but also sometimes AppVeyor or valgrind builds. I've never reproduced it on my machines.

The problem is really due to a bug in the FFI API, in short the length of the output of a particular function always depends on how many PBKDF iterations occur in 300 ms, but this varies between the two times we call it (once to get the length, the other to receive the value). In fact there is already code to attempt to compensate for this, but it apparently is not sufficient, and when the lengths are too far apart the test fails. (`ffi_tst_pubkey_export` in `test_ffi.cpp` has the details). The problem is it is kind of a bogus API as well as a bogus test - the API should be able to accept a fixed iteration count, and it should be possible to get at least an estimate of the size of the encoded private key without actually encoding it and throwing away the result (really, that's what the test is doing now). And the test should be capable of incrementally resizing the buffer to receive the output, if it must.

src/lib/pubkey/xmss/xmss_hash.cpp
```diff
@@ -0,0 +1,88 @@
...
+                 const secure_vector<byte>& root,
+                 const secure_vector<byte>& index_bytes,
+                 const secure_vector<byte>& data)
+   {
+   h_msg_init(randomness, root, index_bytes);
+   m_msg_hash->update(data);
+   return m_msg_hash->final();
+   }
+
+XMSS_Hash& XMSS_Hash::operator=(XMSS_Hash hash)
```
randombit: According to codecov this function is not called anywhere during the tests - is it actually needed/useful? If not it could instead just be declared `=delete`

src/lib/pubkey/xmss/xmss_index_registry.h
```diff
@@ -0,0 +1,109 @@
...
+       *              registered key.
+       *
+       * @return positon of leaf index registry entry for key identified
+       *         by id.
+       **/
+      size_t add(uint64_t id, size_t last_unused = 0);
+
+      std::vector<uint64_t> m_key_ids;
+      std::vector<std::shared_ptr<Atomic<size_t>>> m_leaf_indices;
+      std::mutex m_mutex;
```
randombit: Use `mutex_type` from `<botan/mutex.h>` so this can work on single-threaded platforms.

src/lib/pubkey/xmss/xmss_wots_publickey.h
```diff
@@ -0,0 +1,337 @@
...
+         }
+
+      virtual AlgorithmIdentifier algorithm_identifier() const override
+         {
+         BOTAN_ASSERT(false, "No AlgorithmIdentifier available for XMSS-WOTS.");
+         }
+
+      virtual bool check_key(RandomNumberGenerator&, bool) const override
+         {
+         BOTAN_ASSERT(false, "No key strength check implemented for XMSS-WOTS.");
```
randombit: These should be `throw Not_Implemented(...)` - it makes such cases easier to find with a grep.

src/lib/pubkey/xmss/xmss_tools.h
```diff
@@ -0,0 +1,111 @@
...
+#include <botan/types.h>
+#include <botan/secmem.h>
+
+namespace Botan {
+
+/**
+ * Helper tools for low level byte operations required
+ * for the XMSS implementation.
+ **/
+ class XMSS_Tools
```
randombit: This seems like a lot of extra syntax just to save `m_is_little_endian` - but there was no runtime endian check before in utils. I've added `CPUID::is_little_endian()` to master in 523b2a4ca, cpuid.h header is public. It seems with that call `XMSS_Tools` class/singleton could be removed and the `concat`s do not need to be member functions anymore, also simplifying syntax for callers.

randombit:
I'm reviewing this PR first, then WOTS can be rebased and reviewed after. Still a lot to read through, taking a break for now however. One question I had, re SHAKE support. With how XMSS uses it, is the desired SHAKE output length always known before the input is consumed? Ideally, XMSS only needs a few specific and reasonably small output sizes and we can define new hash types `SHAKE-128(output_len)` and such without having to get into generic XOF interfaces right now.

It would be nice to support both SHA-2 and SHAKE in XMSS from the start and ensure that everything is well tested.

src/lib/pubkey/xmss/xmss_publickey.h
```diff
@@ -0,0 +1,273 @@
...
+         }
+
+      virtual size_t message_part_size() const override
+         {
+         return std::numeric_limits<size_t>::infinity();
+         }
+
+      virtual size_t message_parts() const override
+         {
+         return std::numeric_limits<size_t>::infinity();
```
randombit: ?

src/lib/pubkey/xmss/xmss_index_registry.h
```diff
@@ -0,0 +1,109 @@
...
+#include <botan/secmem.h>
+#include <botan/types.h>
+#include <botan/atomic.h>
+
+namespace Botan {
+
+/**
+ * A registry for XMSS private keys, keeps track of the leaf index for
+ * independend copies of the same key.
+ **/
```
randombit: Is this scheme described in the i-d? From a brief check I could not find it.

Also a general comment, as an aid to reviewers any comments in the source just briefly referring to the relevant sections of the i-d are very helpful.

src/lib/pubkey/xmss/xmss_publickey.h
randombit: Why would this function return something other than 1? Larger values are really only used for DSA/ECDSA pair encoding.

src/lib/pubkey/xmss/xmss_hash.cpp
mgierlings: Yes this is not required any more - I'll delete it.

src/lib/pubkey/xmss/xmss_index_registry.h
mgierlings: Changed the mutex type and lock_guard accordingly.

src/lib/pubkey/xmss/xmss_tools.h
mgierlings: Great, thanks for adding the function, I'll simplify the XMSS_Tools class and remove the singleton.

src/lib/pubkey/xmss/xmss_publickey.h
mgierlings: I think I might be missinterpreting the meaning of this function. I thought this was supposed to specify the number of parts a message may consist of, when signing with the signature scheme, or simpy put how many times `update(message_part[], part_size)` may be called, with message_part_size() limiting the length of the `message_part[]` array for a single call. Since `XMSS_Signature_Operation::update()` may be called arbitrarily often, with parameters of arbitrary size, I figured `infinity()`would be appropriate here. Unfortunately the documentation in the base class does not provide much information either. Could you clarify how the methods `message_parts()`, `message_part_size()`, `max_input_bits()` are expected to act?

src/lib/pubkey/xmss/xmss_index_registry.h
mgierlings: Actually this is not defined in the standard. The standard only states that reuse of one time keys must be prevented. Since the XMSS_Private_Key is copyable, there is a chance that two independent copies of the same private key exist within one application. The leaf indices might be updated independently which would likely result in key reuse. The purpose of the index registry is to map all copies of a private key to a shared leaf index. That way even if a library user accidentally copies the same key and uses multiple instances for signing, the leaf index will still be properly updated across all instances of the same key.

src/lib/pubkey/xmss/xmss_publickey.h
randombit: `max_input_bits` is only an implementation function used by `PK_Ops::{Signature,Verification}_with_EMSA` and never called through the PK_Ops::Signature API. I will move the declaration to the sub-class. XMSS is correct to avoid these subclasses since XMSS does message hashing in an unusual way and does not rely on any external padding scheme.

As to the other functions, these have nothing to do with limitations on `update` (no such limits exist beyond what fits in size_t), instead the (confusingly named) `message_part_size` and `message_parts` should only be overridden if the output signature is a set of BigInts and you want to DER encode them (parts=2, size=order bits). This DER encoding is only used for DSA and ECDSA signatures in some protocols. XMSS should simply not define the functions and assume the default.

randombit:
The buffering problem with verification would also affect EdDSA since there also the message is prefixed with a random value included as part of the signature. I will work on a change to the verify code so buffering is not required in XMSS.

mgierlings:
> One question I had, re SHAKE support. With how XMSS uses it, is the desired SHAKE output length always known before the input is consumed? Ideally, XMSS only needs a few specific and reasonably small output sizes and we can define new hash types SHAKE-128(output_len) and such without having to get into generic XOF interfaces right now.

SHAKE's extensible output capability is not required for use in XMSS, since the [standard proposes] (https://tools.ietf.org/html/draft-irtf-cfrg-xmss-hash-based-signatures-06#page-40) fixed element sizes of length "n" .

That being said I think it might be possible to realize the XOF capabilities with very little changes to the existing Botan interfaces. The only change that would be required IMO is to allow multiple calls to `final()` without clearing the internal state. This way the existing hash output will be extended by another block every time `final()` is called.

[This `clear()`](https://github.com/randombit/botan/blob/master/src/lib/hash/sha3/sha3.cpp#L199) would be replaced by a call to the squeeze operation.

While Merkle-Damgard based hash functions supported currently by Botan work like this:

    x[i] = input blocks
    y[i] = output blocks

    update(x[0]) -> update(x[1]) -> ... -> update(x[n]) -> final(y[0])

The sponge construction used for SHAKE could be realized like this:

    update(x[0]) -> update(x[1]) -> ... -> update(x[n])  -> final(y[0]) -> final(y[1]) -> ... -> final(y[m])
(`m`, `n` can be chosen independently, but *all* calls to update must be performed before the first call to final.)

Every call to final yields a new output block of fixed size. If a hash output length `l`, that is not a multiple of the output block size, is required, hash output `Y = y[0] || y[1] || ... || y[m]` is generated such
that `len(Y) > l`. Then `Y` is truncated to length `l`.

Internally the SHAKE implementation would also need to keep track of whether the transition from absorption to squeeze phase has been made. This could be a simple boolean member `m_squeeze` that is set to true when `final_result()` is called. Calling update during the squeeze phase again would then either trigger an assertion or throw an exception.

For convenience Botan could feature an overload

`final(byte out[], size_t len)`

that generates `ceil(len/block_size)` output blocks and truncates the result to `len`. For non XOF hash functions `len` could be capped to output block size and/or throw an error if `len` is too big. Alternatively only multiples of output block size could be supported leaving the truncation to the user. In this case the additional XOF-related overload is unnecessary.

One might argue then, that the method name `final()` is misleading once it is legal to be called multiple times, but then again final is still correct in the sense that after the first call to final no additional input may be provided via update. So `final()` finalizes the input, not the output for hash functions with XOF capabilities.

src/lib/pubkey/xmss/xmss_publickey.h
mgierlings: Thank you for the explanation, I'm going to fix this accordingly. I think a more verbose doxygen comment on those functions would be helpful.

randombit:
While your proposal does require minimal API change it has some very unfortunate consequences due to this, for an arbitrary HashFunction& h:

```
h.update(msg);
h.final(h_of_msg);
h.final(h_of_empty); // hash of empty string (or not!)
```

If `h` is an XOF `final` behaves differently, and depends on `msg` instead. That's subtle and obnoxious, especially when a particular algorithm might be XOF upgraded in the future (eg, Skein actually does support XOF-like outputs, though support for that was removed as seeming unnecessary - maybe it gets added back someday - suddenly the code above changes behavior).

Also re API changes - what is being offered would be completely new functionality, so if we can choose something which more clearly expresses at the call site WTF is going on that's helpful. Having applications be required to call a function named `final` many times to compute a result is not friendly.

I can see clear uses for XOFs with either a specific and fixed output lengths (eg needing SHA-3(1056) for Ed521) and also for very long and somewhat arbitrary lengths for stream cipher/KDF/RNG usage. But, the first is handled well by simply defining a SHAKE-128 hash that is paramaterized by the output length on instantiation, and the second by the SHAKE-128 stream cipher.  I have a hard time seeing a use case where the application wants a hash like interface (incremental updates) but then also needs a variable number of output blocks. So, I am inclined not to create this API edge case, just to support this hypothetical use.

In the end it may prove useful to express XOF concept with an entirely new API rather than tacked onto `HashFunction` since in fact XOFs are not cryptographic hash functions at all. For example AES-256/CTR keyed with SHA-256(input) is (AFAIK) a legitimate XOF. But, also, NIST has not really offered much guidance on XOF usage yet, nor are they too widely used yet in practice, I am really hesitant to adopt any interface (either your repeated `final` proposal or a new XOF class) that turns out to be insufficient for real usage.

randombit:
@mgierlings I just added SHAKE-128 and SHAKE-256 hashes in b7ae8043e963

mgierlings:
> I just added SHAKE-128 and SHAKE-256 hashes in b7ae804

Great thanks.

You also have a very valid point concerning the downsides of my proposal to implement dynamic output lengths with minimal API changes in botan. There are variations possible, like not recycling final to generate the output but to add another function that could be called `output()`, `squeeze()`, `generate()` ... but as you said it is hard to forsee the future and predict if we need certain functionalities like the "hashlike" incremental output. I just took a quick look at the SHA3/SHAKE implementations and it seems to be suitable for use in XMSS context. I'll take a closer look handle the SHAKE integration in a seperate commit though. It is always possible to add additional functionality later, once it is needed and thus more clear how it is going to be used.

For now I've pushed most of the changes we discussed above.  `max_input_bits()` is still there, since the [public key interface](https://github.com/randombit/botan/blob/master/src/lib/pubkey/pk_keys.h#L78) requires an overload, but it throws an [exception](https://github.com/randombit/botan/pull/717/files#diff-7159043f07651f6189934b71ec709206R21) now since it is not supposed to be called anyway.

Concerning the currently required buffering [workaround](https://github.com/randombit/botan/pull/717/files#diff-337be97fce04c50e906f906dca3a0084R106) for the message in `XMSS_Verification_Operation()`: Do you plan to introduce the [API changes to `PK_Ops::Verification()`](https://github.com/randombit/botan/blob/master/src/lib/pubkey/pk_ops.h#L70) before or after the merge? Either way I'd handle this in a seperate commit as well, what do you think?

src/lib/pubkey/xmss/xmss_signature_operation.h
```diff
@@ -0,0 +1,107 @@
...
+ *     draft-itrf-cfrg-xmss-hash-based-signatures-06
+ *     Release: July 2016.
+ *     https://datatracker.ietf.org/doc/
+ *     draft-irtf-cfrg-xmss-hash-based-signatures/?include_text=1
+ **/
+class BOTAN_DLL XMSS_Signature_Operation : public virtual PK_Ops::Signature,
+                                           public XMSS_Common_Ops
+   {
+   public:
+      typedef XMSS_PrivateKey Key_Type;
```
mgierlings: I think this, and [this](https://github.com/randombit/botan/pull/717/commits/8b06b4fe5fbe189c7d5250becb189bf2b87b9013#diff-f4a0f2b0dad4a5f63f992c7d8e704dd0R33) can go now too because of fc44288?

randombit: Correct

randombit:
I've merged to master. SHAKE support and the changes to unbuffer verification can happen in later PRs. I will work on the verification API part in the next few days.

PR #718, opened 2016-11-06
Adds XMSS WOTS stand-alone components

mgierlings:
The main XMSS implementation does not expose stand-alone access to XMSS
Winternitz One Time Signatures through the Botan::PK_Ops interfaces.
XMSS WOTS is used by XMSS internally though. This PR implements a
possible XMSS WOTS interface for stand-alone use.

XMSS WOTS is tested through the XMSS tests, therefore there are no
dedicated XMSS WOTS test vectors and test bench components implemented
in this PR.


randombit:
This would be a nice additional feature. Can you rebase?

mgierlings:
Sure, just finished the rebase =).

I just want to point out that standalone use of XMSS WOTS has some caveats and I'm not sure if it is going to be useful for library users at all. It might be of use internally and base line for a non XMSS-related WOTS implementation.

- Due to the pseudo-random key (re-)generation used in XMSS I had to introduce some wrapper classes `XMSS_Addressed_Public/Private_Key()`, which solely exist for the purpose of bundling an XMSS_Address with the appropriate key. I'm not really happy with this solution but since I could not add additional parameters for the PK_Ops calls to `sign()`, `is_valid_signature()` I had to solve this differently. This could something to keep in mind when refactoing these interfaces because of the message buffering issue. A way to provide additional, optional arguments either to the `PK_Ops::Signature`/`PK_Ops::Verification` or the call to `sign()`/`is_valid_signature()` might be useful also for other signature schemes that rely on pseudo-random key generation/tree structures etc. Maybe you have a completely different idea how to solve this more elegantly.

- Since the whole address and index management is done by the XMSS classes, when extracting XMSS WOTS as standalone procedure the user has to keep track of which key indices (addresses) have been used up. All the safety nets the XMSS classes provide here are gone. Key reuse can become a real problem here if the library user is not very careful.

- The message digest is created by [XMSS in Algorithm 12](https://www.ietf.org/id/draft-irtf-cfrg-xmss-hash-based-signatures-07.txt#26). Here a message `M'` is prepared and provided as input for the call to `tree_sig()` which then in turn calls `WOTS_sign()` with the pre digested message. So XMSS WOTS already expects a pre hashed message of length `n` as input. Thus it does not offer the hashlike `update()` mechanism for messages of arbitrary length like other signature schemes do. We could do something similar to what `XMSS_sign()` does when generating `M'` but we do not have `getRoot(SK)` at our disposal since this is part of XMSS. Omitting `getRoot(SK)` in the prefix or replacing it with another value would be outside standard territory.

randombit:
On the Address side, does something like this seem sufficient?

```
class PK_Signature_Context {
    // just a handle for dynamic_cast
     virtual ~PK_Signature_Context() {}
};

class XMSS_Address_Context : public PK_Signature_Context {
  XMSS_Address m_addr;
  ...
}
```

With `sign` and `verify` taking optional `PK_Signature_Context*` arguments, which WOTS can require and `dynamic_cast` to an Address and all existing algorithms can at least for now ignore as unused.

mgierlings:
Yes I like this idea. It is a very flexible solution and could be used by other algorithms that require additional arguments as well. If you add this during the `PK::Ops` interface changes I'll remove the addressed key classes and pass the addresses for XMSS WOTS through the `PK_Signature_Context`.

mgierlings:
One more thing just came to my mind: There should also be a setter available for `PK_Signature_Context`, otherwise and instance of `XMSS_WOTS_Signature_Operation`/`XMSS_WOTS_Verification_Operation` could only be used a single time. If we have a setter, we can change the context and reuse the existing object, reducing overhead by avoiding constructions and destructions, if many signatures are generated/verified in a short period of time.

randombit:
@mgierlings I may not have time to do these pk API changes we discussed before the next release (which I am hoping to do by early AM Monday 11/28). What do you think about merging WOTS support now for 1.11.34 (this PR as is) and I will work on the pubkey changes and cooresponding changes to WOTS/XMSS for the following release.

mgierlings:
Sure lets do this.

Issue #694, opened 2016-10-30
Improving the handbook


cordney:
We'd like to improve the handbook. It is currently partly incomplete and outdated and also missing complete topics, such as stream ciphers or PKCS#11. We'd also like to include reordering handbook pages, for example "The Low-Level Interface" should be moved up, probably coming right after "Memory container".

In general, we would like to follow the current structure of each handbook topic, which seems to be:

1) Basic information on the algorithm/protocol/whatever covered, e.g., MACs
2) Generic interface in Botan, e.g., `Botan::MessageAuthenticationCode`
3) One chapter for each algorithm contained in Botan, e.g., CBC-MAC, CMAC, GMAC, HMAC, ...
4) Example code, e.g., calculating an HMAC tag on a message and then verifying the tag

I must admit I really like the [Crypto++ wiki](https://cryptopp.com/wiki/Special:AllPages), where there is a page for almost every algorithm/protocol/whatever, with example code explained and even as a downloadable zip.

In addition to the topics already covered (that need to be checked for being up-to-date with the code and more info + example code added maybe), we think the following topics would be useful:
- Block Ciphers
- MACs
- Hash Functions
- Stream Ciphers
- Providers
- PKCS11
- The CLI
- The test suite

What do you think?

randombit:
All of this sounds great. The algorithm specific chapters would be very useful to provide to the reader some brief guidance on the proper use of that algorithm, for example the pitfalls of not having prefix-free messages with CBC-MAC. These chapters could also document the `BOTAN_HAS_XXX` macro name/value associated with the documented version. This would also be a good place to provide general guidance on safe usage.

Crypto++ wiki is very good, it would be amazing to have this level of detail in our docs though I would prefer most docs/examples be in `docs` as ReST instead of a wiki (because wiki spam is a pain and not everyone reads wikis).

randombit:
For some examples it might be nice to reference the relevant CLI code directly. To that end it may be useful to split up the CLI sources to a single command per file so they can be included from Sphinx. I'm not sure, probably code written for the express purpose of exposition will be clearer. The downside is API snippets that are not compiled tend to bitrot, and I'm not aware of any tools that can compile/test C++ snippets from documentation as there are for Python.

randombit:
Obviously the manual could still be much better in many ways, not sure if you had more in mind past #721/#725 though - should this be closed?

cordney:
[...]

PR #721, opened 2016-11-11
Documentation changes/additions: hash, symmetric crypto and pubkey

neverhub:
As proposed in https://github.com/randombit/botan/issues/694, we  restructured, updated and added various chapters of/ the Botan manual. In addition some sections are now covered with example code.

Not every addition is 100% complete (some sample code is still missing and will be added in the near future), but we would like to know what you think.


randombit:
Overall seems fine to me. I'm wondering if the CLI shouldn't be documented in its own section rather than scattering it through the manual, but probably doesn't matter much until at least some of them are documented.

Are you planning further commits on this PR, or would you prefer I merge this now and you opening new PR for future changes?

cordney:
I too think that we should have a separate page on the CLI in the end.

neverhub:
Thanks for the comments.
I am planning to commit the missing code examples and your remarks to this PR.
I will also add a separate CLI section soon.

randombit:
Is this good to merge at this point?

neverhub:
yes it's done so far.

Issue #711, opened 2016-11-03
State of HKDF


cordney:
I would like to know why HKDF was implemented in the first place and why only the expansion step? Users may be misled by the name `HKDF`, although it's mentioned in the doxygen API doc for the class.

From looking at the [latest TLS 1.3 draft](https://tools.ietf.org/html/draft-ietf-tls-tls13-14), it seems HKDF extraction and expansion steps are used independently, so maybe a good way to go would be renaming the current `HKDF` class to `HKDF_Expand`, adding the extraction step as `HKDF_Extract` and finally a wrapper class `HKDF` that simply invokes both the former.

randombit:
The original reason was to use standard HKDF in HMAC_RNG, instead of the variant it used. HMAC_RNG was based on the HKDF paper and predated standard HKDF by many years. Switching to standard HKDF would have made HMAC_RNG more testable. In that RNG design the extraction step was elsewhere, and only the expansion was needed, so that was what was implemented.

In the end I decided HMAC_DRBG was good enough, and already had 100% test coverage, so this change to HMAC_RNG never ended up happening.

In addition to TLS 1.3, I seem to recall seeing several protocols that used HKDF-Expand only, with the equivalent of the extraction step happening elsewhere in a key exchange. So these need to stay distinct steps. Renaming, and adding `HKDF` and `HKDF_Extract` as you suggest seems good to me.

cordney:
Ok, thx. Will give it a try.

randombit:
Merged, closing

PR #723, opened 2016-11-12
Add full HKDF implementation

cordney:
Adds the full HKDF as class HKDF, renames the existing HKDF, which only implemented the expansion step, to HKDF_Expand and adds the extraction step as HKDF_Extract.

The latter two are usually only used seperately in protocols such as TLS. A normal user would go for the
full HKDF.

Fixes GH #711.


randombit:
Code is fine, but in addition to `HKDF` combined tests there should be `HKDF_Extract` and `HKDF_Expand` tests also based on the same vectors (just create new sections for `HKDF_Extract` and `HKDF_Expand` and rename IKM/XTS values from master appropriately for Extract). Running at least one test for each named algo ensures the lookup code works correctly, that `name` returns the right thing, and so on.

cordney:
b3ee964 should do it. I already thought about that, but unnecessarily worried about duplicating test vectors.

cordney:
Regarding `KDF::name()`, I initially wanted to add that a test for that during this work, but encountered several errors that I will fix with a new PR, after this one is merged.

PR #724, opened 2016-11-14
Add ESP block cipher padding from RFC 4303

cordney:
ESP padding is one of the recommended block cipher padding methods from [BSI-TR-02102-1](https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Publikationen/TechnischeRichtlinien/TR02102/BSI-TR-02102.pdf;jsessionid=F5B2077757545C9B129BC7D57FB1ADEE.2_cid368?__blob=publicationFile&v=2). It simply pads with a monotonically increasing sequence: 1,2,3, ....


src/lib/modes/mode_pad/mode_pad.cpp
```diff
@@ -121,5 +125,44 @@ size_t OneAndZeros_Padding::unpad(const byte block[], size_t size) const
...
+      buffer.push_back(pad_value++);
+      }
+   }
+
+/*
+* Unpad with ESP Padding Method
+*/
+size_t ESP_Padding::unpad(const byte block[], size_t size) const
+   {
+   while(size)
```
randombit: Shouldn't this confirm all the padding bytes? Just truncating at the last found 0x01 seems very dangerous to me - it seems to me this function should both confirm last byte is <= block_size and confirm block[i-1] == block[i] for i >= (block_size - last_byte)

For extra challenge I'm pretty sure this can be done in entirely const time, just loop over the entire last block and ignore the input when i corresponds to plaintext bytes.

randombit:
Worth noting also RFC 4303 allows up to 256 bytes of padding, whereas Botan CBC only supports inspecting final block for padding bytes.

cordney:
I added checking of all padding bytes and also tried at const time, but I'm not sure this is what you meant.
I also added two negative tests to check the `Decoding_Error` branches.

randombit:
Sorry for being unclear, I meant with the helpers in `ct_utils.h` though the loop structure you have there seems like it would adapt well to use with masked operations. But this as it is, is fine with me for merging.

Really though all the cipher mode `unpad` functions should be written as const time functions to minimize the risk of padding oracles. Indeed AIUI IPsec even now allows ESP mode without a MAC, so the possibility of such timing oracles is quite real in some systems.

cordney:
Ok, let's merge as is and create an issue for making all BlockCipherModePaddingMethods const time using `ct_utils.h`. Issue because it should be fixed rather sooner than later.

PR #725, opened 2016-11-14
Add PKCS#11 documentation

neusdan:
#694

### DIFF
--- a/doc/manual/pkcs11.rst
+++ b/doc/manual/pkcs11.rst
@@ -1,0 +1,1244 @@
+.. _pkcs11:
+
+PKCS#11
+========================================
+
+.. versionadded:: 1.11.31
+
+|
+
+PKCS#11 is a platform-independent interface for accessing smart cards and
+hardware security modules (HSM). Vendors of PKCS#11 compatible devices usually
+provide a so called middleware or "PKCS#11 module" which implements the PKCS#11
+standard. This middleware translates calls from the platform-independent PKCS#11
+API to device specific calls. So application developers don't have to write smart card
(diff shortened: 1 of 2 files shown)


doc/manual/pkcs11.rst
```diff
@@ -0,0 +1,1244 @@
...
+
+   .. cpp:function:: PKCS11_ECDSA_KeyPair PKCS11::generate_ecdsa_keypair(Session& session, const EC_PublicKeyGenerationProperties& pub_props, const EC_PrivateKeyGenerationProperties& priv_props)
+
+----------
+
+Code example:
+
+   .. code-block:: cpp
+
+      Botan::PKCS11::secure_string pin = { '1', '2', '3', '4', '5', '6' };
```
randombit: This is not related to the documentation, but I am seeing a need for a `std::string` to `secure_string` conversion function here. ISTM like nearly every application using the PKCS11 interface will end up writing such a function.

mouse07410:
@neusdan, great! **Thank you!**

A question: I would like to use keys on a PKCS#11 token with TLS (that Botan also provides). How can I do that? A small example would be *very much* appreciated!

neusdan:
AFAIK the TLS API currently does not support using other providers.

mouse07410:
:-(

It would be great if all of the PK operations (where it makes sense) could use keys on a PKCS#11 token.

And it would be great if tls client and server could use it. Like OpenSSL does via pkcs11 engine.

cordney:
Discussing on a closed PR seems wrong, please use #730.

PR #727, opened 2016-11-15
Add a test for KDF::name()

cordney:
This required fixing some test vectors that previously omitted the MAC, defaulting to HMAC, that used the SHA-1 alias to SHA-160, and fixing `X942_PRF::name()`. For the latter, the constructor accepts an OID as well as a string, but the name includes the OID only, so test vectors now give the OID only for the `KDF::name()` test to work.


src/tests/data/kdf/kdf1.vec
```diff
@@ -1,4 +1,4 @@
-[KDF1(SHA-1)]
+[KDF1(HMAC(SHA-160))]
```
randombit: This seems wrong - KDF1 uses a hash, not a MAC

cordney: My fault, fixed in 10b8c8e.

PR #732, opened 2016-11-19
TLS-Attacker extended tests and fuzzing

jurajsomorovsky:
Extended server test suite and policy tests with TLS-Attacker.

TLS-Attacker server fuzzing.

### DIFF
--- a/src/cli/tls_server.cpp
+++ b/src/cli/tls_server.cpp
@@ -138,31 +138,42 @@ class TLS_Server final : public Command
                {
                while(!server.is_closed())
                   {
-                  uint8_t buf[4*1024] = { 0 };
-                  ssize_t got = ::read(fd, buf, sizeof(buf));
-
-                  if(got == -1)
+                  try 
                      {
-                     std::cout << "Error in socket read - " << strerror(errno) << std::endl;
-                     break;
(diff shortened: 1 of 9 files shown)


randombit:
Thanks for submitting this. I will merge this now and look at adding the testsuite portion to Travis CI sometime soon, so the tests are run reliably.

From my understanding, with the proper workflows defined TLS-Attacker is also capable of acting as a server which attacks clients which connect to it. It would be great to add tests for that, as well as for DTLS. The system in TLS-Attacker for this looks very easy to use but I have not yet had time to play with it much.

jurajsomorovsky:
Yes, adding further tests for TLS clients or DTLS functionality should be simple, but is not yet well tested. We are currently refactoring TLS-Attacker and adding new functionalities in the internal branches (e.g., new fuzzing strategies or an action system allowing you to change TLS protocol version in the middle of the protocol flow). Once this is done, we will add also new testing capabilities. However, this can take several weeks.

Thank you for accepting my pull request and integrating TLS-Attacker.

jurajsomorovsky:
Btw., you can find the (probably obvious) travis script I used previously here:
https://github.com/Hackmanit/botan-extended-tests/blob/master/.travis.yml

PR #733, opened 2016-11-19
Timing test suite

jurajsomorovsky:
Based on the mona timing library


randombit:
Thanks. Visualization from mona timing report may also prove useful for checking ECC scalar multiply, table based AES implementation and other key codes.

PR #734, opened 2016-11-22
Export work factor functions

cordney:
Now that users can implement custom PK ops via Private_Key and Public_Key outside the library, it makes very much sense to provide the work factor functions to them.


randombit:
Thanks. Would be good to have a couple of explicit tests for these functions also as they are some of the very few floating point operations in the library (#606 for a case where they seem to have been miscomplied on Android/ARM)

PR #736, opened 2016-11-25
Adds SHAKE support for XMSS

mgierlings:
- Enables code for shake support
- Creating SHAKE hash function by name now allows to select output size
  of 256 Bit for SHAKE128 and 512 Bit for SHAKE256.
- Adds *self-generated*, unverified test vectors for XMSS/SHAKE.


src/lib/hash/hash.cpp
```diff
@@ -251,10 +251,18 @@ std::unique_ptr<HashFunction> HashFunction::create(const std::string& algo_spec,
       {
       return std::unique_ptr<HashFunction>(new SHAKE_128(req.arg_as_integer(0, 128)));
       }
+   if(req.algo_name() == "SHAKE-128(256)")
```
mgierlings: This allows to create SHAKE-128/256 instances with 256/512 Bit output length by name. Maybe a more generic solution parsing the output length provided in braces is more desirable. However for the purpose of using SHAKE-128/156 covering these two cases is sufficient.

randombit: I don't think either of these is necessary (or would ever even be hit at runtime) - we already handle arbitrary SHAKE output lengths with the existing code (that's the `req.arg_as_integer(0, def_val)` which takes the first user param if any), f.ex see the test vector files where `SHAKE-128(1120)` and `SHAKE-256(2000)` are tested

PR #738, opened 2016-11-25
[For review] Add OCSP stapling support to TLS client

randombit:
Adds OCSP stapling support to client side TLS only. Tested as working against various servers.

Currently we don't send any ResponderIDs or nonce because as far as I know there is no good reason to send either. Requiring a nonce forces freshness, which means the server can't just send us the OCSP response they've already cached and hurts latency. And requesting any specific responder seems pointless: the server is likely to have (at most) one valid OCSP response to give to us for their end-entity cert.

These restrictions on the OCSP request probably won't fly for server side OCSP stapling support, but I'm not planning on supporting OCSP stapling on the server side at this point. That will probably come in along with TLS 1.3 sometime ~ next year.

Posting as a merge to `ocsp-network` for ease of review but actually planning on merging this PR directly to `master` sometime after `ocsp-network` is merged.

### DIFF
--- a/src/cli/tls_client.cpp
+++ b/src/cli/tls_client.cpp
@@ -11,6 +11,7 @@
 
 #include <botan/tls_client.h>
 #include <botan/x509path.h>
+#include <botan/ocsp.h>
 #include <botan/hex.h>
 
 #if defined(BOTAN_HAS_TLS_SQLITE3_SESSION_MANAGER)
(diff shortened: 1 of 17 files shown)


randombit:
Fixed trivial build error and rebased

randombit:
Merging this now for 1.11.34

PR #740, opened 2016-11-28
define NOMINMAX 1 before every include of windows.h

BMBurstein:
When creating the amalgamation on Windows, the min/max macros get defined and conflict with the code. This solves it.


randombit:
Merged just in time for 1.11.34, thanks!

noloader:
> `define NOMINMAX 1 before every include of windows.h`

Another way to approach the issue that does not involve macros and defines like `NOMINMAX` (which may cross pollinate into other code or user code):

    c = (std::min)(a, b);
    c = (std::max)(a, b);

The parenthesis around the function name avoid the problem with Microsoft supplying a macro for `min` and `max`. The same trick works for `numeric_limits::min()` and  `numeric_limits::max()`, too.

PR #744, opened 2016-11-28
Allow custom extensions when creating certificates

cordney:
We need to create X.509 certificates with custom extension which are used to authenticate clients to servers using a dynamically generated client certificate. For this purpose, we put different custom extension into this certificate. Botan currently does not support this, so this PR aims to add the following functionality:

* Allow custom extensions in `X509_Cert_Options` for creating self-signed certificates
* Allow custom extensions in CA-signed cert requests
* Add templated getter for (custom) extensions

I had to refactor `PKCS10_Request` to use an instance of `Extensions` to encode/decode extensions instead of a `Data_Store`. In case an extension is already present in a `PKCS10_Request` that is to be signed by a CA that overwrites this extension, I added `Extensions::replace()`.

Custom extensions can be added like this:

```cpp
class Custom_Extension : Certificate_Extension
   {
   ...
   };

Extensions extensions;
extensions.add(new Custom_Extension());
```

To read a custom extension from a certificate, I added a template getter:

```cpp
X509_Certificate cert("cert.crt");
auto extensions = cert.v3_extensions();
std::unique_ptr<Custom_Extension> custom_ext = extensions.get_raw<Custom_Extension>(custom_ext_oid);
```

Custom extension types need to `friend Extensions` or make `encode_inner()`, `decode_inner()` and `contents_to()` public.

I would like to hear your opinion on the new API. I must admit that I don't like that we need two lists to manage extensions in `class Extensions` via `m_extensions` and `m_extensions_raw`. That complicates our logic and leads to bugs, as can be seen in `Extensions::encode_into()`, which previously missed to encode any unknown extensions not encoded as `Unknown_Critical_Extension`.

### DIFF
--- a/src/lib/x509/pkcs10.cpp
+++ b/src/lib/x509/pkcs10.cpp
@@ -46,7 +46,7 @@ PKCS10_Request::PKCS10_Request(const std::vector<byte>& in) :
    }
 
 /*
-* Deocde the CertificateRequestInfo
+* Decode the CertificateRequestInfo
 */
 void PKCS10_Request::force_decode()
    {
@@ -120,11 +120,7 @@ void PKCS10_Request::handle_attribute(const Attribute& attr)
       }
    else if(attr.oid == OIDS::lookup("PKCS9.ExtensionRequest"))
       {
-      Extensions extensions;
-      value.decode(extensions).verify_end();
-
-      Data_Store issuer_info;
-      extensions.contents_to(m_info, issuer_info);
+      value.decode(m_extensions).verify_end();
       }
    }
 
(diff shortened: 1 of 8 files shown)


randombit:
I also don't like the two lists, but don't have anything better to suggest at this moment. Besides that the PR looks good to me on a first scan but I'm so tired it barely counts as a review, I am going to read it over again in the morning.

randombit:
Still have not had time to fully review, sorry. Subject is WIP, do you see a need for further changes here?

cordney:
Let me do some cleanups and add docs.

cordney:
Did some small renaming, added doxygen docs and updated the tests. Ready to be reviewed.

randombit:
LGTM, thanks!

PR #745, opened 2016-11-29
Make --via-amalgamation an error

webmaster128:
Given that we should get rid of ` --via-amalgamation` and ` --gen-amalgamation` entirely in Botan 2.0, this is an intermediate step for migration. Configuration should error but include a helpful message about what to do instead.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -2201,8 +2201,7 @@ def main(argv = None):
         raise Exception("--gen-amalgamation was removed. Migrate to --amalgamation.")
 
     if options.via_amalgamation:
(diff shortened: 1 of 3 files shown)


randombit:
The CI scripts and manual need to be updated before the option can be removed/disabled entirely.

webmaster128:
Ah sorry, I did not see the CI fails. I added the changes.

webmaster128:
Converted TEST_FLAGS to Bash array as well for consistency. I guess the remaining fails of 651c730 are independent of this PR (python 2 installation on OS X).

randombit:
LGTM, thanks

PR #747, opened 2016-12-01
1.10.14 build

alonbl:
release-1_10 branch is out of date, and github cannot create pull request for a tag 1.10.14, so sorry for the mess. all this patch does is std:to_string -> to_string.


randombit:
Thanks for the patch. Re branch status, I apparently just pushed the tags and forgot to push the branch.

PR #749, opened 2016-12-01
Updates module policy, fixes test suite errors

mgierlings:
- Moves SHAKE to prohibited algorithms
- Catches an error during XMSS private key generation when SHAKE is disabled.
  This should not be reported as error by the test bench since [SHAKE is
  optional in XMSS](https://tools.ietf.org/html/draft-irtf-cfrg-xmss-hash-based-signatures-07#section-5)


randombit:
Thanks

PR #756, opened 2016-12-04
Remove ECB mode

randombit:
It should not be easy to use ECB mode. ;)


tfussell:
I hope this is the right place to discuss this. I'm the creator of a C++ library for reading and writing xlsx files called [xlnt](https://github.com/tfussell/xlnt). I'm currently using Botan (one of my favorite C++ libraries) for reading files encrypted according to [MS-OFFICECRYPTO specification](https://msdn.microsoft.com/en-us/library/cc313071.aspx) which uses ECB cipher mode for pre-Office 2013 encrypted files. I'm going to stick to the last commit before this feature was removed from Botan for now, but I guess I'll have to switch to a different crypto library eventually. I don't want to tell you all what to do with your excellent library--just wanted to put in a vote for keeping ECB. Thanks.

randombit:
@tfussell Ouch that is really unfortunate. I figured that only possible users of ECB were accidental or experiments - that no real-world software/protocol required it anymore. Unfortunately even though ensuring Botan is usable for a wide range of applications, I think the library really is better off without ECB because outside of tests, or a few carefully contrived scenarios, any use of ECB mode (for example, as Office 2010 likely used it) is a catastrophic error. I am sorry to break your code, but an "easy" ECB mode is just too dangerous to keep.

One thing you can consider (as my "it should not be easy"  statement in the PR alluded to), is just using the low level BlockCipher interface to perform ECB mode. The following code is not tested and without handling for padding but should give the idea:

```
std::unique_ptr<BlockCipher> cipher(BlockCipher::create("DESorWhateverStupidThingTheyDid"));
cipher->set_key(whatever_key_you_passed_to_pipe_before);
// Presumably there is no IV
// Assumes full ciphertext blocks
size_t ciphertext_blocks = ciphertext_len / cipher->block_size();
// the buffers can be the same (but not overlapping)
cipher->decrypt_n(ciphertext_buffer, plaintext_buffer, ciphertext_blocks);
// output in plaintext_buffer, strip padding in last block
```

It is another codepath in your library instead (probably) just passing a different cipher string, but really not bad (what 4 lines, plus 1-3 for padding depending on what they did?). You will not find better with OpenSSL ;) The BlockCipher interface is very stable, and the above code (or the de-typoed version of it) will probably work for a long time. Anyway, something to consider as a simpler change vs switching libs entirely.

If you do end up switching libs, once again my apologies for having busted your code and making work for you. But ECB is unsafe in a crypto library distributed for general use.

Unrelated rant: Microsoft really shipped ECB crypto in Office *2010*? WTF. That should have been an embarrassment in 2000. I really am appalled. :(

tfussell:
I appreciate your taking the time to work through this with me. I agree that it's probably for the best if users don't have access to it in Botan. I was just hoping to avoid implementing my own cryptographic functions since it's so error prone. It looks like the low-level interface isn't terrible though and it's a good chance to learn a bit more about cipher modes. 

Thanks for the example. It certainly still beats all of the boilerplate code needed for OpenSSL.

As for Office, I think they needed random-access to different regions of the document without decrypting the whole thing into memory so ECB seemed like the easy choice. Newer versions use SHA512 with CBC over 4KB segments--a little better.

PR #757, opened 2016-12-05
Add X.509 subject public key info and PKCS#8 private key info getters on key classes

### LINKED ISSUE #685: Wrong behavior of RSA_PublicKey::x509_subject_public_key
The caller of this function would expect that it returns a X.509 subject public key as described in RFC5280:

```
SubjectPublicKeyInfo  ::=  SEQUENCE  {
        algorithm            AlgorithmIdentifier,
        subjectPublicKey     BIT STRING  }

```

But what it actually returns is a PKCS#1 RSAPublicKey:

```
RSAPublicKey ::= SEQUENCE {
          modulus           INTEGER,  -- n
          publicExponent    INTEGER   -- e
      }

```

cordney:
Adds new `Public_Key::subject_public_key()` that does the same as `BER_encode(Public_Key&)` and equivalent `Private_Key::private_key_info()` that does the same as `BER_encode(Private_Key&)`.

Renames existing `Public_Key::x509_subject_public_key()`, which in fact never returned an X.509 SubjectPublicKey, to `public_key_bits()` and equivalent `Private_Key::pkcs8_private_key()`, which never returned a PKCS#8 PrivateKeyInfo, to `private_key_bits()`.

This is an API change, so it's just the right time before 2.0.

Fixes #685.

### DIFF
--- a/src/lib/prov/tpm/tpm.cpp
+++ b/src/lib/prov/tpm/tpm.cpp
@@ -349,7 +349,7 @@ AlgorithmIdentifier TPM_PrivateKey::algorithm_identifier() const
                               AlgorithmIdentifier::USE_NULL_PARAM);
    }
 
-std::vector<byte> TPM_PrivateKey::x509_subject_public_key() const
+std::vector<byte> TPM_PrivateKey::public_key_bits() const
    {
    return DER_Encoder()
       .start_cons(SEQUENCE)
--- a/src/lib/prov/tpm/tpm.cpp
+++ b/src/lib/prov/tpm/tpm.cpp
@@ -361,7 +361,7 @@ std::vector<byte> TPM_PrivateKey::public_key_bits() const
 
 secure_vector<byte> TPM_PrivateKey::private_key_bits() const
    {
-   throw TPM_Error("PKCS #8 export not supported for TPM keys");
+   throw TPM_Error("Private key export not supported for TPM keys");
    }
 
 std::vector<uint8_t> TPM_PrivateKey::export_blob() const
--- a/src/lib/prov/tpm/tpm.cpp
+++ b/src/lib/prov/tpm/tpm.cpp
@@ -359,7 +359,7 @@ std::vector<byte> TPM_PrivateKey::public_key_bits() const
       .get_contents_unlocked();
    }
 
-secure_vector<byte> TPM_PrivateKey::pkcs8_private_key() const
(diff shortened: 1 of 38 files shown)


src/lib/prov/tpm/tpm.cpp
```diff
@@ -359,7 +359,7 @@ std::vector<byte> TPM_PrivateKey::x509_subject_public_key() const
       .get_contents_unlocked();
    }
 
-secure_vector<byte> TPM_PrivateKey::pkcs8_private_key() const
+secure_vector<byte> TPM_PrivateKey::private_key_bits() const
    {
    throw TPM_Error("PKCS #8 export not supported for TPM keys");
```
neusdan: Message has to be adjusted. `export not supported for TPM private keys` or something similar.

src/lib/pubkey/curve25519/curve25519.cpp
```diff
@@ -88,7 +88,7 @@ Curve25519_PrivateKey::Curve25519_PrivateKey(const AlgorithmIdentifier&,
    size_check(m_private.size(), "private key");
    }
 
-secure_vector<byte> Curve25519_PrivateKey::pkcs8_private_key() const
+secure_vector<byte> Curve25519_PrivateKey::private_key_bits() const
    {
    return DER_Encoder()
       .start_cons(SEQUENCE)
```
neusdan: Nothing to do with your change, but is it correct that the public **and** private key bits are encoded here?

cordney: Seems ok when looking into https://tools.ietf.org/html/rfc5915 and https://tools.ietf.org/html/draft-ietf-curdle-pkix-00#page-6. `publicKey` field is optional, but a SHOULD.

src/lib/prov/tpm/tpm.cpp
cordney: Fixed in d56812d, thanks.

src/lib/pubkey/curve25519/curve25519.cpp
randombit: There should be a comment here that what's being encoded is not an `ECPrivateKey` type at all but an ad-hoc botan specific format under a private OID. I suppose it would be possible to stick x25519 into ECPrivateKey but I don't know of any impl that transfers 25519 keys in that way.

randombit:
LGTM, thanks for doing this.

PR #758, opened 2016-12-05
fix heap buffer overflow in Ciphersuite::by_id()

neverhub:
std::lower_bound returns all_suites.end(), if the passed suite id is greater than 0xFFCB.
all_suites.end()  points one past the last element in the vector and thus not to one of the known ciphersuites.

[libfuzzer output](https://0bin.net/paste/53SCgavErbNOZ9c5#h8aJfFFa4XEiGbBetVFuDmRo1PoxoV2420uJr-K2rmA)

### DIFF
--- a/src/lib/tls/tls_ciphersuite.cpp
+++ b/src/lib/tls/tls_ciphersuite.cpp
@@ -43,7 +43,7 @@ bool Ciphersuite::cbc_ciphersuite() const
 Ciphersuite Ciphersuite::by_id(u16bit suite)
    {
    const std::vector<Ciphersuite>& all_suites = all_known_ciphersuites();
-   auto s = std::lower_bound(all_suites.begin(), all_suites.end(), suite);
+   auto s = std::lower_bound(all_suites.begin(), all_suites.end() - 1, suite);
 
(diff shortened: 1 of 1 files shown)


randombit:
Nice find. I fixed this in 8f58094 in a slightly different way, and also added a test which reproduced the problem.

This might be exploitable for a DOS but I do not see any real avenue for code execution. I would guess most of the time the read of uninitialized memory does not result in reading the expected ciphersuite code, in which case nothing happens. If the u16 does match (and it would have to be one of the values 0xFFCC...0xFFFF) then an Ciphersuite struct with invalid values is returned.

The most likely scenario for exploitation I can see if a server sending a client an invalid ciphersuite. But in that case, the client rejects the server hello almost first thing, because the ciphersuite isn't one that the client offered. The TLS stack does call `by_id` first, before throwing the exception (it happens more or less as a side effect of parsing the server hello, which isn't a great design), but AFAICT never uses the invalid Ciphersuite struct because the client rejects the connection almost immediately after.

randombit:
Closing as fixed. Based on analysis above, I don't think this needs a CVE or expedited release, let me know if you disagree.

PR #759, opened 2016-12-05
Add ISO9796-2 Signature Schemes giving message recovery 2 and 3

neverhub:
Adds the signature schemes 2 and 3 from ISO-9796-2:2010 to Botan.

The ISO standard 9796-2 defines 3 signature schemes. Only scheme 2 and 3 with respective test vectors taken from the ISO document and bouncycastle are part of this commit.

Signature scheme 2 is closely related to PSSR and utilizes a random salt. Scheme 3 is essentially scheme 2 without a salt and thus deterministic.


src/lib/pk_pad/iso9796/iso9796.h
```diff
@@ -0,0 +1,94 @@
+/*
+ * ISO-9796-2 - Digital signature schemes giving message recovery schemes 2 and 3
+ * (C) 2016 Tobias Niemann, Hackmanit GmbH
+ *
+ * Botan is released under the Simplified BSD License (see license.txt)
+ */
+
+#ifndef ISO9796_H
```
randombit: Header guard must match the format `BOTAN_XXX_H__` for the amalgamation script to work.

PR #760, opened 2016-12-07
Fix Travis CI docs

neverhub:
This should hopefully fix the docs CI build. But let's wait for travis.


neverhub:
Does fix the package issue, but another error (warning) occurs:
Sphinx 1.5 introduced a new warning.

- C++, show warnings when the kind of role is misleading for the kind of target it refers to (e.g., using the class role for a function).

`/home/travis/build/randombit/botan/doc/manual/pkcs11.rst:1092: WARNING: cpp:class targets a function (PKCS11_RNG::PKCS11_RNG).`

randombit:
Thanks!

PR #762, opened 2016-12-07
Add missing netinet/in.h include for sockaddr_in

louiz:
- Without it, botan doesn’t compile on (at least) freebsd.

- `man 7 ip` list this include

- [Posix states](http://pubs.opengroup.org/onlinepubs/000095399/basedefs/netinet/in.h.html) that sockaddr_in must be defined in netinet/in.h,


randombit:
Thanks for the patch

Issue #728, opened 2016-11-16
Cipher Mode unpadding operations should be constant time


randombit:
All of the `unpad` calls in `mode_pad.cpp` should be constant time code. That means: no conditionals or memory lookups which depend in any way on the block contents. Instead the padding scheme should either return 0 (signaling invalid padding) or the size of the padding (iff valid). Note this is a change from current behavior of throwing an exception on invalid, so ECB and CBC would need to be changed to check the return value, and if invalid truncate the return buffer and throw an exception from there.

Any good protocol is using Encrypt-then-Mac with a strong authentication algorithm so theoretically these unpad operations not being constant time is ok, but not all protocols are EtM or even have a MAC at all (for example old, deprecated but still widely supported versions of PGP and IPsec ESP do not use MACs in their encryption schemes) and acting as an oracle via timing channel is a bad idea.

A very good paper on const time programming is "The Program Counter Security Model" (https://eprint.iacr.org/2005/368.pdf). There are helper functions similar to the routines described in that paper in in `util/ct_utils.h`. Code written in this style is not as easy to understand as standard C, as well as being quite tricky to write at times, but it ensures there is no information leak by timing, cache or branch predictor side channels. See TLS CBC decoding (tls_cbc.cpp) for example of this for another CBC padding check done in const time.

#724 for original discussion.

Oddly, this is actually a really good issue for a first time contributor, if someone wanted to take it up. All that is required is to read the above paper (at least enough to get the idea), read and understand a single header (`ct_utils.h`, ~150 LOC) and translate 4 small functions in mode_pad.cpp - no further understanding of the library is required.

The test framework for padding schemes is currently inadequate - negative tests must be hard coded as was done in #724. I'll be fixing that part shortly.

randombit:
I should have referenced this ticket in the commit db3a868cb, which adds negative tests for cipher padding. I am not certain the current set of invalid inputs is sufficient to test all cases.

randombit:
Merged with @neverhub's changes in #765

PR #765, opened 2016-12-09
Rewrote bc unpad functions as const time operations.

neverhub:
The CT functions are kind of tricky and i hope i got everything right. The unpad functions now return the blocksize as padding position, if the padding is invalid.  

Let me know what you think ;).

[valgrind output](https://0bin.net/paste/AiRw5MYX+fhoc8Nz#a+H0soUap9q+fJ6QDgWyp15kpNbWfuSXQr41qt0eeLu) for proof of const time

corresponding issue: https://github.com/randombit/botan/issues/728

### DIFF
--- a/src/lib/modes/cbc/cbc.cpp
+++ b/src/lib/modes/cbc/cbc.cpp
@@ -19,8 +19,8 @@ CBC_Mode::CBC_Mode(BlockCipher* cipher, BlockCipherModePaddingMethod* padding) :
    {
    if(m_padding && !m_padding->valid_blocksize(cipher->block_size()))
       throw Invalid_Argument("Padding " + m_padding->name() +
-                                  " cannot be used with " +
-                                  cipher->name() + "/CBC");
+                             " cannot be used with " +
+                             cipher->name() + "/CBC");
    }
 
 void CBC_Mode::clear()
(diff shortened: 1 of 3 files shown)


src/lib/modes/cbc/cbc.cpp
```diff
@@ -243,6 +243,10 @@ void CBC_Decryption::finish(secure_vector<byte>& buffer, size_t offset)
 
    const size_t pad_bytes = BS - padding().unpad(&buffer[buffer.size()-BS], BS);
    buffer.resize(buffer.size() - pad_bytes); // remove padding
+   if(pad_bytes == 0 && padding().name() != "NoPadding")
```
randombit: That's a good idea.

randombit:
Nice! Will try to take a look at this soon. Writing code this way is not easy, and subtle bugs can sometimes arise. We may want to add more invalid padding tests and/or add a fuzzer that compares constant time unpadding ops in library with dumb/side channel vulnerable (but obviously 'correct') unpadding, as is done already for PKCSv1.5 decryption.

randombit:
I did a more careful read through, everything looks good to me so I will merge. Thanks for doing this change.

I do want to add some more decoding-invalid testcases for all the mode paddings before the next release, just to make sure we haven't missed anything.

PR #769, opened 2016-12-11
Comments desired: Simplify the license text by using (C) The Botan Authors

randombit:
The license text is getting quite long due to including all the contributors, and literally the *only* thing we require for people to use Botan is to distribute and display the license text. So it should be succinct and save paper/screen space.

I am not a copyright lawyer and have no idea if this is strictly legit. But it seems to be a common convention among open source projects, and cuts the length of the license in half.

### DIFF
--- a/authors.txt
+++ b/authors.txt
@@ -1,0 +1,41 @@
+The original author and current release manager is
+
+Jack Lloyd
+
+With extensive and ongoing contributions by several individuals
+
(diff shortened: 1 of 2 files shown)


mouse07410:
Looks fine to me. Of course, IANAL.

randombit:
OK well 3 oks, no objections and thankfully none of us are lawyers ;) so I'm going to merge.

PR #771, opened 2016-12-11
Convert to using standard uintN_t integer types

randombit:
Mostly a scripted change, and a big one. I've checked over the whole patch but a second set of eyes would be great.


neusdan:
have to stop now. looked till passhash9.h

neusdan:
Didn't see all changes but are there now cstdint includes missing or how is this handled?

randombit:
Already cstdint is included/used (in types.h) the typedefs used right now are

```
using byte   = std::uint8_t;
using u16bit = std::uint16_t;
using u32bit = std::uint32_t;
using u64bit = std::uint64_t;
```

randombit:
Rebased and fixed a merge conflict.

@neusdan If it is an incentive to finish the review, merging this is holding up creating the reindent PR :)

src/lib/ffi/ffi.h
```diff
@@ -89,7 +89,7 @@ int botan_make_error(const char* msg, const char* func, int line);
 
 // This would seem to require both saving the message permanently
 catch(std::exception& e) {
-return botan_make_error_from_transient_string(e.what(), BOTAN_FUNCTION, __LINE__);
+return botan_make_error_fromransient_string(e.what(), BOTAN_FUNCTION, __LINE__);
```
randombit: Some regexp gone gone I am sure. But, how did this pass CI?

src/lib/prov/pkcs11/pkcs11f.h
```diff
@@ -131,7 +131,7 @@ CK_PKCS11_FUNCTION_INFO(C_InitToken)
   CK_SLOT_ID      slotID,    /* ID of the token's slot */
   CK_UTF8CHAR_PTR pPin,      /* the SO's initial PIN */
   CK_ULONG        ulPinLen,  /* length in bytes of the PIN */
-  CK_UTF8CHAR_PTR pLabel     /* 32-byte token label (blank padded) */
+  CK_UTF8CHAR_PTR pLabel     /* 32-uint8_t token label (blank padded) */
```
randombit: Definitely not, thanks

src/lib/base/buf_comp.h
```diff
@@ -59,57 +59,57 @@ class BOTAN_DLL Buffered_Computation
...
+      * @param in the uint8_t to process
       */
-      void update(byte in) { add_data(&in, 1); }
+      void update(uint8_t in) { add_data(&in, 1); }
 
       /**
       * Complete the computation and retrieve the
       * final result.
-      * @param out The byte array to be filled with the result.
+      * @param out The uint8_t array to be filled with the result.
```
randombit: Yeah, some of these I had reverted already prior to my commit, but I left many where either is technically valid. But "byte string" is easier to parse than "uint8_t string", and that it is a uint8_t vs a future byte type vs unsigned char is kind of an implementation detail. I will revisit the changes at least to the comments for the major APIs.

src/lib/ffi/ffi.h
randombit: Oh passed CI because this code is just part of a comment sketching out future work to the API.

randombit:
Reverted most of the comments to use byte again, it is easier to read that way. I will squash this before merging, left fixups separate for now for easier followup review.

src/lib/math/bigint/bigint.h
```diff
@@ -557,17 +557,17 @@ class BOTAN_DLL BigInt
         }
 
      /**
-     * Encode a BigInt to a uint8_t array according to IEEE 1363
+     * Encode a BigInt to a byte array according to IEEE 1363
      * @param n the BigInt to encode
-     * @param bytes the length of the resulting secure_vector<uint8_t>
+     * @param bytes the length of the resulting secure_vector<byte>
```
neusdan: here `uint8_t` makes more sense i think ;-)

randombit: Indeed so, fixed

PR #776, opened 2016-12-17
Add support for brainpool curves in openssl provider

cordney:
OpenSSL 1.0.2 added support for brainpool curves, so we can use it provided the version check succeeds.


randombit:
Nice, thanks

PR #777, opened 2016-12-17
Update the X.509 manual and add more information

cordney:
This originally started as an "only update the section on Certificate Stores after adding the sqlite backend", but ended up adding more information on extension handling in general and custom extension handling in particular as well as updating the whole chapter to the current interfaces.

I was wondering if we should list all extensions implemented in x509_ext.h in the list of supported extensions, is there anything against it? 

On a side note, the OCSP manual page is pretty short. There was a refactoring by @randombit not long ago, would be great to see some bits of this added to the handbook. Maybe OCSP should also go into X.509 certificates handbook chapter.

### DIFF
--- a/doc/manual/x509.rst
+++ b/doc/manual/x509.rst
@@ -35,6 +35,10 @@ in the :doc:`tls` protocol. A X.509 certificate is represented by
 
      Returns the point in time the certificate expires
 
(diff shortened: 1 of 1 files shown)


randombit:
I'll do some updates to the OCSP doc and merge it to the X.509 chapter

PR #778, opened 2016-12-19
Reindent sources using a more compact style

randombit:
Kind of impossible to read the diffs here, so this PR is mostly created for any discussion and to have a historical record of the change.

I wavered on using astyle vs clang-format for some time, settled on clang-format, then hit this showstopper bug https://llvm.org/bugs/show_bug.cgi?id=17362 - right now clang-format moves all preprocessor statements to the first column, but the amalgamation generator needs to have conditional includes indented in order to detect them correctly. Until at least one of these issues is fixed we cannot use clang-format, at least not as a mass reformatter. I'm not entirely happy about astyles formatting in all cases, but it does get the source much closer to a conventional formatting, so it's good enough for now.


cordney:
What about also putting opening angle brackets in a newline. For example, Boost and Crypto++ use this style and we internally also switched to it (from same line) for our enterprise-internal libraries 2 or 3 years ago. IMHO this makes code more readable.

```cpp
uint64_t cc_derank(uint64_t cc_number)
{
   for (size_t i = 0; i != 10; ++i)
   {
      if (luhn_check(cc_number * 10 + i))
      {
      return (cc_number * 10 + i);
      }
   }
}

**Comment by neusdan** [comment]
Thanks a lot for this! Can you document somewhere the Astyle settings/command line options?

**Comment by neusdan** [comment]
What about whitespace after left parenthesis and before right parenthesis? 

```c++
std::unique_ptr<Botan::PBKDF> pbkdf( Botan::PBKDF::create( "PBKDF2(SHA-256)" ) )
```

in favor of

```c++
std::unique_ptr<Botan::PBKDF> pbkdf(Botan::PBKDF::create("PBKDF2(SHA-256)"));
```

webmaster128:
> What about whitespace after left parenthesis and before right parenthesis?

Hmm, the Cordpress code style. I do not like that an don't see any value in this

mgierlings:
Yes, a big +1 for @cordneys proposal to use Allman indent style. I like to have my matching braces in the same column. However I'd rather not like to see additional spaces inbetween braces.

neverhub:
What is exactly the problem with preprocessor directives indentation and clang-format?
I just reformatted a local repository with clang-format and configured and build with `--amalgamation`. I also enabled/disabled some modules for testing purposes.

Everything seems to work just fine, even with unindented conditional includes. Am i missing something here?

randombit:
@neusdan Sorry the options are in src/scripts/indent.sh Also, I think spaces between parens is too noisy sorry.

@neverhub Hm let me look into this again (won't be until tonight), I am pretty sure there is a problem. Even if not I'm not a huge fan of how clang-format changes the indents if ifdefs, but it's at least not a blocker then.

@cordney Actually IMO the main point of the reformat is moving to a more attached brace style to a) improve code/screen density b) at least roughly match style of other large modern C++ codebases. But given the +1s I will think about it some more.

randombit:
Deferring this until 2.0.1

randombit:
Closing in favor of followup to #963

PR #779, opened 2016-12-19
Add ECDH/ECIES blinding, improve DL_group verification and check DH public key against small-subgroup attacks

neverhub:
DH public key check is from [RFC2785](https://tools.ietf.org/html/rfc2785).
ECDH/ECIES blinding uses blinding from ECDSA.

### DIFF
--- a/src/lib/pubkey/dl_algo/dl_algo.cpp
+++ b/src/lib/pubkey/dl_algo/dl_algo.cpp
@@ -35,8 +35,8 @@ std::vector<byte> DL_Scheme_PublicKey::public_key_bits() const
    }
 
 DL_Scheme_PublicKey::DL_Scheme_PublicKey(const AlgorithmIdentifier& alg_id,
-                                         const std::vector<byte>& key_bits,
-                                         DL_Group::Format format)
+      const std::vector<byte>& key_bits,
+      DL_Group::Format format)
    {
    m_group.BER_decode(alg_id.parameters, format);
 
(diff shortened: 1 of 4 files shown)


src/lib/pubkey/dl_algo/dl_algo.cpp
```diff
@@ -67,6 +67,8 @@ bool DL_Scheme_PublicKey::check_key(RandomNumberGenerator& rng,
       return false;
    if(!m_group.verify_group(rng, strong))
       return false;
+   if(power_mod(m_y,group_q(),group_p()) != 1)
```
randombit: q may not be available in all groups

randombit:
Other than missing q == 0 check, LGTM

Are there tests that could be usefully added here? I bet verify_group is not currently tested at all so there may not be an existing test to hang off of.

neverhub:
verify_group() and check_key() are called, when testing dh_keygen. But separate negative/positive tests would be better. I will add the q==0 check.

cordney:
May be related, I was adding DH 2,048 bits test vectors to `dh.vec` from [CAVP KAS FCC file 20.1](http://csrc.nist.gov/groups/STM/cavp/documents/keymgmt/KASTestVectorsFFC2016.zip). It also has negative test vectors for failed public and private key checks. I enabled load checks in the tests but all tests still passed. The checks are listed in [SP 800-56A](http://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar2.pdf):

```
5.6.2.3.1 FFC Full Public-Key Validation Routine
...

1. Verify that 2 <= y <= p-2.
(This step is to ensure that the public key has the unique correct representation and
range for the field.)
2. Verify that 1 = y^q mod p.
(This step is to ensure that the public key has the correct order and thus is in
the correct subgroup.)
```

```
5.6.2.1.2 Owner Assurance of Private-Key Validity

Prior to the use of a key pair in a key-establishment transaction, the owner of a
static or ephemeral key-establishment key pair shall obtain an initial assurance
that the private key is an integer in the correct range
([1, q−1] for FFC or [1, n−1] for ECC), as determined by the domain
parameters (see Sections 5.5.1.1 and 5.5.1.2) ...
```

PR #785, opened 2016-12-27
In path validation, prohibit SHA-1 and 1024 bit RSA by default

randombit:
Effectively disables 1024 bit RSA as well as SHA-1; if you want to turn them on you have to at least change something. At this point broad consensus is that neither provides adequate security margin anymore.

This API for the path validation restrictions is kind of bad, since using 80 vs 110 for this parameter is actually a critical security setting, but that's not going to be obvious at all to most application programmers. It would be better if turning SHA-1 back on required something more explicit by the application such as

validation_restrictions.allow_signature_hash("SHA-1");

because then it's obvious to the reader that the application is doing something bad. But, I don't want to mess with this API just now.

### DIFF
--- a/news.rst
+++ b/news.rst
@@ -24,6 +24,12 @@ Version 1.11.35, Not Yet Released
 
 * Allow use of custom extensions when creating X.509 certificates (GH #744)
 
+* The default Path_Validation_Restrictions constructor has changed to
+  require at least 110 bit signature strength. This means 1024 bit RSA
+  certificates and also SHA-1 certificates are rejected by default.
+  Both settings were already the default for certificate validation in
+  TLS handshake, but this changes it for applications also.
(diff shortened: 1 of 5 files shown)


cordney:
Nice, thanks! I had this on my personal todo list. We don't have any explicit tests for path validation restrictions, have we? It would be nice to have at least some basic tests now making sure SHA-1 and RSA1024 really make path validation fail. We can always add more tests over time.

randombit:
That would be a great test to have. I can at least confirm that both SHA-1 and 1024 bit RSA usage caused the various tests to fail after I changed the default. But this should be checked on an ongoing basis.

cordney:
Not a blocker for me here. We can add it to the todo list.

PR #786, opened 2016-12-28
Fix building this fuzzer

alex:
(untested)


randombit:
Thanks, building these should really be part of CI.

PR #790, opened 2017-01-02
Add positive DH test vectors from NIST CAVP

cordney:
Adds positive tests for Diffie-Hellman KEX from NIST CAVP. This adds tests for DH with 2,048 bits.
Also adds a function for the test suite to load an optional `BigInt` from a vector file, as the CAVP test vectors contain a Q value.


cordney:
Added WIP, still working on something.

cordney:
Added two more commits:

* 7cc746b loads Q from .vec file for DH public keys
* abb7591 adds test vectors for which the public key checks must fail

I added the test vectors to the DH_KAT test suite, but I am not so confident that they really belong there.

randombit:
I split out the invalid keys to their own test vector file.  Anytime the amount of hardcoded test inputs exceeds ~~ the amount of chars it would take to write a new test reading from a file I prefer adding a new .vec file. It makes it easier to add tests over time, as compared to strings encoded in the source.

cordney:
Thanks, that absolutely makes sense.

PR #791, opened 2017-01-02
[WIP] Improve symmetry for CXXFLAGS and LDFLAGS

webmaster128:
For cross-compiling for Android, I need to set additional LDFLAGS and CXXFLAGS (which are not "abi flags").

Before, LDFLAGS could only be set at make time and CXXFLAGS could only be set at configure time via `--cc-abi-flags`.

This PR allows using the LDFLAGS/CXXFLAGS environment variable at make time. It would be possible to move those to configure time as well, if desired.

~~I think this supersedes the `--cc-abi-flags` configure time argument, which tends to be a messy place to hack in all kind of compile flags. Is there a good reason to preserve it?~~

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -993,9 +993,7 @@ class CompilerInfo(object):
         if options.cc_abi_flags != '':
             abi_flags += ' ' + options.cc_abi_flags
 
-        if abi_flags != '':
-            return ' ' + abi_flags
-        return ''
+        return abi_flags
 
     def cc_warning_flags(self, options):
         def gen_flags():
--- a/configure.py
+++ b/configure.py
@@ -956,7 +956,7 @@ class CompilerInfo(object):
     """
     Return the machine specific ABI flags
     """
-    def mach_abi_link_flags(self, options):
+    def mach_abi_flags(self, options):
         def all():
             if options.with_debug_info and 'all-debug' in self.mach_abi_linking:
                 return 'all-debug'
@@ -1516,7 +1516,7 @@ def create_template_vars(build_config, options, modules, cc, arch, osinfo):
         'mp_bits': choose_mp_bits(),
 
         'cxx': (options.compiler_binary or cc.binary_name),
-        'cxx_abi_flags': cc.mach_abi_link_flags(options),
+        'cxx_abi_flags': cc.mach_abi_flags(options),
         'linker': cc.linker_name or '$(CXX)',
 
         'cc_compile_flags': cc.cc_compile_flags(options),
--- a/src/build-data/makefile/gmake.in
+++ b/src/build-data/makefile/gmake.in
@@ -36,11 +36,11 @@ all: $(CLI) $(TEST)
 libs: $(LIBRARIES)
 
 $(CLI): $(LIBRARIES) $(CLIOBJS)
-	$(CLI_LINK_CMD) $(LDFLAGS) $(CLIOBJS) -L%{out_dir} -l%{libname} $(CLI_LINKS_TO) -o $(CLI)
+	$(CLI_LINK_CMD) $(CLIOBJS) -L%{out_dir} -l%{libname} $(CLI_LINKS_TO) -o $(CLI)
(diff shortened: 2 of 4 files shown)


webmaster128:
Updated the initial PR, which is working for me now. The remaining CI failures are related to #792.

webmaster128:
I believe the environment variables LDFLAGS/CXXFLAGS should be read at configure time and then be written into the Makefile. This is how cmake and autoconf and most other build systems I know work. It has the advantage that all configuration is done in one step and all flags are documented in the resulting Makefile.

Moving LDFLAGS from make time to configure time break build setups that work like

```
./configure.py ...
LDFLAGS=xy make
```

instead of

```
export LDFLAGS=xy
./configure.py ...
make
```

Is it okay to perform this incompatible change?

randombit:
@webmaster128 Reading the env variables at configure time seems like the right decision to me, for the reasons you outline. It is OK to perform this change, we should just make sure there is a clear release note on it. Do you want to do this change as part of this PR?

webmaster128:
Thanks for the feedback. I'll rework this one entirely after different things became clear to me.

I am not sure about one thing: is there a difference between CXXFLAGS and what we call --cc-abi-flags? Should both be used for every call of CXX, for compiling and linking?

randombit:
I think classically CXXFLAGS is only used for individual file compilation, whereas --cc-abi-flags original meaning is specifically for flags which affect ABI or which otherwise must be used at compile time, shared object link time, and exe link time (for example -m64, -fstack-protector, or -fsanitize=address). It just gets abused a lot to pass extra flags that should really be CXXFLAGS as they only affect compilation of individual source files.

In your proposal, since configure is already processing the information and exporting it into the makefile templates, it should probably also take --cxxflags= and --ldflags= options (better name? i am not sure if autoconf exposes this via a flag) which can be used instead of (or to override) settings in env variables.

webmaster128:
> it should probably also take --cxxflags= and --ldflags= options (better name? i am not sure if autoconf exposes this via a flag) which can be used instead of (or to override) settings in env variables.

I guess it is okay to provide just one way (for simplicity), and I prefer --cxxflags/--ldflags over CXXFLAGS/LDFLAGS because we use arguments for everything else.

> It just gets abused a lot to pass extra flags that should really be CXXFLAGS as they only affect compilation of individual source files.

Hmm... cmake passes everything from CXXFLAGS to the link call as well: running  `CXXFLAGS=-Dheyho12 cmake ..` results in the link command

> ./test/CMakeFiles/smartsqlite_tests.dir/link.txt:/usr/lib/ccache/c++   **-Dheyho12**  -std=c++11 -pthread -Wall -Wextra -Wpedantic -Wcast-qual -Wformat=2 -Wlogical-op -Wmissing-include-dirs -Wswitch-default -Wno-maybe-uninitialized -Woverloaded-virtual   CMakeFiles/smartsqlite_tests.dir/blob_test.cpp.o CMakeFiles/smartsqlite_tests.dir/connection_test.cpp.o CMakeFiles/smartsqlite_tests.dir/logging_test.cpp.o CMakeFiles/smartsqlite_tests.dir/nullable_test.cpp.o CMakeFiles/smartsqlite_tests.dir/scopedsavepoint_test.cpp.o CMakeFiles/smartsqlite_tests.dir/scopedtransaction_test.cpp.o CMakeFiles/smartsqlite_tests.dir/statement_test.cpp.o CMakeFiles/smartsqlite_tests.dir/version_test.cpp.o  -o smartsqlite_tests -rdynamic ../src/libsmartsqlite.a ../lib/gmock/libgmock.a ../lib/gmock/libgmock_main.a -ldl

And [Gentoo puts ABI information into CXXFLAGS](https://wiki.gentoo.org/wiki/Safe_CFLAGS), that are definitely required at link time. 

So maybe --cxxflags can replace --cc-abi-flags.

webmaster128:
Closing in favour of #1237.

PR #793, opened 2017-01-02
Add os: iOS

### LINKED ISSUE #620: system_rng on iOS
Currently there is no separate "ios" os in the botan build system and "darwin" is used for iOS builds. Thus for iOS configurations, there system_rng is marked as available, because darwin is

> Any system with /dev/random or CryptGenRandom

But there is no access to /dev/random or /dev/urandom on iOS, forcing us to use SecRandomCopyBytes.

Now the question is, how to get this into the build system? Create a separate "ios" os with no system_rng or add a 3rd system_rng implementation (duplicating the darwin_secrandom code)?

webmaster128:
This basically disables the module system_rng for iOS, where it does not work anyway.

Closes #620


randombit:
LGTM, thanks!

Issue #741, opened 2016-11-28
amalgamation with out-of-tree build doesn't work


BMBurstein:
I ran configure.py with --amalgamation and --with-build-dir (which, by the way, is documented in the manual as --build-dir) and could not run the resulting nmake file

webmaster128:
I just tried to reproduce the issue in [the ci](https://botan-ci.kullo.net/) (build job Windows32-static) where it works without a problem.

The trick might be that you need to have the Botan source directory as working directory, e.g.

```
call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\Tools\VsDevCmd.bat"

jom /J %NUMBER_OF_PROCESSORS% /F ..\build-botan\Makefile || goto error

goto :EOF

:error
exit /b 1
```

where `.\botan\` is the source directory and `.\build-botan\` the build directory.

webmaster128:
Did this solve the problem? Can this be closed?

BMBurstein:
I never went back and tried it. This looks like it would probably solve the problem, though whether this is "solved" or not is a question. The fact remains that just running configure/nmake does not work as expected, and s not documented otherwise.

webmaster128:
Being able to build the project from any other directory than Botan is unfortunatly entirely out of scope given the current buid system, available development ressources and use cases of this library. It took me quite some time to make out-of-tree builds possible at all.

What you describe is one of the many things you get for free when using cmake, which is not supported right now.

But why not add a section on out-of-tree builds to the docs?

randombit:
> Being able to build the project from any other directory than Botan is unfortunatly entirely out of scope given the current buid system

Wait, what doesn't work here? I do this all the time (eg, when building the fuzzers in src/extra_tests/fuzzers), which runs

```
../../../configure.py $(CFG_FLAGS) --with-build-dir=afl-build  ....
make -f afl-build/Makefile
```

No problems building the library itself at least. The only issue I know of with this would be that running tests needs --data-dir=.. option to point to the input files.

webmaster128:
Interesting, did not know that. What never worked is

```
./configure.py --with-build-dir=../build/botan
cd ../build/botan
make
```

Changing the directory also does not work inside the tree

```
mkdir -p sub1/sub2
cd sub1
../configure.py --with-build-dir=sub2
cd sub2
make
```

So the key might be that configure and make must be run from the same working directory.

--------------------------

Given that, we still get an a bug with (which is another topic)

```
[...]

PR #801, opened 2017-01-03
Fix relative path calculation for symlinked headers

webmaster128:
Before, the build dir needed to be a child of the source root in order for `dirs_up` to make sense.

This fixes the second issue described in https://github.com/randombit/botan/issues/741#issuecomment-270181476

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1807,20 +1807,10 @@ def portable_symlink(filename, target_dir, method):
         return
 
     if method == 'symlink':
-        def count_dirs(dir, accum = 0):
-            if dir in ['', '/', os.path.curdir]:
-                return accum
-            (dir,basename) = os.path.split(dir)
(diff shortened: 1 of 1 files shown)


randombit:
I pushed a bogus commit to master, which I then removed with reset --hard and a force push. (Which in retrospect was maybe not the best idea but there, I did it.) It looks like you picked up the commit in your branch. That's what is causing the Travis failures. Can you rebase against master? Otherwise looks good, thanks for the var rename.

webmaster128:
Urg, history changes on master, indeed not good. If possible, we should make all ling living branches [protected](https://help.github.com/articles/about-protected-branches/).

I rebased onto 4536a57020cdfb1c58316375999d3e50be704e92 from yesterday to get out of the danger area.

randombit:
Yeah master and release-* protected now.

PR #802, opened 2017-01-04
Update version to 2.0.0, update shared object names

randombit:
Cleans up so object naming since most of the time (across Unix) we follow the exact same naming scheme; just make it the default if only the so suffix is specified in the file. Removes the minor number from the library name, and moves it to after the ABI version in the shared object, before the patch. Then the shared object name follows the standard convention for Linux.

Also updates include header dir to be named botan-${major}

### DIFF
--- a/botan_version.py
+++ b/botan_version.py
@@ -1,8 +1,8 @@
 
-release_major = 1
-release_minor = 11
-release_patch = 35
-release_so_abi_rev = release_patch
+release_major = 2
+release_minor = 0
+release_patch = 0
+release_so_abi_rev = 0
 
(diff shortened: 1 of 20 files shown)


webmaster128:
Is windowsphone.txt intended to be hidden in this PR?

randombit:
Didn't seem to matter much either way, but probably better to include it with the PR adding the other patches for that OS (upcoming). New commit pushed that sticks to the main topic of the PR.

randombit:
Merging this shortly to allow testing. Still have 1 or 2 small PRs before release, ETA tonight.

PR #812, opened 2017-01-06
Return 0 bits in Darwin_SecRandom::poll on SecRandomCopyBytes failure

neusdan:


randombit:
Thanks. Was there a compiler warning on OS X for this? It might be worth exploring enabling -Werror to CI.

neusdan:
Good question, i have no idea. Noticed this while reading the code. I don't have an OS X system ;)

cordney:
It did generate a warning, I just check our macOS and iOS CI build logs.

PR #817, opened 2017-01-06
Add try/catch blocks in PKCS11 noexcept destructors that might throw

randombit:
@neusdan Coverity complains these noexcept destructors might throw Botan::PKCS11::PKCS11_ReturnError, I added try/catch blocks to ignore any errors during object destruction. What do you think?

### DIFF
--- a/src/lib/prov/pkcs11/p11_module.cpp
+++ b/src/lib/prov/pkcs11/p11_module.cpp
@@ -20,7 +20,14 @@ Module::Module(const std::string& file_path, C_InitializeArgs init_args)
 
 Module::~Module() BOTAN_NOEXCEPT
    {
-   m_low_level->C_Finalize(nullptr, nullptr);
+   try
+      {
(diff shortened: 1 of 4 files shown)


randombit:
Tests look OK and I think `try{ } catch(...) {}` is the only safe thing to do in this situation, so I'm going to merge this so another Coverity run can start.

neusdan:
That are false positives from Coverity. The try/catch blocks are unnecessary. When `nullptr` is passed to the PKCS#11 low level function as the last argument (parameter `ReturnValue*`) then the low level function  returns `true` on success or `false` on failure and no exception is thrown.

randombit:
OK thanks. I think it is easier to just leave the try/catch blocks so it is clear no exception escape is possible, yeah?

Maybe the comments should be updated though, to reflect that it should not actually ever happen.

neusdan:
Yeah of course, that's OK.

PR #820, opened 2017-01-07
Remove BCRYPT from todo.rst since it's already implemented

Bulat-Ziganshin:


randombit:
Sorry this todo is confusing, bcrypt as a standard password hash is implemented, this todo refers to creating a PBKDF class that uses bcrypt (which is not implemented). This would allow easily deriving symmetric keys from a passphrase using the bcrypt algorithm. I believe OpenSSH has a format defined for this but I haven't dug into it further.

Issue #821, opened 2017-01-07
Cygwin 6.3 and "ERROR: global name 'self' is not defined"


noloader:
I'm working on a Cygwin 6.3 x86_64 machine hosted by Windows 8.1.

```
$ ./configure.py
   INFO: ./configure.py invoked with options ""
   INFO: Platform: OS="CYGWIN_NT-6.3" machine="x86_64" proc=""
   INFO: Guessing target OS is cygwin (use --os to set)
   INFO: Guessing to use compiler gcc (use --cc to set)
   INFO: Guessing target processor is a x86_64/x86_64 (use --cpu to set)
   INFO: Target is gcc-cygwin-x86_64-x86_64
  ERROR: global name 'self' is not defined
```

And:

```
$ python --version
Python 2.7.12
```

It also shows up on Cygwin i686, which is probably not surprising:

```
$ ./configure.py
   INFO: ./configure.py invoked with options ""
   INFO: Platform: OS="CYGWIN_NT-6.3-WOW" machine="i686" proc=""
   INFO: Guessing target OS is cygwin (use --os to set)
   INFO: Guessing to use compiler gcc (use --cc to set)
   INFO: Guessing target processor is a x86_32/i686 (use --cpu to set)
   INFO: Target is gcc-cygwin-x86_32-i686
  ERROR: global name 'self' is not defined
```

randombit:
Hm, interesting one. Can you re-run with `--verbose` (should produce a backtrace for the exception)

noloader:
@randombit,

```
  DEBUG: Matched 'x86_64' to processor '('x86_64', 'x86_64')'
   INFO: Guessing target processor is a x86_64/x86_64 (use --cpu to set)
   INFO: Target is gcc-cygwin-x86_64-x86_64
  DEBUG: Program sphinx-build not found
  DEBUG: Traceback (most recent call last):
  File "./configure.py", line 2395, in <module>
    main()
  File "./configure.py", line 2261, in main
    logging.warning('Shared libs not supported on %s, disabling shared lib support' % (self.infofile))
NameError: global name 'self' is not defined

  ERROR: global name 'self' is not defined
```

randombit:
Thanks, fixed in master, will be merged to release branch before next release. Seems like a straightforward fix so I'm closing, reopen if c46692e fails to resolve the issue.

[...]

PR #827, opened 2017-01-08
Use pylint to improve code quality of configure.py

webmaster128:
As mentioned in #821, pylint is a tool that can help us identify bugs or potential problems in the configure script before they happen on the user's machine. This is especially valuable since in Python a lot of errors are not detected until the code branch is executed.

The default pylint configuration applies PEP8, so we have snake_case variable names and a bunch of other conventions.

While pylint can be annoying in some cases, in my experience it helps a lot in the longer run. And in those cases, we can silent pylint rules project global, per module or even per line.

The long .pylintrc file is more or less the default. I only applied some things that are useful for us:

- turn off full reports (reports=no)
- disable rule "missing-docstring"
- disable rule "logging-not-lazy" (because we don't care for speed)
- set line length limit to 120
- set line limit per module to 3000
- allow short variable names

I think it is another one or two hours to resolve the remaining issues, but I think for a basic setup to get the idea, this is a good first step.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -40,6 +40,9 @@ import botan_version
 if 'dont_write_bytecode' in sys.__dict__:
     sys.dont_write_bytecode = True
 
+class ConfigureError(Exception):
+    pass
+
 def flatten(l):
     return sum(l, [])
 
@@ -71,9 +74,6 @@ def get_vc_revision():
             return '%s:%s' % (cmdname, rev)
         except OSError as e:
             logging.debug('Error getting rev from %s - %s' % (cmdname, e.strerror))
-            return None
-        except Exception as e:
-            logging.debug('Error getting rev from %s - %s' % (cmdname, e))
             return None
 
     vc_command = ['git', 'rev-parse', 'HEAD']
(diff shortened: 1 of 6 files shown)


configure.py
```diff
@@ -2393,8 +2410,7 @@ def release_date(datestamp):
 if __name__ == '__main__':
     try:
         main()
-    except Exception as e:
-        import traceback
+    except ConfigureError as e:
```
randombit: This last outmost `except`  should stay `Exception` - we want to capture all runtime errors here, not just ones we generated.

src/scripts/ci/travis/install.sh
```diff
@@ -20,6 +20,9 @@ fi
 if [ "$TRAVIS_OS_NAME" = "linux" ]; then
     sudo apt-get -qq update
 
+    pip install --user pylint
+    pip3 install --user pylint
+
```
randombit: The pylint run should go into it's own Travis build without a binary, like docs target does.

That way when `pip pylint` install breaks some day, it just fails the pylint sub build, not all Travis builds. And when it fails it is obvious what is breaking.

configure.py
webmaster128: Do we really want to catch stuff we are not prepared for? I think everything that is thrown without someone thinking about the consequences should crash the program with stack trace.

src/scripts/ci/travis/install.sh
webmaster128: Good idea!

configure.py
randombit: Capturing and logging the stack trace regardless of what happened during `main` is exactly what this top level block is for.

webmaster128:
Good, htanks for feedback. I'll implement it

webmaster128:
Okay, I think I have everything in there now

PR #830, opened 2017-01-09
build: pkg-config: fix include directory

alonbl:

### DIFF
--- a/src/build-data/botan.pc.in
+++ b/src/build-data/botan.pc.in
@@ -1,7 +1,7 @@
 prefix=%{prefix}
 exec_prefix=${prefix}
 libdir=${prefix}/%{libdir}
-includedir=${prefix}/include/botan-%{version_major}
+includedir=${prefix}/include/botan-%{version_major}.%{version_minor}
 
 Name: Botan
 Description: Crypto and TLS for C++11


randombit:
Thank you for noticing this and the patch. But actually this is not the correct fix, `botan config cflags` output is correct - actually the include dir on install is incorrect. :( The minor number should have been removed from the dir name. I'll patch this issue shortly. I think this fix will be worth releasing a 2.0.1 shortly (within next day or so).

Reasoning to remove minor number from include dir name: since SemVer dictates that Botan-2 API only be extended in future releases, there is no reason to ever install two versions of it in parallel, it is always best to just take the latest release. This was unlike the situation for botan-1.10 and botan-1.11 where there are valid reasons to have both installed in parallel indefinitely (due to different software using the two distinct APIs), and so including the minor number ensured both could be installed.

But we do not remove botan-N prefix entirely, because it allows still 1.10+2,  and in the future 3, whose support lifetime will overlap 2.

Hope this makes sense.

alonbl:
If you want side by side then make sure all artifacts are qualified, for
example /use/bin/botan should be botan-2, and also docdir in similar
notation.

בתאריך 10 בינו׳ 2017 3:34,‏ "Jack Lloyd" <notifications@github.com> כתב:

Thank you for noticing this and the patch. But actually this is not the
correct fix, botan config cflags output is correct - actually the include
dir on install is incorrect. :( The minor number should have been removed
from the dir name. I'll patch this issue shortly. I think this fix will be
worth releasing a 2.0.1 shortly (within next day or so).

Reasoning to remove minor number from include dir name: since SemVer
dictates that Botan-2 API only be extended in future releases, there is no
reason to ever install two versions of it in parallel, it is always best to
just take the latest release. This was unlike the situation for botan-1.10
and botan-1.11 where there are valid reasons to have both installed in
parallel indefinitely (due to different software using the two distinct
APIs), and so including the minor number ensured both could be installed.

But we do not remove botan-N prefix entirely, because it allows still
1.10+2, and in the future 3, whose support lifetime will overlap 2.

Hope this makes sense.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/randombit/botan/pull/830#issuecomment-271461390>, or mute
the thread

PR #832, opened 2017-01-10
Fix various pylint warnings

randombit:
Mostly fixing indentation warnings. There are other pylint warnings that I need to look at, but leaving those for another time/PR.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -3,7 +3,7 @@
 """
 Configuration program for botan
 
-(C) 2009,2010,2011,2012,2013,2014,2015,2016 Jack Lloyd
+(C) 2009,2010,2011,2012,2013,2014,2015,2016,2017 Jack Lloyd
 (C) 2015,2016 Simon Warta (Kullo GmbH)
 
 Botan is released under the Simplified BSD License (see license.txt)
@@ -102,7 +102,7 @@ class BuildConfigurationInformation(object):
 
     # This is used on Darwin for dylib versioning
     version_packed = version_major * 1000 + version_minor
-    
+
     """
     Constructor
     """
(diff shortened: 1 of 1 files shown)


webmaster128:
I have a bunch of other pylint fixes ready locally, which I am happy to provide as soon as this one is in.

webmaster128:
@neusdan Could you please have a look at this diff in the `except ValueError:` body? This PR changes behaviour because the f.write is moved out of the body. Which way is the right one?

Please let me do the change because I have a lot of commits ready that touch this code.

neusdan:
I can't remember exactly what this try catch actually handles, but i think it handles the situation when configure.py is called from different directories. This affects for example the `src` var. This could be a full path like `C:\botan\src\cli\compress.cpp` or just `src\cli\compress.cpp`.

Anyway, i think it is not necessary. The following works for me too:

```
@@ -1255,23 +1255,15 @@ def gen_bakefile(build_config, options):
         for src in sources:
             (dir,filename) = os.path.split(os.path.normpath(src))
             dir = dir.replace('\\','/')
-            try:
-                param, dir = dir.split('/src/',1)
-                file.write('\tsources { src/%s/%s } \n' % ( dir, filename))
-            except ValueError:
-                pass
-            file.write('\tsources { %s/%s } \n' % ( dir, filename))
+            param, dir = dir.split('src/',1)
+            file.write('\tsources { src/%s/%s } \n' % ( dir, filename))
 
     def bakefile_cli_headers(file, headers):
         for header in headers:
             (dir, filename) = os.path.split(os.path.normpath(header))
             dir = dir.replace('\\','/')
-            try:
-                param, dir = dir.split('/src/',1)
-                file.write('\theaders { src/%s/%s } \n' % ( dir, filename))
-            except ValueError:
-                pass
-            file.write('\theaders { %s/%s } \n' % ( dir, filename))
+            param, dir = dir.split('src/',1)
+            file.write('\theaders { src/%s/%s } \n' % ( dir, filename))
 
     def bakefile_test_sources(file, sources):
         for src in sources:
```

[0002-removed-unnecessary-try-catch.patch.txt](https://github.com/randombit/botan/files/698620/0002-removed-unnecessary-try-catch.patch.txt)

Issue #828, opened 2017-01-08
Build system issues


neusdan:
I am executing configure.py as follows:

```
configure.py --module-policy bsi --enable-modules=tls,pkcs11,xts
```

Generated output:

```
INFO: configure.py invoked with options "--module-policy bsi --enable-modules=tls,pkcs11,xts"
   INFO: Platform: OS="Windows" machine="AMD64" proc="Intel64 Family 6 Model 94 Stepping 3, GenuineIntel"
   INFO: Guessing target OS is windows (use --os to set)
   INFO: Guessing to use compiler msvc (use --cc to set)
   INFO: Guessing target processor is a x86_64/x86_64 (use --cpu to set)
   INFO: Target is msvc-windows-x86_64-x86_64
   INFO: Skipping, incompatible OS - darwin_secrandom dev_random fd_unix proc_walk
   INFO: Skipping, not requested - adler32 aont bcrypt certstor_sql certstor_sqlite3 chacha_sse2 codec_filt compression crc24 crc32 cryptobox eme_raw ffi filters fpe_fe1 http_util keccak newhope passhash9 pbkdf1 rfc3394 sessions_sql sessions_sqlite3 shake_cipher srp6 tls tls_cbc tss xts
   INFO: Skipping, prohibited by module policy - blake2 blowfish camellia cascade cast cbc_mac ccm cecpq1 cfb chacha chacha20poly1305 comb4p curve25519 des eax elgamal emsa_raw emsa_x931 gost_28147 gost_3410 gost_3411 hkdf idea idea_sse2 kasumi kdf1 kdf2 lion mce mceies md4 misty1 noekeon noekeon_simd ocb ofb poly1305 prf_x942 rc4 rfc6979 rmd160 salsa20 seed serpent serpent_simd shake siphash siv skein threefish threefish_avx2 tiger twofish whirlpool x919_mac xtea
   INFO: Skipping, requires external dependency - boost bzip2 lzma openssl pkcs11 sqlite3 tpm zlib
[...]

PR #835, opened 2017-01-10
Load pkcs11 module by request

cordney:
The pkcs11 module once required the pkcs11 headers as an external dependency, but the headers were included a while ago. Still, the module was set to be load_on vendor. Changes it to be loaded by request. This seems the right choice - instead of enabling it by default - given that not all users would need pkcs11 support.

### DIFF
--- a/src/lib/prov/pkcs11/info.txt
+++ b/src/lib/prov/pkcs11/info.txt
@@ -1,6 +1,6 @@
 define PKCS11 20160219
 
-load_on vendor
+load_on request
 
 <requires>
 dyn_load


cordney:
Note that this fixes the part of #828 about the pkcs11 module showing up in "requires external dependency". It also removes the ``--with-pkcs11`` option, so the pkcs11 module can be enabled/disabled as every other module via ``--enable/disable-modules`` now.

randombit:
Is there any real reason not to just have PKCS11 support enabled by default? It is only necessary for some users, but it does not affect any users who do not need it, aside from larger binary size. But if that is a concern the user would hopefully have read the manual and know to use `--minimized-build`.

Another reason to enable PKCS11 by default is that many package distributions just configure.py && make with effectively no extra options not strictly required by their build environment. They are unlikely to enable PKCS11 explicitly, unless we find and bug each package maintainer. But if PKCS11 support is just enabled by default, then the Linux distros (when they ship Botan-2) will include the PKCS11 support in their packaged versions.

Worth knowing this PR as is does not actually remove configure options such as --with-pkcs11. That's because arg parsing happens before we parse build-data, so the --with-zlib and co are hardcoded into configure.py. Also (AFAIK) --enable-modules=pkcs11 should already work, with --with-pkcs11 effectively just a shorthand.

mouse07410:
I prefer having PKCS#11 enabled by default.

cordney:
Ok, convinced. Opened #837.

PR #837, opened 2017-01-11
Enable PKCS11 module by default and remove --with-pkcs11 option

cordney:
The pkcs11 module once required the pkcs11 headers as an external dependency, but the headers were included a while ago. Still, the module was set to be load_on vendor. Instead, we can enable the module by default now.


cordney:
@neusdan Noticed that --with-pkcs11 is still used in CI scripts and in the pkcs11 manual. I'll update the PR.

PR #858, opened 2017-01-27
Fix private key ctors API docs

cordney:
The `*_PrivateKey` ctors wrongly state that they expect PKCS#8 key bits, where they really expect a key-specific encoding, e.g., RFC 5915's ECPrivateKey for ECC private keys. For loading a key from PKCS#8 we have a different API (`PKCS8::load_key()` functions).


cordney:
Damn, I still sometimes forget to add [ci skip].

PR #868, opened 2017-02-05
Remove function comments n*4/3 and n*3/4 in base64 and add two functions

evpo:
**Commit comments**

The parameter comments were misleading because they did not take
into account that input_length must be rounded up to a multiple of
3 and 4 for encode and decode respectively. Two new functions were
added to calculate the correct maximum output length.

**Rationale**

This change is important because, for example, the recommended length of the output buffer for encode was as below:

  input_length*4/3

This would result in a buffer overrun for some lengths. For example:

  size_t input_length = 32;
  size_t buffer_length = (input_length * 4)/3; // 42

When it's rounded up:

  size_t buffer_length = ceil(static_cast<double>(input_length) * 4)/3); // 43

Both interpretations are not enough. There will be 44 bytes written, which is 1 to 2 bytes of invalid writes.

The calculation is too complex to remain as a comment. It's probably better to document it with help of a function.

**Implementation details**

The functions are defined in the cpp because they will probably not be performance critical and they use round_up that is an internal function.

**More information**

See this [stackoverflow](http://stackoverflow.com/questions/13378815/base64-length-calculation) question.

### DIFF
--- a/src/lib/codec/base64/base64.cpp
+++ b/src/lib/codec/base64/base64.cpp
@@ -78,7 +78,7 @@ size_t base64_encode(char out[],
 std::string base64_encode(const uint8_t input[],
                           size_t input_length)
    {
-   const size_t output_length = (round_up(input_length, 3) / 3) * 4;
+   const size_t output_length = base64_encode_max_output(input_length);
    std::string output(output_length, 0);
 
    size_t consumed = 0;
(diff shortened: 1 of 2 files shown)


randombit:
This is a good improvement, simplifies life for callers and makes it obvious what the correct value should be. Thanks.

PR #869, opened 2017-02-06
Add tests for DHE_DSS ciphersuites

cordney:
As long as we support DSS ciphersuites, we should test them, even if they're deprecated. Adds tests for the DHE_DSS ciphersuites recommended by BSI TR-02102-2.

### DIFF
--- a/src/tests/unit_tls.cpp
+++ b/src/tests/unit_tls.cpp
@@ -1,6 +1,7 @@
 /*
 * (C) 2014,2015 Jack Lloyd
 *     2016 Matthias Gierlings
+*     2017 René Korthaus, Rohde & Schwarz Cybersecurity
 *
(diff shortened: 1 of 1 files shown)


randombit:
Thanks. And for the record I absolutely agree, something being deprecated is no reason not to test it fully.

There is some problem with this PR, where the test policy ends up being unable to choose any usable suite.

Failure 1: TLS v1.2 TLS client with error Policy does not allow any available cipher suite
Failure 2: TLS v1.2 TLS client with error Policy does not allow any available cipher suite

randombit:
Oh I see the problem, both AES CBC ciphersuites for DSS use SHA-256, not SHA-384

cordney:
Thanks so much, I wouldn't have time to look into it until the evening.

PR #872, opened 2017-02-07
Add TLS::Policy::allow_client_initiated_renegotiation

randombit:
Parallel of the server policy flag.

Would be good to test that both of these flags (client and server side) work correctly.

### DIFF
--- a/doc/manual/tls.rst
+++ b/doc/manual/tls.rst
@@ -957,6 +957,14 @@ policy settings from a file.
 
      Default: false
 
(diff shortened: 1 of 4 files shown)


randombit:
This needs tests added before merging.

randombit:
Ugh, well still no test for this (or the corresponding client side policy, either) but going to merge this for 2.3 because as-is there is no good way for a server to disable renegotiation attempts. I plan on doing some TLS work in 2.4 and hopefully we'll get tests for this then.

randombit:
Some reason github isn't seeing this as a merge.

PR #876, opened 2017-02-12
Remove std::cout accidentally left from debugging

cordney:
Though I need to express that I would really like to see a --verbose parameter to botan-test for tracing that, e.g., outputs everything passed to `test_note()`, not only on failure.


randombit:
A `--verbose` flag would be good

PR #878, opened 2017-02-14
Fix botan-test verbose mode

cordney:
Verbose mode was already in, but a missing check prevented it from being enforced. Enable with `--verbose`.


randombit:
Thanks for looking into this.

PR #879, opened 2017-02-16
Fix rsa_blinding tests and add one more test

neusdan:
If Botan is configured without `emsa_raw` the rsa_blinding tests fail:

```
rsa_blinding ran 1 tests 1 FAILED
Failure 1: Could not find any algorithm named "Raw"
Tests complete ran 1 tests in 749.40 msec 1 tests failed
Error: Test error Test suite failure
```

This is caused by a wrong define: `#if defined(BOTAN_HAS_EME_RAW)` instead of `#if defined(BOTAN_HAS_EMSA_RAW)`. The test is using the signature generation/verification operation and therefore is using `emsa_raw` instead of `eme_raw`.

Additionally i wrote a test for rsa_blinding in case `eme_raw` is available and `emsa_raw` is not.

### DIFF
--- a/src/tests/test_rsa.cpp
+++ b/src/tests/test_rsa.cpp
@@ -5,6 +5,7 @@
 */
 
 #include "tests.h"
+#include "test_rng.h"
 
 #if defined(BOTAN_HAS_RSA)
   #include <botan/rsa.h>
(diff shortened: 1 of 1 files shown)


neusdan:
One failure in some of the Travis CI builds.

```
RSA blinding ran 225 tests 1 FAILED
Failure 1:  failed to throw expected exception
```

I have no idea why. This does not occur on my local machine. Tested Windows, and Linux with gcc and clang.

randombit:
No idea why this would happen either. I'll work in improving the test output so this is easier to debug.

randombit:
This patch looks ok outside of the CI failure. I don't see what the problem could be, and I just ran the RSA blinding tests (with this PR merged) 1000 times with no failures, so I'm going to merge.

randombit:
Weird, no problems on my machine but as soon as I merged, *all* the CI builds failed. Wuh?

I tracked it down to the test trying to encrypt an all-zero vector (``std::vector<uint8_t>(16)``) in the final test. Since no padding is in use, that leads to trying to compute `0^e mod n` during encryption, and that computation fails with in invalid argument from `Power_Mod`.  Fixed I think in 8eacc25b

randombit:
Additional problem, if OpenSSL was used then the library blinder code is bypassed completely and the expected exception did not arrive. Changed by forcing that test to use "base" provider.

PR #880, opened 2017-02-16
Add ECDSA timing tests (and discussion of timing leak countermeasures)

neverhub:
Timing attacks exploit algorithms not running in constant time and we might have a ECDSA timing problem in Botan. The current implementation (with default build.h macros) leaks information about the length of the ECDSA nonce k. The problem lies within ECC Blinded_Point_Multiply, which leaks the size of the scalar.

The problem is already known ([manual](https://botan.randombit.net/manual/side_channels.html#ecc-scalar-multiply)) . The problem can be fixed by easy to implement timing countermeasures, as the following timing reports show.

**secp384r1** and
```
k_1 = 01
k_2 = FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
```
were used for the reports.

without scalar blinding and with fixed window exp: 
21-scatterplot-40-50

without scalar blinding and with Montgomery ladder:
21-scatterplot-40-50

**current defaults**: with 20bit scalar blinding and with fixed window exp:
21-scatterplot-40-50

with 20bit scalar blinding and with Montgomery ladder:
21-scatterplot-40-50

with m_order.bits()/2 bit scalar blinding and with fixed window exp:
21-scatterplot-40-50

with m_order.bits()/2 bit scalar blinding and with Montgomery ladder:
21-scatterplot-40-50

with m_order.bits() bit scalar blinding and with fixed window exp:
21-scatterplot-40-50

with m_order.bits() bit scalar blinding and with Montgomery ladder:
21-scatterplot-40-50

It can be seen, that a longer and preferably dynamic blinding factor as proposed in http://csrc.nist.gov/groups/ST/ecc-workshop-2015/papers/session6-schindler-werner.pdf or the usage of the randomized Montgomery ladder algorithm prevents the timing leak. I think one of these countermeasures should be enabled by default. I know that the randomized Montgomery ladder algorithm was once used by default. What was the reason for replacing it with the fixed window exp. algorithm?

Please tell me what you think and which countermeasure you prefer.


randombit:
Thanks, great test and analysis. Any measurable timing differences for these operations must be considered potentially exploitable, given ECDSA's fragility. The difference between the top and bottom of the current default config seems to be about 60000 clock cycles, if I'm reading it correctly.

One question, the second graph refers to "without scalar blinding and with Montgomery ladder:" is that the Montgomery ladder is implemented by PointGFp::operator*, or the randomized algorithm in Blinded_Point_Multiply? Considering it is randomized I would expect that algorithm to not have the completely flat timing results you see.

For countermeasures: I think exposing the blinding bitsize to the user via build.h is not helpful: likely very few users edit this file and none of them understand what it means. The build.h macro should be removed and Blinded_Point_Multiply should choose whatever value best practice dictates. Based on your results, a scalar blinding of half the order size seems sufficient, and the performance is not so impacted as using a full order-sized blinding.

> I know that the randomized Montgomery ladder algorithm was once used by default. What was the reason for replacing it with the fixed window exp. algorithm?

I'm not sure this is true. The reason this algorithm is not currently (and to my recollection never has been) the default is because AFAIK the referenced paper has not been extensively studied. And the nature of the algorithm, which performs a randomized walk through the path of possible addition chains, is (IMO) difficult to be 100% confident is fully tested. Also as you can see in the graphs, the randomized Montgomery is quite a bit slower than the table lookup code. But now OSS-Fuzz is running many millions of randomized ECC tests every day for us, so switching is easier to feel confident about.

So the final response, IMO is remove `BOTAN_POINTGFP_BLINDED_MULTIPLY_USE_MONTGOMERY_LADDER`, `BOTAN_POINTGFP_SCALAR_BLINDING_BITS` (and also `BOTAN_POINTGFP_RANDOMIZE_BLINDING_BITS`) toggles from build.h, fix the scalar blinding to half the order bits, and remove the fixed window code (achieving the goal of "converge on a single side channel silent algorithm" mentioned in the side channel document).

Also these timing tests should all be integrated into the main test suite so they can run automatically, and so nothing bitrots. I'll handle this.

neverhub:
Thanks for the quick reply and your opinion :).
"without scalar blinding and with Montgomery ladder" means that the randomized Montgomery ladder was used. Therefore `BOTAN_POINTGFP_BLINDED_MULTIPLY_USE_MONTGOMERY_LADDER` was set.

I just noticed that the k_2 used in this PR is too long. I forgot to commit this change. It would be great, if you could shorten k_2 in `ecdsa.vec` by exactly 1 byte, so that the results are reproducible.

I will create the countermeasure PR very soon.

randombit:
> "without scalar blinding and with Montgomery ladder" means that the randomized Montgomery ladder was used. Therefore BOTAN_POINTGFP_BLINDED_MULTIPLY_USE_MONTGOMERY_LADDER was set.

OK thanks for the clarification.

> I just noticed that the k_2 used in this PR is too long.

Fixed in b521e251b

> I will create the countermeasure PR very soon.

Looking forward to it, thanks.

PR #881, opened 2017-02-17
Adjust number of Miller-Rabin iterations in dsa_gen and RSA keycheck

neverhub:
FIPS-186 requires 64 M-R iterations for p and q when generating DSA parameters (confer tables in appendix C). Passing a prob. of 126 results in 64 iterations.

In the _RSA_PrivateKey::check_key()_ i think we should perform 65 iterations for p and q (if strong is true), as we do during the sampling of RSA primes. This corresponds to a passed prob. of  128.


randombit:
LGTM, thanks

PR #883, opened 2017-02-18
CLI OCSP: fix expected OK return code

nunojpg:


randombit:
Thanks for the fix. These two enums used to have the same value.

PR #884, opened 2017-02-18
X509_Certificate internals refactor

randombit:
The final goal is to remove DataStore usage entirely, without impacting API or any applications.

Tests mostly pass, but one problem with custom extensions test that I don't understand yet.


randombit:
Figured out the problem with custom extension encoding. Still a lot of refactoring to do.

randombit:
I think `X509_Certificate` is done now. Still need some refactoring in PKCS10 and CRLs but it's pretty straightforward now that I've figured out what to do.

This doesn't eliminate `Data_Store` entirely, still used for extension handling. I don't think there's a realistic way to remove it there without breaking applications that use `subject_info` and `issuer_info` to get at extension data. That can be revisited in 3.x

randombit:
Now complete modulo rebasing to clean up the commit history, and updating docs.

At this point `Data_Store` and `Certificate_Extension::contents_to` are only used to support getting extension info via the `X509_Certificate::subject_info` and `issuer_info` calls. Getting extension data that way is still supported since previously applications had no other way of getting it, but hopefully we can remove that eventually.

The use of `shared_ptr<X509_Certificate>` is kind of a waste now since the X509_Certificate is internally a `shared_ptr` to the decoded data. Maybe parallel APIs could be created for such cases, but I'm not going to take that on here.

@neusdan Paging you to this PR since I seem to recall you were having problems with certificates, though I can't find the issue now. Oh, I think it was that if there was an unusual OID in the extended key usage, it would get filtered out in the return value of `ex_constraints`. That should be resolved here, along with a number of other weird corner cases. Also it's a major overhaul of the X509 internals and probably good to have eyes on it, if you have time for a review.

randombit:
Rebased and cleaned up the commit history, just needs documentation at this point.

randombit:
Ready for launch. Will leave this open for a day or so in case anyone want to review.

PR #885, opened 2017-02-18
[Request for comments] Change security_strength of HMAC_DRBG

securitykernel:
Looking into SP808-90A, it mentions that the max security strength is the security strength
of the hash function for pre-image resistance. The current implementation however always only
uses half of the output length, therefore only providing half of the security strength possible.
However, SP800-90A supports only up to 256 bit security strength, so we fix it to this upper limit.

In a second change, add_entropy() now also resets the reseed counter if enough entropy input provided.

### DIFF
--- a/src/lib/rng/hmac_drbg/hmac_drbg.cpp
+++ b/src/lib/rng/hmac_drbg/hmac_drbg.cpp
@@ -158,12 +158,18 @@ void HMAC_DRBG::update(const uint8_t input[], size_t input_len)
 void HMAC_DRBG::add_entropy(const uint8_t input[], size_t input_len)
    {
    update(input, input_len);
+
+   if(8*input_len >= security_level())
+      {
+      m_reseed_counter = 1;
+      }
    }
 
 size_t HMAC_DRBG::security_level() const
    {
-   // sqrt of hash size
-   return m_mac->output_length() * 8 / 2;
+   // security strength of the hash function
+   // for pre-image resistance (see NIST SP800-57),
+   // but NIST SP800-90A only supports up to 256 bits
+   return std::min(m_mac->output_length(), size_t(32)) * 8;
    }
-
 }
--- a/src/lib/rng/hmac_drbg/hmac_drbg.cpp
+++ b/src/lib/rng/hmac_drbg/hmac_drbg.cpp
@@ -168,8 +168,17 @@ void HMAC_DRBG::add_entropy(const uint8_t input[], size_t input_len)
 size_t HMAC_DRBG::security_level() const
    {
    // security strength of the hash function
-   // for pre-image resistance (see NIST SP800-57),
-   // but NIST SP800-90A only supports up to 256 bits
-   return std::min(m_mac->output_length(), size_t(32)) * 8;
+   // for pre-image resistance (see NIST SP 800-57)
+   // SHA-160: 128 bits, SHA-224, SHA-512/224: 192 bits,
+   // SHA-256, SHA-512/256, SHA-384, SHA-512: >= 256 bits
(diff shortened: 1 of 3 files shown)


randombit:
The intuition leading to the security_level code, was my expectation there exists a generic collision attack on the HMAC that breaks HMAC_RNG without requiring a hash preimage, and such an attack would (might) run in sqrt of the output space. I will review what the SP says about this.

This change means instead of SHA-256: 128, SHA-384: 192, SHA-512: 256 the function always returns 256 for all SHA-2 variants, correct? This seems rather less helpful to the user.

One alternative that might work: leave `security_level` alone, but always require a full 256 bits of input to reset the reseed counter since that is the maximum security level the SP supports, regardless of the PRF. What do you think?

The change to only reset the reseed counter on a 'full input' looks good.

cordney:
Sorry for taking so long to answer, I wanted to get some feedback from BSI, which I finally got. Essentially it confirms that SP 800-90A only supports the four security strengths 112 bits, 128 bits, 192 bits and 256 bits. So there really is an upper limit of what the HMAC_DRBG can provide as a security strength. The matching of hash function used in HMAC to security strength of the DRBG is listed in Table 3 of [SP 800-57  Pt. 1](http://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-57pt1r4.pdf). With HMAC(SHA-1), HMAC_DRBG would support up to 128 bits, with HMAC(SHA-224) and HMAC(SHA-512/224) it would support up to 192 bits and with HMAC(SHA-256), HMAC(SHA-512/256), HMAC(SHA-384), HMAC(SHA-512) and HMAC(SHA3-512) it supports up to 256 bits.

We could add a `virtual MessageAuthenticationCode::security_level()` which for HMAC would return something like `HashFunction::pre_image_resistance()`. `HMAC_DRBG::security_level()` would then return `std::min(MessageAuthenticationCode::security_level(), size_t(32)) * 8`. That would require some work, but I can't think of any other, cheaper solution. What do you think?

randombit:
What you propose in previous comment seems fine but would be a lot of work relative to the gain, I doubt many applications are carefully consulting `security_level` and for those that are, simply matching the values from the NIST SP is what they will expect. I'm ok with this change as is, is this ready to merge?

securitykernel:
Not quite yet, we would need to follow column 2 the following Table 3 from NIST SP 800-57:

Security Strength | Digital Signatures and hash-only applications | HMAC, Key Derivation Functions, Random Number Generation
------------ | ------------- | -------------
≤ 80 | SHA-1 | -
112 | SHA-224, SHA-512/224, SHA3-224 | -
128 | SHA-256, SHA-512/256, SHA3-256 | SHA-1
192 | SHA-384, SHA3-384 | SHA-224, SHA-512/224
≥ 256 | SHA-512, SHA3-512 | SHA-256, SHA-512/256, SHA-384, SHA-512, SHA3-512

So for SHA-1, security level must return **128**, for SHA-224 and SHA-512/224 **192** and for SHA-256+ **256**. So what do you think about something like:

```cpp
	if ( m_mac->output_length() <= 32 )
		return (m_mac->output_length() - 4) * 8;
	else
		return 32 * 8;
```

Note to self: We should have probably limited the PRF to SHA-1 and SHA-2 on the API when we did break it prior to 2.0. :/

randombit:
I don't really understand the logic behind that table, but so it goes. The version you suggest in last comment is ok. Can you make the change, including a comment in the source  explaining the reasoning behind it, and also (if missing) add tests for the return value of `HMAC_DRBG::security_level` with the various relevant hashes.

securitykernel:
Add the code and tests, additionally rebased.

randombit:
Thanks

Issue #838, opened 2017-01-12
FFI docs are incomplete


securitykernel:
The FFI handbook page is incomplete. It does not list all functions currently available and for the functions it lists, there is little to no usage documentation. Some example code would also certainly be helpful, for more examples we could link to the ffi tests.

randombit:
An additional example of usage is the Python ctypes wrapper which calls FFI API.

randombit:
FFI docs are in a much better state these days, closing this. Documentation PRs always welcome though! :)

PR #886, opened 2017-02-19
Improve ffi docs

cordney:
This is the first step to fixing #838. I realized this is much more than just fixing the handbook, but also adding doxygen docs for most of the functions and I would also like to see small code examples in the ffi handbook. I will gradually do these things over time, for now this PR adds:

* List all public ffi types and functions on the ffi handbook page
* Add doxygen API docs for versioning, hash, mac, rng, kdf, pbkdf, bcrypt
* Add handbook descriptions for versioning, hash, mac, rng, kdf, pbkdf, bcrypt

I discovered `int botan_hash_name(botan_hash_t hash, char* name, size_t name_len)` which has a declaration but no definition. Shall we remove it?

### DIFF
--- a/doc/manual/ffi.rst
+++ b/doc/manual/ffi.rst
@@ -7,7 +7,8 @@ FFI Interface
 Botan's ffi module provides a C API intended to be easily usable with
 other language's foreign function interface (FFI) libraries. For
 instance the Python module using the FFI interface needs only the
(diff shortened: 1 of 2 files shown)


randombit:
Thanks. For `botan_hash_name` I would vote remove the declaration since ffi does not offer a similar API for any other type.

Issue #888, opened 2017-02-20
RSA verification test occasionally fails with bad_alloc


randombit:
Seen on CI, reproduced by running the test locally a few thousand times until a failure occurred.

```
$ ./botan-test --drbg-seed=14A5094D2F95C15F rsa_verify
Testing Botan 2.0.1 (unreleased, revision git:63a19e0126fbf93af72baeef99f5b572020008ed, distribution unspecified)
Starting tests rng:HMAC_DRBG with seed '14A5094D2F95C15F'
RSA/EMSA4(RIPEMD-160) signature verification ran 28 tests in 2.94 msec all ok
RSA/EMSA4(SHA-1) signature verification ran 98 tests in 9.99 msec all ok
RSA/EMSA4(SHA-224) signature verification ran 28 tests in 3.07 msec all ok
RSA/EMSA4(SHA-256) signature verification ran 28 tests in 3.06 msec all ok
RSA/EMSA_PKCS1(SHA-256) signature verification ran 7 tests in 0.53 msec all ok
RSA/ISO_9796_DS2(RIPEMD-160,imp) signature verification ran 7 tests in 0.44 msec all ok
RSA/ISO_9796_DS3(RIPEMD-160,imp) signature verification ran 7 tests in 0.42 msec all ok
RSA/ISO_9796_DS3(SHA-1,exp) signature verification ran 7 tests in 0.42 msec all ok
RSA/Raw signature verification ran 7 tests in 4.24 msec all ok
pubkey/rsa_verify.vec ran 1 tests 1 FAILED
Failure 1: test 30 failed with exception 'std::bad_alloc'
Tests complete ran 218 tests in 26.05 msec 1 tests failed
Error: Test error Test suite failure
```

Have not analyzed it yet.

randombit:
Fixed in #889 not seen since

PR #889, opened 2017-02-20
Add missing length check to ISO 9796 unpadding operation

randombit:
Fixes crash mentioned in GH #888

@neverhub Does this change seem right to you? I am not sure if the early exit I added here causes an exploitable timing channel.


neverhub:
Thanks for pointing this out. I created a PR under https://github.com/randombit/botan/pull/891. Nevertheless i am not sure if constant time is required here.

randombit:
Resolved by #891

PR #893, opened 2017-02-22
converge on a single side channel silent ec blinded multiply algorithm

neverhub:
related : https://github.com/randombit/botan/pull/880

Montgomery ladder with order.bits()/2 bit scalar blinding and point
randomization is now used by default. The fixed window algorithm has been removed.

I kept the `BOTAN_POINTGFP_RANDOMIZE_BLINDING_BITS`, because i am not sure about the sufficient bit length here (and it had no impact on timing). In addition, i removed the`BOTAN_POINTGFP_SCALAR_BLINDING_BITS` and `BOTAN_POINTGFP_BLINDED_MULTIPLY_USE_MONTGOMERY_LADDER` switches.

The new `BOTAN_POINTGFP_USE_SCALAR_BLINDING` switch was introduced to optionally disable scalar blinding, if desired.

Just to be sure i performed another short timing test (as in https://github.com/randombit/botan/pull/880):
21-scatterplot-40-50

### DIFF
--- a/src/build-data/buildh.in
+++ b/src/build-data/buildh.in
@@ -80,21 +80,14 @@
 #define BOTAN_USE_VOLATILE_MEMSET_FOR_ZERO 1
 
 /*
-* If enabled the ECC implementation will use Montgomery ladder
-* instead of a fixed window implementation.
+* If enabled the ECC implementation will use scalar blinding with order.bits()/2
+* bit long masks.
 */
-#define BOTAN_POINTGFP_BLINDED_MULTIPLY_USE_MONTGOMERY_LADDER 0
-
-/*
-* Set number of bits used to generate mask for blinding the scalar of
-* a point multiplication. Set to zero to disable this side-channel
-* countermeasure.
-*/
-#define BOTAN_POINTGFP_SCALAR_BLINDING_BITS 20
+#define BOTAN_POINTGFP_USE_SCALAR_BLINDING 1
 
 /*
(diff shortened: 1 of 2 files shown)


randombit:
LGTM. There is a fairly large hit to performance with these changes but by moving to a single (secure) algorithm there is at least one code path to focus all future optimization work.

The measurement work you are doing here is really essential for guiding decisions like this, thanks again.

randombit:
Currently `BOTAN_POINTGFP_RANDOMIZE_BLINDING_BITS` defaults to 80 bits but if as you say it had no impact on timing, should we just use again `order_bits/2`? That it going to be strictly more conservative since the smallest supported curves are 160 bits.

neverhub:
Unfortunately i was unable to find any documents recommending the size of  `BOTAN_POINTGFP_RANDOMIZE_BLINDING_BITS`. Do you know if there are any papers or guidelines on this? I think it would be nice, if we could reference a publication here to justify this change.

Apart from that i think it is a good idea to make this size dynamic too :).

randombit:
I am not aware of any guidance on this issue. I checked Coron's original paper (which suggested the 20 bits used for scalar blinding) but it does not suggest anything about how large a random value to use when blinding projective coordinates.

This PR is good as is I think. `BOTAN_POINTGFP_RANDOMIZE_BLINDING_BITS` can be dealt with in a later PR.

PR #894, opened 2017-02-23
Merge timing analysis tests into command line util

randombit:
Having it installed as part of the `botan` binary makes it easy for end-users to easily test on their platform/compiler, and ensures nothing bitrots.

This doesn't touch the original version in `extra_tests` for now, but ideally `botan timing_test` will be able to replace it. @neverhub @jurajsomorovsky would love your take on this PR since you have been main users/maintainers of these tests so far.

One notable difference is in secret handling in the output for Mona: if there are > 2 inputs, `timing_test` (this PR) labels them as secrets 1,2,3... vs current `extra_tests` which buckets them into just 2 secrets. The current behavior seems wrong to me, but maybe I am misunderstanding what Mona expects here.

### DIFF
--- a/src/cli/timing_tests.cpp
+++ b/src/cli/timing_tests.cpp
@@ -1,0 +1,454 @@
+/*
+* Timing Analysis Tests
+*
+* These tests are not for performance, but verifying that two inputs are not handled
+* in a way that is vulnerable to simple timing attacks.
+*
+* Produces output which can be analyzed with the Mona reporting library
+* git clone https://github.com/seecurity/mona-timing-report.git
+*
+* (C) 2016 Juraj Somorovsky - juraj.somorovsky@hackmanit.de
+* (C) 2017 Neverhub
+* (C) 2017 Jack Lloyd
+*
+* Botan is released under the Simplified BSD License (see license.txt)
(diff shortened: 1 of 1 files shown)


jurajsomorovsky:
Thanks, this looks very good.

Regarding the issue with >2 secrets: We evaluated it with @neverhub and found out that Mona timing reporter indeed had a small bug. @neverhub created a patch and a pull request, which fixes the issue and makes the usage of more secrets possible: https://github.com/seecurity/mona-timing-report/pull/2

randombit:
Thanks for checking out the Mona issue. I looked and @neverhub's PR has been just now merged to their master.

I plan to merge this PR (after a rebase and commit cleanup). @jurajsomorovsky in your view can this fully replace the version in `extra_tests`? I'd like to only have one timing framework and afaik there are no missing features in the CLI version.

I do want to write a document that explains how to run `botan timing_test`, document the output format, and outline how to use the Mona reporter.

jurajsomorovsky:
Yes, you can safely remove the timing test implementation from extra_tests. Here, you only need the test vectors and some instructions how to run Mona timing report library.

PR #895, opened 2017-02-25
Remove verify_end() chained immediatly before end_cons() (close #890)

### LINKED ISSUE #890: BER_Decoder::verify_end() before BER_Decoder::end_cons()
In some code verify_end() is called before end_const(), eg:

https://github.com/randombit/botan/blob/master/src/lib/x509/x509_obj.cpp#L108

Since end_const would throw anyway if there is data left, what is the purpose of verify_end?

Thanks!

nunojpg:
BER_Decoder::end_cons() allready assures the verify_end()
function, so it is redundant.


randombit:
LGTM. Thank you for the patch!

PR #897, opened 2017-02-26
Add generic memory type value BER decoder

nunojpg:
This adds the method:

     template <typename T>
        BER_Decoder &get_next_value(T &out,
                                    ASN1_Tag type_tag,
                                    ASN1_Tag class_tag = CONTEXT_SPECIFIC)

Which essentially does the same as

      BER_Decoder& decode(std::vector<uint8_t>& v,
                          OCTET_STRING,
                          ASN1_Tag type_tag,
                          ASN1_Tag class_tag = CONTEXT_SPECIFIC);

While allowing to use any standard layout type, like uint8_t[size_t] or std::array<uint8_t,size_t>, and at the same time asserting the size is the expected one.

Also in some parts of the code that always process fixed length arrays (let's say a sha256 which is always 32 bytes), I wonder if beside std::vector or secure_vector, a fixed size array shouldn't be used...

### DIFF
--- a/src/lib/asn1/ber_dec.h
+++ b/src/lib/asn1/ber_dec.h
@@ -34,6 +34,38 @@ class BOTAN_DLL BER_Decoder
 
       BER_Decoder& get_next(BER_Object& ber);
 
+      /**
+      * Get next object and copy value to POD type
+      * Asserts value length is equal to POD type sizeof.
+      * Asserts Type tag and optional Class tag according to parameters.
+      * Copy value to POD type (struct, union, C-style array, std::array, etc.).
+      * @param out POD type reference where to copy object value
+      * @param type_tag ASN1_Tag enum to assert type on object read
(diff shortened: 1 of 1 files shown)


randombit:
This looks good. Please add a Doxygen comment describing the usage.

nunojpg:
I've pushed doxygen documentation and fix this is only valid for POD and not for Standard Layout.

Standard Layout would allow for a std::vector, for example, which would not be a smart idea. This is only useful for type struct, union, C-style array, std::array, etc.

Doxygen documentation for BER_Decoder is not ideal, this really needs a table explaining all the method variants. I can try to make a draft for that, since wording would have to be reviewed to be as less confusing as possible according to BER specification and current API.

randombit:
@nunojpg The PR is good, merging.

As to your question about `decode_optional` and  `decode_optional_implicit` I believe the current behavior is correct because the two cases in BER are encoded differently. But it would be nice if there was some comment explaining any of it. :/

PR #898, opened 2017-02-26
Make pylint happier and fix bug in configure

webmaster128:
This fixes 30 pylint warnings and fixes a (minor) variable bug found by pylint.


randombit:
LGTM, thanks! `InfoFile` class is a good addition here.

Issue #896, opened 2017-02-25
FFI test error


nunojpg:
While running botan-test I had a FFI error of the kind "botan_privkey_export_encrypted".

Then I repeated the test and it passed.

Unfortunately I lost the console output and so don't have the seed for this particular test that failed.

Is there a simple way with botan-test to run this test only in a loop until it (hopefully) fails, to increase the chance to find the bug?

Thanks!

randombit:
If you run `botan-test ffi` it will just run the FFI tests, normally if I am trying to replicate an occasional failure I do something like

```
for i in $(seq 1 1000)
do
  ./botan-test ffi | tee -a test_output.log
done
```

and scan the log for failures after. I'm running this now, the FFI tests seem to take 6-7 seconds to complete on my system so it may take a while to replicate.

An option for `botan-test` telling it to run certain tests N times and fail on the first error would be useful for tracking down problems like this.

nunojpg:
Why doesn't it just return a error exit status if there is any failure in any test it run? that way the shell script becomes trivial.

randombit:
Good point, actually `main` in `botan-test` already returns the number of tests that failed so an early exit from the loop is easy.

nunojpg:
I will also run the test, but close this issue for now as it is not (yet) reproducible.

Thanks.

nunojpg:
6 minutes after...


Testing Botan 2.0.1 (unreleased, revision git:2142cf8de355a2c442922e02696ee8e26449f5e6, distribution unspecified)
Starting tests rng:HMAC_DRBG with seed '14A6D65E20AE03BB'
FFI ran 253 tests all ok
Tests complete ran 253 tests in 8.81 sec all tests ok
Testing Botan 2.0.1 (unreleased, revision git:2142cf8de355a2c442922e02696ee8e26449f5e6, distribution unspecified)
Starting tests rng:HMAC_DRBG with seed '14A6D6602E2000D8'
FFI ran 253 tests 1 FAILED
Failure 1: FFI botan_privkey_export_encrypted unexpectedly failed with error code -10
Note 1: FFI PBKDF timed 10 ms 10000 iterations 100 ms 20000 iterations
Tests complete ran 253 tests in 10.13 sec 1 tests failed
Error: Test error Test suite failure

randombit:
I am unable to replicate with either seed, using a clean rebuild of the same commit :(

[...]

PR #899, opened 2017-02-26
Add ability to specify iterations when encrypting a private key

randombit:
Fixes GH #896 among other related issues.


randombit:
Realized since this PR extends FFI interface we need a new version code there, and so probably some changes to `botan.py` so that it can handle either FFI version correctly.

PR #900, opened 2017-02-27
X509 DN hash functions

nunojpg:
Hi,

This implements functionality similar to openssl x509 issuer_hash and subject_hash.

This is usable, AFAIK for certificate hash filename lookup, and OCSP scenarios where we only have the issuer subject name hash, because the issuer pub key is not available at that moment.

The hash algorithm used for Openssl is a little bit complex and I don't really know the rational behind it, I imagine it is to collapse several encoding variants of the same certificate so they all hash to the same. [1]

I used instead a simple SHA-256 of the raw name.

Things that need absolutely review at this PR:

1 - Is my assumption for Certificate_Store::find_crl_for, that it is not implemented, correct?
2 - Do you think any collapsing/re-encoding like Openssl does it important?
3 - On find_cert_by_pubkey_sha1 we use SHA1. For this new methods I used sha256, while for me sha1 would be OK as well. Should we keep consistency on the API for non critical uses of SHA-1 or for any new API just default to SHA2?

Thanks!

[1] http://stackoverflow.com/questions/34095440/hash-algorithm-for-certificate-crl-directory

### DIFF
--- a/src/lib/x509/certstor.cpp
+++ b/src/lib/x509/certstor.cpp
@@ -13,27 +13,23 @@ namespace Botan {
 
 std::shared_ptr<const X509_CRL> Certificate_Store::find_crl_for(const X509_Certificate&) const
    {
-   return std::shared_ptr<const X509_CRL>();
+   return {};
    }
 
 void Certificate_Store_In_Memory::add_certificate(const X509_Certificate& cert)
    {
-   for(size_t i = 0; i != m_certs.size(); ++i)
-      {
-      if(*m_certs[i] == cert)
+   for(const auto& c : m_certs)
+      if(*c == cert)
          return;
-      }
 
    m_certs.push_back(std::make_shared<const X509_Certificate>(cert));
    }
 
 void Certificate_Store_In_Memory::add_certificate(std::shared_ptr<const X509_Certificate> cert)
    {
-   for(size_t i = 0; i != m_certs.size(); ++i)
-      {
-      if(*m_certs[i] == *cert)
+   for(const auto& c : m_certs)
+      if(*c == *cert)
          return;
-      }
 
    m_certs.push_back(cert);
    }
@@ -41,8 +37,8 @@ void Certificate_Store_In_Memory::add_certificate(std::shared_ptr<const X509_Cer
 std::vector<X509_DN> Certificate_Store_In_Memory::all_subjects() const
    {
    std::vector<X509_DN> subjects;
-   for(size_t i = 0; i != m_certs.size(); ++i)
-      subjects.push_back(m_certs[i]->subject_dn());
+   for(const auto& cert : m_certs)
+      subjects.push_back(cert->subject_dn());
    return subjects;
    }
 
@@ -50,22 +46,22 @@ std::shared_ptr<const X509_Certificate>
 Certificate_Store_In_Memory::find_cert(const X509_DN& subject_dn,
                                        const std::vector<uint8_t>& key_id) const
    {
-   for(size_t i = 0; i != m_certs.size(); ++i)
+   for(const auto& cert : m_certs)
       {
       // Only compare key ids if set in both call and in the cert
       if(key_id.size())
          {
-         std::vector<uint8_t> skid = m_certs[i]->subject_key_id();
+         std::vector<uint8_t> skid = cert->subject_key_id();
 
          if(skid.size() && skid != key_id) // no match
             continue;
          }
 
-      if(m_certs[i]->subject_dn() == subject_dn)
-         return m_certs[i];
+      if(cert->subject_dn() == subject_dn)
+         return cert;
       }
 
-   return std::shared_ptr<const X509_Certificate>();
+   return nullptr;
    }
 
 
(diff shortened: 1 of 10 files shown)


randombit:
All the loop cleanups are great, thank you!

> 1 - Is my assumption for Certificate_Store::find_crl_for, that it is not implemented, correct?

It is not implemented, but throwing here is not right. It is just that `find_crl_for` is not pure virtual, and has a kind of default no-op. But `Certificate_Store` is a pure virtual interface class, and has no state to consult - there is nothing that could even hypothetically be written there. This change would cause subclasses of `Certificate_Store` that do not implement CRLs to throw during path validation, which does not seem like the right behavior.

It would be better for `find_crl_for` to be pure virtual instead of the default no-op nonsense, and force any subclasses that do not implement CRLs to explicitly return nothing themselves or throw themselves. Again, API compat. Blerg.

> 2 - Do you think any collapsing/re-encoding like Openssl does it important?

It is probably important to someone somewhere, but a plain hash seems a lot simpler to me.

> 3 - On find_cert_by_pubkey_sha1 we use SHA1. For this new methods I used sha256, while for me sha1 would be OK as well. Should we keep consistency on the API for non critical uses of SHA-1 or for any new API just default to SHA2?

SHA2 for anything new please.

In `find_cert_by_raw_subject_dn_sha256` you should create a single `HashFunction` and process `cert.raw_subject_dn()` in a loop, which will be a lot faster (avoiding the whole lookup process, memory allocations, etc).

nunojpg:
@randombit removed the Not_Implemented throw, refactored two additional lines, reuse the hash at find_cert_by_raw_subject_dn_sha256, and also refactored find_cert_by_pubkey_sha1 in the same way.

I'm using the hash lookup and unique_ptr thing instead of creating a  Botan::SHA_256 directly. I'm doing that just because that's how it appears to be done elsewhere in the code, but I really don't know the reason for this. It seems it could be usefull if we were passing a string with the desired hash method (while that could also be done at compile time with templates).

So, is this std::unique_ptr<HashFunction> hash(HashFunction::create("SHA-256")); (vs SHA_256 hash;) the recommend way? (and why?)

randombit:
> So, is this std::unique_ptr hash(HashFunction::create("SHA-256")); (vs SHA_256 hash;) the recommend way? (and why?)

Using `create` is probably better in most cases. It allows the function to compile even if SHA-256 is unavailable at build time in the library, and it can just fail at runtime in that case. Otherwise we'd have to guard the function definition itself with `#ifdef BOTAN_HAS_SHA2_32`, which then API callers would have to know about and handle specially. If SHA-2 was essential to the overall function of `X509_Certificate` class then it would be more appropriate to make sha2 a hard build-time dependency of `x509` module.

Also using `create` allows using the OpenSSL versions which (at least for right now) are not used if the operation goes through the `SHA_256` class. That doesn't matter much in this instance, since the DN bits are going to fit in a single call to the compression function in most cases.

The distinction mattered at lot more before #623 (#477 has more background on that change), before that PR no CPU specific code would be used by `SHA_256 hash` because CPU specific code (like ARMv8 extensions, in the case of SHA2) would be in a distinct class that was returned by `create`.

> I'm a little lost navigating in the test suite. If you can point me to where you would like the tests to be done, I can try to implement it...

Probably the place for this is in `src/tests/unit_x509.cpp`, add a new freestanding function in that file and call it from `X509_Cert_Unit_Tests::run`. There are some existing test certs in `src/tests/data/certstor` that can be used.

nunojpg:
@randombit thank you so much for the explanation. I want to implement a test suite but I'll only be able to get back to this in 30 days (!). Please leave the pull request in standby or merge as you wish, and I will in any case add or create a new pull request for the test suite.

randombit:
@nunojpg Thanks for the update. As a general policy we avoid adding new code without tests, I may have some time to write up a test for this change in the next weeks but if not it will wait for your return.

nunojpg:
Maybe it should be created instead at test_certstor.cpp to avoid duplication of the certificates file loading and "Skipping due to missing filesystem access" case logic?

randombit:
Yes `test_certstor.cpp` would be much better place for these tests.

nunojpg:
I added some unit tests for:

X509_Certificate::raw_issuer_dn_sha256()
X509_Certificate::raw_subject_dn_sha256()

I can't really implement the certificate store lookup tests this month. Sorry.

Also "needed" to constantify X509_CA::sign_request.

Please let me know how this looks. Thanks!

nunojpg:
@randombit tests are now also included and ready for your "final" review.

Maybe @flanfly would like to review commit "Refactor test_certstor.cpp to prepare for adding tests". I made a lot of changes to this file, some to prepare for adding tests not specific to Certstor SQL, others just to simplify it, but that might not be unanimous. For example I removed the use of std::pair and used instead structs to name the fields, as this makes code review much more simpler.

The rest I believe are quite trivial changes.

Thanks!

ksanderon:
@nunojpg: Unfortunately with clang-format it's nearly impossible :/ However - I will reformat whole botan code base(as started in #963) with astyle(hopefully soon) and provide configuration file for it in repo. If you would like to have automatic code formating which more(or less) fit to currently used style look at that provided in: #967

nunojpg:
@randombit had to make some #if defined changes to work also with --minimized-build, but passing CI now.

randombit:
@nunojpg Thanks! Will review shortly, would be great to have this for 2.1

PR #901, opened 2017-02-27
Allow OCSP requests without the full subject certificate

nunojpg:
A OCSP request doesn't need the full subject certificate.

This extends the API to require instead of the subject certificate:

* OCSP::Request: subject serial.
* OCSP::online_check: subject serial AND ocsp responder url.

API breaking change:

* removal of OCSP::Request::subject() as OCSP::Request doesn't need to hold
  the certificate, but only the serial.

### DIFF
--- a/doc/manual/x509.rst
+++ b/doc/manual/x509.rst
@@ -698,9 +698,14 @@ the subject's issuing certificate.
 .. cpp:class:: OCSP::Request
 
  .. cpp:function:: OCSP::Request(const X509_Certificate& issuer_cert, \
-                                 const X509_Certificate& subject_cert)
+                                 const BigInt& subject_serial)
 
       Create a new OCSP request
+
+ .. cpp:function:: OCSP::Request(const X509_Certificate& issuer_cert, \
+                                 const X509_Certificate& subject_cert)
+
+      Variant of the above, using serial number from ``subject_cert``.
 
  .. cpp:function:: std::vector<byte> BER_encode() const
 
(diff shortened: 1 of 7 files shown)


randombit:
Thank you for the patch. Looks really good on a quick first review. Can you please update `doc/manual/x509.rst` with the added constructor signature.

I am not completely sure about how to handle removal of `OCSP::Request::subject`, in the context of maintaining a stable 2.0 API. This function is not documented (or used/tested within the library), so arguably it is not part of the "public API" and can be safely removed. But, it is still accessible to applications... An alternative which is 'build-time compatible' is to export a deprecated function `OCSP::Request::subject` that always throws `Not_Implemented`. This alternative seems acceptable to me, because there is (unless explicitly documented) never a guarantee of any particular API function always succeeding.

For support reasons incrementing the major version at this time does not work - already we have publicly committed to supporting 2.x API through 2020, and adding another 3.x branch to release from would both confuse matters for end-users and add greatly to our maintenance burden.

@neusdan @cordney @webmaster128 (or anyone else who wants to chime in) what are your thoughts on this issue - does 'public API' mean just the APIs documented in the manual, or all BOTAN_DLL exported functions, or something else?

@nunojpg I'm sure this patch will make it to master in time for 2.1, but having only just recently adopted SemVer there are some issues that we have yet to settle.

webmaster128:
I don't really care for API stability details as long as incompitible changes cause hard compile errors and do not silently make encryption weak.

neusdan:
I think public API is everything that is accessible. However, marking the function as deprecated, throwing  `Not_Implemented`, documenting it in the release notes and increasing only the minor version seems ok to me because this will only affect a few users, if any. 

[Here](https://github.com/mojombo/semver/issues/311) is also some interesting discussion if behavioral changes should result in a major version increase.

nunojpg:
I'll provide documentation when the change is stable.

I've updated the branch with subject() throwing. This required a little hack:

      const X509_Certificate& subject() const {
         throw Not_Implemented("Method have been deprecated");
         return m_issuer; //while this API is not removed completely, we return
                          //a random X509_Certificate object on hand to avoid
                          //compiler warnings.
      }

randombit:
@nunojpg what compiler required the explicit return statement? I would have guessed any modern compiler would understand an unconditional throw as exiting the function without a return value being required.

nunojpg:
Sorry, you are right. I can remove the return completely without any warning.

randombit:
Looks good thank you!

Issue #892, opened 2017-02-22
insufficient checks when parsing RFC 3279 ECKeys


neverhub:
RFC 3279 allows ECKeys with user defined parameters. Therefore the EC domain parameters can be specified by the public key. Botan does not check the parsed domain parameters when parsing a public key. EC domains parsed this way, might have cofactor != 1, which potentially leads to problems like small-subgroup-attacks on ECC.

@cordney imported the test vectors from [wycheproof](https://github.com/google/wycheproof/blob/master/java/com/google/security/wycheproof/testcases/EcKeyTest.java) and extended the testsuite. I added a `EC_Group::verify_group` function, which performs the necessary domain parameter checks. In addition, the `EC_PublicKey::check_key` was expanded to handle curves with cofactor > 1, as it is not sufficient to check if the point is on the curve in this case (conf. Section 5 of [source](https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Zertifizierung/Interpretationen/AIS_46_ECCGuide_e_pdf.pdf?__blob=publicationFile)).  All changes were pushed to https://github.com/Hackmanit/botan/tree/ecc_check.

I think the checks should be executed automatically, when constructing a `EC_PublicKey` or decoding a `EC_Group`. Unfortunately this would require API changes, as we need a `RandomNumberGenerator` for `EC_Group::verify_group`.

What is your opinion on this? Are you okay with an API change or do you have a better idea to resolve the issue?

randombit:
> I think the checks should be executed automatically, when constructing a EC_PublicKey or decoding a EC_Group.

I think an incompatible API change is a non-starter here, because we have just now committed to a 4 year support lifetime with a 2.0-compatible API. Forking the development at this very early stage would cause a lot of problems down the road, and not really improve matters for end-users since if the code cannot make it into a 2.x release it effectively does not exist anyway. This type of checking-during-key-parse behavior was previously in use for all algos, and caused so many problems over the years that it was finally removed in 8d50c82b43b34. I'm rather hesitant to revisit the issue, at least right now.

Adding the missing tests and making them available via `check_key` is certainly an improvement. Also an issue, the documentation should clearly explain that simply loading the key in to memory does not verify that the key satisfies all tests, and explain `check_key`s usage.

securitykernel:
What is left to do here, only updating the docs?

PR #902, opened 2017-02-28
Extend EC_PublicKey check, add EC_Group check and corresponding tests

neverhub:
This pr adds functions, which solve https://github.com/randombit/botan/issues/892. As already referenced in https://github.com/randombit/botan/issues/892, the implementation follows Section 5 of the  BSI [recommendations](https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Zertifizierung/Interpretationen/AIS_46_ECCGuide_e_pdf.pdf?__blob=publicationFile). `openssl` implements similar checks for EC_Groups and EC_Keys. The respective functions can be found [here](https://github.com/openssl/openssl/blob/master/crypto/ec/ec_check.c#L13) and [here](https://github.com/openssl/openssl/blob/master/crypto/ec/ec_key.c#L270). 

In addition invalid ECDSA test vectors from google Wycheproof, FIPS 186-2 and FIPS 186-4 have been added to the test suite as negative tests.


randombit:
Looks good thanks!

PR #908, opened 2017-03-09
Document --with-build-dir correctly in building manual.

bluhm:
The configure option is actually called --with-build-dir not --build-dir.


randombit:
Thanks

PR #909, opened 2017-03-10
tls_client must not pass an IP address as server information

bluhm:
RFC 6066 section 3 says: Literal IPv4 and IPv6 addresses are not
permitted in "HostName".  But if a user passes an IP address to
botan tls_client as connect address, this is also used for SNI.
Some TLS server like libtls from the LibreSSL project check that a
provided hostname is a DNS name.  The TLS connection attempt from
botan is rejected with a fatal alert.

My libtls server provides this error message:
nc: tls handshake failed (handshake failed: error:140270E2:SSL routines:ACCEPT_SR_CLNT_HELLO_C:clienthello tlsext)

While botan tls_client fails with:
./botan tls_client 127.0.0.1 --port=1234
Alert: internal_error

I am not sure whether my proposed commit is in the correct layer.
Maybe a more generic fix in Botan::TLS::Client or
Botan::TLS::Server_Information would be better.

### DIFF
--- a/src/cli/tls_client.cpp
+++ b/src/cli/tls_client.cpp
@@ -25,6 +25,7 @@
 #include <sys/time.h>
 #include <sys/socket.h>
 #include <netinet/in.h>
+#include <arpa/inet.h>
 #include <netdb.h>
 #include <unistd.h>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


randombit:
Thanks for the report. It would be better to solve this in a generic way since likely other applications will use an IP here in various circumstances. In addition even if only the IP is known, that is useful information to the library (for things like saving the session to DB, or comparing with the IP value set in certificate subject alternative name). Instead just SNI extension should not be sent, if it does not seem to be a DNS name (in `msg_client_hello.cpp`).

bluhm:
As suggested, I will fix the bug in the library when the client hello is generated.

PR #910, opened 2017-03-10
TLS client must not pass an IP address as server name

bluhm:
RFC 6066 section 3 says: Literal IPv4 and IPv6 addresses are not
permitted in "HostName".  But if a user passes an IP address to
botan tls_client as connect address, this is also used for SNI.
Some TLS server like libtls from the LibreSSL project check that a
provided hostname is a DNS name.  The TLS connection attempt from
botan is rejected with a fatal alert.

Fix the issue in the botan library.  Do not send a SNI extension
in the client hello if the hostname is empty or an IPv4 or IPv6
address instead of a DNS name.

### DIFF
--- a/src/lib/tls/msg_client_hello.cpp
+++ b/src/lib/tls/msg_client_hello.cpp
@@ -16,6 +16,11 @@
 #include <botan/internal/stl_util.h>
 #include <chrono>
 
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+
 namespace Botan {
 
 namespace TLS {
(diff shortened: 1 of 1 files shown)


bluhm:
Unfortunately this does not compile on Windows.

It looks like some windows.h, winsock2.h, or ws2tcpip.h include
magic is needed.  As I have not used this operating system since
Windows 95, I cannot help there.

I recommend to use the native operating system's functions to detect
IP addresses.  Especially IPv6 has many different representations
that are hard to parse.  LibreSSL libtls also uses inet_pton(3) to
check that the other side does not send IPs and to prevent sending
IPs itself in SNI.

randombit:
> I recommend to use the native operating system's functions to detect
IP addresses. Especially IPv6 has many different representations
that are hard to parse.

Ouch. Any solution relying on using system socket libs in `src/tls` cannot be used. Botan's TLS stack is used on unikernel OS without any POSIX libs. Probably the best solution is to change the TLS command line utility as in your previous patch, and document that all applications should avoid passing IP addresses as the `hostname` parameter.

bluhm:
As suggested fix CLI and doc in pull request #942.

PR #913, opened 2017-03-13
Fix: UTCTime interpreted as GeneralizedTime

neusdan:
Example:
"200305100350Z" interpreted as "2003/05/10 03:50:00 UTC"
correct is "2020/03/05 10:03:50 UTC"

According to RFC 5280:
UTCTime values ... MUST include seconds (i.e., times are YYMMDDHHMMSSZ) -> length 13
GeneralizedTime values ...  MUST include seconds (i.e., times are YYYYMMDDHHMMSSZ) -> length 15

I think we should enforce the RFC5280 rules even if the ASN.1 rules are not that strict.


randombit:
LGTM. I'm OK with following PKIX here, especially given it removes a nasty ambiguity.

PR #915, opened 2017-03-13
Fix OS::get_processor_timestamp() tests

neusdan:
In approximately 8/10 test runs i get the following error from the os_utils tests:

```
OS::get_processor_timestamp ran 1 tests 1 FAILED
Failure 1: OS::get_processor_timestamp Processor timestamp does not duplicate produced unexpected result '0' expected '1'
```

PR #916, opened 2017-03-13
Add VS2017 builds

neusdan:
The "Visual Studio 2015" image contains VS2013 and VS2015. The "Visual Studio 2017" image contains  VS2015 and VS2017. So we have to use two different images now:

https://www.appveyor.com/docs/build-environment/


neusdan:
Ready for review. One failed job because of #915

randombit:
LGTM, thanks!

PR #918, opened 2017-03-15
Set latex papersize to a4

cordney:
I have heard various requests offline for the latex-formatted handbook to be in A4 format.

### DIFF
--- a/src/build-data/sphinx/conf.py
+++ b/src/build-data/sphinx/conf.py
@@ -176,7 +176,8 @@ htmlhelp_basename = 'botandoc'
 
 # -- Options for LaTeX output --------------------------------------------------
 
-# Various latex options, like papersize and pointsize
+# Various latex options, e.g.,
+# paper size option of the document class ('a4paper' or 'letterpaper'), default 'letterpaper'
 latex_elements = {
   'papersize': 'a4paper',
 }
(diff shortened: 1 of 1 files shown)


mouse07410:
And I need it to stay Letter (8.5 x 11 inches).

Make it configurable?

cordney:
I'd rather not like to introduce an additional `configure.py` option to just configure the papersize of the manual when built with the latex output formatter. We are just changing the default setting here, A4 is the default in almost all of the world except the US. You can still change it manually in sphinx `conf.py` whenever you create a latex doc, which I assume is not very often. I added c710fd5 to document which options can be used for papersize.

randombit:
Based on `https://llvm.org/svn/llvm-project/cfe/trunk/docs/Makefile.sphinx` it is possible to override the setting at build time (without forcing enduser to edit any configs) by passing `-D latex_paper_size=[letter|a4]` to `sphinx-build`.

I do not care overmuch which size is the default but since there are legit needs for both it probably should be easy to override by the end user. For example if there were two makefile targets `docs_pdf_a4` and `docs_pdf_letter`, or a PAPER variable such as LLVM uses. A `configure.py` argument seems over the top.

mouse07410:
> We are just changing the default setting here

And I strongly argue for *not* doing that.

> change it manually in sphinx `conf.py`

Well, why don't you do that for A4 instead...

> since there are legit needs for both it probably should be easy to override by the end user

YES!!

>  For example if there were two makefile targets `docs_pdf_a4` and `docs_pdf_letter`, or a PAPER variable such as LLVM uses

A variable `PAPER` or `PAPER_SIZE` seems to make perfect sense.

cordney:
Understand. I'll work on a new PR.

PR #923, opened 2017-03-16
Provide the handbook as a PDF download

### LINKED ISSUE #919: Provide handbook as PDF download on the website
I have heard various requests offline for the handbook also being provided in PDF format downloadable from the website, additionally to the online manual. The link could be available on the index page. Can we do that? Not sure if this could be automated.

securitykernel:
Extends the website script to include a link to the manual as a PDF file for download.
Also includes links to search and index only for html, as it does not work for latex.

Fixes #919.

### DIFF
--- a/src/scripts/website.sh
+++ b/src/scripts/website.sh
@@ -11,6 +11,13 @@ WEBSITE_SRC_DIR=./www-src
 rm -rf $WEBSITE_SRC_DIR $WEBSITE_DIR
 mkdir -p $WEBSITE_SRC_DIR
 
+# build manual as pdf for download
+sphinx-build -t website -c "$SPHINX_CONFIG" -b "latex" doc/manual latex
+cd latex
+pdflatex botan.tex
+cd ..
+
+# build online manual
 cp readme.rst $WEBSITE_SRC_DIR/index.rst
 cp -r news.rst doc/security.rst $WEBSITE_SRC_DIR
 echo -e ".. toctree::\n\n   index\n   news\n   security\n" > $WEBSITE_SRC_DIR/contents.rst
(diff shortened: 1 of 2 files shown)


src/scripts/website.sh
```diff
@@ -11,6 +11,13 @@ WEBSITE_SRC_DIR=./www-src
 rm -rf $WEBSITE_SRC_DIR $WEBSITE_DIR
 mkdir -p $WEBSITE_SRC_DIR
 
+# build manual as pdf for download
+sphinx-build -t website -c "$SPHINX_CONFIG" -b "latex" doc/manual latex
+cd latex
+pdflatex botan.tex
```
mouse07410: I don't think running `pdflatex` once is sufficient, as it won't account for index and TOC.

cordney: The generated TOC is just fine, I just double-checked. As of now, we have no index of functions in the LaTex document. I would love to have one, as we have in the online manual, but I couldn't find any docs on how to do this. If you have a solution, that would be great.

Anyway, the CI build fails here, and I figured out why. sphinx generates a warning in case the botan.pdf file is not found. I explicitly put an `only:: html and website` directive around the download role in `index.rst`, but sphinx seems to ignore it when processing the download role. From reading https://github.com/sphinx-doc/sphinx/issues/2150 it seems the `only::` directive is only applied after processing the rEST, causing the warning regardless of whether the tags are set or not. The only way out I see is `cp`ing the PDF file in `website.sh` and hardcoding a link to https://botan.randombit.net/manual/botan.pdf into `index.rst`.

randombit: I'm good with hardcoding the URL

securitykernel: Updated the PR

Issue #922, opened 2017-03-15
Errors when running tests


ruipacheco:
I've downloaded Botan 2.0.1 and configured with the following parameters: 

`python configure.py --minimized-build --enable-modules=zlib,md5,sha1,tls --single-amalgamation-file`

Upon running `./botan-test` a number of tests failed with the following message: `'Test error No usable RNG in build, and this test requires an RNG'`

At the end of the test suit I found this message:

```
RSA/EMSA_PKCS1(SHA-256) verify invalid signature ran 141 tests in 12.81 msec all ok
tls ran 1 tests 1 FAILED
Failure 1: Test error No usable RNG in build, and this test requires an RNG
alert parsing ran 9 tests in 0.08 msec all ok
cert_status parsing ran 8 tests in 0.13 msec all ok
cert_verify parsing ran 8 tests in 0.24 msec all ok
client_hello parsing ran 18 tests in 0.21 msec all ok
hello_request parsing ran 2 tests in 0.02 msec all ok
hello_verify parsing ran 5 tests in 0.04 msec all ok
hello_verify_request construction ran 1 tests all ok
new_session_ticket parsing ran 5 tests in 0.04 msec all ok
server_hello parsing ran 12 tests in 0.11 msec all ok
tls_policy ran 1 tests 1 FAILED
Failure 1: Test error No usable RNG in build, and this test requires an RNG
Date parsing ran 71 tests in 0.13 msec all ok
calendar_point::to_string ran 1 tests all ok
X509v3 Name Constraints: Invalid_DN_Name_Constraint.crt ran 1 tests all ok
X509v3 Name Constraints: Invalid_Email_Name_Constraint.crt ran 1 tests all ok
X509v3 Name Constraints: Invalid_IP_Name_Constraint.crt ran 1 tests all ok
X509v3 Name Constraints: Valid_DNS_Name_Constraint.crt ran 1 tests all ok
X509v3 Name Constraints: Valid_DN_Name_Constraint.crt ran 1 tests all ok
X509v3 Name Constraints: Valid_IP_Name_Constraint.crt ran 1 tests all ok
NIST path validation ran 71 tests in 802.10 msec all ok
X509test path validation ran 37 tests in 125.18 msec all ok
x509_unit ran 1 tests 1 FAILED
Failure 1: Test error No usable RNG in build, and this test requires an RNG
Tests complete ran 9977 tests in 4.04 sec 2269 tests failed
Error: Test error Test suite failure
```

[...]

PR #924, opened 2017-03-17
Add a note that no RNG is part of the minimized build

cordney:
Also adds a description for the --list-modules and --module-policy option. Related to #922.

@ruipacheco You think this is sufficient?


ruipacheco:
Looks great.

PR #929, opened 2017-03-20
Add ppc64le (POWER8 little endian) as supported cpu

gut:
Before:
```
$ ./configure.py
   INFO: ./configure.py invoked with options ""
   INFO: Platform: OS="Linux" machine="ppc64le" proc="ppc64le"
   INFO: Guessing target OS is linux (use --os to set)
   INFO: Guessing to use compiler gcc (use --cc to set)
  ERROR: Could not determine target CPU; set with --cpu
```

Due to:
```
$ lscpu
Architecture:          ppc64le
Byte Order:            Little Endian
CPU(s):                80
On-line CPU(s) list:   0,8,16,24,32,40,48,56,64,72
Off-line CPU(s) list:  1-7,9-15,17-23,25-31,33-39,41-47,49-55,57-63,65-71,73-79
Thread(s) per core:    1
Core(s) per socket:    5
Socket(s):             2
NUMA node(s):          2
Model:                 2.1 (pvr 004b 0201)
Model name:            POWER8E (raw), altivec supported
CPU max MHz:           3690.0000
CPU min MHz:           2061.0000
L1d cache:             64K
L1i cache:             32K
L2 cache:              512K
L3 cache:              8192K
NUMA node0 CPU(s):     0,8,16,24,32
NUMA node1 CPU(s):     40,48,56,64,72
```

After this change I could run `./botan-test` and it returned:
```
Tests complete ran 415766 tests in 106.17 sec all tests ok
```

### DIFF
--- a/src/build-data/arch/ppc64le.txt
+++ b/src/build-data/arch/ppc64le.txt
@@ -1,0 +1,22 @@
+endian little
+
+family ppc
+wordsize 64
+
+<aliases>
+powerpc64
+</aliases>
+
+<submodel_aliases>
+</submodel_aliases>
+
+<submodels>
+power8
+</submodels>
+
+<submodel_aliases>
+</submodel_aliases>
+
+<isa_extensions>
+altivec
+</isa_extensions>


gut:
BTW: I'm considering contributing with an altivec approach for SHA-2 algorithms like I did for [cryptopp](https://github.com/weidai11/cryptopp/pull/394) which normally run around 4 times faster. Would botan welcome such a contribution?
Thanks

randombit:
@gut Altivec SHA-2 would be a great addition, I would be happy to merge such a PR. One important change would be required vs the Cryptopp patch however, which is Botan does not use bare asm files ever, but only intrinsics (or inline asm when absolutely necessary) because it affords better diagnostics and optimizations and avoids any ABI problems (name mangling, parameter passing and the like).

gut:
@randombit : Nice. We have an [inline assembly version](https://github.com/PPC64/sha2-le/blob/inline_asm/sha256_compress.c) but we didn't like placing so many macros to inline those assemblies, so we ended up with a [m4 solution](https://github.com/PPC64/sha2-le/blob/master/sha256_compress_ppc.m4) which looks much nicer, gave us better control of registers and produces a [readable assembly](https://github.com/PPC64/cryptopp/blob/3f128e4667218d7fba546d25358d9b42a4b08c24/sha256_compress_ppc.s).

At first we used intrinsics but we noticed the performance was not good (as bad as plain C code). But a lot of things changed from our algorithm since then... we'll take a closer look. But that is another task (Thanks for your quick reply!)

What about this one? Is that ok for enabling ppc64le? Do you need more system info in order to confirm these parameters?

gut:
@randombit : ping:
"*What about this one? Is that ok for enabling ppc64le? Do you need more system info in order to confirm these parameters?*"

randombit:
ppc64le build already works fine *if* the user knows to use `--cpu=ppc64 --with-endian=little`. The current behavior is not friendly.

I would prefer adding logic to configure.py to handle bi-endian arches such as `ppc6el`, `mipsel`, `armbe`, ... in a generic way, translating it to the base arch plus an endian override.  I may make this change in the future. But your approach seems ok for 2.1

PR #930, opened 2017-03-21
Allow destructor to throw when building with >=C++11

Peter-Levine:
After C++11, destructors default to `noexcept(true)`.  Thus, any throw statements in destructors become calls to `std::terminate()`.  Building with CXXFLAGS="-Werror=terminate" causes build failure.  To enable the intended throw statement, the destructor must be explicitly marked `noexcept(false)`.


randombit:
It would probably be better to change this destructor to not throw an exception. Even in C++98 it is not a good idea.

Peter-Levine:
> It would probably be better to change this destructor to not throw an exception. Even in C++98 it is not a good idea.

@randombit Agreed.

randombit:
Thanks

PR #931, opened 2017-03-21
Add support for reaching IPv6 hosts in the tls_client CLI

securitykernel:
Tested against OpenSSL 1.1.0 with `openssl s_server --6`.


randombit:
Lovely thank you!

PR #933, opened 2017-03-21
Fix #917: Allow years > 2037 if 64 bit std:time_t is used

### LINKED ISSUE #917: calendar_point::to_std_timepoint() does not support years after 2037
Can we do something about this?

Since Visual Studio 2005 the size of time_t is always 64-bits. But i don't know about the others. Is this still a problem?

neusdan:
Fixes #917

Only throw on systems where 32 bit std::time_t is used.


randombit:
LGTM thanks

PR #936, opened 2017-03-23
Fix path in build variable botan_pkgconfig

webmaster128:
Current master breaks with 

```
./configure.py --amalgamation --with-build-dir hello --prefix wohoo
make -j4 -f hello/Makefile install
```

due to an extra hello/ path segment in `botan_pkgconfig`. This fixes the bug


randombit:
Thanks

PR #937, opened 2017-03-23
Use a hardcoded URL for the PDF download

securitykernel:
Somehow this commit didn't land on master, but should have been part of #923.


randombit:
Website rebuilt and pdf is up at https://botan.randombit.net/manual/botan.pdf

securitykernel:
Thank you!

PR #940, opened 2017-03-23
Add some MSVC compiler flags

neusdan:
/Oi enabled in release builds: generates intrinsic functions for appropriate function calls
https://msdn.microsoft.com/en-us/library/f99tchzc.aspx

/Os if `--optimize-for-size` is used: tells the compiler to favor optimizations for size over optimizations for speed
https://msdn.microsoft.com/en-us/library/f9534wye.aspx

### DIFF
--- a/src/build-data/cc/msvc.txt
+++ b/src/build-data/cc/msvc.txt
@@ -10,8 +10,8 @@ add_lib_option ""
 
 compile_flags "/nologo /c"
 
-optimization_flags "/O2"
-size_optimization_flags "/O1"
+optimization_flags "/O2 /Oi"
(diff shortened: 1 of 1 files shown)


neusdan:
Debug build with run time checks takes too long for Appveyor :-(

randombit:
Would it be reasonable to lump "/RTC1" under `--with-sanitizers` instead of debug mode?

neusdan:
/RTC cannot be used with compiler optimizations, so I think this does not help.

randombit:
Worthwhile to drop RTC1 flag for now and take the other two?

neusdan:
Yes, dropped the `/RTC1` flag for now.

PR #946, opened 2017-03-28
Expose BigInt API subset to C API

randombit:
Also adds RSA key constructors using integer types

Also makes the change to FFI version checking in botan.py mentioned in #899

Also fixes a bcrypt test that wanted to run with work factor 3, after #938 bcrypt requires at least wf 4 to match other implementations.


securitykernel:
Nice! We should also add the new functions to the handbook.

PR #947, opened 2017-03-29
Use getentropy(2) as random source.

bluhm:
Gather entropy from system call getentropy(2).  This is available
since in OpenBSD 5.6 and Solaris 11.3.  It can provide up to 256
bytes entropy from the kernel without blocking.  As a system call
it does not need a file descriptor and works in chroot(2) environments
without device nodes.

OpenBSD provides this system call to seed random number generators
in user land.  See http://man.openbsd.org/getentropy


randombit:
Very nice, thank you!

PR #949, opened 2017-03-29
Use entropy source for cli rng.

bluhm:
With "botan rng --entropy" the global entropy sources are used to
seed the auto seeded random number generator.

This allows to run and test the entropy sources interactively.


randombit:
Thanks

PR #955, opened 2017-03-30
Run tests for a specific provider.

bluhm:
Currently botan runs the tests for all crypto providers it can find.
Add a --provider option for botan-test to specify exactly one
provider.  This allows to see which parts of a specific implementation
have been tested.  Pass down the given provider to a specific test
class.  Start with Block_Cipher_Tests to filter out unwanted
providers, others classes will follow.

My goal is to run the botan test suite with LibreSSL.  The configure
--with-openssl feature makes this possible as LibreSSL and OpenSSL
are API compatible.  I am showing this diff early to find out whether
I am on a reasonable path.  Then all the other Tests can be made
provider aware and it seems that some OpenSSL bindings can be added.

### DIFF
--- a/src/tests/main.cpp
+++ b/src/tests/main.cpp
@@ -35,7 +35,7 @@ namespace {
 class Test_Runner : public Botan_CLI::Command
    {
    public:
-      Test_Runner() : Command("test --threads=0 --run-long-tests --run-online-tests --test-runs=1 --drbg-seed= --data-dir= --pkcs11-lib= --log-success *suites") {}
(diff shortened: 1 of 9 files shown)


randombit:
Yes, this is a good addition. Also I like your idea of using the botan tests to cross-test LibreSSL

bluhm:
The --provider option works for all text based tests now.  If nothing
is specified, all tests run.  But with a provider given on the
botan-test command line, the test run is limited to this one.

This mechanism should be expanded to all tests.  Before I do that,
I would prefer that this step gets integrated.

randombit:
Looks great, thank you

PR #956, opened 2017-03-31
Add TOC link to users manual and API ref for website

securitykernel:
Adds a link to the users manual and the API reference to the website's index.html TOC, so it is easily accessible. I regularly find myself searching for the users manual link, which is currently only in the text, and I heard this request from other users, too.

apiref


randombit:
Good idea, thanks

PR #960, opened 2017-03-31
Check name of test suite.

bluhm:
If botan-test is invoked with an invalid test name, fail with a
usage error.  This is more obvious than not running the test.


randombit:
Thanks!

PR #962, opened 2017-04-02
More pylint happyness

webmaster128:
This PR consists mainly of those two refactorings:

* Change defines in module info.txt from key/value field to dictionary, because in some cases we have more than one define
*  Strip down BuildConfigurationInformation to just store the build paths by extracting Botan version and other unrelated data

Review could happen by individual commits, as I think the individual steps are easier to review as the diff, which contains a lot of noisy moving or indentation.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -111,10 +111,9 @@ class BuildConfigurationInformation(object):
     def __init__(self, options, modules):
         self.build_dir = os.path.join(options.with_build_dir, 'build')
 
-        self.obj_dir = os.path.join(self.build_dir, 'obj')
-        self.libobj_dir = os.path.join(self.obj_dir, 'lib')
-        self.cliobj_dir = os.path.join(self.obj_dir, 'cli')
-        self.testobj_dir = os.path.join(self.obj_dir, 'test')
+        self.libobj_dir = os.path.join(self.build_dir, 'obj', 'lib')
+        self.cliobj_dir = os.path.join(self.build_dir, 'obj', 'cli')
+        self.testobj_dir = os.path.join(self.build_dir, 'obj', 'test')
 
         self.doc_output_dir = os.path.join(self.build_dir, 'docs')
         self.doc_output_dir_manual = os.path.join(self.doc_output_dir, 'manual')
--- a/configure.py
+++ b/configure.py
@@ -125,14 +125,13 @@ class BuildConfigurationInformation(object):
         self.internal_include_dir = os.path.join(self.botan_include_dir, 'internal')
         self.external_include_dir = os.path.join(self.include_dir, 'external')
 
-        self.sources = sorted(flatten([mod.sources() for mod in modules]))
         self.internal_headers = sorted(flatten([m.internal_headers() for m in modules]))
         self.external_headers = sorted(flatten([m.external_headers() for m in modules]))
 
         if options.amalgamation:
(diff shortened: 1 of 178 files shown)


randombit:
Everything LGTM, thank you!

randombit:
On testing, ran into some problems:

```
$ ./configure.py --verbose
INFO:root:./configure.py invoked with options "--verbose"
   INFO: ./configure.py invoked with options "--verbose"
INFO:root:Platform: OS="Linux" machine="x86_64" proc=""
   INFO: Platform: OS="Linux" machine="x86_64" proc=""
DEBUG:root:Traceback (most recent call last):
  File "./configure.py", line 2525, in <module>
    main()
  File "./configure.py", line 2245, in main
    find_files_named('info.txt', options.lib_dir)]])
  File "./configure.py", line 2244, in <listcomp>
    [ModuleInfo(info) for info in
  File "./configure.py", line 692, in __init__
    self._defines = parse_lex_dict(lex.defines)
  File "./configure.py", line 532, in parse_lex_dict
    for key, sep, value in [as_list[3*i:3*i+3] for i in range(0, len(as_list)/3)]:
TypeError: 'float' object cannot be interpreted as an integer

  DEBUG: Traceback (most recent call last):
  File "./configure.py", line 2525, in <module>
    main()
  File "./configure.py", line 2245, in main
    find_files_named('info.txt', options.lib_dir)]])
  File "./configure.py", line 2244, in <listcomp>
    [ModuleInfo(info) for info in
  File "./configure.py", line 692, in __init__
    self._defines = parse_lex_dict(lex.defines)
  File "./configure.py", line 532, in parse_lex_dict
    for key, sep, value in [as_list[3*i:3*i+3] for i in range(0, len(as_list)/3)]:
TypeError: 'float' object cannot be interpreted as an integer

ERROR:root:'float' object cannot be interpreted as an integer
  ERROR: 'float' object cannot be interpreted as an integer
```

First, the crash itself, but also that each log line seems to be getting duplicated. I see this crash on Python 3.6.0 on Arch, but not with 2.7.13. But the log lines seem to be getting duplicated on CI also, both on Ubuntu 

https://travis-ci.org/randombit/botan/jobs/217732684#L540

and OS X

https://travis-ci.org/randombit/botan/jobs/217732702#L1255

webmaster128:
Found the division issue. The duplicate log entries came with https://github.com/randombit/botan/pull/962/commits/4c026d0d97e4d0d8ec10f0b542499db780f330cc. But I have no clue how this change affects logging :/ Any ideas?

randombit:
Aha? Hinted at by a S.O. post which mentions that "Also be advised that if you log before any handler is added, a default StreamHandler(sys.stderr) will be created."

`Version.vc_rev` is:

```
    vc_rev = botan_version.release_vc_rev if botan_version.release_vc_rev else get_vc_revision()
```

And `get_vc_revision` has debug logging in it. Since the variable is static, it is initialized before main runs and sets up the logger. So then two log handlers (a default one and the custom one) are both enabled.

One workaround might be to make `Version.vc_rev` a static function instead of a static variable, then `get_vc_revision` is not called until after main has started.

[I have not tested any of this but it seems right]

webmaster128:
Great, thanks. I'll follow up on this

webmaster128:
ok I think this is done

randombit:
LGTM thanks

PR #963, opened 2017-04-02
WIP: Code refactoring

ksanderon:
I would like to improve in nearest days quality of botan code - this is first step.

Changes which we have in this PR are mostly done in half automated way. I plan to introduce later manual refactoring for some(or all) parts of code - get rid of raw pointers, RAII in some places, removal of not needed virtual, lvalue ref instead or rvalue ref if rvalue not needed etc, but this is precondition for me.

However here we have:
* fixes for deprecated constructions in c++11 and later (explicit rule of 3/5 or implicit rule of 0 and other violations)
* `default` specifier instead of `{}` in some places(not all yet)
* removal of unreachable code (for example `return` after `throw`)
* fix for `throw()` specifier -> `noexcept`
* fixed code formating style: missing braces for one line statements, non whitesmith formating etc... (done mostly automatic way by astyle - some complex macros required some tuning to be correctly interpreted by astyle )
* removed not needed semicolons

Additional things:
* I added astyle configuration which I used to reformat code
* Added cmake support for development purposes (because of IDE integrations) - not for release builds of course ;) It was generated based on `Makefile` generated by `./configure.py` on mac os x. 
* possibly these files should be moved to ./src/scripts? 
* if changes regarding formating style are not acceptable because to much changes - no problem I have another branch with all other.

**TODO:**
* manual check indentations:
    * [ ] cli
    * [ ] tests
    * [ ] lib
        * [ ] asn1
        * [ ] base
        * [ ] block
        * [ ] codec
        * [ ] entropy
        * [ ] ffi
        * [ ] filters
        * [ ] hash
        * [ ] kdf
        * [ ] mac
        * [ ] math
        * [ ] misc
        * [ ] modes
        * [ ] passhash
        * [ ] pbkdf
        * [ ] pk_pad
        * [ ] prov
        * [ ] pubkey
        * [ ] rng
        * [ ] stream
        * [ ] tls
        * [ ] utils
        * [ ]  x509


src/cli/main.cpp
```diff
@@ -37,8 +37,8 @@ std::string main_help()
 int main(int argc, char* argv[])
    {
    std::cerr << Botan::runtime_version_check(BOTAN_VERSION_MAJOR,
-                                             BOTAN_VERSION_MINOR,
-                                             BOTAN_VERSION_PATCH);
+             BOTAN_VERSION_MINOR,
+             BOTAN_VERSION_PATCH);
```
webmaster128: Does this change make sense? I am used to the following two indentation methods from other technologies (Python, Swift):

1. break all arguments and use default indent level

```
std::cerr << Botan::runtime_version_check(
   BOTAN_VERSION_MAJOR,
   BOTAN_VERSION_MINOR,
   BOTAN_VERSION_PATCH
);
```

2. Indent to first argument start

```
std::cerr << Botan::runtime_version_check(BOTAN_VERSION_MAJOR,
                                          BOTAN_VERSION_MINOR,
                                          BOTAN_VERSION_PATCH);
```

So the original code looks good to me the change does not improve things in case rule 1. and 2. are desired.

src/cli/pubkey.cpp
```diff
@@ -99,19 +105,24 @@ class PK_Sign final : public Command
       void go() override
          {
          std::unique_ptr<Botan::Private_Key> key(Botan::PKCS8::load_key(get_arg("key"),
-                                                                        rng(),
-                                                                        get_arg("passphrase")));
+                                                 rng(),
+                                                 get_arg("passphrase")));
```
webmaster128: Same here: the old version did better group load_key arguments

src/cli/timing_tests.cpp
```diff
@@ -57,9 +57,9 @@ class Timing_Test
       virtual ~Timing_Test() {}
 
       std::vector<std::vector<ticks>>
-      execute_evaluation(const std::vector<std::string>& inputs,
-                         size_t warmup_runs,
-                         size_t measurement_runs);
+                                   execute_evaluation(const std::vector<std::string>& inputs,
```
webmaster128: Can this gigantic 30 space indent for manual line breaks be turned off or be reduced to a single tab width (3 spaces)?

src/cli/main.cpp
ksanderon: You are right ;) There could be this kind of deviations - astyle is not so inteligent - however it's easy to fix to follow 1) or 2) just one more rule is needed. I will fix it. and prepare self review here for this kind of issues ;)

src/cli/timing_tests.cpp
ksanderon: Most of these not good looking things are result of "inherited" formating. It should be easy to fix. I will care about it.

randombit:
Hi! Both the formatting fixes and the other C++ changes are most welcome. As are your mentioned future plans.

I'm not happy about the hardcoded file lists, version, compilation flags, etc in CMakeLists as it's one more thing to be maintained over time. Also it misses any files that don't build on x86-64, or that aren't built by default (zlib, TPM support, etc). It would be better if CMakeLists could be generated by configure.py, as bakefiles are (see function gen_bakefile in configure.py, flag --with-bakefile) so that the version, flags, etc remain correct over time. Then an developer using an IDE does a one time `configure.py` to generate the CMakeLists for their platform, and they are off.

If you can do 2 PRs, one for formatting and one for everything else that would be much better.

I've only partially reviewed the changes so far. This change (and the 65 others like it) is wrong:

```
-         for(auto&& version : versions)
+         for(auto && version : versions)
```

`auto && version` reads like a boolean AND operation, rather than `a universal reference. (http://stackoverflow.com/questions/26991393/what-does-auto-e-do-in-range-based-for-loops). Probably most/all of the affected loop variables could be changed to `auto&` or `const auto&` instead.

ksanderon:
@randombit : Yes, I found that rvalue references are interpreted sometimes by astyle incorrect, however from other hand I not found yet in botan place where &&(as universal reference) is really needed, so as you said, probable in most cases where we have universal reference we an put const& or &.

randombit:
The source files `src/lib/tls/tls_suite_info.cpp` and `src/lib/asn1/oids.cpp` are generated by scripts (`src/scripts/tls_suite_info.py` and `src/scripts/oids.py`) which must be changed to emit the corrected format. No use just changing the source file, since it will be lost next time those files are regenerated.

ksanderon:
@randombit : regarding PR split - for me It also will be easier - at least in case of long review and many rebases ;)

ksanderon:
So if you can please just attach only general findings(I will create other PR with refactoring withour core reformat and after it will be accepted next with reformat) - to not duplicate issues next time.

PR #965, opened 2017-04-02
Basic refactor

ksanderon:
Same as we have in #963 , however without code style reformat - so less changes and easier to review. 

Content:
* fixes for deprecated constructions in c++11 and later (explicit rule of 3/5 or implicit rule of 0 and other violations)
* `default` specifier instead of `{}` in some places(probably all)
* removal of unreachable code (for example `return` after `throw`)
* removal of compilation unit only visible, but not used functions
* fix for `throw()` specifier ~~(just removed, because in used context it doesn't matter, while noexcept is not supported on MSVC 2013)~~ **edit: used instead `BOTAN_NOEXCEPT`**
* removed not needed semicolons

### DIFF
--- a/src/lib/base/secmem.h
+++ b/src/lib/base/secmem.h
@@ -46,12 +46,20 @@ class secure_allocator
       typedef std::size_t     size_type;
       typedef std::ptrdiff_t  difference_type;
 
-      secure_allocator() BOTAN_NOEXCEPT {}
+#ifdef BOTAN_BUILD_COMPILER_IS_MSVC_2013
+      secure_allocator() = default;
+      secure_allocator(const secure_allocator&) = default;
+      secure_allocator& operator=(const secure_allocator&) = default;
+      ~secure_allocator() = default;
+#else
+      secure_allocator() BOTAN_NOEXCEPT = default;
+      secure_allocator(const secure_allocator&) BOTAN_NOEXCEPT = default;
+      secure_allocator& operator=(const secure_allocator&) BOTAN_NOEXCEPT = default;
+      ~secure_allocator() BOTAN_NOEXCEPT = default;
+#endif
 
       template<typename U>
       secure_allocator(const secure_allocator<U>&) BOTAN_NOEXCEPT {}
-
-      ~secure_allocator() BOTAN_NOEXCEPT {}
 
       pointer address(reference x) const BOTAN_NOEXCEPT
          { return std::addressof(x); }
--- a/src/lib/pubkey/pk_ops_impl.h
+++ b/src/lib/pubkey/pk_ops_impl.h
@@ -9,6 +9,9 @@
 #define BOTAN_PK_OPERATION_IMPL_H__
 
 #include <botan/pk_ops.h>
+#include <botan/eme.h>
+#include <botan/kdf.h>
+#include <botan/emsa.h>
 
 namespace Botan {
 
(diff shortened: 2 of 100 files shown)


src/lib/base/secmem.h
```diff
@@ -21,7 +21,7 @@
 namespace Botan {
 
 template<typename T>
-class secure_allocator
```
ksanderon: Maybe secure_allocator was designed to be base class for something? (or users of library may use it as base class for their own?)

src/lib/base/secmem.h
```diff
@@ -46,13 +46,6 @@ class secure_allocator
       typedef std::size_t     size_type;
       typedef std::ptrdiff_t  difference_type;
 
-      secure_allocator() BOTAN_NOEXCEPT {}
-
-      template<typename U>
-      secure_allocator(const secure_allocator<U>&) BOTAN_NOEXCEPT {}
```
ksanderon: This looks little strange. What for was that?

src/lib/tls/tls_handshake_io.h
```diff
@@ -177,8 +177,6 @@ class Datagram_Handshake_IO final : public Handshake_IO
          Message_Info(uint16_t e, Handshake_Type mt, const std::vector<uint8_t>& msg) :
             epoch(e), msg_type(mt), msg_bits(msg) {}
 
-         Message_Info(const Message_Info& other) = default;
```
ksanderon: Message_Info is intended to be base class for something? Maybe it should be marked final?

ksanderon:
@randombit 
@webmaster128 

I closed #963 (and created this PR instead) because I suppose with smaller parts it will be easier for reviewers to check changes.

src/lib/base/secmem.h
ksanderon: for msvc 2015 :/

src/lib/tls/tls_handshake_io.h
randombit: No, `Message_Info` is just a struct used internally by (and is private to) `Datagram_Handshake_IO`, used for saving message flights for retransmission. It should be `final`, as should `Handshake_Reassembly` in the same header.

ksanderon: So I will make both final in this PR and try to remember to take care in simillar cases.

src/lib/pubkey/pk_ops_impl.h
```diff
@@ -9,6 +9,9 @@
 #define BOTAN_PK_OPERATION_IMPL_H__
 
 #include <botan/pk_ops.h>
+#include <botan/eme.h>
+#include <botan/kdf.h>
```
ksanderon: please check here if there headers can be moved.

src/lib/tls/tls_messages.h
```diff
@@ -20,17 +20,17 @@
 #include <string>
 #include <set>
 
-namespace Botan {
-
-class Credentials_Manager;
+#if defined(BOTAN_HAS_CECPQ1)
+  #include <botan/cecpq1.h>
+#endif
```
ksanderon: please check if it is acceptable to not use forward declaration here.

randombit: Not ideal, but acceptable as long as the included headers are all public (as here).

src/lib/pubkey/pk_ops_impl.h
randombit: The reason for the empty destructors in pk_ops.cpp was to run ~unique_ptr<T> where T was forward declared (EMSA, KDF, EME), to avoid pulling in these headers. But, all three are public headers belonging to the basic API, and your change simplifies the code overall, so ok.

randombit:
Everything looks good. Can you rebase and squash these 7 commits into a single commit with a log message outlining the changes made? What you have in the opening PR comment is fine.

ksanderon:
No problem I will rebase and squash it.

ksanderon:
Done.

ksanderon:
@randombit : failed job on travis is coverage - codecov hanged for long time.

PR #966, opened 2017-04-03
Add SM3 hash function

dewyatt:
There is some small code duplication here but it is intentional.

I have run a fair amount of random data of varying size through this and 3 other implementations, and everything is matching up perfectly.

Sources:
* https://tools.ietf.org/html/draft-shen-sm3-hash-00
* https://tools.ietf.org/html/draft-shen-sm3-hash-01
* http://www.oscca.gov.cn/UpFile/20101222141857786.pdf
* http://ris.sic.gov.cn/EN/abstract/abstract191.shtml


randombit:
Very nice, thanks. It would probably be much faster to unroll the loops in `SM3::compress_n` (as is done for SHA-2 and company) to avoid the excess variable assignments. It's not strictly necessary, if you're happy with the reduced performance, since this can always be improved later on without any compatibility concerns.

This (in info.txt) is wrong: `define SM3 20131128` the macro value should match the date the API of this module was added/last modified.

I would feel happier with more than 2 test vectors. If you could both include a comment in `sm3.vec` providing a URL for the two official vectors (pointing to the internet draft is fine) and also include randomly generated vectors from another implementation (ideally for each message size from 0 to 128 bytes, to reduce the risk of uncaught padding bugs etc) that would be +.

Do you know if an OID has been allocated for SM3? I could not find one in a brief search. But adding that would allow SM3-signed PKIX certificates and the like. I'm not sure if SM3 is deployed in this way, so it may not matter.

The BSI and NIST policies would need to be updated to prohibit `sm3`. I can do that after merging or you can do it in this PR, either way.

randombit:
Also with #962 merged you will need to additionally change info.txt

```
<defines>
SM3 -> $date
</defines>
```

Thanks!

dewyatt:
Ok I think I addressed all the items mentioned.
I did see a 34% speed up on my hardware with these changes.

There is one test that is commented out as it fails, but it also fails in the origin (https://github.com/guanzhi/GmSSL/blob/master/test/sm3test.c), so I'm not quite sure what to think about that.

Let me know if there is anything else you see that should be changed. Thanks!

dewyatt:
I'm not sure why the CI tests are failing.
`botan-test hash` is passing all 1102 tests for SM3 here with a fresh clone of the sm3 branch at riboseinc/botan and also with randombit/botan with this patch manually applied.

randombit:
Thanks for making the changes. I'll look into the test failures.

randombit:
Everything passes ok for me also... wtf?

dewyatt:
Weird...without the build artifacts it's hard to know what happened. Maybe an issue in some of the build caching?

Closed & reopened to trigger another build...let's see what happens this time.

dewyatt:
Looks like it's continuing to fail...strange.

dewyatt:
I did some testing and confirmed that the tests fail on Haswell-based GCE machines, specifically at the unrolled loops commit (05c6178).

It's surprising that it doesn't fail on our local machines.
I'll see if I can find any obvious mistakes, but it may end up being easier to abandon the unrolling.

randombit:
Also, Clang and GCC seem to be producing different results:

https://travis-ci.org/randombit/botan/jobs/218519650

Produced: A895ECFD0799A07EF456243BD7049F4949D1CB0E09E7B48B3C45964E951F41F8
Expected: 66C7F0F462EEEDD9D1F2D46BDC10E4E24167C4875CF2F7A2297DA02B8F4BA8E0

vs in Clang https://travis-ci.org/randombit/botan/jobs/218519645

Produced: 257813613FF1F2722782E9E0B52D4E29DD9BFFD71629D3ED5272309BB3EC7D25
Expected: 66C7F0F462EEEDD9D1F2D46BDC10E4E24167C4875CF2F7A2297DA02B8F4BA8E0

randombit:
I'm ok with just reverting the unrolling for now, though I am curious what is going on with this.

dewyatt:
Very odd that clang and GCC are producing different results.
I'm curious too, I'll at least try to unroll the loops one at a time and see if I can narrow it down.

dewyatt:
I went ahead and restored the two main compression loops at the bottom for now. I expect the CI tests will pass.

I'm still not sure why unrolling causes the tests to fail but I have been able to reproduce it in local containers (`docker run -ti ubuntu:trusty /bin/bash`).
It seems the [final loop](https://github.com/randombit/botan/pull/966/commits/05c6178a434c4e62e87470dda945be8dbf6b103c#diff-18b430affee47188eece8dd1695ae8ecL102), when unrolled, causes the test failures.

The compiler version also makes a difference:

```
FAIL g++ 4.8 (Ubuntu 14.04)
PASS g++ 5.4 (Ubuntu 16.04)
PASS g++ 6.3 (Arch)
```

No version of clang that I tried passes (3.4, 3.8, 3.9, svn), unless I replace that final loop.
I didn't try any special options to configure.py.

I would guess that comparing the generated assembly would be interesting, but it's probably best that I not spend any more time on this at the moment.

randombit:
Been a while since I've see either GCC or Clang have a codegen bug. Thanks for digging into this. I was also testing with GCC 6.3.

Anyway patch looks great, thanks for the contribution!

PR #967, opened 2017-04-03
Cmake support - generator in configure.py

ksanderon:
Content:
* added cmake generator to configure.py
* added .astylerc
* updated .gitignore to not see generated cmakelists

~~TODO: check on platform(for instance on mac os x) which require "frameworks" if cmake list is generated correctly.~~ **edit: done**

I'm python noob, so please point me places where things could be made smarter/cleaner/better ;)

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1433,7 +1433,7 @@ def gen_bakefile(build_config, options, external_libs):
 
     f.close()
 
-def gen_cmake(build_config, using_mods, cc, options):
+def gen_cmake(build_paths, using_mods, cc, options):
     def create_target_rules(sources):
         target = {'sources': {}, 'frameworks': set(), 'libs': set()}
         for source in sources:
@@ -1454,25 +1454,24 @@ def gen_cmake(build_config, using_mods, cc, options):
                     target['libs'].add(lib)
             if options.os in using_mod.frameworks:
                 for framework in using_mod.frameworks[options.os]:
-                    target['frameworks'].add('"-framework '+framework+'"')
+                    target['frameworks'].add('"-framework %"' % framework)
 
-    def generate_target_sources_list(fd, targetName, target):
-        fd.write('set(' + targetName + '\n')
+    def generate_target_sources_list(fd, target_name, target):
+        fd.write('set(%s\n' % target_name)
         for source in target['sources']:
-            fd.write('    ${CMAKE_CURRENT_LIST_DIR}' + os.sep + os.path.normpath(source) + '\n')
+            fd.write('    ${CMAKE_CURRENT_LIST_DIR}%s%s\n' % (os.sep, os.path.normpath(source)))
         fd.write(')\n\n')
 
     def generate_target_source_files_isa_properties(fd, target):
         for source in target['sources']:
-            joinedIsaFlags = ' '.join(target['sources'][source]['isa_flags'])
-            if joinedIsaFlags:
-                fd.write('set_source_files_properties(${CMAKE_CURRENT_LIST_DIR}' +
-                         os.sep + os.path.normpath(source) +
-                         ' PROPERTIES COMPILE_FLAGS '+ joinedIsaFlags + ')\n')
+            joined_isa_flags = ' '.join(target['sources'][source]['isa_flags'])
+            if joined_isa_flags:
+                fd.write('set_source_files_properties(${CMAKE_CURRENT_LIST_DIR}%s%s PROPERTIES COMPILE_FLAGS %s)\n'
+                         % (os.sep, os.path.normpath(source), joined_isa_flags))
 
-    library_target_configuration = create_target_rules(build_config.lib_sources)
-    tests_target_configuration = create_target_rules(build_config.test_sources)
-    cli_target_configuration = create_target_rules(build_config.cli_sources)
+    library_target_configuration = create_target_rules(build_paths.lib_sources)
+    tests_target_configuration = create_target_rules(build_paths.test_sources)
+    cli_target_configuration = create_target_rules(build_paths.cli_sources)
 
     for module in using_mods:
         add_target_details(library_target_configuration, module)
@@ -1497,8 +1496,8 @@ def gen_cmake(build_config, using_mods, cc, options):
     f.write('option(ENABLED_OPTIONAL_WARINIGS "If enabled more strict warinig policy will be used" OFF)\n')
     f.write('option(ENABLED_LTO "If enabled link time optimization will be used" OFF)\n\n')
 
-    f.write('set(COMPILER_FEATURES ' + cc.cc_compile_flags(options) + ' ' + cc.mach_abi_link_flags(options) + ')\n')
-    f.write('set(COMPILER_WARNINGS ' + cc.cc_warning_flags(options) + ')\n')
+    f.write('set(COMPILER_FEATURES %s %s)\n' % (cc.cc_compile_flags(options), cc.mach_abi_link_flags(options)))
+    f.write('set(COMPILER_WARNINGS %s)\n' % cc.cc_warning_flags(options))
     f.write('set(COMPILER_INCLUDE_DIRS build/include build/include/external)\n')
     f.write('if(ENABLED_LTO)\n')
     f.write('    set(COMPILER_FEATURES ${COMPILER_FEATURES} -lto)\n')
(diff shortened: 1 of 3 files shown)


webmaster128:
Just a comment unrelated to the quality of this PR: Caution before merging: this will clash with the changes in #962 even if git can merge it together. I can easily rebase #962 if this one goes in first. Otherwise this one needs to be updated.

configure.py
```diff
@@ -1420,6 +1424,106 @@ def bakefile_test_sources(file, sources):
...
+    test_module = create_basic_rules(build_config.test_sources)
+    cli_module = create_basic_rules(build_config.cli_sources)
+
+    for module in using_mods:
+        update_rules(lib_module, module)
+        update_rules(test_module, module)
+        update_rules(cli_module, module)
+
+    f = open('CMakeLists.txt', 'w')
+    f.write('cmake_minimum_required(VERSION 3.0.0)\n')
```
webmaster128: cmake 3.0 is a high requirement. Can you make this 2.8.12 compatible? This is what we have ob Ubuntu 14.04 LTS, which is available in the CI

ksanderon:
@webmaster128 : it should not be big problem ;) I will observe #962. I suppose your changes are more important, so should be accepted before mine.

configure.py
```diff
@@ -1420,6 +1424,106 @@ def bakefile_test_sources(file, sources):
...
+                for lib in using_mod.libs[options.os]:
+                    rules['libs'].add(lib)
+            if options.os in using_mod.frameworks:
+                for framework in using_mod.frameworks[options.os]:
+                    rules['frameworks'].add(framework)
+
+    def generate_sources_list(fd, name, rules):
+        fd.write('set(' + name + '\n')
+        for source in rules['sources']:
+            fd.write('    ${CMAKE_CURRENT_LIST_DIR}/' + source[2:] + '\n')
```
webmaster128: Cutting substrings from paths looks dangerous to me. I think [os.path.normpath](https://docs.python.org/2/library/os.path.html#os.path.normpath) is your friend

```
$ python -c "import os; print(os.path.normpath('.'))"
.
$ python -c "import os; print(os.path.normpath('ab.cpp'))"
ab.cpp
$ python -c "import os; print(os.path.normpath('./sub/ab.cpp'))"
sub/ab.cpp

However, this will result in backward slashes on Windows and I don't know if this causes problems with cmake

File: configure.py [hunk 3]
```diff
@@ -1420,6 +1424,106 @@ def bakefile_test_sources(file, sources):
...
+
+    def generate_special_properties(fd, rules):
+        for source in rules['sources']:
+            joined = ' '.join(rules['sources'][source]['isa_flags'])
+            if joined:
+                fd.write('set_source_files_properties(${CMAKE_CURRENT_LIST_DIR}/' + source[2:] + ' PROPERTIES COMPILE_FLAGS '+ joined + ')\n')
+
+    lib_module = create_basic_rules(build_config.sources)
+    test_module = create_basic_rules(build_config.test_sources)
+    cli_module = create_basic_rules(build_config.cli_sources)
```
webmaster128: You are mixing the terms "rules" and "module" here which makes it very hard to follow the code flow even for an experienced cmake/Python coder. Isn't the proper term you are searching for "target" (which is what cmake uses to describe a static/shared library or an executable)?

configure.py
```diff
@@ -1420,6 +1424,106 @@ def bakefile_test_sources(file, sources):
...
+        fd.write('set(' + name + '\n')
+        for source in rules['sources']:
+            fd.write('    ${CMAKE_CURRENT_LIST_DIR}/' + source[2:] + '\n')
+        fd.write(')\n\n')
+
+    def generate_special_properties(fd, rules):
+        for source in rules['sources']:
+            joined = ' '.join(rules['sources'][source]['isa_flags'])
+            if joined:
+                fd.write('set_source_files_properties(${CMAKE_CURRENT_LIST_DIR}/' + source[2:] + ' PROPERTIES COMPILE_FLAGS '+ joined + ')\n')
```
webmaster128: same here (normpath)

webmaster128:
You might want to have a look at https://github.com/webmaster128/botan/commits/single-cmake (newest commit only), where I tried to do this thing in a static fashion some time ago. Please just copy/past of you find useful blocks in there. It's a dead branch an I am not planning to work on it on my own anymore.

configure.py
ksanderon: Yes, It should be not problem - I need only check if in 2.8.* are avaliable: 
* set_source_files_properties, 
* target_link_libraries,
* target_compile_options,
* target_include_directories,
* set_target_properties

As these are very useful to control target dependencies(by "smart" inheritance).

configure.py
ksanderon: Yes, I agree code is unclear here ;)

configure.py
ksanderon: Thank you for hint. I'll fix it.

ksanderon:
merged and rebased

configure.py
```diff
@@ -1429,6 +1433,120 @@ def bakefile_test_sources(file, sources):
...
+
+    def generate_target_sources_list(fd, targetName, target):
+        fd.write('set(' + targetName + '\n')
+        for source in target['sources']:
+            fd.write('    ${CMAKE_CURRENT_LIST_DIR}' + os.sep + os.path.normpath(source) + '\n')
+        fd.write(')\n\n')
+
+    def generate_target_source_files_isa_properties(fd, target):
+        for source in target['sources']:
+            joinedIsaFlags = ' '.join(target['sources'][source]['isa_flags'])
```
webmaster128: We use snake_case [for variable names in Python](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables)

configure.py
```diff
@@ -1429,6 +1433,120 @@ def bakefile_test_sources(file, sources):
...
+                    libs_or_frameworks_needed = True
+                    for isa in using_mod.need_isa:
+                        target['sources'][source_path]['isa_flags'].add(cc.isa_flags[isa])
+        if libs_or_frameworks_needed == True:
+            if options.os in using_mod.libs:
+                for lib in using_mod.libs[options.os]:
+                    target['libs'].add(lib)
+            if options.os in using_mod.frameworks:
+                for framework in using_mod.frameworks[options.os]:
+                    target['frameworks'].add('"-framework '+framework+'"')
```
webmaster128: Using a string formatter makes it easier to understand what the final content will contain, especially in this case with all the quotes:

```python
target['frameworks'].add('"-framework %s"' % framework)
```

webmaster128: And then I doubt that passing "-framework XYZ" including the quotes as a single command line argument is correct. Does that work?

configure.py
```diff
@@ -1429,6 +1433,120 @@ def bakefile_test_sources(file, sources):
 
     f.close()
 
+def gen_cmake(build_config, using_mods, cc, options):
```
webmaster128: Could you rename `build_config` to `build_paths` (for new code only)? I made the class more specific, such that it is easier to understand what is in there

configure.py
```diff
@@ -1429,6 +1433,120 @@ def bakefile_test_sources(file, sources):
...
+                        target['sources'][source_path]['isa_flags'].add(cc.isa_flags[isa])
+        if libs_or_frameworks_needed == True:
+            if options.os in using_mod.libs:
+                for lib in using_mod.libs[options.os]:
+                    target['libs'].add(lib)
+            if options.os in using_mod.frameworks:
+                for framework in using_mod.frameworks[options.os]:
+                    target['frameworks'].add('"-framework '+framework+'"')
+
+    def generate_target_sources_list(fd, targetName, target):
```
webmaster128: snake_case

ksanderon: Power of habit :/ I'm afraid that, I will soon write mails with camelCase ;) Anyway, I will fix it.

configure.py
ksanderon: yes - this is somehow workaround for cmake. Without quotes it think that `-framework` `XYZ` are separate arguments ;) I'm not sure if this is cleanest solution but at least work correctly (generated  by cmake build commands are correct at least for make and ninja).

configure.py
ksanderon: done

configure.py
ksanderon: done

randombit:
This looks ok to me, ready to merge?

webmaster128:
👍 @randombit

configure.py
```diff
@@ -1429,6 +1433,121 @@ def bakefile_test_sources(file, sources):
...
+                    libs_or_frameworks_needed = True
+                    for isa in using_mod.need_isa:
+                        target['sources'][source_path]['isa_flags'].add(cc.isa_flags[isa])
+        if libs_or_frameworks_needed == True:
+            if options.os in using_mod.libs:
+                for lib in using_mod.libs[options.os]:
+                    target['libs'].add(lib)
+            if options.os in using_mod.frameworks:
+                for framework in using_mod.frameworks[options.os]:
+                    target['frameworks'].add('"-framework %"' % framework)
```
webmaster128: pylint just told me that there an "s" missing in this place: "-framework %" -> "-framework %s"

I am already fixing this and will submit erlier than expected because of this. But the version on master is not working

Issue #973, opened 2017-04-04
Refactor tracking issue


ksanderon:
I would like to improve in nearest days(weeks?) quality of botan code - this will be splited in some steps for easier review process.

**Planned content:**
* [x] Basic refactor ( #965 ):
    * [x] fixes for deprecated constructions in c++11 and later (explicit rule of 3/5 or implicit rule of 0 and other violations)
    * [x] `default` specifier instead of `{}`
    * [x] removal of unreachable code (for example `return` after `throw`)
    * [x] removal of compilation unit only visible, but not used functions
    * [x] get rid of  for `throw()` specifier in favor for `BOTAN_NOEXCEPT`
    * [x] removal of not needed semicolons
* [x] Support for cmake in configure.py ( #967  )
    * [x] mac os x
    * [x] linux x86_64
* [ ] Code reformat with astyle:
    * [x] astyle configuration file ( #967 )
    * [ ] reformated code delivery
        * [x] cli ( #980 )
        * [x] tests ( #980 )
        * [ ] lib
             * [ ] asn1
             * [ ] base
             * [ ] block
             * [ ] codec
             * [ ] entropy
             * [ ] ffi
             * [ ] filters
             * [ ] hash
             * [ ] kdf
             * [ ] mac
             * [ ] math
             * [ ] misc
             * [ ] modes
             * [ ] passhash
             * [ ] pbkdf
             * [ ] pk_pad
             * [ ] prov
             * [ ] pubkey
             * [ ] rng
             * [ ] stream
             * [ ] tls
             * [ ] utils
             * [ ] x509
* [ ] get rid of raw pointers - smart pointers instead in 99% of cases (may be some minor interface change will be needed - in case of `X * X::copy/clone() and friends` -> `std::unique_ptr<X> X::pristine()`)
    * [x] helper header file which provide missed stuff in std (+ fixes for compilers incompatibility if needed):
        * [x] `make_unique` (#974)
* [ ] RAII in some places,
* [ ] removal of not needed `virtual` and virtual inheritance, 
* [ ] lvalue ref instead or rvalue ref if rvalue not needed
* [ ] `final` for classes which should not be base for something
* [ ] refactor C++03 style private ctors/dctors style in favour of C++11 `delete`
* [ ] refactor `for` loops refactor in fovour of `for ... in`
* [ ] add install target support in cmakefile generated by config.py

Optional but possible and (maybe) reasonable in some places:
[...]

PR #974, opened 2017-04-04
Add wrapper for missing in C++11 std::make_unique

ksanderon:
This PR is part of #973.

Content:
* provided `Botan::make_unique()` which should be equal to C++14 `std::make_unique`. In case of Botan compilation with C++ 14, stl implementatoin will be used.

This should allow further refactoring with removal explicite `new` and `delete` usage(in 99% cases) in favor of smart pointers.

### DIFF
--- a/src/lib/utils/info.txt
+++ b/src/lib/utils/info.txt
@@ -22,6 +22,7 @@ parsing.h
 rotate.h
 types.h
 version.h
+stl_compatibility.h
 </header:public>
 
 <header:internal>
--- a/src/lib/utils/stl_compatibility.h
+++ b/src/lib/utils/stl_compatibility.h
@@ -1,0 +1,77 @@
+/*
+* STL standards compatibility functions
+* (C) 2017 Tomasz Frydrych
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#ifndef BOTAN_STL_COMPATIBILITY_H__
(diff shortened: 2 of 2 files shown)


ksanderon:
Do we have something like "death tests" in botan (tests for things, which should not work or even compile - for instnce static_asserts etc.)?

Maybe some tests required?

randombit:
A worthy goal and thanks again for these cleanups, explicit calls to `delete` are nearly always a bug waiting to happen.

webmaster128:
:+1: 

I think the entire bridging part should be replaced by

```
#if defined _MSC_VER || defined __cpp_lib_make_unique
    using std::make_unique;
#else

// fallback
```

ksanderon:
Huh! I not know that something like __cpp_lib_* even exist ;) 
Also just `using std::make_unique;` could do the job.

So with next PR I will fix it.

ksanderon:
With 
```
#if defined _MSC_VER || defined __cpp_lib_make_unique
    using std::make_unique;
#else
```
might be only one problem - I'm not sure if MSVC will correctly handle `using` in this context. But I will just try so we will know ;)

webmaster128:
This works fine in production software compiled with MSVS 2013/2015/2017:

https://github.com/kullo/client-libkullo/blob/master/kulloclient/util/misc.h

PR #981, opened 2017-04-05
Cut absolute path from main configuration executable

webmaster128:


webmaster128:
Can this coverage check be made less strict somehow?

It looks like the difference in coverage just comes from random error cases during the test execution:
https://codecov.io/gh/randombit/botan/compare/22e812d...eb0856a64c09e98672cee28a2e3b38ecb74ad9c8/changes

randombit:
Stuck with it I think, I don't see anything in docs for codecov.yaml about this https://docs.codecov.io/docs/codecov-yaml nor are there any relevant options on codecov.io for me as project owner.

We can avoid the random flux by running coverage CI build with some fixed DRBG seed.

PR #984, opened 2017-04-07
Pylint compliance for cmake, module resolution and amalgamtion

webmaster128:
This is a refactoring of the topics mentioned.

I confirmed that the generated amalgamation does not differ from current master using our own project (custom modules list, 12 different target platforms, botan_all_aesni.cpp used on some systems).

Apart from the code refactorings, there are minor improvements in the configuration results:
* include guard name is added to #endif (e.g. `#endif` -> `#endif // BOTAN_AMALGAMATION_H__`)
* botan_all_internal.h now also gets a license banner
* source files list is now sorted for convenience in CMakeLists.txt

After this, we only need those 3 things for green lint light:
1. Refactoring of configure's main()
2. Cleanup botan2.py
3. Check Python 2 pylint
All of those 3 are minor tasks compared to this current PR.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -153,10 +153,10 @@ inlinevar-rgx=[A-Za-z_][A-Za-z0-9_]*$
 inlinevar-name-hint=[A-Za-z_][A-Za-z0-9_]*$
 
 # Regular expression matching correct method names
-method-rgx=[a-z_][a-z0-9_]{2,30}$
+method-rgx=[a-z_][a-z0-9_]{2,45}$
 
 # Naming hint for method names
-method-name-hint=[a-z_][a-z0-9_]{2,30}$
+method-name-hint=[a-z_][a-z0-9_]{2,45}$
 
 # Regular expression matching correct constant names
 const-rgx=(([A-Z_][A-Z0-9_]*)|(__.*__))$
--- a/configure.py
+++ b/configure.py
@@ -2330,6 +2330,21 @@ class AmalgamationGenerator(object):
                 target = 'sse2'
         return target
 
+    def _isas_for_target(self, target):
+        for mod in sorted(self._modules):
+            # Only first module for target is considered. Does this make sense?
+            if self._target_for_module(mod) == target:
+                out = set()
+                for isa in mod.need_isa:
+                    if isa == 'aesni':
+                        isa = "aes,ssse3,pclmul"
+                    elif isa == 'rdrand':
+                        isa = 'rdrnd'
+                    out.add(isa)
+                return out
+        # Return set such that we can also iterate over result in the NA case
+        return set()
+
     def generate(self):
         pub_header_amalag = AmalgamationHeader(self._build_paths.public_headers)
         header_name = '%s.h' % (AmalgamationGenerator.filename_prefix)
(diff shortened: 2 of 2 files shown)


randombit:
Awesome, will review this over the weekend.

randombit:
Only problem I see so far is the verbose logging in `_resolve_dependencies` is just too verbose, and repeats itself a lot.

```
  DEBUG: Searching for base...
  DEBUG: Already loaded.
  DEBUG: Searching for base...
  DEBUG: Already loaded.
  DEBUG: Searching for hash...
  DEBUG: Already loaded.
  DEBUG: Resolving dependencies of block...
  DEBUG: Searching for base...
  DEBUG: Already loaded.
  DEBUG: Resolving dependencies of aont...
  DEBUG: Searching for ctr...
  DEBUG: Already loaded.
  DEBUG: Searching for rng...
  DEBUG: Already loaded.
  DEBUG: Searching for filters...
  DEBUG: Already loaded.
  DEBUG: Searching for base...
  DEBUG: Already loaded.
...
```

I'm not sure this level of logging helps anyone, except perhaps you while you were debugging your changes :)

I'd be ok with either removing the logs entirely (also simplifying `_resolve_dependencies` a bit by removing `level` argument), reducing the log output to just the 'informative bits' (say something like

```
diff --git a/configure.py b/configure.py
index b9b731daa..d8b697de2 100755
--- a/configure.py
+++ b/configure.py
@@ -2046,25 +2046,26 @@ class ModulesChooser(object):

     def _resolve_dependencies(self, modname, level=0):
         indent = "    "*level
-        logging.debug(indent + "Resolving dependencies of %s..." % modname)
+
+        dependencies = self._modules[modname].dependencies()
+
+        logging.debug(indent + "Resolving dependencies of %s [%s]" %
+                      (modname, ' '.join(dependencies)))

         for dependency in dependencies:
             dependency_choices = set(dependency.split('|'))

-            logging.debug(indent + "Searching for %s..." % dependency)
-
             dependency_met = False
             new_loaded_mod = None

             # Prefer what we already loaded
             if not set(dependency_choices).isdisjoint(self._to_load):
-                logging.debug(indent + "Already loaded.")
                 dependency_met = True
             else:
                 possible_mods = dependency_choices.intersection(self._maybe_dep)
                 if possible_mods:
                     new_loaded_mod = possible_mods.pop()
-                    logging.debug(indent + "Adding %s..." % new_loaded_mod)
+                    logging.debug(indent + "Adding module '%s' required by '%s'" % (new_loaded_mod, modname))
                     dependency_met = True
```

or else leaving the logs but guarding them by some additional command line flag (`--trace` or `--really-verbose` or whatever) so when we ask someone who is having problems with the build to post the output of `configure.py --verbose` it's not 1200+ lines (!). Leaving `--trace` logging for developers who are working on changes to configure, and are ok with a log firehose.

webmaster128:
Ahh, damn, I knew about this issue but then forgot about it after I was done and did not use debug anymore.

I think I'll remove the logging code entirely as this is mostly development code. We can add it again when there are issues with dependency resolution.

randombit:
Sounds good, thanks

webmaster128:
okay, done

PR #986, opened 2017-04-08
Fix DSA reduction

randombit:
If the hash value was much larger than `q` this loop could take a long time. Accidentally ran into this issue by using DSA "Raw" hashing (which bypasses EMSA1's bitshifting) then trying to sigh a SHA-256 hash with a 160 bit `q`.

@neusdan This look ok to you?

### DIFF
--- a/src/lib/pubkey/dsa/dsa.cpp
+++ b/src/lib/pubkey/dsa/dsa.cpp
@@ -111,8 +111,8 @@ DSA_Signature_Operation::raw_sign(const uint8_t msg[], size_t msg_len,
    {
    BigInt i(msg, msg_len);
 
-   while(i >= m_q)
-      i -= m_q;
+   if(i >= m_q)
+      i = (i % m_q);
 
 #if defined(BOTAN_HAS_RFC6979_GENERATOR)
    BOTAN_UNUSED(rng);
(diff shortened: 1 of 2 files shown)


randombit:
Fixed building the test when EME_RAW was disabled, and rebased. Think this is ok to merge now but would like a second set of eyes that this is really what we want to do here.

securitykernel:
@neusdan is on vacation this week. I was discussing this with @jurajsomorovsky and @neverhub. ISO 14888-3 says nothing about this, there `|h(m)| <= |q|` always holds. It gives valid combinations for the bitlength of p and q and the hash functions's output length, namely (1024,160,160), (2048,224,224), (3072,256,256), (7680,384,384) and (15360,512,512).

@neverhub found that FIPS-186-4 requires the first `N = |q|` bits of the hash output to be taken into account when the output length of the hash function exceeds N, which should be equivalent to what EMSA1 does. He wanted to look into what other implementations do, if they support "Raw" signatures.

@jurajsomorovsky Pointed out that even with "Raw" signing you would hash manually before (instead of directly signing the message), so using the first N bits in this case too would sound reasonable.

neverhub:
Quick update: I took a look at the openssl DSA [implementation](https://github.com/openssl/openssl/blob/master/crypto/dsa/dsa_ossl.c#L61) first. Note that this DSA signing function does only sign the `|q|` leftmost bits of the passed message/digest, as  FIPS-186-4 requires. Same thing in the nss DSA [implementation](https://github.com/nss-dev/nss/blob/master/lib/freebl/dsa.c#L317). Both do thereby require that the message is hashed manually beforehand.

securitykernel:
I looked up BSI TR-02102-1 again and it also demands to sign only the `|q|` leftmost bits. I also checked with another colleague with a strong crypto background and he confirmed that this would be the right thing to do here. We could think of the modulo as a (very bad) hash function.

The handbook currently states that when the message is too long, an exception is thrown, which neither matches the old nor the new implementation, but could also be an option. I would go with the first option, as it is what other libraries do. That would essentially mean we would have to extract EMSA1's `emsa1_encoding()`. The same holds for ECDSA, ECGDSA and ECKCDSA.

randombit:
No need to port the EMSA1 logic around since in all cases at this point in the signature algorithm we are representing the message as a BigInt. So `bn.mask_bits(q.bits())` should suffice.

I'll need to generate tests at least for DSA/ECDSA for this situation.

randombit:
Closing in favor of #1502

Issue #982, opened 2017-04-06
Botan 2.1.0 does not compile under Cygwin 2.8.0 with g++ 5.4.0


SideChannel:
Hi,

the recent release 2.1.0 does not compile with my system at work while 2.0.1 does (same setup). It has problems compiling os_utils.cpp.

System
--------

- Windows 7 SP1, Cygwin 2.8.0, Compiler g++ 5.4.0

Steps to reproduce the problem
---------------------------------
tar xzvf Botan-2.1.0.tgz
cd Botan-2.1.0
./configure.py
   INFO: ./configure.py invoked with options ""
   INFO: Platform: OS="CYGWIN_NT-6.1" machine="x86_64" proc=""
   INFO: Guessing target OS is cygwin (use --os to set)
   INFO: Guessing to use compiler gcc (use --cc to set)
   INFO: Detected CPU model "Intel(R) Xeon(R) CPU W3690 @ 3.47GHz" in /proc/cpuinfo
   INFO: Guessing target processor is a x86_64/x86_64 (use --cpu to set)
   INFO: Target is gcc-cygwin-x86_64-x86_64
   INFO: Found sphinx-build (use --without-sphinx to disable)
WARNING: Shared libs not supported on cygwin, disabling shared lib support
   INFO: Skipping, dependency failure - certstor_sqlite3 pkcs11 sessions_sqlite3
   INFO: Skipping, incompatible OS - darwin_secrandom dyn_load getentropy locking_allocator
   INFO: Skipping, no enabled compression schemes - compression
   INFO: Skipping, requires external dependency - boost bzip2 lzma openssl sqlite3 tpm zlib
   INFO: Loading modules adler32 aead aes aes_ni aes_ssse3 aont asn1 auto_rng base base64 bcrypt bigint blake2 block blowfish camellia cascade cast cbc cbc_mac
ccm cecpq1 certstor_sql cfb chacha chacha20poly1305 chacha_sse2 clmul cmac codec_filt comb4p crc24 crc32 cryptoapi_rng cryptobox ctr curve25519 des dev_random dh dl_algo dl_group dlies dsa eax ec_gfp ec_group ecc_key ecdh ecdsa ecgdsa ecies
 eckcdsa elgamal eme_oaep eme_pkcs1 eme_raw emsa1 emsa_pkcs1 emsa_pssr emsa_raw
[...]

PR #987, opened 2017-04-08
Cygwin compilation problem

randombit:
Blind fix for GH #982

Looking at the error posted in the issue a bit more, it seems that sigaction and company do exist on Cygwin, but are macros rather than function declarations. So prefixing them with `::` to refer to the global namespace causes weird syntax errors.


randombit:
Didn't fix #982, closing

PR #989, opened 2017-04-09
Avoid recursion in BER_Decoder::get_next_object

randombit:
With a long enough sequence of EOC markers you could cause `BER_Decoder` to exhaust its stack. Found by OSS-Fuzz (bug 813 on their tracker). Security bug insofar as it can cause a crash on parsing untrusted inputs, not otherwise exploitable afaict.

Even minimized, the clusterfuzz test case is large (768K), and even that does not trigger with default Linux stack size, I had to use `ulimit -s 512`. It should be easy enough to generate the malformed BER at runtime instead.

### DIFF
--- a/src/lib/asn1/ber_dec.cpp
+++ b/src/lib/asn1/ber_dec.cpp
@@ -218,20 +218,25 @@ BER_Object BER_Decoder::get_next_object()
       return next;
       }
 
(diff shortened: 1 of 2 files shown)


randombit:
Minimized test case turned out to be easy enough because the EOC marker is 0 bytes - trying to parse `std::vector<uint8_t> in(10000000, 0)` was sufficient to repro.

PR #992, opened 2017-04-10
configure: remove internal sorting of ModuleInfo object

webmaster128:
before, `python3 configure.py --amalgamation` failed with "TypeError: unorderable types: ModuleInfo() < ModuleInfo()" as `__cmp__` is not state of the art anymore (https://docs.python.org/release/3.0.1/whatsnew/3.0.html#ordering-comparisons)


webmaster128:
This should fix the CI fail here: https://ci.kullo.de:8154/go/pipelines/using-botan/375/build-botan/1

Issue #972, opened 2017-04-04
Pipe::process_msg() does not finish encryption to file


lieser:
I noticed a strange behavior regarding `Botan::Pipe::process_msg()` and `Botan::DataSink_Stream`. If i understand the [manual](https://botan.randombit.net/manual/filters.html) and the API of [`Botan::Filter`](https://botan.randombit.net/doxygen/classBotan_1_1Filter.html#a5378ec3bf4adf860a4cfa01cfc51e20e) correctly, `end_msg()` should flush the buffers. Because `process_msg()` is sending `end_msg()` at the end, it too should flush the buffers (and write the output to the file).
But this is not the behavior I observed. After `process_msg()` is called, the output file is still incomplete (see code below). Only after calling `Botan::Pipe::reset()` or `Botan::~Pipe()` is the file completely written to.

I'm a little unsure if this is really a bug, or a misunderstanding of the API on my side. If it is a misunderstanding, I think the documentation should be made more clear.

I also think it would be a good idea to include the openmode `std::ios::binary` in the example code of [Data Sinks](https://botan.randombit.net/manual/filters.html#data-sinks).

``` C++
std::vector<uint8_t> key = Botan::hex_decode("128068ddbb6c5a51b5d5ca93b03b40e2f7b4b823606d2354006a9d7b0296de44" );
Botan::InitializationVector iv = Botan::hex_decode( "FFFFFFFFFFFFFFFFFFFFFFFF" );

// encrypt file in.txt to out_enc.txt
Botan::Pipe pipeEnc( Botan::get_cipher( "AES-256/GCM", key, iv, Botan::ENCRYPTION ), new Botan::DataSink_Stream( "out_enc.txt", std::ios::binary ) );
Botan::DataSource_Stream inEnc( "in.txt", std::ios::binary );
pipeEnc.process_msg( inEnc );

// at this time, the out_enc.txt file is still incomplete

// decrypt file out_enc.txt to out_dec.txt
Botan::Pipe pipeDec( Botan::get_cipher( "AES-256/GCM", key, iv, Botan::DECRYPTION ), new Botan::DataSink_Stream( "out_dec.txt", std::ios::binary ) );
Botan::DataSource_Stream inDec( "out_enc.txt", std::ios::binary );
pipeDec.process_msg( inDec );
// decryption fails, because input file out_dec.txt ist still incomplete.
```

randombit:
[...]

PR #994, opened 2017-04-11
fix missing flush in DataSink_Stream::end_msg

lieser:
This is a fix for #972.

I'm a little unsure if it is OK how I implemented the test case. Current I create a temporary file `botan_test_data_src_sink_flush.tmp` in the directory the test is executed in.


randombit:
Test case looks ok, we probably should extent the test framework to handle temporary files more generally but this is good for now. Thanks!

PR #995, opened 2017-04-11
Add support file for aarch64

danimo:


webmaster128:
Isn't that covered by https://github.com/randombit/botan/blob/master/src/build-data/arch/arm64.txt?

danimo:
@webmaster128 Indeed. Failed to spot the alias section. Sorry for the noise!

Issue #999, opened 2017-04-12
CTR-BE should allow seeking()


reneme:
We are planning to use Botan to encrypt large files before uploading them into cloud storage. Since we need to deal with flaky networks, a possibility for resuming half-way through is desired.

Hence, we would like to use `CTR-BE(AES-256)` for its `StreamCipher` behaviour. Unfortunately, the `CTR_BE` implementation does not allow to use `::seek()`. It just throws a `Not_Implemented` exception.

Are there plans to implement `CTR_BE::seek()` or is there an inherent problem that prevents an implementation?

randombit:
No inherit problem, just hasn't been implemented (same goes for `Salsa20::seek`).

Supporting seeking will require a few changes to `CTR_BE`. Currently the class maintains two buffers, `m_counter` and `m_pad` the first keeps track of the current counter, and is encrypted by the underlying block cipher to produce `m_pad` which is XORed with the plaintext. The counter is initialized using the IV, though normally its actually 2-8 blocks wide and stores (IV+0,IV+1,IV+2,IV+3,...), and the pad is correspondingly also several blocks wide. This allows parallel or bitsliced implementations of the block cipher. (See `CTR_BE::increment_counter`)

This arrangement doesn't work well with `seek` because at the point `seek` is called, `CTR_BE` does not know how many times it has incremented the counter or where it is within the cipher stream, all it knows is the current value of the counter (which is derived from the original IV + # of bytes processed).

Fix is probably to save the original IV in a new member variable. Then if/when `seek` is called, recompute `m_counter` using the original IV plus the seek offset, then encrypt it to produce the new `m_pad` and set `m_pad_pos` to the desired byte within that block.

Thanks for pointing out this problem, would be happy to take a patch and might fix it myself eventually but I have other things I must do right now.

reneme:
Pull request GH #1000 is ready from my point of view.

randombit:
With #1000 merged, closing. Thanks for the contribution

PR #1000, opened 2017-04-13
Implement CTR_BE::seek()

reneme:
This is implementing my feature request in #999.

### Code Changes

* add member `CTR_BE::m_iv` to store original IV (as set in `set_iv()`)
* add implementation for `CTR_BE::seek()` inspired by `CTR_BE::increment_counter()`
* `CTR_BE::set_iv()` now uses `seek(0)` to set up `m_counter`

### Test Suite Extension

* derive basic `seek()` tests from existing test data in `src/tests/data/stream/ctr.vec` for all available algorithms
* add a few "long-seek" tests (seeking several kB/MB), generated like so:
  * encrypt large data block without seeking
  * use last bytes of input and output and corresponding "seek-value" for the tests

### ~Potential ToDos~

* ~`increment_counter()` and `seek()` share very similar implementation of big-endian addition~
  * the former adds a small (< 255) constant value to each "parallel block" (nomenclature?) while the latter adds a potentially large value that is incremented for each "parallel block"
  * Refactoring idea: `seek()` might only initialize `m_counter` with the IVs and base-counter values (0, 1, 2, 3, ...). A generalized version of `increment_counter()` could then add a (potentially large) constant value.
  * ~didn't do that (yet) to simplify initial code review~
  * any thoughts on refactoring this?


reneme:
added the refactoring discussed above. We are using this code since two weeks without problems.

randombit:
Thanks. It will be a couple of days before I have time to look at this.

Issue #1004, opened 2017-04-14
Add ARIA Block Cipher


noloader:
Attached is Korea Internet & Security Agency's ARIA block cipher. It is based on the 32-bit implementation by Aaram Yun from the National Security Research Institute, KOREA. Yun's implementation is based on the 8-bit implementation by Jin Hong. Those fellows should get the credit since they did the hard work.

I think its at a place where its ready for integration into Botan. One of the Botan team members will need to finish the integration. Here, "finish" means closing the gaps that I overlooked.

Here is a website that has a lot of good information, like English documents and test vectors: [Block Cipher Algorithm ARIA](http://210.104.33.10/ARIA/index-e.html). Additional test vectors can be found [here](https://github.com/weidai11/cryptopp/blob/master/TestVectors/aria.txt) and [here](https://github.com/weidai11/cryptopp/blob/master/TestData/aria.dat).

Here are some of the numbers:

* **Skylake, Linux, x86_64**:

```
skylake:botan$ ./botan speed --msec=3000 Camellia-128 Camellia-256 ARIA-128 ARIA-256
Camellia-128 [base] encrypt 109.764 MiB/sec 23.57 cycles/byte (329.31 MiB in 3000.18 ms)
Camellia-128 [base] decrypt 110.215 MiB/sec 23.47 cycles/byte (330.69 MiB in 3000.38 ms)
Camellia-256 [base] encrypt 89.666 MiB/sec 28.85 cycles/byte (269.00 MiB in 3000.03 ms)
Camellia-256 [base] decrypt 89.592 MiB/sec 28.88 cycles/byte (268.81 MiB in 3000.40 ms)
ARIA-128 [base] encrypt 132.606 MiB/sec 19.51 cycles/byte (397.88 MiB in 3000.42 ms)
ARIA-128 [base] decrypt 132.603 MiB/sec 19.51 cycles/byte (397.81 MiB in 3000.03 ms)
ARIA-256 [base] encrypt 98.538 MiB/sec 26.25 cycles/byte (295.62 MiB in 3000.11 ms)
ARIA-256 [base] decrypt 98.553 MiB/sec 26.24 cycles/byte (295.69 MiB in 3000.28 ms)
```

* **Core2 Duo, Linux, x86_64**:

```
core2:botan$ ./botan speed --msec=3000 Camellia-128 Camellia-256 ARIA-128 ARIA-256
[...]

PR #1005, opened 2017-04-14
Add ARIA Block Cipher (GH #1004)

noloader:
This ARIA implementation is based on the 32-bit implementation by Aaram Yun from the National Security Research Institute, KOREA. Aaram Yun's implementation is based on the 8-bit implementation by Jin Hong.

I think its at a place where its ready for integration into Botan. One of the Botan team members will need to finish the integration. Here, "finish" means closing the gaps that I overlooked.

Also see [GitHub Issue 1004](https://github.com/randombit/botan/issues/1004).

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1476,7 +1476,7 @@ class CmakeGenerator(object):
         for source in sorted_sources:
             joined_isa_flags = ' '.join(target['sources'][source]['isa_flags'])
             if joined_isa_flags:
-                fd.write('set_source_files_properties(${CMAKE_CURRENT_LIST_DIR}%s%s PROPERTIES COMPILE_FLAGS %s)\n'
+                fd.write('set_source_files_properties(${CMAKE_CURRENT_LIST_DIR}%s%s PROPERTIES COMPILE_FLAGS "%s")\n'
                          % (os.sep, os.path.normpath(source), joined_isa_flags))
 
     @staticmethod


noloader:
The failed compile on Windows is due to the `alignas(16)`. It should be criminal for Microsoft to claim they are C++11. They can't even handle `alignof` and `alignas`.  And don't get me started on [Dynamic Initialization and Destruction with Concurrency](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2008/n2660.htm), which is a core C++11 feature that MS did not provide until 9 years after it was introduced.

Anyway... What does Botan use to abstract `alignas`?

-----

Here are the Linux numbers with and without `alignas`. 4 MB/s difference on a modern Skylake is non-trivial:

* ***with `alignas`***:

```
ARIA-128 [base] encrypt 132.606 MiB/sec 19.51 cycles/byte (397.88 MiB in 3000.42 ms)
ARIA-128 [base] decrypt 132.603 MiB/sec 19.51 cycles/byte (397.81 MiB in 3000.03 ms)
ARIA-256 [base] encrypt 98.538 MiB/sec 26.25 cycles/byte (295.62 MiB in 3000.11 ms)
ARIA-256 [base] decrypt 98.553 MiB/sec 26.24 cycles/byte (295.69 MiB in 3000.28 ms)
```

* ***without `alignas`***:

```
ARIA-128 [base] encrypt 127.982 MiB/sec 20.21 cycles/byte (384.00 MiB in 3000.43 ms)
ARIA-128 [base] decrypt 127.928 MiB/sec 20.22 cycles/byte (383.81 MiB in 3000.21 ms)
ARIA-256 [base] encrypt 94.965 MiB/sec 27.24 cycles/byte (284.94 MiB in 3000.44 ms)
ARIA-256 [base] decrypt 94.968 MiB/sec 27.23 cycles/byte (284.94 MiB in 3000.35 ms)
```

noloader:
The failure is due to [Issue 1011](https://github.com/randombit/botan/issues/1011):

```
src/lib/ffi/ffi.cpp: In function ‘int botan_pubkey_load_elgamal(botan_pubkey_struct**, botan_mp_t, botan_mp_t, botan_mp_t)’:

./src/lib/ffi/ffi.cpp:1450:22: error: ‘q’ was not declared in this scope

    BOTAN_UNUSED(q);

                      ^

./src/lib/ffi/ffi.cpp:1452:22: error: ‘x’ was not declared in this scope

    BOTAN_UNUSED(x);

                      ^
```

noloader:
This was closed because I was not able to keep Botan's upstream and my upstream sync'd properly. There were too many damn merges. Its too bad Git makes it so damn hard to simply check-out file, check-in files, and compare files for differences.

PR #1014, opened 2017-04-18
Add ARIA Block Cipher (GH #1004)

noloader:
This ARIA implementation is based on the 32-bit implementation by Aaram Yun from the National Security Research Institute, KOREA. Aaram Yun's implementation is based on the 8-bit implementation by Jin Hong.

I think its at a place where its ready for integration into Botan. One of the Botan team members will need to finish the integration. Here, "finish" means closing the gaps that I overlooked.

The original PR was closed because I was not able to keep Botan's upstream and my upstream sync'd properly. There were too many damn merges. I'm pretty sure [GitHub's instructions](https://help.github.com/articles/syncing-a-fork/) are wrong given there were no conflicts. This PR should be cleaner.

Also see [GitHub Issue 1004](https://github.com/randombit/botan/issues/1004).


randombit:
Closing in favor of #1157 thanks for this patch!

PR #1017, opened 2017-04-20
Complete wildcard handling for X.509 certificates

securitykernel:
Hostname validation is used to make sure the certificate hostname matches the hostname of the connected host, e.g., in TLS. RFC 6125 allows one wildcard in the left-most label of a hostname. Up to now, we only supported only the wildcard as the left-most label, e.g., `www.example.com` would match `*.example.com`, but `www.example.com` would not match `www*.example.com`, although it is permitted by the RFC. Also adds test vectors from RFC 6125 as well as the OpenSSL test suite.

Implementation note: The code would have been shorter if `Botan::split_on()` would work on strings with less than two elements instead of throwing an exception, e.g., splitting `"a*"` on `*` into `"a"` and `""`.

### DIFF
--- a/src/lib/utils/parsing.cpp
+++ b/src/lib/utils/parsing.cpp
@@ -2,6 +2,7 @@
 * Various string utils and parsing functions
 * (C) 1999-2007,2013,2014,2015 Jack Lloyd
 * (C) 2015 Simon Warta (Kullo GmbH)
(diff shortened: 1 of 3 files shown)


randombit:
Thanks, I'll want to review this one more carefully before merge (probably this weekend), but ok in principle.

PR #1023, opened 2017-04-25
Add some missing function overrides in TLS::Text_Policy

louiz:
Maybe we should think of a way to check that all future addition to TLS::Policy are also added into Text_Policy


randombit:
Your suggestion on ensuring the two classes stay consistent is good but I'm not sure how to test it effectively. Another area where inconsistencies arise is in `Policy::print`, which misses a number of more recently added fields.

randombit:
The escape hatch you left for applications by making the function `protected` is good too, thanks

PR #1024, opened 2017-04-25
Make configure script Pylint complient and run in CI

webmaster128:
I think this is the last step needed to have run pylint on configure.py on Travis.

As always I tried to keep the disabled rules at a minimum and as local as possible. So this needed to include a lot of refactoring.

I disables the botan2.py lint fail for now in order to get configure.py checking up and running.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+# See http://editorconfig.org/
+
+root = true
+
+[*]
+end_of_line = lf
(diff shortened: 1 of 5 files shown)


webmaster128:
Looks like remaining Travis fails are unrelated to this change. Same fails happen in #1022 and #1023.

randombit:
Nice cleanups in here thank you

PR #1032, opened 2017-04-29
Throw OpenSSL exception if any OpenSSL function failed.

bluhm:
Checking for all failures helps to find problems early.  The
OpenSSL_Error() exception provides the OpenSSL error string.


randombit:
Nice thank you

PR #1033, opened 2017-04-29
Fix botan_bcrypt_is_valid return code.

dewyatt:
Currently it returns 0 on success and BOTAN_FFI_ERROR_EXCEPTION_THROWN otherwise,
which is a bit inaccurate and not consistent with the rest of the FFI interface.

TEST_FFI_FAIL is used in the test, so no update is needed there (unless you would prefer an explicit check on the return code).


dewyatt:
I forgot to mention, the docs for this function do say:
> Returns 0 if if this password/hash combination is valid,
   1 if the combination is not valid (but otherwise well formed),
   negative on error.

So this change would be in compliance with the documentation.

ronaldtse:
Thank you @dewyatt and @randombit !

Issue #1028, opened 2017-04-27
Module dependency resolution fails


randombit:
As example, `./configure.py --disable-modules=prf_tls` disables `tls`, but then `tls_cbc` is left enabled, causing a build failure. 

Think something in #984 broke this.

webmaster128:
On master I currently get

```
g++ -fstack-protector -m64 -pthread -fPIC -fvisibility=hidden -std=c++11 -D_REENTRANT -O3 -momit-leaf-frame-pointer -Wall -Wextra -Wpedantic -Wstrict-aliasing -Wstrict-overflow=5 -Wcast-align -Wmissing-declarations -Wpointer-arith -Wcast-qual -Wzero-as-null-pointer-constant -Wnon-virtual-dtor -Ibuild/include -Ibuild/include/external -c ./src/lib/tls/sessions_sql/tls_session_manager_sql.cpp -o build/obj/lib/tls_sessions_sql_tls_session_manager_sql.o
In file included from ./src/lib/tls/sessions_sql/tls_session_manager_sql.cpp:8:0:
build/include/botan/tls_session_manager_sql.h:11:39: fatal error: botan/tls_session_manager.h: No such file or directory
compilation terminated.
```

for the configuration you posted. Looks like sessions_sql is missing a dependency on tls. This is the case for tls_cbc too.

webmaster128:
Do we have implicit dependencies on modules in parent folder? I don't remember seeing any code handling this

webmaster128:
I am currently bisecting this. Let's see ...

webmaster128:
a4f9e50611 from #984 broke this. But the bug might be somewhere else and only becomes visible by this change, e.g. in ModuleInfo.dependencies(). Unfortunately I don't have the time to debug this this week

randombit:
Yes all modules have implicit dependencies on their parent dir, that's in `ModuleInfo.dependencies` where the parent module (if existing) is added as an implicit dep. You can also see this with eg `--disable-modules=aead` which in 2.1 and prior disabled all AEAD modes but on master they still are included in the build but of course fail since aead.h is not there.

I think the problem is in

```
+                if new_loaded_mod:
 +                    self._maybe_dep.remove(new_loaded_mod)
 +                    self._to_load.add(new_loaded_mod)
 +                    # Added new_loaded_mod, now resolve its dependencies
 +                    self._resolve_dependencies(new_loaded_mod, level+1)
```

[...]

PR #1034, opened 2017-04-29
Rework dependency resolution

webmaster128:
This may do the trick for #1028


randombit:
Works for every case I could think of, thank you for looking into this.

PR #1040, opened 2017-05-02
KDF SP800-56A: Single-step version

flowher:
Patch implements hash and HMAC based concatenation KDF as according to SP800-56A revision 2.
This change is needed in order to support RFC 4880 (ECDH)
Patch includes test vectors which were generated with BouncyCastle (hash based KDF) and cryptography.io (HMAC based KDF)


flowher:
argh.... 32bit vs 64bit...

src/lib/kdf/sp800_56a/sp800_56a.cpp
```diff
@@ -0,0 +1,99 @@
+/*
+* KDF defined in NIST SP 800-56a (Approved Alternative 1)
+*
+* (C) 2017 Ribose Inc.
+* Written by Krzysztof Kwiatkowski.
```
webmaster128: Could you write that into one line to make it easier to add years and additional authors.

src/lib/kdf/sp800_56a/sp800_56a.h
```diff
@@ -0,0 +1,59 @@
+/*
+* KDF defined in NIST SP 800-56a revision 2 (Single-step key-derivation function)
+*
+* (C) 2017 Ribose Inc.
+* Written by Krzysztof Kwiatkowski.
```
webmaster128: same here

src/lib/kdf/sp800_56a/sp800_56a.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+#include <botan/sp800_56a.h>
+#include <botan/hmac.h>
+#include <botan/scan_name.h>
+#include <botan/exceptn.h>
+
+namespace Botan {
+
+namespace {
+static const uint64_t MAX_REPS = (2ULL << 32);
```
webmaster128: static in a cpp file and anonymous namespace is basically the same thing for variables, so the "static" keyword should be removed here;

Regarding naming: The value you define here is not a maximum but a strict upper bound, right?

Regarding value: this is shifting binary 11 to the left, which does not end up as 2^32

flowher: ah... of course! Thanks!

randombit:
Thank you! FYI I made some changes after merging, specifically I did not really like the use of a template for the `SP800_56A` class (it's not really obvious to a user that they can't use say a block cipher as the aux function, and such an error would be diagnosed as a link failure which is not so friendly), so split it out into `SP800_56A_Hash` and `SP800_56A_HMAC` classes which take the respective types.

The code to handle the requirements of sec 5.8.1.1 wrt an empty salt is made much simpler when you realize that with HMAC,  a key of all zeros of the length of the hash block size is precisely equivalent to an empty key, so `hmac->set_key(salt, salt_len)` works for both empty and non-empty salt cases. :) The wording in the NIST SP does not make this obvious.

flowher:
Ah, yes. That's right.
Thanks !

PR #1041, opened 2017-05-03
Make botan2.py pylint compliant and integrate into CI

webmaster128:
btw.: how stable does this file need to be? I would be great if we could rename classes to CamelCase

### DIFF
--- a/src/python/botan2.py
+++ b/src/python/botan2.py
@@ -30,7 +30,7 @@ if sys.platform == 'darwin':
 else:
     botan = CDLL('libbotan-2.so') # pylint: disable=invalid-name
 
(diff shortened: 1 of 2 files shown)


webmaster128:
Only diff hits missing in this PR, which is out of scope for this pice of work

randombit:
@webmaster128 I guess there is nothing really firmly stating the Python API is 'stable'/'frozen'/whatever-you-want-to-call-it, but not sure we want to go crazy with it either. But for this particular case can't we just rename them to CamelCase and then add compatibility 'typedefs' like `hash_function = HashFunction`? Then any code written against the 2.0/2.1 Python modules still works but new code can use the Pythonic names.

PR #1043, opened 2017-05-06
Fix CMake Generator for visual studio

RyuheiMori:
details:
-Most of CMake commands accept only unix style path
-Visual Stduio Generator allows mutiple configurations (debug;release)
-Name collsion of static libary and dynamic libary and exe
-Mismatch of dllexport/dllimport
-Allows "x86_64:avx2" in isa flags

python configure.py --with-cmake --cpu=x64 & cmake . -A x64 -G"Visual Studio 14 2015"

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1447,11 +1447,16 @@ def gen_bakefile(build_config, options, external_libs):
 
 
 class CmakeGenerator(object):
-    def __init__(self, build_paths, using_mods, cc, options):
+    def __init__(self, build_paths, using_mods, cc, options, template_vars):
         self._build_paths = build_paths
         self._using_mods = using_mods
         self._cc = cc
         self._options = options
+        self._template_vars = template_vars
+
+    @staticmethod
+    def _escape(str):
+        return str.replace('(','\\(').replace(')','\\)').replace('#','\\#').replace('$','\\$')
 
     @staticmethod
     def _create_target_rules(sources):
@@ -1467,7 +1472,7 @@ class CmakeGenerator(object):
                 if source_path == mod_source:
                     libs_or_frameworks_needed = True
                     for isa in using_mod.need_isa:
-                        target['sources'][source_path]['isa_flags'].add(self._cc.isa_flags[isa])
+                        target['sources'][source_path]['isa_flags'].add(self._cc.isa_flags_for(isa,self._template_vars['arch']))
         if libs_or_frameworks_needed:
             if self._options.os in using_mod.libs:
                 for lib in using_mod.libs[self._options.os]:
@@ -1481,8 +1486,9 @@ class CmakeGenerator(object):
         fd.write('set(%s\n' % target_name)
         sorted_sources = sorted(target['sources'].keys())
         for source in sorted_sources:
-            fd.write('    ${CMAKE_CURRENT_LIST_DIR}%s%s\n' % (os.sep, os.path.normpath(source)))
+            fd.write(('    ${CMAKE_CURRENT_LIST_DIR}%s%s\n' % (os.sep, os.path.normpath(source))).replace('\\','\\\\'))
         fd.write(')\n\n')
+        fd.write('to_cmake_paths("${%s}" %s)\n\n' %(target_name,target_name))
 
     @staticmethod
     def _generate_target_source_files_isa_properties(fd, target):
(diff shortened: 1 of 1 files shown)


configure.py
```diff
@@ -1481,8 +1487,9 @@ def _generate_target_sources_list(fd, target_name, target):
         fd.write('set(%s\n' % target_name)
         sorted_sources = sorted(target['sources'].keys())
         for source in sorted_sources:
-            fd.write('    ${CMAKE_CURRENT_LIST_DIR}%s%s\n' % (os.sep, os.path.normpath(source)))
+            fd.write(('    ${CMAKE_CURRENT_LIST_DIR}%s%s\n' % (os.sep, os.path.normpath(source))).replace('\\', '\\\\'))
```
webmaster128: Why keep the os.sep and backward slashes in this place? Maybe we should have a cmake_normpath function in Python, that does os.path.normpath and replace("\\", "/").

This makes it easier to review the generated cmake file and to_cmake_paths becomes obsolete, right?

configure.py
```diff
@@ -1496,17 +1503,43 @@ def _generate_target_source_files_isa_properties(fd, target):
     @staticmethod
     def _write_header(fd):
         fd.write('cmake_minimum_required(VERSION 2.8.0)\n')
-        fd.write('project(botan LANGUAGES CXX)\n\n')
+        fd.write('project(botan)\n\n')
+        fd.write('if(POLICY CMP0042)\n\n')
```
webmaster128: "\n\n" -> "\n" for the three lines that belong together without empty lines in between

configure.py
```diff
@@ -1496,17 +1503,43 @@ def _generate_target_source_files_isa_properties(fd, target):
...
+        fd.write('cmake_policy(SET CMP0042 NEW)\n\n')
```
webmaster128: This line is in the code twice now

configure.py
```diff
@@ -1496,17 +1503,43 @@ def _generate_target_source_files_isa_properties(fd, target):
...
         fd.write('cmake_policy(SET CMP0042 NEW)\n\n')
+        fd.write('endif()\n\n')
+        fd.write('function(to_cmake_paths paths results)\n')
+        fd.write('    set(_results)\n')
+        fd.write('	foreach(path ${paths})\n')
```
webmaster128: If we still need this function, let's use 4 space indent for all the function body

configure.py
```diff
@@ -1496,17 +1503,43 @@ def _generate_target_source_files_isa_properties(fd, target):
...
     def _write_footer(self, fd, library_link, cli_link, tests_link):
         fd.write('\n')
 
         fd.write('option(ENABLED_OPTIONAL_WARINIGS "If enabled more strict warinig policy will be used" OFF)\n')
         fd.write('option(ENABLED_LTO "If enabled link time optimization will be used" OFF)\n\n')
-
-        fd.write('set(COMPILER_FEATURES %s %s)\n'
+        original_no_optimizations = self._options.no_optimizations
+        original_with_debug_info = self._options.with_debug_info
+        self._options.no_optimizations = False
```
webmaster128: Changing self._options is not what you want to do here, because self._options is a map, which is a reference type in Python. So whenever you touch your reference of self._options, the entire configure script is affected. I know you revert the changes later but this is easily changed into bugs.

Let's take the safe route and create a local copy of the options that you can youse locally for whatever you want and throw away when you're done:

```python
my_local_options = copy.deepcopy(self._options)
```

configure.py
```diff
@@ -1559,10 +1594,9 @@ def generate(self):
             self._add_target_details(tests_target_configuration, module)
             self._add_target_details(cli_target_configuration, module)
 
-        library_target_libs_and_frameworks = '%s %s %s' % (
+        library_target_libs_and_frameworks = '%s %s' % (
             ' '.join(library_target_configuration['frameworks']),
-            ' '.join(library_target_configuration['libs']),
-            self._cc.mach_abi_link_flags(self._options)
+            ' '.join(library_target_configuration['libs'])
```
webmaster128: Let's keep the optional trailing comma, reducing the diff here and for future changes to the list

ksanderon:
@RyuheiMori : current cmake support is little noobish - because it was made by me, only to have some qt-creator/clang-model features for #973 ;) I think, that If you have experience with cmake on windows(and some spare time), we can try to make it much better and implement cmake as alternative for configure.py - at least for linux, mac os x and windows.

I will add to #973 to implement "install" target to cmake (as I forgot about it last time, while it is very useful)(with generation of `botan-config.cmake` to cleanly use `find_package` in user projects).

PR #1044, opened 2017-05-13
Handle IV carryover in CBC, CFB, and stream ciphers

### LINKED ISSUE #864: False assertion 'The nonce is fresh for this message'
We've hit an assertion failure which is introduced in botan 2.0 (since 1.11 dev branch). It breaks the behaviour in botan-1.10, with which we could re-use nonces for encryption. But with botan-2, an assertion failure is triggered.

However, this is how shadowsocks stream protocol is designed, and other cipher suites work perfectly well (including previous version of botan). The rationale behind this assertion is to enforce "better practice" and "better security". However, it should really be up to downstream developers rather than library providers.

This is not a programmatic error, can we remove this assertion? This is effectively breaking the back-compatibility with software that has been written with botan-1.10 that needs this particular behaviour.

References:
https://marc.info/?l=botan-devel&m=145755126216465&w=2
https://github.com/shadowsocks/libQtShadowsocks/issues/104

randombit:
Allow an empty nonce to mean "continue using the current cipher state".

@librehat Can you check this resolves #864 for you?

@neusdan @securitykernel Would like a quick second opinion on this if you have time.


securitykernel:
Looks good to me on a first sight.

PR #844, opened 2017-01-18
Add support for ARMv8 SHA1/SHA2 instructions

randombit:
Based almost entirely on @noloader's patches in #840 and #841 

Relies on #843 which is merged yet.

PMULL patch comes in different PR, probably not today.

Requires at least GCC 4.9 for the crypto intrinsics, so this will probably fail CI build. I need to work on something for restricting certain modules based on compiler version, this is also needed for Goldmont SHA-1 which requires GCC 5.0

Tested on qemu+GCC 6.3 and an AMD Opteron 1100. Opteron showed 4-5x speedup.

PR #826, opened 2017-01-07
Add support for Intel SHA instructions

randombit:
#824 Need to install SDE to test this. But it compiles at least. :)

Based on GH #807 and #808

PR #1052, opened 2017-05-18
Add compiler build time detection

randombit:
And allow specific modules to require a minimum compiler version.

Needed to finally merge #826 and #844

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -340,7 +340,7 @@ exclude-protected=_asdict,_fields,_replace,_source,_make
 [DESIGN]
 
 # Maximum number of arguments for function / method
-max-args=7
+max-args=8
 
 # Argument names that match this expression will be ignored. Default to name
 # with leading underscore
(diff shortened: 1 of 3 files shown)


webmaster128:
Looks like patch version should be optional in the pattern for at least clang. See

```
$ lalalala/crystax-toolchains/x86/bin/clang -v 
clang version 3.7 
Target: i686-none-linux-android
Thread model: posix
Found candidate GCC installation: lalalala/crystax-toolchains/x86/bin/../lib/gcc/i686-linux-android/4.9
Selected GCC installation: lalalala/crystax-toolchains/x86/bin/../lib/gcc/i686-linux-android/4.9
Candidate multilib: .;@m32
Selected multilib: .;@m32
```

randombit:
Yeah I'll change this to just warn and return `None` if there is not a match - this will disable any modules requiring a specific version but the build can carry on.

Also I'll change to Clang regex to only require just the major.minor since we don't care about the patch version, looks like this output format changed in 3.8.

Issue #991, opened 2017-04-09
Pathlen Constraints


securitykernel:
Taking this from the [mailing list](https://lists.randombit.net/pipermail/botan-devel/2017-April/002206.html) here, it may get lost there.

I can reproduce the issue with a simple chain:

```
EE
  IntCA
    RootCA   pathLenConstraint = 1
```

which returns "Certificate did not validate - Certificate chain too long" when running ./botan cert_verify ee.crt int.crt root.crt.

[pathlen.zip](https://github.com/randombit/botan/files/908524/pathlen.zip)

However, while your proposed fix works for this chain, it fails for others. Run botan-test x509_test_x509test and x509_test_nist and you'll see invalid chains incorrectly accepted where CERT_CHAIN_TOO_LONG was expected, for example:

```
EE
  Lev3CA
    Lev2CA   pathLenConstraint = 1
      Lev1CA   pathLenConstraint = 1
        Root
```

When reaching Lev2CA, 2 < 2 will evaluate to false and thus then chain will validate correctly.

I also tried adopting the algorithm listed in RFC 5280, but can't get it to work, other tests fail. In part this seems not to work because Botan uses zero as the default value for pathLenConstraint on end entity
certificates internally.

Any ideas?

securitykernel:
Fixed in #1053.

PR #1053, opened 2017-05-19
Fix pathLenConstraint validation

neusdan:
Fixes GH #991

The problem with the current implementation of the chain validation code is that is runs from the end certificate to the trust cert.
@securitykernel and me tried to fix the pathLenConstraint validation within this reverse loop but we were not sure if we missed some edge cases.

So we felt safer to use the algorithm listed in RFC 5280 which executes from the top to the bottom. It's probably best to rewrite the code to use the whole algorithm from RFC 5280, i.e. validating the chain from the trust to the end cert.

Additionally, we wrote some tests including the one that raised this issue initially.


randombit:
LGTM thanks.

CI failure was because the online OCSP test hits Let's Encrypt OCSP responder, which had an outage.

PR #1058, opened 2017-05-24
Make Botan compile with LibreSSL again.

bluhm:
Add some #ifdef LIBRESSL_VERSION_NUMBER in addition to the
OPENSSL_VERSION_NUMBER switch.  Narrow down API compatiblity between
LibreSSL and OpenSSL version in docs.


randombit:
Thanks.

PR #1063, opened 2017-06-05
Added std::ios::binary option for istreams in compression. 

volka:
Without this the compress/uncompress methods may produce corrupted files on
Windows.


randombit:
Thank you!

PR #1064, opened 2017-06-06
Fix missing binary mode flag on compression ofstreams

volka:
Missing binary flags on ostreams may corrupt data in compression on Windows. This is the ofstream commit similar to my earlier ifstream pull request. Sorry should have pushed them all together.


randombit:
Thanks - not using Windows much anymore, I forget that binary vs text mode actually makes a difference there.

PR #1066, opened 2017-06-07
Add Ed25519 signatures

randombit:
Based on the public domain ref10 implementation from SUPERCOP.

This work was sponsored by Ribose Inc


randombit:
Visual C++ seems to miscompile the code on 32-bit x86. Affects even VC2017.

Sound be nice to fix this but I don't even have a Windows dev VM around anymore and deducing which of the hundreds of lines of code might be at fault via CI whackamole would take days. So... just disabled Ed25519 entirely for VC.

randombit:
Good news turned out to just be a bug (that I introduced, not in ref10) that broke 32-bit, not a VC problem.

PR #1070, opened 2017-06-09
Add assertion to ensure key is set in OCB_Encryption::encrypt.

dewyatt:
Otherwise we just SIGSEGV.


randombit:
This should already have been caught be the similar assert in `start` but I guess if you neither set a key nor an IV then call encrypt it would crash which is lame also.

PR #1074, opened 2017-06-10
FFI: Add possibility to load ECDSA and ECDH keys from MPI

### LINKED ISSUE #1073: FFI: Add possibility to load key pair for ECDSA sign/verification
It doesn't seem that FFI supports loading of EC keys in format different than BER. We need functions similar to botan_pubkey_load_dsa/botan_prvkey_load_dsa. 

I'll create a patch today

flowher:
This patch resolves #1073

This work was sponsored by Ribose Inc

### DIFF
--- a/src/lib/ffi/ffi.cpp
+++ b/src/lib/ffi/ffi.cpp
@@ -1699,9 +1699,14 @@ int botan_pubkey_ed25519_get_pubkey(botan_pubkey_t key,
 #endif
    }
 
-int botan_privkey_load_ec(botan_privkey_t* key,
-                          const botan_mp_t scalar,
-                          const char* curve_name)
+namespace {
+/* Those functions can't be defined in namespace {} that's located higher in this file
+ * as they use botan_privkey_struct/botan_pubkey_struct structs.
+ */
+int privkey_load_ec(botan_privkey_t* key,
+                    const botan_mp_t scalar,
+                    const char* curve_name,
+                    const std::string& algo_name)
    {
 #if defined(BOTAN_HAS_ECC_PUBLIC_KEY_CRYPTO)
    *key = nullptr;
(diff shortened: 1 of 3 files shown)


src/lib/ffi/ffi.cpp
```diff
@@ -1699,6 +1757,64 @@ int botan_pubkey_ed25519_get_pubkey(botan_pubkey_t key,
 #endif
    }
 
+int botan_pubkey_load_ecdsa(botan_pubkey_t* key,
+                                   const botan_mp_t public_x,
+                                   const botan_mp_t public_y,
+                                   const char* curve_name)
+   {
+
+   std::unique_ptr<Botan::ECDSA_PublicKey> p_key;
```
randombit: This will fail to compile if ECDSA isn't enabled

src/lib/ffi/ffi.cpp
```diff
@@ -1699,6 +1757,64 @@ int botan_pubkey_ed25519_get_pubkey(botan_pubkey_t key,
...
+
+   return -1;
+   }
+
+int botan_pubkey_load_ecdh(botan_pubkey_t* key,
+                                   const botan_mp_t public_x,
+                                   const botan_mp_t public_y,
+                                   const char* curve_name)
+   {
+   std::unique_ptr<Botan::ECDH_PublicKey> p_key;
```
randombit: Ditto for ECDH

src/lib/ffi/ffi.cpp
```diff
@@ -330,8 +330,66 @@ Botan::BigInt privkey_get_field(const Botan::Private_Key& key,
    throw Botan::Exception("Unsupported algorithm type for botan_privkey_get_field");
    }
 
+template<class ECPrivateKey_t>
+int privkey_load_ec(std::unique_ptr<ECPrivateKey_t>& key,
```
randombit: These functions are not exposed so should be in an anon namespace

src/lib/ffi/ffi.cpp
```diff
@@ -1699,6 +1757,64 @@ int botan_pubkey_ed25519_get_pubkey(botan_pubkey_t key,
...
+
+   return -1;
+   }
+
+int botan_privkey_load_ecdsa(botan_privkey_t* key,
+                          const botan_mp_t scalar,
+                          const char* curve_name)
+   {
+   std::unique_ptr<Botan::ECDSA_PrivateKey> p_key;
+   if(!privkey_load_ec(p_key, safe_get(scalar), curve_name))
```
randombit: If the user passes something that isn't a `botan_mp_t` for the scalar, `safe_get` will throw. Instead pass the `botan_mp_t` to the helper and dereference it in the `try`/`catch` block

flowher:
Changes applied. 
It seems to me that travis has some problems with it's environment.

randombit:
I could restart the jobs but looks certainly to be a Travis issue so I'm going to go ahead and merge. Thanks!

flowher:
Cool, thanks!

PR #1079, opened 2017-06-16
updated key checking in manual

neverhub:
Minor doc change to finally close https://github.com/randombit/botan/issues/892. I also added a small key check code example.


randombit:
Thanks

PR #1087, opened 2017-06-23
Update Doxygen for X509_Certificate::subject_info

neusdan:
Took me a while to figure out how to read out the DN email address


randombit:
Thanks

PR #1092, opened 2017-06-28
Fix transcoding of asn1print strings to UTF-8

securitykernel:
When the terminal used supports UTF-8, asn1print CLI should convert strings from internal Latin1 to UTF-8 encoding for printing to terminal. However, it previously tried to convert in the opposite direction, probably because of the misconception that `Charset::transcode()` expects the two encodings as "from, to" instead of "to, from".


randombit:
Thanks

PR #1093, opened 2017-06-28
Allow ECDSA tests for OpenSSL by comparing EC Groups

bluerise:
The ecdsa_sign test does not work for OpenSSL since the public key
that is derived from the private key does not store the OID of the
curve.  Thus, the OpenSSL code cannot find the proper NID for the
given public key and cannot do a verification operation.  We can
fix this by comparing the actual groups and not using the OIDs.

### DIFF
--- a/src/lib/prov/openssl/openssl_ec.cpp
+++ b/src/lib/prov/openssl/openssl_ec.cpp
@@ -62,35 +62,39 @@ secure_vector<uint8_t> PKCS8_for_openssl(const EC_PrivateKey& ec)
       .get_contents();
    }
 
-int OpenSSL_EC_nid_for(const EC_Group& group)
+int OpenSSL_EC_nid_for(const OID& oid)
    {
+   if(oid.empty())
+      return -1;
 
-   if(group == EC_Group("secp192r1"))
+   const std::string name = OIDS::lookup(oid);
+
+   if(name == "secp192r1")
       return NID_X9_62_prime192v1;
-   if(group == EC_Group("secp224r1"))
+   if(name == "secp224r1")
       return NID_secp224r1;
-   if(group == EC_Group("secp256r1"))
+   if(name == "secp256r1")
       return NID_X9_62_prime256v1;
-   if(group == EC_Group("secp384r1"))
+   if(name == "secp384r1")
       return NID_secp384r1;
-   if(group == EC_Group("secp521r1"))
+   if(name == "secp521r1")
       return NID_secp521r1;
 
    // OpenSSL 1.0.2 added brainpool curves
 #if OPENSSL_VERSION_NUMBER >= 0x1000200fL
-   if(group == EC_Group("brainpool160r1"))
+   if(name == "brainpool160r1")
       return NID_brainpoolP160r1;
-   if(group == EC_Group("brainpool192r1"))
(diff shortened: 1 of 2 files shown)


randombit:
A couple of the tests from `ecies_iso` battery fail - the problem is the test curve is P-256 but the curve is constructed manually instead of loading the pre-built curve. So the equality check on `EC_Group` succeeds, but then when we go to encode the key in `PKCS8_for_openssl` it fails because the OID is not set.

Anyway this is a corner case so I'll just add a check that the OID is set on the EC_Group, and if not OpenSSL will not be used. This won't effect normal applications.

randombit:
Wait, looking at this more I'm confused - first, code coverage reports that `OpenSSL_ECDSA_Verification_Operation` is being tested.

Also can you explain what you mean about "since the public key that is derived from the private key does not store the OID of the curve" - this shouldn't be occurring, the OID should get copied over.

bluerise:
Does code coverage specify that the `ecdsa_sign` test actually uses it? Where can I have a look at that?

So the signature test in `test_ecdsa.cpp` is derived from `PK_Signature_Generation_Test` in `test_pubkey.cpp`, and I think that code calls `std::unique_ptr<Botan::Public_Key> pubkey(Botan::X509::load_key(Botan::X509::BER_encode(*privkey)));`.

When it then goes to call the verifier, which is the `OpenSSL_ECDSA_Verification_Operation`, the OID that is passed is empty, so there is no NID returned, and thus it throws a `Lookup_Error`, which means that the test does not actually run for OpenSSL.

The ASN1 parser `EC_Group::EC_Group(const std::vector<uint8_t>& ber_data)` is passed a `SEQUENCE` instead of an `OBJECT_ID`, so the OID is never filled.

Sorry, I don't have a full trace of all the things that are going on to show. However, without this change the OID passed is empty. If that's a bug somewhere else, can you give me a hint where to look?

randombit:
The coverage report is here: https://codecov.io/gh/randombit/botan/src/master/src/lib/prov/openssl/openssl_ec.cpp

However I notice now that verification is only hit 11 times vs several hundred for signing, which suggests it's being called at some other point in a different test but not by `ecdsa_sign`.

Looking at it more, now I understand the problem - when encoding a public key, we always do it as the explicit encoding instead of the OID (at least unless `EC_PublicKey::set_parameter_encoding` is explicitly called). And, the test for signature verification creates the public key used for verification by encoding the key then loading it:

`std::unique_ptr<Botan::Public_Key> pubkey(Botan::X509::load_key(Botan::X509::BER_encode(*privkey)));`

(this is in `PK_Signature_Generation_Test::run_one_test` in `test_pubkey.cpp`)

So you're right that the key is losing the OID, which would cause the effect you mention.

While your change does resolve the issue I'm not sure it's quite the correct fix overall. I think what needs to happen is, if the EC group has an OID set, use the OID encoding when encoding the key instead of the literal encoding.

bluerise:
Does this look better? That works for me. 

The change that allowed the test to run was the change in `EC_PrivateKey::EC_PrivateKey(RandomNumberGenerator& rng,`.  Not sure we need the rest, but I guess it's better than not having it?

randombit:
This looks good to me, thanks

PR #1094, opened 2017-06-28
BearSSL: Initial support with Hash and ECDSA

bluerise:
BearSSL is an implementation of the SSL/TLS protocol in C aiming to be correct and secure, small and highly portable. Thus making it nicer to be included in a rather sparse bootloader.

This pull request adds initial support for BearSSL's hash and ECDSA, with more stuff coming up. The goal is to be able to test BearSSL using Botan's extensive testsuite.

### DIFF
--- a/src/lib/hash/hash.cpp
+++ b/src/lib/hash/hash.cpp
@@ -88,6 +88,10 @@
   #include <botan/blake2b.h>
 #endif
 
+#if defined(BOTAN_HAS_BEARSSL)
+  #include <botan/internal/bearssl.h>
+#endif
+
 #if defined(BOTAN_HAS_OPENSSL)
   #include <botan/internal/openssl.h>
 #endif
@@ -97,6 +101,17 @@ namespace Botan {
 std::unique_ptr<HashFunction> HashFunction::create(const std::string& algo_spec,
                                                    const std::string& provider)
    {
+#if defined(BOTAN_HAS_BEARSSL)
+   if(provider.empty() || provider == "bearssl")
+      {
+      if(auto hash = make_bearssl_hash(algo_spec))
+         return hash;
+
+      if(!provider.empty())
+         return nullptr;
+      }
+#endif
+
 #if defined(BOTAN_HAS_OPENSSL)
    if(provider.empty() || provider == "openssl")
       {
@@ -323,7 +338,7 @@ HashFunction::create_or_throw(const std::string& algo,
 
 std::vector<std::string> HashFunction::providers(const std::string& algo_spec)
    {
-   return probe_providers_of<HashFunction>(algo_spec, {"base", "openssl"});
+   return probe_providers_of<HashFunction>(algo_spec, {"base", "bearssl", "openssl"});
    }
 
 }
--- a/src/lib/hash/hash.cpp
+++ b/src/lib/hash/hash.cpp
@@ -101,10 +101,10 @@ namespace Botan {
 std::unique_ptr<HashFunction> HashFunction::create(const std::string& algo_spec,
                                                    const std::string& provider)
    {
-#if defined(BOTAN_HAS_BEARSSL)
-   if(provider.empty() || provider == "bearssl")
+#if defined(BOTAN_HAS_OPENSSL)
+   if(provider.empty() || provider == "openssl")
       {
-      if(auto hash = make_bearssl_hash(algo_spec))
+      if(auto hash = make_openssl_hash(algo_spec))
          return hash;
 
       if(!provider.empty())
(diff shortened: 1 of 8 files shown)


src/lib/prov/bearssl/bearssl_ec.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+         const SCAN_Name req(emsa);
+         m_hf = make_bearssl_hash(req.arg(0));
+         if (m_hf == NULL)
+            throw Lookup_Error("BearSSL ECDSA does not support hash " + req.arg(0));
+
+         const secure_vector<uint8_t> enc = EC2OSP(ecdsa.public_point(), PointGFp::UNCOMPRESSED);
+         m_key.qlen = enc.size();
+         m_key.q = new uint8_t[m_key.qlen];
+         memcpy(m_key.q, (unsigned char *)enc.data(), m_key.qlen);
```
securitykernel: The cast generates a warning about removing the constness. It is not necessary here.

src/lib/prov/bearssl/bearssl_ec.cpp
```diff
@@ -0,0 +1,194 @@
...
+            throw Lookup_Error("BearSSL ECDSA does not support EMSA " + emsa);
+
+         const SCAN_Name req(emsa);
+         m_hf = make_bearssl_hash(req.arg(0));
+         if (m_hf == NULL)
+            throw Lookup_Error("BearSSL ECDSA does not support hash " + req.arg(0));
+
+         const secure_vector<uint8_t> enc = EC2OSP(ecdsa.public_point(), PointGFp::UNCOMPRESSED);
+         m_key.qlen = enc.size();
+         m_key.q = new uint8_t[m_key.qlen];
```
securitykernel: The bearlssl [docs](https://bearssl.org/apidoc/index.html) state that bearlssl does no dynamic allocations and thus it will not release anything. Don't we leak memory here then?

src/lib/prov/bearssl/bearssl_ec.cpp
```diff
@@ -0,0 +1,194 @@
...
+         }
+
+      secure_vector<uint8_t> sign(RandomNumberGenerator&) override
+         {
+         const size_t order_bytes = (m_order_bits + 7) / 8;
+         secure_vector<uint8_t> sigval(2*order_bytes);
+         size_t sign_len;
+
+         br_ecdsa_sign engine = br_ecdsa_sign_raw_get_default();
+         sign_len = engine(&br_ec_prime_i31, m_hash, m_hf->final().data(), &m_key, sigval.data());
```
securitykernel: You need `sign_len` here first, so it's better to declare and define it here instead of declaring it above an leaving it unitialized. That may not be of a problem now, but may become when you refactor code in the future.

src/lib/hash/hash.cpp
```diff
@@ -97,6 +101,17 @@ namespace Botan {
 std::unique_ptr<HashFunction> HashFunction::create(const std::string& algo_spec,
                                                    const std::string& provider)
    {
+#if defined(BOTAN_HAS_BEARSSL)
+   if(provider.empty() || provider == "bearssl")
```
bluerise: Agreed.  I put it in front because I felt like sorting it alphabetically, but I will move it after OpenSSL, that's better, yes.

src/lib/prov/bearssl/bearssl_ec.cpp
```diff
@@ -0,0 +1,194 @@
...
+   if (emsa == "EMSA1(SHA-512)")
+      return &br_sha512_vtable;
+
+   return NULL;
+   }
+}
+
+#endif
+
+#if defined(BOTAN_HAS_ECDSA) && !defined(OPENSSL_NO_ECDSA)
```
bluerise: Correct, it's one library with all the stuff inside, so there's no disable button. Will correct it.

src/lib/prov/bearssl/bearssl_ec.cpp
bluerise: Sorry, you're right. I'm not particularly used to C++ memory allocations. BearSSL does not do malloc, so we definitely have to free this array again in the deconstructor I guess.

securitykernel:
There are still a few `NULL`s in `bearssl_ec.cpp`. Otherwise, looks good to me.

randombit:
@bluerise @securitykernel Sorry for taking so long to come to a decision on this but - BearSSL provider might be useful to some. I'd rather we were testing everything in CI but we can skip that for the time being. Anyway I will merge this for 2.2

PR #1096, opened 2017-06-29
Allow bit rotation by more than sizeof(T) bits.

dewyatt:
Currently `rotate_left` and `rotate_right` will happily bit shift by `>= sizeof(T)*8` bits.
However, this is undefined behavior, and leads to unexpected results (`0x00`) on at least one platform I've tested.

With this update, you can expect that `rotate_left<uint32_t>(1, 32) == 1` and `rotate_right<uint32_t>(1, 32) == 1`.

Alternatively, this responsibility could be pushed to callers (some [already do this](https://github.com/randombit/botan/blob/62c94693cb1cf5ddc5e8e43a787561e7d8351258/src/lib/block/gost_28147/gost_28147.cpp#L61
)) but personally I found the behavior surprising and thought perhaps it should be adjusted here.

**Note**: I found that my SM3 implementation is broken because of this on certain systems (a big-endian powerpc64 emulated system in qemu). So if this won't be merged, please let me know so I can do a PR to handle it within SM3.

### DIFF
--- a/src/lib/utils/rotate.h
+++ b/src/lib/utils/rotate.h
@@ -20,6 +20,7 @@ namespace Botan {
 */
 template<typename T> inline T rotate_left(T input, size_t rot)
    {
+   rot %= 8 * sizeof(T);
    return (rot == 0) ? input : static_cast<T>((input << rot) | (input >> (8*sizeof(T)-rot)));;
    }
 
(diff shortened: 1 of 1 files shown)


randombit:
Probably best to prevent undefined behavior from occurring rather than trusting the caller to always get it right. And, inlining and constant propagation should ensure that this doesn't have any real performance impact.

PR #1111, opened 2017-07-05
Turn --destdir option into an environment variable, fixes #1101, fixes #996, supersedes PR #997

### LINKED ISSUE #996: destdir vs prefix confusion
Due to Botan 2 using its own toolchain, it seems that some changes were introduced that get in the way of proper packaging. `configure.py` takes a `--destdir` optional argument, which the documentation advertises to be '/usr' or '/usr/local'.

This is what autotools-based build systems call `prefix`. It's the prefix to all other directories (`libdir`, `bindir`, etc), unless they are specified with an absolute path as well.

The purpose of `DESTDIR` (which is a shell variable) is to be able to relocate the entire tree relative to a given directory. This is used in packaging of all RPM and deb based distributions during build (note: It's _not_ meant to create relocatable binaries). `DESTDIR` should not be an argument passed to configure, it should only be evaluated by `install.py` (hence it usually is a shell variable).

We cannot use the current `destdir` for this, because it will cause the (temporary) DESTDIR prefix path to be included in e.g. header files.

### LINKED ISSUE #1101: Install path ignores --prefix when --destdir is set
i'm setting up an OBS rpm build for botan, but i can't figure how the
configure.py scripts designed to work.
For me the expected behavior (coming from autotools and co.) would be that after

./configure.py --destdir=/home/bla/BUILDROOT/ --prefix=/usr

the install tree will be under /home/bla/BUILDROOT/usr, so the following dirs will be created:

/home/bla/BUILDROOT/usr/bin
/home/bla/BUILDROOT/usr/include
...

But actually the --prefix seems to be ignored and the files installed under
/home/bla/BUILDROOT/bin
/home/bla/BUILDROOT/include
...

I could set the whole path '/home/bla/BUILDROOT/usr/' as --prefix, but then
`botan config prefix`  and also the installed .pc pkgconfig file will show that full path,
while the final install dir (after packaging) will be /usr/, so these would be broken.

It might be that i overlooked something, what do you think? For me it seems to be a bug in the build scripts...

I think the preferable behavior would be if the `make install` target would take also
a DESTDIR argument (like autotools and most of the makefile generators do), and install
the whole tree (in the subdirs specified with --libdir, --prefix, etc) under that top-level DESTDIR.

zgyarmati:
This commit removes the (broken) --destdir command line option and introduces the
DESTDIR environment variable for the 'make install' target. When it's set, the installation will
take place in the DESTDIR directory, under the prefix set by --prefix, with all of the internal
references (pkgconfig file, etc) containing only the prefix.
This behavior is more standard (as in autotools and co.), so makes packaging easier.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -458,7 +458,7 @@ def process_command_line(args): # pylint: disable=too-many-locals
     install_group.add_option('--prefix', metavar='DIR',
                              help='set the install prefix')
     install_group.add_option('--destdir', metavar='DIR',
-                             help='set the destination prefix (REMOVED, use DESTDIR '
+                             help='set the destination prefix (REMOVED, use DESTDIR ' \
                                   'environment variable when calling \'make install\')')
     install_group.add_option('--docdir', metavar='DIR',
                              help='set the doc install dir')
--- a/configure.py
+++ b/configure.py
@@ -2849,7 +2849,7 @@ def validate_options(options, info_os, info_cc, available_module_policies):
         raise UserError("Unknown module set %s" % options.module_policy)
 
     if options.destdir:
-       raise UserError("--destdir was removed. Use the DESTDIR environment variable instead when calling 'make install'")
+        raise UserError("--destdir was removed. Use the DESTDIR environment variable instead when calling 'make install'")
 
     # Warnings
 
--- a/configure.py
+++ b/configure.py
@@ -458,7 +458,8 @@ def process_command_line(args): # pylint: disable=too-many-locals
     install_group.add_option('--prefix', metavar='DIR',
                              help='set the install prefix')
     install_group.add_option('--destdir', metavar='DIR',
-                             help='set the destination prefix (REMOVED, use DESTDIR environment variable when calling \'make install\')')
+                             help='set the destination prefix (REMOVED, use DESTDIR '
+                                  'environment variable when calling \'make install\')')
     install_group.add_option('--docdir', metavar='DIR',
                              help='set the doc install dir')
     install_group.add_option('--bindir', metavar='DIR',
@@ -2849,7 +2850,8 @@ def validate_options(options, info_os, info_cc, available_module_policies):
         raise UserError("Unknown module set %s" % options.module_policy)
 
     if options.destdir:
-        raise UserError("--destdir was removed. Use the DESTDIR environment variable instead when calling 'make install'")
+        raise UserError("--destdir was removed. Use the DESTDIR environment "
+                        "variable instead when calling 'make install'")
 
     # Warnings
 
--- a/configure.py
+++ b/configure.py
@@ -458,7 +458,7 @@ def process_command_line(args): # pylint: disable=too-many-locals
     install_group.add_option('--prefix', metavar='DIR',
                              help='set the install prefix')
     install_group.add_option('--destdir', metavar='DIR',
-                             help='set the install directory')
+                             help='set the destination prefix (REMOVED, use DESTDIR environment variable when calling \'make install\')')
     install_group.add_option('--docdir', metavar='DIR',
                              help='set the doc install dir')
     install_group.add_option('--bindir', metavar='DIR',
@@ -1881,7 +1881,6 @@ def create_template_vars(source_paths, build_config, options, modules, cc, arch,
         'program_suffix': options.program_suffix or osinfo.program_suffix,
 
         'prefix': options.prefix or osinfo.install_root,
-        'destdir': options.destdir or options.prefix or osinfo.install_root,
         'bindir': options.bindir or osinfo.bin_dir,
         'libdir': options.libdir or osinfo.lib_dir,
         'includedir': options.includedir or osinfo.header_dir,
@@ -2848,6 +2847,9 @@ def validate_options(options, info_os, info_cc, available_module_policies):
 
     if options.module_policy and options.module_policy not in available_module_policies:
         raise UserError("Unknown module set %s" % options.module_policy)
+
+    if options.destdir:
+       raise UserError("--destdir was removed. Use the DESTDIR environment variable instead when calling 'make install'")
 
     # Warnings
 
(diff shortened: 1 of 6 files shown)


zgyarmati:
See the discussion and the evolution of this at #997. I'm still not sure though about the way the --destdir command line argument is obsoleted, so if @randombit or any of the maintainers would propose something, i'll be happy to implement it, just let me know.

src/scripts/install.py
```diff
@@ -63,6 +63,55 @@ def log_level():
...
+    """
+    Needed because os.path.join() discards the first path if the
+    second one is absolute, which is usually the case here. Still, we
+    want relative paths to work and leverage the os awareness of
+    os.path.join().
+    """
+    try:
+        destdir = os.environ['DESTDIR']
+    except KeyError as e:
+        destdir = ""
```
webmaster128: Those four lines should be replaced by `destdir = os.environ.get('DESTDIR', "")`

src/scripts/install.py
```diff
@@ -63,6 +63,55 @@ def log_level():
...
+          \\foo\bar (Python >3.1 only)
+          ../somewhere/else
+
+        These will be normalized to a relative path and joined with DESTDIR:
+          /absolute/dir
+          relative/dir
+          /dir/with/../inside
+          ./relative/to/me
+          ~/botan-install-test
+        """
```
webmaster128: A Python docstring is always placed right below the symbol it documents, usually a method or class name. Here we have a stray string literal that is not assigned to anything. So in this case, you should use a usual comment format (all lines with #, there are no multi line comments in Python, sorry)

src/scripts/install.py
```diff
@@ -63,6 +63,55 @@ def log_level():
...
+
+        # resolved ~, ~user
+        path = os.path.expanduser(path)
+        # native slashes, ".." inside (not in front of) pathes normalized
+        path = os.path.normpath(path)
+        # Remove / or \ prefixes if existent to accomodate for os.path.join()
+        path = path.lstrip(os.path.sep)
+        path = os.path.join(destdir, path)
+
+    return path
```
webmaster128: empty line missing

src/scripts/install.py
```diff
@@ -63,6 +63,51 @@ def log_level():
...
+def prepend_destdir(path):
+    """
+    Needed because os.path.join() discards the first path if the
+    second one is absolute, which is usually the case here. Still, we
+    want relative paths to work and leverage the os awareness of
+    os.path.join().
+    """
+    destdir = os.environ.get('DESTDIR', "")
+
+    #DESTDIR is non-empty, but we cannot join all prefix paths.
```
webmaster128: Comment is wrong now because it was moved into a wrong code branch

zgyarmati:
Comment placement and too long lines triggering pylint fixed. I see that the other lint error is being handled in #1112.
I'm not sure about the proposed relative path tests, @webmaster123 could you show a usecase from the install process?

configure.py
```diff
@@ -458,7 +458,8 @@ def process_command_line(args): # pylint: disable=too-many-locals
     install_group.add_option('--prefix', metavar='DIR',
                              help='set the install prefix')
     install_group.add_option('--destdir', metavar='DIR',
-                             help='set the destination prefix (REMOVED, use DESTDIR environment variable when calling \'make install\')')
+                             help='set the destination prefix (REMOVED, use DESTDIR '
+                                  'environment variable when calling \'make install\')')
```
webmaster128: Backslash missing (see https://github.com/randombit/botan/blob/master/configure.py#L2295)

webmaster128:
You can use a prefix relative to the current directory, e.g.

```
$ cd /home/me/Botan/botan
$ ./configure.py --prefix=foo && make -j 4 && make install
```

will install into `/home/me/Botan/botan/foo`. When prefixing DESTDIR=/bar, I'd expect `/bar/home/me/Botan/botan/foo` instead of `/bar/foo` as a result.

Since I doubt that relative prefixes are relevant together with DESTDIR, I'd enforce an absolute prefix when DESTDIR is set. But I am not a packager, so maybe I'm wrong here.

zgyarmati:
Thanks for the feedback and the clarification. So you refer to the usecase when --prefix is a relative path.
A couldn't imagine how this could be useful, so I checked how autotools work when prefix is a relative path:
```
./configure --prefix=opt
configure: error: expected an absolute directory name for --prefix: opt
```

On the other hand, cmake:

```
pwd
/home/zgyarmati/sandbox/cmocka/build
cmake -DCMAKE_INSTALL_PREFIX=usr .. && make all install DESTDIR=/tmp/cmocka
(...)
(...)
-- Installing: /tmp/cmocka/home/zgyarmati/sandbox/cmocka/build/usr/lib/libcmocka.so
```

But actually as autotools do so and normally we don't want any path information leaking from the build tree to the installation, i think it's ok if enforce --prefix to be absolute. What do you think? @randombit, any opinion?

randombit:
@zgyarmati Generally configure.py should try to follow autotools (externally visible behavior) when possible, to make life easier for downstream packagers. (Thus in general I approve of replacing `--destdir` with the `DESTDIR` env, not because I think it makes any more sense but because it matches the existing convention.) So probably prohibiting relative `--prefix` is OK, and likely avoids having to deal with nasty corner cases.

The way you deprecated/removed `--destdir` seems ok to me.

zgyarmati:
@randombit, thanks for the input, we'll work out the further details with webmaster128 and notify you when we think it's ready to merge.
@webmaster128, i'll push one more commit which explicitly checks and loudly fails if --prefix is relative, and we can continue with the review.

webmaster128:
:+1: With #1112 merged, you can rebase this onto master and run `./src/scripts/ci/travis/lint.sh` locally to do the linting

webmaster128:
The PR now disallows relative prefixes for all users only because relative prefix does not work with DESTDIR. Does it make sense to preserve the relative prefix feature and only throw an error when prepend_destdir is used with relative prefix? However, I don't know any real word use case for relative prefixes @randombit

webmaster128:
I think we should check for absolute prefix in `prepend_destdir` when `destdir` is set, not in configure. This is because prepend_destdir cannot handle relative paths, thus it should be responsible for throwing an exception. Given that DESTDIR is used by one or two handful of people, I think it is okay to error out in `DESTDIR=xyz make install` and let everyone else use relative prefixes.

PR #1114, opened 2017-07-06
Add Streebog hash (GOST R 34.11-2012).

dewyatt:
Feedback much appreciated.

I'm not sure on the naming for one. I've used Streebog here, but it's also known as GOST R 34.11-2012 of course.
I think the naming of the GOST-R-34.11-94 implementation may make things a bit confusing.


src/lib/hash/streebog/streebog.cpp
```diff
@@ -0,0 +1,165 @@
...
+                  STREEBOG_Ax[4][r[i + 32]] ^
+                  STREEBOG_Ax[5][r[i + 40]] ^
+                  STREEBOG_Ax[6][r[i + 48]] ^
+                  STREEBOG_Ax[7][r[i + 56]];
+      }
+   }
+
+inline void e(uint64_t* K, const uint8_t* m)
+   {
+   uint64_t tmp[8];
```
randombit: It is confusing here how `tmp` is used to refer to what is (in RFC 6986) called `K`, and `K` is what is called `m` in the RFC. Can we rename them in some way that will make the usage more clear?

src/lib/hash/streebog/streebog.h
```diff
@@ -0,0 +1,71 @@
...
+
+#ifndef BOTAN_STREEBOG_H__
+#define BOTAN_STREEBOG_H__
+
+#include <botan/hash.h>
+
+namespace Botan {
+
+/**
+* Streebog
```
randombit: Can you add some brief comment here on the origin of the hash (eg refer to RFC, the GOST standard, etc)

src/lib/hash/streebog/streebog.h
```diff
@@ -0,0 +1,71 @@
...
+*/
+class BOTAN_DLL Streebog_512 final : public Streebog
+   {
+   public:
+      Streebog_512() : Streebog(512) {}
+   };
+
+extern const uint64_t STREEBOG_Ax[8][256];
+extern const uint64_t STREEBOG_C[12][8];
+
```
randombit: Oh I see `lps` and `e` are just anon namespace inlines so these arrays can't be hidden class members. Hrm.

src/lib/hash/streebog/streebog.cpp
```diff
@@ -0,0 +1,165 @@
...
+      xor_buf(K, tmp, 8);
+      }
+   }
+
+inline void g(uint64_t* h, const uint8_t* m, uint64_t N)
+   {
+   uint64_t hN[8];
+   store_le(N, reinterpret_cast<uint8_t*>(&N));
+
+   std::memcpy(hN, h, 64);
```
randombit: As mentioned re the constants, I think this should just be a call to `load_le` (see the 3 arg version at https://github.com/randombit/botan/blob/master/src/lib/utils/loadstor.h#L320) so little-endian values are used throughout.

src/lib/hash/streebog/streebog.h
dewyatt: Yeah I didn't like it either, but it seemed better to keep the functions in the anonymous namespace, so I couldn't think of a nicer way. :(

src/lib/hash/streebog/streebog-precalc.cpp
```diff
@@ -0,0 +1,1680 @@
+/*
+ * Copyright (c) 2013, Alexey Degtyarev <alexey@renatasystems.org>.
+ * All rights reserved.
```
dewyatt: Exactly, the license is the same. :)
Will add more info in the comment.

src/lib/hash/streebog/streebog-precalc.cpp
```diff
@@ -0,0 +1,1680 @@
...
+ */
+
+#include <botan/streebog.h>
+
+namespace Botan {
+
+#if defined(BOTAN_TARGET_CPU_IS_LITTLE_ENDIAN)
```
dewyatt: I had a feeling you would say this since I didn't see it done anywhere else in the code. Will update.

dewyatt:
@randombit Thanks for all the great feedback, I'll get this updated soon.

src/lib/hash/streebog/streebog.h
evpo: `extern const` in `streebog-precalc.cpp` definitions seems to work. See this [patch](https://gist.github.com/evpo/70f2e2cafc1db89b1359a72511c93192) with the changes.

`const uint64_t` is `static const uint64_t` by default in c++. `extern` makes it open for other units.

randombit:
@dewyatt Friendly ping. I'm planning on releasing Botan 2.2 next Monday it would be great to be able to include this.

dewyatt:
@randombit Assuming CI passes, do you think this can make it in?
I don't know if I addressed every comment but I'm trying not to spend much time on this, considering other priorities.

randombit:
@dewyatt This looks ok, main issue was with the duplicated tables the rest is style points.

PR #1116, opened 2017-07-11
Fix loading of unencrypted PKCS#8 key via CLI

securitykernel:
CLI passed an emtpy password to `PKCS8::load_key()`, even if --pass-in was not given, which caused loading of unencrypted private keys to fail. Fixed by calling the corresponding flavour of load_key().


randombit:
Thanks

PR #1123, opened 2017-07-21
Add OCSP::Response::certificates() to access attached certificates

reneme:
This adds `Botan::OCSP::Response::certificates()` to access the (optionally attached) certificate chain of an OCSP response. We would like to access those from our application to do further checks an validation on them.


randombit:
LGTM thank you

PR #1097, opened 2017-06-29
Don't crash if openssl can't allocate a new context and catch ffi exception

catap:
This is a contribution from [Ribose Inc](https://www.ribose.com/) (@riboseinc).

PR #1128, opened 2017-07-25
Replacement for #1097: FFI exception safety

randombit:
Similar result to #1097 but with less code in ffi.cpp instead of more. :)

Also now the argument to `X_destroy` functions are checked to be of the expected type, previously they would just delete anything.

@catap Can you take a look at this if you have a chance?


catap:
@randombit yay, sorry, I forgot about it.

Anyway it looks good.

PR #1129, opened 2017-07-25
Split up ffi.cpp

randombit:
One (large) commit on top of #1128 that spits up `ffi.cpp` into several files.


randombit:
Merged after a rebase.

PR #1133, opened 2017-07-30
MSVC compiler version detection

### LINKED ISSUE #1125: configure.py fails on non-English Windows
Running configure.py on a German Windows 10 with Visual Studio 2017 fails in the compiler version detection.

```
botan>py.exe configure.py
   INFO: configure.py invoked with options ""
   INFO: Platform: OS="Windows" machine="AMD64" proc="Intel64 Family 6 Model 94 Stepping 3, GenuineIntel"
   INFO: Guessing target OS is windows (use --os to set)
   INFO: Guessing to use compiler msvc (use --cc to set)
   INFO: Guessing target processor is a x86_64/x86_64 (use --cpu to set)
   INFO: Canonicalized CPU target x86_64 to x86_64/x86_64
   INFO: Target is msvc-windows-x86_64-x86_64
Exception in thread Thread-2:
  File "C:\Program Files\Python36\lib\threading.py", line 916, in _bootstrap_inner
    self.run()
  File "C:\Program Files\Python36\lib\threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Program Files\Python36\lib\subprocess.py", line 1055, in _readerthread
    buffer.append(fh.read())
  File "C:\Program Files\Python36\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 62: character maps to <undefined>

  ERROR: Traceback (most recent call last):
  File "configure.py", line 3032, in <module>
    sys.exit(main(argv=sys.argv))
  File "configure.py", line 3016, in main
    cc_version = detect_compiler_version(cc, options.compiler_binary or cc.binary_name, osinfo.basename)
  File "configure.py", line 2599, in detect_compiler_version
    universal_newlines=True).communicate()
  File "C:\Program Files\Python36\lib\subprocess.py", line 838, in communicate
    stdout, stderr = self._communicate(input, endtime, timeout)
  File "C:\Program Files\Python36\lib\subprocess.py", line 1107, in _communicate
    stderr = stderr[0]
IndexError: list index out of range

An internal error occurred.

Don't panic, this is probably not your fault!

Please report the entire output at https://github.com/randombit/botan or email
to the mailing list https://lists.randombit.net/mailman/listinfo/botan-devel

You'll meet friendly people happy to help!
```

Byte 0x81 is the German umlaut "ü". Digging further into the code I found that Botan invokes `cl.exe` and parses the output. On a German Windows installation the output is:

```
botan>cl.exe
Microsoft (R) C/C++-Optimierungscompiler Version 19.10.25019 für x86
Copyright (C) Microsoft Corporation. Alle Rechte vorbehalten.

Syntax: cl [ Option... ] Dateiname... [ /link Linkeroption... ]
```

So configure.py chokes on the "ü" in "für x86". Apart from the text encoding problem we may have a more serious issue here, because the regex in `detect_compiler_version()` that attempts to extract the MSVC version also requires the output of `cl.exe` to be English: `r'Compiler Version ([0-9]+).([0-9]+).[0-9\.]+ for'`.

webmaster128:
Closes #1125

After digging into the issue describes in #1125, I found we have two independent problems:

1. cl.exe prints mbcs encoded data to STDOUT/STDERR, that cannot be decoded using the default decoding applied by `subprocess.Popen` with `universal_newlines=True`
2. cl.exe prints a localized text around the version number, so no text before or after the version is fixed and safe to use for regex matching

So there is no easy and solid way to ask cl.exe for it's version. We can however use the C preprocessor to print out the content of the macro `_MSC_VER`, e.g. `1900` for 19.00. This output is in the ASCII range so we don't need to care for encoding at all. This is what this PR does.

Additionally, I separated the compiler call and the output parsing in configure.py, giving us a great way to unittest the second part. This led to some regex fixes for the other compilers, mostly because `.` matches any character and we need `\.` for matching dots.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -2571,38 +2571,49 @@ class AmalgamationGenerator(object):
         return amalgamation_sources
 
 
-def detect_compiler_version(ccinfo, cc_bin, os_name):
-    # pylint: disable=too-many-locals
-
-    cc_version_flag = {
-        'msvc': ([], r'Compiler Version ([0-9]+).([0-9]+).[0-9\.]+ for'),
-        'gcc': (['-v'], r'gcc version ([0-9]+.[0-9])+.[0-9]+'),
-        'clang': (['-v'], r'clang version ([0-9]+.[0-9])[ \.]')
+class CompilerDetector(object):
+    _version_flags = {
+        'msvc': [],
+        'gcc': ['-v'],
+        'clang': ['-v'],
+    }
+    _version_patterns = {
+        'msvc': r'Compiler Version ([0-9]+).([0-9]+).[0-9\.]+ for',
+        'gcc': r'gcc version ([0-9]+.[0-9])+.[0-9]+',
+        'clang': r'clang version ([0-9]+.[0-9])[ \.]',
     }
 
-    cc_name = ccinfo.basename
-    if cc_name not in cc_version_flag.keys():
-        logging.info("No compiler version detection available for %s" % (cc_name))
-        return None
+    def __init__(self, cc_name, cc_bin, os_name):
+        self._cc_name = cc_name
+        self._cc_bin = cc_bin
+        self._os_name = os_name
 
-    (flags, version_re_str) = cc_version_flag[cc_name]
-    cc_cmd = cc_bin.split(' ') + flags
+    def get_version(self):
+        try:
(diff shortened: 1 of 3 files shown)


randombit:
Nice, thank you!

PR #1135, opened 2017-07-30
FFI: Add interface for key wrapping with RFC 3394

flowher:
This is a contribution from Ribose Inc (@riboseinc).


flowher:
We need this in ``rnp`` in order to be able to perform ECDH.

flowher:
Hmm.... travis fails, but it doesn't seem to be related

PR #1136, opened 2017-07-31
Some thoughts on compiler version detection

webmaster128:
Different updates for compiler version detection, most described well in the commit message


randombit:
LGTM thanks

PR #1137, opened 2017-07-31
Add fast ChaCha based RNG

randombit:
Something like this (but less helpful) was already implemented twice, once for NewHope speed tests and another time in the fuzzer code. I'm not sure how many applications necessarily need a very fast PRNG but it's probably better to just provide one rather than have someone decide HMAC_DRBG is too slow for their needs and then make up something on their own.

Note: I made this up on my own. ;) But it's a pretty natural extract-then-expand KDF using HMAC/SHA-256 as the extractor function and ChaCha as the expansion function


randombit:
@securitykernel In 44965d7 I refactored the HMAC_DRBG unit tests to test any `Stateful_RNG` subclass and would like your input.

To simplify the code a bit I build-time required SHA-256 for the HMAC_DRBG unit tests (since without it already almost all of the HMAC_DRBG tests were skipped at runtime).

Also I noticed `test_reseed_kat` did not behave as expected - it generated an output then compared it with the fixed value `output_without_reseed` to ensure they were different ie that a reseed had occurred. But the `output_without_reseed` var had a different size than the generated block, so even if no reseed happened the two values would always be "different". I changed the test so `Request_Counting_RNG` always produces fixed output (0x808080...) and changed `test_reseed_kat` so that it only gets access to the counting RNG (instead of also giving it access to the actual entropy sources), then tests the output "after reseed" for equality with the expected result (with a reseed of 808080...) instead of non-equality with what the result would have been, lacking a reseed. Make sense?

securitykernel:
Probably a test bug. The new code makes more sense.

randombit:
Merged in 8a29dc8

Issue #1109, opened 2017-07-03
http_util and OCSP online checks


securitykernel:
During a source review for fixing #1107, I figured that OCSP online checks are guarded by `#if defined(BOTAN_HAS_ONLINE_REVOCATION_CHECKS)` in [x509_path.cpp](src/lib/x509/x509path.cpp). `BOTAN_HAS_ONLINE_REVOCATION_CHECKS` is defined in [x509_path.h](src/lib/x509/x509path.h):

```
#if defined(BOTAN_TARGET_OS_HAS_THREADS) && defined(BOTAN_HAS_HTTP_UTIL)
  #define BOTAN_HAS_ONLINE_REVOCATION_CHECKS
#endif
```

We can assume that nearly every OS has threading support, so it's only about http_util. That means if module http_util is not enabled in the build, OCSP online checks will be silently omitted. In a standard build, http_util will be enabled. But when using one of the module policies bsi, modern or nist, http_util will not be enabled. Can we do better?

I thought about setting `load_on -> auto` for http_util, which would always enable the module unless --no-autoload or --minimized-build is set. http_util does not have any OS dependencies in info.txt, but instead provides an implementation using boost asio (`BOTAN_HAS_BOOST_ASIO`) and Unix sockets (`BOTAN_TARGET_OS_HAS_SOCKETS`). If neither is available, the functions will throw an exception at runtime. For Windows builds without boost though, X.509 path validation will fail when OCSP online checks are made, which is maybe not good behaviour. Is there any alternative solution?

slicer4ever:
I've modified http_util.cpp to support windows by changing the following(built for msvc17):

#elif defined(BOTAN_TARGET_OS_HAS_SOCKETS)
#ifdef _WIN32
#include <winsock2.h>
#include <WS2tcpip.h>

#define close closesocket

int write(int SockFD, const char *Buf, size_t Len) {
	return send(SockFD, Buf, Len, 0);
}

int read(int SockFD, char *Buf, size_t BufLen) {
	return recv(SockFD, Buf, (int)BufLen, 0);
}

typedef size_t ssize_t;
#else
  #include <sys/types.h>
  #include <sys/socket.h>
  #include <netdb.h>
  #include <unistd.h>
  #include <netinet/in.h>
#endif
#else
  //#warning "No network support enabled in http_util"
#endif

securitykernel:
[...]

PR #1138, opened 2017-08-04
Add support for Windows sockets to http_util

securitykernel:
Based on the work by @slicer4ever suggested in #1109, adds support for Windows sockets to http_util. As a bonus, we get Windows support for tls_client and tls_server CLI.

Edit: Successfully tested tls_client and tls_server CLI against each other on a Windows 10 system. Also confirmed the OCSP online checks run successfully using `botan-test ocsp --run-online-tests`.

### DIFF
--- a/src/build-data/os/windows.txt
+++ b/src/build-data/os/windows.txt
@@ -23,6 +23,7 @@ rtlsecurezeromemory
 stl_filesystem_msvc
 threads
 filesystem
+sockets
 </target_features>
 
 <aliases>
--- a/src/cli/tls_client.cpp
+++ b/src/cli/tls_client.cpp
@@ -1,6 +1,7 @@
 /*
 * (C) 2014,2015 Jack Lloyd
 *     2016 Matthias Gierlings
+*     2017 René Korthaus, Rohde & Schwarz Cybersecurity
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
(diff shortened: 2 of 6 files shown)


randombit:
Nice, thanks. Question on `WSACleanup` calls - will this cause problems if the application is independently using sockets for other purposes? Or do things nest properly with the `WSAStartup`?

AppVeyor build is failing with duplicate symbol for `close`

```
botan.lib(utils_http_util.obj) : error LNK2005: "int __cdecl close(int)" (?close@@YAHH@Z) already defined in tls_client.obj
botan.lib(utils_http_util.obj) : error LNK2005: "int __cdecl read(int,void *,unsigned __int64)" (?read@@YAHHPEAX_K@Z) already defined in tls_client.obj
```

I think putting the declarations in `http_util.cpp` in an anon namespace will fix that.

slicer4ever:
WSAStartup and WSACleanup increment/decrements a counter when they have already been called.  ideally their should be no issue in calling each multiple time.  the only concern might be the extra costs when tearing down/starting up repeatable if the application itself never called them.

randombit:
@slicer4ever OK thanks for clarifying.

@securitykernel can you add a fix for the duplicate symbol issue? Otherwise this seems good to merge.

securitykernel:
Added a fix for the duplicate symbol in 0791842.

securitykernel:
And another failed Windows build, now in the amalgamation. Seems to be some Windows-specific include ordering problem, going with the solution described in https://stackoverflow.com/a/1372836.

PR #1139, opened 2017-08-04
Turn --destdir option into an environment variable

### LINKED ISSUE #996: destdir vs prefix confusion

### LINKED ISSUE #1101: Install path ignores --prefix when --destdir is set

webmaster128:
This is my approach on #1111, including my latest remarks and handing a lot of tests for edge cases I saw during review.

Most of that work is from @danimo and @zgyarmati, with just refinements of the implementation, especially for error handling.

Closes #996, closes #1101, closes #1111


webmaster128:
CI fail here is probably unrelated to the diff

randombit:
LGTM merging thanks for all the review and work on this.

PR #1147, opened 2017-08-09
ECDSA: Reduce message size to base point order before sign/verify

flowher:
Currently deterministic version ECDSA doesn't support messages to be sign which are longer than order of a base point. This is because size of a buffer ``m_rng_in`` in rfc6979.cpp (``RFC6979_Nonce_Generator::nonce_for``) is set to 2*order_length. So if digest to be signed is longer than a order length - it won't fit into buffer.

FIPS 186-3, point 6.4 says that "When the length of the output of the hash function is greater than the bit length of n, then the leftmost n bits of the hash function output block shall  be used in any calculation using the hash function output during the generation or verification of a digital signature"

Similarly RFC 6979 says in 2.3.2 that "if qlen < blen then the qlen left most bits are kept" - this should be done before converting ``msg`` to ``BitInt`` (spec calls this function ``bits2int``).

Proposed patch makes sure that message to be sign has required length.

### DIFF
--- a/src/lib/pubkey/ecdsa/ecdsa.cpp
+++ b/src/lib/pubkey/ecdsa/ecdsa.cpp
@@ -77,8 +77,8 @@ secure_vector<uint8_t>
 ECDSA_Signature_Operation::raw_sign(const uint8_t msg[], size_t msg_len,
                                     RandomNumberGenerator& rng)
    {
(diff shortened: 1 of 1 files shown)


flowher:
@randombit Can you comment if you agree with the patch?
There are tests missing - I'll try to produce vectors with BouncyCastle today.

PR #1149, opened 2017-08-13
FFI test must search test data relative to data dir.

bluhm:
The Test::data_file() method adds the current data dir prefix to
the test data path so that it can be overwritten with the --data-dir
option.  This was missing in the ffi test and could result in a
botan_x509_cert_load_file exception.


randombit:
Thanks!

PR #1153, opened 2017-08-15
Allow signature using `Raw(hashname)`

randombit:
This confirms the message is exactly the size of the expected hash, and also causes RFC 6979 nonces to be generated using the specified hash. See also https://github.com/riboseinc/rnp/issues/367

@flowher This look ok to you?


flowher:
Perfect!

flowher:
Actually, hmmm - I think I should be allowed to use combination where base point order is smaller than hash size (leftmost bitlength(order) should be used). But if I try to do it with this implementation then I get error "Invalid argument Encoding error: encode_1363: n is too large to encode properly" - that's because ``m_rlen`` in ``rfc6979.cpp`` depends on order size.

Ok, so then as a user of ecdsa, I can hash message myself and then reduce the size. But then I get an error in ``botan_pk_op_sign_finish`` as now my hash is too small (see ``encoding_of`` as implemented in this patch).

randombit:
@flowher I think you're right - probably #986 is the right start in this direction but I need to correct it to match the standard behavior.

flowher:
Definitely right start

PR #1154, opened 2017-08-18
Add guards for MAC and hash algo used in passhash9 tests

securitykernel:
Previously, passhash9 tests would fail if for example blowfish was not part of the build. Adds guards for the different MAC and hash algorithms used in passhash9 for calls to `generate_passhash9()`. For `check_passhash9()`, there is no way to know that an algorithm is not supported, but at least
we improve the situation a bit here.


randombit:
Would it be better to add a new library API `bool is_passhash9_alg_supported(uint8_t id)` to encapsulate this logic? It seems like an application might want to be able to do this very same test.

securitykernel:
Seems like a good idea, added that in 51f92a0.

PR #1157, opened 2017-08-22
Add ARIA

randombit:
Based on @noloader's patch in #1004 with some cleanups in 2ac367f

I enabled the ARIA TLS ciphersuites from RFC 6209 in fa477de but hilariously? I can't find even a single host that will negotiate it, even `kisa.or.kr` refuses it. So I will probably just drop that commit.

Edit: I disabled ARIA TLS again.


noloader:
> I enabled the ARIA TLS ciphersuites from RFC 6209 ...  even kisa.or.kr refuses it

Ouch! That is kind of odd.

> So I will probably just drop that commit.

You might want to make it optional.

> Edit: I disabled ARIA TLS again.

Yeah, this sounds like a good idea. I'm guessing an option `--enable-aria` would be a good idea. As an option, folks who don't need the support don't carry around the extra baggage.

randombit:
@noloader There is actually basically zero extra code size cost to adding the ARIA TLS suites, if ARIA itself is enabled. It's just adds some entries in the ciphersuite table. But I don't want to ship something that can't be tested. And if even KISA isn't using ARIA ciphersuites I'm guessing the demand for RFC 6209 ciphersuites is precisely 0.

I guess no reason not to add a `--with-aria` option to `src/scripts/tls_suite_info.py` (which is what generates the ciphersuite table) - we already do that for other unusual modes like OCB, CECPQ1 etc

noloader:
> But I don't want to ship something that can't be tested.

Yeah, good point.

Related, OpenSSL recently added the block cipher at https://github.com/openssl/openssl/pull/3242.  You might be able to use OpenSSL master and setup a local server for testing with `s_server`.

Or, shelve it for a couple of years until OpenSSL trickles into distributions. Once the newer OpenSSL is adopted it should be easier to perform the testing.

randombit:
> Related, OpenSSL recently added the block cipher

So far at least OpenSSL has only added ARIA cipher but not the TLS ciphersuites.

hided62:
It doesn’t seem easy to find a TLS server supporting the ARIA cipher that is accessible by the public.
ARIA is used in some VoIP solutions and SSL VPNs used by the Korean government.
But, ARIA has been included in the Bouncy Castle library as TLS cipher-suites recently. 
It seems taht it will be possible to test with a TLS server running the Bouncy Castle v. 1.57 or later in the near future.

letrhee-nsr:
It doesn’t seem easy to find a TLS server supporting the ARIA cipher that is accessible by the public.
ARIA is used in some VoIP solutions and SSL VPNs used by the Korean government.
But, ARIA has been included in the Bouncy Castle library as TLS cipher-suites recently. 
It seems taht it will be possible to test with a TLS server running the Bouncy Castle v. 1.57 or later in the near future.

randombit:
@letrhee-nsr Thanks for the hint on this, I will try testing with ARIA ciphersuites against Bouncycastle, if that goes well I will enable the ARIA suites in a future PR.

PR #1164, opened 2017-08-28
Tweak Travis config

randombit:
According to the docs in https://docs.travis-ci.com/user/multi-os/ this should work, though the examples in that doc don't pass the validator at https://lint.travis-ci.org/ which doesn't seem promising.


randombit:
This apparently doesn't work at all

PR #1175, opened 2017-09-01
Add ffi flags for PK ops signature format

securitykernel:
ECDSA in [strongswan](https://github.com/strongswan/strongswan/blob/master/src/libstrongswan/credentials/keys/public_key.h#L92) supports the Ecdsa-Sig-Value signature format according to [RFC 3279](https://www.ietf.org/rfc/rfc3279.txt), which is called `DER_SEQUENCE` in botan, as well as just the concatenation of r and s according to [RFC 4754](https://tools.ietf.org/html/rfc4754), which is called `IEEE_1363` in botan. Adds flags to ffi's `botan_pk_op_sign_create()` and `botan_pk_op_verify_create()` and updates the tests.

On that note, I stumbled upon the problem of knowing the signature size beforehand, which was just short-circuited for the tests. I did not want to hard-code the signature size for the `DER_SEQUENCE` formatted signature, too, so I just call update() and finish() twice, but that means any real signer would also need to do the computationally intensive signature operation twice. Could we add a function `botan_pk_op_sign_output_length()`, similar to the hash and hmac output length helper functions?

/cc @kolal23

### DIFF
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -1015,6 +1015,9 @@ BOTAN_PUBLIC_API(2,0) int botan_pk_op_decrypt(botan_pk_op_decrypt_t op,
                                   uint8_t out[], size_t* out_len,
                                   const uint8_t ciphertext[], size_t ciphertext_len);
 
+#define BOTAN_SIGNATURE_FORMAT_IEEE_1363 0
+#define BOTAN_SIGNATURE_FORMAT_DER_SEQUENCE 1
+
 /*
 * Signature Generation
 */
--- a/src/lib/ffi/ffi_pk_op.cpp
+++ b/src/lib/ffi/ffi_pk_op.cpp
@@ -103,10 +103,23 @@ int botan_pk_op_sign_create(botan_pk_op_sign_t* op,
 
       *op = nullptr;
 
-      if(flags != 0)
-         return BOTAN_FFI_ERROR_BAD_FLAG;
+      // before introducing the flags, 0 defaulted to IEEE_1363
+      Botan::Signature_Format sig_format = Botan::IEEE_1363;
 
-      std::unique_ptr<Botan::PK_Signer> pk(new Botan::PK_Signer(safe_get(key_obj),Botan::system_rng(),  hash));
+      if(flags == BOTAN_SIGNATURE_FORMAT_IEEE_1363 || flags == 0)
+         {
+         sig_format = Botan::IEEE_1363;
+         }
+      else if(flags == BOTAN_SIGNATURE_FORMAT_DER_SEQUENCE)
+         {
+         sig_format = Botan::DER_SEQUENCE;
+         }
+      else
+         {
+         return BOTAN_FFI_ERROR_BAD_FLAG;
+         }
+
+      std::unique_ptr<Botan::PK_Signer> pk(new Botan::PK_Signer(safe_get(key_obj), Botan::system_rng(), hash, sig_format));
       *op = new botan_pk_op_sign_struct(pk.release());
       return BOTAN_FFI_SUCCESS;
       });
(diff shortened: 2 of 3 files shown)


randombit:
Patch looks good.

> Could we add a function botan_pk_op_sign_output_length(), similar to the hash and hmac output length helper functions?

This would be great to have, it's a huge hole in the ffi API right now. It would require some changes outside of ffi to support it, basically:

- Add botan_pk_op_sign_output_length, which calls PK_Signer::signature_length()
- Implement PK_Signer::signature_length() which just calls PK_Ops::Signature::signature_length() on `m_op`
- Add the `virtual size_t PK_Ops::Signature::signature_length() const = 0`
- Then implement `signature_length` for all of the `Signature` subclasses (RSA, DSA, ECDSA, XMSS, ...). Most of these would be trivial I think.

To avoid having the additional complexity of considering the DER sequence encoding within each signature scheme that might support it (DSA, ECDSA, ECGDSA, ...) it seems better to have `PK_Signer` call the basic `m_op->signature_length()`, then if `m_sig_format == DER_SEQUENCE` just add the byte count for the expected DER encoding overhead based on `m_parts` and `m_part_size`. Or compute it by encoding dummy values of the appropriate size, though that seems a bit inelegant.

randombit:
Some tests are failing can you take a look

https://travis-ci.org/randombit/botan/jobs/270788064#L1623
https://travis-ci.org/randombit/botan/jobs/270788065#L1266

securitykernel:
Re `botan_pk_op_sign_output_length()`, thanks for the guiding. We will take upon this in another PR.

Re test failures, only occurs with openssl provider enabled. Also only reproducable in 2/10 runs. Debugging this.

securitykernel:
I can also reproduce the problem with the base provider now, it only occurs occassionally though. The problem is that `botan_pk_op_sign_finish()` aka `PK_Signer::signature()` sometimes returns different signature sizes with the same message for DSA and ECDSA with Signature_Format = DER_SEQUENCE. So in the ffi tests, the first call for ECDSA may return 102 bytes but the second call with the same message may return 103 bytes, which generates a `BOTAN_FFI_ERROR_INSUFFICIENT_BUFFER_SPACE` error in the second call, because the buffer passed is to small. `PK_Signer::signature()` decodes the signature parts r and s into a `std::vector<BigInt>` and then lets the ASN1 encoder do the encoding:

```cpp
const std::vector<uint8_t> sig = unlock(m_op->sign(rng));

...

std::vector<BigInt> sig_parts(m_parts);
for(size_t i = 0; i != sig_parts.size(); ++i)
   sig_parts[i].binary_decode(&sig[m_part_size*i], m_part_size);

return DER_Encoder()
   .start_cons(SEQUENCE)
   .encode_list(sig_parts)
   .end_cons()
   .get_contents_unlocked();
```

Now in `DER_Encoder::encode(const BigInt& n, ...)`, depending on the bit size of the passed `BigInt`, an extra zero byte is prepended, changing the size of the DER sequence encoding. Thus it may also happen that the second call will return a signature of a smaller size then in the first call to `PK_Signer::signature()`. This will also affect a potential future `PK_Signer::signature_length()`.

```cpp
DER_Encoder& DER_Encoder::encode(const BigInt& n,
                                 ASN1_Tag type_tag, ASN1_Tag class_tag)
   {
   if(n == 0)
      return add_object(type_tag, class_tag, 0);

   bool extra_zero = (n.bits() % 8 == 0);
   secure_vector<uint8_t> contents(extra_zero + n.bytes());

   ...
```

randombit:
I guess that isn't strictly required for verification, but it seems weird to support signature-as-DER flag on verification but not for signing...

securitykernel:
Thanks for your thoughts. I only need DER signatures in strongswan and it has its own ASN1 parser, I'll probably try DER encoding/decoding in strongswan instead.

randombit:
Yeah probably easiest thing to do here.

PR #1178, opened 2017-09-02
Use botan-ci-tools repo for packages

randombit:
Goal is to reduce dependencies on network resources at build time, and if possible make CI faster.


randombit:
Cuts about 24 minutes from the overall build time, nice.

Issue #842, opened 2017-01-16
Add ARM Carryless Multiply


noloader:
Attached and below is a patch for Carryless Multiplies using ARM Crypto extensions and `PMULL`. Its another partial patch, and others will have to complete it.

The dev-boards used for testing were a Pine64 and LeMaker HiKey. Both have CRC and Crypto extensions. Botan was configured with `./configure.py --cc={gcc|clang} --cc-abi="-march=armv8-a+crc+crypto -mtune=cortex-a53"`.

Here are the relative numbers:

* Botan without ARM `PMULL`, GCC 4.9.2 (Pine64):
```
pine64:botan$ ./botan speed --msec=3000 'GMAC(AES-128)'
GMAC(AES-128) [base] mac 4.647 MiB/sec (13.941 MiB in 3000.230 ms)
pine64:botan$ ./botan speed --msec=3000 'GMAC(AES-128)'
GMAC(AES-128) [base] mac 3.944 MiB/sec (11.832 MiB in 3000.084 ms)
```
* Botan without ARM `PMULL`, GCC 4.9.2 (HiKey):
```
hikey:botan$ ./botan speed --msec=3000 'GMAC(AES-128)'
GMAC(AES-128) [base] mac 7.167 MiB/sec (21.504 MiB in 3000.520 ms)
hikey:botan$ ./botan speed --msec=3000 'GMAC(AES-128)'
GMAC(AES-128) [base] mac 7.209 MiB/sec (21.629 MiB in 3000.265 ms)
```
* Botan without ARM `PMULL`, Clang 3.5.0 (Pine64):
```
pine64:botan$ ./botan speed --msec=3000 'GMAC(AES-128)'
GMAC(AES-128) [base] mac 3.443 MiB/sec (10.328 MiB in 3000.142 ms)
pine64:botan$ ./botan speed --msec=3000 'GMAC(AES-128)'
GMAC(AES-128) [base] mac 3.818 MiB/sec (11.453 MiB in 3000.100 ms)
```
* Botan without ARM `PMULL`, Clang 3.5.0 (HiKey):
```
hikey:botan$ ./botan speed --msec=3000 'GMAC(AES-128)'
GMAC(AES-128) [base] mac 6.360 MiB/sec (19.082 MiB in 3000.461 ms)
hikey:botan$ ./botan speed --msec=3000 'GMAC(AES-128)'
GMAC(AES-128) [base] mac 6.361 MiB/sec (19.086 MiB in 3000.274 ms)
```

[...]

PR #1181, opened 2017-09-03
Add support for ARMv8 PMULL instruction

randombit:
Contributed in GH #842 thanks @noloader 

Only tested under qemu so far, need to check hardware.


randombit:
Tests fine on an AMD Opteron 1100 with GCC 7.1, GMAC perf went from 20 Mib -> 168 Mib

Issue #1146, opened 2017-08-08
Add ARMv8 AES


noloader:
Attached and below is a patch for AES using ARMv8 extensions. Its another partial patch, and it hijacks the C++ implementation. Others will have to complete it. 

The code came from a few sources, including Crypto++ and mbedTLS experimental implementation by Johannes Schneiders, Barry O'Rourke and Skip Hovsmith. An early paper by Cynthia Crutchfield was also used as a reference. mbedTLS  was helpful because it was a working implementation that allowed us to examine state under a debugger.

Here are the numbers from GCC117 (compile farm), which is a 2.0 GHz AMD Opteron. If my calculations are correct, Botan is pushing data at 1.4 cpb for AES-128 on the Opteron (2.048/1.414).

* AMD Opteron @2.0 GHz with ARMv8 AES extensions

```
$  ./botan speed --msec=3000 AES-128 AES-192 AES-256
AES-128 [base] encrypt buffer size 4096 bytes: 1414.382 MiB/sec (4243.188 MiB in 3000.029 ms)
AES-128 [base] decrypt buffer size 4096 bytes: 1414.546 MiB/sec (4243.688 MiB in 3000.034 ms)
AES-192 [base] encrypt buffer size 4096 bytes: 1086.721 MiB/sec (3260.188 MiB in 3000.022 ms)
AES-192 [base] decrypt buffer size 4096 bytes: 1086.507 MiB/sec (3259.562 MiB in 3000.039 ms)
AES-256 [base] encrypt buffer size 4096 bytes: 931.761 MiB/sec (2795.312 MiB in 3000.031 ms)
AES-256 [base] decrypt buffer size 4096 bytes: 931.845 MiB/sec (2795.562 MiB in 3000.031 ms)

algo                          operation  4096 bytes
AES-128 [base]                decrypt    1483259.41
AES-192 [base]                decrypt    1139284.89
AES-256 [base]                decrypt    977109.84
AES-128 [base]                encrypt    1483086.97
AES-192 [base]                encrypt    1139509.75
AES-256 [base]                encrypt    977022.56
```

* AMD Opteron @2.0 GHz without ARMv8 AES extensions

```
$ ./botan speed --msec=3000 AES-128 AES-192 AES-256
AES-128 [base] encrypt buffer size 4096 bytes: 100.720 MiB/sec (302.188 MiB in 3000.262 ms)
[...]

PR #1182, opened 2017-09-03
Add ARMv8 AES

randombit:
Based on the patch in GH #1146

Tested in qemu and Opteron 1100. On hardware it is roughly 17x faster than table lookups.

### DIFF
--- a/src/lib/block/aes/aes.cpp
+++ b/src/lib/block/aes/aes.cpp
@@ -412,6 +412,19 @@ void aes_key_schedule(const uint8_t key[], size_t length,
    DK.resize(length + 24);
    copy_mem(EK.data(), XEK.data(), EK.size());
    copy_mem(DK.data(), XDK.data(), DK.size());
+
+#if defined(BOTAN_HAS_AES_ARMV8)
+   if(CPUID::has_arm_aes())
+      {
+      // ARM needs the subkeys to be byte reversed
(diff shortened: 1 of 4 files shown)


noloader:
@randombit,

Based on related PPC testing with Power8 AES built-ins, we found that running 6x blocks rather than 4x blocks was profitable. 6x blocks achieved another 0.1 to 0.3 cpb, depending on the modes. Also [see here](https://github.com/weidai11/cryptopp/commit/25efb7a14092091af6619117a1ac098655207e6b).

You might want to increase the number of block processed in `aes_encrypt_n`  and `aes_decrypt_n`  to 6.

randombit:
With 32 NEON registers to work with it makes sense, no real register pressure and it hides instruction latencies a bit better. Thanks for the tip.

BTW I tried 6x unroll for AES-NI, for AES-128 we have enough xmm registers (6 blocks + 10 subkeys) that there is no spilling, but it doesn't seem to help any, at least on Skylake.

noloader:
> BTW I tried 6x unroll for AES-NI, for AES-128 we have enough xmm registers (6 blocks + 10 subkeys) that there is no spilling, but it doesn't seem to help any, at least on Skylake.

Ah, thanks. I was thinking about trying that. I know its an evolutionary dead-end.

PR #1184, opened 2017-09-04
Added SHA3_XXX_PKCS OID

fxdupont:
RSA PKCSv1.5 sign/verify does not work with SHA-3 because of these missing OIDs.
BTW as RSA PSS is supposed to be used instead it is not critical...


randombit:
Thanks!

PR #1199, opened 2017-09-09
Run Windows CI build under ci_build.py

randombit:
I added support for clcache (https://github.com/frerich/clcache) but unfortunately it seems useless for our needs because it calls a cache miss whenever we rewrite `build.h`, which is every commit. So it's not used in the AppVeyor build. But left the support in ci_build.py since it might be useful for someone running tests locally or in a buildbot instance where they can do multiple builds targets on a single cache/checkout.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1131,22 +1131,22 @@ class CompilerInfo(InfoObject): # pylint: disable=too-many-instance-attributes
 
         if options.with_coverage_info:
             if self.coverage_flags == '':
-                raise InternalError('No coverage handling for %s' % (self.basename))
+                raise UserError('No coverage handling for %s' % (self.basename))
             abi_link.append(self.coverage_flags)
 
         if options.with_sanitizers:
             if self.sanitizer_flags == '':
-                raise InternalError('No sanitizer handling for %s' % (self.basename))
+                raise UserError('No sanitizer handling for %s' % (self.basename))
             abi_link.append(self.sanitizer_flags)
 
         if options.with_openmp:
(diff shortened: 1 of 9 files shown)


webmaster128:
> but unfortunately it seems useless for our needs because it calls a cache miss whenever we rewrite build.h, which is every commit

There is two modes: non-direct mode compares preprocessed sources and direct mode compares contents of sources and includes. Both modes work as long as build.h contains the same content after re-creating.

randombit:
> Both modes work as long as build.h contains the same content after re-creating.

The problem is `#define BOTAN_VERSION_VC_REVISION ...` in `build.h`, which changes on every new revision. So every file missed with 'header changed'. However I guess the problem may be I set direct mode, and clcache doesn't back down to preprocessor mode on a direct miss? I'll try skipping direct mode and seeing if that helps.

webmaster128:
> However I guess the problem may be I set direct mode, and clcache doesn't back down to preprocessor mode on a direct miss?

As far as I can see in the code, it is not intended to fall back from direct mode to preprocessor mode. This is as far as I can see because the case that the source changed but preprocessed output does not change is rare and not considered. Only the other way is covered (source unchanged but preprocessed changes e.g. using __TIME__).

Thus setting CLCACHE_NODIRECT should help for all compilation units that do not expand BOTAN_VERSION_VC_REVISION when preprocessing.

randombit:
Setting `CLCACHE_NODIRECT`, I see cache hits. Thanks. Testing this now.

PR #1205, opened 2017-09-11
Support large block sizes in OCB

randombit:
This follows draft-krovetz-ocb-wide-d3 (which it seems is still not public but will eventually update RFC 7253) to support larger block ciphers in OCB mode. Specifically 192 bit (that's mostly for the tests, may be removed before merging), 256-bit (SHACAL2) and 512-bit (Threefish).

Not ready to merge, there are a couple of areas (marked with `#if 0` blocks) where the spec and the reference implementation diverge and I'm waiting for clarification on these. Just want to get it up for CI to churn.


randombit:
Waiting on final confirmation of test vectors here

PR #1208, opened 2017-09-13
Remove inner/outer OID check in EC_PrivateKey ctor

securitykernel:
RFC 5915 mandates that the OID of an ECDSA ECPrivateKey shall be id-ecPublicKey with the named curve as the parameters. ECPrivateKey may contain the named curve OID, too, which is compared to id-ecPublicKey. Such keys could not be loaded. We remove this check and add a test vector from strongswan.


randombit:
Nice thanks.

5915 also has

> Though the ASN.1 indicates that the parameters field is OPTIONAL, implementations that conform to this document MUST always include the parameters field.

We currently omit the parameters on encoding (in `EC_PrivateKey::private_key_bits`), so we don't conform to 5915 here.

securitykernel:
Yeah, I saw that one. Shall we fix this right away?

randombit:
Might as well. I had some memory that the reason we were not was because 1.10 did not understand these extra fields, and I didn't want 2.x to produce something that 1.10 could not parse. But it looks like 1.10 has ignored the optional fields since 659e7852  (released in 1.10.3 in 2012), so no problems there.

randombit:
I'm going to merge this now, feel free to take on the encoding side if you like or I'll get to it sometime in the next couple of weeks before 2.3.0

briansmith:
> Nice thanks.
> 
> 5915 also has
> 
> > Though the ASN.1 indicates that the parameters field is OPTIONAL, implementations that conform to this document MUST always include the parameters field.
>
> We currently omit the parameters on encoding (in EC_PrivateKey::private_key_bits), so we don't conform to 5915 here.

See https://www.rfc-editor.org/errata_search.php?rfc=5915&eid=5008.

randombit:
Since effectively the only use of `private_key_bits` is to shove it into a PKCS8 structure, and PKCS11 comparability is rather desirable, we probably should skip on the optional/mandatory parameters. Thanks for the link @briansmith

Issue #1207, opened 2017-09-12
GCM tag length restrictions


securitykernel:
I think we had a discussion about this topic once, but I could not find anything in issues, PRs, commits or by e-mail. Botan currently restricts the tag length for GCM to [either 64 bits or 128 bits](https://github.com/randombit/botan/blob/master/src/lib/modes/aead/gcm/gcm.cpp#L185).

```cpp
   if(m_tag_size != 8 && m_tag_size != GCM_BS)
      throw Invalid_Argument(name() + ": Bad tag size " + std::to_string(m_tag_size));
```

[BSI TR-02102, sec. 2.1.2](https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TG02102/BSI-TR-02102-1.pdf?__blob=publicationFile&v=6) recommends a tag length of at least 96 bits. Also NIST recommends any of the values 128, 120, 112, 104, or 96. Was there a particular reason for only allowing 64 and 128 bits?

Edit: recommends a tag length of **at least** 96 bits
Edit2: NIST recommends **any of the values 128, 120, 112, 104, or 96**

randombit:
I do recall this coming up at one point also, not sure where/when.

And wow 64-bit GCM tags are a terrible idea, don't know why I would have added that in the first place. Anyway if we're going to support truncated tags at all it would be best to allow the 96...128 range. Just a minor code change, would need to dig up some official test vectors if possible.

securitykernel:
The reason this came up now is that we are currently implementing AES-GCM via botan in strongswan and strongswan supports GCM with 64 bits, 96 bits and 128 bits tags as per [RFC 5282](https://tools.ietf.org/html/rfc5282). A plugin can register each of them separately, so we can just support 96 and 128 bits and not 64 bits in the botan plugin for strongswan.

Regarding test vectors: strongswan uses the official NIST test vectors and just [truncates the tag](https://github.com/strongswan/strongswan/blob/master/src/libstrongswan/plugins/test_vectors/test_vectors/aes_gcm.c), can't we do so, too?

randombit:
Ah yes I forgot that GCM truncated tags don't effect the computation in any way, so just truncating an existing test is fine.

randombit:
With #1209 and #1210 merged I think we are all set here.

PR #1210, opened 2017-09-14
Allow GCM tag size in the range 96..128 bits

naskata:
The idea was to actually not support 8 byte tags and only allow 12..16 byte tags (#1207)


randombit:
I think while using them is not a great idea, we probably shouldn't immediately remove support for 64-bit GCM tags right away to avoid breaking any downstream applications which might be using them. Adding 96...128 range is fine. We can deprecate 64-bit tags now (just added it in `doc/deprecated.txt`) and maybe remove it after a suitable period.

naskata:
OK, 64-bit tags are reenabled with a comment that those are deprecated.

randombit:
Thanks

Issue #1211, opened 2017-09-14
Raw RSA PSS with salt


bellgrim:
I am trying to implement CKM_RSA_PKCS_PSS from PKCS#11 using Botan, but I cannot figure out which EMSA string I should use when calling PK_Signer().

> The PKCS #1 RSA PSS mechanism, denoted CKM_RSA_PKCS_PSS, is a mechanism based on the RSA public-key cryptosystem and the PSS block format defined in PKCS #1.  It supports single-part signature generation and verification without message recovery. This mechanism corresponds only to the part of PKCS #1 that involves block formatting and RSA, given a hash value; it does not compute a hash value on the message to be signed.

> It has a parameter, a CK_RSA_PKCS_PSS_PARAMS structure. The sLen field must be less than or equal to k*-2-hLen and hLen is the length of the input to the C_Sign or C_Verify function. k* is the length in bytes of the RSA modulus, except if the length in bits of the RSA modulus is one more than a multiple of 8, in which case k* is one less than the length in bytes of the RSA modulus.

So the input / message is already hashed by the calling application. "EMSA4(Raw)" will not use the salt length. If I use e.g. "EMSA4(SHA-256,MGF1,sLen)", then the message will be hashed again.

Is there an EMSA string that will not hash the incoming message again and just add the PSS padding with salt?

neusdan:
Unfortunately no. 

This is currently hard coded to a salt length of zero as you can see here:

https://github.com/randombit/botan/blob/751a3d36bd6072a1e042c8c2c6662fc7ae7535be/src/lib/prov/pkcs11/p11_mechanism.cpp#L24

I think we need to extend 

https://github.com/randombit/botan/blob/751a3d36bd6072a1e042c8c2c6662fc7ae7535be/src/lib/prov/pkcs11/p11_mechanism.cpp#L73

and the `RSA_SignMechanism` ctor.

Without patching Botan i can see two alternatives at the moment: 

- Padding in software with Botan and using "Raw" RSA as PKCS#11 mechanism.

- Using the Botan PKCS#11 low level API directly as you can see here:

https://github.com/randombit/botan/blob/751a3d36bd6072a1e042c8c2c6662fc7ae7535be/src/lib/prov/pkcs11/p11_rsa.cpp#L201

randombit:
[...]

PR #1212, opened 2017-09-16
Support PSS signing of raw hashes

randombit:
Allows PSS-signing a raw hash while also still specifying the salt length. GH #1211

Adds a full set of tests for both PSS (which was tested but not so well) and the raw variant, using the NIST test vectors.

@bellgrim Can you take a look and make sure this is doing what you need? Basically `PSSR_Raw(hFunc,MGF1,sLen)` will consume a raw hash, verify the input hash matches the hash function output length, generate a random salt of the desired length, then perform the inner H(msg || salt) computation.


bellgrim:
Thanks! Will have a look at this on Wednesday.

bellgrim:
This has now been implemented in https://github.com/opendnssec/SoftHSMv2/pull/340

The tests can sign and verify data by using `PSSR_Raw(hFunc,MGF1,sLen)`. It is working as expected!

randombit:
Hurah, merging

Issue #1222, opened 2017-09-25
Investigate RSA side channel from Usenix Security


randombit:
Looking through papers from this years Usenix Security I come across an interesting one on identifying cache-based timing channels. 

https://www.usenix.org/conference/usenixsecurity17/technical-sessions/presentation/wang-shuai

Hm I wonder what bugs they found... from the abstract "Moreover, we have successfully discovered previously unknown issues in two widely used cryptosystems, OpenSSL and Botan." well then!

They analyze 1.10.3 but the paper mentions "we notice that this vulnerability affects
several other versions of Botan, including 1.10.12, 1.10.11, and 1.11.33." (and so presumably 2.x also).

Haven't read the paper carefully enough to understand the issue yet but hopefully this can be addressed in time for 2.3

FTR I never received any contact about this ... shrug

randombit:
Assigne CVE-2017-14737

vpereira:
will this issue be backported to branch 1.10?

thank you

randombit:
Yes I plan to backport the patch and release 1.10.17 at the same time as 2.3.0

randombit:
Fixed in trunk, releases coming on Monday

flowher:
Any idea if they have released this CacheD tool ? I can't find it.

randombit:
I emailed the authors on that topic, they replied that it was not public yet and they were considering their options.

PR #1224, opened 2017-09-26
Use a side channel silent table look up in the Montgomery exponentiation

randombit:
Addressing #1222 - or at least I believe so.

~~Important todo: I haven't checked what kind of code GCC or Clang generates for `const_time_lookup`.~~ Added valgrind annotations and everything looks good, at least on x86-64 with recent GCC/Clang.

@jurajsomorovsky @neverhub @securitykernel @neusdan would love some additional eyes on this.


securitykernel:
Sorry, I don't have time to look into it this time.

jurajsomorovsky:
Oh, sorry for the slow response. I am not sure I would be helpful here (it is quite out of my competence) and currently have not that much time ...

PR #1232, opened 2017-10-03
Move socket implementation into module http_util

webmaster128:
This removes the requirement of linking socket libraries for applications that do not use http_util.

See also  #1138


randombit:
Yeah makes sense. Can you look at the Windows build?

webmaster128:
Thanks for the quick feedback. Build looks promising now. So I guess this is ready for final review

webmaster128:
Damn, there is an `#include <string.h>` missing for `strerror` and `errno` in the BSD_Socket implementation.

randombit:
Fixed

PR #1236, opened 2017-10-04
openssl: support libressl

alonbl:


randombit:
Thanks, had just realized that report was talking about the LibreSSL build not OpenSSL 1.1.0

PR #1239, opened 2017-10-07
ed25519 signing and verification using a Curve25519 key-pair

rafalsk:
Implemented functionallity which allows to perform ed25519 signing and verification using a Curve25519 key-pair.

Usage:
`	
       //sign
        Botan::PK_Signer signer(prevPrivKey25519, rng, "Pure");
	signer.update(msgv);
	std::vector<uint8_t> signature2 = signer.signature(rng);

	//verify
	Botan::PK_Verifier verifier(pubKey25519, "Pure");
	verifier.update(msgv);
	bool valid = verifier.check_signature(signature2);`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,3 +85,5 @@ src/contrib/perl-xs/*blib
 /*.pub
 /*.crt
 /*.txt
+/.vs/ProjectSettings.json
+/.vs/slnx.sqlite
--- a/src/lib/pubkey/curve25519/curve25519.cpp
+++ b/src/lib/pubkey/curve25519/curve25519.cpp
@@ -12,6 +12,7 @@
 #include <botan/der_enc.h>
 #include <botan/rng.h>
 #include <botan/sha2_64.h>
+#include <stdlib.h>
 
 
 namespace Botan {
@@ -73,9 +74,6 @@ namespace Botan {
 		{
 			if (sig_len != 64)
 				return false;
-
-			unsigned char signature[64];
-
 
 			bool ok = curve25519_verify(sig, m_key.public_key_bits().data(), m_msg.data(), m_msg.size());
 			m_msg.clear();
--- a/src/lib/pubkey/curve25519/curve25519.cpp
+++ b/src/lib/pubkey/curve25519/curve25519.cpp
@@ -6,12 +6,89 @@
 */
 
 #include <botan/curve25519.h>
+#include <botan/internal/ed25519_internal.h>
 #include <botan/internal/pk_ops_impl.h>
 #include <botan/ber_dec.h>
 #include <botan/der_enc.h>
 #include <botan/rng.h>
 
+
 namespace Botan {
+
+
+	/**
+	* Ed25519 signing operation ('C25519' - signs with a Curve25519 key)
+	*/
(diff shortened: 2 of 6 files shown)


randombit:
Thanks! It will take a while for me to review this. In the meantime can you check the CI failures, and also reformat? It appears some functions (like ge_scalarmult_base) got reindented. So it is hard to see what actually changed. There is a configuration for the `astyle` formatter (http://astyle.sourceforge.net/) in `src/configs`

rafalsk:
yes; I'll be taking a look into these matters.

randombit:
@rafalsk Sorry about taking so long to review this.

Big concern, do we even want this in the library? This is a documented approach for signing with x25519 (https://moderncrypto.org/mail-archive/curves/2014/000205.html) but is it implemented by any other library? If we're going to offer such conversions, going in the ed25519 to x25519 direction seems simpler and is implemented by libsodium (https://download.libsodium.org/doc/advanced/ed25519-curve25519.html) and its various wrappers. That's especially a concern as (AIUI) this signing approach is *not* compatible with standard Ed25519 signatures (please correct me If I'm wrong about this.)

Assuming we do want it in the library:
 - There are no tests, these need to be added, preferably with vectors created by an independent implementation.
 - This creates a new internal dependency where x25519 uses Ed25519. This should be optional, with the signing operations disabled if Ed25519 is not included in the build.
 - This adds functions `ed25519_verify_modified` and `ed25519_sign_modified` - how are they modified? Why can't these just do the conversion and call the versions in Ed25519? At the very least there should be a comment clearly explaining this, but overall I'd be a lot happier with the EdDSA algorithm not being duplicated in two places with slightly different implementations.

There are also minor style things (used of C APIs like malloc+free, not using stdint typedefs) but no point addressing those until the larger issues are cleared up.

rafalsk:
I've lots of work right now; I'll give some worthy response ASAP.

rafalsk:
Yes, yes we do want this into the library; the proliferation of x25519 keys seems to be much higher; I know I could go the easier way doing ed25519 to x25519. I'll document everything in detail and implement all the tests etc. at the beginning of January; too busy right now.

rafalsk:
I'm still on the subject. I'll try to find time to document and rationalize everything which has been implemented by me here in weeks to follow. I already do use the proposed functionality.

deemru:
Just tried your `rafalsk:ed25519sigWithCurve25519`, an example of [Waves transaction signing](https://github.com/wavesplatform/Waves/wiki/Cryptographic-practical-details#signing) did worked for me on verify their etalon signature. That is nice.

Secondly, `ed25519_verify` works the same but much better by design against `ed25519_verify_modified`, is there a reason for `_modified` versions?

rafalsk:
@deemru there was a reason; I promise to dive into this in days to follow. We'll need to bring this into production on hardware devices soon; so I'll need to verify this once again.

rafalsk:
The thing is at the heart of our project https://twitter.com/gridnetproject, it's been all validated, we've got a version upgraded to the recent version of Botan running locally, as soon as I find some time I'll update this. never failed on us, passing all tests.

randombit:
Supplanted by #2774

Issue #1220, opened 2017-09-23
Signed overflow warnings


randombit:
Probably nothing but would be good to figure out what's happening here and fix/silence them if we can.

gcc version 7.1.1 20170630 (GCC) on Linux x86-64

```
./src/lib/utils/barrier.cpp: In member function 'void Botan::Barrier::sync()':
./src/lib/utils/barrier.cpp:37:1: warning: assuming signed overflow does not occur when changing X +- C1 cmp C2 to X cmp C2 -+ C1 [-Wstrict-overflow]
./src/lib/utils/semaphore.cpp: In member function 'void Botan::Semaphore::release(std::size_t)':
./src/lib/utils/semaphore.cpp:43:1: warning: assuming signed overflow does not occur when changing X +- C1 cmp C2 to X cmp C2 -+ C1 [-Wstrict-overflow]
./src/lib/utils/semaphore.cpp: In member function 'void Botan::Semaphore::acquire()':
./src/lib/utils/semaphore.cpp:43:1: warning: assuming signed overflow does not occur when changing X +- C1 cmp C2 to X cmp C2 -+ C1 [-Wstrict-overflow]
./src/lib/utils/parsing.cpp: In function 'bool Botan::host_wildcard_match(const string&, const string&)':
./src/lib/utils/parsing.cpp:342:6: warning: assuming signed overflow does not occur when changing X +- C1 cmp C2 to X cmp C2 -+ C1 [-Wstrict-overflow]
 ```

randombit:
These were fixed by #1245

PR #1245, opened 2017-10-09
#1220 - fixed signed overflow warnings

LesnyRumcajs:


randombit:
Thanks for the PR, it would be great to address these warnings. Unfortunately the change seems to have broken `Threaded_Filter` (which IIRC is the only user of these classes currently), you can see this in the CI build where the `filter` tests seem to hang, presumably with a deadlock.

LesnyRumcajs:
My bad, I was working on my release version and current master in two different projects. It should work now.

PR #1246, opened 2017-10-09
Include cstdlib to make os_utils compile with clang.

bluhm:
Header cstdlib is needed to compile on OpenBSD 6.2 with clang 4.0.0.


randombit:
Thanks. Out of curiosity what symbol was missing?

bluhm:
This was the error message:

/usr/ports/pobj/botan2-2.3.0/Botan-2.3.0/src/lib/utils/os_utils.cpp:476:27: error: 
      no member named 'getenv' in the global namespace
   if(const char* env = ::getenv("BOTAN_MLOCK_POOL_SIZE"))
                        ~~^
1 error generated.

Issue #1113, opened 2017-07-06
X.509/ASN.1 charset handling


dewyatt:
As mentioned in the manual:

> Botan doesn’t currently support any of the Unicode variants used in ASN.1 (UTF-8, UCS-2, and UCS-4), any of which could be used for the fields in the DN. This could be problematic, particularly in Asia and other areas where non-ASCII characters are needed for most names. The UTF-8 and UCS-2 string types are accepted (in fact, UTF-8 is used when encoding much of the time), but if any of the characters included in the string are not in ISO 8859-1 (ie 0 ... 255), an exception will get thrown. Currently the ASN1_String type holds its data as ISO 8859-1 internally (regardless of local character set); this would have to be changed to hold UCS-2 or UCS-4 in order to support Unicode (also, many interfaces in the X.509 code would have to accept or return a std::wstring instead of a std::string).

As an example, trying to use asn1print or cert_info, etc, for the certificate [here](http://gostcrypto.com/tool-asn1.html) will cause an exception.

As a workaround I've generally been storing and outputting everything as UTF-8 and using `boost::locale` for conversions, something like the below:

```cpp
// used for getting the local encoding
std::locale loc(boost::locale::generator().generate(""));

//IA5String -> UTF-8 (should be no-op)
boost::locale::conv::between("abcd",
    "UTF-8",
    "646",
    boost::locale::conv::stop);

//UCS-2 to local encoding
boost::locale::conv::between("\x42\x30",
    std::use_facet<boost::locale::info>(loc).encoding(),
    "ucs-2",
    boost::locale::conv::stop);
```

There might be a nicer way, this is just some local hack I've been using.

randombit:
Would be nice to improve this area. I think the manuals comment about using `wstring` is bogus (yes, I wrote that) and it would be better to instead translate everything to UTF-8 `std::string`s.

I think (but am not certain) that C++11 `std::locale` is insufficient for this task. We'll have to use either Boost, iconv (on Unix/Linux) or `MultiByteToWideChar` and co on Windows.

reneme:
We encountered X.509 certificates with non-Latin1 unicode chars in the DN. Obviously they are not readable by Botan. Our workaround is not to use any such characters in the certificates of our application for now, but that's only a short term solution.

After briefly looking into the ASN.1 parser, I am having trouble to estimate what it takes to get UTF-8 support in there. If you could you give us a few pointers, we might be able to allot some resources into this. How do you estimate the amount of changes necessary for this?

[...]

PR #1250, opened 2017-10-11
[WIP] UTF-8 support for ASN.1 parsing and encoding

reneme:
This is an early take on building UTF-8 support into the ASN.1 implementation. I roughly followed the pointers given in #1113. There are a few new test cases reading ASN.1 objects with `UTF8_STRING` tag and an X.509 certificate containing cyrillic letters in its DN.

That's by no means ready, I just want to put it up here for early feedback.


randombit:
Took a quick look, seems right to me. Thanks for adding the tests.

randombit:
@reneme What did you have in mind for the rest of this PR? This looks ok to merge as is and I'm not sure what is left to do.

reneme:
I want to go through the thoughts you posted on #1113 -- help with that appreciated of course. But on the other hand I am planning to add tests for handling of the different ASN.1 string types (esp. UCS2). As far as I can see proper UCS2 support is not too hard to reach but isn't there yet.

src/lib/asn1/asn1_str.cpp
```diff
@@ -58,46 +61,69 @@ ASN1_Tag choose_encoding(const std::string& str,
...
+      const CharT *as_wide_char;
+      };
+
+   as_char = ucsX.data();
+   const size_t wide_char_count = ucsX.size() / sizeof(CharT);
+
+   using converter_t = std::codecvt_utf8<CharT, 0x10ffff, std::consume_header>;
+   std::wstring_convert<converter_t, CharT> convert;
+   return convert.to_bytes(as_wide_char, as_wide_char + wide_char_count);
+   }
```
reneme: That's more a request for comment. The STL actually brings a way to convert UTF/UCS back and forth. It's deprecated (without replacement) in C++17 though...

reneme:
@randombit 

I got UTF-8/16/32 conversions in the `<codecvt>` header to work on windows now (with a nasty little workaround). Despite the header being standardized in STL it is not available in older libstdc++ versions. That is the reason for the failing build on many Travis configs. Furthermore `<codecvt>` is deprecated in C++17. I think that is a dead-end.

Seems that many people use [Boost.Locale](http://www.boost.org/doc/libs/1_65_1/libs/locale/doc/html/index.html) as a replacement. Another option would be to use [UTF8-CPP](https://github.com/nemtrif/utfcpp) which is reasonably light-weight and header-only. It also supports manipulation and checking of UTF-8 strings which we don't need for now, AFAIKS. Third option would be to use operating system facilities ([iconv](http://man7.org/linux/man-pages/man3/iconv.3.html) for POSIX and [WideCharToMultiByte](https://msdn.microsoft.com/en-us/library/windows/desktop/dd374130(v=vs.85).aspx) which I would avoid for portability reasons. Also they seem to be pretty archaic to use.

I'll need your opinion here and potentially a few pointers how to integrate a library into botan.

Thanks!

randombit:
Ughghg. Messy.

That `codecvt` was deprecated in C++17 is no big deal, but it not being available in GCC 4.8 is a problem. Especially since Travis is still stuck on Ubuntu 14.04 and no compiler more recent than 4.8 is available in that release.

Boost.Locale+ICU seems really heavy for this task. Probably a useful library if you need to do a lot of i18n work. But it's big, and ICU has has had a number of scary bugs (3 code exec bugs just this year). And anyway we cannot add Boost as a hard dependency, so there is still the problem of what to do in a Boost-free build.

Leaning towards just implementing the conversions within the library (as we do already for Latin1<->UTF8 etc), then adding a fuzzer that compares the conversion results with a non-portable implementation (eg `iconv`) to ensure correctness. (For further thoughts on this area see https://github.com/randombit/botan/blob/master/doc/contributing.rst#library-dependencies).

Another option maybe - `mbstowcs` from POSIX/C99. It seems tied to locale though? And maybe doesn't help with UTF-32. Probably not useful.

randombit:
I'm working on adding conversion functions that cross-check with `iconv`.

reneme:
Amazing! I'll be out of office pretty much until the end of November and would have left this starving for another while. Thanks a lot for following up!

randombit:
Thanks for letting me know. I will probably try to get this PR merged before then.

One question - I notice the tests you added use little-endian encoding for UCS2/UCS4. But all references I can find (eg https://msdn.microsoft.com/en-us/library/windows/desktop/bb540793(v=vs.85).aspx and https://www-archive.mozilla.org/projects/security/pki/jss/javadoc/org/mozilla/jss/asn1/BMPString.html) say that big-endian is always used. I am having a hard time parsing the OpenSSL ASN.1 code but it appears it also uses big-endian for BMP_STRING and UNIVERSAL_STRING (eg the code in `asn1/a_mbstr.c`). Just checking if there is something I'm missing here, thanks!

reneme:
I assumed that the encoding should be able to deal with a Byte Order Mark at the start as I didn't find a reference clearly stating big-endian. Hence I just went for little-endian with respective BOM.

randombit:
The PKIX RFCs seem to be underspecified here, but https://www.cs.auckland.ac.nz/~pgut001/pubs/x509guide.txt states

> The correct format for BMPStrings is: big-endian 16-bit characters, no Unicode
byte order marks (BOMs), and no terminating null character (ISO 8825-1 section
8.20).

GnuTLS also seems to do big-endian only with no consideration of BOM.

reneme:
Agreed. If I get the time tomorrow, I'll try and incorporate the changes you posted in #1286.

randombit:
@reneme Check out #1287 if you have a moment

reneme:
@randombit #1287 looks very good to me. I added a test that parses a certificate containing `BMPString` in #1289 to be merged into your **asn1-utf8** branch. Thanks a lot for your help on this!

Closing this PR in favour of #1286, #1287 and #1289.

PR #1254, opened 2017-10-13
Add limits.h header for INT_MAX

alonbl:
Gentoo-Bug: https://bugs.gentoo.org/633468


randombit:
Thanks

PR #1255, opened 2017-10-14
Reduce AES to using a single T-table

randombit:
Should have significantly better characteristics wrt cache side channels, though it would be nice to verify this.

Also need to check performance across platforms to makes sure this won't hurt too badly. On Skylake, before (current master) ~10.2 cycles/byte, after ~13 cpb.

### DIFF
--- a/src/lib/block/aes/aes.cpp
+++ b/src/lib/block/aes/aes.cpp
@@ -1,6 +1,6 @@
 /*
 * AES
-* (C) 1999-2010,2015 Jack Lloyd
+* (C) 1999-2010,2015,2017 Jack Lloyd
 *
 * Based on the public domain reference implementation by Paulo Baretto
 *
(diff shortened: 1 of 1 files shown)


randombit:
Almost exactly the same perf as with 4 tables on Pi2, thanks to the ARMv7 barrel shifter.

randombit:
About 35-40% slowdown on POWER7, eg AES-128 ~162 Mb/s -> 120 Mb/s

randombit:
~25% slowdown on AMD Bobcat, ~10% slowdown on Westmere.

randombit:
25% on ARMv8 (X-Gene)

randombit:
I'm going to merge this and take the perf hit. Will try to make it back with support for NEON and VMX using a pshufb-style implementation as is already done with SSSE3.

PR #1262, opened 2017-10-17
GCM and CTR optimizations

randombit:


randombit:
On Skylake: CTR 1.5x GCM 2x GMAC 3x faster
On Westmere: CTR 1.3x GCM 1.5x GMAC 1.6x faster

randombit:
With PMULL changes, on Opteron 1110, GCM 1.7x  GMAC 2.5x faster

I think GCM is held back because AES isn't unrolled for ARMv8 right now

PR #1263, opened 2017-10-18
Add supported groups TLS extension (RFC 7919)

securitykernel:
Adds support for the supported groups TLS extension per RFC 7919. This extension allows negotiating DH groups for the key exchange, similar to the supported elliptic curves extension (in fact, it just renames the supported elliptic curves extension to supported groups and adds codepoints for 5 predefined DH groups).

* Adds the predefined ffdhe groups as named DL groups and extends the tests
* Renames `Supported_Elliptic_Curves` to `Supported_Groups` (and adds a typedef to the old name)
  * Adds codepoints to the TLS extension
  * Adds a function `dh_groups()` that returns the ffdhe groups
* Adds a new `TLS_Policy::allowed_groups()` that may contain ffdhe groups as well as elliptic curves
* Adds code for negotiating the ffhde group in the TLS server
* Adds a check for the negotiated ffdhe group to the TLS client

There are two things left over I think. First is server behaviour. A TLS client also sends a ciphersuite list, which may conflict with the supported groups list. Section 6.1 in the RFC has a good example. In such a case, servers should prefer supported groups preference over ciphersuite preference. Handling this would complicate ciphersuite selection quite a bit.

Second, what do we do with `TLS_Policy::dh_group()`, which was only used by the TLS server before.

*Edit:* Should we deprecate `allowed_ecc_curves()` in favor of `allowed_groups()`?


randombit:
Thanks. Overall looks good.

The behavior of the client if the server selects a non FFDHE group does not seem right to me though. Since the server does not send back a `supported_groups` extension, the client has no way of knowing if it is talking to a server that supports FFDHE or not. So if it talks to an old server, which selects a DH ciphersuite without taking the FFDHE extension into consideration then negotiation will fail. I think practically speaking the client must more or less take whatever the server offers, modulo size checks. Or, we can add a policy toggle to require enforcing the server send a known FFDHE group as you have here. But I don't think it can be on by default. because it breaks DH exchanges with all older servers.

Also, in the event a client does not send any DH groups in the supported groups extension, but does offer DH ciphersuites, `policy.choose_dh_group` will return an empty string and we will fail with an exception in `DL_Group` constructor. Instead in that case we should just select a DH group arbitrarily (ie call `Policy::dh_group()`)

> In such a case, servers should prefer supported groups preference over ciphersuite preference. Handling this would complicate ciphersuite selection quite a bit.

Fortunately 7919 says the server "MAY resolve the contradiction any way it sees fit" so I'd vote for continuing to take ciphersuite ordering as the primary indicator and only after a ciphersuite is chosen taking the relevant groups into consideration. This makes more sense to me, and also has the advantage that we don't need to do anything further. ;)

securitykernel:
Added the fallback to ` Policy::dh_group()` in 85c97aa. Re incompatible servers, I removed the check in the client entirely in 558b66b.

randombit:
LGTM, merging

Issue #1264, opened 2017-10-18
A test fails


ruipacheco:
I tried building Botan thus:

```
> python configure.py --minimized-build --enable-modules=zlib,md5,sha1,tls --single-amalgamation-file
> make
```

and botan-test failed on the following test:

```
X509 Usage ran 8 tests 1 FAILED
Failure 1: X509 Usage test_usage RSA with error False assertion 'RSA sign consistency check' (expression m == c) in raw_sign @botan_all.cpp:20428
```

Reverting back to an [old issue](https://github.com/randombit/botan/issues/922) I had here I tried creating the library with the following command:

`python configure.py --minimized-build --enable-modules=md5,sha1,system_rng,sha2_64,tls --single-amalgamation-file --amalgamation`

but make fails:

```
⟩ make
clang++ -fstack-protector -m64 -pthread -stdlib=libc++ -fPIC -fvisibility=hidden -std=c++11 -D_REENTRANT -O3 -Wall -Wextra -Wpedantic -Wshadow -Wstrict-aliasing -Wstrict-overflow=5 -Wcast-align -Wmissing-declarations -Wpointer-arith -Wcast-qual -Ibuild/include -c botan_all.cpp -o build/obj/lib/botan_all.o
botan_all.cpp:27496:67: warning: unused parameter 'version' [-Wunused-parameter]
Connection_Cipher_State::Connection_Cipher_State(Protocol_Version version,
                                                                  ^
botan_all.cpp:27501:55: warning: unused parameter 'uses_encrypt_then_mac' [-Wunused-parameter]
                                                 bool uses_encrypt_then_mac) :
                                                      ^
botan_all.cpp:30710:25: error: variable has incomplete type 'struct stat'
            struct stat stat_buf;
                        ^
botan_all.cpp:30710:20: note: forward declaration of 'stat'
            struct stat stat_buf;
                   ^
botan_all.cpp:30712:18: error: no member named 'stat' in the global namespace
            if(::stat(full_path.c_str(), &stat_buf) == -1)
               ~~^
botan_all.cpp:30905:13: error: no member named 'getpid' in the global namespace
   return ::getpid();
          ~~^
2 warnings and 3 errors generated.
make: *** [build/obj/lib/botan_all.o] Error 1
```

Finally, GitHub says the latest release is 2.3.0 but the Python script says it's 2.4.0:

[...]

PR #1265, opened 2017-10-19
Only skip unconditional includes when generating amalgamation

webmaster128:
Fixes the broken build on macOS described in #1264: `python configure.py --minimized-build --enable-modules=md5,sha1,system_rng,sha2_64,tls --single-amalgamation-file --amalgamation && make`


randombit:
Excellent thank you

PR #1267, opened 2017-10-19
Implements multithreading support for XMSS

mgierlings:
Implements multithreading support for XMSS. The tree hash function divides the tree structure into _n_ subtrees. _n_ is chosen based on the available threads reported through `std::thread::hardware_concurrency()` and the height of the target node. The roots of the subtrees are then used to calculate the root node of the parent tree.

Adding multithreading requires every thread working on a subtree to own a distinct instance of `XMSS_Hash`. To prevent the overhead for repeated allocations on the stack the `XMSS_Hash` instances are allocated in the main thread prior to parallelization and then passed via reference into the child threads. These changes make the XMSS WOTS API unsuitable for direct use by library users, thus the visibility of XMSS WOTS and related classes is changed to _internal_. Since use of XMSS WOTS outside of XMSS does not seem useful this change should not be of disadvantage to library users.

Exemplary benchmark before and after the multithreading update.
```
Single-threaded XMSS
algo                               operation  sec/op            op/sec           
XMSS_SHA2-256_W16_H10              sign       3.597482          0.277972         
XMSS_SHA2-512_W16_H10              sign       8.825816          0.113304         
XMSS_SHAKE128_W16_H10              sign       4.524819          0.221003         
XMSS_SHAKE256_W16_H10              sign       16.463382         0.060741         
XMSS_SHA2-256_W16_H10              verify     0.002079          481.092313       
XMSS_SHA2-512_W16_H10              verify     0.002443          409.308538       
XMSS_SHAKE128_W16_H10              verify     0.002393          417.887207       
XMSS_SHAKE256_W16_H10              verify     0.008211          121.780908     

Multi-threaded XMSS
algo                               operation  sec/op            op/sec           
XMSS_SHA2-256_W16_H10              sign       2.225256          0.449387         
XMSS_SHA2-512_W16_H10              sign       4.923947          0.203089         
XMSS_SHAKE128_W16_H10              sign       2.545100          0.392912         
XMSS_SHAKE256_W16_H10              sign       7.937064          0.125991         
XMSS_SHA2-256_W16_H10              verify     0.001944          514.432180       
XMSS_SHA2-512_W16_H10              verify     0.004581          218.276982       
XMSS_SHAKE128_W16_H10              verify     0.002550          392.205050       
XMSS_SHAKE256_W16_H10              verify     0.008982          111.339372   
```

### DIFF
--- a/news.rst
+++ b/news.rst
@@ -119,7 +119,7 @@ Version 2.4.0, Not Yet Released
 
 * Fixes for CMake build (GH #1251)
 
-* Avoid some signed overflow warnings (GH #1220 #1245)
+* Add multithreading support for XMSS
 
 Version 2.3.0, 2017-10-02
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/src/lib/pubkey/xmss/xmss.h
+++ b/src/lib/pubkey/xmss/xmss.h
@@ -1,7 +1,18 @@
-/*
- * XMSS
- * Includes XMSS headers.
- * (C) 2016 Matthias Gierlings
+/**
+ * @file xmss.h
+ * Includes headers required for Extended Hash-Based Signatures (XMSS)
+ * as described in [1]. @see Botan::XMSS_PublicKey and @see
+ * Botan::XMSS_PrivateKey for further
+ * details.
+ *
+ * <pre>
+ * [1] XMSS: Extended Hash-Based Signatures,
+ *     draft-itrf-cfrg-xmss-hash-based-signatures-06
+ *     Release: July 2016.
+ *     https://datatracker.ietf.org/doc/draft-irtf-cfrg-xmss-hash-based-signatures/?include_text=1
+ * </pre>
+ *
+ * (C) 2016,2017 Matthias Gierlings
  *
  * Botan is released under the Simplified BSD License (see license.txt)
  **/
--- a/src/lib/pubkey/xmss/xmss_common_ops.cpp
+++ b/src/lib/pubkey/xmss/xmss_common_ops.cpp
@@ -15,16 +15,17 @@ XMSS_Common_Ops::randomize_tree_hash(secure_vector<uint8_t>& result,
                                      const secure_vector<uint8_t>& left,
                                      const secure_vector<uint8_t>& right,
                                      XMSS_Address& adrs,
-                                     const secure_vector<uint8_t>& seed)
+                                     const secure_vector<uint8_t>& seed,
+                                     XMSS_Hash& hash)
    {
    adrs.set_key_mask_mode(XMSS_Address::Key_Mask::Key_Mode);
-   secure_vector<uint8_t> key { m_hash.prf(seed, adrs.bytes()) };
+   secure_vector<uint8_t> key { hash.prf(seed, adrs.bytes()) };
 
    adrs.set_key_mask_mode(XMSS_Address::Key_Mask::Mask_MSB_Mode);
-   secure_vector<uint8_t> bitmask_l { m_hash.prf(seed, adrs.bytes()) };
+   secure_vector<uint8_t> bitmask_l { hash.prf(seed, adrs.bytes()) };
 
    adrs.set_key_mask_mode(XMSS_Address::Key_Mask::Mask_LSB_Mode);
-   secure_vector<uint8_t> bitmask_r { m_hash.prf(seed, adrs.bytes()) };
+   secure_vector<uint8_t> bitmask_r { hash.prf(seed, adrs.bytes()) };
 
    BOTAN_ASSERT(bitmask_l.size() == left.size() &&
                 bitmask_r.size() == right.size(),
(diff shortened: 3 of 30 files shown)


randombit:
Overall this is really good.

We must guard any use of `std::thread`, `<thread>`, etc with a check for the define `BOTAN_TARGET_OS_HAS_THREADS`. That's because a few platforms we support (including IncludeOS and LLVM/Emscripten) don't have them. For platforms without threads XMSS should just work iteratively as before.

Also I think `hardware_concurrency` is not going to always be the right number to use here. Particular examples, on my i7-6700k (4 cores, 8 threads), I saw almost no speedup for XMSS with SHA-256 or SHAKE-128 as a result of your change. However with SHA-512 or SHAKE-256, it was 2-3x faster. When I patched the code to treat the machine as having 4 CPUs instead of 8, I saw good performance increases for all parameter sets (only tried H=10). And I ran some quick tests on a large POWER8 machine (160 CPUs, possibly with SMT I'm not sure) - using threads there actually reduced performance by 30 to 50%. I'd like to better understand what the performance characteristics of using threads will be on some other systems before moving forward.

mgierlings:
Thanks, I think adding the guards shouldn't be too much of an issue. I'll look into that. 

I agree on your concerns about `hardware_concurrency`. While this was the most straight forward, standard library compliant way to determine a reasonable number for parallel threads to be spawned, it will not do well in many scenarios as you pointed out. I do not own very exotic hardware, with the exception of one Raspberry Pi Model 1, all my computers are x86.

I suspect not only hyper-threading and massive physical core counts to be a problem, as you already observered. It can also imagine a CPU having hardware support for SHA2 or not having impact on which number of threads will perform best (which unfortunately I can not test).

I was actually thinking about writing some code to benchmark different thread configurations at runtime and then approximate the number of threads _n_. But even with sth. like this in place we might want to offer some way of limiting the number of maximum concurrent threads in botan somehow, maybe set a default limit at compile time and then allow to adjust this at runtime globally and/or individually per algorithm. We should definitly discuss possible solutions.

randombit:
@mgierlings Sorry about delay here. I think it's ok to merge this and we can continue to look at the correct number of cores to use, possibility of SIMD etc later. However some of the CI tests are crashing, specifically all of the MSVC 32-bit builds (https://ci.appveyor.com/project/randombit/botan/build/1.0.3212/job/2wrgecx5kntkduol), as well as the Wine cross build (https://travis-ci.org/randombit/botan/jobs/290675691). All crash during execution of the `aead` tests - which seems strange.

Also the amalgamation is failing to build https://travis-ci.org/randombit/botan/jobs/290675698

Can you take a look? Thanks

mgierlings:
> Sorry about delay here. I think it's ok to merge this and we can continue to look at the correct number of cores to use, possibility of SIMD etc later. 

@randombit No problem. I added a method to `XMSS_Tools` which will *attempt* to measure if the number of cores reported by `std::thread::hardware_concurrency()` is the number of physical cores or threads. On my hardware it seems to work however I'd recommend to run it on different hardware platforms with and without threading enabled before merging it. 

Once we have the capabilities in `Botan::CPUID` to reliably detect physical core counts across hardware and OS boundaries this code becomes obsolete running on "bare metal" with a single CPU socket. However it might be still a good idea to measure speeds with a different amount of threads in virtualized or cloud environments. Multi-socket systems might also be an interesting corner case. If we decide to keep it around for these purposes I might need to refine the code a little more. At the moment it assumes that there are only two possibilities:

1.  `std::thread::hardware_concurrency()` reports the actual number of physical cores in which case using just as many threads usually will the most speed.
2.  The CPU supports threading in which case `std::thread::hardware_concurrency()` reports twice the number of physical cores, so only half the number of reported cores will be used.

These assumptions will probably hold on most modern desktop hardware. On more exotic hardware, multi-socket systems or in virtualized environments it might not yield optimal results.

In case we decide that we don't want the runtime benchmark we can just drop the latest two commits: https://github.com/randombit/botan/pull/1267/commits/0e28426ca870e1e560b0f89baad92071f6813c4e and https://github.com/randombit/botan/pull/1267/commits/f504a76e3409e16da9701adaa4a297724ae0e4b1.

> However some of the CI tests are crashing, specifically all of the MSVC 32-bit builds (https://ci.appveyor.com/project/randombit/botan/build/1.0.3212/job/2wrgecx5kntkduol), as well as the Wine cross build (https://travis-ci.org/randombit/botan/jobs/290675691). All crash during execution of the aead tests - which seems strange.

This was indeed very weird since I did not touch any of the related code. The issue never occurred on my machines and also disappeared in the CI. I have no clue what happened there.

> Also the amalgamation is failing to build https://travis-ci.org/randombit/botan/jobs/290675698

The failing amalgamation build was my bad. It was caused by the visibility changes I applied to the WOTS classes and is fixed. All CI builds now finish without error.

randombit:
OK I haven't had time to test this on larger systems but I think with the goal of making forward progress I am going to merge to master now and then we can revisit the optimum thread counts, CPUID support etc soon (hopefully before 2.4.0)

Issue #871, opened 2017-02-07
Default padding for RSA signatures in X.509 certs


neusdan:
Can we change the default padding mechanism for RSA signatures to EMSA4/PSS ?

As you can see [here](https://github.com/randombit/botan/blob/master/src/lib/x509/x509_ca.cpp#L240) currently EMSA3/PKCS#1 v1.5 is used.

PSS should be used for new applications.

randombit:
TBH I'm not sure that RSS-PSS certs even work in Botan right now, so at the least some testing would have to happen before changing that.

Also I'm not sure how many other implementations support PSS certs, so changing the default might make certs generated by Botan unreadable by other implementations; but we could certainly make it possible for the application to select generating PSS certs (once it's been tested) instead of hardcoding PKCSv1.5

neusdan:
Yeah, doesn't seem to work right now. Did a quick test and hardcoded it to EMSA4/PSS, but failed because  the OID is missing in oids.cpp.

randombit:
There is more to it than that, one must extract the salt and MGF parameters from the AlgorithmIdentifier, so some special PSS codepaths must be added for signing and verifying X509 objects. The most recent spec seems to be RFC 4055 https://tools.ietf.org/html/rfc4055 updated by RFC 5756 https://tools.ietf.org/html/rfc5756

If RSA-PSS is going to be supported on certificates, it should also be supported by CRLs and OCSP. I think most of the codepaths for all 3 are already shared, but if not it could be taken as a sign to refactor the code so that adding another signature scheme for certificates in the future (like XMSS) is as easy as possible.

Seems Firefox does not support RSA-PSS (https://bugzilla.mozilla.org/show_bug.cgi?id=1088140) but based on comments in that bug Chrome and IE do.

randombit:
RSA-PSS works!

I don't think we want to change the default yet. Maybe in a couple of years...

PR #1270, opened 2017-10-20
Verification of RSA-PSS signed X.509 objects

neusdan:
The PR adds support for verifying RSA-PSS signed CRTs, CSRs and CRLs.
This is done basically by decoding the `RSASSA-PSS-params` structure in the `X509_Object` class.
For more information about this structure also see RFC 4055.

RSA-PSS signed OCSP responses are probably not working. Couldn't find such a response to test with.

Furthermore added a lot of test CRTs, CSRs and CRLs. Sources are documented in `src/tests/data/pss_x509/Sources.txt`

There is one certificate in `src\tests\data\pss_x509\98` which includes a bad salt length: the salt length specified in the `RSASSA-PSS-params` structure is not the one used to compute the signature. Botan verifies this signature/certificate correctly because it infers the salt length from the signature and does not use the salt length specified in the `RSASSA-PSS-params`. I've observed that OpenSSL reports an error when verifying the certificate. I couldn't find any rules in the RFCs how to deal with such a case.

Additionally some more RSA-PSS verification test vectors are added:
* NIST CAVP RSA-PSS verification tests: SigVerPSS_186-3
* From RSA: ftp://ftp.rsa.com/pub/pkcs/pkcs-1/pkcs-1v2-1-vec.zip

Related: #871


randombit:
Oh thank you.

Re the PSS salt length thing - yes right now PSS will verify any salt length. I already planned to address this because TLS v1.3 also requires that PSS signatures be generated with salt length == hash length, and that verifiers check the salt length is of the mandated size.

Basically:

- If a specific salt length is provided by the caller, eg `PSSR(SHA-256,16)`, verify that the salt length is of that size. Right now this only controls the length of salt for signatures we generate.

- If no specific salt length is provided, eg `PSSR(SHA-256)`, then generate using a salt equal in length to the hash output, but do not place any restrictions on verification.

randombit:
I don't have time to review this today but I'll try to look at it this weekend.

randombit:
A bunch of the fuzzer tests crash.

https://travis-ci.org/randombit/botan/jobs/290457314
https://travis-ci.org/randombit/botan/jobs/290457315#L2261

Should be easy to replicate with `./src/scripts/ci_build.py fuzzers`

neusdan:
Oh okay. Will have a look at it.

src/lib/x509/x509_obj.cpp
```diff
@@ -181,6 +220,49 @@ bool X509_Object::check_signature(const Public_Key& pub_key) const
...
+
+         std::string mgf_algo = OIDS::lookup(pss_parameter.mask_gen_algo.oid);
+         if(mgf_algo != "MGF1")
+            {
+            return false;
+            }
+
+         // For MGF1, it is strongly RECOMMENDED that the underlying hash function be the same as the one identified by hashAlgorithm
+         // Must be SHA1, SHA2-224, SHA2-256, SHA2-384 or SHA2-512
+         std::string mgf_hash_algo = OIDS::lookup(pss_parameter.mask_gen_hash.oid);
```
randombit: Could this just compare the OIDs directly (mask_gen_hash.oid == hash_algo.oid)? Not a big deal but seems simpler than decoding twice and comparing the strings. It would break in the case that we supported more than one OID for a hash, and someone mixed them. But I'd be inclined to reject that out of hand on principle.

src/lib/x509/x509cert.cpp
```diff
@@ -112,6 +112,53 @@ void X509_Certificate::force_decode()
...
+            ToDo: Allow salt length to be greater
+            */
+            if(public_key_alg_id != signature_algorithm())
+               {
+               throw Decoding_Error("Algorithm identifier mismatch");
+               }
+            }
+         if(sig_info[1] == "OAEP")
+            {
+            throw Decoding_Error("Currently unsupported");
```
randombit: This should be a more descriptive string, if it triggers during verification it will probably be confusing to the user. "RSA-OAEP certificate verification not supported"?

src/tests/data/pubkey/rsa_invalid.vec
```diff
@@ -148,3 +148,1359 @@ InvalidSignature = 1acce04e348a5c8377c54d8ddd8ec2d8c5cb9b195863c32eb716745f3462b
...
 InvalidSignature = ab9014dc47d44b6d260fc1fef9ab022042fd9566e9d7b60c54100cb6e1d4edc98590467d0502c17fce69d00ac5efb40b2cb167d8a44ab93d73c4d0f109fb5a26c2f8823236ff517cf84412e173679cfae42e043b6fec81f9d984b562517e6febe1f72295dbc3fdfc19d3240aa75515563f31dad83563f3a315acf9a0b351a2300000
+
+# Based on SigVerPSS_186-3.rsp
+# CAVS 11.1
+# "FIPS186-3 - SigVer RSA PKCS#1 RSASSA-PSS" information for "rsa2_check"
+# Combinations selected:Mod Size 1024 with SHA-1(Salt len: 20); SHA-224(Salt len: 20); SHA-256(Salt len: 20); SHA-384(Salt len: 20); SHA-512(Salt len: 20);; Mod Size 2048 with SHA-1(Salt len: 20); SHA-224(Salt len: 28); SHA-256(Salt len: 32); SHA-384(Salt len: 48); SHA-512(Salt len: 64);; Mod Size 3072 with SHA-1(Salt len: 0); SHA-224(Salt len: 0); SHA-256(Salt len: 0); SHA-384(Salt len: 24); SHA-512(Salt len: 0);
+# Generated on Wed May 18 13:42:19 2011
+
+Padding = EMSA4(SHA-1)
+E = 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000090c6d3
```
randombit: Nit - the leading 0x000000000... here (and following cases) is unnecessary

src/lib/pubkey/pk_algs.cpp
```diff
@@ -84,7 +84,9 @@ std::unique_ptr<Public_Key>
 load_public_key(const AlgorithmIdentifier& alg_id,
                 const std::vector<uint8_t>& key_bits)
    {
-   const std::string alg_name = OIDS::lookup(alg_id.oid);
+   const std::vector<std::string> alg_info = split_on(OIDS::lookup(alg_id.oid), '/');
+   const std::string alg_name = alg_info[0];
```
randombit: I think if the OID is unknown this can access an empty vector

randombit:
Couple of minor things + the crash (which might be the same issue in `load_public_key` I commented on, not sure) otherwise LGTM

src/lib/x509/x509_obj.cpp
neusdan: Removed the lookup and now compare the OIDs directly. Not sure if this is what you meant.

src/lib/x509/x509cert.cpp
neusdan: Good point. Should be more descriptive now.

src/tests/data/pubkey/rsa_invalid.vec
neusdan: fixed

src/lib/pubkey/pk_algs.cpp
neusdan: fixed. Seems like the fuzzers don't compile on Windows, so at the moment i have no way to verify local on my machine if this is causing the fuzzer crashes.

neusdan:
Thanks for the review. Changes pushed.

PR #1271, opened 2017-10-21
Deprecate some exceptions

neusdan:
This should work to deprecate these exceptions and prevent hundreds of MSVC warnings.

Related:
53a7136759a38f5cd7b1dfcd79868ffe89d61c35
843a982c3a019e18975aef0277af44a3731f8caa
2e491b4ddbb63543d027a8a84a9fb523383ff3f0


randombit:
Ah right that works. Thanks

PR #1273, opened 2017-10-22
Refactor cli and test runners

randombit:
Allows cleaning up header includes, also somewhat smaller binaries.

### DIFF
--- a/src/cli/argparse.h
+++ b/src/cli/argparse.h
@@ -1,0 +1,278 @@
+/*
+* (C) 2015,2017 Jack Lloyd
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#ifndef BOTAN_ARGPARSE_H_
+#define BOTAN_ARGPARSE_H_
+
+#include <string>
+#include <map>
+#include <set>
+#include <vector>
+#include <botan/parsing.h>
+#include "cli_exceptions.h"
+
+namespace Botan_CLI {
+
+class Argument_Parser
+   {
+   public:
+      Argument_Parser(const std::string& spec,
+                      const std::vector<std::string>& extra_flags = {},
+                      const std::vector<std::string>& extra_opts = {});
+
+      void parse_args(const std::vector<std::string>& params);
+
+      bool flag_set(const std::string& flag) const;
+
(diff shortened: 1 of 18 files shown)


configure.py
```diff
@@ -223,7 +226,8 @@ def find_headers_in(basedir, srcdir):
 
         self.cli_sources = list(find_sources_in(source_paths.src_dir, 'cli'))
         self.cli_headers = list(find_headers_in(source_paths.src_dir, 'cli'))
-        self.test_sources = list(find_sources_in(source_paths.src_dir, 'tests'))
+        self.test_sources = list(find_sources_in(source_paths.src_dir, 'tests')) + \
+                            list(find_sources_in(source_paths.src_dir, 'cli', 'cli'))
```
webmaster128: Why are .cpp files from src/cli added to `self.test_sources`? Shouldn't those folders remain independent?

randombit: Already the tests and cli code are sharing the driver https://github.com/randombit/botan/blob/master/src/tests/main.cpp#L7
but right now cli.h is header only so it's just an include.

I guess there is some argument to be made to have a cli lib which the cli and tests can both depend on. But since I don't see much reason to have more than 2 binaries (multicall `botan` that is installed plus test suite that's not installed) it seems like a needless complexity in the build. Or alternately removing the dependency and making the cli and tests completely independent. But I don't want to repeat (or copy+paste) the options parsing code. :/

webmaster128: Given the extra complexity this will add to the build system as well as complicated dependency trees, I really prefer a shared source. Having an extra library is likely too much, as you said. What comes to my mind now is:

1. Copy & paste code
2. Use a code generator script to write the same file into two locations
3. Use a symbolic link to share the same file between ./src/cli and /src/tests (no duplication but different files for the build system)
4. Have an internal class in the botan lib

randombit: Looking at this a bit more I realize the test runner only depends on a tiny portion of the overall cli support library. I'm going to split that part out (as either header-only code or maybe include in the library itself, not sure yet) and remove the larger cross dependency between the two components.

webmaster128: :+1:

PR #1275, opened 2017-10-25
Perform OIDS lookup before to prevent a guaranteed exception (EC_Group)

fschoenm:
In our TLS stack, this location always raised an exception because the ``curve_name`` is never PEM-encoded here (as far as I could see) but filled from output of ``Supported_Elliptic_Curves::curve_id_to_name()``, which only returns real curve names. I tried to prevent the exception by doing an eager lookup here instead of waiting for the exception handler.

### DIFF
--- a/src/lib/tls/msg_client_kex.cpp
+++ b/src/lib/tls/msg_client_kex.cpp
@@ -7,6 +7,7 @@
 
 #include <botan/tls_messages.h>
 #include <botan/tls_extensions.h>
+#include <botan/oids.h>
 #include <botan/rng.h>
 
 #include <botan/internal/tls_reader.h>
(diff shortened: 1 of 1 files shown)


randombit:
I was confused when I first read this because I thought you meant the exception escaped and the connection failed. But now I understand this is bypassing a needless throw/catch in the EC_Group constructor. The `EC_Group` constructors are kind of weird, it would be nice to clean this behavior up.

Thanks for the PR.

fschoenm:
Sorry for the confusion :) I didn't check all the other places of ``EC_Group`` constructors. Just this one in particular was very distracting during debugging, so thanks for merging.

PR #1285, opened 2017-11-05
Use enumerations for TLS parameters, add PSS signatures

randombit:
The code is overall cleaner, also this makes the change that we distinguish between ciphersuites which are anonymous, and ciphersuites which implicitly authenticate due to the key exchange (SRP/PSK/static RSA). Currently those are treated the same way.

Downside is it's slightly harder to add a new algo (eg #1284) since it requires doing more than just editing the generator script, also enums must be changed and so forth. But this is not exactly a common scenario.

Adds support for RSA-PSS signatures. Tested client-side against a few google servers.

Only breaking changes as a result of this (at least, that I know of):
 - The old Google specific ChaCha ciphersuites are removed. This was because they used a 4th nonce format (distinct from either GCM or standard ChaCha). I checked and OpenSSL doesn't support them anymore either. 
 - Enabling static RSA requires both enabling RSA key exchange and enabling "IMPLICIT" authentication mode.

It would be good to make the same change for cipher modes but this commit is already large and complicated so leaving it for later.


randombit:
Turned out supporting PSS signatures was really easy after changing the signature extension to use enums, so that will end up going into this PR

randombit:
I don't have time to finish this properly for 2.4.0 and don't want to merge it half baked. Plan to merge it early in the 2.5 cycle

PR #1287, opened 2017-11-09
Use UTF-8 for ASN.1 strings

randombit:
#1250 plus a couple further commits using the conversion routines added in #1286

I searched around trying to find a "live" certificate that uses BMP or UCS4 strings, no luck.


randombit:
Fixes #1113

reneme:
Please see PR #1289 that reads a certificate containing `BMPString`-encoded fields on top of this PR.

PR #1290, opened 2017-11-11
Missing include: #include <cstdlib> is needed for std::free.

yurivict:


randombit:
Thanks!

webmaster128:
The new include duplicates line 18 of the same file, doesn't it?

I also noted missing `<cstdlib>` includes for `free` after upgrading to the toolchain in Ubuntu 17.10, where some default includes may have been dropped from compiler or libc++, but the problem must be somewhere else then openssl_rsa.cpp. Could you post your error output @yurivict?

randombit:
LOL this PR was a repeat of #1233, @yurivict presumably was trying to build 2.3.0

yurivict:
Collision of discoveries.

PR #1296, opened 2017-11-16
Fix secure_allocator with std::string on VS2013

### LINKED ISSUE #1295: Using std::string with Botan::secure_allocator
Is this somehow possible? Something like this:

```C++
using secure_string = std::basic_string<char, std::char_traits<char>, Botan::secure_allocator<char> >;	
secure_string test = "abcd";
```

This gives some compile errors. My guess is that the `Botan::secure_allocator` has to be extended to support this.

neusdan:
Closes #1295

### DIFF
--- a/src/lib/base/secmem.h
+++ b/src/lib/base/secmem.h
@@ -40,6 +40,22 @@ class secure_allocator
       secure_allocator(const secure_allocator&) = default;
       secure_allocator& operator=(const secure_allocator&) = default;
       ~secure_allocator() = default;
+
+      template <typename U>
+      struct rebind
+         {
+         typedef secure_allocator<U> other;
+         };
+
+      void construct(value_type* mem, const value_type& value)
+         {
+         std::_Construct(mem, value);
+         }
+
+      void destroy(value_type* mem)
+         {
+         std::_Destroy(mem);
+         }
 #else
       secure_allocator() BOTAN_NOEXCEPT = default;
       secure_allocator(const secure_allocator&) BOTAN_NOEXCEPT = default;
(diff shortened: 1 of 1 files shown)


randombit:
Does this reintroduce the performance problem of #1228? I guess it only affects VC2013 so I don't much care, but someone else might.

neusdan:
I have no TLS setup. Can i test this somehow?

randombit:
Might be easiest to just write a test program (creates a `secure_vector<uint8_t>`, then invoke copy constructor) and look at the resulting assembly with vs without your patch, does it call memcpy or not?

If a problem, the trick using `enable_if` proposed in #1228 would work, but that might conflicts with whatever 2013's `std::string` is expecting. (I assume you added `construct` and `destruct` back because `std::string` doesn't work without them, so probably they are required even if the T is trivial....)

neusdan:
If i see this correctly then in both cases memcpy is called.

Just to be sure i made a test program to measure both versions:

```C++
void measure( int i )
{
	Botan::System_RNG rng;
	Botan::secure_vector<uint8_t> x = rng.random_vec( 1024 );

	std::chrono::steady_clock::time_point begin = std::chrono::steady_clock::now();
	for ( size_t i = 0; i < 1000000; ++i )
	{
		Botan::secure_vector<uint8_t> y( x );
	}
	std::chrono::steady_clock::time_point end = std::chrono::steady_clock::now();

	std::cout << i << ": " << std::chrono::duration_cast<std::chrono::milliseconds>( end - begin ).count() << "ms" << std::endl;
}

int main()
{
	for ( size_t i = 1; i <= 5; ++i )
	{
		measure( i );
	}
	
    return 0;
}
```

Output:

Without patch: 
1: 159ms
2: 158ms
3: 157ms
4: 156ms
5: 159ms

With patch:
1: 161ms
2: 156ms
3: 160ms
4: 154ms
5: 158ms

randombit:
Looks ok to me then, thanks for checking

Issue #1297, opened 2017-11-17
nmake install launches text editor instead of executing install.py


solvingj:
On windows, after running configure.py, the install method from the Makefile is here: 

```
install: $(CLI) docs
	$(SCRIPTS_DIR)\install.py --destdir=C:\Users\botan_dir --build-dir="build" --bindir=bin --libdir=lib --docdir=docs --includedir=include
```
When I run `nmake install`, windows opens the `install.py` file in my default editor for `.py` files.  With windows, one typically has to pass `.py` files as arguments to the `python` binary, as in `python $(SCRIPTS_DIR)\install.py ` .  So, I'm confused how this is working for anyone.

randombit:
It does seem the install target works on our Windows CI builds:

https://ci.appveyor.com/project/randombit/botan/build/1.0.3317/job/tmyhcv840kcoi4mu#L2025

Probably the difference is because you've associated .py files with your editor, and our CI has associated it with the Python interpreter.

It should be an easy fix, all we need to do is export the value of `sys.executable` as an additional template variable, and then invoke the install script as you suggest.

randombit:
(I am however surprised nobody has run into this before - or perhaps they have and just didn't bother creating a bug report...)

solvingj:
Thanks for the clarification, it does make me feel better.

randombit:
Can you check if the patch in #1298 does the trick?

solvingj:
Yes, it works, thanks very much!

PR #1298, opened 2017-11-17
Prefix execution of install.py with Python binary

randombit:
Fixes GH #1297

This is done even on GNU make builds, since the same issue affects MinGW


solvingj:
Sorry, can't test tonight, will try tomorrow.

solvingj:
I think this is good. Thank you so much.

Issue #1306, opened 2017-11-24
pipe constructor


lambdafu:
I tried to use pipe.append() to dynamically construct my filter chain, and got the deprecation warning. This is the best I could come up with now:

```
  Botan::Filter* f1 = new Botan::Hash_Filter(m_algo);
  Botan::Filter* f2 = new Botan::Hex_Encoder();
  Botan::Filter* f3 = new Botan::DataSink_Stream(std::cout);
  std::shared_ptr<Botan::Pipe> pipe;
  if (m_raw)
    pipe = std::make_shared<Botan::Pipe>(f1, f3);
  else
    pipe = std::make_shared<Botan::Pipe>(f1, f2, f3);
```

Is there a better pattern?  I realize that Pipe steals the pointers, and thus using a std::vector<Botan::Filter*> is a bit fragile.  But constructing an initializer_list dynamically is not straight forward.
I am only trying to construct the filters dynamically, I do not want to change them after processing has started.

randombit:
Pipe's constructor will ignore null arguments, so you could do something like this (untested):

```
  std::shared_ptr<Botan::Pipe> pipe;
  pipe = std::make_shared<Botan::Pipe>(
      new Botan::Hash_Filter(m_algo),
      (m_raw) ? nullptr : new Botan::Hex_Encoder(),
      new Botan::DataSink_Stream(std::cout)
  );
```

It may be worth adding a variant of `append` which works like the current version, except that it throws if any messages have already been processed. Thus allowing dynamic setup but prohibiting runtime modification.

lambdafu:
Thanks, that variation at least has only a single constructor, so I don't need the shared pointer.  Nice :)

randombit:
I've added a variant of `append` that only allows modification before the first message is processed in #1307. So I think we're set here, but please reopen or open a new issue if you have further problems/questions.

lambdafu:
Thanks! Looking good! :beach_umbrella:

PR #1307, opened 2017-11-24
Add Pipe::append_filter

randombit:
Similar to append but it only allows modification before start_msg. See GH #1306

### DIFF
--- a/src/lib/filters/pipe.cpp
+++ b/src/lib/filters/pipe.cpp
@@ -40,7 +40,10 @@ Pipe::Pipe(Filter* f1, Filter* f2, Filter* f3, Filter* f4) :
 */
 Pipe::Pipe(std::initializer_list<Filter*> args)
    {
-   init();
+   m_outputs.reset(new Output_Buffers);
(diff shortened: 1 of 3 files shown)


matejk:
append() and prepend() were made deprecated, however only append_filter() was added.

Is prepend_filter() missing also or was not implemented intentionally?

randombit:
Hm, not necessarily intentionally - but I assumed that if we are limiting modification of a Pipe to prior to message processing, then appending is sufficient. Is there some situation I'm not thinking of, where this would be useful?

matejk:
Our software that uses Botan to do processing on the files in certain order, which is not necessarily the same for all objects: compressions, encryptions etc.

The steps done when archiving data are saved with the object.

When getting the data back from the archive, these steps have to be done in reverse order. prepend() is very handy to create proper pipe. Of course before calling any start_msg().

randombit:
Makes sense. I'll add a similar prepend operation for 2.5.0

matejk:
Thank you very much.

PR #1308, opened 2017-11-25
Add test case for bzip2 compression filter.

lambdafu:
Let me break your build :) I think there is a flush() missing when closing the stream.  I could reproduce the problem with Compression_Algorithm instead of the filter when omitting the `update(buffer, offset, true)` stage, see below for an isolated test. I would fix it if I had a good idea where to put the flush in the pipe processing, I guess it needs to be somewhere in end_msg().

```
void test_isolated(bool with_flush) {
  Botan::Compression_Algorithm* comp = Botan::make_compressor("bz2");
  Botan::secure_vector<unsigned char> buffer(2048);
  comp->start(0);
  buffer = {'f', 'o', 'o', '\n' };
  comp->update(buffer, 0, with_flush);
  std::cout.write((char*)buffer.data(), buffer.size());
  buffer.clear();
  comp->finish(buffer, 0);
  std::cout.write((char*)buffer.data(), buffer.size());
}

// OK:
test_isolated(true);
// throws bzip compress error 1
test_isolated(false);
```

### DIFF
--- a/src/tests/test_filters.cpp
+++ b/src/tests/test_filters.cpp
@@ -57,6 +57,7 @@ class Filter_Tests final : public Test
          results.push_back(test_pipe_cbc());
          results.push_back(test_pipe_cfb());
          results.push_back(test_pipe_compress());
+         results.push_back(test_pipe_compress_bzip2());
          results.push_back(test_pipe_codec());
          results.push_back(test_fork());
          results.push_back(test_chain());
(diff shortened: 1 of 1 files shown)


randombit:
Thanks. Bzip2 seems pickier about how it is called than zlib. I think I see the issue.

randombit:
Fixed, thanks!

PR #1310, opened 2017-11-25
Switch to updated default Travis XCode (8.3)

randombit:
Travis is moving to 8.3 as new default XCode image (starting Tuesday) and deprecating 8.2. Test it now.


randombit:
This change seemed to cause the CI build to freeze? No progress after 16 hours. Wut?

Closing, will retry again on Tuesday when the default is switched over.

PR #1311, opened 2017-11-25
Add create and create_or_throw factories for Compression_Algorithm and Decompression_Algorithm.

lambdafu:
I noticed that make_compressor was working differently from the create/create_or_throw factories I already knew for hash functions. Assuming that create/create_or_throw is the newer interface, I added them for compression and decompression (I guess adding an alias CompressionFunction = Compression_Algorithm, and DecompressionFunction = Decompression_Algorithm is also interesting, but I didn't do that).

If this doesn't go in the right direction, please disregard it. :)

### DIFF
--- a/src/lib/compression/compression.h
+++ b/src/lib/compression/compression.h
@@ -20,6 +20,22 @@ class BOTAN_PUBLIC_API(2,0) Compression_Algorithm
    {
    public:
       /**
+      * Create an instance based on a name, or return null if the
+      * algo/provider combination cannot be found. If provider is
+      * empty then best available is chosen.
+      */
+      static std::unique_ptr<Compression_Algorithm>
+         create(const std::string& algo_spec);
+
+      /**
+      * Create an instance based on a name
+      * @param algo_spec algorithm name
+      * Throws Lookup_Error if not not found.
+      */
+      static std::unique_ptr<Compression_Algorithm>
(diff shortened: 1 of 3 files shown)


src/lib/compression/compression.h
```diff
@@ -19,6 +19,22 @@ namespace Botan {
 class BOTAN_PUBLIC_API(2,0) Compression_Algorithm
    {
    public:
+      /**
+      * Create an instance based on a name, or return null if the
+      * algo/provider combination cannot be found. If provider is
+      * empty then best available is chosen.
```
randombit: No provider argument here so this statement is bogus.

src/lib/compression/compression.h
```diff
@@ -65,6 +81,22 @@ class BOTAN_PUBLIC_API(2,0) Compression_Algorithm
 class BOTAN_PUBLIC_API(2,0) Decompression_Algorithm
    {
    public:
+      /**
+      * Create an instance based on a name, or return null if the
+      * algo/provider combination cannot be found. If provider is
+      * empty then best available is chosen.
```
randombit: Same

randombit:
Thanks! Minor nit on the Doxygen comments (references non-existent provider argument) other than that LGTM

lambdafu:
Sorry that I missed those.  I remember removing two other ones :cloud:

PR #475, opened 2016-04-13
[for review] Remove Transform base class

randombit:
Removes the Transform base class, which had effectively two subclasses - Cipher_Mode and Compression_Transform. Encryption and compression are completely different operations and trying to have an API that covers both cases led to bad results for both of them. A lot of operations previously on the compression objects, like querying the nonce length, make no sense.

The Cipher_Mode interface is left intact here, the APIs on Transform just move down to it. The compression interface changes: it splits into two types (Compression_Algorithm and
Decompression_Algorithm). Compression_Algorithm's start() takes the compression level instead of the constructor, which allows varying compression levels with a single object,. Flushing the compression state is moved to a bool param on `Compression_Algorithm::update`.

PR #1312, opened 2017-11-26
Fix documentation of compression/decompression update function.

lambdafu:
This seems to be a cut&paste leftover from the cipher docs.


randombit:
At one point this comment was actually correct, but not since #475. Thanks

PR #1316, opened 2017-11-28
Fix issues found with tlsfuzzer

randombit:
Nothing security critical, but some nice bugs.

FFDHE-4096 was actually FFDHE-3072. :) Also because HMAC had a key size limit of 512 bytes, DH groups > 4096 bits could not actually be used.

Also reject SSLv3 properly - the server didn't recognize it, treated it as a future version and negotiated TLS v1.2!


securitykernel:
> FFDHE-4096 was actually FFDHE-3072. :)

Oops, my fault, sorry. I converted the group parameters from the RFC programatically one by one and must have made a mistake there.

PR #1317, opened 2017-11-29
Try removing ranlib usage for macOS/iOS

randombit:
Supposedly "ar crs" works in all recent versions, lets find out.


randombit:
Huzah it works

PR #1327, opened 2017-12-01
fix install failure when static libraries are disabled

alonbl:
not sure this is the best approach but is working


randombit:
Ack, thanks. I fixed this in a slightly different way in #1328

PR #1332, opened 2017-12-04
Tls external crypto

harrmann:
### Improve support for external crypto devices in TLS handshake

Using an external crypto device for signature generation and verification or (EC)DH key exchange while establishing a TLS session requires hooks to implement the corresponding functionality. To this end some TLS signature and key exchange code is moved into additional callbacks which can be overwritten.

In client side TLS handshake processing it may be necessary to verify the server certificate before using the public key to check the signature in a server key exchange message. The current implementation postpones verification of the server certificate until the server hello done message has been received. This is not necessary if the client does not expect a certificate status message. Therefore support for certificate status messages is made optional via an additional policy entry and the server certificate is verified immediately after receiving it whenever this is appropriate.

### DIFF
--- a/src/lib/tls/msg_cert_verify.cpp
+++ b/src/lib/tls/msg_cert_verify.cpp
@@ -1,6 +1,7 @@
 /*
 * Certificate Verify Message
 * (C) 2004,2006,2011,2012 Jack Lloyd
+*     2017 Harry Reimann, Rohde & Schwarz Cybersecurity
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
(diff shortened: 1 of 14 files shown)


randombit:
Nice, thanks. I'll review this soon.

> In client side TLS handshake processing it may be necessary to verify the server certificate before using the public key to check the signature in a server key exchange message.

Can you explain more about this? Is this some limitation with the external device?

randombit:
Oh and please add (C) statements to relevant files, thanks

randombit:
Callbacks code looks good. I am still confused by the need to disable the stapled response though.

harrmann:
> Callbacks code looks good. I am still confused by the need to disable the stapled response though.
The crypto device does not accept a public key without the certificate that proofs its authenticity.

randombit:
Interesting thanks

Issue #1338, opened 2017-12-05
List supported Elliptic Curves


J08nY:
There is currently no way to enumerate supported Elliptic Curves, apart from trying some known OIDs in `OIDS::lookup`. This would be quite useful. (list their string names as Botan uses them and also their OIDs)

I could do a PR if this feature would be approved, however currently I am not quite sure for a right place for such enumeration.

randombit:
This would be useful, especially over time as groups are added and removed.

Probably best spot is in `ec_named.cpp`, add new static function

```
std::vector<std::string> EC_Group::known_ec_groups()
   {
   return {
      "secp256r1",
        ...
       };
    }
```

There are some refactoring planned for how EC_Groups are stored so this implementation will eventually change, but I could handle that with the rest of the refactoring. The above would do for now.

Someone who needs the OID can call `OIDS:lookup` on the returned strings.

If you do this please also add test that verifies that for each value returned by `known_ec_groups` we are also able to retrieve the representation as well as lookup an OID for that name.

Bonus points for updating `ECC_Randomized_Tests` (`tests/unit_ecc.cpp`) to use this function instead of hardcoding its own list of curves.

randombit:
Hm, things to contemplate:

Does a `std::set` make more more sense? There is no meaning to the ordering of the return value.

Is it useful to cache the value (as a static const) and return a const reference, instead of new object?

J08nY:
Right, both things make sense I think. Will work on the PR soon.

randombit:
Fixed in #1338 thanks!

PR #1339, opened 2017-12-05
Add known_named_groups to EC_Group to list known curves. Fixes GX #1338.

J08nY:
This adds listing of known named elliptic curves, as suggested in #1338.


randombit:
Looks good, thanks!

Issue #1323, opened 2017-11-30
https client library with botan backend?


lambdafu:
I almost feel like misusing the ticket system, because this feels more like a support question. But I don't know how to reach the community in another way, sorry!
Is there already a https client library (such as curl) that supports Botan as backend? I would feel really stupid linking to curl and then sucking in openssl as a dependency, if I am heavily leaning on Botan for all the other crypto in the application;
If there isn't, I guess am raising my hand to volunteer. Is there an important alternative to libcurl in the C++ world that I am missing?  I couldn't immediately find something. @securitykernel told me that the C89 API of Botan does not yet support TLS, so if a C library is used for HTTP, adding that is an obvious starting point.  Anything else to watch out for?
Thanks!

webmaster128:
+1 for working on Botan as a curl TLS backend.

> I would feel really stupid linking to curl and then sucking in openssl as a dependency, if I am heavily leaning on Botan for all the other crypto in the application;

This is what we do in Kullo for Linux. No real issue right now but it would be great to have the choice.

randombit:
Questions or general support issues are fine here. You can also use the mailing list (https://lists.randombit.net/mailman/listinfo/botan-devel) but GH Issues is more active.

There are a couple of relatively widely used C++ HTTP client libraries (for example https://github.com/cpp-netlib/cpp-netlib) but afaik even in C++, curl is far and away the most common for this task. Also the fact that curl already supports a large number of SSL libraries suggests they'd likely accept an additional implementation. Whereas most of the other relevant libs I know of (like cpp-netlib, asio) support exclusively OpenSSL and getting them to take a patch supporting multiple backends might be an uphill slog; for such libs, it may prove easier to just write an OpenSSL compatible stub library.

And besides curl, a C89 TLS interface would be great to have since we could then expose TLS to Python, OCaml, etc. There is a sketch of an interface at the end of `ffi.h` (its `#if 0`ed out and definitely incomplete - for instance the `_init` call don't offer any way to set server certificates, there is no way to get info on ciphersuites, etc). Also certificate verification is not currently exposed to C, and that will certainly be required, at least eventually.

There is a Slack channel used "internally" for developer chat, I'll send you an invite. Optional of course but good if you have a quick question on how to approach something.

lambdafu:
[...]

PR #1341, opened 2017-12-07
[WIP] Some further sketch work on an C interface for TLS

randombit:
Re #1323

To some extent I think this design should co-evolve with a curl backend implementation. Since curl already supports most TLS libraries with a C interface, whatever we do *should* fit into the same model.

Big question is if the C API should follow the C++ API with its callbacks for output, or if it would be better to just buffer outputs in the FFI object and then the application can call a function to find out if output is pending and call `SSL_read` when they feel like it. It looks like curl assumes it can do that; though of course it's possible to implement the same buffering just within curl's vtls layer.

@lambdafu Don't know if you had any plans on taking this up, but feel free to fork from here and continue. I probably won't have time to work on this further before 2.4.0 release.

### DIFF
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -1334,22 +1334,56 @@ int botan_fpe_decrypt(botan_fpe_t fpe, botan_mp_t x, const uint8_t tweak[], size
 /*
 * TLS (WIP)
 */
-#if defined(BOTAN_HAS_TLS) && 0
 
+/**
+* TLS Policy objects
+*/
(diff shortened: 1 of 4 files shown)


lambdafu:
Yes, I planned to work on it this week, but did a new website for neopg instead (I'm lagging behind in documentation). I am thankful for any help I can get, so great to see your sketches, thanks a lot!

randombit:
Closing, not ready for review yet

talregev:
Will you continue?
https://github.com/randombit/botan/discussions/4320

PR #1343, opened 2017-12-09
Generate CMakeLists using a template file

randombit:
@webmaster128 You actually use cmake support so would appreciate a quick review if you have time.

I checked a few common configurations (Linux/gcc, macOS/clang, Windows/MSVC) and for each the output was identical to what master produces, modulo some insignificant whitespace and ordering.


randombit:
Closing in favor of #1346

PR #1344, opened 2017-12-09
Use template file to generate bakefile

randombit:
@neusdan Similar to #1343 I don't use bakefile (nor, unlike cmake, even have a way to test) so a review would be good if you have time. The output seems to be identical to master modulo some minor whitespace and ordering things that shouldn't affect the build.

Also I added a restriction that fails the build if someone attempts to generate bakefile for a target != Windows/MSVC since the bakefile seems fixed to that configuration.


randombit:
Closing in favor of #1346

PR #1345, opened 2017-12-10
Add basic iteration constructs to template files

randombit:
Easier than I thought, once I finally sat down to write it. Eliminates the last bits of makefile-specific syntax from configure.py

Once #1344 and #1343 are merged I'll update the cmake and bakefile generators to also use this, which will allow some additional simplifications.


randombit:
Closing in favor of #1346

PR #1346, opened 2017-12-10
Templatize cmake and bakefile builds

randombit:
Merge of #1343 #1344 #1345

Will rebase before merge to clean up the history.


randombit:
OK this is ready to go now.

neusdan:
Bakefile x64 build works fine. x86 doesn't work:

`python configure.py --with-bakefile --cpu=x86`

Output from Bakefile:

`botan.bkl:488:8: error: variable "archs" (list of architectures): expression "x86_32" is not a valid architecture value: must be one of "x86", "x86_64"`

randombit:
Ah ok I'll fix that, thanks for testing.

configure.py
```diff
@@ -1114,6 +1122,30 @@ def isa_flags_for(self, isa, arch):
...
+
+                if isa is None:
+                    continue
+
+            flagset = self.isa_flags_for(isa, arch.basename)
+            if flagset is None:
+                raise UserError('Compiler %s does not support %s' % (self.basename, isa))
+            flags.add(flagset)
+
+        return ("".join([" %s" % f for f in sorted(flags)])).strip()
```
webmaster128: Isn't this the same as `" ".join(sorted(flags))`?

randombit: I believe so, will change. Thanks

randombit:
@neusdan Fixed the bakefile arch issue, also I noticed in the bakefile docs that it accepts having multiple files listed in the `sources {}` block so I changed the output a bit in a299f08. Should be identical behavior? Let me know if that ends up causing problems for you.

randombit:
Will merge this tomorrow night, assuming no additional problems are identified.

neusdan:
Tested it again. Everything fine!

PR #1352, opened 2017-12-11
Clarify platform log during configuration

webmaster128:
* make clear that the logged platform is just the computer running the script and independent of configuration options
* hide logging in non-verbose mode as user knows the machine running this

ps.: I am just testing the updated configure, which looks fine overall


randombit:
Makes sense, thanks

PR #1354, opened 2017-12-12
Fix pragma message to avoid warning

schregger:
Missing parentheses around pragma message caused warnings in Visual Studio.


randombit:
Thanks!

PR #1365, opened 2017-12-19
Always set domain encoding correctly when loading an ECC key

securitykernel:
When loading an ECC key from file, the domain encoding was always set to explicit instead of depending on the encoded key file read. This resulted in different encodings and therefore different
fingerprints of the same key when encoding the same key twice (once after generation, once after re-reading it).


randombit:
LGTM thanks

PR #1368, opened 2017-12-20
Enable signing X509 structures with rsa-pss

fweissberg:
The code is definitely rough around the edges but I'd like to share it this early since the feature freeze is nigh.
Obviously, there is a lot to do in order to keep the code coverage up. 

* enable rsa-pss signing fo x509 certificates, crls, certificate requests
* make the padding for RSA signatures of x509 certificates configurable
   * modify src/scripts/oids.py so that it reads the src/build-data/oids.txt and extracts valid padding schemes for the available sigining algorithms if run like ./src/scripts/oids.py pads > src/lib/pk_pad/padding.cpp
   * add std::string padding_scheme and a corresponding setter to X509_Cert_Options to control the padding used for self signed certificates and certificate requests
   * add a constructor to X509_CA that takes an options map<string,string>; use {"padding", <padding_scheme_name>}
   * overload the mostly internal choose_sig_format to take an options map<string,string>; use {"padding", <padding_scheme_name>}
      * choose_sig_format constructs an EMSA object and asks it to configure the padding and the signature AlgorithmIdentifier
* update the x509_unit tests to use all valid paddings for the tested algorithms
* augment the following cli commands by a switch "emsa="
   * gen_self_signed
   * gen_pkcs10
   * sign_cert

**common sense demo script; needs to be executed in the same directory as the botan executable/a link to the executable; requires openssl**

#!/bin/bash

echo "generating ca key"
./botan keygen > ca_key.pem

echo "creating ca certificate"
./botan gen_self_signed ca_key.pem test_ca --ca --emsa=EMSA4 > testca.pem.crt

echo "generating end key"
./botan keygen > end_key.pem

echo "creating certificate signing request"
./botan gen_pkcs10 end_key.pem end --emsa=EMSA4 > end.req

echo "creating and signing end certificate"
./botan sign_cert testca.pem.crt ca_key.pem end.req --emsa=EMSA4 > end.pem.crt

echo "checking the ca certificate with openssl"
openssl verify -verbose -CAfile ./testca.pem.crt ./testca.pem.crt
echo "testca.pem.crt signature algorithm according to openssl:"
openssl x509 -in testca.pem.crt -noout -text | grep "Signature Algorithm"

echo "checking path from the end certificate to the root with openssl"
openssl verify -verbose -CAfile ./testca.pem.crt ./end.pem.crt
echo "testca.pem.crt signature algorithm according to openssl:"
openssl x509 -in end.pem.crt -noout -text | grep "Signature Algorithm"


randombit:
Looks ok on a quick look.

For `padding.h` I'm not sure if we want to make this public. Given that, make it internal for now and we can promote it to public API later if it seems needed. Also, instead of ``__attribute__`` for setting symbol visibility use `BOTAN_TEST_API` which works with both Linux and Windows. And there is no reason to set the visbility attribute in the source file (`padding.cpp`) having it on the declaration is sufficient.

For the base case of `EMSA::config_for_x509` that just throws, put it into `emsa.cpp`, and include the output of `name` in the exception message so it's easier to diagnose.

`sig_algo_and_pad_ok` second param should be a const reference.

`check_padding_scheme` (in `x509self.cpp`) both params should be const refs.

`EMSA_Raw::name` seems wrong, should this use "Raw" not "RAW"?

I don't understand the return value of `config_for_x509` - it always returns `name` (which makes sense given the usage in `x509_ca.cpp`). Given this I'd prefer if `config_for_x509` returned the AlgorithmIdentifier (instead of having an output parameter), and `x509_ca.cpp`  called `emsa->name()` to get the name to pass to `PK_Signer`.

fweissberg:
I followed your suggestions.
check_padding_scheme is not needed any more.

fweissberg:
Noticed some more room for improving the code coverage.

randombit:
@fweissberg Looks good to me. Still has WIP in subject, are you planning further changes?

fweissberg:
I was thinking about something mentioned in [RFC4055](https://tools.ietf.org/html/rfc4055):
>    The rsaEncryption object identifier continues to identify the subject
>    public key when the RSA private key owner does not wish to limit the
>    use of the public key exclusively to either RSASSA-PSS or RSAES-OAEP.
>    In this case, the rsaEncryption object identifier MUST be used in the
>    algorithm field within the subject public key information, and the
>    parameters field MUST contain NULL.
> 
>       rsaEncryption  OBJECT IDENTIFIER  ::=  { pkcs-1 1 }
> 
>    Further discussion of the conventions associated with use of the
>    rsaEncryption object identifier can be found in RFC 3279 (see
>    [PKALGS], Section 2.3.1).
> 
>    When the RSA private key owner wishes to limit the use of the public
>    key exclusively to RSASSA-PSS, then the id-RSASSA-PSS object
>    identifier MUST be used in the algorithm field within the subject
>    public key information, and, if present, the parameters field MUST
>    contain RSASSA-PSS-params.  The id-RSASSA-PSS object identifier value
>    and the RSASSA-PSS-params syntax are fully described in Section 3.
At the moment, RSA keys appear to be always of type rsaEncryption so we do need to handle keys of type id-RSASSA-PSS, right?

Moreover, I thought about improving the code coverage. But this could be easily split into a future PR.

fweissberg:
I can squash the second commit and sign tomorrow morning.

randombit:
Yes we will want to support id-RSASSA-PSS, both for generation and on the validation side. But a distinct PR for that  seems better. I will plan on merging this tomorrow, after you have a chance to squash.

PR #1373, opened 2017-12-24
Fixes #1370 UBSan incompatible function ptr type

### LINKED ISSUE #1370: Odd UbSan warning - redc_p384 not of expected type
Clang 5.0 on x86-64 Linux.

I don't understand this warning. Commenting out the P-384 test in `unit_ecc.cpp`, it starts warning about the next run (P-256). Tried converting from `std::function` to a template parameter, no change.

```
nist_redc:
/bin/../lib64/gcc/x86_64-pc-linux-gnu/7.2.1/../../../../include/c++/7.2.1/bits/std_function.h:316:2: runtime error: call to function Botan::redc_p384(Botan::BigInt&, std::vector<unsigned long, Botan::secure_allocator<unsigned long> >&) through pointer to incorrect function
 type 'void (*)(Botan::BigInt &, std::vector<unsigned long, Botan::secure_allocator<unsigned long> > &)'
(/home/jack/work/botan/libbotan-2.so.4+0x15d0fb0): note: Botan::redc_p384(Botan::BigInt&, std::vector<unsigned long, Botan::secure_allocator<unsigned long> >&) defined here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /bin/../lib64/gcc/x86_64-pc-linux-gnu/7.2.1/../../../../include/c++/7.2.1/bits/std_function.h:316:2 in 
```

Maybe some bad interaction between Clang and `libstdc++`? Unfortunately `libc++` doesn't seem to work on my machine.

mgierlings:
Calls `Botan::redc_pXXX` directly inside non-capturing lambda function,
which can be converted to `std::function<void (...)>`, instead of passing an
incompatible `void(*)` to `NIST_Curve_Reduction_Tests::random_redc_test`.

### DIFF
--- a/src/tests/unit_ecc.cpp
+++ b/src/tests/unit_ecc.cpp
@@ -188,12 +188,42 @@ class NIST_Curve_Reduction_Tests final : public Test
          std::vector<Test::Result> results;
 
 #if defined(BOTAN_HAS_NIST_PRIME_REDUCERS_W32)
-         results.push_back(random_redc_test("P-384", Botan::prime_p384(), Botan::redc_p384));
-         results.push_back(random_redc_test("P-256", Botan::prime_p256(), Botan::redc_p256));
-         results.push_back(random_redc_test("P-224", Botan::prime_p224(), Botan::redc_p224));
-         results.push_back(random_redc_test("P-192", Botan::prime_p192(), Botan::redc_p192));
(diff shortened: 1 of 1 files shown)


randombit:
Thanks. I'd be happier if there was a clear reason this was required but I'll take it.

PR #1374, opened 2017-12-27
Remove the unused XMSS-WOTS operations

randombit:
@mgierlings I noticed that after the recent changes the XMSS WOTS operation types are not used (and afaict not usable in any way, since the WOTS key types are not exposed either). Is it ok to just remove them at this point?


mgierlings:
Yes, that is perfectly fine. I implemented them to stay consistent with the library structure. In theory they are a distinct cryptographic scheme. But, as stated in my original XMSS pull request, I don't see any practical use outside XMSS. Removing this will improve coverage and in general it is a good thing to avoid dead code. If we ever need the XMSS WOTS standalone interface we can easily reimplement it. So +1 for removing it from my side.

randombit:
OK thanks

PR #1381, opened 2018-01-03
detect the ca issuers field(1.3.6.1.5.5.7.48.2) in x509

xqp:
We need a solution to detect the ca issuer URI from the x509 certificates with botan
With these patch we can extract the information about the ca issuer from the certificate with the CLI tool and programmatically with the library.

http://oid-info.com/get/1.3.6.1.5.5.7.48.2
http://www.ietf.org/rfc/rfc2459.txt


randombit:
Thank you! You should know that we are now less than 1 week to the 2.4.0 release and well into the feature freeze period so I'm hesitant to merge this right away. But it's a pretty small change so maybe ok for 2.4.0 if we can get the below issues addressed quickly.

- If it is possible can you please also attach a certificate which contains this field so we can add a test for decoding?

- It looks like you edited `src/lib/asn1/oids.cpp` directly to add the OID. Instead add it to `src/build-data/oids.txt` and then generate the source file with `./src/scripts/oids.py oids > src/lib/asn1/oids.cpp`

- In `x509_ext.cpp` instead of `std::for_each` just use a C++11 for-loop (`for (std::string ca_issuer : m_ca_issuers)`).

Other than that lgtm.

randombit:
Thanks for making those changes. As to the test certificate I see the current cert for `microsoft.com` has AIA CA issuers field so we can just use that.

reneme:
We just used some test certificates from our internal CA that contain an AIA with two CA issuers. Thanks a lot.

randombit:
Perfect thank you

PR #1383, opened 2018-01-04
Add note on OAEP version implemented

securitykernel:
When working on a TPM 2.0 software stack, we noticed that OAEP as implemented in Botan did not work with a TPM 2.0. As it turned out, TPM 2.0 uses the PKCS#1 v2.1 (RFC 3447) OAEP version which prepends a zero byte, whereas Botan implements the encoding according to PKCS#1 v2.0 (RFC 2437). Here we only add a note which PKCS#1 version Botan implements.


randombit:
Thanks. I was not aware there were two incompatible encodings of OAEP. We'll probably want to address this more fully in a future release by supporting both

PR #1387, opened 2018-01-04
Print a short command description for cli help

securitykernel:
In order to improve the CLI help, adds a description for each command and prints it git style when CLI is called with no arguments or with `--help` or `-h`. Adds pure virtual functions `group()` and `description()` to `Command` which are used in the output of the help command.

I plan on extending `Command::help_text()` to provide a help text similar to other CLI programs in a later PR.

The new output of `botan --help` looks like this:

```shell
$ ./botan
Usage: botan <cmd> <cmd-options>
All commands support --verbose --help --output= --error-output= --rng-type= --drbg-seed=

Available commands:

Compression:
   compress           Compress a given file
   decompress         Decompress a given compressed archive

Encryption:
   encryption         Encrypt or decrypt a given file

Hash Functions:
   hash               Compute the message digest of given file(s)

HMAC:
   hmac               Compute the HMAC tag of given file(s)

Miscellaneous:
   asn1print          Decode and print file with ASN.1 Basic Encoding Rules (BER)
   base64_dec         Decode Base64 encoded file
   base64_enc         Encode given file to Base64
   cc_decrypt         Decrypt the passed valid ciphertext credit card number using FPE decryption
   cc_encrypt         Encrypt the passed valid credit card number using FPE encryption
   config             Print the used prefix, cflags, ldflags or libs
   cpuid              List available processor flags (aes_ni, SIMD extensions, ...)
   hex_dec            Hex decode a given file
   hex_enc            Hex encode a given file
   http_get           Retrieve resource from the passed http/https url
   rng                Sample random bytes from the specified rng
   speed              Measures the speed of algorithms
   timing_test        Run various timing side channel tests
   version            Print version info

Number Theory:
   factor             Factor a given integer
   gen_prime          Samples one or more primes
   is_prime           Test if the integer n is composite or prime
   mod_inverse        Calculates a modular inverse

Password Hashing:
   check_bcrypt       Checks a given bcrypt hash against hash
   gen_bcrypt         Calculate the bcrypt password digest of a given file

PSK Database:
   psk_get            Read a value saved with psk_set
   psk_list           List all values saved to the database
   psk_set            Save a PSK encrypted in the database

Public Key Cryptography:
   dl_group_info      Print raw Diffie-Hellman parameters (p,g) of the standarized DH group name
   ec_group_info      Print raw elliptic curve domain parameters of the standarized curve name
   fingerprint        Calculate a public key fingerprint
   gen_dl_group       Generate ANSI X9.42 encoded Diffie-Hellman group parameters
   keygen             Generate a PKCS #8 private key
   pkcs8              Open a PKCS #8 formatted key
   sign               Sign arbitrary data
   verify             Verify the authenticity of the given file with the provided signature

TLS:
   tls_ciphers        Lists all ciphersuites for a policy and TLS version
   tls_client         Connect to a host using TLS/DTLS
   tls_client_hello   Parse a TLS client hello message
   tls_http_server    Provides a simple HTTP server
   tls_proxy          Proxies requests between a TLS client and a TLS server
   tls_server         Accept TLS/DTLS connections from TLS/DTLS clients

X.509:
   cert_info          Parse X.509 certificate and display data fields
   cert_verify        Verify if the passed X.509 certificate passes path validation
   gen_pkcs10         Generate a PKCS #10 certificate signing request (CSR)
   gen_self_signed    Generate a self signed X.509 certificate
   ocsp_check         Verify an X.509 certificate against the issuers OCSP responder
   sign_cert          Create a CA-signed X.509 certificate from a PKCS #10 CSR

```

### DIFF
--- a/src/cli/asn1.cpp
+++ b/src/cli/asn1.cpp
@@ -21,6 +21,16 @@ class ASN1_Printer final : public Command
    public:
       ASN1_Printer() : Command("asn1print --pem file") {}
 
+      std::string group() const override
+         {
+         return "misc";
+         }
+
+      std::string description() const override
+         {
+         return "Decode and print file with ASN.1 Basic Encoding Rules (BER)";
+         }
+
       void go() override
          {
          const std::string input = get_arg("file");
--- a/src/cli/cc_enc.cpp
+++ b/src/cli/cc_enc.cpp
@@ -119,6 +119,16 @@ class CC_Encrypt final : public Command
    public:
       CC_Encrypt() : Command("cc_encrypt CC passphrase --tweak=") {}
 
+      std::string group() const override
+         {
+         return "misc";
+         }
+
+      std::string description() const override
+         {
+         return "Encrypt the passed valid credit card number using FPE encryption";
+         }
+
       void go() override
          {
          const uint64_t cc_number = std::stoull(get_arg("CC"));
@@ -143,6 +153,16 @@ class CC_Decrypt final : public Command
    {
    public:
       CC_Decrypt() : Command("cc_decrypt CC passphrase --tweak=") {}
+
+      std::string group() const override
+         {
+         return "misc";
+         }
+
+      std::string description() const override
+         {
+         return "Decrypt the passed valid ciphertext credit card number using FPE decryption";
+         }
 
       void go() override
          {
--- a/src/cli/cli.h
+++ b/src/cli/cli.h
@@ -86,6 +86,10 @@ class Command
 
       int run(const std::vector<std::string>& params);
 
+      virtual std::string group() const = 0;
+
+      virtual std::string description() const = 0;
+
       virtual std::string help_text() const;
 
       const std::string& cmd_spec() const
@@ -98,7 +102,7 @@ class Command
    protected:
 
       /*
-      * The actual functionality of the cli command implemented in subclas.
+      * The actual functionality of the cli command implemented in subclass.
       * The return value from main will be zero.
       */
       virtual void go() = 0;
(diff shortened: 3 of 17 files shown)


randombit:
I like it!

Only thing that bugs me here is in the help printing loops itself, as it is O(#groups * #cmds). Which is not a problem in practice, here, but offends my sensibilities ;)

Related, the ifdefs in the help output are fragile, eg if we added some non-bcrypt password hashing utils but forgot the update `help_text` to check for that macro, then if bcrypt was disabled the (non-bcrypt) password hash cmds wouldn't be listed in the help output. Which is a really easy bug to miss. Similarly, if a new group is added, associated commands will not appear in the help output unless the user also remembers to update this other function.

I think both problems can be solved easily
- Skip the ifdef checks when creating `groups_description`
- Iterate over `Command::registered_cmds`, call `Command::get_cmd` for each, and populate a map of `map<string, std::vector<unique_ptr<Command>>>` - mapping the group onto a list of Command objects. Basically `if(cmd) { m[cmd->group].push_back(move(cmd)); }`
- Iterate over the map (now ordered by grouping, and with each sublist ordered by name -
 this is because return of `registered_cmds` is sorted) and print the output. If a group doesn't have a pretty name (in `groups_description`) just print the basic group string instead.

If a group happens to be empty (eg because HMAC is disabled), the relevant key in the map would just not be populated, without anyone having to remember to manually change anything within this function.

I don't think there is any reason to expose `Command::global_registry`, with `Botan::map_keys_as_set(Botan_CLI::Command::global_registry())` you are just duplicating result of `Botan_CLI::Command::registered_cmds()`.

securitykernel:
Thanks for the comments. I‘ll address them sometime in the next week. As 2.4 is feature freeze we won‘t merge it before Monday anyway.

securitykernel:
Comments addressed, now the code is much cleaner :). I also rebased onto master.

randombit:
Looks great thank you

randombit:
Ughhhh I merged without pulling the new PR into to my workspace :( I'll fix

PR #1388, opened 2018-01-04
Change default userid for SM2 signatures.

dewyatt:
GM/T 0009-2012 specifies that the SM2 userid should default to the 16-byte hex sequence 31,32,33,34,35,36,37,38,31,32,33,34,35,36,37,38.

See [here](https://riboseinc.github.io/gbt-0009-2012/) for more info.

Since other implementations default to this value and the standard calls for it, it seems like it would be best to follow suit here. With this PR, a blank userid will actually use the value specified in the standard.


randombit:
Thanks

PR #1399, opened 2018-01-09
Remove dependency: base depends on mac

webmaster128:


randombit:
OK I guess, but why?

One thing, `lookup.h` should have guards for MACs being disabled. You probably missed this because nothing in the lib or tests includes that header.

webmaster128:
This came to my mind when thinking about a "hash depends on mac" dependency.

Just a low hanging fruit trying to reduce complicated and especially circular dependencies. I think base should not depend on anything in order to move from a chaotic dependency graph to something that looks more tree-ish. With this change, you can build the library without the mac module and if needed, the dependency should be added explicitly or via parent-folder dependencies.

webmaster128:
Thanks for feedback, updated.

webmaster128:
:+1:

Issue #1397, opened 2018-01-09
Keyed-hashing support for blake2


webmaster128:
I tried to implement the blake2b self test from [RFC 7693](https://tools.ietf.org/html/rfc7693#appendix-E) but found that keyed hashing is not supported in Botan.

Are there any plans to implement keyed hashing for blake2 or the hash interface in general?

randombit:
I think it would be better to expose a hash that supports keying as a MAC, instead of tacking on keying operations to the hash function interface.

But it's hard to see a good way to implement this without duplicating a lot of code. The best I can come up with is implement `BLAKE2b_MAC` and move most of the hash logic (eg the compression function) there. Then implement the hash function in terms of the MAC, just always setting the key to empty. `MessageAuthenticationCode::start_msg` also adds a place to support the BLAKE salting mode.

PR #1400, opened 2018-01-09
Add keyed Blake2b [WIP]

webmaster128:
this is a proof-of-concept for keyed Blake2b (see #1397). It follows the idea used in skein where additional data is added when constructing a Blake2b hasher.

In my opinion this interface is far from beautiful but it is a small (probably compatible) change to the existing interface without the need to add a mac module just to add the key argument to the Blake2b constructor.

I'd be happy to throw this away if there is a more elegant solution.


randombit:
Interesting approach. Have to think about this one more.

randombit:
Sorry it took so long to get back to you on this. I don't think this is the right approach. It does offer keying but prevents all the usual keying mechanisms from working (eg you can't change the key, query the key length, ...). I think the algo parameter should be limited to "fixed" values like the output length or personalization field.

Issue #1411, opened 2018-01-14
random_safe_prime() is slow when using DL_Group::Strong


bellgrim:
I am having issues with creating a DL_Group for DH. When tracing the code, it looks like it cannot find a random_safe_prime. It gets stuck on an infinite loop (I aborted after some minutes): https://github.com/randombit/botan/blob/master/src/lib/math/numbertheory/make_prm.cpp#L121

The DL_Group is created similar to this:

```
Botan::RandomNumberGenerator* rng = new Botan::AutoSeeded_RNG();
Botan::DL_Group* group = new Botan::DL_Group(*rng, Botan::DL_Group::Strong, 1024);
```

I do not have this issue on all of my machines, only Ubuntu 14.04 (64-bit). Do you have any hints on how I can debug this?

randombit:
Second 1024-bit generator finished after exactly 4 minutes.

bellgrim:
It did return a value, but after 11 minutes.

```
$ time botan gen_dl_group --type=strong --pbits=1024
...
real	11m8.379s
user	11m0.733s
sys	0m7.156s
```

Shall run my code again and see how long it takes.

bellgrim:
My code returned a result after 12 minutes. So it is not an infinite loop, just that it takes a long time.

randombit:
That's too slow. We're doing something wrong here.

randombit:
The paper https://eprint.iacr.org/2003/186.pdf was helpful.

First, we should generate q to be == 2 mod 3, otherwise 2*q+1 will be a multiple of 3 and never prime.

Secondly, using a larger sieve seems to be helpful.

Also, there is no reason to do a very strong prime test on q until we have verified that p is prime. As it is we go to great length making sure q is prime, then throw it away because p is obviously not prime. So we should switch to a 2-stage verification of q, first a fast check, then check 2*q+1, then if p proves to be prime go back and do a stronger check of q.

With those changes, generating a 1024-bit safe prime is taking 10-40 seconds.

PR coming soon.

randombit:
Should be fixed by #1413, now seeing 1024-bit strong DH params in 1 - 20 seconds. Would be great if you can confirm, especially with your VM.

bellgrim:
Thanks for the fix. This was a real improvement!

```
$ time botan gen_dl_group --type=strong --pbits=1024
....
real	0m3.777s
user	0m3.753s
sys	0m0.013s

[...]

PR #1413, opened 2018-01-16
Improve speed of prime generation especially safe primes

randombit:
First, correct a bug in the sieve code. It would break early if a value did not pass the sieve. However in that case, the sieve values would be out of sync with the value of p, and would be returning effectively random results. This caused prime generation to be slower than it should be, both because the sieve was incorrectly rejecting values that were not multiples of any small prime and was allowing values that were multiples of small primes to move on to the Miller-Rabin test. (That bug was introduced in 7335eefcf)

In the sieve, also sieve so that 2*q+1 is also not a multiple of the small primes. This speeds up safe prime generation.

On master, RSA-3072 keygen takes ~~ 2-4 seconds. With this change .5-2 seconds.

On master, 1024-bit strong DH group gen typically takes over 60 seconds and sometimes much longer. With this change, over 10 trials, max was 15 seconds, min .5 seconds, median 9 seconds. Also ran 2048-bit strong prime gen several times, saw results in under a minute sometimes as fast as 10 seconds.

Fixes GH #1411

### DIFF
--- a/src/lib/math/mp/mp_core.cpp
+++ b/src/lib/math/mp/mp_core.cpp
@@ -436,10 +436,14 @@ word bigint_divop(word n1, word n0, word d)
 */
 word bigint_modop(word n1, word n0, word d)
    {
+#if defined(BOTAN_HAS_MP_DWORD)
+   return ((static_cast<dword>(n1) << MP_WORD_BITS) | n0) % d;
+#else
    word z = bigint_divop(n1, n0, d);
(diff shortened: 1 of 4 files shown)


randombit:
Interesting this seems to cause some kind of horrible perf regression in other tests investigating

randombit:
Oh oops because the sieve actually works we loop forever trying to generate very small primes

PR #1414, opened 2018-01-17
Do not generally choose eg++ compiler on OpenBSD.

bluhm:
As three OpenBSD architectures have moved to clang as base compiler,
use clang if it is available.  Use gcc from ports as fallback.  Then
the binary name has to be set manually as auto configuration would
confuse the OpenBSD ports framework.  The old assembler is used
only by gcc.


randombit:
Thanks!

PR #1415, opened 2018-01-17
Adds the feature to register custom elliptic curves at runtime and to use them in TLS

neverhub:
Back in https://github.com/randombit/botan/pull/710 @securitykernel introduced the house-curve feature. This feature allows us to set a single application specific curve at compile time. While this practically enables the use of a custom curve in TLS it is arguably highly restrictive.

The requested changes allow the import of multiple custom curves at once at runtime and to use them for TLS. Therefore i made the following changes

1. OID registry with ability to register OIDs at runtime (as it was before https://github.com/randombit/botan/commit/62e55f484a7a03e2532875696eb2479a577878e9)
2. dynamic registry of custom EC_Groups
3. dynamic registry of custom TLS CurveIDs

With these changes it is now possible to register custom EC_Groups with name, OID and TLS CurveID. After a successful registration it is now possible to use the curve in TLS (if the TLS policy allows it). To verify that the implementation works as intended please take a quick look at https://github.com/Hackmanit/botan/tree/custom_curves_cli.

For testing purposes i registered the testcurves curves in `ecc_custom_params.txt` to the 
reserved range 0xfe00 to 0xfe05 and allowed them in the tls policy `(tls_ecc_custom.txt)`. In my tests Botan was able to complete TLS 1.2 handshakes (`ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256`) on the registered testcurves.

I look forward to your comments and naturally appreciate further testing.

rel.: https://github.com/randombit/botan/pull/710, https://github.com/randombit/botan/issues/636


randombit:
Thanks. It will take me a while to review this fully and I think some changes will be required but I'll get back to you soon on that.

One question though, it seems like this ability completely displaces the existing house curve support. Does it make sense as part of this change to also remove that code?

neverhub:
I think it makes perfect sense to remove the house curve as part of this change. I actually wanted to do this before before creating the pull request. I will add this very soon.

randombit:
OK some initial thoughts

- Adding back modifyable OIDs - don't really like bringing back a global lock here, but probably fine

- I don't like the TLS curve ID registry. It feels very heavy to have all this machinery just to look up a uint16_t (and only in an uncommon case, at that). I don't have concrete suggestions on a different approach just yet. I'm wondering though, maybe it be better to associate the TLS curve ID with the `EC_Group` itself? Or can we leverage `TLS::Callbacks` somehow? (Eg adding a new callback that applications which want to use a custom curve must implement.)

- Broadly speaking the custom curve registry is fine, except I don't think it goes far enough, in that I'd like it to apply to all the curves (rather than treating the builtin curves and the "custom" curves differently). There should be a registry mapping OID to the EC group data, on first use of a particular curve this is populated from the hardcoded data, or alternately the application can insert data directly. Basically the same format as you implement for custom curves could also be used to specify the builtin curves (instead of using PEM formatting, which is really opaque). #1435 changes the groups to use a shared representation which is a push in this direction.

Plan of action?
 - You create a new PR with just the first 2 commits (modifying OIDs). I'll merge that.
 - I'll merge #1435 
 - I'll write some `EC_Group` registry code, probably using some code from this PR. (Leave this PR open as discussion area until we get everything sorted.)
 - Everyone considers other options for handling the TLS curve ids

randombit:
#1441 adds the EC_Group registry code, which I think will work out pretty well. One thing I especially like is it gives us a much easier path to removing weak curves (P-192, etc) since it will be possible - for applications that require them - to easily re-enable such curves without modifying the library.

I played with some options on handling the TLS curve IDs but didn't find anything really satisfying yet. It is kind of a mess since the same code has to handle ECC, DH, and x25519 (which yeah is ECC but we special case it). Also fitting it into the existing TLS::Policy APIs is hard. Anyway, will look at it more soon.

randombit:
Major bug affecting `EC_Group_Common` described in #1441 :/

PR #1421, opened 2018-01-19
Fix documentation of DataSource::end_of_data.

lambdafu:
It's very confusing if the text is inverted :)


randombit:
Oops thank you!

PR #1426, opened 2018-01-21
FFI function for DSA and Elgamal key generation

flowher:
*   Adds function for ElGamal key generation that allows
    usage of 'p' chosen by the caller.  
   
*   Adds function for DSA key generation that allows
    usage of 'p' and 'q' chosen by the caller.

This is a contribution from Ribose Inc (@riboseinc).

### DIFF
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -702,6 +702,7 @@ BOTAN_PUBLIC_API(2,0) int botan_privkey_create_ecdsa(botan_privkey_t* key, botan
 BOTAN_PUBLIC_API(2,0) int botan_privkey_create_ecdh(botan_privkey_t* key, botan_rng_t rng, const char* params);
 BOTAN_PUBLIC_API(2,0) int botan_privkey_create_mceliece(botan_privkey_t* key, botan_rng_t rng, size_t n, size_t t);
 BOTAN_PUBLIC_API(2,0) int botan_privkey_create_dh(botan_privkey_t* key, botan_rng_t rng, const char* param);
+BOTAN_PUBLIC_API(2,0) int botan_privkey_create_dsa(botan_privkey_t* key, botan_rng_t rng, size_t pbits, size_t qbits);
 
 /*
 * Input currently assumed to be PKCS #8 structure;
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -703,6 +703,7 @@ BOTAN_PUBLIC_API(2,0) int botan_privkey_create_ecdh(botan_privkey_t* key, botan_
 BOTAN_PUBLIC_API(2,0) int botan_privkey_create_mceliece(botan_privkey_t* key, botan_rng_t rng, size_t n, size_t t);
 BOTAN_PUBLIC_API(2,0) int botan_privkey_create_dh(botan_privkey_t* key, botan_rng_t rng, const char* param);
 BOTAN_PUBLIC_API(2,0) int botan_privkey_create_dsa(botan_privkey_t* key, botan_rng_t rng, size_t pbits, size_t qbits);
+BOTAN_PUBLIC_API(2,0) int botan_privkey_create_elgamal(botan_privkey_t* key, botan_rng_t rng, size_t pbits);
 
 /*
 * Input currently assumed to be PKCS #8 structure;
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -702,8 +702,56 @@ BOTAN_PUBLIC_API(2,0) int botan_privkey_create_ecdsa(botan_privkey_t* key, botan
 BOTAN_PUBLIC_API(2,0) int botan_privkey_create_ecdh(botan_privkey_t* key, botan_rng_t rng, const char* params);
 BOTAN_PUBLIC_API(2,0) int botan_privkey_create_mceliece(botan_privkey_t* key, botan_rng_t rng, size_t n, size_t t);
 BOTAN_PUBLIC_API(2,0) int botan_privkey_create_dh(botan_privkey_t* key, botan_rng_t rng, const char* param);
-BOTAN_PUBLIC_API(2,0) int botan_privkey_create_dsa(botan_privkey_t* key, botan_rng_t rng, size_t pbits, size_t qbits);
-BOTAN_PUBLIC_API(2,0) int botan_privkey_create_elgamal(botan_privkey_t* key, botan_rng_t rng, size_t pbits);
+
+
+/*
+ * Generates DSA key pair. Gives to a caller control over key length
+ * and order of a subgroup 'q'.
+ *
+ * @param   key   handler to the resulting key
+ * @param   rng   initialized PRNG
+ * @param   pbits length of the key in bits. Must be between in range (1024, 3072)
+ *          and multiple of 64. Bit size of the prime 'p'
+ * @param   qbits order of the subgroup. Must be in range (160, 256) and multiple
+ *          of 8
+ *
+ * @returns BOTAN_FFI_SUCCESS Success, `key' initialized with DSA key
+ * @returns BOTAN_FFI_ERROR_NULL_POINTER  either `key' or `rng' is NULL
+ * @returns BOTAN_FFI_ERROR_BAD_PARAMETER unexpected value for either `pbits' or
+ *          `qbits'
+ * @returns BOTAN_FFI_ERROR_NOT_IMPLEMENTED functionality not implemented
+ *
+-------------------------------------------------------------------------------- */
+BOTAN_PUBLIC_API(2,5) int botan_privkey_create_dsa(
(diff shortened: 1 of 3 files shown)


src/lib/ffi/ffi.h
```diff
@@ -702,6 +702,8 @@ BOTAN_PUBLIC_API(2,0) int botan_privkey_create_ecdsa(botan_privkey_t* key, botan
 BOTAN_PUBLIC_API(2,0) int botan_privkey_create_ecdh(botan_privkey_t* key, botan_rng_t rng, const char* params);
 BOTAN_PUBLIC_API(2,0) int botan_privkey_create_mceliece(botan_privkey_t* key, botan_rng_t rng, size_t n, size_t t);
 BOTAN_PUBLIC_API(2,0) int botan_privkey_create_dh(botan_privkey_t* key, botan_rng_t rng, const char* param);
+BOTAN_PUBLIC_API(2,0) int botan_privkey_create_dsa(botan_privkey_t* key, botan_rng_t rng, size_t pbits, size_t qbits);
+BOTAN_PUBLIC_API(2,0) int botan_privkey_create_elgamal(botan_privkey_t* key, botan_rng_t rng, size_t pbits);
```
randombit: These should be 2,5 since that is the first release they will appear.

src/lib/ffi/ffi_pkey_algs.cpp
```diff
@@ -333,6 +333,24 @@ int botan_pubkey_rsa_get_n(botan_mp_t n, botan_pubkey_t key)
...
+int botan_privkey_create_dsa(botan_privkey_t* key, botan_rng_t rng_obj, size_t pbits, size_t qbits)
+   {
+#if defined(BOTAN_HAS_DSA)
+
+    if(rng_obj == nullptr)
+      return BOTAN_FFI_ERROR_NULL_POINTER;
+
+    return ffi_guard_thunk(BOTAN_CURRENT_FUNCTION, [=]() -> int {
+      Botan::RandomNumberGenerator& rng = safe_get(rng_obj);
+      Botan::DL_Group group(rng, Botan::DL_Group::Prime_Subgroup, pbits, qbits);
```
randombit: For DSA this should probably instead use the NIST approved DSA parameter generator (DL_Group::DSA_Kosherizer)

src/lib/ffi/ffi_pkey_algs.cpp
```diff
@@ -333,6 +333,24 @@ int botan_pubkey_rsa_get_n(botan_mp_t n, botan_pubkey_t key)
    }
 
 /* DSA specific operations */
+int botan_privkey_create_dsa(botan_privkey_t* key, botan_rng_t rng_obj, size_t pbits, size_t qbits)
+   {
+#if defined(BOTAN_HAS_DSA)
+
+    if(rng_obj == nullptr)
```
randombit: Should also check that key is not null.

src/lib/ffi/ffi_pkey_algs.cpp
```diff
@@ -441,6 +459,24 @@ int botan_privkey_load_ecdsa(botan_privkey_t* key,
    }
 
 /* ElGamal specific operations */
+int botan_privkey_create_elgamal(botan_privkey_t* key, botan_rng_t rng_obj, size_t pbits)
+   {
+#if defined(BOTAN_HAS_ELGAMAL)
+
+    if(rng_obj == nullptr)
```
randombit: Also key

src/tests/test_ffi.cpp
```diff
@@ -1300,13 +1300,7 @@ class FFI_Unit_Tests final : public Test
          return result;
          }
```
randombit: Add a copyright notice to this file.

src/lib/ffi/ffi_pkey_algs.cpp
randombit: Also, otherwise it is possible to generate a DSA key with params that are considered invalid for DSA (eg a 1536 bit p with 384 bit q). Such params are fine, technically speaking, but some implementations may reject such keys outright for not following the NIST param lengths (1024+160, 2048+256, etc)

flowher: I'm not using DSA_Kosherizer because in OpenPGP we want to be able to generate any key between 1024 and 3074 with size that's multiple of 64.

src/lib/ffi/ffi_pkey_algs.cpp
```diff
@@ -441,6 +459,24 @@ int botan_privkey_load_ecdsa(botan_privkey_t* key,
...
+      return BOTAN_FFI_ERROR_NULL_POINTER;
+
+    return ffi_guard_thunk(BOTAN_CURRENT_FUNCTION, [=]() -> int {
+      Botan::RandomNumberGenerator& rng = safe_get(rng_obj);
+      Botan::DL_Group group(rng, Botan::DL_Group::Strong, pbits);
```
flowher: I've noticed that build for ARM fails because it takes long time to run tests. I think, it's because of the fact strong prime generation takes some time. I don't have immediate idea on how to improve performance here. So I'm wondering on either blocking the test or using Prime_Subgroup (ideally I would prefer strong primes, but from the other hand elgamal is not very popular anymore). What do you think?

src/tests/test_ffi.cpp
flowher: added

src/lib/ffi/ffi_pkey_algs.cpp
```diff
@@ -441,6 +465,28 @@ int botan_privkey_load_ecdsa(botan_privkey_t* key,
...
+    if ((rng_obj == nullptr) || (key == nullptr))
+      return BOTAN_FFI_ERROR_NULL_POINTER;
+
+    if (pbits < 1024) {
+      return BOTAN_FFI_ERROR_BAD_PARAMETER;
+    }
+
+    return ffi_guard_thunk(BOTAN_CURRENT_FUNCTION, [=]() -> int {
+      Botan::RandomNumberGenerator& rng = safe_get(rng_obj);
+      Botan::DL_Group group(rng, Botan::DL_Group::Strong, pbits);
```
flowher: Test for ARM fails because generation of strong primes is slow. I unfortunately don't have immediate idea on how perf could be improved (if possible at all). I'm wondering on either blocking the test or using Prime_Subgroup (q=0). I would prefer strong primes, but from the other hand elgamal is not very popular anyway. Opinions?

randombit: After #1413 it is probably not going to be easy to speed up safe prime gen too much. If it is taking 10 minutes under qemu, it seems like even if we really work on it and cut runtime in half that still means this single test adds 5 minutes per qemu test run, which is pretty horrible in the long term.

FWIW libgcrypt's ElGamal uses smaller subgroups. (Not that libgcrypt is necessarily the best guide for the right thing to do, but it does indicate current behavior of GPG). See `cipher/elgamal.c`.

Easiest resolution I think is, if qbits == pbits-1, generate a safe prime, otherwise a prime subgroup. Have the test use a 1024/160 prime, which should be fast, and than do safe primes in your production keygen.

src/lib/ffi/ffi.h
```diff
@@ -702,6 +702,8 @@ BOTAN_PUBLIC_API(2,0) int botan_privkey_create_ecdsa(botan_privkey_t* key, botan
 BOTAN_PUBLIC_API(2,0) int botan_privkey_create_ecdh(botan_privkey_t* key, botan_rng_t rng, const char* params);
 BOTAN_PUBLIC_API(2,0) int botan_privkey_create_mceliece(botan_privkey_t* key, botan_rng_t rng, size_t n, size_t t);
 BOTAN_PUBLIC_API(2,0) int botan_privkey_create_dh(botan_privkey_t* key, botan_rng_t rng, const char* param);
+BOTAN_PUBLIC_API(2,5) int botan_privkey_create_dsa(botan_privkey_t* key, botan_rng_t rng, size_t pbits, size_t qbits);
```
randombit: One more thing - these really should have Doxygen comments of some kind. Can be brief. Should mention the handling of qbits for `create_elgamal` as well as restrictions on sizes.

src/lib/ffi/ffi_pkey_algs.cpp
flowher: Thanks. I havn't checked libgcrypt implementation before, but now I see they mention work of Weiner in paper from  eurocrypt'96. I need to take a look at it, not familiar with it yet.

I've used your suggestion about playing with qbits and pbits and now I use smaller key in test so it will be faster

flowher:
Execution time looks OK  to me now, just stuck in queue

| Mode | New FFI | PR #1415  |
|---|---|---|
| BUILD_MODE="shared" |  4 min 18 sec | 5 min 47 sec |
| BUILD_MODE="cross-arm32" |16 min 51 sec |  13 min 32 sec  |
| BUILD_MODE="cross-arm64" | 16 min 51 sec | 18 min 21 sec |

PR #1432, opened 2018-01-29
[WIP] Add pairings over 256-bit Barreto Naehrig curve

randombit:
- [ ] Provide an abstract interface? TBH I'm not even sure how to approach this in C++ given the need for several sub classes (G1,G2,GT) which are associated with the pairing.
- [ ] Optimizations. Quite slow due to excessive dynamic allocation. Should use a sparse representation over 5 64-bit words, hardwire the Montgomery params in GFp1 class.
- [ ] Add hashing on the curve (APIs are there but not implemented)
- [ ] Documentation
- [ ] More tests (going to fuzz vs dclxvi implementation to generate test vectors and provide more assurance of correctness)


randombit:
Closing, this is never going to merge. BLS12-381 is ecosystem wise the correct choice at this point, also a new pairing implementation should certainly follow in the footsteps of pcurves rather than using BigInt for anything.

PR #1433, opened 2018-01-30
Avoid dynamic allocation at Keccak_1600

deemru:
Avoid an unnecessary dynamic allocation of padding at Keccak_1600


randombit:
Thank you. This is a good change. I must have been being lazy when I wrote that code. On looking at what we do here, I wonder why we allocate anything at all? Mostly we are creating a bunch of zero bits, which we dutifully xor into the sponge. It is better to just XOR the 2 padding bytes into the state directly - as in #1434 -  so I am closing this in favor of that change.

PR #1436, opened 2018-02-01
Load every certificates of files found.

souch:
This is especially useful for CentOS/RedHat where the root CA certificates are bundled into one file.


randombit:
Thanks. A test would be good. If you can write it that would be great, otherwise I will merge this when I have time to write one. There are existing test data files that have multiple concatenated certs such as `src/tests/data/x509/x509test/ValidCert.pem`, just load this file and verify that you are able to look up each cert.

souch:
Ok, I'll make a test ASAP. 
I do not understand why some codecov checks failed ?

randombit:
> I do not understand why some codecov checks failed ?

"0% of diff hit" = no test, so the changed code was not hit during the test run.

The failure related to overall % drop is not your fault, the problem is some Travis CI builds run on CPUs with AVX2 and others do not, so there is a large % drop when codecov compares a run that did have AVX2 with one that did not, because suddenly all the AVX2 specific code is all skipped.

randombit:
@souch Thanks test looks good. Two small changes required. First, the `#includes` in the test should be inside the `#if defined(BOTAN_HAS_CERTSTOR_SQL)` block otherwise the test will fail to compile if x509 code is disabled (as seen in https://travis-ci.org/randombit/botan/jobs/336564412). Also, `bundledcertdir` should not be under `x509test`, that directory should only contain values generated by https://github.com/yymax/x509test. Move it to under `src/tests/data/x509/misc`.

randombit:
Thanks!

souch:
I did not see your message, so I didn't move cert to src/tests/data/x509/misc . I will make another pull request to fix that monday.

randombit:
No worries I fixed it post-merge.

souch:
Ok, thanks.

PR #1441, opened 2018-02-04
Create a persistent registry for ECC group data

randombit:
Now a single copy is maintained of each EC group info

Discussion in #1415

### DIFF
--- a/doc/manual/deprecated.rst
+++ b/doc/manual/deprecated.rst
@@ -58,7 +58,10 @@ in the source.
 
 - Old (Google specific) ChaCha20 TLS ciphersuites
 
-- All built in ECC groups < 256 bits
+- Weak or rarely used ECC builtin groups including "secp160k1", "secp160r1",
+  "secp160r2", "secp192k1", "secp192r1", "secp224k1", "secp224r1",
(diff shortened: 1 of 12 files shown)


randombit:
I discovered a total show stopper bug in this just before merging. If the ``mlock_allocator`` destructor runs first, then the `BigInts` cached in the EC group info structures will try to zero memory that has already been `munmap'ed`. By chance the tests didn't run into it but I would see it with `./botan ec_group_info secp256r1`.

The exact same problem affects ``EC_Group_Custom`` map in #1415

I really don't know how to fix this. Best I can come up with is `secure_allocator` could hold a `shared_ptr` reference to the pool allocator, and then we are assured the pool allocator would be freed until all `secure_vector`s had been destroyed. However that brings atomic operations in all over the place where they weren't before, I don't know what implications it has for performance or code size. Or ditch the mlock allocator entirely, but that seems a little harsh.

randombit:
Ugh just adding a shared_ptr to the allocator increases library binary size by over 15%

randombit:
OK I guess another option (though definitely feels like a hack) would be to ensure we reference the allocator before creating `EC_Group_Map` since (AIUI) that ensure the allocator destructor runs after the `EC_Group_Map` destructor.

Issue #1396, opened 2018-01-08
Botan 2.5.0 release


randombit:
Release date April 2, 2018, feature freeze March 23

- [ ] Improve tests of Python module
- [x] Update docs wrt #1448 #1395 #1394 
- [ ] Fix check_crl_online #951 
- [x] Fix DSA #1502
- [x] Fix #210
- [x] Support multiple DNS names in `gen_pkcs10` and `gen_self_signed`
- [x] Fix #983
- [x] Check #500
- [x] Fix random_integer test #1476
- [x] Issue burndown: all tickets either categorized or closed
- [x] Improve tests of CLI util
- [x] Fix Windows build issues (in #1456)
- [x] Fix timing_test cli (#1465)
- [x] Refactor CurveGFp and EC_Group to support shared representation
- [x] Make use of shared repr and precomputations to optimize scalar mult
- [x] AES for POWER (#1206)
- [x] Clean up x25519 code (#1391)
- [x] Finish and merge #1285, support PSS in TLS
- [x] Add ability to control/customize TLS extensions (#1186)
- [x] Add callback for client auth giving list of CAs (#1261)
- [x] Remove vestigial compression support from TLS

randombit:
Tagged and released.

PR #1448, opened 2018-02-08
Support custom curves in TLS

randombit:
Same idea as #1415 but different approach. Adds a new TLS callback `std::string tls_decode_group_param(Group_Params group_param)` which, for applications that require custom curves, they must implement in order to match against their custom curve IDs. This also works for DH groups, as long as the custom assignment is within the DH range (256-511, FFDHE leaves 508-511 for private assignment).

This is a little contorted in that an application that wants to use a custom curve must both specially set their Policy (to return the curve ID) and implement this additional Callback to handle decoding the curve ID to a string. But this really seems like a niche case, and it avoids having a special global (+ global lock) just to maintain this mapping.

I'm not actually 100% sure this is the right approach either, and am contemplating trying to include the TLS ids within the EC_Group/DL_Group data directly. In which case the extra callback would go away, and the TLS id would just be specified when the ECC group is created. However this requires a big refactoring of DL_Group (ala #1435/#1441) so may take a while.

Anyway posting now for @neverhub and @securitykernel to look at.


randombit:
I wrote up an implementation of shared_ptr DL_Groups and then realized it's probably a bad idea since the set of modp groups is basically unbounded, and keeping around a reference to every DL_Group an application ever processes is a recipe for trouble. We might be able to work around this by using `weak_ptr` in the registry but things get increasingly complicated for a seemingly small gain.

I'm still not completely happy with this PR but it's good enough I think, as long as @neverhub thinks it meets the same goals as #1415

randombit:
Rebased

securitykernel:
I contacted @neverhub last week, but it seems he's not available. First of all, thanks for implementing this! To me it seems like it handles the use case just fine. And I like the TLS callback and policy more than encapsulating the curve ID in the DL/EC group. Modulo docs (aka handbook) I think we're set with what we wanted to achieve with #1415.

neverhub:
Sorry for the late reply. I was quite busy recently :/. I did a bunch of testing today and yesterday and i am very satisfied with the callback approach. Thank you Jack :+1:  .

This achieves all set goals of #1415. I still have a few tests with more exotic curves at hand. I will add them, if you think they are helpful.

I think i can close #1415 now.

randombit:
@securitykernel This still needs docs but I'm going to merge now to minimize risk of conflicts, I've left a note in the 2.5 release ticket (#1396) that docs are still needed.

securitykernel:
@randombit Thanks! We‘ll take care of it.

PR #1450, opened 2018-02-12
[WIP] Fix make error on windows

pshurgal:
Fixed make error:
make: *** No rule to make target '.\libbotan-2.a', needed by 'libs'.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -55,6 +55,11 @@ def normalize_source_path(source):
     cmake needs this, and nothing else minds
     """
     return os.path.normpath(source).replace('\\', '/')
+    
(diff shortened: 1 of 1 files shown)


randombit:
Change seems ok but can you be more clear about the problem and how this fixes it? Is this a problem one runs into with any MinGW build or in a specific configuration?

Also there are a couple of pylint failures in Travis build, should be easy to fix.

webmaster128:
Is this because you're using a version of python with `os.path.sep` of `\`  in MinGW?

randombit:
Should be fixed now (in a slightly different way) by #1475. If you see any additional problems with MinGW build don't hesitate to open an issue.

PR #1454, opened 2018-02-19
Used shared_ptr rep for DL_Group

randombit:
Moves precomputations for mod-p reductions and g^x mod p calcs to the DL_Group itself since they can be shared by all users.


randombit:
Merged after rebase

Issue #1458, opened 2018-02-23
Soft-fail certificate revocation check implementation


souch:
HI,

I would like to have soft fail revocation check implemented in Botan. I don't think it can do that right now.
The best would be to have it on OCSP and CRL but right now only OCSP is plugged to the x509_path_validate. So I will try first to implement it with OCSP only.

But I don't know which way is best to implement it. And if you are ok with that. I will try to make a pull request once sthg ok has been done.
I first tried to add a parameter "revocation_softfail" to "Path_Validation_Restrictions" 

```
      * @param revocation_softfail if true, revocation check do not fail if
      *        revocation server is not available or revocation URL not set on
      *        certificate.
      */
      Path_Validation_Restrictions(bool require_rev = false,
                                   size_t minimum_key_strength = 110,
                                   bool ocsp_all_intermediates = false,
                                   bool revocation_softfail = false);
```

But I am a bit stuck right now on how to put that nicely in check_ocsp, check_ocsp_online,  Path_Validation_Result::successful_validation and x509_path_validate functions. I think I will need some new Certificate_Status_Code: NO_REVOCATION_URL and REVOCATION_SERVER_NOT_AVAILABLE.

I know this issue is a bit of a draft, but it is mainly to have your opinion and to have some guidelines and tips.

randombit:
If I'm understanding what you want here, I don't think there is any reason to add an additional argument to Path_Validation_Restrictions. Instead just add the new status codes as warnings, and insert them for every cert where no revocation data was available (that happens in `PKIX::merge_revocation_status`). Applications which ignore the warning state of path validation results will work as usual, but an application which cares about revocation status can check.

randombit:
Fixed with #1462?

souch:
Yep. At least for OCSP. I don't need CRL right now.

PR #1462, opened 2018-02-26
OCSP softfail revocation check

souch:
Here is a patch to have softfail OCSP revocation check:
- if the ocsp URL is empty or there is an exception in the POST_sync ocsp request, the status is set to OCSP_RESPONSE_GOOD and a warning is added to the status codes.
- it could be perhaps better to softfail in the POST_sync ocsp request only if the connection to the server fails (i.e.: do not softfail if connect ok but errors when reading data) ?
- to make the fewest modification possible, I added a special constructor to OCSP::Response that returns a specified status_code.
- note: like you said in the issue #1458, now, to have a softfail check one must use `Path_Validation_Result::successful_validation(),` to have a hardfail check one must call `Path_Validation_Result::no_warnings()` in addition
- if you think this pull request is ok, I can add a test.


randombit:
This seems ok. One thing I don't like so much is the addition of m_return_status to indicate that m_status_code is in use. It seems like instead you could check if `m_responses.empty()`, and if so return the "dummy" value. The only situation where this would be a problem is if we received a completely empty OCSP response from the server. That could be handled by pre-initializing `m_status_code` in all cases (currently it is left undefined unless the special constructor is used). `OCSP_RESPONSE_INVALID` seems like the correct default value.

Also the name of `m_return_status` should make it clear this is a faked up response code not associated with an actual response (just to avoid confusion for future readers). `m_dummy_response_status` or something like that.

randombit:
Changes look good. However on a final pass I do not understand this

```
+            else if(ocsp_signature_status == Certificate_Status_Code::OSCP_NO_REVOCATION_URL ||
+                    ocsp_signature_status == Certificate_Status_Code::OSCP_SERVER_NOT_AVAILABLE)
+               {
+               status.insert(ocsp_signature_status);
+               // softfail
+               status.insert(Certificate_Status_Code::OCSP_RESPONSE_GOOD); 
+               }
```

Specifically, I don't think we want to insert OCSP_RESPONSE_GOOD in there. That code means (quoting the to_string text) "OCSP response accepted as affirming unrevoked status for certificate" which is not at all what we want to claim, if the OCSP server was not available. I think instead you should just insert the warning status. Then certificate verification will still succeed, but not include the OCSP_RESPONSE_GOOD status.

souch:
I did not like the insert of `OCSP_RESPONSE_GOOD` too.
But, if I do not insert `OCSP_RESPONSE_GOOD`, the func `merge_revocation_status` will insert `NO_REVOCATION_DATA` which is an error status, then `Path_Validation_Result::successful_validation()` will return false.

souch:
Or I modify `merge_revocation_status` instead, e.g.:
```
if(code == Certificate_Status_Code::OCSP_RESPONSE_GOOD ||
   code == Certificate_Status_Code::OSCP_NO_REVOCATION_URL ||
   code == Certificate_Status_Code::OSCP_SERVER_NOT_AVAILABLE)
   {
   had_ocsp = true;
   }```

souch:
I am preparing a new pull request "ocsp_softfail_v2" that will, hopefully, take your remarks into account.

souch:
Moved to pull request  #1470 .

PR #1464, opened 2018-02-27
Added pkcs11 test results

perec-jar:
Added table with pkcs11 test results to doc/manual/pkcs11.rst.


randombit:
Thanks for running/documenting these tests. Hopefully the errors seen on CardOS can be easily addressed.

PR #1466, opened 2018-02-28
Add ChaCha20Poly1305 to CLI encryption tool

webmaster128:
Using the Botan CLI I was able to generate Test data for libsodium's crypto_aead_chacha20poly1305_ietf_encrypt/crypto_aead_chacha20poly1305_ietf_decrypt.

All green, so the implementations seem to match :)

Demo call: `echo -n "" | ./botan encryption --mode=chacha20poly1305 --iv=000000000000000000000000 --ad= --key=0000000000000000000000000000000000000000000000000000000000000000 | xxd -p`


randombit:
Nice thanks!

PR #1468, opened 2018-02-28
Path validation result warnings

souch:
Fix `no_warnings()` func : `find_warnings` add empty set into m_warnings vector even if no warning founds.


randombit:
Looks good thanks

PR #1470, opened 2018-03-01
OCSP softfail revocation check v2

souch:
I gathered pull request #1462 into one commit to make it cleaner.

If you are ok with this pull request, I think we should update the doc to explain how hardfail and softfail can be obtained (`success_validation()` + `no_warnings()`...). 
At the beginning, I started to add a param `revocation_softfail` to `Path_Validation_Restrictions`, so that by default, botan behave the same as before, and we have to explicitly enable softfail, and this can avoid the need to update the doc.


randombit:
Looks ok. Softfail is probably a more reasonable default and is (AIUI) how most systems work these days.

souch:
Yes AIUI softfail is the default, and OCSP stapling is recommended. Here is another commit that document the behaviour (softfail/hardfail) of Botan.

randombit:
Thanks

PR #1471, opened 2018-03-02
Fix check_crl_online segfault, need tests.

souch:
This pull request fix at least the segfaults :-(. I tried to make a test, but it takes quite some times to make it properly. And OCSP check will be good enough for our needs for the moment.


souch:
This pull request fixes also the problem: takes timeout time when there is no CRL URL.

randombit:
Yeah it is kind of a hard function to test well. I will merge as is since not crashing is certainly an improvement...

PR #1472, opened 2018-03-02
Implement product-scanning Montgomery reduction

randombit:
Results in approx 10-20% improvement for DH and RSA, 5% for ECC curves that use Montgomery form.


randombit:
Interesting, seems to actually cause a slowdown on x86-32 with gcc 7. However, it is also 10-12% faster on Aarch64 so I'm calling it a win.

randombit:
Also 10-20% faster on POWER8

PR #1481, opened 2018-03-09
Add missing include file to enable genentropy() on MacOS 10.12+

mouse07410:
MacOS 10.12 and newer support `getentropy()` function call. It is defined in the `<sys/random.h>` file, so that file should be included - but only if it's (a) MacOS, and (b) version 10.12 or newer.

### DIFF
--- a/src/lib/entropy/getentropy/getentropy.cpp
+++ b/src/lib/entropy/getentropy/getentropy.cpp
@@ -8,6 +8,18 @@
 #include <botan/internal/getentropy.h>
 
 #include <unistd.h>
+ 
+#if defined(__APPLE__)
+#if defined(__MAC_OS_X_VERSION_MIN_REQUIRED)
+#if __MAC_OS_X_VERSION_MIN_REQUIRED >= 101200
+/*
+ * To enable genentropy() method on MacOS, you also need
+ * to add "--with-os-features=getentropy
+ */
+#include <sys/random.h>
+#endif
+#endif /* __MAC_OS_X_VERSION_MIN_REQUIRED */
+#endif /* __APPLE__ */
 
 namespace Botan {
 


randombit:
Is there any reason for these version checks before including the header?

AIUI, if this is compiled on an older version of MacOS, not only does the header not exist, but `getentropy` will not either. So it will still not compile.

Why not just

```
#if defined(BOTAN_TARGET_OS_IS_DARWIN)
   #include <sys/random.h>
#endif
```

?

Also you commit message reads 10.11+ but the code only enables it for 10.12+

mouse07410:
> Is there any reason for these version checks before including the header?

Yes, because it not present in *any* Darwin - only Darwin 10.12 or newer.

> Also you commit message reads 10.11+ but the code only enables it for 10.12+

The code is right.

randombit:
> Yes, because it not present in any Darwin - only Darwin 10.12 or newer.

That's my point. What happens if you try to compile this file on 10.11?

mouse07410:
> That's my point. What happens if you try to compile this file on 10.11?

It will compile fine, and will *not* attempt to include `<sys/random.h>`. Which IMHO is right, because even if such a file exists on MacOS prior to 10.12 (and I've no more 10.11 to check that, nor do I care much), there won't be `getentropy()` call defined/available. So may as well not include it at all.

randombit:
Wait. You are saying, if compiled on 10.11 a function calling `getentropy` will compile fine even though `getentropy` does not exist? This seems ... not correct.

mouse07410:
> You are saying, if compiled on 10.11 a function calling getentropy will compile fine even though getentropy does not exist? This seems ... not correct.

**No**. I'm saying that **this file compiles correctly** on any MacOS (like it did before this PR), **but** it only includes `<sys/random.h>` if the version of MacOS is at least 10.12.

It does pull the necessary file on MacOS 10.12+, which is needed when `--with-os-features=getentropy` is configured. It does not include that file (because it probably does not exist, and it does not define `getentropy()` in any case) if MacOS version is 10.11 or earlier - in which case an attempt to configure `--with-os-features=getentropy` would cause compiler error, just like it does without this PR now.

What seems to be the problem/concern?

randombit:
I don't like ugly or confusing things.

mouse07410:
@ranfombit so what do you want it to look like to be non-confusing? It has to include that for only if it's Darwin, and then only if the target MacOS version is 10.12 or newer. If my guards/checks are ugly - how would you propose to do these checks?

mouse07410:
Again, this feature is available on Darwin, but only from 10.12, and it requires an extra include file. 

@randombit how do you propose to check for those two conditions, to ensure that <sys/random.h> is not included on systems/versions where it probably doesn't even exist?

randombit:
That is completely harmless. Because either the file fails to compile because sys/random.h doesn't exist. Or it fails to compile because `getentropy` is not defined. There doesn't seem much difference. Therefore, the version checks were unnecessary.

mouse07410:
If entropy.cpp gets compiled on MacOS only if the build is configured with "--with-os-features=getentropy" then indeed it's harmless, and we can omit the version check.

If however this file is compiled regardless - then the compilation would fail on systems where it used to compile ok.

randombit:
> If (get)entropy.cpp gets compiled on MacOS only if the build is configured with "--with-os-features=getentropy" then indeed it's harmless

This is the case. The entire module is enabled/disabled based on available OS feature flags.

mouse07410:
Ok. Would you like me then to update this PR, or open a new one? Or would you rather make that change yourself?

randombit:
Done in acecc97

mouse07410:
Thanks!

Issue #1490, opened 2018-03-14
Custom attributes in X509_DN are not encoded


phwork:
It is possible to add custom attributes to `X509_DN` by calling `add_attribute`.
The attributes are not written by `X509_DN::encode_into` because only a list of hardcoded attributes is considered.
This affects the certificate generation using `X509_CA::make_cert` and prevents the use of custom attributes in the subject. Is this behaviour intentional?

See also #1489 - same question here: are there already any plans to change the api?

randombit:
That behavior is intentional in the sense that I wrote the code and that's what it does. But it is not really optimal, no.

There probably shouldn't actually be any need to change the API here. Just encode all the DN entries rather than only picking the ones to encode out of a hardcoded list.

randombit:
Should be fixed by #1492

phwork:
Thanks, I don't know if I can test/verify it soon, but the tests lgtm.

randombit:
OK I believe everything here is all set  with #1492 merged, so I'm going to close the issue but feel free to reopen if something is missing.

PR #1492, opened 2018-03-14
Support custom DN entries

randombit:
GH #1490

### DIFF
--- a/src/lib/x509/x509_dn.cpp
+++ b/src/lib/x509/x509_dn.cpp
@@ -221,13 +221,13 @@ void X509_DN::encode_into(DER_Encoder& der) const
       {
       /*
       If we decoded this from somewhere, encode it back exactly as
-      we recieved it
+      we received it
       */
       der.raw_bytes(m_dn_bits);
       }
    else
       {
-      for(auto dn : m_dn_info)
+      for(const auto& dn : m_dn_info)
          {
          der.start_cons(SET)
             .start_cons(SEQUENCE)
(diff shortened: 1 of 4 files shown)


src/lib/x509/x509_dn.cpp
```diff
@@ -201,58 +210,32 @@ bool operator<(const X509_DN& dn1, const X509_DN& dn2)
...
    {
-   auto dn_info = get_attributes();
-
    der.start_cons(SEQUENCE);
 
    if(!m_dn_bits.empty())
+      {
+      /*
+      If we decoded this from somewhere, encode it back exactly as
+      we recieved it
```
phwork: `received` ;)

src/lib/x509/x509_dn.cpp
```diff
@@ -201,58 +210,32 @@ bool operator<(const X509_DN& dn1, const X509_DN& dn2)
...
    else
       {
-      do_ava(der, dn_info, PRINTABLE_STRING, "X520.Country");
-      do_ava(der, dn_info, DIRECTORY_STRING, "X520.State");
-      do_ava(der, dn_info, DIRECTORY_STRING, "X520.Locality");
-      do_ava(der, dn_info, DIRECTORY_STRING, "X520.Organization");
-      do_ava(der, dn_info, DIRECTORY_STRING, "X520.OrganizationalUnit");
-      do_ava(der, dn_info, DIRECTORY_STRING, "X520.CommonName");
-      do_ava(der, dn_info, PRINTABLE_STRING, "X520.SerialNumber");
+      for(auto dn : m_dn_info)
```
phwork: could use `const auto&`

randombit:
@phwork thank you kindly for the review

Paging also @neusdan + @securitykernel - this PR changes the ordering that DN entries are encoded in. Previously the order was specified by the sequence of calls to `do_ava`. Now we encode by OID ordering. Thus the changes to the DN hash test. I don't think there are any troublesome implications to this but wanted to run it by you.

neusdan:
LGTM!

Issue #210, opened 2015-07-19
make mach_abi_linking configurable


cdesjardins:
Add option to configure script to allow the user to use static run time with MSVC, in other words allow the user to select the `[MT|MTd]` or the `[MD|MDd]` flags at configure.py time.

webmaster128:
Could you help me with naming? Is CRT library (C Run-Time library?) the name you'd use for a `./configure` argument? Why has the C++ standard library a C naming?

Is it `--msvc-crt` or `--windows-crt`? What about MinGW and Cygwin?

Which options do exist? Is this the full list? https://support.microsoft.com/en-us/kb/140584 Which option values do you preferr? `/MLd`, `MLd` or `LIBCD`

cdesjardins:
In visual studio they call it the "Runtime library", which I think is appropriate for all platforms and languages. As far as I am concerned the ML flavors are totally pointless, I have never used them. So basically I would suggest just a switch to turn on "Static runtime library" or "Shared runtime library", then the logic is something like this:

````
if (staticRuntime == True):
    flag = "/MT"
else:
    flag = "/MD"

if (debug == True):
    flag += "d"
````

Regarding MinGW, and Cygwin, I never use them and have no input regarding them.

webmaster128:
I think MingGW and Cygwin ship their own STLs with the compiler, so it is `--msvc-runtime=`. In order to avoid too much magic that needs to be fixed later, I just use "/ML" to "/MDd" as choices. Adding the slash privides the MSCV look and feel.

randombit:
Sorry about the long delay getting this resolved, fixed now by #1499

Issue #294, opened 2015-10-12
Question: How to switch between CRT linkage type for MSVC


diorahman:
Hi,

How to configure the CRT linkage type in MSVC when building Botan from source?

Thanks!

cdesjardins:
Currently there is no built in way to do this, I just use find/replace on the Makefile, sed works fine. It is an open issue:
https://github.com/randombit/botan/issues/210

There is also a branch where they are starting to use cmake as the build system, which should support this, but it is not a very mature build system yet.

randombit:
Closing as dup of #210

PR #1499, opened 2018-03-20
Set the MSVC runtime library appropriately depending on flags

randombit:
Fixes GH #210

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -321,6 +321,9 @@ def process_command_line(args): # pylint: disable=too-many-locals
     target_group.add_option('--ar-command', dest='ar_command', metavar='AR', default=None,
                             help='set path to static archive creator')
 
+    target_group.add_option('--msvc-runtime', metavar='RT', default=None,
+                            help='specify MSVC runtime (MT, MD, MTd, MDd)')
+
     target_group.add_option('--with-endian', metavar='ORDER', default=None,
                             help='override byte order guess')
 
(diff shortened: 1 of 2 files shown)


randombit:
I don't use Windows/MSVC but it seems like this resolves #210 by picking /MT(d) if `--disable-shared` is used and /MD(d) if building the DLL. Any comments @webmaster128 @neusdan @cdesjardins

webmaster128:
I don't think static/dynamic STL linking should be tied to how Botan is built. I think those are orthogonal.

 . | Botan dynamic | Botan static
------------ | -------------| -------------
STL dynamic | 1 | 2
STL static | 3 | 4

1 and 2 are supported at the moment, are valid and work. At Kullo we use 2 because we want Botan to be static but need dynamic STL because we also link Qt dynamically and don't want dynamic and static STL in the same application.

I don't see how 3. helps anyone but I am not an expert here. 4 (all static) is nice to reduce dependencies and I guess this is what users in #210 and #294 want.

This PR changes choices in from 1+2 to 1+4. But I think we should support all four.

randombit:
OK makes sense I will rework this PR. Thanks

webmaster128:
Looks good to me. However, I don't see a reason so change default behaviour for static build here. Static Botan with dynamic STL is probably a good choice for many applications and changing the defaults from case 2 to 4 breaks builds in ways that may be hard to debug for no real benefit.

PR #1505, opened 2018-03-22
Remove use of rpath=ORIGIN and similar tricks on Darwin

randombit:
These complicated the build steps, and distros don't like (or explicitly prohibit) use of rpath=ORIGIN for security reasons.

The only reason for doing this was so that the user would not be confused, by trying to run botan-test before install without their LD_LIBRARY_PATH pointing in the right spot. But it is easier to just fix that problem by adding a script that sets up the environment and then runs the binary.

This also makes out of tree builds easier, because it is possible for the script to account for where the test data dir is and pass the relevant flag down to botan-test.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -180,6 +180,7 @@ class SourcePaths(object):
         self.scripts_dir = os.path.join(self.src_dir, 'scripts')
 
         # subdirs of src/
+        self.test_data_dir = os.path.join(self.src_dir, 'tests/data')
         self.sphinx_config_dir = os.path.join(self.configs_dir, 'sphinx')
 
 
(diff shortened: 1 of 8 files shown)


randombit:
OK this doesn't quite work as is since CI needs qemu, etc...

noloader:
@randombit,

Another way to handle this is using `install_name_tool` on OS X. For example, here is what [Crypto++ does in its makefile](https://github.com/weidai11/cryptopp/blob/master/GNUmakefile#L895) after an install (i.e., Botan's post-link step):

```
ifneq ($(wildcard libcryptopp.dylib),)
	@-$(MKDIR) $(DESTDIR)$(LIBDIR)
	$(INSTALL_PROGRAM) libcryptopp.dylib $(DESTDIR)$(LIBDIR)
	-install_name_tool -id $(DESTDIR)$(LIBDIR)/libcryptopp.dylib $(DESTDIR)$(LIBDIR)/libcryptopp.dylib
endif
```

You should be able to set the `install_name` to PWD after building the library so the self tests can run. Then, after installation, set the `install_name` again to the final location.

devnexen:
Once rebased looks good to me, indeed this is not necessary anymore.

PR #1506, opened 2018-03-22
Update modules dependency graph

webmaster128:
This improves the visual representation of the module dependency graph.

Old:

graph

New:

graph-new


randombit:
Nice, new format is much easier to read. Thanks

PR #1508, opened 2018-03-26
Support passing an OAEP label in EME name

securitykernel:
TPM 1.2 expects passing the owner and SRK secret encrypted with the public endorsement key.
For asymmetric encryption, the TPM 1.2 uses OAEP with the label "TCPA".


randombit:
Looks OK (though these code paths are getting a little messy) but *tests please* :)

randombit:
There is one OAEP with label test here https://github.com/pyca/cryptography/blob/master/vectors/cryptography_vectors/asymmetric/RSA/oaep-label.txt

randombit:
Oh you're right, that is unfortunate. I'm not sure where else to source a test vector beyond writing an OpenSSL test prog.

securitykernel:
Ok, will try OpenSSL.

securitykernel:
Add one test vector in c94c227 generated using OpenSSL and rebased.

randombit:
Nice thanks. I will merge after 2.5 release goes out

PR #1513, opened 2018-03-28
Update TLS manual to reflect custom curve changes

neverhub:
I recently updated the manual to include the changes performed in https://github.com/randombit/botan/pull/1448. In addition, i added a server and client example with custom curves based on the already existing client/server examples.

While doing this, i noticed a few minor issues in the client/server examples. I fixed them in this PR as well.


randombit:
Nice thank you

PR #1525, opened 2018-04-06
Make tests pass by default in UBSAN mode

mgierlings:
- Adds macros to check if botan was compiled with a certain sanitizers.
- Automatically excludes the tests that are intended to provoke undefined
  behaviour from the test bench, when botan is compiled with UBSAN.
- Changes option `--avoid-undefined` to `--no-avoid-undefined` so the
  failing tests can be explicitly activated when needed.

Disclaimer: I'm not 100% sure if the `sanitizer_types` list I added to 
`configure.py` is in the right place, however I'll be happy to make
any adjustments here.


mgierlings:
Yes I like the better readability of your suggestion, thanks for the hint. I'll change the PR accordingly.

randombit:
LGTM thanks!

PR #1526, opened 2018-04-06
Fixes memory leak in benchmark

mgierlings:
The test bench allocates raw pointers and fails to delete these.

### DIFF
--- a/src/cli/speed.cpp
+++ b/src/cli/speed.cpp
@@ -783,12 +783,18 @@ class Speed final : public Command
                }
 #endif
 #if defined(BOTAN_HAS_CIPHER_MODES)
-            else if(auto enc = Botan::get_cipher_mode(algo, Botan::ENCRYPTION))
+            else if(auto enc =
+                       std::unique_ptr<std::remove_reference<
+                           decltype(*Botan::get_cipher_mode(
+                                     algo, Botan::ENCRYPTION))>::type>
+                       { Botan::get_cipher_mode(algo, Botan::ENCRYPTION) })
                {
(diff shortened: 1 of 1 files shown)


randombit:
Nice find, thanks. But, this is not the right fix. Explicit `delete` is always asking for trouble. Instead the raw pointers should be captured by a `unique_ptr`.

randombit:
A while back I went on a tear to remove (almost) all explicit `delete` calls 096d99fe1528195c523ee6ed7b3defb7519140e2 f9051c02b7b1ce4c97a16a218b0cea0439722f4e ba89dd9a71016b318847fe471dc086504e18cf42 ee45da7befa4b5dcc28105df13b6248029dbf036 bac9f93cd68994710f2a2b916a0f121f715458b1 de581bbf6dfc1fadc0755925300d047f392bf986 d7330e586aee1912838952c6583d9449060d776a

mgierlings:
Okay 9d480840c09375116ae908fa182ad4adebc4d2af captures the raw pointers inside `std::unique_ptr`. Readability wise this is not really pretty though and therefore I'd have rather gone with deleting the raw pointers. I think the problem here is not to delete the raw pointers but the fact that we are returning raw pointers in the first place. So the best solution would be to have `Botan::get_cipher_mode` return smart pointers.

randombit:
You are right about that. We can't change these existing APIs since that would break callers, but we can introduce `create` and `create_or_throw` which matches the interface used for other types. Done in #1527, which fixes the leak in "the general case" in that ``auto cipher = Cipher_Mode::create(...)` will automatically free. So closing this PR.

PR #1530, opened 2018-04-10
Use mkstemp(3) to create file for test.

bluhm:
Creating a temporary file in the current directory may fail due to
permissions.  Use POSIX mkstemp(3) to create a temporary file in a
secure way for the filter test.

Although mkstemp(3) is the right thing for UNIX to create temporary
files, I don't know what is suitable for Windows.  The current directory
is bad for the stand alone botan-test tool.  Also when using the OpenBSD
ports framework with privilege separation the test user is not allowed
to write into the build user's tree.

### DIFF
--- a/src/tests/test_filters.cpp
+++ b/src/tests/test_filters.cpp
@@ -32,6 +32,7 @@
 
 #if defined(BOTAN_TARGET_OS_HAS_FILESYSTEM)
    #include <fstream>
+   #include <stdlib.h>
 #endif
 
 namespace Botan_Tests {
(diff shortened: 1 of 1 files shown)


bluhm:
I have added a #include <stdlib.h>.  Let's see whether this fixes the CI build.

randombit:
Using `mkstemp` for POSIX is OK. I would want to generalize this a bit so other future tests needing temp files do not need to replicate the logic; add `std::string Test::temp_filename(const std::string& basename)` which for POSIX systems (`#if defined(BOTAN_TARGET_OS_HAS_POSIX1)`) does `mkstemp` and for other systems (for now anyway) drops into cwd as before. This would also fix the Windows build.

If you want to amend PR as above awesome, otherwise I will do this in next few days.

PR #1531, opened 2018-04-10
Improve XMSS coverage

mgierlings:
Codecov does not reach all parts of the `XMSS_PrivateKey` code because too few
cores are detected during the CI run. To cover the missed codepaths always
return a large enough core count if botan is compiled with coverage.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1857,6 +1857,8 @@ def create_template_vars(source_paths, build_paths, options, modules, cc, arch, 
 
         'build_fuzzers': options.build_fuzzers,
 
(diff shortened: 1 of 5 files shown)


randombit:
This doesn't interact well if the end user wants coverage #s for their own usage, but that is an edge case for sure.

It may be better to add a "test mode" that is enabled for CI builds, instead of going off of coverage information. This could additionally do things like enabling expensive assert checks that we don't want in a production binary.

I'll merge this now but I think ^ is better.

PR #1532, opened 2018-04-11
Add missing errno.h include to entropy/dev_random

fwsGonzo:
This was necessary for me when building for IncludeOS with musl (https://www.musl-libc.org/) as the C library


randombit:
Thanks!

PR #1534, opened 2018-04-11
Optimize EC point doubling for a == 0 and a == -3

randombit:
For a == -3 (NIST P- curves) it does not really help much, more work needed there I suspect.

Improves secp256k1 ECDSA verification and ECDH agreement by 10%

Improves secp{256,384,521}r1 ECDH by ~5%. ECDSA verification slightly faster but just a nudge (~1-2%).

### DIFF
--- a/doc/todo.rst
+++ b/doc/todo.rst
@@ -40,7 +40,6 @@ Public Key Crypto, Math
 * X448 and Ed448
 * FHMQV
 * Use GLV decomposition to speed up secp256k1 operations
-* Optimize ECC point doubling for a=-3 and a=0 curves
 * wNAF ECC point multiply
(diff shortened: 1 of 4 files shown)


randombit:
This test failure on AppVeyor is concerning

> ECC randomized secp224r1 ran 1 tests in 55.10 sec 1 FAILED
> Failure 1: ECC randomized secp224r1 secp224r1 with error bad allocation

randombit:
With MSVC it seems an infinite loop hits in `ecc_randomized`. Disabled a=-3 case to try to isolate issue.

randombit:
MSVC 2017 (only) doesn't like something about the a=-3 code, don't really see the problem.

PR #1535, opened 2018-04-12
Do not create shared library symlinks on OpenBSD.

bluhm:
Symlinks to shared libraries confuse the OpenBSD dynamic linker.
We need one file with two numbers.  The problem became apparent
when the abi_rev and the OpenBSD ports shared libs numbers diverged.
Add a new conditional variable symlink_shared_lib to suppress the
symlink in the makefile.


randombit:
Thanks

PR #1542, opened 2018-04-17
RSA keygen: optimize and avoid side channel

randombit:
Side channel is same idea as OpenSSL's CVE-2018-0737, gcd algorithm is not constant time and so it is possible to construct information about the prime via the gcd behavior. No idea if it is exploitable with our particular implementation.

Creates a specialized function for RSA keygen, which might change algorithms in the future (eg to whatever is in FIPS 186-4).

Adds calculated probabilities for 1/2**128 error in Miller-Rabin, which lets us use reduced M-R iterations for the default keygen params.

Also, a surprisingly important optimization, move the check for p.bits() being too large until just before the primality check. Otherwise we are calling `bits` in a loop many times when it is very unlikely to have overflowed.

Now even 4K bit RSA keygen takes under a second on my machine.

### DIFF
--- a/src/lib/math/numbertheory/make_prm.cpp
+++ b/src/lib/math/numbertheory/make_prm.cpp
@@ -12,6 +12,61 @@
 
 namespace Botan {
 
+namespace {
+
+class Prime_Sieve
+   {
+   public:
+      Prime_Sieve(const BigInt& init_value) : m_sieve(PRIME_TABLE_SIZE)
+         {
+         for(size_t i = 0; i != m_sieve.size(); ++i)
+            m_sieve[i] = static_cast<uint16_t>(init_value % PRIMES[i]);
+         }
+
+      void step(word increment)
+         {
+         for(size_t i = 0; i != m_sieve.size(); ++i)
+            {
+            m_sieve[i] = (m_sieve[i] + increment) % PRIMES[i];
+            }
+         }
+
+      bool passes(bool check_2p1 = false) const
+         {
+         for(size_t i = 0; i != m_sieve.size(); ++i)
+            {
+            /*
+            In this case, p is a multiple of PRIMES[i]
+            */
+            if(m_sieve[i] == 0)
+               return false;
+
+            if(check_2p1)
+               {
+               /*
+               In this case, 2*p+1 will be a multiple of PRIMES[i]
+
+               So if potentially generating a safe prime, we want to
+               avoid this value because 2*p+1 will certainly not be prime.
+
+               See "Safe Prime Generation with a Combined Sieve" M. Wiener
+               https://eprint.iacr.org/2003/186.pdf
+               */
+               if(m_sieve[i] == (PRIMES[i] - 1) / 2)
+                  return false;
+               }
+            }
+
+         return true;
+         }
+
+   private:
+      std::vector<uint16_t> m_sieve;
+   };
+
+}
+
+
 /*
 * Generate a random prime
 */
@@ -64,7 +119,6 @@ BigInt random_prime(RandomNumberGenerator& rng,
          }
       }
 
-   secure_vector<uint16_t> sieve(PRIME_TABLE_SIZE);
    const size_t MAX_ATTEMPTS = 32*1024;
 
    while(true)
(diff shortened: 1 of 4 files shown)


AnnieYousar:
Hi randombit,
I'm fine with your changes in `numtry.cpp` in https://github.com/randombit/botan/commit/04d15b435183299d7cacf180316f19b636300ecb
Concerning the error probs, I used in the Magma script the monster (formula 2) from the NIST Appendix F.1 (p. 117). Additionally I checked the script
```
f2:= function(k,t)
 if ((t eq 2) and (k ge 88)) or ((t ge 3) and (9*t le k) and (k ge 21)) then
  return RealField(4)!-Log(2,k^1.5*2^t*4^(2-Sqrt(t*k))/Sqrt(t));
 else return "";
 end if;
end function;

for i:=256 to 1536 by 256 do
 s:="";
 for t:=1 to 28 do
  rs:=Sprintf("%o ",f2(i,t)); s*:=rs;
  s*:=": ";
 end for;
 printf "%4o : %o\n",i,s;
end for;
```
from HAC (A. Menezes, P. van Oorschot und O. Vanstone. Handbook of Applied Cryptography. CRC Press, 1996). Some minor differences to your results:  for (n_bits >= 256) 28 test give only 2^-127.7 and not 2^-128.

AnnieYousar:
The idea of the Miller-Rabin test is as follows (https://en.wikipedia.org/wiki/Miller%E2%80%93Rabin_primality_test):
> Let n be prime with n > 3, then we can write n−1=2^s·d, where s and d are positive integers and d is odd. Then for each a≠0 in (Z_n), either a^d ≡ 1 (n) or (a^d)^(2^r) ≡ −1 (n) for some 0 ≤ r ≤ s−1. 

This is used as a oracle:

> Let n be a candidate with  n−1=2^s·d, then n is not a prime if there is an a≠0 with a^d ≢ 1 (n) and for all r with 0 ≤ r ≤ s − 1 we have  (a^d)^(2^r) ≢ −1 (n). In this case a is called a witness (for not being a prime).

Note that the check, whether (a^d)^(2^r) is 1 modulo n, is not necessary.

The Fermat test  is based on

> Let n be prime with n > 3, then for each a≠0 in (Z_n) a^(n-1) ≡ 1 (n).

And the test is:

> Let n be a candidate. Then n is not a prime if there is an a≠0 with a^(n-1) ≢ 1 (n).

The advantage of Miller-Rabin is that all composite number have witnesses. This is not true for the Fermat test. A Carmichael number is composite and have no Fermat witnesses at all.
 
So in `mr_witness()` we got the true Miller-Rabin test and btw the code snippet should be
```
    if(y == 1 || y == n_minus_1)
       return false;
 
  for(size_t i = 1; i != s; ++i)
       {
       y = reducer_n.square(y);
 
       if(y == n_minus_1) 
          return false; // not a witness, so give up
       }
  return true; // candidate fails Miller-Rabin test
```

randombit:
> Some minor differences to your results: for (n_bits >= 256) 28 tests give only 2^-127.7 and not 2^-128.

My script also gives 127.7 for `p(256, 28)` but here I rounded it up since even `p(257,28)` is under 2^-128. But since generation of such small primes is not operationally important (we don't even allow generating RSA keys under 1024 bit) I will instead bump this to 29 trials to assure we are always under the desired bound.

```
  for(size_t i = 1; i != s; ++i)
       {
       y = reducer_n.square(y);
 
       if(y == n_minus_1) 
          return false; // not a witness, so give up
       }
```

However the existing early exit for `y == 1` seems fine since in that case we know that `y` will be 1 through the end of the loop, and thus never n - 1, and so the remaining squarings are wasted. There also doesn't seem to be any useful timing channel since this only occurs when the integer is not prime.

AnnieYousar:
I checked that many times, the test `y==1` is useless in many cases. I set a break point in this test, but I never reached it. Yes, there may be very rare cases, where you waste time. But the remaining squarings cost you almost nothing, because they are very, very fast: y is already 1. OK, I didn't check the costs of the extra `y==1` test, so feel free to keep it as it is.

AnnieYousar:
check the spelling in line 504 ;-)
```
-        * Test for primaility using Miller-Rabin
+        * Test for primality using Miller-Rabin
```

randombit:
Fixed thanks

Issue #1479, opened 2018-03-08
Faster modular inversion


randombit:
At this point modular inversion is one of the major bottlenecks in ECDSA signature generation, at about 30% of the total runtime. Two inversions are required, one for the nonce and the other to convert the point to affine.

Niels Moller sent me an email where he contended that the fastest approach for const time inversion at ECC sizes was using Fermat's little theorem. However in Botan with P-521, using the const time inversion algorithm (that Niels invented) is over twice as fast. So maybe the issue is (also) that our modular exponentiation algorithm is too slow.

I ran some quick checks, OpenSSL seems to take ~210k cycles for const time modular inversion, vs 360k cycles in ct_inverse_mod_odd_modulus. Simply matching that would improve ECDSA signature perf by 15%

randombit:
WRT using Fermat's little theorem it would be beneficial to use the known addition chains for computing `x^(p-2) % p` https://briansmith.org/ecc-inversion-addition-chains-01#p256_scalar_inversion has chains for inversions mod order and mod p for P-256 and P-384.

randombit:
https://eprint.iacr.org/2014/852.pdf has the addition chain for P-521 inversion.

randombit:
With #1546 and #1547 we have faster field inversions for P-256, P-384, and P-521 all of which improved ECDSA signature performance by ~~ 10-15%. ECDSA verification and ECDH also improved but not as much (% wise) because getting the affine coordinate is less of the total runtime there.

We may also want specialized inversions modulo the curve order.  But this doesn't really seem worthwhile because the only algorithm that benefits is ECDSA. It would be better at this point to work on improving the performance of the generic const time modular inversion, which would have a lot of benefits across the codebase.

henrydcase:
Have you looked at https://eprint.iacr.org/2019/266.pdf already?

randombit:
@henrydcase I have seen that paper but currently don't understand at all how the algo works. It seems much faster, eg they report for inversion modulo the 511-bit M-511 prime 30K Skylake cycles, while Botan's const time algo takes 200K+ Skylake cycles for a randomly chosen 512-bit prime. And the paper claims "Our advantage is also larger in applications that use “random” primes rather than special primes" [vs Fermat]. So overall certainly promising. Even if we assume a 5x slowdown going from DJB hand coded asm to C++ that's still a decent speedup.

Also our current const time algorithm only works for odd modulus which means the generation of `d` during RSA keygen is not protected.

[...]

PR #1547, opened 2018-04-20
Add Fermat based inversion of P-384 field elements

randombit:
Cuts about 100K cycles from the inversion, improving ECDSA sign by 10% and ECDH by ~2%

Addition chain from https://briansmith.org/ecc-inversion-addition-chains-01

GH #1479


briansmith:
IMO it would help understand the code to include the link to the addition chain documentation in the code instead of in the commit message.

PR #1554, opened 2018-04-24
WIP: Add support for Linux AF_ALG interface

randombit:
Right now just hashes as a prototype. Motivation is mostly for supporting crypto hardware that is accessible to the kernel but not userspace (old style ARM co-processors and the like). However I have no such devices on hand. It looks like the Beagle Bone Black is a good test candidate (has AES, SHA-1, MD5).

### DIFF
--- a/src/build-data/os/linux.txt
+++ b/src/build-data/os/linux.txt
@@ -9,6 +9,7 @@ dev_random
 proc_fs
 clock_gettime
 getauxval
+af_alg
 
 sockets
 threads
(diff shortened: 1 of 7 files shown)


randombit:
Looking at adding the cipher/AEAD support, things get hairy. Nothing too terrible but it ends up being a lot of boilerplate code and dealing with strange corner cases of the kernel API.

Going to switch to using https://github.com/smuellerDD/libkcapi is a dual BSD/GPL helper library for interfacing with AF_ALG. It is included in latest Debian, Ubuntu, and Fedora.

randombit:
Closing, not sure we want this

PR #1555, opened 2018-04-25
Implementing Base32

Wambou:
Here is the implementation of the Base32 in such a way that Base64 can be changed to reuse most of the code.

The main idea is to use a class that includes data used in the algorithm.
It works for Base32 and Base64 encoding, but might not work for Base16 (existing Hex) with minimal change.

Because I cannot use constexpr, I had to use a std::vector in the algorithm instead of an std::array, this can be an improvement.

I compiled with:
- gcc 6.3.0-18
- clang 3.8.1-24
- msvc 2013
- msvc 2017

I also used astyle automatic formating.
I can squash commits 75ad320 to cda9db2, I preferred iterative steps to implement everything.


randombit:
Nice, thank you. I will review this soon.

One immediate comment: the newly added files should have your copyright info. Since it looks like you based the code somewhat on the existing base64 code it is fine to have (C) me in there also but you also wrote code in those files, and as we do not require a CLA the copyright on your code remains yours.

randombit:
@Wambou Sorry for the long delay on reviewing. Looks good overall. A few small changes

- The copyright issue I noticed in previous comment
- For now at least, `codec_base.h` should be an internal header so we don't need to commit to keeping that specific API moving forward. (And anyway it includes an internal header, so it could not practically be used by a end-user in any event.)
- In `base32.h` the `BOTAN_PUBLIC_API` annotations should be 2,7 since 2.7 is the first release they will appear in.
- There doesn't seem to be any need for the empty `codec_base.cpp`

I still need to review the actual encoding logic in `base_encode` and `base_decode` more carefully but in the meantime can you make the above changes as well as rebasing against master to handle the merge conflict in the news file?

Wambou:
Thank you for the review.
I will make the changes ASAP.

Wambou:
I updated the code to master and applied the modifications in tests to match your refactoring (#1592).

randombit:
Merged now thank you!

Issue #1553, opened 2018-04-24
Add OIDS for OpenPGP Curve25519


lambdafu:
RFC4880bis specifies the following private enterprise OIDS:

    1.3.6.1.4.1.11591.15.1   Ed25519
    1.3.6.1.4.1.3029.1.5.1   Curve25519

It's a bit sad that everybody rushes to specify their own identifiers, and then we get multiple identifiers for essentially the same thing (although in the case of GnuPG, it uses Curve25519 with ECDSA+ECDH instead of X25519, but that's a different story altogether and it's not resolved yet whether this is a bug that will be fixed or if it justifies using a different OID).

These identifiers are already in the wild (about 1750 and 750 keys/signatures on the public key servers).

Botan recognizes the OIDS from "Algorithm Identifiers for Ed25519, Ed448, X25519 and X448 for use in the Internet X.509 Public Key Infrastructure" https://datatracker.ietf.org/doc/draft-ietf-curdle-pkix

     id-X25519    OBJECT IDENTIFIER ::= { 1 3 101 110 }
     id-X448      OBJECT IDENTIFIER ::= { 1 3 101 111 }
     id-Ed25519   OBJECT IDENTIFIER ::= { 1 3 101 112 }
     id-Ed448     OBJECT IDENTIFIER ::= { 1 3 101 113 }

Maybe we can find some names for the OpenPGP variants and add them to the OIDS recognized by Botan. I'd make a patch, but some obvious names ("Ed25519", "Curve25519") are already taken, and I really don't want to add to the confusion.  Some guidance for an appropriate naming scheme would be nice.  Maybe "Ed25519/OpenPGP", "Curve25519/OpenPGP" ?  I think it's important to have unique names, aliasing would only ask for even more trouble.

randombit:
I did some searches and it seems these OIDs are not really used outside of OpenPGP (effectively all hits were in `gpg` or `libgcrypt`) so an OpenPGP specific tag seems right.

The convention (within Botan) for protocol-specific OIDs is to prefix it with the protocol name: `OpenPGP.Ed25519`. Slash (especially after a pubkey algo name) is normally used to provide extra parameters so for instance `Ed25519/X` would be used to to specify Ed25519 with hash X.

randombit:
Fixed thanks for the PR!

PR #1560, opened 2018-05-02
Add OpenPGP-specific curve OIDs

lambdafu:
This takes care of #1553.


randombit:
Close but not quite: edit `src/build-data/oids.txt` then regenerate the OID map

```
./src/scripts/oids.py oids > src/lib/asn1/oid_maps.cpp
```

lambdafu:
Lol, should have read the comments. Thanks!

PR #1563, opened 2018-05-07
Fix signature of SystemFunction036 / RtlGenRandom

### LINKED ISSUE #1562: Run-Time Check Failure in system_rng on Windows (32 bit)
When compiling and running

```
$ python configure.py --debug-mode --cxxflags="/RTC1" --cpu=x86_32
$ jom -j 6
$ botan-test.exe system_rng
```

I get the error

> Run-Time Check Failure #0: The value of ESP was not properly saved across a function call.  This is usually a result of calling a function declared with one calling convention with a function pointer declared with a different calling convention.

bildschirmfoto von 2018-05-07 18-22-19

The same happened when I ran my application in debugger, where I found that the line https://github.com/randombit/botan/blob/2.6.0/src/lib/rng/system_rng/system_rng.cpp#L44 triggers the run-time check failure.

The problem does not occur without `/RTC1` or in a 64 bit build.

[Looking at how Chromium uses RtlGenRandom](https://github.com/chromium/chromium/blob/9899ff1dc3/base/rand_util_win.cc), it seems possible and easier to use RtlGenRandom directly.

webmaster128:
Closes #1562

With this patch applied, I cannot reproduce #1562 anymore.

### DIFF
--- a/src/lib/rng/system_rng/system_rng.cpp
+++ b/src/lib/rng/system_rng/system_rng.cpp
@@ -50,7 +50,7 @@ class System_RNG_Impl final : public RandomNumberGenerator
       void clear() override { /* not possible */ }
       std::string name() const override { return "RtlGenRandom"; }
    private:
-      typedef BOOL (*RtlGenRandom_f)(PVOID, ULONG);
+      typedef BOOL (NTAPI *RtlGenRandom_f)(PVOID, ULONG);
 
       Dynamically_Loaded_Library m_advapi;
       RtlGenRandom_f m_rtlgenrandom;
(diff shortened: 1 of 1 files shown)


webmaster128:
I can reproduce locally that this does not work with amalgamation but works without. Any ideas?

randombit:
Ugh, yes `ASN1_Tag` dumps everything into the namespace. Switching to `enum class` for everything should have happened before 2.0 stabilization. I have added it to the 'for 3.x' incompatible changes list.

randombit:
Using `BYTE` as you did here seems like a decent workaround.

webmaster128:
okay, fine. Thanks for your guidance! Coverage is probably down because this is based on the 2.6.0 tag, not on master

randombit:
Thanks

PR #1578, opened 2018-05-24
Correct XMSS key leaf index

quantumcthulhu:


mgierlings:
A tree of height 10 actually does have 512 leaves which is about half of its 1023 nodes. Lets take a look at a smaller toy example a tree of height 3:

```
                  1
	    2           3
        4     5     6      7
```

As you can see the tree has (2^3)-1 = 7 nodes. Out of all these the nodes four are leaves, specifically the nodes numbered 4, 5, 6, 7. So we have 4 = 2^2 = 2^(3 - 1) leaf nodes. Generalized you would calculate the leaf count: 2^(h-1), wheras your formula (2^h) - 1 calculates the number of *nodes*.

Receiving an error for leaf index 512 in a tree of hight 10 is desired, because it has 512 leaves, which are numbered with the leaf indices 0 ... 511.

quantumcthulhu:
Thanks, you helped me to understand my mistake

PR #1579, opened 2018-05-24
Keep cxx_abi_flags out of CXX, which may get overridden

lanurmi:
Overriding CXX with the make command results in cxx_abi_flags being ignored, which in turn may lead to a linking error.

For example: CXX is set to value `ccache g++`, which leads to linker being called without `-fstack-protector`, which in turn can cause unresolved symbols.

### DIFF
--- a/src/build-data/makefile.in
+++ b/src/build-data/makefile.in
@@ -1,12 +1,13 @@
 # Paths to relevant programs
 
-CXX            = %{cxx} %{cxx_abi_flags}
+CXX            = %{cxx}
 LINKER         = %{linker}
 AR             = %{ar_command}
 PYTHON_EXE     = %{python_exe}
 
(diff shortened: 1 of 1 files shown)


randombit:
Makes sense, thanks. One of the CI builds is failing (https://travis-ci.org/randombit/botan/jobs/383364656) I think because we are using `--coverage` flag but it is not being included when the fuzzer binaries are linked. I think all that is need is to add the new ABI flags variable also to that link command (last one in the makefile template).

lanurmi:
Done. Not sure why Codecov says the coverage decreases, though.

randombit:
The coverage decrease is because Travis alternates between building on systems with and without AVX2 and so sometimes the AVX2 code is not hit which looks like a coverage decrease.

Issue #1580, opened 2018-05-25
XMSS signatures max count


quantumcthulhu:
I want to make 2^10 signatures with XMSS tree height=10. But I can sign only 512 messages. Why?

```
#include <iostream>
#include <botan/xmss.h>
#include <botan/auto_rng.h>

bool sign_and_verify_message(Botan::XMSS_PrivateKey & key, Botan::secure_vector<uint8_t> const & msg, Botan::AutoSeeded_RNG & rng)
{
	// create signature operation using the private key.
	std::unique_ptr<Botan::PK_Ops::Signature> sig_op = key.create_signature_op(rng, "", "");

	sig_op->update(msg.data(), msg.size());
	Botan::secure_vector<uint8_t> sig = sig_op->sign(rng);

	// create verification operation using the public key
	std::unique_ptr<Botan::PK_Ops::Verification> ver_op = key.create_verification_op("", "");

	// verify the signature for the previously generated message.
	ver_op->update(msg.data(), msg.size());
	return ver_op->is_valid_signature(sig.data(), sig.size());
}

int main()
{
	int i = 0;
	try
	{
		Botan::AutoSeeded_RNG rng;

		Botan::XMSS_PrivateKey key(Botan::XMSS_Parameters::XMSS_SHA2_256_W16_H10, rng);

		std::string str("test message");
		Botan::secure_vector<uint8_t> msg(str.begin(), str.end());

		for (; i <= 1024; ++i)
		{
			sign_and_verify_message(key, msg, rng);
			std::cout << i+1 << " messages are signed" << std::endl;
		}
	}
	catch (std::exception const & e)
	{
		std::cout << "Exception: " << e.what() << std::endl;
	}

	std::cout << "totally " << i << " messages are signed" << std::endl;

    return 0;
}
```

returns:

> Exception: Integrity failure: XMSS private key, one time signatures exhausted.
> totally 512 messages are signed

mgierlings:
This is a bug in the sanity check for the leaf index bounds you already referred to with #1578. The reason why you're running out of leaves early, is that the leaf index bounds are currently calculated based on the tree height, whereas they should be calculated based on the number of layers in the tree, which is equal to `tree_height + 1`. The relevant part is in [section 4.1.1](https://tools.ietf.org/html/draft-irtf-cfrg-xmss-hash-based-signatures-12#section-4.1.1) of the XMSS standard. The problem should be fixed with #1582 . Thanks a lot for pointing this out, this was a nice find!

PR #1582, opened 2018-05-25
Fixes XMSS leaf index bounds sanity check

mgierlings:
Prior to this patch the sanity check for XMSS leaf indices was wrongly based on
the tree height. As a result only half of the one-time keys could be used.

Instead base leaf index sanity check on the number of levels in a tree which equals
tree height + 1. 

(see also: https://tools.ietf.org/html/draft-irtf-cfrg-xmss-hash-based-signatures-12#section-4.1.1)

### DIFF
--- a/src/lib/pubkey/xmss/xmss_privatekey.cpp
+++ b/src/lib/pubkey/xmss/xmss_privatekey.cpp
@@ -50,8 +50,7 @@ XMSS_PrivateKey::XMSS_PrivateKey(const secure_vector<uint8_t>& raw_key)
       unused_leaf = ((unused_leaf << 8) | *i);
       }
 
-   if(unused_leaf >= (1ull << (XMSS_PublicKey::m_xmss_params.tree_height() -
-                      1)))
+   if(unused_leaf >= (1ull << XMSS_PublicKey::m_xmss_params.tree_height()))
       {
       throw Integrity_Failure("XMSS private key leaf index out of "
                               "bounds.");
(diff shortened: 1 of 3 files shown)


randombit:
Ref #1580 #1578

randombit:
Thanks @mgierlings and @quantumcthulhu

It would be nice if there was a test for this, but XMSS signatures are so slow that doing 1024 would take 30-60 minutes even on a fast machine...

randombit:
Would it be possible to "edit" the key state to fast forward it to 1022 signatures then verify the key is exhausted after 1024?

mgierlings:
@randombit Actually we can do even better than that. You can fast forward the key using [XMSS_PrivateKey.set_unused_leaf_index(size_t)](https://github.com/randombit/botan/blob/f256e603ab93e2587fe0e38bcf06c437fcd41abf/src/lib/pubkey/xmss/xmss_privatekey.h#L114). This function is a one way street though. Once you skipped forward you cant go back, this is to prevent accidental key reuse. The universal solution to test for arbitrary XMSS parameter sets would be sth. like this:
```
Botan::AutoSeeded_RNG rng;
Botan::XMSS_PrivateKey key(Botan::XMSS_Parameters::XMSS_SHA2_256_W16_H10, rng);
size_t max_idx = (1ull << Botan::XMSS_Parameters(Botan::XMSS_Parameters::XMSS_SHA2_256_W16_H10)).tree_height()) - 1;
key.set_unused_leaf_index(max_idx);
...
```
That way you can sign one message before you run out of keys.

PR #1584, opened 2018-05-27
Add BMI2-specific SHA-256

randombit:
Currently just a copy of the baseline compression function, but compiled with BMI2 flags. On Skylake improves performance by about 40%.

Still a ways to matching OpenSSL though. Hashing 16K buffers on Skylake

Before (non-BMI2): 229 Mb/s
BMI2: 361 Mb/s
OpenSSL 1.1.0h: 541 Mb/s

Probably easiest win is to rewrite the code so GCC can figure out how to use AVX2 for the message expansion, then compile this with both BMI2 and AVX2 enabled.


randombit:
Quick tests show SHA-512 sees a 30% speedup from BMI2 flags so probably worth doing something similar for there.

randombit:
Enabled BMI2 across entire build and compared performance with baseline. Besides SHA-256 and SHA-512 only algorithm that saw a large speedup from BMI2 was SHA-3 (~30% faster).

Issue #1586, opened 2018-05-28
Windows UWP / Windows Phone platform


steipete:
Hi! We compile Botan in Windows UWP and use the Windows Phone platform as target, which has now been deprecated. Please reverse this decision. Botan is used in PSPDFKit, including PSPDFKit for Windows: https://pspdfkit.com/pdf-sdk/windows/

The regular Windows platform is for WIN32, which uses API not available in UWP.

We recently noticed this change as we're in the process of updating to Botan 2.6.

If we're doing it wrong, and there's a separate UWP target that we're missing, please let me know. 
If you need more information on this issue we can expand on our case as well.

Thanks! Botan is awesome! We're so glad we could ditch OpenSSL for it!

randombit:
My understanding is that Windows Phone is basically dead upstream @ Microsoft. This presumably is not true, if you are shipping applications for it. Can you explain the difference (if any) between Windows Phone platform and UWP? OK, from https://docs.microsoft.com/en-us/windows/uwp/get-started/universal-application-platform-guide it sounds like UWP is the general cross-device Microsoft API, Windows Phone is just the mobile side, and UWP is still used for if you say wanted to use the same binary for both Windows 10 and Xbox? This about right? (Sorry I know next to nothing about Windows dev...)

If UWP is still in use and supported by Microsoft then we do want to support it.

You should know that starting in 2.5.0 the `System_RNG` switched to using `RtlGenRandom` on Windows, which is not (AIUI) available on UWP. (Which means the system RNG class is absent on UWP builds.) We dropped it because it allows removing any use of CryptoAPI/CryptoNG, RtlGenRandom is supported on all version of Windows desktop we support, and because I did not know anyone was actually using UWP support. We'll probably need to revert at least part of #1494 to bring back the CryptoNG RNG for UWP.

randombit:
@steipete #1587 adds back support for the CryptoNG RNG and reverts the deprecation. It would be great if someone could test this change before merge as we do not have UWP/WinPhone in the CI builds.

steipete:
We'll test it and report back. Huge thanks!! ❤️

randombit:
Build success was confirmed so #1587 has been merged to master now, thanks for raising this.

vivanov-dp:
[...]

PR #1587, opened 2018-05-28
Add back support for Windows Phone RNG, undeprecate UWP

randombit:
See #1586. Reverts part of #1494


danielmartin:
Thanks, @randombit, I can confirm this patch passes our UWP CI. Feel free to merge it!

PR #1588, opened 2018-05-28
Try to fix docs build

randombit:
Seems to be some problem introduced in Sphinx 1.7.5 (released today)


randombit:
Moving back to v1.7.4 seems to fix it. Opened an issue with Sphinx https://github.com/sphinx-doc/sphinx/issues/5022

PR #1598, opened 2018-06-05
Conditionally use concurrency with sphinx-build.

dewyatt:
Sphinx added support for the `-j` option at 1.2. CentOS 7, for example, currently has 1.1.3.

There might be a better way to handle this, this was just my first thought.
I guess another approach might be to try running with `-j` first, then try without if it fails.

### DIFF
--- a/src/scripts/build_docs.py
+++ b/src/scripts/build_docs.py
@@ -120,6 +120,22 @@ def parse_options(args):
 
     return options
 
+def sphinx_supports_concurrency():
+    import re
+    from distutils.version import StrictVersion
+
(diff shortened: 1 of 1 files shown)


randombit:
Thanks.

Just trying to use `-j` then falling back on error might be cleaner overall but doesn't fit in well with what the doc build script is doing wrt building up a list of commands and then executing them later.

Another thing I thought of was maybe there is some escape hatch available within the Sphinx configuration, where we could set the flag then. I looked and this is probably possible, but would be very fragile in the face of any refactorings within Sphinx.

So, this seems good enough.

PR #1604, opened 2018-06-13
Address DSA/ECDSA side channel

randombit:
CVE-2018-0495


carnil:
FTR: The CVE id for this issue is CVE-2018-12435 (the CVE-2018-0495 was specific assigned for libgcrypt)

randombit:
@carnil Yes there was a miscommunication among the various parties and my understanding at the time was CVE-2018-0495 was being used for all affected implementations. I have already updated the release notes to use -12435 instead.

PR #1606, opened 2018-06-15
Make Montgomery exponentiation completely const time

randombit:
With exception that the size of the exponent leaks via the number of loop iterations, which I think we can fix, at least to only leaking the size of the exponent on a word-size granularity. Which is not ideal either but probably good enough.

Some slowdown as a result of this change but not too bad, seeing 1-5% for DH and RSA. Fortunately lots of optimizations in this release cycle so even with this change, RSA/DH/DSA are much faster than in 2.6.0

### DIFF
--- a/doc/manual/side_channels.rst
+++ b/doc/manual/side_channels.rst
@@ -109,8 +109,10 @@ Modular Exponentiation
 ------------------------
 
 Modular exponentiation uses a fixed window algorithm with Montgomery
-representation. A side channel silent table lookup is used to access the
-precomputed powers. See monty_exp.cpp
+representation. A side channel silent table lookup is used to access
+the precomputed powers. Currently the bit length of the exponent is
+leaked (with a granularity based on the window size, typically 4 bits)
+due to the number of loop iterations. See monty_exp.cpp
 
 Karatsuba multiplication algorithm avoids any conditional branches; in
 cases where different operations must be performed it instead uses masked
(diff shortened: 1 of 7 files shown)


briansmith:
> With exception that the size of the exponent leaks via the number of loop iterations, which I think we can fix, at least to only leaking the size of the exponent on a word-size granularity. Which is not ideal either but probably good enough.

It is important to not leak the size of the exponent. BoringSSL (and *ring*) solved this by fixing the exponent length to the modulus length, padding with zeros as necessary.

randombit:
> It is important to not leak the size of the exponent. BoringSSL (and ring) solved this by fixing the exponent length to the modulus length, padding with zeros as necessary.

Our situation is a little more complicated because for RSA we use exponent blinding (so the exponent > modulus), and for finite field DH/DSA we use short exponents that are sized to balance the strength of Pollard rho vs NFS (so exponent << modulus, and performance would suffer significantly if we just padded to size of modulus). But it occurs to me we can address both cases by having the caller indicate during the precomputation step the maximum expected bit length of the exponent, taking into account the blinding factors (for RSA) or the size of q (for DSA). Then the loop iteration count can be based on this static information, and will not depend on the exponent itself.

This is still slightly leaky for RSA case since loop iterations will still depend on bit length of p and q. If they are wildly imbalanced (say a 2048 key with 512 + 1536 bit factors) that will be visible. Of course such a key has more serious problems than side channels. Perhaps we should add something to verify when loading an externally generated key, that p and q differ by no more than 1 bit in length.

Thanks for the prod.

PR #1622, opened 2018-07-10
Add support for ARMv8 SM4 instructions

randombit:
Unfortunately almost none of this is testable in CI (old GCC, old qemu)


src/build-data/cc/gcc.txt
```diff
@@ -65,6 +65,10 @@ altivec -> "-maltivec"
 ppccrypto -> "-mcrypto"
 
 arm64:armv8crypto -> ""
+arm64:armv8sm3 -> "-march=armv8.2-a+sm4"
+arm64:armv8sm4 -> "-march=armv8.2-a+sm4"
+arm64:armv8sha512 -> "-march=armv8.2-a+sha3"
+arm64:armv8sha3 -> "-march=armv8.2-a+sha3"
```
noloader: Oh, this is good stuff. I think this may need `march=armv8.4-a` (plus the crypto, like `sm4`). Also see [Aarch64 ISA additions (SM3, SM4, SHA3)](http://gcc.1065356.n8.nabble.com/Aarch64-ISA-additions-SM3-SM4-SHA3-td1449695.html).

randombit: Apparently the new crypto extensions can be optionally implemented by 8.2-A architecture onwards, so this allows for that situation. Otherwise if we built for 8.4-a then ran on on 8.2-A processor with the extensions (such a processor may well never end up existing but hard to be sure), it would crash if the generated code used any additional instructions added in 8.3/8.4

PR #1629, opened 2018-07-17
Attempt to fix Pylint install issue

randombit:
Pylint just a 2.0 release which apparently breaks pip on 16.04 (if I'm reading the build log correctly https://travis-ci.org/randombit/botan/jobs/404994589)

Just force to last non-broken release.


randombit:
Oh pylint 2.0 dropped Python support.  https://github.com/PyCQA/pylint/issues/2294

So WTF is `python2-pip` trying to install something that isn't supported? :man_shrugging: Whatever.

PR #1640, opened 2018-08-07
Add XChaCha and XChacha20Poly1305

randombit:
plus a XSalsa bug fix


webmaster128:
Wow, finally a solid, independent implementation that can be tested against libsodium. Great, thank you!

PR #1648, opened 2018-08-13
Add XChaCha20Poly1305 test vectors generated by libsodium

webmaster128:
This adds a set of XChaCha20Poly1305 test vectors generated by libsodium. Those are systematically created in https://github.com/webmaster128/xchacha20poly1305-testvectors and include tests with empty messages and non-empty additional data.

This PR adds 1792 tests with ~0.5 MB of uncompressed plain text (size=2 from the repo).

Additionally I'd like to let you know that all 23392 tests (size=4) pass locally. If more or less tests should be checked in, I can change that easily.


randombit:
Nice, thanks for checking this.

I think there is a bit too much overlap between the different tests, eg there are 16 different tests with empty AD+text with different keys/nonces. I would rather that more AD/input lengths were tested, but each with just a single key/nonce pair. [Say every AD length 0...32 plus every plaintext length 0...48, for 1536 tests - plus maybe some say 512 byte inputs to test processing of longer messages.] That would be more likely to catch bugs like the one fixed in 646ddaef388.

webmaster128:
Thanks for the feedback, Jack!

I made the size levels much more structured now (https://github.com/webmaster128/xchacha20poly1305-testvectors#size-levels) and checked-in size level 3. This is
* a single key/nonce-pair
* AD length 0-21,31,32,33,63,64,65,128,533 (30)
* message lengths 0-48,63,64,65 (52)
giving us 1560 tests and ~0.5 MB of uncompressed plain text.

Again, all 100k tests from the greatest size level pass locally.

randombit:
Looks great thanks again

PR #1650, opened 2018-08-19
Regenerate xchacha20poly1305 testvectors

webmaster128:
This adds some more sense to message lengths. Vector sets of size >= 2 now ensure to include messages and ADs of length 1024 (2^10) and 577 (prime), i.e. divisible by almost everything and divisible by nothing. Those lengths replace the randomly chosen lengths 128 and 533.

See https://github.com/webmaster128/xchacha20poly1305-testvectors/commit/e89cc6d0eb70596a56c2983b47c5d68597eee96d


randombit:
Makes sense thanks

PR #1651, opened 2018-08-21
Default disable support for TLS v1.0/v1.1 and all CBC and CCM suites

randombit:
@securitykernel @neusdan @jurajsomorovsky  @webmaster128 Any reason this is not a good idea?

It will require some applications to change their policy, but having this be opt-in by those few application developers who need to interop with ancient peers means everyone else is protected by default.


webmaster128:
@randombit thanks for reaching out. We're not using TLS directly and I don't have any opinion on that.

mouse07410:
Disabling CCM is a very bad idea because some HSM devices use it for "wrapping" data blobs.  If you mean "CCM in TLS only" - I have no strong opinion, but would still like to keep an alternative to GCM, just in case.

jurajsomorovsky:
Thank you, Jack. We do not have any products/customers directly using TLS or statistics about its usage, so I cannot give you any recommendations. From a scientific and security perspective, I would say this is a great idea. 
And as you say, if somebody wants to use them, he can re-enable them.

securitykernel:
Thanks for reaching out. We don't allow TLS 1.0/1.1 anymore for a while now. Changing the default looks good to me now, also in the light of TLS 1.3 now being an official RFC.

PR #1652, opened 2018-08-22
Discard unused data in X509v3 CRL Distribution Points extension

RuecklS:
Currently Botan fails to parse X.509 certificates of the German smart meter root CAs (SM-Root.CA / SM-Test-Root.CA, cf. https://www.telesec.de/de/smartmetering-pki/downloadbereich/category/144-zertifikate) due to the "CRL Issuer" field within the "X509v3 CRL Distribution Points" extension.

I'm aware of the fact that this information would be crucial to perform revocation checks with indirect CRL issuers. However, in my opinion, this should not prevent the certificate from being loaded in the first place.

For your convenience I've attached an example certificate:
[sm-test-root.ca_sn2.txt](https://github.com/randombit/botan/files/2310005/sm-test-root.ca_sn2.txt)

### DIFF
--- a/doc/authors.txt
+++ b/doc/authors.txt
@@ -85,6 +85,7 @@ Simon Cogliani
 Simon Warta (Kullo GmbH)
 slaviber
 souch
+Stefan Rückl (Sagemcom Fröschl GmbH)
 t0b3
 tcely
 Technische Universitat Darmstadt
--- a/news.rst
+++ b/news.rst
@@ -71,6 +71,8 @@ Version 2.8.0, Not Yet Released
   removal of support using the configure.py flag ``--ack-vc2013-deprecated``
   (GH #1557)
 
+* Ignore additional but unused data in the X509v3 CRL Distribution Points extension
+
 Version 2.7.0, 2018-07-02
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/src/lib/x509/x509_ext.cpp
+++ b/src/lib/x509/x509_ext.cpp
@@ -953,7 +953,9 @@ void CRL_Distribution_Points::Distribution_Point::decode_from(class BER_Decoder&
         .decode_optional_implicit(m_point, ASN1_Tag(0),
                                   ASN1_Tag(CONTEXT_SPECIFIC | CONSTRUCTED),
                                   SEQUENCE, CONSTRUCTED)
-      .end_cons().end_cons();
+        .end_cons()
+        .discard_remaining()
+      .end_cons();
    }
 
 std::vector<uint8_t> CRL_Issuing_Distribution_Point::encode_inner() const


mouse07410:
> ... Certificate parsing should not fail...

This friends on whether the extension on question is flagged as Critical. 

If it is - then the parser must abort the processing of the certificate if it cannot completely parse/understand the given extension. 

If it is flagged as Non-Critical - then the parser can ignore the extension (or any parts of it) that it cannot understand (but it is allowed to ignore only what it cannot understand - this rule is not a carte blanche to ignore any Non-Critical extension the parser doesn't care for).

randombit:
@RuecklS The fix may end up being more complicated than what you propose in order to handle validation correctly. Can you please attach a full certificate chain ending at the trusted smartmeter root (+ CRLs) to allow us to test patches?

RuecklS:
Thank you for the quick response.

I think the whole "indirect CRL" thing is way too complex to offer any real security benefit. However, I have to deal with it and know that it's not that easy to implement it. The revocation checks are done by an external system in our use case, so it was sufficient for me to just get the certificates parsed.

In a general purpose implementation I agree with you that's not ok to silently ignore parsing errors in critical extensions (disclaimer: I've never seen a certificate that marked this extension critical so far, but you certainly should be able to handle this correctly nonetheless).

As requested I've attached some certificates you can use for your tests. They are publicly available at the site mentioned in my first post. Unfortunately, almost all information about this topic is written in German language only (for obvious reasons).

- [sm-test-root.ca_sn2.pem.txt](https://github.com/randombit/botan/files/2313235/sm-test-root.ca_sn2.pem.txt) -> trust anchor
- [sm-test-root.ca.tls-s_sn4.pem.txt](https://github.com/randombit/botan/files/2313234/sm-test-root.ca.tls-s_sn4.pem.txt) -> intermediate certificate authority
- [sm-test-root.ca.tls_sn6.pem.txt](https://github.com/randombit/botan/files/2313232/sm-test-root.ca.tls_sn6.pem.txt) -> end entity certificate
- [sm-test-root.ca.crl-s_sn2.pem.txt](https://github.com/randombit/botan/files/2313230/sm-test-root.ca.crl-s_sn2.pem.txt) -> CRL issuer
- [SM-Test-Root.CA.SM-Test-PKI-DE.crl.txt](https://github.com/randombit/botan/files/2313231/SM-Test-Root.CA.SM-Test-PKI-DE.crl.txt) -> CRL of root CA (issued by CRL-S)
- [SM-Test-Root.CA.TLS-S_SN4.crl.txt](https://github.com/randombit/botan/files/2313233/SM-Test-Root.CA.TLS-S_SN4.crl.txt) -> CRL of intermediate CA

randombit:
The change in #1657 allows parsing this certificate, except that the data that would be available in CRL DP extension is missing.

Next step is handling the full CRL DP extension.

randombit:
Closing. Full CRPDP handling PR incoming in next weeks.

PR #1658, opened 2018-08-23
Fix build without x509 module

tobiasbrunner:
Ifdefs were missing in the functions that were added with dc85761ef02c ("Add path validation to FFI").


randombit:
Thanks!

PR #288, opened 2015-09-25
Add the Darwin_SecRandom entropy source

tiwoc:
It uses the SecRandomCopyBytes function from the Security framework of OS X and iOS. We need this because it is the official way to get cryptographically secure random numbers on iOS, where /dev/random is not accessible due to sandboxing.

To get access to Security.framework, this PR also introduces framework support to the build system.

PR #1668, opened 2018-09-04
Remove Darwin SecRandomCopyBytes

randombit:
This API is the same RNG as arc4random and /dev/urandom. And arc4random seems to be working well for iOS and macOS, so lets remove this redundant source.

But, #288 says "We need this because it is the official way to get cryptographically secure random numbers on iOS" @tiwoc do you know if this is still true or is ``arc4random`` also considered acceptable these days, for purposes of app store review?

cc @webmaster128


tiwoc:
Yes, `arc4random` is totally fine on iOS, so I think removing `darwin_secrandom` is a good thing to remove some redundancy.

LGTM

mouse07410:
Thanks. *I'm speechless.*

PR #1670, opened 2018-09-05
New password hashing interface

### LINKED ISSUE #384: Be able to provide passwords as secure_vector<byte>
It seems undesirable  to have to pass in a string for the passphrase in various places when we could provide it as a secure vector. If we've already gone to the trouble of making sure the data is in a secure vector, it seems a waste of time (and a potential security problem) turning it into a string just to pass it to a function.

e.g.

```
OctetString PBKDF::derive_key(size_t output_len, const std::string &passphrase, const byte *salt, size_t salt_len, size_t iterations) const
```

randombit:
The current PBKDF interface assumes the configuration consists of a single size_t ("iterations"), but this does not work for Scrypt, Argon2, and others.

This adds a new interface which handles multiple parameters, as well as provides automatic tuning (eg choosing parameters for a particular runtime) and default parameters. PBKDF2, PGP-S2K, and Scrypt have been updated to the new interface (in addition to still supporting previous PBKDF API). I believe it will be possible to support Bcrypt-PBKDF and Argon2 within the same framework. PBKDF1 has not been updated, it stays PBKDF only and is eventually removed.

Fixes #384 now it is possible to hash by providing just a pointer and length, and so the password is not forced into a `std::string` which might leak bytes of the secret into the heap.

This project is a contribution by Ribose Inc (@riboseinc) 

cc @ronaldtse

### DIFF
--- a/doc/manual/pbkdf.rst
+++ b/doc/manual/pbkdf.rst
@@ -1,33 +1,47 @@
 
 .. _pbkdf:
 
-PBKDF Algorithms
+Password Based Key Derivation
 ========================================
 
-There are various procedures for turning a passphrase into a arbitrary
-length key for use with a symmetric cipher. A general interface for
-such algorithms is presented in ``pbkdf.h``. The main function is
-``derive_key``, which takes a passphrase, a salt, an iteration count,
-and the desired length of the output key, and returns a key of that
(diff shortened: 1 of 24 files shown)


ronaldtse:
Thank you @randombit !!

PR #1671, opened 2018-09-05
Fix build issues on windows when there is a space in the prefix folder

raffienficiaud:
I am having a little issue since my home folder contains a space. Everything works fine otherwise.
Tested on Win10.


randombit:
Thanks

PR #1675, opened 2018-09-07
Fix in-memory certificate store tests not running

oNaiPs:
If sqlite dependency is not used, tests should still be able to run.


randombit:
Nice, thanks

PR #1703, opened 2018-10-11
OCSP stapling enabled in TLS server

cryptosource-GmbH:
quote from previously exchanged email:
=>
for a project I implemented OCSP stapling in Botan in a rudimentary
version.

Previously the status_request TLS extension and the Certificate_Status
message were in the code, but the Botan server could do nothing else
than repeating the status_request extension to the client (which is
legitimate). It was not possible to actually send an OCSP response. I
implemented this now by adding a corresponding callback. The callback
will set the encoded OCSP-response, while the existing implementation
intended to provide it as an object. In our project we definitely need
to use the encoded response, since we want to use the server to also
send invalid responses for testing purposes. However, I generally think
that unnecessary encoding / decoding should be avoided. Any unsupported
feature of OCSP-responses or a bug in the encoding or decoding routines
would unnecessarily affect the OCSP stapling. So I suggest to remove the
original approach using the response object altogether (as I wrote above
it was so far unusable anyway). This I have not yet done.
<=

In the meanwhile I have also completed the signature of the OCSP-response providing callback with the server's certificate chain and the decoded status_request extension from the clientHello.

### DIFF
--- a/src/lib/tls/msg_cert_status.cpp
+++ b/src/lib/tls/msg_cert_status.cpp
@@ -41,23 +41,41 @@ Certificate_Status::Certificate_Status(Handshake_IO& io,
    {
    hash.update(io.send(*this));
    }
+Certificate_Status::Certificate_Status(Handshake_IO& io,
+                                       Handshake_Hash& hash,
+                                       std::vector<uint8_t> const& raw_response_bytes) :
+   m_raw_response_bytes(raw_response_bytes)
+   {
+   hash.update(io.send(*this));
+   }
 
 std::vector<uint8_t> Certificate_Status::serialize() const
    {
-   BOTAN_ASSERT_NONNULL(m_response);
-   const std::vector<uint8_t>& m_resp_bits = m_response->raw_bits();
+   const std::vector<uint8_t>* resp_bits;
+   if(m_raw_response_bytes.size() == 0)
+      {
+      BOTAN_ASSERT_NONNULL(m_response);
+      resp_bits = &m_response->raw_bits();
+      }
+   else
+      {
+      BOTAN_ASSERT(m_raw_response_bytes.size() != 0,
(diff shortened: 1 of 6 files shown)


randombit:
Sorry about delay here, looks OK though I want to write some covering tests before merge.

falko-strenzke:
Ok, what exactly do you want to cover with tests? Maybe I can help...

randombit:
Automated tests covering at least 1) a successful transfer/processing of a response 2) a transfer of an invalid response which the client rejects.

Plus ideally some way of doing OCSP in the CLI so that for example stapling can be tested with `tls_http_server` and a standard browser.

falko-strenzke:
> Automated tests covering at least 1) a successful transfer/processing of a response 2) a transfer of an invalid response which the client rejects.
In order to determine the effort for the implementation of such tests as a precondition to get the approval to realize these in the scope of our current project, I have some further questions regarding the tests:
What do you mean exactly by "transfer/processing"? Would that be an execution of a TLS handshake in the course of the unit tests? Where would such tests be implemented exactly within the Botan test code?
What I implemented is the support of the extension in Botan TLS server, so I would only implement tests covering the behaviour regarding Botan's TLS server...

> Plus ideally some way of doing OCSP in the CLI so that for example stapling can be tested with `tls_http_server` and a standard browser.
This is straightforward, I basically have already done that within our project. I would add a command-line parameter "--stapled_response=..." the file to which this points the TLS server would then send whenever the client requests the certificate status in the ClientHello.

randombit:
With BoGo merged we now have a way of comprehensively testing server OCSP handling, so this just needs a rebase plus enabling the relevant tests in BoGo runner. I'll look at this soon.

randombit:
Taken up in #1967

Thanks again for contributing this

PR #1704, opened 2018-10-14
move instead of copy

LesnyRumcajs:
prevent some unnecessary copies with usage of emplace_back where it's feasible


randombit:
I doubt any of these make much different wrt performance but I guess it can't hurt

PR #1726, opened 2018-11-02
Add missing <stdexcept> <string> necessary

Cheney-W:
This causes compile failures with VS2019 previews. Add \<stdexcept> include to use std::invalid_argument and add \<string> include to use std::string.


randombit:
Thank you

PR #1728, opened 2018-11-04
Add generic SIMD ChaCha and fix counter bugs

randombit:
Adds a generic ChaCha using SIMD_4x32 that will work with SSE2, NEON or AltiVec. Also it is smarter and avoids the shuffles each round so it's about 10% faster on x86.

Amusingly? all 4 implementations of ChaCha (scalar, SSE2, NEON and AVX2) had buggy behavior right around the 256 GB point (when the low 32-bit part of the counter overflows). They produced correct output before and after.


randombit:
NEON sees a slowdown here even though less instructions are used in the main loop. Best guess is NEON load/store in SIMD_4x32 is non-optimal.

randombit:
Will deal with NEON perf issues in a different PR

PR #1735, opened 2018-11-08
Switch LibreSSL provider to OpenSSL 1.1 API

bluhm:
LibreSSL 2.7 has added parts of the OpenSSL 1.1 API.  Remove the
defined(LIBRESSL_VERSION_NUMBER) as it is not necessary anymore.
Switch to the newer API unconditionally, older LibreSSL version are
not supported.


randombit:
Nice thanks

PR #1736, opened 2018-11-08
Run RSA blinding tests with provider

bluhm:
The RSA blinding test was always run with the base provider.  Run
these tests in a loop with possible providers base and openssl.
The random number generator and the blinder are not used by the
openssl provider, so skip these tests.


randombit:
I don't really understand the purpose of this change. The main point of these tests it to check on the behavior of the blinding used in RSA, but we don't use blinding when calling OpenSSL. So why run the test at all? As it is, it is essentially just generating signatures for a sequence of messages and checking that they verify - but that behavior should already be tested by the other known answer tests.

bluhm:
I was confused that the test was also run with --provider=openssl.   Thanks for pointing out that this makes no sense.  So let's run this explicitly only with the base provider.

randombit:
OK makes sense now, thanks

PR #1737, opened 2018-11-09
Inline the contents of mp_core.cpp

randombit:
Allows better optimizations. Especially improves ECC. ECDSA P-256 15-20% faster, lesser improvements for larger curves (7% for 512-bit). Tiny (< 1%) improvements for RSA/DH.

Downside is it makes profiling a little trickier, and will affect code size negatively, so I'm not entirely sure this should merge.

Todo before merge: compare code size (compiling with `-Os`) with/without this change on x86-64 and ARM.


randombit:
When optimizing for size, there is basically no difference in the size of the shared lib between master and this branch.

Arch | Size (`master`) | Size (`mp-inline`)
-------| -----------------------| ------------------------
aarch64 | 5106592   | 5106624
x86-64 | 5085880 | 5085616

Issue #1742, opened 2018-11-15
Mapping all exceptions to error_codes


ChrisBFX:
Somewhat related to #1299
I'm currently trying to map all exceptions from Botan to boost::system::error_codes and it somewhat works (see https://github.com/krumedia/asio-botan-stream/blob/master/include/asio/botan/detail/ConvertExceptions.h).
The hundreds of throws inside Botan, which don't use special exception classes but only have a error message string are quite annoying though. These make it quite hard to debug cause i can only map those to "unknown error".

Would it be feasible to change those throws to include a distinct error-code from an enum? This should also include the TLS_Exceptions, cause the alert type is often not enough.

randombit:
> Would it be feasible to change those throws to include a distinct error-code from an enum? 

Do you mean a distinct error code attached to each and every exception? That seems kind of excessive to me.

But, some changes related to that that do seem reasonable:

- Never throw bare `Exception` (make the constructors protected) but always some descriptive subclass, since it is true, catching a base `Exception` just tells you "something went wrong" with no real hint unless you parse the message. (And, specific content of exception messages are not considered part of the stable API, so that doesn't really work anyway.)
- Have a reason code enum ala
```
enum class ErrorType {
  InvalidArgument,
  InternalError,
  ...
}
```
exposed by new virtual `Exception::error_type() const nothrow` which kind of reflects the existing structure of Exception types. That would also simplify code like https://github.com/randombit/botan/blob/master/src/lib/ffi/ffi.cpp#L38 which is doing something much like what you linked to.
- Capture `__FILE__`, `__LINE__`, and `__func__` in exceptions (probably with the help of throwing helpers that capture these values for us). And, if all throws go through some small set of functions that makes it possible to capture those for logging/debugging which is often useful.

Just got back from a trip so ^ is just initial reaction I'll think about this more. Let me know what you think.

randombit:
@ChrisBFX There are some initial improvements in this area in #1744, I would welcome your comments.

randombit:
I've merged #1744 if you can think of something else that would help in this area I would certainly consider it.

randombit:
We have error code enums that are attached onto `Exception`.

PR #1744, opened 2018-11-17
Avoid throwing base Botan::Exception type

randombit:
As it is difficult to determine what the error is when such an exception is caught.

See GH #1742


ChrisBFX:
Thanks a lot for looking into and working on this!
Removing the usage of the base exception and the added error_code looks good.
I'm not sure if I understand the motivation behind the ErrorType enum though. It looks like it only changes a bunch of catch() clauses into a switch case?
The combination of error_code and error_type looks a lot like what the std::error_code does (value and category), at least semantically. Anyhow i think it would be better if the enum starts at 1 (most conventions are that 0 is not an error).

randombit:
Looking at it more, `std::error_code` and co seems kind of oddly designed and I'm not sure we want to use it directly. Or at least not until I understand it better. For now I've stuck with `error_type` and `error_code`, and if we want to support the std APIs later I think that can be built on top of these.

I renumbered the `ErrorType` enum to start at 1 thanks for the suggestion.

dewyatt:
@randombit Just a note, making the `Exception` constructor protected breaks the softhsm build at the moment. Not sure if other projects actually use that. It seems odd to use Botan's exception class but I guess it's one of those "if you build it they will come" type of things.

randombit:
@dewyatt Thanks for letting me know, I'll revert the part that makes the base class constructor protected.

I also just added build/test SoftHSM using latest code as part of the release checklist.

dewyatt:
> Thanks for letting me know, I'll revert the part that makes the base class constructor protected.

@randombit That's great! I went ahead and did a PR for SoftHSM to throw `Encoding_Error` instead, which I figure should make it compatible with more commits of botan regardless.

> I also just added build/test SoftHSM using latest code as part of the release checklist.

Sweet! One thing to note just in case you're not aware, I think they use a slightly unusual branching strategy where `master` is the latest release and `develop` (their default branch) is where development really happens (more akin to Botan's master).

Issue #730, opened 2016-11-17
Using PKCS#11 with TLS


securitykernel:
This is a follow-up of #725, where @mouse07410 asked about using PKCS#11 objects with TLS. Correct me if I'm wrong, but from my understanding it should already be possible to have a TLS server with the server's long-term key stored in a PKCS#11 device, now as the PK ops are tied to the `Private_Key` and `Public_Key` objects (which was introduced in #625). All you have to do is to create a PKCS11 key, e.g., a `PKCS11_RSA_PrivateKey`, and return it from `Credentials_Manager::private_key_for()`. `TLS_Server` will forward it to `PK_Signer`, which will select and sign with the PK signature op returned by `PKCS11_RSA_PrivateKey::create_signature_op()`.

If you want more control over the handshake, e.g., you want do the ECDH key exchange ops in a PKCS#11 device, there is currently no way to do it. We are currently thinking about a new interface, let it be named TLS Adapter for now, for doing these operations in hardware. We will have to do these operations in a non-PKCS#11 compliant HSM, but we want to make it generic enough that PKCS#11 devices can be supported easily.

We will come up with a proposal for this use case probably in the next few weeks.

*Update: Removed the chain validation part, of course it's already possible by overriding `Credentials_Manager::verify_certificate_chain()`.*

randombit:
> All you have to do is to create a PKCS11 key, e.g., a PKCS11_RSA_PrivateKey, and return it from Credentials_Manager::private_key_for(). TLS_Server will forward it to PK_Signer

That is how it should work but in fact the server currently `dynamic_cast`s its key to a `RSA_PrivateKey` when processing the client key exchange msg, so it would reject a PKCS11 key. It should probably just check `algo_name` matches instead.

cordney:
Thanks for the hint. I am currently implementing this with a `TPM_PrivateKey` type that uses our custom TSS for a TPM key. I will fix whatever is necessary.

mouse07410:
1. Yes, it would be great to have the TLS server (and the client!) fixed, so they can accept a PKCS11 key.

2. Yes it would be great if ECDH would become supported.

Thanks!

cordney:
[...]

PR #1769, opened 2018-12-04
Extend the SignMechanisms map for Signature_Scheme padding strings

hrantzsch:
This addresses #730. The tls_sign_message callback was unable to handle
emsa strings produced as padding strings for TLS::Signature_Scheme, due
to inconsistent naming (mostly between EMSA3 and EMSA_PKCS1).

### DIFF
--- a/src/lib/prov/pkcs11/p11_mechanism.cpp
+++ b/src/lib/prov/pkcs11/p11_mechanism.cpp
@@ -75,11 +75,6 @@ static std::map<std::string, RSA_SignMechanism> SignMechanisms =
    {
       { "Raw", RSA_SignMechanism(MechanismType::RsaX509) },
 
-      { "EMSA1(SHA-1)", RSA_SignMechanism(MechanismType::DsaSha1) },
-      { "EMSA1(SHA-256)", RSA_SignMechanism(MechanismType::DsaSha256) },
-      { "EMSA1(SHA-384)", RSA_SignMechanism(MechanismType::DsaSha384) },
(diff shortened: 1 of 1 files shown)


src/lib/prov/pkcs11/p11_mechanism.cpp
```diff
@@ -75,6 +75,11 @@ static std::map<std::string, RSA_SignMechanism> SignMechanisms =
    {
       { "Raw", RSA_SignMechanism(MechanismType::RsaX509) },
 
+      { "EMSA1(SHA-1)", RSA_SignMechanism(MechanismType::DsaSha1) },
```
hrantzsch: Note that these are ambiguous. We could come from either DSA or ECDSA Signature Schemes. Do you think it would be better not to allow them at all here?

randombit:
Thanks I'll try to look at this soon.

src/lib/prov/pkcs11/p11_mechanism.cpp
randombit: As we cannot handle this unambiguously it is probably better to not have them and fail instead.

randombit:
I'm not happy with this as complete fix for the issues discussed in #730, I feel like we should be doing better. That said it seems like this fixes the immediate problem wrt TLS + PKCS11, so if you are good with this then I suppose I am as well.

Remove the EMSA1 encodings though, since as you say we cannot determine the correct P11 mechanism. Also - looking at it further I think the DSA/ECDSA params should not go into this struct anyway as it is just for RSA, right?

CC @neusdan as the PKCS11 expert

randombit:
Merged, thanks!

PR #1781, opened 2018-12-09
Test GCC 4.8 as 32-bit x86

randombit:
We are currently not testing the GCC x86-32 inline asm anywhere in CI


randombit:
For whatever reason this don't work

PR #1800, opened 2018-12-29
New Memory_Pool implementation

randombit:
Quite a bit faster than the old version, and with better properties wrt alignment

Not for 2.9, want to test this thoroughly so will merge after 2.9 release. Also the bucket sizes probably need tuning, current ones are guesswork. Finally, want to write a fuzzer which tests the pool operation.

### DIFF
--- a/src/build-data/buildh.in
+++ b/src/build-data/buildh.in
@@ -120,13 +120,6 @@
 /* How much to allocate for a buffer of no particular size */
 #define BOTAN_DEFAULT_BUFFER_SIZE 1024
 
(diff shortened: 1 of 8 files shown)


randombit:
With this design each page is used separately, no allocations cross page boundaries. So we can modify `allocate_locked_pages` to return a list of individual locked pages, each one preceded by and followed by a guard page (`mprotect(PROT_NONE)` / `VirtualProtect(PAGE_NOACCESS)`). That turns most out of bound reads/writes into crashes.

PR #1805, opened 2019-01-03
Reduce AppVeyor matrix

randombit:
AppVeyor is slow and compiler caching doesn't help because of how AppVeyor implements it (splits cache among the various VM providers, ensuring a near 0% hit rate)

Remove VC2015 x86-64 build, leaving only the x86-32 build. Choice of platform to leave is based on the fact that x86-32 is the faster build/test cycle.

Use VC2019 for the sanitizer tests and drop the VC2017 sanitizers. That gives the benefit of testing a compile under VC2019, and also uses the latest and greatest? debug iterator support, while eliminating a job.

This should cut about 20 minutes from the CI build.


randombit:
Closing in favor of #1807

PR #1808, opened 2019-01-06
Proposal of sandboxing feature.

devnexen:
For now only used by the TLS server.

### DIFF
--- a/src/cli/sandbox.cpp
+++ b/src/cli/sandbox.cpp
@@ -1,0 +1,44 @@
+/*
+* (C) 2019 David Carlier <devnexen@gmail.com>
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include "sandbox.h"
+#include <botan/build.h>
+
+#if defined(BOTAN_TARGET_OS_HAS_PLEDGE)
+  #include <unistd.h>
+#elif defined(BOTAN_TARGET_OS_HAS_CAP_ENTER)
+  #include <sys/capsicum.h>
+#endif
+
+namespace Botan_CLI {
+
+Sandbox::Sandbox()
+   {
+#if defined(BOTAN_TARGET_OS_HAS_PLEDGE)
+   m_name = "pledge";
+#elif defined(BOTAN_TARGET_OS_HAS_CAP_ENTER)
+   m_name = "capsicum";
+#else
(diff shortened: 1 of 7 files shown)


randombit:
Sorry for the delay here, was traveling.

In general I certainly approve of the concept, it would be great to sandbox the CLI to minimize attack surface.

While the docs state that any OS specific code should go in os_utils I don't think it is appropriate in this situation because there is no circumstance (at least that I can think of at the moment) where it would be appropriate to enter sandbox mode from within the library - the calling application would need to handle that. So it should be in `src/cli/sandbox.{h,cpp}` instead.

I think `pledge` should be called with `error` set so it returns ENOSYS rather than aborting - an accidental sandbox violation crashing makes for an unpleasant and hard to debug user experience. Perhaps in the future when we are more confident that the logic is good (ie that we are opening all resources we need prior to the sandbox being entered) then we can switch to abort mode.

After merge I will toy a bit with getting other CLI items sandboxed and adding a mode for seccomp.

src/cli/sandbox.cpp
```diff
@@ -0,0 +1,44 @@
+/*
+* (C) 2019
```
randombit: This should have your name or a company name, whoever owns copyright on this work.

src/cli/sandbox.h
```diff
@@ -0,0 +1,27 @@
+/*
+* (C) 2019
```
randombit: Same

src/cli/sandbox.h
```diff
@@ -0,0 +1,27 @@
...
+   public:
+	   explicit Sandbox();
+	   bool Init();
+	   virtual ~Sandbox();
+	   const std::string& name() const
+              {
+              return s_name;
+	      } 
+   private:
+	   std::string s_name;
```
randombit: This should be `m_name` since it is not a static

src/cli/sandbox.h
```diff
@@ -0,0 +1,27 @@
...
+
+#include <string>
+
+namespace Botan_CLI {
+
+class Sandbox
+   {
+   public:
+	   explicit Sandbox();
+	   bool Init();
```
randombit: Should be `init` to match style convention used within the library.

randombit:
Looks good thank you!

PR #1810, opened 2019-01-10
Corrects TLS alerts

mgierlings:
When parsing an invalid/unknown signature scheme in `tls_algos.cpp`
an Invalid_State exception was thrown. This exception could then
bubble up to `TLS:Channel:received_data`, where it would be caught by
a generic `catch(...)` statement and mapped to an INTERNAL_ERROR
alert. Since the inability to select a proper signature algorithm
is definitely protocol related an INTERNAL_ERROR alert may not
be thrown here. Instead an ILLEGAL_PARAMETER alert is more
appropriate. According to RFC5246 it is to be used in case "a
field in the handshake was out of range or inconsistent with
other fields. This commit implements the according fix.

Note: This patch is probably not exhaustive and similar issues may be 
found elsewhere in the code base.

### DIFF
--- a/src/lib/tls/msg_cert_req.cpp
+++ b/src/lib/tls/msg_cert_req.cpp
@@ -1,10 +1,13 @@
 /*
 * Certificate Request Message
 * (C) 2004-2006,2012 Jack Lloyd
+*     2019 Matthias Gierlings
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
 
+#include <botan/tls_alert.h>
+#include <botan/tls_exceptn.h>
 #include <botan/tls_messages.h>
 #include <botan/tls_extensions.h>
 #include <botan/internal/tls_reader.h>
@@ -43,7 +46,7 @@ uint8_t cert_type_name_to_code(const std::string& name)
    if(name == "ECDSA")
       return 64;
 
-   throw Invalid_Argument("Unknown cert type " + name);
+   throw TLS_Exception(Alert::CERTIFICATE_UNKNOWN, "Unknown cert type " + name);
    }
 
 }
@@ -74,7 +77,7 @@ Certificate_Req::Certificate_Req(const std::vector<uint8_t>& buf,
                                  Protocol_Version version)
    {
    if(buf.size() < 4)
-      throw Decoding_Error("Certificate_Req: Bad certificate request");
+      throw TLS_Exception(Alert::DECODE_ERROR, "Certificate_Req: Bad certificate request");
 
    TLS_Data_Reader reader("CertificateRequest", buf);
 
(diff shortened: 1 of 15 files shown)


randombit:
Thanks reasoning make sense but I don't think it is good to convert any `Invalid_Argument` into `ILLEGAL_PARAMETER` since this may have been thrown due to an internal problem (ie we called some other library API with a bogus input due to some bug in the TLS layer).

I think best fix is to modify the throws in `tls_algos.cpp` to be `throw TLS_Exception(Alert::ILLEGAL_PARAMETER, ...)` instead, then we get correct behavior for this case without converting other invalid arguments into a potentially misleading alert.

randombit:
> Once we are in TLS territory sending INTERNAL_ERROR is always the wrong thing to do.

I don't think I agree with your interpretation of `internal_error`. My read of "An internal error unrelated to the peer or the correctness of the protocol" is that "an error has occurred that makes it impossible to continue, but the occurrence of this error is not the fault of anything the peer sent"

Consider for example a throw in `msg_server_kex.cpp` -
`throw Internal_Error("Negotiated X25519 somehow, but it is disabled")`. This occurs if we have agreed to use X25519 despite it being disabled in the build. This should never happen. If it does ever trigger, it implies a bug in the library. That is, an internal error. (Certainly not other alert type seems to apply to such a situation.)

mgierlings:
> 
> 
> > Once we are in TLS territory sending INTERNAL_ERROR is always the wrong thing to do.
> 
> I don't think I agree with your interpretation of `internal_error`. My read of "An internal error unrelated to the peer or the correctness of the protocol" is that "an error has occurred that makes it impossible to continue, but the occurrence of this error is not the fault of anything the peer sent"
> 
> Consider for example a throw in `msg_server_kex.cpp` -
> `throw Internal_Error("Negotiated X25519 somehow, but it is disabled")`. This occurs if we have agreed to use X25519 despite it being disabled in the build. 

I think that if we have agreed upon a cryptographic algorithm i.e. X25519 as part of a TLS handshake this is related to the correctness of the protocol. There can be several reasons for this problem including:
- we offered X25519 during handshake despite not supporting it (violation of TLS on our side, thus protocol related)
- we didn't offer X25519 during handshake and use it despite another algorithm was negotiated (violation of TLS on our side, thus protocol related)
- the peer chose to use X25519 despite us explicitly not supporting it (TLS violation on client side and additionally peer related)
- the peer uses X25519 but it was never agreed upon (TLS violation on client side and additionally peer related)

Agreement on used algorithms is completely covered by the TLS standard, either it is a protocol violation on our side or it is a peer related problem. In both cases it can not be an `INTERNAL_ERROR`. 

> This should never happen. If it does ever trigger, it implies a bug in the library.  That is, an internal error. (Certainly not other alert type seems to apply to such a situation.)

Even if it should never happen and it is caused by some bug in the library it is still a protocol related error. The `INTERNAL_ERROR` is only suitable for those errors that are not explicitly covered by the standard. A typical `INTERNAL_ERROR` would be: Everything went fine according to protocol, we allocate some memory, the OS does not grant us additional memory (maybe because we're out of RAM) or due to some limit, as a result we can not go on. The TLS Standard does not deal with OS/plattform specifics, or resource constraints imposed by those. That is what the `INTERNAL_ERROR` is for.

Issue #1731, opened 2018-11-06
Working on an asio::ssl::stream<> replacement based on botan's TLS


reneme:
We are actively developing a client application that mainly communicates via HTTP and already uses botan for several cryptographic tasks. Now, we would like to replace its networking layer (currently WinHTTP and SChannel) with an open source implementation.

For a socket abstraction we are planning to use boost::asio. To avoid the addition of OpenSSL as a second crypto library we are looking into a replacement for asio's `ssl::stream<>` based on botan. Its not yet clear whether this will be a compatible drop-in replacement but we would take `asio::ssl::stream<>` as a guide line.

To the best of our knowledge to date there is no such implementation available as open source. Since some (optional) functionality in botan is already based on ASIO: Would you be interested in accepting such an ssl stream implementation as a contribution to the library?

randombit:
Keep in mind also it is perfectly possible to use the existing TLS API with asio, see for example https://github.com/randombit/botan/blob/master/src/cli/tls_proxy.cpp and https://github.com/randombit/botan/blob/master/src/cli/tls_http_server.cpp

That said an explicit asio integration would probably be useful to others and yes I would accept a PR for this. This could be packaged in a sub-module of tls.

Style thing I'll mention from the outset, in my install of Boost the asio SSL support is spread out over 27 headers which to me seems really excessive especially as most are under 200 LOC. Something like say 3-5 headers seems more reasonable to me here.

I'm going to be unavailable quite a bit in the next few weeks but if you have any questions about direction or how to do something feel free to ask here.

ChrisBFX:
@reneme  I've worked on exactly the same thing last week. It's still pretty much work in progress but the basic functionality is done. You can check it out here: https://github.com/krumedia/asio-botan-stream
Would be great if you can provide some kind of feedback if this would be useful for you, or if don't like the architecture at all.

reneme:
Thanks for your reply @ChrisBFX. Indeed your implementation seems to be on a good track and might serve as a basis for what we have in mind as well. Are you planning to develop this to become production ready?

Tymolc:
We opened PR #1839 to implement this feature.

reneme:
Since this is merged and integrated into our application, I take the liberty to close this.
Thanks a lot to all contributors. :)

pedro-vicente:
Is there a version for asio standalone (non boost) ?
and if not , would it be possible to change the current botan source to use the standalone asio ?

https://think-async.com/Asio/

[...]

PR #1830, opened 2019-02-11
Access to System Certificate Store on macOS

reneme:
This is ongoing work towards an adapter for the macOS system certificate store onto Botan's `Certificate_Store` interface. I will add documentation, description and tests along the way and would be happy about a quick code review. Particularly, whether my build-system integration is reasonable. :-)

Currently implemented are:

* `Certificate_Store::find_cert()`
* `Certificate_Store::find_all_certs()`
* `Certificate_Store::find_cert_by_pubkey_sha1()`
* `Certificate_Store::all_subjects()`

However, I'm not sure whether `::find_crl_for()` and `::find_cert_by_raw_subject_dn_sha256()` are in reach.

Furthermore, I'm a bit worried about testing this. The tests added for now try to fetch the root certificate that signs the TLS certificate of "botan.randombit.net" in different ways. However, more sophisticated tests would require root access to the CI machine, to alter the root certificate chains. I don't think that this is feasible.

### DIFF
--- a/src/build-data/os/darwin.txt
+++ b/src/build-data/os/darwin.txt
@@ -21,6 +21,8 @@ commoncrypto
 sockets
 threads
 filesystem
+
+apple_keychain
 </target_features>
 
 <aliases>
--- a/src/lib/x509/certstor_system_macos/certstor_macos.cpp
+++ b/src/lib/x509/certstor_system_macos/certstor_macos.cpp
@@ -1,0 +1,472 @@
+/*
+* Certificate Store
+* (C) 1999-2019 Jack Lloyd
+* (C) 2019      René Meusel
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/build.h>
+
+#include <algorithm>
+#include <array>
+
+#include <CoreFoundation/CoreFoundation.h>
+#include <CoreServices/CoreServices.h>
+
+#include <botan/assert.h>
+#include <botan/ber_dec.h>
+#include <botan/certstor_macos.h>
+#include <botan/data_src.h>
+#include <botan/der_enc.h>
+#include <botan/exceptn.h>
+#include <botan/x509_dn.h>
+
+namespace Botan {
+
+namespace {
+
+/**
+ * Abstract RAII wrapper for CFTypeRef-style object handles
+ * All of those xxxRef types are eventually typedefs to void*
+ */
+template<typename T>
+class scoped_CFType
+   {
+   public:
+      explicit scoped_CFType(T value)
+         : m_value(value)
+         {
+         }
+
+      scoped_CFType(const scoped_CFType<T>& rhs) = delete;
+      scoped_CFType(scoped_CFType<T>&& rhs) :
+         m_value(std::move(rhs.m_value))
+         {
+         rhs.m_value = nullptr;
+         }
+
+      ~scoped_CFType()
+         {
+         if(m_value)
+            {
+            CFRelease(m_value);
+            }
+         }
+
+      operator bool() const { return m_value != nullptr; }
+
+      void assign(T value)
+         {
+         BOTAN_ASSERT(m_value == nullptr, "scoped_CFType was not set yet");
+         m_value = value;
+         }
+
+      T& get() { return m_value; }
+      const T& get() const { return m_value; }
+
+   private:
+      T m_value;
+   };
+
+/**
+ * Apple's DN parser "normalizes" ASN1 'PrintableString' into upper-case values
+ * and strips leading, trailing as well as multiple white spaces.
+ * See: opensource.apple.com/source/Security/Security-55471/sec/Security/SecCertificate.c.auto.html
+ */
+X509_DN normalize(const X509_DN& dn)
+   {
+   X509_DN result;
+
+   for(const auto& rdn : dn.dn_info())
+      {
+      // TODO: C++14 - use std::get<ASN1_String>(), resp. std::get<OID>()
+      const auto oid = rdn.first;
+      auto str = rdn.second;
+
+      if(str.tagging() == ASN1_Tag::PRINTABLE_STRING)
+         {
+         std::string normalized;
+         normalized.reserve(str.value().size());
+         for(const char c : str.value())
+            {
+            if(c != ' ')
+               {
+               // store all 'normal' characters as upper case
+               normalized.push_back(::toupper(c));
+               }
+            else if(!normalized.empty() && normalized.back() != ' ')
+               {
+               // remove leading and squash multiple white spaces
+               normalized.push_back(c);
+               }
+            }
+
+         if(normalized.back() == ' ')
+            {
+            // remove potential remaining single trailing white space char
+            normalized.erase(normalized.end() - 1);
+            }
+
+         str = ASN1_String(normalized, str.tagging());
+         }
+
+      result.add_attribute(oid, str);
+      }
+
+   return result;
+   }
+
+std::string to_string(const CFStringRef cfstring)
+   {
+   const char* ccstr = CFStringGetCStringPtr(cfstring, kCFStringEncodingUTF8);
+
+   if(ccstr != nullptr)
+      {
+      return std::string(ccstr);
+      }
+
+   auto utf16_pairs    = CFStringGetLength(cfstring);
+   auto max_utf8_bytes = CFStringGetMaximumSizeForEncoding(utf16_pairs, kCFStringEncodingUTF8);
+
+   std::vector<char> cstr(max_utf8_bytes, '\0');
+   auto result = CFStringGetCString(cfstring,
+                                    cstr.data(), cstr.size(),
+                                    kCFStringEncodingUTF8);
+
+   return (result) ? std::string(cstr.data()) : std::string();
+   }
+
+std::string to_string(const OSStatus status)
+   {
+   scoped_CFType<CFStringRef> eCFString(
+      SecCopyErrorMessageString(status, nullptr));
+   return to_string(eCFString.get());
+   }
+
+void check_success(const OSStatus status, const std::string context)
+   {
+   if(errSecSuccess == status)
+      {
+      return;
+      }
+
+   throw Internal_Error(
+      std::string("failed to " + context + ": " + to_string(status)));
+   }
+
+template <typename T>
+void check_notnull(const scoped_CFType<T>& value, const std::string context)
+   {
+   if(value)
+      {
+      return;
+      }
+
+   throw Internal_Error(std::string("failed to ") + context);
+   }
+
+SecCertificateRef to_SecCertificateRef(CFTypeRef object)
+   {
+   if(!object || CFGetTypeID(object) != SecCertificateGetTypeID())
+      {
+      throw Internal_Error("cannot convert CFTypeRef to SecCertificateRef");
+      }
+
+   return static_cast<SecCertificateRef>(const_cast<void*>(object));
+   }
+
+/**
+ * Create a CFDataRef view over some provided std::vector<uint8_t. The data is
+ * not copied but the resulting CFDataRef uses the std::vector's buffer as data
+ * store. Note that the CFDataRef still needs to be manually freed, hence the
+ * scoped_CFType wrapper.
+ */
+scoped_CFType<CFDataRef> createCFDataView(const std::vector<uint8_t>& data)
+   {
+   return scoped_CFType<CFDataRef>(
+             CFDataCreateWithBytesNoCopy(kCFAllocatorDefault,
+                                         data.data(),
+                                         data.size(),
+                                         kCFAllocatorNull));
+   }
+
+/**
+ * Convert a SecCertificateRef object into a Botan::X509_Certificate
+ */
+std::shared_ptr<const X509_Certificate> readCertificate(SecCertificateRef cert)
+   {
+   scoped_CFType<CFDataRef> derData(SecCertificateCopyData(cert));
+   check_notnull(derData, "read extracted certificate");
+
+   // TODO: factor this out into a createDataSourceView() as soon as this class
+   //       gets a move-constructor
+   const auto data   = CFDataGetBytePtr(derData.get());
+   const auto length = CFDataGetLength(derData.get());
+
+   DataSource_Memory ds(data, length);
+   return std::make_shared<Botan::X509_Certificate>(ds);
+   }
+
+}
+
+/**
+ * Internal class implementation (i.e. Pimpl) to keep the required platform-
+ * dependent members of Certificate_Store_MacOS contained in this compilation
+ * unit.
+ */
+class Certificate_Store_MacOS_Impl
+   {
+   private:
+      static constexpr const char* system_roots =
+         "/System/Library/Keychains/SystemRootCertificates.keychain";
+      static constexpr const char* system_keychain =
+         "/Library/Keychains/System.keychain";
+
+   public:
+      using Query = std::vector<std::pair<CFStringRef, CFTypeRef>>;
+
+   public:
+      Certificate_Store_MacOS_Impl() :
+         m_policy(SecPolicyCreateBasicX509()),
+         m_system_roots(nullptr),
+         m_system_chain(nullptr),
+         m_keychains(nullptr)
+         {
+         check_success(SecKeychainOpen(system_roots, &m_system_roots.get()),
+                       "open system root certificates");
+         check_success(SecKeychainOpen(system_keychain, &m_system_chain.get()),
+                       "open system keychain");
+         check_notnull(m_system_roots, "open system root certificate chain");
+         check_notnull(m_system_chain, "open system certificate chain");
+
+         // m_keychains is merely a convenience list view into all open keychain
+         // objects. This list is required in prepareQuery().
+         std::array<const void*, 2> keychains{{
+               m_system_roots.get(),
+               m_system_chain.get()
+               }};
+
+         m_keychains.assign(
+            CFArrayCreate(kCFAllocatorDefault,
+                          keychains.data(),
+                          keychains.size(),
+                          &kCFTypeArrayCallBacks));
+         check_notnull(m_keychains, "initialize keychain array");
+         }
+
+      CFArrayRef keychains() const { return m_keychains.get(); }
+      SecPolicyRef policy() const { return m_policy.get(); }
+
+      /**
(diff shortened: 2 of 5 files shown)


randombit:
Oh nice, thanks.

I don't have time to review this right now but will do so asap, probably in next few days.

We should (I think) have root on the CI so it would be possible to test it there. However we probably should not ask for or require root access to run the tests normally, eg when installing through Homebrew. Nor should we mess with an end users certificate store ;)

To some extent that could be worked around by adding a flag to `botan-test` that signals "I'm running in continuous integration environment and it's ok to run potentially dangerous or destructive tests." and any test modifying the local certstore could check that flag before running.

randombit:
Have not had time for a full review but a few quick comments:

- Question: Is it possible to replace `scoped_CFType` with `unique_ptr` which is passed a release function? As done for example in https://github.com/randombit/botan/blob/master/src/lib/prov/openssl/openssl_ec.cpp#L141

- Use `BOTAN_ASSERT` or `BOTAN_ASSERT_EQUAL` instead of C stdlib `assert`. I think `assert_notnull` could be replaced by `BOTAN_ASSERT_NONNULL`.

- Question: is it possible this works on iOS as well?

reneme:
Regarding the `scoped_CFType`: I failed to fully replace it with `unique_ptr`. Those `CFTypeRef`-like types are eventually typedefs of `void*`, so they already _are_ a pointer themselves. I tried to replace it like so:

```C++
template <typename T>
struct CFType_deleter {
  void operator()(T cftype) {
    if (cftype != nullptr) {
      CFRelease(cftype);
    }
  }
};

template <typename T>
using scoped_CFType = std::unique_ptr<typename std::remove_pointer<T>::type, CFType_deleter<T>>;
```

This sort-of works, but doesn't allow to take the address of the wrapped `CFTypeRef`. That's needed when using it as output parameter of the `SecKeychainOpen()` call. One could of course add a short-lived bare `CFTypeRef` and wrap it after the call, but I didn't like that approach.

reneme:
I replaced the `assert()` calls with `BOTAN_ASSERT` and renamed the `assert_nonnull` and `assert_success` to `check_xxx`, as I intend to throw exceptions in those conditions.

reneme:
Regarding iOS: I would imagine that this is mostly going to work on iOS as well. I think, accessing the keychain itself is handled a bit differently there, though. In particular there is only _one_ keychain in the whole system, so the calls to `SecKeychainOpen()` can probably be omitted. A quick glance in the documentation of the used API calls suggests that it should work from iOS 10.6 onwards with minor adaptions. That said, I don't have a test setup readily available to try it.

reneme:
Finally, I found a somewhat major bummer that sounds pretty odd to me: In order to successfully find certificates using Apple's `SecItemCopyMatching()` call, I needed to "normalize" the DN objects. See [here](https://github.com/randombit/botan/pull/1830/commits/a27f1c2ef6efb7d90b2f2a44c5e08a13903d7407#diff-ce3705d4ff7e58a182e5ff3f3334b904R78) for the actual method.

Apple states in their [keychain implementation](https://opensource.apple.com/source/Security/Security-55471/sec/Security/SecCertificate.c.auto.html) that ASN.1 strings of the type 'PrintableString' are not case sensitive. Hence, they transform those strings into uppercase and furthermore strip leading and trailing whitespace. Also the certificate lookup seems to be rather slow: The current test body, takes nearly 500ms on my laptop, despite fetching just a handful certificates.

As an alternative of using Apple's APIs for certificate lookups, we might be better off to pull the entire system certificate store into the `Certificate_Store_In_Memory` implementation upon initialization. Initially, I rejected that idea as rather dirty. What do you think?

src/lib/x509/certstor_system_macos/certstor_macos.cpp
```diff
@@ -0,0 +1,472 @@
...
+
+   if(!result)
+      {
+      return {};  // no certificates found
+      }
+
+   const auto count = CFArrayGetCount(result.get());
+   BOTAN_ASSERT(count > 0, "certificate result list contains data");
+
+   std::vector<std::shared_ptr<const X509_Certificate>> output;
```
reneme: Pre-allocate buffer?
```suggestion
   std::vector<std::shared_ptr<const X509_Certificate>> output;
   output.reserve(count);
```

randombit:
> Apple states in their [keychain implementation](https://opensource.apple.com/source/Security/Security-55471/sec/Security/SecCertificate.c.auto.html) that ASN.1 strings of the type 'PrintableString' are not case sensitive. Hence, they transform those strings into uppercase and furthermore strip leading and trailing whitespace.

So, this follows from the requirements of RFC 3280 which states that when comparing DNs, "attribute values in PrintableString are not case sensitive" as well as the whitespace normalization. And in fact X.509 proper requires such for all string types, and we implement that when comparing DNs (function `x500_name_cmp` in `parsing.cpp`).

However Apple's behavior here is completely broken: how could "case insensitive" turn out to mean "you must take into account casing"?!? I think they were just too lazy to implement the correct DN comparison algorithm and did this as a hack.

> Also the certificate lookup seems to be rather slow: The current test body, takes nearly 500ms on my laptop, despite fetching just a handful certificates.

Wow that is quite slow. Maybe worth timing each test individually to help diagnose the issue? Call `start_timer`/`end_timer` on your `Test::Result` instances to enable this.

> As an alternative of using Apple's APIs for certificate lookups, we might be better off to pull the entire system certificate store into the `Certificate_Store_In_Memory` implementation upon initialization. Initially, I rejected that idea as rather dirty. What do you think?

Yeah that seems messy, for example consider a long running server process: you would probably expect that adding a cert to the system store would make it available in the server process without a process restart.

I wonder too if perhaps there is some startup overhead to using the system store, say 400ms on first use, but then all subsequent operations are fast. You might be able to detect this using `botan-test --test-runs=10 <test names>` which runs the named tests several times in a row.

src/lib/x509/certstor_system_macos/certstor_macos.cpp
randombit: It shouldn't matter much since you are just reserving the size for the `shared_ptr`s. Either way is fine as far as I am concerned.

reneme:
> I wonder too if perhaps there is some startup overhead to using the system store, say 400ms on first use, but then all subsequent operations are fast.

Thanks for the pointer with `start_timer`/`end_timer`. That was helpful. Turns out that the `.all_subjects()` test takes 95% of the full time, i.e. some 450ms. Which is unfortunate, but somewhat expected. It needs to pull in and parse _all_ trusted certificates, after all. Initialization of the `SecKeychain` object is in the ballpark of 1ms. So are individual certificate lookups.

> However Apple's behavior here is completely broken: how could "case insensitive" turn out to mean "you must take into account casing"?!?

Its annoying for sure. I'll have another look whether we can somehow do the "normalization" with some public Apple API, rather than re-implementing it. At least that would be less brittle, I guess.

> you would probably expect that adding a cert to the system store would make it available in the server process without a process restart.

That's exactly what I had in mind as well, when deciding against this option.

randombit:
> I'll have another look whether we can somehow do the "normalization" with some public Apple API, rather than re-implementing it. At least that would be less brittle, I guess.

Looks like https://developer.apple.com/documentation/security/1396030-seccertificatecopynormalizedsubj?language=objc is the intended API for this. And yeah, good to use the public API here since they might decide to apply the same normalization to other string types in the future.

src/lib/x509/certstor_system_macos/certstor_macos.cpp
```diff
@@ -0,0 +1,475 @@
...
+Certificate_Store_MacOS::find_cert_by_raw_subject_dn_sha256(const std::vector<uint8_t>& subject_hash) const
+   {
+   BOTAN_UNUSED(subject_hash);
+   throw Not_Implemented("Certificate_Store_MacOS::find_cert_by_raw_subject_dn_sha256");
+   }
+
+std::shared_ptr<const X509_CRL> Certificate_Store_MacOS::find_crl_for(const X509_Certificate& subject) const
+   {
+   BOTAN_UNUSED(subject);
+   throw Not_Implemented("Certificate_Store_MacOS::find_crl_for");
```
reneme: I'm currently testing this along with our under-construction TLS asio stream (see #1731) and it works quite nicely. However, for the certificate validation to work, I need to statically return an empty list of CRLs (or respectively don't override this function at all, hence falling back to the base class' method returning an empty list). As far as I can see, the macOS keychain does not support any CRL handling. Is "not overriding" a reasonable implementation strategy here?

randombit: Yes if the keychain does not support CRLs at all then it is fine to either omit implementing the function entirely, or (to make it obvious the omission is intentional) override it and have it return an empty list with a comment to the effect that it is not supported by the system API. `Not_Implemented` is more to indicate a TODO type situation, where the functionality is possible but the code is just not written yet.

src/lib/x509/certstor_system_macos/certstor_macos.h
```diff
@@ -0,0 +1,75 @@
...
+namespace Botan {
+
+class Certificate_Store_MacOS_Impl;
+
+/**
+* Certificate Store that is backed by the system trust store on macOS. This
+* opens a handle to the macOS keychain and serves certificate queries directly
+* from there.
+*/
+class BOTAN_PUBLIC_API(2, 0) Certificate_Store_MacOS final : public Certificate_Store
```
randombit: This should be 2,10 since that will be the version version that includes this.

src/tests/test_certstor_macos.cpp
```diff
@@ -0,0 +1,336 @@
...
+   {
+   Test::Result result("macOS Certificate Store - Open Keychain");
+
+   try
+      {
+      result.start_timer();
+      auto certstore = new Botan::Certificate_Store_MacOS();
+      result.end_timer();
+
+      delete certstore;
```
randombit: I know it is just a test but would be better to use RAII here - explicit `delete` is always trouble.

src/build-data/os/darwin.txt
```diff
@@ -21,6 +21,8 @@ commoncrypto
 sockets
 threads
 filesystem
+
+certstore
```
randombit: Having this feature named certstore is going to get confusing since Windows etc also has a certificate store. Maybe something like `apple_keychain`?

src/tests/test_certstor_macos.cpp
```diff
@@ -0,0 +1,336 @@
...
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include "tests.h"
+
+#include <botan/build.h>
+
+#include <iostream>
+
+#if defined(BOTAN_TARGET_OS_IS_DARWIN)
```
randombit: This should be checking for `BOTAN_HAS_CERTSTOR_MACOS` since it is possible to build on OSX but with the certstore disabled.

src/lib/x509/certstor_system_macos/certstor_macos.h
```diff
@@ -0,0 +1,75 @@
...
+   {
+   public:
+      Certificate_Store_MacOS();
+
```
randombit: `=default` copy/move constructors/assignments?

src/lib/x509/certstor_system_macos/certstor_macos.cpp
```diff
@@ -0,0 +1,475 @@
...
+* Certificate Store
+* (C) 1999-2019 Jack Lloyd
+* (C) 2019      René Meusel
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/build.h>
+
+#if defined(BOTAN_TARGET_OS_IS_DARWIN)
```
randombit: No need for this - if the file is being compiled we know (by virtue of restrictions set in `info.txt`) that we are on OSX

reneme:
> Looks like https://developer.apple.com/documentation/security/1396030-seccertificatecopynormalizedsubj?language=objc is the intended API for this.

Unfortunately this API is deprecated and also takes a `SecCertificateRef`. In [Security/SecCertificate.c](https://opensource.apple.com/source/Security/Security-55471/sec/Security/SecCertificate.c.auto.html) I found an internal API called `SecDistinguishedNameCopyNormalizedContent` (defined in [Security/SecCertificateInternal.h](https://opensource.apple.com/source/Security/Security-55471/sec/Security/SecCertificateInternal.h.auto.html)) which does pretty much what we need. It is not defined in the public header though, so I needed to define it in the compilation unit like so:

```C++
extern "C" CFDataRef SecDistinguishedNameCopyNormalizedContent(CFDataRef distinguished_name);
```

With a bit of further mangling (the call strips the enclosing ASN.1 `SEQUENCE` from the DN data), it actually does what it should. However, I obviously don't feel comfortable linking against internal Apple's internal API... :-/

src/lib/x509/certstor_system_macos/certstor_macos.cpp
```diff
@@ -0,0 +1,475 @@
...
+   const char* ccstr = CFStringGetCStringPtr(cfstring, kCFStringEncodingUTF8);
+
+   if(ccstr != NULL)
+      {
+      return std::string(ccstr);
+      }
+
+   auto chars = CFStringGetLength(cfstring);
+
+   // Worst case: 2 bytes per character + NUL
```
randombit: Is that really true? UTF-8 can be up to 4 bytes / character

src/lib/x509/certstor_system_macos/certstor_macos.cpp
```diff
@@ -0,0 +1,475 @@
...
+   auto cstrlen = chars * 2 + 1;
+   auto cstr = std::unique_ptr<char[]>(new char[cstrlen]);
```
randombit: would prefer `vector` to explicit `new`

src/tests/test_certstor_macos.cpp
```diff
@@ -0,0 +1,336 @@
+/*
+* (C) 1999-2019 Jack Lloyd
+* (C) 2019      René Meusel
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include "tests.h"
+
+#include <botan/build.h>
```
randombit: You can assume "tests.h" includes `botan/build.h` since most tests want to do feature macro checks.

randombit:
> However, I obviously don't feel comfortable linking against internal Apple's internal API...

Yeah that is no good. OK I guess we use the normalization you wrote and hope Apple doesn't change their minds about what is normalized or how.

src/build-data/os/darwin.txt
reneme: Judging from "filesystem", "threads" and so on, I assumed that this is a list of "features" a certain platform provides. I.e. if we'd add a similar implementation for Windows (which we are actually planning in the near future), we would just add "certstore" to `windows.txt` as well.
Is that a wrong assumption?

reneme:
Thanks for the review. I fixed the simple things right away and will do the rest tomorrow.

src/build-data/os/darwin.txt
randombit: Here feature is in terms of specific APIs, so for `threads` we mean that `std::thread` and company work, and `sockets` means BSD socket API (which is why Windows defines `winsock2` instead as the APIs differ). If say there was a platform which had threads but the `std::` thread API didn't work and we had to use some system specific API, it would get its own distinct feature and not define `threads`.

reneme: That's good to know, thanks.

src/lib/x509/certstor_system_macos/certstor_macos.cpp
reneme: According to the [documentation](https://developer.apple.com/documentation/corefoundation/1542853-cfstringgetlength?language=objc) `CFStringGetLength()` returns the number of UTF-16 code pairs. So, for example this Emoji "🧖" requires two UTF-16 code pairs (Big Endian: 0xD83E 0xDDD6) and four UTF-8 bytes (0xF0 0x9F 0xA7 0x96). For a string containing only this emoji `CFStringGetLength()` would return "2", hence the calculation would be correct.

Nevertheless, while reading this up, I found `CFStringGetMaximumSizeForEncoding()` which is probably the better choice anyway.

reneme:
I'm done with applying your comments.

randombit:
There are a few warnings:

```
src/lib/x509/certstor_system_macos/certstor_macos.cpp:245:49: warning: suggest braces around initialization of subobject [-Wmissing-braces]
         std::array<const void*, 2> keychains({ m_system_roots.get(),
                                                ^~~~~~~~~~~~~~~~~~~~~
                                                {
src/lib/x509/certstor_system_macos/certstor_macos.cpp:463:95: warning: unused parameter 'subject' [-Wunused-parameter]
std::shared_ptr<const X509_CRL> Certificate_Store_MacOS::find_crl_for(const X509_Certificate& subject) const
                                                                                              ^
2 warnings generated.
src/tests/test_certstor_macos.cpp:81:44: warning: empty parentheses interpreted as a function declaration [-Wvexing-parse]
      Botan::Certificate_Store_MacOS unused();
                                           ^~
src/tests/test_certstor_macos.cpp:81:44: note: remove parentheses to declare a variable
      Botan::Certificate_Store_MacOS unused();
                                           ^~
src/tests/test_certstor_macos.cpp:81:38: warning: unused function 'unused' [-Wunused-function]
      Botan::Certificate_Store_MacOS unused();
                                     ^
2 warnings generated.
```

Can you fix these and then squash all the commits?

Also, I realized now why all_subjects is slow. In addition to accessing every cert, it has to actually load/parse the certificates to work around Apple's DN mangling. This triggers a signature check on each CA certificate (this is necessary to test self-signed vs self-issued certificates). Likely if that logic were made to be lazy (computed just when/if X509_Certificate::is_self_signed is called) then the operation would be much faster.

reneme:
> Can you fix these and then squash all the commits?

Done. ... waiting for CI to see if all the warnings have vanished.

> Also, I realized now why all_subjects is slow. ...

I consider this outside the scope of that PR. Another option would be to let `all_subjects()` throw `Not_Implemented` for the time being. Not sure whether this list is of much value anyway.

randombit:
> I consider this outside the scope of that PR.

Oh absolutely - I meant that I will look at addressing that later, so at some point in the future this slow operation will become fast.

reneme:
> waiting for CI to see if all the warnings have vanished.

Looks like the code doesn't produce warnings related to this PR on the macOS build anymore. :)

randombit:
Merged! Thanks again.

PR #1832, opened 2019-02-15
Renamed m_handle to m_prov as this is the variable name used elsewhere.

rasmartins:
The code in the class System_RNG_Impl uses m_prov but the variable was called m_handle.


randombit:
Thanks - Windows Phone isn't part of our CI build right now so this was missed.

PR #1834, opened 2019-02-18
Inline CPUID::state

randombit:
This has to be in a function for MSVC but it seems to work even if the function is inlined.


randombit:
This doesn't work for reasons I don't understand. MSVC is ok but GCC in certain builds fails.

PR #1835, opened 2019-02-18
Update docs to replace from_configuration.

kelvinhammond:
PasswordHash::from_configuration has bee replace with with PasswordHash::from_params.


randombit:
Thank you

PR #1839, opened 2019-02-21
Botan::TLS::Stream

hrantzsch:
This PR introduces a `boost::asio`-based `Botan::TLS::Stream`.
The stream strives to be a drop-in replacement for `boost::asio::ssl` that is based on Botan, rather than openSSL.
The provided interface is not _yet_ fully compatible though.

In the following I will provide a quick fly-over through the code and point out some of the design decisions and open questions.

## `StreamBase` and `StreamCore`

The stream is based on `Botan::TLS::Channel` subclasses.
`StreamBase` is a templated class and can be specialized for the individual channels.
Currently, only `Botan::TLS::Client` is implemented.
Implementing `Server` should be straight forward, but we neither use nor test it at the moment, so we decided not to provide this specialization.

`StreamBase` instantiates the respective `Channel` using `Botan::TLS::StreamCore` as `Botan::TLS::Callbacks` instance.

## `Stream`

`Botan::TLS::Stream`'s main responsibility lies in implementing `read_some` and `write_some` to be used by the respective `boost::asio` methods.
It also implements `handshake` and `shutdown`, as well as a number of compatibility methods to comply to the interface of `boost::asio::ssl`.
Some of these compatibility methods are not implemented and will throw an exception accordingly when called.

As mentioned above, the stream can currently only be used for the client side.
In contrast to `asio::ssl`, our stream decides at the point of instantiation what the type of its underlying channel is.
Thus, `handshake(handshake_type t)` methods from the compatibility interface will only check if the provided type matches the channel and throw an exception otherwise.

## Exceptions

Boost usually relies on communicating errors via `boost::system::error_code`, while Botan usually throws an exception when an error occurs.
In order to provide a boost-compatible interface most calls to the `Botan::TLS::Channel` instance are wrapped in a `try` block and exceptions are converted to appropriate error codes (see `asio_convert_exceptions.h`).

## Tests

The unit tests mock the `Botan::TLS::Channel` instance by providing a `StreamBase<MockChannel>` specialization and the network socket by using `boost::beast::test::stream`.
This means all entities "behind" the interface of the stream (like `StreamCore` and async operation classes) are tested implicitly through the stream.

# State of the PR

We are opening this PR as a work-in-progress.
Currently, we are battle-testing the stream in our real-world application and we expect to have some fixes and improvements in the near future.
Nonetheless, we'd be very grateful for some feedback for the current code.


randombit:
Great! I will review this asap, probably over the weekend.

src/lib/tls/asio/asio_stream.h
```diff
@@ -0,0 +1,487 @@
...
+      async_handshake(HandshakeHandler&& handler)
+         {
+         BOOST_ASIO_HANDSHAKE_HANDLER_CHECK(HandshakeHandler, handler) type_check;
+
+         boost::asio::async_completion<HandshakeHandler,
+               void(boost::system::error_code)>
+               init(handler);
+
+         auto op = create_async_handshake_op(std::move(init.completion_handler));
+         op(boost::system::error_code{}, 0, 1);
```
vinniefalco: The system of launching composed operations is different now (and a lot better). See: https://github.com/boostorg/asio/commit/0568d3bf0b492887c6c48d05f5492a782dafa746

Beast supports all these as of v217

vinniefalco: You could probably write `op({}, 0, 1);` instead

src/lib/tls/asio/asio_stream.h
```diff
@@ -0,0 +1,487 @@
...
+         this->m_core.consumeSendBuffer(writtenBytes);
+         return writtenBytes;
+         }
+
+      template <typename Handler>
+      Botan::TLS::AsyncHandshakeOperation<Channel, StreamLayer, Handler>
+      create_async_handshake_op(Handler&& handler)
+         {
+         return Botan::TLS::AsyncHandshakeOperation<Channel, StreamLayer, Handler>(
+                   native_handle(), this->m_core, m_nextLayer, std::forward<Handler>(handler));
```
vinniefalco: You need to decay the handler when you pass it as a template parameter, e.g. `typename std::decay<Handler>::type`

src/lib/tls/asio/asio_stream.h
```diff
@@ -0,0 +1,487 @@
...
+      template <typename Handler, typename MutableBufferSequence>
+      Botan::TLS::AsyncReadOperation<Channel, StreamLayer, Handler,
+            MutableBufferSequence>
+            create_async_read_op(Handler&& handler,
+                                 const MutableBufferSequence& buffers)
+         {
+         return Botan::TLS::AsyncReadOperation<Channel, StreamLayer, Handler,
+                MutableBufferSequence>(
+                   native_handle(), this->m_core, m_nextLayer, std::forward<Handler>(handler),
+                   buffers);
```
vinniefalco: These helpers are kind of pointless to be honest, I would just inline them at the call site. Less code and easier to read.

src/lib/tls/asio/asio_stream.h
```diff
@@ -0,0 +1,487 @@
...
+      template <typename Handler>
+      Botan::TLS::AsyncWriteOperation<Handler>
+      create_async_write_op(Handler&& handler, std::size_t plainBytesTransferred)
+         {
+         return Botan::TLS::AsyncWriteOperation<Handler>(
+                   this->m_core, std::forward<Handler>(handler), plainBytesTransferred);
+         }
+
+   protected:
+      StreamLayer m_nextLayer;
```
vinniefalco: `protected` member in a `final` class? Hmm...

src/lib/tls/asio/asio_stream.h
```diff
@@ -0,0 +1,487 @@
...
+                MutableBufferSequence>(
+                   native_handle(), this->m_core, m_nextLayer, std::forward<Handler>(handler),
+                   buffers);
+         }
+
+      template <typename Handler>
+      Botan::TLS::AsyncWriteOperation<Handler>
+      create_async_write_op(Handler&& handler, std::size_t plainBytesTransferred)
+         {
+         return Botan::TLS::AsyncWriteOperation<Handler>(
```
vinniefalco: All these handlers in template parameter lists need to be decayed

src/lib/tls/asio/asio_stream_base.h
```diff
@@ -0,0 +1,95 @@
...
+            throw Invalid_Argument("wrong handshake_type");
+            }
+         }
+
+      //! \brief validate the OpenSSL compatibility enum `handshake_type`
+      bool validate_handshake_type(handshake_type type, boost::system::error_code& ec)
+         {
+         if(type != handshake_type::client)
+            {
+            ec = make_error_code(Botan::TLS::error::invalid_argument);
```
vinniefalco: This should work:
```
ec = Botan::TLS::error::invalid_argument;
```

src/lib/tls/asio/asio_stream_core.h
```diff
@@ -0,0 +1,120 @@
...
+           input_buffer(boost::asio::buffer(m_input_buffer_space)) {}
+
+      virtual ~StreamCore() = default;
+
+      void tls_emit_data(const uint8_t data[], size_t size) override
+         {
+         auto buffer = m_send_buffer.dynamicBuffer.prepare(size);
+         auto copySize =
+            boost::asio::buffer_copy(buffer, boost::asio::buffer(data, size));
+         m_send_buffer.dynamicBuffer.commit(copySize);
```
vinniefalco: Or you could just write in one statement:
```
m_send_buffer.dynamicBuffer.commit(net::buffer_copy(
    m_send_buffer.dynamicBuffer.prepare(size), net::buffer(data, size)));
```

src/lib/tls/asio/asio_stream_core.h
```diff
@@ -0,0 +1,120 @@
...
+         {
+         Buffer() : dynamicBuffer(data_buffer) {}
+         std::vector<uint8_t> data_buffer;
+         boost::asio::dynamic_vector_buffer<
+         uint8_t, typename decltype(data_buffer)::allocator_type>
+         dynamicBuffer;
+         };
+
+      StreamCore()
+         : m_input_buffer_space(17 * 1024, '\0'), // enough for a TLS Datagram
```
vinniefalco: 17KB always? That's a lot...who am I to judge though.

src/tests/unit_asio_stream.cpp
```diff
@@ -0,0 +1,560 @@
...
+constexpr std::size_t TEST_DATA_SIZE = 142;
+static_assert(sizeof(TEST_DATA) == TEST_DATA_SIZE, "size of TEST_DATA must match TEST_DATA_SIZE");
+
+// use memcmp to check if the data in a is a prefix of the data in b
+bool contains(const void* a, const void* b, const std::size_t size) { return memcmp(a, b, size) == 0; }
+
+/**
+ * Mocked Botan::TLS::Channel. Pretends to perform TLS operations and triggers appropriate callbacks in StreamCore.
+ */
+class MockChannel
```
vinniefalco: If you have access to beast (or if you can just copy the code) you might want to try this stream as your next layer:
https://github.com/boostorg/beast/blob/develop/include/boost/beast/_experimental/test/stream.hpp#L37

It is quite solid, it can simulate failures (great for testing all branches and achieving high coverage), and you can "stuff" the stream with bytes. The Beast tests make heavy use of it. In this code, you can see how the test stream is stuffed with an HTTP request, and the call to `http::read` is run in a "failure loop" which fails at the `n` I/O operation this is how Beast gets close to 100% coverage:
https://github.com/boostorg/beast/blob/develop/test/beast/http/read.cpp#L277

src/lib/tls/asio/asio_async_handshake_op.h
```diff
@@ -0,0 +1,100 @@
...
+
+namespace Botan {
+
+namespace TLS {
+
+template <class Channel, class StreamLayer, class Handler>
+struct AsyncHandshakeOperation
+   {
+      AsyncHandshakeOperation(Channel* channel, StreamCore& core,
+                              StreamLayer& nextLayer, Handler&& handler)
```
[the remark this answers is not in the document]
Tymolc: The type of  the handler argument should stay `Handler&&` and not `Handler_&&`?
If I understand correctly, the `Handler_` template should be deduced by the type of `handler`, right? How can the compiler deduce this type in your example, though?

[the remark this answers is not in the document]
vinniefalco: Well obviously I made a mistake.. lol... I will edit my reply.

src/lib/tls/asio/asio_stream_core.h
randombit: TLS maximum record size is 16K plus some overhead for headers and padding.

randombit:
> In order to provide a boost-compatible interface most calls to the `Botan::TLS::Channel` instance are wrapped in a `try` block and exceptions are converted to appropriate error codes (see `asio_convert_exceptions.h`).

I think some of this conversion logic may be duplicating code recently added wrt `Exception::error_type()` and the `ErrorType` enum.

It may be worthwhile offering APIs returning errors natively (ie directly on `TLS::Channel`), I will think about this.

src/lib/tls/asio/asio_stream_core.h
vinniefalco: When you say a "TLS Record" is that this thing:
https://github.com/vinniefalco/beast/blob/59347b3928a56b08b274221bb649e0fb60c229f4/include/boost/beast/core/detect_ssl.hpp#L108

src/lib/tls/asio/asio_async_handshake_op.h
```diff
@@ -0,0 +1,100 @@
...
+            {
+            auto read_buffer =
+               boost::asio::buffer(m_core.input_buffer, bytesTransferred);
+            try
+               {
+               m_channel->received_data(
+                  static_cast<const uint8_t*>(read_buffer.data()),
+                  read_buffer.size());
+               }
+            catch(...)
```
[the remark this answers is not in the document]
hrantzsch: I agree. @randombit, can I assume that `TLS::Channel` and its subclasses will throw instances of `Botan::Exception`? I noticed that `(...)` is caught and re-thrown in `Channel::received_data`.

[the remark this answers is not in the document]
randombit: The only time we intentionally throw something other than `Botan::Exception` (or a subclass thereof) is if `calloc` returns `nullptr` in which case `std::bad_alloc` is thrown. It is possible though for a `std::exception` to be raised due to `std::vector::at` with a bad index or the like.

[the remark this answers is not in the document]
vinniefalco: Suggested:
```
try
{
    ...
}
catch(Botan::Execption const&)
{
    ...
}
catch(std::exception const&)
{
    ...
}
```

src/tests/unit_asio_stream.cpp
```diff
@@ -0,0 +1,560 @@
+#include "tests.h"
+
+#include <iostream>
```
randombit: Is this include needed?

src/tests/unit_asio_stream.cpp
```diff
@@ -0,0 +1,560 @@
+#include "tests.h"
```
randombit: There should be a copyright/license declaration at the top if this file as well.

src/lib/tls/asio/info.txt
```diff
@@ -0,0 +1,23 @@
+<defines>
+ASIO -> 20181218
```
randombit: This should be something like `TLS_ASIO_STREAM` to distinguish it from other uses of asio

randombit:
@hrantzsch Sorry, I still owe you a proper review on this, work has been very busy for me.
But a quick question - what is the reason for requiring Boost 1.66.0? It looks like Beast was added in that version, but does this integration (the stuff in src/lib) actually require Beast? Or (as was my understanding) just asio, and Beast is used in the tests?

For testing, I think the best thing to do is install Beast from git in the coverage build - `src/scripts/ci/setup_travis.sh` so we can run the tests there and get test coverage numbers. If need be we can disable use of Boost in the Mac build to simplify things.

Tymolc:
io_context was also introduced in boost 1.66.0 and deprecated the old io_service. We could try using io_service and check which boost version we can use then.
What should be our target boost version? Ubuntu 18.04 has boost 1.65 in apt. Would that be ok?

randombit:
Oh I see. On the one hand it would be nice to support Ubuntu 18.04 since that's pretty common. But it appears to be an outlier here - Fedora 28+, Alpine, Arch, Debian testing, and Ubuntu 18.10 all have at least Boost 1.66. And, supporting a build with Boost 1.65 means we have a configuration that is going to be subtly different from the normally tested one (due to asio differences) *and* it won't have tests (since Beast will be missing). So it is probably better overall to just assume and require 1.66 or higher for this component.

For CI, which is stuck on 16.04, we need to figure out how to get a modern Boost included. With luck it is possible to install the Boost 1.67 packages from Cosmic onto Xenial. I will look into this in the next day or two.

vinniefalco:
if I recall, Boost 1.65 asio has a defect, so you would have to ensure 1.64 (need to double check this)

Tymolc:
We created a small test project based on Boost.Beast, using our Stream: https://gist.github.com/Tymolc/dee30dd4e9453a1478dd52ef94cf0643

src/lib/tls/asio/asio_stream.h
```diff
@@ -396,13 +396,15 @@ class Stream : public StreamBase<Channel>
...
                {
-               const std::size_t to_send =
-                  std::min<std::size_t>(MAX_PLAINTEXT_SIZE - sent, boost::asio::buffer_size(*it));
-               native_handle()->send(static_cast<const uint8_t*>(it->data()), to_send);
-               sent += to_send;
+               if(sent >= MAX_PLAINTEXT_SIZE) return;
+               boost::asio::const_buffer buffer = *it;
+               const auto amount =
+                  std::min<std::size_t>(MAX_PLAINTEXT_SIZE - sent, buffer.size());
+               native_handle()->send(static_cast<const uint8_t*>(buffer.data()), amount);
```
vinniefalco: unfortunate that `send` wants a `uint8_t*`

randombit:
@hrantzsch What's the state of this PR? Is it ready for another review or do you have more work planned?

hrantzsch:
We have successfully integrated the stream in a fork of the cpprestsdk library and fixed some last issues. Thus, the PR is ready for review.
We will continue to battle-test it with our application, but I will remove the WIP label. Oh, and we will add some documentation, especially for the async operations.

src/lib/tls/asio/asio_stream_core.h
```diff
@@ -0,0 +1,120 @@
...
+   {
+      StreamCore()
+         : m_input_buffer_space(17 * 1024, '\0'), // enough for a TLS Datagram
+           input_buffer(m_input_buffer_space.data(), m_input_buffer_space.size()) {}
+
+      virtual ~StreamCore() = default;
+
+      void tls_emit_data(const uint8_t data[], std::size_t size) override
+         {
+         m_send_buffer.commit(
```
reneme: as far as I understand `tls_emit_data()` is (naïvely) meant to write the data onto a socket of some sort. It makes sense that we just prepare the send buffer here, but probably we should explain ourselves with a short comment. Especially, that this implementation assumes some other entity to actually go ahead and perform the actual "send".

src/lib/tls/asio/asio_stream.h
```diff
@@ -0,0 +1,585 @@
...
+            }
+         }
+
+      template <typename ConstBufferSequence>
+      std::size_t tls_encrypt_some(const ConstBufferSequence& buffers,
+                                   boost::system::error_code& ec)
+         {
+         std::size_t sent = 0;
+         // NOTE: This is not asynchronous: it encrypts the data synchronously.
+         // Only writing to the socket is asynchronous.
```
reneme: This comment is misleading when trying to understand the synchronous `write_some()` function. 

```suggestion
         // The data encrypted by native_handle()->send() is synchronously stored in the send_buffer of m_core,
         // but is not actually written to the wire, yet.
```

src/lib/tls/asio/asio_stream.h
```diff
@@ -0,0 +1,585 @@
...
+         op{std::move(init.completion_handler), *this, this->m_core, buffers};
+         return init.result.get();
+         }
+
+   protected:
+      size_t writePendingTlsData(boost::system::error_code& ec)
+         {
+         auto writtenBytes = boost::asio::write(m_nextLayer, this->m_core.sendBuffer(), ec);
+
+         this->m_core.consumeSendBuffer(writtenBytes);
```
reneme: For my understanding of the buffer handling in the `StreamCore`. Imagine the following:

* application calls the synchronous `write_some()` several times with the biggest possible buffer
* `tls_encrypt_some` encrypts the data (probably causing some overhead) and stores it in the send buffer
* `asio::write()` pushes "some" data out through the socket, but leaves some data behind in the send buffer
* since `write_some()` calls `writePendingTlsData()` only once after `tls_encrypt_data`
  * shouldn't we see some data build-up in the send buffer that could eventually cause a dead lock or data not being sent as quickly as possible or being dropped?

Is this taken care of? and how?

src/lib/tls/asio/asio_stream.h
```diff
@@ -0,0 +1,585 @@
...
+      BOOST_ASIO_INITFN_RESULT_TYPE(WriteHandler,
+                                    void(boost::system::error_code, std::size_t))
+      async_write_some(const ConstBufferSequence& buffers, WriteHandler&& handler)
+         {
+         BOOST_ASIO_WRITE_HANDLER_CHECK(WriteHandler, handler) type_check;
+
+         boost::asio::async_completion<WriteHandler, void(boost::system::error_code, std::size_t)> init(handler);
+
+         boost::system::error_code ec;
+         std::size_t sent = tls_encrypt_some(buffers, ec);
```
reneme: Does it make sense to move the encryption into the asynchronous operation as well? As far as I can see, `tls_encrypt_data` just manages the `buffers` and then calls `native_handle()->send()`. I think this should be doable in the `AsyncWriteOperation` as well. However, at the expense of duplicating the logic in `tls_encrypt_data` to a certain extend, no?

My main argument for this is symmetry though. The `AsyncReadOperation` takes care of the deciphering internally.

src/lib/tls/asio/asio_stream.h
```diff
@@ -0,0 +1,585 @@
...
+         this->m_core.consumeSendBuffer(writtenBytes);
+         return writtenBytes;
+         }
+
+      void tls_decrypt_some(boost::system::error_code& ec)
+         {
+         boost::asio::const_buffer read_buffer =
+            {
+            this->m_core.input_buffer.data(),
+            m_nextLayer.read_some(this->m_core.input_buffer, ec)
```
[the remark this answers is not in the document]
hrantzsch: Correct.

src/lib/tls/asio/asio_stream.h
```diff
@@ -0,0 +1,583 @@
...
+               { return; }
+
+            try
+               {
+               native_handle()->received_data(static_cast<const uint8_t*>(read_buffer.data()),
+                                              read_buffer.size());
+               }
+            catch(const std::exception& ex)
+               {
+               ec = Botan::TLS::convertException();
```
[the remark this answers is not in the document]
hrantzsch: Ok, we now catch both `Botan::TLS::TLS_Exception` and `Botan::Exception` wherever we used `convertException` before. The error_code this creates will either be from the `BotanAlertCategory` or the `BotanErrorCategory` and will have values from `Botan::TLS::Alert` or `Botan::ErrorType`, respectively. I hope that's the correct way of handling this.

We are still also catching `std::exception` because our code should not throw if the user expects it to set an error_code. It introduces a lot of try-catch blocks. Should we re-introduce `convertException` as a helper function that only does the throwing-and-catching again, or don't you mind the code duplication?

src/lib/tls/asio/asio_error.h
```diff
@@ -0,0 +1,176 @@
...
+   invalid_oid,
+   stream_io_error,
+   self_test_failure,
+   not_implemented,
+   unknown
+   };
+
+namespace detail {
+// TLS Alerts
+struct BotanAlertCategory : boost::system::error_category
```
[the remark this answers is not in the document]
hrantzsch: That's not easy, unfortunately. `error_code` creation relies on providing overloads for `make_error_code` and we cannot define both `std::error_code make_error_code(...)` and `boost::error:code make_error_code(...)` (from the same enum class type). Hence, if we changed all this code to work on `std::error_code`, we'd no longer be able to automatically create `boost::error_code` from our enums.

We need to create `boost::error_code` because asio relies on them in its read and write methods. Of course we could have the internal machinery to automatically create `std::error_code` and then cast to `boost::error_code` wherever necessary. Just makes the code duplication discussed above larger, but maybe we can come up with a solution for this as well.

src/lib/tls/asio/info.txt
```diff
@@ -0,0 +1,25 @@
...
+TLS_ASIO_STREAM -> 20181218
+</defines>
+
+<header:public>
+asio_context.h
+asio_stream.h
+asio_error.h
+</header:public>
+
+<header:internal>
```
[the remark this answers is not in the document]
hrantzsch: We have started to restructure the code to address this. On that occasion we were also able to move the interface closer to boost::asio::ssl::stream (see commit message baa0725) and hide the detail whether the stream is a client or a server from the using code, much like Vinnie suggested above.
We're not quite done cleaning up the headers and will continue tomorrow.

src/lib/tls/asio/asio_stream.h
```diff
@@ -0,0 +1,583 @@
...
+#include <boost/version.hpp>
+#if BOOST_VERSION >= 106600
+
+#include <botan/internal/asio_async_handshake_op.h>
+#include <botan/internal/asio_async_read_op.h>
+#include <botan/internal/asio_async_write_op.h>
+#include <botan/internal/asio_convert_exceptions.h>
+#include <botan/internal/asio_includes.h>
+#include <botan/internal/asio_stream_base.h>
+#include <botan/internal/asio_stream_core.h>
```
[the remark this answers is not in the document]
hrantzsch: "resolving" to keep discussion of this issue in the comment in info.txt

src/tests/unit_asio_stream.cpp
```diff
@@ -0,0 +1,771 @@
...
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include "tests.h"
+
+#if defined(BOTAN_HAS_TLS) && defined(BOTAN_HAS_BOOST_ASIO)
+
+#include <boost/version.hpp>
+#if BOOST_VERSION >= 106600
```
hrantzsch: Unfortunately this is not sufficient for 'boost/beast/experimental/test/stream.hpp'. We need boost 1.68 for this. Either we stop using the test stream, or we require 1.68 for this unit test (the rest of the stream works fine with 1.66).

src/lib/tls/asio/asio_error.h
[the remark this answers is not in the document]
randombit: OK thanks for explanation, in that case let's just stick with the Boost versions for now.

src/lib/tls/asio/asio_stream.h
```diff
@@ -0,0 +1,708 @@
...
+            void tls_record_received(uint64_t, const uint8_t data[], std::size_t size) override
+               {
+               m_receive_buffer.commit(
+                  boost::asio::buffer_copy(m_receive_buffer.prepare(size), boost::asio::const_buffer(data, size))
+               );
+               }
+
+            void tls_alert(Botan::TLS::Alert alert) override
+               {
+               if(alert.type() == Botan::TLS::Alert::CLOSE_NOTIFY)
```
Tymolc: According to https://tools.ietf.org/html/rfc6455#section-7.1.7 a fatal error should also close the socket connection, so we can treat those the same

Tymolc:
```suggestion
               if(alert.type() == Botan::TLS::Alert::CLOSE_NOTIFY || alert.is_fatal())
```

This would be possible. We do still need to close the socket, though.

Tymolc: (In reference to https://github.com/randombit/botan/pull/1839#discussion_r276399517)

src/tests/unit_asio_stream.cpp
hrantzsch: I did as suggested above.

src/lib/tls/asio/info.txt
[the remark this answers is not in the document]
hrantzsch: I think we were able to make the interface much clearer now. We have four headers left:

- `asio_stream.h`
- `asio_context.h`, which corresponds to `asio::ssl::context` and is used to configure the stream
- `asio_error.h`, only relevant to a user in that it defines the error categories we set for error codes we create
- `asio_async_ops`, not user relevant, but a fair amount of code; placed in detail namespace now

src/scripts/ci/setup_travis.sh
```diff
@@ -74,5 +74,5 @@ if [ "$TRAVIS_OS_NAME" = "linux" ]; then
     fi
 
 elif [ "$TRAVIS_OS_NAME" = "osx" ]; then
-    HOMEBREW_NO_AUTO_UPDATE=1 brew install ccache
+    HOMEBREW_NO_AUTO_UPDATE=1 brew install ccache boost
```
reneme: i believe this proved unnecessary, didn't it?

src/lib/tls/asio/asio_stream.h
```diff
@@ -0,0 +1,708 @@
...
+      void set_verify_mode(verify_mode v,
+                           boost::system::error_code& ec)
+         {
+         BOTAN_UNUSED(v);
+         ec = Botan::ErrorType::NotImplemented;
+         }
+
+      //
+      // -- -- accessor methods for send and receive buffers
+      //
```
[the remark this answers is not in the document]
hrantzsch: Either that, or we friend the async operations. Strictly speaking, they should be an implementation detail of the stream anyways. I'll reconsider.

src/lib/tls/asio/asio_stream.h
```diff
@@ -0,0 +1,708 @@
...
+       * The function call will block until handshaking is complete or an error occurs.
+       *
+       * @param type The type of handshaking to be performed, i.e. as a client or as a server.
+       * @param ec Set to indicate what error occurred, if any.
+       */
+      void handshake(Connection_Side side, boost::system::error_code& ec)
+         {
+         setup_native_handle(side, ec);
+         if(ec)
+            { return; }
```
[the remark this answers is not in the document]
hrantzsch: yes

src/lib/tls/asio/asio_stream.h
```diff
@@ -0,0 +1,708 @@
...
+         BOTAN_UNUSED(v);
+         ec = Botan::ErrorType::NotImplemented;
+         }
+
+      //
+      // -- -- accessor methods for send and receive buffers
+      //
+
+      const boost::asio::mutable_buffer& input_buffer() { return m_input_buffer; }
+      boost::asio::const_buffer sendBuffer() const { return m_send_buffer.data(); }  // TODO: really .data() ?
```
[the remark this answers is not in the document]
hrantzsch: not available before boost 1.70 :/

src/lib/tls/asio/asio_stream.h
```diff
@@ -0,0 +1,708 @@
...
+               ec = e.error_type();
+               return;
+               }
+            catch(const std::exception&)
+               {
+               ec = Botan::ErrorType::Unknown;
+               return;
+               }
+
+            sendPendingEncryptedData(ec);
```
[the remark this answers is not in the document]
hrantzsch: I moved the first send out of the loop and left a note what it does there. i think thank clarifies it.

src/scripts/ci/setup_travis.sh
hrantzsch: true. un-cherry-picked.

randombit:
> ```
> PKCS11 RSA test 5 ran 1 tests 1 FAILED
> Failure 1: PKCS11 error 5
> ```
> 
> Is that flaky?

Yes this test occasionally fails (in CI only, have never been able to repro on my machines) and I have not been able to determine the cause. Err 5 in PKCS11 is "general error" :/

src/tests/unit_asio_stream.cpp
```diff
@@ -0,0 +1,772 @@
...
+         test_async_write_some_success(results);
+         test_async_write_some_buffer_sequence(results);
+         test_async_write_some_error(results);
+         test_async_write_throw(results);
+
+         return results;
+         }
+   };
+
+BOTAN_REGISTER_TEST("asio_stream", Asio_Stream_Tests);
```
randombit: Test should be named "tls_asio_stream" so it gets grouped with the rest of the TLS tests.

(I need to add a proper grouping mechanism to the tests but for now common prefixes suffice well enough)

src/tests/unit_asio_stream.cpp
```diff
@@ -0,0 +1,772 @@
...
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include "tests.h"
+
+#if defined(BOTAN_HAS_TLS) && defined(BOTAN_HAS_BOOST_ASIO)
+
+// first boost version to include boost/beast/experimental/test/stream.hpp
+#include <boost/version.hpp>
+#if BOOST_VERSION >= 106800
```
randombit: As I understand it asio_stream.h itself is ok starting with Boost 1.66. In that case can we here include asio_stream.h if Boost >= 1.66 but then only compile the tests if Boost >= 1.68. Which at least allows us to check that the header parses ok with the current compiler.

src/tests/unit_asio_stream.cpp
```diff
@@ -0,0 +1,772 @@
...
+ * The stream's channel, i.e. TLS_Client or TLS_Server, is mocked and pretends to perform TLS operations (noop) and
+ * provides the test data to the stream.
+ * The underlying network socket, claiming it read / wrote a number of bytes.
+ */
+class Asio_Stream_Tests final : public Test
+   {
+      // use memcmp to check if the data in a is a prefix of the data in b
+      bool contains(const void* a, const void* b, const std::size_t size) { return memcmp(a, b, size) == 0; }
+
+      boost::string_view test_data() const { return boost::string_view((const char*)TEST_DATA, TEST_DATA_SIZE); }
```
randombit: Old style cast

src/tests/unit_asio_stream.cpp
hrantzsch: That makes sense. I'll just move the check for 1.68 below the include, as the stream itself checks for 1.66 already.

hrantzsch:
Ok, we are done with the final polishing works.

It turns out `asio::ssl::stream` doesn't close the socket, so it should be fine if we don't act on `ALERT::CLOSE_NOTIFY`.

With this, the PR can be merged from our behalf. In the next two weeks we will need to implement `set_verify_callback`, overriding `tls_verify_cert_chain` in the StreamCore, in order to allow our application to do it's own OCSP validation.
For the sake of not bloating this PR even further, we would like to address this in a future PR.

randombit:
Sounds good. I will try to do final pass review tonight and hopefully merge. No problem adding OCSP in a future PR. Thank you @hrantzsch @Tymolc @reneme  for all the work on this.

randombit:
Merged!

hrantzsch:
Great, thank you! Thank you also to @vinniefalco for a lot of help in getting the ASIO specifics right.

pedro-vicente:
Ideally, the dependency should be on the non-Boost ASIO 

https://think-async.com/Asio/

PR #1840, opened 2019-02-21
macOS fix for tls_server in UDP/DTLS mode

witaly-iwanow:
tls_server doesn't work on my MBP/macOS 10.14.3 in UDP mode:
./botan tls_server test.crt test.key --port=5555 --type=udp
Listening for new connections on udp port 5555
Error: Could not connect UDP socket

Apparently it happens due to specific recvfrom() behavior on macOS when it's passed a nullptr/zero size buffer. So here's my workaround


randombit:
Thanks

PR #1842, opened 2019-02-22
Roughtime client

### LINKED ISSUE #1825: WIP: Roughtime client (https://roughtime.googlesource.com/roughtime/)
I plan to start working this week in roughtime client.

If anyone is already on it please let me know to avoid duplication.

Thanks!

nunojpg:
This pull request implements a Roughtime client API and CLI.

Roughtime is a authenticated time protocol developed by [Google](https://roughtime.googlesource.com/roughtime).

The protocol is under [standardization](https://tools.ietf.org/html/draft-roughtime-aanchal-03), but there initial version is available one multiple public servers, including from Google and Cloulflare.

This PR implements this initial version since it's the one actually being used on the wild and the standard one is not finished yet.

- [x] UDP client
- [x] Roughtime client
- [x] CLI
- [x] Complete test suite


nunojpg:
@randombit Pull request should be feature complete but I've not yet implemented a non-asio socket layer, and tests are incomplete.

I ask for a API review at this time, since I never implemented any feature of this size for Botan, so a lot of mistakes are expected.

- I added a URI class to parse strings in the format 10.98.31.21:3233, [2001:abcd::0001]:3232, and botan.randombit.net:3232., I believe this can be of general use.
- Current implementations use the "roughtime-chain.json" format for accountability. I've implemented this and are currently depending on nlohmann/json library. Please let me know how you would like to deal with this. Either if this dependency is ok, if a in-tree json implementation is preferable, or if this feature should be removed.
- I've used std::array for fixed length interfaces. This requires some ugly casts, to be compatible with current API. It should not look ugly anymore after std::span becomes used.

src/lib/misc/roughtime/roughtime.h
```diff
@@ -0,0 +1,119 @@
...
+*/
+BOTAN_PUBLIC_API(2,10)
+std::vector<uint8_t> online_request(const std::string& uri,
+                        const std::array<uint8_t, 64>& nonce,
+                        std::chrono::milliseconds timeout = std::chrono::seconds(3));
+
+
+struct BOTAN_PUBLIC_API(2,10) Server final
+  {
+    std::string name;
```
randombit: Avoid public members: instead use `class`, a constructor, and getters returning `const &` for the various members. This adds boilerplate but prevents modifications of a `Server` once deserialized.

src/lib/misc/roughtime/roughtime.h
```diff
@@ -0,0 +1,119 @@
...
+      const std::array<uint8_t,72> m_cert_dele;
+      const std::array<uint8_t,64> m_cert_sig;
+      const sys_microseconds64 m_utc_midpoint;
+      const microseconds32 m_utc_radius;
+   };
+
+struct BOTAN_PUBLIC_API(2,10) Link final
+   {
+      std::vector<uint8_t> response;
+      Ed25519_PublicKey public_key;
```
randombit: Avoid public members

src/lib/misc/roughtime/roughtime.h
```diff
@@ -0,0 +1,119 @@
...
+
+namespace Roughtime {
+
+constexpr unsigned request_min_size = 1024;
+
+/**
+* An Roughtime request.
+*/
+BOTAN_PUBLIC_API(2,10)
+std::array<uint8_t, request_min_size> encode_request(const std::array<uint8_t, 64>& nonce);
```
randombit: I'd prefer this was returning `std::vector` since (AIUI) a Roughtime request *could* (at least in theory) exceed the min length.

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,370 @@
...
+      {
+      {"publicKeyType", "ed25519"},
+      {"serverPublicKey", Botan::base64_encode(link.public_key.get_public_key())},
+      {"nonceOrBlind", Botan::base64_encode(link.nonce_or_blind.data(), link.nonce_or_blind.size())},
+      {"reply", Botan::base64_encode(link.response)},
+      };
+    all["links"].push_back(l);
+    }
+  const auto dump = all.dump(2);
+  ds.write((const uint8_t*)dump.data(), dump.size());
```
randombit: For char*<->uint8_t* casts, use `cast_char_ptr_to_uint8` and `cast_uint8_ptr_to_char` from mem_ops.h

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,370 @@
...
+      {
+      nonce = blind;
+      }
+  }
+void Chain::append(const std::vector<uint8_t>& response,
+                  const Ed25519_PublicKey& public_key,
+                  const std::array<uint8_t, 64>& nonce_or_blind,
+                  size_t max_chain_size)
+  {
+  if(max_chain_size <= 0)
```
randombit: Since the max_chain_size is unsigned this should be just == 0 right?

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,370 @@
...
+  return buf;
+  }
+
+Response Response::from_bits(const std::vector<uint8_t>& response,
+                             const std::array<uint8_t, 64>& nonce)
+  {
+  const auto response_v = split(response);
+  const auto cert = split(get_v(response_v,"CERT"));
+  const auto cert_dele = get<std::array<uint8_t,72>>(cert,"DELE");
+  const auto cert_sig =  get<std::array<uint8_t,64>>(cert,"SIG");
```
randombit: Is this ordering in the response guaranteed by the protocol?

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,370 @@
...
+  if(srep_midp > cert_dele_maxt)
+    throw Invalid_Argument("midpoint later than delegation end");
+  return {cert_dele, cert_sig, srep_midp, srep_radi};
+  }
+
+bool Response::validate(const Ed25519_PublicKey& pk) const
+  {
+  const char context[] = "RoughTime v1 delegation signature--";
+  PK_Verifier verifier(pk, "Pure");
+  verifier.update((const uint8_t*)context, sizeof(context)); //add context including \0
```
randombit: `cast_char_ptr_to_uint8`

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,370 @@
...
+  verifier.update(m_cert_dele.data(), m_cert_dele.size());
+  return verifier.check_signature(std::vector<uint8_t>(m_cert_sig.data(), m_cert_sig.data() + m_cert_sig.size()));
```
randombit: Instead of creating a new vector here just pass the pointer+length as two params: `m_cert_sig.data(), m_cert_sig.size()`

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,370 @@
...
+  if(!socket)
+     throw Not_Implemented("No socket support enabled in build");
+
+  const auto encoded = encode_request(nonce);
+  socket->write(encoded.data(), encoded.size());
+
+  if(std::chrono::system_clock::now() - start_time > timeout)
+    throw System_Error("Timeout during socket write");
+
+  std::array<uint8_t, 1024> buffer;
```
randombit: Since a vector is being returned, why not specify `buffer` as a vector also, then truncate to the resulting size? Then no copy is required.

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,370 @@
...
+  std::unique_ptr<Botan::HashFunction> hash(Botan::HashFunction::create("SHA-512"));
+  hash->update(previous_response);
+  hash->update(hash->final());
+  hash->update(blind.data(), blind.size());
+  hash->final(ret.data());
+
+  return ret;
+  }
+
+Chain::Chain(DataSource& ds)
```
randombit: This seems to just read to EOF. In that case I would rather this just took a string containing the entire JSON blob (`string_view` if we were C++17 land, but alas not an option right now) and the user can do the IO however they like before initializing the `Chain`. In the CLI you can use `slurp_file_as_str` to grab the entire contents of a file.

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,370 @@
...
+    unsigned step = 1024;
+    for(;;)
+    {
+      v.resize(read+step);
+      auto n = ds.read(v.data() + read, step);
+      read += n;
+      if(n!=step) break;
+    }
+    v.resize(read);
+    return nlohmann::json::parse(v.begin(), v.end());
```
randombit: Same as `Chain` - since this is just reading from the `DataSource&` to EOF and consuming all input, instead pass a `std::string&`

src/lib/misc/roughtime/roughtime.h
```diff
@@ -0,0 +1,119 @@
...
+                  size_t max_chain_size);
+      void persist(DataSink_Stream& ds) const;
+   private:
+      std::vector<Link> m_links;
+   };
+
+/**
+*/
+BOTAN_PUBLIC_API(2,10)
+std::array<uint8_t, 64> nonce_from_blind(const std::vector<uint8_t>& previous_response,
```
randombit: Instead of raw array this could return a `Roughtime::Nonce` as a firstclass type, then same type could be passed to `online_request`.

src/lib/misc/roughtime/roughtime.h
nunojpg: > https://roughtime.googlesource.com/roughtime/+/HEAD/PROTOCOL.md
> it‘s expected that a UDP request contain exactly 1024 bytes as there’s no point padding the request to more than the minimum.

I'm fine to change it you think vector will is "just in case" for API stability.

src/lib/misc/roughtime/roughtime.cpp
nunojpg: Is similiar expected for writing back to file instead of DataSink_Stream?

src/lib/misc/roughtime/roughtime.cpp
nunojpg: Yes. The check currently is just doing that. But I believe this is the right check to me made, since it still is correct if the variable is changed to signed.

src/lib/misc/roughtime/roughtime.cpp
randombit: Yes Chain::persist should just return a string

src/lib/misc/roughtime/roughtime.cpp
randombit: Fair enough

src/lib/misc/roughtime/roughtime.h
nunojpg: I did not change this one, confirm if I should.

nunojpg:
Updated with your comments. I'm sure there is still a lot to fix.

src/lib/misc/roughtime/roughtime.cpp
nunojpg: I don't see I make any assumptions on order. I construct a map with the tags, and then look them up, throwing if they don't exist of if size is not exactly the one expected.

randombit: Right sorry

randombit:
@nunojpg Sorry I have not had time to look at this further, but hope to soon. I think partially we are blocked on what to do about JSON support. I haven't figured out what to do here. Is it possible to just drop the JSON for now, and continue with the rest?

Also there is a merge conflict, can you rebase?

nunojpg:
I am still to add non ASIO network functions. Please give me 2 to 4 weeks to finish that, tests and rebase. Then we take a final look and decision about the json issue.

giorgiozoppi:
@nunojpg  You really need such runtime, why simply no using grpc, if you need udp, just using Quic support.

nunojpg:
Hi @randombit,

I now have a BSD Socket UDP implementation, and I removed JSON, I replaced with trivial space separated formats. For Botan alone there is no difference, JSON only made it easy to use the other implementations de facto format, which is not part of the standard anyway.

I would call this feature and API final, while it's not ready for merge until tests, documentation and eventually a updated of the protocol for the final Roughtime version, depending on when it comes.

I you would like to test, it requires minimum config `--minimized-build --enable-modules=roughtime,system_rng`, and you should create a file with the on-line servers:

file ecosystem:
```
Chainpoint-Roughtime ed25519 bbT+RPS7zKX6w71ssPibzmwWqU9ffRV5oj2OresSmhE= udp roughtime.chainpoint.org:2002
Cloudflare-Roughtime ed25519 gD63hSj3ScS+wuOeGrubXlq35N1c5Lby/S+T7MNTjxo= udp roughtime.cloudflare.com:2002
Google-Sandbox-Roughtime ed25519 etPaaIxcBMY1oUeGpwvPMCJMwlRVNxv51KK/tktoJTQ= udp roughtime.sandbox.google.com:2002
int08h-Roughtime ed25519 AW5uAoTSTDfG5NfY1bTh08GUnOqlRb+HVhbJ3ODJvsE= udp roughtime.int08h.com:2002
ticktock ed25519 cj8GsiNlRkqiDElAeNMSBBMwrAl15hYPgX50+GWX/lA= udp ticktock.mixmin.net:5333
```
This file have the same information as for eg: https://github.com/cloudflare/roughtime/blob/master/ecosystem.json

Run the CLI as:

`botan roughtime --servers-file=ecosystem`

This will create by default file "roughtime-chain", which can be replayed for proof of misbehaviour.

`botan roughtime_check roughtime-chain`

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,527 @@
...
+   if(bytes.size() < 8)
+      { throw Invalid_Argument("invalid structure"); }
+   const auto buf = bytes.data();
+   const uint32_t num_tags = buf[0];
+   const uint32_t start_content = num_tags * 8;
+   if(start_content > bytes.size())
+      { throw Invalid_Argument("invalid structure"); }
+   uint32_t start = start_content;
+   std::map<std::string, std::vector<uint8_t>> tags;
+   for(unsigned i=0; i<num_tags; ++i)
```
randombit: `i` should be a `uint32_t` to match `num_tags`

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,527 @@
...
+   hash->final(ret.data());
+   return ret;
+   }
+
+template<size_t N>
+std::array<uint8_t, N> vector_to_array(std::vector<uint8_t> vec)
+   {
+   if(vec.size() != N)
+      { throw Invalid_Argument("invalid vector size"); }
+   return *(std::array<uint8_t, N>*)vec.data();
```
randombit: This case/conversion looks really strange to me. Is this a legit transformation? It seems like a violation of strict aliasing.

src/lib/misc/roughtime/roughtime.h
```diff
@@ -0,0 +1,163 @@
...
+private:
+   std::string m_name;
+   Botan::Ed25519_PublicKey m_public_key;
+   std::vector<std::string> m_addresses;
+
+   };
+
+
+BOTAN_PUBLIC_API(2, 10)
+std::vector<Server> servers_from_file(const std::string&);
```
randombit: name the param

src/lib/utils/socket/uri.h
```diff
@@ -0,0 +1,49 @@
...
+#define BOTAN_URI_H_
+
+#include <cstdint>
+#include <string>
+
+#include <botan/build.h>
+
+namespace Botan {
+
+struct BOTAN_TEST_API URI
```
randombit: This class needs tests + a fuzzer

src/lib/utils/socket/uri.cpp
```diff
@@ -0,0 +1,155 @@
...
+   const auto port_pos = uri.find(':');
+   if(port_pos != std::string::npos)
+      {
+      for(char c : uri.substr(port_pos+1))
+         {
+         if(!std::isdigit(c))
+            return {};
+         port = port*10 + c - '0';
+         if(port > 65535)
+            return {};
```
devnexen: Would it be a bad idea if it was an exception thrown instead for all the `from...` ?

src/lib/misc/roughtime/roughtime.cpp
nunojpg: It's a violation... I only understood it is unacceptable recently.

nunojpg:
No review required for now. There is still WIP.

nunojpg:
@randombit should we add something like c++20 bit_cast next to other casts like cast_char_ptr_to_uint8?

This are based on the cppreference example:

```
//similar to c++20 bit_cast
template <class To, class From>
typename std::enable_if<
    (sizeof(To) == sizeof(From)) &&
    std::is_trivially_copyable<From>::value &&
    std::is_trivial<To>::value,
    To>::type
bit_cast(const From &src) noexcept
{
    To dst;
    std::memcpy(&dst, &src, sizeof(To));
    return dst;
}

//memcpy from pointer without checking size
template <class To, class From>
typename std::enable_if<
    std::is_trivially_copyable<From>::value &&
    std::is_trivial<To>::value,
    To>::type
bit_cast(const From *src) noexcept
{
    To dst;
    std::memcpy(&dst, src, sizeof(To));
    return dst;
}
```
I actually don't use the first version, and the second should probably not have the same name.

src/lib/utils/socket/uri.cpp
nunojpg: Done.

src/lib/utils/socket/uri.h
nunojpg: I've changed this to use inet_pton. Is the fuzzer to test for memory leaks or crashes? Can you give me some advice about it?

nunojpg:
I consider this PR to be API and feature complete. Only documentation is missing. As so at this time I ask for a review.

The CLI roughtime command works only by connecting to a server. Only the TLS commands also require a server, but the code implements both sides, so CLI tests can loopback them. But for roughtime the server is not implemented, so the CLI test would be always online, without a option to disable it. What do you think about it?

randombit:
Is the LGTM warning legit? If so it should be addressed. If not, let me know and I will report it to LGTM - I've found they are very responsive about addressing false positives.

nunojpg:
> Is the LGTM warning legit? If so it should be addressed. If not, let me know and I will report it to LGTM - I've found they are very responsive about addressing false positives.

It's definitively a false positive.

randombit:
Reported @ https://discuss.lgtm.com/t/false-positive-about-unused-static-constexpr-integer/2238

I'm going to be away and mostly not available for much of this week (at Blackhat) but I will try to get an initial review done of this before I leave - probably tonight or tomorrow night.

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,413 @@
...
+   hash->update(0);
+   hash->update(leaf.data(), leaf.size());
+   hash->final(ret.data());
+   return ret;
+   }
+
+std::array<uint8_t, 64> hashNode(const std::array<uint8_t, 64>& left, const std::array<uint8_t, 64>& right)
+   {
+   std::array<uint8_t, 64> ret;
+   std::unique_ptr<HashFunction> hash(HashFunction::create("SHA-512"));
```
randombit: `create_or_throw`

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,413 @@
...
+
+   const auto indx = get<uint32_t>(response_v, "INDX");
+   const auto path = get_v(response_v, "PATH");
+   const auto srep_root = get<std::array<uint8_t, 64>>(srep_v, "ROOT");
+   const auto size = path.size();
+   const auto levels = size/64;
+
+   if(size % 64)
+      { throw Invalid_Argument("Merkle tree path size must be multiple of 64 bytes"); }
+   if(indx >= std::exp2(levels))
```
randombit: Use of floating point here makes me queasy. Can't we do `indx >= (1 << levels)`?

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,413 @@
...
+      { throw Invalid_Argument("Merkle tree path is too short"); }
+
+   auto hash = hashLeaf(nonce.get_nonce());
+   auto index = indx;
+   auto level = 0u;
+   while(level < levels)
+      {
+      hash = index&1
+             ? hashNode(typecast_copy<std::array<uint8_t, 64>>(path.data() + level*64), hash)
+             : hashNode(hash, typecast_copy<std::array<uint8_t, 64>>(path.data() + level*64));
```
randombit: These are the only calls to hashNode. They copy the path data to another array buffer. But there is no reason to do that - just pass the raw pointers instead and have hashNode assume both inputs are 64 bytes.

(I wish we had a `span` type for things like this, but we don't get that in `std` until C++20 and I haven't gotten around to writing one for utils ...)

src/lib/misc/roughtime/roughtime.h
```diff
@@ -0,0 +1,158 @@
...
+
+namespace Botan {
+
+class RandomNumberGenerator;
+
+namespace Roughtime {
+
+constexpr unsigned request_min_size = 1024;
+
+class BOTAN_PUBLIC_API(2, 10) Nonce final
```
randombit: All version numbers should be 2, 12 now

src/lib/misc/roughtime/roughtime.h
```diff
@@ -0,0 +1,158 @@
...
+* @param url Roughtime server UDP endpoint (host:port)
+* @param timeout a timeout on the UDP request
+* @return Roughtime response
+*/
+BOTAN_PUBLIC_API(2, 10)
+std::vector<uint8_t> online_request(const std::string& uri,
+                                    const Nonce& nonce,
+                                    std::chrono::milliseconds timeout = std::chrono::seconds(3));
+
+struct BOTAN_PUBLIC_API(2, 10) Server final
```
randombit: Doxygen comment should make this clear that it's just the server information and not a roughtime server implementation.

[Actually to leave `Roughtime::Server` name open for a future work adding an actual Roughtime server, perhaps this should be named `Server_Information` instead?]

src/lib/misc/roughtime/roughtime.cpp
randombit: Also cleaner to remove the ternary here, instead pass `index&1` to `hashNode` then it can hash left/right or right/left as needed.

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,413 @@
...
+   {
+   std::vector<Server> servers;
+   std::stringstream ss(str);
+   for(std::string s; std::getline(ss, s);)
+      {
+      size_t start = 0, end = 0;
+      end = s.find(' ', start);
+      if(end == std::string::npos)
+         {
+         throw Invalid_Argument("wrong file format");
```
randombit: Decoding Error + better error messages here and rest of this function

src/tests/data/misc/roughtime_response.vec
```diff
@@ -0,0 +1,68 @@
+[Botan::Invalid_Argument:tag not found]
```
randombit: Don't hardcode exception messages into the test cases. Instead just have two blocks, valid and invalid, and (optionally) for invalid cases, comments explaining why they are invalid.

src/lib/utils/socket/socket_udp.cpp
```diff
@@ -0,0 +1,337 @@
...
+#elif defined(BOTAN_TARGET_OS_HAS_WINSOCK2)
+   #include <ws2tcpip.h>
+#endif
+
+namespace Botan {
+
+namespace {
+
+#if defined(BOTAN_HAS_BOOST_ASIO)
+class Asio_SocketUDP final : public OS::SocketUDP
```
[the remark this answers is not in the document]
nunojpg: I don't know how much effort is that. BTW, why don't we use standalone ASIO instead of raw OS sockets? Standalone ASIO (non-boost) is header only, and does not depend on any libraries. Basically, why support boost ASIO and OS sockets when standalone ASIO would cover all uses? (I am hoping this does not impact amalgation)

src/lib/misc/roughtime/roughtime.cpp
nunojpg: you don't want a return copy, so just hashNode(hash, path.data() + level*64, index&1), and hash is in/out parameter?

src/lib/misc/roughtime/roughtime.h
nunojpg: Agree.

src/lib/misc/roughtime/roughtime.cpp
nunojpg: Yes, this was very unsmart...

src/lib/misc/roughtime/roughtime.cpp
nunojpg: @randombit ping

randombit: Sorry missed this, yes in/out param is good.

nunojpg:
Hi @randombit 

I pushed changes on top on a new commit so you can take a quick look. Then I will rebase them were appropriate.

To be addressed from my side:

- Documentation
- URI class fuzzer
- Some more UDP class cleanups

From your side:

- I've implemented a a CLI test with a Python UDP server with hardcoded request and responses (obtained from a real server). I used the CLI with the drbg seed "802" * 32. so the default cli tests will pass. I see that some tests force this drbg_seed (see def cli_uuid_tests(_tmp_dir):), regardless of the cli test global setting. I should do the same correct?
- Answer my Aug 4 question:

> Also, using CLI roughtime command can only be done with external reference on the file format, as it is Botan specific. Yes, usually people would Google it and find on-line documentation. Not only for roughtime, but also for most commands, I think the CLI should have a option of printing a long (multi-line) help text per command.

randombit:
> I've implemented a a CLI test with a Python UDP server with hardcoded request and responses (obtained from a real server). I used the CLI with the drbg seed "802" * 32. so the default cli tests will pass. I see that some tests force this drbg_seed (see def cli_uuid_tests(_tmp_dir):), regardless of the cli test global setting. I should do the same correct?

Yes forcing the DRBG seed in order to verify the output exactly matches what we expect is good as then any (unintentional) change becomes visible via the tests.

> Also, using CLI roughtime command can only be done with external reference on the file format, as it is Botan specific. Yes, usually people would Google it and find on-line documentation. Not only for roughtime, but also for most commands, I think the CLI should have a option of printing a long (multi-line) help text per command.

This is already possible, just not currently used. If a `Command` instance overrides `Command::help_text` it can return any arbitrary string which will be printed if invoked with `--help`.

nunojpg:
> This is already possible, just not currently used. If a `Command` instance overrides `Command::help_text` it can return any arbitrary string which will be printed if invoked with `--help`.

In that case, the base implementation is:

      const std::string& cmd_spec() const
         {
         return m_spec;
         }

m_spec is non-const private. Can I make it const protected so that when I override I can still use it?

randombit:
> m_spec is non-const private. Can I make it const protected so that when I override I can still use it?

I don't understand. `Command::cmd_spec` is not virtual and afaict doesn't need to be - it is the way a subclass can access the original command template. `help_text` should be overridden, and can invoke `cmd_spec` to get the value.

nunojpg:
I mean when I override help_text, I can still read cmd_spec. It's not cmd_spec itself that I want to override!

Currently it's private and I want to make it protected.

Also making it const I think is appropriate.

randombit:
But `cmd_spec` is not private? So I don't understand what making`m_spec` const+protected changes, since for read only access in the subclass, you can just call the public `cmd_spec` to get the value.

Making it `const` is fine.

nunojpg:
> But `cmd_spec` is not private? So I don't understand what making`m_spec` const+protected changes, since for read only access in the subclass, you can just call the public `cmd_spec` to get the value.
> 
> Making it `const` is fine.

Sorry, I did not see the getter cmd_spec()

src/lib/misc/roughtime/info.txt
```diff
@@ -0,0 +1,9 @@
+<defines>
+ROUGHTIME -> 20190220
+</defines>
+
+<requires>
+rng
+ed25519
+socket
```
randombit: Since this uses SHA-512 directly it should give that as a dependency. (`sha2_64`)

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,413 @@
...
+         }
+      const auto serverPublicKey = Botan::Ed25519_PublicKey(Botan::base64_decode(s.substr(start, end-start)));
+
+      start = end + 1;
+      end = s.find(' ', start);
+      if(end == std::string::npos)
+         {
+         throw Invalid_Argument("wrong file format");
+         }
+      const auto nonceOrBlind = Nonce(vector_to_array<64>(Botan::unlock(Botan::base64_decode(s.substr(start, end-start)))));
```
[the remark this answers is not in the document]
nunojpg: 88 bytes

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,413 @@
...
+      { throw Not_Implemented("No socket support enabled in build"); }
+
+   const auto encoded = encode_request(nonce);
+   socket->write(encoded.data(), encoded.size());
+
+   if(std::chrono::system_clock::now() - start_time > timeout)
+      { throw System_Error("Timeout during socket write"); }
+
+   std::vector<uint8_t> buffer;
+   buffer.resize(1024);
```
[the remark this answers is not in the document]
nunojpg: @randombit  No, they are unrelated. This is for the response and not the request. I just checked and the typical answer is 360 bytes, and increases according to the levels of the merkle tree: 360+64*n. So I set it to 360+64*10+1 and left a comment
```
//response basic size is 360 bytes + 64 bytes for each level of merkle tree
//add one additional byte to be able to differentiate if datagram got truncated
```

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,413 @@
...
+   return typecast_copy<std::array<uint8_t, N>>(vec.data());
+   }
+}
+
+namespace Roughtime {
+
+Nonce::Nonce(const std::vector<uint8_t>& nonce)
+   {
+   if(nonce.size() != 64)
+      { throw Decoding_Error("invalid length for Nonce"); }
```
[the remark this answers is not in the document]
nunojpg: @randombit should the exception go into a new file?

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,413 @@
...
+std::vector<Response> Chain::responses() const
+   {
+   std::vector<Response> responses;
+   for(unsigned i = 0; i < m_links.size(); ++i)
+      {
+      const auto& l = m_links[i];
+      const auto nonce = i ? nonce_from_blind(m_links[i-1].response(), l.nonce_or_blind()) : l.nonce_or_blind();
+      const auto response = Response::from_bits(l.response(), nonce);
+      if(!response.validate(l.public_key()))
+         { throw Invalid_Argument("invalid signature or public key"); }
```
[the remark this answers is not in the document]
nunojpg: @randombit I have no knowledge of that. Why would we ignore it? If it fails then you can retry a request until you get a valid one or give up. I don't understand what you mean.

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,413 @@
...
+   for(uint32_t i=0; i<num_tags; ++i)
+      {
+      const uint32_t end = ((i+1) == num_tags) ? bytes.size() : start_content + typecast_copy<uint32_t>(buf + 4 + i*4);
+      if(end > bytes.size())
+         { throw Invalid_Argument("invalid structure"); }
+      if(end < start)
+         { throw Invalid_Argument("invalid structure"); }
+      const char* label_ptr = cast_uint8_ptr_to_char(buf) + (num_tags+i)*4;
+      const char label[] = {label_ptr[0], label_ptr[1], label_ptr[2], label_ptr[3], 0};
+      tags[label] = std::vector<uint8_t>(buf+start, buf+end);
```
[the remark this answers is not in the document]
nunojpg: Ok, I throw if a duplicated label appears.

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -0,0 +1,413 @@
...
+   if(tag->second.size() != sizeof(T))
+      { throw Invalid_Argument("tag content wrong size"); }
+   return typecast_copy<T>(tag->second.data());
+   }
+
+const std::vector<uint8_t>& get_v(const std::map<std::string, std::vector<uint8_t>>& map, const std::string& label)
+   {
+   const auto& tag = map.find(label);
+   if(tag == map.end())
+      { throw Invalid_Argument("tag not found"); }
```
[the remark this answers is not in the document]
nunojpg: @randombit  If I understand correctly, Invalid_Argument is for invalid human generated data, and Decoding_Error for invalid computer generated data?

src/lib/misc/roughtime/roughtime.cpp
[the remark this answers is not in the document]
nunojpg: @randombit I don't think this is valid. base_decode requires the output array to be larger than the  final length. While base64_decode returns 64, it actually wrote over the 64 buts. The sanitizer caught this for me. This appears to be the intended implementation, see base_decode:
`size_t written = (out_ptr - output) - base.bytes_to_remove(final_truncate);`
Do you have anothe suggestion?

src/lib/misc/roughtime/roughtime.cpp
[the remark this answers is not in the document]
randombit: Since the exception is only being thrown from Roughtime it is probably fine to define it in `roughtime.h`.

src/lib/misc/roughtime/roughtime.cpp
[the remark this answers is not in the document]
randombit: > While base64_decode returns 64, it actually wrote over the 64 boundary.

That seems like a bug. I will investigate.

src/lib/misc/roughtime/roughtime.cpp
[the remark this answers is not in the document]
randombit: From https://roughtime.googlesource.com/roughtime/+/HEAD/ECOSYSTEM.md
"Rather we plan on having Roughtime servers return invalid, bogus answers to a small fraction of requests. These bogus answers would contain the wrong time, but would also be invalid in another way. For example, one of the signatures might be incorrect, or the tags in the message might be in the wrong order. Client implementations that don’t implement all the necessary checks would find that they get nonsense answers". I don't know for certain if any current Roughtime server implements these behavior but we should assume it will occur.

You are right though, it is probably better to fail and have the application retry.

src/lib/misc/roughtime/roughtime.cpp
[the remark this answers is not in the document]
randombit: Correct yes - if the exception could have reasonably been avoided within the application code due to following documented restrictions on parameters, Invalid_Argument. If the data came from an external source possibly outside the applications control (eg from a file or network) Decoding_Error.

src/lib/misc/roughtime/roughtime.cpp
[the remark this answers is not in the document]
nunojpg: > > While base64_decode returns 64, it actually wrote over the 64 boundary.
> 
> That seems like a bug. I will investigate.

@randombit please don't forget this.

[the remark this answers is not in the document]
randombit: Thanks I had forgotten. I wrote a fuzzer trying to find the issue but it has found nothing so far. Was the issue reliably reproducible for you?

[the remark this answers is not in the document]
nunojpg: Yes, still the fail you can see I'm CI. I don't understand what are you looking for. See the code I copied above. The part of "base.bytes_to_remove(final_truncate)". This function is designed to need a array with some extra bytes which are trimmed at the end.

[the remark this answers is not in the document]
randombit: Sorry I didn't realize the CI failure was related here.

I don't think it is a bug, exactly, though certainly points to a bad API.

When decoding, the length of the output buffer is not specified. Instead it is assumed to always be at least base64_decode_max_output bytes, which is asserted (in a sense) by calling clear_mem on the output buffer zeroing that many bytes. Because the value passed to base64_decode is 88 bytes, with 2 bytes of `=` padding, it decodes to 64 bytes. But base64_decode_max_output doesn't look at contents and only length, so it assumes you have passed 88 bytes of base64 with no padding or whitespace. In that case the output would be 66 bytes, so that is the amount that is memset to zero at the beginning of decoding. But of course 64 is returned since that is the value actually decoded.

Short term fix is to check that the last 2 bytes are == then remove them before doing base64 decoding, which will produce exactly the 64 byte value expected.

It would be best to add some new base64 functions which take an explicit length for the output buffer, which would have found this issue immediately. This length assumption in the API is definitely a nice footgun.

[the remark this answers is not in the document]
nunojpg: I agree it's not a big. The function is just not intended to do what you asked me to use it for. I can though still avoid one copy by copying from secure_vector to array directly. 
I will not be changing base64 code, I don't have time to get this ready this week. And I really want to close this.
Are you doing some work on this or should I go ahead and go with the intermediate solution of memcpy secure vector to array?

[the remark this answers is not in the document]
randombit: I also would like to see this merged in time for 2.12 it has been long enough already.

Just copy it for now. I will do the base64 changes + update Roughtime to use them early in 2.13 cycle.

randombit:
@nunojpg CI is green - is this ready for final review before merge?

nunojpg:
No, but I finish tomorrow.

src/lib/misc/roughtime/roughtime.cpp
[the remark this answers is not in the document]
nunojpg: Do I throw Roughtime_Error instead of Decoding_Error? And is the base std::runtime_error or Decoding_Error?

When the API is used incorrectly, should I throw a Roughtime_Error or a Invalid_Argument?

src/lib/utils/socket/socket_udp.h
```diff
@@ -0,0 +1,73 @@
...
+      virtual ~SocketUDP() = default;
+
+      /**
+      * Write to the socket. Blocks until all bytes sent.
+      * Throws on error.
+      */
+      virtual void write(const uint8_t buf[], size_t len) = 0;
+
+      /**
+      * Reads up to len bytes, returns bytes written to buf.
```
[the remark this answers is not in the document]
nunojpg: umhh. The problem is this: I provide a buffer of 1024 bytes. If I read 1024 bytes I have no way of knowing if the packet was truncated or not. The only way of knowing that is to always pass one extra buffer byte and check if it is unused.
So either we have to specify that the user always provide a buffer 1 byte longer, or we need to dynamically allocate a temporary buffer inside.
I say this is up to the application using this class, since this is a typical limitation with UDP and the alternatives are not nice.

nunojpg:
I commented into two issues above.

I made a commit with all changes requested on top for easier review. When you give a OK I split the changes into the respective commits.

randombit:
Replying to earlier comments here. GH interface really falls down on longer convos

> Do I throw Roughtime_Error instead of Decoding_Error? And is the base std::runtime_error or Decoding_Error?

Base should be either `Exception` or `Decoding_Error`. We want to ensure that any exception we intentionally throw is catchable via `catch(Botan::Exception&)`. Probably Decoding_Error since (??) any such throw indicates some problem with decoding the messages.

> When the API is used incorrectly, should I throw a Roughtime_Error or a Invalid_Argument?

Invalid_Argument

> UDP interface

Ugh, OK I understand. I would like to come up with an alternate way of handling this but it is not pressing as this API is only internal use so we can always change it later on.

FYI the 2.12.0 release is scheduled for tomorrow, following usual schedule of releasing on first Monday of Jan/April/July/October. I think merging into that release is cutting it a little too close wrt limited testing. So I plan to merge this shortly after the 2.12 release and then we have a full release cycle for testing/fuzzing/any interface tweaks.

nunojpg:
@randombit OK. I push those changes now. You can merge it or not for this release. I think there have been no logic changes for months now.

Do you want to do a incremental review yet or can I split all the changes into the final commits?

nunojpg:
Btw:

class BOTAN_PUBLIC_API(2,0) Decoding_Error : public Invalid_Argument

I don't know if this is fully logical...

randombit:
Split up as expected for final merge. I will do a last review but don't expect there to be any issues.

Re Decoding_Error the derivation does not make sense at all. It is retained for now for compat, I will fix it in Botan 3. There is a comment above

```
* This exception derives from Invalid_Argument for historical reasons, and it                                                                                                         
* does not make any real sense for it to do so. In a future major release this                                                                                                        
* exception type will derive directly from Exception instead.                                                                                                                         
```

nunojpg:
Oups, looks like we had the same tought...

nunojpg:
Should I bump the API to 2, 13 now?

randombit:
Yes. Or I can do it after merge. Either way is fine.

nunojpg:
Ok, for me it's done, unless there is a CI surprise. Maybe you can also cancel all but the latest jobs I racked up in Appveyor.

randombit:
Looks great. Thanks for all your work (and patience) on this, it is a nice addition to the library.

PR #1846, opened 2019-02-28
Fix: cmake build

Tymolc:
One of the last commits broke the cmake build due to a missed variable renaming. This fixes it.


randombit:
Thanks

PR #1847, opened 2019-02-28
Run coverage tests in threaded mode

randombit:
Not sure if this will work but if it does it will save several minutes on the build.


randombit:
Doesn't actually help much

PR #1851, opened 2019-03-07
Android can possibly support arc4random API.

devnexen:


randombit:
I could not figure out exactly which Android version added support for `arc4random` but it seems to be at least 4.3 which is old enough that we don't need to care about anything earlier.

Issue #1853, opened 2019-03-10
RFC4880_encode_count doesn't return consistent results when  processing exact iterations


evpo:
The function returns RFC4880 code + 1 when given exact iterations. The code below demonstrates this behavior:

```
using namespace Botan;
std::stringstream size_t decode_result = OpenPGP_S2K::decode_count(160U);
stm << " decode(160)=" << decode_result << '\n';
stm << " encode(" << decode_result << ")=" <<
    static_cast<unsigned>(OpenPGP_S2K::encode_count(decode_result)) << '\n';
```

Output:
```
decode(160)=1048576
encode(1048576)=161
```

This issue is related: https://github.com/evpo/EncryptPad/issues/82

The behavior makes it impossible to open encrypted files.

PR is coming

PR #1854, opened 2019-03-10
Replace upper_bound with lower_bound in pgp_s2k

evpo:
Fix: RFC4880_encode_count doesn't return consistent results when processing exact iterations. It returns RFC4880 code + 1.

Fixes: #1853


randombit:
Sorry about that and thanks for the patch. Can you add something to the test (in `test_pbkdf.cpp`) that would detect this issue?

evpo:
No worries, man. I will look at the test.

evpo:
done

PR #1858, opened 2019-03-15
XMSS RFC 8391 Update

mgierlings:
Updates XMSS (draft status) to conform to RFC 8391


src/lib/pubkey/xmss/xmss_parameters.h
```diff
@@ -26,18 +26,18 @@ class BOTAN_PUBLIC_API(2,0) XMSS_Parameters
...
-         XMSS_SHA2_512_W16_H16 = 0x05000005,
-         XMSS_SHA2_512_W16_H20 = 0x06000006,
-         XMSS_SHAKE128_W16_H10 = 0x07000007,
-         XMSS_SHAKE128_W16_H16 = 0x08000008,
-         XMSS_SHAKE128_W16_H20 = 0x09000009,
-         XMSS_SHAKE256_W16_H10 = 0x0a00000a,
-         XMSS_SHAKE256_W16_H16 = 0x0b00000b,
-         XMSS_SHAKE256_W16_H20 = 0x0c00000c
+         XMSS_SHA2_10_256 = 0x00000001,
+         XMSS_SHA2_16_256 = 0x00000002,
```
randombit: So the format changed between draft-16 and RFC? :/

Is there any reasonable way we can continue supporting older keys/sigs? It seems this change would invalidate all existing uses.

At the very least we'll also want to use a different OID for keys so it is not possible to mix them.

randombit:
I sent out a warning letting anyone who is actively using current draft-06 support to let us know https://lists.randombit.net/pipermail/botan-devel/2019-March/002278.html and added a warning to the 2.10 release notes. If we don't hear from anyone by say May 1st, I say go ahead and merge this.

In the mean time we should still work to minimize disturbance:

* Increase XMSS datestamp in info.txt
* Change to a new OID arc (update also `doc/oids.txt`)
* Maybe even change `BOTAN_HAS_XMSS` to `BOTAN_HAS_XMSS_RFC8391`?

src/lib/pubkey/xmss/xmss_parameters.h
mgierlings: Yes, unfortunately the OID format did change. There exists another, new draft status document of interest regarding XMSS [Algorithm Identifiers for HSS and XMSS for Use in the Internet X.509 Public Key Infrastructure](https://tools.ietf.org/html/draft-vangeest-x509-hash-sigs-03). Also somewhere down the road XMSS will likely be assigned an "official" OID. So this may not be the last incompatible change to XMSS we see. There is also a disclaimer in RFC 8391:

> This document is not an Internet Standards Track specification; it is
> published for informational purposes.

>   This document is a product of the Internet Research Task Force
>   (IRTF).  The IRTF publishes the results of Internet-related research
>   and development activities.  These results might not be suitable for
>   deployment.  This RFC represents the consensus of the Crypto Forum
>   Research Group of the Internet Research Task Force (IRTF).  Documents
>   approved for publication by the IRSG are not candidates for any level
>   of Internet Standard; see Section 2 of RFC 7841.

>   Information about the current status of this document, any errata,
>   and how to provide feedback on it may be obtained at
>   https://www.rfc-editor.org/info/rfc8391.

A possibility to support older formats would be to integrate a key conversion utility into the botan cli.

randombit:
Haven't heard from anyone about this so likely this is good to merge, but lets wait until end of the month just to be sure.

randombit:
@mgierlings Can you rebase to address the merge conflict?

mgierlings:
@randombit Will do. I also have another update to this PR in the pipeline, which I'll push soon.

randombit:
It looks like you merged master into your branch rather than rebasing, this makes history very hard to read. Please rebase instead.

mgierlings:
@randombit The broken commit should be fixed. I'm not sure though what causes the softhsm error in some of the Travis CI builds.

randombit:
@mgierlings Thanks. I think the problem is `bench_xmss` in speed.cpp needs to be updated for the new param names.

randombit:
I don't think any change in performance was expected either way, but ftr I did check master vs this branch, they had only insignificant differences in sign and verify time.

randombit:
@mgierlings I fixed the test issue in 10630cc. Is this PR ready to merge?

mgierlings:
@randombit Looks good to me! One remark though on the XMSS parameter name changes introduced with 10630cc: The RFC actually uses "-" between XMSS and the hash function name, see [RFC 8391 Section 5.3 XMSS Parameters](https://tools.ietf.org/html/rfc8391#section-5.3). By replacing the dash with an underscore we technically use non standard compliant algorithm names.

securitykernel:
@mgierlings BouncyCastle uses the [same naming scheme](https://github.com/bcgit/bc-java/pull/513), maybe this is a thing? Anyway, whatever we decide here, we should probably make it consistent in both libraries.

mgierlings:
@securitykernel Agreed, we should be consistent. If others start implementing XMSS, they will likely use names from the standard document. The cleanest solution would probably be to adhere to the RFC's naming scheme (in both BouncyCastle and Botan). If we come up with our own now we might need to revisit this and change naming again in the future.

randombit:
I renamed the params because that's what the patched handbook uses for the param names. I find combining hyphen and underscore really awkward. But if that is the standard naming scheme, probably best to follow it.

mgierlings:
@randombit The algorithm names in the handbook are supposed to correspond to the `xmss_algorithm_t` enum items, since this is what is passed as argument in the code snippet below. However I did omit the bulky prefix `Botan::XMSS_Parameters::xmss_algorithm_t`, as a result the algorithm list in the handbook is somewhat ambiguous. To be more clear, we could change the handbook to list the available algorithms as follows:

```
Botan::XMSS_Parameters::xmss_algorithm_t::XMSS_SHA2_10_256 ("XMSS-SHA2_10_256")
Botan::XMSS_Parameters::xmss_algorithm_t::XMSS_SHA2_16_256 ("XMSS-SHA2_16_256")
Botan::XMSS_Parameters::xmss_algorithm_t::XMSS_SHA2_20_256 ("XMSS_SHA2_20_256")
...
```
or
```
"XMSS-SHA2_10_256" (Botan::XMSS_Parameters::xmss_algorithm_t::XMSS_SHA2_10_256)
"XMSS-SHA2_16_256" (Botan::XMSS_Parameters::xmss_algorithm_t::XMSS_SHA2_16_256)
"XMSS-SHA2_20_256" (Botan::XMSS_Parameters::xmss_algorithm_t::XMSS_SHA2_20_256)
...
```

randombit:
I switched param names back. So I think this is now good to merge? It's been almost two months since the notification went out warning about the change and it looks like nobody sees an issue with it.

mgierlings:
@randombit Yes, I'd say we're good to merge.

PR #1862, opened 2019-03-24
Darwin supports getentropy too

devnexen:


randombit:
Support already existed for it in the code but was disabled by default. However appears it was added in 10.12 and all earlier versions are now out of support, so safe to assume it is there.

webmaster128:
I just hit this when re-activating the macOS 10.11 CI agent: https://ci.kullo.de:8154/go/tab/build/detail/using-botan/416/test-botan/1/macOS10.11. The successor 10.12 is almost 3 years available now, so it is probably not wrong to dismiss 10.11.

Is there a good place in the docs to specify runtime system support?

PR #1864, opened 2019-03-25
Use a thread pool for XMSS signatures

randombit:
@mgierlings what do you think?

* i7-6700K (4 core w/SMT): 10% improvements
* Ryzen 7 2700 (8 core w/SMT): 25% improvement except SHA-256 which is almost 2x
* POWER8 (160 cores w/SMT): between 25 and 60% faster

Running all the signatures in a single thread pool also avoids some pathological behavior if an application performs several XMSS signatures in different threads which would otherwise lead to the CPU being massively oversubscribed.


mgierlings:
Very nice, the changes look good to me.

PR #1871, opened 2019-03-29
Tracing socket feature for the CLI TLS server.

devnexen:


randombit:
This is FreeBSD specific right? Please amend the commit message to describe usage/purpose. Otherwise ok.

PR #1874, opened 2019-03-31
Fix the incorrect syntax .PHONY=

evpo:
PHONY targets should be .PHONY: (see this [link](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html) )

The issue doesn't manifest because there are no files / directories matching targets for now.


randombit:
Thanks

PR #1877, opened 2019-04-01
Using available cores rather than total for the pools on linux/android/BSD.

devnexen:
Issue with the C++ api it might not reflect the reality of the H/W,
and on Android, in battery saving mode for example, might be more
reasonable to use what the system really offers.

### DIFF
--- a/src/lib/utils/os_utils.cpp
+++ b/src/lib/utils/os_utils.cpp
@@ -15,6 +15,10 @@
 #include <chrono>
 #include <cstdlib>
 
+#if defined(BOTAN_TARGET_OS_HAS_THREADS)
+  #include <thread>
+#endif
+
 #if defined(BOTAN_TARGET_OS_HAS_EXPLICIT_BZERO)
   #include <string.h>
 #endif
@@ -161,6 +165,36 @@ uint64_t OS::get_cpu_cycle_counter()
 #endif
 
    return rtc;
+   }
+
+size_t OS::get_cpu_total()
+   {
+   size_t tt = 1;
+#if defined(BOTAN_TARGET_OS_HAS_POSIX1) && defined(_SC_NPROCESSORS_CONF)
+   long res;
+   if ((res = ::sysconf(_SC_NPROCESSORS_CONF)) <= 0)
+     {
+          return 1;
+     }
+     tt = static_cast<size_t>(res);
+#elif defined(BOTAN_TARGET_OS_HAS_THREADS)
+     tt = static_cast<size_t>(std::thread::hardware_concurrency());
+#endif
+     return tt;
+   }
+
+size_t OS::get_cpu_available()
+   {
+#if defined(BOTAN_TARGET_OS_HAS_POSIX1) && defined(_SC_NPROCESSORS_ONLN)
+   long res;
+   if ((res = ::sysconf(_SC_NPROCESSORS_ONLN)) <= 0)
+     {
+       return 1;
+     }
+   return static_cast<size_t>(res);
+#else
+   return get_cpu_total();
+#endif
    }
 
 uint64_t OS::get_high_resolution_clock()
--- a/src/lib/utils/os_utils.h
+++ b/src/lib/utils/os_utils.h
@@ -50,6 +50,9 @@ bool running_in_privileged_state();
 */
 uint64_t BOTAN_TEST_API get_cpu_cycle_counter();
 
+size_t BOTAN_TEST_API get_cpu_total();
+size_t BOTAN_TEST_API get_cpu_available();
+
 /*
 * @return best resolution timestamp available
 *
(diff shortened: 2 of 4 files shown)


src/build-data/os/android.txt
```diff
@@ -18,4 +18,5 @@ getauxval
 sockets
 threads
 filesystem
+sysconf
```
randombit: sysconf function is part of POSIX.1-2001 so these can be removed and go off `defined(BOTAN_TARGET_OS_HAS_POSIX1)`

src/lib/utils/os_utils.cpp
```diff
@@ -15,6 +15,10 @@
 #include <chrono>
 #include <cstdlib>
 
+#if !defined(BOTAN_TARGET_OS_HAS_SYSCONF)
+  #include <thread>
```
randombit: This include should be guarded by `BOTAN_TARGET_OS_HAS_THREADS`  since there are platforms that have neither POSIX APIs nor thread support.

src/lib/utils/os_utils.cpp
```diff
@@ -163,6 +167,27 @@ uint64_t OS::get_cpu_cycle_counter()
    return rtc;
    }
 
+long OS::get_cpu_total()
+   {
+   static long tt = 0;
```
randombit: Remove the caching, this will cause warnings from race checkers since the access is not serialized. I know the call is slightly expensive but usually it will be followed by spawning some threads which is far more costly, so I don't think the overhead matters.

src/lib/utils/os_utils.cpp
```diff
@@ -163,6 +167,27 @@ uint64_t OS::get_cpu_cycle_counter()
...
+   if(tt == 0)
+#if defined(BOTAN_TARGET_OS_HAS_SYSCONF)
```
randombit: POSIX includes sysconf but does not guarantee existence of `_SC_NPROCESSORS_xx` so this guard should be `defined(BOTAN_TARGET_OS_HAS_POSIX1) && defined(_SC_NPROCESSORS_CONF)`

I'm not sure if POSIX actually guarantees that sysconf args are macros (vs an enum) but worst case, such (weird) platforms don't get CPU count info. So whatever.

src/lib/utils/os_utils.cpp
```diff
@@ -163,6 +167,27 @@ uint64_t OS::get_cpu_cycle_counter()
...
+      tt = sysconf(_SC_NPROCESSORS_CONF);
```
randombit: Use `::` to refer to functions in the global namespace, ie `::sysconf`

Also, check return value, if <= 0 then fall back to using `std::thread::hardware_concurrency` if available (`HAS_THREADS`) or just return 1

src/lib/utils/os_utils.cpp
```diff
@@ -163,6 +167,27 @@ uint64_t OS::get_cpu_cycle_counter()
...
+#else
+      tt = std::thread::hardware_concurrency();
+#endif
+   return tt;
+   }
+
+long OS::get_cpu_available()
+   {
+#if defined(BOTAN_TARGET_OS_HAS_SYSCONF)
+   return sysconf(_SC_NPROCESSORS_ONLN);
```
randombit: Ditto re `::sysconf`, checking return value, and changing macro guard.

src/lib/utils/os_utils.h
```diff
@@ -50,6 +50,9 @@ bool running_in_privileged_state();
 */
 uint64_t BOTAN_TEST_API get_cpu_cycle_counter();
 
+long BOTAN_TEST_API get_cpu_total();
+long BOTAN_TEST_API get_cpu_available();
```
randombit: Both functions should return size_t since the result should never be negative.

(Ideally we would have a `NonZero<T>` guard so as to statically check the result is also != 0, but that's a problem for later)

src/lib/utils/os_utils.cpp
randombit: Condition here is inverted (should be HAS_THREADS, not !HAS_THREADS)

src/lib/utils/os_utils.cpp
```diff
@@ -163,6 +167,36 @@ uint64_t OS::get_cpu_cycle_counter()
...
+   {
+   size_t tt = 0;
+#if defined(BOTAN_TARGET_OS_HAS_POSIX1) && defined(_SC_NPROCESSORS_CONF)
+   long res;
+   if ((res = ::sysconf(_SC_NPROCESSORS_CONF)) <= 0)
+     {
+          return 1;
+     }
+     tt = static_cast<size_t>(res);
+#else
```
randombit: This should be guarded with `HAS_THREADS`, then a final `#else` (for both no threads and no sysconf) that just sets tt to 1.

src/lib/utils/os_utils.cpp
randombit: Missing include due to this is causing Windows CI to fail.

devnexen: AH sorry ... typo.

Issue #1869, opened 2019-03-28
Implement a CertificateStore for CAs on Linux


alnkpa:
## Problem Description

In the effort to provide a platform-independent CA trust store (started with the [macOS integration](https://github.com/randombit/botan/pull/1830)), we are thinking about adding an implementation on Linux.
In contrast to Windows and macOS which both offer a standard interface to access the root CAs trusted by the system, Linux does not provide something like that.
Instead every distribution puts the trusted CAs either in a large amalgamation PEM file, in one or more directories one PEM-file at a time or some conbination of both.
Frankly, it seems to be (slightly) different for every distribution and other crypto libraries have various approaches to the issue (see [this blog post](https://www.happyassassin.net/2015/01/12/a-note-about-ssltls-trusted-certificate-stores-and-platforms/) for a pretty comprehensive writeup).

So before we start and try to implement something, we'd like to have your opinion on that.

## Ideas towards a solution

* Let the consuming application figure out where to find the trusted CAs by
  * compile-time configuration (assuming their packaging depends on the distribution anyway)
  * provide *the correct* trust store location in the `Certificate_Store_Linux` constructor
  * provide a list of potential locations in the constructor

* Let Botan decide internally by means of magic ;-)
  * probably a prioritized list of potential file system locations
  * by having a compile-time configurable default location (like OpenSSL does, for example), encouraging distributions to set their specific default location properly in their packaged Botan
* Let Botan ship an individual list of trusted CAs (the [NSS approach](https://wiki.mozilla.org/NSS_Shared_DB_And_LINUX))
  * Chrome [goes that way on Linux](https://github.com/chromium/chromium/blob/master/docs/linux_cert_management.md) while using the dedicated system store APIs on the other platforms
* Let OpenSSL decide
  * as stated above, OpenSSL (apparently) is shipped with the correct path [baked into the distribution-provided packages](https://github.com/openssl/openssl/blob/master/crypto/x509/x509_def.c)
  * we could [get the path from OpenSSL](https://stackoverflow.com/questions/4138139/how-to-find-out-the-path-for-openssl-trusted-certificate); obvious at the expense of a required dependency to OpenSSL
* don't implement `Certificate_Store_Linux` at all and let the application take care of it entirely :-(

randombit:
> Let the consuming application figure out where to find the trusted CAs

This dumps work on the application developers which is lame.

> Let Botan decide internally by means of magic ;-)

[...]

PR #1878, opened 2019-04-03
allow pem-decoding of file including multiple PEMs

alnkpa:
This PR starts the way to a linux certificate store (#1869). As trusted CAs are canonically stored in a single file, we introduce the ability to decode it.

### DIFF
--- a/src/lib/pubkey/pem/pem.cpp
+++ b/src/lib/pubkey/pem/pem.cpp
@@ -123,6 +123,23 @@ secure_vector<uint8_t> decode(DataSource& source, std::string& label)
    return base64_decode(b64.data(), b64.size());
    }
 
+std::vector<secure_vector<uint8_t>> decode_all(DataSource& source)
+   {
+   std::vector<secure_vector<uint8_t>> pems;
+
+   while(!source.end_of_data())
+      {
+      std::string label;
+      try
+         {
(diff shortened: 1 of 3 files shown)


src/lib/pubkey/pem/pem.cpp
```diff
@@ -123,6 +123,23 @@ secure_vector<uint8_t> decode(DataSource& source, std::string& label)
...
+   std::vector<secure_vector<uint8_t>> pems;
+
+   while(!source.end_of_data())
+      {
+      std::string label;
+      try
+         {
+         pems.push_back(decode(source, label));
+         }
+      catch(const Decoding_Error&) {}
```
alnkpa: If the file is completely parsed, but still includes newlines, it will throw when we try to `decode`. Is there a better way to mitigate that?

randombit: I think PEM interface is due for a design refresh (this code is very old) so that we can indicate errors in a cleaner way. But this works for now.

Anyway we probably want to catch exceptions here because (AIUI) the CA bundle file might include stray text/comments that we'll want to ignore.

randombit:
Changes look good but I'm going to close this in favor of #1885 so the changes can be reviewed as a whole.

PR #1884, opened 2019-04-08
FIX: Disable conflicting assertion macros from Apple headers

reneme:
That is a follow-up for "[Access to System Certificate Store on macOS]https://github.com/randombit/botan/pull/1830)" which fails to build as an amalgamation on macOS prior to and including XCode 9. Unfortunately, I noticed it only when trying to [package the new release for conan](https://github.com/bincrafters/conan-botan/pull/8). The [Travis CI build](https://travis-ci.com/bincrafters/conan-botan/builds/107369074) fails due to this problem for older macOS installations.

The actual problem is a legacy assertion macro defined in Apple's `AssertMacros.h` (i.e. `verify()`, `check()`, `require()` and so forth). Apparently, newer Xcodes ship a version of this header where these macros are disabled by default and replaced by `__verify()`, `__check()` and so on. Meh!

### DIFF
--- a/src/lib/x509/certstor_system_macos/certstor_macos.cpp
+++ b/src/lib/x509/certstor_system_macos/certstor_macos.cpp
@@ -11,6 +11,7 @@
 #include <algorithm>
 #include <array>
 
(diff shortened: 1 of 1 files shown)


randombit:
Down with macro namespace pollution.

PR #1888, opened 2019-04-10
Add a option for setting path to trusted CA list

randombit:
Related to #1885


randombit:
I'm not sure about the name ("system cert bundle") but it's the best I can come up with at the moment

PR #1892, opened 2019-04-12
Proposal to silent, when wished, some "false positives" when using sa…

devnexen:
…nitizers.

To be improved overtime eventually.

### DIFF
--- a/src/scripts/ubsan-blacklist.txt
+++ b/src/scripts/ubsan-blacklist.txt
@@ -1,2 +1,0 @@
-src:*dl_group.cpp
-src:*ecc_key.cpp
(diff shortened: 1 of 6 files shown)


randombit:
I don't understand. With Clang 7.0 I see no warnings running the test suite under UbSan. If there are any UbSan issues, we should fix them.

devnexen:
Well sure but I think the test assign on purpose some bad values thus do not if to be considered as false positive or not ... Tested with clang 7 on FreeBSD though.

randombit:
Interesting. We do have a few cases like that but they should all be guarded and only used if `--no-avoid-undefined` flag is set. Maybe something about the compilation (specific optimization flags or whatever) causes the undefined behavior to still be executed even if the flag is not set at runtime.

May be better to remove the flag and runtime tests, and guard these at build time with `#if !defined(BOTAN_HAS_SANITIZER_UNDEFINED)`

randombit:
I think you missed a git add

devnexen:
Yes all in fact

randombit:
Looks good thank you

PR #1893, opened 2019-04-12
Add a wrapper class that wraps macOS or Linux certificate stores.

randombit:
@alnkpa @reneme can you review?

### DIFF
--- a/src/cli/credentials.h
+++ b/src/cli/credentials.h
@@ -13,6 +13,10 @@
 #include <botan/data_src.h>
 #include <memory>
 
+#if defined(BOTAN_HAS_CERTSTOR_SYSTEM)
+   #include <botan/certstor_system.h>
+#endif
+
 inline bool value_exists(const std::vector<std::string>& vec,
                          const std::string& val)
    {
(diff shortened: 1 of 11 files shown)


src/cli/credentials.h
```diff
@@ -30,22 +34,18 @@ class Basic_Credentials_Manager : public Botan::Credentials_Manager
...
+            std::shared_ptr<Botan::Certificate_Store> cs(new Botan::Certificate_Store_In_Memory(ca_path));
+            m_certstores.push_back(cs);
             }
 
-         if(paths.empty() == false)
+         if(use_system_store)
             {
-            load_certstores(paths);
+            std::shared_ptr<Botan::Certificate_Store> cs(new Botan::System_Certificate_Store);
+            m_certstores.push_back(cs);
```
alnkpa: Why not:

```suggestion
            m_certstores.push_back(std::make_shared<Botan::Certificate_Store>());
```

src/lib/x509/certstor_system/info.txt
```diff
@@ -0,0 +1,8 @@
+<defines>
+CERTSTOR_SYSTEM -> 20190411
+</defines>
+
+<os_features>
+apple_keychain
+filesystem
```
alnkpa: It's either `apple_keychain` or `filesystem`, isn't it? If both are needed, `System_Certificate_Store` will only ever be included on macOS.

randombit: Syntax here is confusing, each line represents an alternative. AND is done with ',', like in https://github.com/randombit/botan/blob/master/src/lib/rng/system_rng/info.txt#L5

src/cli/credentials.h
randombit: GCC 4.8 has a bug converting unique_ptr<U> to unique_ptr<T> when U derives T. I'm not sure if shared_ptr has the same problem, I just reflexively worked around it. I'll check if it is a problem for shared_ptr as well.

randombit: shared_ptr works, unique_ptr doesn't.

randombit:
It seems my assumption in #1885 that the certificate store would never include non-CA certificates was not well founded; Ubuntu (and presumably then also Debian) include such certs in their trust stores. Who knows how this came about. I added an option so for the system cert store, we just ignore such certs, while by default we continue to throw an error. (Thus ignoring a problem we can do nothing about, but still sanity checking when someone uses the flatfile store with their own custom roots).

src/lib/x509/certstor_system/info.txt
reneme: Out of curiosity: Is there a write-up of the build system somewhere? So far I always defaulted to reading `./configure.py`...

randombit: There is not. But if I'm going to force people to deal with a one-off build system it really should be documented.

randombit: @reneme Still needs work but https://github.com/randombit/botan/blob/master/doc/manual/configure.rst

reneme: Thanks for the overview! :)

PR #1895, opened 2019-04-15
Provides auxiliary vector support for FreeBSD.

devnexen:
somewhat similar to Linux's counterpart but mainly for cpu features.


randombit:
Nice thank you

PR #1908, opened 2019-04-21
Add missing 'void' to C function declarations in ffi.h

JornVernee:
The C API in ffi.h defines functions which take no arguments using an empty formal parameter list. e.g.

    BOTAN_PUBLIC_API(2,0) uint32_t botan_ffi_api_version();

But in C, a function declaration with an empty parameter list declares a varargs function. See also: https://softwareengineering.stackexchange.com/a/286494/209013

To improve type-safety, it would be good to update these function declarations. Though, in my specific case, this is tripping up a tool that (_correctly_) translates these functions as varargs when automatically generating foreign language bindings.

### DIFF
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -99,7 +99,7 @@ BOTAN_PUBLIC_API(2,8) const char* botan_error_description(int err);
 * expressed in the form YYYYMMDD of the release date of this version
 * of the API.
 */
-BOTAN_PUBLIC_API(2,0) uint32_t botan_ffi_api_version();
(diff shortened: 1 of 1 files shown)


randombit:
Thank you! I don't write much C anymore these days and it is easy to slip into C++ thinking.

If your bindings work becomes public please add a link in https://github.com/randombit/botan/wiki/Language-Bindings

JornVernee:
@randombit Just experimenting at this point, but will do if it becomes substantial! 👍

giorgiozoppi:
I am interested in helping this because is needed for the project panama java binding

PR #1910, opened 2019-04-23
Warning compile fix for OpenBSD

devnexen:


randombit:
Use `BOTAN_UNUSED(id)` for this. Otherwise ok.

PR #1911, opened 2019-04-24
Install Boost via Homebrew on Travis macOS build

randombit:
Needed for testing #1839


hrantzsch:
I cherry-picked this commit on the stream PR, but it seems boost was already installed:

> Warning: boost 1.67.0_1 is already installed and up-to-date
To reinstall 1.67.0_1, run `brew reinstall boost`

Let me check if the beast::test::stream we use for the unit tests is not available in 1.67 yet.

randombit:
Wow, upgrading Boost on Homebrew involves installing Qt, Postgres, Python2, Python3, glib, NSS, and about 50 other packages. Took about 13 minutes to finish (!). Unfortunate but I think I'll pass, we'll just wait for new Travis images before the stream code can be tested in CI.

PR #1918, opened 2019-04-28
Remove explicit mips64 -mabi parameter

ffontaine:
This breaks when building for n32 ABI on mips64.

Signed-off-by: Baruch Siach <baruch@tkos.co.il>
[Bernd: rebased for botan-2.7.0]
Signed-off-by: Bernd Kuhls <bernd.kuhls@t-online.de>
[Retrieved from:
https://git.buildroot.net/buildroot/tree/package/botan/0001-remove-mips64-explicit-mabi.patch]


randombit:
Yeah makes sense. It's been a long time since I used any mips machines.

PR #1919, opened 2019-04-28
Project Panama Java API for Botan

giorgiozoppi:
Dear Botan maintainers, 
i am going on with this project using FFI API and I would like to merge. For now i have setup the enviroment, tested on Ubuntu with a simple hmac init. I have provided a simple unit test as well.
The development will be driven by test using your samples in the FFI interface.
In order to try this experimental extension you should:
1. Install Jetbrains Idea in /opt/jetbrains/idea or change botan.properties
1. Install Project Panama JDK Early Build access JVM at https://jdk.java.net/panama/
2. We suppose that your JAVA_HOME is /opt/java, so extract OpenJDK 13 the build at /opt/java
3. Install Botan at /usr/local or modify make_botan.sh to point to your botan path.
4. Run make_botan.sh
5. For build and test ant -f build.xml test
# Gradle support
The gradle support is here, but at the current state Gradle is not supporting
JDK 13 since it is not in general availability so unfortunately we have to live with ant until that time.
When we will have part of FFI covered will be interesting doing concurrency tests and performance tests agains the standard JCE provider and Bouncy castle....but for now thanks to provide 
such good cryptolibrary,
Best Regards,
Giorgio


randombit:
Thanks. Since this can't really "do" anything yet it seems to me not ready to merge, but we can keep the PR open for review/tracking. This does not have to expose all of FFI to be merged, even if it just does say hashing, MACs, and symmetric ciphers, that is sufficient functionality to be worth merging/distributing.

Few comments to start
 - All files should have a copyright statement by you (and/or your company, depending on what your rules may be re work for hire etc) and statement of license.
 - If it's at all possible I would prefer the gradle scripts be in Python as that is our house scripting language. Having a single script (instead of 1 batch + 1 shell) also ensures Windows and Unix automatically stay in sync.
 - It should be Hmac or HMAC, HMac looks strange. Also, that MAC interface allows other MACs (SipHash, Poly1305, CMAC, etc) so probably better to just name it MessageAuthenticationCode.
 - If the code is part of the main repo, it should also be build and tested in CI. It should probably be its own CI job.
 - Ideally we have some way to also build this as part of the native configure.py/make sequence. But this can come later.

giorgiozoppi:
Ok i keep doing in my cloned repo and when we have ready something. I will do another PR.

randombit:
No reason to open a new PR, just continue pushing commits to this branch.

randombit:
@giorgiozoppi I'm closing this PR since it seems the work is not active but we would still be interested in changes along the lines of what you started here.

PR #1927, opened 2019-05-03
Fix: make template friends compile on clang and MSVC

hrantzsch:
It turns out my final refactoring only compiles on GCC, sorry about that.

I found out that AppVeyor should actually provide a sufficiently new boost: https://www.appveyor.com/docs/windows-images-software/#boost. Can we enable building `--with-boost` on AppVeyor? Maybe by always adding the boost flag in `ci_build.py`?


randombit:
Thanks.

Testing Boost on AppVeyor sounds good.

PR #1931, opened 2019-05-03
Add a system certificate store for Windows

Tymolc:
Completing the trinity of system certificate stores.

It allows access into the windows system certificate store, using the Windows API and wincrypt.

Currently implemented are:

* `Certificate_Store::find_cert()`
* `Certificate_Store::find_all_certs()`
* `Certificate_Store::find_cert_by_pubkey_sha1()`
* `Certificate_Store::all_subjects()`

The Windows API for finding multiple certificates works iteratively by passing the result of a previous call of `CertFindCertificateInStore` into the next call. This function will then take care of freeing the previous `PCCERT_CONTEXT`. The RAII-wrapper around those handles is mainly there for exception-safety.

### DIFF
--- a/src/lib/x509/certstor_system_windows/certstor_windows.cpp
+++ b/src/lib/x509/certstor_system_windows/certstor_windows.cpp
@@ -231,6 +231,6 @@ Certificate_Store_Windows::find_cert_by_raw_subject_dn_sha256(const std::vector<
 std::shared_ptr<const X509_CRL> Certificate_Store_Windows::find_crl_for(const X509_Certificate& subject) const
    {
    BOTAN_UNUSED(subject);
-   throw Not_Implemented("Certificate_Store_Windows::find_crl_for");
+   return {};
    }
 }
--- a/src/lib/x509/certstor_system_windows/certstor_windows.cpp
+++ b/src/lib/x509/certstor_system_windows/certstor_windows.cpp
@@ -118,7 +118,7 @@ Cert_Vector search_cert_stores(const _CRYPTOAPI_BLOB& blob, const DWORD& find_ty
                                bool return_on_first_found)
    {
    Cert_Vector certs;
-   for(auto& store_name : cert_store_names)
+   for(const auto store_name : cert_store_names)
       {
       Handle_Guard<HCERTSTORE> windows_cert_store = open_cert_store(store_name);
       Handle_Guard<PCCERT_CONTEXT> cert_context = nullptr;
@@ -188,7 +188,7 @@ Certificate_Store_Windows::Certificate_Store_Windows() {}
 std::vector<X509_DN> Certificate_Store_Windows::all_subjects() const
    {
    std::vector<X509_DN> subject_dns;
-   for(auto& store_name : cert_store_names)
+   for(const auto store_name : cert_store_names)
       {
       Handle_Guard<HCERTSTORE> windows_cert_store = open_cert_store(store_name);
       Handle_Guard<PCCERT_CONTEXT> cert_context = nullptr;
@@ -230,7 +230,7 @@ Cert_Pointer Certificate_Store_Windows::find_cert_by_pubkey_sha1(const std::vect
    blob.cbData = static_cast<DWORD>(key_hash.size());
    blob.pbData = const_cast<BYTE*>(key_hash.data());
 
-   auto filter = [&](const Cert_Vector&, Cert_Pointer) { return true; };
+   auto filter = [](const Cert_Vector&, Cert_Pointer) { return true; };
 
    const auto certs = search_cert_stores(blob, CERT_FIND_KEY_IDENTIFIER, filter, true);
    return certs.empty() ? nullptr : certs.front();
--- a/src/lib/x509/certstor_system_windows/certstor_windows.cpp
+++ b/src/lib/x509/certstor_system_windows/certstor_windows.cpp
@@ -93,6 +93,9 @@ class Handle_Guard
          {
          if(m_context)
             {
+            // second parameter is a flag that tells the store how to deallocate memory
+            // using the default "0", this function works like decreasing the reference counter
+            // in a shared_ptr
             CertCloseStore(m_context, 0);
             }
          }
@@ -245,6 +248,7 @@ Cert_Pointer Certificate_Store_Windows::find_cert_by_raw_subject_dn_sha256(
 
 std::shared_ptr<const X509_CRL> Certificate_Store_Windows::find_crl_for(const X509_Certificate& subject) const
    {
+   // TODO: this could be implemented by using the CertFindCRLInStore function
    BOTAN_UNUSED(subject);
    return {};
    }
--- a/src/lib/x509/certstor_system_windows/certstor_windows.cpp
+++ b/src/lib/x509/certstor_system_windows/certstor_windows.cpp
@@ -18,12 +18,32 @@
 #include <Windows.h>
 #include <Wincrypt.h>
 
-
 namespace {
-const std::array<std::string, 4> cert_store_names{"MY", "Root", "Trust", "CA"};
-}
 
-namespace Botan {
+const std::array<std::string, 4> cert_store_names{"MY", "Root", "Trust", "CA"};
+
+HCERTSTORE openCertStore(const std::string& cert_store_name)
+   {
+   auto store = CertOpenSystemStore(0, cert_store_name.c_str());
+   if(!store)
+      {
+      throw Botan::Internal_Error(
+         "failed to open windows certificate store '" + cert_store_name +
+         "' (Error Code: " +
+         std::to_string(::GetLastError()) + ")");
+      }
+   return store;
+   }
+
+bool already_contains_certificate(
+   const std::vector<std::shared_ptr<const Botan::X509_Certificate>>& certs, std::shared_ptr<Botan::X509_Certificate> cert)
+   {
+   return std::any_of(certs.begin(), certs.end(),
+                      [&](std::shared_ptr<const Botan::X509_Certificate> c)
+      {
+      return *c == *cert;
+      });
+   }
 
 /**
  * Abstract RAII wrapper for PCCERT_CONTEXT and HCERTSTORE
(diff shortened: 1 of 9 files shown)


src/lib/x509/certstor_system_windows/info.txt
```diff
@@ -0,0 +1,15 @@
...
+<os_features>
+win32
+</os_features>
+
+<header:public>
+certstor_windows.h
+</header:public>
+
+<libs>
+all!windows -> crypto
```
randombit: Is this for Wine?

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -0,0 +1,236 @@
...
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/certstor_windows.h>
+#include <botan/der_enc.h>
+
+#include <array>
+#include <vector>
+#include <iostream>
```
randombit: stray iostream include, was for debugging perhaps?

src/lib/x509/certstor_system_windows/info.txt
Tymolc: Uhm ... I stole this from `src\lib\prov\openssl\info.txt` 
I think it didn't work before, but worked after ... it's entirely possible that another change fixed it though. I can check if the `all!` is still necessary.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -0,0 +1,236 @@
...
+#include <Windows.h>
+#include <Wincrypt.h>
+
+namespace {
+
+const std::array<std::string, 4> cert_store_names{"MY", "Root", "Trust", "CA"};
+
+HCERTSTORE openCertStore(const std::string& cert_store_name)
+   {
+   auto store = CertOpenSystemStore(0, cert_store_name.c_str());
```
reneme: `0` => `NULL`?

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+#include <array>
+#include <vector>
+#include <iostream>
+
+#define NOMINMAX 1
+#define _WINSOCKAPI_ // stop windows.h including winsock.h
+#include <Windows.h>
+#include <Wincrypt.h>
+
```
randombit: Botan namespace should begin here

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -0,0 +1,236 @@
...
+namespace {
+
+const std::array<std::string, 4> cert_store_names{"MY", "Root", "Trust", "CA"};
```
reneme: could be `const char *` because its global and const anyway.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
randombit: Use nullptr instead of 0/NULL

src/lib/x509/certstor_system_windows/certstor_windows.cpp
randombit: Does it make sense to hardcode these? They can (AIUI) change depending on configuration. They can be enumerated with CertEnumSystemStore

src/lib/x509/certstor_system_windows/info.txt
randombit: On Linux/etc -lcrypto == openssl
It looks like Crypt32.lib is needed for the Cert APIs so I think this should just be
windows -> crypt32

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -0,0 +1,236 @@
...
+   if(key_hash.size() != 20)
+      {
+      throw Invalid_Argument("Certificate_Store_Windows::find_cert_by_pubkey_sha1 invalid hash");
+      }
+
+   CRYPT_HASH_BLOB blob;
+   blob.cbData = static_cast<DWORD>(key_hash.size());
+   blob.pbData = const_cast<BYTE*>(key_hash.data());
+
+   for(auto& store_name : ::cert_store_names)
```
randombit: Design q. This enumerates all of the cert stores. Is it in some contexts reasonable to only use some subset of them? That would argue for the class to take a list of which specific ones to use to allow the user to control this if needed.
(I don't know the answer to this, I'm not really that familiar with how Windows cert stores are used in practice.)

src/lib/x509/certstor_system_windows/certstor_windows.cpp
reneme: `nullptr` doesn't seem to work...
The type of this parameter is `HCRYPTPROV_LEGACY` which is a typedef to `ULONG_PTR` which is a typedef to `unsigned __int3264` which is a (sort of?) integer type discussed [here](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/21eec394-630d-49ed-8b4a-ab74a1614611).

randombit: That's terrifying. OK.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -0,0 +1,236 @@
...
+      {
+      Handle_Guard<HCERTSTORE> windows_cert_store = openCertStore(store_name);
+      Handle_Guard<PCCERT_CONTEXT> cert_context = CertFindCertificateInStore(
+               windows_cert_store.get(), PKCS_7_ASN_ENCODING | X509_ASN_ENCODING,
+               0, CERT_FIND_KEY_IDENTIFIER,
```
randombit: Should this be CERT_FIND_SHA1_HASH?

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -0,0 +1,236 @@
...
+   std::vector<uint8_t> dn_data;
+   DER_Encoder encoder(dn_data);
+   subject_dn.encode_into(encoder);
+
+   CERT_NAME_BLOB blob;
+   blob.cbData = static_cast<DWORD>(dn_data.size());
+   blob.pbData = reinterpret_cast<BYTE*>(dn_data.data());
+
+   std::vector<std::shared_ptr<const X509_Certificate>> certs;
+   for(auto& store_name : ::cert_store_names)
```
randombit: If key_id is set it seems better to search by key id since that should be unique or nearly so, which will reduce the number of certs that must be parsed.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
Tymolc: Assuming we only want to find trust roots, it seems to be sufficient to search `Root` and `CA`. According to the docs, `MY` contains certificates of the current user with associated private keys.

I don't feel confident enough to use `CertEnumSystemStore` correctly... AFAICS it shouldn't be necessary.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+   return subject_dns;
+   }
+
+std::shared_ptr<const X509_Certificate>
+Certificate_Store_Windows::find_cert(const Botan::X509_DN&           subject_dn,
+                                     const std::vector<uint8_t>& key_id) const
+   {
+   const auto certs = find_all_certs(subject_dn, key_id);
+   return certs.empty() ? nullptr : certs.front();
```
randombit: Doing it this way does simplify the code a bit but it means that all cert stores must be searched/parsed even if the cert is found in the first one. Maybe pull the logic out to an anon namespace function which takes a bool `return_on_first_found` and then it either appends to the list or returns immediately depending on the value.

src/tests/test_certstor_system.cpp
```diff
@@ -131,6 +158,34 @@ Test::Result find_certs_by_subject_dn_and_key_id(Botan::Certificate_Store& certs
...
+      result.start_timer();
+      auto certs = certstore.find_all_certs(dn, std::vector<uint8_t>());
+      result.end_timer();
+
+      if(result.confirm("result not empty", !certs.empty()) &&
+            result.test_eq("exactly one certificate", certs.size(), 1))
+         {
+         auto cns = certs.front()->subject_dn().get_attribute("CN");
+         result.test_is_eq("exactly one CN", cns.size(), size_t(1));
+         result.test_eq("CN", cns.front(), "DST Root CA X3");
```
randombit: This fails because Ubuntu certstore has two certs with this DN. Maybe add the key id also? Or just verify >= 1 instead of == 1

src/lib/x509/certstor_system_windows/certstor_windows.cpp
Tymolc: s.a. It might make sense in some cases, but I don't know if there is much gain in doing that right now.

src/tests/test_certstor_system.cpp
```diff
@@ -76,6 +76,33 @@ Test::Result find_cert_by_subject_dn(Botan::Certificate_Store& certstore)
...
+
+      result.start_timer();
+      auto cert = certstore.find_cert(dn, std::vector<uint8_t>());
+      result.end_timer();
+
+      if(result.test_not_null("found certificate", cert.get()))
+         {
+         auto cns = cert->subject_dn().get_attribute("CN");
+         result.test_is_eq("exactly one CN", cns.size(), size_t(1));
+         result.test_eq("CN", cns.front(), "D-TRUST Root CA 3 2013");
```
randombit: Fails on Ubuntu "Failure 1: System Certificate Store - Find Certificate by UTF8 subject DN CN produced unexpected result 'D-TRUST Root Class 3 CA 2 2009' expected 'D-TRUST Root CA 3 2013'"

src/lib/x509/certstor_system_windows/certstor_windows.cpp
randombit: OK. If it comes up later we can add a new constructor taking the list and leaving the default constructor with the current behavior as default.

src/tests/test_certstor_system.cpp
Tymolc: This was supposed to be a regression test for the deduplication ...
I was assuming that two certificates with the same DN are the same. If this is not the case, the test doesn't make that much sense I guess (a test with the key id already exists)

src/lib/x509/certstor_system_windows/certstor_windows.cpp
Tymolc: Good point.

Signing off for today, though. I will address this and the other review comments on Monday.

src/tests/test_certstor_system.cpp
Tymolc: That's really strange ... Apparently the D-TRUST Root CA 3 2013 cert is not available on the Ubuntu agents, but it really shouldn't find a certificate with another name (and sadly it's also not yet reproducible for me ...)
I will investigate further.

Tymolc: I added a test for the flatfile certstore, which tries to find the same DN and should succeed if it doesn't find it (as this failing test does find a cert, but the wrong one) and that works.
On the Linux agents, the system certstore is just a flatfile_cerstor with `/etc/ssl/certs/ca-certificates.crt` as the path, right? Could you maybe extract this file from the agents? I'd like to use the flatfile certstore test with the default system CA list and find out what might be going on (tried it with the list of Ubuntu 18.10, but that also worked as expected)

randombit: Weird. I don't have any way to directly access the build VMs (that I know of) but for debugging purposes you can add arbitrary shell commands at the start of https://github.com/randombit/botan/blob/master/src/scripts/ci/setup_travis.sh and see the output in the build log.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
Tymolc: `CERT_FIND_SHA1_HASH` looks fo the calculated hash of the certificate, while `CERT_FIND_KEY_IDENTIFIER` searches by the field `subjectKeyIdentifier`, which is encoded into the certificate itself.
The test suggested that the latter is the expected behaviour.

src/tests/test_certstor_system.cpp
Tymolc: Indeed, with the CA list of the linux agent it's reproducible.

Tymolc:
I had to exclude MinGW for now, because `CertFindCertificateInStore` is not implemented there.

Tymolc:
Alright, I'm finally done fighting with the Travis-side CI. Appveyor still seems to have some problems. @randombit could you maybe cancel all the appveyor builds but the last one, so it doesn't take all that long to see if there are still issues?
I also think this is ready for another review.

Tymolc:
Hm, Windows doesn't seem to find the `D-TRUST Root Class 3 CA 2 EV 2009`, which according to https://social.technet.microsoft.com/wiki/contents/articles/51151.microsoft-trusted-root-certificate-program-participants-as-of-january-30-2018.aspx#D should be available. Now I am unsure if the cert, for some reason, isn't available on the Appveyor agent or if there is a bug. Do you have more access to the AppVeyor agents than the Travis agents by any chance?

randombit:
It appears it is possible to RDP to AppVeyor build VMs https://www.appveyor.com/docs/how-to/rdp-to-build-worker/ I can check into this but probably not until late tonight.

randombit:
The build VM has an oddly sparse list of CA certs.

image

Tymolc:
Alright, thanks. I think I will just disable the test on Windows, then. It somewhat uncovered a bug on Linux, so I feel like it has some worth there (even though what it is testing isn't actually the reason it failed on Linux...)

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -0,0 +1,251 @@
...
+      }
+   return store;
+   }
+
+Cert_Vector search_cert_stores(const _CRYPTOAPI_BLOB& blob, const DWORD& find_type,
+                               std::function<bool(const Cert_Vector& certs, Cert_Pointer cert)> filter,
+                               bool return_on_first_found)
+   {
+   Cert_Vector certs;
+   for(auto& store_name : cert_store_names)
```
reneme:
```suggestion
   for(const auto store_name : cert_store_names)
```

its a pointer, no point to iterate via reference...

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -0,0 +1,251 @@
...
+         }
+
+      T operator->()
+         {
+         return m_context;
+         }
+
+   private:
+      template<class T2 = T>
+      typename std::enable_if<std::is_same<T2, PCCERT_CONTEXT>::value>::type close()
```
reneme: Just curious: An old-school overload didn't do the trick here? As in:

```C++
void close(PCCERT_CONTEXT ctx) { /* ... */ }
void close(HCERTSTORE ctx) { /* ... */ }
```

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -0,0 +1,251 @@
...
+            CertFreeCertificateContext(m_context);
+            }
+         }
+
+      template<class T2 = T>
+      typename std::enable_if<std::is_same<T2, HCERTSTORE>::value>::type close()
+         {
+         if(m_context)
+            {
+            CertCloseStore(m_context, 0);
```
reneme: what's the 0 magic number here? Mind to comment?

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -0,0 +1,251 @@
...
+         {
+         auto cert = std::make_shared<X509_Certificate>(cert_context->pbCertEncoded, cert_context->cbCertEncoded);
+         if(filter(certs, cert))
+            {
+            if(return_on_first_found)
+               {
+               return {cert};
+               }
+            certs.push_back(cert);
+            }
```
reneme: That's a bit of a guess, but I believe, if you do it like that, you should benefit from copy elision on return:
```C++
if(filter(certs, cert))
   {
   certs.push_back(cert);
   if(return_on_first_found)
      {
      return certs;
      }
   }
```

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -0,0 +1,251 @@
...
+   return search_cert_stores(blob, find_type, filter, return_on_first_found);
+   }
+} // namespace
+
+Certificate_Store_Windows::Certificate_Store_Windows() {}
+
+std::vector<X509_DN> Certificate_Store_Windows::all_subjects() const
+   {
+   std::vector<X509_DN> subject_dns;
+   for(auto& store_name : cert_store_names)
```
reneme:
```suggestion
   for(const auto store_name : cert_store_names)
```

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -0,0 +1,251 @@
...
+   if(key_hash.size() != 20)
+      {
+      throw Invalid_Argument("Certificate_Store_Windows::find_cert_by_pubkey_sha1 invalid hash");
+      }
+
+   CRYPT_HASH_BLOB blob;
+   blob.cbData = static_cast<DWORD>(key_hash.size());
+   blob.pbData = const_cast<BYTE*>(key_hash.data());
+
+   auto filter = [&](const Cert_Vector&, Cert_Pointer) { return true; };
```
reneme:
```suggestion
   auto filter = [](const Cert_Vector&, Cert_Pointer) { return true; };
```

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -0,0 +1,251 @@
...
+   const std::vector<uint8_t>& subject_hash) const
+   {
+   BOTAN_UNUSED(subject_hash);
+   throw Not_Implemented("Certificate_Store_Windows::find_cert_by_raw_subject_dn_sha256");
+   }
+
+std::shared_ptr<const X509_CRL> Certificate_Store_Windows::find_crl_for(const X509_Certificate& subject) const
+   {
+   BOTAN_UNUSED(subject);
+   return {};
```
reneme: Is that implementable with the Windows API? Maybe add a comment at least, in case it is.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
Tymolc: It would have, but we liked the `enable_if` version better.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
Tymolc: Hmm. It's difficult to validate this. Playing around on godbolt for a bit seemed to show that not pushing into the vector and returning (i.e. the current version) seems to use less instructions, but I don't know if the example may be too minimal ... https://godbolt.org/z/vf56r2

reneme:
At least one windows installation seems to fail with
```
System Certificate Store - Find all Certificate Subjects ran 1 tests 1 FAILED
Failure 1: Time 30160807230117Z does not seem to be valid
```
:-(

randombit:
In the screenshot you can see their local "master-win2016" cert expires in 3016. Weird that MS even lets you create something like this. Currently we reject as a general sanity check any dates > 2200. I was expecting to be dead by the time it became a problem.

I'll write a patch to update the limit to 3100 tonight. Then we are, uh, future proofed? Something like that.

randombit:
Rebase to pick up 249edbf. PR looks good, ready to merge when CI is green.

pgp:
Hi everyone,
is there any chance of including this certificate store implementation also for mingw os configuration?
Using latest botan master, I currently manage to compile and use an HTTPS url download example through mingw toolchain, but using amalgamation generated with:
```
python configure.py --amalgamation --disable-modules=pkcs11,tls_10 --cpu=x64 --os=windows --cc=gcc
```
and then patching it by including the definition of inet_pton, that mingw misses, from [here](https://stackoverflow.com/questions/15370033/how-to-use-inet-pton-with-the-mingw-compiler)

Since the compiled example seems to work correctly, I assume that windows certstore implementation could also be enabled naturally for mingw (currently, if I configure with --os=mingw, the feature certstore_system_windows is listed as skipped due to OS incompatiblity)

Tymolc:
I believe the reason why we had to disable mingw was that its crypt32 library didn't implement some of the necessary Windows API methods. Specifically, `CertFindCertificateInStore` seems to have been the problem.
If you want to try getting it to work, you just have to include 
```mingw -> crypt32`

**Comment by pgp** [comment]
Hi, trying to apply the above modification in the info.txt doesn't yet allow me to configure with mingw and certstor_system_windows enabled (still figures as skipped due to incompatible OS), it's still unclear to me how exactly feature incompatiblity is detected through the configure process (because in the info.txt file, I believe you specify the linker dependencies, right?)

**Comment by Tymolc** [comment]
Ah right. I remember now. First of all, [here](https://botan.randombit.net/handbook/dev_ref/configure.html) is a summary of what the `info.txt` files can do.
Second, the `os_features` are actually what prevent mingw from being compatible. We introduced the `certificate_store` os_feature in this PR. 
You can look at `src/build-data/os/mingw.txt` and `src/build-data/os/windows.txt` to see which os_features they provide.
So the way to go would be to either remove the `certificate_store` feature from `info.txt` or add it to `mingw.txt`

**Comment by pgp** [comment]
Ok, just tried adding certificate_store in mingw.txt, and then configuring amalgamation with:
```
python configure.py --amalgamation --disable-modules=pkcs11,tls_10 --cpu=x64 --os=mingw --cc=gcc
```
and now the example successfully builds. I even tried a dummy usage of the `CertFindCertificateInStore` function and it works (i'm linking against crypt32, clearly).
I'm using MinGW from [here](https://nuwen.net/mingw.html), maybe the culprit of the missing `CertFindCertificateInStore` definition was an older mingw distribution?

Tymolc:
That's very possible. I didn't have a mingw setup of my own back then. My only indication was the CI and some googling. Try opening a pull request and see if the CI will be green.

pgp:
Ok, just done..

PR #1936, opened 2019-05-07
use map of vectors instead of multimap in flatfile certstor

Tymolc:
There appeared to be a problem using the std::multimap::equal_range function for retrieving certificates to a given subject DN. Using the ca-certificates.crt from the Linux Travis Agent, `find_all_certs` for a subject_dn and no key_id returned wrong results.
The reason for that is still unclear to me, but using a map of vectors mitigates the problem.

This should also fix the tests for #1931


randombit:
I bet the problem is X509_DN operator< is buggy. I will look at this. In the meantime will merge this since either approach is fine and it unblocks your other work.

PR #1957, opened 2019-05-21
Fix: allow comments after #endif in amalgamation

hrantzsch:
The regex does not match `#endif` statements that have a comment, such as here: https://github.com/randombit/botan/blob/b1f974a29d6fab6ff8ff6ff91fd4ba7abc211e57/src/lib/tls/asio/asio_stream.h#L674

As far as I can see this only occurs in our tls/asio module and in two files in the pkcs11 module. Of course, if you prefer to keep the regex as it is I can remove the comments instead.


randombit:
This way is fine

PR #1958, opened 2019-05-21
Allow multiple external include and library dirs in ./configure.py

reneme:
To further simplify the integration of `./configure.py` and [bincrafter's conan recipe](https://github.com/bincrafters/conan-botan), I'd like to be able to define multiple include and library search paths. This proposes to allow multiple specifications of `--with-external_(lib|include)dir=`, one for each additional search path.

The change should be backward-compatible.


randombit:
Looks good, thanks

PR #1960, opened 2019-05-21
Allow specification of external compile-time definitions

reneme:
This is related to [my PR earlier today](https://github.com/randombit/botan/pull/1958) and aims to allow compile-time definitions when compiling the library with external dependencies (e.g `-DBOOST_USE_STATIC_LIB`).

I opted for a new `./configure.py` switch, because the compiler parameter differs between MSVC and all other compilers. Otherwise `--extra-cxxflags` would have probably sufficed. In that respect: Did I extend the compiler abstractions properly?

Also: Is piggy-backing the compile-time defines onto the `%{cc_compile_flags}` acceptable? I wanted to avoid touching all build system output templates.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -420,8 +420,8 @@ def process_command_line(args): # pylint: disable=too-many-locals,too-many-state
     build_group.add_option('--with-external-libdir', metavar='DIR', default=[],
                            help='use DIR for external libs', action='append')
 
-    build_group.add_option('--with-external-compile-definition', default=[], action='append',
-                           help='set compile-time definition like KEY[=VALUE]')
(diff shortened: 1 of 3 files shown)


randombit:
Yes this is exactly how you would do something along these lines.

If someone had asked I would have leaned to directing them to `--extra-cxxflags` but as you say MSVC has a different syntax here.

Main issue is, I find the name very awkward and it is not clear from that alone what it is doing. Maybe we can use some simpler/"snappier" name, like `--define-build-macro` or ???

reneme:
I changed it to `--define-build-macro`. Wasn't happy with the name either.

PR #1962, opened 2019-05-21
Providing little getauxval implementation for Android<18

devnexen:
tested with armv7 and ndk 16.


randombit:
Thanks. One comment: `environ` and `elf.h` are only needed for `defined(BOTAN_TARGET_OS_IS_ANDROID)` case so move both of these to a new section in the preamble. Otherwise lgtm.

Issue #1930, opened 2019-05-03
Test Boost in AppVeyor MSVC builds


randombit:
Re #1927

reneme:
While finishing up our Windows build we realised that boost has [peculiar naming conventions](https://www.boost.org/doc/libs/1_66_0/more/getting_started/unix-variants.html#library-naming) for its libraries. On Windows, static libraries are prefixed with `lib` (e.g. `libboost_system.lib`), "import libraries" are not (e.g. `boost_system.lib` stubbing for some `boost_system.dll`). And further there are some layers of name-based versioning, that are configurable at build time.

We tried to work around the `*.lib` suffix in #1940. That is not enough though, as we want to link statically, i.e. the `lib...` prefix would be required as well on windows. At this point we're not sure how to allow for both dynamic and static linkage with the `info.txt` machinery on Windows. Any thoughts on this?

We'll have a closer look at this tomorrow and (hopefully) also get to test it on AppVeyor in the process.

reneme:
@randombit In case you read this soon: It would be cool if you could kill all AppVeyor builds except the latest one (with git sha: 2f64492).

randombit:
@reneme I just killed all but the most recent commit.

For testing might help to temporarily prune the build matrix to a single build, get that one working, then re-add the others.

reneme:
Thanks for reviewing and merging this. Would be great to run the asio tests on the other platforms some day too...

randombit:
Sadly Travis really lags behind, Ubuntu 18.04 was released before 16.04 became available for CI. At that rate, we'll get 18.04 support in 2020 - and because in turn Ubuntu also specializes in releasing obsolete software, 18.04's Boost is too old anyway, so that doesn't help anyway :(

I think best chance is to move Travis to a Docker setup so we can control the image and pre-populate useful packages.

But for this ticket, closing, fixed in #1964

PR #1963, opened 2019-05-22
Allow specification of platform/build dependent boost library name

reneme:
See #1930 for details.

This allows to specify a list of boost library names for `./configure.py` that are then associated with the boost library names to be linked from `info.txt` files.

For example:
```
# info.txt
<libs>
boost_system,boost_filesystem
</libs>
```

```
# on windows
$ ./configure.py --verbose --with-boost --boost-library-name libboost_system.lib --boost-library-name libboost_filesystem.lib
DEBUG: Replacing boost library name boost_system -> libboost_system.lib
DEBUG: Replacing boost library name boost_filesystem -> libboost_filesystem.lib

$ nmake
link /LIBPATH:C:/.... botan.lib libboost_system.lib libboost_filesystem.lib
```

Note that [starting from 1.69.0](https://www.boost.org/doc/libs/1_70_0/libs/system/doc/html/system.html#changes_in_boost_1_69) `boost_system` is actually header-only. Hence, linking wouldn't be necessary anymore. Particularly, since `boost_system` is the only boost library in botan that (used to) require linking -- now that `boost_filesystem` is out. However, to keep compatibility with older versions of boost, we believe it still makes sense to have this build system enhancement.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -488,6 +488,8 @@ def process_command_line(args): # pylint: disable=too-many-locals,too-many-state
                            help=optparse.SUPPRESS_HELP)
     build_group.add_option('--without-pkg-config', dest='with_pkg_config', action='store_false',
                            help=optparse.SUPPRESS_HELP)
+    build_group.add_option('--boost-library-name', dest='boost_libnames', default=[],
+                           help="file name of some boost library to link", action='append')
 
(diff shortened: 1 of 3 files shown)


reneme:
Note that the CI just turned red, because @randombit (inadvertently) killed the AppVeyor job just before the last stage.

reneme:
closing in favour of #1964

PR #1964, opened 2019-05-22
Allow specification of platform/build dependent boost library name + AppVeyor uses boost 

reneme:
This is based on #1963 and uses the pre-installed boost 1.69.0 on AppVeyor. The `appveyor.yml` now specifies which version of boost to use for each matrix configuration. Note that we opted to use the static library versions of boost in all cases. Probably its also possible to use the dynamic ones, but we didn't test that.

~~Once AppVeyor turns green for the first time, I'll squash the comments, if you want.~~

Re: Re: #1930


randombit:
It works! Thanks for getting this set up.

PR #1968, opened 2019-05-23
TLS::Stream documentation

hrantzsch:
Add documentation for TLS::Stream to the manual section about TLS.
The description includes a very basic but working example to make a single https request.

I also added the TLS stream to the list of "Other Useful Things" in the readme and I took the liberty to add myself to the list of contributors :)


randombit:
Thanks!

PR #1974, opened 2019-05-24
expose this_update and next_update in OCSP::Response

alnkpa:
As per [RFC 6960](https://tools.ietf.org/html/rfc6960#section-3.2), when receiving an OCSP response, a client shall confirm that:

> 5. The time at which the status being indicated is known to be correct (thisUpdate) is sufficiently recent

Currently, a user can't examine that field of an OCSP response parsed by botan. This PR exposes both `thisUpdate` and `nextUpdate` as a new struct.

As the requested `issuer` and `subject` might not be in the response, the return struct also includes a `bool m_has_value`.


randombit:
I wish we could use C++17 `std::optional` here. :/ But this is ok. Thanks.

randombit:
OK, looking at this more, I am unsure why it is required. `Response::status_for` takes a timestamp and verifies that this_update <= now <= next_update, returning `OCSP_NOT_YET_VALID` or `OCSP_HAS_EXPIRED` otherwise (or `OCSP_CERT_NOT_LISTED` if the cert isn't given in the response). Is that not sufficient?

alnkpa:
When `next_update` is not set (it is optional, after all), we want to check if a response is "suffiently recent" to mitigate potential replays of the response.

randombit:
How about instead adding a new param to `status_for` which specifies the maximum age allowed for the response (in seconds), if next_update is not set.

Then change this logic:

```
         if(response.next_update().time_is_set() && x509_ref_time > response.next_update())
            return Certificate_Status_Code::OCSP_HAS_EXPIRED;
```

to something like

```
         if(response.next_update().time_is_set())
            {
            if(x509_ref_time > response.next_update())
                return Certificate_Status_Code::OCSP_HAS_EXPIRED;
            }
         else if(max_age > 0)
             {
             if(x509_ref_time - this_update() > max_age)
                 return Certificate_Status_Code::OCSP_IS_TOO_OLD; // new error code
             }
```

alnkpa:
Sounds good to me. The only question here is what happens if the `max_age` has been reached, but `next_update` is still in the future.

src/lib/x509/cert_status.h
```diff
@@ -48,6 +48,7 @@ enum class Certificate_Status_Code {
    OCSP_HAS_EXPIRED = 2003,
    CRL_NOT_YET_VALID = 2004,
    CRL_HAS_EXPIRED = 2005,
+   OCSP_IS_TOO_OLD = 2006,
```
randombit: Update `to_string` with an error message for this

src/tests/test_ocsp.cpp
```diff
@@ -154,6 +154,37 @@ class OCSP_Tests final : public Test
...
+         const auto valid_time = Botan::calendar_point(2016, 11, 20, 8, 30, 0).to_std_timepoint();
+         const auto max_age = std::chrono::hours(1);
+         const auto ocsp_status = Botan::PKIX::check_ocsp(cert_path, { ocsp }, { &certstore }, valid_time, max_age);
+
+         if(result.test_eq("Expected size of ocsp_status", ocsp_status.size(), 1))
+            {
+            if(result.test_eq("Expected size of ocsp_status[0]", ocsp_status[0].size(), 1))
+               {
+               result.confirm("Status good", ocsp_status[0].count(Botan::Certificate_Status_Code::OCSP_RESPONSE_GOOD));
+               }
```
randombit: Should also be a test (with different max_age) that the expected error is returned

src/tests/test_ocsp.cpp
```diff
@@ -225,6 +266,7 @@ class OCSP_Tests final : public Test
          results.push_back(test_response_parsing());
          results.push_back(test_response_certificate_access());
          results.push_back(test_response_verification());
+         results.push_back(test_response_verification_too_old());
```
reneme: We should add another test that uses an OCSP response with a `next_update` specification and validate that the `max_age` does not have an effect.

src/lib/x509/cert_status.h
reneme: ✔️

src/tests/test_ocsp.cpp
reneme: ✔️

alnkpa:
This should address all problems that were mentioned.

PR #1979, opened 2019-05-26
Fix PKCS#11 C_Decrypt, C_Encrypt, C_Sign, C_SignFinal buffer output size.

dewyatt:
~~The spec does say that modules should set pulBufLen to the exact byte
count when pBuf is not NULL. However, some modules fail to do this and
it's easy to work around here.~~

EDIT: See below.

### DIFF
--- a/src/lib/prov/pkcs11/p11.h
+++ b/src/lib/prov/pkcs11/p11.h
@@ -2305,7 +2305,12 @@ class BOTAN_PUBLIC_API(2,0) LowLevel
             }
 
          signature.resize(signature_size);
-         return C_SignFinal(session, signature.data(), &signature_size, return_value);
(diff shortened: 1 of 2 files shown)


dewyatt:
Actually I'm a little tired and it looks like this is actually a bug in botan and not the PKCS#11 module.
The first size is an estimate, which is allowed to be larger than the actual output, according to the spec (section 5.2). The vector needs to be resized after that second call.

Will update title and such.

dewyatt:
Copy & paste error on the title fixed :P, I should get some sleep before I hurt myself.

randombit:
@neusdan can you review?

dewyatt:
It looks like there are similar bugs in other places, so I'll add some commits here and update the title.

PR #1980, opened 2019-05-27
Unnecessary lambda parameter capture fix

devnexen:


devnexen:
with win build, I understand better why it has been like this ...

randombit:
Yeah. Disappointing this hasn't been fixed even in VC2019 :/

PR #1997, opened 2019-06-16
clock_gettime supported by mac since 10.12

devnexen:


randombit:
Nice thanks. AFAICT 10.11 is end of life now so probably safe to enable it by default.

PR #2012, opened 2019-06-30
Enable more LGTM alerts

randombit:


randombit:
LGTM has a default set that runs for each language, along with others that are not included by default. Here I'm trying to enable tests that are not in the default set. But I also want to exclude the fixme-comment alert which is enabled by default.

randombit:
Modified on master

PR #1916, opened 2019-04-28
makefile.in: move LDFLAGS after EXE_LINKS_TO

ffontaine:
Move LDFLAGS after EXE_LINKS_TO to allow the user to give additional
libraries such as -latomic otherwise static build will fail:

cli/tls_server.o build/obj/cli/tls_utils.o build/obj/cli/utils.o build/obj/cli/x509.o -L. -lbotan-2 -lboost_filesystem -lboost_system -lbz2 -lcrypto -llzma -lsqlite3 -lz -o botan
[...]

PR #2014, opened 2019-07-02
Fix android build

Bjoe:
I try to create a [botan hunter package](https://github.com/ruslo/hunter/issues/1897) but I have some issues to build botan for android.

The first [compile issue: cannot find -lrt](https://travis-ci.org/Bjoe/hunter/jobs/553046275#L1294) is solved when I removed `linux -> rt` in the `src/lib/utils/socket/info.txt` file. But I still don't know if this is the right solution!

The second [compile issue: .. undefined reference to ...](https://travis-ci.org/Bjoe/hunter/jobs/553440703#L1315) is solved when I use the same approach like in pr #1916 

With this both fixes, the [android build passed](https://travis-ci.org/Bjoe/hunter/jobs/553476099) :-)

### DIFF
--- a/src/build-data/makefile.in
+++ b/src/build-data/makefile.in
@@ -70,7 +70,7 @@ TESTOBJS = %{join test_objs}
 # Executable targets
 
 $(CLI): $(LIBRARIES) $(CLIOBJS)
-	$(EXE_LINK_CMD) $(ABI_FLAGS) $(LDFLAGS) $(CLIOBJS) $(EXE_LINKS_TO) %{output_to_exe}$@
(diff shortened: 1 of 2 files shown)


randombit:
First I want to say: thank you. The Android build is definitely kind of a 2nd class citizen right now and any improvements are very welcome.

If you are building for Android then you should certainly use `--os=android` in which case any Linux specific paths are avoided. [I know Android is Linux, kind of - except that it's not in all kinds of ways that are relevant to us.] From your build log it looks like the Android build is being invoked with `--os=linux`. The socket lib doesn't try to link with librt on Android so I think if you ensure your package uses `--os=android` as appropriate then modifying the info.txt file is unnecessary.

The change to the makefile template seems good and mirrors the change made in #1916

Bjoe:
Ok, that's fast ;-) ... I'm still testing ... but ok. If I get in trouble, I will create a now pr :-)

PR #2016, opened 2019-07-03
Android docker build proposal.

devnexen:
Using external image just for the NDK and focusing on the two
most popular architectures.

### DIFF
--- a/src/scripts/Dockerfile.android
+++ b/src/scripts/Dockerfile.android
@@ -1,0 +1,22 @@
+FROM debian:stretch AS debian
+ARG SDK
+RUN apt-get update && apt-get -y --no-install-recommends install \
+	libboost-all-dev \
+	make \
+	python3-sphinx
+
+RUN mkdir -p /botan/android
+WORKDIR /botan
+COPY --from=bitriseio/android-ndk /opt/android-ndk android-ndk
+COPY configure.py configure.py
+COPY src src
+COPY doc doc
+COPY license.txt license.txt
+COPY news.rst news.rst
+ENV PATH=$PATH:android-ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/
+RUN ./configure.py --prefix=android/arm --os=android --cpu=arm --cc=clang --cc-bin=armv7a-linux-androideabi${SDK}-clang++ --ar-command=arm-linux-androideabi-ar --with-sphinx
+RUN make -j`getconf _NPROCESSORS_ONLN`
+RUN make install && make clean
+RUN ./configure.py --prefix=android/arm64 --os=android --cpu=arm64 --cc=clang --cc-bin=aarch64-linux-android${SDK}-clang++ --ar-command=aarch64-linux-android-ar --with-sphinx
+RUN make -j`getconf _NPROCESSORS_ONLN`
+RUN make install
--- a/src/scripts/Dockerfile.android
+++ b/src/scripts/Dockerfile.android
@@ -3,8 +3,7 @@ ARG ANDORID_SDK_VER
 ARG ANDROID_ARCH
 ARG ANDROID_TOOLCHAIN_SUF
 RUN apt-get update && apt-get -y --no-install-recommends install \
-	make \
-	python3-sphinx
+	make
 
 RUN mkdir -p /botan/android
 WORKDIR /botan
@@ -15,6 +14,6 @@ COPY doc doc
 COPY license.txt license.txt
 COPY news.rst news.rst
 ENV PATH=$PATH:android-ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/
-RUN ./configure.py --prefix=android/arm --os=android --cpu=${ANDROID_ARCH} --cc=clang --cc-bin=armv7a-linux-android${ANDROID_TOOLCHAIN_SUF}${SDK}-clang++ --ar-command=arm-linux-android${ANDROID_TOOLCHAIN_SUF}-ar --with-sphinx
+RUN ./configure.py --prefix=android/arm --os=android --cpu=${ANDROID_ARCH} --cc=clang --cc-bin=armv7a-linux-android${ANDROID_TOOLCHAIN_SUF}${SDK}-clang++ --ar-command=arm-linux-android${ANDROID_TOOLCHAIN_SUF}-ar
 RUN make -j`getconf _NPROCESSORS_ONLN`
 RUN make install
--- a/src/scripts/Dockerfile.android
+++ b/src/scripts/Dockerfile.android
@@ -1,7 +1,8 @@
 FROM debian:stretch AS debian
-ARG SDK
+ARG ANDORID_SDK_VER
+ARG ANDROID_ARCH
+ARG ANDROID_TOOLCHAIN_SUF
 RUN apt-get update && apt-get -y --no-install-recommends install \
-	libboost-all-dev \
 	make \
 	python3-sphinx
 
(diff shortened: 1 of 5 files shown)


devnexen:
I left on purpose the x86 family but I can add them if wished.

randombit:
Scripts (and docker configs I guess) should be in `src/scripts`. Toplevel dir especially should be kept clean.

I generally prefer Python for anything scripting but shell here probably makes sense.

This is probably good as a start, at least it makes it possible to easily test locally. I think what I'd really like to see is a single Docker image that we can publish on Docker hub that contains the NDK + everything else we currently install for CI (qemu, cross compilers, clone of https://github.com/randombit/botan-ci-tools, etc), and then configure Travis to run all CI builds within that container.

OK with files moved to `src/scripts`

devnexen:
No problems. Will wait your approval ;-)

src/scripts/Dockerfile.android
```diff
@@ -0,0 +1,22 @@
+FROM debian:stretch AS debian
```
webmaster128: What about deriving from `bitriseio/android-ndk` directly? If there is anyone in the industry I'd trust for maintaining an Android build image, it is the bitrise team

randombit:
Docker is taking forever to download the images on my slow internet so I haven't been able to test this. But it's a start and can't break anything so I'll merge now.

src/scripts/Dockerfile.android
randombit: I have no opinion on this. @devnexen wdyt?

devnexen: I ve done that in another project and thought it was a bit ancient but that s a fair point maybe that would just work in botan's case.

webmaster128:
Good job! Could you add an Android build to the Travis build matrix? Then we can see in this PR build if it works

devnexen:
Good point, will try if necessary I admit I m not a travis wizard :-)

webmaster128:
> I m not a travis wizard :-)

No worries, there are travis wizards around happy to help out

src/scripts/ci/travis.yml
```diff
@@ -88,6 +93,9 @@ matrix:
     - os: linux
       compiler: clang
       env: BUILD_MODE="cross-arm64"
+    # No docker build on osx
+    - os: osx
+      env: BUILD_MODE="docker-android"
```
randombit: This should also exclude with gcc since NDK only uses Clang.

devnexen: In fact the docker build uses the clang binaries from the ndk, not the system one but maybe it is really necessary in travis case.

src/scripts/ci/travis.yml
```diff
@@ -34,6 +38,7 @@ env:
     - BUILD_MODE="nist"
     - BUILD_MODE="docs"
     - BUILD_MODE="lint"
+    - BUILD_MODE="docker-android"
```
randombit: This should get moved up (maybe to just after the other cross builds). Travis does 5 builds in parallel, in the order specified here. If any of the final builds take a long time it ends with just 1 build running after everything else has completed, which lengthens the overall CI build time.

src/scripts/Dockerfile.android
```diff
@@ -0,0 +1,22 @@
...
+RUN mkdir -p /botan/android
+WORKDIR /botan
+COPY --from=bitriseio/android-ndk /opt/android-ndk android-ndk
+COPY configure.py configure.py
+COPY src src
+COPY doc doc
+COPY license.txt license.txt
+COPY news.rst news.rst
+ENV PATH=$PATH:android-ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/
+RUN ./configure.py --prefix=android/arm --os=android --cpu=arm --cc=clang --cc-bin=armv7a-linux-androideabi${SDK}-clang++ --ar-command=arm-linux-androideabi-ar --with-sphinx
```
randombit: For running this in CI might be best to take an additional ARG which specifies aarch64 vs armv7, then splitting it into two Travis builds.

src/scripts/Dockerfile.android
```diff
@@ -0,0 +1,22 @@
+FROM debian:stretch AS debian
+ARG SDK
+RUN apt-get update && apt-get -y --no-install-recommends install \
+	libboost-all-dev \
```
randombit: Is boost being used here? I don't think we really need it in this context since mostly it just enables asio support in the CLI utils.

devnexen: That s a good point, the lesser dependencies the better.

src/scripts/Dockerfile.android
devnexen: I would need in this case to handle clang binary name difference between those two arches too.

src/scripts/ci/travis.yml
```diff
@@ -11,9 +11,13 @@ compiler:
   - clang
   - gcc
 
+services:
+  - docker
+
 env:
   global:
     - CCACHE_MAXSIZE=800M
+    - SDK=21
```
randombit: For just invoking the script on the side a short var name like `SDK` was fine but setting it here it is not obvious what it controls. Rename to `ANDROID_SDK_VER` or something.

src/scripts/ci/travis.yml
randombit: The problem is right now there are two builds which are identical except for g++ vs clang. https://travis-ci.org/randombit/botan/jobs/554123055 https://travis-ci.org/randombit/botan/jobs/554123058

As you say it is kind of irrelevant which is used, since in either case we ignore CXX, but it would be best to have the Android build run as a Clang build so it is easy to see on the front page what compiler is being used, since it is *a* Clang that's doing the work.

src/scripts/Dockerfile.android
```diff
@@ -0,0 +1,20 @@
+FROM debian:stretch AS debian
+ARG ANDORID_SDK_VER
+ARG ANDROID_ARCH
+ARG ANDROID_TOOLCHAIN_SUF
+RUN apt-get update && apt-get -y --no-install-recommends install \
+	make \
+	python3-sphinx
```
randombit: Similarly I don't think we need Sphinx here, that's what the `docs` Travis build is for.

src/scripts/ci/travis.yml
```diff
@@ -88,6 +94,12 @@ matrix:
     - os: linux
       compiler: clang
       env: BUILD_MODE="cross-arm64"
+    # No docker build on osx
+    - os: osx
+      env: BUILD_MODE="docker-android"
+    # Nor for gcc's
+    - compiler: gcc
+      env: BUILD_MODE="docker-android"
```
randombit: I think these are not being excluded from the build because now env contains multiple values and Travis wants to see a complete match of the env for an exclusion.

I think the best fix is

ANDROID_ARCH=arm ANDROID_TOOLCHAIN_SUF=eabi BUILD_MODE="docker-android" -> BUILD_MODE="android-arm7"

ANDROID_ARCH=arm64 BUILD_MODE="docker-android" -> BUILD_MODE="android-aarch64"

Then exclusions work.

However, we now need to get the relevant env vars down to the docker script. I'm going to push a change adding that to the CI build script in this branch.

randombit:
So, the good news is the build from your initial attempt does seem to be working https://travis-ci.org/randombit/botan/jobs/554123058

The problem is, the build has been running now for almost 30 minutes. Which to put it in context, currently *all* the Travis builds finish in about 1h20m, and usual start-to-finish time is about 30 minutes.

First, running it within docker means we can't (AIUI) use ccache to speed up the builds. That's bad news since it means we always recompile everything. For the vast majority of PRs ccache has a very high hit rate.

Also, pulling the docker images seemed to take quite a while. (I'm glad it's not just my network connection.)

I think this Docker script is good as a demo and to make it easy for an end user to generate binaries but doesn't work well for CI. For CI instead we just should wget https://dl.google.com/android/repository/android-ndk-r20-linux-x86_64.zip during the Travis setup script and invoke it directly, which allows using ccache and (presumably) will be faster than redownloading most of Ubuntu via Docker.

Since the NDK version doesn't change very often we could even stash the NDK zip in the Travis build cache and save the download.

devnexen:
Would only revert the .travis part be good enough for you ? Maybe working on travis in a separated branch ?

randombit:
Yes I'm good with the script as you initially posted, and it will be useful for end users. Sorry for sending you down this rabbithole. I'm off work today (Independence Day here in US) so I'm working on the Travis side now.

devnexen:
Np :-) was worth a shot.

src/scripts/ci_build.py
```diff
@@ -438,6 +438,9 @@ def main(args=None):
         if use_python3 and options.use_pylint3:
             cmds.append(['python3', '-m', 'pylint'] + pylint_flags + [py3_flags] + full_paths)
 
+    elif target == 'docker-android':
```
randombit: This should be reverted too

PR #2017, opened 2019-07-04
Build Android via NDK

randombit:


src/scripts/ci_build.py
```diff
@@ -141,6 +143,28 @@ def determine_flags(target, target_os, target_cpu, target_cc, cc_bin, ccache, ro
                 flags += ['--cpu=arm64', '--cc-abi-flags=-arch arm64 -stdlib=libc++']
             else:
                 raise Exception("Unknown cross target '%s' for iOS" % (target))
+        elif target_os == 'android':
+
+            if os.getenv('ANDROID_NDK') is None:
+                raise Exception('Android CI build requires ANDROID_NDK env variable be set')
+            if os.getenv('ANDROID_API_LEVEL') is None:
```
devnexen: Would a "sane" default be relevant eventually, knowing the ndk version ?

devnexen:
Should work as is ... I admit it s not super easy to automate android ndk builds :-)

src/scripts/ci_build.py
randombit: Now defaults to 16 for 32-bit and 28 for 64-bit, so we are testing both compat features (like your impl of getauxval for < 18) and latest shiny (like getentropy added in 28).

devnexen:
I think getentropy needs sys/random.h as mac. the api level is correct to me.

PR #2020, opened 2019-07-05
Remove BearSSL provider

randombit:
Reverting #1094 

Comparing the situation with the OpenSSL provider:

* OpenSSL is (sometimes) faster
* OpenSSL or a compatible fork is installed everywhere

BearSSL

* Seems to be much slower
* Is not included in any distribution I know of
* As a result of not being distributed, the integration isn't tested in CI either

BearSSL has good qualities and makes sense for bootloader or small embedded systems. But in such environments it doesn't make sense to bring along Botan for the ride. And in environments where Botan fits, BearSSL doesn't bring anything to the table.

CC @bluerise


bluerise:
It was very helpful to have Botan as a provider to test if BearSSL is doing its job correctly. But I understand and I'm fine with it being removed, I can live with it.

PR #2023, opened 2019-07-05
Skip allocation if OctetString is constructed with empty string.

LooseGears:
`explicit OctetString(const std::string& str = "");` called with default param.

results in `m_data.resize(1);`

This change just skips that allocation.


randombit:
Change is good. Can you fix indentation?

LooseGears:
Sure, I tried to follow the indentation style but looks like I got it wrong.

Look good?

randombit:
Perfect thank you

PR #2025, opened 2019-07-07
[WIP] locked pages transparent huge page support on Linux.

devnexen:
It is assumed the consumer knows its system settings and
huge page size supported. Needs to remain an option for the
vast majority of cases the page size is sufficient.

### DIFF
--- a/src/build-data/buildh.in
+++ b/src/build-data/buildh.in
@@ -128,7 +128,11 @@
 * Total maximum amount of RAM (in KiB) we will lock into memory, even
 * if the OS would let us lock more
 */
-#define BOTAN_MLOCK_ALLOCATOR_MAX_LOCKED_KB 512
+#if !defined(BOTAN_TARGET_OS_HAS_HUGETLB)
+  #define BOTAN_MLOCK_ALLOCATOR_MAX_LOCKED_KB 512
+#else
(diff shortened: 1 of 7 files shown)


randombit:
I'm not sure this is a good idea. For one, most Linux systems default to 64 KiB of `mlock`able memory for unprivileged users. So at the least one would have to modify `/etc/security/limits.conf` to enable at least one large page size worth of lockable data. I suspect what's happening right now is, mlock fails and the pool is simply not used.

Also this breaks `BOTAN_MEM_POOL_USE_MMU_PROTECTIONS` which uses `mprotect` to disable access to pages on the free list. `Memory_Pool` is still working the basis of the normal system page size (since that's what is passed to the constructor in `locked_allocator`) and it's not possible to simply pass the huge page size there instead because `Memory_Pool` splits up buckets on the basis of the page size. (Ie, each page is only used to allocate objects of 1 particular size, so if you pass just 1 large page then the allocator ends up being usable only for one size of allocation, basically whatever is requested first).

A single large page also reduces the benefit of having guard pages.

All of this might be resolvable but I don't see easy solutions.

devnexen:
Yes I was just toying with the idea  to see where it would lead more about perf in few cases than security ... in my own set clock does not fail but I admit my own case is not necessarily the most spread. I was just thinking about the bucket allocation type issue but did not come up with something too much breaking ...

PR #2026, opened 2019-07-11
Add support for POWER9 DARN RNG

randombit:


randombit:
Further testing with this shows some very bad behavior, namely repeated outputs:

```
        for(size_t i = 0; i != 32; ++i)
          {
            const uint64_t d = __builtin_darn();
            printf("%d %016llX\n", i, d);
            rng.add_entropy_T(d);
          }
```

->

```
0 57DF91918C8793AD
1 57DF91918C8793AD
2 57DF91918C8793AD
3 57DF91918C8793AD
4 57DF91918C8793AD
5 57DF91918C8793AD
6 57DF91918C8793AD
7 57DF91918C8793AD
8 57DF91918C8793AD
9 57DF91918C8793AD
10 57DF91918C8793AD
11 57DF91918C8793AD
12 57DF91918C8793AD
13 57DF91918C8793AD
14 57DF91918C8793AD
15 57DF91918C8793AD
16 57DF91918C8793AD
17 57DF91918C8793AD
18 57DF91918C8793AD
19 57DF91918C8793AD
20 57DF91918C8793AD
21 57DF91918C8793AD
22 57DF91918C8793AD
23 57DF91918C8793AD
24 57DF91918C8793AD
25 57DF91918C8793AD
26 57DF91918C8793AD
27 57DF91918C8793AD
28 57DF91918C8793AD
29 57DF91918C8793AD
30 57DF91918C8793AD
31 57DF91918C8793AD
```

I don't know if this a compiler bug or CPU bug but in any case this looks pretty sketchy. Will revisit if we can find more info and a resolution.

randombit:
It looks to be a GCC problem - if I use inline asm instead of the builtin it seems to work ok

Issue #424, opened 2016-02-15
Code style template


neusdan:
It would be really really nice to have a clang-format style (#284) and of course one that is more optimized for readability than the current one.

Until then i've found a solution to format the code with the current style with [Astyle](http://astyle.sourceforge.net/astyle.html)

`--style=whitesmith --indent=spaces=3 --align-pointer=type --align-reference=type --max-code-length=120 --attach-namespaces --indent-preproc-block --indent-preproc-define --unpad-paren --keep-one-line-blocks --keep-one-line-statements --mode=c`

randombit:
Re the reformat, I'm waiting for a quiet release where ideally the reformat can be the only major change. I'm also planning to use clang-tidys fixit mode to add braces around all single-statement blocks at the same time. clang-tidy can insert {} but does not try to match any particular indentation style so needs a format tool to help out.

I was kind of frustrated that clang-format insists on mangling closing namespace brackets (this seems to not be controllable via any clang-format toggle I could find). But then I decided it does not matter that much :P

al45tair:
Have you considered e.g.

```c++
#define BEGIN_BOTAN_NS namespace Botan {
#define END_BOTAN_NS   }
```

Has the added benefit of making many editors' auto indent work the way you'd hope...

ksanderon:
sorry, that I'm  necormancer - with clang-format is not possible (in 99%) to have whitesmith style even "strict allman" is sometimes hard - mostly regarding lambda functions. For more I not recommend it because of major changes in format style between versions - if rules are more complex than just:
```
BasedOnStyle: XYZ
Standard: C++XY
IndentWidth: cde
ColumnLimit: fgh
```

However it work well with complex templates ;)

From other hand astyle should do the job in botan and not break to much in macros and templates, because botan simply don't have them too much / complicated.

Maybe after succesful "code reformat" it will be worth to add to CI job which will run something like this during PR:
```
find * -iname '*.h' -o -iname '*.cpp' | xargs clang-format -i -style=file
git status --porcelain > ../status.log
git diff
test ! -s ../status.log
```
just to keep code style.

webmaster128:
> Maybe after succesful "code reformat" it will be worth to add to CI job which will run something like this during PR:

Something like that could nicely be added to the "lint" job (see Travis matrix) as it does not require compilation

randombit:
As of #3502 `clang-format` is now in use for Botan3

PR #2037, opened 2019-07-16
Some OS:: cleanups/changes

randombit:


src/lib/utils/os_utils.cpp
```diff
@@ -206,32 +206,28 @@ uint64_t OS::get_cpu_cycle_counter()
...
-   if ((res = ::sysconf(_SC_NPROCESSORS_CONF)) <= 0)
-     {
-          return 1;
-     }
-     tt = static_cast<size_t>(res);
-#elif defined(BOTAN_TARGET_OS_HAS_THREADS)
-     tt = static_cast<size_t>(std::thread::hardware_concurrency());
-#endif
-     return tt;
+   const long res = ::sysconf(_SC_NPROCESSORS_CONF);
```
devnexen: I see you rarely use `auto` (maybe for clarity's sake) so I just followed but nothing against.

devnexen:
Was thinking if a clang format config would be useful with this PR.

src/lib/utils/os_utils.cpp
randombit: Yeah in general I only use `auto` when the type name is either very long (`std::map<...>::const_iterator`) or is actually untypable (lambda, based on template param, ...).

randombit:
Re code style there is an open issue #424 actually the oldest unresolved issue. It has been a while since I looked at it. What I wanted was something that gets within ~90% of the current codestyle and apply the formatter across the codebase, then we could from there consider future format changes. I was not able to get there with clang-format at the time, but I imagine it has improved in last years. There is an `astyle` template posted in that issue that gets pretty close though (also in `src/configs/astyle.rc`). But anyway I think a code reformat should happen in its own PR with no other (functional) code changes.

PR #2043, opened 2019-07-18
Try enabling threads on MinGW

randombit:
Seems like Debian/Ubuntu does something weird here.


randombit:
Apparently that doesn't work either oh well

PR #2050, opened 2019-07-27
Darwin, monitoring anonymous locked pages via vm_stat tool.

devnexen:
Userland applications are allowed taking ID from 240->255.
The ID can be made dynamic later on if wished.

### DIFF
--- a/src/lib/utils/os_utils.cpp
+++ b/src/lib/utils/os_utils.cpp
@@ -430,11 +430,19 @@ std::vector<void*> OS::allocate_locked_pages(size_t count)
    #define PROT_MAX(p) 0
 #endif
       const int pflags = PROT_READ | PROT_WRITE;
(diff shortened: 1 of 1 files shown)


devnexen:
Allows to display values as
`Memory Tag 255                    1024K    1024K    1024K       0K       0K       0K       0K      256`

src/lib/utils/os_utils.cpp
```diff
@@ -55,6 +55,12 @@
   extern "C" char **environ;
 #endif
 
+#if defined(BOTAN_TARGET_OS_IS_IOS) || defined(BOTAN_TARGET_OS_IS_MACOS)
+  #define LOCKED_PAGES_FD (255<<24)
+#else
+  #define LOCKED_PAGES_FD -1
```
randombit: Use `const int locked_fd = ...` instead of a define here, and add a comment explaining meaning of/use for the Apple value. LGTM otherwise

PR #2051, opened 2019-07-30
Add Private_Key::stateful_operation and use it in cli signer

randombit:
If you generated an XMSS key and used it in the CLI, the key would not be updated which is bad news for stateful signatures.

@mgierlings did I get this right? Is there anything else that is needed to do for updating the key?

### DIFF
--- a/src/cli/pubkey.cpp
+++ b/src/cli/pubkey.cpp
@@ -21,6 +21,8 @@
 #include <botan/workfactor.h>
 #include <botan/data_src.h>
 
+#include <fstream>
+
 #if defined(BOTAN_HAS_DL_GROUP)
    #include <botan/dl_group.h>
 #endif
@@ -215,6 +217,15 @@ class PK_Sign final : public Command
          this->read_file(get_arg("file"), onData);
 
          output() << Botan::base64_encode(signer.signature(rng())) << "\n";
+
+         if(key->stateful_operation())
+            {
+            std::ofstream updated_key(key_file);
+            if(passphrase.empty())
+               updated_key << Botan::PKCS8::PEM_encode(*key);
+            else
+               updated_key << Botan::PKCS8::PEM_encode(*key, rng(), passphrase);
+            }
          }
    };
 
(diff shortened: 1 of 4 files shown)


src/cli/pubkey.cpp
```diff
@@ -215,6 +217,15 @@ class PK_Sign final : public Command
          this->read_file(get_arg("file"), onData);
 
          output() << Botan::base64_encode(signer.signature(rng())) << "\n";
```
mgierlings: At this point a valid signature with the current leaf index is generated and immediately written to stdout...

src/cli/pubkey.cpp
```diff
@@ -215,6 +217,15 @@ class PK_Sign final : public Command
...
+
+         if(key->stateful_operation())
+            {
+            std::ofstream updated_key(key_file);
+            if(passphrase.empty())
+               updated_key << Botan::PKCS8::PEM_encode(*key);
```
mgierlings: If the application crashes i.e. due to an out of memory error, power outage etc. between [L219](https://github.com/randombit/botan/blob/db1b06d2ee44a1bc72de3bdd4c4098cc8c149dcc/src/cli/pubkey.cpp#L219) and the key update here the signature might have been used or stored already but the key has not been updated. Once the application is restarted the next signature operation will reuse the leaf index breaking XMSS's security.

src/cli/pubkey.cpp
```diff
@@ -215,6 +217,15 @@ class PK_Sign final : public Command
...
+            else
+               updated_key << Botan::PKCS8::PEM_encode(*key, rng(), passphrase);
+            }
          }
```
mgierlings: The signature output should be generated here instead of in [L219](https://github.com/randombit/botan/blob/db1b06d2ee44a1bc72de3bdd4c4098cc8c149dcc/src/cli/pubkey.cpp#L219). This has been corrected in #2069.

randombit:
#2069

randombit:
Thanks @mgierlings. With #2069 merged into this branch are we ok here?

PR #2057, opened 2019-08-04
Clean up OID code and internal APIs

randombit:
There is more that needs to done in this area, esp the code for parsing X.509 signature algorithms is old and a mess with lots of duplicated logic. Pushing that off to a future PR.


randombit:
Merged in 4f5abd1 but forgot to pull had to apply last revision again on master.

PR #2058, opened 2019-08-05
The allocator cb can possibly return null

devnexen:


randombit:
You're right. Even before I added an explicit return null in 402b727, `calloc` is allowed to return null in that situation.

PR #2061, opened 2019-08-06
Deprecate many publically available headers

randombit:
Currently Botan has 300 (!) public headers.

This change deprecates public access to many of these headers, which in a future (major) release will bring the total header count down to 'just' 161. Probably this can/will be trimmed further in future PRs.

The benefits are several:
 * Reducing the number of APIs which a developer must learn (which of the 300 headers has the functionality you want?)
 * Freeing us to change (now internal) APIs whenever useful.
 * Reducing ABI surface. We typically have to increment the soname with each release, sometimes due to trivial reasons like a single struct changing size. Hiding as much as possible reduces soname churn. This in particular would make it easier for updates to get into Debian and downstream distros since they don't like soname bumps.
 * Reducing soname exports. If the compiler knows a symbol is not exported (and thus cannot be overridden eg via `LD_PRELOAD`) then it can optimize somewhat better. This will be a bit marginal since a lot of functionality is already exported (only) for tests via internal headers.

There are a few major categories of headers so deprecated:
 * Algorithm headers like ``aes.h``, ``gcm.h``, ``kdf2.h``, etc. Instead the application should use ``X::create``. This does not apply to public key headers since there is some functionality only available via direct call right now, namely keygen and accessing fields. But we could look at that in the future, since that could reduce the header count by a further 20 or 30.
 * The EME/EMSA headers are entirely deprecated, even the base class - there doesn't seem to be any reason for an application to instantiate these.
 * Utility headers like ``bswap.h``, ``rotate.h``,  ``cpuid.h``, ``parsing.h`` - these are used to implement the library but aren't part of our core functionality offering.
 * Implementation detail headers like ``xmss_common_ops.h``, ``divide.h``, ``keypair.h`` which an application has no obvious reason to use directly.

### DIFF
--- a/doc/deprecated.rst
+++ b/doc/deprecated.rst
@@ -1,96 +1,25 @@
 Deprecated Features
 ========================
 
-The following functionality is currently deprecated, and will likely
-be removed in a future release. If you think you have a good reason to
-be using one of the following, contact the developers to explain your
-use case if you want to make sure your code continues to work.
+Certain functionality is deprecated and is likely to be removed in
+a future major release.
 
-This is in addition to specific API calls marked with BOTAN_DEPRECATED
-in the source.
+To help warn users, macros are used to annotate deprecated functions
+and headers. These warnings are enabled by default, but can be
+disabled by defining the macro ``BOTAN_NO_DEPRECATED_WARNINGS`` prior
+to including any Botan headers.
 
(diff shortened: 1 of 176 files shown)


randombit:
https://lists.randombit.net/pipermail/botan-devel/2019-August/002286.html

PR #2063, opened 2019-08-08
Some thoughts on the Python module installation

risa2000:
I have made some testing of the different ways of installing the module (well, the one file module) and come up with this PR as a base for a review/discussion on how to proceed.

The standard way is to use `setup.py` (I added one with some tangible values, others need to be completed). Now even with `setup.py` there are basically three ways how to use it:

1) `distutils` = ancient way, no longer supported or recommended
2) `setuptool` = less ancient, more supported, should work OK
3) `pip` = current way, fully supported and promoted

While `pip` is usually associated with installation of web packages, it can perfectly work on local ones:
```python
pip install ./src/python
```
when called from the project root will install the package (as proposed in this PR). It can also be executed from the script:
```python
import subprocess
import sys

def install(package):
    subprocess.call([sys.executable, "-m", "pip", "install", package])

install(r'./src/python')
```
The advantage is that everything else automatically works - it is listed in `pip list`, can be uninstalled, or upgraded (this does not work if the package is installed via distutils).
There is even a handy feature for development:
```python
pip install -e ./src/python
```
which installs only a "link" into site-packages and the actual code will run from the original place (i.e. you can work on it and it is automatically pulled everytime you `import botan2`).

Now, (maybe) disadvantage is that the user needs `pip` (and `setuptools`) installed. Not sure if it would be a hurdle for `botan`'s audience?

The other point to consider is whether run the installation automatically, or let user run it (it is relatively straightforward as described above). The advantage is that it gives the control over all `pip` options, so user can choose where and how exactly to install the package (in site dir, virtual env, etc.)

The `setup.py` can also be used directly with `setuptools` or, when modified a bit, with `distutils`, but it does not offer really any advantage expect not being dependent on `pip` and the comfort is lower.

I did not figure out how to instruct `pip` to pull in `readme.rst` from the root dir, so I had to copy it there, probably could be done automatically by some other script. Also, I guess the script will need some additional work to correctly import the project version.

(I will be away from keyboard for a week or two).


randombit:
Thanks for looking into this. But I don't think we would want to go this route. In particular there are many systems that do not have `pip` (or even `distutils`) installed, since they are not part of the base Python install. We absolutely don't want to block installing Botan itself (which is primarily a C++ library) just because some Python tool which is not installed by default is missing. I don't think we even want to block installing the Python module just because `pip` is missing. Also for the sake of Linux package managers we *don't* want to install via `pip` in some cases, because the installation of `botan2.py` should be "sticky" with the main library package; if you remove the botan package it should remove the Python wrapper as well, but here it would be managed via `pip` instead which the package manager would not know about.

The Python wrapper is kind of in a funny place due to it being included with the main source tree. The Rust, Ruby and OCaml wrappers, which are distributed separately, make use of their languages respective package management systems. If Python was also separate, then it would make sense for it to use `pip` or the like. We could such a split, but one key difference is Python is already required to build the library, and so presumably is available on the host system. [And a big reason for the Python module is to act as a testbed/testcase for the C wrapper, so it is useful for development purposes to keep it in the main repo.]

This proves unfortunate for people who are wanting to program in straight Python. It probably *mostly* affects people on Windows since on Unix/macOS installing `botan` via the standard package managers should get them the Python wrapper as a side effect. So maybe the resolution is, make it easier to get Botan installed on Windows? [And make sure `botan2.py` ends up in the right place, since that probably is not behaving correctly on Windows.]

Something else that might work ok is to add an option to `configure.py` like `--install-python-module-via=` which could take options `no` (don't install), `copy` (current logic) or `pip` (something like what you've done here). Then if in `pip`-mode the install script could create a temp directory for pip (copy `botan2.py` to `botan2/__init__py`, copy readme, copy `setup.py` from `build-data`, then invoke `pip`). It would be a bit more work... and I'm not entirely positive this solves the problem you're having.

risa2000:
I guess the difference is in the perspective.

I understand that on Linux, the system package manager can copy `botan2.py` to `site-packages` while installing the native libraries and it is perfectly sufficient. The packaging system records the metadata about the Python file and when the package is uninstalled or upgraded the file is handled accordingly together with its binary counterpart. 

From the Python perspective though, copying a Python script into `site-packages` is just a "hack", which just conveniently uses the fact that `site-packages` are on `PYTHONPATH` and as such anything in there is directly available for `import`.

Apart from that there is no metadata in Python's own packaging system, so strictly speaking, the file is not installed. It would work exactly the same way if the file was put in some other place (on PYTHONPATH), or directly into the application folder, etc.

While on Linux Python is usually part of the system install and as such is also kept up to date, and follows some "standard rules" on Windows Python is a "standalone app", which has no system dependencies, can be of any version and type and installed anywhere as would be the most convenient to the user. There is no equivalent for `/usr/lib` or `/usr/local/lib` either. So trying to guess where to copy `botan2.py` does not really make sense as it can be anywhere.

Since there is no `botan` Python "package" for Windows and only way to install `botan` on Windows seems to be to build it locally, I guess it would make sense to create a Python package for Windows users, ie. a binary package with the DLL _and_ the Python file), but it seems it is not really needed (people do not use it in Python on Windows, or build their own?). 

For someone like me, who build the lib from the source and at the same time wants to use it in Python on Windows, I can do exactly what the installer does on Linux, i.e. copy the file to PYTHONPATH and it will be no better and no worse than what the Linux counterpart does.

It does not make sense to have it done by installer, which is targeting a particular Linux directory structure, though.

It does not make sense to try to figure out where `site-packages` are and write it there either, because it has no advantage over doing it manually, while forcing the particular config on the user, who may want to put it somewhere else (in virtual env, or simply somewhere else than into system protected `site-packages`). 

I believe that the way it is handled now is really only fitting Linux usage (or Linux-like systems) and eventually handling the Python bindings as proper Python package is the right way to go, but I guess until there are more Windows users asking for it, it is not worthy the effort.

Anyway the current state is not a problem which needs to be solved (for me), so you can close this PR.

PR #2073, opened 2019-08-18
Add remove function to Botan::Extension class

ahura24:
Hello and thanks for your awesome library.

We need to remove an extension from the Botan::Extensions class. This patch fix the bug of Botan::Extensions::replace function when throw and exception at create time of Extensions_Info structure and add a "remove" function for this class.

Regards,
[Add_remove_function_to_extensions.zip](https://github.com/randombit/botan/files/3512801/Add_remove_function_to_extensions.zip)


randombit:
Taken up in #2101

Issue #2086, opened 2019-08-30
Make it possible to disable cli from the build


randombit:
See #2085 from background

randombit:
Maybe generalize it to `--build-targets=` which can accept options `static`, `shared`, `cli`, and `tests`. I believe there is an open issue about how disable-shared, disable-static is confusing especially for Windows targets.

reneme:
> I believe there is an open issue about how disable-shared, disable-static is confusing 

I guess you mean [this one](https://github.com/randombit/botan/issues/2040)?

reneme:
There are already switches like `--build-fuzzers` and `--build-bogo-shim`. In order to keep the default behaviour stable (i.e. the CLI and tests are built), we would need to add "negative" switches like `--no-build-cli` and `--no-build-tests`. That feels inconsistent.

I suggest to go with (something like) `--build-extra-targets=` (accepting a list of `cli`, `tests`, `fuzzers`, `bogo`) that defaults to `cli,tests` and strive to deprecate the above mentioned switches?

I would argue to keep `static` and `shared` out of this in a first step, to keep it simple.

randombit:
Done with #2098

PR #2098, opened 2019-09-05
Add configure parameter --build-targets=

reneme:
As a first step towards #2086, this introduces `--build-extra-targets=`. It allows to select additional build targets (apart from the actual library) during `./configure.py`. A minimal build of only the library, would be configured like so:

```
./configure.py --build-extra-targets="" --minimized-build
```

Todo:

- [x] sanity checks
- [x] dealing with `static` and `shared` library builds?
- [ ] ~~how to deal with fuzzer build targets? -- e.g. one target per fuzzer-type (i.e. `afl_fuzzer`, `libfuzzer`, `klee_fuzzer`, ...)~~
- [x] CI integration?

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -490,6 +490,9 @@ def process_command_line(args): # pylint: disable=too-many-locals,too-many-state
     build_group.add_option('--build-bogo-shim', action='store_true', default=False,
                            help=optparse.SUPPRESS_HELP)
 
+    build_group.add_option('--build-extra-targets', default='cli,tests', dest="build_targets",
+                           help="build additional targets and tools (cli, tests, bogo_shim, docs)")
+
     build_group.add_option('--with-pkg-config', action='store_true', default=None,
                            help=optparse.SUPPRESS_HELP)
     build_group.add_option('--without-pkg-config', dest='with_pkg_config', action='store_false',
@@ -1957,14 +1960,18 @@ def create_template_vars(source_paths, build_paths, options, modules, cc, arch, 
             return path
         return os.path.join(build_dir, path)
 
-    def all_targets():
+    def all_targets(options):
+        targets = [ t.strip().lower() for t in options.build_targets.split(',') ]
         yield 'libs'
-        yield 'cli'
-        yield 'tests'
-        yield 'docs'
+        if 'cli' in targets:
+            yield 'cli'
+        if 'tests' in targets:
+            yield 'tests'
+        if 'docs' in targets:
+            yield 'docs'
         if options.build_fuzzers:
             yield 'fuzzers'
-        if options.build_bogo_shim:
+        if options.build_bogo_shim or 'bogo_shim' in targets:
             yield 'bogo_shim'
 
     variables = {
@@ -1983,7 +1990,7 @@ def create_template_vars(source_paths, build_paths, options, modules, cc, arch, 
         'macos_so_compat_ver': '%s.%s.0' % (Version.packed(), Version.so_rev()),
         'macos_so_current_ver': '%s.%s.%s' % (Version.packed(), Version.so_rev(), Version.patch()),
 
-        'all_targets': ' '.join(all_targets()),
+        'all_targets': ' '.join(all_targets(options)),
 
         'base_dir': source_paths.base_dir,
         'src_dir': source_paths.src_dir,
--- a/configure.py
+++ b/configure.py
@@ -276,7 +276,7 @@ class BuildPaths(object): # pylint: disable=too-many-instance-attributes
         else:
             raise InternalError("Unknown src info type '%s'" % (typ))
 
-__acceptable_build_targets = ["static", "shared", "cli", "tests", "bogo_shim", "docs"]
+ACCEPTABLE_BUILD_TARGETS = ["static", "shared", "cli", "tests", "fuzzers", "bogo_shim", "docs"]
 
 def read_build_targets(options):
     if options.build_targets is None:   # not set => no list
@@ -500,7 +500,7 @@ def process_command_line(args): # pylint: disable=too-many-locals,too-many-state
                            help=optparse.SUPPRESS_HELP)
 
     build_group.add_option('--build-targets', default=None, dest="build_targets",
-                           help="build specific targets and tools (%s)" % ', '.join(__acceptable_build_targets))
+                           help="build specific targets and tools (%s)" % ', '.join(ACCEPTABLE_BUILD_TARGETS))
 
     build_group.add_option('--with-pkg-config', action='store_true', default=None,
                            help=optparse.SUPPRESS_HELP)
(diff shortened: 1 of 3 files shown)


randombit:
Nice. For future evolution I'd prefer this is named `--build-targets` so we can extend it to things like

--build-targets=static,shared,cli # build both libs
--build-targets=shared,tests # shared lib only

In the short term we can reject shared/static being specified here and continue to use disable-shared/disable-static to guide.

It should definitely sanity check at least that the things mentioned on the command line are targets we know about.

For fuzzers, we could support multiple fuzzer types in one build but I don't think we need it - KLEE doesn't actually work (doesn't support C++ last I checked though was a WIP), and AFL is less commonly used these days. And all new upcoming fuzzers just use the libFuzzer interface because it is convenient and does everything needed. Instead we should split up the current `--build-fuzzers=type` option to ``--fuzzer-interface=type`` plus ``--build-targets=...,fuzzers``.

A good start for testing this in CI would be to remove the --build-fuzzer and --build-bogo-shim options and instead run those through --build-targets, and update the CI script accordingly.

randombit:
An interesting thing we can do with this - if the tests are not being built then we don't need to export symbols which are only exported for testing purposes, ie define `BOTAN_TEST_API` to an empty string instead of a visibility attribute. [This doesn't have to happen as part of this PR.]

configure.py
```diff
@@ -2982,6 +2998,37 @@ def canonicalize_options(options, info_os, info_arch):
...
+        elif bool(options.build_shared_lib) != bool("shared" in build_targets):
+            raise UserError("inconsistent usage of --enable/disable-shared-library and --build-targets")
+
+        # building the static lib desired and without contradiction?
+        if options.build_static_lib == None:
+            options.build_static_lib = "static" in build_targets
+        elif bool(options.build_static_lib) != bool("static" in build_targets):
+            raise UserError("inconsistent usage of --enable/disable-static-library and --build-targets")
+
+    canonicalize_build_targets(options)
```
reneme: This turned out to be fairly complex, because it shall keep the default and legacy behaviour (using `--(enable/disable)-(shared/static)-library`)  and consistently introduce `--build-targets=(static/shared)`.

The behaviour for `static` and `shared` lib builds is now as follows:

```
                 options.build_shared_lib
                 None    True   False
 build_targets     
        static    yes     yes    fail
       !static     no    fail     yes
          None     no     yes      no
```

I.e. if the legacy switches are used, they behave as expected (switching library builds on and off). If only the new `--build-targets=` switch is used, it builds exactly what is listed by it. When using a combination of both legacy and new switches, their decisions are checked for consistency. For example this would fail with a `UserError`:

```
./configure.py --disable-shared-library --build-targets="static,tests"
```

configure.py
```diff
@@ -2982,6 +2998,37 @@ def canonicalize_options(options, info_os, info_arch):
...
         logging.info('Canonicalized CPU target %s to %s', options.cpu, options.arch)
 
+    # select and sanity check build targets
+    def canonicalize_build_targets(options):
+        build_targets = read_build_targets(options)
+
+        # --build-targets was not provided: build default targets
+        if build_targets == None:
+            options.build_targets = "cli,tests"
+            return
```
reneme: This ensures legacy behaviour by selecting the targets to `cli,tests`. Furthermore, it must skip the sanity checks below, to fully leave control over static and shared libraries to the legacy switches `--(enable/disable)-(shared/static)-library)`.

reneme:
I removed the `fuzzers` `--build-targets` for now, to keep the CI scripting complexity down for this pull request. I'd appreciate a quick review to tell me whether this is on the right track. I'll aim to add the rest in a follow-up PR.

reneme:
... mmh, [the "coverage" Travis Job](https://travis-ci.org/randombit/botan/jobs/582811592) failed with some package installation it seems. Otherwise things are green.

configure.py
```diff
@@ -276,6 +276,15 @@ def src_info(self, typ):
         else:
             raise InternalError("Unknown src info type '%s'" % (typ))
 
+ACCEPTABLE_BUILD_TARGETS = ["static", "shared", "cli", "tests", "bogo_shim"]
+
+def read_build_targets(options):
```
randombit: The code here should happen once, in `canonicalize_build_targets`, and then stores the result in `options.build_targets`.

configure.py
```diff
@@ -2982,6 +2994,37 @@ def canonicalize_options(options, info_os, info_arch):
...
+            return
+
+        # --build-targets was set to empty string: invalid input
+        if not build_targets:
+            raise UserError("select at least one --build-targets")
+
+        for build_target in build_targets:
+            if build_target not in ACCEPTABLE_BUILD_TARGETS:
+                raise UserError("unknown build target: %s" % build_target)
+
```
randombit: I think it is possible here for a user on MSVC to request static+shared which doesn't work due to how Windows DLLs work. It should probably error instead.

configure.py
```diff
@@ -487,8 +496,8 @@ def process_command_line(args): # pylint: disable=too-many-locals,too-many-state
     build_group.add_option('--with-debug-asserts', action='store_true', default=False,
                            help=optparse.SUPPRESS_HELP)
 
-    build_group.add_option('--build-bogo-shim', action='store_true', default=False,
-                           help=optparse.SUPPRESS_HELP)
+    build_group.add_option('--build-targets', default=None, dest="build_targets",
+                           help="build specific targets and tools (%s)" % ', '.join(ACCEPTABLE_BUILD_TARGETS))
```
randombit: If you use an `append` target then it becomes possible to provide multiple --build-target options on the command line. That is sometimes useful when the configure.py invocation is being created using a script (as is done, in various ways, in most distribution packages). Ex some distribution script might have logic like

```
if build_cli:
   cmd += ['--build-target=cli']
if build_shared:
   cmd += ['--build-target=shared']
...
```

Depending on the language used it may be easier to provide multiple options vs performing a join to form a single --build-targets=cli,shared.

configure.py
randombit: Thanks for addressing this. It would be nice to be able to deprecate and remove the older options eventually as --build-targets is much cleaner - but doing so will break everyone's build scripts, so we probably have to support both methods for at least a few releases.

Roughly: 2.12: add --build-targets, 2.13: start printing a deprecation warning when older options are used, 2.xx?/3.0?: remove the options.

configure.py
reneme: Thats true. If a user calls `./configure --build-targets="static,shared` on windows, she will be presented with an error due to a sanity check that is already part of the script (see [here](https://github.com/randombit/botan/blob/0ee18280a798c3f311259036ae70e503ae9ff3a3/configure.py#L3057).

In fact, I ran into this issue [in the CI](https://ci.appveyor.com/project/randombit/botan/builds/27287793/job/g8g2l2yki2wc8cc3) once as well, before adding https://github.com/randombit/botan/pull/2098/commits/577ac1a01b8b3ead65655de5ef84176253549428.

configure.py
reneme: Both is possible now. Even something like `--build-targets=cli --build-targets="static,shared"`, is flattened accordingly.

configure.py
reneme: True. However, as stated in a different reply, this will be caught just a few lines below. Admittedly, this is somewhat hard to spot, as `canonicalize_build_targets` sets `options.build_static_lib` and `options.build_shared_lib` accordingly.

randombit: Ah sorry I missed this, what's there is fine then.

PR #2105, opened 2019-09-13
tls cli: getting real number of cores as possible.

devnexen:


src/cli/tls_http_server.cpp
```diff
@@ -513,7 +513,7 @@ class TLS_HTTP_Server final : public Command
          {
          if(size_t t = get_arg_sz("threads"))
             return t;
-         if(size_t t = std::thread::hardware_concurrency())
+         if(size_t t = get_cpu_total())
```
randombit: Should be `Botan::OS::` prefixed. Also, seems better to use `get_cpu_available`? Since if a CPU is offline it doesn't make sense to schedule an extra thread for it.

devnexen: yes it s clearer prefixed even though it compiled. Your last point is valid too.

randombit:
CI is failing looks like include is missing

devnexen:
ah yes on windows I see.

PR #2106, opened 2019-09-18
Add X.509 Certificate Usage_Type::ENCRYPTION

reneme:
This introduces a new entry to the `Usage_Type` enum in x509cert.h: `ENCRYPTION` that translates to "X509v3 Key Usage" containing either "Key Encipherment" or "Data Encipherment".

In our application we would like to check the key usage type as part of `x509_path_validate` rather than adding an extra manual check. However, the implementation of `X509_Certificate::allowed_usage()` states that it is based on [RFC 5280 4.2.1.12](https://tools.ietf.org/html/rfc5280#section-4.2.1.12) which does not explicitly list a fitting category.

What do you think?

### DIFF
--- a/src/tests/unit_x509.cpp
+++ b/src/tests/unit_x509.cpp
@@ -990,6 +990,8 @@ Test::Result test_usage(const Botan::Private_Key& ca_key,
    result.test_eq("key usage cRLSign not allowed",
                   user1_cert.allowed_usage(
                      Key_Constraints(Key_Constraints::DIGITAL_SIGNATURE | Key_Constraints::CRL_SIGN)), false);
+   result.test_eq("encryption is not allowed",
+                  user1_cert.allowed_usage(Usage_Type::ENCRYPTION), false);
 
    // cert only allows digitalSignature, so checking for only that should be ok
    result.confirm("key usage digitalSignature allowed", user1_cert.allowed_usage(Key_Constraints::DIGITAL_SIGNATURE));
@@ -1014,6 +1016,9 @@ Test::Result test_usage(const Botan::Private_Key& ca_key,
    result.confirm("key usage multiple cRLSign allowed", mult_usage_cert.allowed_usage(Key_Constraints::CRL_SIGN));
    result.confirm("key usage multiple digitalSignature and cRLSign allowed", mult_usage_cert.allowed_usage(
                      Key_Constraints(Key_Constraints::DIGITAL_SIGNATURE | Key_Constraints::CRL_SIGN)));
+   result.test_eq("encryption is not allowed",
+                  mult_usage_cert.allowed_usage(Usage_Type::ENCRYPTION), false);
+
 
    opts.constraints = Key_Constraints::NO_CONSTRAINTS;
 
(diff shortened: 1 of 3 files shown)


randombit:
Seems ok to me

src/tests/unit_x509.cpp
```diff
@@ -1027,6 +1028,27 @@ Test::Result test_usage(const Botan::Private_Key& ca_key,
    result.confirm("key usage digitalSignature allowed", no_usage_cert.allowed_usage(Key_Constraints::DIGITAL_SIGNATURE));
    result.confirm("key usage cRLSign allowed", no_usage_cert.allowed_usage(Key_Constraints::CRL_SIGN));
 
+   if (sig_algo == "RSA")
```
randombit: I'd like to see some additional tests here, using the original constraints and for all key types:

- If key encipherment or data encipherment are set, test that ENCRYPTION is allowed
- Otherwise, test that ENCRYPTION is denied

As it is (with the current tests) allowed_usage(ENCRYPTION) could simply be `return true` regardless of usage or algorithm and everything would pass.

[The test you have already written is also fine to include.]

src/tests/unit_x509.cpp
```diff
@@ -1026,6 +1032,28 @@ Test::Result test_usage(const Botan::Private_Key& ca_key,
    // cert allows every usage
    result.confirm("key usage digitalSignature allowed", no_usage_cert.allowed_usage(Key_Constraints::DIGITAL_SIGNATURE));
    result.confirm("key usage cRLSign allowed", no_usage_cert.allowed_usage(Key_Constraints::CRL_SIGN));
+   result.confirm("key usage encryption allowed", no_usage_cert.allowed_usage(Usage_Type::ENCRYPTION));
```
reneme: I find this "special case" questionable. Except for RSA these certificates theoretically "allow" everything (including encryption) but practically don't support encryption due to their asymmetric algorithm. Should `allowed_usage` still "claim" encryption?

The creation of a certificate with such an algorithm will throw when specifying either `KEY_ENCIPHERMENT` or `DATA_ENCIPHERMENT` specifically.

src/tests/unit_x509.cpp
reneme: I added 'negative' tests for the already existing constraints. None of those previously neither allowed  `KEY_ENCIPHERMENT` nor `DATA_ENCIPHERMENT`. The "cert allows every usage" [above](https://github.com/randombit/botan/pull/2106/files#r327026610) is a bit questionable though.

src/tests/unit_x509.cpp
randombit: This is because allowed_usage only looks at the KeyUsage extension which if missing (or wrong) we have no guidance and accept anything, eg a DH cert will be marked as ok for digital signature.

To fix this we must deserialize the public key when parsing and compute the possible constraints. This adds some overhead to cert parsing, but also has some advantages - we can add a new function to `X509_Certificate` that returns a const& to the `Public_Key` so other functions which need to access the public key eg during validation can avoid the extra deserialization step as done in `load_subject_public_key`

reneme: I see. Thanks for the detailed rationale.
Hence, I understand that the behaviour as sketched out in this test is somewhat expected and "desired", for the time being, right?

randombit: Yes, expected at least. Probably not really desirable but at least the test does confirm the current behavior. I'd be happy with a PR changing this so we verify or mask the certificate constraints against what the key can actually do, but doesn't need to happen here.

PR #2107, opened 2019-09-21
build-data: fix up a bug in botan.pc.in

liuqun:
This bug was found when I ran pkg-config to see the botan lib path:
`pkg-config --libs botan-2`
```
-L/usr///usr/lib/x86_64-linux-gnu -lbotan-2 -fstack-protector -m64 -pthread
```


liuqun:
Updated in Issue-2108 and PR-2109 ...

PR #2109, opened 2019-09-21
build-data: fix up a bug in botan.pc.in

### LINKED ISSUE #2108: configure.py libdir bug was found on libbotan-2-dev from software center of Ubuntu 19.04
Most other projects with an autoconf-based configure script support an option of "--libdir=/path/to/x86_64-linux-gnu", for example:
```
./configure --prefix="/usr" --libdir=/usr/lib/x86_64-linux-gnu
```

However, the python script `configure.py` in botan does support this full-path option style.

Ubuntu 19.04 and Debian's maintainer build botan-2.9.0 in their software build systems. And their build script for botan-2.9.0 uses full-path options --libdir.

On my Ubuntu 19.04, install the pre-built libbotan-2-9 from the deb package, then trying to run `pkg-config` to get the lib path:
 * `sudo apt-get install -y libbotan-2-9 libbotan-2-dev`
 * `pkg-config --libs botan-2`

You will got a wrong -L option:

    -L/usr///usr/lib/x86_64-linux-gnu -lbotan-2 -fstack-protector -m64 -pthread

Since Ubuntu 19.04 always search lib\*.so from "/usr/lib/x86_64-linux-gnu", the buggy -L option could be always ignored by the C++ compilers.

---

A patch for the support on full-path styled "--libdir=/usr/lib/x86_64-linux-gnu" is at PR-2109 now

liuqun:
`configure.py` should support full path options such as  `--libdir=/full/path/to/lib/x86_64-linux-gnu`

Fixes #2108


liuqun:
travis-ci is passing now...

randombit:
Thanks. I didn't know `--libdir` was supposed to be abolute path instead of relative to prefix. That behavior doesn't really make any sense to me. But we do want configure.py to behave as close as possible to autoconf, especially in cases where an option name is shared.

However we still want to continue accepting relative libdir/bindir so as to not break any build scripts which are using configure.py's interpretation of libdir as being relative to the prefix. I've pushed a change to your branch that allows both versions to work; if a non-absolute path is provided to --libdir/--bindir, it continues to act relative to prefix. WDYT?

randombit:
@liuqun ping - do my modifications seem ok to you?

PR #2110, opened 2019-09-23
add status code to OCSP responses

alnkpa:
We suggest removing the exception on parsing a OCSP response with `resp_status != 0`.

Instead, we'd add a `status` field to `OCSP::Response` that needs to be checked before querying the `Response` for revocation information.


src/lib/x509/ocsp.h
```diff
@@ -63,6 +63,16 @@ class BOTAN_PUBLIC_API(2,0) Request final
...
    };
 
+/**
+* OCSP response status.
+*
+* see https://tools.ietf.org/html/rfc6960#section-4.2.1
+*/
+enum class BOTAN_PUBLIC_API(2,12) Response_Status_Code
+{
+   SUCCESSFUL = 0, MALFORMED_REQUEST = 1, INTERNAL_ERROR = 2, TRY_LATER = 3, SIG_REQUIRED = 5, UNAUTHORIZED = 6
```
randombit: Put each enum member on its own line.

For `enum class` it seems cleaner to use a more natural case since unlike old style `enums` they are not all dropped into the global namespace. So eg `Successful`, `Malformed_Request`, ...

Also, style nit with enum we put first `{` on same line as `enum`

randombit:
CI is failing, looks related to this change.

randombit:
Also please add some decoding tests of error-status OCSP responses.

randombit:
Looks good thanks

PR #2118, opened 2019-09-26
unoptimise inputs for the timing safe comparison purpose.

devnexen:


src/lib/utils/mem_ops.cpp
```diff
@@ -58,9 +58,13 @@ uint8_t ct_compare_u8(const uint8_t x[],
                       size_t len)
    {
    volatile uint8_t difference = 0;
```
randombit: We should support using `timingsafe_bcmp` here (and any other similar APIs that may exist, though I don't know of any)

randombit: Actually `timingsafe_bcmp` turns out to be ... well not unsafe exactly but certainly very fragile wrt future compiler optimizations, so we probably don't want to do this.

devnexen:
So you are right for timingsafe_bcmp, the volatile usage here was to make sure data inputs are read and not optimised away by the compiler in every case. However the delta being volatile itself might be good enough after second thoughts.

randombit:
I think we should leave this as is until we have a firm idea if and how additional use of `volatile` helps here. Thanks for looking at this.

PR #2121, opened 2019-09-28
Avoid preview MSVC and re-enable Serpent/AVX2

randombit:
Discussion in GH #2120


randombit:
Also fails with cl 19.23.28105.4 for x64

PR #2126, opened 2019-09-29
Improve SM4 speed

xiaojuna8:
SM4 by rotr in compile time，make 3 arrays to store roted values, save time to rot in runtime.  
on my computer
```
./botan speed SM4
# before
SM4 encrypt buffer size 1024 bytes: 169.450 MiB/sec 16.75 cycles/byte (84.73 MiB in 500.00 ms)
SM4 decrypt buffer size 1024 bytes: 169.344 MiB/sec 16.77 cycles/byte (84.67 MiB in 500.00 ms)
---
# after
SM4 encrypt buffer size 1024 bytes: 184.208 MiB/sec 15.40 cycles/byte (92.10 MiB in 500.00 ms)
SM4 decrypt buffer size 1024 bytes: 184.494 MiB/sec 15.38 cycles/byte (92.25 MiB in 500.00 ms)
```


randombit:
This implementation was deliberate as a single 1 Kb table reduces the cache signature vs 4 Kb of tables, especially as the performance is only slightly impacted. So while I thank you for the contribution I do not think we want to accept this change; performance is a goal but side channels should be reduced or eliminated whenever possible and that takes priority.

If you are interested in SM4 performance I would instead suggest trying to adapt the technique in https://github.com/mjosaarinen/sm4ni which uses AES-NI instructions to accelerate SM4. Since SM4 sbox is based on GF(2^n) inversion it also seems possible that it could be executed using SSSE3 vector permutes as in https://www.shiftleft.org/papers/vector_aes/vector_aes.pdf and https://github.com/randombit/botan/blob/master/src/lib/block/aes/aes_vperm/aes_vperm.cpp - on Skylake the SSSE3 version of AES is about 1.5x the speed of the table based version, and also executes in constant time. 

Especially as SM4 is mandated in China and widely used there we would be happy to accept such optimizations, even if in limited form. By that I mean, it does not need to necessarily improve all scenarios. For example the SM4 using AES-NI code I linked to processes 4 blocks in parallel in encryption mode only, which accelerates CTR and GCM (in either direction), or OCB and XTS in encrypt direction only, but does nothing for CBC mode. But still would be of interest since it improves performance and side channel safety for at least some important modes, even if it does not cover all cases.

PR #2127, opened 2019-10-01
fix deprecation warnings on MSVC

alnkpa:


randombit:
Thanks. What was the issue with the current version? Just didn't work correctly, compilation failure, ?

alnkpa:
the compiler warned about it

Issue #2128, opened 2019-10-01
klocwork warning - get_affine_x() and get_affine_y()


MartinOsterta:
When we checked the code in Klocwork we got a warning in the following code part:	
BigInt PointGFp::get_affine_y() const
   {
   if(is_zero())
      throw Illegal_Transformation("Cannot convert zero point to affine");
   secure_vector<word> monty_ws;
   if(is_affine())
==>      return m_curve.from_rep(m_coord_y, monty_ws);

The warning is: "Non-void function returns void value" which is correct since from_rep has two different implmentations: One with return type void and a second returning type BigInt. The compiler chooses on the first parameter which method to use (passing first  parameter as const vs. non-const)

First It is strange to me that the same overloaded methods have different return parameters. Is this an error or intention?

I assume the code above can be fixed in the following way as the for affine points the follwing is valid: (X', Y') == (X', Y', Z=1) and is_affine() checks whether Z == 1.
   if(is_affine())
      return m_coord_y;

The same issue applies to BigInt PointGFp::get_affine_x() const.

Do you agree with that solution?

randombit:
No, solution is incorrect as the coordinates are usually held in Montgomery form, thus a conversion must be applied to get the actual coordinate value.

The different return types for from_rep is confusing but intentional; one is in-place operation which mutates its argument and returns void, the other creates a temporary and returns it. So I think the actual warning from Klocwork is a false positive. That said it is confusing, and we have a policy about avoiding static analysis warnings whenever possible even if false positives. So we should rename the one that creates a temporary to `from_rep_to_tmp`, which matches the pattern of other functions on CurveGFp which create temporary values to return rather than doing in-place operations. Then overload resolution is not an issue.

randombit:
Thanks for the patch

PR #2129, opened 2019-10-02
fix for botan issue #2128: klocwork warning - get_affine_x() and get_affine_y() 

MartinOsterta:
This is a fix for issue #2128. Klocwork gives warning about "Non-void function returns void value" since there are overloaded methods in class CurveGFp which have different return types (one is void).
As a fix one of the overloaded methods is renamed to resolve the issue and make Klocwork happy.


randombit:
Nice thank you

PR #2136, opened 2019-10-08
Attempt to address some LGTM warnings in Roughtime

randombit:


nunojpg:
I believed this warnings would only be triggered for the pull request were line was introduced, otherwise I would have fixed it, as upstream they also seem interested. Sorry.

PR #2138, opened 2019-10-09
`./configure.py --without-documentation` disables build target 'docs'

reneme:
Trying to [update Botan's conan package](https://github.com/bincrafters/conan-botan/pull/13) we ran into some python problem in `build_docs.py`. While I'm pretty sure that the root cause is not in Botan itself, I'd have loved to just disable the script invocation via `./configure.py --without-documentation`.

Currently, the `build_docs.py` script simply exits if `build_config.json` is set up to not build documentation. With this patch, a simple call to `make` won't even start the script anymore.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1962,7 +1962,8 @@ def create_template_vars(source_paths, build_paths, options, modules, cc, arch, 
             yield 'fuzzers'
         if 'bogo_shim' in options.build_targets:
             yield 'bogo_shim'
-        yield 'docs'
+        if options.with_documentation:
+            yield 'docs'
 
     def absolute_install_dir(p):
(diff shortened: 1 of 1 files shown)


reneme:
Turns out that the issue was due to [a change in Botan](https://github.com/randombit/botan/pull/2131/commits/f4d69b8062d5223008d6b610558cba6074c3ca3d) after all: see https://github.com/randombit/botan/issues/2141.

randombit:
@reneme This change is fine. Can you rebase? I think these issues (especially the nmake one) require a 2.12.1 release. So I have removed from master the recent merges. In the future I will avoid merging anything major until at least a week after the release so there is time to find and address any issues that require a patch release.

PR #2140, opened 2019-10-09
FIX: `make install` forces build of unwanted targets

reneme:
Trying to [update Botan's conan package](https://github.com/bincrafters/conan-botan/pull/13) we noticed that `make install` forces the build of the CLI utility even though it was explicitly disabled via the new `--build-targets=` config flag.

This explicitly lists all targets that are (a) enabled via `--build-targets` and (b) required for installation. Furthermore `install.py` won't even try to install the CLI binary if it was disabled.


randombit:
@reneme Change looks good. Needs rebase then will merge.

PR #2154, opened 2019-10-15
Fix some dangling pointer problems

danielmartin:
After running Clang's experimental -Wlifetime over the Botan codebase,
I discovered a few dangling pointer problems that resulted from
returning a pointer to a temporary object. The temporary object will
be destroyed at the end of the scope/expression and the pointer will dangle.

Here's a Godbolt that shows the problem: https://godbolt.org/z/fJyWl4

Instead of returning a dangling pointer, return a `std::unique_ptr` to
make more explicit that it's the responsibility of the caller to
destroy the resource that was handled in by the function. In some
cases, like `subject_public_key()`, it was explicitly said in
comments ("This object is owned by the caller."), but returning a
explicit type that transfers ownership (a unique pointer) is better.

I've executed every Botan test with ASAN on and it looks good.

### DIFF
--- a/src/lib/modes/aead/aead.h
+++ b/src/lib/modes/aead/aead.h
@@ -111,9 +111,9 @@ class BOTAN_PUBLIC_API(2,0) AEAD_Mode : public Cipher_Mode
 * @param name AEAD name
 * @param direction ENCRYPTION or DECRYPTION
 */
-inline AEAD_Mode* get_aead(const std::string& name, Cipher_Dir direction)
+inline std::unique_ptr<AEAD_Mode> get_aead(const std::string& name, Cipher_Dir direction)
    {
-   return AEAD_Mode::create(name, direction, "").release();
+   return AEAD_Mode::create(name, direction, "");
    }
 
 }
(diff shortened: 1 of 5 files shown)


randombit:
Hi. Thanks for testing with this new option, it looks very useful. That said I really don't understand the issue here. We are returning the result of `unique_ptr::release`, which causes the unique_ptr to no longer retain ownership. So ~unique_ptr would be a no-op, and we effectively return a raw pointer allocated by `new`. This seems to me like a bug in the Clang warning, in that it is failing to account for semantics of `release`?

Now, these APIs are certainly bad news; we should not be returning raw pointers with the expectation that the application will free them. But we cannot change them because that would break compat with existing code; there is a reason most of these APIs are just calling some newer API which does return a proper `unique_ptr`. In fact though all of them should be officially deprecated so we can remove them later on.

danielmartin:
Yes, I agree with the incorrect lifetime diagnostic here. Ideally, the API should return a smart pointer that represents ownership better, but I understand this is a big change in public API that will affect tons of library clients.

I'm handling the Clang Lifetime bug separately (https://github.com/mgehre/llvm-project/issues/70) so I think I'll abandon this pull request.

PR #2155, opened 2019-10-15
prefer std::shared_ptr to boost::shared_ptr

nunojpg:


randombit:
Thanks. I had some memory this didn't work for some reason, but CI is happy so :tada:

PR #2168, opened 2019-10-20
fix roughtime coverity warning

nunojpg:
This fixes the coverity false positive. I did refactor to improve the design, but the warning itself could only be silenced by passing sizeof with 1 unit off.


randombit:
Thanks

PR #2169, opened 2019-10-20
silence trivial warnings

nunojpg:


src/lib/pubkey/rsa/rsa.cpp
```diff
@@ -413,7 +413,7 @@ class RSA_Private_Operation
          */
          m.sig_words();
 
-         auto future_j1 = std::async(std::launch::async, [this, &m, &d1_mask, powm_window]() {
```
randombit: This will hit the problem in MSVC where it doesn't realize this capture is not required.

Maybe it is possible to rewrite the code in such a way that Clang doesn't warn but MSVC still builds

src/tests/test_modes.cpp
```diff
@@ -148,7 +148,7 @@ class Cipher_Mode_Tests final : public Text_Based_Test
          const size_t large_nonce_size = 65000;
          result.test_eq("Large nonce not allowed", mode.valid_nonce_length(large_nonce_size), false);
          result.test_throws("Large nonce causes exception",
-                            [&mode,large_nonce_size]() { mode.start(nullptr, large_nonce_size); });
```
randombit: Similarly this is for the capture bug in MSVC

src/lib/pubkey/rsa/rsa.cpp
nunojpg: the easiest fix is to remove const from the variable. Then it becomes a required capture. What do you think?

nunojpg: Maybe constexpr would also work for MSVC. I will push and check if it compiles.

randombit: Re making it non-const kind of ugly. I really want anything that is not modified `const`, since then I don't need to think about if it is modified and if so how... but warning isn't great either.

Based on https://godbolt.org/z/ofmS89 it looks like MSVC accepts if the const is in the global namespace instead of within the function. So I think I'd prefer that. Just rename to `RSA_POWM_WINDOW_BITS` and hoist it out.

randombit: Looks like constexpr won't work https://godbolt.org/z/q5Irxr

nunojpg: ...static constexpr works. I think keeping it local like that is preferable.

nunojpg: static const/static constexpr

randombit: static is good

PR #2172, opened 2019-10-22
Add support for XMSS X.509 certificates

securitykernel:
Add support for generating and parsing X.509 certificates with XMSS keys/signatures. As discussed via email, we follow [RFC draft-vangeest-x509-hash-sigs-03](https://www.ietf.org/archive/id/draft-vangeest-x509-hash-sigs-03.txt), even though it has expired. This is coordinated with ISARA Corp., which is one of the authors of the RFC draft, and BouncyCastle (via @dghgit), so BouncyCastle will implement the same format in order to achieve interoperability between Botan and BouncyCastle.

Therefore, this PR makes the following changes:

* adds OID for XMSS from draft-vangeest-x509-hash-sigs-03 and sets as the default (the former are still retained as XMSS-draft12)
* encode XMSS public keys as ASN.1 OCTET STRING
* adds support for generating X.509 certificates in `X509_Object`
* adds tests for parsing two self-signed XMSS certificates, one generated by ISARA Corp., one by BouncyCastle

Note that both XMSS public and private key were not encoded as OCTET STRINGs before. In order not to have to update the test vectors and to support previously encoded public and private keys, decoding now falls back to the old format if the key bits could not be encoded as and OCTET STRING. Is this desired?


randombit:
> In order not to have to update the test vectors and to support previously encoded public and private keys, decoding now falls back to the old format if the key bits could not be encoded as and OCTET STRING. Is this desired?

Yes we should retain ability to parse previously generated keys if possible.  OTOH this is less pressing since we have already broken previous use of XMSS in 2.11 when we switched from draft-6 to final RFC format. So if it proves to be a problem it is not essential that we retain this.

src/lib/pubkey/xmss/xmss_privatekey.cpp
```diff
@@ -17,15 +17,34 @@
...
 #endif
 
 namespace Botan {
 
-XMSS_PrivateKey::XMSS_PrivateKey(const secure_vector<uint8_t>& raw_key)
-   : XMSS_PublicKey(unlock(raw_key)),
+namespace {
+
+secure_vector<uint8_t> extract_raw_key(const secure_vector<uint8_t>& key_bits)
+{
```
randombit: There should be a comment here explaining what is going on

src/tests/test_x509_path.cpp
```diff
@@ -895,6 +895,42 @@ class Path_Validation_With_OCSP_Tests final : public Test
...
+
+      std::vector<Test::Result> run() override
+      {
+      if(Botan::has_filesystem_impl() == false)
+         {
+         return {Test::Result::Note("XMSS path validation",
+                                    "Skipping due to missing filesystem access")};
+         }
+
+      return  {validate_self_signed("XMSS path validation with certificate created by ISARA corp", "xmss_isara_root.pem"),
```
randombit: I think the example certs are missing from this commit

randombit:
@securitykernel Looks good. But the CLI tests must be updated - they are testing XMSS stateful update and the new key encoding has broken the tests.

``./src/scripts/test_cli.py <path to botan cli> xmss``

securitykernel:
> @securitykernel Looks good. But the CLI tests must be updated - they are testing XMSS stateful update and the new key encoding has broken the tests.
> 
> `./src/scripts/test_cli.py <path to botan cli> xmss`

Updated in 03eb48c7c.

src/tests/test_x509_path.cpp
```diff
@@ -895,6 +895,42 @@ class Path_Validation_With_OCSP_Tests final : public Test
 
 BOTAN_REGISTER_TEST("x509_path_with_ocsp", Path_Validation_With_OCSP_Tests);
 
+class XMSS_Path_Validation_Tests final : public Test
```
randombit: Alas one more problem - this fails if XMSS is not available, as when building using the NIST build policy.

securitykernel: Added the missing guards in a7b7fa47c.

PR #2173, opened 2019-10-22
Label test vectors and add remaining test vectors from IEEE 1619-2007

### LINKED ISSUE #2146: Investigate XTS test vectors
The comment in the data file has them all coming from P1619 and most of them seem to be, but some of the vectors do not appear in http://libeccio.di.unisa.it/Crypto14/Lab/p1619.pdf and others from the spec are missing in the vector file.

ssd71:
Fixes #2146

I labeled the test vectors as best as I could find, I did find some duplicates but I wasn't sure if should delete them, I also added the rest of the vectors from the spec. Please check it out

Edit: added issue number


randombit:
@ssd71 Thanks. I think some of the test vectors are using AES-256 and need to go in `[AES-256/XTS]` section.

randombit:
Looks great thank you!

PR #2175, opened 2019-10-23
Add check for permitted range of reseed_interval

securitykernel:
Adds a check and corresponding test in HMAC_DRBG whether reseed_interval is in the permitted range (Table 2 in NIST SP 800-90A, Rev. 1).

Stumbled upon this while reading https://eprint.iacr.org/2018/349.pdf.

### DIFF
--- a/src/lib/rng/hmac_drbg/hmac_drbg.cpp
+++ b/src/lib/rng/hmac_drbg/hmac_drbg.cpp
@@ -10,6 +10,26 @@
 
 namespace Botan {
 
+namespace {
+
+void check_limits(size_t reseed_interval,
+                  size_t max_number_of_bytes_per_request)
+   {
+   // SP800-90A permits up to 2^48, but it is not usable on 32 bit
+   // platforms, so we only allow up to 2^24, which is still reasonably high
+   if(reseed_interval == 0 || reseed_interval > static_cast<size_t>(1) << 24)
+      {
+      throw Invalid_Argument("Invalid value for reseed_interval");
+      }
+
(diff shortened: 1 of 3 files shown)


src/lib/rng/hmac_drbg/hmac_drbg.cpp
```diff
@@ -20,6 +20,11 @@ HMAC_DRBG::HMAC_DRBG(std::unique_ptr<MessageAuthenticationCode> prf,
    {
    BOTAN_ASSERT_NONNULL(m_mac);
 
+   if(reseed_interval == 0 || reseed_interval > static_cast<size_t>(1) << 48)
```
randombit: This overflows and causes problems on 32-bit builds.

We could either omit the second check when sizeof(size_t) == 4 (since it is not possible to actually express a reseed interval exceeding the requirement in that case), or else reduce the limit below the SP mandated maximum. Which seems preferable to me since 2**48 is a *lot* of output. Maybe drop our max to 2**24 or so??

You forgot to update the constructor taking just an `EntropySource&`. I think it would be best to hoist these checks up to an anon namespace function to validate both reseed_interval and max bytes / request parameters so the logic is not duplicated 3 times.

securitykernel: I totally forgot about 32 bit platforms. 2^24 seems OK to me, I also documented that in the API doc. Introduced a new function `check_limits()` to deduplicate code as you suggested, all in 91f925164.

randombit:
LGTM

PR #2188, opened 2019-11-09
Allow disabling TLS 1.0/1.1 and DTLS 1.0 at build time

randombit:
FYI @neusdan @securitykernel this disables TLS 1.0/1.1, TLS CBC ciphersuites, and MD5+SHA at build time in the BSI policy. I know 1.0/1.1 are prohibited and MD5+SHA are only enabled for benefit of TLS. Wasn't sure about TLS CBC wrt BSI perspective, though using CBC ciphersuies when GCM/ChaCha is available seems like a bad idea IMO.

### DIFF
--- a/doc/dev_ref/todo.rst
+++ b/doc/dev_ref/todo.rst
@@ -89,7 +89,6 @@ TLS
 * Make DTLS support optional at build time
 * Improve/optimize DTLS defragmentation and retransmission
 * Implement logging callbacks for TLS
-* Make TLS v1.0 and v1.1 optional at build time
 * Make RSA optional at build time
 * Make finite field DH optional at build time
 * Authentication using TOFU (sqlite3 storage)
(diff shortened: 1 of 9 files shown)


securitykernel:
@randombit Thanks. TLS CBC is still permitted where GCM is not available, maybe some special scenario. #2195 is an update of the TLS policy which we wanted to do these days anyway. It brings the implementation in line with the latest version 2019-01 of the underlying BSI technical guideline. In this context it adds CCM ciphersuites to the BSI TLS policy, amongst other changes.

randombit:
@securitykernel Should I remove `tls_cbc` from the banned list in the BSI build policy?

randombit:
Oh ftr turns out we cannot disable SHA-1 even with TLS v1.0 removed as there is a hard dependency on it in X.509, so only MD5 is actually turned off.

securitykernel:
@randombit Yes, TLS CBC should still be allowed in the BSI policy. Thank you.

PR #2193, opened 2019-11-10
CLI fingerprint read from stdin

nunojpg:
This allows at least to verify a DER public key from stdin, while not to convert to PEM.


randombit:
Nice thanks

PR #2207, opened 2019-11-30
attemp_srp is overriding thus marking explicitally

devnexen:


randombit:
Thanks. I noticed this via #2206 but it will be a bit before I can address the remaining issues there.

Issue #1906, opened 2019-04-19
Implement missing Python functions


randombit:
Following functions are defined in the C interface but not used in the Python binding:

 - [ ] botan_constant_time_compare
 - [ ] botan_hex_encode
 - [ ] botan_hex_decode
 - [ ] botan_base64_encode
 - [ ] botan_base64_decode
 - [x] botan_hash_copy_state
 - [x] botan_hash_block_size
 - [ ] botan_cipher_output_length
 - [ ] botan_pwdhash
 - [ ] botan_pwdhash_timed
 - [x] botan_block_cipher_init
 - [x] botan_block_cipher_destroy
 - [x] botan_block_cipher_clear
 - [x] botan_block_cipher_set_key
 - [x] botan_block_cipher_block_size
 - [x] botan_block_cipher_encrypt_blocks
 - [x] botan_block_cipher_decrypt_blocks
 - [x] botan_block_cipher_name
 - [x] botan_block_cipher_get_keyspec
 - [ ] botan_mp_set_from_radix_str
 - [x] botan_mp_to_bin
 - [ ] botan_mp_from_bin
 - [x] botan_mp_is_positive
 - [x] botan_mp_is_zero
 - [ ] botan_mp_mod_mul
 - [x] botan_mp_rand_bits
 - [x] botan_mp_rand_range
 - [ ] botan_mp_gcd
 - [ ] botan_privkey_ed25519_get_privkey
 - [ ] botan_privkey_export_encrypted
 - [ ] botan_privkey_export_encrypted_pbkdf_iter
 - [x] botan_privkey_load_dh
 - [x] botan_privkey_load_dsa
 - [x] botan_privkey_load_ecdh
 - [x] botan_privkey_load_ecdsa
 - [ ] botan_privkey_load_ed25519
 - [x] botan_privkey_load_elgamal
 - [x] botan_privkey_load_rsa
 - [ ] botan_privkey_load_rsa_pkcs1
 - [x] botan_privkey_load_sm2
 - [ ] botan_privkey_load_x25519
 - [ ] botan_privkey_rsa_get_privkey
 - [ ] botan_privkey_x25519_get_privkey
 - [ ] botan_pubkey_ed25519_get_pubkey
 - [x] botan_pubkey_load_dh
 - [x] botan_pubkey_load_dsa
 - [x] botan_pubkey_load_ecdh
 - [x] botan_pubkey_load_ecdsa
 - [ ] botan_pubkey_load_ed25519
[...]

PR #2208, opened 2019-12-02
Implementing a few functions missing in the Python binding

zacikpa:
I tried to add another bit of functionality mentioned in #1906.


randombit:
Thanks change looks good. There are two pylint warnings that need to be addressed https://travis-ci.org/randombit/botan/jobs/619691102

The trailing whitespace should be removed. The too-many-public-methods warnings should just be disabled within the test file

PR #2213, opened 2019-12-03
Add basic support for CRL in FFI API 

zacikpa:
Hi, I tried to add some basic support of CRLs in the FFI API.
I plan to add more CRL related functions and move the support to the Python binding as well.

I've got a few questions:

- Is it okay if it stays in the ffi_cert.cpp file?
- Should my `botan_x509_cert_verify_with_crl` stay separate from `botan_x509_cert_verify` to be backwards compatible?
- I couldn't figure out what is the meaning of the `magic` number in `BOTAN_FFI_DECLARE_STRUCT`. Is it just some random number that each struct has assigned to prevent some errors? What should I set it to?

Thanks beforehand for all suggestions.

### DIFF
--- a/doc/api_ref/ffi.rst
+++ b/doc/api_ref/ffi.rst
@@ -1152,7 +1152,51 @@ X.509 Certificates
     Set ``reference_time`` to be the time which the certificate chain is
     validated against. Use zero to use the current system clock.
 
+.. cpp:function:: int botan_x509_cert_verify_with_crl(int* validation_result, \
+                  botan_x509_cert_t cert, \
+                  const botan_x509_cert_t* intermediates, \
+                  size_t intermediates_len, \
+                  const botan_x509_cert_t* trusted, \
+                  size_t trusted_len, \
+                  const botan_x509_crl_t* crls, \
+                  size_t crls_len, \
+                  const char* trusted_path, \
+                  size_t required_strength, \
+                  const char* hostname, \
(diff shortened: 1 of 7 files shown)


randombit:
Great to see this happening. I will review fully later.

> Is it okay if it stays in the ffi_cert.cpp file?

Yes that's ok.

> Should my botan_x509_cert_verify_with_crl stay separate from botan_x509_cert_verify to be backwards compatible?

Yes it should -  ffi.h is trying for maximum stability to make life easier for the various language wrappers.

> I couldn't figure out what is the meaning of the magic number in BOTAN_FFI_DECLARE_STRUCT. Is it just some random number that each struct has assigned to prevent some errors? What should I set it to?

Yes exactly. This is just a safeguard to tag each struct type, which prevent users from accidentally passing the wrong pointer type to a function. That's quite easy to do in some languages like C, Python, or Ruby where typing is loose. When you call `safe_get` or use `BOTAN_FFI_DO` macro, the magic value in the passed structs are checked against the expected value before the actual data field is dereferenced.

Any random bytes will do here, eg `botan rng 4` or `openssl rand -hex 4`.

Also please add CRL support in the Python wrapper and add some tests at the Python level as well. (This is a general rule for anything being added to ffi.h) If anything about how do that is unclear, @ me.

zacikpa:
Thanks for the quick answer. Python binding is now done as well, together with tests and documentation.

PR #2228, opened 2019-12-13
Extend PKCS#11 RSA paddings

neusdan:
Fixes: `Unexpected exception occured: PKCS#11 RSA sign/verify does not support EMSA EMSA4(SHA-512,MGF1,64)`


randombit:
This is needed for TLS with PKCS11 key + PSS right? Can this be tested in CI using SoftHSM?

randombit:
@neusdan ping re testing. I am ok merging this as is but it would be nice if we could add testing to prevent regressions here

Issue #2053, opened 2019-07-31
ASIO TLS Stream: Client only?


bazineta:
Looking over the new ASIO stream implementation in 2.11.0, I see examples and tests for client side, but not for server side. As this is a new implementation, I'm curious if server side is supported, or if this is a client-side only implementation. Thanks.

randombit:
I believe it is only client side right now @hrantzsch can you confirm?

hrantzsch:
Correct, the server side is not yet implemented. We won't get around to implement it in the near future I'm afraid.
@bazineta if you are interested in implementing it, I don't think it will be a huge amount of work and we will be happy to assist you. A good starting point would be to implement setting up the native_handle for the server side: https://github.com/randombit/botan/blob/fdf9970f921bf6b3e99c2a99ebc251b6e5dd760e/src/lib/tls/asio/asio_stream.h#L650

nunojpg:
@hrantzsch btw, DTLS is supported or do you have any comments about how to get it working?

hrantzsch:
DTLS is not supported by the Stream unfortunately. The first issue is that the Stream always instantiates the underlying `TLS::Client` with the default protocol version, which currently is TLS 1.2. (It would probably be good to do this more explicitly.)
The next, and more problematic, issue is that `boost::asio::udp::socket` does not offer the `read_some` and `write_some` methods that the Stream relies on.

reneme:
Server-side is now implemented and merged. Any feedback on the feature is highly appreciated. Otherwise, I believe, this ticket can be closed now, right?

PR #2229, opened 2019-12-16
Server Side Asio TLS Stream

hrantzsch:
This is PR allows wrapping TLS::Server using TLS::Stream.

In addition, the most notable change is the fix for the incorrect handling of stream truncation (aka short-read) errors.

Furthermore,

 * `async_shutdown()` is now available
 * trying to use the stream before performing the handshake fails gracefully
 * the stream can now be used with Beast's `tcp_stream`
 * the names of the error categories (for error codes) have changed

### DIFF
--- a/src/lib/tls/asio/asio_async_ops.h
+++ b/src/lib/tls/asio/asio_async_ops.h
@@ -1,7 +1,7 @@
 /*
 * Helpers for TLS ASIO Stream
-* (C) 2018-2019 Jack Lloyd
-*     2018-2019 Hannes Rantzsch, Tim Oesterreich, Rene Meusel
+* (C) 2018-2020 Jack Lloyd
+*     2018-2020 Hannes Rantzsch, Tim Oesterreich, Rene Meusel
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
@@ -144,24 +144,18 @@ class AsyncReadOperation : public AsyncBase<Handler, typename Stream::executor_t
                {
                // We have received encrypted data from the network, now hand it to TLS::Channel for decryption.
                boost::asio::const_buffer read_buffer{m_stream.input_buffer().data(), bytes_transferred};
-               try
-                  {
-                  m_stream.native_handle()->received_data(
-                     static_cast<const uint8_t*>(read_buffer.data()), read_buffer.size()
-                  );
-                  }
-               catch(const TLS_Exception& e)
-                  {
-                  ec = e.type();
-                  }
-               catch(const Botan::Exception& e)
-                  {
-                  ec = e.error_type();
-                  }
-               catch(...)
-                  {
-                  ec = Botan::ErrorType::Unknown;
-                  }
+               m_stream.process_encrypted_data(read_buffer, ec);
+               }
+
+            if (m_stream.shutdown_received())
+               {
+               // we just received a 'close_notify' from the peer and don't expect any more data
+               ec = boost::asio::error::eof;
+               }
+            else if (ec == boost::asio::error::eof)
+               {
+               // we did not expect this disconnection from the peer
+               ec = StreamError::StreamTruncated;
                }
 
             if(!m_stream.has_received_data() && !ec && boost::asio::buffer_size(m_buffers) > 0)
@@ -242,6 +236,12 @@ class AsyncWriteOperation : public AsyncBase<Handler, typename Stream::executor_
                return;
                }
 
+            if (ec == boost::asio::error::eof && !m_stream.shutdown_received())
+               {
+               // transport layer was closed by peer without receiving 'close_notify'
+               ec = StreamError::StreamTruncated;
+               }
+
             if(!isContinuation)
                {
                // Make sure the handler is not called without an intermediate initiating function.
@@ -293,28 +293,16 @@ class AsyncHandshakeOperation : public AsyncBase<Handler, typename Stream::execu
          {
          reenter(this)
             {
+            if(ec == boost::asio::error::eof)
+               {
+               ec = StreamError::StreamTruncated;
+               }
+
             if(bytesTransferred > 0 && !ec)
                {
                // Provide encrypted TLS data received from the network to TLS::Channel for decryption
                boost::asio::const_buffer read_buffer {m_stream.input_buffer().data(), bytesTransferred};
-               try
-                  {
-                  m_stream.native_handle()->received_data(
-                     static_cast<const uint8_t*>(read_buffer.data()), read_buffer.size()
-                  );
-                  }
-               catch(const TLS_Exception& e)
-                  {
-                  ec = e.type();
-                  }
-               catch(const Botan::Exception& e)
-                  {
-                  ec = e.error_type();
-                  }
-               catch(...)
-                  {
-                  ec = Botan::ErrorType::Unknown;
-                  }
+               m_stream.process_encrypted_data(read_buffer, ec);
                }
 
             if(m_stream.has_data_to_send() && !ec)
@@ -325,11 +313,10 @@ class AsyncHandshakeOperation : public AsyncBase<Handler, typename Stream::execu
                // operation will eventually call `*this` as its own handler, passing the 0 back to this call operator.
                // This is necessary because the check of `bytesTransferred > 0` assumes that `bytesTransferred` bytes
                // were just read and are available in input_buffer for further processing.
-               AsyncWriteOperation<
-               AsyncHandshakeOperation<typename std::decay<Handler>::type, Stream, Allocator>,
-                                       Stream,
-                                       Allocator>
-                                       op{std::move(*this), m_stream, 0};
+               AsyncWriteOperation<AsyncHandshakeOperation<typename std::decay<Handler>::type, Stream, Allocator>,
+                                   Stream,
+                                   Allocator>
+                                   op{std::move(*this), m_stream, 0};
                return;
                }
 
--- a/src/lib/tls/asio/asio_error.h
+++ b/src/lib/tls/asio/asio_error.h
@@ -1,7 +1,7 @@
 /*
 * TLS Stream Errors
-* (C) 2018-2019 Jack Lloyd
-*     2018-2019 Hannes Rantzsch, Tim Oesterreich, Rene Meusel
+* (C) 2018-2020 Jack Lloyd
+*     2018-2020 Hannes Rantzsch, Tim Oesterreich, Rene Meusel
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
@@ -22,6 +22,42 @@
 
 namespace Botan {
 namespace TLS {
+
+enum StreamError
+   {
+   StreamTruncated = 1
+   };
+
+class StreamCategory : public boost::system::error_category
+   {
+   public:
+      const char* name() const noexcept override
+         {
+         return "asio.ssl.stream";
+         }
+
+      std::string message(int value) const override
+         {
+         switch(value)
+            {
+            case StreamTruncated:
+               return "stream truncated";
+            default:
+               return "asio.botan.tls.stream error";
+            }
+         }
+   };
+
+inline const StreamCategory& botan_stream_category()
+   {
+   static StreamCategory category;
+   return category;
+   }
+
+inline boost::system::error_code make_error_code(Botan::TLS::StreamError e)
+   {
+   return boost::system::error_code(static_cast<int>(e), Botan::TLS::botan_stream_category());
+   }
 
 //! @brief An error category for TLS alerts
 struct BotanAlertCategory : boost::system::error_category
@@ -82,6 +118,11 @@ namespace boost {
 namespace system {
 
 template<> struct is_error_code_enum<Botan::TLS::Alert::Type>
+   {
+   static const bool value = true;
+   };
+
+template<> struct is_error_code_enum<Botan::TLS::StreamError>
    {
    static const bool value = true;
    };
--- a/src/lib/tls/asio/asio_error.h
+++ b/src/lib/tls/asio/asio_error.h
@@ -20,6 +20,13 @@
 #include <botan/tls_alert.h>
 #include <botan/tls_exceptn.h>
 
+/*
+ * This file defines Botan-specific subclasses of boost::system::error_category.
+ * In addition to the class definition, each category class is accompanied by function `make_error_code` used to create
+ * a `boost::system::error_code` of the category from some other kind of error in Botan (for example, a TLS alert).
+ * Since error_category instances should be singletons, there's also a method to get/create the instance for each class.
+ */
+
 namespace Botan {
 namespace TLS {
 
@@ -28,12 +35,13 @@ enum StreamError
    StreamTruncated = 1
    };
 
-class StreamCategory : public boost::system::error_category
+//! @brief An error category for errors from the TLS::Stream
+struct StreamCategory : public boost::system::error_category
    {
    public:
       const char* name() const noexcept override
          {
-         return "asio.ssl.stream";
+         return "Botan TLS Stream";
          }
 
       std::string message(int value) const override
@@ -43,7 +51,7 @@ class StreamCategory : public boost::system::error_category
             case StreamTruncated:
                return "stream truncated";
             default:
-               return "asio.botan.tls.stream error";
+               return "generic error";
             }
          }
    };
@@ -64,7 +72,7 @@ struct BotanAlertCategory : boost::system::error_category
    {
    const char* name() const noexcept override
       {
-      return "asio.botan.tls.alert";
+      return "Botan TLS Alert";
       }
 
    std::string message(int ev) const override
@@ -92,7 +100,7 @@ struct BotanErrorCategory : boost::system::error_category
    {
    const char* name() const noexcept override
       {
-      return "asio.botan.tls";
+      return "Botan";
       }
 
    std::string message(int ev) const override
@@ -114,6 +122,10 @@ inline boost::system::error_code make_error_code(Botan::ErrorType e)
 
 }  // namespace Botan
 
+ /*
+ * Add a template specialization of `is_error_code_enum` for each kind of error to allow automatic conversion to an
+ * error code.
+ */
 namespace boost {
 namespace system {
 
--- a/src/lib/tls/asio/asio_stream.h
+++ b/src/lib/tls/asio/asio_stream.h
@@ -29,7 +29,7 @@
 // which interferes with Botan's amalgamation by defining macros like 'B0' and 'FF1'.
 #define BOOST_ASIO_DISABLE_SERIAL_PORT
 #include <boost/asio.hpp>
-#include <boost/beast/core/flat_buffer.hpp>
+#include <boost/beast/core.hpp>
 
 #include <algorithm>
 #include <memory>
@@ -102,18 +102,31 @@ class Stream
       //! @{
 
       using next_layer_type = typename std::remove_reference<StreamLayer>::type;
-      using lowest_layer_type = typename next_layer_type::lowest_layer_type;
-      using executor_type = typename next_layer_type::executor_type;
-      using native_handle_type = typename std::add_pointer<ChannelT>::type;
-
-      executor_type get_executor() noexcept { return m_nextLayer.get_executor(); }
 
       const next_layer_type& next_layer() const { return m_nextLayer; }
       next_layer_type& next_layer() { return m_nextLayer; }
 
+#if BOOST_VERSION >= 107000
+      /*
+       * From Boost 1.70 onwards Beast types no longer provide public access to the member function `lowest_layer()`.
+       * Instead, the new free-standing functions in Beast need to be used.
+       * See also: https://github.com/boostorg/beast/commit/6a658b5c3a36f8d58334f8b6582c01c3e87768ae
+       */
+      using lowest_layer_type = typename boost::beast::lowest_layer_type<StreamLayer>;
+
+      lowest_layer_type& lowest_layer() { return boost::beast::get_lowest_layer(m_nextLayer); }
+      const lowest_layer_type& lowest_layer() const { return boost::beast::get_lowest_layer(m_nextLayer); }
+#else
+      using lowest_layer_type = typename next_layer_type::lowest_layer_type;
+
       lowest_layer_type& lowest_layer() { return m_nextLayer.lowest_layer(); }
       const lowest_layer_type& lowest_layer() const { return m_nextLayer.lowest_layer(); }
+#endif
 
+      using executor_type = typename next_layer_type::executor_type;
+      executor_type get_executor() noexcept { return m_nextLayer.get_executor(); }
+
+      using native_handle_type = typename std::add_pointer<ChannelT>::type;
       native_handle_type native_handle()
          {
          if(m_native_handle == nullptr)
--- a/src/lib/tls/asio/asio_stream.h
+++ b/src/lib/tls/asio/asio_stream.h
@@ -23,6 +23,7 @@
 #include <botan/tls_channel.h>
 #include <botan/tls_client.h>
 #include <botan/tls_magic.h>
+#include <botan/tls_server.h>
 
 // We need to define BOOST_ASIO_DISABLE_SERIAL_PORT before any asio imports. Otherwise asio will include <termios.h>,
 // which interferes with Botan's amalgamation by defining macros like 'B0' and 'FF1'.
@@ -39,8 +40,6 @@ namespace TLS {
 
 /**
  * @brief boost::asio compatible SSL/TLS stream
- *
- * Currently only the TLS::Client specialization is implemented.
  *
  * @tparam StreamLayer type of the next layer, usually a network socket
  * @tparam ChannelT type of the native_handle, defaults to Botan::TLS::Channel, only needed for testing purposes
@@ -200,7 +199,7 @@ class Stream
        * The function call will block until handshaking is complete or an error occurs.
        *
        * @param side The type of handshaking to be performed, i.e. as a client or as a server.
-       * @throws boost::system::system_error if error occured, or if the chosen Connection_Side is not available
+       * @throws boost::system::system_error if error occured
        */
       void handshake(Connection_Side side)
          {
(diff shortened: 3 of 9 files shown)


randombit:
Closes #2053 

Re testing, the simplest approach may be to implement a fully functioning server using this in the cli, then add some tests of it using Python (as part of `src/scripts/test_cli.py`). Currently in the CLI we have a TLS proxy (using asio), a HTTPS test server (asio) and an echo server (sockets). Possibly one of these could be adapted to using the ASIO TLS interface, but maybe easier to adapt your current test code to add some new functionality. Perhaps combine with Beast and offer a simple static file server over HTTPS?

Re timeouts, I'm not sure if how Botan behaves here interacts well with Beast or asio. But, essentially, timeouts are expected to be handled by the application because the TLS layer is not touching the network. So one strategy might be when starting to connect, set a timer that goes off in X seconds, when it fires check Channel::is_active and if still not active abort the connection.

Sessions being closed without an explicit close notify can be detected by overriding `tls_alert` and monitoring for a close notify, if seen set a boolean to true. When the socket closes, check the flag to see if the session was explicitly closed or was truncated.

hrantzsch:
Thanks for the input. Sounds very good--we'll do it like that :)

src/cli/tls_stream_http_server.cpp
```diff
@@ -0,0 +1,556 @@
...
+* (C) 2014,2015,2017,2019 Jack Lloyd
+* (C) 2016 Matthias Gierlings
+* (C) 2019 Hannes Rantzsch
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*
+*
+* This code is based on the Beast Example: HTTP SSL server, asynchronous.
+*
+* Copyright (c) 2016-2019 Vinnie Falco (vinnie dot falco at gmail dot com)
```
hrantzsch: This is a simple file server for the `botan-cli` based on the Beast example "HTTP SSL server, asynchronous". @randombit @vinniefalco can you confirm that including the Beast copyright notice is OK with both the license requirements of Botan and Beast?

vinniefalco: Looks fine to me, and I'm glad that the code is useful to you.

randombit: Alas we cannot take this. There are many organizations who carefully track exactly which license(s) are being used in their open source dependencies, I am in fairly regular contact with many large industrials, etc on this topic, and combining multiple licenses is a complication and uncertainty for them that I don't want to bring in.

For similar reasons Botan is unable to make use of MIT or BSD-3 licensed code even though of course MIT/BSD-2/BSD-3/BSL are all basically the same thing.

[One could argue that few of these orgs are paying me, so I shouldn't care too much about their concerns, but the main reason I maintain this lib at all is so something exists which can be usable in large commercial projects that is not named OpenSSL, and anything which deters such usage makes the work less satisfying for me on a personal level]

Sorry :/

vinniefalco: I understand.

> MIT/BSD-2/BSD-3/BSL are all basically the same thing.

Hmm, no not at all there is a world of difference. BSD-2 requires attribution in binaries while the BSL does not. BSD-2 is less permissive.

hrantzsch: > combining multiple licenses is a complication and uncertainty for them that I don't want to bring in

That makes sense. To my understanding though, we weren't releasing Botan or any part of it under BSL license. The code makes use of BSL licensed code, maintaining the copyright notice with the source code, as required. But it is still licensed BSD-2, like the rest of Botan, isn't it?

Did I misunderstand this, or can we solve this by a clearer wording?

hrantzsch:
Update on timeouts: like Jack said, timeouts should be handled by the application. It turns out that beast, for example, makes this very easy with `beast::tcp_stream::expires_after` ([see here how it's used in the example](https://github.com/randombit/botan/pull/2229/files#diff-cb84b72127de44326c930bb468cfc0cbR329)).

Unfortunately, a `TLS::Stream<beast::tcp_stream>` can currently not be used, because the `TLS::Stream` wants to access `lowest_layer_type`, which is private in the `tcp_stream`. I believe it would be very useful to users if we made our Stream compatible with Beast's `tcp_stream`.
For now, I was able to make it work by simply removing the (internally) unused accessors to `lowest_layer`. However, we might want to find another workaround, since we had the functions in the public API already.

hrantzsch:
I replaced the `test_cli.py`-based test by a unit(?) test of a simple client-server communication. The code of the file server lives on [here](https://github.com/hrantzsch/botan-tls-stream-server-example). I would like to link it as an example in the documentation.

randombit:
I will do a new review within next few days. CI is failing, looks like a Boost version issue?

hrantzsch:
Yeah, looks like beast is not where I thought it was.

We're not quite done, so no hurry with the review.

hrantzsch:
What's your opinion on the `lowest_layer` situation? The explanation is a bit longer I'm afraid:

Currently, our Stream offers the type `lowest_layer_type` and the function `lowest_layer()`, which gives you access to the underlying socket, for example to call `connect`. [`asio::ssl::stream`](https://www.boost.org/doc/libs/1_72_0/doc/html/boost_asio/reference/ssl__stream.html) also offers this.

In addition to `lowest_layer`, there's also `next_layer`. Both are the same if we wrap our Stream around a socket (`Botan::TLS::Stream<asio::ip::tcp::socket>`). They are different if, for example, we wrap our Stream around another stream (`Botan::TLS::Stream<beast::tcp_stream>`). In the latter case `lowest_layer` would be the socket but we have the `beast::tcp_stream` in between as the `next_layer`.

Unfortunately, we currently cannot use our Stream with `beast::tcp_stream`, because `beast::tcp_stream::lowest_layer_type` is private and we implemented `TLS::Stream::lowest_layer_type` simply as `next_layer_type::lowest_layer_type`.

I see three options here:
 * keep `Botan::TLS::Stream::lowest_layer()` and don't support `beast::tcp_stream`
    - this is my least favorite option, as I think `tcp_stream` is very nice to use
 * deviate from the interface of `asio::ssl::stream` and change our Stream's interface by removing `lowest_layer`
    - could break compatibility for someone, including our client example in the docs
    - users would probably need to use `beast::get_lowest_layer` to access the Stream's socket instead ([like here](https://github.com/hrantzsch/botan/blob/8dc5e9364305a4788c13dddcd15d2c77c23e5b34/src/tests/test_server_stream.cpp#L153))
 * find a workaround to offer `lowest_layer` without having to access `next_layer_type::lowest_layer_type`
    - this probably means re-implementing `beast::get_lowest_layer`, which recursively asks for `next_layer` until there is none

src/lib/tls/asio/asio_async_ops.h
```diff
@@ -149,6 +149,9 @@ class AsyncReadOperation : public AsyncBase<Handler, typename Stream::executor_t
                   m_stream.native_handle()->received_data(
                      static_cast<const uint8_t*>(read_buffer.data()), read_buffer.size()
                   );
+
+                  if (m_stream.native_handle()->is_closed())
+                     { ec = boost::asio::error::eof; }
```
hrantzsch: `native_handle()->is_closed()` tells means `received_data` has processed a `close_notify` alert and closed the channel. We set the EOF error to communicate this. Below, `map_error_code` might change that EOF to `StreamTruncated` (short read).
Can we make the assumption that we received `close_notify` from `is_closed()` here? Do I need to add more comments here?

In addition, `received_data` will have put the corresponding `close_notify` response into the send buffer.

src/lib/tls/asio/asio_error.h
```diff
@@ -23,6 +23,42 @@
 namespace Botan {
 namespace TLS {
 
+enum StreamError
+   {
+   StreamTruncated = 1
+   };
+
+class StreamCategory : public boost::system::error_category
```
hrantzsch: This category has only one error at the moment, but I think we need to make our own specific error for this case and cannot reuse some existing error_code.

src/lib/tls/asio/asio_error.h
```diff
@@ -56,6 +92,7 @@ struct BotanErrorCategory : boost::system::error_category
    {
    const char* name() const noexcept override
       {
+      // TODO: this seems wrong given the method brief
```
hrantzsch: Strictly speaking, this string is exposed to the outside and could be used by someone to check what kind of error occurred. But I think we can still change it, right? Maybe we should refine all the names. `StreamCategory`'s name looks inconsistent as well.

src/lib/tls/asio/asio_stream.h
```diff
@@ -110,6 +115,7 @@ class Stream
       const next_layer_type& next_layer() const { return m_nextLayer; }
       next_layer_type& next_layer() { return m_nextLayer; }
 
+      // TODO: since we are using beast already, use beast::get_lowest_layer for beast::tcp_stream compatibility
```
hrantzsch: This is needed to make the beast file server (now moved to another repo) work. But we can leave it for another PR.

src/lib/tls/asio/asio_stream.h
```diff
@@ -214,31 +220,23 @@ class Stream
          {
          setup_native_handle(side, ec);
 
-         // send client hello, which was written to the send buffer on client instantiation
-         send_pending_encrypted_data(ec);
+         if(side == CLIENT)
```
hrantzsch: This check isn't actually needed, as `send_pending_encrypted_data` won't have anything to send if there's no client native handle that put the client-hello into the send buffer.
In fact, the async handshake doesn't explicitly check this. Should we try to `send_pending...` unconditionally?

src/lib/tls/asio/asio_stream.h
```diff
@@ -516,6 +537,18 @@ class Stream
 
       //! @}
 
+      // TODO: do we need these public? asio ssl stream doesn't have them, but they are very useful for testing
```
hrantzsch: as it says in the TODO

src/lib/tls/asio/asio_stream.h
```diff
@@ -533,28 +566,32 @@ class Stream
       class StreamCore : public Botan::TLS::Callbacks
          {
          public:
-            StreamCore(boost::beast::flat_buffer& receive_buffer, boost::beast::flat_buffer& send_buffer, Context& context)
-               : m_receive_buffer(receive_buffer), m_send_buffer(send_buffer), m_tls_context(context) {}
+            StreamCore(Stream& stream) : m_stream(stream) {}
```
hrantzsch: I'm not really happy with using the stream reference here directly, and also not with accessing the private members of the stream directly below. This was changed because we need to communicate receiving `close_notify` to the stream in some way. We could keep the parameters as they were and add an additional `std::function<> alert_callback` or similar, but I'm not sure if it's worth it.

src/lib/tls/asio/asio_stream.h
```diff
@@ -676,25 +721,72 @@ class Stream
...
+            }
+         }
+
+      /**
+       * @brief Map an EOF error_code returned by the underlying transport according to state of the SSL session.
+       *
+       * Returns a const reference to the error code object, suitable for passing to a completion handler.
+       * This method corresponds to the method map_error_code in Boost.Asio's SSL engine.
+       */
+      const boost::system::error_code map_error_code(boost::system::error_code& ec)
```
hrantzsch: The purpose of this function is to map an EOF error to a StreamTruncated error when appropriate. The Boost.Asio method mentioned above can be [found here](https://github.com/boostorg/asio/blob/4f11be2/include/boost/asio/ssl/detail/impl/engine.ipp#L199-L223).
It is called in async operations always before calling the handler (before `complete_now`) and in synchronous methods after asio::read and asio::write.

src/lib/tls/asio/asio_stream.h
```diff
@@ -676,25 +721,72 @@ class Stream
...
+         {
+         // We only want to map the error::eof code.
+         if(ec != boost::asio::error::eof)
+            {
+            return ec;
+            }
+
+         // TODO
```
hrantzsch: I am very confused by this one. @reneme @Tymolc let's talk about it in the chat.

src/tests/test_server_stream.cpp
```diff
@@ -0,0 +1,520 @@
...
+         results.insert(results.end(), r3.cbegin(), r3.cend());
+
+         auto r4 = run_test_case<Test_No_Shutdown_Response>();
+         results.insert(results.end(), r4.cbegin(), r4.cend());
+
+         return results;
+         }
+   };
+
+BOTAN_REGISTER_TEST("tls_server_stream", Tls_Server_Stream_Tests);
```
hrantzsch: This test is currently misnamed. Suggestions?

src/tests/test_server_stream.cpp
```diff
@@ -0,0 +1,520 @@
...
+
+static std::string server_cert() { return Botan_Tests::Test::data_dir() + "/x509/certstor/cert1.crt"; }
+static std::string server_key() { return Botan_Tests::Test::data_dir() + "/x509/certstor/key01.pem"; }
+
+class Timeout_Exception : public std::runtime_error
+   {
+      using std::runtime_error::runtime_error;
+   };
+
+struct Side
```
hrantzsch: This struct bundles all the Botan components that both the server and the client will need.

src/tests/test_server_stream.cpp
```diff
@@ -0,0 +1,520 @@
...
+         {
+         m_result.confirm(msg, condition);
+         }
+
+   private:
+      net::system_timer m_timer;
+      Result m_result;
+   };
+
+class Server : public Side, public std::enable_shared_from_this<Server>
```
hrantzsch: The server side is rather passive in all tests. It almost always performs the same checks. `expect_short_read` and `shutdown` allow tests cases to change it's behavior when needed.

src/tests/test_server_stream.cpp
```diff
@@ -0,0 +1,520 @@
...
+
+      void handle_write(const error_code& ec)
+         {
+         m_result.check_ec("send_response", ec);
+         m_result.set_timer("read_message");
+         net::async_read(*m_stream, buffer(),
+                         std::bind(&Server::handle_read, shared_from_this(), _1, _2));
+         }
+
+      void handle_read(const error_code& ec, size_t bytes_transferred=0)
```
hrantzsch: As long as no error occurs, the server will simply echo what it read, looping between `handle_write` and `handle_read`.

src/tests/test_server_stream.cpp
```diff
@@ -0,0 +1,520 @@
...
+
+   private:
+      Client m_client;
+      Result_Wrapper m_result;
+      std::shared_ptr<Server> m_server;
+   };
+
+#include <boost/asio/unyield.hpp>
+
+template<typename TestT>
```
hrantzsch: If you find a more elegant solution than my test case classes, I'll be very grateful.

src/tests/unit_asio_stream.cpp
```diff
@@ -177,7 +178,7 @@ class Asio_Stream_Tests final : public Test
          {
          net::io_context ioc;
          // fail right away
-         FailCount  fc{0, net::error::eof};
+         FailCount  fc{0, net::error::no_recovery};
```
hrantzsch: These were changed to avoid confusion when `eof` is mapped to `StreamTruncated`. The error code used here is chosen arbitrarily anyway.

src/lib/tls/asio/asio_stream.h
Tymolc: "Testing" as in "Debugging"?
But the stream should just work (tm), right? :) I feel like if a user has to read those flags, they are probably deep enough in the code to just print them themselves, but I don't have a strong feeling 🤷‍♂

src/lib/tls/asio/asio_error.h
hrantzsch: Because boost asio ssl is the openssl mplementation / requires openssl and that is what we want to replace.

src/lib/tls/asio/asio_stream.h
hrantzsch: No, I mean I find them really useful in `test_server_stream` in order to verify the `close_notify` was sent/received.

src/tests/test_server_stream.cpp
```diff
@@ -0,0 +1,520 @@
...
+
+            m_result.set_timer("receive_response");
+            // Start the async_read but do not yield for it to complete. Instead, close the socket and expect the read
+            // to be aborted.
+            auto self = shared_from_this();
+            /* no yield! */ net::async_read(m_client.stream(), m_client.buffer(),
+                                            [this, self](const error_code &ec, size_t)
+               {
+               // check that "waiting" for the server's shutdown turns out to be aborted
+               m_result.confirm("async_read is aborted", ec == net::error::operation_aborted);
```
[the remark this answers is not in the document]
hrantzsch: Yes, and that the error_code is propagated appropriately. Both this check and the `!shutdown_received()` check below are not super interesting I guess, but it confirms the expected behavior ;)

src/lib/tls/asio/asio_stream.h
hrantzsch: It's not really possible to commit to the send_buffer before doing the handshake, because the `handshake()` is what creates the underlying client or server `Channel`. However, I found that trying to do so was not handled gracefully: https://github.com/randombit/botan/pull/2229/commits/b8f2f06631885a7933ec700e42f0d094f1876a30

Is `Invalid_State` the right exception to use here?

src/lib/tls/asio/asio_stream.h
```diff
@@ -335,22 +328,66 @@ class Stream
...
+       *       [h = std::forward<Handler>(handler)] (...) mutable { return h(ec); }
+       */
+      template <typename Handler>
+      class Wrapper
+         {
+         public:
+            Wrapper(Handler&& handler) : _handler(std::forward<Handler>(handler)) {}
+            void operator()(boost::system::error_code ec, size_t)
+               {
+               if(!ec)
```
hrantzsch: Alternatively, we could try to check `native_handle()->is_closed()` in `AsyncWriteOperation`.

src/lib/tls/asio/asio_stream.h
Tymolc:
```
// Invalid state was encountered. A request was made on an object while the
// object was in a state where the operation cannot be performed.
```

Sounds good to me 👍

hrantzsch:
Review hints for @reneme @tymolc:

The two major issues still to be addressed are the failing tests on Windows and the correct handling SSL short read errors via `map_error_code`.

Tests on Windows pass occasionally, but fail with an `bad_record_mac` error most of the time. There's very strong indication that this is a race in the asynchronous test code.

The desired behavior regarding short read errors is described very well here: https://stackoverflow.com/a/25703699. We need to check that the current strategy of setting `eof` (in sync and async read operations) and mapping it to `StreamTruncated` if required is sound.

A final change I'd like to make is to rename the `test_server_stream` test case more appropriately. Still looking for a good name.

hrantzsch:
@randombit I would consider this PR ready for review now.

Enabling the stream to be used as the server-side is actually the smallest part in this PR.

Most of it is concerned with correctly handling stream truncation a.k.a. short-read errors. This is a fix that is probably relevant for the client-side as well.
Before now, we failed to inform users of the stream when an SSL connection was not closed properly. This can facilitate a truncation attack.
Now the stream transforms `EOF` errors to `StreamTruncated` errors when appropriate. Thanks @reneme for the help in sorting out all the cases.

In addition we added an "integration test" instead of the previously discussed test in `test_cli.py`. The test instantiates two streams, one as a client and one as a server, and lets them talk on a local socket.

Minor things remain to be done:

 * [ ] the integration test runs too long on Linux
 * [ ] naming of error categories ([discussion](https://github.com/randombit/botan/pull/2229/files#r374548865))
 * [ ] shutdown_sent is not set correctly in async shutdown ([discussion](https://github.com/randombit/botan/pull/2229/files#r375107425)); `shutdown_sent()` method is also never used
 * [ ] squash commits

reneme:
@hrantzsch I hate to say it, but the test we tried to fix today morning, still fails on windows :(

```
tls_stream_integration:
Server ran 52 tests 1 FAILED
Failure 1: Server handshake with error Unexpected error code: stream truncated
Test Close Without Shutdown ran 3 tests all ok
Test Close Without Shutdown Sync ran 2 tests all ok
Test Conversation ran 8 tests all ok
Test Conversation Sync ran 8 tests all ok
Test Eager Close ran 4 tests all ok
Test Eager Close Sync ran 4 tests all ok
Test No Shutdown Response ran 5 tests 2 FAILED
Failure 1: Test No Shutdown Response read gives EOF with error Unexpected error code: bad_record_mac
Failure 2: Test No Shutdown Response received close_notify produced unexpected result '0' expected '1'
Test No Shutdown Response Sync ran 5 tests all ok
```

src/lib/tls/asio/asio_stream.h
reneme: We decided to get rid of it.

reneme:
@randombit In case you're already awake: It would be cool to kill this (deadlocked) AppVeyor task: https://ci.appveyor.com/project/randombit/botan/builds/31039695

src/lib/tls/asio/asio_async_ops.h
```diff
@@ -325,11 +313,10 @@ class AsyncHandshakeOperation : public AsyncBase<Handler, typename Stream::execu
...
                // were just read and are available in input_buffer for further processing.
-               AsyncWriteOperation<
-               AsyncHandshakeOperation<typename std::decay<Handler>::type, Stream, Allocator>,
-                                       Stream,
-                                       Allocator>
-                                       op{std::move(*this), m_stream, 0};
+               AsyncWriteOperation<AsyncHandshakeOperation<typename std::decay<Handler>::type, Stream, Allocator>,
+                                   Stream,
+                                   Allocator>
+                                   op{std::move(*this), m_stream, 0};
```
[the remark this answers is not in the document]
hrantzsch: Well, I found the previous formatting quite misleading. I moved things around until the formatter and I could agree on this, in my mind less confusing, version :)

src/lib/tls/asio/asio_async_ops.h
```diff
@@ -144,24 +144,18 @@ class AsyncReadOperation : public AsyncBase<Handler, typename Stream::executor_t
...
-                  }
-               catch(const Botan::Exception& e)
-                  {
-                  ec = e.error_type();
-                  }
-               catch(...)
-                  {
-                  ec = Botan::ErrorType::Unknown;
-                  }
+               m_stream.received_data(read_buffer, ec);
```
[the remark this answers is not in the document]
hrantzsch: Renamed and commented. Hope it's clearer now.

src/lib/tls/asio/asio_stream.h
```diff
@@ -295,29 +291,21 @@ class Stream
...
-            {
-            ec = e.error_type();
-            }
-         catch(const std::exception&)
-            {
-            ec = Botan::ErrorType::Unknown;
-            }
+            }, ec);
+
+         send_pending_encrypted_data(ec);
```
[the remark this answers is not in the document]
hrantzsch: It's probably not the most consistent use of error code. But it's very convenient in a number of places, where we'd have to check the error code before proceeding otherwise.
I added a comment to the method.
Do you think we should change it?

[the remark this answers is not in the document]
reneme: No, I think that's fine. I just pointed it out for further review passes to keep it off my head. The comment is helpful anyway. Thanks.

src/lib/tls/asio/asio_async_ops.h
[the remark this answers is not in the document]
reneme: Check.

src/lib/tls/asio/asio_async_ops.h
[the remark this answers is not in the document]
reneme: Thanks.

hrantzsch:
> Probably we should try to build a simple server based on the TLS server code paths before finishing this up.

I made the Stream use Beast functionality to access its lowest layer. This allows us to run Beast's HTTP server example with our Stream rather than Asio's ssl_stream.

hrantzsch:
I decided to change the names of the error categories. I didn't find any hint that they should follow a specific format (as they seemed to do before). [This official example](https://www.boost.org/doc/libs/1_71_0/libs/outcome/doc/html/motivation/plug_error_code2.html) seems to use a free-text name.

I believe the naming is reasonable now. **Note** however, the name of the error categories is public. Theoretically, it was possible for users to
```
if (errorcode.category().name() == "old name") { }
```
Although this is really not a sensible way to do it.

src/lib/tls/asio/asio_error.h
hrantzsch: Addressed in the comment below.

randombit:
@hrantzsch :+1: I will review sometime before next Monday.

src/tests/test_tls_stream_integration.cpp
```diff
@@ -0,0 +1,722 @@
...
+
+static std::string server_cert() { return Botan_Tests::Test::data_dir() + "/x509/certstor/cert1.crt"; }
+static std::string server_key() { return Botan_Tests::Test::data_dir() + "/x509/certstor/key01.pem"; }
+
+class Timeout_Exception : public std::runtime_error
+   {
+      using std::runtime_error::runtime_error;
+   };
+
+struct Side
```
randombit: Should be a `class`, use `struct` only for POD with no constructors

src/tests/test_tls_stream_integration.cpp
```diff
@@ -0,0 +1,722 @@
...
+* TLS ASIO Stream Client-Server Interaction Test
+* (C) 2018-2020 Jack Lloyd
+*     2018-2020 Hannes Rantzsch, Rene Meusel
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include "tests.h"
+
+#if defined(BOTAN_HAS_TLS) && defined(BOTAN_HAS_TLS_ASIO_STREAM)
```
randombit: You'll also want `BOTAN_TARGET_OS_HAS_THREADS` so this test is disabled on baremetal, IncludeOS etc

src/tests/test_tls_stream_integration.cpp
hrantzsch: Done. Same for Result_Wrapper just below, I assume.

hrantzsch:
Alright, thanks @randombit. CI also seems fine after rebase.

Issue #2197, opened 2019-11-14
Spurious failure in GOST test


randombit:
Only ever seen this in CI, not been able to reproduce:

gost_3410_sign:
pubkey/gost_3410_sign.vec ran 1 tests 1 FAILED
Failure 1: Test # 1  failed with exception 'OID::encode_into: OID is invalid' [Key=01456C64BA4642A1653C235A98A60249BCD6D3F746B631DF928014F6C5BF9C4041AA28D2F1AB148280CD9ED56FEDA41974053554A42767B83AD043FD39DC0493]

It seems more common on the Aarch64 fuzzer build which has 32 cores, so it may be some kind of threading bug. The EC group being tested here is a custom registered one which I suspect is part of the issue.

randombit:
Possibly this can be resolved by just serializing the test execution like we do for things that manipulate CPUID. But I'd like to understand what the actual issue is since it seems to indicate some kind of threading bug.

PR #2233, opened 2020-01-03
install the DLL to bin-dir rather than lib-dir

hrantzsch:
On Windows, the DLL should be installed to the binary directory instead of the library directory. Otherwise it will not be found in the `PATH` when trying to link Botan to the using code.

### Context

We are currently trying to get Botan included in Conan's package index. The respective PR [can be found here](conan-io/conan-center-index#437).
Up until now the [conan-botan recipe](https://github.com/bincrafters/conan-botan) used the workaround to declare the lib folder (that contained the dll) as a binary directory. However, such workaround is not desired for packages in the official index.

### Possibly Breaking

If somebody relied on finding the DLL in the library folder, this could possibly break something for them.
Do you think this is a problem and can we ask users to adjust?

### `--libdir` and `--bindir`

Also note that we are aware of the `--libdir` option, but currently it places both the `.lib` and the `.dll` in the given directory, while actually the DLL should always go to the binary directory.
Now the DLL placement will be determined by `--bindir`.

### DIFF
--- a/src/scripts/install.py
+++ b/src/scripts/install.py
@@ -191,7 +191,7 @@ def main(args):
             libname = cfg['libname']
             soname_base = libname + '.dll'
             copy_executable(os.path.join(out_dir, soname_base),
-                            prepend_destdir(os.path.join(lib_dir, soname_base)))
+                            prepend_destdir(os.path.join(bin_dir, soname_base)))
         else:
             soname_patch = cfg['soname_patch']
             soname_abi = cfg['soname_abi']


reneme:
There's one test failure on Linux on arm64. I doubt that this change is responsible for it, though:
```
pubkey/gost_3410_sign.vec ran 1 tests 1 FAILED
Failure 1: Test # 1  failed with exception 'OID::encode_into: OID is invalid' [Key=01456C64BA4642A1653C235A98A60249BCD6D3F746B631DF928014F6C5BF9C4041AA28D2F1AB148280CD9ED56FEDA41974053554A42767B83AD043FD39DC0493]
```

randombit:
It's ok to change how the build/installation process works as long as it is documented in the release notes. This will probably break someone's build script somewhere, but SemVer only applies to the public API surface not the build process. Here it seems like DLL in the binary is the correct thing to do, so we should do it. I'll mention it in the relnotes.

Test failure is a known flaky test #2197  I have not been able to figure out the problem.

By the way thanks for doing the work getting us into conan index - anything that makes it easier to install and use by developers means more users, which hopefully (eventually) means more development resources coming our way.

randombit:
I accidentally merged this without `--no-ff` which caused it to be commited with a different id dbce0c6c8b9781bf776c14729694520d09e6af5f so github doesn't see this as merged.

PR #2269, opened 2020-02-14
Update TLS policy default values in docs

nametoolong:
It seems that `TLS::Policy::allow_client_initiated_renegotiation` has been changed to `false` for some time, but the documentation is somehow not updated.
https://github.com/randombit/botan/blob/master/src/lib/tls/tls_policy.cpp#L332


randombit:
Thank you!

PR #2274, opened 2020-02-16
Default to NoPadding if no padding specified

hekkihek:
commoncrypto_opts_from_algo is called from make_commoncrypto_block_cipher, on creating a block cipher no mode/padding is specified which made commoncrypto_opts_from_algo always throw an "Unsupported cipher mode padding!" exception. This issue was introduced when PKCS7 padding was disabled.


hekkihek:
Apparently one of the builds fails, I'll look into it and submit a new pull request.

PR #2275, opened 2020-02-16
Commoncrypto incorrect minimum keysize Blowfish

hekkihek:


hekkihek:
kCCKeySizeMinBlowfish is defined as 8 by CommonCrypto

randombit:
I don't think this is right, because we adjust the key length in commoncrypto_adjust_key_size. For Blowfish, short keys are just repeated enough times to fill out the entire block. So if given a very short key (say two bytes AB) we repeat it to ABABABAB to satisfy the commoncrypto minimum, then commoncrypto is happy but we have the same effect as using a short key.

Is there a specific issue you are seeing due to this? If so we should have a test that triggers it.

hekkihek:
You are quite right, I didn't spot the commoncrypto_adjust_key_size function. My use case has 448 bit keys exclusively, I just spotted the line that looked as an error on first sight so thought I'd be better to report it.

Just read https://www.schneier.com/academic/archives/1994/09/description_of_a_new.html) and this comment is pretty clear: "For every short key, there is at least one equivalent longer key; for example, if A is a 64-bit key, then AA, AAA, etc., are equivalent keys.".

I'll close this pull request.

Issue #2255, opened 2020-02-05
PKCS8::load_key is affected by a compiler bug in clang 8


reneme:
## TL;DR

This is an informational ticket in the hope of helping other people that might stumble over the issue. A [bug in clang 8](https://bugs.llvm.org/show_bug.cgi?id=41810) can cause programs using Botan to crash when calling `PKCS8::load_key` with an incorrect password of a certain length.

Feel free to close this right away.

## Symptoms

We recently received a bug report claiming that our application crashes for long (>800 characters) user-provided passwords. We quickly realized that crashes occur for (wrong) passwords that are longer than only 23 (!) characters.

This was reproducible and the stack trace looked roughly like that:

```
spike(44594,0x10081d5c0) malloc: *** error for object 0x100a07340: pointer being freed was not allocated
spike(44594,0x10081d5c0) malloc: *** set a breakpoint in malloc_error_break to debug

* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007fff720962c2 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff72151bf1 libsystem_pthread.dylib`pthread_kill + 284
    frame #2: 0x00007fff720006a6 libsystem_c.dylib`abort + 127
    frame #3: 0x00007fff7210f077 libsystem_malloc.dylib`malloc_vreport + 545
    frame #4: 0x00007fff7210ee38 libsystem_malloc.dylib`malloc_report + 151
    frame #5: 0x00000001001b11fc spike`Botan::PKCS8::load_key(Botan::DataSource&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) [inlined] std::__1::_DeallocateCaller::__do_call(__ptr=<unavailable>) at new:320:12 [opt]
    frame #6: 0x00000001001b11f7 spike`Botan::PKCS8::load_key(Botan::DataSource&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) [inlined] std::__1::_DeallocateCaller::__do_deallocate_handle_size(__ptr=<unavailable>) at new:278 [opt]
[...]

PR #2277, opened 2020-02-17
Avoid double-free in PKCS8::load_key

hrantzsch:
This PR adds a test case to reproduce #2255 and a fix to avoid the issue.

 * [x] reproduce the issue on CI
 * [x] "fix" load_key to avoid the issue
 * [x] ~~run test only on clang 8 (or not?)~~

### DIFF
--- a/src/tests/test_clang_bug.cpp
+++ b/src/tests/test_clang_bug.cpp
@@ -1,0 +1,117 @@
+/*
+* (C) 2020 Jack Lloyd, René Meusel, Hannes Rantzsch
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+/*
+ * This test case ensures that we avoid crashing in PKCS8::load_key due to a bug in Clang 8.
+ *
+ * A detailed description of the issue can be found here: https://github.com/randombit/botan/issues/2255.
+ * In short, Clang 8 performs a double-free on captured objects if an exception is thrown within a lambda. This case can
+ * occur when PKCS8::load_key fails to load the key due to a wrong password. The password needs to be long enough to not
+ * be small buffer optimized.
+ *
+ * The Clang bug is tracked here and apparently won't be fixed in Clang 8: https://bugs.llvm.org/show_bug.cgi?id=41810.
+ * Other Clang versions seem not to be affected.
+ */
+
+#include "tests.h"
+
+#if defined(BOTAN_HAS_PUBLIC_KEY_CRYPTO)
+   #include <botan/system_rng.h>
+   #include <botan/pkcs8.h>
+   #include <botan/data_src.h>
+#endif // BOTAN_HAS_PUBLIC_KEY_CRYPTO
+
+namespace Botan_Tests {
+
+#if defined(BOTAN_HAS_PUBLIC_KEY_CRYPTO)
+
+namespace {
+
+// The key needs to be a PKCS #8 encoded and encrypted private key.
+std::string getPrivateKey()
+   {
+   return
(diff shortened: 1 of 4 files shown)


src/tests/test_clang_bug.cpp
```diff
@@ -0,0 +1,119 @@
...
+   {
+   public:
+      std::vector<Test::Result> run() override
+         {
+         Test::Result result("PKCS8::load_key does not crash when compiled with Clang 8");
+
+         Botan::System_RNG rng;
+
+         auto pem = getPrivateKey();
+         std::string pw = "01234567890123456789"; // sufficiently long password
```
reneme: Sure that this is sufficiently long? On macOS I needed at least 23 bytes (including the `\0` byte, I believe).

hrantzsch: It was enough in my local tests.
The strange thing is, testing the MVE you posted in the issue, very long passwords (~100 chars) didn't cause a crash but the issue was still visible in valgrind. 20 and 23 characters did cause a crash (double-free).

hrantzsch:
The bug was reproduced in the first commit, before adding the fix. [Here's the failed job](https://travis-ci.org/randombit/botan/jobs/651382999?utm_medium=notification&utm_source=github_status).

Regarding running the test only on the clang8 target, what's the best way to do this?
Also, I noticed that there already is a Valgrind target in the travis matrix. Should we also use `--with-valgrind` just to be sure, or is it enough that we've seen it crashing in the job linked above?

src/tests/test_clang_bug.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+         Botan::DataSource_Memory ds(reinterpret_cast<const uint8_t*>(pem.data()), pem.size());
+
+         try
+            {
+            Botan::PKCS8::load_key(ds, rng, pw);
+            }
+         catch(const std::exception&) { }
```
reneme: Should we fail the test if `PKCS8::load_key` reports anything but "wrong password"?

src/tests/test_clang_bug.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+#if defined(BOTAN_HAS_PUBLIC_KEY_CRYPTO)
+
+namespace {
+
+// The key needs to be a PKCS #8 encoded and encrypted private key.
+std::string getPrivateKey()
+   {
+   return
+      "-----BEGIN ENCRYPTED PRIVATE KEY-----"
```
reneme: For future reference: Do you remember the correct password? The test itself must use a wrong password, you mention that in the long description above. Future-me would find it helpful to see the "correct" password as another hint when quickly reading this test case

reneme:
> run test only on clang 8

Actually, I think it doesn't hurt to run this regression test on all compilers.

src/tests/test_clang_bug.cpp
hrantzsch: Hm, initially I had it fail if there's no exception, but now I think that's not really what we want to test here. So I left it out to avoid confusion. I'm pretty sure there's another test that ensures this already.

randombit:
Looks great thanks. I agree there is no reason to only run this under Clang 8 given the test should run very quickly. Always running it also ensures this never regresses under another compiler (at least not without us noticing..)

hrantzsch:
I changed the key for one where I still know the password and noted it, as @reneme suggested. Also, I changed my mind about checking whether `load_key` throws. The test now fails if it doesn't throw and succeeds only if it does.

All done from my side :)

PR #2278, opened 2020-02-18
CommonCrypto: improve name parsing for block cipher

hekkihek:
make_commoncrypto_block_cipher called commoncrypto_opts_from_algo to parse the supplied name.
commoncrypto_opts_from_algo requires the supplied string to contain a non-empty padding which made make_commoncrypto_block_cipher always fail.
The logic to parse just the algo_name has been moved to a seperate function commoncrypto_opts_from_algo_name.

### DIFF
--- a/src/lib/prov/commoncrypto/commoncrypto_block.cpp
+++ b/src/lib/prov/commoncrypto/commoncrypto_block.cpp
@@ -139,7 +139,7 @@ make_commoncrypto_block_cipher(const std::string& name)
 
    try
       {
-      CommonCryptor_Opts opts = commoncrypto_opts_from_algo(name);
+      CommonCryptor_Opts opts = commoncrypto_opts_from_algo_name(name);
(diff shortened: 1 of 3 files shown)


randombit:
Thanks I'll review this asap but may be a day or two.

hekkihek:
To provide some more context to ease reviewing:

Compiling for iOS with --with-commoncrypto

std::unique_ptr<Botan::BlockCipher> cipher = Botan::BlockCipher::create("AES-256");
NSLog(@"DEBUG: provider: %s", cipher->provider().c_str());

Using the master branch this outputs
DEBUG: provider: base

So, although commoncrypto support is compiled in, create() still returns the base provided implementation (due to the issue described in the pull request).
Explicitly requesting the commoncrypto implementation (since we now support is compiled in) using create("AES-256", "commoncrypto") will even return nullptr.

With the patch applied:
DEBUG: provider: commoncrypto

Issue #2268, opened 2020-02-14
MacOS 10.15: system RNG test crashes


mouse07410:
MacOS Catalina 10.15.3, Xcode-11.3.1, current Botan master.

Builds OK, but fails RSA tests:
```
. . . . .
rsa_kem:
RSA/KDF1-18033(SHA-1) KEM ran 6 tests in 23.70 msec all ok
RSA/KDF1-18033(SHA-256) KEM ran 6 tests in 38.40 msec all ok
RSA/KDF1-18033(SHA-512) KEM ran 3 tests in 20.54 msec all ok
RSA/KDF2(SHA-1) KEM ran 6 tests in 28.22 msec all ok
RSA/KDF2(SHA-256) KEM ran 6 tests in 35.69 msec all ok
RSA/KDF2(SHA-512) KEM ran 3 tests in 19.77 msec all ok
rsa_keygen:
Abort trap: 6
```

Here's the crash data:
```
Thread 4 Crashed:
0   libsystem_kernel.dylib        	0x00007fff657767fa __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff65833bc1 pthread_kill + 432
2   libsystem_c.dylib             	0x00007fff656fda1c abort + 120
3   libcorecrypto.dylib           	0x00007fff6546dd78 cc_abort + 9
4   libcorecrypto.dylib           	0x00007fff65496cdb cc_try_abort + 9
5   libcorecrypto.dylib           	0x00007fff6542792c ccrng_cryptographic_generate + 538
6   botan-test                    	0x0000000105690175 Botan_Tests::(anonymous namespace)::System_RNG_Tests::run() + 1317
7   botan-test                    	0x00000001056a3c8d Botan_Tests::(anonymous namespace)::run_a_test(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 157
[...]

PR #2279, opened 2020-02-19
Fix a crash in System_RNG on macOS 10.15

randombit:
[Potential fix, need @mouse07410 or someone else with access to macOS 10.15 to confirm this works]

See #2268

### DIFF
--- a/src/lib/rng/system_rng/system_rng.cpp
+++ b/src/lib/rng/system_rng/system_rng.cpp
@@ -119,7 +119,11 @@ class System_RNG_Impl final : public RandomNumberGenerator
 
       void randomize(uint8_t buf[], size_t len) override
          {
-         ::arc4random_buf(buf, len);
+         // macOS 10.15 arc4random crashes if called with len == 0
+         if(len > 0)
+            {
+            ::arc4random_buf(buf, len);
+            }
          }
 
       bool accepts_input() const override { return false; }


reneme:
LGTM. Though, I cannot reproduce the crash on macOS 10.15.3 with or without this patch. I used the following test program that might be too naive?:

```
#include <botan/system_rng.h>
#include <botan/hex.h>

#include <iostream>

int main() {
    uint8_t buffer[32] = {0};

    auto &rng = Botan::system_rng();

    std::cout << rng.name() << std::endl;
    std::cout << Botan::hex_encode(buffer, 32) << std::endl;

    rng.randomize(buffer, 0);

    std::cout << Botan::hex_encode(buffer, 32) << std::endl;

    return 0;
}
```

randombit:
@reneme My understanding is your prog should be triggering this. Possibly it is only a bug in certain patch levels of 10.15? Another repro would be

```
#include <stdlib.h>

int main() {
  unsigned char buf[32] = { 0 };
  arc4random_buf(buf, 0);
  return 0; // didn't crash ...
}
```

Also are you sure `System_RNG` is using arc4random in your build? It should be unless configured with an explicit disable of ``--without-os-features=arc4random`` - you can confirm by looking for `BOTAN_TARGET_OS_HAS_ARC4RANDOM` in ``build.h``

randombit:
In #2268 @mouse07410 also reports using Catalina 10.15.3 so wtf? Why does it only reproduce on some machines.

reneme:
I wasn’t sure about arc4random, hence the `rng.name()` print to confirm. Also a printf in `System_RNG_Impl::randomize()` got triggered. So yeah, I’m pretty sure that I went into the

mouse07410:
First, the test-code was
```c++
std::vector<uint8_t> out_buf(i);
```
Is that the same as 
```c++
uint8_t buffer[32] = {0};
```
I think not.

@reneme does `./botan-test system_rng` work for you? Are you on Xcode-11.3.1?

Finally, even if the #2268 problem doesn't manifest on *all* the Macs, this PR fixes the problem, and doesn't seem to cause any hurt elsewhere. Let's merge (I've merged it in my fork regardless).

randombit:
Oh good point @mouse07410, the problem is likely specifically `arc4random_buf(nullptr, 0)` crashing.

Anyway agree fix is simple and seems to resolve the issue. Just waiting for CI to merge.

randombit:
Ok CI has already finished just isn't reporting as such probably because of the outage GH had earlier today.

randombit:
@reneme Just for curiosity, can you check `arc4random_buf(nullptr, 0)` on your machine?

reneme:
@randombit @mouse07410 Will do tomorrow morning.

reneme:
This:
```C++
#include <stdlib.h>

int main() {
    arc4random_buf(nullptr, 0);
    return 0;
}
```
comes back with a crash like that:
```
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007fff6ed4d7fa libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff6ee0abc1 libsystem_pthread.dylib`pthread_kill + 432
    frame #2: 0x00007fff6ecd4a1c libsystem_c.dylib`abort + 120
    frame #3: 0x00007fff6ea44d78 libcorecrypto.dylib`cc_abort + 9
    frame #4: 0x00007fff6ea6dcdb libcorecrypto.dylib`cc_try_abort + 9
    frame #5: 0x00007fff6e9fe92c libcorecrypto.dylib`ccrng_cryptographic_generate + 538
    frame #6: 0x0000000100000f8e spike`main + 30
    frame #7: 0x00007fff6ec067fd libdyld.dylib`start + 1
```

reneme:
I'm on macOS 10.15.3 and Xcode 11.3.1.

Both `./botan-test system_rng` and `./botan rng 0` crash similarly. If I adapt my test program posted earlier to use `std::vector<uint8_t> buffer(0)` I see the same stack trace. Obviously all of those WITHOUT applying the patch in this pull request. With the patch all of them work fine.

Though, I guess the actual culprit is the `nullptr` parameter rather than `len` being zero, right? Shouldn't we check that instead?

randombit:
@reneme The specific condition we are checking for is nullptr/0 because that is generally considered a valid span to pass to the library. Whereas, passing say nullptr/1 should crash, or maybe shouldn't, but in any case right now that is status quo and adding null pointer checks everywhere would be a lot of checks for unclear value.

My comment about the problem being len == 0 is incorrect so I will fix that.

mouse07410:
> My comment about the problem being `len == 0` is incorrect so I will fix that.

IMHO, it would be good to check for *both* `nullptr` *and* `len == 0`, as in either case one shouldn't call `arc4random...()`.

Issue #2254, opened 2020-02-05
Consistent use of string pointer relative to UNICODE in MSVC build


pgp:
(Using Visual Studio Build Tools 2019)

The method ```HCERTSTORE open_cert_store(const char* cert_store_name)``` in ```src/lib/x509/certstor_system_windows/certstor_windows.cpp``` accepts a non-templated string pointer (const char*), but internally calls the ```CertOpenSystemStore``` API, whose actual signature depends on whether a build is UNICODE or not (that is, resolving to CertOpenSystemStoreW or CertOpenSystemStoreA, respectively). I am aware of the fact that developing a robust unicode-aware management for windows requires a lot of effort, but it would be useful at least to explicitly use **CertOpenSystemStoreA** in this case, since the signature of open_cert_store doesn't use templates, in order to avoid build failure if UNICODE is enabled.

randombit:
Thanks for the report. We need to add a Unicode build to CI to detect issues like this.

hrantzsch:
The issue can easily be reproduced by setting the `-DUNICODE -D_UNICODE` cxx flags. What's the best way to teach the CI to do this?
`ci_build.py` could either get a new option `--extra-cxxflags`, which it forwards to `configure.py`, or directly an option `--with-unicode` that sets the extra flags.

randombit:
@hrantzsch In #2256 I went with --extra-cxxflags exactly as you suggested, since we may well need similar functionality for other cases in the future. But I only used `/DUNICODE` - is `/D_UNICODE` also required here?

hrantzsch:
`/DUNICODE` is enough to reproduce the issue.

`_UNICODE` is for the headers of the C runtime (https://stackoverflow.com/a/7953476) and it seems common practice to define both. I'm not sure if it can cause any trouble but I think we want to use both.

randombit:
Fixed in master, and now we check Unicode builds in CI which should prevent similar regressions. Thanks again for the report.

PR #2280, opened 2020-02-20
Enable certstore_system_windows also for mingw configuration

pgp:
After discussing possible issues in this old [PR](https://github.com/randombit/botan/pull/1931), and performing the suggested modifications, I tried configure.py and make with os=mingw and cpu=x64 with a recent MinGW [distro](https://nuwen.net/mingw.html), which successfully builds.

### DIFF
--- a/src/build-data/os/mingw.txt
+++ b/src/build-data/os/mingw.txt
@@ -28,4 +28,5 @@ virtual_lock
 threads
 thread_local
 filesystem
+certificate_store
 </target_features>
--- a/src/lib/x509/certstor_system_windows/certstor_windows.cpp
+++ b/src/lib/x509/certstor_system_windows/certstor_windows.cpp
@@ -129,7 +129,7 @@ Cert_Vector search_cert_stores(const _CRYPTOAPI_BLOB& blob, const DWORD& find_ty
       Handle_Guard<PCCERT_CONTEXT> cert_context = nullptr;
       while(cert_context.assign(CertFindCertificateInStore(
                                    windows_cert_store.get(), PKCS_7_ASN_ENCODING | X509_ASN_ENCODING,
-                                   NULL, find_type,
+                                   WINCRYPT_UNUSED_PARAM, find_type,
                                    &blob, cert_context.get())))
          {
          auto cert = std::make_shared<X509_Certificate>(cert_context->pbCertEncoded, cert_context->cbCertEncoded);
--- a/src/lib/x509/certstor_system_windows/certstor_windows.cpp
+++ b/src/lib/x509/certstor_system_windows/certstor_windows.cpp
@@ -17,6 +17,8 @@
 #include <windows.h>
 #include <wincrypt.h>
 
+#define WINCRYPT_UNUSED_PARAM 0 // for avoiding warnings when passing NULL to unused params in win32 api that accept integer types
+
 namespace Botan {
 namespace {
 
@@ -105,7 +107,7 @@ class Handle_Guard
 
 HCERTSTORE open_cert_store(const char* cert_store_name)
    {
-   auto store = CertOpenSystemStoreA(NULL, cert_store_name);
+   auto store = CertOpenSystemStoreA(WINCRYPT_UNUSED_PARAM, cert_store_name);
    if(!store)
       {
       throw Botan::Internal_Error(
--- a/src/lib/x509/certstor_system_windows/info.txt
+++ b/src/lib/x509/certstor_system_windows/info.txt
@@ -12,4 +12,5 @@ certstor_windows.h
 
 <libs>
 windows -> crypt32
+mingw -> crypt32
 </libs>
--- a/src/scripts/ci_build.py
+++ b/src/scripts/ci_build.py
@@ -100,6 +100,10 @@ def determine_flags(target, target_os, target_cpu, target_cc, cc_bin,
         flags += ['--with-doxygen', '--with-sphinx', '--with-rst2man']
         test_cmd = None
 
+    if target == 'cross-win64':
+        # this test compiles under MinGW but fails when run under Wine
+        test_cmd += ['--skip-tests=certstor_system']
+
     if target == 'coverage':
         flags += ['--with-coverage-info', '--with-debug-info', '--test-mode']
         build_targets += ['bogo_shim']
--- a/src/tests/test_certstor_system.cpp
+++ b/src/tests/test_certstor_system.cpp
@@ -76,6 +76,12 @@ Test::Result find_cert_by_subject_dn(Botan::Certificate_Store& certstore)
    return result;
    }
 
+/*
+ * test case conditionally disabled for now, see:
+ * https://github.com/randombit/botan/pull/1931
+ * https://github.com/randombit/botan/pull/2280
+*/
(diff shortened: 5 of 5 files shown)


reneme:
Mmmh, unfortunately this [fails to build the windows certificate store](https://ci.appveyor.com/project/randombit/botan/builds/30948466/job/8rl82cblbgths9tx):

```
botan_all.cpp: In function 'void* Botan::{anonymous}::open_cert_store(const char*)':
botan_all.cpp:21384:38: error: passing NULL to non-pointer argument 1 of 'void* CertOpenSystemStoreA(HCRYPTPROV_LEGACY, LPCSTR)' [-Werror=conversion-null]
21384 |    auto store = CertOpenSystemStoreA(NULL, cert_store_name);
      |                                      ^~~~
In file included from C:/msys64/mingw64/x86_64-w64-mingw32/include/windows.h:95,
                 from botan_all.cpp:21293:
C:/msys64/mingw64/x86_64-w64-mingw32/include/wincrypt.h:4738:69: note:   declared here
 4738 |   WINIMPM HCERTSTORE WINAPI CertOpenSystemStoreA (HCRYPTPROV_LEGACY hProv, LPCSTR szSubsystemProtocol);
      |                                                   ~~~~~~~~~~~~~~~~~~^~~~~
botan_all.cpp: In function 'Botan::{anonymous}::Cert_Vector Botan::{anonymous}::search_cert_stores(const _CRYPTOAPI_BLOB&, const DWORD&, std::function<bool(const std::vector<std::shared_ptr<const Botan::X509_Certificate> >&, std::shared_ptr<const Botan::X509_Certificate>)>, bool)':
botan_all.cpp:21406:36: error: passing NULL to non-pointer argument 3 of 'const CERT_CONTEXT* CertFindCertificateInStore(HCERTSTORE, DWORD, DWORD, DWORD, const void*, PCCERT_CONTEXT)' [-Werror=conversion-null]
21406 |                                    NULL, find_type,
      |                                    ^~~~
In file included from C:/msys64/mingw64/x86_64-w64-mingw32/include/windows.h:95,
                 from botan_all.cpp:21293:
C:/msys64/mingw64/x86_64-w64-mingw32/include/wincrypt.h:4069:116: note:   declared here
 4069 |   WINIMPM PCCERT_CONTEXT WINAPI CertFindCertificateInStore (HCERTSTORE hCertStore, DWORD dwCertEncodingType, DWORD dwFindFlags, DWORD dwFindType, const void *pvFindPara, PCCERT_CONTEXT pPrevCertContext);
      |                                                                                                              ~~~~~~^~~~~~~~~~~
cc1plus.exe: all warnings being treated as errors
```

Admittedly, I never worked with mingw and can't give you much guidance on it. The CI platform AppVeyor comes with [a few options for mingw versions](https://www.appveyor.com/docs/windows-images-software/#mingw-msys-cygwin), though. Maybe its merely a configuration issue in the CI?

pgp:
For what I can see, those are actually warnings (that actually appear also in my local build) that are treated as errors (due to -Wall maybe?), something, which is a rather common scenario when using mingw and win32 apis.

pgp:
Also, [MSDN](https://docs.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certopensystemstorea) for that function clearly says:
```
hProv

This parameter is not used and should be set to NULL.
```
So, it's only a warning treated as error, since the actual type of HCRYPTPROV_LEGACY is maybe not a pointer type, but NULL (or equivalently 0) is supposed to be passed anyway.

Tymolc:
That indeed doesn't sound like a big problem. Maybe you can mitigate the error by casting `NULL` (or 0) to `HCRYPTPROV_LEGACY`? FWICS, `HCRYPTPROV` and `HCRYPTPROV_LEGACY` are defined as an unsigned long type, while `DWORD` is an unsigned integer, so it's possible that it's sufficient to use 0 there.
In c++ `NULL` can be defined as either the literal `0` or as `nullptr`, so it looks like the MinGW implementation went with the pointer type, while MSVC uses `0`. That's probably where the ambiguity comes from.

pgp:
Fine. Should I define WINCRYPT_UNUSED_PARAM in certstore_windows.h, or in some higher-level header?

Tymolc:
Hm. Only certstor_windows.cpp uses wincrypt, so for that specific `define` it should be fine to put it into the cpp.

pgp:
Still 1 error, due to unused function find_cert_by_utf8_subject_dn - only (conditional) usage is in test_certstor_system.cpp... Maybe this ifdef block:
```cpp
#if !defined(BOTAN_HAS_CERTSTOR_WINDOWS)
         results.push_back(find_cert_by_utf8_subject_dn(*system));
#endif
```
should be modified?

randombit:
@pgp I'm not sure why that test is disabled for Windows certstore (@reneme @Tymolc any ideas on this?).

Fix (at least assuming the test really should be disabled) would be to also #ifdef out the definition of the function starting on line 79 in same file.

reneme:
I could imagine this to be related to [this recent issue](https://github.com/randombit/botan/issues/2254). Looking at `git blame`: the disabled test just claims to "fix CI". It's worth looking into independently. @Tymolc, what do you think?

Anyway, for this PR I would also suggest to just `#ifdef` the implementation of the culprit function.

Tymolc:
Looking back through the comments on the pull request, we were talking about how the windows CI machine had a very sparse list of Root CAs and especially the one needed in the test was missing. It seems like we decided that the test wasn’t so valuable that fixing that on the Appveyor-side was worth it, so we #ifdefed it out. 
I‘m not sure if anything changed with the Root CA set on the build machines. For now, the quick fix seems reasonable. To properly include the test, we could either find a way to add the Root CA or find a website to test that has an included root ca as trust anchor.

randombit:
Tests are failing on Travis either because Wine doesn't support some of these certificate APIs or else Wine isn't looking in the default system certificate stores. Either way probably simplest to continue to build but to skip the test under Wine. I think this patch would do it

```
diff --git a/src/scripts/ci_build.py b/src/scripts/ci_build.py
index ee1262f36..0c9e7dd1f 100755
--- a/src/scripts/ci_build.py
+++ b/src/scripts/ci_build.py
@@ -100,6 +100,10 @@ def determine_flags(target, target_os, target_cpu, target_cc, cc_bin,
         flags += ['--with-doxygen', '--with-sphinx', '--with-rst2man']
         test_cmd = None
 
+    if target == 'cross-win64':
+        # this test compiles under MinGW but fails when run under Wine
+        test_cmd += ['--skip-tests=certstor_system']
+
     if target == 'coverage':
         flags += ['--with-coverage-info', '--with-debug-info', '--test-mode']
         build_targets += ['bogo_shim']
```

pgp:
Ok, some updates: after seeing the Coverage job failed on running botan-test.exe (via wine), I tried to reproduce it on a VM running Ubuntu 18.04 (bionic) (the Travis VM runs 16.04 - xenial), after having built the windows executable from mingw on Windows 10 (not cross compiling). In this case, all the tests run fine. Maybe it would be worth to try update the coverage job machine from xenial to bionic, since other jobs use bionic as well?

randombit:
@pgp The coverage build failure is an unrelated issue (there is one PKCS11 test that is flaky in CI and we have not figured out what the issue is), I have restarted that job and it will likely come back green.

PR #2284, opened 2020-02-27
Coalescing TLS records

nametoolong:
This is an ad-hoc patch that tries to coalesce multiple TLS handshake records into one packet, so that handshakes are made slightly faster, especially for DTLS.

I am not familiar enough with Botan's code, not could I figure out how should I coalesce multiple TLS messages into a single record when their epochs interleave, so this patch just concatenates individual records. Its approach might be troublesome (exceptions are completely not handled), and it might have introduced memory hazards.

I would like to ask for such a feature to be added into the library instead of sending a badly-written PR, but I think submitting my patch might be helpful for other users. Please review with care.

**EDIT:**
Alternatively we can implement this on a per-message basis for DTLS, as there is already a queue of messages in `Datagram_Handshake_IO`. Stream-oriented TLS can be left as-is.

### DIFF
--- a/src/lib/tls/tls_channel.cpp
+++ b/src/lib/tls/tls_channel.cpp
@@ -196,7 +196,12 @@ Handshake_State& Channel::create_handshake_state(Protocol_Version version)
 bool Channel::timeout_check()
    {
    if(m_pending_state)
-      return m_pending_state->handshake_io().timeout_check();
+      {
+      start_buffering_records();
+      bool retrans_needed = m_pending_state->handshake_io().timeout_check();
+      send_buffered_records();
+      return retrans_needed;
+      }
 
    //FIXME: scan cipher suites and remove epochs older than 2*MSL
    return false;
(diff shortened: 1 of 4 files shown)


randombit:
Thanks. Right now we rely heavily on transport layer (ie whatever the IO callback layer is doing) to coalesce writes. That works ok for TCP most of the time but is certainly not optimal for DTLS.

I'm sick and not really in any condition to review a patch touching code this complicated right now but I'll take a look soon and hopefully it can either be made to work or at least inspire another implementation.

nametoolong:
I should also note some 802.1x protocols like PEAP or EAP-TTLS use stream TLS but in a request/response way. Write coalescence is a must for EAP supplicants.

reneme:
FWIW: The TLS 1.3 implementation (currently under active development) does implement coalescing of handshake messages into single records: https://github.com/randombit/botan/blob/ffe1c30395ed2c6f8d7c89e024185f1c7513b46f/src/lib/tls/tls13/tls_channel_impl_13.h#L26-L66

PR #2290, opened 2020-03-03
Replace PKCS11 OASIS headers with a version from p11-kit

alvinchchen:
Hello,

I found headers provided by OASIS PKCS11 TC are not-exactly free license
(they do not allow modification).
I use an alternative header from p11-kit to replace them.
Which is licensed under a more liberal license.

Regards,
Alvin


mouse07410:
Is their content identical? I thought p11-kit guess were missing some definitions?

Also, why would you ever want to modify those files that define PKCS#11 standard/required values?

randombit:
@mouse07410 It's an issue with the licensing of the OASIS versions (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=952951)

Seems ok to me but I would like input from @neusdan or @securitykernel here

neusdan:
That’s fine for me if they are identical. But I also think that there should never be any need to modify the files.

mouse07410:
First, I see the Debian issue, but don't understand why inability to modify those headers is a problem.

Second, I *don't* think they're identical. There are three or four different variants of PKCS#11 headers floating around the Internet, and I don't think p11-kit has the most complete one.

Unless these files are identical (which I doubt), or the p11-kit variants is more complete - I'm against this change.

randombit:
As best I can tell the p11-kit headers are derived from the the OASIS headers but with the OASIS copyright removed. We are not violating the OASIS license right now, but by distributing p11-kit headers we would be. So I am not going to accept this change.

An alternative that would work would be to add a build flag to use some externally provided PKCS11 headers instead.

PR #2294, opened 2020-03-04
Remove support for running macOS binaries from local dir

randombit:
This is a lot of complications just to work around a user misaprehension about ld paths. And likely most macOS users are getting it via Homebrew anyway.

Taken from #1505

cc @reneme since I know you are using macOS.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1476,7 +1476,6 @@ class OsInfo(InfoObject): # pylint: disable=too-many-instance-attributes
                 'doc_dir': 'share/doc',
                 'man_dir': 'share/man',
                 'use_stack_protector': 'true',
-                'so_post_link_command': '',
                 'cli_exe_name': 'botan',
                 'lib_prefix': 'lib',
                 'library_name': 'botan{suffix}-{major}',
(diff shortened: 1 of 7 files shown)


mouse07410:
I take an exception about most users "getting it via Homebrew", but what would this PR do? I.e., I build Botan on MacOS from sources, run tests, and then install (into `/opt/local` if it matters). Works there be any difference for me if this PR is merged?

randombit:
@mouse07410 Only change is that if you want to run the binaries from the working directory without having installed the lib to a system location, you must first set `DYLD_LIBRARY_PATH` appropriately (eg to `.` or to the full path where the checkout is).

mouse07410:
Understood, thank you! In that case - no problem whatsoever.

reneme:
I don't see a problem in removing this either. I think, it's fairly common practise to set `DYLD_LIBRARY_PATH` in such cases. Personally, I usually build things with static linking anyway or use conan to put things together.

Generally, we worry about rpath only when packaging our application for releases to customers.

PR #2311, opened 2020-03-28
cpu arm: Dead code as the symbol was never defined.

devnexen:


src/lib/utils/cpuid/cpuid_arm.cpp
```diff
@@ -16,6 +16,10 @@
 #else
   #include <botan/internal/os_utils.h>
 
+#if defined(BOTAN_TARGET_OS_HAS_GETAUXVAL) || defined(BOTAN_TARGET_OS_HAS_ELF_AUX_INFO)
+  #include <sys/auxv.h>
```
randombit: Why add this include? It should only be needed if we are invoking `getauxval` directly instead of through `OS::` wrapper, right?

devnexen: Not so much about getauxval but more AT_DBCACHESIZE

randombit: Ah got it thanks

randombit:
LGTM but btw while pushing to a branch on this repo is fine, please prefix with a username or id of some kind. I have recently started prefixing all my WIP branches with `jack/` so it is clear these are not something that should be used directly.

PR #2312, opened 2020-03-29
Make CBC padding constant time

randombit:
Maximilian Blochberger points out that while unpadding was constant
time, padding operation leaked the length of the plaintext. This is
probably not too serious in most circumstances but is not desirable
behavior.

### DIFF
--- a/doc/security.rst
+++ b/doc/security.rst
@@ -15,6 +15,23 @@ mail please use::
 This key can be found in the file ``doc/pgpkey.txt`` or online at
 https://keybase.io/jacklloyd and on most PGP keyservers.
 
+2020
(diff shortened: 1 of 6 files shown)


thmo:
@randombit : In order to prepare updates for the botan2 packages in F31 and EPEL8 ([bz](https://bugzilla.redhat.com/show_bug.cgi?id=1849743)) and avoid an soname change, I'd go and cherry pick commits 7318a40 and a56547e ontop of 2.11.0 (for F31) and 2.12.1 (for EPEL8).

Any objections?

randombit:
@thmo Backporting those should be fine. The CBC padding fixes are very low severity imo (OpenSSL still even has variable time CBC *unpadding*, which is a far more serious issue which can leak plaintext information, not just the length, and apparently nobody cares)

Issue #2318, opened 2020-04-03
Stateless DTLS cookie verification


nametoolong:
It seems like that Botan requires every new DTLS client hello be handled in a new `TLS_Server`. Therefore, DTLS session cookies only provide protection against amplification, but not against resource exhaustion attacks. An attacker can easily consume a few hundred megabytes of memory by creating many TLS servers, without actually responding to hello verify requests.

Finding no clue in the documentation, I end up writing a very ugly solution like this:
```cpp
std::vector<uint8_t> msg_contents(record_buf.begin() + 12, record_buf.begin() + length + 8);
Client_Hello client_hello(msg_contents);
Hello_Verify_Request verify(client_hello.cookie_input_data(), client_identity, cookie_secret);

if (client_hello.cookie() == verify.cookie())
{
	std::unique_ptr<Server> server(new Server(/* blablabla */));
	// monkey-patch the sequence numbers in Handshake_IO here
}
else
{
	std::unique_ptr<Datagram_Handshake_IO>
		handshake_io(new Datagram_Handshake_IO(/* blablabla */));
	handshake_io->send(verify);
}
```
Is there any facility in Botan to verify a DTLS cookie statelessly, like GnuTLS's `gnutls_dtls_cookie_verify`? If not, will there be any plan to add such an interface?

randombit:
We don't have a good story on this but certainly something that should be improved upon given the increasing number of DTLS users I am aware of.

PR #2320, opened 2020-04-04
Stateless DTLS cookie verify

nametoolong:
#2318
I've ported the approach in our production code back into Botan. It works fine but is a bit hacky.

Use static method `TLS::Server::pre_verify_cookie` to verify the client hello. Pass the resultant `DTLS_Prestate` into the newly-added `TLS::Server` constructor. Call `TLS::Server::received_data` again with the verified client hello.

As far as I have tested, no old code is broken. Interoperability is untested.

Possible concerns:
- Can we avoid passing the client hello twice? This might require more architectural change.
- Several methods in `Datagram_Handshake_IO` are made static. Should it be done this way?
- We could refactor `TLS::write_record` and remove `unconnected_write_record` from `tls_server.cpp`.
- Some shared code can be reused between `pre_verify_cookie` and other places in `TLS::Channel`.


randombit:
Nice thank you. I will review this next week.

src/lib/tls/tls_channel.cpp
```diff
@@ -104,6 +104,18 @@ void Channel::reset_active_association_state()
       m_sequence_numbers->reset();
    }
 
+void Channel::set_prestate(uint16_t in_message_seq, uint16_t out_message_seq)
```
randombit, line 107: I think these should be uint64_t since theoretically the initial client hello could have been retransmitted and lost many times before we got it and sent the hello verify request.

src/lib/tls/tls_channel.cpp
```diff
@@ -168,6 +180,15 @@ Handshake_State& Channel::create_handshake_state(Protocol_Version version)
          m_sequence_numbers.reset(new Stream_Sequence_Numbers);
       }
 
+   for (uint64_t seq = 0; seq < pre_in_message_seq; seq++)
+      {
+      m_sequence_numbers->read_accept(seq);
+      }
+   for (uint16_t seq = 0; seq < pre_out_message_seq; seq++)
+      {
+      m_sequence_numbers->next_write_sequence(0);
```
randombit, line 189: Use `/*epoch=*/0` to make it clear what this is - at first I thought this was a typo and should have been `seq`

src/lib/tls/tls_channel.cpp
```diff
@@ -168,6 +180,15 @@ Handshake_State& Channel::create_handshake_state(Protocol_Version version)
          m_sequence_numbers.reset(new Stream_Sequence_Numbers);
       }
 
+   for (uint64_t seq = 0; seq < pre_in_message_seq; seq++)
+      {
+      m_sequence_numbers->read_accept(seq);
+      }
+   for (uint16_t seq = 0; seq < pre_out_message_seq; seq++)
```
randombit, line 187: uint64_t

src/lib/tls/tls_handshake_io.h
```diff
@@ -204,11 +216,11 @@ class Datagram_Handshake_IO final : public Handshake_IO
       uint64_t m_last_write = 0;
       uint64_t m_next_timeout = 0;
 
-      uint16_t m_in_message_seq = 0;
```
randombit, line 207: Why this change?

src/lib/tls/tls_server.cpp
```diff
@@ -290,6 +293,32 @@ get_server_certs(const std::string& hostname,
    return cert_chains;
    }
 
+void unconnected_write_record(
```
randombit, line 296: This should be in tls_record.cpp - I have a refactoring PR for this coming in

src/lib/tls/tls_server.h
```diff
@@ -20,13 +20,22 @@ namespace TLS {
 
 class Server_Handshake_State;
 
+struct BOTAN_UNSTABLE_API DTLS_Prestate final
+   {
+   public:
+   bool cookie_valid = false;
+   uint16_t in_message_seq = 0;
```
randombit, line 27: uint64_t

src/lib/tls/tls_server.cpp
```diff
@@ -1020,6 +1064,126 @@ void Server::session_create(Server_Handshake_State& pending_state,
...
+      {
+      throw TLS_Exception(Alert::Type::DECODE_ERROR,
+         "Bad length in client hello");
+      }
+
+   const uint16_t msg_seq = make_uint16(record_buf[4], record_buf[5]);
+
+   if (msg_seq > 2)
+      {
+      // is it possible for a Client Hello to get beyond seq=2?
```
randombit, line 1122: I think yes if retransmitted/lost, eg RFC 6347 section 4.2.2

>   Note, however, that from the perspective of the DTLS record layer,
   the retransmission is a new record.  This record will have a new
   DTLSPlaintext.sequence_number value.

So it could go all the way to 2**48-1. [I mean not really, but at a protocol layer it is possible]

Also should verify here that epoch == 0.

src/lib/tls/tls_channel.cpp
```diff
@@ -168,6 +180,15 @@ Handshake_State& Channel::create_handshake_state(Protocol_Version version)
          m_sequence_numbers.reset(new Stream_Sequence_Numbers);
       }
 
+   for (uint64_t seq = 0; seq < pre_in_message_seq; seq++)
```
randombit, line 183: Add sanity check that if `pre_in_message_seq > 0 || pre_out_message_seq > 0` then we are in DTLS state.

src/lib/tls/tls_server.cpp
```diff
@@ -1020,6 +1064,126 @@ void Server::session_create(Server_Handshake_State& pending_state,
...
+   // a hacky use of Stream_Sqeuence_Numbers
+   SymmetricKey cookie_secret;
+   size_t consumed = 0;
+
+   const Record_Header record =
+      read_record(true, readbuf, input,
+                  buf_size, consumed,
+                  record_buf, &seq, [](uint16_t) -> std::shared_ptr<Connection_Cipher_State> {return nullptr;},
+                  true);
+   // should read_record be refactored to get rid of those quirks?
```
randombit, line 1087: I agree below checks seem ugly here -- working on a refactor for record layer to help

src/lib/tls/tls_channel.h
```diff
@@ -309,6 +313,9 @@ class BOTAN_PUBLIC_API(2,0) Channel
       secure_vector<uint8_t> m_record_buf;
 
       bool m_has_been_closed;
+
+      uint16_t pre_in_message_seq = 0;
+      uint16_t pre_out_message_seq = 0;
```
randombit, line 318: Replace these with a `unique_ptr<DTLS_Prestate>`

src/lib/tls/tls_channel.cpp
```diff
@@ -190,6 +212,8 @@ Handshake_State& Channel::create_handshake_state(Protocol_Version version)
    if(auto active = active_state())
       m_pending_state->set_version(active->version());
 
+   reset_prestate();
```
randombit, line 215: If the prestate is held in a unique_ptr this could just be `m_prestate.reset()`

src/lib/tls/tls_server.cpp
```diff
@@ -1020,6 +1064,126 @@ void Server::session_create(Server_Handshake_State& pending_state,
...
+   secure_vector<uint8_t> record_buf;
+   Stream_Sequence_Numbers seq;
+   // a hacky use of Stream_Sqeuence_Numbers
+   SymmetricKey cookie_secret;
+   size_t consumed = 0;
+
+   const Record_Header record =
+      read_record(true, readbuf, input,
+                  buf_size, consumed,
+                  record_buf, &seq, [](uint16_t) -> std::shared_ptr<Connection_Cipher_State> {return nullptr;},
```
randombit, line 1085: Eek yeah I don't like this use of dummy seq, a refactor of the record reader code would help and also avoid the no-op lambda.

src/lib/tls/tls_server.h
```diff
@@ -20,13 +20,22 @@ namespace TLS {
 
 class Server_Handshake_State;
 
+struct BOTAN_UNSTABLE_API DTLS_Prestate final
+   {
+   public:
```
randombit, line 25: Instead of struct this should be a proper class with constructors [2, one for new prestate and the other for accepted prestate] and an accessor for cookie_valid, plus a `friend` decl for Channel so it can get at the internal values.

src/lib/tls/tls_server.cpp
randombit: Actually I think you can just use `nullptr` for the seq here

src/lib/tls/tls_handshake_io.h
nametoolong: Ugh, this is to fix the GCC -Wreorder warning in CI.

src/lib/tls/tls_server.cpp
nametoolong: We are reading the sequence in the message here, not the record. And yes it is possible to get beyond 2 - when the server sends more than one Verify Request here, and I wonder if `Datagram_Sequence_Numbers::read_accept` calls in `Channel::create_handshake_state` should be made O(1) to avoid DoS possibility.

nametoolong:
Thanks for reviewing and I am sorry to issue such an hackish PR (though it has been in production for a while :( ). I will try to fix those issues.

src/lib/tls/tls_handshake_io.h
randombit: OK

src/lib/tls/tls_server.cpp
randombit: Oh right sorry. But as you point out we can ping-pong repeatedly if eg cookie is corrupted several times, cookie key is rotated, etc, and so end up with a higher message sequence.

Good point on read_accept complexity - a client could send a (valid) client hello with a very high record sequence number and then we would spin.

src/lib/tls/tls_channel.cpp
nametoolong: 16-bit message sequence here, not the 48-bit record sequence number. It seems that we don't need to take care of record sequence since it only affects decryption? Correct me if I am wrong.

nametoolong:
@randombit What's the idiom in Botan to restrict a `DTLS_Prestate` to be used exactly once, like returning an `unique_ptr<DTLS_Prestate>`? If we store the client hello in prestate I think it makes sense to prevent each prestate from being used more than once.

randombit:
Can you rebase against master to pick up #2338 I think the new `write_unecrypted_record` fn replaces the code you have in tls_server.cpp for same purpose.

nametoolong:
Rebased (and kind of messed up commit timestamps). Cleaned up tls_server.cpp.

nametoolong:
> But parsing it is only required to get at the version, type, and cookie (which are relatively easy to find) plus to compute the hello verify MAC input which currently relies on reserializing the cookie.

This is no longer the case with DTLS 1.3 draft (which puts the cookie in an extension). I am not sure whether the parsing should be avoided or not. Maybe we need a way to construct a half-parsed `Client_Hello` instead of computing MAC input separately.

Issue #761, opened 2016-12-07
FreeBSD and libc++: Segmentation fault in secure_allocator, in a static object


louiz:
On FreeBSD 11, I have an easily reproducible segmentation fault when a static object allocates something using secure_allocator.

I’m using the default libc, libc++ and clang version 3.8
Using the latest dev botan version (1.11.35, 8faed8bf9a1e6c0)

With this simple source code.
```c++
#include <botan/auto_rng.h>

static Botan::AutoSeeded_RNG rng;

int main()
{}
```
Compiled like this

```bash
clang++ -g3 -I/usr/local/include/botan-1.11/ -lbotan-1.11 -L/usr/local/lib -std=c++14 test.cpp -Wall -Wextra
```
We get the follow backtrace in GDB:
```
$ gdb -q ./a.out 
(gdb) r
Starting program: /usr/home/tester/a.out 
[New LWP 100112]
[New Thread 802416000 (LWP 100112/a.out)]

Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 802416000 (LWP 100112/a.out)]
0x0000000801b29a54 in memset () from /lib/libc.so.7
(gdb) bt full
#0  0x0000000801b29a54 in memset () from /lib/libc.so.7
No symbol table info available.
#1  0x0000000800f2fbad in Botan::secure_scrub_memory (ptr=0x8006700c0, n=48) at ./src/lib/utils/mem_ops.cpp:30
        memset_ptr = (void *(* const volatile)(void *, int, size_t)) 0x801b29a10 <memset>
#2  0x0000000800ad64d1 in Botan::secure_allocator<unsigned char>::deallocate (this=0x80241c048, p=0x8006700c0 <Error reading address 0x8006700c0: Bad address>, n=48) at secmem.h:76
No locals.
#3  0x0000000800ad6451 in std::__1::__vector_base<unsigned char, Botan::secure_allocator<unsigned char> >::~__vector_base (this=0x80241c038) at memory:1496
No locals.
#4  0x0000000800ad6305 in std::__1::vector<unsigned char, Botan::secure_allocator<unsigned char> >::~vector (this=0x80241c038) at vector:457
No locals.
#5  0x0000000800e0b7f4 in Botan::HMAC_DRBG::~HMAC_DRBG (this=0x80241c000) at hmac_drbg.h:21
No locals.
[...]

PR #2337, opened 2020-04-19
make sure that Botan::mlock_allocator is constructed before any botan objects in the user code

Denys-Linke:


randombit:
Does this actually address the issue though? My understanding of static initialization order is that, across different translation units, the order of static initialization is undefined. So it would still be possible that the application static is created first and then the same ordering difficulty occurs.

Denys-Linke:
Thanks for pointing that out. I have to admit that it doesn't fix issue #761 in general. It just fixes my version of the issue which I described here:
https://github.com/randombit/botan/issues/761#issuecomment-616226307

Commit c8836ebadec8f9d0a422a43080d66cdae134b0e2 makes the mlock_allocator instance a static object and therefore it will be created before main(). That is sufficient for my needs, since my singleton object is created within main().

To address global static botan objects as well, I will have a look at the _nifty_ _counter_ _idiom,_ which I found here: https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Nifty_Counter

matejk:
One option is to implement a function to return a reference to a static instance that is allocated in the function like this:

declaration:

```
class mlock_allocator {
...
static mlock_allocator& mlock_allocator::instance();
...
```

definition

```
mlock_allocator& mlock_allocator::instance()
{
  static mlock_allocator& allocator;
  return allocator;
}
```

The instance of the allocator is created when the function is called for the first time.

Direct access to global allocator shall be replaced with calls to the `instance()` static function.

Denys-Linke:
I applied the nifty counter idiom to the mlock_allocator class. This idiom ensures the correct order of static initialization across multiple translation units. This should solve the general issue as well.

The nifty counter idiom is described in detail here: https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Nifty_Counter

src/lib/utils/locking_allocator/locking_allocator.cpp
```diff
@@ -68,8 +70,32 @@ mlock_allocator::~mlock_allocator()
...
+   }
+
+/*
+Schwarz counter / nifty counter idiom
+see: https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Nifty_Counter
+*/
+static int nifty_counter;
+static typename std::aligned_storage<sizeof (mlock_allocator), alignof (mlock_allocator)>::type
+   mlock_allocator_buf;
+mlock_allocator& mlock_allocator_instance = reinterpret_cast<mlock_allocator&> (mlock_allocator_buf);
```
randombit, line 83: Can these all live in an anon namespace?

src/lib/utils/locking_allocator/locking_allocator.h
```diff
@@ -36,10 +36,20 @@ class BOTAN_PUBLIC_API(2,0) mlock_allocator final
...
 
+      friend struct mlock_allocator_initializer;
+
       std::unique_ptr<Memory_Pool> m_pool;
       std::vector<void*> m_locked_pages;
    };
 
+extern mlock_allocator& mlock_allocator_instance;
+
+// this class needs to remain in a public header, even if locking_allocator.h becomes internal
```
randombit, line 47: Currently locking_allocator.h is public but it is only ever included in one source file (since we now abstract the allocation behind allocate_memory fn in `mem_ops.cpp`) and so without explicitly including it no application ever sees it. So is this still going to be effective? This may need to be in `secmem.h` instead.

Denys-Linke: That's right. I assumed locking_allocator.h would be included indirectly but I haven't checked it.
After moving `mlock_allocator_initializer` to `secmem.h` I had problems with undefined references when building botan. It took me some time to figure out, why `mlock_allocator_initializer` gets added to the static library but not to the shared library. I realized that the declaration needs to be tagged with  `BOTAN_DLL` or one if its aliases.
Since I'm not sure which one to choose, I selected `BOTAN_DLL.` Please feel free to replace it with the appropriate one.

src/lib/utils/locking_allocator/locking_allocator.cpp
Denys-Linke: The first two of them can, but `mlock_allocator_instance` needs to remain outside that namespace.

PR #2345, opened 2020-04-30
Add some missing building and configure documentation

lieser:
- Add some missing building and configure documentation
- Fix a small error in `configure.py`


randombit:
Thank you

PR #2351, opened 2020-05-08
Update CLI documentation

lieser:
- Update CLI documentation
- fix doc string for `base58_check_encode` and `base58_check_decode`


lieser:
Note that I was unsure if some of the older missing commands there left out deliberately from the documentation (e.g. `compress` and `decompress`). I can remove them again if you want.

randombit:
No the omissions were not intentional. Thanks very much!

PR #2361, opened 2020-05-31
Updated cpuid.cpp bit_from_string to support all flags

danieldjewell:
cpuid.cpp :: bit_from_string did not support parsing all potential flags
printed by "botan cpuid"; some were simply not present and some were not
matching the string printed in "botan cpuid". e.g. aes_ni is the flag
printed by cpuid but bit_from_string would only recognize "aesni";
likewise for ARM, "arm_sha2" would be printed by cpuid but the string
for matching was "armv8sha2". This made it *extraordiarily* difficult to
use "botan speed --clear-cpuid=..." to benchmark with H/W acceleration.

Tested compilation on aarch64 and x86_64 - unable to test on PPC;
shouldn't cause any breaking changes as no strings were removed...

### DIFF
--- a/src/lib/utils/cpuid/cpuid.cpp
+++ b/src/lib/utils/cpuid/cpuid.cpp
@@ -149,45 +149,63 @@ CPUID::bit_from_string(const std::string& tok)
       return {Botan::CPUID::CPUID_SSE2_BIT};
    if(tok == "ssse3")
       return {Botan::CPUID::CPUID_SSSE3_BIT};
-   if(tok == "aesni")
+   if(tok == "sse41")
+      return {Botan::CPUID::CPUID_SSE41_BIT};
+   if(tok == "sse42")
+      return {Botan::CPUID::CPUID_SSE42_BIT};
+   // aes_ni is the string printed on the console when running "botan cpuid"
+   if(tok == "aesni" || tok == "aes_ni")
       return {Botan::CPUID::CPUID_AESNI_BIT};
    if(tok == "clmul")
       return {Botan::CPUID::CPUID_CLMUL_BIT};
    if(tok == "avx2")
       return {Botan::CPUID::CPUID_AVX2_BIT};
-   if(tok == "sha")
+   if(tok == "avx512f")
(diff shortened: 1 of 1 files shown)


src/lib/utils/cpuid/cpuid.cpp
```diff
@@ -149,45 +149,63 @@ CPUID::bit_from_string(const std::string& tok)
...
       return {Botan::CPUID::CPUID_SHA_BIT};
+   if(tok == "rdtsc")
+      return {Botan::CPUID::CPUID_RDTSC_BIT};
+   if(tok == "bmi1")
+      return {Botan::CPUID::CPUID_BMI1_BIT};
    if(tok == "bmi2")
       return {Botan::CPUID::CPUID_BMI2_BIT};
    if(tok == "adx")
       return {Botan::CPUID::CPUID_ADX_BIT};
-   if(tok == "intel_sha")
```
danieldjewell, line 164: Just so it's clear - Since this had the same return value as ```c++ tok=="sha"``` I streamlined it into an OR on the "sha" if

randombit:
Oh thank you, this inconsistency would occasionally bug me and then I would be busy with other things and forget to look at it.

danieldjewell:
No worries! (It was bugging me too 😁) ...  I'm just happy that I could help. 

On a related note: I was thinking that it would be kinda cool/convenient to have a ```--clear-cpuid=all``` or equivalent to disable all possible CPU acceleration in one go - what do you think? 

(Possibly even some "high level" keywords like ```--clear-cpuid=sse``` to disable all flavors of SSE?) Not sure, but maybe even making something that matches *regardless of architecture* (e.g. ```clear-cpuid=aes``` that would disable aes_ni on x86 and arm_aes on ARM? Not suggesting removing anything, just adding some shortcuts that would make that portion of the code more consistent between architectures? Obviously there's a limit - I certainly don't want to suggest completely irrelevant things that just bloat the code... ) 

Let me know - I can also open a regular issue to track the discussion if you'd like :smile:

PR #2362, opened 2020-06-01
Attempt to avoid leaking the HMAC key length

randombit:
As sometimes this itself is sensitive information, as in PBKDF2 for instance.

Pointed out by Robert Merget of Ruhr-Universität Bochum

### DIFF
--- a/src/lib/mac/hmac/hmac.cpp
+++ b/src/lib/mac/hmac/hmac.cpp
@@ -1,12 +1,13 @@
 /*
 * HMAC
-* (C) 1999-2007,2014 Jack Lloyd
+* (C) 1999-2007,2014,2020 Jack Lloyd
 *     2007 Yves Jerschow
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
 
 #include <botan/hmac.h>
+#include <botan/internal/ct_utils.h>
 
 namespace Botan {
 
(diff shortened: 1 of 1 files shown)


src/lib/mac/hmac/hmac.cpp
```diff
@@ -54,27 +55,51 @@ void HMAC::key_schedule(const uint8_t key[], size_t length)
...
-
-      for(size_t i = 0; i != m_hash_output_length; ++i)
+      }
+   else if(length > 0)
+      {
+      for(size_t i = 0; i != m_hash_block_size; ++i)
          {
-         m_ikey[i] ^= ipad;
+         auto in_range = CT::Mask<size_t>::is_lt(i, length);
+         m_ikey[i] = static_cast<uint8_t>(in_range.if_set_return(key[i % length]));
```
briansmith, line 95: If `length` is secret then it would be unfortunate to use `% length` since `%` is probably somewhat leaky w.r.t. the the value of `length` since division doesn't usually have value-invariant timing.

Consider:

```
size_t i_mod_length = 0;
for(size_t n = 0; n != m_hash_block_size; ++n) {
   i_mod_length = constant_time_select(constant_time_lt(i_mod_length, length), i_mod_length, 0);
   m_ikey[i] = key[i_mod_length];
```

randombit: Wow I totally forgot here that division/mod isn't usually constant time, good catch thank you

src/lib/mac/hmac/hmac.cpp
```diff
@@ -54,27 +55,51 @@ void HMAC::key_schedule(const uint8_t key[], size_t length)
...
+   * branch to hash first reveals that. In addition, counting the number of
+   * compression functions executed reveals the size at the granularity of the
+   * hash function's block size.
+   *
+   * The greater concern is for smaller keys; being able to detect when a
+   * passphrase is say 4 bytes may assist choosing weaker targets. Even though
+   * the loop bounds are constant, we can only actually read key[0..length] so
+   * it doesn't seem possible to make this truly constant time. This should
+   * still hide the length of the key to a leak at the granularity of the cache
+   * line size, which is sufficient for most purposes and similar to the leak
```
briansmith, line 76: NIT: I think this assertion about the leak being at the granularity of the cache line is unfounded. Shay Gueron already made a pretty convincing argument that it is not true on x86_64, which he used to justify replacing "small table" AES implementations with ones that with data-invariant memory access patterns. I forgot the details but IIRC at the very least the timing was different when `(uintptr_t)p % CACHE_LINE_WIDTH == 0`.

randombit: Interesting. I am not sure I've seen that one. There are cache bank conflicts [ala https://eprint.iacr.org/2016/224] but my understanding is those only affect x86 and do not affect the most recent core designs.

My argument implicitly assumed that `key` didn't cross a cache line which is not something we can control, so even without cache bank conflicts it was not valid.

src/lib/block/sm4/sm4.cpp
```diff
@@ -117,15 +117,15 @@ inline uint32_t SM4_Tp(uint32_t b)
...
+   B##2 ^= F(B##3 ^ B##0 ^ B##1 ^ m_RK[4*R+2]);            \
    B##3 ^= F(B##0 ^ B##1 ^ B##2 ^ m_RK[4*R+3]);            \
    } while(0)
 
 #define SM4_D_RNDS(B, R, F) do {                           \
    B##0 ^= F(B##1 ^ B##2 ^ B##3 ^ m_RK[4*R+3]);            \
-   B##1 ^= F(B##0 ^ B##2 ^ B##3 ^ m_RK[4*R+2]);            \
-   B##2 ^= F(B##0 ^ B##1 ^ B##3 ^ m_RK[4*R+1]);            \
+   B##1 ^= F(B##2 ^ B##3 ^ B##0 ^ m_RK[4*R+2]);            \
+   B##2 ^= F(B##3 ^ B##0 ^ B##1 ^ m_RK[4*R+1]);            \
```
briansmith, line 128: It seems some unrelated changes got in?

src/lib/x509/x509cert.cpp
```diff
@@ -240,6 +240,13 @@ std::unique_ptr<X509_Certificate_Data> parse_x509_cert_body(const X509_Object& o
       {
       if(ext->get_is_ca() == true)
          {
+         /*
+         * RFC 5280 section 4.2.1.3 requires that CAs include KeyUsage in all
+         * intermediate CA certificates they issue. Currently we accept it being
+         * missing, as do most other implementations. But it may be worth
+         * removing this entirely, or alternately adding a warning level
+         * validation failure for it.
+         */
```
briansmith, line 249: Ditto. Unrelated changes.

PR #2363, opened 2020-06-05
Fix cppcheck findings

securitykernel:
* Variable vs. function shadowing with `CLI::group()`
* Wrong XMSS parameter string to enum mapping
* Duplicate `if` statements in TLS
* Unused variables in PKCS#10 and pubkey tests

### DIFF
--- a/src/cli/pubkey.cpp
+++ b/src/cli/pubkey.cpp
@@ -384,19 +384,19 @@ class EC_Group_Info final : public Command
 
       void go() override
          {
-         Botan::EC_Group group(get_arg("name"));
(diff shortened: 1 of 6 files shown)


src/cli/pubkey.cpp
```diff
@@ -425,16 +425,16 @@ class DL_Group_Info final : public Command
 
       void go() override
          {
-         Botan::DL_Group group(get_arg("name"));
+         Botan::DL_Group ec_group(get_arg("name"));
```
randombit, line 428: Should be `dl_group`

securitykernel: Sure, fixed in 3738dd8

PR #2371, opened 2020-06-16
Fix memory leak

hekkihek:
CommonCrypto_BlockCipher::clear() did not release the cryptor objects.
When assigning a new key the previously created cryptor objects were not released.


hekkihek:
The Travis CI error looks like a (temporary) issue on their side.

hekkihek:
Done.

Issue #2375, opened 2020-06-22
Botan is wrongly configured when using --with-build-dir


PetrJarosik:
When the configure.py is running with **--with-build-dir** switch, so CmakeLists.txt is wrongly generated and subsequent build fails then.

PetrJarosik:
The main issue is that **COMPILER_INCLUDE_DIRS** variable is hardcoded in **src/build-data/cmake.in** file as shown below:
`set(COMPILER_INCLUDE_DIRS build/include build/include/external)`

It means that manually defined build directory is not taken into account in this case.

PR #2376, opened 2020-06-22
#2375 Fix hardcoded COMPILER_INCLUDE_DIRS

PetrJarosik:
The 'COMPILER_INCLUDE_DIRS' variable contains hardcoded value which is wrong when specified custom build directory by --with-build-dir switch. This change fixes the functionality and set the value according to specified build directory.


randombit:
Thanks. This is fine. But you should be aware that CMake only exists because some people who contribute to the library prefer CMake due to IDE integration, and it is not supported for production builds. If you want to send patches that is fine but especially as I myself do not use it at all I would not have the ability or interest to help that much with CMake specific issues.

PR #2380, opened 2020-06-26
Remove some duplicated logic for Certstore find_cert

randombit:
Many of the implementations had the same logic of `find_all_certs` followed by selecting one value arbitrarily, usually first in the list, though in the case of SQL, the last instead.

Make that behavior the default implementation.

Leave `find_cert` as virtual as some implementations can (Windows) or could (MacOS, SQL) limit the number of retreived certs in order to reduce overhead.

### DIFF
--- a/src/lib/x509/certstor_sql/certstor_sql.cpp
+++ b/src/lib/x509/certstor_sql/certstor_sql.cpp
@@ -52,27 +52,24 @@ Certificate_Store_In_SQL::find_cert(const X509_DN& subject_dn, const std::vector
 
    if(key_id.empty())
       {
-      stmt = m_database->new_statement("SELECT certificate FROM " + m_prefix + "certificates WHERE subject_dn == ?1");
+      stmt = m_database->new_statement("SELECT certificate FROM " + m_prefix + "certificates WHERE subject_dn == ?1 LIMIT 1");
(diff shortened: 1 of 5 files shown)


src/lib/x509/certstor_sql/certstor_sql.cpp
```diff
@@ -43,38 +43,6 @@ Certificate_Store_In_SQL::Certificate_Store_In_SQL(std::shared_ptr<SQL_Database>
...
-      }
-
-   std::shared_ptr<const X509_Certificate> cert;
-   while(stmt->step())
-      {
-      auto blob = stmt->get_blob(0);
-      cert = std::make_shared<X509_Certificate>(
-            std::vector<uint8_t>(blob.first,blob.first + blob.second));
-
-      }
```
reneme, line 73: Adapting that implementation for less overhead would have been easy, no? Just add ` LIMIT 1` to the SQL statements... Could be reintroduced in another PR of course.

randombit: Yeah I was thinking about doing it in a later PR but no reason not to do it here.

randombit:
Rebased, also changed SQL to use LIMIT 1

PR #2381, opened 2020-06-29
Overhaul Certificate_Store_MacOS

reneme:
This overhauls the implementation of `Certificate_Store_MacOS` to allow distinction between `kSecMatchLimitAll` and `kSecMatchLimitOne`. It is related to [my previous PR](https://github.com/randombit/botan/pull/2379) which merely quick-fixed an issue where `Certificate_Store_MacOS::find_cert()` would fail for ambiguous root certificate search results.

Key points implemented:

* Add an internal class `Certificate_Store_MacOS_Impl::Query` that wraps the creation (and parameter data lifetime management) of the search query to be passed into [`SecItemCopyMatching`](https://developer.apple.com/documentation/security/1398306-secitemcopymatching?language=objc)
* Refactor some internal helper functions for better readability and code reuse
* Add the internal methods `Certificate_Store_MacOS_Impl::findOne()` and `::findAll()` that take a `::Query` object, perform a certificate store search with it and parse the result into a `Botan::X509_Certificate` or a `std::vector<>` of those
* As a result: Remove most Apple-specific implementation details from the interface implementation of `Botan::Certificate_Store`

### DIFF
--- a/src/lib/x509/certstor_system_macos/certstor_macos.cpp
+++ b/src/lib/x509/certstor_system_macos/certstor_macos.cpp
@@ -209,7 +209,51 @@ std::shared_ptr<const X509_Certificate> readCertificate(SecCertificateRef cert)
    return std::make_shared<Botan::X509_Certificate>(ds);
    }
 
-}
+/**
+ * Reads all certificates from a given CFArrayRef into a result vector
+ */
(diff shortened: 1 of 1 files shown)


reneme:
This is likely going to conflict with https://github.com/randombit/botan/pull/2380 as it removes `Certificate_Store_MacOS::find_cert()` entirely. The merge should be fairly straight forward, though.

Tymolc:
LGTM 👍

randombit:
Nice thanks

PR #2385, opened 2020-07-08
solaris/illumos build update and cli sandbox support proposal.

devnexen:


src/cli/sandbox.cpp
```diff
@@ -63,6 +67,56 @@ bool Sandbox::init()
       }
 
    return (::cap_enter() == 0);
+#elif defined(BOTAN_TARGET_OS_HAS_SETPPRIV)
+   priv_set_t *ps;
+
+   if ((ps = ::priv_allocset()) == nullptr)
```
randombit, line 73: Can we use `unique_ptr<priv_set_t, ::priv_freeset>` to avoid the code duplication of freeing on error?

src/cli/sandbox.cpp
```diff
@@ -63,6 +67,56 @@ bool Sandbox::init()
...
+      {
+      return false;
+      }
+
+   ::priv_basicset(ps);
+
+   if (::priv_delset(ps, PRIV_PROC_FORK) == -1)
```
randombit, line 80: These and following could be replaced by a loop, something like:

const auto privs[] = { PRIV_PROC_FORK, PRIV_PROC_EXEC, ... }
for(auto priv : privs) {
  priv_delset(ps, priv);
}

src/cli/sandbox.cpp
```diff
@@ -63,6 +67,56 @@ bool Sandbox::init()
...
+      return false;
+      }
+
+   if (::setppriv(PRIV_SET, PRIV_INHERITABLE, ps) == -1)
+      {
+      ::priv_freeset(ps);
+      return false;
+      }
+
+   ::priv_emptyset(ps);
```
randombit, line 116: Elsewhere we `freeset` without using `emptyset` - is that ok?

devnexen: might be better indeed.

randombit (approved):

Issue #2397, opened 2020-09-02
RNG segfault for multi threaded application.


larssilven:
I have developed a pkcs#11 proxy implementation that transfers p11 calls done by an application to a p11 implementation on another host (the server).
When I do a stress test using the [SoftHSM2](https://github.com/opendnssec/SoftHSMv2) p11 implementation on the server I got a segfault. See [gdb.txt](https://github.com/randombit/botan/files/5162224/gdb.txt).
I used the [p11speed](https://github.com/opendnssec/p11speed) tool on the client side. Testing with 100 threads with each thread trying to do 100 signings has always failed. I will estimate that something in between 10 to 20 signings is usually done before the segfault.
By applying [patch-softhsm2.txt](https://github.com/randombit/botan/files/5162379/patch-softhsm2.txt) to softhsm version 2.6.1 everything works fine. I have done 100 threads with 10000 signings for each thread.
So it seems to be something wrong with Botan::Stateful_RNG when used with multi threaded applications.
Using openssl instead of botan when building softhsm is also working fine.
But the strange thing is that p11speed is working if I test directly (not using the proxy) with softhsm-botan.

Cheers,
Lars

randombit:
Is SoftHSM using a single shared `AutoSeeded_RNG` among multiple threads? That is not good. These objects do not serialize access.

Can you try

```
-	rng = new Botan::AutoSeeded_RNG();
+	rng = new Botan::Serialized_RNG(new Botan::AutoSeeded_RNG());
```

and see if that resolves the crash?

[Possibly not always serializing is a mistake, I considered adding locks here sometime in the pre-2.0 timeframe but rejected it due to overhead. Then again, at least on Linux uncontended locks are very fast, so maybe the overhead doesn't matter]

larssilven:
Yes with the serialization the auto RNG is working.

Comparing the RNGs I get this result.
System RNG:
100 threads, 1000 signatures per thread, 877.77 sig/s (RSA_PKCS 2048 bits)
Serialized auto RNG:
100 threads, 1000 signatures per thread, 847.97 sig/s (RSA_PKCS 2048 bits)

I also got an idea why the proxy is not working while running without the proxy is working:
Each thread in the p11speed test is using its own session handle when accessing the p11. But in the proxy server there is one TCP socket thread for each p11 call that is currently handled by the softhsm p11 and each of these threads are reused when new calls are received.
[...]

PR #2399, opened 2020-09-12
Refactor Stateful_RNG and add a lock

randombit:
This differs from our general practice of requiring external locks if
multiple threads wish to use the same object, but reduces risk wrt
incorrect usage causing catastrophic failure such as duplicating RNG
outputs.

See also #2397


reneme (approved):
LGTM :+1:

PR #2401, opened 2020-09-17
Updated HTTP Utils so that OCSP responder URLs that contain a non sta…

j3ll3:
Updated HTTP Utils so that an OCSP responder with a non-standard port (80) will work.

Problem:
A certificate containing an URI with a (non-standard) port number (e.g. http://ocsp-responder.local:8080) results in the following error:
Error: HTTP error HTTP connection to ocsp-responder.local:8080 failed: Name resolution failed for ocsp-responder.local:8080 error code -2

Root cause:
The root cause is that the hostname + portnumber is passed to the getaddrinfo in socket_udp.cpp which is unable to resolve the combination.

Solution:
To solve this problem the hostname and portnumber must be seperated before passing to the getaddrinfo
In my opinion this would be in the http_util.cpp function http_transact where the socket is opened. 
If no port can be found in the hostname the service "http" is used as already implemented.

### DIFF
--- a/src/lib/utils/http_util/http_util.cpp
+++ b/src/lib/utils/http_util/http_util.cpp
@@ -31,16 +31,29 @@ std::string http_transact(const std::string& hostname,
    std::unique_ptr<OS::Socket> socket;
 
    const std::chrono::system_clock::time_point start_time = std::chrono::system_clock::now();
+   std::string service, host;
+
+   const auto port_sep = hostname.find(":");
+   if(port_sep == std::string::npos)
+      {
+      service = "http";
+      host = hostname;
+      }
+   else
+      {
+      service = hostname.substr(port_sep + 1, std::string::npos);
+      host = hostname.substr(0, port_sep);
+      }
 
    try
       {
-      socket = OS::open_socket(hostname, "http", timeout);
+      socket = OS::open_socket(host, service, timeout);
       if(!socket)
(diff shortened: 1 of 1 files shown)


randombit:
Thanks, I'll want to test this but change looks ok. Do you know if there are any public CAs which use non-standard ports for OCSP?

j3ll3:
If I understand correct some public certificates solve this problem by requiring a host header in the OCSP HTTP request but this is contractionary as the OCSP specification prescribes HTTP 1.0.

Openssl has a good discussion on this topic: https://github.com/openssl/openssl/issues/1986

In my opinion both solutions (non-standard port numbers and the host header) are the best solutions to allow interoperability with both HTTP 1.0 and 1.1. Main goal is to connect to an address that hosts multiple OCSP responders.

I checked with openssl and wolfssl where a port number is allowed in the OCSP URI.

randombit (approved):

j3ll3:
I found a potential issue with this pull request in http_util.cpp. 
The http_transact function will now work but the issue with the port number remains in http_sync on line 136.

Is it an idea to extend the interface of the http_transact function so that the port/service is passed as an argument from the http_sync function?

randombit:
Ack I missed your comment before merging. But yes your proposal sounds ok. Can you open a new PR changing this?

PR #2406, opened 2020-09-26
Moved seperation of port number from hostname to http_sync function

j3ll3:
Improvement of pull request https://github.com/randombit/botan/pull/2401
The non-standard port number is now also removed from the HTTP host header

### DIFF
--- a/src/lib/utils/http_util/http_util.cpp
+++ b/src/lib/utils/http_util/http_util.cpp
@@ -25,35 +25,23 @@ namespace {
 * closes the socket.
 */
 std::string http_transact(const std::string& hostname,
+		                  const std::string& service,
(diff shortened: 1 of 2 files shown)


src/lib/utils/http_util/http_util.cpp
```diff
@@ -143,6 +131,18 @@ Response http_sync(http_exch_fn http_transact,
       loc = url.substr(host_loc_sep, std::string::npos);
       }
 
+   const auto port_sep = hostname.find(":");
+   if(port_sep == std::string::npos)
+      {
+      service = "http";
+      hostname = hostname;
```
randombit, line 138: This triggers a warning under clang, which is causing CI to fail

PR #2413, opened 2020-10-03
Converting raw pointers to smart pointers

TheOverLordDude:
I was having issues with heap corruption do to issues related to pipes and their use of raw pointers with filters.
To fix this I am proposing the use of smart pointers to manage the pointers as opposed to manual allocation/deallocation by the pipe.
This may break old software, however it is a matter of time until the Pipe system fails by using raw pointers and improper memory management.
I only edited the related files. I have not gone through the whole library yet to see where else pipes are used to correct the issue.
Using share_ptr for filters is significantly safer and much cleaner and to the modern C++ standard anyway.

### DIFF
--- a/src/lib/filters/filter.cpp
+++ b/src/lib/filters/filter.cpp
@@ -7,6 +7,7 @@
 
 #include <botan/filter.h>
 #include <botan/exceptn.h>
+#include <memory>
 
 namespace Botan {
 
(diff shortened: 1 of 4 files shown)


TheOverLordDude:
I may have also forgot a few headers as I did this directly in github as opposed to downloading the source and editing it.
In my defense I need to sleep in about ten minutes and did this in the heat of passion.

TheOverLordDude:
Don't merge this I have a lot more work to do. My bad. As I said, heat of passion.

PR #2419, opened 2020-10-09
improve check if signature algorithm supports DER SEQUENCE encoding

lieser:
Recently we added in some test code RSA signing which used the DER SEQUENCE format. This of course failed, as it is currently not supported.

But because the error message was only `Encoding error: Unexpected size for DER signature`, it was not immediately obvious that this was not an error in our code, but that it is simply not supported. This PR should improve the given error message.

Note that this would break code there currently the following is done:
```c++
// key: a Key of an algorithm there DER SEQUENCE format is not supported, e.g. RSA
PK_Verifier verifier(key, "EMSA4(SHA-512)", Signature_Format::DER_SEQUENCE);
verifier.set_input_format(Signature_Format::IEEE_1363);
// code for verification ...
```

Also added a test case for using ECDSA signatures with the DER SEQUENCE format.

/cc @securitykernel

### DIFF
--- a/src/lib/pubkey/pubkey.cpp
+++ b/src/lib/pubkey/pubkey.cpp
@@ -223,6 +223,12 @@ SymmetricKey PK_Key_Agreement::derive_key(size_t key_len,
    return m_op->agree(key_len, in, in_len, salt, salt_len);
    }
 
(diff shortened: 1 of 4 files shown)


randombit:
Thanks!

PR #2420, opened 2020-10-11
Fix order of linker flags

antonio-rojas:
Linked libraries should go after other linker flags, otherwise some flags (such as --as-needed) are not applied correctly


randombit:
Thanks!

PR #2422, opened 2020-10-15
src/build-data/arch/arm64.txt: add aarch64_be

ffontaine:
Allow the user to set cpu value to aarch64_be.
Endianness will be correctly guessed by choose_endian function because
this alias ends with "be"

Fixes:
 - http://autobuild.buildroot.org/results/69ebf03c59b2af4140a39bc26f17d0396b6ec15d

### DIFF
--- a/src/build-data/arch/arm64.txt
+++ b/src/build-data/arch/arm64.txt
@@ -5,6 +5,7 @@ family arm
 
 <aliases>
 aarch64
+aarch64_be
 armv8
 armv8-a
 </aliases>


randombit:
Thanks, this is fine, but can you give some background on big-endian Aarch64? I had thought while it was still bi-endian, literally nobody ran it in big-endian mode, so that there is any call for this is surprising to me.

ffontaine:
This build failure was raised by one of the buildroot autobuilders with the linaro aarch64_be toolchain which can be found here: https://releases.linaro.org/components/toolchain/binaries/7.3-2018.05/aarch64_be-linux-gnu/

PR #2426, opened 2020-10-21
fix parsing nested cipher name

winterland1989:
During testing a Haskell binding, I find a bug when cipher name nesting becomes deep, for example. 

`Lion(SHA-3(384),OFB(Cascade(Lion(MD4,CTR-BE(AES-128,16),128),AES-128)),128)/CBC/PKCS7`

botan will incorrectly translate the block cipher into 

`Lion(SHA-3(384),OFB(Cascade(Lion(MD4,CTR-BE(AES-128,16),128),AES-128),128))`

Note the `)` is written at a wrong positon. 

In reality, nobody should construct such a complex cipher, but still a bug to be fixed.

### DIFF
--- a/src/lib/base/scan_name.cpp
+++ b/src/lib/base/scan_name.cpp
@@ -32,8 +32,11 @@ std::string make_arg(const std::vector<std::pair<size_t, std::string>>& name, si
          }
       else if(name[i].first < level)
          {
-         output += ")," + name[i].second;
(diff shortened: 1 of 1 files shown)


randombit:
Thanks! Also, great to see a Haskell binding happening. There is a list of bindings in https://github.com/randombit/botan/wiki/Language-Bindings if you care to add a link.

PR #2434, opened 2020-10-28
Remove various deprecated algorithms

randombit:
[Not for merge until after 2.17.0 is released and master becomes 3.0.0-alpha1]


mouse07410:
Can I ask why GCM 64-bit tags are deprecated? Austere comm links may need them. NIST standard allows as short as 32-bit auth tag.

Same question about CBC-MAC. It's still used by some apps. E.g., think AES-CCM that, unlike GCM, did not fail catastrophically if nonce value is (accidentally) reused.

XTEA - while not a "production" cipher, it's a nice toy, useful for teaching purposes. I'd rather have it stay (and I don't believe it is costly to maintain ;).

randombit:
Like other polynomial macs, GCM falls over very hard once you can forge a single tag. Which is why NIST insists on very strict message length and message count limitations when short tags are in use (Appendix C of SP 800-38d), restrictions which we do not enforce. On the other hand it appears both OpenSSL and BearSSL even allow a 0-byte tag, wups :hankey: For links where long authentication tags are too expensive it would be better to use CCM, EAX, or a CTR+HMAC construction, avoiding polynomial macs entirely. However given the apparent laxity everyone else treats this maybe continuing to allow 64 bits is ok.

Is CBC-MAC still used anywhere? Can you provide a link? X9.19 CBC-MAC is retained since I know that is still in use in old financial systems. But plain CBC-MAC is so error prone I am hesitant to keep it around unless there is a very clear use case.

XTEA is nice as a toy but it is not used anywhere in any kind of production system I have been able to find. Literally the only reason it was not removed in the runup to 2.0 was because it was included in Golang's x/crypto library - but shortly after 2.0 was released, Go deprecated their version as well.

And FWIW there *are* costs to every algorithm in ongoing maintenance. This especially comes up when I need to make a change which affects every {BlockCipher,HashFunction,etc} instance.

mouse07410:
It is my understanding that CBC-MAC is a part of CCM (which is CBC-MAC plus CTR encryption). So, I'm confused - how can you remove CBC-MAC without removing CCM? And I'm strongly urging to keep CCM, as we are using it. Sorry, no URL - the code is not public :-(.
Also, CBC-MAC is used as an option in NIST KDF. It makes sense to keep supporting/providing it. 

Re. XTEA - I hear you. :-)

Re. GCM - IMHO, the problem is but with the size of the auth tag. So, I'm asking to continue allowing 64-bit tags.

Thanks

randombit:
CCM is a CTR+CBC-MAC construction but in Botan it is implemented standalone without reference to the CBC-MAC code, because the BSI profile wants CCM but prohibits CBC-MAC. CCM is not going anywhere.

randombit:
Can you provide a reference re CBC-MAC use in NIST KDFs? I can understand how CMAC would be allowed there, but CBC-MAC is AFAIK not NIST approved anymore, so I'm surprised they would allow it to be used in a SP 800-108 KDF.

mouse07410:
> Can you provide a reference re CBC-MAC use in NIST KDFs? 

NIST SP 800-108 KDF (page 8) refers to NIST SP 800-38B, which in turn specifies CMAC (a variant of CBC-MAC proposed by Black & Rogaway, later improved by Iwata & Kurosawa under the name OMAC1). I'd call it CBC-MAC with a twist.

NIST SP 800-38C explicitly approves CBC-MAC within the context of CCM, and **disapproves** it outside of CCM.

Honestly, as long as CMAC and CCM are provided - my apps won't need the "raw" unmodified CBC-MAC. Our KDF is SP 800-108 compliant (uses CMAC, I just checked).

Sorry for the noise on CBC-MAC.

halvors:
Why is Tiger hash function being completly removed? I get that it should be disabled by default, but why not let use it if explicitly defining that we want to do so?

I use this for interfacing with older applications, they are not security critical at all.

randombit:
@halvors The point of removal was to remove algorithms which are both insecure/obsolescent *and* apparently not in use. Which is why MD4/MD5 are still included, for example. Can you point to some existing, in current use, code or protocol which uses Tiger? It doesn't have to be directly Botan related; the point of the library is to be able to implement any existing protocol so if anyone is using it, anywhere, I'd be fine with reverting. There is a similar issue in #2764 but the reporter never got back to me.

PR #2445, opened 2020-11-01
Allows FreeBSD and OpenBSD to do not dump possibly sensitive data

devnexen:
as Linux.

### DIFF
--- a/src/lib/utils/os_utils.cpp
+++ b/src/lib/utils/os_utils.cpp
@@ -60,6 +60,14 @@
   #include <mach/vm_statistics.h>
 #endif
 
+#if defined(BOTAN_TARGET_OS_IS_FREEBSD)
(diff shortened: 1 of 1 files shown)


randombit:
We already have logic to use `MAP_CONCEAL` (OpenBSD) or `MAP_NOCORE` (FreeBSD), since for those OSes we can mark as nocore without the extra `madvise` call. Is there some reason to use both the `MAP_` and `MADV_` variants?

devnexen:
> We already have logic to use `MAP_CONCEAL` (OpenBSD) or `MAP_NOCORE` (FreeBSD), since for those OSes we can mark as nocore without the extra `madvise` call. Is there some reason to use both the `MAP_` and `MADV_` variants?

Oh good point sorry.

PR #2457, opened 2020-11-06
Update of DragonFlyBSD platform.

devnexen:
explicit_bzero supports since 5.5 branch.
getrandom since 5.7's

actually it is 5.8.x.


randombit:
Thanks

PR #2464, opened 2020-11-07
Hash the entire client hello (except the cookie) for hello verify

randombit:
This makes it simpler to statelessly verify a DTLS cookie without
having to fully parse the initial client hello. (GH #2320)

### DIFF
--- a/src/lib/tls/msg_client_hello.cpp
+++ b/src/lib/tls/msg_client_hello.cpp
@@ -243,21 +243,10 @@ std::vector<uint8_t> Client_Hello::serialize() const
 
 std::vector<uint8_t> Client_Hello::cookie_input_data() const
    {
-   std::vector<uint8_t> buf;
+   if(m_cookie_input_bits.empty())
+      throw Invalid_State("Client_Hello::cookie_input_data called but was not computed");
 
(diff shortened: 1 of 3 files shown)


src/lib/tls/msg_client_hello.cpp
```diff
@@ -292,14 +281,26 @@ Client_Hello::Client_Hello(const std::vector<uint8_t>& buf)
    const uint8_t minor_version = reader.get_byte();
 
    m_version = Protocol_Version(major_version, minor_version);
-
    m_random = reader.get_fixed<uint8_t>(32);
-
    m_session_id = reader.get_range<uint8_t>(1, 0, 32);
 
+   std::unique_ptr<HashFunction> sha256;
+   if(m_version.is_datagram_protocol())
```
MartinOsterta, line 288: why not merge this if with the if in line 294 and the if in line 297?

MartinOsterta (changes requested):

src/lib/tls/msg_client_hello.cpp
randombit: Thanks that approach is much clearer I do not remember why I wrote it the way I did.

PR #2510, opened 2020-11-24
Fix C++17 TODOs

hrantzsch:
Just did two minor TODOs that we marked to be done with C++17.

### DIFF
--- a/src/lib/tls/asio/asio_stream.h
+++ b/src/lib/tls/asio/asio_stream.h
@@ -677,6 +677,8 @@ class Stream
        */
       void setup_native_handle(Connection_Side side, boost::system::error_code& ec)
          {
+         BOTAN_UNUSED(side);  // workaround: GCC 9 produces a warning claiming side is unused
+
          // Do not attempt to instantiate the native_handle when a custom (mocked) channel type template parameter has
          // been specified. This allows mocking the native_handle in test code.
          if constexpr(std::is_same<ChannelT, Channel>::value)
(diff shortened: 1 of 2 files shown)


src/lib/tls/asio/asio_stream.h
```diff
@@ -683,34 +675,37 @@ class Stream
        * @param side The desired connection side (client or server)
        * @param ec Set to indicate what error occurred, if any.
        */
-      template<class T = ChannelT>
-      typename std::enable_if<std::is_same<Channel, T>::value>::type
-      setup_native_handle(Connection_Side side, boost::system::error_code& ec)
+      void setup_native_handle(Connection_Side side, boost::system::error_code& ec)
```
randombit, line 678: This is triggering a strange error under GCC 9

```
src/tests/unit_asio_stream.cpp:191:46:   required from here
build/include/botan/asio_stream.h:678:48: error: parameter ‘side’ set but not used [-Werror=unused-but-set-parameter]
  678 |       void setup_native_handle(Connection_Side side, boost::system::error_code& ec)
```

AFAICT this is a compiler bug, maybe it is being fooled by the combination of `if constexpr` and the lambda. I think resolution is just to add `BOTAN_UNUSED(side)` at the beginning of the function?

randombit: (Only an error because we compile in CI with -Werror but I'd like to keep CI warning free)

hrantzsch: That seems to fix it. Thanks for the hint, would have been quite confusing to debug this issue otherwise!

randombit:
Looks great and thank you

PR #2511, opened 2020-11-24
Remove support for AES-128 OCB ciphersuites

randombit:


deknos:
why is this done? OCB provides AEAD and is not patent encumbered anymore, why did you delete it?

randombit:
This is only OCB with AES-128 - the OCB ciphersuites with AES-256 are retained. There didn't seem to be any reason to retain both given that even now nobody but us (AFAIK) supports the OCB TLS ciphersuites. Outside of TLS you can use OCB with any cipher you like.

lieser:
@randombit To me it looks like also some OCB ciphersuites with AES-256 were removed:
- `DHE_RSA_WITH_AES_256_OCB_SHA256`
- `DHE_PSK_WITH_AES_256_OCB_SHA256` (not really relevant, as `DHE_PSK` was removed completely in a later commit)

Was the removal of `DHE_RSA_WITH_AES_256_OCB_SHA256` intended?

randombit:
@lieser I think this was accidental!

PR #2512, opened 2020-11-24
Remove support for DHE_PSK

randombit:
Retains PSK and ECDHE_PSK


pist-eb (approved):
Looks good.
One question: what is the reason of removing the DHE_PSK?

randombit:
I would like to remove DHE from TLS all together as it is slower than ECDH by several factors (when compared on equivalent security level) and also TLS DH is poorly designed, in a way that FFDHE extension unfortunately does not fix. First the p,g parameters are always sent over the wire even when a known FFDHE group is used, wasting hundreds of bytes per handshake. Secondly in TLS the server is allowed to send any kind of arbitrarily bad DH group (too small, too large, poorly generated, small subgroups, etc), unlike ECDH where only known groups can be negotiated. DHE negotiations actually have much more overhead than ECDHE due to having to validate (however imperfectly) the server provided group.

The main reason to continue supporting DHE at all is to support old or misconfigured servers which do not support ECDHE. However this does not apply to DHE_PSK since in scenarios where TLS-PSK makes sense, like VPNs or embedded systems, typically the same party controls both sides and can configure them appropriately. In addition for embedded in particular, it would be better to use ECDHE due to improved computational performance and handshake bandwidth, unless the device is too limited in which case plain PSK make sense.

pist-eb:
Thank you for answer @randombit. Sounds reasonable / I asked because I thought that maybe there is any recent security issue with DHE_PSK found.

PR #2517, opened 2020-11-27
Allow custom RSA Private_Key in Client_Key_Exchange message

claudiushauptmann:
Custom RSA Private_Key implementations are currently not accepted by Client_Key_Exchange messages. To be able to integrate cryptography hardware that does not implement PKCS11 we need custom implementations.


randombit:
Looks good thank you

PR #2524, opened 2020-11-29
Add support for keyed Blake2b and also MAC via Blake2b.

AtomicNibble:
Hello,

I've added support for keyed Blake2b https://github.com/randombit/botan/issues/1397 just opening this PR for review / feedback on approach.

I quite like this way as this is now supported:

```cpp
auto mac = Botan::MessageAuthenticationCode::create("Blake2b(256)");  
mac->set_key(macKey);  
mac->update(data);  
auto digest= mac->final();  
```

But you can also set a key on Blake2b instance itself.

Tom.


src/lib/hash/blake2/blake2b.cpp
```diff
@@ -32,26 +32,58 @@ const uint64_t blake2b_IV[BLAKE2B_IVU64COUNT] = {
 }
 
 BLAKE2b::BLAKE2b(size_t output_bits) :
+   BLAKE2b(nullptr, 0, output_bits)
+   {
+   }
+
+BLAKE2b::BLAKE2b(const SymmetricKey& key, size_t output_bits) :
```
randombit, line 39: This is inconsistent with how the rest of the library treats objects, which are initialized un-keyed, later can be keyed/rekeyed. So I think both this constructor and the second taking keys as bytes should be removed here.

src/lib/hash/blake2/blake2b.cpp
```diff
@@ -181,26 +213,50 @@ void BLAKE2b::final_result(uint8_t output[])
...
+
 std::string BLAKE2b::name() const
    {
    return "BLAKE2b(" + std::to_string(m_output_bits) + ")";
    }
 
 HashFunction* BLAKE2b::clone() const
    {
-   return new BLAKE2b(m_output_bits);
+   return new BLAKE2b(m_padded_key_buffer.data(), m_key_size, m_output_bits);
```
randombit, line 228: This clones with the key set which is inconsistent with the rest of the library (`clone` is a terrible name given that it just copies the object type and not the object state, but changing it for just one type is more confusing than anything else)

src/lib/hash/blake2/blake2b.h
```diff
@@ -17,27 +18,41 @@ namespace Botan {
...
+      Key_Length_Specification key_spec() const override;
 
       HashFunction* clone() const override;
       std::string name() const override;
       void clear() override;
 
       std::unique_ptr<HashFunction> copy_state() const override;
 
-   private:
+   public:
```
randombit, line 49: `key_schedule` as well as `add_data` and `final_result` should be `private` - they are only called via other functions defined on the base class.

[This is mostly harmless now that the algorithm types aren't publicaly exposed but I'd like to keep it clear what is or is not callable APIs]

src/lib/mac/blake2mac/blake2bmac.cpp
```diff
@@ -0,0 +1,36 @@
+/*
+* BLAKE2b MAC
+* (C) 1999-2007,2014 Jack Lloyd
```
randombit, line 3: You should add your (C) statement here also

src/lib/mac/blake2mac/blake2bmac.h
```diff
@@ -0,0 +1,62 @@
...
+      verify_key_set(m_cipher.key_size() > 0);
+      m_cipher.final_result(out);
+      }
+
+   explicit BLAKE2bMAC(size_t output_bits = 512);
+
+   BLAKE2bMAC(const BLAKE2bMAC&) = delete;
+   BLAKE2bMAC& operator=(const BLAKE2bMAC&) = delete;
+private:
+   BLAKE2b m_cipher;
```
randombit, line 55: `m_cipher` is a weird name for this - maybe `m_blake`?

src/lib/mac/blake2mac/info.txt
```diff
@@ -0,0 +1,7 @@
+<defines>
+BLAKE2BMAC -> 20202311
```
randombit, line 2: Date format is YYYYMMDD not YYYYDDMM

(I should probably add a check for this in configure)

src/tests/data/mac/blake2bmac.vec
```diff
@@ -0,0 +1,1051 @@
+[BLAKE2b(224)]
```
randombit, line 1: Add a comment indicating where these came from

randombit (changes requested):
Overall looks great, some small changes required but happy to accept this feature and your general approach is fine.

src/lib/mac/blake2mac/info.txt
AtomicNibble: Yeah I tried to guess what it was haha

AtomicNibble:
Will make the changes suggested, thanks for taking the time to look over it.

src/tests/data/mac/blake2bmac.vec
AtomicNibble: The ones for 224, 256 and 384 I made manually(could not find any online not even in the RFC 😢) but did check they match outpout of Go for 256 and 384: https://play.golang.org/p/ktgwCpIfc19

Will add comments and link to source of 512 test vectors.

src/lib/hash/blake2/blake2b.cpp
AtomicNibble: I only added them as it's quite common in other libs to support keyed construction for Blake2b, but happy to remove it.

mouse07410:
I think it makes sense to have a keyed construction.

randombit:
A keyed constructor is fine in principle but it does not make sense that just one particular algorithm supports it when the rest do not.

randombit:
Looks good. I am not happy that all of the test vectors use a single key length (seems like it would be easy to miss a bug when handling shorter keys) but that seems to be a problem with the published BLAKE2 vectors. I may try generating some with OpenSSL or something to test other key lengths.

PR #2526, opened 2020-11-30
Test install command

hrantzsch:
While trying to update the Botan Conan package to 2.17.2 we discovered that `make install` silently fails on Windows ([conan-center PR](https://github.com/conan-io/conan-center-index/pull/3607))

The issue seem to be the forward slashes introduced to `C:/PythonXX/python.exe` in 72bbed1c5444455a51ce0d5f122b8ab750b1c1a6. When run through nmake, Windows fails to find python at this path and cannot run install.py. We only noticed because Conan complains about building an empty package.

An additional problem occurs when python is not installed in `C:\PythonXX`, but for example in `C:\Program Files\Python`. In this case, python cannot be found due to the space character in the path.

Before trying to fix the issue we wanted to reproduce it in the Botan CI. Unfortunately, I haven't found a good place for this test, yet. The way it is implemented now is quite hacky. Also, I hit different errors when I wanted to test it on appveyor, so I'm curious what our CI will say here.

### DIFF
--- a/src/scripts/ci_check_install.py
+++ b/src/scripts/ci_check_install.py
@@ -1,12 +1,19 @@
 #!/usr/bin/env python
 
-"""This script is used to validate the results of `make install`"""
+"""
+Botan CI check installation script
+This script is used to validate the results of `make install`
+
+(C) 2020 Jack Lloyd, René Meusel, Hannes Rantzsch
+
+Botan is released under the Simplified BSD License (see license.txt)
+"""
 
 import os
 import sys
 
 def has_extension(filename, extensions):
-    for ext in [ext for ext in extensions if ext != ""]:
+    for ext in [ext for ext in extensions]:
         if filename.endswith(".%s" % ext):
             return True
     return False
@@ -20,12 +27,12 @@ def is_header_file(filename):
 def main():
     if len(sys.argv) < 2:
         print("Usage: %s <install_prefix>" % sys.argv[0])
-        sys.exit(1)
+        return 1
     install_prefix = sys.argv[1]
 
     if not os.path.isdir(install_prefix):
         print('Error: install_prefix "%s" is not a directory' % install_prefix)
-        sys.exit(1)
+        return 1
 
     found_libs = False
     found_headers = False
--- a/src/scripts/ci_check_install.py
+++ b/src/scripts/ci_check_install.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python
+
+"""This script is used to validate the results of `make install`"""
+
+import os
+import sys
+
+def has_extension(filename, extensions):
+    for ext in [ext for ext in extensions if ext != ""]:
+        if filename.endswith(".%s" % ext):
+            return True
+    return False
+
+def is_lib_file(filename):
+    return has_extension(filename, ["so", "a", "dll", "dylib", "lib"])
+
+def is_header_file(filename):
+    return has_extension(filename, ["h", "hpp", "h++", "hxx", "hh"])
+
+def main():
+    if len(sys.argv) < 2:
+        print("Usage: %s <install_prefix>" % sys.argv[0])
+        sys.exit(1)
+    install_prefix = sys.argv[1]
+
+    if not os.path.isdir(install_prefix):
+        print('Error: install_prefix "%s" is not a directory' % install_prefix)
+        sys.exit(1)
+
+    found_libs = False
+    found_headers = False
+
+    for (_, _, filenames) in os.walk(install_prefix):
+        for filename in filenames:
+            if is_header_file(filename):
+                found_headers = True
+            elif is_lib_file(filename):
+                found_libs = True
+        if found_libs and found_headers:
+            return 0
+
+    print("Error: installation incomplete. Found headers: %s. Found libs: %s. install_prefix was %s"
+            % (found_headers, found_libs, install_prefix))
+    return 1
+
+
+if __name__ == '__main__':
+    sys.exit(main())
--- a/configure.py
+++ b/configure.py
@@ -1994,7 +1994,7 @@ def create_template_vars(source_paths, build_paths, options, modules, cc, arch, 
     def choose_python_exe():
         exe = sys.executable
 
-        if exe[1] == ':': # Windows style paths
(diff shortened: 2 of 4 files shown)


hrantzsch:
Only VS 14 (2015) seems to have this problem.

```
Running 'nmake -k install' ...
Microsoft (R) Program Maintenance Utility Version 14.00.24210.0
Copyright (C) Microsoft Corporation.  All rights reserved.
	C:/Python27/python.exe src/scripts/build_docs.py --build-dir="build"
	C:/Python27/python.exe src/scripts/install.py --prefix="c:\users\appveyor\appdata\local\temp\1\botan-install" --build-dir="build" --bindir=c:\users\appveyor\appdata\local\temp\1\botan-install\bin --libdir=c:\users\appveyor\appdata\local\temp\1\botan-install\lib --docdir=docs --includedir=include
  File "src\scripts\ci_build.py", line 675, in <module>
    sys.exit(main())
  File "src\scripts\ci_build.py", line 670, in main
    run_cmd(cmd, root_dir)
  File "src\scripts\ci_build.py", line 349, in run_cmd
    if not os.listdir(lib_install_dir):
WindowsError: [Error 3] The system cannot find the path specified: u'c:\\users\\appveyor\\appdata\\local\\temp\\1\\botan-install\\*.*'
```

2017 and 2019 work fine. My test doesn't work on MinGW yet, which runs the successfully installation (and says `INFO: Botan 2.17.2 installation complete`) but the path is still not found. Maybe due to escaping and back-slashes again.

reneme:
Mhh, both [2.17.0](https://github.com/conan-io/conan-center-index/pull/3741) and [2.17.1](https://github.com/conan-io/conan-center-index/pull/3460) are building smoothly on CCI's CI. It really seems to be the commit you mentioned above...

reneme:
Oh, for reference: I just found that the `Makefile` used to rely on `python.exe` being available in the `$PATH` until very recently (https://github.com/randombit/botan/commit/f617a60cbdc825d1798e2d61575a5cfeb2bc2aa5), i.e. 2.17.1.

hrantzsch:
It should work now. We
 * added a script to validate the installation in the CI
 * added more quotes to the install command in makefile.in. without this, mingw seemed to be confused by the unquoted bindir and libdir
 * changed the `\` to `/` fix in configure.py to only run on mingw

hrantzsch:
Two things we might want to do later: Probably this should be ported to 3.0 as well; We didn't handle spaces in the python path yet.

randombit:
For context on the Python-in-PATH change - it seems that recent versions of `macOS` have a `python` in `PATH` but it is a symlink to some directory which is not in `PATH` and the actual executable is installed as `Python` rather than `python` (WHY APPLE WHY?!?!). So if you try to invoke `basename(sys.executable)` you end with `Python` which is not found in the `PATH`.

Possibly a partial fix here is to only use the full path on  macOS and go back to the basename logic elsewhere?

src/scripts/ci_check_install.py
```diff
@@ -0,0 +1,48 @@
...
+
+def main():
+    if len(sys.argv) < 2:
+        print("Usage: %s <install_prefix>" % sys.argv[0])
+        sys.exit(1)
+    install_prefix = sys.argv[1]
+
+    if not os.path.isdir(install_prefix):
+        print('Error: install_prefix "%s" is not a directory' % install_prefix)
+        sys.exit(1)
```
randombit, line 28: Use return here

src/scripts/ci_check_install.py
```diff
@@ -0,0 +1,48 @@
...
+def is_lib_file(filename):
+    return has_extension(filename, ["so", "a", "dll", "dylib", "lib"])
+
+def is_header_file(filename):
+    return has_extension(filename, ["h", "hpp", "h++", "hxx", "hh"])
+
+def main():
+    if len(sys.argv) < 2:
+        print("Usage: %s <install_prefix>" % sys.argv[0])
+        sys.exit(1)
```
randombit, line 23: Just return 1

src/scripts/ci_check_install.py
```diff
@@ -0,0 +1,48 @@
...
+    install_prefix = sys.argv[1]
+
+    if not os.path.isdir(install_prefix):
+        print('Error: install_prefix "%s" is not a directory' % install_prefix)
+        sys.exit(1)
+
+    found_libs = False
+    found_headers = False
+
+    for (_, _, filenames) in os.walk(install_prefix):
```
randombit, line 33: This logic seems easy to fool in the sense that any random directory with an `.h` and a `.so` will pass. A better test would be to have `build_config.json` export the list of installed headers, then the script could verify that every header so named is in the expected spot.

That said this at least verifies that something was installed at all and can always be improved from there.

randombit (commented):
Some minor changes but seems fine, and it's good to have some tests that install does something.

I do wonder if we should switch (back) to the basename(sys.executable) logic except on macOS which would I think solve the spaces in path problem for Windows/minGW

src/scripts/ci_check_install.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python
+
+"""This script is used to validate the results of `make install`"""
+
+import os
+import sys
+
+def has_extension(filename, extensions):
+    for ext in [ext for ext in extensions if ext != ""]:
```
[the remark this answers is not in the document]
hrantzsch: Right, that's not needed anymore

src/scripts/ci_check_install.py
hrantzsch: I suggest we improve the test script on master later and go with the simple version for now.

hrantzsch:
We kept using the full path but added quotes for both the python executable path and the path to the script in the make file. This works for the platforms we have tried manually (VS 14, Linux, MacOS Catalina). I hope it works on all other platforms as well.

src/scripts/ci_check_install.py
randombit: Yeah sorry didn't mean to suggest you do that here, what you have is fine for now and we can do something a little nicer/more extensive later

randombit:
Thanks. Can you create a similar PR for master?

hrantzsch:
Oh no! VS doesn't like René's `é` :sob: 
Can we revert?

PR #2527, opened 2020-11-30
Additional blake2b mac test vectors.

AtomicNibble:
Found some more test vectors using different key lengths for blake2b mac if intrested.

https://raw.githubusercontent.com/torvalds/linux/master/crypto/testmgr.h

PS: I noticed you fixed the formating for blake2b mac class, sorry for the incorrect formating.


randombit:
Thanks!

PR #2531, opened 2020-12-02
Improve installation checks

randombit:
(And also some changes to the install script itself)

### DIFF
--- a/src/scripts/ci_check_install.py
+++ b/src/scripts/ci_check_install.py
@@ -23,7 +23,16 @@ def verify_library(build_config):
 
     found_libs = set([])
 
-    lib_re = re.compile(r'^(lib)?botan-3\.(so|a|dll|dylib|lib)$')
+    major_version = int(build_config["version_major"])
+
+    if build_config['compiler'] == 'msvc':
+        expected_lib_format = r'^botan\.(dll|lib)$'
+    elif build_config['os'] == 'macos':
+        expected_lib_format = r'^libbotan-%d\.(a|dylib)$' % (major_version)
+    else:
+        expected_lib_format = r'^libbotan-%d\.(a|so)$' % (major_version)
+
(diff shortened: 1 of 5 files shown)


hrantzsch (approved):
Looks great. At least on Linux, Conan is also still happy with packaging the installation.

src/scripts/ci_check_install.py
```diff
@@ -12,44 +12,81 @@
...
+        print('Error: libdir "%s" is not a directory' % lib_dir)
+        return False
 
-def is_lib_file(filename):
-    return has_extension(filename, ["so", "a", "dll", "dylib", "lib"])
+    found_libs = set([])
 
-def is_header_file(filename):
-    return has_extension(filename, ["h", "hpp", "h++", "hxx", "hh"])
+    lib_re = re.compile(r'^(lib)?botan-3\.(so|a|dll|dylib|lib)$')
```
reneme, line 26: I know the hard coded major version changes only once every few years: But wouldn't it be worthwhile to pull it from the `build_config['version_major']`? Or even better: Implement the `verify_library` similarly to the `verify_includes` where `./configure.py` stores a list of expected library files in the `build_config`?

reneme (approved):
It appears to me that the build on AppVeyor `VC2019p` fails because it produces a `botan.dll` and `botan.lib` while the `ci_check_install.py` is unsuccessfully checking for a `botan-3.***`.

Some relevant link commands from the CI output:

```
link /DLL [... many botan object files ...] crypt32.lib user32.lib ws2_32.lib /OUT:.\botan.dll
```

```
link [... many botatn test object files ...]  .\botan.lib crypt32.lib user32.lib ws2_32.lib /OUT:.\botan-test.exe
```

The extended installation checking looks good to me though, despite one suggestion.

PR #2537, opened 2020-12-09
Remove support for SHA-1 signatures in TLS 1.2

randombit:


MartinOsterta (commented):
why remove SHA-1 completely in TLS1.2. Wouldn't it be enough to just disable it by default by configuration, but a user might have the possibility to enable it if for whatever reason SHA-1 is still required in specific scenarios?

randombit:
https://tools.ietf.org/html/draft-ietf-tls-md5-sha1-deprecate-04 calls for SHA-1 to be removed/disabled in TLS 1.2 due to actively known collision attacks. We may be only a few years away from it being possible to real-time create a collision and MITM connections authenticated with a SHA-1 ServerKeyExchange. It does not seem safe to support any longer, even as a disabled-by-default option (in fact in 2.x it is already disabled by default). Since 3.0 will be supported for some time into the future (say through 2027 or perhaps later) I would like to remove as many obviously problematic features as possible now from the start - rather than having an attack be proven and then have to scramble to address it.

Are there specific scenarios you have in mind where SHA-1 is still required to be negotiated in TLS 1.2? The only one I can think of is that you might have a hardware device which supports SHA-1 signatures but not SHA-256. But I would think any such tokens would be very old by now.

mouse07410:
My opinion, based on experience with hardware tokens and HSM devices:
- Keep SHA-1 in the crypto library;
- Remove SHA-1 from all the TLS code.

Thanks

randombit:
There is no chance of SHA-1 being removed from the library in general. There are many uses where SHA-1 is still ok(-ish) like HMAC, and it's still very commonly used. But I'm trying to position TLS to be in the best possible position for long term support, and with as few obviously insecure "features" as possible.

MartinOsterta:
I do not have a specific scenario in mind, I was just wondering if there are any old servers running in the wilderness which still need SHA-1 for very specific scenarios, in which case it could be helpful to enable SHA-1. More like a safety net. But probably you are right and in this case older/adapted botan library has to be used.

PR #2544, opened 2020-12-13
Solaris and Illumos distributions support getentropy.

devnexen:


randombit:
Thanks.

Looks like it was added in Solaris 11.3 (October 2017) and per https://www.solaris.wtf/blog/solaris-11-3-support-end-is-near/ 11.3 is already no longer supported so seems safe to have it enabled by default.

PR #2552, opened 2020-12-15
Convert ASN1_Tag to an enum class

randombit:


src/lib/asn1/asn1_obj.cpp
```diff
@@ -228,7 +230,8 @@ bool maybe_BER(DataSource& source)
       throw Stream_IO_Error("ASN1::maybe_BER: Source was empty");
       }
 
-   if(first_u8 == (SEQUENCE | CONSTRUCTED))
+   // CONSTRUCTED SEQUENCE
+   if(first_u8 == 0x30)
```
MartinOsterta, line 234: why use a hardcoded value here instead of defines?

MartinOsterta (changes requested):

PR #2570, opened 2020-12-31
Mac M1 build update, cpu features detection.

devnexen:


src/lib/utils/cpuid/cpuid_arm.cpp
```diff
@@ -196,6 +196,14 @@ uint64_t CPUID::CPUID_Data::detect_cpu_features(size_t* cache_line_size)
 
 #elif defined(BOTAN_USE_GCC_INLINE_ASM) && defined(BOTAN_TARGET_ARCH_IS_ARM64)
 
+#if defined(BOTAN_TARGET_OS_IS_MACOS)
```
randombit, line 199: This is inside a `USE_GCC_INLINE_ASM` block which doesn't seem necessary. I think a new preprocessor arm is needed eg

```
#elif defined(BOTAN_TARGET_OS_IS_MACOS) && defined(BOTAN_TARGET_ARCH_IS_ARM64)
  // sysctlbyname
#elif defined(BOTAN_USE_GCC_INLINE_ASM) && defined(BOTAN_TARGET_ARCH_IS_ARM64)
  // existing code
#endif
```

Also, does this API work on iOS, or only macOS?

randombit (changes requested):

src/lib/utils/cpuid/cpuid_arm.cpp
devnexen: yes good point :-)

Issue #2581, opened 2021-01-09
Wrong value for ASN1_Tag::PRIVATE


sledgehammer999:
According to [this](https://en.wikipedia.org/wiki/X.690#Encoding) the PRIVATE class has a value of `0b1100'000` aka `0xc0`

But it in https://github.com/randombit/botan/blob/854b79287818a01137aa78e59e36ab10240bf13b/src/lib/asn1/asn1_obj.h#L24
it is defined as
```c++
enum class ASN1_Tag : uint32_t {
   UNIVERSAL        = 0x00,
   APPLICATION      = 0x40,
   CONTEXT_SPECIFIC = 0x80,

   CONSTRUCTED      = 0x20,

   PRIVATE          = CONSTRUCTED | CONTEXT_SPECIFIC,
```

aka `0b1010'0000` or `0xa0`

Git blame shows that this was introduced with https://github.com/randombit/botan/commit/4e43080954be57e362feb1cc8202bfd42117e286 but I can't find any further info.

Is this truly a bug or is there a reason for defining the value as `0xa0`?

randombit:
Thanks for pointing this out. I think actually the problem is that `PRIVATE` is named incorrectly and has nothing to do with private class in DER (which is not used by anything I've encountered). It actually is being used to handle context-specific explicitly tagged items. I'm not sure why I used private for this. Certainly it's wrong.

sledgehammer999:
OK, that makes more sense now.

Btw, I noticed that you very recently converted it to `enum class`. Have you considered splitting the class and type tags to different `enum class`es to improve type safety even more?
Woud you object to such a PR?

randombit:
Such a change makes sense. And if it is going to happen, now is the time (`master` is in runup to a 3.0-alpha where we already are breaking compat guarantees, but anything that is in final 3.0.0 we'll be stuck with for years).

sledgehammer999:
Closing this as fixed by @2584.

PR #2584, opened 2021-01-11
Split ASN1 tags enum class into 2 enum classes

sledgehammer999:
This should improve type safety even more.

I did this fairly quickly without delving much into the code logic.

I changed the old `ASN1_Tag::PRIVATE` to `ASN1_Class::EXPLICIT_CONTEXT_SPECIFIC` as per #2581. Of course this is a placeholder name if you want to change it.
I replace all instaces of `PRIVATE` to the new name, HOWEVER I am not able to verify that those places used it with the correct meaning. I haven't read the relevant RFCs.

Disclaimer: My feelings won't be hurt if you don't want this PR.


src/lib/asn1/asn1_obj.h
```diff
@@ -19,17 +19,24 @@ class BER_Decoder;
...
+    APPLICATION      = 0b0100'0000,
+    CONTEXT_SPECIFIC = 0b1000'0000,
+    PRIVATE          = 0b1100'0000,
 
-   CONSTRUCTED      = 0x20,
+    CONSTRUCTED      = 0b0010'0000,
+    EXPLICIT_CONTEXT_SPECIFIC = CONSTRUCTED | CONTEXT_SPECIFIC,
 
-   PRIVATE          = CONSTRUCTED | CONTEXT_SPECIFIC,
+    NO_OBJECT        = 0xFF00
```
sledgehammer999, line 33: I didn't know how to correctly handle code that used this enum value so both enum classes have this value.
Maybe instead of a sentinel value, use an `std::optional`? Sadly, I am not very familiar with your code base so I can't make any serious proposal here.

src/lib/asn1/asn1_obj.h
```diff
@@ -19,17 +19,24 @@ class BER_Decoder;
...
-   APPLICATION      = 0x40,
-   CONTEXT_SPECIFIC = 0x80,
+enum class ASN1_Class : uint32_t {
+    UNIVERSAL        = 0b0000'0000,
+    APPLICATION      = 0b0100'0000,
+    CONTEXT_SPECIFIC = 0b1000'0000,
+    PRIVATE          = 0b1100'0000,
 
-   CONSTRUCTED      = 0x20,
+    CONSTRUCTED      = 0b0010'0000,
```
sledgehammer999, line 30: I wrote the class values as binary literals because it was easier for me to understand what was going on. If you want, I can convert them to hex literals.

src/lib/asn1/asn1_obj.h
randombit: Using `std::optional` is probably viable. Only a few places care about or have to consider an unset tag (mostly in the decoder)

src/lib/asn1/asn1_obj.h
randombit: Yes for this binary is better (it wasn't binary before since until this past November we were still C++11-only)

src/lib/asn1/asn1_obj.h
```diff
@@ -118,34 +140,35 @@ class BOTAN_PUBLIC_API(2,0) BER_Object final
 
       BER_Object& operator=(BER_Object&& other) = default;
 
-      bool is_set() const { return m_type_tag != ASN1_Tag::NO_OBJECT; }
+      bool is_set() const { return m_type_tag != ASN1_Type::NO_OBJECT; }
 
-      ASN1_Tag tagging() const { return type_tag() | class_tag(); }
+      uint32_t tagging() const { return type_tag() | class_tag(); }
```
randombit, line 145: I don't like how we lose the type here but don't really see a good way of avoiding it.

sledgehammer999: Yeah my me too. I am open to suggestions.

sledgehammer999:
Wow! You merged it? Thank you! I didn't expect that it would be done so quickly.

PS: Are you sure you want leave this `EXPLICIT_CONTEXT_SPECIFIC` as the future name? I intended it as a placeholder for review. Anyway, I don't care much.

randombit:
I think the name is fine, or at least I can't think of a better one.

For cleanup changes like this where review is easy I like to either merge or reject outright pretty quickly. It's harder for functionality changes because then I have to think about if the functionality makes sense for us to support at all, does it expand attack surface, how do we test this, etc etc. Here, either we want to do it or we don't, don't bikeshed it to death.

PR #2587, opened 2021-01-12
Various ASN1 API changes

randombit:


sledgehammer999:
Suggestion: How about dropping the `ASN1_` prefix on the various enums/classes/functions and instead enclosing them in an `ASN1` namespace?

PR #2596, opened 2021-01-17
Fix System_RNG only filling the first 4GB of the buffer for len >4GB

AtomicNibble:
I noticed that some System_Rng impl will silenty only fill the first 4GB if a buffer larger than 4GB is passed.
Is likley a super rare use case but I think it should either be supported or error.

I have included a test but problem is that tests requires allocating a >4GB buffer so not sure if should be anabled by default.

Feedback is welcome.
Tom.

### DIFF
--- a/src/lib/rng/system_rng/system_rng.cpp
+++ b/src/lib/rng/system_rng/system_rng.cpp
@@ -1,6 +1,7 @@
 /*
 * System RNG
 * (C) 2014,2015,2017,2018 Jack Lloyd
+* (C) 2021 Tom Crowley
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
@@ -48,9 +49,24 @@ class System_RNG_Impl final : public RandomNumberGenerator
 
       void randomize(uint8_t buf[], size_t len) override
          {
-         bool success = m_rtlgenrandom(buf, ULONG(len)) == TRUE;
-         if(!success)
-            throw System_Error("RtlGenRandom failed");
+         const size_t limit = std::numeric_limits<ULONG>::max();
+
+         uint8_t* pData = buf;
+         size_t bytesLeft = len;
+         while (bytesLeft > 0)
+            {
+            const ULONG blockSize = static_cast<ULONG>(std::min(bytesLeft, limit));
+
+            const bool success = m_rtlgenrandom(pData, blockSize) == TRUE;
+            if (!success)
+               {
+               throw System_Error("RtlGenRandom failed");
+               }
+
+            BOTAN_ASSERT(bytesLeft >= blockSize, "Block is oversized");
+            bytesLeft -= blockSize;
+            pData += blockSize;
+            }
          }
 
       void add_entropy(const uint8_t[], size_t) override { /* ignored */ }
@@ -86,9 +102,24 @@ class System_RNG_Impl final : public RandomNumberGenerator
 
       void randomize(uint8_t buf[], size_t len) override
          {
-         NTSTATUS ret = ::BCryptGenRandom(m_prov, static_cast<PUCHAR>(buf), static_cast<ULONG>(len), 0);
-         if(ret != STATUS_SUCCESS)
-            throw System_Error("System_RNG call to BCryptGenRandom failed", ret);
+         const size_t limit = std::numeric_limits<ULONG>::max();
+
+         uint8_t* pData = buf;
+         size_t bytesLeft = len;
+         while (bytesLeft > 0)
+            {
+            const ULONG blockSize = static_cast<ULONG>(std::min(bytesLeft, limit));
+
+            const NTSTATUS ret = BCryptGenRandom(m_prov, static_cast<PUCHAR>(pData), blockSize, 0);
+            if (ret != STATUS_SUCCESS)
+               {
+               throw System_Error("System_RNG call to BCryptGenRandom failed", ret);
+               }
+
+            BOTAN_ASSERT(bytesLeft >= blockSize, "Block is oversized");
+            bytesLeft -= blockSize;
+            pData += blockSize;
+            }
          }
 
       void add_entropy(const uint8_t in[], size_t length) override
--- a/src/tests/test_rng.cpp
+++ b/src/tests/test_rng.cpp
@@ -771,6 +771,21 @@ class System_RNG_Tests final : public Test
             rng.add_entropy(out_buf.data(), out_buf.size());
             }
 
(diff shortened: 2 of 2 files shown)


src/lib/rng/system_rng/system_rng.cpp
```diff
@@ -48,9 +49,24 @@ class System_RNG_Impl final : public RandomNumberGenerator
 
       void randomize(uint8_t buf[], size_t len) override
          {
-         bool success = m_rtlgenrandom(buf, ULONG(len)) == TRUE;
-         if(!success)
-            throw System_Error("RtlGenRandom failed");
+         const uint32_t limit = std::numeric_limits<ULONG>::max();
```
randombit, line 52: If this was a `size_t` you can avoid having to specify the type for `std::min(byteLeft, limit)`

src/lib/rng/system_rng/system_rng.cpp
```diff
@@ -48,9 +49,24 @@ class System_RNG_Impl final : public RandomNumberGenerator
...
+
+         uint8_t* pData = buf;
+         size_t bytesLeft = len;
+         while (bytesLeft > 0)
+            {
+            ULONG blockSize = static_cast<ULONG>(std::min<size_t>(bytesLeft, limit));
```
randombit, line 58: const

src/lib/rng/system_rng/system_rng.cpp
```diff
@@ -48,9 +49,24 @@ class System_RNG_Impl final : public RandomNumberGenerator
...
+
+            bool success = m_rtlgenrandom(pData, blockSize) == TRUE;
```
randombit, line 60: const

src/lib/rng/system_rng/system_rng.cpp
```diff
@@ -86,9 +102,24 @@ class System_RNG_Impl final : public RandomNumberGenerator
 
       void randomize(uint8_t buf[], size_t len) override
          {
-         NTSTATUS ret = ::BCryptGenRandom(m_prov, static_cast<PUCHAR>(buf), static_cast<ULONG>(len), 0);
-         if(ret != STATUS_SUCCESS)
-            throw System_Error("System_RNG call to BCryptGenRandom failed", ret);
+         const uint32_t limit = std::numeric_limits<ULONG>::max();
```
randombit, line 105: size_t

src/lib/rng/system_rng/system_rng.cpp
```diff
@@ -86,9 +102,24 @@ class System_RNG_Impl final : public RandomNumberGenerator
...
+
+         uint8_t* pData = buf;
+         size_t bytesLeft = len;
+         while (bytesLeft > 0)
+            {
+            ULONG blockSize = static_cast<ULONG>(std::min<size_t>(bytesLeft, limit));
```
randombit, line 111: const

src/lib/rng/system_rng/system_rng.cpp
```diff
@@ -86,9 +102,24 @@ class System_RNG_Impl final : public RandomNumberGenerator
...
+
+            NTSTATUS ret = BCryptGenRandom(m_prov, static_cast<PUCHAR>(pData), blockSize, 0);
```
randombit, line 113: const

src/tests/test_rng.cpp
```diff
@@ -771,6 +771,21 @@ class System_RNG_Tests final : public Test
             rng.add_entropy(out_buf.data(), out_buf.size());
             }
 
+#if (BOTAN_MP_WORD_BITS == 64)
```
randombit, line 774: Use `if constexpr(sizeof(size_t) > 4)` for this

src/tests/test_rng.cpp
```diff
@@ -771,6 +771,21 @@ class System_RNG_Tests final : public Test
...
+         // Pass buffer with a size greater than 32bit
+         const size_t size32BitsMax = std::numeric_limits<uint32_t>::max();
+         const size_t checkSize = 1024;
+         std::vector<uint8_t> large_buf(size32BitsMax + checkSize);
+         std::memset(large_buf.data() + size32BitsMax, 0xfe, checkSize);
+
+         rng.randomize(large_buf.data(), large_buf.size());
+
+         std::vector<uint8_t> check_buf(checkSize);
+         std::memset(check_buf.data(), 0xfe, checkSize);
```
randombit, line 784: This could just be `std::vector<uint8_t> check_buf(checkSize, 0xFE)`

randombit (approved):
Thanks for catching this and overall the patch looks good, just some minor style comments.

One thing I'm a bit worried about is the test. On Windows 64-bit the odds are good you have way more than 4 GB of memory and we can run this. But I have plenty of hardware (eg little ARMv8 SOC boards) with 64-bit CPUs but <= 4G total memory, where this test will just crash or worse (eg OOM kill something useful).

randombit:
CI failure may be related, probably S390x image has < 4 G memory available.

AtomicNibble:
Yes I was also concerned about the test but was not sure how else could test such a case.
Since the system_rng will write data to the whole range can't do stuff like only commit some of the pages.

I will make the style changes.

src/tests/test_rng.cpp
AtomicNibble: True!

src/tests/test_rng.cpp
AtomicNibble: I did have this originally but reverted it as I'm still on 2.x but I see 3.0 supports c++17 so will change.

randombit:
It might be possible to set up the test by using mmap plus huge pages, 4 G = 2048 2M huge pages, but instead of creating 2K pages, you create one and use `mremap` with `MREMAP_FIXED` so that all of the virtual addresses are linear, so using only 2M of physical memory (plus page table overhead etc). Overall kind of a pain though, and will be fragile, *and* doesn't work on Windows which is where the actual bugs were anyway... so probably not worth while.

I considered adding a function is `OS` namespace that returns the total physical memory since that's available at least on Linux, FreeBSD, and Windows, probably most other systems too. But ... there is really no call for this function I can think of in the library proper, so supporting it feels a little weird. (And, just because the machine has say 6 GB of memory doesn't mean our test should, or even can, take up 4 GB of it).

So I guess what I'll do is just guard this test with `--run-long-tests` flag, and modify CI so that at least one Windows build uses it (which should have already been happening).

AtomicNibble:
Sounds like a plan, thanks!

PR #2598, opened 2021-01-19
get_cpu_cycle_counter on ARM64 implementation.

devnexen:
recommended register for Darwin devices as well.


randombit:
It was my understanding that the cycle counter is protected from user space in at least some situations. On my Odroid-C2 (Cortex-A53) this

```
  asm volatile("isb; mrs %0, cntvct_el0" : "=r" (rtc));
```

Crashes with SIGILL. So it's not clear to me we can use this reliably.

(We can probably use the Mac version, since if Apple is allowing it now they will presumably allow it in future versions also)

PR #2599, opened 2021-01-22
Update ECDSA timing test vectors

securitykernel:
According to @jurajsomorovsky, under normal circumstances the current ones can not be generated in an attacker scenario. More realisitic is a difference in one byte only.

Already prepared this in July, but forgot to push.


randombit:
I would agree that 1 vs max will never be seen in practice but wouldn't those values serve to accentuate any possible channel? That is if 1 bit vs max is indistinguishable then certainly max-8 and max are also.

jurajsomorovsky:
You are true that in certain circumstances, this can give a hint about the side channel. The question is how practical it is and whether the test catches what you want to see.
The problem with the 1 vs max test is that in typical crypto implementations, you will always see a side channel; the side channel will come from the bignum library, which allocates buffers based on the integer size. Allocating buffers for 1 vs max will definitely cause differences, which will be visible in the ECDSA computation as a side channel and we cannot get rid of this side channel without making big numbers completely constant time which is super hard (I am sorry, I currently do not know how exactly botan handles big numbers). 
Therefore, I was proposing to make practical tests with a difference of 1 byte (we can extend this for 2/3/4 bytes ?), which will at least catch practical attacks and give us hints whether the implementation is vulnerable in those scenarios.

David Benjamin also gave us a hint on the constant time implementation of bignums in boringssl while performing the raccoon disclosure: https://github.com/openssl/openssl/issues/6640 (I think the issue is relevant in this context)

Issue #2595, opened 2021-01-16
Possibility to register custom RNG with FFI API


securitykernel:
It is currently not possible to use a custom RNG with the FFI API, which is one of the main advantages of the C++ API. `botan_rng_init()` only supports passing an RNG name that maps to one of the built-in RNGs, such as `AutoSeeded_RNG` or `System_RNG`. This is fine for most standard use cases, but falls short in more advanced scenarios where you have a custom RNG inside a system and you need to use it from several implementations, including projects which are C only. One such scenario is in the strongswan VPN project. Strongswan offers a plugin interface for custom RNG implementations. Other plugin interfaces, e.g., ECDSA signature generation, can use this RNG interface to request random bits. The Botan plugin for strongswan however makes no use of the RNG interface in strongswan, simply because all Botan FFI functions require passing a `botan_rng_t`, which can only be created using `botan_rng_init()`.

What would be required is the possibility to register a custom RNG implementation with the FFI API, something along the lines of:

```c++
BOTAN_PUBLIC_API(2,17) int botan_rng_init_custom(botan_rng_t* rng_out, const char* rng_name, void* context, void(* init_cb)(), void(* destroy_cb)(),  void(* get_cb)(void* context, uint8_t* out, size_t out_len));

...

int botan_rng_init_custom(botan_rng_t* rng_out, const char* rng_name, void* context, void(* init_cb)(), void(* destroy_cb)(),  void(* get_cb)(void* context, uint8_t* out, size_t out_len))
{
return ffi_guard_thunk(__func__,[=]() -> int {
   if(rng_out == nullptr)
      return BOTAN_FFI_ERROR_NULL_POINTER;

   std::unique_ptr<Botan::RandomNumberGenerator> rng;
   class Custom_RNG : public Botan::RandomNumberGenerator
      {
      public:
         Custom_RNG(const std::string& name, void* context, void(* init_cb)(), void(* get_cb)(void* context, uint8_t* out, size_t out_len)) :
            m_name(name)
            {
               init_cb();
               m_cbs.context = context;
               m_cbs.get_cb = get_cb;
            }

        void randomize(uint8_t output[], size_t length) overrride
        {
            m_cbs.get_cb(m_cbs.context, output, length);
        }

        bool accepts_input() const override
        {
           // TODO
           return false;
        }

        void add_entropy(const uint8_t input[], size_t length) override
        {
            // TODO
        }    

        std::string name() const override
        {
           return m_name;
        }

        void clear() override
        {
[...]

PR #2600, opened 2021-01-22
Add API to register custom RNG with FFI API

securitykernel:
Adds a C API function that allows to register a custom `Botan::RandomNumberGenerator` implementation via callback functions as outlined in #2595.

For our use case in strongswan (https://github.com/strongswan/strongswan/pull/192), having the ability to register a function for `RandomNumberGenerator::randomize()` is sufficient, but I added a callback for `add_entropy()`, too, for generality. I am not sure what to do with `clear()`. And I am also not sure whether the exceptions that are generated are fitting here, so I'll leave this as draft for the moment waiting for feedback.

### DIFF
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -201,6 +201,17 @@ typedef struct botan_rng_struct* botan_rng_t;
 BOTAN_PUBLIC_API(2,0) int botan_rng_init(botan_rng_t* rng, const char* rng_type);
 
 /**
+* Initialize a custom random number generator with a set of callbacks
+* @param rng rng object
+* @param rng_name name of the rng
+* @param get_cb Callback for getting random bytes from the rng
+* @param add_entry_cb Callback for adding entropy to the rng
+*/
+BOTAN_PUBLIC_API(3,0) int botan_rng_init_custom(botan_rng_t* rng_out, const char* rng_name, void* context,
+                                                 int(* get_cb)(uint8_t* out, size_t out_len, void* context),
+                                                 int(* add_entropy_cb)(const uint8_t input[], size_t length, void* context));
+
+/**
 * Get random bytes from a random number generator
 * @param rng rng object
 * @param out output buffer of size out_len
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -1,6 +1,7 @@
 /*
 * FFI (C89 API)
 * (C) 2015,2017 Jack Lloyd
+* (C) 2021 René Fischer
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
@@ -201,15 +202,18 @@ typedef struct botan_rng_struct* botan_rng_t;
 BOTAN_PUBLIC_API(2,0) int botan_rng_init(botan_rng_t* rng, const char* rng_type);
 
 /**
-* Initialize a custom random number generator with a set of callbacks
+* Initialize a custom random number generator from a set of callback functions
 * @param rng rng object
 * @param rng_name name of the rng
+* @param context An application-specific context passed to the callback functions
 * @param get_cb Callback for getting random bytes from the rng
-* @param add_entry_cb Callback for adding entropy to the rng
+* @param add_entry_cb Callback for adding entropy to the rng, may be NULL
+* @param destroy_cb Callback called when rng is destroyed, may be NULL
 */
 BOTAN_PUBLIC_API(3,0) int botan_rng_init_custom(botan_rng_t* rng_out, const char* rng_name, void* context,
-                                                 int(* get_cb)(uint8_t* out, size_t out_len, void* context),
-                                                 int(* add_entropy_cb)(const uint8_t input[], size_t length, void* context));
+                                                 int(* get_cb)(void* context, uint8_t* out, size_t out_len),
+                                                 int(* add_entropy_cb)(void* context, const uint8_t input[], size_t length),
+                                                 void(* destroy_cb)(void* context));
 
 /**
 * Get random bytes from a random number generator
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -206,8 +206,8 @@ BOTAN_PUBLIC_API(2,0) int botan_rng_init(botan_rng_t* rng, const char* rng_type)
 * @param rng rng object
 * @param rng_name name of the rng
 * @param context An application-specific context passed to the callback functions
-* @param get_cb Callback for getting random bytes from the rng
-* @param add_entry_cb Callback for adding entropy to the rng, may be NULL
+* @param get_cb Callback for getting random bytes from the rng, return 0 for success
+* @param add_entry_cb Callback for adding entropy to the rng, return 0 for success, may be NULL
 * @param destroy_cb Callback called when rng is destroyed, may be NULL
 */
 BOTAN_PUBLIC_API(3,0) int botan_rng_init_custom(botan_rng_t* rng_out, const char* rng_name, void* context,
--- a/src/lib/ffi/ffi_rng.cpp
+++ b/src/lib/ffi/ffi_rng.cpp
@@ -1,5 +1,6 @@
 /*
 * (C) 2015,2017 Jack Lloyd
+* (C) 2021 René Fischer
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
@@ -74,7 +75,6 @@ return ffi_guard_thunk(__func__,[=]() -> int {
    if(get_cb == nullptr)
       return BOTAN_FFI_ERROR_NULL_POINTER;
 
-   std::unique_ptr<Botan::RandomNumberGenerator> rng;
    class Custom_RNG : public Botan::RandomNumberGenerator
       {
       public:
@@ -100,9 +100,10 @@ return ffi_guard_thunk(__func__,[=]() -> int {
 
          void randomize(uint8_t output[], size_t length) override
          {
-            if(m_get_cb(m_context, output, length))
+            int rc = m_get_cb(m_context, output, length);
+            if(rc)
             {
-               throw Botan::Invalid_State("Failed to get random from C callback");
+               throw Botan::Invalid_State("Failed to get random from C callback, rc=" + rc);
             }
          }
 
@@ -117,10 +118,11 @@ return ffi_guard_thunk(__func__,[=]() -> int {
             {
                return;
             }
-            
-            if(m_add_entropy_cb(m_context, input, length))
+
+            int rc = m_add_entropy_cb(m_context, input, length);
+            if(rc)
             {
-               throw Botan::Invalid_State("Failed to add entropy via C callback");
+               throw Botan::Invalid_State("Failed to add entropy via C callback, rc=" + rc);
             }
          }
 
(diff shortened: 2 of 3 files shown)


securitykernel:
Linked the corresponding [strongswan PR](https://github.com/strongswan/strongswan/pull/192).

Oh and since this is going into 3.0 which is a while away from release, I'd like to backport this to 2.x.

src/lib/ffi/ffi_rng.cpp
```diff
@@ -57,6 +57,93 @@ int botan_rng_init(botan_rng_t* rng_out, const char* rng_type)
...
+         bool accepts_input() const override
+         {
+            return m_cbs.add_entropy != nullptr;
+         }
+
+         void add_entropy(const uint8_t input[], size_t length) override
+         {
+            if(m_cbs.add_entropy == nullptr)
+            {
+               throw Botan::Invalid_Argument("No callback set for add_entropy");
```
randombit, line 99: Conventionally RNG instances which do not accept entropy (eg RDRAND_RNG, or System_RNG in most cases) just ignore any inputs.

src/lib/ffi/ffi_rng.cpp
```diff
@@ -57,6 +57,93 @@ int botan_rng_init(botan_rng_t* rng_out, const char* rng_type)
       });
    }
 
+int botan_rng_init_custom(botan_rng_t* rng_out, const char* rng_name, void* context,
+                          int(* get_cb)(uint8_t* out, size_t out_len, void* context),
+                          int(* add_entropy_cb)(const uint8_t input[], size_t length, void* context))
+{
+return ffi_guard_thunk(__func__,[=]() -> int {
+   if(rng_out == nullptr)
```
randombit, line 65: Should also be checking rng_name and the callback functions. (I suppose `context` can legitimately be NULL, if the functions themselves do not need it eg for calling a HW RNG)

src/lib/ffi/ffi.h
```diff
@@ -200,6 +200,17 @@ typedef struct botan_rng_struct* botan_rng_t;
...
+/**
+* Initialize a custom random number generator with a set of callbacks
+* @param rng rng object
+* @param rng_name name of the rng
+* @param get_cb Callback for getting random bytes from the rng
+* @param add_entry_cb Callback for adding entropy to the rng
+*/
+BOTAN_PUBLIC_API(3,0) int botan_rng_init_custom(botan_rng_t* rng_out, const char* rng_name, void* context,
+                                                 int(* get_cb)(uint8_t* out, size_t out_len, void* context),
+                                                 int(* add_entropy_cb)(const uint8_t input[], size_t length, void* context));
```
randombit, line 212: Let's put `context` always first in the arguments

src/lib/ffi/ffi.h
```diff
@@ -200,6 +200,17 @@ typedef struct botan_rng_struct* botan_rng_t;
 */
 BOTAN_PUBLIC_API(2,0) int botan_rng_init(botan_rng_t* rng, const char* rng_type);
 
+/**
+* Initialize a custom random number generator with a set of callbacks
+* @param rng rng object
+* @param rng_name name of the rng
+* @param get_cb Callback for getting random bytes from the rng
```
randombit, line 207: Document context

src/lib/ffi/ffi.h
```diff
@@ -200,6 +200,17 @@ typedef struct botan_rng_struct* botan_rng_t;
 */
```
randombit, line 200: Should update (C) for this file

src/lib/ffi/ffi_rng.cpp
```diff
@@ -57,6 +57,93 @@ int botan_rng_init(botan_rng_t* rng_out, const char* rng_type)
       });
```
randombit, line 57: Update (C)

src/lib/ffi/ffi_rng.cpp
```diff
@@ -57,6 +57,93 @@ int botan_rng_init(botan_rng_t* rng_out, const char* rng_type)
...
 
+int botan_rng_init_custom(botan_rng_t* rng_out, const char* rng_name, void* context,
+                          int(* get_cb)(uint8_t* out, size_t out_len, void* context),
+                          int(* add_entropy_cb)(const uint8_t input[], size_t length, void* context))
+{
+return ffi_guard_thunk(__func__,[=]() -> int {
+   if(rng_out == nullptr)
+      return BOTAN_FFI_ERROR_NULL_POINTER;
+
+   std::unique_ptr<Botan::RandomNumberGenerator> rng;
```
randombit, line 68: Move this decl to after `Custom_RNG`, then it can just be

`std::unique_ptr<Botan::RandomNumberGenerator> rng(new Custom_RNG(...))`

without the rest

src/lib/ffi/ffi.h
```diff
@@ -200,6 +200,17 @@ typedef struct botan_rng_struct* botan_rng_t;
...
 BOTAN_PUBLIC_API(2,0) int botan_rng_init(botan_rng_t* rng, const char* rng_type);
 
+/**
+* Initialize a custom random number generator with a set of callbacks
+* @param rng rng object
+* @param rng_name name of the rng
+* @param get_cb Callback for getting random bytes from the rng
+* @param add_entry_cb Callback for adding entropy to the rng
```
randombit, line 208: Document that return values should be 0 for success

src/lib/ffi/ffi_rng.cpp
```diff
@@ -57,6 +57,93 @@ int botan_rng_init(botan_rng_t* rng_out, const char* rng_type)
...
+               m_cbs.context = context;
+               m_cbs.get = get_cb;
+               m_cbs.add_entropy = add_entropy_cb;
+            }
+
+         void randomize(uint8_t output[], size_t length) override
+         {
+            if(m_cbs.get(output, length, m_cbs.context))
+            {
+               throw Botan::Invalid_State("Failed to get random from C callback");
```
randombit, line 86: Maybe capture the error return code here?

Currently we throw away the exception messages so this is lost unless you enable the stdout logging, but I have a plan to address that described in https://github.com/randombit/botan-rs/issues/16, I've now used that approach in implementing the C bindings for https://github.com/signalapp/libsignal-client and it works well.

src/lib/ffi/ffi_rng.cpp
```diff
@@ -57,6 +57,93 @@ int botan_rng_init(botan_rng_t* rng_out, const char* rng_type)
...
+         {
+            void* context;
+            int(* get)(uint8_t* out, size_t out_len, void* context);
+            int(* add_entropy)(const uint8_t input[], size_t length, void* context);
+         } m_cbs;
+   };
+
+   rng.reset(new Custom_RNG(rng_name, context, get_cb, add_entropy_cb));
+
+   if(!rng)
```
randombit, line 137: Could this ever fire? Seems like we would have already thrown `bad_alloc`

src/lib/ffi/ffi_rng.cpp
```diff
@@ -57,6 +57,93 @@ int botan_rng_init(botan_rng_t* rng_out, const char* rng_type)
...
+         }
+
+      private:
+         std::string m_name;
+         struct callbacks
+         {
+            void* context;
+            int(* get)(uint8_t* out, size_t out_len, void* context);
+            int(* add_entropy)(const uint8_t input[], size_t length, void* context);
+         } m_cbs;
```
randombit, line 132: I don't feel like this extra struct is adding much vs them being individual members of the class

randombit: I think these can be stored as `std::function` where the context is pre-bound in the right place.

src/lib/ffi/ffi_rng.cpp
```diff
@@ -57,6 +57,93 @@ int botan_rng_init(botan_rng_t* rng_out, const char* rng_type)
...
+
+         std::string name() const override
+         {
+            return m_name;
+         }
+
+         void clear() override
+         {
+            // TODO?
+            // post: is_seeded() == false!
```
randombit, line 116: Given the context it's probably best to assume that the RNG is already seeded or that it is capable of reseeding itself on demand if needed, leaving this as a no-op and `is_seeded` always returning true.

src/tests/test_ffi.cpp
```diff
@@ -247,6 +248,40 @@ class FFI_Unit_Tests final : public Test
             TEST_FFI_OK(botan_rng_add_entropy, (rng, not_really_entropy, 32));
             }
 
+         size_t cb_counter = 0;
+
+         auto custom_get_cb = +[](uint8_t* out, size_t out_len, void* context) -> int
```
randombit, line 253: TIL about `+[]` to force the lambda to be a function pointer, hadn't seen that before.

src/tests/test_ffi.cpp
```diff
@@ -247,6 +248,40 @@ class FFI_Unit_Tests final : public Test
...
+            {
+            BOTAN_UNUSED(out, out_len);
```
randombit, line 255: Maybe write to out with some fixed pattern (0x010203...) and verify the output is really copied from the callback all the way back to the caller.

randombit (commented):
Approach looks good. Some style and testing comments etc but nothing major.

randombit:
Oh and no problem with backporting this.

src/lib/ffi/ffi_rng.cpp
securitykernel: IMHO only `get_cb` should be mandatory, all others may be set to `NULL`, e.g., the strongswan RNG API does not offer adding entropy to the pool.

src/lib/ffi/ffi_rng.cpp
securitykernel: Done

src/tests/test_ffi.cpp
securitykernel: TBH: Found on SO

src/tests/test_ffi.cpp
securitykernel: Done

securitykernel:
Applied all review comments. I also added back the `destroy_cb` I had proposed in #2595. Using it in https://github.com/strongswan/strongswan/pull/192 removed quite a bit of duplicated code. In the strongswan PR, we need to add a fallback for older Botan versions and if that is set, we can merge.

src/lib/ffi/ffi_rng.cpp
```diff
@@ -57,6 +60,101 @@ int botan_rng_init(botan_rng_t* rng_out, const char* rng_type)
...
+               m_destroy_cb(m_context);
+            }
+         }
+
+         void randomize(uint8_t output[], size_t length) override
+         {
+            int rc = m_get_cb(m_context, output, length);
+            if(rc)
+            {
+               throw Botan::Invalid_State("Failed to get random from C callback, rc=" + rc);
```
randombit, line 106: std::to_string(rc)

src/lib/ffi/ffi_rng.cpp
```diff
@@ -57,6 +60,101 @@ int botan_rng_init(botan_rng_t* rng_out, const char* rng_type)
...
+         {
+            if(m_add_entropy_cb == nullptr)
+            {
+               return;
+            }
+
+            int rc = m_add_entropy_cb(m_context, input, length);
+            if(rc)
+            {
+               throw Botan::Invalid_State("Failed to add entropy via C callback, rc=" + rc);
```
randombit, line 125: to_string(rc)

randombit (commented):
Change looks good but one bug (trying to string+int which does not do what one might expect)

securitykernel:
Removed draft status, ready to merge.

PR #2601, opened 2021-01-22
get_auxval implementation for NetBSD.

devnexen:
seemingly did not make a nice wrapper as FreeBSD (yet), this is the most straightforward way in this platform.

### DIFF
--- a/src/build-data/os/netbsd.txt
+++ b/src/build-data/os/netbsd.txt
@@ -14,7 +14,7 @@ sockets
 threads
 thread_local
 filesystem
-_dlauxinfo
+auxinfo
 </target_features>
 
 <feature_macros>
--- a/src/build-data/os/netbsd.txt
+++ b/src/build-data/os/netbsd.txt
@@ -14,6 +14,7 @@ sockets
 threads
 thread_local
 filesystem
+_dlauxinfo
 </target_features>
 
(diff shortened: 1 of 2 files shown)


src/build-data/os/netbsd.txt
```diff
@@ -14,6 +14,7 @@ sockets
 threads
 thread_local
 filesystem
+_dlauxinfo
```
randombit, line 17: The leading underscore here makes the macros weird imo.

Otherwise looks good. Not really well documented (couldn't find a man page for it) but seems to have been introduced some time ago - https://mail-index.netbsd.org/tech-userlevel/2012/02/14/msg006088.html so safe to use by default.

randombit (changes requested):

src/build-data/os/netbsd.txt
devnexen: I know but this is the real call :-) but I can remove it if you insist.

randombit: Yeah but I think of the OS feature macros as helpful shorthands, sometimes that happens to map onto the exact function name that the feature is indicating exists but we also have `posix_mlock`, or things that aren't function calls at all but some other visible feature eg `proc_fs`.

PR #2606, opened 2021-01-26
Add getrandom and BCryptGenRandom to RNG docs

securitykernel:
* ` BCryptGetRandom` was missing in the ` System_RNG`  docs
* ` BCryptGetRandom` and ` getrandom` were missing in the entropy sources docs


randombit:
Thanks

PR #2624, opened 2021-02-15
Fix/Update PasswordHashFamily::from_params docs

### LINKED ISSUE #2576: PasswordHashFamily::from_params documents Argon2 parameters out-of-order
According to the docs,
> Argon2{i,d,id} would use iterations, memory, parallelism for i{1-3}

However, it seems like iterations and memory are switched here. Take this snippet:
```cpp
std::unique_ptr<Botan::PasswordHashFamily> family = Botan::PasswordHashFamily::create("Argon2id");
std::unique_ptr<Botan::PasswordHash> hash = family->from_params(16, 1000000, 1);

std::cout << "Iterations: " << hash->iterations() << std::endl;
std::cout << "Memory: " << hash->memory_param() << std::endl;
```
The docs say that this should output:
```
Iterations: 16
Memory: 1000000
```
but it actually outputs:
```
Iterations: 1000000
Memory: 16
```
This error is present both in the handbook and the internal function docs. Seems like the memory and iterations need to be swapped, whether it be in the docs or the function itself.

It's also worth noting that the docs for this function could be a lot better, honestly. Something like this:
```
Create a password hash using some scheme specific format. Parameters are as follows:
- For PBKDF2, PGP-S2K, and Bcrypt-PBKDF, i1 is iterations
- Scrypt uses N, r, p for i{1-3} (potentially include a link to what these parameters do, such as here: https://blog.filippo.io/the-scrypt-parameters/)
- Argon2 family uses memory (in KB), iterations, and parallelism for i{1-3}
All unneeded parameters should be set to 0 or left blank.
```

binex-dsk:
Fixes/Closes #2576.

Updated to be more clear, fixes Argon2 parameters being out of order.

### DIFF
--- a/doc/api_ref/pbkdf.rst
+++ b/doc/api_ref/pbkdf.rst
@@ -99,13 +99,11 @@ The ``PasswordHashFamily`` creates specific instances of ``PasswordHash``:
    .. cpp:function:: std::unique_ptr<PasswordHash> from_params( \
          size_t i1, size_t i2 = 0, size_t i3 = 0) const
 
-      Create a password hash using some scheme specific format.
-      Eg PBKDF2 and PGP-S2K set iterations in i1
-      Scrypt uses N,r,p in i{1-3}
-      Bcrypt-PBKDF just has iterations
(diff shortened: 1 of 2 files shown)


randombit:
Thank you

PR #2627, opened 2021-02-18
Backport #2600

securitykernel:
Fixing #2595

### DIFF
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -1,6 +1,7 @@
 /*
 * FFI (C89 API)
 * (C) 2015,2017 Jack Lloyd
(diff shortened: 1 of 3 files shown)


randombit:
CI failure is related - looks like MSVC 2015 doesn't understand `+[]` lambda syntax :(

https://ci.appveyor.com/project/randombit/botan/builds/37832648/job/scdaq4g48qfqu8oi#L970

randombit:
[I'd be fine with just disabling the test at compile time on MSVC 2015]

securitykernel:
Removed the `+[]` lambda syntax. Its only needed when the lambda captures, which the tests do not. May be a leftover from coding.

PR #2633, opened 2021-02-22
cli sandbox implementation proposal for mac.

devnexen:
it s being marked as deprecated however still widely used (e.g. openssh)
 as there is no other genuine programatic replacements.

### DIFF
--- a/src/build-data/cc/clang.txt
+++ b/src/build-data/cc/clang.txt
@@ -6,7 +6,7 @@ lang_flags "-std=c++17 -D_REENTRANT"
 
 warning_flags "-Wall -Wextra -Wpedantic -Wshadow -Wstrict-aliasing -Wstrict-overflow=5 -Wcast-align -Wmissing-declarations -Wpointer-arith -Wcast-qual -Wshorten-64-to-32"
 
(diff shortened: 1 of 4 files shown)


src/build-data/cc/clang.txt
```diff
@@ -6,7 +6,7 @@ lang_flags "-std=c++17 -D_REENTRANT"
 
 warning_flags "-Wall -Wextra -Wpedantic -Wshadow -Wstrict-aliasing -Wstrict-overflow=5 -Wcast-align -Wmissing-declarations -Wpointer-arith -Wcast-qual -Wshorten-64-to-32"
 
-werror_flags "-Werror -Wno-error=unused-parameter -Wno-error=unreachable-code -Wno-error=unused-lambda-capture"
+werror_flags "-Werror -Wno-error=unused-parameter -Wno-error=unreachable-code -Wno-error=unused-lambda-capture -Wno-error=deprecated-declarations"
```
randombit, line 9: Don't like adding a no-error until we really have to. Is there some way to disable it just at the invocation point?

devnexen: Surely, but I tried to find an "elegant" solution rather than adding pragma in compiler etc while it is not really hiding hideous bug here just deprecations. but if you insist I ll update that part.

Issue #2638, opened 2021-03-04
Bigint: -1 < -2


guidovranken:
```
#include <botan/bigint.h>

int main(void)
{
    printf("%d\n", Botan::BigInt("-1") < Botan::BigInt("-2"));
    return 0;
}
```

This prints 1. Is this intentional?

randombit:
Not intentional - investigating

randombit:
Bug introduced in 00b6842

randombit:
Fixed on master and release-2 now thanks for the report.

PR #2639, opened 2021-03-04
Fix BigInt::operator< when both integers are negative

randombit:
Broken in 00b6842a54

GH #2638


mouse07410:
I think this problem exists in Botan-2 as well. Fix may need to be backported.

randombit:
Yes bug was introduced in 2.9.0

PR #2641, opened 2021-03-05
Fix BigInt::operator< when both integers are negative

randombit:
Broken in 00b6842a54

GH #2638


randombit:
Merged but GH is confused as I used the wrong base branch on the PR

Issue #2649, opened 2021-03-11
EC_Group - inconsistent behavior when converting from domain parameters to OID


finalglyph:
Hi,

One of my use cases is reading the DER-encoded domain parameters of an elliptic curve from an external source and finding out which "named" curve they correspond to.

In other words, I sometimes need to get the OID of an elliptic curve (that is known by Botan) from its domain parameters.

I found that, while `Botan::EC_Group` **can** do this "DP to OID" conversion, it does not do it ***reliably*** unless a workaround is used.

The following test program demonstrates the specific situations I will describe below. My environment is Arch Linux, with gcc 10.2 and Botan 2.17.3.

```cpp
#include <botan/ec_group.h>
#include <botan/asn1_obj.h> // for Botan::OID
#include <fstream>
#include <vector>
#include <cassert>

void write_buffer_to_file(const std::vector<uint8_t>& buffer, const char *file_name)
{
	std::ofstream stream{file_name, std::ios::binary};
	stream.write(reinterpret_cast<const char *>(buffer.data()), buffer.size());
}

std::vector<uint8_t> read_buffer_from_file(const char *file_name)
{
	std::ifstream stream{file_name, std::ios::binary | std::ios::ate};
	std::vector<uint8_t> buffer(stream.tellg());
	stream.seekg(0);
	stream.read(reinterpret_cast<char *>(buffer.data()), buffer.size());
	return buffer;
}

void write_ec_param_to_file(const Botan::OID& ec_oid, const char *file_name)
{
	Botan::EC_Group ec_group{ec_oid};

	auto ec_param_der = ec_group.DER_encode(Botan::EC_Group_Encoding::EC_DOMPAR_ENC_EXPLICIT);

	write_buffer_to_file(ec_param_der, file_name);
}

Botan::OID read_ec_oid_from_file(const char *file_name)
{
	auto ec_param_der = read_buffer_from_file(file_name);

	Botan::EC_Group ec_group{ec_param_der};

	return ec_group.get_curve_oid();
}

int main(int argc, char *argv[])
{
	write_ec_param_to_file(Botan::OID::from_string("secp256r1"), "secp256r1.der");

	// expected and desirable behavior
	assert(read_ec_oid_from_file("secp256r1.der").to_formatted_string() == "secp256r1");

	// simulate application restart
	Botan::EC_Group::clear_registered_curve_data();

	// unexpected and undesirable behavior
	assert(read_ec_oid_from_file("secp256r1.der").has_value() == false);

[...]

PR #2654, opened 2021-03-12
Lookup OIDs for EC_Groups using the order as a key

randombit:
#2649


src/lib/pubkey/ec_group/ec_group.cpp
```diff
@@ -241,26 +241,38 @@ class EC_Group_Data_Map final
...
+            }
+         else
+            {
+            // Here try to use the order as a hint to look up the group id, to identify common groups
+            const OID oid_from_store = EC_Group::EC_group_identity_from_order(order);
+            if(oid_from_store.has_value())
                {
-               m_registered_curves.push_back(data);
-               return data;
+               std::shared_ptr<EC_Group_Data> data = EC_Group::EC_group_info(oid);
```
finalglyph, line 266: Should be `EC_group_info(oid_from_store)` instead of `EC_group_info(oid)`.

src/lib/pubkey/ec_group/ec_group.cpp
```diff
@@ -241,26 +241,38 @@ class EC_Group_Data_Map final
...
+               if(data == nullptr || new_group->match(*data))
```
finalglyph, line 267: Calling `EC_Group::EC_group_info` with any non-empty OID returned by `EC_Group::EC_group_identity_from_order` must be **guaranteed** by the library to return a non-null `std::shared_ptr<EC_Group_Data>`. So `data` must always be non-null here. This should probably be `assert`ed.

src/tests/test_ec_group.cpp
```diff
@@ -704,13 +704,71 @@ Test::Result test_ec_group_from_params()
...
+
+   const Botan::OID oid("1.3.6.1.4.1.25258.100.0"); // some other random OID
+
+   Botan::EC_Group reg_group(p, a, b, g_x, g_y, order, 1, oid);
+   result.test_success("Registration success");
    result.confirm("Group has correct OID", reg_group.get_curve_oid() == oid);
 
+   // We can now get it by OID:
+   Botan::EC_Group hc_group(oid);
+   result.confirm("Group has correct OID", reg_group.get_curve_oid() == oid);
```
finalglyph, line 765: Should be `hc_group.get_curve_oid()` instead of `reg_group.get_curve_oid()`.

src/lib/pubkey/ec_group/ec_group.cpp
```diff
@@ -241,26 +241,38 @@ class EC_Group_Data_Map final
...
+         std::shared_ptr<EC_Group_Data> new_group =
+            std::make_shared<EC_Group_Data>(p, a, b, g_x, g_y, order, cofactor, oid, source);
 
          if(oid.has_value())
             {
-            // Not located in existing store - try hardcoded data set
             std::shared_ptr<EC_Group_Data> data = EC_Group::EC_group_info(oid);
-
-            if(data)
+            if(data != nullptr && !new_group->match(*data))
```
finalglyph, line 257: `new_group->match_param(*data)` would be more explicit (and still short). Also, the function should take a `const` reference.

src/lib/pubkey/ec_group/ec_group.cpp
```diff
@@ -223,7 +223,7 @@ class EC_Group_Data_Map final
                {
                /*
                * If the same curve was previously created without an OID
-               * but has been registered again using an OID, save that OID.
+               * but is now being registered again using an OID, save that OID.
                */
                if(oid.empty() == false)
                   {
```
finalglyph, line 229: Off-topic, but I believe the `throw Invalid_Argument("Cannot register ECC group ...` below is **unreachable** due to the `continue` above.

src/lib/pubkey/ec_group/ec_group.cpp
randombit: Ouch nice catch thank you

src/lib/pubkey/ec_group/ec_group.cpp
randombit: Not understanding this - we still don't want to deref a nullptr, right? Agree re the arg being `const` ref.

src/lib/pubkey/ec_group/ec_group.cpp
randombit: This logic was pretty contorted overall. I rewrote the whole sequence to avoid any nested conditionals.

src/lib/pubkey/ec_group/ec_group.cpp
finalglyph: My suggestion was to rename function `EC_Group_Data::match` to `EC_Group_Data::match_param`. This would indicate to the reader that a match **by domain parameters** is attempted, which isn't obvious from the function signature alone. I didn't mean the `nullptr` check should be removed. Sorry for the ambiguous phrasing.

src/lib/pubkey/ec_group/ec_group.cpp
finalglyph: OK, I think it's better now.

Issue #2635, opened 2021-02-27
TLS boost::asio::ssl compatible library does not work


fxdupont:
I tried to make a small server tool using the TLS "boost::asio::ssl compatible" library but it fails in handshake with an Unknown error from a "basic_flat_buffer too long" exception.

This looks like a mismatch between Botan and boost::beast so I give here the exact environment:
 - I use a x86 mac running macOS Big Sur 11.1
 - the Xcode version is 12.4 and I use the C++ compiler with the --std=c++11 option 
 - the botan homebrew package is a 2.17.3 where I patched the formula to add --with-boost
 - the boost homebrew package is 1.75.0

I attached the botan and boost formulas (in the case a config flag matters) and the tool which is derived from the boost c++11 sample for the boost::asio::ssl library. I tried the client derived from the same sample directory and also the tls_client from Botan cli: both made the handshake to fail on the server side. I used wireshark to see the packets: there is nothing particular: the client sends a standard (and between 280-290 bytes) TLS ClientHello and the server closes the connection.

I can help if you need more details or to run test codes. The idea of the boost::asio::ssl compatible library was **great** and   I really want to make it to work in the software I work on which offers the choice between Botan and OpenSSL crypto backends. Even I currently is not about security software I worked many times on security related codes: in fact this is something which comes back often: I think I never stay 10 years without it...

Another (very minor) point: the classes derived from boost::system::error_code should have a virtual destructor to not get problems with the clang/gcc -Wnon-virtual-dtor flag. BTW it is clear these classes will be never used through a pointer to the base class but the compiler does not know... Alternately you can disable locally the warning but IMHO this is a dubious practice.

[boostbotan.rb.txt](https://github.com/randombit/botan/files/6054356/boostbotan.rb.txt)
[boost.rb.txt](https://github.com/randombit/botan/files/6054358/boost.rb.txt)

[botan_boost_sample_server.cpp.txt](https://github.com/randombit/botan/files/6054366/botan_boost_sample_server.cpp.txt)

randombit:
@hrantzsch @reneme any ideas?

hrantzsch:
Strange! I'm not sure if the homebrew packages are really the issue here. To be certain, can you try compiling this example of an HTTP server with those same versions of Botan and Boost?
https://github.com/hrantzsch/botan-tls-stream-server-example

fxdupont:
[...]

PR #2655, opened 2021-03-12
fix: TLS::Stream broke on being std::move'd

hrantzsch:
As reported in #2635, `Botan::TLS::Stream` did not handle being `std::move`d well. The reason was the nested `StreamCore` which held a reference to the Stream itself.

 * Add a regression test
 * Avoid breaking Stream reference within StreamCore by inverting dependencies

### DIFF
--- a/src/lib/tls/asio/asio_stream.h
+++ b/src/lib/tls/asio/asio_stream.h
@@ -63,8 +63,7 @@ class Stream
       explicit Stream(Context& context, Args&& ... args)
          : m_context(context)
          , m_nextLayer(std::forward<Args>(args)...)
-         , m_core(*this)
-         , m_shutdown_received(false)
+         , m_core(context)
          , m_input_buffer_space(MAX_CIPHERTEXT_SIZE, '\0')
          , m_input_buffer(m_input_buffer_space.data(), m_input_buffer_space.size())
          {}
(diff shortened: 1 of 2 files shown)


hrantzsch:
@randombit should I open another PR to backport to Botan 2?

src/lib/tls/asio/asio_stream.h
```diff
@@ -627,25 +627,29 @@ class Stream
...
                   }
                else
                   {
                   Callbacks::tls_verify_cert_chain(cert_chain, ocsp_responses, trusted_roots, usage, hostname, policy);
                   }
                }
 
+            bool shutdown_received;
+            boost::beast::flat_buffer receive_buffer;
+            boost::beast::flat_buffer send_buffer;
```
reneme, line 642: should those be prefixed with `m_`?

src/tests/test_tls_stream_integration.cpp
```diff
@@ -196,7 +202,14 @@ class Server : public Side, public std::enable_shared_from_this<Server>
 
          // Note: If this was a real server, we should create a new session (with its own stream) for each accepted
          // connection. In this test we only have one connection.
-         m_stream = std::unique_ptr<ssl_stream>(new ssl_stream(std::move(socket), m_ctx));
+
+         if (m_move_before_accept) {
+            // regression for #2635
```
reneme, line 207: regression _test_ for ...

src/tests/test_tls_stream_integration.cpp
```diff
@@ -286,6 +299,9 @@ class Server : public Side, public std::enable_shared_from_this<Server>
       tcp::acceptor m_acceptor;
       Result_Wrapper m_result;
       bool m_short_read_expected;
+
+      // regression for #2635
```
reneme, line 303: see above

reneme (approved):
LGTM despite minor nits.

src/lib/tls/asio/asio_stream.h
hrantzsch: I removed the prefix because they are public in the StreamCore. StreamCore is a nested class that is only accessible to Stream though.

randombit:
Thanks for chasing this down, much appreciated.

Issue #2498, opened 2020-11-17
The dTLS server (1.2) is not handling properly the re-transmissions of flight 5 from the client


pist-eb:
Problem:
The retransmissions of Flight 5 made by the dTLS client are not well handled and end us with Alert.

Prerequisities:
•	Botan cli as a dTLS server
•	Openssl as a dTLS client
•	Delay introduces in the dTLS server (on receiving the Client Key Exchange) to trigger the dTLS retransmissions by the client

Reproduction:
•	Running Botan cli server, i.e.:
botan tls_server my_cert.pem  my_key.pem --port=8080 --type=udp
•	Running openssl client, i.e.:
openssl s_client -connect 127.0.0.1:8080 -debug -msg -dtls1_2 -CAfile my_cert.pem
•	Sending any data through the client

Current behavior:
•	dTLS session not established correctly, even if the client thinks that everything went fine
•	Application Data received triggers an alert on the server side (*)

Expected behavior:
•	dTLS session successfully established, despite of the retransmissions of the flight 5 messages (Client Key Exchange, Change Cipher Spec, Finished),
•	Application Data received and decrypted on the server side

More details / output from test:

Server:
```
botan tls_server my_cert.pem  my_key.pem --port=8080 --type=udp
Listening for new connections on udp port 8080
Handshake complete, DTLS v1.2 using ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256
Session ID 5FB2367026312B133DC67909107C3805729C841AD509ACE077E1A56319CDA548
Connection problem: Can't interleave application and handshake data (*)
```

Client:
```
openssl s_client -connect 127.0.0.1:8080 -debug -msg -dtls1_2 -CAfile my_cert.pem 
CONNECTED(00000003)
...
New, TLSv1.2, Cipher is ECDHE-ECDSA-CHACHA20-POLY1305
Server public key is 256 bit
Secure Renegotiation IS supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : DTLSv1.2
    Cipher    : ECDHE-ECDSA-CHACHA20-POLY1305
    Session-ID: 5FB2367026312B133DC67909107C3805729C841AD509ACE077E1A56319CDA548
    Session-ID-ctx: 
    Master-Key: ...
    PSK identity: None
    PSK identity hint: None
    SRP username: None
    Start Time: 1605514864
    Timeout   : 7200 (sec)
[...]

PR #2657, opened 2021-03-12
Retransmission of Flight 5 fixed in dTLS

pstanisz:
This change shall fix the issue: https://github.com/randombit/botan/issues/2498

In case of any delay during the dTLS handshake and retransmission of the Flight 5 messages by the client, the server might receive Flight 5 messages few times. First time it finishes the handshake, for next (retransmitted) messages it re-creates the pending_state. This prevents from further communication, even if the dTLS connection is well established.

### DIFF
--- a/src/bogo_shim/config.json
+++ b/src/bogo_shim/config.json
@@ -134,7 +134,7 @@
         "SRTP-Server-IgnoreMKI-*": "Non-empty MKI is rejected (bug)",
 
         "Renegotiate-Client-Packed": "Packing HelloRequest with Finished loses the HelloRequest (bug)",
-        "SendHalfHelloRequest*PackHandshake": "Packing HelloRequest with Finished loses the HelloRequest (bug)",
+        "SendHalfHelloRequest*": "Packing HelloRequest with Finished loses the HelloRequest (bug), DTLS fix for retransmissions breaks this test",
 
         "PartialClientFinishedWithClientHello": "Need to check for buffered messages when CCS (bug)",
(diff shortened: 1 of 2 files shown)


randombit:
Thank you. I'll review this shortly. The CI failure looks to be related to this change - the `SendHalfHelloRequest` tests from BoringSSL's suite are now failing.

randombit:
I think it's OK to just disable `SendHalfHelloRequest*` tests (in `src/bogo_shim/config.json`); this test is verifying a behavior that is not spec mandated but just how BoringSSL happens to behave.

It would be nice though if there was some test of the reflight behavior though. We can get though via the BoringSSL suite but this requires building timeout support for DTLS, right now all of the DTLS retransmit tests are disabled for this reason.

randombit (approved):
Change is fine please update the BoGo configuration to disable the failing test.

pstanisz:
Thank you for comments @randombit. Currently the fix is just verified manually with Botan & openssl, hope that at some point it would be possible to extend the DTLS testing with Bogo.

pstanisz:
Not good, another test failed: DTLS-SendExtraFinished*. I'll try to look what is happening, maybe the fix causes some other issue(s)

nametoolong:
IMHO deduplication should be done in `Handshake_Reassembly` - otherwise we have to add a new code path for each type of message which is hard to maintain. https://github.com/randombit/botan/blob/master/src/lib/tls/tls_handshake_io.cpp#L245 might be relevent.

oviano:
Just wondering about this as I have seen this too?

I can see there was a proposed fix but judging from it failing the tests it wasn't propagated to master.

Issue #2648, opened 2021-03-11
TLS::Stream only works with callbacks as completion tokens


finalglyph:
Hi,

First of all, thank you for developing this library. It certainly made my work easier.

I wrote a TLS server using boost.asio, boost.coroutine and `Botan::TLS::Stream`. Unfortunately, during development, I discovered that `Botan::TLS::Stream` does not work with coroutines.

Actually, the problem is that `Botan::TLS::Stream` does not correctly participate in boost.asio's ***extensible asynchronous model*** and, as a result, only supports callback functions as completion tokens.

While investigating this issue, I found C++ standard proposal [N4045](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4045.pdf), which does a great job of explaining the extensible asynchronous model (in chapters 7, 8 and 9).

The following minimal TLS (echo) server using boost stackful coroutines does not compile (Arch Linux, gcc 10.2, C++17, boost 1.75, Botan 2.17.3).

```cpp
#include <botan/credentials_manager.h>
#include <botan/x509cert.h> // for Botan::X509_Certificate
#include <botan/pk_keys.h> // for Botan::Private_Key
#include <botan/data_src.h> // for Botan::DataSource_Stream
#include <botan/pkcs8.h> // for Botan::PKCS8::load_key
#include <string>
#include <vector>
#include <memory> // for std::unique_ptr

class credentials_manager : public Botan::Credentials_Manager
{
public:

	credentials_manager(const std::string& cert_chain_file, const std::string& private_key_file)
	{
		for (Botan::DataSource_Stream stream{cert_chain_file}; !stream.end_of_data(); /**/)
		{
			m_cert_chain.emplace_back(stream);
		}

		if (Botan::DataSource_Stream stream{private_key_file}; !stream.end_of_data())
		{
			m_private_key = Botan::PKCS8::load_key(stream);
		}
	}

private:

	std::vector<Botan::X509_Certificate> find_cert_chain(const std::vector<std::string>& cert_key_types, const std::vector<Botan::X509_DN>& acceptable_cas, const std::string& type, const std::string& context) override
	{
		if (!std::count(cert_key_types.begin(), cert_key_types.end(), m_private_key->algo_name()))
		{
			return {};
		}

		return m_cert_chain;
	}

	Botan::Private_Key *private_key_for(const Botan::X509_Certificate& cert, const std::string& type, const std::string& context) override
	{
		if (cert != m_cert_chain.front())
		{
[...]

PR #2667, opened 2021-03-22
TLS::Stream compatibility with the Boost.Asio extensible model

hrantzsch:
Improvements and fixes suggested in #2648.

- TLS::Stream async functions now also support completion tokens other than callback functions
- TLS::Stream::async_shutdown's internal handler wrapper now correctly forwards the executor
- add virtual d'tors to boost error_category subclasses

### DIFF
--- a/src/lib/tls/asio/asio_stream.h
+++ b/src/lib/tls/asio/asio_stream.h
@@ -332,28 +332,36 @@ class Stream
 
    private:
       /**
-       * @brief Internal wrapper type to adapt the expected signature of `async_shutdown`
-       *        to the completion handler signature of `AsyncWriteOperation`.
+       * @brief Internal wrapper type to adapt the expected signature of `async_shutdown` to the completion handler
+       *        signature of `AsyncWriteOperation`.
        *
-       * This is boilerplate to ignore the `size_t` parameter that is passed to the
-       * completion handler of `AsyncWriteOperation`.
-       *
-       * @todo in C++14 and above this could be implemented as a mutable lambda expression
-       *       that captures `handler` by perfect forwarding, like so:
-       *
-       *       [h = std::forward<Handler>(handler)] (...) mutable { return h(ec); }
+       * This is boilerplate to ignore the `size_t` parameter that is passed to the completion handler of
+       * `AsyncWriteOperation`. Note that it needs to retain the wrapped handler's executor.
        */
-      template <typename Handler>
-      class Wrapper
+      template <typename Handler, typename Executor>
+      struct Wrapper
          {
-         public:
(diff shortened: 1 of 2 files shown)


src/lib/tls/asio/asio_stream.h
```diff
@@ -539,18 +548,23 @@ class Stream
        *                the handler must be:
```
hrantzsch, line 548: Fix parameter name in comment

src/lib/tls/asio/asio_stream.h
```diff
@@ -502,32 +508,35 @@ class Stream
...
+             {
+                using completion_handler_t = std::decay_t<decltype(completion_handler)>;
+
+                BOOST_ASIO_WRITE_HANDLER_CHECK(completion_handler_t, completion_handler) type_check;
+
+                boost::system::error_code ec;
+                tls_encrypt(bufs, ec);
+
+                if (ec)
+                {
```
hrantzsch, line 525: run formatter

Tymolc (approved):
LGTM 👍

hrantzsch:
Thank you again @finalglyph. I can't add you as a reviewer as it seems, but I'd be happy if you could give it another skim.

By the way, I mostly applied the changes you had already explained so nicely in the issue. If you like, feel free to directly open a PR in the future.

hrantzsch:
@randombit none of the appveyor jobs on master runs with `--with-boost`, do they? From the [backport PR](https://github.com/randombit/botan/pull/2673) it seems Visual Studio is not happy about the changes, so I'll look into that again.

randombit:
No the Boost builds were dropped when all old MSVC builds were, but that leaves a big hole in CI. I'll look at re-adding MSVC 2019 Boost build.

randombit:
#2674 added back a Boost MSVC build

finalglyph:
@hrantzsch, thank you for reading my lengthy explanations and preparing this pull request. The code changes look OK to me. (Actually, making the destructors of those `error_category` classes `virtual` seems unnecessary, but I suppose there are other reasons behind this.)

My server, which uses that `tls_stream_replacement` workaround I showed in #2648, hasn't had any issues (e.g. crashes) since I made the "executor forwarding" fix two weeks ago. This pull request basically "merges" the (latest) `tls_stream_replacement` code from #2648 into `Botan::TLS::Stream`, so I'm pretty confident it will fully solve #2648. Please close that issue once this pull request is merged.

Thanks again for these fixes! `Botan::TLS::Stream` now properly makes use of boost.asio's "extensible asynchronous model", so its interface is now as flexible as that of `boost::asio::ssl::stream`.

PR #2686, opened 2021-04-03
src/build-data/arch/riscv64.txt: set wordsize to 64

ffontaine:


randombit:
Wow that probably had a significant perf impact, thanks for noticing this!

PR #2730, opened 2021-04-26
Fix #2237 forcing OID::to_string to use C locale, avoiding thousand s…

### LINKED ISSUE #2237: TLS Server
Hello All,

I'm using Botan::TLS_Server on top of standalone ASIO.
I'm using a self-signed certificate as CA and another one as server certificate (signed with that self-signed one), both made with botan command line using ECDSA algorithm; session manager is Session_Manager_Noop, rng is AutoSeeded_RNG  and the policy is Strict_Policy.

I'm using version 2.11, and everything works fine on Windows and ArchLinux.

Now my question/problem:
If I upgrade botan on my ArchLinux, going to version 2.12.4 or when I built it on an Ubuntu 18.04 box (botan 2.4.0), I get an "alert TLS handshake failure." on the client side.

At the moment I have no idea what is wrong, or what else to look for.
If I use openssl as a client, I get the following:

```
openssl s_client -enable_pha -tls1_2 -state -msg -showcerts -connect localhost:8989 -CAfile /tmp/cert/tempcerts.pem

CONNECTED(00000003)
SSL_connect:before SSL initialization
>>> ??? [length 0005]
    16 03 01 00 db
>>> TLS 1.2, Handshake [length 00db], ClientHello
    01 00 00 d7 03 03 04 78 82 59 cd 66 a7 fe 49 17
    9c a6 cb b8 22 80 c9 d9 d3 cd f2 21 64 3d 8e 8b
    cd cd aa 5a 75 1a 00 00 38 c0 2c c0 30 00 9f cc
    a9 cc a8 cc aa c0 2b c0 2f 00 9e c0 24 c0 28 00
    6b c0 23 c0 27 00 67 c0 0a c0 14 00 39 c0 09 c0
    13 00 33 00 9d 00 9c 00 3d 00 3c 00 35 00 2f 00
    ff 01 00 00 76 00 00 00 1a 00 18 00 00 15 61 74
    65 6c 69 65 72 64 61 73 61 75 64 65 2e 65 73 70
    2e 62 72 00 0b 00 04 03 00 01 02 00 0a 00 0c 00
    0a 00 1d 00 17 00 1e 00 19 00 18 00 23 00 00 00
    16 00 00 00 17 00 00 00 0d 00 30 00 2e 04 03 05
    03 06 03 08 07 08 08 08 09 08 0a 08 0b 08 04 08
    05 08 06 04 01 05 01 06 01 03 03 02 03 03 01 02
    01 03 02 02 02 04 02 05 02 06 02
SSL_connect:SSLv3/TLS write client hello
<<< ??? [length 0005]
    15 03 03 00 02
<<< TLS 1.2, Alert [length 0002], fatal handshake_failure
    02 28
SSL3 alert read:fatal:handshake failure
SSL_connect:error in error
140040023487744:error:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure:ssl/record/rec_layer_s3.c:1543:SSL alert number 40
---
no peer certificate available
---
No client certificate CA names sent
---
SSL handshake has read 7 bytes and written 224 bytes
Verification: OK
---
New, (NONE), Cipher is (NONE)
Secure Renegotiation IS NOT supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : 0000
    Session-ID: 
    Session-ID-ctx: 
    Master-Key: 
    PSK identity: None
    PSK identity hint: None
    SRP username: None
    Start Time: 1578354362
    Timeout   : 7200 (sec)
    Verify return code: 0 (ok)
    Extended master secret: no
---
```
Am I missing some initialization/configuring for other versions of botan?

leobackes:
…eparators from the process locale.

### DIFF
--- a/src/lib/asn1/asn1_oid.cpp
+++ b/src/lib/asn1/asn1_oid.cpp
@@ -98,6 +98,7 @@ OID::OID(const std::string& oid_str)
 std::string OID::to_string() const
    {
    std::ostringstream oss;
+   oss.imbue(std::locale("C"));
    for(size_t i = 0; i != m_id.size(); ++i)
       {
       oss << m_id[i];


randombit:
Yikes. I had no idea that locale would affect ostringstream integer formatting. This probably implies that every use of ostringstream within the library is buggy.

leobackes:
@randombit  Yes, iirc all streams from stdlib (ofstream, stringstream, etc...) are all locale-aware, and they will use the locale set either by the environment (in this case LC_NUMERIC), by the program itself (std::setlocale, std::locale::global), or specifically on the stream with .imbue();

I assumed you chose to use the stream version due some reason, but the fact that this code should not be locale-aware, I think std::to_string does the job just as well, and I have the impression that it will do so with a smaller footprint.

For the other uses, it will be buggy if it is used internally, for the user facing messages, I think it's fine to be locale-aware.

randombit:
Per https://en.cppreference.com/w/cpp/string/basic_string/to_string `to_string` also uses locale and cannot be relied on here. :/

leobackes:
Oh, I didn't see that, so it is also locale-aware.
The weird part is that std::to_string and the streams are doing different things when they should be doing the same, as the to_string didn't use the thousand separator, while the stream did.
So, yeah, stream and imbue() seem to be the way for these cases.

randombit:
I just checked and both libc++ and libstdc++ implement `to_string` in a way that ignores locale. But the C++17 spec requires that `to_string` be implemented as if it was generated using `sprintf` which does respect locale. I think for reliability we need to avoid integer formatting using either ostringstream or `to_string`, at least for cases where locale should not be applied. :/ Thanks for digging into this issue.

leobackes:
Well, if the standard will require std::to_string to be locale-aware, then the sstream with imbued C locale should fix it consistently.
Despite int formatting should be easy to implement, seems like going out of the way for doing something stdlib should do out of the box.

As for digging into it, no worry, this is an awesome library, with code really easy to understand and navigate. Thank you :-)

PR #2740, opened 2021-05-11
add certstore

winterland1989:
For our usage, a cert store with O(log(n)) lookup is better, so this patch bring the flat file cert store to FFI.

### DIFF
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -1628,6 +1628,27 @@ BOTAN_PUBLIC_API(2,13) int botan_x509_cert_verify_with_crl(
    const char* hostname,
    uint64_t reference_time);
 
+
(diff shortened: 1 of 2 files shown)


randombit:
Change looks fine. CI failure is related - the include should just be `botan/certstor_flatefile.h`

winterland1989:
OK, I'll fix it

randombit:
Still some CI failures. Also can you add tests? These could be either in C++ in `test_ffi.cpp` or done by exposing these to the Python interfaces and then adding the tests to `src/scripts/test_python.py` (or, ideally, both)

winterland1989:
I'm afraid I'm not familiar with C++ or python enough to do that, the CI failure also looks strange to me.

randombit:
No need to close this - the change is fine. But if you are not able to add tests that merging will have to wait until I have the time to do so.

Issue #2754, opened 2021-05-31
RFC-compliance


mmaehren:
Hi,
we (@jurajsomorovsky @ic0ns @mmaehren @XoMEX @Kavakuo) are performing an analysis of the RFC-compliance of open-source TLS implementations. Below we list our findings for this implementation. We admit that some are rather nit-picky, but we added them for the sake of completeness. We tried to keep the descriptions brief and didn’t want to spam the issues section so feel free to split up the list into individual issues as you see fit.
If you disagree with our interpretation of certain RFC statements, please leave feedback as we’re interested in your view.

Our results apply to the default configuration of version ***2.17.3***. We used the example implementations for client and server.

[S] = Applies to server
[C] = Applies to client
[C+S] = Applies to both

## Session not aborted

- [C] when a server chooses a non-CBC cipher suite but negotiates encrypt-then-MAC
    - RFC 7366 - 3. Applying Encrypt-then-MAC
        > If a server receives an encrypt-then-MAC request extension from a client and then selects a stream or Authenticated Encryption with Associated Data (AEAD) ciphersuite, it MUST NOT send an encrypt-then-MAC response extension back to the client.
- [C] when a server omits the ChangeCipherSpec message and directly sends an encrypted Finished message, Botan client does not reject the received record immediately based on the record content type handshake (instead of change cipher spec)
    - For ChaCha20_Poly1305 ciphertexts, the bytes of the ciphertext are parsed and likely result in a seemingly long handshake message. If the given Record does not contain enough bytes, Botan waits for additional records until enough bytes have been received before rejecting the message as unexpected.
    - The record can be rejected immediately based on the record content type or as soon as the handshake message type can be parsed (the latter can also be used if a server is expected to send a NewSessionTicket message)

- [S] upon receiving a Padding extension that contains bytes other than 0x00
    - RFC 7685 - 3. Padding Extension
        > The client MUST fill the padding extension completely with zero bytes, although the padding extension_data field may be empty.

- [S] upon receiving a SupportedPointFormat extension that only accepts compressed points or an invalid format
    - 8422 -  5.1. Client Hello Extensions
        > If the client sends the extension and the extension does not contain the uncompressed point format, and the client has used the Supported Groups extension to indicate support for any of the curves defined in this specification, then the server MUST abort the handshake and return an illegal_parameter alert.

- [S] upon receiving a ClientHello that contains elliptic curve extensions but no ECC cipher suite
    - 8422 -  4. TLS Extensions for ECC
[...]

PR #2755, opened 2021-06-01
Check the TLS record type immediately on decoding

randombit:
We would have rejected this later on anyway but there is no reason not to immediately close.

GH #2754


src/lib/tls/tls_record.cpp
```diff
@@ -340,10 +340,13 @@ Record_Header read_tls_record(secure_vector<uint8_t>& readbuf,
...
-   This is a little hacky but given how TLS 1.3 versioning works it
-   is probably safe
+   Verify that the record type and record version are within some expected
+   range, so we can quickly reject totally invalid packets.
+
+   The version check is a little hacky but given how TLS 1.3 versioning works
+   this is probably safe
    */
-   if(readbuf[1] != 3)
+   if(readbuf[0] < 20 || readbuf[0] > 23 || readbuf[1] != 3 || readbuf[2] >= 4)
```
pstanisz, line 349: Content type check will be different in TLS 1.3, where there are additional allowed values: invalid(0) and heartbeat(24). But it will anyway need separate handling/implementation probably.

pstanisz (approved):
This check might be valuable to give some valuable information when crap data is received instead of TLS record.

PR #2756, opened 2021-06-07
FIX: export RSA pubkey via PKCS #11 w/o accessing sensitive data

reneme:
Requesting `p` and `q` via PKCS#11 is likely to fail as those parameters are sensitive key information and usually not extractable from a smart card. Instead, one can fetch `CKA_MODULUS`. Note that this is also done already in [the key generation constructor of `PKCS11_RSA_PrivateKey`](https://github.com/randombit/botan/blob/edd94567b92aadfca79a9117cfb649cb024fc567/src/lib/prov/pkcs11/p11_rsa.cpp#L89).

This was introduced with #2520.


randombit:
Looking back I have no idea why I did it that way. Thanks for the patch.

PR #2771, opened 2021-07-03
tls_server add linux for marking the socket

devnexen:


src/cli/tls_server.cpp
```diff
@@ -16,7 +16,14 @@
...
-#if defined(SO_USER_COOKIE) || defined(SO_RTABLE)
+#if defined(SO_MARK) || defined(SO_USER_COOKIE) || defined(SO_RTABLE)
+#if defined(SO_MARK)
+#define BOTAN_SO_SOCKETID SO_MARK
+#elif defined(SO_USER_COOKIE)
+#define BOTAN_SO_SOCKETID SO_USER_COOKIE
+#else
+#define BOTAN_SO_SOCKETID SO_RTABLE
+#endif
 #define SOCKET_ID 1
```
randombit, line 27: Can we remove the definition of `SOCKET_ID` and just use the existence of `BOTAN_SO_SOCKETID` to decided if we should call `setsockopt` or not?

randombit (commented):
Question / possible change request

src/cli/tls_server.cpp
devnexen: valid point

randombit: Looks good, thanks!

PR #2774, opened 2021-07-05
Curve25519 signatures made compatible with 2.18.1 release

rafalsk:
I have updated my signatures on Curve 25519 to be compatible with Botan 2.18.1 Release.

Wanted to make them compatible with 3.0 alpha straight away but from the first looks on it, additional work is needed.

Description of changes and rationale: Curve25519 is a Montgomery curve over a prime field, while Ed25519 is a Twisted Edwards curve. Further,  Ed25519 is birationally equivalent to Curve25519. Here, the representations are converted on-the-fly (montgomery x-coordinate  is translated into edwards y during signature verification). The sign-information is kept within the resulting signature, which would otherwise be lost.

When making signature the steps are as follows:

- the Curve25519 private keys is converted to an Ed25519 public key through a scalar multiplication (ge_scalarmult_base)
- a sign bit is retrieved  from the Edward's point representation
- an Ed25519 signature is performed
-  the sign bit is encoded into signature (in an unused high bit)

Once I have more time in 1-2 months I can write more detailed description, for now also upgrade to 3.0+ for now here it is.

The initial pull-request to the master branch is here https://github.com/randombit/botan/pull/1239

It's been all tested.

### DIFF
--- a/news.rst
+++ b/news.rst
@@ -1,6 +1,7 @@
 Release Notes
 ========================================
 
+<<<<<<< HEAD
 Version 2.18.1, 2021-05-09
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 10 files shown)


rafalsk:
@randombit let me know if any work needed here on my part. We've implemented this in JavaScript as well and keep using all time long.

Issue #2779, opened 2021-07-09
Windows: System Certificate Store might fail to find certs by their public key fingerprint


reneme:
While writing tests for our application, I noticed that Botan's `System_Certificate_Store` fails to find certain certificates by their public key SHA-1 ([`Certificate_Store_Windows::find_cert_by_pubkey_sha1`](https://github.com/randombit/botan/blob/master/src/lib/x509/certstor_system_windows/certstor_windows.cpp#L231)) on Windows. For instance `D-TRUST Root Class 3 CA 2 2009` (below) is such a certificate. 

The reason: This certificate does not adhere to [RFC 3280's suggestion](https://datatracker.ietf.org/doc/html/rfc3280#section-4.2.1.2) to calculate the "subject key identifier" as a SHA-1 *without* prepending the algorithm identifier in the BER encoding. Hence, the "subject key identifier" differs from the canonical "public key SHA-1" for this particular certificate. In my understanding, that is not the usual case, though.

```
# D-TRUST Root Class 3 CA 2 2009
Subject Key Identifier: fdda14c49f30de21bd1e4239fcab632349e0f184
Public Key SHA-1:       a737b46280e401211faff74eeccd1c05eb8947ce

# Counter Example: DigiCert Global Root G2
Subject Key Identifier: 4e2254201895e6e36ee60ffafab912ed06178f39
Public Key SHA-1:       4e2254201895e6e36ee60ffafab912ed06178f39
```

Unfortunately, [the Windows API](https://docs.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certfindcertificateinstore) does not allow to search for a SHA1 of a certificate's public key directly (it only supports MD5 -- via `CERT_FIND_PUBKEY_MD5_HASH`). Instead, Botan must rely on the convention described above to implement finding certificates via their SHA1 public key fingerprint and actually filter for the "subject key identifier" (via `CERT_FIND_KEY_IDENTIFIER`). Hence, certificates like `D-TRUST Root Class 3 CA 2 2009` will not be found via `::find_cert_by_pubkey_sha1` on Windows despite the certificate being in the trust store. For macOS and Linux this issue doesn't appear.

I am opening this ticket mainly for documentation purposes, in case somebody else stumbles over this. Nevertheless, any suggestion how to tackle this would be appreaciated. It remains to be tested, how many root CA certificates suffer from this issue on a commodity desktop operating system installation.

```
-----BEGIN CERTIFICATE-----
[...]

PR #2780, opened 2021-07-12
FIX: Certificate Store might not find certificates where SHA1(pubkey) != key_id

reneme:
This fixes [GH #2779](https://github.com/randombit/botan/issues/2779). It implements a fallback exhaustive search in `Certificate_Store_Windows::find_cert_by_pubkey_sha1()` that enumerates all available certificates, hashes their public keys and compares them to the requested key hash. Unfortunately, the Windows API doesn't seem to provide a built-in option to query for SHA-1(pubkey). It only provides MD5.

The fallback is invoked only if the existing query comes back with empty hands. The first query result is cached (both positive and negative outcome). This is meant to speed up repetitive searches of the same key hash. That seems particularly important for the negative case, where repetitive queries would otherwise always pay the price of the O(n)-fallback.

Furthermore, I added a regression test that now successfully finds `D-TRUST Root Class 3 CA 2 2009` which wouldn't be found without the proposed fix. On my machine, this exhaustive search fallback test runs for less than 10 milliseconds, FWIW.

I'm happy to provide a backport of that fix to the Botan2 branch if you find it worth the effort.

### DIFF
--- a/src/lib/x509/certstor_system_windows/certstor_windows.cpp
+++ b/src/lib/x509/certstor_system_windows/certstor_windows.cpp
@@ -7,11 +7,13 @@
 */
 
 #include <botan/certstor_windows.h>
+#include <botan/hash.h>
 #include <botan/pkix_types.h>
 #include <functional>
 
 #include <array>
 #include <vector>
+#include <map>
 
 #define NOMINMAX 1
 #define _WINSOCKAPI_ // stop windows.h including winsock.h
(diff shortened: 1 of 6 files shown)


src/tests/test_certstor_system.cpp
```diff
@@ -49,6 +49,37 @@ Test::Result find_certificate_by_pubkey_sha1(Botan::Certificate_Store& certstore
...
 
+Test::Result find_certificate_by_pubkey_sha1_with_unmatching_key_id(Botan::Certificate_Store& certstore)
+   {
+   Test::Result result("System Certificate Store - Find Certificate by SHA1(pubkey) - regression test for GH #2779");
+
+   if(!certstore.find_cert(get_dn_of_cert_with_different_key_id(), {}).has_value())
+      {
+      result.note_missing("OS does not trust the certificate used for this regression test, skipping");
+      return result;
+      }
```
reneme, line 60: This was required, because AppVeyor doesn't come with any of the "vulnerable" certificates. Hence, on AppVeyor the regression test won't run. The MinGW stage provides `SecureTrust CA`, though.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -185,6 +188,52 @@ std::vector<X509_Certificate> find_cert_by_dn_and_key_id(const Botan::X509_DN& s
...
+ *  that is not equal to the SHA-1 of the public key (w/o algorithm identifier)
+ *
+ *  This method lazily builds a cache of certificates found in previous queries as well
+ *  as negative results for @p key_hash queries that didn't find a certificate.
+ *
+ *  See here for further details: https://github.com/randombit/botan/issues/2779
+ */
+std::optional<X509_Certificate> find_cert_by_pubkey_sha1_via_exhaustive_search(const std::vector<uint8_t> &key_hash)
+   {
+   static std::map<std::vector<uint8_t>, std::optional<X509_Certificate>> non_rfc3289_certs;
```
randombit, line 202: It seems like there could be concurrency issues here, if one created multiple certificate stores and performed queries in different threads. Should this instead be a member variable?

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -185,6 +188,52 @@ std::vector<X509_Certificate> find_cert_by_dn_and_key_id(const Botan::X509_DN& s
...
+
+   if(const auto cache_hit = non_rfc3289_certs.find(key_hash); cache_hit != non_rfc3289_certs.end())
+      {
+      return cache_hit->second;
+      }
+
+   auto sha1 = HashFunction::create("SHA-1");
```
randombit, line 209: Should use `create_or_throw` here - it shouldn't ever fail especially as you added SHA-1 as an explicit dependency, but if it did it would return null and then we would null-deref instead of throwing.

randombit (commented):
Looks good, thanks for taking this one. One very minor comment re `create_or_throw` and some concern about the static variable, I think that should be a member var to avoid concurrency sadness.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
reneme: True. My intention was to share the cache with multiple instances of the certificate store to accomodate people that re-instanciate it for every lookup. Nevertheless, the potential race is definitely a fair point.

I changed it to a member variable, with one caveat: All methods of this class are `const`-qualified: I.e. I was forced to qualify the member variable as `mutable`. This might cause usage-dependent threading issues as well, as a user might feel safe to use a `const`-only class without locking.

Would you go for a mutex in that situation, to play it safe?

reneme:
What happened in the coverage build? It didn't hit the error handling in the test case and hence reports a failure because only 80-ish% of the diff was hit? :(

randombit:
Sorry about the delay here. This looks good. I don't think we need a mutex - generally speaking even with `const` qualified functions we do not guarantee that concurrent access to a shared object is safe without some external locking (excepting some types where we explicitly document and guarantee this, like `System_RNG`). We could consider adding one, if it turns out sharing a certstore among threads makes sense / is common.

PR #2781, opened 2021-07-12
Improve Performance of Certificate_Store_Windows::all_subjects()

reneme:
While researching for solutions for [GH #2779](https://github.com/randombit/botan/issues/2779) I stumbled across the fact that the Windows API provides BER encoded buffers for some common certificate constituents. This provides an opportunity to improve the performance of `Certificate_Store_Windows::all_subjects()` by avoiding the full `X509_Certificate` parsing. Instead this patch now parses the Subject DN directly from the provided buffer.

On my Windows VM the speedup of the respective unit-test is fairly significant:

```
# Before
System Certificate Store - Find all Certificate Subjects ran 3 tests in 100.16 msec all ok

# After
System Certificate Store - Find all Certificate Subjects ran 3 tests in 4.31 msec all ok
```

### DIFF
--- a/src/lib/x509/certstor_system_windows/certstor_windows.cpp
+++ b/src/lib/x509/certstor_system_windows/certstor_windows.cpp
@@ -1,16 +1,18 @@
 /*
 * Certificate Store
-* (C) 1999-2019 Jack Lloyd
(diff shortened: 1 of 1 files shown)


randombit:
Nice improvement!

PR #2790, opened 2021-08-03
Avoid using short exponents with ElGamal

randombit:
Some off-brand PGP implementation generates keys where p - 1 is smooth, as a result short exponents can leak enough information about k to allow decryption.

Needs backport for 2.x

cc @ronaldtse


ronaldtse:
Thank you for the ping @randombit !

CC @ni4 @dewyatt @antonsviridenko

antonsviridenko:
@randombit what measures should be taken by rnp developers to address this issue?

src/lib/pubkey/elgamal/elgamal.cpp
```diff
@@ -113,8 +113,12 @@ ElGamal_Encryption_Operation::raw_encrypt(const uint8_t msg[], size_t msg_len,
    if(m >= m_group.get_p())
       throw Invalid_Argument("ElGamal encryption: Input is too large");
 
-   const size_t k_bits = m_group.exponent_bits();
-   const BigInt k(rng, k_bits);
+   /*
+   Some ElGamal implementations foolishly use prime fields where p - 1 is
+   smooth, as a result it is unsafe to use short exponents.
+   */
```
defeo, line 119: Hi. I'm one of the discoverers of the vulnerability (described [here](https://ibm.github.io/system-security-research-updates/2021/07/20/insecurity-elgamal-pt1) and [here](https://eprint.iacr.org/2021/923)).

It seems there is a misunderstanding on the nature of the vulnerability (or maybe it's just an improper use of the word "smooth"): the affected public keys strictly follow El Gamal's original description, and do nothing "foolish". In particular, p-1 is not smooth: it contains a large prime factor, but also a large smooth factor, just like in DSA keys. The blame cannot fall solely on the library that does public key generation: it's a complex issue involving the whole OpenPGP ecosystem.

This patch properly fixes the vulnerability, but the comment is misleading. You may want to edit it, so that in, say, 10 years time some maintainer doesn't feel it's ok to ignore "foolish" implementations. Note that there is nothing you can do against really foolish implementations that use truly smooth p-1: for these, message recovery is always possible no matter what the encrypting party does.

randombit: Thanks, I did in fact just skim and clearly misunderstand the paper. I thought the issue was these keys had chosen `p` completely at random rather than via a structured way and due to poor luck had generated a `p` where `p - 1` truly had only small factors. I will edit the comment.

defeo: This is not very important, I'm just being a 🤓: if you took a `p` of 2048 bits at random, the chances that all the prime factors of `p-1` are smaller than, say, 200 bits are ≈10^-11. You would need some **criminally** poor luck for that to happen!

Relatedly, we would definitely appreciate if the comment and/or commit message contained a link to our paper 😊

PR #2795, opened 2021-09-02
Clarify behavior of RNG in docstring

lieser:
Clarify behavior of the RNG interface in docstring. Not just for potential users, but also to make it easier to implement a custom RNG.


randombit:
Thanks!

PR #2796, opened 2021-09-09
secure_nodump api proposal.

devnexen:
malloc_conceal functions just like a normal malloc call
but the content won t appear in core dumps and goes to extra step
 zeroing the memory.

### DIFF
--- a/src/lib/utils/mem_ops.cpp
+++ b/src/lib/utils/mem_ops.cpp
@@ -25,7 +25,11 @@ BOTAN_MALLOC_FN void* allocate_memory(size_t elems, size_t elem_size)
       return p;
 #endif
 
+#if defined(BOTAN_TARGET_OS_HAS_ALLOC_CONCEAL)
+   void *ptr = ::calloc_conceal(elems, elem_size);
+#else
    void* ptr = std::calloc(elems, elem_size);
+#endif
    if(!ptr)
       throw std::bad_alloc();
    return ptr;
@@ -42,7 +46,6 @@ void deallocate_memory(void* p, size_t elems, size_t elem_size)
    if(mlock_allocator::instance().deallocate(p, elems, elem_size))
       return;
 #endif
-
    std::free(p);
    }
 
--- a/src/build-data/os/openbsd.txt
+++ b/src/build-data/os/openbsd.txt
@@ -16,6 +16,7 @@ arc4random
 getentropy
 explicit_bzero
 pledge
+alloc_conceal
 
 atomics
 sockets


randombit:
This is a feature we should take advantage of but I don't think this is the right approach. For one we don't want to use `mlock` here - the inputs are not going to be page sized or paged aligned, so we are "wasting" our typically very limited `mlock` rlimit. (On BSD I think this is a bit higher but default Linux is only 64K total per process).

For similar reason I don't think using `DONTDUMP` is a good approach here as it is likely to overlap with other general allocations and make debugging very difficult. If we wanted to go this path I think the better way would be to extend the pool allocator, which already handles page alignment, to also use unlocked pages. But I am not sure this is worth the bother - we don't really have a good approach for core dump security (due to use of `std::string`, etc) and the current uses of DONTDUMP are just in there as an extra because it is easy to do.

We should tie this into the normal `allocate_memory` function instead, since that function is only ever used for (what might be) key material.

I think the approach is to modify `allocate_memory` with something like

```
-   void* ptr = std::calloc(elems, elem_size);
+  #if defined(...)
+  void* ptr = malloc_conceal(...);
+  memset(ptr, 0, len);
+ #else
+  void* ptr = std::calloc(elems, elem_size);
+ #endif
```

which also avoids having to create any new API endpoints.

src/lib/utils/mem_ops.cpp
```diff
@@ -25,9 +25,17 @@ BOTAN_MALLOC_FN void* allocate_memory(size_t elems, size_t elem_size)
       return p;
 #endif
 
+#if defined(BOTAN_TARGET_OS_HAS_ALLOC_CONCEAL)
+   size_t len = elems * elem_size;
+   void *ptr = ::malloc_conceal(len);
+   if(!ptr)
+      throw std::bad_alloc();
+   ::memset(ptr, 0, len);
```
randombit, line 33: Could this use `calloc_conceal` instead?

src/lib/utils/mem_ops.cpp
```diff
@@ -42,8 +50,11 @@ void deallocate_memory(void* p, size_t elems, size_t elem_size)
    if(mlock_allocator::instance().deallocate(p, elems, elem_size))
       return;
 #endif
-
+#if defined(BOTAN_TARGET_OS_HAS_ALLOC_CONCEAL)
+   ::freezero(p, elems * elem_size);
```
randombit, line 54: The OpenBSD docs say "The malloc_conceal() and calloc_conceal() functions behave the same as malloc() and calloc() respectively, with the exception that the allocation returned is marked with the MAP_CONCEAL mmap(2) flag and **calling free() on the allocation will discard the contents explicitly.**" so AIUI we don't need this? Just calling `free` will cause the value to be zeroed as a side effect.

(Not that it matters in any case given we start the function off by zeroing the bits before calling any kind of free function)

randombit (changes requested):

src/lib/utils/mem_ops.cpp
devnexen: yes I missed the scrub_memory call...

PR #2797, opened 2021-09-14
Fix a minor OAEP side channel

randombit:
OAEP had some logic to handle how RSA used to work, but this was already fixed way back in b8966d0f8.

Thanks to @lieser for pointing out this issue.


lieser:
Looks good to me.

PR #2798, opened 2021-09-14
Clarify some docstrings and comments

lieser:
Some small clarifications in docstrings and comments, mostly related to OAEP.


randombit:
Thanks!

PR #2804, opened 2021-09-25
Add shared library support for mingw

lazka:
This adds the following output files when building the shared library
with mingw (tested with mingw-w64 and both gcc and clang):

bin/libbotan-3.dll
lib/libbotan-3.dll.a

Also look for the DLL name when loading the Python bindings
(looking for both versions to mirror the MSVC logic)


randombit:
Thank you!

PR #2806, opened 2021-09-28
TLS 1.2 refactoring and TLS 1.3 development started

pist-eb:
This PR consists of 2 commits.

### TLS 1.2 refactoring

1.  Using private implementation for TLS Channel, Client and Server. Motivation:
- Common interface for TLS 1.2 and TLS 1.3 classes,
- Protocol-version specific internal implementation,
- Clear separation of TLS 1.2 and TLS 1.3 implementation,
- Ability to switch the internal implementation during the handshake, i.e. TLS client proposes TLS 1.3 and TLS 1.2, while TLS server selects the TLS 1.2.
2.  Using private implementation for TLS handshake messages. Motivation:
- Handshake messages which appear in both versions of protocol shall have common interface,
- Internal implementation and interpretation of data shall be protocol-version specific.
3. Creating the factories to instantiate proper versions of internal implementation. Motivation:
- Unifying the way of creating an internal implementation depending on selected protocol version.

### TLS 1.3 development started

1. Adding new module for Botan: `tls13`,
2. Extending the `Protocol_Version` structure with `TLS_V13`,
3. Extending the `Policy` with `TLS_V13` support,
4. Implementing/testing new extensions:
- Cookie (44),
- Key_share (51),
- Signature_algorithms_cert (50),
- Supported_versions (43),
- Supported_groups (10),
- Signature_algorithms (13).
5. Implementing the skeleton of Channel and Client for TLS 1.3,
6. Implementing the skeleton of Client_Hello for TLS 1.3.


pist-eb:
Please note, that there are changes in configure.py, which enables include dependencies - the change was added to properly detect the changes in header files and speed up the development (temporary solution, as configuration time is much longer then.

PR #2808, opened 2021-10-01
TLS 1.2 Refactoring

hrantzsch:
The refactoring of the TLS 1.2 code done by @pist-eb and his colleagues in #2806 only. This way we can review and merge into the TLS 1.3 dev branch independently of the added TLS 1.3 code.


pist-eb:
> In a first pass over the changes I focussed on the refactoring of the TLS message implementations. Mainly sanity checking the PIMPL-scheme layout. Furthermore, I double-checked the upstream protocol implementation code and the associated relocated `_impl_12` classes.
> 
> No real surprises, but I'd like to discuss whether we should revert the splitting of the `Server_Hello` parsing code into the abstract `_Impl` and concrete `_Impl_12` classes. I feel it would be better to do those (logic altering) changes in the next step. That said: I didn't find obvious issues with the separation either. @pist-eb, what do you think?
> 
> Anyway! Great to see the protocol version abstraction falling into place. 👍

Honestly, the border between the 'refactoring of TLS 1.2' and 'implementation of TLS 1.3' was quite thin and maybe we partially started some TLS 1.3 logic changes already during refactoring. Please, feel free to adapt refactored code to your needs and ideas :)

reneme (approved):
Good for now. I'm sure we'll find many more things while getting our feet wet when implementing TLS 1.3.

PR #2809, opened 2021-10-01
Replace expired 'DST Root CA X3' by 'ISRG Root X1

reneme:
This certificate is used for some system certificate store tests. 'DST Root CA X3' expired yesterday, failing those tests on CI.


reneme:
Meh. Windows still fails, because it seems to be unable to find `ISRG Root X1`. I don't have access to a windows machine right now. @trafo any chance you try to compile this PR and run the tests on your machine next week?

Even though I'd be surprising, but could it be that AppVeyor doesn't include the `ISRG Root X1` into their trust store? Especially as this is now Lets Encrypt's root, I would actually rule this out. 

Is there a better way to find this, than [RDP-ing into an AppVeyor machine](https://github.com/randombit/botan/pull/1931#issuecomment-491349599), @randombit ?

PR #2810, opened 2021-10-01
Backport: Replace expired 'DST Root CA X3' by 'ISRG Root X1'

reneme:
see also: https://github.com/randombit/botan/pull/2809


reneme:
Closed in favour of #2820

PR #2831, opened 2021-10-27
FIX: potential segfault in TLS::Ciphersuite::to_string()

reneme:
`m_iana_id` is a `const char*` that defaults to `nullptr`. Hence, the following code would currently crash:

```
uint16_t unexpected_ciphersuite_id = 1337;
TLS::Ciphersuite::by_id(unexpected_ciphersuite_id).to_string();
```

Now it returns "unknown cipher suite". Perhaps, a meaningful `throw` would be reasonable instead?!

### DIFF
--- a/src/lib/tls/tls_handshake_state.cpp
+++ b/src/lib/tls/tls_handshake_state.cpp
@@ -219,7 +219,15 @@ void Handshake_State::client_hello(Client_Hello* client_hello)
 void Handshake_State::server_hello(Server_Hello* server_hello)
    {
    m_server_hello.reset(server_hello);
-   m_ciphersuite = Ciphersuite::by_id(m_server_hello->ciphersuite());
+
+   auto suite = Ciphersuite::by_id(m_server_hello->ciphersuite());
+   if (!suite.has_value())
+      {
+      throw Decoding_Error("Failed to find cipher suite for ID " +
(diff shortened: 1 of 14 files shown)


src/lib/tls/tls_session.cpp
```diff
@@ -113,6 +114,13 @@ Session::Session(const uint8_t ber[], size_t ber_len)
                            " no longer supported");
       }
 
+   if(!Ciphersuite::by_id(ciphersuite_code))
+      {
+      throw Decoding_Error("Serialized TLS session contains unknown cipher suite "
```
hrantzsch, line 119: Maybe `Invalid_State` would be more fitting than `Decoding_Error`.

hrantzsch (approved):

src/lib/tls/tls_handshake_state.cpp
```diff
@@ -289,6 +289,15 @@ void Handshake_State::client_finished(Finished* client_finished)
    note_message(*m_client_finished);
    }
 
+const Ciphersuite& Handshake_State::ciphersuite() const
+   {
+   if (!m_ciphersuite.has_value())
+      {
+      throw Decoding_Error("Cipher suite is not set");
```
hrantzsch, line 296: Maybe `Invalid_State` would be more fitting than `Decoding_Error`.

PR #2840, opened 2021-10-28
CI fuzz test2 [ DO NOT MERGE]

DavidKorczynski:


randombit:
@DavidKorczynski so sorry I forgot I had enabled that as a precaution against someone opening a PR using a malicious action. With `google/oss-fuzz*` added to the allowed list everything seems to be working correctly https://github.com/randombit/botan/actions/runs/1403483785 

Thanks again for the original PR and helping debug this.

PR #2850, opened 2021-11-16
Fix compilation issue on older mac (< 10.12)

tenzap:
Error was:
```
In file included from src/lib/entropy/getentropy/getentropy.cpp:13:
/usr/include/sys/random.h:37:32: error: unknown type name 'u_int'
void read_random(void* buffer, u_int numBytes);
                               ^
/usr/include/sys/random.h:38:33: error: unknown type name 'u_int'
void read_frandom(void* buffer, u_int numBytes);
                                ^
/usr/include/sys/random.h:39:33: error: unknown type name 'u_int'
int  write_random(void* buffer, u_int numBytes);
                                ^
```

sys/types.h header which provides u_int was missing.

Discovered in 2.18.2 but PR is towards master. Maybe this could be backported.

### DIFF
--- a/src/lib/entropy/getentropy/getentropy.cpp
+++ b/src/lib/entropy/getentropy/getentropy.cpp
@@ -10,6 +10,10 @@
 #if defined(BOTAN_TARGET_OS_IS_OPENBSD) || defined(BOTAN_TARGET_OS_IS_FREEBSD) || defined(BOTAN_TARGET_OS_IS_SOLARIS)
    #include <unistd.h>
 #else
+   #if defined(BOTAN_TARGET_OS_HAS_POSIX1)
+      // Allows successful compilation on macOS older than 10.12: Provides a missing typedef for `u_int`.
+      #include <sys/types.h>
+   #endif
    #include <sys/random.h>
 #endif
 


reneme:
Frankly, I'm not sure whether we should add patches targeting old macOS platforms that are no longer supported by Apple. Taking [this python issue as evidence](https://bugs.python.org/issue27596#msg271069), macOS 10.11 doesn't even provide `getentropy()` in the first place. Furthermore the broken function declarations in question (`read_random()` and `write_random()`) from this old `sys/random.h` aren't used by Botan anyways.

Let me suggest to add `--disable-modules=getentropy` to the `configure.py` invocation when building on macOS < 10.12 which should disable this module altogether.

tenzap:
The background is that I'm updating the Portfile to build botan on MacPorts.
I can of course use`--disable-modules=getentropy` as you suggest but this is not the prefered way on MacPorts since it is able to make some missing functions available on older systems. Typically getentropy can be made available on osx < 10.12.

Although osx 10.11 is not supported anymore by Apple, it doesn't mean it is not used anymore. Unfortunately some older macs which are still totally usable nowadays can't be updated to newer osx versions.

This change allows to have botan on older macs. If it isn't available on older macs it blocks some software to be available on older macs too if they have a hard dependency on botan. This is for example the case of keepassxc 2.7 which is still in development.

Of course we could keep that patch (which simply includes a `missing` header) only in macports, but I think it is better if it also merged upstream. That may benefit someone else.

src/lib/entropy/getentropy/getentropy.cpp
```diff
@@ -10,6 +10,9 @@
 #if defined(BOTAN_TARGET_OS_IS_OPENBSD) || defined(BOTAN_TARGET_OS_IS_FREEBSD) || defined(BOTAN_TARGET_OS_IS_SOLARIS)
    #include <unistd.h>
 #else
+   #if defined(BOTAN_TARGET_OS_HAS_POSIX1)
+      #include <sys/types.h>
```
reneme, line 14:
```suggestion
      // Allows successful compilation on macOS older than 10.12: Provides a missing typedef for `u_int`.
      #include <sys/types.h>
```

reneme (approved):
I understand your point and agree that it makes sense to upstream this patch; particularly given that it is such a simple fix. I added a comment for later reference. Thanks for the contribution. Could you open a second PR for the 2.18.x backport please?

tenzap:
Which branch should I commit to for 2.18.x ?

reneme:
> Which branch should I commit to for 2.18.x ?

Your other PR is correctly targeted towards `release-2`. Thanks again.

PR #2853, opened 2021-11-22
add missing header: cstdlib (2.x branch)

tenzap:
for compilation on older systems (like macos 10.11)

Needed for std::free() & std::calloc() on older systems [1]
That file is included indirectly on newer systems.

[1] https://en.cppreference.com/w/cpp/memory/c/free

### DIFF
--- a/src/fuzzer/mem_pool.cpp
+++ b/src/fuzzer/mem_pool.cpp
@@ -11,7 +11,7 @@
 #include <map>
 #include <utility>
 
-#include <stdlib.h>
+#include <cstdlib>
 
 namespace {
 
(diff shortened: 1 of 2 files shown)


src/fuzzer/mem_pool.cpp
```diff
@@ -12,6 +12,7 @@
 #include <utility>
 
 #include <stdlib.h>
+#include <cstdlib>
```
reneme, line 15: See https://github.com/randombit/botan/pull/2854#discussion_r754964517.

reneme (approved):

PR #2854, opened 2021-11-22
add missing header: cstdlib

tenzap:
for compilation on older systems (like macos 10.11)

Needed for std::free() & std::calloc() on older systems [1]
That file is included indirectly on newer systems.

[1] https://en.cppreference.com/w/cpp/memory/c/free

### DIFF
--- a/src/fuzzer/mem_pool.cpp
+++ b/src/fuzzer/mem_pool.cpp
@@ -11,7 +11,7 @@
 #include <map>
 #include <utility>
 
-#include <stdlib.h>
+#include <cstdlib>
 
 namespace {
 
--- a/src/lib/compat/sodium/sodium_utils.cpp
+++ b/src/lib/compat/sodium/sodium_utils.cpp
@@ -11,6 +11,7 @@
 #include <botan/internal/os_utils.h>
 #include <botan/internal/ct_utils.h>
 #include <botan/internal/loadstor.h>
+#include <cstdlib>
 
 namespace Botan {
 


src/fuzzer/mem_pool.cpp
```diff
@@ -12,6 +12,7 @@
 #include <utility>
 
 #include <stdlib.h>
+#include <cstdlib>
```
reneme, line 15: Shouldn't it be okay to remove `#include <stdlib.h>` if you include `cstdlib`? If both need to stay for some sort of legacy workaround, please leave a comment explaining why.

reneme (approved):

src/fuzzer/mem_pool.cpp
tenzap: I don't know, I didn't check further. 

If I remove stdlib.h does part of the code have to be rewritten to conform to cstdlib? I'm not at my ease in c/c++, sorry.

reneme: Usually, `cstdlib` should be enough. Could you try to remove `stdlib.h` and see if it still compiles on the old platforms?

On my macOS (11.6.1) it works just fine without `stdlib.h`.

tenzap: I can only check on 10.13. since it builds on macports with this patch on 10.11 it will be difficult to get such a test (just to test that it will builds) accepted in macports' production servers.

tenzap: It builds fine on 10.13. I will re-commit.

PR #2860, opened 2021-12-11
Cleanup SHA-3 Clang workaround

randombit:
This makes it more clear what the real workaround is vs what the code "should be". It will also make it easier to remove this workaround when it is no longer required. Finally sharing the SHA-3 round in a header reduces the confusion wrt the two duplicate implementations.


reneme (approved):
I agree, duplicating the affected code block reads clearer than the `MAYBE_INLINE` macro fairy dust.

Issue #2852, opened 2021-11-18
unknown file type: botan-test.bc when converting tests to wasm


radekgs:
Following the instructions on https://botan.randombit.net/handbook/building.html#emscripten-webassembly, after successfully compiling the framework and trying to folllow the second step of converting the tests to wasm I get the following error:

`~/git/botan$ em++ -s ALLOW_MEMORY_GROWTH=1 -s DISABLE_EXCEPTION_CATCHING=0 -s WASM=1    --preload-file src/tests/data botan-test.bc -o botan-test.html
wasm-ld: error: unknown file type: botan-test.bc
em++: error: '/home/radek/git/emsdk/upstream/bin/wasm-ld -o botan-test.wasm botan-test.bc -L/home/radek/git/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten -lGL -lal -lhtml5 -lstubs-debug -lc -lcompiler_rt -lc++ -lc++abi -ldlmalloc -lc_rt_wasm -lsockets -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-cxx-exceptions -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr --import-undefined --strip-debug --export-if-defined=main --export-if-defined=emscripten_stack_get_end --export-if-defined=emscripten_stack_get_free --export-if-defined=emscripten_stack_init --export-if-defined=stackSave --export-if-defined=stackRestore --export-if-defined=stackAlloc --export-if-defined=__wasm_call_ctors --export-if-defined=fflush --export-if-defined=__errno_location --export-if-defined=malloc --export-if-defined=free --export-if-defined=__cxa_is_pointer_type --export-if-defined=__cxa_can_catch --export-if-defined=setThrew --export-if-defined=__start_em_asm --export-if-defined=__stop_em_asm --export-table -z stack-size=5242880 --initial-memory=16777216 --no-entry --max-memory=2147483648 --global-base=1024' failed (returned 1)`

reneme:
Sorry for the long waiting time, @radekgs. Admittedly, we're no experts on WASM (, yet). It would be great if you could try out the patch we provided in #2864 and verify that this helps.

lieser:
If the mentioned patch did not help please reopen this issue and provide more information.

PR #2864, opened 2021-12-14
Improve emscripten build target

reneme:
This tries to improve the emscripten/wasm build support in botan by providing deeper integration into the build configuration scripts. The previous rudimentary emscripten support was integrated in 2018 and [seems to have degraded in the meantime](https://github.com/randombit/botan/issues/2852).

We tried this with Emscripten 3.0.

Building the static library and a `botan-test.html` "executable" should now work like this:

```
./configure.py --cpu=wasm --os=emscripten
make
```

@radekgs Could you try this out and see if it fits your needs, please?


randombit:
Thanks! I'll review tonight. Maybe we should add this to CI to prevent further regressions?

reneme:
CI would be great indeed! Building shouldn't be a big issue and would likely catch the worst regressions.
Admittedly, I didn't get my feet wet on the web platform so far. So I'm not sure about how to run the tests. We could do some research, though.

In the browser, a few tests fail currently (one seems flaky):

```
[...]
certstor_system:
System Certificate Store - Open Keychain ran 1 tests 1 FAILED
Failure 1: I/O error: DataSource: Failure opening file /etc/ssl/cert.pem
[...]
filter:
filter ran 1 tests 1 FAILED
Failure 1: I/O error: Pipe input operator (unixfd) has failed
[...]
os_utils:
OS::get_cpu_available ran 1 tests all ok
OS::get_cpu_cycle_counter ran 1 tests all ok
OS::get_high_resolution_clock ran 2 tests 1 FAILED
Failure 1: OS::get_high_resolution_clock high resolution clock eventually changes value unexpected result 453 >= 128
[...]
Tests complete ran 2540233 tests in 531.16 sec 3 tests failed (in certstor_system filter os_utils)
```

System certificate store fails for good reasons, I guess. We could probably just disable this module on emscripten. The `os_utils` test doesn't seem to be happy with the resolution of the high resolution clock. That's flaky and doesn't fail in every run. I suppose that's because browsers introduce random jitter into their clock APIs. Not sure about the `filter` test, to be honest.

hrantzsch:
Maybe we can use https://github.com/bytecodealliance/wasmtime to run the tests.

randombit:
Even just a build-only CI would go a long way to preventing problems.

randombit (approved):
CI would be nice but is optional. Changes lgtm

reneme:
Added a CI build step for emscripten (running on `macos-latest` w/ `brew install emscripten`). Unfortunately `ubuntu-20.04` doesn't seem to provide emscripten in its default `apt-get` repos.

CI currently builds emscripten but doesn't run the tests....

src/build-data/os/emscripten.txt
```diff
@@ -15,3 +17,8 @@ filesystem
 dev_random
 posix1
 </target_features>
+
+<aliases>
+emscripten
```
randombit, line 22: Why the self-alias?

reneme: Otherwise the [`.matches_name("emscripten")`](https://github.com/randombit/botan/pull/2864/files#diff-4d5f3192809ec1b9add6b33007e0c50031ad9a0a2f3f55a481b506468824db2cR2161) didn't work.

randombit:
No response from @radekgs so far but I think this is certainly an improvement over current state so IMO we should merge and if there are additional issues try to address them in a followup.

PR #2867, opened 2021-12-17
Backport #2860 - Cleanup SHA-3 Clang workaround

randombit:


src/lib/hash/sha3/sha3_round.h
```diff
@@ -0,0 +1,123 @@
...
+
+namespace Botan {
+
+#if defined(BOTAN_HAS_BROKEN_CLANG_SHA3)
+
+namespace {
+
+__attribute__((noinline)) decltype(auto) xor_CNs(const uint64_t A[25])
+   {
+   return std::tuple<uint64_t, uint64_t, uint64_t, uint64_t, uint64_t>(
```
reneme, line 37: For the record: For this to [work with older GCC](https://github.com/randombit/botan/pull/2849/files) the `std::tuple<...>` type needed to be repeated both for the return statement and the function declaration. Though, this code path being compiled only with the broken clang version, it shouldn't be a problem.

reneme (approved):

src/lib/hash/sha3/sha3_round.h
randombit: One reason I wanted to have totally distinct codepaths for broken-Clang vs everything else - we only have to worry about the workaround in the specific case that we're being compiled under this particular compiler+version instead of having to work everywhere.

Oh that reminds me though, I think GCC 4.4 is still missing from the new GH actions CI and if that continues we'll certainly have more build regressions...

PR #2872, opened 2021-12-22
WIP: Add support for Kyber PQ-Safe KEM

boricm:
Adding support for Kyber (post-quantum-safe KEM) as specified in Round 3 of the NIST post-quantum project. Specification and link to NIST submission package: https://pq-crystals.org/kyber/resources.shtml

Contributed by:
@reneme @hrantzsch @omarama @boricm


randombit:
This is great to see! I'll get a review in ASAP.

Until then, a topic for discussion -- do we really need the 90s mode? It seems like it adds a relatively large amount of complexity. Will anyone use it?

mouse07410:
I don't know. 

-90s is faster, but my benchmarks of the "normal" code showed performance sufficient for my needs.

Still, if it's not too big a headache, why not support both...

hrantzsch:
> do we really need the 90s mode?

Hm, good question. I can't say that we need it specifically, but I think it's a great use case for Botan's module system. If someone does have a configuration where the more modern primitives are not available, we can support that very well. The differences between the modes are also pretty local so it shouldn't be too difficult to separate the shared code parts from the specific parts.

randombit (commented):
An initial review. Many comments here and there will be more later but overall a good direction.

Another question wrt 90s mode. Currently the code makes a distinction between 90s and modern at key generation time. Would it make more sense to have this be an attribute of the KEM? That is, `KyberMode` is just 512/768/1024, which encapsulates the algorithmic params like N, eta1, etc. And then there is some additional abstraction (`KyberSymmetricPrimitives` or something like that) that implements the hashes, PRFs, etc, and the KEM instantiates one implementation or another (90s/modern) based on the param strings which are currently unused.

I think this design would make it a little simpler to actually realize the promise of supporting Kyber in limited configurations as it would cleanly split out the AES/SHA-2 vs SHA_3/SHAKE code into two implementations of `KyberSymmetricPrimitives` and these could be implemented in submodules which in turn depend on the relevant primitives.

One concern here is that maybe the Kyber designers do not recommend mixing modes for any particular key. It's been a long time since I read the Kyber specs and I don't recall what if anything they have to say about this.

src/tests/test_kyber.cpp
```diff
@@ -0,0 +1,373 @@
...
+                }
+
+                Test::Result result( test_name );
+
+                uint8_t salt[1];
+                size_t shared_secret_length = 32;
+
+                // Alice
+                std::unique_ptr<Botan::RandomNumberGenerator> kyber_rng_alice;
+                kyber_rng_alice.reset( new Botan::AutoSeeded_RNG );
```
randombit, line 195: This should use `Test::rng()` so that the tests are reproducible

src/tests/test_kyber.cpp
```diff
@@ -0,0 +1,373 @@
...
+                auto priv_key = Botan::Kyber_PrivateKey( *kyber_rng_alice, mode );
+
+                // Bob
+                std::unique_ptr<Botan::RandomNumberGenerator> kyber_rng_bob;
+                kyber_rng_bob.reset( new Botan::AutoSeeded_RNG );
```
randombit, line 200: `Test::rng()`

src/tests/test_kyber.cpp
```diff
@@ -0,0 +1,373 @@
...
+                case Botan::KyberMode::Kyber1024:
+                    test_name = "kyber1024 test API";
+                    break;
+                default:
+                    throw std::runtime_error( "unknown kyber mode in run_kyber_test()" );;
+                }
+
+                Test::Result result( test_name );
+
+                uint8_t salt[1];
```
randombit, line 190: Uninitialized??

src/tests/test_kyber.cpp
```diff
@@ -0,0 +1,373 @@
...
+                kyber_rng_alice.reset( new Botan::AutoSeeded_RNG );
+                auto priv_key = Botan::Kyber_PrivateKey( *kyber_rng_alice, mode );
+
+                // Bob
+                std::unique_ptr<Botan::RandomNumberGenerator> kyber_rng_bob;
+                kyber_rng_bob.reset( new Botan::AutoSeeded_RNG );
+                auto pub_key = Botan::Kyber_PublicKey( priv_key.public_key_bits(), mode );
+                auto enc = pub_key.create_kem_encryption_op( *kyber_rng_bob, "", "" );
+                Botan::secure_vector<uint8_t> cipher_text, key_bob;
+                enc->kem_encrypt( cipher_text, key_bob, shared_secret_length, *kyber_rng_bob, salt, 0 );
```
randombit, line 204: If `salt_len` is zero then `nullptr` should be accepted

src/tests/test_kyber.cpp
```diff
@@ -0,0 +1,373 @@
...
+            std::vector<uint8_t> sk_in = vars.get_req_bin( "sk" );
+            std::vector<uint8_t> ct_in = vars.get_req_bin( "ct" );
+            std::vector<uint8_t> ss_in = vars.get_req_bin( "ss" );
+
+            uint8_t salt[1];
+            size_t shared_secret_length = 32;
+
+            // Kyber test RNG
+            std::unique_ptr<Botan::RandomNumberGenerator> kyber_test_rng;
+            kyber_test_rng.reset( new Kyber_Test_RNG( seed_in ) );
```
randombit, line 269: `auto keyber_test_rng = std::make_unique<Kyber_Test_RNG>(seed_in)`

randombit:
Answering my own question

> Another question wrt 90s mode. Currently the code makes a distinction between 90s and modern at key generation time. Would it make more sense to have this be an attribute of the KEM? 

After reviewing the Kyber spec I remember why this is not possible, the `A` matrix is derived via the PRF and so every key is either modern or 90s, forever.

For this reason I think we should probably have two distinct OIDs, so the keys can be distinguished.

Or possibly 6 OIDs? One per size + 90s vs modern.

mouse07410:
IMHO, offhand, I'd prefer 6 OIDs.

src/lib/pubkey/kyber/kyber.cpp
```diff
@@ -0,0 +1,1817 @@
...
+     * Description: Add two polynomials
+     *
+     * Arguments: - poly *r:       pointer to output polynomial
+     *            - const poly *a: pointer to first input polynomial
+     *            - const poly *b: pointer to second input polynomial
+     **************************************************/
+    Polynomial &operator+=(const Polynomial &other)
+    {
+        for (size_t i = 0; i < this->coeffs.size(); ++i)
+            this->coeffs[i] = this->coeffs[i] + other.coeffs[i];
```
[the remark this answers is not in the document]
omarama: 1. reduces are always called after the "+=" ( poly_add in the reference implementation ). For performance reasons the reduce will not executed after every "+=". There are a few "+=" operator's in the code but mostly after four "+=" operations there is a reduce call. 
2. If we assume that we have a reduced polynomial, we can add around 18 times before a buffer overflow will happen. Nevertheless we can insert a buffer overflow check if you like. We think that this check is actually not necessary but it will also not harm the code.

src/tests/test_kyber.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+        // Kyber test RNG
+#define RNG_SUCCESS      0
+#define RNG_BAD_MAXLEN  -1
+#define RNG_BAD_OUTBUF  -2
+#define RNG_BAD_REQ_LEN -3
+
+        void AES256_ECB( unsigned char* key, unsigned char* ctr, unsigned char* buffer );
+
+        void AES256_CTR_DRBG_Update( unsigned char* provided_data, unsigned char* Key, unsigned char* V )
```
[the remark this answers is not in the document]
hrantzsch: We went with a `Fixed_Output_RNG` and hardcoding the output in the test vectors instead.

src/lib/pubkey/kyber/kyber.cpp
```diff
@@ -0,0 +1,1817 @@
...
+        extkey.reserve(seed.size() + 1);
+        extkey.insert(extkey.end(), seed.begin(), seed.end());
+        extkey.push_back(nonce);
+
+        secure_vector<uint64_t> sponge_state(25);
+        size_t sponge_state_pos =
+            Botan::SHA_3::absorb(KyberConstants::kShake256Rate, sponge_state, 0, extkey.data(), extkey.size());
+
+        Botan::SHA_3::finish(KyberConstants::kShake256Rate, sponge_state, sponge_state_pos, 0x1F, 0x80);
+        Botan::SHA_3::expand(KyberConstants::kShake256Rate, sponge_state, out.data(), out.size());
```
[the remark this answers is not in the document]
reneme: Yep, we were struggling with the interface in the beginning. Done now.

src/lib/pubkey/kyber/kyber.cpp
```diff
@@ -0,0 +1,1817 @@
...
+    }
+    else
+    {
+        // 90s mode
+        std::unique_ptr<Botan::StreamCipher> cipher( Botan::StreamCipher::create_or_throw( "CTR-BE(AES-256)" ) );
+        cipher->set_key( seed );
+        // IV is zero padded to block length internally
+        uint8_t iv[12] = { 0 };
+        iv[0] = nonce;
```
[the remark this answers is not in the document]
reneme: Nice! TIL... thanks.

src/lib/pubkey/kyber/kyber.cpp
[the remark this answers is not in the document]
randombit: Thanks. Please add a comment indicating these bounds and the need to call reduce eventually, along with a `BOTAN_DEBUG_ASSERT` that the add doesn't overflow. (Similarly for `-=`)

src/lib/pubkey/kyber/kyber.cpp
```diff
@@ -0,0 +1,1817 @@
...
+    {
+        return polynomial_vector_byte_length() + kSeedLength;
+    }
+
+    size_t private_key_byte_length() const
+    {
+        return polynomial_vector_byte_length() + public_key_byte_length() + kPublicKeyHashLength + kZLength;
+    }
+
+    std::unique_ptr<HashFunction> H() const
```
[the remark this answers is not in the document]
hrantzsch: Obsolete after module refactoring, I believe.

src/lib/pubkey/kyber/kyber_common/kyber.cpp
```diff
@@ -117,6 +117,8 @@ class KyberConstants
                 return "Kyber-1024-r3";
             case KyberMode::Kyber1024_90s:
                 return "Kyber-1024-90s-r3";
+            default:
+                BOTAN_ASSERT(false, "unreachable code");
```
hrantzsch, line 121: Do you think it'd be worthwhile to introduce an explicit `BOTAN_UNREACHABLE`? Either simply as an alias of the above or via `__builtin_unreachable` and friends.

src/lib/pubkey/kyber/kyber_common/kyber.h
```diff
@@ -0,0 +1,106 @@
...
+
+#include <botan/pk_keys.h>
+#include <botan/exceptn.h>
+#include <botan/der_enc.h>
+
+#include <botan/internal/pk_ops_impl.h>
+
+#if !defined(BOTAN_HAS_KYBER_90S) && !defined(BOTAN_HAS_KYBER)
+   static_assert(false, "botan module 'kyber_common' is useful only when enabling modules 'kyber', 'kyber_90s' or both");
+#endif
```
reneme, line 25: We opted to introduce three modules `kyber` (modern), `kyber_90s` and `kyber_common`. The latter contains the actual algorithm and supporting math code. The former two contain adapters to the required symmetric algorithms for XOF, KDF and more.

Hence, users enabling `kyber_common` exclusively at `./configure.py` wouldn't end up with a functional kyber implementation. They'd need at least one of the symmetric algorithm adapters in the other mentioned modules. This `static_assert` is an attempt to warn them of their mistake at build time.

To the best of our knowledge, Botan's module system doesn't have a notion of an "internal module" for such use cases, right? Is that potentially something we'd want to introduce?

mouse07410:
IMHO, it makes no sense to allow `kyber_common` for _explicit_ configuration. 
I'd have only `kyber` and `kyber_90s` as "configurable", each of which would force inclusion of `kyber_common`. Adding it explicitly without one of the other two makes no sense, disabling it when at least one of the others is enabled doesn't make sense either.

src/lib/pubkey/kyber/kyber_common/kyber.h
reneme: Replying to @mouse07410 inline, to keep the context:

> IMHO, it makes no sense to allow kyber_common for explicit configuration.
I'd have only kyber and kyber_90s as "configurable", each of which would force inclusion of kyber_common. Adding it explicitly without one of the other two makes no sense, disabling it when at least one of the others is enabled doesn't make sense either.

I agree with you! Though, given the inner workings of Botan's module system, we didn't figure out how to do that without introducing the `kyber_common` module. I.e. this solution is a technical crutch for the time being. Is there a better way, @randombit?

src/lib/pubkey/kyber/kyber.cpp
```diff
@@ -0,0 +1,1817 @@
+#include <botan/hash.h>
+#include <botan/kyber.h>
+#include <botan/loadstor.h>
+#include <botan/mem_ops.h>
+#include <botan/pubkey.h>
+#include <botan/rng.h>
+#include <botan/sha3.h>
+#include <botan/shake.h>
```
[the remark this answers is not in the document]
reneme: We replaced the bare-metal `absorb`/`expand` calls for the XOF by `SHAKE_128_Cipher`. That interface does the job just fine for now and hides quite a bit of iffy complexity. Adding a specific XOF interface would be a nice improvement, nevertheless.

src/lib/pubkey/kyber/kyber_common/kyber.cpp
randombit: Sounds good. I can do this in a new PR to master (or you can add one in this PR, either way is fine with me)

src/lib/pubkey/kyber/kyber.cpp
```diff
@@ -0,0 +1,1817 @@
...
+
+class PolynomialVector
+{
+  public:
+    PolynomialVector() = delete;
+    explicit PolynomialVector(const size_t k) : vec(k)
+    {
+    }
+
+    std::vector<Polynomial> vec;
```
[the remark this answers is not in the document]
reneme: `Polynomial`, `PolynomialVector`, `PolynomialMatrix` and `Ciphertext` are all tightly bound due to their interplay. For now they are internal to the kyber compilation unit (anonymous namespace), so this frankly didn't feel too 'dangerous'.

Future work in the scope of [the KBLS project](https://www.forschung-it-sicherheit-kommunikationssysteme.de/projekte/kbls) will introduce Dilithium whose [reference implementation](https://github.com/pq-crystals/dilithium) is based on the same (or similar) math helpers. We hope to be able to abstract those into a dedicated "crystals" module. From a quick glance, also NewHope uses a very similar mathematical base.

I suggest to re-visit the public members in the context of that refactoring, either with simple accessors or by befriending the classes as needed.

`m_` prefix was added, though.

src/lib/pubkey/kyber/kyber.cpp
```diff
@@ -0,0 +1,1817 @@
...
+        cipher->encrypt( out );
+    }
+
+    return out;
+}
+
+class Polynomial
+{
+  public:
+    std::array<int16_t, KyberConstants::N> coeffs;
```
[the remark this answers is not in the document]
reneme: Left it `public` for now as described in https://github.com/randombit/botan/pull/2872#discussion_r782138228.

src/lib/pubkey/kyber/kyber_common/kyber.h
randombit: Ironically there *used* to be a good way of doing this, a module could have an `A or B` dependency, which would work perfectly in this situation. It looks like I removed this in the otherwise unrelated commit ba8a26f17d - the functionality hadn't been used in some time at that point, and in that PR I was making other changes to dependency resolution to handle conditional dependencies based on OS features and apparently decided to remove `|` at the same time.

This current setup is certainly going to be confusing to end users. I'll see how hard it would be to add alternatives back, probably not bad. Then we would have `kyber` which would depend on `kyber_modern | kyber_90s`.

hrantzsch: Sounds good, if it's not too complicated to get alternatives back in. Would the module `kyber` default to the modern mode then? Or would users have to explicity enable `kyber,kyber_modern`?

Note that currently the `kyber` module is the modern mode and `kyber_90s` the 90s mode. So end users would need to explicitly only enable `kyber_common` to get a broken build. So naively using `kyber` would be fine. We hoped this would avoid confusion at least a bit.

reneme: As an alternative, we might take a look at creating "internal modules" (i.e. by adding some new keyword in `info.txt`) so that the current `kyber_common` module could be "required" by `kyber` and `kyber_90s` but wouldn't show up in the user-facing `./configure.py --list-modules`.

src/lib/pubkey/kyber/kyber_common/kyber.cpp
hrantzsch: we'll add it in a cherry-pickalbe commit on this branch

hrantzsch: it's b0025e10381afde1e1d9823c3f08886235a7bb85, in case it's needed before we merge this

reneme:
> Regarding OIDs: We'll switch to the ones mentioned in this IETF draft standard. Also, this document specifies an ASN.1 encoding for Kyber keys. Probably we should support those as well.

@randombit Regarding ASN.1 encoding: Do you think it makes sense to support the reference implementation's encoding (simple byte concatenation of parameters) as well? If yes, what's the preferred way of selecting which encoding should be read/emitted?

Further: Should there be constructors for `Kyber_PublicKey`/`Kyber_PrivateKey` that take raw algorithm-specific parameters? This might come in handy for users that want to interface with implementations that provide some other unsupported encoding. Or could this be left as future improvement?

@omarama @boricm Do you have some specific requirements here?

mouse07410:
I have some reservations about ASN.1 encoding from that document. Let me double-check and get back here.

boricm:
> > Regarding OIDs: We'll switch to the ones mentioned in this IETF draft standard. Also, this document specifies an ASN.1 encoding for Kyber keys. Probably we should support those as well.
> 
> @randombit Regarding ASN.1 encoding: Do you think it makes sense to support the reference implementation's encoding (simple byte concatenation of parameters) as well? If yes, what's the preferred way of selecting which encoding should be read/emitted?
> 
> Further: Should there be constructors for `Kyber_PublicKey`/`Kyber_PrivateKey` that take raw algorithm-specific parameters? This might come in handy for users that want to interface with implementations that provide some other unsupported encoding. Or could this be left as future improvement?
> 
> Edit: The (preliminary) ASN.1 spec defines [a partial and full encoding of the private key](https://www.ietf.org/id/draft-uni-qsckeys-00.html#section-4.3). Same questions arise for that.
> 
> @omarama @boricm Do you have some specific requirements here?

Thanks for finding this document, @reneme! Back when we thought about the API and IODs, this document was not yet published. That's the only reason why we didn't use the encoding and OIDs specified there.
@omarama and me would like to support both - the reference implementations encoding as well as the ASN.1 encoding from this document.

randombit:
Certainly would want to support any IETF-standardized ASN.1 encoding (even if suboptimal). No strong opinion on supporting the reference implementation encoding.

randombit:
Re the ASN.1 I would much rather they have just said "everything is a single OCTET STRING which encapsulates whatever the reference implementation uses", rather than reflecting the algorithm specific details in the ASN.1 for no real reason. I'll send a comment, once I figure out what WG this is going through...

mouse07410:
All the NIST PQC KEM candidates represent their public keys and ciphertexts as byte arrays - aka, direct making onto ASN.1 OCTET STRING. 

The referred IETF document is a proposal, a draft - not a standard - and in my opinion this draft is not good.

src/lib/pubkey/kyber/kyber.cpp
```diff
@@ -0,0 +1,1817 @@
...
+     * Name:        poly_tomont
+     *
+     * Description: Inplace conversion of all coefficients of a polynomial
+     *              from normal domain to Montgomery domain
+     *
+     * Arguments:   - poly *r: pointer to input/output polynomial
+     **************************************************/
+    void tomont()
+    {
+        const int16_t f = (1ULL << 32) % KyberConstants::Q;
```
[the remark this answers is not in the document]
reneme: We just changed `const` to `constexpr`. `KyberConstants::Q` is `constexpr` as well. Not sure what to call this, otherwise. :)

src/lib/pubkey/kyber/kyber.cpp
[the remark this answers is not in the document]
reneme: done.

hrantzsch:
We're done with all the improvements and refactorings we had in mind and discussed above. Ready for another round of review :)

boricm:
Hi @hrantzsch and @reneme!
@omarama and me noticed that you added a `random = ` line to the test vectors. Where did you get this random from? Did you capture the output of the kyber-rng (which was only used for the KAT-tests)?
We like the approach with the unmodified test vectors, as you can simply add new vectors if the NIST provides them. What are your thoughts on this?

reneme:
> Where did you get this random from? Did you capture the output of the kyber-rng (which was only used for the KAT-tests)?

Yes, exactly. It's the random string produced by the RNG that used to be implemented in the KAT tests. We opted for this "static" approach instead of shipping an entire RNG implementation. See also @randombit's comment [here](https://github.com/randombit/botan/pull/2872#discussion_r774677831).

> We like the approach with the unmodified test vectors, as you can simply add new vectors if the NIST provides them. What are your thoughts on this?

If NIST provides more test vectors, we can do the same approach and hard-code the random outputs, no? Its a bit of extra work up-front but saves the additional homebrew RNG in the code base.

mouse07410:
With the misguided API that NIST enforced, I think keeping a homebrew NIST-specified (P)RNG in the code is the only way to deal with KAT. 😟

src/lib/utils/compiler.h
```diff
@@ -155,7 +155,27 @@
...
 
+/*
+* Define Botan::unreachable()
+*
+* There is a pending WG21 proposal for `std::unreachable()`
+*   http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0627r3.pdf
+*/
+namespace Botan {
+
+[[noreturn]] BOTAN_FORCE_INLINE void unreachable()
```
randombit, line 168: This has to either be a macro, or defined somewhere else - `compiler.h` is included into `build.h` which is included into `ffi.h`, which needs to be parseable by a C89 compiler.

[There should be a CI test for this]

reneme: Fair point. I wasn't aware of this. Moved it to assert.h

src/lib/pubkey/kyber/kyber_common/kyber.cpp
```diff
@@ -0,0 +1,1496 @@
...
+         // Overwrite pre-k with z on re-encryption failure (constant time)
+         secure_vector<uint8_t> lower_g_out_final;
+         if(constant_time_compare(encap_key, cmp.data(), len_encap_key))
+            {
+            std::copy(lower_g_out.begin(), lower_g_out.end(), std::back_inserter(lower_g_out_final));
+            }
+         else
+            {
+            std::copy(m_key.m_private->z().begin(), m_key.m_private->z().end(), std::back_inserter(lower_g_out_final));
+            }
```
randombit, line 1228: This will be constant time, but still falls to a cache based side channel that detects if `z` is read from or not.

I would suggest something like

```
lower_g_out_final.resize(lower_g_out.size());
CT::conditional_copy_mem(lower_g_out_final.data(), lower_g_out_final.size(), lower_g_out.data(), m_key.m_private->z().data());
```

randombit (approved):
One side channel issue that should be resolved before merge, but besides that this is looking really good.

I have not had time to work on the modules issue where we would like have the Kyber module require that at least one implementation be available at build time. I don't think that should hold up this merging.

Since the Kyber specification is still somewhat subject to change I would be against backporting this to the 2.x branch at least until NIST announces something (which I have heard should be by end of March).

src/lib/pubkey/kyber/kyber_common/kyber.cpp
hrantzsch: Thanks, we were not aware of that.
We spotted another bug nearby, where we didn't actually use `lower_g_out_final`. See [ded2a96](https://github.com/randombit/botan/pull/2872/commits/ded2a969bfa6ac9b2c4da50a2c6720dbf20c53e2).

PR #2873, opened 2021-12-23
SystemRNG update for macOs.

devnexen:
From Yosemite, there is a new native API, the little potential issue
 with arc4random is possible errors (even tough most of the time
unlikely happening) are ignored.

The `__builtin_available` would be another way to detect its availability
if preferred.


randombit:
10.10 was already 7 years ago so fine to start relying on it, someone in the rare case needing to build for something that old can disable it at build time.

PR #2878, opened 2022-01-12
Check OSXSAVE and AVX flags before enabling AVX2 support.

bluhm:
OpenBSD on i386 architecture does not save the extended FPU state
and does not set this feature in the CPU.  In this case AVX2
instructions cause an illegal instruction exception.  Botan must
also check the flags set by the OS in the CPU.


randombit:
Thanks!

PR #2891, opened 2022-01-30
Create build_docs.py

niclasr:
Find rst2man from pkgsrc.
Pkgsrc installs rst2man as rst2man-${python_version} (rst2man-3.9).


niclasr:
Wrong commit message

PR #2892, opened 2022-01-30
Find rst2man from pkgsrc

niclasr:
Find rst2man from pkgsrc.
Pkgsrc installs rst2man as rst2man-${python:version} (rst2man-3.9).


randombit:
Thanks!

PR #2893, opened 2022-02-01
Text_Policy ignores 25519 if unsupported by build

hrantzsch:
While introducing a new `key_exchange_groups_to_offer` field to TLS policies, we discovered an inconsistency when `Text_Policy` deals with groups that aren't supported by the Botan build.

Previously, curve25519 would end up in the policy's `supported_groups` list, even if the module was not enabled. Using a text policy that includes x25519, this could lead to TLS clients offering this group, even if it isn't compiled into the library.

We changed the Text_Policy's behavior to (quietly) ignore such groups.

This also removes the need for a hack in the `tls_policy_text` test.


randombit:
Looks good!

Worth backporting to the 2.x release branch?

PR #2894, opened 2022-02-01
Text_Policy ignores 25519 if unsupported by build

hrantzsch:
Backporting #2893


reneme:
It failed on on the windows "sanitizer" platform while running `scrypt` tests with an error that seems unrelated and flaky. @randombit could you re-trigger? Or can we trigger a build ourselves somehow?

Issue #2885, opened 2022-01-24
[pkcs11] derive secret key with hardware private key


ncoder-1:
Hi,

I'm trying to derive a secret key from a private key stored on a pkcs11 token and a generated ecdh public key. Unfortunately, I'm doing something wrong as I'm getting the following error:

```
terminate called after throwing an instance of 'Botan::PKCS11::PKCS11_ReturnError'
  what():  PKCS11 error 113
```

I'm able to login onto the token and find the private key ojbect. I'm also able to do the opposite and derive a secret key from a generated private key and a token-stored public key. Is my `PKCS11_ECDH_PrivateKey` object improperly referenced or missing the attributes required to derive?

This works:

```cpp
Botan::PKCS11::AttributeContainer pubkey_search_template;
pubkey_search_template.add_string(Botan::PKCS11::AttributeType::Label, "Public key for Key Management");
auto found_objs = Botan::PKCS11::Object::search<Botan::PKCS11::Object>(session, pubkey_search_template.attributes());
if (found_objs.size() != 1) return 1;

Botan::PKCS11::PKCS11_ECDH_PublicKey hsm_pub_key(session, found_objs.at(0).handle());

Botan::AutoSeeded_RNG rng;

Botan::ECDH_PrivateKey privkey_ecdh(rng, Botan::EC_Group("secp384r1"));

Botan::PK_Key_Agreement ecdhKA(privkey_ecdh, rng, "Raw");

Botan::SymmetricKey secret_key_A = ecdhKA.derive_key(48, hsm_pub_key.public_point().encode(Botan::PointGFp::UNCOMPRESSED));

std::cout << "Secret Key A: " << '\n' << secret_key_A.to_string() << '\n';
```

This doesn't:

```cpp
Botan::PKCS11::AttributeContainer privkey_search_template;
privkey_search_template.add_string(Botan::PKCS11::AttributeType::Label, "Private key for Key Management");
auto found_objs_priv = Botan::PKCS11::Object::search<Botan::PKCS11::Object>(session, privkey_search_template.attributes());
if (found_objs_priv.size() != 1) return 1;
  
Botan::PKCS11::PKCS11_ECDH_PrivateKey hsm_priv_key(session, found_objs_priv.at(0).handle());

Botan::AutoSeeded_RNG rng;

Botan::ECDH_PrivateKey privkey_ecdh(rng, Botan::EC_Group("secp384r1"));

[...]

PR #2895, opened 2022-02-03
Allow setting the public point encoding of a P11 ECC private key

reneme:
See GH #2885 for further details.

@ncoder-1 Would that work for you? This new method simply allows setting the preferred public point encoding on a PKCS#11 private key without providing its associated public key. Nevertheless, you mentioned, that you ran into trouble with the `Public point not set. Inferring the public key from a PKCS#11 ec private key is not possible.` error. Even though I don't see why the key agreement would try to access the private key's public point, I'd like to make sure that I'm not missing something.

For the record: I still find this interface rather awkward and would prefer setting the preferred public point encoding along with the actual key agreement parameters. However, the high-level `PK_Key_Agreement` class doesn't really provide a means to configure such things for the user. Maybe @randombit has a better idea regarding this new API?

### DIFF
--- a/src/lib/prov/pkcs11/p11_ecc_key.h
+++ b/src/lib/prov/pkcs11/p11_ecc_key.h
@@ -176,6 +176,16 @@ class BOTAN_PUBLIC_API(2,0) PKCS11_EC_PrivateKey : public virtual Private_Key,
          }
 
       /**
+       * Sets the public desired public point encoding of this private key, when it is passed to cryptoki functions.
+       * This could be either `PublicPointEncoding::Raw` or `PublicPointEncoding::Der`. By default this is set to `Der`,
+       * but some tokens might expect `Raw`-encoded public keys, e.g. when using this private key for key agreement.
+       */
+      void set_point_encoding(PublicPointEncoding point_encoding)
+         {
+         m_point_encoding = point_encoding;
+         }
+
+      /**
       * Gets the public_point
       * @note the public key must be set using `set_public_point`
       *       because it is not possible to infer the public key from a PKCS#11 EC private key


ncoder-1:
I just tested a build with the patch, sure enough all I needed was this:

`hsm_priv_key.set_point_encoding(Botan::PKCS11::PublicPointEncoding::Raw);`

I would say that it fixes my issue, but I'm somewhat concerned that it's very close to the already existing `set_public_point()` and seems like it could cause some confusion by having 2 functions nearly named the same. Both functions exist to set the public point encoding, one requires passing the actual public points, the other doesn't.

Also, like I mentioned in GH [#2885](https://github.com/randombit/botan/issues/2885), it seems odd that the default would be `::Der` and that we'd have to convert to `::Raw`. I actually think it should be the other way around, `::Raw` and if required, convert to `::Der` via `set_point_encoding()`.

That said, it does work and now I have a proper workaround. I'd be curious to see what @randombit thinks.

randombit:
I'm fine with this PR in the sense that it is simple and seems to solve the problem. But it is pretty awkward; going from "I'm getting this weird PKCS11 error" to "I need to write code to find the peer ECDH key in the store and set this flag on it" is non-obvious (unless a search happens to surface this discussion).

I'm wondering if there is some way we can query the card to see what format it expects the other key in. Otherwise how would any ECDH app using PKCS11 work without having to potentially retry the operation, first with the raw key then with DER encoding as a fallback?

> However, the high-level PK_Key_Agreement class doesn't really provide a means to configure such things for the user.

Re this, there *should* be some mechanism to provide some potentially unbounded set of parameters when creating a pubkey operation object. Currently the configuration is a pair of strings (`params` and `provider`). Changing these to be some general struct which can take many possible parameters would be nice as it would give an easy way to extend to handle cases like this.

randombit:
Somewhat related issue - looking at https://botan.randombit.net/handbook/api_ref/pkcs11.html#tested-supported-smartcards I see that a number of tests were failing against CardOS hardware with "Invalid argument OS2ECP: Unknown format type <random number>" I bet in these cases we are getting back DER encoding instead of raw point.

PR #2905, opened 2022-02-08
Update GMAC code example

ncoder-1:
Key is required again to verify, otherwise the following error occurs:
terminate called after throwing an instance of 'Botan::Key_Not_Set'
  what():  Key not set in AES-256


randombit:
This shouldn't be required. I think this is a bug in GMAC.

randombit:
Bug in GMAC - #2908

Issue #2714, opened 2021-04-16
Support for the TLS 1.3 - development planned


pist-eb:
We'd like to contribute to Botan by starting the implementation of the TLS 1.3 protocol support.
The motivation is that our use-cases require the TLS 1.3 from the client side and therefore we could support the development in that topic. The plan is to start in May (or even earlier with rough design and development plan).

The rough plan:
- design draft, to agree the direction of development with the community,
- refactoring of the existing TLS code, i.e. splitting the common code and the protocol version specific, introducing pimpl idiom to have generic interfaces and version specific implementation, etc.,
- verification of refactored code to keep the quality on high level and avoid any regression,
- start the development of TLS 1.3 protocol,
- next steps...

The important thing would be to select proper start point (version) for the development. It seems that the release 2 (with the latest 2.18) is already mature and development on master branch is probably meant for upcoming 3.x release.
Is there any rough plan, when first 3.x version will be released, as we would like to pick up some 'official tag' for the development?

Please, comment whether this topic sounds interesting and implementing the TLS 1.3 support would be valuable here?

randombit:
Yes I'd be very happy to accept patches adding TLS 1.3 support and would work with you to make this happen.

This work should happen on master branch for 3.0 release. 2.x at this point is effectively bug fix only, or at most backporting small features that are merged first onto master. Adding 1.3 support to master would probably be a bit simpler as we've shed a number of obsolete or obscure features (TLS 1.0, SRP, DSA support, etc) in master, as compared to 2.x which still retains them. There isn't a particular tag to pick up but, modulo adding TLS 1.3 support, I don't anticipate further major changes to the TLS code at this point, as the feature removals that are going to happen for 3.0 have already occurred on master.

There is no explicit schedule for 3.0.0 final release. I am planning on doing an initial alpha for community preview purposes in the next month or so. The final release will happen when it is ready (but I would certainly hope sometime this year).

cc @reneme @securitykernel

pist-eb:
Thank you. We'll then pick up the master for development.
I'll share some design concepts, plans here when we start.

pist-eb:
@randombit: We need to pick up the Botan version to start working on the TLS 1.3 protocol implementation. As agreed, master branch shall be used. Due to the formal reasons, it would be good to have any official tag to start with and to have some reference for that starting point later. When do you expect to release mentioned 3.0.0 preview/alpha version?

[...]

PR #2922, opened 2022-02-24
[TLS 1.3] Limited TLS 1.3 Client Implementation

hrantzsch:
# Limited TLS 1.3 Client

## Pull Request Dependencies

 * The RFC 8448 based test uses the `Test::Result` consolidation constructor from [this pull request](https://github.com/randombit/botan/pull/2967). For simplicity the required commit is duplicated here and should be removed after the consolidation constructor is finalized and merged.

### Follow-up Pull Requests

The number of open pull requests that are based on this are getting somewhat out of hand. Below is a list to keep track of the dependencies:

   * **THIS ONE:** https://github.com/randombit/botan/pull/2922
      * https://github.com/randombit/botan/pull/2957
      * https://github.com/randombit/botan/pull/2974
      * https://github.com/randombit/botan/pull/2977
      * https://github.com/randombit/botan/pull/2983
      * https://github.com/randombit/botan/pull/2988
      * https://github.com/randombit/botan/pull/2989
      * **INFORMATIVE:** https://github.com/randombit/botan/pull/2981
      * **INFORMATIVE:** https://github.com/randombit/botan/discussions/2954

### Potential Future Work

This is a non-exhaustive list of potential improvements or further development. Note that those points are not tackled by the follow-up PRs. I'll leave this here for future reference.

* [ ] **Overhaul of `TLS_Data_Reader`**
   * Some of its methods leave room for performance improvements
   * The interface could be easier to use
   * More invariant checks (e.g. assert `!has_remaining()` on destruction).
   * Also a "sub-reader" concept: i.e. a new `::get_tls_length_value_as_reader()` could return a sub-reader that can then be passed into a sub-parser (e.g. Client Hello parsing its extensions)
   * Use it consistently for parsing
* [ ] **Allow disabling TLS 1.2 at compile time** 
* [ ] **`std::span` to avoid copying**
   * C++17 doesn't actually provide it, but a custom implementation should be straight forward
* [ ] **Separate TLS 1.2 and 1.3 cipher suites**
   * TLS 1.3 suites are more focussed
     (some protocol parameters are negotiated differently)
   * the `Ciphersuite` class provides methods that are strictly not defined for TLS 1.3 (potential cause for bugs)
   * cipher suite handling code becomes more convoluted that necessary
     (e.g. `Policy::ciphersuite_list()`, generation via `tls_suite_info.py`)

## This PR adds limited TLS 1.3 support. Available functionality:

 * functional TLS 1.3 client
     * automatically downgrades to TLS 1.2 if allowed and required by the server
     * usable via Botan-CLI `tls_client`
 * additional extensions as required per RFC 8446:
     * Key_Share
     * Supported_Versions
     * Cookie
     * Signature Algorithms
     * Signature Algorithms Certificate
     * Negotiated Groups
 * additional extensions (not strictly required by RFC 8446)
     * Certificate Status Request (OCSP stapling)
     * Record_Size_Limit (RFC 8449)
     * Application Layer Protocol Notification (RFC 7301)
     * Server Name Indication
 * Utility and Misc
     * `./botan tls_client --debug` to print raw TLS traffic
     * `Fixed_Output_RNG` can optionally fall back to another RNG when its pool is empty
 * Tests integrated or performed
     * New infrastructure has unit tests
     * Integration tests using the test vectors from RFC 8448
     * BoGo tests that do not depend on not-yet-implemented protocol features
       (see [here](https://github.com/randombit/botan/pull/2922/commits/7967664f730cb015f8d4e6ddb93ec3bb6d4e7491) for a list of issues found by BoGo)
     * Handshake and data exchange works using `./botan tls_client` with the big players (google.com, cloudflare.com, ...)

## Limitations:

 * TLS 1.3 support cannot be compiled without enabling TLS 1.2
   (TLS 1.2 can be disabled in the TLS policy though)
 * no TLS 1.3 compatible server
 * no client authentication *(added in: https://github.com/randombit/botan/pull/2957)*
 * no session resumption via Tickets *(added in: https://github.com/randombit/botan/pull/2974)*
 * no Pre-Shared-Keys *(technical foundation added in: https://github.com/randombit/botan/pull/2974)*
 * no 0-RTT support (early data) *(technical foundation added in: https://github.com/randombit/botan/pull/2974)*
 * no DTLS
 * no fuzzing of the TLS 1.3 implementation *(basic fuzz target for parser: https://github.com/randombit/botan/pull/2977)*

# Implementation Overview

To make this large PR a bit more approachable, we'll provide an overview to the newly introduced components and their responsibilities. Note that this work-in-progress pull request strives towards a usable implementation of "Minimal Viable Product" as outlined in the ToDo-List of this [issue comment](https://github.com/randombit/botan/issues/2714#issuecomment-1031621985).

## TODO

* [x] server certificate validation
    * [x] OCSP stapling
* [x] Hello Retry Request
* [x] handle `CLOSE_NOTIFY` properly (difference between TLS 1.2 and 1.3)
* [x] protocol version downgrade
* [x] Key Update
* [x] Key Material Export (RFC 8446 7.5 / RFC 5705)
* [x] review TLS callbacks invocations
* [x] bogo test suite integration
* [x] check occurrences of extensions as indicated in [the table in RFC 8446 4.2](https://www.rfc-editor.org/rfc/rfc8446.html#page-37)
* [x] configure ClientHello settings (e.g. ALPN)
* [x] honor the record size limit extension
* ~~Session Resumption (PSK w/o 0-RTT)~~ not part of this PR

## Intro

The descriptions mostly refer to the components in the `tls13` module, as the TLS 1.2 implementation remains unchanged as far as possible. As initiated in the refactoring by Elektrobit Automotive GmbH, the user-facing interface of `TLS::Client` and `TLS::Server` also remains unchanged by means of a Pimpl-pattern.

We nevertheless anticipate some API changes in the public API. Please refer to [this issue comment](https://github.com/randombit/botan/issues/2714#issuecomment-1057175631) for further details.

## Components

Screenshot 2022-03-04 at 12 18 24
https://excalidraw.com/#json=uJjsg_me6QrHsybCbkgxs,Ql21IMba1ZERgxfIiFElKg

### Channel

The Channel acts as "composition root" of the entire TLS 1.3 implementation. It implements the library's public APIs via a Pimpl-construction as detailed in [this issue comment](https://github.com/randombit/botan/issues/2714#issuecomment-864965990).
This class implements the client/server agnostic parts of the protocol. Hence, orchestrating the `Record_Layer` and `Handshake_Layer` to transform received bytes from the network into handshake messages to be interpreted by the `Client` (and later `Server`) implementations or application data to be passed to the using code. Furthermore, it takes care of handling TLS alerts.

#### Record_Layer

This layer implements the record layer level of the protocol. It parses raw data from the wire to individual records and decrypts protected records using `Cipher_State`. This corresponds to tasks performed by the Channel (`Channel_Impl_12`) and the `Record_Header` class in the 1.2 implementation.
When sending data it add record headers and encryption to the records.

#### Handshake_Layer

This class takes care of parsing and marshalling of `Handshake_Messages` as well as equipping them with the appropriate handshake protocol headers. As this class handles the byte-representation of all TLS handshake messages (i.e. before parsing or after serialization), it is responsible for updating the `Transcript_Hash_State` appropriately.

In TLS 1.2 the task of the `Handshake_Layer` were mostly implemented in the `Handshake_IO` class. Note that `Handshake_Layer` does not perform any handshake state validations. It simply parses/serializes messages as they come in from the wire or the downstream protocol implementation.

#### Cipher_State

This class implments the Key Schedule mechanism. Most importantly, it derives and holds traffic secrets and provides interfaces for the `Record_Layer` protect and deprotect records. The `Client` advances the `Cipher_State` through the Key Schedule "state machine".

#### Transcript_Hash_State

This class keeps track of the transcript hash while sending/receiving messages in the `Handshake_Layer`. `Client` (and later `Server`) consult the `Transcript_Hash_State` for relevant hash-data when updating the `Cipher_State` appropriately.

### Client

The Client's main responsibility continues to lie in `process_handshake_message`, implemented via individual `handle`-methods for each specific message type. Hence, the client's TLS state machine is implemented here. During message processing, it updates the state of other components:
 * advancing the Key Schedule in `Cipher_State`
 * consult the `Transcript_Hash_State`
 * manage expected next messages in `Handshake_Transitions`

#### Handshake_Transitions

This class aids the Client implementation in validating state transitions. It is merely an extraction of the `confirm_transition_to` and `set_expected_next` functionality from TLS 1.2's `Handshake_State`.

#### Handshake_State

`HandshakeState` manages the incoming and outgoing messages and keeps a record of them for later reference. Before storing the messages it filters them regarding the `Outbound_Message` and `Inbound_Message` variants. Hence, a misuse of `Handshake_State` will fail to compile (e.g. when a client implementation tries to send a `Server_Hello_13`). Similarly, the compiler can check that `Client` implements `handle()` methods for _exactly_ the relevant handshake messages.

### Handshake_Messages

The class hierarchy of `Handshake_Messages` implements the specifics of individual handshake messages. In the TLS 1.3 implementation we don't rely on the actual polymorphic nature of `Handshake_Message` but replace it with `std::variant` constructions to specifically define which handshake message types are expected where.

Each handshake message takes care of "local" protocol semantic checks. I.e. all validations that can be performed without contextual information is happening right in the parsing code. The same holds true for handshake message extensions (e.g. `Key_Share`, `Supported_Groups`, ...). Protocol logic that is tied to specific messages or extensions is implemented locally in those classes. For instance, `Key_Share` implements Diffie-Hellman in its `exchange` method and `Client_Hello_13::retry()` implements necessary updates and validations for handling "hello retry requests".

This is a different paradigm compared to the messages in TLS 1.2, where many messages where mere "Data Transfer Objects" without much logic. Hence, messages derive a `_12` and `_13` sub-class and share the parsing code in the parent class where possible. This is also useful in situations where the interface and usage of the same message type differs between TLS 1.2 and TLS 1.3, but the wire representation is unchanged.


.github/workflows/ci.yml
```diff
@@ -10,9 +10,9 @@ permissions:
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, dev/tls-13 ]
```
hrantzsch, line 13: This was only needed while we worked from a PR _towards_ `dev/tls-13` and should be removed now.

src/lib/tls/msg_cert_req.cpp
```diff
@@ -70,11 +76,11 @@ Certificate_Req::Certificate_Req(const std::vector<uint8_t>& buf)
 
    TLS_Data_Reader reader("CertificateRequest", buf);
 
-   const auto cert_type_codes = reader.get_range_vector<uint8_t>(1, 1, 255);
+   std::vector<uint8_t> cert_type_codes = reader.get_range_vector<uint8_t>(1, 1, 255);
 
-   for(const auto cert_type_code : cert_type_codes)
+   for(size_t i = 0; i != cert_type_codes.size(); ++i)
```
hrantzsch, line 81: revert: clang-tidy fix was lost during rebase

randombit:
@hrantzsch @reneme Guys so sorry about how long review is taking on my side, work is kind of a death march at the moment. I will see what I can do over this weekend.

hrantzsch:
Status update: the "MVP" is now pretty much done and even usable. We started testing using the current (TLS 1.2) Bogo configuration. Some of those tests downgrade to TLS 1.2 and continue to (successfully) test the 1.2 implementation, some succeed without downgrading, and a number of tests (specifically those testing NYI features like resumption) will fail.

Next week we will look into the remaining Bogo test failures and tick the other minor items in the TODO list.

src/cli/tls_utils.cpp
```diff
@@ -140,7 +140,8 @@ class TLS_Client_Hello_Reader final : public Command
 
          try
             {
-            Botan::TLS::Client_Hello hello(input);
+            // TODO: deal with Client_Hello_13
```
hrantzsch, line 143: - [ ] allow printing `Client_Hello_13` infos in cli utils

We'll need to allow deserializing a Client Hello w/o knowing if it'll be 1.2 or 1.3, just as we'll eventually need to do for the TLS 1.3 server.

src/fuzzer/tls_client_hello.cpp
```diff
@@ -12,7 +12,7 @@ void fuzz(const uint8_t in[], size_t len)
    try
       {
       std::vector<uint8_t> v(in, in + len);
-      Botan::TLS::Client_Hello ch(v);
+      Botan::TLS::Client_Hello_12 ch(v);  // TODO: We might want to do that for TLS 1.3 as well
```
hrantzsch, line 15: - [ ] fuzz `Client_Hello_13`

src/lib/tls/asio/asio_stream.h
```diff
@@ -716,7 +716,7 @@ class Stream
                                 m_context.m_policy,
                                 m_context.m_rng,
                                 m_context.m_server_info,
-                                Protocol_Version::latest_tls_version()));
+                                Protocol_Version::TLS_V12));  // TODO don't hardcode
```
hrantzsch, line 719: - [ ] allow using asio stream (client) with TLS 1.3

src/lib/tls/msg_cert_req.cpp
```diff
@@ -1,6 +1,7 @@
 /*
 * Certificate Request Message
 * (C) 2004-2006,2012 Jack Lloyd
+*     2021 Elektrobit Automotive GmbH
```
hrantzsch, line 4: - [ ] add all the copyright headers

src/scripts/ci/setup_gh_actions.sh
```diff
@@ -53,7 +53,8 @@ if type -p "apt-get"; then
         pip install --user codecov
         echo "$HOME/.local/bin" >> "$GITHUB_PATH"
 
-        git clone --depth 1 --branch jack/runner-20210401 https://github.com/randombit/boringssl.git
+        # TODO: merge changes to Botan's boring fork
+        git clone --depth 1 --branch rene/runner-20220322 https://github.com/reneme/boringssl.git
```
hrantzsch, line 57: - [ ] merge BoGo runner changes

src/scripts/tls_suite_info.py
```diff
@@ -227,7 +244,7 @@ def main(args = None):
                 if ns in name:
                     should_use = False
 
-            if should_use and name.find('_WITH_') > 0:
+            if should_use:# and name.find('_WITH_') > 0:
```
hrantzsch, line 247: is this right?

randombit:
I'm working on a review of this now. It will take a while and we should probably conference on the design but only after I've had a chance to look at everything.

I would suggest as an intermediate step, we should try to reduce the overall size of this PR. You've already done that with the test code, I'd suggest similarly creating split out PRs for

* The comments in `pubkey.h`
* The handshake state transition tracker
* `tls_reader.h` additions
* `tls_alert.*` changes (I know this alert is TLS 1.3 only but no reason we can't recognize it immediately)

Ideally we'd also do the pimpl changes for TLS 1.2 only in a separate PR as well. I don't know how easy that is to do at this point. If it is easy then please do it as that is largely renaming and would be fast to review.

reneme:
FWIW: The diff statistics to #2946 look "promising" indeed. Almost only additions and about 1/3 test code 2/3 implementation.

```
> git diff --stat tls13/pimpl_refactoring | tail -n1
95 files changed, 10032 insertions(+), 78 deletions(-)

> git diff --stat tls13/pimpl_refactoring -- src/tests | tail -n1
 32 files changed, 3879 insertions(+), 7 deletions(-)

> git diff --stat tls13/pimpl_refactoring -- src/lib/tls | tail -n1
 54 files changed, 5796 insertions(+), 21 deletions(-)
```

reneme:
Rebased again after the confusion over #2953.

src/lib/tls/tls_extensions_key_share.cpp
```diff
@@ -0,0 +1,553 @@
...
+            m_private_key = std::move(skey);
+#if defined(BOTAN_HAS_CURVE_25519)
+            }
+         else if(is_x25519(group))
+            {
+            auto skey = std::make_unique<X25519_PrivateKey>(rng);
+            m_key_exchange = skey->public_value();
+            m_private_key = std::move(skey);
+#endif
+            }
```
hrantzsch, line 117:
```suggestion
            }
#if defined(BOTAN_HAS_CURVE_25519)
         else if(is_x25519(group))
            {
            auto skey = std::make_unique<X25519_PrivateKey>(rng);
            m_key_exchange = skey->public_value();
            m_private_key = std::move(skey);
            }
#endif
```
This is more readable

src/lib/tls/tls13/tls_handshake_layer_13.cpp
```diff
@@ -0,0 +1,197 @@
...
+         case CERTIFICATE:
+            return Certificate_13(msg, policy, peer_side);
+         // case CERTIFICATE_REQUEST:
+         //    return Certificate_Req_13(msg);
+         case CERTIFICATE_VERIFY:
+            return Certificate_Verify_13(msg, peer_side);
+         case FINISHED:
+            return Finished_13(msg);
+         default:
+            BOTAN_ASSERT(false, "cannot be reached"); // make sure to update handshake_type_from_byte
```
hrantzsch, line 102: If the messages commented above are received, this _will_ be reached. Handle NYI messages better.

hrantzsch, line 102: If the messages commented above are received, this _will_ be reached. Handle NYI messages better.

src/lib/tls/msg_cert_verify.cpp
```diff
@@ -97,6 +96,135 @@ bool Certificate_Verify_12::verify(const X509_Certificate& cert,
...
+
+Certificate_Verify_13::Certificate_Verify_13(const std::vector<uint8_t>& buf,
+      const Connection_Side side)
+   : Certificate_Verify(buf)
+   , m_side(side) {}
+
+namespace {
+
+std::pair<std::string, Signature_Format>
+parse_sig_format(const std::string& key_type,
```
hrantzsch, line 119: This is duplicated from handshake state and needs cleanup.

src/lib/tls/tls13/tls_handshake_layer_13.cpp
reneme: Fixed in https://github.com/randombit/botan/pull/2922/commits/8976bf250a3e7c299a4b36485f0328a843f2627e by commenting the culprit message types in `handshake_type_from_byte()` (invoked above) resulting in an UNEXPECTED_MESSAGE alert instead of failed assertion here.

The unimplemented messages are currently not negotiated in the Client Hello, so a compliant server will not send them. Except the server might still request client authentication (sending a Certificate Request message) which we would ungracefully reject right now. Client authentication will come in the next PR today or tomorrow, though.

reneme: See https://github.com/randombit/botan/pull/2922#discussion_r849104658

src/lib/tls/asio/asio_stream.h
reneme: Just reverting this hunk was fine and resulted in a working asio wrapper.

src/lib/tls/tls12/tls_handshake_state.cpp
```diff
@@ -374,13 +386,36 @@ Handshake_State::parse_sig_format(const Public_Key& key,
...
+   //   RSA signatures MUST use an RSASSA-PSS algorithm, regardless of whether
+   //   RSASSA-PKCS1-v1_5 algorithms appear in "signature_algorithms".
+   if(version() == Protocol_Version::TLS_V13 && key_type == "RSA" &&
+         (scheme == Signature_Scheme::RSA_PKCS1_SHA1
+          || scheme == Signature_Scheme::RSA_PKCS1_SHA256
+          || scheme == Signature_Scheme::RSA_PKCS1_SHA384
+          || scheme == Signature_Scheme::RSA_PKCS1_SHA512))
+      {
+      throw TLS_Exception(Alert::ILLEGAL_PARAMETER, "RSA signatures must use an RSASSA-PSS algorithm");
+      }
```
reneme, line 417: This hunk should be removed as `version() == Protocol_Version::TLS_V13` won't ever be true. This is a left-over from when the premature TLS 1.3 code still depended on this handshake state implementation.

src/lib/tls/msg_certificate_13.cpp
```diff
@@ -196,20 +196,8 @@ Certificate_13::Certificate_13(const std::vector<uint8_t>& buf,
...
-      {
-      append_tls_length_value(entries, entry.certificate.BER_encode(), 3);
-      append_tls_length_value(entries, entry.extensions.serialize(m_side), 2);
-      }
-
-   append_tls_length_value(buf, entries, 3);
-
-   return buf;
+   // Needed only for server implementation or client authentication
+   throw Not_Implemented("NYI");
```
reneme, line 200: Removed the implementation as it was not covered by unit tests and is unused as of now. Removing those lines carried the coverage over the line and the build is now finally green. 😃

reneme:
I rebased this onto `master` and removed the merge commit that was introduced by merging https://github.com/randombit/botan/pull/2968 in here. Come to think of it, it would probably have been wiser to apply this to `master` straight away. :-/

randombit (commented):
Here are some initial comments. I'm going to continue with a more in-depth review but wanted to give some feedback. Overall looks very good.

src/lib/tls/msg_client_hello.cpp
```diff
@@ -3,44 +3,53 @@
...
 std::vector<uint8_t> make_hello_random(RandomNumberGenerator& rng,
                                        const Policy& policy)
    {
    std::vector<uint8_t> buf(32);
    rng.randomize(buf.data(), buf.size());
 
-   auto sha256 = HashFunction::create_or_throw("SHA-256");
-   sha256->update(buf);
-   sha256->final(buf);
+   if(policy.hash_hello_random())
```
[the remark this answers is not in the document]
reneme: So that we can disable it for the RFC 8448 test.

src/lib/tls/msg_server_hello.cpp
```diff
@@ -408,4 +472,197 @@ std::vector<uint8_t> Server_Hello_Done::serialize() const
...
+      }
+
+   // RFC 8446 4.1.3
+   //    legacy_compression_method:  A single byte which MUST have the value 0.
+   if(compression_method() != 0x00)
+      {
+      throw TLS_Exception(Alert::DECODE_ERROR, "compression is not supported in TLS 1.3");
+      }
+
+   const auto& exts = extensions();
```
randombit, line 541: This is just introduced for one usage. Simpler to write as `if(!extensions().has<Supported_Versions>())`

src/lib/tls/msg_server_hello.cpp
```diff
@@ -408,4 +472,197 @@ std::vector<uint8_t> Server_Hello_Done::serialize() const
...
+   // RFC 8446 4.1.3
+   //    The ServerHello MUST only include extensions which are required to
+   //    establish the cryptographic context and negotiate the protocol version.
+   //    [...]
+   //    Other extensions (see Section 4.2) are sent separately in the
+   //    EncryptedExtensions message.
+   //
+   // Note that further validation dependent on the client hello is done in the
+   // TLS client implementation.
+   std::set<Handshake_Extension_Type> allowed =
```
randombit, line 579: `const`

src/lib/tls/tls_policy.cpp
```diff
@@ -563,6 +619,14 @@ std::vector<std::string> Strict_Policy::allowed_key_exchange_methods() const
    }
 
 bool Strict_Policy::allow_tls12()  const { return true;  }
+bool Strict_Policy::allow_tls13()  const
+{
+#if defined(BOTAN_HAS_TLS_13)
+   return true;
+#else
+   return false;
+#endif
```
randombit, line 628: Should this just inherit the default Policy impl instead?

src/tests/test_tls_transcript_hash_13.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+#include <array>
+
+#include <botan/internal/tls_transcript_hash_13.h>
+
+using namespace Botan::TLS;
+
+namespace {
+
+using Test = Botan_Tests::Test;
```
randombit, line 20: Should be in `namespace Botan_Tests` instead

src/tests/test_tls_signature_scheme.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+#if defined(BOTAN_HAS_TLS)
+
+#include <botan/tls_signature_scheme.h>
+
+#include <botan/ec_group.h>
+
+namespace {
+
+using Test = Botan_Tests::Test;
```
randombit, line 18: `namespace Botan_Tests {`

src/tests/test_tls_record_layer_13.cpp
```diff
@@ -0,0 +1,978 @@
...
+#include <botan/tls_magic.h>
+#include <botan/internal/stl_util.h>
+#include <botan/internal/tls_reader.h>
+#include <botan/internal/tls_cipher_state.h>
+#include <botan/tls_exceptn.h>
+#include <botan/tls_ciphersuite.h>
+
+#include <botan/internal/tls_record_layer_13.h>
+
+namespace {
```
randombit, line 21: Should be in `Botan_Tests` namespace

src/lib/tls/msg_cert_verify.cpp
```diff
@@ -93,9 +100,140 @@ bool Certificate_Verify::verify(const X509_Certificate& cert,
...
+   const std::string hash_algo = hash_function_of_scheme(scheme);
+
+   // RFC 8446 4.4.3:
+   //   The SHA-1 algorithm MUST NOT be used in any signatures of
+   //   CertificateVerify messages.
+   if(scheme == Signature_Scheme::RSA_PKCS1_SHA1
+         || scheme == Signature_Scheme::ECDSA_SHA1
+         || scheme == Signature_Scheme::DSA_SHA1)
+      {
+      throw TLS_Exception(Alert::ILLEGAL_PARAMETER, "SHA-1 algorithm must not be used");
```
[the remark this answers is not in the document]
reneme: Apparently this review comment clashed with my rebase. This particular message got removed by the `Signature_Scheme` refactoring. Now it says: "Peer sent signature algorithm that is not suitable for TLS 1.3".

src/lib/tls/msg_key_update.cpp
```diff
@@ -0,0 +1,46 @@
+/*
+* Key Update message
+* (C) 2022 Jack Lloyd
+*     2022 René Meusel, Hannes Rantzsch - neXenio GmbH
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/build.h>
```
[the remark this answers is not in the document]
reneme: Indeed, along with `msg_certificate_13.cpp`, `msg_excrypted_extensions.cpp`, `tls_extensions_key_share.cpp` and `msg_certificate_12.cpp` (in TLS 1.2, ofc).

src/lib/tls/tls13/tls_cipher_state.cpp
```diff
@@ -0,0 +1,445 @@
...
+constexpr size_t NONCE_LENGTH = 12;
+}
+
+std::unique_ptr<Cipher_State> Cipher_State::init_with_server_hello(
+   const Connection_Side side,
+   secure_vector<uint8_t>&& shared_secret,
+   const Ciphersuite& cipher,
+   const Transcript_Hash& transcript_hash)
+   {
+   auto cs = std::unique_ptr<Cipher_State>(new Cipher_State(side, cipher));
```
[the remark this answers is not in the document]
reneme: Unfortunately, the used constructor is private. Hence, calling it from the `make_unique` template does not work.

src/lib/tls/msg_server_hello.cpp
```diff
@@ -408,4 +472,197 @@ std::vector<uint8_t> Server_Hello_Done::serialize() const
...
+      {
+      TLSEXT_COOKIE,
+      TLSEXT_SUPPORTED_VERSIONS,
+      TLSEXT_KEY_SHARE,
+      };
+
+   // As the Hello Retry Request shall only contain essential extensions, we
+   // don't give any slack for extensions not implemented by Botan here.
+   if(extensions().contains_other_than(allowed))
```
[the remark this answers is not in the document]
reneme: ... refactoring is the root of all evil. ;)

src/lib/tls/tls_policy.cpp
reneme: Sounds reasonable. At least until "strict" doesn't mean to prohibit certain TLS versions. I'm removing all three "allows"...

src/fuzzer/tls_client_hello.cpp
reneme: see: https://github.com/randombit/botan/pull/2977

src/lib/tls/tls12/tls_handshake_state.cpp
reneme: This got replaced by `Signature_Scheme::is_compatible_with`.

src/cli/tls_utils.cpp
reneme: > We'll need to allow deserializing a Client Hello w/o knowing if it'll be 1.2 or 1.3, just as we'll eventually need to do for the TLS 1.3 server.

I feel, it would be fine to postpone that until the TLS 1.3 server is implemented.

src/lib/tls/tls13/tls_channel_impl_13.cpp
```diff
@@ -0,0 +1,397 @@
...
+         const auto& record = std::get<Record>(result);
+
+         // RFC 8446 5.1
+         //   Handshake messages MUST NOT be interleaved with other record types.
+         if(record.type != HANDSHAKE && m_handshake_layer.has_pending_data())
+            { throw Unexpected_Message("Expected remainder of a handshake message"); }
+
+         if(record.type == HANDSHAKE)
+            {
+            m_handshake_layer.copy_data(unlock(record.fragment));  // TODO: record fragment should be an ordinary std::vector
```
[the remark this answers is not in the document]
reneme: The `Record_Layer` knows the exact length of the record fragment upfront. So the `record.fragment` secure vector is not reallocated. In contrast to `Handshake_Layer` which performs a good bit of copying when parsing individual messages from records.

Hence, https://github.com/randombit/botan/pull/2922/commits/574b8f733ff034059ebd159ef24faca42b53cc4c changes the `Handshake_Layer::copy_data()` to take a secure vector. Internally, the contents are then copied into a standard vector for further processing.

reneme:
@randombit Thanks for looking deeper into this! Your review is much appreciated.
I added your proposed changes up to this point.

Apart from the last unresolved tasks and discussions (https://github.com/randombit/botan/pull/2922#discussion_r887700253, https://github.com/randombit/botan/pull/2922#discussion_r887715983, https://github.com/randombit/botan/pull/2922#discussion_r839591154) I'd like to draw your attention to the list of "Potential Future Work" in this PR's description. Albeit uncritical, I believe those improvements would be beneficial for both performance and maintainability. Should we open a separate issue to track those?

Further, I feel we should invest some time into reviewing the application-facing interfaces (particularly `TLS::Policy` and `TLS::Callbacks`). Some of the callbacks don't make sense for TLS 1.3 or lack information to be useful. We started assembling a list a while back: https://github.com/randombit/botan/issues/2714#issuecomment-1057175631 (probably not complete). Obviously, most of those potential API adaptions could also be tackled when the need comes up. Though, some might require breaking API changes.

Lastly, I would really appreciate a beta-test by a potential future consumer of the TLS 1.3 implementation. Maybe @pist-eb is still interested?

randombit:
> Should we open a separate issue to track those?

I saw these and don't think any should block merging this PR. We could track the remaining work either as distinct issues, or (maybe simpler) by opening a single issue that contains a checklist of all of the todos and fixmes that would otherwise block releasing a "final" 1.3 implementation.

> Further, I feel we should invest some time into reviewing the application-facing interfaces (particularly TLS::Policy and TLS::Callbacks).

I agree on this. Some may already be obsolete with the removal of TLS 1.0/1.1 also.

For the key agreement callbacks I think we can just remove or change them suitably to cover both 1.2 and 1.3. In general a gradual deprecation is nicer but those particular callbacks are very niche and IMO it's simpler for us to just document what changed and give guidance on how to adapt to whatever we come up with that works for both 1.2 and 1.3

reneme:
> > Further, I feel we should invest some time into reviewing the application-facing interfaces (particularly TLS::Policy and TLS::Callbacks).
> 
> I agree on this. Some may already be obsolete with the removal of TLS 1.0/1.1 also.

I'm working on a `TLS::Callbacks` overhaul: https://github.com/randombit/botan/pull/2988 Please feel free to put in your feedback on the go.

reneme:
No issues with merging it. Thanks for the time reviewing all of this. 😊

reneme:
@randombit we're waiting for your final approval. :)

randombit (approved):

hrantzsch:
Thank you so much @randombit and @reneme. Really happy about this 🥳

PR #2929, opened 2022-03-08
Fixed more -Wredundant-tags warnings

jankoerner:
I had to add new forward declarations, because otherwise it won't compile.


randombit:
Looks good, thanks!

PR #2932, opened 2022-03-10
Provide Editor Configurations via .editorconfig and specifically for Sublime Text

reneme:
This adds a generic `.editorconfig` setting up basic editor behaviour. Also it adds a Sublime Text project file with a few convenience scripts.

Both editor configs are not enabled by default but potential users are required to symlink the config they wish to use into the checkout's root directory. E.g. for sublime: `ln -s src/editors/sublime/sublime-project botan.sublime-project`.

The sublime project provides some best-effort build automations that are explained in further detail in the respective [README.md](https://github.com/randombit/botan/compare/master...reneme:convenience/editorconfigs?expand=1#diff-150a0978cea9c9bf574825073172533da4fdfa2a294fe4d9c38355ecfcdfab6a). Note that this is fairly work-and-progress and just reflects what I personally found useful right now.

Also, I noticed a similar config for Emacs in `src/configs` which might move into `src/editors` in case it is still actively used and up to date.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,7 @@ vgcore.*
 *.pyc
 .DS_Store
 *.swp
+/*.cache
 
 # ctags/etags files
 /TAGS
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,11 @@ vgcore.*
 \#*\#
 .\#*
 
+# Editor configuration files (top level)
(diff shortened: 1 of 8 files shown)


randombit:
Nice.

Emacs config might be slightly out of date (would have to check) but still works anyway. I'll move it.

Issue #2931, opened 2022-03-09
x509 module without sha1 dependency


securitykernel:
Our high-level goal is to be able to use the BSI module and additionally enable the tls module, but leave the sha1 module disabled. This is currently not possible, as tls depends on x509, which depends on sha1. From looking at the code, it seems the x509 module already handles non-existence of the sha1 module by throwing, e.g., at runtime in `X509Cert::subject_public_key_bitstring_sha1()`. This function seems to be only used by `Flatfile_Certificate_Store`, which we could potentially replace. Also, ocsp relies on `subject_public_key_bitstring_sha1()`. Additionally, the `CertificateStore` interface and implementations rely on it via `CertificateStore::find_cert_by_pubkey_sha1()`, I am not sure what the consequences would be here.

Temporarily removing sha1 from x509's info.txt builds just fine, but unsurprisingly some tests fail:

`Tests complete ran 171854 tests in 23.68 sec 129 tests failed (in ecdsa_unit ocsp x509_path_name_constraint x509_path_nist x509_path_rsa_pss x509_path_with_ocsp x509_path_x509test x509_v1_ca)`

Probably we can replace most of the test certificates using SHA-1 with SHA-256 ones, that depends on the source. I will check these.

randombit:
I think one issue is that OCSP relies on SHA-1 as a key identifier, though I don't recall the details at the moment.

Absolutely in favor of any changes which allow disabling SHA-1 in more configurations.

Possibly for OCSP we can use some different key identifier mechanism instead.

securitykernel:
So from what I can see, the following things need to be done to fix the tests:

- [x] Fix ecdsa_unit test: Check whether SHA-1 is available
- [ ] Fix ocsp tests: Possibly we can use SHA-256 here, need to check
- [ ] Fix x509_path_name_constraint: xxx
- [ ] Fix x509_path_nist: Hard without NIST's help, "All certificates are signed with RSA PKCS#1 v1.5 and SHA-1", can we get NIST to update their test suite to use SHA-256?
- [ ] Fix x509_path_rsa_pss: Certificates from different sources, some already use SHA-256, some still SHA-1
- [ ] Fix x509_path_with_ocsp: See ocsp
- [ ] Generate x509_path_x509test certificates with SHA-256: https://github.com/yymax/x509test hasn't been updated since 8 years ago and does not work out of the box with python 3.9, maybe we can use an older Ubuntu container image 
[...]

PR #2933, opened 2022-03-10
Fix ecdsa_unit test failures when SHA-1 is not available

securitykernel:
* Certificates used in `test_decode_ver_link_SHA1()` are signed with SHA-1, so disable the test case without SHA-1
* Use SHA-256 fingerprint in `test_decode_ecdsa_X509()`

Part of fixes for #2931

### DIFF
--- a/src/tests/unit_ecdsa.cpp
+++ b/src/tests/unit_ecdsa.cpp
@@ -106,7 +106,6 @@ Test::Result test_decode_ver_link_SHA256()
    return result;
    }
 
-#if defined(BOTAN_HAS_SHA1)
 Test::Result test_decode_ver_link_SHA1()
    {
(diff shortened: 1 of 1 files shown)


src/tests/unit_ecdsa.cpp
```diff
@@ -106,6 +106,7 @@ Test::Result test_decode_ver_link_SHA256()
    return result;
    }
 
+#if defined(BOTAN_HAS_SHA1)
 Test::Result test_decode_ver_link_SHA1()
```
randombit, line 110: Instead of compiling this test out have it still load the certificates, then check that if SHA1 is disabled the verification fails. That (albeit somewhat implicitly) tests that we can still parse a SHA-1 signed certificate even if we cannot verify it.

securitykernel: Changed

PR #2943, opened 2022-04-01
performance: don't reserve before buffer inserting

### LINKED ISSUE #2920: Substantial performance hit encoding ASN1 Sequence between 2.9 and 2.18
We are working on an application, where we need to create a large ASN1 SEQUENCE with 100K OCTET-STRING entries.

So far we used libbotan-2-dev that comes with Debian Buster in version 2.9. There was no notable time required to encode the structure.

This is completely different with 2.18, where encoding the list takes a dramatic 90 seconds.

Looking at the git revision history I can not really see what may cause this decrease in performance. 

Any idea ?

hrantzsch:
Fixes #2920

Calling `reserve` before inserting into a vector with `operator+=`
caused a significant performance degradation when called very often (see
referenced GH issue).
In this case, letting `insert` reserve exponentially more memory when
needed is a better strategy than manually calling `reserve`, which would
need to allocate new memory on every call.

[cppreference.com](https://en.cppreference.com/w/cpp/container/vector/reserve) notes that "a function that receives an arbitrary
vector by reference and appends elements to it should usually not call
reserve() on the vector, since it does not know of the vector's usage
characteristics."

### DIFF
--- a/src/lib/base/secmem.h
+++ b/src/lib/base/secmem.h
@@ -79,7 +79,6 @@ std::vector<T, Alloc>&
 operator+=(std::vector<T, Alloc>& out,
            const std::vector<T, Alloc2>& in)
    {
-   out.reserve(out.size() + in.size());
    out.insert(out.end(), in.begin(), in.end());
    return out;
    }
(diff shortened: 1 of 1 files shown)


randombit (approved):
Thanks for following up on this

PR #2944, opened 2022-04-02
new util to name the data pages for now linux and especially android

devnexen:
which supports it since a lot longer.

### DIFF
--- a/src/lib/utils/os_utils.cpp
+++ b/src/lib/utils/os_utils.cpp
@@ -67,6 +67,14 @@
   #include <mach/vm_statistics.h>
 #endif
 
+#if defined(BOTAN_TARGET_OS_HAS_PRCTL)
+  #include <sys/prctl.h>
+  #if !defined(PR_SET_VMA)
+    #define PR_SET_VMA 0x53564d41
+    #define PR_SET_VMA_ANON_NAME 0
+  #endif
+#endif
+
 namespace Botan {
 
 // Not defined in OS namespace for historical reasons
@@ -587,6 +595,8 @@ std::vector<void*> OS::allocate_locked_pages(size_t count)
 
       std::memset(ptr, 0, 3*page_size); // zero data page and both guard pages
 
+      // Attempts to name the data page
+      page_named(ptr, 3*page_size);
       // Make guard page preceeding the data page
       page_prohibit_access(static_cast<uint8_t*>(ptr));
       // Make guard page following the data page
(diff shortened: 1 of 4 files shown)


src/lib/utils/os_utils.cpp
```diff
@@ -655,6 +665,16 @@ void OS::free_locked_pages(const std::vector<void*>& pages)
       }
    }
 
+void OS::page_named(void* page, size_t size)
+   {
+   constexpr char name[] = "Botan";
```
randombit, line 670: Should this also be `static`?

src/lib/utils/os_utils.cpp
```diff
@@ -655,6 +665,16 @@ void OS::free_locked_pages(const std::vector<void*>& pages)
...
+#if defined(BOTAN_TARGET_OS_HAS_PRCTL)
+   (void)prctl(PR_SET_VMA, PR_SET_VMA_ANON_NAME, reinterpret_cast<uintptr_t>(page), size, name);
+#else
+   (void)name;
```
randombit, line 674: Use `BOTAN_UNUSED` here

src/lib/utils/os_utils.cpp
```diff
@@ -655,6 +665,16 @@ void OS::free_locked_pages(const std::vector<void*>& pages)
       }
    }
 
+void OS::page_named(void* page, size_t size)
+   {
+   constexpr char name[] = "Botan";
+#if defined(BOTAN_TARGET_OS_HAS_PRCTL)
+   (void)prctl(PR_SET_VMA, PR_SET_VMA_ANON_NAME, reinterpret_cast<uintptr_t>(page), size, name);
```
randombit, line 672: Use

```
int rc = prctl(...);
BOTAN_UNUSED(rc); // ignore errors
```

src/lib/utils/os_utils.h
```diff
@@ -149,6 +149,11 @@ void page_prohibit_access(void* page);
 */
 void page_allow_access(void* page);
 
+/**
+* Set a ID to a page's range
+*/
+void page_named(void* page, size_t size);
```
randombit, line 155: Doc comment should specify if the size is the page count or the byte count.

randombit (commented):
Overlall fine, had a few comments

src/lib/utils/os_utils.cpp
randombit: Should be `BOTAN_UNUSED(name, size)` since both otherwise MSVC will warn about unused params.

PR #2947, opened 2022-04-04
[TLS 1.3] Extract Handshake_Transitions from Handshake_State

reneme:
Depends on: #2941 (requires rebase after merge)

This extracts the handshake state machine transition checker from `Handshake_State` into an extra class (`Handshake_Transitions`). Mainly, the TLS 1.3 implementation will need this functionality independent of the TLS 1.2 `Handshake_State` class.

Note that this PR is currently based on #2941 and contains the test utility code. I'll rebase once that PR has been merged.

### DIFF
--- a/src/tests/test_tls_handshake_transitions.cpp
+++ b/src/tests/test_tls_handshake_transitions.cpp
@@ -1,0 +1,91 @@
+/*
+* (C) 2022 Jack Lloyd
+* (C) 2022 Hannes Rantzsch, René Meusel - neXenio
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include "tests.h"
+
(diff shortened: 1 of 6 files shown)


randombit:
LGTM, merged #2941 so this is ready for a rebase

src/tests/test_tls_handshake_transitions.cpp
```diff
@@ -0,0 +1,91 @@
...
+*/
+
+#include "tests.h"
+
+#if defined(BOTAN_HAS_TLS)
+
+#include <botan/internal/tls_handshake_transitions.h>
+
+using namespace Botan::TLS;
+using namespace Botan_Tests;
```
randombit, line 15: Instead of being in global namespace with a `using` this should be within `Botan_Tests` namespace

PR #2953, opened 2022-04-05
FIX: Unused variables in OS::page_named

reneme:
Not sure what happened in #2944. Forgot to `git pull -r` before merge?
Anyway, lets reapply @devnexen 's changes. :)


reneme:
@randombit I think we're stepping on each others toes here. Nevertheless, I think [1846f35](https://github.com/randombit/botan/pull/2953/commits/1846f35887747e586f8d98a37513e93f367a329b) is relevant.

randombit:
I just fixed with a push to master, yeah I forgot to repull the branch before merging :cry: but should be fixed now

PR #2958, opened 2022-04-14
[TLS 1.3] Signature_Scheme Class

hrantzsch:
This PR introduces a class for signature schemes, replacing the existing `Signature_Scheme` enum and the free-standing functions around it (`signature_algorithm_of_scheme`, `hash_function_of_scheme`, etc).

Previously, it was not clear whether these functions would throw if an invalid or unknown scheme was provided. This has been changed: the member functions never throw. Rather, `Signature_Scheme::is_available` should be used to check if a scheme can be used (is implemented).

We intend to extend this class with higher level predicates, such as `bool is_allowed_for(TLS::Version)`, or similar.

Note that this affects the public API of TLS::Policy and Signature_Scheme. Users can still create schemes as, e.g., `Botan::TLS::Signature_Scheme::ECDSA_SHA256`, because the enum class moved into the actual class (but is no longer an enum _class_).


src/lib/tls/msg_cert_verify.cpp
```diff
@@ -62,11 +62,11 @@ std::vector<uint8_t> Certificate_Verify::serialize() const
    {
    std::vector<uint8_t> buf;
 
-   if(m_scheme != Signature_Scheme::NONE)
+   // TODO: both for TLS 1.2 and 1.3 this should never happen and the message would be invalid
+   if(m_scheme.is_set())
```
hrantzsch, line 66: This check should not be necessary. Otherwise the message we create would be malformed.

randombit: In this case should it be a `BOTAN_ASSERT` instead?

src/lib/tls/msg_cert_verify.cpp
```diff
@@ -62,11 +62,11 @@ std::vector<uint8_t> Certificate_Verify::serialize() const
...
       {
-      const uint16_t scheme_code = static_cast<uint16_t>(m_scheme);
-      buf.push_back(get_byte<0>(scheme_code));
-      buf.push_back(get_byte<1>(scheme_code));
+      buf.push_back(get_byte<0>(m_scheme.wire_code()));
+      buf.push_back(get_byte<1>(m_scheme.wire_code()));
```
randombit, line 69: This is a very minor point but perhaps easier to read with a variable to hold the result of `m_scheme.wire_code`

src/lib/tls/tls_signature_scheme.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+const std::vector<Signature_Scheme>& Signature_Scheme::all_available_schemes()
+   {
+   /*
+   * This is ordered in some approximate order of preference
+   */
+   static const std::vector<Signature_Scheme> all_schemes = {
+
+#if defined(BOTAN_HASH_ED25519)
+      EDDSA_25519,
```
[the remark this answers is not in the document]
randombit: What you are doing seems fine. It is an odd omission and I'm not sure why that would have been the case. I'll look into what the current code is doing.

src/lib/tls/tls_signature_scheme.h
```diff
@@ -0,0 +1,100 @@
...
+
+namespace Botan::TLS {
+
+class BOTAN_PUBLIC_API(3,0) Signature_Scheme
+{
+public:
+/*
+* Matches with wire encoding
+*/
+enum Code : uint16_t {
```
randombit, line 27: My initial reaction was "can we make this an `enum class`"? But I see that this breaks other code that is using `Signature_Scheme` as if it was still an `enum`. So maybe add a comment here, that the bare `enum` is intentional because the overall class is emulating an enum.

src/lib/tls/tls_signature_scheme.h
```diff
@@ -0,0 +1,100 @@
...
+
+   std::string to_string() const noexcept;
+   std::string hash_function_name() const noexcept;
+   std::string padding_string() const noexcept;
+   std::string algorithm_name() const noexcept;
+   AlgorithmIdentifier algorithm_identifier() const noexcept;
+   std::optional<Signature_Format> format() const noexcept;
+
+   bool is_sha1() const noexcept;
+   bool is_pkcs1() const noexcept;
```
randombit, line 89: So this is more clear/explicit lets name it `is_rsa_pkcs1`

src/tests/test_tls_signature_scheme.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+#if defined(BOTAN_HAS_TLS)
+
+#include <botan/tls_signature_scheme.h>
+
+#include <botan/ec_group.h>
+
+namespace {
+
+using namespace Botan::TLS;
```
randombit, line 18: Remove this and use `Botan::TLS` prefix explicitly. Or if worst case it is truly too verbose for some reason, contain the using namespace to within a function.

randombit (commented):
This looks great. In the last years I've often bemoaned that C++ does not have Rust-style enums which allow you to attach functions, but this is a nice approach to getting that behavior.

A few minor comments but overall :+1:

src/lib/tls/tls_signature_scheme.cpp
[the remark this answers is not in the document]
randombit: Oh wait, I forgot - right now we don't support Ed25519 certificates for authentication. So this should not be included in the return, until we do have such support.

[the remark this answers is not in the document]
reneme: I'll just comment this code with a reference to your comment here.

src/lib/tls/tls_signature_scheme.h
reneme: From your comment I figure you're fine with this end user-facing API change. I'll add a comment describing the intention regarding the bare enum and that it is supposed to resemble the old `Signature_Scheme` enum class.

reneme:
Please don't merge this yet. While finishing up the client authentication code I made a few adaptions to the API of `Signature_Scheme` which I'd like to discuss with @hrantzsch first tomorrow.

src/lib/tls/tls_signature_scheme.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+      if(m_code == ECDSA_SHA256 && !(keylen >= 250 && keylen <= 350))
+         return false;
+
+      if(m_code == ECDSA_SHA384 && !(keylen >= 350 && keylen <= 450))
+         return false;
+
+      if(m_code == ECDSA_SHA512 && !(keylen >= 450 && keylen <= 550))
+         return false;
+      }
```
reneme, line 338: BoGo tests whether the curve's bit length matches the associated hash function output length. We we weren't sure how to test for this and went with a "fuzzy" range for the ECC key length. The idea was to incorporate custom curves whose bit key length don't fall exactly on predefined thesholds. Does it make sense to implement it like that? @randombit

src/lib/tls/tls_signature_scheme.h
reneme: Discussion moved [here](https://github.com/randombit/botan/pull/2968/files#r859412553).

src/lib/tls/tls_signature_scheme.cpp
reneme: Disscussion moved [here](https://github.com/randombit/botan/pull/2968/files#r859411641).

reneme:
Closed in favour of https://github.com/randombit/botan/pull/2968

PR #2964, opened 2022-04-25
Correct the case of ressol modulo 2

### LINKED ISSUE #2962: Special case in ressol() for p == 2 is never reached
The `ressol()` function contains a special case for `p == 2`, so I assume this is a valid input.

But in 55fa3685c5053d66533a7a9e0f08403ffa95b323 the input checks there rearranged so that the `p.is_even()` check is now done at the beginning, which results in an exception being thrown before the special case for `p == 2` is reached.

randombit:
Also rename ressol to sqrt_modulo_prime to make the meaning more clear.

Fixes #2962

### DIFF
--- a/src/fuzzer/ecc_helper.h
+++ b/src/fuzzer/ecc_helper.h
@@ -33,7 +33,7 @@ Botan::BigInt decompress_point(bool yMod2,
    g += curve_b;
    g = g % curve_p;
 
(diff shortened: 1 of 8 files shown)


lieser (commented):
Sorry for the late review, noticed only now that my comment I already added a while ago was never visible because I failed to press the submit review button.

PR #2967, opened 2022-04-27
[TLS 1.3] Add a "consolidation c'tor" for Test::Result

reneme:
This is somewhat a request for comment in the hope to iterate towards a better solution.

For the RFC 8448 based test in the main TLS pull request, I would like to use the new `CHECK` helper to structure sub sections of individual tests. Furthermore, I want to use `Text_Based_Tests` to store the transcript records of RFC 8448.

Now, using `CHECK` results in a vector of `Test::Result` objects but `Text_Based_Tests::run_one_test` expects a single `Test::Result` as return value. Hence, I added a "consolidation constructor" to somewhat merge the fine grained downstream `Test::Result` objects into a single `Test::Result` to return into the test framework. I'm not super happy with this approach, as I'm not sure whether it swallows vital failure information that might make test debugging harder.

### DIFF
--- a/src/tests/tests.cpp
+++ b/src/tests/tests.cpp
@@ -30,9 +30,9 @@
 
 namespace Botan_Tests {
 
-void Test::Result::merge(const Result& other)
+void Test::Result::merge(const Result& other, bool ignore_test_name)
    {
-   if(who() != other.who())
+   if(!ignore_test_name && who() != other.who())
       {
       throw Test_Error("Merging tests from different sources");
       }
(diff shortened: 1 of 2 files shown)


reneme:
For reference, a realistic failure in a downstream `Test::Result` looks about like that:

```
tls_rfc8448:
Client Authentication (Client side) ran 4 tests in 2.74 msec 2 FAILED
Failure 1: other handshake messages and client auth failed unexpectedly with error Test error Fixed output RNG ran out of bytes, test bug?
Failure 2: Close Connection unexpected result for Client close_notify
Produced: 1703030013BD3A77FF4BD1F015E4ECB853A0E6B392981B2A
Expected: 1703030013E4AD7D44C29245339D355962C779B89EF44C58
XOR Diff: 000000000059970ABB8943B52679D9E131679F0B0C6C5772
Note 1: Client Authentication (Client side) Test # 1 Client_Authentication_Handshake failed ClientFinished=170303... ClientHello_1=160301... Client_CloseNotify=170303... RNG_Pool=6a47... ServerHandshakeMessages=170303... ServerHello=160303... Server_CloseNotify=170303...
```

randombit:
Seems kind of hackish but I don't have an immediate suggestion on improvement and I do see the need for this functionality, so I'm fine with this PR as is.

randombit (approved):

PR #2968, opened 2022-04-27
[TLS 1.3] Signature_Scheme Class

reneme:
### Reincarnation of this [pull request](https://github.com/randombit/botan/pull/2958)

@hrantzsch accidentally opened it from his private Botan fork. As this doesn't allow me to do amendments  without his help (while he is enjoying his PTO), I reopened the PR here. Note that I re-posted the open discussions from the old PR.

### Pull Request Dependencies

This is based on [the basic TLS 1.3 implementation](https://github.com/randombit/botan/pull/2958), should be tracked along with it and re-targeted to master once the underlying pull request is merged.

### Original PR description

This PR introduces a class for signature schemes, replacing the existing `Signature_Scheme` enum and the free-standing functions around it (`signature_algorithm_of_scheme`, `hash_function_of_scheme`, etc).

Previously, it was not clear whether these functions would throw if an invalid or unknown scheme was provided. This has been changed: the member functions never throw. Rather, `Signature_Scheme::is_available` should be used to check if a scheme can be used (is implemented).

We intend to extend this class with higher level predicates, such as `bool is_allowed_for(TLS::Version)`, or similar.

Note that this affects the public API of TLS::Policy and Signature_Scheme. Users can still create schemes as, e.g., `Botan::TLS::Signature_Scheme::ECDSA_SHA256`, because the enum class moved into the actual class (but is no longer an enum _class_).


src/lib/tls/tls_signature_scheme.cpp
```diff
@@ -0,0 +1,336 @@
...
+      return false;
+
+   if(m_code == ECDSA_SHA384 && !(keylen >= 350 && keylen <= 450))
+      return false;
+
+   if(m_code == ECDSA_SHA512 && !(keylen >= 450 && keylen <= 550))
+      return false;
+
+   return true;
+   }
```
reneme, line 334: BoGo tests whether the curve's bit length matches the associated hash function output length. We we weren't sure how to test for this and went with a "fuzzy" range for the ECC key length. The idea was to incorporate custom curves whose bit key length don't fall exactly on predefined thesholds. Does it make sense to implement it like that? @randombit

src/lib/tls/tls_signature_scheme.h
```diff
@@ -0,0 +1,106 @@
...
+{
+public:
+/**
+ * Matches with wire encoding
+ *
+ * Note that this is intentionally left as a bare enum. It emulates the Botan 2
+ * API where `Signature_Scheme` was an enum class with associated free-standing
+ * functions. Leaving it as a bare enum resembles the legacy user-facing API.
+ */
+enum Code : uint16_t {
```
reneme, line 33: @randombit [said](https://github.com/randombit/botan/pull/2958#discussion_r851222378):

> My initial reaction was "can we make this an `enum class`"? But I see that this breaks other code that is using `Signature_Scheme` as if it was still an `enum`. So maybe add a comment here, that the bare `enum` is intentional because the overall class is emulating an enum. 

From your comment I figure you're fine with this end user-facing API change. I'll add a comment describing the intention regarding the bare enum and that it is supposed to resemble the old `Signature_Scheme` enum class.

PR #2974, opened 2022-05-05
[TLS 1.3] Session Resumption

reneme:
This adds session resumption to [the TLS 1.3 implementation](https://github.com/randombit/botan/pull/2922).

Central changes in this pull request:

* Extend the `TLS::Session` class to allow storing TLS 1.3 sessions
* Provide a new entry point into the TLS 1.3 key schedule: `init_with_psk()`
* Allow premature serialization of `Client_Hello_13` to calculate PSK binders
  * Implement a `Truncate(Client_Hello)` function as outlined in [RFC 8446 4.2.11.2](https://www.rfc-editor.org/rfc/rfc8446.html#section-4.2.11.2)
  * Extend the TLS client state machine to accomodate resumption
* ~~Provide `TLS::Callbacks::tls_current_timestamp()` for deterministic timestamps for testing~~ (now extracted [here](https://github.com/randombit/botan/pull/2975))

Currently, the TLS 1.3 PSK mechanism is exclusively used for session resumption. It should be fairly straight forward to extend it for out-of-band PSKs. Furthermore, this pull request leaves support for early data and early key export to future work.

Caveat: [RFC 8446 suggests to not reuse session tickets](https://www.rfc-editor.org/rfc/rfc8446.html#appendix-C.4) to prevent passive client tracking. This is not implemented by the client implementation (neither in TLS 1.2). I suppose this is meant to be implemented by a custom `Session_Manager` if needed?!


src/lib/tls/tls_session.cpp
```diff
@@ -90,9 +145,27 @@ Session::Session(const uint8_t ber[], size_t ber_len)
...
+      {
+      seqdec
+         .decode(m_early_data_allowed)
+         .decode_integer_type(m_max_early_data_bytes)
+         .decode_integer_type(m_ticket_age_add)
+         .decode_integer_type(m_lifetime_hint);
+      }
+
+   seqdec.end_cons();
+   dec.verify_end();
```
reneme, line 168: This is compatibility code to allow the new session implementation to read legacy session objects and resume sessions that were saved before migrating to Botan 3.0.

It introduces quite a bit of complexity due to handling the `BER_Decoder` objects properly. Is this even necessary? I feel it might be okay that Botan 2.x sessions are lost when updating to Botan 3.0?

src/lib/tls/tls_session.h
```diff
@@ -139,8 +153,11 @@ class BOTAN_PUBLIC_API(2,0) Session final
 
       /**
       * Get the session identifier
+      *
+      * Note that for TLS 1.3 sessions this will be equal to the session ticket
+      * and might therefore be a up to 2^16-1 bytes long.
       */
-      const std::vector<uint8_t>& session_id() const { return m_identifier; }
+      const std::vector<uint8_t>& session_id() const;
```
reneme, line 160: This is worth noting! Probably even more so for users of the library, as they might need to deal with the fact that session IDs are suddenly not 32-byte buffers anymore but potentially much larger. Will there be a "migration guide to Botan 3.0" of some sort where this could be mentioned?

TLS 1.3 removes the concept of session IDs entirely, because resumption was merged with Pre-Shared-Key negotiation. Instead PSK negotiations that stem from session tickets are identified by the opaque session ticket buffer itself. However, our `Session_Manager` API relies on the session id as a database index to retrieve stored sessions on the server side.

I considered "inventing" session IDs by simply hashing the TLS 1.3 session tickets, but found that to be intransparent magic that might better be implemented by user's application code. Especially when we want to enable library users to actually use the session ticket as a means to keep encrypted state on the client side.

Thoughts?

src/lib/tls/tls13/tls_cipher_state.cpp
```diff
@@ -27,7 +31,7 @@
  *                                     |
  *                                     *
  *                             STATE EARLY TRAFFIC
- * This state is reached by constructing Cipher_State using init_with_psk() (not yet implemented).
+ * This state is reached by constructing Cipher_State using init_with_psk().
```
reneme, line 34: Needs revision: State is reached by initializing with `init_with_psk()` and then advancing with `advance_with_client_hello()`.

src/lib/tls/tls13/tls_cipher_state.cpp
```diff
@@ -114,12 +118,43 @@ std::unique_ptr<Cipher_State> Cipher_State::init_with_server_hello(
...
    return cs;
    }
 
+std::unique_ptr<Cipher_State> Cipher_State::init_with_psk(
+         const Connection_Side side,
+         const Cipher_State::PSK_Type type,
+         secure_vector<uint8_t>&& psk,
+         const Ciphersuite& cipher)
+   {
+   auto cs = std::unique_ptr<Cipher_State>(new Cipher_State(side, cipher.prf_algo()));
```
reneme, line 133: I read the RFC so that for PSK we just need to know the hash function from the start but during the remaining negotiation we might select a different cipher suite (that must share the same hash function, though). We should have tests that verify this possibility (maybe boGo has some?) and add a check in `advance_with_server_hello()` that the negotiated cipher suite doesn't change the pre-determined hash function.

src/bogo_shim/config.json
```diff
@@ -40,33 +41,10 @@
...
-        "Ticket-Forbidden-TLS13": "TLS 1.3 session resumption is NYI",
-        "Resume-Client-PRFMismatch-TLS13": "TLS 1.3 session resumption is NYI",
-        "TLS13-HelloRetryRequest-Client-*": "TLS 1.3 session resumption is NYI",
-        "TLS13-TicketAgeSkew-*": "TLS 1.3 session resumption is NYI",
-        "CurveID-Resume-Client-TLS13": "TLS 1.3 session resumption is NYI",
-        "ALPNClient-TLS-TLS13": "TLS 1.3 session resumption is NYI",
+        "TLS12SessionID-TLS13": "We don't offer TLS 1.3 when a TLS 1.2 session was found",
+        "Ticket-Forbidden-TLS13": "We don't offer TLS 1.3 when a TLS 1.2 session was found",
+        "Resume-Client-NoResume-TLS12-TLS13-TLS": "We don't offer TLS 1.3 when a TLS 1.2 session was found",
+        "Resume-Client-Mismatch-TLS12-TLS13-TLS": "We don't offer TLS 1.3 when a TLS 1.2 session was found",
```
reneme, line 47: We basically query the `Session_Manager` before instantiating the TLS 1.2 or TLS 1.3 implementation. If we find a TLS 1.2 session we don't even bother offering TLS 1.3 in the first place. Hence, we cannot fulfil those tests. Though, I think it's worth discussing whether the described behaviour is actually wise.

src/lib/tls/msg_client_hello.cpp
```diff
@@ -23,6 +23,11 @@
 #include <botan/internal/tls_handshake_io.h>
 #include <botan/internal/tls_handshake_hash.h>
 
+#ifdef BOTAN_HAS_TLS_13
+#include <botan/internal/tls_transcript_hash_13.h>
+#include <botan/internal/tls_handshake_layer_13.h>
+#endif
```
reneme, line 29: Indentation?

src/lib/tls/tls13/tls_client_impl_13.cpp
```diff
@@ -421,18 +476,34 @@ void Client_Impl_13::handle(const Finished_13& finished_msg)
    m_cipher_state->advance_with_server_finished(m_transcript_hash.previous());
    m_cipher_state->advance_with_client_finished(m_transcript_hash.current());
 
-   // TODO: save session and invoke tls_session_established callback
+   // TODO: Create a dummy session object and invoke tls_session_established.
+   //       Alternatively, consider changing the expectations described in the
+   //       callback's doc string.
```
reneme, line 481: The callback `tls_session_established` has the underlying assumption that a successful TLS handshake generates exactly one resumable session. In TLS 1.2 this session was constructed and reported via this callback. TLS 1.3 breaks with this assumption by allowing an arbitrary number of "New Session Ticket" messages at the server's discretion.

Therefore, I feel this callback doesn't really fit for TLS 1.3. Creating a "dummy" session as suggested by the comment might be a way out, but it feels wrong...

src/lib/tls/tls13/tls_transcript_hash_13.cpp
```diff
@@ -58,15 +63,125 @@ Transcript_Hash_State Transcript_Hash_State::recreate_after_hello_retry_request(
...
+//       as well as basic parsing of extensions. We should resolve this.
+//
+// Ad-hoc idea: When parsing the production objects, we could keep markers into
+//              the original buffer. E.g. the PSK extensions would keep its off-
+//              set into the entire client hello buffer. Using that offset we
+//              could quickly identify the offset of the binders list slice the
+//              buffer without re-parsing it.
+//
+// Finds the truncation offset in a serialization of Client Hello as defined in
+// RFC 8446 4.2.11.2 used for the calculation of PSK binder MACs.
```
reneme, line 78: ... yeah. The binder MACs are unfortunately a major cross-cutting concern through our layered implementation architecture. Probably it would be a good idea how other implementations are handling this. Maybe I missed a short cut. The current implementation works, but its sad to see this.

reneme:
Rebased and retargeted to master.

src/lib/tls/tls13/tls_client_impl_13.cpp
reneme: Repeated "New Session Ticket" messages establishe multiple resumable session tickets. I.e. each session ticket can be used to resume a TLS session independently. The cryptographic parameters are fixed after the handshake and won't differ between individual session tickets. However: The server might opt to hand out additional session tickets after certain events have occured (e.g. after a post-handshake client authentication).

To the best of my knowledge, this behaviour is application-specific, though. RFC 8446 just specifies the possibility to hand out an arbitrary number of session tickets at any point after the handshake.

src/lib/tls/tls13/tls_cipher_state.h
```diff
@@ -58,9 +59,25 @@ class Ciphersuite;
  */
 class BOTAN_TEST_API Cipher_State
    {
+   public:
+      enum class PSK_Type
```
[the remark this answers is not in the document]
reneme: I tend to use (redundant) visibility markers as visual cues to separate "internal data structures", "methods" and "member variables" in my class implementations. To the best of my knowledge, that is not Botan's "official" style guide, but more of my personal preference.

src/lib/tls/tls13/tls_client_impl_13.h
```diff
@@ -81,13 +81,17 @@ class Client_Impl_13 : public Channel_Impl_13
       void handle(const New_Session_Ticket_13& new_session_ticket);
       void handle(const Key_Update& key_update);
 
+   private:
+      std::optional<Session> find_session_for_resumption();
```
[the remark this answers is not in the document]
reneme: Indeed, this one is odd to me. Will remove it.

src/lib/tls/tls13/tls_extensions_psk.cpp
```diff
@@ -0,0 +1,279 @@
...
+   //    Clients MUST verify that the server's selected_identity is within the
+   //    range supplied by the client, [...].  If these values are not
+   //    consistent, the client MUST abort the handshake with an
+   //    "illegal_parameter" alert.
+   if(id >= ids.size())
+      {
+      throw TLS_Exception(Alert::ILLEGAL_PARAMETER, "PSK identity selected by server is out of bounds");
+      }
+
+   auto cipher_state = std::exchange(ids[id].cipher_state, nullptr);
```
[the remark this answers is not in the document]
reneme: Most likely, yes. I believe I used `std::exchange` to explicitly show that `ids[id].cipher_state` will be reset to `nullptr`. The behaviour of `std::move()` would likely be the same but less "explicit". I don't feel very strongly about that, though.

src/lib/tls/tls13/tls_extensions_psk.cpp
```diff
@@ -0,0 +1,279 @@
...
+   //    indicating a Hash associated with the PSK [...].  If these values
+   //    are not consistent, the client MUST abort the handshake with an
+   //   "illegal_parameter" alert.
+   if(!cipher_state->is_compatible_with(cipher))
+      {
+      throw TLS_Exception(Alert::ILLEGAL_PARAMETER, "PSK and ciphersuite selected by server are not compatible");
+      }
+
+   // destroy cipher states and PSKs that were not selected by the server
+   ids.clear();
```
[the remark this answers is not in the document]
reneme: More than memory efficiency this zaps unused PSK master secrets to avoid keeping unused secrets exposed in memory.

src/lib/tls/msg_session_ticket.cpp
```diff
@@ -51,14 +53,45 @@ std::vector<uint8_t> New_Session_Ticket_12::serialize() const
...
+
+std::optional<uint32_t> New_Session_Ticket_13::early_data_byte_limit() const
    {
-   // TODO: Implement
+   if(!m_extensions.has<EarlyDataIndication>())
+      return std::nullopt;
+
+   const auto ext = m_extensions.get<EarlyDataIndication>();
+   BOTAN_ASSERT_NOMSG(ext->max_early_data_size().has_value());
+   return ext->max_early_data_size().value();
```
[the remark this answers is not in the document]
reneme: That was certainly not intentional. Though, I would argue that `.value()` is much easier to read/understand than the `operator*`. As in: `.value()` somewhat implies that we're dealing with a wrapper type (most likely a `std::optional`), while `operator*` might be anything from a smart pointer, raw pointer, ...

src/lib/tls/tls13/tls_extensions_psk.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+   BOTAN_ASSERT_NOMSG(std::holds_alternative<std::vector<Client_PSK>>(m_impl->psk));
+   return std::get<std::vector<Client_PSK>>(m_impl->psk).empty();
+   }
+
+
+std::unique_ptr<Cipher_State> PSK::select_cipher_state(const PSK& server_psk, const Ciphersuite& cipher)
+   {
+   BOTAN_STATE_CHECK(std::holds_alternative<std::vector<Client_PSK>>(m_impl->psk));
+   BOTAN_STATE_CHECK(std::holds_alternative<Server_PSK>(server_psk.m_impl->psk));
```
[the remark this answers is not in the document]
reneme: Technically, that would be possible. Though, the concept of "an extension being polymorphic depending on the TLS message it appear in" affects other extension types as well (e.g. Key Share). #consistency

src/lib/tls/tls_session.cpp
reneme: dropped.

src/lib/tls/tls_session.h
reneme: For now I just added a remark in the migration guide. Thanks for pointing out this document, by the way!

src/bogo_shim/config.json
lieser: Would of course be nice to switch to TLS 1.3. even than an older session exist. But unless you think the new some API change for that I don't think it is important to have it in the first version.

src/lib/tls/tls13/tls_extensions_psk.cpp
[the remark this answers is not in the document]
lieser: Ok, still not that happy with it but fine with leaving as is for now.

lieser (approved):

PR #2977, opened 2022-05-19
[TLS 1.3] Fuzz Target for Handshake Message Parsing

reneme:
This is a basic fuzz target that shoves the fuzzer input into `Handshake_Layer_13::copy_data()` and tries to parse the messages. We've seen fairly good coverage of the handshake message parsing code with a seed corpus based on the RFC 8448 handshake messages.

~~We weren't able to figure out how to integrate this target into the OSS-Fuzz run, though.~~ OSS-Fuzz seems to pick up the new target but (obviously) fails to find a seed corpus for it.

* [ ] Add `--terminate-on-asserts` in the [OSS-Fuzz build step](https://github.com/google/oss-fuzz/blob/master/projects/botan/build.sh) once this is merged

For CI convenience we left the base commits of [the TLS 1.3 PR](https://github.com/randombit/botan/pull/2922) in here. The relevant fuzz target can be found in this commit: https://github.com/randombit/botan/pull/2977/commits/3151d0d31b8dfd645dec09a0f2ae4380e863c821

### DIFF
--- a/src/fuzzer/tls_13_handshake_layer.cpp
+++ b/src/fuzzer/tls_13_handshake_layer.cpp
@@ -1,0 +1,40 @@
+/*
+* (C) 2022 Jack Lloyd
+* (C) 2002 René Meusel - neXenio GmbH
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include "fuzzers.h"
+#include <botan/internal/tls_handshake_layer_13.h>
+#include <botan/internal/tls_transcript_hash_13.h>
+
+
+namespace {
+
+Botan::TLS::Handshake_Layer prepare(const std::vector<uint8_t>& data)
+   {
+   Botan::TLS::Handshake_Layer hl(Botan::TLS::Connection_Side::CLIENT);
+   hl.copy_data(data);
+   return hl;
+   }
+
+}  // namespace;
+
+
+void fuzz(const uint8_t in[], size_t len)
+   {
+   static Botan::TLS::Default_Policy policy;
+
+   try
+      {
+      std::vector<uint8_t> v(in, in + len);
+      auto hl1 = prepare(v);
+      Botan::TLS::Transcript_Hash_State ths("SHA-256");
+      while (hl1.next_message(policy, ths).has_value()) {};
+
+      auto hl2 = prepare(v);
+      while (hl2.next_post_handshake_message(policy).has_value()) {};
+      }
+   catch(Botan::Exception& e) {}
+   }


lz101010:
It seems the new fuzzing target is indeed picked up - we're guessing this is due to the fact that it starts with `"tls"` which is configured somewhere in https://oss-fuzz.com/ to which we don't have access.

In order to speed up the fuzzers coverage, we added a corpus over at https://github.com/randombit/crypto-corpus/pull/1 which oss-fuzz should then be able to pick up here: https://github.com/google/oss-fuzz/blob/master/projects/botan/Dockerfile#L20

----

Regarding our local test setup, we had to do a little workaround in order to run libFuzzer:
```shell
./configure.py --compiler-cache=ccache --cc=clang --cc-bin=/usr/local/opt/llvm@13/bin/clang++ --build-fuzzer=libfuzzer2 --unsafe-fuzzer-mode --enable-sanitizers=coverage,address,undefined,fuzzer
make -j$(nproc) fuzzers
DYLD_LIBRARY_PATH=. build/fuzzer/tls_13_handshake_layer msg_corpus
```

Notes:
- `libfuzzer2` is essentially a copy of `libfuzzer`, but we couldn't get it to work with the latter
- `msg_corpus` contains the corpus from https://github.com/randombit/crypto-corpus/pull/1

We also visualized the coverage of the corpus (without additional fuzzing) via
```shell
mkdir coverage
./configure.py --compiler-cache=ccache --cc=clang --cc-bin=/usr/local/opt/llvm@13/bin/clang++ --build-fuzzer=libfuzzer2 --unsafe-fuzzer-mode --enable-sanitizers=coverage,address,undefined,fuzzer --with-coverage-info
make -j$(nproc) fuzzers
DYLD_LIBRARY_PATH=. build/fuzzer/tls_13_handshake_layer -runs=0 msg_corpus
gcovr --html --html-details --object-directory=build/obj/lib -r . -o coverage/coverage.html
open coverage/coverage.html
```
See https://github.com/google/fuzzing/issues/41#issuecomment-1031942660 for details.

reneme:
Rebased to master.

src/fuzzer/tls_13_handshake_layer.cpp
```diff
@@ -0,0 +1,40 @@
...
+}  // namespace;
+
+
+void fuzz(const uint8_t in[], size_t len)
+   {
+   static Botan::TLS::Default_Policy policy;
+
+   try
+      {
+      std::vector<uint8_t> v(in, in + len);
```
lz101010, line 31: @reneme I'm unable to push the following change (403) to the vector's type:
```suggestion
Botan::TLS::Handshake_Layer prepare(const Botan::secure_vector<uint8_t>& data)
   {
   Botan::TLS::Handshake_Layer hl(Botan::TLS::Connection_Side::CLIENT);
   hl.copy_data(data);
   return hl;
   }

}  // namespace;


void fuzz(const uint8_t in[], size_t len)
   {
   static Botan::TLS::Default_Policy policy;

   try
      {
      Botan::secure_vector<uint8_t> v(in, in + len);
```

Can you approve the suggestion so the build can hopefully pass?

lz101010: I decided to open a new PR anyway, the fuzzer now works as intended: #3013 

I suggest closing this PR (I apparently can't do this myself) in favor of the other one, because (a) the other one is ready to merge (b) I can rebase the other PR if something else changes in the meantime.

PR #2980, opened 2022-05-23
./configure.py --terminate-on-asserts

reneme:
As [discussed](https://github.com/randombit/botan/discussions/2978), this adds `./configure.py --terminate-on-asserts` causing the library to `std::abort` when an assertion is not met.

The code base seems to use `BOTAN_ASSERT*` in some places where `BOTAN_ARG_CHECK` or `BOTAN_STATE_CHECK` would be more appropriate. Hence, I enabled the new option for the coverage CI run in the hope to catch most of them.

### DIFF
--- a/src/lib/pubkey/dsa/dsa.cpp
+++ b/src/lib/pubkey/dsa/dsa.cpp
@@ -23,10 +23,21 @@ namespace Botan {
 /*
 * DSA_PublicKey Constructor
 */
+
+DSA_PublicKey::DSA_PublicKey(const AlgorithmIdentifier& alg_id,
+                             const std::vector<uint8_t>& key_bits) :
+   DL_Scheme_PublicKey(alg_id, key_bits, DL_Group_Format::ANSI_X9_57)
+   {
+   BOTAN_ARG_CHECK(group_q().bytes() > 0, "Q parameter must be set for DSA");
+   }
+
+
 DSA_PublicKey::DSA_PublicKey(const DL_Group& grp, const BigInt& y1)
    {
    m_group = grp;
    m_y = y1;
+
+   BOTAN_ARG_CHECK(grp.q_bytes() > 0, "Q parameter must be set for DSA");
    }
 
 /*
--- a/src/lib/pubkey/dsa/dsa.cpp
+++ b/src/lib/pubkey/dsa/dsa.cpp
@@ -52,7 +52,10 @@ DSA_PrivateKey::DSA_PrivateKey(RandomNumberGenerator& rng,
    if(x_arg == 0)
       m_x = BigInt::random_integer(rng, 2, group_q());
    else
+      {
+      BOTAN_ARG_CHECK(m_x.bits() <= m_group.q_bits(), "x must not be larger than q");
       m_x = x_arg;
+      }
 
    m_y = m_group.power_g_p(m_x, m_group.q_bits());
    }
@@ -61,6 +64,7 @@ DSA_PrivateKey::DSA_PrivateKey(const AlgorithmIdentifier& alg_id,
                                const secure_vector<uint8_t>& key_bits) :
    DL_Scheme_PrivateKey(alg_id, key_bits, DL_Group_Format::ANSI_X9_57)
    {
+   BOTAN_ARG_CHECK(m_x.bits() <= m_group.q_bits(), "x must not be larger than q");
    m_y = m_group.power_g_p(m_x, m_group.q_bits());
    }
 
--- a/src/lib/pubkey/dsa/dsa.cpp
+++ b/src/lib/pubkey/dsa/dsa.cpp
@@ -48,12 +48,13 @@ DSA_PrivateKey::DSA_PrivateKey(RandomNumberGenerator& rng,
                                const BigInt& x_arg)
    {
    m_group = grp;
+   BOTAN_ARG_CHECK(x_arg.is_positive(), "x must be positive");
 
    if(x_arg == 0)
       m_x = BigInt::random_integer(rng, 2, group_q());
    else
       {
-      BOTAN_ARG_CHECK(m_x.bits() <= m_group.q_bits(), "x must not be larger than q");
+      BOTAN_ARG_CHECK(m_x < m_group.get_q(), "x must not be larger than q");
       m_x = x_arg;
       }
 
@@ -64,7 +65,8 @@ DSA_PrivateKey::DSA_PrivateKey(const AlgorithmIdentifier& alg_id,
                                const secure_vector<uint8_t>& key_bits) :
    DL_Scheme_PrivateKey(alg_id, key_bits, DL_Group_Format::ANSI_X9_57)
    {
-   BOTAN_ARG_CHECK(m_x.bits() <= m_group.q_bits(), "x must not be larger than q");
+   BOTAN_ARG_CHECK(m_x > 0, "x must be greater than zero");
+   BOTAN_ARG_CHECK(m_x < m_group.get_q(), "x must not be larger than q");
    m_y = m_group.power_g_p(m_x, m_group.q_bits());
    }
 
--- a/src/lib/utils/assert.cpp
+++ b/src/lib/utils/assert.cpp
@@ -6,7 +6,13 @@
 */
 
 #include <botan/exceptn.h>
+#include <botan/build.h>
 #include <sstream>
+
+#if defined(BOTAN_TERMINATE_ON_ASSERTS)
+  #include <cstdlib>
+  #include <iostream>
+#endif
 
 namespace Botan {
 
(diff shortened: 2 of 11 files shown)


src/lib/utils/assert.cpp
```diff
@@ -6,7 +6,13 @@
 */
 
 #include <botan/exceptn.h>
+#include <botan/build.h>
 #include <sstream>
+#include <cstdlib>
```
randombit, line 11: Is the `cstdlib` required in the general case or only with `BOTAN_TERMINATE_ON_ASSERTS`?

randombit (approved):
LGTM. Still a few CI failures which I assume are missing cases but once CI is green good to merge

reneme:
The assertion in the coverage build job seems like an actual bug, actually.

There are a few BoGo tests that send a prefix string to the TLS server under test. E.g. BoGo test `HttpGET` sends `GET / HTTP/1.0\n"` and expects the TLS server under test to notice that and fail. The TLS 1.2 record implementation detects this situation and throws a `TLS_Exception`:

https://github.com/randombit/botan/blob/15228710faef1b85aaad0cad4f207b8b6dfcdf74/src/lib/tls/tls12/tls_record.cpp#L361-L368

The TLS channel then catches this exception and proceeds to sending a TLS alert in response. `TLS_Channel::send_alert()` used to swallow all exceptions that happened while trying to send an alert:

https://github.com/randombit/botan/blob/15228710faef1b85aaad0cad4f207b8b6dfcdf74/src/lib/tls/tls12/tls_channel_impl_12.cpp#L573-L577

But now, the `sequence_numbers()` accessor is called downstream, causing an assertion because the server didn't actually finish initializing and `m_sequence_numbers` is still a nullptr:

https://github.com/randombit/botan/blob/15228710faef1b85aaad0cad4f207b8b6dfcdf74/src/lib/tls/tls12/tls_channel_impl_12.cpp#L548-L552

I see two possible fixes:

* initialize `m_sequence_numbers` before any data is processed, or
* detect this situation and handle it

But is it actually legal for a TLS server to send a TLS alert as their first flight?

reneme:
> But is it actually legal for a TLS server to send a TLS alert as their first flight?

At least for Botan's (D)TLS 1.2 implementation it doesn't seem to be legal for the TLS server to send messages before it successfully received a Client Hello. Only after that do we know whether we speak TLS or DTLS.

Hence, let's just not send TLS alerts in the very early error case described above: https://github.com/randombit/botan/pull/2980/commits/9c70d9d268241120104ad780bb3783e909a6750a

src/lib/pubkey/dsa/dsa.cpp
```diff
@@ -23,10 +23,21 @@ namespace Botan {
...
+   }
+
+
 DSA_PublicKey::DSA_PublicKey(const DL_Group& grp, const BigInt& y1)
    {
    m_group = grp;
    m_y = y1;
+
+   BOTAN_ARG_CHECK(grp.q_bytes() > 0, "Q parameter must be set for DSA");
    }
```
reneme, line 41: That is suppose to fix asserts in `PK_Verifier` for a DSA public key in the `cert` corpus/fuzzer. I left the assert untouched and instead introduce sanity checks at construction time of the DSA public key.
@randombit I could use a little help with this, as I'm not sure whether other things should also be validated here.

By the way: The currently still failing assert in the `pkcs8` fuzzer is related to a DSA private key:

```
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007ff81db8e00e libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007ff81dbc41ff libsystem_pthread.dylib`pthread_kill + 263
    frame #2: 0x00007ff81db0fd24 libsystem_c.dylib`abort + 123
    frame #3: 0x00000001009a9778 pkcs8`Botan::assertion_failure(char const*, char const*, char const*, char const*, int) + 696
    frame #4: 0x000000010048ed39 pkcs8`Botan::Montgomery_Exponentation_State::exponentiation(Botan::BigInt const&, unsigned long) const + 1001
    frame #5: 0x000000010048fc46 pkcs8`Botan::monty_execute(Botan::Montgomery_Exponentation_State const&, Botan::BigInt const&, unsigned long) + 22
    frame #6: 0x0000000100651218 pkcs8`Botan::DL_Group_Data::power_g_p(Botan::BigInt const&, unsigned long) const + 72
    frame #7: 0x000000010065127b pkcs8`Botan::DL_Group::power_g_p(Botan::BigInt const&, unsigned long) const + 59
    frame #8: 0x0000000100660e4f pkcs8`Botan::DSA_PrivateKey::DSA_PrivateKey(Botan::AlgorithmIdentifier const&, std::__1::vector<unsigned char, Botan::secure_allocator<unsigned char> > const&) + 767
    frame #9: 0x000000010080913c pkcs8`std::__1::__unique_if<Botan::DSA_PrivateKey>::__unique_single std::__1::make_unique<Botan::DSA_PrivateKey, Botan::AlgorithmIdentifier const&, std::__1::vector<unsigned char, Botan::secure_allocator<unsigned char> > const&>(Botan::AlgorithmIdentifier const&, std::__1::vector<unsigned char, Botan::secure_allocator<unsigned char> > const&) + 76
    frame #10: 0x000000010080851b pkcs8`Botan::load_private_key(Botan::AlgorithmIdentifier const&, std::__1::vector<unsigned char, Botan::secure_allocator<unsigned char> > const&) + 1003
    frame #11: 0x000000010083fe17 pkcs8`Botan::PKCS8::(anonymous namespace)::load_key(Botan::DataSource&, std::__1::function<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > ()> const&, bool) + 471
    frame #12: 0x0000000100840587 pkcs8`Botan::PKCS8::load_key(Botan::DataSource&) + 231
    frame #13: 0x0000000100002c8f pkcs8`fuzz(unsigned char const*, unsigned long) + 255
    frame #14: 0x0000000100002b8a pkcs8`LLVMFuzzerTestOneInput + 26
    frame #15: 0x0000000100003483 pkcs8`(anonymous namespace)::fuzz_files(char**) + 899
    frame #16: 0x0000000100002eb7 pkcs8`main + 231
    frame #17: 0x0000000101df551e dyld`start + 462
```

src/lib/pubkey/dsa/dsa.cpp
```diff
@@ -50,6 +64,7 @@ DSA_PrivateKey::DSA_PrivateKey(const AlgorithmIdentifier& alg_id,
                                const secure_vector<uint8_t>& key_bits) :
    DL_Scheme_PrivateKey(alg_id, key_bits, DL_Group_Format::ANSI_X9_57)
    {
+   BOTAN_ARG_CHECK(m_x.bits() <= m_group.q_bits(), "x must not be larger than q");
```
reneme, line 67: I have a hunch that this check should better be constant time, because:

https://github.com/randombit/botan/blob/15228710faef1b85aaad0cad4f207b8b6dfcdf74/src/lib/math/numbertheory/monty_exp.cpp#L102-L105

src/lib/pubkey/dsa/dsa.cpp
randombit: I think this assert would trigger if you load a DSA key with a secret key that is greater than the group order.

This likely affects other key types as well.

I think we need to rethink key validation somewhat. Right now all validation is dumped on `check_key`, because some of the checks (testing primality and ECC point validity) are expensive. But most are simple range checks and could be cheaply done during construction or deserialization.

This is resolved with 1735f95 right?

src/lib/pubkey/dsa/dsa.cpp
randombit: I think this is safe enough to not be constant time in this context. `bits` itself is constant time, and if we reject that means the key itself is just manifest invalid. So there is no real info leak.

We should check `x <= q` though rather than just the bit size, since `x == q` would be accepted, but isn't a valid DSA key.

randombit (approved):
LGTM, modulo changing DSA key check to use value instead of bit length.

src/lib/pubkey/dsa/dsa.cpp
```diff
@@ -50,6 +65,8 @@ DSA_PrivateKey::DSA_PrivateKey(const AlgorithmIdentifier& alg_id,
                                const secure_vector<uint8_t>& key_bits) :
    DL_Scheme_PrivateKey(alg_id, key_bits, DL_Group_Format::ANSI_X9_57)
    {
+   BOTAN_ARG_CHECK(m_x > 0, "x must be greater than zero");
+   BOTAN_ARG_CHECK(m_x < m_group.get_q(), "x must not be larger than q");
```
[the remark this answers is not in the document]
reneme: Nevertheless, I propose this is something for a follow-up PR. I wanted to keep this to the minimum to just start enforcing assertions in the first place.

reneme:
I rebased to master to let the CI run once again after this was stale for more than a month. Once that comes back clean, I'll merge.

PR #2983, opened 2022-05-31
[TLS 1.3] Post-Quantum Readiness via Hybrid Key Exchange

reneme:
### Pull-Request Dependencies

* https://github.com/randombit/botan/pull/2922
* https://github.com/randombit/botan/pull/2982

Both change sets are currently also displayed in this pull request. Hence, review and merge of those should make this PR fairly small (~700 lines added).

### TODO

* [ ] Add a "hybrid" key exchange method to the TLS policy
* [ ] Add a `minimum_kyber_group_size()` to the TLS policy (??)

### Description

This enables the TLS 1.3 implementation to perform hybrid key exchanges using a classical KEX (ECDH or X25519) and a post-quantum KEM (Kyber or Kyber90s). The implementation is based on [this IETF draft](https://datatracker.ietf.org/doc/html/draft-ietf-tls-hybrid-design-04) and the group identifiers for the Key Share extension are taken from [OQS](https://github.com/open-quantum-safe/openssl/blob/OQS-OpenSSL_1_1_1-stable/oqs-template/oqs-kem-info.md).

### Demo

```bash
./configure.py                 \
    --build-targets=static,cli \
    --minimized-build          \
    --without-documentation    \
    --enable-modules=tls13,tls13_pqc,auto_rng,system_rng,chacha20poly1305,curve25519,kyber

make -j$(nproc) cli

./botan tls_client                                           \
    --policy=src/tests/data/tls-policy/default_tls13_pqc.txt \
    --port=443                                               \
    kms.eu-central-1.amazonaws.com
```

Using the snippets above, one should obtain a TLS 1.3 connection to Amazon's KMS endpoint (that is [already PQC-enabled](https://docs.aws.amazon.com/kms/latest/developerguide/pqtls.html) using their [s2n-tls library](https://github.com/aws/s2n-tls/)). Simply typing "GET / HTTP/1.1" [Enter][Enter] should yield an (admittedly useless) "Bad Request" response.

The PQC TLS 1.3 policy file passed to the CLI uses X25519/Kyber512 as the hybrid key exchange scheme.

PR #2989, opened 2022-06-08
[TLS 1.3] Overhaul TLS::Policy

reneme:
### Cipher suite related: `allowed_ciphers()`, `allowed_macs()`, ...

My current working assumption: TLS 1.3 cipher suites are simply subject to those policy configurations as well. 

* `allowed_key_exchange_methods()` needs to be extended for [hybrid key exchange for PQC](https://github.com/randombit/botan/pull/2983).

### `only_resume_with_exact_version()`

Sessions that originally negotiated TLS 1.2 cannot be resumed as TLS 1.3 and vice versa in the current implementation. I'm not sure whether this would even be technically possible. Since TLS protocol versions prior to 1.2 were removed, this will not have an effect.

### `record_size_limit()`

Currently, the record size limit extension is supported by the TLS 1.3 implementation only. Hence, it cannot be offered if our Botan-client signals willingness to negotiate either 1.2 or 1.3. Otherwise a peer might choose to negotiate TLS 1.2 and use the record size limit extension which would break. That's a bit of a bummer (and a non-obvious cause for user frustration), but the extension is fairly niche anyway. We mainly implemented it for compliance with the RFC 8448 tests that use it.

### Removed `use_extended_master_secret()`

TLS 1.2 enables this extension by default (and ignores the setting of this policy config). TLS 1.3 connections don't have a notion of "extended master secret" anymore. Hence the removal.

### `key_exchange_groups_to_offer()`

TLS 1.3 client specific: the groups (in order of preference) that should be offered in the initial Client Hello's KeyShare extension. By default, this will offer the first supported group (as returned by `Policy::key_exhange_groups()`) but applications may adjust this. Somewhat obvious: only groups that are supported can be offered.

### `choose_key_exchange_group()` has a new parameter

This method existed before. It is responsible for deciding which key exchange group a peer should use given the counter-party's supported groups. Now, it additionally obtains a list of groups that the counter-party send TLS 1.3 key share offers for. When called from the TLS 1.2 implementation the new parameter is always empty.
With that, it is up to the application whether a Hello Retry Request should be issued (e.g. to choose a PQ-safe group over an offered conventional group). By default, this will try to avoid Hello Retry Requests by choosing a group that was offered by the counter-party.

### `tls_13_middlebox_compatibility_mode()`

Lets a TLS 1.3 client behave as specified in [RFC 8446 Appendix D.4](https://datatracker.ietf.org/doc/html/rfc8446#appendix-D.4). TL;DR: Make the TLS 1.3 handshake appear to be a TLS 1.2 session resumption to fool non-compliant middleboxes. This is on by default and (obviously) has no effect on the TLS 1.2 implementation. Not added in this PR.

### `hash_hello_random()`

Whether or not the random string in client/server hello should be hashed with SHA-256 after being drawn from the RNG. This is enabled by default and was added to allow deterministic testing (i.e. RFC 8448). Not added in this PR.

### Documentation

Added `@note` paragraphs for policy configs that have no effect for either TLS 1.2 or 1.3 respectively.

### DIFF
--- a/src/lib/tls/tls_policy.h
+++ b/src/lib/tls/tls_policy.h
@@ -117,9 +117,16 @@ class BOTAN_PUBLIC_API(2,0) Policy
 
       /**
       * Select a key exchange group to use, from the list of groups sent by the
-      * peer. If none are acceptable, return Group_Params::NONE
+      * peer. In TLS 1.3 handshakes the peer might have provided cryptographic material
+      * for a subset of its available groups. Choosing a group for which no share was
+      * provided will result in an additional round trip. If none are acceptable, return
+      * Group_Params::NONE.
+      *
+      * By default this will try to optimize for less round trips even if this results
+      * in the usage of a less preferred group.
       */
-      virtual Group_Params choose_key_exchange_group(const std::vector<Group_Params>& peer_groups) const;
+      virtual Group_Params choose_key_exchange_group(const std::vector<Group_Params>& supported_by_peer,
+                                                     const std::vector<Group_Params>& offered_by_peer) const;
 
       /**
       * Allow renegotiation even if the counterparty doesn't
--- a/src/lib/tls/tls_policy.h
+++ b/src/lib/tls/tls_policy.h
@@ -93,24 +93,25 @@ class BOTAN_PUBLIC_API(2,0) Policy
       virtual std::vector<Group_Params> key_exchange_groups() const;
 
       /**
-      * TLS 1.3 specific
       * Return a list of groups to provide prepared key share offers in the
       * initial client hello for. Groups in this list must be reflected in
       * key_exchange_groups() and in the same order. By default this returns
       * the most preferred group from key_exchange_groups().
       * If an empty list is returned, no prepared key share offers are sent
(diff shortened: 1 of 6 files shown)


src/lib/tls/tls_policy.h
```diff
@@ -290,7 +308,11 @@ class BOTAN_PUBLIC_API(2,0) Policy
       *   up to the server Finished, implementations which support both TLS 1.3
       *   and earlier versions SHOULD indicate the use of the Extended Master
       *   Secret extension in their APIs whenever TLS 1.3 is used.
+      *
+      * @deprecated This policy configuration does not have an effect anymore, as
+      *             extended master secret is always enabled by the implementation.
       */
+      BOTAN_DEPRECATED("extended master secret is always enabled for TLS 1.2 and not applicable for TLS 1.3")
       virtual bool use_extended_master_secret() const;
```
randombit, line 316: This being the case we should just remote it as a toggle

randombit (approved):
Similar comment re the callbacks PR - for EMS support, if it is always enabled and the policy does nothing, we might as well just remove it entirely in 3.0. Otherwise LGTM

PR #2998, opened 2022-06-20
Update BSI policy

lieser:
- prohibit `ec_h2c`
- prohibit `blake2mac` (was implicitly prohibited before #2930 was merged)
- explicitly prohibited sub modules that there missing (for consistency)
- some alphabetical reordering

/cc @securitykernel


randombit:
LGTM, thanks

PR #3000, opened 2022-06-23
Fix mul128.h to support MSVC ARM64 target

nursik:
I assume this is the only change required to bring ARM64 support for Windows. Build was done using VCPKG (arm64-windows triplet). Tests are OK.


nursik:
I would like to know if there is any progress or any concerns about the change?

vladimir-aubrecht (approved):
Looks good!

vladimir-aubrecht:
@randombit Do you think we could get this tiny change merged? 😊

randombit:
Sorry about the delay here

PR #3003, opened 2022-07-11
Documentation: Module Tree

reneme:
# Description

As a user of the library I often struggle with managing the available modules, especially when trying to select the ones to enable/disable.
Hence, this is a first step towards a more accessible module system by providing auto-generated documentation of the available modules, their inter-dependencies, platform requirements and contents.

For that, I added short documentation-friendly "module infos" to each module's `info.txt` file and rendered them into the [Doxygen documentation as `@defgroup` tags](https://www.doxygen.nl/manual/commands.html#cmddefgroup).

For instance (AES-NI's `info.txt`):
```
<defines>
AES_NI -> 20131128
</defines>

<module_info>
name -> "AES-NI"
brief -> "AES using AES-NI and SIMD instructions"
</module_info>

<isa>
sse2
ssse3
aesni
</isa>
```

Potentially, the same information could be used to improve the UX of `./configure.py --list-modules` and related configuration CUI features like error messages.

# Demo

Here are some screenshots of the resulting Doxygen renderings:

## Module Overview Page

This renders a tree-view of all modules and their sub-modules. Optionally, the module author can add a short description of the module's content. Note the new "Modules" main-menu entry that Doxygen automatically creates for this view.

Screenshot 2022-07-11 152915

## Sub-Module Description Page

Each module is represented by a description page listing the contained source files and providing some extra information on the module: e.g. which other modules it depends on. Also note the bread crumbs describing the sub-module relationships.

Screenshot 2022-07-11 153342
 
## Platform-specific Requirements

The platform-requirements (that are defined in the `info.txt`) can also be rendered into the detailed module description.

Screenshot 2022-07-11 153904

# Limitations

## Sub-Module Relationship

The module list is still fairly hard to grasp, I feel. Improving the grouping of specific modules might be really helpful here. For instance, create a "virtual" module for "codec" to group Base64, Base58 and Base32.

Also, the sub-module dependency is currently not detected for modules that reside in sub-sub-folders without an intermediate `info.txt`. For instance, the "Kyber" `info.txt` files are in `pubkey/kyber/kyber_90s`. But there is no `info.txt` at the `pubkey/kyber` level. As a result, `Kyber 90s` is currently treated as a top-level module. That might even cause functional bugs, as it probably breaks the implicit "parent modules are always required" assumption. Edit: e2ccb17db5636570dc164697833ba2a0af6b6055 tacles that.

## Assign APIs to Modules

It would be great if Doxygen could annotate APIs (classes, functions, enum, ...) to their respective module (or "group" in Doxygen-terms). Technically, we could do that manually, by annotating each item with the correct `@ingroup ...` tag, but that doesn't seem feasible.

As an example, I annotated `Certificate_Store_MacOS` resulting in the module bread-crumbs under the class' title as well as a list of contained APIs in the module description page.

Screenshot 2022-07-11 155826

Screenshot 2022-07-11 160159

It remains to be seen whether there is an automatic approach to that without auto-altering the header files.

## Detailed Documentation

Currently, the `info.txt` `<module_info>` section doesn't provide a means to add more detailed descriptions for a specific module. Mainly, because I don't think it's desirable to add extensive text in there. If we'd want to add richer documentation to modules we'd need to see how those could be outsourced to dedicated documentation files.

# RFC 

What do you think about this attempt to improve the documenntation of Botans modules?

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -795,10 +795,13 @@ class InfoObject(object):
         self.lives_in = dirname
         if basename == 'info.txt':
             (obj_dir, self.basename) = os.path.split(dirname)
-            if os.access(os.path.join(obj_dir, 'info.txt'), os.R_OK):
-                self.parent_module = os.path.basename(obj_dir)
-            else:
-                self.parent_module = None
+            self.parent_module = None
+
+            while obj_dir:
+                if os.access(os.path.join(obj_dir, 'info.txt'), os.R_OK):
+                    self.parent_module = os.path.basename(obj_dir)
+                    break
+                (obj_dir, _) = os.path.split(obj_dir)
         else:
             self.basename = basename.replace('.txt', '')
 
--- a/configure.py
+++ b/configure.py
@@ -19,6 +19,8 @@ CPython 2.5 and earlier are not supported.
 On Jython target detection does not work (use --os and --cpu).
 """
 
+# pylint: disable=too-many-lines
+
 import collections
 import copy
 import json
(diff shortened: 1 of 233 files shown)


randombit:
I still need to review the code, but the concept is great and I would certainly encourage this kind of thing.

Issue #2736, opened 2021-05-04
X.509 certificates with non-critical name constraints extension are being rejected


robertsuska:
Hi,

I have a question (not a bug to report) regarding Botan's behavior:

What is the reasoning behind rejecting X.509 certificates that contain a non-critical name constraints extension?

---

[RFC 5280](https://www.rfc-editor.org/rfc/rfc5280.txt) (article 4.2.1.10) says

_"Conforming CAs MUST mark this extension as critical and SHOULD NOT impose name constraints on the x400Address, ediPartyName, or registeredID name forms."_

and

_"If a name constraints extension that is marked as critical imposes constraints on a particular name form, and an instance of that name form appears in the subject field or subjectAltName extension of a subsequent certificate, then the application MUST either process the constraint or reject the certificate."_

but it does not say anything about the case, when the encountered extension is not marked as critical.

Botan's behavior makes sense given that such certificates should not really exist, but I am still curious since I've tested this with other libraries, namely OpenSSL, pyOpenSSL, GnuTLS, and MbedTLS, and none of them behaves the same way.

---

I am using Botan in version 2.17.2.
The condition for criticality of the extension is [here](https://github.com/randombit/botan/blob/2.17.2/src/lib/x509/x509_ext.cpp#L660).

randombit:
I'm pretty sure this was done based on the "Conforming CAs MUST ..." wording. But if other libraries accept non-criticial we probably should as well, since they are relatively likely to be found in practice if we are the one oddball who doesn't accept.

randombit:
CC @securitykernel @fweissberg  We have an explicit test from the BSI test suite that added in #1363 that non-critical name constraints are rejected. Given that other implementations do accept (I've verified that OpenSSL accepts non-critical) I'm going to modify the BSI test. My read of RFC 5280 is that the CA is non-conforming, not to mention idiotic, to issue a non-critical name constraint, but there does not seem to be an intrinsic problem with accepting it on the verification side.

randombit:
Fixed in master and 2.18.1 thanks for reporting this

PR #3006, opened 2022-07-15
Add a Note on a BSI Test Description (re: GH #2736)

reneme:
The expected outcome of this BSI test was changed a while back. To avoid future confusion I added a note to the test's description.txt.


randombit (approved):
:+1:

PR #3007, opened 2022-07-15
CI: Use GitHub Actions for Windows builds

reneme:
# Description

This did escalate quite a bit from "just adding windows" to "overhauling the CI". So _please_ don't hesitate to call me out for going too far with it. I'm using the opportunity to do a deep-dive on GitHub Actions and splitting this up into several PRs or rolling back most of the changes shouldn't be too hard.

As a side-note: While experimenting, I found a few buggy build configurations and documented them below. I would suggest to keep them in mind and fix them after this is integrated.

The pipeline as it is now should provide at least as much coverage as before, plus some extra configurations. Some new bits and pieces bump the minimal requirements up, though. Below is an overview:

## Requirements

* Boost 1.69 (released on christmas 2018)
  * no more linking of `libboost_system` (header-only as of 1.69)

* Visual Studio 2019
  * because of `/external:I`

* Not sure about clang's `--system-header-prefix=`
  * though that is only used in the CI not in production

### Basic Matrices

* Windows 32-Bit/64-Bit, Linux, macOS
* `static`, `shared`, `amalgamation`
* including most external dependency integrations on all platforms (particularly boost)
* some configs are disabled (see "Build Failures" below)

### Specials and Metrics

* coverage build (with basically everything included)
* MinGW smoke test
* sanitizer (windows)
* minimized
* bsi
* fuzzers
* valgrind
* emscripten
* baremetal
* lint

### Cross-Compilation

* same configurations as before

### `setup_gh_actions.ps1/sh`

This didn't change much. The windows flavour of the script basically just installs `jom` and `sccache`. One thing was added: The scripts now find out where the compiler cache is located and communicates this upstream via an environment variable. Hence, no more hard-coded cache paths in `actions/cache`.

### `setup_gh_actions_after_vsvars.ps1`

A new script installing 3rd-party dependencies via `nuget` (is there a better way?) after the Visual Studio environment was set up. That setup should also facilitate caching the windows dependencies, as `nuget` takes like 4 to 5 minutes to install `boost`. -.-

### Multiple Matrices

Windows brought more complexity into the `ci.yml` than expected. Mainly due to platform-dependent setup steps, handling `vcvarsall.bat`, extra build configuration variables (`make_tool`, `arch`) and their inter-dependency. Also the list of build configurations became hard to grasp, so I grouped the configurations into several build matrices. As a side effect, we can now actually use the "matrix" feature effectively. E.g. to build `static`, `shared` and `amalgamation` on all major platforms.

#### VS-Shell

This is a [small 3rd-party action](https://github.com/egor-tensin/vs-shell) by @egor-tensin. I figured that it would be easier to copy the code and adapt it here, instead of relying on the 3rd-party repo. It basically locates `vsvarsall.bat` and slurps it into the build agent's environment. Currently, the newest Visual Studio version available is selected. Though, it should be extensible to configure older VS versions as well.

#### Setup Build Agent

This is a tiny custom action that I extracted to re-use the platform dependent agent setup steps. Mainly, it takes care of interacting with the `setup_gh_actions` script, set up compiler caching and calling VS-Shell (see above) on Windows.

### `ci_build.py`

As the only new thing, this script can now auto-detect `sccache` on Windows machines.

## Production Code Changes

To make this compile, a few changes to the production code were required. Most notably, the `boost` module does not link against `libboost_system` anymore. This boost library is [effectively header-only since Boost 1.69](https://www.boost.org/doc/libs/1_70_0/libs/system/doc/html/system.html#changes) and only shipped with an empty compatibity library we were linking against. For reference: Boost 1.69 [was released in the end of 2018](https://www.boost.org/users/history/version_1_69_0.html) making this effectively the oldest supported version.

## Build Infrastructure Changes

This introduces a notion of "system headers" (`-isystem` on gcc/clang and `/external:I` on msvc). That way, we can suppress warnings in external header files like boost.

# TODO

* [x] Remove trailing dash from cache restore key
* [x] Make sure Boost works on windows
* [x] Add older versions of Visual Studio to the pipeline
* [x] Clean-up the commit history
* [x] Sunset AppVeyor

## TODO -- Build Failures in Specific Configurations
* [ ] Argon2id [seems to fail](https://github.com/randombit/botan/runs/7390426222?check_suite_focus=true#logs) on Windows 32 bit builds
* [ ] 32 bit build on windows causes an [unreachable code warning](https://github.com/randombit/botan/runs/7390425958?check_suite_focus=true#step:8:1257)
* [ ] [macOS fails to build](https://github.com/randombit/botan/runs/7390425464?check_suite_focus=true#step:8:105) with gcc (lacking `BOTAN_PARALLEL_SIMD_FOR`)
* [ ] [Amalgamation build fails](https://github.com/randombit/botan/runs/7390903061?check_suite_focus=true#step:8:55) on ubuntu (as well as [windows](https://github.com/randombit/botan/runs/7390903878?check_suite_focus=true#step:8:485) and [macOS with clang](https://github.com/randombit/botan/runs/7390903422?check_suite_focus=true#step:8:70)) with clang (abstract class marked `final`)

# Nice to have

* [ ] Clang sanitizers
* [ ] Custom error/warning messages in the Checks summary
* [x] Maybe group individual build jobs
   * e.g. "standard builds" (static, shared, amalg.), "analysis" (coverage, static analy. linting, ...), "cross compilation"
* [ ] Maybe have a "smoke test" stage run first, and deeper analysis later, to save on free-tier GitHub runners?
* [ ] Cache installed dependency (particularly `nuget install boost` takes long)


randombit:
Thanks for looking into this. AppVeyor's serial execution really hurts CI times and as a result over time the number of Windows configurations being tested has been pruned.

reneme:
That's pretty much done from my side, now. 

If you wish, @randombit, I could make that into three PRs, namely:

1. Changes somewhat unrelated to the CI:
   * [stop linking against legacy boost-system library](https://github.com/randombit/botan/pull/3007/commits/7ba456b8eef5e7f43047ebe856357c0fa7629491)
   * [add compiler support for system header includes](https://github.com/randombit/botan/pull/3007/commits/951ab6d25c60432489221546432f499fd54e96fd)
2. CI Refactorings and Improvements
   * [auto-detect sccache in CI build script](https://github.com/randombit/botan/pull/3007/commits/26fe03034ed2a6c408dbd3ae9f5ef3d30499d28c)
   * [cancel CI runs on frequent consecutive pushes](https://github.com/randombit/botan/pull/3007/commits/502062d3abff893ece4cec6f3ddbd7b1c867f050)
   * [Refactor: find compiler cache directory dynamically](https://github.com/randombit/botan/pull/3007/commits/6cdcddc0b43c63c42b59404903e3407f500ef835)
   * [Refactor: extract build agent setup into custom action](https://github.com/randombit/botan/pull/3007/commits/19e097e2407ccff1ab1ec6384c62cddfe44ad13a)
   * [Refactor: compartmentalize and matrix-ize build configs](https://github.com/randombit/botan/pull/3007/commits/67a1f18adf2131e6fb1b16265d492d8fc0bf2a80)
3. Moving Windows build configurations to Github Actions
   * [add vs-shell actions by Egor Tensin](https://github.com/randombit/botan/pull/3007/commits/d6910f788ca60d697be0f44d090c6a1095c8249e)
   * [add windows build configurations](https://github.com/randombit/botan/pull/3007/commits/5822e38fb637d8a86fb2a21d118881b62757fdae)
   * [add MinGW build configuration](https://github.com/randombit/botan/pull/3007/commits/7c4ade8665b8e0758b3c4f04d9007146d2ee1125)
   * [build with boost for basic configurations](https://github.com/randombit/botan/pull/3007/commits/f0dd6737da0a89e77648fdbf09e7e6397ec2a231)
   * [remove AppVeyor configurations and docs](https://github.com/randombit/botan/pull/3007/commits/e3ed2b47fff8fafc60fb07da4a3e21bcb7350c27)
   * [remove mentions of Travis CI](https://github.com/randombit/botan/pull/3007/commits/fc4f04fe63d6eedf07be508858fa0dcdccbf80ba)

.github/workflows/ci.yml
```diff
@@ -14,85 +14,210 @@ on:
...
+      fail-fast: false
+
+      matrix:
+        # basic configuration combinations should run on all target platforms
+        target:    [ static, shared, amalgamation ]
+        arch:      [ x86_64, x86 ]
+        host_os:   [ windows-2022, windows-2019 ]
+
+        exclude:
+          # fails due to "abstract class marked final" (see PR description) / TODO
```
reneme, line 36: Reference to PR should either be removed or backed with a link to this PR.

.github/workflows/ci.yml
```diff
@@ -14,85 +14,210 @@ on:
...
+    strategy:
+      fail-fast: false
+
+      matrix:
+        # basic configuration combinations should run on all target platforms
+        target:   [ static, shared, amalgamation ]
+        compiler: [ clang, gcc ]
+
+        exclude:
+          # fails due to "abstract class marked final" (see PR description) / TODO
```
reneme, line 68: Reference to PR should either be removed or backed with a link to this PR.

.github/workflows/ci.yml
```diff
@@ -14,85 +14,210 @@ on:
...
+          - compiler: clang
+            target: amalgamation
+          # fails (see PR description) / TODO
+          - compiler: gcc
```
reneme, line 101: References to PR should either be removed or backed with a link to this PR.

readme.rst
```diff
@@ -36,14 +36,6 @@ what is new. Botan is also available through most
...
 such as Fedora, Debian, Arch and Homebrew.
 
-.. image:: https://api.travis-ci.com/randombit/botan.svg?branch=master
-    :target: https://travis-ci.com/github/randombit/botan
-    :alt: Travis CI status
-
-.. image:: https://ci.appveyor.com/api/projects/status/n9f94dljd03j2lce/branch/master?svg=true
-    :target: https://ci.appveyor.com/project/randombit/botan/branch/master
-    :alt: AppVeyor CI status
-
```
reneme, line 46: We probably want to add a GitHub Actions badge here.

src/lib/utils/boost/info.txt
```diff
@@ -5,5 +5,10 @@ BOOST_ASIO -> 20131228
 load_on vendor
 
 <libs>
-all -> boost_system
+linux -> rt
+mingw -> ws2_32
```
reneme, line 9: It turned out that MinGW doesn't link without additionally adding `wsock32`. From my understanding wsock32 is a backward-compatibility wrapper around `ws2_32` though. For now, I added the include as an `--ldflags` addition in `ci_build.py`.

src/lib/utils/boost/info.txt
```diff
@@ -5,5 +5,10 @@ BOOST_ASIO -> 20131228
...
+windows -> ws2_32
+haiku -> network
+solaris -> socket,nsl
+qnx -> socket
 </libs>
```
reneme, line 14: I guess it is fair to assume that Boost ASIO will (usually) need the same external libraries as the `socket` module (where this list is copied from).

src/scripts/ci_build.py
```diff
@@ -271,20 +275,21 @@ def determine_flags(target, target_os, target_cpu, target_cc, cc_bin,
...
 
-            if boost_root and boost_libs and boost_system:
-                flags += ['--with-boost',
-                          '--with-external-includedir', boost_root,
-                          '--with-external-libdir', boost_libs,
-                          '--boost-library-name', boost_system]
+            if target_os == 'mingw':
+                # apparently mingw needs this legacy socket library version for reasons
+                # as per: https://stackoverflow.com/questions/38770895/how-to-fix-undefined-reference-to-getacceptexsockaddrs-boost-asio-in-clion#comment105791579_38771260
+                flags += ['--ldflags=-static -lwsock32']
```
reneme, line 292: Technically, we could also add this to the `boost/info.txt` as a global dependency. But since `wsock32` is a somewhat legacy library I was hoping that we need to add this here more as a workaround than an actual required library. I.e. the windows-build of MinGW doesn't need it to pass.

reneme (commented):
Self-review...

randombit:
@reneme Sounds good. Please split into 3 PRs as you outlined.

reneme:
@randombit This is now based on #3010 and #3011. Note that I integrated the changes in my self-review into the respective commits.

reneme:
Rebased to master to pick up the emscripten fix.

reneme:
Rebased to master to pick up the emscripten fix.

.github/actions/vs-shell/README.md
```diff
@@ -0,0 +1,39 @@
+Visual Studio shell
```
randombit, line 1: I've added `egor-tensin/vs-shell` as an allowed source so I think it would be possible to remove this. I think that's a better option since maintaining a fork of this is very much outside the scope of the library itself.

randombit (approved):
This is fine, but let's try using the upstream version of the action rather then embedding it. I've added the upstream version to the approved list in the repo.

.github/actions/vs-shell/README.md
reneme: I think it makes sense to also specify the desired version of the 3rd party action. Though, apparently, your allow list is too restrictive for that:

> Bad request - egor-tensin/vs-shell@v2 is not allowed to be used in randombit/botan. Actions in this workflow must be: within a repository owned by randombit, created by GitHub, or matching the following: google/oss-fuzz/*, egor-tensin/vs-shell.

I guess you want to allow `egor-tensin/vs-shell*`, `.../vs-shell@*` or even `.../vs-shell@2`...

reneme: @randombit I need your help with a repository config. :)

randombit: Sorry I missed this! I updated the rule to `egor-tensin/vs-shell@*`

reneme: The 3rd party usage seems to work now. But something still fails in the windows builds. I'll look into it tomorrow. Thanks for giving this another look with the configs.

reneme: Heureka!

PR #3010, opened 2022-07-20
Stop linking against libboost_system / start using -isystem compiler option

reneme:
Changes in #3007 that are not related to the CI.


randombit (approved):
Looks good! No problem with bumping the minimum Boost requirement but we (I) should make sure this is noted in the release notes.

PR #3011, opened 2022-07-20
CI: Improvements and preparation for Windows on GitHub Actions

reneme:
Changes in #3007 that prepare the CI setup for the integration of the new Windows build configs.
This contains #3010, i.e. probably #3010 should be reviewed/merged first.

* auto-detect `sccache` in `ci_build.py`
* automatically cancel builds on consecutive pushes to pull requests
* compiler cache directory is dynamically detected in `setup_gh_actions.sh`
* build agent setup is pulled into a custom action
* build configurations are sorted into several job matrices

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,12 @@ on:
   pull_request:
     branches: [ master ]
 
+# cancel running workflows when new commits are being pushed in pull requests
+# but not on the master branch
+concurrency:
+  group: ${{ github.workflow }} @ ${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   ci:
     runs-on: ${{ matrix.host_os }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,74 +21,23 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  ci:
-    runs-on: ${{ matrix.host_os }}
-
-    env:
-      ANDROID_NDK: android-ndk-r21d
-
+  linux:
+    name: "Linux (Basic)"
     strategy:
       fail-fast: false
 
       matrix:
-        include:
-          - target: coverage
-            compiler: gcc
-            host_os: ubuntu-20.04
-          - target: shared
-            compiler: gcc
-            host_os: ubuntu-20.04
-          - target: shared
-            compiler: clang
-            host_os: ubuntu-20.04
-          - target: valgrind
-            compiler: gcc
-            host_os: ubuntu-20.04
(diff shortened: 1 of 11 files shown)


src/scripts/ci_build.py
```diff
@@ -482,10 +481,16 @@ def main(args=None):
             print('Error unknown compiler %s' % (options.cc))
             return 1
 
-    if options.compiler_cache is None and options.cc != 'msvc':
-        # Autodetect ccache
-        if have_prog('ccache'):
+    if options.compiler_cache is None:
+        # Autodetect compiler cache
+        if have_prog('sccache'):
+            print('Found \'sccache\' installed, will use it...')
```
randombit, line 487: Instead of these two `print`, have one `print` at the end that prints `options.compiler_ccache` if it is not `None`

randombit: Also you can use "" for the outer quotes and then don't have to quote the `'`

.github/workflows/ci.yml
```diff
@@ -14,85 +14,172 @@ on:
...
             compiler: gcc
-            host_os: ubuntu-20.04
+            host_os: ubuntu-latest
+            make_tool: make
           - target: lint
             compiler: gcc
-            host_os: ubuntu-20.04
-          - target: shared
+            host_os: ubuntu-latest
+            make_tool: make
```
randombit, line 125: I dislike having to explicitly mention that (of course) we use `make` for these targets. But I know this is in preparation for supporting `nmake`/`jom` for Windows. It would be nice if we could avoid it. But no support for conditionals... YAML is not my favorite programming language.

randombit (approved):
Couple minor comments but looks good.

The repeated `make_tool: make` isn't a must fix, I don't like it so if you see an easy way to resolve it :+1: but don't worry about it too much.

.github/workflows/ci.yml
reneme: Same here! I didn’t find a good way to default a variable to some value without repeating it.
In fact that was a big motivation to add multiple job matrices on the first place. Boiling it down to the `make_tool` here was the best compromise I could come up with.

I wouldn’t want to rule out that I missed something, though. Seems like a typical pain point many yaml-programmers must have…

reneme: Here's a suggestion: Exploit that undefined variables in yaml default to the empty string and teach `ci_build.py` to deal with `--make-tool=''` by explicitly defaulting to `make`. I find it a bit arbitrary to be honest, but it cleans up the yaml for sure.

randombit: Bit of a hack but still an improvement overall, :+1:

PR #3012, opened 2022-07-21
CI: Run sanitizers on the entire test corpus

reneme:
This adds CI jobs running clang's ASan and UBsan as well as MSVC's ASan and checked iterators on the entire test corpus (most notably BoGo tests). This revealed a memory leak in the `bogo_shim`.

Note that parallel building (via `jom`) with debug symbols did not work on windows. This would require multiple processes to access `botan.pdb` concurrently which fails. The suggested fix (adding `/FS`) didn't work either -- likely due to lacking support of file-system locking in GitHub Action's VMs. Therefore, the windows sanitizer build uses `nmake` (and therefore builds sequentially).

This also uses `BOTAN_FFI_RETURNING` more consistently in `ffi_cert.cpp` to silence warnings showing up in the MSVC sanitizer build. See [here for details on this macro](https://github.com/randombit/botan/blob/master/src/lib/ffi/ffi_util.h#L96-L107).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,7 +155,14 @@ jobs:
           - target: sanitizer
             compiler: msvc
             host_os: windows-latest
-            make_tool: jom
+            make_tool: nmake # to force a serial build (otherwise multi-process PDB handling chokes)
+                             # See: https://github.com/randombit/botan/pull/3012#discussion_r1033621569
+          - target: sanitizer
+            compiler: clang
+            host_os: ubuntu-22.04
+          - target: sanitizer
+            compiler: gcc
+            host_os: ubuntu-22.04
           - target: lint
             compiler: gcc
             host_os: ubuntu-22.04
@@ -177,7 +184,7 @@ jobs:
           repository: reneme/boringssl
           ref: rene/runner-20220322 # TODO: merge changes to Botan's boring fork
           path: ./boringssl
-        if: matrix.target == 'coverage'
+        if: matrix.target == 'coverage' || matrix.target == 'sanitizer'
 
       - name: Setup Build Agent
         uses: ./source/.github/actions/setup-build-agent
--- a/src/scripts/ci_build.py
+++ b/src/scripts/ci_build.py
@@ -75,6 +75,8 @@ def build_targets(target, target_os):
 
     if target in ['coverage']:
         yield 'bogo_shim'
+    if target in ['sanitizer'] and target_os not in ['windows']:
+        yield 'bogo_shim'
 
 def determine_flags(target, target_os, target_cpu, target_cc, cc_bin, ccache,
                     root_dir, build_dir, test_results_dir, pkcs11_lib, use_gdb,
@@ -168,7 +170,10 @@ def determine_flags(target, target_os, target_cpu, target_cc, cc_bin, ccache,
         flags += ['--ldflags=-static']
 
     if target == 'coverage':
-        flags += ['--with-coverage-info', '--with-debug-info', '--test-mode', '--terminate-on-asserts']
+        flags += ['--with-coverage-info']
+
+    if target in ['coverage', 'sanitizer']:
+        flags += ['--with-debug-info', '--test-mode', '--terminate-on-asserts']
 
     if target == 'valgrind':
         flags += ['--with-valgrind']
@@ -203,7 +208,7 @@ def determine_flags(target, target_os, target_cpu, target_cc, cc_bin, ccache,
         if target_cc in ['clang', 'gcc']:
             flags += ['--enable-sanitizers=address,undefined']
         else:
-            flags += ['--with-sanitizers']
+            flags += ['--enable-sanitizers=address']
 
     if target in ['valgrind', 'sanitizer', 'fuzzers']:
         flags += ['--disable-modules=locking_allocator']
@@ -301,7 +306,7 @@ def determine_flags(target, target_os, target_cpu, target_cc, cc_bin, ccache,
         if target_os in ['osx', 'ios']:
             flags += ['--with-commoncrypto']
 
-        if target in ['coverage', 'shared', 'static', 'amalgamation']:
+        if target in ['coverage', 'sanitizer', 'shared', 'static', 'amalgamation']:
             flags += ['--with-boost']
             if target_cc == 'clang':
                 # make sure clang ignores warnings in boost headers
@@ -591,7 +596,7 @@ def main(args=None):
             if target in ['coverage', 'fuzzers']:
                 make_targets += ['fuzzer_corpus_zip', 'fuzzers']
 
-            if target in ['coverage']:
+            if target in ['coverage', 'sanitizer'] and options.os not in ['windows']:
                 make_targets += ['bogo_shim']
 
             cmds.append(make_prefix + make_cmd + make_targets)
@@ -602,7 +607,7 @@ def main(args=None):
         if run_test_command is not None:
             cmds.append(run_test_command)
 
-        if target == 'coverage':
+        if target in ['coverage', 'sanitizer'] and options.os != 'windows':
             if not options.boringssl_dir:
                 raise Exception('coverage build needs --boringssl-dir')
 
@@ -619,11 +624,11 @@ def main(args=None):
                          os.path.join(build_dir, 'fuzzer_corpus'),
                          os.path.join(build_dir, 'build/fuzzer')])
 
-        if target in ['shared', 'coverage'] and options.os != 'windows':
+        if target in ['shared', 'coverage', 'sanitizer'] and options.os != 'windows':
             botan_exe = os.path.join(build_dir, 'botan-cli.exe' if options.os == 'windows' else 'botan')
 
             args = ['--threads=%d' % (options.build_jobs)]
-            if target == 'coverage':
+            if target in ['coverage', 'sanitizer']:
                 args.append('--run-slow-tests')
             if root_dir != '.':
                 args.append('--test-data-dir=%s' % root_dir)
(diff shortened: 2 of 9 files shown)


src/scripts/ci_build.py
```diff
@@ -152,7 +154,12 @@ def determine_flags(target, target_os, target_cpu, target_cc, cc_bin,
         flags += ['--ldflags=-static']
 
     if target == 'coverage':
-        flags += ['--with-coverage-info', '--with-debug-info', '--test-mode', '--terminate-on-asserts']
+        flags += ['--with-coverage-info']
+
+    if target in ['coverage', 'sanitizer']:
+        if target_os != 'windows':
+            flags += ['--with-debug-info']
```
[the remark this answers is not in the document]
lieser: > Shall we try it?


I would say yes

lieser:
For me it is currently a little unclear what the distinguishing between the `fuzzers` and `sanitizer` job is supposed to be.

After this PR we would have:

1. fuzzers, gcc
2. sanitizer, clang
3. sanitizer, msvc

Some observations/questions:

- All three are building the test fuzzer. But only the fuzzers job is actually executing them.
- The fuzzers job is also executing at least part of the normal tests, but not e.g. the bogo tests. Is this intended and needed?
- The sanitizer jobs are also running the bogo and cli tests
- The fuzzer job is also using sanitizers. Is the sanitizer config different from the sanitizer job? Or why do we need/want to distinguish between them?

Seems to be currently like we probably still want to have three such jobs (one for each compiler), but let them run the same tests and name them the same (I would tend to name them sanitizer).

randombit:
The intent of the fuzzers build is just to build and run the fuzzers to ensure they work. The reason for using the sanitizers there is that this is typically how the sanitizers are built (eg in OSS-Fuzz) and will detect any memory errors caused by the corpus itself.

>  All three are building the test fuzzer.

I don't think that's correct. If it is correct, it was certainly unintentional!

```
    if target in ['fuzzers', 'coverage']:
        flags += ['--build-fuzzers=test']
```

> The fuzzers job is also executing at least part of the normal tests, but not e.g. the bogo tests. Is this intended and needed?

It's certainly intentional that we skip the BoGo tests in the fuzzer job since that doesn't relate to the fuzzers. It actually would be better if the fuzzers job skipped the normal tests as well, is it is redundant with the other builds.

> Seems to be currently like we probably still want to have three such jobs (one for each compiler), but let them run the same tests and name them the same (I would tend to name them sanitizer).

It would be fine to have an equivalent to the Clang sanitizer build for GCC. That might be helpful as occasionally GCC's ASan behaves a little differently from Clang's. ISTR a couple of bugs that were caught by one but not the other. But there is no need to build and run the fuzzers under all three compilers IMO.

Also the fuzzers job enables `--unsafe-fuzzer-mode` which is not in general how we'd want to test the library, since it disables various signature verifications and so on.

src/scripts/ci_build.py
```diff
@@ -178,7 +185,7 @@ def determine_flags(target, target_os, target_cpu, target_cc, cc_bin,
     if target == 'fuzzers':
         flags += ['--unsafe-fuzzer-mode', '--terminate-on-asserts']
 
-    if target in ['fuzzers', 'coverage']:
+    if target in ['fuzzers', 'coverage', 'sanitizer']:
         flags += ['--build-fuzzers=test']
```
lieser, line 189: With the clarifications of the differences between the fuzzers and sanitizer jobs is does not seem like we want to build the fuzzer tests also for the sanitizer jobs.

reneme: That makes sense to me, too. Changed.

reneme:
I removed the `--build-fuzzers=test` flag from `sanitizer` builds and added a GCC sanitizer job. Maybe it would be helpful when the CI pipeline would generate some sort of input/output matrix and publish that as a build result. I'd hope that this would provide a quick overview over which job is building and running what. Something along the lines of:

| CI Flag | sanitizer/gcc | sanitizer/clang | fuzzers/gcc | coverage | ... |
| --- | --- | --- | --- | --- | --- |
| `--build-fuzzers=test` | | | :heavy_check_mark: | :heavy_check_mark: | |
| `--enable-sanitizers=address,undefined` | :heavy_check_mark: | :heavy_check_mark: | | | |
| `--with-debug-asserts` | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | | |
| ... | | | | | |

src/scripts/ci_build.py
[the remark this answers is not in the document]
reneme: [Seems to work just fine](https://github.com/randombit/botan/actions/runs/3548877661/jobs/5960581074) without this. Some macOS-based configurations failed as there seems to be some outage at homebrew.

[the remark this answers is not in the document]
lieser: You commented out the code, but I think it should be removed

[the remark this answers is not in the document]
lieser: Also you now removed `--with-debug-info` now completely. But for nice outputs we probably want debug info in our sanitizer builds.
So I think it should be added for all, not removed for all.

lieser (changes requested):
I think the `--with-debug-info` discussion is still open.

src/scripts/ci_build.py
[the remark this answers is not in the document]
reneme: [Another bread-crumb](https://stackoverflow.com/a/21855956): Someone had this error while building inside a Dropbox share, because the dropbox daemon locked files at will and interfered with the build. Maybe some anti-virus software is meddling with the file.

lieser (approved):

reneme:
Note: This cannot be merged, yet. MSVC complained about more locations where `BOTAN_FFI_DO` should be replaced by `BOTAN_FFI_RETURNING`. Though, now that we have C++17 we can do better than this workaround and let the compiler figure out what to do: https://github.com/randombit/botan/pull/3079

Let's merge #3079 first and rebase this to get the CI to pass.

reneme:
Rebased after #3079 was merged. Please be nice, MSVC!

src/scripts/ci_build.py
```diff
@@ -201,7 +206,7 @@ def sanitize_kv(some_string):
         if target_cc in ['clang', 'gcc']:
             flags += ['--enable-sanitizers=address,undefined']
         else:
-            flags += ['--with-sanitizers']
+            flags += ['--enable-sanitizers=address']
```
reneme, line 209: This explicitly disables the `-D_ITERATOR_DEBUG_LEVEL=1` as it doesn't seem to be compatible with ASan. Linking of the botan library fails with:

```
VCASAN.lib(vcasan.obj) : error LNK2038: mismatch detected for '_ITERATOR_DEBUG_LEVEL': value '0' doesn't match value '1' in asn1_alg_id.obj
   Creating library .\botan.lib and object .\botan.exp
.\botan.dll : fatal error LNK1319: 1 mismatches detected
```

lieser (approved):

.github/workflows/ci.yml
```diff
@@ -155,7 +155,16 @@ jobs:
           - target: sanitizer
             compiler: msvc
             host_os: windows-latest
-            make_tool: jom
+            make_tool: nmake # to force a serial build (otherwise multi-process PDB handling chokes)
+                             # See: https://github.com/randombit/botan/pull/3012#discussion_r1033621569
+          - target: sanitizer
+            compiler: clang
+            # in Oct. 2022 'ubuntu-latest' would mean 'ubuntu-20.04' only
```
randombit, line 162: These comments can be removed since we are on `ubuntu-22.04` for all CI now

randombit (approved):
Changes look good but there is still a merge conflict and some CI failures. Also please rebase+squash the commits before merge.

reneme:
Rebased, squased and now waiting for CI. I will force push again tomorrow. Don't have my YubiKey for commit signing handy. >.<

src/scripts/ci_build.py
```diff
@@ -326,6 +331,10 @@ def sanitize_kv(some_string):
                 test_cmd += ['--pkcs11-lib=%s' % (pkcs11_lib)]
 
     if target in ['coverage', 'sanitizer']:
+        if target_os != 'windows' or target != 'sanitizer':
+            # GitHub Actions worker intermittently ran out of memory when
+            # asked to allocate multi-gigabyte buffers under MSVC's ASan.
+            test_cmd += ['--run-memory-intensive-tests']
```
randombit, line 337: Let's invert this option (to `--skip-memory-intensive-tests`) because otherwise nobody (including eg an end user building on some weird CPU/compiler we've never seen) will have these tests run when they build locally, unless they know about this magic flag. That doesn't matter so much for this one particular RNG test but would be if it started being used to gate for instance an Argon2 test with very large memory.

randombit (approved):
Approved modulo the option comment

src/scripts/ci_build.py
reneme: I considered that for the same reason but went with the 'disabled by default' to stay consistent with `--run-long-tests`. We could turn both around? Or just don't care about consistency here. I don't have a strong feeling either way.

randombit: Eh. Not really strong feelings here I guess.

Let's make one minor change here: add `--run-memory-intensive-tests` to the default CI run and then *remove* it for the Windows sanitizer build only. That way we are still running the RNG 4GB test on macOS and Windows. Otherwise, AIUI, we are only ever running it within the Linux coverage build, so we could miss some problem that creeps into how we implement `System_RNG` on another platform.

reneme: Done. For the record: The same is true for the "long" tests. `--run-long-tests` is only ever executed for the coverage build (and -- with this PR -- with sanitizers on all three platforms)

randombit: Thanks for pointing that out. I'll address this.

PR #3020, opened 2022-08-01
CI: Export test results as xUnit files

reneme:
Based on https://github.com/randombit/botan/pull/3007

Eventually, this should enable rich visualization of failed tests in GitHub Action's UI. Also, it will facilitate the automatic generation of test report documents.


reneme:
Closing in favour of #3028.

PR #3021, opened 2022-08-01
Fix XMSS example

lieser:
- add missing `BOTAN_PUBLIC_API` in public XMSS headers
- use `PK_Signer`/`PK_Verifier` in XMSS example
  - previously used `PK_Ops::Signature`/`PK_Ops::Verification` is no longer part of the public API in version 3.0.0


randombit (approved):
Thanks

PR #3027, opened 2022-08-04
Refactor: Log Test results via a dedicated reporter class

reneme:
This separates the concerns of "test running" from "test result reporting". I introduced an abstract base class `Botan_Tests::Reporter`. The existing stdout log reporting moved to `StdoutReporter` and care was taken that the output is compatible (see below for differences).

Also, I moved the test runner classes into a sub directory and refactored the `Test_Runner::run_tests()` method to better separate multi-threaded test execution.

Note that care was taken to make each commit self-contained to ease reviewability.

This is in preparation for a new test result reporter emitting JUnit XML reports.

### Differences in the Log Output

The only difference is in the preamble. I introduced the concept of "properties" to manage things like the DRBG-seed and build configurations. Hence, the reporting of those are slightly different now. As far as I'm aware, there are no scripts parsing this output anyway, though.

One "dynamic" change: Previously, the program logged the just-started test-suite name and later the results. Below, `block:\n` would have been logged before the (potentially lengthy) tests started. Now, both the label `block:` and the associated test results are logged after the tests have finished. That might be a minor inconveniece when actively looking at the running tests. @randombit Worth "fixing"?

#### Before

```
Testing Botan 3.0.0-alpha0 (unreleased, revision git:52810a2f6b1ef2f0fd6a3597060ba8dd7ced67a0, distribution unspecified)
CPU flags: sse2 ssse3 sse41 sse42 avx2 avx512f avx512dq avx512bw avx512_icelake rdtsc bmi1 bmi2 adx aes_ni clmul rdrand rdseed intel_sha avx512_aes avx512_clmul
Starting tests drbg_seed:17082A4528C48D4C
block:
AES-128 ran 17235 tests in 16.49 msec all ok
AES-192 ran 20250 tests in 19.07 msec all ok
AES-256 ran 23130 tests in 22.38 msec all ok
ARIA-128 ran 30 tests in 0.03 msec all ok
ARIA-192 ran 30 tests in 0.04 msec all ok
ARIA-256 ran 30 tests in 0.03 msec all ok
[...]
Twofish ran 16545 tests in 17.38 msec all ok
Tests complete ran 225585 tests in 222.21 msec all tests ok
```

#### After

```
Testing Botan 3.0.0-alpha0 (unreleased, revision git:4008aa072b5cf3267c7181f5487ddf489b0b9dec, distribution unspecified)
Properties:
  CPU flags: sse2 ssse3 sse41 sse42 avx2 avx512f avx512dq avx512bw avx512_icelake rdtsc bmi1 bmi2 adx aes_ni clmul rdrand rdseed intel_sha avx512_aes avx512_clmul
  drbg_seed: 17082A0EBAF5AF2B
block:
AES-128 ran 17235 tests in 17.24 msec all ok
AES-192 ran 20250 tests in 21.49 msec all ok
AES-256 ran 23130 tests in 24.17 msec all ok
ARIA-128 ran 30 tests in 0.03 msec all ok
ARIA-192 ran 30 tests in 0.03 msec all ok
ARIA-256 ran 30 tests in 0.03 msec all ok
[...]
Twofish ran 16545 tests in 17.63 msec all ok
Tests complete ran 225585 tests in 234.05 msec all tests ok
```


src/tests/main.cpp
```diff
@@ -98,7 +98,7 @@ int main(int argc, char* argv[])
 
       Botan_Tests::Test_Runner tests(std::cout);
 
-      return tests.run(opts);
+      return tests.run(opts) ? 0 : 1;
```
reneme, line 101: Previously that returned the number of failed tests as the program's exit code. Keeping that behaviour would mean extra logic in the new reporter. I hope that doesn't matter too much.

src/tests/runner/test_reporter.cpp
```diff
@@ -0,0 +1,62 @@
...
+      const std::string who = result.who();
+      auto i = combined.find(who);
+      if(i == combined.end())
+         {
+         combined.insert(std::make_pair(who, Botan_Tests::Test::Result(who)));
+         i = combined.find(who);
+         }
+
+      i->second.merge(result);
+      }
```
reneme, line 48: I copied that verbatim from the old `Test_Result_State::record()` method without fully grasping what it's supposed to achieve. @randombit any idea?

src/tests/tests.cpp
```diff
@@ -429,6 +429,7 @@ Test::Result::Result(std::string who, std::vector<Result> downstream_results)
       merge(result, true /* ignore non-matching test names */);
    }
 
+// TODO: this should move to `StdoutReporter`
```
reneme, line 432: This method purely implements "rendering" for the stdout logging. Hence, it might make sense to move it over to `StdoutReporter`. There's a test directly accessing this though, and I wanted to keep the change "small" for now.

src/tests/main.cpp
randombit: No it was just easy to do but I don't think anything cares except about pass/fail.

src/tests/runner/test_reporter.cpp
randombit: IIRC this is to combine multiple tests that have the same summary within a single test. Eg if two `Test::Result("foo")` are created this has the effect of combining them into a single report in the final output.

randombit (approved):
One comment where I think the logic is wrong, otherwise lgtm

src/tests/runner/test_runner.cpp
```diff
@@ -185,116 +186,30 @@ int Test_Runner::run(const Test_Options& opts)
...
-      }
-
-   for(auto const& result : combined)
-      {
-      out << result.second.result_string();
-      m_tests_run += result.second.tests_run();
-
-      const size_t failed = result.second.tests_failed();
+      const bool passed =
+         (opts.test_threads() > 1)
```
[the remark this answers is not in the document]
reneme: Good point! I didn't realize that `==0` means: "pick something". Though, I implemented it a bit differently now: https://github.com/randombit/botan/pull/3027/commits/2bcd84ba60724776c07a16eec94d5529e57d07c4 as:

```C++
      const bool passed =
         (opts.test_threads() == 1)
            ? run_tests(req)
            : run_tests_multithreaded(req, opts.test_threads());
```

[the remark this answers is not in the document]
randombit: Yeah that's fine

PR #3031, opened 2022-08-05
Remove unused get_emsa() declaration

### LINKED ISSUE #3022: get_emsa() is declared but not defined
I see a declaration of `BOTAN_TEST_API EMSA* get_emsa(const std::string& algo_spec)` in `src/lib/pk_pad/emsa.h`, but don't see any definition of it.

This is caused by the raw pointer removal #2697.

Either:
1. The declaration should be removed
2. Similar to other removals, a small definition using the new `EMSA::create()` should be added again.
    - Maybe declare it as deprecated. But in most similar cases (e.g. `BlockCipher::clone()`) the old API using raw pointer is not yet marked deprecated.

lieser:
Fixes #3022

### DIFF
--- a/src/lib/pk_pad/emsa.h
+++ b/src/lib/pk_pad/emsa.h
@@ -106,14 +106,6 @@ class BOTAN_TEST_API EMSA
    };
 
 /**
-* Factory method for EMSA (message-encoding methods for signatures
-* with appendix) objects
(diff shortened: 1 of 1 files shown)


reneme (approved):
One test is failing, though that seems to be a flaky test that is unrelated to this change. I've seen it spuriously fail in other builds as well.

PR #3032, opened 2022-08-08
Feature: <module_info> in info.txt files

reneme:
This is to improve on [my previous draft PR](https://github.com/randombit/botan/pull/3003) and split things up a little.

### `<module_info>`

I added `<module_info>` tags to all `info.txt` module descriptions along with the necessary parsing code in `configure.py`.

`<module_info>` contains three variables:

* `name` -- a human-friendly name for the module
* `brief` -- a short description of the module's content (optional)
* `type` -- the type of the module (`Public`, `Internal` or `Virtual`) (defaults to `Public`)

Note that new modules are rejected if they do not contain a `<module_info>` tag that specify the `name` variable at least.

#### `<module_info> \n type -> "..."`

This introduces module types (inspired by [a previous discussion in the initial Kyber PR](https://github.com/randombit/botan/pull/2872#discussion_r782044034)), to define a module's visibility and library user interaction.

`Public` modules behave like the current modules. They usually contain some implementation files, may or may not contain sub-modules, can be depended upon or depend on other modules and can be enables/disables at build time by the library user.

`Internal` modules behave essentially like `Public` modules, but the library user cannot actively enable/disable them. They are supposed to contain code that shouldn't be interacted with directly by the library user. Frankly, I don't know of any such module apart from the recently introduced `kyber_common`.

`Virtual` modules essentially tackle the first limitation stated [in the previous PR](https://github.com/randombit/botan/pull/3003). They can be seen as "Internal" but must not contain implementation files and cannot be depended upon. Instead they are meant as container modules to other (`Public`) modules. Examples are: `prov`, `math`, `misc`, `passhash` each containing sub-modules. We require them as a location to place the `<module_info>` for those "container directories".

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+# pylint: disable=too-many-lines
+
 """
 Configuration program for botan
 
--- a/configure.py
+++ b/configure.py
@@ -814,7 +814,7 @@ class ModuleInfo(InfoObject):
             infofile,
             ['header:internal', 'header:public', 'header:external', 'requires',
              'os_features', 'arch', 'isa', 'cc', 'comment', 'warning'],
-            ['defines', 'libs', 'frameworks'],
+            ['defines', 'libs', 'frameworks', 'module_info'],
             {
                 'load_on': 'auto',
                 'endian': 'any',
@@ -872,6 +872,7 @@ class ModuleInfo(InfoObject):
         self.requires = lex.requires
         self.warning = combine_lines(lex.warning)
         self.endian = lex.endian
+        self._parse_module_info(lex)
 
         # Modify members
         self.source = [normalize_source_path(os.path.join(self.lives_in, s)) for s in self.source]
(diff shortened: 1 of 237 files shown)


configure.py
```diff
@@ -791,10 +793,13 @@ def __init__(self, infofile):
...
-                self.parent_module = os.path.basename(obj_dir)
-            else:
-                self.parent_module = None
+            self.parent_module = None
+
+            while obj_dir:
+                if os.access(os.path.join(obj_dir, 'info.txt'), os.R_OK):
+                    self.parent_module = os.path.basename(obj_dir)
+                    break
+                (obj_dir, _) = os.path.split(obj_dir)
```
reneme, line 802: This fixes an issue where parent modules could not be found when modules where nested into intermediate directories that did not contain an `info.txt`. One might argue that this is not strictly required anymore, given that such directories are likely going to be `Virtual` modules now. Nevertheless, this isn't enforced so probably it does not hurt to keep this fix.

reneme:
I wanted a way to group the `math` modules together in the documentation. Otherwise they would have showed up top-level in the Doxygen module list. Now there's a "virtual" module with some documentation `@brief` combining them.

I.e. the "virtual" module idea is purely for documentation purposes.

Here's a screenshot from the former draft PR where "base**" and "hex" show up toplevel. In the new PR they are grouped under the "virtual" module "Codecs".

## without "virtual" modules

178282978-79b1d246-269c-4608-93cd-e66b5776e817

## with "virtual" modules

183438799-bbc7c49b-2bb4-499b-ada4-e031bff9e62a

randombit (approved):

PR #3033, opened 2022-08-08
Doc: Render <module_info> into Doxygen

reneme:
This is based on [the `<module_info>` tags added before](https://github.com/randombit/botan/pull/3032) and renders that information into Doxygen. The results are similar to [last month's previous draft PR](https://github.com/randombit/botan/pull/3003) but using the `Virtual` modules, the tree view of modules now makes a lot more sense.

# Main "Modules" list

Screenshot 2022-08-08 161432

# Module Descriptions

Modules have a "Details" page that list all their sub-modules, dependencies and also marks `Virtual` and `Internal` modules as such.

Screenshot 2022-08-08 161633

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -211,6 +211,7 @@ class BuildPaths(object): # pylint: disable=too-many-instance-attributes
         self.doc_output_dir = os.path.join(self.build_dir, 'docs')
         self.handbook_output_dir = os.path.join(self.doc_output_dir, 'handbook')
         self.doc_output_dir_doxygen = os.path.join(self.doc_output_dir, 'doxygen') if options.with_doxygen else None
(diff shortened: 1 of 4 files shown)


randombit:
With #3032 merged I think this needs to be rebased

reneme:
> With https://github.com/randombit/botan/pull/3032 merged I think this needs to be rebased

Done! I don't have my yubikey in reach, will sign and re-push once I'm home.

lieser (approved):
LGTM, but then I tried it locally I noticed that in `src/lib/kdf/hkdf/info.txt` it incorrectly has HMAC as the name. Should get fixed here or in a follow up PR.

lieser (approved):

Issue #3008, opened 2022-07-19
[3.0.0] Support unpadding for blinded rsa decryption using PKCS#11


weberph2:
To ensure that blinding is used for PKCS#11 rsa decryption, it is necessary to set the `padding` of `Botan::PK_Decryptor_EME` to `"Raw"`.
This prevents sensitive information from being passed through the reader, the usb connection and the middleware.
If the result needs unpadding (e.g. OAEP) then `Botan::EME::unpad` can be used as long as the header / EME base class are public.

Would it be an option to keep exposing the EME headers or base class as part of the public API or are there any alternatives?

randombit:
I thought the PKCS11 RSA wrapper already implemented this blinding? So I wouldn't think there would be any need for such operations at the application level as well. Or am I misunderstanding?

weberph2:
Yes, the PKCS11 RSA wrapper performs blinding, but only for "Raw" operations: https://github.com/randombit/botan/blob/master/src/lib/prov/pkcs11/p11_rsa.cpp#L149

Blinding is not possible if the smart card or middleware is asked to perform the unpadding.

As long as Botan does not provide a way to explicitly perform "unpadding with (blinded) raw decryption", an application must use "Raw" decryption, then perform unpadding at the application level (no blinding at app level).

PR #3034, opened 2022-08-09
PKCS#11 RSA allows choosing software EME

reneme:
When enabling ` set_use_software_padding(bool)` the PKCS#11 module won't perform the unpadding. Instead, we let it decrypt a blinded ciphertext in Raw-mode and strip the EME in software afterwards.

Hence,  the PKCS#11 module (and all middleware on the way) won't gain knowledge of the decrypted secret value. Note that this might require some refactoring.

@weberph2, does that roughly address your issue? (See #3008)


weberph2:
@reneme Yes, thank you. This would resolve the issue described in #3008.

neusdan:
Is this pull request still WIP or ready for review? :)

randombit:
:+1:

PR #3035, opened 2022-08-10
Berlekamp-Welch error correction for zfec module

thepaul:
The zfec module provides Reed-Solomon encoding and decoding bytewise compatible with the FECpp library and the zfec library from Tahoe-LAFS. Interestingly, although Reed-Solomon is indeed a forward error-correcting code and each of these implementations have "FEC" in the name, none of them actually perform the error correction part.

This patch adds a new API providing error correction for Reed-Solomon encoded data, using the Berlekamp-Welch algorithm. I ported it from the Go-language "infectious" library from Vivint, at https://github.com/vivint/infectious. That library is MIT-licensed, so that might be a problem if you don't want to add the extra license information and permission notice to Botan.

### DIFF
--- a/news.rst
+++ b/news.rst
@@ -143,6 +143,8 @@ Version 3.0.0, Not Yet Released
 
 * Generate a ``compile_commands.json`` for use with Clang tooling.
 
+* Add error-correction support to the FEC implementation.
+
 Version 2.19.2, 2022-06-03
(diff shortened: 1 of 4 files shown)


randombit:
This looks useful but unfortunately the license is an issue. In retrospect I should have used MIT instead of BSD, as MIT is of course more common but there is no fixing that now, and I don't want to complicate the licensing situation.

PR #3045, opened 2022-09-14
FIX: emscripten object file suffix is .o

reneme:
This is an attempt to make the emscripten build pass again.


randombit (approved):
I wonder what version of Emscripten broke this, but I suppose working with the latest version is the best we can do here.

PR #3046, opened 2022-09-14
Sunset Python 2 support

reneme:
I guess its time to say goodbye to Python 2. 

This updates the documentation accordingly, uses `/usr/bin/env python3` in all python scripts I found in the repo and removes some python2/3 specific switches in the CI.

Note that on the current master a simple `./configure.py` doesn't work anymore for modern distributions (and macOS for that matter). They don't ship with a `python` executable in the $PATH anymore. Hence, one needs to work-around with some `ln -s /usr/bin/python3 /usr/local/bin/python` hackery.

Also, some minor script adjustments were necessary. I smoke-tested all scripts that I updated, though cannot rule out that things in some old dev-scripts fall apart.

This is very much work in progress to see CI results.


reneme:
AppVeyor doesn't seem to find a `python3` binary. I guess we can ignore that and later see how [GitHub Actions](https://github.com/randombit/botan/pull/3007) fares with that.

Also the emscripten target fails, [that's handled independently](https://github.com/randombit/botan/pull/3045).

randombit:
At least `configure.py`, `test_cli.py, and `botan2.py` have a number of conditional jumps based on `sys.version_info` to handle Python2 vs Python3 differences, which should be removed.

reneme:
> At least configure.py, test_cli.py, and botan2.py have a number of conditional jumps based on sys.version_info to handle Python2 vs Python3 differences, which should be removed.

Deleting code! ❤️

src/scripts/test_cli.py
```diff
@@ -975,10 +969,7 @@ def cli_tls_http_server_tests(tmp_dir):
     if resp.status != 405:
         logging.error('Unexpected response status %d', resp.status)
 
-    if sys.version_info.major >= 3:
-        rc = tls_server.wait(5) # pylint: disable=too-many-function-args
-    else:
-        rc = tls_server.wait()
+    rc = tls_server.wait(5) # pylint: disable=too-many-function-args
```
randombit, line 972: I think this pylint annotation can be removed now, it was getting confused between 2.x and 3.x APIs

src/scripts/test_cli.py
```diff
@@ -1071,10 +1062,7 @@ def do_GET(self): # pylint: disable=invalid-name
         if body != server_response:
             logging.error('Unexpected response from server %s', body)
 
-    if sys.version_info.major >= 3:
-        rc = tls_proxy.wait(5) # pylint: disable=too-many-function-args
-    else:
-        rc = tls_proxy.wait()
+    rc = tls_proxy.wait(5) # pylint: disable=too-many-function-args
```
randombit, line 1065: Similarly, can be removed (I think)

randombit (approved):
If we're going to do it, now's the time.

src/scripts/test_cli.py
reneme: LGTM (and to the linter): https://github.com/randombit/botan/actions/runs/3068176208/jobs/4955320144

reneme:
This works, now. Though it breaks the build on AppVeyor. `master` won't build once this is merged. ..

Instead, lets hold on until [the switch to GitHub Actions (Windows)](https://github.com/randombit/botan/pull/3007#discussion_r970662405) is done and merged, then rebase and (probably) make it green on Windows.

randombit:
> Note that on the current master a simple ./configure.py doesn't work anymore for modern distributions (and macOS for that matter). They don't ship with a python executable in the $PATH anymore.

Good to know. I use Arch which still ships with a symlink by default (pointing to `python3`). I guess in CI we didn't notice this because of explicit `python2` / `python3` invocations.

PR #3047, opened 2022-09-15
Nit: remove std::optional<> where it is not needed

reneme:
The references point into a vector containing X509_Certificate objects. No need for an optional that will always be set anyway. This also eliminates a few (presumably inadvertant) copies by using `const&`.

This was raising some alarms for me, as the code seemed to access an optional without safe-guarding it with `.has_value()`. Turns out the optional is always set and hence shouldn't exist in the first place.


randombit (approved):
:+1:

PR #3049, opened 2022-09-25
socket code consolidation

KerrickC:
new file with consolidated code at path: \botan\src\lib\utils\socket\sockets.h

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -1,0 +1,21 @@
+{
+    "configurations": [
+        {
+            "name": "Win32",
+            "includePath": [
+                "${workspaceFolder}/**"
+            ],
+            "defines": [
+                "_DEBUG",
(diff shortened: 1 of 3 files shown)


lieser:
Thanks for the contribution. Didn't do a review but noticed that the new code is not used anywhere. For this to be merged we probably want to change the existing socket code in the same PR to see how the new consolidated code integrates.

Did you maybe wanted to open this as a draft?

randombit:
This still has 4 socket implementations (asio+tcp, asio+udp, sockets+tcp, sockets+udp) while the idea of the socket consolidation todo is to combine this 2: asio+(tcp and udp) and sockets+(tcp and udp)

randombit:
Closing this for now, @KerrickC we'd be happy to have this in some form eventually. Feel free to open up a ticket to discuss what we're looking for in this todo.

PR #3058, opened 2022-10-11
Report Code Locations of test failures

reneme:
This allows rudimentary reporting of unittest failure code locations. Like so:

```
tls_rfc8448:
Handshake involving Hello Retry Request (Client side) ran 52 tests in 21.09 msec all ok
Middlebox Compatibility Mode (Client side) ran 7 tests in 19.48 msec 1 FAILED
Failure 1: Close connection client connection was terminated produced unexpected result '0' expected '1' (at src/tests/test_tls_rfc8448.cpp:833)
```

The reported code location points to the respective `BOTAN_REGISTER_TEST` macro. Unfortunately, that's currently the only place where we can extract the code location conveniently (via `__FILE__` and `__LINE__`). With C++20 we'll be able to use [`std::source_location`](https://en.cppreference.com/w/cpp/utility/source_location) to obtain more specific source coordinates without the use of macros.

Still, I think this addition is worthwhile. For instance to build a convenient test result integration with VS code problem matchers. See screenshots below:

image

---

image

### DIFF
--- a/src/tests/runner/test_runner.cpp
+++ b/src/tests/runner/test_runner.cpp
@@ -227,7 +227,7 @@ std::vector<Test::Result> run_a_test(const std::string& test_name)
          std::vector<Test::Result> test_results = test->run();
          for (auto& result : test_results)
             {
-            if(test->registration_location() && !result.code_location())
+            if(!result.code_location() && test->registration_location())
                {
                // If a test result has no specific code location associated to it,
                // we fall back to the test case's registration location.
--- a/src/tests/runner/test_runner.cpp
+++ b/src/tests/runner/test_runner.cpp
@@ -225,6 +225,15 @@ std::vector<Test::Result> run_a_test(const std::string& test_name)
       else if(std::unique_ptr<Test> test = Test::get_test(test_name))
          {
          std::vector<Test::Result> test_results = test->run();
+         for (auto& result : test_results)
+            {
+            if(test->registration_location() && !result.code_location())
+               {
+               // If a test result has no specific code location associated to it,
+               // we fall back to the test case's registration location.
+               result.set_code_location(test->registration_location().value());
+               }
+            }
          results.insert(results.end(), test_results.begin(), test_results.end());
          }
       else
--- a/src/tests/runner/test_runner.cpp
+++ b/src/tests/runner/test_runner.cpp
@@ -18,6 +18,8 @@
    #include <botan/internal/rwlock.h>
 #endif
 
+#include <shared_mutex>
+
 namespace Botan_Tests {
 
 Test_Runner::Test_Runner(std::ostream& out) : m_output(out) {}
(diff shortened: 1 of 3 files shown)


src/tests/runner/test_runner.cpp
```diff
@@ -223,6 +225,15 @@ std::vector<Test::Result> run_a_test(const std::string& test_name)
       else if(std::unique_ptr<Test> test = Test::get_test(test_name))
          {
          std::vector<Test::Result> test_results = test->run();
+         for (auto& result : test_results)
+            {
+            if(test->registration_location() && !result.code_location())
```
lieser, line 230: I would switch the checks as this fits better with the logic we want then reading from left to right.

src/tests/tests.h
```diff
@@ -623,15 +641,20 @@ template<typename Test_Class>
 class TestClassRegistration
    {
    public:
-      TestClassRegistration(const std::string& category, const std::string& name)
+      TestClassRegistration(std::string category, std::string name, CodeLocation registration_location)
```
lieser, line 644: What is the reason behind taking by value if `Test::register_test()` is still taking them by reference? Same in `TestFnRegistration`.

src/tests/tests.cpp
```diff
@@ -32,9 +32,14 @@ namespace Botan_Tests {
...
    {
-   if(!ignore_test_name && who() != other.who())
+   if(!ignore_test_name)
       {
-      throw Test_Error("Merging tests from different sources");
+      if(who() != other.who())
+         throw Test_Error("Merging tests from different sources");
+
+      // When deliberately merging results with different names, the code location is
+      // likely inconsistent and must be discarded.
```
lieser, line 41: If they are inconsistent would it maybe make sense to also unset the maybe already set `m_where`? Currently only the location of `other` is discarded.

src/tests/tests.h
reneme: Nothing particular. I wanted to modernize this but didn't follow through all the way. Didn't remove those changes in the end, though.

src/tests/tests.cpp
reneme: Good point. I changed it around so that `m_where` is now discarded if names don't match and `other.m_where` is taken if names match.

src/tests/tests.cpp
```diff
@@ -32,15 +32,18 @@ namespace Botan_Tests {
...
          throw Test_Error("Merging tests from different sources");
 
       // When deliberately merging results with different names, the code location is
       // likely inconsistent and must be discarded.
-      m_where = other.m_where;
+      m_where.reset();
       }
+   else {
+      m_where = other.m_where;
+   }
```
lieser, line 46: we should stay consistent with braces placement.

reneme: agreed.

lieser (approved):

PR #3061, opened 2022-10-18
VS Code: Basic Integrations and Helpers

reneme:
This adds some convenience functionality for Visual Studio Code. Note that this is somewhat best-effort and might need further adaptions for other people. In my current setup (Windows host + Ubuntu 22.04 in WSL) works reasonably well with this.

### Included

* C++ formatting adapter for AStyle (using chiehyu.vscode-astyle plugin)
* Tasks for:
  * Configure/Building
  * Running Unittests
    * either all tests at once or just the selected test source file
    * problem matcher to visualize test failures in the editor
  * Running BoGo TLS Tests
* Debug configuration for Unittests using GDB
* Code snippets for the copyright headers in Python/C++ files
* IntelliSense finds and uses `compile_commands.json`
* PyLint integration

### Usage

Create a symlink in the project root like so: `ln -s src/editors/vscode .vscode`, then start VS Code and profit. It might be beneficial to also use `ln -s src/editors/editorconfig/.editorconfig .editorconfig`.

#### Unittest Problem Matcher

image

#### Running BoGo Tests

image

### DIFF
--- a/src/editors/vscode/scripts/bogo.py
+++ b/src/editors/vscode/scripts/bogo.py
@@ -4,27 +4,27 @@ import os
 from common import run_cmd, get_concurrency
 
 
-boring_repo = "https://github.com/reneme/boringssl.git"
-boring_branch = "rene/runner-20220322"
+BORING_REPO = "https://github.com/reneme/boringssl.git"
+BORING_BRANCH = "rene/runner-20220322"
 
-boring_path = "build_deps/boringssl"
-bogo_path = os.path.join(boring_path, "ssl", "test", "runner")
+BORING_PATH = "build_deps/boringssl"
+BOGO_PATH = os.path.join(BORING_PATH, "ssl", "test", "runner")
 
-shim_path = "./botan_bogo_shim"
-shim_config = "src/bogo_shim/config.json"
+SHIM_PATH = "./botan_bogo_shim"
+SHIM_CONFIG = "src/bogo_shim/config.json"
 
 
 def main():
-    if not os.path.isdir(boring_path):
+    if not os.path.isdir(BORING_PATH):
         # check out our fork of boring ssl
         run_cmd("git clone --depth 1 --branch %s %s %s" %
-                (boring_branch, boring_path, boring_path))
+                (BORING_BRANCH, BORING_REPO, BORING_PATH))
 
     # make doubly sure we're on the correct branch
-    run_cmd("git -C %s checkout %s" % (boring_path, boring_branch))
+    run_cmd("git -C %s checkout %s" % (BORING_PATH, BORING_BRANCH))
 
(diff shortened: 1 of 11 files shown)


src/editors/vscode/scripts/bogo.py
```diff
@@ -0,0 +1,52 @@
...
+    sys.stdout.flush()
+
+    try:
+        subprocess.run(cmd, shell=shell, check=True, cwd=workingdir)
+    except subprocess.CalledProcessError:
+        raise RuntimeError('External command failed, aborting...')
+
+
+boring_repo = "https://github.com/reneme/boringssl.git"
+boring_branch = "rene/runner-20220322"
```
reneme, line 35: I guess it would be great to centralize that somewhere. The same config is duplicated in the CI setup.

src/editors/vscode/tasks.json
```diff
@@ -0,0 +1,145 @@
...
+            "label": "Build All",
+            "group": "build",
+            "type": "shell",
+            "linux": {
+                "command": "make -j $(nproc)"
+            },
+            "osx": {
+                "command": "make -j $(sysctl -n hw.logicalcpu)"
+            },
+            "dependsOn": "Configure",
```
lieser, line 35: Same for the other tasks depending on `Configure`:
Do we really want to enforce a new configure for each run of the build task?
I guess the build performance is still acceptable because of the use of ccache. But it would still prevent a user from just running a different configure and reuse the other tasks (without having to change the configure task).

src/editors/vscode/scripts/bogo.py
lieser: Agree that it would be good to have a central configuration. In general I would try to avoid doing similar stuff for CI and local development in different scripts. And instead try to write a general one that can be reused.

But maybe not that easy here and a little out of scope for the PR.

reneme: Agree. Let's have a cleanup PR after this one. Also, I'm duplicating some scripts from the Sublime integration (which probably no-one uses or maintains now).

src/editors/vscode/tasks.json
reneme: Fair point. I'll remove those dependencies.

src/editors/vscode/settings.json
```diff
@@ -0,0 +1,18 @@
+{
+    "[cpp]": {
+        "editor.defaultFormatter": "chiehyu.vscode-astyle"
+    },
+    "astyle.astylerc": "${workspaceRoot}/src/configs/astyle.rc",
+    "astyle.cmd_options": [
+        "--suffix=none"
+    ],
+    "files.trimTrailingWhitespace": true
```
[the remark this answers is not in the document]
reneme: Indeed we don't. I forgot to symlink `.editorconfig` into the project root. >.<

Issue #2990, opened 2022-06-09
TLS 1.3 - Current State and TODOs


reneme:
With the [TLS 1.3 client implementation now on the finish line](https://github.com/randombit/botan/pull/2922), I'd like to summarize the steps needed towards a feature-complete TLS 1.3 implementation in Botan.

## Current State (June 2022)

Once merged, Botan will have *basic client-side support* for TLS 1.3 with some limitations and with minimal API changes. See https://github.com/randombit/botan/pull/2988 and https://github.com/randombit/botan/pull/2989 for details regarding changes in `TLS::Callbacks` and `TLS::Policy` configurations and https://github.com/randombit/botan/issues/2714#issuecomment-1057175631 for expected changes for some missing features.

The implementation is covered by unit and integration tests (based on [RFC 8448 test vectors](https://datatracker.ietf.org/doc/html/rfc8448)) and passes the relevant BoringSSL BoGo tests. A fuzzing target for the handshake messages is implemented and run in the CI.

### Current Limitations (a.k.a. TODOs)

* [ ] TLS 1.3 server implementation
* [ ] DTLS 1.3
* [ ] Implementation of the ["signature_algorithms_cert" extension](https://www.rfc-editor.org/rfc/rfc8446.html#section-4.2.3)
* [ ] Support for [early data (0-RTT)](https://www.rfc-editor.org/rfc/rfc8446.html#section-2.3)
* [ ] Support for [early secret exporters](https://www.rfc-editor.org/rfc/rfc8446.html#section-7.5)
* [ ] Support for externally provided PSKs
* [ ] Support ["RawPublicKey" in the Certificate handshake message](https://www.rfc-editor.org/rfc/rfc8446.html#section-4.4.2)
* [ ] [Post-handshake client authentication](https://www.rfc-editor.org/rfc/rfc8446.html#section-4.2.6)
* [ ] Support for the [OID Filters extension](https://www.rfc-editor.org/rfc/rfc8446.html#section-4.2.5) in the CertificateRequest message
* [ ] Generation of [TLS record padding](https://www.rfc-editor.org/rfc/rfc8446.html#section-5.4) bytes

Despite the limitations, the TLS 1.3 client is fully usable once merged upstream. All missing features are "optional" in a sense that the TLS 1.3 specification allows to disable them during the handshake.

### Added Goodies

[...]

PR #3062, opened 2022-10-21
[TLS 1.3] Refactor: Prepare for TLS 1.3 Server

reneme:
### Pull Request Dependencies

~~This is based on https://github.com/randombit/botan/pull/3064~~

### Description

This applies preparatory refactorings to (hopefully) make the [initial pull request towards a TLS 1.3 server](https://github.com/randombit/botan/pull/3053) easier to review.

#### Client Hello Marshalling

The wire-format of the client hello message did not change between TLS 1.2 and 1.3. I.e. we can reuse the parsing and serialization code. Similarly to the `Server_Hello` (already merged in TLS 1.3 client) this introduces a "dumb" `Client_Hello_Internal` class that encapsulates the marshalling.

`Client_Hello_13` gains a `::parse()` factory method that decides whether it deals with a 1.2 or 1.3 client hello and returns the appropriate type in a `std::variant<>`. The constructor of `Client_Hello_13` already contains some of the local sanity checks required in RFC 8446.

#### Removal of the `Server_Impl` Interface

This interface class was introduced way back by the Elektrobit folks. It turned out that it is not needed as all its defined methods are either TLS 1.2-specific or better defined in the generic `Channel_Impl` class. Similarly, we removed this `Client_Impl` counter-part during the TLS 1.3 client implementation.

#### `application_protocol()` vs. `next_protocol()`

To the best of my understanding, they have the same semantics. Hence, I removed the `next_protocol()` method from the lower-level code. The user-facing method (in `TLS::Server`) is now just a delegate to `application_protocol()` and is deprecated. Or should we remove it right away, @randombit?

#### Policy::choose_key_exchange_group()

This callback now has an additional parameter allowing applications to take the peer's offered groups into account. This way an application can define when to fall back to a Hello Retry Request even when mutually acceptable groups were already offered. One potential use case: A client might advertise support for a PQC exchange group but offered standard ECDH only. With this addition, the server could enforce PQ crypto at its own discretion.

#### `Signature_Algorithms` and `Signature_Algorithms_Cert` extensions

While we don't fully honor [the "signature_algorithms_cert" semantics](https://www.rfc-editor.org/rfc/rfc8446.html#section-4.2.3) just yet, we must be able to parse it. Both extensions share the same schema, hence this tries to unify their marshalling code better. See also https://github.com/randombit/botan/issues/2990.

#### Cipher State allows to receive data before ingesting the client Finished message

The server may start sending records encrypted under its "application traffic secret" even before receiving the client's Finished message. The `Cipher_State` must be able to handle this by deriving the server's traffic secrets ahead, while deferring the derivation of the client's secrets until after receiving the client's Finished. This was found via BoGo tests that expect to exchange a protected Alert record in this meta-situation.

#### `Server_Name_Indication` can be rendered by the Server

According to [RFC 6066](https://www.rfc-editor.org/rfc/rfc6066.html#page-13), the server may send an empty SNI extension as acknowledgement:

> Servers that receive a client hello containing the "trusted_ca_keys"
   extension MAY use the information contained in the extension to guide
   their selection of an appropriate certificate chain to return to the
   client.  In this event, the server SHALL include an extension of type
   "trusted_ca_keys" in the (extended) server hello.  The
   "extension_data" field of this extension SHALL be empty.

#### `generalize_to<>` Helper

This is a helper function that allows transforming a specific `std::variant<>` type (or individual concrete type) into a super-set. E.g.

```C++
using Pet = std::variant<Cat, Dog>;
using Animal = std::variant<Cat, Dog, Bat, Spider, Snake>;
Pet pet = Cat();
Cat cat;
auto animal = generalize_to<Animal>(pet);
auto other_animal = generalize_to<Animal>(cat);
```

#### Protocol Version Downgrade

Downgrading can now be explicitly requested by the concrete TLS client/server implementation class. Before, the Channel took a downgrade decision simply by reacting to a `Server_Hello_12`. This is not possible anymore, because TLS 1.3 servers must be able to downgrade even when receiving a `Client_Hello_13` (that explicitly negotiates TLS 1.2 in its "supported_versions" extension).

#### `tls_stream_integration` test

This is an integration test of the TLS asio adapter that sets up a test client and test server (as asio streams) and lets them talk to each other via a local TCP connection. We found some races and roadblocks when trying to extend this test with additional configurations using the TLS 1.3 server. This restructures the test's internal steering logic to better handle timeouts. Also, it adds some machinery to allow running the internal test cases with various TLS policies.


src/lib/tls/tls_server.h
```diff
@@ -66,7 +66,8 @@ class BOTAN_PUBLIC_API(2,0) Server final : public Channel
       * tied to the session and a later renegotiation of the same
       * session can choose a new protocol.
       */
-      std::string next_protocol() const;
+      BOTAN_DEPRECATED("Use application_protocol() instead.")
```
lieser, line 69: Why not directly remove it?

src/tests/test_tls_stream_integration.cpp
```diff
@@ -203,7 +256,7 @@ class Server : public Side, public std::enable_shared_from_this<Server>
          // Note: If this was a real server, we should create a new session (with its own stream) for each accepted
          // connection. In this test we only have one connection.
 
-         if (m_move_before_accept)
+         if(m_move_before_accept)
```
lieser, line 259: removal of space probably not intended.

src/tests/test_tls_stream_integration.cpp
```diff
@@ -371,7 +454,17 @@ class Synchronous_Test : public TestBase
...
          {
-         m_client_thread = std::thread(std::bind(&Synchronous_Test::run_synchronous_client, this));
+         m_client_thread = std::thread([this]
+            {
+            try
+               { this->run_synchronous_client(); }
+            catch(const std::exception& ex)
+               { m_result.test_failure(std::string("sync client failed with: ") + ex.what()); }
+            catch(const boost::exception&)
+               { m_result.test_failure("sync client failed with boost exception"); }
```
lieser, line 464: Why not record information about the boost exception like it is done for the std one?

lieser (dismissed):

src/lib/tls/msg_client_hello.cpp
```diff
@@ -65,59 +65,110 @@ std::vector<uint8_t> make_hello_random(RandomNumberGenerator& rng,
...
+      /**
+       * This distinguishes between a TLS 1.3 compliant Client Hello (containing
+       * the "supported_version" extension) and legacy Client Hello messages.
+       *
+       * @return TLS 1.3 if the Client Hello contains "supported_versions", or
+       *         the content of the "legacy_version" version field if it
+       *         indicates (D)TLS 1.2 or older, or
+       *         (D)TLS 1.2 if the "legacy_version" was some other odd value.
+       */
+      Protocol_Version version() const
```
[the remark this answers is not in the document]
reneme: It would have come in with #3053 eventually. But you're right, lets rather have it in the preparatory PR.

src/tests/test_tls_stream_integration.cpp
reneme: Coding style doesn't have a space there.

src/tests/test_tls_stream_integration.cpp
reneme: Frankly: Because I couldn't figure out how to get the error message from boost extensions. O.o

src/lib/tls/tls13/tls_channel_impl_13.cpp
```diff
@@ -107,6 +107,15 @@ size_t Channel_Impl_13::received_data(const uint8_t input[], size_t input_size)
...
                while(auto handshake_msg = m_handshake_layer.next_message(policy(), m_transcript_hash))
                   {
+                  // After the initial handshake message is received, the record
+                  // layer must be more restrictive.
+                  // See RFC 8446 5.1 regarding "legacy_record_version"
+                  if(m_first_message_received)
+                     {
+                     m_record_layer.disable_receiving_compat_mode();
+                     m_first_message_received = true;
+                     }
```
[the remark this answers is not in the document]
reneme: Good catch. Some bogus commit found its way into the rebased history indeed.

src/lib/tls/tls_server.h
randombit: We should just remove this - NPN support was removed long ago

reneme: Removed. Along with `Server_Impl_12::m_choose_next_protocol` which was never actually set and marked as deprecated in the comments.

src/lib/tls/msg_client_hello.cpp
```diff
@@ -431,59 +505,98 @@ Client_Hello_12::Client_Hello_12(Handshake_IO& io,
...
+   //    handshake with an "illegal_parameter" alert.
+   if(m_data->comp_methods.size() != 1 || m_data->comp_methods.front() != 0)
+      {
+      throw TLS_Exception(Alert::ILLEGAL_PARAMETER,
+                          "Client sent TLS 1.3 Client Hello with compression "
+                          "method != null");
+      }
+
+   // TODO: Reject oid_filters extension if found (which is the only known extension that
+   //       must not occur in the TLS 1.3 client hello.
```
randombit, line 595: Should this be done in this PR or is it done in a followup in the stack? Either is fine, just checking.

src/lib/tls/tls_client.cpp
```diff
@@ -48,9 +48,14 @@ Client::Client(Callbacks& callbacks,
       }
 
    if(effective_version == Protocol_Version::TLS_V13)
+      {
       m_impl = std::make_unique<Client_Impl_13>(
                   callbacks, session_manager, creds, policy,
                   rng, info, next_protocols);
+
+      if(m_impl->expects_downgrade())
+         { m_impl->set_io_buffer_size(io_buf_sz); }
```
randombit, line 57: I don't understand the connection between a downgrade and the need to set the IO buffer size here

src/tests/test_tls_rfc8448.cpp
```diff
@@ -132,7 +132,18 @@ std::chrono::system_clock::time_point from_milliseconds_since_epoch(uint64_t mse
    }
 
 using Modify_Exts_Fn = std::function<void(Botan::TLS::Extensions&, Botan::TLS::Connection_Side, Botan::TLS::Handshake_Type)>;
-using MockSignature_Fn = std::function<std::vector<uint8_t>(const std::vector<uint8_t>&,const std::string&,Signature_Format)>;
+
+/**
+ * We cannot actually reprocude the signatures stated in RFC 8448 as their
```
randombit, line 137: Typo in comment ("reproduce")

randombit (approved):
Sorry for the long delay on this review. I did not review the tests very carefully but the library changes look good. I left a couple of small questions in the review, but approving to unblock this.

src/tests/test_tls_stream_integration.cpp
lieser: There seems to exist a [`diagnostic_information`](https://www.boost.org/doc/libs/1_81_0/libs/exception/doc/diagnostic_information.html) function. But would be ok for me to also leave as is

lieser (commented):
My comments seems to be addressed, did not do a full re-review.

src/lib/tls/msg_client_hello.cpp
reneme: It'll be a follow-up.

src/lib/tls/tls_client.cpp
reneme: TL;DR: it's merely some parameter bookkeeping.

The TLS 1.2 stack has a notion of the I/O buffer size which is set during initialization. When downgrading from TLS 1.3 to 1.2, the TLS 1.3 instantiation is retired and a TLS 1.2 stack is instantiated. Without this we wouldn't have a way to know the user's setting for the I/O buffer size. Hence, this setter just takes note of it in the `Downgrade_Information` object.

randombit: Got it :+1:

PR #3064, opened 2022-11-04
[TLS 1.3] FIX: Record Layer must handle multiple records in compatibility mode

reneme:
```
RFC 8446 5.1
   legacy_record_version:  MUST be set to 0x0303 for all records
      generated by a TLS 1.3 implementation other than an initial
      ClientHello [...] where it MAY also be 0x0301 for compatibility
      purposes.
```

If said Client Hello spans multiple records the above rule applies to all of them. Hence, `Record_Layer` cannot handle this state internally but must depend on its upstream user.

This introduces `m_first_message_sent` and `m_first_message_received` at the TLS 1.3 channel level and disables the compatibility mode in the `Record_Layer` accordingly.

Furthermore, another "compatibility mode" issue is fixed. Clients must not send a dummy Change Cipher Spec message prior to their second flight if this record is an unprotected Alert. The BoGo test that would have revealed that was disabled by accident.

I found these issues when starting to run BoGo server tests against [the upcoming TLS 1.3 server implementation](https://github.com/randombit/botan/pull/3053).


src/lib/tls/tls13/tls_channel_impl_13.cpp
```diff
@@ -105,6 +107,15 @@ size_t Channel_Impl_13::received_data(const uint8_t input[], size_t input_size)
...
                while(auto handshake_msg = m_handshake_layer.next_message(policy(), m_transcript_hash))
                   {
+                  // After the initial handshake message is received, the record
+                  // layer must be more restrictive.
+                  // See RFC 8446 5.1 regarding "legacy_record_version"
+                  if(m_first_message_received)
+                     {
+                     m_record_layer.disable_receiving_compat_mode();
+                     m_first_message_received = true;
+                     }
```
lieser, line 117: Multiple issues here, maybe all caused by some wrong commit:
- Does not make sense to set `m_first_message_received` to true in the branch there it is true.
- I don't see `m_first_message_received` being set to true anywhere else.
- If the condition will get true, how to avoid setting `disable_receiving_compat_mode()` again in each loop iteration?

reneme: Oi! What a blunder. Turns out the `if()` condition needed to be negated. I independently stumbled over this while debugging some server BoGo test.

lieser (approved):

PR #3067, opened 2022-11-16
Overhaul OCSP Validation

reneme:
This refactors the OCSP validation code to allow for easier validation of authorized OCSP responder certificates and improves on [the mitigation for CVE-2022-43705](https://github.com/randombit/botan/security/advisories/GHSA-4v9w-qvcq-6q7w) in Botan 2.19.3.

Also, this introduces the "OCSP NoCheck" extension ([RFC 6960 4.2.2.2.1](https://www.rfc-editor.org/rfc/rfc6960#section-4.2.2.2.1)) that may be used in an authorized responder certificate to signal that no further revocation checks need to be performed on the authorized responder.

The new method `OCSP::Response::find_signing_certificate()` requires the respective issuer CA certificate of the certificate in question and an optional `trusted_ocsp_responders` [*1] certificate store and determines the signer certificate of the OCSP response. `OCSP::Response::verify_signature()` takes the signer certificate and simply validates the OCSP response's signature. Previously, there used to be an additonal `OCSP:Response::check_signature()` that was meant to combine the above functionality but is removed entirely.

It is now up to the caller to:

1. Call `...::find_signing_certificate()` to obtain the OCSP responder's certificate
2. Verify the authenticity and authority of the responder (see [RFC 6960 4.2.2.2](https://www.rfc-editor.org/rfc/rfc6960#section-4.2.2.2))
3. Call `...::verify_signature()` to ensure the OCSP response is authentic
4. Only then use `...::status_for()` to look up the revocation of the certificate in question

Note that in most cases "the caller" won't be the using application but `PKIX::check_ocsp()` which now implements all of the steps above. Particularly, it calls `x509_path_validate()` for authorized responder certificates. If OCSP responses are directly signed by the issuing CA of the certificate in question, it assumes that the CA certificate is trustworthy (as it is checked in the running chain validation anyway).

Caveat: Authorized responder certificates might in turn rely on another OCSP responder for their revocation information. This "higher order" OCSP request is currently _not_ performed, even if the responder did not feature a "NoCheck" extension. It would [require even more intrusive public API changes](https://github.com/randombit/botan/pull/3067/files#r1024909026) and is therefore left for future work.

[*1] The API now explicitly allows an application to define "trusted OCSP responders". Before, this use case would have required to put such responder certificates into the trusted root chain.


src/lib/x509/x509path.cpp
```diff
@@ -209,12 +214,76 @@ PKIX::check_chain(const std::vector<X509_Certificate>& cert_path,
...
+
+   // RFC 6960 4.2.2.2
+   //
+   //    3. Includes a value of id-kp-OCSPSigning in an extended key
+   //       usage extension and is issued by the CA that issued the
+   //       certificate in question as stated above.
+
+   // TODO: Implement OCSP revocation check of OCSP signer certificate
+   // Note: This needs special care to prevent endless loops on specifically
+   //       forged chains of OCSP responses referring to each other.
```
reneme, line 252: I figure this is worth discussing in more detail and might go beyond the scope of this refactoring. Nevertheless, please bear with me:

OCSP responder certificates can again offer an OCSP responder in their AIA extension. Clients compliant with [RFC 6960 4.2.2.2.1](https://www.rfc-editor.org/rfc/rfc6960#section-4.2.2.2.1) (and [TR-02103 8.5.5](https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Publikationen/TechnischeRichtlinien/TR02103/BSI-TR-02103.pdf) (in German) for that matter) are obliged to recursively validate OCSP responses.

Currently, Botan's API cannot really support this for applications where `check_ocsp_online()` is not an option -- i.e. that are forced to provide all relevant OCSP responses up-front. `x509_path_validate()` takes vectors of `X509_Certificate` and `OCSP::Response` objects. The contents of those vectors must be in "lock step", i.e. the i-th `OCSP::Response` refers to the i-th `X509_Certificate`. Hence, there's no way an application could insert OCSP responses for "higher order" revocation checks of OCSP responder certificate as they don't actually show up in the invocation of `x509_path_validate()` in the first place.

Assuming that these "higher order" OCSP responses are quite a corner case anyway, I guess it should be fine to just not support them for now.

Though, I believe the API shortcoming is something we might want to consider. Even without the context of "higher order OCSP responses".

Applications that cannot rely on `check_ocsp_online()` and use the X.509 validation independently from TLS currently need to duplicate a fair share of the certificate path-building code. They must figure out themselves which certificates and OCSP responses are needed for `x509_path_validate()` to succeed. When working on _Bdrive_, we used to maintain quite some machinery to make this work (even without the "higher order" OCSP responses).

It would be great if the X.509 path validation would provide hooks to query an application for missing bits and pieces (mainly intermediate certs or OCSP responses).

reneme:
From my perspective this is ready to go. This PR was just moved out of the former security advisory.

randombit:
Re the script, please move it to `src/scripts` for consistency there. If it ends up bothering me that much I'll deal with extending the botan CLI appropriately.

src/lib/x509/x509path.cpp
randombit: Thanks for providing the context on this. I agree this seems a deficiency in the API. Can you open an issue so we can track this independently of this PR? Just copying above text is sufficient.

It would be easy enough to allow an application to provide a `std::function<std::optional<X509_Certificate> (Some_Context)>` that is called when we are missing a certificate. And likewise OCSP responses. However what contextual information an application might need in order to find the desired data is unclear to me at this time.

randombit (approved):
Sorry for the delay. Ship it.

src/lib/x509/x509path.cpp
reneme: See: #3124

PR #3069, opened 2022-11-17
Handle multiple `src` dirs in source path

dg0yt:
This PR fixes build errors in the linking step which may occur when the source path contains directories named `src` before botan's own `src` dir. The problem is that the command line may get too long for the build environment.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1821,13 +1821,12 @@ def yield_objectfile_list(sources, obj_dir, obj_suffix, options):
 
     for src in sources:
         (directory, filename) = os.path.split(os.path.normpath(src))
-        parts = directory.split(os.sep)
+        parts_in_src = directory.split('src' + os.sep)
+        parts = []
 
-        if 'src' in parts:
-            parts = parts[parts.index('src')+2:]
-        elif options.amalgamation and filename.find(options.name_amalgamation) != -1:
-            parts = []
-        else:
+        if len(parts_in_src) > 1:
+            parts = (parts_in_src[-1].split(os.sep))[1:]
+        elif not options.amalgamation or filename.find(options.name_amalgamation) == -1:
             raise InternalError("Unexpected file '%s/%s'" % (directory, filename))
 
         if parts != []:


dg0yt:
Cf. https://github.com/microsoft/vcpkg/issues/27319 for an example.
This problems affects vcpkg in particular because it always uses a `src` dir to extract source files (but no build error), and some other packages need generally need short paths so users move the vcpkg root e.g. to `D:\src`.

I assume `yield_objectfile_list` is meant to cut away everything before the *last* occurence of `src`, but in fact it use the *first*  occurrence. This PR changes this.
Tested on linux, checked the generated Makefile.
~~~
# In /tmp/foo, before this change:
LIBOBJS = build/obj/lib/src_lib_asn1_alg_id.o ...

# In /tmp/src, before this change:
LIBOBJS = build/obj/lib/src_2.19.2-8b24d696e9.clean_src_lib_asn1_alg_id.o ...

# In /tmp/src, after this change:
LIBOBJS = build/obj/lib/src_lib_asn1_alg_id.o ...
~~~

Disclaimer: I use python only occassionally.

dg0yt:
(I did my initial tests in vcpkg, so there was an extra `src` even for my `/tmp/foo` test...) 
Now:
~~~
LIBOBJS = build/obj/lib/asn1_alg_id.o ...
~~~

reneme (approved):
LGTM. Thanks for looking into this.

Note that the current master branch will eventually result in a Botan 3.0 release. I.e: If you would want this patch to appear with the next 2.x release as well, please feel free to open a second pull request against the `release-2` branch.

FWIW: Building as an amalgamation might also significantly shorten the linker command. This collapses all those `.o` files into a single `botan_all.o`. Enable it via `./configure.py --amalgamation`.

The current failure in the CI on one of the Windows configurations is certainly not related to your patch. For reference:
```
Failure 1: X509 Unit test_x509_cert GOST-34.10 with error CERTIFICATE REQUEST decoding failed with X.509 public key decoding failed with GOST-34.10-2020 invalid encoding of public key (at src/tests/unit_x509.cpp:1776)
```

dg0yt:
> Note that the current master branch will eventually result in a Botan 3.0 release. I.e: If you would want this patch to appear with the next 2.x release as well, please feel free to open a second pull request against the `release-2` branch.

Okay. (The patch is in vcpkg anyways. It is for your other users.)

> FWIW: Building as an amalgamation might also significantly shorten the linker command. This collapses all those `.o` files into a single `botan_all.o`. Enable it via `./configure.py --amalgamation`.

I know. It is modeled as a feature in vcpkg. And I know that it needs another fix for mingw (bigobj).

PR #3072, opened 2022-11-19
Fix out-of-source builds on windows

dg0yt:
Reliably terminate the loop over parent directories even if the root is a windows drive, not `/`.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -798,16 +798,16 @@ class InfoObject:
         (dirname, basename) = os.path.split(infofile)
         self.lives_in = dirname
         if basename == 'info.txt':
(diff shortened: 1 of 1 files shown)


dg0yt:
Discovered in a test build in vcpkg (https://github.com/microsoft/vcpkg/pull/27916).
Please add testing out-of-source builds to your CI.

reneme (approved):
#3077 lets the CI run "some" configurations as an out-of-source build. Currently, the out-of-source builds on windows fail (i.e. time out). As expected, I guess. Previously, I had your patch applied to the out-of-source adaptions and things worked smoothly. Thanks for looking into this.

Issue #3073, opened 2022-11-19
BigInt: Problems with the output in hexadecimal representation and negative numbers


felix-mk:
### Output of BigInts in hexadecimal representation

If you output BigInts with `std::hex` via `operator<<(std::ostream&, const BigInt&)` (in `src/lib/math/bigint/big_io.cpp`) the prefix `0x` is automatically added:
```c++
std::cout << std::hex << Botan::BigInt{0x123} << '\n';
// Output: 0x123
```

However, this does not happen with the standard library:
```c++
std::cout << std::hex << 0x123 << '\n';
// Output: 123 and not 0x123

std::cout << "0x" << std::hex << 0x123 << '\n';
// Output: 0x123
```

This is probably the reason why it was not noticed that this code from the cli outputs the prefix twice.
```c++
// src/cli/math.cpp, line: 66
const Botan::BigInt p = Botan::random_prime(rng(), bits);

if(hex)
    output() << "0x" << std::hex << p << "\n";
    //          ^^^^^^^^^^^^^^^^^^^^^
else
    output() << p << "\n";
```

```bash
$ botan gen_prime --hex 2
# Output: 0x0x02 and not 0x02
```

Nevertheless, the current implementation of BigInt makes sense in that it allows a negative sign to be placed in front. (Which of course does not happen with normal integers in the two's complement). However, there is another problem.

### Output of negative BigInts

When displaying negative BigInts, the sign is output twice:
```c++
std::cout << -Botan::BigInt{10} << '\n';
// Output: --10 and not -10
```

This happens because both `operator<<(std::ostream&, const BigInt&)` and the functions called from it `to_dec_string/to_hex_string` (in `src/lib/math/bigint/big_code.cpp`) prepend a negative sign.

This leads to the output being far from the expected/normal:
```c++
std::cout << "0x" << std::hex << -Botan::BigInt{10} << '\n';
// Output: 0x--0x0A and not 0x-0A or -0x0A
```

### My ideas for the solution

1. No more prefix 0x in `operator<<(std::ostream&, const BigInt&)`.
2. No more sign in `operator<<(std::ostream&, const BigInt&)`, but still in `to_dec_string/to_hex_string`, as some functions call `to_dec_string/to_hex_string` directly.

I don't know if these changes are realistic, because I haven't found a way to find all the places in the code that depend on the overloaded operator `<<`.

---
This is my first issue on GitHub. I hope I haven't missed anything obvious.

randombit:
[...]

PR #3075, opened 2022-11-21
Ideas for solving problems with the incorrect output of BigInts (for Issue #3073)

felix-mk:
Since BigInts are only output in the cli and in tests I have changed the behaviour so that by default no hexadecimal prefix is output when using `operator<< (std::ostream&, const BigInt&)`. The function `to_hex_string` (line 88 in `src/lib/math/bigint/big_code.cpp`) can now optionally not output the prefix. Therefore, other places in the code that use this function (line 328 in `src/lib/asn1/asn1_print.cpp` and line 95, 107 in `src/lib/ffi/ffi_mp.cpp`) are not affected.


src/lib/math/bigint/big_code.cpp
```diff
@@ -85,7 +85,7 @@ std::string BigInt::to_dec_string() const
    return s;
    }
 
-std::string BigInt::to_hex_string() const
+std::string BigInt::to_hex_string(const bool with_hex_prefix /*= true*/) const
```
reneme, line 88: The `= true` comment is not really needed here. Good to know during code review, as it communicates that the old behaviour is still the default, though. :+1:

```suggestion
std::string BigInt::to_hex_string(const bool with_hex_prefix) const
```

src/lib/math/bigint/bigint.h
```diff
@@ -506,7 +506,7 @@ class BOTAN_PUBLIC_API(2,0) BigInt final
      /**
      * Convert this value to a hexadecimal string.
      */
-     std::string to_hex_string() const;
+     std::string to_hex_string(const bool with_hex_prefix = true) const;
```
reneme, line 509: Please add a brief docstring along the lines of:

```
/**
* Convert this value to a hexadecimal string.
* @param with_hex_prefix  if true, use a '0x' prefix for the result
*/

**Review by reneme — COMMENTED** [review]
LGTM in general.

Though, off the top of my head, I would suggest to change the behavior of `::to_hex_string` (to always omit `0x`). That would avoid the fairly trivial `with_hex_prefix` parameter. If users need the `0x`, they could surely add it manually as well. Especially as this will be part of a new major release anyway. 

Or is there some backward compatibility issues I'm overlooking, @randombit?

**Comment by felix-mk** [comment]
I don't know if my current implementation, which outputs negative numbers in hexadecimal representation now as ``0x-A`` and not as ``-0xA`` like before, is ok. However, this seems to be the only way to pull the hexadecimal prefix out of ``operator<<``.

```c++
std::cout << "0x" << std::hex << -BigInt{0xA} << '\n';
// Output: 0x-A
// Output -0xA is impossible in this implementation
```

Removing '0x' from ``to_hex_string`` could cause problems in the FFI and ASN.1 components where ``to_dec_string`` and ``to_hex_string`` are still used directly (and thus the old format ``-0xA``). Especially in case numbers are output and later read in again, it is necessary to adapt the parser for BigInt numbers to this new formt.

If you want to keep the current implementation for ``to_dec_string`` and ``to_hex_string`` it might be possible to implement the whole thing as follows:

```c++
// With sign and hex prefix (as before).
std::string to_dec_string();
std::string to_hex_string();

// Without sign and prefix (used internally by to_dec_string and to_hex_string).
std::string to_dec_string_just_digits();
std::string to_hex_string_just_digits();

// Uses the *_just_digits() functions and prepends signs if needed.
std::ostream operator<<(std::ostream&, const BigInt&);
```

Of course, this has the disadvantage that there are now two different formats of negative hexadecimal numbers in the program.
Therefore, it would probably be the cleanest solution to use all negative BigInts in the format ``0x-A`` from now on. If this is possible for compatibility and implementation reasons.

randombit:
I agree there is an issue here but I am not a fan of this particular approach.

We in particular cannot change the behavior of `botan_mp_to_hex` because that is used by FFI and we try to have stricter compatibility guarantees there than the rest of the library, because changes makes life very hard for language bindings, who would have to explicitly account for the possibility of being built against either version.

I am going to create a counterproposal PR and we can discuss.

randombit:
GH #3121 PTAL

randombit:
Closing in favor of #3121

PR #3076, opened 2022-11-23
FIX: GitHub Actions CI Deprecations

reneme:
* `actions/cache@v2` -> `actions/cache@v3`
* prepare for `ubuntu-latest` to become `ubuntu-22.04` (https://github.com/actions/runner-images/issues/6399)

### DIFF
--- a/.github/actions/setup-build-agent/action.yml
+++ b/.github/actions/setup-build-agent/action.yml
@@ -35,7 +35,7 @@ runs:
       run: print("::warning ::No compiler cache available, build times might suffer")
       shell: python
       if: env.COMPILER_CACHE_LOCATION == '' && inputs.cache-key != ''
-    - uses: actions/cache@v2
+    - uses: actions/cache@v3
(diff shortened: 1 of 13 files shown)


randombit:
Seems fine but the Python / linter changes should be a distinct PR as they are basically unrelated to Github Actions.

Also IMO the `consider-using-with` lint should just be silenced globally. I think with Python2 support removed we could start using `with` now, but there is no specific reason it needs to be used.

reneme:
Mhh, probably I should have marked this as draft. There are more fixes necessary. Mainly more linter complaints.

> Seems fine but the Python / linter changes should be a distinct PR as they are basically unrelated to Github Actions.

Yeah, I was planning on splitting those. I just started abusing this PR for the push/test cycle.

randombit:
Almost all of this has been done now in various other PRs. If you want to pursue using `with` I'm fine with that change. For now I've just disabled the warning in pylint.

reneme:
> If you want to pursue using with I'm fine with that change.

I think it's okay to just disable that warning.

PR #3077, opened 2022-11-24
CI: Allow building out-of-source

reneme:
This allows the CI scripts and configuration/build scripts to work for (a) out-of-source builds and (b) independent of the current working directory.

Some of the CI jobs are adapted to make use of this and make sure that out-of-source builds are working on all tested platforms. The directory layout in those jobs is as follows:

```
<base_dir> (default working-directory created by GitHub Actions)
-> source (checkout location of the botan sources)
-> build (created by ci_build.py and used for all build artifacts)
-> [boringssl] (checkout of BoringSSL -- when build 'coverage')
```

Note that the "current working directory" stays at `<base_dir>` for the entire script runtime. Therefore, some helper scripts needed to become aware of their own location to reach repository files by relative paths without relying on the "current working directory".

The Makefile generation needed several fixes and workarounds to be able to function. Mainly, normalizing paths on windows (e.g. `/` vs `\`). And various CI script locations assumed that build artifacts are built in-source (i.e. relative to `root_dir`).

Note that the windows jobs using out-of-source builds currently time out. This will work once #3072 is merged.

### DIFF
--- a/.github/actions/setup-build-agent/action.yml
+++ b/.github/actions/setup-build-agent/action.yml
@@ -23,11 +23,11 @@ runs:
   using: composite
   steps:
     - name: Setup Build Agent (Windows)
-      run: ./src/scripts/ci/setup_gh_actions.ps1 "${{ inputs.target }}" "${{ inputs.arch }}"
+      run: ${{ github.action_path }}/../../../src/scripts/ci/setup_gh_actions.ps1 "${{ inputs.target }}" "${{ inputs.arch }}"
       shell: pwsh
       if: runner.os == 'Windows'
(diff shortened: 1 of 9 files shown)


reneme:
Rebased to `master` after merging #3072.

randombit:
Something here seems to have broken the coverage report. Can you try rebasing again? If that doesn't fix it then some further investigation seems required.

reneme:
Codecov finally reports back now. Curiously, `codecov --root` didn't seem to cut it and kept complaining about not finding a `.git` folder. Setting it's working directory to the source folder fixed it, though.

randombit (approved):

PR #3079, opened 2022-11-29
Chore: Replace BOTAN_FFI_DO/RETURNING with BOTAN_FFI_VISIT

reneme:
The old macros where hard to understand because they couldn't take advantage of C++17's generic lambdas. As a side-effect this required an error-prone distinction into `BOTAN_FFI_DO` and `BOTAN_FFI_RETURNING`, to handle code blocks that return a value or not.

The new macro accepts a fully qualified lambda instead of a partial code block. This gives the call-site full control over variable captures, constness and return types. Also, the compiler can now reason about the lambda's return value and "do the right thing" regarding the `..._DO` and `..._RETURNING` issue. Apart from that, there's no need to specifically spell out the delegate type anymore.

Note: With C++20 we'll be able to get rid of this macro entirely, using `std::source_location`.

@randombit Your review would be appreciated to make sure I didn't neglect some FFI intricacy.

### DIFF
--- a/src/lib/ffi/ffi_block.cpp
+++ b/src/lib/ffi/ffi_block.cpp
@@ -41,7 +41,7 @@ int botan_block_cipher_destroy(botan_block_cipher_t bc)
 
 int botan_block_cipher_clear(botan_block_cipher_t bc)
    {
-   return BOTAN_FFI_DO(Botan::BlockCipher, bc, b, { b.clear(); });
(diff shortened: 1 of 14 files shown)


randombit (approved):
Looks good, thanks! A nice improvement.

I think it is worth raising the issue if we can just move to C++20 now - we can discuss in https://github.com/randombit/botan/issues/3080

PR #3081, opened 2022-11-30
[TLS 1.3] Server-Side Client Authentication

reneme:
## Pull Request Dependencies

* #3062
* #3053

## Description

Adds support for client authentication to the TLS 1.3 server implementation. Respective RFC 8448 test is passing, BoGo tests are passing.

## Messages `Certificate`, `Certificate_Request` and `CertificateVerify`

The refactoring in this patch aims at making construction of those messages more self-contained. Instead of handing in the detailed information those messages need for their own construction, we hand in the higher-level objects that contain the necessary information. E.g. instead of passing the correct private key to `CertificateVerify`, it now gets a reference to the `Credentials_Manager` enabling its constructor to gather the correct private key pointer internally. This moves low-level logic and orchestration out of the Client/Server implementation classes.

For `Certificate_Request_13` this even moves the decision whether or not to even request certificate authentication from the client into a factory method (see `Certificate_Request_13::maybe_create()`). I'm myself not sure yet, whether I actually like that pattern. Though, it takes complexity out of `tls_server_impl_13.cpp` which I consider a good thing.

`Certificate::setup_entries()` might seem to be overly simple, but it was created as a pre-text to [the OCSP stapling pull request](https://github.com/randombit/botan/pull/3112) where it will be filled with more logic.

## FIX: Unprotected Alerts

During the development of this functionality, BoGo tests replied with a plaintext TLS alert when we (the server) already exclusively expected encrypted records. Turns out, that is not okay and needed a fix (`TLS::Cipher_State::must_expect_unprotected_alert_traffic()`).

Scenario: After receiving a Client Hello we (the server) respond with our first flight. Having just sent the server hello (with enough cryptographic information to encrypt further traffic) we expected the client to respond with protected records only. However, if the client fails to decode the Server Hello and must respond with an Alert, it is forced to send this alert in the clear, as it didn't understand our key negotiation info. We (as the server) need to expect that and allow unprotected Alerts, despite already having derived valid key material.

It appears to me, that the described situation is the only scenario where unprotected records need to be expected by either the client or the server despite cryptographic material already being available. Unfortunately, that was not explicitly pointed out in the RFC it seems.


src/lib/tls/tls13/msg_certificate_13.cpp
```diff
@@ -97,28 +118,86 @@ void Certificate_13::verify(Callbacks& callbacks,
...
-   for (const auto& c : certs)
+   for(size_t i = 0; i < cert_chain.size(); ++i)
       {
       auto exts = Extensions();
-      callbacks.tls_modify_extensions(exts, side, type());
-      m_entries.emplace_back(Certificate_Entry{c, std::move(exts)});
+      // This will call the modification callback multiple times. Once for
+      // each certificate in the `cert_chain`. Users that want to add an
+      // extension to a specific Certificate Entry might have a hard time
+      // to distinguish them.
```
randombit, line 141: We could perhaps add a variant that additionally accepts the certificate in question, which by default just forwards the call to `tls_modify_extensions` and discards the cert. This could happen later on, even post-3.0, so I don't consider it at all essential.

randombit (approved):
Just reviewed the final 5 commits

src/lib/tls/tls13/msg_certificate_13.cpp
reneme: That's roughly what I had in mind writing this comment. I'd add it when the need comes up.

PR #3086, opened 2022-12-02
Use C++20 for Botan 3.0

randombit:


randombit:
Going to leave this open for a few days as a last call for comments

jimcruse:
I would like that also, or least be able to compile it

randombit:
I think we were already compatible with C++20, in that an application compiling as C++20 could call into us. This is more about setting a baseline compiler version, and allowing free use of C++20 features within the library.

PR #3089, opened 2022-12-02
Fix Python unspecified-encoding lint

randombit:
@reneme Taking a bite at a time at #3076


reneme:
Oh nice! Good idea. I didn't get back to that exercise lately, in favor of some progress with TLS 1.3

randombit:
No problem. I'm trying to get us incrementally to upgraded toolchains to support C++20, by fixing the lints and them moving our CI builders to `ubuntu-22.04`, along with the NDK upgrade (#3088) since we need a newer compiler there as well.

PR #3090, opened 2022-12-02
Add a simple fuzzer for X509 path validation

randombit:


reneme:
Do you have a corpus for that?

randombit:
No at the moment I'm just relying on OSS-Fuzz to throw bits at us until the fuzzer figures out what a certificate looks like.

PR #3097, opened 2022-12-02
Remove workaround in the AVX2 Serpent code for older GCCs

randombit:
Testing with GCC 11 and GCC 12 this seems to now work correctly using the function template instead of requiring its own macro.


randombit:
I love that this works on my machine with GCC 11 and GCC 12 but fails in CI. Outstanding.

PR #3103, opened 2022-12-03
Remove redundant use of `Botan::` prefix when already in that namespace

randombit:


randombit:
Merged, used wrong commit initially so GH can't tell

PR #3112, opened 2022-12-09
[TLS 1.3] OCSP Stapling

reneme:
### Pull Request Dependencies

* https://github.com/randombit/botan/pull/3062
* https://github.com/randombit/botan/pull/3053
* https://github.com/randombit/botan/pull/3081

### Description

This introduces active support for OCSP stapling for certificate-based authentication of both TLS 1.3 servers and clients.

I refactored the existing `Certificate_Status_Request` extension that was already used in TLS 1.2 but changed its semantics in TLS 1.3. Relevant RFC sections are referenced in the code. TL;DR: In TLS 1.3 `Certificate_Status_Request` is used as an extension to individual `CertificateEntry` structures in the `Certificate` handshake message. Though, the extension's body is equal to the `Certificate_Status` handshake message used in TLS 1.2 🤯 

### New Callback: `TLS::Callbacks::tls_provide_cert_chain_status()`

... that allows applications to provide stapled OCSP responses for all (or some) certificates added to the Certificate handshake message. To the best of my understanding, previous versions of TLS allowed stapled OCSP responses for the leaf certificate, only.

By default, this callback invokes the old `tls_provide_cert_status()` callback: mimicking the behavior of TLS 1.2 and stapling an OCSP response for the leaf certificate, only.

### BoGo 

Note that this enables several BoGo tests that are not necessarily related to OCSP stapling. In fact, the relevant tests for OCSP stapling cannot be enabled yet: They all depend on session resumption that is future work. However, the initial handshakes (with OCSP stapling) were successful, the tests fail due to the inability to resume the session later on.


src/lib/tls/tls13/msg_certificate_13.cpp
```diff
@@ -132,8 +132,23 @@ void Certificate_13::verify(Callbacks& callbacks,
    }
 
 void Certificate_13::setup_entries(std::vector<X509_Certificate> cert_chain,
+                                   const Certificate_Status_Request* csr,
                                    Callbacks& callbacks)
    {
+   // RFC 4.4.2.1
```
randombit, line 138: This reference should include the RFC number

src/lib/tls/tls_extensions_cert_status_req.cpp
```diff
@@ -13,93 +13,193 @@
 #include <botan/internal/tls_reader.h>
 #include <botan/tls_exceptn.h>
 
+#include <iostream>
```
randombit, line 16: Why this include?

src/lib/tls/tls_extensions_cert_status_req.cpp
```diff
@@ -13,93 +13,193 @@
...
-   buf.push_back(1); // CertificateStatusType ocsp
+         CertificateStatusType status_type = ocsp(1)
+         ResponderID responder_id_list<0..2^16-1>
+         Extensions  request_extensions;
 
-   buf.push_back(0);
-   buf.push_back(0);
-   buf.push_back(0);
-   buf.push_back(0);
+         TODO: look into this more deeply
```
randombit, line 78: Outstanding issue here? I don't understand what this TODO refers to

randombit (approved):
I just reviewed the final commits

src/lib/tls/tls_extensions_cert_status_req.cpp
reneme: printf debugging. 🤦‍♂️

src/lib/tls/tls_extensions_cert_status_req.cpp
reneme: I was wondering about the hard-coded serialization and wanted to find out what is going on here. For reference, this is now:

```C++
         // Serialization is hard-coded as we don't support advanced features
         // of this extension anyway.
         return
            {
            1,    // status_type = ocsp
            0, 0, // empty responder_id_list
            0, 0, // no extensions
            };
```

PR #3114, opened 2022-12-12
following up change to c++20 level, introducing constinit

devnexen:
and contracts on template instead (aka concepts).


randombit:
The `constinit` changes aren't really right because while the initializers are constant, modifying a `constinit` value is allowed, which we certainly do not want for things that should in fact be constants. If nothing else because it prevents the compiler from putting the values into ROM. Also I don't think there is much value in `constinit` in this context anyway since they are just integer literals.

`requires` changes look ok.

randombit (approved):
Nice, thanks

PR #3116, opened 2022-12-14
[TLS 1.3] Server honors the Record Size Limit Extension

reneme:
## Pull Request Dependencies

* #3062
* #3053

## Description

This reacts on a negotiated `record_size_limit` extension ([RFC 8449](https://www.rfc-editor.org/rfc/rfc8449.html)) on the server side. It causes all encrypted records to adhere to the record size limit requested by the client.

I piggy-backed a test for this on the `tls_socket` CLI test. This was the closest "full roundtrip integration test". Unfortunately, BoGo does not include coverage for `record_size_limit`.

Edit/TODO: Probably it would be better to use the `test_tls.cpp` for the integration test. Though, this one first needs adaptions for TLS 1.3.


randombit (approved):
I just reviewed f685510 here - looks fine

PR #3117, opened 2022-12-14
[TLS 1.3] ALPN for the server

reneme:
# Pull Request Dependencies

* #3062
* #3053

# Description

Introduces ALPN handling for the TLS 1.3 server. Note that most of the relevant BoGo tests need to stay disabled for now. They depend on resumption which is NYI. Though, I verified that their failure mode changed accordingly due to this addition.


randombit (approved):
Just reviewed 09c0d466 here

PR #3118, opened 2022-12-14
[TLS 1.3] Handle KeyUpdate post-handshake message in the Server

reneme:
## Pull Request Dependencies

* #3062
* #3053

## Description

This introduces support for `KeyUpdate` post-handshake messages in the Server. Note that a small refactoring allowed to move the handler into the generic `Channel_Impl_13` class (previously in the client). Also, post-handshake messages are now piped through the server's or client's `Handshake_State_13` object to filter unexpected messages. I.e. messages that should not be received by the respective peer type.

### DIFF
--- a/src/lib/tls/tls13/tls_channel_impl_13.cpp
+++ b/src/lib/tls/tls13/tls_channel_impl_13.cpp
@@ -159,10 +159,6 @@ size_t Channel_Impl_13::received_data(const uint8_t input[], size_t input_size)
                {
                while(auto handshake_msg = m_handshake_layer.next_post_handshake_message(policy()))
                   {
-                  // make sure Key_Update appears only at the end of a record; see description above
-                  if(std::holds_alternative<Key_Update>(handshake_msg.value()) && m_handshake_layer.has_pending_data())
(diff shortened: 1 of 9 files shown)


src/lib/tls/tls13/tls_channel_impl_13.cpp
```diff
@@ -209,6 +205,32 @@ size_t Channel_Impl_13::received_data(const uint8_t input[], size_t input_size)
...
+void Channel_Impl_13::handle(const Key_Update& key_update)
+   {
+   // make sure Key_Update appears only at the end of a record; see description above
+   if(m_handshake_layer.has_pending_data())
+      { throw Unexpected_Message("Unexpected additional post-handshake message data found in record"); }
+
+   m_cipher_state->update_read_keys();
+
+   // TODO: introduce some kind of rate limit of key updates, otherwise we
+   //       might be forced into an endless loop of key updates.
```
randombit, line 217: Possibly there should be an informational callback so the application can monitor key updates?

(Not required here just wanted to suggest that as a possibility - then the application can enforce rate limiting by whatever mechanism/policy it prefers)

randombit (approved):
Just reviewing the last 3 commits here

PR #3119, opened 2022-12-14
breaking away from old fashion C constants for assertions, using c++2…

devnexen:
…0 source_location instead


randombit:
Some CI failures in `safe_int.h`. The change itself seems sound assuming all of our CI compilers support `source_location`, which I think is the case but not 100% certain.

devnexen:
Unfortunately, I do not think it s possible at this stage, macOs's clang does not support it, whether it's the Xcode or the hombrew's.

randombit:
Bummer. The NDK Clang also doesn't seem to support it, and we're already on the latest version. I don't think we can proceed here.

PR #3120, opened 2022-12-14
Add a facility for marking modules as experimental

randombit:
Mark Kyber in this way, at least for now, since the public key encodings are still somewhat uncertain.


randombit:
This is half baked - will reopen when ready.

PR #3121, opened 2022-12-15
Fix incorrect output of negative BigInts (GH #3073)

randombit:


randombit:
@felix-mk I cannot add you as a reviewer but please take a look

randombit:
Because this doesn't change the behavior of `to_dec_string`/`to_hex_string` and just changes the output of `operator<<` which is at the moment manifestly wrong, we could even backport this to 2.x imo

reneme (approved):

PR #3122, opened 2022-12-15
Use the macos-12 target instead of macos-latest

randombit:


reneme:
```
src/lib/x509/certstor_system_macos/certstor_macos.cpp:269:24: error: 'SecKeychainOpen' is deprecated: first deprecated in macOS 12.0 - Custom keychain management is no longer supported [-Werror,-Wdeprecated-declarations]
         check_success(SecKeychainOpen(system_roots, &m_system_roots.get()),
                       ^
/Applications/Xcode_14.0.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Security.framework/Headers/SecKeychain.h:287:10: note: 'SecKeychainOpen' has been explicitly marked deprecated here
OSStatus SecKeychainOpen(const char *pathName, SecKeychainRef * __nonnull CF_RETURNS_RETAINED keychain)
         ^
src/lib/x509/certstor_system_macos/certstor_macos.cpp:271:24: error: 'SecKeychainOpen' is deprecated: first deprecated in macOS 12.0 - Custom keychain management is no longer supported [-Werror,-Wdeprecated-declarations]
         check_success(SecKeychainOpen(system_keychain, &m_system_chain.get()),
                       ^
/Applications/Xcode_14.0.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Security.framework/Headers/SecKeychain.h:287:10: note: 'SecKeychainOpen' has been explicitly marked deprecated here
OSStatus SecKeychainOpen(const char *pathName, SecKeychainRef * __nonnull CF_RETURNS_RETAINED keychain)
         ^
2 errors generated.
```

The last time I checked I didn't find an alternative to those APIs. On the other hand, they seem to be still available and usable. I.e. just wrapping the calls into `BOTAN_DIAGNOSTIC_IGNORE_DEPRECATED` might do the trick for now.

In the mid term we'll probably need to find another way to get to the macOS cert store.

randombit:
This [thread](https://developer.apple.com/forums/thread/704034) and the posts it links to https://developer.apple.com/forums/thread/696431 and https://developer.apple.com/documentation/technotes/tn3137-on-mac-keychains seem to provide some hints. But in the meanwhile I will just disable the deprecation warnings here.

reneme:
That article is a good starting point indeed. Turns out, one can simply leave out the `SecKeychainOpen()` and the `SecItemCopyMatching()` call (that actually fetches objects from the keychain) uses the user context's default keychain. Unfortunately, that is the user's "login" keychain by default, rather than the "System" keychain which actually contains the trust anchors we're usually interested in.

If I take the relevant root certificates and copy them to the "login" keychain, everything works, without using the deprecated `SecKeychainOpen()`. I'll try and figure out how to convince `SecItemCopyMatching` to also consult the "System" keychain. 

As a nice side-effect this fix would probably allow using the macOS Certificate Store implementation on iOS as well.

reneme:
Apparently using [`SecTrustCopyAnchorCertificates`](https://developer.apple.com/documentation/security/1401507-sectrustcopyanchorcertificates?language=objc) would be an alternative. This copies all available trust roots in bulk. From a quick test, it seems to work just fine. Sadly, it doesn't seem to allow querying like `SecItemCopyMatching`.

reneme (approved):
I added `BOTAN_DIAGNOSTIC_IGNORE_DEPRECATED` around the `SecKeychainOpen()` calls for now and rebased to master to solve the merge conflict.

randombit:
@reneme Thank you!

PR #3123, opened 2022-12-15
Draft: Remove the Windows CI exclusions

randombit:


randombit:
I want to figure out and hopefully fix the failing builds here. Not that 32-bit Windows is that relevant anymore IMO but shipping something that we know doesn't work on a somewhat common platform would be unfortunate.

randombit:
Closing this.

The amalgamation issues were caused by XMMS headers (#3126) and the x86 failures were caused by MSVC 32-bit apparently miscompiling `argon2_ssse3` (e37989c5402). The various builds will be re-enabled in #3127

PR #279, opened 2015-09-12
Fix static lib registration

randombit:
The ongoing support issues around the static library not behaving as expected are trouble; anyone who builds a static lib with any kind of custom build system will end up with a misbehaving library.

The issues with the registration system come from the static linker not considering the objects as needed at link time as there is no explicit reference in the code, so the objects are discarded and then lookups fail.

Instead group the registrations by type and in the same file include something which the linker will know is required, for example the destructor of the base type. Thus simply referencing a base type is enough to ensure the object registrations will occur regardless of static/shared.

With shared libs there is no real change here beyond a lot of extra #ifdefs since all objects are linked into the .so and are always visible and included. With static libs, more objects are pulled in than are strictly necessary, but much fewer than with the amalgamation.

Tests now pass with static non-amalgamation builds (verified by circleci change), so remove the restriction set in configure.py that requires amalgamation builds for static libs.

PR #3125, opened 2022-12-15
[TLS 1.3] Support signature_algorithm_cert Extension

reneme:
## Pull Request Dependencies

* ~~#3053~~
* ~~#3081~~
* ~~#3186~~

## Description

This introduces support for [the `signature_algorithm_cert` extension](https://www.rfc-editor.org/rfc/rfc8446.html#section-4.2.3) introduced with TLS 1.3. With that, one can create a finer-grained signature algorithm policy; independently specifying acceptable algorithms for signatures in X.509 certificates and in the TLS protocol itself (i.e. in the `CertificateVerify` message).

To communicate the peer's restrictions to the application, `CredentialsManager::find_cert_chain()` (and friends) will now additionally get a vector of `AlgorithmIdentifier`s specifying which algorithms are acceptable.

Note that also the TLS 1.2 implementation can process this new extension (as suggested by RFC 8446).

## New policy: `tls_acceptable_certificate_signature_schemes()`

By default, this returns `std::nullopt` meaning that the restrictions from the existing `::tls_acceptable_signature_schemes()` should also apply for signatures in certificates. If anything other than `std::nullopt` is returned, clients will add an explicit `signature_algorithm_cert` extension to the Client Hello; servers likewise in the `CertificateRequest` message.

### DIFF
--- a/src/lib/tls/tls_signature_scheme.cpp
+++ b/src/lib/tls/tls_signature_scheme.cpp
@@ -11,6 +11,9 @@
 #include <botan/tls_exceptn.h>
 #include <botan/tls_version.h>
 #include <botan/internal/stl_util.h>
+#include <botan/hash.h>
+#include <botan/der_enc.h>
+#include <botan/internal/emsa.h>
 
 namespace Botan::TLS {
 
@@ -208,7 +211,7 @@ std::string Signature_Scheme::algorithm_name() const noexcept
       }
    }
 
-AlgorithmIdentifier Signature_Scheme::algorithm_identifier() const noexcept
+AlgorithmIdentifier Signature_Scheme::key_algorithm_identifier() const noexcept
    {
    switch(m_code)
       {
@@ -235,6 +238,14 @@ AlgorithmIdentifier Signature_Scheme::algorithm_identifier() const noexcept
       default:
          return AlgorithmIdentifier();
       }
+   }
+
+AlgorithmIdentifier Signature_Scheme::algorithm_identifier() const noexcept
+   {
+   auto emsa = EMSA::create(padding_string());
+   if(!emsa)
+      { return AlgorithmIdentifier(); }
+   return emsa->config_for_x509(algorithm_name(), hash_function_name());
    }
 
 std::optional<Signature_Format> Signature_Scheme::format() const noexcept
@@ -308,6 +319,14 @@ bool Signature_Scheme::is_suitable_for(const Private_Key &private_key) const noe
       return false;
 
    return true;
+   }
+
+std::vector<AlgorithmIdentifier> to_algorithm_identifiers(const std::vector<Signature_Scheme>& schemes)
+   {
+   std::vector<AlgorithmIdentifier> result;
+   std::transform(schemes.begin(), schemes.end(), std::back_inserter(result),
+                  [](const auto& scheme) { return scheme.algorithm_identifier(); });
+   return result;
    }
 
 }  // Botan::TLS
--- a/src/bogo_shim/bogo_shim.cpp
+++ b/src/bogo_shim/bogo_shim.cpp
@@ -1289,6 +1289,7 @@ class Shim_Credentials final : public Botan::Credentials_Manager
 
       std::vector<Botan::X509_Certificate> cert_chain(
          const std::vector<std::string>& cert_key_types,
+         const std::vector<Botan::AlgorithmIdentifier>& /*cert_signature_schemes*/,
          const std::string& /*type*/,
          const std::string& /*context*/) override
          {
--- a/src/cli/tls_helpers.h
+++ b/src/cli/tls_helpers.h
@@ -117,10 +117,13 @@ class Basic_Credentials_Manager : public Botan::Credentials_Manager
 
       std::vector<Botan::X509_Certificate> find_cert_chain(
          const std::vector<std::string>& algos,
+         const std::vector<Botan::AlgorithmIdentifier>& cert_signature_schemes,
          const std::vector<Botan::X509_DN>& acceptable_cas,
          const std::string& type,
          const std::string& hostname) override
          {
+         BOTAN_UNUSED(cert_signature_schemes);
+
          if(type == "tls-client")
             {
             for(const auto& dn : acceptable_cas)
--- a/src/examples/tls_client.cpp
+++ b/src/examples/tls_client.cpp
@@ -53,6 +53,7 @@ public:
   }
 
   std::vector<Botan::X509_Certificate> cert_chain(const std::vector<std::string> &cert_key_types,
+                                                  const std::vector<Botan::AlgorithmIdentifier> &cert_signature_schemes,
                                                   const std::string &type,
                                                   const std::string &context) override {
     // when using tls client authentication (optional), return
--- a/src/examples/tls_custom_curves_client.cpp
+++ b/src/examples/tls_custom_curves_client.cpp
@@ -66,6 +66,7 @@ public:
   }
 
   std::vector<Botan::X509_Certificate> cert_chain(const std::vector<std::string> &cert_key_types,
+                                                  const std::vector<Botan::AlgorithmIdentifier> &cert_signature_schemes,
                                                   const std::string &type,
                                                   const std::string &context) override {
     // when using tls client authentication (optional), return
--- a/src/examples/tls_custom_curves_server.cpp
+++ b/src/examples/tls_custom_curves_server.cpp
@@ -79,6 +79,7 @@ public:
   }
 
   std::vector<Botan::X509_Certificate> cert_chain(const std::vector<std::string> &cert_key_types,
+                                                  const std::vector<Botan::AlgorithmIdentifier> &cert_signature_schemes,
                                                   const std::string &type,
                                                   const std::string &context) override {
     // return the certificate chain being sent to the tls client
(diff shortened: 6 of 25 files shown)


src/lib/tls/tls_signature_scheme.cpp
```diff
@@ -256,6 +258,76 @@ AlgorithmIdentifier Signature_Scheme::algorithm_identifier() const noexcept
...
+      case DSA_SHA256:
+         return AlgorithmIdentifier("DSA/EMSA1(SHA-256)", AlgorithmIdentifier::USE_EMPTY_PARAM);
+      case DSA_SHA384:
+         return AlgorithmIdentifier("DSA/EMSA1(SHA-384)", AlgorithmIdentifier::USE_EMPTY_PARAM);
+      case DSA_SHA512:
+         return AlgorithmIdentifier("DSA/EMSA1(SHA-512)", AlgorithmIdentifier::USE_EMPTY_PARAM);
+
+      default:
+         return AlgorithmIdentifier();
+      }
```
reneme, line 328: This is somewhat a duplicate of the implementations of `EMSA::config_for_x509()` scattered around the different EMSAs. I'm not quite happy with that, to be honest. Simply instantiating the respective EMSA and then actually using `::config_for_x509()` doesn't work out-of-the-box, as it would currently require to pass a respective private key. I think it would be helpful to abstract both use cases and centralize this somehow.

Anyway: As is, this is fairly inefficient, as the `AlgorithmIdentifier`s are re-generated for each invocation. It would be an interesting exercise to `constexpr` the heck out of this. 🤡

I'm open for better ideas. :)

reneme:
Whoops, that almost broke the code examples. @lieser thanks for the CI job checking their successful compilation. 😅

src/lib/tls/tls_signature_scheme.cpp
randombit: Yeah pretty ugly. Also we recompute the PSS params each time when in practice it will always be SHA-{256,512} :sob: 

Looking at `config_for_x509` ... why the heck does this interface take a `Private_Key`? It could easily get away with a `Public_Key`? (Not that this change would address the issue here, but still kind of odd).

Anyway this falls in the category of definitely not great and pointing to a place where some refactoring time is needed but doesn't IMO need to block the PR merging.

src/lib/tls/tls_signature_scheme.cpp
```diff
@@ -256,6 +258,76 @@ AlgorithmIdentifier Signature_Scheme::algorithm_identifier() const noexcept
...
+
+      // for DSA, ECDSA, GOST parameters "SHALL" be empty
+      case DSA_SHA1:
+         return AlgorithmIdentifier("DSA/EMSA1(SHA-160)", AlgorithmIdentifier::USE_EMPTY_PARAM);
+      case DSA_SHA256:
+         return AlgorithmIdentifier("DSA/EMSA1(SHA-256)", AlgorithmIdentifier::USE_EMPTY_PARAM);
+      case DSA_SHA384:
+         return AlgorithmIdentifier("DSA/EMSA1(SHA-384)", AlgorithmIdentifier::USE_EMPTY_PARAM);
+      case DSA_SHA512:
+         return AlgorithmIdentifier("DSA/EMSA1(SHA-512)", AlgorithmIdentifier::USE_EMPTY_PARAM);
```
randombit, line 324: Hold on a minute - do we still support DSA ciphersuites?

It appears we do?? Or at least there is something left over after #2505 which I had though removed DSA entirely...

A 3.0 release seems like a very opportune time to drop all support for DSA from TLS, if we haven't already ...

randombit: OK it appears the only real usage is in the `Signature_Scheme::Code` - will open a PR

randombit: Removed in #3143

randombit (approved):
Needs a rebase in order to (hopefully) make use of #3142, but otherwise lgtm.

src/lib/tls/tls_signature_scheme.cpp
```diff
@@ -237,6 +240,18 @@ AlgorithmIdentifier Signature_Scheme::algorithm_identifier() const noexcept
...
+   auto emsa = EMSA::create(padding_string());
+   if(!emsa)
+      { return AlgorithmIdentifier(); }
+   auto hash = hash_function_name();
+   if(hash == "SHA-1")
+      { hash = "SHA-160"; }
+
+   return emsa->config_for_x509(algorithm_name(), hash);
+   }
+
```
reneme, line 254: This is now using `EMSA::config_for_x509()` instead of [copy-pasting most of its scattered implementations](https://github.com/randombit/botan/pull/3125#discussion_r1049941524).

src/lib/tls/tls_signature_scheme.cpp
reneme: After merging #3142, this now became: https://github.com/randombit/botan/pull/3125#discussion_r1070925009
Thanks for the help in adapting `config_for_x509()` to make it suitable for this use case!

src/lib/tls/tls_signature_scheme.cpp
reneme: The `config_for_x509()` implementations check that the passed hash function name is equal to the hash function used in the EMSA. This is done by string comparison (e.g. from emsa1.cpp):

```c++
if(cert_hash_name != m_hash->name())
   throw Invalid_Argument(...);
```

The library can deal with different string-identifiers for hash functions (e.g "SHA-1", "SHA1" and "SHA-160"). But `m_hash->name()` usually returns a hard-coded identifier. I tried dealing with that using the "SHA-1" to "SHA-160" mapping above. However, this PR's macOS builds fail, because the software-implementation (sha1.cpp) returns "SHA-160" when invoking `Hash_Function::name()` and the CommonCrypto provider (`CommonCrypto_HashFunction`) that might or might not be enabled on macOS returns "SHA-1". 😨

In my opinion, these string comparisons are a generic footgun that will cause recurring issues. 😞 Especially when they force us to scatter such mappings around the code base. Lets try to centralize the string-identifier mappings in the algorithm implementations? @randombit

This could be done by requiring a static method `::implements(std::string_view algo_name)` in each concrete implementation. Now, that implementations "know" all their string-identifiers, a `virtual bool Hash_Function::is(std::string_view algo_name)` could allow convenient polymorphic checks as in `config_for_x509()`. Even the `Hash_Function::create()` factory could benefit from `::implements()` and stop duplicating the string-identifiers. See [this godbolt for a sketch](https://godbolt.org/z/MGxEfjj3M).

One might even consider to auto-generate the `Hash_Function::create()` code based on self-registering sub-classes. But that might take it too far.

Obviously, getting rid of the ambiguous string-identifiers would be even better, but I doubt the massive public API change would be worth it.

randombit: > However, this PR's macOS builds fail, because the software-implementation (sha1.cpp) returns "SHA-160" when invoking Hash_Function::name() and the CommonCrypto provider (CommonCrypto_HashFunction) that might or might not be enabled on macOS returns "SHA-1"

Can you take a look at #3184?

> In my opinion, these string comparisons are a generic footgun that will cause recurring issues. disappointed Especially when they force us to scatter such mappings around the code base. Lets try to centralize the string-identifier mappings in the algorithm implementations?

They are. It was one of the big design mistakes.

> One might even consider to auto-generate the Hash_Function::create() code based on self-registering sub-classes.

This used to be how it worked, actually. But that approach ran into weird problems with static libraries, and ended up having to be removed (in https://github.com/randombit/botan/pull/279)

> This could be done by requiring a static method ::implements(std::string_view algo_name) in each concrete implementation. Now, that implementations "know" all their string-identifiers, a virtual bool Hash_Function::is(std::string_view algo_name) could allow convenient polymorphic checks as in config_for_x509().

This might work but I wonder how this handles parameterizations such as output length or personalization flags. Quick example

```
auto hash = HashFunction::create("BLAKE2b(384)");
assert!(hash->is("BLAKE2b")); // true?
assert!(hash->is("BLAKE2b(384)")); // true?
```

As far as I can recall the main problematic case for this is SHA-1, largely because we use the idiosyncratic/non-standard "SHA-160" name. Maybe the real fix here is to just remove that entirely as a known name, and relnote it? It's not even a big compatibility issue since "SHA-1" always worked.

> Obviously, getting rid of the ambiguous string-identifiers would be even better, but I doubt the massive public API change would be worth it.

We could consider having an enum approach and then implementing the string method on top of that, but I agree that we cannot really remove the string-based interface at this point since it's very widely baked into how the CLI, the tests, the FFI layer, etc all work.

reneme: > This might work but I wonder how this handles parameterizations such as output length or personalization flags.

Fair point. A static `BLAKE2b::implements("BLAKE2b(384)")` should return `true`. But a specific instance of it probably shouldn't.

> Maybe the real fix here is to just remove that entirely as a known name, and relnote it?

I tend to agree. Given that the string-identifiers are such a footgun its probably a good idea to make them as dumb as possible. As in: Don't allow for ambiguities during construction.

> We could consider having an enum approach [...]

Though, I don't see how parameterization would work well with enums either. `Hash_Types::BLAKE2b_384`? That feels rigid to me.

Would it make sense to use `class AlgorithmIdentifier` for those use cases -- as its name suggests? De facto this is already implemented in the specific algorithm implementations. As a library user I always found it inconvenient/confusing to construct `AlgorithmIdentifier` out of thin air, but that might be fixable by some user-friendly API.

reneme:
Rebased and cleaned up after #3186 was merged. This should be ready for prime time.

PR #3126, opened 2022-12-15
Remove XMSS_WOTS_Addressed_{Public,Private}Key

randombit:
These headers are unused, untested, and actually broken, which caused amalgamation build failures.


randombit:
cc @reneme @securitykernel just in case these headers have some use case that I'm missing

securitykernel:
Not that I know of.

reneme (approved):

PR #3127, opened 2022-12-15
Modify the CI matrix to improve coverage

randombit:
CI turnaround time is becoming unweildly with 38 jobs (and with some, such as the amalgamation builds disabled because of the problem addressed in #3126)

@reneme @securitykernel would welcome your take on this. IMO none of the jobs removed are really critical signals for us with respect to improving testing coverage. In particular the Windows and macOS builds are very slow to start which causes a lot of CI wait time so it is worthwhile to trim those to just those that provide real value.

In particular this removes:

* Most of the `windows-2019` builds except for a shared/x86_64
* Most of the `x86` builds (but those have been disabled already for some time, due to a problem with Argon2)
* For macOS we test shared/static/amalgamation with Clang and just `shared` for GCC. (Again currently GCC macOS build was disabled anyway, apparently due to problems with Boost)
* Removes the clang static library Linux build as I don't think it's going to catch anything not caught by either the GCC static library build or the various other Clang builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,15 +29,22 @@ jobs:
       fail-fast: false
 
       matrix:
-        # basic configuration combinations should run on all target platforms
-        target:    [ static, shared, amalgamation ]
-        arch:      [ x86_64, x86 ]
-        host_os:   [ windows-2022, windows-2019 ]
-
-        exclude:
-          # currently failing
-          # see TODOs here: https://github.com/randombit/botan/pull/3007
-          - arch: x86
+        include:
+          - target: shared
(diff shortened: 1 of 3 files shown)


randombit:
Somehow I've increased the size of the matrix, mostly by re-enabling previously disabled targets ...

reneme:
No objection to reducing the build matrix. I'd wanted to look into the jobs that got disabled due to issues but it slipped off my mind.

randombit:
Changed the PR title since in the end we go from 37 targets to ... 37 targets. But they are different targets now - we do test Windows x86 and macOS GCC again for instance.

PR #3135, opened 2022-12-18
Limit compiler cache sizes in CI to 200MB

randombit:
Github Actions caching system is very poorly thought out, easily my least favorite thing about it.


lieser:
Out of interest, could you elaborate what kind of issue a bigger cache cause for the Github Actions?

PR #3138, opened 2022-12-20
Fix overwriting an attribute in an AttributeContainer

lieser:
If an already added attribute was again added in an `AttributeContainer`, elements were shifted left via move assignment by the `std::remove_if`.
This could result in wrong references to the values of existing attributes.

### DIFF
--- a/src/tests/test_pkcs11_high_level.cpp
+++ b/src/tests/test_pkcs11_high_level.cpp
@@ -489,9 +489,24 @@ Test::Result test_attribute_container()
    attributes.add_binary(AttributeType::Value, bin);
 
    attributes.add_bool(AttributeType::Sensitive, true);
-   attributes.add_numeric(AttributeType::ObjectId, 12);
+   attributes.add_numeric(AttributeType::ObjectId, 10);
+   attributes.add_numeric(AttributeType::Id, 20);
+   attributes.add_numeric(AttributeType::PixelX, 30);
+   // Test that overwriting the existing Id attribute works. The numeric attributes above should not be affected by this.
+   attributes.add_numeric(AttributeType::Id, 21);
+   attributes.add_numeric(AttributeType::PixelY, 40);
 
-   result.test_eq("Five elements in attribute container", attributes.count(), 5);
+   result.test_eq("8 elements in attribute container", attributes.count(), 8);
+
+   const std::vector<Botan::PKCS11::Attribute>& storedAttributes = attributes.attributes();
+   result.test_int_eq("ObjectId type", storedAttributes.at(4).type, AttributeType::ObjectId);
+   result.test_int_eq("ObjectId value", *reinterpret_cast< uint64_t* >(storedAttributes.at(4).pValue), 10);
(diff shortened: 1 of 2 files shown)


src/tests/test_pkcs11_high_level.cpp
```diff
@@ -489,9 +489,23 @@ Test::Result test_attribute_container()
...
 
    attributes.add_bool(AttributeType::Sensitive, true);
-   attributes.add_numeric(AttributeType::ObjectId, 12);
-
-   result.test_eq("Five elements in attribute container", attributes.count(), 5);
+   attributes.add_numeric(AttributeType::ObjectId, 10);
+   attributes.add_numeric(AttributeType::Id, 20);
+   attributes.add_numeric(AttributeType::PixelX, 30);
+   attributes.add_numeric(AttributeType::Id, 21);
+   attributes.add_numeric(AttributeType::PixelY, 40);
```
reneme, line 496: Maybe add a comment giving a brief description what this is supposed to check. I assume, that `AttributeType::Id` is successfully overwritten, no?

reneme (approved):
Just to make sure I understood correctly: The free-standing `std::remove_if()` did actually move-construct some elements in the list to push "to-be-removed" items to the back of the list. Despite the underlying container being a linked-list. Right?

Now, I assume that C++20's `std::list::remove_if` takes advantage of the linked-list data structure and just re-wires the elements as one would expect, leaving references to the "shifted" elements intact.

Under that assumption: looks reasonable to me.

lieser:
> The free-standing std::remove_if() did actually move-construct some elements in the list to push "to-be-removed" items to the back of the list. Despite the underlying container being a linked-list. Right?

Yes

> Now, I assume that C++20's std::list::remove_if takes advantage of the linked-list data structure and just re-wires the elements as one would expect, leaving references to the "shifted" elements intact.

Yes.
But note that at least on https://en.cppreference.com/w/cpp/container/list/remove I do not see it being mentioned as a guarantee. It is however guaranteed for the [`erase`](https://en.cppreference.com/w/cpp/container/list/erase), which hopefully means it will essentially also hold for the `remove`.

Example that shows the difference: https://godbolt.org/z/KecsoM4b1

src/tests/test_pkcs11_high_level.cpp
```diff
@@ -489,9 +489,24 @@ Test::Result test_attribute_container()
    attributes.add_binary(AttributeType::Value, bin);
 
    attributes.add_bool(AttributeType::Sensitive, true);
-   attributes.add_numeric(AttributeType::ObjectId, 12);
-
-   result.test_eq("Five elements in attribute container", attributes.count(), 5);
+   attributes.add_numeric(AttributeType::ObjectId, 10);
+   attributes.add_numeric(AttributeType::Id, 20);
+   attributes.add_numeric(AttributeType::PixelX, 30);
+   // Test that overwriting the existing Id attribute works. The over numeric attributes should not be unaffected by this.
```
reneme, line 495:
```suggestion
   // Test that overwriting the existing Id attribute works. The numeric attributes above should not be affected by this.
```

randombit (approved):

PR #3139, opened 2022-12-20
adding new kind of assertions for release builds usage as in

devnexen:
hinting the compiler to optimise assuming the pre-condition is met.


randombit:
I don't think this is a good approach, because it leads to undefined behavior if the assumption is violated.

PR #3142, opened 2022-12-29
Modify EMSA::config_for_x509 to no longer require a private key

randombit:
In EMSA1 we always now set parameters to empty since that is the convention for DSA, ECDSA, etc. Previously if some other algorithm (eg RSA) was used, the parameters would be set. However the `sig_algo_and_padding_ok` check already prevented one from using EMSA1 padding with RSA, so there is no change to functionality as a result.

### DIFF
--- a/src/lib/pk_pad/emsa.cpp
+++ b/src/lib/pk_pad/emsa.cpp
@@ -35,7 +35,7 @@
 
 namespace Botan {
 
-AlgorithmIdentifier EMSA::config_for_x509(const Private_Key& /*unused*/,
+AlgorithmIdentifier EMSA::config_for_x509(const std::string& /*unused*/,
                                           const std::string& /*unused*/) const
    {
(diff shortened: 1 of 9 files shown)


randombit:
@reneme Ping. Do you think this would help with #3125?

reneme:
> Do you think this would help with https://github.com/randombit/botan/pull/3125?

Yes, I'm pretty sure it will help. Thanks. I'll look deeper into it next week. Frankly, this got swamped by the sessions.

reneme (approved):
This works well for #3125. The only small caveat: The TLS code uses "SHA-1" as a hash function name. And the sanity check in `EMSA::config_for_x509()` compare it to the result of `HashFunction::name()` which comes back as "SHA-160". For now, I just mapped the string accordingly.

Let's look into that in #3125 once this is merged.

PR #3144, opened 2022-12-29
Use an inner TLS_Data_Reader when decoding TLS extensions

randombit:
This prevents an extension from accidentally reading information that is actually part of an adjacent extension. It also allows systematically checking that the extension parser did look at all of the bytes of the extension.

The `size` parameter provided to the extensions is now redundant, but mostly harmless. I think that's worth a cleanup as well, but I do not have any more time this morning, and already this seems a nice improvement.


src/tests/test_tls_rfc8448.cpp
```diff
@@ -987,6 +987,7 @@ class Test_TLS_RFC8448_Client : public Test_TLS_RFC8448
                   "tls_verify_cert_chain",
                   "tls_verify_message"
                   });
+               result.require("certificate exists", ctx->certs_verified().size() > 0);
```
randombit, line 990: This change is somewhat unrelated to the PR but at one point I broke the extension parsing which caused the list to become empty here, which resulted in the test segfaulting instead of failing nicely.

reneme (approved):

reneme:
That's certainly a good iteration, thanks. My idea would be that `TLS_Data_Reader` takes care of that by itself in the future, like:

```C++
this->add(make_extension(reader.get_subreader("Extension", extension_size), extension_code, from, message_type));
// extn_reader.assert_done();
//   maybe throw in ~TLS_Data_Reader if it was not fully consumed
```

Sub-readers could then operate on the same container as their parents and avoid copies.

randombit:
@reneme Right, and your proposed approach seems totally superior to what I did here wrt to copies, and simplicity of usage. I would be completely in favor of something like that. This PR is just picking the extremely low hanging fruit, wrt extension decoding being error prone.

randombit:
Rebased to resolve merge conflicts

PR #3145, opened 2022-12-29
Convert Handshake_Extensions_Type to an enum class

randombit:
Also renames the enum itself to `Extensions_Code` to indicate it is not (given 1.3) handshake-specific, and converts the enum elements to CamelCase.


randombit:
Also given how prevalent extensions are in 1.3, a better name would be `TLS::Extensions_Code` since these are no longer handshake specific. Deferring that rename also.

reneme (approved):

reneme:
Rebased to master after merging TLS 1.3 server stuff.

Issue #3107, opened 2022-12-03
Implement Classic McEliece


securitykernel:
We will implement [NIST PQC contest round 4](https://csrc.nist.gov/News/2022/pqc-candidates-to-be-standardized-and-round-4) KEM submission [Classic McEliece](https://classic.mceliece.org/) during the [BSI](https://www.bsi.bund.de/EN) Project 481.

~Preliminary schedule for the implementation is Nov '23 - Jan '24.~

Updated schedule below due to slight project plan rescheduling.

securitykernel:
Updated schedule for the implementation is Dec '23 - Mar '24.

reneme:
Done: #3883

PR #3147, opened 2022-12-29
Remove the HyMES based McEliece implementation

randombit:
With the expectation of eventually replacing it in 3.x with Classic McEliece #3107


randombit:
Going to leave this in for 3.0 and we'll replace it with #3107 and remove it in 4.0

PR #3148, opened 2022-12-30
Move the emscripten CI build to a Linux target

randombit:


randombit:
@reneme IIRC you initially set this job up; do you see any reason not to move it to Linux given that the 22.04 image seems to have a recent enough Emscripten? It seems to be ~2x faster to build this on Linux vs Mac, and at least on my local Linux machine, ccache is able to cache Emscripten, whereas it apparently doesn't work correctly on Mac.

reneme (approved):
Totally fine with moving this to Linux. IIRC we added this due to an issue a user reported. And I put it on macOS because this was the platform I could test locally on with the least friction. 😨

randombit:
Just to follow up in this, `ccache` on Linux does seem to correctly cache Emscripten compilation, hot cache builds take less than 2 minutes now. (Versus ~ 22 minutes we were seeing on Mac builders, regardless of cache status)

PR #3152, opened 2023-01-01
Rename PointGFp to EC_Point

randombit:
The former name suggests support for GF(2^m) curves either existing or possibly existing in the future. Since such curves are difficult to implement in software and possibly are entirely broken, there doesn't seem to be much call for this.

### DIFF
--- a/src/lib/pubkey/ec_group/ec_point.h
+++ b/src/lib/pubkey/ec_group/ec_point.h
@@ -7,8 +7,8 @@
 * Botan is released under the Simplified BSD License (see license.txt)
 */
 
-#ifndef BOTAN_POINT_GFP_H_
-#define BOTAN_POINT_GFP_H_
+#ifndef BOTAN_EC_POINT_H_
+#define BOTAN_EC_POINT_H_
 
 #include <botan/curve_gfp.h>
 #include <botan/exceptn.h>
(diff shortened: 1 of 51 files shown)


src/lib/pubkey/ec_group/ec_point.h
```diff
@@ -378,17 +378,20 @@ std::pair<BigInt, BigInt> BOTAN_UNSTABLE_API OS2ECP(const uint8_t data[], size_t
...
 template<typename Alloc>
-PointGFp OS2ECP(const std::vector<uint8_t, Alloc>& data, const CurveGFp& curve)
+EC_Point OS2ECP(const std::vector<uint8_t, Alloc>& data, const CurveGFp& curve)
    { return OS2ECP(data.data(), data.size(), curve); }
 
-class PointGFp_Var_Point_Precompute;
+class EC_Point_Var_Point_Precompute;
+
+// The name used for this type in older versions
+typedef EC_Point PointGFp;
```
reneme, line 387: If we want to keep supporting the old identifier, I thike it would make sense to also add a "forwarding header" `point_gfp.h` that simply includes `ec_point.h`. Maybe mark it as deprecated right from the start.

reneme (approved):

PR #3155, opened 2023-01-03
FIX/Update: Apple Clang version map

reneme:
The last update broke the mapping so that `./configure.py` always detected clang 7.0. Now its a more fine-grained detection, FWIW.

### DIFF
--- a/src/build-data/detect_version.cpp
+++ b/src/build-data/detect_version.cpp
@@ -32,16 +32,27 @@
 
    This is a rough map from the XCode Clang to the upstream Clang
    versions. It is not correct, but is sufficient for our purposes.
+
+   Wikipedia has a mapping table from Apple Clang version to underlying
+   LLVM version: https://en.wikipedia.org/wiki/Xcode
    */
 
-   #if __clang__major__ <= 10
-     CLANG 7 0
-   #elif __clang__major == 11
+   #if __clang_major__ == 11 && __clang_minor__ >= 4
      CLANG 9 0
-   #elif __clang__major == 12
-     CLANG 10 0
-   #elif __clang__major >= 13
-     CLANG 12 0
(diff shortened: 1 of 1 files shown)


src/build-data/detect_version.cpp
```diff
@@ -32,16 +32,27 @@
...
    */
 
-   #if __clang__major__ <= 10
-     CLANG 7 0
-   #elif __clang__major == 11
+   #if __clang_major__ == 11 && __clang_minor__ >= 4
      CLANG 9 0
-   #elif __clang__major == 12
-     CLANG 10 0
-   #elif __clang__major >= 13
```
reneme, line 43: Note the typos in the macro name:

```
__clang__major
__clang__major__
__clang_major__     <= this is correct
```

src/build-data/detect_version.cpp
```diff
@@ -32,16 +32,27 @@
...
+         CLANG 11 0
+      #endif
+   #elif __clang_major__ == 13
+      #if __clang_minor__ < 3
+         CLANG 12 0
+      #else
+         CLANG 13 0
+      #endif
+   #elif __clang_major__ >= 14
+     CLANG 14 0
```
reneme, line 55: Technically we could prune this a lot more. Clang 9 started having `-std=c++20` but it'll likely choke as we're starting to actually use more and more C++20 things.

randombit: Given that our stated minimum version is Clang 14 and we are only testing in CI with Clang 14, I suspect we will over time become incompatible even with Clang 13. That said having the `configure.py` output map the XCode version (approximately) correctly may aid in assisting users trying to compile with old versions of XCode.

randombit (approved):

PR #3156, opened 2023-01-03
Use sccache 0.4.0-pre4

randombit:
No idea if these script modifications will work :sweat_smile:


randombit:
I give up. Will wait for sccache 0.4.0 to be released and packaged

PR #3158, opened 2023-01-04
starting to remove handmade thread facilities in favor of their C++20…

devnexen:
… counterparts.


reneme (approved):
Thanks!

src/lib/filters/threaded_fork.cpp
```diff
@@ -22,7 +22,7 @@ struct Threaded_Fork_Data
    * Semaphore for indicating that there is work to be done (or to
    * quit)
    */
-   Semaphore m_input_ready_semaphore;
+   std::counting_semaphore<256> m_input_ready_semaphore{0};
```
randombit, line 25: What happens if the `Threaded_Fork` has more than 256 filters attached to it?

PR #3159, opened 2023-01-05
Remove pointless addition in the script-generated Montgomery redc

randombit:
This read was pointless but also harmless because the calling requirements for the redc functions call for `2*p_words + 2` words be allocated, even though z being at most `p*p` will be at most `2*p_words`, leaving the higher words as always zero.

Also in both the script generated and generic redc code, just copy out p_words words from the result, instead of p_words+1, because - assuming the result really was reduced - we know that the highest word must have been zero.


randombit:
There is a lot more pretty sketchy stuff in the MP<->BigInt layer that needs cleanup. A lot of this code is very old.

Issue #3163, opened 2023-01-07
Always scaling GCM update granularity causes issues with non-ffi use


novasharper:
Since the block size is 16 bytes, the size of the (un)encrypted data is not guaranteed to be a multiple of 64 bytes. For example, the size of the encrypted data in a password-protected ssh private key is only guaranteed to be a multiple of 16 bytes.

novasharper:
Looking at how `botan_cipher_update` does
```c++
const size_t ud = cipher.update_granularity();
BOTAN_ASSERT(cipher.update_granularity() > cipher.minimum_final_size(), "logic error");
mbuf.resize(ud);
```
would it make sense to instead do
```c++
const size_t ud = cipher.update_granularity();
size_t buf_size = std::max<size_T>(cipher.update_granularity(), cipher.minimum_final_size());
mbuf.resize(buf_size);
```

randombit:
@novasharper Do you think #3168 resolves this for you?

Another question is, assuming #3168 resolves things, are you in desperate need of this to be backported to the 2.x release branches or are you willing/able to wait a few months for 3.0 to be released?

novasharper:
I need to try building against 3.0, but #3168  should resolve this. My need for this is not urgent. I can just use an alternate cipher until 3.0 is released. I had just noticed the issue and figured that I would bring it up and try to create a fix for it.

PR #3164, opened 2023-01-08
Allow disabling FFI compat features for Cipher_Mode

novasharper:
If `ffi_compat=false` is passed to `::create_or_throw` or `::create`, then create the cipher mode object with ffi compatability fixes disabled (if there are any).

For example, in GCM_Mode, the the size of the update buffer is required to be a multiple of 64 bytes (block size * scaling factor, default 4) instead of the more usual default of 16 bytes. This can cause issues when trying to decrypt content that was not encrypted with botan.

GH #3163


randombit:
I agree #3163 is an issue that we should resolve (and happily with 3.0 still not yet released we make changes here, even incompatible ones) but I don't think this is the right approach. I will follow up in #3163 to discuss.

randombit:
Closing in favor of #3168

PR #3165, opened 2023-01-09
Introduce a strong-type template

reneme:
Usage:

```C++
// declare a strong type (`struct SessionID_` is needed to make this
// a unique type among all the other byte vector based types)
using SessionID = Strong<std::vector<uint8_t>, struct SessionID_>

SessionID sid{unlock(rng.random_vec(32))};
// I'm hoping this will eventually be:
// auto sid = rng.random_range<SessionID>(32);

// Strong Types try hard to act like their underlying type without
// doing implicit conversion between each other
std::cout << hex_encode(sid); // assumg `hex_encode` has a `std::span<>` overload
```

With that, constructors that take two byte vectors are much more ergonomic. No more mixing up parameters. Overloading constructors (despite both overloads being byte vectors):

```C++
using SessionID = Strong<std::vector<uint8_t>, struct SessionID_>
using SessionTicket = Strong<std::vector<uint8_t>, struct SessionTicket_>

// A Session_Handle must be constructed with an ID, a Ticket or both
// but it cannot be constructed without either of them.
struct Session_Handle {
  Session_Handle(SessionID id);
  Session_Handle(SessionTicket ticket);
  Session_Handle(SessionID id, SessionTicket ticket);
};
```

### DIFF
--- a/src/lib/utils/concepts.h
+++ b/src/lib/utils/concepts.h
@@ -1,0 +1,100 @@
+/**
+ * Useful concepts that are available throughout the library
+ * (C) 2023 Jack Lloyd
+ *     2023 René Meusel - Rohde & Schwarz Cybersecurity
+ *
+ * Botan is released under the Simplified BSD License (see license.txt)
+ */
+
+#ifndef BOTAN_CONCEPTS_H_
+#define BOTAN_CONCEPTS_H_
+
+#include <compare>
+#include <concepts>
+#include <type_traits>
+
+namespace Botan::concepts {
+
+// TODO: C++20 use std::convertible_to<> that was not available in Android NDK
+//       as of this writing. Tested with API Level up to 33.
+template <class FromT, class ToT>
+concept convertible_to =
+  std::is_convertible_v<FromT, ToT> &&
+  requires {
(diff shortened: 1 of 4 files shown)


reneme:
Ugh... I already tried to work around that. Apparently without much luck. I'll look into it once more. So far 2019 did keep better pace than other compilers. But I agree, it's likely that it'll become painful.

reneme:
>  I'll look into it once more.

I found another (IMO more elegant) way to get 2019 to do what I want. For now, it may stay. ;)

randombit:
OK. Should we in the [https://github.com/randombit/botan/blob/master/doc/support.rst#supported-platforms](supported platform list) for 3.0 say instead of "Visual C++ 2019 or higher" something like "Visual C++ 2022, with VC 2019 supported on an interim basis"?

Toolchains tend to be sticky in projects (more so on the Windows side of things, in my experience) and I don't want to bait and switch someone, where they start a project using Botan 3 plus VC 2019, and everything works great until without warning in some minor release we drop VC2019 support because it got inconvenient.

randombit (approved):
Fine to merge, we can sort out the VC2019 support thing in a ticket.

PR #3168, opened 2023-01-10
Add Cipher_Mode::ideal_granularity

randombit:
This reflects the size that would be best for performance, while update_granularity is modified to return the true minimum value.

GH #3163

### DIFF
--- a/src/lib/ffi/ffi_cipher.cpp
+++ b/src/lib/ffi/ffi_cipher.cpp
@@ -30,17 +30,29 @@ namespace {
 size_t ffi_choose_update_size(Botan::Cipher_Mode& mode)
    {
    const size_t update_granularity = mode.update_granularity();
-   const size_t ideal_granularity = mode.ideal_granularity();
    const size_t minimum_final_size = mode.minimum_final_size();
 
-   if(minimum_final_size == 0)
+   /*
+   * Return the minimum possible granularity given the FFI API constraints that
+   * we require the returned size be > minimum final size.
+   *
+   * If the minimum final size is zero, or the update_granularity is
+   * already greater, just use that.
+   *
+   * Otherwise scale the update_granularity to a sufficient size
+   * to be greater than the minimum.
+   */
+   if(minimum_final_size == 0 || update_granularity > minimum_final_size)
       {
       BOTAN_ASSERT_NOMSG(update_granularity > 0);
       return update_granularity;
       }
 
-   BOTAN_ASSERT_NOMSG(ideal_granularity > minimum_final_size);
-   return ideal_granularity;
+   size_t buf_size = std::max(update_granularity, minimum_final_size + 1);
+   if(buf_size % update_granularity != 0)
+      buf_size += update_granularity - (buf_size % update_granularity);
+
+   return buf_size;
    }
 
 }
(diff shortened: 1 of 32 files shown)


src/lib/ffi/ffi_cipher.cpp
```diff
@@ -217,7 +244,12 @@ int botan_cipher_get_default_nonce_length(botan_cipher_t cipher, size_t* nl)
 
 int botan_cipher_get_update_granularity(botan_cipher_t cipher, size_t* ug)
    {
-   return BOTAN_FFI_VISIT(cipher, [=](const auto& c) { *ug = c.update_granularity(); });
+   return BOTAN_FFI_VISIT(cipher, [=](const auto& c) { *ug = c.ideal_granularity(); });
```
novasharper, line 247: Should this still be `c.update_granularity()`?

src/tests/test_aead.cpp
```diff
@@ -452,8 +452,20 @@ class AEAD_Tests final : public Text_Based_Test
...
+                        enc->update_granularity(), 0);
+
+         result.test_eq("enc and dec ideal granularity is the same",
+                        enc->ideal_granularity(), dec->ideal_granularity());
+
+         result.test_gt("ideal granularity is at least update granularity",
+                        enc->ideal_granularity(), enc->update_granularity());
+
+         result.confirm("ideal granularity is a multiple of update granularity",
+                        enc->ideal_granularity() % enc->ideal_granularity() == 0);
```
novasharper, line 468: This should be
```c++
enc->ideal_granularity() % enc->update_granularity() == 0
```

src/tests/test_modes.cpp
```diff
@@ -64,6 +64,18 @@ class Cipher_Mode_Tests final : public Text_Based_Test
...
+                           enc->update_granularity(), 0);
+
+            result.test_eq("enc and dec ideal granularity is the same",
+                           enc->ideal_granularity(), dec->ideal_granularity());
+
+            result.test_gt("ideal granularity is at least update granularity",
+                           enc->ideal_granularity(), enc->update_granularity());
+
+            result.confirm("ideal granularity is a multiple of update granularity",
+                           enc->ideal_granularity() % enc->ideal_granularity() == 0);
```
novasharper, line 77: This should be
```c++
enc->ideal_granularity() % enc->update_granularity() == 0
```

src/tests/test_aead.cpp
randombit: Nice catch thank you

src/tests/test_modes.cpp
randombit: Fixed

src/lib/ffi/ffi_cipher.cpp
randombit: No because that "idea" granularity is still the required update size on the FFI side of things. If we adopt your suggestion to use the smallest possible multiple from #3163 (which seems like a good idea, but I'll do it in a followup MR) then this would instead return that value.

randombit: I applied your suggestion from #3163 here and modified this function to return that value instead.

PR #3170, opened 2023-01-12
Follow up: Sanity check for Shake_Cipher base class

reneme:
I merged #3167 prematurely, omitting the sanity check in `Shake_Cipher`. See: https://github.com/randombit/botan/pull/3167#discussion_r1065815364
Sorry for that.


randombit (approved):
:+1:

PR #3178, opened 2023-01-14
Support encoding/decoding Kyber keys through the generic interfaces

randombit:


reneme (approved):
I wasn't aware of the `PK_Key_Generation_Test` helper. That's good to know also for the upcoming algorithms in BSI Project 481. Thanks for giving this some TLC.

PR #3184, opened 2023-01-16
Have CommonCrypto hash impl always return a consistent name.

randombit:
Previously it's name() would return the name it was requested by instead of the canonical name.


reneme (approved):
Pretty sure that this will mitigate the issue pointed out in https://github.com/randombit/botan/pull/3125#discussion_r1071037931. Thanks.

randombit:
Closing in favor of #3186

PR #3186, opened 2023-01-17
Only allow "SHA-1" as a valid name for SHA-1

randombit:
Removing SHA-160 and SHA1 aliases


reneme (approved):
Thanks. I'll rebase #3125 onto this to see if it fixes macOS.

reneme:
> I'll rebase https://github.com/randombit/botan/pull/3125 onto this to see if it fixes macOS.

It works!

PR #3188, opened 2023-01-17
concept using directly std::convertible_to instead.

devnexen:


devnexen:
let's revisit it as next ndk update.

reneme:
First of all: thanks for looking into that.

> let's revisit it as next ndk update.

There are more things blocked by NDK (and XCode Clang) that forced me to add (potentially half-baked) concepts in `concepts.h` which actually have better alternatives in the C++20 STL. Can't wait! 😉

PR #2973, opened 2022-05-05
Dilithium pqc

omarama:
Dilithium PQC NIST Round 3 implementation

Co-authored-by: Michael Boric <Michael.Boric@rohde-schwarz.com>

PR #3189, opened 2023-01-17
Implement write_keystream for CTR and SHAKE

randombit:
Also fix a bug in the SHAKE cipher

### DIFF
--- a/src/cli/speed.cpp
+++ b/src/cli/speed.cpp
@@ -963,10 +963,20 @@ class Speed final : public Command
             while(encrypt_timer->under(runtime))
                {
                encrypt_timer->run([&]() { cipher.encipher(buffer); });
-               //encrypt_timer->run([&]() { cipher.write_keystream(buffer.data(), buffer.size()); });
                }
 
             record_result(encrypt_timer);
+
+            if(verbose())
+               {
+               auto ks_timer = make_timer(cipher.name(), buffer.size(), "write_keystream", provider, buf_size);
+
+               while(ks_timer->under(runtime))
+                  {
+                  ks_timer->run([&]() { cipher.write_keystream(buffer.data(), buffer.size()); });
+                  }
+               record_result(ks_timer);
+               }
             }
          }
 #endif
--- a/src/cli/speed.cpp
+++ b/src/cli/speed.cpp
@@ -963,6 +963,7 @@ class Speed final : public Command
             while(encrypt_timer->under(runtime))
                {
                encrypt_timer->run([&]() { cipher.encipher(buffer); });
+               //encrypt_timer->run([&]() { cipher.write_keystream(buffer.data(), buffer.size()); });
                }
 
             record_result(encrypt_timer);
--- a/src/lib/stream/ctr/ctr.cpp
+++ b/src/lib/stream/ctr/ctr.cpp
@@ -126,40 +126,32 @@ void CTR_BE::write_keystream(uint8_t out[], size_t length)
    {
    verify_key_set(m_iv.empty() == false);
 
-   const uint8_t* pad_bits = &m_pad[0];
-   const size_t pad_size = m_pad.size();
+   const size_t avail = m_pad.size() - m_pad_pos;
+   const size_t take = std::min(length, avail);
+   copy_mem(out, &m_pad[m_pad_pos], take);
+   length -= take;
+   out += take;
+   m_pad_pos += take;
 
-   if(m_pad_pos > 0)
+   while(length >= m_pad.size())
       {
-      const size_t avail = pad_size - m_pad_pos;
-      const size_t take = std::min(length, avail);
-      copy_mem(out, pad_bits + m_pad_pos, take);
-      length -= take;
-      out += take;
-      m_pad_pos += take;
+      add_counter(m_ctr_blocks);
+      m_cipher->encrypt_n(m_counter.data(), out, m_ctr_blocks);
 
-      if(take == avail)
-         {
-         add_counter(m_ctr_blocks);
-         m_cipher->encrypt_n(m_counter.data(), m_pad.data(), m_ctr_blocks);
-         m_pad_pos = 0;
-         }
+      length -= m_pad.size();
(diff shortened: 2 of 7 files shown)


src/lib/stream/ctr/ctr.cpp
```diff
@@ -122,6 +122,44 @@ void CTR_BE::cipher(const uint8_t in[], uint8_t out[], size_t length)
...
+      out += pad_size;
+
+      add_counter(m_ctr_blocks);
+      m_cipher->encrypt_n(m_counter.data(), m_pad.data(), m_ctr_blocks);
+      }
+
+   copy_mem(out, pad_bits, length);
+   m_pad_pos += length;
+   }
+
```
reneme, line 162: Maybe worthwhile to avoid the code duplication with `CTR_BE::cipher()`? Like so: https://github.com/reneme/botan/commit/3725095adeede3a83477d5d48c76e23c72e0c65c

I'd bet that the compiler inlines that and (hopefully) makes it a zero-cost abstraction.

src/lib/stream/shake_cipher/shake_cipher.cpp
```diff
@@ -62,6 +62,44 @@ void SHAKE_Cipher::cipher(const uint8_t in[], uint8_t out[], size_t length)
...
+      length -= m_shake_rate;
+      out += m_shake_rate;
+      }
+
+   copy_out_le(m_buffer.data(), m_shake_rate, m_state.data());
+
+   copy_mem(out, &m_buffer[0], length);
+   m_buf_pos += length;
+   }
+
```
reneme, line 102: Potentially, a similar refactoring as for CTR could be used, didn't look into that so far.

src/lib/stream/shake_cipher/shake_cipher.cpp
```diff
@@ -72,6 +110,7 @@ void SHAKE_Cipher::key_schedule(const uint8_t key[], size_t length)
    const size_t S_pos = SHA_3::absorb(SHAKE_BITRATE, m_state, 0, key, length);
    SHA_3::finish(SHAKE_BITRATE, m_state, S_pos, 0x1F, 0x80);
    copy_out_le(m_buffer.data(), m_buffer.size(), m_state.data());
+   m_buf_pos = 0;
```
reneme, line 113: I suppose that's the bug that you fixed along the way? Worth back-porting?

randombit: Yes this needs to be backported, bug found by the new tests

src/cli/speed.cpp
```diff
@@ -963,6 +963,7 @@ class Speed final : public Command
             while(encrypt_timer->under(runtime))
                {
                encrypt_timer->run([&]() { cipher.encipher(buffer); });
+               //encrypt_timer->run([&]() { cipher.write_keystream(buffer.data(), buffer.size()); });
```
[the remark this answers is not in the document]
randombit: It's nice to be able to see the difference (if any) in performance. I changed this so we test both.

src/lib/stream/stream_cipher.h
```diff
@@ -60,6 +60,9 @@ class BOTAN_PUBLIC_API(2,0) StreamCipher : public SymmetricAlgorithm
 
       /**
       * Write keystream bytes to a buffer
+      *
+      * The contents of out are ignored/overwritten
```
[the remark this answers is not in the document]
randombit: :+1:

src/lib/stream/ctr/ctr.cpp
randombit: You're right this would work *but* I don't want to do it because it eliminates the possibility of a nice optimization which is missing here in CTR but was already done for SHAKE in 5b5baed264eb1e371cb69983b0e2585454b3b52a

For encrypt the data flow is

buffer <- process(internal_state)
output <- output xor buffer

But for keystream we have:

buffer <- process(internal_state)
output <- buffer

but the buffer isn't doing anything for us. So for SHAKE in the main loop what we do is instead is

output <- process(internal_state)

skipping a memory copy.

I added a todo for CTR mentioning we should do the same. It actually will make a much bigger difference for CTR, because for SHAKE it seems to save ~~ .22 cycles per byte. But CTR + AES-128 (with AES-NI) runs already at about 1.25 cycles per byte - bringing that down to ~1 cpb would make for a large performance improvement.

src/lib/stream/shake_cipher/shake_cipher.cpp
randombit: See other comment

src/lib/stream/ctr/ctr.cpp
reneme: Mhm, I wasn't happy about this additional copy as well. Fair enough.

reneme (approved):

reneme:
As a side-note: The refactorings in #2973 re:XOF suggest that it might be worthwhile to add a XOF interface to the library. The stream cipher approach works, but `XOF::generate_bytes(size_t)` would feel more natural than `Stream_Cipher::write_keystream()`.

src/lib/stream/ctr/ctr.cpp
randombit: Added a somewhat faster `write_keystream` for CTR as described above in 5aaf48e313bd173e6f6704eaa298fccb923362e2, now down to .78 cpb with large buffers.

randombit:
XOF may be tricky in the long run because we need SHAKE and AES/CTR, since that's what pretty much all of the PQC algorithms use, but I suspect that many of the schemes instantiate the AES/CTR XOF in slightly different ways. It's probably better to wait until we have more experience with what the different PQC algorithms do before committing to something.

src/lib/stream/ctr/ctr.cpp
reneme: Agreed. Trying to incorporate that into my proposed abstraction will result in too much added complexity to save a duplicated lines.

PR #3194, opened 2023-01-19
[TLS 1.3] tls_reader benefiting std span usage

devnexen:


src/lib/tls/tls_reader.h
```diff
@@ -24,8 +25,8 @@ namespace TLS {
 class TLS_Data_Reader final
    {
    public:
-      TLS_Data_Reader(const char* type, const std::vector<uint8_t>& buf_in) :
-         m_typename(type), m_buf(buf_in), m_offset(0) {}
+      TLS_Data_Reader(const char* type, const std::span<const uint8_t>& buf_in) :
```
reneme, line 28: `std::span<>` should always be taken by value instead of `const&`.

src/lib/tls/tls13/tls_transcript_hash_13.cpp
```diff
@@ -169,8 +167,7 @@ void Transcript_Hash_State::update(std::span<const uint8_t> serialized_message_s
       if(serialized_message_length > 0 && *serialized_message == CLIENT_HELLO)
          {
          truncation_mark =
-            find_client_hello_truncation_mark({serialized_message,
-                                               serialized_message + serialized_message_length});
+            find_client_hello_truncation_mark(serialized_message_s);
```
reneme, line 170: I noticed that you're using the `_s` suffix for `std::span<>`. Is that somewhat idiomatic? I don't mind it, but I also wouldn't want to enforce it.

reneme (commented):
Yeah! That's definitely a good point to use `std::span<>`. It was high up on my list as well.

src/lib/tls/tls_reader.h
```diff
@@ -24,8 +25,8 @@ namespace TLS {
...
+         m_typename(type), m_buf(buf_in), m_size(buf_in.end() - buf_in.begin()), m_offset(0) {}
```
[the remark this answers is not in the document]
devnexen: nah it s my fault, I did not see there was a size member in the C++ spec..

reneme (approved):

Issue #752, opened 2016-12-02
Guideline Support Library


neusdan:
What do you think about using the [Guideline Support Library](https://github.com/Microsoft/GSL)?
It's small, header only and seems to have a good platform support. Would probably the best to ship it with Botan. I especially like the idea to replace all pointer arithmetic with `span`.

randombit:
I hadn't seen this library before. `span` concept does seem very useful, I'm not so sure about the rest. I will try it out soon and think it over.

webmaster128:
If you're a video guy you might be interested in the GSL introduction on CppCon 2015 for the overall idea and the motivation

https://www.youtube.com/watch?v=1OEu9C51K2A&list=PLHTh1InhhwT75gykhs7pqcR_uSiG601oh&index=1
https://www.youtube.com/watch?v=hEx5DNLWGgA&index=2&list=PLHTh1InhhwT75gykhs7pqcR_uSiG601oh

randombit:
Closing this - now with C++20 on master we can use `std::span` which seems like the main attractor for GSL

PR #3195, opened 2023-01-19
std::span (and more) for Botan::RandomNumberGenerator

reneme:
### Open Things to Do in this Pull Request

* [ ] **Deprecate `random_vec(std::span)`?**
       This replaced `template<typename AllocT> random_vec(std::vector<uint8_t, AllocT>)`, so we can't just get rid of it. But it is equivalent to `randomize(std::span)` which has the better fitting name.
* [ ] **Deprecate `template<typename T> random_vec(size_t)` (and friends)?**
       ... and rename it simply to `template<typename T> random(size_t)` (but that may be just a nit-pick).
* [x] **Stateful_RNG::randomize_with_ts_input()**
       Its implementation deviates from the base-class implementation (presumably to be more conservative with stateful RNGs). We [have to find a way to incorporate that](https://github.com/randombit/botan/pull/3195#issuecomment-1449602320) into the new sub-class API.

### Description

This adapts the RNG's internal API to `std::span<>` so that subclasses do not need to provide or use ptr-length style method overloads. The class hierarchy's heavy lifting (i.e generating random bits and ingesting entropy bits) is unified into a single method `fill_bytes_with_input()`. All sub-classes need to implement this interface starting with Botan 3.0. Details [here](https://github.com/randombit/botan/pull/3195#issuecomment-1397813174).

For most sub-classes this change applies nicely. `Stateful_RNG` needed some extra thought (and special handling) to ingest initial entropy (via `add_entropy()`) into its concrete implementations (`ChaCha_RNG` and `HMAC_DRBG`) in an implementation-agnostic manner. Along with that, an [outstanding API inconsistency was fixed](https://github.com/randombit/botan/pull/3195#discussion_r1122935125) and nailed down with a test.

---

### Old (rejected) Idea Description

This generalizes the API of `Botan::RandomNumberGenerator` to work with arbitrary contiguous byte-containers beyond just `std::vector<>` or C-style pointer-length crutches. Judging from the massive internal usage of the adapted APIs this seems to be  backward-compatible change (especially regarding `::random_vec()`).

Particularly, `::random_vec()` should be useful within the library as well. Such as in #3150:

```C++
using Session_ID = Botan::Strong<std::vector<uint8_t>, struct Session_ID_>;

// generate a random TLS Session ID
const auto random_id = rng.random_vec<Session_ID>(32);
```

I opted to keep the old pointer-length APIs and add overloads using `std::span<>` for backward compatibility. Albeit convenient for legacy software, this forces sub-classes of `RandomNumberGenerator` to actively make those overloads available, when they override (say) `randomize(uint8_t*, size_t)` (see [the adaptions to `test_rng.h`](https://github.com/randombit/botan/compare/feature/random_something?expand=1#diff-ebeec3911bd550c2748b6ee38865e4762dabc699da9832ae0b4eee9bc7246144)).

Hence, I'd like to make the argument to ditch the C-style APIs in favor of fully embracing `std::span<>`. Doing this consistently across the API is a _huge_ break in compatibility. But I think it's worth considering.

For reference #752.

### DIFF
--- a/src/lib/rng/auto_rng/auto_rng.cpp
+++ b/src/lib/rng/auto_rng/auto_rng.cpp
@@ -7,6 +7,10 @@
 #include <botan/auto_rng.h>
 #include <botan/entropy_src.h>
 #include <botan/hmac_drbg.h>
+#include <botan/internal/loadstor.h>
+#include <botan/internal/os_utils.h>
+
+#include <array>
 
 #if defined(BOTAN_HAS_SYSTEM_RNG)
   #include <botan/system_rng.h>
@@ -83,11 +87,6 @@ std::string AutoSeeded_RNG::name() const
    return m_rng->name();
    }
 
-void AutoSeeded_RNG::add_entropy(const uint8_t in[], size_t len)
-   {
-   m_rng->add_entropy(in, len);
-   }
-
 size_t AutoSeeded_RNG::reseed(Entropy_Sources& srcs,
                               size_t poll_bits,
                               std::chrono::milliseconds poll_timeout)
@@ -95,15 +94,17 @@ size_t AutoSeeded_RNG::reseed(Entropy_Sources& srcs,
    return m_rng->reseed(srcs, poll_bits, poll_timeout);
    }
 
-void AutoSeeded_RNG::randomize(uint8_t output[], size_t output_len)
+void AutoSeeded_RNG::fill_bytes_with_input(std::span<uint8_t> out, std::span<const uint8_t> in)
    {
-   m_rng->randomize_with_ts_input(output, output_len);
-   }
+   std::array<uint8_t, 16> additional_input;
+   if(in.empty() && m_rng->accepts_input())
+      {
+      store_le(OS::get_system_timestamp_ns(), additional_input.data());
+      store_le(OS::get_high_resolution_clock(), additional_input.data() + 8);
+      in = additional_input;
+      }
 
-void AutoSeeded_RNG::randomize_with_input(uint8_t output[], size_t output_len,
-                                          const uint8_t ad[], size_t ad_len)
-   {
-   m_rng->randomize_with_input(output, output_len, ad, ad_len);
+   m_rng->randomize_with_input(out, in);
    }
 
 }
--- a/src/lib/rng/auto_rng/auto_rng.cpp
+++ b/src/lib/rng/auto_rng/auto_rng.cpp
@@ -96,15 +96,10 @@ size_t AutoSeeded_RNG::reseed(Entropy_Sources& srcs,
 
 void AutoSeeded_RNG::fill_bytes_with_input(std::span<uint8_t> out, std::span<const uint8_t> in)
    {
-   std::array<uint8_t, 16> additional_input;
-   if(in.empty() && m_rng->accepts_input())
-      {
-      store_le(OS::get_system_timestamp_ns(), additional_input.data());
-      store_le(OS::get_high_resolution_clock(), additional_input.data() + 8);
-      in = additional_input;
-      }
-
-   m_rng->randomize_with_input(out, in);
+   if(in.empty())
+      { m_rng->randomize_with_ts_input(out); }
+   else
+      { m_rng->randomize_with_input(out, in); }
    }
 
 }
--- a/src/lib/rng/chacha_rng/chacha_rng.cpp
+++ b/src/lib/rng/chacha_rng/chacha_rng.cpp
@@ -58,20 +58,21 @@ void ChaCha_RNG::clear_state()
    m_chacha->set_key(m_hmac->final());
    }
 
-void ChaCha_RNG::generate_output(uint8_t output[], size_t output_len,
-                                 const uint8_t input[], size_t input_len)
+void ChaCha_RNG::generate_output(std::span<uint8_t> output, std::span<const uint8_t> input)
    {
-   if(input_len > 0)
+   BOTAN_ASSERT_NOMSG(!output.empty());
+
+   if(!input.empty())
       {
-      update(input, input_len);
+      update(input);
       }
 
-   m_chacha->write_keystream(output, output_len);
+   m_chacha->write_keystream(output.data(), output.size());
    }
 
-void ChaCha_RNG::update(const uint8_t input[], size_t input_len)
+void ChaCha_RNG::update(std::span<const uint8_t> input)
    {
-   m_hmac->update(input, input_len);
+   m_hmac->update(input.data(), input.size()); // TODO: fix after GH #3294
    m_chacha->set_key(m_hmac->final());
 
    secure_vector<uint8_t> mac_key(m_hmac->output_length());
--- a/src/lib/rng/chacha_rng/chacha_rng.cpp
+++ b/src/lib/rng/chacha_rng/chacha_rng.cpp
@@ -16,12 +16,12 @@ ChaCha_RNG::ChaCha_RNG() : Stateful_RNG()
    clear();
    }
 
-ChaCha_RNG::ChaCha_RNG(const secure_vector<uint8_t>& seed) : Stateful_RNG()
+ChaCha_RNG::ChaCha_RNG(std::span<const uint8_t> seed) : Stateful_RNG()
    {
    m_hmac = MessageAuthenticationCode::create_or_throw("HMAC(SHA-256)");
    m_chacha = StreamCipher::create_or_throw("ChaCha(20)");
    clear();
-   add_entropy(seed.data(), seed.size());
+   add_entropy(seed);
    }
 
 ChaCha_RNG::ChaCha_RNG(RandomNumberGenerator& underlying_rng,
@@ -67,16 +67,14 @@ void ChaCha_RNG::generate_output(std::span<uint8_t> output, std::span<const uint
       update(input);
       }
 
-   m_chacha->write_keystream(output.data(), output.size());
+   m_chacha->write_keystream(output);
    }
 
 void ChaCha_RNG::update(std::span<const uint8_t> input)
    {
-   m_hmac->update(input.data(), input.size()); // TODO: fix after GH #3294
+   m_hmac->update(input);
    m_chacha->set_key(m_hmac->final());
-
-   secure_vector<uint8_t> mac_key(m_hmac->output_length());
-   m_chacha->write_keystream(mac_key.data(), mac_key.size());
+   const auto mac_key = m_chacha->keystream_bytes(m_hmac->output_length());
    m_hmac->set_key(mac_key);
    }
 
--- a/src/lib/rng/hmac_drbg/hmac_drbg.cpp
+++ b/src/lib/rng/hmac_drbg/hmac_drbg.cpp
@@ -139,49 +139,52 @@ std::string HMAC_DRBG::name() const
 * HMAC_DRBG generation
 * See NIST SP800-90A section 10.1.2.5
 */
-void HMAC_DRBG::generate_output(uint8_t output[], size_t output_len,
-                                const uint8_t input[], size_t input_len)
+void HMAC_DRBG::generate_output(std::span<uint8_t> output, std::span<const uint8_t> input)
    {
-   if(input_len > 0)
+   BOTAN_ASSERT_NOMSG(!output.empty());
+
+   if(!input.empty())
       {
-      update(input, input_len);
+      update(input);
       }
 
+   uint8_t* output_ptr = output.data();
+   size_t output_len = output.size();
    while(output_len > 0)
       {
       const size_t to_copy = std::min(output_len, m_V.size());
       m_mac->update(m_V.data(), m_V.size());
       m_mac->final(m_V.data());
-      copy_mem(output, m_V.data(), to_copy);
+      copy_mem(output_ptr, m_V.data(), to_copy);
 
-      output += to_copy;
+      output_ptr += to_copy;
       output_len -= to_copy;
       }
 
-   update(input, input_len);
+   update(input);
    }
 
 /*
 * Reset V and the mac key with new values
 * See NIST SP800-90A section 10.1.2.2
 */
-void HMAC_DRBG::update(const uint8_t input[], size_t input_len)
+void HMAC_DRBG::update(std::span<const uint8_t> input)
    {
    secure_vector<uint8_t> T(m_V.size());
    m_mac->update(m_V);
    m_mac->update(0x00);
-   m_mac->update(input, input_len);
+   m_mac->update(input.data(), input.size()); // TODO: pass span after merging GH #3294
    m_mac->final(T.data());
    m_mac->set_key(T);
 
    m_mac->update(m_V.data(), m_V.size());
    m_mac->final(m_V.data());
 
-   if(input_len > 0)
+   if(!input.empty())
       {
       m_mac->update(m_V);
       m_mac->update(0x01);
-      m_mac->update(input, input_len);
+      m_mac->update(input.data(), input.size()); // TODO: pass span after merging GH #3294
       m_mac->final(T.data());
       m_mac->set_key(T);
 
--- a/src/lib/rng/hmac_drbg/hmac_drbg.cpp
+++ b/src/lib/rng/hmac_drbg/hmac_drbg.cpp
@@ -148,17 +148,14 @@ void HMAC_DRBG::generate_output(std::span<uint8_t> output, std::span<const uint8
       update(input);
       }
 
-   uint8_t* output_ptr = output.data();
-   size_t output_len = output.size();
-   while(output_len > 0)
+   while(!output.empty())
       {
-      const size_t to_copy = std::min(output_len, m_V.size());
-      m_mac->update(m_V.data(), m_V.size());
-      m_mac->final(m_V.data());
-      copy_mem(output_ptr, m_V.data(), to_copy);
+      const size_t to_copy = std::min(output.size(), m_V.size());
+      m_mac->update(m_V);
+      m_mac->final(m_V);
+      copy_mem(output.data(), m_V.data(), to_copy);
 
-      output_ptr += to_copy;
-      output_len -= to_copy;
+      output = output.subspan(to_copy);
       }
 
    update(input);
@@ -173,23 +170,23 @@ void HMAC_DRBG::update(std::span<const uint8_t> input)
    secure_vector<uint8_t> T(m_V.size());
    m_mac->update(m_V);
    m_mac->update(0x00);
-   m_mac->update(input.data(), input.size()); // TODO: pass span after merging GH #3294
-   m_mac->final(T.data());
+   m_mac->update(input);
+   m_mac->final(T);
    m_mac->set_key(T);
 
-   m_mac->update(m_V.data(), m_V.size());
-   m_mac->final(m_V.data());
+   m_mac->update(m_V);
+   m_mac->final(m_V);
 
    if(!input.empty())
       {
       m_mac->update(m_V);
       m_mac->update(0x01);
-      m_mac->update(input.data(), input.size()); // TODO: pass span after merging GH #3294
-      m_mac->final(T.data());
+      m_mac->update(input);
+      m_mac->final(T);
       m_mac->set_key(T);
 
-      m_mac->update(m_V.data(), m_V.size());
-      m_mac->final(m_V.data());
+      m_mac->update(m_V);
+      m_mac->final(m_V);
       }
    }
 
--- a/src/lib/rng/rng.cpp
+++ b/src/lib/rng/rng.cpp
@@ -11,7 +11,7 @@
 
 namespace Botan {
 
-void RandomNumberGenerator::randomize_with_ts_input(uint8_t output[], size_t output_len)
+void RandomNumberGenerator::randomize_with_ts_input(std::span<uint8_t> output)
    {
    if(this->accepts_input())
       {
@@ -23,19 +23,12 @@ void RandomNumberGenerator::randomize_with_ts_input(uint8_t output[], size_t out
       store_le(OS::get_system_timestamp_ns(), additional_input);
       store_le(OS::get_high_resolution_clock(), additional_input + 8);
 
-      this->randomize_with_input(output, output_len, additional_input, sizeof(additional_input));
+      this->fill_bytes_with_input(output, additional_input);
       }
    else
       {
-      this->randomize(output, output_len);
+      this->fill_bytes_with_input(output, {});
       }
-   }
-
-void RandomNumberGenerator::randomize_with_input(uint8_t output[], size_t output_len,
-                                                 const uint8_t input[], size_t input_len)
-   {
-   this->add_entropy(input, input_len);
-   this->randomize(output, output_len);
    }
 
 size_t RandomNumberGenerator::reseed(Entropy_Sources& srcs,
@@ -56,9 +49,7 @@ void RandomNumberGenerator::reseed_from_rng(RandomNumberGenerator& rng, size_t p
    {
    if(this->accepts_input())
       {
-      secure_vector<uint8_t> buf(poll_bits / 8);
-      rng.randomize(buf.data(), buf.size());
-      this->add_entropy(buf.data(), buf.size());
+      this->add_entropy(rng.random_vec(poll_bits / 8));
       }
    }
 
--- a/src/lib/rng/rng.cpp
+++ b/src/lib/rng/rng.cpp
@@ -9,19 +9,37 @@
 #include <botan/internal/loadstor.h>
 #include <botan/internal/os_utils.h>
 
+#if defined(BOTAN_HAS_SYSTEM_RNG)
+  #include <botan/system_rng.h>
+#endif
+
+#include <array>
+
 namespace Botan {
 
 void RandomNumberGenerator::randomize_with_ts_input(std::span<uint8_t> output)
    {
    if(this->accepts_input())
       {
-      /*
-      Form additional input which is provided to the PRNG implementation
-      to paramaterize the KDF output.
-      */
-      uint8_t additional_input[16] = { 0 };
-      store_le(OS::get_system_timestamp_ns(), additional_input);
-      store_le(OS::get_high_resolution_clock(), additional_input + 8);
+      constexpr auto s_hd_clk = sizeof(decltype(OS::get_high_resolution_clock()));
+      constexpr auto s_sys_ts = sizeof(decltype(OS::get_system_timestamp_ns()));
+      constexpr auto s_pid    = sizeof(decltype(OS::get_process_id()));
+
+      std::array<uint8_t, s_hd_clk + s_sys_ts + s_pid> additional_input = {0};
+      auto s_additional_input = std::span(additional_input.begin(), additional_input.end());
+
+      store_le(OS::get_high_resolution_clock(), s_additional_input.data());
+      s_additional_input = s_additional_input.subspan(s_hd_clk);
+
+#if defined(BOTAN_HAS_SYSTEM_RNG)
+      System_RNG system_rng;
+      system_rng.randomize(s_additional_input);
+#else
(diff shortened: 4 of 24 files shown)


src/lib/rng/rng.h
```diff
@@ -254,7 +254,7 @@ class BOTAN_PUBLIC_API(2,0) RandomNumberGenerator
       template<typename T = secure_vector<uint8_t>>
       requires(concepts::contiguous_container<T> &&
                concepts::resizable_container<T> &&
-               std::default_initializable<T> &&
+               concepts::default_initializable<T> &&
```
devnexen, line 257: we ll get there :-)

randombit:
> Hence, I'd like to make the argument to ditch the C-style APIs in favor of fully embracing std::span<>. Doing this consistently across the API is a huge break in compatibility. But I think it's worth considering.

I think this is a total non-starter. That would break every single application out there, probably piss a lot of people off, and anytime you're forced to modify your perfectly good application code to deal with a new library version that broke its API for no good reason, your mind will inevitably be drawn to the fact that if you're rewriting it anyway, you might as well pick a different library. I personally do not want to be responsible for any new code being written that uses OpenSSL.

> I opted to keep the old pointer-length APIs and add overloads using std::span<> for backward compatibility. 

This also seems to me to be not the correct approach. It only provides the benefits of `span` to our users (who may well not appreciate them!) while not providing us, as the library developers, any benefit at all.

Also I think there is an important difference here, wrt compatability:

* I'm very much not ok with breaking totally typical application code that, say, instantiates a `System_RNG` and uses it for RNG-ish things.
* I'm fine with us breaking code that *derives a new instance of one of our base classes*, because that is rare and you're probably doing something weird that is literally impossible with most other libraries, so you'll accept some breakage as part of the process.

I would sketch something like this instead (here just doing `randomize` but it extends naturally):

```
class RandomNumberGenerator
   {
   public:

   // no longer virtual
   // we could deprecate this eventually
   void randomize(uint8_t out[], size_t len)
     {
     this->fill_bytes(std::span<uint8_t>(out, len));
     }

    // This is what all subclasses of RandomNumberGenerator must implement in 3.0
    virtual void fill_bytes(std::span<uint8_t> output) = 0;
    ...
```

randombit:
If we are revisiting how this base class works we could also unify `randomize`, `add_entropy`, and `randomize_with_input` with just a single call:

`fill_bytes_with_input(span input, span output)` where

`randomize` -> `fill_bytes_with_input` with an empty input buffer
`add_entropy` -> empty output buffer
`randomize_with_input` -> both spans (possibly) populated.

reneme:
> I think this is a total non-starter.That would break every single application out there, [...]

It would break applications that use pointer-length-style APIs in a way that is easily and unambiguously fixable by using a modern idiom for it. Certainly far from a "rewrite" IMHO. I agree though that it might be too much of a burden on users, nevertheless. No hard feelings, I just found it worth discussing, despite being a radical proposal.

> Re: compatibility of deriving classes vs. typical applications

Fair enough. That's a distinction that wasn't obvious to me, but that makes a whole lot of sense. I'll sketch out your proposal of replacing the virtual `randomize*()` and `add_entropy()` with `fill_bytes_with_input(std::span)`.

reneme:
I converted this to a draft PR and split off #3201 that is technically unrelated from the changes to the internals of `RandomNumberGenerator`.

randombit:
> It would break applications that use pointer-length-style APIs in a way that is easily and unambiguously fixable by using a modern idiom for it. Certainly far from a "rewrite" IMHO.

I think now that std::span is available we should absolutely use it internally everywhere (eventually). I don't think it's that difficult to leave (as in my `randomize` example above) ptr/length APIs available for public compat even if we never use them internally ourselves.

I would envision a timeline for removing such APIs entirely as something like

1) Full set of std::span based APIs introduced in 3.0, with appropriate ptr/length compat
2) If not in 3.0, eventually all internals are rewritten to use span
3) Some where along the way we deprecate the ptr/length versions
4) In 4.0 the ptr/length versions are removed and everything is a span.

reneme:
> I would envision a timeline for removing such APIs entirely as something like [...]

Sounds fair to me. I'll work on my patience regarding API changes, going forward. 😊

reneme:
I finally came back to this and [applied your suggestion](https://github.com/randombit/botan/pull/3195#issuecomment-1397813174) across the code base. Most things fit nicely with this interface. The `Stateful_RNG` sub-tree needed a bit of mind-bending but the test coverage helps a lot here! Also `AutoSeeded_RNG` does some conservative mixing with a high-resolution timer which needs a thorough review to ensure I didn't screw something up.

The public API is backward-compatible, with all C-style methods now being accompanied py a `std::span<>` sibling and simply map to that one. Virtual methods only use with `std::span<>`.

Note that the public `randomize_with_input()` and the virtual (protected) `fill_bytes_with_input()` are one and the same. Though, when adapting [`Buffered_Computation`](https://github.com/randombit/botan/pull/3294) it came in handy that public and private-virtual interface was disjoint. So I kept this for good measure.

Also, I think, we could clean up the vector-centric API a bit more (definitely in a follow-up, though):

* [ ] *Deprecate `random_vec(std::span)`*
       This replaced `template<typename AllocT> random_vec(std::vector<uint8_t, AllocT>)`, so we can't just get rid of it. But it is equivalent to `randomize(std::span)` which has the better fitting name.
* [ ] *Deprecate `template<typename T> random_vec(size_t)` (and friends)*
       ... and rename it simply to `template<typename T> random(size_t)` (but that may be just a nit-pick).

src/lib/rng/stateful_rng/stateful_rng.cpp
```diff
@@ -34,68 +34,47 @@ bool Stateful_RNG::is_seeded() const
...
-   add_entropy(input, len);
-   }
-
-void Stateful_RNG::randomize(uint8_t output[], size_t output_len)
-   {
-   randomize_with_input(output, output_len, nullptr, 0);
+   add_entropy(input);
    }
 
-void Stateful_RNG::randomize_with_ts_input(uint8_t output[], size_t output_len)
```
lieser, line 62: Do you know why for the `Stateful_RNG:` the used input for `randomize_with_ts_input()` was a little different than for the general implementation?

I think it is good to unify what input is used, just want to make sure the change is intentional and OK to do.

src/lib/rng/stateful_rng/stateful_rng.cpp
```diff
@@ -34,68 +34,47 @@ bool Stateful_RNG::is_seeded() const
...
-#else
-   store_le(OS::get_system_timestamp_ns(), additional_input + 8);
-   store_le(OS::get_process_id(), additional_input + 16);
-#endif
+   if(output.empty())
+      {
+      // We won't produce any output but we must give the derived stateful
+      // RNG a chance to use the provided input for seeding. Therefore, no
+      // reseed_check() is performed before calling generate_output().
+      this->generate_output({}, input);
```
lieser, line 54: This changes the behavior from before:

1. A call to `randomize_with_input()` then no output was requested was a noop before. I think it is good to change this, as the default implementation did behave different. But still something to keep in mind.
2. When calling `randomize_with_input()` with both input and output, we now do this additional `generate_output({}, input)`, even if the input will be later already included via the other calls to `generate_output()`. Is this intentionally? The biggest problem I see here is that e.g. HMAC would no longer follow strictly *NIST SP800-90A section 10.1.2.2*. Even if the additional update should not hurt, not following the specification without a good reason is I think something we should avoid. 
3. Calling `add_entropy()` for a RNG that has no limit (i.e. `max_number_of_bytes_per_request() == 0`, which at least for HMAC should never be the case) will result in a call to `reseed_check()` and `generate_output()`. Is this something we want?

src/lib/rng/rng.h
```diff
@@ -40,18 +40,28 @@ class BOTAN_PUBLIC_API(2,0) RandomNumberGenerator
       RandomNumberGenerator(const RandomNumberGenerator& rng) = delete;
       RandomNumberGenerator& operator=(const RandomNumberGenerator& rng) = delete;
 
+   protected:
```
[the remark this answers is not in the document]
reneme: That's the main (functional) method that needs to be overridden by subclasses and I wanted it to be "prominent" in the interface. No hard feelings about moving it down though.

src/lib/rng/rng.h
```diff
@@ -165,13 +178,13 @@ class BOTAN_PUBLIC_API(2,0) RandomNumberGenerator
       /**
       * Fill a given byte container with @p bytes random bytes
       *
+      * @todo deprecate this overload (in favor of randomize())
```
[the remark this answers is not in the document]
reneme: I wanted to put it up for discussion. See [my last top-level comment](https://github.com/randombit/botan/pull/3195#issuecomment-1435001817).

src/lib/rng/stateful_rng/stateful_rng.h
```diff
@@ -72,7 +72,11 @@ class BOTAN_PUBLIC_API(2,0) Stateful_RNG : public RandomNumberGenerator
       * of the length of the input or the current seeded state of
       * the RNG.
       */
-      void initialize_with(const uint8_t input[], size_t length);
+      void initialize_with(std::span<const uint8_t> input);
+      void initialize_with(const uint8_t input[], size_t length)
+         {
+         this->initialize_with(std::span(input, length));
+         }
```
reneme, line 79: This might be another instance of the virtual dispatch issue [discussed here](https://github.com/randombit/botan/pull/3317). Probably we want to also add a protected interface with another name for good measure.

src/lib/rng/stateful_rng/stateful_rng.cpp
```diff
@@ -34,68 +34,47 @@ bool Stateful_RNG::is_seeded() const
...
    {
-   uint8_t additional_input[20] = { 0 };
-
-   store_le(OS::get_high_resolution_clock(), additional_input);
+   lock_guard_type<recursive_mutex_type> lock(m_mutex);
 
-#if defined(BOTAN_HAS_SYSTEM_RNG)
-   System_RNG system_rng;
-   system_rng.randomize(additional_input + 8, sizeof(additional_input) - 8);
-#else
```
reneme, line 71: I believe this vanished and should probably come back.

src/lib/rng/chacha_rng/chacha_rng.cpp
```diff
@@ -58,25 +58,20 @@ void ChaCha_RNG::clear_state()
...
-   m_hmac->update(input, input_len);
-   m_chacha->set_key(m_hmac->final());
+      secure_vector<uint8_t> mac_key(m_hmac->output_length());
+      m_chacha->write_keystream(mac_key.data(), mac_key.size());
+      m_hmac->set_key(mac_key);
+      }
 
-   secure_vector<uint8_t> mac_key(m_hmac->output_length());
-   m_chacha->write_keystream(mac_key.data(), mac_key.size());
-   m_hmac->set_key(mac_key);
```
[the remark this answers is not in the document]
reneme: Previously, `Stateful_RNG` had `::update()` and `::generate_output()`. IIUC that was done so that `add_entropy()` could just call `::update()` and don't bother with the generation code at all. Now that `add_entropy` is implemented in terms of `::fill_bytes_with_input()` we don't need that distinction anymore. Merging `::update()` and `::generate_output()` increases the control flow complexity a little, though.

src/lib/rng/stateful_rng/stateful_rng.cpp
reneme: My understanding is that `Stateful_RNG` wants to be a bit more conservative to make sure that the state is always reasonably seeded.

src/lib/rng/auto_rng/auto_rng.cpp
```diff
@@ -83,27 +87,24 @@ std::string AutoSeeded_RNG::name() const
...
+void AutoSeeded_RNG::fill_bytes_with_input(std::span<uint8_t> out, std::span<const uint8_t> in)
    {
-   m_rng->randomize_with_ts_input(output, output_len);
-   }
+   std::array<uint8_t, 16> additional_input;
+   if(in.empty() && m_rng->accepts_input())
+      {
+      store_le(OS::get_system_timestamp_ns(), additional_input.data());
+      store_le(OS::get_high_resolution_clock(), additional_input.data() + 8);
+      in = additional_input;
```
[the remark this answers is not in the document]
reneme: The diff seems to be a bit misleading here. This code merges the behaviours of the former `::randomize()` and `::randomize_with_input()`. Therefore it has a special case for `in` being empty. Nevertheless, we might indeed replace the special case with a call to `randomize_with_ts_input()` and call `randomize_with_input()` in an `else` branch.

This would essentially make the call recursive, though. `randomize_with_ts_input()` will eventually call `fill_bytes_with_input()` again (but not trigger the special case, because `in` won't be empty anymore. I'm frankly not sure whether this little code duplication justifies this.

Sketch:

```C++
void AutoSeeded_RNG::fill_bytes_with_input(std::span<uint8_t> out, std::span<const uint8_t> in)
   {
   if(in.empty())
      {
      // Will call `fill_bytes_with_input()` recursively after filling `in`
      // with entropy pulled from the high resolution system clock.
      m_rng->randomize_with_ts_input(out);
      }
   else
      {
      m_rng->randomize_with_input(out, in);
      }
   }
```

It certainly is tempting, but should we really risk an endless recursion (when `randomize_with_ts_input()` does break its oath)?

src/lib/rng/stateful_rng/stateful_rng.cpp
reneme: > A call to randomize_with_input() then no output was requested was a noop before. I think it is good to change this, as the default implementation did behave different. But still something to keep in mind.

I agree with that. Input would now be incorporated. I also agree that this is probably fine and should likely have been the case before.

Re 2:

If `output` is non-empty we won't enter this branch (line 49: `if(output.empty()) // ...`). Otherwise this would indeed be a problem. There are KAT-tests for HMAC DRBG that back this implementation and I would expect them to fail if we'd deviate from the spec like that.

Re 3:

We wouldn't want that, and it won't happen either. I believe you missed the `return` statement (line 61) that early-returns from this special-case branch. I.e. for `output.empty()` this special case will only update the underlying implementation's state but won't actually try to draw any random bits.

Nevertheless, I think this discussion is making the point that the control flow in this method is actually too hard to disect. That's what I meant with "increases the control flow complexity" [here](https://github.com/randombit/botan/pull/3195/files#r1121292472).

reneme:
@randombit

There's a gotcha with making `randomize_with_ts_input()` non-virtual. As @lieser [pointed out](https://github.com/randombit/botan/pull/3195/files#r1118951688) `Stateful_RNG` used to override the default implementation of this method. In short, it uses the `System_RNG` additional to a high-res clock reading or, if `System_RNG` is not available, it incorporated another high-res clock reading and the current process ID.

To my mind, that is done to be more conservative in the stateful RNG case (particularly the process ID is likely meant to be a counter-measure to `fork()` duplicating the RNG state). Note that `Stateful_RNG::reseed_check()` has an explicit fork-detection.

Without `randomize_with_ts_input()` being `virtual` I don't see a way to keep this exact behavior, unfortunately. Except by moving it up into the base implementation of `RandomNumberGenerator` which might be somewhat overkill.

I see three ways forward:

1. Move the conservative implementation up into the base implementation,
2. Leave the `randomize_with_ts_input()` as `virtual` and override it in `Stateful_RNG` as before. Note that we need to keep the [virtual overload dispatch issue](https://github.com/randombit/botan/pull/3317) in mind -- i.e. introduce a `protected` alias that gets overridden,
3. Remove the conservative special case and rely on the `reseed_check()`.

src/lib/rng/stateful_rng/stateful_rng.cpp
lieser: 2) You are right. I think for some reason I thought of the `if(output.empty())` being `if(!input.empty())` then I wrote this.

3) Yes I misses the `return` completely.

And agree that it is not the easiest to follow how the `Stateful_RNG::fill_bytes_with_input()` and `generate_output()` interact now.
Don't know if keeping the `Stateful_RNG::add_entropy()` and `Stateful_RNG::randomize_with_input()` as private methods would help here or if that is just more indirection without real benefit. I.e.:

```c++
void Stateful_RNG::fill_bytes_with_input(std::span<uint8_t> output, std::span<const uint8_t> input)
   {
   if(output.empty())
      {
      this->add_entropy(input);
      return;
      }
   this->randomize_with_input(output, input);
   }
```

But I think a lot of the problem came also from me trying to understand it just viewing it in the unified diff view.

src/lib/rng/rng.h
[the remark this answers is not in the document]
lieser: I would tend to say it is more important to write it in a way that is useful for users of a `RandomNumberGenerator`, instead of people wanting to create their own derived subclass.

But I'm also OK with leaving as is, hopefully the documentation is good enough that users don't need to read the header to understand how to use a RNG in Botan.

src/lib/rng/stateful_rng/stateful_rng.cpp
reneme: I'll add a bit more documentation to the control flow in the hope that this will be enough.

src/lib/rng/stateful_rng/stateful_rng.cpp
```diff
@@ -34,83 +34,63 @@ bool Stateful_RNG::is_seeded() const
...
+      while(output_length > 0)
          {
-         const size_t this_req = std::min(max_per_request, output_len);
-
-         /*
-         * We split the request into several requests to the underlying DRBG but
-         * pass the input to each invocation. It might be more sensible to only
-         * provide it for the first invocation, however between 2.0 and 2.15
-         * HMAC_DRBG always provided it for all requests so retain that here.
-         */
```
reneme, line 105: We altered the behavior in the sense described here. I.e. using the input only with the first round. As a result `a` and `b` result in the same byte stream (assuming equally seeded RNGs),

```C++
auto a = rng.random_vec(64*1024) + rng.random_vec(64*1024) + rng.random_vec(64*1024);
auto b = rng.random_vec(3*64*1024)
```

See also: https://github.com/randombit/botan/pull/3195/commits/fff82489dabf6786734119ca3e74246a5169db10

@randombit Do you think that's okay?

lieser: If we keep this we probably also want to mention it int the PR description.

src/lib/rng/rng.h
```diff
@@ -40,18 +40,21 @@ class BOTAN_PUBLIC_API(2,0) RandomNumberGenerator
       RandomNumberGenerator(const RandomNumberGenerator& rng) = delete;
       RandomNumberGenerator& operator=(const RandomNumberGenerator& rng) = delete;
 
+   public:
```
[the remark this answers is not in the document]
reneme: Just a left-over...

lieser (approved):

src/lib/rng/stateful_rng/stateful_rng.cpp
randombit: This seems OK to me. When I introduced `Stateful_RNG` I wanted to retain the existing behavior of `HMAC_DRBG`, but this new behavior actually does make more sense and if we are going to change it, 3.0 is a good time for it.

randombit:
@reneme regarding `randomize_with_ts_input` I definitely want to keep this as conservative as possible since userspace RNGs are prone to all kinds of fun problems and a bug here could be catastrophic. But I think just moving the conservative version from `Stateful_RNG` to the base class is fine.

src/lib/rng/hmac_drbg/hmac_drbg.cpp
```diff
@@ -139,49 +139,52 @@ std::string HMAC_DRBG::name() const
...
    m_mac->update(m_V.data(), m_V.size());
    m_mac->final(m_V.data());
 
-   if(input_len > 0)
+   if(!input.empty())
       {
       m_mac->update(m_V);
       m_mac->update(0x01);
-      m_mac->update(input, input_len);
+      m_mac->update(input.data(), input.size()); // TODO: pass span after merging GH #3294
```
randombit, line 187: This and similar could be updated now, right?

src/lib/rng/chacha_rng/chacha_rng.cpp
```diff
@@ -58,20 +58,21 @@ void ChaCha_RNG::clear_state()
    m_chacha->set_key(m_hmac->final());
    }
 
-void ChaCha_RNG::generate_output(uint8_t output[], size_t output_len,
-                                 const uint8_t input[], size_t input_len)
+void ChaCha_RNG::generate_output(std::span<uint8_t> output, std::span<const uint8_t> input)
    {
-   if(input_len > 0)
+   BOTAN_ASSERT_NOMSG(!output.empty());
```
randombit, line 63: This seems wrong - `randomize(nullptr, 0)` should succeed and do nothing. (There should be a test for this)

randombit: It looks like there are tests along these lines for `System_RNG` and `Processor_RNG` but not the stateful implementations

src/lib/rng/auto_rng/auto_rng.cpp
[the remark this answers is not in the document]
randombit: This is not really recursive though, because `AutoSeeded_RNG` is calling its internal RNG, not itself?

src/lib/rng/chacha_rng/chacha_rng.cpp
randombit: Added a test covering this in a379e17340ce26a6f71fd59e3b5ec352d2cce152

reneme: The `generate_output()` method is the internal virtual entrypoint of `Stateful_RNG`. The implementation in `Stateful_RNG::fill_bytes_with_input()` is marked as `final` and won't call into `generate_output()` if the given output sink is empty. I.e. this is just a safety net.

src/lib/rng/auto_rng/auto_rng.cpp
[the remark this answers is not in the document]
reneme: Oh! Indeed. What I had in mind when writing that comment before was:

```C++
void AutoSeeded_RNG::fill_bytes_with_input(std::span<uint8_t> out, std::span<const uint8_t> in)
   {
   if(in.empty())
      {
      // Will call `fill_bytes_with_input()` recursively after filling `in`
      // with entropy pulled from the high resolution system clock.
      this->randomize_with_ts_input(out);
   // ^~ !! RECURSE !!
      }
   else
      {
      m_rng->randomize_with_input(out, in);
      }
   }
```

I.e. the call stack would have looked roughly like that:

```
...
application_code()
AutoSeeded_RNG::fill_bytes_with_input(out: <some buffer>, in: {})
AutoSeeded_RNG::fill_bytes_with_ts_input(out: <some buffer>)
AutoSeeded_RNG::fill_bytes_with_input(out: <some buffer>, in: {timestamp}) <= "in" is now filled (recursion abort)
Internal_RNG::fill_bytes_with_input(out: <some buffer>, in: {timestamp})
...
```

But the way I (accidentially) typed out 3 weeks ago would actually make much more sense. Especially with `randomize_with_ts_input()` not being `virtual` anymore. I.e. the internal RNG implementation cannot play any tricks on us. Yikes, object orientation spinning in my head.

src/lib/rng/stateful_rng/stateful_rng.cpp
reneme: Does this mandate a mention in the migration guide? It might break people's "deterministic" algorithms under certain circumstances after all.

randombit: Yes definitely worth a mention. Hopefully people aren't relying on this behavior since it's not entirely standard, but Hyrum's Law applies. The change in behavior is at least easy to work around in application code, by breaking up calls that are too long and intentionally repeating the input for each call, mimicking what we did internally.

src/lib/rng/rng.cpp
```diff
@@ -9,35 +9,44 @@
...
+#endif
+
+      this->fill_bytes_with_input(output, additional_input);
       }
    else
       {
-      this->randomize(output, output_len);
+      this->fill_bytes_with_input(output, {});
       }
    }
```
reneme, line 48: /cc @randombit

The base implementation of `RandomNumberGenerator::randomize_with_ts_input()` now adopted the behavior of `Stateful_RNG` (as discussed [here](https://github.com/randombit/botan/pull/3195#issuecomment-1449602320) and [here](https://github.com/randombit/botan/pull/3195#issuecomment-1477692624)). Additionally, this method is not virtual any longer: I.e. all RNGs will inherit exactly this implementation.

There's one caveat: Naturally, also the `System_RNG` is a subclass of `RandomNumberGenerator`. In other words: if the platform's RNG reports that it can accept input, it ends up seeding itself. Technically, that shouldn't be problematic though its worth noting.

There shouldn't be a chance of this becoming an endless loop either. `system_rng.randomize()` will resolve to `System_RNG::fill_bytes_with_input()` (without actual input). By design, this is the most low-level method that will not call back into higher-level members (like `randomize_with_ts_input()` to actually create an endless recursion).

reneme, line 48: /cc @randombit

The base implementation of `RandomNumberGenerator::randomize_with_ts_input()` now adopted the behavior of `Stateful_RNG` (as discussed [here](https://github.com/randombit/botan/pull/3195#issuecomment-1449602320) and [here](https://github.com/randombit/botan/pull/3195#issuecomment-1477692624)). Additionally, this method is not virtual any longer: I.e. all RNGs will inherit exactly this implementation.

There's one caveat: Naturally, also the `System_RNG` is a subclass of `RandomNumberGenerator`. In other words: if the platform's RNG reports that it can accept input, it ends up seeding itself. Technically, that shouldn't be problematic though its worth noting.

There shouldn't be a chance of this becoming an endless loop either. `system_rng.randomize()` will resolve to `System_RNG::fill_bytes_with_input()` (without actual input). By design, this is the most low-level method that will not call back into higher-level members (like `randomize_with_ts_input()` to actually create an endless recursion).

src/lib/rng/stateful_rng/stateful_rng.cpp
reneme: It did: https://github.com/randombit/botan/pull/3195#discussion_r1143575248

src/lib/rng/rng.cpp
randombit: That seems OK to me. (As long as there is no infinite recursions! which seems we are safe here.) Typically one would not use `randomize_with_ts_input` on a system RNG but rather only a userspace RNG. Also at this point only a few systems (if any?) use the `/dev/urandom` based interface anymore, which is the only `System_RNG` instance that supports writing at all.

randombit (approved):

src/lib/rng/rng.cpp
reneme: We could consider `final` overriding `System_RNG::accepts_input()` and let it return `false` to be explicit about that. But that can be done in a follow-up.

src/lib/rng/stateful_rng/stateful_rng.h
reneme, line 79: General remark (out of scope of this pull request):

The doxygen string says:

```
      /**
      * Consume this input and mark the RNG as initialized regardless
      * of the length of the input or the current seeded state of
      * the RNG.
      */
```

Assuming that "mark the RNG as initialized" means "considering the RNG as seeded": I don't think that is the case in general!

The `input` will eventually flow down to the following code:

```C++
this->update(input);

if(8*input.size() >= security_level())
   {
   reset_reseed_counter();
   }
```

Which will reset the reseed counter only if the input was long enough. This contradicts with "regardless of the length of the input" in the doxygen comment above. I.e. `stateful_rng->initialize_with("baadcafe")` wouldn't result in a seeded `stateful_rng` object.

@randombit I think we should consider fixing this.

However: The behavior was the same before the refactoring, so I consider that out-of-scope.

Issue #3190, opened 2023-01-18
Expose ZFEC functionality via the C-compatible FFI interface


exarkun:
It would be nice to use Botan's ZFEC implementation from languages/runtimes other than C++.  The obvious way to enable this to add a `extern "C"` style interface on top of the C++ interface, similar to how other functionality is exposed in `src/lib/ffi/`.

A reasonably simple interface to use might be something like:

```
// input gives the primary shares
// size gives the length of the primary shares (all equal)
// output gives a place to write a uint8_t** giving the generated secondary shares
// outputSizes gives a place to write size_t* giving the size of the generated shares (all equal)
// memory ownership of *output and all contained pointers and *outputSizes is given to the caller on success
int botan_zfec_encode(size_t K, size_t N, const uint8_t input[], size_t size, uint8_t ***output, size_t **outputSizes);

// shares gives primary or secondary shares to decode
// share_indexes gives the index of corresponding elements of shares
// num_shares gives the length of shares and share_indexes
// share_size gives the length of each share
// output and outputSizes are like for encode
int botan_zfec_decode(size_t K, size_t N, const uint8_t shares[], const size_t[] share_indexes, size_t num_shares, size_t share_size, uint8_t ***output, size_t **outputSizes);
```

A different interface could use `output_cb_t` instead of output buffers which might enable less copying but it's also harder (but maybe not impossible to consume across an FFI interface).

This interface does not preclude the addition of such an interface later on, if it proves desirable enough to do the extra work.

I have a preliminary implementation of `botan_fec_encode` that I've demonstrated works when called from Haskell.  I can finish it (docs, tests, whatever else it needs) and work on `botan_fec_decode` if this addition is welcome.

randombit:
Yes we'd be happy to take this as a PR (probably would need some additional tests/etc before merge but the general approach seems fine)

exarkun:
I think #3196 is in reasonable shape for a first review.  I'm sure I got many things wrong there is at least a passing test that verifies the round-trip behavior for encode/decode.  Looking forward to any feedback.

randombit:
Done! Will be available in 3.0.0 which is estimated to be released sometime next month.

PR #3196, opened 2023-01-19
Add an FFI-friendly interface for ZFEC encoding and decoding

exarkun:
Relates to #3190

### DIFF
--- a/src/lib/ffi/ffi_zfec.cpp
+++ b/src/lib/ffi/ffi_zfec.cpp
@@ -1,5 +1,5 @@
 /*
-* (C) 2015,2017,2018 Jack Lloyd
+* (C) 2023 Least Authority TFA GmbH
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
--- a/src/lib/ffi/ffi_zfec.cpp
+++ b/src/lib/ffi/ffi_zfec.cpp
@@ -20,12 +20,12 @@ extern "C" {
     });
   }
 
-  int botan_zfec_decode(size_t K, size_t N, uint8_t *const*const inputs, size_t size, uint8_t **outputs, size_t *sizes) {
-    std::map<size_t, const uint8_t*> shares;
-    for (size_t k = 0; k < K; ++k) {
-      shares.insert(std::pair<size_t, const uint8_t*>(k, inputs[k]));
-    }
+  int botan_zfec_decode(size_t K, size_t N, const size_t *indexes, uint8_t *const*const inputs, size_t size, uint8_t **outputs, size_t *sizes) {
     return Botan_FFI::ffi_guard_thunk(__func__, [=]() -> int {
+      std::map<size_t, const uint8_t*> shares;
+      for (size_t k = 0; k < K; ++k) {
+	shares.insert(std::pair<size_t, const uint8_t*>(indexes[k], inputs[k]));
+      }
       Botan::ZFEC(K, N).decode_shares(shares, size, [=](size_t index, const uint8_t block[], size_t blockSize) -> void {
 	  outputs[index] = new uint8_t[blockSize];
 	  std::copy(block, block + blockSize, outputs[index]);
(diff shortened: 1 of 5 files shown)


src/lib/ffi/ffi_zfec.cpp
```diff
@@ -0,0 +1,37 @@
+/*
+* (C) 2023 Least Authority TFA GmbH
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/ffi.h>
+#include <botan/zfec.h>
```
randombit, line 8: The include and the functions themselves should be guarded by `BOTAN_HAS_ZFEC`, so that it is possible to compile the C layer even if ZFEC is disabled from the build.

The functions themselves should still be defined mind you! And just return `BOTAN_FFI_ERROR_NOT_IMPLEMENTED` in that case.

randombit (commented):
Thanks @exarkun! This already looks very good.

I think the big remaining issue here is to also amend the Python wrapper in `src/python/botan2.py` with the appropriate ctypes declarations, plus a reasonably Pythonic wrapper and a simple test (along the lines of what you wrote in C++ already) that it works.

exarkun:
@randombit I think this is ready for another look.  Thanks!

randombit:
@exarkun Great! I'll review asap. One immediate comment: please rebase and remove the merge commit.

src/python/botan3.py
```diff
@@ -1825,3 +1831,88 @@ def srp6_client_agree(username, password, group, hsh, salt, b, rng):
...
+    ])
+
+    # actual C call
+    x = _DLL.botan_zfec_encode(
+        c_size_t(k), c_size_t(n), input_bytes, c_size_t(input_size), c_outputs
+    )
+    if x != 0:
+        raise RuntimeError(
+            "Unexpected error, code={}".format(x)
+        )
```
randombit, line 1868: This check on the return value should not be required, because `ffi_api` sets the `errcheck` property to a function (`_errcheck`) which already checks that the return value is 0 and throws otherwise.

src/python/botan3.py
```diff
@@ -1825,3 +1831,88 @@ def srp6_client_agree(username, password, group, hsh, salt, b, rng):
...
+    c_outputs = p_p_kbytes(*[
+        addressof(output)
+        for output in outputs
+    ])
+
+    # actual C call
+    x = _DLL.botan_zfec_decode(
+        c_size_t(k), c_size_t(n), c_indexes, c_inputs, c_size_t(share_size), c_outputs
+    )
+    if x != 0:
```
randombit, line 1914: Similarly this check should not be necessary.

src/scripts/test_python.py
```diff
@@ -718,6 +719,66 @@ def test_srp6(self):
...
+    def test_encode(self):
+        """
+        Simple encoder test.
+
+        Could benefit from more variations
+        """
+        n = 3
+        k = 2
+        input_bytes = b"abcdefgh" + b"ijklmnop"
+        output_shares = botan2.zfec_encode(k, n, input_bytes)
```
randombit, line 737: These tests are fine, but they need to be adapted to the new import. Specifically here in the tests we are importing `botan3` as `botan` (so we don't have to modify all of the tests every time we bump the major version). This is currently causing some CI failures.

randombit (commented):
This looks very good and basically ready to merge. There are couple of very small issues, basically just the unnecessary checks on the return value in the Python wrapper, plus needing to change the references to the module in the Python tests.

src/python/botan3.py
exarkun: Fixed in d251adf1b

src/python/botan3.py
exarkun: Fixed in d251adf1b

src/scripts/test_python.py
exarkun: Fixed in 7136dd7ce

exarkun:
Thanks!  I believe I have addressed those points.

randombit:
@exarkun Looks good! Just waiting on CI now

PR #3199, opened 2023-01-19
Use Clang for the coverage build

randombit:
No idea if this is faster but worth a try


randombit:
It seems that codecov's tool can't handle this

PR #3201, opened 2023-01-20
Feature: Generalization of RandomNumberGenerator::random_vec()

reneme:
This is a spin-off from #3195 to harvest the low-hanging fruit. Namely, allowing to use `RNG::random_vec()` with arbitrary contiguous byte buffers.

Example usage (e.g. as desired in #3150):

```C++
// generate a random TLS Session ID
using Session_ID = Botan::Strong<std::vector<uint8_t>, struct Session_ID_>;
const auto random_id = rng.random_vec<Session_ID>(32);

// filling an arbitrary buffer with random bytes
std::array<uint8_t, 32> buffer;
rng.random_vec(buffer);
```

### DIFF
--- a/src/lib/utils/concepts.h
+++ b/src/lib/utils/concepts.h
@@ -12,6 +12,7 @@
 #include <compare>
 #include <concepts>
 #include <type_traits>
+#include <ostream>
 
 namespace Botan::concepts {
 
@@ -70,6 +71,7 @@ concept container = requires(T a)
    { a.cbegin() } -> container_iterator<T>;
    { a.cend() } -> container_iterator<T>;
    { a.size() } -> std::same_as<typename T::size_type>;
+   typename T::value_type;
    };
 
 template<typename T>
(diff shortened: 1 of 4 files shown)


src/lib/utils/concepts.h
```diff
@@ -12,6 +12,7 @@
 #include <compare>
 #include <concepts>
 #include <type_traits>
+#include <ostream>
```
randombit, line 15: Either this was a include left over from debugging, or there is something I'm missing about how these concepts work.

randombit (approved):
This part is fine

src/lib/utils/concepts.h
reneme: TL;DR: include what you use

The include is needed for the `streamable<>` concept. This was not introduced here, but the compiler needed the header when including `<botan/concepts.h>` in some compilation unit that didn't slurp it in through other ways.

Its likely that this particular PR wouldn't have needed it. Nevertheless I left it in for completeness.

randombit: OK I was only looking at the diff and missed the preexisting concept. Yeah that's fine.

PR #3203, opened 2023-01-20
[std::span] X509:: and PKCS8:: load_key(std::span)

reneme:
This introduces `PKCS8::load_key` overloads for `std::span`, as well as for `X509::load_key`. The latter already had an overload for `std::vector<uint8_t>`, and for the sake of consistency, I think the former should take a bare memory region, too. To simplify things, `Data_Source_Memory` now also has a `std::span<>` c'tor.

As a library user, having to wrap things into `Data_Source_Memory` is always a small but noticable obstacle. Hence, I'm a fan of the memory container overloads. On the other hand, to be fully consistent, we should also add `std::secure_vector<uint8_t>` overloads to `PKCS8::load_key()`; but due to additional parameters the combinatorics of the overloads blow up in `PKCS8`.

Hence, I'm not sure this is actually a good idea and worth the added code.

### DIFF
--- a/src/lib/utils/data_src.h
+++ b/src/lib/utils/data_src.h
@@ -12,6 +12,7 @@
 #include <botan/secmem.h>
 #include <string>
 #include <iosfwd>
+#include <span>
 
 namespace Botan {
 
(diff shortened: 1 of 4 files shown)


src/lib/utils/data_src.h
```diff
@@ -121,8 +122,15 @@ class BOTAN_PUBLIC_API(2,0) DataSource_Memory final : public DataSource
       * Construct a memory source that reads from a secure_vector
       * @param in the MemoryRegion to read from
       */
-      explicit DataSource_Memory(const secure_vector<uint8_t>& in) :
-         m_source(in), m_offset(0) {}
+      explicit DataSource_Memory(secure_vector<uint8_t> in) :
+         m_source(std::move(in)), m_offset(0) {}
```
reneme, line 126: `DataSource_Memory` is creating an internal copy of the data. With this patch, users can avoid that if necessary by moving the data buffer in.

randombit (approved):

reneme:
Fixed a merge conflict. Waiting for CI, then merging...

PR #3205, opened 2023-01-21
Add AVX2 implementation of Argon2's BLAMKA function

randombit:
About 10-20% faster on Skylake


randombit:
Amalgamation build with GCC fails, *only* with `-O3`, works fine on my machine with `-O2`. Very likely a codegen bug, unfortunately with it only affecting the amalgamation, reducing it sufficiently that GCC devs will be willing to look may be difficult.

PR #3206, opened 2023-01-21
Initial AVX512 support

randombit:
Only tested with Intel SDE at the moment.

I'm sure Serpent can be made much faster using vpternlog


randombit:
Waiting until I have hardware to check performance before merging. Probably for 3.1

Issue #3208, opened 2023-01-22
Using Dilithium as PKI Signer Algo.


obalak:
To generate a self signed root CA for PSQ PKI, I tried the following:

`botan keygen --algo=Dilithium --params=Dilithium-AES-r3/6x5 --passphrase=xxxx --pbe= --pbe-millis=300 --provider= > ~/tmp/rootca_key.pem`

Which generates correctly the key. Then I generate the root CA certificate with the following:

`botan gen_self_signed ~/tmp/rootca_key.pem "My Root CA" --country=CH --dns=cloud.xxxx.ch --organization=xxxx --email=info@xxxx.ch --path-limit=1 --days=7300  --key-pass=xxxx --ca  --hash=SHA-512 --emsa= > ~/tmp/rootca_crt.pem`

The certificate is not generated and an error is issued:
`Error: Unknown X.509 signing key type: Dilithium-AES-r3/6x5`

I checked the source code and find out that the algorithm is not yet supported.
When do you think this will be supported? Or do you recommend other approaches?

randombit:
We should support this if we can and I believe we already have everything needed to make it work (SPKI encodings, etc) it probably just requires a bit of glue so the X.509 handling code knows what to do.

obalak:
Thanks. Waiting for the alpha1 to test.

reneme:
This should now work as expected. Be aware that the algorithm name changed from "Dilithium-AES-r3/6x5" to "Dilithium-6x5-AES-r3". The slash in the old name had an internal semantic in the X.509 code that made this change necessary.

Feel free to test it with the latest master commit and re-open if you still run into issues.

obalak:
It is working fine now with the fixed scheme name for all botan main PKI operations except for certificate verification.

`/usr/local/bin/botan cert_verify ~/build/pkigen/certs/entity_crt.pem ~/build/pkigen/certs/subca_crt.pem ~/build/pkigen/certs/rootca_crt.pem
Error: Internal error: Invalid name format found for 1.3.6.1.4.1.25258.1.10.2
`
It should be related to the name of the signature algorithm ASN.1.

result of : `botan cert_info --fingerprint certs/entity_crt.pem ` is 
`
Signature algorithm: Dilithium-6x5-AES-r3
...
Public Key [Dilithium-6x5-AES-r3-1952]
`
Let me know if you need help to have a makefile to generate all the test material.
I'm testing on RockyLinux 9.1 in docker.

obalak:
[...]

PR #3212, opened 2023-01-23
Dilithium/Kyber in X.509

reneme:
This introduces glue code to allow using Kyber keys in X.509 certificates (for usage 'key exchange' only). Also, Dilithium keys can now be used to sign X.509 objects. I updated `unit_x509` on a best effort basis. Are there more tests that need adaption?

I posted some potential inconsistencies for discussion below.

Closes https://github.com/randombit/botan/issues/3208.

#### Algorithm Naming Conventions

Kyber and Dilithium are incorporating the algorithm parameters ("size" and "modern vs. 90s/AES") into their algo name. Therefore, `Public_Key::algo_name()` returns something along the lines of "Kyber-1024-90s-r3" or "Dilithium-6x5-r3". For other common asymmetric algorithms `Public_Key::algo_name()` is static (e.g. "RSA" or "ECDSA" regardless of the key size). The same reflects in their (preliminary) OID assignments -- one OID per algorithm configuration. Latest IETF drafts for [Kyber](https://datatracker.ietf.org/doc/draft-uni-qsckeys-kyber/00/) and [Dilithium](https://datatracker.ietf.org/doc/draft-uni-qsckeys-dilithium/00/) suggest that this approach is correct. Nevertheless, I find it inconsistent with the rest of the library.

#### OID Assignment

When we added Kyber, the prelimenary OID assignments were taken from the (now outdated) [draft-uni-qsckeys 00](https://datatracker.ietf.org/doc/draft-uni-qsckeys/00/). To the best of my understanding, latest drafts deliberately leave the OIDs to be defined. Hence, for Dilithium we use OIDs under Botan's private arc. Should we do the same for the Kyber OIDs? (I saw that [oids.rst](https://github.com/randombit/botan/blob/master/doc/dev_ref/oids.rst) actually already reserved a number for Kyber in 2019.

#### X.509 Key Usage for Kyber

I'm not entirely sure how to deal with Kyber keys here. Technically Kyber is a KEM, not a KEX: I.e. it could be used for key encipherment, no? Should we allow 'encryption' in `verify_cert_constraints_valid_for_key_type()`, therefore? Or is there a specific key usage for KEMs already?


randombit:
> Algorithm Naming Conventions

I don't think we are tied to anything here, in that we could change how this works in the future.

I think in a few places (in X509 mostly) we assume that it's possible to go from an algorithm name to an OID in an unambiguous way. That would have to be fixed, but those would be good cleanups to have in any event.

> OID Assignment

These formats are still in flux, and we do not control these OIDs. We should probably ship 3.0 using private OIDs and then if/when a final format/OIDs are defined, transition. (For example by encoding using the new format while still accepting both formats for decoding).

> X.509 Key Usage for Kyber

I'm not aware of a standard key usage specifically for KEMs. To me key encipherment seems a reasonable approximation of what a KEM is doing.

randombit (approved):

reneme:
> I'm not aware of a standard key usage specifically for KEMs. To me key encipherment seems a reasonable approximation of what a KEM is doing.

The only adaption I did from the initial version of this PR is that Kyber may now be used for "agreement" and "encipherment".

PR #3215, opened 2023-01-24
Chore: Improve Test Registration

reneme:
This introduces an internal `Test_Registry` class to centralize test case registration and filtering.

Mostly, this allows to filter tests by category like: `./botan-test tls` would run all tests related to TLS. The category annotation in `BOTAN_REGISTER_TEST(category, test_name, ...)` did exist but was unused so far. To do that in a user-friendly way, category names and test names need to be disjoint.

Test filters can be mixed (i.e. some categories and some specific test names). Skipping specific tests within a category can be done as before using `--skip-tests`. There's no such thing as `--skip-category`, yet. Note that filtering is implemented fairly naively and inefficiently (with a bunch of lists), to keep it simple. But probably `O(n^2)` is fine for mangling a list of tests. 🤡 

Currently available test categories for reference:

```
Available test categories
----------------
asn1 block codec compat compression dilithium ec_h2c ffi filters
hash kdf keywrap kyber mac math misc modes otp pake pbkdf pkcs11
pubkey rng roughtime stream tls tls_extensions utils x509 zfec
```

Technically, "kyber" and "dilithium" should move into "pubkey", I guess. Or maybe "PQC"?

While changing a few test names, I started seeing race conditions in multi-threaded execution: Turns out that there were centralized lists of test names to identify "smoke tests" and "serialized tests" hidden in the test runner.
To avoid this tight coupling between test names and runner behavior, tests are now annotated accordingly during registration using `BOTAN_REGISTER_SERIALIZED_TEST` and `BOTAN_REGISTER_SMOKE_TEST`.

The above-described changes are split into individual commits for easier review. If needed, we could also split it into two or three PRs.

### DIFF
--- a/src/lib/utils/stl_util.h
+++ b/src/lib/utils/stl_util.h
@@ -44,6 +44,20 @@ std::set<K> map_keys_as_set(const std::map<K, V>& kv)
    return s;
    }
 
+/**
+* Return the keys of a multimap as a std::set
+*/
+template<typename K, typename V>
+std::set<K> map_keys_as_set(const std::multimap<K, V>& kv)
+   {
+   std::set<K> s;
+   for(auto&& i : kv)
+      {
+      s.insert(i.first);
+      }
+   return s;
+   }
+
 /*
 * Searching through a std::map
 * @param mapping the map to search
--- a/src/tests/main.cpp
+++ b/src/tests/main.cpp
@@ -16,6 +16,28 @@
 
 namespace {
 
+void print_item_list(std::ostringstream& err, std::set<std::string> list)
+   {
+   size_t line_len = 0;
+
+   for(auto const& item : list)
+         {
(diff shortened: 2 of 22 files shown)


randombit (approved):
I don't like that for categories that consist of a single test, we have to come up with a distinct name. Eg roughtime vs roughtime_tests. It feels like these tests should be categorized as without-a-category. Or alternately, it is a test that is also equal to a category. I guess this isn't a big deal though and handling it would probably complicate the logic.

Anyway this is fine to merge, certainly an improvement over current behavior, just expect to see a followup PR from me with additional changes. :)

reneme:
Frankly I'm not a big fan of renaming those tests either. I did it to disambiguate categories and test names as much as possible. No strong feelings about it either way.

Issue #3216, opened 2023-01-24
Remove raw pointers from the public interface


reneme:
`X509_Object::check_signature(const Public_Key*)` did give us a run for our money yesterday. In some application code we (accidentially) switched from the overload that takes a `const Public_Key&` and suddenly started seeing segfaults. One might say, we should have read the documentation:

```C++
/**
* Check the signature on this data
* @param key the public key purportedly used to sign this data
*        the object will be deleted after use (this should have
*        been a std::unique_ptr<Public_Key>)
* @return true if the signature is valid, otherwise false
*/
```

Should we try and get such things out of the API before 3.0?

Other places where raw pointers without "clear" ownership contracts are used:

* `Public_Key* X509::load_key(...)`
* `Public_Key* X509::copy_key(...)`
* `Private_Key* PKCS8::load_key(...)`
  (raw pointer overloads are deprecated, but could be removed)
* `bool X509_Object::check_signature(const Public_Key*)`
  (takes ownership and deletes the pointer 😨)
* `Public_Key* PKCS10_Request::subject_public_key()`
* `Private_Key* Credentials_Manager::private_key_for()`
* Various semi-internal APIs in the TLS implementations
* probably more

Unfortunately, some of those APIs aren't deprecated, yet. Still, I think we should see what to do with them.

reneme:
See also: https://github.com/randombit/botan/issues/2677

randombit:
For `PKCS8::load_key` we should absolutely remove all the deprecated versions; the ones returning `unique_ptr` have been there since 2.3

Unfortunately `X509::load_key`/`copy_key`  for whatever reason never got the same smart pointer treatment.

`check_signature` we should just change and note it in the migration guide; I expect it's not that common for applications to use this directly.

`PKCS10_Request::subject_public_key` should get the same treatment as `X509_Certificate::subject_public_key`; that function just calls `load_subject_public_key` and releases the pointer. (We should also deprecate the versions returning a raw pointer, currently `X509_Certificate::subject_public_key` is not deprecated)

`Credentials_Manager::private_key_for` not sure if there is an easy way to transition. Maybe simpler to just change the return type and note it in the migration guide.

randombit:
#3220 addressed all of the ones mentioned in your original comment aside from `Credentials_Manager::private_key_for` but there are still several also from #2677 which should be dealt with before 3.0

randombit:
[...]

PR #3220, opened 2023-01-24
Remove various uses of raw pointers in the public API

randombit:
GH #3216

While some of these are changing the return types of non-deprecated functions, I think this is mostly OK; it's easy enough to use either variant as (for example) `std::unique_ptr<Public_Key> key(cert.subject_public_key())` and if it doesn't work at least the compiler gives a hint what is wrong (couldn't convert unique_ptr to raw ptr) vs the function just vanishing.

### DIFF
--- a/src/lib/x509/x509_obj.h
+++ b/src/lib/x509/x509_obj.h
@@ -81,15 +81,6 @@ class BOTAN_PUBLIC_API(2,0) X509_Object : public ASN1_Object
       bool check_signature(const Public_Key& key) const;
 
       /**
-      * Check the signature on this data
-      * @param key the public key purportedly used to sign this data
-      *        the object will be deleted after use (this should have
-      *        been a std::unique_ptr<Public_Key>)
-      * @return true if the signature is valid, otherwise false
-      */
-      bool check_signature(const Public_Key* key) const;
-
-      /**
       * DER encode an X509_Object
       * See @ref ASN1_Object::encode_into()
       */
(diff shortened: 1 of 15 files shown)


src/lib/x509/x509_obj.h
```diff
@@ -80,15 +80,6 @@ class BOTAN_PUBLIC_API(2,0) X509_Object : public ASN1_Object
...
       bool check_signature(const Public_Key& key) const;
 
-      /**
-      * Check the signature on this data
-      * @param key the public key purportedly used to sign this data
-      *        the object will be deleted after use (this should have
-      *        been a std::unique_ptr<Public_Key>)
-      * @return true if the signature is valid, otherwise false
-      */
-      bool check_signature(const Public_Key* key) const;
```
reneme, line 90: Good riddance. 😌

reneme (approved):
Thanks a lot!

src/lib/x509/x509_obj.h
randombit: Yeah a function that deletes its argument as a side effect of doing something else, wtf was I thinking.

src/lib/x509/x509cert.h
```diff
@@ -38,15 +38,15 @@ class BOTAN_PUBLIC_API(2,0) X509_Certificate : public X509_Object
...
+      * certificate. If the public key bits was valid for X.509 encoding
+      * purposes but invalid algorithmically (for example, RSA with an even
+      * modulus) that will be detected at this point, and an exception will be
+      * thrown.
       *
-      * @return public key
+      * @return subject public key of this certificate
       */
-      Public_Key* subject_public_key() const;
+      std::unique_ptr<Public_Key> subject_public_key() const;
```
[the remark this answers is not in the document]
randombit: Makes sense

PR #3221, opened 2023-01-25
TEST: test_cli.py tries additional sig algos for X.509 cert chains

reneme:
This extends `test_cli.py` (`cli_cert_issuance_tests`) to build and verify certificate chains using algorithms other than RSA. Also, it tries to build chains with heterogeneous algorithms.

That was created in an attempt to reproduce https://github.com/randombit/botan/issues/3208#issuecomment-1402253925, but everything seems to work smoothly. Still, I figured, the test is worth adding.


reneme:
Will re-open with a fix and based on #3222.

PR #3222, opened 2023-01-25
FIX: test_cli.py does not fail on 'unknown command'

reneme:
This was found while investigating https://github.com/randombit/botan/issues/3208#issuecomment-1402253925 using `cli_cert_issuance_tests`. I expected the test to fail with Dilithium (as @obalak suggested), but it didn't.

Turns out that there was a subtle bug in the test itself (missing `,`) resulting in the CLI reporting "Unknown Command". The latter was not seen as an error, because it was reported via stdout. Both are fixed here.

### DIFF
--- a/src/cli/main.cpp
+++ b/src/cli/main.cpp
@@ -28,7 +28,7 @@ int main(int argc, char* argv[])
 
    if(!cmd)
       {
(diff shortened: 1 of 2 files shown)


randombit (approved):
Ouch

PR #3223, opened 2023-01-25
FIX: Verify X.509 with Dilithium

### LINKED ISSUE #3208: Using Dilithium as PKI Signer Algo.

reneme:
### Pull Request Dependencies

* #3222 

### Discussion

This fixes an issue reported in #3208 that X.509 certificates signed by Dilithium could not be verified. A regression test is added in `test_cli.py` (`cli_cert_issuance_alternative_algos_tests`).

Dilithium -- just like Ed25519 -- hard-codes the hash function it uses for compressing the message to sign. Internally, Dilithium uses `SHAKE-256(512)`, so I added that as a special case in `X509_Object::hash_used_for_signature()` when Dilithium is used.

Though, that feels a bit dirty for two reasons:

* `::hash_used_for_signature()` now bears internal knowledge of Dilithium (just as it did for Ed25519).
* `SHAKE-256(512)` is not trusted in the path restriction's defaults. Adding it should be fine, but it makes me wonder what to do with the out-length parameter.

Currently, I just added `SHAKE-256(512)` as trusted by default in the `Path_Validation_Restrictions` as a proof of concept. I don't really like it, though.

I think we should find a more general approach to signature algorithms that do not require an EMSA. And find a solution for how things like `X509_Object::hash_used_for_signature()` should behave for such signatures.

I wouldn't be surprised if similar constructions are needed for other protocols that use signatures (e.g. the TLS implementation).

Closes #3208.

### DIFF
--- a/src/cli/main.cpp
+++ b/src/cli/main.cpp
@@ -28,7 +28,7 @@ int main(int argc, char* argv[])
 
    if(!cmd)
       {
-      std::cout << "Unknown command " << cmd_name << " (try --help)\n";
+      std::cerr << "Unknown command " << cmd_name << " (try --help)\n";
       return 1;
       }
 
--- a/src/lib/x509/x509_obj.cpp
+++ b/src/lib/x509/x509_obj.cpp
@@ -140,8 +140,13 @@ std::string X509_Object::hash_used_for_signature() const
    const OID& oid = m_sig_algo.get_oid();
    const std::vector<std::string> sig_info = split_on(oid.to_formatted_string(), '/');
 
-   if(sig_info.size() == 1 && sig_info[0] == "Ed25519")
-      return "SHA-512";
+   if(sig_info.size() == 1)
+      {
+      if(sig_info[0] == "Ed25519")
+         return "SHA-512";
+      else if(sig_info[0].starts_with("Dilithium-"))
+         return "SHAKE-256(512)";
+      }
    else if(sig_info.size() != 2)
       throw Internal_Error("Invalid name format found for " + oid.to_string());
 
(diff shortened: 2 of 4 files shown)


randombit (approved):
:+1: thanks for looking into this

randombit:
> I think we should find a more general approach to signature algorithms that do not require an EMSA. And find a solution for how things like X509_Object::hash_used_for_signature() should behave for such signatures.

There is a lot of weird tangling where we hard code the capabilities of different algorithms into the X509 layer. It would be nice to clean this up somehow. I ran into this also in the [key constraint handling code](https://github.com/randombit/botan/blob/master/src/lib/x509/key_constraint.cpp#L77-L85) while looking at #3211 

Your approach here seems basically sound with regards to what we are doing already so I'd say lets try to get it merged so that things at least basically work, and then revisit.

reneme:
I opened a ticket for this. CI failures were caused by an outage of GitHub Actions, I believe. Trying them again and merging if green...

Issue #3169, opened 2023-01-11
Stablize Supported Platforms for Botan 3.0


reneme:
[Currently supported platforms](https://github.com/randombit/botan/blob/master/doc/support.rst#supported-platforms) are roughly:

* GCC 11+
* Clang 14+
* MSVC 2019+

* latest XCode Clang
* latest NDK

Currently, both XCode and NDK are pushing the breaks on (some) C++20 features. For instance, they are behind with adoption of `<ranges>` and `<concepts>` STL headers. This will likely get better with future versions, as we pledge to only support the latest version of those.

Assuming XCode and NDK caught up, MSVC 2019 is [the most likely culprit to hinder adoption](https://github.com/randombit/botan/pull/3165#issuecomment-1377317013) of new features.

Hence, we might want to put MSVC 2019 on probation and not pledge support for the whole 3.0 life-cycle.

randombit:
We are constrained in multiple directions:

* We can't reasonably require a compiler that's later than what is available in the CI images, since we wouldn't be able to test it.
* We want to support as few old compilers as possible, because they have bugs and limitations that must be worked around.
* Nothing stops us from arbitrarily requiring a new compiler version at any time, when we decide we're tired of doing workarounds. *But* doing so is pretty unfriendly; in my experience toolchain versions tend to be very sticky within a certain project, and when we drop supporting a certain compiler that probably will cut off some number of people from being able to use upgraded versions of the library. And it's also sticky *for us* because we have to finally decide at some point "ok, these hacks really aren't worth it anymore, let's drop 2019 support in the next release"
* In principle we could maintain branches to "backport" security fixes to the final release that supported some certain compiler, in order to prevent anyone from becoming stranded. There is just not the manpower for this.

The options as I see them are:
1) Support VC2019 over the entire release lifetime, which might be many years (currently 2.x has been supported for almost 6 years, with another year of support lifetime already advertised, and we will almost certainly have to extend that)
2) Support VC2019 until we get tired of supporting it
3) Support VC2019 until some specific date (say Jan 1, 2024)
4) Drop support for VC2019 right now, and 3.x never supports 2019. (Since it works right now, the final release 3.0 might well compile even if we dropped support today. But we would no longer test it in CI, and not claim it to be supported.)

I would personally advocate for option 4 because it is overall the most predicable for users and the least work for us:

1) A lot of work over the long haul, and resources are scarce.
2) Unpredictable to users, also places a burden on us in deciding when we've had enough.
[...]

PR #3224, opened 2023-01-25
Drop VC 2019 support

randombit:
GH #3169


randombit:
I'm guessing there are some hacks here and there specifically for handling VC2019 but nothing that's #ifdef'ed on the version, so no code changes.

PR #3225, opened 2023-01-25
Make Key_Constraints an "enum class"-like class

randombit:
#3211

### DIFF
--- a/src/lib/x509/pkix_enums.h
+++ b/src/lib/x509/pkix_enums.h
@@ -165,12 +165,7 @@ class BOTAN_PUBLIC_API(3,0) Key_Constraints
       // Return true if any of the bits provided are set
       bool includes_any(auto&& ...bits) const
          {
-         for(auto bit: std::initializer_list<Key_Constraints::Bits>{ bits... })
-            {
-            if(this->includes(bit))
-               return true;
-            }
-         return false;
+         return (m_value & (bits | ...)) > 0;
          }
 
       bool empty() const { return m_value == 0; }
--- a/src/lib/x509/pkix_enums.h
+++ b/src/lib/x509/pkix_enums.h
@@ -8,7 +8,7 @@
 #define BOTAN_X509_PKIX_ENUMS_H_
 
 #include <botan/types.h>
-
+#include <string>
 namespace Botan {
 
 class Public_Key;
@@ -142,11 +142,7 @@ class BOTAN_PUBLIC_API(3,0) Key_Constraints
 
       Key_Constraints() : m_value(0) {}
 
-      bool operator==(const Key_Constraints& other) const { return m_value == other.m_value; }
-      bool operator!=(const Key_Constraints& other) const { return m_value != other.m_value; }
-
-      bool operator==(Key_Constraints_Bits other) const { return m_value == other; }
-      bool operator!=(Key_Constraints_Bits other) const { return m_value != other; }
+      bool operator==(const Key_Constraints&) const = default;
 
       void operator|=(Key_Constraints_Bits other)
          {
--- a/src/lib/x509/pkix_enums.h
+++ b/src/lib/x509/pkix_enums.h
@@ -1,5 +1,5 @@
 /*
-* (C) 2013 Jack Lloyd
+* (C) 2013,2023 Jack Lloyd
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
@@ -105,7 +105,7 @@ BOTAN_PUBLIC_API(2,0) const char* to_string(Certificate_Status_Code code);
 class BOTAN_PUBLIC_API(3,0) Key_Constraints
    {
    public:
-      enum Key_Constraints_Bits : uint32_t {
+      enum Bits : uint32_t {
          None              = 0,
          DigitalSignature  = 1 << 15,
          NonRepudiation    = 1 << 14,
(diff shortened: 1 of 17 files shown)


src/lib/x509/pkix_enums.h
```diff
@@ -100,18 +102,76 @@ BOTAN_PUBLIC_API(2,0) const char* to_string(Certificate_Status_Code code);
...
+         DIGITAL_SIGNATURE = DigitalSignature,
+         NON_REPUDIATION   = NonRepudiation,
+         KEY_ENCIPHERMENT  = KeyEncipherment,
+         DATA_ENCIPHERMENT = DataEncipherment,
+         KEY_AGREEMENT     = KeyAgreement,
+         KEY_CERT_SIGN     = KeyCertSign,
+         CRL_SIGN          = CrlSign,
+         ENCIPHER_ONLY     = EncipherOnly,
+         DECIPHER_ONLY     = DecipherOnly,
+      };
```
reneme, line 132: Let's split this up into two enums (new and deprecated). This might increase the number of required overloads in the methods and operators below, but allow us to mark the old stuff as deprecated.

Here's a sketch in Godbolt: https://godbolt.org/z/z81r4oGrz

src/lib/x509/pkix_enums.h
```diff
@@ -100,18 +102,76 @@ BOTAN_PUBLIC_API(2,0) const char* to_string(Certificate_Status_Code code);
...
+
+      explicit Key_Constraints(uint32_t bits) : m_value(bits) {}
+
+      Key_Constraints() : m_value(0) {}
+
+      bool operator==(const Key_Constraints& other) const { return m_value == other.m_value; }
+      bool operator!=(const Key_Constraints& other) const { return m_value != other.m_value; }
+
+      bool operator==(Key_Constraints_Bits other) const { return m_value == other; }
+      bool operator!=(Key_Constraints_Bits other) const { return m_value != other; }
```
reneme, line 149: C++20 fanciness:

```suggestion
      bool operator==(const Key_Constraints&) const = default;
```

* [Default comparisons](https://en.cppreference.com/w/cpp/language/default_comparisons)
* [Operator rewriting](https://en.cppreference.com/w/cpp/language/overload_resolution#Call_to_an_overloaded_operator)

The operators taking `Key_Constraints_Bits` can be omitted due to implicit conversion via `Key_Constraints(Key_Constraint_Bits)`.

src/lib/x509/pkix_enums.h
```diff
@@ -100,18 +102,76 @@ BOTAN_PUBLIC_API(2,0) const char* to_string(Certificate_Status_Code code);
...
+
+      void operator|=(Key_Constraints_Bits other)
+         {
+         m_value |= other;
+         }
```
reneme, line 154: Maybe implement that in terms of implicit conversion as well. When splitting the enum into "deprecated" and "new", this would avoid an overload.

```suggestion
      void operator|=(const Key_Constraints& other)
         {
         m_value |= other.m_value;
         }
```

Though, it sort-of widens the scope of the operator in ways that we might not want.

src/lib/x509/x509_ca.cpp
```diff
@@ -81,22 +81,25 @@ Extensions choose_extensions(const PKCS10_Request& req,
...
-      constraints = Key_Constraints(KEY_CERT_SIGN | CRL_SIGN);
+      constraints |= Key_Constraints::KeyCertSign;
+      constraints |= Key_Constraints::CrlSign;
       }
    else
       {
-      std::unique_ptr<Public_Key> key(req.subject_public_key());
-      verify_cert_constraints_valid_for_key_type(*key, req.constraints());
       constraints = req.constraints();
       }
```
reneme, line 90: With a few free-standing `operator|` this could be spiced up to:

```C++
const auto constraints =
  (req.is_CA())
  ? Key_Constraints::KeyCertSign | Key_Constraints::CrlSign
  : req.constraints();
```

My main objective here: Make the class "feel" like any enum type, despite the added cleverness. I.e. you'd be able to write this and still get the 'clever' type:

```C++
const auto constraints = Key_Constraints::KeyCertSign | Key_Constraints::CRL_SIGN;
BOTAN_ASSERT(!constraints.empty());
```

Godbolt: https://godbolt.org/z/oWdxKKqGf

src/lib/x509/x509_ext.h
```diff
@@ -70,7 +72,7 @@ class BOTAN_PUBLIC_API(2,0) Key_Usage final : public Certificate_Extension
       std::string oid_name() const override { return "X509v3.KeyUsage"; }
 
       bool should_encode() const override
-         { return (m_constraints != NO_CONSTRAINTS); }
+         { return (m_constraints != Key_Constraints::None); }
```
reneme, line 75:
```suggestion
         { return !m_constraints.empty(); }
```

src/lib/x509/x509self.cpp
```diff
@@ -70,19 +70,22 @@ X509_Certificate create_self_signed_cert(const X509_Cert_Options& opts,
...
       {
-      constraints = Key_Constraints(KEY_CERT_SIGN | CRL_SIGN);
+      constraints |= Key_Constraints::KeyCertSign;
+      constraints |= Key_Constraints::CrlSign;
       }
    else
       {
-      verify_cert_constraints_valid_for_key_type(key, opts.constraints);
       constraints = opts.constraints;
       }
```
reneme, line 79: See suggestion re: "free-standing `operator|`" above

src/lib/x509/x509self.cpp
```diff
@@ -119,22 +122,26 @@ PKCS10_Request create_cert_req(const X509_Cert_Options& opts,
...
```
reneme, line 131: likewise.

reneme (commented):
I very much like the added cleverness by the convenience methods. 👍
Just a few suggestions to make it even smarter and benefit from C++20.

src/lib/x509/pkix_enums.h
```diff
@@ -11,6 +11,8 @@
```
reneme, line 11:
```suggestion
#include <string>
```

Appeasing the MSVC build.

src/lib/x509/pkix_enums.h
randombit: Nice :+1:

src/lib/x509/pkix_enums.h
randombit: Very nice thank you

src/lib/x509/pkix_enums.h
randombit: I discovered there is a way that is (IMO) cleaner:

```
enum E {
  Something = 0,

  SOME_LOUD_THING [[deprecated("Use Something")]] = Something,
}
```

warns in exactly the way we would want, and without having to create a second type.

src/lib/x509/x509_ca.cpp
randombit: I went another route here, adding a `ca_constraints()` helper that returns KeyCertSign | CrlSign. That neatly solves every single case we used `|` operator outside of tests and allows use of the ternary operator as in your example.

IMO the fact that a "real" enum, `E::Val1 | E::Val2` does not result in an `E` but rather an integer, and you need to say `E(E::Val1 | E::Val2)`, it seems to me fine to require similar here.

src/lib/x509/pkix_enums.h
```diff
@@ -100,18 +102,92 @@ BOTAN_PUBLIC_API(2,0) const char* to_string(Certificate_Status_Code code);
...
+      // Return true if any of the bits provided are set
+      bool includes_any(auto&& ...bits) const
+         {
+         for(auto bit: std::initializer_list<Key_Constraints::Bits>{ bits... })
+            {
+            if(this->includes(bit))
+               return true;
+            }
+         return false;
+         }
```
reneme, line 174: Nice, I didn't know about `std::initializer_list<>` like that. However, I think that's a perfect case for C++17 fold expressions:

```suggestion
      bool includes_any(auto&& ...bits) const
         {
         return (m_value & (bits | ...)) > 0;
         }
```

reneme (approved):
One more litte C++-feature nit.

PR #3228, opened 2023-01-26
hw rng attempt to read value from arm64 devices.

devnexen:


src/lib/rng/processor_rng/processor_rng.cpp
```diff
@@ -111,6 +130,8 @@ bool Processor_RNG::available()
    {
 #if defined(BOTAN_TARGET_CPU_IS_X86_FAMILY)
    return CPUID::has_rdrand();
+#elif defined(BOTAN_TARGET_ARCH_IS_ARM64) && !defined(BOTAN_TARGET_OS_IS_MACOS) && defined(BOTAN_TARGET_IS_IOS)
```
randombit, line 133: I don't see any reason to here (or elsewhere in the file) guard against macOS/iOS - if CPUID doesn't indicate support, the code will never be invoked, and the only way CPUID can detect support is with getauxval which doesn't exist on Apple.

randombit:
You'll also need to update `processor_rng/info.txt` to mark that the code can compile on aarch64

src/lib/rng/processor_rng/processor_rng.cpp
devnexen: in fact it s more because the assembly does not compile on these.

randombit: Oh that's unfortunate.

Maybe instead we can use `__builtin_aarch64_rndrrs` / `__builtin_arm_rndrrs` builtins? And then only enable support if the builtins exist.

devnexen: ah good point

randombit:
Seems like this is stalled - should we close?

devnexen:
> Seems like this is stalled - should we close?

It s ready for review.

randombit:
Seems OK but needs to be rebased to address the conflict

randombit:
```
src/lib/rng/processor_rng/processor_rng.cpp: In member function ‘virtual void Botan::Processor_RNG::fill_bytes_with_input(std::span<unsigned char>, std::span<const unsigned char>)’:
src/lib/rng/processor_rng/processor_rng.cpp:173:1: error: unrecognizable insn:
  173 | }
      | ^
(insn 55 54 56 10 (parallel [
            (set (reg:DI 132)
                (unspec_volatile:DI [
                        (const_int 0 [0])
                    ] UNSPEC_RNDRRS))
            (set (reg:CC_Z 66 cc)
                (unspec_volatile:CC_Z [
                        (const_int 0 [0])
                    ] UNSPEC_RNDRRS))
        ]) "src/lib/rng/processor_rng/processor_rng.cpp":83:37 -1
     (nil))
during RTL pass: vregs
src/lib/rng/processor_rng/processor_rng.cpp:173:1: internal compiler error: in extract_insn, at recog.c:2770
0x185906c internal_error(char const*, ...)
	???:0
0x6adbf9 fancy_abort(char const*, int, char const*)
	???:0
0x68d0ed _fatal_insn(char const*, rtx_def const*, char const*, int, char const*)
	???:0
0x68d10f _fatal_insn_not_found(rtx_def const*, char const*, int, char const*)
	???:0
```

We may need a newer compiler here?

randombit:
IIUC the situation here is at best finnicky, and maybe it only works reliably if we are running in kernel mode. It seems even mbedtls is hesitant to support this (https://github.com/Mbed-TLS/mbedtls/issues/7509) so I think until the situation is more clear - or at least until we can reliably test this - I'm going to close.

PR #3233, opened 2023-01-28
compress utils overflow check prior to allocation.

devnexen:


src/lib/compression/compress_utils.cpp
```diff
@@ -13,7 +13,17 @@ namespace Botan {
...
 void* Compression_Alloc_Info::do_malloc(size_t n, size_t size)
    {
-   // TODO maximum length check here?
+   size_t t;
+#if BOTAN_COMPILER_HAS_BUILTIN(__builtin_mul_overflow)
+   if(__builtin_mul_overflow(n, size, &t)) [[unlikely]]
+#else
+   t = n * size;
+   if(size && t / size != n) [[unlikely]]
+#endif
```
randombit, line 22: Would be better to handle this as a new function in `safeint.h` along the lines of

`std::optional<size_t> checked_mul(size_t x, size_t y)`

returning `nullopt` if the result would overflow.

There are a couple of similar overflow checking multiplications in `locking_allocator.cpp` btw

devnexen: Yes I though of that hesitated because we re not supposed to throw exceptions here but I will update

randombit (approved):

PR #3234, opened 2023-01-29
Make CPUID data thread local

randombit:
If this is possible then we can avoid serializing most of the tests. IIRC it caused problems with MSVC but lets see if it is still an issue with 2022.


reneme (approved):
CI seems to work. Don't we need to apply additional measures when `thread_local` is not available for the platform, though?

randombit:
@reneme If CPUID data is thread local, we can remove a lot of the serialization from the test suite. But I will look into this more later, now that I know VC2022 is ok with thread local in DLL-exported symbols.

Issue #3146, opened 2022-12-29
Convert the remaining TLS enums to enum class


randombit:
Deferring this to avoid creating too many conflicts for the 1.3 work

- [x] `Record_Type`
- [x] `Connection_Side`
- [x] `Handshake_Type`
- [x] `Protocol_Version::Version_Code`
  (technically that is already wrapped forming a "smart enum")
- [x] `Cipher_State::PSK_Type`
  (already is an `enum class` but doesn't adhere to the style conventions)

reneme:
Grepping through the `tls` folder, I believe, this is done now. There are some enums in `tls_algos.h` that are still all-caps, but those are algorithm lists where most names are abbreviations anyway. So, I guess, that's fine.

Thanks. 😄

PR #3235, opened 2023-01-30
Make TLS::Connection_Side an enum class

randombit:
GH #3146


reneme (approved):
Wow! This enum is truely used everywhere! 😲
(I added a patch for the Examples and Fuzzers)

randombit:
Thanks!

PR #3237, opened 2023-01-31
Make TLS::Record_Type an enum class

randombit:
This also merges `NO_RECORD` with the TLS 1.3 `Invalid` enum value, since my read of TLS 1.3 is that this is reservation is done explicitly so that implementations can use 0 as the invalid/`None` record type instead of having to use an out of range value (as `NO_RECORD` was) or optional type.

#3146


src/lib/tls/tls_magic.h
```diff
@@ -50,18 +50,15 @@ enum class Connection_Side {
...
-
-   CHANGE_CIPHER_SPEC = 20,
-   ALERT              = 21,
-   HANDSHAKE          = 22,
-   APPLICATION_DATA   = 23,
+enum class Record_Type: uint8_t {
+   Invalid            = 0,  // RFC 8446 (TLS 1.3)
 
-   HEARTBEAT          = 24, // RFC 6520 (TLS 1.3)
+   CCS                = 20,
```
reneme, line 56: Might be a nit, but I'd prefer to expand the abbreviation as `ChangeCipherSpec` for readability.

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -526,19 +526,24 @@ void Channel_Impl_12::write_record(Connection_Cipher_State* cipher_state, uint16
...
+                                    static_cast<uint8_t>(record_type),
+                                    record_version, next_seq,
                                     input, length);
       }
    else
       {
-      TLS::write_record(m_writebuf, record_type, record_version, next_seq,
+      TLS::write_record(m_writebuf,
+                        static_cast<uint8_t>(record_type),
+                        record_version, next_seq,
```
reneme, line 538: Why not adapt `write[_unprotected]_record` with the `Record_Type` parameter as well?

src/lib/tls/tls_magic.h
randombit: Went back and forth on this one myself. Will change to `ChangeCipherSpec`.

src/lib/tls/tls12/tls_channel_impl_12.cpp
randombit: I was lazy ;) done now

src/lib/tls/tls_magic.h
randombit: Renamed

PR #3246, opened 2023-02-01
Modify TLS Alert enum names

randombit:
A handful of alert types that it seemed likely applications would special case on (like close notify and protocol version problems) have the old names retained as deprecated variants. For the rest only the new name is available.


randombit:
Been a while since I've seen GCC ICE

src/lib/tls/tls_alert.h
```diff
@@ -15,60 +15,69 @@ namespace Botan {
...
-         UNKNOWN_PSK_IDENTITY            = 115,
-         CERTIFICATE_REQUIRED            = 116, // RFC 8446
-
-         NO_APPLICATION_PROTOCOL         = 120, // RFC 7301
-
-         // pseudo alert values
-         NULL_ALERT                      = 256
-      };
+      typedef AlertType Type;
+      using enum AlertType;
```
reneme, line 75: It seems that `using enum` is not understood by MinGW and the baremetal job:

```
build/include/botan/tls_alert.h:75:13: error: expected nested-name-specifier before ‘enum’
   75 |       using enum AlertType;
      |             ^~~~
```

Why not stay with the current construction of having a plain enum nested inside `class Alert`? Isn't that the same "smart-enum" idiom we used in other places?

src/lib/tls/tls_exceptn.h
```diff
@@ -42,7 +42,7 @@ class BOTAN_PUBLIC_API(2,0) Unexpected_Message final : public TLS_Exception
    {
    public:
       explicit Unexpected_Message(const std::string& err) :
-         TLS_Exception(Alert::UNEXPECTED_MESSAGE, err) {}
+         TLS_Exception(AlertType::None, err) {}
```
reneme, line 45: Not sure how this happened, but it broke the ASIO stream test.

```suggestion
         TLS_Exception(AlertType::UnexpectedMessage, err) {}
```

src/lib/tls/tls_alert.h
reneme: Also, my gut feeling is that this is why GCC is ICE-ing. Which it still does, by the way.

For instance in the GCC sanitizer job:

```
/home/runner/work/botan/botan/build/build/include/botan/asio_stream.h:628:47: internal compiler error: in tsubst_copy, at cp/pt.c:16822
  628 |                if(alert.type() == TLS::Alert::CloseNotify)
      |                                   ~~~~~~~~~~~~^~~~~~~~~~~
[...]
```

reneme (changes requested):

src/lib/tls/tls_alert.h
randombit: > Why not stay with the current construction of having a plain enum nested inside class Alert? Isn't that the same "smart-enum" idiom we used in other places?

It is, but that's only because I just found out about `using enum` in C++20. If it works I plan to modify at least some of the other enumerations similarly. Having the enumeration being an enum class prevents certain errors, for example adding two enumerations is valid for plain `enum` but invalid for `enum class`.

src/lib/tls/tls_exceptn.h
randombit: I think I edited that one manually instead of via a script

src/lib/tls/tls_alert.h
randombit: > It seems that using enum is not understood by MinGW and the baremetal job:

It appears both of these are actually GCC 10!

randombit: > Which it still does, by the way.

I missed the case in asio that triggers the ICE

randombit:
Merged the wrong commit - this has landed

PR #3250, opened 2023-02-03
Fix removal of the wrong CI job

randombit:
The GCC build on Windows is ok (GCC 11) while Ubuntu is stuck with a GCC 10 MinGW

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -229,9 +229,10 @@ jobs:
           - target: cross-android-arm64
             compiler: clang
             host_os: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


devnexen:
off-topic a bit but do you think it would be worth trying hwasan on android (possibly linux too) on arm64 instead of asan ?

randombit:
Don't think it would help much considering our ARM "builders" are just cross compilers + qeum. If it even worked, it would certainly be slower than ASan on x86-64.

reneme (approved):
I suppose this will fix https://github.com/randombit/botan/pull/3246, right?

PR #3255, opened 2023-02-05
c++ 20 introducing bit_cast usage.

devnexen:


src/lib/utils/timer.cpp
```diff
@@ -84,7 +84,7 @@ std::string Timer::result_string_bps() const
    {
    const size_t MiB = 1024 * 1024;
 
-   const double MiB_total = static_cast<double>(events()) / MiB;
+   const double MiB_total = std::bit_cast<double>(events()) / MiB;
```
randombit, line 87: This (and the similar cases in this file) seems wrong to me - we want an integer value converted to a double value, not a double with the bit pattern of the integer

randombit:
Closing this as inactive / needing work. We can look into `bit_cast` more, I think in particular we *do* want this in cases where we are currently using `memcpy` for this, as IIUC that's the whole point behind `std::bit_cast`.

PR #3258, opened 2023-02-07
Change Credentials_Manager::private_key_for to return a shared_ptr

randombit:
The previous design introduced serious ownership management problems because there was no way for the Credentials_Manager to know when (or even if) the TLS layer had dropped the pointer, so it had to be kept alive forever.

### DIFF
--- a/src/tests/test_tls_rfc8448.cpp
+++ b/src/tests/test_tls_rfc8448.cpp
@@ -63,23 +63,6 @@ void add_entropy(Botan_Tests::Fixed_Output_RNG& rng, const std::vector<uint8_t>&
    rng.add_entropy(bin.data(), bin.size());
    }
 
-Botan::Private_Key* server_private_key()
-   {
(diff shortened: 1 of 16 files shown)


reneme (approved):
`std::shared_ptr<>` certainly communicates the intent and ownership expectations much better than the old raw pointer. Certainly not a must, but maybe also change `TLS::Callbacks::tls_sign_message()` to accept the `std::shared_ptr<>` for consistency?

src/tests/test_tls_rfc8448.cpp
```diff
@@ -1847,6 +1857,7 @@ class Test_TLS_RFC8448_Server : public Test_TLS_RFC8448
 
             Botan_Tests::CHECK("Expect Server close_notify", [&](Test::Result& result)
                {
+               BOTAN_ASSERT_NONNULL(ctx);
```
reneme, line 1860: Thanks!

PR #3263, opened 2023-02-08
Refactor: Make xmss_wots.h and xmss_hash.h internal

reneme:
This moves some implementation details of `XMSS_PrivateKey` out of the public interface. With that `xmss_wots.h` and `xmss_hash.h` can be internal headers (and not marked as public API classes anymore).

I'm hoping to reduce the public cross section of the XMSS implementation further in later PRs. Particularly, I don't really see a reason to expose implementation details with public getters **and setters** 😲.

To me, parts of the implementation (particularly the WOTS public/private key handling) seem more complicated than they need to be. Here is a first step towards cleaning this up.

Later, we're planning to adapt the XMSS key generation to the recommendations in [NIST SP.800-208](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-208.pdf) -- which should be backward compatible even for existing private keys generated by Botan 2.x and add the parameter sets described in the same document.

Eventually, an implementation for XMSS-MT is planned, but I wouldn't want to commit that for Botan 3.0.

### DIFF
--- a/src/lib/pubkey/xmss/xmss_privatekey.cpp
+++ b/src/lib/pubkey/xmss/xmss_privatekey.cpp
@@ -19,6 +19,8 @@
 #include <botan/internal/xmss_signature_operation.h>
 #include <botan/internal/xmss_index_registry.h>
 #include <botan/internal/xmss_common_ops.h>
+#include <botan/internal/loadstor.h>
+#include <botan/internal/stl_util.h>
 #include <botan/ber_dec.h>
 #include <botan/der_enc.h>
 #include <iterator>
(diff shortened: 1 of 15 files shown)


randombit (approved):
Thank you, these headers being public bugged me

Also looking at these getters/setters, WTF ... if we can remove these I'll be happy.

randombit:
The CI failures look relevant, though I don't understand why it is failing only on the coverage build

reneme:
> The CI failures look relevant, though I don't understand why it is failing only on the coverage build

Yes, they are reproducible locally when running `test_cli.py --run-slow-tests`. The slow tests are run exclusively for the coverage build. Should we maybe enable them for "sanitizers" as well?

src/lib/pubkey/xmss/xmss_privatekey.cpp
```diff
@@ -32,105 +34,201 @@ namespace Botan {
...
-                     public_seed(),
-                     rng),
-     m_hash(xmss_hash_function()),
-     m_prf(rng.random_vec(XMSS_PublicKey::m_xmss_params.element_size())),
-     m_index_reg(XMSS_Index_Registry::get_instance())
+   : XMSS_PublicKey(xmss_algo_id, rng)
+   , m_private(std::make_shared<XMSS_PrivateKey_Internal>(
+      m_xmss_params, m_wots_params, m_public_seed,
+      rng.random_vec(m_xmss_params.element_size()),
+      rng.random_vec(m_xmss_params.element_size())))
```
reneme, line 212: Re: CLI test failure: This reverses the order that secret seed and the PRF value are drawn from the RNG. The `cli_xmss_sign_tests` test verifies the hash of the private key given a fixed RNG-seed and therefore failed.

We could switch out the parameters in the `XMSS_PrivateKey_Internal` constructor to avoid changing the expecations of the test. I did that locally to make sure the test would turn green: ✔️
Though, I think, a sensible order of parameters trumps changing the test.

randombit: OK I've given this some consideration and I think we should *not* change how the key is generated, and instead swap the order of the parameters.

In general, right now we do not guarantee that a particular RNG state will lead to a particular private key, if you re-run key generation. We do implicitly guarantee it for Kyber and Dilitihium, and probably will also for most of the other new PQC algorithms simply because of how the KATs are generated. This is largely a problem of underspecification for older algorithms.

I want to start moving in that direction by adding explicit tests (#3269) that check this behavior, without guaranteeing long term stability to end users (yet). That is, if we really decide to change how ECDSA keygen is done, we can do so, and just also update the tests. But eventually we would make such a guarantee, and the test data then becomes set in stone.

reneme: Generally speaking: I'm fully in favor of this idea. Specifically for XMSS, though: Currently the key generation in Botan is not compliant to the suggestions in NIST SP.800-208 anyway. TL;DR: They introduce an additional PRF dedicated to key-generation to protect against a multi-target attack. Currently, Botan's implementation adheres to the recommendations in [RFC 8391](https://www.rfc-editor.org/rfc/rfc8391) where the specifics of key generation are not set in stone.

Long story short: When changing the key generation to the NIST spec, the relationship between RNG state and resulting private key will break for XMSS anyway. Note though, that the change in the key generation would not invalidate legacy private keys and/or signatures.

randombit: OK given above situation, fine to proceed.

PR #3264, opened 2023-02-08
Rename Public_Key::get_oid to object_identifier

randombit:


reneme (approved):
I guess this is a step towards consistent naming of getters without the `get_` prefix?

randombit:
Exactly so, though that will be a long road. `get` only makes sense when there is a `set` and in most contexts we don't have a set.

PR #3265, opened 2023-02-08
Deprecate get_ prefix functions on AlgorithmIdentifier

randombit:
And update callers within the library


src/lib/asn1/asn1_obj.h
```diff
@@ -444,7 +444,10 @@ class BOTAN_PUBLIC_API(2,0) AlgorithmIdentifier final : public ASN1_Object
       const OID& oid() const { return m_oid; }
```
reneme, line 444: For consistency with #3264: shouldn't this be `object_identifier()`?

reneme (approved):

src/lib/asn1/asn1_obj.h
randombit: That would be better. I'll rework this.

PR #3266, opened 2023-02-08
Add Asymmetric_Key base class

randombit:
This will help with transitioning a Public_Key / Private_Key split

### DIFF
--- a/src/lib/pubkey/pk_keys.cpp
+++ b/src/lib/pubkey/pk_keys.cpp
@@ -14,6 +14,19 @@
 
 namespace Botan {
 
+const BigInt& Asymmetric_Key::get_int_field(const std::string& field) const
+   {
+   throw Unknown_PK_Field_Name(algo_name(), field);
+   }
+
+OID Asymmetric_Key::object_identifier() const
+   {
+   OID o = OIDS::str2oid_or_empty(algo_name());
+   if(o.empty())
+      throw Lookup_Error("PK algo " + algo_name() + " has no defined OIDs");
+   return o;
+   }
+
 std::string create_hex_fingerprint(const uint8_t bits[],
                                    size_t bits_len,
                                    const std::string& hash_name)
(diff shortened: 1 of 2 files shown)


reneme:
I'm guessing you're planning to go from this:

PlantUML model

to this, eventually. To resolve the diamond-shape multiple inheritance under `Public_Key`.



I'm not quite sure what the `Public_Key_Algorithm` abstraction brings to the table, though. Sure, you can get basic algorithm characteristics from it, but I don't see a way to implement algorithmic abstractions based on it. Compared to the virtual interface that (say) `Hash_Function` offers. Don't get me wrong, I don't oppose having such a base class, but **I would suggest to rename it to `Asymmetric_Key` which fits the modeled concept better than `Public_Key_Algorithm`** IMO.

In a `Public_Key_Algorithm` I would expect methods like `sign`, `encrypt`, `verify`, `encapsulate_key` and so forth. But that cannot be nicely modeled with an inheritance hierarchy. This could be a high-level convenience wrapper and model a composition of `Public_Key` and/or `Private_Key`.

Below is a sketch. I'm on the fence whether such a thing would be worth adding, though. Abstact algorithms may (and probably should) directly take `Public_Key*` and/or `Private_Key*` as arguments and not deal with the state mess such a composition class would impose on them. Just my two cents, about what my intuition on a `Public_Key_Algorithm` class would be.

```C++
using Signature = Strong<std::vector<uint8_t>, struct Signature_>;

class Public_Key_Algorithm
{
  public:
    Public_Key_Algorithm(std::unique_ptr<Public_Key>) {}
    Public_Key_Algorithm(std::unique_ptr<Private_Key>) {}
    Public_Key_Algorithm(std::unique_ptr<Public_Key>, std::unique_ptr<Private_Key>) {}

    PK_Signer make_signer(RandomNumberGenerator& rng) {
        BOTAN_STATE_CHECK(m_private && m_private->can_sign());
        return PK_Signer(m_private, rng, m_private->default_emsa());
    }

    Signature sign(RandomNumberGenerator& rng, std::span<const uint8_t> message) {
        return make_signer(rng).sign_message(message);
    }

  private:
    std::unique_ptr<Public_Key> m_public;
    std::unique_ptr<Private_Key> m_private;
};
```

randombit:
Your guess as to the intended eventual hierarchy is spot on.

> you can get basic algorithm characteristics from it, but I don't see a way to implement algorithmic abstractions based on it.

Right, that's not the goal. The primary reason for this class, *and* the reason for including it now, in 3.0, when this split can't possibly happen before 4.0, is to give applications which need a common base class for keys something to use. Right now `Public_Key*` works for that, but when `Private_Key` doesn't derive from `Public_Key` that doesn't work.

(Arguably `std::variant` over `Public_Key` + `Private_Key` would be a better approach to the problem, but the goal here is to allow an easy as possible transition of code doing something that works right now, to how things "should be" in the future)

Your suggested name does better capture the notion expressed by the class.

reneme (approved):

PR #3269, opened 2023-02-10
Add support for key generation stability tests

randombit:
Right now we don't claim in the docs that keygen is stable, for any algorithm. But I'd like to fix this eventually. In the meantime, explicit stability tests provide a guardrail to detect otherwise unintentional changes.


src/tests/test_ecdsa.cpp
```diff
@@ -162,6 +162,13 @@ class ECDSA_Keygen_Tests final : public PK_Key_Generation_Test
          }
    };
 
+class ECDSA_Keygen_Stability_Tests final : public PK_Key_Generation_Stability_Test
```
reneme, line 165:
```suggestion
/**
 * Ensures that equal RNG input leads to equivalent key pairs across library
 * versions. We try our best to ensure key generation stability.
 */
class ECDSA_Keygen_Stability_Tests final : public PK_Key_Generation_Stability_Test
```

Note to self?

reneme (approved):
LGTM. Maybe leave a comment to future contributors that end up breaking such a test?

PR #3270, opened 2023-02-10
Reduce the size of Dilithium and Kyber test data

randombit:
For Dilithium, store just H(sig) instead of the signature

For Kyber, generalize the CTR_DRBG implementation we already had for Dilithium, and use it. This allows removing the precomputed CTR_DRBG outputs. (I imagine this will come in handy for some of the future PQC work as well.)

This saves a bit over 8000 KiB in `src/tests/data`

I'm going to follow up on this one by also modifying the Kyber tests to just store `H(pk),H(sk),H(ct)` instead of the actual values, since after this change Kyber is 30% of the entire pubkey test data. But this PR is already big enough.


src/tests/test_dilithium.cpp
```diff
@@ -25,193 +25,60 @@
...
 
          Botan::Dilithium_PublicKey pub_key(priv_key.public_key_bits(), DerivedT::mode, Botan::DilithiumKeyEncoding::Raw);
          auto verificator = Botan::PK_Verifier(pub_key,"");
          verificator.update(ref_msg.data(), ref_msg.size());
-         result.confirm("Signature Verification", verificator.check_signature(signature.data(), signature.size()));
+         result.confirm("signature verifies",
+                        verificator.check_signature(signature.data(), signature.size()));
 
-         // wrong signagture
+         // test validating incorrect wrong signagture
```
reneme, line 78:
```suggestion
         // test validating incorrect wrong signature
```

reneme (approved):
I like! Certainly, this will be helpful for other PQC KATs.

Issue #3268, opened 2023-02-10
Does Botan 1.10.17 support AES/GCM cipher


adnan-kamili:
We have a legacy app where we use Botan 1.10.17 (can't use newer version of Botan as they don't support older Linux Distros). We use amalgamation build and wanted to use AES/GCM ciper. I tried following:

```
python3 configure.py --cc=clang --os=darwin --cpu=arm64 --no-autoload --gen-amalgamation --enable-modules=rsa,eme_pkcs,emsa3,emsa4,sha1,auto_rng,cbc,kdf2,asm_engine,pthreads,unix_procs,pbes2,sha2_64,des,
gcm,aead
```

But is gives a warning:

```
WARNING: Unknown enabled module aead
WARNING: Unknown enabled module gcm
```

randombit:
Sorry no, GCM was added in 1.11.3

> can't use newer version of Botan as they don't support older Linux Distros

Just out of curiosity what distro/gcc are you using that doesn't have GCC 4.8?? Even RHEL7 has that  

Also FWIW a old distro should be fine as long as the compiler is updated.

adnan-kamili:
We have to support Debian 7 with glibc v2.13 and gcc 4.7, and as per some previous discussion I remember Botan 2 doesn't work with this configuration.

randombit:
You might be able to get away with compiling 1.11.3, or even a somewhat later version, with GCC 4.7. 1.11.0 was released before GCC 4.8, and the change notes claim it worked with GCC 4.7. 1.11.3 was released only a couple of weeks after GCC 4.8 was. I don't see in the notes where exactly we dropped GCC 4.7 support along the way, and I certainly don't remember since this was literally a decade ago, but you might get lucky. Especially if you're able to disable features you're not otherwise using. TLS was seeing very active work at that time and I imagine the first uses of GCC 4.8-only features was there.

If you decide to try this I really can't offer much in the way of support, 1.10 has been EOL for over 4 years and officially speaking 1.11 were development snapshots and never supported. But something to try if you're really stuck here.

... further digging ...

You may be in luck here, 30edf16ca6fabc22f mentions GCC 4.7, which means it may have worked as late as 1.11.27 (or even later, who knows).

[...]

PR #3273, opened 2023-02-11
Patches to enable GCC 4.7 to compile Botan2

randombit:
Wasn't so bad actually.

Have only tested this on a MIPS box (only machine I can find that still has GCC 4.7 installed) so there may be little problems in x86-specific code.

#3268


reneme:
Off-topic: Apparently `release-2` does not have the same branch protections as `master` (i.e. I can merge without a review). I would assume that's not intentional.

randombit:
It was not. Thanks for mentioning. I just updated the branch protection rules.

PR #3276, opened 2023-02-13
Remove Accessors to XMSS internals

reneme:
### Pull Request Dependencies

* #3263

### Description

This removes (or at least hides) accessors to XMSS algorithm internals from the public interface. Also, it deprecates `XMSS_PrivateKey::unused_leaf_index()` in favor of a new and more user-friendly `::remaining_signatures()` method.

Finally, it adds a test for `remaining_signatures()` to ensure that no more signatures can be generated once the WOTS private keys in the XMSS tree are exhausted.

### DIFF
--- a/src/lib/pubkey/xmss/info.txt
+++ b/src/lib/pubkey/xmss/info.txt
@@ -8,8 +8,6 @@ name -> "XMSS"
 
 <header:public>
 xmss.h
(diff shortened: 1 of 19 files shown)


randombit (approved):
Great cleanups here, thank you

PR #3277, opened 2023-02-14
Chore: Simplify (internal)WOTS Public/Private Key Implementation

reneme:
### Pull Request Dependencies

* #3276
* #3263

### Description

Before, the WOTS public and private key had two roles. Namely to hold the the long-lived private/public seeds and to be able to derive short-lived concrete WOTS keys from those seeds. This didn't quite make sense. Now only the latter is implemented here, while the seeds are kept upstream with the XMSS public/private keys.

Additionally, there was a lot of boiler plate accessors and a number of constructors that didn't initialize the objects fully and were dependent on the call sites to set the missing fields correctly.

Finally, WOTS public and private key classes were derived from the `Public_*` and `Private_Key` base classes. Albeit technically correct, this pulled in override obligations and usage quirks unneccesarily.

This PR alone gets rid of roughly 450 lines of code. 🎈


randombit (approved):
:tada:

PR #3280, opened 2023-02-14
Implement XMSS PRF_keygen as in NIST SP.800-208 Section 6.2

reneme:
### Pull Request Denpendencies

* #3283 

### Description

This adapts the XMSS keygen to be compliant to [NIST SP.800-208](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-208.pdf). It introduces another `XMSS_Hash` mode `prf_keygen` that it used exclusively for deriving WOTS keys. Additionally, the WOTS private key now also depends on the public seed.

I updated the affected test vectors (`xmss_sig.vec` and the CLI test) and added a dedicated "keygen reference" test. The latter gets the verbatim `public_seed`, `private_seed` and `prf` values, sticks them (correctly ordered) into a `Fixed_Output_RNG` and checks that the resulting key pair is equal to [the reference implementation](https://github.com/XMSS/xmss-reference). There's also a new helper script to reproduce said test vectors.

The test data in `xmss_sig.vec` with an XMSS tree height > 10 was commented out on `master`. Certainly, to keep the test run time managable. (A single signature-test with tree height = 20 took about 3 minutes with multithreading on my laptop.) Nevertheless, I re-created the full test vector and commented out the same configurations. This results in roughly 450kB of mostly commented-out test data. Doesn't feel super useful to me, though. What do you think?


src/lib/pubkey/xmss/xmss_wots.cpp
```diff
@@ -142,17 +144,17 @@ wots_keysig_t XMSS_WOTS_PrivateKey::sign(const secure_vector<uint8_t>& msg,
...
    {
-   const auto r = hash.prf(private_seed, adrs.bytes());
-
    for(size_t i = 0; i < m_params.len(); ++i)
       {
-      XMSS_Tools::concat<size_t>(m_key_data[i], i, 32);
-      hash.prf(m_key_data[i], r, m_key_data[i]);
+      adrs.set_chain_address(i);
+      const auto data = concat_as<std::vector<uint8_t>>(public_seed, adrs.bytes());
+      hash.prf_keygen(m_key_data[i], private_seed, data);
```
reneme, line 157: This is the actual keygen change.

reneme:
Test failures are relevant. I will look into it tomorrow.

src/lib/pubkey/xmss/xmss_hash.h
```diff
@@ -61,6 +61,50 @@ class XMSS_Hash final
          return m_hash->final();
          }
 
+      /**
+       * Pseudoranom function creating a hash out of a key and data using
```
randombit, line 65: Typo - Pseudorandom

src/lib/pubkey/xmss/xmss_hash.h
```diff
@@ -61,6 +61,50 @@ class XMSS_Hash final
...
+         secure_vector<uint8_t> result;
+         m_hash->update(m_zero_padding);
+         m_hash->update(m_id_prf_keygen);
+         m_hash->update(key.data(), key.size());
+         m_hash->update(data.data(), data.size());
+         return m_hash->final();
+         }
+
+      /**
+       * Pseudoranom function creating a hash out of a key and data using
```
randombit, line 87: Pseudorandom

randombit (approved):

reneme:
Any thoughts on this (from the PR description)? @randombit 

> The test data in xmss_sig.vec with an XMSS tree height > 10 was commented out on master. Certainly, to keep the test run time managable. (A single signature-test with tree height = 20 took about 3 minutes with multithreading on my laptop.) Nevertheless, I re-created the full test vector and commented out the same configurations. This results in roughly 450kB of mostly commented-out test data. Doesn't feel super useful to me, though. What do you think?

randombit:
Yeah I guess just shipping these bits and not testing them isn't buying us anything, and running these tests regularly would be quite painful.

IIUC we can always rederive these values using the XMSS reference implementation and your new script, right? That being the case maybe lets just remove everything that is too slow to test, and if we need to debug some XMSS issue one can use the script to introduce new test vectors to local dev environment.

reneme:
> IIUC we can always rederive these values using the XMSS reference implementation and your new script, right?

That is correct. However, by default (I didn't look into the configuration options at all), the reference implementation was fairly slow. Generating this file (on a single core) took like 36 hours (!) on my laptop. So yes: Re-generation is possible but it takes a while.

That said: chances are that we won't have the right input vector handy when we end up debugging some XMSS corner case. 🤡 I'll get rid of them.

reneme:
> Test failures are relevant. I will look into it tomorrow.

C++ does not define the evaluation order of parameters to a function call. This can lead to [surprising outcomes](https://github.com/randombit/botan/pull/3280/commits/a30cd63dc0e2d87a7ff687a9c02561a6fc628a8d) when trying to test for key generation stability with a given RNG seed.

# 🤕

PR #3287, opened 2023-02-15
Clean up `XMSS_Hash` and prepare for NIST SP.800-208 parameter sets

reneme:
### Pull Request Denpendencies

* #3283 
* #3280

### Description

This simplifies the `XMSS_Hash` implementation somewhat. Two main things: The hash functions are not overloaded anymore, just the version writing into an out-param remains. And `f()`, `h()`, `prf()`, `prf_keygen()` now use the same base implementation. `h_msg()` is somewhat different and keeps its own implementation.

For the record: I did consider removing the `XMSS_Hash` class altogether and replace it by free-standing functions that instantiate the correct hash function, pre-condition it and return the `HashFunction` instance to the caller. Albeit much more clean and flexible, I figured that re-instantiating the hash object each time would significantly hurt XMSS' performance. I assume it's better to live with the (thread-local) hash state management as is. I didn't measure this, though.

To prepare for the NIST parameter sets, the length of the domain-separating "Hash ID" is now configurable in `XMSS_Parameters`.


randombit (approved):
Only reviewed final commit, lgtm

PR #3294, opened 2023-02-16
Chore: Buffered_Computation can deal with generic containers

reneme:
This should be a transparent change for the most part but enables modern support for all sorts of containers without resorting to the C-style pointer APIs.

Allows passing all sorts of contiguous ranges to `::update()` and `::process()` as well as storing results of `Buffered_Computation` in user-defined contiguous containers.

The latter is most-useful for the strong types:

```C++
using MessageHash = Strong<std::vector<uint8_t>, struct MessageHash_>;

auto sha = HashFunction::create("SHA-256");
sha->update("some message");
auto message_hash = sha->final<MessageHash>();
```

### DIFF
--- a/src/lib/base/buf_comp.h
+++ b/src/lib/base/buf_comp.h
@@ -80,7 +80,9 @@ class BOTAN_PUBLIC_API(2,0) Buffered_Computation
       * final result as a container of your choice.
       * @return a contiguous container holding the result
       */
-      template<concepts::contiguous_container T>
+      template<typename T>
+         requires(concepts::contiguous_container<T> &&
+                  concepts::resizable_container<T>)
          T final()
          {
          T output(output_length());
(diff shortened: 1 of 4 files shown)


randombit (approved):

reneme:
Sorry, I should have marked this as a draft while I was still working on it.

We added another overload for `::process<>()` that allows custom output containers (same as `::final<>()`). And also there's a test now, that makes sure all this convenience is actually working (aka compiling).

reneme:
I deliberately left the virtual methods (below) untouched to get the low hanging fruits first.

```C++
virtual void add_data(const uint8_t input[], size_t length) = 0;
virtual void final_result(uint8_t out[]) = 0;
```

Though, both are good candidates for `std::span<>`. Particularly with sub-classes being forced to trust their base class(s) and write into an unbounded array. 😨

What's your take on this? Should we take this on?

randombit:
> What's your take on this? Should we take this on?

Sure, using span internally here would be a lot cleaner overall. But also a lot of work since you'll have to touch every hash/MAC. Those kind of changes tend to be a slog.

reneme:
@randombit I didn't merge this yet because I pushed some API changes after you approved the PR. Most notably: https://github.com/randombit/botan/pull/3294/commits/c4582aaeabc67e607a805ea3ec6ffeaa52e80245

randombit (approved):
:+1:

PR #3297, opened 2023-02-17
Chore: Symmetric_Algorithm can deal with generic containers

reneme:
Introducing `::set_key(std::span)` would have caused an ambiguous overload with `::set_key(SymmetricKey)`. That's what initially side-tracked me resulting in [the OctetString question](https://github.com/randombit/botan/pull/3296).

To resolve this ambiguity, I made the c'tors of `OctetString` explicit.

Burried under this is a bigger question, though: Should the API even allow anything else than a `SymmetricKey` container for `::set_key()`? Or should Botan's public API leverage the type system to try and prevent programming errors?

The latter would probably mean that we'd eventually ask library users to use our strong types as part of the public API. That might or might not be intrusive for some people.

### DIFF
--- a/src/cli/cc_enc.cpp
+++ b/src/cli/cc_enc.cpp
@@ -67,7 +67,7 @@ uint64_t cc_derank(uint64_t cc_number)
    }
 
 uint64_t encrypt_cc_number(uint64_t cc_number,
-                           const Botan::secure_vector<uint8_t>& key,
+                           const Botan::SymmetricKey& key,
                            const std::vector<uint8_t>& tweak)
    {
    const Botan::BigInt n = 1000000000000000;
@@ -90,7 +90,7 @@ uint64_t encrypt_cc_number(uint64_t cc_number,
    }
 
 uint64_t decrypt_cc_number(uint64_t enc_cc,
-                           const Botan::secure_vector<uint8_t>& key,
+                           const Botan::SymmetricKey& key,
                            const std::vector<uint8_t>& tweak)
    {
    const Botan::BigInt n = 1000000000000000;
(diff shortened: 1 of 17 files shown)


randombit:
> Or should Botan's public API leverage the type system to try and prevent programming errors?

That was the original idea behind `SymmetricKey` and `InitializationVector`. But I think in practice this doesn't really help that much because if you can "mistakenly" write

```
aead->set_key(thats_not_a_key_you_dummy);
```

you can just as easily write

```
// this stupid API makes you convert a key to its internal key type
aead->set_key(SymmetricKey(thats_not_a_key_you_dummy));
```

which is why `SymmetricKey` and co are not really used in new APIs and tbh should be deprecated and eventually removed. (I don't think that can happen right now because there are various APIs that still require them, but that should probably occur eventually.)

The *real* errors, the subtle problems that cause systems to actually fail: nonce reuse, reusing a key in two different contexts, allowing cut-and-paste attacks, etc etc are not fixable with the typesystem, and the way to prevent those is to provide good out of the box tools, for those who don't really know what they are doing (eg, TLS, PGP, Signal protocol, whatever) and for those who need low level functionality (ie, implementations of different protocols), offer that functionality with as little mess and inconvenience as possible.

**I'm all in favor of using the type system to prevent errors when that is feasible** but in this instance I think the costs (syntactic, performance overhead, etc) outweigh the benefits (which tbh seem near zero, in terms of real bugs prevented).

reneme:
> The real errors, the subtle problems that cause systems to actually fail: nonce reuse, reusing a key in two different contexts, allowing cut-and-paste attacks, etc etc are not fixable with the typesystem.

Agreed. Bottom line: Low-level programming convenience over type safety in this particular case. You nailed my concern regarding "intrusive API" with your example, btw. 😏 I'll close #3296 on this basis. Thanks.

One more thing: Making `OctetString`'s c'tors explicit _is_ a breaking API change after all. Is that acceptable? Otherwise this wrapper is actually standing in the way of rolling out `std::span` here (and potentially other places).

randombit (approved):
I think some of these changes point to areas where `SymmetricKey` can/should be outright removed from the API. But the change to `set_key` and making the constructors `explicit` is fine.

reneme:
Rebased to master to solve conflicts. Now waiting for CI, then merging.

Re:SymmetricKey: Yeah, I think it isn't used consistently throughout the API and causes more confusion than its actually useful.

PR #3300, opened 2023-02-17
http_utils update using std::span

devnexen:


nunojpg:
We are storing a copy of the data at m_body, not a reference/pointer.

With std::span you just store a pointer, to you need to make sure the original passed data remains available.

Think of it as string vs string_view.

src/lib/x509/x509_crl.cpp
```diff
@@ -43,7 +43,7 @@ X509_CRL::X509_CRL(DataSource& src)
    load_data(src);
    }
 
-X509_CRL::X509_CRL(const std::vector<uint8_t>& vec)
+X509_CRL::X509_CRL(std::span<const uint8_t> vec)
    {
    DataSource_Memory src(vec.data(), vec.size());
```
reneme, line 48:
```suggestion
   DataSource_Memory src(vec);
```

`DataSource_Memory` already learned to deal with `std::span<>`

lieser:
> We are storing a copy of the data at m_body, not a reference/pointer.

I agree that we should be careful about this. And make really sure that not storing a copy is what we want and brings us enough performance that we care about that it is worth the potential lifetime issue.

If we stay with the current storing a copy of data in `Response`, we should change the constructor to takes a `std::vector<uint8_t>` (and not a reference). If the caller no longer needs the body this would allow to avoid creating a copy of the body without the lifetime issue just storing a pointer to it has.

randombit:
This code has evolved quite a bit on master and I think most of the changes don't apply anymore, closing.

Issue #3302, opened 2023-02-17
Dilithium occasionally generates signatures that it rejects


randombit:
After writing #3301 I see occasional failures - Dilithium is generating a signature which it then rejects:

```
Dilithium-4x4-r3 757 keygen/sec; 1.32 ms/op 5267480 cycles/op (1 op in 1 ms)
2 generated signatures rejected in PK signature bench
Dilithium-4x4-r3  3023 sign/sec; 0.33 ms/op 1325232 cycles/op (9075 ops in 3001 ms)
Dilithium-4x4-r3  12220 verify/sec; 0.08 ms/op 327820 cycles/op (36662 ops in 3000 ms)
Dilithium-6x5-r3 3486 keygen/sec; 0.29 ms/op 1149190 cycles/op (1 op in 0 ms)
2 generated signatures rejected in PK signature bench
Dilithium-6x5-r3  1890 sign/sec; 0.53 ms/op 2120116 cycles/op (5671 ops in 3000 ms)
Dilithium-6x5-r3  7435 verify/sec; 0.13 ms/op 538877 cycles/op (22308 ops in 3000 ms)
Dilithium-8x7-r3 2182 keygen/sec; 0.46 ms/op 1836240 cycles/op (1 op in 0 ms)
Dilithium-8x7-r3  1506 sign/sec; 0.66 ms/op 2659674 cycles/op (4521 ops in 3000 ms)
Dilithium-8x7-r3  4386 verify/sec; 0.23 ms/op 913457 cycles/op (13162 ops in 3000 ms)
```

It's not known at this time if this is because the signature is invalid, or because verification has a false-reject.

a) Bad news, we need to either figure this out or in the worst case pull Dilithium from 3.0
b) This points to a major hole in our testing - this should have been caught before Dilithium ever merged

I will look at resolving (b) namely adding randomized testing for all signature schemes.

@reneme @securitykernel @boricm @omarama any takers on (a)

randombit:
Seeing this with all parameter sizes, and AES vs SHAKE doesn't seem to matter.

securitykernel:
That's bad. I was not involved in Dilithium, @reneme and @boricm were. Guys, can you look into this?

reneme:
[...]

PR #3304, opened 2023-02-18
Fix Dilithium verification

randombit:
If the signature was rejected due to a structural issue, then the SHAKE context was never finalized. If this happens, a subsequent verification would fail since the SHAKE context contained some extra bits from the previous, failed, verification.

GH #3302


reneme (approved):
Yikes! Good catch, truely.

boricm:
Good finding, @randombit! Also thanks for the fast fix! 👍 
I used to place the valid signature tests at the beginning of a test suite and the invalid signature tests at the end - guess I'll change my behavior now and add a everything-is-still-working-test at the end of a suite.

PR #3308, opened 2023-02-20
[TLS 1.3] Update Migration Guide

reneme:
That's merely a brain dump of the current state of incompatible API changes in the TLS stack. Note that it includes the (currently unmerged) changes in #3150 and #3140 (which are ready for review, btw. 😉). Feel free to edit my English. 😅

I'll do a review of the APIs in the coming days and might amend this as necessary.

### DIFF
--- a/doc/migration_guide.rst
+++ b/doc/migration_guide.rst
@@ -26,8 +26,15 @@ algorithm headers (such as ``aes.h``) have been removed. Instead you should
 create objects via the factory methods (in the case of AES,
 ``BlockCipher::create``) which works in both 2.x and 3.0
 
-TLS Functionality Removed
----------------------------
+TLS
+---
+
+Starting with Botan 3.0 TLS 1.3 is supported.
+This development required a number of backward-incompatible changes to
+accomodate the protocol differences to TLS 1.2, which is still supported.
+
+Removed Functionality
+^^^^^^^^^^^^^^^^^^^^^
 
 Functionality removed from the TLS implementation includes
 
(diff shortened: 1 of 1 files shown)


doc/migration_guide.rst
```diff
@@ -26,8 +26,15 @@ algorithm headers (such as ``aes.h``) have been removed. Instead you should
...
 ``BlockCipher::create``) which works in both 2.x and 3.0
 
-TLS Functionality Removed
----------------------------
+TLS
+---
+
+Starting with Botan 3.0 TLS 1.3 is supported.
+This development required a number of backward-incompatible changes to
+accomodate the protocol differences to TLS 1.2 which is still supported.
```
randombit, line 34: ", which is still supported"

doc/migration_guide.rst
```diff
@@ -41,19 +48,118 @@ Functionality removed from the TLS implementation includes
...
+A number of new callbacks were added with TLS 1.3. None of those new callbacks
+is mandatory to implement by applications, though. Additionally there are a few
+backward incompatible changes in callbacks that might require attention by some
+applications:
+
+tls_verify_cert_chain()
+"""""""""""""""""""""""
+
+The parameter `ocsp_responses` is now a vector of `std::optional<OCSP::Response>`
+rather than potentially null `std::shared_ptr<OCSP::Response>`.
```
randombit, line 70: This is confusingly worded IMO in that it sounds like we are guaranteeing that the `std::optional`s are not `nullopt`, which is not the case, right?

Might be easier to just strike the "rather than potentially null":

> The parameter `ocsp_responses`, which was previously `std::shared_ptr<OCSP::Response>`, is now 
`std::optional<OCSP::Response>`

doc/migration_guide.rst
```diff
@@ -41,19 +48,118 @@ Functionality removed from the TLS implementation includes
...
-Sessions
-^^^^^^^^
+Policy
+^^^^^^
+
+choose_key_exchange_group()
+"""""""""""""""""""""""""""
+
+The new parameter `offered_by_peer` identifies the key exchange groups a peer
+has sent public exchange offerings for (in TLS 1.3 handshakes only).
```
randombit, line 87: Maybe include something like "In TLS 1.2, this vector is always empty" just so it's clear that this function is still called in TLS 1.2 (IIUC), just missing this information.

randombit (approved):
A few minor comments, but looks good, thanks

reneme:
Thanks. There will be more edits re: the session management. But merging for now.

PR #3317, opened 2023-02-22
Chore: AEAD_Mode can deal with generic containers

reneme:
This introduces `AEAD_Mode::set_associated_data(std::span<>)` alongside the ptr-length-overloads. It switches the virtual method to the `std::span<>` overload and adapts the AEAD implementations accordingly. Also `set_associated_data_vec()` is marked as deprecated as it is just an alias now.

### Caveat

This might be somewhat a show-stopper for the overload-based introduction of `std::span<>` we had in mind ~~(also affecting #3297, #3294, #3195, #3201)~~. In short: overloads of virtual methods in the base class are not necessarily dispatched to from an object of a derived type. A [godbolt says more than a hundred words](https://godbolt.org/z/11GWWrKhW).

Same code for reference:

```C++
class AEAD_Mode {
public:
    virtual void set_associated_data(std::span<const uint8_t> ad) = 0;
    void set_associated_data(const uint8_t ad[], size_t ad_len)
        { set_associated_data(std::span(ad, ad_len)); }
};

class GCM : public AEAD_Mode {
public:
    // Otherwise the compiler does not statically dispatch to the
    // ptr-length overload in the base class
    using AEAD_Mode::set_associated_data;

    void set_associated_data(std::span<const uint8_t> ad) override final
        { std::cout << "Yeah!" << std::endl; }

};

int main()
{
    GCM gcm;

    std::vector<uint8_t> ad{'a', 'b', 'c'};

    // this works
    gcm.set_associated_data(ad);
    
    // this does not (without the `using` declaration in GCM)
    gcm.set_associated_data(ad.data(), ad.size());

    // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

    std::unique_ptr<AEAD_Mode> gcm_ptr = std::make_unique<GCM>();
    
    // this works regardless of the `using` declaration
    gcm_ptr->set_associated_data(ad);
    gcm_ptr->set_associated_data(ad.data(), ad.size());

    return 0;
}
```

Frankly, I'm quite surprised that the method dispatch doesn't find the (publicly available) overload in the base class. 😢 One can work around it by explicitly `using Base::method` in the derived class, but that's easy to miss in large refactorings.
On the other hand, one could simply use other method names for the `std::span` overloads. But that complicates the public API even more and forces us to come up with new names for established concepts.

What a mess! Or am I actually missing something?

### DIFF
--- a/src/lib/modes/aead/aead.h
+++ b/src/lib/modes/aead/aead.h
@@ -10,6 +10,8 @@
 
 #include <botan/cipher_mode.h>
 
+#include <span>
+
 namespace Botan {
 
 /**
@@ -54,9 +56,11 @@ class BOTAN_PUBLIC_API(2,0) AEAD_Mode : public Cipher_Mode
       * once (after set_key) is the optimum.
       *
       * @param ad the associated data
-      * @param ad_len length of add in bytes
       */
-      virtual void set_associated_data(const uint8_t ad[], size_t ad_len) = 0;
+      void set_associated_data(std::span<const uint8_t> ad)
+         { set_associated_data_n(0, ad); }
+      void set_associated_data(const uint8_t ad[], size_t ad_len)
+         { set_associated_data(std::span(ad, ad_len)); }
 
       /**
       * Set associated data that is not included in the ciphertext but
@@ -71,11 +75,14 @@ class BOTAN_PUBLIC_API(2,0) AEAD_Mode : public Cipher_Mode
       * all other modes only nominal AD input 0 is supported; all
       * other values of idx will cause an exception.
       *
+      * Derived AEADs must implement this. For AEADs where
+      * `maximum_associated_data_inputs()` returns 1 (the default), the
+      * @p idx must simply be ignored.
+      *
       * @param idx which associated data to set
       * @param ad the associated data
-      * @param ad_len length of add in bytes
       */
-      virtual void set_associated_data_n(size_t idx, const uint8_t ad[], size_t ad_len);
+      virtual void set_associated_data_n(size_t idx, std::span<const uint8_t> ad) = 0;
 
       /**
       * Returns the maximum supported number of associated data inputs which
@@ -104,7 +111,7 @@ class BOTAN_PUBLIC_API(2,0) AEAD_Mode : public Cipher_Mode
       template<typename Alloc>
       void set_associated_data_vec(const std::vector<uint8_t, Alloc>& ad)
          {
-         set_associated_data(ad.data(), ad.size());
+         set_associated_data(ad);
          }
 
       /**
@@ -116,10 +123,9 @@ class BOTAN_PUBLIC_API(2,0) AEAD_Mode : public Cipher_Mode
       *
       * @param ad the associated data
       */
-      template<typename Alloc>
-      void set_ad(const std::vector<uint8_t, Alloc>& ad)
+      void set_ad(std::span<const uint8_t> ad)
          {
-         set_associated_data(ad.data(), ad.size());
+         set_associated_data(ad);
          }
 
       /**
--- a/src/lib/modes/aead/aead.h
+++ b/src/lib/modes/aead/aead.h
@@ -109,6 +109,7 @@ class BOTAN_PUBLIC_API(2,0) AEAD_Mode : public Cipher_Mode
       * @param ad the associated data
       */
       template<typename Alloc>
+      BOTAN_DEPRECATED("Simply use set_associated_data")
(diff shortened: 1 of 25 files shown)


reneme:
Based on a suggestion by @lieser I added an indirection to work around above-mentioned caveat. The protected pure-virtual method `set_ad_n()` is now derived by all AEAD implementations. The `idx` parameter is ignored by most (except SIV), while the public API wrapper ensures that only SIV can successfully receive anything but `0` for the `idx`.

Come to think of it, that's exactly what we ended up implementing in #3195.

src/lib/modes/aead/aead.h
```diff
@@ -116,8 +122,7 @@ class BOTAN_PUBLIC_API(2,0) AEAD_Mode : public Cipher_Mode
       *
       * @param ad the associated data
       */
-      template<typename Alloc>
-      void set_ad(const std::vector<uint8_t, Alloc>& ad)
+      void set_ad(std::span<const uint8_t> ad)
```
reneme, line 125: Should that be deprecated as well, on the same basis as `set_associated_data_vec`. Its just an alias for `set_associated_data()`.

src/lib/modes/aead/ccm/ccm.cpp
```diff
@@ -91,18 +91,18 @@ void CCM_Mode::key_schedule(const uint8_t key[], size_t length)
    m_cipher->set_key(key, length);
    }
 
-void CCM_Mode::set_associated_data(const uint8_t ad[], size_t length)
+void CCM_Mode::set_ad_n(size_t /* ignored */, std::span<const uint8_t> ad)
```
randombit, line 94: Ignoring the specified AD index seems like a really bad idea to me. There is a reason the current code throws if you attempt to use an idx > 0 with a mode that doesn't support more than one AD, to do otherwise is to invite really subtle bugs.

src/lib/modes/aead/aead.h
```diff
@@ -129,6 +134,14 @@ class BOTAN_PUBLIC_API(2,0) AEAD_Mode : public Cipher_Mode
...
 
       virtual ~AEAD_Mode() = default;
+
+   protected:
+      /**
+       * Derived AEADs must implement this.
+       * If `maximum_associated_data_inputs()` returns 1 (the default), the
+       * @p idx must simply be ignored.
+       */
+      virtual void set_ad_n(size_t idx, std::span<const uint8_t> ad) = 0;
```
randombit, line 144: The overload thing is a pain. I have no idea why C++ does that - this is why for example the `HashFunction`s have an `update` call that delegates to the virtual `add_data`, instead of the `update` directly being the virtual.

In this specific case I think what we could do is keep `set_associated_data_n` as the virtual entry point for setting the AD, and have it take a `span`. This function is new (only added in 2.17.0) and probably very rarely used since it only is useful for SIV at the moment, so it's probably fine to just change the signature and not provide any overloads.

src/lib/modes/aead/aead.h
randombit: Kind of an alias, but kind of not. The point of this is to bypass exactly the virtual overloading problem you encountered. Otherwise they all would have just been named `set_associated_data`.

src/lib/modes/aead/ccm/ccm.cpp
reneme: I introduced `set_ad_n` as the internal interface to get around the virtual dispatch issue. The idea was to combine both virtual methods `set_additional_data` and `set_additional_data_n` for the internal implementation, "saving" one virtual method. The public interface stays the same and throws when misused as before.

src/lib/modes/aead/aead.h
reneme: With the (protected) `set_ad_n` described in the other comment, they can all be called `set_associated_data` now. Hence the suggestion to deprecate it.

src/lib/modes/aead/aead.h
reneme: Do we need to mention that in the migration guide?

src/lib/modes/aead/ccm/ccm.cpp
reneme: Done.

reneme:
Rebased to master after one of yesterday's PRs conflicted with it.

src/lib/modes/aead/aead.h
randombit: Yeah fine to deprecate given the other changes in this PR

randombit (approved):

PR #3322, opened 2023-02-24
[TLS 1.3] Generalize ephemeral key handling callbacks

reneme:
In TLS 1.2 the key exchange is initiated by the server:

```
            Client                   Server
              |                        | gen key pair
              | <---- KEX pubkey ----- |
gen key pair  |                        |
derive secret |                        |
              | ----- KEX pubkey ----> |
              |                        | derive secret
```

The existing callbacks `tls_dh_agree()` and `tls_ecdh_agree()` implemented both "gen key pair" and "derive secret" in one step. This allowed customization of the client-side key exchange logic by the application.

In TLS 1.3 the key exchange procedure is "reversed", in a sense that the client first sends a public key and the combined "gen key pair"/"derive secret" happens on the server:

```
            Client                   Server
gen key pair  |                        |
              | ----- KEX pubkey ----> |
              |                        | gen key pair
              |                        | derive secret
              | <---- KEX pubkey ----- |
derive secret |                        |
```

Therefore, said callbacks cannot be used to allow client applications to introduce custom key exchange logic for TLS 1.3 -- their implemented behaviour lives on the server side now. Additionally, the existing callbacks did not allow the introduction of entirely new algorithms for key exchange (read: PQC). As the server side was not customizable.

This PR replaces them by `tls_ephemeral_key_generate()` and `tls_ephemeral_key_agreement()` as atomic operations for a customizable key exchange. These callbacks can be used on the client and server side both in TLS 1.2 and 1.3 and provide much more flexibility to applications that wish to customize the key exchange.

As a side-effect the callback `tls_decode_group_param()` was obsoleted and removed.

@harrmann please verify that this new design is compatible with your requirements.


src/lib/tls/tls_algos.h
```diff
@@ -99,6 +99,32 @@ enum class Group_Params : uint16_t {
...
+
+constexpr bool is_dh(const Group_Params group)
+   {
+   return
+      group == Group_Params::FFDHE_2048 ||
+      group == Group_Params::FFDHE_3072 ||
+      group == Group_Params::FFDHE_4096 ||
+      group == Group_Params::FFDHE_6144 ||
+      group == Group_Params::FFDHE_8192;
+   }
```
reneme, line 126: Note to self: we might want to make `Group_Params` a "smart enum" that provides these predicates as methods.

reneme:
Force-pushed again to fix commit signatures.

lieser (approved):

src/lib/tls/tls_callbacks.cpp
```diff
@@ -147,87 +143,123 @@ bool TLS::Callbacks::tls_verify_message(
...
-      throw TLS_Exception(Alert::InsufficientSecurity,
-                          "DH group validation failed");
+}
 
-   DH_PublicKey peer_key(group, Y);
+std::unique_ptr<PK_Key_Agreement_Key> TLS::Callbacks::tls_generate_ephemeral_key(
+   const std::variant<TLS::Group_Params, DL_Group> group,
+   RandomNumberGenerator& rng)
+   {
+   if(is_dh_group(group))
```
randombit, line 174: Just to check - there is something higher in the stack that is ensuring that in 1.3 we are only using group IDs and not explicit groups, right?

randombit (approved):
Looks good to me

src/lib/tls/tls_callbacks.cpp
reneme: TLS 1.3 doesn't support explicit statement of groups. The only place where this callback is called in the TLS 1.3 implementation is here: 

https://github.com/randombit/botan/pull/3322/files#diff-2cbf91ed19123a472976d7d1e316ab16b24e1a4062d6c97dfcb5af7d630e4d1fR44-R46

```
      Key_Share_Entry(const TLS::Group_Params group, Callbacks& cb, RandomNumberGenerator& rng)
         : m_group(group)
         , m_private_key(cb.tls_generate_ephemeral_key(group, rng))
```

... and we only have a group ID handy there.

PR #3323, opened 2023-02-25
Fix clang-tidy warnings in recently introduced code

randombit:
cc @reneme on the TLS session manager changes

### DIFF
--- a/src/lib/tls/tls_session_manager_memory.cpp
+++ b/src/lib/tls/tls_session_manager_memory.cpp
@@ -107,8 +107,7 @@ size_t Session_Manager_In_Memory::remove_internal(const Session_Handle& handle)
          std::erase_if(m_sessions, [&](const auto& item)
             {
             const auto& [_, session_and_handle] = item;
-            const auto& [__, this_handle] = session_and_handle;
-
+            const auto& [_unused, this_handle] = session_and_handle;
             return this_handle.is_ticket() && this_handle.ticket().value() == ticket;
             });
          return before - m_sessions.size();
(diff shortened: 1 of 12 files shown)


src/lib/tls/tls_session_manager_memory.cpp
```diff
@@ -107,8 +107,7 @@ size_t Session_Manager_In_Memory::remove_internal(const Session_Handle& handle)
          std::erase_if(m_sessions, [&](const auto& item)
             {
             const auto& [_, session_and_handle] = item;
-            const auto& [__, this_handle] = session_and_handle;
-
+            const auto& [_unused, this_handle] = session_and_handle;
```
reneme, line 110: Maybe `_unused1` and `_unused2` for consistency? I found double-underscore was kind-of good enough to work around the 'repeated identifier' issue. But if clang-tidy doesn't like that, so be it.

reneme (approved):
LGTM. Is there something that stops us from running `clang-tidy` in CI to catch those issues early?

src/lib/tls/tls_session_manager_memory.cpp
randombit: Yeah that's fine. For context the complaint here is because C++ reserves all identifiers containing `__` for the implementation.

randombit:
The main reason to not run it in CI is runtime. It takes about 15 minutes to run `clang-tidy` over the entire codebase on my 4 core + HT machine. So might take as long as an hour on the GH Actions images.

Maybe we can hack up something that only runs it over source files that changed in a PR - that will still miss some problems but would catch the majority before merge, which would be an improvement over the current situation.

Issue #3227, opened 2023-01-26
Improve (internal) handling of Signature Schemes that don't require an EMSA


reneme:
Parts of the code assume that all signature schemes use an EMSA (and along with that a configurable hash function). That's not true for e.g. Ed25519 and also newer algorithms like Dilithium.

See also #3223, #3208 for more details.

Somewhat related: 

 * #3039

randombit:
I think this situation has been improved somewhat by the recent work in #3315 and #3320 but a lot of that only addressed the signature generation side, going from a `PK_Signer` to the `AlgorithmIdentifier` that we should use to identify the signature in an X509 structure.

The other half, verification, is still an issue, `X509_Object::verify_signature` and `hash_used_for_signature` still contain a lot of algorithm specific logic.

Maybe the fix here is to allow `PK_Verifier` to be directly instantiated with a `AlgorithmIdentifier` and then each key can decide for itself what it is going to do with it. That plus adding `PK_Verifier::hash_function` should handle this neatly.

I think I avoided this in the past because it seems to entangle the X509 code with the public key code, but X509 isn't exactly going away anytime soon. And as more PQC algorithms are added the situation will only become worse if something isn't done.

randombit:
I think we are doing pretty well here at this point. The main remaining centralization of logic in X.509 is now `x509_signature_padding_for` in `x509_obj.cpp` where we pick an appropriate padding scheme based on the tuple of `(pk algorithm,user specified hash,user specified padding scheme)`. I don't see an obvious cleanup on this one.

reneme:
Great progress on that, thank you! I really love to see all those centralized if-else-heaps go. Could you elaborate on what is the obstacle for `x509_signature_padding_for`?

I'm assuming you would want `X509_Object::choose_sig_format()` to stay stable. 

Maybe add a `Private_Key::preferred_padding(std::string hash_fn_preference = "")` that returns the algo-specific default cases in `x509_signature_padding_for` (potentially incorporating the user's hash preference if applicable).

Then add a constructor to `PK_Signer` that additionally takes the `hash_fn_preference` and uses the new `Private_Key::preferred_padding()`.

My hope would be, that this could unify the double-lookup in `X509_Object::choose_sig_format()` and distribute the algo-specific code in the `Private_Key` implementations.

Though, frankly, I think I'm lacking the overview to see whether this will cover all major cases.

randombit:
[...]

PR #3325, opened 2023-02-25
Add PK_Verifier::hash_function, remove X509_Object::hash_used_for_signature

randombit:
Further push on #3227

### DIFF
--- a/src/lib/x509/x509_obj.h
+++ b/src/lib/x509/x509_obj.h
@@ -48,11 +48,6 @@ class BOTAN_PUBLIC_API(2,0) X509_Object : public ASN1_Object
       const AlgorithmIdentifier& signature_algorithm() const { return m_sig_algo; }
 
       /**
-      * @return hash algorithm that was used to generate signature
-      */
-      std::string hash_used_for_signature() const;
-
-      /**
       * Create a signed X509 object.
       * @param signer the signer used to sign the object
       * @param rng the random number generator to use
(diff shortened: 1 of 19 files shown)


src/lib/x509/x509_obj.h
```diff
@@ -47,11 +47,6 @@ class BOTAN_PUBLIC_API(2,0) X509_Object : public ASN1_Object
       */
       const AlgorithmIdentifier& signature_algorithm() const { return m_sig_algo; }
 
-      /**
-      * @return hash algorithm that was used to generate signature
-      */
-      std::string hash_used_for_signature() const;
```
lieser, line 53: For my own understanding: when is an existing `BOTAN_PUBLIC_API(major,minor)` supposed to get changed?

- Then the API is extended? (probably not)
- Then the API is broken? E.g.
  - By removing a part of the API
  - Or changing a part of the API
- Never, as it is just supposed to indicated when some version of the API was added, regardless if it is still compatibel with it.

src/lib/x509/x509_obj.h
```diff
@@ -69,9 +64,11 @@ class BOTAN_PUBLIC_API(2,0) X509_Object : public ASN1_Object
       * Check the signature on this data
       * @param key the public key purportedly used to sign this data
       * @return status of the signature - OK if verified or otherwise an indicator of
-      *         the problem preventing verification.
+      *         the problem preventing verification, along with the hash function that
+      *         was used, for further policy checks. The second parameter is empty
+      *         unless the validation was sucessful.
       */
-      Certificate_Status_Code verify_signature(const Public_Key& key) const;
+      std::pair<Certificate_Status_Code, std::string> verify_signature(const Public_Key& key) const;
```
lieser, line 71: I personally am not a fan of `std::pair` in public APIs. Did you consider adding a small helper struct for it so the parts are properly named then accessed?

randombit: Might not be a bad idea. Though this function while nominally public mostly exists to help implement other functionality so I'm not sure it matters that much in this particular case.

PR #3329, opened 2023-02-27
Allow initializing a PK_Verifier from an AlgorithmIdentifier

randombit:
This moves the logic for understanding/parsing an algid down to key-specific code instead of forcing it to be in one centralized spot.

Also this fixes a number of corner cases that were previously missed, eg for Dilithium we would accept signatures where the signature parameters was NULL instead of absent.

GH #3227

### DIFF
--- a/src/lib/pubkey/pk_keys.h
+++ b/src/lib/pubkey/pk_keys.h
@@ -229,6 +229,22 @@ class BOTAN_PUBLIC_API(2,0) Public_Key : public virtual Asymmetric_Key
       virtual std::unique_ptr<PK_Ops::Verification>
          create_verification_op(const std::string& params,
                                 const std::string& provider) const;
+
+      /**
+      * This is an internal library function exposed on key types.
+      * In all cases applications should use wrappers in pubkey.h
+      *
+      * Return a verification operation for this combination of key and
+      * signature algorithm or throw.
+      *
+      * @param signature_algorithm is the X.509 algorithm identifier encoding the padding
+      * scheme and hash hash function used in the signature if applicable.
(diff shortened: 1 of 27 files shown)


src/lib/pubkey/pk_keys.h
```diff
@@ -207,6 +207,18 @@ class BOTAN_PUBLIC_API(2,0) Public_Key : public virtual Asymmetric_Key
...
                                 const std::string& provider) const;
+
+      /**
+      * This is an internal library function exposed on key types.
+      * In all cases applications should use wrappers in pubkey.h
+      *
+      * Return a verification operation for this key/params or throw
+      * @param signature_algorithm the X509 algorithm identifier
+      * @param provider the provider to use
+      */
```
reneme, line 218: I think this deserves a few more details. Maybe:

```suggestion
      /**
      * This is an internal library function exposed on key types.
      * In all cases applications should use wrappers in pubkey.h
      *
      * Return a verification operation for this combination of key and
      * signature algorithm or throw. @param signature_algorithm is the
      * X.509 algorithm identifier encoding the padding scheme and hash
      * hash function used in the signature if applicable.
      *
      * @param signature_algorithm the X509 algorithm identifier
      * @param provider the provider to use
      */
```

reneme (approved):
That's definitely a worthwhile cleanup! I like that the special cases (if any) are now handled in the algorithm implementations and not centrally.

src/lib/pubkey/pk_keys.h
randombit: :+1:

PR #3337, opened 2023-03-03
[TLS 1.3] Use TLS::Callbacks::tls_session_established()

reneme:
### Pull Request Dependencies

* ~~#3316~~

### Descripton

This adapts `TLS::Callbacks::tls_session_established()` to be usable with TLS 1.3 (similar to [the discussed approach here](https://github.com/randombit/botan/pull/2988#issuecomment-1351250465)). Previously, this callback had two tasks:

1. Provide the application with an overview of a just-negotiated TLS connection. If the application did not agree with the negotiated features for some reason it could throw an exception and stop the finalization of the handshake
2. Decide whether the resumption information associated to the just-negotiated connection should be persisted for a resumption later on.

Unfortunately, in TLS 1.3 resumption information and individual TLS connections are not as tightly bound as in TLS 1.2.

There's a new `class Session_Summary` that is similar to the existing `Session` but does not contain the actual resumption information (most notably the `master_secret`). Also, the above-mentioned callback objectives are split across two callbacks:

1. `tls_session_established(const Session_Summary&)` (e.g.) to give applications a last-second chance to veto the connection.
2. `tls_should_persist_resumption_information(const Session&)` to give applications a chance to decide whether this particular resumption "ticket" should be kept for later or not.

Both are now called in a usage-compatible fashion in TLS 1.2 and TLS 1.3 implementations. **TLS 1.3 will call `tls_should_persist_resumption_information()` for each session ticket** it creates or receives.

Also note that **`tls_session_established()` is no longer a mandatory callback**. By default it is a NOOP. Also `tls_should_persist_resumption_information()` is not mandatory and returns `true` in most cases.

### Outlook

Technically, `tls_should_persist_resumption_information()` is not really needed. Applications that wish to selectively store sessions, could also provide their own `Session_Manager` implementation and do the filtering there. We decided to keep it, to make the API-adaption as pain-free as possible for applications. Though, this new callback could be marked as deprecated somewhere on the way towards Botan 4.0.0, IMO.

### TODO

* [x] Update migration guide

### DIFF
--- a/doc/api_ref/tls.rst
+++ b/doc/api_ref/tls.rst
@@ -74,12 +74,10 @@ information about the connection.
 
  .. cpp:function:: void tls_session_established(const Botan::TLS::Session_Summary& session)
 
-     Mandatory. Called whenever a negotiation completes. This can happen more
-     than once on any connection, if renegotiation occurs. The *session* parameter
+     Optional - default implementation is a no-op
+     Called whenever a negotiation completes. This can happen more than once on
+     TLS 1.2 connections, if renegotiation occurs. The *session* parameter
      provides information about the session which was just established.
-
-     If this function returns false, the session will not be cached
-     for later resumption.
 
      If this function wishes to cancel the handshake, it can throw an
      exception which will send a close message to the counterparty and
--- a/doc/api_ref/tls.rst
+++ b/doc/api_ref/tls.rst
@@ -72,7 +72,7 @@ information about the connection.
      received before the handshake is complete are not authenticated and
      could have been inserted by a MITM attacker.
 
- .. cpp:function:: bool tls_session_established(const TLS::Session& session)
+ .. cpp:function:: void tls_session_established(const Botan::TLS::Session_Summary& session)
 
      Mandatory. Called whenever a negotiation completes. This can happen more
      than once on any connection, if renegotiation occurs. The *session* parameter
@@ -195,7 +195,7 @@ information about the connection.
      Optional. Called by the server when a client hello includes a list of supported groups in the
      supported_groups extension and by the client when decoding the server key exchange including the selected curve identifier.
      The function should return the name of the DH group or elliptic curve the passed
-     TLS group identifier should be mapped to. Therefore this callback enables the use of custom 
+     TLS group identifier should be mapped to. Therefore this callback enables the use of custom
      elliptic curves or DH groups in TLS, if both client and server map the custom identifiers correctly.
      Please note that it is required to allow the group TLS identifier in
      in the used :cpp:class:`TLS::Policy`.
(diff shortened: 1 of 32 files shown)


doc/api_ref/tls.rst
```diff
@@ -72,7 +72,7 @@ information about the connection.
      received before the handshake is complete are not authenticated and
      could have been inserted by a MITM attacker.
 
- .. cpp:function:: bool tls_session_established(const TLS::Session& session)
+ .. cpp:function:: void tls_session_established(const Botan::TLS::Session_Summary& session)
 
      Mandatory. Called whenever a negotiation completes. This can happen more
```
reneme, line 77: Not mandatory anymore.

reneme: Also, (a few lines below) the return value is now `void`

src/lib/tls/tls_callbacks.h
```diff
@@ -125,19 +122,20 @@ class BOTAN_PUBLIC_API(2,0) Callbacks
          }
 
        /**
-       * Optional callback: New session ticket received
-       * Called when we receive a session ticket from the server at any point
-       * after the initial handshake has finished. Clients may decide to keep or
-       * discard the session ticket in the configured session manager.
+       * Optional callback: Resumption information was received/established
+       *
+       * TLS 1.3 calls this when we sent or received a TLS 1.3 session ticket at
```
lieser, line 127: Maybe make it more clear that (on the server side) for sending, this is called *before* the session is sent (or even the session manager asked for one).

So it will also prevent sending of the information no just the caching.

src/lib/tls/tls13/tls_channel_impl_13.h
```diff
@@ -188,8 +188,9 @@ class Channel_Impl_13 : public Channel_Impl
       */
       bool secure_renegotiation_supported() const override
          {
-         // No renegotiation supported in TLS 1.3
-         return false;
+         // Secure renegotiation is not supported in TLS 1.3, though BoGo
+         // tests expect us to claim that it is available.
+         return true;
```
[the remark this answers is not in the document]
reneme: My best guess: This is for legacy purposes of applications that somehow required secure renegotiation on an API level. Secure renegotiation was a TLS feature [added to the protocol later-on](https://www.ietf.org/rfc/rfc5746.txt), after a vulnerability was found.

Technically, renegotiation (secure or insecure) was removed with TLS 1.3 altogether. Or (more specifically) was replaced by the new but better scoped [KeyUpdate](https://www.rfc-editor.org/rfc/rfc8446.html#section-4.6.3) and [Post-Handshake client authentication](https://www.rfc-editor.org/rfc/rfc8446.html#section-4.6.2) features.

`TLS::Policy::allow_insecure_renegotiation()` (defaults to false) still allows to disable secure renegotiation for TLS 1.2. So we can't remove this method (yet) and need to answer something in TLS 1.3. In a way not providing renegotiation on a protocol level also makes it secure. 🤡 Hence, I find it makes sense-ish.

lieser (approved):

reneme:
@randombit Do you want to have a look at this. From our side, this can go in.

randombit (approved):

PR #3346, opened 2023-03-04
Remove the RNG argument from PK_KEM_Encryptor

randombit:
It was entirely unused by all algorithms


mouse07410:
Current implementation (mostly) follows NIST API. My concern with that API is that in order to run KAT the caller has to "rig" the RNG. Modifying the NIST API and passing RNG explicitly would alleviate the need for that hack.

randombit:
@mouse07410 I do not really understand what issue you are raising here nor how it may relate to this PR

mouse07410:
The issue is really with the NIST API. So, probably, shouldn't've brought it up here.

Issue #3342, opened 2023-03-04
`tls_session_established` only accepts exception of type `Botan::TLS::TLS_Exception` as valid type for aborting handshake


marty1885:
Hi,

In the [document](https://botan.randombit.net/handbook/api_ref/tls.html) of `tls_session_established`. It states that

> If this function wishes to cancel the handshake, it can throw an exception which will send a close message to the counterparty and reset the connection state.

This seems to indicate that any exception will cause Botan to reset the connection. However, after some experiments. I find that it only accepts `Botab::TLS::TLS_Exception`. Others like `std::runtimer_error` does not work.

Maybe this is an documentation error?

randombit:
This behavior is surprising to me. Certainly if you want to control which alert is sent, you would need to throw a `TLS_Exception` to achieve that. But I would have thought any exception would be caught and end up as an internal error alert to the peer. This deserves a look, thank you for reporting this.

reneme:
I'm working on this. Currently building a regression test in `unit_tls.cpp` to see whether this is easily reproducible. Judging from the code, I'm also baffled why that wouldn't work as expected.

reneme:
I'm working on this. Currently building a regression test in `unit_tls.cpp` to see whether this is easily reproducible. Judging from the code, I'm also baffled why that wouldn't work as expected.

reneme:
Unfortunately I didn't manage to reproduce this (based on the Botan 2.x branch) with [a regression test](https://github.com/randombit/botan/pull/3348). When throwing `Botan::Invalid_Authentication_Tag` or `std::runtime_error` the peer receives a `BAD_RECORD_MAC` or `INTERNAL_ERROR` [as expected](https://github.com/randombit/botan/blob/9747b8ad75c33df958fe26bdc335521bf5c9cd83/src/lib/tls/tls_channel.cpp#L430-L449).

Could you elaborate on what faulty behaviour you are experiencing?

reneme:
Any news here? Otherwise, I'd close as "cannot reproduce" and merge #3348 to keep it that way. Also: Probably it would make sense to forward-port this to `master`.

reneme:
Closing as *not reproducible*. Please don't hesitate to re-open if you have further information, @marty1885.

PR #3348, opened 2023-03-05
Regression Test: Trying to reproduce GH #3342

reneme:
This validates that throwing an exception in TLS::Callbacks::tls_session_established results in a graceful closure of the TLS connection with an appropriate alert code.

### DIFF
--- a/src/tests/unit_tls.cpp
+++ b/src/tests/unit_tls.cpp
@@ -3,6 +3,7 @@
 *     2016 Matthias Gierlings
 *     2017 René Korthaus, Rohde & Schwarz Cybersecurity
 *     2017 Harry Reimann, Rohde & Schwarz Cybersecurity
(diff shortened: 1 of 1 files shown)


src/tests/unit_tls.cpp
```diff
@@ -627,11 +695,6 @@ void TLS_Handshake_Test::go()
       if(server_handshake_completed == false && m_server->is_active())
          server_handshake_completed = true;
 
-      if(client->is_closed() || m_server->is_closed())
-         {
-         break;
-         }
```
reneme, line 633: This early-exit needed to be removed so that the peer of an aborting client/server receives the alert (and we can check our expectation in `Callbacks::tls_alert()`).

randombit (approved):
Weird that we cannot repro the issue as described, but this is in any event a good test to have.

PR #3349, opened 2023-03-05
std::span/std::string_view for hex_encode/decode

reneme:


src/lib/codec/hex/hex.cpp
```diff
@@ -155,12 +155,19 @@ size_t hex_decode(uint8_t output[],
...
    {
    return hex_decode(output, input.data(), input.length(), ignore_ws);
    }
 
+size_t hex_decode(std::span<uint8_t> output,
+                  std::string_view input,
+                  bool ignore_ws)
+   {
+   return hex_decode(output.data(), input.data(), input.length(), ignore_ws);
+   }
```
reneme, line 169: Initially I wanted to add a sanity check like `output.size() >= input.size() / 2`. Though, that has subtle problems when handling short inputs and ignoring whitespace. From what I understand, the decoding logic always needs at least one byte as scratch space.

We could add the sanity-check for inputs longer than 2 bytes, but, strictly speaking, that would not be a constant-time check anymore.

randombit (approved):

PR #3351, opened 2023-03-06
Improve VSCode integration

FAlbertDev:


reneme:
FYI: Fabian is our new colleague. We're using the opportunity to improve the dev environment while setting things up for him.

randombit:
So this seems to be removing the Sublime config which is fine as far as I'm concerned, if nobody is using/maintaining that anymore, but doesn't seem to be explicitly mentioned.

Re moving the VSCode setup ... I'd rather not unless this is really necessary, in that I like keeping the root directory as clean as possible. I'd rather if `.github` and `.codecov.yml` were not there, but rather in a CI specific subdir, but this is unavoidable. For the editor, it seems like a local one time command

`ln -s src/editors/vscode .vscode`

would suffice? Am I missing something with this? It feels like the real fix is maybe better dev onboarding docs? There is `contributing.rst` but it hasn't been touched in a while.

reneme:
Re: clean repo-root: I used it exactly like that (pulling in .editorconfig and .vscode via symlinks). Moving it up into the root just makes for a more out-of-the-box experience for new contributers, I think. Though, I'm totally fine with your veto and updating the docs accordingly.

Re: sublime: I added that at some point, never brought it to a point where it was actually useful in a sustainable way. I don't think we break anyone's workflow by removing that.

reneme:
Regarding `clang-tidy`: I just saw that the `clangd` extension in VSCode actually integrates with `clang-tidy` as will. I.e. if one adds a `.clang-tidy` file (probably top-level as well 😨) it would (IIUC) show relevant clang-tidy warnings right during development. Though, this would mean we'd need to switch from the `run_clang_tidy.py` to a `.clang-tidy` file.

randombit:
On that I'd want to keep the script since it can parallelize running the linter (have I mentioned `clang-tidy` is REALLY slow?), amusingly even LLVM has their own equivalent script for the same reasons https://clang.llvm.org/extra/doxygen/run-clang-tidy_8py_source.html but we could generalize it a bit so that it can also generate a suitable configuration file.

randombit:
I'm going to decline on the basis of keeping root dir as clean as possible. This is arbitrary but so it goes.

Fine to open another PR removing the Sublime stuff if nobody is using/maintaining that.

PR #3363, opened 2023-03-13
Public TLS API uses std::span

reneme:
This adds overloads for `std::span<>` to the TLS channel API: Namely `TLS::Channel::received_data()` and `::send()`. Internally, those are now dispatched to the new (virtual) protected methods `::from_peer()` and `::to_peer()`. Subclasses of the basic TLS channel now overload those. The ptr-length style overloads of `::received_data()` and `::send()` are left intact.

The `TLS::Callbacks::tls_emit_data()` and `TLS::Callbacks::tls_record_received()` now use `std::span<>` as well. This is not backward compatible, unfortunately. The migration guide has been adapted accordingly.

Also, the TLS 1.3 implementation is adapted slightly to push the `std::span<>` deeper into layers. More adaptions are left for future work.

### DIFF
--- a/doc/migration_guide.rst
+++ b/doc/migration_guide.rst
@@ -63,6 +63,12 @@ is mandatory to implement by applications, though. Additionally there are a few
 backward incompatible changes in callbacks that might require attention by some
 applications:
 
+tls_record_received() / tls_emit_data()
+"""""""""""""""""""""""""""""""""""""""
+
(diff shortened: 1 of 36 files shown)


randombit (approved):
Especially with the callback changes, if we are ever going to do it now is the time. Thanks for your efforts on this.

PR #3364, opened 2023-03-13
Chore: Update BoringSSL's test suite to latest master

reneme:
Note that a few tests needed to be disabled, but no additional failures were found. In a future release, we might consider to build explicit support for a FIPS 202205 policy.

TODO: @randombit could you push `reneme/rene/20230313` to your fork of BoringSSL, please? After that I'd point the CI to that. Currently, also `master` runs against my fork.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,7 +155,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: reneme/boringssl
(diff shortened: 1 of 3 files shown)


randombit:
@reneme I just added you as a collaborator on that repo so you can push a branch directly.

randombit (approved):
:tada:

Issue #3367, opened 2023-03-14
Dilithium speed tests suddenly failing


randombit:
Abruptly, `speed Dilitihum` starts reporting generation of invalid signatures.

First seen in CI for #3366 but now I can replicate it on master on my local machine. And, checking out older revisions, I can replicate it there as well. No idea what is happening.

a) What's wrong with the signatures
b) Why did it not fail in CI before today
c) Why did nobody notice it failing locally before today

Real WTF

randombit:
Verification had the same bug as fixed in #3320 but I do not get why this suddenly manifested, weeks later.

randombit:
OK, a theory on why this wasn't detected earlier.

We only noticed the problem because of the CLI tests. These run `botan speed` but using `--msec=1` to minimize the test runtime.

As a consequence of this, the cli only tended to actually verify a few signatures. If the CI VM was relatively slow, then we might just generate a single signature, verify that signature, then verify an incorrect signature, and then exit. And then perhaps GH upgraded their machines, or just the timing of things changed in some other way, and then the test failed because we started trying to verify more than a single signature during the cli tests.

At least that's the best I can come up with. I did confirm that locally, running the CLI tests failed from a revision last week.

Still a pretty big wtf but I'm not sure what could be checked further wrt post mortem, so closing.

PR #3368, opened 2023-03-14
Fix Dilithium verification

randombit:
GH #3367


reneme (approved):
😲

The speed test did try multiple verifications before, right?

randombit:
Yes! And it used a cached `PK_Verifier`. Nothing has changed here in at least two weeks. And the most baffling part is, of course it *never worked correctly* - I can check out a revision from last week, and it fails for me on my local machine. And yet CI passed this entire time, and the CLI test script checks the output against a reject that the error message does not match. I have no idea what happened here, and I don't like it.

PR #3376, opened 2023-03-14
Regression Test: trying to reproduce GH #3342

reneme:
This validates that throwing an exception in TLS::Callbacks::tls_session_established results in a graceful closure of the TLS connection with an appropriate alert code.

Forward port to `master` of: https://github.com/randombit/botan/pull/3348


randombit (approved):
:+1:

PR #3385, opened 2023-03-15
Support signing with legacy XMSS private keys

reneme:
With the move towards the NIST SP.800-208 key generation, we broke the relationship between legacy private keys and their public keys. This is working towards a "legacy XMSS mode" to (optionally) allow re-enabling the old key derivation.


reneme:
Closing in favor of #3388

PR #3388, opened 2023-03-16
Implement legacy support for old XMSS_PrivateKeys

FAlbertDev:
XMSS private keys created before the NIST SP.800-208 adaptations cannot create new valid signatures in the newer version. Therefore, private keys can now be explicitely loaded using the `from_legacy_key` factory. These private keys are handled without the adaptions and can therefore be used as before.


reneme (changes requested):
Looks good already, thanks. Two general things:

1. [ ] We'll need to mention this in the `migration.rst` so that potential users with legacy private keys get fair warning.
2. [ ] We might want to be more precise than "legacy", see below.

This part of the key generation is deliberately under-specified in RFC 8391. Botan 2.x (roughly) followed the approach sketched in [RFC 8391 Section 3.1.7](https://www.rfc-editor.org/rfc/rfc8391.html#section-3.1.7). NIST has published other recommendations with SP.800-208. But other implementations might go their own way.

To clarify this in the API, we might consider introducing an enum like so:

```C++
enum class WOTSP_Derivation_Method
   {
   /// This roughly followed the suggestions in RFC 8391 but is vulnerable
   /// to a multi-target attack. For new private keys, we recommend using
   /// the derivation as suggested in NIST SP.800-208.
   /// Private keys generated with Botan 2.x will need to stay with this mode,
   /// otherwise they won't be able to generate valid signatures any longer.
   Botan2x,

   /// Derivation as specified in NIST SP.800-208 to avoid a multi-target attack
   /// on the WOTS+ key derivation suggested in RFC 8391. New private keys
   /// should use this mode.
   NIST_SP800_208,
   };
```

and use that instead of the boolean `is_legacy_key`. In principle we could even add more derivation methods in the future.

src/lib/pubkey/xmss/xmss_privatekey.cpp
```diff
@@ -437,7 +442,9 @@ XMSS_WOTS_PublicKey XMSS_PrivateKey::wots_public_key_for(XMSS_Address& adrs, XMS
 
 XMSS_WOTS_PrivateKey XMSS_PrivateKey::wots_private_key_for(XMSS_Address& adrs, XMSS_Hash& hash) const
    {
-   return XMSS_WOTS_PrivateKey(m_private->wots_parameters(), m_public_seed, m_private->private_seed(), adrs, hash);
+   return (is_legacy_key())
+      ? XMSS_WOTS_PrivateKey(m_private->wots_parameters(), m_private->private_seed(), adrs, hash)
+      : XMSS_WOTS_PrivateKey(m_private->wots_parameters(), m_public_seed, m_private->private_seed(), adrs, hash);
```
reneme, line 447: Rule number 41712: Never trust my inline code suggestions. 😶 

```
src/lib/pubkey/xmss/xmss_privatekey.cpp:446:97: error: ‘Botan::XMSS_WOTS_PrivateKey::XMSS_WOTS_PrivateKey(Botan::XMSS_WOTS_Parameters, std::span<const unsigned char>, Botan::XMSS_Address, Botan::XMSS_Hash&)’ is private within this context
  446 |       ? XMSS_WOTS_PrivateKey(m_private->wots_parameters(), m_private->private_seed(), adrs, hash)
      |                                                                                                 ^
In file included from build/include/botan/internal/xmss_signature.h:16,
                 from build/include/botan/internal/xmss_signature_operation.h:14,
                 from src/lib/pubkey/xmss/xmss_privatekey.cpp:20:
build/include/botan/internal/xmss_wots.h:171:7: note: declared private here
  171 |       XMSS_WOTS_PrivateKey (XMSS_WOTS_Parameters params,
      |       ^~~~~~~~~~~~~~~~~~~~
make: *** [Makefile:904: build/obj/lib/pubkey_xmss_privatekey.o] Error 1
```

src/lib/pubkey/xmss/xmss.h
```diff
@@ -158,6 +158,24 @@ template<typename> class Atomic;
...
+/// the derivation as suggested in NIST SP.800-208.
+/// Private keys generated with Botan 2.x will need to stay with this mode,
+/// otherwise they won't be able to generate valid signatures any longer.
+Botan2x,
+
+/// Derivation as specified in NIST SP.800-208 to avoid a multi-target attack
+/// on the WOTS+ key derivation suggested in RFC 8391. New private keys
+/// should use this mode.
+NIST_SP800_208,
+};
```
reneme, line 177: Code style: 
```suggestion
   {
   /// This roughly followed the suggestions in RFC 8391 but is vulnerable
   /// to a multi-target attack. For new private keys, we recommend using
   /// the derivation as suggested in NIST SP.800-208.
   /// Private keys generated with Botan 2.x will need to stay with this mode,
   /// otherwise they won't be able to generate valid signatures any longer.
   Botan2x,
   
   /// Derivation as specified in NIST SP.800-208 to avoid a multi-target attack
   /// on the WOTS+ key derivation suggested in RFC 8391. New private keys
   /// should use this mode.
   NIST_SP800_208,
   };
```

src/lib/pubkey/xmss/xmss.h
```diff
@@ -250,9 +274,13 @@ class BOTAN_PUBLIC_API(2,0) XMSS_PrivateKey final : public virtual XMSS_PublicKe
        **/
       secure_vector<uint8_t> raw_private_key() const;
 
+      WOTS_Derivation_Method wots_derivation_method() const { return m_wots_derivation_method; }
+
    private:
       friend class XMSS_Signature_Operation;
 
+      XMSS_PrivateKey(std::span<const uint8_t> raw_key, bool is_legacy_key);
+
```
reneme, line 283: That's a left-over from the previous iteration (using the `is_legacy_key` boolean). It should go, because it doesn't have an implementation and is unused.

src/lib/pubkey/xmss/xmss.h
```diff
@@ -293,6 +321,8 @@ class BOTAN_PUBLIC_API(2,0) XMSS_PrivateKey final : public virtual XMSS_PublicKe
                              XMSS_Hash& hash);
 
       std::shared_ptr<XMSS_PrivateKey_Internal> m_private;
+
+      const WOTS_Derivation_Method m_wots_derivation_method;
```
reneme, line 325: Defining this as `const` will (implicitly) remove the class' move constructor and move assignment operator. I don't think we want that.

```suggestion
      WOTS_Derivation_Method m_wots_derivation_method;
```

src/lib/pubkey/xmss/xmss_privatekey.cpp
```diff
@@ -437,7 +441,15 @@ XMSS_WOTS_PublicKey XMSS_PrivateKey::wots_public_key_for(XMSS_Address& adrs, XMS
 
 XMSS_WOTS_PrivateKey XMSS_PrivateKey::wots_private_key_for(XMSS_Address& adrs, XMSS_Hash& hash) const
    {
-   return XMSS_WOTS_PrivateKey(m_private->wots_parameters(), m_public_seed, m_private->private_seed(), adrs, hash);
+   switch (wots_derivation_method())
```
reneme, line 444: Code style:

```suggestion
   switch(wots_derivation_method())
```

src/lib/pubkey/xmss/xmss_wots.h
```diff
@@ -127,6 +127,24 @@ class XMSS_WOTS_PrivateKey : public XMSS_WOTS_Base
...
+       * @param params       The WOTS parameters to use
+       * @param private_seed The private seed for the private key generation
+       * @param adrs         The address of the key to retrieve.
+       * @param hash         Instance of XMSS_Hash, that may only be used by the
+       *                     thread executing it.
+       **/
+      XMSS_WOTS_PrivateKey (XMSS_WOTS_Parameters params,
+                            std::span<const uint8_t> private_seed,
+                            XMSS_Address adrs,
+                            XMSS_Hash& hash);
```
reneme, line 146: Code style:

```suggestion
      XMSS_WOTS_PrivateKey(XMSS_WOTS_Parameters params,
                           std::span<const uint8_t> private_seed,
                           XMSS_Address adrs,
                           XMSS_Hash& hash);
```

src/tests/test_xmss.cpp
```diff
@@ -207,12 +207,122 @@ std::vector<Test::Result> xmss_statefulness()
...
+         result.confirm("legacy private key generates signatures that are still verifable",
+                        verifier.verify_message(message, signature));
+         }),
+
+      Botan_Tests::CHECK("Verify a legacy signature", [&](auto& result)
+         {
+         Botan::PK_Verifier verifier(public_key, algo_name);
+         result.confirm("legacy private key generates signatures that are still verifable",
+                        verifier.verify_message(message, legacy_signature));
+         }),
```
reneme, line 316: Maybe we should add another test that verifies a legacy signature with a `Public_Key` object that was obtained by straight from a legacy public key. This particular `public_key` was obtained as the base class of the legacy private key.

I.e. have  another public key that is instantiated like so:

```C++
const auto legacy_public_key = Botan::XMSS_PublicKey(Botan::hex_decode("04818800000001C49B128365E23D97C2D36A2AC26E0B448FCEF77A5ED68EFDC460E787D0EDB6768EAC3E82B04B84571C49684ED822F8179FDFF1C825DA9F9EA78F5D50A21E6341");
```

and create a third `CHECK`-context that verifies a new signature (created by the legacy private key) with the `legacy_public_key`. My main objective: Make sure that the `WOTS_Derivation_Method::Botan2x` used in the creation of the existing `public_key` object did not have a side effect. I.e. even if no such extra step is involved, the validation still works.

We can do that in pair-programming, if you want. Not sure whether I made myself clear enough, tbh. 😄

reneme (changes requested):
Yeah, I think the `WOTS_Derivation_Method` enum makes it much clearer. Thanks.
Below are a few nits and code style comments. Plus an idea for another test that we might want to tackle together.

src/tests/test_xmss.cpp
FAlbertDev: I think I understood your suggestion. Let's discuss tomorrow, how to procede :) 
Thanks for the review btw!

src/lib/pubkey/xmss/xmss.h
FAlbertDev: Ops, I missed that...

reneme (approved):

randombit (approved):
So, I missed that #3280  was breaking something, and maybe everyone missed that until after it was merged - but I'm put off that we didn't have a test that noticed this. I'm not sure if there is anything specifically actionable there, but it's disquieting.

The change itself seems fine but I'm concerned that PKCS8 encoded XMSS keys don't have any metadata (AIUI) that allow us to check which mode we are in. For anyone using XMSS, this will lead to very hard to debug issues, where their existing keys no longer work and it will not really be clear why.

Approving this change since it seems like a good start but I think some further work is required. Perhaps we can add some extra field in the XMSS DER encoded private key to indicate which mode we are in?

reneme:
> So, I missed that #3280 was breaking something, and maybe everyone missed that until after it was merged - but I'm put off that we didn't have a test that noticed this. I'm not sure if there is anything specifically actionable there, but it's disquieting.

This is on me! I had to regenerate the `xmss_sig.vec` but not `xmss_verify.vec` in #3280 after switching to the NIST generation. Unfortunately, those vectors do not cover the scenario that an old key is capable of generating valid signatures with the new code -- the regenerated `xmss_sig.vec` contains private key generated with the new NIST approach.

Nevertheless, I convinced myself that no further action was required to keep backward compatibility. Despite having discussed that it might become a concern internally and with BSI. I should have written the tests added in this patch already then. Sorry for that. 😞 

Eventually, we discovered the issue while writing up documentation for the adapted XMSS key generation for BSI.

> Perhaps we can add some extra field in the XMSS DER encoded private key to indicate which mode we are in?

Fully on board with this! It seems that interoperability in the private key format of XMSS is not really a concern anyway. Thinking about it: Due to the statefulness one might even consider it a non-goal. Following up in #3402.

PR #3392, opened 2023-03-20
std::span for Botan::Cipher_Mode

reneme:
Two things:

* allow `Cipher_Mode::process()` and `Cipher_Mode::start()` for arbitrary contiguous containers
* allow `concepts::resizable_byte_buffer` for the in/out parameters of `Cipher_Mode::update()` and `Cipher_Mode::finish()`.

To do that properly a "virtual entrypoint" indirection was introduced. Namely `Cipher_Mode::process_msg()` and `Cipher_Mode::finish_msg()`.

Note that `Cipher_Mode::finish()` with anything but `Botan::secure_vector<uint8_t>` will introduce additional data copies. We might be able to address this later on (by generalizing the virtual entrypoint for `Cipher_Mode`). But that felt entirely out of scope here. FWIW: Users that had their data stored in some other container were previously forced to do the same copy ceremony before in their application code. That's a hot take, maybe. I'm open for better approaches.

### DIFF
--- a/src/lib/modes/aead/ccm/ccm.cpp
+++ b/src/lib/modes/aead/ccm/ccm.cpp
@@ -117,7 +117,7 @@ void CCM_Mode::start_msg(const uint8_t nonce[], size_t nonce_len)
    m_msg_buf.clear();
    }
 
-size_t CCM_Mode::process(uint8_t buf[], size_t sz)
+size_t CCM_Mode::process_msg(uint8_t buf[], size_t sz)
    {
    BOTAN_STATE_CHECK(!m_nonce.empty());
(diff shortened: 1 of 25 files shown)


randombit (approved):

randombit:
> I read that you don't foresee this for Botan 3.0, right?

No, not at this stage of things certainly. I don't even know what exactly a new API would look like at this point.

> With today's PRs I'm trying to reach a somewhat consistent API re: the adoption of std::span and concepts::resizable_byte_buffer.

:+1: absolutely, this PR series is a good improvement over status quo

randombit (approved):

Issue #2742, opened 2021-05-12
ECKCDSA incorrect hash truncation


randombit:
Current behavior of ECKCDSA does not match ISO 14888-3; hash truncation occurs by taking bits from the left rather than the right. It seems this is due to an error in BSI TR 03111.

Test vectors taken from the ISO spec are included in
https://github.com/ANSSI-FR/libecc/blob/f3abf297e84874278129fb91bc1b89356dd66627/src/tests/ec_self_tests_core.h#L1042

@securitykernel I plan to address this in 3.x by modifying ECKCDSA, which will break compat with 2.x behavior when the hash and group order are not equal in size. LMK if this is an issue / if we need to form some alternate approach.

securitykernel:
Interesting. I remember it was hard to find any test vectors at all. BSI TR 03111 does not mention ECKCDSA anymore specifically, but BSI TR 02102-1 still lists it as a recommended algorithm. I will check older versions of TR 031111 on Monday and if there is an error, check back with BSI. Do you have a recent version of ISO 14888-3 at hand?

randombit:
> Do you have a recent version of ISO 14888-3 at hand?

I do not. This was reported to me by @rb-anssi who indicated that the tests in libecc were taken directly from the ISO doc.

The spec referenced in our test vectors (http://www.tta.or.kr/include/Download.jsp?filename=stnfile/TTAK.KO-12.0015_R2.pdf) does seem to include some tests where the curve order and hash do not match, but only for GF(2^m) curves. However this indicates the spec must somewhere specify how truncation occurs. I'll try feeding pieces into Google translate to see if I can figure that out. [Of course this Korean spec may not match the final ISO spec in any case, standards being what they are]

rb-anssi:
Hi,

Thanks @randombit for the references, and your link on the Korean standard rang a bell!

After some digging, it seems that there is actually a **third revision R3** of this Korean standard (http://www.tta.or.kr/include/Download.jsp?filename=stnfile/TTAK.KO-12.0015_R3.pdf) where the test vector from ISO14888-3 can be found in appendix II.2 (secp224r1 with SHA-256). This is actually in line with the 2016 revision of ISO where this hash truncation matter has been revised (or rather made unambiguous) and detailed compared to the previous editions of the standards.

My Korean being as bad as yours, after a "Google translate" operation, the truncation explanation seems to be detailed in the last paragraph of 9.1, and seems (as intelligible as the translation can be) to be in line with the ISO description.

[...]

PR #3393, opened 2023-03-20
Support hash truncation in ECKCDSA (#2742)

lieser:


randombit (commented):
It looks like this only handles truncation in the case where the difference between the group order and the hash length is a multiple of 8, is that correct? If so I think we should still reject the other cases rather than producing an incorrect signature. (Or ideally, extend this patch to also cover them.)

Examples of curves with this issue are secp160r1 and x962_p239v1 since there there curve order is not a multiple of 8

lieser:
I will take a look again into the ISO standard (and maybe extend the docstring) but from my current understanding the truncation is defined on a byte and not bit level.

See also the implantation in libecc and the detailed explanation:
https://github.com/libecc/libecc/blob/eb4b76c3499c745f6fab45604bcb2f2412e0e4af/src/sig/eckcdsa.c#L101

randombit (approved):
Approving modulo the bit vs byte truncation question. If it's really is bytes (as the libecc comment seems to indicate [*]), just updating the comment to explain is sufficient.

[*] In which case, WTF

PR #3398, opened 2023-03-20
std::span for Key Derivation Function

reneme:
### Pull Request Dependencies

* #3392 
  (because we need `concepts::resizable_byte_buffer`)

### Description

Simply translated `KDF::derive_key()` overloads to use `std::span`, `std::string_view` and made its return type configurable as a `concepts::resizable_byte_buffer`.

I feel we should think about deprecating a few of those overloads, though. Not that they are expensive, but the combinatorics with the provided parameter types make the API rather extensive.

### DIFF
--- a/src/lib/kdf/kdf.h
+++ b/src/lib/kdf/kdf.h
@@ -9,8 +9,11 @@
 #define BOTAN_KDF_BASE_H_
 
 #include <botan/secmem.h>
+#include <botan/concepts.h>
 #include <botan/exceptn.h>
 #include <string>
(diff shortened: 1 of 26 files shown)


randombit:
> I feel we should think about deprecating a few of those overloads, though. Not that they are expensive, but the combinatorics with the provided parameter types make the API rather extensive.

KDF is one of those types that has accreted over time without ever being pruned, likely we could deprecate half the interfaces without problem.

randombit (approved):

PR #3407, opened 2023-03-21
Use sccache 0.4.0

randombit:


randombit:
Ugh OK whatever, just going to wait until Chocolately promotes it to a standard install.

PR #3409, opened 2023-03-22
[TLS 1.3] Dependency Lifetimes

### LINKED ISSUE #3321: Clarify lifetime of TLS stack dependencies
Currently, `TLS::Client` and `TLS::Server` take a number of dependency objects by reference in their constructors. The application is responsible for keeping those objects alive as long as the TLS stack is alive.
Some or all of those dependencies might be shared with multiple instances of the TLS stack.

This is a follow-up to [a discussion in the `Session_Manager` restructuring PR](https://github.com/randombit/botan/pull/3150#discussion_r1113287945).

The "dependency objects" in question are:

* `TLS::Callbacks`
* `TLS::Session_Manager`
* `TLS::Policy`
* `Credentials_Manager`
* `RandomNumberGenerator`

To me, the most obvious change would be to replace the references by `std::shared_ptr<>`.

reneme:
This makes use of `std::shared_ptr<>` for auxiliary objects in `TLS::Server` and `TLS:Client`. The implementations keep ownership of the `shared_ptr`s internally. See (mostly) `TLS_Channel_Impl_12` and `TLS_Channel_Impl_13`. The `Channel`s retain ownership and hand references to the auxiliary objects downstream through the rest of the implementations.
Also, the `Session_Manager`s now retain a `shared_ptr` of `RandomNumberGenerator` (and `Credentials_Manager` for some).

Most of the changes just adapt the consuming code in CLI, tests, fuzzers and examples. I tried to split minor preparatory refactorings into their own commits.

Closes #3321


src/lib/tls/tls12/tls_client_impl_12.cpp
```diff
@@ -61,19 +61,20 @@ class Client_Handshake_State_12 final : public Handshake_State
...
-                               const Server_Information& info,
+Client_Impl_12::Client_Impl_12(std::shared_ptr<Callbacks> callbacks,
+                               std::shared_ptr<Session_Manager> session_manager,
+                               std::shared_ptr<Credentials_Manager> creds,
+                               std::shared_ptr<const Policy> policy,
+                               std::shared_ptr<RandomNumberGenerator> rng,
+                               Server_Information info,
                                bool datagram,
-                               const std::vector<std::string>& next_protocols,
+                               std::vector<std::string> next_protocols,
```
randombit, line 71: a) I have a feeling clang-tidy will complain about this
b) I don't really understand the motivation for the change here

randombit:
@reneme Just wanted to check in on this one, is there anything you need from me here? I am holding of on the RC until this merges since it has a big impact on the public facing API. No worries if there is a little delay, we can stick to April 10th final release.

reneme:
Ready to go.

PR #3410, opened 2023-03-22
Fix some clang-tidy warnings

randombit:


reneme (approved):
Thanks! The changes in `tls_callbacks.h` need to be reflected in `examples/tls_custom_curves_client.cpp`, though.

Issue #3328, opened 2023-02-26
Use simplified encodings for Kyber and Dilithium keys


randombit:
Currently we follow (just approximately) an internet draft for the X.509 encoding, but using private allocation OIDs.

This is actually the worst possible situation because

* If the draft is finalized in a way that doesn't match exactly what we're doing, then we'll have to support the final RFC as well as our current format. I know we are incompatible with the current draft, and the current draft anyway seems very half-baked with numerous underspecified parts, so I don't think it's even possible to *be* compatible with the draft without guessing.
* If the draft is abandoned, then we'll be using this one off format that is incompatible with OQS and similar code

To minimize long term pain I think the best course of action is to remove the existing ASN.1 based key encodings for Kyber and Dilithium, and have our private allocation OIDs simply export the "Raw" encoding as used by the reference implementation as an OCTET STRING. That's a key format we'll have to continue to support *anyway* since it looks like the ecosystem will end up splitting into "raw octets as represented by the reference code" vs "destructuring the public key into ASN.1 for no particular reason". But at least this way we only end up having to support at most one of the latter.

@reneme @securitykernel thoughts on this?

reneme:
I tend to agree, too. As long as these algorithms are not-yet-standardized they'll likely be used for proof of concept and experimental work anyway. Compatibility with the reference implementations and OQS should therefore be the first priority. We implemented those extra encodings mainly because "we could" without an actual use case.

That said, I'm wondering about the overall compatibility guarantees we want to give for those algorithms in the near future. After all they might still change during the standardization process. Maybe we should [reconsider the experimental modules](https://github.com/randombit/botan/discussions/3110) for Kyber and Dilithium: In that case I wouldn't see much harm for leaving the extra encodings in.

reneme:
I tend to agree, too. As long as these algorithms are not-yet-standardized they'll likely be used for proof of concept and experimental work anyway. Compatibility with the reference implementations and OQS should therefore be the first priority. We implemented those extra encodings mainly because "we could" without an actual use case.

That said, I'm wondering about the overall compatibility guarantees we want to give for those algorithms in the near future. After all they might still change during the standardization process. Maybe we should [reconsider the experimental modules](https://github.com/randombit/botan/discussions/3110) for Kyber and Dilithium: In that case I wouldn't see much harm for leaving the extra encodings in.

Issue #3318, opened 2023-02-23
Extend public APIs with `std::span` overloads


reneme:
Alongside existing `std::vector<T, Alloc>` and C-style ptr-length APIs we're gradually introducing overloads taking [a `std::span`](https://en.cppreference.com/w/cpp/container/span) from C++20. This ticket is an attempt to create/maintain a list of interfaces that should get such overloads.

### Public APIs and Interfaces

* [ ] `RandomNumberGenerator` (https://github.com/randombit/botan/pull/3195)
* [ ] `Buffered_Computation` (https://github.com/randombit/botan/pull/3294)
* [ ] `Symmetric_Algorithm` (https://github.com/randombit/botan/pull/3297)
* [ ] `AEAD_Mode` (https://github.com/randombit/botan/pull/3317)

### Internal APIs and Interfaces

* [ ] `Signature` (pk_ops.h)

randombit:
Another one, from #3411 

```
      Kyber_PublicKey(const AlgorithmIdentifier& alg_id,
-                      const std::vector<uint8_t>& key_bits);
+                      std::span<const uint8_t> key_bits);
```

Would be nice to do this for all of the key types.

(This might require some work on `BER_Decoder` though ...)

crazydef:
It would be nice if Botan detected the C++ version in the headers and didn't expose unsupported/new features to C++ compilers that didn't support them.

I am currently trying to integrate Botan into a codebase that can't be built with C++20 (because other libraries we're using don't support it yet) and having to wrap everything is becoming a royal pain in the ass.

randombit:
@crazydef Don't do that. Use Botan2, which is still supported and probably will be for years, and is C++11.

Or (for some functionality) you can use the C89 API in `ffi.h`

lieser:
> because other libraries we're using don't support it yet

Do this libraries really break if compiled with C++20? I would expect most older code to work without a problem if compiled with a newer C++ standard.

crazydef:
@lieser C++17 deprecated a lot of stuff and removed even more.  One particular library I have to deal with still only supports C++14 because of the mess that is C++17.

CharlesJQuarra:
I'm just trying to build from tag 3.2.0, and I'm seeing the `#include <span>` breaking compilation for C++17. make is building with the following options:

[...]

PR #3411, opened 2023-03-23
Kyber Encoding Improvements

reneme:
This removes the half-baked DER encoding from Kyber (see #3328). Also it reduces copies and code complexity in Kyber's parsing code. The two changes are in individual commits.


src/lib/pubkey/kyber/kyber_common/kyber.h
```diff
@@ -56,24 +58,16 @@ class BOTAN_PUBLIC_API(3, 0) KyberMode
...
    {
    public:
-      Kyber_PublicKey(const std::vector<uint8_t>& pub_key,
-                      KyberMode mode,
-                      KyberKeyEncoding encoding);
+      Kyber_PublicKey(std::span<const uint8_t> pub_key, KyberMode mode);
 
       Kyber_PublicKey(const AlgorithmIdentifier& alg_id,
-                      const std::vector<uint8_t>& key_bits);
+                      std::span<const uint8_t> key_bits);
```
randombit, line 70: It would be nice to make this change for all of the key types before 3.0

(This is the sort of minor thing that can be done after the RC)

Left a comment in #3318

randombit (approved):
Thanks for taking this on

PR #3412, opened 2023-03-23
Remove DER encoding support in Dilithium

reneme:
As discussed [here](https://github.com/randombit/botan/issues/3328). Additionally, this sprinkles a little more `std::span<>` into the public interface (in an independent commit).


randombit (approved):
:+1:

PR #3413, opened 2023-03-23
XMSS encodes the WOTS+ key derivation into the private key

reneme:
### Pull Request Dependencies

* #3411
  (to be able to use `BufferSlicer` that was made globally available there)

### Description

First, this improves the parsing-code of XMSS keys using `BufferSlicer`, resulting in much less manual buffer offset calculation. 

More importantly, it adds the selected WOTS+ key derivation method to the private key encoding (as a one-byte-suffix). Legacy keys are detected by the absence of this byte and are treated accordingly. Once a legacy private key was used for a signature, it will gain the declaring byte due to the stateful nature of XMSS. See also: https://github.com/randombit/botan/pull/3388

This re-introduces a number of test vector entries from Botan 2.x to verify proper functionality.


randombit (approved):
Only reviewed the last 2 commits here

Issue #2710, opened 2021-04-10
Re: Slow DH derivation


guidovranken:
In response to your question why I constrained inputs to DH_Derive: https://github.com/guidovranken/cryptofuzz/commit/52fa502b21ad37ad49cd9062955ec38c28c7dd85#r49315676

I'm using Botan as an oracle in several of my OSS-Fuzz projects (Nettle, wolfCrypt, BearSSL, relic etc) and I noticed that Botan's DH derivation would consistently be involved in time outs, so I decided to constrain inputs to random low values (I picked 80 bytes) so that doesn't happen anymore.

Here's an example:

Compile flags used:

```
export CC="clang"
export CXX="clang++"
export CFLAGS="-m32 -fsanitize=fuzzer-no-link -g -O1 -fsanitize=address -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION"
export CXXFLAGS="-m32 -fsanitize=fuzzer-no-link -g -O1 -fsanitize=address -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION"
export LDFLAGS="-m32 -fsanitize=fuzzer-no-link -fsanitize=address"
```

Botan configuration used:

```
./configure.py --cpu=x86_32 --cc-bin=$CXX --cc-abi-flags="$CXXFLAGS" --disable-shared --disable-modules=locking_allocator --build-targets=static --without-documentation
```

```cpp
#include <botan/dh.h>
#include <botan/pubkey.h>
#include <botan/system_rng.h>

int main(void)
{
    ::Botan::System_RNG rng;

    const ::Botan::BigInt g("2251799813685248");
[...]

PR #3424, opened 2023-03-25
Remove blinding from Diffie-Hellman

randombit:
Investigating #2710 showed that the majority of the reason for Diffie-Hellman being slower than other implementations was the use of blinding. This converts short-exponent Diffie-Hellman into full-exponent Diffie-Hellman; fully a factor of 5x to 16x decrease in performance, depending on the group size. This is not really worthwhile given our modular exponentiation is anyway constant time.

This does leak the bit length of the Diffie-Hellman secret key, but unlike say an ECDSA nonce this does not help an attacker.

### DIFF
--- a/src/lib/pubkey/dh/dh.cpp
+++ b/src/lib/pubkey/dh/dh.cpp
@@ -8,7 +8,6 @@
 #include <botan/dh.h>
 #include <botan/internal/dl_scheme.h>
 #include <botan/internal/pk_ops_impl.h>
-#include <botan/internal/blinding.h>
 
 namespace Botan {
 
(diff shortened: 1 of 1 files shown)


randombit:
Wow this analysis is totally wrong, actually the DH blinding is done by `((k*v)^x)/(k^x)` and does not affect the exponent sizes at all. (Because in DH we don't have a good way of even knowing the group order, so exponent blinding isn't really possible)

The real problem was we were not indicating the size of `x` to the modular exponentiation function so it defaulted to using a loop equal to `p` bits.

randombit:
#3426

PR #3426, opened 2023-03-25
In DH bound the exponent size by the private key size

randombit:
This leaks the length of the private key to side channels but this is not a problem, and doing so improves the performance by ~5-16 times.


randombit:
Was looking into backporting this but it turned out to be a performance regression vs 2.x

PR #3428, opened 2023-03-26
Fixed bcrypt WinRT build

oviano:
Without these changes there are a ton of build errors with WinRT/UWP.

### DIFF
--- a/src/lib/rng/system_rng/system_rng.cpp
+++ b/src/lib/rng/system_rng/system_rng.cpp
@@ -17,7 +17,9 @@
 #if defined(BOTAN_TARGET_OS_HAS_RTLGENRANDOM)
   #include <botan/internal/dyn_load.h>
 #elif defined(BOTAN_TARGET_OS_HAS_CRYPTO_NG)
+  #include <windows.h>
   #include <bcrypt.h>
+  #define STATUS_SUCCESS 0x00000000
 #elif defined(BOTAN_TARGET_OS_HAS_CCRANDOM)
   #include <CommonCrypto/CommonRandom.h>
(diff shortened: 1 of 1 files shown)


src/lib/rng/system_rng/system_rng.cpp
```diff
@@ -17,7 +17,9 @@
 #if defined(BOTAN_TARGET_OS_HAS_RTLGENRANDOM)
   #include <botan/internal/dyn_load.h>
 #elif defined(BOTAN_TARGET_OS_HAS_CRYPTO_NG)
+  #include <windows.h>
   #include <bcrypt.h>
+  #define STATUS_SUCCESS 0x00000000
```
randombit, line 22: The include I can understand but what's the purpose of this define? We don't use `STATUS_SUCCESS` anywhere in our own code.

oviano: My apologies that define was related to an earlier version of Botan and was fixed a while back (https://github.com/randombit/botan/commit/00a6d3832905531328f14ebc5ffc736d0d4297c7)

I will remove it.

randombit: Oh you may need to use the 

```
  #define NOMINMAX 1
  #define _WINSOCKAPI_ // stop windows.h including winsock.h
```

dance before including `windows.h` or the amalgamation builds will run into issues.

oviano: Well apparently not, as I use the amalgamation build myself and it seems fine without those lines.

randombit: OK great if it works, maybe this crap is unnecessary on modern systems.

randombit (approved):

PR #3429, opened 2023-03-26
Added Ed25519_PrivateKey/Curve25519_PrivateKey raw_private_key_bits

oviano:
I needed access to the raw private key data, not the encoded one available from private_key_bits, so I added these functions.


randombit:
If I'm understanding correctly, this functionality already exist in the form of the (poorly named!) `Curve25519_PrivateKey::get_x` and `Ed25519_PrivateKey::get_private_key`

(OTOH your names are better, we might want to introduce them anyway and deprecate the old names)

oviano:
No you are not mistaken, I overlooked that, thank you!

PR #3430, opened 2023-03-26
Added external_allocator

oviano:
This allows using your own memory allocator in Botan.

### DIFF
--- a/src/lib/utils/external_allocator/external_allocator.h
+++ b/src/lib/utils/external_allocator/external_allocator.h
@@ -1,0 +1,39 @@
+/*
+* External Allocator
+* (C) 2022 Oliver Collyer
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#ifndef BOTAN_EXTERNAL_ALLOCATOR_H_
+#define BOTAN_EXTERNAL_ALLOCATOR_H_
+
+#include <botan/types.h>
+
+//BOTAN_FUTURE_INTERNAL_HEADER(external_allocator.h)
+
+namespace Botan {
+
+class BOTAN_PUBLIC_API(2,0) external_allocator final
+   {
+   public:
+      static external_allocator& instance();
+
+      void* allocate(size_t num_elems, size_t elem_size);
+
+      bool deallocate(void* p, size_t num_elems, size_t elem_size) noexcept;
+
+      external_allocator(const external_allocator&) = delete;
+
+      external_allocator& operator=(const external_allocator&) = delete;
+
+   private:
+      external_allocator();
+
+      ~external_allocator();
+   };
+
+}
+
+#endif
+
--- a/src/lib/utils/external_allocator/info.txt
+++ b/src/lib/utils/external_allocator/info.txt
@@ -1,0 +1,12 @@
+<defines>
+EXTERNAL_ALLOCATOR -> 20131128
+</defines>
+
+<module_info>
+name -> "External Allocator"
+brief -> "STL allocator using an external allocator"
+</module_info>
+
+<header:public>
+external_allocator.h
+</header:public>
--- a/src/lib/utils/mem_ops.cpp
+++ b/src/lib/utils/mem_ops.cpp
@@ -12,6 +12,8 @@
 
 #if defined(BOTAN_HAS_LOCKING_ALLOCATOR)
   #include <botan/internal/locking_allocator.h>
+#elif defined(BOTAN_HAS_EXTERNAL_ALLOCATOR)
+  #include <botan/external_allocator.h>
 #endif
 
 namespace Botan {
@@ -28,6 +30,9 @@ BOTAN_MALLOC_FN void* allocate_memory(size_t elems, size_t elem_size)
 
 #if defined(BOTAN_HAS_LOCKING_ALLOCATOR)
    if(void* p = mlock_allocator::instance().allocate(elems, elem_size))
+      return p;
+#elif defined(BOTAN_HAS_EXTERNAL_ALLOCATOR)
+   if(void* p = external_allocator::instance().allocate(elems, elem_size))
       return p;
 #endif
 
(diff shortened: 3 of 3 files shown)


randombit (commented):
I'm confused as to how this would work. Is the idea that the application would define the implementation of `Botan::external_allocator` within its own source? It seems like then it would fail to link if the user did not define such an allocator.

Also can you speak a bit to the motivation here? The main purpose of this allocator interface is to provide backing storage just for key material type things; we use `std::vector` or `std::string` which are backed by the standard heap for everything else. If there is something that we could do better, wrt protecting key material (mlock, zeroization, etc) we should just do it in the library.

oviano:
Sure - so if you wanted to use your own allocator, you'd enable this module and then yes, it would require you to provide the code for the missing functions. Otherwise you just wouldn't define it. I think it needs disabling by default, how would I do that?

Motivation - yes, my project uses its own memory allocator, and it's nice to be able to continue to use the same one.

randombit:
> I think it needs disabling by default, how would I do that?

If you add `load_on request` to the `info.txt` file then it will only be built if `--enable-modules=external_allocator` is used.

OTOH I'm not sure if this gets you what you are looking for because then your external allocator is only called if you have a special one-off build, and otherwise it is ignored.

It seems like a better approach here might be something like

```
typedef std::function<void* (size_t, size_t)> alloc_fn;
typedef std::function<void (void*, size_t, size_t)> free_fn;
void set_memory_allocation_functions(alloc_fn alloc, free_fn free);
```

```
// when allocating ...
if(g_alloc_fn)
   return g_alloc_fn(...);
```

I also want to reiterate again that this allocator interface is only ever called for `secure_vector` and in general Botan uses the standard C++ `new`/`delete` heap for most things.

Is the source of this allocator you are using public? If not can you describe what it is doing that we are missing from the current mlock pool that makes using your version desirable?

oviano:
So my external allocator is this:

```
/*
* External Allocator
* (C) 2022 Oliver Collyer
*
*/

#include "botan.h"

#include "../../memory_allocator.h"

namespace Botan {

void* external_allocator::allocate(size_t num_elems, size_t elem_size)
   {
      return ovcore::allocate_memory(num_elems * elem_size);
   }

bool external_allocator::deallocate(void* p, [[maybe_unused]] size_t num_elems, [[maybe_unused]] size_t elem_size) noexcept
   {
      ovcore::free_memory(p);
      return true;
   }

external_allocator::external_allocator()
   {
   }

external_allocator::~external_allocator()
   {
   }

external_allocator& external_allocator::instance()
   {
   static external_allocator external;
   return external;
   }

}
```

...and these chain onto a bunch of functions that calling rpmalloc:

```
// ----------------------------------------------------------------------------------------------------

#include <atomic>
#include <iostream>
#include <thread>

#define ENABLE_STATISTICS   1
#define ENABLE_ASSERTS      1
#include "thirdparty/rpmalloc/rpmalloc.h"

#include "memory_allocator.h"

// ----------------------------------------------------------------------------------------------------

namespace ovcore {

// ----------------------------------------------------------------------------------------------------

static std::atomic<size_t> global_initialiser_count = 0;
static thread_local size_t thread_initialiser_count = 0;

// ----------------------------------------------------------------------------------------------------

struct INITIALISER
{
    INITIALISER()
    {
        if (thread_initialiser_count ++ == 0)
        {
            if (global_initialiser_count ++ == 0)
            {
                rpmalloc_initialize();
            }

            rpmalloc_thread_initialize();
        }
    }

    ~INITIALISER()
    {
        if (-- thread_initialiser_count == 0)
        {
            rpmalloc_thread_finalize(1);

            if (-- global_initialiser_count == 0)
            {
                rpmalloc_finalize();
            }
        }
    }

};

// ----------------------------------------------------------------------------------------------------

void initialise_memory_allocator()
{
    // initialise
    static thread_local INITIALISER initialisr;
}

// ----------------------------------------------------------------------------------------------------

void* allocate_memory(size_t size)
{
    // allocate memory
    return allocate_memory(size, (std::align_val_t)0);
}

// ----------------------------------------------------------------------------------------------------

void* allocate_memory(size_t size, std::align_val_t align)
{
    // initialise
    static thread_local INITIALISER initialisr;

    // align valid?
    if ((size_t)align > 0)
    {
        // allocate memory
        return rpaligned_alloc((size_t)align, size);
    }
    else
    {
        // allocate memory
        return rpmalloc(size);
    }
}

// ----------------------------------------------------------------------------------------------------

void free_memory(void* ptr)
{
    // initialise
    static thread_local INITIALISER initialisr;

    // free memory
    rpfree(ptr);
}

// ----------------------------------------------------------------------------------------------------

}
```

So for me it allows me to continue using my allocator (a wrapper around rpmalloc) in your excellent library too.

But for someone else it could do whatever they wanted it to, that's the point, it allows you to plug-in your own.

These PRs I've dumped today are just a collection of bits and bobs I've accumulated over the past year or two, and with 3.0 around the corner it seemed like a good time to offload them to you, in case they were of interest - no worries at all if it doesn't fit with your plans/design or whatever though.

oviano:
> > I think it needs disabling by default, how would I do that?
> 
> If you add `load_on request` to the `info.txt` file then it will only be built if `--enable-modules=external_allocator` is used.
> 
> OTOH I'm not sure if this gets you what you are looking for because then your external allocator is only called if you have a special one-off build, and otherwise it is ignored.
> 
> It seems like a better approach here might be something like
> 
> ```
> typedef std::function<void* (size_t, size_t)> alloc_fn;
> typedef std::function<void (void*, size_t, size_t)> free_fn;
> void set_memory_allocation_functions(alloc_fn alloc, free_fn free);
> ```
> 
>  ```
> // when allocating ...
> if(g_alloc_fn)
>    return g_alloc_fn(...);
> ```
> 
> I also want to reiterate again that this allocator interface is only ever called for `secure_vector` and in general Botan uses the standard C++ `new`/`delete` heap for most things.
> 
> Is the source of this allocator you are using public? If not can you describe what it is doing that we are missing from the current mlock pool that makes using your version desirable?

Ok, so it's not using it everywhere, so yes maybe your suggested approach would be better and encompass more allocations, I do take your point.

oviano:
Closing this one as I agree there are better solutions to achieve what I want.

PR #3431, opened 2023-03-26
Added secure_vector<T>::lock

oviano:
I found I sometimes needed to convert a regular vector into a secure_vector when switching between Botan/non-Botan code, so I added the counterpart lock() function.


randombit (approved):
Makes sense given that we have APIs that require using `secure_vector` in spots.

reneme:
Already merged, however: Using `std::span<>` would have been more versatile for that, no?

```C++
Botan::secure_vector<uint8_t> lock(std::span<uint8_t> buffer);
```

randombit:
Oh indeed I should have thought of this given all the recent span work. I'll open a PR

PR #3432, opened 2023-03-26
Adopt std::string_view in various APIs

randombit:


src/lib/asn1/asn1_oid.cpp
```diff
@@ -88,20 +88,24 @@ OID OID::from_string(const std::string& str)
    if(!raw.empty())
       return OID(std::move(raw));
 
-   throw Lookup_Error("No OID associated with name " + str);
+   std::ostringstream err;
+   err << "No OID associated with name " << str;
+   throw Lookup_Error(err.str());
```
reneme, line 93: Maybe not in this pull-request: But perhaps we want to start adopting `std::format` for things like that? (pending working compiler/library support)

src/lib/pubkey/pbes2/pbes2.cpp
```diff
@@ -199,24 +195,21 @@ secure_vector<uint8_t> derive_key(const std::string& passphrase,
...
-      throw Encoding_Error("PBE-PKCS5 v2.0: Invalid cipher spec " + cipher);
-
-   if(!known_pbes_cipher_mode(cipher_spec[1]))
-      throw Encoding_Error("PBE-PKCS5 v2.0: Don't know param format for " + cipher);
-
    auto enc = Cipher_Mode::create(cipher, Cipher_Dir::Encryption);
 
-   if(!enc)
-      throw Decoding_Error("PBE-PKCS5 cannot encrypt no cipher " + cipher);
+   const std::vector<std::string> cipher_spec = split_on(cipher, '/');
```
reneme, line 207: Technically, `split_on()` is the perfect candidate to return a vector of `std::string_view`. Maybe, at least use `auto` here, so that line will be agnostic to it.

```suggestion
   const auto cipher_spec = split_on(cipher, '/');
```

src/lib/pubkey/pk_algs.cpp
```diff
@@ -346,47 +325,82 @@ create_private_key(const std::string& alg_name,
...
-      const size_t rsa_bits = (params.empty() ? 3072 : to_u32bit(params));
-      return std::make_unique<RSA_PrivateKey>(rng, rsa_bits);
+      const size_t modulus_bits = [&]() -> size_t
+         {
+         if(params.empty())
+            return 3072;
+         return to_u32bit(params);
+         }();
+
+      return std::make_unique<RSA_PrivateKey>(rng, modulus_bits);
```
reneme, line 335: Out of curiosity: What was wrong with the ternary?

src/lib/pubkey/pk_algs.cpp
```diff
@@ -346,47 +325,82 @@ create_private_key(const std::string& alg_name,
...
-         throw Invalid_Argument("create_private_key bad McEliece parameters " + params);
-
-      size_t mce_n = to_u32bit(mce_param[0]);
-      size_t mce_t = to_u32bit(mce_param[1]);
-
-      return std::make_unique<McEliece_PrivateKey>(rng, mce_n, mce_t);
+      const auto [n, t] = [&]() -> std::pair<size_t, size_t>
+         {
+         if(params.empty())
+            return std::make_pair(2960, 57);
```
reneme, line 345: Nit: I believe this would do the trick with less code, because the expected return type is stated in the lambda declaration above. Caveat: I didn't try it. 😨

```suggestion
            return {2960, 57};
```

src/lib/pubkey/pk_algs.cpp
```diff
@@ -346,47 +325,82 @@ create_private_key(const std::string& alg_name,
...
+         if(mce_params.size() != 2)
+            {
+            std::ostringstream err;
+            err << "create_private_key: invalid McEliece parameters " << params;
+            throw Invalid_Argument(err.str());
+            }
+
+         const size_t mce_n = to_u32bit(mce_params[0]);
+         const size_t mce_t = to_u32bit(mce_params[1]);
+         return std::make_pair(mce_n, mce_t);
```
reneme, line 358: See above:

```suggestion
         return {mce_n, mce_t};
```

src/lib/utils/exceptn.cpp
```diff
@@ -68,104 +69,131 @@ std::string to_string(ErrorType type)
...
+* Should be replaced with std::format once that's available
+* on all our supported compilers.
+*/
+template<typename... T>
+std::string simple_format(T... args)
+   {
+   std::ostringstream oss;
+   ((oss << args), ...);
+   return oss.str();
+   }
```
reneme, line 86: Should we have that more prominently in (say) `stl_util.h`?

src/lib/utils/parsing.cpp
```diff
@@ -27,15 +27,16 @@ uint16_t to_uint16(const std::string& str)
    return static_cast<uint16_t>(x);
    }
 
-uint32_t to_u32bit(const std::string& str)
+uint32_t to_u32bit(std::string_view str_view)
    {
+   const std::string str(str_view);
```
reneme, line 32: One could consider using [`std::from_chars()`](https://en.cppreference.com/w/cpp/utility/from_chars) to avoid the copy. Though, the API seems a bit low-level for my taste. Still worth considering.

reneme (approved):
I like! Just a few nits.

src/lib/pubkey/pbes2/pbes2.cpp
randombit: I'm a bit nervous about returning string_view from anywhere given C++ doesn't really offer the lifetime tracking that Rust does, so you can end up with something like

```
auto uhoh = split_on(some_temporary(), '/')
```

The `auto` is fine though

src/lib/pubkey/pk_algs.cpp
randombit: You know I think here actually the ternary worked ok and I just autopiloted the conversion after I did the other cases. I'll revert this bit.

src/lib/utils/exceptn.cpp
randombit: I'm hoping to write a (very simple) variant of `std::format` that just handles `{}` with no flags, etc and promote that to `stl_util.h`, then we have a very easy conversion path when `std::format` is available. We could even start using it today on MSVC, AIUI:

```
#if feature_check_for_std_format
  using simple_format = std::format;
#else
  std::string simple_format(...) {}
#endif
```

but I guess in the meantime we could promote this version and start using it

src/lib/asn1/asn1_oid.cpp
randombit: Not until GCC 13, which isn't even released yet :disappointed:

src/lib/utils/exceptn.cpp
randombit: I started on this but it actually becomes quite wide ranging as we use `std::ostringstream` to format exception messages and the like in a number of spots. I'll open a second followup PR here.

src/lib/pubkey/pk_algs.cpp
randombit: Reverted

src/lib/pubkey/pk_algs.cpp
randombit: :+1:

src/lib/utils/parsing.cpp
randombit: I just realized this whole function is actually exceptionally inefficient (not that it really matters) since we scan the string twice. I'll open another PR here.

src/lib/utils/exceptn.cpp
randombit: #3440

PR #3436, opened 2023-03-26
Small cleanup for up how CommonCrypto hashes are created

randombit:


randombit:
Apparently despite Apple's own docs claiming there is a `CC_SHA224_CTX` ... there isn't. Excellent work guys.

Issue #3438, opened 2023-03-27
Session_Manager_SQL::store immediately deletes record when it calls prune_session_cache() if max sessions is zero


oviano:
Slightly obscure one, but according to the documentation:

```
      /**
      * @param db A connection to the database to use
               The table names botan_tls_sessions and
               botan_tls_sessions_metadata will be used
      * @param passphrase used to encrypt the session data
      * @param rng a random number generator
      * @param max_sessions a hint on the maximum number of sessions
      *        to keep in memory at any one time. (If zero, don't cap)
      */
```

However, this code here in prune_session_cache() will delete them all if m_max_sessions is zero:

```
   auto remove_oldest = m_db->new_statement("DELETE FROM tls_sessions WHERE session_id NOT IN "
                                            "(SELECT session_id FROM tls_sessions ORDER BY session_start DESC LIMIT ?1)");
   remove_oldest->bind(1, m_max_sessions);
   remove_oldest->spin();
```

I guess it just doesn't need to do anything inside prune (or not call it) if m_max_sessions == 0.

PR #3439, opened 2023-03-27
When m_max_sessions is zero, don't prune the session cache.

oviano:
Fixes https://github.com/randombit/botan/issues/3438.

### DIFF
--- a/src/lib/tls/sessions_sql/tls_session_manager_sql.cpp
+++ b/src/lib/tls/sessions_sql/tls_session_manager_sql.cpp
@@ -296,7 +296,7 @@ void Session_Manager_SQL::prune_session_cache()
    {
    // internal API: assuming that the lock is held already
 
-   if(!m_max_sessions)
+   if(m_max_sessions == 0)
       return;
 
    auto remove_oldest = m_db->new_statement("DELETE FROM tls_sessions WHERE session_id NOT IN "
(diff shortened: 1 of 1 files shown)


randombit (commented):
Seems fine to me. @reneme?

src/lib/tls/sessions_sql/tls_session_manager_sql.cpp
```diff
@@ -296,6 +296,9 @@ void Session_Manager_SQL::prune_session_cache()
    {
    // internal API: assuming that the lock is held already
 
+   if(!m_max_sessions)
+      return;
```
reneme, line 300: For clarity, maybe:

```suggestion
   if(m_max_sessions == 0)
      return;
```

reneme (changes requested):
It would be great to add a test in `test_tls_session_manager.cpp`. More specifically add a `CHECK` section in the `test_session_manager_sqlite()` function as a regression test.

src/lib/tls/sessions_sql/tls_session_manager_sql.cpp
oviano: I'm not familiar with how the tests work/what exactly you want it to test, so I'll have to leave that to somebody else.

reneme:
I'll add the test in a follow-up. Thanks @oviano.

reneme (approved):

PR #3440, opened 2023-03-27
Add a simple formatting helper

randombit:
This can serve as a standin for `std::format` until either a more `std::format` like helper can be devised or (longer term) we can switch to a compiler provided `std::format`.

### DIFF
--- a/src/lib/pubkey/pbes2/pbes2.cpp
+++ b/src/lib/pubkey/pbes2/pbes2.cpp
@@ -12,6 +12,7 @@
 #include <botan/der_enc.h>
 #include <botan/ber_dec.h>
 #include <botan/internal/parsing.h>
+#include <botan/internal/fmt.h>
 #include <botan/asn1_obj.h>
 #include <botan/rng.h>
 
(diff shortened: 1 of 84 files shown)


randombit:
I just realized it's very easy to make this `std::format` compatible using `{}` but sadly I can't push it because GH is having problems.

reneme:
FWIW: Also had trouble with GitHub half an hour ago. Though, at least from Berlin, it works again.

src/lib/utils/fmt.h
```diff
@@ -0,0 +1,68 @@
...
+         return do_fmt(oss, format.substr(i + 2), rest...);
+         }
+      else
+         {
+         oss << format[i];
+         }
+
+      i += 1;
+      }
+   }
```
reneme, line 43: There's an off-by-one here, no?

Consider `format = "I will make the sanitizer unhappy {"`. That will cause an out-of bounds read when looking for the closing brace.

```suggestion
   while(i < format.size() - 1)
      {
      if(format[i] == '{' && format[i + 1] == '}')
         {
         oss << val;
         return do_fmt(oss, format.substr(i + 2), rest...);
         }
      else
         {
         oss << format[i];
         }

      i += 1;
      }
   }
```

src/lib/pubkey/pbes2/pbes2.cpp
```diff
@@ -85,15 +87,15 @@ secure_vector<uint8_t> derive_key(std::string_view passphrase,
...
+      pwdhash->hash(derived_key, passphrase, salt);
 
       return derived_key;
       }
    else
-      throw Decoding_Error("PBE-PKCS5 v2.0: Unknown KDF algorithm " +
-                           kdf_algo.oid().to_string());
+      {
+      throw Decoding_Error(fmt("PBE-PKCS5 v2.0: Unknown KDF algorithm {}",
+                                kdf_algo.oid().to_string()));
```
reneme, line 97: As a side note: When we eventually move to `std::format` these `to_string()` gymnastics should be handled centrally with our custom types IMO. Didn't look into it in detail, but I'd be surprised if one could not define this generically for user-defined types in `std::format`.

For now, I think, this is totally sufficient.

reneme (approved):
Nice! LGTM, modulo the potential out-of-bounds read and maybe a few basic unit tests.

src/lib/utils/fmt.h
randombit: Fixed now, also added tests

src/lib/pubkey/pbes2/pbes2.cpp
randombit: I haven't looked into in detail either (seems somewhat academic without a compiler to test anything with) but it appears [std::formatter](https://en.cppreference.com/w/cpp/utility/format/formatter) will let us add custom flags support and other fancy stuff.

For the short term I added an operator<< for OID, which is sufficient to avoid the explicit `to_string` calls, since our `fmt` will happily eat anything it can put into an `ostream`

randombit:
Fixed the off by one, added some unit tests and also expanded usage somewhat to other places where we were doing string concatenations, especially `"Foo(" + x + ")"`. Adopting a parenthesized notation for algorithm names should probably go into [`mistakes.rst`](https://github.com/randombit/botan/blob/master/doc/dev_ref/mistakes.rst) 

I was a little worried adopting this widely was going to blow up the binary size but happily it seems like (at least on my machine/compiler) the resulting binary actually got a bit smaller, vs `master`.

PR #3442, opened 2023-03-27
Use span for the argument to lock

randombit:


reneme (approved):

randombit:
Oh that is unfortunate. Given that I think I'll just revert this change.

PR #3449, opened 2023-03-28
Removed redundant matrix creation in dilithium keygen

FAlbertDev:
The `matrix` created by `rho` is not used in this function. It is (newly) created and used in `calculate_t0_and_t1`.
I also see no reason for this piece of code from a side-channel perspective, since `rho` is public.


lieser:
I'm a little surprised that we didn't get a warning here about an unused variable.

PR #3451, opened 2023-03-28
FIX: Session_Manager::find_all() is inefficient

reneme:
Now that we're talking about the efficiency of the `Session_Manager` class hierarchy: This is likely something that should go in before the release candidate. It replaces `Session_Manager::find_all()` with `::find_some()`.

Technically, nothing would have forced application-defined session managers to actually always return "all" found sessions, but the method name very much implied that. Instead, it is now called `find_some()` and gets a "max sessions hint" that implementations can (and should) follow.

Additionally, the `Session_Manager` base class now has some mechanics to call `find_some()` more than once if the returned `Session` objects don't pass its filters -- currently mainly a session expiry check. This algorithm might see some evolution, but I think the `find_some()` interface is more versatile and better suited.


reneme:
Rebased after merging #3450.

PR #3453, opened 2023-03-29
Added botan_system_rng_get convenience function

oviano:
I am writing a botan CRYSPR for SRT and I found there was no obvious hook to create and destroy a random number generator via the callbacks SRT provides, so this function, which mirrors what is already available in the C++ API makes things a lot easier.

Needs checking though - I'm not too sure about the meaning/purpose of the ffi_guard_thunk.

### DIFF
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -264,6 +264,14 @@ BOTAN_PUBLIC_API(3,0) int botan_rng_init_custom(botan_rng_t* rng_out, const char
 BOTAN_PUBLIC_API(2,0) int botan_rng_get(botan_rng_t rng, uint8_t* out, size_t out_len);
 
 /**
+* Get random bytes from system random number generator
+* @param out output buffer of size out_len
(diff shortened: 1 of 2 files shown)


src/lib/ffi/ffi.h
```diff
@@ -263,6 +263,14 @@ BOTAN_PUBLIC_API(3,0) int botan_rng_init_custom(botan_rng_t* rng_out, const char
 */
 BOTAN_PUBLIC_API(2,0) int botan_rng_get(botan_rng_t rng, uint8_t* out, size_t out_len);
 
+/**
+* Get random bytes from system random number generator
+* @param out output buffer of size out_len
+* @param out_len number of requested bytes
+* @return 0 on success, negative on failure
+*/
+BOTAN_PUBLIC_API(2,0) int botan_system_rng_get(uint8_t* out, size_t out_len);
```
randombit, line 272: Just a note, this should be `3,0` since the first release this function is available in will be 3.0

PR #3455, opened 2023-03-29
cli: tls_client: support custom DTLS version

nunojpg:


randombit:
I don't think this works as you would expect since we don't have DTLS 1.3 support yet

nunojpg:
This commit fixes the case where you specify tls_version == "1.2" for UDP, otherwise it throws.

If you specify 1.3 it throws either way since DTLS 1.3 is not supported.

Anyway it fixes a case, and I don't see why DTLS 1.3 should be handled as a special case as the result is a throw anyway.

PR #3456, opened 2023-03-31
Various clang-tidy fixes

randombit:


reneme (approved):
LGTM.

One question, though: What's your rationale behind the strong encapsulation in internal structs? Like `TLSPlaintext_Header` or the `Client_Hello`/`Server_Hello` internal structs.

IMO, there won't ever be any loose-coupling or code reuse outside their narrow scope (compilation unit). Therefore, I usually see them as a pure implementation detail and value developer convenience over encapsulation. But maybe I'm missing something?

randombit:
> Therefore, I usually see them as a pure implementation detail and value developer convenience over encapsulation. But maybe I'm missing something?

It allows a reader to more easily understand the situations in which the value might be modified. As an extreme case `TLSPlaintext_Header` does not have any mutations required, only getters. So it's very clear that, after deserialization, the values are not modified by anywhere else in the code. And in the server hello, only the `Extensions` are ever modified, again everything else is constant after the constructor finishes.

reneme:
> It allows a reader to more easily understand the situations in which the value might be modified.

Okay, thanks. I agree that this is worth the extra methods.

Issue #3459, opened 2023-03-31
Flaky PKCS11 RSA test, possible PKCS11 RSA bug


randombit:
Finally managed to narrow down and reproduce the occasional failure in the PKCS11 RSA tests that have been seen up until now only in CI.

The issue is in `set_use_software_padding` configuration. Occasionally (~ once per 200 tests) the decryption fails, with either PKCS1 or OAEP padding. In the version that uses padding generated by the token, the test always (or at least after 1000s of tries) passes.

```
diff --git a/src/tests/test_pkcs11_high_level.cpp b/src/tests/test_pkcs11_high_level.cpp
index 542ed71e8..9ea7ce647 100644
--- a/src/tests/test_pkcs11_high_level.cpp
+++ b/src/tests/test_pkcs11_high_level.cpp
@@ -861,10 +861,13 @@ Test::Result test_rsa_encrypt_decrypt()

    plaintext = { 0x00, 0x01, 0x02, 0x03, 0x04, 0x00 };
    encrypt_and_decrypt(plaintext, "EME-PKCS1-v1_5", false);
-   encrypt_and_decrypt(plaintext, "EME-PKCS1-v1_5", true);
+   for(size_t i = 0; i != 1000; ++i)
+      encrypt_and_decrypt(plaintext, "EME-PKCS1-v1_5", true);

    encrypt_and_decrypt(plaintext, "OAEP(SHA-1)", false);
-   encrypt_and_decrypt(plaintext, "OAEP(SHA-1)", true);
+   for(size_t i = 0; i != 1000; ++i)
+      encrypt_and_decrypt(plaintext, "OAEP(SHA-1)", true);

    keypair.first.destroy();
    keypair.second.destroy();
```

cc @reneme @securitykernel

randombit:
Disabling the blinding steps in `PKCS11_RSA_Decryption_Operation` seems to work around the problem (ie padding in userspace works fine, but doing padding plus blinding causes failures)

randombit:
SoftHSMv2 debug log

`OSSLRSA.cpp(1301): Invalid amount of input data supplied for RSA decryption`

randombit:
Fixed by #3460

PR #3460, opened 2023-03-31
When performing PKCS11 RSA padding in userspace, pad to modulus size

randombit:
At least SoftHSM requires this, and missing it caused occasional CI failures.

GH #3459


reneme (approved):
Nice! Thanks for looking into that. I don't see how this change could case problems for us. But it might be worth looking at for @weberph2 or @neusdan.

Issue #3377, opened 2023-03-15
Use Breathe for API Reference documentation


reneme:
Currently, the documentation in `docs/api_ref` duplicates quite a bit of the API details already documented in Doxygen. This adds a maintenance burden and leads to outdated documentation.

Instead, I propose to use [Breathe](https://breathe.readthedocs.io/en/latest/index.html) for our Sphinx documentation. This allows pulling API documentation from Doxygen into Sphinx documentation.

### Example

Say, I want to describe the mandatory callbacks to be implemented by an application using TLS.

First I would use a [doxygen member group](https://www.doxygen.nl/manual/grouping.html#memgroup) to define the mandatory members in the code.

```C++
namespace Botan::TLS {
class Callbacks {
public:
  /**
   * @name Mandatory
   * Those callbacks must be overridden by all applications wishing to use TLS.
   */
  ///@{
  void tls_emit_data();
  void tls_record_received();
  // ...
  /// @}
};
}
```

Then, [instead of recreating all the information about the callback methods](https://github.com/randombit/botan/blob/b67c32643c1bc2e84896cd43967e087b44149346/doc/api_ref/tls.rst), one would simply add the following in the reStructuredText documentation for Sphinx:

```rST
.. doxygenclass:: Botan::TLS::Callbacks
    :members:
    :membergroups: Mandatory
```

And the result would look like that. Further tuning is possible, of course:

![image](https://user-images.githubusercontent.com/1562139/225262078-128683a5-5e65-4c1c-8341-196f7043954c.png)

randombit:
Sounds good to me

randombit:
It does seem breathe is available in most distros

https://repology.org/project/python:breathe/versions
and
https://repology.org/project/breathe/versions

But the question arises what can/should we do if breathe is not available. Can we proceed with building the docs at that point, in some degraded fashion? Or is there no chance of proceeding?

reneme:
We can detect that breathe is not installed in the system (via `except ImportError` in Sphinx's `conf.py`) and simply not register it as an extension in Sphinx. This will result in warnings during build:

```
doc/api_ref/tls.rst:39: WARNING: Unknown directive type "doxygenclass".

.. doxygenclass:: Botan::TLS::Callbacks
    :members:
    :membergroups: Mandatory
```

and simply leave out the Doxygen information.

Probably, we could even mock breathe's directives in this case and render some "API documentation not available"-warning into the documentation in lieu of the actual breathe output.

[...]

PR #3464, opened 2023-04-03
Doc: Set up Breathe

reneme:
This adds [breathe](https://github.com/breathe-doc/breathe) as a bridge between Doxygen and Sphinx. It should allow us to replace _a lot_ of duplicated information simply with pointers to the information from Doxygen comments.

Also, I added a (super simplistic) stand-in if breathe is disabled or not available during `./configure.py`. Basically, it just writes something along these lines into the documentation:

```
[Missing Breathe Content: Botan::TLS::Callbacks]
```

No adaptions to the reStructuredText documentation have been done, yet. That will come in later PRs. If you want to test the integration locally, just go to `hash.rst` and replace the entire `.. cpp:class:: HashFunction` block with:

```
.. doxygenclass:: Botan::HashFunction
   :members:
```

... more fine tuning might be needed to further filter the displayed information, but it should show that it works.

### DIFF
--- a/doc/migration_guide.rst
+++ b/doc/migration_guide.rst
@@ -12,10 +12,7 @@ Documentation
 
 Most of the API references are now powered by `Breathe
 <https://github.com/breathe-doc/breathe>`_. Without Breathe the documentation
-will lack almost all API details and will be much less helpful. Additionally,
-the Sphinx extension `togglebutton
-<https://pypi.org/project/sphinx-togglebutton/>`_ must be installed to
-successfully build the documentation.
+will lack almost all API details and will be much less helpful.
 
 Headers
 --------
--- a/doc/migration_guide.rst
+++ b/doc/migration_guide.rst
@@ -7,6 +7,16 @@ This guide attempts to be, but is not, complete. If you run into a problem while
 converting code that does not seem to be described here, please open an issue on
 Github.
 
+Documentation
+-------------
+
+Most of the API references are now powered by `Breathe
+<https://github.com/breathe-doc/breathe>`_. Without Breathe the documentation
+will lack almost all API details and will be much less helpful. Additionally,
+the Sphinx extension `togglebutton
+<https://pypi.org/project/sphinx-togglebutton/>`_ must be installed to
+successfully build the documentation.
+
 Headers
 --------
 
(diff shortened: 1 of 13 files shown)


reneme:
Note that this now also installs `sphinx-togglebutton` (rationale: #3470)

randombit (commented):
Since lacking Breathe results in somewhat substandard documentation, the usage should be noted in the manual, changelog, and also the migration guide. I expect even with all that the initial packaging for distributions will miss the new dependency.

doc/migration_guide.rst
```diff
@@ -7,6 +7,16 @@ This guide attempts to be, but is not, complete. If you run into a problem while
...
 
+Documentation
+-------------
+
+Most of the API references are now powered by `Breathe
+<https://github.com/breathe-doc/breathe>`_. Without Breathe the documentation
+will lack almost all API details and will be much less helpful. Additionally,
+the Sphinx extension `togglebutton
+<https://pypi.org/project/sphinx-togglebutton/>`_ must be installed to
+successfully build the documentation.
```
randombit, line 18: Woah - is this really correct? We *must* have this extension, to build the docs at all? I don't think that can work.

Per https://repology.org/project/python:sphinx-togglebutton/versions effectively zero Linux distributions include this package, which would mean that everyone installing the Linux packages from their distro would get no (or broken) documentation.

randombit:
I think there are some major unresolved issues here and since our adopting vs not adopting this is not a SemVer issue, I propose postponing to after 3.0 is released.

doc/migration_guide.rst
reneme: Mhm, indeed. Its likely possible to ship this (or a subset of this) extension so that we don't need to pull it out of `pip` or similar.

reneme:
> I think there are some major unresolved issues here and since our adopting vs not adopting this is not a SemVer issue, I propose postponing to after 3.0 is released.

I've jotted down my thoughts on this in https://github.com/randombit/botan/issues/3377#issuecomment-1505155573. Let's move further discussion there.

TL;DR: In its current shape, Breathe will unfortunately not work for us, I'm afraid.

reneme:
Rebased to latest master, added `sphinx-togglebutton` to the repo be independent of distro's repos and disabled `XML_PROGRAMLISTING`.

randombit:
With #4185 merging the doc  comments, and Breathe apparently abandoned/on hold upstream I think we should close this and related PRs

PR #3466, opened 2023-04-04
FIX: build with TLS 1.3 disabled

reneme:
Building a shared library on Windows with `--module-policy=bsi --enable-modules=tls` failed due to missing symbols.


src/lib/tls/tls_messages.h
```diff
@@ -452,7 +452,6 @@ class BOTAN_UNSTABLE_API Hello_Retry_Request final : public Server_Hello_13
       Handshake_Type wire_type() const override { return Handshake_Type::ServerHello; }
    };
 
-#endif // BOTAN_HAS_TLS_13
```
lieser, line 455: Multiple empty lines intended?

PR #3468, opened 2023-04-04
[TLS 1.3] Build Module Configuration

reneme:
This marks the `tls` module "Internal" (users cannot explicitly enable it via `--enable-module=` anymore). Therefore, users must deliberately select either `tls12` or `tls12,tls13`. Note that exclusive `tls13` is NYI. Also adapt the migration guide accordingly.

Additionally, this mentions that Windows now suffixes the DLL name with the library's major version.


mouse07410:
Do I need to guess what you mean by NYI?

reneme:
Not Yet Implemented. Currently its not possible to compile botan with TLS 1.3 support but without TLS 1.2 support. You'd need to disable TLS 1.2 in the policy at runtime for the time being.

PR #3470, opened 2023-04-04
Doc: Use Breathe for BigInt and Block_Cipher chapters

reneme:
### Pull Request Dependencies

* #3464 

### Description

This replaces the API description in `bigint.rst` and `block_cipher.rst` with Breathe directives. I stopped after a few chapters to collect some feedback.

The manual descriptions were actually not just copy-and-pasted from the Doxygen comments. I.e. I tried to conserve information by migrating it into the header files where it seemed necessary.

#### Fold away the API Overview

TL;DR: I suggest adding [`sphinx-togglebutton`](https://sphinx-togglebutton.readthedocs.io/en/latest/) (or some similar extension) to make the documentation less cluttered by collapsing the Breathe API overview by default.

image

---

image

It is unquestionably more maintainable to pull the API documentation from Doxygen via Breathe. Though, it produces more detail than strictly necessary to get a feeling for the API.

For instance, it renders all subclasses of `BlockCipher`, producing a lot of clutter (other people [have pointed that out](https://github.com/breathe-doc/breathe/issues/840)).

image

If there are several overloads of a single method, it always renders all of them:

image

Personally, I find the inlined API overview is very helpful and probably saves the user from digging through Doxygen most of the time. But being exposed to all this detail on every reference page seems offputting.

#### Undocumented Overloads

Turns out that undocumented overloads are not added to the Breathe output. For instance, this would not display the c-style overload in the Breathe output:

```C++
/**
* Encrypt one or more blocks in place. The @p block will hold the result
* when the function returns.
* @param block the plaintext block to be encrypted (multiple of
*              block_size())
*/
void encrypt(std::span<uint8_t> block) const
   { /* ... */ }
void encrypt(uint8_t block[]) const
   { /* ... */ }
```

I.e. by not adding a Doxygen comment to method overloads we might nudge users towards the more modern overloads.

In fact, I've been doing that in other `std::span`-refactorings already. E.g. in the `Random_Number_Generator`:

https://github.com/randombit/botan/blob/0a4ab5d796e823cf7e7807757ec842b932869694/src/lib/rng/rng.h#L43-L56

What do you think, @randombit @lieser?

### DIFF
--- a/doc/api_ref/block_cipher.rst
+++ b/doc/api_ref/block_cipher.rst
@@ -13,118 +13,6 @@ operations such as authenticated encryption.
    modes or MACs), or in the very rare situation where ECB is required,
    eg for compatibility with an existing system.
 
-.. cpp:class:: BlockCipher
-
-  .. cpp:function:: static std::unique_ptr<BlockCipher> create(const std::string& algo_spec, \
-                                                               const std::string& provider = "")
-
-      Create a new block cipher object, or else return null.
-
-  .. cpp:function:: static std::unique_ptr<BlockCipher> create_or_throw(const std::string& algo_spec, \
-                                                                        const std::string& provider = "")
-
-      Like ``create``, except instead of returning null an exception is thrown
-      if the cipher is not known.
-
-  .. cpp:function:: void set_key(const uint8_t* key, size_t length)
-
-      This sets the key to the value specified. Most algorithms only accept keys
-      of certain lengths. If you attempt to call ``set_key`` with a key length
-      that is not supported, the exception ``Invalid_Key_Length`` will be
-      thrown.
-
-      In all cases, ``set_key`` must be called on an object before any data
-      processing (encryption, decryption, etc) is done by that object. If this
-      is not done, an exception will be thrown.
-      thrown.
-
-  .. cpp:function:: bool valid_keylength(size_t length) const
-
-     This function returns true if and only if *length* is a valid keylength for
-     this algorithm.
-
-  .. cpp:function:: size_t minimum_keylength() const
-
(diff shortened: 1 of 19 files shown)


doc/api_ref/block_cipher.rst
```diff
@@ -134,6 +22,14 @@ block of plaintext with AES-256 using two different keys.
 .. literalinclude:: /../src/examples/aes.cpp
    :language: cpp
 
+API Overview
+------------
+
+.. container:: toggle
+
+   .. doxygenclass:: Botan::BlockCipher
+      :members: set_key,minimum_keylength,maximum_keylength,encrypt,decrypt,encrypt_n,decrypt_n
```
lieser, line 31: Before it also documented the static `create()` and `create_or_throw()`. I would say we probably want to continue mentioning them.

lieser:
I also don't like that for `BlockCipher` all the subclasses are shown. Probably a little confusing for users to so preeminently mention all this different `Block_Cipher_Fixed_Params`.

But I would say the advantage of ensuring that the API documentation stays up to date is worth it.

randombit:
I also find that very cluttered and unpleasant, but I suppose worst case is that it annoys me enough that I go off and spend a year writing a Doxygen+Sphinx+Breathe replacement tool.

randombit:
(My frustrations with our doc stack go beyond format nits - Doxygen has a lot of parsing bugs, and Sphinx, while pleasant to use, is absurdly slow)

doc/api_ref/block_cipher.rst
reneme: Good point. Done.

reneme:

PR #3478, opened 2023-04-05
Doc: Breathe for version info

reneme:
### Pull Request Dependencies

* #3464


reneme:

PR #3480, opened 2023-04-05
Use std::string_view and fmt more extensively

randombit:


src/lib/pubkey/xmss/xmss_parameters.h
```diff
@@ -176,12 +176,12 @@ class BOTAN_PUBLIC_API(2,0) XMSS_Parameters
...
        * @return XMSS registry name for the chosen parameter set.
        **/
-      const std::string& name() const
+      const std::string name() const
          {
          return m_name;
          }
 
-      const std::string& hash_function_name() const
+      const std::string hash_function_name() const
```
reneme, line 184: Sorry for pulling out this old PR. Though: I think these should have stayed `const std::string&`, no?

Issue #3469, opened 2023-04-04
Botan in JVM + SoftHSM


antoinelochet:
Hello,
We use SoftHSM for testing in our Java application.
We would like to have Botan to test our application through SoftHSM becuase of the PQC support which do not exist (and will probably never exist) in OpenSSL 1.1
We encounter a SIGSEGV when SoftHSM is stopping with the JVM and is destructing the Botan instance:

```
Current thread (0x00007fe48811a570):  VMThread "VM Thread" [stack: 0x00007fe48c6b0000,0x00007fe48c7b0000] [id=5978]

Stack: [0x00007fe48c6b0000,0x00007fe48c7b0000],  sp=0x00007fe48c7ae448,  free space=1017k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [libc.so.6+0x16173c]
C  [libbotan-2.so.19+0x7e0e21]  Botan::deallocate_memory(void*, unsigned long, unsigned long)+0x36
C  [libsofthsm-develop-with-kdf-and-date-encryption.so+0x13b428]  Botan::secure_allocator<unsigned long>::deallocate(unsigned long*, unsigned long)+0x2c
C  [libsofthsm-develop-with-kdf-and-date-encryption.so+0x13b386]  std::allocator_traits<Botan::secure_allocator<unsigned long> >::deallocate(Botan::secure_allocator<unsigned long>&, unsigned long*, unsigned long)+0x2b
C  [libsofthsm-develop-with-kdf-and-date-encryption.so+0x13b186]  std::_Vector_base<unsigned long, Botan::secure_allocator<unsigned long> >::_M_deallocate(unsigned long*, unsigned long)+0x32
C  [libsofthsm-develop-with-kdf-and-date-encryption.so+0x13b03a]  std::_Vector_base<unsigned long, Botan::secure_allocator<unsigned long> >::~_Vector_base()+0x3e
C  [libsofthsm-develop-with-kdf-and-date-encryption.so+0x13b07f]  std::vector<unsigned long, Botan::secure_allocator<unsigned long> >::~vector()+0x41
C  [libsofthsm-develop-with-kdf-and-date-encryption.so+0x13acfe]  Botan::BigInt::Data::~Data()+0x18
C  [libsofthsm-develop-with-kdf-and-date-encryption.so+0x13ad26]  Botan::BigInt::~BigInt()+0x24
[...]

PR #3481, opened 2023-04-05
Use init_priority attribute on the pool allocator

randombit:
This will probably address many of the static init problems, at least for GCC/Clang users.

See GH #761 and #3469 among many others


antoinelochet:
I have tested this patch on two different environnements without a hitch.
Thanks again.

PR #3482, opened 2023-04-06
Update of BSI policy for BSI guidelines 2023

FAlbertDev:
I've checked Botan's BSI policy against the newest ['BSI – Technical Guideline' document](https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TG02102/BSI-TR-02102-1.pdf?__blob=publicationFile&v=6) to see if the current policy is still up to date. This PR contains some adaptions. Since I'm still quite new to Botan and are not familiar with all algorithms, I hope that you can verify and double check my changes. In the following I give the reasoning for all changes I've done:

1. Removed the XTS mode, since it does not appear in [Table 2.2 of the Guideline](https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TG02102/BSI-TR-02102-1.pdf?__blob=publicationFile&v=6#page=25).
2. Added the PQC signature scheme Dilithium (Kyber/XMSS were already added). The BSI states [the following](https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TG02102/BSI-TR-02102-1.pdf?__blob=publicationFile&v=6#page=39):
> The international standardisation of PQC mechanisms, in particular by the American National Institute of Standards and Technology (NIST), has meanwhile started and first mechanisms have been selected for standardisation. Corresponding standards are expected in the coming years. Introducing current, non-standardised mechanisms in new cryptographic systems is therefore always associated with the risk of creating systems that are incompatible with standards that are foreseeable for the near future. However, in applications that are intended to guarantee the confidentiality of information with a high value and a long-term need for protection, these problems weigh less heavily in the BSI’s view than the possibility of future attacks. In general, it is recommended to put great stress on cryptoagility in the design of new systems;

For me it sounds like PQC algorithms should be used with caution, but are basically allowed.
3. Added `argon2_avx2` for password based KDFs. I see no reason why not (argon2_ssse3 is already enabled).
4. The BSI states [the following](https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TG02102/BSI-TR-02102-1.pdf?__blob=publicationFile&v=6#page=72):
> If the use of a a cryptographic hardware token for password-based key derivation is not possible, the hash function Argon2id should be used. The security parameters of Argon2id and the requirements for the passwords depend on the application scenario and should be discussed with an expert.

For me it sounds like only Argon2 is recommended for PBKDFs. Therefore, i would forbid all other PBKDFs, i.e, ``pbkdf2``, `bcrypt_pbkdf`, `pgp_s2k`, and `scrypt`. Also the password hashes `passhash9` and `bcrypt`(already prohibited).
5. Added `getentropy` as entropy collection. I don't know much about there different entropy collection systems, but it seems that BSD's getentropy has high security claims. Also the BSI does not specify specific algorithms as far as I know.
6. Reenabled `hkdf`. I was told it is required for TLS 1.3.
7. Disabled the public key padding modules `eme_raw`. Should be sensible.
8. Disabled the hash module ``raw_hash``. At first glance I could not find other modules using `raw_hash`. Otherwise, it should definitely be disabled.

### DIFF
--- a/src/build-data/policy/bsi.txt
+++ b/src/build-data/policy/bsi.txt
@@ -38,6 +38,8 @@ iso9796
 # pubkey
 dlies
 dh
+#dilithium // not (yet) recommended
+#dilithium_aes // not (yet) recommended
 rsa
 dsa
 ecdsa
@@ -45,8 +47,8 @@ ecgdsa
 ecies
 eckcdsa
 ecdh
-kyber
-kyber_90s
+#kyber // not (yet) recommended
+#kyber_90s // not (yet) recommended
 xmss
 
 # rng
@@ -73,10 +75,12 @@ sha2_64_bmi2
 sha3_bmi2
 
 # entropy sources
+getentropy
 rdseed
 win32_stats
 
 # pbkdf
+argon2_avx2
 argon2_ssse3
 
 # rng
@@ -134,7 +138,7 @@ salsa20
 #shake_cipher # not recommended, but needed for kyber
 
 # kdf
-hkdf
+#hkdf // needed for tls 1.3
 kdf1
 kdf2
 prf_x942
@@ -153,6 +157,7 @@ sm2
 # pk_pad
 #eme_pkcs1 // needed for tls
 #emsa_pkcs1 // needed for tls
+#eme_raw // allows custom paddings
 emsa_raw
 emsa_x931
 


src/build-data/policy/bsi.txt
```diff
@@ -123,6 +127,7 @@ eax
 ocb
 siv
 cfb
+xts
```
reneme, line 130: Should not be prohibited, just not enabled by default.

```suggestion
```

src/build-data/policy/bsi.txt
```diff
@@ -134,12 +139,18 @@ salsa20
...
 kdf2
 prf_x942
 sp800_56a
 
+# pbkdf
+bcrypt_pbkdf
+pbkdf2
+pgp_s2k
+scrypt
+
```
reneme, line 153: As long as BSI doesn't explicitly discourage their usage, we should not prohibit them. Just not enable by default. E.g. `pgp_s2k`: I could imagine that the PGP-related BSI project might even depend on it.

src/build-data/policy/bsi.txt
```diff
@@ -153,8 +164,10 @@ sm2
 # pk_pad
 #eme_pkcs1 // needed for tls
 #emsa_pkcs1 // needed for tls
+eme_raw
 emsa_raw
 emsa_x931
+raw_hash
```
reneme, line 170: That's a module to implement certain special cases using Botan. Shouldn't be enabled by default but also not prohibited.

src/build-data/policy/bsi.txt
```diff
@@ -153,8 +164,10 @@ sm2
 # pk_pad
 #eme_pkcs1 // needed for tls
 #emsa_pkcs1 // needed for tls
+eme_raw
```
reneme, line 167: Don't prohibit, just not part of the default module set, please.

src/build-data/policy/bsi.txt
```diff
@@ -182,5 +195,6 @@ x919_mac
 
 # passhash
 bcrypt
+passhash9
```
reneme, line 198: Do they explicitly prohibit PBKDF2? If no, let's just leave it to the user.

reneme (changes requested):
For Botan 3.0 we should just update the policy so that new things can be used. (e.g. by removing HKDF from the prohibited-list in favor of TLS 1.3). Additional prohibitions should be used scarcely. We might re-iterate that at a later stage.

src/build-data/policy/bsi.txt
lieser: This new module is from my understanding now used in e.g. ECDSA there before `emsa_raw` was used (as they required EMSA1(hash) before instead of the hash directly).

Don't know why we currently forbid `emsa_raw`, but we may want to keep it consistent an either forbid both or none.

src/build-data/policy/bsi.txt
FAlbertDev: I would suggest that we try to stay consistent. For example, why prohibiting the EAX mode, if there is no security flaw (as far as I know). IMO we have three possibilities:
1) We continue prohibiting XTS and all other modes not explicitly recommended
2) We only prohibit modes with discovered security flaws (however,this requires literature research, which is the reason guidelines like the one of BSI exist)
3) We activate the recommended modes but prohibit none.

I think we should get some feedback on these three approaches and apply them to other categories as well. Until then I think its the best approach to stay consistent with approach 1), i.e., prohibiting `xts`. Or is `xts` some required dependency?

src/build-data/policy/bsi.txt
FAlbertDev: The guideline does not explicitly prohibit anything as far as I know. They only recommend the most secure stuff. In this case [they only recommend Argon2](https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TG02102/BSI-TR-02102-1.pdf?__blob=publicationFile&v=6#page=72): 
> If the use of a a cryptographic hardware token for password-based key derivation is not possible,
the hash function Argon2id [14] should be used.

I would suggest the same as for the cipher modes: If there is no important dependency we should stay consistent and add or remove both. 


Additional Note: [Wikipedia quote](https://en.wikipedia.org/wiki/PBKDF2#Alternatives_to_PBKDF2):

> One weakness of PBKDF2 is that while its number of iterations can be adjusted to make it take an arbitrarily large amount of computing time, it can be implemented with a small circuit and very little RAM, which makes brute-force attacks using [application-specific integrated circuits](https://en.wikipedia.org/wiki/Application-specific_integrated_circuit) or [graphics processing units](https://en.wikipedia.org/wiki/Graphics_processing_unit) relatively cheap.

Not critical, but the reason to create a competition to create a better algorithm (which was Argon2).

FAlbertDev: For now, I will remove the prohibition as suggested.

src/build-data/policy/bsi.txt
FAlbertDev: I'll remove it for now.

src/build-data/policy/bsi.txt
mouse07410: > I would suggest that we try to stay consistent. For example, why prohibiting the EAX mode, if there is no security flaw (as far as I know). IMO we have three possibilities:
> 
> 1) We continue prohibiting XTS and all other modes not explicitly recommended

I don't like this option.

> 2) We only prohibit modes with discovered security flaws (however,this requires literature research, which is the reason guidelines like the one of the BSI exist)

I could **easily** live with this choice.

> 3) We activate the recommended modes but prohibit none.

This seems the best and the soundest approach. Though might want to combine with "prohibit nodes work discovered [practical, exploitable] security flaws".



> I think we should get some feedback on these three approaches and apply them to other categories as well.

Indeed.

FAlbertDev:
I've disabled the PQC algorithms kyber and dilithium for now, since they are not yet wanted in the current policy. Maybe these will be activated in the next iteration. (4fe8b03)

FAlbertDev:
I would suggest that we leave the current changes for now. In the following weeks we are planning to agree with the BSI about the goals of the policy (including prohibition strategy, etc.). Afterward, we may need further adaptions. 
Until then, these changes should (at least) enable the TLS 1.3 modules in this policy.

src/build-data/policy/bsi.txt
reneme: From what I understand it is not strictly needed by any other module. Though, it allows applications to implement their own RSA encryption padding scheme. Basically, do the padding yourself and then encrypt with "EME Raw". No reason to prohibit that.

reneme (approved):

randombit:
Please rebase and squash before merge

src/build-data/policy/bsi.txt
lieser: I agree the `needed` is misleading here. I'm not aware of any module there `eme_raw` and `raw_hash` are needed.
It's just that we don't want to prohibit them in case a user needs them to implement a special (corner) case.

Issue #3324, opened 2023-02-25
Reduce proliferation of public headers related to TLS


randombit:
I just noticed that in the course of the session manager changes we've introduced a number of new public headers related to the session managers. We can probably consolidate these somewhat (cc @reneme)

More generally we are now up to 24 public headers for TLS which seems quite excessive - lets see if some of these can be merged or otherwise eliminated.

reneme:
The bulk work on TLS 1.3 for Botan 3.0.0 is coming to a close with #3322, #3316 and one more change to generalize the `tls_session_established()` callback ([along those lines](https://github.com/randombit/botan/pull/2988#issuecomment-1351250465)). Once those are in, I'll look into header consolidation.

For some guidance: Do you prefer merging class declarations into bigger headers, or should we introduce virtual headers (e.g. `tls_session_managers.h`) that just include the specific downstream headers?

randombit:
On the whole I'd prefer consolidation where it makes sense because for a user trying to figure out what header(s) they need, it's imo easier if there is just one big header to look through.

Particular ones to look at:

*  "tls_session_manager*" + tls_ticket_13.h should probably all be a single header.
* Any reason that `tls_algos.h` does not contain also `tls_signature_scheme.h`?
* `tls_magic.h` is public but parts are clearly public API (eg `Connection_Side`) while others parts are purely internal. This should probably be split into two headers, one public one internal

randombit:
That said we may well want to also consider a virtual header `<botan/tls.h>` that includes more or less everything a user might need. (Probably skipping headers that are only necessary if you are doing esoteric handshake inspection or extension tampering stuff, like `tls_messages.h` and `tls_extensions.h`)

randombit:
Eh the virtual header helps, and at this point we're stuck with whatever we have until 4.0 I guess

PR #3483, opened 2023-04-06
Add virtual TLS header

lieser:
- Add virtual `tls.h` header for commonly used TLS headers
- Make `Record_Type` internal

#3324

### DIFF
--- a/src/lib/tls/tls_channel_impl.h
+++ b/src/lib/tls/tls_channel_impl.h
@@ -29,6 +29,17 @@ namespace TLS {
 class Client;
 class Server;
 
+enum class Record_Type: uint8_t {
+   Invalid            = 0,  // RFC 8446 (TLS 1.3)
+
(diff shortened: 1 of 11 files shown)


src/lib/tls/tls_channel_impl.h
```diff
@@ -29,6 +29,17 @@ namespace TLS {
 class Client;
 class Server;
 
+enum class Record_Type: uint8_t {
```
lieser, line 32: Unsure if this is the best place, but did not want to create another internal header just for this enum.

src/lib/tls/tls_magic.h
lieser, line 1: The `Transcript_Hash` defined in here seems to be also only needed internally, at least the full declaration. But don't see how the using directive could be forward declared.

randombit (approved):

PR #3484, opened 2023-04-06
Doc: Breathe for TLS

reneme:
### Pull Request Dependencies

* #3464 

### Description

This moves API documentation to Breathe and at the same time updates the documentation to reflect the changes introduced with TLS 1.3. More could be done here, but I think it's a good first increment.

This is currently not touching the `TLS::Stream` documentation yet. That'll happen later, with a fresh mind.


reneme:

PR #3485, opened 2023-04-06
FIX: Avoid unnecessary exception

reneme:
While testing 3.0 RC1 internally, @weberph2 found that `std::stoul()` threw an exception while trying to read `BOTAN_MLOCK_POOL_SIZE`. For reasons, the underlying `std::getenv_s` on Windows returns true despite the environment variable not being set.

This is just adding another check before provoking an unnecessary exception.

@weberph2: Please check whether I paraphrased your finding correctly.


weberph2:
@reneme Yes, thank you!

randombit (approved):

PR #3489, opened 2023-04-09
Allow the user to disable internal modules

### LINKED ISSUE #3488: ./configure.py --disable-modules=tls fails
```
$ ./configure.py --disable-modules=tls
...
Module 'tls' is meant for internal use only
```

Fallout from changing it to a virtual. But this error does not make any sense; disabling the module, and all of its dependencies, is perfectly reasonable.

I'd like to get this fixed before 3.0.0 if possible.

(cc @reneme)

randombit:
Fixes #3488


reneme (approved):
Makes sense, thanks.

Issue #3487, opened 2023-04-09
'++crypto' is not a recognized feature for this target (ignoring feature)


oviano:
I get this warning together with a similar one about "++simd" on my macOS arm64 host when building native macOS, arm64 iOS or arm64 Android.

I also get a similar warning about when building Android armv7 about "+fpu=neon".

Here's an example with iOS arn64:

```
python3 configure.py --minimized --enable-modules=tls12,tls13,sqlite3,sessions_sqlite3,srp6,ed25519,curve25519,system_rng,bcrypt,crc32,hkdf,chacha20poly1305,nist_keywrap,ffi,aes_armv8,aes_ni,aes_power8,aes_vperm,idea_sse2,serpent_avx2,shacal2_armv8,shacal2_avx2,shacal2_x86,sm4_armv8,rdseed,sha1_armv8,sha1_sse2,sha1_x86,sha2_32_armv8,sha2_32_bmi2,sha2_32_x86,sha2_64_bmi2,sha3_bmi2,zfec_sse2,zfec_vperm,argon2_avx2,argon2_ssse3,processor_rng,chacha_avx2,ghash_cpu,ghash_vperm,simd,simd_avx2 --os=ios --cpu=arm64 --cc=clang --cc-min-version=13.0 --cc-abi-flags="-arch arm64" --disable-shared
   INFO: configure.py invoked with options "--minimized --enable-modules=tls12,tls13,sqlite3,sessions_sqlite3,srp6,ed25519,curve25519,system_rng,bcrypt,crc32,hkdf,chacha20poly1305,nist_keywrap,ffi,aes_armv8,aes_ni,aes_power8,aes_vperm,idea_sse2,serpent_avx2,shacal2_armv8,shacal2_avx2,shacal2_x86,sm4_armv8,rdseed,sha1_armv8,sha1_sse2,sha1_x86,sha2_32_armv8,sha2_32_bmi2,sha2_32_x86,sha2_64_bmi2,sha3_bmi2,zfec_sse2,zfec_vperm,argon2_avx2,argon2_ssse3,processor_rng,chacha_avx2,ghash_cpu,ghash_vperm,simd,simd_avx2 --os=ios --cpu=arm64 --cc=clang --cc-min-version=13.0 --cc-abi-flags=-arch arm64 --disable-shared"
   INFO: Configuring to build Botan 3.0.0-rc1 (revision git:604672a06366b7ddb94632d1027c546c49a67caa)
[...]

PR #3491, opened 2023-04-10
This deals with incompatibilities between clang and gcc regarding names of function target attribute isas for arm targets

oviano:
This deals with the warnings described in https://github.com/randombit/botan/issues/3487

From clang 16.0.0 there is greater compatibility between the two compilers regarding this but since Apple clang has only just upgraded to clang 15.0 with Xcode 14.3 it seems that is some way off at least within the Apple ecosystem.

### DIFF
--- a/src/lib/block/aes/aes_armv8/aes_armv8.cpp
+++ b/src/lib/block/aes/aes_armv8/aes_armv8.cpp
@@ -54,7 +54,11 @@ namespace Botan {
 /*
 * AES-128 Encryption
 */
+#if defined(BOTAN_BUILD_COMPILER_IS_CLANG)
+BOTAN_FUNC_ISA("crypto")
+#else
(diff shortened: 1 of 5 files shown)


randombit:
Sorry this got ignored. In the current tree these attribute strings are centralized in a util header so there is an easier place to update them if required. Are you seeing these warnings with a recent XCode? Possibly the warnings are gone anyway and we can just use the GCC strings.

oviano:
> Sorry this got ignored. In the current tree these attribute strings are centralized in a util header so there is an easier place to update them if required. Are you seeing these warnings with a recent XCode? Possibly the warnings are gone anyway and we can just use the GCC strings.

I don't think it's an issue anymore.

randombit:
I'll close this for now, if it reoccurs we are anyway in a better position to address it

Issue #3510, opened 2023-04-15
--library-suffix giving strange results


crazydef:
Hi,

I'm building botan (master from about 12 hours ago) on Windows with the command line argument `--library-suffix=.x64.Debug` (for debug builds - I have other configurations where `Debug` is replaced with different things).

The resulting lib and pdb files are named slightly and really strangely, respectively.

The lib files are named `botan.x64.Debug-3.lib` (I would have expected `botan-3.x64.Debug.lib`) and `botan.x64.Debug-3.x64.Debug.pdb` (I would have expected just the file extension to change from the corresponding lib file.)

randombit:
Wow that is not right at all, thanks for the report, I'll look at getting this fixed.

PR #3511, opened 2023-04-19
Fix bugs related to --library-suffix

randombit:
GH #3510


configure.py
lieser, line 1: The `soname_pattern_base` is currently still set to `libbotan{lib_suffix}-{version_major}.%s`. Probably makes sense to keep them consistent.

The question is how `soname_pattern_abi` and `soname_pattern_patch` should then behave. The `.{abi_rev`/`.{version_minor}.{version_patch}` should probably continue to be directly behind `{version_major}`.

PR #3520, opened 2023-04-25
Fixes: RFC-compliance of TLS (zero premaster sec + disabled ECC extensions)

FAlbertDev:
Changes as requested in Issue #2754:
1) Abort if an all-zero premaster secret was computed by the client or the server using the group X25519.
2) Enforce that the ECC extensions are not used when no ECC cipher suite is offered in the ClientHello. Also abort, if a client violates this rule.


src/lib/tls/tls12/msg_client_kex.cpp
```diff
@@ -22,6 +23,25 @@
...
+      //   computed premaster secret is the all-zero value and abort the
+      //   handshake if so, as described in Section 6 of [RFC7748].
+      if(named_group == Group_Params::X25519/* || named_group == Group_Params::X448 (not implemented)*/)
+         {
+         // Side channel resistent (!!!) check for 0
+         uint8_t or_sum = 0;
+         for(uint8_t s_byte : shared_secret)
+            or_sum |= s_byte;
+
+         return (or_sum != 0);
```
randombit, line 39: For this loop use the const time utilities in `ct_utils.h`.

randombit: Added `CT::all_zeros` in #3522 to make this easy

FAlbertDev: Nice, thanks!

randombit (changes requested):
Please rebase against master rather than merging commits from master into a branch.

randombit (approved):
LGTM. @reneme I don't know if you want to review as well.

src/lib/tls/msg_client_hello.cpp
```diff
@@ -463,6 +463,24 @@ Client_Hello_12::Client_Hello_12(std::unique_ptr<Client_Hello_Internal> data)
...
+   //    propose any ECC cipher suites.
+   if(m_data->extensions().has<Supported_Groups>() || m_data->extensions().has<Supported_Point_Formats>())
+      {
+      auto offered_ciphers = m_data->ciphersuites();
+      if (std::none_of(offered_ciphers.begin(), offered_ciphers.end(), [](uint16_t cipherSuiteCode) {
+         std::optional<Ciphersuite> cs = Ciphersuite::by_id(cipherSuiteCode);
+         // true for ECC cipher suites or TLS 1.3 cipher suites
+         return cs.has_value() && (cs.value().ecc_ciphersuite() ||
+                                   cs.value().auth_method() == Auth_Method::UNDEFINED ||
+                                   cs.value().kex_method() == Kex_Algo::UNDEFINED); }))
```
reneme, line 478: Maybe simplify to:

```suggestion
         return cs.has_value() && (cs.value().ecc_ciphersuite() ||
                                   cs.usable_in_version(Protocol_Version::TLS_V13)); }))
```

Or even add a predicate method to `Ciphersuite`, like:

```bool Ciphersuite::is_tls12_suite()```

File: src/lib/tls/msg_client_hello.cpp [hunk 3]
```diff
@@ -463,6 +463,24 @@ Client_Hello_12::Client_Hello_12(std::unique_ptr<Client_Hello_Internal> data)
...
+         {
+         throw TLS_Exception(Alert::HandshakeFailure,
+                             "Client sent elliptic curves extension, but no ECC cipher suite was proposed");
```
reneme, line 481: It not otherwise stated in the RFC, I think `IllegalParameter` would be more consistent here.

cf. RFC 8446 6.2

```
illegal_parameter:  A field in the handshake was incorrect or
      inconsistent with other fields.  This alert is used for errors
      which conform to the formal protocol syntax but are otherwise
      incorrect.
handshake_failure:  Receipt of a "handshake_failure" alert message
      indicates that the sender was unable to negotiate an acceptable
      set of security parameters given the options available.
```

```suggestion
         throw TLS_Exception(Alert::IllegalParameter,
                             "Client sent elliptic curves extension, but no ECC cipher suite was proposed");
```

src/lib/tls/msg_client_hello.cpp
```diff
@@ -463,6 +463,24 @@ Client_Hello_12::Client_Hello_12(std::unique_ptr<Client_Hello_Internal> data)
          m_data->extensions().add(new Renegotiation_Extension());
          }
       }
+   // RFC 8446 4.2.3
```
reneme, line 466: I believe this reference is not correct. Can't find the quoted paragraph in RFC 8446. Instead its in [RFC 8422 4](https://www.rfc-editor.org/rfc/rfc8422#section-4).

src/lib/tls/msg_client_hello.cpp
```diff
@@ -513,12 +531,22 @@ Client_Hello_12::Client_Hello_12(Handshake_IO& io,
    if(policy.support_cert_status_message())
       m_data->extensions().add(new Certificate_Status_Request({}, {}));
 
-   auto supported_groups = std::make_unique<Supported_Groups>(policy.key_exchange_groups());
-   if(!supported_groups->ec_groups().empty())
+   // RFC 8446 4.2.3
```
reneme, line 534: Same as above: Not in RFC 8446 but in [RFC 8422 4](https://www.rfc-editor.org/rfc/rfc8422#section-4).

src/lib/tls/msg_client_hello.cpp
```diff
@@ -513,12 +531,22 @@ Client_Hello_12::Client_Hello_12(Handshake_IO& io,
...
+   //    The client MUST NOT include these extensions [Supported Elliptic Curves Extension
+   //    and Supported Point Formats Extension] in the ClientHello message if it does not
+   //    propose any ECC cipher suites.
+   auto offered_ciphers = m_data->ciphersuites();
```
reneme, line 538:
```suggestion
   const auto& offered_ciphers = m_data->ciphersuites();
```

src/lib/tls/msg_client_hello.cpp
```diff
@@ -513,12 +531,22 @@ Client_Hello_12::Client_Hello_12(Handshake_IO& io,
...
+      std::optional<Ciphersuite> cs = Ciphersuite::by_id(cipherSuiteCode);
+      return cs.has_value() && (cs.value().ecc_ciphersuite()); }))
       {
-      m_data->extensions().add(new Supported_Point_Formats(policy.use_ecc_point_compression()));
+      auto supported_groups = std::make_unique<Supported_Groups>(policy.key_exchange_groups());
+      if(!supported_groups->ec_groups().empty())
+         {
+         m_data->extensions().add(new Supported_Point_Formats(policy.use_ecc_point_compression()));
+         }
+      m_data->extensions().add(supported_groups.release());
```
reneme, line 548: Nit. Though, no need to recreate the `unique_ptr`.
```suggestion
      m_data->extensions().add(std::move(supported_groups));
```

src/lib/tls/msg_client_hello.cpp
```diff
@@ -513,12 +531,22 @@ Client_Hello_12::Client_Hello_12(Handshake_IO& io,
...
+   if (std::any_of(offered_ciphers.begin(), offered_ciphers.end(), [](uint16_t cipherSuiteCode) {
+      std::optional<Ciphersuite> cs = Ciphersuite::by_id(cipherSuiteCode);
+      return cs.has_value() && (cs.value().ecc_ciphersuite()); }))
       {
-      m_data->extensions().add(new Supported_Point_Formats(policy.use_ecc_point_compression()));
+      auto supported_groups = std::make_unique<Supported_Groups>(policy.key_exchange_groups());
+      if(!supported_groups->ec_groups().empty())
+         {
+         m_data->extensions().add(new Supported_Point_Formats(policy.use_ecc_point_compression()));
+         }
```
reneme, line 547: Question: Is it possible/legal that we might offer some ECC-based ciphersuites (outer `if` statement is true) but don't advertise any supported ECC groups (this `if` statement is also true)?

src/lib/tls/tls12/msg_client_kex.cpp
```diff
@@ -133,6 +133,15 @@ Client_Key_Exchange::Client_Key_Exchange(Handshake_IO& io,
                                                           rng,
                                                           policy);
 
+         // RFC 8422 - 5.11.
+         //   With X25519 and X448, a receiving party MUST check whether the
+         //   computed premaster secret is the all-zero value and abort the
+         //   handshake if so, as described in Section 6 of [RFC7748].
```
reneme, line 139: I suppose it doesn't hurt to do that check in TLS 1.3 as well. It should go into `tls_extensions_key_share.cpp` line 109 after calling `Callbacks::tls_ephemeral_key_agreement`.

src/lib/tls/tls12/msg_client_kex.cpp
```diff
@@ -133,6 +133,15 @@ Client_Key_Exchange::Client_Key_Exchange(Handshake_IO& io,
...
+         if(curve_id == Group_Params::X25519 && CT::all_zeros(shared_secret.data(), shared_secret.size()).is_set())
+            {
+            throw TLS_Exception(Alert::InternalError, "Bad X25519 key exchange");
```
reneme, line 142: Internal errors are meant for intrinsic errors of the implementation (e.g. out-of-memory). I guess:

```suggestion
            throw TLS_Exception(Alert::DecryptError, "Bad X25519 key exchange");
```

RFC 8446:
```
decrypt_error:  A handshake (not record layer) cryptographic
      operation failed, including being unable to correctly verify a
      signature or validate a Finished message or a PSK binder.
```

src/lib/tls/tls12/msg_client_kex.cpp
```diff
@@ -318,6 +327,20 @@ Client_Key_Exchange::Client_Key_Exchange(const std::vector<uint8_t>& contents,
...
+               {
+                  // RFC 8422 - 5.11.
+                  //   With X25519 and X448, a receiving party MUST check whether the
+                  //   computed premaster secret is the all-zero value and abort the
+                  //   handshake if so, as described in Section 6 of [RFC7748].
+                  BOTAN_ASSERT_NOMSG(state.server_kex()->params().size() >= 3);
+                  Group_Params group = static_cast<Group_Params>(state.server_kex()->params().at(2));
+                  if(group == Group_Params::X25519 && CT::all_zeros(shared_secret.data(), shared_secret.size()).is_set())
+                     {
+                     throw TLS_Exception(Alert::InternalError, "Bad X25519 key exchange");
```
reneme, line 340:
```suggestion
                     throw TLS_Exception(Alert::DecryptError, "Bad X25519 key exchange");
```

src/lib/tls/tls12/msg_client_kex.cpp
```diff
@@ -318,6 +327,20 @@ Client_Key_Exchange::Client_Key_Exchange(const std::vector<uint8_t>& contents,
             if(ka_key->algo_name() == "DH")
                shared_secret = CT::strip_leading_zeros(shared_secret);
```
reneme, line 329: Unrelated note: Above is a hard-coded key exchange. This should probably use `Callbacks::tls_ephemeral_key_agreement()` as of #3322 😨. I'll look into this separately.

reneme (changes requested):
Several minor nits and TLS alert consistency suggestions. Actual logic LGTM.

src/lib/tls/tls12/msg_client_kex.cpp
reneme: See: https://github.com/randombit/botan/pull/3539

src/lib/tls/msg_client_hello.cpp
randombit: I think technically it is allowed, but wouldn't make sense to do since the ECC suites could not be negotiated.RFC 8422 states that if impl is "unable to complete the ECC handshake while restricting itself to the enumerated curves and point formats, it MUST NOT negotiate the use of an ECC cipher suite." and TLS 1.3 has similar language.

src/lib/tls/msg_client_hello.cpp
FAlbertDev: Oops, this happens when you copy paste and forget to adept all fields...

src/lib/tls/msg_client_hello.cpp
FAlbertDev: I simply copied the previous code and put it in the new if clause. I think an additional check can't hurt, but I'll remove it to keep it more simple.

src/lib/tls/tls12/msg_client_kex.cpp
FAlbertDev: I remember that I selected the same alert OpenSSL uses in the same situation. I was also a bit confused by the term "decrypt", because it is no decryption but a key exchange. However, I think you are right because it is indeed a "cryptographic operation". that has (somehow) "failed".

FAlbertDev:
Uff, it seems that the boringssl testsuite is not happy about the strict prohibition of ECC extensions in case of no ECC ciphersuites.
Any suggestions how to solve this problem?

From the pipeline:
```
FAILED (Basic-Server-RSA-DTLS-Async-SplitHandshakeRecords)
unexpected failure: local error 'remote error: illegal parameter', child error 'exit status 1', stdout:

stderr:
Unmapped error: 'Client sent elliptic curves extension, but no ECC cipher suite was proposed'


PASSED (ClientAuth-RSA-Client-TLS13-DTLS-Async-SplitHandshakeRecords)
...
FAILED (EmptyPSKHint-Server-DTLS-Async-SplitHandshakeRecords)
unexpected failure: local error 'remote error: illegal parameter', child error 'exit status 1', stdout:

stderr:
Unmapped error: 'Client sent elliptic curves extension, but no ECC cipher suite was proposed'


PASSED (EmptyPSKHint-Client-DTLS-Async-SplitHandshakeRecords)
...
FAILED (Basic-Server-RSA-DTLS-Async-PackHandshake)
unexpected failure: local error 'remote error: illegal parameter', child error 'exit status 1', stdout:

stderr:
Unmapped error: 'Client sent elliptic curves extension, but no ECC cipher suite was proposed'


PASSED (ClientAuth-Server-DTLS-Async-PackHandshake)
...
FAILED (EmptyPSKHint-Server-DTLS-Async-PackHandshake)
unexpected failure: local error 'remote error: illegal parameter', child error 'exit status 1', stdout:

stderr:
Unmapped error: 'Client sent elliptic curves extension, but no ECC cipher suite was proposed'


PASSED (EmptyPSKHint-Client-DTLS-Async-PackHandshake)
...
exit status 1
FAIL	boringssl.googlesource.com/boringssl/ssl/test/runner	127.859s
Ran for 144 seconds
Command 'go test -pipe -num-workers 8 -shim-path /home/runner/work/botan/botan/build/botan_bogo_shim -shim-config /home/runner/work/botan/botan/source/src/bogo_shim/config.json' failed with error code 1
Error: Process completed with exit code 1.
```

src/lib/tls/msg_client_hello.cpp
```diff
@@ -513,12 +530,19 @@ Client_Hello_12::Client_Hello_12(Handshake_IO& io,
...
+   //    and Supported Point Formats Extension] in the ClientHello message if it does not
+   //    propose any ECC cipher suites.
+   const auto& offered_ciphers = m_data->ciphersuites();
+   if (std::any_of(offered_ciphers.begin(), offered_ciphers.end(), [](uint16_t cipherSuiteCode) {
+      std::optional<Ciphersuite> cs = Ciphersuite::by_id(cipherSuiteCode);
+      return cs.has_value() && cs.value().ecc_ciphersuite(); }))
       {
+      auto supported_groups = std::make_unique<Supported_Groups>(policy.key_exchange_groups());
       m_data->extensions().add(new Supported_Point_Formats(policy.use_ecc_point_compression()));
+      m_data->extensions().add(std::move(supported_groups));
```
reneme, line 544: Seeing it like that now: you could avoid the variable altogether, no?

```suggestion
      m_data->extensions().add(new Supported_Point_Formats(policy.use_ecc_point_compression()));
      m_data->extensions().add(new Supported_Groups(policy.key_exchange_groups());
```

reneme (changes requested):
Two small nits to go.

If BoGo doesn't like it, then we might consider just dropping this check. Its likely that other implementations don't adhere to the RFCs that strictly either.

And as far as I can tell, there's no security implication to ignoring this constraint, right? No need to be overly strict here, I think.

For completeness: We could also introduce a `TLS::Policy::strict_ecc_extension_mode()` (or similar) that could disable this check explicitly. But I don't find it worth the headache, TBH.

FAlbertDev:
> If BoGo doesn't like it, then we might consider just dropping this check. Its likely that other implementations don't adhere to the RFCs that strictly either.
> 
> And as far as I can tell, there's no security implication to ignoring this constraint, right? No need to be overly strict here, I think.

Yeah, I also think it might not be worth the trouble. I also can't see any security implications leaving this check out.

reneme (approved):
Thanks. Please squash and rebase to latest `master`.

PR #3526, opened 2023-04-26
Detect and error during configuration if an unsupported compiler is detected

randombit:
Rather than allowing the build to continue and eventually fail in a way that does not immediately indicate the reason.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1194,6 +1194,7 @@ class CompilerInfo(InfoObject): # pylint: disable=too-many-instance-attributes
                 'binary_name': None,
                 'linker_name': None,
                 'macro_name': None,
+                'minimum_supported_version': None,
                 'output_to_object': '-o ',
                 'output_to_exe': '-o ',
                 'add_include_dir_option': '-I',
@@ -1264,6 +1265,7 @@ class CompilerInfo(InfoObject): # pylint: disable=too-many-instance-attributes
         self.visibility_build_flags = lex.visibility_build_flags
         self.warning_flags = lex.warning_flags
         self.werror_flags = lex.werror_flags
+        self.minimum_supported_version = lex.minimum_supported_version
 
     def cross_check(self, os_info, arch_info, all_isas):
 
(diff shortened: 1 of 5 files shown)


configure.py
```diff
@@ -3186,28 +3186,37 @@ def calculate_cc_min_version(options, ccinfo, source_paths):
...
     cc_version = "%d.%d" % (major_version, minor_version)
-    logging.info('Auto-detected compiler version %s', cc_version)
+    logging.info('Auto-detected compiler version %s %s', cxx, cc_version)
+
+    if cxx == 'gcc' and major_version < 11:
+        logging.error('This version of Botan requires at least GCC 11')
+    elif cxx == 'clang' and major_version < 14:
+        logging.error('This version of Botan requires at least Clang 14')
+    elif cxx == 'msvc' and (major_version < 19 or (major_version == 19 and minor_version < 30)):
+        logging.error('This version of Botan requires at least Visual C++ 2022')
```
reneme, line 3219: Two things:

* Would it be worthwhile to move the supported versions out of the script? E.g. configurable in `version.txt`.
* Shouldn't this differentiate between clang and XCode?

src/build-data/detect_version.cpp
```diff
@@ -35,24 +35,28 @@
 
    Wikipedia has a mapping table from Apple Clang version to underlying
    LLVM version: https://en.wikipedia.org/wiki/Xcode
+
+   We omit any mapping logic for versions before XCode 13 since we no longer
+   support those versions. The only reason to detect them is so we can tell that
+   the version of Clang is too old and stop configuration immediately.
```
reneme, line 41: That answers my question re:XCode from above. 😅

reneme (approved):
I think that's a useful addition that'll avoid confusion due to the new supported set. Externalizing the supported version into a config file is purely optional, but would be cleaner IMO.

configure.py
randombit: That would be nice. That also allows us to generalize this check for other compilers (ICC, etc), if/when the minimum version for those become apparent.

Best place is probably the compiler info files since that is nominally where all knowledge about the compiler features is stored.

src/build-data/detect_version.cpp
randombit: We may want to split up Clang vs XCode anyway - they are kind of the same, but in a lot of ways it feels like the difference between GCC and Clang is smaller than the difference between Clang and XCode.

PR #3539, opened 2023-05-03
FIX: allow custom KEX logic for TLS 1.2 server

reneme:
When implementing #3322 I missed applying the KEX customization hook for the TLS 1.2 server (cf. `TLS::Callbacks::tls_ephemeral_key_agreement()`). Before Botan 3.0 this customization point wasn't available at all: So hopefully this is not a big deal. 😞

To fix this, I had to add `Server_Key_Exchange::shared_group()` to be able to access the negotiated KEX group in the `Client_Key_Exchange` c'tor.

To avoid some confusion in understanding the control-flow: The c'tor of `Server_Key_Exchange` is called on the server after receiving a ClientHello. This is where the server negotiates a KEX group and generates an ephemeral key using `TLS::Callbacks::tls_generate_ephemeral_key()`. That was adapted in #3322 already.
In the next flight, the server invokes the c'tor of `Client_Key_Exchange` (deserialization) and derives the shared secret with the client's public value using `TLS::Callbacks::tls_ephemeral_key_agreement()`. This was hard-coded in the c'tor so far.

Also, I changed the `Server_Key_Exchange::server_kex_key()` to return a `PK_Key_Agreement_Key` rather than a generic `Private_Key` to avoid the need for a `dynamic_cast`. The class is marked as `BOTAN_UNSTABLE_API`, so breaking the API in fine, right?

### DIFF
--- a/src/lib/tls/tls12/msg_server_kex.cpp
+++ b/src/lib/tls/tls12/msg_server_kex.cpp
@@ -258,7 +258,7 @@ bool Server_Key_Exchange::verify(const Public_Key& server_key,
 #endif
    }
 
-const Private_Key& Server_Key_Exchange::server_kex_key() const
+const PK_Key_Agreement_Key& Server_Key_Exchange::server_kex_key() const
    {
    BOTAN_ASSERT_NONNULL(m_kex_key);
    return *m_kex_key;
--- a/src/lib/tls/tls12/msg_server_kex.cpp
+++ b/src/lib/tls/tls12/msg_server_kex.cpp
@@ -51,7 +51,7 @@ Server_Key_Exchange::Server_Key_Exchange(Handshake_IO& io,
       {
       const std::vector<Group_Params> dh_groups = state.client_hello()->supported_dh_groups();
 
-      Group_Params shared_group = Group_Params::NONE;
+      m_shared_group = Group_Params::NONE;
 
       /*
       If the client does not send any DH groups in the supported groups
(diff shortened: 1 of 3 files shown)


randombit:
> Also, I changed the Server_Key_Exchange::server_kex_key() to return a PK_Key_Agreement_Key rather than a generic Private_Key to avoid the need for a dynamic_cast. The class is marked as BOTAN_UNSTABLE_API, so breaking the API in fine, right?

Seems fine to me, especially as we're returning a subclass of the previous return type.

src/lib/tls/tls12/msg_server_kex.cpp
```diff
@@ -60,20 +60,28 @@ Server_Key_Exchange::Server_Key_Exchange(Handshake_IO& io,
...
 
-      if(shared_group == Group_Params::NONE)
+      if(m_shared_group.value() == Group_Params::NONE)
          throw TLS_Exception(Alert::HandshakeFailure,
                "Could not agree on a DH group with the client");
 
-      BOTAN_ASSERT(group_param_is_dh(shared_group), "DH groups for the DH ciphersuites god");
-
-      m_kex_key = state.callbacks().tls_generate_ephemeral_key(shared_group, rng);
+      BOTAN_ASSERT(group_param_is_dh(m_shared_group.value()), "DH groups for the DH ciphersuites good");
```
randombit, line 74: I'm sure you thought "god" was a typo, actually was a reference to https://knowyourmeme.com/memes/blood-for-the-blood-god

We should change this to something reasonable eg "DH ciphersuite is using a finite field group"

randombit (approved):
LGTM

src/lib/tls/tls12/msg_server_kex.cpp
reneme: Yikes! 🤣 made my day.

Issue #3099, opened 2022-12-03
Implement SPHINCS+


randombit:
Hash based signatures are great but the stateful nature of XMSS is a real footgun and difficult operationally.

securitykernel:
An implementation is planned during [BSI](https://www.bsi.bund.de/EN) project 481.
 Preliminary schedule for the implementation is Mar '23 - May '23.

securitykernel:
Updated schedule for the implementation is May '23 - July '23.

PR #3549, opened 2023-05-17
SPHINCS+

reneme:
## Pull Request Dependencies

* ~~#3564~~

## Description

This is adding SPHINCS+ as a PQC signature algorithm.

Note that the implementation makes heavy use of the strong type additions in #3564. With SPHINCS+ handling various different trees, buffers, hashes, ... we think it improves understandability of the implementation substantially. Also, those auxiliary building blocks allow for an (almost) zero-copy implementation of SPHINCS+ by making heavy use of std::span<> (and the new `StrongSpan<>` that carries type information).

The inner algorithms and building blocks of SPHINCS+ take a lot of inspiration from [the reference implementation](https://github.com/sphincs/sphincsplus/) and apply modern C++ mechanics where appropriate.

## Future Work (Potential Optimizations)

SPHINCS+ is built from WOTS+, XMSS (along with XMSS_MT) and FORS. Currently, there is no shared code between the XMSS implementation and SPHINCS+. Deeper integration may or may not be feasible. In any case it would be substantial future work. Though, given that the XMSS implementation can use multi-threading it might be worthwhile to look into at a later point.

All "tweaked hash" invocations in SPHINCS+ always hash a common prefix before their payload (namely, the `Public Seed`). Caching the hash state of `Thash(Public Seed)` is a natural optimization also used by the reference implementation. We could use `HashFunction::copy_state()` for this, but it forces us to (a) re-allocate a `std::unique_ptr<>` for each hash calculation and (b) somewhat stands in the way of avoiding the virtual method calls in `HashFunction`, despite knowing the hash type at compile time. Those allocations result in about 15% of the run time on my machine (from a quick and dirty gprof profiling). As a result, just re-calculating the state is faster than using `copy_state()`. We'd need some (internal) API that allows more direct access here.

## Open TODOs

* [x] Internal refactorings
* [x] Finish up integration in X.509
* [x] Clean up commit history
* [x] Tests
  * [x] Generic round-trip tests
  * [x] Decide which KAT vectors to use and whether to disable some vectors due to timing constraints. The full KAT body would likely take way to long.

Issue link: #3099


randombit (commented):
Did a very quick scan

src/lib/pubkey/sphincsplus/sp_hash.h
```diff
@@ -0,0 +1,129 @@
...
+#include <botan/sp_parameters.h>
+#include <botan/secmem.h>
+#include <botan/internal/sp_address.h>
+#include <botan/internal/sp_types.h>
+
+namespace Botan {
+
+#if defined(BOTAN_HAS_SHA2_32) && defined(BOTAN_HAS_SHA2_64) && defined(BOTAN_HAS_TRUNCATED_HASH) && defined(BOTAN_HAS_MGF1) && defined(BOTAN_HAS_HMAC)
+  #define BOTAN_CAN_USE_SHA_FOR_SHPINCSPLUS
+#endif
```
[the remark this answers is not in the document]
reneme: The idea was to allow building the library with SPHINCS+ support but without SHA2 or SHA3 respectively. For SHA2 one additionally needs the above modules.

I introduced the indirection via the additional `#define` to reuse it for enabling the relevant test code.

src/lib/pubkey/sphincsplus/sphincsplus.cpp
```diff
@@ -0,0 +1,421 @@
...
+#include <botan/internal/sp_hash.h>
+#include <botan/internal/sp_treehash.h>
+#include <botan/internal/sp_types.h>
+#include <botan/internal/sp_wots.h>
+#include <botan/internal/sp_xmss.h>
+#include <botan/internal/stl_util.h>
+
+#if not defined(BOTAN_HAS_SHAKE) and not defined(BOTAN_CAN_USE_SHA_FOR_SHPINCSPLUS)
+   static_assert(false, "botan module 'sphincsplus' is useful only when enabling at least 'shake' or 'sha2_32,sha2_64,mgf1,hmac,trunc_hash'");
+#endif
```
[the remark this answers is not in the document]
reneme: > I'd be inclined here to just hard require everything

That's what we did initially. But I found it a worthwhile use case to build the library with SPHINCS+ but without support for one of the hashes. There isn't much additional complexity to allow for this flexibility, in my opinion.

The static assert was added to provide guidance to users that might configure as `--minimized-build --enable-modules=sphincsplus` and end up with a dysfunctional module. A similar approach was implemented for Kyber (to require at least `kyber` or `kyber_90s` to be enabled). In contrast, the Kyber example uses actual sub modules that can model their respective transitive dependencies. We could do this here as well, maybe.

reneme:
> please rebase and apply clang-format.

Done.

src/lib/pubkey/sphincsplus/sp_address.h
```diff
@@ -0,0 +1,178 @@
...
+         m_address.fill(0);
+         set_type(type);
+         }
+
+      Sphincs_Address(std::array<uint32_t, 8> address)
+         {
+         std::copy(address.begin(), address.end(), m_address.begin());
+         }
+
+      Sphincs_Address& set_layer(HypertreeLayerIndex layer)
```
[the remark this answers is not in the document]
FAlbertDev: I like your suggestion. However:
Some of the setters of the Sphincs_Address are called in the most critical path of Sphincs+'s logic.
Every single hash computation of Sphincs+ is performed using a different address, so every hash computation requires an address update using one of the setters. We cannot copy the address each time without significant runtime loss.

src/tests/data/pubkey/sphincsplus.vec
```diff
@@ -0,0 +1,90 @@
...
...4A83614859B0E025549B18C03D2894F37B9C584E0B371212D68BCFED60B18570B0CF63645C808286B9765725754E4F751DAF403815182B21C4B301183E2799F6BC144F16F98A96BAFFFC8B0B63B505762E2F66E9DCB44802D04D80CA8B33731D7B62DA025AB83B4430CEB4CDD0DE2433C396227E90F0DBD0C827826F9209A7672846B78BACDB935A271F3F18BE486952826C29EDD709D910EDCA1A8FB598712A3776DDECE903F247A88E4247A093B5DB0AA8FBC51572263F12063BDC8DA8140843F581DABF765FCFCE299CE315E67E7C965D51DABFC2AD5EB72426E0C43D60551E1C56170CA73F89B6782C4027B5ECBD7EA87D611E82C3BF54CCE971A60A719BE00E3C81B60E7B375C1532852CEF9DAFB88A130B218AC0F4C232C8418120E14EEF266C2299F640C73B09F70779771C226FA4C3B8B00D52A195E2671201379E5245D5DCE08DF4AB4CBB9C2478D9B96F32413453988C875299C5E4A315623DE8ACB3F4F3039D2D7395A424C7C998FE270B6A21214AC1BDE4730FD8E991DFF3D38120BDC8223003A5A524E04A2FB9BA7FFBC2570076BF2A378E7C8F97F2CA4C74C4AA2B6B2A8EAC571053B9553A8A96A6BD83BF665D473C742B6B69CFEBB8763AF518A716C180A41C2A66E6F5394CC8262779200C3279995A5D7452C25B864A9C8EA424E1334262871B2AF99859CF79F71A256EC6F580068359DBD704BA65C4926132F39B723BD145C88479D7154E5F561AA0919217174BE0204B0E85F8ECB3850ADA9412E69DD919873AE9D2849BA615EB46D7B0EE036F2DD81C4D8D734FCBFBEADE3D3F8A039FAA2A2C9957E835AD55B22E75BF57BB556AC8
```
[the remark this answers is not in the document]
FAlbertDev: Yeah, there are pros and cons for your suggestion.
Pro: Obviously, much shorter and smaller file size. 
Con: The full signature helps for debugging, since the Sphincs+ signature is a concatenation of multiple parts. When searching for errors one could inspect which part of the signature is false to identify the respective part of the program.

IMO we should apply your suggestion anyway.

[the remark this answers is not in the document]
reneme: Done

src/lib/pubkey/sphincsplus/sp_address.h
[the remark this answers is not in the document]
randombit: OK if it is a measurable performance impact, fine to use mutations

reneme:
We did a few final clean-ups today and with #3564 merged this is now ready for review from our side.
A few things worth discussing:

### Intermediate Unit Tests

Along the way we added "intermediate" unit tests for FORS, WOTS and utils. Those were useful during development of those building blocks but should now be fully covered by the top-level integration tests (in `test_sphincsplus.cpp`). Hence, we could remove those intermediates for simplicity.

### Internal APIs

The module-internal APIs for the building blocks (basically the things in `sp_*.h`) are not exactly picture perfect (long parameter lists, inconsistent ordering, ...). Typically these APIs are used in just one or very few locations inside the module, though. IMHO, the strong types make this bearable. Also, if we ever look into unifying XMSS and SPHINCS+ we'll trip over those no matter what. My verdict: it's good enough as it is.

### Haraka

Currently, the API suggests that we support it. Though, it'll just throw `Not_Implemented`. Also, NIST doesn't seem to be interested in standardizing this.

### Robust Mode

This is neither implemented nor is there an API that would suggest otherwise. [NIST specifically said that it won't be part of the standard](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/4MBurXr58Rs/m/-1Ja0ZYyAQAJ) either.

src/lib/pubkey/sphincsplus/sphincsplus.cpp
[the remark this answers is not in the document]
randombit: OK. I would somewhat prefer a submodules approach here eg `sphincsplus_sha2` and `sphinxplus_sha3` which themselves depend on the relevant dependencies. That could also provide a path for encapsulating the operation of the hash algorithms, which would probably prove useful if/when Haraka support is added.

Also explicit submodules could expose `BOTAN_HAS_SPHINCSPLUS_SHAn` feature macros which would allow applications to test which hash functions are supported, without having to know about this one off macro.

*And* (if I ever get around to adding back support for conditional module dependencies) we could avoid the static assert by having `sphinxplus` module depend on `sphincsplus_sha2 | sphincsplus_sha3`.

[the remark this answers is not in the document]
reneme: Yeah. Plus it is consistent with Kyber and Dilithium where the AES modes were handled as submodules. On it.

src/cli/speed.cpp
```diff
@@ -319,79 +323,78 @@ class Speed final : public Command {
          points of the most interesting or widely used algorithms.
          */
 
-         return {
```
randombit, line 322: Maybe add a `clang-format off` annotation here since I don't think the reformat improves readability in the first place, and makes the diff noisy.

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_address.h
```diff
@@ -0,0 +1,158 @@
...
+   ForsTree = 3,
+   ForsTreeRootsCompression = 4,
+   WotsKeyGeneration = 5,
+   ForsKeyGeneration = 6
+};
+
+class BOTAN_TEST_API Sphincs_Address final {
+   private:
+      static constexpr size_t layer_offset = 0;
+      static constexpr size_t tree_offset = 1;  // 96bits
```
randombit, line 33: This comment is a little confusing/oblique - maybe a reference to the spec is warranted?

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_address.h
```diff
@@ -0,0 +1,158 @@
...
+
+   public:
+      using enum Sphincs_Address_Type;
+
+      Sphincs_Address(Sphincs_Address_Type type) {
+         m_address.fill(0);
+         set_type(type);
+      }
+
+      Sphincs_Address(std::array<uint32_t, 8> address) { std::copy(address.begin(), address.end(), m_address.begin()); }
```
randombit, line 49: I think this formatting is not accepted anymore after #3568 - you may need to rebase again and reformat

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_hash.cpp
```diff
@@ -0,0 +1,95 @@
...
+         throw Not_Implemented("Haraka is not yet implemented");
+      case Sphincs_Hash_Type::Shake256:
+#if defined(BOTAN_HAS_SPHINCS_PLUS_WITH_SHAKE)
+         return std::make_unique<Sphincs_Hash_Functions_Shake>(sphincs_params, pub_seed);
+#else
+         throw Not_Implemented("SHAKE is not available in this build");
+#endif
+   }
+
+   Botan::unreachable();
```
randombit, line 53: This whole function is pretty hard to understand, and it's not entirely obvious at first glance that the unreachable really is unreachable. Also if we are passed an invalid `enum` value somehow, we have UB.

I'd suggest something like

```
switch(sphinx_params.hash_type()) {
   #if defined(BOTAN_HAS_SPHINCS_PLUS_WITH_SHA2)
   case Sphincs_Hash_Type::Sha256:
            return std::make_unique<Sphincs_Hash_Functions_Sha2>(sphincs_params, pub_seed);
   #endif 
   // likewise for SHA-3

   default:
      throw Not_Implemented("SPHINCS+ is not implemented for this hash type");
}
```

So that the function always either returns a value or throws.

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_hash.cpp
```diff
@@ -0,0 +1,95 @@
...
+   const auto& p = m_sphincs_params;
+   BufferSlicer s(digest);
+   auto msg_hash = s.copy<SphincsHashedMessage>(p.fors_message_bytes());
+   auto tree_index_bytes = s.take(p.tree_digest_bytes());
+   auto leaf_index_bytes = s.take(p.leaf_digest_bytes());
+   BOTAN_ASSERT_NOMSG(s.empty());
+
+   return {std::move(msg_hash),
+           from_first_n_bits<XmssTreeIndexInLayer>(p.h() - p.xmss_tree_height(), tree_index_bytes),
+           from_first_n_bits<TreeNodeIndex>(p.xmss_tree_height(), leaf_index_bytes)};
```
randombit, line 92: It would be nice to assign names to the two values computed by `from_first_n_bits` which would make the `return` statement a bit easier to read imo.

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_parameters.cpp
```diff
@@ -0,0 +1,206 @@
...
+   switch(type) {
+      case Sphincs_Hash_Type::Sha256:
+         return "sha2";
+      case Sphincs_Hash_Type::Shake256:
+         return "shake";
+      case Sphincs_Hash_Type::Haraka:
+         return "haraka";
+   }
+
+   Botan::unreachable();
```
randombit, line 74: I'd rather just have a `default:` block that has an assertion failure, if we somehow don't have a match, rather than UB.

randombit: Or returning a `nullptr` or `std::optional<const char*>` if we want this to be `nothrow`

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_parameters.cpp
```diff
@@ -0,0 +1,206 @@
...
+         return "192s-r3.1";
+      case Sphincs_Parameter_Set::Sphincs192Fast:
+         return "192f-r3.1";
+      case Sphincs_Parameter_Set::Sphincs256Small:
+         return "256s-r3.1";
+      case Sphincs_Parameter_Set::Sphincs256Fast:
+         return "256f-r3.1";
+   }
+
+   Botan::unreachable();
```
randombit, line 93: As above wrt UB and `unreachable`

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_parameters.cpp
```diff
@@ -0,0 +1,206 @@
...
+}
+
+template <concepts::integral T>
+T ceil_tobytes(T bits) {
+   return (bits + 7) / 8;
+}
```
randombit, line 99: A quick grep for " + 7) / 8" shows we do this all over the place - can you move this function to a util header like `bit_ops.h` so we can make use of it elsewhere?

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_parameters.cpp
```diff
@@ -0,0 +1,206 @@
...
+      case Sphincs_Parameter_Set::Sphincs192Fast:
+         return Sphincs_Parameters(set, hash, 24, 66, 22, 8, 33, 16, 194);
+
+      case Sphincs_Parameter_Set::Sphincs256Small:
+         return Sphincs_Parameters(set, hash, 32, 64, 8, 14, 22, 16, 255);
+      case Sphincs_Parameter_Set::Sphincs256Fast:
+         return Sphincs_Parameters(set, hash, 32, 68, 17, 9, 35, 16, 255);
+   }
+
+   Botan::unreachable();
```
randombit, line 172: Again this should be an assert rather than UB

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_parameters.cpp
```diff
@@ -0,0 +1,206 @@
...
+   switch(m_hash_type) {
+      case Sphincs_Hash_Type::Sha256:
+         return "SHA-256";
+      case Sphincs_Hash_Type::Shake256:
+         return fmt("SHAKE-256({})", 8 * n());
+      case Sphincs_Hash_Type::Haraka:
+         return "Haraka";
+   }
+
+   Botan::unreachable();
```
randombit, line 189: Assert

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_parameters.h
```diff
@@ -0,0 +1,235 @@
...
+#include <botan/asn1_obj.h>
+
+#include <string_view>
+
+namespace Botan {
+
+enum class Sphincs_Hash_Type {
+   Shake256,
+   Sha256,
+   Haraka,
```
randombit, line 21: Add a comment here noting that Haraka is unsupported currently as a hint for someone trying to use it and confused why it always fails.

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_wots.cpp
```diff
@@ -0,0 +1,191 @@
+/*
+* WOTS+ - Winternitz One Time Signature+
+* (C) 2023 Jack Lloyd
+*     2023 Fabian Albert, René Meusel, Amos Treiber - Rohde & Schwarz Cybersecurity
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+**/
+
+#include <botan/internal/sp_hash.h>
+#include <botan/internal/sp_wots.h>
```
randombit, line 10: `sp_wots.h` should be the first include with a newline separating it from the rest of the includes

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.cpp
```diff
@@ -0,0 +1,340 @@
...
+   if(provider.empty() || provider == "base")
+      return std::make_unique<SphincsPlus_Verification_Operation>(m_public);
+   throw Provider_Not_Found(algo_name(), provider);
+}
+
+std::unique_ptr<PK_Ops::Verification> SphincsPlus_PublicKey::create_x509_verification_op(
+   const AlgorithmIdentifier& signature_algorithm, std::string_view provider) const {
+   if(provider.empty() || provider == "base") {
+      if(signature_algorithm != this->algorithm_identifier())
+         throw Decoding_Error("Unexpected AlgorithmIdentifier for SPHINCS+ signature");
```
randombit, line 184: This is missing braces, which will make `clang-tidy` unhappy. This won't be caught by CI until after #3571 is merged

randombit: In the meantime I'd suggest doing a full `clang-tidy` run and addressing any issues.

src/tests/data/pubkey/sphincsplus_fors.vec
```diff
@@ -0,0 +1,82 @@
...
...0cd0ac61867e348df041e9570614808601c9f43dbaa54b9bb75ce0f309e72f083863a541b31336a042e3d28fd86eda2d7c10f340355f62777bdee01ab8efe9cb340c1ce22eeee93eae324a94e148b45f00243bc1c2e1fcce1ac063c45a270e83be90f3e2d483bca5488b821fe32fcea4841e58c7969273c45ff47f1ad46ba8e791d40106f232959315ee0695e3d83d6fc18b763d37fce98f740bc5e8dcefaa511902227244d3f804814733c56a1763f6e8dd8e3669bf32918e3b7343d2a627e60bb93a455e63331a953b62307163d0d820775faf34ecfa016859d0cb51d379efc3f97145cfab7b5510af9d74ccd866771c9d651da1d4a53a6944ee3e779adee1b2b47311a46a1aeddf91f7ca690d56466871a43bd2897a0aede78876e29350ece1ae3525c23557b2870891e0197cf491ddde0ecfb47a12800710fe87896a23ef8c93dc72ccfed7d5db35157be46d39333a17c457f8bc11af25f9c2bab7f59dec653ea29f370eadd5cfab2e8c9c2f9c8308d9d61e5f0b35b14f7c9ddcb092a57446329f36f6606018a691cb810cdacb39c679ae256dfc5baf2fe7f052e52ccf8638330c0eca69acb447fba35c0b9ca8caaa8a14c8f5e16d7f5b7548ff5dfae726a05749baf6e1f00bbaefccda9f7311ce37c8d81f895c515bc238a2d1bc2158dd37fb699939930dd5343303dc45a5a508846a3d278edab1329b93bc72cada10e6493fe316a3e1f84494382dd1d49e8eb58f5ca291cb3d6c1e5e96d4680f61f4a51d8cc40f688f2cf9636038638b58e62b40bd19b3d9b9f303328cfcef40ce612b5007c3c85b74c3d6173ad8ec1a06d4311e14c68c2ef22b9d
```
randombit, line 14: Here again including only the hashes of the signatures would reduce the amount of test data by quite a log.

src/tests/test_sphincsplus.cpp
```diff
@@ -0,0 +1,143 @@
...
+ * Test all implemented SPHINCS+ instances using the data of the KAT files
+ * generated by the reference implementation (Round 3.1
+ *  (https://github.com/sphincs/sphincsplus/))
+ */
+class SPHINCS_Plus_Test final : public Text_Based_Test {
+   public:
+      SPHINCS_Plus_Test() : Text_Based_Test("pubkey/sphincsplus.vec", "SphincsParameterSet,seed,pk,sk,msg,HashSig") {}
+
+      bool skip_this_test(const std::string&, const VarMap& vars) override {
+         [[maybe_unused]] auto params = Botan::Sphincs_Parameters::create(vars.get_req_str("SphincsParameterSet"));
```
randombit, line 40: This should never really be unused because we know that at least one of SHA-2 or SHAKE is enabled, right?

src/tests/test_sphincsplus.cpp
```diff
@@ -0,0 +1,143 @@
...
+         // Depending on the SPHINCS+ configuration the resulting signature is
+         // hashed either with SHA-3 or SHA-256 to reduce the inner dependencies
+         // on other hash function modules.
+         auto hash = [&]() -> std::unique_ptr<Botan::HashFunction> {
+            if(params.hash_type() == Botan::Sphincs_Hash_Type::Shake256) {
+               return Botan::HashFunction::create("SHA-3(256)");
+            } else {
+               return Botan::HashFunction::create("SHA-256");
+            }
+         }();
```
randombit, line 77: It might be clearer if the lambda just returned the string, and then we call `Botan::HashFunction::create` on the result of the lambda.

src/tests/test_sphincsplus_fors.cpp
```diff
@@ -0,0 +1,96 @@
...
+         return Botan::Sphincs_Address(adrs);
+      }
+
+   public:
+      SPHINCS_Plus_FORS_Test() :
+            Text_Based_Test("pubkey/sphincsplus_fors.vec",
+                            "SphincsParameterSet,Address,SecretSeed,PublicSeed,PublicKey,Msg,Signature") {}
+
+      bool skip_this_test(const std::string&, const VarMap& vars) override {
+         [[maybe_unused]] auto params = Botan::Sphincs_Parameters::create(vars.get_req_str("SphincsParameterSet"));
```
randombit, line 43: Again, shouldn't ever really be unused since at least one of the two modes must be available.

src/tests/test_sphincsplus_wots.cpp
```diff
@@ -0,0 +1,117 @@
...
+         return Botan::Sphincs_Address(adrs);
+      }
+
+   public:
+      SPHINCS_Plus_WOTS_Test() :
+            Text_Based_Test("pubkey/sphincsplus_wots.vec",
+                            "SphincsParameterSet,Address,SecretSeed,PublicSeed,HashedWotsPk,WotsPk,Msg,WotsSig") {}
+
+      bool skip_this_test(const std::string&, const VarMap& vars) override {
+         [[maybe_unused]] auto params = Botan::Sphincs_Parameters::create(vars.get_req_str("SphincsParameterSet"));
```
randombit, line 43: As with previous comments, this annotation shouldn't be required

randombit (commented):
This looks very good, thanks. I left a bunch of comments, most should be easy to address.

randombit:
Please also add at least a brief entry in `pubkey.rst` describing what SPHINCS is, the version it was added in, etc.

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_hash.cpp
reneme: My two cents: If we shy away from `unreachable()` we should at least move the `Not_Implemented` outside and under the entire switch statement. Omitting the `default` case in enum-based switches enables useful compiler warnings when the enum is amended but the switch is forgotten.

reneme:
Regarding `Botan::unreachable()`:

I see your point with catching the (potentially) invalid enum values. If we make this a general rule -- which is probably a good idea in terms of defensive programming -- we should also fix it in Dilithium and Kyber where a similar `unreachable` idiom is used. And then we may remove `Botan::unreachable()` altogether, as it isn't used anywhere else at the moment.

src/tests/test_sphincsplus.cpp
FAlbertDev: The lines below are not `#if defined(...)` but `#if not defined(...)`. If both hash functions are enabled, `params` is unused.

src/tests/test_sphincsplus_fors.cpp
FAlbertDev: As with the previous one. If both hash functions are enabled, `params` is unused.

src/tests/test_sphincsplus_wots.cpp
FAlbertDev: Same case. If both hash functions are enabled, `params` is unused.

src/tests/test_sphincsplus.cpp
FAlbertDev: Done. Fixed in 3dec39d

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.cpp
FAlbertDev: Done!

FAlbertDev:
Thanks for your review @randombit! All suggestions should be implemented by now. Also, the branch is rebased to the current master. Let me know, if you have further suggestions. Else, I would squash the commits to finalize PR ready.

randombit:
Re `unreachable` - it is a tricky situation. It can certainly be helpful for optimization. I've certainly used the equivalent Rust macro `unreachable!` a few times. But if you think about if this function was instead named `execute_arbitrary_code_here`, the situations where you'd really want it, versus just taking the cost of an exception throw and avoiding UB, become slimmer and slimmer.

We may want to consider defining something that is `unreachable`-like but without the UB, eg:

```
[[noreturn]] void unexpected_codepath(const char* file, size_t line, ... other debug information maybe ...);

void unexpected_codepath(...) {
#if defined(BOTAN_TERMINATE_ON_ASSERTS)
   std::cerr << ... << '\n';
   std::abort();
#else
   throw Internal_Error(...);
#endif
}
```

src/tests/test_sphincsplus_wots.cpp
randombit: Ah right I see now, thanks

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_hash.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <memory>
+
+namespace Botan {
+
+Sphincs_Hash_Functions::Sphincs_Hash_Functions(const Sphincs_Parameters& sphincs_params,
+                                               const SphincsPublicSeed& pub_seed) :
+      m_sphincs_params(sphincs_params), m_pub_seed(pub_seed) {}
+
+std::unique_ptr<Sphincs_Hash_Functions> Sphincs_Hash_Functions::create(
+   const Sphincs_Parameters& sphincs_params, [[maybe_unused]] const SphincsPublicSeed& pub_seed) {
```
randombit, line 35: OK another ``[[maybe_unused]]`, and here I do think it must be used since it is used if either SHA2 or SHAKE is enabled.

randombit (approved):
One minor nit to check on, but otherwise LGTM!

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_hash.cpp
```diff
@@ -0,0 +1,97 @@
...
+         return std::make_unique<Sphincs_Hash_Functions_Shake>(sphincs_params, pub_seed);
+#else
+         throw Not_Implemented("SPHINCS+ with SHAKE is not available in this build");
+#endif
+
+      case Sphincs_Hash_Type::Haraka:
+         throw Not_Implemented("Haraka is not yet implemented");
+   }
+
+   throw Not_Implemented("Hash type is invalid");
```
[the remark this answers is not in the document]
FAlbertDev: Done. I like the new assert!

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_hash.cpp
FAlbertDev: Agreed, thanks! I fixed it.

FAlbertDev:
Thanks! I rebased the PR and applied all suggestions. Let me know, if you find anything else or if we are ready to squash and merge.

randombit:
@reneme Keep in mind that we are on time based release schedule, and barring unforeseen circumstances 3.1.0 will be going out the door on July 10th.

FAlbertDev (approved):
Good job @reneme, I like your new test cases. The only disadvantage is that the tests are now twice as slow.

src/tests/data/pubkey/sphincsplus_wots.vec
```diff
@@ -0,0 +1,15 @@
+# This test data was generated using customly written tests using the
+# reference implementation of SPHINCS+ Round 3.1 (https://github.com/sphincs/sphincsplus/)
+#
```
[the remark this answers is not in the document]
FAlbertDev: Yeah, I think this would be more consistent and single WOTS+ signatures are fast. Otherwise, we could drop the WOTS+ tests, since they are only Sphincs+ internal logic and the Sphincs+ tests test WOTS+ implicitly million fold. However, IMO additional (fast) tests never hurt ;)

reneme:
Running it [only for `Sphincs_Parameter_Set::Sphincs128Fast` now](https://github.com/randombit/botan/pull/3549/commits/eb38be7a8955d779dca568d6ca6495dfa90a3835). That should run the additional stuff once for each hash function.

FAlbertDev:
> Should we extend this WOTS+ unit test body with a few more test cases?

Done! I also changed the tests, so that the comparison is performed on hash values as in the FORS and Sphincs+ tests. I also added FORS tests for all instances (some SHA-2 instances were missing).

randombit:
> . Maybe we should run the additional tests on just one or two configurations.

Yeah 30 seconds is pretty terrible considering currently the entire main test suite takes < 20 seconds when run on multiple cores.

We could gate these on the result of `Test::run_long_tests`

reneme:
> We could gate these on the result of `Test::run_long_tests`

I'm always a bit hesitant to disable foundational tests with that, TBH. One more idea: we could run all the "slow" parameterizations only on`Test::run_long_tests()`. This would speed things up by quite a bit, I think. That would mean that half the implemented SPHINCS+ configurations would run for the special configs only.

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.h
```diff
@@ -0,0 +1,94 @@
...
+   public:
+      SphincsPlus_PrivateKey(std::span<const uint8_t> private_key, Sphincs_Parameter_Set type, Sphincs_Hash_Type hash);
+      SphincsPlus_PrivateKey(std::span<const uint8_t> private_key, Sphincs_Parameters params);
+      SphincsPlus_PrivateKey(const AlgorithmIdentifier& alg_id, std::span<const uint8_t> key_bits);
+      SphincsPlus_PrivateKey(RandomNumberGenerator& rng, Sphincs_Parameter_Set type, Sphincs_Hash_Type hash);
+      SphincsPlus_PrivateKey(RandomNumberGenerator& rng, Sphincs_Parameters params);
+
+      ~SphincsPlus_PrivateKey() override;
+
+      secure_vector<uint8_t> private_key_bits() const override;
```
reneme, line 79: We should override `raw_private_key_bits()` here (as an alias for `private_key_bits()`). The default implementation of that method would otherwise throw `Not_Implemented`.

FAlbertDev: Done! 077c5be

FAlbertDev:
The slow instances of SPHINCS+ are now only tested with `Test::run_long_tests()` (4756744). This reduces the test runtime to around 1 sec. The slow instances differ from the fast ones only in parameters; the logic stays the same. Therefore, the fast tests should be feasible for the default case.

FAlbertDev:
> Could you squash the latest changes?

Done! I think this PR is finally ready to go :)

Dezzj (approved):
Approved

Dezzj (approved):
Approved

PR #3550, opened 2023-05-17
Remove MDx_HashFunction

randombit:


src/lib/hash/mdx_hash/mdx_hash.h
```diff
@@ -1,62 +1,120 @@
...
+
+template<MD_Endian ENDIAN,
+         typename DIGEST_T,
+         size_t DIGEST_ELEM,
+         void init_fn(DIGEST_T[DIGEST_ELEM]),
+         void compress_fn(DIGEST_T[DIGEST_ELEM], const uint8_t[], size_t),
+         size_t BLOCK_BYTES = 64,
+         size_t DIGEST_LENGTH = DIGEST_ELEM * sizeof(DIGEST_T),
+         size_t CTR_BYTES = 8>
+class MD_Hash final
```
reneme, line 28: Let me suggest an alternative design using C++20 concepts and a CRTP-ish structure: Instead of this long list of parameters, we could write a concept to tie down the parameters each MDx-hash needs to define (as static constexpr members). Then, the instantiation of `MDx_Hash` in the concrete implementations shrinks to: `MDx_Hash<MD5> m_md;`:

```C++
template <typename T>
concept mdx_hash_implementation =
    requires(T::digest_type digest, uint8_t input[], size_t blocks, MD_Endian endian) {
        typename T::digest_type;
        T::endianness;
        T::block_bytes;
        T::ctr_bytes;
        T::init(digest);
        T::compress_n(digest, input, blocks);
    } &&
    T::block_bytes >= 64 && is_power_of_2(T::block_bytes) &&
    T::ctr_bytes >= 8 && is_power_of_2(T::ctr_bytes) &&
    T::ctr_bytes < T::block_bytes &&
    sizeof(typename T::digest_type) >= 16;
    

template <mdx_hash_implementation T>
class MDx
{
    std::array<uint8_t, T::block_bytes> m_buffer;
    T::digest_type m_digest;
    uint64_t m_count;
    size_t m_position;
};

class MD5
{
public:
    using digest_type = std::array<uint32_t, 4>;
    static constexpr MD_Endian endianness = MD_Endian::Little;
    static constexpr size_t block_bytes = 64;
    static constexpr size_t ctr_bytes = 8;
    
public:
    static void init(digest_type& digest);
    static void compress_n(digest_type& digest, const uint8_t input[], size_t blocks);

private:
    MDx<MD5> m_md;
};
```

That surely is a bit more boilerplate in the concrete implementations. But I see a few advantages:

* It lends itself from the usual STL members: e.g. `std::vector<>::value_type`
* Readability: Instead of a long list of magic numbers in the `MDx_Hash` instantiation, each parameter has a speaking name
* DRY: Parameter definitions can be referred to in the implementation classes (if needed) -> less implicit knowledge: e.g. `digest_type` encapsulates the `digest_length` and `digest_element_type` and the `init()`/`compress_n()` methods don't need to repeat it
* Parameter constraints are declared as a concept, hopefully resulting in better compiler errors

Certainly this can be further refined.

Godbolt: https://godbolt.org/z/1h1nzcYoh

src/lib/hash/md5/md5.cpp
```diff
@@ -63,81 +46,90 @@ inline void II(uint32_t& A, uint32_t B, uint32_t C, uint32_t D, uint32_t M)
...
+
+std::unique_ptr<HashFunction> MD5::new_object() const
+   {
+   return std::make_unique<MD5>();
+   }
+
+std::unique_ptr<HashFunction> MD5::copy_state() const
+   {
+   return std::make_unique<MD5>(*this);
    }
```
reneme, line 133: Now that each MDx hash inherits straight from `HashFunction`, these "plumbing methods" are repeated for each class. From what I can see, they all simply forward to their respective `m_md` or create a copy of themselves. Its just copy-and-paste boilerplate, but it does bug me somewhat.

With the full CRTP-approach [I suggested earlier](https://github.com/randombit/botan/pull/3525#discussion_r1184680854) this would go away. At the expense of inheriting from an `MDx` template.

reneme (commented):
Thanks for this example to help moving the Keccak discussion forward. I believe that makes your point clearer. To my mind the concept of `Keccak_c` and `MDx_Hash` is very similar.

Nevertheless, I'm still trying to wrap my head around why this inheritance was a problem. Is there a technical reason that I'm still missing?

For the sake of discussion, let me use this PR as a base for an alternative suggestion. Don't get me wrong: I'm fine with it either way. I just got hooked by the design space and would like to get to the bottom of it. 😇

src/lib/hash/sha2_64/sha2_64.h
```diff
@@ -52,47 +34,79 @@ class SHA_512 final : public MDx_HashFunction
...
+
+      void clear() override;
+   private:
+      void add_data(const uint8_t input[], size_t length) override;
+      void final_result(uint8_t output[]) override;
+
+      static void init(uint64_t digest[8]);
+
+      MD_Hash<MD_Endian::Big, uint64_t, 16,
+              SHA_384::init, SHA_512::compress_digest, 128, 48, 16> m_md;
```
reneme, line 82: Is that correct? The compress and init methods take `[8]` elements.

```suggestion
      MD_Hash<MD_Endian::Big, uint64_t, 8,
              SHA_384::init, SHA_512::compress_digest, 128, 48, 16> m_md;
```

Do I misunderstand something or is that a copy-and-paste error? (also for the SHA-flavours in this file)

reneme:
For reference as promised: https://github.com/randombit/botan/pull/3553. I've taken this patch as a starting point and moved common stuff into a CRTP base class. What do you think?

reneme:
This needs some consideration:

1. `clang-format` 😨
2. Interdependence with #3681
3. Consideration of `https://github.com/randombit/botan/pull/3550#discussion_r1198937199`

I'm willing to take this over. But would suggest to put it on hold until #3681 is merged.

reneme:
> I'm willing to take this over. But would suggest to put it on hold until https://github.com/randombit/botan/pull/3681 is merged

I opened a follow-up PR (#3705).

randombit:
Closing in favor of #3705

PR #3553, opened 2023-05-19
CRTP for MDx_Hash

reneme:
That's still somewhat sketchy, but I think it makes my point clear about having a CRTP-Adapter to move shared code (i.e. `HashFunction` interface implementation) into a central location. As I said: That's not a must at all. Albeit, I do believe it cleans things up quite nicely.


reneme:
> I fear I don't capture the whole design idea for MD_Hash from this sketch.

I tried capturing the design idea as a class diagram (click to enlarge).

image

PR #3558, opened 2023-05-29
Add braces around all if/else statements

randombit:


randombit:
@reneme Can you double check my change to `TLS::Client::Client`

PR #3563, opened 2023-05-30
Configure VSCode for clangd and clang-format

reneme:
Switch default formatter to `clang-format` and disable automatic addition of includes for `clangd`.

Personally, I'm using VS Code on Windows connected to an Ubuntu 22.04 (in WSL2). To use `clang-format` I needed to set the formatter executable to `clang-format-15` in the user-specific settings in VS Code. Obviously, that shouldn't be added to the shared configuration.

Re:clangd auto-include: This is annoying as it adds library-internal includes as relative paths using quotes. Disclaimer: I didn't look into taming this behavior but just disabled it for now.


src/editors/vscode/extensions.json
```diff
@@ -5,7 +5,7 @@
     "recommendations": [
         "ms-vscode.cpptools",
         "ms-python.python",
-        "chiehyu.vscode-astyle",
+        "xaver.clang-format",
```
lieser, line 8: What is the advantage of using this separate extension for formatting? Did not yet try it out, but clang-format support seems to be also part of the C/C++ extension.

reneme: I didn't get it to work properly. This extension kept using other formatters (apparently VS formatter or astyle) with some default formatting configuration. I didn't find a way to convince it to use clang-format and gave up eventually.

reneme: In the hope to make this more robust, I switched off the `C/C++`'s formatting feature explicitly.

lieser (approved):

PR #3565, opened 2023-05-30
Disable noisy/pointless pylint warnings

randombit:


src/configs/pylint.rc
```diff
@@ -37,7 +37,8 @@ confidence=
 #   not libraries
 # * fixme is marking TODOs in the future
 # * no-else-return is often awkward
-disable=missing-docstring,fixme,duplicate-code,no-else-return,consider-using-f-string,consider-using-with,broad-exception-raised
+# * too-many-x is just like, your opinion man
```
reneme, line 40: :rofl:

reneme (approved):
Thanks!

src/configs/pylint.rc
```diff
@@ -37,7 +37,8 @@ confidence=
...
+disable=missing-docstring,fixme,duplicate-code,no-else-return,consider-using-f-string,consider-using-with,broad-exception-raised,too-many-lines,too-many-locals,too-many-statements,too-many-branches,too-many-arguments,too-many-instance-attributes,broad-except,line-too-long,too-many-public-methods
```
lieser, line 41: Note that something like the following I think works too:
```
disable=
    missing-docstring, # not important because these are scripts not libraries
    fixme, # is marking TODOs in the future
    duplicate-code,
    no-else-return, # is often awkward
    consider-using-f-string,
    too-many-lines, # too-many-x is just like, your opinion man
```

I personally think the multi line style is easier to read, and it allows kind of inline comments for explaining why it is disabled.

But no strong opinion on this.

randombit: That syntax works fine, and is certainly much easier to read.

PR #3566, opened 2023-05-30
In CI use an explicit build dir distinct from current working directory

randombit:


reneme:
In case you're not aware: The out of source build option was added on user request and back then I set up the "Special" and "Analysis" build groups to build in dedicated directories (in `ci.yml`). The other groups build with the default so that both cases are exercised in CI.

randombit:
Thanks for mentioning. I don't know if this needs to be merged in any form - I'm trying to debug https://github.com/randombit/botan-rs/issues/82 which seems to be a Windows specific build failure that, iiuc, is a bug in the build system.

PR #3570, opened 2023-06-01
KMAC, 2nd: added keccak-fips as standalone; added KMAC-256

falko-strenzke:
This is a new PR for the addition of KMAC. This time, Keccak[c], the building block underlying all SHA-3 derived hash functions and also KMAC (build on  cSHAKE, currently not implemented in Botan) is added as a freestanding class that is used as a member in the SHA-3 derived functions.

This PR supersedes the previous PR for KMAC: https://github.com/randombit/botan/pull/3525


src/lib/hash/keccak/keccak.cpp
```diff
@@ -35,11 +35,12 @@ void Keccak_1600::clear() {
 }
 
 void Keccak_1600::add_data(const uint8_t input[], size_t length) {
-   m_S_pos = SHA_3::absorb(m_bitrate, m_S, m_S_pos, input, length);
+   m_S_pos = Keccak_FIPS::absorb(m_bitrate, m_S, m_S_pos, input, length);
 }
 
 void Keccak_1600::final_result(uint8_t output[]) {
-   SHA_3::finish(m_bitrate, m_S, m_S_pos, 0x01, 0x80);
+   //SHA3::finish(m_bitrate, m_S, m_S_pos, 0x01, 0x80);
```
reneme, line 42: Remove commented out code.

src/lib/hash/sha3/keccak_fips.h
```diff
@@ -0,0 +1,117 @@
...
+*
+*
+* [1] FIPS PUB 202 – FEDERAL INFORMATION PROCESSING STANDARDS PUBLICATION – SHA-3 Standard: Permutation-Based Hash and Extendable-Output Functions
+*       https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.202.pdf#page=28 
+* [2] https://csrc.nist.gov/projects/hash-functions/sha-3-project
+*/
+
+class Keccak_FIPS {
+   public:
+      virtual ~Keccak_FIPS();
```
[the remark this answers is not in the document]
reneme: In fact, the d'tor is probably not needed at all.

src/lib/hash/sha3/keccak_fips.cpp
```diff
@@ -0,0 +1,208 @@
...
+   {
+   zeroise(m_S);
+   m_S_pos = 0;
+   }
+
+
+void Keccak_FIPS::add_data(const uint8_t input[], size_t length)
+   {
+   m_S_pos = Keccak_FIPS::absorb(m_bitrate, m_S, m_S_pos, input, length);
+   }
```
reneme, line 187: Two things:

1. We don't need to stick to the `HashFunction` interface here. Let's consider to adopt the Keccak-nomenclature and call this `absorb` (just not static)
2. For new interfaces, let's use `std::span<>` consistently

```suggestion
void Keccak_FIPS::absorb(std::span<const uint8_t> input)
   {
   m_S_pos = Keccak_FIPS::absorb(m_bitrate, m_S, m_S_pos, input.data(), length.size());
   }
```

Usage of the interface will need adaption, of course.

src/lib/hash/sha3/keccak_fips.cpp
```diff
@@ -0,0 +1,208 @@
...
+   Keccak_FIPS::finish(m_bitrate, m_S, m_S_pos, m_custom_padd, m_custom_padd_bit_len);
+
+   /*
+   * We never have to run the permutation again because we only support
+   * limited output lengths
+   */
+   copy_out_vec_le(output, m_output_bits/8, m_S);
+
+   clear();
+   }
```
reneme, line 201: Same as with `add_data()`, I think we should consider using the Keccak interface instead of the `HashFunction` interface:

```suggestion
void Keccak_FIPS::finish(std::span<uint8_t> output)
   {
   Keccak_FIPS::finish(m_bitrate, m_S, m_S_pos, m_custom_padd, m_custom_padd_bit_len);

   BOTAN_ASSERT_NOMSG(output.size() >= m_output_bits/8);

   /*
   * We never have to run the permutation again because we only support
   * limited output lengths
   */
   copy_out_vec_le(output.data(), m_output_bits/8, m_S);
   }
```

I removed the final `clear()` under the assumption that we shouldn't force that on the use when calling this method. It's a low-level interface, users should know what they are doing.

reneme (commented):
Good direction, I think! Some remarks/questions regarding the basic design of `Keccak_FIPS`:

1. Is there a use-case for handling the basic data structures (`m_S`, `m_S_pos`, ...) outside of the class?
Maybe we should just fully encapsulate them and replace the low-level static method `absorb()`, `finish()`, `expand()`, `permute()` with ordinary members that handle the data structures internally. Exposing those data structures via getters and setters would probably give enough flexibility if needed, while hiding some of the Keccak complexity inside the class.
2. For new code, I'd love to see full adoption of `std::span<>` and `std::string_view` where applicable. Instead of C-style pointer-length-parameters or even raw pointer+hope kind of interfaces.
3. With `Keccak_FIPS` deliberately being designed as a component for composition rather than a base for inheritance, I think it should be a `final` class without any virtual methods.

Some more details are in inline comments.

src/lib/hash/sha3/keccak_fips.h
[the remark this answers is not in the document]
falko-strenzke: yes, this is a leftover from the previous version...
I remove the dtor in the next commit

src/lib/hash/keccak/keccak.cpp
falko-strenzke: Done in upcoming commit

src/lib/hash/sha3/keccak_fips.cpp
falko-strenzke: Done in upcoming commit.

src/lib/hash/sha3/keccak_fips.cpp
falko-strenzke: I adopted the proposed changes with the exception of removing the call to `clear()`. When removing that call, HMAC tests fail. Of course that could be resolved in the HMAC code, but currently I am not certain whether this will have other implications not necessarily showing up in the test results.

src/lib/hash/sha3/keccak_fips.h
```diff
@@ -0,0 +1,123 @@
...
+* [2] https://csrc.nist.gov/projects/hash-functions/sha-3-project
+*/
+
+class Keccak_FIPS final {
+   public:
+      /**
+      * @param output_bits the size of the hash output; must be one of
+      *                    224, 256, 384, or 512
+      */
+      explicit Keccak_FIPS(size_t output_bits, size_t capacity_bits, uint64_t custom_padd, uint8_t custom_padd_bit_len);
```
reneme, line 44: Would it make sense to default the `custom_padd` variables to `0`?

src/lib/hash/sha3/keccak_fips.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+Keccak_FIPS::Keccak_FIPS(size_t output_bits, size_t capacity, uint64_t custom_padd, uint8_t custom_padd_bit_len) :
+      m_output_bits(output_bits),
+      m_bitrate(1600 - capacity),
+      m_capacity(capacity),
+      m_custom_padd(custom_padd),
+      m_custom_padd_bit_len(custom_padd_bit_len),
+      m_S(25),  // 1600 bit
+      m_S_pos(0) {
+   // We only support the parameters for Keccak-FIPS in this constructor
```
reneme, line 136: Maybe `BOTAN_ASSERT_NOMSG` (or maybe `BOTAN_ARG_CHECK()`) that the passed in bitrates are divisible by 8. We're dividing them by 8 in some getters and should be certain that this will not lead to surprises.

src/lib/hash/sha3/keccak_fips.h
```diff
@@ -0,0 +1,123 @@
...
+
+      size_t output_bits() const { return m_output_bits; }
+
+      uint32_t bit_rate() const { return m_bitrate; }
+
+      std::string name() const;
+      void clear();
+      std::string provider() const;
+
+      secure_vector<uint64_t>& get_internal_state() { return m_S; }
```
reneme, line 58: I believe this is not exactly consistent, but for newly added getters we usually omit the `get_` prefix. Setters on the other hand come with a `set_` prefix. Also, please provide a const overload for the getter.

src/lib/hash/sha3/keccak_fips.h
```diff
@@ -0,0 +1,123 @@
...
+
+      void set_internal_pos(uint32_t pos) { m_S_pos = pos; }
+
+      uint32_t get_internal_pos() { return m_S_pos; }
```
reneme, line 62:
```suggestion
      uint32_t internal_pos() const { return m_S_pos; }
```

reneme (commented):
Thanks! A few more minor things. Also, I'd be really curious what @randombit has to say to this proposal. :)

reneme:
As a more generic remark: Would it make sense to pull the `Keccak_FIPS` into its own module? It might be a bit contrived, but some consumer might want to enable SHAKE but not SHA-3. That's currently not possible as the underlying `Keccak_FIPS` is (somewhat arbitrarily) a part of the sha3 module. [Jack suggested to put it into a new `permutations` subdir`](https://github.com/randombit/botan/pull/3525#issuecomment-1543797448).

src/lib/hash/sha3/keccak_fips.h
falko-strenzke: Hard to say, I think. Setting them to `0` means that the unofficial Keccak would be the default. That will matter in exactly a single class. I wouldn't really see when and where an implementer could really have any benefit from this.

src/lib/hash/sha3/keccak_fips.cpp
falko-strenzke: OK, will come in the next commit.

src/lib/hash/sha3/keccak_fips.h
falko-strenzke: Comes in next commit.

falko-strenzke:
> As a more generic remark: Would it make sense to pull the `Keccak_FIPS` into its own module? It might be a bit contrived, but some consumer might want to enable SHAKE but not SHA-3. That's currently not possible as the underlying `Keccak_FIPS` is (somewhat arbitrarily) a part of the sha3 module. [Jack suggested to put it into a new `permutations` subdir`](https://github.com/randombit/botan/pull/3525#issuecomment-1543797448).

Done as well.

src/lib/hash/sha3/keccak_fips.h
reneme: You're right. Doesn't really make a lot of sense.

reneme:
Could you squash your commits and rebase to master?  There are a few minor conflicts. If possible, maybe separate the SHA3/Keccak refactoring from the addition of KMAC in the commit history.

falko-strenzke:
I would like to inquire about the state  of this PR as we need KMAC integrated upstream in our project. If there any open issues remaining with this PR please let me know.

lieser:
@falko-strenzke I don't know about the state of the review, but as minimum for this to get merged the CI would need to be green.

src/lib/mac/kmac/kmac.cpp
```diff
@@ -0,0 +1,183 @@
+/*
+* KMAC
+* (C) 2023 Falko Strenzke
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include "botan/assert.h"
+#include "botan/exceptn.h"
+#include "botan/secmem.h"
```
FAlbertDev, line 10: This fixes the amalgamation CI builds:
```suggestion
#include <botan/assert.h>
#include <botan/exceptn.h>
#include <botan/secmem.h>
```

src/lib/permutations/keccak_fips/keccak_fips.cpp
```diff
@@ -0,0 +1,179 @@
...
+   return static_cast<uint32_t>(S_pos);
+}
+
+//static
+
+void Keccak_FIPS::finish(
+   size_t bitrate, secure_vector<uint64_t>& S, size_t S_pos, uint64_t custom_padd, uint8_t custom_padd_bit_len) {
+   BOTAN_ARG_CHECK(bitrate % 64 == 0, "Keccak-FIPS bitrate must be multiple of 64");
+   // append the first bit of the final padding after the custom padding
+   uint8_t init_pad = static_cast<uint8_t>(custom_padd | static_cast<uint64_t>(1 << custom_padd_bit_len));
```
reneme, line 98: Convince MSVC:

```suggestion
   uint8_t init_pad = static_cast<uint8_t>(custom_padd | uint64_t(1) << custom_padd_bit_len);
```

src/lib/permutations/keccak_fips/keccak_fips.h
```diff
@@ -0,0 +1,127 @@
...
+* (C) 2010,2016 Jack Lloyd
+* (C) 2023 Falko Strenzke
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#ifndef BOTAN_KECCAK_FIPS_H_
+#define BOTAN_KECCAK_FIPS_H_
+
+#include <botan/hash.h>
```
reneme, line 12: Please remove this superfluous include. It prevents a successful build with ` ./configure.py --minimized-build --enable-modules=keccak_fips`.

```suggestion
```

reneme (changes requested):
The requested changes below should fix the remaining CI issues.

reneme:
Once the CI is fixed, I would suggest the following:

1. Split this PR into two (for better reviewability):
  1.1. Just the Keccak related refactoring (#3671 sort-of depends on it)
  1.2. The addition of KMAC on top of the refactoring
2. Squash commits (individually in both proposed PRs)
3. Have another round of review and eventually bring this upstream

@falko-strenzke If you need any assistance, please don't hesitate to contact me on my work email or schedule a Teams meeting.

src/lib/permutations/keccak_fips/keccak_fips.h
falko-strenzke: Thanks for the corrections

src/lib/mac/kmac/kmac.cpp
falko-strenzke: Thanks for these corrections.

falko-strenzke:
> Once the CI is fixed, I would suggest the following:
> 
>     1. Split this PR into two (for better reviewability):
>        1.1. Just the Keccak related refactoring ([eXtendable Output Functions as first-class citizen #3671](https://github.com/randombit/botan/pull/3671) sort-of depends on it)
>        1.2. The addition of KMAC on top of the refactoring
> 
>     2. Squash commits (individually in both proposed PRs)
> 
>     3. Have another round of review and eventually bring this upstream
> 
> 
> @falko-strenzke If you need any assistance, please don't hesitate to contact me on my work email or schedule a Teams meeting.

I rewrote history now so that there are only two commits, one for SHA-3 refactoring and one for the addition to keccak.

So do I understand correctly, that the goal is to have both PRs on top of master:
```
master
  - PR 1 for SHA-3 refactoring
  - PR 2 for addition of KMAC
```
where PR 2 will compile only after rebasing onto master after the merge of PR 1.

reneme:
Thanks for rewriting the history. That helps a lot already.

Usually, I'd open two PRs on top of each other: I.e. (without rebasing any commits): One PR that contains 8f126dd0fe7b8b8615074797c91b66025e4fb510 and 1b089e175be27f86ed8aad329e79fba6092a66ed (the refactoring) and another one containing 8f126dd0fe7b8b8615074797c91b66025e4fb510, 1b089e175be27f86ed8aad329e79fba6092a66ed and 45bd4503f7fc4db112501a6919c8a9c5147053a4 (the refactoring + KMAC). In fact, the second proposed PR basically is this one already.

Typically, GitHub is clever enough to notice once the "smaller" PR is merged and adapts the diff view of the "bigger" PR accordingly. At least that's the workflow @randombit and me are typically using when splitting change sets like that.

falko-strenzke:
> Thanks for rewriting the history. That helps a lot already.
> 
> Usually, I'd open two PRs on top of each other: I.e. (without rebasing any commits): One PR that contains [8f126dd](https://github.com/randombit/botan/commit/8f126dd0fe7b8b8615074797c91b66025e4fb510) and [1b089e1](https://github.com/randombit/botan/commit/1b089e175be27f86ed8aad329e79fba6092a66ed) (the refactoring) and another one containing [8f126dd](https://github.com/randombit/botan/commit/8f126dd0fe7b8b8615074797c91b66025e4fb510), [1b089e1](https://github.com/randombit/botan/commit/1b089e175be27f86ed8aad329e79fba6092a66ed) and [45bd450](https://github.com/randombit/botan/commit/45bd4503f7fc4db112501a6919c8a9c5147053a4) (the refactoring + KMAC). In fact, the second proposed PR basically is this one already.
> 
> Typically, GitHub is clever enough to notice once the "smaller" PR is merged and adapts the diff view of the "bigger" PR accordingly. At least that's the workflow @randombit and me are typically using when splitting change sets like that.
> 
> One noteworthy downside: In case the reviewer of the "smaller" PR desires changes, it becomes a bit messy as the "bigger" PR must be rebased to the "smaller" one.

OK, I have now created  #3673.

reneme:
> OK, I have now created https://github.com/randombit/botan/pull/3673.

Thanks. I'll review (hopefully) tomorrow morning.

reneme:
Before going deeper into a review: [NIST SP.800-185](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-185.pdf) defines KMAC in terms of cSHAKE (customizable SHAKE) which is defined in terms of SHAKE. As I understand it, cSHAKE is basically SHAKE with incorporated domain separation byte strings.

You are intermingling the cSHAKE implementation with KMAC, which is technically fine. Though, I'm wondering whether we should split this up to ease future implementations of other functions defined in SP.800-185.

We could either extend Botan's SHAKE implementation to optionally take the domain separators `N` and `S`. They could both default to the empty string (resulting in the vanilla SHAKE function). Or we add another dedicated cSHAKE implementation that builds on the existing SHAKE and requires `N` and `S`. Personally, I'd go for the first option, as I believe the fairly simple domain separation approach in cSHAKE doesn't really justify the addition of another class.

What do you think? @falko-strenzke @randombit

falko-strenzke:
> Before going deeper into a review: [NIST SP.800-185](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-185.pdf) defines KMAC in terms of cSHAKE (customizable SHAKE) which is defined in terms of SHAKE. As I understand it, cSHAKE is basically SHAKE with incorporated domain separation byte strings.
> 
> You are intermingling the cSHAKE implementation with KMAC, which is technically fine. Though, I'm wondering whether we should split this up to ease future implementations of other functions defined in SP.800-185.
> 
> We could either extend Botan's SHAKE implementation to optionally take the domain separators `N` and `S`. They could both default to the empty string (resulting in the vanilla SHAKE function). Or we add another dedicated cSHAKE implementation that builds on the existing SHAKE and requires `N` and `S`. Personally, I'd go for the first option, as I believe the fairly simple domain separation approach in cSHAKE doesn't really justify the addition of another class.
> 
> What do you think? @falko-strenzke @randombit

Either option is fine for me. Given that apparently it is forbidden to use inheritance other than for mere interface inheritance, I would fear that a cSHAKE implementation would introduce a great deal of code duplication. If the SHAKE interface is extended as you propose, it should be well documented how to use it as cSHAKE.

reneme:
General remark: Consider adding KMAC to the `./botan speed` benchmark tool. See `src/cli/speed.cpp`.

reneme:
@falko-strenzke This now needs a rebase to `master` after #3673 was merged. It will reduce the patch displayed by GitHub to the relevant changes to introduce KMAC.

reneme:
> . Though, I'm wondering whether we should split this up to ease future implementations of other functions defined in SP.800-185.

In #3671 I now added a XOF implementation of cSHAKE-128/256. This required implementing some of the helper functions in SP.800-185 a bit more generically. See `src/lib/permutations/keccak_perm/keccak_helpers.h` and `src/lib/xof/shake_xof.h`.

Basically, `KMAC` would need to depend on the `shake_xof` module, include `botan/internal/shake_xof.h` and instantiate cSHAKE like so:

```C++
cSHAKE_256_XOF xof("KMAC");
xof.start(S, {});
xof.update(/* incorporate key using helpers in keccak_helpers.h */);
xof.update(msg)
xof.output(output_len);
```

@falko-strenzke If you're okay with it, I'm going to give the adaption of the KMAC implementation a try tomorrow.

reneme:
See: #3689

reneme:
Closing in favor of #3689.

PR #3573, opened 2023-06-02
Enable some additional clang-tidy warnings

randombit:
We have exceptions where we violate the check, but for those cases add NOLINT annotations. This at least helps prevent creating more instances of the problem in the future, and makes existing violations easy to find.

### DIFF
--- a/src/lib/ffi/ffi.cpp
+++ b/src/lib/ffi/ffi.cpp
@@ -17,6 +17,7 @@
 
 namespace Botan_FFI {
 
+// NOLINTNEXTLINE(*-avoid-non-const-global-variables)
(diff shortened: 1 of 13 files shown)


randombit:
@reneme It seems like the `Server_Hello_13` tag types can be const without issue but I wanted to check with you

reneme:
> It seems like the Server_Hello_13 tag structs can be const without issue but I wanted to check with you

Yes, for sure. Their only purpose is to steer constructor-overloading in a readable way.

reneme (approved):
LGTM. Should I look into fixing all the exceptions in the new TLS code as some follow up?

Issue #3578, opened 2023-06-05
XMSS private key does not update when signing X.509 things from the CLI tool


thewalker77:
I am not 100% sure I am doing things correctly, but assuming I am. XMSS keys are stateful and they should keep track of how many times they have been used, but the cli tool does not do that:

```
$ botan -V
2.19.1
$ botan keygen --algo=XMSS  > key.priv
$ shasum key.priv
e938f6.............ef key.priv
$ botan gen_self_signed key.priv --emsa=pssr random_file > cert
$ shasum key.priv
e938f6.............ef key.priv # does not change
```

I expected key.priv to update on every usage,otherwise it is security risk ?

reneme:
At first sight that sounds surprising as we test that [with a CLI test](https://github.com/randombit/botan/blob/f896922083ec913a2fe9a875c49158aa772b751b/src/scripts/test_cli.py#L409). Though, this does sign directly and not via an X.509 certificate. I'll keep you posted.

reneme:
Yes, this is reproducible. Thanks for reporting, I'll look into it.

reneme:
With #3579 merged this is fixed. Thanks again for reporting.

PR #3579, opened 2023-06-05
FIX: Missing update of XMSS private key in CLI's X.509 operations

reneme:
XMSS was not explicitly handled in `x509_obj.cpp` (`x509_signature_padding_for()`). Using XMSS private keys to sign certificates and CSR therefore resulted in: "Unknown X.509 signing key type: XMSS". One could conclude that we didn't want to allow XMSS for X.509 signing in the first place.
Anyway: By just passing an arbitrary EMSA to the CLI (like: `--emsa=lol`) this check could be circumvented and an XMSS signature was generated just fine. Though, update of the private key file was missed as reported in #3578.

This patch conditionally updates the private key file (as is done in `./botan sign`). Also, it adds XMSS to `x509_signature_padding_for()`, so `--emsa=` is not required (aka ignored) for XMSS.

IMO, this is a security issue and should be back-ported to the 2.x branch. Does it also justify a security advisory and/or CVE? @randombit

### DIFF
--- a/src/cli/x509.cpp
+++ b/src/cli/x509.cpp
@@ -26,7 +26,40 @@
       #include <botan/certstor_system.h>
    #endif
 
+   #include <fstream>
+
 namespace Botan_CLI {
+
+namespace {
+
+std::unique_ptr<Botan::Private_Key> load_private_key(const std::string& key_file, const std::string& passphrase) {
+   Botan::DataSource_Stream key_stream(key_file);
+   auto key = Botan::PKCS8::load_key(key_stream, passphrase);
+
+   if(!key) {
+      throw CLI_Error("Failed to load key from " + key_file);
+   }
+
+   return key;
+}
+
+void update_stateful_private_key(const Botan::Private_Key& key,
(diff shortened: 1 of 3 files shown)


reneme:
> If XMSS was not intended to sign X.509, what would be the expected usecase?

Apart from X.509 you could use it as a building block for your own protocols. Whenever you can establish trust in a public key without certification. E.g. to sign software updates for some proprietary gadget and bake the XMSS public key into your hardware.

Anyway, it appears that the missing XMSS handling in X.509 was merely an oversight.

mouse07410:
Probably the "largest" use case for XMSS works be firmware signing. IMHO, _definitely_ not X.509.

thewalker77:
Well, even for firmware signing, I would say that some standards use X.509 like DICE, and caliptra, x509 is  (to the best of my knowledge) just data exchange format

randombit:
> If XMSS was not intended to sign X.509, what would be the expected usecase ?

Keep in mind the limitation here is just in the command line utility

PR #3586, opened 2023-06-15
FIX: minor quirks in DL Scheme implementation

reneme:
This fixes a few typos and inconsistency that surfaced while reviewing #3210.

### DIFF
--- a/src/lib/pubkey/dsa/dsa.cpp
+++ b/src/lib/pubkey/dsa/dsa.cpp
@@ -61,11 +61,15 @@ DSA_PublicKey::DSA_PublicKey(const DL_Group& group, const BigInt& y) {
 }
 
 DSA_PrivateKey::DSA_PrivateKey(RandomNumberGenerator& rng, const DL_Group& group) {
+   BOTAN_ARG_CHECK(group.has_q(), "Q parameter must be set for DSA");
+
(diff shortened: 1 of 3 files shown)


src/lib/pubkey/dsa/dsa.cpp
```diff
@@ -61,18 +61,24 @@ DSA_PublicKey::DSA_PublicKey(const DL_Group& group, const BigInt& y) {
...
    m_private_key = std::make_shared<DL_PrivateKey>(group, x);
    m_public_key = m_private_key->public_key();
 }
 
 DSA_PrivateKey::DSA_PrivateKey(const AlgorithmIdentifier& alg_id, std::span<const uint8_t> key_bits) {
    m_private_key = std::make_shared<DL_PrivateKey>(alg_id, key_bits, DL_Group_Format::ANSI_X9_57);
    m_public_key = m_private_key->public_key();
+
+   BOTAN_ARG_CHECK(m_private_key->group().has_q(), "Q parameter must be set for DSA");
 }
```
reneme, line 82: The existence of a Q in the provided group is checked when instantiating a DSA public key. But from what I can see, it should be worthwhile to check it for the private key instantiation/generation as well.

randombit (approved):
Thanks

PR #3587, opened 2023-06-16
Segmentation Fault Protection is test_tls_rfc8448

FAlbertDev:
I stumbled on this while testing out some internal math stuff, where I've got a segfault in the respective test file. If the `Client_Context` in these tests cannot be created, the next checks run into a segfault. It seems that some checks were forgotten in these three lines (other places already have this check), so I added them quickly.


reneme (approved):
Yeah. We've been falling into this trap enough times. That's actually quite the drawback with the `Botan_Tests::CHECK` helper. It does somewhat imply (at least to me) that following invocations are cancelled when one of them fails. But that's not the case, resulting in those state mess ups.

Perhaps we could make this happen, but it would probably require us to wrap the `return {..}` into another helper.

PR #3588, opened 2023-06-16
Add assert_unreachable to API

FAlbertDev:
The CI workflow in https://github.com/randombit/botan/pull/3549 currently fails, since I used the new `BOTAN_ASSERT_UNREACHABLE` in the Sphincs+ tests. I think we could add this assert to the test/public API. I've chosen the `BOTAN_UNSTABLE_API` macro, but you may choose which one fits the best.


reneme (approved):
UNSTABLE makes enough sense to me.

randombit (approved):
:+1:

Issue #3585, opened 2023-06-13
Align OIDs of Kyber and Dilithium for IETF InteropTests


janklaussner:
Since end of last year several parties are conducting interoperability tests of the new PQC algorithms within the X.509 environment (see https://github.com/IETF-Hackathon/pqc-certificates).
Since there are no official OIDs yet, the participants agreed on a temporary mapping: https://github.com/IETF-Hackathon/pqc-certificates/blob/master/docs/oid_mapping.md
Botan needs the following patch for dilithium and kyber OIDs so that compatible certificates can be created or verified.
[align-oids-of-kyber-and-dilithium-for-ietf-interoptests.patch.txt](https://github.com/randombit/botan/files/11731610/align-oids-of-kyber-and-dilithium-for-ietf-interoptests.patch.txt)

Also, it would be good if the planned PQC algorithm implementations would also use this mapping (SPHINCS+, Falcon...)

randombit:
@reneme Your proposal seems generally fine. We would not want to break the OID mappings, but we already have a mechanism to accept multiple OIDs when decoding, with a single canonical OID used for encoding (this is used for instance for RSA keys which were assigned both 1.2.840.113549.1.1.1 and 2.5.8.1.1).

Only concern here is that we use the same OID also when encoding public keys and private keys. Currently our encoding is the simple, obvious, sane encoding (a single `OCTET STRING` containing the bytestring that is equal to how Kyber reference implementation encodes keys) but that may not match other implementations.

randombit:
@janklaussner Can you provide any reference or hints about how these other implementations encode public and private keys for these schemes?

janklaussner:
@randombit I have no closer insight into the impementations yet. At least Dilithium keys seemed to be compatible once the OIDs matched, I tested this. 
You can check samples if you look for the artifacts.zip files in the repo, e.g. https://github.com/IETF-Hackathon/pqc-certificates/blob/master/providers/bc/java-artifacts/artifacts.zip for Bouncycastle. There are certificate chains for all supported algorithms with encoded Public Keys and their private counterparts.
Does this help?

randombit:
@janklaussner Thanks - if the public keys were accepted that seems a good sign that everyone is converging on the same idea wrt encoding the keys, and with luck that will carry over into the private keys as well. I'll take a look at these artifacts, hopefully that gives some insights into where things are headed.

randombit:
[...]

PR #3589, opened 2023-06-19
Adopt (temporary) PQC OID mappings from IETF hackathon

reneme:
As suggested in #3585, this changes the (nonetheless temporary) OID mappings for Kyber, Dilithium and SPHINCS+ to the [mappings used in the IETF interoperability hackathon](https://github.com/IETF-Hackathon/pqc-certificates/blob/b5b1cdd95ce805e08756d7ba2539f9cca3816519/docs/oid_mapping.md).

For backward compatibility to Botan 3.0.0, we will be able to read the OIDs for Kyber/Dilithium that were previously defined in Botan's private region.

With SPHINCS+ [merged earlier today](https://github.com/randombit/botan/pull/3549) no backward compatibility will be needed here. However, the hackathon does not seem to define OIDs for SPHINCS+ with SHAKE (let alone Haraka). We therefore have no other choice than to keep those in Botan's private region for now.

 @janklaussner, @randombit Thoughts on that?


randombit:
@reneme Looks fine to me modulo the concerns about key encoding I raised in #3585. I have not had a chance to follow up on this aspect.

randombit:
I'm going to defer on this one until 3.2 - the worst possible outcome here is we switch the OIDs but then the other implementations are using an incompatible key encoding. That would leave us with a bigger problem than the current situation.

randombit:
@reneme Should we actually ship these OID mappings at all? I'm assuming at this point any standardization of formats will be for the final FIPS variants.

reneme:
Agreed. Let's close this.

PR #3592, opened 2023-06-22
Generate .clang-tidy files from run_clang_tidy.py

reneme:
This extends `run_clang_tidy.py` to render the rule-configuration into an inline `.clang-tidy` file (also added in this PR). Note that there is no top-level `.clang-tidy` file in the repository's root directory. I.e. no unnecessary clutter.

To regenerate this file, simply run:

```bash
 src/scripts/dev_tools/run_clang_tidy.py --regenerate-inline-config-file
```

`clangd` automagically picks up those config files and shows `clang-tidy` warnings inline while editing. This would certainly reduce the frequency of the nightly build scolding us for being dumb programmers. 🤡

image

Alternatively, one could perhaps manually manage those `.clang-tidy` files (e.g. using `InheritParentConfig` to build exclusive rules for non-lib directories) and let `run_clang_tidy.py` pick them up as well. Unfortunately, the `.clang-tidy` file is a YAML document that doesn't allow inline comments between or behind the globs. The list of rules is basically [a folded multi-line string](https://yaml-multiline.info/) that doesn't allow inline comments 😠.

Auto-generating the `.clang-tidy` files felt like the next best approach.


randombit:
Having the script be able to generate a yaml config for direct consumption seems fine and useful.

Having the multiple different `.clang-tidy` files seems really noisy though :/ tbh I would rather remove the `disabled_checks_non_lib` option and use a universal set of `clang-tidy` rules that cover both the core library + tests/etc so we just have the one file in the root dir.

reneme:
This is aiming to reflect the status quo. Certainly, I agree that unifying the checks for the entire code-base would simplify this a lot.

randombit:
Fair.

Over the weekend I'll look at clearing the exceptions currently in place for tests/cli, which will simplify this PR

src/scripts/dev_tools/run_clang_tidy.py
```diff
@@ -174,7 +204,7 @@ def file_matches(file, args):
             return True
     return False
 
-def main(args = None):
+def main(args = None): # pylint: disable=too-many-return-statements
```
randombit, line 207: TBH I think we should just disable that pylint, as already done with the other `too-many-...` warnings

randombit:
Oops - I just realized our current `clang-tidy` nightly run isn't covering the BoGo shim or the examples!

randombit:
WIP @ #3595

randombit:
With #3595 merged things become much simpler here.

As already discussed :slightly_smiling_face: I'm not a fan of random files in the root directory, but it is the lesser of evils compared to either 5 subtly different automatically generated configuration files, *or* people being unable to know during development that they're going to break the nightly checks, so fine to add a single generated `.clang-tidy` to the root dir.

reneme:
Technically, this particular file doesn't need to go into the repo root. `./src` is enough. FWIW.

Now that it is just one file, should we still auto-generate it? Or should we make it the ground truth and come up with some wonky workaround for the commenting issue?

lieser:
> Or should we make it the ground truth and come up with some wonky workaround for the commenting issue?

Note that I already tried and failed to add comments to a `.clang-tidy` file.

reneme:
> I'd lean to keeping the file being generated by the script, as you're doing here. That gives us options if in the future we want to enable or disable particular warnings based on eg the clang-tidy version (version X is buggy, useful lint Y is only available in clang-tidy 19, etc)

Meh! I interpreted your "thumbs up" as though you would favor the .clang-tidy as ground truth. 😐

The comments are still annoying, though. (I simply added them all at the top now, which is okay-ish but not optimal).

Obviously, no big deal to revert to the generator again.

src/.clang-tidy
```diff
@@ -0,0 +1,123 @@
...
+#   cppcoreguidelines-pro-type-vararg                          idiocy
+#   hicpp-vararg                                               idiocy
+#   hicpp-signed-bitwise                                       impossible to avoid in C/C++, int promotion rules :/
+#   modernize-loop-convert                                     sometimes very ugly
+#   modernize-raw-string-literal                               usually less readable
+#   modernize-use-trailing-return-type                         fine, but we're not using it everywhere
+#   modernize-return-braced-init-list                          thanks I hate it
+#   modernize-use-using                                        fine not great
+#   readability-identifier-length                              lol, lmao
+#   readability-redundant-access-specifiers                    reneme likes doing this
```
reneme, line 44: ❤️

randombit:
>  Meh! I interpreted your "thumbs up" as though you would favor the .clang-tidy as ground truth.

Ah, sorry. That was just in response to that it was (obviously, in retrospect) not needed for the config to be in the top level since all of the C++ files are under `src`

reneme:
This is now reverted to `run_clang_tidy.py --regenerate-inline-config-file`. It just generates one `.clang-tidy` file in `./src`.

randombit (approved):
:+1:

src/scripts/dev_tools/run_clang_tidy.py
randombit: I disabled this check on master

PR #3593, opened 2023-06-23
Chore: Make clang-tidy believe that StrongSpan<> really is cheap to copy

### LINKED ISSUE #3591: SPHNICS+ introduces some clang-tidy warnings
I fixed all of them on master except

```
Checked src/lib/pubkey/sphincsplus/sphincsplus_common/sp_treehash.cpp
/home/runner/work/botan/botan/src/lib/pubkey/sphincsplus/sphincsplus_common/sp_treehash.cpp:27:53: warning: the parameter 'out_auth_path' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param]
              StrongSpan<SphincsAuthenticationPath> out_auth_path,
                                                    ^
              const                                &
```

where I'm not sure if `clang-tidy`s suggestion is really correct. @reneme can you take a look at this?

reneme:
Closes #3591


randombit (approved):
Thank you

PR #3598, opened 2023-06-24
Update CI to XCode 14.3

randombit:


randombit:
This is fairly confusing - we are still detecting Xcode as `clang 14` but we have an explicit check for 14.3 which should map to `clang 15`.

randombit:
@reneme FYI - this may allow removing some of the `concepts.h` compatability hacks (I haven't investigate this)

reneme (approved):
Yikes.

Re:concepts.h: Apart from XCode, Android NDK was causing the same issues. Did you update that recently?

randombit:
No, the new NDK with a updated LLVM/libc++ is not expected until Q3 sometime. I plan to upgrade as soon as it's released.

Some of the concepts.h headers mention only XCode as a problem and not NDK (ex for `container_iterator`) possibly just an omission in the comments but I hoped this might move things forward slightly there. Not necessarily any reason to look into this now; might be easier to wait until NDK 26 is also out.

reneme:
> Not necessarily any reason to look into this now; might be easier to wait until NDK 26 is also out.

That's probably the safer bet. IIRC they both had similar issues. (Basically not implementing any C++20 std concepts.)

randombit:
Looking at https://xcodereleases.com/ it appears the XCode 15 schedule is closely tracking what they did for XCode 14. If that holds, XCode 15 will be released in September. We'll probably be able to move to XCode 15 + NDK 26 in time for 3.2.0, and at that point fix XCode 15 and NDK 26 as our minimum supported versions (rather than only supporting the latest version), since at that point we'll probably be held back by GCC 11 and not XCode/NDK.

PR #3599, opened 2023-06-24
Debugging XCode detection

randombit:


randombit:
What in the absolute fuckery is this? Apple Clang apparently always sets `__clang_minor__` to zero???????????????? regardless of the XCode version.

Which would work out fine if they didn't also upgrade the LLVM version within a major version .... wtf

randombit:
wtf apple

PR #3603, opened 2023-06-27
Fix some clang-tidy warnings in ARM specific code

randombit:


src/lib/block/aes/aes_armv8/aes_armv8.cpp
```diff
@@ -325,26 +353,30 @@ BOTAN_FUNC_ISA("+crypto") void AES_256::hw_aes_encrypt_n(const uint8_t in[], uin
    const uint8x16_t K13 = vld1q_u8(skey + 13 * 16);
    const uint8x16_t K14 = vld1q_u8(skey + 14 * 16);
 
+   using namespace AES_AARCH64;
+
+   using namespace AES_AARCH64;
```
reneme, line 358: Germans would say: "doppelt hält besser" -- twice holds stronger 🤡

randombit: :joy:

PR #3608, opened 2023-06-28
[TLS 1.3] Use a KEM interface in the Callbacks

reneme:
This refactors the TLS 1.3 asymmetric key establishment to use a KEM-like interface. I.e. `TLS::Callbacks` now has three new methods:

* `tls_kem_generate_key()`
* `tls_kem_encapsulate()`
* `tls_kem_decapsulate()`

Their default implementations simply use the existing `tls_generate_ephemeral_key()` and `tls_ephemeral_key_agreement()`. So "legacy" code based on Botan 3.0 won't break. With that, we gain the flexibility to implement the TLS 1.3 key establishment with PQ KEMs or hybrid approaches.

TLS 1.2 stays with the existing KEX-based callbacks mentioned above. We're not planning on contributing a KEM-based key establishment for TLS 1.2 anyway.

### Future Work

Frankly, the internal implementation of the `Key_Share` extension is a mess. I'll look into cleaning this up independently.


src/lib/tls/tls13/tls_extensions_key_share.cpp
```diff
@@ -139,21 +161,22 @@ class Key_Share_ServerHello {
...
+      const Key_Share_Entry& get_singleton_entry() const { return m_server_share; }
 
       std::vector<Named_Group> offered_groups() const { return {selected_group()}; }
 
       Named_Group selected_group() const { return m_server_share.group(); }
 
-      void erase() { m_server_share.erase(); }
+      secure_vector<uint8_t> take_shared_secret() {
+         BOTAN_STATE_CHECK(!m_shared_secret.empty());
+         return std::exchange(m_shared_secret, {});
```
randombit, line 174: It seems clang 11 does not know about `std::exchange` :disappointed:

src/lib/tls/tls13/tls_extensions_key_share.cpp
```diff
@@ -426,22 +449,27 @@ bool Key_Share::empty() const {
...
+                                                              Callbacks& cb,
+                                                              RandomNumberGenerator& rng) {
+   return std::unique_ptr<Key_Share>(new Key_Share(selected_group, client_keyshare, policy, cb, rng));
+}
+
+secure_vector<uint8_t> Key_Share::decapsulate(const Key_Share& server_keyshare,
+                                              const Policy& policy,
+                                              Callbacks& cb,
+                                              RandomNumberGenerator& rng) {
+   return std::visit(overloaded{[&](Key_Share_ClientHello& ch, const Key_Share_ServerHello& sh) {
```
randombit, line 464: I really hope pattern matching makes it in C++26 ...

src/lib/tls/tls_callbacks.h
```diff
@@ -255,6 +255,98 @@ class BOTAN_PUBLIC_API(2, 0) Callbacks {
                                       const std::vector<uint8_t>& msg,
                                       const std::vector<uint8_t>& sig);
 
+      struct Encapsulation_Result {
+            std::vector<uint8_t> encapsulated_bytes;
+            secure_vector<uint8_t> shared_secret;
+      };
```
randombit, line 261: I'd prefer this be a class with a constructor, private members, and accessors. Specifically then, given an `Encapsulation_Result`, one can know that we never eg modify the encapsulated bytes.

randombit (approved):
Looks good to me, thanks. Left a few minor comments. Fine to merge after addressing.

src/lib/tls/tls13/tls_extensions_key_share.cpp
reneme: Yeah! Though, as stated in the "future work" above: The internal implementation of `Key_Share` may be more complicated than it needs to.

src/lib/tls/tls13/tls_extensions_key_share.cpp
reneme: We're using it in other places, though. I rather think, an explicit `#include <utility>` should do the trick.

src/lib/tls/tls_callbacks.h
reneme: At first we just wanted to use a `std::pair` but went for this simple struct to be able to name the members of the returned tuple in `tls_kem_encapsulate`.

Though, thinking about it, we could add a `PK_KEM_Encryptor::encrypt()` override that also uses this struct instead of the currently used out-params. Having this as a "full blown" class might be more worthwhile.

Maybe as a follow-up?

randombit: Yeah having something like this be the return type from the basic KEM operations would be quite reasonable, the current KEM approach of having the user pass two references is a little confusing. Main concern would be this is public API and whatever we ships in 3.1.0 (July 10th) we're mostly stuck with. If you think you would have time to look at this before then, fine to merge as is

reneme: See here for a suggestion to generalize this result struct: https://github.com/randombit/botan/pull/3611

reneme:
Merging for better overview in #3609.

PR #3609, opened 2023-06-30
[TLS 1.3] Hybrid PQ/T key establishment

reneme:
### Pull Request Dependencies

* ~~#3608~~
* ~~#3616~~
* ~~#3611~~

### Things to do

* [x] Tests (at least a round-trip test -- e.g. as a new configuration in `test_cli.py cli_tls_socket_tests`)
* [x] Decide and introduce preliminary code points for compatibility with existing PQ/T endpoints -- e.g. cloudflare, Amazon KMS, Microsoft, IBM, ??)
  * How to handle support for those?
* [x] Try to build/test with hybrid KEX module disabled
* [x] Update the BoGo tests, that [seem to support Kyber since May '23](https://github.com/google/boringssl/commit/b811a6c6ab09b8294f2e2ed0b8080874a7d8bef4)
* [x] Act on https://github.com/randombit/botan/pull/3609#issuecomment-1628415582

### Description

This (finally) follows up on #2983 and is a more serious attempt in delivering hybrid key exchange based on [draft-ietf-tls-hybrid-design-06](https://datatracker.ietf.org/doc/html/draft-ietf-tls-hybrid-design).

Hybrid key exchange is handled by an adapter class `Hybrid_KEM_PrivateKey` that combines 2 or more KEX/KEM keys as specified in draft-ietf-tls-hybrid-design-06. To upstream users it provides a KEM interface.

Internally `Hybrid_KEM_PrivateKey` uses another adapter `KEX_to_KEM_Adapter_PrivateKey` that takes a single Key Agreement key (e.g. X25519) and exposes a KEM interface based on it. "KEM-Encaps" is implemented by generating a matching ephemeral key pair and using the ephemeral public key as the "encapsulated secret".

There are plenty of `TODO` where we needed to workaround short-comings of the `Public_Key` API. Potential improvements are collected in #3706.

IANA did not define code points for PQ and PQ/T key exchange groups in TLS, yet. As a result, different (beta) services define incompatible algorithm identifiers. We're providing a number of algorithm aliases to be compatible to both libOQS and Cloudflare's implementation. This is certainly a moving target that might change in the near future.

To give this a test run and connect to cloudflare.com using PQC (see [their blog post](https://blog.cloudflare.com/post-quantum-for-all/) -- section "What we deployed" -- for more details):

1. Create a TLS policy file (`pqc_tls.txt`):
```
allow_tls13 = true
allow_tls12 = false
x25519/Kyber-512-r3 secp256r1/Kyber-512-r3 x25519/Kyber-512-r3/cloudflare
```

2. Start the CLI like so:
```
./botan tls_client --tls-version=1.3 --debug cloudflare.com --policy=./pqc_tls.txt
```

You'll see that the exchanged Client Hello and Server Hello messages in the debug output are unusually large. The kyber public key weighs about 800bytes.

### DIFF
--- a/src/lib/pubkey/pubkey.cpp
+++ b/src/lib/pubkey/pubkey.cpp
@@ -135,7 +135,7 @@ secure_vector<uint8_t> PK_Decryptor_EME::do_decrypt(uint8_t& valid_mask, const u
 PK_KEM_Encryptor::PK_KEM_Encryptor(const Public_Key& key, std::string_view param, std::string_view provider) {
    m_op = key.create_kem_encryption_op(param, provider);
    if(!m_op) {
-      throw Invalid_Argument(fmt("Key type {} does not support KEM encryption"));
+      throw Invalid_Argument(fmt("Key type {} does not support KEM encryption", key.algo_name()));
    }
 }
 
--- a/src/lib/pubkey/pubkey.h
+++ b/src/lib/pubkey/pubkey.h
@@ -546,32 +546,32 @@ class BOTAN_PUBLIC_API(2, 0) PK_Decryptor_EME final : public PK_Decryptor {
 };
 
 /**
-   * Result of a key encapsulation operation.
-   */
+* Result of a key encapsulation operation.
+*/
 class KEM_Encapsulation final {
    public:
       KEM_Encapsulation(std::vector<uint8_t> encapsulated_shared_key, secure_vector<uint8_t> shared_key) :
             m_encapsulated_shared_key(std::move(encapsulated_shared_key)), m_shared_key(std::move(shared_key)) {}
 
       /**
-         * @returns the encapsulated shared secret (encrypted with the public key)
-         */
+      * @returns the encapsulated shared secret (encrypted with the public key)
+      */
       const std::vector<uint8_t>& encapsulated_shared_key() const { return m_encapsulated_shared_key; }
 
+      std::vector<uint8_t>& encapsulated_shared_key() { return m_encapsulated_shared_key; }
+
       /**
-         * @returns the plaintext shared secret
-         */
+      * @returns the plaintext shared secret
+      */
       const secure_vector<uint8_t>& shared_key() const { return m_shared_key; }
+
+      secure_vector<uint8_t>& shared_key() { return m_shared_key; }
 
    private:
       friend class PK_KEM_Encryptor;
 
       KEM_Encapsulation(size_t encapsulated_size, size_t shared_key_size) :
             m_encapsulated_shared_key(encapsulated_size), m_shared_key(shared_key_size) {}
-
-      std::vector<uint8_t>& encapsulated_shared_key() { return m_encapsulated_shared_key; }
-
-      secure_vector<uint8_t>& shared_key() { return m_shared_key; }
 
    private:
       std::vector<uint8_t> m_encapsulated_shared_key;
--- a/src/lib/pubkey/pubkey.h
+++ b/src/lib/pubkey/pubkey.h
@@ -15,6 +15,7 @@
 #include <span>
 #include <string>
 #include <string_view>
+#include <utility>
 
 namespace Botan {
 
@@ -558,14 +559,17 @@ class KEM_Encapsulation final {
       */
       const std::vector<uint8_t>& encapsulated_shared_key() const { return m_encapsulated_shared_key; }
 
-      std::vector<uint8_t>& encapsulated_shared_key() { return m_encapsulated_shared_key; }
-
       /**
       * @returns the plaintext shared secret
       */
       const secure_vector<uint8_t>& shared_key() const { return m_shared_key; }
 
-      secure_vector<uint8_t>& shared_key() { return m_shared_key; }
+      /**
+       * @returns the pair (encapsulated key, key) extracted from @p kem
+       */
+      static std::pair<std::vector<uint8_t>, secure_vector<uint8_t>> destructure(KEM_Encapsulation&& kem) {
+         return std::make_pair(std::exchange(kem.m_encapsulated_shared_key, {}), std::exchange(kem.m_shared_key, {}));
+      }
 
    private:
       friend class PK_KEM_Encryptor;
@@ -651,13 +655,11 @@ class BOTAN_PUBLIC_API(2, 0) PK_KEM_Encryptor final {
       KEM_Encapsulation encrypt(RandomNumberGenerator& rng,
                                 size_t desired_shared_key_len = 32,
                                 std::span<const uint8_t> salt = {}) {
-         KEM_Encapsulation result(encapsulated_key_length(), shared_key_length(desired_shared_key_len));
-         encrypt(std::span{result.encapsulated_shared_key()},
-                 std::span{result.shared_key()},
-                 rng,
-                 desired_shared_key_len,
-                 salt);
-         return result;
+         std::vector<uint8_t> encapsulated_shared_key(encapsulated_key_length());
+         secure_vector<uint8_t> shared_key(shared_key_length(desired_shared_key_len));
+
+         encrypt(std::span{encapsulated_shared_key}, std::span{shared_key}, rng, desired_shared_key_len, salt);
+         return KEM_Encapsulation(std::move(encapsulated_shared_key), std::move(shared_key));
       }
 
       /**
--- a/src/lib/tls/msg_client_hello.cpp
+++ b/src/lib/tls/msg_client_hello.cpp
@@ -423,7 +423,7 @@ void Client_Hello_12::add_tls12_supported_groups_extensions(const Policy& policy
    const std::vector<Group_Params> kex_groups = policy.key_exchange_groups();
    std::vector<Group_Params> compatible_kex_groups;
    std::copy_if(kex_groups.begin(), kex_groups.end(), std::back_inserter(compatible_kex_groups), [](const auto group) {
-      return is_ecdh(group) || is_dh(group) || is_x25519(group);
+      return !is_post_quantum(group);
    });
 
    auto supported_groups = std::make_unique<Supported_Groups>(std::move(compatible_kex_groups));
--- a/src/lib/tls/msg_client_hello.cpp
+++ b/src/lib/tls/msg_client_hello.cpp
@@ -414,6 +414,27 @@ std::vector<uint8_t> Hello_Request::serialize() const {
    return std::vector<uint8_t>();
 }
 
+void Client_Hello_12::add_tls12_supported_groups_extensions(const Policy& policy) {
+   // RFC 7919 3.
+   //    A client that offers a group MUST be able and willing to perform a DH
+   //    key exchange using that group.
+   //
+   // We don't support hybrid key exchange in TLS 1.2
+   const std::vector<Group_Params> kex_groups = policy.key_exchange_groups();
+   std::vector<Group_Params> compatible_kex_groups;
+   std::copy_if(kex_groups.begin(), kex_groups.end(), std::back_inserter(compatible_kex_groups), [](const auto group) {
+      return is_ecdh(group) || is_dh(group) || is_x25519(group);
+   });
+
+   auto supported_groups = std::make_unique<Supported_Groups>(std::move(compatible_kex_groups));
+
+   if(!supported_groups->ec_groups().empty()) {
+      m_data->extensions().add(new Supported_Point_Formats(policy.use_ecc_point_compression()));
+   }
+
+   m_data->extensions().add(std::move(supported_groups));
+}
+
 Client_Hello_12::Client_Hello_12(std::unique_ptr<Client_Hello_Internal> data) : Client_Hello(std::move(data)) {
    if(offered_suite(static_cast<uint16_t>(TLS_EMPTY_RENEGOTIATION_INFO_SCSV))) {
       if(Renegotiation_Extension* reneg = m_data->extensions().get<Renegotiation_Extension>()) {
@@ -478,11 +499,7 @@ Client_Hello_12::Client_Hello_12(Handshake_IO& io,
       m_data->extensions().add(new Certificate_Status_Request({}, {}));
    }
 
-   auto supported_groups = std::make_unique<Supported_Groups>(policy.key_exchange_groups());
-   if(!supported_groups->ec_groups().empty()) {
-      m_data->extensions().add(new Supported_Point_Formats(policy.use_ecc_point_compression()));
-   }
-   m_data->extensions().add(supported_groups.release());
+   add_tls12_supported_groups_extensions(policy);
 
    m_data->extensions().add(new Signature_Algorithms(policy.acceptable_signature_schemes()));
    if(auto cert_signing_prefs = policy.acceptable_certificate_signature_schemes()) {
@@ -561,13 +578,7 @@ Client_Hello_12::Client_Hello_12(Handshake_IO& io,
       m_data->extensions().add(new Certificate_Status_Request({}, {}));
    }
 
-   auto supported_groups = std::make_unique<Supported_Groups>(policy.key_exchange_groups());
-
-   if(!supported_groups->ec_groups().empty()) {
-      m_data->extensions().add(new Supported_Point_Formats(policy.use_ecc_point_compression()));
-   }
-
-   m_data->extensions().add(supported_groups.release());
+   add_tls12_supported_groups_extensions(policy);
 
    m_data->extensions().add(new Signature_Algorithms(policy.acceptable_signature_schemes()));
    if(auto cert_signing_prefs = policy.acceptable_certificate_signature_schemes()) {
--- a/src/lib/tls/tls13_pqc/hybrid_public_key.cpp
+++ b/src/lib/tls/tls13_pqc/hybrid_public_key.cpp
@@ -1,0 +1,370 @@
+/**
+* Composite key pair that exposes the Public/Private key API but combines
+* multiple key agreement schemes into a hybrid algorithm.
+*
+* (C) 2023 Jack Lloyd
+*     2023 Fabian Albert, René Meusel - Rohde & Schwarz Cybersecurity
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/internal/hybrid_public_key.h>
+
+#include <botan/pk_algs.h>
+
+#include <botan/internal/fmt.h>
+#include <botan/internal/kex_to_kem_adapter.h>
+#include <botan/internal/pk_ops_impl.h>
+#include <botan/internal/stl_util.h>
+
+namespace Botan::TLS {
+
+namespace {
+
+std::vector<std::pair<std::string, std::string>> algorithm_specs_for_group(Group_Params group) {
+   BOTAN_ARG_CHECK(is_hybrid(group), "Group is not hybrid");
+
+   switch(group) {
+      case Group_Params::HYBRID_X25519_KYBER_512_R3_OQS:
+      case Group_Params::HYBRID_X25519_KYBER_512_R3_CLOUDFLARE:
+         return {{"Curve25519", "Curve25519"}, {"Kyber", "Kyber-512-r3"}};
+      case Group_Params::HYBRID_X25519_KYBER_768_R3_OQS:
+      case Group_Params::HYBRID_X25519_KYBER_768_R3_CLOUDFLARE:
+         return {{"Curve25519", "Curve25519"}, {"Kyber", "Kyber-768-r3"}};
+
+      case Group_Params::HYBRID_SECP256R1_KYBER_512_R3_OQS:
+         return {{"ECDH", "secp256r1"}, {"Kyber", "Kyber-512-r3"}};
+      case Group_Params::HYBRID_SECP384R1_KYBER_768_R3_OQS:
+         return {{"ECDH", "secp384r1"}, {"Kyber", "Kyber-768-r3"}};
+      case Group_Params::HYBRID_SECP521R1_KYBER_1024_R3_OQS:
+         return {{"ECDH", "secp521r1"}, {"Kyber", "Kyber-1024-r3"}};
+
+      default:
+         return {};
+   }
+}
+
+std::vector<AlgorithmIdentifier> algorithm_identifiers_for_group(Group_Params group) {
+   BOTAN_ASSERT_NOMSG(is_hybrid(group));
+
+   const auto specs = algorithm_specs_for_group(group);
+   std::vector<AlgorithmIdentifier> result;
+   result.reserve(specs.size());
+
+   // This maps the string-based algorithm specs hard-coded above to OID-based
+   // AlgorithmIdentifier objects. The mapping is needed because
+   // load_public_key() depends on those while create_private_key() requires the
+   // strong-based spec.
+   //
+   // TODO: This is inconvenient, confusing and error-prone. Find a better way
+   //       to load arbitrary public keys.
+   for(const auto& spec : specs) {
+      result.push_back(AlgorithmIdentifier(spec.second, AlgorithmIdentifier::USE_EMPTY_PARAM));
+   }
+
+   return result;
+}
+
+std::vector<size_t> public_value_lengths_for_group(Group_Params group) {
+   BOTAN_ASSERT_NOMSG(is_hybrid(group));
+
+   // This duplicates information of the algorithm internals.
+   //
+   // TODO: Find a way to expose important algorithm constants globally
+   //       in the library, to avoid violating the DRY principle.
+   switch(group) {
+      case Group_Params::HYBRID_X25519_KYBER_512_R3_CLOUDFLARE:
+      case Group_Params::HYBRID_X25519_KYBER_512_R3_OQS:
+         return {32, 800};
+      case Group_Params::HYBRID_X25519_KYBER_768_R3_CLOUDFLARE:
+      case Group_Params::HYBRID_X25519_KYBER_768_R3_OQS:
+         return {32, 1184};
+
+      case Group_Params::HYBRID_SECP256R1_KYBER_512_R3_OQS:
+         return {32, 800};
+      case Group_Params::HYBRID_SECP384R1_KYBER_768_R3_OQS:
+         return {48, 1184};
+      case Group_Params::HYBRID_SECP521R1_KYBER_1024_R3_OQS:
+         return {66, 1568};
+
+      default:
+         return {};
+   }
+}
+
+}  // namespace
+
+std::unique_ptr<Hybrid_KEM_PublicKey> Hybrid_KEM_PublicKey::load_for_group(
+   Group_Params group, std::span<const uint8_t> concatenated_public_values) {
+   const auto public_value_lengths = public_value_lengths_for_group(group);
+   auto alg_ids = algorithm_identifiers_for_group(group);
+   BOTAN_ASSERT_NOMSG(public_value_lengths.size() == alg_ids.size());
+
+   const auto expected_public_values_length =
+      reduce(public_value_lengths, size_t(0), [](size_t acc, size_t len) { return acc + len; });
+   BOTAN_ARG_CHECK(expected_public_values_length == concatenated_public_values.size(),
+                   "Concatenated public values have an unexpected length");
+
+   BufferSlicer public_value_slicer(concatenated_public_values);
+   std::vector<std::unique_ptr<Public_Key>> pks;
+   for(size_t idx = 0; idx < alg_ids.size(); ++idx) {
+      pks.emplace_back(load_public_key(alg_ids[idx], public_value_slicer.take(public_value_lengths[idx])));
+   }
+   BOTAN_ASSERT_NOMSG(public_value_slicer.empty());
+   return std::make_unique<Hybrid_KEM_PublicKey>(std::move(pks));
+}
+
+Hybrid_KEM_PublicKey::Hybrid_KEM_PublicKey(std::vector<std::unique_ptr<Public_Key>> pks) {
+   BOTAN_ARG_CHECK(!pks.empty(), "List of public keys must not be empty");
+   BOTAN_ARG_CHECK(std::all_of(pks.begin(), pks.end(), [](const auto& pk) { return pk != nullptr; }),
+                   "List of public keys contains a nullptr");
+   BOTAN_ARG_CHECK(std::all_of(pks.begin(),
+                               pks.end(),
+                               [](const auto& pk) {
+                                  return pk->supports_operation(PublicKeyOperation::KeyEncapsulation) ||
+                                         pk->supports_operation(PublicKeyOperation::KeyAgreement);
+                               }),
+                   "Some provided public key is not compatible with this hybrid wrapper");
+
+   std::transform(
+      pks.begin(), pks.end(), std::back_inserter(m_public_keys), [](auto& key) -> std::unique_ptr<Public_Key> {
+         if(key->supports_operation(PublicKeyOperation::KeyAgreement) &&
+            !key->supports_operation(PublicKeyOperation::KeyEncapsulation)) {
+            return std::make_unique<KEX_to_KEM_Adapter_PublicKey>(std::move(key));
+         } else {
+            return std::move(key);
+         }
+      });
+
+   m_key_length =
+      reduce(m_public_keys, size_t(0), [](size_t kl, const auto& key) { return std::max(kl, key->key_length()); });
+   m_estimated_strength = reduce(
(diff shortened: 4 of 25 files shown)


src/lib/tls/tls13_pqc/info.txt
```diff
@@ -0,0 +1,20 @@
+<defines>
+TLS_13_PQC -> 20210721
```
reneme, line 2:
```suggestion
TLS_13_PQC -> 20230703
```

reneme:
We tested successfully against:

* **cloudflare.com** ([documentation](https://blog.cloudflare.com/post-quantum-for-all/))
  Unfortunately they use incompatible algo identifiers (hence the `x25519/Kyber-512-r3/cloudflare` hack in the policy). They use Curve25519 and don't support the NIST curves.
* **qsc.eu-de.kms.cloud.ibm.com** ([documentation](https://cloud.ibm.com/docs/key-protect?topic=key-protect-quantum-safe-cryptography-tls-introduction))
  IBM uses the algo identifiers [proposed by liboqs](https://github.com/open-quantum-safe/oqs-provider/blob/main/oqs-template/oqs-kem-info.md). It does not support Curve25519. Apart from the hybrid exchange this endpoint can also handle just `Kyber-512-r3`.
* **test.openquantumsafe.org** ([documentation](https://test.openquantumsafe.org/))
  Yeah. Everything should role.

We couldn't negotiate against:

* **kms.eu-central-1.amazonaws.com** ([documentation](https://docs.aws.amazon.com/kms/latest/developerguide/pqtls.html))
  This used to work with #2983. But the hybrid encoding proposed by IETF draft changed in the mean time. Apparently, Amazon's endpoint still speaks the old draft. (Individual public values were not just concatenated but a length-value-encoding was used.

src/lib/tls/tls13_pqc/info.txt
randombit: :+1:

reneme:
> Do we need to establish one or more new key_exchange_method in the policy?

I believe we should. See for example `Session_Summary::kex_algo()` which would currently report `UNDEFINED` for hybrid and/or KEM.

reneme:
Rebased to `master` after merging #3611.

src/lib/pubkey/pubkey.cpp
```diff
@@ -129,7 +129,7 @@ secure_vector<uint8_t> PK_Decryptor_EME::do_decrypt(uint8_t& valid_mask, const u
 PK_KEM_Encryptor::PK_KEM_Encryptor(const Public_Key& key, std::string_view param, std::string_view provider) {
    m_op = key.create_kem_encryption_op(param, provider);
    if(!m_op) {
-      throw Invalid_Argument(fmt("Key type {} does not support KEM encryption"));
+      throw Invalid_Argument(fmt("Key type {} does not support KEM encryption", key.algo_name()));
```
randombit, line 132: Eep :see_no_evil:

src/lib/pubkey/pubkey.h
```diff
@@ -540,33 +540,33 @@ class BOTAN_PUBLIC_API(2, 0) PK_Decryptor_EME final : public PK_Decryptor {
...
+      std::vector<uint8_t>& encapsulated_shared_key() { return m_encapsulated_shared_key; }
+
       /**
-         * @returns the plaintext shared secret
-         */
+      * @returns the plaintext shared secret
+      */
       const secure_vector<uint8_t>& shared_key() const { return m_shared_key; }
 
+      secure_vector<uint8_t>& shared_key() { return m_shared_key; }
```
randombit, line 562: Why would we want to allow this kind of mutation?

src/lib/tls/msg_client_hello.cpp
```diff
@@ -478,7 +478,19 @@ Client_Hello_12::Client_Hello_12(Handshake_IO& io,
...
+   // RFC 7919 3.
+   //    A client that offers a group MUST be able and willing to perform a DH
+   //    key exchange using that group.
+   //
+   // We don't support hybrid key exchange in TLS 1.2
+   const std::vector<Group_Params> kex_groups = policy.key_exchange_groups();
+   std::vector<Group_Params> compatible_kex_groups;
+   std::copy_if(kex_groups.begin(), kex_groups.end(), std::back_inserter(compatible_kex_groups), [](const auto group) {
+      return is_ecdh(group) || is_dh(group) || is_x25519(group);
+   });
```
randombit, line 490: I would suggest pulling these two common blocks into a utility function in this file, something like

`std::unique_ptr<Supported_Groups> make_tls12_supported_groups(const Policy& p)`

src/lib/pubkey/pubkey.h
randombit: This seems to be driven by wanting to avoid copying the key, but IMO it's not worth it given the potential for footguns in general purpose code.

If we really wanted to support this I'd propose something akin to Rust's `fn consume(self)`:

`static std::pair<std::vector<uint8_t>, secure_vector<uint8_t>> KEM_Encapsulation::destructure(KEM_Encapsulation kem)`

which internally moves the kem and then moves the values out of the kem.

src/lib/tls/tls13_pqc/info.txt
randombit: Date still should be updated

src/lib/tls/tls_algos.h
```diff
@@ -114,12 +134,28 @@ constexpr bool is_dh(const Group_Params group) {
           group == Group_Params::FFDHE_6144 || group == Group_Params::FFDHE_8192;
 }
 
-constexpr bool is_kem(const Group_Params) {
-   return false;  // no KEMs implemented, yet
+constexpr bool is_kyber(const Group_Params group) {
```
randombit, line 137: To avoid confusion with the hybrid groups, call this `is_pure_kyber`

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+namespace {
+
+template <typename RetT, typename KeyT, typename ReducerT>
+RetT reduce(const std::vector<KeyT>& keys, RetT acc, ReducerT reducer) {
+   for(const KeyT& key : keys) {
+      acc = reducer(std::move(acc), key);
+   }
+   return acc;
+}
```
randombit, line 30: Maybe throw this into `stl_util.h`?

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
```diff
@@ -0,0 +1,373 @@
...
+      case Group_Params::HYBRID_SECP521R1_KYBER_1024_R3_OQS:
+         return {{"ECDH", "secp521r1"}, {"Kyber", "Kyber-1024-r3"}};
+
+      default:
+         return {};
+   }
+}
+
+std::vector<AlgorithmIdentifier> algorithm_identifiers_for_group(Group_Params group) {
+   BOTAN_ASSERT_NOMSG(is_hybrid(group));
```
randombit, line 56: `BOTAN_ARG_CHECK`?

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
```diff
@@ -0,0 +1,373 @@
...
+         if(key->supports_operation(PublicKeyOperation::KeyAgreement)) {
+            return std::make_unique<KEX_to_KEM_Adapter_PublicKey>(std::move(key));
+         } else {
+            return std::move(key);
+         }
+      });
+}
+
+std::string Hybrid_KEM_PublicKey::algo_name() const {
+   std::string algo_name = "Hybrid(";
```
randombit, line 142: Use `std::ostringstream` here

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
```diff
@@ -0,0 +1,373 @@
...
+   for(size_t i = 0; i < m_public_keys.size(); ++i) {
+      algo_name += m_public_keys[i]->algo_name();
+      if(i < m_public_keys.size() - 1) {
```
randombit, line 145: Either way is fine so feel free to ignore this one, but I somewhat prefer

```
   for(size_t i = 0; i < m_public_keys.size(); ++i) {
      if(i > 0) {
         algo_name += ",";
      }
      algo_name += m_public_keys[i]->algo_name();
   }
```

src/lib/tls/tls13_pqc/hybrid_public_key.h
```diff
@@ -0,0 +1,107 @@
...
+namespace Botan::TLS {
+
+/**
+ * Composes a number of public keys as defined in this IETF draft:
+ * https://datatracker.ietf.org/doc/html/draft-ietf-tls-hybrid-design-04
+ *
+ * To an upstream user this composite key pair is presented as a KEM.
+ * Compositions of at least two (and potentially more) public keys are legal.
+ * Each individual key pair must either work as a KEX or as a KEM. Currently,
+ * the class can deal with ECC keys anc Kyber.
```
randombit, line 30: s/anc/and/

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
```diff
@@ -0,0 +1,373 @@
+/**
+* Composite key pair that exposes the Public/Private key API but combines
+* multiple key agreement schemes into a hybrid algorithm.
+*
+* (C) 2023 Jack Lloyd
+*     2023 Fabian Albert, René Meusel - Rohde & Schwarz Cybersecurity
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
```
randombit, line 10: General comment on this: it seems like the implementation is rather complicated by supporting an arbitrary number of keys, instead of exactly 2. The draft does allow it but it seems implausible to me that anyone would ever use it.

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
```diff
@@ -0,0 +1,373 @@
...
+   std::vector<std::unique_ptr<Public_Key>> pks;
+   for(size_t idx = 0; idx < alg_ids.size(); ++idx) {
+      pks.emplace_back(load_public_key(alg_ids[idx], public_value_slicer.take(public_value_lengths[idx])));
+   }
+   BOTAN_ASSERT(public_value_slicer.empty(), "Concatenated public value has the correct length");
+   return std::make_unique<Hybrid_KEM_PublicKey>(std::move(pks));
+}
+
+Hybrid_KEM_PublicKey::Hybrid_KEM_PublicKey(std::vector<std::unique_ptr<Public_Key>> pks) {
+   BOTAN_ARG_CHECK(!pks.empty(), "List of public keys must not be empty");
```
randombit, line 120: We should probably check that `pks.size() >= 2` or otherwise this isn't really a hybrid

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
```diff
@@ -0,0 +1,373 @@
...
+   return algo_name;
+}
+
+size_t Hybrid_KEM_PublicKey::estimated_strength() const {
+   return reduce(
+      m_public_keys, size_t(0), [](size_t es, const auto& key) { return std::max(es, key->estimated_strength()); });
+}
+
+size_t Hybrid_KEM_PublicKey::key_length() const {
+   return reduce(m_public_keys, size_t(0), [](size_t kl, const auto& key) { return kl + key->key_length(); });
```
randombit, line 159: These values could be precomputed in the constructor.

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
```diff
@@ -0,0 +1,373 @@
...
+   public:
+      Hybrid_KEM_Encryption_Operation(const Hybrid_KEM_PublicKey& key,
+                                      std::string_view kdf,
+                                      std::string_view provider) :
+            PK_Ops::KEM_Encryption_with_KDF(kdf) {
+         std::transform(
+            key.public_keys().begin(),
+            key.public_keys().end(),
+            std::back_inserter(m_kem_encryptors),
+            [&](const auto& pubkey) { return std::make_unique<PK_KEM_Encryptor>(*pubkey, "Raw", provider); });
```
randombit, line 216: I know above is considered more modern C++ style (I guess) but why not

```
for(const auto& k: key.public_keys()) {
  m_kem_encryptors.push_back(std::make_unique<PK_KEM_Encryptor>(*k, "Raw", provider));
}
```

Should have identical performance characteristics, and 3 fewer lines with less repetition.

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
```diff
@@ -0,0 +1,373 @@
...
+      size_t raw_kem_shared_key_length() const override {
+         return reduce(m_kem_encryptors, size_t(0), [](auto acc, const auto& kem_enc) {
+            return acc + kem_enc->shared_key_length(0 /* no KDF */);
+         });
+      }
+
+      size_t encapsulated_key_length() const override {
+         return reduce(m_kem_encryptors, size_t(0), [](auto acc, const auto& kem_enc) {
+            return acc + kem_enc->encapsulated_key_length();
+         });
```
randombit, line 228: These also could be precomputed in the constructor

```
for(const auto& k: key.public_keys()) {
  auto kem = std::make_unique<PK_KEM_Encryptor>(*k, "Raw", provider);
  m_raw_kem_shared_key_length += kem->shared_key_length(0 /* no KDF */);
  m_encapsulated_key_length += kem->encapsulated_key_length();
  m_kem_encryptors.push_back(std::move(kem));
}
```

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
```diff
@@ -0,0 +1,373 @@
...
+            kem_enc->encrypt(encaps_key_stuffer.next(kem_enc->encapsulated_key_length()),
+                             shared_key_stuffer.next(kem_enc->shared_key_length(0 /* no KDF */)),
+                             rng);
+         }
+      }
+
+   private:
+      // Note: PK_KEM_Encryptor can neither be moved nor copied. Hence, we wrap
+      //       it into a std::unique_ptr<> before storing it in a variant/vector.
+      std::vector<std::unique_ptr<PK_KEM_Encryptor>> m_kem_encryptors;
```
randombit, line 250: We should consider making these types movable. (Certainly not copyable in any circumstances given they contain a `unique_ptr` themselves)

 #3641

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
```diff
@@ -0,0 +1,373 @@
...
+   const auto specs = algorithm_specs_for_group(group);
+   std::vector<AlgorithmIdentifier> result;
+
+   // This maps the string-based algorithm specs hard-coded above to OID-based
+   // AlgorithmIdentifier objects. The mapping is needed because
+   // load_public_key() depends on those while create_private_key() requires the
+   // strong-based spec.
+   //
+   // TODO: This is inconvenient, confusing and error-prone. Find a better way
+   //       to load arbitrary public keys.
```
randombit, line 67: I agree this is awkward. Can you open a ticket about this?

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
```diff
@@ -0,0 +1,373 @@
...
+                               pks.end(),
+                               [](const auto& pk) {
+                                  return pk->supports_operation(PublicKeyOperation::KeyEncapsulation) ||
+                                         pk->supports_operation(PublicKeyOperation::KeyAgreement);
+                               }),
+                   "Some provided public key is not compatible with this hybrid wrapper");
+
+   std::transform(
+      pks.begin(), pks.end(), std::back_inserter(m_public_keys), [](auto& key) -> std::unique_ptr<Public_Key> {
+         if(key->supports_operation(PublicKeyOperation::KeyAgreement)) {
```
randombit, line 133: If a key supports both key agreement and key encapsulation, we will use `KEX_to_KEM_Adapter_PublicKey` - is that really the right choice?

randombit (commented):
This is only a partial review - I haven't reviewed the KEX->KEM adapter at all and only partially reviewed `hybrid_public_key.cpp`. I will finish the review asap.

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
reneme: Mhm, that's the conclusion we drew after finishing the general implementation. Perhaps even more simplification would be possible by supporting exactly one KEM and one KEX algorithm. Right now this class supports an arbitrary combination of 1..n such algorithms.

On the plus side: If one instantiates this as a wrapper around a single KEX, it acts as a KEM adapter. That may or may not be a useful bit.

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
reneme: @FAlbertDev will have a good laugh about this. 🤣 Yeah! I agree. Looking forward to when we can leverage ranges. (I believe macOS and NDK weren't yet there, when I last tried)

reneme:
Thanks for the review. Unfortunately, I won't be able to look into it before August 7th. Perhaps @FAlbertDev wants to give it a go, though. 😃

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+class Hybrid_KEM_Decryption final : public PK_Ops::KEM_Decryption_with_KDF {
+   public:
+      Hybrid_KEM_Decryption(const Hybrid_KEM_PrivateKey& key,
+                            RandomNumberGenerator& rng,
+                            const std::string_view kdf,
+                            const std::string_view provider) :
+            PK_Ops::KEM_Decryption_with_KDF(kdf) {
+         for(const auto& private_key : key.private_keys()) {
+            PK_KEM_Encryptor enc(*private_key, "Raw");
```
randombit, line 329: With #3653 it will not be required anymore to create a `PK_KEM_Encryptor` just to get the encapsulated length

randombit: #3653 is merged now so this can use `dec.encapsulated_key_length`

FAlbertDev: Very nice! I rebased the PR and used the new `encapsulated_key_length` methods in 0988317.

reneme:
Rebased to `master` to solve conflicts in `test_cli.py` after merging #3618.

reneme:
I accidentally over-force-pushed @FAlbertDev's commits. Fixed.
Could you please verify that I didn't forget anything?

FAlbertDev:
> Could you please verify that I didn't forget anything?

Looks correct!

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
reneme: For the record: I gave this one more iteration, removing the `unique_ptr` (now that [`PK_KEM_Decryptor` is movable](https://github.com/randombit/botan/pull/3655)) and simplified the encapsulated key length book-keeping (now that [we have `PK_KEM_Decryptor::encapsulated_key_length()`](https://github.com/randombit/botan/pull/3653))

reneme:
Rebased to latest master and worked on a few more nits. I'll do a thorough self-review and squash the history somewhat. Then, this should (finally) be ready from our side.

reneme:
From my perspective this is now ready for a final review. I did [successfully re-try handshakes with the services mentioned above](https://github.com/randombit/botan/pull/3609#issuecomment-1620039445).

There are still quite a few workarounds in the KEX-to-KEM adapter. Though, given that PQ/T hybrid TLS 1.3 is a moving target in itself, I'd suggest to address them at a later stage.

I'll also add a few remarks on how to use this in `tls.rst` this afternoon.

src/lib/tls/tls13_pqc/kex_to_kem_adapter.cpp
```diff
@@ -0,0 +1,266 @@
...
+ * TODO: Find a way to get this information without duplicating those
+ *       implementation details of the key agreement algorithms.
+ */
+size_t kex_shared_key_length(const Public_Key& kex_public_key) {
+   return std::visit(
+      overloaded{[](const ECDH_PublicKey& ecdh_public_key) { return ecdh_public_key.domain().get_p_bytes(); },
+                 [](const DH_PublicKey& dh_public_key) { return dh_public_key.group().p_bytes(); },
+                 [](const Curve25519_PublicKey&) { return size_t(32) /* TODO: magic number */; }},
+      as_kex_public_key(kex_public_key));
+}
```
reneme, line 99: It would be great if we could access certain basic information of specific algorithms statically.

randombit: Statically might be a bit tricky (given that DH/ECDH are parameterized by the group), but we easily could/should add a function on `PK_Key_Agreement_Key` which specifies what the output length is when you do a raw key agreement.

src/lib/tls/tls13_pqc/kex_to_kem_adapter.cpp
```diff
@@ -0,0 +1,266 @@
...
+   }
+
+   throw Not_Implemented(fmt("Cannot use unknown key agreement public key of type '{}' in the hybrid KEM key",
+                             kex_public_key.algo_name()));
+}
+
+/**
+ * This helper converts a key agreement public key into its raw public
+ * value. In contrast to the private key (cf. PK_Key_AgreementKey) there
+ * is no generic interface class to do this.
```
randombit, line 76: Oof. This PR is definitely pointing to some areas where the API needs to be improved.

(What you're doing in this PR for this seems fine in the short term, and no need to complicate it with other machinery to make it more generic - but this should be fixed.)

randombit (commented):
Looks very good, left a few comments, ptal

src/lib/tls/msg_client_hello.cpp
```diff
@@ -414,6 +414,27 @@ std::vector<uint8_t> Hello_Request::serialize() const {
...
+void Client_Hello_12::add_tls12_supported_groups_extensions(const Policy& policy) {
+   // RFC 7919 3.
+   //    A client that offers a group MUST be able and willing to perform a DH
+   //    key exchange using that group.
+   //
+   // We don't support hybrid key exchange in TLS 1.2
+   const std::vector<Group_Params> kex_groups = policy.key_exchange_groups();
+   std::vector<Group_Params> compatible_kex_groups;
+   std::copy_if(kex_groups.begin(), kex_groups.end(), std::back_inserter(compatible_kex_groups), [](const auto group) {
+      return is_ecdh(group) || is_dh(group) || is_x25519(group);
```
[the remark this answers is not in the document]
reneme: > Maybe we should have a `is_post_quantum` predicate?

Good idea! :+1:

> Also I really want to make `Group_Params` a Rust-style class-around-enum

Yes, please!

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
```diff
@@ -0,0 +1,370 @@
...
+            return std::move(key);
+         }
+      });
+}
+
+secure_vector<uint8_t> Hybrid_KEM_PrivateKey::private_key_bits() const {
+   throw Not_Implemented("Hybrid private keys cannot be serialized");
+}
+
+std::unique_ptr<Public_Key> Hybrid_KEM_PrivateKey::public_key() const {
```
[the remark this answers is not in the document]
reneme: Indeed. Thanks.

src/lib/tls/tls13_pqc/hybrid_public_key.h
```diff
@@ -0,0 +1,111 @@
...
+
+#ifndef BOTAN_TLS_13_HYBRID_KEM_PUBLIC_KEY_H_
+#define BOTAN_TLS_13_HYBRID_KEM_PUBLIC_KEY_H_
+
+#include <botan/pubkey.h>
+
+#include <botan/tls_algos.h>
+
+#include <memory>
+#include <vector>
```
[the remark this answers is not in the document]
reneme: They could. Though, actually, I got used to separating STL from botan includes like that on purpose. For better visual separation. 😅

Should I stop doing that?

src/lib/tls/tls13_pqc/hybrid_public_key.h
```diff
@@ -0,0 +1,111 @@
...
+
+namespace Botan::TLS {
+
+/**
+ * Composes a number of public keys as defined in this IETF draft:
+ * https://datatracker.ietf.org/doc/html/draft-ietf-tls-hybrid-design-04
+ *
+ * To an upstream user, this composite key pair is presented as a KEM.
+ * Compositions of at least two (and potentially more) public keys are legal.
```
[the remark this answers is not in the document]
reneme: I added a `BOTAN_ARG_CHECK()` to the constructors for "at least two keys" and extended the documentation a bit.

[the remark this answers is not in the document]
reneme: https://github.com/randombit/botan/pull/3609/commits/c59061b776d7d78074b9b30fb174804144b39e0c

src/lib/tls/tls13_pqc/kex_to_kem_adapter.cpp
reneme: I created a ticket to collect TODOs: https://github.com/randombit/botan/issues/3706
(writeup is still to be completed)

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
reneme: https://github.com/randombit/botan/issues/3706

reneme:
Done with your latest review comments. I updated the PR description (mainly for future reference) and created #3706 to collect future improvements on the Pubkey-APIs.

I committed the review fixes into new commits for easy review. Let me know if I should squash before we merge eventually.

reneme:
Rebased to master.

src/lib/tls/tls13_pqc/kex_to_kem_adapter.cpp
randombit: Thought I had left a comment on this earlier? Hm.

Statically might be difficult given how DH/ECDH are defined over arbitrary groups, but it seems pretty possible to have something like `virtual size_t raw_shared_key_length() = 0` the challenge is on what type to put it since `PK_Key_Agreement_Key` is only for the private keys ...

Conceivably we could have something on `Public_Key` like so

````
/**
* Return the length of the relevant artifact of this key.
* 
* Throws Not_Implemented if the key does not support an operation of this type.
*
* For KeyAgreement, returns the raw (non-KDF) output length.
*/
virtual size_t output_length(PublicKeyOperation op) = 0
```

src/lib/tls/tls13_pqc/info.txt
```diff
@@ -0,0 +1,20 @@
...
+<header:public>
+</header:public>
+
+<header:internal>
+hybrid_public_key.h
+kex_to_kem_adapter.h
+</header:internal>
+
+<requires>
+tls13
```
randombit, line 19: Should this also require `kyber`?  From a technical perspective maybe the code compiles without it, but you won't get something useful without it.

src/lib/tls/tls13_pqc/kex_to_kem_adapter.cpp
```diff
@@ -0,0 +1,266 @@
...
+             std::reference_wrapper<const ECDH_PublicKey>,
+             std::reference_wrapper<const Curve25519_PublicKey>>
+as_kex_public_key(const Public_Key& kex_public_key) {
+   BOTAN_ASSERT_NOMSG(kex_public_key.supports_operation(PublicKeyOperation::KeyAgreement));
+
+   try {
+#if defined(BOTAN_HAS_ECDH)
+      if(kex_public_key.algo_name() == "ECDH") {
+         return dynamic_cast<const ECDH_PublicKey&>(kex_public_key);
+      }
```
randombit, line 51: For these it seems better to do dynamic_cast on the pointer; otherwise if someone provides eg `PKCS11_ECDH_PublicKey` which also has an `algo_name` of "ECDH" we'll get a bad dynamic cast, which you do catch, but it's going to be a confusing error message.

Something like

```
if(const auto* ecdh = dynamic_cast<const ECDH_PublicKey*>(&kex_public_key)) {
  return *ecdh;
}
```

src/tests/test_tls_hybrid_kem_key.cpp
```diff
@@ -0,0 +1,258 @@
...
+   #include <botan/internal/hybrid_public_key.h>
+   #include <botan/internal/kex_to_kem_adapter.h>
+   #include <botan/internal/stl_util.h>
+
+namespace Botan_Tests {
+
+namespace {
+
+std::unique_ptr<Botan::Private_Key> kem() {
+   static auto kem_key = Botan::create_private_key("Kyber", Test::rng(), "Kyber-512-r3");
```
randombit, line 23: Why are these test keys `static`? I don't understand.

src/lib/tls/tls13_pqc/hybrid_public_key.h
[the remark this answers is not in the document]
randombit: I think it's fine as long as you follow the same ordering that `clang-format` would otherwise based on `IncludeCategories` configuration, eg first botan headers then std headers. (which you do here)

randombit (approved):
Did a full review, left a few minor comments. Approving to unblock since I didn't notice anything major. This certainly points out some nasty holes in our API surface; in principle it should be possible to implement the hybrid and kex->kem adaptors without referencing any algorithm specific details.

src/lib/tls/tls13_pqc/info.txt
reneme: I think it shouldn't. It's currently just an artifact of Kyber being the only available algorithm. Once #3679 is ready, we could also provide FrodoKEM-based hybrid instantiation. Consequently, we'd need to then depend on Kyber and Frodo, which doesn't really make sense to me.

src/tests/test_tls_hybrid_kem_key.cpp
reneme: The actual key isn't important for those tests. They just need "some" Kyber key. So this will generate one key and then re-use it to save some time. Now, generating a Kyber key is reasonably fast, so  we could avoid doing this. Or add a comment explaining the idea.

src/lib/tls/tls13_pqc/kex_to_kem_adapter.cpp
reneme: Fair, I'll fix it.

However, regarding: PKCS #11: That's an issue, no? Shouldn't we be able to use those here?

reneme: Another issue with `as_kex_public_key()`: It assumes that all algorithms are enabled via `--enable-modules=`. Otherwise the return type `std::variant<>` won't compile. Yikes.

reneme: In 512e0ffa9117e95a2090294759dc02c1a50b5eaf I got rid of `as_kex_public_key()` entirely, and duplicated the `#if defined(...) \n dynamic_cast<>()` machinery in all downstream helper methods. That way, one can disable (say) Curve25519 at compile time and the KEX-to-KEM adapter still works for the other algorithms.

That's messy. 😞 But its a provisional workaround until #3706 is being tackled.

src/lib/tls/tls13_pqc/info.txt
randombit: OK good point, fine to leave it like this.

src/lib/tls/tls13_pqc/kex_to_kem_adapter.cpp
randombit: Oof, right. How you did it seems fine given the current limitations of the PK interface.

src/tests/test_tls_hybrid_kem_key.cpp
randombit: Ah I see. I'd somewhat prefer non-static, and a new key is created for each invocation, unless there are performance difficulties as a result. But at the least there should be a comment, since otherwise the why-is-this-static is not so clear.

reneme: I tried removing the `static` and realized that the tests assume that these functions return the same key with every invocation. I aborted an attempt to refactor the test case to remove this assumption. It appeared that this would result in less clarity and/or added template complexity.

randombit: :+1:

PR #3616, opened 2023-07-05
Chore: Update BoGo tests

reneme:
Most notably, BoringSSL now disabled 3DES by default. This required some additional hacks to selectively enable it for some tests. Also, a "shim-id" was introduced that is used to identify individual connections. Apparently to save on used port numbers.
Oh, and it contains test for hybrid PQ/T exchanges that will come in handy for #3609.

Otherwise, no new issues were found. 😄


randombit (approved):
Thanks!

PR #3617, opened 2023-07-06
[TLS 1.3] Preparations for externally provided PSKs

reneme:
The `TLS::Ticket` class was introduced in Botan 3.0.0 (#3150). Given that externally provided PSKs will use the same structure, it would be great to generalize it a bit and rename it to `TLS::PskIdentity` (similar to its name in RFC 8446). This is breaking the public API of 3.0 somewhat (we added a deprecated typedef in case someone already started relying on it). @randombit Is that acceptable?

We doubt that we'll finish the actual implementation of the PSK support before 3.1, but it might be worthwhile to consider adding this already now.


randombit (approved):
:+1:

SemVer is a promise we'll do our best, and in a situation like this where a very new API that isn't quite what we want a break is fine. I would guess most people would be much happier with a tiny API break and a full TLS 1.3 implementation vs "true SemVer" and then waiting 3 more years for TLS PSK support.

PR #3618, opened 2023-07-06
[TLS 1.3] PSK Support

reneme:
### Pull Request Dependencies

* ~~#3617~~
* ~~#3622~~

### Description

This adds support for externally provisioned PSKs in TLS 1.3. Note that currently, only the `PSK_DHE_KE` mode is implemented. `PSK_KE` is left as future-work, same with 0-RTT/early data support.

External PSKs are provided by deriving the `Credentials_Manager` and overriding at least the `find_preshared_keys()` method. For fine-grained control, applications should additionally override `choose_preshared_key()`.

Note that the low-level key derivation machinery for PSK support was already implemented due to the existing TLS 1.3 resumption support. Most of this pull request is focussing on managing the offering and choosing of the PSK(s) during the early negotiation.

### Tests

Unfortunately, BoGo doesn't come with PSK tests for TLS 1.3 (yet?). We added an integration test in `test_cli.py` that performs a PSK-based handshake between a botan-cli client and server. Additionally we sucessfully confirmed compatibility with `openssl s_client` and `gnutls-cli` both in TLS 1.2 and 1.3.


src/lib/tls/tls13/tls_preshared_key.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include <botan/secmem.h>
+#include <botan/strong_type.h>
+
+#include <vector>
+
+namespace Botan::TLS {
+
+/// @brief holds a PSK identity as used in TLS 1.3
+using PresharedKeyID = Strong<std::vector<uint8_t>, struct PresharedKeyID_>;
```
reneme, line 20: Maybe use `std::string` for the external PSK identity? Typically, this is a human-readable label for a PSK. Maybe take inspiration from other TLS implementations.

src/lib/tls/credentials_manager.h
```diff
@@ -161,6 +161,17 @@ class BOTAN_PUBLIC_API(2, 0) Credentials_Manager {
...
       virtual SymmetricKey psk(const std::string& type, const std::string& context, const std::string& identity);
+
+      /**
+       TODO: describe me
+       */
+      virtual std::vector<TLS::PresharedKey> tls13_preshared_keys_for(std::string_view host);
+
+      /**
+       TODO: describe me
+       */
```
reneme, line 172: Make sure to mention that implementations need to cope with the fact that all identities provided to them might be resumption PSKs. I.e. the have to ignore _all_ members of the list and return `std::nullopt`. Same is true for `choose_session_to_resume()` (or similar name) in `Session_Manager`.

src/lib/tls/tls13/tls_preshared_key.h
```diff
@@ -0,0 +1,52 @@
...
+            m_identity(std::move(identity)), m_prf_algo(prf_algo), m_master_secret(std::move(psk)) {}
+
+      const PresharedKeyID& identity() const { return m_identity; }
+
+      secure_vector<uint8_t> extract_master_secret() {
+         BOTAN_STATE_CHECK(!m_master_secret.empty());
+         return std::exchange(m_master_secret, {});
+      }
+
+      const std::string& prf_algo() const { return m_prf_algo; }
```
reneme, line 42: Documentation?

src/lib/tls/tls_extensions.h
```diff
@@ -622,9 +633,11 @@ class BOTAN_UNSTABLE_API PSK final : public Extension {
        * @param session_to_resume  the session to be resumed; note that the
        *                           master secret will be taken away from the
        *                           session object.
+       * @param psks               a list of non-resumption PSKs that should be
+       *                           offered to the server
        * @param callbacks          the application's callbacks
        */
-      PSK(Session_with_Handle& session_to_resume, Callbacks& callbacks);
+      PSK(std::optional<Session_with_Handle>& session_to_resume, std::vector<PresharedKey> psks, Callbacks& callbacks);
```
reneme, line 640: Should `session_to_resume` be passed by value or maybe `const&`? Non-`const&` feels like ano out-param? 😕

src/lib/tls/tls_messages.h
```diff
@@ -223,7 +223,8 @@ class BOTAN_UNSTABLE_API Client_Hello_13 final : public Client_Hello {
                       RandomNumberGenerator& rng,
                       std::string_view hostname,
                       const std::vector<std::string>& next_protocols,
-                      std::optional<Session_with_Handle>& session);
+                      std::optional<Session_with_Handle>& session,
```
reneme, line 226: `const&` or mybe by-value?

src/lib/tls/tls_session.h
```diff
@@ -248,8 +248,29 @@ class BOTAN_PUBLIC_API(3, 0) Session_Summary : public Session_Base {
...
 
+      /**
+       * Indicates that the session was established using an externally provided
+       * PSK. Session resumptions in TLS 1.3 (while technically implemented
+       * using a PSK) are not considered here. @sa was_resumption()
+       *
+       * @returns true if the session was established using an externally
+       *          provided PSK.
+       */
+      bool psk_used() const { return m_external_psk_identity.has_value(); }
```
reneme, line 267: Note: With Botan 3.0 this reported `true` when a resumption was performed. This was technically correct, but semantically debatable at least. Now, `psk_used()` reports `true` for externally provided PSKs only.

reneme (commented):
Self-review. We'll address those issues (+ the currently failing CI) today. External PSK should then be finished from our side.

reneme:
Rebased after #3622 was merged. As mentioned in the PR description, most of this patch adds details to manage PSKs and let them co-exist with sessions during the early stages of the handshake. Along with the introduction of the public API `Channel::external_psk_identity()` this might be somewhat hard to follow. Should we spend the time and split this into two or three more focused commits?

src/lib/tls/tls13/tls_extensions_psk.cpp
```diff
@@ -92,32 +105,32 @@ class Client_PSK {
...
       uint16_t selected_identity() const { return m_selected_identity; }
 
-      Session take_session_to_resume() {
-         BOTAN_STATE_CHECK(m_session_to_resume.has_value());
-         Session s = std::move(m_session_to_resume.value());
-         m_session_to_resume = std::nullopt;
-         return s;
+      std::variant<std::monostate, Session, ExternalPSK> take_session_to_resume_or_psk() {
+         BOTAN_STATE_CHECK(!std::holds_alternative<std::monostate>(m_session_to_resume_or_psk));
+         return std::exchange(m_session_to_resume_or_psk, std::monostate());
```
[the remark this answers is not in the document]
reneme: I fully agree that `std::monostate` is a hack and shouldn't leak into the public interface. We use it because it is easier to use than the suggested `std::optional<std::variant<>>` combination. The best we can do in terms of pattern matching is the `overloaded{}` helper, and that becomes more compact using `std::monostate` instead of `std::optional<>`.

All that to say: This particular method is an internal class (`Server_PSK` is defined in an anonymous namespace). The public interface already decomposes this and returns `std::variant<Session, ExternalPSK>` as you suggest. See `PSK::take_session_to_resume_or_psk()` in line 360ish.

src/lib/tls/credentials_manager.cpp
```diff
@@ -9,6 +9,7 @@
 
 #include <botan/pkix_types.h>
 #include <botan/internal/fmt.h>
+#include <botan/internal/stl_util.h>
```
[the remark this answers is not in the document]
reneme: Probably a left-over from some intermediate prototyping.

src/lib/tls/credentials_manager.h
```diff
@@ -16,6 +16,7 @@
 #include <botan/tls_external_psk.h>
 #include <botan/tls_magic.h>
 #include <botan/x509cert.h>
+
```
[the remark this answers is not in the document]
reneme: Force of habit: trying to visually separate Botan and STL includes.

reneme:
> I'd like to see at least one C++ test since IIUC we are currently only testing this functionality with the cli, which seems not great.

Mhm, that bugged me as well. Especially, since we're just doing a round-trip test in CLI against our own implementation. If we break something along the way, chances are that we would stay compatible with ourselves but not with the outside world. 😨

Unfortunately, neither BoGo nor RFC8448 provide tests for externally provided PSK. And just implementing another round-trip test in C++ didn't seem like a big win to me.

Suggestion: Let's generate our own RFC8448-ish test vector and freeze it in a `Text_Based_Test`. This current state is manually cross-tested against OpenSSL and gnutls at least. I'll look into that.

reneme:
> Suggestion: Let's generate our own RFC8448-ish test vector and freeze it in a Text_Based_Test. This current state is manually cross-tested against OpenSSL and gnutls at least. I'll look into that.

Done. This worked surprisingly smoothly by hacking the CLI-based roundtrip test we had implemented already.

For future reference:

1. Fix the RNG seed to make things deterministic `./botan tls_server/tls_client --rng-type=drbg --drbg-seed=<some fixed seed>`
2. Configure the roundtrip test to what we need (mostly setting up appropriate policies for client and server)
3. Print the relevant information in the C++ TLS implementation
4. Spend some nerves on debugging and formatting it for `test_tls_rfc8448.cpp` and building a test flow for it

I've created [a temporary branch with the hacked bits](https://github.com/randombit/botan/tree/hack/rfc8448ish_transcript_from_tls_cli_test). This might come in handy if we want to repeat this exercise for `PSK_KE` (w/o ephemeral key).

src/lib/tls/tls13/tls_extensions_psk.cpp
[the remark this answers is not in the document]
randombit: I saw that this is already cleaned up in `PSK::take_session_to_resume_or_psk`, which is the important one - it just feels like we could avoid the possible assertion there if the `ServerPSK` return type was tighter.

> We use it because it is easier to use than the suggested std::optional<std::variant<>> combination. The best we can do in terms of pattern matching is the overloaded{} helper

Rust has really spoiled me in terms of this kind of thing being easy :/

Resolving because this is really not that big a deal.

randombit:
Thanks for adding the new tests @reneme

randombit (approved):

PR #3622, opened 2023-07-11
[TLS 1.3] Extend CredentialsManager for upcoming TLS 1.3 PSK

reneme:
This attempts to step away from the generic `Credentials_Manager::psk()` method that is parameterized with `type` and `context`. Instead, the class now has dedicated getter-methods for the specific keys it manages.

| Before | After |
| --- | --- |
| type:"tls-server", context:"session-ticket" | `Credentials_Manager::session_ticket_key()` |
| type:"tls-server", context:"dtls-cookie-secret" | `Credentials_Manager::dtls_cookie_secret()` |
| type"tls-server"/"tls-client", context:"my.host.name" | `Credentials_Manager::find_preshared_keys()` |

The TLS 1.2 implementation keeps using the `::psk()` method as before -- so existing applications that override it will continue to work as before. We think, dedicated methods for specific artefacts is much more on-par with the other TLS dependency classes (e.g. `Callbacks` and `Policy`). Additionally, its easier to understand the API like that.

New applications can make use of the added default implementation of `::psk()` that orchestrates the new dedicated methods as shown in the table above. With Botan 4.0 the `::psk()` method should probably disappear entirely. (And `Credentials_Manager` should perhaps move into the `TLS` namespace.)

The new methods `find_preshared_keys()` and `choose_preshared_key()` take into account that one can offer multiple PSKs in a TLS 1.3 Client Hello. The server then gets to choose which PSK it wants to negotiate with.

Given that TLS 1.2 pre-negotiates a singular PSK identity prior to actually requesting the actual PSK-value, it can benefit from the same `find_preshared_keys()` (through the legacy `psk()`). If exactly one PSK identity is requested, it should be guaranteed that at most one PSK is returned by `find_preshared_keys()`.

The actual TLS 1.3 PSK implementation will come in a follow-up.

### DIFF
--- a/src/lib/tls/credentials_manager.cpp
+++ b/src/lib/tls/credentials_manager.cpp
@@ -8,6 +8,7 @@
 #include <botan/credentials_manager.h>
 
 #include <botan/pkix_types.h>
+#include <botan/internal/fmt.h>
 
 namespace Botan {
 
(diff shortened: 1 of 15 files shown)


src/lib/tls/credentials_manager.cpp
```diff
@@ -21,10 +22,60 @@ std::string Credentials_Manager::psk_identity(const std::string& /*unused*/,
...
+         return SymmetricKey(std::move(key));
+      }
+   } else if(side == TLS::Connection_Side::Server && context == "dtls-cookie-secret") {
+      if(auto key = dtls_cookie_secret(); !key.empty()) {
+         return SymmetricKey(std::move(key));
+      }
+   } else /* context is a host name */ {
+      // Assuming that find_preshared_keys returns _exactly_ one or no keys when
+      // searching for a single specific identity.
+      if(auto psks = find_preshared_keys(context, side, {std::string(identity)}); psks.size() == 1) {
```
reneme, line 54:
```suggestion
      if(auto psks = find_preshared_keys(context, side, {identity}); psks.size() == 1) {
```

... artefact from trying to move to `std::string_view` method parameter types.

src/lib/tls/tls13/tls_psk_identity_13.h
```diff
@@ -40,16 +40,12 @@ class BOTAN_PUBLIC_API(3, 1) PskIdentity {
       /**
        * Construct from an externally provided PSK in the client
        */
-      PskIdentity(PresharedKeyID identity) :
-            m_identity(std::move(identity.get())),
-
-            // RFC 8446 4.2.11
-            //    For identities established externally, an obfuscated_ticket_age of
-            //    0 SHOULD be used, and servers MUST ignore the value.
-            m_obfuscated_age(0) {}
```
reneme, line 49: We moved away from the strong type here (`PresharedKeyID`). Though, I'm still somewhat on the fence whether or not to re-introduce it for clarity.

reneme (commented):
Self-review for tomorrow.

src/lib/tls/credentials_manager.cpp
```diff
@@ -21,10 +22,60 @@ std::string Credentials_Manager::psk_identity(const std::string& /*unused*/,
...
+      } else if(type == "tls-server") {
+         return TLS::Connection_Side::Server;
+      } else {
+         throw Internal_Error(fmt("No PSK set for type {}", type));
+      }
+   }();
+
+   // New applications should use the appropriate credentials methods. This is a
+   // retrofit of the behaviour before Botan 3.2.0 and will be removed in a
+   // future major release.
```
randombit, line 40: We should note this in `doc/deprecated.rst`

randombit (approved):
LGTM

PR #3624, opened 2023-07-12
Remove unnecessary variable in BLAKE2b

FAlbertDev:
This is only a small and non critical finding.
While looking at the BLAKE2b code for our Botan 3.1.0 Audit, I found this unused variable.

The variable m_F[1] is always zero and only used for an XOR operation. Therefore, I simply removed it.


randombit (approved):
Thanks. I was curious what the other bits of F might be for, but I don't find any indicators; https://datatracker.ietf.org/doc/html/rfc7693#section-3.2 seems to indicate this is just for the finalization bit.

We could probably simplify this a little by passing a boolean to `compress` to indicate if we are on the last block or not, in a manner very similar to how the RFC describes it. But this is already a nice improvement.

PR #3626, opened 2023-07-13
Add missing size limits checks for argon2 on 64-bit arch

lieser:
The limits are described in RFC 9106.
We also do some casting to uint32_t that would break.

### DIFF
--- a/src/lib/pbkdf/argon2/argon2.cpp
+++ b/src/lib/pbkdf/argon2/argon2.cpp
@@ -12,6 +12,7 @@
 #include <botan/internal/fmt.h>
 #include <botan/internal/loadstor.h>
 #include <botan/internal/rotate.h>
+#include <limits>
 
 #if defined(BOTAN_HAS_THREAD_UTILS)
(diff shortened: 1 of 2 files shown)


reneme (approved):

reneme:
Good catch! @randombit Perhaps get this into 3.1.1?

src/lib/pbkdf/argon2/argon2.cpp
```diff
@@ -21,6 +21,8 @@
    #include <botan/internal/cpuid.h>
 #endif
 
+#include <limits>
```
randombit, line 24: Include this in the main block of `#include` rather than after the conditionals

randombit (approved):
Thanks!

Sorry, didn't see this until after I pushed 3.1.1

src/lib/pbkdf/argon2/argon2.cpp
lieser: Fixed

PR #365, opened 2015-12-16
Replace deprecated Win API calls in Win32_EntropySource

neusdan:
* GetTickCount is replaced by GetTickCount64(): see https://msdn.microsoft.com/en-us/library/windows/desktop/ms724408(v=vs.85).aspx for details

* GlobalMemoryStatus is replaced by GlobalMemoryStatusEx: see https://msdn.microsoft.com/en-us/library/windows/desktop/aa366589(v=vs.85).aspx for details

PR #3628, opened 2023-07-15
thead pool debugging support proposal using native capabilities if po…

devnexen:
…ssible.

### DIFF
--- a/src/lib/utils/os_utils.cpp
+++ b/src/lib/utils/os_utils.cpp
@@ -16,16 +16,13 @@
 #include <chrono>
 #include <cstdlib>
 
-#if defined(BOTAN_TARGET_OS_HAS_THREADS)
-   #include <thread>
-#endif
-
 #if defined(BOTAN_TARGET_OS_HAS_EXPLICIT_BZERO)
    #include <string.h>
 #endif
 
 #if defined(BOTAN_TARGET_OS_HAS_POSIX1)
    #include <errno.h>
+   #include <pthread.h>
    #include <setjmp.h>
    #include <signal.h>
    #include <stdlib.h>
@@ -70,6 +67,10 @@ extern "C" char** environ;
 
 #if defined(BOTAN_TARGET_OS_HAS_PRCTL)
    #include <sys/prctl.h>
+#endif
+
+#if defined(BOTAN_TARGET_IS_FREEBSD) || defined(BOTAN_TARGET_IS_OPENBSD)
+   #include <pthread_np.h>
 #endif
 
 namespace Botan {
@@ -613,6 +614,25 @@ void OS::page_named(void* page, size_t size) {
    BOTAN_UNUSED(page, size);
 #endif
 }
+
+#if defined(BOTAN_TARGET_OS_HAS_THREADS)
+void OS::set_thread_name(std::thread& thread, const std::string& name) {
+   #if defined(BOTAN_TARGET_OS_IS_LINUX) || defined(BOTAN_TARGET_OS_IS_FREEBSD)
+   static_cast<void>(pthread_setname_np(thread.native_handle(), name.c_str()));
+   #elif defined(BOTAN_TARGET_OS_IS_OPENBSD)
+   static_cast<void>(pthread_set_name_np(thread.native_handle(), name.c_str()));
+   #elif defined(BOTAN_TARGET_OS_IS_NETBSD)
+   static_cast<void>(pthread_set_name_np(thread.native_handle(), "%s", const_cast<char*>(name.c_str())));
+   #elif defined(BOTAN_TARGET_OS_HAS_WIN32) && defined(BOTAN_BUILD_COMPILER_IS_MSVC)
+   // Using SetThreadDescription from Win10
+   BOTAN_UNUSED(thread, name);
+   #else
+   // TODO other possible oses ?
+   // macOs does not seem to allow to name threads other than the current one.
+   BOTAN_UNUSED(thread, name);
+   #endif
+}
+#endif
 
 #if defined(BOTAN_TARGET_OS_HAS_POSIX1) && !defined(BOTAN_TARGET_OS_IS_EMSCRIPTEN)
 
--- a/src/lib/utils/os_utils.h
+++ b/src/lib/utils/os_utils.h
@@ -13,6 +13,10 @@
 #include <string>
 #include <vector>
 
+#if defined(BOTAN_TARGET_OS_HAS_THREADS)
+   #include <thread>
+#endif
+
 namespace Botan::OS {
 
 /*
@@ -147,6 +151,10 @@ void page_allow_access(void* page);
 * Set a ID to a page's range expressed by size bytes
 */
 void page_named(void* page, size_t size);
+
+#if defined(BOTAN_TARGET_OS_HAS_THREADS)
+void set_thread_name(std::thread& thread, const std::string& name);
+#endif
 
 /**
 * Run a probe instruction to test for support for a CPU instruction.
--- a/src/lib/utils/thread_utils/thread_pool.cpp
+++ b/src/lib/utils/thread_utils/thread_pool.cpp
@@ -41,6 +41,8 @@ Thread_Pool& Thread_Pool::global_instance() {
 
 Thread_Pool::Thread_Pool(std::optional<size_t> opt_pool_size) {
    m_shutdown = false;
+   // On Linux, it is 16 length max, including terminator
+   const std::string tname = "Botan thread";
 
    if(!opt_pool_size.has_value()) {
       return;
@@ -65,8 +67,11 @@ Thread_Pool::Thread_Pool(std::optional<size_t> opt_pool_size) {
       }
    }
 
+   m_workers.resize(pool_size);
+
    for(size_t i = 0; i != pool_size; ++i) {
-      m_workers.push_back(std::thread(&Thread_Pool::worker_thread, this));
+      m_workers[i] = std::thread(&Thread_Pool::worker_thread, this);
+      OS::set_thread_name(m_workers[i], tname);
    }
 }
 


src/lib/utils/os_utils.cpp
```diff
@@ -614,6 +620,35 @@ void OS::page_named(void* page, size_t size) {
...
 }
 
+#if defined(BOTAN_TARGET_OS_HAS_THREADS)
+void OS::set_thread_name(std::thread& thread, const char* const name) {
+   int r;
+   auto hdl = thread.native_handle();
+   #if defined(BOTAN_TARGET_OS_IS_LINUX)
+   r = pthread_setname_np(hdl, name);
+   #elif defined(BOTAN_TARGET_OS_IS_FREEBSD)
+   r = pthread_set_name_np(hdl, name);
```
randombit, line 630: FreeBSD also seems to support `pthead_setname_np` (https://man.freebsd.org/cgi/man.cgi?query=pthread_set_name_np) which would reduce the number of conditionals.

src/lib/utils/os_utils.h
```diff
@@ -148,6 +152,10 @@ void page_allow_access(void* page);
 */
 void page_named(void* page, size_t size);
 
+#if defined(BOTAN_TARGET_OS_HAS_THREADS)
+void set_thread_name(std::thread& thread, const char* const name);
```
randombit, line 156: Is it required (by any OS) that this point to static storage? If so that should be documented here. Otherwise, `name` should be a `const std::string&` instead of a C style string.

(I would say a `std::string_view` but you can't convert that to a NULL terminated string)

src/lib/utils/os_utils.cpp
```diff
@@ -614,6 +620,35 @@ void OS::page_named(void* page, size_t size) {
...
+   #elif defined(BOTAN_TARGET_OS_IS_OPENBSD)
+   r = pthread_set_name_np(hdl, const_cast<char*>(name));
```
randombit, line 632: This cast doesn't seem to be necessary? (https://man.openbsd.org/pthread_set_name_np)

src/lib/utils/os_utils.cpp
```diff
@@ -614,6 +620,35 @@ void OS::page_named(void* page, size_t size) {
...
+   #elif defined(BOTAN_TARGET_OS_IS_NETBSD)
+   r = pthread_set_name_np(hdl, "%s", const_cast<char*>(name));
```
randombit, line 634: What a weird interface

src/lib/utils/os_utils.cpp
```diff
@@ -614,6 +620,35 @@ void OS::page_named(void* page, size_t size) {
...
+   #elif defined(BOTAN_TARGET_OS_HAS_WIN32) && defined(BOTAN_BUILD_COMPILER_IS_MSVC)
+   // This approach (as opposed to trigger an exception) is simpler
+   // also gives a chance to be in coredumps.
+   wchar_t val[16];
+   if(MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, name, -1, val, 16) > 0) {
+      r = SetThreadDescription(hdl, val);
```
randombit, line 640: IIUC https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setthreaddescription this function is only supported on Windows 10 or higher, which I think would be problematic

src/lib/utils/os_utils.cpp
```diff
@@ -614,6 +620,35 @@ void OS::page_named(void* page, size_t size) {
 #endif
 }
 
+#if defined(BOTAN_TARGET_OS_HAS_THREADS)
+void OS::set_thread_name(std::thread& thread, const char* const name) {
+   int r;
+   auto hdl = thread.native_handle();
+   #if defined(BOTAN_TARGET_OS_IS_LINUX)
+   r = pthread_setname_np(hdl, name);
```
randombit, line 628: Instead of assigning to `r` and then using `BOTAN_UNUSED` later on in the function just `static_cast<void>(pthread_set...)` within each block.

src/lib/utils/os_utils.cpp
randombit: In general the policy is that we don't support operating systems that are no longer supported by the OS vendor, but for Windows we are basically forced to be more lax than that simply due to the huge installed base of out of date systems.

I think practically speaking we are limited to using features that are available in the oldest version of Windows that is supported by the oldest version of VC++ we support. For VC that's VC2022 now. But I can't find anything definite about the oldest OS that VC2022 will allow you to target. I know we reverted #365 because people were still building XP-compatible binaries in 2016.

src/lib/utils/os_utils.cpp
```diff
@@ -614,6 +620,35 @@ void OS::page_named(void* page, size_t size) {
 #endif
 }
 
+#if defined(BOTAN_TARGET_OS_HAS_THREADS)
+void OS::set_thread_name(std::thread& thread, const char* const name) {
+   int r;
+   auto hdl = thread.native_handle();
```
randombit, line 626: `std::thread::native_handle` isn't necessarily guaranteed to exist, which would cause this to fail to compile on a (probably weird) OS that supported threads but didn't let you access native handles.

Instead call `thread.native_handle()` within each block which ensures we only call it if we have some idea what OS we are on.

src/lib/utils/os_utils.cpp
devnexen: If we consider windows the only exception in term of supporting EOL releases then yes.

src/lib/utils/os_utils.cpp
devnexen: fair point, you end up 99% of the time with a string anyway..

src/lib/utils/os_utils.cpp
devnexen: ah right I thought HAS_THREADS meant "all the way", fair enough then.

src/lib/utils/os_utils.cpp
```diff
@@ -614,6 +615,44 @@ void OS::page_named(void* page, size_t size) {
...
+#if defined(BOTAN_TARGET_OS_HAS_THREADS)
+void OS::set_thread_name(std::thread& thread, std::string& name) {
+   #if defined(BOTAN_TARGET_OS_IS_LINUX) || defined(BOTAN_TARGET_OS_IS_FREEBSD)
+   static_cast<void>(pthread_setname_np(thread.native_handle(), name.c_str()));
+   #elif defined(BOTAN_TARGET_OS_IS_OPENBSD)
+   static_cast<void>(pthread_set_name_np(thread.native_handle(), name.c_str()));
+   #elif defined(BOTAN_TARGET_OS_IS_NETBSD)
+   static_cast<void>(pthread_set_name_np(thread.native_handle(), "%s", const_cast<char*>(name.c_str())));
+   #elif defined(BOTAN_TARGET_OS_HAS_WIN32) && defined(BOTAN_BUILD_COMPILER_IS_MSVC)
+   const DWORD MS_VC_EXCEPTION = 0x406D1388;
```
randombit, line 627: This approach doesn't seem so useful since it's works if you are currently attached to a debugger.

If it's possible to check at compile time what the target OS is, I'd suggest calling the Windows 10 function if it is available and for older systems just do nothing. If that's not possible I'd say just remove this block entirely and leave a comment that we can use `SetThreadDescription` in the future.

In 33dee2f4f I deprecated support for older Windows but we'll probably want to wait at least a release or two before pulling the trigger on it.

src/lib/utils/os_utils.h
```diff
@@ -148,6 +152,10 @@ void page_allow_access(void* page);
 */
 void page_named(void* page, size_t size);
 
+#if defined(BOTAN_TARGET_OS_HAS_THREADS)
+void set_thread_name(std::thread& thread, std::string& name);
```
randombit, line 156: Ah I guess `name` cannot be a `string_view` since we need to get the null terminated value, so I think it should be a `const std::string&`

src/lib/utils/thread_utils/thread_pool.cpp
```diff
@@ -41,6 +41,8 @@ Thread_Pool& Thread_Pool::global_instance() {
 
 Thread_Pool::Thread_Pool(std::optional<size_t> opt_pool_size) {
    m_shutdown = false;
+   // On Linux, it is 16 length max, including terminator
+   std::string tname = "Botan thread";
```
randombit, line 45: This should be `const`

src/lib/utils/os_utils.cpp
```diff
@@ -614,6 +615,25 @@ void OS::page_named(void* page, size_t size) {
...
+#if defined(BOTAN_TARGET_OS_HAS_THREADS)
+void OS::set_thread_name(std::thread& thread, std::string& name) {
+   #if defined(BOTAN_TARGET_OS_IS_LINUX) || defined(BOTAN_TARGET_OS_IS_FREEBSD)
+   static_cast<void>(pthread_setname_np(thread.native_handle(), name.c_str()));
+   #elif defined(BOTAN_TARGET_OS_IS_OPENBSD)
+   static_cast<void>(pthread_set_name_np(thread.native_handle(), name.c_str()));
+   #elif defined(BOTAN_TARGET_OS_IS_NETBSD)
+   static_cast<void>(pthread_set_name_np(thread.native_handle(), "%s", const_cast<char*>(name.c_str())));
+   #elif defined(BOTAN_TARGET_OS_HAS_WIN32) && defined(BOTAN_BUILD_COMPILER_IS_MSVC)
+   // Using SetThreadDescription from Win10
```
randombit, line 627: There should be a `BOTAN_UNUSED` here as well - currently Windows CI is failing because the parameters are not used

src/lib/utils/os_utils.cpp
```diff
@@ -614,6 +615,25 @@ void OS::page_named(void* page, size_t size) {
...
+   #else
+   // TODO other possible oses ?
+   // macOs does not seem to allow to name threads other than the current one.
+   BOTAN_UNUSED(thread);
+   BOTAN_UNUSED(name);
```
randombit, line 632: You can pass multiple args to a single invocation of `BOTAN_UNUSED` - `BOTAN_UNUSED(thread, name)`

randombit (commented):
Thanks! This is looking very good now. I left a few comments.

randombit (approved):
Looks great. Thanks for addressing the review comments.

PR #3634, opened 2023-07-20
Fix: Segfault in tls_http_server

FAlbertDev:
I stumbled on this one while playing around with [tlsfuzzer](https://github.com/tlsfuzzer/tlsfuzzer) on Botan's `tls_http_server` CLI. Somehow the `TLS_Asio_HTTP_Session::stop()` method can be called multiple times for the same object, causing a segmentation fault when trying to call `m_tls->close()` on a nullptr `m_tls`. 

I don't know much about ASIO stuff, so I simply added some guarding _if_ statements, which should fix the segfault for now. I also added some guards at other positions, just to be sure that `m_tls` is never accessed after close.

### DIFF
--- a/src/cli/tls_http_server.cpp
+++ b/src/cli/tls_http_server.cpp
@@ -204,6 +204,11 @@ class TLS_Asio_HTTP_Session final : public std::enable_shared_from_this<TLS_Asio
       }
 
       void stop() {
+         if(!m_tls) {
+            // Server is already closed
+            return;
+         }
+
          m_tls->close();
 
          // Need to explicitly destroy the TLS::Server object to break the
@@ -287,6 +292,10 @@ class TLS_Asio_HTTP_Session final : public std::enable_shared_from_this<TLS_Asio
       }
 
       void handle_http_request(const HTTP_Parser::Request& request) override {
+         if(!m_tls) {
+            log_error("Received client data after close");
+            return;
+         }
          std::ostringstream response;
          if(request.verb() == "GET") {
             if(request.location() == "/" || request.location() == "/status") {
@@ -382,6 +391,10 @@ class TLS_Asio_HTTP_Session final : public std::enable_shared_from_this<TLS_Asio
       }
 
       void tls_alert(Botan::TLS::Alert alert) override {
+         if(!m_tls) {
+            log_error("Received client data after close");
+            return;
+         }
          if(alert.type() == Botan::TLS::Alert::CloseNotify) {
             m_tls->close();
          } else {


randombit:
This change as it is seems fine, but it does raise further questions

* Is there something systematic we should be changing with our asio command line utils?
* Is there something off about the TLS state machine itself that is leading us into a bad state?

reneme (approved):
I'm quite sure that the root cause for this particular crash is not running deeper than what @FAlbertDev fixed in this patch. Namely, that `::stop()` can be called more than once and its implementation didn't account for that. It is easy to reproduce by sending a valid Client Hello and then immediately closing the connection on the client side. See the script at the end.

Here's what happens:

1. The HTTP server waits for clients to connect
2. A Client Hello is received and pushed into the TLS implementation
   (`TLS_Asio_HTTP_Session::client_read()`)
3. An asynchronous write containing the Server Hello response is scheduled
   (`TLS_Asio_HTTP_Session::tls_emit_data()`)
4. Control flow goes back to the TLS implementation and eventually to `TLS_Asio_HTTP_Session::client_read()` (without an error)
5. An asynchronous read is started (in `::client_read()`) to wait for more data from the client
6. The asynchronous write fails (because the client disconnected in the mean time) and `::handle_client_write_completion()` calls `::stop()` for the first time.
7. The asynchronous read fails (because the server system closes the broken socket) and calls `::stop()` for the second time
8. Ouch!

Note that 6. and 7. might appear in the opposite order, e.g. when the system closes the socket before the async write is attempted.
In any case, the crash happens because the `tls_http_server` get into a state where it waits for two asynchronous I/O operations concurrently. If both fail (e.g. due to a broken socket) the TLS channel is destroyed twice and the second attempt causes a nullptr dereference.

In the happy case, this concurrent I/O situation does occur as well. However, they don't run into an error simultaneously because in a clean connection termination `::handle_client_write_completion()` does not schedule another async write after all pending data has been written.

Reproducer:

```python
import socket
import binascii

# A valid TLS 1.3 client hello (taken from openssl s_client -debug ...)
CLIENT_HELLO = "16030101200100011c030344338cf9eef97fda9f081804430af03c2ee9f98a9cd4fe761e6d5f9cc1ebdd6e20d9f9fd980f02f7cac917f6692667291de47a78ef0a755da52736d3af5f96dcda003e130213031301c02cc030009fcca9cca8ccaac02bc02f009ec024c028006bc023c0270067c00ac0140039c009c0130033009d009c003d003c0035002f00ff01000095000b000403000102000a00160014001d0017001e0019001801000101010201030104002300000016000000170000000d002a0028040305030603080708080809080a080b080408050806040105010601030303010302040205020602002b00050403040303002d00020101003300260024001d00201daa045ceed56fa669f9dc35dff4ba65690edfc9c52d8238492c2a93f6d46e74"

TCP_IP = '127.0.0.1'
TCP_PORT = 8443
BUFFER_SIZE = 1024

s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.connect((TCP_IP, TCP_PORT))

# Send the client hello and close the connection immediately
s.send(binascii.unhexlify(CLIENT_HELLO))
s.close()
```

PR #3639, opened 2023-07-24
Modify the CI script to fall back to nmake if jom is missing

### LINKED ISSUE #3629: CI breaking because jom download is flaky
Occasionally Windows CI fails because `choco` fails to download `jom` due to a timeout.

Not really sure how to fix this atm

randombit:
Fixes #3629

### DIFF
--- a/src/scripts/ci_build.py
+++ b/src/scripts/ci_build.py
@@ -522,6 +522,23 @@ def have_prog(prog):
                 return True
     return False
 
+def validate_make_tool(make_tool, build_jobs):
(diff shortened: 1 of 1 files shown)


reneme (approved):
Certainly just a hack. I agree, though, it's better than random failures.

At the expense of some vendor lock-in: We might consider [leaving a note in the job summary](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary) about that.

PR #3640, opened 2023-07-25
Try some GitHub Actions Workflow Commands

reneme:
I felt like playing with GitHub Actions Commands. This is meant as a demo/test bed. If we want to keep it, I could split it into individual PRs.

See here for details: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions

### Grouping the Build Log

That change is simple but potentially simplifies the log exploration significantly. The GitHub Actions specific stuff is printed only when `export $GITHUB_ACTIONS=true` (as [stated in GitHub's documentation](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)):

> Always set to true when GitHub Actions is running the workflow. You can use this variable to differentiate when tests are being run locally or by GitHub Actions.

image

### Clang Tidy Warnings in the Pull Request

`clang-tidy` allows exporting its findings as a YAML file. I added a small python helper that reads this and translates it into GitHub Actions annotations with file coordinates.

image

### DIFF
--- a/src/scripts/ci/gh_clang_tidy_fixes_in_pr.py
+++ b/src/scripts/ci/gh_clang_tidy_fixes_in_pr.py
@@ -1,0 +1,120 @@
+#!/usr/bin/env python3
+
+"""
+(C) 2023 Jack Lloyd
+    2023 René Meusel - Rohde & Schwarz Cybersecurity
+
+Botan is released under the Simplified BSD License (see license.txt)
+"""
+
+import argparse
+import glob
+import os
+import sys
+
+import yaml
+
+class FileLocation:
+    def __init__(self, line : int, column : int, endline : int | None = None, endcolumn : int | None = None):
+        self.line = line
+        self.column = column
+        self.endline = endline
+        self.endcolumn = endcolumn
+
+    def has_range(self):
+        return self.endline is not None and self.endcolumn is not None
+
+
+class Diagnostic:
+    @staticmethod
+    def read_diagnostics_from(yaml_file : str):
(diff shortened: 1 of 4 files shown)


reneme:
Other additions worth looking into:

* Translate the JUnit results into GitHub Actions Commands
* Visualize compiler warnings/errors similarly to clang-tidy
   (off-the-shelf [problem matchers](https://michaelheap.com/getting-started-problem-matchers/) exist for this)
* ...

src/scripts/ci/gh_clang_tidy_fixes_in_pr.py
```diff
@@ -0,0 +1,120 @@
...
+            if "Diagnostics" not in fixes:
+                raise RuntimeError(f"No Diagnostics found in {yaml_file}")
+            return [Diagnostic(diag) for diag in fixes["Diagnostics"]]
+
+
+    def __map_file_path(self, file_path, base_path): # pylint: disable=unused-argument
+        if file_path.endswith(".h"):
+            raise RuntimeError(f"Header file {file_path} cannot be handled")
+        # TODO: try to map include files (residing in build/include) onto their
+        #       origin path in src/lib etc.
```
reneme, line 42: This might become urgent once https://github.com/randombit/botan/issues/3644 is done.

reneme (commented):
No need that this was still a draft. @randombit both CI features should be ready to merge, if you think they could be useful.

randombit (approved):
:+1:

PR #3643, opened 2023-07-25
Remove bswap_4

randombit:
This function made a little more sense when it was coded to use SSE2 to directly perform the byteswaps. But these days we're better of just writing a simple loop and letting the compiler figure it out.

Checked performance of some big-endian algorithms on a LE machine (SHA-1, SM3, CAST-128) if anything performance slightly improved.


reneme (approved):

randombit:
Merged, not sure what happened here

Issue #3444, opened 2023-03-27
Add TLS-Anvil to testing setup


randombit:
https://tls-anvil.com/ seems to be derived from/ from the same team as TLS-Attacker, but :sparkles:  fully automated :sparkles: :heart:

FAlbertDev:
Yeah, I've worked with/for TLS-Anvil in my master thesis. It uses so called "combinatorial testing" to systematically test various combinations of TLS handshake parameters (e.g. cipher suites, enabled extensions, etc.). Its test suite tries to cover all RFC statements, so yeah: Its a nice tool :D  
If we want to use TLS-Anvil for Botan, we should make some considerations:

1. When using TLS-Anvil we can define a combinatorial testing strength defining how many combinations of handshake parameters are tested (e.g. `strength=2` requires that all possible combinations of two parameters are covered by the test vector). These combinations are tested in TLS-Anvil's test suite. Consequently, we can choose freely, if want a more extensive testing using a high strength (e.g. `strength=3`), which is much slower than testing with low strenghts. Maybe we should run a one-time manual analysis with high strength and  integrate a CI task with low strength.

2. If we want to consider a CI/CD integration, we need to discuss how to automatically interpret the results. As far as I remember, TLS-Anvil outputs multiple .json files which can be processed by an analyser for manual analysis in the browser. To process them automatically, extra work will be required.

Maybe I can invest some time in the following weeks to setup and run TLS-Anvil with Botan (manually first). I will then share the results. However, as far as I know, the authors of TLS-Anvil already ran TLS-Anvil with various TLS Libraries including Botan. I guess, if there were any findings at this point, they would have reported them. However, a CI task for TLS-Anvil is a good addition in my opinion...

securitykernel:
We're running this in nightly CI now! 🚀

PR #3651, opened 2023-07-31
TLS Anvil Server Tests in Nightly CI

FAlbertDev:
(Refers to #3444)
This adds [TLS Anvil](https://github.com/tls-attacker/TLS-Anvil) server tests to Botan's nightly CI. Unfortunately, a current bug of the origin TLS-Anvil repo required me to create a [fork](https://github.com/FAlbertDev/TLS-Anvil) and work around it (I already created an issue for the TLS-Anvil repo). Until the bug is fixed, the CI script uses my fork. 
The job's workflow is the following:
1) Build Botan and TLS-Anvil
2) Run the `script src/scripts/ci/ci_tlsanvil_test.py` to setup and start a botan tls_server and the TLS-Anvil process. The TLS-Anvil process tests the server by playing a client  (repeatedly) connecting to the server. TLS-Anvil creates a TestSuiteResults folder for storing the results with additional test information.
3) The TestSuiteResults folder and the TLS-Anvil log files are uploaded as an artifact, so we can analyze them if necessary.
4) The script `src/scripts/ci/ci_tlsanvil_check.py` checks if the tests are successful, i.e., if the results in TestSuiteResults meet our expectations. Currently, some tests fail. I will look into these together with @reneme after his holidays. Until then, I allowed the respective tests to fail.

An example workflow execution with failing tests (without deactivating the currently failing tests) is [here](https://github.com/FAlbertDev/botan/actions/runs/5690338681/job/15423485761). When allowing the tests to fail, the workflow looks like [this](https://github.com/FAlbertDev/botan/actions/runs/5692102195/job/15428489867). Since the tests take around 1.5h, they are only executed nightly. 
I also plan to add the TLS-Anvil client tests in a future PR.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,5 +1,6 @@
 
 # (C) 2023 Jack Lloyd
+# (C) 2023 Fabian Albert, Rohde & Schwarz Cybersecurity
 #
 # Botan is released under the Simplified BSD License (see license.txt)
 
@@ -54,3 +55,35 @@ jobs:
 
       - name: Valgrind Checks
         run: python3 ./src/scripts/ci_build.py --cc=gcc --make-tool=make valgrind-full
+
+  tls_anvil_server_test:
+    name: "TLS-Anvil - Botan Server Test"
+
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Fetch Botan Repository
+        uses: actions/checkout@v3
+
+      - name: Setup Build Agent
+        uses: ./.github/actions/setup-build-agent
+        with:
+          target: tlsanvil
+          cache-key: linux-x86_64-tlsanvil
+
+      - name: Build and Test Botan Server with TLS-Anvil
+        run: >
+          python3 ./src/scripts/ci/ci_tlsanvil_test.py
+          --botan-dir .
+          --test-target server
+          --parallel $(nproc)
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: tls-anvil-server-test-results
+          path: |
+            ./TestSuiteResults/
+            ./logs/
+
+      - name: Check TLS-Anvil Test Results
+        run: python3 ./src/scripts/ci/ci_tlsanvil_check.py --verbose ./TestSuiteResults
--- a/src/scripts/ci/ci_tlsanvil_test.py
+++ b/src/scripts/ci/ci_tlsanvil_test.py
@@ -1,0 +1,139 @@
+# Script to run inside the CI container to test the botan
+# TLS client/server with TLS-Anvil
+#
+# (C) 2023 Jack Lloyd
+# (C) 2023 Fabian Albert, Rohde & Schwarz Cybersecurity
+#
+# Botan is released under the Simplified BSD License (see license.txt)
+import sys
+import argparse
+import os
+import subprocess
+
+
+class Config:
+    """ Hardcoded configurations for this CI script """
+    tls_anvil_docker_image = "ghcr.io/tls-attacker/tlsanvil"
+    tls_anvil_version_tag = "v1.2.2"
+    key_and_cert_storage_path = "/tmp/"
+    test_suite_results_dest = "."
+    test_suite_results_dir_name = "TestSuiteResults"
+    tmp_key_file_name = "tmp_rsa_key.pem"
+    tmp_cert_file_name = "tmp_rsa_cert.pem"
+    server_dest_ip = "127.0.0.1"
+    server_dest_port = 4433
+    botan_server_log = "./logs/botan_server.log"
+    botan_config_args = ["--compiler-cache=ccache", "--build-targets=static,cli",
+                          "--without-documentation", "--with-boost"]
+
+def group_output(group_title: str, func):
+    """
+    Wraps a function to be called within a GitHub actions group, so that
+    the console output is expandable.
+
+    Returns the wrapped function
+    """
+    def wrapped_func(*args, **kwargs):
+        print(f"::group::{group_title}", flush=True)
+        ret = func(*args, **kwargs)
+        print("\n::endgroup::", flush=True)
+        return ret
+    return wrapped_func
+
+
+def create_cert_and_key(botan_dir_path):
+    """
+    Create a X.509 certificate and associated RSA key at Config.key_and_cert_storage_path
+    using Botan's CLI.
+
+    Returns: (<cert path>, <key path>)
+    """
+
+    key_path = os.path.join(Config.key_and_cert_storage_path, Config.tmp_key_file_name)
+    cert_path = os.path.join(Config.key_and_cert_storage_path, Config.tmp_cert_file_name)
+
+    with open(key_path, 'w', encoding='utf-8') as keyfile:
+        subprocess.run([botan_dir_path, "keygen", "--algo=RSA", "--params=2048"], stdout=keyfile, check=True)
+
+    with open(cert_path, 'w', encoding='utf-8') as certfile:
+        subprocess.run([botan_dir_path, "gen_self_signed", key_path, "localhost"], stdout=certfile, check=True)
+
+    return (cert_path, key_path)
+
+
+def build_botan(botan_dir: str, parallel_jobs: int) -> str:
+    """
+    Configure and build botan.
+
+    Returns the botan executable path
+    """
+    group_output("Configure Botan", subprocess.run)(["python3", "./configure.py"] + Config.botan_config_args, check=True, cwd=botan_dir)
+    group_output("Build Botan with Make", subprocess.run)(["make", f"-j{parallel_jobs}"], check=True, cwd=botan_dir)
+
+    return os.path.join(botan_dir, "botan")
+
+
+def server_test(botan_dir_path: str, parallel: int):
+    """ Test the Botan TLS server """
+    cert_path, key_path = create_cert_and_key(botan_dir_path)
+    docker_img = f"{Config.tls_anvil_docker_image}:{Config.tls_anvil_version_tag}"
+
+    group_output("Pull TLS-Anvil image", subprocess.run)(["docker", "pull", docker_img], check=True)
+
+    tls_anvil_cmd = [
+        "docker", "run",
+        "--network", "host",
+        "-v", f"{Config.test_suite_results_dest}:/output",
+        docker_img,
+        "-strength", "1",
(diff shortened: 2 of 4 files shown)


src/scripts/ci/ci_tlsanvil_test.py
```diff
@@ -0,0 +1,113 @@
...
+    using Botan's CLI.
+
+    Returns: (<cert path>, <key path>)
+    """
+
+    key_path = os.path.join(Config.key_and_cert_storage_path, Config.tmp_key_file_name)
+    cert_path = os.path.join(Config.key_and_cert_storage_path, Config.tmp_cert_file_name)
+
+    with open(key_path, 'w', encoding='utf-8') as keyfile:
+        subprocess.run([botan_exe_path, "keygen", "--algo=RSA"], stdout=keyfile, check=True)
```
randombit, line 37: `keygen` with RSA defaults to 3072 bits since that is ~~ 128 bit security level, but for testing this is just some needless computational overhead - let's generate 2048 bit instead since that is still sufficient and may save a bit of test time.

src/scripts/ci/ci_tlsanvil_test.py
```diff
@@ -0,0 +1,113 @@
...
+    return (cert_path, key_path)
+
+
+def server_test(botan_exe_path: str, tls_anvil_jar_path: str):
+    cert_path, key_path = create_cert_and_key(botan_exe_path)
+
+    tls_anvil_cmd = [
+        "java", "-jar", tls_anvil_jar_path,
+        "-strength", "1",
+        "-parallelHandshakes", "1",
```
randombit, line 51: The runners we get have 2 cores - would it make sense / be possible to use 2-3 handshakes in parallel here?

src/scripts/ci/ci_tlsanvil_test.py
```diff
@@ -0,0 +1,113 @@
...
+        action="store_true",
+        default=False,
+        help="Test the Botan TLS server",
+    )
+    parser.add_argument(
+        "--client-test",
+        action="store_true",
+        default=False,
+        help="Test the Botan TLS client",
+    )
```
randombit, line 91: Maybe instead a single option `--test-target=` which could take on any of `client`, `server,` or `client,server`?

.github/workflows/nightly.yml
```diff
@@ -54,3 +55,52 @@ jobs:
...
+        uses: actions/checkout@v3
+
+      - name: Setup Build Agent
+        uses: ./.github/actions/setup-build-agent
+        with:
+          target: static
+          cache-key: linux-clang-x86_64-static
+
+      - name: Build Botan
+        run: ./configure.py --compiler-cache=ccache --build-targets=static,cli --without-documentation && make -j8
```
randombit, line 75: TBH I don't know if this even works in GH Actions but should this be something like

`make -j$(nproc)`

so the build scales to using all (and only) the number of cores available?

.github/workflows/nightly.yml
```diff
@@ -54,3 +55,52 @@ jobs:
...
+
+    steps:
+      - name: Fetch Botan Repository
+        uses: actions/checkout@v3
+
+      - name: Setup Build Agent
+        uses: ./.github/actions/setup-build-agent
+        with:
+          target: static
+          cache-key: linux-clang-x86_64-static
```
randombit, line 72: This cache key seems wrong since we are using gcc to build

I think if we use the `linux-gcc-x86_64-static` key, then we can share caches with this build and the normal static library GCC CI build.

randombit (commented):
Thanks for setting this up @FAlbertDev, I'll be happy when this is part of our test coverage.

I have a few minor comments/questions but overall looks very good.

src/scripts/ci/ci_tlsanvil_test.py
FAlbertDev: Yeah, parallel handshakes would be great. However, I run TLS-Anvil against the `./botan tls_server` process, which only allows one connection at a time.

.github/workflows/nightly.yml
FAlbertDev: It seems to work 👍

src/scripts/ci/ci_tlsanvil_test.py
FAlbertDev: Seems to be not significantly faster with 2048 bit. Won't hurt, though.

FAlbertDev:
Thanks for your review, @randombit! I addressed your suggestions in ba484ec :)

src/scripts/ci/ci_tlsanvil_test.py
randombit: Could we use `tls_http_server` instead?

.github/workflows/nightly.yml
```diff
@@ -54,3 +55,56 @@ jobs:
...
+        uses: actions/setup-java@v3
+        with:
+          java-version: '11'
+          distribution: 'zulu'
+
+      - name: Fetch TLS-Anvil fork for Botan tests
+        uses: actions/checkout@v3
+        with:
+          repository: FAlbertDev/TLS-Anvil
+          path: ./tlsanvil
```
reneme, line 87: If we need a fork, maybe it should live under @randombit's GitHub account. Similarly to the `boringssl` fork we maintain for BoGo.

.github/workflows/nightly.yml
```diff
@@ -54,3 +55,56 @@ jobs:
...
+        uses: actions/checkout@v3
+
+      - name: Setup Build Agent
+        uses: ./.github/actions/setup-build-agent
+        with:
+          target: static
+          cache-key: linux-gcc-x86_64-static
+
+      - name: Build Botan
+        run: ./configure.py --compiler-cache=ccache --build-targets=static,cli --without-documentation && make -j$(nproc)
```
reneme, line 75: Maybe configure and build botan in the `ci_tlsanvil_test.py` instead?

.github/workflows/nightly.yml
```diff
@@ -54,3 +55,56 @@ jobs:
...
+        run: mvn install -DskipTests -Dspotless.apply.skip
+
+      - name: Test Botan Server with TLS-Anvil
+        run: >
+          python3 ./src/scripts/ci/ci_tlsanvil_test.py
+          --botan-exe ./botan
+          --tlsanvil-jar ./tlsanvil/TLS-Testsuite/apps/TLS-Testsuite.jar
+          --test-target server
+        env:
+          DOCKER: 1 # TLS-Anvil specific to disable the loading bar
```
reneme, line 100: Please set this inside the `ci_tlsanvil_test.py` script, if possible. Rationale: We try to keep the CI YAML files as "dumb" as possible and move as much of the configuration and tool-intricacies into the dedicated scripts.

src/scripts/ci/ci_tlsanvil_test.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+
+    key_path = os.path.join(Config.key_and_cert_storage_path, Config.tmp_key_file_name)
+    cert_path = os.path.join(Config.key_and_cert_storage_path, Config.tmp_cert_file_name)
+
+    with open(key_path, 'w', encoding='utf-8') as keyfile:
+        subprocess.run([botan_exe_path, "keygen", "--algo=RSA", "--params=2048"], stdout=keyfile, check=True)
+
+    with open(cert_path, 'w', encoding='utf-8') as certfile:
+        subprocess.run([botan_exe_path, "gen_self_signed", key_path, "localhost"], stdout=certfile, check=True)
```
reneme, line 40: As a general remark: At one point we should spin-off some internal `Utilities` python module that we can reuse across the internal python machinery. As a center piece this will likely include a convenience wrapper around `subprocess.run` and/or `subprocess.Popen`. No idea, how many scripts already have a copy of `run_commend()`. 🤡

reneme (commented):
Nice! Just a few general remarks below. Detailed review will need to come later.

.github/workflows/nightly.yml
reneme: (I'm guessing you hope to get rid of the fork after your upstream issue is fixed)

FAlbertDev: > (I'm guessing you hope to get rid of the fork after your upstream issue is fixed)

True. In fact, the same hour you wrote this comment, the issue was fixed ;)

src/scripts/ci/ci_tlsanvil_test.py
FAlbertDev: Yeah. But we need to evaluate if an additional abstraction layer is worth it. I think, in this case, the calls are slim enough.

FAlbertDev:
My current state for this PR: 
I'm currently working on the usage of the `tls_http_server` instead of `tls_server` to speed up the tests by parallelization. Some test results are currently inconsistent between both CLI modules (probably because of a small issue with the tls_http_server). I'll investigate this and will probably open another PR addressing this issue after understanding the problem better.
Also, I need to implement your suggestions, @reneme, and redirect the checkout of TLS-Anvil to the origin repo, which is fixed now.

.github/workflows/nightly.yml
```diff
@@ -54,3 +55,38 @@ jobs:
...
+
+    steps:
+      - name: Fetch Botan Repository
+        uses: actions/checkout@v3
+
+      - name: Setup Build Agent
+        uses: ./.github/actions/setup-build-agent
+        with:
+          target: static
+          cache-key: linux-gcc-x86_64-static
```
FAlbertDev, line 72: Do we need a separate cache key because we build with `--with-boost`?

randombit: Oh probably so - maybe best to just use a new unique cache id for each job and let the cache management system sort it out rather than trying to share caches and causing conflicts and cache misses if/when the configuration of different jobs changes over time.

randombit: Also: if we use a unique target here (eg `anvil`) we can install boost in `setup_gh_actions.sh` as a side effect of setting up the build agent, rather than having an explicit install of boost later on in the yaml

FAlbertDev: Done. (see a589ed2)

Thanks for your feedback!

randombit:
LGTM. Only remaining comment from me would be please squash before merging.

FAlbertDev:
> please squash before merging

Done.

randombit (approved):

reneme (approved):

PR #3656, opened 2023-08-04
For swap use ADL instead of specializing std::swap

randombit:
Not using ADL is a remnant of pre-C++11

### DIFF
--- a/src/lib/math/bigint/bigint.h
+++ b/src/lib/math/bigint/bigint.h
@@ -171,6 +171,8 @@ class BOTAN_PUBLIC_API(2, 0) BigInt final {
          std::swap(m_signedness, other.m_signedness);
       }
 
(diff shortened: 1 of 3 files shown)


reneme (commented):
Isn't that (technically) a breaking API change that requires users to switch from `std::swap()` to just `swap()`?

At least that's how understand [the example code in the STL documentation](https://en.cppreference.com/w/cpp/algorithm/swap#Example).

randombit:
I don't think this is a breaking change. It does mean that using `std::swap`, at least with the usual implementation of

```
T c = move(a);
a = move(b);
b = move(c);
```

will be a bit slower than it was since we do 3 swaps instead of 1. But it still works just fine.

I'm basing this change on comments from Dave Abrahams and Howard Hinnant in this SO post https://stackoverflow.com/questions/11562/how-to-overload-stdswap

reneme (approved):

Issue #3664, opened 2023-08-09
Add a way of detecting explicit curves


randombit:
https://github.com/randombit/botan/discussions/3630

PR #3665, opened 2023-08-09
Extend EC_Group to track when a group used the explicit encoding

randombit:
GH #3630 #3664

### DIFF
--- a/src/lib/pubkey/ec_group/ec_group.h
+++ b/src/lib/pubkey/ec_group/ec_group.h
@@ -351,6 +351,14 @@ class BOTAN_PUBLIC_API(2, 0) EC_Group final {
       EC_Group_Source source() const;
 
       /**
+      * Return true if this EC_Group was derived from an explicit encoding
+      *
+      * Explicit encoding of groups is deprecated; when support for explicit curves
+      * is removed in a future major release, this function will also be removed.
+      */
+      bool used_explicit_encoding() const { return m_explicit_encoding; }
+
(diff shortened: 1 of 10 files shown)


src/lib/pubkey/ec_group/ec_group.h
```diff
@@ -373,9 +381,9 @@ class BOTAN_PUBLIC_API(2, 0) EC_Group final {
    private:
       static EC_Group_Data_Map& ec_group_data();
 
-      static std::shared_ptr<EC_Group_Data> BER_decode_EC_group(const uint8_t bits[],
-                                                                size_t len,
-                                                                EC_Group_Source source);
+      static std::pair<std::shared_ptr<EC_Group_Data>, bool> BER_decode_EC_group(const uint8_t bits[],
+                                                                                 size_t len,
+                                                                                 EC_Group_Source source);
```
reneme, line 386: Instead of returning a tuple, couldn't we store the `explicit_encoding` flag in `EC_Group_Data` as well? It's an internal structure anyway...

randombit: No - because `EC_Group_Data` is shared among all instances of the same group. While here the goal is to track the specific source of *this one* `EC_Group` regardless of if it points to a known group or not. If we tracked it in `EC_Group_Data` then we would be still unable to distinguish between P-256 that came from an OID, and P-256 that came from an explicit encoding.

PR #3669, opened 2023-08-12
test cpu probe adding version for RISCV.

devnexen:


devnexen:
Work for [both 32 and 64 bits](https://godbolt.org/z/38Kz8sb5s).

devnexen:
does not seem to behave as I thought it would.

Issue #3543, opened 2023-05-09
Kyber Standardization Updates


randombit:
We shouldn't change anything right away but the Kyber designers are [making changes](https://github.com/pq-crystals/kyber/tree/standard) to their code to target the 'final' version of Kyber.

So far these seem to include

- [ ] No more 90s mode (this will allow a nice simplification of the codebase)
- [ ] ~~Using SHAKE128 instead of CTR-DRBG to generate the test vectors~~
- [ ] Using SHAKE256 instead of SHA-3 for PRF

Looking over the commits I don't see any changes to the core encap/decap algorithms

randombit:
https://www.federalregister.gov/documents/2023/08/24/2023-18197/request-for-comments-on-draft-fips-203-draft-fips-204-and-draft-fips-205

randombit:
https://csrc.nist.gov/pubs/fips/203/ipd

randombit:
In large part it depend on how big the difference is between the proposed and final versions. From what I've seen it is pretty minor, and we could accommodate it with just a new `KyberMode`.

I think we should deprecate Kyber 90s mode, since it is not only not going to be in the final version but also many Kyber implementations already use only the SHAKE based primitives, and 90s mode leads to some significant extra complexity in the code, eg the CTR XOF. We could probably consider an accelerated (pre 4.0) removal here.

I rather suspect that round2 SHAKE based Kyber is here to stay. There are already a number of deployed systems that use it (for instance Signal) and in the near term that is likely to increase since nobody implements the final NIST version yet, so any new systems adopting Kyber today are going to pick up the non-final version.

Re terminology, I'd prefer continuing to call it Kyber since that is the name under which it is known generally. I've already had some experience with calling something under the name from a standard rather than the conventional name ("EMSA3" vs "PKCS1v15") and it just leads to user confusion.

falko-strenzke:
> * ML-KEM should be [marked "experimental"](https://github.com/randombit/botan/discussions/3110) in a sense that draft-standard updates would simply be applied in the next release, without backward compatibility

I would prefer to refrain from this approach. Marking it "experimental" isn't solving the problem of those who will use it for experimental implementations that are supposed to be interoperable with other implementations. Silently changing the implementation is not a good idea. In https://github.com/randombit/botan/issues/3697 I proposed a different naming which would prevent such confusion.

falko-strenzke:
[...]

PR #3671, opened 2023-08-15
eXtendable Output Functions as first-class citizen

reneme:
### Motivation

Most PQC schemes use XOFs as a building block in one or the other way. This is a proposal to establish `XOF` as a new public API base class. The only implementation currently provided is `SHAKE-128` and `SHAKE-256`.

### Pull Request Dependencies

* ~#3673~

### Find included

* New top-level module "xof" with `class XOF` as abstract base class
  * `start()` allows priming certain XOFs with a `salt` and `key` (otherwise its a NOOP) 
  * `update()` allows streamed input of an arbitrarily long message
  * `output()` allows streamed access to the XOF's infinitely long output stream
  * typical implementations won't allow `update()`s after the first call to `output()`
* `SHAKE_128_XOF` and `SHAKE_256_XOF` as sub-classes of `XOF`
* `cSHAKE_128_XOF` and `cSHAKE_256_XOF` as sub-classes of `XOF` (for internal use only)
* internal helper functions (roughly) as defined in [NIST SP.800-185](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-185.pdf) Section 2.3
  (those functions can be reused to implement KMAC -- see #3570)

#### XOFs based on stream ciphers

Kyber and Dilithium build their "90s-crypto" variants on XOF constructions based on AES-256/CTR. We could look into providing support for such stream-cipher XOFs independent of the CRYSTALS algorithms. (See #3672)

#### Retrofitting PQC algorithms

That's future work and will be done in a follow-up. (See #3672)


reneme:
Rebased onto @falko-strenzke's Keccak refactoring as a sanity checks. Findings are in #3673.

src/lib/base/buf_input.h
```diff
@@ -0,0 +1,73 @@
...
+* internal state with an arbitrary number of streamlined update() calls.
+*/
+class BOTAN_PUBLIC_API(3, 2) Buffered_Input {
+   public:
+      /**
+      * Add new input to process.
+      * @param in the input to process as a byte array
+      * @param length of param in in bytes
+      */
+      void update(const uint8_t in[], size_t length) { add_data(in, length); }
```
randombit, line 31: `Buffered_Computation` is already deprecated, I don't think we should use it for the XOF

This also allows some modernization and minimization of the API. In fact we could literally go to just taking a `std::span<const uint8_t>` and be done with it.

src/lib/hash/shake/shake.h
```diff
@@ -9,71 +10,70 @@
...
+}  // namespace detail
+
 /**
-* SHAKE-256
+* SHAKE-128 with a fixed output length
+*
+* For variable output lengths consider using the XOF interface
 */
-class SHAKE_256 final : public HashFunction {
+class SHAKE_128 final : public detail::SHAKE_Adapter<SHAKE_128_XOF, SHAKE_128> {
```
randombit, line 64: All of this seems completely unnecessary. Why not just hold a `std::unique_ptr<XOF>` of the appropriate type?

src/lib/stream/shake_cipher/shake_cipher.h
```diff
@@ -1,77 +1,128 @@
...
-      void set_iv_bytes(const uint8_t iv[], size_t iv_len) final;
-
-   protected:
-      size_t m_shake_rate;
+      void cipher_bytes(const uint8_t in[], uint8_t out[], size_t length) final {
+         assert_key_material_set();
+
+         const auto block_size = m_keystream_buffer.size();
+
+         auto cipher_some = [&](size_t bytes) {
```
randombit, line 65: I don't see why this logic has to be in a header

src/lib/base/buf_input.h
reneme: Mhm, fair enough. Albeit, I found it quite handy to have `update_le()`, for instance.

You're proposing to omit the common base interface altogether, right?

reneme:
The whole Keccak-Refactoring is becoming a mess with several inter-dependent pull requests. Sorry for that. Lets put this one on-hold until #3673 and #3675 are integrated.

src/lib/stream/shake_cipher/shake_cipher.h
reneme: Yeah, that was still sketchy. Addressed in #3675.

Now that `Keccak_Permutation::squeeze()` can handle multiple calls to pull the sponge output in arbitrarily-sized chunks, I decided not the base `SHAKE_Cipher` on the `SHAKE_XOF` but directly on `Keccak_Permutation`. There's no difference in level of abstraction anyway.

src/lib/hash/shake/shake.h
reneme: The rationale is similar to [the one discussed back in May](https://github.com/randombit/botan/pull/3525#discussion_r1186276249). I think, we never really came to a conclusion whether this is just a code smell or might cause actual trouble.

Regardless: I decided not to touch the `HashFunction` implementation of SHAKE in this PR at all. With #3675 introducing `Keccak_Permutation::squeeze()`, there's no significant difference in level of abstraction between `Keccak_Permutation` and `SHAKE_XOF`. Hence, I think it's just fine to avoid the indirection via `SHAKE_XOF` here.

reneme:
Re: XOFs based on stream ciphers

Maybe we should step back and rethink the XOF interface somewhat. Looking at [NIST SP.800-185](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-185.pdf) they define things like KMAC-XOF that is KMAC with a variable output length. I.e. just like the AES/CTR-XOF used in CRYSTALS Kyber/Dilithium, it takes a key (and potentially other customization parameters).

I think it would be great to incorporate such XOFs off the bat, when introducing a new API.

reneme:
To my mind an abstract XOF should have two phases:

1. Parameterization
2. Output

The "output phase" is simple and inherently generic. The user just calls `::output()` as often as they want to read the indefinite stream of XOF output bytes.

The "parameterization phase" is algorithm-specific, unfortunately. Here are some relevant examples:

| Algorithm   | Parameters                                |
| ----------- | ----------------------------------------- |
| SHAKE       | an arbitrary-length message `msg`         |
| cSHAKE      | arbitrary-lengths `N` and `S`, and `msg`  |
| AES/CTR-XOF | a short `iv` and `key`                    |
| KMAC-XOF    | arbitrary-length `key`, `msg` and `salt`  |

Reflecting this parameter space in a common `XOF` base-interface is possible but probably won't result in a convenient API. Except, if the only "common" feature of this interface is the `::output()` method which could be shared by all XOFs.

reneme:
Given the above-described constraints, it seems reasonable to abandon the ubiquitous `::create("AlgoSpec")` factory methods for XOFs. Instead, the user could instantiate the concrete XOF subclass, parameterize it and then (maybe) use it via the abstract `XOF` base class that just provides the `::output()` functionality.

@randombit Does that sound like a reasonable abstraction to you?

reneme:
Benchmarking the SHAKE-128/256 XOF showed comparable results for `update()` and `output()` and correspond to [the measured performance of SHA-3(256)/SHA-3(512)](https://github.com/randombit/botan/pull/3673#issuecomment-1691130234). No obvious bottlenecks.

|            | absorb()    | squeeze()   |
| ---------- | ----------- | ----------- |
| SHAKE-128  | 555MiB/sec  | 530MiB/sec  |
| SHAKE-256  | 448MiB/sec  | 432MiB/sec  |

randombit:
>  Instead, the user could instantiate the concrete XOF subclass, parameterize it and then (maybe) use it via the abstract XOF base class that just provides the ::output() functionality.

I very much dislike this. A lot of effort was spent in 2->3 in removing all of the concrete subclasses from the public API; this minimizes API and ABI surface. This would also complicate the FFI interface.

I don't think CTR mode "XOF" is worth optimizing for. It is a weird corner case that should not be exposed to applications. And it seems Kyber is dropping "90s mode" entirely (#3543) so it is probably going to be removed soonish anyway.

Then we're left with 

* SHAKE: arbitrary length msg
* cSHAKE: arbitrary length msg. `N` is not for us - quoting SP 800-185 "N is a function-name bit string, used **by NIST** to define functions based on cSHAKE. When no function other than cSHAKE is desired, N is set to the empty string." and so can be ignored. And `S` is a domain separator and can be instantiated in the usual way eg `cSHAKE(256, my_domain_dep)`
* KMAC-XOF: arbitrary length msg, key, and salt.

```
class XOF {
  virtual void start(std::span<const uint8_t> salt = {}, std::span<uint8_t> key = {}) = 0;
  virtual bool valid_salt_length(size_t salt_len) = 0;
  // returns Key_Length_Specification(0) if key is not supported
  virtual Key_Length_Specification key_spec() const = 0;
  virtual void update(std::span<const uint8_t> input);
  //...
};
```

CTR, which doesn't support inputs, is handled by the funny case that `XOF::update` throws `Not_Implemented`.

reneme:
@randombit Thanks for your opinion. Your points make sense and I agree in general. I'll be AFK until the beginning of September and will come back to this after.

randombit:
Needs a rebase now that the SHA-3 refactor is merged; I'll review after.

reneme:
> Needs a rebase now that the SHA-3 refactor is merged; I'll review after.

Rebased. I hope, I get time today to work on the extended API [discussed here](https://github.com/randombit/botan/pull/3671#issuecomment-1691595134).

src/tests/data/xof/shake.vec
```diff
@@ -0,0 +1,539 @@
+# These vectors are selected from the NIST CAVS file for SHAKE
+#  Generated on Thu Jan 28 14:46:49 2016
+
+[SHAKE-128]
```
reneme, line 4: Note to self: the Keccak permutation provides a `bmi2` implementation. Probably, we want to insert

```
#test cpuid bmi2
```

in that test vector file to exercise the alternative provider.

src/lib/base/buf_input.h
randombit: Exactly so, yes

src/lib/xof/shake_xof/shake_xof.cpp
```diff
@@ -0,0 +1,43 @@
...
+SHAKE_XOF::SHAKE_XOF(size_t capacity) : m_keccak(capacity, 0xF, 4), m_output_generated(false) {
+   BOTAN_ASSERT_NOMSG(m_keccak.byte_rate() >= 72 && m_keccak.byte_rate() <= 168);
+}
+
+void SHAKE_XOF::clear() {
+   m_keccak.clear();
+   m_output_generated = false;
+}
+
+void SHAKE_XOF::add_data(const uint8_t key[], size_t length) {
```
randombit, line 25: Calling this param `key` seems odd since there is no requirement that XOF input be secret, and is inconsistent with the other declarations which call this `input` which seems a better name to capture the non-specific nature of the bytes here.

reneme: For sure. Not sure where this came from. Perhaps a copy & paste quirk. Good catch.

reneme:
For tomorrow:

* [ ] Revert the `Buffered_Input` refactoring (see https://github.com/randombit/botan/pull/3671#discussion_r1297073573)
* [ ] Rebase #3672 and try the extended API (`XOF::start()`)
* [ ] Look into cSHAKE

src/lib/xof/xof.h
```diff
@@ -0,0 +1,196 @@
...
+      void output(std::span<uint8_t> output) { generate_bytes(output); }
+
+      /**
+       * @return the next single output byte
+       */
+      uint8_t output_next_byte() {
+         uint8_t out;
+         generate_bytes({&out, 1});
+         return out;
+      }
```
reneme, line 165: `XOF` does not inherit from a generic `Buffered_Input` class anymore. Instead it directly defines `update(std::span<>)` and no additional convenience methods. For consistency, we should probably remove the convenience methods of `output()` as well, no?

My two cents: I did enjoy having some convenience overloads (particularly taking `std::string_view` and Big/Little-Endian integers). Re-implementing them here as well as later in `HashFunction` and `MessageAuthenticationCode` feels like violating the DRY principle, though. Hence, I wouldn't do that.

Nevertheless, I'd suggest keeping two flavors of `output()`:

* `void output(std::span<>)`
   to allow the user to handle memory allocation
* `T output(size_t bytes)` with `T` being a user-defined container that defaults to `secure_vector<>`
   to let the user conveniently pull data wrapped in their favorite vector-ish.

randombit (commented):
First pass review

src/lib/xof/shake_xof/shake_xof.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+void SHAKE_XOF::start(std::span<const uint8_t> salt, std::span<const uint8_t> key) {
+   BOTAN_UNUSED(key);
+   BOTAN_STATE_CHECK(!m_output_generated);
+
+   if(!m_function_name.empty() || !salt.empty()) {
+      // We deliberately do not check the salt length here. There's no technical
+      // reason to reject a salt buffer passed in by an application that exceeds
+      // the (arbitrarily set) valid_salt_length() restriction.
+      keccak_absorb_padded_strings_encoding(*this, block_size(), m_function_name, salt);
```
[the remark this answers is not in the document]
reneme: Yikes, indeed! The condition above was previously meant to exclude non-cSHAKE instances of this base class. But I must have broken it in some restructuring. Anyway, that's certainly not optimal.

I'll look into giving the cSHAKE variants their own base class and leave the salt logic out of the vanilla SHAKE implementation.

src/lib/xof/shake_xof/shake_xof.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+cSHAKE_128_XOF::cSHAKE_128_XOF(std::vector<uint8_t> function_name) :
+      SHAKE_XOF(256, std::move(function_name), 0b00, 2) {}
+
+cSHAKE_128_XOF::cSHAKE_128_XOF(std::span<const uint8_t> function_name) :
+      cSHAKE_128_XOF(std::vector<uint8_t>{function_name.begin(), function_name.end()}) {}
+
+cSHAKE_128_XOF::cSHAKE_128_XOF(std::string_view function_name) :
+      cSHAKE_128_XOF(std::vector<uint8_t>{cast_char_ptr_to_uint8(function_name.data()),
+                                          cast_char_ptr_to_uint8(function_name.data()) + function_name.size()}) {}
```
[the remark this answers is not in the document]
reneme: I'd love that. Unfortunately, the base-class constructor additionally takes the `capacity`. This value is hard-coded in the concrete overloads in cSHAKE-128/256 classes. Therefore we can't just conveniently use the delegates in the derived classes. 😞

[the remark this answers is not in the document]
reneme: Nevertheless, moving the data type conversions into the base class helps a little bit. Then the derived class constructors simply delegate to their base-class counter part.

src/lib/xof/xof.h
```diff
@@ -0,0 +1,196 @@
...
+       */
+      virtual bool valid_salt_length(size_t salt_len) const = 0;
+
+      /**
+       * @returns an object describing limits on the key size
+       */
+      virtual Key_Length_Specification key_spec() const = 0;
+
+      /**
+       * @return the intrinsic processing block size of this XOF
```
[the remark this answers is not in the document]
reneme: I changed this to be a pure-virtual, forcing downstream implementations to make up their minds.

src/tests/test_xof.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+            const std::string provider(xof->provider());
+            result.test_is_nonempty("provider", provider);
+            result.test_eq(provider, xof->name(), algo);
+
+            // Some XOFs don't accept input at all. We assume that this stays the same
+            // after calling `XOF::clear()`.
+            const auto new_accepts_input = xof->accepts_input();
+
+            result.confirm("advertised block size is > 0", xof->block_size() > 0);
```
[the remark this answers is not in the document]
reneme: [Yes](https://github.com/randombit/botan/pull/3671#discussion_r1321161215).

src/lib/xof/xof.h
```diff
@@ -0,0 +1,196 @@
...
+
+      /**
+       * @return the hash function name
+       */
+      virtual std::string name() const = 0;
+
+      /**
+       * Some XOFs can be parameterized with a @p salt and/or @p key. If required,
+       * this must be called before calling XOF::update(). XOFs that don't support
+       * either of these parameters will simply ignore it silently.
```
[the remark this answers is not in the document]
reneme: Changed as described [here](https://github.com/randombit/botan/pull/3671#discussion_r1321161215).

src/lib/permutations/keccak_perm/keccak_helpers.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include <botan/internal/bit_ops.h>
+#include <botan/internal/loadstor.h>
+
+namespace Botan {
+
+namespace {
+
+size_t int_encoding_size(uint64_t x) {
+   return ceil_tobytes(std::max(uint8_t(1), ceil_log2(x + 1)));
```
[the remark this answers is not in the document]
reneme: True, that'll result in a wrong result. The int-encodings are mostly used for encoding bitstring-lengths (for salts, function names and friends). I.e. in practice that shouldn't ever happen. I added an assert to be on the safe side, though. Good catch!

reneme:
Thanks for the review! I rebased to latest master (after merging the `std::span` refactorings) and addressed your review comments.

Most notably, I split the `SHAKE_XOF` and `cSHAKE_XOF` into two separate base classes (and modules). That results in a handful duplicated lines (`::add_data()`, `::generate_bytes()`) but splits off the salt handling in `::start()` much more logically.

randombit (commented):
Looks good now. Should we have some tests for cSHAKE as well?

falko-strenzke:
There are four [test vectors from NIST](https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/cSHAKE_samples.pdf), all of which have an empty value for N.

I just implemented cSHAKE for another libgcrpyt and created two test vectors with non empty N using  https://asecuritysite.com/golang/cs. I checked their results for non-empty N and that is fine.

```
  { /* Created with https://asecuritysite.com/golang/cs */
    GCRY_MD_CSHAKE128,
    "00010203",
    "ABC",
    "Email Signature",
    32,
    "5CF74DC523ADC0B97EC3614E703835277E9F818879AA1EAE5B2B4E4472EB6A68" },
  { /* Created with https://asecuritysite.com/golang/cs */
   CSHAKE256,
    "00010203", // data hex
    "ABC", // N as C string
    "Email Signature", // S as C string
    32, // output size in bytes
    "0C34C14C4A56E5FC01BE8C04C759DA61437E86B88DF3E21A934436D427A85E9D"  // expected output hex
}, 
```

randombit:
Thanks for checking this @falko-strenzke. @reneme can we get these included here so we have some baseline check on cSHAKE? I guess it is a little complicated by cSHAKE not being exposed to applications so you can't immediately use the normal XOF test.

reneme:
@falko-strenzke Nice! Thanks for the test vectors. I did look for something "official" for cSHAKE but wasn't successful.

@randombit I hacked the XOF allocation in `test_xof.cpp` so that it explicitly deals with the internal cSHAKE XOFs when needed. Similarly, I also added vectors for the AES/CTR-XOF in #3672 a while ago.

randombit:
@reneme looks good thanks. I couldn't find the tests themselves though - are you missing a `git add` of `cshake.vec`?

reneme:
I was fighting some linker visibility issues. Had to move the `cSHAKE_XOF::provider()` and `::block_size()` implementations into the *.cpp file to avoid exposing the `Keccak_Permutation` in the shared library.

Now everything should turn green. ☺️

randombit (approved):

reneme:
# 🥳

PR #3672, opened 2023-08-16
Use XOF interface in Kyber and Dilithium

reneme:
### Pull Request Dependencies

* ~#3673~
* ~#3671~
* ~#3701~

### Description

Instead of misusing `StreamCipher` for this job, we now have the top-level `XOF` interface. Also, I added a common XOF implementation based on AES-256/CTR for both Kyber and Dilithium in their 90s-modes.

### Potential Future Work

The polymorphic construction to distinguish between 90s and modern modes in Kyber and Dilithium is instantiating a lot of hash functions that are then just used once by the business logic. There might be room for performance improvements.


reneme:
WIP: At the moment the adaption of the `SHAKE_Cipher` is buggy. I'll need to look into that tomorrow.

reneme:
This will need reconsideration once #3671 is finished and merged. Converted to a draft-PR for now.

reneme:
Rebased to the latest revision of #3671 and adapted to the `XOF::start()` method to apply the IV and key to the AES-256/CTR XOF.

reneme:
Rebased to the latest revision of https://github.com/randombit/botan/pull/3671 and adapted to the XOF::valid_salt_length() method. Also, fixes a missing module dependency in `kyber`.

reneme:
Rebased to the latest revision of https://github.com/randombit/botan/pull/3671 again to resolve expected conflicts with [the cSHAKE test hack](https://github.com/randombit/botan/pull/3671#issuecomment-1717994514).

reneme:
Rebased to master after #3671 was merged.

src/lib/pubkey/kyber/kyber_common/kyber.cpp
```diff
@@ -441,38 +442,13 @@ class Polynomial {
...
-
-         auto xof_reader = XOF_Reader(xof);
+      static Polynomial sample_rej_uniform(XOF& xof) {
          Polynomial p;
 
          size_t count = 0;
          while(count < p.size()) {
-            auto buf = xof_reader.next_3_bytes();
+            std::array<uint8_t, 3> buf;
+            xof.output(buf);
```
randombit, line 451: Does not buffering anymore cause a performance regression? IIRC this change had a pretty noticeable impact

src/lib/pubkey/kyber/kyber_common/kyber_symmetric_primitives.h
```diff
@@ -41,7 +33,7 @@ class Kyber_Symmetric_Primitives {
       virtual std::unique_ptr<HashFunction> H() const = 0;
       virtual std::unique_ptr<HashFunction> KDF() const = 0;
 
-      virtual std::unique_ptr<Kyber_XOF> XOF(std::span<const uint8_t> seed) const = 0;
+      virtual Botan::XOF& XOF(std::span<const uint8_t> seed, std::tuple<uint8_t, uint8_t> matrix_position) const = 0;
```
randombit, line 36: No need to namespace prefix within `Botan`

src/lib/pubkey/kyber/kyber_common/kyber.cpp
reneme: For keygen it does, indeed. I'll look into it.

|                   | keygen/s (master) | keygen/s (XOF) |
| ----------------- | ----------------- | -------------- |
| Kyber-512-r3      | 16095             | 15587          |
| Kyber-512-90s-r3  | 27645             | 14944          |
| Kyber-768-r3      | 13160             | 15992          |
| Kyber-768-90s-r3  | 26266             | 14622          |
| Kyber-1024-r3     | 12685             | 12856          |
| Kyber-1024-90s-r3 | 16779             | 7289           |

src/lib/pubkey/kyber/kyber_common/kyber_symmetric_primitives.h
reneme: Some compilers are confused about the double meaning of the identifier:

```
  botan_all.cpp:8275:20: error: declaration of ‘virtual Botan::XOF& Botan::Kyber_Symmetric_Primitives::XOF(std::span<const unsigned char>, std::tuple<unsigned char, unsigned char>) const’ changes meaning of ‘XOF’ [-fpermissive]
   8275 |       virtual XOF& XOF(std::span<const uint8_t> seed, std::tuple<uint8_t, uint8_t> matrix_position) const = 0;
        |                    ^~~
```

src/lib/pubkey/kyber/kyber_common/kyber.cpp
reneme: The "XOF w/ buffer" uses the following patch. I didn't push that, as it adds an allocation and doesn't really speed up the SHA-3 mode.

```diff
diff --git a/src/lib/pubkey/kyber/kyber_common/kyber.cpp b/src/lib/pubkey/kyber/kyber_common/kyber.cpp
index 107e93f28..fa43c2055 100644
@@ -445,10 +445,12 @@ class Polynomial {
       static Polynomial sample_rej_uniform(XOF& xof) {
          Polynomial p;

+         const auto xof_output = xof.output<secure_vector<uint8_t>>(p.size() * 3);
+         BufferSlicer r(xof_output);
+
          size_t count = 0;
          while(count < p.size()) {
-            std::array<uint8_t, 3> buf;
-            xof.output(buf);
+            const auto buf = r.take(3);

             const uint16_t val0 = ((buf[0] >> 0) | (static_cast<uint16_t>(buf[1]) << 8)) & 0xFFF;
             const uint16_t val1 = ((buf[1] >> 4) | (static_cast<uint16_t>(buf[2]) << 4)) & 0xFFF;
```

reneme:
Rebased ont #3701 to address a similar issues as pointed out here: https://github.com/randombit/botan/pull/3701#discussion_r1327963781

src/lib/pubkey/kyber/kyber_common/kyber.cpp
randombit: OK if the regression is only for 90s mode, and only for keygen, seems fine to leave as is.

src/lib/pubkey/kyber/kyber_common/kyber_symmetric_primitives.h
randombit: WTF???

Anyway fine to leave like that if not prefixing is causing problems otherwise.

randombit (approved):

PR #3673, opened 2023-08-16
Refactoring SHA3: based on new permutation keccak-fips

falko-strenzke:
This PR is needed as a prerequisite of #3570.

PR #3675, opened 2023-08-18
Refactor: Keccak Permutation

reneme:
### Pull Request Dependencies

This builds on top of #3673. @falko-strenzke feel free to incorporate my two patches into your PR and squash it all together. We could then just close this one and merge your's.

### Description

This applies further cleanup to `Keccak_Permutation` and widens its versatility somewhat. Most notably:

* **Replace single-shot `::expand()` with more versatile `::squeeze()`**
  Just like `::absorb()` before, `::squeeze()` can be called multiple times with arbitrary-length buffers. That's really useful for implementing XOFs (see #3671).

* **use `BufferStuffer`/`BufferSlicer` in `::squeeze()` and `::absorb()`**
  That's more of a style thing to avoid some raw pointers and lots of offset shifting. IMHO, that makes the code as "readable" as it gets. 🤡

* **remove low-level access to Keccak state and static mutation methods**
  I believe that's actually the whole point of this new class: No more fiddling with Keccak-states and offset positions in the implementations of "SHA3", "SHAKE" and the "Keccak" hash. @falko-strenzke please make sure that this interface is enough to implement KMAC as well.

* **::permute() is now private**
  Its an implementation detail of the sponge construction and should be handled that way, I think. Or are there use cases where a downstream user would want to have control over when exactly `permute()` should be called?

### My two cents

Was this refactoring worth it? I think so! The implementations of "SHA3", "SHAKE" and the upcoming "SHAKE" XOF are now super slim an concise. Note though, that the implementations for `SHA_3` and `Keccak_1600` (the hash) are basically identical. I hope that's roughly what @randombit had in mind when we started arguing about the architecture back in February (#3279)


randombit:
:+1:

reneme:
Closing, after @falko-strenzke squashed my changes into #3673.

PR #3678, opened 2023-08-23
FIX: explicitly prohibit CT::Mask<bool>

reneme:
This explicitly disables the usage of `bool` in `CT::Mask<>`. We tripped over this when working on FrodoKEM.
Specifically, `CT::Mask<bool>::is_equal(bool, bool)` compiled but generated unexpected results. The internally used `is_zero()` function assumes a multi-bit data type.


randombit:
Even if it worked it would be best avoided as often compilers will do range analysis on `bool` and convert seemingly constant time code into something with branches.

randombit (approved):

PR #3679, opened 2023-08-23
PQC: FrodoKEM

### LINKED ISSUE #3106: Implement FrodoKEM
We will implement [NIST PQC contest round 3](https://csrc.nist.gov/Projects/post-quantum-cryptography/post-quantum-cryptography-standardization/round-3-submissions) alternate candidate [FrodoKEM](https://frodokem.org/) during the [BSI](https://www.bsi.bund.de/EN) Project 481.

~Preliminary schedule for the implementation is Aug '23 - Oct' 23.~

Updated schedule below due to slight project plan rescheduling.

atreiber94:
Closes #3106 by introducing FrodoKEM as specified in the [ISO proposal](https://frodokem.org/files/FrodoKEM-standard_proposal-20230314.pdf).

- Includes all parameter sets:
  - FrodoKEM and ephemeral FrodoKEM ("eFrodoKEM").
  - AES and SHAKE modes for generating the public matrix.
  - All supported matrix sizes (640, 976, 1344).
- Includes optimizations from the reference implementation
- Includes integration into TLS 1.3, X.509, and CLI speed. Compatibility was tested with a liboqs TLS server. If desired, we can also split this integration into separate PRs.
- We did not integrate FrodoKEM into FFI yet.

Open Questions:
- Since FrodoKEM is still in the proposal stage of standardization, should we use separate identifiers (similar to "R3" for Kyber)?
- How to generalize FFI (see Future Work)?

Future Work:
- FFI: The current issue is that a solution similar to Kyber's does not work, since we can't determine the parameter set uniquely based on the given `key_len`. It would be possible to load FrodoKEM keys with an additional parameter specifying the parameter set. However, it may be more worthwhile to provide a general `botan_privkey_load` interface, similar to `botan_privkey_create`.
- Further optimizations may still be possible (SSE, ...).


lieser (commented):
I did just a very quick look at the code, no detailed code review or trying to follow the Frodo algorithm.

src/lib/pubkey/frodokem/frodo_types.h
```diff
@@ -0,0 +1,44 @@
...
+
+using FrodoDomainSeparator = Strong<std::array<uint8_t, 1>, struct FrodoDoaminSeparator_>;
+
+/// u/u' in the spec
+using FrodoPlaintext = Strong<secure_vector<uint8_t>, struct FrodoPlaintext_>;
+
+using FrodoSalt = Strong<std::vector<uint8_t>, struct FrodoSalt_>;
+
+// TODO: Find a better name for this
+using FrodoK = Strong<secure_vector<uint8_t>, struct FrodoK_>;
```
FAlbertDev, line 40: Maybe you can add some comments about the types (+ their respective symbol in the spec)

src/lib/pubkey/frodokem/frodokem.h
```diff
@@ -0,0 +1,105 @@
...
+#include <tuple>
+#include <vector>
+
+namespace Botan {
+
+class FrodoKEM_PublicKeyInternal;
+class FrodoKEM_PrivateKeyInternal;
+
+/**
+ * FrodoKEM is a unstructured lattice-based post-quantum secure KEM. It is a
```
FAlbertDev, line 28:
```suggestion
 * FrodoKEM is an unstructured lattice-based post-quantum secure KEM. It is a
```
(sorry for this nitpicking 😝)

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
```diff
@@ -30,15 +30,33 @@ std::vector<std::pair<std::string, std::string>> algorithm_specs_for_group(Group
          return {{"Curve25519", "Curve25519"}, {"Kyber", "Kyber-512-r3"}};
       case Group_Params::HYBRID_X25519_KYBER_768_R3_OQS:
          return {{"Curve25519", "Curve25519"}, {"Kyber", "Kyber-768-r3"}};
+      case Group_Params::HYBRID_X25519_eFRODOKEM_640_SHAKE_OQS:
+         return {{"Curve25519", "Curve25519"}, {"FrodoKEM", "eFrodoKEM-640-SHAKE"}};
+      case Group_Params::HYBRID_X25519_eFRODOKEM_640_AES_OQS:
+         return {{"Curve25519", "Curve25519"}, {"FrodoKEM", "eFrodoKEM-640-AES"}};
```
FAlbertDev, line 36: They're taking the hobbits to TLS 1.3. Nice!

src/scripts/dev_tools/gen_frodo_kat.py
```diff
@@ -0,0 +1,131 @@
...
+
+def main(args = None):
+    """Set True for eFrodo, otherwise the non-ephemeral variant is assumed.
+    Necessary because the files have the same name for either variant."""
+    is_ephemeral = False
+
+    if args is None:
+        args = sys.argv
+
+    output = open('src/tests/data/pubkey/frodokem_kat.vec', 'w')
```
FAlbertDev, line 110: More modern Python would be: `with open(...) as output:`.

src/tests/test_frodokem.cpp
```diff
@@ -0,0 +1,224 @@
+/*
+ * Tests for FrodoKEM
```
FAlbertDev, line 2: Add more LOTR easter eggs in this file ;)

FAlbertDev (commented):
I didn't compare the code against the specification. I only left some remarks. All in all, LGTM!

src/lib/pubkey/frodokem/frodo_constants.h
```diff
@@ -0,0 +1,105 @@
...
+class BOTAN_TEST_API FrodoKEMConstants {
+   public:
+      FrodoKEMConstants(FrodoKEMMode mode);
+
+      ~FrodoKEMConstants();
+
+      FrodoKEMConstants(const FrodoKEMConstants& other) : FrodoKEMConstants(other.m_mode) {}
+
+      FrodoKEMConstants(FrodoKEMConstants&& other) = default;
+      FrodoKEMConstants& operator=(const FrodoKEMConstants& other) = delete;
```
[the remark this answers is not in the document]
atreiber94: `m_shake_xof` has a deleted copy assignment, so the default does not work and we'd have to build it ourselves. Since we only use the copy constructor, we just deleted it.

src/lib/pubkey/frodokem/frodo_constants.cpp
```diff
@@ -0,0 +1,110 @@
...
+            m_len_se = 256;
+         } else if(mode.is_ephemeral()) {
+            m_len_se = 128;
+         } else {
+            BOTAN_ASSERT_UNREACHABLE();
+         }
+
+         m_cdf_table = {4643, 13363, 20579, 25843, 29227, 31145, 32103, 32525, 32689, 32745, 32762, 32766, 32767};
+
+         m_shake = "SHAKE-128";
```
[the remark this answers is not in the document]
atreiber94: It's not, `m_shake` is 128 for all 640 Frodos and 256 otherwise.

src/lib/pubkey/frodokem/frodo_matrix.h
```diff
@@ -0,0 +1,108 @@
...
+
+   private:
+      size_t m_dim1;
+      size_t m_dim2;
+
+      // This may hold sensitive information (e.g. S and E matrices), though
+      // it is not modeled as a secure_vector (because it's uint16_t).
+      //
+      // TODO: consider allowing to use secure_vector for this
+      std::vector<uint16_t> m_elements;
```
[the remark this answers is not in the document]
atreiber94: Right, it works. @reneme I don't recall our discussion on this, do we miss anything by just using `secure_vector` here?

src/lib/pubkey/frodokem/frodo_matrix.cpp
```diff
@@ -0,0 +1,545 @@
...
+
+      // Do 8 invocations to fill 8 rows
+      row_generator(a_row.next(sizeof(uint16_t) * constants.n()), static_cast<uint16_t>(i + 0));
+      row_generator(a_row.next(sizeof(uint16_t) * constants.n()), static_cast<uint16_t>(i + 1));
+      row_generator(a_row.next(sizeof(uint16_t) * constants.n()), static_cast<uint16_t>(i + 2));
+      row_generator(a_row.next(sizeof(uint16_t) * constants.n()), static_cast<uint16_t>(i + 3));
+      row_generator(a_row.next(sizeof(uint16_t) * constants.n()), static_cast<uint16_t>(i + 4));
+      row_generator(a_row.next(sizeof(uint16_t) * constants.n()), static_cast<uint16_t>(i + 5));
+      row_generator(a_row.next(sizeof(uint16_t) * constants.n()), static_cast<uint16_t>(i + 6));
+      row_generator(a_row.next(sizeof(uint16_t) * constants.n()), static_cast<uint16_t>(i + 7));
```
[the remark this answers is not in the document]
atreiber94: We tested it and this manual loop unrolling from the reference implementation did indeed give us a significant speed up. Template magic in the end may look just as cluttered as the current state.

src/lib/pubkey/frodokem/frodo_matrix.h
```diff
@@ -0,0 +1,108 @@
...
+            m_dim1(std::get<0>(dimensions)), m_dim2(std::get<1>(dimensions)), m_elements(std::move(elements)) {}
+
+   private:
+      size_t m_dim1;
+      size_t m_dim2;
+
+      // This may hold sensitive information (e.g. S and E matrices), though
+      // it is not modeled as a secure_vector (because it's uint16_t).
+      //
+      // TODO: consider allowing to use secure_vector for this
```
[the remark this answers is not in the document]
atreiber94: TODO is done

src/lib/pubkey/frodokem/frodo_constants.cpp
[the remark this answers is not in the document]
FAlbertDev: Oh, you are right!

doc/dev_ref/oids.rst
```diff
@@ -15,6 +15,24 @@ Values currently assigned are::
   -- { publicKey 5 } previously used for XMSS draft 6
   gost-3410-with-sha256 OBJECT IDENTIFIER ::= { publicKey 6 1 }
 
+  frodokem-shake  OBJECT IDENTIFIER ::= { publicKey 13 }
+  efrodokem-shake OBJECT IDENTIFIER ::= { publicKey 15 }
+  frodokem-aes    OBJECT IDENTIFIER ::= { publicKey 14 }
+  efrodokem-aes   OBJECT IDENTIFIER ::= { publicKey 16 }
```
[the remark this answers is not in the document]
randombit: Let's just give 13 to LMS and shift Frodo to 14-17, then we don't have to deal with a conflict after the fact.

src/lib/pubkey/frodokem/frodo_constants.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include <botan/internal/frodo_constants.h>
+
+#include <botan/xof.h>
+
+namespace Botan {
+
+FrodoKEMConstants::FrodoKEMConstants(FrodoKEMMode mode) : m_mode(mode), m_len_a(128), m_n_bar(8) {
+#if !defined(BOTAN_HAS_AES)
+   BOTAN_ARG_CHECK(!mode.is_aes(), "cannot instantiate AES-based FrodoKEM: This build does not support AES");
+#endif
```
randombit, line 22: Instead of having to do a macro check along with `mode.is_aes()` let's add

`mode.is_available()` (or `is_supported` or ...)

and use it here, in `speed.cpp`, and anywhere else we're doing an explicit `BOTAN_HAS_AES` check.

This also generalizes if we decide to make SHAKE modes optional in the future.

src/tests/test_frodokem.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+   auto modes = std::vector {
+      Botan::FrodoKEMMode::eFrodoKEM1344_SHAKE, Botan::FrodoKEMMode::eFrodoKEM976_SHAKE,
+         Botan::FrodoKEMMode::eFrodoKEM640_SHAKE, Botan::FrodoKEMMode::FrodoKEM1344_SHAKE,
+         Botan::FrodoKEMMode::FrodoKEM976_SHAKE, Botan::FrodoKEMMode::FrodoKEM640_SHAKE,
+   #if defined(BOTAN_HAS_AES)
+         Botan::FrodoKEMMode::eFrodoKEM1344_AES, Botan::FrodoKEMMode::eFrodoKEM976_AES,
+         Botan::FrodoKEMMode::eFrodoKEM640_AES, Botan::FrodoKEMMode::FrodoKEM1344_AES,
+         Botan::FrodoKEMMode::FrodoKEM976_AES, Botan::FrodoKEMMode::FrodoKEM640_AES
+   #endif
```
randombit, line 134: Instead of excluding the modes at compile time in this way lets just `continue` if `!mode.is_supported()`

src/tests/test_frodokem.cpp
```diff
@@ -0,0 +1,224 @@
...
+      }
+
+      std::string algo_name() const override { return "FrodoKEM"; }
+};
+
+}  // namespace
+
+   #if defined(BOTAN_HAS_AES)
+BOTAN_REGISTER_TEST("frodokem", "frodo_kat_tests", Frodo_KAT_Tests);
+   #endif
```
randombit, line 217: We don't run any of the KATs if AES is unavailable??? That puts us in a situation where we are willing to configure and build something and then not test it at all :(

randombit: As with elsewhere we should just skip AES tests at runtime if we don't have support.

src/tests/test_frodokem.cpp
```diff
@@ -0,0 +1,224 @@
+/*
+ * Tests for FrodoKEM ("You SHALL Pass")
```
randombit, line 2: :joy:

src/tests/test_frodokem.cpp
```diff
@@ -0,0 +1,224 @@
...
+      return Botan::FrodoKEMMode::FrodoKEM1344_AES;
+   } else if(header == "eFrodoKEM-640-AES") {
+      return Botan::FrodoKEMMode::eFrodoKEM640_AES;
+   } else if(header == "eFrodoKEM-976-AES") {
+      return Botan::FrodoKEMMode::eFrodoKEM976_AES;
+   } else if(header == "eFrodoKEM-1344-AES") {
+      return Botan::FrodoKEMMode::eFrodoKEM1344_AES;
+   }
+
+   BOTAN_ASSERT_UNREACHABLE();
```
randombit, line 64: This is reachable at runtime if someone flubs the test data. It should probably throw `Test_Error`

src/lib/tls/tls_algos.h
```diff
@@ -114,11 +121,21 @@ enum class Group_Params_Code : uint16_t {
...
+   HYBRID_SECP256R1_eFRODOKEM_640_SHAKE_OQS = 0x2F01,
+   HYBRID_SECP256R1_eFRODOKEM_640_AES_OQS = 0x2F00,
+
    HYBRID_SECP384R1_KYBER_768_R3_OQS = 0x2F3C,
+   HYBRID_SECP384R1_eFRODOKEM_976_SHAKE_OQS = 0x2F03,
+   HYBRID_SECP384R1_eFRODOKEM_976_AES_OQS = 0x2F02,
+
    HYBRID_SECP521R1_KYBER_1024_R3_OQS = 0x2F3D,
+   HYBRID_SECP521R1_eFRODOKEM_1344_SHAKE_OQS = 0x2F05,
+   HYBRID_SECP521R1_eFRODOKEM_1344_AES_OQS = 0x2F04,
```
randombit, line 138: IMO it's a bit more readable if all the eFRODOKEM suites are grouped rather than sorting by NIST security level

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
```diff
@@ -79,13 +97,31 @@ std::vector<size_t> public_value_lengths_for_group(Group_Params group) {
          return {32, 800};
       case Group_Params::HYBRID_X25519_KYBER_768_R3_OQS:
          return {32, 1184};
+      case Group_Params::HYBRID_X25519_eFRODOKEM_640_SHAKE_OQS:
```
randombit, line 100: Not something to address here, but this hardcoding of parameters in something that should theoretically not have to care at all is getting quite dire.

src/lib/pubkey/frodokem/frodo_mode.cpp
```diff
@@ -0,0 +1,113 @@
+/*
+ * FrodoKEM modes and constants
+ * Based on the MIT licensed reference implementation by the designers
+ * (https://github.com/microsoft/PQCrypto-LWEKE/tree/master/src)
```
randombit, line 4: Is anything in this file actually derived from the reference implementation?

src/lib/pubkey/frodokem/frodo_mode.h
```diff
@@ -0,0 +1,78 @@
+/*
+ * FrodoKEM modes and constants
+ * Based on the MIT licensed reference implementation by the designers
+ * (https://github.com/microsoft/PQCrypto-LWEKE/tree/master/src)
```
randombit, line 4: Is anything in this file actually derived from the reference implementation?

src/lib/pubkey/frodokem/frodo_constants.h
```diff
@@ -0,0 +1,105 @@
+/*
+ * FrodoKEM constants
+ * Based on the MIT licensed reference implementation by the designers
+ * (https://github.com/microsoft/PQCrypto-LWEKE/tree/master/src)
```
randombit, line 4: Is anything in this file actually derived from the reference implementation?

src/tests/test_frodokem.cpp
```diff
@@ -0,0 +1,224 @@
...
+   // dependency of the FrodoKEM module and will always be available
+   // when FrodoKEM is enabled in the Botan module configuration.
+   auto xof = Botan::XOF::create_or_throw("SHAKE-256");
+   xof->update(data);
+   return xof->output<std::vector<uint8_t>>(16);
+}
+
+class Frodo_KAT_Tests final : public Text_Based_Test {
+   public:
+      Frodo_KAT_Tests() : Text_Based_Test("pubkey/frodokem_kat.vec", "seed,ss,pk,sk,ct") {}
```
randombit, line 79: Typically variables in the test data files are capitalized (`Seed`) or all caps (`PK`) rather than all lowercase.

src/tests/test_frodokem.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+      Test::Result run_one_test(const std::string& header, const VarMap& vars) override {
+         Test::Result result(Botan::fmt("FrodoKEM KAT {}", header));
+
+         const auto mode = get_mode(header);
```
randombit, line 84: I think this is the spot where we should skip the test if the mode is unavailable.

src/lib/pubkey/frodokem/frodo_mode.h
```diff
@@ -0,0 +1,78 @@
...
+#ifndef BOTAN_FRODOKEM_MODE_H_
+#define BOTAN_FRODOKEM_MODE_H_
+
+#include <botan/asn1_obj.h>
+
+#include <vector>
+
+namespace Botan {
+
+class BOTAN_PUBLIC_API(3, 0) FrodoKEMMode {
```
randombit, line 22: All the `BOTAN_PUBLIC_API` should be `3, 3`

src/lib/pubkey/frodokem/frodokem.cpp
```diff
@@ -0,0 +1,398 @@
...
+   }
+   return *this;
+}
+
+AlgorithmIdentifier FrodoKEM_PublicKey::algorithm_identifier() const {
+   return AlgorithmIdentifier(object_identifier(), AlgorithmIdentifier::USE_EMPTY_PARAM);
+}
+
+OID FrodoKEM_PublicKey::object_identifier() const {
+   return OID::from_string(m_public->constants().mode().to_string());
```
randombit, line 287: Couldn't this use `mode().object_identifier()`?

src/lib/pubkey/frodokem/frodokem.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+         shake.update(consts.encapsulation_domain_separator());
+         shake.update(seed_se);
+
+         const auto s_p =
+            FrodoMatrix::sample(consts,
+                                std::tuple(consts.n_bar(), consts.n()),
+                                shake.output<FrodoSampleR>(sizeof(uint16_t) * consts.n_bar() * consts.n()));
+
+         const auto e_p =
```
randombit, line 121: We have here three repetitions of

```
FrodoMatrix::sample(consts,
    std::tuple(consts.n_bar(), consts.XXX()),
    shake.output<FrodoSampleR>(sizeof(uint16_t) * consts.n_bar() * consts.XXX())
```

I'm not sure if this is the best approach but: at the minimum it seems like a lambda that captures `shake` as a mutable reference and takes `XXX` (either `n_bar` or `n`) as an argument would clean this up.

randombit: There are three similar invocations during decrypt - maybe a `sample` helper on `FrodoMatrix` or a function in an anon namespace in this file would be better than a lambda which would still have to be duplicated.

src/lib/pubkey/frodokem/frodokem.cpp
```diff
@@ -0,0 +1,398 @@
...
+         // and performs a single CT comparison. For convenience we compare the
+         // matrices individually in CT and CT-&& the result.
+         //
+         // Note: b_p and c are unpacked values that are reduced by definition.
+         //       b_pp and c_p are calculated values that need the reduction for
+         //       an unambiguous comparison.
+         b_pp.reduce(consts);
+         c_p.reduce(consts);
+         const bool cmp_b_p_b_pp = b_p.constant_time_compare(b_pp);
+         const bool cmp_c_c_p = c.constant_time_compare(c_p);
```
randombit, line 229: These `constant_time_compare` calls should probably return `CT::Mask` instead of bools

src/lib/pubkey/frodokem/frodo_matrix.cpp
```diff
@@ -0,0 +1,513 @@
...
+
+#include <botan/internal/frodo_matrix.h>
+
+#include <botan/assert.h>
+#include <botan/frodokem.h>
+#include <botan/hex.h>
+#include <botan/mem_ops.h>
+#if defined(BOTAN_HAS_AES)
+   #include <botan/internal/aes.h>
+#endif
```
randombit, line 21: Larger question: is there really a reason

a) To include the AES support at all? It seems like everyone is moving to SHAKE for pqc

b) If AES support is required/useful (for standards conformance or contract reasons, etc): can we just make the AES support also mandatory? 99% of the time you want AES anyway for other reasons, and while you maybe don't need *Frodo-AES*, in the current approach there isn't any way to disable that anyway; if you enable plain old AES, plus Frodo, you get Frodo-AES. This contrasts with how Kyber was done where Kyber-90s is a distinct module that can be toggled off.

src/lib/pubkey/frodokem/frodo_matrix.cpp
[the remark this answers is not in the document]
randombit: Sometimes the stupid ways are best :)

randombit (commented):
This is a first pass review of everything *except* the `frodo_matrix.cpp` file

doc/dev_ref/oids.rst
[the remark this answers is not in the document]
reneme: done

src/lib/tls/tls13_pqc/hybrid_public_key.cpp
reneme: On that note: We should rebase this to `master` and add FrodoKEM to the nightly online interoperability test against the oqs implementation.

src/scripts/dev_tools/gen_frodo_kat.py
reneme: Super modern python would also be to add `encoding='utf8'`. 😨 This script is more documentation than actual functionality, though. So not super important IMO.

src/lib/pubkey/frodokem/frodo_matrix.cpp
reneme: > To include the AES support at all?

In contrast to the NIST drafts, currently the [ISO draft for (e)FrodoKEM](https://frodokem.org/files/FrodoKEM-standard_proposal-20230314.pdf) explicitly specifies the AES variant. Otherwise we would have left it out.

> This contrasts with how Kyber was done where Kyber-90s is a distinct module that can be toggled off.

Initially, we planned to build two sub-modules for FrodoKEM as well. However, in contrast to Kyber, AES in FrodoKEM is somewhat an "add-on", as both FrodoKEM-SHAKE and FrodoKEM-AES depend on SHAKE. 😞 So FrodoKEM without SHAKE doesn't make sense while FrodoKEM without AES still works partially.
You make a fair point, though, that we currently cannot switch off AES-support in FrodoKEM if both `frodokem` and `aes` are enabled in the build.

Let us try to extract the AES portions into a very thin module, see how much complexity that adds and then decide whether its worth it or not.

src/tests/test_frodokem.cpp
reneme: done.

src/lib/pubkey/frodokem/frodo_matrix.cpp
reneme: > Let us try to extract the AES portions into a very thin module, see how much complexity that adds and then decide whether its worth it or not.

Done here: https://github.com/randombit/botan/pull/3679/commits/0694f1db15e0082450fe4b6727935e5f4ea86390
In my opinion the AES sub module doesn't bring much added complexity for giving users the flexibility to switch it off.

src/tests/test_frodokem.cpp
reneme: I introduced an overload for `skip_this_test()` to skip AES modes if they are not enabled in the build.

src/lib/pubkey/frodokem/frodo_constants.cpp
atreiber94: I added this method and used it when possible, but in some instances (row generation, KATs, and `PK_Key_Generation_Test`) compile time checks were still needed.

src/tests/test_frodokem.cpp
atreiber94: There's a distinction between (not) having FrodoKEM-AES and not having AES, since the KATs use AES for seed expansion even for FrodoKEM-SHAKE.

With the new FrodoKEM-AES module we can make the following minimized builds:

- `--enable-modules=frodokem_aes` (Can run FrodoKEM-SHAKE/-AES and all FrodoKEM-SHAKE/-AES tests)
- `--enable-modules=aes,frodokem` (Can run FrodoKEM-SHAKE and all FrodoKEM-SHAKE tests)
-  `--enable-modules=frodokem` (Can run FrodoKEM-SHAKE but not test its KATs, this does not compile right now because we have to use `CTR_DRBG_AES256` in the KATs, which was previously only compiled if `BOTAN_HAS_AES`)

I believe the story is similar in the Kyber implementation, where we skip the KATs  if we do not have AES available.

A possible fix is to manually enter the expanded seed values into the test vector file but I'm not sure if we want to do that or just skip like before/in Kyber?

src/lib/tls/tls_algos.h
atreiber94: Done

src/lib/pubkey/frodokem/frodo_mode.cpp
atreiber94: Fixed

src/lib/pubkey/frodokem/frodo_mode.h
atreiber94: Fixed

src/lib/pubkey/frodokem/frodo_constants.h
atreiber94: Fixed

src/lib/pubkey/frodokem/frodokem.cpp
atreiber94: Fixed

src/lib/pubkey/frodokem/frodokem.cpp
atreiber94: Added a helper function to `FrodoMatrix`. This nicely reduces complexity, thanks for the tip!

src/lib/pubkey/frodokem/frodokem.cpp
atreiber94: Agreed, thanks for pointing this out!

`conditional_copy_mem` now also directly uses the mask instead of the value.

atreiber94:
Thanks for the review! We now addressed all comments except the discussions on the modules and whether to skip KATs without AES support.

src/lib/pubkey/frodokem/frodo_matrix.cpp
reneme: For the sake of consistency, we could certainly do that. Note that you will need to explicitly enable `frodokem_shake` and `frodokem_aes`. There's (AFAIK) no concept of a "virtual" module feature that would enable both variants. Similarly, `kyber` enables the Keccak-based Kyber and `kyber_90s` enables the AES variant. There's an (internal) `kyber_common` with the base implementation that both flavors depend on.

src/tests/test_frodokem.cpp
reneme: @atreiber94 is right. We skip KATs because they depend on the test-helper `CTR_DRBG_AES256` that depends on AES. 
Meanwhile, I fixed the compilation error, @atreiber94 pointed out. By disabling the KAT tests for builds w/o AES. 😭
Note, though, that at least we do roundtrip tests with FrodoKEM if the KATs are disabled. Obviously, that's then not hinged to an independent test vector.

In any case, other PQC algorithms currently have the very same problem. I would suggest addressing this independent of this pull request. For instance, we could create a reduced set of KATs that don't rely on `CTR_DRBG_AES256` to at least have an independent sanity check.

Additionally, and somewhat orthogonally, I suggest to add a generic KAT test class (along the lines of the base classes in `src/tests/test_pubkey.h`) for all PQC-KATs to share. This could replace the individual roundtrip and encoding/decoding tests that we currently have for each algorithm.

reneme: > I suggest to add a generic KAT test class (along the lines of the base classes in `src/tests/test_pubkey.h`) for all PQC-KATs to share.

Done in #3807.

reneme:
Rebased to latest `master` and squashed commit history.

src/tests/test_frodokem.cpp
reneme: With #3807 merged, I closed #3808 and incorporated the relevant adaptions into this pull request.

reneme:
Rebased again, incorporated (the now closed) #3808.

src/lib/pubkey/frodokem/frodokem_common/frodo_matrix.cpp
```diff
@@ -0,0 +1,494 @@
...
+FrodoMatrix FrodoMatrix::mul_bs(const FrodoKEMConstants& constants, const FrodoMatrix& b, const FrodoMatrix& s) {
+   Dimensions dimensions = {constants.n_bar(), constants.n_bar()};
+   auto elements = make_elements_vector(dimensions);
+
+   for(size_t i = 0; i < constants.n_bar(); ++i) {
+      for(size_t j = 0; j < constants.n_bar(); ++j) {
+         auto& current = elements.at(i * constants.n_bar() + j);
+         current = 0;
+         for(size_t k = 0; k < constants.n(); ++k) {
+            current += static_cast<uint16_t>(
```
randombit, line 326: This expression is very hard to read thanks to how the inner comments get formatted. Splitting it out to a couple of variables with leading comments would probably help a lot, eg

```
const uint32_t b_ink = b.elements_at(...);
// Since ...
const uint32_t s_jnk = s.elements_at(...);
current += ...
```

src/lib/pubkey/frodokem/frodokem_common/frodo_matrix.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+   auto elements = make_elements_vector(dimensions);
+   BOTAN_ASSERT_NOMSG(n == elements.size());
+
+   load_le<uint16_t>(elements.data(), r.data(), n);
+
+   for(size_t i = 0; i < n; ++i) {
+      uint32_t sample = 0;                   // Avoid integral promotion
+      uint16_t prnd = elements.at(i) >> 1;   // Drop the least significant bit
+      uint16_t sign = elements.at(i) & 0x1;  // Pick the least significant bit
```
randombit, line 84: These two `uint16_t` can be `const` right?

randombit (approved):
LGTM, left a few style comments on bits that are probably leftover from the C

reneme:
Thanks, @randombit! We addressed the review comments. Currently, we're looking deeper into the application of [the newly added endian load/store mechanics](https://github.com/randombit/botan/pull/3707) in FrodoKEM. But maybe we should leave these for after the Botan 3.3.0 release. What do you think?

reneme:
> But maybe we should leave these for after the Botan 3.3.0 release.

I'm quite certain we want to keep those refactorings for after the release. They require parts of #3869 which we certainly don't want to merge just before the release.

Let's revisit this 3.3.0 is out.

Issue #2876, opened 2022-01-09
Kuznyechik block cipher


huven:
Is there any interest in adding [Kuznyechik](https://en.wikipedia.org/wiki/Kuznyechik) as block cipher to Botan?

Just finished adding it to a project, basically took the (public domain) code from http://cppcrypto.sourceforge.net/ and adapted the interface to match Botan::BlockCipher. Pretty straightforward, seems to work nicely.

This block cipher is used by VeraCrypt (https://www.veracrypt.fr/en/Encryption%20Algorithms.html).

If this would be a welcome addition to Botan, I will do more work to integrate it in Botan's framework and make a pull request.

reneme:
For the record: There seems to be a questionable internal structure hidden in both Streebog and Kuznyechik [1]. FWIW.

[1]

hekkihek:
Yeah, the Wikipedia page I linked to references similar studies.
Still, users tend to want to decrypt whatever they encrypted in the past using this algorithm 😄

huven:
Added in commit 4ec8531e169d1241b1ab6731a2d7eebd25b02d49

PR #3680, opened 2023-08-25
Kuznyechik block cipher

huven:
Add the Kuznyechik block cipher (issue #2876)


huven:
Removed the sync merge commits from this PR, it is now based again on 3.1.1 as intended

src/lib/block/kuznyechik/kuznyechik.h
```diff
@@ -0,0 +1,41 @@
...
+      void decrypt_n(const uint8_t in[], uint8_t out[], size_t blocks) const override;
+
+      void clear() override;
+
+      std::string name() const override { return "Kuznyechik"; }
+
+      std::unique_ptr<BlockCipher> new_object() const override { return std::make_unique<Kuznyechik>(); }
+
+      bool has_keying_material() const override;
+      ~Kuznyechik() override;
```
randombit, line 30: What is the purpose of this destructor?

src/lib/block/kuznyechik/kuznyechik.cpp
```diff
@@ -0,0 +1,4520 @@
...
+   193, 240, 35,  231, 37,  94,  181, 30,  162, 223, 166, 254, 172, 34,  249, 226, 74,  188, 53,  202, 238, 120,
+   5,   107, 81,  225, 89,  163, 242, 113, 86,  17,  106, 137, 148, 101, 140, 187, 119, 60,  123, 40,  171, 210,
+   49,  222, 196, 95,  204, 207, 118, 44,  184, 216, 46,  54,  219, 105, 179, 20,  149, 190, 98,  161, 59,  22,
+   102, 233, 92,  108, 109, 173, 55,  97,  75,  185, 227, 186, 241, 160, 133, 131, 218, 71,  197, 176, 51,  250,
+   150, 111, 110, 194, 246, 80,  255, 93,  169, 142, 23,  27,  151, 125, 236, 88,  247, 31,  251, 124, 9,   13,
+   122, 103, 69,  135, 220, 232, 79,  29,  78,  4,   235, 248, 243, 62,  61,  189, 138, 136, 221, 205, 11,  19,
+   152, 2,   147, 128, 144, 208, 36,  52,  203, 237, 244, 206, 153, 16,  68,  64,  146, 58,  1,   38,  18,  26,
+   72,  104, 245, 129, 139, 199, 214, 32,  10,  8,   0,   76,  215, 116};
+
+const uint64_t T[16][256][2] = {{{0x9680c07a0cd5fbe9, 0xb5aec91b93871119}, {0xce1a17b28ce00819, 0x0af2fef8abfc327b},
```
randombit, line 47: This is 64K of ROM which is really more than I'd like to spend on something that is very marginal in terms of utility. Can we decrease these tables?

src/lib/block/kuznyechik/kuznyechik.cpp
```diff
@@ -0,0 +1,4520 @@
...
+#include <botan/internal/kuznyechik.h>
+#include <botan/internal/loadstor.h>
+
+namespace Botan {
+
+namespace {
+
+namespace Kuznyechik_F {
+
+const unsigned char S[256] = {
```
randombit, line 19: For `S` and `IS` use `alignas(256)`

randombit, line 19: All integer types should be `uintX_t` (`uint8_t` here)

src/lib/block/kuznyechik/kuznyechik.cpp
```diff
@@ -0,0 +1,4520 @@
...
+                                  {0x7372e73091f7182e, 0xe804faf08c271499}, {0xdd81255817ac2c7f, 0xd7b3e6fbd4863ab2},
+                                  {0xe034d7a13509b1ad, 0xd1d664608616085a}, {0x590ae18b6498a406, 0x468f99e21c88f6b9},
+                                  {0xc8bd26e48c7a726b, 0x801736f29c304ba2}, {0x9c4d2a51a5a4c93b, 0x1fc2f0d67fad4ce2},
+                                  {0x32bee83923fffd6a, 0x2075ecdd270c62c9}, {0x2782eb85b829a37e, 0x77d13cd46fba13d9},
+                                  {0x3f4405076eb9e23c, 0xf92fb31bd8199330}, {0x304f1fc76fe38284, 0xdf3fdd5dad85c311},
+                                  {0x7aa927314489f967, 0x0ccac7f5a4e36413}, {0x8a19446c546036b6, 0x2909e81ff8372d46},
+                                  {0x223a5cc5061f8316, 0xd460a757fdce23c6}, {0x825b1e12a7100988, 0x53e22c5a9556eca0},
+                                  {0x88e8b392187c4958, 0xd643d99f72be8c9e}, {0x6107a432f80b19bc, 0xe35b80fadc6cf44e},
+                                  {0x8fa1f32cea5616de, 0x8ab8739c6a431d59}, {0x46854fb74022bac2, 0x948abc2eb3d6e797},
+                                  {0xaad2ef571e63ca4e, 0x02237ec88f
```
randombit, line 4163: These should not be `get_byte_var` - that is only for use if the rotation amount is not known at compile time. Use `get_byte<N>`

src/tests/data/block/kuznyechik.vec
```diff
@@ -0,0 +1,7 @@
+[Kuznyechik]
+
+# RFC 7801
+
+Key = 8899AABBCCDDEEFF0011223344556677FEDCBA98765432100123456789ABCDEF
+In = 1122334455667700FFEEDDCCBBAA9988
+Out = 7F679D90BEBC24305A468D42B9D4EDCD
```
randombit, line 7: I would definitely want more than a single test for this. I would imagine at most 10% of the `T` table is used by this one particular vector.

If no tests are available from any other official sources (in which case ?????) then create random ones using another implementation.

src/lib/block/kuznyechik/kuznyechik.cpp
randombit, line 4355: These memcpys look strange/wrong. Has this been tested in a big endian system?

randombit, line 4207: Should be `uint8_t`

randombit (commented):
So here is a first pass review. In general it looks pretty good.

I think there is a larger question here that must be addressed first, which is if we really want to accept this at all.

On the technical merits this has no advantage over AES or Serpent or ... - but certainly one can say the same about similar vanity ciphers like ARIA or Camellia, so that's not in and of itself a disqualifier.

The hidden sbox structure (https://who.paris.inria.fr/Leo.Perrin/pi.html) is sketchy as hell considering the designers seemingly still refuse to acknowledge that the sbox is not random as they had claimed.

In general I'm pretty tolerant of the weird national stuff from countries in the top ~10 GDP, and less so lower down; for example the similar ciphers from Belarus and Ukraine are almost certainly never going to be supported (absent some pretty significant economic changes). OTOH while on a GDP list Russia is number 11 and South Korea is 12, SK's GDP is ~10% from software/tech while Russia's is perhaps 1%. So the total set of software developers who might possibly use this seems small relative to similar vanity ciphers. (I'm assuming nobody sane would ever use this cipher unless forced to by gov mandate.) And since every cipher we support is an ongoing maintenance cost on us, I am hesitant.

Is there a specific use case you have in mind here?

huven:
> I think there is a larger question here that must be addressed first, which is if we really want to accept this at all.
> 
> On the technical merits this has no advantage over AES or Serpent or ... - but certainly one can say the same about similar vanity ciphers like ARIA or Camellia, so that's not in and of itself a disqualifier.
> 
> The hidden sbox structure (https://who.paris.inria.fr/Leo.Perrin/pi.html) is sketchy as hell considering the designers seemingly still refuse to acknowledge that the sbox is not random as they had claimed.
> 
> In general I'm pretty tolerant of the weird national stuff from countries in the top ~10 GDP, and less so lower down; for example the similar ciphers from Belarus and Ukraine are almost certainly never going to be supported (absent some pretty significant economic changes). OTOH while on a GDP list Russia is number 11 and South Korea is 12, SK's GDP is ~10% from software/tech while Russia's is perhaps 1%. So the total set of software developers who might possibly use this seems small relative to similar vanity ciphers. (I'm assuming nobody sane would ever use this cipher unless forced to by gov mandate.) And since every cipher we support is an ongoing maintenance cost on us, I am hesitant.
> 
> Is there a specific use case you have in mind here?

Yes, this cipher is needed for an app (https://disk-decipher.app/) that provides support for VeraCrypt containers. This is one of the ciphers offered by VeraCrypt (https://veracrypt.fr/en/Encryption%20Algorithms.html), so I need this cipher for full compatibility with VeraCrypt.

Previously I maintained this cipher out-of-tree. The upgrade to v3 removed the required API for that, so I needed to move the implementation inside the library.

src/lib/block/kuznyechik/kuznyechik.h
huven: The destructor takes care of clearing the m_rke and m_rkd ivars.

src/lib/block/kuznyechik/kuznyechik.cpp
huven: Changed. Will be in the next update to this PR.

huven: Changed. Will be in the next update to this PR.

src/lib/block/kuznyechik/kuznyechik.cpp
huven: Changed. Will be in the next update to this PR.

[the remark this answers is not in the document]
huven: Changed to uint8_t (as they are limited by 32)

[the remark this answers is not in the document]
huven: Removed. The code was to generate the tables.

huven: Changed. Will be in the next update to this PR.

huven: Not by me. Do the RISC tests of this PR test this? The original code also has this strangeness, I made it more explicit by converting it to memcpy.

randombit: No, up until #3686 we did not have a CI target for big endian systems. (Previously there was a MIPS build but that had to be disabled due to toolchain limitations on Ubuntu 22.04) Can you rebase to pick this up?

huven: Will do, if any BE issue turns up, I will address it before the next update to this PR.

src/tests/data/block/kuznyechik.vec
huven: Created 64 random key/plain-text pairs and encrypted them with the 8-bit (slow) implementation available at https://github.com/mjosaarinen/kuznechik

Added the resulting vectors:

Kuznyechik ran 1235 tests in 2.73 msec all ok

src/lib/block/kuznyechik/kuznyechik.cpp
huven: Ahh, the S390x checks failed nicely, will address that in the next commit.

randombit: The utils in `loadstor.h` handle most of this (eg `load_le<uint64_t>`)

huven:
The last commit completes the changes for your first pass review.

If there are any other changes needed, just let me know.

The more general question if you want to include this cipher into the library at all is something only you can decide. For full compatibility with VeraCrypt it is required.

No hard feelings if you reject this PR. In that case I will keep it private for my own use.

src/lib/block/kuznyechik/kuznyechik.cpp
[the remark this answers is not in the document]
randombit: This is otherwise not the convention we use in the library so I'd prefer variables renamed with decimal index

randombit (changes requested):
Thanks for addressing the review comments.

This is definitely right on the edge of what I'd be willing to include but you present a realistic use case where it's needed so I suppose I'm ok with merging and we'll slap some big warnings in the documentation to never use it unless you really have to.

Can you rename the variables that are currently using binary notation? That's not the convention we use in the library and it's best to keep things as consistent as possible.

reneme:
@huven Please rebase to the latest `master`. #3684 and #3681 migrated the virtual base-class methods from ptr-length-parameters to a `std::span`.

huven:
> @huven Please rebase to the latest `master`. #3684 and #3681 migrated the virtual base-class methods from ptr-length-parameters to a `std::span`.

Thanks for pointing that out, just committed the necessary changes to the PR.

randombit:
@huven Sorry one final hurdle, can you please rebase and squash to a single commit? Thanks!

huven:
> @huven Sorry one final hurdle, can you please rebase and squash to a single commit? Thanks!

No problem, I was expecting that, I left the previous commits separately for easy review.

Just did the rebase/squash/force push.

src/lib/block/kuznyechik/kuznyechik.cpp
```diff
@@ -0,0 +1,4394 @@
+/*
+* GOST R 34.12-2015: Block Cipher "Kuznyechik"
```
randombit, line 2: Please add a Copyright notice from yourself following the usual convention

src/lib/block/kuznyechik/kuznyechik.cpp
```diff
@@ -0,0 +1,4394 @@
...
+* RFC 7801
+*
+* This code is written by kerukuro for cppcrypto library (http://cppcrypto.sourceforge.net/)
+* and released into public domain.
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/mem_ops.h>
+#include <botan/internal/kuznyechik.h>
```
randombit, line 12: Include of `kuznyechik.h` should be the first include and separated by a newline from the remainder of the headers (so clang-format doesn't reorder it)

randombit (changes requested):
Sorry, two more very minor things I noticed in a final review. Almost there!

src/lib/block/kuznyechik/kuznyechik.cpp
huven: Done

src/lib/block/kuznyechik/kuznyechik.cpp
huven: Also done

randombit (approved):

PR #3681, opened 2023-08-26
std::span for Buffered_Computation

reneme:
### Pull Request Dependencies

This interferes with the Keccak refactoring, KMAC and the work done in `MDx_HashFunction`. It shouldn't be too hard to fix those conflicts in any order, though.

### Description

This is upgrading the internal virtual methods of `Buffered_Computation` (i.e. `add_data()` and `final_result()`) to use `std::span<>`. Also some of the implementations in the "hash" and "mac" modules benefit from `BufferSlicer` and `BufferStuffer` to abstract some of the pointer/offset mangling.

... courtesy of a long boring flight. :)

### Future Work

* make `buffer_insert`, `copy_mem`, `xor_buf`, and friends `std::span`-aware to further improve on the expressiveness of those methods
* look into std::spanifying `MDx_HashFunction` (see also #3550)
* `Symmetric_Algorithm::key_schedule()` (see #3684)
* A bit of a long-shot: Maybe we could spike an abstraction over the recurring pattern of the buffer alignment code: i.e. "process data until alignment is reached, process data in buffer alignment, process remaining bytes causing disalignment"

### DIFF
--- a/src/lib/base/buf_comp.cpp
+++ b/src/lib/base/buf_comp.cpp
@@ -13,37 +13,37 @@ namespace Botan {
 void Buffered_Computation::update_be(uint16_t val) {
    uint8_t inb[sizeof(val)];
    store_be(val, inb);
-   add_data(inb, sizeof(inb));
+   add_data({inb, sizeof(inb)});
 }
 
 void Buffered_Computation::update_be(uint32_t val) {
    uint8_t inb[sizeof(val)];
    store_be(val, inb);
-   add_data(inb, sizeof(inb));
+   add_data({inb, sizeof(inb)});
 }
 
 void Buffered_Computation::update_be(uint64_t val) {
    uint8_t inb[sizeof(val)];
    store_be(val, inb);
-   add_data(inb, sizeof(inb));
+   add_data({inb, sizeof(inb)});
 }
 
 void Buffered_Computation::update_le(uint16_t val) {
    uint8_t inb[sizeof(val)];
    store_le(val, inb);
-   add_data(inb, sizeof(inb));
+   add_data({inb, sizeof(inb)});
 }
 
 void Buffered_Computation::update_le(uint32_t val) {
    uint8_t inb[sizeof(val)];
    store_le(val, inb);
-   add_data(inb, sizeof(inb));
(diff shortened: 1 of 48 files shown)


randombit:
>  Maybe we could spike an abstraction over the recurring pattern of the buffer alignment code: i.e. "process data until alignment is reached, process data in buffer alignment, process remaining bytes causing disalignment"

That would be pretty nice. I've made several attempts over the years at this and never got anything that worked (ie was actually cleaner/easier to understand and did not introduce perf issues somwhere)

reneme:
> I've made several attempts over the years at this and never got anything that worked

Just a pure gut feeling: it might be a good playground for C++20 coroutines.

randombit:
I would be concerned about using coroutines wrt baremetal systems, I don't know what degree of runtime support this requires.

randombit (commented):
Seems basically fine. I checked and it does not seem to introduce any performance regressions.

I wonder for `final` if we should be checking the length of the span. This would be a new check in that the current approach just writes to the provided pointer and if you don't provide enough data we'll overwrite something we shouldn't. And I guess we can continue that here; it just feels a little unfriendly to do so given that we do know the length of the output span.

reneme:
> I wonder for final if we should be checking the length of the span.

I deliberately didn't add checks to the `private virtual` `add_data()` and `final_result()` implementations. They should all be invoked via the public `update()` and `final()` calls in `Buffered_Computation`. Those public wrappers have asserts for the lengths where possible. Or did I miss the check somewhere? 

Maybe we could replace the asserts with `BOTAN_ARG_CHECK`, though.

randombit:
> They should all be invoked via the public update() and final() calls in Buffered_Computation. Those public wrappers have asserts for the lengths where possible.

Sorry I missed this. Looks fine then.

> Maybe we could replace the asserts with BOTAN_ARG_CHECK, though.

Yeah concur on this. `BOTAN_ASSERT` *should* be only for conditions where we believe it could never fail but want to double check. But `BOTAN_ASSERT` has been around a lot longer than `BOTAN_ARG_CHECK` etc and so we have a number of cases where we are using assert to check user inputs.

randombit (approved):

PR #3684, opened 2023-08-30
std::span for SymmetricAlgorithm::key_schedule()

reneme:
### Pull Request Dependencies

* #3681

Also this probably interferes with the Keccak refactoring, KMAC and the work done in `MDx_HashFunction`. It shouldn't be too hard to fix those conflicts in any order, though.

### Description

This upgrades the internal virtual method of `SymmetricAlgorithm::key_schedule()` to use `std::span<>`.


randombit (approved):
Only reviewed the second commit, LGTM thanks!

Issue #2607, opened 2021-01-26
Enable getrandom() by default?


securitykernel:
Support for ` getrandom()` was introduced in 9a41b24157260eb9830e6a441c1131d6be09b3df, but disabled by default as it required a newer kernel and glibc. Two years down the road, would it make sense to enable it by default now?

Additionally, in recent Linux kernels starting with 5.6, there have been some major changes to the RNG ([source](https://www.bsi.bund.de/DE/Publikationen/Studien/LinuxRNG/index_htm.html)):

* Unrestricted generation of random numbers is available with /dev/urandom and the in-kernel function
get_random_bytes. This implies that irrespective whether the entropy pool or the ChaCha20 DRNG
received sufficient entropy, random data is generated.
* When accessing /dev/random, random numbers are only generated when the entropy pool or the
ChaCha20 DRNG received at least 128 bits of initial entropy. After reaching that threshold of 128 bits of
entropy once, /dev/random will operate non-blocking for the lifetime of the system and thus operate
identically to /dev/urandom.

In addition, the Linux kernel offers the getrandom system call documented by its respective man page
which provides access to the Linux-RNG as follows:

* When invoking getrandom where the flag field is zero, the system call accesses the ChaCha20 DRNG
identically to /dev/random.
* When invoking getrandom with a flag of GRND_INSECURE, the system call behaves like
/dev/urandom.
* The flag GRND_RANDOM is currently unused.

So it seems the blocking behaviour of /dev/random, one of its main criticism, was changed to just address the ` not enough entropy during early system startup` problem and once seeded, never blocks anymore, so behaves like /dev/urandom from then on. ` getrandom()` was changed to access /dev/random by default with flags set to `0`, which Botan does. So to me it seems ` getrandom()`  is the best choice for a `System_RNG` on Linux these days.

randombit:
The syscall itself was introduced in Linux 3.7 [Dec 2012] and even 2 years ago that was old enough to rely on. However the libc wrappers were only added in glibc 2.25 [Feb 2017] and musl 1.1.20 [Sep 2018], which made switching in Sept 2018 a problem. However consulting https://repology.org/project/glibc/versions and https://repology.org/project/musl/versions it seems most active distros have a recent enough version, with the perhaps notable exception of RHEL7. However the fallback for old systems is easy.

So, seems good to me particularly for 3.0. I'm less certain about doing this on 2.x branch but suppose it's fine there if you see the need.

securitykernel:
[...]

Issue #1322, opened 2017-11-30
getentropy on Linux?


lambdafu:
`src/lib/entropy/getentropy/info.txt` seems to suggest getentropy is only used on openbsd. The system call is also available on Linux systems now, and it's obviously a nice improvement over /dev/urandom (which, if I interpret things correctly, is currently used by default on Linux). Are there any plans to enable getentropy for Linux and make it the new default? Thanks!

randombit:
No immediate plans. The main issue is that support was only added in glibc in 2.25, released earlier this year. So it is not available in most distros, basically right now only in rolling release distros like Arch and Gentoo.

And, we cannot really enable it conditionally right now, because of a limitation of the build system. Basically the capabilities of the OS are static (specified in files in `src/build-data/os`) rather than detected based on the target system.

A few options:
 a) Add build time feature detection ala autoconf. I have this on the agenda but it's a big project and I don't have a huge amount of time for botan right now.
 b) Add an option that lets a user enable features that we don't know about. Eg `./configure.py --enable-os-features=getentropy`. This would be fairly simple to implement, and useful in several different scenarios.
 c) Ignore the issue until the "stable" Linux distros catch up, circa 2019, and we can just use it knowing that we probably have a recent enough glibc. Worth mentioning that musl still doesn't implement getrandom or getentropy :(

For OpenBSD it works out because getentropy has been available there for a while, and (IIRC) OpenBSD only supports their 2 most recent releases, so basically all supported versions of OpenBSD include the syscall.

In short term (b) is probably right choice.

lambdafu:
Thanks for the explanations, which are very helpful! In this case it's more important to have an answer than an immediate solution. As a workaround, I think concerned users of botan could simple call getentropy (or even the getrandom system call directly) during initialization.  Once it returns one byte, there would be no difference anyway.

randombit:
With #1378 merged `getentropy` can be enabled at build time with `--enable-os-features=getentropy` so closing this.

mouse07410:
I want to add that `getentropy()` is also available on MacOS starting with 10.12 (Sierra). But it requires adding `#include <sys/random.h>` to `src/lib/entropy/entropy_src.h` file.

Could you add it?

mouse07410:
This is the patch I'm thinking of:
```
diff --git a/src/lib/entropy/entropy_src.h b/src/lib/entropy/entropy_src.h
index 56e5bd53e..90aacd4ce 100644
[...]

PR #3685, opened 2023-08-31
Use syscall for getrandom

uilianries:
The [getrandom](https://man7.org/linux/man-pages/man2/getrandom.2.html) is actually a syscall that can be call through `syscall` method. The advantage here is using older glibc versions that does not have `getrandom()` exposed in headers, but available as pure syscall.

In case not accepting it, due Botan progression of always opting for newer compilers, is still possible to add a new option like `with-os-syscall`? 

Related to issues #2607 and #1322

### DIFF
--- a/src/lib/rng/system_rng/system_rng.cpp
+++ b/src/lib/rng/system_rng/system_rng.cpp
@@ -26,6 +26,7 @@
 #elif defined(BOTAN_TARGET_OS_HAS_GETRANDOM)
    #include <errno.h>
    #include <sys/random.h>
+   #include <sys/syscall.h>
 #elif defined(BOTAN_TARGET_OS_HAS_DEV_RANDOM)
    #include <errno.h>
    #include <fcntl.h>
@@ -211,7 +212,7 @@ class System_RNG_Impl final : public RandomNumberGenerator {
          uint8_t* buf = output.data();
          size_t len = output.size();
          while(len > 0) {
-            const ssize_t got = ::getrandom(buf, len, flags);
+            const ssize_t got = ::syscall(SYS_getrandom, buf, len, flags);
 
             if(got < 0) {
                if(errno == EINTR) {


src/lib/rng/system_rng/system_rng.cpp
```diff
@@ -211,7 +212,7 @@ class System_RNG_Impl final : public RandomNumberGenerator {
          uint8_t* buf = output.data();
          size_t len = output.size();
          while(len > 0) {
-            const ssize_t got = ::getrandom(buf, len, flags);
```
securitykernel, line 214: You accidentally removed the variable name `got`, re-adding it should fix the CI.

src/lib/rng/system_rng/system_rng.cpp
```diff
@@ -211,7 +212,7 @@ class System_RNG_Impl final : public RandomNumberGenerator {
...
+            const ssize_t = ::syscall(SYS_getrandom, buf, len, flags);
```
securitykernel, line 215: You accidentally removed the variable name got, re-adding it should fix the CI.

uilianries, line 215:
```suggestion
            const ssize_t got = ::syscall(SYS_getrandom, buf, len, flags);
```

uilianries: Thank you! Fixed.

randombit:
So, this doesn't quite work because we also use `getrandom` on other systems that support it (eg FreeBSD)

Due to basically an accident of the code, we always prefer `arc4random_buf` vs `getrandom` so this would still compile but we need to figure out the correct approach. That might be as simple as flagging `getrandom` support only on Linux so we can just use the syscall interface there.

securitykernel:
But FreeBSD seams to provide the same syscall as Linux ([table](https://alfonsosiciliano.gitlab.io/posts/2021-01-02-freebsd-system-calls-table.html), [code](https://cgit.freebsd.org/src/tree/sys/kern/sys_getrandom.c)). The same for DragonFly BSD ([code](https://github.com/DragonFlyBSD/DragonFlyBSD/blob/master/sys/sys/syscall.h#L388C17-L388C17)). That are all OSes we marked as supporting getrandom.

randombit:
Same syscall, sure, but do they have the same interface for calling raw syscalls? Ie would the code in this PR compile on FreeBSD?

securitykernel:
Yeah, the same interface: [FreeBSD syscall()](https://man.freebsd.org/cgi/man.cgi?query=syscall&sektion=2&format=html) | [Linux syscall()](https://man7.org/linux/man-pages/man2/syscall.2.html). I compiled it and even ran `botan rng`, it works.

uilianries:
@randombit Done, I just rebased and squashed the PR.

randombit (approved):

PR #3687, opened 2023-09-08
Implement Certificate Authority serialization

mateuszb:
Implement Certificate Authority serialization as part of the TLS 1.3 client certificate request and authentication flow.

### DIFF
--- a/src/lib/tls/tls_extensions.cpp
+++ b/src/lib/tls/tls_extensions.cpp
@@ -4,6 +4,7 @@
 *     2016 Juraj Somorovsky
 *     2021 Elektrobit Automotive GmbH
 *     2022 René Meusel, Hannes Rantzsch - neXenio GmbH
+*     2023 Mateusz Berezecki
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
@@ -11,6 +12,7 @@
 #include <botan/tls_extensions.h>
 
 #include <botan/ber_dec.h>
+#include <botan/der_enc.h>
 #include <botan/tls_exceptn.h>
 #include <botan/tls_policy.h>
 #include <botan/internal/stl_util.h>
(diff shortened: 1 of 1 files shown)


src/lib/tls/tls_extensions.cpp
```diff
@@ -761,7 +764,27 @@ PSK_Key_Exchange_Modes::PSK_Key_Exchange_Modes(TLS_Data_Reader& reader, uint16_t
...
+       bytes.emplace_back();
+       auto old_sz = bytes.size();
+       dn.encode_into(encoder);
+       auto sz = static_cast<uint16_t>(bytes.size() - old_sz);
+
+       bytes[sz_idx] = get_byte<0>(sz);
+       bytes[sz_idx+1] = get_byte<1>(sz);
+   }
+
+   BOTAN_ASSERT_NOMSG(bytes.size() >= 2 && bytes.size() <= std::numeric_limits<uint16_t>::max());
```
reneme, line 782: `append_tls_length_value()` also checks the size limits and throws if things don't fit. Once you moved your application to use it, this assertion should be removed.

As a general remark: We use assertions to validate things that should never happen in any typical use case. However, this might be triggered by an application that configured too many trusted DNs. It should therefore not be an assertion but a sanity check.

reneme (commented):
Thank you for the patch! This particular bit must have fallen through, due to a lack of test coverage in both the RFC8448-based tests and BoGo.

Please use our TLS serialization helper function as outlined below. That should make your code quite a bit simpler, roughly along those lines:

```C++
std::vector<uint8_t> out;
std::vector<uint8_t> dn_list;

for(const auto& dn : m_distinguished_names) {
   std::vector<uint8_t> encoded_dn;
   dn.encode_into(encoded_dn);
   append_tls_length_value(dn_list, encoded_dn, 2);
}

append_tls_length_value(out, dn_list, 2);
```

Admittedly, that's less memory efficient, but easier to reason about. Improving the TLS marshaling helpers has been on my list for a while now. 😞

Also: Our CI checks consistency of the `.clang-format` rules. That's why your patch currently fails to build. Please have a quick read through [our (short) style guide](https://botan.randombit.net/handbook/dev_ref/contributing.html#style-conventions).

reneme (approved):
Thanks! Looks good to me.

I successfully smoke-tested that locally against an OpenSSL client.
@FAlbertDev Any chance we could get coverage for that from TLS Anvil?

This marshalling code is needed to pass "accepted client-auth CA certificate distinguished names" from a server to a client in a `CertificateRequest` message. For completeness: Also `ClientHello` can contain this extension to communicate CAs that a client is willing to accept, though we don't implement that at the moment.

randombit (approved):
LGTM, thanks @mateuszb

FAlbertDev:
> I successfully smoke-tested that locally against an OpenSSL client.
@FAlbertDev Any chance we could get coverage for that from TLS Anvil?

It seems that the Certificate Authorities Extension is currently not supported by TLS-Attacker and TLS-Anvil. Of course, it could be implemented, but not without additional work.

PR #3688, opened 2023-09-08
Use the raw getrandom syscall only if we detect an old glibc

randombit:
As invoking the raw syscall is not instrumented by MSan, it causes false positives of using uninitialized memory.

OSS-Fuzz 62131

FYI @uilianries


uilianries (approved):
Excellent, thank you for spotting it!

reneme (approved):

PR #3689, opened 2023-09-08
Feature: KMAC-128/256, NIST SP.800-185

reneme:
### Pull Request Dependencies

* ~#3671~

### Description

This finishes up the KMAC implementation based on the cSHAKE-128/256 added in #3671. The test vectors are based on [these NIST samples](https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/KMAC_samples.pdf).

@falko-strenzke Feel free to rebase your pull request on top of #3671 and pull these changes. I don't want to take the principal authorship away from you. :)


falko-strenzke:
> @falko-strenzke Feel free to rebase your pull request on top of #3671 and pull these changes. I don't want to take the principal authorship away from you. :)

No problem, my commit is in the history anyway, so I suggest we just continue with this PR and close https://github.com/randombit/botan/pull/3570

reneme:
Rebased and adapted after std::span related PRs got merged.

reneme:
Rebased to master after #3671 got merged.

src/lib/mac/kmac/kmac.h
```diff
@@ -0,0 +1,71 @@
...
+
+#include <botan/mac.h>
+
+namespace Botan {
+
+class cSHAKE_XOF;
+
+class KMAC : public MessageAuthenticationCode {
+   protected:
+      KMAC(std::unique_ptr<cSHAKE_XOF> cshake, size_t output_byte_length);
```
randombit, line 21: In the implementation this is treated as a bit length

src/lib/mac/kmac/kmac.cpp
```diff
@@ -0,0 +1,96 @@
...
+   m_cshake->output(output.first(output_length()));
+   m_cshake->clear();
+}
+
+void KMAC::key_schedule(std::span<const uint8_t> key) {
+   clear();
+   m_key.insert(m_key.end(), key.begin(), key.end());
+}
+
+KMAC128::KMAC128(size_t output_byte_length) : KMAC(std::make_unique<cSHAKE_128_XOF>("KMAC"), output_byte_length) {}
```
randombit, line 76: I think this is actually `output_bit_length` (ditto for KMAC256)

src/lib/mac/kmac/kmac.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+void KMAC::start_msg(const uint8_t nonce[], size_t nonce_len) {
+   assert_key_material_set(!m_key.empty());
+   m_cshake->start({nonce, nonce_len}, {});
+   keccak_absorb_padded_strings_encoding(*m_cshake, m_cshake->block_size(), m_key);
+}
+
+KMAC::KMAC(std::unique_ptr<cSHAKE_XOF> cshake, size_t output_bit_length) :
+      m_output_bit_length(output_bit_length), m_cshake(std::move(cshake)) {
+   BOTAN_ARG_CHECK(m_output_bit_length % 8 == 0, "KMAC output length must be full bytes");
```
randombit, line 54: Should we reject `output_bit_length == 0`?

reneme: Yeah, makes sense.

src/lib/mac/kmac/kmac.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+void KMAC::add_data(std::span<const uint8_t> data) {
+   assert_key_material_set(!m_key.empty());
+   m_cshake->update(data);
+}
+
+void KMAC::final_result(std::span<uint8_t> output) {
+   assert_key_material_set(!m_key.empty());
+   std::array<uint8_t, keccak_max_int_encoding_size()> encoded_output_length_buffer;
+   m_cshake->update(keccak_int_right_encode(encoded_output_length_buffer, m_output_bit_length));
```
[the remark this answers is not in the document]
reneme: That would force us to include `keccak_helpers.h` in the header (for `keccak_max_int_encoding_size()`). I don't think that's worth it either.

src/lib/mac/kmac/kmac.cpp
```diff
@@ -0,0 +1,96 @@
...
+bool KMAC::has_keying_material() const {
+   return !m_key.empty();
+}
+
+std::string KMAC::provider() const {
+   return m_cshake->provider();
+}
+
+void KMAC::start_msg(const uint8_t nonce[], size_t nonce_len) {
+   assert_key_material_set(!m_key.empty());
```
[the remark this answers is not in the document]
reneme: > I would lean to just require a non-empty key

I agree. FWIW: All other MACs require the key to be non-empty as well. Except HMAC that allows for keys between 0 and 4k.

src/lib/mac/kmac/kmac.cpp
```diff
@@ -0,0 +1,96 @@
...
+   std::array<uint8_t, keccak_max_int_encoding_size()> encoded_output_length_buffer;
+   m_cshake->update(keccak_int_right_encode(encoded_output_length_buffer, m_output_bit_length));
+   m_cshake->output(output.first(output_length()));
+   m_cshake->clear();
+}
+
+void KMAC::key_schedule(std::span<const uint8_t> key) {
+   clear();
+   m_key.insert(m_key.end(), key.begin(), key.end());
```
reneme, line 73: Two things:

1) I think this should use `.assign()` (or `assign_range()` in C++23):

```suggestion
   m_key.assign(key.begin(), key.end());
```

2. This should check that the key is non-empty (for the same reason @randombit [pointed out](https://github.com/randombit/botan/pull/3689/files#r1326080790))
Though, I think, we shouldn't actually check the upper bound of `key_spec()`, no problem if a user decides that they need more.

src/lib/mac/kmac/kmac.cpp
```diff
@@ -0,0 +1,96 @@
...
+}
```
reneme, line 74: Side note: To allow for zero-length strings without an extra boolean flag, we could pre-compute the "padded string encoding" and store that (rather than just the actual key). This encoding would result in a non-empty buffer even for empty keys. It might also give a slight performance benefit when the same MAC object is used for many messages.

reneme: The two commits I just added implement it like suggested (and therefore allow for zero-length keys).

For the sake of some C++20 tech talk: Have a look at https://github.com/randombit/botan/pull/3689/commits/73d27329b5580e9553fdb6d7a178334a8c54eb9b which enables `keccak_absorb_padded_strings_encoding()` to deal with a byte vector as a data-sink (additionally to a XOF).

Now, I'm not saying that this is the most beautiful solution ever. But just imagine the amount of infrastructure needed to pull this off before C++20 concepts. Let alone as a zero-cost abstraction. :heart:

src/lib/mac/kmac/kmac.cpp
reneme: Obsoleted by: https://github.com/randombit/botan/pull/3689#discussion_r1326265225

src/lib/mac/kmac/kmac.cpp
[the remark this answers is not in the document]
reneme: In the end I opted for allow zero-length keys. After all, SP.800-185 specifies them as acceptable. See: https://github.com/randombit/botan/pull/3689#discussion_r1326265225

src/lib/mac/kmac/kmac.cpp
```diff
@@ -0,0 +1,96 @@
...
+   return m_output_bit_length / 8;
+}
+
+Key_Length_Specification KMAC::key_spec() const {
+   // KMAC supports key lengths from zero up to 2²⁰⁴⁰ (2^(2040)) bits:
+   // https://nvlpubs.nist.gov/nistpubs/specialpublications/nist.sp.800-185.pdf#page=28
+   //
+   // However, we restrict the key length to 64 bytes in order to avoid allocation of overly
+   // large memory stretches when client code works with the maximal key length.
+   return Key_Length_Specification(0, 64);
```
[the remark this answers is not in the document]
randombit: 64 bytes is arbitrary but also 512 bits seems more than sufficient as a MAC key

HMAC allows 0..4095 because in PBKDF2 the HMAC "key" is the passphrase and we don't want to artificially limit the length of passphrases that people can use. This was also required for the TLS 1.2 PRF when using eg DH with large groups since there the shared secret will be the size of the field elements and TLS PRF uses the shared secret as an HMAC key.

src/lib/mac/kmac/kmac.cpp
randombit: I had no idea you could use `||` to combine concepts!

src/lib/mac/kmac/kmac.cpp
```diff
@@ -0,0 +1,97 @@
...
+}
+
+std::string KMAC::provider() const {
+   return m_cshake->provider();
+}
+
+void KMAC::start_msg(const uint8_t nonce[], size_t nonce_len) {
+   assert_key_material_set();
+   m_cshake->start({nonce, nonce_len}, {});
+   m_cshake->update(m_encoded_key);
```
randombit, line 49: I don't think this works given the way the MAC contract is specified. We currently only require `start_msg` if you are providing a nonce. But, the nonce is optional with KMAC. So someone using KMAC without a nonce, and just doing init-update-final, will end up with unkeyed KMAC.

This is missed by the tests because `test_mac.cpp` currently always calls start even if the `iv` is empty. This should be extended so we also (when the iv/nonce is not in use) invoking the MAC without calling `start`. (I'll do a PR)

reneme: Yikes! 😶

randombit: #3699

reneme: I'll rebase and fix after #3699 is merged (potentially tomorrow morning).

reneme: Rebased and fixed. On that note: `cSHAKE_XOF` has the very same issue: If no-one calls `start()` it won't absorb the prefix (i.e. `N` and `S`) and generate wrong results.

Now, with the XOF API still being fresh, we could simply enforce that any XOF must be `start()`ed, but I think that would be confusing and error prone for things like vanilla SHAKE. I'll fix this in an independent PR.

randombit (approved):

PR #3693, opened 2023-09-12
Refactor: AlignmentBuffer<> helper for block-oriented Hashes

reneme:
### Introduction and Motivation

[As promised](https://github.com/randombit/botan/pull/3681#issuecomment-1705313378) I am looking into a way to simplify the alignment code in block-oriented algorithms. To my mind, there is quite a bit of code duplication in many implementations. Also, its fairly critical code that handles raw memory, offsets and internal state that could easily go wrong.

In fact, I even found an issue in the existing code, while working on this. Below is [the `::add_data()` implementation for `MDx_HashFunction` in Botan 3.1.1](https://github.com/randombit/botan/blob/f60608b8818c7bb8579fe797122ed6116f4134af/src/lib/hash/mdx_hash/mdx_hash.cpp#L49-L75). Latest master is slightly different but contains the same issue.

Let's assume this is called twice on a pristine instance of SHA-512 (block size 128 bytes):

1. `add_data({...}, 1)`
2. `add_data({...}, 126)`

The first call fills `m_buffer` with one byte and sets `m_position = 1`. Both is happening at `buffer_insert` at the very end of the function (because `m_position` was initially 0).
In the second call we append an additional 126 bytes to `m_buffer` (in the first `buffer_insert()`, as `m_position` is initially 1). The block still isn't full, so no `compress_n()` is invoked.
**The issue:** We again set `m_position = 127` only at the very end after the second `buffer_insert` (that copies the 126 bytes once more into the same buffer location). The offset calculations check out, so there's no functional problem. But frankly, to me this doesn't look like intentional control flow.

Side note: one needs to know that `buffer_insert()` "fills `m_buffer` with as many bytes from `input` as possible, but at most `m_buffer.size() - m_position`".

```C++
void MDx_HashFunction::add_data(const uint8_t input[], size_t length) {
   const size_t block_len = static_cast<size_t>(1) << m_block_bits;

   m_count += length;

   if(m_position) {
      buffer_insert(m_buffer, m_position, input, length);

      if(m_position + length >= block_len) {
         compress_n(m_buffer.data(), 1);
         input += (block_len - m_position);
         length -= (block_len - m_position);
         m_position = 0;
      }
   }

   // Just in case the compiler can't figure out block_len is a power of 2
   const size_t full_blocks = length >> m_block_bits;
   const size_t remaining = length & (block_len - 1);

   if(full_blocks > 0) {
      compress_n(input, full_blocks);
   }

   buffer_insert(m_buffer, m_position, input + full_blocks * block_len, remaining);
   m_position += remaining;
}
```

### Description of Proposed Improvement

This is an attempt on encapsulating the intermittent buffering into an opinionated helper class with a (hopefully) clear and minimal interface. @randombit I'd be grateful for feedback early-on. I'll describe the details of the helper class' API in Doxygen comments.

With the helper, `MDx_HashFunction::add_data()` would roughly look like that and not pose the issue described above:

```C++
void MDx_HashFunction::add_data(std::span<const uint8_t> input) {
   BufferSlicer in(input);

   while(!in.empty()) {
      if(const auto one_block = m_buffer.handle_unaligned_data(in)) {
         compress_n(one_block->data(), 1);
      }

      if(m_buffer.in_alignment()) {
         const auto [aligned_data, full_blocks] = m_buffer.aligned_data_to_process(in);
         if(full_blocks > 0) {
            compress_n(aligned_data.data(), full_blocks);
         }
      }
   }

   m_count += input.size();
}
```

A quick trail benchmark run (average of 4 runs `./botan speed SHA-1`) :

| Branch | MiB/s |
| ------ | ----- |
| this1  | 1830  |
| master | 1829  |

### DIFF
--- a/src/lib/hash/blake2/blake2b.cpp
+++ b/src/lib/hash/blake2/blake2b.cpp
@@ -14,33 +14,26 @@
 #include <botan/internal/loadstor.h>
 #include <botan/internal/rotate.h>
 #include <botan/internal/stl_util.h>
+
 #include <algorithm>
+#include <array>
 
 namespace Botan {
 
 namespace {
 
-enum blake2b_constant { BLAKE2B_BLOCKBYTES = 128, BLAKE2B_IVU64COUNT = 8 };
-
-const uint64_t blake2b_IV[BLAKE2B_IVU64COUNT] = {0x6a09e667f3bcc908,
-                                                 0xbb67ae8584caa73b,
-                                                 0x3c6ef372fe94f82b,
-                                                 0xa54ff53a5f1d36f1,
-                                                 0x510e527fade682d1,
-                                                 0x9b05688c2b3e6c1f,
-                                                 0x1f83d9abfb41bd6b,
-                                                 0x5be0cd19137e2179};
+constexpr std::array<uint64_t, 8> blake2b_IV{0x6a09e667f3bcc908,
+                                             0xbb67ae8584caa73b,
+                                             0x3c6ef372fe94f82b,
+                                             0xa54ff53a5f1d36f1,
+                                             0x510e527fade682d1,
+                                             0x9b05688c2b3e6c1f,
+                                             0x1f83d9abfb41bd6b,
+                                             0x5be0cd19137e2179};
 
 }  // namespace
 
-BLAKE2b::BLAKE2b(size_t output_bits) :
-      m_output_bits(output_bits),
-      m_buffer(BLAKE2B_BLOCKBYTES),
-      m_bufpos(0),
-      m_H(BLAKE2B_IVU64COUNT),
-      m_T(),
-      m_F(),
-      m_key_size(0) {
+BLAKE2b::BLAKE2b(size_t output_bits) : m_output_bits(output_bits), m_H(blake2b_IV.size()), m_T(), m_F(), m_key_size(0) {
    if(output_bits == 0 || output_bits > 512 || output_bits % 8 != 0) {
       throw Invalid_Argument("Bad output bits size for BLAKE2b");
    }
@@ -49,17 +42,14 @@ BLAKE2b::BLAKE2b(size_t output_bits) :
 }
 
 void BLAKE2b::state_init() {
-   copy_mem(m_H.data(), blake2b_IV, BLAKE2B_IVU64COUNT);
+   copy_mem(m_H.data(), blake2b_IV.data(), blake2b_IV.size());
    m_H[0] ^= (0x01010000 | (static_cast<uint8_t>(m_key_size) << 8) | static_cast<uint8_t>(output_length()));
    m_T[0] = m_T[1] = 0;
    m_F = 0;
 
-   if(m_key_size == 0) {
-      m_bufpos = 0;
-   } else {
-      BOTAN_ASSERT_NOMSG(m_padded_key_buffer.size() == m_buffer.size());
-      copy_mem(m_buffer.data(), m_padded_key_buffer.data(), m_padded_key_buffer.size());
-      m_bufpos = m_padded_key_buffer.size();
+   m_buffer.clear();
+   if(m_key_size > 0) {
+      m_buffer.append(m_padded_key_buffer);
    }
 }
 
@@ -149,45 +139,28 @@ void BLAKE2b::compress(const uint8_t* input, size_t blocks, uint64_t increment) 
 }
 
 void BLAKE2b::add_data(std::span<const uint8_t> input) {
-   if(input.empty()) {
-      return;
-   }
-
    BufferSlicer in(input);
 
-   if(m_bufpos > 0) {
-      if(m_bufpos < BLAKE2B_BLOCKBYTES) {
-         const auto part = in.take(std::min(BLAKE2B_BLOCKBYTES - m_bufpos, in.remaining()));
-         copy_mem(&m_buffer[m_bufpos], part.data(), part.size());
-         m_bufpos += part.size();
+   while(!in.empty()) {
+      if(const auto one_block = m_buffer.handle_unaligned_data(in)) {
+         compress(one_block->data(), 1, BLAKE2B_BLOCKBYTES);
       }
 
-      if(m_bufpos == m_buffer.size() && !in.empty()) {
-         compress(m_buffer.data(), 1, BLAKE2B_BLOCKBYTES);
-         m_bufpos = 0;
+      if(m_buffer.in_alignment()) {
+         const auto [aligned_data, full_blocks] = m_buffer.aligned_data_to_process(in);
+         if(full_blocks > 0) {
+            compress(aligned_data.data(), full_blocks, BLAKE2B_BLOCKBYTES);
+         }
       }
    }
-
-   if(in.remaining() > BLAKE2B_BLOCKBYTES) {
-      const size_t full_blocks = ((in.remaining() - 1) / BLAKE2B_BLOCKBYTES);
-      compress(in.take(BLAKE2B_BLOCKBYTES * full_blocks).data(), full_blocks, BLAKE2B_BLOCKBYTES);
-   }
-
-   if(!in.empty()) {
-      const auto take = in.remaining();
-      copy_mem(&m_buffer[m_bufpos], in.take(take).data(), take);
-      m_bufpos += take;
-   }
-
-   BOTAN_ASSERT_NOMSG(in.empty());
 }
 
 void BLAKE2b::final_result(std::span<uint8_t> output) {
-   if(m_bufpos != BLAKE2B_BLOCKBYTES) {
-      clear_mem(&m_buffer[m_bufpos], BLAKE2B_BLOCKBYTES - m_bufpos);
-   }
+   const auto pos = m_buffer.elements_in_buffer();
+   m_buffer.fill_up_with_zeros();
+
    m_F = 0xFFFFFFFFFFFFFFFF;
-   compress(m_buffer.data(), 1, m_bufpos);
+   compress(m_buffer.consume().data(), 1, pos);
    copy_out_vec_le(output.data(), output_length(), m_H);
    state_init();
 }
@@ -229,9 +202,8 @@ void BLAKE2b::key_schedule(std::span<const uint8_t> key) {
 
 void BLAKE2b::clear() {
    zeroise(m_H);
-   zeroise(m_buffer);
+   m_buffer.clear();
    zeroise(m_padded_key_buffer);
-   m_bufpos = 0;
    m_key_size = 0;
    state_init();
 }
--- a/src/lib/hash/blake2/blake2b.h
+++ b/src/lib/hash/blake2/blake2b.h
@@ -10,12 +10,15 @@
 
 #include <botan/hash.h>
 #include <botan/sym_algo.h>
+#include <botan/internal/alignment_buffer.h>
 #include <memory>
 #include <string>
 
 namespace Botan {
 
 class BLAKE2bMAC;
+
+constexpr size_t BLAKE2B_BLOCKBYTES = 128;
 
 /**
 * BLAKE2B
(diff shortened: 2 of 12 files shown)


randombit:
Haven't reviewed this carefully yet but the broad idea looks ok to me

reneme:
> Haven't reviewed this carefully yet but the broad idea looks ok to me

That'll do for now, thanks. I'll push this forward tomorrow.

reneme:
This converged nicely for all hash implementations. The `HashFunction::add_data()` implementations are now basically all equivalent and unifying that into a single base implementation (using an abstract `compress()` method) should be easy to do.

I'll add more unit tests and documentation for `AlignmentBuffer<>` after lunch and will remove the "Draft" status afterwards.

### Future Work

Applying this to the hash implementations is probably enough complexity for this pull request. Other places (e.g. block cipher modes) might benefit from this refactoring as well, though. 

The **Keccak permutation** state may or may not benefit from `AlignmentBuffer<>` as well. Though, with the Keccak state not being a simple byte buffer, there's more work to be done.

@randombit Are there more places?

src/lib/hash/blake2/blake2b.cpp
```diff
@@ -149,45 +145,28 @@ void BLAKE2b::compress(const uint8_t* input, size_t blocks, uint64_t increment)
...
-
-   BOTAN_ASSERT_NOMSG(in.empty());
 }
 
 void BLAKE2b::final_result(std::span<uint8_t> output) {
-   if(m_bufpos != BLAKE2B_BLOCKBYTES) {
-      clear_mem(&m_buffer[m_bufpos], BLAKE2B_BLOCKBYTES - m_bufpos);
-   }
+   const auto pos = m_buffer.elements_in_buffer();
+   m_buffer.fill_up_with_zeros();
```
randombit, line 166: It's common (though not universal) that you call `fill_up_with_zeros` and then immediately `consume` the buffer. A helper that performed both operations in a single call might be useful.

src/lib/utils/alignment_buffer.h
```diff
@@ -0,0 +1,205 @@
...
+template <concepts::contiguous_container T>
+class AlignmentBuffer {
+   public:
+      using value_type = typename T::value_type;
+
+      /**
+       * @param buffer_size        typically the block size of the using algorithm
+       * @param defer_final_block  refrains from consuming the final block of an input
+       */
+      explicit AlignmentBuffer(size_t buffer_size, bool defer_final_block = false) :
```
randombit, line 45: Just for sanity check we should assert the buffer size isn't zero.

src/lib/utils/alignment_buffer.h
```diff
@@ -0,0 +1,205 @@
...
+      /**
+       * Once the buffer reached alignment, this can be used to consume as many
+       * input bytes from the given @p slider as possible. The output always
+       * contains data elements that are a multiple of the intrinsic block size.
+       *
+       * @returns a view onto the aligned data from @p slicer and the number of
+       *          full blocks that are represented by this view.
+       */
+      [[nodiscard]] std::tuple<std::span<const uint8_t>, size_t> aligned_data_to_process(BufferSlicer& slicer) const {
+         BOTAN_ASSERT_NOMSG(in_alignment());
```
randombit, line 102: This should probably should be `BOTAN_STATE_CHECK`

src/lib/utils/alignment_buffer.h
```diff
@@ -0,0 +1,205 @@
...
+ *
+ * This helper encapsulates such an alignment buffer. The API of this class
+ * is designed to minimize user errors in the algorithm implementations.
+ * Therefore, it is strongly opinionated on its use case. Don't try to use it
+ * for anything but the described circumstance.
+ *
+ * @tparam T  the container type to use for the alignment buffer
+ */
+template <concepts::contiguous_container T>
+class AlignmentBuffer {
```
randombit, line 37: WDYT about making `defer_final_block` a template parameter? Since I don't think we do (or would ever) have a use case where you would toggle between them at runtime.

randombit: Similarly, but perhaps more contentiously, if we made the buffer size a template parameter then

a) we might gain some minimal performance improvements due to the the block size being known to be fixed, eg when dividing by the buffer size.
b) (more importantly) we could skip using a container entirely and store the buffer in a simple array, avoiding the heap allocation.

This would be less flexible though, which *might* matter for some future case where we are doing runtime dispatch of the specific algorithm based on hardware support. Eg we want to align to 16 bytes if running regular C++ but 128 bytes if we're doing something with SIMD.

Also it would require exposing the internal block size in the headers of the various algorithms, but these headers are now internal only anyway so this doesn't seem so bad.

src/lib/utils/alignment_buffer.h
```diff
@@ -0,0 +1,205 @@
...
+
+      /**
+       * Once the buffer reached alignment, this can be used to consume full
+       * blocks from the input data represented by @p slicer.
+       *
+       * @returns a view onto the next full block from @p slicer or std::nullopt
+       *          if not enough data is available in @p slicer.
+       */
+      [[nodiscard]] std::optional<std::span<const uint8_t>> next_aligned_block_to_process(BufferSlicer& slicer) const {
+         BOTAN_ASSERT_NOMSG(in_alignment());
```
randombit, line 119: `BOTAN_STATE_CHECK`

randombit (commented):
This is nice, thanks. I left a few nit comments plus some ideas for possible improvements, ptal.

src/lib/utils/alignment_buffer.h
reneme: I'm certainly in favor of making things more "`constexpr`" here. Both the block size and the `defer_final_block` make sense as template parameters. Also I've been tinkering with a template-customizable `append()` function (to allow using this in Keccak where -- among other things -- we'd need to XOR instead of "overwrite"). The latter is certainly future work, though.

Regarding `std::array<>`: I had that in mind, but didn't go for it because we can't use the mlocked allocator for stack allocations. Is that something we can (and want to) live with?

reneme:
Thanks for the swift review! Re:`BOTAN_STATE_CHECK()`: [I changed the asserts (and also used `BOTAN_ARG_CHECK` where applicable)](https://github.com/randombit/botan/pull/3693/commits/5a3ff509f2619c6f7bf1f0f7def9cde0e9c1eff4). But, is that really correct? This class is internal and tightly coupled with the buffered computation use-case. If any of those ever throw, it is because we made a programming mistake in the consumer implementations. No particularly hard feelings about this, though, just double-checking.

src/lib/utils/alignment_buffer.h
randombit: I don't think the mlock aspect is that critical tbh. We spray intermediate values everywhere on the stack in the course of computing anything, and it's pretty unavoidable that someone with access to a core dump or swap file is going to be able to recover information.

If we go the array route should in any case scrub the memory on `clear` or destructor; that already puts us ahead of the game compared to most implementations I've seen.

randombit:
Hmm you have a point there. In a nominal sense `Invalid_State` is the correct throw ("A request was made on an object while the object was in a state where the operation cannot be performed.") but you're right that since this is an internal-only API it throwing implies a bug elsewhere within the library, rather than some usage error by the application.

reneme:
> since this is an internal-only API it throwing implies a bug elsewhere within the library, rather than some usage error by the application.

I read that as "let's revert 5a3ff50", right?

This commit contains a few additional tests for the state checks. I'm strongly assuming that it is not sensible to try and check asserts in the unittests, though. Particularly after introducing `./configure.py --terminate-on-assert`. 😅

src/lib/utils/alignment_buffer.h
reneme: Fair enough. I'll look into that tomorrow.

randombit:
Yeah let's revert that, sorry for the confusion.

src/lib/utils/alignment_buffer.h
randombit: :+1:

src/lib/hash/blake2/blake2b.cpp
randombit: No strong opinion on this, so if you're not enthused about the idea, fine to leave it as is.

src/lib/utils/alignment_buffer.h
reneme: I applied both suggestions, making `defer_final_block` and the `block_size` into template parameters. The latter proved problematic for `MDx_HashFunction` as this info is currently a runtime information (behind a virtual method). Once #3550 is finished, that shouldn't be a problem anymore, though.

src/lib/utils/alignment_buffer.h
```diff
@@ -0,0 +1,219 @@
...
+
+      ~AlignmentBuffer() { clear(); }
+
+      AlignmentBuffer(const AlignmentBuffer& other) = default;
+      AlignmentBuffer(AlignmentBuffer&& other) noexcept = default;
+      AlignmentBuffer& operator=(const AlignmentBuffer& other) = default;
+      AlignmentBuffer& operator=(AlignmentBuffer&& other) noexcept = default;
+
+      void clear() {
+         clear_mem(m_buffer.data(), m_buffer.size());
```
randombit, line 60: Should be `secure_scrub_memory` to prevent being elided during the destructor.

randombit: (Or alternately continue to use `clear_mem` here and just invoke `secure_scrub_memory` once, in the destructor)

src/lib/hash/blake2/blake2b.h
```diff
@@ -10,13 +10,16 @@
...
 #include <botan/sym_algo.h>
+#include <botan/internal/alignment_buffer.h>
 #include <memory>
 #include <string>
 
 namespace Botan {
 
 class BLAKE2bMAC;
 
+enum blake2b_constant { BLAKE2B_BLOCKBYTES = 128, BLAKE2B_IVU64COUNT = 8 };
```
randombit, line 21: There doesn't seem to be a reason to expose `BLAKE2B_IVU64COUNT` here

reneme: True. I just used `std::array<>` in the *.cpp file and now use `blakeIV.size()` where needed.

src/lib/hash/skein/skein_512.h
```diff
@@ -27,6 +29,8 @@ class Skein_512 final : public HashFunction {
       */
       Skein_512(size_t output_bits = 512, std::string_view personalization = "");
 
+      Skein_512(const Skein_512& other);
```
randombit, line 32: This is declared but does not seem to be defined anywhere.

src/lib/utils/alignment_buffer.h
```diff
@@ -0,0 +1,219 @@
...
+
+#include <array>
+#include <optional>
+#include <span>
+
+namespace Botan {
+
+enum class AlignmentBufferFinalBlock : size_t {
+   IS_NOT_SPECIAL = 0,
+   MUST_BE_DEFERRED = 1,
```
randombit, line 23: IMO `enum class` members should not be `SHOUTING_CASE`

randombit: Also for future users, briefly document the usage of this enum in a comment.

randombit (commented):
Looks really good - left a couple of nitpicky comments

src/lib/hash/skein/skein_512.h
reneme: Good catch! I was running into issues when state-cloning this hash along the way. This is a remnant of a debugging attempt.

randombit (approved):

src/lib/utils/alignment_buffer.h
reneme: Thanks for the approval. I just pushed another change (had forgotten to add documentation of the enum class).

reneme:
For the record: I pushed another change after your approval (documenting the enum class `AlignmentBufferFinalBlock`). Please check and merge at your convenience.

randombit (approved):
Thanks

PR #3695, opened 2023-09-13
Prepare for Kyber standardization updates

randombit:
Use _Round3 suffix on the KyberMode to allow us to easily support the final standardized version. This also makes it clear to users that they are using the non-final Kyber.

This deprecates Kyber 90s mode since it is not included in the draft FIPS, seems to be not nearly as widely implemented as the SHA-3 based version, and supporting it complicates the implementation.

### DIFF
--- a/doc/api_ref/pubkey.rst
+++ b/doc/api_ref/pubkey.rst
@@ -12,7 +12,7 @@ Public and private keys are represented by classes ``Public_Key`` and
 ``Private_Key``. Both derive from ``Asymmetric_Key``.
 
 Currently there is an inheritance relationship between ``Private_Key`` and
-``Public_Key``, so that a private key can also be used as the cooresponding
+``Public_Key``, so that a private key can also be used as the corresponding
 public key. It is best to avoid relying on this, as this inheritance will be
 removed in a future major release.
 
@@ -102,7 +102,7 @@ Post-quantum key encapsulation scheme based on lattices.
 .. note::
 
    Currently two modes for Kyber are defined: the round3 specification
-   from the NIST PQC competetition, and the "90s mode" (which uses
+   from the NIST PQC competition, and the "90s mode" (which uses
    AES/SHA-2 instead of SHA-3 based primitives). The 90s mode Kyber is
    deprecated and will be removed in a future release.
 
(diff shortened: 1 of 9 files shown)


doc/api_ref/pubkey.rst
```diff
@@ -99,6 +99,13 @@ Kyber
 
 Post-quantum key encapsulation scheme based on lattices.
 
+.. note::
+
+   Currently two modes for Kyber are defined: the round3 specification
+   from the NIST PQC competetition, and the "90s mode" (which uses
```
reneme, line 105:
```suggestion
   from the NIST PQC competition, and the "90s mode" (which uses
```

src/lib/pubkey/kyber/kyber_common/kyber.h
```diff
@@ -29,7 +29,19 @@ namespace Botan {
 
 class BOTAN_PUBLIC_API(3, 0) KyberMode {
    public:
-      enum Mode { Kyber512, Kyber512_90s, Kyber768, Kyber768_90s, Kyber1024, Kyber1024_90s };
+      enum Mode {
+         Kyber512_Round3,
+         Kyber768_Round3,
+         Kyber1024_Round3,
```
reneme, line 35: The respective `to_string()` method and the OID-map use the short-hand "r3" already. I think, "Round3" makes it more obvious to the average user that this refers to a pre-final Kyber spec and not just some obscure technicality. But, I think, we should be consistent either way.

randombit: Yeah good point, especially if the round3 variant ends up (as I expect) having a long life here, vs 90s mode which is probably going to be removed soonish.

randombit: Oof - I was looking at Dilithium and there we use r3 suffix as well - and then there is a note in `dilithium.h` that we are not compatible with the round 3 submission :see_no_evil:

randombit:
I went with using `_R3` suffix on the enums. I think that plus sufficient documentation is enough to point people in the right direction - especially once plain `Kyber512` etc points to the final NIST version.

doc/api_ref/pubkey.rst
```diff
@@ -99,6 +99,15 @@ Kyber
...
```
reneme, line 105: Lost in rebase or force push?

```suggestion
   from the NIST PQC competition, and the "90s mode" (which uses
```

reneme (approved):
LGTM, apart from the re-introduced typo in the documentation.

doc/api_ref/pubkey.rst
randombit: There were several spelling errors in this file, fixed in baea85c

PR #3696, opened 2023-09-13
rsa, raw_op using std::span view for raw_kem_decrypt to avoid

devnexen:
copying again the output.

### DIFF
--- a/src/lib/pubkey/rsa/rsa.cpp
+++ b/src/lib/pubkey/rsa/rsa.cpp
@@ -445,21 +445,27 @@ class RSA_Private_Operation {
             m_max_d1_bits(m_private->p_bits() + m_blinding_bits),
             m_max_d2_bits(m_private->q_bits() + m_blinding_bits) {}
 
-      secure_vector<uint8_t> raw_op(const uint8_t input[], size_t input_len) {
-         if(input_len > public_modulus_bytes()) {
+      void raw_op(std::span<uint8_t> out, std::span<const uint8_t> input) {
+         if(input.size() > public_modulus_bytes()) {
             throw Decoding_Error("RSA input is too long for this key");
          }
-         const BigInt input_bn(input, input_len);
+         const BigInt input_bn(input.data(), input.size());
          if(input_bn >= m_public->get_n()) {
             throw Decoding_Error("RSA input is too large for this key");
          }
-
          // TODO: This should be a function on blinder
          // BigInt Blinder::run_blinded_function(std::function<BigInt, BigInt> fn, const BigInt& input);
 
          const BigInt recovered = m_blinder.unblind(rsa_private_op(m_blinder.blind(input_bn)));
          BOTAN_ASSERT(input_bn == m_public->public_op(recovered), "RSA consistency check");
-         return BigInt::encode_1363(recovered, m_public->public_modulus_bytes());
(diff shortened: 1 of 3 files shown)


src/lib/pubkey/rsa/rsa.cpp
```diff
@@ -445,6 +445,11 @@ class RSA_Private_Operation {
             m_max_d1_bits(m_private->p_bits() + m_blinding_bits),
             m_max_d2_bits(m_private->q_bits() + m_blinding_bits) {}
 
+      std::span<uint8_t> raw_op(std::span<const uint8_t> input) {
+         auto sv = raw_op(input.data(), input.size());
+         return std::span<uint8_t>(sv.begin(), sv.end());
```
randombit, line 450: This doesn't seem right - `sv` is a temporary within this function so on return the `span` will point to deallocated memory.

FWIW I am not sure exactly what the comment is suggesting to do - @reneme do you remember?

src/lib/pubkey/rsa/rsa.cpp
```diff
@@ -445,6 +445,20 @@ class RSA_Private_Operation {
             m_max_d1_bits(m_private->p_bits() + m_blinding_bits),
             m_max_d2_bits(m_private->q_bits() + m_blinding_bits) {}
 
+      void raw_op(std::span<const uint8_t> input, std::span<uint8_t> out) {
```
reneme, line 448: Not sure this is a consistent convention, but typically, we put out-params first in the list.

src/lib/pubkey/rsa/rsa.cpp
reneme: For the record: I agree that this isn't correct (it would be good if tooling would warn about it).

> I am not sure exactly what the comment is suggesting to do

I'm assuming that you mean

```C++
// TODO: avoid copy by letting raw_op() work on std::span<>
```

which hints at making `raw_op()` work on a `std::span<uint8_t>` out-param just like `raw_kem_decrypt()` does. This would save a copy.

Meanwhile @devnexen adapted the patch accordingly.

src/lib/pubkey/rsa/rsa.cpp
```diff
@@ -445,6 +445,20 @@ class RSA_Private_Operation {
...
+         if(input_bn >= m_public->get_n()) {
+            throw Decoding_Error("RSA input is too large for this key");
+         }
+         const BigInt recovered = m_blinder.unblind(rsa_private_op(m_blinder.blind(input_bn)));
+         BOTAN_ASSERT(input_bn == m_public->public_op(recovered), "RSA consistency check");
+         BOTAN_ASSERT(m_public->public_modulus_bytes() == out.size(), "output size check");
+         BigInt::encode_1363(out, recovered);
+      }
+
       secure_vector<uint8_t> raw_op(const uint8_t input[], size_t input_len) {
```
reneme, line 462: Please avoid the duplicated implementation by using the out-param based method. Along those lines:

```C++
secure_vector<uint8_t> raw_op(const uint8_t input[], size_t input_len) {
   secure_vector<uint8_t> out(m_public->public_modulus_bytes);
   raw_op(out, {input, input_len});
   return out;
}
```

... as a follow-up we should probably see if we can get rid of this overload entirely.

reneme (commented):
Thanks!

reneme (approved):
LGTM now.

PR #3700, opened 2023-09-14
tls_session unfold the input as std::span::subspan to remove raw poin…

devnexen:
…ters usage.


src/lib/tls/tls_session.cpp
```diff
@@ -428,15 +428,15 @@ Session Session::decrypt(std::span<const uint8_t> in, const SymmetricKey& key) {
...
-      const uint8_t* key_name = magic + TLS_SESSION_CRYPT_MAGIC_LEN;
-      const uint8_t* key_seed = key_name + TLS_SESSION_CRYPT_KEY_NAME_LEN;
-      const uint8_t* aead_nonce = key_seed + TLS_SESSION_CRYPT_AEAD_KEY_SEED_LEN;
-      const uint8_t* ctext = aead_nonce + TLS_SESSION_CRYPT_AEAD_NONCE_LEN;
+      size_t inoff = 0;
+      const auto key_name = in.subspan(inoff += TLS_SESSION_CRYPT_MAGIC_LEN).data();
+      const auto key_seed = in.subspan(inoff += TLS_SESSION_CRYPT_KEY_NAME_LEN).data();
+      const auto aead_nonce = in.subspan(inoff += TLS_SESSION_CRYPT_AEAD_KEY_SEED_LEN).data();
+      const auto ctext = in.subspan(inoff += TLS_SESSION_CRYPT_AEAD_NONCE_LEN).data();
+      BOTAN_ASSERT(inoff < in.size(), "input consistency check");
```
reneme, line 436: Actually, I think we want to use the `BufferSlicer` helper here as it keeps track of what is here called `inoff` internally.

Like so:

```C++
BufferSlicer s(in);
const auto magic =  s.take(TLS_SESSION_CRYPT_MAGIC_LEN);
const auto key_name = s.take(TLS_SESSION_CRYPT_KEY_NAME_LEN);
...
```

reneme (changes requested):
I think the `BufferSlicer` in `stl_util.h` is perfect for this task.

src/lib/tls/tls_session.cpp
```diff
@@ -428,12 +428,12 @@ Session Session::decrypt(std::span<const uint8_t> in, const SymmetricKey& key) {
...
-      const uint8_t* key_seed = key_name + TLS_SESSION_CRYPT_KEY_NAME_LEN;
-      const uint8_t* aead_nonce = key_seed + TLS_SESSION_CRYPT_AEAD_KEY_SEED_LEN;
-      const uint8_t* ctext = aead_nonce + TLS_SESSION_CRYPT_AEAD_NONCE_LEN;
+      BufferSlicer sub(in);
+      const auto magic = sub.take(TLS_SESSION_CRYPT_MAGIC_LEN).data();
+      const auto key_name = sub.take(TLS_SESSION_CRYPT_KEY_NAME_LEN).data();
+      const auto key_seed = sub.take(TLS_SESSION_CRYPT_AEAD_KEY_SEED_LEN).data();
+      const auto aead_nonce = sub.take(TLS_SESSION_CRYPT_AEAD_NONCE_LEN).data();
+      const auto ctext = sub.take(TLS_SESSION_CRYPT_HDR_LEN).data();
       const size_t ctext_len = in.size() - TLS_SESSION_CRYPT_HDR_LEN;  // includes the tag
```
reneme, line 437:
```suggestion
      auto ctext = sub.copy_as_secure_vector(sub.remaining());
```

... and then just use the `ctext` variable in `aead->finish()` down below straight away without copying into `buf` before.

reneme (approved):
Nice! Thank you.

PR #3701, opened 2023-09-15
FIX: cSHAKE should not depend on the user calling start()

reneme:
[Similar to the `MessageAuthenticationCode`](https://github.com/randombit/botan/pull/3689#discussion_r1326853886), XOFs provide a `start()` method that is optional when a user doesn't want to use a salt (or the XOF itself doesn't support one). Nevertheless, we need to make sure that preconditioning of the XOF's state is not dependent on the user calling `start()`.

The `XOF` base class is taking care of this, by keeping track whether `start()` was called before the first `update()` invocation. If not, it calls it with default values (i.e. "empty"). Also, this adds a test to make sure that `start()` is optional (similar to #3699).

Also, this adds a test that cSHAKE fails when neither `N` nor `S` is set. This introduces many `try {} catch {}` blocks in test_xof.cpp. It's beneficial to review the two commits separately.


src/lib/xof/xof.h
```diff
@@ -203,6 +215,14 @@ class BOTAN_PUBLIC_API(3, 2) XOF {
        * @param output  the span to be filled entirely with output bytes
        */
       virtual void generate_bytes(std::span<uint8_t> output) = 0;
+
+      /**
+       * Clear the XOF's internal state and allow for new input.
+       */
+      virtual void reset() = 0;
```
randombit, line 222: TBH I do not understand the semantic difference between `reset` and `clear` here; when/why would a user choose one vs the other? I have a feeling `reset` should be protected and only used internally

src/tests/data/xof/cshake.vec
```diff
@@ -20,6 +20,17 @@ Salt = 456d61696c205369676e6174757265
...
 Out = C5221D50E4F822D96A2E8881A961420F294B7B24FE3D2094BAED2C6524CC166B
 
+# cSHAKE instantiation with a Name but no Salt should work
+Name = 426f74616e2d54657374
+In = 000102030405060708090A0B0C0D0E0F101112131415161718191A1B1C1D1E1F202122232425262728292A2B2C2D2E2F303132333435363738393A3B3C3D3E3F404142434445464748494A4B4C4D4E4F505152535455565758595A5B5C5D5E5F606162636465666768696A6B6C6D6E6F707172737475767778797A7B7C7D7E7F808182838485868788898A8B8C8D8E8F909192939495969798999A9B9C9D9E9FA0A1A2A3A4A5A6A7A8A9AAABACADAEAFB0B1B2B3B4B5B6B7B8B9BABBBCBDBEBFC0C1C2C3C4C5C6C7
+Out = 429F4B1FFD82BAAD7EB191BCFD1C9A89486C2C8D
+
+# cSHAKE instantiations without neither a Name nor a Salt should fail
+# (we don't implement the NIST-specified fallback to vanilla SHAKE in that case)
+Fail = cSHAKE-128 cannot accept a salt length of 0
```
randombit, line 30: Conventionally these KAT tests are only used for checking a map from Inputs to Output and while yeah nominally an exception is the "output" here, I think it would be clearer to skip this part of the change and then just have a one-off test that checks cSHAKE rejects in this case.

src/lib/xof/xof.h
reneme: > I have a feeling reset should be protected and only used internally

That was exactly my intention. I wanted to separate the public interface (`clear()`, not virtual) from the internal one (`reset()` virtual). So that the base class can safely handle the `m_xof_started` state.

In the sub-classes I forgot to move the `reset()` overrides into the `private:` section, though. Maybe that caused your confusion?

src/tests/data/xof/cshake.vec
reneme: Moved it into a simpler test inside `run_final_tests()`.

src/lib/xof/xof.h
randombit: Sorry, I missed in the diff that `XOF::reset` was declared in the private block!

src/tests/data/xof/cshake.vec
randombit: Thanks

randombit (approved):

PR #3702, opened 2023-09-15
Add some minimal API sanity checks for PK_Signer/PK_Verifier

lieser:
Add some minimal API sanity checks for the `PK_Signer`/`PK_Verifier` that are not covered by other general public key tests.

The main motivation behind this is to detect use after free problems in case the `PK_Signer`/`PK_Verifier` wrongly only hold a reference to the keys.
Bellow is a table of affected classes. Together with a status if it was detected in the CI by the sanitizers/valgrind, and if it is fixed in this PR.

Possible use-after-free:

|                 class                 | sanitizers |    valgrind    | fixed |
| ------------------------------------- | ---------- | -------------- | ----- |
| `Dilithium_Signature_Operation`       | ✔          | ✔              | ✔     |
| `ECDSA_Signature_Operation`           | ✔          | ✔              | ✔     |
| `ECGDSA_Signature_Operation`          | ✔          | ✔              | ✔     |
| `ECKCDSA_Signature_Operation`         | ✔          | ✔              | ✔     |
| `GOST_3410_Signature_Operation`       | ✔          | ✔              | ✔     |
| `PKCS11_ECDSA_Signature_Operation`    | ❌          | ❌ (no PKCS#11) | ✔     |
| `PKCS11_RSA_Signature_Operation`      | ❌          | ❌ (no PKCS#11) | ✔     |
| `SM2_Signature_Operation`             | ✔          | ✔              | ✔     |
| `TPM_Signing_Operation`               | ❌          | ❌ (no TPM)     | ✔     |
| `XMSS_PrivateKey_Internal`            | ✔          | ❌ (skipped)    | ✔     |
| `Ed25519_Pure_Verify_Operation`       | ✔          | ❌ (skipped)    | ✔     |
| `Ed25519_Hashed_Verify_Operation`     | ✔          | ❌ (skipped)    | ✔     |
| `PKCS11_ECDSA_Verification_Operation` | ❌          | ❌ (no PKCS#11) | ✔     |
| `PKCS11_RSA_Verification_Operation`   | ❌          | ❌ (no PKCS#11) | ✔     |
| `XMSS_Verification_Operation`         | ✔          | ❌ (skipped)    | ✔     |

Possible null pointer dereferencing:

|            class             | sanitizers | valgrind | fixed |
| ---------------------------- | ---------- | -------- | ----- |
| `SM2_Signature_Operation`    | ✔          | ✔        | ✔     |
| `SM2_Verification_Operation` | ✔          | ✔        | ✔     |

Also added a note to the documentation that for PKCS#11 the session needs to remain alive as long as any object using it is still alive.

### DIFF
--- a/src/tests/data/pubkey/api_sign.vec
+++ b/src/tests/data/pubkey/api_sign.vec
@@ -1,0 +1,73 @@
+# [Dilithium]
+
+Algorithm = Dilithium
+AlgoParams = Dilithium-4x4-r3
+SigParams =
+
+# [DSA]
+
+Algorithm = DSA
+AlgoParams = dsa/jce/1024
+SigParams = SHA-256
+
+# [ECDSA]
+
+Algorithm = ECDSA
+AlgoParams = secp256k1
+SigParams = SHA-256
+
+# [ECGDSA]
+
+Algorithm = ECGDSA
+AlgoParams = secp256r1
+SigParams = SHA-256
+
+# [ECKCDSA]
+
+Algorithm = ECKCDSA
+AlgoParams = secp256r1
+SigParams = SHA-256
+
+# [Ed25519]
+
+Algorithm = Ed25519
+AlgoParams =
+SigParams = Pure
+
+Algorithm = Ed25519
+AlgoParams =
+SigParams = Ed25519ph
+
+# [GOST-34.10]
+
+Algorithm = GOST-34.10
+AlgoParams = gost_256A
+SigParams = Raw
+
+# [RSA]
+
+Algorithm = RSA
+AlgoParams = 2048
+SigParams = PSSR(SHA-256)
+
+# [SM2]
+
+Algorithm = SM2
+AlgoParams = secp256r1
+SigParams = ALICE123@YAHOO.COM,SM3
+
+Algorithm = SM2
+AlgoParams = secp256r1
+SigParams = ALICE123@YAHOO.COM,Raw
+
+# [SPHINCS+]
+
+Algorithm = SPHINCS+
+AlgoParams = SphincsPlus-shake-128s-r3.1
+SigParams =
+
+# [XMSS]
+
+Algorithm = XMSS
+AlgoParams = XMSS-SHA2_10_256
+SigParams = SHA2_10_256
--- a/src/tests/test_pubkey.cpp
+++ b/src/tests/test_pubkey.cpp
@@ -17,10 +17,13 @@
    #include <botan/pkcs8.h>
    #include <botan/pubkey.h>
    #include <botan/x509_key.h>
+   #include <botan/internal/fmt.h>
 
    #if defined(BOTAN_HAS_HMAC_DRBG)
       #include <botan/hmac_drbg.h>
    #endif
+
+   #include <array>
 
 namespace Botan_Tests {
 
@@ -113,7 +116,7 @@ Test::Result PK_Signature_Generation_Test::run_one_test(const std::string& pad_h
    auto pubkey = Botan::X509::load_key(Botan::X509::BER_encode(*privkey));
 
    result.confirm("public key claims to support signatures",
-                  privkey->supports_operation(Botan::PublicKeyOperation::Signature));
+                  pubkey->supports_operation(Botan::PublicKeyOperation::Signature));
 
    std::vector<std::unique_ptr<Botan::PK_Verifier>> verifiers;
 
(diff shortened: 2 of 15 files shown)


lieser:
Open point for discussion: I'm unsure if it is a good idea to add the new tests to the `PK_Signature_Generation_Test` test, or if it would not be better to add a new general test for it.

Advantage for `PK_Signature_Generation_Test`:

- The test are automatically added form most public key algorithms.

Advantage for a separate test:

- Can be reused for algorithms that do not have a `PK_Signature_Generation_Test` test.
- Would probably be enough to run the new test once per algorithm instead of for each KAT test.

randombit:
I would suggest a new test for this.

My thinking right now would be to use `Text_Based_Test` that has a format along the lines of

```
Algorithm = RSA
AlgoParams = 2048
SigParams = PSSR(SHA-256)

Algorithm = ECDSA
AlgoParams = secp256k1
SigParams = SHA-256
...
```

which then does the basic test that you have written here but using `create_private_key(Algorithm, AlgoParams)` then creates a `PK_Signer`/`PK_Verifier` using `SigParams`.

We have to remember to update the test each time a new algorithm is added but that isn't that commonly done, and it avoids running the same test with each KAT which seems somewhat excessive.

A similar test could/should be written for the encryptor and key agreement `pubkey.h` types

randombit:
The ideal fix here would be, as with RSA and DH already, for the keys to be stored internally as a `shared_ptr` and then the operations objects can just capture the key information that way. I started on changing this for the EC objects prior to 3.0 and never got it over the line; maybe I'll take this up again next week.

lieser:
> My thinking right now would be to use Text_Based_Test that has a format along the lines of [...]

With a separate test I was thinking about creating a base class that all algorithms then have to derive from, similar too `PK_Signature_Generation_Test`.
But I like your suggestion, and will try to find time this week to rewrite it as a single `Text_Based_Test`.

Writing something similar for the encryptor and key agreement is probably a good idea. At least `PKCS11_RSA_Encryption_Operation` seems to have the same problem. But I will probably leave it as a follow up.

> The ideal fix here would be, as with RSA and DH already, for the keys to be stored internally as a shared_ptr and then the operations objects can just capture the key information that way.

Yeah I saw that some algorithms go this way. And if the key data is big enough that we don't want to copy it around it is probably the way to go long term.
For this PR I will however keep the fixes as simple as possible if that is OK.

randombit:
> Yeah I saw that some algorithms go this way. And if the key data is big enough that we don't want to copy it around it is probably the way to go long term.
For this PR I will however keep the fixes as simple as possible if that is OK.

For now just capturing by value as you're doing is absolutely fine, thanks for looking into this.

randombit:
Sorry stray mouse click there

lieser:
Reworked the test to be a single `Text_Based_Test` test. Most classes are now fixed, only the PKCS11 and TPM related ones are still missing. Have to see if/how we want to test that with sanitizers in the CI too.

A note about PKCS11:
There all objects hold a reference to a `Session`. This is also a potential for memory problems, but probably not something we want to change for now. That seems to be to integrated into the PKCS11 code.
But at least at a glance I also did not see a big warning in the manual that the user has to ensure that a `Session` exists as long a any of the associated object still exists.

randombit:
@lieser Is this ready for review? Even if it does not completely address all the issues, what you have already seems an improvement so it would be good to merge this for 3.2 if we can.

randombit:
Here I think tiny commits is fine since it makes it easier to selectively backport fixes, if someone needs to do that at some point.

src/tests/data/pubkey/api_sign.vec
```diff
@@ -0,0 +1,73 @@
+# [Dilithium]
```
randombit, line 1: I know I suggested the current layout, but actually the headers would make it a bit easier to read. Maybe let's drop `Algorithm` as a key and instead use the header block to specify which algo we're testing?

src/tests/test_pubkey.cpp
```diff
@@ -742,6 +745,72 @@ Test::Result PK_Key_Generation_Stability_Test::run_one_test(const std::string&,
...
+
+         std::unique_ptr<Botan::PK_Signer> signer = std::make_unique<Botan::PK_Signer>(
+            *privkey, Test::rng(), sig_params, Botan::Signature_Format::Standard, provider);
+         std::unique_ptr<Botan::PK_Verifier> verifier =
+            std::make_unique<Botan::PK_Verifier>(*pubkey, verify_params, Botan::Signature_Format::Standard, provider);
+         result.confirm("Creating PK_Signer works", signer != nullptr);
+         result.confirm("Creating PK_Signer works", verifier != nullptr);
+
+         result.test_is_nonempty("PK_Signer should report some hash", signer->hash_function());
+         result.test_is_nonempty("PK_Verifier should report some hash", verifier->hash_function());
```
randombit, line 797: Should we check that the two objects agree on the hash function?

(Maybe there is some corner case where that isn't the case?)

src/tests/test_pubkey.cpp
```diff
@@ -742,6 +745,72 @@ Test::Result PK_Key_Generation_Stability_Test::run_one_test(const std::string&,
...
+         std::unique_ptr<Botan::Public_Key> pubkey = Botan::X509::load_key(Botan::X509::BER_encode(*privkey));
+         result.confirm("Storing and loading public key works", pubkey != nullptr);
+
+         result.confirm("private key claims to support signatures",
+                        privkey->supports_operation(Botan::PublicKeyOperation::Signature));
+         result.confirm("public key claims to support signatures",
+                        pubkey->supports_operation(Botan::PublicKeyOperation::Signature));
+         result.test_gt("Public key length must be greater than 0", privkey->key_length(), 0);
+
+         std::unique_ptr<Botan::PK_Signer> signer = std::make_unique<Botan::PK_Signer>(
```
randombit, line 789: For cases like this where the type is being named already in the expression that initializes the type, I think `auto` is a little easier.

randombit (approved):
Looks good, thanks! I left a few minor comments/suggestions for your consideration, all optional - fine to merge this as is if you prefer.

lieser:
Your suggestions all look reasonable.
Unfortunately I will not be able to do the changes before the 9.10.

If that is too late for the release I could do changes in a follow up PR.
Or maybe you or @reneme could just apply the suggestions to my PR.

PR #3705, opened 2023-09-18
Refactor: Remove MDx_HashFunction

reneme:
### Pull Request Dependencies

* ~#3693~

### Description

This is a re-iteration of #3550 which became stale mostly due to `clang-format`. Note that I took the liberty to implement the alternative suggestion on configuring the instances' static configurations. I posted this on the original pull request: https://github.com/randombit/botan/pull/3550#discussion_r1198937199.

In my opinion, this has two important advantages:

1. **readability:** All "magic numbers" in the MDx instances have clear names, instead of just being an unordered template parameter list. Also, those names are available throughout the hash's and the helper's implementation.
2. **"global" static availability:** The "magic numbers" are available from anywhere within the library. For instance: If some internal downstream code needs the output length of SHA-256 as a static information, it can just include `sha2_32.h` and go for `SHA_256::output_bytes`.

@randombit I'd be interested in your opinion on the usefulness of exposing some algorithm specifics statically. FWIW: Here's another example where this would be helpful: https://github.com/randombit/botan/pull/3609#discussion_r1328867982. To be clear: I think, this should be an internal feature and not exposed via public API.


reneme:
Rebased to master after #3693 got merged.

randombit (commented):
I haven't done a complete review yet, but yeah looks pretty good.

Please do check that this doesn't introduce performance regressions.

randombit:
On my machine anyway this seems very slightly slower, but it's under .1 cycles per byte and mostly notable for MD4/MD5 where the compression function is very cheap - seems fine.

src/lib/hash/sha2_64/sha2_64.cpp
```diff
@@ -29,23 +30,11 @@ std::string sha512_provider() {
...
-   return std::make_unique<SHA_384>(*this);
-}
-
-std::unique_ptr<HashFunction> SHA_512::copy_state() const {
-   return std::make_unique<SHA_512>(*this);
-}
-
-std::unique_ptr<HashFunction> SHA_512_256::copy_state() const {
-   return std::make_unique<SHA_512_256>(*this);
-}
```
randombit, line 42: I think the various `copy_state` and such should *not* be inlined; it's already an expensive operation due to requiring a heap allocation.

In the predecessor PR I actually went in the other direction and moved all of the `new_object` to be not inline.

src/lib/hash/sha2_32/sha2_32.h
```diff
@@ -16,70 +16,96 @@ namespace Botan {
...
       std::string provider() const override;
 
-      SHA_224() : MDx_HashFunction(64, true, true), m_digest(8) { clear(); }
-
    private:
-      void compress_n(const uint8_t[], size_t blocks) override;
-      void copy_out(uint8_t[]) override;
+      void add_data(std::span<const uint8_t> input) override { m_md.update(input); }
+
+      void final_result(std::span<uint8_t> output) override { m_md.final(output); }
```
randombit, line 50: I don't think inlining these makes sense. For application users, they can't take advantage of the inlined nature. And it blows up code size for anything within the library that's operating on the concrete types.

randombit (commented):
Did a full review, this is basically fine outside of some quibbling about what should or should not be inline.

reneme:
Thanks for the review! I'm glad you like this design.

I'm done with addressing your comments. I.e. renaming the MD hash template variable from `T` to `MD` and moving the implementations of `copy_state`, `new_object`, `add_data` and `final_result` into the respective *.cpp files of the concrete hash implementations.

randombit (approved):
Thanks @reneme looks great

PR #3707, opened 2023-09-19
[std::span] {load,store}_{le,be}

reneme:
### Pull Request Dependencies

* ~~#3715~~

### Description

This has two objectives:

1. **range-based overloads for big/little endian helpers**
    The lack of this has been annoying me for some time now. Using ranges facilitates maximum usage convenience while enabling opportunistic compile-time buffer size checks when the passed range has static length information (read: `std::array` or `std::span<T, 42>`). Otherwise runtime checks will be performed (`BOTAN_ARG_CHECK`).
2. **Reduce the overload complexity**
    ... by letting the compiler generate most of the generic code as `constexpr`. I didn't see any slow-downs (in the runtime of the test suite) after applying this.

Here's a godbolt (of a somewhat stripped-down version) to play around with that, if needed: https://godbolt.org/z/sxjhnbePo

Currently, this includes legacy (pointer-based) overloads of the existing helpers. Those just assert that the passed in pointers are big enough and wrap them into properly sized `std::span`s to forward the calls. In an ideal world we'd need to go through the code base and replace all those calls with statically checkable invocations.

### DIFF
--- a/src/lib/utils/concepts.h
+++ b/src/lib/utils/concepts.h
@@ -33,6 +33,14 @@ struct is_strong_type<Strong<Ts...>> : std::true_type {};
 template <typename... Ts>
 constexpr bool is_strong_type_v = is_strong_type<std::remove_const_t<Ts>...>::value;
 
+template <typename T0, typename... Ts>
+struct all_same {
+      static constexpr bool value = (std::is_same_v<T0, Ts> && ...);
+};
+
+template <typename... Ts>
+static constexpr bool all_same_v = all_same<Ts...>::value;
+
 namespace ranges {
 
 /**
@@ -204,6 +212,10 @@ concept contiguous_strong_type = strong_type<T> && contiguous_container<T>;
 // TODO: C++20 - replace with std::integral
 template <typename T>
 concept integral = std::is_integral_v<T>;
+
+// TODO: C++20 - replace with std::unsigned_integral
+template <typename T>
+concept unsigned_integral = std::is_integral_v<T> && std::is_unsigned_v<T>;
 
 }  // namespace concepts
 
--- a/src/lib/utils/ghash/ghash.cpp
+++ b/src/lib/utils/ghash/ghash.cpp
@@ -13,6 +13,8 @@
 #include <botan/internal/ct_utils.h>
 #include <botan/internal/loadstor.h>
 
+#include <array>
+
 namespace Botan {
 
 std::string GHASH::provider() const {
@@ -177,9 +179,12 @@ void GHASH::add_final_block(secure_vector<uint8_t>& hash, size_t ad_len, size_t 
    * stack buffer is fine here since the text len is public
    * and the length of the AD is probably not sensitive either.
    */
-   uint8_t final_block[GCM_BS];
-   store_be<uint64_t>(final_block, 8 * ad_len, 8 * text_len);
-   ghash_update(hash, {final_block, GCM_BS});
+   std::array<uint8_t, GCM_BS> final_block;
+
+   const uint64_t ad_bits = 8 * ad_len;
+   const uint64_t text_bits = 8 * text_len;
+   store_be(final_block, ad_bits, text_bits);
+   ghash_update(hash, final_block);
 }
 
 void GHASH::final(std::span<uint8_t> mac) {
--- a/src/lib/utils/loadstor.h
+++ b/src/lib/utils/loadstor.h
@@ -133,6 +133,57 @@ inline constexpr T load_le(InR&& in_range) {
    }
 }
 
+namespace details {
+
+/**
+* Given a statically sized range @p r of size 1, 2, 4, or 8, this finds the
+* corresponding unsigned integer type to hold the value of @p r. Note, that
+* this only maps the type, and does not perform any loading.
+*/
+template <ranges::contiguous_range<uint8_t> InR>
+consteval auto default_load_type(InR&& r) {
+   std::span s{r};
+   static_assert(decltype(s)::extent != std::dynamic_extent,
+                 "Cannot determine the output type based on a dynamically sized bytearray");
+   constexpr size_t e = decltype(s)::extent;
+
+   // clang-format off
+   using type =
+      std::conditional_t<e == 1, uint8_t,
+      std::conditional_t<e == 2, uint16_t,
+      std::conditional_t<e == 4, uint32_t,
+      std::conditional_t<e == 8, uint64_t, void>>>>;
+   // clang-format on
+
+   static_assert(
+      !std::is_void_v<type>,
+      "Cannot determine the output type based on a statically sized bytearray with length other than those: 1, 2, 4, 8");
+
+   return type{};
+}
+
+}  // namespace details
+
+/**
+* Load a little-endian unsigned integer, auto-detect the output type
+* @param in_range a statically-sized range with some bytes
+* @return T loaded from in, as a little-endian value
+*/
+template <ranges::contiguous_range<uint8_t> InR>
+inline constexpr auto load_le(InR&& in_range) {
+   return load_le<decltype(details::default_load_type(in_range)), InR>(std::forward<InR>(in_range));
+}
+
+/**
+* Load a big-endian unsigned integer, auto-detect the output type
+* @param in_range a statically-sized range with some bytes
+* @return T loaded from in, as a big-endian value
+*/
+template <ranges::contiguous_range<uint8_t> InR>
+inline constexpr auto load_be(InR&& in_range) {
+   return load_be<decltype(details::default_load_type(in_range)), InR>(std::forward<InR>(in_range));
+}
+
 /**
 * Load a big-endian word
 * @param in a pointer to some bytes
@@ -195,7 +246,7 @@ inline constexpr T load_le(const uint8_t in[], size_t off) {
 * @param outs a arbitrary-length parameter list of unsigned integers to be loaded
 */
 template <ranges::contiguous_range<uint8_t> InR, concepts::unsigned_integral... Ts>
-   requires all_same_v<Ts...>
+   requires(sizeof...(Ts) > 0) && all_same_v<Ts...>
 inline constexpr void load_be(InR&& in, Ts&... outs) {
    ranges::assert_exact_byte_length<(sizeof(Ts) + ...)>(in);
    auto load_one = [off = 0]<typename T>(auto i, T& o) mutable {
@@ -212,7 +263,7 @@ inline constexpr void load_be(InR&& in, Ts&... outs) {
 * @param outs a arbitrary-length parameter list of unsigned integers to be loaded
 */
 template <ranges::contiguous_range<uint8_t> InR, concepts::unsigned_integral... Ts>
-   requires all_same_v<Ts...>
+   requires(sizeof...(Ts) > 0) && all_same_v<Ts...>
 inline constexpr void load_le(InR&& in, Ts&... outs) {
    ranges::assert_exact_byte_length<(sizeof(Ts) + ...)>(in);
    auto load_one = [off = 0]<typename T>(auto i, T& o) mutable {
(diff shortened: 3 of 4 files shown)


src/lib/utils/ghash/ghash.cpp
```diff
@@ -178,7 +178,7 @@ void GHASH::add_final_block(secure_vector<uint8_t>& hash, size_t ad_len, size_t
    * and the length of the AD is probably not sensitive either.
    */
    uint8_t final_block[GCM_BS];
-   store_be<uint64_t>(final_block, 8 * ad_len, 8 * text_len);
+   store_be(final_block, 8 * static_cast<uint64_t>(ad_len), 8 * static_cast<uint64_t>(text_len));
```
reneme, line 181: `store_be<>` now takes a variadic number of (same-length) integers. `ad_len` and `text_len` are `size_t` (which is 32-bit on 32 bit machines). Therefore, we need to specifically cast those to `uint64_t`.

src/lib/utils/loadstor.h
```diff
@@ -82,6 +83,52 @@ inline constexpr uint64_t make_uint64(
...
+   } else {
+#if defined(BOTAN_TARGET_CPU_IS_BIG_ENDIAN)
+      return typecast_copy<T>(in);
+#elif defined(BOTAN_TARGET_CPU_IS_LITTLE_ENDIAN)
+      return reverse_bytes(typecast_copy<T>(in));
+#else
+      return [&]<size_t... i>(std::index_sequence<i...>) {
+         return ((static_cast<T>(in[i]) << ((sizeof(T) - i - 1) * 8)) | ...);
+      }
+      (std::make_index_sequence<sizeof(T)>());
```
randombit, line 104: What kind of code does this generate on CPUs where the endian is unknown at compile time? You can test this by commenting out `endian little` in `build-data/arch/x86_64.txt`

randombit (commented):
In general this looks like a nice improvement, main concern would be if there are any performance implications. I would think not for the case where endian is known at compile time so we can make use of memcpy, and the targets where it's not known (or anyway assumed) are all relatively obscure, but I'd like to not introduce a regression if we don't have to.

src/lib/utils/ghash/ghash.cpp
randombit: This looks fine, but it might be a bit cleaner to do something like

```
const uint64_t ad_bits = 8 * static_cast<uint64_t>(ad_len);
// similar for text_len
store_be(final_block, ad_bits, text_bits);
```

src/lib/utils/loadstor.h
reneme: Trying [on godbolt, with clang 12 x86_64](https://godbolt.org/z/aexr9G5rT), it boils it down to a `bswap` (for `load_be`) `mov` (for `load_le`). Same with [GCC 11.3](https://godbolt.org/z/Ev8h6fE7E). [MSVC v19](https://godbolt.org/z/jYYonqqdY) seems to make it into an unrolled loop at least.

src/lib/utils/ghash/ghash.cpp
reneme: aye aye.

src/lib/utils/loadstor.h
randombit: OK looks fine. MSVC codegen is always terrible anyway *and* the platforms where endianness is uncertain are typically obscure bi-endian systems like MIPS or SH where GCC/Clang are the dominant compilers.

(Occurs to me that we could start using `cpp`-based macro sniffing as we do already to check the compiler version which could probably eliminate this uncertain-endian state entirely. No need to address that here unless you want to.)

reneme: > No need to address that here unless you want to.

I think I'll pass. 😏 I'd rather have another look at `copy_out_vec_be`, but maybe in a follow-up.

> MSVC codegen is always terrible anyway

I was actually quite amazed that GCC/clang figured out that we're just `bswap`-ing here. I did expect an unrolled loop at most.

src/lib/utils/mem_ops.h
```diff
@@ -162,6 +162,20 @@ inline constexpr To typecast_copy(const FromT* src) noexcept
...
+   return typecast_copy<To>(src.data());
+}
+
+template <class FromT>
+inline constexpr void typecast_copy(std::span<uint8_t, sizeof(FromT)> out, FromT in)
+   requires std::is_trivial_v<typename std::decay_t<FromT>>
+{
+   typecast_copy(out.data(), in);
+}
+
```
reneme, line 178: In fact, maybe we should turn this around. Put the implementation w/ `memcpy` in here and let the existing bare-pointer overloads call those. I did it like that for the endian-conversions.

reneme:
Rebased after #3711 fixed the CI build.

src/lib/utils/mem_ops.h
reneme: done.

reneme:
Lets postpone this for a moment. I'm hoping that the iOS/Android update will bring more tools to make this easier.

randombit:
I'm going to either update the emscripten build (#3720) or just disable the emscripten CI build, at which point we should be able to rely on having most of C++20.

[*] Outside of `std::source_location`, and probably 20 other things that we just haven't hit yet...

src/lib/utils/ghash/ghash.cpp
```diff
@@ -177,9 +179,12 @@ void GHASH::add_final_block(secure_vector<uint8_t>& hash, size_t ad_len, size_t
...
    */
-   uint8_t final_block[GCM_BS];
-   store_be<uint64_t>(final_block, 8 * ad_len, 8 * text_len);
-   ghash_update(hash, {final_block, GCM_BS});
+   std::array<uint8_t, GCM_BS> final_block;
+
+   const uint64_t ad_bits = 8 * ad_len;
+   const uint64_t text_bits = 8 * text_len;
+   store_be(final_block, ad_bits, text_bits);
+   ghash_update(hash, final_block);
```
reneme, line 187: This patch was needed to explicitly convert `size_t` to `uint64_t` on 32-bit platforms.

reneme:
Rebased after #3760 conflicted with the underlying #3715.

reneme:
This will need a rebase. But let's review/merge #3715 first and get this done after.

FAlbertDev:
I'll review this PR in the following days!

src/lib/utils/loadstor.h
```diff
@@ -82,6 +83,56 @@ inline constexpr uint64_t make_uint64(
            (static_cast<uint64_t>(i6) << 8) | (static_cast<uint64_t>(i7)));
 }
 
+/**
+* Load a big-endian unsigned integer
```
FAlbertDev, line 87: This sounds like the unsigned integer itself must be stored big-endian (or little-endian). Maybe we want to chose a wording like: "Load an unsigned integer into a byte range in big-endian order".

src/lib/utils/loadstor.h
```diff
@@ -382,219 +299,133 @@ inline constexpr void load_be(T out[], const uint8_t in[], size_t count) {
...
-   out[0] = get_byte<0>(in);
-   out[1] = get_byte<1>(in);
-#endif
-}
-
-/**
-* Store a little-endian uint16_t
-* @param in the input uint16_t
-* @param out the byte array to write to
+* Store a big-endian unsigned integer
```
FAlbertDev, line 302: Same as with Load. Maybe: "Store an unsigned integer to bytes in big-endian order"?

src/lib/utils/loadstor.h
```diff
@@ -115,169 +166,87 @@ inline constexpr T load_le(const uint8_t in[], size_t off) {
...
+* Load many big-endian unsigned integers
+* @param in   a fixed-length span to some bytes
+* @param outs a arbitrary-length parameter list of unsigned integers to be loaded
 */
-template <>
-inline constexpr uint64_t load_be<uint64_t>(const uint8_t in[], size_t off) {
-   in += off * sizeof(uint64_t);
+template <ranges::contiguous_range<uint8_t> InR, concepts::unsigned_integral... Ts>
+   requires all_same_v<Ts...>
+inline constexpr void load_be(InR&& in, Ts&... outs) {
```
FAlbertDev, line 199: Do we need tests for multiple _out_ params for load_le/be (or multiple _in_ params for store_le/be)?

FAlbertDev (commented):
Nice! I'm looking forward to using these new interfaces! This will beautify many places, where load/store is used 😃

src/lib/utils/concepts.h
```diff
@@ -123,6 +222,10 @@ concept contiguous_strong_type = strong_type<T> && contiguous_container<T>;
 template <typename T>
 concept integral = std::is_integral_v<T>;
 
+// TODO: C++20 - replace with std::unsigned_integral
+template <typename T>
+concept unsigned_integral = std::is_integral_v<T> && std::is_unsigned_v<T>;
+
```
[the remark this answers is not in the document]
reneme: In fact we might be able to use it now. Xcode and NDK used to bring incomplete STL implementations for the C++20 additions -- lacking the `<concepts>` header, in particular.

src/lib/utils/loadstor.h
reneme: Here you have it right. I'm guessing that [your earlier comment](https://github.com/randombit/botan/pull/3707#discussion_r1398761064) was just a wording mixup.

src/lib/utils/concepts.h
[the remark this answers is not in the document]
reneme: Let's leave it as is for now, though. We should soon try to remove some of those self-defined concepts. There are more of them here, and I suggest doing that in a follow up.

src/lib/utils/loadstor.h
FAlbertDev: Yeah, sorry for the confusion. I meant that it sounds like the integer itself, which is the output of this function, is big-endian. However, of course, the input bytes are interpreted in big-endian order. But you are right. Probably it's clear enough 👍

src/lib/utils/loadstor.h
```diff
@@ -382,219 +350,157 @@ inline constexpr void load_be(T out[], const uint8_t in[], size_t count) {
...
-* @param x1 the second word
-* @param x2 the third word
-* @param x3 the fourth word
-* @param x4 the fifth word
-* @param x5 the sixth word
-* @param x6 the seventh word
-* @param x7 the eighth word
+* Store a big-endian unsigned integer
+* @param in the input unsigned integer
+* @return a byte array holding the integer value in big-endian byte order
```
FAlbertDev, line 497:
```suggestion
* Store a little-endian unsigned integer
* @param in the input unsigned integer
* @return a byte array holding the integer value in little-endian byte order
```

FAlbertDev (approved):
Yeah, I like the static return type deduction! We could discuss if we still want to add load/store members to the BufferSlicer and BufferStuffer. For now, however, I think this PR is ready to merge.

reneme:
Thanks for the thorough review! @randombit no particular rush with this, but I'd love to have it in (after discussing https://github.com/randombit/botan/pull/3707#issuecomment-1820766105) to get Frodo and LMS adapted to this. Kyber and Dilithium will follow.

randombit:
Change itself looks fine, and a nice cleanup overall.

I do see a minor but persistent performance regression with Blowfish (about .2 cycles/byte slower) but for whatever reason that's the only algorithm I can find that seems slower, so not a problem.

src/lib/utils/concepts.h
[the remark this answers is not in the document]
randombit: With NDK and XCode upgraded we might be able to use this now?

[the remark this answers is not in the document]
randombit: (Fine also to defer it all to a distinct PR.)

randombit (approved):

reneme:
:tada:

PR #3708, opened 2023-09-21
Doc: Update TLS Chapter in Handbook

reneme:
Since the work on the [Breathe-based documentation](#3484) has stalled for the time being, this is an update addressing the most pressing issues with the TLS documentation for Botan 3.x.


reneme:
Rebased after #3711 fixed the CI build.

randombit (approved):
Thanks!

PR #3712, opened 2023-09-23
X509: Allow writing CDP extension into certificates

martin-schiffner:
Adding the required methods to write CRL Distribution Points into the respective extension of X.509 certificates.
Also, adding tests in the X.509 unit tests file.

### DIFF
--- a/src/lib/x509/x509_ext.cpp
+++ b/src/lib/x509/x509_ext.cpp
@@ -724,10 +724,7 @@ void CRL_ReasonCode::decode_inner(const std::vector<uint8_t>& in) {
 
 std::vector<uint8_t> CRL_Distribution_Points::encode_inner() const {
    std::vector<uint8_t> output;
-   DER_Encoder(output)
-       .start_sequence()
-           .encode_list(m_distribution_points)
-       .end_cons();
+   DER_Encoder(output).start_sequence().encode_list(m_distribution_points).end_cons();
    return output;
 }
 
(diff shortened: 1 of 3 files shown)


randombit:
Thank you for your contribution! I'll have to look over the relevant RFCs to make sure we're not missing something important but this looks good from a first glance.

randombit:
@martin-schiffner Can you please rebase against current master? Currently your test code is failing to compile because the NDK 25 compiler does not support `std::range` but this should have been addressed by #3718. Also `clang-format` is unhappy; while on the whole I think your formatting is better we are trying to be consistent. You can fix this by running `make fmt` or by applying the patch that is printed in the `fmt` CI job.

I still have not checked the relevant RFCs but I'll look at this soon.

martin-schiffner:
Did a rebase and also ran `make fmt` on the code.

src/lib/x509/x509_ext.cpp
```diff
@@ -742,8 +744,14 @@ void CRL_Distribution_Points::decode_inner(const std::vector<uint8_t>& buf) {
...
-void CRL_Distribution_Points::Distribution_Point::encode_into(DER_Encoder& /*to*/) const {
-   throw Not_Implemented("CRL_Distribution_Points encoding");
+void CRL_Distribution_Points::Distribution_Point::encode_into(DER_Encoder& der) const {
+   der.start_sequence()
+      .start_cons(ASN1_Type(0), ASN1_Class::ContextSpecific)
+      .start_cons(ASN1_Type(0), ASN1_Class::ContextSpecific)
+      .add_object(ASN1_Type(6), ASN1_Class::ContextSpecific, m_point.get_first_attribute("URI"))
+      .end_cons()
+      .end_cons()
+      .end_cons();
```
randombit, line 754: This doesn't behave correctly if the distribution point doesn't contain a URL. (We'd encode an empty URL)

I'm not sure what the best fix is. The simplest thing to do would be to check that the distribution point alt name does contain a URL; if it does, do what you've done here. Otherwise, throw `Not_Implemented`. That would be fine with me and would at least make this more functional than the current state.

Also I suppose this doesn't behave correctly if the alternative name includes multiple URLs; we'd just encode the first. That would be actively bad since someone would think their distribution points were encoded but they were not. This could be fixed by either checking that there is *exactly* 1 URL (and throwing `Not_Implemented` if otherwise) or by iterating and encoding all of the URLs.

randombit (changes requested):
The encoding issue re URLs should be fixed; otherwise this looks good to go.

src/lib/x509/x509_ext.cpp
martin-schiffner: Fully agree, I've reworked the method addressing your concerns. It should now accommodate these cases.

randombit (approved):
Thanks @martin-schiffner LGTM

PR #3714, opened 2023-09-26
std::span and AlignmentBuffer<> for GHASH

reneme:
### Pull Request Dependency

* #3713

### Description

This was a side-quest of #3713. I realized that `AlignmentBuffer<>` is useful in `GHASH` and applied it along with the usual `std::span<>` moderizations.


randombit (approved):
LGTM but please update the PR description/commit message as this applies outside of just GHASH

randombit:
Ah sorry I missed this was a stack PR, n/m

PR #3715, opened 2023-09-27
Range-based mem_ops

reneme:
This adds C++20 range-based overloads to the relevant functions in `mem_ops.h`. This header is public, so I kept the legacy ptr-based overloads (also because they are still used throughout the code base). But perhaps those should be deprecated and removed at one point, @randombit?

For convenience, this adds a few helpers to `concepts.h`. Details can be found in their doxygen comments.

### DIFF
--- a/src/lib/utils/concepts.h
+++ b/src/lib/utils/concepts.h
@@ -9,10 +9,14 @@
 #ifndef BOTAN_CONCEPTS_H_
 #define BOTAN_CONCEPTS_H_
 
+#include <botan/assert.h>
+
 #include <compare>
 #include <concepts>
 #include <cstdint>
 #include <ostream>
+#include <ranges>
+#include <span>
 #include <type_traits>
 
 namespace Botan {
@@ -28,6 +32,84 @@ struct is_strong_type<Strong<Ts...>> : std::true_type {};
 
 template <typename... Ts>
 constexpr bool is_strong_type_v = is_strong_type<std::remove_const_t<Ts>...>::value;
+
+namespace ranges {
+
+/**
+ * Models a std::ranges::contiguous_range that (optionally) restricts its
+ * value_type to ValueT. In other words: a stretch of contiguous memory of
+ * a certain type (optional ValueT).
+ */
+template <typename T, typename ValueT = std::ranges::range_value_t<T>>
+concept contiguous_range = std::ranges::contiguous_range<T> && std::same_as<ValueT, std::ranges::range_value_t<T>>;
+
+/**
+ * Models a std::ranges::contiguous_range that satisfies
+ * std::ranges::output_range with an arbitrary value_type. In other words: a
+ * stretch of contiguous memory of a certain type (optional ValueT) that can be
+ * written to.
+ */
+template <typename T, typename ValueT = std::ranges::range_value_t<T>>
+concept contiguous_output_range = contiguous_range<T, ValueT> && std::ranges::output_range<T, ValueT>;
+
+/**
+ * Models a range that can be turned into a std::span<>. Typically, this is some
+ * form of ranges::contiguous_range.
+ */
+template <typename T>
+concept spanable_range = std::constructible_from<std::span<const std::ranges::range_value_t<T>>, T>;
+
+/**
+ * Find the length in bytes of a given contiguous range @p r.
+ */
+inline constexpr size_t size_bytes(spanable_range auto&& r) {
+   return std::span{r}.size_bytes();
+}
+
+/**
+ * Check that a given range @p r has a certain statically-known byte length. If
+ * the range's extent is known at compile time, this is a static check,
+ * otherwise a runtime argument check will be added.
+ *
+ * @throws Invalid_Argument  if range @p r has a dynamic extent and does not
+ *                           feature the expected byte length.
+ */
+template <size_t expected, spanable_range R>
+inline constexpr void assert_exact_byte_length(R&& r) {
+   const std::span s{r};
+   if constexpr(decltype(s)::extent != std::dynamic_extent) {
+      static_assert(s.size_bytes() == expected, "memory region does not have expected byte lengths");
+   } else {
+      BOTAN_ASSERT(s.size_bytes() == expected, "memory region does not have expected byte lengths");
+   }
+}
+
+/**
+ * Check that a list of ranges (in @p r0 and @p rs) all have the same byte
+ * lengths. If the first range's extent is known at compile time, this will be a
+ * static check for all other ranges whose extents are known at compile time,
+ * otherwise a runtime argument check will be added.
+ *
+ * @throws Invalid_Argument  if any range has a dynamic extent and not all
+ *                           ranges feature the same byte length.
+ */
+template <spanable_range R0, spanable_range... Rs>
+inline constexpr void assert_equal_byte_lengths(R0&& r0, Rs&&... rs)
+   requires(sizeof...(Rs) > 0)
+{
+   const std::span s0{r0};
+
+   if constexpr(decltype(s0)::extent != std::dynamic_extent) {
+      constexpr size_t expected_size = s0.size_bytes();
+      (assert_exact_byte_length<expected_size>(rs), ...);
+   } else {
+      const size_t expected_size = s0.size_bytes();
+      BOTAN_ARG_CHECK(((std::span<const std::ranges::range_value_t<Rs>>{rs}.size_bytes() == expected_size) && ...),
+                      "memory regions don't have equal lengths");
+   }
+}
+
+}  // namespace ranges
 
 namespace concepts {
 
--- a/src/lib/utils/mem_ops.cpp
+++ b/src/lib/utils/mem_ops.cpp
@@ -14,4 +14,11 @@ uint8_t ct_compare_u8(const uint8_t x[], const uint8_t y[], size_t len) {
    return CT::is_equal(x, y, len).value();
 }
 
+bool constant_time_compare(std::span<const uint8_t> x, std::span<const uint8_t> y) {
+   const auto min_size = CT::Mask<size_t>::is_lte(x.size(), y.size()).select(x.size(), y.size());
+   const auto equal_size = CT::Mask<size_t>::is_equal(x.size(), y.size());
+   const auto equal_content = CT::Mask<size_t>::expand(CT::is_equal(x.data(), y.data(), min_size));
+   return (equal_content & equal_size).as_bool();
+}
+
 }  // namespace Botan
--- a/src/lib/utils/mem_ops.h
+++ b/src/lib/utils/mem_ops.h
@@ -167,39 +167,98 @@ inline constexpr void copy_mem(OutR&& out, InR&& in)
    }
 }
 
+/**
+ * Copy a range of a trivially copyable type into another range of trivially
+ * copyable type of matching byte length.
+ */
+template <ranges::contiguous_output_range ToR, ranges::contiguous_range FromR>
+inline constexpr void typecast_copy(ToR&& out, FromR&& in)
+   requires std::is_trivially_copyable_v<std::ranges::range_value_t<FromR>> &&
+            std::is_trivially_copyable_v<std::ranges::range_value_t<ToR>>
+{
+   ranges::assert_equal_byte_lengths(out, in);
+   std::memcpy(std::ranges::data(out), std::ranges::data(in), ranges::size_bytes(out));
+}
+
+/**
+ * Copy a range of trivially copyable type into an instance of trivially
+ * copyable type with matching length.
+ */
+template <typename ToT, ranges::contiguous_range FromR>
+inline constexpr void typecast_copy(ToT& out, FromR&& in) noexcept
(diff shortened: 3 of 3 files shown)


randombit:
Let's hold on this for a little bit: NDK 26 and XCode 15 are both out now, I have been meaning to do a CI upgrade

src/lib/utils/mem_ops.cpp
```diff
@@ -67,10 +67,11 @@ void initialize_allocator() {
...
-uint8_t ct_compare_u8(const uint8_t x[], const uint8_t y[], size_t len) {
-   volatile uint8_t difference = 0;
+uint8_t ct_compare_u8(std::span<const uint8_t> x, std::span<const uint8_t> y) {
+   const auto min_size = CT::Mask<size_t>::is_lte(x.size(), y.size()).select(x.size(), y.size());
 
-   for(size_t i = 0; i != len; ++i) {
+   volatile uint8_t difference = 0;
+   for(size_t i = 0; i != min_size; ++i) {
       difference = difference | (x[i] ^ y[i]);
    }
```
reneme, line 77: @randombit please triple-check for constant time correctness. :)

src/lib/utils/mem_ops.h
```diff
@@ -127,6 +176,20 @@ inline constexpr void copy_mem(T* out, const T* in, size_t n)
...
+* Copy memory
+* @param out the destination array
+* @param in the source array
+*/
+template <ranges::contiguous_output_range OutR, std::ranges::contiguous_range InR>
+inline constexpr void copy_mem(OutR&& out, InR&& in)
+   requires std::is_same_v<std::ranges::range_value_t<OutR>, std::ranges::range_value_t<InR>> &&
+            std::is_trivially_copyable_v<std::ranges::range_value_t<InR>>
+{
+   ranges::assert_equal_byte_lengths(out, in);
```
reneme, line 189: If both `in` and `out` have static extents (e.g. static-extent `std::span<>` and/or `std::array<>`), `assert_equal_byte_lengths` will only verify the extents statically. There won't be another check during runtime.

src/lib/utils/mem_ops.h
```diff
@@ -61,24 +63,57 @@ class Allocator_Initializer final {
...
+/**
+* Scrub memory contents in a way that a compiler should not elide,
+* using some system specific technique. Note that this function might
+* not zero the memory.
+*
+* @param data  the data region to be scrubbed
+*/
+void secure_scrub_memory(ranges::contiguous_output_range auto&& data) {
+   secure_scrub_memory(std::ranges::data(data), ranges::size_bytes(data));
+}
```
reneme, line 75: TIL: simple concept annotations can be added to parameters without the verbose `template<typename T> requires some_concept<T>` syntax.

devnexen:
So I gave a shot with the NDK 26 locally and got this

`In file included from build/include/botan/buf_comp.h:11:
build/include/botan/concepts.h:148:94: error: use of undeclared identifier 'uint8_t'
   contiguous_container<T> && resizable_container<T> && std::same_as<typename T::value_type, uint8_t>;
                                                                                             ^
In file included from src/lib/base/buf_comp.cpp:7:
build/include/botan/buf_comp.h:76:27: error: no type named 'resizable_byte_buffer' in namespace 'Botan::concepts'
      template <concepts::resizable_byte_buffer T = secure_vector<uint8_t>>
                ~~~~~~~~~~^
build/include/botan/buf_comp.h:76:75: error: expected '(' for function-style cast or type construction
      template <concepts::resizable_byte_buffer T = secure_vector<uint8_t>>`

once fixed locally, everything else passes.

reneme:
> once fixed locally, everything else passes.

I'm guessing this was just lacking an include to `<cstdint>` somewhere, right?

randombit:
On thing we should consider here is if we want to have these additions in the public `mem_ops.h` at all.

This header is kind of a grab bag of various things, some of which are required to be public for API or ABI reasons (especially `allocate_memory` which is used by the `secure_vector`) but much of this is not really intended to be for use by applications.

It might be better to add a new internal header which contains the span versions, convert the library internals to use this, and then deprecate most of the functions in `mem_ops.h` which were never intended for end users.

src/lib/utils/mem_ops.h
randombit: Nice!

reneme:
> One thing we should consider here is if we want to have these additions in the public mem_ops.h at all.

I think it makes a lot of sense to split this header for a cleaner structure. The challenge with making it private, though, is that it'll prevent us from using these helpers in templates of other (public) headers.

randombit:
This is true. I'll take the first step here - split the parts that are required to implement `secmem.h` into a new header `allocator.h`, and modify `secmem.h` to no longer include `mem_ops.h` [*] This will force all headers or source files which use functionality from this header to include the header explicitly. At that point it will be easier to evaluate how to split everything else out.

[*] And to maintain compatability `mem_ops.h` just additionally includes `allocator.h`

lieser:
> The challenge with making it private, though, is that it'll prevent us from using these helpers in templates of other (public) headers.

A solution could be to keep what is needed in the public header, but in a e.g. a `detailed` namespace. That should make it clear to users that it is not part of the public API.

reneme:
Rebased to master after #3760 conflicted with the changes in here.

randombit:
@reneme This needs a rebase with #3752 merged but I'll review after

reneme:
> This needs a rebase with https://github.com/randombit/botan/pull/3752 merged but I'll review after

Done. Will also do a self-review tomorrow. Too many ghosts around here to concentrate right now.

# 👻 

... Happy Halloween.

src/lib/utils/mem_ops.h
```diff
@@ -110,39 +157,114 @@ inline constexpr void copy_mem(T* out, const T* in, size_t n)
...
+
+/**
+ * Copy a rande of a trivially copyable type into another range of trivially
+ * copyable type of matching byte length.
+ */
+template <ranges::contiguous_output_range ToR, ranges::contiguous_range FromR>
+inline constexpr void typecast_copy(ToR&& out, FromR&& in)
+   requires std::is_trivially_copyable_v<std::ranges::range_value_t<FromR>> &&
+            std::is_trivially_copyable_v<std::ranges::range_value_t<ToR>>
```
reneme, line 183: This is very closely related to C++20's [`std::bit_cast`](https://en.cppreference.com/w/cpp/numeric/bit_cast), right? I didn't find a way to use `std::bit_cast` on arbitrary ranges, though. Probably that's by design, as it is supposed to work closely with the type system.

But given that Botan's `typecast_copy` and STL's `bit_cast` are so closely related: Perhaps we should adopt the naming here and deprecate `typecast_copy` in the public API. For instance, we could call these functions `bitcast_copy`.

src/lib/utils/mem_ops.h
```diff
@@ -44,15 +47,39 @@ namespace Botan {
...
+
 /**
 * Memory comparison, input insensitive
 * @param x a pointer to an array
 * @param y a pointer to another array
 * @param len the number of Ts in x and y
 * @return 0xFF iff x[i] == y[i] forall i in [0...n) or 0x00 otherwise
 */
+BOTAN_DEPRECATED("This function is deprecated, use constant_time_compare()")
 BOTAN_PUBLIC_API(2, 9) uint8_t ct_compare_u8(const uint8_t x[], const uint8_t y[], size_t len);
```
reneme, line 71: We don't use it anywhere in the code base anymore. Okay to deprecate?

src/lib/utils/concepts.h
```diff
@@ -29,6 +33,93 @@ struct is_strong_type<Strong<Ts...>> : std::true_type {};
...
+ */
+template <size_t expected, spanable_range R>
+inline constexpr void assert_exact_byte_length(R&& r) {
+   const std::span s{r};
+   if constexpr(has_static_extent<R>()) {
+      static_assert(s.size_bytes() == expected, "memory region does not have expected byte lengths");
+   } else {
+      BOTAN_ARG_CHECK(s.size_bytes() == expected, "memory region does not have expected byte lengths");
+   }
+}
```
FAlbertDev, line 93: It's interesting, because a call of this method works, but only with type deduction. For example:
```cpp
    std::array<uint8_t,32> a;
    // Does not work 
    assert_exact_byte_length<32, std::array<uint8_t,32>>(a);
    // Works
    assert_exact_byte_length<32>(a);
```
It seems the input type is implicitly converted to a span.

src/lib/utils/concepts.h
```diff
@@ -29,6 +33,93 @@ struct is_strong_type<Strong<Ts...>> : std::true_type {};
...
+
+/**
+ * Check that a list of ranges (in @p r0 and @p rs) all have the same byte
+ * lengths. If all ranges' extents are known at compile time, this will be a
+ * static check, otherwise a runtime argument check will be added.
+ *
+ * @throws Invalid_Argument  if any range has a dynamic extent and not all
+ *                           ranges feature the same byte length.
+ */
+template <spanable_range R0, spanable_range... Rs>
```
FAlbertDev, line 103: Same for this one. I think a fix of `has_static_extent()` should fix all of this.

src/lib/utils/mem_ops.h
```diff
@@ -44,15 +47,39 @@ namespace Botan {
...
 
+#endif
+
+/**
+ * Memory comparison, input insensitive
+ * @param x a range of bytes
+ * @param y another range of bytes
+ * @return true iff x and y have equal lengths and x[i] == y[i] forall i in [0...n)
+ */
+BOTAN_PUBLIC_API(3, 3) bool constant_time_compare(std::span<const uint8_t> x, std::span<const uint8_t> y);
```
FAlbertDev, line 81: Do you think we want to move this to ct_utils.h?

src/lib/utils/mem_ops.h
```diff
@@ -226,75 +343,110 @@ size_t buffer_insert(std::vector<T, Alloc>& buf, size_t buf_offset, const std::v
...
+* @param in2 the second input buffer
+* @param length the length of the three buffers
+*/
+inline void xor_buf(uint8_t out[], const uint8_t in[], const uint8_t in2[], size_t length) {
+   // simply assumes that *out, *in, and *in2 point to "length" allocated bytes at least
+   xor_buf(std::span{out, length}, std::span{in, length}, std::span{in2, length});
+}
+
+// TODO: deprecate and replace, use .subspan()
 inline void xor_buf(std::span<uint8_t> out, std::span<const uint8_t> in, size_t n) {
```
FAlbertDev, line 429: I agree this function interface is somehow weird...

FAlbertDev (commented):
All in all, LGTM. Some improvements of this PR and #3707 will be very helpful for implementing Classic McEliece :)

src/lib/utils/concepts.h
```diff
@@ -29,6 +33,93 @@ struct is_strong_type<Strong<Ts...>> : std::true_type {};
...
+   return std::span{r}.size_bytes();
+}
+
+/**
+ * Check whether a given range type has a length known at compile time.
+ */
+template <spanable_range R>
+bool consteval has_static_extent() {
+   return decltype(std::span{std::declval<R>()})::extent != std::dynamic_extent;
+}
```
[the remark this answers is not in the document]
reneme: I really like your approach, apart from being fairly verbose. In the hope that we won't need this in many more places: I had another pass at this, and ended up simply spelling it out, like:

```C++
const std::span s{r};
if constexpr(decltype(s)::extent != std::dynamic_extent) { /*...*/ }
// ...
```

No idea why `has_static_extent<R>()` seemed  to work for me, before.

src/lib/utils/concepts.h
reneme: That's somewhat expected, because of the forwarding reference (`R&&`). When passing `a` with type deduction, you'll end up with `R = std::array<>&` (an lvalue reference). I.e. If you were to add the `&` to your explicit declaration, it should work. Here's some reading: https://medium.com/factset/modern-c-in-depth-perfect-forwarding-570f242261f8

src/lib/utils/mem_ops.h
reneme: Probably. This PR has more than enough complexity, though. So, I suggest we do that as a follow up.

FAlbertDev (approved):
👍

reneme:
@randombit Could we go ahead with this? I'd love to have this and #3707 merged before doing another pass over the Kyber/Dilithium implementations.

src/lib/utils/mem_ops.h
randombit: Yeah fine to deprecate

src/lib/utils/mem_ops.h
randombit: ct_utils is our internal header while `constant_time_compare` is *intended* for library users since this comes up for application also.

All internal uses should use `CT::Mask` instead - I have a WIP branch on this somewhere.

randombit (approved):
Yeah fine to merge, sorry I've been sick since early Nov so way behind on reviews right now.

reneme:
> sorry I've been sick since early Nov so way behind on reviews right now

No worries at all. Hope you're feeling better now. 😄 
Thanks for the review

Issue #3105, opened 2022-12-03
Implement LMS


securitykernel:
We will implement LMS from [NIST SP 800-208](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-208.pdf) during the [BSI](https://www.bsi.bund.de/EN) Project 481.

~~Preliminary schedule for the implementation is Jun '23 - Aug '23.~~

Updated schedule below due to slight project plan rescheduling.

securitykernel:
Updated schedule for the implementation is Aug '23 - Oct '23.

randombit:
Merged for 3.5.0 :tada:

PR #3716, opened 2023-09-27
HSS-LMS Signature Algorithm Implementation

FAlbertDev:
### Pull Request Dependencies

* #3869 

### Description

This is adding the Hierarchical Signature System (HSS) with Leighton-Micali Hash-Based Signatures (LMS) as defined in [RFC 8554](https://datatracker.ietf.org/doc/html/rfc8554). 

The first commit (9dddcf6) contains some preparations, mainly common utility functions, and minor cleanup. HSS-LMS is implemented in the second commit (ba36c5e).

For this algorithm, we tried to find a way to generalize the logic to construct a Merkle tree, which is used in all hash-based signature algorithms so far. We introduce the tree_hash.h containing a generalized (and highly flexible) logic for creating Merkle trees. We only apply it for HSS-LMS in this PR but aim to apply it for (at least) SPHINCS+ in a follow-up PR. 

If you are interested, you can drop some comments and suggestions.

#3105


src/tests/test_hss_lms.cpp
```diff
@@ -0,0 +1,232 @@
...
+         }
+         return result;
+      }
+
+      std::vector<Test::Result> run() final {
+         return {test_too_short_signature(), test_too_short_private_key(), test_too_short_public_key()};
+      }
+};
+
+// TODO: this currently only calls APIs not covered by other tests.
```
lieser, line 183: Probably adapt this a little if https://github.com/randombit/botan/pull/3702 gets merged first.

src/lib/pubkey/hss_lms/hss.cpp
```diff
@@ -0,0 +1,421 @@
...
+
+bool HSS_LMS_PublicKeyInternal::verify_signature(std::span<const uint8_t> msg, const HSS_Signature& sig) const {
+   if(HSS_Level(checked_cast_to<uint32_t>(sig.Nspk())) + 1 != L()) {
+      // HSS levels in the public key does not match with the signature's
+      return false;
+   }
+
+   const LMS_PublicKey* lms_pk = &lms_pub_key();
+
+   // Verify the signature by the above layer over the LMS public keys for layer 1 to Nspk.
```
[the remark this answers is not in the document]
FAlbertDev: Indeed, I only did this check when parsing a private key, but not when creating one or when verifying a signature. I'll add both checks, thanks!

src/lib/pubkey/hss_lms/hss_lms.cpp
```diff
@@ -0,0 +1,175 @@
...
+      }
+
+      bool is_valid_signature(const uint8_t* sig, size_t sig_len) override {
+         std::vector<uint8_t> message_to_verify = std::move(m_msg_buffer);
+         m_msg_buffer = std::vector<uint8_t>();
+         try {
+            const auto signature = HSS_Signature::from_bytes_or_throw({sig, sig_len});
+            bool sig_valid = m_public->verify_signature(message_to_verify, signature);
+            return sig_valid;
+         } catch(const Decoding_Error&) {
```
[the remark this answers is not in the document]
FAlbertDev: IMO, a caller does not care about the reason for a signature to be invalid, so I think catching this here is the right decision.

src/lib/pubkey/hss_lms/hss_lms.cpp
```diff
@@ -0,0 +1,175 @@
...
+   public:
+      HSS_LMS_Signature_Operation(std::shared_ptr<HSS_LMS_PrivateKeyInternal> private_key,
+                                  std::shared_ptr<HSS_LMS_PublicKeyInternal> public_key) :
+            m_private(std::move(private_key)), m_public(std::move(public_key)) {}
+
+      void update(const uint8_t msg[], size_t msg_len) override {
+         m_msg_buffer.insert(m_msg_buffer.end(), msg, msg + msg_len);
+      }
+
+      secure_vector<uint8_t> sign(RandomNumberGenerator&) override { return m_private->sign(m_msg_buffer); }
```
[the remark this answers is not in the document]
FAlbertDev: Good catch!

src/lib/pubkey/hss_lms/hss_lms.h
```diff
@@ -0,0 +1,167 @@
...
+ - Hash signatures in tests
+ - Full list of OTS tests (possible?)
+ - Process TODOs in tests
+
+ Final Cleanup:
+ - Minimize all includes
+
+
+
+*/
```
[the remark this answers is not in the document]
FAlbertDev: These TODOs have already been processed ;)

FAlbertDev:
This PR is now ready! Any review is welcome ❤️

src/lib/utils/concepts.h
```diff
@@ -29,6 +29,18 @@ struct is_strong_type<Strong<Ts...>> : std::true_type {};
...
+ * Checks whether a strong type has the @p Capability included in its @p Tags type pack.
+ */
+template <typename Capability, typename T, typename... Tags>
+constexpr auto strong_type_has_capability(Strong<T, Tags...>) {
+   if constexpr((std::is_same_v<Capability, Tags> || ...)) {
+      return std::true_type();
+   } else {
+      return std::false_type();
+   }
+}
```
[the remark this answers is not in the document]
reneme: And another architecture suggestion: Why not let the `Strong<>` template tell you what it can do via a static template method in `class Strong<>`:

```C++
      template <typename CapabilityT>
      constexpr static bool has_capability() {
         return (std::is_same_v<CapabilityT, Capabilities> || ...);
      }
```

See here for a quick sketch: https://godbolt.org/z/jxebM7fvK

src/lib/utils/concepts.h
```diff
@@ -117,12 +129,38 @@ concept strong_type = is_strong_type_v<T>;
 template <class T>
 concept contiguous_strong_type = strong_type<T> && contiguous_container<T>;
 
+template <typename T, typename Capability>
+concept strong_type_with_capability = requires(T a) {
+                                         { strong_type_has_capability<Capability>(a) } -> std::same_as<std::true_type>;
+                                      };
```
reneme, line 135: Given my suggestion of the static template method in `Strong<>` above, this would become:

```C++
template <typename T, typename Capability>
concept strong_type_with_capability = T::template has_capability<Capability>();
```

Godbolt: https://godbolt.org/z/8sPcc3vbo

src/lib/utils/loadstor.h
```diff
@@ -228,6 +229,28 @@ inline constexpr uint64_t load_le<uint64_t>(const uint8_t in[], size_t off) {
...
+* Load a big-endian enum value
+* @param in a pointer to some bytes
+* @param off an offset into the array
+* @return off'th enum of in, as a big-endian value
+*/
+template <concepts::enum_type T>
+inline constexpr T load_be(const uint8_t in[], size_t off) {
+   return static_cast<T>(load_be<std::underlying_type_t<T>>(in, off));
+}
+
```
reneme, line 253: Nice! This has some implications/synergies on #3707. I'll look into that.

src/lib/utils/stl_util.h
```diff
@@ -206,6 +218,17 @@ class BufferStuffer {
...
+/**
+ * @brief Size guarded copy using spans.
+ *
+ * @param out output buffer
+ * @param in input buffer
+ */
+inline void copy_into(std::span<uint8_t> out, std::span<const uint8_t> in) {
+   BOTAN_ASSERT_NOMSG(in.size() == out.size());
+   std::copy(in.begin(), in.end(), out.begin());
+}
```
reneme, line 230: We already have `copy_mem` in `mem_ops.h`. #3715 will introduce that for ranges as well.

src/lib/utils/stl_util.h
```diff
@@ -150,6 +152,11 @@ class BufferSlicer final {
          std::copy(data.begin(), data.end(), sink.begin());
       }
 
+      template <typename T>
+      auto copy_be() {
+         return load_be<T>(take(sizeof(T)).data(), 0);
+      }
+
```
reneme, line 159: I like. Let's introduce little-endian right away.

src/lib/utils/tree_hash.h
```diff
@@ -0,0 +1,269 @@
...
+concept tree_node = contiguous_container<T>;
+
+/**
+ * @brief An index of a node in a layer.
+ *
+ * This is a separate index for each layer.
+ * The left most node of a layer has the index 0.
+ */
+template <typename T>
+concept tree_node_index = strong_type_with_capability<T, EnableArithmeticWithPlainNumber>;
```
reneme, line 36: Just an alternative suggestion, because `strong_type_with_capability` seems to be so verbose that we'll probably always wrap it into more specific concepts: Maybe not have `strong_type_with_capability` at all but formulate those as a conjunction:

```suggestion
template <typename T>
concept tree_node_index = is_strong_type<T>() && T::template has_capability<EnableArithmeticWithPlainNumber>;
```

src/lib/utils/tree_hash.h
```diff
@@ -0,0 +1,269 @@
...
+/**
+ * @brief A layer in a Tree.
+ *
+ * The bottom layer is the layer 0.
+ */
+template <typename T>
+concept tree_layer_index = strong_type_with_capability<T, EnableArithmeticWithPlainNumber>;
+
+template <strong_type S>
+struct strong_span_type {};
```
reneme, line 47: ?

src/lib/utils/tree_hash.h
```diff
@@ -0,0 +1,269 @@
...
+inline void treehash(
+   StrongSpan<TreeNode> out_root,
+   std::optional<AuthPathSS> out_auth_path,
+   std::optional<TreeNodeIndex> leaf_idx,
+   size_t node_size,
+   TreeLayerIndex total_tree_height,
+   uint32_t idx_offset,
+   concepts::tree_hash_node_pair<TreeNodeIndex, TreeLayerIndex, Address, StrongSpan<TreeNode>> auto node_pair_hash,
+   concepts::tree_gen_leaf<TreeNodeIndex, TreeLayerIndex, Address, StrongSpan<TreeNode>> auto gen_leaf,
+   Address& tree_address) {
```
reneme, line 123: # 😲 

Any chance we could wrap that into a reasonable class that removes some complexity from this interface? Off the top of my head, I don't know enough to give useful suggestions. Though, on first glance, that seems like a nightmare to use and/or understand.

Can we go through the usage of this in a pair-programming session at one point please?

reneme (commented):
No full review at all. Mainly just looked at the stuff in `src/lib/utils`.

randombit (commented):
Started a review. I haven't gotten into the core implementation yet so this is just some initial comments, I will do a full review asap.

src/lib/utils/stl_util.h
FAlbertDev: Yeah. Well, the current `copy_mem` only works on pointers by passing the size, but we wanted a function working on spans with an integrated size check. The implementation with ranges is exactly what we need. What do you suggest to do until then? I would keep `copy_into` for now and remove it after #3715 is merged.

src/lib/utils/concepts.h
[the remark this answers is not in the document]
FAlbertDev: Nice, it works! Thanks!

src/lib/utils/concepts.h
FAlbertDev: 👍

src/lib/utils/stl_util.h
FAlbertDev: Nice, I'll add little-endian!

src/lib/utils/tree_hash.h
FAlbertDev: Do you think so? I would guess that it's not a rare use case. And since strong types are the only classes with Capabilities it is reasonable IMO to have a concept for this. What do you think?

src/lib/utils/tree_hash.h
FAlbertDev: I missed this artifact. Thanks!

src/lib/utils/tree_hash.h
FAlbertDev: Yeah, I admit that this construct looked much better in my head. The goal of this template hell is that the calling party only calls this function with its own strong types while all types of `treehash` are deduced automatically without the need to pass any template parameters. But yeah, I would love to get your input on this in a pair-programming session!

src/build-data/oids.txt
```diff
@@ -29,6 +29,14 @@
 1.3.6.1.4.1.25258.1.10.2 = Dilithium-6x5-AES-r3
 1.3.6.1.4.1.25258.1.10.3 = Dilithium-8x7-AES-r3
 
+# HSS-LMS
+# draft-gazdag-x509-hash-sigs-01
+1.2.840.113549.1.9.16.3.17 = HSS-LMS
+
+# HSS-LMS private key (since the format of an HSS-LMS private key is not specified,
+#  we define an OID for the one we use)
+1.3.6.1.4.1.25258.1.13 = HSS-LMS-Private-Key
```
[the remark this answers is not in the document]
FAlbertDev: Yeah, thanks!

src/lib/utils/tree_hash.h
```diff
@@ -0,0 +1,269 @@
+/**
+ * Treehash logic used for hash-based signatures
```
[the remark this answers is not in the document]
FAlbertDev: Agreed!

src/lib/pubkey/hss_lms/lms.h
```diff
@@ -0,0 +1,339 @@
...
+ * @brief Enum of available LMS algorithm types.
+ *
+ * See RFC 8554 Section 5.1.
+ */
+enum class LMS_Algorithm_Type : uint32_t {
+   // --- RFC 8554 ---
+   RESERVED = 0x00,
+
+   // SHA-256 based
+   SHA256_M32_H5 = 0x05,
```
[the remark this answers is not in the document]
FAlbertDev: Good point 👍

src/lib/pubkey/hss_lms/hss_lms_utils.h
```diff
@@ -0,0 +1,75 @@
...
+      /**
+       * @brief Create a hash value using the preconfigured prefix and a @p seed
+       */
+      void gen(std::span<uint8_t> out, HashFunction& hash, std::span<const uint8_t> seed) const;
+
+   private:
+      std::vector<uint8_t> m_input_buffer;
+      std::span<uint8_t> m_q;
+      std::span<uint8_t> m_i;
+      uint8_t* m_j;
```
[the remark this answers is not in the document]
FAlbertDev: 👍

src/lib/pubkey/hss_lms/hss_lms.h
```diff
@@ -0,0 +1,153 @@
...
+      std::unique_ptr<Public_Key> public_key() const override;
+
+      AlgorithmIdentifier pkcs8_algorithm_identifier() const override;
+
+      bool stateful_operation() const override { return true; }
+
+      /**
+       * Retrieves the number of remaining signatures for this private key.
+       */
+      uint64_t remaining_signatures() const;
```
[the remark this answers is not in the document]
FAlbertDev: Yeah, good idea. Much better than forcing applications to down-cast the private key!

src/lib/pubkey/hss_lms/lms.h
```diff
@@ -0,0 +1,339 @@
...
+#include <string_view>
+#include <utility>
+#include <vector>
+
+namespace Botan {
+
+/**
+ * @brief Enum of available LMS algorithm types.
+ *
+ * See RFC 8554 Section 5.1.
```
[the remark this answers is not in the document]
FAlbertDev: Yeah. I didn't know you have experience with LMS. Then I'll look forward to your feedback :)

src/tests/test_hss_lms.cpp
lieser: The PR is merged, so we should add LMS to it (after a rebase if that is not yet done)

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,351 @@
...
+   TreeAddress lms_tree_address(lms_sk.lms_params().h());
+
+   treehash(out_root,
+            out_auth_path,
+            leaf_idx,
+            lms_sk.lms_params().m(),
+            LMS_TreeLayerIndex(lms_sk.lms_params().h()),
+            0,
+            hash_pair_func,
+            gen_leaf,
```
reneme, line 114: As stated above: When these are lambdas that have a closured `unique_ptr<>` they won't be able to be copied.

```suggestion
            std::move(hash_pair_func),
            std::move(gen_leaf),
```

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,351 @@
...
+   LMS_Tree_Node root(lms_params.m());
+
+   compute_root(StrongSpan<LMS_Tree_Node>(root),
+                sig.auth_path(),
+                sig.q(),
+                StrongSpan<const LMS_Tree_Node>(tmp),
+                lms_params.m(),
+                LMS_TreeLayerIndex(lms_params.h()),
+                0,
+                hash_pair_func,
```
reneme, line 341: As stated above: When these are lambdas that have a closured `unique_ptr<>` they won't be able to be copied.

```suggestion
                std::move(hash_pair_func),
```

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,351 @@
...
+      return LMS_Params(type, "Truncated(SHA-256,192)", h);
+   }
+   if(type >= LMS_Algorithm_Type::SHAKE_M32_H5 && type <= LMS_Algorithm_Type::SHAKE_M32_H25) {
+      uint8_t h = 5 * (type_value - checked_cast_to<uint8_t>(LMS_Algorithm_Type::SHAKE_M32_H5) + 1);
+      return LMS_Params(type, "SHAKE-256(256)", h);
+   }
+   if(type >= LMS_Algorithm_Type::SHAKE_M24_H5 && type <= LMS_Algorithm_Type::SHAKE_M24_H25) {
+      uint8_t h = 5 * (type_value - checked_cast_to<uint8_t>(LMS_Algorithm_Type::SHAKE_M24_H5) + 1);
+      return LMS_Params(type, "SHAKE-256(192)", h);
+   }
```
reneme, line 138: I'm both scared and amazed. But frankly, I think enum-arithmetics is married to pointer-arithmetics and they made a bunch of children called "obscure bug" and "undefined behavior". 🤡

Would a `switch()` in some helper function really be so cumbersome? 😨

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,351 @@
...
+   } else if(hash_name == "SHAKE-256(256)") {
+      base_type = LMS_Algorithm_Type::SHAKE_M32_H5;
+   } else if(hash_name == "SHAKE-256(192)") {
+      base_type = LMS_Algorithm_Type::SHAKE_M24_H5;
+   } else {
+      throw Decoding_Error("Unsupported hash function");
+   }
+   auto type = checked_cast_to<LMS_Algorithm_Type>(checked_cast_to<uint8_t>(base_type) + type_offset);
+   return LMS_Params(type, hash_name, h);
+}
```
reneme, line 161: Same as above: enum-arithmetics, really? It's beautiful though, I'll admit that.
Please, leave a big and blinking warning in the `enum` declaration that you're doing this, at least. Future-you will be grateful.

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,351 @@
...
+   BOTAN_ASSERT_NOMSG(sig_stuffer.full());
+
+   TreeAddress lms_tree_address(lms_params().h());
+   LMS_Tree_Node pk_buffer(lms_params().m());
+   lms_treehash(StrongSpan<LMS_Tree_Node>(pk_buffer.get()), auth_path_buffer, q, *this);
+
+   return LMS_PublicKey(lms_params(), lmots_params(), identifier(), std::move(pk_buffer));
+}
+
+LMS_PublicKey LMS_PublicKey::from_bytes_of_throw(BufferSlicer& slicer) {
```
reneme, line 191:
```suggestion
LMS_PublicKey LMS_PublicKey::from_bytes_or_throw(BufferSlicer& slicer) {
```

Please don't forget the header.

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,351 @@
...
+   size_t total_remaining_bytes = slicer.remaining();
+   // Alg. 6. 1. (4 bytes are sufficient until the next check)
+   if(total_remaining_bytes < sizeof(LMS_Algorithm_Type)) {
+      throw Decoding_Error("To few bytes while parsing LMS public key.");
```
reneme, line 195:
```suggestion
      throw Decoding_Error("Too few bytes while parsing LMS public key.");
```

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,351 @@
...
+   }
+   // Alg. 6. 2.a.
+   auto lms_type = slicer.copy_be<LMS_Algorithm_Type>();
+   // Alg. 6. 2.c.
+   auto lms_params = LMS_Params::create_or_throw(lms_type);
+   // Alg. 6. 2.d.
+   if(total_remaining_bytes < size(lms_params)) {
+      throw Decoding_Error("To few bytes while parsing LMS public key.");
```
reneme, line 203:
```suggestion
      throw Decoding_Error("Too few bytes while parsing LMS public key.");
```

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,351 @@
...
+   std::vector<uint8_t> bytes(size(lms_params()));
+   BufferStuffer stuffer(bytes);
+
+   stuffer.append_be(lms_params().algorithm_type());
+   stuffer.append_be(lmots_params().algorithm_type());
+   stuffer.append(identifier());
+   stuffer.append(m_lms_root);
+   BOTAN_ASSERT_NOMSG(stuffer.full());
+
+   return bytes;
```
reneme, line 231: Note to self: if an overload of `store_be()` that returns a properly-sized `std::array<uint8_t, x>` was available, this could have been formulated as:

```C++
return concat_at<std::vector<uint8_t>>(store_be(lms_params(return concat_at<std::vector<uint8_t>>(
    store_be(lms_params().algorithm_type()),
    store_be(lmots_params().algorithm_type()),
    identifier(),
    m_lms_root);
```

which would have been nice.

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+size_t LMS_PublicKey::size(const LMS_Params& lms_params) {
+   return sizeof(LMS_Algorithm_Type) + sizeof(LMOTS_Algorithm_Type) + LMS_IDENTIFIER_LEN + lms_params.m();
+}
+
+LMS_Signature LMS_Signature::from_bytes_or_throw(BufferSlicer& slicer) {
+   size_t total_remaining_bytes = slicer.remaining();
+   // Alg. 6a 1. (next 4 bytes are checked in LMOTS_Signature::from_bytes_or_throw)
+   if(total_remaining_bytes < sizeof(LMS_Tree_Node_Idx)) {
+      throw Decoding_Error("To few signature bytes while parsing LMS signature.");
```
reneme, line 251:
```suggestion
      throw Decoding_Error("Too few signature bytes while parsing LMS signature.");
```

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,351 @@
...
+   if(slicer.remaining() < sizeof(LMS_Algorithm_Type)) {
+      throw Decoding_Error("To few signature bytes while parsing LMS signature.");
+   }
+   // Alg. 6a 2.f.
+   auto lms_type = slicer.copy_be<LMS_Algorithm_Type>();
+   // Alg. 6a 2.h.
+   LMS_Params lms_params = LMS_Params::create_or_throw(lms_type);
+   // Alg. 6a 2.i. (signature is not exactly [...] bytes long)
+   if(total_remaining_bytes < size(lms_params, lmots_params)) {
+      throw Decoding_Error("To few signature bytes while parsing LMS signature.");
```
reneme, line 269:
```suggestion
      throw Decoding_Error("Too few signature bytes while parsing LMS signature.");
```

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,351 @@
...
+   }
+   // Alg. 6a 2.a.
+   auto q = slicer.copy_be<LMS_Tree_Node_Idx>();
+
+   // Alg. 6a 2.b.-e.
+   auto lmots_sig = LMOTS_Signature::from_bytes_or_throw(slicer);
+   LMOTS_Params lmots_params = LMOTS_Params::create_or_throw(lmots_sig.algorithm_type());
+
+   if(slicer.remaining() < sizeof(LMS_Algorithm_Type)) {
+      throw Decoding_Error("To few signature bytes while parsing LMS signature.");
```
reneme, line 261:
```suggestion
      throw Decoding_Error("Too few signature bytes while parsing LMS signature.");
```

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,351 @@
...
+   LMS_Params lms_params = LMS_Params::create_or_throw(lms_type);
+   // Alg. 6a 2.i. (signature is not exactly [...] bytes long)
+   if(total_remaining_bytes < size(lms_params, lmots_params)) {
+      throw Decoding_Error("To few signature bytes while parsing LMS signature.");
+   }
+
+   // Alg. 6a 2.j.
+   auto auth_path = slicer.take<LMS_AuthenticationPath>(lms_params.m() * lms_params.h());
+
+   return LMS_Signature(q, std::move(lmots_sig), lms_type, LMS_AuthenticationPath(auth_path));
```
reneme, line 275:
```suggestion
   return LMS_Signature(q, std::move(lmots_sig), lms_type, std::move(auth_path));
```

... looks like some refactoring left over to me.

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,351 @@
...
+   if(lms_params().algorithm_type() != sig.lms_type() ||
+      lmots_params().algorithm_type() != sig.lmots_sig().algorithm_type()) {
+      return std::nullopt;
+   }
+
+   const LMS_Params lms_params = LMS_Params::create_or_throw(sig.lms_type());
+   const LMOTS_Signature& lmots_sig = sig.lmots_sig();
+   const LMOTS_Params lmots_params = LMOTS_Params::create_or_throw(lmots_sig.algorithm_type());
+   const LMOTS_K Kc = lmots_compute_pubkey_from_sig(lmots_sig, msg, identifier(), sig.q());
+   const auto hash = HashFunction::create_or_throw(lms_params.hash_name());
```
reneme, line 321: Not sure at all, but: should we catch those potential exceptions and return `std::nullopt`? Personally, I'd be surprised if a method threw if it claimed to "just maybe" return a value by having a `std::optional<>` return type.

reneme (changes requested):
Please also have a look into `src/tests/unit_x509.cpp` where you could register LMS.

src/lib/pubkey/hss_lms/hss_lms_utils.h
```diff
@@ -0,0 +1,79 @@
...
+
+      /**
+       * @brief Specify the value for the u16str(i) hash input field
+       */
+      void set_i(uint16_t i) { store_be(i, m_i.data()); }
+
+      /**
+       * @brief Specify the value for the u8str(j) hash input field
+       */
+      void set_j(uint8_t j) { *m_j = j; }
```
reneme, line 48: Maybe its worth to hold the locations of `q`, `j` and `i`. But perhaps we should calculate the spans inline? Does this have an impact on the performance?

```suggestion
      void set_q(uint32_t q) { store_be(q, std::span(m_input_buffer).last<7>().first<4>().data()); }

      /**
       * @brief Specify the value for the u16str(i) hash input field
       */
      void set_i(uint16_t i) { store_be(i, std::span(m_input_buffer).last<3>().first<2>().data()); }

      /**
       * @brief Specify the value for the u8str(j) hash input field
       */
      void set_j(uint8_t j) { m_input_buffer.back() = j; }
```

reneme: At least the `m_j` raw pointer can be avoided like that without a performance impact, for sure.

src/lib/pubkey/hss_lms/hss.cpp
```diff
@@ -0,0 +1,411 @@
...
+   stuffer.append_be(get_idx());
+
+   for(HSS_Level layer(1); layer <= hss_params().L(); ++layer) {
+      const auto& params = hss_params().params_at_level(layer - 1);
+      stuffer.append_be(params.lms_params().algorithm_type());
+      stuffer.append_be(params.lmots_params().algorithm_type());
+   }
+
+   sk_bytes.insert(sk_bytes.end(), seed().begin(), seed().end());
+   sk_bytes.insert(sk_bytes.end(), identifier().begin(), identifier().end());
```
reneme, line 184: Can't we use the `stuffer` here? If no: leave a note to future-us. :)

reneme (commented):
Some more comments. Really like the structure of the whole implementation. Well done!

src/lib/utils/tree_hash/info.txt
```diff
@@ -0,0 +1,12 @@
+<defines>
+TREE_HASH -> 20231006
+</defines>
+
+<module_info>
+name -> "Tree Hash"
+</module_info>
```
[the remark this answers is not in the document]
FAlbertDev: Sure, I didn't know that this config is possible.

src/lib/utils/stl_util.h
```diff
@@ -198,6 +210,11 @@ class BufferStuffer {
          std::copy(buffer.begin(), buffer.end(), sink.begin());
       }
 
+      template <typename T>
+      auto append_be(T value) {
+         return store_be(value, next(sizeof(T)).data());
+      }
+
```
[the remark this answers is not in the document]
FAlbertDev: Yeah, I'll add `append_le` and remove the return value (I can't remember why we defined a return value at all...). Regarding the `std::unsigned_integral` concept, we also want to support strong types. Therefore, I'll add the additional overload (also for le):
```cpp
      template <concepts::strong_type T>
         requires std::unsigned_integral<typename T::wrapped_type>
      void append_be(T value) {
         append_be(value.get());
      }
  ```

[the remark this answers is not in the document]
Comment (FAlbertDev) [reply src/lib/utils/stl_util.h:243]: Well, for now I simply end up with a new concept allowing all three cases:
```cpp
template <typename T>
concept unsigned_type = std::unsigned_integral<T> ||  // Normal uint
                        (concepts::strong_type<T> &&
                         std::unsigned_integral<typename T::wrapped_type>) ||  // Strong type on uint
                        (std::is_enum_v<T> && std::unsigned_integral<std::underlying_type_t<T>>);  // Enum on uint
```

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+std::function<void(
+   StrongSpan<LMS_Tree_Node>, const TreeAddress&, StrongSpan<const LMS_Tree_Node>, StrongSpan<const LMS_Tree_Node>)>
+get_hash_pair_func_for_identifier(const LMS_Params& lms_params, LMS_Identifier identifier) {
+   // hash object must be shared, otherwise std::function would not be copyable, which is not allowed
+   std::shared_ptr<HashFunction> hash = HashFunction::create_or_throw(lms_params.hash_name());
+   return [hash, I = std::move(identifier)](StrongSpan<LMS_Tree_Node> out,
+                                            const TreeAddress& address,
+                                            StrongSpan<const LMS_Tree_Node> left,
+                                            StrongSpan<const LMS_Tree_Node> right) {
```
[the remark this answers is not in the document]
FAlbertDev: Nice!

src/lib/utils/stl_util.h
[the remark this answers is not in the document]
reneme: Probably that's what we want. Naming this is hard, though.
* `unsigned_integralish`? 🤡
* `potentially_wrapped_unsigned_integral`?

Another alternative that the STL seems to do in such cases: Have a trait that new "unsigned integral-ish" types can specialize to explicitly enable the concept for them. But I think that's taking it a bit too far.

src/lib/pubkey/hss_lms/lms.cpp
FAlbertDev: If you insist 😝

src/lib/pubkey/hss_lms/lms.cpp
FAlbertDev: I'll also use switches for lm_ots, where we did the same 👍

src/lib/pubkey/hss_lms/lms.cpp
FAlbertDev: This is indeed a copy since we decided to copy the auth path into the signature object instead of storing only spans. Better would be to directly call `slicer.copy` instead of  `slicer.take`, though.

src/lib/pubkey/hss_lms/lms.cpp
FAlbertDev: Agreed.

src/lib/pubkey/hss_lms/hss_lms_utils.h
FAlbertDev: I could not measure a significant performance boost. However, I like getting rid of the member spans.

FAlbertDev:
Thanks again for your review @reneme! I addressed your suggestions in https://github.com/randombit/botan/pull/3716/commits/a4c5df28bc9126a3f4de5fc3b02ec523863184a4.

src/lib/utils/stl_util.h
reneme: Sooner or later, we'll need to go through the code base and clean up these new range-based overloads once they are merged. No need to address here, I guess.

src/lib/utils/tree_hash.h
reneme: Don't have a super strong opinion on that. You have a point that it is more readable for sure.

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,432 @@
...
+         case LMS_Algorithm_Type::SHAKE_M24_H15:
+            return {"SHAKE-256(192)", 15};
+         case LMS_Algorithm_Type::SHAKE_M24_H20:
+            return {"SHAKE-256(192)", 20};
+         case LMS_Algorithm_Type::SHAKE_M24_H25:
+            return {"SHAKE-256(192)", 25};
+         case LMS_Algorithm_Type::RESERVED:
+            throw Decoding_Error("Unsupported LMS algorithm type");
+      }
+      throw Decoding_Error("Unsupported LMS algorithm type");
```
reneme, line 161: Please use unique error messages: If you see one of those in some bug report log, you'll want to know which code path was taken.

reneme: ... same below, for the hash functions.

src/lib/pubkey/hss_lms/lms.cpp
reneme: Thank you.

reneme (commented):
Just some drive-by comment.

Personally, I'm perfectly okay with you self-resolving my comments if they aren't controversial. As a rule of thumb: small to medium annotation with an obvious resolution in the code: simply resolve yourself. Especially, if you just apply my code suggestion. :+1:

lieser:
rebased to current master

reneme:
I force pushed the changes here, to:

1. take advantage of the changes in #3869 
2. squash and clean up the history

reneme (commented):
Another review pass, mostly looking at programming and style concepts. Overall, this looks well structured and concise. 👍

src/lib/pubkey/hss_lms/hss_lms_utils.h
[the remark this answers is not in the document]
reneme: Mhh, not sure whether we're now turning in circles. Though, here's a suggestion related to this: https://github.com/randombit/botan/pull/3716#discussion_r1452185525

src/lib/pubkey/hss_lms/hss.h
```diff
@@ -0,0 +1,402 @@
...
+
+      /**
+       * @brief Returns the LMS an LM-OTS parameters at the specified @p level of the HSS tree.
+       */
+      const LMS_LMOTS_Params_Pair& params_at_level(HSS_Level level) const { return m_lms_lmots_params.at(level.get()); }
+
+      /**
+       * @brief Returns the number of layers the HSS tree has.
+       */
+      HSS_Level L() const { return HSS_Level(checked_cast_to<uint32_t>(m_lms_lmots_params.size())); }
```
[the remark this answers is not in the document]
reneme: Works now.

src/lib/pubkey/hss_lms/hss.h
```diff
@@ -0,0 +1,402 @@
...
+
+      /**
+       * @brief Returns the seed contained in the private key used for key derivation
+       */
+      const LMS_Seed& seed() const { return m_hss_seed; }
+
+      /**
+       * @brief The identifier of the top level LMS tree
+       */
+      const LMS_Identifier& identifier() const { return m_identifier; }
```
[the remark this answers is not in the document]
FAlbertDev: Probably not :+1:

src/lib/pubkey/hss_lms/hss.h
```diff
@@ -0,0 +1,402 @@
...
+   private:
+      /**
+       * @brief Private constructor using the individual signature fields.
+       */
+      HSS_Signature(uint8_t Nspk, std::vector<Signed_Pub_Key> signed_pub_keys, LMS_Signature sig) :
+            m_Nspk(Nspk), m_signed_pub_keys(std::move(signed_pub_keys)), m_sig(std::move(sig)) {}
+
+      uint8_t m_Nspk;
+      std::vector<Signed_Pub_Key> m_signed_pub_keys;
+      LMS_Signature m_sig;
```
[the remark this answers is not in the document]
FAlbertDev: Nspk is one field of the HSS signature that defines how many signed_pub_keys are expected. However, after parsing the signature in the constructor, this field becomes indeed redundant with the size of signed_pub_keys. I suggest removing the member but keeping the accessor `Nspk()` as an alias to signed_pub_keys.size() for better spec accordance. Good catch :+1:

src/lib/pubkey/hss_lms/hss.h
```diff
@@ -0,0 +1,402 @@
...
+       */
+      uint8_t Nspk() const { return m_Nspk; }
+
+      /**
+       * @brief Returns the signed LMS key signed by a specific layer.
+       *
+       * @param layer The layer by which the LMS key is signed.
+       * @return The LMS key and the signature by its parent layer.
+       */
+      const Signed_Pub_Key& signed_pub_key(uint8_t layer) const { return m_signed_pub_keys.at(layer); }
```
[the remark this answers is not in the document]
FAlbertDev: This smells like the already existing strong type HSS_Level. I don't know why we didn't use this one.

src/lib/pubkey/hss_lms/hss_lms_utils.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Botan {
+PseudorandomKeyGeneration::PseudorandomKeyGeneration(std::span<const uint8_t> identifier) {
+   m_input_buffer.resize(identifier.size() + sizeof(uint32_t) + sizeof(uint16_t) + sizeof(uint8_t));
+   BufferStuffer input_stuffer(m_input_buffer);
+   input_stuffer.append(identifier);
+}
+
+void PseudorandomKeyGeneration::gen(std::span<uint8_t> out, HashFunction& hash, std::span<const uint8_t> seed) const {
+   hash.update(m_input_buffer);
```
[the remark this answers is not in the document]
FAlbertDev: This code section is a bottleneck. I haven't measured how expensive the `clear` is in comparison to the hash itself, but I want to avoid any unnecessary operation here.

src/lib/pubkey/hss_lms/hss_lms_utils.cpp
```diff
@@ -0,0 +1,26 @@
...
+   hash.update(seed);
+   hash.final(out);
```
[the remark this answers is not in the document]
FAlbertDev: As mentioned above, this code path is very hot, so we need it to be as fast as possible. The reason for splitting these fields is also performance. When computing the hash chains, we compute the hash many times with inputs with almost the same prefix. Therefore, we store the prefix and only change the part of the prefix that needs to be changed. By passing this buffer to the hash function all at once, we hope to gain some advantage over passing it in multiple smaller chunks.

src/lib/pubkey/hss_lms/hss_lms_utils.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+#include <botan/internal/hss_lms_utils.h>
+
+#include <botan/internal/stl_util.h>
+
+namespace Botan {
+PseudorandomKeyGeneration::PseudorandomKeyGeneration(std::span<const uint8_t> identifier) {
+   m_input_buffer.resize(identifier.size() + sizeof(uint32_t) + sizeof(uint16_t) + sizeof(uint8_t));
+   BufferStuffer input_stuffer(m_input_buffer);
+   input_stuffer.append(identifier);
```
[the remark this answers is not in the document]
FAlbertDev: I like this idea 👍

src/lib/pubkey/hss_lms/lm_ots.h
```diff
@@ -0,0 +1,341 @@
...
+
+      /**
+       * @brief The n-byte randomizer of the signature.
+       */
+      std::span<const uint8_t> C() const { return m_C; }
+
+      /**
+       * @brief Returns the part of the signature for @p chain_idx.
+       */
+      StrongSpan<const LMOTS_Node> y(uint16_t chain_idx) const { return m_y.at(chain_idx); }
```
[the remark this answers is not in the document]
FAlbertDev: Sadly, I have no access to `params.n()` at this point to recompute the span. I could create a params object from m_algorithm_type, store n as a member, or add a function parameter for n. But none of these possibilities are very nice. I will keep it as it is since it's not expensive, and the buffer is never changed, so there is no danger.

src/lib/pubkey/hss_lms/lm_ots.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+         if(start == end) {
+            copy_into(out, in);
+            return;
+         }
+         m_gen.set_i(chain_idx);
+
+         // Unroll first iteration of the loop
+         m_gen.set_j(start++);
+         m_gen.gen(out, hash, in);
```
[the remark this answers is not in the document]
FAlbertDev: Yeah, since this is only one extra copy per hash chain, this is ok, I guess. It improves readability a lot 👍 
I can also remove `if(start == end) { return; }` since it is covered by the loop.

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,432 @@
...
+                  HashFunction& hash) {
+   hash.update(lmots_pk.identifier());
+   hash.update_be(tree_address.r());
+   hash.update_be(D_LEAF);
+   hash.update(lmots_pk.K());
+   hash.final(out);
+}
+
+auto lms_gen_leaf_func(const LMS_PrivateKey& lms_sk) {
+   return [hash = HashFunction::create_or_throw(lms_sk.lms_params().hash_name()), lms_sk](
```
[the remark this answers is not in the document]
FAlbertDev: This is expected, as far as I remember. `lms_sk` is not very big, and from a memory safety perspective, it's more elegant when returning the lambda object that it holds no reference. In the places where we use `lms_gen_leaf_func`, the lifetime of lms_sk would be long enough to work with a reference, but we decided to go the idiomatic way.

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,432 @@
...
+   // Alg. 6. 2.d.
+   if(total_remaining_bytes < size(lms_params)) {
+      throw Decoding_Error("Too few bytes while parsing LMS public key.");
+   }
+   // Alg. 6. 2.b.
+   auto lmots_type = load_be<LMOTS_Algorithm_Type>(slicer.take<sizeof(LMOTS_Algorithm_Type)>());
+   auto lmots_params = LMOTS_Params::create_or_throw(lmots_type);
+
+   if(lms_params.hash_name() != lmots_params.hash_name()) {
+      throw Decoding_Error("No support for HSS-LMS instances with multiple hash functions.");
```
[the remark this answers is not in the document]
FAlbertDev: I wanted to stay consistent so that every parsing error of the public key creates a `Decoding_Error`. In my interpretation, `Not_Implemented` is used when a feature is missing but could be added in the future. However, at this point, we decided strictly against the support of multiple hash functions per tree (which is also not recommended in the respective documents). 
But I'm open for discussion.

src/lib/utils/tree_hash/tree_hash.h
```diff
@@ -0,0 +1,269 @@
...
+ *                 creates a one-time/few-time-signature's public key which is hashed to be the leaf node.
+ * @param tree_address The address that is passed to gen_leaf or node_pair hash. This function will update the
+ *                     address accordings to the currently processed node. This object may contain further
+ *                     algorithm specific information, like the position of this merkle tree in a hypertree.
+ */
+template <concepts::contiguous_strong_type TreeNode,
+          concepts::strong_span AuthPathSS,
+          concepts::tree_node_index TreeNodeIndex,
+          concepts::tree_layer_index TreeLayerIndex,
+          typename Address>
```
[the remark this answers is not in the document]
FAlbertDev: Well, we call `.get()` on them in the function body, which is only available for strong types. Can we abstract this? What do you think?

src/lib/utils/tree_hash/tree_hash.h
```diff
@@ -0,0 +1,269 @@
...
+            copy_into(out_root, current_node);
+            return;
+         }
+
+         // Check if the node we have is a part of the authentication path; if
+         // it is, write it out. The XOR sum of both nodes (at internal_idx and internal_leaf)
+         // is 1 iff they have the same parent node in the FORS tree
+         if(is_signing && (internal_idx ^ internal_leaf.value()) == 0x01U) {
+            auto auth_path_location = out_auth_path.value().get().subspan(h.get() * node_size, node_size);
+            copy_into(auth_path_location, current_node);
```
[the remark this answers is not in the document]
FAlbertDev: The algorithm does not visit the nodes in the same order in which they are written into the `out_auth_path` buffer. We could precompute these locations with a buffer stuffer but I don't think this is optimal too. The best would be some sort of mdspan for this.

src/lib/utils/tree_hash/tree_hash.h
```diff
@@ -0,0 +1,269 @@
...
+   BOTAN_ASSERT(out_auth_path.has_value() == leaf_idx.has_value(),
+                "Both leaf index and auth path buffer is given or neither.");
+   bool is_signing = leaf_idx.has_value();
+   if(is_signing) {
+      BOTAN_ASSERT_NOMSG(out_auth_path.value().size() == node_size * total_tree_height.get());
+   }
+
+   const TreeNodeIndex max_idx(uint32_t((1 << total_tree_height.get()) - 1));
+
+   std::vector<uint8_t> stack(total_tree_height.get() * node_size);
```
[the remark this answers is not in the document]
FAlbertDev: Well, it seems the "stack" is not a stack. It is a storage for the last visited node at a given tree height. We should rename this variable for sure. And you are right; this span computation is inelegant. I suggest using a two-dimensional vector instead. IMO, creating a separate class is overkill.

FAlbertDev:
Thanks for your detailed and comprehensive review, @reneme 👍 I addressed your suggestions in 
1f3befe.

src/lib/pubkey/hss_lms/hss_lms_utils.cpp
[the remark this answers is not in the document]
reneme: For the record: we went with this alternative suggestion: https://github.com/randombit/botan/pull/3716#discussion_r1452185525.

src/lib/pubkey/hss_lms/lms.cpp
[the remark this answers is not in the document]
reneme: Yeah, sounds good to me. It's a proper nit-pick anyway.

src/lib/utils/tree_hash/tree_hash.h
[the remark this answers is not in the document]
reneme: Certainly, this could be abstracted. Though, let's leave this for when we try to use this in XMSS or SPX+.

src/lib/utils/tree_hash/tree_hash.h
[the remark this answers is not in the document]
reneme: Much better now, thanks!

src/lib/utils/tree_hash/tree_hash.h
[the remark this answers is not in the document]
reneme: Okay, no need right now.

src/lib/pubkey/hss_lms/hss_lms_utils.cpp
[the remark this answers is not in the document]
reneme: I took the liberty to iterate on your implementation: https://github.com/randombit/botan/pull/3716/commits/6b6f6fdf717ef58a6b5e8c39bef3301a3827898a. Without overkill-template magic we can't get rid of some of the magic numbers, so let's embrace it. :) But we can get rid of the raw pointers!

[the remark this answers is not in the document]
FAlbertDev: Nice, I like it 👍

src/lib/pubkey/hss_lms/hss.h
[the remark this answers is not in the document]
lieser: > This smells like the already existing strong type HSS_Level. I don't know why we didn't use this one.

Not using the same strong type here was a deliberate choice by me, and I still would prefer to not use the same strong types for `Nspk` and `L`.


The definitions of `L` and `Nspk` according to the RFC:
> The number of levels is denoted as L and is between one and eight, inclusive.

> Nspk = L-1 denotes the number of signed public keys.

Levels are indexing the different layers of the HSS tree,
the range is `[0,L)`, 0 being the root.
Note that the RFC nether uses the term layer, unsure why we started to use it.

For me the subtle differences were very confusing at the beginning, and so I wanted to especially use different types for `L` and `Nspk` to make it very clear that they can not just be interchanged with each other. Although they are of course very closely related to each other.

Now do we really want to have three different strong types for this, one each for level, `Nspk` and `L`?
I'm unsure but probably that is a little overkill.
But again we should at least avoid using the same strong type for both `Nspk` and `L`, as otherwise the point of using strong types to avoid misuse is kind of defeated.

reneme:
Rebased to (latest) `master` after #3888.

reneme:
Rebased after the underlying #3869 needed to get conflicts resolved after #3908 was merged. Also, the code here needed minor conflict resolving due to #3908.

randombit:
#3869 is approved; I'll review LMS after this is rebased post merge

reneme:
> I'll review LMS after this is rebased post merge

Done. I merged #3869 and rebased. No history rewrite was done.

src/lib/pubkey/hss_lms/hss.cpp
```diff
@@ -0,0 +1,417 @@
...
+         checked_cast_to_or_throw<uint8_t, Invalid_Argument>(scan_layer.arg_as_integer(0), "Invalid parameter");
+      const auto w =
+         checked_cast_to_or_throw<uint8_t, Invalid_Argument>(scan_layer.arg_as_integer(1), "Invalid parameter");
+      m_lms_lmots_params.push_back({LMS_Params::create_or_throw(hash, h), LMOTS_Params::create_or_throw(hash, w)});
+   }
+   m_max_sig_count = calc_max_sig_count();
+}
+
+HSS_Sig_Idx HSS_LMS_Params::calc_max_sig_count() const {
+   uint32_t total_hight_counter = 0;
```
randombit, line 104: Typo: `height`

src/lib/pubkey/hss_lms/hss_lms.cpp
```diff
@@ -0,0 +1,196 @@
...
+      void update(const uint8_t msg[], size_t msg_len) override {
+         m_msg_buffer.insert(m_msg_buffer.end(), msg, msg + msg_len);
+      }
+
+      secure_vector<uint8_t> sign(RandomNumberGenerator&) override {
+         std::vector<uint8_t> message_to_sign = std::exchange(m_msg_buffer, {});
+         return m_private->sign(message_to_sign);
+      }
+
+      size_t signature_length() const override { return HSS_Signature::size(m_private->hss_params()); }
```
randombit, line 172: This length could be computed and cached once, either here or in the private key itself.

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,432 @@
...
+   // Alg. 6. 1. (4 bytes are sufficient until the next check)
+   if(total_remaining_bytes < sizeof(LMS_Algorithm_Type)) {
+      throw Decoding_Error("Too few bytes while parsing LMS public key.");
+   }
+   // Alg. 6. 2.a.
+   auto lms_type = load_be<LMS_Algorithm_Type>(slicer.take<sizeof(LMS_Algorithm_Type)>());
+   // Alg. 6. 2.c.
+   auto lms_params = LMS_Params::create_or_throw(lms_type);
+   // Alg. 6. 2.d.
+   if(total_remaining_bytes < size(lms_params)) {
```
randombit, line 281: This would be easier to follow as `LMS_PublicKey::size(lms_params)`

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,432 @@
...
+            throw Decoding_Error("Unsupported LMS algorithm type");
+      }
+      throw Decoding_Error("Unsupported LMS algorithm type");
+   }(type);
+
+   return LMS_Params(type, hash_name, height);
+}
+
+LMS_Params LMS_Params::create_or_throw(std::string_view hash_name, uint8_t height) {
+   if(height != 5 && height != 10 && height != 15 && height != 20 && height != 25) {
```
randombit, line 168: This seems redundant with the later checks

randombit (approved):
Various small comments/questions but overall lgtm

src/lib/pubkey/hss_lms/lms.h
```diff
@@ -0,0 +1,342 @@
...
+class BOTAN_TEST_API LMS_Signature {
+   public:
+      /**
+       * @brief Parse the bytes of a lms signature into a LMS Signature object
+       *
+       * @param slicer A BufferSlicer object at the position of the LMS_Signature to parse
+       * @return LMS_Signature object
+       * @throws Decoding_Error If parsing the signature fails.
+       */
+      static LMS_Signature from_bytes_or_throw(BufferSlicer& slicer);
```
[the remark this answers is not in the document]
FAlbertDev: We chose this interface for LMS signatures because it is challenging to precompute the length of the signature sections. We need to parse the first fields of the signature to obtain the used LMS instance. Then, the first fields of the OTS signature define the OTS instance. Only then can we reconstruct the length. 

Fishing those values from the entire signature is messy.  Alternatively, we could pass the entire span as input and return a subspan of the remaining bytes. That's exactly the information the BufferSlicer contains.

src/lib/pubkey/hss_lms/lms.h
```diff
@@ -0,0 +1,342 @@
...
+class BOTAN_TEST_API LMS_PublicKey : public LMS_Instance {
+   public:
+      /**
+       * @brief Parse a public LMS key.
+       *
+       * @param slicer The BufferSlicer at the public key bytes' position
+       * @return The LMS public key.
+       * @throws Decoding_Error If parsing the public key fails.
+       */
+      static LMS_PublicKey from_bytes_or_throw(BufferSlicer& slicer);
```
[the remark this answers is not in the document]
FAlbertDev: I agree that we can pass a span here instead. This would be suitable but inconsistent with signature parsing, where passing the slicer is quite sensible (see below). I have no hard feelings for both options.

src/lib/pubkey/hss_lms/lm_ots.cpp
```diff
@@ -0,0 +1,350 @@
...
+   const auto Q_with_cksm = gen_Q_with_cksm(params, identifier, q, sig.C(), msg);
+
+   // Prefill the final hash object
+   const auto pk_hash = HashFunction::create_or_throw(params.hash_name());
+   pk_hash->update(identifier);
+   pk_hash->update(store_be(q));
+   pk_hash->update(store_be(D_PBLC));
+
+   Chain_Generator chain_gen(identifier, q);
+   const auto hash = HashFunction::create_or_throw(params.hash_name());
```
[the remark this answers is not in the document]
FAlbertDev: I like your suggestions! It also shows more explicitly that both hash functions are equal.

reneme:
Rebased to fix conflicts on master (likely introduced by merging #3933).

src/lib/pubkey/hss_lms/lms.h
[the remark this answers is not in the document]
reneme: Perhaps my two cents; talking from experience in #4024: I totally agree, that any loosely coupled API shouldn't pass around `BufferSlicer` or `BufferSlicer` references. Though, for tightly coupled internal APIs (e.g. utility functions in the same compilation unit), I find this very much convenient.

Accepting a span, as @randombit suggested, has the drawback that _the caller must know how many bytes_ the callee will want to consume or emit. Usually, that can be determined, but it typically requires some extra calculations that were already done to determine the overall size of the final serialized structure. Just passing the `Buffer***` reference and trusting that the callee will be responsible, is saving some cycles and also noise in the code.

That said: We should always remember to validate that the `Buffer***` is full/empty when expected at the very end of the marshalling, to check that everything checked out as expected.

src/lib/utils/int_utils.h
```diff
@@ -47,14 +48,71 @@ constexpr inline std::optional<T> checked_mul(T a, T b) {
...
 }
 
-template <std::unsigned_integral ReturnT>
-constexpr inline std::optional<ReturnT> checked_cast(std::unsigned_integral auto input) {
-   if(std::numeric_limits<ReturnT>::max() < input) {
-      return {};
+namespace detail {
+
+template <typename T>
+concept int_or_strong_type = std::integral<T> || concepts::integral_strong_type<T>;
```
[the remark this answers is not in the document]
reneme: Future work: https://github.com/randombit/botan/issues/4029

doc/api_ref/pubkey.rst
```diff
@@ -1241,3 +1249,42 @@ signature:
...
+used hash function and the Winternitz Parameter of the underlying one-time signature
+can be chosen for each tree layer. For a sensible selection of parameters refer to
+`RFC 8554 Section 6.4. <https://datatracker.ietf.org/doc/html/rfc8554#section-6.4>`_.
+
+.. warning::
+
+   HSS-LMS is stateful, meaning the private key must be updated after
+   each signature. If the same private key is ever used to generate
+   two different signatures, then the scheme becomes insecure. For
+   this reason it can be challening to use HSS-LMS securely.
```
reneme, line 1271: Perhaps, give clear guidance what "updating the private key" means exactly. Along the lines: "After each signature, the output 
 of `sk->private_key_bits()` changes and MUST be used to overwrite any existing persisted private key"

reneme: (we should probably add a similar line to the XMSS documentation if it doesn't exist, yet)

src/lib/pubkey/hss_lms/hss.h
```diff
@@ -0,0 +1,392 @@
...
+#include <string_view>
+#include <utility>
+#include <vector>
+
+namespace Botan {
+
+/**
+ * @brief The index of a node within a specific LMS tree layer
+ */
+using HSS_Sig_Idx = Strong<uint64_t, struct HSS_Sig_Idx_, EnableArithmeticWithPlainNumber>;
```
reneme, line 28: Without a strong opinion: In many other algorithms we centralized the existing strong types into a `algo_types.h`. I did find this quite useful because it provides a first overview of the artefacts involved.

src/lib/utils/concepts.h
```diff
@@ -198,9 +198,16 @@ concept strong_type = is_strong_type_v<T>;
 template <class T>
 concept contiguous_strong_type = strong_type<T> && contiguous_container<T>;
 
+template <class T>
+concept integral_strong_type = strong_type<T> && std::integral<typename T::wrapped_type>;
```
reneme, line 202: For reference: #4029.

src/lib/pubkey/hss_lms/hss.cpp
```diff
@@ -0,0 +1,418 @@
...
+   std::vector<uint8_t> current_pk;
+   for(int32_t layer_it = hss_params().L().get() - 1; layer_it >= 0; --layer_it) {
+      HSS_Level layer(layer_it);
+      if(layer == hss_params().L() - 1) {
+         current_pk =
+            lms_key_at_layer.at(layer.get())
+               .sign_and_get_pk(out_lms_sig_buffer_at_layer.at(layer.get()), q.at(layer.get()), LMS_Message(msg))
+               .to_bytes();
+      } else {
+         std::copy(current_pk.begin(), current_pk.end(), out_child_pk_buffer_at_layer.at(layer.get()).begin());
```
reneme, line 257:
```suggestion
         copy_mem(out_child_pk_buffer_at_layer.at(layer.get()), current_pk);
```

... didn't try this locally, but should work and runtime-check the range sizes.

src/lib/pubkey/hss_lms/hss.cpp
```diff
@@ -0,0 +1,418 @@
...
+std::vector<uint8_t> HSS_LMS_PublicKeyInternal::to_bytes() const {
+   std::vector<uint8_t> lms_pub_key_bytes = m_top_lms_pub_key.to_bytes();
+   std::vector<uint8_t> bytes(sizeof(m_L) + lms_pub_key_bytes.size());
+   BufferStuffer stuffer(bytes);
+
+   stuffer.append(store_be(m_L));
+   stuffer.append(lms_pub_key_bytes);
+   BOTAN_ASSERT_NOMSG(stuffer.full());
+
+   return bytes;
```
reneme, line 335:
```suggestion
   return concat<std::vector<uint8_t>>(store_be(m_L), m_top_lms_pub_key.to_bytes());
```

... this should work now (using https://github.com/randombit/botan/pull/3994).

reneme (commented):
Just a few ad-hoc remarks. Will go through the implementation one final time tomorrow.

src/lib/pubkey/hss_lms/hss.cpp
```diff
@@ -0,0 +1,418 @@
...
+}
+
+std::shared_ptr<HSS_LMS_PrivateKeyInternal> HSS_LMS_PrivateKeyInternal::from_bytes_or_throw(
+   std::span<const uint8_t> key_bytes) {
+   if(key_bytes.size() < sizeof(HSS_Level) + sizeof(HSS_Sig_Idx)) {
+      throw Decoding_Error("Too few private key bytes.");
+   }
+   BufferSlicer slicer(key_bytes);
+
+   const auto L = load_be<HSS_Level>(slicer.take<sizeof(HSS_Level)>());
```
reneme, line 127: This, being so verbose, is bothering me quite a bit, tbh.

For future work (and related to https://github.com/randombit/botan/pull/3716#discussion_r1576194649): We might want to consider allowing to pass `BufferSlicer` into a `load_be/le()` in-lieu of a "contiguous-range". Conceptually, they're almost the same anyway. It would be quite handy, particularly when working with strongly-typed integers. Many locations in this file could be simplified along those lines:

```suggestion
   const auto L = load_be<HSS_Level>(slicer);
```

... I'm on the fence to propose the same for `store_be/le()`. There, we can do `stuffer.append(store_be(someval))` which I find acceptable. For the record, the new overload would allow `store_be(stuffer, someval)` which might even be less readable. Regardless, a stuffer-based overload would also allow things like multi-input stores, e.g. `store_le(stuffer, val1, val2, val3, val4)` or `store_be(stuffer, array_of_vals)`.

(I'll create a ticket for it, if there's no objection. No reason to hold back on merging because of that, IMO.)

src/lib/pubkey/hss_lms/hss.cpp
```diff
@@ -0,0 +1,418 @@
...
+      params.push_back({LMS_Params::create_or_throw(lms_type), LMOTS_Params::create_or_throw(lmots_type)});
+   }
+   std::string hash_name = params.at(0).lms_params().hash_name();
+   if(!std::all_of(params.begin(), params.end(), [&hash_name](HSS_LMS_Params::LMS_LMOTS_Params_Pair& lms_lmots_params) {
+         bool valid_lmots_hash = lms_lmots_params.lmots_params().hash_name() == hash_name;
+         bool valid_lms_hash = lms_lmots_params.lms_params().hash_name() == hash_name;
+         return valid_lmots_hash && valid_lms_hash;
+      })) {
+      throw Decoding_Error("Inconsistent hash functions are not allowed.");
+   }
```
reneme, line 150: Perhaps use `std::any_of` and reverse the predicate?

reneme (approved):
Still didn't find time for another full review pass. But no general objection to merging this from my side.

src/lib/pubkey/hss_lms/lms.cpp
```diff
@@ -0,0 +1,428 @@
...
+   std::vector<uint8_t> bytes(size(lms_params()));
+   BufferStuffer stuffer(bytes);
+
+   stuffer.append(store_be(lms_params().algorithm_type()));
+   stuffer.append(store_be(lmots_params().algorithm_type()));
+   stuffer.append(identifier());
+   stuffer.append(m_lms_root);
+   BOTAN_ASSERT_NOMSG(stuffer.full());
+
+   return bytes;
```
reneme, line 306: Technically, we could now formulate that with `concat()`:

```suggestion
   // clang-format off
   return concat<std::vector<uint8_t>>(
      store_be(lms_params().algorithm_type()),
      store_be(lmots_params().algorithm_type()),
      identifier(),
      m_lms_root);
   // clang-format on
```

... your call on what is easier to read.

reneme (approved):

src/lib/pubkey/hss_lms/hss.cpp
reneme: For the record: https://github.com/randombit/botan/issues/4068

src/lib/pubkey/hss_lms/hss.h
FAlbertDev: We consciously decided against it to keep the hierarchy of LMS' building blocks separate.

src/lib/pubkey/hss_lms/hss.cpp
FAlbertDev: fair

src/lib/pubkey/hss_lms/lms.cpp
FAlbertDev: I like your version better!

doc/api_ref/pubkey.rst
FAlbertDev: Done!

reneme (approved):

PR #3724, opened 2023-10-04
Some cleanups for Kuznyechik

randombit:
No performance change one way or another vs master.

@huven fyi

### DIFF
--- a/src/lib/block/kuznyechik/kuznyechik.cpp
+++ b/src/lib/block/kuznyechik/kuznyechik.cpp
@@ -4162,80 +4162,50 @@ const uint64_t C[32][2] = {{0xb87a486c7276a26e, 0x019484dd10bd275d}, {0xb3f490d8
                            {0x45aba4f6433784cc, 0x1dffec46132c75de}, {0x4e257c42d5ada17e, 0x1e80a3e223281c39},
                            {0xf65f342ea7db0310, 0x1f14273f33953b64}, {0x619b141e58d8a75e, 0x20a8ed9c45c16af1}};
 
-inline void LS(uint64_t x1, uint64_t x2, uint64_t& t1, uint64_t& t2) {
-   t1 = T[0][get_byte<7>(x1)][0] ^ T[1][get_byte<6>(x1)][0] ^ T[2][get_byte<5>(x1)][0] ^ T[3][get_byte<4>(x1)][0] ^
-        T[4][get_byte<3>(x1)][0] ^ T[5][get_byte<2>(x1)][0] ^ T[6][get_byte<1>(x1)][0] ^ T[7][get_byte<0>(x1)][0] ^
-        T[8][get_byte<7>(x2)][0] ^ T[9][get_byte<6>(x2)][0] ^ T[10][get_byte<5>(x2)][0] ^ T[11][get_byte<4>(x2)][0] ^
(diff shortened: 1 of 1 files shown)


src/lib/block/kuznyechik/kuznyechik.cpp
```diff
@@ -4162,80 +4162,60 @@ const uint64_t C[32][2] = {{0xb87a486c7276a26e, 0x019484dd10bd275d}, {0xb3f490d8
...
-        IT[15][S[get_byte<0>(x2)]][1];
+inline void ILS(uint64_t x1, uint64_t x2, uint64_t& t1, uint64_t& t2) {
+   t1 = 0;
+   t2 = 0;
+   for(size_t i = 0; i != 16; ++i) {
+      const uint8_t x = get_byte_var(7 - (i % 8), (i < 8) ? x1 : x2);
+      t1 ^= IT[i][x][0];
+      t2 ^= IT[i][x][1];
+   }
 }
```
reneme, line 4186: Indeed, this overload doesn't seem to be used anywhere.

src/lib/block/kuznyechik/kuznyechik.cpp
```diff
@@ -4257,43 +4237,59 @@ bool Kuznyechik::has_keying_material() const {
...
 
-   uint64_t k0, k1, k2, k3;
-   k0 = load_le<uint64_t>(key.data(), 0);
-   k1 = load_le<uint64_t>(key.data(), 1);
-   k2 = load_le<uint64_t>(key.data(), 2);
-   k3 = load_le<uint64_t>(key.data(), 3);
+   uint64_t k0 = load_le<uint64_t>(key.data(), 0);
+   uint64_t k1 = load_le<uint64_t>(key.data(), 1);
+   uint64_t k2 = load_le<uint64_t>(key.data(), 2);
+   uint64_t k3 = load_le<uint64_t>(key.data(), 3);
```
reneme, line 4247: FYI: With #3707 I so far didn't add overloads for `load_le(std::span<>, size_t offset);`. My assumption was that we'd either use `BufferSlicer` to do the offsetting explicitly, or formulate such stride-loads as:

```C++
uint64_t k0, k1, k2, k3;
load_le(key, k0, k1, k2, k3);
```

To avoid pre-defining variables and initializing them later, a (not yet implemented) option might be:

```C++
auto [k0, k1, k2, k3] = load_le<uint64_t>(key, 4 /* number of uint64_t to load */);
```

Might be worth discussing this design in #3707.

reneme (commented):
Somewhat off-topic comment about `load_le`

src/lib/block/kuznyechik/kuznyechik.cpp
randombit: Yeah, just leftover. I guess only clang warns about this at the moment. I suppose the compiler has to keep track of which TU contained the definition, since if the inline came from a header the warning would be noise.

src/lib/block/kuznyechik/kuznyechik.cpp
randombit: The destructuring approach would be nice. I generally don't like leaving unitialized vars but in the sort of code we'd be looking at (like your first snippet) they are immediately initialized in the next line so ... meh. It would be fairly easy to typo and leave something uninitialized (eg `load_le(key, k0, k1, k2, k2)`), unlike the approach that returns a tuple, but that situation should be caught by any reasonable tests.

PR #3729, opened 2023-10-05
Wrap TLS::Group_Params enum in a class

randombit:
This is as it stands a minor compatibility break

* It's not possible anymore to `static_cast<uint16_t>` a `Group_Params`. We could allow this by including an implicit conversion to an integer (but I'd rather not)
* It removes various free functions (`is_ecdh`, etc) and replaced by member functions on the new class `Group_Params`. We could retain those as deprecated inline functions which just forward to the member function.

I'm not sure if anyone in practice would make much use of either functionality. Potentially if they were reimplementing some parts of `TLS::Callbacks`. I could be argued into either or both I suppose.

### DIFF
--- a/src/lib/tls/tls_algos.h
+++ b/src/lib/tls/tls_algos.h
@@ -10,6 +10,7 @@
 #include <botan/asn1_obj.h>
 #include <botan/pk_keys.h>
 #include <botan/types.h>
+#include <optional>
 #include <string>
 #include <vector>
 
@@ -78,7 +79,7 @@ Auth_Method BOTAN_TEST_API auth_method_from_string(std::string_view str);
 /*
 * Matches with wire encoding
 */
-enum class Group_Params : uint16_t {
+enum class Group_Params_Code : uint16_t {
    NONE = 0,
 
    SECP256R1 = 23,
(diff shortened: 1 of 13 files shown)


src/lib/tls/tls_algos.h
```diff
@@ -117,48 +118,81 @@ enum class Group_Params : uint16_t {
...
+      constexpr bool is_post_quantum() const { return is_pure_kyber() || is_pqc_hybrid(); }
+
+      constexpr bool is_pqc_hybrid() const {
+         return m_code == Group_Params_Code::HYBRID_X25519_KYBER_512_R3_CLOUDFLARE ||
+                m_code == Group_Params_Code::HYBRID_X25519_KYBER_768_R3_CLOUDFLARE ||
+                m_code == Group_Params_Code::HYBRID_X25519_KYBER_512_R3_OQS ||
+                m_code == Group_Params_Code::HYBRID_X25519_KYBER_768_R3_OQS ||
+                m_code == Group_Params_Code::HYBRID_SECP256R1_KYBER_512_R3_OQS ||
+                m_code == Group_Params_Code::HYBRID_SECP384R1_KYBER_768_R3_OQS ||
+                m_code == Group_Params_Code::HYBRID_SECP521R1_KYBER_1024_R3_OQS;
```
reneme, line 179: Take care when merging. This list is adapted in #3733 and as it stands this may produce a rather unclear merge conflict.

src/lib/tls/tls_algos.h
```diff
@@ -117,48 +118,81 @@ enum class Group_Params : uint16_t {
...
+      constexpr bool is_kem() const { return is_pure_kyber() || is_pqc_hybrid(); }
+
+      // Returns std::nullopt if the param has no known name
+      std::optional<std::string> to_string() const;
+
+      Group_Params(const Group_Params& other) = default;
+      Group_Params(Group_Params&& other) = default;
+      Group_Params& operator=(const Group_Params&) = default;
+      Group_Params& operator=(Group_Params&&) = default;
+      ~Group_Params() = default;
```
reneme, line 191: If all are default, none needs to be defined. No hard feelings either way.

src/lib/tls/tls_algos.h
```diff
@@ -117,48 +118,81 @@ enum class Group_Params : uint16_t {
...
-   return group == Group_Params::HYBRID_X25519_KYBER_512_R3_CLOUDFLARE ||
-          group == Group_Params::HYBRID_X25519_KYBER_768_R3_CLOUDFLARE ||
-          group == Group_Params::HYBRID_X25519_KYBER_512_R3_OQS ||
-          group == Group_Params::HYBRID_X25519_KYBER_768_R3_OQS ||
-          group == Group_Params::HYBRID_SECP256R1_KYBER_512_R3_OQS ||
-          group == Group_Params::HYBRID_SECP384R1_KYBER_768_R3_OQS ||
-          group == Group_Params::HYBRID_SECP521R1_KYBER_1024_R3_OQS;
-}
+      /**
+      * Returns std::nullopt if an unknown name
```
reneme, line 132:
```suggestion
      * @returns std::nullopt if an unknown name
```

src/lib/tls/tls_algos.h
```diff
@@ -117,48 +118,81 @@ enum class Group_Params : uint16_t {
...
-   return is_pure_kyber(group) || is_hybrid(group);
-}
+      constexpr bool operator==(Group_Params_Code code) const { return m_code == code; }
 
-constexpr bool is_post_quantum(const Group_Params group) {
-   return is_pure_kyber(group) || is_hybrid(group);
-}
+      constexpr bool operator==(Group_Params other) const { return m_code == other.m_code; }
+
+      constexpr bool operator<(Group_Params other) const { return m_code < other.m_code; }
```
reneme, line 140: C++20 doesn't need to get this spelled out anymore. Simply default the spaceship operator:

```C++
constexpr auto operator<=>(const Group_Params&) const = default;
```

Godbolt: https://godbolt.org/z/4jovGh4EY

src/lib/tls/tls_algos.h
```diff
@@ -117,48 +118,81 @@ enum class Group_Params : uint16_t {
...
+      constexpr bool is_dh_named_group() const {
+         return m_code == Group_Params_Code::FFDHE_2048 || m_code == Group_Params_Code::FFDHE_3072 ||
+                m_code == Group_Params_Code::FFDHE_4096 || m_code == Group_Params_Code::FFDHE_6144 ||
+                m_code == Group_Params_Code::FFDHE_8192;
+      }
+
+      constexpr bool is_diffie_hellman() const {
+         const uint16_t group_id = wire_code();
+         return (group_id >= 256 && group_id < 512);
+      }
```
reneme, line 163: 😮 I'm sure that is some legacy definition in the allocation of this TLS identifier value space. Maybe add a comment at least?

src/lib/tls/tls_extensions.cpp
```diff
@@ -365,7 +365,8 @@ const std::vector<Group_Params>& Supported_Groups::groups() const {
 std::vector<Group_Params> Supported_Groups::ec_groups() const {
    std::vector<Group_Params> ec;
    for(auto g : m_groups) {
-      if(group_param_is_dh(g) == false) {
+      // This conditional assumes that we do not support a PQC hybrid with DH
+      if(g.is_diffie_hellman() == false && g.is_pure_kyber() == false) {
          ec.push_back(g);
       }
```
reneme, line 371: Frankly, I'm not exactly sure what this does. I'm assuming "any group that uses ECC, either hybrid or not".
Should we promote this into a method on the new `Group_Params` class? E.g. We're planning to allow FrodoKEM for hybrid key exchange soon. I'm sure I'll forget to add `.is_pure_frodo()` here.

src/lib/tls/tls_algos.h
```diff
@@ -117,48 +118,81 @@ enum class Group_Params : uint16_t {
...
+      constexpr bool is_ecdh_named_curve() const {
+         return m_code == Group_Params_Code::SECP256R1 || m_code == Group_Params_Code::SECP384R1 ||
+                m_code == Group_Params_Code::SECP521R1 || m_code == Group_Params_Code::BRAINPOOL256R1 ||
+                m_code == Group_Params_Code::BRAINPOOL384R1 || m_code == Group_Params_Code::BRAINPOOL512R1;
+      }
+
+      constexpr bool is_dh_named_group() const {
+         return m_code == Group_Params_Code::FFDHE_2048 || m_code == Group_Params_Code::FFDHE_3072 ||
+                m_code == Group_Params_Code::FFDHE_4096 || m_code == Group_Params_Code::FFDHE_6144 ||
+                m_code == Group_Params_Code::FFDHE_8192;
```
reneme, line 157: I know this is somewhat redundant, but for clarity?

```suggestion
         if(!is_diffie_hellman()) {
            return false;
         }
         return m_code == Group_Params_Code::FFDHE_2048 || m_code == Group_Params_Code::FFDHE_3072 ||
                m_code == Group_Params_Code::FFDHE_4096 || m_code == Group_Params_Code::FFDHE_6144 ||
                m_code == Group_Params_Code::FFDHE_8192;
```

src/lib/tls/tls_policy.cpp
```diff
@@ -568,9 +568,13 @@ void print_vec(std::ostream& o, const char* key, const std::vector<std::string>&
...
-      o << group_param_to_string(v[i]);
-      if(i != v.size() - 1) {
-         o << ' ';
+      auto name = v[i].to_string();
+
+      if(name.has_value()) {
+         o << name.value();
+         if(i != v.size() - 1) {
+            o << ' ';
+         }
```
reneme, line 577: There's way too little pattern matching in C++. 😞

```suggestion
      if(auto name = v[i].to_string()) {
         o << name.value();
         if(i != v.size() - 1) {
            o << ' ';
         }
      }
```

reneme (changes requested):
Mostly minor things. However, I do think we should introduce something along the lines of `Group_Params::uses_ecc()`. If all relevant predicates are centralized its just much more likely that future-us will sanity check the predicates when new algorithms and combinations of hybrid algorithms are added.

src/lib/tls/tls_algos.h
randombit: :+1:

src/lib/tls/tls_algos.h
randombit: This is from https://www.rfc-editor.org/rfc/rfc7919 "Codepoints in the "Supported Groups Registry" with a high byte of 0x01 (that is, between 256 and 511, inclusive) are set aside for FFDHE groups" - that said it's unlikely we'll ever support other named DH groups and I don't think we'd ever actually call this function unless we specifically share the group with the peer.

Possibly something about this made more sense in TLS 1.0/1.1? I don't recall.

src/lib/tls/tls_extensions.cpp
randombit: I had made this change on the basis of "any group that uses ECC, hybrid or not" but on further checking, we only use this for TLS 1.2 negotiation where we should not attempt to negotiate a hybrid. So probably this should only return the list of standard ECDH curves, if any.

reneme:
Force-pushed to fix a superfluous curly brace introduced in one of the suggested patches.

src/lib/tls/tls_algos.h
randombit: Unfortunately that doesn't work here as it seems the additional overload of operator== (to handle `Group_Params_Code`) prevents the compiler from synthesizing the operator== we want using <=>

src/lib/tls/tls_algos.h
randombit: I checked over all uses, everywhere we did use it we were implicitly assuming it was a named FFDHE group. So this predicate is removed entirely.

src/lib/tls/tls_algos.h
randombit: :+1: removed

src/lib/tls/tls_extensions.cpp
randombit: Changed this to just filter using `is_ecdh_named_curve`

src/lib/tls/tls_algos.h
randombit: No longer required since `is_diffie_hellman` is gone

src/lib/tls/tls_policy.cpp
```diff
@@ -568,9 +568,11 @@ void print_vec(std::ostream& o, const char* key, const std::vector<std::string>&
...
    o << key << " = ";
    for(size_t i = 0; i != v.size(); ++i) {
-      o << group_param_to_string(v[i]);
-      if(i != v.size() - 1) {
-         o << ' ';
+      if(auto name = v[i].to_string()) {
+         o << name.value();
+         if(i != v.size() - 1) {
+            o << ' ';
+         }
```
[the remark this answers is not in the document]
reneme: I guess, for the time being, we'll just have to either assert that `to_string()` never fails or conditionally unwrap and push_back into a temporary `std::vector<std::string>`. 😞

reneme (changes requested):

src/lib/tls/tls_policy.cpp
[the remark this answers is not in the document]
randombit: Eh, std::vector it is.

I think we actually misbehaved in a similar way before; instead of returning an `optional`, to_string indicated failure by returning an empty string, so we'd get a double space. Which I think the parser should handle but IDK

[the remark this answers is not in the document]
randombit: Changed to vector

reneme (approved):

PR #3730, opened 2023-10-05
Update XLC build configuration

randombit:
Tested with XLC 17.1 on AIX 7.3.3 (gcc119)

Compiles, but there are a number of apparent miscompilations causing test failures, including in big integer multiplication, base58, Poly1305, and Curve25519. The BigInt miscompilation prevents running many tests since RSA being broken causes most certificate tests to fail or hang.

A number of other tests (including Kyber, Dilithium, XMSS, ciphers, hashes, KDFs, Argon2, scrypt, Ed25519) do pass.


randombit:
Just checked, even with `-O0` the miscompilations persist :/

PR #3732, opened 2023-10-06
[TLS 1.3] Test Hybrid Schemes Online

### LINKED ISSUE #3731: Tests against test.openquantumsafe.org fail for x25519/Kyber
Everything works fine for pure Kyber and the NIST+Kyber hybrids, but fails with a handshake failure for X25519+Kyber

Cloudflare's X25519+Kyber seems ok testing against cloudflare.com

fyi @reneme - we should probably address this or else disable OQS X25519+Kyber before shipping 3.2.

reneme:
This introduces an online test in `test_cli.py` that attempts TLS 1.3 handshakes using hybrid PQ/T key exchange schemes against a number of pre-production services. Namely:
* test.openquantumsafe.org
* pq.cloudflareresearch.com, and
* qsc.eu-de.kms.cloud.ibm.com.

To facilitate that, I adapted `./botan tls_client` to terminate with a non-zero status code if neither the peer nor us have sent/received a proper _close_notify_ alert. Most notably: Handshake failures now result in a non-zero result code. Also the `test_cli()` helper method now supports an optional `timeout`.

The online tests are run only when `test_cli.py` is launched with `--run-online-tests`. I did not reconfigure any CI configuration to do that so that we're not dependent on external (beta) services. @randombit Should we have some nightly that runs those tests?

To try, after building the CLI, run this:

```bash
src/scripts/test_cli.py --run-online-tests $(pwd)/botan pqc
```

Closes #3731

### DIFF
--- a/src/cli/tls_client.cpp
+++ b/src/cli/tls_client.cpp
@@ -39,11 +39,13 @@ namespace {
 
 class Callbacks : public Botan::TLS::Callbacks {
    public:
-      Callbacks(TLS_Client& client_command) : m_client_command(client_command) {}
+      Callbacks(TLS_Client& client_command) : m_client_command(client_command), m_peer_closed(false) {}
 
       std::ostream& output();
       bool flag_set(const std::string& flag_name) const;
       void send(std::span<const uint8_t> buffer);
+
+      int peer_closed() const { return m_peer_closed; }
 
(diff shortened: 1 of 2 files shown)


randombit (approved):
Nice, thanks for digging into this. Nightly run of this seems good enough; I expect otherwise we'll get (more) flaky CI builds due to networking problems. We can do this in a followup.

PR #3733, opened 2023-10-06
[TLS 1.3] Clean up support Hybrid PQ/T algorithms

reneme:
### Pull Request Dependencies

Note that this is not a hard dependency. If we decide to not include the CLI online test, this can be rebased/merged without an issue.

* #3732

### Description

In the meantime, Cloudflare as well as OQS use the same `Group_Param` for _x25519/Kyber768_ (namely the one defined in [X25519Kyber768Draft00](https://datatracker.ietf.org/doc/draft-tls-westerbaan-xyber768d00)). Therefore, we can omit the [Extrawurst](https://en.wikipedia.org/wiki/Extrawurst#German_idiomatic_expression) for this combination.

Also, this adds _p256/Kyber768_ (supported by OQS) that might become relevant in case people don't trust Kyber512 enough.

### DIFF
--- a/src/cli/tls_client.cpp
+++ b/src/cli/tls_client.cpp
@@ -39,11 +39,13 @@ namespace {
 
 class Callbacks : public Botan::TLS::Callbacks {
    public:
-      Callbacks(TLS_Client& client_command) : m_client_command(client_command) {}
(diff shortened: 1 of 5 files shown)


randombit (approved):
:+1:

PR #3736, opened 2023-10-07
FIX: out-of-bounds access of std::array

### LINKED ISSUE #3734: Undefined behavior in AlignmentBuffer::fill_up_with_zeros
Set `CXXFLAGS` to `-D_GLIBCXX_DEBUG` (if you're using the GNU libc) or to `-D_LIBCPP_DEBUG=1` (if you're using the LLVM libc). Or set both if you're unsure. Then compile Botan and the reproducer below.

```cpp
#include <botan/kdf.h>

int main(void)
{
    const std::vector<uint8_t> password{0xc0, 0x99, 0xf7, 0x37, 0x45, 0x13, 0xe6, 0xc8, 0xd6, 0xa8};
    const std::vector<uint8_t> salt{0xb4, 0xad, 0xdb, 0x86, 0x49, 0x8b, 0xb9, 0xcc, 0x8a, 0x02, 0xd5, 0xa0, 0x60, 0xc0, 0x97, 0x30,
        0xa8, 0x6a, 0x66, 0x98, 0x52, 0x9f, 0xa4, 0x04, 0x41, 0x0a, 0xbd, 0x9e, 0x59, 0x55, 0x0f, 0x82,
        0x7f, 0x1f, 0x06, 0xda, 0xa0, 0x0f, 0x03, 0xcc, 0x97, 0xcf, 0x71, 0x82, 0x54, 0x00, 0xb7, 0x1e,
        0xbc, 0x50, 0x34, 0xc3, 0xac, 0x5a, 0x57, 0x9e, 0x16, 0xcf, 0x7d, 0xd2, 0xd5, 0x4a, 0x12, 0x86,
        0xd8, 0x79, 0x7a, 0xd9, 0xfd, 0x03, 0x1c, 0xfb, 0x34, 0x86, 0x5a, 0x0d, 0x58, 0xaf, 0x9f, 0x36,
        0x58, 0x22, 0x50, 0x99, 0x14, 0x83, 0xd0, 0x64, 0x74, 0x7a, 0x89, 0xb2, 0x5d, 0xab, 0xe1, 0xb2,
        0xb9, 0xc3, 0x38, 0x87, 0x69, 0x5f, 0xfb, 0x54, 0x58, 0x28, 0x8d, 0x42, 0x67, 0xc7, 0x54, 0x79,
        0x7e, 0xdb, 0xb5, 0x1d, 0x84, 0x16, 0x1a, 0x51, 0x5d, 0x10, 0xe4, 0x41, 0xf1, 0x78, 0x7a, 0x0d,
        0x22, 0xf4, 0x42, 0xde, 0xdd, 0x25, 0xa6, 0xda, 0x05, 0xb4, 0x09, 0x25, 0x7e, 0x47, 0xd4, 0xe7,
        0x30, 0x85, 0x66, 0x49, 0xfd, 0x10, 0x29, 0x6b, 0xd3, 0x92, 0x85, 0x78, 0x8d, 0xb0, 0xdd, 0x91,
        0x1b, 0x0f, 0x9b, 0xe0, 0x5f, 0x58, 0x1f, 0x86, 0x3d, 0x2d, 0xd5, 0xb4, 0xa2, 0x37, 0x1e, 0x84,
        0x6c, 0x18, 0x61, 0x1f, 0x85, 0x7b, 0x09, 0xae, 0xd2, 0x00, 0x75, 0x5a, 0x38, 0x55, 0x0d};
    const std::vector<uint8_t> info{0x42};
    std::unique_ptr<::Botan::KDF> hkdf = nullptr;
    hkdf = ::Botan::KDF::create("HKDF(SHA-256)");
    hkdf->derive_key(1264, password, salt, info);
    return 0;
}
```

Output:

```
/usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/debug/array:155:
In function:
    std::array::reference std::array<unsigned char, 
    64>::operator[](std::array::size_type) [_Tp = unsigned char, _Nm = 64]

Error: attempt to subscript container with out-of-bounds index 64, but 
container only holds 64 elements.

Objects involved in the operation:
    sequence "this" @ 0x0x55ec329a3818 {
      type = std::__debug::array<unsigned char, 64ul>;
    }
Aborted (core dumped)
```

The reason is that `m_buffer` (which has 64 elements) is accessed at element 64:

https://github.com/randombit/botan/blob/30ecb87197a4bb94fc660a2e6d215b20a9ac7f06/src/lib/utils/alignment_buffer.h#L80

It is illegal to access a container with an out-of-bounds index, even if it isn't dereferenced.

reneme:
This adds a check to skip filling an `AlignmentBuffer<>` with zeros when it is already full (i.e. `ready_to_consume()`). Many thanks to @guidovranken for spotting this just before the 3.2.0 release and the detailed write-up.

From what I can see, this wouldn't have resulting in an actual memory bug, though. Albeit the pointer passed to `clear_mem` being bogus, it would have been ignored due to the `size` being zero. Nevertheless, such things are accidents waiting to happen. This will benefit from another look after the release (and once https://github.com/randombit/botan/pull/3715 is merged).

@guidovranken I'd be glad if you could confirm that the patch actually fixes this.

Closes #3734.

### DIFF
--- a/src/lib/utils/alignment_buffer.h
+++ b/src/lib/utils/alignment_buffer.h
@@ -77,8 +77,10 @@ class AlignmentBuffer {
        * Fills the currently unused bytes of the buffer with zero bytes
        */
       void fill_up_with_zeros() {
-         clear_mem(&m_buffer[m_position], elements_until_alignment());
-         m_position = m_buffer.size();
+         if(!ready_to_consume()) {
+            clear_mem(&m_buffer[m_position], elements_until_alignment());
+            m_position = m_buffer.size();
+         }
       }
 
       /**
--- a/src/tests/test_utils_buffer.cpp
+++ b/src/tests/test_utils_buffer.cpp
@@ -269,6 +269,17 @@ std::vector<Test::Result> test_alignment_buffer() {
 
                result.test_is_eq("prefix", v(out.first(16)), v(first_half_data));
                result.test_is_eq("zero-padding", v(out.last(16)), std::vector<uint8_t>(16, 0));
+
+               // Regression test for GH #3734
+               // fill_up_with_zeros() must work if called on a full (ready to consume) alignment buffer
+               b.append(data);
+               result.confirm("ready_to_consume()", b.ready_to_consume());
+               result.test_eq("elements_until_alignment()", b.elements_until_alignment(), 0);
+
+               b.fill_up_with_zeros();
+               const auto out_without_padding = b.consume();
+
+               result.test_is_eq("no padding", v(out_without_padding), v(data));
             }),
 
       CHECK("Handle unaligned data in Alignment Buffer (no block-defer)",


guidovranken:
Testing now.

guidovranken:
Confirmed that this patch resolves the issue.

randombit (approved):
I'm a little put off that somehow nothing in the test suite caught this before merge, even though we do run CI with ASan on 3 different compilers. I'm not sure why that is.

Anyway patch looks good, thanks @reneme, and of course @guidovranken for CryptoFuzz

reneme:
ASan doesn't pick this up because there was no out-of-bounds memory access. "Only" the pointer arithmetic in std::array went out of bounds, though the memory behind the pointer was never accessed.

I guess that the debug flag enables bounds checks on std::array which picks up on it.

thesamesam:
> ASan doesn't pick this up because there was no out-of-bounds memory access. "Only" the pointer arithmetic in std::array went out of bounds, though the memory behind the pointer was never accessed.
> 
> I guess that the debug flag enables bounds checks on std::array which picks up on it.

For libstdc++, `-D_GLIBCXX_ASSERTIONS` should catch this. For libc++, you need `-D_LIBCPP_ENABLE_ASSERTIONS=1` or for >= 18 `-D_LIBCPP_ENABLE_HARDENED_MODE=1`.

You _can_ use the debug options for either libstdc++ or libc++, but those break ABI. In CI though, that should be fine for you to use and will catch more.

PR #3738, opened 2023-10-07
OS::set_thread_name attempt to port to windows >= 10

devnexen:


src/lib/utils/os_utils.cpp
```diff
@@ -624,8 +628,22 @@ void OS::set_thread_name(std::thread& thread, const std::string& name) {
    #elif defined(BOTAN_TARGET_OS_IS_NETBSD)
    static_cast<void>(pthread_set_name_np(thread.native_handle(), "%s", const_cast<char*>(name.c_str())));
    #elif defined(BOTAN_TARGET_OS_HAS_WIN32) && defined(BOTAN_BUILD_COMPILER_IS_MSVC)
-   // Using SetThreadDescription from Win10
-   BOTAN_UNUSED(thread, name);
+   // SetThreadDescription
+   HMODULE kern = GetModuleHandleA("KernelBase.dll");
+   HRESULT(WINAPI * set_thread_name)
+   (HANDLE, PCWSTR) = reinterpret_cast<HRESULT(WINAPI*)(HANDLE, PCWSTR)>(GetProcAddress(kern, "SetThreadDescription"));
```
randombit, line 634: A local typedef would probably make this declaration and cast a little easier on the eyes.

src/lib/utils/os_utils.cpp
```diff
@@ -624,8 +628,22 @@ void OS::set_thread_name(std::thread& thread, const std::string& name) {
...
+   if(set_thread_name) {
+      std::wstring w;
+      auto sz = MultiByteToWideChar(CP_UTF8, 0, name.data(), static_cast<int>(name.size()), nullptr, 0);
+      if(sz > 0) {
+         w.resize(sz);
+         if((sz = MultiByteToWideChar(CP_UTF8, 0, name.data(), static_cast<int>(name.size()), &w[0], sz)) > 0) {
+            w.resize(sz - 1);
```
randombit, line 641: Why the resize here?

randombit (commented):
Yeah this should work. Left a couple questions/comments

src/lib/utils/os_utils.cpp
devnexen: Because sz is possibly updated earlier.

randombit: Still don't understand :sweat_smile: 

First, can the two calls to `MultiByteToWideChar` really return different lengths depending on if the output buffer is set or not? The input string is the same.

Secondly, IIUC  `MultiByteToWideChar` returns the number of characters written. Doesn't this resize of `sz - 1` chop off the last character then?

devnexen: Your second point makes sense, let's see.

randombit (approved):
Looks good, thanks!

PR #3739, opened 2023-10-08
In the CLI tests also test Kyber kex against google.com

randombit:
I think this is worthwhile since I'd guess this is BoringSSL under the hood, rather than the Golang fork that Cloudflare is running.


reneme (approved):

reneme:
Nice! I wasn't aware that vanilla google.com also allows Kyber already.

PR #3740, opened 2023-10-09
CI: Test nightly Hybrid PQ/T TLS interoperability

reneme:
This adds a nightly job to run [the PQ/T interoperability tests](https://github.com/randombit/botan/pull/3732) against online (beta) services. We don't do that on normal CI runs to avoid introducing a(nother) direct dependency to 3rd party services there.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -56,6 +56,23 @@ jobs:
       - name: Valgrind Checks
         run: python3 ./src/scripts/ci_build.py --cc=gcc --make-tool=make valgrind-full
 
+  hybrid_tls_interop:
+    name: "PQ/T TLS 1.3"
+
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Setup Build Agent
+        uses: ./.github/actions/setup-build-agent
+        with:
+          target: hybrid-tls13-interop-test
+          cache-key: linux-x86_64-hybrid_tls
+
+      - name: Hybrid PQ/T TLS 1.3 Online Interop Checks
+        run: python3 ./src/scripts/ci_build.py --cc=gcc --make-tool=make hybrid-tls13-interop-test
+
   tls_anvil_server_test:
     name: "TLS-Anvil (server)"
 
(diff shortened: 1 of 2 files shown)


src/scripts/ci_build.py
```diff
@@ -684,7 +688,10 @@ def main(args=None):
             if options.compiler_cache is not None:
                 cmds.append([options.compiler_cache, '--show-stats'])
 
-            make_targets = ['libs', 'tests', 'cli']
+            make_targets = ['libs', 'cli']
+
+            if run_test_command is not None:
+                make_targets += ['tests']
```
reneme, line 694: Is that a good idea? Basically not to build the test executable if this particular CI run isn't planning to run it anyway?

Also: This selection of make targets is sort-of a repetition of the selection of build targets in the helper method in this very same script. Perhaps we should consider cleaning that up at some point:

https://github.com/randombit/botan/blob/68ca91208e2c640fab835f3ba40d5e37145beac4/src/scripts/ci_build.py#L89-L111

.github/workflows/nightly.yml
```diff
@@ -11,6 +11,12 @@ permissions:
   # implicitly all other scopes not listed become none
 
 on:
+  workflow_dispatch:
+  push:
+    paths:
+      # Run if a pull request changes this workflow to
+      # validate it works properly before merging.
+      - '.github/workflows/nightly.yml'
```
reneme, line 19: This doesn't seem to work. My suspicion is that this configuration takes effect only once it was merged. I.e. with the next change to `nightly.yml`. Though, admittedly that might be wishful thinking.

randombit: I guess the easiest thing to do is just to merge this and then open a test PR that tries modifying it and see if it works.

src/scripts/ci_build.py
randombit: Hrm - that will disable building the tests also on CI jobs where we don't currently run the tests (Android NDK, iOS, Emscripten) which isn't great. Considering this just saves on building the tests in a nightly job, I'd be inclined to continue building the tests here.

I agree this script is getting a little out of control :sweat_smile: I had ideas on rebuilding it such that most of this decision logic is driven by a text configuration rather than embedding everything about the targets into the script itself, which might help.

reneme: Reverted.

randombit (approved):

PR #3741, opened 2023-10-09
Add missing simd requires to aes_ni

lieser:


reneme (approved):
I've stumbled over that recently as well, while working on the module dependencies of FrodoKEM (#3679). You beat me to submitting a fix. Thanks.

randombit:
Thanks!

BTW if unaware, there is a script that is helpful for detecting issues like this - `src/scripts/test_all_configs.py` which tries to do a build and optionally run the tests in two modes - a build with any single module disabled, and also minimized builds with exactly one named module included. I only run it occasionally, and always on very big machines. I think it would be to slow even for a nightly job given just 2 cores, but maybe we could adapt some variation of this idea for a nightly.

PR #3743, opened 2023-10-09
Fix FFDHE selection in TLS to conform to RFC 7919

### LINKED ISSUE #3742: Fix failing TLS-Anvil test caused by not recognizing DH groups
Oops - I knew that did something.

I only noticed this after spinning the 3.2.0 release :/ probably a lesson that the previous idea of no major code changes in the 5-7 days before the release was there for a reason.

```
Error: server.tls12.rfc7919.FfDheShare.abortsWhenGroupsDontOverlap - Unexpected result 'FULLY_FAILED'
  
  Class Name: 'de.rub.nds.tlstest.suite.tests.server.tls12.rfc7919.FfDheShare'
  Method Name: 'abortsWhenGroupsDontOverlap'
  
  RFC 7919, Section 4. Server Behavior:
  If a compatible TLS server receives a Supported Groups extension from a client that includes any FFDHE group (i.e., any codepoint between 256 and 511, inclusive, even if unknown to the server), and if none of the client-proposed FFDHE groups are known and acceptable to the server, then the server MUST NOT select an FFDHE cipher suite.[...]If the extension is present with FFDHE groups, none of the client's offered groups are acceptable by the server, and none of the client's proposed non-FFDHE cipher suites are acceptable to the server, the server MUST end the connection with a fatal TLS alert of type insufficient_security(71).
  
  Result: FULLY_FAILED (expected STRICTLY_SUCCEEDED)
```

randombit:
This logic was broken by the change in #3729

Closes #3742

### DIFF
--- a/src/lib/tls/tls12/msg_server_kex.cpp
+++ b/src/lib/tls/tls12/msg_server_kex.cpp
@@ -50,9 +50,12 @@ Server_Key_Exchange::Server_Key_Exchange(Handshake_IO& io,
       m_shared_group = Group_Params::NONE;
 
       /*
-      If the client does not send any DH groups that we recognize in
-      the supported groups extension, but does offer DH ciphersuites,
-      we select a group arbitrarily
+      RFC 7919 requires that if the client sends any groups in the FFDHE
+      range, that we must select one of these. If this is not possible,
+      then we are required to reject the connection.
+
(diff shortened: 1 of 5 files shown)


reneme (approved):
I like the added details in the comments around that.

PR #3745, opened 2023-10-10
Fix: Unintended Kyber Warnings

atreiber94:
Examples are not supposed to be part of the library so due to the -DBOTAN_IS_BEING_BUILT flag, warnings are suppressed (e.g., the Kyber example used deprecated Kyber enum values).

Additionally, simply including kyber.h resulted in deprecation warnings, which is fixed in the PR.

Noticed by @henningM1r


randombit (approved):
LGTM, thanks!

PR #3752, opened 2023-10-11
Split out allocator helpers to allocator.h

randombit:
See #3715 for context

### DIFF
--- a/src/lib/utils/mem_ops.h
+++ b/src/lib/utils/mem_ops.h
@@ -14,35 +14,18 @@
 #include <type_traits>
 #include <vector>
 
+/*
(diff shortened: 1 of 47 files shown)


src/lib/utils/mem_ops.h
```diff
@@ -8,6 +8,7 @@
 #ifndef BOTAN_MEMORY_OPS_H_
 #define BOTAN_MEMORY_OPS_H_
 
+#include <botan/allocator.h>
```
reneme, line 11: Do we need to mention in a comment that this include is here for backward compatibility? Maybe even with a link to the discussion on GitHub. There's a non-zero chance that we'll run include-what-you-use or similar and will just drop that.

PR #3754, opened 2023-10-13
FIX: TLS 1.3 server hit assertion if application did not provide a cert chain

reneme:
### Description

Server applications are expected to provide a certificate chain via `Credentials_Manager::find_cert_chain()`. This callback defines constraints that the certificate must fulfill. Applications may respond with an empty chain, if the requested certificate constraints are not met. In this case, the TLS 1.3 server implementation did not abort the handshake but tried replying with an empty Certificate message. This, in turn, resulted in the triggering of an assertion when trying to create a CertificateVerify message accompanying the (empty) Certificate message.

### Impact

As some of the certificate constraints can be defined by the peer, this bug may potentially be triggered remotely. In the default build configuration, botan-assertions result in an exception. The TLS server would then report an `internal_error` alert, which should be okay in the described situation. However, if the build configuration enabled `--terminate-on-assert` this would be a potential DoS vector.

### Mitigation

Applications can avoid this issue by throwing an exception in their implementation of `Credentials_Manager::find_cert_chain()` instead of returning an empty certificate chain in the described situation. For instance: `Botan::TLS::TLS_Exception(Botan::TLS::Alert::HandshakeFailure, "<some message>")`.

### DIFF
--- a/src/lib/tls/tls13/msg_certificate_13.cpp
+++ b/src/lib/tls/tls13/msg_certificate_13.cpp
@@ -174,14 +174,20 @@ Certificate_13::Certificate_13(const Client_Hello_13& client_hello,
       m_request_context(), m_side(Connection_Side::Server) {
    BOTAN_ASSERT_NOMSG(client_hello.extensions().has<Signature_Algorithms>());
 
-   setup_entries(
+   auto cert_chain =
       credentials_manager.find_cert_chain(filter_signature_schemes(client_hello.signature_schemes()),
                                           to_algorithm_identifiers(client_hello.certificate_signature_schemes()),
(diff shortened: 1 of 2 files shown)


randombit:
> However, if the build configuration enabled --terminate-on-assert this would be a potential DoS vector.

I don't really consider this an issue in that `terminate-on-assert` is very much intended for testing/fuzzing and should not be used in production.

randombit (approved):
Change looks good

reneme:
> I don't really consider this an issue in that terminate-on-assert is very much intended for testing/fuzzing and should not be used in production.

That was my understanding as well.

PR #3756, opened 2023-10-13
Use concat in DLIES and ECIES

randombit:


src/lib/pubkey/dlies/dlies.cpp
```diff
@@ -76,9 +77,10 @@ std::vector<uint8_t> DLIES_Encryptor::enc(const uint8_t in[], size_t length, Ran
...
    secure_vector<uint8_t> out(m_own_pub_key.size() + ciphertext.size() + tag.size());
-   buffer_insert(out, 0, m_own_pub_key);
-   buffer_insert(out, 0 + m_own_pub_key.size(), ciphertext);
-   buffer_insert(out, 0 + m_own_pub_key.size() + ciphertext.size(), tag);
+   BufferStuffer stuffer(out);
+   stuffer.append(m_own_pub_key);
+   stuffer.append(ciphertext);
+   stuffer.append(tag);
 
    return unlock(out);
```
reneme, line 85: For simple concatenations, just go for: `return concat(m_own_pub_key, ciphertext, tag)`. This uses the container type of the first parameter as the target container type. If that's not what you want, `concat_as<std::vector<uint8_t>>()` has your back. Maybe even avoid copying the `tag` variable by directly finalizing the MAC into the `concat()`:

```C++
m_mac->update(ciphertext);

return concat_as<std::vector<uint8_t>>(m_own_pub_key, ciphertext, m_mac->final_stdvec());
```

src/lib/pubkey/ecies/ecies.cpp
```diff
@@ -272,19 +273,21 @@ std::vector<uint8_t> ECIES_Encryptor::enc(const uint8_t data[],
...
+
+   // concat elements
+
+   std::vector<uint8_t> out(m_eph_public_key_bin.size() + encrypted_data.size() + mac.size());
+   BufferStuffer stuffer(out);
+   stuffer.append(m_eph_public_key_bin);
+   stuffer.append(encrypted_data);
+   stuffer.append(mac);
 
    return out;
```
reneme, line 292: Same as above, simply:

```C++
return concat_as<std::vector<uint8_t>>(m_eph_public_key_bin, encrypted_data, m_mac->final_stdvec());
```

reneme (changes requested):

src/lib/pubkey/dlies/dlies.cpp
randombit: Even shorter :)

reneme (approved):

PR #3758, opened 2023-10-13
OS::set_thread_name for Haiku using native BeOS api

devnexen:


src/lib/utils/os_utils.cpp
```diff
@@ -73,10 +73,14 @@ extern "C" char** environ;
    #include <sys/prctl.h>
 #endif
 
-#if defined(BOTAN_TARGET_IS_FREEBSD) || defined(BOTAN_TARGET_IS_OPENBSD)
+#if defined(BOTAN_TARGET_OS_IS_FREEBSD) || defined(BOTAN_TARGET_OS_IS_OPENBSD)
```
randombit, line 76: Oof on my part missing this in review

randombit (approved):
Looks good thanks

src/lib/utils/os_utils.cpp
devnexen: yes easy to miss :)

PR #3757, opened 2023-10-13
Remove use of buffer_insert from filters, MACs, stream ciphers

randombit:

PR #3760, opened 2023-10-15
Move const time memory comparison utils to ct_utils.h

randombit:


src/lib/utils/ct_utils.h
```diff
@@ -324,6 +331,30 @@ inline CT::Mask<T> all_zeros(const T elem[], size_t len) {
...
+template <typename T>
+inline CT::Mask<T> is_equal(const T x[], const T y[], size_t len) {
+   volatile T difference = 0;
+
+   for(size_t i = 0; i != len; ++i) {
+      difference = difference | (x[i] ^ y[i]);
+   }
+
+   return CT::Mask<T>::is_zero(difference);
+}
```
reneme, line 347: This could be based on ranges, no? Something along these lines, with `CT::Mask<>` fairy-dust sprinkled in:

```suggestion
template <std::ranges::contiguous_range X, std::ranges::contiguous_range Y>
    requires std::same_as<std::ranges::range_value_t<X>, std::ranges::range_value_t<Y>>
constexpr auto is_equal(X&& x, Y&& y)
{
    volatile auto difference = static_cast<std::ranges::range_value_t<X>>(std::ranges::size(x) != std::ranges::size(y));

    auto ix = std::ranges::begin(x);
    auto iy = std::ranges::begin(y);

    // TODO C++23: std::views::zip(x, y)
    //             future-me, please make sure that zip() is constant-time
    while(ix != std::ranges::end(x) && iy != std::ranges::end(y))
    {
        difference |= *ix ^ *iy;
        ++ix;
        ++iy;
    }

    return difference;
}```

File: src/lib/utils/ct_utils.h [hunk 2]
```diff
@@ -77,9 +77,11 @@ inline void unpoison(T& p) {
 * since you never know what a compiler might do.
 */
 template <typename T>
-   requires(std::is_unsigned<T>::value && !std::is_same<bool, T>::value)
 class Mask final {
    public:
+      static_assert(std::is_unsigned<T>::value && !std::is_same<bool, T>::value,
+                    "Only unsigned integer types are supported by CT::Mask");
+
```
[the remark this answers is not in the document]
randombit: No problem, just that the error message you get with `static_assert` is much more clear and to the point than when the concept restrictions don't apply to the type.

src/lib/utils/ct_utils.h
randombit: Maybe so. But tbh for this kind of code I like writing things in the most brain-dead way possible because it's usually (fairly) obvious what the compiler is going to do with it.

Some span based overloads here might be nice though. As with #3757 the goal here is primarily to remove usage of functions in `mem_ops.h` especially when they were already redundant with other functionality.

PR #3762, opened 2023-10-16
Introduce TLS::Channel::is_handshake complete()

reneme:
... and fix a minor issue in the ASIO stream using it. Namely, that `async_handshake` did prematurely notify "completion" in the TLS 1.3 server use case. For a typical application, nothing bad will happen except that it'll get the handshake completion callback prematurely.

Background: Technically, RFC 8446 Section 4.4.4 allows the server to send application data as soon as it sent its `Finished` message -- without waiting for the client's `Finished`:

> Servers MAY send data after sending their first flight, but because the handshake is not yet complete, they have no assurance of either the peer's identity or its liveness [...]

The semantics of `TLS::Channel::is_active() == true` is that "we can write". Therefore, the predicate method `TLS::Channel::is_active()` becomes `true` _before_ the TLS 1.3 server receives the client's `Finished`.

The ASIO stream's `AsyncHandshakeOperation` used to rely on `is_active()` to detect success. Now it relies on `is_handshake_complete()`, fixing the above-described discrepancy.

### DIFF
--- a/src/lib/tls/asio/asio_async_ops.h
+++ b/src/lib/tls/asio/asio_async_ops.h
@@ -278,7 +278,7 @@ class AsyncHandshakeOperation : public AsyncBase<Handler, typename Stream::execu
                return;
             }
 
-            if(!m_stream.native_handle()->is_active() && !ec) {
+            if(!m_stream.native_handle()->is_handshake_complete() && !ec) {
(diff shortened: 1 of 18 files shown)


randombit (approved):
LGTM

PR #3763, opened 2023-10-17
Rewrite ./botan tls_http_server using Boost Beast

### LINKED ISSUE #3659: Rewrite ./botan tls_http_server using the ASIO stream
It seems the implementation for said toy HTTP server is quite dated and bug prone. I envision to use Boost Beast to replace the minimal HTTP responder that is currently used on top of the ASIO stream. To be clear: this does not aim to implement a fully functional HTTP server. Instead, it should serve as a nice showcase for how to combine ASIO, Botan's TLS and Beast to build an HTTPS server.

Perhaps, @hrantzsch's [example code](https://github.com/hrantzsch/botan-tls-stream-server-example/blob/master/server.cpp) could be a good starting point for this. This example was initially adapted from [Boost Beast's server example](https://www.boost.org/doc/libs/master/libs/beast/example/advanced/server/advanced_server.cpp) in early 2020 and may or may not need some refreshing with the latest Boost code.

reneme:
### Pull Request Dependencies

* ~~#3764~~

### Description

Using the modernizations of #3764, this re-implements the `./botan tls_http_server` using Botan's TLS asio stream and boost Beast with C++20 co-routines.

Currently, in CI this will compile on Windows, macOS and Linux (with GCC). On Linux (with clang) the boost version provided by `apt install` isn't recent enough to support co-routines. As a result, the `./botan tls_http_server` command is not available and the respective tests in `test_cli.py` are skipped.

Closes #3659.

### DIFF
--- a/src/cli/tls_http_server.cpp
+++ b/src/cli/tls_http_server.cpp
@@ -1,37 +1,47 @@
 /*
-* (C) 2014,2015,2017,2019 Jack Lloyd
+* (C) 2014,2015,2017,2019,2023 Jack Lloyd
 * (C) 2016 Matthias Gierlings
-* (C) 2023 René Meusel, Rohde & Schwarz Cybersecurity
+* (C) 2023 Fabian Albert, René Meusel - Rohde & Schwarz Cybersecurity
+*
+* This implementation is roughly based on this BSL-licensed example
+* by Klemens D. Morgenstern:
+*   www.boost.org/doc/libs/1_83_0/libs/beast/example/http/server/awaitable/http_server_awaitable.cpp
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
 
 #include "cli.h"
 
-#if defined(BOTAN_HAS_TLS) && defined(BOTAN_HAS_BOOST_ASIO) && defined(BOTAN_TARGET_OS_HAS_FILESYSTEM)
+#if defined(BOTAN_HAS_BOOST_ASIO)
+   #include <botan/boost_compat.h>
+#endif
 
-   #include <atomic>
-   #include <fstream>
+#if defined(BOTAN_FOUND_COMPATIBLE_BOOST_VERSION)
+   #include <boost/asio/awaitable.hpp>
+#endif
+
+// If your boost version is too old, this might not be defined despite
+// your toolchain supporting C++20 co_await.
+#if defined(BOOST_ASIO_HAS_CO_AWAIT) && defined(BOTAN_HAS_TLS) && defined(BOTAN_TARGET_OS_HAS_FILESYSTEM)
+
+   #include <ctime>
    #include <iomanip>
-   #include <iostream>
    #include <memory>
    #include <string>
    #include <thread>
-   #include <utility>
    #include <vector>
 
-   #include <botan/internal/os_utils.h>
-   #include <boost/asio.hpp>
-   #include <boost/bind.hpp>
+   #include <boost/asio/co_spawn.hpp>
+   #include <boost/asio/ip/tcp.hpp>
+   #include <boost/asio/use_awaitable.hpp>
+   #include <boost/beast/http.hpp>
 
-   #include <botan/hex.h>
-   #include <botan/pkcs8.h>
-   #include <botan/rng.h>
+   #include <botan/asio_stream.h>
    #include <botan/tls_messages.h>
-   #include <botan/tls_server.h>
    #include <botan/tls_session_manager_memory.h>
    #include <botan/version.h>
-   #include <botan/x509cert.h>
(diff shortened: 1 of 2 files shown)


src/cli/tls_http_server.cpp
```diff
@@ -43,305 +49,51 @@ namespace Botan_CLI {
...
-
-            if(headers.size() > 1024) {
-               throw Botan::Invalid_Argument("Too many HTTP headers sent in request");
-            }
-         }
+   #if defined(BOTAN_BUILD_COMPILER_IS_MSVC) || defined(BOTAN_TARGET_OS_IS_MINGW) || \
+      defined(BOTAN_TARGET_OS_IS_CYGWIN) || defined(BOTAN_TARGET_OS_IS_WINDOWS)
+         localtime_s(&tm, &t);
+   #else
+         localtime_r(&t, &tm);
```
randombit, line 70: Can we use `std::chrono::local_time` here now?

src/cli/tls_http_server.cpp
```diff
@@ -393,79 +145,147 @@ class TLS_Asio_HTTP_Session final : public std::enable_shared_from_this<TLS_Asio
...
+         return resp;
+      } else if(req.target() == "/" || req.target() == "/status") {
+         auto resp = std::make_shared<http::response<http::string_body>>(http::status::ok, req.version());
+         resp->body() = callbacks->summary() + summarize_request(tls_stream, req);
+         return resp;
+      } else {
+         auto resp = std::make_shared<http::response<http::string_body>>(http::status::not_found, req.version());
+         resp->body() = std::string("Not found");
+         return resp;
+      }
```
randombit, line 210: I guess it's not a huge deal but this lambda seems really odd to me. There are really only two things that change depending on the params: the response code and the body. Instead there is a lot of repetition in the three cases. Could we do something like

```
// 
auto [response_code, body] = ...;
auto resp = std::make_shared<...>(response_code, req.version());
resp->body() = std::move(body);
```

randombit (commented):
Don't have time for a full review right now. Generally a fine idea (the existing HTTP "parser" is :sob: so happy to see that go)

src/cli/tls_http_server.cpp
reneme: It seems that the ["Calendar and timezone" features](https://en.cppreference.com/w/cpp/compiler_support#C.2B.2B20_library_features) aren't implemented in the STL, yet. :(

lieser (approved):

src/cli/tls_http_server.cpp
```diff
@@ -1,37 +1,43 @@
...
-   #include <boost/bind.hpp>
+   #include <boost/asio/co_spawn.hpp>
+   #include <boost/asio/ip/tcp.hpp>
+   #include <boost/asio/use_awaitable.hpp>
+   #include <boost/beast/http.hpp>
 
-   #include <botan/hex.h>
-   #include <botan/pkcs8.h>
-   #include <botan/rng.h>
+   #include <botan/asio_stream.h>
```
reneme, line 35: After #3765 is merged, this should probably double-check the boost compatibility using `#include <botan/boost_compat.h>`.

src/cli/tls_http_server.cpp
```diff
@@ -43,305 +49,51 @@ namespace Botan_CLI {
...
+      Logger(std::ostream& out, std::ostream& err) : m_out(out), m_err(err) {}
 
-         // Defer the setup of the TLS server to make use of
-         // shared_from_this() which wouldn't work in the c'tor.
-         session->setup(session_manager, credentials, policy);
-
-         return session;
+      void log(std::string_view out) {
+         std::scoped_lock lk(m_mutex);
+         m_out << Botan::fmt("[{}] {}", timestamp(), out) << std::endl;
```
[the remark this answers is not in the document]
reneme: Just muscle memory. Fixed.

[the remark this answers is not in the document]
lieser: You changed it to `"\n"`, not `'\n'`, but I guess it doesn't matter that much 😅,

src/cli/tls_http_server.cpp
randombit: :sob:

randombit (approved):
Thanks looks good and a very nice improvement. Also a good way of testing the waters re coroutines

src/cli/tls_http_server.cpp
reneme: Rebased and amended the commit accordingly.

reneme:
> Also a good way of testing the waters re coroutines

Its definitely nice to see that coroutines pretty much "just worked" with the existing code base. I'm super happy to see this fall into place so nicely.

PR #3764, opened 2023-10-18
Refactor: Modernize asio TLS Stream

reneme:
This updates the asio stream for C++20:

* using boost-provided concepts instead of macros
* preparing its async methods to be used with C++20 co-routines
* reflects the need for at least boost 1.73.0 (released in April 2020) in the code base.

Additionally, it now allows applications to (optionally) override the `TLS::Callbacks` object. This wasn't possible before and therefore limited the customizability of the asio stream significantly.

Re: co-routines: Given a recent-enough toolchain and boost version, one can actually do:

```C++
   auto tcp_stream = /*  some boost-provided TCP stream wrapper */
   Botan::TLS::Stream<tcp_stream> tls_stream(std::move(tcp_stream), std::move(tls_ctx));
   co_await tls_stream.async_handshake(Botan::TLS::Connection_Side::Server);
   co_await tls_stream.write_some(...);
```

### DIFF
--- a/src/lib/tls/asio/asio_stream.h
+++ b/src/lib/tls/asio/asio_stream.h
@@ -127,6 +127,16 @@ class StreamCallbacks : public Callbacks {
       std::weak_ptr<TLS::Context> m_context;
 };
 
+namespace detail {
+
+template <typename T>
+concept basic_completion_token = boost::asio::completion_token_for<T, void(boost::system::error_code)>;
+
+template <typename T>
+concept byte_size_completion_token = boost::asio::completion_token_for<T, void(boost::system::error_code, size_t)>;
+
+}  // namespace detail
+
 /**
  * @brief boost::asio compatible SSL/TLS stream
  *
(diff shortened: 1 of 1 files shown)


src/lib/tls/asio/asio_async_ops.h
```diff
@@ -12,7 +12,7 @@
 #include <botan/types.h>
 
 #include <boost/version.hpp>
-#if BOOST_VERSION >= 106600
+#if BOOST_VERSION >= 107300
```
randombit, line 15: My main concern here is this increases the required Boost version (which is fine) but also it's hard for downstream users to know if the asio integration works for their botan+boost combination. They can test if it's available, using the usual feature macro - but they don't know if it will work if they include the header.

Maybe we should define somewhere (in an existing or new header) what our minimum boost version is, wrt this asio integration, which would allow applications to check for availability. (Or maybe you have some better idea of how to address this)

randombit (approved):
lgtm modulo this thing with the versioning; doesn't have to be addressed here - it was always an issue and increasing the Boost version just makes the issue more immediate.

src/lib/tls/asio/asio_async_ops.h
reneme: That's indeed a valid point. I'll make a proposal in a follow-up PR.

PR #3765, opened 2023-10-18
Introduce <botan/boost_compat.h> as feature flag

reneme:
This introduces `#include <botan/boost_compat.h>` that defines `BOTAN_MINIMUM_SUPPORTED_BOOST_VERSION` as the minimum version of boost to be usable by this version of Botan. If the available boost headers are new enough, `BOTAN_FOUND_COMPATIBLE_BOOST_VERSION` is defined as well. In this case, it is safe to `#include <botan/asio_stream.h>`, otherwise a macro-error would be generated.

Usage example:

```C++
#include <botan/boost_compat.h>

#if defined(BOTAN_FOUND_COMPATIBLE_BOOST_VERSION)
#include <botan/asio_stream.h>
// profit...
#else
// fallback...
#endif

```

### DIFF
--- a/src/lib/utils/boost/boost_compat.h
+++ b/src/lib/utils/boost/boost_compat.h
@@ -1,0 +1,47 @@
+/*
+* Checks compatibility between the existing headers of Botan and boost
+* (C) 2023 Jack Lloyd
+*     2023 René Meusel - Rohde & Schwarz Cybersecurity
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#ifndef BOTAN_BOOST_COMPAT_H_
+#define BOTAN_BOOST_COMPAT_H_
+
+#include <botan/build.h>
+
+#if defined(BOTAN_HAS_BOOST_ASIO)
+
+   #include <boost/version.hpp>
+
+   /** @brief minimum supported boost version for the TLS ASIO wrapper
+    *
+    *  BOOST_VERSION % 100 is the patch level
+    *  BOOST_VERSION / 100 % 1000 is the minor version
+    *  BOOST_VERSION / 100000 is the major version
+    *
+    * Botan may still work with older versions of boost. Though, the asio TLS
+    * wrapper won't work with versions older than the one specified below.
+    *
+    * Also note the changelog with rationales for the required versions:
+    *
+    * until Botan 3.2.0
+    *   1.66.0 - first version to be compatible with Networking TS (N4656) and boost::beast
+    *
+    * as of Botan 3.3.0
+    *   1.73.0 - first version supporting the C++20 concepts syntax
+    */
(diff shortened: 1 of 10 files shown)


src/tests/unit_asio_stream.cpp
```diff
@@ -736,5 +736,5 @@ BOTAN_REGISTER_TEST("tls", "tls_asio_stream", Asio_Stream_Tests);
 
 }  // namespace Botan_Tests
 
-   #endif  // BOOST_VERSION
-#endif     // BOTAN_HAS_TLS && BOTAN_HAS_BOOST_ASIO
+   #endif
+#endif  // BOTAN_HAS_TLS && BOTAN_HAS_BOOST_ASIO
```
lieser, line 740:
```suggestion
#endif  // BOTAN_HAS_TLS && BOTAN_HAS_TLS_ASIO_STREAM
```

src/lib/utils/boost/boost_compat.h
```diff
@@ -0,0 +1,41 @@
...
+    *
+    * until Botan 3.2.0
+    *   1.66.0 - first version to be compatible with Networking TS (N4656) and boost::beast
+    *
+    * as of Botan 3.3.0
+    *   1.73.0 - first version supporting the C++20 concepts syntax
+    */
+   #define BOTAN_MINIMUM_SUPPORTED_BOOST_VERSION 107300
+
+   #include <boost/version.hpp>
```
lieser, line 30: Move include before the define?

src/lib/utils/boost/boost_compat.h
```diff
@@ -0,0 +1,41 @@
...
+    *  BOOST_VERSION / 100 % 1000 is the minor version
+    *  BOOST_VERSION / 100000 is the major version
+    *
+    * until Botan 3.2.0
+    *   1.66.0 - first version to be compatible with Networking TS (N4656) and boost::beast
+    *
+    * as of Botan 3.3.0
+    *   1.73.0 - first version supporting the C++20 concepts syntax
+    */
+   #define BOTAN_MINIMUM_SUPPORTED_BOOST_VERSION 107300
```
randombit, line 28: This comment should be clarified that it only applies to the asio integrations; we can still use older Boost for eg socket wrappers.

randombit (approved):
Left a couple of minor comments (and I concur on the comments @lieser left) but basically LGTM.

Can you please amend the relevant docs/examples?

src/tests/unit_asio_stream.cpp
reneme: Mhh, somehow, I thought `clang-format` was responsible for that. Fixed.

reneme:
I applied the suggestions, extended `src/examples/tls_stream_client.cpp` and added a note about `BOTAN_FOUND_COMPATIBLE_BOOST_VERSION` in the documentation.

lieser (approved):

PR #3769, opened 2023-10-23
Ninja: use header dependencies for better rebuild support

lieser:
When locally developing, currently the build system does not properly track changes to headers.
This can result in broken rebuilds, that are currently often workaround by re-invoking `./configure.py` before a rebuild.

Ninja supports for gcc, clang and msvc to track the header dependencies (see https://ninja-build.org/manual.html#ref_headers for more details).
This PR enables usage of the header dependencies feature then using ninja.
When changing headers this can greatly improve the rebuild because only `ninja` needs to be invoked, and a new call to `./configure.py` is not needed.

A note about Windows:
This only works with hard links, as ninja will fail to detect a new modification time of the target file.
For hard links there seems to be also some caching involved, and when modifying the file under `src/` and not `build/`, it sometimes can take a while for ninja to detect the change.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1249,6 +1249,9 @@ class CompilerInfo(InfoObject):
                 'ar_output_to': '',
                 'werror_flags': '',
                 'supports_gcc_inline_asm': 'no',
+                'ninja_header_deps_style': '',
+                'header_deps_flag': '',
+                'header_deps_out': '',
             })
 
         self.add_framework_option = lex.add_framework_option
@@ -1293,6 +1296,9 @@ class CompilerInfo(InfoObject):
         self.warning_flags = lex.warning_flags
         self.werror_flags = lex.werror_flags
         self.minimum_supported_version = lex.minimum_supported_version
+        self.ninja_header_deps_style = lex.ninja_header_deps_style
+        self.header_deps_flag = lex.header_deps_flag
+        self.header_deps_out = lex.header_deps_out
 
     def cross_check(self, os_info, arch_info, all_isas):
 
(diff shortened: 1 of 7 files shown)


lieser:
Opened as a draft as I'm not sure that adding the logic to `build.ninja` is the best approach, or if it would be better to move more into `configure.py` and/or the compiler information under `src/build-data/cc/`.

reneme:
This works well for me and actually helps with one of the bigger pains I have with Botan's build system. 😍

> I'm not sure that adding the logic to build.ninja is the best approach, or if it would be better to move more into configure.py and/or the compiler information under src/build-data/cc/.

I would indeed suggest to add the compiler flags to `src/build-data/cc/*.txt`. 
That would reduce the complexity in the `ninja.in` file quite a bit, I think. Not sure how to handle the difference in the invocations, though. The `-MF $out.d` seems to be necessary, despite it being the default anyway.

randombit:
>  I would indeed suggest to add the compiler flags to src/build-data/cc/*.txt.

Strong :+1: on this aspect - ideally we should not be anywhere asking "is this compiler MSVC" or "is this compiler GCC" but instead "does the compiler have the capability X, and if so how do I invoke it". Then later someone can update the compiler info for say Intel C++, and then everything Just Works without any additional changes.

lieser:
> I would indeed suggest to add the compiler flags to `src/build-data/cc/*.txt`

The problem I see here is that for gcc/clang the file argument `$out.d` to `-MF` is ninja specific. I am a little hesitant to add something build tool specific to this files, but maybe still the best solution.

> Could we create another "abstract" ninja rule, like:

Don't know much about ninja, but from what I read in the specification I don't see any way for rules to kind of inherit properties from each other.

We could try to use top level variables to get rid of some repetition.

```
%{if compiler_is_gcc_or_clang}
write_dependencies_flags = -MD -MF $out.d
dep_file =$out.d
%{endif}
%{if compiler_is_msvc}
write_dependencies_flags = /showIncludes
dep_file = 
%{endif}

rule compile_lib
  depfile = $dep_file
  deps = %{compiler}
  command = %{cxx} $write_dependencies_flags ...
```

But a few thinks are unclear:
- Would an empty `depfile = ` cause problems for msvc? My guess would be no (and a small test so far confirms it)
- If we specify a ninja variable `write_dependencies_flags = -MD -MF $out.d` that is used in the `command`, how would the `$out.d` variable in it behave? Same question for `dep_file`/`depfile`.

randombit (approved):

Issue #3706, opened 2023-09-19
Improve API of asymmetric algorithms


reneme:
This is a collection of potential API improvements found in #3609.

* [ ] **Parameter consistency for generic key handling functions (`pk_algs.h`)**
       The generic function `create_private_key()` takes string-typed `algo_name` and `algo_params` to identify the algorithm (see also #3275). In contrast `load_{public/private}_key()` requires an `AlgorithmIdentifier` object. This is inconvenient, as it requires the user to find a way to translate between the two.

randombit:
Another one: LMS and XMSS both have stateful operation, but there is no generic way to detect how many operations remain on a key. I think the fix here is fairly easy, add `std::optional<uint64_t> remaining_operations() const` on `Private_Key` which returns `nullopt` for non-stateful algorithms.

reneme:
For the record, here's an example of the hoops one needs to jump through to load a private key that is not wrapped in PKCS#8: https://github.com/randombit/botan/discussions/3902#discussioncomment-8476815.

PR #3770, opened 2023-10-23
Feature: `AsymmetricKey::generate_another()`

reneme:
As suggested in #3706, this adds `::generate_another(RNG&)`. Given some (abstract) asymmetric key object, one can simply generate another equivalent key pair that uses the same algorithm and parameters. This comes in handy if one needs to generate an ephemeral key pair for a public key received from a peer.

Currently, the new method is used in the `KEX_to_KEM_Adapter` class, only. Though, the DLIES and ECIES implementation might benefit from them as well. At first glance, their APIs appeared to require a bit of a make-over to benefit from `::generate_another()`. I've left this for future work.
Similarly, the PKCS#11 key variants are left for future work -- currently throwing `Not_Implemented`. It remains to-be-discussed whether `::generate_another()` should generate a software instance of the given asymmetric algorithm or whether they should attempt to generate an equivalent keypair via PKCS#11.


randombit (approved):
:+1:

PR #3771, opened 2023-10-24
Feature: RawPublicKey authentication in TLS 1.3 (RFC 7250)

reneme:
## Pull Request Dependencies

* ~~#3792~~

## Description

This adds support for authentication with "raw public keys" to the TLS 1.3 implementation. This is useful for applications that want to avoid or don't need the complexity of a PKI. Such applications bear the responsibility to manage the trust-relationship of the public key pairs they use.

**The commits build the feature from the ground up**, starting with the public API, new TLS extensions, integration into `Certificate`/`CertificateVerify` messages, and the actual implementation in client and server. I hope this aids in reviewing this patch.

## New user-facing APIs

* **`TLS::Policy`**
  * `::accepted_client_certificate_types()` / `::accepted_server_certificate_types()` determines what trust credentials are supported for client and server authentication respectively. By default, this returns X.509. Applications that want to use raw public keys need to override or configure this accordingly.
* **`Credentials_Manager`**
  * `::find_raw_public_key()` is called when usage of raw public keys was negotiated to obtain a public key from the downstream application. This is the very same concept as the existing `::find_cert_chain()`
  * `::private_key_for(Public_Key&)` is called to obtain the associated private key. It is an overload of the existing `::private_key_for(X509_Certificate&)` method.
* **`TLS::Callbacks`**
  * `::tls_verify_raw_public_key()` must be implemented by the downstream application to establish trust in a raw public key received from a peer. In contrast to `tls_verify_cert_chain()` this does not provide a best-effort default implementation; instead it rejects all raw public keys.
* **`TLS::Channel`**
  * `::peer_raw_public_key()` returns the raw public key used to authenticate this Channel's active session (or `nullptr`) if no raw public key was used.
* **`TLS::Session_Summary`**
  *  `::peer_raw_public_key()` returns the raw public key used to authenticate the session (or `nullptr`) if no raw public key was used.

#### Start a Botan Server

Create a *policy.txt*:

```
allow_tls13 = true
allow_tls12 = false
accepted_server_certificate_types = RawPublicKey
```

... and run the server:

```bash
./botan keygen --algo=RSA > rsa_priv.pem
./botan pkcs8 --pub-out rsa_priv.pem > rsa_pub.pem
./botan tls_server rsa_pub.pem rsa_priv.pem --port=5555 --policy=policy.txt
```

#### Connect with a GnuTLS Client

```bash
gnutls-cli --port=5555 --priority="NORMAL:+CTYPE-RAWPK" --insecure --print-cert localhost
```

#### Start a gnutls server

```bash
./botan keygen --algo=RSA > rsa_priv.pem
./botan pkcs8 --pub-out rsa_priv.pem > rsa_pub.pem
gnutls-serv --rawpkfile=rsa_pubkey.pem --rawpkkeyfile=rsa_privkey.pem --disable-client-cert --priority="NORMAL:+CTYPE-RAWPK" --port=5555
```

#### Connect with a Botan client

Create a *policy.txt*:

```
allow_tls13 = true
allow_tls12 = false
accepted_server_certificate_types = RawPublicKey
```

... and run the client:

```bash
# without explicitly trusting the Raw Public Key
./botan tls_client --port=5555 --policy=policy.txt localhost

# with a trusted public key fingerprint
./botan fingerprint --algo=SHA-256 rsa_pub.pem
# -> copy the fingerprint string into the placeholder below:
./botan tls_client --port=5555 --debug --policy=rawpk_policy.txt --trusted-pubkey-sha256="ba:ad:ca:fe:..." localhost
```


reneme:
This should be ready for a review now.

Two things: 

1. That's the second feature I implemented for TLS 1.3 only (after [RecordSizeLimit](https://www.rfc-editor.org/rfc/rfc8449)). I feel that is something we should track somewhere.
2. It feels inconsistent that we use `std::shared_ptr` or `std::unique_ptr` for asymmetric keys. Having those classes explicitly allow/prohibit to be copied and/or moved would be more modern, in my opinion (like `X509_Certificate`). That might be something to consider for Botan 4.0 when splitting the `Public_Key` and `Private_Key` base classes. _Where would I note that, so that we don't forget it?_

FAlbertDev (commented):
Good job. Looks very clean! I only left some comments and suggestions.

Do we also want to update doc/cli.rst?

src/lib/tls/tls13/tls_client_impl_13.h
```diff
@@ -62,6 +62,11 @@ class Client_Impl_13 : public Channel_Impl_13 {
       */
       std::vector<X509_Certificate> peer_cert_chain() const override;
 
+      /**
+      * @return raw public key of the peer (may be nullptr)
```
lieser, line 66: Docstring indention.

src/lib/tls/tls13/tls_server_impl_13.cpp
```diff
@@ -45,13 +45,25 @@ std::string Server_Impl_13::application_protocol() const {
...
+             m_handshake_state.client_certificate().has_certificate_chain()) {
       return m_handshake_state.client_certificate().cert_chain();
    } else {
       return {};
    }
 }
 
+std::shared_ptr<const Public_Key> Server_Impl_13::peer_raw_public_key() const {
+   if(m_resumed_session.has_value()) {
+      return m_resumed_session->peer_raw_public_key();
```
lieser, line 58: Is there a reason the server prefers the resumed value other the current one, and the client does it the other way around?

src/lib/tls/tls_channel.h
```diff
@@ -137,6 +137,11 @@ class BOTAN_PUBLIC_API(2, 0) Channel {
       */
       virtual std::vector<X509_Certificate> peer_cert_chain() const = 0;
 
+      /**
+      * @return raw public key of the peer (may be nullptr)
```
lieser, line 141: Indention

src/lib/tls/tls_channel_impl.h
```diff
@@ -109,6 +109,11 @@ class Channel_Impl {
       */
       virtual std::vector<X509_Certificate> peer_cert_chain() const = 0;
 
+      /**
+      * @return raw public key of the peer (may be nullptr)
```
lieser, line 113: Indention

src/lib/tls/tls_extensions.cpp
```diff
@@ -356,6 +363,108 @@ std::vector<uint8_t> Application_Layer_Protocol_Notification::serialize(Connecti
...
+}
+
+Certificate_Type_Base::Certificate_Type_Base(TLS_Data_Reader& reader, uint16_t extension_size, Connection_Side from) :
+      m_from(from) {
+   if(extension_size == 0) {
+      throw Decoding_Error("Certificate type extension cannot be empty");
+   }
+
+   if(from == Connection_Side::Client) {
+      const auto type_bytes = reader.get_tls_length_value(1);
```
lieser, line 429: Do we need to check that the `extension_size` matches with how much we read here?

src/lib/tls/tls_messages.h
```diff
@@ -548,23 +548,42 @@ class Certificate_Request_13;
...
+
+            std::vector<uint8_t> serialize() const;
+
+            Extensions& extensions() { return m_extensions; }
+
+            const Extensions& extensions() const { return m_extensions; }
+
+         private:
+            std::optional<X509_Certificate> m_certificate;
+            std::shared_ptr<Public_Key> m_raw_public_key;
```
lieser, line 570: Did you consider using a variant for this?

src/lib/tls/tls_messages.h
```diff
@@ -585,15 +605,19 @@ class BOTAN_UNSTABLE_API Certificate_13 final : public Handshake_Message {
...
       * Deserialize a Certificate message
       * @param buf the serialized message
       * @param policy the TLS policy
       * @param side is this a Connection_Side::Server or Connection_Side::Client certificate message
       */
-      Certificate_13(const std::vector<uint8_t>& buf, const Policy& policy, Connection_Side side);
+      Certificate_13(const std::vector<uint8_t>& buf,
+                     const Policy& policy,
+                     Connection_Side side,
+                     Certificate_Type cert_type);
```
lieser, line 620: param missing in docstring

src/lib/tls/tls_policy.cpp
```diff
@@ -585,6 +593,17 @@ void print_vec(std::ostream& o, const char* key, const std::vector<Group_Params>
    o << "\n";
 }
 
+void print_vec(std::ostream& o, const char* key, const std::vector<Certificate_Type>& types) {
```
lieser, line 596: Use `std::string_view` as type for key?

src/lib/tls/tls_policy.h
```diff
@@ -364,6 +365,28 @@ class BOTAN_PUBLIC_API(2, 0) Policy {
       */
       virtual bool request_client_certificate_authentication() const;
 
+      /**
+      * Returns a list of accepted certificate types for client authentication
```
lieser, line 369: Indention, same below

reneme:
Thanks for the reviews. I had a quick skim-through. Lot's of valuable remarks, I believe. I'm hoping to get back to this on Monday, but it might slip to Thursday.

src/lib/tls/tls_callbacks.h
```diff
@@ -171,6 +171,38 @@ class BOTAN_PUBLIC_API(2, 0) Callbacks {
...
+       * This function should throw an exception derived from
+       * std::exception with an informative what() result if the
+       * raw public key cannot be verified.
+       *
+       * @param raw_public_key specifies the raw public key to be used
+       *        for peer authentication
+       * @param usage what this cert chain is being used for
+       *        Usage_Type::TLS_SERVER_AUTH for server chains,
+       *        Usage_Type::TLS_CLIENT_AUTH for client chains,
+       *        Usage_Type::UNSPECIFIED for other uses
```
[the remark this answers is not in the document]
reneme: Fair.

src/lib/tls/tls_extensions.cpp
```diff
@@ -356,6 +363,108 @@ std::vector<uint8_t> Application_Layer_Protocol_Notification::serialize(Connecti
...
 
+std::string certificate_type_to_string(Certificate_Type type) {
+   switch(type) {
+      case Certificate_Type::X509:
+         return "X509";
+      case Certificate_Type::RawPublicKey:
+         return "RawPublicKey";
+   }
+
+   return "Unknown";
```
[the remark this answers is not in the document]
reneme: I'd rather not do that here. The certificate_type extension may come with other certificate types that we don't support. So we might be calling this for integers that we didn't register in the enum. For instance the proposed Merkle Tree Certificates would make use of this mechanism with another entry in the enum.

src/cli/tls_helpers.h
```diff
@@ -161,10 +172,18 @@ class Basic_Credentials_Manager : public Botan::Credentials_Manager {
          return {};
       }
 
+      std::shared_ptr<Botan::Public_Key> find_raw_public_key(const std::vector<std::string>& algos,
+                                                             std::string_view type,
+                                                             std::string_view hostname) override {
+         BOTAN_UNUSED(type, hostname);
```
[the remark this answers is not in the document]
reneme: Fair point. It was inconsistent before already, though. I'd rather keep the diff small here, to be honest.

src/lib/tls/credentials_manager.h
```diff
@@ -75,6 +75,23 @@ class BOTAN_PUBLIC_API(2, 0) Credentials_Manager {
          const std::string& type,
          const std::string& context);
 
+      /**
+      * Return a raw public key to be used for authentication or nullptr if no
```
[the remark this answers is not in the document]
reneme: This also bugs me all the time. For new classes I don't do it like that. Though, when extending existing code that formats it like that, I stay consistent with the surroundings. Perhaps, we could smooth this out with a `clang-format` rule independent of this PR.

src/lib/tls/credentials_manager.h
```diff
@@ -134,6 +151,18 @@ class BOTAN_PUBLIC_API(2, 0) Credentials_Manager {
...
+      * This function should either return nullptr or throw an exception if
+      * the key is unavailable.
+      *
+      * @return private key associated with this raw public key if we should
+      *         use it with this context. @p raw_public_key was returned by
+      *         find_raw_public_key()
+      */
+      virtual std::shared_ptr<Private_Key> private_key_for(const Public_Key& raw_public_key,
+                                                           const std::string& type,
+                                                           const std::string& context);
```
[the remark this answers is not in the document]
reneme: Went for `const std::string&`. Thanks.

src/lib/tls/tls12/tls_channel_impl_12.h
```diff
@@ -107,6 +107,14 @@ class Channel_Impl_12 : public Channel_Impl {
       */
       std::vector<X509_Certificate> peer_cert_chain() const override;
 
+      /**
+      * Note: Raw public key for authentication (RFC7250) is currently not
```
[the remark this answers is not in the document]
reneme: As mentioned before: I think, we should stay consistent with neighboring existing code.

src/lib/tls/tls13/tls_client_impl_13.cpp
```diff
@@ -586,6 +620,18 @@ std::vector<X509_Certificate> Client_Impl_13::peer_cert_chain() const {
    return {};
 }
 
+std::shared_ptr<const Public_Key> Client_Impl_13::peer_raw_public_key() const {
+   if(m_handshake_state.has_client_certificate_msg() && m_handshake_state.server_certificate().is_raw_public_key()) {
```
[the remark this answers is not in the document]
reneme: Yes! Very good catch. 😨

src/lib/tls/tls13/tls_client_impl_13.h
reneme: As above: local consistency weighs more IMO.

src/lib/tls/tls13/tls_server_impl_13.cpp
reneme: The order doesn't really matter: A resumed session does not have a "current" certificate. When resuming a session, the server won't authenticate again. And vice-versa: when we have a Certificate message, we are not in a resumed session.

I.e. the order doesn't really matter. Though I'll fix it for consistency, thanks.

src/lib/tls/tls_extensions.cpp
reneme: Certainly doesn't hurt, thanks.

src/lib/tls/tls_policy.cpp
reneme: Yeah, we could certainly do that. Though, there are more implementations of `print_vec` here, that I wanted to stay consistent with.

src/lib/tls/tls_messages.h
reneme: I tried, and it certainly would work. However, the method `Certificate_Entry::public_key()` is supposed to return the credential's public key for both cases (i.e. Certificate and RawPublicKey).

Consider we have a certificate: If we'd use a variant, we would need to call `subject_public_key()` on the `X509_Certificate` every time. This returns a `std::unique_ptr<Public_Key>` and therefore re-allocates and decodes the public key for every invocation.

By splitting the members, we avoid this and decode the certificate's public key exactly once (in the constructor of `Certificate_Entry`). Albeit, it is debatable whether this is worth the additional complexity of two members.

src/lib/tls/tls13/msg_certificate_13.cpp
```diff
@@ -141,60 +173,180 @@ void Certificate_13::setup_entries(std::vector<X509_Certificate> cert_chain,
...
+         credentials_manager.find_cert_chain(key_types,
+                                             to_algorithm_identifiers(cert_request.certificate_signature_schemes()),
+                                             cert_request.acceptable_CAs(),
+                                             op_type,
+                                             std::string(hostname)),
+         cert_request.extensions().get<Certificate_Status_Request>(),
+         callbacks);
+   } else if(cert_type == Certificate_Type::RawPublicKey) {
+      auto raw_public_key = credentials_manager.find_raw_public_key(key_types, op_type, hostname);
+
```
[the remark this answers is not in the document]
reneme: No there wasn't! Very important catch. Thanks.

reneme:
@FAlbertDev @lieser  I addressed your comments. Thanks a lot. You've both found very relevant issues.

lieser (approved):

FAlbertDev (approved):

PR #3777, opened 2023-10-27
Add an example for X.509 path validation

securitykernel:
Adds an example for simple X.509 path validation of an end entity certificate against a Root CA.

### DIFF
--- a/src/examples/x509_path.cpp
+++ b/src/examples/x509_path.cpp
@@ -1,13 +1,21 @@
+#include <botan/certstor_system.h>
 #include <botan/x509cert.h>
 #include <botan/x509path.h>
 
 int main() {
-   // Create a certificate store and add the trusted CA certificate
-   Botan::Certificate_Store_In_Memory store;
-   store.add_certificate(Botan::X509_Certificate("ca.crt"));
+   // Create a certificate store and add a locally trusted CA certificate
+   Botan::Certificate_Store_In_Memory customStore;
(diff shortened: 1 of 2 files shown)


src/examples/x509_path.cpp
```diff
@@ -0,0 +1,30 @@
...
+   Botan::Path_Validation_Restrictions restrictions;
+
+   // Optional: Specify usage type, compared against the key usage in endEntityCert
+   Botan::Usage_Type usage = Botan::Usage_Type::UNSPECIFIED;
+
+   // Optional: Specify hostname, if not empty, compared against the DNS name in endEntityCert
+   std::string hostname = "";
+
+   Botan::Path_Validation_Result validationResult =
+      Botan::x509_path_validate(end_entity, restrictions, store, hostname, usage);
```
reneme, line 22: Perhaps use the overload that takes a list of certificates for `end_entity` to showcase the path building capability as well. Also, there's another overload that takes a list of `Certificate_Store*`, with that you could showcase the `System_Certificate_Store` as well; basically validating against "anything the OS trusts AND this application specific root".

reneme (approved):
Kudos for working on the documentation!

src/examples/x509_path.cpp
securitykernel: Good point, I'll change it.

securitykernel: Done

randombit:
:+1:

PR #3782, opened 2023-10-27
Add EC_PrivateKey::check_key() to check private value of EC keys (#3648)

### LINKED ISSUE #3648: ECDSA private key check
bool ECDSA_PrivateKey::check_key(RandomNumberGenerator& rng, bool strong) const

Does not verify that private value is in the range of 1 <= k < curve order.

Suggested fix:
bool ECDSA_PrivateKey::check_key(RandomNumberGenerator& rng, bool strong) const {
     if(m_private_key < 1 || m_private_key >= m_domain_params.get_order()) {
      return false;
   }

   if(!public_point().on_the_curve()) {
      return false;
   }
....

lieser:
Add `EC_PrivateKey::check_key(...)` to check private EC keys. The check includes:

- Check that the private value is in the range [1, q-1] (q being the size of the group).
- Delegate checking the public values by calling `EC_PublicKey::check_key(...)`.

Also include a call to `EC_PrivateKey::check_key(...)` in the `check_key(...)` methods of `ECDSA_PrivateKey`, `ECGDSA_PrivateKey`, `ECKCDSA_PrivateKey` and `SM2_PrivateKey`.

Fixes #3648.

### DIFF
--- a/src/lib/pubkey/ecc_key/ecc_key.cpp
+++ b/src/lib/pubkey/ecc_key/ecc_key.cpp
@@ -158,6 +158,14 @@ EC_PrivateKey::EC_PrivateKey(const AlgorithmIdentifier& alg_id,
    }
 }
 
+bool EC_PrivateKey::check_key(RandomNumberGenerator& rng, bool strong) const {
+   if(m_private_key < 1 || m_private_key >= m_domain_params.get_order()) {
+      return false;
+   }
+
+   return EC_PublicKey::check_key(rng, strong);
+}
+
 const BigInt& EC_PublicKey::get_int_field(std::string_view field) const {
    if(field == "public_x") {
       BOTAN_ASSERT_NOMSG(this->public_point().is_affine());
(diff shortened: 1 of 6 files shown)


randombit (commented):
Change looks good but: should ECDH and GOST 34.10 also be updated?

lieser:
> should ECDH and GOST 34.10 also be updated?

The `ECDH_PrivateKey` and `GOST_3410_PrivateKey` do not override the `check_key(...)` method.
So they should automatically use the new `EC_PrivateKey::check_key(...)`.

randombit (approved):

guidovranken:
@randombit @lieser 

Technically SM2 private keys should be 1..order-2, not 1..order-1 like other curves. `SM2_PrivateKey::check_key` accepts order-1, unless `strong` is `true`. Not sure if this is worth addressing?

Some commentary on this here: https://github.com/libecc/libecc/blob/b9329e2826f4d622dbb9ffdd9316e98fda7a023f/src/sig/sm2.c#L27-L36

PR #3786, opened 2023-10-31
FIX: Non-KEX-algos may choose to not implement ::generate_another()

reneme:
Some algorithms (like HSS/LMS #3716) cannot even implement it at all, because their public key does not provide enough information about the private key's internal parameters.

### DIFF
--- a/src/lib/pubkey/pk_keys.h
+++ b/src/lib/pubkey/pk_keys.h
@@ -104,7 +104,9 @@ class BOTAN_PUBLIC_API(3, 0) Asymmetric_Key {
 
       /**
        * Generate another (cryptographically independent) key pair using the
-       * same algorithm parameters as this key.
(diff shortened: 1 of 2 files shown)


src/lib/pubkey/pk_keys.h
```diff
@@ -104,7 +104,9 @@ class BOTAN_PUBLIC_API(3, 0) Asymmetric_Key {
 
       /**
        * Generate another (cryptographically independent) key pair using the
-       * same algorithm parameters as this key.
+       * same algorithm parameters as this key. This is most useful for algorithms
+       * that support PublicKeyOperation::KeyAgreement to generate a fitting ephemeral
+       * key pair. For other key types it might throw NotImplemented.
```
randombit, line 109: `Not_Implemented`

randombit (approved):

PR #3787, opened 2023-10-31
examples target should require boost

securitykernel:
Fail if `examples` target should be build, but boost is not enabled.


reneme (approved):

randombit:
This doesn't really seem right to me. Just a single one of the examples requires Boost; you could make a similar argument that we should require Kyber, RSA, ... since at least one of the examples will fail to compile in such a configuration.

securitykernel:
In general I agree, we should improve the examples to not be all-or-nothing. I just stumbled upon this resulting in being simply an uncompilable configuration and I at least wanted to make it explicitly fail on configure instead, as a first improvement.

PR #3788, opened 2023-10-31
sqlite3 engine update proposal.

devnexen:


reneme (approved):
Looks good to me. This requires SQLite 3.37 from November 2021, which is the version shipped in Ubuntu 22.04.

PR #3791, opened 2023-11-01
FIX: ./botan tls_http_server does not compile for --with-boost --disable-modules=asio

reneme:
I moved the new `boost_compat.h` from the `boost` module to the `asio` TLS stream and renamed it to `asio_compat.h`. Hopefully, this better communicates its intent of feature-flagging the ASIO stream but not all boost-based implementations in the library.

This was sparked when I found that the `./botan tls_http_server` CLI command did not compile when configuring the build `--with-boost` and `--disable-modules=asio`. The header macros are somewhat misleading:

* `BOTAN_HAS_BOOST_ASIO` means that the build is configured `--with-boost` but
  (and therefore contains the meta-module `boost`)
* `BOTAN_HAS_TLS_ASIO_STREAM` means that the `asio` module (with the TLS stream) is available.


src/cli/tls_http_server.cpp
```diff
@@ -12,11 +12,11 @@
 
 #include "cli.h"
 
-#if defined(BOTAN_HAS_BOOST_ASIO)
-   #include <botan/boost_compat.h>
+#if defined(BOTAN_HAS_TLS_ASIO_STREAM)
+   #include <botan/asio_compat.h>
 #endif
```
reneme, line 17: This checked for the wrong module (the meta-module `boost` that also used to contain `boost_compat.h`). It is available when configured `--with-boost`. Downstream in this file we didn't check for `BOTAN_HAS_TLS_ASIO_STREAM` but assumed that, if `BOTAN_FOUND_COMPATIBLE_BOOST_VERSION` was set, that `asio_stream.h` would be available. This is not true when configured with `--disable-modules=asio`, though.

randombit (approved):
Makes sense, looks good to me

PR #3794, opened 2023-11-02
Improve examples by e.g. using create_or_throw() and fix missing include in stream_cipher.cpp

lieser:
Improve examples:
- Use `create_or_throw(...)` to avoid possible null dereferencing.
- Use '\n' for newline in examples

Also fix a missing include in `src/lib/stream/stream_cipher.cpp`.


reneme (approved):
Nice! That `::create()` factory is actually a bit of a footgun, IMO. So it's a great idea to not push people to use it in the examples. :+1:

Issue #3778, opened 2023-10-27
Nightly TLS-Anvil CI failure with new HTTP server


FAlbertDev:
It seems that TLS-Anvil is not (yet) happy with the new TLS HTTP server (see [failing CI](https://github.com/randombit/botan/actions/runs/6662530573/job/18107006051) ). At first glance, it is a problem with a clean TLS session shutdown. Either some wrong handling of the close notify TLS alert (which looks OK in Wireshark) or some issue with closing the socket. I'll look into it.

reneme:
After a first investigation we believe that the ASIO stream fails to send TLS alerts while handling the handshake. For instance, if the TLS implementation generates a `TLS_Exception(Alert::HandshakeFailure)`, it would close the socket immediately but not report the alert to the peer first.

This seems to be a long-standing bug in the ASIO stream.

reneme:
Another issue seems to be that we didn't close the connection gracefully in `./botan tls_http_server` after a protocol error had occured. Details [here](https://github.com/randombit/botan/pull/3795#issuecomment-1792589832). @FAlbertDev has run [another round of TLS-Anvil](https://github.com/FAlbertDev/botan/actions/runs/6746976008/job/18342090752) with now just one failing test. More next week!

PR #3795, opened 2023-11-02
FIX: Alert handling in ASIO stream and HTTP Server

reneme:
Turns out that #3778 seems to be caused by a long-standing bug in the ASIO stream wrapper. Consider a client sending a ClientHello that a server has to reject with a lack of compatible ciphersuites. Botan's TLS server would signal that with a `TLS_Exception(Alert::HandshakeFailure, "...")`.

The TLS stream wrapper (as the server) didn't properly handle that (as in: send a fatal alert message, terminate the connection and report a reasonable error code back to ASIO). Instead it just closed the connection immediately.
Likewise, when receiving a fatal alert message from a peer, the ASIO stream wrapper didn't report that back to the application as an error code but also simply terminated the connection.

This surfaced because the TLS Anvil nightly test is based on `./botan tls_http_server`, which is now based on the ASIO stream: #3763. Here we add proper handling for both cases and a suitable regression test.

Fortunately, the buggy behavior was arguable "okayish": The connection was hard-reset. Just, the application and the peer didn't get a proper error message. TLS-Anvil isn't satisfied with just that, though.

@FAlbertDev Let's verify tomorrow that this helps with #3778.

### DIFF
--- a/src/cli/tls_http_server.cpp
+++ b/src/cli/tls_http_server.cpp
@@ -226,6 +226,8 @@ net::awaitable<void> do_session(tcp_stream stream,
    auto callbacks = std::make_shared<TlsHttpCallbacks>();
    Botan::TLS::Stream<tcp_stream> tls_stream(std::move(stream), std::move(tls_ctx), callbacks);
 
+   std::exception_ptr protocol_exception;
+
    try {
       // Perform a TLS handshake with the peer
       co_await tls_stream.async_handshake(Botan::TLS::Connection_Side::Server);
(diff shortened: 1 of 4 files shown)


reneme:
Force pushed to explain it again for MSVC. It produced warnings for unreachable code that were false positives. 🤷‍♂️

FAlbertDev:
Nice! TLS-Anvil is not yet fully satisfied but many failing tests were fixed by these changes. Before we had 62 failing tests, now we only have 8. We'll look into these as well.

reneme:
https://github.com/randombit/botan/pull/3795/commits/9a21fe31598dec7afa3f0798fe2ba06db68e7da1 contains another fix. On application level (`./botan tls_http_server`) this time. We need to make sure that the server always calls `.async_shutdown()`, even if a TLS exception is thrown. This will flush outstanding data from the server's write buffer: e.g. a TLS alert message emitted by the stream during `async_read()`.

randombit (approved):
:+1:

PR #3796, opened 2023-11-03
Added Blake2s hash function

huven:
For this I used the reference implementation from RFC 7693, have tried to make as few changes to the code as possible. All lines from the RFC are marked as such.

I also kept the "key" parameter, although it is not used for digest calculation, to allow extending this into a MAC similar to Blake2b.


reneme:
Please mention Blake2s in the documentation (`doc/api_ref/hash.rst`). Also note that these are about to [mention the respective algorithm specs](https://github.com/randombit/botan/pull/3802) for each primitive.

huven:
> Please mention Blake2s in the documentation (`doc/api_ref/hash.rst`). Also note that these are about to [mention the respective algorithm specs](https://github.com/randombit/botan/pull/3802) for each primitive.

done!

huven:
Not sure how to deal with the failing check, seems related to some TLS check failing (due to another recently accepted commit) so not related to this PR.

Any help appreciated..

reneme (commented):
Sorry for the long waiting time. I marked a few nits here and there.

~~One architectural remark: I think we should consider implementing this algorithm in terms of [the `XOF` interface](https://github.com/randombit/botan/blob/b7c737ff816148fc05f634d2e3b25974bfd19452/src/lib/xof/xof.h). (Independently, the same holds for Blake2b, of course.)~~

~~Once we have a `BLAKE2s_XOF`, we could simply define `BLAKE2s` hash and (as future work) `BLAKE2s_MAC` in terms of the more generic XOF. Similarly, to what we did [for SHAKE earlier this year](https://github.com/randombit/botan/pull/3671).~~

Never mind the XOF idea. The implementation of the BLAKE2X variants is slightly different.

huven:
I have applied the requested changes

reneme (approved):

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -0,0 +1,208 @@
...
+                                  {14, 10, 4, 8, 9, 15, 13, 6, 1, 12, 0, 2, 11, 7, 5, 3},
+                                  {11, 8, 12, 0, 5, 2, 15, 13, 10, 14, 3, 6, 7, 1, 9, 4},
+                                  {7, 9, 3, 1, 13, 12, 11, 14, 2, 6, 5, 10, 4, 0, 15, 8},
+                                  {9, 0, 5, 7, 2, 4, 10, 15, 14, 1, 11, 12, 6, 8, 3, 13},
+                                  {2, 12, 6, 10, 0, 11, 8, 3, 4, 13, 7, 5, 15, 14, 1, 9},
+                                  {12, 5, 1, 15, 14, 13, 4, 10, 0, 7, 6, 3, 9, 2, 8, 11},
+                                  {13, 11, 7, 14, 12, 1, 3, 9, 5, 0, 15, 4, 8, 6, 2, 10},
+                                  {6, 15, 14, 9, 11, 3, 0, 8, 12, 2, 13, 7, 1, 4, 10, 5},
+                                  {10, 2, 8, 4, 7, 6, 1, 5, 15, 11, 9, 14, 3, 12, 13, 0}};
+   int i;
```
randombit, line 52: Instead of global `i` use an index scoped to the loops. Also use `size_t` for loop iteration variables of this kind unless the signedness properties are actually used.

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -0,0 +1,208 @@
...
+   }
+
+   v[12] ^= ctx.t[0];  // low 32 bits of offset
+   v[13] ^= ctx.t[1];  // high 32 bits
+   if(last) {          // last block flag set ?
+      v[14] = ~v[14];
+   }
+   for(i = 0; i < 16; i++) {  // get little-endian words
+      m[i] = load_le<uint32_t>(&ctx.b[4 * i], 0);
+   }
```
randombit, line 67: There is a variant of `load_le` that writes to an array, which is more efficient since it boils down to just a `memcpy` on little endian platforms.

src/lib/hash/blake2s/blake2s.h
```diff
@@ -0,0 +1,57 @@
...
+
+// RFC7693 - copy begin
+
+// state context
+typedef struct {
+      uint8_t b[64];  // input buffer
+      uint32_t h[8];  // chained state
+      uint32_t t[2];  // total number of bytes
+      uint8_t c;      // pointer for b[]
+      size_t outlen;  // digest size
```
randombit, line 21: Above should just be moved directly as member variables of `BLAKE2s` class

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+   while(ctx->c < 64) {  // fill up with zeros
+      ctx->b[ctx->c++] = 0;
+   }
+   blake2s_compress(*ctx, true);  // final block flag = 1
+
+   // little endian convert and store
+   for(i = 0; i < ctx->outlen; i++) {
+      out[i] = (ctx->h[i >> 2] >> (8 * (i & 3))) & 0xFF;
```
randombit, line 160: This should use `store_le` utilities

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -0,0 +1,208 @@
...
+   blake2s_init(&m_ctx, m_ctx.outlen, nullptr, 0);
+}
+
+void BLAKE2s::add_data(std::span<const uint8_t> in) {
+   blake2s_update(&m_ctx, in.data(), in.size());
+}
+
+void BLAKE2s::final_result(std::span<uint8_t> out) {
+   blake2s_final(&m_ctx, out.data());
+   blake2s_init(&m_ctx, m_ctx.outlen, nullptr, 0);
```
randombit, line 185: This could just be a call to `clear`

randombit (commented):
While I do understand your reasoning for taking the original code from the RFC, following that structure isn't really that great for us as a library. We then are faced with the following options in the future

* Continue to keep them as a RFC reference, and never use helpers that we work on over time (for memory accesses, buffer management, etc)
* Adapt the RFC based code over time, so diverging from it while still claiming to be a copy/reference.
* Just don't do that and avoid the issue entirely :)

I would suggest the third. Specifically I'd request removing the inner ctx struct (the class `this` *is* the context struct!) and also the various `blake2s_` functions, and the references to the RFC beyond "see this RFC for a description of BLAKE2s" and/or a mention that the code is originally based on that but without claiming to be identical.

huven:
All requested changes have been made.

randombit (approved):
LGTM thanks!

PR #3799, opened 2023-11-04
New Example: Minimal HTTPS client

reneme:
This is an attempt to implement a minimal example of a working HTTPS client using ASIO, Beast and Botan. I think it is making a relevant point, that one can write a basic HTTPS client using Botan's TLS in about 30 lines of C++20.

To minimize the amount of Botan-related boilerplate I added a few convenience constructions to the library. I'm somewhat ambivalent about actually having them in the library, hence the "draft pull request". I'm very much open for feedback about this.

In detail, I added:

* `Default_Credentials_Manager` that uses `System_Certificate_Store` as the default trust store (if available)
* a convenience constructor for `TLS::Stream<>` that uses defaults for all aspects of the stream's `Context` object, namely:
  * `AutoSeeded_RNG`
  * `Default_Credentials_Manager`
  * `TLS::Session_Manager_In_Memory` and
  * `TLS::Default_Policy`
* Some deduction guides for `TLS::Stream<>` to make construction less verbose
  (those, I think, we should add in any case)


reneme:
Mhh, the boost version on CI doesn't seem to be recent enough for this example. I'll look into that later.

src/lib/tls/asio/asio_context.h
```diff
@@ -47,6 +52,20 @@ class Context {
...
       using Verify_Callback = detail::fn_signature_helper<decltype(&Callbacks::tls_verify_cert_chain)>::type;
 
+   #if defined(BOTAN_HAS_AUTO_SEEDING_RNG)
+      /**
+       * @brief Construct a TLS stream context with typical defaults
+       *
+       * @param server_info  Basic information about the host to connect to (SNI)
+       */
+      Context(Server_Information server_info = Server_Information()) :
+            m_credentials_manager(std::make_shared<Default_Credentials_Manager>()),
```
randombit, line 62: This `Default_Credentials_Manager` seems odd to me tbh. If the system certs are not available, it seems like nothing will work anyway. Why not just require the system certs for the example?

reneme: Currently, the `Credentials_Manager` base class is mostly an interface without useful default implementations. Therefore, users are forced to provide their own `Credentials_Manager` subclass, even if they just want to use system certificates, anyway.

In my opinion that adds quite a bit of friction. Especially, for people that want to try the ASIO stream as replacement for the OpenSSL stream shipping with Boost. Hence, this is an attempt to have a functional default for the `TLS::Context` object with minimal parameters.

Admittedly, the `Default_Credentials_Manager` was just the shortest path to get this. Perhaps, it could be more generic and (optionally) accept user-provided roots. If no root is provided, it could default to the system certs (when available). Like so:

```C++
/**
 * @param server_info  the hostname you're trying to connect to
 * @param root_certs   the root certificates you're willing to trust
 *                     (defaults to the operating system's certificates)
 */
Context(Server_Information server_info = Server_Information(),
        std::vector<X509_Certificate> root_certs = {}) :
    m_credentials_manager(std::make_shared<Default_Credentials_Manager>(std::move(root_certs))),
    [...]
```

reneme: Note: I rebased this onto #3901 which adds the very same example but without the convenience additions in this PR.

reneme: I moved the `Default_Credentials_Manager` out of the public interface and demoted it to an implementation detail. See here: https://github.com/randombit/botan/pull/3799#issuecomment-2072156100.

randombit (approved):

PR #3805, opened 2023-11-12
Link examples from getting started page

securitykernel:
During working on an OpenSSL -> Botan migration guide for the docs, which is heavily based on example code, I noticed that we have no listing of examples other than referencing the CLI in the getting started guide. This PR adds such a list with clickable references to the relevant pages in the docs, which the migration guide can simply refer to later.

I played a bit with [automatically labeling sections](https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#automatically-label-sections) and it worked quite well. During experimenting I removed duplicate references already (having multiple subsections `Code Example` in the same doc; fixed in this PR). By not having to explicitly create labels for sections we want to refer to, IMHO we would make it easier to reference inside the docs in general, so we should think about it.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -17,8 +17,31 @@ source may not be required on your system.
 Examples
 ----------
 
-Some examples of usage are included in this documentation. However a better
-source for example code is in the implementation of the
+Examples of usage are included in this documentation, some of which
+are listed below:
+
+* :ref:`Block Ciphers <block_cipher_example>`
+* :ref:`Cipher Modes <cipher_modes_example>`
+* :ref:`Hash Functions <hash_example>`
+* :ref:`KDFs <key_derivation_function_example>`
+* :ref:`MACs <mac_example>`
+* :ref:`PBKDFs <pbkdf_example>`
+* :ref:`Key Agreement <ecdh_example>`
+* :ref:`ECDSA <ecdsa_example>`
+* :ref:`Kyber <kyber_example>`
+* :ref:`RSA <rsa_example>`
+* :ref:`XMSS <xmss_example>`
+* :ref:`Stream Ciphers <stream_ciphers_example>`
+* :ref:`TLS Client <tls_client_example>`
+* :ref:`TLS Client (PQC/hybrid) <tls_hybrid_client_example>`
+* :ref:`HTTPS Client <https_client_example>`
+* :ref:`TLS Server <tls_server_example>`
+* :ref:`X.509 <x509_certificates_example>`
+
+You'll find additional examples of usage in the
+`src/examples <https://github.com/randombit/botan/tree/master/src/examples>`_ directory.
+
+An additional source for example code is in the implementation of the
 `command line interface <https://github.com/randombit/botan/tree/master/src/cli>`_,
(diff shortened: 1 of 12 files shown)


doc/index.rst
```diff
@@ -17,8 +17,31 @@ source may not be required on your system.
...
+* :ref:`TLS Client <tls_client_example>`
+* :ref:`TLS Client (PQC/hybrid) <tls_hybrid_client_example>`
+* :ref:`HTTPS Client <https_client_example>`
+* :ref:`TLS Server <tls_server_example>`
+* :ref:`X.509 <x509_certificates_example>`
+
+You'll find additional examples of usage in the
+`src/examples <https://github.com/randombit/botan/tree/master/src/examples>`_ directory.
+
+However a better source for example code is in the implementation of the
```
reneme, line 44: Not sure the CLI is a "better" source. Maybe just "another"? The examples are actually quite complete, I think. Except maybe the baremetal TLS examples, which are mostly just starting points (w/o socket code for brevity).

src/configs/sphinx/conf.py
```diff
@@ -188,7 +188,7 @@ def parse_version_file(version_path):
 # Grouping the document tree into LaTeX files. List of tuples
 # (source start file, target name, title, author, documentclass [howto/manual]).
 
-authors = u'Jack Lloyd \\and Daniel Neus \\and Ren\u00e9 Korthaus \\and Juraj Somorovsky \\and Tobias Niemann'
```
reneme, line 191: :+1:

reneme (approved):
Nice!

doc/index.rst
securitykernel: Yeah, probably that was true when this section was written, but as more and more examples were added over time, I also favor calling the CLI "another" source.

doc/index.rst
```diff
@@ -17,8 +17,31 @@ source may not be required on your system.
...
+* :ref:`ECDSA <ecdsa_example>`
+* :ref:`Kyber <kyber_example>`
+* :ref:`RSA <rsa_example>`
+* :ref:`XMSS <xmss_example>`
+* :ref:`Stream Ciphers <stream_ciphers_example>`
+* :ref:`TLS Client <tls_client_example>`
+* :ref:`TLS Client (PQC/hybrid) <tls_hybrid_client_example>`
+* :ref:`HTTPS Client <https_client_example>`
+* :ref:`TLS Server <tls_server_example>`
+* :ref:`X.509 <x509_certificates_example>`
```
securitykernel, line 39: I wasn't sure if mixing categories, e.g., hash functions, with algorithms, e.g. RSA, here is a good idea. I tend to believe that people rather look for examples on concrete algorithms, and then we should rename everything here to algorithms.

doc/index.rst
securitykernel: Changed in a2e862613c8a91c7decea11429f515fd1949eb29

doc/index.rst
reneme: Yeah, you have a point there, I guess. But then again: I think most people should be able to figure that out regardless. Having the list is certainly a nice improvement.

PR #3815, opened 2023-11-27
Add an OpenSSL 1.1 migration guide

securitykernel:
Adds a guide for migrating applications from OpenSSL 1.1 to Botan 3.x. Since OpenSSL 1.1 has gone EOL in September, 2023, users are forced to migrate to OpenSSL 3.x. The idea of writing this guide is to catch these users and offer some guidance for migrating to Botan instead.

The guide is heavily driven by comparing OpenSSL and Botan example code for certain use cases, and then elaborates on the specifics in Botan, as well as provides deep links into the documentation where possible. The Botan examples were mostly reused from the existing examples. The corresponding OpenSSL 1.1 example code was created using GitHub CoPilot, so it may not be 100% accurate, but IMHO for the purpose of comparison is sufficient.

This PR also enables automatic labeling of sections, so I did not need to add any new jump marks for the deep links.

### DIFF
--- a/doc/openssl_migration_guide.rst
+++ b/doc/openssl_migration_guide.rst
@@ -1,0 +1,464 @@
+OpenSSL 1.1 to Botan 3.x Migration
+==================================
+
+This aims to be a rough guide for migrating applications from OpenSSL 1.1 to Botan 3.x.
+
+This guide attempts to be, but is not, complete. If you run into a problem while
+migrating code that does not seem to be described here, please open an issue on
+`GitHub <https://github.com/randombit/botan/issues>`_.
+
+.. note::
+   The OpenSSL code snippets in this guide may not be 100% correct. They are
+   intended to show the differences in using OpenSSL's and Botan's APIs
+   rather to be a complete and correct example.
+
+General Remarks
+----------------
+
+* Botan is a C++ library, whereas OpenSSL is a C library
+* Botan also provides a :doc:`C API <api_ref/ffi>` for most of its functionality,
+  but it is not a 1:1 mapping of the C++ API
+* With OpenSSL's API, there are sometimes multiple ways to achieve the same result,
+  whereas Botan's API is more consistent
+* OpenSSL's API is mostly underdocumented, whereas Botan targets 100% Doxygen
+  coverage for all public API
+* It is often hard to find example code for OpenSSL, whereas Botan provides
+  many :ref:`examples <index:examples>` and lots of
+  `test code <https://github.com/randombit/botan/tree/master/src/tests>`_.
+
+X.509
+------
+
+Consider the following application code that uses OpenSSL to verify a
+certificate chain consisting of an end-entity certificate, two untrusted intermediate
+certificates, and a trusted root certificate.
+
+.. code-block:: cpp
+
+    #include <openssl/x509.h>
+    #include <openssl/pem.h>
+    #include <openssl/ssl.h>
+
+    int main() {
+        // Create a new X.509 store
+        X509_STORE *store = X509_STORE_new();
+
+        // Load the root certificate
+        FILE* rootCertFileHandle = fopen("root.crt", "r");
+        X509* rootCert = PEM_read_X509(rootCertFileHandle, NULL, NULL, NULL);
+        X509_STORE_add_cert(store, rootCert);
+        fclose(rootCertFileHandle);
+
+        // Create a new X.509 store context
+        X509_STORE_CTX *ctx = X509_STORE_CTX_new();
+        X509_STORE_CTX_init(ctx, store, NULL, NULL);
+
+        // Load the intermediate certificates
+        FILE* intermediateCertFileHandle1 = fopen("int2.crt", "r");
+        FILE* intermediateCertFileHandle2 = fopen("int1.crt", "r");
+        X509* intermediateCert1 = PEM_read_X509(intermediateCertFileHandle1, NULL, NULL, NULL);
+        X509* intermediateCert2 = PEM_read_X509(intermediateCertFileHandle2, NULL, NULL, NULL);
+        X509_STORE_CTX_trusted_stack(ctx, sk_X509_new_null());
+        sk_X509_push(X509_STORE_CTX_get0_untrusted(ctx), intermediateCert1);
+        sk_X509_push(X509_STORE_CTX_get0_untrusted(ctx), intermediateCert2);
+        fclose(intermediateCertFileHandle1);
+        fclose(intermediateCertFileHandle2);
+
+        // Load the end-entity certificate
+        FILE* endEntityCertFileHandle = fopen("ee.crt", "r");
+        X509* endEntityCert = PEM_read_X509(endEntityCertFileHandle, NULL, NULL, NULL);
+        X509_STORE_CTX_set_cert(ctx, endEntityCert);
+        fclose(endEntityCertFileHandle);
+
+        // Verify the certificate chain
+        int result = X509_verify_cert(ctx);
+        if(result != 1) {
+            // Verification failed
+            X509_STORE_CTX_free(ctx);
+            X509_STORE_free(store);
+            return -1;
+        }
+
+        // Verification succeeded
+        X509_STORE_CTX_free(ctx);
+        X509_STORE_free(store);
+        return 0;
+    }
+
+First, we create a new ``X509_STORE`` object and add the trusted root certificate.
+Then we add the intermediate certificates to the untrusted certificate stack.
+Finally, we set the end-entity certificate and call ``X509_verify_cert()`` to verify
+the whole certificate chain.
+
+Here is the equivalent C++ code using Botan:
+
+.. literalinclude:: /../src/examples/x509_path.cpp
+   :language: cpp
+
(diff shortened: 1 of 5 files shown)


doc/openssl_migration_guide.rst
```diff
@@ -0,0 +1,389 @@
...
+
+This guide attempts to be, but is not, complete. If you run into a problem while
+migrating code that does not seem to be described here, please open an issue on
+`GitHub <https://github.com/randombit/botan/issues>`_.
+
+General Remarks
+----------------
+
+* Botan is a C++ library, whereas OpenSSL is a C library
+* Botan also provides a :doc:`C API <api_ref/ffi>` for most of its functionality,
```
reneme, line 14: Is there a difference between the `:doc:` and `:ref:` markups?

doc/openssl_migration_guide.rst
```diff
@@ -0,0 +1,389 @@
...
+
+    #include <botan/auto_rng.h>
+    #include <botan/hex.h>
+    #include <iostream>
+
+    int main() {
+        Botan::AutoSeeded_RNG rng;
+
+        Botan::secure_vector<uint8_t> buffer(16); // Buffer to hold 16 random bytes
+        rng.randomize(buffer.data(), buffer.size());
```
reneme, line 146:
```suggestion
        const Botan::secure_vector<uint8_t> buffer = rng.random_vec(16);
```

or maybe even:

```suggestion
        const auto buffer = rng.random_vec(16);
```

... but, I think there's some value to spelling out the type for this example.

Alternatively, one can use the more modern `std::span<>` based API and avoid the raw-pointer handling:

```suggestion
        Botan::secure_vector<uint8_t> buffer(16); // Buffer to hold 16 random bytes
        rng.randomize(buffer);
```

doc/openssl_migration_guide.rst
```diff
@@ -0,0 +1,389 @@
...
+
+        // Print the random bytes in hexadecimal format
+        std::cout << Botan::hex_encode(buffer) << std::endl;
+
+        return 0;
+    }
+
+This snipped uses the ``AutoSeeded_RNG`` class to generate the 16 random bytes. Botan provides different
```
reneme, line 154:
```suggestion
This snippet uses the ``AutoSeeded_RNG`` class to generate the 16 random bytes. Botan provides different
```

doc/openssl_migration_guide.rst
```diff
@@ -0,0 +1,389 @@
...
+:ref:`api_ref/rng:random number generators`, including system-specific as well as system-independent
+software and hardware-based generators, and a comprehensive interface for implementing
+your own random number generator, if required. ``AutoSeeded_RNG`` is the recommended random number
+generator for most applications.
+The ``randomize()`` function takes a pointer to the buffer and the size of the buffer as arguments.
```
reneme, line 159: Depending on the changed example code above this sentence needs an adaption.

doc/openssl_migration_guide.rst
```diff
@@ -0,0 +1,389 @@
...
+are hex-encoded and converted to binary before encryption.
+The ``AES_set_decrypt_key()`` and ``AES_decrypt()`` functions are used to decrypt the ciphertext.
+
+Here is the equivalent C++ code using Botan:
+
+.. literalinclude:: /../src/examples/aes.cpp
+   :language: cpp
+
+This example uses the ``AES_256`` cipher to encrypt a 128-bit plaintext block with a 256-bit
+key using AES. The key and plaintext block are hex-encoded and converted to binary before encryption.
```
reneme, line 224:
```suggestion
key using AES. The key and plaintext block are hex-decoded and converted to binary before encryption.
```

... no?

doc/openssl_migration_guide.rst
```diff
@@ -0,0 +1,389 @@
...
+            printf("%02X", decrypted[i]);
+        }
+        printf("\n");
+
+        return 0;
+    }
+
+This example uses the ``AES_set_encrypt_key()`` and ``AES_encrypt()`` functions to encrypt
+a 128-bit plaintext block with a 256-bit key using AES. The key and plaintext block
+are hex-encoded and converted to binary before encryption.
```
reneme, line 215:
```suggestion
are hex-decoded and converted to binary before encryption.
```

doc/openssl_migration_guide.rst
```diff
@@ -0,0 +1,389 @@
...
+   :language: cpp
+
+This example uses the ``AES_256`` cipher to encrypt a 128-bit plaintext block with a 256-bit
+key using AES. The key and plaintext block are hex-encoded and converted to binary before encryption.
+``set_key()`` is used to set the key and ``encrypt()`` is used to encrypt the plaintext block.
+Then, the ``AES_256`` object is even reused to encrypt a second plaintext with a different key.
+
+To learn more about the ``BlockCipher`` and ``CipherMode`` interfaces, including a list of all
+available block ciphers and cipher modes, see the :ref:`api_ref/block_cipher:block ciphers` and
+:ref:`api_ref/cipher_modes:cipher modes` handbook sections.
```
reneme, line 230: I think we should rather showcase AES with a safe mode of operation (eg. CBC with padding -- we have an example for that). Maybe even better: an AEAD. Or at least note more prominently, that bare-metal usage of AES is not a very typical use case for end users.

doc/openssl_migration_guide.rst
```diff
@@ -0,0 +1,389 @@
...
+        if(privKeyFile == NULL) {
+            fprintf(stderr, "Error opening private key file.\n");
+            return 1;
+        }
+        EVP_PKEY* privKey = PEM_read_PrivateKey(privKeyFile, NULL, NULL, NULL);
+        fclose(privKeyFile);
+
+        // String to encrypt
+        unsigned char* plaintext = (unsigned char*)"Your great-grandfather gave this watch to your granddad for good luck. Unfortunately, Dane's luck wasn't as good as his old man's.";
+        size_t plaintext_len = strlen((char*)plaintext);
```
reneme, line 267: Do we actually need those casts?

doc/openssl_migration_guide.rst
```diff
@@ -0,0 +1,389 @@
...
+.. literalinclude:: /../src/examples/rsa_encrypt.cpp
+   :language: cpp
+
+This example uses the ``PK_Encryptor_EME`` and ``PK_Decryptor_EME`` classes to
+:ref:`encrypt and decrypt <api_ref/pubkey:public key encryption/decryption>`.
+a message using :ref:`api_ref/pubkey:rsa`. The public and private keys are
+:ref:`loaded from files <api_ref/pubkey:serializing private keys using pkcs #8>`.
+The padding scheme and :doc:`hash function <api_ref/hash>` are passed as a string parameter.
+
+Consider the following application code to sign some data with ECDSA using OpenSSL.
```
reneme, line 325: Let's have another headline here. (Split asym encryption and signing into two sections)

doc/openssl_migration_guide.rst
```diff
@@ -0,0 +1,389 @@
...
+        // Sign the hash
+        ECDSA_SIG* sig = ECDSA_do_sign(hash, SHA256_DIGEST_LENGTH, ec_key);
+        if(sig == NULL) {
+            std::cerr << "Error signing: " << ERR_error_string(ERR_get_error(), NULL) << "\n";
+            return 1;
+        }
+
+        // Print the signature
+        const BIGNUM* r;
+        const BIGNUM* s;
```
reneme, line 360: Nit, but: Do we need to free those?

doc/openssl_migration_guide.rst
```diff
@@ -0,0 +1,389 @@
...
+.. code-block:: cpp
+
+    #include <openssl/ecdsa.h>
+    #include <openssl/pem.h>
+    #include <openssl/sha.h>
+    #include <iostream>
+
+    int main() {
+        // Load private key
+        FILE* privKeyFile = fopen("private.pem", "r");
```
reneme, line 336: The Botan example is generating a private key. Maybe we want to do that here as well, to make it more "equivalenter" 😁.

doc/openssl_migration_guide.rst
securitykernel: With `:doc:` you reference whole documents, with `:ref:` you reference sections. See https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#cross-referencing-using-roles.

doc/openssl_migration_guide.rst
securitykernel: Removed them

doc/openssl_migration_guide.rst
securitykernel: OK

doc/openssl_migration_guide.rst
securitykernel: Good point, changed it.

doc/openssl_migration_guide.rst
securitykernel: Good point, changed it to use the AES CBC example.

securitykernel:
Covered everything I think. I also added a note that the OpenSSL code snippets in this guide may not be 100% correct, but are more illustrative examples to showcase OpenSSL's vs Botan's API.

reneme (approved):

PR #3817, opened 2023-11-30
session manager in memory, addressing TODO.

devnexen:
does not seem to bring down performance.


reneme (commented):
Given that iOS and Android were updated a while ago, perhaps we might try and get our feet wet with ranges here. Also that might actually improve readability.

devnexen:
ok giving a try..

randombit:
Blocked on toolchain updates, could be revisited for Botan4 when that happens. Closing.

PR #3819, opened 2023-12-02
Use furo Sphinx theme

### LINKED ISSUE #3818: Proposal: Use furo theme for Sphinx docs
We currently use the [readthedocs](https://sphinx-rtd-theme.readthedocs.io/en/stable/index.html) theme for publishing documentation on https://botan.randombit.net with Sphinx.

I recently discovered the [furo](https://github.com/pradyunsg/furo) theme for Sphinx, which is [used](https://github.com/pradyunsg/furo#used-by) by pip, python-gitlab, urllib3 and more. IMHO it looks much cleaner than readthedocs, and optionally allows the user to switch to dark mode. Below are some screenshots for comparison. I also uploaded a full recent Botan docs made with furo to https://docs.botan-crypto.org/furo/ to browse.

## Start page

image
image

## Building The Library

image
image

## Random Number Generators

image
image

## TLS Code Example

image
image

What do you think?

securitykernel:
Use [furo](https://github.com/pradyunsg/furo) Sphinx theme instead of readthedocs, which looks much cleaner and features a dark mode.

* Use furo theme by default
* Add an edit button to each document that allows visitors to jump to GH for improving the docs
* Don't hardcode `pygments_style` for code blocks, but let the theme figure out which style is best

Closes #3818

### DIFF
--- a/src/configs/sphinx/conf.py
+++ b/src/configs/sphinx/conf.py
@@ -88,7 +88,7 @@ add_function_parentheses = False
 highlight_language = 'cpp'
 
 # The name of the Pygments (syntax highlighting) style to use.
-pygments_style = 'sphinx'
+#pygments_style = 'sphinx'
 
 # A list of ignored prefixes for module index sorting.
 #modindex_common_prefix = []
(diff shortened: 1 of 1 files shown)


reneme (approved):
On that note: A while ago we found that [we don't build the documentation in CI anymore](https://github.com/randombit/botan/pull/3464#issuecomment-1494490616). Maybe we should re-enable the `docs` build target and make sure the theme package is installed.

securitykernel:
> On that note: A while ago we found that [we don't build the documentation in CI anymore](https://github.com/randombit/botan/pull/3464#issuecomment-1494490616). Maybe we should re-enable the `docs` build target and make sure the theme package is installed.

Good point! Let's check this next week when we meet.

PR #3829, opened 2023-12-11
Add sig algos and key agreements to bench.py

securitykernel:
Extends `bench.py` in order to be ready to run comparative benchmarks against OpenSSL to be published.

* From the already supported algorithms, remove MD5 and RIPEMD and add SHA-3 instead
* Add signature algorithms, in particular RSA and ECDSA (more can be added easily)
* Add key agreements, in particular DH and ECDH (more can be added easily)

Tested successfully with OpenSSL 3.0 and 3.2. OpenSSL seems to have changed their machine-readable output format in between these versions, which forced us to make the regexs more flexible.

Also changed the DH groups used in `botan speed dh` from `modp` to `ffdhe`, because the latter is used by `openssl speed`. Had to remove some key sizes not supported by the FFDHE groups.

`bench.py` now also supports optionally specifying one or more algorithms to benchmark instead of benchmarking all algorithms from the internal lists:

```sh
src/scripts/bench.py # still the default: benchmark all algorithms
src/scripts/bench.py RSA # NEW: benchmark only RSA
src/scripts/bench.py RSA ECDSA AES-128/GCM # NEW: benchmark RSA, ECDSA and AES-128/GCM
```

### DIFF
--- a/src/cli/speed.cpp
+++ b/src/cli/speed.cpp
@@ -1843,8 +1843,8 @@ class Speed final : public Command {
 
 #if defined(BOTAN_HAS_DIFFIE_HELLMAN)
       void bench_dh(const std::string& provider, std::chrono::milliseconds msec) {
-         for(size_t bits : {1024, 1536, 2048, 3072, 4096, 6144, 8192}) {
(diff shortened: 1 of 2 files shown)


randombit (approved):
Thanks

PR #3831, opened 2023-12-12
Update docstring of draft-irtf-cfrg-hash-to-curve

### LINKED ISSUE #3806: Update draft-irtf-cfrg-hash-to-curve to RFC 9380
[draft-irtf-cfrg-hash-to-curve](https://datatracker.ietf.org/doc/rfc9380/) was added in 3.0.0 and was now finalized as RFC 9380. We should review the differences and update the implementation accordingly. Guessing from 8bf306dbff69188f23ae1b6932ba392e4e04e814 it was likely implemented against draft-irtf-cfrg-hash-to-curve-09 or draft-irtf-cfrg-hash-to-curve-10, which is 8 or 7 revisions behind RFC. Luckily, IETF's data tracker provides a good diff viewer.

securitykernel:
Fixes #3806


securitykernel:
clang builds seem to be broken again, opened #3832.

securitykernel:
Rebased after #3833 was merged.

reneme (approved):

PR #3833, opened 2023-12-13
Update GitHub Actions workaround / Fix emcc build

reneme:
... as suggested in https://github.com/actions/runner-images/issues/8659#issuecomment-1853177960

The `brew` step in the "emscripten" build started failing as well. Trying to fix it while we're on it.


securitykernel (approved):

securitykernel:
Thanks!

PR #3838, opened 2023-12-14
Build docs in CI and make it fail on warnings or errors

### LINKED ISSUE #3654: CI doesn't fail on Sphinx formatting errors
Some formatting errors slipped in (fixed in 2c748bf9a5); this should never have passed CI.

lieser:
Fixes #3654.

Note that the missing `--keep-going` for `src/scripts/website.py` is deliberate.
`--keep-going` is nice for the CI because we see all warnings/errors, but it means Sphinx kind of still successfully builds the document. Making it easier to ignore/miss the error.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -214,6 +214,9 @@ jobs:
           - target: bsi
             compiler: gcc
             host_os: ubuntu-22.04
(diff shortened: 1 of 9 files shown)


reneme (approved):
Thank you!

PR #3839, opened 2023-12-14
FIX: flatten boost buffer sequence for TLS

reneme:
This adapts the ASIO Stream wrapper to flatten incoming buffer sequences (if needed) into a single buffer. Before, we simply iterated the buffer sequence and called `TLS::Channel::send()` for each buffer. Though, that can be sub optimal, because it may create many tiny TLS records.

For instance, Boost Beast renders its HTTP response into many small buffers that may contain just a single line of HTTP headers or perhaps even less. As a result, `./botan tls_http_server` generated several tiny TLS records to transfer the HTTP response. It turns out that `tlsfuzzer` chokes on this, as it expects the HTTP response to be shipped in a single TLS record.

It goes without saying that this data copy isn't optimal either. At one point we should look into extending `TLS::Channel` to be able to deal with scattered data buffers.

### DIFF
--- a/src/lib/tls/asio/asio_stream.h
+++ b/src/lib/tls/asio/asio_stream.h
@@ -848,19 +848,30 @@ class Stream {
        */
       template <typename ConstBufferSequence>
       void tls_encrypt(const ConstBufferSequence& buffers, boost::system::error_code& ec) {
+         // TODO: Once the TLS::Channel can deal with buffer sequences we can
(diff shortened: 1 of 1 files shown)


FAlbertDev (approved):

randombit (approved):
Oof, yeah it would be good to provide something to help with this in the base TLS channel interface.

PR #3841, opened 2023-12-15
Fix some doxygen warnings

lieser:


lieser:
Locally the still open warnings disappear with the newer doxygen 1.9.8 (and new warnings about `CLASS_DIAGRAMS` and `DOT_TRANSPARENT` appear).

A few open questions/todos:

- Do we want to use a newer doxygen version in the CI or try to fix the remaining warnings for the older version too?
- Do we want to change the requires to make doxygen happy or wait until https://github.com/doxygen/doxygen/pull/10486 is merged and released?
- If we are able to make the CI green, do we want to fail doxygen if warnings appear?
- Do we want to just delete the obsolete config options, or let doxygen make a complete update of the old configuration file?

reneme:
Nice work! I think there's quite some value in a warning-free Doxygen build. So my vote is on failing the build if new warnings show up. Caveat: I'm assuming there's a mechanism to selectively disable warnings when they cannot be fixed (or if we want to opt not to fix them for whatever reason).

Do we depend on the bleeding edge Doxygen, even after shuffling the `requires` clauses around? If not, I'd stay with what we have and avoid the extra complication of installing a newer version. I don't have a strong opinion where to put the `requires`.

lieser:
Suppressing specific warnings without removing the code from doxygen processing seems not possible.

reneme (approved):

PR #3845, opened 2023-12-18
Fix CLI tests under windows and enabled them in the CI

### LINKED ISSUE #1198: CLI tests fail on Windows
I don't understand the problem, and debugging via CI logging sucks, so I'm just going to skip running the tests on Windows for now. Example run:

https://ci.appveyor.com/project/randombit/botan/build/1.0.2958/job/bmhy8ys9rf09m9ij

```
Running 'python .\src/scripts/cli_tests.py .\botan-cli.exe':
FF.......FFF.....F.....F.......F...F..F...FFF...FFFF..FFFF..FF...FF.F.FFFFFFFFFF.F...F..F...FFF...FFFF..FFFF..FF......FFFFFFFFFFFFF.F.F.F.FFFFF.F.F.F.F.F.F.F.F.F.FFFFFFFFFFFFFF........F.FFF..F...FFFFF....F..FF.FF........F.FF......F.FFFF.....F......FF....FF.FF...FFFFFFFFFFF...FFF....FF...FFF.FFFFF...FFF..F.FF...FFF..FFFF.F.F.F.F.F.F.F.F.FFF.FFFFFFFFFF
======================================================================
FAIL: test_aes-128_cfb_no_001_decrypt (__main__.TestSequence)
----------------------------------------------------------------------
  File ".\src/scripts/cli_tests.py", line 98, in do_test_expected
    self.assertEqual(expected, actual)
AssertionError: '6bc1bee22e409f96e93d7e117393172aae2d8a571e03ac9c9eb76fac45af8e5130c81c46a35ce411e5fbc1191a0a52eff69f2445df4f9b17ad2b417be66c3710' != u'6bc1bee22e409f96e93d7e117393172aae2d8a571e03ac9c9eb76fac45af8e5130c81c46a35ce411e5fbc1191a0d0a52eff69f2445df4f9b17ad2b417be66c3710'
...
```

lieser:
- Fix reading string environment variables under windows (e.g. `BOTAN_CLEAR_CPUID`).
- Let the CLI write TSS shares in binary mode to avoid potential issues with automatic conversations.
- Fix CLI test under windows.
- Enable CLI test under windows. Note that `cli_tls_proxy_tests` is disabled because it is flaky.
- Use the binary mode under windows for reading/wrinting stdin/stdout where we handle binary data.

Fixes #1198.

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -23,6 +23,11 @@
    #include <botan/base58.h>
 #endif
 
+#ifdef _WIN32
+   #include <fcntl.h>
+   #include <io.h>
+#endif
+
 namespace Botan_CLI {
 
 Command::Command(const std::string& cmd_spec) : m_spec(cmd_spec) {
@@ -151,6 +156,16 @@ std::ostream& Command::output() {
    return std::cout;
 }
 
+std::ostream& Command::output_binary() {
+   if(m_output_stream) {
+      return *m_output_stream;
+   }
+#ifdef _WIN32
+   _setmode(_fileno(stdout), _O_BINARY);
+#endif
+   return std::cout;
+}
+
 std::ostream& Command::error_output() {
    if(m_error_output_stream) {
       return *m_error_output_stream;
@@ -176,6 +191,9 @@ void Command::read_file(const std::string& input_file,
                         const std::function<void(uint8_t[], size_t)>& consumer_fn,
                         size_t buf_size) {
    if(input_file == "-") {
+#ifdef _WIN32
+      _setmode(_fileno(stdin), _O_BINARY);
+#endif
       do_read_file(std::cin, consumer_fn, buf_size);
    } else {
       std::ifstream in(input_file, std::ios::binary);
--- a/src/scripts/ci_build.py
+++ b/src/scripts/ci_build.py
@@ -732,7 +732,7 @@ def main(args=None):
                          os.path.join(build_dir, 'fuzzer_corpus'),
                          os.path.join(build_dir, 'build/fuzzer')])
 
-        if target in ['shared', 'coverage', 'sanitizer'] and options.os != 'windows':
+        if target in ['shared', 'coverage', 'sanitizer']:
             botan_exe = os.path.join(build_dir, 'botan-cli.exe' if options.os == 'windows' else 'botan')
 
             args = ['--threads=%d' % (options.build_jobs)]
--- a/src/scripts/test_cli.py
+++ b/src/scripts/test_cli.py
@@ -1295,7 +1295,9 @@ def cli_tls_http_server_tests(tmp_dir):
         tls_server.kill()
 
 def cli_tls_proxy_tests(tmp_dir):
-    if not run_socket_tests() or not check_for_command("tls_proxy"):
+    # In the Windows GitHub CI this sometimes fails, for reasons unknown.
+    # The connectionn to the TLS proxy is forcibly closed by the remote host.
+    if not run_socket_tests() or platform.system() == 'Windows' or not check_for_command("tls_proxy"):
         return
 
     server_port = port_for('tls_proxy_backend')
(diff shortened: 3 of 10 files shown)


src/cli/cli.cpp
```diff
@@ -151,6 +156,16 @@ std::ostream& Command::output() {
    return std::cout;
 }
 
+std::ostream& Command::output_binary() {
+   if(m_output_stream) {
+      return *m_output_stream;
+   }
+#ifdef _WIN32
+   _setmode(_fileno(stdout), _O_BINARY);
+#endif
```
reneme, line 165: Mhh, juicy global state. 😞

Assuming one mixes `output()` and `output_binary()` for whatever reasons, we might end up with subtle misuse:

```C++
auto binstream = output_binary(); // silently sets the mode on `std::cout` to `_O_BINARY`
auto txtstream = output(); // `std::cout` is still set to binary
```

... or:

```C++
auto txtstream = output();
auto binstream = output_binary(); // silently sets the mode on `std::cout` to `_O_BINARY` (affecting txtstream as well)
```

Let's at least document that and discourage the mixed usage.

src/scripts/ci_build.py
```diff
@@ -732,7 +732,7 @@ def main(args=None):
                          os.path.join(build_dir, 'fuzzer_corpus'),
                          os.path.join(build_dir, 'build/fuzzer')])
 
-        if target in ['shared', 'coverage', 'sanitizer'] and options.os != 'windows':
+        if target in ['shared', 'coverage', 'sanitizer']:
```
reneme, line 735: # 🥳

src/scripts/test_cli.py
```diff
@@ -170,12 +173,18 @@ def cli_config_tests(_tmp_dir):
     ldflags = test_cli("config", "ldflags")
     libs = test_cli("config", "libs")
 
-    if len(prefix) < 4 or prefix[0] != '/':
-        logging.error("Bad prefix %s", prefix)
+    if os.name == 'nt':
```
reneme, line 176: For reference: Another (perhaps slightly more readable option):

```suggestion
    if platform.system() == 'Windows':
```

reneme (approved):
Nice! ❤️

src/cli/cli.cpp
lieser: I agree that changing the global state is not nice. But I don't think it is a big problem in this case.

If one writes binary data from stdin to e.g. pipe it to somewhere, any additional "normal" test output would likely break the use case anyway.

lieser: added some doc strings

src/scripts/test_cli.py
lieser: changed

lieser:
Nota that the Windows CLI test is currently flaky in the CI. Looks like some strange environment issues.
Unsure if we want to still merge it with the test enabled.

src/cli/cli.cpp
reneme: Thanks.

reneme:
> Unsure if we want to still merge it with the test enabled.

If its flaky we shouldn't merge it. But perhaps its worth another try to find the reason for the flakiness. Admittedly I didn't look at the log so far.

Nevertheless, there's certainly value in having the CLI covered under windows.

randombit (approved):

randombit:
Ideally we can fix the flakiness since otherwise we'll inevitably end up with regressions, but if not fine to merge with the test step disabled.

randombit:
Is the flakyness confined to just the socket tests? If so we could just skip those and still get significant test coverage for the cli. Already there is a `run_socket_tests` guard that skips those tests on FreeBSD (which is no longer really needed since we don't use Travis CI anymore...) we could change this so the tests are similarly skipped on Windows.

lieser:
I decided to just disable the `cli_tls_proxy_tests` test and not all socket tests, as it is the only one I so far saw failing in the Windows CI.

PR #3846, opened 2023-12-19
FIX: potential non-constant time division in Kyber

### LINKED ISSUE #3844: Possible variable-time division when decapsulating in Kyber
This bit of code is used in compressing a polynomial ring element into a (secret) message: https://github.com/randombit/botan/blob/3dcf8af5a77e9e7643cc58cd5498d3950c1470ef/src/lib/pubkey/kyber/kyber_common/kyber.cpp#L389-L390

To do so, it performs a division by Q that might not necessarily compile to a multiplication instruction: looking at the output of some C compilers using https://godbolt.org/z/sKn3TKKGq and https://godbolt.org/z/8GqKoTfYh for example, a division instruction is emitted even when -O3 is specified. Should a division instruction be emitted, its execution time would likely be variable and leak information about its secret input.

We reported a similar issue in the CRYSTALS-Kyber reference implementation; you may want to use their fix: https://github.com/pq-crystals/kyber/commit/dda29cc63af721981ee2c831cf00822e69be3220

reneme:
Some compilers generate a non-constant time division operation for the division by `KyberConstants::Q`.

Fix is based on the patch in the reference implementation: https://github.com/pq-crystals/kyber/commit/dda29cc63af721981ee2c831cf00822e69be3220

Thanks @xvzcf for reporting this.

@randombit we'll also open a pull request on against the 3.2.0 release.

closes #3844

References:

* [Thread in the PQC forum mailing list](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/hWqFJCucuj4)
* [kyberslash.cr.yp.to](https://kyberslash.cr.yp.to/)

### DIFF
--- a/src/lib/pubkey/kyber/kyber_common/kyber.cpp
+++ b/src/lib/pubkey/kyber/kyber_common/kyber.cpp
@@ -383,12 +383,21 @@ class Polynomial {
 
          this->csubq();
 
+         auto compress = [](uint32_t t) {
+            // (t << 1) + ((KyberConstants::Q / 2) / KyberConstants::Q) & 1
+            // Note that magic numbers assume that ::Q = 3329
+            t <<= 1;
+            t += 1665;
+            t *= 80635;
+            t >>= 28;
+            t &= 1;
+            return static_cast<uint8_t>(t);
+         };
+
          for(size_t i = 0; i < size() / 8; ++i) {
             result[i] = 0;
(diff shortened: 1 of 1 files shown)


randombit:
Given 3.3.0 is coming out in a few weeks and this is just a side channel I don't see any real need to backport. Keep in mind we only support the latest release in any series. If someone wants to pull this they can always cherry pick it.

randombit (approved):

PR #3849, opened 2023-12-19
Add overview of supported hardware acceleration

### LINKED ISSUE #3824: Overview of hardware acceleration support
I'd like to see an overview of all hardware acceleration features utilized by Botan, on which platforms they are supported, whether they are enabled by default and how to enable/disable them, how to find out if they are supported on a specific system, and probably also a rough estimate on the factor of performance gain when using them. The best location of such an overview probably is the handbook. I'd also think hardware random number generators should be part of this overview, so I am not sure whether "hardware acceleration" would be the right headline.

None of the major crypto libraries currently provide such an overview, so that would also set Botan apart.

securitykernel:
Adds an overview of supported hardware acceleration by platform to the handbook.

Includes:

* Algorithm
* CPU instruction/extension
* Botan module
* Added in Botan version

Also adds information on how to configure hardware support at build time via `configure.py` and at runtime using environment variables.

One thing that came out during documentation of the `configure.py` options, which Sphinx can automatically highlight via forwarding to [docutils](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#option-lists): `--disable-sse4.1` is not a [POSIX-compliant command line option](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_01), as it only allows letters, numbers, and hyphens (not dots). ~I changed `configure.py` accordingly, so this means that:~

* ~`configure.py --disable-sse4.1` is now `configure.py --disable-sse41`~
* ~`configure.py --disable-sse4.2` is now `configure.py --disable-sse42`~

Decided to postpone this change to 4.0, if ever.

Closes #3824

### DIFF
--- a/doc/hardware_acceleration.rst
+++ b/doc/hardware_acceleration.rst
@@ -1,0 +1,175 @@
+Hardware Acceleration
+==============================
+
+Botan provides built-in support for hardware acceleration of certain algorithms
+on certain platforms. These alternate implementations use special CPU instructions
+that are not available on all platforms and either speed up the algorithm
+or improve security in terms of side channel resistance.
+
+A “base” software implementation is always provided. For example, for the AES-128
+block cipher three implementations are available. All of the AES-128 implementations
+are immune to common cache/timing based side channels.
+
+* If AES hardware support is available (AES-NI, POWER8, Aarch64) use that
+* If 128-bit SIMD with byte shuffles are available (SSSE3, NEON, or Altivec),
+  use the vperm technique published by Mike Hamburg at CHES 2009
+* If no hardware or SIMD support, fall back to a constant time bitsliced implementation
+
+The following sections list the platforms and algorithms for which hardware acceleration
+is available. If the CPU specific optimizations are available at runtime, they are
+automatically used if enabled in the build. If not, the base implementation is used.
+
+x86
+--------------
+
+On x86-64 and x86-32 platforms, the following CPU specific optimizations are available:
+
++-----------+--------------------------------------------+--------------------+------------+
+| Algorithm | Extension                                  | Module             | Added in   |
++===========+============================================+====================+============+
+| AES       | AES-NI                                     | `aes_ni`           | 1.9.3      |
+|           |                                            |                    |            |
+|           | SSSE3                                      | `aes_vperm`        | 1.9.10     |
++-----------+--------------------------------------------+--------------------+------------+
+| AES-GCM   | CLMUL                                      | `ghash_cpu`        | 1.11.6     |
+|           |                                            |                    |            |
+|           | SSSE3                                      | `ghash_vperm`      | 1.9.10     |
++-----------+--------------------------------------------+--------------------+------------+
+| Argon2    | AVX2                                       | `argon2_avx2`      | 3.0.0      |
+|           |                                            |                    |            |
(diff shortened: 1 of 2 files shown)


configure.py
```diff
@@ -380,12 +380,12 @@ def process_command_line(args):
...
 
     for isa_extn_name in isa_extensions:
-        isa_extn = isa_extn_name.lower().replace(' ', '')
+        isa_extn = isa_extn_name.lower().replace(' ', '').replace('.', '')
 
         target_group.add_option('--disable-%s' % (isa_extn),
                                 help='disable %s intrinsics' % (isa_extn_name),
                                 action='append_const',
-                                const=isa_extn.replace('-', '').replace('.', '').replace(' ', ''),
+                                const=isa_extn.replace('-', '').replace(' ', ''),
```
reneme, line 388: For the sake of backward compatibility (e.g. with the Conan recipe), maybe add the switches that don't have a period (`.`) instead of replacing them with the existing ones.

https://github.com/conan-io/conan-center-index/blob/f89d800e5d4eb8d42af1560bf2cc28ac894428f6/recipes/botan/all/conanfile.py#L329-L333

reneme (approved):

configure.py
securitykernel: Good point, will take care.

doc/hardware_acceleration.rst
```diff
@@ -0,0 +1,157 @@
...
+==============================
+
+Botan provides built-in support for hardware acceleration of certain algorithms
+on certain platforms. These alternate implementations use special CPU instructions
+that are not available on all platforms and either speed up the algorithm
+or improve security in terms of side channel resistance.
+
+A “base” software implementation is always provided. For example, for the AES-128
+block cipher three implementations are available. All of them are immune to
+common cache/timing based side channels.
```
randombit, line 11: This is slightly confusing in that someone could misread it as "All of [the "base" software implementations of all ciphers]" vs "All of [the AES implementations]"

doc/hardware_acceleration.rst
```diff
@@ -0,0 +1,157 @@
...
+| NOEKEON   | AltiVec                                    | `noekeon_simd`     | 1.9.4      |
++-----------+--------------------------------------------+--------------------+------------+
+
+Configuring Acceleration
+------------------------------
+
+Hardware acceleration can be disabled at *compile time* by passing certain
+``--disable-*`` options to ``configure.py``. This will cause the base software
+implementation to be used instead of the hardware accelerated one.
+
```
randombit, line 139: Might be worth mentioning here that `--disable-modules=` can be used to remove a specific feature, if this is desirable.

configure.py
securitykernel: I reverted this change, it is not worth changing this now. Maybe make it POSIX-compliant in 4.0. I formatted the options manually instead.

doc/hardware_acceleration.rst
securitykernel: Made it more clear that it only applies to the AES-128 example.

doc/hardware_acceleration.rst
securitykernel: Added that, thanks!

PR #3852, opened 2023-12-22
Work around a warning in GCC 13

### LINKED ISSUE #3851: Dangling reference warning with GCC 13
Introduced in #3820 

```
src/cli/tls_utils.cpp: In static member function ‘static std::string Botan_CLI::TLS_Client_Hello_Reader::format_hello(const std::variant<Botan::TLS::Client_Hello_13, Botan::TLS::Client_Hello_12>&)’:
src/cli/tls_utils.cpp:126:22: error: possibly dangling reference to a temporary [-Werror=dangling-reference]
  126 |          const auto& hello_base =
      |                      ^~~~~~~~~~
src/cli/tls_utils.cpp:127:23: note: the temporary was destroyed at the end of the full expression ‘std::visit<Botan_CLI::TLS_Client_Hello_Reader::format_hello(const std::variant<Botan::TLS::Client_Hello_13, Botan::TLS::Client_Hello_12>&)::<lambda(const auto:62&)>, const variant<Botan::TLS::Client_Hello_13, Botan::TLS::Client_Hello_12>&>(<lambda closure object>Botan_CLI::TLS_Client_Hello_Reader::format_hello(const std::variant<Botan::TLS::Client_Hello_13, Botan::TLS::Client_Hello_12>&)::<lambda(const auto:62&)>(), (* & hello))’
  127 |             std::visit([](const auto& ch) -> const Botan::TLS::Client_Hello& { return ch; }, hello);
      |             ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
make: *** [Makefile:1519: build/obj/cli/tls_utils.o] Error 1
```

cc @reneme maybe this is harmless, idk, haven't looked into it carefully yet.

reneme:
... which claims that the replaced code creates a 'possibly dangling reference'. I'm quite sure that this warning is a false positive but let's be safe.

At least in Godbolt that [seems to be fine with GCC 13](https://godbolt.org/z/6GvMqcf9h).

Closes #3851.

### DIFF
--- a/src/cli/tls_utils.cpp
+++ b/src/cli/tls_utils.cpp
@@ -123,8 +123,8 @@ class TLS_Client_Hello_Reader final : public Command {
          const std::variant<Botan::TLS::Client_Hello_13, Botan::TLS::Client_Hello_12>& hello) {
          std::ostringstream oss;
 
-         const auto& hello_base =
-            std::visit([](const auto& ch) -> const Botan::TLS::Client_Hello& { return ch; }, hello);
+         const auto* hello_base =
+            std::visit([](const auto& ch) -> const Botan::TLS::Client_Hello* { return &ch; }, hello);
 
          const auto version = std::visit(Botan::overloaded{
                                             [](const Botan::TLS::Client_Hello_12&) { return "1.2"; },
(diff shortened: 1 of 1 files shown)


randombit (approved):
Thanks

Issue #3836, opened 2023-12-14
Error building CLI Target using Emscripten compiler.


Satancito:
**Configure command in powershell**     
`prompt> & "/home/x/.emsdk/upstream/emscripten/emconfigure" python "$BOTAN_UNZIPPED_DIR/configure.py" $options $BotanOptions`

**Generated Makefile**
```
# Makefile for Botan 3.2.0 generated using
# 'configure.py --debug-mode --with-debug-info --no-optimizations --link-method=copy --cpu=wasm --os=emscripten --cc=emcc --disable-shared-library --prefix=/mnt/c/Users/Megam/.CppLibs/Botan-3.2.0-Debug-Emscripten-Full'

# Paths to relevant programs
CXX            = /home/x/.emsdk/upstream/emscripten/em++
LINKER         = /home/x/.emsdk/upstream/emscripten/em++
AR             = /home/x/.emsdk/upstream/emscripten/emar
AR_OPTIONS     = cr
PYTHON_EXE     = /usr/bin/python

# Compiler Flags
ABI_FLAGS      =  
LANG_FLAGS     = -s DISABLE_EXCEPTION_CATCHING=0 -std=c++20 -D_REENTRANT 
LANG_EXE_FLAGS = -s ALLOW_MEMORY_GROWTH=1 -s WASM=1 -s NO_DISABLE_EXCEPTION_CATCHING
CXXFLAGS       = -g -DBOTAN_IS_BEING_BUILT
WARN_FLAGS     = -Wall -Wextra -Wpedantic -Wshadow -Wstrict-aliasing -Wstrict-overflow=5 -Wcast-align -Wmissing-declarations -Wpointer-arith -Wcast-qual -Wshorten-64-to-32
LIB_FLAGS      = 
LDFLAGS        = 

EXE_LINK_CMD   = em++

LIB_LINKS_TO        =  
BUILD_DIR_LINK_PATH = -L.
EXE_LINKS_TO        = -lbotan-3 $(LIB_LINKS_TO) 

# Note: Remember to adapt the generation of compile_commands.json, when changing
#       any of those variables or their associations to compilation units.
#       See compile_commands.json.in and configure.py (generate_build_info)
BUILD_FLAGS    = $(ABI_FLAGS) $(LANG_FLAGS) $(CXXFLAGS) $(WARN_FLAGS)

SCRIPTS_DIR    = /home/x/.PsBotan/Botan-3.2.0/src/scripts
[...]

PR #3854, opened 2023-12-22
Use variable to refer to Emscripten when linking

randombit:
This allows overriding the path to emscripten using CXX and having it work consistently.

GH #3836


reneme (approved):
Good to know!

PR #3855, opened 2023-12-22
[std::span] BigInt - use std::span<> in some internal implementations

reneme:
### Pull Request Dependencies

* https://github.com/randombit/botan/pull/3865
* https://github.com/randombit/botan/pull/3866

### Description

Note: I split that into multiple commits for easier reviewability. Especially the first commit is fixing the indentation of `BigInt`'s doxygen comments. I'm guessing that this is an artifact of the introduction of `clang-format`.

This introduces `std::span<>` overloads for `BigInt` (mostly in the dependended upon PRs) and moves the high-level implementations to be based on `std::span<>`. It doesn't touch the implementations in `big_code.cpp`, `big_rand.cpp`, or `big_ops*.cpp`. We could do that as a follow-up, though.

### DIFF
--- a/src/lib/math/bigint/bigint.h
+++ b/src/lib/math/bigint/bigint.h
@@ -686,10 +686,16 @@ class BOTAN_PUBLIC_API(2, 0) BigInt final {
       void binary_decode(std::span<const uint8_t> buf);
 
       /**
-       * Place the value into out, zero-padding up to size words
-       * Throw if *this cannot be represented in size words
+       * Place the value into @p out, zero-padding up to @p size words
+       * Throw if *this cannot be represented in @p size words
        */
-      void encode_words(word out[], size_t size) const;
+      void encode_words(word out[], size_t size) const { encode_words(std::span{out, size}); }
+
+      /**
+       * Place the value into @p out, zero-padding to fill the @p out
+       * Throw if *this cannot be represented in the size of @p out.
+       */
+      void encode_words(std::span<word> out) const;
 
       /**
        * If predicate is true assign other to *this
--- a/src/lib/math/bigint/bigint.h
+++ b/src/lib/math/bigint/bigint.h
@@ -14,6 +14,7 @@
 #include <botan/secmem.h>
 #include <botan/types.h>
 #include <iosfwd>
+#include <span>
 
 namespace Botan {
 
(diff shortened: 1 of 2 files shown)


src/lib/math/bigint/bigint.h
```diff
@@ -567,144 +574,156 @@ class BOTAN_PUBLIC_API(2, 0) BigInt final {
...
 
       /**
-     * Read integer value from a byte array with given size
-     * @param buf byte array buffer containing the integer
-     * @param length size of buf
-     */
-      void binary_decode(const uint8_t buf[], size_t length);
+       * Store BigInt-value in a given byte array. If the buffer length
+       * is less than the size of the value, then it will be truncated.
+       * If it is greater than the size of the value, it will be zero-padded.
```
reneme, line 674: @randombit The implementation didn't seem to do actually ensure a zero-padding. I would have expected a call to `clear_mem()` at least. I think that needs to be addressed somehow, but I'm not sure about the implications.

Similarly, in `BigInt::encode_words()`.

randombit:
So the constructor changes (and of course the indentation) look fine.

However for the rest, I think we're in a similar situation as with the `mem_ops.h` changes - `span` interfaces are certainly preferable *BUT* do we want to have those interfaces at all? That is should we `span`ify them, or should we instead deprecate and work to remove them entirely?

I *really* wish there was some way of getting behavior like `pub(crate)` acts in Rust where an interface is available freely within the library itself but completely unavailable for consumers. I guess we could do things like

```
#if defined(BOTAN_IS_BEING_BUILT)
void some_function_we_need_but_dont_want_to_expose_to_users();
#endif
```

but this may violate ODR, IDK.

For now can you split this into 3 PRs

- The reformatting (instant approve+merge)
- The constructor changes (same)
- The `binary_decode` etc changes (we'll have to figure out what we want to do)

src/lib/math/bigint/bigint.h
```diff
@@ -649,41 +656,53 @@ class BOTAN_PUBLIC_API(2, 0) BigInt final {
...
+      void binary_decode(std::span<const uint8_t> buf);
 
       /**
-       * Place the value into out, zero-padding up to size words
-       * Throw if *this cannot be represented in size words
+       * Place the value into @p out, zero-padding up to @p size words
+       * Throw if *this cannot be represented in @p size words
        */
-      void encode_words(word out[], size_t size) const;
+      void encode_words(word out[], size_t size) const { encode_words(std::span{out, size}); }
```
randombit, line 699: Should we deprecate some of these ptr+len interfaces?

randombit:
Going with an `_` prefix to indicate to users that it’s public for (reasons) but not intended to application use sounds pretty good.

src/lib/math/bigint/bigint.h
randombit: I’ll look into what’s going with this

src/lib/math/bigint/bigint.h
reneme: Yes please! :)

randombit:
I think this has been pre-empted by #4056

PR #3856, opened 2023-12-22
BufferSlicer::take<count>() & BufferStuffer::next<bytes>()

reneme:
Those return std::span<> with a statically known extent. The new range-based memory operators and endian-loaders can use those to statically check for matching buffer sizes. See also: https://github.com/randombit/botan/pull/3707#issuecomment-1820766105.


randombit:
LGTM

securitykernel (approved):

PR #3858, opened 2023-12-29
Optimize by using /Zc:throwingNew with MSVC

lanurmi:
Assume that `operator new` throws, which is the standard behavior since C++98. For compatibility with very old MSVC versions, MSVC by default assumes `new` can return zero on error.

Further explanation:
https://learn.microsoft.com/en-us/cpp/build/reference/zc-throwingnew-assume-operator-new-throws

Disclaimer: I don't have any statistics to prove that this optimization actually does any good to either speed or size. Still, this seems the right thing to do in a project that does not care about compatibility with ancient MSVC versions.

### DIFF
--- a/src/build-data/cc/msvc.txt
+++ b/src/build-data/cc/msvc.txt
@@ -16,7 +16,7 @@ add_lib_option "%s.lib"
 
 compile_flags "/nologo /c"
 
-optimization_flags "/O2 /Oi"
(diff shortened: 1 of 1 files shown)


randombit (approved):
LGTM thanks! I am surprised this isn't the default behavior in the year 1998+25

PR #3863, opened 2023-12-31
Small cleanups for compiler.h

randombit:


src/lib/utils/compiler.h
```diff
@@ -156,32 +156,19 @@
...
-#endif
-
-#if defined(BOTAN_BUILD_COMPILER_IS_GCC)
-   #define BOTAN_DIAGNOSTIC_PUSH _Pragma("GCC diagnostic push")
-   #define BOTAN_DIAGNOSTIC_IGNORE_DEPRECATED_DECLARATIONS \
-      _Pragma("GCC diagnostic ignored \"-Wdeprecated-declarations\"")
-   #define BOTAN_DIAGNOSTIC_IGNORE_INHERITED_VIA_DOMINANCE
-   #define BOTAN_DIAGNOSTIC_POP _Pragma("GCC diagnostic pop")
-#elif defined(BOTAN_BUILD_COMPILER_IS_CLANG)
+#if defined(__clang__)
```
reneme, line 159: Interesting catch! Maybe leave a note to future-us, why that isn't using `BOTAN_BUILD_COMPILER_IS_*`?

reneme (approved):

src/lib/utils/compiler.h
randombit: Done

PR #3870, opened 2024-01-03
[std::span] Pass std::span buffers into Block_Cipher Implementations

reneme:
This demotes Block_Cipher::encrypt_n/decrypt_n to top-level methods and introduces new (private) virtual methods (encrypt/decrypt_blocks) that use std::span for the in/out buffers. Also, this adapts all block cipher implementations in the library to use the new API.

### DIFF
--- a/src/lib/block/seed/seed.cpp
+++ b/src/lib/block/seed/seed.cpp
@@ -7,6 +7,7 @@
 
 #include <botan/internal/seed.h>
 
+#include <botan/assert.h>
 #include <botan/internal/loadstor.h>
 #include <botan/internal/prefetch.h>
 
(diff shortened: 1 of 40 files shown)


src/lib/block/seed/seed.cpp
```diff
@@ -173,9 +169,7 @@ void SEED::key_schedule(std::span<const uint8_t> key) {
 
    secure_vector<uint32_t> WK(4);
 
-   for(size_t i = 0; i != 4; ++i) {
-      WK[i] = load_be<uint32_t>(key.data(), i);
-   }
+   load_be(key.first<BLOCK_SIZE>(), WK[0], WK[1], WK[2], WK[3]);
```
reneme, line 172: With https://github.com/randombit/botan/pull/3869 this could become:

```C++
auto WK = load_be<secure_vector<uint32_t>>(key.first<BLOCK_SIZE>());
```

randombit:
Sorry just getting back to this.

I'm not sold on this change as it stands because it has two effects

1) We are now accepting arguments from the user, and ... just not checking then. An implicit assumption of the interface (with or without span) is that both arrays point to the same length of memory, namely `blocks*BLOCK_SIZE` bytes, and also that they don't overlap. However now (with span) we are allowing the user to tell us via the `span` lengths "oh no these input and output arrays aren't the same length, lol lmao" but we blithely ignore it. Likewise we still require the user to tell us the number of blocks, even though we can deduce it from the span lengths.

2) Overall the `span` based flow is harder to understand (at least for me) than ptr + index.

That said `span` *does have potential* to improve the internal implementations of the block ciphers (for example #3942) However I think a better approach here is instead of a wholesale conversion of the ciphers is to instead pick a single one [*] and use it as a testbed for experimentation of approaches to ease implementation. Effectively

```
void encrypt_n(const uint8_t inb[], uint8_t outb[], size_t blocks) {
   std::span<const uint8_t> in(inb, inb + blocks*block_size());
   std::span<uint8_t> out(outb, outb + blocks*block_size());

   // span-oriented cipher follows
}
```

Then when things reach the point that it's clear there are substantial improvements in usability we can switch over wholesale.

[*] I've traditionally used Blowfish as my arbitrary choice for these kinds of things, because it's very fast so any performance impacts are visible, and it's not "important" so if bugs are introduced it's not the end of the world.

reneme:
> I think a better approach here is instead of a wholesale conversion of the ciphers is to instead pick a single one [*] and use it as a testbed for experimentation of approaches to ease implementation.

Agreed. See #4151.

PR #3871, opened 2024-01-04
Fix mul128.h to support MSVC ARM64 target

lanurmi:
Backport compilation fix from master.

(cherry picked from commit c2baf41bd06d5c9012be83c83d831baf4ecee4a5)

This is the commit from PR #3000 cherry-picked and adapted for the release-2 branch. With this fix, (cross-)compilation for the ARM64 target is successful. While the changes appear to be pretty trivial, I should point out that *I do not have access to Windows ARM64 hardware, so I cannot run any tests*.

Perhaps @nursik has the proper hardware to test this branch?


reneme:
Could you please rebase and force-push your commit? CI is currently hanging, as [we had to fix the GCC 4.8 build](https://github.com/randombit/botan/pull/3875).

reneme:
The change looks good to me. Nevertheless, it would be good to be able to test it before merging. Maybe @Trafo has access to an ARM64 box and is willing to compile this branch, check that the new `#ifdef` is taking effect and unit tests are running fine? 😄

randombit:
We're due for a new 2.x release anyway. Even lacking testing, this is probably worth merging since the status quo is that it absolutely doesn't compile on this target. Best case, the backport is correct and everything works. Worst case it (still) fails to compile.

randombit (approved):

PR #3874, opened 2024-01-04
Fix: Potential side-channel in Kyber's Ciphertext::compress functions

FAlbertDev:
This is a follow-up to #3846 and relates to the [KyberSlash](https://kyberslash.cr.yp.to/faq.html) side channel. It also fixes the `Ciphertext::compress` method's potential division side channel.

To remove the critical division operation, we enforce the optimization most compilers perform to replace the division by multiplication and various shifts (see [godbolt](https://godbolt.org/z/vfn7Psn5K)).

### DIFF
--- a/src/lib/pubkey/kyber/kyber_common/kyber.cpp
+++ b/src/lib/pubkey/kyber/kyber_common/kyber.cpp
@@ -71,6 +71,20 @@ KyberMode::Mode kyber_mode_from_string(std::string_view str) {
    throw Invalid_Argument(fmt("'{}' is not a valid Kyber mode name", str));
 }
 
+/**
+ * Constant time implementation for computing an unsigned integer division
+ * with KyberConstants::Q = 3329.
+ *
+ * It enforces the optimization of various compilers,
+ * replacing the division operation with multiplication and shifts.
+ *
+ * @returns (a / KyberConstants::Q)
+ */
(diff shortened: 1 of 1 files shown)


src/lib/pubkey/kyber/kyber_common/kyber.cpp
```diff
@@ -71,6 +71,20 @@ KyberMode::Mode kyber_mode_from_string(std::string_view str) {
...
+/**
+ * Constant time implementation for computing an unsigned integer division
+ * with KyberConstants::Q = 3329.
+ *
+ * It enforces the optimization of various compilers,
+ * replacing the division operation with multiplication and shifts.
+ *
+ * @returns (a / KyberConstants::Q)
+ */
+uint16_t int_div_kyber_q(uint32_t a) {
```
reneme, line 83: Nit: functions that carry some constant-time guarantee are prefixed with `ct_` by convention. Maybe so:

```suggestion
uint16_t ct_int_div_kyber_q(uint32_t a) {
```

reneme (approved):

FAlbertDev:
With #3898, a better, less "magic number" approach is introduced; however, it depends on the open PR #3887, which we will probably not merge until the 3.3 release. I recommend merging this quick fix for the 3.3 release and then going for the improved approach afterward.

randombit (approved):
:+1:

PR #3876, opened 2024-01-07
Plot benchmark graphs and create HTML report

securitykernel:
Adds a new option `bench.py --html-report` that plots graphs using `matplotlib` and outputs a HTML report including Botan and OpenSSL version info, system environment, and `build.h` contents.

A sample report is available for browsing at https://securitykernel.io/bench_html/.

### DIFF
--- a/src/scripts/bench.py
+++ b/src/scripts/bench.py
@@ -5,12 +5,9 @@ Compare Botan with OpenSSL using their respective benchmark utils
 
 (C) 2017,2022 Jack Lloyd
     2023 René Meusel, Rohde & Schwarz Cybersecurity GmbH
-    2023 René Fischer
+    2023,2024 René Fischer
 
 Botan is released under the Simplified BSD License (see license.txt)
-
-TODO
- - Output pretty graphs with matplotlib
 """
 
 import logging
@@ -20,6 +17,12 @@ import optparse # pylint: disable=deprecated-module
 import subprocess
 import re
 import json
+import math
+import platform
+from datetime import datetime
+import shutil
+import numpy as np
+import matplotlib.pyplot as plt
 
 def setup_logging(options):
     if options.verbose:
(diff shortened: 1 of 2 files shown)


src/scripts/bench.py
```diff
@@ -20,6 +17,12 @@
 import subprocess
 import re
 import json
+import math
+import platform
+from datetime import datetime
+import shutil
+import numpy as np
+import matplotlib.pyplot as plt
```
reneme, line 25: Maybe import those optionally, so that numpy and matplotlib isn't a hard dependency for this script?

Along those lines:

```suggestion
try:
    import numpy as np
except:
    pass
```

... w/o those dependencies we could just not provide the `--html-report` option. No strong opinion either way, from my side, though.

reneme (commented):
I like the idea, and the resulting diagrams.

securitykernel:
> I like the idea, and the resulting diagrams.

Yeah, one could argue whether those are pretty, but at least they do what we want. We could use [plotly express](https://plotly.com/python/plotly-express/#gallery) instead, but don't know if it's worth. I thought about using [horizontal bar charts](https://matplotlib.org/stable/gallery/lines_bars_and_markers/barh.html) instead of vertical ones, as they seem to be used more commonly for benchmarking.

src/scripts/bench.py
```diff
@@ -380,6 +461,61 @@ def bench_key_agreement_algo(openssl, botan, algo):
...
+    os_info += ' ('
+    os_info += platform.machine()
+    os_info += ')'
+
+    include_path = re.search(r'-I(.*)', run_command(['botan', 'config', 'cflags']).strip()).group(1)
+    build_h = 'N/A'
+
+    try:
+        with open(os.path.join(include_path, 'botan/build.h'), 'r', encoding='utf-8') as f:
+            build_h = f.read()
```
FAlbertDev, line 506: One thing I noticed while playing around with the benchmarks. This part is incompatible with a user-defined botan path (i.e., `--botan-cli`). Great work otherwise!

src/scripts/bench.py
randombit: No strong opinion from me either. For scripts the end user has to run to build/test the library, stock Python only (or with purely optional use of modules).

For scripts that are informational or for developer use, as here, anything goes.

randombit (approved):

src/scripts/bench.py
securitykernel: Good point, you're right. We try our best and ask the CLI where to find the headers. But even a local binary `./botan` created without installing it will output `/usr/local/include/botan-3`, the designated install location. So there is nothing I can image we can do about it.

src/scripts/bench.py
securitykernel: Thanks for the input. I'll leave it as-is.

Issue #3878, opened 2024-01-08
Botan 3.0: Cannot create custom sub-classes of `Botan::Private_Key`


reneme:
With https://github.com/randombit/botan/commit/a55e4b22b6cbeeb30ca787d4ea4e3933ccccbdf1 the [`pk_ops.h` header](https://github.com/randombit/botan/blob/fb4df9a1b05c9c519562916d4fad5c49e78c0d6a/src/lib/pubkey/pk_ops.h) became private, only the [`pk_ops_fwd.h` header](https://github.com/randombit/botan/blob/master/src/lib/pubkey/pk_ops_fwd.h) stayed public. That's not enough to create custom sub-classes of `Botan::PK_Ops::*` and prevents downstream applications from implementing their own public-key algorithms.

For instance, applications might want to do that to offload public-key crypto to custom hardware. Another use case is implementing custom hybrid schemes; e.g., combine a classical with a PQC algorithm. 

This may become a showstopper for some applications migrating from Botan 2.x to 3.x. I'm assuming that preventing applications from doing that wasn't intentional, @randombit? The obvious fix is to promote `pk_ops.h` to a public header and write a dedicated example applications demonstrating a custom public-key algorithm, as a sort-of regression test.

This is closely related to https://github.com/randombit/botan/pull/3867.

securitykernel:
This is IMHO really a crucial feature and I think what partially led to the current regression is that we (1) failed to document it properly and (2) have tests for it. A good place for tests may be examples?

reneme:
I do think that a comprehensive example (as proposed in #3879) is the best way to "regression test" this.

On that note: Currently, the build system (read: `make examples`) does not prevent us from including internal headers in example files. We should find a way to fix this.

reneme:
> Currently, the build system does not prevent us from including internal headers in example files.

See: #3880

PR #3879, opened 2024-01-09
FIX: Bring pk_ops.h back as public header

reneme:
See #3878 for details.

This also adds a new example, with three objectives:

* Showcase how to integrate custom public-key algorithms
* Explain how one could use Botan to build a hybrid public-key scheme (e.g. in the transition to PQC)
* Act as a regression test for the ability of applications to build custom public-key schemes using Botan's API

### DIFF
--- a/src/lib/pubkey/pk_ops.h
+++ b/src/lib/pubkey/pk_ops.h
@@ -16,6 +16,11 @@
 * Unless you're doing something like that, you don't need anything
 * here. Instead use pubkey.h which wraps these types safely and
 * provides a stable application-oriented API.
+*
+* Note: This header was accidentally pulled from the public API between
+*       Botan 3.0.0 and 3.2.0, and then restored in 3.3.0. If you are
+*       maintaining an application which used this header in Botan 2.x,
+*       you should make sure to use Botan 3.3.0 or later when migrating.
 */
 
 #include <botan/pk_keys.h>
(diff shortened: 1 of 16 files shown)


reneme:
Yeah, user convenience was the only reasoning. And I actually had my doubts, because the interplay between the base-classes in `pk_ops.h` with the helpers isn't really obvious. If you're saying that it wasn't public historically either, let's just hide it again.

I can't work on that today anymore, though. Will rework this first thing tomorrow.

securitykernel (approved):

src/lib/pubkey/pk_ops.h
```diff
@@ -33,7 +38,7 @@ namespace PK_Ops {
 /**
 * Public key encryption interface
 */
-class Encryption {
+class BOTAN_PUBLIC_API(3, 3) Encryption {
```
randombit, line 41: These should be `BOTAN_UNSTABLE_API`.

Using this interface requires deriving, which is not covered by our current SemVer guarantee (`doc/sem_ver.rst`) and already I know we have to change at least the decryption API to support some side channel mitigations.

randombit (changes requested):

src/lib/pubkey/pk_ops.h
reneme: Done.

randombit (approved):
LGTM, thanks

PR #3880, opened 2024-01-09
Chore: Split public and internal headers

reneme:
This is a follow-up for #3878 and causes a compilation error if we accidentally include an internal header in an example.

For that, I changed the layout of the header files linked into the build directory, like so:

```bash
# public headers (those are installed with the library)
./build/include/public/botan/*

# private headers (those are not installed and an application cannot use them)
./build/include/internal/botan/internal/*

# external headers that are installed with the library (used for PKCS#11 only)
./build/include/external/*
```

The build system templates can then selectively decide for each compilation unit which include paths are required. Practically, we always need all three types, except for the 'examples' which do not see the 'internal' headers anymore.


securitykernel:
LGTM. Maybe add a short one-liner explanation for the future you to `configure.py` on why we split up the include paths this way.

securitykernel (approved):

Issue #3837, opened 2023-12-14
Discrepancies between `T::provider()` and `T::providers()`


securitykernel:
I was looking into adding output of the provider (= hardware acceleration) used for each algorithm benchmarked to `bench.py` and noticed that the provider is only printed for cipher modes, e.g., AES-128/GCM ("clmul" in my case), but not for e.g., hash functions. From looking into the speed CLI, hash functions and cipher modes are benchmarked differently:

* For hash functions (as well as XOFs, block ciphers, stream ciphers, and MACs), `bench_providers_of()` calls `T::providers()` and then runs benchmarks for every provider returned
* For all other algorithms, (a single) `T` is directly instantiated and benchmarked, making the library choose the best available implementation, and then the output of `T::provider()` is printed along with `T`'s name

The first question is, what is the reasoning of these two different types of benchmarking? For running `bench.py` to compare with OpenSSL, only the second option is really useful, since what you want is both Botan and OpenSSL to choose the best implementation available and benchmark these against each other. I don't see a high value in benchmarking e.g., the `clmul`, `pmull` and software implementations of OpenSSL and Botan against each other separately. It makes more sense to be able to benchmark different implementations inside Botan against each other in order to check improvements of hardware-accelerated implementations against software implementations for example. To accommodate this, we could add an optional `--all-providers` switch to the speed CLI, implementing the first option. The default mode would implement the second option (for all algorithms equally).

Looking into why `armv8` was not printed as a provider for SHA-256 on my system, I found that `HashFunction::providers()` would ever only return at most {"base", "commoncrypto"}. But `SHA_256::provider()` returns `armv8` on my system, which made me wonder what is the correlation between these two functions? The value of `SHA_256::provider()` is selected at compile-time based on which hardware acceleration is available, but `HashFunction::providers()` is always hardcoded and never probes `HashFunction::provider()` at any time.

randombit:
This got messed up (I think in changes just prior to 2.0); previously "base" was precisely the baseline implementation, and to get eg "sse2" SHA-1 you would request it explicitly, and in doing so got a completely different `class` than the baseline. That is there was a 1:1 between provider strings and classes which implemented that algorithm. But now "base" can in fact be implemented via hardware accel and so it's instead 1->N.

We could address this by changing the providers to always return "base", and then add some additional getter that specifies implementation specific detail (eg "aes_ni" vs "ssse3").

>  The first question is, what is the reasoning behind these two different types of benchmarking?

[...]

PR #3884, opened 2024-01-11
Add docs on providers

securitykernel:
This is an attempt to document what *providers* in Botan are, how to use them, and how to implement them (at least roughly, I tried to make it appeal to both users and implementors) as a follow-up of #3837. IMHO this really needs a thorough review. It may serve as a starting point of discussions on the future of the provider interfaces.

### DIFF
--- a/doc/api_ref/providers.rst
+++ b/doc/api_ref/providers.rst
@@ -1,0 +1,169 @@
+External Providers
+==============================
+
+Botan ships with a variety of cryptographic algorithms in both pure software
+as well as with support from :doc:`hardware acceleration <../hardware_acceleration>`.
+
+Additionally, Botan allows to use external implementations to provide algorithms ("providers").
+
+Integrated Providers
+------------------------------
+
+PKCS#11
+^^^^^^^^^^^^^
+
+PKCS#11 is a standard API for accessing cryptographic hardware. Botan
+ships a :doc:`PKCS#11 provider <pkcs11>` for interacting with PKCS#11
+devices which provide cryptographic algorithms. It is enabled by default.
+
+TPM 1.2
+^^^^^^^^^^^^^
+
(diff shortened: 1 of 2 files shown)


doc/api_ref/providers.rst
```diff
@@ -0,0 +1,169 @@
+External Providers
+==============================
+
+Botan ships with a variety of cryptographic algorithms in both pure software
+as well as with support from :doc:`hardware acceleration <../hardware_acceleration>`.
```
reneme, line 5: Should we link to an rST label rather than a file, maybe?

doc/api_ref/providers.rst
```diff
@@ -0,0 +1,169 @@
...
+* ``Cipher_Mode``
+* ``Hash``
+* ``KDF``
+* ``MAC``
+* ``PasswordHashFamily``
+* ``PBKDF``
+* ``StreamCipher``
+* ``XOF``
+
+Each of the interfaces provides a factory method which takes string arguments
```
reneme, line 59:
```suggestion
Each of the interfaces provide a factory method which takes string arguments
```

randombit (approved):

doc/api_ref/providers.rst
securitykernel: If we would want to do that, we would need to add a label "hardware acceleration" to `hardware_acceleration.rst`, which is kind of double the work. After #3849 was merged, we are now able to reference whole documents as well as headlines with autogenerated labels. To see all labels sphinx generates, run `python3 -m sphinx.ext.intersphinx build/docs/handbook/objects.inv`.

doc/api_ref/providers.rst
securitykernel: Fixed

Issue #3776, opened 2023-10-26
Clang CI recently started failing


randombit:
Definitely looks to be some problem with the Ubuntu packaging, or something about the GHA image. It appears Clang 14 is installed but it is attempting to use libc++ 13

randombit:
Already reported https://github.com/actions/runner-images/issues/8659

randombit:
I think that would only help if we were to start using a newer clang, since the issue is that clang 14 doesn't understand what to do with gcc 13's libstdc++ `chrono`. But if we use a newer clang then we are no longer testing with our minimum required version, so inevitably we'll encounter build regressions by introducing something that works with other compilers but not clang 14.

reneme:
Looks like GitHub Actions is rolling out a new image bringing this back yet again...
https://github.com/randombit/botan/actions/runs/7543153103/job/20533520177?pr=3887

KaganCanSit:
Hello @randombit, @reneme, @devnexen,

I investigated this issue to assess the current situation and determine if any changes were required. My goal was to create a PR to revert back to the previous configuration if the issue was resolved and to observe the current situation in CI.

**Research Summary:**
- I reviewed the related [issue report](https://github.com/actions/runner-images/issues/8659). This bug report was moved to a [discussion](https://github.com/actions/runner-images/discussions/9446) due to the length of the resolution process. Then, after a long review, @lkfortuna notified the community that the bug was fixed on February 13th with the following message:

> "closing as fixed by [#9679](https://github.com/actions/runner-images/issues/9679)"

I have observed that the bug has been resolved and many projects have switched their CI builds from workarounds to standard operation.

- I analyzed `src/scripts/ci/setup_gh_action.sh` and `src/scripts/ci/gha_linux_packages.py` on the main branch.
- I confirmed that version-specific workarounds are no longer needed.

**Current Status:**
In applying the workaround, I noticed that the debian packages to be installed in `src/scripts/ci/setup_gh_action.sh` were updated with the notation (`*`). Then I analyzed the same script file on the `master` branch and noticed that the packages to be installed are now managed by `src/scripts/ci/gha_linux_packages.py` and no version is specified.

**Conclusion:**
Based on this analysis, I believe that this issue can be closed, as the problem is solved both in the use of CI and in the existing codebase (since there is no need to test it separately in CI).

Please let me know if I have missed any aspect or if there are specific concerns that still need to be addressed.

Best regards,

randombit:
Thanks for looking into this

PR #3888, opened 2024-01-16
CI: Update GitHub Actions Ubuntu workaround

reneme:
See: https://github.com/actions/runner-images/issues/8659
See also: https://github.com/randombit/botan/issues/3776

... this time using wild cards as suggested in: https://github.com/mjp41/workaround8649/blob/c8550b715ccdc17f89c8d5c28d7a48eeff9c94a8/action.yml


randombit (approved):
Thanks

You would think they could just roll the change back

PR #3892, opened 2024-01-17
Prepare PQC KEM KAT Tests for Classic McEliece

FAlbertDev:
This PR relates to the generalized PQC KEM KAT tests introduced in #3807. The following changes have been made.

- Refactor the PK_PQC_KEM_KAT_Test class to use polymorphism instead of a templated Delegator. 
- Prepare the class for Classic McEliece, i.e., allow the child class to decide which KAT values are hashed in the KAT files. Also, allow the child class to override a callback to modify how the Fixed_Output_RNG is checked after keygen/encaps. In Classic McEliece, we do not know how much randomness is required; therefore, we need to soften the check in this case.
- Apply the changes to the Tests of Kyber and FrodoKEM

### DIFF
--- a/src/tests/test_pubkey_pqc.h
+++ b/src/tests/test_pubkey_pqc.h
@@ -15,29 +15,10 @@
    #include "test_rng.h"
 
    #include <botan/hash.h>
+   #include <botan/pk_algs.h>
    #include <botan/internal/fmt.h>
 
 namespace Botan_Tests {
-
-namespace detail {
-
-template <typename T>
-concept PQC_KEM_KAT_Test_Implementation =
-   std::derived_from<typename T::private_key_t, Botan::Private_Key> &&
-   std::derived_from<typename T::public_key_t, Botan::Public_Key> &&
-   std::convertible_to<decltype(T::input_file), std::string> &&
-   std::convertible_to<decltype(T::algo_name), std::string> &&
-   requires(
-      T impl, Botan::RandomNumberGenerator& rng, std::span<const uint8_t> crypto_artefact, std::string_view algo_spec) {
-      { T(algo_spec) } -> std::same_as<T>;
-      { impl.rng_for_keygen(rng) } -> std::same_as<Botan_Tests::Fixed_Output_RNG>;
-      { impl.rng_for_encapsulation(rng) } -> std::same_as<Botan_Tests::Fixed_Output_RNG>;
-      { impl.map_value(crypto_artefact) } -> std::same_as<std::vector<uint8_t>>;
-      { impl.available() } -> std::convertible_to<bool>;
-      { std::is_constructible_v<typename T::private_key_t, Botan::RandomNumberGenerator&, decltype(impl.mode())> };
-   };
-
-}
 
 /**
  * This is an abstraction over the Known Answer Tests used by the KEM candidates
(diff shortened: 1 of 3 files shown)


src/tests/test_pubkey_pqc.h
```diff
@@ -49,22 +30,54 @@ concept PQC_KEM_KAT_Test_Implementation =
...
-      using Public_Key = typename Delegate::public_key_t;
+      PK_PQC_KEM_KAT_Test(const std::string& algo_name, const std::string& input_file) :
+            PK_Test(algo_name, input_file, "Seed,SS,PK,SK,CT") {}
+
+      // --- Callbacks ---
+
+      /// Map a recomputed value to the expected value from the KAT vector (e.g. apply a hash function if Botan's KAT entry is hashed)
+      virtual std::vector<uint8_t> map_value(const std::string& params,
+                                             std::span<const uint8_t> value,
+                                             VarType) const = 0;
```
reneme, line 46:
```suggestion
      virtual std::vector<uint8_t> map_value(const std::string& params,
                                             std::span<const uint8_t> value,
                                             VarType value_type) const = 0;
```

src/tests/test_pubkey_pqc.h
```diff
@@ -49,22 +30,54 @@ concept PQC_KEM_KAT_Test_Implementation =
...
+      virtual void inspect_rng_after_encaps(const std::string& params,
+                                            const Fixed_Output_RNG& rng_encaps,
+                                            Test::Result& result) const {
+         BOTAN_UNUSED(params);
+         result.confirm("All prepared random bits used for encapsulation", rng_encaps.empty());
+      }
 
    private:
-      bool skip_this_test(const std::string& header, const VarMap&) override {
+      bool skip_this_test(const std::string& params, const VarMap&) override {
```
reneme, line 75: Due to the (partial) default implementation and the fact that this is delegating to a new interface, I would say this should be marked final:

```suggestion
      bool skip_this_test(const std::string& params, const VarMap&) final {
```

src/tests/test_pubkey_pqc.h
```diff
@@ -77,47 +90,54 @@ class PK_PQC_KEM_KAT_Test : public PK_Test {
...
-         result.test_is_eq("Generated private key", d.map_value(sk.raw_private_key_bits()), vars.get_req_bin("SK"));
-         result.confirm("All prepared random bits used for key generation", rng_keygen.empty());
+         auto sk = Botan::create_private_key(algo_name(), rng_keygen, params);
+         result.test_is_eq("Generated private key",
+                           map_value(params, sk->raw_private_key_bits(), VarType::PrivateKey),
+                           vars.get_req_bin("SK"));
+         inspect_rng_after_keygen(params, rng_keygen, result);
 
          // Algorithm properties
-         result.test_eq("algorithm name", sk.algo_name(), algo_name());
```
reneme, line 98: I think we should keep this. It checks that the class under test correctly identifies itself with the correct name. E.g. in Kyber this was broken for a while (as in: it identified as "Kyber512/768/1024" when in Botan the `algo_name()` doesn't contain the key length by convention).

Unfortunately, the algo name string is duplicated at several places across the code base. Therefore, it makes sense to pin it in the test.

src/tests/test_pubkey_pqc.h
```diff
@@ -77,47 +90,54 @@ class PK_PQC_KEM_KAT_Test : public PK_Test {
...
          // Encapsulation
-         auto enc = Botan::PK_KEM_Encryptor(pk2, "Raw");
+         auto enc = Botan::PK_KEM_Encryptor(*pk2, "Raw");
          const auto encaped = enc.encrypt(rng_encaps, 0 /* no KDF */);
-         result.test_is_eq("Shared Secret", encaped.shared_key(), Botan::lock(vars.get_req_bin("SS")));
-         result.test_is_eq("Ciphertext", d.map_value(encaped.encapsulated_shared_key()), vars.get_req_bin("CT"));
-         result.confirm("All prepared random bits used for encapsulation", rng_encaps.empty());
+         result.test_is_eq("Shared Secret",
+                           encaped.shared_key(),
+                           Botan::lock(map_value(params, vars.get_req_bin("SS"), VarType::SharedSecret)));
```
reneme, line 131: I think the mapping should happen on the algorithm's result not the KAT input, no?

```suggestion
                           map_value(params, encaped.shared_key(), VarType::SharedSecret),
                           Botan::lock(vars.get_req_bin("SS"));
```

reneme (approved):
Nice. Certainly, this is more ergonomic than the delegate template. I'm still trying to remember why I went the template route initially: I believe, I didn't see a way to abstract from the different algorithm "mode" objects and didn't even think of using the high-level API in `pk_algs.h`. Thanks for spotting that!

A few things, but generally LGTM.

src/tests/test_pubkey_pqc.h
FAlbertDev: Sure. I thought the algo name was implicitly tested when creating the keys. But they are not used in these create calls, you are right 👍

src/tests/test_pubkey_pqc.h
FAlbertDev: Good catch 😅

FAlbertDev:
Thanks for your review, @reneme! I applied your suggestions.

PR #3893, opened 2024-01-18
PQC: ML-KEM

### LINKED ISSUE #4346: How to access ML-KEM?
Please forgive my ignorance. I thought that the current master included the updates of Kyber Round 3 to ML-KEM, but all I can find via API seems to be "Kyber-<XXX>-r3", not "ML-KEM-<XXX>".

Am I not looking in the right place, or is ML-KEM support just not there yet? If not there - could you share your plans for merging it in? Thanks!

reneme:
### Pull Request Dependencies

* #4373 - Mostly for the test abstractions. Will rebase again once this is merged.

### Descriptions

This builds on the existing Kyber implementation (and the flexibility added in #3387 and #4024) to provide an implementation of ML-KEM as specified in FIPS 203. The OIDs for ML-KEM are also updated to [the official definition](https://csrc.nist.gov/Projects/computer-security-objects-register/algorithm-registration#KEM).

While reviewing the implementation for consistency with the final FIPS 203 we did some minor cleanups/refactorings. Those are currently intertwined in this PR, but could be moved into an independent commit or PR.

Note that each flavor of Kyber/ML-KEM can be disabled at compile time via the modules `kyber`, `kyber_90s`, `ml_kem`.

Integration in FFI as well as into the Python bindings is also included and deprecates some of the kyber-specific FFI functions that are now replaced by algorithm-agnostic key encoding functions.

### TODOs

* [x] Currently, we use the KATs from [Kris Kwiatkowski](https://github.com/post-quantum-cryptography/KAT) which were [verified](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/s_Wez9FanHw) against the PQShield implementations.
      We should probably also look into the official ACVP tests: [#1](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-KEM-encapDecap-FIPS203), [#2](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-KEM-keyGen-FIPS203)
* [x] Use the private seeds rather than expanded structures for storing the ML-KEM private key (cc @FiloSottile)
  ~~We already have `private_key_bits()` and `raw_private_key_bits()`, the latter could return the seed (if it is available). We'll somehow have to distinguish serialized from non-serialized keys for loading: simply by size? 🤔~~
* [x] Look into the ACVP decaps KAT (currently disabled due to a lack of the `d||z`-type private key format)
_Either remove the tests entirely, or monitor whether an updated test vector is being released._
* [x] Go through FIPS 203 again and see if all validations are implemented (add tests)
* [x] ~~Decide on the API (see below)~~ (as with the other FIPS-PRs, let's not bloat this even more and build an API bridge as a follow up)
* [x] Squash (or at least clean up) the commit history
* [x] Remove ML-KEM specific key encoding functions from FFI/Python (once #4368 is merged)

### Open Questions re: API

Currently, there's no user-facing API that explicitly states "ML-KEM". Instead, users are expected to use the Kyber-classes with the ML-KEM parameter set. Technically, that should be fine, but I believe this will confuse people in a few years, when they didn't follow the PQC competition at the time and don't associate Kyber with ML-KEM anymore.

I see three options:

1. Do nothing, apart from guidance in documentation
2. Define aliases in a dedicated `ml_kem.h` header (e.g. by flat-out inheriting from the Kyber classes)
3. Rename all "Kyber" classes to "ML_KEM" (so that ML-KEM is the first-class API) and define aliases for Kyber (as suggested in 2)

Also, the FFI currently provides kyber-specific APIs. They should just work (tm) with ML-KEM key material as well, but like above, the names might become confusing for some users in the near future.

### Caveats

This pull request initially implemented the Initial Public Draft version of FIPS 203. It was never merged and eventually got updated to implement ML-KEM as specified in the final standard. As a result, no released version of Botan will implement ML-KEM-IPD.

Note that it is an addition! Kyber-R3 (and 90s mode) continue to work as before. 

### Example

```C++
#include <botan/pk_algs.h>
#include <botan/pubkey.h>
#include <botan/system_rng.h>

#include <cassert>
#include <iostream>

using namespace Botan;

int main() {
   // Alice
   System_RNG rng;
   auto sk = create_private_key("ML-KEM", rng, "ML-KEM-768");
   if(!sk) {
      std::cerr << "Failed to create private key\n";
      return 1;
   }
   auto pk = sk->public_key();

   // Bob
   const auto [ciphertext, shared_key_bob] = KEM_Encapsulation::destructure(PK_KEM_Encryptor(*pk, "Raw").encrypt(rng));

   // Alice
   auto shared_key_alice = PK_KEM_Decryptor(*sk, rng, "Raw").decrypt(ciphertext);

   // Test
   assert(shared_key_alice == shared_key_bob);
   std::cout << "Alice and Bob can now communicate securely!\n";
}

```

Closes #4346


falko-strenzke:
@reneme, that's great news. I highly appreciate that the different versions can be controlled in the build and invoked at run-time independently. 
As to whether the ipd version should be merged into the master branch: I clearly tend say it should, for the following reasons:
- Keeping it in a separate branch leaves the work of keeping the code up to date with the master branch either to a maintainer of the user of the branch.
- The "kyber" version is already in master and released. So the addition of ML-KEM-idp does not introduce really a new problem regarding the maintenance of pre-final ML-KEM versions.
- I think the decision whether and when to deprecate them and finally drop the pre-final versions of the PQC algorithms should best be taken after the final versions have been published. Considering the case of "Keccak", where the pre-final version is still contained in Botan – due to real world implementations still using it – it seems there might be a chance that the pre-final versions of the PQC algorithms also remain in the field. If that is not the case, they can be dropped after an adequate grace period.

reneme:
Rebased after underlying #3887 had to get some conflicts resolved.

reneme:
With #4024 merged, this is now directly based on master and (finally) ready for prime time. The code to actually make ML-KEM-ipd happen isn't a lot, apart from the various `#ifdef`'ed blocks to weave it into the code base. So, in my opinion, this could actually be merged -- given that it is somewhat unclear when the final ML-KEM spec will come.

Right now the module is marked as 'experimental' and won't be built unless `./configure.py --enable-experimental-features`. So the Kyber situation would be as follows:

| Module | Description | Status |
| --- | --- | --- |
| `kyber` | The round 3 spec using SHA-3 (Kyber 3.1 spec) | **Stable** |
| `kyber_90s` | Same as above but using SHA-2 and AES | **Deprecated** |
| `ml_kem_ipd` | This pull request... | **Experimental** |

randombit:
https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.203.pdf

reneme:
Judging from appendix C.2, the only relevant addition is the domain separation in the seed expansion in Algorithm 13 (KeyGen). Basically, they just append a single byte `k = {2,3,4}` (depending on the parameter set) to the `G()` hash function. That's a minimal change.

falko-strenzke:
> Judging from appendix C.2, the only relevant addition is the domain separation in the seed expansion in Algorithm 13 (KeyGen). Basically, they just append a single byte `k = {2,3,4}` (depending on the parameter set) to the `G()` hash function. That's a minimal change.

That means in the ML-KEM ipd implementation the swapping of the matrix indices that happened in the ipd draft version were never implemented?

reneme:
> That means in the ML-KEM ipd implementation the swapping of the matrix indices that happened in the ipd draft version were never implemented?

I had taken that as an editorial error in the IPD document. When implementing the IPD version we aligned with [the reference implementation](https://github.com/pq-crystals/kyber/commit/11d00ff1f20cfca1f72d819e5a45165c1e0a2816) and I don't think we actually swapped these indices.

reneme:
The latest push contains a first round of adaptions (mostly updating the inline document references). As of now, I didn't come across any credible KATs to verify functional changes, albeit them being minimal.

# To do

* [ ] User-facing changes
  * [ ] rename `ml_kem_ipd` module
  * [ ] rename methods such as `mode.is_ml_kem_ipd`
  * [ ] rename `KyberMode` enum entries
  * [ ] ❓ Introduce `class ML_KEM_PublicKey` / `...PrivateKey`
         (and aliasing the existing `Kyber_...` classes for backward compatibility)
* [ ] internals
  * [ ] rename implementation classes (e.g. `ML_KEM_IPD_Encryptor`)
  * [ ] rename include guards
* [ ] Integrate updated KATs
* [ ] Incorporate the new domain separator in the seed calculation

reneme:
This patch is now updated to implement ML-KEM as specified in FIPS 203. I updated the original pull request description accordingly and noted a few TODOs before this can eventually go in. (The pull request is open for a long time and used to implement ML-KEM-ipd.)

@randombit Your insight on the open question regarding API (Kyber vs. ML-KEM) would be appreciated. Especially about your experience with the maintenance burden of aliases in the API for backward (or in this case: forward) compatibility.

My feeling is that we would have to alias essentially all definitions in the public `kyber.h`, namely:

```C++
namespace Botan {

class KyberMode;
class Kyber_PublicKey;
class Kyber_PrivateKey;

}
```

FiloSottile:
> * Add support for handling private keys as seeds rather than expanded structures (cc @FiloSottile)

💯 

> We already have `private_key_bits()` and `raw_private_key_bits()`, the latter could return the seed (if it is available). We'll somehow have to distinguish serialized from non-serialized keys for loading: simply by size? 🤔

Have you considered just not supporting the extended encoding? It's what I plan to do in Go. It solves the problem of how to distinguish them, and it means the seed is always available, so you don't have the problem of "what if someone loads an extended key and exports a seed".

reneme:
> Have you considered just not supporting the extended encoding?

I certainly understand the appeal of that idea, my only concern would (obviously) be interoperability with keys that were generated and exported by another implementation that didn't get the memo. :smirk: ... the standards are out now and contain the expanded format, after all. Nevertheless, one could make an argument that moving private keys around isn't such common use case for real-world systems anyway, I guess. And adding an interface for expanded keys in a later version is always possible. :thinking:

Perhaps @randombit has an opinion on this as well?

atreiber94:
Included the KATs from [Kris Kwiatkowski](https://github.com/post-quantum-cryptography/KAT) which were [verified](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/s_Wez9FanHw) against the PQShield implementations. They are working fine and were easy to integrate since they were already in the previous formats. I still want to also look into the ACVP tests since these are official.

FiloSottile:
> > Have you considered just not supporting the extended encoding?
> 
> I certainly understand the appeal of that idea, my only concern would (obviously) be interoperability with keys that were generated and exported by another implementation that didn't get the memo. 😏 

FWIW, it sounds like the IETF may specify seed encodings, judging from the conversation in the LAMPS WG.

Another reason to start with seeds is that if you internally you have seeds you can always add an API to export expanded keys, but not vice versa.

atreiber94:
ACVP tests are included now as well. I added them alongside the existing PQC KAT infrastructure since they do not seem to be generated by a provided seed.
This would also allow us to remove the "old" KATs when phasing out Kyber and solely rely on the ACVP formats for the finished standards since these are from NIST.

FiloSottile:
An update on the private key format: the IETF reached consensus on using seeds as the private key encoding. 
https://mailarchive.ietf.org/arch/msg/spasm/KPg3lbPeNms2H5PNDBt0iJByuE0/

src/tests/data/pubkey/ml_kem.vec
```diff
@@ -1,203 +1,203 @@
-# This file was auto-generated from the KATs at https://github.com/post-quantum-cryptography/KAT/tree/main/MLKEM
+# This file was auto-generated from the reference implemention's KATs
```
atreiber94, line 1: This change stems from the `gen_kyber_kat.py` but is not correct since these KATs are from the `post-quantum-cryptography/KAT` repository and not directly from the reference implementation.

doc/api_ref/pubkey.rst
```diff
@@ -120,14 +120,27 @@ Kyber
```
FAlbertDev, line 120: [Arbitrary code location]
pubkey.rst:1167
```rst
Botan implements the following KEM schemes:

1. RSA
#. Kyber
#. FrodoKEM
#. McEliece
```
We can now write something like `#. ML-KEM (Kyber)`

Also pubkey.rst:1176 
```rst
Code Example: Kyber
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

The code below demonstrates key encapsulation using the Kyber post-quantum scheme.

.. literalinclude:: /../src/examples/kyber.cpp
   :language: cpp
```
... can be adopted accordingly. Maybe we should change the Kyber example to an ML-KEM example and rename the file accordingly.

doc/api_ref/pubkey.rst
```diff
@@ -120,14 +120,27 @@ Kyber
...
+   third round of the PQC competition. It is not compatible with the
+   "Initial Public Draft" version of FIPS 203 for which Botan does not
+   offer an implementation.
+
+ML-KEM
+~~~~~~
 
-   The final NIST specification version of Kyber is not yet implemented.
+Post-quantum key encapsulation scheme based on (structured) lattices,
+as standardized in FIPS 203.
```
FAlbertDev, line 144: Arbitrary code location:

- credits.rst:166 can be updated with `ML-KEM/Kyber` instead of `Kyber`.
- FFI should not work for ML-KEM yet. But this is future work, I guess.

FAlbertDev (approved):
The changes look very good to me! I like the modular handling of seed-based private keys and the logic that differs between Kyber and ML-KEM. 
I've got only some minor/editorial comments :)

src/lib/pubkey/kyber/kyber_common/kyber.cpp
```diff
@@ -135,10 +164,14 @@ KyberMode Kyber_PublicKey::mode() const {
 }
 
 std::string Kyber_PublicKey::algo_name() const {
-   return "Kyber";
+   return mode().is_ml_kem() ? "ML-KEM" : "Kyber";
```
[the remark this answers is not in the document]
reneme: Yeah, I think we should make a distinction here. This is a public-facing algorithm name and it might lead to confusion in the wild.

src/lib/pubkey/kyber/kyber_common/kyber_keys.h
```diff
@@ -19,20 +19,18 @@
 
 namespace Botan {
 
+class Kyber_Keypair_Codec {
+   public:
+      virtual ~Kyber_Keypair_Codec() = default;
+      virtual secure_vector<uint8_t> encode_keypair(KyberInternalKeypair keypair) const = 0;
+      virtual KyberInternalKeypair decode_keypair(std::span<const uint8_t> private_key, KyberConstants mode) const = 0;
+};
```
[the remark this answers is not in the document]
reneme: It does need/produce the entire keypair though. I.e. it returns the public portion or consumes it.

src/lib/pubkey/kyber/kyber_common/kyber.h
```diff
@@ -21,8 +21,10 @@
 
 #include <span>
 
-#if !defined(BOTAN_HAS_KYBER_90S) && !defined(BOTAN_HAS_KYBER)
-static_assert(false, "botan module 'kyber_common' is useful only when enabling modules 'kyber', 'kyber_90s' or both");
+#if !defined(BOTAN_HAS_KYBER_90S) && !defined(BOTAN_HAS_KYBER) && !defined(BOTAN_HAS_ML_KEM)
+static_assert(
+   false,
+   "botan module 'kyber_common' is useful only when enabling at least one of those modules: 'kyber', 'kyber_90s', 'ml_kem'");
 #endif
```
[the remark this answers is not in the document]
reneme: Hosenträger-Gürtel-Prinzip. 😏 Doesn't hurt to have that explicitly in the code, I would say. E.g. when someone breaks or "hacks" `configure.py`.

src/lib/pubkey/kyber/kyber_common/kyber_symmetric_primitives.h
```diff
@@ -43,15 +43,28 @@ class Kyber_Symmetric_Primitives {
...
-      std::pair<KyberSeedRho, KyberSeedSigma> G(StrongSpan<const KyberSeedRandomness> seed) const {
-         return G_split<KyberSeedRho, KyberSeedSigma>(seed);
+      std::pair<KyberSeedRho, KyberSeedSigma> G(StrongSpan<const KyberSeedRandomness> seed,
+                                                const KyberConstants& mode) const {
+         return std::apply(
+            [this](auto&&... vals) {  // binds `this` for the call to G_split
+               return G_split<KyberSeedRho, KyberSeedSigma>(std::forward<decltype(vals)>(vals)...);
+            },
+            domain_separate_seed(seed, mode));
       }
```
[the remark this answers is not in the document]
reneme: Your's assumes the structure of the return value. Mine is more generic. That said: you're right. No need to be overly fancy here. I made the customization point return an optional and now it looks like this:

```C++
         if(auto domsep = seed_expansion_domain_separator(mode)) {
            return G_split<KyberSeedRho, KyberSeedSigma>(seed, *domsep);
         } else {
            return G_split<KyberSeedRho, KyberSeedSigma>(seed);
         }
```

src/tests/test_kyber.cpp
```diff
@@ -132,81 +140,111 @@ class Kyber_KAT_Tests final : public PK_PQC_KEM_KAT_Test {
...
+      }
+};
+
+}  // namespace
+
+BOTAN_REGISTER_TEST("kyber", "kyber_kat", KyberR3_KAT_Tests);
+BOTAN_REGISTER_TEST("kyber", "ml_kem_kat", ML_KEM_KAT_Tests);
+BOTAN_REGISTER_TEST("kyber", "ml_kem_acvp_kat_keygen", ML_KEM_ACVP_KAT_KeyGen_Tests);
+BOTAN_REGISTER_TEST("kyber", "ml_kem_acvp_kat_encap", ML_KEM_PQC_KEM_ACVP_KAT_Encap_Test);
+BOTAN_REGISTER_TEST("kyber", "ml_kem_acvp_kat_decap", ML_KEM_PQC_KEM_ACVP_KAT_Decap_Test);
```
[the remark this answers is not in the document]
reneme: Actually, we should probably register all new algorithms under `pubkey`. 😅

doc/api_ref/pubkey.rst
```diff
@@ -120,14 +120,27 @@ Kyber
...
```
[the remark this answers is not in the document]
reneme: Good idea. Done.

doc/api_ref/pubkey.rst
reneme: Lets update the credits independently! :)

Re:FFI: The FFI is tricky, actually. `botan_privkey_view_kyber_raw_key` and `botan_pubkey_view_kyber_raw_key` should just work with ML-KEM keys as well, I think.
For `botan_privkey_load_kyber` and `botan_pubkey_load_kyber` the idea of using the buffer length as an indicator for the required parameter set is coming to bite us. The public key lengths are now ambiguous between Kyber and ML-KEM. And the private key of ML-KEM is 64 bytes, regardless of the the parameter set. Probably we should just add new methods `botan_*_mlkem_*()` and deprecate the kyber variants in due time.

doc/api_ref/pubkey.rst
reneme: Done. Also renamed the example to `ml_kem.cpp`.

src/lib/pubkey/kyber/ml_kem/ml_kem_impl.cpp
```diff
@@ -0,0 +1,81 @@
...
+                                   RandomNumberGenerator& rng) {
+   const auto& sym = m_public_key->mode().symmetric_primitives();
+
+   const auto m = rng.random_vec<KyberMessage>(KyberConstants::SEED_BYTES);
+   const auto [K, r] = sym.G(m, m_public_key->H_public_key_bits_raw());
+   m_public_key->indcpa_encrypt(out_encapsulated_key, m, r, precomputed_matrix_At());
+
+   // TODO: avoid this copy by letting sym.G() directly write to the span.
+   copy_mem(out_shared_key, K);
+}
```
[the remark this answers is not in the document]
reneme: Good catch!

doc/api_ref/pubkey.rst
reneme: Re:FFI, see here: https://github.com/randombit/botan/pull/3893/commits/06c817558ce72495225e80aee515c271f716e6c3

src/lib/ffi/ffi_pkey_algs.cpp
```diff
@@ -1027,6 +1028,96 @@ int botan_pubkey_view_kyber_raw_key(botan_pubkey_t key, botan_view_ctx ctx, bota
...
+   BOTAN_UNUSED(key, ctx, view);
+   return BOTAN_FFI_ERROR_NOT_IMPLEMENTED;
+#endif
+}
+
+int botan_pubkey_view_ml_kem_raw_key(botan_pubkey_t key, botan_view_ctx ctx, botan_view_bin_fn view) {
+#if defined(BOTAN_HAS_ML_KEM)
+   return BOTAN_FFI_VISIT(key, [=](const auto& k) -> int {
+      if(auto mlkem_key = dynamic_cast<const Botan::Kyber_PublicKey*>(&k)) {
+         return invoke_view_callback(view, ctx, mlkem_key->public_key_bits());
```
FAlbertDev, line 1110:
```suggestion
         return invoke_view_callback(view, ctx, mlkem_key->raw_public_key_bits());
```
(I know both methods do the same)

src/lib/pubkey/kyber/kyber_common/kyber.cpp
```diff
@@ -252,6 +262,27 @@ secure_vector<uint8_t> Kyber_PrivateKey::private_key_bits() const {
...
+      return false;
+   }
+
+   PK_KEM_Encryptor enc(*this, "Raw");
+   PK_KEM_Decryptor dec(*this, rng, "Raw");
+
+   const auto [c, K] = KEM_Encapsulation::destructure(enc.encrypt(rng));
+   const auto K_prime = dec.decrypt(c);
+
+   return K == K_prime;
```
FAlbertDev, line 283: You want to check that the public key is consistent with the private key, don't you?  Isn't this only necessary for Kyber R3, since ML-KEM uses seeds?

reneme:
@randombit This is now ready for final review. It would be amazing to have this merged before the 3.6.0 release. Please let us know if we can help in some way.

Left for future work:

* **Clean-up of the KATs**, hopefully we'll converge on some official test vector, pending availability
* **API Future-Proofing**, currently the Public API is called `Kyber_***` and we should probably provide `ML_KEM_***` aliases (see also the original PR description)

Note:

The commits that introduce `Bounded_XOF` and that alter the `PK_Keygen_Test` class are repeated in the ML-DSA PR. We'll rebase and resolve conflicts once one of them got merged.

src/tests/test_kyber.cpp
```diff
@@ -108,18 +108,26 @@ class KYBER_Tests final : public Test {
          results.push_back(run_kyber_test("Kyber768 API", Botan::KyberMode::Kyber768_R3, 192, 768));
          results.push_back(run_kyber_test("Kyber1024 API", Botan::KyberMode::Kyber1024_R3, 256, 1024));
    #endif
+   #if defined(BOTAN_HAS_ML_KEM)
+         results.push_back(run_kyber_test("ML-KEM-512 API", Botan::KyberMode::ML_KEM_512, 128, 512));
+         results.push_back(run_kyber_test("ML-KEM-768 API", Botan::KyberMode::ML_KEM_768, 192, 768));
+         results.push_back(run_kyber_test("ML-KEM-1024 API", Botan::KyberMode::ML_KEM_1024, 256, 1024));
+   #endif
```
FAlbertDev, line 116: Maybe resolve these ifdefs using KyberMode::is_available

FAlbertDev (commented):
The changes since my last review look good to me 👍 
I left some comments

src/lib/pubkey/kyber/kyber_common/kyber.cpp
reneme: Essentially, that's what FIPS 203 requests us to do to "check the keys". But you are totally right: This is not needed when extracting the private key from the seeds.

src/tests/test_kyber.cpp
reneme: Good idea! Thanks.

reneme:
Work on it was finished on Friday, it is now awaiting review. So feature-wise it's ready to go.

Feel free to pull it into a local build if you want to start prototyping with ML-KEM. Any feedback is appreciated.

FAlbertDev (approved):

tobiasbrunner:
While trying to integrate these changes into our Botan plugin I noticed (see https://github.com/strongswan/strongswan/issues/2228) that these changes allocate the seed for the key pair in a different order, first d and then z instead of the other way around, which is what FIPS-203 describes and other implementations do (and e.g. the referenced KAT vectors do as well). So it's quite awkward to test this implementation using the API. Is there a specific reason this order was chosen?

reneme:
>  [...] these changes allocate the seed for the key pair in a different order, first d and then z instead of the other way around, which is what FIPS-203 describes and other implementations

For reference: I added my view on this [here](https://github.com/strongswan/strongswan/issues/2228#issuecomment-2399196433). TL;DR: In my opinion, downstream users should not pose assumptions on the RNG invocation pattern of any implementation of ML-KEM. For testing with KAT vectors, they should prepare a concatenation of `d || z` taken from the KAT and pass those 64 bytes into the "private key loading procedure". For Botan that would be: `Botan::Kyber_PrivateKey(d_and_z, Botan::KyberMode("ML-KEM-512"))`.

>  Is there a specific reason this order was chosen?

Not really. We followed the order of operations as stated in FIPS 203, Algorithm 19.

tobiasbrunner:
> For testing with KAT vectors, they should prepare a concatenation of `d || z` taken from the KAT and pass those 64 bytes into the "private key loading procedure". For Botan that would be: `Botan::Kyber_PrivateKey(d_and_z, Botan::KyberMode("ML-KEM-512"))`.

OK, thanks. That sounds like a way to handle this. I've successfully used `botan_privkey_load_ml_kem()` with my original test vectors that assume a 64-byte pull from the DRBG.

> Not really. We followed the order of operations as stated in FIPS 203, Algorithm 19.

Sorry about that. I read the order in FIPS-203 wrong for some reason, it's definitely d before z (also for the inputs to Algorithm 16). So it's more the referenced KAT vectors, which allocate z before d from the DRBG, that made this awkward.

src/lib/ffi/ffi_pkey_algs.cpp
```diff
@@ -1027,6 +1028,96 @@ int botan_pubkey_view_kyber_raw_key(botan_pubkey_t key, botan_view_ctx ctx, bota
...
+      return BOTAN_FFI_SUCCESS;
+   });
+#else
+   BOTAN_UNUSED(key, key_len, privkey, mlkem_mode);
+   return BOTAN_FFI_ERROR_NOT_IMPLEMENTED;
+#endif
+}
+
+int botan_privkey_view_ml_kem_raw_key(botan_privkey_t key, botan_view_ctx ctx, botan_view_bin_fn view) {
+#if defined(BOTAN_HAS_ML_KEM)
```
randombit, line 1092: Could we instead just define these generically for any key? Since `raw_public_key_bits` and `raw_private_key_bits` are available for a range of algorithms, and it avoids (further) algorithm specific functions. I'd rather we in fact implement the generic case and deprecate the algorithm specific ones. Discussion in #4366

reneme: Yeah. Makes sense for sure. I'd propose to just remove the ML-KEM specific one from this pull request and add the generic one independently.

src/lib/ffi/ffi_pkey_algs.cpp
```diff
@@ -1027,6 +1028,96 @@ int botan_pubkey_view_kyber_raw_key(botan_pubkey_t key, botan_view_ctx ctx, bota
 #endif
 }
 
+/*
+* Algorithm specific key operations: ML-KEM
+*/
+
+int botan_privkey_load_ml_kem(botan_privkey_t* key, const uint8_t privkey[], size_t key_len, const char* mlkem_mode) {
+#if defined(BOTAN_HAS_ML_KEM)
+   *key = nullptr;
```
reneme, line 1037: Check before dereferencing!

reneme:
Rebased after #4367 and #4372 caused conflicts.

src/lib/pubkey/kyber/kyber_common/kyber_keys.h
```diff
@@ -60,37 +58,49 @@ class Kyber_PublicKeyInternal {
...
+      const KyberImplicitRejectionValue& z() const { return m_seed.z; }
 
       const KyberConstants& mode() const { return m_mode; }
 
       Kyber_PrivateKeyInternal() = delete;
 
-      void _const_time_poison() const { CT::poison_all(m_s, m_z); }
+      void _const_time_poison() const {
+         CT::poison_all(m_s, m_seed.z);
+         if(m_seed.d.has_value()) {
```
randombit, line 88: We might want to consider a variant of `poison` that takes an `optional<T>` and poisons it if it has a value, as a convenience. (No need to do so in this PR)

src/lib/pubkey/kyber/kyber_common/kyber_symmetric_primitives.h
```diff
@@ -95,8 +104,12 @@ class Kyber_Symmetric_Primitives {
...
 
    protected:
+      virtual std::optional<std::array<uint8_t, 1>> seed_expansion_domain_separator(
```
randombit, line 107: Why the `array` and not just `std::optional<uint8_t>`?

randombit: Oh maybe this just makes life easier wrt `concat`?

randombit (approved):

src/lib/pubkey/kyber/kyber_common/kyber_symmetric_primitives.h
reneme: Yep. That's the reason. I did look into extending concat to take scalar values along with ranges but didn't find an easy enough way to allow that.

We might want to look into that later though.

src/lib/pubkey/kyber/kyber_common/kyber_keys.h
reneme: Sounds like a good idea. I'll look at this independently.

PR #3898, opened 2024-01-24
Kyber constant time division

FAlbertDev:
### Pull request dependencies
* #3887 

This PR adds logic to perform constant-time division by replacing division with a multiplication and right-shift. The constants for division and right-shift are computed at compile time using an algorithm from Hacker's Delight, Chapter 10 - 9.


reneme:
Perhaps describe the counter measure in side_channel.rst?

reneme:
Rebased and resolved conflicts after #3874 was merged and #3887 was adapted to it.

randombit (commented):
TBH I don’t see how this helps keep the compiler from turning the multiplication into a div, since the multiplicand will anyway be known at compile time.

One thing I find interesting here is that after the (comple time) computation of `pm`, the final result is computed using just a multiplication and a single shift. Wheras the current approach uses a multiplication and several shifts, add/sub, etc. If we just took the value that would be computed for `pm` for the Kyber q and used it directly (`return (n * MAGIC) >> MAGIC_SHIFT`), is there really any risk of variable time operations being compiled in? I would think no (non-malicous) compiler would convert 2 instructions into a div or jump.

Issue #3896, opened 2024-01-22
`test_compression` fails on the latest Fedora Rawhide


mrc0mmand:
Hey!

A while ago I got a notification that the `botan2` package fails to build on Fedora Rawhide due to failing `make check`, and while trying to finally address it I noticed that it is reproducible with the latest `master` branch as well:

```
$ podman run -it --rm fedora:rawhide
# dnf install -y git dnf-plugins-core
# dnf builddep -y botan2
# git clone https://github.com/randombit/botan --depth=1
# cd botan
# ./configure.py --enable-modules=bzip2,zlib
# make -j $(nproc)
# make check |& tee log.txt
...
ZFEC encoding/decoding ran 10449 tests in 6.43 msec all ok
Tests complete ran 2938384 tests in 22.04 sec 3 tests failed (in compression_tests)
ERROR:root:Error running ./botan-test --data-dir=./src/tests/data
make: *** [Makefile:65: check] Error 1
# grep "Failure" log.txt
Failure 1: deflate compression Level 9 compresses text at least as well as level 1 unexpected result 443 < 444 (at src/tests/test_compression.cpp:164)
Failure 1: gzip compression Level 9 compresses text at least as well as level 1 unexpected result 461 < 462 (at src/tests/test_compression.cpp:164)
Failure 1: zlib compression Level 9 compresses text at least as well as level 1 unexpected result 449 < 450 (at src/tests/test_compression.cpp:164)
```

reneme:
Admittedly, that test seems quite opinionated on how a compression library should work and how its performance characteristics should look like. Botan just wraps the compression libraries and the test should probably just make sure that this wrapper works as expected.

From what I can tell, we could just remove the "opinionated" portion of this test. But perhaps, @randombit could add some context on that?

randombit:
I guess this is opinionated but I am surprised that level 1 compression compresses *BETTER THAN* level 9 compression. This kind of implies we should just use level 1 and save the extra CPU costs ....

mrc0mmand:
> I guess this is opinionated but I am surprised that level 1 compression compresses _BETTER THAN_ level 9 compression. This kind of implies we should just use level 1 and save the extra CPU costs ....

Yeah, I find this really suspicious, and I still feel like there's a bug somewhere else. But it's weird that it fails the same way with three different compression methods.

mrc0mmand:
[...]

PR #3900, opened 2024-01-31
FIX: False Positive for Compression Test 

reneme:
This test recently started failing on Fedora Rawhide because of some `zlib-ng` constellation where Level 1 could actually compress (slightly) better than Level 9. See #3896 and https://github.com/zlib-ng/zlib-ng/issues/1656 for details.

Here's how to reproduce this via docker:

```bash
cd my_botan_source_dir

docker run -v .:/botan -it fedora:rawhide bash

# now inside the container
yum install g++ ninja-build bzip2-devel xz-devel zlib-ng-compat-devel
cd /botan
./configure.py --with-zlib --with-lzma --with-bzip2 --disable-shared --build-tool=ninja --minimized-build
ninja
./botan-test compression_tests
```


randombit (approved):
Thanks

PR #3905, opened 2024-02-13
Add documented botan_cipher_is_authenticated() to FFI

reneme:
This function is documented since 2018 but was actually not implemented in the Foreign Function Interface. It's merely a convenience function for `botan_cipher_get_tag_length()` returning a value greater than zero, though.

### DIFF
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -539,6 +539,11 @@ BOTAN_FFI_EXPORT(2, 0) int botan_cipher_valid_nonce_length(botan_cipher_t cipher
 BOTAN_FFI_EXPORT(2, 0) int botan_cipher_get_tag_length(botan_cipher_t cipher, size_t* tag_size);
 
 /**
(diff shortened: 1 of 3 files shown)


src/lib/ffi/ffi.h
```diff
@@ -538,6 +538,11 @@ BOTAN_FFI_EXPORT(2, 0) int botan_cipher_valid_nonce_length(botan_cipher_t cipher
 */
 BOTAN_FFI_EXPORT(2, 0) int botan_cipher_get_tag_length(botan_cipher_t cipher, size_t* tag_size);
 
+/**
+* Returns 1 iff the cipher provides authentication as well as confidentiality.
+*/
+BOTAN_FFI_EXPORT(3, 2) int botan_cipher_is_authenticated(botan_cipher_t cipher);
```
randombit, line 544: Should be 3,3

randombit (approved):
Wonder what happened here :/

PR #3908, opened 2024-02-16
Remove workarounds for compilers that didn't implement C++20 concepts

randombit:


reneme (approved):
Uh! I've been procrastinating on that. Now you've beat me to it. Thanks!

PR #3909, opened 2024-02-19
Use binary GCD

reneme:
Judging from the runtime of the gcd test vector, this is faster than the existing implementation. It should also improve the resilience against timing side channels.


randombit (approved):

randombit:
I'm working on a backport to 2.x

Issue #3853, opened 2023-12-22
Need support for module deprecation


randombit:
A bit of discussion at the end of #3807 about this.

One question is how a deprecated module behaves. My take

* If a deprecated module is being built, we warn about it
* It continues to be included by default, you can explicitly request it if you want, etc. Basically its status as deprecated does not affect module selection.
* There is a new option (`--disable-deprecated-features`?) that disables them, and which possibly also sets a macro that can be used to selectively disable other deprecated features (eg 64 bit GCM tags).

reneme:
This is probably useful, especially when paired with the `--disable-deprecated-features` flag. Otherwise, I'm not exactly convinced whether a warning during configuration is prominent enough for downstream users to actually notice.

While we're discussing to introduce a notion of "lifecycle" for modules: let me bring up [this idea from last year](https://github.com/randombit/botan/discussions/3110). I.e. flagging modules as "experimental". The behaviour of "experimental" modules would likely be similar to "deprecated" modules. Except that they might not be enabled by default and hence require an `--enable-experimental-features` flag.

randombit:
It would be easy to introduce both of these at the same time.

PR #3910, opened 2024-02-20
Feature: Module Lifecycle and Experimental/Deprecated Features

reneme:
This aims at implementing the suggestions in #3110 and #3853. Namely, it introduces a "lifecycle" to Botan's module system. In their `info.txt` file, individual modules may be flagged as "Experimental" or "Deprecated" ("Stable" is the default), like so:

```
<defines>
MY_MODULE -> 20240220
</defines>

<module_info>
name -> "My fancy new module"
brief -> "A new feature that isn't fully ready for prime time, yet."
lifecycle -> "Experimental"
</module_info>
```

### Lifecycle

| State | Meaning |
| --- | --- |
| "Stable" | The (automatic) default. Today, all modules are considered "Stable", essentially. Minor releases may not change the API contract. |
| "Deprecated" | Such a module behaves exactly the same as "Stable", though it is scheduled for removal in a future major release. While being built by default, `./configure.py` will subtly warn about the deprecated modules that go into the build. |
| "Experimental" | **Work in progress**. API and functionality might not be stable; implementation quality may be "beta". These modules are not built by default and `./configure.py` warns about any experimental modules that go into the build. |

### Configuration

This adds a few switches to `./configure.py` to handle modules (and features) that are not "Stable":

* `--enable-experimental-features` defines `BOTAN_ENABLE_EXPERIMENTAL_FEATURES` in the code base and enables all experimental modules.
* `--disable-experimental-features` (the default)
* `--enable-deprecated-features` (the default)
* `--disable-deprecated-features` defines `BOTAN_DISABLE_DEPRECATED_FEATURES` in the code base and disables all deprecated modules

Additionally, the default build behaviour of experimental and deprecated modules may be explicitly changed using `--enable-modules=MODS` and `--disable-modules=MODS`.

Dependants of a disabled experimental/deprecated module won't be built (due to dependency failure), but they won't be transitively marked as experimental/deprecated. Library maintainers have to keep this in mind when changing modules' life cycle state.

### Usage

This just introduces the feature but does not mark any modules/features as deprecated or experimental. A good first candidate for "experimental" might be ML-KEM-ipd (#3893). In retrospect, TLS with hybrid PQ/T key exchange (#3609) would have been a good candidate for an experimental module. Soon, Kyber 90s could become the first "deprecated" module: https://github.com/randombit/botan/pull/3807#issuecomment-1867376879.

Also, the macros `BOTAN_ENABLE_EXPERIMENTAL_FEATURES` and `BOTAN_DISABLE_DEPRECATED_FEATURES` are just introduced but not used, yet. For instance, GCM tags with just 64bits could become a deprecated feature (that is not a full module) at one point.

### Documentation

The new switches and the life cycle states are mentioned in the relevant locations of the handbook. Also, the generated Doxygen pages for the modules contain information about the module's life cycle status, if they are not "Stable", like so:

image

image

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -385,6 +385,16 @@ def process_command_line(args):
     target_group.add_option('--without-os-features', action='append', metavar='FEAT',
                             help='specify OS features to disable')
 
+    target_group.add_option('--enable-experimental-features', dest='enable_experimental_features',
+                            action='store_true', default=False, help='enable building of experimental features and modules')
+    target_group.add_option('--disable-experimental-features', dest='enable_experimental_features',
+                            action='store_false', help=optparse.SUPPRESS_HELP)
+
+    target_group.add_option('--enable-deprecated-features', dest='enable_deprecated_features',
+                            action='store_true', default=True, help=optparse.SUPPRESS_HELP)
+    target_group.add_option('--disable-deprecated-features', dest='enable_deprecated_features',
+                            action='store_false', help='disable building of deprecated features and modules')
+
     isa_extensions = [
         'SSE2', 'SSSE3', 'SSE4.1', 'SSE4.2', 'AVX2', 'BMI2', 'RDRAND', 'RDSEED',
         'AES-NI', 'SHA-NI',
@@ -2290,6 +2300,9 @@ def create_template_vars(source_paths, build_paths, options, modules, disabled_m
         'os_name': osinfo.basename,
         'cpu_features': arch.supported_isa_extensions(cc, options),
         'system_cert_bundle': options.system_cert_bundle,
+
+        'enable_experimental_features': options.enable_experimental_features,
+        'disable_deprecated_features': not options.enable_deprecated_features,
 
         'fuzzer_mode': options.unsafe_fuzzer_mode,
         'building_fuzzers': options.build_fuzzers,
--- a/configure.py
+++ b/configure.py
@@ -951,9 +951,12 @@ class ModuleInfo(InfoObject):
         self.name = info["name"]
         self.brief = info.get("brief") # possibly None
         self.type = info.get("type") or "Public"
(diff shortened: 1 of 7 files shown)


randombit (approved):
One thing I’m not completely sure about - and this is a policy issue related to how we’ll use this, rather than the change itself - previously (or since 2.0.0, anyway) there was a clear distinction about when any functionality could be removed, namely in a new major version. Consider we introduce experimental module X, then decide it doesn’t work out. What is the procedure for making it clear this is something that “was” experimental, but is going to be removed soon? We can’t mark it as deprecated - that causes it to start building by default! I guess under this contract we can just yank it without notice. I suppose this can be decided on a case by case basis as we go forward.

Please update `sem_ver.rst` to make it clear experimental features are not covered by any SemVer guarantee. Otherwise lgtm.

reneme:
>  What is the procedure for making it clear this is something that “was” experimental, but is going to be removed soon?

Mhm, fair point. But frankly, another state (e.g. "deprecated experimental") feels somewhat overkill for this use case. At least for adding it right now. In any case, if we'd really need such a lifecycle state for some module in the future, we could always just add it to the build system later on.

Regarding communication: We could make it a policy to open a discussion on GitHub about every experimental feature. This could be used for early feedback and also as a platform for the deprecation notice. To make this discoverable, we might even add another entry to the `<module_info>` (e.g. `feedback_link`, or perhaps something more general). Then, if someone builds with `--enable-experimental-features` we could print a list of all experimental modules with these feedback links in the `configure.py`-run.

I would consider that future-work, though.

reneme:
> Please update sem_ver.rst to make it clear experimental features are not covered by any SemVer guarantee.

Done. Ready for merge from my side.

PR #3915, opened 2024-02-20
Avoid including build.h into ffi.h

randombit:
This leaves ffi.h as completely free standing

### DIFF
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -56,14 +56,28 @@ API follows a few simple rules:
  - TLS
 */
 
-#include <botan/build.h>
 #include <stddef.h>
 #include <stdint.h>
 
 /**
-* Notes the version that this FFI function was first added
+* BOTAN_FFI_EXPORT indicates public FFI functions.
+*
+* The arguments to the macro are to indicate the version that
+* that particular FFI function was first available
 */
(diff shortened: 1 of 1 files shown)


reneme:
`./configure.py` sets the `BOTAN_DLL` macro to the empty string for the static build.

```python
    def gen_visibility_attribute(self, options):
        if options.build_shared_lib:
            return self.visibility_attribute
        return ''
```

reneme:
I'm reasonably sure that, [this](https://github.com/randombit/botan/pull/3915/commits/90da35e9a792d64aae0116ba98d7fce85c73dfb8) should work on Windows. When building a shared lib, `BOTAN_DLL` is defined (as `__declspec(dllexport)`). When building a static lib `BOTAN_DLL` isn't defined, but `BOTAN_IS_BEING_BUILT` is. Hence, the `BOTAN_FFI_EXPORT` defaults to the empty string.

When consuming the library, neither `BOTAN_DLL` nor `BOTAN_IS_BEING_BUILT` is defined, so `BOTAN_FFI_EXPORT` defaults to `__declspec(dllimport)`; which should be fine for both shared and static linkage.

reneme (approved):

randombit:
Squashed and rebased

Issue #3916, opened 2024-02-21
[2.19.4] Sporadic test failure with `_GLIBCXX_ASSERTIONS`


thesamesam:
Hi all!

When running tests for Botan 2.19.4 when packaging for Gentoo, I got:
```
bcrypt:
bcrypt ran 380 tests in 2.33 sec all ok
bcrypt_pbkdf:
/usr/lib/gcc/x86_64-pc-linux-gnu/14/include/g++-v14/bits/stl_vector.h:1127: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = unsigned char; _Alloc = Botan::secure_allocator<unsigned char>; reference = unsigned char&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Aborted (core dumped)
```

It only happens on some runs, unfortunately.

Backtraces (hit this with gcc 13 too):
```
Core was generated by `./botan-test2 --test-threads=32'.
Program terminated with signal SIGABRT, Aborted.
#0  0x00007f403f2bec7c in ?? () from /usr/lib64/libc.so.6
[Current thread is 1 (Thread 0x7f4032a136c0 (LWP 2878430))]
(gdb) bt
#0  0x00007f403f2bec7c in ?? () from /usr/lib64/libc.so.6
#1  0x00007f403f268fc2 in raise () from /usr/lib64/libc.so.6
#2  0x00007f403f2514f2 in abort () from /usr/lib64/libc.so.6
#3  0x00007f403f4e4dfb in std::__glibcxx_assert_fail (file=file@entry=0x7f403fdac4f0 "/usr/lib/gcc/x86_64-pc-linux-gnu/14/include/g++-v14/bits/stl_vector.h", line=line@entry=1127,
    function=function@entry=0x7f403fdacee0 "std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = unsigned char; _Alloc = Botan::secure_allocator<unsigned char>; reference = unsigned char&; size_type = "..., condition=condition@entry=0x7f403fda401a "__n < this->size()")
    at /usr/src/debug/sys-devel/gcc-14.0.9999/gcc-14.0.9999/libstdc++-v3/src/c++11/assert_fail.cc:41
[...]

PR #3918, opened 2024-02-21
Fix an iterator debugging error in TLS CBC decryption

randombit:
GH #3916


reneme (approved):
My assumption is, that this catches the case where both `offset` and `msg_size` are 0, hence the `buffer.resize(0)` empties the vector, so that `buffer[0]` is actually wrong.

randombit:
This should probably be ported also into 3.x since the same code is there - though for whatever reason I was unable to repro the crash there. I'm not sure why.

Issue #3917, opened 2024-02-21
[3.3.0] Sporadic dilithium_kat_4x4_AES_Deterministic failure


thesamesam:
I was trying to reproduce https://github.com/randombit/botan/issues/3916 with 3.3.0 but not yet succeeded.

I can reproduce with both:
* `gcc (Gentoo Hardened 14.0.9999 p, commit 568b3f9940f87a775060aea789b2f438af477e9f) 14.0.1 20240218 (experimental) e63ae9085aca9d306a2f16445b473289b9186e10`
* `gcc-13 (Gentoo Hardened 13.2.1_p20240217 p13) 13.2.1 20240217`

I did however hit another failure once in many runs:
```
Testing Botan 3.3.0 (release, dated 20240220, revision git:9074b04c1303a24e2084f8325fa570a5ad4f2478, distribution Gentoo 3.3.0)
Properties:
  CPU flags: rdtsc sse2 ssse3 avx2 bmi2 adx aes_ni clmul rdrand rdseed intel_sha
  drbg_seed: 17B5D570B30DF7D1
[...]
dilithium_kat_4x4_AES_Deterministic:
Dilithium_4x4_AES ran 601 tests in 92.28 msec 4 FAILED
Failure 1: Dilithium_4x4_AES unexpected result for generated expected private key hash
Produced: B882DEC292A05E8890222097C4C00FD9774B49E0BDBE981EBB9F8371A181CE6C
Expected: BD0C8516566A958D12A458E7A1B5AC2763208AA9988BEC64E3A695DD12D93DD8
XOR Diff: 058E5BD4C4CACB05828678706575A3FE146BC3492535747A583916ACB358F3B4 (at src/tests/test_dilithium.cpp:101)
Failure 2: Dilithium_4x4_AES unexpected result for generated expected public key hash
Produced: F40020D52D757DA856F42F7FFE03E7C78F4F9ECC5BF3BCBFA79713EE82D5BDB5
Expected: 1ECBF9C21C96E7C287901D8853995FC2EDBE0A86680E42A0747E2DEB4BD11056
XOR Diff: EACBD91731E39A6AD16432F7AD9AB80562F1944A33FDFE1FD3E93E05C904ADE3 (at src/tests/test_dilithium.cpp:101)
Failure 3: Dilithium_4x4_AES unexpected result for generated expected signature hash
Produced: 3D00CA1AA521DE5FEECE2238E6618287854A38DE269D2EF46AB8075A6EFF3C3D
[...]

PR #3919, opened 2024-02-22
Chore: Check Serialized Tests

reneme:
This aborts `Text_Based_Test`s when it tries to mangle with the global `CPUID` settings but was not marked "serialized".
There's other global-state that requires serialization and could benefit from similar checks. @randombit could you braindump some locations? :)

Obviously, this approach is flawed, but perhaps it saves some headaches in the future. I'm hoping that this helps with #3917.

### DIFF
--- a/src/tests/test_utils.cpp
+++ b/src/tests/test_utils.cpp
@@ -635,14 +635,18 @@ class CPUID_Tests final : public Test {
 #if defined(BOTAN_TARGET_CPU_IS_X86_FAMILY)
 
          if(Botan::CPUID::has_sse2()) {
-            result.confirm("Output string includes sse2", cpuid_string.find("sse2") != std::string::npos);
+            if(!this->is_serialized()) {
+               result.test_failure("Cannot proceed, test is not serialized");
+            } else {
+               result.confirm("Output string includes sse2", cpuid_string.find("sse2") != std::string::npos);
 
-            Botan::CPUID::clear_cpuid_bit(Botan::CPUID::CPUID_SSE2_BIT);
+               Botan::CPUID::clear_cpuid_bit(Botan::CPUID::CPUID_SSE2_BIT);
 
-            result.test_eq("After clearing cpuid bit, has_sse2 returns false", Botan::CPUID::has_sse2(), false);
+               result.test_eq("After clearing cpuid bit, has_sse2 returns false", Botan::CPUID::has_sse2(), false);
 
-            Botan::CPUID::initialize();  // reset state
-            result.test_eq("After reinitializing, has_sse2 returns true", Botan::CPUID::has_sse2(), true);
+               Botan::CPUID::initialize();  // reset state
+               result.test_eq("After reinitializing, has_sse2 returns true", Botan::CPUID::has_sse2(), true);
+            }
          }
 #endif
 
@@ -650,7 +654,7 @@ class CPUID_Tests final : public Test {
       }
 };
 
-BOTAN_REGISTER_TEST("utils", "cpuid", CPUID_Tests);
+BOTAN_REGISTER_SERIALIZED_TEST("utils", "cpuid", CPUID_Tests);
 
 #if defined(BOTAN_HAS_UUID)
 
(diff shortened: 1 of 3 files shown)


src/tests/tests.h
```diff
@@ -566,6 +566,10 @@ class Test {
 
       const std::optional<CodeLocation>& registration_location() const { return m_registration_location; }
 
+      void set_serialization(bool needs_serialization) { m_needs_serialization = needs_serialization; }
```
randombit, line 569: This feels weird in two ways

- Serialization state can nominally change (eg a `Test` can call `set_serialization(true)` then `set_serialization(false)`)
- A `Test` can set serialization late (after initialization)

I think both are remedied by making it an argument to the (newly added) `Test::initialize`

src/tests/test_utils.cpp
```diff
@@ -635,22 +635,26 @@ class CPUID_Tests final : public Test {
...
+            }
          }
 #endif
 
          return {result};
       }
 };
 
-BOTAN_REGISTER_TEST("utils", "cpuid", CPUID_Tests);
+BOTAN_REGISTER_SERIALIZED_TEST("utils", "cpuid", CPUID_Tests);
```
randombit, line 657: Very good odds imo that this is the cause of #3917

src/tests/tests.cpp
```diff
@@ -1060,6 +1060,9 @@ std::vector<Test::Result> Text_Based_Test::run() {
          }
 
          m_cpu_flags = parse_cpuid_bits(pragma_tokens);
+         if(!m_cpu_flags.empty() && !this->is_serialized()) {
+            throw Test_Error(Botan::fmt("'{}' used cpuid control but is not serialized", this->test_name()));
+         }
```
randombit, line 1065: Isn't this redundant with the prior check just before this block?

randombit: (If not, there should be a comment explaining why both are necessary)

src/tests/test_utils.cpp
```diff
@@ -635,22 +635,26 @@ class CPUID_Tests final : public Test {
 #if defined(BOTAN_TARGET_CPU_IS_X86_FAMILY)
 
          if(Botan::CPUID::has_sse2()) {
-            result.confirm("Output string includes sse2", cpuid_string.find("sse2") != std::string::npos);
+            if(!this->is_serialized()) {
```
randombit, line 638: Is this necessary? Shouldn't we (if we registered as serialized) *always* be serialized?

src/tests/tests.h
```diff
@@ -566,6 +566,10 @@ class Test {
...
+
+      bool is_serialized() const { return m_needs_serialization; }
```
randombit, line 571: When would a test need to check if it was serialized or not? Since this is set (effectively) at runtime based on which macro is invoked.

randombit (commented):
I'm somewhat confused by some of these changes.

Serializing the `cpuid` test seems right, and is likely the fix for #3917.

The rest though seems redundant with other checks and/or unnecessary - what am I missing?

randombit:
Other global state to watch for

* mlock allocator (probably ok, modulo performance issues with sharing the pool among all threads)
* The system RNG handle (ditto)
* FFI error reporting (but this is `thread_local`) so probably ok
* The internal cached `EC_Group` state - mostly set up once at runtime but can be modified to eg add a new curve, so could cause problems.

src/tests/test_utils.cpp
reneme: That's what I thought. 🤞

src/tests/tests.h
reneme: Agreed.

reneme:
Closing this, and re-opening with just the potential fix for #3917.

This whole thing was a knee-jerk idea to "make sure that tests are serialized properly". An idea you had implemented more than half ago, [and I even liked it](https://github.com/randombit/botan/pull/3625#discussion_r1261114374). I didn't bother reading the code left and right of my changes. My bad. 😊

Nevertheless, it would be great to have a similar safe-guard for the `EC_Group` registration. (Or, perhaps, just wrap this registration with a mutex in the library?)

PR #3920, opened 2024-02-22
Remove the global RNG from the tests

randombit:
Instead tests instantiate a new RNG for their own use. This prevents non-determinism from creeping in due to thread interleavings altering the output of the test RNG.

This PR doesn't quite remove the global test RNG - it's still used in `test_tls_hybrid_kem_key.cpp` and `test_tls_rfc8448.cpp`. I think those tests may need some substantial refactoring to avoid using the global RNG.

### DIFF
--- a/src/tests/tests.h
+++ b/src/tests/tests.h
@@ -638,7 +638,6 @@ class Test {
       static std::unique_ptr<Botan::RandomNumberGenerator> new_rng(std::string_view test_name);
       static std::shared_ptr<Botan::RandomNumberGenerator> new_shared_rng(std::string_view test_name);
 
-      static Botan::RandomNumberGenerator& global_rng();
       static std::string random_password(Botan::RandomNumberGenerator& rng);
       static uint64_t timestamp();  // nanoseconds arbitrary epoch
 
@@ -646,7 +645,6 @@ class Test {
 
    private:
       static Test_Options m_opts;
-      static std::shared_ptr<Botan::RandomNumberGenerator> m_global_test_rng;
       static std::string m_test_rng_seed;
 
       /// The string ID that was used to register this test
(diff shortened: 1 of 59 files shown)


reneme:
> This PR doesn't quite remove the global test RNG - it's still used in test_tls_hybrid_kem_key.cpp and test_tls_rfc8448.cpp. I think those tests may need some substantial refactoring to avoid using the global RNG.

Done. In `test_tls_rfc8448.cpp` the only location using the global RNG needed it for API-pleasing only -- `Null_RNG` is enough. For `test_tls_hybrid_kem_key.cpp` it was indeed more tricky. I ended up replacing the "global global" with a lazily instantiated "test global" RNG for now.

src/tests/tests.h
```diff
@@ -664,8 +668,7 @@ class TestClassRegistration {
                             CodeLocation registration_location) {
          Test::register_test(category, name, smoke_test, needs_serialization, [=] {
             auto test = std::make_unique<Test_Class>();
-            test->set_test_name(name);
-            test->set_registration_location(registration_location);
+            test->initialize(name, registration_location);
```
reneme, line 671: FYI: This will cause a merge conflict with #3919. Same as the hunk below.

reneme (approved):
Overall this looks good to me. Also, it looks like _a lot of work_. 😲

Frankly, I found the existence of `Test::rng()` very useful. But for sure: it does ridicule the existence of the test DRBG seed when running multi-threaded. Are you hoping that this helps with #3917?

randombit:
Thanks for fixing the last holdout tests!

> Also, it looks like a lot of work.

It was really not so bad with a clear plan in mind, which I had already thought about probably years ago. 90% of the work was Perl one liners and emacs macros, with some cleanup at the end.

> But for sure: it does ridicule the existence of the test DRBG seed when running multi-threaded. Are you hoping that this helps with https://github.com/randombit/botan/issues/3917?

Indeed, the main goal here is to make the tests (more) reproducible. I suspect there that the culprit in #3917 is more likely cpuid bit mangling - I don't see how RNG interleavings would affect those particular Dilithium tests.

PR #3814, opened 2023-11-23
FIX: Access Violation due to stale reference in OCB

reneme:
This contains a proposal for a fix of the crash in the OCB mode reported in #3812. Personally, I'm not very familiar with OCB, so this is the simplest fix I could come up with that likely won't break any optimization assumptions.

@randombit I'd be glad if you could take another look at this. The regression test in the commit should work reliably on typical platforms.

Closes #3812.

Issue #3812, opened 2023-11-21
AV on certain AEAD OCB decryption input calls under Windows.


ni4:
We use Botan as cryptography backend for RNP OpenPGP library. And recently, after updating CI runners to Botan 3.1.1, started to receive failures, only for Windows platform. After isolating the issue and enabling sanitizers got the following call stack. Suspecting something in RNP which may corrupt the memory (however sanitizers doesn't repor that), I was able to isolate the issue to pure botan calls with some certain input.

Do you have any idea why this happens?
Should I provide some more input, like test data file/order of decryption calls causing the failure?

```
2: ==6280==ERROR: AddressSanitizer: access-violation on unknown address 0x00004e000017 (pc 0x7ff9c7f0f990 bp 0x00e1edaf5fb0 sp 0x00e1edaf5eb0 T0)
2: ==6280==The signal is caused by a READ memory access.
2:     #0 0x7ff9c7f0f98f in Botan::L_computer::compute_offsets(unsigned __int64, unsigned __int64) C:\vcpkg\buildtrees\botan\src\3.1.1-d5e04c5040.clean\src\lib\modes\aead\ocb\ocb.cpp:68
2:     #1 0x7ff9c7f1024e in Botan::OCB_Decryption::decrypt(unsigned char *const, unsigned __int64) C:\vcpkg\buildtrees\botan\src\3.1.1-d5e04c5040.clean\src\lib\modes\aead\ocb\ocb.cpp:413
2:     #2 0x7ff9c7f10617 in Botan::OCB_Decryption::finish_msg(class std::vector<unsigned char, class Botan::secure_allocator<unsigned char>> &, unsigned __int64) C:\vcpkg\buildtrees\botan\src\3.1.1-d5e04c5040.clean\src\lib\modes\aead\ocb\ocb.cpp:451
2:     #3 0x7ff9c7e59139 in Botan::Cipher_Mode::finish C:\vcpkg\buildtrees\botan\x64-windows-rel\build\include\botan\cipher_mode.h:146
2:     #4 0x7ff9c7e59139 in `botan_cipher_update'::`2'::<lambda_1>::operator() C:\vcpkg\buildtrees\botan\src\3.1.1-d5e04c5040.clean\src\lib\ffi\ffi_cipher.cpp:159
[...]

PR #3924, opened 2024-02-27
Backport/FIX: Access Violation in OCB

reneme:
This is a back port of #3814, in response to [the reminder](https://github.com/randombit/botan/issues/3812#issuecomment-1966248869) by @ni4. Sorry, again, for that.

Anyway, it turns out, that this bug was actually exposed by [this patch during the 3.x development](https://github.com/randombit/botan/commit/510a8318e9596b795700681dcd9828f03c5e3265#diff-6b0a60b5ffc4d50b4acf24238f3502c6d6d2d31a6f2bf42b29f4be11178ce413). It reduced the allowed update granularity of the OCB mode, making it much easier to provoke a bogus state of OCB.

Albeit more contrived, its still possible to create an example to hit the access violation in 2.19.4.

### DIFF
--- a/src/lib/modes/aead/ocb/ocb.cpp
+++ b/src/lib/modes/aead/ocb/ocb.cpp
@@ -24,6 +24,17 @@ class L_computer final
          m_L_star.resize(m_BS);
          cipher.encrypt(m_L_star);
          m_L_dollar = poly_double(star());
+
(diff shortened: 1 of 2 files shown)


randombit (approved):
Nice thank you for following up on this

Issue #3928, opened 2024-03-02
Getting "bignum is negative" from ssh_agent using Botan generated RSA keys


droidmonkey:
Hello, this is a bit of a shot in the dark, but our users (and our own CI) are starting to report anomalies when adding generated RSA keys to ssh_agent. The ssh_agent outputs an error message stating the "bignum is negative" and refuses the key. This just started happening recently with no code changes to the relevant bits on our side. We suspect this may be related to a library change, Botan is the obvious one in this case.

Background issue: https://github.com/keepassxreboot/keepassxc/issues/10320

Relevant KeePassXC code:

https://github.com/keepassxreboot/keepassxc/blob/dff2f186ce0e063df5842731631c0f8e8c2af0e6/src/sshagent/OpenSSHKeyGen.cpp#L51-L82

``` C++
bool generateRSA(OpenSSHKey& key, int bits)
    {
        auto rng = randomGen()->getRng();

        try {
            Botan::RSA_PrivateKey rsaKey(*rng, bits);

            QByteArray publicData;
            BinaryStream publicStream(&publicData);
            // intentionally flipped n e -> e n
            bigIntToStream(rsaKey.get_e(), publicStream);
            bigIntToStream(rsaKey.get_n(), publicStream, 1);

            QByteArray privateData;
            BinaryStream privateStream(&privateData);
            bigIntToStream(rsaKey.get_n(), privateStream, 1);
            bigIntToStream(rsaKey.get_e(), privateStream);
            bigIntToStream(rsaKey.get_d(), privateStream);
            bigIntToStream(rsaKey.get_c(), privateStream, 1);
            bigIntToStream(rsaKey.get_p(), privateStream, 1);
            bigIntToStream(rsaKey.get_q(), privateStream, 1);

            key.setType("ssh-rsa");
            key.setCheck(randomGen()->randomUInt(std::numeric_limits<quint32>::max() - 1) + 1);
            key.setPublicData(publicData);
            key.setPrivateData(privateData);
            key.setComment("id_rsa");
            return true;
        } catch (std::exception& e) {
            return false;
        }
    }
```

Was the RSA code recently changed in Botan?

Is there something we can do to prevent negative big numbers?

reneme:
There was indeed a change that may be relevant. Did you update to 2.19.4?

reneme:
Interesting. The change I had in mind (#3909) was released more recently (2.19.4 and 3.3.0). Thanks for reporting, I will look into it.

[...]

PR #3929, opened 2024-03-04
Test: make sure that BigInt::binary_encode() writes padding

reneme:
As [mentioned here](https://github.com/randombit/botan/issues/3928#issuecomment-1976811349) I wanted to make sure that `BigInt::binary_encode()` correctly generates a zero-padding even for longer buffers. At first glance the implementation looked flawed to me. Though, it works as expected, albeit somewhat inefficient for long paddings (which isn't really a viable use case anyway).

Now that I've written some tests, we might as well keep them. 👼

### DIFF
--- a/src/tests/test_bigint.cpp
+++ b/src/tests/test_bigint.cpp
@@ -1,5 +1,6 @@
 /*
 * (C) 2009,2015,2016 Jack Lloyd
+* (C) 2024           Fabian Albert, René Meusel -  Rohde & Schwarz Cybersecurity
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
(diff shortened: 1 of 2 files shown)


src/tests/test_bigint.cpp
```diff
@@ -774,6 +773,57 @@ class DSA_ParamGen_Test final : public Text_Based_Test {
...
+std::vector<Test::Result> test_bigint_serialization() {
+   auto rng = Test::new_rng("test_bigint_serialization");
+
+   return {
+      Botan_Tests::CHECK("BigInt binary serialization",
+                         [](Test::Result& res) {
+                            Botan::BigInt a(0x1234567890ABCDEF);
+                            auto enc = Botan::BigInt::encode(a);
+                            res.test_eq("BigInt::encode()", enc, Botan::hex_decode("1234567890ABCDEF"));
+
```
FAlbertDev, line 785: Maybe, while you're at it:
```suggestion
                            Botan::BigInt a(0x1234567890ABCDEF);
                            auto enc = Botan::BigInt::encode(a);
                            res.test_eq("BigInt::encode()", enc, Botan::hex_decode("1234567890ABCDEF"));

                            auto enc_locked = Botan::BigInt::encode_locked(a);
                            res.test_eq(
                               "BigInt::encode_locked()", enc_locked, Botan::hex_decode_locked("1234567890ABCDEF"));

```

FAlbertDev (approved):

PR #3930, opened 2024-03-05
FIX: Correct sha512 compilation errors under clang 16, ARMv8.

### LINKED ISSUE #3927: Build failure, 3.3.0 amalgamation arm64
When building using clang 16.0.6, on Ubuntu arm64, I found that I had to modify:
```
src/lib/hash/sha2_64/sha2_64_armv8/sha2_64_armv8.cpp
```
to specify `BOTAN_FUNC_ISA("+crypto,sha3")`, instead of just `"+crypto"`, as without doing so, the compilation failed with errors of the format:
```
botan_all.cpp:88261:30: error: always_inline function 'vsha512su0q_u64' requires target feature 'sha3', but would be inlined into function 'compress_digest_armv8' that is compiled without support for 'sha3'
```
Excessive compiler spew omitted for brevity; many such errors.

Unlikely to be an optimal solution, but it did resolve the issue for me.

bazineta:
Corrects sha512 compilation errors issued when using clang 16, ARMv8. Fixes #3927. Environment tested, Ubuntu 22, clang 16.0.6, M3 Max.

### DIFF
--- a/src/lib/hash/sha2_64/sha2_64_armv8/sha2_64_armv8.cpp
+++ b/src/lib/hash/sha2_64/sha2_64_armv8/sha2_64_armv8.cpp
@@ -14,7 +14,7 @@ namespace Botan {
 /*
 * SHA-512 using CPU instructions in ARMv8
 */
-BOTAN_FUNC_ISA("+crypto,sha3")
+BOTAN_FUNC_ISA("arch=armv8.2-a+sha3")
 void SHA_512::compress_digest_armv8(digest_type& digest, std::span<const uint8_t> input8, size_t blocks) {
    alignas(128) static const uint64_t K[] = {
       0x428A2F98D728AE22, 0x7137449123EF65CD, 0xB5C0FBCFEC4D3B2F, 0xE9B5DBA58189DBBC, 0x3956C25BF348B538,
--- a/src/lib/hash/sha2_64/sha2_64_armv8/sha2_64_armv8.cpp
+++ b/src/lib/hash/sha2_64/sha2_64_armv8/sha2_64_armv8.cpp
@@ -14,7 +14,7 @@ namespace Botan {
 /*
 * SHA-512 using CPU instructions in ARMv8
 */
-BOTAN_FUNC_ISA("+crypto")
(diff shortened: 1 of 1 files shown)


reneme:
One CI failure (the arm64 cross compile target using qemu) is relevant. The others seem to be flukes.

```
  src/lib/hash/sha2_64/sha2_64_armv8/sha2_64_armv8.cpp:18:104: error: pragma or attribute ‘target("sha3")’ is not valid
     18 | void SHA_512::compress_digest_armv8(digest_type& digest, std::span<const uint8_t> input8, size_t blocks) {
        |                                                                                                        ^
  make: *** [Makefile:431: build/obj/lib/hash_sha2_64_sha2_64_armv8.o] Error 1
```

Perhaps the compiler we use here isn't up to the job? It's gcc:11.4-linux-arm64.
I'll also look into it later.

src/lib/hash/sha2_64/sha2_64_armv8/sha2_64_armv8.cpp
```diff
@@ -14,7 +14,7 @@ namespace Botan {
 /*
 * SHA-512 using CPU instructions in ARMv8
 */
-BOTAN_FUNC_ISA("+crypto")
+BOTAN_FUNC_ISA("+crypto,sha3")
```
reneme, line 17: @bazineta Could you instead please try this on your machine:

```suggestion
BOTAN_FUNC_ISA("arch=armv8.2-a+sha3")
```

That seems to work with the CI's qemu-based arm64 target.

reneme: 🎉 Hooray!

bazineta: Will revise this PR to contain that change.

reneme:
Much better now! I did incorporate this into #3931, along with some build system related discussion. Would you be terribly sad if we'd close this and continue discussions over there? 😶

bazineta:
> Would you be terribly sad if we'd close this and continue discussions over there? 😶

I am already over it....thanks. 😶

PR #3780, opened 2023-10-27
Add missing aes and sha2 ISA target for armv8

lieser:
This fixes the armv8 amalgamation build using the clang compiler from the Android NDK r26.

Fixes #3774.

Issue #3774, opened 2023-10-26
ARM64-v8a compiler error when amalgamation and armv8crypto is used


Trafo:
I am trying to make the botan Conan recipe compatible with Android and Emscripten. There I got the issue, that I could not build the Android ARM64-v8a version with amalgamation, because of compiler errors. If I deactivate it, everything builds.

Console output:
```
(linux) INFO: Downloading conan_sources.tgz
(linux) INFO: botan/3.2.0-emscripten@conan-pkg+botan/testa: Configuring sources in /builds/llt-at-nordlocker/ncrypto/.cache/.conan/data/botan/3.2.0-emscripten/conan-pkg+botan/testa/source/src
(linux) INFO: botan/3.2.0-emscripten@conan-pkg+botan/testa:
(linux) INFO: botan/3.2.0-emscripten@conan-pkg+botan/testa: Copying sources to build folder
(linux) INFO: botan/3.2.0-emscripten@conan-pkg+botan/testa: Building your package in /builds/llt-at-nordlocker/ncrypto/.cache/.conan/data/botan/3.2.0-emscripten/conan-pkg+botan/testa/build/c2911c17b898bcf0bfa856abf449c956bafd6e6b
(linux) INFO: botan/3.2.0-emscripten@conan-pkg+botan/testa: Generator txt created conanbuildinfo.txt
(linux) INFO: botan/3.2.0-emscripten@conan-pkg+botan/testa: Calling generate()
(linux) INFO: botan/3.2.0-emscripten@conan-pkg+botan/testa: Aggregating env generators
(linux) INFO: botan/3.2.0-emscripten@conan-pkg+botan/testa: Calling build()
(linux) INFO: botan/3.2.0-emscripten@conan-pkg+botan/testa: Apply patch (portability): Use WASM esxceptions instead of Emscripten emulation
[...]

PR #3931, opened 2024-03-05
FIX: arm64 amalgamations + better CI coverage

### LINKED ISSUE #3927: Build failure, 3.3.0 amalgamation arm64

reneme:
This continues the work @lieser started in #3780. It aims to fix #3927 and #3774. It also incorporates #3930.

Along with fixes to the relevant `BOTAN_FUNC_ISA()` annotations, this adds CI targets for amalgamation builds on arm64 (using GCC) and Android NDK arm64 (using clang). Also, this adds builds on the `macos-14` runner [released a month ago](https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/), that is based on Apple Silicon, i.e. arm64 😅. 

**Note:** This also fixes an issue in the Python wrapper, where the `ffi_api()` declaration failed to explicitly specify a `c_int` buffer length parameter. On Apple silicon that resulted in this parameter being `0` on the C/C++ side. Not entirely sure how that ever worked on all other platforms we test in CI.

@randombit I'm somewhat worried that these inline ISA-annotations aren't necessarily in sync with whatever is defined in the `src/build-data/cc/*.txt` files. Currently, for arm64 the definitions are as such:

clang.txt
```
arm64:armv8crypto -> "-march=armv8+crypto"
arm64:armv8sha512 -> "-march=armv8.2-a+sha3"
```

gcc.txt
```
arm64:armv8crypto -> ""
arm64:armv8sm3 -> "-march=armv8.2-a+sm4"
arm64:armv8sm4 -> "-march=armv8.2-a+sm4"
arm64:armv8sha512 -> "-march=armv8.2-a+sha3"
arm64:armv8sha3 -> "-march=armv8.2-a+sha3"
```

... i.e. `clang.txt` doesn't even define all of them. I'm guessing that the inline annotations using `BOTAN_FUNC_ISA()` are saving the day for those, no?


reneme:
Rebased after merging #3932

randombit (approved):
I agree things are not necessarily in sync with the info files. I also don’t like the repetition of having these magic (compiler-specific, sometimes undocumented) strings everywhere - ideally we’d have a header that encapsulates this like we do for the SIMD API annotations. (No need to do this here.)

Change itself lgtm.

randombit:
Good to merge? Not sure if more was going to happen here.

reneme:
Should be fine to merge. I didn't plan to address the discussed ISA flag duplication here.

PR #3934, opened 2024-03-07
Fix build using clang/libc++/winpthread on Windows.

solemnwarning:


randombit (approved):

lazka:
fyi, this breaks the mingw build:

```
src/lib/utils/os_utils/os_utils.cpp:668:22: error: use of undeclared identifier 'pthread_setname_np'
  668 |    static_cast<void>(pthread_setname_np(thread.native_handle(), name.c_str()));
      |                      ^
1 error generated.
```

PR #3935, opened 2024-03-11
Doc: Example to encrypt a message using a password

reneme:
I think this is a fairly typical use case for PBKDFs and perhaps warrants a more comprehensive example.

The function `encrypt_by_password` stretches a password using Argon2id into enough keying material to initialize AES-256/GCM with a key and IV. It then encrypts a message and prepends the salt used for Argon2 to the encrypted message. The final ciphertext therefore contains: `salt || encrypted bytes || GCM tag`.

The opposite direction is implemented by `decrypt_by_password`. If the GCM tag validation fails it concludes that the provided password was probably wrong.

Perhaps we should add a few pointers for how to define reasonable Argon2 parameters. The example uses M=256MiB, i = 4, p = 2.

### DIFF
--- a/src/examples/password_encryption.cpp
+++ b/src/examples/password_encryption.cpp
@@ -1,0 +1,140 @@
+#include <botan/aead.h>
+#include <botan/assert.h>
+#include <botan/auto_rng.h>
+#include <botan/hex.h>
+#include <botan/pwdhash.h>
+
+#include <iostream>
+
+namespace {
+
+template <typename OutT = std::vector<uint8_t>, typename... Ts>
+OutT concat(Ts&&... buffers) {
+   OutT out;
+   out.reserve((buffers.size() + ... + 0));
+   (out.insert(out.end(), buffers.begin(), buffers.end()), ...);
+   return out;
+}
+
+template <typename Out, typename In>
+Out as(const In& data) {
+   return Out(data.data(), data.data() + data.size());
+}
+
+constexpr size_t salt_length = 16;
+
+Botan::secure_vector<uint8_t> derive_key_material(std::string_view password,
+                                                  std::span<const uint8_t> salt,
+                                                  size_t output_length) {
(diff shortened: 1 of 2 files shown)


src/examples/password_encryption.cpp
```diff
@@ -0,0 +1,140 @@
...
+}
+
+std::unique_ptr<Botan::AEAD_Mode> prepare_aead(std::string_view password,
+                                               std::span<const uint8_t> salt,
+                                               Botan::Cipher_Dir direction) {
+   auto aead = Botan::AEAD_Mode::create_or_throw("AES-256/GCM", direction);
+
+   // Stretch the password into enough cryptographically strong key material
+   // to initialize the AEAD with a key and nonce (aka. initialization vector).
+   const auto keydata_needed = aead->key_spec().minimum_keylength() + aead->default_nonce_length();
```
randombit, line 56: Maybe use `maximum_keylength` here? It doesn't matter at all for `AES-256/GCM` since the key is specified, but would matter somewhat if someone copied this code and then (unwisely!!!) changed the AEAD to something like "Blowfish/EAX" which is an AEAD which supports a very wide range of keys, in fact down to a single byte.

src/examples/password_encryption.cpp
```diff
@@ -0,0 +1,140 @@
...
+   const auto keydata = derive_key_material(password, salt, keydata_needed);
+   const auto key = std::span{keydata}.first(aead->key_spec().minimum_keylength());
+   const auto nonce = std::span{keydata}.last(aead->default_nonce_length());
```
randombit, line 59: Grabbing the keylength and nonce lengths into local variables would avoid a bit of repetition, save on a couple of virtual calls, and (perhaps marginally) improve readability.

randombit (approved):

src/examples/password_encryption.cpp
reneme: Good to know, thanks.

PR #3885, opened 2024-01-11
Chore: Modernize copy_out_be/le

reneme:
### Pull Request Dependencies

* #3869

### Description

This removes the legacy `copy_out_{be/le}` helpers in `loadstor.h`. We replaced them with more versatile overloads of `load_*` and `store_*` that can handle collections of unsigned integers.

Here's the general idea of the helpers that can take collections:

```C++
// if we can "copy out" with word alignment...
std::vector<uint32_t> some_hash_state(4};
auto digest = store_be<std::vector<uint8_t>>(some_hash_state);

// if the final digest is not word aligned, we still need something like copy_out_x
// (but it is now based on the existing loadstore helpers and uses ranges)
std::vector<uint8_t> digest(15);
copy_out_be(digest, some_hash_state);
```

also added tests for the new implementations of `copy_out_x`

PR #3942, opened 2024-03-25
Chore: BufferTransformer

reneme:
### Pull Request Dependencies

* #3870

### Description

This is in response of https://github.com/randombit/botan/pull/3885#issuecomment-2016581381: Combining the `BufferStuffer` and `BufferSlicer` into a `BufferTransformer` that always acts on two buffers (one input, one output) in lock-step. Block ciphers (and in particular implementations of `BlockCipher::encrypt_n()`/`::decrypt_n()`) are a typical use case for this helper.

I added this spike on #3870, which introduced `std::span` in the relevant methods, to simplify the integration. See https://github.com/Rohde-Schwarz/botan/commit/00ff565707278174ab565f21eece4fed4fd1fd0e for the relevant patch.

@randombit Is that what you had in mind?

### DIFF
--- a/src/lib/block/aes/aes.cpp
+++ b/src/lib/block/aes/aes.cpp
@@ -11,6 +11,7 @@
 #include <botan/internal/ct_utils.h>
 #include <botan/internal/loadstor.h>
 #include <botan/internal/rotate.h>
+#include <botan/internal/stl_util.h>
 
 namespace Botan {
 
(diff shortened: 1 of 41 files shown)


randombit:
> Is that what you had in mind?

Basically so, yeah. I wonder if we can fit in some other common functionality there, namely the word level loads/stores. That's a very common task (encrypt+decrypt for N ciphers plus alternative implementations) so if there is a nice way to make that work it would cut some complexity. Not necessary for the initial PR. I just found an explict slicer+stuffer quite verbose in the original change.

reneme:
This is superseded by #4151

PR #3944, opened 2024-03-26
FIX: wrong module name for AES ARM acceleration

### LINKED ISSUE #3943: aes_neon module is not presesnt
as per documentatiom of botan https://botan.randombit.net/handbook/hardware_acceleration.html
aes_neon need to be enabled for hardware accelaration of botan for arm architecure.
But i can not see any moduled named aes_neon in botan 2.12.1 & 2.19.3

reneme:
Closes #3943


securitykernel (approved):
Thanks!

PR #3951, opened 2024-03-27
Improve FFI's botan_cipher_update() performance for stream ciphers

### LINKED ISSUE #3925: AES-CTR (and probably more stream ciphers) are slow when used via the FFI's Stream_Cipher mode.
Hello everyone,

I have created a benchmarking tool to compare performance between libraries and have noticed a significant drop in performance with Botan's AES CTR. In my benchmark, I generated plaintexts with a length of 4096 Bytes for each round and measured the time it takes for 1 million iterations. For my performance test I have used the Botan C interface ffi.h.

Results:

AES-NI and SSSE3 enabled:

| Library   | Algorithm    | Time (seconds) |
|-----------|--------------|----------------|
|Botan 3.2.0|     	 AES-128     	 |0.554971       
|Botan 3.2.0 |   	 AES-128/CTR 	|77.982720
|Botan 3.2.0  |   	 AES-128/CBC 	|5.963908
|Botan 3.2.0   |  	 AES-128/CFB 	|7.617746

AES-NI disabled:

| Library   | Algorithm    | Time (seconds) |
|-----------|--------------|----------------|
| Botan 3.2.0 | AES-128      | 9.187737       |
| Botan 3.2.0 | AES-128/CTR  | 81.234424      |
| Botan 3.2.0 | AES-128/CBC  | 19.137964      |
| Botan 3.2.0 | AES-128/CFB  | 20.943194      |

 With AES-NI and SSSE3 disabled:

| Library    | Algorithm    | Time (seconds) |
|------------|--------------|----------------|
| Botan 3.2.0 | AES-128      | 66.311669     |
| Botan 3.2.0 | AES-128/CTR  | 135.916029    |
| Botan 3.2.0 | AES-128/CBC  | 153.400266    |
| Botan 3.2.0 | AES-128/CFB  | 151.787276    |

### Environment:
- **Operating System:** Ubuntu 22.04.3 LTS (Codename: jammy)
- **CPU:** AMD Ryzen 5 5600U with Radeon Graphics
  - **Architecture:** x86_64
  - **CPU(s):** 12
  - **Thread(s) per core:** 2
  - **Core(s) per socket:** 6
  - **CPU max MHz:** 4289,0000
  - **CPU min MHz:** 400,0000
 ```bash
lscpu | grep "Byte Order"
Byte Order:                         Little Endian
```
I suspect that the AES CTR encryption process is not being parallelized, which could be causing the significant drop in performance. However, I'm uncertain about the exact reason for this issue and would appreciate your assistance in investigating further.

Thank you and best regards,
Ahmed 

Edit 1: add Byte Order
Edit 2: fix a comma typo

### PR DESCRIPTION
**Pull request by reneme**
### Pull Request Dependencies

* #3969 
* #3971 

### Description

This reworks the implementation of FFI `botan_cipher_update()`, making it comply with [the general "rules of engagement" of the FFI](https://botan.randombit.net/handbook/api_ref/ffi.html#rules-of-engagement). **Note that this subtly changes the behavior when finalizing a cipher without providing a big enough output buffer up-front.**

When the caller failed to provide a big enough output buffer, it now returns `BOTAN_FFI_ERROR_INSUFFICIENT_BUFFER_SPACE` instead of `BOTAN_FFI_ERROR_INVALID_INPUT` and sets `output_written` to the required number of bytes. In that case, re-invoking `botan_cipher_update()` with the "final" flag and a sufficiently sized output buffer then obtains the remaining bytes and resets the cipher object. 

Before, the re-invocation had to happen _without_ setting the final flag again to obtain the same behavior.
@randombit Frankly, I doubt that the existing implementation supported such a usage explicitly. It did feel like an inadvertent hack to me. That's why I changed it to an (IMHO) more logical interface. Nevertheless, there might be an argument to keep the existing behavior, albeit inconsistent with the generic "rules of engagement".

The rework should also significantly improve the performance of stream cipher modes (e.g. CTR-mode) via the FFI by opportunistically leveraging the `ideal_update_granularity()` of the cipher mode. Before, we generally processed data in chunks of `update_granularity()`, which is "1 byte" for stream ciphers. That resulted in very poor performance, as described in #3925.

**TODO:** We might look into the internal function `ffi_choose_update_size()` and whether it still makes sense in this new setup. Frankly, I'm lacking the historic context to conclusively judge that.

Closes #3925.

### DIFF
--- a/src/lib/ffi/ffi_cipher.cpp
+++ b/src/lib/ffi/ffi_cipher.cpp
@@ -8,6 +8,7 @@
 
 #include <botan/aead.h>
 #include <botan/internal/ffi_util.h>
+#include <botan/internal/stl_util.h>
 
 extern "C" {
 
@@ -15,18 +16,24 @@ using namespace Botan_FFI;
 
 struct botan_cipher_struct final : public botan_struct<Botan::Cipher_Mode, 0xB4A2BF9C> {
    public:
-      explicit botan_cipher_struct(std::unique_ptr<Botan::Cipher_Mode> x, size_t update_size) :
-            botan_struct(std::move(x)), m_update_size(update_size) {
-         m_buf.reserve(m_update_size);
+      explicit botan_cipher_struct(std::unique_ptr<Botan::Cipher_Mode> x,
+                                   size_t update_size,
+                                   size_t ideal_update_size) :
+            botan_struct(std::move(x)), m_update_size(update_size), m_ideal_update_size(ideal_update_size) {
+         BOTAN_DEBUG_ASSERT(ideal_update_size >= update_size);
+         m_buf.reserve(m_ideal_update_size);
       }
 
       Botan::secure_vector<uint8_t>& buf() { return m_buf; }
 
       size_t update_size() const { return m_update_size; }
 
+      size_t ideal_update_size() const { return m_ideal_update_size; }
+
    private:
       Botan::secure_vector<uint8_t> m_buf;
       size_t m_update_size;
+      size_t m_ideal_update_size;
 };
 
 namespace {
@@ -71,8 +78,9 @@ int botan_cipher_init(botan_cipher_t* cipher, const char* cipher_name, uint32_t 
       }
 
       const size_t update_size = ffi_choose_update_size(*mode);
+      const size_t ideal_update_size = std::max(mode->ideal_granularity(), update_size);
 
-      *cipher = new botan_cipher_struct(std::move(mode), update_size);
+      *cipher = new botan_cipher_struct(std::move(mode), update_size, ideal_update_size);
       return BOTAN_FFI_SUCCESS;
    });
 }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/lib/ffi/ffi_cipher.cpp [hunk 1]
```diff
@@ -173,39 +177,38 @@ int botan_cipher_update(botan_cipher_t cipher_obj,
...
 
       if(input_size == 0) {
-         // Currently must take entire buffer in this case
-         *output_written = mbuf.size();
-         if(output_size >= mbuf.size()) {
-            copy_mem(output, mbuf.data(), mbuf.size());
-            mbuf.clear();
-            return BOTAN_FFI_SUCCESS;
-         }
-
```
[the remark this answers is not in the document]
reneme: In fact, the coverage report claims that these lines are hit, though: https://coveralls.io/builds/66545020/source?filename=source%2Fsrc%2Flib%2Fffi%2Fffi_cipher.cpp#L175

# 🤔

randombit:
One thing that's not really clear here is - do the newly added tests pass on master, without the other changes to `ffi_cipher.cpp`?

I'd prefer multiple PRs here

- `scoped_cleanup` + tests
- `botan_cipher_requires_entire_message`
- New FFI tests

finally the buffering logic changes themselves.

reneme:
> I'd prefer multiple PRs here

No problem. I'll do that once CI is green on the entire patch set.

> do the newly added tests pass on master, without the other changes

Frankly, I don't think so. At least not out-of-the-box. The existing implementation of `botan_cipher_update()` didn't seem to _explicitly_ handle the final flag and an insufficiently sized output buffer. Instead of `BOTAN_FFI_ERROR_INSUFFICIENT_BUFFER_SPACE` it just returned `-1` in that case.

When removing this bit of the new tests, they should pass, though.

reneme:
While trying to port the AEAD test down to the existing implementation I found a couple more bugs in [the inner loop of `botan_cipher_update()`](https://github.com/randombit/botan/blob/master/src/lib/ffi/ffi_cipher.cpp#L192-L205) when used with "SIV" (and probably also "CCM"):

* passing some input bytes but no output buffer results in "0 bytes" being consumed (despite those modes not producing any streamed output anyway -- they return `require_entire_message() -> true`). This is somewhat surprising but not catastrophic.
* providing a "dummy" output buffer, however, results in the plaintext being copied into the output buffer. The current implementation fails to detect that `cipher.update()` actually calls `.resize(0)` on `mbuf` (for SIV) and blindly copies the plaintext bytes out of `mbuf.data()`. This might even be UB, but to the best of my knowledge `.resize(0)` doesn't actually de-allocate the vector's underlying memory. Nevertheless, the plaintext ending up in the "encryption output" buffer is bad enough. 😨

reneme:
Rebased to master after the underlying pull requests were merged.

reneme:
Force-pushed to add GPG signatures to the commit. I'm back at home with access to a USB-C to USB-A adapter for my YubiKey. 🤦‍♂️

src/lib/ffi/ffi_cipher.cpp
```diff
@@ -71,8 +78,9 @@ int botan_cipher_init(botan_cipher_t* cipher, const char* cipher_name, uint32_t
       }
 
       const size_t update_size = ffi_choose_update_size(*mode);
+      const size_t ideal_update_size = mode->ideal_granularity();
```
randombit, line 81: This is potentially wrong if `ideal_granularity` is small and/or `minimum_final_size` is large. `ffi_choose_update_size` will scale up the granularity to satisfy the minimum final size. So we might end up here (in a weird situation) where the ideal update size was actually less than the update size.

I don't think this should happen absent an oversight in implementing a mode but should guard against it. Easiest path seems to be just use `std::max(mode->ideal_granularity(), update_size)`

randombit (approved):
Thanks. This is a lot cleaner and easier to understand. Really demonstrates the benefits of `scoped_cleanup` and the span oriented slicers :+1:

src/lib/ffi/ffi_cipher.cpp
reneme: Fair enough. I applied your suggestion.

reneme:
Rebased to master and addressed the review comment.

ni4:
@reneme This broke things for RNP AEAD implementation, as we relied on the previous behaviour where `botan_cipher_update()` for non-final chunk processes the whole chunk, writing out the equal size as well. Anywa, we'll fix that. The only question is whether it is correct to assume that calling (non-final) `botan_cipher_update()` for AEAD cipher will not return more data then it is provided by the input?

reneme:
> This broke things for RNP AEAD implementation, as we relied on the previous behaviour where `botan_cipher_update()` for non-final chunk processes the whole chunk, writing out the equal size as well.

@ni4 Sorry about that!

For clarification (and for the record): Also previously, `botan_cipher_update()` might have left some input bytes untouched. Namely, when the number of input bytes wasn't a multiple of some internal update granularity. For AEADs this internal granularity used to be "1 byte" inadvertently ([resulting in poor performance](https://github.com/randombit/botan/issues/3925)). I.e. it used to always consume every byte.

@randombit Perhaps we should mention that gotcha in the changelog?

> whether it is correct to assume that calling (non-final) `botan_cipher_update()` for AEAD cipher will not return more data then it is provided by the input?

Yes, it won't ever produce more bytes than provided as input. Unless _during finalization_ to emit the AEAD's authentication tag. I clarified that in the documentation of the function as well: https://github.com/randombit/botan/pull/4088.

src/lib/ffi/ffi_cipher.cpp
```diff
@@ -132,81 +140,117 @@ int botan_cipher_start(botan_cipher_t cipher_obj, const uint8_t* nonce, size_t n
...
+
+            while(in.remaining() >= granularity && out.remaining_capacity() >= expected_output_per_iteration) {
+               copy_mem(mbuf, in.take(granularity));
+               const auto written_bytes = cipher.process(mbuf);
+               BOTAN_DEBUG_ASSERT(written_bytes == expected_output_per_iteration);
+               if(written_bytes > 0) {
+                  BOTAN_ASSERT_NOMSG(written_bytes <= granularity);
+                  copy_mem(out.next(written_bytes), std::span(mbuf).first(written_bytes));
+               }
+            }
```
reneme, line 236: @randombit There's more room for optimization here. We could detect that the user passed the same buffer for input and output and omit the copy to the internal `mbuf` in that case.

ni4:
> For clarification (and for the record): Also previously, `botan_cipher_update()` might have left some input bytes untouched. Namely, when the number of input bytes wasn't a multiple of some internal update granularity. For stream-cipher-ish ciphers (like AES-GCM for instance) this internal granularity used to be "1 byte" inadvertently ([resulting in poor performance](https://github.com/randombit/botan/issues/3925)). I.e. it used to always consume every byte.

@reneme We used the update granularity, and I believe that at some point in the past it was like 16 bytes or so. Probably at some point that changed.

> Yes, it won't ever produce more bytes than provided as input. Unless _during finalization_ to emit the AEAD's authentication tag. I clarified that in the documentation of the function as well: #4088.

Yeah, tag should not be a problem. We just use the same preallocated buffer to encrypt/decrypt (and it is the same for input and output, like you mentioned in the further comment) so it would be unwilling to do allocations during the call.

Thanks!

PR #3952, opened 2024-03-27
Don't stop the ninja build if errors occur

randombit:
This mirrors the behavior used for make


reneme (approved):
:+1:

... we should also look into parsing the compiler errors and rendering them into the pull request diff. Hopefully, that would avoid quite a bit of log scrolling. It's on my list for a long time. 😅

PR #3956, opened 2024-03-29
Add a generic Comba multiply/square function

randombit:
Use it instead of the specialized for 7 word Comba in the Curve448 code, which is the only place a 7 word multiply is useful.


FAlbertDev (approved):
Thank you! Looks good to me.

PR #3959, opened 2024-03-30
Use a simpler formula for Kyber division by q

randombit:
Simpler approach than #3898

### DIFF
--- a/src/lib/pubkey/kyber/kyber_common/kyber.cpp
+++ b/src/lib/pubkey/kyber/kyber_common/kyber.cpp
@@ -78,11 +78,20 @@ KyberMode::Mode kyber_mode_from_string(std::string_view str) {
  * It enforces the optimization of various compilers,
  * replacing the division operation with multiplication and shifts.
  *
+ * This implementation is only valid for integers <= 2**20
+ *
  * @returns (a / KyberConstants::Q)
  */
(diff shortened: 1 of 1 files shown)


FAlbertDev (approved):
I'm okay with using this more straightforward approach. We can stay with this simple method until we need other CT divisions in the future.
Since we already have a test for our ct_div function in #3898, we may also want to apply it here.
```cpp
// src/tests/test_kyber.cpp
namespace {

std::vector<Test::Result> test_kyber_utilities() {
   return {
      Botan_Tests::CHECK("constant-time division",
                         [](Test::Result& result) {
                            constexpr auto Q = Botan::KyberConstants::Q;
                            // Check ct_int_div_kyber_q with all possible inputs that may appear in Kyber.
                            for(uint32_t i = 0; i < (1 << 18); ++i) {
                               result.test_is_eq(
                                  Botan::fmt("{}/Q", i), Botan::ct_int_div_kyber_q(i), i / Q);
                            }
                         }),
   };
}

}  // namespace

BOTAN_REGISTER_TEST_FN("kyber", "kyber_utils", test_kyber_utilities);
```

randombit:
The function isn't currently exposed anywhere and I'm inclined not to expose it simply to test it, considering it is quite small and should already be very well tested as a side effect of the Kyber tests themselves. I did do an offline test that the expression works for all integers < 2**20.

Issue #3958, opened 2024-03-30
botan 3.3.0 fails to build on powerpc G4


he32:
Hi, it appears that botan 3.3.0 fails to build on 32-bit powerpc, on a G4 specifically.
This appears to be because someone has snuck power8-only-operations in under the "has Altivec" flag, and the 32-bit powerpc CPUs do not have the power8-only operations.  The specific error I get is:

```
In file included from src/lib/block/noekeon/noekeon_simd/noekeon_simd.cpp:10:
build/include/botan/internal/simd_32.h: In member function 'Botan::SIMD_4x32 Botan::SIMD_4x32::bswap() const':
build/include/botan/internal/simd_32.h:483:35: error: '__builtin_vsx_revb_v4si' requires the '-mcpu=power8' and '-mvsx' options
  483 |          return SIMD_4x32(vec_revb(m_simd));
      |                                   ^
build/include/botan/internal/simd_32.h:483:35: note: overloaded builtin '__builtin_vec_revb' is implemented by builtin '__builtin_vsx_revb_v4si'
gmake: *** [Makefile:191: build/obj/lib/block_noekeon_simd.o] Error 1
*** Error code 2
```

From cursory checking, it looks like the `__VSX__` and `__POWER8_VECTOR__` macros are defined when those corresponding extensions are available, but botan 3.3.0 doesn't check for any of those before using those extensions.

As for how to implement the primitive using just altivec and not vsx, I don't know, and could therefore use some hints.

randombit:
Sorry about that. These days the only machines I have access to for testing are POWER 7/8/9 machines so its easy for issues like this to slip in. This operation should be easy to do with a single `vec_perm` since we’re just flipping the bytes in each 32 bit portion. Something like

```
__vector unsigned char rev = { 3, 2, 1, 0, 7, 6, 5, 4, 11, 10, … };
return vec_perm(a, a, rev);
```

Indices may be wrong here, been a while since I did much with Altivec permutes.

he32:
Ah, thanks for the hint.  I prepared a patch which doesn't use altivec, and it looks like the below.  I'll adapt it to use your suggestion above, though.

```
--- src/lib/utils/simd/simd_32.h.orig   2024-02-20 17:28:54.000000000 +0000
+++ src/lib/utils/simd/simd_32.h
@@ -21,6 +21,12 @@
    #undef vector
    #undef bool
    #define BOTAN_SIMD_USE_ALTIVEC
[...]

PR #3960, opened 2024-03-30
Fix build for non-VSX altivec powerpc CPUs.

he32:
Fixes https://github.com/randombit/botan/issues/3958


randombit:
LGTM but the formatter check is complaining. You should just be able to apply the diff printed in the failing job.

he32:
> LGTM but the formatter check is complaining. You should just be able to apply the diff printed in the failing job.

I think I fixed the formatting and squashed the two commits back into one.

he32:
And ... made another formatting fix and squash.

randombit (approved):

PR #3966, opened 2024-04-03
Apply constexpr to BufferStuffer

randombit:


reneme (approved):
We could probably add `constexpr` to basically all methods, no?

randombit:
Likely so. These changes are not really principled, rather entirely driven by what I'm working on atm (fully `constexpr` elliptic curve library). Probably `BufferSlicer` and elsewhere could get the same treatment.

PR #3969, opened 2024-04-04
Add FFI API botan_cipher_requires_entire_message()

reneme:
... as it says on the box.

@randombit  One question: Should this be added to the python wrapper? I'm unsure, given that the python module just collects whatever shared object it finds. What happens, if it get a hold of an older library version whose FFI didn't provide this method. What's our contract here?


reneme:
Updated accordingly. I took the liberty to also add `ideal_update_granularity` and make `is_authenticated` a wrapper around `botan_cipher_is_authenticated`.

randombit (approved):
Looks great thanks!

PR #3973, opened 2024-04-04
Split out XMD from hash2curve to its own module

randombit:


src/lib/kdf/xmd/info.txt
```diff
@@ -0,0 +1,11 @@
+<defines>
+XMD -> 20240404
+</defines>
+
+<module_info>
+name -> "XMD"
```
reneme, line 6: Perhaps add a `brief -> ""` here? That string shows up in the module overview in doxygen.

src/lib/kdf/xmd/xmd.h
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Botan {
+
+void BOTAN_TEST_API expand_message_xmd(std::string_view hash_fn,
+                                       uint8_t output[],
+                                       size_t output_len,
+                                       const uint8_t input[],
+                                       size_t input_len,
+                                       const uint8_t domain_sep[],
+                                       size_t domain_sep_len);
```
reneme, line 21: `std::span<>` while being on it, perhaps? No strong feelings about it, though.

src/lib/kdf/xmd/info.txt
reneme, line 6: As far as I can tell, the module exposes an internal API, only. Perhaps, mark it as `type -> "Internal"`?

reneme (approved):
Some suggestions regarding the module configuration.

src/lib/kdf/xmd/info.txt
randombit: Right thanks

randombit: Done

src/lib/kdf/xmd/xmd.h
randombit: `span` conversion completed, worked out nice except that converting a `string_view` to a `span<const uint8_t>` is fugly.

src/lib/kdf/xmd/xmd.cpp
```diff
@@ -0,0 +1,81 @@
...
+      }
+
+      // Now compute the next b_i if needed
+
+      b_i ^= b_0;
+      hash->update(b_i);
+      hash->update(cnt);
+      hash->update(domain_sep);
+      hash->update(domain_sep_len);
+      hash->final(b_i.data());
```
reneme, line 76:
```suggestion
      hash->final(b_i);
```

... `b_i` was initially created with `hash->final()`, hence should be the correct length. Passing it as a range double-checks that, however.

src/lib/kdf/xmd/xmd.h
```diff
@@ -0,0 +1,40 @@
...
+void BOTAN_TEST_API expand_message_xmd(std::string_view hash_fn,
+                                       std::span<uint8_t> output,
+                                       std::span<const uint8_t> input,
+                                       std::span<const uint8_t> domain_sep);
+
+inline void expand_message_xmd(std::string_view hash_fn,
+                               std::span<uint8_t> output,
+                               std::string_view input_str,
+                               std::string_view domain_sep_str) {
+   std::span<const uint8_t> input(reinterpret_cast<const uint8_t*>(input_str.data()), input_str.size());
```
reneme, line 31: That's always annoying. Perhaps we want some utility for this at some point.

src/lib/pubkey/ec_group/ec_group.cpp
```diff
@@ -631,7 +631,9 @@ EC_Point EC_Group::hash_to_curve(std::string_view hash_fn,
       throw Not_Implemented("EC_Group::hash_to_curve not available for this curve type");
    }
 
-   return hash_to_curve_sswu(*this, hash_fn, input, input_len, domain_sep, domain_sep_len, random_oracle);
+   std::span input_span(input, input_len);
+   std::span dst_span(domain_sep, domain_sep_len);
+   return hash_to_curve_sswu(*this, hash_fn, input_span, dst_span, random_oracle);
```
reneme, line 636: For the record (it's debatable what form is better), this would work as well and is more compact:

```suggestion
   return hash_to_curve_sswu(*this, hash_fn, {input, input_len}, {domain_sep, domain_sep_len}, random_oracle);
```

src/lib/kdf/xmd/xmd.cpp
```diff
@@ -0,0 +1,81 @@
...
+   hash->update(domain_sep);
+   hash->update(domain_sep_len);
+
+   secure_vector<uint8_t> b_i = hash->final();
+
+   uint8_t cnt = 2;
+   for(;;) {
+      const size_t produced = std::min(output.size(), hash_output_size);
+
+      copy_mem(&output[0], b_i.data(), produced);
```
reneme, line 62: Maybe:

```suggestion
      copy_mem(std::span<uint8_t>(output).first(produced), std::span<const uint8_t>(b_i).first(produced));
```

... I've used this idiom in some places, but I'm not super happy with it. It's quite chatty, and IIUC `.first()` doesn't even range-check the `span`, so technically it isn't even saver than the bare pointer stuff.

reneme (approved):
Thanks! Some nits and suggestions. Nothing of which is truly necessary, though.

src/lib/kdf/xmd/xmd.h
randombit: Likely so. There are already some conversions like this for bare points in `mem_ops.h` since this comes up pretty often.

reneme (approved):

src/lib/kdf/xmd/xmd.cpp
randombit: Lot of noise with not a lot of benefit imo. Probably worth investigating options for some span-centric operations like this for the general case.

PR #3974, opened 2024-04-05
Add an option to run_clang_format to skip the version check

randombit:


reneme:
I'm guessing that you're thinking of relaxing the check from requiring version 15 explicitly to a list of versions that are known to be compatible?

reneme (approved):

randombit:
Eventually that. Right now it's because I'm on a new laptop that has only `clang-format 17` which is *almost* compatible, but not quite. So it's nice to be able to run the formatter and know it's at least close to what CI will expect. I do think with a bit of work we can extend the supported versions.

PR #3976, opened 2024-04-05
Apply some formatting changes from clang-format 17

randombit:
These are some changes from `clang-format 17` that I think our CI clang would also agree on.


randombit:
Oh excellent even the ones that looked totally innocuous like `x&0xFF` vs `x & 0xFF` are rejected. I guess we really are stuck with exactly `clang-format 15` for the moment.

PR #3979, opened 2024-04-06
Add library for compile time instantiation of elliptic curves

randombit:
This new approach has a lot of benefits. It avoids the side channels that plague anything based on `BigInt` (or any general purpose arbitrary length integer type), it also avoids almost all heap allocations and inlines nicely.

Performance advantage depends on curve and compiler. For smaller curves (eg secp256r1) ECDSA seems be almost twice as fast. For larger such as secp521r1 the advantage is more like 40%. Clang seems to be able to optimize everything much better than GCC here, for reasons I haven't been able to nail down.

Performance can be improved significantly from here; besides P-521 we're not taking advantage of specialized reductions possible with many primes, nor is there a system to express the precomputed addition chains used for inversions. The BigInt based code makes use of both of these already.

One disadvantage is that the code size is quite significant. The worst overhead is actually the symbol names. I've applied various tricks to reduce these but still over half the size of `math_pcurves.o` is (per Google's bloaty tool) symbols. I'd like to reduce these in the future, but I don't think it's a blocker to merge.


reneme:
Awesome! I'll certainly have a look. Not before tomorrow, though.

randombit:
No rush! This is certainly not going into 3.4 so there is plenty of time.

Most interesting thing about the sea of red in CI - GCC 11.4 on x86 fails with constexpr timeout, while same version of GCC on say Aarch64 or S390 accepts. So constexpr limits are architecture dependent :( and thus intrinsically flaky since you can be near some limit without knowing it

randombit:
Clang seems to have a nasty bug where `std::is_constant_evaluated` returns false incorrectly in `constexpr` context, then it rejects the code because we used `asm` or `volatile` in `constexpr` context in the `!std::is_constant_evaluated` branch :rage: 

It works fine for me in Clang 17 on my machines so I'm assuming this was a Clang bug that was fixed subsequently. I'd vote for just bumping the Clang minimum version - that's an absolutely insane bug and nearly impossible to work around - but unfortunately the version in Android NDK also has the bug and we are stuck with that version at least until the next NDK release.

randombit:
Downside of this approach is that (due to name mangling being somewhat horribly thought out for this case) object sizes are really big. Even with just 3 curves, on my machine the object file is one of the largest from the whole library, that will get a lot worse with 27. And compile times may prove untenable.

I'm sure 90% of this can be salvaged but I suspect in the end the string based instantiation won't work in practice :sob: which is a shame since it's quite pretty imo

randombit:
Clang bug might be https://github.com/llvm/llvm-project/issues/55638

randombit:
This error

```
 call to consteval function 'Botan::p_minus<(unsigned char)'\x01', unsigned long, 4UL>' is not a constant expression
```

Looks like https://github.com/llvm/llvm-project/issues/51182

randombit:
I can repro the `std::if_constant_evaluated` problem using Clang 16 on my machine so it looks to be a bug fixed in Clang 17, though I don't see anything that looks relevant in the release notes.

src/lib/pubkey/pcurves/pcurves_util.h
```diff
@@ -0,0 +1,188 @@
...
+
+      const W carry = bigint_sub3(t.data(), r.data(), N + 1, p.data(), N);
+
+      if(carry == 0) {
+         std::swap(r, t);
+      }
+   }
+
+   std::array<W, N> rs;
+   std::copy(r.begin(), r.begin() + N, rs.begin());
```
reneme, line 42: Now that `copy_mem` has a constexpr-variant, I think we should prefer that. If both passed-in ranges have static length, it will do a `static_assert` of the correct lengths.

```suggestion
   std::copy(std::span{r}.first<N>(), rs);
```

src/lib/pubkey/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,816 @@
...
+         auto bytes = store_be(v);
+
+         if constexpr(Self::BYTES == Self::N * WordInfo<W>::bytes) {
+            return bytes;
+         } else {
+            // Remove leading zero bytes
+            const size_t extra = Self::N * WordInfo<W>::bytes - Self::BYTES;
+            std::array<uint8_t, Self::BYTES> out;
+            copy_mem(out.data(), &bytes[extra], Self::BYTES);
+            return out;
```
reneme, line 228: Perhaps using range-based `copy_mem`?

```suggestion
            // Remove leading zero bytes
            constexpr size_t extra = Self::N * WordInfo<W>::bytes - Self::BYTES;
            std::array<uint8_t, Self::BYTES> out;
            copy_mem(out, std::span{bytes}.template subspan<extra, Self::BYTES>();
            return out;
```

src/lib/pubkey/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,816 @@
...
+         }
```
reneme, line 229: Perhaps we could get away without the `if constexpr` and the potential memory copy entirely?

```suggestion
         auto v = bigint_monty_redc(m_val, Self::P, Self::P_dash);
         std::reverse(v.begin(), v.end());
         
         std::array<uint8_t, Self::BYTES> out = {0};
         static_assert(Self::N * WordInfo<W>::bytes >= Self::BYTES); // is this guranteed somehow anyway?
         const size_t zero_padding_offset = Self::N * WordInfo<W>::bytes - Self::BYTES;
         store_be(std::span{out}.template subspan<zero_padding_offset>(), v);
         return out;
```

... this is untested. Just a sketch of the idea. `store_be` with a statically-sized out-param will static_assert that the byte buffer matches the input range exactly.

src/lib/pubkey/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,816 @@
...
+            const std::array<W, N> hi = bytes_to_words<W, N, BYTES>(&padded_bytes[0]);
+            const std::array<W, N> lo = bytes_to_words<W, N, BYTES>(&padded_bytes[BYTES]);
+            return Self(hi) * R3 + Self(lo) * R2;
+         }
+      }
+
+      static constexpr Self random(RandomNumberGenerator& rng) {
+         std::array<uint8_t, Self::BYTES> buf;
+         for(;;) {
+            rng.randomize(buf.data(), buf.size());
```
reneme, line 281:
```suggestion
            rng.randomize(buf);
```

src/lib/pubkey/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,816 @@
...
+      }
+
+      constexpr std::array<uint8_t, Self::BYTES> serialize() const {
+         std::array<uint8_t, Self::BYTES> r = {};
+         BufferStuffer pack(r);
+         pack.append(0x04);
+         pack.append(m_x.serialize());
+         pack.append(m_y.serialize());
+         return r;
+      }
```
reneme, line 357: It would be nice if we could use the `concat` helpers for this. Currently, that doesn't work, because `concat` requires the output type to have `.insert()`. But, technically, we could extend it to be used like that:

```suggestion
      constexpr std::array<uint8_t, Self::BYTES> serialize() const {
         return concat_as<std::array<uint8_t, Self::BYTES>(
            store_le<uint8_t>(0x04), // store_le is just to make it look like a byte-array
            m_x.serialize(),
            m_y.serialize());
      }
```

This could then also (statically) assert that the output array was filled entirely.

src/lib/pubkey/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,816 @@
...
+      template <typename Pt>
+      constexpr void conditional_add(bool cond, const Pt& pt) {
+         conditional_assign(cond, *this + pt);
+      }
+
+      void conditional_assign(bool cond, const Self& pt) {
+         m_x.conditional_assign(cond, pt.x());
+         m_y.conditional_assign(cond, pt.y());
+         m_z.conditional_assign(cond, pt.z());
+      }
```
reneme, line 442: I'm assuming this is needed for a constant-time implementation? So far, I saw it as an unspoken rule that all methods that are "supposed to be constant-time" are prefixed with `ct_...`. I found this a helpful convention, that might be applicable here as well.

reneme (commented):
Interesting! 😄

I merely skimmed through the changes and left a few suggestions and comments here and there. No thorough review whatsoever.

randombit:
GCC 11 miscompilation, excellent

src/lib/pubkey/pcurves/pcurves_impl.h
randombit: In this code I'm going for the opposite convention namely that everything is constant time by default and anything that is intentionally variable time has a `_vartime` suffix.

(There are some current exceptions eg in the point multiplication that will be fixed before merging.)

src/lib/pubkey/pcurves/pcurves_impl.h
reneme: #3994 is work towards this. It makes `concat()` `constexpr` and also automagically inferring the output array size. So this should actually work as:

```suggestion
      constexpr std::array<uint8_t, Self::BYTES> serialize() const {
         return concat(
            store_le<uint8_t>(0x04), // store_le is just to make it look like a byte-array
            m_x.serialize(),
            m_y.serialize());
      }
```

src/lib/pubkey/pcurves/pcurves_util.h
```diff
@@ -0,0 +1,188 @@
+/*
+* (C) 2024 Jack Lloyd
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#ifndef BOTAN_PCURVES_UTIL_H_
+#define BOTAN_PCURVES_UTIL_H_
```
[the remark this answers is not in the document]
randombit: Main problem is for some technical reason I'm not bothering to learn the details of, C++ can't deduce the conversion from a statically sized `array` to a statically sized `span`. https://stackoverflow.com/questions/70983595

This can be worked around with some additional noise but the additional flexibility doesn't buy us anything in this context so I'd rather keep things as simple as possible.

src/lib/pubkey/pcurves/pcurves_impl.h
randombit: Something seems to be missing still

```
build/include/internal/botan/internal/loadstor.h: In instantiation of ‘constexpr auto Botan::store_le(ParamTs&& ...) [with ModifierT = unsigned char; ParamTs = {int}]’:
build/include/internal/botan/internal/pcurves_impl.h:351:30:   required from here
build/include/internal/botan/internal/loadstor.h:699:67: error: no matching function for call to ‘store_any<Botan::detail::Endianness::Little, unsigned char>(int)’
  699 |    return detail::store_any<detail::Endianness::Little, ModifierT>(std::forward<ParamTs>(params)...);
      |           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

src/lib/pubkey/pcurves/pcurves_util.h
randombit: :+1:

src/lib/pubkey/pcurves/pcurves_impl.h
randombit: :+1:

randombit:
Hilariously, considering my initial guess that VC would be the problem compiler, we've had

* Several Clang constexpr bugs
* A GCC miscompilation
* Loss of a days worth of debugging time to macOS thread stacks being 512 KiB (?!?!)

And zero problems with Visual C++

reneme:
> And zero problems with Visual C++

That was my impression in recent times as well. Microsoft is doing a good job implementing the new parts of the standard, lately. 😊

randombit:
@reneme I think this is ready to go.

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1404 @@
...
+      // Reduces large input modulo the order
+      template <size_t L>
+      static constexpr Self from_wide_bytes(std::span<const uint8_t, L> bytes) {
+         static_assert(8 * L <= 2 * Self::BITS);
+         std::array<uint8_t, 2 * BYTES> padded_bytes = {};
+         copy_mem(padded_bytes.data() + 2 * BYTES - L, bytes.data(), L);
+         return Self(Rep::wide_to_rep(bytes_to_words<W, 2 * N, 2 * BYTES>(&padded_bytes[0])));
+      }
+
+      static constexpr Self random(RandomNumberGenerator& rng) {
```
reneme, line 472: Not sure about the `constexpr` here, because of `RNG::randomize()`.

```suggestion
      static Self random(RandomNumberGenerator& rng) {
```

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1404 @@
...
+         std::array<uint8_t, Self::BYTES> buf;
+
+         for(;;) {
+            rng.randomize(buf);
```
reneme, line 476: A broken RNG (e.g. that produces only 1-bits) would send this into an endless loop. Is that something we would want to catch? For instance, with an upper bound of iterations of this `for` loop? We've had similar instances in the sampling code of variaous PQC algorithms.

reneme:
For the record: I'm working on some high-level overview to structure my reading. Perhaps its helpful for others as well:

image
[Excalidraw.com](https://excalidraw.com/#json=G_nQMoTSr443xfEWC3EBp,sGaR3y-BAFI_yIOY-IEzSA)

src/tests/test_pcurves.cpp
```diff
@@ -0,0 +1,304 @@
+/*
+* (C) 2014,2015,2019 Jack Lloyd
```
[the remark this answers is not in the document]
randombit: Copy(right)pasta

src/lib/math/pcurves/pcurves_wrap.h
```diff
@@ -0,0 +1,236 @@
...
+      Scalar scalar_from_u32(uint32_t x) const override { return stash(C::Scalar::from_word(x)); }
+
+      Scalar random_scalar(RandomNumberGenerator& rng) const override { return stash(C::Scalar::random(rng)); }
+
+      PrimeOrderCurveImpl() : m_mul_by_g(C::G) {}
+
+      static std::shared_ptr<const PrimeOrderCurve> instance() {
+         static auto g_curve = std::make_shared<const PrimeOrderCurveImpl<C>>();
+         return g_curve;
+      }
```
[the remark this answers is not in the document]
randombit: There is no (mutable) state in these - just the precomputed base mul table, which we certainly want to share.

I had hoped the whole table could be `constexpr`/`constinit` (in principle it could be) but clearly compilers are not anywhere close to up to it - several couldn't even handle a single field inversion before bailing out.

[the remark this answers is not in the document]
reneme: Mhm, that was my intuition as well. Yeah, I guess there's no way around the singleton then. Out of the blue: perhaps it would make sense to define the precomputed table itself a singleton (rather then the entire `PrimeOrderCurve`) to make this clearer?

src/tests/tests.cpp
```diff
@@ -857,7 +857,17 @@ std::vector<uint8_t> VarMap::get_req_bin(const std::string& key) const {
...
+         if(i->second.size() % 2 == 0) {
+            return Botan::hex_decode(i->second.substr(2));
+         } else {
+            std::string z = i->second;
+            std::swap(z[0], z[1]);  // swap 0x to x0 then remove x
+            return Botan::hex_decode(z.substr(1));
+         }
+      } else {
+         return Botan::hex_decode(i->second);
+      }
```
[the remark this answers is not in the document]
randombit: I think it can be removed at this point, I'll take a look

src/lib/math/pcurves/pcurves_impl.h
randombit: Oh indeed, thanks. I keep expecting the compiler will catch this for me.

src/lib/math/pcurves/pcurves_util.h
```diff
@@ -0,0 +1,244 @@
...
+
+   while(get_bit(b - 1) == 0) {
+      b -= 1;
+   }
+
+   return b;
+}
+
+template <WordType W, size_t N, size_t L>
+inline constexpr auto bytes_to_words(const uint8_t bytes[L]) {
```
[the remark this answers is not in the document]
randombit: I don't think there was a particular conscious decision on not using span here, probably just mental autocomplete.

src/lib/math/pcurves/pcurves.h
```diff
@@ -0,0 +1,300 @@
...
+            return PrimeOrderCurve::from_id(id.value());
+         } else {
+            return {};
+         }
+      }
+
+      static std::shared_ptr<const PrimeOrderCurve> from_id(PrimeOrderCurveId id);
+
+      /// Creates a generic non-optimized version
+      //static std::shared_ptr<const PrimeOrderCurve> from_params(...);
```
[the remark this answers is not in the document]
randombit: I think at this point a left over - removed.

It was (somewhat still is) unclear how to handle curves that are not baked in (either application specific curves but also the various "misc" curves like secp192k1, x962_p239v1, etc). One approach would be to somehow implement `PrimeOrderCurve` for these in a generic way and that's what this commented out decl was leading to.

In practice though, examining the list of curves remaining, some have some rather unappealing properties that would be difficult to fit into this model. Particularly curves where p % 4 != 3, or where the order one bit larger than the field.

So I think what's going to happen, at least in the short term (ie prior to Botan 4 where we can maybe remove some of the most unappealing builtin curves, plus remove explicit encoding of EC curves in keys/ASN.1, plus put some real restrictions on what applications use for custom curves) is that for anything that's not a builtin, we default back to the old generic BigInt approach.

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1404 @@
...
+#include <botan/internal/stl_util.h>
+#include <botan/internal/xmd.h>
+#include <vector>
+
+namespace Botan {
+
+template <typename Params>
+class MontgomeryRep final {
+   public:
+      typedef MontgomeryRep<Params> Self;
```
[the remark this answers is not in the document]
randombit: Not one I have an opinion on either way tbh, I find them about equal wrt readability so I default to typedef out of habit. I'll change to `using`

src/lib/math/pcurves/pcurves_impl.h
randombit: I'm not sure. I suppose an actually infinite loop can be unfortunate. Added a bound

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1404 @@
...
+      static const constexpr auto AW = Params::AW;
+
+      // Simplifying assumption
+      static_assert(PW.size() == NW.size());
+
+      class ScalarParams final : public IntParams<W, NW.size(), NW> {};
+
+      typedef IntMod<MontgomeryRep<ScalarParams>> Scalar;
+
+      class FieldParams final : public IntParams<W, PW.size(), PW> {};
```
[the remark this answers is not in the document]
randombit: Done

src/tests/tests.cpp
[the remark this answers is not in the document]
randombit: Consolidated the test vectors

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1428 @@
...
+      constexpr static std::array<W, N> redc(const std::array<W, 2 * N>& z) {
+         constexpr W TOP_MASK = static_cast<W>(0x1FF);
+
+         std::array<W, N> hi = {};
+         std::copy(z.begin() + N - 1, z.begin() + 2 * N - 1, hi.begin());
+         shift_right<9>(hi);
+
+         std::array<W, N> lo = {};
+         std::copy(z.begin(), z.begin() + N, lo.begin());
+         lo[N - 1] &= TOP_MASK;
```
reneme, line 109:
```suggestion
         std::array<W, N> hi = {};
         copy_mem(hi, std::span{z}.template subspan<N - 1, N>());
         shift_right<9>(hi);

         std::array<W, N> lo = {};
         copy_mem(lo, std::span{z}.template first<N>());
         lo[N - 1] &= TOP_MASK;
```

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1428 @@
...
+      template <size_t L>
+      static constexpr Self from_words(std::array<W, L> w) {
+         if constexpr(L == N) {
+            return Self(Rep::to_rep(w));
+         } else {
+            static_assert(L < N);
+            std::array<W, N> ew = {};
+            for(size_t i = 0; i != L; ++i) {
+               ew[i] = w[i];
+            }
```
reneme, line 178:
```suggestion
            copy_mem(std::span{ew}.template first<L>(), w);
```

reneme: I feel that at this point we should build a small helper like:

```C++
std::array<W, L> ew = copy_prefix<L>(w);
```

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1428 @@
...
+
+      friend constexpr Self operator-(const Self& a, const Self& b) { return a + b.negate(); }
+
+      friend constexpr Self operator*(uint8_t a, const Self& b) { return b * a; }
+
+      friend constexpr Self operator*(const Self& a, uint8_t b) {
+         // We assume b is a small constant and allow variable time
+         // computation
+
+         // In practice this function is called for 2, 3, 4, or 8
```
reneme, line 227: Frankly, I don't have an intuition about the side-channel vulnerability of this. Would it make sense to have a `BOTAN_DEBUG_ASSERT` in the generic `else` branch to ensure that the constant is "small enough"?

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1428 @@
...
+            }
+
+            return z;
+         }
+      }
+
+      friend constexpr Self operator*(const Self& a, const Self& b) {
+         std::array<W, 2 * N> z;
+         comba_mul<N>(z.data(), a.data(), b.data());
+         return Self(Rep::redc(z));
```
reneme, line 256: Nit: Perhaps implement this in terms of `operator*=` like:

```suggestion
         auto result = a;
         result *= b;
         return result;
```

... same for `operator+` and the others, of course. That would reduce code duplication a tiny bit.

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1425 @@
...
+      // Returns nullopt if the input is an encoding greater than or equal P
+      constexpr static std::optional<Self> deserialize(std::span<const uint8_t> bytes) {
+         // We could allow either short inputs or longer zero padded
+         // inputs here, however it seems best to avoid non-canonical
+         // representations unless required
+         if(bytes.size() != Self::BYTES) {
+            return {};
+         }
+
+         const auto words = bytes_to_words<W, N, BYTES>(std::span{bytes}.subspan<0, Self::BYTES>());
```
reneme, line 434:
```suggestion
         const auto words = bytes_to_words<W, N, BYTES>(bytes.first<Self::BYTES>());
```

reneme (commented):
So far I mostly looked into the structure of the implementation. We'll probably go deeper into the math next week (when @FAlbertDev is back from vacation 😰). Some more nits and suggestions I found on the way are below.

src/lib/math/pcurves/pcurves_impl.h
randombit: Well the argument is limited to a `uint8_t`. So we could say take a secret value, split it up into bytes, and then use this function. But there are easier and more efficient ways to introduce a side channel.

src/lib/math/pcurves/pcurves.h
```diff
@@ -0,0 +1,300 @@
...
+
+      static std::shared_ptr<const PrimeOrderCurve> from_name(std::string_view name) {
+         if(auto id = PrimeOrderCurveId::from_string(name)) {
+            return PrimeOrderCurve::from_id(id.value());
+         } else {
+            return {};
+         }
+      }
+
+      static std::shared_ptr<const PrimeOrderCurve> from_id(PrimeOrderCurveId id);
```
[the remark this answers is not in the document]
randombit: > Though, it does pose a limitation for potential (internal) use cases where the required curve is known at compile time.

I don't think we have many scenarios like that, at least at the moment.

It's certainly something we could do but I don't see a reason to do it preemptively lacking a clear use.

> This could (and perhaps should) be for internal use only

Keep in mind that none of this is or will be exposed to users. The public facing side of this is #4042

src/lib/math/pcurves/pcurves.cpp
```diff
@@ -0,0 +1,184 @@
...
+   "FFFFFFFF00000001000000000000000000000000FFFFFFFFFFFFFFFFFFFFFFFF",
+   "FFFFFFFF00000001000000000000000000000000FFFFFFFFFFFFFFFFFFFFFFFC",
+   "5AC635D8AA3A93E7B3EBBD55769886BC651D06B0CC53B0F63BCE3C3E27D2604B",
+   "FFFFFFFF00000000FFFFFFFFFFFFFFFFBCE6FAADA7179E84F3B9CAC2FC632551",
+   "6B17D1F2E12C4247F8BCE6E563A440F277037D812DEB33A0F4A13945D898C296",
+   "4FE342E2FE1A7F9B8EE7EB4A7C0F9E162BCE33576B315ECECBB6406837BF51F5",
+   -10> {
+};
+
+class Curve final : public EllipticCurve<Params> {};
```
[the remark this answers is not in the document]
randombit: That's one of the many tricks to reduce object size. Saves about 34KB (4% of total size) on my machine.

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1404 @@
...
+
+   private:
+      std::vector<AffinePoint> m_table;
+};
+
+template <typename C, size_t W>
+class WindowedMul2Table final {
+   public:
+      // We look at 2*W bits of scalar per iteration
+      static_assert(W >= 1 && W <= 4);
```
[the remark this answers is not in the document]
randombit: I find `static_assert` easier to parse in most of these cases, and it has the same semantic effect

src/lib/math/pcurves/pcurves.h
```diff
@@ -0,0 +1,261 @@
...
+
+            const auto& _y() const { return m_y; }
+
+            static AffinePoint _make(CurvePtr curve, StorageUnit x, StorageUnit y) {
+               return AffinePoint(std::move(curve), std::move(x), std::move(y));
+            }
+
+         private:
+            AffinePoint(CurvePtr curve, StorageUnit x, StorageUnit y) :
+                  m_curve(std::move(curve)), m_x(std::move(x)), m_y(std::move(y)) {}
```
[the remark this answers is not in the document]
randombit: Ha, fixed

src/lib/math/pcurves/pcurves_impl.h
randombit: We could also remove the generic algorithm and just fail if someone provides a value other than one of the constants used in the doubling algorithm.

I toyed with making the multiplier a compile time constant (template parameter ala `mul<3>()`) but it makes the doubling code harder to read, and doesn't have any effect on performance since the whole thing is inlined anyway.

reneme: Sounds reasonable to me.

randombit: I changed this to a few specific functions (`mul2`, `mul3` ...) which is still reasonable to read but removes any possibility of doing something silly and introducing a side channel.

src/lib/math/pcurves/pcurves.h
```diff
@@ -0,0 +1,300 @@
...
+
+            Scalar invert() const { return m_curve->scalar_invert(*this); }
+
+            bool is_zero() const { return m_curve->scalar_is_zero(*this); }
+
+            const auto& _curve() const { return m_curve; }
+
+            const auto& _value() const { return m_value; }
+
+            static Scalar _make(CurvePtr curve, StorageUnit v) { return Scalar(curve, v); }
```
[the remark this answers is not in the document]
randombit: s/_make/_create/g

src/lib/math/pcurves/pcurves.cpp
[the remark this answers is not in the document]
randombit: All of the symbols with names that are like

```
Botan::PCurve::PrimeOrderCurveImpl<Botan::PCurve::(anonymous namespace)::brainpool512r1::Curve>
```

instead become

```
Botan::PCurve::PrimeOrderCurveImpl<Botan::EllipticCurve<Botan::PCurve::(anonymous namespace)::brainpool512r1::Params, Botan::MontgomeryRep> >
```

And replicate those extra bytes across many functions on `PrimeOrderCurve`, then across 9 groups, and that's a lot of symbol data.

Per bloaty 329K (~42%) of the object file is symbols (either directly in the symbol tables, or things like relocation tables where longer symbols also increases size). My strategy so far has been

1. Run `nm` on the object file
2. Eyeball it for things that look really long and appear often.
3. Figure out a way to make that name shorter.

The real win is making the Point types shorter. I think by making them concrete types

This is currently a symbol that appears:

```
Botan::AffineCurvePoint<Botan::IntMod<Botan::MontgomeryRep<Botan::EllipticCurve<Botan::PCurve::(anonymous namespace)::brainpool384r1::Params, Botan::MontgomeryRep>::FieldParams> >, Botan::PCurve::(anonymous namespace)::brainpool384r1::Params>::x3_ax_b(Botan::IntMod<Botan::MontgomeryRep<Botan::EllipticCurve<Botan::PCurve::(anonymous namespace)::brainpool384r1::Params, Botan::MontgomeryRep>::FieldParams> > const&)
```

and it *could* be something as short as

```
Botan::PCurve::brainpool384r1::AffinePoint::x3_ax_b(Botan::PCurve::brainpool384r1::FieldElement const&)
```

Replicate that across many functions and it's a lot of symbol data.

src/lib/math/pcurves/pcurves_impl.h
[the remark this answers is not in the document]
randombit: Changed to using

src/lib/math/pcurves/pcurves_wrap.h
```diff
@@ -0,0 +1,236 @@
...
+         return from_stash(pt).serialize_to_vec(compress);
+      }
+
+      secure_vector<uint8_t> affine_point_x_bytes(const AffinePoint& pt) const override {
+         return from_stash(pt).serialize_x_coordinate();
+      }
+
+      std::vector<uint8_t> serialize_scalar(const Scalar& scalar) const override {
+         return from_stash(scalar).serialize_to_vec();
+      }
```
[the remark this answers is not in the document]
randombit: All of the inner serialization functions (in pcurves_impl.h/pcurves_wrap.h) now serialize into a `span`, and at the top level we serialize into a `concepts::resizeable_byte_buffer`

src/lib/pubkey/pcurves/pcurves_impl.h
randombit: With move to span arg we don’t need concat anymore

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1344 @@
...
+            We then perform WindowBits doublings, so accum's dlog at the point
+            of the addition in the first iteration of the loop (when i == 1) is
+            at least 2^W * w_0.
+
+            Since we know w_0 > 0, then in every iteration of the loop, accums
+            dlog will always be greater than the dlog of the table element we
+            just looked up (something between 0 and 2^W-1), and thus the
+            addition into accum cannot be a doubling.
+            */
+            accum += AffinePoint::ct_select(m_table, w_i);
```
FAlbertDev, line 1190: Is this true for all cases? Since we use blinding (with scalars bigger than the group order), we could end up with an accumulator of, for example, $accum = (order + 1)w_0$ (i.e., we have a dlog of $order + 1 \equiv 1$ ), which is effectively $w_0$.
Of course, this would be extremely unlikely.

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1344 @@
...
+            W mask_n[2 * n_words] = {0};
+
+            const auto sw = scalar.to_words();
+
+            // Compute masked scalar s + k*n
+            comba_mul<n_words>(mask_n, mask, C::NW.data());
+            bigint_add2_nc(mask_n, 2 * n_words, sw.data(), sw.size());
+
+            std::reverse(mask_n, mask_n + 2 * n_words);
+            m_bytes = store_be<secure_vector<uint8_t>>(mask_n);
```
reneme, line 1014: This would become more compact if `comba_mul` and `bigint_add2_nc` would be able to handle spans.

```suggestion
            constexpr size_t mask_words = BlindingBits / WordInfo<W>::bits;
            constexpr size_t n_words = C::NW.size();

            std::array<W, n_words> mask = {};
            std::array<W, 2 * n_words> masked_scalar = {};

            auto effective_mask = std::span{mask}.template first<mask_words>();
            load_le(effective_mask, rng.random_array<effective_mask.size_bytes()>());
            effective_mask.back() |= WordInfo<W>::top_bit;

            const auto sw = scalar.to_words();

            // Compute masked scalar s + k*n
            comba_mul<n_words>(masked_scalar.data(), mask.data(), C::NW.data());
            bigint_add2_nc(masked_scalar.data(), masked_scalar.size(), sw.data(), sw.size());

            std::reverse(masked_scalar.begin(), masked_scalar.end());
            m_bytes = store_be<secure_vector<uint8_t>>(masked_scalar);
```

... also, this assumes that we add the following overload to `RandomNumberGenerator`:

```C++
/**
* Create an array of @p N random bytes.
*
* @tparam N  number of desired random bytes
* @return    an array of @p N random bytes
* @throws Exception if RNG fails
*/
template <size_t N>
std::array<uint8_t, N> random_array() {
   std::array<uint8_t, N> result;
   random_vec(std::span{result});
   return result;
}
```

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1344 @@
...
+         } else {
+            m_bytes.resize(C::FieldElement::BYTES);
+            scalar.serialize_to(std::span{m_bytes}.template subspan<0, C::FieldElement::BYTES>());
+         }
```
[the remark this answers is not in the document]
randombit: This could happen if `BlindingBits` is explicitly set to zero for testing.

But maybe it's easier for such cases (testing performance implications of blinding) to just use`UnblindedScalarBits` in the relevant multiplication algos

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1344 @@
...
+      }
+
+      // Extract a WindowBits sized window out of s, depending on offset.
+      size_t get_window(size_t offset) const { return read_window_bits<WindowBits>(std::span{m_bytes}, offset); }
+
+   private:
+      secure_vector<uint8_t> m_bytes;
```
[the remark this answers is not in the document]
randombit: I think a stack allocation here is pretty cheap, the main motivation was to avoid putting the scalar onto the stack. But maybe we can just admit that zeroization is a lost cause and use the stack for everything.

src/lib/math/pcurves/pcurves_impl.h
randombit: Oh good point, this analysis did not consider "overflow" of the dlog of the accumulator due to blinding. Alas.

Possibly there is some other scalar multiplication algorithm we can use here, I'll have to investigate. IMO it doesn't block merging -- the situation wrt side channels here is still strictly better than with the `BigInt` based code, it even uses the same algorithm for this -- but this should be addressed.

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1344 @@
...
+};
+
+template <typename C, size_t WindowBits>
+class BlindedScalarBits final {
+   private:
+      typedef typename C::W W;
+
+      // Use 1/3 the order, rounded up to the next word for blinding
+      static const constinit size_t BlindingBits =
+         ((C::OrderBits / 3 + WordInfo<W>::bits - 1) / WordInfo<W>::bits) * WordInfo<W>::bits;
```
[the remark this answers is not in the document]
randombit: I think I just don't have enough experience in C++20 to have a good intuition wrt `constinit` vs `constexpr`

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1344 @@
...
+            Since we know w_0 > 0, then in every iteration of the loop, accums
+            dlog will always be greater than the dlog of the table element we
+            just looked up (something between 0 and 2^W-1), and thus the
+            addition into accum cannot be a doubling.
+            */
+            accum += AffinePoint::ct_select(m_table, w_i);
+
+            if(i <= 3) {
+               accum.randomize_rep(rng);
+            }
```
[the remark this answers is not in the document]
randombit: TBH no. This is not well motivated. The idea is that in the first few iterations we only have a few secret bits that have been introduced into `accum`, and so rerandomziation might be helpful. But this is extremely handwavey.

I'd be open to removing this entirely, or just doing rerandomization once during the first iteration.

If you contacts at BSI -- they may have better ideas on how to approach this.

src/lib/math/pcurves/pcurves_impl.h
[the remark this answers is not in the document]
reneme: Frankly, I'm also not sure about this. Let me read up on it and come back with a summary.  😏

src/lib/math/pcurves/pcurves_impl.h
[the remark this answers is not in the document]
reneme: Or just do it in the destructor of this helper class?

reneme:
While reviewing the multiplication table helpers, we introduced a few refactorings and added some comments for better comprehension (in #4095): https://github.com/randombit/botan/pull/4095/commits/030c919b7fb7ec959cec1a3307e61f66481919de

reneme (approved):
I like this a lot! Thanks for bearing with the load of comments we've had about that. There are certainly some more nits to be found, but I feel its more efficient if we address them in follow-ups.

randombit:
:+1: Thanks for the reviews! Some of your review comments still seem worth fixing here, also I need to clean up this dumpster fire of a commit history before merge. I'd also like #4096 to merge prior to this landing so we have some assurance wrt constant time, just in case more substantial reworking is required.

src/lib/math/pcurves/pcurves_impl.h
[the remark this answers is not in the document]
randombit: > Do we even need blinding here?

* Programmers make mistakes
* Compilers are smart (sometimes)
* Might help with EM / power analysis (questionable)

src/lib/math/pcurves/pcurves_impl.h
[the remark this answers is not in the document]
randombit: Going off https://www.cppstories.com/2022/const-options-cpp20/ I think I did want `constexpr` here

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1344 @@
...
+};
+
+template <typename C, size_t WindowBits>
+class UnblindedScalarBits final {
+   public:
+      static constexpr size_t Bits = C::Scalar::BITS;
+
+      UnblindedScalarBits(const typename C::Scalar& scalar) {
+         m_bytes.resize(C::Scalar::BYTES);
+         scalar.serialize_to(std::span{m_bytes}.template subspan<0, C::Scalar::BYTES>());
```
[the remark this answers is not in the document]
randombit: Fixed

src/lib/math/pcurves/pcurves_impl.h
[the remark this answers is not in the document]
randombit: I added an explicit boolean toggle for turning blinding on and off for such testing, easier to understand the flow

src/lib/math/pcurves/pcurves_impl.h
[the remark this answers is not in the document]
reneme: Defense in depth. Thanks for the background.

src/lib/math/pcurves/pcurves_wrap.h
```diff
@@ -0,0 +1,249 @@
...
+         return stash(from_stash(a) + from_stash(b));
+      }
+
+      ProjectivePoint point_negate(const ProjectivePoint& pt) const override { return stash(from_stash(pt).negate()); }
+
+      bool affine_point_is_identity(const AffinePoint& pt) const override { return from_stash(pt).is_identity(); }
+
+      void serialize_point(std::span<uint8_t> bytes, const AffinePoint& pt) const override {
+         BOTAN_ARG_CHECK(bytes.size() == C::AffinePoint::BYTES, "Invalid length for serialize_point");
+         from_stash(pt).serialize_to(bytes.subspan<0, C::AffinePoint::BYTES>());
```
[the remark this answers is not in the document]
randombit: Fixed

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1469 @@
...
+*
+* This is done using a windowed variant of what is usually called
+* Shamir's trick.
+*
+* The W = 1 case is simple; we precompute an extra point GH = G + H,
+* and then examine 1 bit in each of x and y. If one or the other bits
+* are set then add G or H resp. If both bits are set, add GH.
+*/
+template <typename C, size_t W>
+class WindowedMul2Table final {
```
reneme, line 1315: Fabian and I had done a refactoring on this, here: https://github.com/randombit/botan/pull/4095/commits/030c919b7fb7ec959cec1a3307e61f66481919de

I don't remember receiving feedback about this commit from you. Just making sure it doesn't fall through the cracks.
At this point it might not apply neatly anymore, though. I'll rebase later.

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -0,0 +1,1469 @@
...
+
+/**
+* Effect 2-ary multiplication ie x*G + y*H
+*
+* This is done using a windowed variant of what is usually called
+* Shamir's trick.
+*
+* The W = 1 case is simple; we precompute an extra point GH = G + H,
+* and then examine 1 bit in each of x and y. If one or the other bits
+* are set then add G or H resp. If both bits are set, add GH.
```
reneme, line 1312: In #4095 we added a bit of ASCII art for the W = 2 case. Perhaps that's worth adding:

```C++
     /**
       * The example below is a precomputed table for W=2. The flattened table
       * begins at (x_i,y_i) = (1,0), i.e. the identity element is omitted.
       * The indices in each cell refer to the cell's location in m_table.
       *
       *  x->           0          1          2         3
       *       0  |/ (ident) |0  x     |1  2x      |2  3x     |
       *       1  |3    y    |4  x+y   |5  2x+y    |6  3x+y   |
       *  y =  2  |7    2y   |8  x+2y  |9  2(x+y)  |10 3x+2y  |
       *       3  |11   3y   |12 x+3y  |13 2x+3y   |14 3x+3y  |
       */
```

reneme (approved):
Thanks for the added documentation and the co-authorship! 🤩

src/lib/math/pcurves/pcurves_impl.h
reneme: Rebased: https://github.com/randombit/botan/pull/4095/commits/b6e847e4272641d12955e7b64e73a4ad5dc54f73

src/lib/math/pcurves/pcurves_impl.h
reneme: (this is also part of https://github.com/randombit/botan/pull/3979#discussion_r1630787145)

src/lib/math/pcurves/pcurves_impl.h
randombit: Thanks for the rebase. I did see the commit but still need to think that one over.

src/lib/math/pcurves/pcurves_impl.h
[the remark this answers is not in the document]
randombit: Moved to using a `std::vector` plus a scrub during destruction. Using an array here is possible but it's a bit tricky to handle because there are leading 0 words in the mask that we'd like to omit. I left that as a todo.

src/lib/math/pcurves/pcurves_impl.h
randombit: I updated the comment to describe this case. We can move to another multiplication algorithm to avoid this.

Or just disable blinding :rocket:

src/lib/math/pcurves/pcurves_impl.h
randombit: Added this comment

src/lib/math/pcurves/pcurves_brainpool256r1/info.txt
```diff
@@ -0,0 +1,9 @@
+<defines>
+PCURVES_BRAINPOOL256R1 -> 20240608
+</defines>
+
+<module_info>
+name -> "PCurve brainpool256r1"
+brief -> "brainpool256r1"
+type -> "Internal"
```
reneme, line 8: Not 100% sure, but `type -> "Internal"` hides the module from the application user. I.e. they would be barred from enabling/disabling it via `./configure.py`. 😞

src/lib/math/pcurves/pcurves_brainpool256r1/info.txt
```diff
@@ -0,0 +1,9 @@
+<defines>
+PCURVES_BRAINPOOL256R1 -> 20240608
+</defines>
+
+<module_info>
+name -> "PCurve brainpool256r1"
+brief -> "brainpool256r1"
```
reneme, line 7: Just in case: the `brief` statement can be omitted entirely.

src/lib/math/pcurves/pcurves_secp521r1/pcurves_secp521r1.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+namespace Botan::PCurve {
+
+namespace {
+
+// clang-format off
+namespace secp521r1 {
+
+template <typename Params>
+class P521Rep final {
```
reneme, line 19: Nice! I like how this moved closer to the actual usage.

src/lib/math/pcurves/pcurves.cpp
```diff
@@ -6,178 +6,177 @@
...
+      PrimeOrderCurveId::secp384r1,
+      PrimeOrderCurveId::secp521r1,
+      PrimeOrderCurveId::secp256k1,
+      PrimeOrderCurveId::brainpool256r1,
+      PrimeOrderCurveId::brainpool384r1,
+      PrimeOrderCurveId::brainpool512r1,
+      PrimeOrderCurveId::frp256v1,
+      PrimeOrderCurveId::sm2p256v1,
+   };
+}
```
reneme, line 116: Perhaps this should only report the curves that are actually available in this build? I'm a bit on the fence about it, given that the curve-specific factories do exist regardless and just return a default value.

src/lib/math/pcurves/pcurves.cpp
```diff
@@ -6,178 +6,177 @@
...
-};
-
-class Curve final : public EllipticCurve<Params> {};
+namespace Botan::PCurve {
 
+#if !defined(BOTAN_HAS_PCURVES_SECP256R1)
+//static
+std::shared_ptr<const PrimeOrderCurve> PCurveInstance::secp256r1() {
+   return nullptr;
 }
```
reneme: The entire pcurves library is meant for internal use only, right? Otherwise, I'd really prefer a `std::optional<>` here. Missing the `nullptr` check is an easy thing to happen in downstream code. And we might even miss it in CI, if we neglect testing with a restricted build configuration. (That said: The BSI policy might be a good candidate to test this.)

randombit: Right, this is all completely for internal use with zero expectation we'll ever expose this to library users.

What happens is users (and really, even almost all uses within the library) use `EC_Group` plus `EC_Scalar` and `EC_AffinePoint` and if the right curve is in use things go fast because we divert from the current EC code to here. The main place I can see where we *might* use this interface directly if if we were implementing some new algorithm (like say FROST) and did not expect or desire it to support arbitrary curves.

Regarding `optional<shared_ptr<T>>` IMO then you just have to remember to check twice :) once for the `optional` and then the `shared_ptr`

src/lib/math/pcurves/pcurves.cpp
randombit: Yeah, I'm not sure what's more helpful. The main usage of this is for the tests though; I don't really see why we would ever call this within the library. Maybe it's best to just move this list to the tests.

Also the tests still assume that all the curve are always available

```
            auto curve = Botan::PCurve::PrimeOrderCurve::from_id(id);

            if(!curve) {
               result.test_failure("ID exists but curve does not");
               continue;
            }
```

so I'll have to fix that regardless.

src/lib/math/pcurves/pcurves_brainpool256r1/info.txt
randombit: Seems to work

```
$  ./configure.py --disable-modules=pcurves_secp256k1
...
   INFO: Skipping (disabled by user): pcurves_secp256k1
...
```

reneme: What doesn't work, however:

```
$  ./configure.py --minimized-build --enable-modules=pcurves_secp256k1
...
  ERROR: Module 'pcurves_secp256r1' is meant for internal use only
...
```

... that's inconsistent with the `--disable-modules` switch. 😓 Let's tackle that independently.

I remember that I [deliberately set up the internal modules that way](https://github.com/randombit/botan/pull/3032):

> `Internal` modules behave essentially like Public modules, but the library user cannot actively enable/disable them. They are supposed to contain code that shouldn't be interacted with directly by the library user. Frankly, I don't know of any such module apart from the recently introduced kyber_common.

... this was apparently changed later: https://github.com/randombit/botan/pull/3489 (for `--disable-modules`).

I don't have a hard objection for allowing the user to interact with internal modules, though, my initial model was that they aren't visible to the user altogether.

src/lib/math/pcurves/pcurves.cpp
reneme: Mhh, I think there's value in having this closer to the implementation, rather than moving it to the tests where it might get missed when adding curves. Perhaps just leave a docstring of warning?

src/lib/math/pcurves/pcurves_brainpool256r1/info.txt
randombit: > initial model was that they aren't visible to the user altogether.

The problem here is there are two different notions of internal we could use

* Something internal, which if enabled the user can't see (eg in a header) but which affects behavior (`pcurves_secp256r1` being enabled or not changes, for example, if hash2curve works for P-256)
* Something truly internal which only exists to satisfy some other dependency and otherwise does not have a visible change in behavior (if KMAC is not included, then the existence or absence of `cshake_xof` changes nothing, so there is no reason to allow the user to enable it)

Maybe the answer here is that none of these modules should be "Internal" since their absence *does* introduce user visible changes?

randombit:
OK there is probably a lot more that could/should be done here but I'm going to go ahead and merge

src/lib/math/pcurves/pcurves_impl.h
FAlbertDev: @randombit During our Botan 3.6.0 audit, we came back to this section, where we are wondering if this issue is resolved. For our audit, we want to argue soundly that this issue is not exploitable (which is hopefully true). Here is a summary of our general understanding of the situation:

See: [pcurves_impl.h](https://github.com/randombit/botan/blob/master/src/lib/math/pcurves/pcurves_impl/pcurves_impl.h#L1369)

This is part of the elliptic curve multiplication logic. In Botan, we use the window method for multiplication:

```
multiply_scalar_by_point(d, P):
  scalar d = | d_0 | d_1 | ... | d_(m-1) |  ( d_i w-bit windows of bits of d), d_0 > 0
  precomputed table: [0, P, 2P, 3P, ... , (2^w - 1)P]

  Q ← d_0
  for i from 1 to m-1 do:
      Q ← double_w_times(Q)
      Q ← add(Q, d_i*P) # d_i*P using the pre-computed table
  return Q
```

A potential side-channel lies in the point_add function, which is (very simplified) implemented as:

```
add(P, Q):
   if P == Q:
      return point_double(P)
   else:
      return point_add(P, Q)
```

When the scalar is smaller than the order of point $P$, it can be shown that the two operands are distinct for all calls to `add` in our multiplication method.

However, Botan uses blinding as a side-channel countermeasure. This involves adding a random multiple of the group order to the scalar used for multiplication. Consequently, the scalar becomes larger than the order of point $P$, breaking the assumption that the two operands are distinct for all calls to `add`. This scenario introduces a potential side-channel vulnerability. Unfortunately, the only ways to mitigate this side-channel are to disable blinding, use a significantly slower addition algorithm, or adopt an entirely different (and likely slower) multiplication algorithm.

We are uncertain if this side-channel is practically exploitable. Intuitively, it seems that for a random scalar and/or random blinding, the probability of the two operands being equal is cryptographically negligible. We are not aware of any research or proof supporting this intuition. Note that in practice (e.g., in ECDH), an attacker can choose any EC point for multiplication while the scalar remains secret.

What are your thoughts? Do you still have any concerns about the current implementation? Thanks!

PR #3980, opened 2024-04-07
Deprecate creation of application specific groups over 521 bits

randombit:
We have to provide an upper bound on the group in order to allow for fixed length on stack representations and 521 bits seems more than enough.

@reneme IIRC RSCS was using some application specific groups in the past, don't know if this would affect you so I wanted to flag it.

### DIFF
--- a/doc/deprecated.rst
+++ b/doc/deprecated.rst
@@ -58,6 +58,11 @@ in a future major release.
   None of the builtin groups have composite order, and in the future
   it will be impossible to create composite order EC_Groups.
 
+- Currently it is possible to create an application specific EC_Group
+  with parameters of effectively arbitrary size. In a future release
+  the maximum allowed bitlength of application provided groups will be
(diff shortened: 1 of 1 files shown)


reneme:
I'm looking into it. Personally, I find 521bits good enough as well, but I'm investigating. Might take a few days, though.

reneme:
Fair enough. Please don't let this block the release. Even in (the very unlikely) case that we'd need bigger curves, we'd come back to this.

randombit:
Doesn't have to be exactly 521 bits in the end - we'll anyway round up to word size which would lead to a practical upper bound of 544 bits.

randombit:
Merging now so the notification is included in this release, we can always increase the specific number if it proves necessary.

reneme:
For the record: no objection from RSCS regarding this deprecation. Thanks.

PR #3981, opened 2024-04-07
Deprecate support for hybrid elliptic curve point encoding

randombit:
Weirdly I can't even find out where this so called hybrid format was specified. It doesn't appear in SEC1. Maybe IEEE 1363? Anyway I've never seen any other implementation support it and it should die asap.

### DIFF
--- a/doc/deprecated.rst
+++ b/doc/deprecated.rst
@@ -54,6 +54,14 @@ in a future major release.
 
 - Kyber 90s mode is deprecated and will be removed.
 
+- Elliptic curve points can be encoded in several different ways.  The
+  most common are "compressed" and "uncompressed"; both are widely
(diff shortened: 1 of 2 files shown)


doc/deprecated.rst
```diff
@@ -54,6 +54,14 @@ in a future major release.
 
 - Kyber 90s mode is deprecated and will be removed.
 
+- Elliptic curve points can be encoded in several different ways.  The
+  most common are "compressed" and "uncompressed"; both are widely
+  used in various systems. Botan additional supports a "hybrid"
```
reneme, line 59:
```suggestion
  used in various systems. Botan additionally supports a "hybrid"
```

reneme (approved):

PR #3985, opened 2024-04-08
Introduce Public_Key::raw_public_key_bits()

reneme:
### Context

As suggested in #3706, this introduces a new top-level method for all public keys: `Public_Key::raw_public_key_bits()`, additional to `::public_key_bits()`. Similarly, `Private_Key` already provides `public_key_bits()` and `raw_public_key_bits()`.

### TODO

I'm opening this as a draft, and am very much open for feedback. For a reference of how this new method is useful see the changes in the TLS 1.3 hybrid key handling.

* [ ] Needs tests
* [ ] Look into constructing `Public_Key` objects using the raw encoding and the right algorithm and parameter specifiers where reasonable.

### Generic Description

For most algorithms this returns the canonical byte encoding of the public key's intrinsic object(s). The byte buffer of the `raw_...()` methods typically does not include any algorithm or parameter specifiers.

This is particularly useful for the algorithms of the PQC NIST competition where the public key encoding is a simple byte-concatenation of the internal objects.
Some algorithms may not have such a canonical byte serialization and may throw `Not_Implemented` in this case (e.g. RSA).

### Key Exchange Schemes

Note that key exchange schemes will return the canonical 'public_value' which wasn't easily accessible from a generic `Public_Key` or `Private_Key` before. Until now, one had to downcast the private key into a `PK_Key_Agreement_Key` and invoke `public_value()`. This was cumbersome (for instance in the implementation of hybrid PQ/T TLS 1.3 - #3609)

### Future Work

For the new PQC algorithms we currently return the same bare byte concatenation of the key for both `public_key_bits()` and `raw_public_key_bits()`. Technically this is incorrect. Earlier, we did opt to _not implement_ any pre-mature ASN.1 specification for PQC keys.

I suggest to depart from this behavior when implementing the final standards and let `public_key_bits()` either throw `Not_Implemented`, or (eventually) return a standardized ASN.1 encoding.


src/lib/tls/tls13_pqc/kex_to_kem_adapter.cpp
```diff
@@ -152,14 +112,21 @@ class KEX_to_KEM_Adapter_Encryption_Operation final : public PK_Ops::KEM_Encrypt
...
       size_t raw_kem_shared_key_length() const override { return kex_shared_key_length(m_public_key); }
 
-      size_t encapsulated_key_length() const override { return kex_public_value(m_public_key).size(); }
+      size_t encapsulated_key_length() const override {
+         // Serializing the public value into a short-lived heap-allocated
+         // vector is not ideal.
+         //
+         // TODO: Find a way to get the public value length without copying
+         //       the public value into a vector.
+         return m_public_key.raw_public_key_bits().size();
```
randombit, line 121: Couldn’t you just call this once in the constructor and save the result?

randombit (commented):
Seems fine to me. We should document somewhere the anticipated change to `public_key_bits` once an asn1 encoding for the pqc algorithms is defined.

FAlbertDev (commented):
LGTM so far :)

src/lib/tls/tls13_pqc/kex_to_kem_adapter.cpp
reneme: Certainly. Though, I don't think that'll help much. Most typical use cases will likely call this exactly once.
I somewhat had "point 5" of #3706 in mind when adding this TODO. I.e. providing a library-wide internal API to provide such magic numbers consistently. Though, admittedly, I didn't really look into the feasibility of it so far.

reneme:
> We should document somewhere the anticipated change to `public_key_bits` once an asn1 encoding for the pqc algorithms is defined.

I added some details in pubkey.rst.

src/lib/tls/tls13_pqc/kex_to_kem_adapter.cpp
randombit: I interpreted the comment more as indicating that the overhead of encoding/etc was wasteful in this context, thus my suggestion for caching the value.

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
```diff
@@ -532,7 +532,13 @@ size_t Dilithium_PublicKey::estimated_strength() const {
    return m_public->mode().nist_security_strength();
 }
 
+std::vector<uint8_t> Dilithium_PublicKey::raw_public_key_bits() const {
+   return m_public->raw_pk();
+}
+
 std::vector<uint8_t> Dilithium_PublicKey::public_key_bits() const {
+   // Technically, that's not really correct. The docstring for public_key_bits()
+   // states that it should return a BER-encoding of the public key.
```
randombit, line 541: I think the doc for `public_key_bits` is not really correct/precise. It should return the bytes that, if we stuff it into an X.509 SPKI, result in a valid public key DER encoding. For example for X25519 we just return the 32 byte encoded point with no outer encoding, because that's what RFC 8410 specifies we should do.

src/tests/test_pubkey.cpp
```diff
@@ -630,6 +630,31 @@ std::vector<Test::Result> PK_Key_Generation_Test::run() {
                            !public_key->supports_operation(Botan::PublicKeyOperation::KeyAgreement));
          }
 
+         // Test that the raw public key can be encoded. RSA is excused, as it
+         // does not have an obvious raw encoding.
+         if(public_key->algo_name() != "RSA") {
```
randombit, line 635: I'd rephrase this test slightly.

1) Always invoke `raw_public_key_bits` (for example, here we never call it on RSA, so if it simply crashed or threw some unexpected exception, we wouldn't notice).

2) The test may throw an exception. If the exception is not `Not_Implemented`, failure. If the exception is `Not_Implemented`, failure unless `algo_name() == "RSA"`.

3) Test of test as is.

src/lib/pubkey/pk_keys.h
```diff
@@ -145,6 +145,17 @@ class BOTAN_PUBLIC_API(2, 0) Public_Key : public virtual Asymmetric_Key {
...
       virtual AlgorithmIdentifier algorithm_identifier() const = 0;
 
+      /**
+      * @return binary public key bits, with no additional encoding
+      *
+      * For key agreements this is an alias for PK_Key_Agreement_Key::public_value.
+      *
+      * Note: some algorithms (for example RSA) do not have an obvious encoding
+      * for this value due to having many different values, and thus not implement
+      * this function.
```
randombit, line 155: Also mention that this function (should) always throw `Not_Implemented` in this case.

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
randombit: So the effect of the PQC algorithms returning `raw_public_key_bits` here is that we just stuff the "native" representation of the key as defined by the various reference implementations into the `BIT STRING` of the ASN.1 structure. Which is IMO the correct thing to do, but who knows what will end up getting standardized.

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
```diff
@@ -532,7 +532,13 @@ size_t Dilithium_PublicKey::estimated_strength() const {
...
    return m_public->raw_pk();
```
randombit, line 542: For cases like this where we are currently having an equivalence between `raw_public_key_bits` and `public_key_bits` I'd prefer the latter invoke the former so this relation is clear.

randombit (commented):
Another pass

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
reneme: I changed this (and the other similar) comments to:

```C++
   // Currently, there isn't a finalized definition of an ASN.1 structure for
   // Dilithium aka ML-DSA public keys. Therefore, we return the raw public key bits.
```

Regarding API stability and "the principle of least surprise": I'm hoping that a finalized implementation of ML-DSA (and the other to-be-standardized algorithms) would return a standardized ASN.1 structure here. While the pre-standard implementations stay with `raw_public_key_bits()` until they are deprecated and eventually phased out. (I'm open for better suggestions.)

reneme:
Rebased to latest `master` (resolved a few conflicts) and addressed all review remarks. This is now ready for a (hopefully) final review pass. /cc @randombit

src/tests/test_pubkey.cpp
```diff
@@ -647,6 +648,33 @@ std::vector<Test::Result> PK_Key_Generation_Test::run() {
...
+         try {
+            auto raw = public_key->raw_public_key_bits();
+            result.test_ne("raw_public_key_bits is not empty", raw.size(), 0);
+
+            if(public_key->supports_operation(Botan::PublicKeyOperation::KeyAgreement)) {
+               // For KEX algorithms, raw_public_key_bits must be equal to the canonical
+               // public value obtained by PK_Key_Agreement_Key::public_value().
+               const auto* ka_key = dynamic_cast<const Botan::PK_Key_Agreement_Key*>(&key);
+               result.require("is a key agreement private key", ka_key != nullptr);
+               result.test_eq("public_key_bits has same encoding", raw, ka_key->public_value());
```
randombit, line 663: If the dynamic cast fails this will crash with a null deref

randombit:
```
if(const auto* ka_key = dynamic_cast...) {
  result.test_eq("public_key_bits has same encoding", raw, ka_key->public_value());
} else {
  result.test_failure("Not a PK_Key_Agreement_Key");
}
```

randombit (approved):
LGTM. One minor test nit

src/tests/test_pubkey.cpp
randombit: Ah yes sorry I missed this and thought you were using `confirm` there. All good then.

PR #3987, opened 2024-04-09
Remove all 32-bit x86 inline asm related to multiprecision math

randombit:


reneme (approved):
Looks good to me, assuming that you simply want to stop supporting special code paths for 32bit x86. I find that totally reasonable.

PR #3988, opened 2024-04-09
Remove some obsolete and/or rarely used STL utilities

randombit:


src/lib/x509/asn1_alt_name.cpp
```diff
@@ -56,7 +56,8 @@ void AlternativeName::add_othername(const OID& oid, std::string_view value, ASN1
    if(value.empty()) {
       return;
    }
-   multimap_insert(m_othernames, oid, ASN1_String(value, type));
+   ASN1_String str(value, type);
+   m_othernames.insert(std::make_pair(oid, str));
```
reneme, line 60: If you use `emplace`, you could even save the `std::make_pair` and end up with more or less the same line:

```suggestion
   m_othernames.emplace(oid, ASN1_String{value, type}));
```

reneme, line 60: If you use `emplace`, you could even save the `std::make_pair` and end up with more or less the same line:

```suggestion
   m_othernames.emplace(oid, ASN1_String{value, type}));
```

src/lib/x509/asn1_alt_name.cpp
```diff
@@ -70,7 +71,9 @@ std::multimap<std::string, std::string> AlternativeName::contents() const {
    }
 
    for(const auto& othername : m_othernames) {
-      multimap_insert(names, othername.first.to_formatted_string(), othername.second.value());
+      std::string typ = othername.first.to_formatted_string();
+      std::string value = othername.second.value();
+      names.insert(std::make_pair(typ, value));
```
reneme, line 76:
```suggestion
      names.emplace(othername.first.to_formatted_string(), othername.second.value());
```

... might also save an unnecessary copy-c'tor of the `std::string` objects.

src/lib/x509/x509_dn.cpp
```diff
@@ -113,7 +113,7 @@ std::multimap<OID, std::string> X509_DN::get_attributes() const {
    std::multimap<OID, std::string> retval;
 
    for(auto& i : m_rdn) {
-      multimap_insert(retval, i.first, i.second.value());
+      retval.insert(std::make_pair(i.first, i.second.value()));
```
reneme, line 116:
```suggestion
      retval.emplace(i.first, i.second.value());
```

src/lib/x509/x509_dn.cpp
```diff
@@ -125,7 +125,7 @@ std::multimap<std::string, std::string> X509_DN::contents() const {
    std::multimap<std::string, std::string> retval;
 
    for(auto& i : m_rdn) {
-      multimap_insert(retval, i.first.to_formatted_string(), i.second.value());
+      retval.insert(std::make_pair(i.first.to_formatted_string(), i.second.value()));
```
reneme, line 128:
```suggestion
      retval.emplace(i.first.to_formatted_string(), i.second.value());
```

reneme (approved):
Nice! Using `emplace()` you could simplify things a bit further.

PR #3989, opened 2024-04-09
Add word3 type and take advantage of Clang's _BitInt extension

randombit:
Clang generates much better code with `_BitInt` than otherwise. This improves performance with Clang by ~5% for ECDSA/ECDH and 10-20% improvement for FFDHE depending on sizes. Unfortunately until Clang 16 the maximum `_BitInt` size was 128 bits, so it is only useful with newer Clang.


randombit:
@reneme Don't know if you want to review this one or not. Diff is huge but it's 99% generated code

src/lib/math/mp/mp_asmi.h
```diff
@@ -749,6 +749,77 @@ inline constexpr void word3_muladd_2(W* w2, W* w1, W* w0, W x, W y) {
    *w2 = word_add(*w2, top, &carry);
 }
 
+#if defined(__BITINT_MAXWIDTH__) && (__BITINT_MAXWIDTH__ >= 3 * 64)
+template <WordType W>
+class word3 {
+   public:
```
reneme, line 755: Perhaps add a short rationale doxygen string for future-us to remember why this has a special case?

reneme (approved):
`_BitInt()`! That's fantastic. 👀

src/lib/math/mp/mp_asmi.h
randombit: :+1:

PR #3990, opened 2024-04-09
Take advange of CT::Mask::select_n

randombit:


reneme (approved):
The CLI tests failed on macOS w/ Apple Silicon with return code -11. I don't think this is related, but pointer arithmetics and segfaults in one sentence... 🤔

randombit:
@reneme I've seen that this morning on a few different PRs now #3991

PR #3992, opened 2024-04-09
Update copyright year in license.txt

jiep:


randombit:
Fixed on master now - thanks for pointing this out

PR #3994, opened 2024-04-09
Refactor: concat() is constexpr and can deal with std::array<>

reneme:
This reworks the `concat()` helper, giving it similar smartness as `load_le` and friends. Most notably, it is now `constexpr` and gains support for statically sized containers (`std::array<>` and `std::span<T, size>`). Also, the desired output container can now be defined without using the (now removed) `concat_as<>` crutch.

The output container to use is determined along the following rules:
1. explicitly specified by the caller: `concat<std::vector<uint8_t>>(...)` or `concat<std::array<uint8_t, 32>>(...)`,
2. if all input containers have static extent, the output container is a `std::array` with the appropriate size,
3. if the first input container is dynamically allocated, use this one (that's how the old `concat` handled it),
4. fail compilation and urge the caller to explicitly specify an output container.

Example usages:

```C++
constexpr std::array<uint8_t, 4> a1{1,2,3,4};
constexpr std::array<uint8_t, 2> a2{5,6};
std::vector<uint8_t> v1{7,8,9};

constexpr std::array<uint_8_t, 6> c1 = concat(a1,a2); // {1,2,3,4,4,5}
std::vector<uint8_t> c2 = concat(v1, a2);             // {7,8,9,5,6}
auto c3 = concat<std::array<uint8_t, 4>>(v1, a2);     // runtime exception: size mismatch (4 != 5)
auto c4 = concat<StrongVector>(v1, a1);               // {7,8,9,1,2,3,4}
// ...
```

Currently, this implementation cannot handle range elements of move-only types (as it uses `std::copy()` internally). Technically, this could probably be resolved, but I refrained from doing that as we mostly concatenate strings anyway. Nevertheless, its worth to keep that in mind and extend it when necessary.

### DIFF
--- a/src/lib/utils/concepts.h
+++ b/src/lib/utils/concepts.h
@@ -33,9 +33,9 @@ struct is_strong_type<Strong<Ts...>> : std::true_type {};
 template <typename... Ts>
 constexpr bool is_strong_type_v = is_strong_type<std::remove_const_t<Ts>...>::value;
 
-template <typename T0, typename... Ts>
+template <typename T0 = void, typename... Ts>
 struct all_same {
-      static constexpr bool value = (std::is_same_v<T0, Ts> && ...);
+      static constexpr bool value = (std::is_same_v<T0, Ts> && ... && true);
 };
 
 template <typename... Ts>
--- a/src/lib/utils/concepts.h
+++ b/src/lib/utils/concepts.h
@@ -41,6 +41,18 @@ struct all_same {
 template <typename... Ts>
 static constexpr bool all_same_v = all_same<Ts...>::value;
 
+namespace detail {
+
+/**
+ * Helper type to indicate that a certain type should be automatically
+ * detected based on the context.
+ */
+struct AutoDetect {
+      constexpr AutoDetect() = delete;
+};
+
+}  // namespace detail
+
 namespace ranges {
 
 /**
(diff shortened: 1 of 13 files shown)


src/lib/utils/concepts.h
```diff
@@ -33,14 +33,26 @@ struct is_strong_type<Strong<Ts...>> : std::true_type {};
 template <typename... Ts>
 constexpr bool is_strong_type_v = is_strong_type<std::remove_const_t<Ts>...>::value;
 
-template <typename T0, typename... Ts>
+template <typename T0 = void, typename... Ts>
 struct all_same {
-      static constexpr bool value = (std::is_same_v<T0, Ts> && ...);
+      static constexpr bool value = (std::is_same_v<T0, Ts> && ... && true);
```
randombit, line 38: What's up with this?

randombit (approved):
This is really nice, thanks.

Left one technical question just for my own understanding.

Also I noticed in the library src itself you avoid using the autodetection, instead always specifying the output type. Is there a reason for this?

src/lib/utils/concepts.h
reneme: That's a default value, if the parameter pack (`T0`) is empty.

reneme:
> Also I noticed in the library src itself you avoid using the autodetection, instead always specifying the output type. Is there a reason for this?

The changes in this pull request in the usage locations remove the `concat_as<>` overload. That might give this impression. However, `concat_as<>` was specifically added to allow overriding the auto-detected default: mostly because the first component had an unfavorable type. Now, with the addition of the `AutoDetect` meta-type we don't need a dedicated "override" function anymore.

There are plenty places where no explicit out-type is given, they just don't show up in this diff.

PR #3996, opened 2024-04-10
Fix enabling stack protector

Kentzo:
The action needs to be `store_true` for the option to work as intended.


randombit (approved):
Thanks!

PR #3997, opened 2024-04-10
Add helper for defining binary option pairs

randombit:
Inspired by #3996 in the hope of preventing future errors of this type.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -318,6 +318,26 @@ def process_command_line(args):
     available before logging is setup.
     """
 
+    def define_option_pair(group, verb, nverb, what, default, with_help=None, without_help=None):
+        dest = '%s_%s' % (verb, what.replace('-', '_'))
+
+        group.add_option('--%s-%s' % (verb, what),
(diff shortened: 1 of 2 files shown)


reneme (approved):
Good idea to make this type of error less likely.

Note: I pushed a suggestion that simplifies the help-text handling somewhat. Instead of a `with` and `without` param, there's now only a single `help` param that is automagically assigns the help text to the option that switches away from the default. That's the convention anyway, and I think it makes sense to set it in stone.

Please feel free to squash or drop that suggestion, of course. I just thought it'd be easiest to just go ahead and do it instead of outlining it in prose.

randombit:
Ah yeah nice improvement, I wasn't so happy with passing two variables when we never use them both. Your logic is slightly off because we have a couple of cases where we use `None` as the default, as a signal that the user provided no opinion (absence of both `--with-foo` *and* `--without-foo`) that instead we should autodetect `foo`. And Python considers `not None == True`

PR #3999, opened 2024-04-10
Move checked integer arithmetic to int_utils.h

randombit:


src/lib/utils/int_utils.h
```diff
@@ -0,0 +1,57 @@
...
+}
+
+template <std::unsigned_integral T, std::unsigned_integral RT>
+std::optional<RT> checked_cast(T input) {
+   const RT cast = static_cast<RT>(input);
+   if(static_cast<T>(cast) != input) {
+      return {};
+   }
+   return cast;
+}
```
reneme, line 53: When the user is supposed to partially define the template at the call-site, the to-be-defined template parameters should come first, no?

If we wanted to overengineer this a little: C++20 seems to allow mixing fully-fledged template parameters and the new `auto` parameter syntax. That way, the user is forced to rely on type deduction for the function parameter, and also there's no confusion in the order of the template parameters:

```suggestion
template <std::unsigned_integral ReturnT>
std::optional<ReturnT> checked_cast(std::unsigned_integral auto input) {
   if(std::numeric_limits<ReturnT>::max() < input) {
      return {};
   }
   return static_cast<ReturnT>(input);
}
```

(Note that this might require `#include <limits>`. Or did you have something specific in mind when implementing it using the reversed cast for the check?)

reneme (approved):

src/lib/utils/int_utils.h
randombit: > When the user is supposed to partially define the template at the call-site, the to-be-defined template parameters should come first, no?

Yes - I did this as a cleanup of something that existed but was unused in `safeint.h` and then didn't think about the template parameter ordering.

> Or did you have something specific in mind when implementing it using the reversed cast for the check?

No, that's just how the preexisting code did it and it seemed reasonable. Your approach seems to lead to an interface that's easier to use correctly.

src/tests/test_utils.cpp
```diff
@@ -47,6 +51,81 @@ class Utility_Function_Tests final : public Test {
...
+            auto sum2 = Botan::checked_add<size_t>(large, zero, zero, zero, i);
+
+            result.confirm("checked_add looks at all args", sum1 == sum2);
+
+            if(i < 5) {
+               result.test_eq("checked_add worked", sum1.value(), i + large);
+            } else {
+               result.confirm("checked_add did not return a result", !sum1.has_value());
+            }
+         }
```
reneme, line 71: That won't ever hit the failure case, no?

```suggestion
         for(size_t i = 0; i <= 5; ++i) {
            auto sum1 = Botan::checked_add<size_t>(i, zero, zero, zero, large);
            auto sum2 = Botan::checked_add<size_t>(large, zero, zero, zero, i);

            result.confirm("checked_add looks at all args", sum1 == sum2);

            if(i < 5) {
               result.test_eq("checked_add worked", sum1.value(), i + large);
            } else {
               result.confirm("checked_add did not return a result", !sum1.has_value());
            }
         }
```

reneme (changes requested):

src/tests/test_utils.cpp
randombit: Fixed! This turned out to expose the fact that `checked_add` did not behave correctly. The check was

`if(r - b != a)`

but if `a + b` overflowed in one direction, the subtraction would just overflow back...

I wrote it in mental analogy to * vs /, where that works.

reneme (approved):

PR #4006, opened 2024-04-12
Small cleanup to the freestanding argon2 function

randombit:


src/lib/pbkdf/argon2/argon2.h
```diff
@@ -155,20 +155,21 @@ inline void argon2(uint8_t output[],
...
+         case 1:
+            return "Argon2i";
+         case 2:
+            return "Argon2id";
+         default:
+            throw Not_Implemented("Unknown Argon2 family type");
+      }
+   };
+
+   auto pwdhash_fam = PasswordHashFamily::create_or_throw(y_to_argon2(y));
```
reneme, line 171: I did start using anonymous lambdas for things like that. No hard feelings either way, but I find the idiom would fit here quite well:

```suggestion
   auto pwdhash_fam = PasswordHashFamily::create_or_throw([y] {
      switch(y) {
         case 0:
            return "Argon2d";
         case 1:
            return "Argon2i";
         case 2:
            return "Argon2id";
         default:
            throw Not_Implemented("Unknown Argon2 family type");
      }
   }());
```

reneme (approved):

src/lib/pbkdf/argon2/argon2.h
randombit: That's nice

PR #4008, opened 2024-04-12
Add constexpr shift_right mp helper

randombit:


src/lib/math/mp/mp_core.h
```diff
@@ -845,15 +845,29 @@ inline constexpr W shift_left(std::array<W, N>& x) {
...
+template <size_t S, WordType W, size_t N>
+inline consteval W shift_right(std::array<W, N>& x) {
+   static_assert(S < WordInfo<W>::bits, "Shift too large");
+
+   W carry = 0;
+   for(size_t i = 0; i != N; ++i) {
+      const W w = x[N - 1 - i];
+      x[N - 1 - i] = (w >> S) | carry;
+      carry = w << (WordInfo<W>::bits - S);
+   }
```
reneme, line 866: If we'd let the loop run backwards from `i = N - 1` to `i >0`, would that make the indices easier to understand?

reneme (approved):

src/lib/math/mp/mp_core.h
randombit: I agree the indices would be clearer going in the other direction. But for whatever reason I find the pattern of consistently mapping from `i \in [0..N)` to an action on that `i` easier to follow.

PR #4009, opened 2024-04-12
Debugging PCurves mac failure

randombit:


randombit:
Conclusion: `-O3` with XCode 15.2 on x86-64 causes sadness

15.3 on Aarch64 is ok

Is it a bug that was fixed in 15.3? Or just a bug that only affects x86 builds? ¯\\_(ツ)_/¯

Not entirely sure how to proceed short of just using -O2. Not interested in paying Apple $99 to get a dev account in order to report a bug they'll probably ignore.

PR #4010, opened 2024-04-12
Distinguish between LLVM Clang and XCode Clang

randombit:


reneme:
That idea makes a lot of sense! Let me know if I should take this up. Judging from the commit messages you don't have a Mac handy. ;)

reneme:
Locally, this seems to work for me. Would probably make sense to have a sharp look at all of the flags in xcode.txt.

randombit:
OK things are looking good to me at this point.

randombit:
I'll squash before merging

PR #4012, opened 2024-04-12
Rename Curve25519 to X25519

randombit:


reneme (approved):
Nice! Consistency with X448 restored, I guess. :)

Regarding semver: Names of modules aren't covered in the semver guarantees?
This would now cause issues across versions in 3.x:
```
./configure.py --minimized-build --enable-modules=curve25519
```

... not sure whether this justifies a 'module alias mechanism' of some sort, though.

randombit:
No - building the library (module names, `configure.py` options, module policy details) has never been covered by SemVer. Only using the resulting binary. This has "always" been the case and I know there have been similar breaks in the past though I don't recall the details off the top of my head. The written SemVer policy should make all this clear though.

PR #4024, opened 2024-04-19
Refactor: Have a common base for Kyber/Dilithium structures

reneme:
This is essentially a complete overhaul of the Kyber and Dilithium implementations aiming at as much shared code as possible between the two algorithms. The rough structure is outlined below. I split-up the changes into three commits, that could be separate pull requests, if needed.

### Module: `pqcrystals`

* **Polynomial Structures**
  This is a templated base for the polynomials (as well as poly vectors/matrices) used in both implementations. The base implementation shares math where possible and is pluggable where necessary. It also exposes its domain (NTT or "normal") in the type system. Also, it makes some initial optimization effort (mostly memory layout), but more work could be done here.
* **Polynomial Encoding/Decoding**
  Both algorithms contain bit-packing algorithms to efficiently encode polynomial coefficients of various known value ranges. Both Kyber and Dilithium now share a generic and pluggable implementation of this.

### Module: `kyber`

This now makes use of the `pqcrystals` module. The polynomial instantiation happens in `kyber_polynomial.h` (replacing the old `kyber_structures.h`). Most low-level "algorithms" (pseudo-codes of the spec) live in `kyber_algos.cpp`. The indcpa "K-PKE" scheme is implemented in `kyber_keys.cpp` and the cca2-secure wrapper can be found in `kyber.cpp`. Auxiliary stuff like constants and strong types are in other headers.

Note that this now contains almost no verbatim code from the reference implementation anymore.

### Module: `dilithium`

Likewise, this uses the `pqcrystals` module (instantiating the polynomials in `dilithium_polynomial.h` (replacing the almost verbatim reference implementation code). The `dilithium_algos.cpp` is home to most low-level "algorithms" (as seen in the spec). `dilithium.cpp` houses the actual signature/verification algorihtms that bind it all together..

Care has been taken to have a consistent implementation structure between Kyber and Dilithium. Note, though, that Dilithium's flexibility might require some love before a clean integration of ML-KEM (similar to what happened to Kyber in #3887).

### Benchmark

~~Despite not explicitly aiming to increase speed,~~ `./botan speed` shows nice gains for Kyber and no losses for Dilithium. ~~It might be worth looking into why Kyber is so much faster, though. Feels suspicious.~~ Edit: The new implementation pre-computes and stores the expanded public matrix A. So as long as one re-uses the `PK_KEM_Encryptor` for multiple encapsulations, one saves _a lot of_ invocations to the XOF. That explains. 😏

GCC 13.2 vs. Clang 18.1  on Linux (in WSL)

| Algorithm            | This PR (GCC)    | Master (GCC)      | This PR (Clang)     | Master (Clang)   |
| -------------------- | ---------------- | ----------------- | ------------------- | ---------------- |
| Dilithium-4x4-r3     | 4602 sign/sec    | 4332 sign/sec     | 5358 sign/sec       | 4681 sign/sec    |
| Dilithium-4x4-r3     | 28913 verify/sec | 31303 verify/sec  | 37226 verify/sec    | 35459 verify/sec |
| Dilithium-4x4-AES-r3 | 5050 sign/sec    | 4689 sign/sec     | 6363 sign/sec       | 5007 sign/sec    |
| Dilithium-4x4-AES-r3 | 27824 verify/sec | 31180 verify/sec  | 38452 verify/sec    | 34830 verify/sec |
| Dilithium-6x5-r3     | 2648 sign/sec    | 2695 sign/sec     | 2893 sign/sec       | 2830 sign/sec    |
| Dilithium-6x5-r3     | 19250 verify/sec | 20999 verify/sec  | 26201 verify/sec    | 24643 verify/sec |
| Dilithium-6x5-AES-r3 | 2848 sign/sec    | 2850 sign/sec     | 3398 sign/sec       | 3033 sign/sec    |
| Dilithium-6x5-AES-r3 | 18503 verify/sec | 21491 verify/sec  | 26030 verify/sec    | 24351 verify/sec |
| Dilithium-8x7-r3     | 2389 sign/sec    | 2251 sign/sec     | 2853 sign/sec       | 2543 sign/sec    |
| Dilithium-8x7-r3     | 12768 verify/sec | 14551 verify/sec  | 18622 verify/sec    | 17211 verify/sec |
| Dilithium-8x7-AES-r3 | 2758 sign/sec    | 2490 sign/sec     | 3044 sign/sec       | 2970 sign/sec    |
| Dilithium-8x7-AES-r3 | 13856 verify/sec | 13881 verify/sec  | 18468 verify/sec    | 17629 verify/sec |

| Algorithm         | This PR (GCC)      | Master (GCC)       | This PR (Clang)   | Master (Clang)    |
| ----------------- | ------------------ | ------------------ | ----------------- | ----------------- |
| Kyber-512-r3      | 69864  encrypt/sec | 39354  encrypt/sec | 59956 encrypt/sec | 38009 encrypt/sec |
| Kyber-512-r3      | 51452  decrypt/sec | 32157  decrypt/sec | 48114 decrypt/sec | 32214 decrypt/sec |
| Kyber-512-90s-r3  | 81350  encrypt/sec | 39878  encrypt/sec | 75388 encrypt/sec | 44527 encrypt/sec |   
| Kyber-512-90s-r3  | 57531  decrypt/sec | 32740  decrypt/sec | 57515 decrypt/sec | 37094 decrypt/sec |   
| Kyber-768-r3      | 50542  encrypt/sec | 22767  encrypt/sec | 45523 encrypt/sec | 21180 encrypt/sec |
| Kyber-768-r3      | 37306  decrypt/sec | 19418  decrypt/sec | 36505 decrypt/sec | 18593 decrypt/sec |
| Kyber-768-90s-r3  | 48522  encrypt/sec | 22167  encrypt/sec | 59785 encrypt/sec | 24793 encrypt/sec |   
| Kyber-768-90s-r3  | 34923  decrypt/sec | 19070  decrypt/sec | 46596 decrypt/sec | 21613 decrypt/sec |   
| Kyber-1024-r3     | 35053  encrypt/sec | 14597  encrypt/sec | 40132 encrypt/sec | 14884 encrypt/sec | 
| Kyber-1024-r3     | 26609  decrypt/sec | 12840  decrypt/sec | 32397 decrypt/sec | 13338 decrypt/sec | 
| Kyber-1024-90s-r3 | 36312  encrypt/sec | 14504  encrypt/sec | 43981 encrypt/sec | 16118 encrypt/sec |    
| Kyber-1024-90s-r3 | 27100  decrypt/sec | 13312  decrypt/sec | 34729 decrypt/sec | 14435 decrypt/sec |


reneme:
clang's UBsan is complaining about a signed integer overflow in Kyber's montgomery reduction. Did we drop a `reduce()` somewhere?

```
/home/runner/work/botan/botan/build/build/include/internal/botan/internal/kyber_constants.h:50:42: runtime error: signed integer overflow: 3611157 * -3327 cannot be represented in type 'int'
  SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/runner/work/botan/botan/build/build/include/internal/botan/internal/kyber_constants.h:50:42 in 
  Ran for 354 seconds
```

reneme:
FYI: I won't continue working on this before end of May. ⏳

reneme:
This still has a few rough edges (e.g. some test cases cause out-of-bounds array accesses in static analysis). But I'm confident and happy this is finally converging.

src/lib/pubkey/pqcrystals/pqcrystals_encoding.h
```diff
@@ -0,0 +1,220 @@
...
+
+constexpr auto as_byte_source(Botan::XOF& xof) {
+   return [&](std::span<uint8_t> out) { xof.output(out); };
+}
+
+}  // namespace detail
+
+template <typename T>
+concept byte_source =
+   requires(T& t) { requires std::invocable<decltype(detail::as_byte_source(t)), std::span<uint8_t>>; };
```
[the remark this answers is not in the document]
reneme: From a technical perspective, one might even implement it like shown here: As a C++20 concept that tries to build a thin wrapper around "the source". That way, new data sources could just hook themselves into this "helper" and decide for themselves to be a "data source" or not.

src/lib/pubkey/kyber/kyber_common/kyber.cpp
```diff
@@ -207,25 +206,27 @@ Kyber_PrivateKey::Kyber_PrivateKey(RandomNumberGenerator& rng, KyberMode m) {
...
+   KyberPolynomialSampler ps(sigma, mode);
 
-   auto a = PolynomialMatrix::generate(rho, false /* not transposed */, mode);
-   auto s = PolynomialVector::getnoise_eta1(sigma, 0 /* N */, mode);
-   auto e = PolynomialVector::getnoise_eta1(sigma, mode.k() /* N */, mode);
+   // TODO: Remove the need for the montgomery transformation
+   //
+   // -> When calculating A*s below, A is not in montgomery form, but s is. The
+   //    operation uses fqmul internally, which performs a montgomery reduction.
+   auto A = montgomery(kyber_sample_matrix(rho, false /* not transposed */, mode));
```
reneme, line 217: This was different in the reference implementation. @FAlbertDev we did look into this a while ago, I feel we should take this up again and get to the bottom of it.

src/lib/pubkey/kyber/kyber_common/kyber_types.h
```diff
@@ -14,12 +14,21 @@
...
 #include <vector>
 
 namespace Botan {
 
+using KyberPolyNTT = Botan::CRYSTALS::Polynomial<KyberPolyTraits, Botan::CRYSTALS::Domain::NTT>;
+using KyberPolyVecNTT = Botan::CRYSTALS::PolynomialVector<KyberPolyTraits, Botan::CRYSTALS::Domain::NTT>;
+using KyberPolyMat = Botan::CRYSTALS::PolynomialMatrix<KyberPolyTraits>;
+
+using KyberPoly = Botan::CRYSTALS::Polynomial<KyberPolyTraits, Botan::CRYSTALS::Domain::Normal>;
+using KyberPolyVec = Botan::CRYSTALS::PolynomialVector<KyberPolyTraits, Botan::CRYSTALS::Domain::Normal>;
```
reneme, line 30: As mentioned before: it would be good if these could be the basis for strong types to disambiguate them in interfaces E.g. to replace the multiple `poly_pack_t0`, `..._t1`, ... methods in `dilithium_algos.cpp` with an overloaded `poly_pack()` that just "does the right thing" ™ based on the strong type.

src/lib/pubkey/pqcrystals/pqcrystals_helpers.h
```diff
@@ -0,0 +1,143 @@
...
+   if(a != b) {
+      while(a != 0) {
+         const T q = b / a;
+         std::tie(a, b) = std::make_tuple(b - q * a, a);
+         std::tie(u1, v1, u2, v2) = std::make_tuple(u2, v2, u1 - q * u2, v1 - q * v2);
+      }
+   }
+
+   return {.gcd = b, .u = u1, .v = v1};
+}
```
reneme, line 83: Perhaps, this exists somewhere else already? Or would be useful somewhere else? Same with the monty-param precalculations above.

src/lib/xof/xof.h
```diff
@@ -158,6 +158,16 @@ class BOTAN_PUBLIC_API(3, 2) XOF {
...
+      /**
+       * @return the next @p count output bytes as a std::array<>.
+       */
+      template <size_t count>
+      std::array<uint8_t, count> output() {
+         std::array<uint8_t, count> out;
+         generate_bytes(out);
+         return out;
+      }
+
```
reneme, line 170: This type of "data source" method is really useful, we should probably add it to other similar classes. E.g. the RNG. That would allow things like: `uint32_t i = load_le(rng.random_array<4>())`. Now this would be:

```C++
std::array<uint8_t, 4> buffer;
rng.randomize(buffer);
uint32_t i = load_le(buffer);
```

src/lib/pubkey/dilithium/dilithium_common/dilithium_constants.h
```diff
@@ -0,0 +1,171 @@
...
+
+      enum DilithiumGamma1 : uint32_t { ToThe17th = (1 << 17), ToThe19th = (1 << 19) };
+
+      enum DilithiumGamma2 : uint32_t { Qminus1DevidedBy88 = (Q - 1) / 88, Qminus1DevidedBy32 = (Q - 1) / 32 };
+
+      enum DilithiumEta : uint32_t { _2 = 2, _4 = 4 };
+
+      enum DilithiumBeta : uint32_t { _78 = 78, _196 = 196, _120 = 120 };
+
+      enum DilithiumOmega : uint32_t { _80 = 80, _55 = 55, _75 = 75 };
```
reneme, line 75: Dilithium has many integer constants that change depending on the instantiation of the algorithm. I found it worthwhile to restrict them with explicit enums. I'm okay with this concept, also given the restricted (algorithm-internal) scope. But I feel there must be a better way for this. :)

src/lib/pubkey/dilithium/dilithium_common/dilithium_algos.cpp
```diff
@@ -0,0 +1,781 @@
...
+
+   for(const auto& p : s2) {
+      dilithium_poly_pack_eta(p, stuffer, mode);
+   }
+
+   for(const auto& p : t0) {
+      dilithium_poly_pack_t0(p, stuffer);
+   }
+
+   BOTAN_ASSERT_NOMSG(stuffer.full());
```
reneme, line 349: Those became ubiquitous and at this point I do start wondering whether we should just put that in the destructors of `Stuffer` and `Slicer` (perhaps explicitly disengagable). What do y'all think?

reneme (commented):
Below are a few highlights and potential conversion starters of things I saw along the way. :)

src/lib/pubkey/dilithium/dilithium_common/dilithium_algos.cpp
randombit: Sounds good to me.

src/lib/pubkey/pqcrystals/pqcrystals_helpers.h
FAlbertDev: Yeah, quite surprising if it is not implemented anywhere else. The only EEA I found is within the old mceliece implementation which uses an EEA (`polyn_gf2m::eea_with_coefficients`) on GF elements.

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
```diff
@@ -521,19 +358,19 @@ AlgorithmIdentifier Dilithium_PublicKey::algorithm_identifier() const {
 }
 
 OID Dilithium_PublicKey::object_identifier() const {
-   return m_public->mode().oid();
+   return m_public->mode().mode().object_identifier();
 }
 
 size_t Dilithium_PublicKey::key_length() const {
    return m_public->mode().public_key_bytes();
 }
```
FAlbertDev, line 366: I get confused by this functions every single time...
```cpp
/**
 * Return an integer value best approximating the length of the
 * primary security parameter. For example for RSA this will be
 * the size of the modulus, for ECDSA the size of the ECC group,
 * and for McEliece the size of the code will be returned.
 */
virtual size_t key_length() const = 0;
```
I'm not sure which value fits best here. We should, at least, provide some size in bits.
Same for Kyber.

src/lib/pubkey/dilithium/dilithium_common/dilithium_symmetric_primitives.h
```diff
@@ -12,198 +12,140 @@
...
+         scoped_cleanup clean([this]() { m_xof.clear(); });
+         m_xof.update(seed);
+
+         // Note: The order of invocations in an initializer list is not
+         //       guaranteed by the C++ standard. Hence, we have to store the
+         //       results in variables to ensure the correct order of execution.
+         auto rho = m_xof.output<DilithiumSeedRho>(DilithiumConstants::SEED_RHO_BYTES);
+         auto rhoprime = m_xof.output<DilithiumSeedRhoPrime>(DilithiumConstants::SEED_RHOPRIME_BYTES);
+         auto k = m_xof.output<DilithiumSigningSeedK>(DilithiumConstants::SEED_SIGNING_KEY_BYTES);
+         return {std::move(rho), std::move(rhoprime), std::move(k)};
```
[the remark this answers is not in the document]
reneme: The idea is to make the cleanup exception-safe. Arguably, if an exception is actually thrown, it is likely due to some assertion, out-of-memory or some other byzantine fault and then all bets are off, anyway.

Fair point, its probably not worth the extra conplexity.

src/lib/pubkey/pqcrystals/pqcrystals_encoding.h
```diff
@@ -0,0 +1,224 @@
...
+   // This is the inverse operation of the bit-packing algorithm above. Please
+   // refer to the comments there for a detailed explanation of the algorithm.
+   for(size_t i = 0; i < p.size(); i += trait::coeffs_per_pack) {
+      get_bytes(std::span{bytes}.template first<trait::bytes_per_pack>());
+      const auto collectors = load_le<typename trait::collector_array>(bytes);
+
+      for(size_t j = 0, bit_offset = 0, c = 0; j < trait::coeffs_per_pack; ++j) {
+         typename trait::sink_t coeff_value = collectors[c] >> bit_offset;
+         bit_offset += trait::bits_per_coeff;
+         if(bit_offset > trait::bits_in_collector) {
```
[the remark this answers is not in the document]
reneme: See here: https://github.com/randombit/botan/pull/4024#discussion_r1653120130

src/lib/pubkey/dilithium/dilithium_common/dilithium_algos.cpp
```diff
@@ -0,0 +1,781 @@
...
+template <DilithiumConstants::T b>
+constexpr std::make_unsigned_t<DilithiumConstants::T> map_range(DilithiumConstants::T c) {
+   BOTAN_DEBUG_ASSERT(b - c >= 0);
+   return b - c;
+}
+
+template <DilithiumConstants::T b>
+constexpr DilithiumConstants::T unmap_range(std::make_unsigned_t<DilithiumConstants::T> c) {
+   return static_cast<DilithiumConstants::T>(b - c);
+}
```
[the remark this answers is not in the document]
reneme: Yeah, I was surprised by that as well.

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
```diff
@@ -114,402 +81,272 @@ std::string DilithiumMode::to_string() const {
...
-            sig[position + mode.omega() + i] = static_cast<uint8_t>(k);
-         }
-         return sig;
-      }
-
-      const Dilithium_PrivateKey m_priv_key;
-      const Dilithium::PolynomialMatrix m_matrix;
-      SHAKE_256 m_shake;
+      std::shared_ptr<Dilithium_PrivateKeyInternal> m_priv_key;
+      const DilithiumConstants& m_mode;
```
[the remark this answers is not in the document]
reneme: Nu kloar!

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
```diff
@@ -114,402 +81,272 @@ std::string DilithiumMode::to_string() const {
...
 
-      std::string hash_function() const override { return "SHAKE-256(512)"; }
+      std::string hash_function() const override { return m_h.name(); }
 
    private:
       std::shared_ptr<Dilithium_PublicKeyInternal> m_pub_key;
-      const Dilithium::PolynomialMatrix m_matrix;
-      const std::vector<uint8_t> m_pk_hash;
-      SHAKE_256 m_shake;
+      const DilithiumConstants& m_mode;
```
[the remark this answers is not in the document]
reneme: Ja.

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
```diff
@@ -114,402 +81,272 @@ std::string DilithiumMode::to_string() const {
...
-         SHAKE_256 shake256_variable(DilithiumModeConstants::SEEDBYTES * 8);
-         shake256_variable.update(mu.data(), mu.size());
-         shake256_variable.update(packed_w1.data(), packed_w1.size());
-         auto c2 = shake256_variable.final();
-
-         BOTAN_ASSERT_NOMSG(c.size() == c2.size());
-         return std::equal(c.begin(), c.end(), c2.begin());
+         const auto c_hat = ntt(dilithium_sample_in_ball(c1, m_mode));
+         auto w_approx = m_A * ntt(std::move(z));
+         w_approx -= c_hat * ntt(m_pub_key->t1() << DilithiumConstants::D);
```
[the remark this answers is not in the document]
reneme: This is worth it, for sure! I did some profiling (mostly Kyber, but whatever) and `ntt()` was one of the most expensive operations. Here's some benchmark:

| Algorithm             | precomputed ntt(t1) | baseline         |
| --------------------- | ------------------- | ---------------- |
| Dilithium-4x4-r3      | 44824 verify/sec    | 35705 verify/sec |
| Dilithium-4x4-AES-r3  | 42568 verify/sec    | 35049 verify/sec |
| Dilithium-6x5-r3      | 28036 verify/sec    | 24382 verify/sec |
| Dilithium-6x5-AES-r3  | 26499 verify/sec    | 24993 verify/sec |
| Dilithium-8x7-r3      | 19843 verify/sec    | 16034 verify/sec |
| Dilithium-8x7-AES-r3  | 21071 verify/sec    | 17046 verify/sec |

That's a nice 20% speedup right there! Very good catch. 👍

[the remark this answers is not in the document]
reneme: I did the pre-computation in the `Dilithium_Verification_Operation` along with the precomputation of A and the preparation of the message hash.

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
reneme: For Kyber, we could return the ML-KEM-x values: i.e. {512, 768, 1024}. What do these even mean though? My best guess: `k*N` as the coefficients in the private vector `s`?! At least that has some foundation in the names of the specified parameter sets, but its not a bit length. We could multiply those by 12 as the bit-length of Q, but that will result in fairly arbitrary values {6144, 9216, 12288}.

For Dilithium, we could go for the bit lengths of `s1` and `s2` perhaps?

@randombit Do you have any preference?

src/lib/rng/rng.h
```diff
@@ -14,6 +14,7 @@
 #include <botan/mutex.h>
 #include <botan/secmem.h>
 
+#include <array>
```
FAlbertDev, line 17: Remove

src/tests/test_crystals.cpp
```diff
@@ -0,0 +1,483 @@
...
+                            auto pv2_ntt = pv_ntt.clone();
+                            for(const auto& poly : pv2_ntt) {
+                               res.confirm("cloned vector polynomial (NTT) don't own memory", !poly.owns_storage());
+                            }
+                         }),
+
+      Botan_Tests::CHECK("hamming weight of polynomials",
+                         [](Test::Result& res) {
+                            Kyberish_Poly<Domain::Normal> p;
+                            res.test_is_eq<size_t>("hamming weight of 0", p.hamming_weight(), 0);
```
FAlbertDev, line 169: already tested below

src/tests/test_crystals.cpp
```diff
@@ -0,0 +1,483 @@
...
+                            random_encoding_roundtrips<Dilithiumish_Trait, 2 * eta2>(res, *rng, 3);
+                            random_encoding_roundtrips<Dilithiumish_Trait, 2 * eta4>(res, *rng, 4);
+                            random_encoding_roundtrips<Dilithiumish_Trait, 2 * twotothed - 1>(res, *rng, 13);
+                         }),
+   };
+}
+
+}  // namespace
+
+BOTAN_REGISTER_TEST_FN("pubkey", "CRYSTALS", test_extended_euclidean_algorithm, test_polynomial_basics, test_encoding);
```
FAlbertDev, line 479:
```suggestion
BOTAN_REGISTER_TEST_FN("pubkey", "crystals", test_extended_euclidean_algorithm, test_polynomial_basics, test_encoding);
```
it seems all other test suits use snake_case (see `./botan-test --help`). Maybe, while you're at it, adapt the names for dilithium as well (e.g.`dilithium_kat_4x4_AES_Deterministic`, `dilithium_kat_4x4_Randomized`, etc.).

FAlbertDev (approved):
Final review iteration. 
I really like your work. IMO it's the better reference implementation ;)

src/lib/pubkey/dilithium/dilithium_common/dilithium_constants.h
FAlbertDev: I also like this idea. While the implied switch-cases are annoying, they guarantee we do not forget a value.

src/lib/pubkey/pqcrystals/pqcrystals_encoding.h
```diff
@@ -0,0 +1,224 @@
...
+         // Bit-pack the coefficient into the collectors array and keep track of
+         // the bit-offset within the current collector. Note that this might
+         // shift some high-bits of the coefficient out of the current collector.
+         collectors[c] |= coeff_value << bit_offset;
+         bit_offset += trait::bits_per_coeff;
+
+         // If the bit-offset now exceeds the collector's bit-width, we fill the
+         // next collector with the high-bits that didn't fit into the previous.
+         // The bit-offset is adjusted to now point into the new collector.
+         if(bit_offset > trait::bits_in_collector) {
```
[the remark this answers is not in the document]
FAlbertDev: I see 👍

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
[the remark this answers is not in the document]
FAlbertDev: Wow, I didn't expect such an impact. Nice!

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
```diff
@@ -114,402 +81,272 @@ std::string DilithiumMode::to_string() const {
...
-
-            /* Compute z, reject if it reveals secret */
-            s1.polyvec_pointwise_poly_montgomery(z, cp);
-
-            z.invntt_tomont();
-            z.add_polyvec(y);
-
+         for(uint32_t nonce = 0; nonce <= std::numeric_limits<uint16_t>::max(); nonce += m_mode.l()) {
+            const auto y = dilithium_expand_mask(rhoprime, static_cast<uint16_t>(nonce), m_mode);
+
```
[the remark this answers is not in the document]
reneme: We could, but this is a theoretical thing anyway that happens if a hash implementation turns out to be broken. I think its fine to have an arbitrary cutoff.

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
FAlbertDev: When looking at where `key_length` is used at all, it seems that it should be some metric beside `estimated_strength` to identify specific instances of an algorithm while being generic enough to be a number. Also, the higher the number, the higher the security. Therefore, I would guess everything with these properties should be sufficient, even {1,2,3}. Therefore, I think {512, 768, 1024} should suffice. But I'll submit to the judgment of our Botan master ;)

src/lib/pubkey/pqcrystals/pqcrystals_helpers.h
randombit: I don't think this exists elsewhere currently. Likewise the Montgomery R params, since so far Montgomery over a machine word hasn't been of interest

src/lib/pubkey/pqcrystals/info.txt
```diff
@@ -0,0 +1,18 @@
...
+</defines>
+
+<module_info>
+name -> "CRYSTALS"
+brief -> "Base utilities for CRYSTALS-Kyber/ML-KEM and CRYSTALS-Dilithium/ML-DSA. CRYptographic SuiTe for Algebraic LatticeS"
+type -> "Internal"
+</module_info>
+
+<requires>
+</requires>
```
randombit, line 12: `xof`

src/lib/pubkey/kyber/kyber_common/kyber_helpers.h
```diff
@@ -0,0 +1,75 @@
...
+
+   // This is a mitigation for a potential side channel called "KyberSlash".
+   // It implements the division by Q using a multiplication and a shift. Most
+   // compilers would generate similar code for such a division by a constant.
+   // Though, in some cases, compilers might use a variable-time int division,
+   // resulting in a potential side channel.
+   //
+   //   See "Hacker's Delight" (Second Edition) by Henry S. Warren, Jr.
+   //   Chapter 10-9 "Unsigned Division by Divisors >= 1"
+   auto divide_by_q = [](uint32_t n) -> unsigned_T {
```
randombit, line 45: Why the lambda? Doesn't seem necessary here vs

```
   const uin32_t n = ((static_cast<uint32_t>(x) << d) + KyberConstants::Q / 2;
   BOTAN_DEBUG_ASSERT(n < (1 << 23));

   // These constants work for all values that appear in Kyber with the
   // greatest being 3328 * 2^11 + Q // 2 = 6,817,408 < 2**23 = 8,388,608.
   constexpr uint64_t m = 2580335;
   constexpr size_t p = 33;
   return static_cast<unsigned_T>((n * m) >> p);
```

src/lib/pubkey/kyber/kyber_common/kyber_helpers.h
```diff
@@ -0,0 +1,75 @@
...
+inline uint32_t load_le3(std::span<const uint8_t, 3> in) {
+   return Botan::load_le(std::array<uint8_t, 4>{in[0], in[1], in[2], 0});
+}
+
+/**
+ * NIST FIPS 203 IPD, Formula 4.5 (Compress)
+ */
+template <size_t d>
+   requires(d > 0 && d < 12)
+constexpr std::make_unsigned_t<KyberConstants::T> kyber_compress(KyberConstants::T x) {
```
randombit, line 33: Seems like a decent case for `auto` return type imo

randombit (commented):
Partial review still need to look at changes in `dilithium` and `kyber`

src/lib/xof/xof.h
reneme: See #4152

src/lib/pubkey/kyber/kyber_common/kyber_algos.h
```diff
@@ -0,0 +1,116 @@
...
+                               const KyberPolyVec& u,
+                               const KyberPoly& v,
+                               const KyberConstants& m_mode);
+std::pair<KyberPolyVec, KyberPoly> kyber_decompress_ciphertext(StrongSpan<const KyberCompressedCiphertext> ct,
+                                                               const KyberConstants& mode);
+KyberPolyMat kyber_sample_matrix(StrongSpan<const KyberSeedRho> seed, bool transposed, const KyberConstants& mode);
+void kyber_sample_polynomial_from_cbd(KyberPoly& poly,
+                                      KyberConstants::KyberEta eta,
+                                      const KyberSamplingRandomness& randomness);
+
```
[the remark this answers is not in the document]
reneme: aye aye.

src/tests/test_crystals.cpp
reneme: One is testing poly, the other polyvec. The more the merrier I would say.

src/tests/test_crystals.cpp
reneme: Fair. Regarding the dilithium KATs: These names are currently tightly coupled with  the `DilithiumMode` constants. 😨 This will need a cleanup anyway, but lets not do that in this PR.

src/lib/pubkey/pqcrystals/pqcrystals_helpers.h
```diff
@@ -0,0 +1,143 @@
...
+   std::conditional_t<sizeof(T) == 4, int64_t, void>>>;
+
+// clang-format on
+
+template <std::integral T>
+   requires(size_t(sizeof(T)) <= 4)
+constexpr T montgomery_R(T q) {
+   using T_unsigned = std::make_unsigned_t<T>;
+   using T2 = next_longer_uint_t<T_unsigned>;
+   return (T2(1) << (sizeof(T) * 8)) % q;
```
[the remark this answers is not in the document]
reneme: `consteval` it is.

src/lib/pubkey/pqcrystals/pqcrystals_encoding.h
```diff
@@ -0,0 +1,224 @@
...
+      using sink_t = uint64_t;
+
+      static_assert(range <= std::numeric_limits<T>::max());
+
+      constexpr static size_t bits_in_collector = sizeof(sink_t) * 8;
+      constexpr static size_t bits_per_coeff = bitlen(range);
+      constexpr static size_t bits_per_pack = [] {
+         // Ensure that the bit-packing is byte-aligned and scale it
+         // to utilize the collector's bit-width as much as possible.
+         size_t smallest_aligned_pack = lcm(bits_per_coeff, size_t(8));
```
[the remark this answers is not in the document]
reneme: Sad that there's no extended euclid.

src/tests/test_crystals.cpp
```diff
@@ -0,0 +1,483 @@
...
+
+template <Domain D>
+using Kyberish_Poly = Botan::CRYSTALS::Polynomial<Kyberish_Trait, D>;
+
+template <Domain D>
+using Kyberish_PolyVec = Botan::CRYSTALS::PolynomialVector<Kyberish_Trait, D>;
+
+std::vector<Test::Result> test_polynomial_basics() {
+   return {
+      Botan_Tests::CHECK("polynomial owning storage",
```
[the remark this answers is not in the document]
reneme: noice!

src/lib/pubkey/pqcrystals/info.txt
reneme: Mhh, actually this requires `xof` only when used in conjunction with Diltihium. Kyber doesn't unpack from a XOF output.

src/lib/pubkey/kyber/kyber_common/kyber_helpers.h
reneme: Mhh, yes and no. I feel there's some value in having the explicit return type for methods that are used elsewhere (and where the type isn't overly excessive).

src/lib/pubkey/kyber/kyber_common/kyber_helpers.h
reneme: I did that for readability, as it shows the intent of "dividing by Q" more clearly. Changed it though, as the comment also makes this clear in detail.

src/lib/pubkey/kyber/kyber_common/kyber.cpp
reneme: Nice! Thanks double-checking and for the detailed write-up! It would probably be really nice to track the montgomery parameter in the type system as well, but it does feel like overkill.

Also: be applying `montgomery()` to the `t` vector, we can remove the overload for `PolynomialMatrix` altogether. :)

src/lib/pubkey/pqcrystals/info.txt
randombit: I see you added ifdef guard, that's fine too, just has to be able to compile one way or another

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
randombit: @FAlbertDev has the right idea about the intent. It doesn't really have anything to do with the length of the key in bytes in any direct way (after all a ECDSA P-256 public key is 64 bytes being two field elements, but saying the key length of ECDSA P-256 is 512 bits is not going to illuminate the situation!).

For Kyber I see quite commonly ML-KEM-{512,768,1024} so that seems the obvious choice. NIST uses the same naming.

For Dilithium I have seen less press. The NIST spec (as well as https://www.ietf.org/id/draft-ietf-lamps-dilithium-certificates-03.html) uses 44, 65, and 87.  TBH I have no idea where those numbers come from, but it's as good as anything.

src/lib/pubkey/kyber/kyber_common/kyber.cpp
```diff
@@ -188,7 +187,7 @@ std::vector<uint8_t> Kyber_PublicKey::public_key_bits() const {
 
 size_t Kyber_PublicKey::key_length() const {
    // TODO: this should report 512, 768, 1024
-   return m_public->mode().public_key_byte_length();
+   return m_public->mode().public_key_bytes();
 }
```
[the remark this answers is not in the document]
randombit: IMO these should be 512, 768, 1024

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
reneme: 4x4, 6x5 and 8x7 are the dimensions of the generated public matrix A for the Dilithium instances.

randombit:
Everything so far looks good but I'm not enthused about dropping big refactors just before a release so pushing this to 3.6.0

src/lib/pubkey/pqcrystals/pqcrystals_encoding.h
```diff
@@ -0,0 +1,228 @@
...
+ * Note that this bit-packing algorithm is inefficient if the bit-length of the
+ * coefficients is a multiple of 8. In that case, a byte-level encoding (that
+ * might need to take endianess into account) would be more efficient. However,
+ * neither Kyber nor Dilithium instantiate bit-packings with such a value range.
+ *
+ * @tparam range the upper bound of the coefficient range.
+ */
+template <int32_t range, crystals_trait PolyTrait, Domain D, coeff_map_fn<typename PolyTrait::T> MapFnT>
+constexpr void pack(const Polynomial<PolyTrait, D>& p, BufferStuffer& stuffer, MapFnT map) {
+   using trait = BitPackingTrait<range, PolyTrait>;
```
reneme, line 117: Could the gist of this be useful beyond the CRYSTALS scope? E.g. FrodoKEM has a pack/unpack routine that -- on first glance at least -- looks like it could use it as well. @randombit perhaps there are more examples?

https://github.com/randombit/botan/blob/3657a72dfe697b4e56f0f0192bbf62cc164be2d6/src/lib/pubkey/frodokem/frodokem_common/frodo_matrix.cpp#L340-L384

src/lib/pubkey/dilithium/dilithium_common/dilithium_algos.h
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+#ifndef BOTAN_DILITHIUM_ALGOS_H_
+#define BOTAN_DILITHIUM_ALGOS_H_
+
+#include <botan/internal/dilithium_types.h>
+
+namespace Botan {
+
+DilithiumPolyMatNTT dilithium_expand_A(StrongSpan<const DilithiumSeedRho> rho, const DilithiumConstants& mode);
```
randombit, line 19: Nit/bikeshed: I think these should all be in a sub-namespace eg `Dilithium_F` and then without the `dilithium_` prefix

src/lib/pubkey/kyber/kyber_common/kyber_algos.h
```diff
@@ -0,0 +1,123 @@
...
+#include <botan/xof.h>
+#include <botan/internal/fmt.h>
+#include <botan/internal/kyber_symmetric_primitives.h>
+#include <botan/internal/kyber_types.h>
+#include <botan/internal/loadstor.h>
+
+namespace Botan {
+
+void kyber_encode_polynomial_vector(std::span<uint8_t> out, const KyberPolyVecNTT& p);
```
randombit, line 26: Likewise with Dilithium, bikeshedding for `namespace Kyber_F` or the like

randombit (approved):
This looks very good. I left one bikeshed suggestion that to me feels a little cleaner - but it's marginal, feel free to ignore if you disagree.

randombit:
Final comment, could use some history cleanup before merge

src/lib/pubkey/dilithium/dilithium_common/dilithium_algos.h
reneme: Done.

src/lib/pubkey/kyber/kyber_common/kyber_algos.h
reneme: Done.

reneme:
> Final comment, could use some history cleanup before merge

Rebased to master, cleaned up history and built the bikeshed. Waiting for CI, then finally drawing a line here. 🥳

PR #4032, opened 2024-04-25
Fix a bug in SAN handling when creating certs and PKCS 10 requests

randombit:
If the Subject Alternative Name extension was set in opts.extensions, then we would ignore any values set in other fields (eg opts.dns)

### DIFF
--- a/src/lib/x509/x509self.cpp
+++ b/src/lib/x509/x509self.cpp
@@ -21,25 +21,48 @@ namespace {
 /*
 * Load information from the X509_Cert_Options
 */
-void load_info(const X509_Cert_Options& opts, X509_DN& subject_dn, AlternativeName& subject_alt) {
+X509_DN load_dn_info(const X509_Cert_Options& opts) {
+   X509_DN subject_dn;
+
    subject_dn.add_attribute("X520.CommonName", opts.common_name);
    subject_dn.add_attribute("X520.Country", opts.country);
    subject_dn.add_attribute("X520.State", opts.state);
    subject_dn.add_attribute("X520.Locality", opts.locality);
    subject_dn.add_attribute("X520.Organization", opts.organization);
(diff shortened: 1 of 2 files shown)


src/lib/x509/x509self.cpp
```diff
@@ -21,25 +21,52 @@ namespace {
...
+   if(auto ext = extensions.get_extension_object_as<Cert_Extension::Subject_Alternative_Name>()) {
+      AlternativeName subject_alt = ext->get_alt_name();
+      load_alt_name(opts, subject_alt);
+      extensions.replace(std::make_unique<Cert_Extension::Subject_Alternative_Name>(subject_alt));
+   } else {
+      AlternativeName subject_alt;
+      load_alt_name(opts, subject_alt);
+      extensions.replace(std::make_unique<Cert_Extension::Subject_Alternative_Name>(subject_alt));
+   }
+}
```
FAlbertDev, line 68: Maybe remove the partly redundant if-statement by something like.
```cpp
AlternativeName subject_alt = [&]{
   /*
      If the extension was already created in opts.extension we need to
      merge that with any values that were set in the other fields
   */
   if(auto ext = 
   extensions.get_extension_object_as<Cert_Extension::Subject_Alternative_Name>()) {
      return ext->get_alt_name();
   } else {
      return AlternativeName(); // new
   }
}();
load_alt_name(opts, subject_alt);
extensions.replace(std::make_unique<Cert_Extension::Subject_Alternative_Name>(subject_alt));

```

FAlbertDev (approved):
👍

src/lib/x509/x509self.cpp
randombit: I realized we can just create an empty one then assign to it if the extension is set, much cleaner

PR #4034, opened 2024-04-26
Refactor AlternativeName internals

randombit:
This provides a completely new and rather more sensible interface, with the old (now deprecated) interface being implemented in terms of the new interface.

It also resolves a performance issue when parsing certificates with many (thousands) of alternative names; previously this triggered some accidentally quadratic code when checking for duplicates.

Performance issue was reported by Bing Shi.

### DIFF
--- a/src/lib/x509/alt_name.cpp
+++ b/src/lib/x509/alt_name.cpp
@@ -1,0 +1,171 @@
+/*
+* (C) 2024 Jack Lloyd
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/pkix_types.h>
+
+#include <botan/ber_dec.h>
+#include <botan/der_enc.h>
+#include <botan/internal/loadstor.h>
+#include <botan/internal/parsing.h>
+
+namespace Botan {
+
+void AlternativeName::add_uri(std::string_view uri) {
+   if(!uri.empty()) {
+      m_uri.insert(std::string(uri));
+   }
+}
+
+void AlternativeName::add_email(std::string_view addr) {
+   if(!addr.empty()) {
+      m_email.insert(std::string(addr));
+   }
+}
+
+void AlternativeName::add_dns(std::string_view dns) {
+   if(!dns.empty()) {
+      m_dns.insert(std::string(dns));
+   }
+}
+
+void AlternativeName::add_other_name(const OID& oid, const ASN1_String& value) {
+   m_othernames.insert(std::make_pair(oid, value));
+}
+
+void AlternativeName::add_dn(const X509_DN& dn) {
+   m_dn_names.insert(dn);
+}
+
+void AlternativeName::add_ip_address(std::string_view ip) {
+   if(!ip.empty()) {
+      m_ip_addr.insert(std::string(ip));
+   }
+}
+
+bool AlternativeName::has_items() const {
+   if(!this->dns().empty()) {
+      return true;
+   }
+   if(!this->uris().empty()) {
+      return true;
+   }
+   if(!this->email().empty()) {
+      return true;
+   }
+   if(!this->ip_address().empty()) {
+      return true;
+   }
+   if(!this->directory_names().empty()) {
+      return true;
+   }
+   if(!this->other_names().empty()) {
+      return true;
+   }
+   return false;
+}
+
+void AlternativeName::encode_into(DER_Encoder& der) const {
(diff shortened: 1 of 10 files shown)


src/lib/x509/alt_name.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+   for(const auto& name : m_dns) {
+      ASN1_String str(name, ASN1_Type::Ia5String);
+      der.add_object(ASN1_Type(2), ASN1_Class::ContextSpecific, str.value());
+   }
+
+   for(const auto& name : m_dn_names) {
+      // THIS IS WRONG!!
+      der.encode(name);
+   }
```
FAlbertDev, line 108: `THIS IS WRONG`? Maybe elaborate?

src/lib/x509/asn1_alt_name.cpp
```diff
@@ -65,170 +76,107 @@ void AlternativeName::add_othername(const OID& oid, std::string_view value, ASN1
...
+      names.emplace("IP", nm);
+   }
 
-std::string AlternativeName::get_first_attribute(std::string_view attr) const {
-   auto i = m_alt_info.lower_bound(attr);
-   if(i != m_alt_info.end() && i->first == attr) {
-      return i->second;
+   for(const auto& nm : this->ip_address()) {
+      names.emplace("IP", nm);
    }
```
FAlbertDev, line 97: The code here is a duplicate.

src/lib/x509/asn1_alt_name.cpp
```diff
@@ -65,170 +76,107 @@ void AlternativeName::add_othername(const OID& oid, std::string_view value, ASN1
...
-   encode_entries(der, m_alt_info, "DNS", ASN1_Type(2));
-   encode_entries(der, m_alt_info, "DN", ASN1_Type(4));
-   encode_entries(der, m_alt_info, "URI", ASN1_Type(6));
-   encode_entries(der, m_alt_info, "IP", ASN1_Type(7));
+std::vector<std::string> AlternativeName::get_attribute(std::string_view attr) const {
+   auto set_to_vector = [](const std::set<std::string>& s) {
+      std::vector<std::string> v(s.size());
+      std::copy(s.begin(), s.end(), v.begin());
+      return v;
+   };
```
FAlbertDev, line 139: Alternatively:
```cpp
   auto set_to_vector = [](const std::set<std::string>& s) -> std::vector<int> {
      return {s.begin(), s.end()};
   };
```

src/lib/x509/x509cert.cpp
```diff
@@ -510,38 +510,44 @@ const AlternativeName& X509_Certificate::issuer_alt_name() const {
...
+namespace {
 
-   if(subject_dn().has_field(req)) {
-      return subject_dn().get_attribute(req);
+std::vector<std::string> get_cert_user_info(std::string_view req, const X509_DN& dn, const AlternativeName& alt_name) {
+   auto set_to_vector = [](const std::set<std::string>& s) {
+      std::vector<std::string> v(s.size());
+      std::copy(s.begin(), s.end(), v.begin());
+      return v;
+   };
```
FAlbertDev, line 520: Same as above

src/lib/x509/alt_name.cpp
randombit: Oops I meant to fix that! What's here matches the current behavior of the code on master, but I believe it is incorrect since we're omitting the tagging. I still need to confirm using something like OpenSSL. If this is the case we've never been able to successfully encode an alternative name with a DN :/

src/lib/x509/asn1_alt_name.cpp
randombit: Thanks I would not have expected that to work

src/lib/x509/asn1_alt_name.cpp
randombit: Fixed

src/lib/x509/x509cert.cpp
randombit: Fixed

src/lib/x509/alt_name.cpp
randombit: Checked with OpenSSL, indeed we encoded DNs in SAN incorrectly :sob:

Decoding always worked.

Fixed now, and with a test that we can read what we write.

PR #4037, opened 2024-05-01
fixed retransmission handled ClientHello with another protocol version.

volok-aleksej:
according to https://www.rfc-editor.org/rfc/rfc9147.html#section-4-6.2

> This value MUST be set to {254, 253} for all records other than the initial ClientHello (i.e., one not generated after a HelloRetryRequest), where it may also be {254, 255} for compatibility purposes. It MUST be ignored for all purposes. See [[TLS13](https://www.rfc-editor.org/rfc/rfc9147.html#RFC8446)], [Appendix D.1](https://www.rfc-editor.org/rfc/rfc8446#appendix-D.1) for the rationale for this.

for all ClientHello we have to ignore checking protocol version on record layer, real version of protocol we have checking on handshake layer as well as the checking retransmitted messages.

### DIFF
--- a/src/lib/tls/tls12/tls_channel_impl_12.cpp
+++ b/src/lib/tls/tls12/tls_channel_impl_12.cpp
@@ -316,6 +316,11 @@ size_t Channel_Impl_12::from_peer(std::span<const uint8_t> data) {
          BOTAN_ASSERT_IMPLICATION(epoch0_restart, allow_epoch0_restart, "Allowed state");
 
          const bool initial_record = epoch0_restart || (!pending_state() && !active_state());
+         bool initial_handshake_message = false;
+         if(record.type() == Record_Type::Handshake && !m_record_buf.empty()) {
+            Handshake_Type type = static_cast<Handshake_Type>(m_record_buf[0]);
+            initial_handshake_message = (type == Handshake_Type::ClientHello);
+         }
 
          if(record.type() != Record_Type::Alert) {
             if(initial_record) {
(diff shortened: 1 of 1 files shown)


src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -316,6 +316,11 @@ size_t Channel_Impl_12::from_peer(std::span<const uint8_t> data) {
          BOTAN_ASSERT_IMPLICATION(epoch0_restart, allow_epoch0_restart, "Allowed state");
 
          const bool initial_record = epoch0_restart || (!pending_state() && !active_state());
+         bool initial_handshake_message = false;
+         if(record.type() == Record_Type::Handshake) {
+            Handshake_Type type = static_cast<Handshake_Type>(m_record_buf[0]);
```
randombit, line 321: I think you need to check that `m_record_buf` is not empty before doing this. At least I'm guessing that is the cause of the CiFuzz failure.

volok-aleksej: yes, fixed

volok-aleksej: sorry, i am first time doing this. now it's done

volok-aleksej:
Hi, Guys, what is progress? I have updated own master branch, therefor need aprove ci again. Sorry, I'll be waiting for you.

randombit:
@volok-aleksej I need to refresh my memory of the RFC here, I'll review tonight

randombit (approved):
LGTM thank you!

randombit:
@reneme Tagging you just as an FYI

PR #4040, opened 2024-05-04
Correct an error in RFC 6979 ECDSA

randombit:
The RFC 6979 mode of ECDSA invoked using the shifted/truncated hash, without reducing modulo the group order. See Section 2.3.4 of RFC 6979 for the definition.

Unfortunately this situation is missed by the official RFC 6979 tests, and for other curves we simply generated the test vectors ourselves, which replicated the error.

DSA always reduced correctly.

This doesn't affect interop or security. However the current tests trip up an alternate (future) implementation of ECDSA where all of the variables used in the computation are held in integers which are strictly bounded modulo the group order, rather than generic BigInts.

The second commit fixes an only slightly related issue, namely that we accepted EC private keys larger than the order. They would effectively be reduced modulo the order during computations. There is no reason to support this. Unfortunately some tests actually did this, particularly the Brainpool RFC 6979 tests.


randombit:
Going to merge this now since test failures are blocking other work and I'm trying to avoid a too high tower of PR dependencies.

FYI @reneme tagging this in case it has relevance to your BSI documentation

src/lib/pubkey/ecdsa/ecdsa.cpp
```diff
@@ -158,7 +158,7 @@ AlgorithmIdentifier ECDSA_Signature_Operation::algorithm_identifier() const {
 secure_vector<uint8_t> ECDSA_Signature_Operation::raw_sign(const uint8_t msg[],
                                                            size_t msg_len,
                                                            RandomNumberGenerator& rng) {
-   BigInt m = BigInt::from_bytes_with_max_bits(msg, msg_len, m_group.get_order_bits());
+   BigInt m = m_group.mod_order(BigInt::from_bytes_with_max_bits(msg, msg_len, m_group.get_order_bits()));
```
FAlbertDev, line 161: I looks like the `m_group.mod_order` below in line 179 can be removed with this change. 
```cpp
// Line 179
m = m_group.multiply_mod_order(m_b, m_group.mod_order(m));
```

FAlbertDev (commented):
Nice, LGTM! Sorry for the late review.

PR #4045, opened 2024-05-08
During X509 path validation, return immediately if a signature is invalid

randombit:
The remainder of path validation logic is still subject to attacker controlled inputs, but the range of inputs is reduced to that which a legitimate certificate authority was willing to sign.


FAlbertDev (approved):
LGTM. I agree that checking the signature chain first is very sensible.

Issue #3877, opened 2024-01-08
Future TPM support


securitykernel:
The [ToDo list](https://github.com/randombit/botan/blob/eb8384c1af5ec48d311542f8e8252546d145e8f4/doc/dev_ref/todo.rst) currently lists this for TPM support:

* Better TPM support: NVRAM, PCR measurements, sealing
* Add support for TPM 2.0 hardware

Is the first one still desired? Given that TPM 1.2 standard was first released in 2003 and superseded by TPM 2.0 in 2014, I'd suggest to deprecate TPM 1.2 support for removal in 4.0 and add basic TPM 2.0 support as a replacement (if still desired).

reneme:
## Current Status

TPM 1.2 is deprecated as of https://github.com/randombit/botan/commit/90001378a0f2484731b99526f0b7ed89b6b33f44. @atreiber94 and me will work on (limited) TPM 2.0 support.

## Work Items

* [ ] Set up a test environment for CI (based on [the TPM simulator](https://github.com/tpm2-software/tpm2-software-container/blob/master/ubuntu-20.04.docker.m4))
* [ ] TPM-based **random number generator**
* [ ] Support **asymmetric crypto** (aka. TPM keypair)
* [ ] **FFI** support
* [ ] **Python** bindings

reneme:
The pull requests merged so far (slated for release with Botan 3.6.0) provide limited support to create and use TPM-hosted keys (both RSA and ECDSA), as well as an interface to the TPM's random number generator. Also, a feature-complete "crypto backend" (requires TSS 4.x) was added. With that, the underlying TSS can utilize Botan's crypto algorithms for securely communicating with the TPM.

Below there's a list of current limitations (adapted from https://github.com/randombit/botan/pull/4337) that may stand as an (incomplete) list of TODOs. Currently we don't have immediate plans to tackle any of those, but that may change. Therefore, this ticket is not assigned to any release milestone for the time being.

* Configuration of key pair properties on creation (perhaps based on [#4318](https://github.com/randombit/botan/pull/4318))
  * Restrict key usage (sign vs. X.509 certification vs. verify)
  * Restrict signature schemes (e.g. signing only with PSS)
  * Duplicatability ("fixed to TPM", "fixed to parent key")
  * Usage restrictions
  * Hierarchies (currently only "Owner")
  * Primary and/or Derived Keys (currently only "Ordinary" Keys)
  * ...
* Sessions
  * Audit sessions
  * Policy sessions
[...]

PR #4050, opened 2024-05-11
Improve notification and documentation of deprecated features

randombit:
In #3910 we introduced a nice way of marking specific modules as deprecated. However this change did not set the lifecycles of any specific modules.

Additionally, officially deprecate the TPM support which is pretty obsolete and these days not well tested. (GH #3877)


reneme (approved):
Perhaps Dilithium AES may also go on this list. Same argument as with Kyber 90s.

PR #4058, opened 2024-05-14
Add a much faster IPv4 dotted quad parser

randombit:


src/lib/utils/parsing.cpp
```diff
@@ -153,36 +153,70 @@ std::string string_join(const std::vector<std::string>& strs, char delim) {
...
+         accum = 0;
+      } else if(c >= '0' && c <= '9') {
+         const auto d = static_cast<uint8_t>(c - '0');
 
-   for(auto part = parts.begin(); part != parts.end(); ++part) {
-      uint32_t octet = to_u32bit(*part);
+         // prohibit leading zero in quad
+         if(d == 0 && accum == 0) {
+            return {};
+         }
```
FAlbertDev, line 185: Currently, a zero quad is not possible. E.g. something like "127.0.0.1" would fail.

FAlbertDev (approved):

src/lib/utils/parsing.cpp
randombit: Nice catch. Added a test suite and extended the fuzzer. Also did an offline exhaustive test for all possible 32 bit addresses that string_to_ipv4 and ipv4_to_string round trip correctly.

PR #4063, opened 2024-05-18
Fix several bugs relating to OID encoding and decoding

### LINKED ISSUE #4023: Invalid BER decoding of OIDs
Hi, we've found an issue with Botan's BER decoding of OIDs.

The byte sequence `0x06 0x02 0x79 0x01` should yield the OID `2.41.1`. However, Botan decodes it as `3.1.1`. _Please note_ that `3.1.1` is explicitly forbidden by the standard (see [ITU-T X.660 Annex A.2](https://www.itu.int/rec/T-REC-X.660/en)). 

According to the standard, the first "byte" is encoded by multiplying the first _object identifier component_ by 40 and adding the second _object identifier component_: `2 * 40 + 41 = 121 (0x79)` (see [ITU-T X.690 Section 8.19.4](https://www.itu.int/ITU-T/studygroups/com17/languages/X.690-0207.pdf)). When decoding this byte, Botan interprets this as `3 * 40 + 1`. For arc `0` and `1` only subidentifiers up to `39` are allowed (they couldn't be distinguished otherwise) (see [ITU-T X.660 Section 7.6](https://www.itu.int/rec/T-REC-X.660/en)). For arc `2`, there's no such limitation.

Furthermore, Botan does treat the first byte as a single byte. That is another issuer: The valid OID [`2.999.1234`](http://oid-info.com/get/2.999) cannot be encoded/decoded by Botan. For this to work, Botan must use the same _variable length encoding_ (bit 8 set to 1) for the first "byte" (see [ITU-T X.690 Section 8.19.2](https://www.itu.int/ITU-T/studygroups/com17/languages/X.690-0207.pdf)).

Helpful Links:
- https://misc.daniel-marschall.de/asn.1/oid-converter/online.php
- https://misc.daniel-marschall.de/asn.1/oid_facts.html

randombit:
The handling for OID encoding did not correctly handle OIDs that begin with 2.x where x >= 40.

Fixes #4023


jdoubleu (commented):
Looks good overall 👍.

I'm not sure if it helps to clarify, that Botan only really supports OIDs in the context of X.503v3 certificates. In theory OIDs can have _infinitely_ large components, while RFC5280 (Appendix B) restricts them to only 28 bits.

I think _negative_ tests would be beneficial as well: Using invalid encodings and seeing what happens (hopefully an exception). I found some test vectors on this page: https://misc.daniel-marschall.de/asn.1/oid_facts.html#chap3

src/lib/asn1/asn1_oid.cpp
```diff
@@ -11,15 +11,27 @@
...
 
 namespace {
 
+void oid_valid_check(std::span<const uint32_t> oid) {
+   BOTAN_ARG_CHECK(oid.size() >= 2, "OID too short to be valid");
+   BOTAN_ARG_CHECK(oid[0] <= 2, "OID root out of range");
+   BOTAN_ARG_CHECK(oid[1] <= 39 || oid[0] == 2, "OID second arc too large");
+   // This last is a limitation of using 32 bit integers when decoding
+   // not a limitation of ASN.1 object identifiers in general
+   BOTAN_ARG_CHECK(oid[1] <= 0xFFFFFFAF, "OID second arc too large");
```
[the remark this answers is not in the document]
randombit: It's `(2**32-1) - 80` which ensures that we can hold the combined first + second in a single `uint32_t`

randombit:
> I'm not sure if it helps to clarify, that Botan only really supports OIDs in the context of X.503v3 certificates. In theory OIDs can have infinitely large components, while RFC5280 (Appendix B) restricts them to only 28 bits.

Indeed. In the short term we can't fix this because we have some APIs (which become deprecated in this PR) which return the OID components as 32 bit integers. So it's not possible for us to support anything larger that this in the meantime.

> I think negative tests would be beneficial as well:

Good point thanks, I'll add this.

src/lib/asn1/asn1_oid.cpp
```diff
@@ -194,35 +210,73 @@ void OID::decode_from(BER_Decoder& decoder) {
...
-            break;
+            if(!more) {
+               return std::make_pair(data, b);
+            }
          }
+         throw Decoding_Error("Truncated OID value");
+      }
+   };
+
+   std::span<const uint8_t> span(obj.bits(), obj.length());
```
reneme, line 252: Side note: I think it would be good to have `BER_Object::data() -> std::span<const uint8_t>` here.

src/lib/asn1/asn1_oid.cpp
```diff
@@ -194,35 +210,73 @@ void OID::decode_from(BER_Decoder& decoder) {
...
-
-   if(length < 2 && !(length == 1 && bits[0] == 0)) {
+   if(obj.length() == 0) {
       throw BER_Decoding_Error("OID encoding is too short");
    }
 
-   m_id.clear();
-   m_id.push_back(bits[0] / 40);
-   m_id.push_back(bits[0] % 40);
+   auto consume = [](std::span<const uint8_t> data) -> std::pair<std::span<const uint8_t>, uint32_t> {
```
reneme, line 217: IMHO this is a good use case for a `BufferSlicer&` and letting `consume` call `BufferSlicer::take_byte()` as often as needed.  That would remove most of the buffer bookkeeping from the implementation in `consume` and also simplify the return type to just a `uint32_t`.

src/lib/asn1/asn1_oid.cpp
```diff
@@ -194,35 +210,73 @@ void OID::decode_from(BER_Decoder& decoder) {
...
-
-         if(!(bits[i] & 0x80)) {
-            break;
+            if(!more) {
+               return std::make_pair(data, b);
+            }
          }
+         throw Decoding_Error("Truncated OID value");
+      }
+   };
```
reneme, line 250: Let me take the liberty to propose an alternative implementation for `consume` that uses `BufferSlicer&` and has a (IMHO) simpler control flow:

```C++
   auto consume = [](BufferSlicer& data) -> uint32_t {
      BOTAN_ASSERT_NOMSG(!data.empty());
      uint32_t b = data.take_byte();

      if(b > 0x7F) {
         b &= 0x7F;

         // Even BER requires that the OID have minimal length, ie that
         // the first byte of a multibyte encoding cannot be zero
         // See X.690 section 8.19.2
         if(b == 0) {
            throw Decoding_Error("Leading zero byte in multibyte OID encoding");
         }

         while(true) {
            if(data.empty()) {
               throw Decoding_Error("Truncated OID value");
            }

            if((b >> (32 - 7)) > 0) {
               throw Decoding_Error("OID component overflow");
            }

            const auto next = data.take_byte();
            b <<= 7;
            b |= (next & 0x7F);

            if(const bool more = (next & 0x80); !more) {
               break;
            }
         }
      }

      return b;
   };
```

src/lib/asn1/asn1_oid.cpp
```diff
@@ -194,35 +210,73 @@ void OID::decode_from(BER_Decoder& decoder) {
...
+
+   std::span<const uint8_t> span(obj.bits(), obj.length());
+   std::vector<uint32_t> parts;
+   while(!span.empty()) {
+      auto [subspan, comp] = consume(span);
+      span = subspan;
```
reneme, line 256: Here's the invocation of `consume` that uses `BufferSlicer&` (as suggested above):

```suggestion
   BufferSlicer data({obj.bits(), obj.length()});
   std::vector<uint32_t> parts;
   while(!data.empty()) {
      const auto comp = consume(data);
```

reneme (approved):
Thanks for taking this on! Looks good to me, too. I added a few nits and a potential control flow simplification for the decoder.

src/lib/asn1/asn1_oid.cpp
```diff
@@ -159,28 +170,33 @@ void OID::encode_into(DER_Encoder& der) const {
       throw Invalid_Argument("OID::encode_into: OID is invalid");
```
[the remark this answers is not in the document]
randombit: Unfortunately with how the ASN.1 library works right now any type you decode has to be default constructable. :/

src/lib/asn1/asn1_oid.cpp
```diff
@@ -81,24 +93,29 @@ OID OID::from_string(std::string_view str) {
...
+   } catch(...) {}
 
    throw Lookup_Error(fmt("No OID associated with name '{}'", str));
 }
 
+OID::OID(std::initializer_list<uint32_t> init) : m_id(init) {
+   oid_valid_check(m_id);
+}
+
+OID::OID(std::vector<uint32_t>&& init) : m_id(init) {
```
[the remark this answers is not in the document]
randombit: Yeah tbh I'm not sure what the rvalue is doing here. I'll deprecate this.

It's less critical to deprecate the constructors, wrt supporting longer OID components - 99.99% of the time every OID component fits in to a 32 bit integer just fine, and constructing an OID on that basis is sound. The problem with the accessors is that they assume anything we just decoded must also fit into 32 bit integers, which causes the problem.

In the future I expect we'll hold OID components using something like `std::variant<uint32_t, std::vector<uint8_t>>` - if it fits into 32 bits use that, otherwise bump in to a variable length integer stored on the heap. But of course the current accessors are not compatible with this.

src/lib/asn1/asn1_oid.cpp
randombit: I seem to recall adding exactly this recently, and went looking for which open PR of mine had it but it seems to have gotten lost ¯\_(ツ)_/¯

I'll add it here.

src/lib/asn1/asn1_oid.cpp
randombit: :+1:

src/lib/asn1/asn1_oid.cpp
randombit: I tweaked the control flow a bit but yeah `BufferSlicer` simplifies things here

src/lib/asn1/asn1_oid.cpp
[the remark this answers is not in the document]
randombit: (I'm working on a new ASN.1 library but this is a very long term project, Botan 4 at earliest)

reneme (approved):

jdoubleu:
Thanks for fixing this! 👏

PR #4066, opened 2024-05-21
Tweak XCode version detection, update CI to XCode 15.4

randombit:


src/build-data/detect_version.cpp
```diff
@@ -30,11 +30,19 @@ EMCC __EMSCRIPTEN_major__ __EMSCRIPTEN_minor__
 #elif defined(__clang__) && defined(__apple_build_version__)
 
 /*
-Bizarrely, XCode uses the patch level to indicate the minor number,
-and __clang_minor__ is always set to 0.
+* Even __apple_build_version__ doesn't always get updated to reflect
+* the XCode version, eg XCode 15.4 and XCode 15.3 both define
+* __apple_build_version__ to 15000309
```
reneme, line 35: Judging from the release notes of [15.3](https://developer.apple.com/documentation/Xcode-Release-Notes/xcode-15_3-release-notes) and [15.4](https://developer.apple.com/documentation/Xcode-Release-Notes/xcode-15_4-release-notes), they didn't seem to have updated clang from .3 to .4 (no clang related entries in the notes). Perhaps that's why the pre-processor define didn't get updated either.

On the plus side: That probably means, we wouldn't really benefit from being able to tell .3 and .4 apart anyway.

That said: It's still ridiculous. 🤣

reneme (approved):

PR #4070, opened 2024-05-24
FIX: some code smells found by clang-tidy

reneme:
Introduced in the example code of #4043.

### DIFF
--- a/src/examples/tls_ssl_key_log_file.cpp
+++ b/src/examples/tls_ssl_key_log_file.cpp
@@ -24,8 +24,8 @@
    #include <sys/socket.h>
 #endif
 
-#define SERVER_PORT 5060
-#define CLIENT_PORT 5070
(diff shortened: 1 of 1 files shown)


randombit (approved):
Thanks

Are we not running `clang-tidy` on the examples in CI?

reneme:
> Are we not running clang-tidy on the examples in CI?

Fixed here: https://github.com/randombit/botan/pull/4075/commits/17dd83ccaf14a4ecfbb7e290d0d09bf5a322181f
Problem was, that the CI's `clang-tidy` job did not enable 'examples' in the build configuration (among other things). The `run_clang_tidy.py` is consulting `compile_commands.json` for files to be evaluated: which only lists files in components that were actually enabled in the build configuration. Therefore, examples were not clang-tidy'ed in CI before.

PR #4072, opened 2024-05-24
Fix Python wrapper when X509 cert or CRL parsing fails

randombit:
In Python __del__ is called even if __init__ fails with an exception. In the X509 objects, if parsing failed, then __obj would remain unset. This would then cause another exception in __del__ due to the failed member access. This would be ignored by Python, but it causes a warning to be printed.

By coincidence of how the FFI API works, none of the other wrappers were affected by this.


reneme (approved):
Would it make sense to guard the `botan_x509_crl_destroy()` for nullptr as well, or instead?

PR #4078, opened 2024-05-27
Chore: Update BoGo test suite

reneme:
BoGo now puts some extra emphasis on the certificate chain. Most notably, it provides an ad-hoc self-signed certificate or (sometimes) a certificate chain that the shim should validate against. That's implemented now.

Now, that `Credentials_Manager::trusted_certificate_authorities()` is actually implemented in the shim, `TLS13-Empty-Client-CA-List` started failing... as it is not empty anymore. Incidentally this uncovered a TODO I overlooked earlier: Namely to call `Callbacks::tls_modify_extensions()` on a "CertificateRequest" in TLS 1.3. This callback is now used in the shim to actually empty the CA list in said message: We don't have another way to do that. Botan's TLS server implementation always reports the `::trusted_certificate_authorities()` automagically.

Finally, `DTLS-Replay-NonMonotonic` started failing. Most certainly due to this change: https://github.com/google/boringssl/commit/f94f3ed3965ea033001fb9ae006084eee408b861 (Increase DTLS window size from 64 to 256). My DTLS-knowledge isn't enough to debug this ad-hoc. **I feel we should look into it, though.**


src/bogo_shim/config.json
```diff
@@ -183,6 +192,7 @@
 
         "DTLS-StrayRetransmitFinished-ClientFull": "Needs investigation",
         "DTLS-StrayRetransmitFinished-ServerResume": "Needs investigation",
+        "DTLS-Replay-NonMonotonic": "Needs investigation, started failing after https://github.com/google/boringssl/commit/f94f3ed3965ea033001fb9ae006084eee408b861",
```
reneme, line 195: This should be investigated!

src/lib/tls/tls13/msg_certificate_req_13.cpp
```diff
@@ -96,9 +98,7 @@ Certificate_Request_13::Certificate_Request_13(std::vector<X509_DN> acceptable_C
 
    // TODO: Support cert_status_request for OCSP stapling
 
-   // TODO: give the application a chance to modifying extensions
-   //       (after GH #2988 is merged)
-   // callbacks.tls_modify_extensions(m_extensions, Connection_Side::Server);
+   callbacks.tls_modify_extensions(m_extensions, Connection_Side::Server, type());
```
reneme, line 101: This is a noteworthy change!

src/bogo_shim/config.json
randombit: I think this is ok - we just need to have a DTLS window that's at least as big as what the test expects. I can take this. #4079

randombit (approved):

PR #3883, opened 2024-01-11
PQC: Classic McEliece

FAlbertDev:
### Pull Request Dependencies

* #4373 - for the generic FFI/Python tests

### Description

This PR relates to the Classic McEliece KEM as specified [in this ISO draft](https://classic.mceliece.org/iso-mceliece-20230419.pdf). It also contains the instances defined in the [NIST Round 4 submission](https://classic.mceliece.org/mceliece-spec-20221023.pdf). The test cases were generated using the NIST submissions reference implementation. Note that Classic McEliece (module `cmce`) is not the original McEliece Algorithm that is implemented in Botan's `mce` module. See the [Classic McEliece homepage](https://classic.mceliece.org/), for a brief comparison.

#### TODO Tracker
- [x] Satisfy Botan's CI
- [x] Process TODOs in the code
- [x] Clean-up tests
- [x] Add Botan internal documentation
- [x] Refine/Document code taken from the reference implementation
- [x] Side-channel analysis
- [x] Cleanup Git history
- [x] ~~Use uint8_t as bitvector return type~~ (see: https://github.com/randombit/botan/pull/3883#issuecomment-1907491054)
- [x] Add strong type adapter for bitvector
- [x] Remove test instances from this PR
- [x] Apply ceil_division introduced with HSS/LMS (after merge of https://github.com/randombit/botan/pull/3716). See [Suggestion](https://github.com/randombit/botan/pull/3883#discussion_r1470710399).
- [x] Apply improved load/store methods (after merge of https://github.com/randombit/botan/pull/3869). See suggestions [A](https://github.com/randombit/botan/pull/3883#discussion_r1466414063), [B](https://github.com/randombit/botan/pull/3883#discussion_r1470754242), [C](https://github.com/randombit/botan/pull/3883#discussion_r1470782657).
- [x] Reduce the test vector somewhat (https://github.com/randombit/botan/pull/3883#issuecomment-2025231164)
- [x] Add dedicated raw key loading functions to FFI/Python (see also https://github.com/randombit/botan/discussions/4366)

PR #4081, opened 2024-05-29
Skip the Frodo KAT tests under valgrind and arm32-qemu

randombit:
These test consumes a couple of minutes for each CI run.


reneme (approved):
Perhaps it makes sense to support some kind of "reduced" KAT tests (potentially generically for all PQC algorithms)? A similar issue will very likely arise [with Classic McEliece](https://github.com/randombit/botan/pull/3883#issuecomment-2025231164).

randombit:
> Perhaps it makes sense to support some kind of "reduced" KAT tests (potentially generically for all PQC algorithms)?

Possibly so. I think CI runtimes are generally fine except for valgrind and the arm32 qemu (which for whatever reason is *much* slower than the other qemus) so it doesn't seem that pressing. The only possible issue is that we would miss an ARM32 specific bug since we never run those tests in CI. We could consider add a nightly similar to the valgrind full run.

Issue #4082, opened 2024-05-29
HMS test failure


randombit:
Seen on CI

```
$ ./botan-test --drbg-seed=17D3F1D99C7385E0 hss_lms_sign
Testing Botan 3.5.0 (unreleased, revision git:94c81d62b7e652c752a83ed257eeb0f6c5815b97, distribution unspecified)
Properties:
  CPU flags: rdtsc sse2 ssse3 aes_ni clmul
  drbg_seed: 17D3F1D99C7385E0
hss_lms_sign:
HSS-LMS/ signature generation ran 24 tests in 2.26 sec 1 FAILED
Failure 1: HSS-LMS/ signature generation Modified signature rejected with exception with error Internal error: Codepath that was marked unreachable was reached @src/lib/pubkey/hss_lms/lms.cpp:160 (at src/tests/test_hss_lms.cpp:297)
Note 1: HSS-LMS/ signature generation Accepted invalid signature 0000000100000004000000043B4C...
```

@reneme @FAlbertDev

randombit:
Fixed in #4083

PR #4083, opened 2024-05-29
Reject all HSS-LMS unknown parameter types

randombit:
GH #4082


FAlbertDev (approved):
Thanks. I'll look into the code to see if there are other places where forged signatures may provoke an Exception instead of a Decode_Error.

PR #4085, opened 2024-05-29
Follow-up: Use load_be more efficiently

### LINKED ISSUE #4080: Somehow load_{be,le} of integer without offset doesn't work
In most cases it doesn't matter much but when there is a lone `load_be(array, 0)` it would be nicer if we could omit the index.

https://github.com/randombit/botan/pull/4056#discussion_r1618336576

@reneme can you take a look?

reneme:
See: https://github.com/randombit/botan/pull/4056#discussion_r1615735532

Closes #4080

### DIFF
--- a/src/lib/math/bigint/bigint.cpp
+++ b/src/lib/math/bigint/bigint.cpp
@@ -420,13 +420,15 @@ void BigInt::assign_from_bytes(std::span<const uint8_t> bytes) {
    secure_vector<word> reg((round_up(full_words + (extra_bytes > 0 ? 1 : 0), 8)));
 
    for(size_t i = 0; i != full_words; ++i) {
-      reg[i] = load_be<word>(&bytes[length - sizeof(word) * (i + 1)], 0);
+      reg[i] = load_be<word>(bytes.last<sizeof(word)>());
+      bytes = bytes.first(bytes.size() - sizeof(word));
    }
 
-   if(extra_bytes > 0) {
-      for(size_t i = 0; i != extra_bytes; ++i) {
-         reg[full_words] = (reg[full_words] << 8) | bytes[i];
-      }
+   if(!bytes.empty()) {
+      BOTAN_ASSERT_NOMSG(extra_bytes == bytes.size());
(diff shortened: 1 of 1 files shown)


src/lib/math/bigint/bigint.cpp
```diff
@@ -420,13 +420,15 @@ void BigInt::assign_from_bytes(std::span<const uint8_t> bytes) {
    secure_vector<word> reg((round_up(full_words + (extra_bytes > 0 ? 1 : 0), 8)));
 
    for(size_t i = 0; i != full_words; ++i) {
-      reg[i] = load_be<word>(&bytes[length - sizeof(word) * (i + 1)], 0);
+      reg[i] = load_be<word>(bytes.last<sizeof(word)>());
```
randombit, line 423: Hmm so the loaders can omit the index when loading from a span but not an array?

reneme: If by "array" you mean "raw pointer to data", yes. My guess is, that this wasn't possible before the load/store refactoring either. Or perhaps I forgot to translate a legacy overload; but given that the rest of the code base compiled, I doubt that.

randombit: That's true, the index was always required previously. My point is now the behavior is inconsistent depending on the type of argument, which is obviously surprising.

https://en.wikipedia.org/wiki/Principle_of_least_astonishment

reneme: Oh, right. I believe, leaving out the index-overload for ranges was my deliberate decision. I.e. rather set up the range or view accordingly before invoking the loader than pass an offset/index.

On more general terms: The structure of the loaders (with `detail::load_any` and the constexpr-magic) has the disadvantage that it is fairly hard to see for us which overloads are actually legal. 🤔

randombit (approved):
I'm not super happy about the inconsistent behavior depending on the argument, that was definitely not clear to me in the initial PR. That said this change looks fine.

reneme:
FWIW: I would hope to eventually purge the raw-pointer overloads altogether; making it consistent again...

Issue #4090, opened 2024-06-01
FFI: botan_cipher_get_update_granularity() returns 17


ni4:
....which is quite unlogical if application relies on this value.
Root of this is in the `ffi_choose_update_size()` function and the fact that `mode.update_granularity()` returns 1.

reneme:
I confirm!

This is true for the **decryption** of some AEAD modes. On the C++ level, some report a `minimum_final_size()` (of the tag length) and an update granularity of 1; notably ChaChaPoly, EAX, SIV and CCM. As a result, `ffi_choose_update_size()` produces 17.
Also surprising: for GCM it reports 32 (due to both update granularity and minimum final size reporting 16 and `ffi_choose_update_size()` rounding up "17" to the next multiple of "update granularity").

Note that `botan_cipher_get_ideal_update_granularity()` (introduced in Botan 3.0) always reports reasonable values. However, documentation claims that "ideal granularity" is always a multiple of "update granularity", which obviously isn't the case for the configurations affected by this issue.

We should have caught that in #3951 already, but the tests that sanity-check the granularities aren't up to the job. Thanks a lot for scrutinizing this even before it gets released!

reneme:
@randombit I feel the safest fix for this is to let `botan_cipher_get_update_granularity()` pass-through whatever is returned by `Cipher_Mode::update_granularity()` (see https://github.com/randombit/botan/pull/4094). This would remove the bug-prone logic in `ffi_choose_update_size()`. Functionally, it should be just fine after the changes in #3951 -- most notably the `BOTAN_FFI_ERROR_INSUFFICIENT_BUFFER_SPACE` handling added there.

Nevertheless, one could make a case about semver: Users that rely on the old behavior with the sufficiently sized update granularity, might trip over this. When updating to 3.5.0, they would either need to switch to `botan_cipher_get_IDEAL_update_granularity()` or handle `BOTAN_FFI_ERROR_INSUFFICIENT_BUFFER_SPACE ` properly.

As an alternative to #4094, we could extend the logic in `ffi_choose_update_size()` to scale the minimal update granularity to ensure ideal granularity is a multiple of it. My gut tells me, that this will just make it more error-prone, though: e.g. when `Cipher_Mode`'s update granularity is equal to its ideal granularity.

PR #4094, opened 2024-06-03
FIX: botan_cipher_get_update_granularity()

reneme:
With #3951 `botan_cipher_update()` gained a documented way to communicate the required buffer size to finalize certain cipher modes if not enough bytes were provided by the caller. I feel that this actually obsoletes the internal helper function `ffi_choose_update_size()` altogether. Instead, I suggest to just pass through what the `Cipher_Mode` object reports.

Note that this might harm performance for users that currently use `botan_cipher_get_update_granularity()` (similarly to #3925). Such users should switch to `botan_cipher_get_ideal_update_granularity()`, as explained [in the Botan 3.0 migration guide](https://github.com/randombit/botan/blob/5649a10ec3399d28c8b84ae18fd65fce57bd7bca/doc/migration_guide.rst#cipher-mode-granularity).

### DIFF
--- a/src/lib/ffi/ffi_cipher.cpp
+++ b/src/lib/ffi/ffi_cipher.cpp
@@ -36,37 +36,6 @@ struct botan_cipher_struct final : public botan_struct<Botan::Cipher_Mode, 0xB4A
       size_t m_ideal_update_size;
 };
 
-namespace {
-
-size_t ffi_choose_update_size(Botan::Cipher_Mode& mode) {
-   const size_t update_granularity = mode.update_granularity();
-   const size_t minimum_final_size = mode.minimum_final_size();
-
-   /*
-   * Return the minimum possible granularity given the FFI API constraints that
-   * we require the returned size be > minimum final size.
-   *
-   * If the minimum final size is zero, or the update_granularity is
-   * already greater, just use that.
-   *
-   * Otherwise scale the update_granularity to a sufficient size
-   * to be greater than the minimum.
-   */
-   if(minimum_final_size == 0 || update_granularity > minimum_final_size) {
-      BOTAN_ASSERT_NOMSG(update_granularity > 0);
-      return update_granularity;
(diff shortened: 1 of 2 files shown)


src/lib/ffi/ffi_cipher.cpp
```diff
@@ -36,37 +36,6 @@ struct botan_cipher_struct final : public botan_struct<Botan::Cipher_Mode, 0xB4A
...
 
-namespace {
-
-size_t ffi_choose_update_size(Botan::Cipher_Mode& mode) {
-   const size_t update_granularity = mode.update_granularity();
-   const size_t minimum_final_size = mode.minimum_final_size();
-
-   /*
-   * Return the minimum possible granularity given the FFI API constraints that
-   * we require the returned size be > minimum final size.
```
randombit, line 47: Does this condition no longer hold? I would think this would break callers.

reneme: To the best of my understanding it doesn't hold anymore (after the introduction of #3951). For instance: an AEAD that allows single-byte updates, could now be used via `botan_cipher_update()` with single-byte buffers. When trying to finalize the cipher with a single-byte bulffer, it will return a `BOTAN_FFI_ERROR_INSUFFICIENT_BUFFER_SPACE` and report the size of the required buffer via the out-variable. Repeating the finalize-call with a big-enough buffer will work.

Nevertheless, I totally agree that this could potentially break callers and force them to add similar handling as I explained above. See also the discussion here: https://github.com/randombit/botan/issues/4090#issuecomment-2144841857.

reneme: FYI: #4122 is an alternative that keeps this condition alive. ... to the best of my understanding.

reneme:
Closing as obsolete after #4122 got merged. We might bring this back with 4.0 as a new API.

PR #4095, opened 2024-06-04
Stacked on the ECC PR: Test validation of ECDSA signatures

reneme:
Please feel free to discard or just cherry-pick. We just didn't want to interfere in your PR.


src/tests/test_pcurves.cpp
```diff
@@ -120,10 +120,28 @@ class Pcurve_Ecdsa_Sign_Tests final : public Text_Based_Test {
...
+
+         const auto z = curve.scalar_from_bits_with_trunc(msg);
 
-         return Botan::concat(r_bytes, s_bytes);
+         const auto s_inv = curve.scalar_invert(s);
+         const auto u_1 = curve.scalar_mul(z, s_inv);
+         const auto u_2 = curve.scalar_mul(r, s_inv);
+
+         const auto table = curve.mul2_setup(curve.generator(), pk);
+         const auto x = curve.mul2_vartime(*table, u_1, u_2).to_affine();
```
randombit, line 142: r is not the x coordinate but rather the x coord reduced modulo the group order. Since this applies also to most ECC signature schemes there is a helper for this.

randombit:
Thanks! I’ll pull this in shorly

reneme:
> Thanks! I’ll pull this in shorly

We'll amend this PR with a few clean-up suggestions (instead of these lengthy code review suggestions).

src/tests/test_pcurves.cpp
reneme: d'accord. Done.

PR #4096, opened 2024-06-04
Add CT::value_barrier

randombit:


src/lib/utils/ct_utils.h
```diff
@@ -75,6 +75,29 @@ constexpr inline void unpoison(T& p) {
    BOTAN_UNUSED(p);
 }
 
+/**
+* This function returns its argument, but (if called in a non-constexpr context)
+* attemps to prevent the compiler from reasoning about the value or the possible
```
reneme, line 80:
```suggestion
* attempts to prevent the compiler from reasoning about the value or the possible
```

src/lib/utils/ct_utils.h
```diff
@@ -75,6 +75,29 @@ constexpr inline void unpoison(T& p) {
...
+constexpr inline T value_barrier(T x) {
+   if(!std::is_constant_evaluated()) {
+      /*
+      * For compilers without inline asm, is there something else we can do?
+      *
+      * For instance we could potentially launder the value through a
+      * `volatile T` or `volatile T*`. This would require some experimentation.
+      */
+#if defined(BOTAN_USE_GCC_INLINE_ASM)
+      asm("" : "+r"(x) : /* no input */);
```
reneme, line 93: As per [this blog post](https://theunixzoo.co.uk/blog/2021-10-14-preventing-optimisations.html):

```suggestion
      asm volatile("" : "+r,m"(x) : : "memory);
```

... i don't have much experience with such inline assembler kung-fu, though.

reneme: Here's the original source that the blog post above dissects: https://github.com/google/benchmark/blob/7f0e99af540a333108b92d792923ec7fc9e9fad9/include/benchmark/benchmark.h#L470-L477

... for reasons, they distinguish clang and non-clang.

src/lib/utils/ct_utils.h
randombit: Fixed

src/lib/utils/ct_utils.h
randombit: BoringSSL is currently using `"+r"(x)` for their value barrier. And per valgrind it does seem sufficient, at least with current Clang. Clobbering all memory, especially in the very low level loops where this kind of thing is useful, may really degrade optimizations. Contra to a benchmark, where the clobbering happens at the very top of the call stack. I left a comment though that we may need a "stronger" asm statement in the future, along with a link to that blog post.

reneme:
What was your strategy to introduce `value_barrier` invocations in `ct_uitls.h`? For instance, `ct_is_zero()` seems like a good candidate to get one as well.

randombit:
I eyeballed it but yes I missed several cases - fixed

src/lib/utils/ct_utils.h
```diff
@@ -75,6 +75,38 @@ constexpr inline void unpoison(T& p) {
...
+* range of values. Such optimizations have a way of breaking constant time code.
+*/
+template <typename T>
+constexpr inline T value_barrier(T x) {
+   if(!std::is_constant_evaluated()) {
+      /*
+      * For compilers without inline asm, is there something else we can do?
+      *
+      * For instance we could potentially launder the value through a
+      * `volatile T` or `volatile T*`. This would require some experimentation.
```
reneme, line 90: Looking at [subtle](https://github.com/dalek-cryptography/subtle), they [seem to rely on volatile reads](https://github.com/dalek-cryptography/subtle/blob/6b6a81ad9a6a00c0b42c327eaf4b2f785774377e/src/lib.rs#L217-L243) to prevent the compiler from assuming certain values.

[`core::ptr::read_volatile`](https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/core/ptr/fn.read_volatile.html) is [translated to](https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/src/core/ptr.rs.html#453-455) `intrinsics::volatile_load` which, to the best of my understanding, is compiled to a [volatile memory access in LLVM](https://llvm.org/docs/LangRef.html#volatile-memory-accesses) bit code. So perhaps the suggested `volatile` laundering is worth a try.

I'll have a go at it.

reneme: This does indeed look promising: https://godbolt.org/z/bKTePqYeb

The resulting code is slightly different, but in both cases the compiler optimization is omitted.

src/lib/utils/ct_utils.h
```diff
@@ -126,17 +158,23 @@ class Mask final {
...
+      static constexpr Mask<T> is_zero(T x) { return Mask<T>(ct_is_zero<T>(value_barrier<T>(x))); }
 
       /**
       * Return a Mask<T> which is set if x == y
       */
-      static constexpr Mask<T> is_equal(T x, T y) { return Mask<T>::is_zero(static_cast<T>(x ^ y)); }
+      static constexpr Mask<T> is_equal(T x, T y) {
+         const T diff = x ^ y;
+         return Mask<T>::is_zero(value_barrier<T>(diff));
+      }
```
reneme, line 169: I believe it is not enough to apply the `value_barrier` on the result of the xor. If the compiler can somehow infer that both `x` and `y` have a small value-space, it might still come up with some sort of jump table. At this point I'm not sure when an example becomes too contrived, tbh.

Same is true for the other factory methods as well.

randombit: Unfortunately I think this kind of work is necessarily reactive rather than proactive. The best we can do IMO is dramatically increase the use of constant time tainting which will help detect cases where things are going bad.

That said I do agree is_equal is prone to issues especially since it's not uncommon for one arg to be a constant and the other a secret. I inserted value_barrier on both args

src/lib/utils/ct_utils.h
randombit: If you compare the codegen at `-O3` for both GCC and Clang, codegen is much worse when using `volatile`. Many redundant reads/writes to memory

```
        mov     dword ptr [rsp - 4], r8d
        mov     r8d, dword ptr [rsp - 4]
```

Which is fine since we're using inline asm for those compilers anyway.

I'm inclined to merge this as is since the only outlier is MSVC and that compiler is not (that I have heard of) capable of the kind of range analysis that causes these problems with rewriting of constant time code. That might change if MSVC ever switches to LLVM as their backend though!

randombit:
@reneme FYI I'm planning on merging this tonight to unblock #3979

reneme (approved):

Issue #4049, opened 2024-05-11
Update clang-format rules


randombit:
Not going to do this anytime soon, best to wait for when things are quiet in terms of active PRs. Maybe end of the year around holidays? This is just to track/discuss possible changes. It would be best to make any such changes all at once to avoid churn.

* Change `AllowShortFunctionsOnASingleLine` from `Inline` to `Empty`
* Change `AllowAllArgumentsOnNextLine` from `true` to `false`
* Change `AllowAllParametersOfDeclarationOnNextLine` from `true` to `false`

As with any rules based formatter, sometimes the result of any change is worse formatting vs what was before, but overall I think these would lead to more readable code.

By EOY Github Actions should have Ubuntu 24.04 available and we could consider switching from `clang-format 15` to `clang-format 17` at the same time.

KaganCanSit:
> If we are able to upgrade `clang-format`
> 
> * `BreakAfterAttributes = Always`
> * `BreakBeforeInlineAsmColon = OnlyMultiLine`

The second style option content is written as "BreakBeforeInlineASMColon". It will save time while editing. In the final case, the new two lines in the `.clang-format` file will be as follows.

```
BreakAfterAttributes: Always
BreakBeforeInlineASMColon: OnlyMultiline
```
You can use [link here](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) as a resource.


**Update:**
Hello @randombit, @reneme,
To help you out, I've implemented all the changes you mentioned in the issue as separate commits in my own branch [here](https://github.com/KaganCanSit/botan/tree/update-clang-format-rules).

However, I think reviewing this branch will be quite difficult, even with separate commits.

I thought it might be better to ask here rather than create a request. My idea was to review them with separate commits and then squash them together before merge.  

Best regards.

~**Update:**
Perhaps I can only help with the new lines to be added. They have a fairly minor impact. It might be better for you maintainers to handle the rest.~

KaganCanSit:
Hello @randombit,

You indicated that you wanted to hold the issue notification here. However, I wanted to address it while I had time. You can view it on the [branch/comment here](https://github.com/KaganCanSit/botan/commit/1af04dba5b08f0f71f1876b2a887ca70edb8ae1b), or I can create a PR request if you want. I'll leave it to you.

What I did;
* I applied the changes you specified to the `.clang-format` file and added the style options content that came in version 17.
* I ran the call below.

``` bash
[...]

PR #4100, opened 2024-06-04
Use clang-format 17

randombit:
This may be slightly disruptive wrt ongoing PRs but the reformatting seems quite minimal and mostly more "correct" so I don't think this will be a problem.

This upgrade does enable some new options that I'd like to apply (#4049) but that part is deferred until the PR queue is clear.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,7 +177,7 @@ jobs:
             host_os: ubuntu-24.04
           - target: format
             compiler: gcc
-            host_os: ubuntu-22.04
+            host_os: ubuntu-24.04
 
     runs-on: ${{ matrix.host_os }}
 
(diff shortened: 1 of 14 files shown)


reneme (approved):
The changes it makes look sensible to me. Especially happy about the more reasonable formatting of concepts!

reneme:
Ubuntu 24.04 provides clang-format-18 as well. Any specific reason to restrict to 17?

randombit:
> Ubuntu 24.04 provides clang-format-18 as well. Any specific reason to restrict to 17?

Not really. Sort of. Arch has still not shipped LLVM 18 for whatever reason and I don't want to have to build it from source on each of my machines. :sweat_smile: 

Bumped it to 18, if the diff isn't overly large lets do that.

Issue #4103, opened 2024-06-07
MSVC 19.40 miscompilation/crash


randombit:
GH actions builds started failing recently. I've been seeing occasional crashes in CI, which retries fixed. Now they seem to be becoming more extensive

> INFO: Auto-detected compiler version msvc 19.39

Works fine

> INFO: Auto-detected compiler version msvc 19.40

  Command '.\botan-test --data-dir=.\src\tests\data --run-memory-intensive-tests --test-results-dir=junit_results --report-properties=ci_target:static,os:windows' failed with error code 3221225477

randombit:
Possibly related: https://github.com/actions/runner-images/issues/10004

randombit:
Combo Visual C++ and Github fuckup #4105

reneme:
Yikes! I saw the occasional MSVC fails as well. Probably we should start trusting our CI setup more and investigate earlier... 😇

Hopefully this CI issue doesn't become a long-standing feature as the previous clang problem. 😨

PR #4104, opened 2024-06-07
Drop MSVC optimizations to O1

randombit:
Due to probable miscompilation/crash in 19.40 (#4103)


randombit:
Not an MSVC bug, rather yet another Github Actions image fuckup

PR #4106, opened 2024-06-07
Use 24.04 image for valgrind and cross-i386 CI images

randombit:


randombit:
The Ubuntu 24.04 image used by GH actions is configured in a way that prevents executing 32 bit binaries, so the i386 build remains on 22.04https://github.com/actions/runner-images/issues/9977

PR #4108, opened 2024-06-07
Use CT::Mask instead of multiplication in polynomial doubling

randombit:


src/lib/utils/poly_dbl/poly_dbl.cpp
```diff
@@ -29,14 +30,22 @@ enum class MinWeightPolynomial : uint32_t {
...
 };
 
+/**
+* If the top bit of c is set, returns the carry (the polynomial)
+*
+* Otherwise returns zero.
+*/
+template <MinWeightPolynomial P>
+inline uint64_t return_carry(uint64_t c) {
+   return CT::Mask<uint64_t>::expand(c >> 63).if_set_return(static_cast<uint64_t>(P));
```
reneme, line 40: Perhaps use `expand_top_bit<>` in `bit_ops.h`?

reneme (approved):

src/lib/utils/poly_dbl/poly_dbl.cpp
randombit: Added `CT::Mask::expand_top_bit` since all uses of `expand_top_bit` really should be using `value_barrier` as well

PR #4110, opened 2024-06-10
Fix: EC private key encoding

FAlbertDev:
[RFC 5915 - Section 3](https://datatracker.ietf.org/doc/html/rfc5915#section-3):
> privateKey is the private key.  It is an octet string of length ceiling (log2(n)/8) (where n is the order of the curve) obtained from the unsigned integer via the Integer-to-Octet-String-Primitive (I2OSP) defined in [[RFC3447](https://datatracker.ietf.org/doc/html/rfc3447)].

I2OSP encoding is basically big-endian encoding with fixed length. Currently, however, the length is not fixed, i.e., a secret key starting with a zero byte is encoded with one byte less. For most curves, this is unlikely; however, for curve secp521r1, this occurs in every second case (because order: 521 bits = 65 bytes + 1 bit). 

Also, this induces a side channel that leaks the number of zero bytes at the beginning of the secret key when encoding or decoding it.

### DIFF
--- a/src/lib/pubkey/ecc_key/ecc_key.cpp
+++ b/src/lib/pubkey/ecc_key/ecc_key.cpp
@@ -117,7 +117,7 @@ EC_PrivateKey::EC_PrivateKey(RandomNumberGenerator& rng,
 }
 
 secure_vector<uint8_t> EC_PrivateKey::raw_private_key_bits() const {
-   return m_private_key.serialize<secure_vector<uint8_t>>();
+   return m_private_key.serialize<secure_vector<uint8_t>>(domain().get_order_bytes());
 }
 
 secure_vector<uint8_t> EC_PrivateKey::private_key_bits() const {


randombit (approved):
Change looks fine however I would request changing the commit message as this doesn't actually impact key encoding at all (that's `private_key_bits`) so the change is potentially confusing.

Actually I don't recall offhand what #3437 is for. I guess just to consolidate what had been some one-off/algorithm specific accessors for this information? Impressive considering I wrote that barely a year ago. @reneme do you remember?

FAlbertDev:
Thanks for the quick review! 
I wrote 'fix encoding' because `private_key_bits` calls `raw_private_key_bits` internally:
```cpp
secure_vector<uint8_t> EC_PrivateKey::private_key_bits() const {
   return DER_Encoder()
      .start_sequence()
      .encode(static_cast<size_t>(1))
      .encode(raw_private_key_bits(), ASN1_Type::OctetString)
      .start_explicit_context_specific(1)
      .encode(m_public_key.encode(EC_Point_Format::Uncompressed), ASN1_Type::BitString)
      .end_cons()
      .end_cons()
      .get_contents();
}
```

randombit:
Oh so it does! Thankfully that only changed in bc5137cd9d49af037a22b43d09421b45d026cf3f so this bad behavior never made it out to a release. Ideally we would catch this with a test though - I'll attend to that

PR #4115, opened 2024-06-11
Add CT::Choice

randombit:


src/lib/utils/ct_utils.h
```diff
@@ -107,6 +107,54 @@ constexpr inline T value_barrier(T x) {
...
+      /**
+      * If v == 0 return an unset (false) Choice, otherwise a set Choice
+      */
+      template <typename T>
+      constexpr static Choice from_int(T v)
+         requires(std::is_unsigned<T>::value)
+      {
+         const auto v_is_0 = ct_is_zero<T>(value_barrier<T>(v));
+         return Choice(ct_is_zero<uint32_t>(static_cast<uint32_t>(v_is_0)));
+      }
```
reneme, line 126:
```suggestion
      template <std::unsigned_integral T>
      constexpr static Choice from_int(T v)
      {
         const auto v_is_0 = ct_is_zero<T>(value_barrier<T>(v));
         return Choice(ct_is_zero<uint32_t>(static_cast<uint32_t>(v_is_0)));
      }
```

src/lib/utils/ct_utils.h
```diff
@@ -107,6 +107,54 @@ constexpr inline T value_barrier(T x) {
...
+
+      template <typename T>
+      constexpr T choose(T a, T b)
+         requires(std::is_unsigned<T>::value)
+      {}
```
reneme, line 131:
```suggestion
      template <std::unsigned_integral T>
      constexpr T choose(T a, T b)
      {}
```

... also: this is lacking an implementation (and perhaps a docstring: what is chosen over what?).

src/lib/utils/ct_utils.h
```diff
@@ -107,6 +107,54 @@ constexpr inline T value_barrier(T x) {
...
+
+      constexpr Choice operator!() const { return Choice(~value()); }
+
+      constexpr Choice operator&&(const Choice& other) const { return Choice(value() & other.value()); }
+
+      constexpr Choice operator||(const Choice& other) const { return Choice(value() | other.value()); }
+
+      constexpr Choice operator!=(const Choice& other) { return Choice(value() ^ other.value()); }
```
reneme, line 139:
```suggestion
      constexpr Choice operator!=(const Choice& other) const { return Choice(value() ^ other.value()); }
```

src/lib/utils/ct_utils.h
```diff
@@ -107,6 +107,54 @@ constexpr inline T value_barrier(T x) {
...
+
+      /// Unsafe conversion to bool
+      constexpr bool as_bool() const { return m_value != 0; }
+
+      /// Return the masked value
+      constexpr uint32_t value() const { return value_barrier(m_value); }
```
reneme, line 145: If we ensure that all "constructors" use a `value_barrier`, we (technically) shouldn't need it for this accessor, as the value in `m_value` should never be predictable for the compiler. But then again: better safe than sorry, I guess.

src/lib/utils/ct_utils.h
```diff
@@ -107,6 +107,54 @@ constexpr inline T value_barrier(T x) {
...
+
+      Choice(const Choice& other) = default;
+      Choice(Choice&& other) = default;
+      Choice& operator=(const Choice& other) = default;
+      Choice& operator=(Choice&& other) = default;
```
reneme, line 150:
```suggestion
      constexpr Choice(const Choice& other) = default;
      constexpr Choice(Choice&& other) = default;
      constexpr Choice& operator=(const Choice& other) noexcept = default;
      constexpr Choice& operator=(Choice&& other) noexcept = default;
```

src/lib/utils/ct_utils.h
```diff
@@ -107,6 +107,54 @@ constexpr inline T value_barrier(T x) {
    }
 }
 
+/**
+* A Choice is used for constant-time conditionals.
+*
+* Internally it always is either all 0 bits or all 1 bits.
```
reneme, line 113:
```suggestion
* A Choice is used for constant-time conditionals.
*
* Internally it always is either all 0 bits or all 1 bits and measures are
* taken to block compilers from reasoning about the expected value of
* a Choice.
```

src/lib/utils/ct_utils.h
```diff
@@ -146,6 +194,20 @@ class Mask final {
...
+      static constexpr Mask<T> from_choice(Choice c) {
+         if constexpr(sizeof(T) <= sizeof(uint32_t)) {
+            // Take advantage of the fact that Choice's mask is always
+            // either zero or all ones
+            return Mask<T>(static_cast<T>(c.value()));
+         } else {
+            return ~Mask<T>::is_zero(c.value());
+         }
+
+      }
```
reneme, line 209:
```suggestion
         }
      }
```

src/lib/utils/ct_utils.h
```diff
@@ -107,6 +107,54 @@ constexpr inline T value_barrier(T x) {
...
+      constexpr Choice operator!() const { return Choice(~value()); }
+
+      constexpr Choice operator&&(const Choice& other) const { return Choice(value() & other.value()); }
+
+      constexpr Choice operator||(const Choice& other) const { return Choice(value() | other.value()); }
+
+      constexpr Choice operator!=(const Choice& other) { return Choice(value() ^ other.value()); }
+
+      /// Unsafe conversion to bool
+      constexpr bool as_bool() const { return m_value != 0; }
```
reneme, line 142: In what sense is this unsafe? Given that the compiler shouldn't be able to reason about `m_value` it shouldn't be able to reason about the `bool` either. Isn't it only the potential downstream usage of the bool that makes it "unsafe"?

src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -769,8 +763,9 @@ class ProjectiveCurvePoint {
 
          bool any_identity = false;
          for(size_t i = 0; i != N; ++i) {
-            if(projective[i].is_identity()) {
+            if(projective[i].is_identity().as_bool()) {
                any_identity = true;
+               break;
```
reneme, line 768: Perhaps this warrants a comment. :)

reneme (approved):
A few coding nits and questions. Perhaps some things should be validated by tests. E.g. the double-application of `ct_is_zero()` in the `from_int()` factory wasn't immediately clear to me. A test that uses `from_int()` based on `uint8_t` through `uint64_t` would ensure that no one ever removes it.

src/lib/utils/ct_utils.h
randombit: This is stray - I decided to go with converting Choice to a Mask for selecting a T but forgot to remove this

src/lib/utils/ct_utils.h
randombit: Yes and no :)

`bool` has a constrained value (it is either `true` or `false`) which provides a choke point on the possible range of values, so it's entirely possible for the compiler to reason about the cases. Consider a hypothetical constant time call:

```
bool b = choice.as_bool();
uint32_t f = expensive_and_pure_fn(inputs);
uint32_t mask = b * 0xFFFFFFFF;
return f & mask;
```

The compiler doesn't know what `b` is, but it knows it is either `true` or `false`. From there it is easy to see that `f` can be omitted entirely if `b` is false, so transforms into

```
return (b) ? expensive_and_pure_fn(inputs) : 0;
```

That's why the internals of this are using a uint32_t - be a large enough state that, assuming the compiler is willing to forget what it knows about the contents, it won't try to optimize on a case-by-case enumeration of the possible states.

src/lib/utils/ct_utils.h
randombit: I would agree with you in theory!

There are I suppose two different ways we could approach this

1. Use what seems like it should be sufficient (eg, just in the constructor). Test const-timeness with latest available compilers. If something fails, we need [more `value_barrier`](https://www.youtube.com/watch?v=cVsQLlk-T0s)
2. Use it kind of wherever. (Current shotgun approach)

Part of the problem is we're trying to anticipate any compiler optimizations that might be made in the future. But I think the bigger issue is that our current const timeness testing is just insufficient to get a good signal from (1)

src/lib/utils/ct_utils.h
randombit: Removed

src/lib/math/pcurves/pcurves_impl.h
randombit: Added a comment here

randombit:
I added some tests, mostly of `from_int` since that's the only tricky part. I also added some comments to explain what's happening and hopefully ward off someone trying to "simplify" the code in the future.

src/lib/utils/ct_utils.h
randombit: I added a comment in the source explaining more about why the conversion itself is potentially unsafe

reneme: So the relevant point is that, even if the compiler cannot infer the actual `bool` value, it might still optimize for the case that it is one or the other. Docstring is also making this clear. Thanks.

reneme (approved):
More verbose comments (along with the test) are really helpful. Thanks.

PR #4116, opened 2024-06-12
x509: add TNAuthList extension

furmur:
support parsing of the [TNAuthList](https://www.rfc-editor.org/rfc/rfc8226) extension for STIR certificates


src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,66 @@ class OCSP_NoCheck final : public Certificate_Extension {
...
+            ASN1_String spc;                          //ServiceProviderCode
+            std::vector<TelephoneNumberRange> range;  //TelephoneNumberRange
+            ASN1_String one;                          //TelephoneNumber
+      };
+
+      TNAuthList() = default;
+
+      std::unique_ptr<Certificate_Extension> copy() const override { return std::make_unique<TNAuthList>(*this); }
+
+      static OID static_oid() { return OID("1.3.6.1.5.5.7.1.26"); }
```
reneme, line 545: Please register this OID in the `oids.txt` and `oid_maps.cpp`. Here's how to do that:

Add the new OID to `oids.txt`, like so:

```diff

diff --git a/src/build-data/oids.txt b/src/build-data/oids.txt
index 781efa928..6facb03a6 100644
--- a/src/build-data/oids.txt
+++ b/src/build-data/oids.txt
@@ -317,7 +317,9 @@
 2.5.29.35 = X509v3.AuthorityKeyIdentifier
 2.5.29.36 = X509v3.PolicyConstraints
 2.5.29.37 = X509v3.ExtendedKeyUsage
+
 1.3.6.1.5.5.7.1.1 = PKIX.AuthorityInformationAccess
+1.3.6.1.5.5.7.1.26 = PKIX.TNAuthList

 2.5.29.32.0 = X509v3.AnyPolicy
```

Then run this script to auto-generate `oid_maps.cpp` from `oids.txt`. (Perhaps take note of https://github.com/randombit/botan/pull/4120)

`python3 src/scripts/dev_tools/gen_oids.py oids > src/lib/asn1/oid_maps.cpp`

src/lib/x509/x509_ext.cpp
```diff
@@ -764,6 +768,39 @@ void CRL_Issuing_Distribution_Point::decode_inner(const std::vector<uint8_t>& bu
    BER_Decoder(buf).decode(m_distribution_point).verify_end();
 }
 
+void TNAuthList::TNEntry::encode_into(DER_Encoder&) const {}
```
reneme, line 771: Please throw `Botan::Not_Implemented`.

src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,66 @@ class OCSP_NoCheck final : public Certificate_Extension {
       void decode_inner(const std::vector<uint8_t>&) override;
 };
 
+/**
+* TNAuthList extension
+*
+* RFC8226 Secure Telephone Identity Credentials
+*   https://www.rfc-editor.org/rfc/rfc8226#section-9
+*/
+class BOTAN_PUBLIC_API(3, 6) TNAuthList final : public Certificate_Extension {
```
reneme, line 509:
```suggestion
class BOTAN_PUBLIC_API(3, 5) TNAuthList final : public Certificate_Extension {
```

... Botan 3.5.0 is coming soon. Let's be optimistic. :)

src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,66 @@ class OCSP_NoCheck final : public Certificate_Extension {
...
+   public:
+      class TNEntry final : public ASN1_Object {
+         public:
+            enum Type { TN_ServiceProviderCode = 0, TN_TelephoneNumberRange, TN_TelephoneNumber };
```
reneme, line 513: Given that this is burried so deep into a class structure, I think we can live without the `TN_` prefix here.

src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,66 @@ class OCSP_NoCheck final : public Certificate_Extension {
...
+      OID oid_of() const override { return static_oid(); }
+
+      const std::vector<TNEntry>& get_entries() const { return m_tn_entries; }
+
+   private:
+      std::string oid_name() const override { return "X509v3.TNAuthList"; }
+
+      bool should_encode() const override { return false; }
+
+      std::vector<uint8_t> encode_inner() const override { return {}; }
```
reneme, line 556: Please throw `Botan::Not_Implemented`.

src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,66 @@ class OCSP_NoCheck final : public Certificate_Extension {
...
+      std::unique_ptr<Certificate_Extension> copy() const override { return std::make_unique<TNAuthList>(*this); }
+
+      static OID static_oid() { return OID("1.3.6.1.5.5.7.1.26"); }
+
+      OID oid_of() const override { return static_oid(); }
+
+      const std::vector<TNEntry>& get_entries() const { return m_tn_entries; }
+
+   private:
+      std::string oid_name() const override { return "X509v3.TNAuthList"; }
```
reneme, line 552: I'm guessing this should come under `PKIX`, aka `1.3.6.1.5.5.7.1`.

src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,66 @@ class OCSP_NoCheck final : public Certificate_Extension {
...
+
+      bool should_encode() const override { return false; }
```
reneme, line 554: Why shouldn't it? :) .. its just not implemented, which is fine, IMO.

src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,66 @@ class OCSP_NoCheck final : public Certificate_Extension {
...
+         public:
+            enum Type { TN_ServiceProviderCode = 0, TN_TelephoneNumberRange, TN_TelephoneNumber };
+
+            struct TelephoneNumberRange {
+                  ASN1_String start;  //TelephoneNumber (IA5String)
+                  size_t count;       //2..MAX
+            };
+
+            TNEntry() = default;
+            virtual ~TNEntry() = default;
```
reneme, line 521: No need to make this explicit.

src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,66 @@ class OCSP_NoCheck final : public Certificate_Extension {
...
+            void encode_into(DER_Encoder&) const override;
+            void decode_from(class BER_Decoder& from) override;
+
+            Type get_type() const { return type; }
+
+            const std::string& getServiceProviderCode() const { return spc.value(); }
+
+            const std::vector<TelephoneNumberRange>& getTelephoneNumberRange() const { return range; }
+
+            const std::string& getTelephoneNumber() const { return one.value(); }
```
reneme, line 532: Typically, getters in Botan aren't prefixed and methods are in snake case:

```suggestion
            Type type() const { return type; }

            const std::string& service_provider_code() const { return spc.value(); }

            const std::vector<TelephoneNumberRange>& telephone_number_range() const { return range; }

            const std::string& telephone_number() const { return one.value(); }
```

src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,66 @@ class OCSP_NoCheck final : public Certificate_Extension {
...
+
+      TNAuthList() = default;
+
+      std::unique_ptr<Certificate_Extension> copy() const override { return std::make_unique<TNAuthList>(*this); }
+
+      static OID static_oid() { return OID("1.3.6.1.5.5.7.1.26"); }
+
+      OID oid_of() const override { return static_oid(); }
+
+      const std::vector<TNEntry>& get_entries() const { return m_tn_entries; }
```
reneme, line 549:
```suggestion
      const std::vector<TNEntry>& entries() const { return m_tn_entries; }
```

src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,66 @@ class OCSP_NoCheck final : public Certificate_Extension {
...
+                  size_t count;       //2..MAX
+            };
+
+            TNEntry() = default;
+            virtual ~TNEntry() = default;
+
+            void encode_into(DER_Encoder&) const override;
+            void decode_from(class BER_Decoder& from) override;
+
+            Type get_type() const { return type; }
```
reneme, line 526:
```suggestion
            Type type() const { return type; }
```

src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,66 @@ class OCSP_NoCheck final : public Certificate_Extension {
...
+
+            const std::vector<TelephoneNumberRange>& getTelephoneNumberRange() const { return range; }
+
+            const std::string& getTelephoneNumber() const { return one.value(); }
+
+         private:
+            Type type;
+            ASN1_String spc;                          //ServiceProviderCode
+            std::vector<TelephoneNumberRange> range;  //TelephoneNumberRange
+            ASN1_String one;                          //TelephoneNumber
```
reneme, line 538: Members in Botan are always prefixed with `m_`:

```suggestion
            Type m_type;
            ASN1_String m_spc;                          //ServiceProviderCode
            std::vector<TelephoneNumberRange> m_range;  //TelephoneNumberRange
            ASN1_String m_one;                          //TelephoneNumber
```

src/lib/x509/x509_ext.cpp
```diff
@@ -764,6 +768,39 @@ void CRL_Issuing_Distribution_Point::decode_inner(const std::vector<uint8_t>& bu
...
+
+void TNAuthList::TNEntry::decode_from(class BER_Decoder& ber) {
+   BER_Object obj = ber.get_next_object();
+
+   type = static_cast<decltype(type)>(obj.type_tag());
```
reneme, line 776: Casting the `type_tag()` looks a bit funky to me. Admittedly though, my knowledge about the `BER_Decoder` is somewhat limited. @randombit please have a look. :)

src/tests/test_x509_tn_auth_list.cpp
```diff
@@ -0,0 +1,80 @@
+/*
+* (C) 2006,2011,2012,2014,2015 Jack Lloyd
+* (C) 2022 René Meusel, Rohde & Schwarz Cybersecurity
```
reneme, line 3: Copy-pasta. :) ... Feel free to add your credits in the files you modify (significantly).

src/tests/test_x509_tn_auth_list.cpp
```diff
@@ -0,0 +1,80 @@
...
+         result.test_eq("one entry type", tn_entries[2].get_type() == TNAuthList::TNEntry::TN_TelephoneNumber, true);
+         result.test_eq("one entry data", tn_entries[2].getTelephoneNumber(), "333");
+
+         results.push_back(result);
+         result.end_timer();
+         return results;
+      }
+};
+
+BOTAN_REGISTER_TEST("x509", "x509_tn_auth_list", X509test_TNAuthList_Tests);
```
reneme, line 76: Thanks for the test case! I think this could live in `unit_x509.cpp`, though. For instance, look at the `test_x509_authority_information_access()` test for an example. Also removes the need of having a `std::vector<Test::Result>`. Simply return your single test result from a free-standing test function and call/push-back it in `X509_Cert_Unit_Tests::run()`.

reneme (commented):
Thanks for this contribution! I've added several comments and would appreciate a scroll-through by @randombit, especially for the BER-decoding bits.

src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,66 @@ class OCSP_NoCheck final : public Certificate_Extension {
...
 
+/**
+* TNAuthList extension
+*
+* RFC8226 Secure Telephone Identity Credentials
+*   https://www.rfc-editor.org/rfc/rfc8226#section-9
+*/
+class BOTAN_PUBLIC_API(3, 6) TNAuthList final : public Certificate_Extension {
+   public:
+      class TNEntry final : public ASN1_Object {
```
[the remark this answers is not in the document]
furmur: rewritten to use `std::variant`

src/tests/test_x509_tn_auth_list.cpp
furmur: moved to the `unit_x509.cpp`

src/lib/x509/x509_ext.cpp
furmur: `type_tag()` contains enum value here because of the explicit context-specific tagging. on the out-of-range values it would cast without errors anyway and throw exception in the default switch case below.

furmur:
thank you for so meticulous code review. modified according to the comments.

src/lib/x509/x509_ext.cpp
```diff
@@ -764,6 +768,51 @@ void CRL_Issuing_Distribution_Point::decode_inner(const std::vector<uint8_t>& bu
    BER_Decoder(buf).decode(m_distribution_point).verify_end();
 }
 
+void TNAuthList::Entry::encode_into(DER_Encoder&) const {
+   throw Botan::Not_Implemented("TNAuthList extension serialization is not supported");
```
reneme, line 772:
```suggestion
   throw Botan::Not_Implemented("TNAuthList extension entry serialization is not supported");
```

... unlikely that we'll ever see this error message. Just making it unique for good measure. `TNAuthList::encode_inner()` used the exact same error string, making it harder to interpret error logs.

src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,65 @@ class OCSP_NoCheck final : public Certificate_Extension {
...
+            void encode_into(DER_Encoder&) const override;
+            void decode_from(class BER_Decoder& from) override;
+
+            Type type() const { return m_type; }
+
+            const std::string& service_provider_code() const { return std::get<ASN1_String>(m_data).value(); }
+
+            const RangeContainer& telephone_number_range() const { return std::get<RangeContainer>(m_data); }
+
+            const std::string& telephone_number() const { return std::get<ASN1_String>(m_data).value(); }
```
reneme, line 532: Yay! `std::variant<>`. Thanks a lot!

One small user experience thing: If someone calls the wrong accessor, they'll get a weird `std::bad_variant_access` exception that they may or may not understand.

Let's make those a bit more friendly:

```suggestion
            const std::string& service_provider_code() const {
                BOTAN_STATE_CHECK(type() == Type::TN_ServiceProviderCode);
                return std::get<ASN1_String>(m_data).value();
            }

            const RangeContainer& telephone_number_range() const {
                BOTAN_STATE_CHECK(type() == Type::TN_TelephoneNumberRange);
                return std::get<RangeContainer>(m_data);
            }

            const std::string& telephone_number() const {
                BOTAN_STATE_CHECK(type() == Type::TN_TelephoneNumber);    
                return std::get<ASN1_String>(m_data).value();
            }
```

... that'll produce `Botan::Invalid_State` with a reasonable error message when we're not in the correct state for the respective accessor.

src/lib/x509/x509_ext.h
reneme: I believe you missed this. :)

src/lib/x509/x509_ext.cpp
```diff
@@ -764,6 +768,51 @@ void CRL_Issuing_Distribution_Point::decode_inner(const std::vector<uint8_t>& bu
...
+         BER_Decoder(obj).decode(one_string);
+         m_data = std::move(one_string);
+      } break;
+      default:
+         throw Decoding_Error("unexpected entry type");
+   };
+}
+
+std::vector<uint8_t> TNAuthList::encode_inner() const {
+   throw Botan::Not_Implemented("TNAuthList extension serialization is not supported");
```
randombit, line 809: No need for `Botan::` prefix here, we are inside the namespace

src/lib/x509/x509_ext.h
randombit: Not only that - since the variant shares a type (`ASN1_String`) for two of the possible data types, currently if `type() == TN_TelephoneNumber` then `service_provider_code()` works and returns the wrong thing.

src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,65 @@ class OCSP_NoCheck final : public Certificate_Extension {
...
+* TNAuthList extension
+*
+* RFC8226 Secure Telephone Identity Credentials
+*   https://www.rfc-editor.org/rfc/rfc8226#section-9
+*/
+class BOTAN_PUBLIC_API(3, 5) TNAuthList final : public Certificate_Extension {
+   public:
+      class Entry final : public ASN1_Object {
+         public:
+            enum Type { TN_ServiceProviderCode = 0, TN_TelephoneNumberRange, TN_TelephoneNumber };
```
randombit, line 513: Currently the code implicitly assumes that the values of this `enum` match the ASN.1. Let's make this a bit more explicit by directly initializing each with their respective value, along with a comment that these values match the respective tagging of `TNEntry` in https://datatracker.ietf.org/doc/html/rfc8226#section-9

src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,65 @@ class OCSP_NoCheck final : public Certificate_Extension {
...
+
+            Type type() const { return m_type; }
+
+            const std::string& service_provider_code() const { return std::get<ASN1_String>(m_data).value(); }
+
+            const RangeContainer& telephone_number_range() const { return std::get<RangeContainer>(m_data); }
+
+            const std::string& telephone_number() const { return std::get<ASN1_String>(m_data).value(); }
+
+            const DataContainer& data() { return m_data; }
```
randombit, line 534: Is there a good reason to expose this to users? It seems like `type` plus the accessors gives the end user everything they need. It also seems to me error prone, vs the specific accessors which verify the type.

src/lib/x509/x509_ext.cpp
```diff
@@ -764,6 +768,51 @@ void CRL_Issuing_Distribution_Point::decode_inner(const std::vector<uint8_t>& bu
...
+         }
+         list.end_cons();
+      } break;
+      case TN_TelephoneNumber: {
+         ASN1_String one_string;
+         BER_Decoder(obj).decode(one_string);
+         m_data = std::move(one_string);
+      } break;
+      default:
+         throw Decoding_Error("unexpected entry type");
```
randombit, line 804: Let's make this a bit more explicit to help out the poor user who encounters this one day

```
throw Decoding_Error(fmt("Unexpected TNEntry type {}", m_type));
```

src/lib/x509/x509_ext.cpp
```diff
@@ -764,6 +768,51 @@ void CRL_Issuing_Distribution_Point::decode_inner(const std::vector<uint8_t>& bu
...
+   };
+}
+
+std::vector<uint8_t> TNAuthList::encode_inner() const {
+   throw Botan::Not_Implemented("TNAuthList extension serialization is not supported");
+}
+
+void TNAuthList::decode_inner(const std::vector<uint8_t>& in) {
+   BER_Decoder(in).decode_list(m_tn_entries).verify_end();
```
randombit, line 813: The RFC requires the list be non-empty (`TNAuthorizationList ::= SEQUENCE SIZE (1..MAX) OF TNEntry`) so lets check that here

src/lib/x509/x509_ext.cpp
```diff
@@ -764,6 +768,51 @@ void CRL_Issuing_Distribution_Point::decode_inner(const std::vector<uint8_t>& bu
...
+            list.decode(entry.start);
+            list.decode(entry.count);
+            range_items.emplace_back(std::move(entry));
+         }
+         list.end_cons();
+      } break;
+      case TN_TelephoneNumber: {
+         ASN1_String one_string;
+         BER_Decoder(obj).decode(one_string);
+         m_data = std::move(one_string);
```
randombit, line 801: The `TelephoneNumber` type is supposed to be of a limited size and from a strict subset of ASCII

```
TelephoneNumber ::= IA5String (SIZE (1..15)) (FROM ("0123456789#*"))
```

We should check this rather than potentially pass on invalid data to the user.

randombit (changes requested):
Thanks for the contribution @furmur this is absolutely within scope for the library. I have a few more comments on top of @reneme's review but this is already looking really good.

src/lib/x509/x509_ext.h
furmur: removed data() accessor

src/lib/x509/x509_ext.cpp
furmur: fixed

src/lib/x509/x509_ext.cpp
furmur: * validator `valid_TelephoneNumber()`  for `TelephoneNumberRange.start` and `one` entries
* added checks for all ASN.1 constraints

src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,76 @@ class OCSP_NoCheck final : public Certificate_Extension {
...
+            /* TNEntry choice values
+             * see: https://datatracker.ietf.org/doc/html/rfc8226#section-9 */
+            enum Type { TN_ServiceProviderCode = 0, TN_TelephoneNumberRange = 1, TN_TelephoneNumber = 2 };
+
+            struct TelephoneNumberRange {
+                  ASN1_String start;  //TelephoneNumber (IA5String)
+                  size_t count;       //2..MAX
+            };
+
+            static bool valid_TelephoneNumber(const ASN1_String& tn);
```
reneme, line 522: No need to expose this here. Just a free-standing helper function in the .cpp file should be enough.
Namely:

```C++
namespace {

bool is_valid_telephone_number(...);

}
```

src/tests/unit_x509.cpp
```diff
@@ -1514,6 +1514,54 @@ Test::Result test_hashes(const Botan::Private_Key& key, const std::string& hash_
...
+   using Botan::Cert_Extension::TNAuthList;
+
+   auto tn_auth_list = cert.v3_extensions().get_extension_object_as<TNAuthList>();
+
+   auto& tn_entries = tn_auth_list->entries();
+
+   result.confirm("cert has TNAuthList extension", tn_auth_list != nullptr, true);
+
+   result.test_eq("spc entry type", tn_entries[0].type() == TNAuthList::Entry::TN_ServiceProviderCode, true);
+   result.test_eq("spc entry data", tn_entries[0].service_provider_code(), "1001");
```
reneme, line 1548: Perhaps also test that the other types throw an error. For instance:

```C++
result.test_throws("SPC is no phone number", [&] { tn_entries[0].telephone_number(); });
```

src/lib/x509/x509_ext.h
```diff
@@ -500,6 +500,76 @@ class OCSP_NoCheck final : public Certificate_Extension {
...
+* RFC8226 Secure Telephone Identity Credentials
+*   https://www.rfc-editor.org/rfc/rfc8226#section-9
+*/
+class BOTAN_PUBLIC_API(3, 5) TNAuthList final : public Certificate_Extension {
+   public:
+      class Entry final : public ASN1_Object {
+         public:
+            /* TNEntry choice values
+             * see: https://datatracker.ietf.org/doc/html/rfc8226#section-9 */
+            enum Type { TN_ServiceProviderCode = 0, TN_TelephoneNumberRange = 1, TN_TelephoneNumber = 2 };
```
reneme, line 515: Nit: I think we can spare the prefix.

reneme (commented):
Just some minor comments now.

src/lib/x509/x509_ext.h
furmur: * renamed to `is_valid_telephone_number`
* moved to the top of the `x509_ext.cpp`

src/tests/unit_x509.cpp
furmur: added throw tests for all 3 accessors

src/lib/x509/x509_ext.h
furmur: removed `TN_` prefix

randombit: Looks like some callers did not get updated

furmur: yep. my bad. fixed

randombit (approved):
LGTM!

I'll leave this open for a bit in case @reneme has any final comments.

reneme (approved):
Thanks! LGTM, too.

PR #4117, opened 2024-06-12
Add TPM2 RNG

atreiber94:
This PR is for #3877 regarding the **RNG integration** of TPM2.  It already includes FFI and Python integration.

For current progress on the entire TPM2 integration beyond RNG check #3877.

There's no need to merge this for 3.5.0 (this API may still possibly be subject to changes within broader TPM2 integration of #3877).

## Setup with `swtpm`

To emulate the TPM2, we use [swtpm](https://github.com/stefanberger/swtpm). The following steps are necessary to setup `swtpm`to work with our implementation:

1. Install [tpm2-tools](https://github.com/tpm2-software/tpm2-tools), [libtss2-dev](https://github.com/tpm2-software/tpm2-tss) and [swtpm](https://github.com/stefanberger/swtpm): `sudo apt install tpm2-tools libtss2-dev swtpm` 
2. Configure Botan `--with-tpm2` and build
4.  `mkdir /tmp/myvtpm`
5. `swtpm socket --tpmstate dir=/tmp/myvtpm --tpm2 --ctrl type=tcp,port=2322 --server type=tcp,port=2321 --flags not-need-init` (terminal has to stay open)
6. `tpm2_startup -c`

By default, Botan's new `TPM2_Context` will use the TPM2-TSS default strategy for identifying the TCTI. To directly use `swtpm` or another TCTI, pass the corresponding TCTI name configuration to `Botan::TPM2_Context::create`.

## Python example (because that's what we need)

```python
import botan3 as botan
from binascii import hexlify

tpm2_ctx = botan.TPM2Context("swtpm")
tpm2_rng = botan.RandomNumberGenerator("tpm2", tpm2_ctx)

print(hexlify(tpm2_rng.get(32)).decode("utf-8"))
```

## ToDos

- [x] CI Integration
- [ ] Documentation


src/lib/prov/tpm2/tpm2_rng.cpp
```diff
@@ -0,0 +1,47 @@
...
+   while(!out.full()) {
+      TPM2B_DIGEST* digest = nullptr;
+      const auto requested_bytes = std::min(sizeof(digest->buffer), out.remaining_capacity());
+      check_tss2_rc("GetRandom",
+                    Esys_GetRandom(m_ctx->get(), ESYS_TR_NONE, ESYS_TR_NONE, ESYS_TR_NONE, requested_bytes, &digest));
+
+      BOTAN_ASSERT_NOMSG(digest->size == requested_bytes);
+      out.append({digest->buffer, digest->size});
+
+      Esys_Free(digest);
```
reneme, line 43: Might be a good idea to use `scoped_cleanup` (or some variant of it) for this. Its unlikely, but if either the assert or the append above throw, the `digest` won't be freed.

src/python/botan3.py
```diff
@@ -2138,3 +2149,21 @@ def zfec_decode(k, n, indexes, inputs):
         c_size_t(k), c_size_t(n), c_indexes, c_inputs, c_size_t(share_size), c_outputs
     )
     return [output.raw for output in outputs]
+
+#
+# TPM2
+#
+class TPM2Context:
+    def __init__(self, tcti_nameconf: str = None) -> None:
+        tcti = c_char_p(0 if not tcti_nameconf else tcti_nameconf.encode("utf-8"))
```
reneme, line 2158: Other locations (for instance the constructor of `RandomNumberGenerator`) use a `_ctype_str()` helper to transform a string object into a (presumable) `const char*`. Perhaps we should use that here as well? Let's look into it once more...

src/python/botan3.py
```diff
@@ -2138,3 +2149,21 @@ def zfec_decode(k, n, indexes, inputs):
...
+
+        self.__obj = c_void_p(0)
+        rc = _DLL.botan_tpm2_ctx_init(byref(self.__obj), tcti)
+        if rc == -40: # 'Not Implemented'
```
reneme, line 2162: Any chance we could pull this magic number from the enum? 😅 
(there are other places that would benefit from it)

reneme: Mhh, there seems to be some secret sauce around the translation of parameter/result types from C to Python and back. This blog post provides a little inspiration: https://v4.chriskrycho.com/2015/ctypes-structures-and-dll-exports.html

Perhaps we could actually mirror the error-enum into `botan3.py`. That's a manual duplication, but should be okay, given that this enum probably doesn't change all the time. And then we could make CDLL do the translation for us.

Any of this should certainly not go into this pull request, though.

src/lib/ffi/ffi_tpm2.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+      auto tcti = [=]() -> std::optional<std::string> {
+         if(tcti_nameconf == nullptr) {
+            return {};
+         } else {
+            return std::string(tcti_nameconf);
+         }
+      }();
+
+      ctx->ctx = Botan::TPM2_Context::create(std::move(tcti));
```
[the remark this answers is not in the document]
atreiber94: Good point. I don't think reusing `BOTAN_FFI_ERROR_NOT_IMPLEMENTED` is all that helpful since the function would be implemented and TPM2 usage desired in that case (as the build is configured with `--with-tpm2`).

The exception that would be thrown here is `TPM2_Error` with error type `ErrorType::TPMError`. I added a corresponding FFI error type such that the caller will now get that error translated within `ffi_guard_thunk` instead of `BOTAN_FFI_ERROR_EXCEPTION_THROWN`. That way, one can distinguish between a runtime TPM error when configured with TPM and a not implemented error when TPM is not configured.

atreiber94:
Applied review suggestions except the parts that require re-structuring `tpm2.h`, which I will pick up later on. As mentioned, we do not intend to merge this soon (not for 3.5.0)

doc/api_ref/ffi.rst
```diff
@@ -177,6 +177,10 @@ The following enum values are defined in the FFI header:
    calling :cpp:func:`botan_hash_destroy` on a ``botan_rng_t`` object will cause
    this error.
 
+.. cpp:enumerator:: BOTAN_FFI_TPM_ERROR = -78
```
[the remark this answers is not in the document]
atreiber94: See https://github.com/randombit/botan/pull/4117#discussion_r1638348006 - this FFI error translates the `ErrorType::TPMError`, which is also the error type for `TPM2_Error`.  Or should we just introduce a new TPM2 `ErrorType` and keep them separated entirely?

src/lib/prov/tpm2/tpm2.h
```diff
@@ -0,0 +1,65 @@
...
+
+      TPM2_Context(const TPM2_Context&) = delete;
+      TPM2_Context(TPM2_Context&& ctx) noexcept = default;
+      ~TPM2_Context();
+
+      TPM2_Context& operator=(const TPM2_Context&) = delete;
+      TPM2_Context& operator=(TPM2_Context&& ctx) noexcept = default;
+
+      /// @return an ESYS_CONTEXT* for use in other TPM2 functions.
+      void* get();
```
randombit, line 53: We’ve (starting very recently) started using a `_` prefix for functions which are publically accessible but that users should not touch. This seems like a good candidate for that.

reneme: Frankly, I'm not so sure. Applications might want to access the wrapped context object; to do some custom configuration, for instance. I think that's a worthwhile use case, no?

Of course, one could argue that this is still possible and the `_` just clearly communicates that a user is voiding their warranty with this. 😅

randombit: I guess I'm ok with that argument, but let's a least make this function name a little longer, eg `inner_context_object`

src/lib/prov/tpm2/tpm2_rng.cpp
```diff
@@ -0,0 +1,54 @@
...
+   BufferSlicer in(input);
+
+   while(!in.empty()) {
+      TPM2B_SENSITIVE_DATA data;
+      data.size = std::min(in.remaining(), MAX_STIR_RANDOM_SIZE);
+      in.copy_into({data.buffer, data.size});
+
+      check_tss2_rc(
+         "StirRandom",
+         Esys_StirRandom(static_cast<ESYS_CONTEXT*>(m_ctx->get()), ESYS_TR_NONE, ESYS_TR_NONE, ESYS_TR_NONE, &data));
```
randombit, line 31: Might be worth adding a helper in the internal util function that both performs the cast and also checks pointer validity.

```
inline ESYS_CONTEXT* inner(std::shared_ptr<TPM2_Context> ctx) {
   BOTAN_ASSERT_NOMSG(ctx != nullptr);
   void* inner = ctx->get();
   BOTAN_ASSERT_NOMSG(inner != nullptr);
   return static_cast<ESYS_CONTEXT*>(inner);
}
```

then

`Esys_StirRandom(inner(m_ctx), ESYS_TR_NONE, ...)`

src/lib/ffi/ffi.h
```diff
@@ -259,6 +260,33 @@ BOTAN_FFI_EXPORT(2, 3) int botan_base64_decode(const char* base64_str, size_t in
 */
 typedef struct botan_rng_struct* botan_rng_t;
 
+/**
```
randombit, line 263: These declarations are in the middle of the RNG declarations. They should all be moved, probably down to the very end of the file.

src/lib/ffi/ffi_tpm2.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+int botan_tpm2_ctx_init(botan_tpm2_ctx_t* ctx_out, const char* tcti_nameconf) {
+#if defined(BOTAN_HAS_TPM2)
+   return ffi_guard_thunk(__func__, [=]() -> int {
+      if(ctx_out == nullptr) {
+         return BOTAN_FFI_ERROR_NULL_POINTER;
+      }
+      auto ctx = std::make_unique<botan_tpm2_ctx_wrapper>();
+
+      auto tcti = [=]() -> std::optional<std::string> {
```
randombit, line 42: Is the empty string (potentially) meaningful here? Currently we map nullptr to none, and anything else including the empty string to some. Just wondering if we should also map empty to nullopt here.

[Disclaimer I have no idea what this string is doing]

src/tests/test_tpm2.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+std::vector<Test::Result> test_tpm2_rng() {
+   std::string use_tpm2_emulator = "swtpm";
+   auto ctx = Botan::TPM2_Context::create(std::move(use_tpm2_emulator));
+
+   auto rng = Botan::TPM2_RNG(ctx);
+
+   return {
+      Botan_Tests::CHECK("Basic functionalities",
```
randombit, line 41: No need for `Botan_Tests::` here since we are in that namespace

src/lib/ffi/ffi_tpm2.cpp
atreiber94: Both a nullptr and the empty string indicate the default Transmission Interface. Mapping empty to nullopt here would be functionally equivalent - however in TPM2_Context the none gets translated back to a nullptr given to the TSS so I guess with the current way we keep the translations more consistent.

src/tests/test_tpm2.cpp
```diff
@@ -0,0 +1,85 @@
...
+#if defined(BOTAN_HAS_TPM2)
+   #include <botan/tpm2_rng.h>
+#endif
+
+namespace Botan_Tests {
+
+#if defined(BOTAN_HAS_TPM2)
+namespace {
+
+bool not_zero_64(std::span<const uint8_t> in) {
```
randombit, line 21: I guess this fails randomly 1/2**64 of the time?

randombit:
LGTM. Needs some history cleanup

thillux:
Do you plan to fetch random numbers over an encrypted (e.g. HMAC session) in the future? Linux kernel does so from 6.10 on.

reneme:
Yes! This PR is somewhat outdated. We're currently working on a more comprehensive wrapper, also for TPM hosted keys.

PR #4118, opened 2024-06-12
PBES2 improvements and fix PBES2-related scrypt bug

randombit:


src/lib/pbkdf/scrypt/scrypt.cpp
```diff
@@ -83,18 +83,20 @@ std::unique_ptr<PasswordHash> Scrypt_Family::tune(size_t output_length,
 
    uint64_t est_nsec = measured_time;
 
-   // First move increase r by 8x if possible
+   // In below code we invoke scrypt_memory_usage with p == 0 as p contributes
+   // (very slightly) to memory consumption, but N is the driving factor.
+   // Including p leads to using an N half as large as what the user would expect.
 
-   if(max_memory_usage == 0 || scrypt_memory_usage(N, r * 8, p) <= max_memory_usage) {
+   // First move increase r by 8x if possible
```
reneme, line 90:
```suggestion
   // First increase r by 8x if possible
```
... I guess.

randombit: Fixed that and also a now-incorrect comment regarding the removed addition of 512 to the memory bound

PR #4119, opened 2024-06-13
Ounsworth KEM Combiner

FAlbertDev:
# Ounsworth KEM Combiner
This KEM Combiner is based on [this draft specification](https://github.com/EntrustCorporation/draft-ounsworth-cfrg-kem-combiners/blob/475ff53eb8fb7213f6e5ab26dd23e5dc3203f7fa/draft-ounsworth-cfrg-kem-combiners.txt). Unlike X-Wing, the Ounsworth combiner (named after its main author) offers great flexibility by supporting any KEM combination. X-Wing, on the other hand, only specifies a single KEM combination (ML-KEM-768 with X25519). This flexibility is beneficial when using BSI-recommended algorithms such as FrodoKEM and brainpool curves.

Since the current draft is still in its early stages and lacks test vectors, the specification may change in the future. Additionally, I may have misinterpreted certain parts of the specification. Therefore, I have marked the implementation as experimental using the `BOTAN_UNSTABLE_API` macro.

## Structure of the Implementation
The Ounsworth KEM Combiner is implemented as a regular KEM. It combines sub-KEMs with different configurations using an `Ounsworth::Mode` object. This mode consists of a list of sub-KEMs and an enum value determining which of the four specified KDFs is used. To provide flexibility in defining sub-KEMs, I want to create an interface that allows users to define and use their own KEMs in the Ounsworth Combiner. This is useful in one of our research projects where experimental PKCS#11-based instances must be combined with the Ounsworth Combiner.

However, I also understand that most applications have more straightforward requirements. To accommodate this, I provide a straightforward constructor for predefined algorithms such as Kyber, FrodoKEM for PQC, and X25519, X448, and ECDH for the classic choice. These algorithms can be selected using an enum value.

Some limitations with the current public/private key interface make generalization a bit messy. For example, to concatenate public keys without any encoding, we need to know the length of the public key in advance. The same goes for private keys. To address this, we may want to add methods like `pk_byte_length()` and `sk_byte_length()` for the private and public key interfaces in the future. For now, I worked around this limitation using algorithm-dependent methods.

I decided to compose the Ounsworth key using the **raw** public and private key bytes. Although the encoding is not specified in the draft, I believe this is the best approach. However, in most cases, this prevents me from using Botan's generic interfaces, which adds further complexity to the implementation (see `ounsworth_mode.cpp`).

## Testing
Since no test vectors are available yet, I used some tests covering most of the draft's requirements.

- The underlying KDF is the one-step KDM defined in [NIST.SP.800-56C](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Cr2.pdf) Section 4.1. Although I couldn't find exhaustive open-source implementations, I did find test vectors from ACVP. I used these test vectors to perform a KAT test on the KDF.

- To ensure that the predefined KEMs are configured correctly, I conducted a roundtrip test using all predefined algorithms simultaneously.

- The only potential source of error remaining is the creation of the KDF inputs, specifically the composition of various fields and lengths to form the KDF input. To test this, I created a handcrafted test vector that should represent other combinations.

- I also tested the key interfaces using the generic keygen tests.

## General Remarks
The draft is still in its early stages, so I had to make some assumptions. For instance, the structure of the private and public keys is not specified in the draft. To address this, I documented my decisions in the doxygen documentation of the private key class.

I've already defined some OIDs in Botan's private arc to make it easier to use in protocols. These OIDs use sensible combinations for different key strengths. I've even included combinations with Frodo and Brainpool curves, which the BSI recommends. I crafted these combinations myself, but feel free to suggest other options.

I welcome any feedback on the implementation.

## Pull request dependencies
- https://github.com/randombit/botan/pull/4067


randombit:
Initial comment is that the KDFs should be split out into their own PR, and should be actual `KDF`s

The hash based one in already implemented as SP800-56A, with the apparent extension required that 56A hash KDF doesn't support a salt (LOL LMAO WTAF) and 56C does.

The KMAC one should be implemented in a dedicated way - that is to say requiring and using only KMAC rather than being theoretically generic over the mechanism.

FAlbertDev:
Ops, you are right. I thought I'd looked into the existing KDFs and decided they didn't match. I'll create a separate PR where I address this 👍

reneme:
With #4126 merged, please rebase to fix the macOS build.

FAlbertDev:
Rebased: ounsworth_kem_combiner->kem-combiner-abstraction->master

FAlbertDev:
I've rebased to master and did some major refactoring. Using a single _Mode_ is too inflexible. Now, applications can either pass lists of public and private keys directly or define callbacks to generate/load private/public keys directly. These callbacks are no longer collected in a single _Ounsworth\_Mode_but are used individually for the respective constructors. (I updated the PR description above)

PR #4121, opened 2024-06-13
One-Step Key Derivation Method with KMAC

FAlbertDev:
As mentioned in PR #4119, this pull request introduces the One-Step Key Derivation Method (KDM) of [SP800-56Cr2](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Cr2.pdf) Section 4 using KMAC.

Previously, the One-Step KDM was defined in [SP800-56Ar2](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar2.pdf). However, with the revision [SP800-56Ar3](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf), it has been moved to [SP800-56Cr2](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Cr2.pdf). To reflect this change, I have updated comments and internal interfaces accordingly. The public API and external interfaces, including module names, remain unchanged.

Additionally, this pull request includes a KMAC option for the one-step Key Derivation Function (KDF), which is utilized by the Ounsworth Key Encapsulation Mechanism (KEM) combiner. Also, I added the test vectors I got for the Ounsworth PR. These include tests for the KMAC option and some for SHA-3 instances, which were not yet covered.


src/tests/data/kdf/sp800_56a.vec
```diff
@@ -2411,3 +2414,342 @@ Salt =
...
...7813555FCB977B032B77C609BF33A372FE9CB2F200E1124F3E11277BFFD55ED4FB5334F6128369083E34BEBED737751666CCF8F4D295A7AC0A68E3359FB937EA369E40344A093FF6918488A78D1DB72146C73437BF94733C877E00527AFA8CC6DC4CDB8F59BBDE9A04B78945D22E2B0BBA5D2F2C7F1437122D99F029A444AE1ACA5AD6B928717F76F4B34BF6A6A1D677A8A59629145720B2060400D831D34E50AE2FE7A9A6D3B051864484A0C63D162E7121FCF3945D4AB1037C8CD9C305BA78541233BDCF4899A4992B865E09242F92DDC3DDEC4BA262E2EDB6F14078497C0DA38DC4F5CFF3F24F0E4896F73F3768F5BC04233F125D06180803B0C198A8EC2F22C596721AEB970003F04A52BB6AE69D788AE864799FD33900DFE167F3F540589A526125D4E180A4699CDFE59706F7E96BA4B7D800878184EB6A539A90019CE3D6E259B35451413063E8ECFB24BFA45479726890D9163AF405B6771612E38D569804A43C5B715BD43491EF3B7C873A6381667D1E5735E77B1EB424AF2575C2522468825F2225BAF790C0C5C19259A92A669D8CD435B2BF1A63C7611568EC94FA52829425B1A0010D3AB3D0164F4B4380CA8D9CBA6F14D4B88D34002332F4AA61D7150CD5D973B70258FCE80A6A4D4F64EB940A134D6C0F5E576D9868A3590923BE3B639EAF2E15D4A0EA267F732C26269CA37D3C17699C3BB4BC04EE4758EC7E7B5B379173EB0869D1227DEAA7C31A5B5CDFC190D8847EF6CCD78AB797E6F61ED4CD438D8882055E3455C1965F69123F739CD43D76C14CBEE9C01478493CBE92B8DC54B9038EF8301D3752244368B6E04FD4F35678AFFB29
```
randombit, line 2435: Some of these test vectors are really long and I don't see much justification for them considering nothing about the KDF varies on the input lengths

src/lib/kdf/sp800_56a/sp800_56c_one_step.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+namespace {
+/**
+ * @brief One-Step Key Derivation as defined in SP800-56Cr2 Section 4
+ */
+void kdm_internal(std::span<uint8_t> output_buffer,
+                  std::span<const uint8_t> z,
+                  std::span<const uint8_t> fixed_info,
+                  Buffered_Computation& h,
+                  const std::function<void(Buffered_Computation*)>& reset_h_callback) {
```
randombit, line 30: Seems better to template this over a `T` to avoid a `dynamic_cast` for each iteration of the loop.

Also `Buffered_Computation` as a base class is anyway deprecated so we'd have to rewrite this when it's removed.

src/lib/kdf/sp800_56a/sp800_56c_one_step.cpp
```diff
@@ -0,0 +1,192 @@
...
+   kdm_internal({key, key_len}, {secret, secret_len}, {label, label_len}, *m_mac, [&](Buffered_Computation* kdf) {
+      MessageAuthenticationCode* kdf_mac = dynamic_cast<MessageAuthenticationCode*>(kdf);
+      BOTAN_ASSERT_NONNULL(kdf_mac);
+      kdf_mac->clear();
+      // 4.1 Option 2 and 3 - An implementation dependent byte string, salt,
+      //     whose (non-null) value may be optionally provided in
+      //     OtherInput, serves as the HMAC#/KMAC# key ..
+
+      // SP 800-56A specifies if the salt is empty then a block of zeros
+      // equal to the hash's underlying block size are used. However this
```
randombit, line 131: This is an issue in the original comment but we should fix it while we're here

> However *for HMAC* this is equivalent

Since that's not generally true for eg KMAC or whatever

src/lib/kdf/sp800_56a/sp800_56c_one_step.cpp
```diff
@@ -0,0 +1,192 @@
...
+      // is equivalent to setting a zero-length key, so the same call
+      // works for either case.
+      kdf_mac->set_key(std::span{salt, salt_len});
```
randombit, line 168: Duplicated call to `set_key`

src/lib/kdf/sp800_56a/sp800_56c_one_step.h
```diff
@@ -98,12 +100,73 @@ class SP800_56A_HMAC final : public KDF {
...
+      * @param key derived keying material K_M
+      * @param key_len the desired output length in bytes
+      * @param secret shared secret Z
+      * @param secret_len size of Z in bytes
+      * @param salt ignored
+      * @param salt_len ignored
+      * @param label label for the expansion step
+      * @param label_len size of label in bytes
+      *
+      * @throws Invalid_Argument key_len > 2^32 or MAC is not a HMAC
```
randombit, line 129: Incorrect comment re HMAC

src/lib/kdf/sp800_56a/sp800_56c_one_step.h
```diff
@@ -98,12 +100,73 @@ class SP800_56A_HMAC final : public KDF {
...
+/**
+ * NIST SP800-56Cr2 One-Step KDF using KMAC-128
+ */
+class SP800_56C_One_Step_KMAC128 final : public SP800_56A_One_Step_KMAC_Abstract {
+   public:
+      std::string name() const override { return "SP800-56A(KMAC-128)"; }
+
+      std::unique_ptr<KDF> new_object() const override { return std::make_unique<SP800_56C_One_Step_KMAC128>(); }
+
+   protected:
```
randombit, line 153: Why not `private`?

src/lib/kdf/sp800_56a/sp800_56c_one_step.h
```diff
@@ -98,12 +100,73 @@ class SP800_56A_HMAC final : public KDF {
...
+/**
+ * NIST SP800-56Cr2 One-Step KDF using KMAC-256
+ */
+class SP800_56C_One_Step_KMAC256 final : public SP800_56A_One_Step_KMAC_Abstract {
+   public:
+      std::string name() const override { return "SP800-56A(KMAC-256)"; }
+
+      std::unique_ptr<KDF> new_object() const override { return std::make_unique<SP800_56C_One_Step_KMAC256>(); }
+
+   protected:
```
randombit, line 166: `private`?

src/lib/kdf/sp800_56c/sp800_56c_two_step.cpp
```diff
@@ -0,0 +1,41 @@
+/*
```
randombit, line 1: Why is `git`s rename logic so stupid

randombit (commented):
Left some comments but overall seems fine

src/lib/kdf/sp800_56a/sp800_56c_one_step.cpp
FAlbertDev: I didn't know Buffered_Computation is deprecated. I'll use a template.

src/lib/kdf/sp800_56a/sp800_56c_one_step.h
FAlbertDev: This does not even check if MAC is an HMAC. This check is performed in the constructor. The max key length is also wrong 😅

src/tests/data/kdf/sp800_56a.vec
FAlbertDev: Agreed. I removed all vectors with excessive lengths!

FAlbertDev:
Thanks, @randombit, for your input! It seems I left some copy-past issues in the new implementation. Also, I did not correctly handle default_salts for KMAC. The doxygen comments were also deprecated in various places.

I addressed your suggestions and the other mentioned issues in the latest commit.

randombit (approved):

PR #4122, opened 2024-06-13
FIX: Better scaling for botan_cipher_get_update_granularity()

### LINKED ISSUE #4090: FFI: botan_cipher_get_update_granularity() returns 17

reneme:
This is an alternative suggestion to address #4090. Essentially, it keeps the constraint that `botan_cipher_get_update_granularity()` should always return a size greater than the `minimal_final_size()` of the underlying cipher mode. But, it tries to produce more sensible outputs that actually honor the documented behavior ("ideal granularity" being a multiple of this).

_To the best of my understanding, this is needed to handle the authentication tag of AEADs without special treatment of the final API buffer. But frankly, I don't fully understand the (historic) reasoning about it._

With the changes that came in #3951 this extra complexity is not needed anymore. At the expense of users handling the `BOTAN_FFI_ERROR_INSUFFICIENT_BUFFER_SPACE` error code properly. Therefore, I strongly suggest we implement the (vaguely) API changing suggestion of #4094 in the next major release. (How do we ensure that this won't be forgotten?)

Closes #4090.

### DIFF
--- a/src/lib/ffi/ffi_cipher.cpp
+++ b/src/lib/ffi/ffi_cipher.cpp
@@ -9,6 +9,7 @@
 #include <botan/aead.h>
 #include <botan/internal/ffi_util.h>
 #include <botan/internal/stl_util.h>
+#include <botan/internal/bit_ops.h>
 
 extern "C" {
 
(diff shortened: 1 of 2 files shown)


randombit:
> But frankly, I don't fully understand the (historic) reasoning about it.

You know we are kind of in the same boat on that one. *But* it is how the this interface has behaved in the past. And generally I put a much higher standard on dont-change-behavior for FFI vs everything else. For one because the FFI is our long lived API *and* ABI compatible interface. Whereas we happily break ABI for the C++ with every release, it's quite possible for the same Python script to link to and run against anything between Botan 2.0.0 and Botan 3.4.0 with no changes to the source and no change in behavior.

randombit (approved):
I am concerned this sequence of changes will still end up breaking callers but I guess we'll just have to wait and see. I checked the Rust bindings and the tests still pass there anyway. Thanks for adding the tests.

randombit:
> How do we ensure that this won't be forgotten?

I've been thinking about starting a doc with "checklist of things to not forget in 3->4", we can put it there. Might work on this over the weekend.

PR #4123, opened 2024-06-14
clang-format-17 everything

reneme:
This is a follow-up for #4100 that missed a few spots. Also, explicitly sets `clang-format-17` in the VS Code configuration.

### DIFF
--- a/src/scripts/ci/setup_gh_actions.sh
+++ b/src/scripts/ci/setup_gh_actions.sh
@@ -95,8 +95,6 @@ if type -p "apt-get"; then
     elif [ "$TARGET" = "docs" ]; then
         sudo apt-get -qq install doxygen python3-docutils python3-sphinx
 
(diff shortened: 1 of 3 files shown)


src/scripts/ci/setup_gh_actions.sh
```diff
@@ -95,8 +95,6 @@ if type -p "apt-get"; then
     elif [ "$TARGET" = "docs" ]; then
         sudo apt-get -qq install doxygen python3-docutils python3-sphinx
 
-    elif [ "$TARGET" = "format" ]; then
-        sudo apt-get -qq install clang-format-15
```
reneme, line 99: FYI: GH Actions' Ubuntu 24.04 image brings `clang-format-17` by default. No need to explicitly install any clang-format anymore.

randombit (approved):

PR #4126, opened 2024-06-14
FIX: store_be(size_t(42)) would not compile on Xcode

reneme:
Xcode defines uint64_t as "unsigned long long" and size_t as "unsigned long int". Both are 64bit types, but the compiler failed to call reverse_bytes() for size_t regardless. It claimed the call is ambiguous. This was initially found by @FAlbertDev in #4119.

The workaround makes `reverse_bytes()` an (awful) template that just assumes to get a `std::unsigned_integral T` and disambiguates on `sizeof(T)` using if constexpr. I'm certainly open to better ideas. But then again: C++23 will obsolete this anyway.

### DIFF
--- a/src/lib/utils/bswap.h
+++ b/src/lib/utils/bswap.h
@@ -2,6 +2,7 @@
 * Byte Swapping Operations
 * (C) 1999-2011,2018 Jack Lloyd
 * (C) 2007 Yves Jerschow
+* (C) 2024 René Meusel - Rohde & Schwarz Cybersecurity
 *
 * TODO: C++23: replace this entire implementation with std::byteswap
 *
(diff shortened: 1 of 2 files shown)


FAlbertDev (approved):
Thanks! XCode hates this one weird trick.

randombit:
I'm confused why you would ever want to do this. The number of bytes written will depend on the size of size_t, which means it's not really useful for serialization. What am I missing?

reneme:
This is not supposed to change any behavior. It's just working around Xcode's compiler not being able to dispatch the overload of `reverse_bytes`.

See here for a build failure on macOS: https://github.com/randombit/botan/actions/runs/9515856655/job/26230969296?pr=4119

randombit:
> This is not supposed to change any behavior.

Sure but my point was that we had not hit this so far because saying `store_be(some_size_t)` isn't really something that makes sense because the number of bytes emitted depends on the size of `size_t`, and so isn't really usable for anything. Serialize it, hash it, whatever, the difference in byte count will mean you have different output.

I looked at the specific code in  `ounsworth.cpp`  that triggered this and I guess `store_be` with a `size_t` is at least usable in that case, because you're post-processing it to truncate unused bytes. This feels like a real edge case to me (wtf is up with these overly complicated variable length encodings) but meh.

randombit (approved):

reneme:
> Sure but my point was that we had not hit this so far because saying `store_be(some_size_t)` isn't really something that makes sense because the number of bytes emitted depends on the size of `size_t`, and so isn't really usable for anything.

True! I didn't even think of that. But that answers: "How did that ever work?" 🤣

Issue #4125, opened 2024-06-14
[botan2] Intermittent assertions in `dh_invalid` test case on ppc64le


mrc0mmand:
Hey!

In the past couple of weeks I started seeing intermittent fails in botan2 package builds in both Fedora Rawhide and Fedora 40. I managed to reproduce them with the latest `release-2` branch as well:

```
# git describe
1.11.30-6444-g068974b9e
# g++ --version
g++ (GCC) 14.1.1 20240607 (Red Hat 14.1.1-5)
Copyright (C) 2024 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

## Note: this is a subset of compiler options Fedora uses that I was able to reproduce the issue with
# CXXFLAGS="-O2 -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -m64 -mcpu=power8 -mtune=power8 -g" python3 ./configure.py --enable-modules bzip2,zlib --cpu=ppc64le --os=linux --cc=gcc
...
# make -j $(nproc)
...

# while LD_LIBRARY_PATH=$PWD/ ./botan-test; do :; done
...
dh_invalid:
/usr/include/c++/14/bits/stl_vector.h:1130: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = unsigned char; _Alloc = Botan::secure_allocator<unsigned char>; reference = unsigned char&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Aborted (core dumped)
```

Full stack trace: https://gist.githubusercontent.com/mrc0mmand/9659e94bb21117a6489d68fc995f9ef6/raw/daae6c587f1a863c9931d301e7e6be20828a8d40/botan2-ppc64le

randombit:
Thanks for the report. So this is only being seen on ppc64le?

I'm going to guess this timeline (last couple of weeks) is associated with a GCC upgrade

randombit:
I think the output is a bit of a red herring

```
dh_invalid:
[...]

PR #4131, opened 2024-06-15
Fix some possible iterator debugging errors in TLS-CBC

randombit:
These have never been seen or reported as affecting 3.x (for some unknown reason), but we know they caused problems in 2.x so might as well address them.

See #3916 #4125 #4130


reneme (approved):
LGTM. In the long run it would probably make sense to convert this implementation to std::spans to avoid such things from the get-go.

PR #4142, opened 2024-06-22
Pcurves fixes and additions

randombit:


src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -120,7 +120,7 @@ class IntMod final {
 
       constexpr CT::Choice is_even() const {
          auto v = Rep::from_rep(m_val);
-         return CT::Choice::from_int(v[0] & 0x01);
+         return CT::Choice::from_int(0x01 ^ (v[0] & 0x01));
```
reneme, line 123: @randombit Isn't the value barrier coming too late, here? I feel like the compiler might come up with branches while calculating `0x01 ^ (v[0] & 0x01)`, no?

Sorry for digging up this merged pull request.

PR #4143, opened 2024-06-22
Bridge pcurves into the main elliptic curve arithmetic

randombit:
This results in roughly a 2x-3x speedup for all elliptic curve algorithms, when using a supported curve.


reneme:
Without a full picture, yet: I think it would make sense to allow disabling the BN-based ECC implementation at compile time.

randombit:
> I think it would make sense to allow disabling the BN-based ECC implementation at compile time.

In principle yes I agree. However it's greatly complicated by the fact that we have large swaths of public API which expose various BigInt based ec functionality (`EC_Group::multiply_mod_order`, etc). This is one of those things that I think is best deferred to Botan 4 - once all of these various APIs are removed we have a much better shot at being able to do this cleanly.

reneme:
To be able to get an idea of the big picture, I extended the diagram I started at some point. Feedback appreciated. :)

image
[excalidraw.com](https://excalidraw.com/#json=pg24mSP40E2cfKzxuONBD,P36KYnTJt9BR0_-DGNma0A)

_Note that excalidraw isn't a shared collaboration space. To publish any modifications, one has to re-create a new snapshot via "save as... > shareable link"_

src/lib/pubkey/ec_group/ec_inner_pc.cpp
```diff
@@ -0,0 +1,210 @@
...
+                                               PCurve::PrimeOrderCurve::AffinePoint pt) :
+      m_group(std::move(group)), m_pt(std::move(pt)) {
+   m_bytes = m_pt.serialize<secure_vector<uint8_t>>();
+}
+
+EC_AffinePoint_Data_PC::EC_AffinePoint_Data_PC(std::shared_ptr<const EC_Group_Data> group,
+                                               std::span<const uint8_t> bytes) :
+      m_group(std::move(group)), m_pt(deserialize_pcurve_pt(m_group->pcurve(), bytes)) {
+   m_bytes = m_pt.serialize<secure_vector<uint8_t>>();
+}
```
reneme, line 96: Two things:

1. `m_bytes` could be initialized in the initializer-list as well

```suggestion
EC_AffinePoint_Data_PC::EC_AffinePoint_Data_PC(std::shared_ptr<const EC_Group_Data> group,
                                               PCurve::PrimeOrderCurve::AffinePoint pt) :
      m_group(std::move(group)), m_pt(std::move(pt)), m_bytes(m_pt.serialize<secure_vector<uint8_t>>()) {
}

EC_AffinePoint_Data_PC::EC_AffinePoint_Data_PC(std::shared_ptr<const EC_Group_Data> group,
                                               std::span<const uint8_t> bytes) :
      m_group(std::move(group)), m_pt(deserialize_pcurve_pt(m_group->pcurve(), bytes)), m_bytes(m_pt.serialize<secure_vector<uint8_t>>()) {
}
```

2. Given that `m_bytes` is never written to, perhaps its worthwhile to add `std::span<>`s that centrally refer to the segments of the serialization. That would avoid spreading the knowledge about the structure in `m_bytes` across several methods. Like so:

```C++
EC_AffinePoint_Data_PC::EC_AffinePoint_Data_PC(std::shared_ptr<const EC_Group_Data> group,
                                               PCurve::PrimeOrderCurve::AffinePoint pt) :
      m_group(std::move(group)), m_pt(std::move(pt)), m_bytes(m_pt.serialize<secure_vector<uint8_t>>()),
      m_bytes_x(std::span{m_bytes}.subspan(1, field_element_bytes())),
      m_bytes_y(std::span{m_bytes}.last(field_element_bytes())) {
}
```

src/lib/pubkey/ec_group/ec_inner_pc.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+const EC_AffinePoint_Data_PC& EC_AffinePoint_Data_PC::checked_ref(const EC_AffinePoint_Data& data) {
+   const auto* p = dynamic_cast<const EC_AffinePoint_Data_PC*>(&data);
+   if(!p) {
+      abort();
+      throw Invalid_State("Failed conversion to EC_AffinePoint_Data_PC");
+   }
```
reneme, line 103: `abort()` seems a bit harsh.

src/lib/pubkey/ec_group/ec_inner_pc.cpp
```diff
@@ -0,0 +1,210 @@
...
+   stuffer.append(y_is_odd ? 0x03 : 0x02);
+   this->serialize_x_to(stuffer.next(fe_bytes));
+}
+
+EC_Point EC_AffinePoint_Data_PC::to_legacy_point() const {
+   const size_t fe_bytes = this->field_element_bytes();
+   auto x = BigInt::from_bytes(std::span{m_bytes}.subspan(1, fe_bytes));
+   auto y = BigInt::from_bytes(std::span{m_bytes}.last(fe_bytes));
+
+   return EC_Point(m_group->curve(), std::move(x), std::move(y));
```
reneme, line 163: clang-tidy says: move won't happen, because this constructor takes the `BigInt` objects by `const&`. I think, it shouldn't do that, though. Perhaps change the constructor?

src/lib/pubkey/ec_group/info.txt
```diff
@@ -11,12 +11,14 @@ brief -> "Wrapper for elliptic curve groups"
 <requires>
 asn1
 numbertheory
+pcurves
```
reneme, line 14: Is that really required? If `pcurves` is not compiled, I can still fall back to the bigint implementation also for the pcurves-supported curves, no?

Or do we want to model "no pcurves at compile time" by disabling all `pcurves_*` submodules?

src/lib/pubkey/ec_group/ec_inner_data.cpp
```diff
@@ -0,0 +1,227 @@
...
+   }
+
+   if(m_pcurve) {
+      if(auto s = m_pcurve->deserialize_scalar(bytes)) {
+         return std::make_unique<EC_Scalar_Data_PC>(shared_from_this(), *s);
+      } else {
+         return nullptr;
+      }
+   } else {
+      BigInt r(bytes.data(), bytes.size());
```
reneme, line 152:
```suggestion
      BigInt r(bytes);
```

src/lib/pubkey/ec_group/ec_inner_data.cpp
```diff
@@ -0,0 +1,227 @@
...
+}
+
+std::unique_ptr<EC_Scalar_Data> EC_Group_Data::scalar_from_bytes_mod_order(std::span<const uint8_t> bytes) const {
+   BOTAN_ARG_CHECK(bytes.size() <= 2 * order_bytes(), "Input too large");
+
+   if(m_pcurve) {
+      if(auto s = m_pcurve->scalar_from_wide_bytes(bytes)) {
+         return std::make_unique<EC_Scalar_Data_PC>(shared_from_this(), std::move(*s));
+      } else {
+         throw Invalid_Argument("Failed to complete reduction of scalar bytes");
```
reneme, line 77: That's somewhat inconsistent. Many other methods that return a unique_ptr simply return `nullptr` if something fails but don't throw. Frankly, I'd prefer things to throw over passing a `nullptr` that I have to remember to check for.

src/lib/pubkey/ec_group/ec_inner_pc.cpp
randombit: Oops, was for debugging :see_no_evil:

src/lib/pubkey/ec_group/ec_inner_data.h
```diff
@@ -0,0 +1,247 @@
...
+      // Returns nullptr if g*x + h*y was point at infinity
+      virtual std::unique_ptr<EC_AffinePoint_Data> mul2_vartime(const EC_Scalar_Data& x,
+                                                                const EC_Scalar_Data& y) const = 0;
+
+      // Returns nullptr if g*x + h*y was point at infinity
+      virtual std::unique_ptr<EC_Scalar_Data> mul2_vartime_x_mod_order(const EC_Scalar_Data& x,
+                                                                       const EC_Scalar_Data& y) const = 0;
+};
+
+class EC_Group_Data final : public std::enable_shared_from_this<EC_Group_Data> {
```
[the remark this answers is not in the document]
randombit: That would probably be cleaner. Main issue is that even with pcurves enabled we still have to support the full set of BigInt based APIs so we're still carrying around all of the same fields. I guess we could have the pcurves `EC_Group_Data` derive from the generic one to avoid too much duplication ...

src/lib/pubkey/ec_group/info.txt
randombit: Not really required in that if pcurves wasn't available at all we could just fall back to the generic code without issue. I'm just trying to avoid too much combinatoric explosion (more than is already in place anyway ...) with yet another module / yet more ifdef. IDK. I guess there are probably users who would prefer just using the generic code for size reasons.

src/lib/pubkey/ec_group/ec_inner_data.cpp
randombit: The behavior between pcurves and the generic code are somewhat inconsistent on this - generic always succeeds no matter how big the input is. Maybe the answer here is to also introduce a (somewhat artificial) limit for the generic case so they behave the same.

randombit:
Comparing `ECDSA` vs `ECDSA-pcurves` (with RFC 6979 disabled) it appears the overhead of the various abstraction layers is betwee 2K and 8K cycles, which is not horrible in this context.

src/lib/pubkey/ec_group/info.txt
randombit: #4212 allows the user to disable all curves by disabling `pcurves_impl` which removes all of the implementations while still leaving the top level interfaces in place.

src/lib/pubkey/ec_group/ec_inner_data.cpp
randombit: Fixed

src/lib/pubkey/ec_group/ec_inner_pc.cpp
```diff
@@ -0,0 +1,210 @@
...
+void EC_AffinePoint_Data_PC::serialize_xy_to(std::span<uint8_t> bytes) const {
+   const size_t fe_bytes = this->field_element_bytes();
+   BOTAN_ARG_CHECK(bytes.size() == 2 * fe_bytes, "Invalid output size");
+   copy_mem(bytes, std::span{m_bytes}.last(2 * fe_bytes));
+}
+
+void EC_AffinePoint_Data_PC::serialize_compressed_to(std::span<uint8_t> bytes) const {
+   const size_t fe_bytes = this->field_element_bytes();
+   BOTAN_ARG_CHECK(bytes.size() == 1 + fe_bytes, "Invalid output size");
+   const bool y_is_odd = (m_bytes[m_bytes.size() - 1] & 0x01) == 0x01;
```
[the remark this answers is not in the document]
randombit: Done

src/lib/pubkey/ec_group/ec_inner_data.cpp
randombit: I changed this so that all failures return nullptr. I also added doc comments indicating which functions can be expected to fail vs which normally always succeed

src/lib/pubkey/ec_group/ec_inner_pc.cpp
randombit: Both done

src/lib/pubkey/ec_group/ec_inner_pc.cpp
randombit: Updated the constructor

randombit:
@reneme Ready for another look

src/lib/pubkey/ec_group/ec_inner_data.cpp
```diff
@@ -75,63 +81,122 @@ void EC_Group_Data::set_oid(const OID& oid) {
...
    try {
-      auto pt = Botan::OS2ECP(bytes.data(), bytes.size(), curve());
-      return std::make_unique<EC_AffinePoint_Data_BN>(shared_from_this(), std::move(pt));
+      if(m_pcurve) {
+         auto pt = m_pcurve->deserialize_point(bytes);
+         if(pt) {
+            return std::make_unique<EC_AffinePoint_Data_PC>(shared_from_this(), std::move(*pt));
+         } else {
+            return nullptr;
+         }
```
reneme, line 195: For consistency with other methods:

```suggestion
         if(auto pt = m_pcurve->deserialize_point(bytes)) {
            return std::make_unique<EC_AffinePoint_Data_PC>(shared_from_this(), std::move(*pt));
         } else {
            return nullptr;
         }
```

src/lib/pubkey/ec_group/ec_inner_data.cpp
```diff
@@ -75,63 +81,122 @@ void EC_Group_Data::set_oid(const OID& oid) {
...
+      return std::make_unique<EC_Scalar_Data_BN>(shared_from_this(), BigInt::one());
+   }
 }
 
 std::unique_ptr<EC_Scalar_Data> EC_Group_Data::scalar_from_bigint(const BigInt& bn) const {
-   // Assumed to have been already checked as in range
-   return std::make_unique<EC_Scalar_Data_BN>(shared_from_this(), bn);
+   if(bn <= 0 || bn >= m_order) {
+      return {};
+   }
```
reneme, line 136: The dcoumentation of this method claims:

> This returns a value only if bn is in [0,n) where n is the group order. Otherwise it returns nullptr.

I.e. 0 would be an acceptable input.

src/lib/pubkey/ec_group/ec_inner_data.cpp
```diff
@@ -75,63 +81,122 @@ void EC_Group_Data::set_oid(const OID& oid) {
...
+      }
+   } else {
+      BigInt r(bytes);
 
-   if(r.is_zero() || r >= m_order) {
-      return nullptr;
-   }
+      if(r.is_zero() || r >= m_order) {
+         return nullptr;
+      }
```
reneme, line 181: Documentation says:

> an integer that is at most n, where n is the group order. [...] It also rejects an all-zero input.

I'm guessing the documentation should be "lower than the group order".

src/lib/pubkey/ec_group/ec_inner_pc.cpp
```diff
@@ -0,0 +1,209 @@
...
+   m_group->pcurve().serialize_scalar(bytes, m_v);
+}
+
+EC_AffinePoint_Data_PC::EC_AffinePoint_Data_PC(std::shared_ptr<const EC_Group_Data> group,
+                                               PCurve::PrimeOrderCurve::AffinePoint pt) :
+      m_group(std::move(group)),
+      m_pt(std::move(pt)),
+      m_bytes(m_pt.serialize<secure_vector<uint8_t>>()),
+      m_x_bytes(std::span{m_bytes}.subspan(1, field_element_bytes())),
+      m_y_bytes(std::span{m_bytes}.last(field_element_bytes())) {}
```
reneme, line 92: Perhaps add a `BOTAN_ASSERT` (potentially `BOTAN_DEBUG_ASSERT`) to the constructor's body to check that `m_bytes` contains the expected number of bytes (`1+2*field_element_bytes()`). If, for any reason, `m_pt.serialize()` didn't produce the expected number of bytes, the spans into that buffer would potentially point to invalid memory. Same in the constructor below.

reneme: Alternatively, one might explicitly allocate `m_bytes` in the constructor's body (as `m_bytes.resize(2*field_element_bytes()+1)`) and then use `m_pt.serialize_to(m_bytes)`. That would also make sure that the spans aren't pointing into the abyss.

src/lib/pubkey/ec_group/ec_inner_pc.cpp
```diff
@@ -0,0 +1,209 @@
...
+}
+
+size_t EC_AffinePoint_Data_PC::field_element_bytes() const {
+   return m_group->pcurve().field_element_bytes();
+}
+
+void EC_AffinePoint_Data_PC::serialize_x_to(std::span<uint8_t> bytes) const {
+   const size_t fe_bytes = this->field_element_bytes();
+   BOTAN_ARG_CHECK(bytes.size() == fe_bytes, "Invalid output size");
+   copy_mem(bytes, m_x_bytes);
```
reneme, line 136: Nit: `copy_mem` does that size check internally, same in some methods that follow. Obviously, with a less clear error message.
Please bear with me if I did point that out before and you decided against it. 😓

src/lib/pubkey/ec_group/ec_inner_pc.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+   BufferStuffer stuffer(bytes);
+   stuffer.append(y_is_odd ? 0x03 : 0x02);
+   this->serialize_x_to(stuffer.next(fe_bytes));
+}
+
+EC_Point EC_AffinePoint_Data_PC::to_legacy_point() const {
+   auto x = BigInt::from_bytes(m_x_bytes);
+   auto y = BigInt::from_bytes(m_y_bytes);
+   return EC_Point(m_group->curve(), std::move(x), std::move(y));
```
reneme, line 164: Nit: The modern C++ language nerds consider explicit `std::move` invocations as a code smell (outside of c'tor member initializer lists). Technically, this could be:

```suggestion
   return EC_Point(m_group->curve(), BigInt::from_bytes(m_x_bytes), BigInt::from_bytes(m_y_bytes));
```

... without a real impact on readability in this particular instance.

Without saying that it is relevant for performance here, but passing the rvalue directly into the function is also more efficient. The compiler can just create the return value of `from_bytes()` right into the argument location of the `EC_Point()` constructor. There won't be an invocation of the move c'tor of `BigInt` and the short-lived (moved-from) `BigInt` object also doesn't need to be destructed.

See: https://godbolt.org/z/c1TK3v9nK

reneme (approved):
Some minor documentation inconsistencies, a suggestion for more explicit memory management and some C++ [Klugscheißerei](https://www.wordreference.com/deen/Klugschei%C3%9Ferei).

src/lib/pubkey/ec_group/ec_inner_data.cpp
randombit: Oops yes off by one here

src/lib/pubkey/ec_group/ec_inner_pc.cpp
randombit: Intentional for the error message - this function is (at least through several layers) reachable by user code, and while they could probably figure it out just from a generic error message it seemed nicer to be a bit more explicit. TBH I think of `copy_mem`s check as more of a fallback for cases we forget.

src/lib/pubkey/ec_group/ec_inner_data.cpp
randombit: Fixed

src/lib/pubkey/ec_group/ec_inner_pc.cpp
randombit: Added an assert here

PR #4145, opened 2024-06-22
Remove fixed block size implementations of XEX encryption

randombit:
Somewhat inexplicably this saves 120 KB in the final shared object


reneme (approved):
> Somewhat inexplicably this saves 120 KB in the final shared object

Even though it is not used anywhere? Curious!

PR #4146, opened 2024-06-24
Add specialized reduction for P256 in pcurves

randombit:
For 32-bit x86, this reduction results in point arithmetic operations that are 25-35% faster than when using Montgomery.

Sadly for 64-bit x86 it is at best about even with using Montgomery, and for Clang 64-bit it's even somewhat slower.


randombit:
Replaced by #4147

PR #4147, opened 2024-06-24
Faster pcurves reductions for P-256 and P-384

randombit:


reneme (approved):
Is that some standard approach that warrants a reference, perhaps?

With some templates the code could be quite a bit more compact, and IMHO also easier to understand; i.e. self-contained functions vs. classes; arrays vs. long lists of variables. See here: https://github.com/randombit/botan/commit/18dc508cf7ea1851cfeba2c2817f74c400df00e5

And here's some Godbolt to play with this implementation: https://godbolt.org/z/4WGGzzYn8

src/lib/math/pcurves/pcurves_secp256r1/pcurves_secp256r1.cpp
```diff
@@ -6,15 +6,115 @@
...
+         sum.accum(S1);
+         sum.accum(S2);
+         sum.accum(S3);
+         sum.accum(S4);
+         sum.accum(S5);
+         sum.accum(S6);
+         sum.accum(S7);
+         const auto S = sum.final_carry(S8);
+
+         CT::unpoison(S);
```
reneme, line 69: This seems to be a stray. Didn't see any corresponding `CT::poison()`. Same for secp384r1. Or am I missing something?

src/lib/math/pcurves/pcurves_secp256r1/pcurves_secp256r1.cpp
```diff
@@ -6,15 +6,115 @@
...
+template <typename Params>
+class Secp256r1Rep final {
+   public:
+      static constexpr auto P = Params::P;
+      static constexpr size_t N = Params::N;
+      typedef typename Params::W W;
+
+      // Adds 4 * P-256 to prevent underflow
+      static constexpr auto P256_4 =
+         hex_to_words<uint32_t>("0x3fffffffc00000004000000000000000000000003fffffffffffffffffffffffc");
```
reneme, line 25: If you decide to adopt my suggestion, it would be worthwhile to replace this by a constexpr multiplication to save a magic string/number, if we have that somewhere already.

src/lib/math/pcurves/pcurves_secp256r1/pcurves_secp256r1.cpp
randombit: The input is (when running under valgrind) already poisoned. As a result `S` is also poisoned, being itself derived from a value that (valgrind is treating as) undefined. This makes the assert trigger an error since it directly jumps if the value is something vs another thing. However the code here is quite resilient should the offset by wrong, so I replaced this with a debug assert.

randombit:
> Is that some standard approach that warrants a reference, perhaps?

Yes this is the standard Solinas reduction (SP 800-186) just done in columns instead of forming each integer directly. I added some references.

> With some templates the code could be quite a bit more compact, and IMHO also easier to understand; i.e. self-contained functions vs. classes; arrays vs. long lists of variables. See here: https://github.com/randombit/botan/commit/18dc508cf7ea1851cfeba2c2817f74c400df00e5

I'll admit this is cleaner but it's also slower than the original Montgomery code, at least on my machine. :)

reneme:
> ... but it's also slower than the original Montgomery code, at least on my machine. :)

Interesting! On my machine (M2 MacBook Air, Xcode 15.3) there's no difference to your implementation.

### Setup

```bash
./configure.py \
   --build-tool=ninja \
   --build-targets=static,cli \
   --compiler-cache=ccache \
   --enable-experimental-features \
   --disable-modules=pcurves_secp256r1,pcurves_secp521r1,pcurves_brainpool256r1,pcurves_brainpool384r1,pcurves_brainpool512r1
ninja cli
./botan speed --msec=1000 pcurves
```

### Results

| Exercise                | René        | Jack         | master      |
| ----------------------- | ----------- | ------------ | ----------- |
| secp384r1 base mul      | 5950/sec    | 5997/sec     | 4828 /sec   |
| secp384r1 var mul       | 1517/sec    | 1514/sec     | 1227 /sec   |
| secp384r1 mul2 setup    | 2034/sec    | 2029/sec     | 1594 /sec   |
| secp384r1 mul2          | 2217/sec    | 2250/sec     | 1815 /sec   |
| secp384r1 proj->affine  | 26603/sec   | 25764/sec    | 20005 /sec  |
| secp384r1 scalar invert | 19650/sec   | 19973/sec    | 20544 /sec  |

randombit:
GCC 14.1.1, Linux, i5-2520M


| Exercise                | René        | Jack         | master      |
| ----------------------- | ----------- | ------------ | ----------- |
| secp384r1 base mul      | 3350/sec    | 5149/sec     | 4159/sec   |
| secp384r1 var mul       | 879/sec    | 1366/sec     | 1117 /sec   |
| secp384r1 mul2 setup    | 1136/sec    | 1880/sec     | 1441 /sec   |
| secp384r1 mul2          | 1307/sec    | 2053/sec     | 1636/sec   |
| secp384r1 proj->affine  | 15319/sec   | 26978/sec    | 17389 /sec  |
| secp384r1 scalar invert | 21183/sec   | 21183/sec    | 21078 /sec  |

reneme:
> I just checked with LLVM Clang on my machine and there I see similar results between the two approaches. Kind of depressing that GCC regresses so badly.

Indeed. I think I should have another look at the performance of #4024 w/ gcc. 😨

reneme:
> I think I should have another look at the performance of https://github.com/randombit/botan/pull/4024 w/ gcc

FTR: [Seems to be fine](https://github.com/randombit/botan/pull/4024#issue-2253206839).

PR #4149, opened 2024-06-24
Make use of __builtin_addc/__builtin_subc where available

randombit:
Both Clang and GCC seem to generate very good code when this builtin is available, at least on x86-64.

This improves performance of a wide range of integer based algorithms by ~5% or so on x86-64. It likely improves the situation even more on platforms, which don't have dedicated assembly.

### DIFF
--- a/src/lib/math/mp/mp_asmi.h
+++ b/src/lib/math/mp/mp_asmi.h
@@ -197,7 +197,7 @@ inline constexpr auto word_add(W x, W y, W* carry) -> W {
 
    W z = x + y;
    W c1 = (z < x);
-   z += *carry;
+   z += *carry & 1;
    *carry = c1 | (z < *carry);
    return z;
 }
@@ -300,7 +300,7 @@ inline constexpr auto word_sub(W x, W y, W* carry) -> W {
 
    W t0 = x - y;
    W c1 = (t0 > x);
-   W z = t0 - *carry;
+   W z = t0 - (*carry & 1);
    *carry = c1 | (z > t0);
    return z;
 }
--- a/src/lib/math/mp/mp_asmi.h
+++ b/src/lib/math/mp/mp_asmi.h
@@ -175,15 +175,25 @@ inline constexpr auto word_madd3(W a, W b, W c, W* d) -> W {
 */
 template <WordType W>
 inline constexpr auto word_add(W x, W y, W* carry) -> W {
-#if defined(BOTAN_MP_USE_X86_64_ASM)
-   if(std::same_as<W, uint64_t> && !std::is_constant_evaluated()) {
-      asm(ADD_OR_SUBTRACT(ASM("adcq %[y],%[x]"))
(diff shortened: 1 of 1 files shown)


randombit:
Before merging, need to double check that neither Clang or GCC does something surprising/unpleasant with this on a range of common CPUs (x86-32, arm32, aarch64, powerpc64, riscv)

randombit:
Indeed the mask seems to be important since, for whatever bad reason, these builtins are actually defined as full-word a+b+c except that the carry is always at most 1, even though if all of a=b=c=max then you might (mathematically) end up with a carry out of 2. Masking allows the compiler to perform range analysis so it knows this is really a+b+carrybit.

Good point that we should apply this to the baseline version. It anyway makes the intent of the code more clear, and might help out with compilers that don't support a builtin.

The performance of the individual `word_add` is not that important - no matter what, the setup cost of getting the carry bit in place is really high. The main thing is that, given some definition of `word_add`, and an unrolled/known at compile time length sequence of those calls, does the compiler convert it to an `adc` carry chain

```
        mov     edx, DWORD PTR [rsi]
        adc     DWORD PTR [rdi], edx
        mov     edx, DWORD PTR [rsi+4]
        adc     DWORD PTR [rdi+4], edx
        mov     edx, DWORD PTR [rsi+8]
        adc     DWORD PTR [rdi+8], edx
```

https://godbolt.org/z/PMzaE845n

randombit:
Per godbolt, powerpc64-clang produces conditional jumps using `__builtin_addc` *or* using the baseline version. There may be no saving us there.

MSVC doesn't seem to produce jumps, but the codegen is horrible :sob:

```
        mov     edx, DWORD PTR [rbx+12]
        mov     r8d, edi
        setb    r9b
        add     edx, DWORD PTR [rcx+12]
        or      r9d, eax
        mov     eax, edi
        cmp     edx, DWORD PTR [rcx+12]
        setb    al
        add     edx, r9d
        cmp     edx, r9d
        mov     DWORD PTR [rcx+12], edx
```

I'm sure it's always been this way I just had never checked.

Codegen for CPUs without a carry flag (RISCV, MIPS) is terrible as well, but they did that to themselves.

On x86-64, GCC without the builtin (it was added in GCC 14), codegen is terrible also, so we have to retain the inline asm for now to avoid a regression. Interestingly Clang *does* realize that it's a carry chain even without the use of the builtin.

randombit:
armv7 and aarch64 - on both, clang generates a carry chain even without the builtin, GCC doesn't generate a carry chain even with the builtin. :cry:  No jumps.

reneme:
Thanks for the details! Very enlightening.

randombit:
I'm going to open a bug with GCC regarding the missed optimization for carry chains on ARM. I think there has been some recent work here related to the new `_BitInt` support, but it required specific backend support that apparently has only been done for x86 so far.

PR #4150, opened 2024-06-26
Use std::lcm in Cascade_Cipher

randombit:


reneme (approved):
There is `std::lcm` and it is constexpr and it was available in C++17 already?

# 🤯

PR #4153, opened 2024-06-27
Properly mask the carry bit in word_add

randombit:
In #4149 a mask was added for the low bit of the carry/borrow, since this can improve codegen, however it was not done correctly for word_add.


reneme (approved):
Good catch. 😶

PR #4154, opened 2024-06-27
Add CT::Choice::yes() and ::no()

reneme:
Allow creating a `CT::Choice` representing true or false (e.g. for unit testing). Also: restrict `CT::Choice::from_int()` to not work for `bool` as this produces compiler warnings* and is probably undesired (i.e. a code-smell) anyway. I need those for the requested adaptions to `bitvector<>` in #3107.

[*] The compiler is complaining that we try to negate a bool using `~` instead of `!`.

### DIFF
--- a/src/lib/utils/ct_utils.h
+++ b/src/lib/utils/ct_utils.h
@@ -119,7 +119,8 @@ class Choice final {
       /**
       * If v == 0 return an unset (false) Choice, otherwise a set Choice
       */
-      template <std::unsigned_integral T>
+      template <typename T>
+         requires std::unsigned_integral<T> && (!std::same_as<bool, T>)
       constexpr static Choice from_int(T v) {
          // Mask of T that is either |0| or |1|
          const T v_is_0 = ct_is_zero<T>(value_barrier<T>(v));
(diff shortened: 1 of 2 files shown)


src/lib/utils/ct_utils.h
```diff
@@ -134,6 +135,10 @@ class Choice final {
          return Choice(ct_is_zero<uint32_t>(static_cast<uint32_t>(v_is_0)));
       }
 
+      constexpr static Choice yes() { return Choice::from_int(1U); }
+
+      constexpr static Choice no() { return Choice::from_int(0U); }
```
randombit, line 140: These could just directly initialize the masked values (either all zero or all ones resp), no need for them to be guarded by the value barrier.

reneme: Given that the constexpr-ness of the factory those delegate to, I was hoping that the compiler would do that for me and not actually compile a value barrier. But yeah, no harm in doing that explicitly.

randombit: That would only apply if you created the variable as `constexpr` though, right?

```
constexpr auto yes = Choice::yes(); // compile time evaluated, no value barrier
const auto yes = Choice::yes(); // evaluted at runtime, presumably with value barrier
```

randombit:
BTW agree that we don't want/need any obvious conversion from `bool` -> Choice, since if the sensitive value is already a `bool` then it's too late, you've been eaten by the value range propagator.

randombit (approved):

PR #4155, opened 2024-06-27
Fix various module dependency bugs

randombit:


reneme (approved):
Does it make sense to try and establish a CI job (maybe as a nightly) for checking this? E.g. like so:

```python
for module in modules:
   if not module.internal():
      configure_minimized_build_plus(module)
      compile()
```

randombit:
Maaayybe. This was found using `src/scripts/test_all_configs.py` btw. But this is probably too slow even for nightly, it takes like an hour on a 128 core machine. I run it manually every release or so. The impact if we miss something is low severity (fail to build in a specific non-default configuration, with a trivial workaround) so that seems good enough to me.

reneme:
Severity is low for sure, but as a user, if I want to restrict the module policy, I'd be annoyed if I constantly need to work around  missing dependencies, I guess. Especially if I need this to work across several minor releases.

Another idea: perhaps we could restrict the test corpus of `test_all_configs.py` to just the modules that are touched by a specific PR, maybe with a timing cutoff. I think there's value in catching such things while developing.

PR #4159, opened 2024-06-29
Bump some of the more expensive or obscure cross builds to nightly

randombit:


.github/workflows/ci.yml
```diff
@@ -258,34 +258,19 @@ jobs:
...
-            host_os: ubuntu-24.04
           - target: cross-ppc64
             compiler: gcc
             host_os: ubuntu-24.04
-          - target: cross-riscv64
-            compiler: gcc
-            host_os: ubuntu-24.04
-          - target: cross-s390x
-            compiler: gcc
-            host_os: ubuntu-24.04
```
reneme, line 272: Not sure about this one. Isn't it the only platform on big endian? Would be good to catch those before merging things.

randombit: Ah yeah you're right. Let's see if MIPS64 is faster here #4167

randombit: Time to run the tests under QEMU

S390 847 seconds
MIPS64 357 seconds

8 minutes faster :grimacing: 

So I would propose bumping S390 to nightly and keeping MIPS64 as our big endian test.

reneme (approved):

Issue #4172, opened 2024-07-02
Add CT::Option


randombit:
In #4171 we return a pair of `T` plus a `CT::Choice` indicating validity. Which is good enough for that one usage. But it would be nice to generalize this to a `CT::Option<T>` which either contains a value or does not.

Can probably copy some ideas from https://docs.rs/subtle/latest/subtle/struct.CtOption.html

PR #4175, opened 2024-07-02
Add CT::Option

randombit:
GH #4172

### DIFF
--- a/src/lib/utils/ct_utils.h
+++ b/src/lib/utils/ct_utils.h
@@ -16,8 +16,8 @@
 
 #include <botan/secmem.h>
 #include <botan/internal/bit_ops.h>
+#include <optional>
 #include <type_traits>
-#include <vector>
 
 #if defined(BOTAN_HAS_VALGRIND)
    #include <valgrind/memcheck.h>
@@ -172,6 +172,12 @@ class Choice final {
 
       uint32_t m_value;
 };
+
+/**
+* A concept for a type which is conditionally assignable
+*/
+template <typename T>
+concept ct_conditional_assignable = requires(T lhs, const T& rhs, Choice c) { lhs.conditional_assign(c, rhs); };
 
 /**
 * A Mask type used for constant-time operations. A Mask<T> always has value
@@ -420,6 +426,87 @@ class Mask final {
       constexpr Mask(T m) : m_mask(m) {}
 
       T m_mask;
+};
+
+/**
+* A CT::Option<T> is either a valid T, or not
+*
+* To maintain constant time behavior a value must always be stored.
+* A CT::Choice tracks if the value is valid or not. It is not possible
+* to access the inner value if the Choice is unset.
+*/
+template <typename T>
+class Option {
+   public:
+      /// Construct an Option which contains the specified value, and is set or not
+      constexpr Option(T v, Choice valid) : m_has_value(valid), m_value(std::move(v)) {}
+
+      /// Construct a set option with the provided value
+      constexpr Option(T v) : Option(std::move(v), Choice::yes()) {}
+
+      /// Construct an unset option with a default inner value
+      constexpr Option()
+         requires std::default_initializable<T>
+            : Option(T(), Choice::no()) {}
+
+      /// Return true if this Option contains a value
+      constexpr Choice has_value() const { return m_has_value; }
+
+      /**
+      * Apply a function to the inner value and return a new Option
+      * which contains that value. This is constant time only if @p f is.
+      *
+      * @note The function will always be called, even if the Option is None. It
+      *       must be prepared to handle any possible state of T.
+      */
+      template <std::invocable<const T&> F>
+      constexpr auto transform(F f) const -> Option<std::remove_cvref_t<std::invoke_result_t<F, const T&>>> {
+         return {f(m_value), m_has_value};
+      }
+
+      /// Either returns the value or throws an exception
+      constexpr const T& value() const {
+         BOTAN_STATE_CHECK(m_has_value.as_bool());
+         return m_value;
+      }
+
+      /// Returns either the inner value or the alternative, in constant time
+      ///
+      /// This variant is used for types which explicitly define a function
+      /// conditional_assign which takes a CT::Choice as the conditional.
+      constexpr T value_or(T other) const
+         requires ct_conditional_assignable<T>
+      {
+         other.conditional_assign(m_has_value, m_value);
+         return other;
+      }
+
+      /// Returns either the inner value or the alternative, in constant time
+      ///
+      /// This variant is used for integer types where CT::Mask can perform
+      /// a constant time selection
+      constexpr T value_or(T other) const
+         requires std::unsigned_integral<T>
+      {
+         auto mask = CT::Mask<T>::from_choice(m_has_value);
+         return mask.select(m_value, other);
+      }
+
+      /// Convert this Option into a std::optional
+      ///
+      /// This is not constant time, leaking if the Option had a
+      /// value or not
+      constexpr std::optional<T> as_optional_vartime() const {
+         if(m_has_value.as_bool()) {
+            return {m_value};
+         } else {
+            return {};
+         }
+      }
+
+   private:
+      Choice m_has_value;
+      T m_value;
 };
 
 template <typename T>
--- a/src/tests/test_ct_utils.cpp
+++ b/src/tests/test_ct_utils.cpp
@@ -1,0 +1,180 @@
+/*
+* (C) 2018,2024 Jack Lloyd
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include "tests.h"
+#include <botan/internal/ct_utils.h>
+#include <botan/internal/fmt.h>
+#include <functional>
+
+namespace Botan_Tests {
+
+class CT_Mask_Tests final : public Test {
+   public:
+      std::vector<Test::Result> run() override {
+         Test::Result result("CT::Mask");
+
+         result.test_eq_sz("CT::is_zero8", Botan::CT::Mask<uint8_t>::is_zero(0).value(), 0xFF);
+         result.test_eq_sz("CT::is_zero8", Botan::CT::Mask<uint8_t>::is_zero(1).value(), 0x00);
+         result.test_eq_sz("CT::is_zero8", Botan::CT::Mask<uint8_t>::is_zero(0xFF).value(), 0x00);
+
+         result.test_eq_sz("CT::is_zero16", Botan::CT::Mask<uint16_t>::is_zero(0).value(), 0xFFFF);
+         result.test_eq_sz("CT::is_zero16", Botan::CT::Mask<uint16_t>::is_zero(1).value(), 0x0000);
+         result.test_eq_sz("CT::is_zero16", Botan::CT::Mask<uint16_t>::is_zero(0xFF).value(), 0x0000);
+
+         result.test_eq_sz("CT::is_zero32", Botan::CT::Mask<uint32_t>::is_zero(0).value(), 0xFFFFFFFF);
+         result.test_eq_sz("CT::is_zero32", Botan::CT::Mask<uint32_t>::is_zero(1).value(), 0x00000000);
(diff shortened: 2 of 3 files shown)


src/lib/utils/ct_utils.h
```diff
@@ -173,6 +173,63 @@ class Choice final {
       uint32_t m_value;
 };
 
+/**
+* A CT::Option<T> is either a valid T (Some), or a default T (None)
+*
+* T must be default construtible
+*/
+template <typename T>
+class Option {
```
reneme, line 182:
```suggestion
/**
* A CT::Option<T> is either a valid T (Some), or a default T (None)
*/
template <std::default_initializable T>
class Option {
```

src/lib/utils/ct_utils.h
```diff
@@ -173,6 +173,63 @@ class Choice final {
...
+   public:
+      constexpr Option(T v, Choice valid) : m_is_some(valid), m_value(v) {}
+
+      constexpr Option(T v) : m_is_some(Choice::yes()), m_value(v) {}
+
+      constexpr Option() : m_is_some(Choice::no()), m_value(T()) {}
```
reneme, line 188:
```suggestion
      constexpr Option(T v, Choice valid) : m_is_some(valid), m_value(std::move(v)) {}

      constexpr Option(T v) : m_is_some(Choice::yes()), m_value(std::move(v)) {}

      constexpr Option() : m_is_some(Choice::no()), m_value(T()) {}
```

src/lib/utils/ct_utils.h
```diff
@@ -173,6 +173,63 @@ class Choice final {
...
+
+      /**
+      * Apply a function to the inner value and return a new Option
+      * which contains that value. This is constant time only if `f` is.
+      */
+      template <typename F>
+      constexpr auto map(F f) const {
+         typedef typename std::invoke_result<F, T>::type U;
+         return Option<U>(f(m_value), m_is_some);
+      }
```
reneme, line 204:
```suggestion
      /**
      * Apply a function to the inner value and return a new Option
      * which contains that value. This is constant time only if @p f is.
      *
      * @note The function will always be called, even if the Option is None. It
      *       must be prepared to handle any possible state of T.
      */
      template <std::invocable<const T&> F>
      constexpr auto map(F f) const -> Option<std::invoke_result_t<F, const T&>> {
         return {f(m_value), m_is_some};
      }
```

src/lib/utils/ct_utils.h
```diff
@@ -173,6 +173,63 @@ class Choice final {
...
+
+      /// Variable time
+      std::optional<T> as_optional() const {
+         if(m_is_some.as_bool()) {
+            return {m_value};
+         } else {
+            return {};
+         }
+      }
```
reneme, line 226:
```suggestion
      /// Variable time
      std::optional<T> as_optional_vartime() const {
         if(m_is_some.as_bool()) {
            return {m_value};
         } else {
            return {};
         }
      }
```

src/lib/utils/ct_utils.h
```diff
@@ -173,6 +173,63 @@ class Choice final {
...
+         BOTAN_STATE_CHECK(m_is_some.as_bool());
+         return m_value;
+      }
+
+      /// Returns either the inner value or the alternative, in constant time
+      T unwrap_or(const T& other) const {
+         T r = other;
+         r.conditional_assign(m_is_some, m_value);
+         return r;
+      }
```
reneme, line 217: Probably a good idea to scope this to types that actually provide this method. And also make the copy of `other` visible via the interface (to optimize usage with rvalue refs for instance):

```suggestion
      /// Returns either the inner value or the alternative, in constant time
      T unwrap_or(T other) const 
         requires concepts::ct_conditional_assignable<T>
      {
         other.conditional_assign(m_is_some, m_value);
         return other;
      }
```

... and the concept itself:

```C++
namespace concepts {

template <typename T>
concept ct_conditional_assignable = requires(T lhs, const T& rhs, Choice c) { lhs.conditional_assign(c, rhs); };

}  // namespace concepts

}
```

... if you wanted to, you could even provide a variable time overload if the concept isn't fulfilled. Though, I'd consider this fairly reckless, as it would potentially fool the user if their type doesn't model the concept (perhaps even subtly inadvertantly).

reneme (commented):
Nice idea!

Frankly, I would suggest to stay with the STL's nomenclature of things, though. I.e.

* `unwrap()` -> `value()`
* `unwrap_or()` -> `value_or()`
* `is_some()` -> `has_value()`
* `map()` -> `transform()` (C++23: [std::optional<>::transform](https://en.cppreference.com/w/cpp/utility/optional/transform))

src/lib/utils/ct_utils.h
reneme: It might make sense to look into the other overloads of [`std::optional::transform`](https://en.cppreference.com/w/cpp/utility/optional/transform). For instance, if `this` is an rvalue-ref we could `std::move(*this)` into the function.

reneme: Another thing to consider: If `m_is_some` is indicating "None", the resulting `Option<U>` will be "None" but its "value" won't be a default-initialized `U` but whatever came back from `f()`. That may or may not be problematic, but it is probably worth documenting that `m_is_some == None` does not implicate that `m_value` is default inited.

src/lib/utils/ct_utils.h
randombit: I'd be quite against a variable time overload. That said it would be nice if we could handle this more generically, ie `CT::Option<uint32_t>` is entirely a sensible thing, and it's quite possible to implement `unwrap_or` in constant time as well, but requires additional help. I'll ponder this.

randombit: Nice way of expressing the conditional assignment btw.

I made a change such that `value_or` can handle either types which explicitly provided an appropriate `conditional_assign`, and for unsigned integral types which make use of `CT::Mask` to perform the selection.

src/tests/test_ct_utils.cpp
```diff
@@ -0,0 +1,180 @@
...
+*/
+
+#include "tests.h"
+#include <botan/internal/ct_utils.h>
+#include <botan/internal/fmt.h>
+#include <functional>
+
+namespace Botan_Tests {
+
+class CT_Mask_Tests final : public Test {
```
randombit, line 14: I moved the CT tests to a new file as `test_utils.cpp` is getting quite large. The Mask and Choice tests are unchanged

src/lib/utils/ct_utils.h
```diff
@@ -422,6 +428,87 @@ class Mask final {
       T m_mask;
 };
 
+/**
+* A CT::Option<T> is either a valid T, or not
+*
+* To maintain constant time behavior a value must always be stored.
+* A CT::Choice tracks is the value is valid or not. It is not possible
+* to access the inner value if the Choice is unset.
+*/
```
reneme, line 437:
```suggestion
/**
* A CT::Option<T> is either a valid T, or not
*
* To maintain constant time behavior a value must always be stored.
* A CT::Choice tracks if the value is valid or not. It is not possible
* to access the inner value if the Choice is unset.
*/
```

src/lib/utils/ct_utils.h
```diff
@@ -422,6 +428,87 @@ class Mask final {
...
+      /// Returns either the inner value or the alternative, in constant time
+      ///
+      /// This variant is used for integer types where CT::Mask can perform
+      /// a constant time selection
+      T value_or(T other) const
+         requires std::unsigned_integral<T>
+      {
+         auto mask = CT::Mask<T>::from_choice(m_is_some);
+         return mask.select(m_value, other);
+      }
```
reneme, line 493: Nice!

src/lib/utils/ct_utils.h
```diff
@@ -422,6 +428,87 @@ class Mask final {
...
+      template <std::invocable<const T&> F>
+      constexpr auto transform(F f) const -> Option<std::invoke_result_t<F, const T&>> {
+         return {f(m_value), m_is_some};
+      }
+
+      /// Either returns the value or throws an exception
+      T value() const {
+         BOTAN_STATE_CHECK(m_is_some.as_bool());
+         return m_value;
+      }
```
reneme, line 471: Should that really return a copy of `m_value`? Maybe:

```suggestion
      /// Either returns the value or throws an exception
      const T& value() const {
         BOTAN_STATE_CHECK(m_is_some.as_bool());
         return m_value;
      }
```

... for completeness, [`std::optional` defines this both for lvalue and rvalue `this`](https://en.cppreference.com/w/cpp/utility/optional/value). Not sure we really need that, though. It just optimizes cases where a temporary is passed on.

```C++
takes_an_Option_by_value(returns_an_option_by_value().value())
```

src/lib/utils/ct_utils.h
```diff
@@ -422,6 +428,87 @@ class Mask final {
...
+      * Apply a function to the inner value and return a new Option
+      * which contains that value. This is constant time only if @p f is.
+      *
+      * @note The function will always be called, even if the Option is None. It
+      *       must be prepared to handle any possible state of T.
+      */
+      template <std::invocable<const T&> F>
+      constexpr auto transform(F f) const -> Option<std::invoke_result_t<F, const T&>> {
+         return {f(m_value), m_is_some};
+      }
```
reneme, line 465: Mhh, here the return type might need some extra case. Imagine `F` returns a reference for some reason, then you would end up instantiating an `Option<T&>` which most certainly isn't what we want.

Either do

```suggestion
      template <std::invocable<const T&> F>
      constexpr auto transform(F f) const -> Option<std::remove_cvref_t<std::invoke_result_t<F, const T&>>> {
         return {f(m_value), m_is_some};
      }
```

and/or exclude references for `typename T` in general with

```C++
template <typename T>
   requires !std::is_reference_t<T>
class Option {
```

reneme (commented):
Perhaps also `constexpr` the remaining member functions. I don't see why any of them wouldn't be `constexpr`.

src/lib/utils/ct_utils.h
randombit: I suppose I was thinking of `Option` as largely for holding types that are cheap to copy (integers, simple arrays, etc) where this wouldn't matter, but yeah no reason not to return a reference here, `CT::Option<vector<uint8_t>>` is entirely sensible and might indeed be something worth using in PK decryption or the like.

src/lib/utils/ct_utils.h
randombit: `Option<&T>` might make sense in some scenario. I can't think of an immediate use case though

reneme: Perhaps. But this is what `std::reference_wrapper<>` is for, IMHO. Otherwise, imagine this:

```C++
auto foo(Option<MyConstTimeType> o) {
   return o.transform([](const MyConstTimeType& ov) {
      return ov.get_some_member_by_reference();
   });
}
```

... `foo()` would return an `Option<T&>` that would most-certainly be a dangling reference, given that the getter returns a reference to a member of the local copy of `Option<T> o`. That's a source of very subtle memory bugs.

reneme (approved):

src/lib/utils/ct_utils.h
reneme: When we foresee to use it with `std::vector<uint8_t>` it would actually make sense to implement the lvalue/rvalue overloads that `std::optional<>` implements. In fact, this just made me re-think the `strong_type_unwrap()` helper: https://github.com/randombit/botan/pull/4170/files#r1672372281

reneme:
Perhaps its worth to reconsider [this](https://github.com/randombit/botan/pull/4175#discussion_r1672377336) before merging.

src/lib/utils/ct_utils.h
randombit: Reading https://stackoverflow.com/questions/24824432/what-is-use-of-the-ref-qualifier-const I am unconvinced by `const &&` and don't really see benefit. I'm going to merge this as is and if/when Option of non-trivally-copyable types comes up we can consider the best course.

Issue #4112, opened 2024-06-10
Somewhat regular failures of gcc sanitizer tls_proxy cli test


randombit:
Nothing but

>    ERROR: Unexpected return code from tls_proxy 1

https://github.com/randombit/botan/actions/runs/9439667258/job/25999341641?pr=4042

I've seen this several times, always on this specific build (not for instance clang)

Probably better logging from the CLI test runner would help

randombit:
I've tried running it locally a few hundred times with sanitizers enabled, no crash `¯\_(ツ)_/¯`

randombit:
https://github.com/randombit/botan/actions/runs/9533945244/job/26277782022

```
2024-06-16T06:42:46.2160516Z    INFO: Stderr: =================================================================
2024-06-16T06:42:46.2162791Z ==16939==ERROR: AddressSanitizer: heap-use-after-free on address 0x515000002890 at pc 0x5601f64f5e52 bp 0x7fd5cbbfeeb0 sp 0x7fd5cbbfeea0
2024-06-16T06:42:46.2164664Z READ of size 8 at 0x515000002890 thread T2
2024-06-16T06:42:46.2169102Z     #0 0x5601f64f5e51 in boost::asio::detail::iterator_connect_op<boost::asio::ip::tcp, boost::asio::any_io_executor, boost::asio::ip::basic_resolver_iterator<boost::asio::ip::tcp>, boost::asio::detail::default_connect_condition, Botan_CLI::(anonymous namespace)::tls_proxy_session::tls_session_activated()::{lambda(boost::system::error_code, boost::asio::ip::basic_resolver_iterator<boost::asio::ip::tcp> const&)#1}>::operator()(boost::system::error_code, int) (/home/runner/work/botan/botan/build/botan+0xb06e51) (BuildId: c57781d1296d22b403cc28959fc40e0ba7e36ad3)
[...]

PR #4177, opened 2024-07-04
FIX: potential use-after-free in `./botan tls_proxy`

reneme:
This fixes an issue that, in rare cases, causes a use-after-free in the `tls_proxy` CLI tool.

```mermaid
sequenceDiagram
    Client->>+Proxy: Connect
    Client->>Proxy: TLS Handshake
    Proxy-->>Client: 
    Proxy->>+Server: Connect
    Client->>Proxy: Disconnect
    Proxy-->>-Client: 
    Note left of Proxy: this-ptr invalidated
    Server-->>-Proxy: ACK
    Note over Proxy: Crash!
```

If the TLS session to the client was established (when `tls_session_activated()` was called), and the connection to the server was also established successfully (error code in `onConnect()` callback was not set); but (in the mean time) the `this`-pointer was deallocated via `std::enable_shared_from_this`, we end up in a use-after-free situation.

This sporadically appeared in CI after updating to Ubuntu 24.04 but wasn't reproducible locally, see #4112. I am confident that we did [see this race on Windows in January](https://github.com/randombit/botan/blame/d24c2c3b2e74c1de10b7bd47317eaf50cebf5fb4/src/scripts/test_cli.py#L1302-L1305) but didn't investigate any further.

NOTE: _This alone won't fix the glitches in CI._ They just won't make `./botan tls_proxy` crash any longer. I'll address the CI glitches independently.

### DIFF
--- a/src/cli/tls_proxy.cpp
+++ b/src/cli/tls_proxy.cpp
@@ -282,13 +282,20 @@ class tls_proxy_session final : public std::enable_shared_from_this<tls_proxy_se
       }
 
       void tls_session_activated() override {
-         auto onConnect = [this](boost::system::error_code ec, const tcp::resolver::iterator& /*endpoint*/) {
+         auto onConnect = [self = weak_from_this()](boost::system::error_code ec,
+                                                    const tcp::resolver::iterator& /*endpoint*/) {
             if(ec) {
                log_error("Server connection", ec);
                return;
             }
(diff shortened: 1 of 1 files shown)


randombit:
Subtle! Nice work tracking this down.

randombit (approved):

Issue #3991, opened 2024-04-09
Odd CLI test failure on MacOS 14 runners


randombit:
I've seen this twice now:

```
python3 ./src/scripts/test_cli.py --threads=3 ./botan
     INFO: Ran cli_base32_tests in 0.02 sec
     INFO: Ran cli_asn1_tests in 0.03 sec
     INFO: Ran cli_argon2_tests in 0.05 sec
     INFO: Ran cli_base64_tests in 0.01 sec
     INFO: Ran cli_base58_tests in 0.03 sec
     INFO: Ran cli_cycle_counter in 0.01 sec
     INFO: Ran cli_bcrypt_tests in 0.02 sec
     INFO: Ran cli_cc_enc_tests in 0.05 sec
     INFO: Ran cli_compress_tests in 0.02 sec
     INFO: Ran cli_config_tests in 0.02 sec
     INFO: Ran cli_cpuid_tests in 0.06 sec
     INFO: Ran cli_dl_group_info_tests in 0.07 sec
     INFO: Ran cli_cert_issuance_tests in 0.22 sec
     INFO: Ran cli_entropy_tests in 0.01 sec
     INFO: Ran cli_ec_group_info_tests in 0.02 sec
     INFO: Ran cli_cert_issuance_alternative_algos_tests in 0.27 sec
     INFO: Ran cli_gen_prime_tests in 0.01 sec
     INFO: Ran cli_hash_tests in 0.03 sec
     INFO: Ran cli_help_tests in 0.01 sec
     INFO: Ran cli_hex_tests in 0.01 sec
     INFO: Ran cli_hmac_tests in 0.01 sec
     INFO: Ran cli_is_prime_tests in 0.02 sec
     INFO: Ran cli_factor_tests in 0.15 sec
     INFO: Ran cli_marvin_tests in 0.06 sec
     INFO: Ran cli_mod_inverse_tests in 0.01 sec
     INFO: Ran cli_gen_dl_group_tests in 0.33 sec
     INFO: Ran cli_pk_encrypt_tests in 0.10 sec
     INFO: Ran cli_pk_workfactor_tests in 0.03 sec
     INFO: Ran cli_psk_db_tests in 0.06 sec
     INFO: Ran cli_rng_tests in 0.06 sec
     INFO: Ran cli_roughtime_check_tests in 0.03 sec
     INFO: Ran cli_roughtime_tests in 0.05 sec
     INFO: Ran cli_pbkdf_tune_tests in 0.55 sec
[...]

PR #4178, opened 2024-07-04
FIX: Race conditions in `test_cli.py` when spawning test server processes

### LINKED ISSUE #4112: Somewhat regular failures of gcc sanitizer tls_proxy cli test

reneme:
### Pull Request Dependencies

* #4177 

### Description

The `test_cli.py` script handles and orchestrates a few long-running processes (namely `./botan tls_server`, `./botan tls_client`, `./botan tls_proxy`). For those tests to work properly, we have to ensure that the server processes are ready before trying to interact with them. Especially on the CI where runtime behavior is notoriously unpredictable. Until now, this was done by invoking `time.sleep(1)` after launching a server process. This worked most of the time, but introduced unnecessary waiting time while still being potentially racey.

This now uses Python's `asyncio` module to handle such long-running processes and interact with them. Each CLI server now reports a "Listening for new connections..." string on stdout once its ready to receive connections. The `asyncio`-based wrapper can wait for this string and thus ensure that the server is ready before the test continues. Also, the wrapper makes sure to timeout on any of those waits (hard-coded 15 seconds) and kill the process if necessary.

### Result

Such tests can now rely on their server processes being in a defined state, **without sleeping** some arbitrary time and hoping for the best. As a side effect, running `./test_cli.py --threads=4` now takes 2.5 seconds on my laptop, instead of the more than 20 seconds it used to take due to the `time.sleep()` workarounds.

Fixes #4112

### DIFF
--- a/src/cli/tls_client.cpp
+++ b/src/cli/tls_client.cpp
@@ -125,6 +125,7 @@ class Callbacks : public Botan::TLS::Callbacks {
                output() << certs[i].PEM_encode();
             }
          }
+         output() << std::flush;
       }
 
       void tls_emit_data(std::span<const uint8_t> buf) override {
@@ -141,6 +142,7 @@ class Callbacks : public Botan::TLS::Callbacks {
          for(const auto c : buf) {
             output() << c;
          }
+         output() << std::flush;
       }
 
       std::vector<uint8_t> tls_sign_message(const Botan::Private_Key& key,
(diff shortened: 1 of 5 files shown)


randombit (approved):
Very nice work!

The pylint failures in CI look relevant. If there is anything it's warning about that isn't sensible to change, feel free to locally disable the warning.

randombit:
I think the macOS failure is #3991 so unrelated to this change.

reneme:
Let's hope so. In it goes!

randombit:
Couple of Windows builds in the merge are failing, maybe we need more than 15 seconds in practice on CI?

PR #4179, opened 2024-07-04
Chore: Pylint in VS Code and Pull Requests

reneme:
Set up a pylint reporter plugin to display warnings as annotations in pull requests. Also, fix the VS Code integration of pylint.


src/scripts/ci/pylint_plugins/gh_actions_reporter/__init__.py
```diff
@@ -0,0 +1,47 @@
+"""
+Pylint reporter for GitHub Actions
+"""
+from __future__ import annotations
```
randombit, line 4: What this is doing seems quite simple so we should just do it ourselves. And especially NAK on anything that is not either strictly BSD-2 or public domain.

I know it seems like MIT vs BSD-2 is basically the same, but there are many organizations where us including a second LICENSE file, even for something not used in the final binary, will require an expensive review process. Seems silly I know! But this is a thing. If this seems incredible, be grateful.

If you want I'll do this part. I have barely looked at this code but given the knowledge that pylint has an API is sufficient.

randombit (changes requested):
VSCode changes are fine.

Fine to report pylint warnings in Actions- that's a nice feature we should work on leveraging more. But hard no on anything in this repo not BSD-2 or explicitly public domain.

(As an alternative to rewriting it, we could also fetch it at build time in the CI setup script. That's fine, because then it's not in git or a release.)

reneme:
Fair. Wasn't aware that this might be a problem. Apparently, one can pull it in via `pip` as well, but I wanted to avoid `pip`.

randombit:
Using `pip` seems the way to go here tbh

reneme:
Turns out `pylint` has a `--msg-template=` command line switch. This doesn't let us map the pylint warning levels to GitHub's. Hence, all pylint findings will be displayed as "Warning" in the pull request. But that should be just fine...

image

src/scripts/ci_build.py
```diff
@@ -684,8 +687,7 @@ def main(args=None):
             'src/editors/vscode/scripts/common.py',
             'src/editors/vscode/scripts/test.py']
 
-        full_paths = [os.path.join(root_dir, s) for s in py_scripts]
-        cmds.append([py_interp, '-m', 'pylint'] + pylint_flags + full_paths)
+        cmds.append(["indir:%s" % root_dir, py_interp, '-m', 'pylint'] + pylint_flags + py_scripts)
```
reneme, line 690: The paths reported in the log must be relative to the repository root. To accomplish that without a custom script, we'll have to set CWD to the repo root when running `pylint`.

randombit:
Oh well that turned out nicely. Everything being a warning seems fine - we anyway treat anything that pylint outputs as ==  CI failed, so distinguishing the display based on pylint’s categorization doesn’t seem helpful.

src/scripts/ci_build.py
randombit: Maybe add a comment here along these lines

randombit (approved):
Looks good thanks

src/scripts/ci_build.py
reneme: Haha! That's pure retaliation for https://github.com/randombit/botan/pull/4181#discussion_r1665689550. 😏

randombit: 😅

PR #4180, opened 2024-07-04
Increase timeout in cli tests

randombit:
Sometimes Windows builds fail with

ERROR: Proxy did not terminate in time, will kill it...

So clearly 15 seconds is insufficient at least some of the time on Github Actions.

### DIFF
--- a/src/scripts/test_cli.py
+++ b/src/scripts/test_cli.py
@@ -33,7 +33,7 @@ from http.server import HTTPServer, BaseHTTPRequestHandler
 # pylint: disable=global-statement,unused-argument
 
 CLI_PATH = None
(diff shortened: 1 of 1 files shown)


randombit:
Oof even 30 seconds was not entirely sufficient - there may be a problem here

reneme:
Yeah, perhaps we should disable windows again.

```
ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host
```

... potentially, there's something else wrong, regarding: https://github.com/randombit/botan/blame/d24c2c3b2e74c1de10b7bd47317eaf50cebf5fb4/src/scripts/test_cli.py#L1302-L1305

randombit:
Closing as insufficient/ineffective

PR #4181, opened 2024-07-04
Disable tls_proxy cli test on Windows again

randombit:
This was disabled in #3845 due to flakyness, then thought possibly fixed and enabled again in #4178. However with #4178 merged the test still occasionally fails. Disable it again pending diagnosis.

### DIFF
--- a/src/scripts/test_cli.py
+++ b/src/scripts/test_cli.py
@@ -1433,6 +1433,9 @@ def cli_tls_http_server_tests(tmp_dir):
     asyncio.run(run_async_test())
 
 def cli_tls_proxy_tests(tmp_dir):
+    # This was disabled in GH #3845 due to flakyness, then thought possibly
+    # fixed and enabled again in Gh #4178. However the test still occasionally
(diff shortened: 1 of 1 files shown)


src/scripts/test_cli.py
```diff
@@ -1433,7 +1433,7 @@ async def run_async_test():
     asyncio.run(run_async_test())
 
 def cli_tls_proxy_tests(tmp_dir):
-    if not run_socket_tests() or not check_for_command("tls_proxy"):
+    if not run_socket_tests() or platform.system() == 'Windows' or not check_for_command("tls_proxy"):
```
reneme, line 1436:
```suggestion
    # This was disabled in GH #3845 due to flakyness, then thought possibly
    # fixed and enabled again in Gh #4178. However the test still occasionally
    # fails. Disable it again pending diagnosis...
    if not run_socket_tests() or platform.system() == 'Windows' or not check_for_command("tls_proxy"):
```

reneme (approved):
:sad:

PR #4182, opened 2024-07-05
Test: Run a self-test for our valgrind setup

reneme:
This runs some basic scenarios we'd like to find with valgrind and checks that they are detected as expected. Otherwise, to the best of our understanding, there wouldn't be any negative signal from valgrind in a typical CI run.

We introduced a new build target "valgrind_selftest" for that and added a python-based steering script that runs each of the C++ test cases individually. The "valgrind" CI job builds and runs the selftest before running the unit tests. This is built as a minimal framework to add more test cases over time.

[One of the test cases isn't an _obvious_ side channel](https://github.com/randombit/botan/pull/4182#discussion_r1666833485), it is a reproduction of the latest Kyber issue found by PQShield and will trigger valgrind only in certain compiler configurations. In a follow-up, we might add a nightly valgrind run with such a configuration in the hope to increase the valgrind signal overall. The python-based steering script detects vulnerable compiler settings (from `build-config.json`) and skips this test otherwise.

[Another one is a de-facto regression test for the recently-introduced value-barrier](https://github.com/randombit/botan/pull/4182#discussion_r1666834518). The regression is obviously not visible unless run with vulnerable compiler settings either. Then, with a disabled value-barrier, valgrind does detect a secret-dependent branch successfully.

### DIFF
--- a/src/ct_selftest/ct_selftest.cpp
+++ b/src/ct_selftest/ct_selftest.cpp
@@ -1,0 +1,332 @@
+/*
+ * Basic tests for the CT::poison annotations.
+ * Some of those are expected to fail, because they deliberately
+ * branch on secret memory.
+ *
+ * (C) 2024 Jack Lloyd
+ * (C) 2024 Fabian Albert, René Meusel - Rohde & Schwarz Cybersecurity
+ *
+ * Botan is released under the Simplified BSD License (see license.txt)
+ */
+
+#include <iostream>
+
+#include <botan/hex.h>
+#include <botan/system_rng.h>
+
+#include <botan/internal/ct_utils.h>
+#include <botan/internal/fmt.h>
+
+#include <functional>
+#include <map>
+
+namespace {
+
+void test_conditional_jump_on_poisoned_data(Botan::RandomNumberGenerator& rng) {
+   const uint8_t poisoned_byte = rng.next_byte();
+   Botan::CT::poison(poisoned_byte);
+
+   // Performing a conditional jump on a "secret" value would introduce
+   // a potential side channel.
+   if(poisoned_byte == 0) {
+      std::cout << "I may or may not be printed." << std::endl;
+   }
+}
+
+void test_array_access_on_poisoned_data(Botan::RandomNumberGenerator& rng) {
+   const uint8_t poisoned_byte = rng.next_byte();
+   const auto lookup_table = rng.random_array<1 << (sizeof(poisoned_byte) * 8)>();
+   Botan::CT::poison(poisoned_byte);
+
+   // Accessing memory with a "secret" value would introduce a potential side channel.
+   std::cout << lookup_table[poisoned_byte] << std::endl;
+}
+
+void test_conditional_jump_on_transitively_poisoned_data(Botan::RandomNumberGenerator& rng) {
+   const uint8_t poisoned_byte = rng.next_byte();
+   const uint8_t innocent_byte = rng.next_byte();
+   Botan::CT::poison(poisoned_byte);
+
+   const uint8_t derived_byte = poisoned_byte ^ innocent_byte;
+
+   // Performing a conditional jump on a value that was calculated from a "secret" value
+   // would (in general) introduce a potential side channel.
+   if(derived_byte == 0) {
+      std::cout << "I may or may not be printed." << std::endl;
+   }
+}
+
+void test_array_access_on_transitively_poisoned_data(Botan::RandomNumberGenerator& rng) {
+   const uint8_t poisoned_byte = rng.next_byte();
+   const uint8_t innocent_byte = rng.next_byte();
+   Botan::CT::poison(poisoned_byte);
+
+   const uint8_t derived_byte = poisoned_byte ^ innocent_byte;
+   const auto lookup_table = rng.random_array<1 << (sizeof(derived_byte) * 8)>();
+
+   // Performing a memory access on a value that was calculated from a "secret" value
+   // would introduce a potential side channel.
+   std::cout << lookup_table[poisoned_byte] << std::endl;
+}
+
+void test_unpoisen_transitively_poisoned_data(Botan::RandomNumberGenerator& rng) {
+   const uint8_t poisoned_byte = rng.next_byte();
+   const uint8_t innocent_byte = rng.next_byte();
+   Botan::CT::poison(poisoned_byte);
+
+   const uint8_t derived_byte = poisoned_byte ^ innocent_byte;
+   Botan::CT::unpoison(derived_byte);
+
+   // This is okay, because the value is not secret anymore.
+   if(derived_byte == 0) {
+      std::cout << "I may or may not be printed." << std::endl;
+   }
+}
+
+void test_poisoned_and_cleared_data(Botan::RandomNumberGenerator& rng) {
+   const uint8_t poisoned_byte = rng.next_byte();
+   Botan::CT::poison(poisoned_byte);
+
+   uint8_t derived_byte = poisoned_byte & 0x00;
+   derived_byte ^= rng.next_byte();  // To prevent the compiler from optimizing the whole thing away.
+
+   // This is okay, because the value was cleared (i.e. is not secret-dependent anymore)
+   if(derived_byte == 0) {
+      std::cout << "I may or may not be printed." << std::endl;
+   }
+}
+
+void test_conditional_branch_on_unpoisoned_bit(Botan::RandomNumberGenerator& rng) {
+   const auto poisoned_byte = rng.next_byte();
+   Botan::CT::poison(poisoned_byte);
+
+   // Clears the last poisoned bit (effectively unpoisoning it)
+   // All other bits are still poisoned (aka secret dependent)
+   uint8_t derived_byte = poisoned_byte & 0b11111110;
+
+   derived_byte ^= rng.next_byte();  // To prevent the compiler from optimizing the whole thing away.
+
+   // This conditional jump is okay, because the jump does not depend on the "secret" value.
+   if((derived_byte & 0b00000001) == 0) {
+      std::cout << "I may or may not be printed." << std::endl;
+   }
+}
+
+void test_conditional_branch_on_poisoned_bit(Botan::RandomNumberGenerator& rng) {
+   const auto poisoned_byte = rng.next_byte();
+   Botan::CT::poison(poisoned_byte);
+
+   // Clears all but one poisoned bit (effectively unpoisoning them)
+   // One bit remains poisoned (aka secret dependent)
+   uint8_t derived_byte = poisoned_byte & 0b00000010;
+
+   derived_byte ^= rng.next_byte() & 0b00000010;  // To prevent the compiler from optimizing the whole thing away.
+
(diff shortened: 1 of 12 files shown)


src/valgrind_selftest/valgrind_selftest.cpp
```diff
@@ -140,6 +141,57 @@ void test_conditional_branch_on_poisoned_bit() {
...
+   // Certain versions of Clang, namely 15, 16, 17 and 18 (maybe more) with
+   // specific optimization flags (-Os, -O1, -O2 -fno-vectorize, ...) do
+   // realize that `poisoned_mask` can only ever be all-zero or all-one and
+   // conditionally jump over the loop execution below.
+   //
+   // See: https://pqshield.com/pqshield-plugs-timing-leaks-in-kyber-ml-kem-to-improve-pqc-implementation-maturity/
+   const uint8_t poisoned_mask = -static_cast<uint8_t>(poisoned_byte & 1);
+   for(size_t i = 0; i < sizeof(output_bytes); ++i) {
+      output_bytes[i] &= poisoned_mask;
+   }
```
reneme, line 165: Given the right [*] compiler and flags, this actually triggers valgrind.

[*]
```bash
./configure.py --without-documentation --cc=clang --compiler-cache=ccache --build-tool=ninja --build-targets=static,valgrind_selftest --with-valgrind --minimized-build --enable-modules=auto_rng,hmac,sha2_64,system_rng --with-debug-info
ninja valgrind_selftest
valgrind --error-exitcode=2 --track-origins=yes ./botan_valgrind_selftest  clang_vs_bare_metal_ct_mask
```

src/valgrind_selftest/valgrind_selftest.cpp
```diff
@@ -140,6 +141,57 @@ void test_conditional_branch_on_poisoned_bit() {
...
+   std::array<uint8_t, 16> output_bytes;
+   std::memset(output_bytes.data(), 0x42, sizeof(output_bytes));
+
+   // Before the introduction of CT::value_barrier, this did generate a
+   // conditional jump when compiled with clang using certain compiler
+   // optimizations. See the test case above for further details.
+   auto poisoned_mask = Botan::CT::Mask<uint8_t>::expand(poisoned_byte & 1);
+   for(size_t i = 0; i < sizeof(output_bytes); ++i) {
+      output_bytes[i] = poisoned_mask.select(output_bytes[i], 0);
+   }
```
reneme, line 189: If you comment-out the gist of `CT::value_barrier()` this also triggers valgrind under the same conditions explained above: https://github.com/randombit/botan/pull/4182/commits/487eb5034c76eb2dd80b13f83d6a42b526da1179#r1666833485

src/valgrind_selftest/valgrind_selftest.py
```diff
@@ -0,0 +1,143 @@
...
+            # whether or not to run certain special-case valgrind tests.
+            with open(args.build_config_path, encoding="utf-8") as build_info_file:
+                build_info = json.load(build_info_file)
+                cc_compile_flags = build_info["cc_compile_flags"]
+                cc_macro = build_info["cc_macro"]
+
+                # This is certainly not an exhaustive list of compiler configurations
+                # that could be problematic for this specific test
+                if cc_macro == "CLANG" and "-Os" in cc_compile_flags:
+                    test_list.append(ValgrindTest("clang_vs_bare_metal_ct_mask", True))
```
reneme, line 124: As a sanity check, we could run `./botan_valgrind_selftest --list-special` and double-check that the "clang_vs_bare_metal_ct_mask" test is in this list. Also, we could take the info whether or not it is _expected to fail_ from there, rather than hard-coding it.

randombit:
Generally a good idea. I'll review fully later. For now, a bikeshed:

This should be named something like `ct_selftest` - that we are currently using `valgrind` for our constant time testing is an implementation detail. In fact I recently learned ASan also supports similar annotations, and might be a better option already. In the future I hope we'll be able to adopt a special purpose tool which could additionally detect use of variable time opcodes like divisions, as well as memory access/jumps which are conditional on the secret.

reneme:
>This should be named something like `ct_selftest`

Fair. Renamed the target (and the python steering script to `ct_selftest`), the actual cpp-file stayed with `valgrind_selftest.cpp` because there it's not just an implementation detail. :)

randombit:
> because there it's not just an implementation detail.

How so? Oh I see, because we have to see if we're running under valgrind or not. Let's just fix that

`ct_utils.h` should export a macro and a function

* `BOTAN_HAS_CT_POSION` - currently just set if `BOTAN_HAS_VALGRIND` is set
* `CT::ct_poison_has_effect()` returns `RUNNING_ON_VALGRIND` if `BOTAN_HAS_VALGRIND` otherwise `false`

Let's not make our abstractions leakier than they have to be :) there should be no need for this test to know how we've implemented the constant time checking.

In particular this would (in the future) allow us to run this same test using multiple tools, which will be helpful to check that they are doing what we expect.

reneme:
> Let's not make our abstractions leakier than they have to be :) there should be no need for this test to know how we've implemented the constant time checking.

Yeah, that's indeed a very valid point. We've implemented it as suggested, except that the define is called `BOTAN_CT_POISON_ENABLED` as `BOTAN_HAS_CT_POISON` did sound like `CT::poison()` wouldn't be defined at all; as is the convention for all other modules.

reneme:
As said in https://github.com/randombit/botan/pull/4202#issuecomment-2225475402, let's keep this open so that we can add another test that puts the new `poison()` abstractions introduced in #4197 through its paces.

src/ct_selftest/ct_selftest.cpp
```diff
@@ -0,0 +1,297 @@
...
+ * (C) 2024 Fabian Albert, René Meusel - Rohde & Schwarz Cybersecurity
+ *
+ * Botan is released under the Simplified BSD License (see license.txt)
+ */
+
+#include <iostream>
+
+#include <botan/internal/ct_utils.h>
+
+#if defined(BOTAN_CT_POISON_ENABLED)
```
randombit, line 16: I’d rather we compile everything regardless, with a small ifdef in main checking the macro

src/ct_selftest/ct_selftest.cpp
```diff
@@ -0,0 +1,297 @@
...
+   #include <botan/hex.h>
+   #include <botan/internal/fmt.h>
+
+   #include <functional>
+   #include <map>
+
+namespace {
+
+void test_conditional_jump_on_poisoned_data() {
+   Botan::AutoSeeded_RNG rng;
```
randombit, line 28: Let’s use System_RNG here - it should anyway be faster, and (hypothetically) some tools may have special understanding of `getentropy` etc

src/ct_selftest/ct_selftest.cpp
```diff
@@ -0,0 +1,297 @@
...
+      {"unpoison_transitive_poisoned", {SHOULD_SUCCEED, test_unpoisen_transitively_poisoned_data}},
+      {"poisoned_and_cleared", {SHOULD_SUCCEED, test_poisoned_and_cleared_data}},
+      {"conditional_jump_on_unpoisoned_bit", {SHOULD_SUCCEED, test_conditional_branch_on_unpoisoned_bit}},
+      {"conditional_jump_on_poisoned_bit", {SHOULD_FAIL, test_conditional_branch_on_poisoned_bit}},
+      {"regression_test_clang_vs_ct_mask", {SHOULD_SUCCEED, regression_test_conditional_jump_in_ct_mask}},
+   };
+
+   // These tests require special conditions to run. E.g. certain compiler
+   // versions and optimization flags.
+   std::map<std::string, Test> special_case_tests = {
```
randombit, line 241: I don’t understand why the two maps here

src/lib/math/bigint/bigint.h
```diff
@@ -800,7 +800,7 @@ class BOTAN_PUBLIC_API(2, 0) BigInt final {
        */
       void cond_flip_sign(bool predicate);
 
-#if defined(BOTAN_HAS_VALGRIND)
+#if defined(BOTAN_CT_POISON_ENABLED)
       void const_time_poison() const;
```
randombit, line 804: I think this has a merge conflict

src/ct_selftest/ct_selftest.cpp
reneme: Done. Instantiation moved to `main()`, as suggested in  https://github.com/randombit/botan/pull/4182#discussion_r1671811179

src/ct_selftest/ct_selftest.cpp
reneme: `ct_selftest.py` uses `ct_selftest --list` to detect which tests are available and runs them one-by-one using valgrind (assuming it to fail or not). The "special case tests" don't fail on all compilers with all flags, and hence need special treatment.

Alternatively, we could add another flag (next to "should fail") to signal that.

reneme: > Alternatively, we could add another flag (next to "should fail") to signal that.

That's what I did, along with a slight refactor in `ct_selftest.PY` to make adding new "special case tests" easier and fail the test if one forgets to do so.

randombit (approved):
Thanks, this is really helpful for improving assurance that our tests and CT techniques are doing what we expect

reneme:
Thanks for the approval, @randombit. I was working on some more tests, though. Sorry. 😅 

> As said in https://github.com/randombit/botan/pull/4202#issuecomment-2225475402, let's keep this open so that we can add another test that puts the new poison() abstractions introduced in https://github.com/randombit/botan/pull/4197 through its paces.

The low-level wrappers (for ranges and integers, as well as `scoped_poison()`) are now also tested in `ct_selftest`.

The higher-level abstractions (`_const_time_poison()`, `poison_all()`, `poison_range()`, and again `scoped_poison()`) are tested as a "normal" unit test. By creating a sentinel object that implements an observable side-effect in `_const_time_poison()`.

Note that particularly the `scoped_poison()` in `ct_selftest` would have been able to find the regression in `scoped_cleanup`. See #4202.

src/tests/test_ct_utils.cpp
```diff
@@ -177,4 +177,79 @@ class CT_Option_Tests final : public Test {
...
+               result.confirm("all not poisoned", !p1.poisoned && !p2.poisoned && !p3.poisoned);
+               Botan::CT::poison_all(p1, p2, p3);
+               result.confirm("all poisoned", p1.poisoned && p2.poisoned && p3.poisoned);
+               Botan::CT::unpoison_all(p1, p2, p3);
+               result.confirm("all unpoisoned", !p1.poisoned && !p2.poisoned && !p3.poisoned);
+            }),
+
+      CHECK("scoped poison",
+            [](Test::Result& result) {
+               // template is useless, but p1, p2, and p3 are different types and we
```
randombit, line 220: No p3 here

randombit (approved):

src/tests/test_ct_utils.cpp
reneme:
```suggestion
               // template is useless, but p1, and p3 are different types and we
```

randombit:
Squash plz

PR #4185, opened 2024-07-07
Update various documentation comments

randombit:
@reneme This is a merge of (as best I could handle after the various merge conflicts) the doc updates you made for Breathe. I'm pretty unconvinced of Breathe still (and especially since it seems to have been abandoned by the upstream dev now...) but the doc comments obviously have a lot of value no matter what. In retrospect I wish we had merged the doc comment updates immediately at the time, and deferred just the Breathe integration part, but so it goes.


randombit:
Feel free to take over this PR/open a new one with this as the base/force push/whatever

reneme:
Oh wow! Thanks for taking that apart.  I'll check it out soon. Agreed though, if breathe is abandoned by now, that's no good. 😔

src/lib/modes/cipher_mode.h
```diff
@@ -84,6 +84,16 @@ class BOTAN_PUBLIC_API(2, 0) Cipher_Mode : public SymmetricAlgorithm {
    public:
       /**
       * Begin processing a message with a fresh nonce.
+      * Start processing a message with a fresh nonce.
+      *
```
reneme, line 88: One should be enough...

```suggestion
      /**
      * Begin processing a message with a fresh nonce.
      *
```

src/lib/tls/credentials_manager.h
```diff
@@ -93,24 +93,23 @@ class BOTAN_PUBLIC_API(2, 0) Credentials_Manager {
...
-      * Return a cert chain we can use, ordered from leaf to root,
-      * or else an empty vector.
+      * Return a certificate chain we can use to identify ourselves, ordered
+      * from leaf to root, or else an empty vector.
       *
-      * This virtual function is deprecated, and will be removed in a
-      * future release. Use (and override) find_cert_chain instead.
+      * @deprecated This virtual function is deprecated, and will be removed in
+      *             a future release. Use (and override) find_cert_chain()
+      *             instead.
```
reneme, line 101: Clang warns that the docstring claims a deprecation but there's no `[[deprecated]]` attribute in the function's declaration.
This would be fixable by adding `BOTAN_DEPRECATED()`. However, when we build with the deprecation annotations being disabled (e.g. amalgamation builds), this clang warning might still show up and break things (due to `-Werror`).

I think we should either disable this clang warning (`-Wdocumentation-deprecated-sync`), or not use `@deprecated` in the Doxygen comments. This seems to be the only/first location where we use it anyway.

reneme: That said: clang has a point. We should actually mark this as deprecated in favor of `find_cert_chain()`. When invoked in a client-auth context, `find_cert_chain()` conveniently delivers information about the advertised acceptable CAs of the server.

src/lib/tls/credentials_manager.h
```diff
@@ -142,10 +151,14 @@ class BOTAN_PUBLIC_API(2, 0) Credentials_Manager {
...
-      * This function should either return null or throw an exception if
-      * the key is unavailable.
+      * Return a `shared_ptr` to the private key for this certificate. The
+      * @p cert will be the leaf cert of a chain returned previously by
+      * find_cert_chain() or cert_chain_single_type().
+      *
+      * @return private key associated with this certificate if we should use it
+      *         with this context. cert was returned by cert_chain This function
+      *         should either return null or throw an exception if the key is
+      *         unavailable.
```
reneme, line 161:
```suggestion
      * Return a `shared_ptr` to the private key for this certificate. The
      * @p cert will be the leaf cert of a chain returned previously by
      * find_cert_chain() or cert_chain_single_type().
      *
      * This function should either return nullptr or throw an exception if
      * the matching private key is unavailable.
      *
      * @return private key associated with this certificate if we should use it
      *         in this context.
```

src/lib/tls/tls_callbacks.h
```diff
@@ -45,50 +45,98 @@ class BOTAN_PUBLIC_API(2, 0) Callbacks {
...
-       * Again the array is overwritten immediately after the function returns.
-       *
-       * @param seq_no the underlying TLS/DTLS record sequence number
-       *
-       * @param data a contiguous data buffer containing the received record
-       */
+      * Mandatory callback: process application data
+      *
+      * Called when application data record is received from the peer. Again
+      * the array is overwritten immediately after the function returns.
```
reneme, line 78:
```suggestion
      * Called when application data record is received from the peer. The
      * array is overwritten immediately after the function returns.
```

... can't really assume that people read this cover-to-cover.

reneme (approved):

reneme:
Once this is merged, I'll close the various breathe PRs to reduce clutter. I agree that breathe doesn't seem to deliver on what we need. Nevertheless, it would be great to somehow interface between the sphinx and doxygen documentation to avoid duplicating too much content.

src/lib/tls/credentials_manager.h
randombit: I removed the `@deprecated` comment but also did introduce a `BOTAN_DEPRECATED` annotation here.

randombit:
> Nevertheless, it would be great to somehow interface between the sphinx and doxygen documentation to avoid duplicating too much content.

I do agree, there is a real need for something that is basically Doxygen+Sphinx combined (but ideally also faster than either...)

PR #4197, opened 2024-07-11
Convenience overloads for CT::poison()

reneme:
This introduces a few convenience helpers for `CT::poison()`. For instance

```C++
CT::poison(some_integer_type);
CT::poison(some_range);
CT::poison(some_type_that_has_method_const_time_poison);
CT::poison_range(a_list_of_big_ints); // iterates the list and poisons each range element individually
CT::poison_all(some_int, some_range, some_type_with_const_time_poison);

auto scope = CT::scoped_poison(m_private); // will be unpoisoned at the end of scope
```
... and of course the respective `CT::unpoison()` functions.

### DIFF
--- a/src/lib/utils/ct_utils.h
+++ b/src/lib/utils/ct_utils.h
@@ -14,9 +14,13 @@
 #ifndef BOTAN_CT_UTILS_H_
 #define BOTAN_CT_UTILS_H_
 
+#include <botan/concepts.h>
 #include <botan/secmem.h>
 #include <botan/internal/bit_ops.h>
+#include <botan/internal/stl_util.h>
+
 #include <optional>
+#include <span>
 #include <type_traits>
 
 #if defined(BOTAN_HAS_VALGRIND)
@@ -24,6 +28,9 @@
 #endif
 
 namespace Botan::CT {
+
+/// @name Constant Time Check Annotation Helpers
+/// @{
 
 /**
 * Use valgrind to mark the contents of memory as being undefined.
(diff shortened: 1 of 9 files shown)


src/lib/utils/ct_utils.h
```diff
@@ -64,15 +67,89 @@ constexpr inline void unpoison(const T* p, size_t n) {
...
+
+/**
+ * Unpoison a class type that provides a public `const_time_unpoison()` method
+ * For instance: BigInt, CT::Mask<>, FrodoMatrix, ...
+ */
+template <typename T>
+   requires requires(const T& x) { x.const_time_unpoison(); }
+constexpr void unpoison(const T& x) {
+   x.const_time_unpoison();
+}
```
reneme, line 116: We saw that `BitInt::const_time_poison()` was a thing already and just extended the idea to general types. This will be convenient for the PQC stuff in particular, as it often has helper-structures that could encapsulate their poisoning.

Ideally, this method should be `_const_time_poison()`, I guess, but we didn't want to break the existing `BigInt` API.

randombit: We could alternately name it `_const_time_poison` and on `BigInt` just define `const_time_poison` as a deprecated forward.

*Or* just define another instance and handle both cases --

```
template <typename T>
   requires requires(const T& x) { x._const_time_unpoison(); }
constexpr void unpoison(const T& x) {
   x._const_time_unpoison();
}
```

src/lib/utils/ct_utils.h
```diff
@@ -64,15 +67,89 @@ constexpr inline void unpoison(const T* p, size_t n) {
...
 
+//
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+//
+// Convenience overloads for poisoning/unpoisoning various types and ranges
+//
+
+template <std::integral T>
+constexpr void poison(T& p) {
+   poison(&p, sizeof(T));
```
randombit, line 78: The argument to `poison` is the number of `T` not the bytes

src/lib/utils/ct_utils.h
```diff
@@ -64,15 +67,89 @@ constexpr inline void unpoison(const T* p, size_t n) {
...
+}
+
+template <std::integral T>
+constexpr void unpoison(T& p) {
+   unpoison(&p, sizeof(T));
```
randombit, line 83: Should be 1

src/lib/utils/ct_utils.h
```diff
@@ -64,15 +67,89 @@ constexpr inline void unpoison(const T* p, size_t n) {
...
+}
+
+template <ranges::spanable_range R>
+constexpr void poison(R&& r) {
```
randombit, line 87: Why not just take a `span` as the argument?

src/lib/utils/ct_utils.h
```diff
@@ -64,15 +67,89 @@ constexpr inline void unpoison(const T* p, size_t n) {
...
+// Convenience overloads for poisoning/unpoisoning various types and ranges
+//
+
+template <std::integral T>
+constexpr void poison(T& p) {
+   poison(&p, sizeof(T));
+}
+
+template <std::integral T>
+constexpr void unpoison(T& p) {
```
randombit, line 82: There is earlier in the file a generic `unpoison(T&)` that should probably be removed - I think it only exists precisely for the case of unpoisoning an integer that was somehow derived from a secret.

reneme: We did remove this; or more precisely: restrict it to `std::integral` here. Otherwise the overloads might have been ambiguous.

In the codebase it was used for integers _and_ CT::Mask unpoisening. The latter is now handled via the new `CT::Mask::const_time_(un)poison` members.

src/lib/utils/ct_utils.h
reneme: That would mean to template on the span-ed type which does not work with implicit conversion. Like:

```C++
template <typename T>
constexpr void poison(std::span<T> s) {
   /* ... */
}
```

You can't just call that with a vector (say `std::vector<uint32_t> x`) like: `poison(x)`, because it won't deduce the `T` as the `std::ranges::range_value_t<decltype(x)>`.

The `spanable_range` concept restricts to every type for which `std::span{x}` would compile, thus making the conversion to span explicit.

src/lib/utils/ct_utils.h
randombit: Ah, oops, I missed that being deleted already in the diff

src/lib/utils/ct_utils.h
reneme: Good catch!

src/lib/utils/ct_utils.h
```diff
@@ -416,6 +493,10 @@ class Mask final {
       */
       constexpr T value() const { return value_barrier<T>(m_mask); }
 
+      constexpr void const_time_poison() const { CT::poison(m_mask); }
+
+      constexpr void const_time_unpoison() const { CT::unpoison(m_mask); }
```
randombit, line 498: Wait, this doesn't seem right. Are we poisoning/unpoisoning the masks currently somewhere?

src/lib/utils/ct_utils.h
reneme: I'll rename them to `_....` which is anyway the new convention.

src/lib/utils/ct_utils.h
reneme: Yes. Here:

https://github.com/randombit/botan/blob/1af5545ae66ab7c56c35780769722db85217e982/src/lib/pk_pad/iso9796/iso9796.cpp#L192
https://github.com/randombit/botan/blob/1af5545ae66ab7c56c35780769722db85217e982/src/lib/tls/tls12/tls_cbc/tls_cbc.cpp#L452

reneme:
I applied all review suggestions and tweaked the requirements for the overloads a little. Most notably: `poison(Range&&)` now requires that the range's members are trivially copyable. Otherwise a (hypothetical) range of complex objects would have unexpected outcomes. Say `poison(std::vector<std::vector<uint8_t>>)` would have worked and (probably) poisoned the inner vectors' meta-data, which is certainly not what we would have expected.

Additionally, I added `poison_range(R&&)` which iterates any range of poisonable objects and poisons them individually. I deliberately decided to _not make this an overload of `poison()`_ to avoid inadvertently binding to the wrong overload for some unforeseen reason.

src/lib/utils/ct_utils.h
randombit: Ah ok, fallout from not having `CT::Choice` for this kind of logic.

Fine to merge this then, it's just keeping status quo. These should be fixed though.

randombit (approved):

PR #4198, opened 2024-07-11
CT::poison() for FrodoKEM

reneme:
### Pull Request Dependencies

* #4197 

### Description

Our assumption was that poisoning the relevant (secret) values in the highest-level algorithms should be enough, as valgrind will propagate it down through the low-level implementations.
Before storing or returning, values are unpoisoned again. Anything that goes out of scope, anyway, will not be explicitly unpoisoned.

@randombit Some feedback on those assumptions would be appreciated.

### DIFF
--- a/src/lib/pubkey/frodokem/frodokem_common/frodo_matrix.h
+++ b/src/lib/pubkey/frodokem/frodokem_common/frodo_matrix.h
@@ -127,6 +127,10 @@ class FrodoMatrix {
 
       void reduce(const FrodoKEMConstants& constants);
 
+      constexpr void _const_time_poison() const { CT::poison(m_elements); }
+
+      constexpr void _const_time_unpoison() const { CT::unpoison(m_elements); }
+
    private:
       FrodoMatrix(const Dimensions& dimensions, secure_vector<uint16_t> elements) :
             m_dim1(std::get<0>(dimensions)), m_dim2(std::get<1>(dimensions)), m_elements(std::move(elements)) {}
--- a/src/lib/pubkey/frodokem/frodokem_common/frodokem.cpp
+++ b/src/lib/pubkey/frodokem/frodokem_common/frodokem.cpp
@@ -66,6 +66,10 @@ class FrodoKEM_PrivateKeyInternal {
 
       const FrodoMatrix& s_trans() const { return m_s_trans; }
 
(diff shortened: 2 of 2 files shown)


randombit (approved):
Frodo changes (455da8b6bc61830d5c2aa5e35697e48b5ea51fe0) approved

randombit:
> Our assumption was that poisoning the relevant (secret) values in the highest-level algorithms should be enough, as valgrind will propagate it down through the low-level implementations.

Yes just poisoning the top level secret values is best practice.

> Before storing or returning, values are unpoisoned again. Anything that goes out of scope, anyway, will not be explicitly unpoisoned.

This is all correct with the exception that if anything was heap allocated you have to unpoison explicitly. Otherwise there will be memory in the heap that valgrind is still considering undefined, this can cause really confusing errors.

reneme:
> if anything was heap allocated you have to unpoison explicitly.

I'm not sure I understand fully. Consider this:

```C++
secure_vector<uint8_t> decaps(PrivateKey& pk, std::span<const uint8_t> ciphertext) {
   auto scope = CT::scoped_poison(pk); // because private key will live on outside this scope

   std::vector<uint8_t> intermediate_result = ciphertext * pk; // some intermediate (that is transitively poisoned)

   std::vector<uint8_t> shared_secret = sha256(intermediate_result);
   CT::unpoison(shared_secret); // because we're handing this memory to the user

   /*
    * Here, intermediate_result is heap-allocated, but we're not unpoisoning it as it
    * will be deallocated at the end of this scope anyway.
    */

   return shared_secret;
}
```

... our assumption is that, `intermediate_result` won't need to be unpoisoned, despite it being allocated on the heap. When someone else allocates that same memory from the heap later, valgrind will need to handle those bits as "uninitialized" anyway, no?

Bottom line: Our mental model is: anything that lives on beyond the relevant scope needs to be unpoisoned. Anything that is deallocated or will be popped from the stack does not need to be unpoisoned.

... just making sure we're on the same page here.

randombit:
Not sure tbh.

What you say wrt deallocation makes sense. I just remember running into issues like this in the past. But possibly it was something similar but not quite the same, like a BigInt or vector being poisoned and then reused in a different context, or something like that.

I suppose the worst thing that happens is we start seeing sporadic valgrind errors, in which case we know to go looking. It wouldn't affect end users in any event.

PR #4202, opened 2024-07-12
FIX: move() of scoped_cleanup

reneme:
TIL: Moving an optional<> (as the default move c'tor would simply do) actually moves the contained object from `other&&` to this. But **IT DOES NOT reset() the other optional**. As a result, the d'tor of `other&&` (at the end of the move c'tor invocation) may or may not call the cleanup callback prematurely. Different scenarios and compilers produce different results. 🤯 

We unfortunately didn't find a way to write a reliable regression test for this.

### DIFF
--- a/src/tests/test_utils.cpp
+++ b/src/tests/test_utils.cpp
@@ -1232,6 +1232,8 @@ class UUID_Tests : public Test {
 
 BOTAN_REGISTER_TEST("utils", "uuid", UUID_Tests);
 
+#endif
+
 class Formatter_Tests : public Test {
    public:
       std::vector<Test::Result> run() override {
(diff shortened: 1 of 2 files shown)


src/tests/test_utils.cpp
```diff
@@ -1232,6 +1232,8 @@ class UUID_Tests : public Test {
 
 BOTAN_REGISTER_TEST("utils", "uuid", UUID_Tests);
 
+#endif
```
reneme, line 1235: Drive-by repair: this is the closing part of `#if defind(BOTAN_HAS_UUID)` and it enclosed too much code.

randombit (approved):
A fun one

reneme:
I mean, it totally makes sense to not mandate this additional store of a single (presumed) bool flag in the standard, I guess. After all its an rvalue ref and typically that's going away anyway. The performance gain must have been amazing. We're all happy to pay for it in years and years of debugging fun and obscure bugs. </irony>

... We'll look into extending the `ct_selftest` to actually test the `poison()` abstractions to have a higher probability to find things like that.

Issue #4196, opened 2024-07-11
Some build flags are not passed to link-time optimizer when building with `-flto=auto`


mrc0mmand:
While preparing another batch of botan2 updates in Fedora, I noticed a failing check in `annocheck`:

```
$ annocheck /usr/bin/botan 
annocheck: Version 12.46.
Hardened: botan: FAIL: stack-prot test because insufficient protection enabled (based on contents of DW_AT_producer string) 
Hardened: Rerun annocheck with --verbose to see more information on the tests.
Hardened: botan: Overall: FAIL.
```

Making `annocheck` a bit more verbose shows where this information comes from:
```
$ annocheck -v -v /usr/bin/botan                                                                                                                                                                                    
annocheck: Version 12.46.  
...
Hardened: /usr/bin/botan: DW_AT_producer = GNU GIMPLE 13.2.1 20230728 (Red Hat 13.2.1-1) -mno-omit-leaf-frame-pointer -m64 -mtune=generic -march=x86-64 -g -O2 -fno-openmp -fno-openacc -fasynchronous-unwind-tables
 -fcf-protection=full -fPIE -fstack-protector -fltrans -fplugin=annobin.                                  
Hardened: /usr/bin/botan: info: record producer: Gimple version: 13 source: DW_AT_producer string.                                                                                                                  
Hardened: /usr/bin/botan: info: set binary producer to Gimple version 13.                                                                                                                                           
Hardened: /usr/bin/botan: PASS: pic test because option found in DW_AT_producer string  (source: DW_AT_producer string)                                                                                             
Hardened: /usr/bin/botan: FAIL: stack-prot test because insufficient protection enabled (based on contents of DW_AT_producer string) (source: DW_AT_producer string)                                                
Hardened: /usr/bin/botan: info: For more information visit: https://sourceware.org/annobin/annobin.html/Test-stack-prot.html                                                    
...
```

With this information I was also able to reproduce this with a reduced set of build flags on the latest main (i.e. botan3):
```
$ CXXFLAGS="-fstack-protector-strong -O2 -g -flto=auto" ./configure.py
...
$ make -j $(nproc)
...
$ objdump -Wi botan | grep DW_AT_producer 
[...]

PR #4206, opened 2024-07-12
Reflect CXXFLAGS into the linker invocation as well

randombit:
This seems to better match the behavior of other build systems, and apparently at least some packaging setups expect it to work this way.

GH #4196


reneme:
Ah, looked at this too late, sorry. I think we should apply this to the Ninja template as well, to match the behaviour.

randombit:
Oh indeed

PR #4207, opened 2024-07-12
Adopt new convention for CT poisoning in pcurves

randombit:


src/lib/math/pcurves/pcurves_impl.h
```diff
@@ -831,16 +831,16 @@ class ProjectiveCurvePoint {
...
       constexpr const FieldElement& z() const { return m_z; }
 
-      constexpr void ct_poison() const {
-         x().ct_poison();
-         y().ct_poison();
-         z().ct_poison();
+      constexpr void _const_time_poison() const {
+         CT::poison(m_x);
+         CT::poison(m_y);
+         CT::poison(m_z);
```
reneme, line 837: For short, `CT::poison_all(m_x, m_y, m_z)` would also work. Same for unpoison, of course.

reneme (approved):
Nice!

reneme (approved):

PR #4213, opened 2024-07-15
Faster scalar inversions for P-384 and P-521

randombit:
#1479


src/lib/math/pcurves/pcurves_impl/pcurves_wrap.h
```diff
@@ -12,6 +12,18 @@
...
+struct CurveHasFeInvert2 : std::false_type {};
+
+template <typename C>
+struct CurveHasFeInvert2<C, decltype(&C::fe_invert2, 0)> : std::true_type {};
+
+template <typename C, typename = int>
+struct CurveHasScalarInvert : std::false_type {};
+
+template <typename C>
+struct CurveHasScalarInvert<C, decltype(&C::scalar_invert, 0)> : std::true_type {};
```
reneme, line 25: Just as [in the other pull request](https://github.com/randombit/botan/pull/4209#discussion_r1677433043); let's make this a concept:

```C++
template <typename C>
concept supports_fast_scalar_inversion = requires(const typename C::Scalar& fe) {
   { C::scalar_invert(fe) } -> std::same_as<typename C::Scalar>;
};
```

reneme (approved):

src/lib/math/pcurves/pcurves_impl/pcurves_wrap.h
```diff
@@ -17,6 +17,11 @@ concept curve_supports_fe_invert2 = requires(const typename C::FieldElement& fe)
    { C::fe_invert2(fe) } -> std::same_as<typename C::FieldElement>;
 };
 
+template <typename C>
+concept curve_supports_scalar_invert = requires(const typename C::FieldElement& fe) {
+   { C::scalar_invert(fe) } -> std::same_as<typename C::FieldElement>;
+};
```
reneme, line 23:
```suggestion
template <typename C>
concept curve_supports_scalar_invert = requires(const typename C::Scalar& s) {
   { C::scalar_invert(s) } -> std::same_as<typename C::Scalar>;
};
```

Copy&Paste issue? Note that this would just silently disable your optimized code. There won't be a compiler error; the concept just wouldn't fit.

randombit: Eep this would explain why I was suddenly struggling to repro the improvements on my other machine ... thank you :sweat_smile:

reneme: It's all internal APIs, but perhaps a smoke test would be a good idea. 😏 I tested the concepts locally with `std::cout` in the respective branches.

PR #4214, opened 2024-07-15
Followup on #4206 for Ninja build

randombit:
Update the Ninja template to also reflect compliation flags into the linker invocation.


reneme (approved):
Merci.

PR #4216, opened 2024-07-15
Close the multiprocessing ThreadPool

randombit:
Possible fix for #3991

We'll only know if it works if we merge this and a few weeks go by without the crash recurring...

### DIFF
--- a/src/scripts/test_cli.py
+++ b/src/scripts/test_cli.py
@@ -1921,13 +1921,13 @@ def main(args=None):
     start_time = time.time()
 
     if threads > 1:
-        pool = ThreadPool(processes=threads)
-        results = []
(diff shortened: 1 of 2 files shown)


src/scripts/test_cli.py
```diff
@@ -1928,6 +1928,7 @@ def main(args=None):
 
         for result in results:
             result.get()
+        pool.close()
```
reneme, line 1931: I believe, `TreadPool` implements a "context manager". I.e. you could use this along those lines:

```python
with ThreadPool() as pool:
   # ...

randombit: Done

reneme (approved):
# 🤞

PR #4217, opened 2024-07-15
Refactor pcurves affine conversions and inversion usage

randombit:
We did not consistently take advantage of any optimized addition chain for inversions, particularly in table setup and hash to curve code.

Also refactor hash to curve such that we avoid a pointless affine conversion in the NU case; previously we returned a projective point even in that case, then later converted it to affine.


randombit:
@reneme ping

Issue #4027, opened 2024-04-20
Replace BigInt based elliptic curve library


randombit:
* [ ] Initial pcurves (point arithmetic, fixed curve params) - that's #3979
* [ ] Support for providing parameterized curves, where we eg compute Montgomery params at runtime. This is required not just for user provided/application specific curves but also I don't think it's worthwhile to provide the fully parameterized/hardcoded support for obscure curves like secp160r1.
* [ ] Bridge between our public APIs (`EC_Point`, `EC_Group`) and pcurves.
* [ ] Implement our EC public key code in terms of pcurves. Probably one algorithm at a time initially.
* [ ] Deprecate all the functionality that existed just to support ellptic curves using `BigInt`, eg `mod_sub`, `ct_reduce_below`, many more.

reneme:
We should consider compile time configuration of which curves get instantiated (as pointed out [here](https://github.com/randombit/botan/pull/3979#discussion_r1617098503)). Much like the compile-time configuration of having "kyber" and/or "kyber_90s", for instance.

Adding one build module per curve should do the trick, and probably provides the best affordance for users. On the other hand, it produces quite a bit of boilerplate due to subdirectories, `info.txt` files and code machinery.

Alternatively, I could see this as a dedicated compile-time switch: Like:

* `./configure.py --enable-elliptic-curves=all` (the default)
* `./configure.py --enable-elliptic-curves=p256,brainpool256` (a restricted set)

and then simply `#ifdef` the instantiations in `pcurves.cpp` accordingly.

randombit:
The `--enable-elliptic-curves` approach is potentially confusing since even if you disable the pcurves module entirely, the elliptic curves still exist. All that ends up happening is you fall back to the slower BigInt code. [*]

[*] I mean right now ECDSA etc still use BigInt, even after both #3979 and #4042 land. There are several more steps here until the whole thing hangs together. But in the end state, we'll have fast ECC for specific curves, plus fallback code for weird curves, application curves, etc. If you disable the fast curve, it doesn't disable the curve, just disables it going fast.

That said there may be environments where the code size becomes an issue. I tried out a module per curve and it was not so bad. There are likely to not be more than ~5 new curves added here over time so I don't think it's an issue in an ongoing way.

In the end we could consider also having a way of disabling the slowpath BigInt stuff, which would benefit environments that are using *just* P-256 or something.

reneme:
[...]

PR #4218, opened 2024-07-16
Add pcurves P-224

randombit:
This one is a little more involved since this is the first curve where p != 3 mod 4, and in fact since P-224 is == 1 mod 16 we must use Shanks-Tonelli

GH #4027


randombit:
Emscripten failure seems to be a Homebrew/Emscripten mess https://github.com/emscripten-core/emscripten/issues/22249 hopefully addressed soon but not related to the change

randombit:
@reneme Ping. Wasn't sure if you cared about this one. Main point I'd be interested in more eyes is the Shanks-Tonelli.

PR #4220, opened 2024-07-16
Add more cross compilers to CI

randombit:


reneme (approved):
Any motivation here, or just trying to build on as many configurations as possible?

randombit:
Somewhat just "because it's there" but also two other points

1) There have been some cases in the past where GCC on weird architectures miscompiled us. Typically I'd only find out by looking at something like Debian build logs and seeing eg that the last 3 releases have crashed during testing on m68k, or something like that. This gives us some early warning if this occurs.

2) It seems `constexpr` limits are somewhat implicitly not just compiler specific but also architecture dependent; I've seen numerous cases now pushing constexpr pretty hard, where everything worked on my machine, but then failed eg on aarch64 build (with the same GCC version) with a constexpr timeout. I imagine it's just some side effect of how the evaluation steps are implemented. So this is to somewhat limit the chances that I accidentally do something that is un-compilable on a specific architecture. (I guess we could start messing with the various constexpr limits flags compilers expose, but this seems fragile, especially for amalgamation users.)

PR #4221, opened 2024-07-16
Add fast iterated point doubling

randombit:
Improves ECDH performance for curves with A == -3 by 5-9%

#4027

### DIFF
--- a/src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
+++ b/src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
@@ -176,7 +176,7 @@ class IntMod final {
          return Self(Rep::redc(z));
       }
 
-      void square_n(size_t n) {
+      constexpr void square_n(size_t n) {
          std::array<W, 2 * N> z;
          for(size_t i = 0; i != n; ++i) {
             comba_sqr<N>(z.data(), this->data());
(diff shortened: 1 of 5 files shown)


src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
```diff
@@ -698,14 +705,48 @@ class ProjectiveCurvePoint {
       }
 
       constexpr Self dbl_n(size_t n) const {
-         // TODO it is possible to optimize this by carrying over values from
-         // the previous iteration into the next
+         if constexpr(Self::A_is_minus_3) {
```
reneme, line 708: In case you want to express this as a good-old function overload, I think you could do this:

```C++
constexpr Self dbl_n(size_t n const)
   requires(Self::A_is_minus_3)
{
   // ...
}

constexpr Self dbl_n(size_t n const)
   requires(!Self::A_is_minus_3)
{
   // ...
}
```

... to avoid the `if constexpr` spanning the entirety of the function. No strong opinion on what is more idiomatic from my side

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
```diff
@@ -698,14 +705,48 @@ class ProjectiveCurvePoint {
...
+            /*
+            Repeated doubling for A == -3
+            Cost: 2S + *2 + n*(4S + 4M + 2*2 + 1*3 + 4A)
```
reneme, line 711:
```
2S + *2 + n*(4S + 4M + 2*2 + 1*3 + 4A)
      ^ typo?
```

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
```diff
@@ -698,14 +705,48 @@ class ProjectiveCurvePoint {
...
+            Algorithm 3.23 in
+            "Guide To Elliptic Curve Cryptography"
+            Hankerson, Menezes, Vanstone
+            */
+
+            auto nx = x();
+            auto ny = y();
+            auto nz = z();
+            ny = ny.mul2();
+            auto w = nz.square().square();
```
reneme, line 726: Does it make sense to early-return on `n == 0`?

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
```diff
@@ -530,6 +530,13 @@ class AffineCurvePoint {
       FieldElement m_y;
 };
 
+template <typename FE>
+const FE& inv_2() {
+   // TODO could be a constexpr, but some compilers get cranky
+   static const FE INV_2 = FE::from_word(2).invert();
+   return INV_2;
+}
```
reneme, line 538: 😞 

Perhaps it's worth allow-listing the compilers that are happy to do this as consteval?

Like so:

```C++
#if GOOD_BOI_COMPILER
template <typename FE>
consteval FE inv_2() {
   return FE::from_word(2).invert();
}
#else
// ... the runtime lazy-evaluation
#endif
```

reneme (approved):

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
randombit: Unfortunately it seems to be not just compiler type to type but for example GCC 14 on i386 bad, GCC 14 on x86-64 OK ... I think the best we could do would be doing constexpr on specific known compilers/versions that can handle it. Which might still be worthwhile

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
randombit: I expect in the future this will expand to cover other types of curves (A == 0 and/or generic A) in which case the single function will be cleaner.

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
randombit: Implicit one *2 but yeah everywhere else that is explicit

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
randombit: Not really in practice - we call this with `n` derived from window sizes so `n` is always > 0

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
randombit: Aha just figured out a nice trick that makes this fast for constexpr - if P is prime then x=(P/2)+1 is the inverse of 2 modulo P

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
randombit: Fixed

PR #4228, opened 2024-07-18
Use Linux for Emscripten build

randombit:
Homebrew's emscripten is broken and it's impossible for me to report this fact to them.


randombit:
24.04 ships Emscripten 3.1.6 which was released in Feb 2024

Ubuntu ... when only the most thoroughly obsolete software will do.

PR #4229, opened 2024-07-18
Disable Emscripten build

randombit:
Emscripten 2.1.63 is broken


randombit:
Tried, Ubuntu 24.04's Emscripten is 2 years old and compilation fails :disappointed:

reneme:
Oh, yeah. Just saw your attempt. 🤣

PR #4230, opened 2024-07-18
Workaround: Older version of Emscripten

reneme:
For science! Let's see whether we could just install an older version of `emscripten`.

According to [this tutorial](https://cmichel.io/how-to-install-an-old-package-version-with-brew/), we should be able to install an older version of emscripten, by:

1. Cloning the entire homebrew/core repo, which is about 800MB
2. Casting some brew-spell to get access to the older recipe
3. Installing the older version, which compiles from source LLVM...

For the record: I don't think that's feasible for a workaround. Just toying around with this.


randombit:
:joy:

reneme:
Less than an hour... given that this is a nightly job.

Your call...

randombit:
Yeah let's do it. Thanks for giving this a try. I would definitely have guessed it would be more like 6+ hours.

It seems like Emscripten releases often (if on an irregular schedule); I'm hoping in a few weeks we can just pick up a correct release once it makes it into homebrew.

randombit:
Sounds like a new version will be released next week https://github.com/emscripten-core/emscripten/issues/22249

randombit (approved):

randombit:
So somehow 

a) Github Actions *DID NOT* run the nightly flow here
b) This broke the nightly flow

:disappointed:

reneme:
> a) Github Actions DID NOT run the nightly flow here

I tested it in `ci.yml` and then just re-enabled the `nightly.yml` but apparently screwed up the indentation.

Mhh, I just saw the

``
on:
  push:
    paths:
      # Run if a pull request changes this workflow to
      # validate it works properly before merging.
      - '.github/workflows/nightly.yml'
```

in the nightly. And, I agree, this should have run when I changed the nightly workflow.

reneme:
Oh, and in fact it tried: https://github.com/randombit/botan/actions/runs/9997453932

But apparently "one of many broken workflows" still gives you the all-clear mark? 🤔

Issue #4231, opened 2024-07-18
Ninja install places .cmake files in the wrong folder.


learn-more:
Reproduction steps (Win10, VS2022, Python 3.12):

- Open `x64 Native Tools Command Prompt for VS 2022`
- Navigate to the folder where `botan` is cloned (`git describe` results in `1.11.30-9833-gb8af12b5c`)
- `python3 configure.py --cc=msvc --os=windows --build-tool=ninja --prefix=build/install --disable-shared-library`
- `ninja install`

Observe the placement of the cmake files:
```
> tree build\install
BUILD\INSTALL
├───bin
├───build
│   └───install
│       └───lib
│           └───cmake
│               └───Botan-3.6.0
├───docs
│   └───botan-3.6.0
│       └───handbook
│           ├───api_ref
│           └───dev_ref
├───include
│   └───botan-3
│       └───botan
└───lib
```
They are placed in `build/install/build/install/lib` instead of in `build/install/lib`.

This seems to originate from:
https://github.com/randombit/botan/blob/b8af12b5cd515b133500ccdc1eb04c6cd341dea4/src/scripts/install.py#L214

The `lib_dir` already includes the `prefix`, so this `os.path.join` should probably not add `lib_dir`.

reneme:
Confirmed! Thanks. As a workaround: if you specify the `--prefix` as an absolute path, it works as expected. I'm guessing `os.path.join()` realizes that the absolute base path in `lib_dir` and `prefix` are the same and "does the right thing".

learn-more:
Thanks for the quick fix!

PR #4236, opened 2024-07-19
FIX: install location of pkgconfig and cmake configs

reneme:
As reported in #4231. Turns out, the same is true for the pkgconfig file. I'm guessing, we didn't catch that, because for an absolute `--prefix=` path this works as expected.


learn-more (approved):
This fix looks like the one I did in my local copy (minus the moving around), so from an end-user perspective this 👍

randombit (approved):

PR #4238, opened 2024-07-20
Refactor internal usage of PK encryption padding

randombit:
Revamp EME class with std::span and CT::Option

### DIFF
--- a/src/lib/pk_pad/eme_oaep/oaep.cpp
+++ b/src/lib/pk_pad/eme_oaep/oaep.cpp
@@ -1,6 +1,6 @@
 /*
 * OAEP
-* (C) 1999-2010,2015,2018 Jack Lloyd
+* (C) 1999-2010,2015,2018,2024 Jack Lloyd
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
@@ -18,38 +18,45 @@ namespace Botan {
 /*
 * OAEP Pad Operation
 */
-secure_vector<uint8_t> OAEP::pad(const uint8_t in[],
-                                 size_t in_length,
-                                 size_t key_length,
-                                 RandomNumberGenerator& rng) const {
+size_t OAEP::pad(std::span<uint8_t> output,
+                 std::span<const uint8_t> input,
+                 size_t key_length,
+                 RandomNumberGenerator& rng) const {
    key_length /= 8;
 
-   if(in_length > maximum_input_size(key_length * 8)) {
+   if(input.size() > maximum_input_size(key_length * 8)) {
       throw Invalid_Argument("OAEP: Input is too large");
    }
 
-   secure_vector<uint8_t> out(key_length);
-   BufferStuffer stuffer(out);
+   const size_t output_size = key_length;
+
+   output = output.first(output_size);  // remainder ignored
+
+   BufferStuffer stuffer(output);
 
    // We always use a seed len equal to the underlying hash
    rng.randomize(stuffer.next(m_Phash.size()));
    stuffer.append(m_Phash);
-   stuffer.append(0x00, stuffer.remaining_capacity() - (1 + in_length));
+   stuffer.append(0x00, stuffer.remaining_capacity() - (1 + input.size()));
    stuffer.append(0x01);
-   stuffer.append({in, in_length});
+   stuffer.append(input);
    BOTAN_ASSERT_NOMSG(stuffer.full());
 
-   mgf1_mask(*m_mgf1_hash, out.data(), m_Phash.size(), &out[m_Phash.size()], out.size() - m_Phash.size());
+   const size_t hlen = m_Phash.size();
 
-   mgf1_mask(*m_mgf1_hash, &out[m_Phash.size()], out.size() - m_Phash.size(), out.data(), m_Phash.size());
+   mgf1_mask(*m_mgf1_hash, output.first(hlen), output.subspan(hlen));
 
-   return out;
+   mgf1_mask(*m_mgf1_hash, output.subspan(hlen), output.first(hlen));
+
+   return key_length;
 }
 
 /*
 * OAEP Unpad Operation
 */
-secure_vector<uint8_t> OAEP::unpad(uint8_t& valid_mask, const uint8_t in[], size_t in_length) const {
+CT::Option<size_t> OAEP::unpad(std::span<uint8_t> output, std::span<const uint8_t> input) const {
+   BOTAN_ASSERT_NOMSG(output.size() >= input.size());
+
    /*
    Must be careful about error messages here; if an attacker can
    distinguish them, it is easy to use the differences as an oracle to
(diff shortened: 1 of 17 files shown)


src/lib/pubkey/pk_ops.cpp
```diff
@@ -45,7 +46,13 @@ secure_vector<uint8_t> PK_Ops::Decryption_with_EME::decrypt(uint8_t& valid_mask,
                                                             const uint8_t ciphertext[],
                                                             size_t ciphertext_len) {
    const secure_vector<uint8_t> raw = raw_decrypt(ciphertext, ciphertext_len);
-   return m_eme->unpad(valid_mask, raw.data(), raw.size());
+
+   secure_vector<uint8_t> ctext(raw.size());
+   auto len = m_eme->unpad(ctext, raw);
+
+   valid_mask = CT::Mask<uint8_t>::from_choice(len.has_value()).if_set_return(0xFF);
+   ctext.resize(len.value_or(0));
```
reneme, line 54: That strikes me as dangerous!
If the `CT::Option` is set, this is probably not a (de)allocation, because the size of the initial allocation was -- more or less -- correct). If the `CT::Option` is not set, this might deallocate the vector's memory (as the `resize(0)` suggests).

Couldn't this turn into a padding oracle?

reneme: In the long run, it would probably be better to not return a vector but let the user provide enough memory in an out-param and just return the number of bytes written, I guess.

reneme (commented):
Yay!

src/lib/pubkey/pk_ops.cpp
randombit: Wait `resize(0)` can cause a free? :(

randombit: “Vector capacity is never reduced when resizing to smaller size because that would invalidate all iterators” - https://en.cppreference.com/w/cpp/container/vector/resize

randombit: (And we’re always shrinking, never growing - indeed a reallocation would cause an oracle)

src/lib/pk_pad/eme_raw/eme_raw.cpp
```diff
@@ -1,29 +1,40 @@
...
-                                    RandomNumberGenerator& /*rng*/) const {
-   return secure_vector<uint8_t>(in, in + in_length);
+size_t EME_Raw::pad(std::span<uint8_t> output,
+                    std::span<const uint8_t> input,
+                    size_t key_length,
+                    RandomNumberGenerator& rng) const {
+   BOTAN_UNUSED(rng);
+   BOTAN_ASSERT_NOMSG(input.size() < maximum_input_size(8 * key_length));
+   BOTAN_ASSERT_NOMSG(output.size() >= input.size());
+   copy_mem(std::span{output.data(), input.size()}, input);
```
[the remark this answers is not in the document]
randombit: I forgot `output` was already a span here 😅

src/lib/utils/ct_utils.cpp
```diff
@@ -52,46 +58,55 @@ secure_vector<uint8_t> copy_output(CT::Mask<uint8_t> bad_input_u8,
...
+
+   /*
+   This is potentially not const time, depending on how std::vector is
+   implemented. But since we are always reducing length, it should
+   just amount to setting the member var holding the length.
+   */
+   output.resize(written.value_or(0));
+
+   return output;
 }
```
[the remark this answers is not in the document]
randombit: I would take it up seperately but yeah returning a vector in this way is not ideal.

src/lib/pubkey/pk_ops.cpp
reneme: Mhh, okay. That sounds authoritative enough to rely on. Perhaps leave a comment about this, though. And/or a `BOTAN_DEBUG_ASSERT(raw.size() >= len.value_or(0))` to really make this clear. 😅

src/lib/pk_pad/eme_oaep/oaep.cpp
```diff
@@ -70,39 +77,39 @@ secure_vector<uint8_t> OAEP::unpad(uint8_t& valid_mask, const uint8_t in[], size
...
+      return CT::Option<size_t>();
+   }
+
+   auto scope = CT::scoped_poison(input);
+
+   const auto has_leading_0 = CT::Mask<uint8_t>::is_zero(input[0]).as_choice();
 
-   secure_vector<uint8_t> input(in + 1, in + in_length);
+   secure_vector<uint8_t> decoded(input.begin() + 1, input.end());
+   auto buf = std::span{decoded};
```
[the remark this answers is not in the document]
randombit: Not really for constant time concerns; with or without `BufferSlicer` we're at the mercy of the compiler and are relying on tooling to detect issues. I do find that here `BufferSlicer` obscures more than it helps (IMO); `slicer_take.byte()` / `slicer.remaining()` obscure the fixed indices being used.

src/lib/pk_pad/eme_pkcs1/eme_pkcs.cpp
```diff
@@ -11,65 +11,69 @@
...
 
-   bad_input_m |= ~CT::Mask<uint8_t>::is_equal(in[0], 0);
-   bad_input_m |= ~CT::Mask<uint8_t>::is_equal(in[1], 2);
+   bad_input_m |= ~CT::Mask<uint8_t>::is_equal(input[0], 0);
+   bad_input_m |= ~CT::Mask<uint8_t>::is_equal(input[1], 2);
 
-   for(size_t i = 2; i < inlen; ++i) {
-      const auto is_zero_m = CT::Mask<uint8_t>::is_zero(in[i]);
+   for(size_t i = 2; i < input.size(); ++i) {
+      const auto is_zero_m = CT::Mask<uint8_t>::is_zero(input[i]);
```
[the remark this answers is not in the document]
randombit: As with OAEP it seems easier to parse to me with fixed/visible index rather than being implicit in the `BufferSlicer` internal state

src/lib/pk_pad/eme_oaep/oaep.cpp
```diff
@@ -70,39 +77,39 @@ secure_vector<uint8_t> OAEP::unpad(uint8_t& valid_mask, const uint8_t in[], size
...
-
-   size_t delim_idx = 2 * hlen;
+   size_t delim_idx = 2 * phash.size();
    CT::Mask<uint8_t> waiting_for_delim = CT::Mask<uint8_t>::set();
    CT::Mask<uint8_t> bad_input_m = CT::Mask<uint8_t>::cleared();
 
-   for(size_t i = delim_idx; i < input_len; ++i) {
+   for(size_t i = delim_idx; i < input.size(); ++i) {
       const auto zero_m = CT::Mask<uint8_t>::is_zero(input[i]);
       const auto one_m = CT::Mask<uint8_t>::is_equal(input[i], 1);
```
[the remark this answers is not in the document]
randombit: Done (phrased slightly differently)

reneme (approved):

PR #4244, opened 2024-07-22
Use array-based load_be/le in MD hashes

reneme:
Using the `load_le`/`load_be` overloads with statically sized ranges results in less code ~~(and also less runtime checks)~~ (not compared to the raw-ptr overloads). ... please excuse my indentation OCD. 😏

Quick measurement doesn't show any accidents (in MiB/s).

|            | master    | Chore/mdx_revamp |
| ---------- | --------- | ---------------- |
| MD4        | 545.190   | 547.178          |
| MD5        | 305.430   | 304.652          |
| SHA-1      | 341.379   | 339.967          |
| SHA-256    | 180.383   | 180.365          |
| SHA-512    | 277.988   | 278.101          |
| RIPEMD-160 | 270.644   | 270.539          |
| SM3        | 163.148   | 163.073          |


src/lib/hash/sha2_32/sha2_32.cpp
```diff
@@ -68,95 +68,84 @@ void SHA_256::compress_digest(digest_type& digest, std::span<const uint8_t> inpu
    uint32_t A = digest[0], B = digest[1], C = digest[2], D = digest[3], E = digest[4], F = digest[5], G = digest[6],
             H = digest[7];
 
+   std::array<uint32_t, 16> W;
+
    BufferSlicer in(input);
 
    for(size_t i = 0; i != blocks; ++i) {
```
randombit, line 75: There is also the BMI2 SHA-2 which should probably be updated for parity

randombit (approved):

src/lib/hash/sha2_32/sha2_32.cpp
reneme: Thanks. Done.

PR #4245, opened 2024-07-22
FIX: ./configure.py --with-build-dir=

### LINKED ISSUE #4232: configure.py drops build.ninja in the repository root
Reproduction steps (Win10, VS2022, Python 3.12):

- Open `x64 Native Tools Command Prompt for VS 2022`
- Navigate to the folder where `botan` is cloned (`git describe` results in `1.11.30-9833-gb8af12b5c`)
- `python3 configure.py --cc=msvc --os=windows --build-tool=ninja --prefix=build/install --disable-shared-library`

The `build.ninja` is placed in the repository root instead of in the `build` folder

reneme:
Running `./configure.py --with-build-dir=mybuild` [didn't work properly](https://github.com/randombit/botan/issues/4232#issuecomment-2239058582), because it produced a Makefile/ninja.build with paths relative to the base-directory of the `configure.py` invocation (i.e. `argv[0]`). It did work when invoking it like that: `$(pwd)/configure.py --with-build-dir=mybuild`.

The idea behind this patch is to always produce absolute paths in the Makefile/ninja.build. As a side effect, this allows us to remove a mingw workaround (I think). Also, I cleaned up the ninja.build of `botan_bogo_shim` and `botan_ct_selftest` that both dropped their intermediate object files in the top-level build directory.

Closes #4232


configure.py
```diff
@@ -3555,7 +3548,7 @@ def main(argv):
 
     setup_logging(options)
 
-    source_paths = SourcePaths(os.path.dirname(argv[0]))
+    source_paths = SourcePaths(os.path.dirname(os.path.realpath(__file__)))
```
reneme, line 3551: This caused the explained issue. `argv[0]` may be a valid relative path (seen from `$($pwd)`), but it is not valid when `--with-build-dir=` drops the generated build files somewhere else and paths in there are relative to whatever `argv[0]` was.

reneme:
Rebased to master (to resolve a conflict with #4336).

reneme:
Turns out that listing all the object files with their absolute path [can overwhelm Windows' upper limit on command lengths](https://github.com/randombit/botan/issues/4342#issuecomment-2339857706). I'll have a look into that as soon as I can.

reneme:
Rebased after #4350 was merged.

reneme:
> Turns out that listing all the object files with their absolute path can overwhelm Windows' upper limit on command lengths. I'll have a look into that as soon as I can.

... okay. One year later. 🫨 That slipped, big time. Anyway, I recently needed `--with-build-dir=` for some Conan shenanigans and realized that it still doesn't work as expected on master. So here we are.

@randombit This should now finally be ready to go in. I added one more commit that uses relative paths for object files, relative to the `Makefile` or `build.ninja`. That should hopefully fix the remaining "path too long" issues on Windows.

Issue #3703, opened 2023-09-16
"WARNING: The windows target is oriented towards MSVC; maybe you want --os=cygwin or --os=mingw"


oviano:
So what does this mean in the context of building with the clang-cl compiler optionally installed with MSVC?

The only side-effect I have noticed so far since switching from the MSVC cl.exe compiler to clang-cl.exe was the need to explicitly link the executable with clang_rt.builtins-x86_64.lib so perhaps the warning above can be ignored in this context?

Or am I supposed to be telling Botan that the CC is still msvc, when taking this approach?

randombit:
Interesting question! I think this warning is now obsolete; we previously would do some feature enablement based on the target being Windows vs Cygwin, but that's been all removed now I think.

Probably for `clang-cl` the correct target is `windows`.

I'll look into if we can create a CI target for this to make sure it works reliably.

oviano:
Yep. I'm using the target Windows, and the CMAKE_COMPILER_ID is coming through as clang.

The only other change I had/wanted to make was that using clang-cl.exe brings in a warning in the code below, so you will see I expanded the #if defined condition to also include BOTAN_BUILD_COMPILER_IS_CLANG.

Otherwise it moans about deprecated/insecure getenv() function.

```
bool OS::read_env_variable(std::string& value_out, std::string_view name_view) {
   value_out = "";

   if(running_in_privileged_state())
      return false;

#if defined(BOTAN_TARGET_OS_HAS_WIN32) && (defined(BOTAN_BUILD_COMPILER_IS_MSVC) || defined(BOTAN_BUILD_COMPILER_IS_CLANG))
   const std::string name(name_view);
   char val[128] = {0};
   size_t req_size = 0;
   if(getenv_s(&req_size, val, sizeof(val), name.c_str()) == 0) {
      value_out = std::string(val, req_size);
      return true;
   }
#else
   const std::string name(name_view);
   if(const char* val = std::getenv(name.c_str())) {
      value_out = val;
      return true;
   }
#endif
```

Zimbrando:
I'm trying to compile Botan version 2.19.5 using Clang-CL on Windows, as part of setting up a project for [QSsh](https://github.com/sandsmark/QSsh). However, I'm running into multiple errors during the build process.

I first tried configuring the build using the following command:

[...]

PR #4246, opened 2024-07-23
Explicitly enable rdrand intrinsics for processor_rng.

solemnwarning:
This fixes the build when using an sufficiently old `-march` that RDRAND isn't guaranteed, or not having `-mrdrand` enabled globally.

### DIFF
--- a/src/lib/rng/processor_rng/info.txt
+++ b/src/lib/rng/processor_rng/info.txt
@@ -18,5 +18,7 @@ processor_rng.h
 </header:public>
 
 <isa>
+x86_32:rdrand
+x86_64:rdrand
 ppc64:power9
 </isa>


randombit:
The change is fine I guess but can you explain more the circumstances where you see a problem? We only use the intrinsics with MSVC; for GCC/Clang we instead use inline asm which isn't affected by processor flags either way. I do quite a bit of development on an old laptop that doesn't have RDRAND and have never had a problem here since the assembler is fine with emitting the code and we don't jump to it unless CPUID indicates support.

solemnwarning:
I'm building with the Clang included in with VS2022 using the clang-cl wrapper, so I'm in-between the supported toolchains.

I was going to open another issue to discuss this, but might as well do it here; I'm using [this patch](https://github.com/solemnwarning/rehex/blob/533e8472ddf8b7d583b47fe2e23f59068aa74e74/msvc/vcpkg-deploy/ports/botan/msvc-clang-isa-flags.patch) to add the `-msse2` etc flags when using clang-cl, which is obviously a massive hack... should I be doing something different? #3703 seems to suggest clang-cl isn't cleanly supported yet.

randombit:
Ah interesting. Yeah this is not well supported atm but you may be in the best position to do something about it; I haven't touched a Windows machine in years and afaik none of the other regular developers care about Windows much either.

I'd suggest basically copying your `msvc.txt` as `clangcl.txt` and making a new compiler target. In fact if that works for you, then please feel free to open a PR adding `clangcl.txt` as such. It may not be perfect but it's probably better than our current situation, and we can then try to iterate on CI/etc.

randombit (approved):

PR #4249, opened 2024-07-24
Try std::ranges::views

reneme:
As [suggested](https://github.com/randombit/botan/pull/4203#discussion_r1687720408), let's try whether our toolchain is now ready to use views.


reneme:
Nope! [Clang 14 is holding us back](https://github.com/randombit/botan/actions/runs/10072152327/job/27843561026?pr=4249) here. Essentially, it doesn't manage to digest the `std::span<>` in a `views::reverse`. According to [the C++20 compiler support table](https://en.cppreference.com/w/cpp/compiler_support/20), full ranges support requires clang 15 (and actually libc++16).

... I'll be patient and promise to be grateful for the things we can use. 😥

PR #4251, opened 2024-07-24
Add NUMS curve numsp512d1

randombit:
The NUMS curves are a set of curves generated using a simple and straighforward set of criteria, while also admitting very efficient implementation. Further details can be found in the IETF draft draft-black-numscurves-02


guidovranken:
@randombit  Will you also support nums t1 curves? I have no need for it personally, but SymCrypt by Microsoft supports them (but not nums d1 curves), and they are in fact the only library supported by Cryptofuzz which supports nums natively. Having an oracle for nums t1 curves could benefit the fuzzing of both libraries.

https://github.com/guidovranken/cryptofuzz/blob/8e2f0ee62596cb1dec7e149ccf047a2506049af9/modules/symcrypt/module.cpp#L1170-L1175

randombit:
@guidovranken Interesting, I was unaware of this. t1 is a completely different thing (Edwards curve with cofactor 4) so wouldn't happen as part of this ongoing project anyway.

PR #4253, opened 2024-07-24
Fix the lcov path rewriter

randombit:


src/scripts/rewrite_lcov.py
```diff
@@ -19,14 +19,20 @@ def main(args = None):
         return 1
 
     # SF:/home/jack/work/botan/build/include/botan/pkix_types.h
-    header_re = re.compile('^SF:(.*/build/include/botan/.*)')
+    header_re = re.compile('^SF:(.*/build/include/botan/(public|internal|external)/botan/.*)')
```
reneme, line 22: Isn't that one too many 'botan'?
On my machine the paths are: `build/include/(public|internal|external)/botan/(internal)?/.*`

reneme (commented):
Oh wow! I didn't know this existed.

src/scripts/rewrite_lcov.py
randombit: Thanks. Fixed - https://coveralls.io/jobs/147601360

reneme (approved):

PR #4254, opened 2024-07-24
Add expand_bit method to CT::Mask

FAlbertDev:
We want to use this in the Classic McEliece PR, where we've done some refactoring with valgrind. 

Note that if we would apply something like this
```cpp
return CT::Mask<T>::expand((v >> bit) & 1);
```
modern compiler could use the [bt instruction](https://en.wikipedia.org/wiki/Bit_Test) which [seems to confuse valgrind](https://valgrind.org/docs/manual/mc-tech-docs.html/mc-tech-docs.html#mc-tech-docs.x86instr). See [this comment](https://github.com/randombit/botan/pull/3883#discussion_r1689637418) of the CMCE PR.

### DIFF
--- a/src/lib/utils/ct_utils.h
+++ b/src/lib/utils/ct_utils.h
@@ -364,6 +364,14 @@ class Mask final {
       static constexpr Mask<T> expand_top_bit(T v) { return Mask<T>(Botan::expand_top_bit<T>(value_barrier<T>(v))); }
 
       /**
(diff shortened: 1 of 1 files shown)


randombit:
> But evidently not; for memory locations the index can be arbitrary, and the processor will index arbitrarily into memory as a result. 

What in the everloving fuck is going on at Intel.

PR #4258, opened 2024-07-25
Remove Homebrew workaround for broken Emscripten

randombit:
3.1.64 is out which is reported to resolve the issue


randombit:
Bleh, I think we have to wait for Actions to update

PR #4260, opened 2024-07-25
CT::Poison: Return input

FAlbertDev:
This adds more convenience to the new poison/unpoison interface by forwarding the input to the output. Especially for PQC algorithms like Classic McEliece, we have rejection sampling and other aborts, where we would need to transform something like
```cpp
if(value != 0){
   return std::nullopt;
}
```
into
```cpp
bool abort = (value != 0);
CT::unpoison(abort);
if(abort){
   return std::nullopt;
}
```
With this PR we can write:
```cpp
if(CT::unpoison(value != 0)){
   return std::nullopt;
}
```
-------------------------------

However, while writing this message, I'm asking myself if this might be a foot gun... For example, one could also write:
 ```cpp
if(CT::unpoison(value) != 0){
   return std::nullopt;
}
```
and expect that `value` is still poisoned after the if-block is skipped, which is wrong. Do we only want to allow this syntax for booleans (we return an unpoisoned copy of the input boolean)? Or do we want to drop it entirely and use `CT::Choice` instead (which unpoisons internally)? What do you think?

### DIFF
--- a/src/lib/utils/ct_utils.h
+++ b/src/lib/utils/ct_utils.h
@@ -200,6 +200,28 @@ template <typename... Ts>
    return scope;
 }
 
+/**
+ * Poisons an r-value @p v and forwards it as the return value.
+ */
+template <poisonable T>
+[[nodiscard]] decltype(auto) driveby_poison(T&& v)
+   requires(std::is_rvalue_reference_v<decltype(v)>)
+{
+   poison(v);
+   return std::forward<T>(v);
+}
+
+/**
+ * Unpoisons an r-value @p v and forwards it as the return value.
+ */
+template <unpoisonable T>
+[[nodiscard]] decltype(auto) driveby_unpoison(T&& v)
+   requires(std::is_rvalue_reference_v<decltype(v)>)
+{
+   unpoison(v);
+   return std::forward<T>(v);
+}
+
 /// @}
 
 /**
--- a/src/tests/test_ct_utils.cpp
+++ b/src/tests/test_ct_utils.cpp
@@ -246,6 +246,18 @@ std::vector<Test::Result> test_higher_level_ct_poison() {
                Botan::CT::unpoison_range(v);
                result.confirm("all unpoisoned", std::none_of(v.begin(), v.end(), is_poisoned));
             }),
+
+      CHECK("poison a poisonable objects with driveby_poison",
+            [](Test::Result& result) {
+               Poisonable p;
+               result.confirm("not poisoned", p.poisoned == false);
+               Poisonable p_poisoned =
+                  Botan::CT::driveby_poison(std::move(p));  // NOLINT(hicpp-move-const-arg,performance-move-const-arg)
+               result.confirm("poisoned", p_poisoned.poisoned == true);
+               Poisonable p_unpoisoned = Botan::CT::driveby_unpoison(
+                  std::move(p_poisoned));  // NOLINT(hicpp-move-const-arg,performance-move-const-arg)
+               result.confirm("unpoisoned", p_unpoisoned.poisoned == false);
+            }),
    };
 }
 

### CODE REVIEW

File: src/lib/utils/ct_utils.h [hunk 1]
```diff
@@ -96,30 +96,36 @@ inline bool poison_has_effect() {
...
 
 /**
  * Poison a contiguous buffer of trivial objects (e.g. integers and such)
  */
 template <ranges::spanable_range R>
-   requires std::is_trivially_copyable_v<std::ranges::range_value_t<R>>
-constexpr void poison(R&& r) {
+   requires std::is_trivially_copyable_v<std::ranges::range_value_t<R>> &&
+            (!requires(const R& x) { x._const_time_poison(); })
+constexpr R&& poison(R&& r) {
```
reneme, line 116: Do we have class that adhere to `spanable_range` and have the `const_time_poison()` helpers? Or what do we need the new `requires` for?

randombit:
`CT::Mask` has a concept `unpoisoned_value` for precisely this kind of thing, maybe it's better to generalize this idea (return an unpoisoned copy of a poisoned value) rather than overloading `unpoison` in this way, it seems very easy to footgun. `CT::Mask::unpoisoned_value` is not used much/at all currently - I think code that previously used it now uses `CT::Choice` and `CT::Option` instead.

I don't see how `poison` returning its input is useful.

src/lib/utils/ct_utils.h
```diff
@@ -96,30 +96,36 @@ inline bool poison_has_effect() {
...
    poison(s.data(), s.size());
+   return std::forward<R>(r);
 }
 
 template <ranges::spanable_range R>
-   requires std::is_trivially_copyable_v<std::ranges::range_value_t<R>>
-constexpr void unpoison(R&& r) {
+   requires std::is_trivially_copyable_v<std::ranges::range_value_t<R>> &&
+            (!requires(const R& x) { x._const_time_unpoison(); })
+constexpr R&& unpoison(R&& r) {
```
reneme, line 125: Here and in the other overloads: Please use `decltype(auto)` as return value. The `std::forward<R>()` will "do the right thing", by either returning as an rvalue or lvalue reference. But I'm not 100% sure that the explicit return type `R&&` always resolves to the correct reference type. Using `decltype(auto)` let's the compiler figure it out from the return value of `std::forward`.

```suggestion
constexpr decltype(auto) unpoison(R&& r) {
```

This might even obsolete the `std::forward()` in the return statement. I'll have to ponder over this.

src/ct_selftest/ct_selftest.cpp
```diff
@@ -209,6 +209,61 @@ void test_scoped_poison_outer(Botan::RandomNumberGenerator& rng) {
...
+      std::cout << "I may or may not be printed too." << std::endl;
+   }
+
+   struct Poisanable {
+         uint8_t a;
+
+         void _const_time_poison() { Botan::CT::poison(a); }
+
+         void _const_time_unpoison() { Botan::CT::unpoison(a); }
+   };
```
reneme, line 251: I deliberately didn't test the `_const_time_poison()` overload in `ct_selftest.cpp` (instead its tested in `test_ct_utils.cpp`). After all these helper methods allow to establish an _observable side effect_ that can be nicely tested in a non-valgrind-enabled unit test.

What you're actually testing below is the proper functioning of the _implementation_ of `const_time_poison()`; and you already did test `poison(uint8_t)` above.

src/ct_selftest/ct_selftest.cpp
```diff
@@ -209,6 +209,61 @@ void test_scoped_poison_outer(Botan::RandomNumberGenerator& rng) {
...
+      std::cout << "I may or may not be printed." << std::endl;
+   }
+
+   // Test that the value is unpoisoned
+   if(poisoned_byte == 1) {
+      std::cout << "I may or may not be printed too." << std::endl;
+   }
+
+   auto poisoned_range = rng.random_vec(8);
+   Botan::CT::poison_range(poisoned_range);
```
reneme, line 233: No need to test the range overload here. Just move this to `test_ct_utils.cpp`. Same rationale as explained for the `const_time_poison()` overloads.

reneme (changes requested):

reneme:
Good point with the footgun! That certainly needs some thought.

I do think that there's value here, though. Eg. the getter below could be much simpler (and there's no footgun there!).

```C++
      DilithiumSerializedPrivateKey raw_sk() const {
         auto scope = CT::scoped_poison(*this);
         auto result = Dilithium_Algos::encode_private_key(m_rho, m_tr, m_signing_seed, m_s1, m_s2, m_t0, m_mode);
         CT::unpoison(result);
         return result;
      }
```

vs.

```C++
      DilithiumSerializedPrivateKey raw_sk() const {
         auto scope = CT::scoped_poison(*this);
         return CT::unpoison(Dilithium_Algos::encode_private_key(m_rho, m_tr, m_signing_seed, m_s1, m_s2, m_t0, m_mode));
      }
```

As a suggestion: we could restrict the pass-through convenience to rvalue-parameters. Because then there's no way to refer to the object that was manipulated in any other way but the returned reference. Like so:

```C++
std::vector<uint8_t> thing;
auto poisoned_thing = poison(thing); // compile error -> lvalue-reference overload does return void
auto poisoned_thing = poison(make_thing()); // ok -> make_thing passed an rvalue of thing
auto poisoned_thing = poison(std::move(thing)); // also ok, because one must not use `thing` after move.
```

src/lib/utils/ct_utils.h
randombit: I still do not see a point in `poison` returning anything

src/ct_selftest/ct_selftest.cpp
FAlbertDev: Well, I wanted to test that the parameters are correctly passed through. To create this struct was simply a tool to test the pass-through of `poison(class)`.

src/lib/utils/ct_utils.h
FAlbertDev: I had a problem with the Crystal Polynomials, which fulfill the range concept AND have the const_time_poison methods (and are seemingly trivially copyable?). After implementing the pass-through, I encountered compiler errors, claiming that `poison` was ambiguous. I decided that the `const_time_poison()` method should always take precedence in such cases. I wonder why this did not occur without the pass-through, though...

src/lib/utils/ct_utils.h
FAlbertDev: I bow to the master :D I tried using `auto`, which did not handle references correctly. But indeed, `decltype(auto)` does the right thing. I admit I didn't know there was a difference between both variants.

src/ct_selftest/ct_selftest.cpp
FAlbertDev: Alright :)

FAlbertDev:
> Godbolt of a basic implementation: https://godbolt.org/z/qh49Ys6e1

Wow, I like it. 
I also like the idea of allowing r-values only. What do you think, @randombit? Are you okay with introducing this pass-through for r-values?

src/lib/utils/ct_utils.h
reneme: > I still do not see a point in poison returning anything

@randombit I feel the discussion about the general utility of this PR encountered a bit of a race condition. 😏
To avoid misunderstandings, let's keep the "yay or nay" discussion to the main thread. I'm not sure you've seen my argument about [restricting this pass-through to rvalue parameters](https://github.com/randombit/botan/pull/4260#issuecomment-2250223444).

reneme:
For the record: I do admit that this adds complexity in a sensitive utility that is fairly hard to test, and that we shouldn't do that lightly. As a compromise: instead of adding this rvalue-passthrough to all `poison/unpoison` overloads, we could add an explicit function that provides this functionality: like `poison_and_forward(some_rvalue())`, `driveby_poison()` (:wink:), ...

FAlbertDev:
I've implemented the `driveby_poison` functions as @reneme suggested. I think that's a suitable compromise that cannot be misused since it is explicit, and only r-values are allowed as input. I currently work on valgrind for HSS/LMS, and these functions are highly convenient in this context.

randombit (approved):

FAlbertDev:
Thanks for the approve! @reneme could you briefly check the template magic before I merge this PR?

reneme (approved):
LGTM. Also, the explicit helper methods reduce the implementation complexity significantly. 👍

PR #4275, opened 2024-07-30
Add SWAR based hex encoding/decoding

randombit:
Stacked on #4271

### DIFF
--- a/src/lib/codec/hex/hex.cpp
+++ b/src/lib/codec/hex/hex.cpp
@@ -12,31 +12,31 @@
 #include <botan/internal/charset.h>
 #include <botan/internal/ct_utils.h>
 #include <botan/internal/fmt.h>
+#include <botan/internal/int_utils.h>
+#include <botan/internal/loadstor.h>
 
 namespace Botan {
 
 namespace {
 
-char hex_encode_nibble(uint8_t n, bool uppercase) {
-   BOTAN_DEBUG_ASSERT(n <= 15);
+uint16_t hex_encode_2nibble(uint8_t n8, bool uppercase) {
+   // Offset for upper or lower case 'a' resp
+   const uint16_t a_mask = uppercase ? 0x0707 : 0x2727;
 
-   const auto in_09 = CT::Mask<uint8_t>::is_lt(n, 10);
-
-   const char c_09 = n + '0';
-   const char c_af = n + (uppercase ? 'A' : 'a') - 10;
-
-   return in_09.select(c_09, c_af);
+   const uint16_t n = (static_cast<uint16_t>(n8 & 0xF0) << 4) | (n8 & 0x0F);
+   // n >= 10? If so add offset
+   const uint16_t diff = swar_lt<uint16_t>(0x0909, n) & a_mask;
+   // Can't overflow between bytes, so don't need explicit SWAR addition:
+   return n + 0x3030 + diff;
 }
 
 }  // namespace
 
 void hex_encode(char output[], const uint8_t input[], size_t input_length, bool uppercase) {
    for(size_t i = 0; i != input_length; ++i) {
-      const uint8_t n0 = (input[i] >> 4) & 0xF;
-      const uint8_t n1 = (input[i]) & 0xF;
-
-      output[2 * i] = hex_encode_nibble(n0, uppercase);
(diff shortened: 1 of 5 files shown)


src/lib/codec/hex/hex.cpp
```diff
@@ -12,31 +12,38 @@
...
 void hex_encode(char output[], const uint8_t input[], size_t input_length, bool uppercase) {
    for(size_t i = 0; i != input_length; ++i) {
-      const uint8_t n0 = (input[i] >> 4) & 0xF;
-      const uint8_t n1 = (input[i]) & 0xF;
-
-      output[2 * i] = hex_encode_nibble(n0, uppercase);
-      output[2 * i + 1] = hex_encode_nibble(n1, uppercase);
+      const uint16_t h = hex_encode_2nibble(input[i], uppercase);
+      output[2 * i] = get_byte<0>(h);
+      output[2 * i + 1] = get_byte<1>(h);
```
reneme, line 46: Not sure that's at all easier to read. Mostly due to the cast. :(

```suggestion
      const uint16_t h = hex_encode_2nibble(input[i], uppercase);
      store_be(reinterpret_cast<uint8_t*>(&output[2 * i]), h);
```

src/lib/codec/hex/hex.cpp
```diff
@@ -53,27 +60,36 @@ std::string hex_encode(const uint8_t input[], size_t input_length, bool uppercas
...
+      const uint64_t a_hi = sub & hi;
+      return (lo7 - a_lo + upper) & hi & ~a_hi;
+   };
 
-   const auto is_whitespace =
-      CT::Mask<uint8_t>::is_any_of(c, {uint8_t(' '), uint8_t('\t'), uint8_t('\n'), uint8_t('\r')});
+   // Starting value: '0', 'a', 'A', ' ', '\n', '\t', '\r'
+   constexpr uint64_t v_lo = 0x00306141200a090d;
+   // Valid ranges: 10, 6, 6, 1, 1, 1, 1
+   constexpr uint64_t v_range = 0x0A060601010101;
```
reneme, line 82: Similar to the base64 version: I find this resolves the "magic numbers" quite nicely.

```suggestion
   // Starts of valid value ranges (v_lo) and their lengths (v_range)
   constexpr uint64_t v_lo = make_uint64(0, '0', 'a', 'A', ' ', '\n', '\t', '\r');
   constexpr uint64_t v_range = make_uint64(0, 10, 6, 6, 1, 1, 1, 1);
```

src/lib/codec/hex/hex.cpp
```diff
@@ -53,27 +60,36 @@ std::string hex_encode(const uint8_t input[], size_t input_length, bool uppercas
...
+      constexpr uint64_t lo7 = 0x7F7F7F7F7F7F7F7F;
 
-   const auto is_alpha_upper = CT::Mask<uint8_t>::is_within_range(c, uint8_t('A'), uint8_t('F'));
-   const auto is_alpha_lower = CT::Mask<uint8_t>::is_within_range(c, uint8_t('a'), uint8_t('f'));
-   const auto is_decimal = CT::Mask<uint8_t>::is_within_range(c, uint8_t('0'), uint8_t('9'));
+      const uint64_t sub = ((v | hi) - (lower & lo7)) ^ ((v ^ (~lower)) & hi);
+      const uint64_t a_lo = sub & lo7;
+      const uint64_t a_hi = sub & hi;
+      return (lo7 - a_lo + upper) & hi & ~a_hi;
+   };
```
reneme, line 77: This (and perhaps other helpers) might be a good candidate for reusable SWAR utilities. On a first glance this looks very similar to the implementation in base64. Or am I missing the fine details?

reneme (approved):
Sweet! Just a few suggestions to reuse some utilities we have anyway. None of that is critical at all.

src/lib/codec/hex/hex.cpp
randombit: I hesitated on these since the way these are formulated is not totally general (in that they cannot handle inputs where the high bit is set). This can be done but adds several more operations which I’d like to avoid given that the whole point is to make encoding/decoding faster than the current CT::Mask code. But they are otherwise basically identical (at most differing based on the size of the type in use) so I moved these to `int_utils.h` with doc comments indicating the current limitations.

PR #4277, opened 2024-07-31
FIX: allow multiple BOTAN_REGISTER_TEST_FN()

reneme:
This allows to register multiple test functions under different names in a single test_*.cpp file. Before that would have cause compile errors due to a redeclaration of the static registration variable. This now generates the variable name by incorporating the first registered function's name.

Note that this also enables [`/Zc:preprocessor` in MSVC](https://learn.microsoft.com/en-us/cpp/build/reference/zc-preprocessor) to make its preprocessor behave in conformance with the C++ standard. Progress!

### DIFF
--- a/src/tests/tests.h
+++ b/src/tests/tests.h
@@ -757,14 +757,18 @@ class TestFnRegistration {
       }
 };
 
+#define BOTAN_REGISTER_TEST_FN_IMPL(category, name, smoke_test, needs_serialization, fn0, ...) \
+   static const TestFnRegistration register_##fn0(                                             \
+      category, name, smoke_test, needs_serialization, {__FILE__, __LINE__}, fn0 __VA_OPT__(, ) __VA_ARGS__)
+
 #define BOTAN_REGISTER_TEST_FN(category, name, ...) \
-   static const TestFnRegistration reg_##fn_name(category, name, false, false, {__FILE__, __LINE__}, __VA_ARGS__)
+   BOTAN_REGISTER_TEST_FN_IMPL(category, name, false, false, __VA_ARGS__)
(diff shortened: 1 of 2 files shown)


src/tests/tests.h
```diff
@@ -757,14 +757,18 @@ class TestFnRegistration {
       }
 };
 
+#define BOTAN_REGISTER_TEST_FN_IMPL(category, name, smoke_test, needs_serialization, fn0, ...) \
+   static const TestFnRegistration register_##fn0(                                             \
+      category, name, smoke_test, needs_serialization, {__FILE__, __LINE__}, fn0 __VA_OPT__(, ) __VA_ARGS__)
```
reneme, line 762: Disambiguates the `static const TestFnRegistration` variable names by incorporating the first registered function name (`fn0`) into the variable identifier.

As a result, one can now do things like:

```C++

Test::Result test_sha1() {/* ...*/}

Test::Result test_sha256() {/* ...*/}
Test::Result test_sha384() {/* ...*/}
Test::Result test_sha512() {/* ...*/}

BOTAN_REGISTER_TEST_FN("hash", "sha1", test_sha1);
BOTAN_REGISTER_TEST_FN("hash", "sha2", test_sha256, test_sha384, test_sha512);
```

randombit: Its a shame Clang’s `__COUNTER__` isnt standard

randombit (approved):

src/tests/tests.h
reneme: Today I learnt! Googling that, and someone suggests to use `__LINE__` instead. And in fact that would actually work just fine for this use case. Good to keep in mind, perhaps.

PR #4279, opened 2024-07-31
Remove internal annotation from pcurves modules

randombit:
This prevents a user from using the entirely sensible configuration of

--minimized --enable-modules=pcurves_secp256r1,sha2_32

### DIFF
--- a/src/lib/math/pcurves/pcurves_brainpool256r1/info.txt
+++ b/src/lib/math/pcurves/pcurves_brainpool256r1/info.txt
@@ -5,7 +5,6 @@ PCURVES_BRAINPOOL256R1 -> 20240608
 <module_info>
 name -> "PCurve brainpool256r1"
 brief -> "brainpool256r1"
-type -> "Internal"
 </module_info>
 
 <requires>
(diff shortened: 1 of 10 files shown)


reneme:
We did [have this discussion](https://github.com/randombit/botan/pull/3979#discussion_r1632698035) before, but `--enable-modules` and `--disable-modules` currently aren't consistent. The latter allows to manipulate internal modules, the former does not. Perhaps, instead we could just resolve this inconsistency.

Nevertheless: I think we should pick up the discussion about the definition of "internal module" and (if needed) update it in `configure.rst`:

https://github.com/randombit/botan/blob/03d8b16bbe4c007e879f2061938d2466247e18ed/doc/dev_ref/configure.rst?plain=1#L211-L212

I was of the impression that the `pcurves` modules don't expose any public API, hence they should be internal. Or am I missing something?

randombit:
> I was of the impression that the pcurves modules don't expose any public API, hence they should be internal. Or am I missing something?

You are correct that there is no public API. And yet their availability or lack thereof is visible to the application developer via the side channels of performance and code size.

reneme:
> And yet their availability or lack thereof is visible to the application developer via the side channels of performance and code size.

😏 Sure, but that's likely true for *any* internal module.

reneme:
Sounds reasonable to me. Also the 'disableability' argument. Thanks.

reneme (approved):

PR #4281, opened 2024-07-31
Doc: clarify definition of 'Internal' modules

reneme:
As discussed in #4279 and [here](https://github.com/randombit/botan/pull/3979#discussion_r1632698035).


randombit (approved):
:+1:

PR #4151, opened 2024-06-26
Chore: BufferTransformer and Blowfish Refactor

reneme:
This picks up the work started in #3942 by proposing a `BufferTransformer` class; which is essentially an orchestrator for `BufferSlicer` and `BufferStuffer`. [As mentioned yesterday](https://github.com/randombit/botan/pull/4042#discussion_r1652432882), I added some template fanciness, to handle the "block processing" necessary for various block ciphers. [As suggested previously](https://github.com/randombit/botan/pull/3870#issuecomment-2039228807), this uses Blowfish to demonstrate the functionality (+ spanifying the blowfish implementation).

@randombit Is the `process_blocks_of<>` implementation in line with [what you had in mind previously](https://github.com/randombit/botan/pull/3942#issuecomment-2039203630):

>  I wonder if we can fit in some other common functionality there, namely the word level loads/stores. That's a very common task (encrypt+decrypt for N ciphers plus alternative implementations)

Here's the idea of `BufferTransformer::process_blocks_of`, where the multi-block loop handling and buffer management is entirely abstracted. Also, as an additional goodie, we now have _explicit compile-time knowledge_ of the blocked data length inside the processing lambdas. Whether the compiler uses this or not has to be seen. FWIW: I don't see a difference in the performance of Blowfish on my Mac.

```C++
void encrypt_n(const uint8_t inb[], uint8_t outb[], size_t blocks) {
   std::span<const uint8_t> in(inb, inb + blocks*BS);
   std::span<uint8_t> out(outb, outb + blocks*BS);
   
   BufferTransformer bt(in, out);
   bt.process_blocks_of<BS*4, BS>(
      overloaded{
         [](std::span<const uint8_t, BS*4> in, std::span<uint8_t, BS*4> out) {
            // encrypt four blocks at a time
         },
         [](std::span<const uint8_t, BS> in, std::span<uint8_t, BS> out) {
            // encrypt a single block
         }
      }
   );
}
```

No hard feelings if that doesn't make it into 3.5.0 at all. I just nerd-sniped myself yesterday and wanted to sketch this out. And frankly, I think the `BufferTransformer` implementation is quite concise. 😄

PR #4288, opened 2024-08-05
Load/Store for SIMD type wrappers

reneme:
This is an attempt for adding support to `load_le` and/or `store_be` for custom types. Essentially, any custom type can implement adapter methods `static T::load_{be/le} -> T` and `T::store_{be/le} -> void` to hook into this. Essentially the same concept as we established with `_const_time_poison()`.

By implementing this for `SIMD_4x32`, combining this with [the proposed `BufferTransformer`](https://github.com/randombit/botan/pull/4151) and a little bit of glue, I ended up with this for AES-NI-128 decrypt function (that provides the same functionality as [the original implementation](https://github.com/randombit/botan/blob/13edb926d2b9f493d062677406738966eed6748b/src/lib/block/aes/aes_ni/aes_ni.cpp#L188-L246)):

```C++
BOTAN_FUNC_ISA("ssse3,aes") void AES_128::hw_aes_decrypt_n(const uint8_t in[], uint8_t out[], size_t blocks) const {
   constexpr size_t rounds = 10;
   const auto K = load_le<std::array<SIMD_4x32, rounds + 1>>(Botan::as_bytes(std::span{m_DK}));

   constexpr size_t four = 4 * BLOCK_SIZE;
   constexpr size_t one = 1 * BLOCK_SIZE;

   BufferTransformer(std::span{in, blocks * BLOCK_SIZE}, std::span{out, blocks * BLOCK_SIZE})
      .process_blocks_of<four, one>([&](auto i, auto o) {
         constexpr size_t blocks = i.size() / BLOCK_SIZE;
         auto Bs = load_le<std::array<SIMD_4x32, blocks>>(i);

         keyxor_new(K[0], Bs);
         for(size_t round = 1; round != rounds; ++round) {
            aesdec_new(K[round], Bs);
         }
         aesdeclast_new(K[10], Bs);

         store_le(o, Bs);
      });
}
```

When pulling the `rounds` variable into a template parameter we might even be able to share this implementation for the other AES keylengths.

### Mini-Benchmark

The generated assembly is exactly the same (for clang 18). GCC 13 is doing an equally good job. No perceivable slowdown on either compiler. MSVC doesn't seem to care to unroll the statically sized for-loops and ~~perhaps other things~~ and in particular seems to have trouble seeing through the load/store magic, resulting in an "amazing" 8x slowdown. 😒 This is really sad!

### DIFF
--- a/src/lib/utils/loadstor.h
+++ b/src/lib/utils/loadstor.h
@@ -168,11 +168,33 @@ constexpr bool native_endianness_is_unknown() {
 }
 
 /**
+ * Models a custom type that provides factory methods to be loaded in big- or
+ * little-endian byte order.
+ */
+template <typename T>
+concept custom_loadable = requires(std::span<const uint8_t, sizeof(T)> data) {
+   { T::load_be(data) } -> std::same_as<T>;
+   { T::load_le(data) } -> std::same_as<T>;
+};
+
+/**
+ * Models a custom type that provides store methods to be stored in big- or
+ * little-endian byte order.
+ */
+template <typename T>
+concept custom_storable = requires(std::span<uint8_t, sizeof(T)> data, const T value) {
+   { value.store_be(data) };
+   { value.store_le(data) };
+};
+
+/**
  * Models a type that can be loaded/stored from/to a byte range.
  */
 template <typename T>
-concept unsigned_integralish = std::unsigned_integral<strong_type_wrapped_type<T>> ||
-                               (std::is_enum_v<T> && std::unsigned_integral<std::underlying_type_t<T>>);
+concept unsigned_integralish =
+   std::unsigned_integral<strong_type_wrapped_type<T>> ||
+   (std::is_enum_v<T> && std::unsigned_integral<std::underlying_type_t<T>>) ||
+   (custom_loadable<strong_type_wrapped_type<T>> || custom_storable<strong_type_wrapped_type<T>>);
 
 template <typename T>
 struct wrapped_type_helper_with_enum {
@@ -276,6 +298,7 @@ inline constexpr void fallback_store_any(InT in, OutR&& out_range) {
  * @return T loaded from @p in_range, as a big-endian value
  */
 template <Endianness endianness, unsigned_integralish WrappedOutT, ranges::contiguous_range<uint8_t> InR>
+   requires(!custom_loadable<strong_type_wrapped_type<WrappedOutT>>)
 inline constexpr WrappedOutT load_any(InR&& in_range) {
    using OutT = detail::wrapped_type<WrappedOutT>;
    ranges::assert_exact_byte_length<sizeof(OutT)>(in_range);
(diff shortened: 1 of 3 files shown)


randombit:
> GCC 13 is doing an equally good job.

In my experience GCC only unrolls loops with a `constexpr` or equivalent loop count with `-O3` - with `-O2` or lower it fails to unroll. That's fine, cause we use `-O3` ... except that most Linux distros compile us with their "approved" set of compiler flags which uses `-O2` :disappointed:

reneme:
> with -O3 - with -O2 or lower it fails to unroll.

We could force the unrolling by using an index sequence and fold expressions. We're using this trick already for [the bswap fallback implementation](https://github.com/randombit/botan/blob/13edb926d2b9f493d062677406738966eed6748b/src/lib/utils/loadstor.h#L215-L230) and IIRC compilers even recognized this as an actual bswap.

Wrapping that into an "okay-to-use" template, doesn't even lead to unreadable code, and unrolls regardless of the optimization settings ([godbolt.org](https://godbolt.org/z/qfnoxc5Ev)).

```C++
template <size_t begin, size_t end, std::invocable<size_t> FnT>
    requires(end >= begin)
constexpr void unrolled_for(FnT&& fn) {
    [&]<size_t... indices>(std::index_sequence<indices...>) {
        (fn(indices + begin), ...);
    }(std::make_index_sequence<end - begin>());
}

int main() {
  unrolled_for<2, 10>([](size_t i) {std::cout << i << '\n'; });
}
```

Frankly, I'm much more concerned about MSVC here. And mostly because the mentioned slowdown seemed to come from the poor optimization of the load/store abstractions. I toyed with its flags a bit but no luck. 😞

src/tests/test_utils.cpp
```diff
@@ -466,6 +470,42 @@ class Utility_Function_Tests final : public Test {
          result.test_is_eq(Botan::store_le<std::vector<uint8_t>>(TestEnum32::_1), Botan::hex_decode("78563412"));
          result.test_is_eq<std::array<uint8_t, 4>>(Botan::store_be(TestEnum32::_2), {0x78, 0x56, 0x34, 0x12});
 
+#if defined(BOTAN_HAS_SIMD_32)
+
```
randombit, line 474: You also have to guard this at runtime with `CPUID::has_simd_32` since it's quite possible (if somewhat less likely these days) to compile with SIMD enabled but it's not available at runtime on whatever CPU we land on.

randombit (approved):

randombit:
CI failures are relevant. Easiest fix is probably to move the test to `test_simd.cpp`

reneme:
I tried adding the same load/store helpers to `SIMD_8x32` and `SIMD_16x32`. Though, this isn't as straight-forward due to the inline ISA annotations. Namely, the `load_any` wrappers aren't marked with those ISA flags and therefore cannot inline the implementations. I'll leave this as future work here.

PR #4291, opened 2024-08-06
SLH-DSA integration into SPHINCS+

FAlbertDev:
# PQC: SLH-DSA (Initial Public Draft)
Similar to PRs #3893 and #4270, this PR integrates the [SLH-DSA IPD](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.205.ipd.pdf) instances into our SPHINCS<sup>+</sup> implementation. The difference to the current SPHINCS<sup>+</sup> round 3.1 implementation is marginal. I also named all instances without the IPD suffix to make the transition to the final standard as smooth as possible.

## Module Hierarchy
I added new modules to allow users to activate only the final SLH-DSA instances. Since the logic is almost the same, no new logic is added to these modules. However, I think it's quite handy for users to allow only SLH-DSA instances via modules (for example, using a policy).

## State of this Pull Request
This PR is in a draft state since **it won't be merged until the final standard is published**. Applications that want to experiment with the IPD instances may work with this PR for now. However, the IPD instances (and temporary OIDs) are dropped before the final merge.

## SLH-DSA Specification Release - TODOs
- [ ] Exchange IPD with SLH-DSA OIDs
- [ ] Check (and possibly adopt) SLH-DSA test vectors


src/lib/pubkey/sphincsplus/sphincsplus_common/sp_fors.cpp
```diff
@@ -34,7 +31,11 @@ std::vector<TreeNodeIndex> fors_message_to_indices(std::span<const uint8_t> mess
...
    for(auto& idx : indices) {
       for(uint32_t i = 0; i < params.a(); ++i, ++offset) {
-         idx ^= (((message[offset >> 3] >> (offset & 0x7)) & 0x1) << i);
+         if(params.is_slh_dsa()) {
+            idx ^= (((message[offset >> 3] >> (~offset & 0x7)) & 0x1) << (params.a() - 1 - i));
+         } else {
+            idx ^= (((message[offset >> 3] >> (offset & 0x7)) & 0x1) << i);
+         }
       }
    }
```
FAlbertDev, line 40: This is the only logic that differs between SPHINCS+ Round 3.1 and SLH-DSA IPD .

randombit:
Rumors that NIST will produce final specs next week

randombit:
To me seems fine to implement the prehashed variant later, or not at all.

FAlbertDev:
Update: I have now implemented SLH-DSA (without context and without prehash). However, I already prepared contexts and pre-hashes for the next iteration. Also, I did not rename the classes and files to not bloat this PR. A follow-up PR will handle this, which will not contain any logical changes. 
Test vectors are still missing, but they are coming soon. Also, I still want to double-check the current implementation with the SLH-DSA spec, so we do not miss anything new.

FAlbertDev:
Update: Test vectors and double-checks are integrated. Next week, I will look into SLH-DSA with X.509. Otherwise, this PR should be ready for review (and side-channel analysis (@aewag)) :)

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.h
```diff
@@ -68,6 +66,26 @@ class BOTAN_PUBLIC_API(3, 1) SphincsPlus_PublicKey : public virtual Public_Key {
...
 BOTAN_DIAGNOSTIC_IGNORE_INHERITED_VIA_DOMINANCE
 
+/**
+ * @brief An SLH-DSA private key.
+ *
+ * This class represents an SLH-DSA private key (or a SPHINCS+ Round 3.1 private key).
+ * Supported are all parameter sets defined in FIPS 205, Section 11. Parameter
+ * sets are specified using the Sphincs_Parameter_Set enum and the with a
+ * Sphincs_Hash_Type enum, for example SLH-DSA-SHA2-128s is defined as
+ * Sphincs_Parameter_Set::SLHDSA128Small and Sphincs_Hash_Type::Sha256.
```
reneme, line 76: Huh?

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.h
```diff
@@ -68,6 +66,26 @@ class BOTAN_PUBLIC_API(3, 1) SphincsPlus_PublicKey : public virtual Public_Key {
...
+ *
+ * For legacy usage of SPHINCS+ Round 3 (not recommended), the parameter sets
+ * Sphincs128Small, ..., Sphincs256Fast are used.
+ *
+ * Note that the parameter sets denoted as 'small' are much slower in signature
+ * creation than the 'fast' sets but with smaller signature sizes.
```
reneme, line 82:
```suggestion
 * Note that the parameter sets denoted as 'small' optimize for signature size
 * at the expense of signing speed, whereas 'fast' trades larger signatures for
 * faster signing speeds.
```

reneme (commented):
This looks really great already. I appreciate all the references into the FIPS 205 document for the algorithms.

Two more pressing things that we should address in my opinion:

1. Hide the specifics of SLH-DSA and SPHINCS+ behind customization points to be able to properly disable that code when selectively enabling/disabling modules. Especially for `prepare_message` that should be doable. Don't overstretch this, though: the added complexity should stay reasonable.
2. The handling of prehash should not be intertwined with the keys. Details and rationale in the respective comment.

Regarding (1.): We might have to try it, but I feel it would help to re-organize the modules a little to facilitate composability. Similar to what I've done in ML-DSA with the `dilithium_shake` internal module. Like so:

```
* sphincsplus_common
  * sphincsplus_sha2_base
  * sphincsplus_shake_base
* sphincsplus_base (might not even be needed)
  * sphincsplus_sha2
  * sphincsplus_shake
* slh_dsa_base (may or may not be needed)
  * slh_dsa_sha2
  * slh_dsa_shake
```

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_parameters.cpp
```diff
@@ -18,84 +18,304 @@
...
+
+   if(hash == Sphincs_Hash_Type::Shake256 && input_mode == SlhDsaInputMode::PreHash) {
+      switch(set) {
+         case Sphincs_Parameter_Set::Sphincs128Small:
+         case Sphincs_Parameter_Set::Sphincs128Fast:
+         case Sphincs_Parameter_Set::Sphincs192Small:
+         case Sphincs_Parameter_Set::Sphincs192Fast:
+         case Sphincs_Parameter_Set::Sphincs256Small:
+         case Sphincs_Parameter_Set::Sphincs256Fast:
+            throw Invalid_Argument("Cannot serialize invalid parameter combination");
```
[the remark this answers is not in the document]
FAlbertDev: I'll throw the prehash out of the params.

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_parameters.cpp
```diff
@@ -152,29 +386,68 @@ Sphincs_Parameters::Sphincs_Parameters(Sphincs_Parameter_Set set,
...
+      case Sphincs_Parameter_Set::SLHDSA256Fast:
+         return Sphincs_Parameters(set, hash, input_mode, 32, 68, 17, 9, 35, 16, 255);
    }
    BOTAN_ASSERT_UNREACHABLE();
 }
 
 Sphincs_Parameters Sphincs_Parameters::create(std::string_view name) {
-   return Sphincs_Parameters::create(set_from_name(name), hash_from_name(name));
+   auto [param_set, hash_type, input_mode] = set_and_hash_and_input_mode_from_name(name);
+   BOTAN_ARG_CHECK(input_mode == SlhDsaInputMode::Pure, "Prehash mode instances not yet supported");
```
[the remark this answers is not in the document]
FAlbertDev: I'll throw the prehash instances out.

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.cpp
```diff
@@ -150,26 +198,32 @@ class SphincsPlus_Verification_Operation final : public PK_Ops::Verification {
...
-      /*
-      * Perform a verification operation
-      */
+      /**
+       * Perform a verification operation
+       */
       bool is_valid_signature(std::span<const uint8_t> sig) override {
+         const auto msg = prepare_message(std::exchange(m_msg_buffer, {}), m_public->parameters(), m_context);
+         return slh_verify_internal(msg, sig);
+      }
```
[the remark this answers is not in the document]
FAlbertDev: Good point. I'll remove the input mode info from the parameters and remove the prehash OIDs for now. 
Regarding OIDs: Interesting... I always thought that the `AlgorithmIdentifier` of `create_x509_verification_op`  is the same as the one of the key (which would be pretty senseless in retrospect). Good to know!

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.cpp
```diff
@@ -19,14 +20,53 @@
...
+      }();
+      prefix_stuffer.append(input_mode_byte);
+      prefix_stuffer.append(checked_cast_to<uint8_t>(context.size()));
+      prefix_stuffer.append(context);
+      if(params.input_mode() == SlhDsaInputMode::PreHash) {
+         throw Not_Implemented("PreHash mode not implemented yet");
+      }
+      BOTAN_ASSERT_NOMSG(prefix_stuffer.full());
+
+      msg.insert(msg.begin(), prefix.begin(), prefix.end());
```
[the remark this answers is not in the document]
FAlbertDev: Good idea, I didn't think about that possibility!

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.cpp
```diff
@@ -333,28 +416,20 @@ class SphincsPlus_Signature_Operation final : public PK_Ops::Signature {
...
+      std::vector<uint8_t> m_context;
 };
 
 std::unique_ptr<PK_Ops::Signature> SphincsPlus_PrivateKey::create_signature_op(RandomNumberGenerator& rng,
                                                                                std::string_view params,
                                                                                std::string_view provider) const {
    BOTAN_UNUSED(rng);
    BOTAN_ARG_CHECK(params.empty() || params == "Deterministic" || params == "Randomized",
-                   "Unexpected parameters for signing with SPHINCS+");
+                   "Unexpected parameters for signing with SLH-DSA");
```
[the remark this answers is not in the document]
FAlbertDev: I think only very few people will use the round 3.1. instances in the future. I can't hurt to replace the messages with `SLH-DSA/SPHINCS+`, though.

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.h
FAlbertDev: >Huh?

I think at least the private key should have at least some documentation at it's class. It had none so far. Or what do you mean?

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.cpp
```diff
@@ -19,14 +20,53 @@
...
+      prefix_stuffer.append(checked_cast_to<uint8_t>(context.size()));
+      prefix_stuffer.append(context);
+      if(params.input_mode() == SlhDsaInputMode::PreHash) {
+         throw Not_Implemented("PreHash mode not implemented yet");
+      }
+      BOTAN_ASSERT_NOMSG(prefix_stuffer.full());
+
+      msg.insert(msg.begin(), prefix.begin(), prefix.end());
+   }
+   // else: SPHINCS+ Round 3.1 uses the message without any prefix
```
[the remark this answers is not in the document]
FAlbertDev: The problem is that the logic for prepare_message differs in the SLH-DSA or SPHINCS+ dimension, while the hash function is in the SHA2 and SHAKE dimension. These two dimensions are orthogonal. Since the logic for the hashes is shared between SLH-DSA and SPHINCS+, I currently only need two derived classes (one for SHA2 and one for SHAKE). Adding an additional layer of module and class hierarchy for the prepare_message function (15 lines of code) alone is overkill, in my opinion.

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_fors.cpp
```diff
@@ -34,7 +34,12 @@ std::vector<TreeNodeIndex> fors_message_to_indices(std::span<const uint8_t> mess
 
    for(auto& idx : indices) {
       for(uint32_t i = 0; i < params.a(); ++i, ++offset) {
-         idx ^= (((message[offset >> 3] >> (offset & 0x7)) & 0x1) << i);
+         // This is one of the few places where SPHINCS+ round 3.1 and SLH-DSA differ
+         if(params.is_slh_dsa()) {
+            idx ^= (((message[offset >> 3] >> (~offset & 0x7)) & 0x1) << (params.a() - 1 - i));
+         } else {
+            idx ^= (((message[offset >> 3] >> (offset & 0x7)) & 0x1) << i);
+         }
```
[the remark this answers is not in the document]
FAlbertDev: Yeah, I think this is overkill. (Same argument as below for `prepare_message`)
Also, sadly, it's `(offset & 0x7)`  (instead of `(~offset & 0x7)`) in the second case.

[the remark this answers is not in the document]
reneme: Yeah, I agree.

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.cpp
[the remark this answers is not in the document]
reneme: Alternatively, you could `#ifdef` the implementations depending on whether `BOTAN_HAS_SLH_DSA` and/or `BOTAN_HAS_SPHINCSPLUS` and add a default error case if necessary.

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.h
FAlbertDev: Oh, I see ;)

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.cpp
```diff
@@ -378,14 +368,9 @@ class SphincsPlus_Signature_Operation final : public PK_Ops::Signature {
...
-            if(addrnd) {
-               // Randomized Variant
-               return SphincsOptionalRandomness(addrnd->begin(), addrnd->end());
-            }
-            // Deterministic Variant
-            return SphincsOptionalRandomness(m_public->seed().begin(), m_public->seed().end());
-         }();
+         // Use addrng for the randomized variant. Use the public seed for the deterministic one.
+         auto opt_rand =
+            (addrnd.has_value()) ? addrnd.value() : StrongSpan<const SphincsOptionalRandomness>(m_public->seed());
```
reneme, line 372: Perhaps even:

```C++
const auto opt_rand = addrnd.value_or(StrongSpan<const SphincsOptionalRandomness>(m_public->seed()));
```

... not compiled, grain of 🧂 please.

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.cpp
```diff
@@ -251,7 +240,7 @@ std::unique_ptr<PK_Ops::Verification> SphincsPlus_PublicKey::create_x509_verific
    const AlgorithmIdentifier& signature_algorithm, std::string_view provider) const {
    if(provider.empty() || provider == "base") {
       if(signature_algorithm != this->algorithm_identifier()) {
-         throw Decoding_Error("Unexpected AlgorithmIdentifier for SLH-DSA signature");
+         throw Decoding_Error("Unexpected AlgorithmIdentifier for SLH-DSA(or SPHINCS+) signature");
```
reneme, line 242: Missing a space. Here and in some other places.

FAlbertDev:
Thanks for your review, @reneme. I applied your review suggestions and sprinkled in some StrongTypes. Regarding the logic separation between SLH-DSA and SPHINCS+:
The total difference between both versions is about 15 lines of code. Since we cannot simply reflect this logic in the `sp_hash` hierarchy, I think it's not worth building an entirely new hierarchy(including new modules and classes).

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.cpp
```diff
@@ -19,14 +20,40 @@
...
+   if(params.is_slh_dsa()) {
+      // prefix (no pre-hash mode): input mode byte + |ctx| + ctx
+      msg_internal.prefix = SphincsMessagePrefix(1 + 1 + context.size());
+      BufferStuffer prefix_stuffer(msg_internal.prefix);
+      const uint8_t input_mode_byte = 0x00;  // Pure (TODO: pre-hash mode: 0x01)
+      prefix_stuffer.append(input_mode_byte);
+      prefix_stuffer.append(checked_cast_to<uint8_t>(context.size()));
+      prefix_stuffer.append(context);
+
+      BOTAN_ASSERT_NOMSG(prefix_stuffer.full());
```
reneme, line 50: Alternatively (to avoid fiddling with the stuffer and precomputing the required buffer length):

```C++
msg_internal.prefix = concat<SphincsMessagePrefix>(store_be(input_mode_byte),
                                                   store_be(checked_cast_to<uint8_t>(context.size())),
                                                   context);
```

... unfortunately, `concat()` can't deal directly with the atomic underlying type (`uint8_t`), hence the `store_be` invocations (that transform the single byte into a 1-byte array which is a "range"). One day, I'll look into allowing this. :)

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.cpp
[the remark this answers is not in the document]
FAlbertDev: I'll try that 👍

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.cpp
```diff
@@ -19,14 +20,40 @@
...
    false,
-   "botan module 'sphincsplus_common' is useful only when enabling at least 'sphincsplus_sha2' or 'sphincsplus_shake'");
+   "botan module 'sphincsplus_common' is useful only when enabling at least 'sphincsplus_sha2', 'sphincsplus_shake', 'slh_dsa_sha2', or 'slh_dsa_shake'");
 #endif
 
 namespace Botan {
 
+namespace {
+// FIPS 205, Algorithm 22, line 8
+SphincsMessageInternal prepare_message(SphincsInputMessage&& msg,
```
reneme, line 36: Do you have a good reason to enforce that `SphincsInputMessage` is an rvalue. (I see that it is coming from a `std::exchange` but that doesn't demand the rvalue-ref here, or does it?)

```suggestion
SphincsMessageInternal prepare_message(SphincsInputMessage msg,
```

reneme (commented):
I think the bikeshed should be orange!

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.cpp
FAlbertDev: Since the input message has been moved into the SphincsMessageInternal struct, I need an rvalue or a copy. I want to prevent us from accidentally copying the full message, so I used an rvalue-ref. Do you think this is problematic?

reneme: There's nothing (technically) wrong with it. Especially not for such an internal interface.

This is really more of a personal style preference.

FAlbertDev: I'll change it then, since both versions are fine to me :)

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_fors.cpp
```diff
@@ -32,14 +32,26 @@ std::vector<TreeNodeIndex> fors_message_to_indices(std::span<const uint8_t> mess
 
    uint32_t offset = 0;
 
+   // This is one of the few places where the logic of SPHINCS+ round 3.1 and SLH-DSA differs
+   auto update_idx = [&]() -> std::function<void(TreeNodeIndex&, uint32_t)> {
+#if defined(BOTAN_HAS_SLH_DSA_WITH_SHA2) or defined(BOTAN_HAS_SLH_DSA_WITH_SHAKE)
```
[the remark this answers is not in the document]
FAlbertDev: As far as I know both versions work. But for consistency I'll go with `||`.

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.cpp
```diff
@@ -19,14 +20,45 @@
...
+                                       const Sphincs_Parameters& params,
+                                       StrongSpan<const SphincsContext> context) {
+   BOTAN_ARG_CHECK(params.is_slh_dsa() || context.empty(), "Context is not supported for SPHINCS+");
+#if defined(BOTAN_HAS_SLH_DSA_WITH_SHA2) || defined(BOTAN_HAS_SLH_DSA_WITH_SHAKE)
+   if(params.is_slh_dsa()) {
+      // prefix (no pre-hash mode): input mode byte + |ctx| + ctx
+      const uint8_t input_mode_byte = 0x00;  // Pure (TODO: pre-hash mode: 0x01)
+      return {.prefix = concat<SphincsMessagePrefix>(
+                 store_be(input_mode_byte), store_be(checked_cast_to<uint8_t>(context.size())), context),
+              .message = std::move(msg)};
```
reneme, line 46: If you put the extra comma behind the `std::move` it'll come up with a nicer formatting:

```suggestion
              .message = std::move(msg),};
```

(not in the suggestion above)

reneme (approved):

reneme (approved):

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_parameters.h
```diff
@@ -18,7 +18,7 @@ namespace Botan {
 enum class Sphincs_Hash_Type {
    Shake256,
    Sha256,
-   Haraka,  ///< Haraka is currently not supported
+   BOTAN_DEPRECATED("Haraka is not and will not be supported") Haraka,  ///< Haraka is currently not supported
```
TJ-91, line 21: When using SLH-DSA, I get a syntax error since `[[deprecated]]` needs to be after the enum element

```suggestion
   Haraka BOTAN_DEPRECATED("Haraka is not and will not be supported"),  ///< Haraka is currently not supported
```

FAlbertDev: True! Thank you!

src/lib/ffi/ffi_pkey_algs.cpp
```diff
@@ -923,6 +927,57 @@ int botan_pubkey_x448_get_pubkey(botan_pubkey_t key, uint8_t output[56]) {
...
+   });
+#else
+   BOTAN_UNUSED(key, key_len, privkey, slhdsa_mode);
+   return BOTAN_FFI_ERROR_NOT_IMPLEMENTED;
+#endif
+}
+
+int botan_pubkey_load_slh_dsa(botan_pubkey_t* key, const uint8_t pubkey[], size_t key_len, const char* slhdsa_mode) {
+#if defined(BOTAN_HAS_SLH_DSA_WITH_SHA2) || defined(BOTAN_HAS_SLH_DSA_WITH_SHAKE)
+   *key = nullptr;
```
randombit, line 959: Missing nullptr check

randombit (approved):

randombit:
IMO fine to ship this in 3.6.0 but we can wait for 3.7.0 if you want context support, prehashing etc which is blocked on #4318

Issue #4051, opened 2024-05-11
CI needs to test deprecated and experimental builds


randombit:
Right now we don't afaik check that these configurations build/work.

Don't need to cover the entire matrix can just do a GCC+Linux build.

PR #4293, opened 2024-08-07
Fix: Deprecated Module in BSI Policy

FAlbertDev:
If I build with 
`./configure.py ... --module-policy=bsi --disable-deprecated-features` 
I get the following error:
`ERROR: Module policy requires module dlies not usable on this platform`

I think it's better to enable only modules that are not deprecated in the respective policies. Users who want to use them anyway can still add them via `--enable-modules`. Therefore, I removed the deprecated `dlies` module from the BSI policy. 

For regression testing, I also added the `--disable-deprecated-features` flag to the BSI/NIST policy CI builds.

### DIFF
--- a/src/lib/pubkey/ec_group/ec_point.h
+++ b/src/lib/pubkey/ec_group/ec_point.h
@@ -223,6 +223,7 @@ class BOTAN_PUBLIC_API(2, 0) EC_Point final {
 #if defined(BOTAN_DISABLE_DEPRECATED_FEATURES)
 
    private:
+      friend class EC_Mul2Table_Data_BN;
 #endif
 
       /**
--- a/src/build-data/policy/bsi.txt
+++ b/src/build-data/policy/bsi.txt
@@ -36,7 +36,6 @@ emsa_pssr
 iso9796
 
 # pubkey
-dlies
 dh
 #dilithium // not (yet) recommended
 #dilithium_aes // not (yet) recommended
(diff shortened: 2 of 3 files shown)


FAlbertDev:
There is currently some issue with deprecated features and the `EC_Mul2Table_Data_BN` class (which uses the deprecated members). See [CI log](https://github.com/randombit/botan/actions/runs/10284183915/job/28459677296?pr=4293#step:4:166). As a quick and easy (but maybe dirty) fix, I added this class as a friend in the deprecated case. 
Maybe you want to look into this @randombit for a more elegant fix.

src/lib/pubkey/ec_group/ec_point.h
```diff
@@ -223,6 +223,7 @@ class BOTAN_PUBLIC_API(2, 0) EC_Point final {
 #if defined(BOTAN_DISABLE_DEPRECATED_FEATURES)
 
    private:
+      friend class EC_Mul2Table_Data_BN;
```
randombit, line 226: ?

FAlbertDev: > As a quick and easy (but maybe dirty) fix, I added this class as a friend in the deprecated case.

randombit: Ah right, I think this got introduced in #4211 and we missed it due to not having #4051

randombit (approved):

PR #4294, opened 2024-08-08
Fix: Kyber MSVC Warnings

FAlbertDev:
When building Botan on windows with:
```console
./configure.py --enable-shared --cc=msvc --cpu=x64 --disable-deprecated-features --werror-mode --enable-modules=kyber --module-policy=bsi --with-debug-info --msvc-runtime=MD
```
and an MSVC compiler of version 19.39, I get warnings like:
```log
warning C4244: '=': conversion from '_Ty' to 'T', possible loss of data
```
I wonder why this happens on our machine and not in Botan's CI. It could be the one version difference (we use MSVC 19.40).
Nevertheless, this PR fixes the warnings by using static_casts instead of implicit conversions.


reneme (approved):
Integer promotion. I'm sure it made sense once... 🙄

PR #4297, opened 2024-08-08
Fix: MSVC Unreachable Code Debug Warnings

FAlbertDev:
One would guess that the compiler warnings are the same for debug and release builds. It seems this is not the case for MSVC's unreachable code warnings (see [this experimental CI run](https://github.com/Rohde-Schwarz/botan/actions/runs/10300789793/job/28510934798#step:4:641)). At least some of them are sensible. This PR fixes them. 

Do we need a Windows debug build in the CI? It's weird that some warnings are only displayed in debug mode. Maybe an MSVC bug?

### DIFF
--- a/src/lib/ffi/ffi.cpp
+++ b/src/lib/ffi/ffi.cpp
@@ -129,8 +129,6 @@ int ffi_guard_thunk(const char* func_name, const std::function<int()>& thunk) {
    } catch(...) {
       return ffi_error_exception_thrown(func_name, "unknown exception");
    }
-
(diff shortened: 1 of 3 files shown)


reneme (approved):
Perhaps some sort of optimization related thing? I don't think we really need a debug build. Perhaps just build w/o -Werror in debug?

randombit:
25 minutes per build doesn’t seem worth it just for this. That’s same cost as eg sanitizer builds which have a lot more value.

PR #4303, opened 2024-08-13
Fix: Deprecated BigInt::Base Enum Warnings

FAlbertDev:
The deprecation of BigInt::Base is problematic when including _bigint.h_ in applications. Since the BigInt::Base is deprecated, we cannot declare methods like:
```cpp
BOTAN_DEPRECATED("For hex/decimal use from_string") BigInt(const uint8_t buf[], size_t length, Base base);
```
This produces errors in applications like (at least with MSVC):
```log
build/include/public\botan/bigint.h(132): warning C4996: 'Botan::BigInt::Base': All functions using this enum are deprecated
build/include/public\botan/bigint.h(888): warning C4996: 'Botan::BigInt::Base': All functions using this enum are deprecated
build/include/public\botan/bigint.h(896): warning C4996: 'Botan::BigInt::Base': All functions using this enum are deprecated
```
Instead, I propose only deprecating the enum values.

I'm working on a regression test by writing an example using the amalgamation header in amalgamation ci jobs. Similar issues should be discovered this way.

### DIFF
--- a/src/lib/math/bigint/bigint.h
+++ b/src/lib/math/bigint/bigint.h
@@ -28,10 +28,10 @@ class BOTAN_PUBLIC_API(2, 0) BigInt final {
       /**
        * Base enumerator for encoding and decoding
        */
-      enum BOTAN_DEPRECATED("All functions using this enum are deprecated") Base {
-         Decimal = 10,
-         Hexadecimal = 16,
-         Binary = 256
+      enum Base {
+         Decimal BOTAN_DEPRECATED("All functions using this enum are deprecated") = 10,
+         Hexadecimal BOTAN_DEPRECATED("All functions using this enum are deprecated") = 16,
+         Binary BOTAN_DEPRECATED("All functions using this enum are deprecated") = 256
       };
 
       /**


reneme:
> I'm working on a regression test [...]. Similar issues should be discovered this way.

Let me add that this type of warning is easily missed by our current CI. As far as I know, we don't really "consume" our headers from a user perspective: namely, without `-DBOTAN_IS_BEING_BUILT`. This hides such (likely compiler-dependent) issues.

randombit (approved):

reneme:
Leveraging our time zone difference here and fishing for early feedback: 😏 

@FAlbertDev's idea is to extend the 'amalgamation' CI target (as it practically includes all headers at once) and create a super simple example file that showcases the include of `botan_all.h`. The `make examples` target does not set `-DBOTAN_IS_BEING_BUILT`, already now. Perhaps as the only build target in the project.

Hence, compiling the examples in the amalgamation build with `-Werror` could be a reasonable way to catch such things in the future, on all three major compilers, with fairly minimal effort.

PR #4305, opened 2024-08-13
Enable the post-quantum x25519+ML-KEM-768 TLS 1.3 ciphersuite by default

randombit:
Tested with cloudflare.com, google.com and ourselves.

This adjusts the default logic for both which groups to offer and which group to select during negotiation.

For offering, we send the first pure ECC group in the preference list. This avoids sending large PQ shares to servers that don't support them. If the client for whatever reason does not have any pure ECC groups, then we send a share of whatever their top preference is.

On the server side, if the client indicated support for any mutually supported PQ algorithm, we always select it, even if the client sent some other type of keyshare. Previously we would always prefer to select a group that the client sent a share of, to reduce round trips.


src/lib/tls/tls_session.cpp
```diff
@@ -103,6 +103,43 @@ Session_Summary::Session_Summary(const Session_Base& base,
...
+   } else {
+      BOTAN_ASSERT_NOMSG(group.has_value());
+      if(group->is_dh_named_group()) {
+         return kex_method_to_string(Kex_Algo::DH);
+      } else if(group->is_ecdh_named_curve() || group->is_x25519() || group->is_x448()) {
+         return kex_method_to_string(Kex_Algo::ECDH);
+      } else if(group->is_pure_kyber()) {
+         return kex_method_to_string(Kex_Algo::KEM);
+      } else if(group->is_pqc_hybrid()) {
+         return kex_method_to_string(Kex_Algo::HYBRID);
```
reneme, line 132:
```suggestion
      } else if(group->is_kem() && !group->is_pqc_hybrid()) {
         return kex_method_to_string(Kex_Algo::KEM);
      } else if(group->is_pqc_hybrid()) {
         return kex_method_to_string(Kex_Algo::HYBRID);
```

src/tests/data/tls-policy/datagram.txt
```diff
@@ -1,23 +1,28 @@
...
-allow_dtls10 = false
 allow_dtls12 = true
+allow_ssl_key_log_file = false
 ciphers = ChaCha20Poly1305 AES-256/GCM AES-128/GCM
 macs = AEAD
 signature_hashes = SHA-512 SHA-384 SHA-256
 signature_methods = ECDSA RSA
 key_exchange_methods = ECDH DH
-key_exchange_groups = x25519 x448 secp256r1 brainpool256r1 secp384r1 brainpool384r1 secp521r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072 ffdhe/ietf/4096 ffdhe/ietf/6144 ffdhe/ietf/8192
+key_exchange_groups = x25519 secp256r1 x25519/Kyber-512-r3/cloudflare x448 secp384r1 secp521r1 brainpool256r1 brainpool384r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072
```
reneme, line 10: No need to advertise the hybrid kex here. DTLS 1.3 is not supported and hybrid kexes aren't available in 1.2.

src/tests/data/tls-policy/datagram.txt
```diff
@@ -1,23 +1,28 @@
...
+support_cert_status_message = true
+tls_13_middlebox_compatibility_mode = true
+accepted_client_certificate_types = X509
+accepted_server_certificate_types = X509
+hash_hello_random = true
+maximum_session_tickets_per_client_hello = 1
 session_ticket_lifetime = 86400
-dh_group = modp/ietf/2048
+reuse_session_tickets = false
+new_session_tickets_upon_handshake_success = 1
```
reneme, line 25: Informational: Most of those are TLS 1.3 specific policies (except `hash_hello_random`). None of these will currently have an effect on DTLS 1.2. No objection to add them, though.

src/tests/data/tls-policy/default.txt
```diff
@@ -1,15 +1,12 @@
...
 allow_tls13 = false
-allow_dtls10 = false
 allow_dtls12 = true
 ciphers = ChaCha20Poly1305 AES-256/GCM AES-128/GCM
 macs = AEAD SHA-256 SHA-384 SHA-1
 signature_hashes = SHA-512 SHA-384 SHA-256
 signature_methods = ECDSA RSA
 key_exchange_methods = ECDH DH
-key_exchange_groups = x25519 x448 secp256r1 brainpool256r1 secp384r1 brainpool384r1 secp521r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072 ffdhe/ietf/4096 ffdhe/ietf/6144 ffdhe/ietf/8192
+key_exchange_groups = x25519 secp256r1 x25519/Kyber-512-r3/cloudflare x448 secp384r1 secp521r1 brainpool256r1 brainpool384r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072
```
reneme, line 9: Let's use the recommended and most widely used X25519Kyber768Draft00: 

```suggestion
key_exchange_groups = x25519 secp256r1 x25519/Kyber-768-r3 x448 secp384r1 secp521r1 brainpool256r1 brainpool384r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072
```

src/tests/data/tls-policy/default_tls13.txt
```diff
@@ -1,22 +1,28 @@
...
-allow_dtls10 = false
 allow_dtls12 = true
+allow_ssl_key_log_file = false
 ciphers = ChaCha20Poly1305 AES-256/GCM AES-128/GCM
 macs = AEAD SHA-256 SHA-384 SHA-1
 signature_hashes = SHA-512 SHA-384 SHA-256
 signature_methods = ECDSA RSA
 key_exchange_methods = ECDH DH
-key_exchange_groups = x25519 x448 secp256r1 brainpool256r1 secp384r1 brainpool384r1 secp521r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072 ffdhe/ietf/4096 ffdhe/ietf/6144 ffdhe/ietf/8192
+key_exchange_groups = x25519 secp256r1 x25519/Kyber-512-r3/cloudflare x448 secp384r1 secp521r1 brainpool256r1 brainpool384r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072
```
reneme, line 10:
```suggestion
key_exchange_groups = x25519 secp256r1 x25519/Kyber-768-r3 x448 secp384r1 secp521r1 brainpool256r1 brainpool384r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072
```

src/tests/data/tls-policy/strict.txt
```diff
@@ -1,22 +1,28 @@
...
-allow_dtls10 = false
 allow_dtls12 = true
+allow_ssl_key_log_file = false
 ciphers = ChaCha20Poly1305 AES-256/GCM AES-128/GCM
 macs = AEAD
 signature_hashes = SHA-512 SHA-384
 signature_methods = ECDSA RSA
 key_exchange_methods = ECDH
-key_exchange_groups = x25519 x448 secp256r1 brainpool256r1 secp384r1 brainpool384r1 secp521r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072 ffdhe/ietf/4096 ffdhe/ietf/6144 ffdhe/ietf/8192
+key_exchange_groups = x25519 secp256r1 x25519/Kyber-512-r3/cloudflare x448 secp384r1 secp521r1 brainpool256r1 brainpool384r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072
```
reneme, line 10:
```suggestion
key_exchange_groups = x25519 secp256r1 x25519/Kyber-768-r3 x448 secp384r1 secp521r1 brainpool256r1 brainpool384r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072
```

src/tests/data/tls-policy/strict_tls13.txt
```diff
@@ -1,22 +1,28 @@
...
-allow_dtls10 = false
 allow_dtls12 = true
+allow_ssl_key_log_file = false
 ciphers = ChaCha20Poly1305 AES-256/GCM AES-128/GCM
 macs = AEAD
 signature_hashes = SHA-512 SHA-384
 signature_methods = ECDSA RSA
 key_exchange_methods = ECDH
-key_exchange_groups = x25519 x448 secp256r1 brainpool256r1 secp384r1 brainpool384r1 secp521r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072 ffdhe/ietf/4096 ffdhe/ietf/6144 ffdhe/ietf/8192
+key_exchange_groups = x25519 secp256r1 x25519/Kyber-512-r3/cloudflare x448 secp384r1 secp521r1 brainpool256r1 brainpool384r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072
```
reneme, line 10:
```suggestion
key_exchange_groups = x25519 secp256r1 x25519/Kyber-768-r3 x448 secp384r1 secp521r1 brainpool256r1 brainpool384r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072
```

src/lib/tls/tls_algos.cpp
```diff
@@ -134,6 +134,82 @@ Auth_Method auth_method_from_string(std::string_view str) {
...
+
+   return true;
+}
+
+std::optional<Group_Params_Code> Group_Params::pqc_hybrid_ecc() const {
+   switch(m_code) {
+      case Group_Params_Code::HYBRID_X25519_KYBER_512_R3_CLOUDFLARE:
+      case Group_Params_Code::HYBRID_X25519_KYBER_512_R3_OQS:
+      case Group_Params_Code::HYBRID_X25519_KYBER_768_R3_OQS:
+
```
reneme, line 182:
```suggestion
```

reneme (approved):
🚀 Nice! The PQ-era is here!

As a side note: This points at an open tech debt regarding the management of ciphersuites (also noted for a long time in https://github.com/randombit/botan/issues/2990). As it is now, there's a lot of `if {} else if {} else` scattered around many places in the TLS implementation. Whenever algorithms are added, there's a lot of potential to miss something.

src/lib/tls/tls_policy.cpp
```diff
@@ -161,20 +161,33 @@ Group_Params Policy::default_dh_group() const {
...
    return {
+      // clang-format off
 #if defined(BOTAN_HAS_X25519)
       Group_Params::X25519,
 #endif
+
+      Group_Params::SECP256R1,
+
+#if defined(BOTAN_HAS_X25519) && defined(BOTAN_HAS_KYBER_ROUND3) && defined(BOTAN_HAS_TLS_13_PQC)
+      Group_Params::HYBRID_X25519_KYBER_512_R3_CLOUDFLARE,
```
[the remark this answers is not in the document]
randombit: Oh thanks! I tried `HYBRID_X25519_KYBER_512_R3_OQS` with eg gmail.com and that didn't work so I assumed Google hadn't rolled out PQ yet.

src/tests/data/tls-policy/datagram.txt
randombit: No need I agree - it's just a fallout of how the Policy is defined; it doesn't know if we are trying to negotiate 1.2 or 1.3

randombit:
Something is wrong - I know I tested that our client connecting to our server would negotiate Kyber (highly relevant since Botan being on both sides of the connection is very common) but now this doesn't work ....

randombit:
Oh I see. I tested us<->us in my initial attempt which put Kyber at the absolute top of the preference list. However I realized later this caused us to send a large keyshare that is ignored much of the time, which is non-optimal.

But it seems - unlike the stacks in `google.com` and `cloudflare.com` - we will, if we receive a keyshare for say x25519, we won't ignore it even if the top preference of both the client and the server are Kyber - we just negotiate x25519. So we won't negotiate Kyber with ourselves :sob:

randombit:
OK I think we just need to adjust the logic in `Policy::choose_key_exchange_group`.

reneme:
> OK I think we just need to adjust the logic in Policy::choose_key_exchange_group.

... I was about to say that. Currently, the code there optimizes for round-trips and avoids sending a HelloRetryRequest whenever it can. I.e.: it will go for the non-PQ group if it is offered and fits with our needs.

randombit:
I'll take a look at that BoringSSL and co do here. I expect it looks something like a 2-tier selection

* If both peers share a PQC algorithm of any kind, then we're using PQC. If the client offered a PQC share, use that (even if it's not our favorite PQC). Otherwise choose a PQC, either clients favorite or servers favorite, depending on `server_uses_own_ciphersuite_preferences`.

* If there is no mutually agreed upon PQC group, then use effectively the existing logic.

reneme:
> I expect it looks something like a 2-tier selection

Sounds reasonable to me as a default policy. I was thinking to propose an additional policy setting like `prefer_pqc_groups_when_possible()`. But then again: if one is explicitly offering PQC support at this time, I guess they are also fine with using it anyway.

randombit:
`prefer_pqc_groups_when_possible()` might be worth having as an easy to use toggle for those who are willing to use PQC but would like to save the round trip where possible.

randombit:
OTOH we already have a lot of fucking policy toggles :sweat_smile:

src/lib/tls/tls_policy.cpp
```diff
@@ -192,13 +192,22 @@ std::vector<Group_Params> Policy::key_exchange_groups() const {
...
+   Likewise we skip DH since the keys are large
+   */
+   for(auto group: key_exchange_groups()) {
+      if(group.is_pure_ecc_group()) {
+         return {group};
+      }
    }
-   return groups_to_offer;
+   // Send no initial key share
+   return {};
```
reneme, line 210: Mhh, frankly, if we reach here, we should probably just offer whatever is the "most preferred group" instead of not offering anything.

I'm thinking of specialized applications that configure a short list of non-ECC groups for some reason. They would always end up with a HelloRetryRequest roundtrip for no obvious reason. Sure, such applications could override this policy... but do they care? And how long will it take them to notice the performance regression otherwise?

src/lib/tls/tls_session.cpp
```diff
@@ -138,39 +175,26 @@ Session_Summary::Session_Summary(const Server_Hello_13& server_hello,
...
-         } else if(group.is_ecdh_named_curve() || group.is_x25519() || group.is_x448()) {
-            return Kex_Algo::ECDH;
-         } else if(group.is_pure_kyber()) {
-            return Kex_Algo::KEM;
-         } else if(group.is_pqc_hybrid()) {
-            return Kex_Algo::HYBRID;
-         }
+         return keyshare->selected_group();
       }
+   }();
```
reneme, line 191: Now, this could probably also just be:

```C++
   const auto group = [&]() -> std::optional<Named_Group> {
      if(const auto keyshare = server_hello.extensions().get<Key_Share>()) {
         return keyshare->selected_group();
      } else {
         return {};
      }
   }();
```

This removes the assertion in the non-PSK case, but if the handshake made it all the way to this session summary, that won't stop any bug here anymore, I guess. Edit: also, there's an equivalent assertion in (`tls13_kex_to_string()`, currently line 124).

src/tests/unit_tls_policy.cpp
```diff
@@ -140,15 +140,27 @@ class TLS_Policy_Unit_Tests final : public Test {
          Botan::TLS::Policy default_policy;
          result.test_eq(
             "default TLS Policy offers exactly one", default_policy.key_exchange_groups_to_offer().size(), 1);
-         result.confirm(
-            "default TLS Policy offers preferred group",
-            default_policy.key_exchange_groups().front() == default_policy.key_exchange_groups_to_offer().front());
+
+         auto first_pure_ecc_group =
+            [](const std::vector<Botan::TLS::Named_Group>& groups) -> std::optional<Botan::TLS::Named_Group> {
```
reneme, line 145: Nit: `std::span<>` would work here as well and erase the `std::vector<>` type, if we were to return any other type from `Policy::key_exchange_groups()` for some reason in the future.

```suggestion
            [](std::span<const Botan::TLS::Named_Group> groups) -> std::optional<Botan::TLS::Named_Group> {
```

src/lib/tls/tls_session.h
```diff
@@ -288,6 +288,8 @@ class BOTAN_PUBLIC_API(3, 0) Session_Summary : public Session_Base {
 
       std::string kex_algo() const { return m_kex_algo; }
 
+      std::optional<std::string> kex_parameters() const { return m_kex_parameters; }
```
reneme, line 291:
```suggestion
      const std::optional<std::string>& kex_parameters() const { return m_kex_parameters; }
```

... there might even be some `clang-tidy` hiccup about that, I think.

src/lib/tls/tls_session.cpp
```diff
@@ -138,39 +175,26 @@ Session_Summary::Session_Summary(const Server_Hello_13& server_hello,
...
+
+   if(group.has_value()) {
+      m_kex_parameters = group->to_string();
+   }
```
reneme, line 195: C++23 will bring [monadic stuff to `std::optional<>`](https://en.cppreference.com/w/cpp/utility/optional/transform) finally. Along those lines:

```C++
m_kex_parameters = group.transform([](auto& x) { return x.to_string(); });
```

... the bulky C++ lambdas somewhat spoil the fun; still better than three lines of `if`, I guess.

reneme (approved):
BoGo isn't happy about our new cleverness in selecting KEX algos, I guess. Probably, we'll just have to override the policy with the old logic there.

src/lib/tls/tls_session.cpp
randombit: I'm definitely looking forward to it, I miss Rust's monadic operations on `std::optional` a lot of the time working in C++

As for pattern matching (sigh)

src/lib/tls/tls_policy.cpp
```diff
@@ -161,31 +161,53 @@ Group_Params Policy::default_dh_group() const {
 }
 
 std::vector<Group_Params> Policy::key_exchange_groups() const {
-   // Default list is ordered by performance
    return {
+      // clang-format off
+#if defined(BOTAN_HAS_X25519) && defined(BOTAN_HAS_KYBER_ROUND3) && defined(BOTAN_HAS_TLS_13_PQC)
+      Group_Params::HYBRID_X25519_KYBER_768_R3_OQS,
+#endif
```
reneme, line 168: Probably, now this should offer ML-KEM instead. I'm expecting the community to move away from the pre-standard algorithm fairly quickly.

randombit: Concur, especially since this change isn't going to be going out until February

src/lib/tls/tls_session.cpp
```diff
@@ -103,6 +103,43 @@ Session_Summary::Session_Summary(const Session_Base& base,
 
 #if defined(BOTAN_HAS_TLS_13)
 
+namespace {
+
+std::string tls13_kex_to_string(bool psk, std::optional<Named_Group> group) {
+   if(psk && group) {
+      if(group->is_dh_named_group()) {
+         return kex_method_to_string(Kex_Algo::DHE_PSK);
+      } else if(group->is_ecdh_named_curve() || group->is_x25519() || group->is_x448()) {
```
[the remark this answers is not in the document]
randombit: We already had `is_pure_ecc_group` for this

src/tests/data/tls-policy/default.txt
```diff
@@ -1,15 +1,12 @@
...
 allow_tls13 = false
-allow_dtls10 = false
 allow_dtls12 = true
 ciphers = ChaCha20Poly1305 AES-256/GCM AES-128/GCM
 macs = AEAD SHA-256 SHA-384 SHA-1
 signature_hashes = SHA-512 SHA-384 SHA-256
 signature_methods = ECDSA RSA
 key_exchange_methods = ECDH DH
-key_exchange_groups = x25519 x448 secp256r1 brainpool256r1 secp384r1 brainpool384r1 secp521r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072 ffdhe/ietf/4096 ffdhe/ietf/6144 ffdhe/ietf/8192
+key_exchange_groups = x25519 secp256r1 x25519/Kyber-768-r3 x448 secp384r1 secp521r1 brainpool256r1 brainpool384r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072
```
reneme, line 9: That should use ML-KEM now, no? Just like the default_tls13.txt

```suggestion
key_exchange_groups = x25519 secp256r1 x25519/ML-KEM-768 x448 secp384r1 secp521r1 brainpool256r1 brainpool384r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072
```

src/tests/data/tls-policy/strict.txt
```diff
@@ -1,22 +1,28 @@
...
-allow_dtls10 = false
 allow_dtls12 = true
+allow_ssl_key_log_file = false
 ciphers = ChaCha20Poly1305 AES-256/GCM AES-128/GCM
 macs = AEAD
 signature_hashes = SHA-512 SHA-384
 signature_methods = ECDSA RSA
 key_exchange_methods = ECDH
-key_exchange_groups = x25519 x448 secp256r1 brainpool256r1 secp384r1 brainpool384r1 secp521r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072 ffdhe/ietf/4096 ffdhe/ietf/6144 ffdhe/ietf/8192
+key_exchange_groups = x25519/Kyber-768-r3 x25519 secp256r1 x448 secp384r1 secp521r1 brainpool256r1 brainpool384r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072
```
reneme, line 10:
```suggestion
key_exchange_groups = x25519/ML-KEM-768 x25519 secp256r1 x448 secp384r1 secp521r1 brainpool256r1 brainpool384r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072
```

src/lib/tls/tls_session.cpp
randombit: I adjusted the lambda as you suggest but added an assert later in the function where we examine if `group` has a value or not, since if this situation ever did happen the eventual fatal error (wherever it would occur) would probably be confusing and difficult to debug without it.

src/lib/tls/tls_policy.cpp
```diff
@@ -161,31 +183,63 @@ Group_Params Policy::default_dh_group() const {
...
 
 std::vector<Group_Params> Policy::key_exchange_groups_to_offer() const {
-   // by default, we offer a key share for the most-preferred group, only
-   std::vector<Group_Params> groups_to_offer;
-   const auto supported_groups = key_exchange_groups();
-   if(!supported_groups.empty()) {
-      groups_to_offer.push_back(supported_groups.front());
+   /*
+   By default, we offer a key share for the most-preferred pure ECC group
+   by default, if any pure ECC group is enabled in the policy.
```
reneme, line 219:
```suggestion
   By default, we offer a key share for the most-preferred pure ECC group
   if any pure ECC group is enabled in the policy.
```

src/lib/tls/tls_algos.h
```diff
@@ -198,7 +203,7 @@ class BOTAN_PUBLIC_API(3, 2) Group_Params final {
                 m_code == Group_Params_Code::FFDHE_8192;
       }
 
-      constexpr bool is_pure_kyber() const {
+      constexpr bool is_pure_kyber_r3() const {
```
reneme, line 206: This is a public API. I think we'll have to go with deprecation of `is_pure_kyber()` here.

reneme (changes requested):
Let's not start offering Kyber by default anymore.

src/tests/data/tls-policy/strict.txt
randombit: We're not really offering Kyber here. It being in the policy text is a leftover from earlier iterations of the PR. But now the Kyber/MKLEM suites are being filtered out and I don't understand why.

randombit: Something was behaving weirdly here because the tests for me passed with Kyber in the expected policy text (even though the policy doesn't include Kyber anymore) and then started failing when I switched it to ML-KEM. No idea what happened, went away after a clean rebuild.

randombit:
@reneme just wanted to ping on this, ideally this PR and #4393 would both merge, and then I'll remove the r3 support.

src/lib/tls/tls_session.cpp
```diff
@@ -103,6 +103,43 @@ Session_Summary::Session_Summary(const Session_Base& base,
...
+      }
+   } else if(psk) {
+      return kex_method_to_string(Kex_Algo::PSK);
+   } else {
+      BOTAN_ASSERT_NOMSG(group.has_value());
+      if(group->is_dh_named_group()) {
+         return kex_method_to_string(Kex_Algo::DH);
+      } else if(group->is_ecdh_named_curve() || group->is_x25519() || group->is_x448()) {
+         return kex_method_to_string(Kex_Algo::ECDH);
+      } else if(group->is_pure_kyber()) {
```
reneme, line 129:
```suggestion
      } else if(group->is_pure_kyber() || group->is_pure_frodokem()) {
```

src/lib/tls/tls_session.cpp
```diff
@@ -103,6 +103,43 @@ Session_Summary::Session_Summary(const Session_Base& base,
...
 
+namespace {
+
+std::string tls13_kex_to_string(bool psk, std::optional<Named_Group> group) {
+   if(psk && group) {
+      if(group->is_dh_named_group()) {
+         return kex_method_to_string(Kex_Algo::DHE_PSK);
+      } else if(group->is_ecdh_named_curve() || group->is_x25519() || group->is_x448()) {
+         return kex_method_to_string(Kex_Algo::ECDHE_PSK);
+      } else if(group->is_pure_kyber()) {
```
reneme, line 114:
```suggestion
      } else if(group->is_pure_kyber() || group->is_pure_frodokem()) {
```

src/tests/data/tls-policy/default.txt
reneme, line 9: This should now be hybrid ML-KEM, no?

```suggestion
key_exchange_groups = x25519 secp256r1 x25519/ML-KEM-768 x448 secp384r1 secp521r1 brainpool256r1 brainpool384r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072
```

src/tests/data/tls-policy/strict.txt
reneme, line 10:
```suggestion
key_exchange_groups = x25519 secp256r1 x25519/ML-KEM-768 x448 secp384r1 secp521r1 brainpool256r1 brainpool384r1 brainpool512r1 ffdhe/ietf/2048 ffdhe/ietf/3072
```

reneme (approved):
Note that `tls13_kex_to_string` will need to be adapted when merged with #4393, due to the introduction of `is_pure_ml_kem()` in there. It's probably a good idea to merge that first and then rebase this PR onto it.

src/tests/data/tls-policy/default.txt
randombit: Oops not sure how that happened

randombit:
@reneme Rebased after merging #4393

reneme (approved):
I'd suggest to split Kyber and ML-KEM explicitly in `tls_algos.h`. Otherwise, this looks good to me. Thanks, and Merry Christmas! 🎄

src/lib/tls/tls_algos.h
```diff
@@ -242,7 +247,7 @@ class BOTAN_PUBLIC_API(3, 2) Group_Params final {
          BOTAN_DIAGNOSTIC_POP
       }
 
-      constexpr bool is_pqc_hybrid() const {
+      constexpr bool is_pqc_hybrid_kyber() const {
```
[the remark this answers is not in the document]
randombit: Removed the ML-KEM ids from the Kyber (-r3) predicate

PR #4307, opened 2024-08-15
CI: Amalgamation Linkage Test and Fixes

FAlbertDev:
Warnings like the one in #4303 or https://github.com/randombit/botan/pull/4291#discussion_r1782486246 are currently not checked by our CI. This is because we do not test including Botan headers while the `BOTAN_IS_BEING_BUILT` macro is undefined. The only target that is compiled without this flag is _examples_.
To fix this shortcoming, I propose adding the examples target to the amalgamation CI build and adding an example that includes Botan's amalgamation header. This has the following advantages:
- Warnings in headers that depend on `BOTAN_IS_BEING_BUILT` are caught by CI (like the one in PR 4303)
- We can test the _examples_ target on different operating systems.
- Since the amalgamation builds are quite fast compared to the normal ones, the additional overhead by adding boost capabilities and building the examples is not too painful.

As you can see [in this experimental CI run](https://github.com/Rohde-Schwarz/botan/actions/runs/10401063766/job/28802820988#step:4:903), the issue of #4303 is found with this setup (seems to be an exclusive MSVC "feature"). Also, I found a few other issues in the examples when using macOS. Since these are related to TLS, @reneme, could you check and verify my proposed fixes?

Here are the errors for the fixed code sections:
<details>
  <summary>src/examples/tls_ssl_key_log_file.cpp:</summary>
  
  [CI log:](https://github.com/Rohde-Schwarz/botan/actions/runs/10400356216/job/28800747414#step:4:250)
  
  ```log
  ccache clang++ -fstack-protector -m64 -pthread -stdlib=libc++ -std=c++20 -D_REENTRANT  -O3 --system-header-prefix=boost/ -Wall -Wextra -Wpedantic -Wshadow -Wstrict-aliasing -Wstrict-overflow=5 -Wcast-align -Wmissing-declarations -Wpointer-arith -Wcast-qual -Wshorten-64-to-32 -Wcomma -Wdocumentation -Werror -Wno-error=unused-parameter -Wno-error=unreachable-code -Wno-error=unused-lambda-capture  -I build/include/public  -isystem build/include/external -isystem /usr/local/opt/boost/include -c src/examples/x509_path.cpp -o build/obj/examples/x509_path.o
  src/examples/tls_ssl_key_log_file.cpp:204:15: error: no type named 'close' in the global namespace
              ::close(fd);
              ~~^
  src/examples/tls_ssl_key_log_file.cpp:204:21: error: declaration shadows a field of '(anonymous namespace)::DtlsConnection' [-Werror,-Wshadow]
              ::close(fd);
                      ^
  src/examples/tls_ssl_key_log_file.cpp:132:11: note: previous declaration is here
        int fd;
            ^
  2 errors generated.
  make: *** [build/obj/examples/tls_ssl_key_log_file.o] Error 1
  ```

</details>
<details>
  <summary>src/examples/tls_stream_coroutine_client.cpp:</summary>
  
  [CI log:](https://github.com/Rohde-Schwarz/botan/actions/runs/10400805898/job/28802067085#step:4:298)
  
  ```log
  ccache clang++ -fstack-protector -pthread -stdlib=libc++ build/obj/examples/tls_stream_coroutine_client.o -L.  -O3 --system-header-prefix=boost/ -lbotan-3 -lbz2 -ldl -lsqlite3 -lz -framework CoreFoundation -framework Security   -o build/examples/tls_stream_coroutine_client
  Undefined symbols for architecture arm64:
    "Botan::TLS::Context::Context(Botan::TLS::Server_Information)", referenced from:
        void std::__1::allocator_traits<std::__1::allocator<Botan::TLS::Context>>::construct[abi:ue170006]<Botan::TLS::Context, Botan::TLS::Server_Information, void, void>(std::__1::allocator<Botan::TLS::Context>&, Botan::TLS::Context*, Botan::TLS::Server_Information&&) in tls_stream_coroutine_client.o
  ld: symbol(s) not found for architecture arm64
  clang: error: linker command failed with exit code 1 (use -v to see invocation)
  make: *** [build/examples/tls_stream_coroutine_client] Error 1
  make: Target `examples' not remade because of errors.
  ```

</details>

### DIFF
--- a/src/examples/amalgamation_header.cpp
+++ b/src/examples/amalgamation_header.cpp
@@ -1,0 +1,13 @@
+#include <iostream>
+#if __has_include(<botan/botan_all.h>)
+   #include <botan/botan_all.h>
+#else
+   // if the amalgamation header isn't available, you have to IWYU.
+   #include <botan/hex.h>
+#endif
+
+int main() {
+   std::cout << "With an amalgamation build you can include everything using the botan_all header.\n";
+   std::cout << "That's " << Botan::hex_encode(std::vector<uint8_t>{0xC0, 0x01}) << "\n";
+   return 0;
+}
--- a/src/examples/tls_ssl_key_log_file.cpp
+++ b/src/examples/tls_ssl_key_log_file.cpp
@@ -23,6 +23,9 @@
    #include <sys/ioctl.h>
    #include <sys/socket.h>
 #endif
+#if defined(BOTAN_TARGET_OS_HAS_POSIX1)
+   #include <unistd.h>
+#endif
 
 namespace {
 
@@ -142,6 +145,8 @@ class DtlsConnection : public Botan::TLS::Callbacks {
          remote_addr.sin_family = AF_INET;
          inet_aton(r_addr.c_str(), &remote_addr.sin_addr);
          remote_addr.sin_port = htons(r_port);
+#else
+         BOTAN_UNUSED(r_addr, r_port);
 #endif
          auto tls_callbacks_proxy = std::make_shared<BotanTLSCallbacksProxy>(*this);
          auto rng = std::make_shared<Botan::AutoSeeded_RNG>();
@@ -169,6 +174,7 @@ class DtlsConnection : public Botan::TLS::Callbacks {
 #if defined(BOTAN_TARGET_OS_HAS_SOCKETS)
          sendto(fd, data.data(), data.size(), 0, reinterpret_cast<const sockaddr*>(&remote_addr), sizeof(sockaddr_in));
 #else
+         BOTAN_UNUSED(data);
          // send data to the other side
          // ...
 #endif
@@ -201,7 +207,9 @@ class DtlsConnection : public Botan::TLS::Callbacks {
          if(fd) {
 #if defined(BOTAN_TARGET_OS_HAS_SOCKETS)
             shutdown(fd, SHUT_RDWR);
+   #if defined(BOTAN_TARGET_OS_HAS_POSIX1)
             ::close(fd);
+   #endif
 #endif
          }
       }
--- a/src/examples/tls_stream_coroutine_client.cpp
+++ b/src/examples/tls_stream_coroutine_client.cpp
@@ -7,6 +7,10 @@
 // in clang 14 and newer. Older versions of Boost might work with other
 // compilers, though.
 #if defined(BOTAN_FOUND_COMPATIBLE_BOOST_ASIO_VERSION) && BOOST_VERSION >= 108100
+   #define BOOST_VERSION_IS_COMPATIBLE
+#endif
+
+#if defined(BOOST_VERSION_IS_COMPATIBLE) && defined(BOTAN_HAS_HAS_DEFAULT_TLS_CONTEXT)
 
    #include <botan/asio_stream.h>
    #include <botan/version.h>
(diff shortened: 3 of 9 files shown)


reneme:
> Since the amalgamation builds are quite fast compared to the normal ones, the additional overhead by adding boost capabilities and building the examples is not too painful.

Let me add that there's value in building the amalgamation with boost enabled in and by itself. There's a good chance that we break the amalgamation in headers that require boost (e.g. the TLS adapter).

src/examples/tls_stream_coroutine_client.cpp
```diff
@@ -3,10 +3,17 @@
...
+int main() {
+   std::cout << "Your system needs an auto seeded RNG and a certificate store.\n";
+   return 1;
+}
+
 // Boost 1.81.0 introduced support for the finalized C++20 coroutines
 // in clang 14 and newer. Older versions of Boost might work with other
 // compilers, though.
-#if defined(BOTAN_FOUND_COMPATIBLE_BOOST_ASIO_VERSION) && BOOST_VERSION >= 108100
+#elif defined(BOTAN_FOUND_COMPATIBLE_BOOST_ASIO_VERSION) && BOOST_VERSION >= 108100
```
reneme, line 16: Perhaps just check for the default TLS context macro here as well and update the fallback-error message at the very end of the example to be more ambigous.

src/examples/tls_stream_coroutine_client.cpp
```diff
@@ -43,6 +50,7 @@ net::awaitable<void> request(std::string host, std::string port, std::string tar
    const auto dns_result = co_await resolver.async_resolve(host, port);
 
    // Connect to host and establish a TLS session
+   // TODO: This constructor needs BOTAN_HAS_HAS_DEFAULT_TLS_CONTEXT. Is there any other way?
```
reneme, line 53: remove TODO?

src/examples/amalgamation_header.cpp
```diff
@@ -0,0 +1,15 @@
+#include <iostream>
+#if __has_include(<botan/botan_all.h>)
+   #include <botan/botan_all.h>
```
reneme, line 3:
```suggestion
#if __has_include(<botan/botan_all.h>)
   #include <botan/botan_all.h>
#else
   // if the amalgamation header isn't available, you have to IWYU.
   #include <botan/hex.h>
#endif
```

... no need for the fallback main function then.

src/examples/tls_ssl_key_log_file.cpp
```diff
@@ -137,7 +140,11 @@ class DtlsConnection : public Botan::TLS::Callbacks {
       std::function<void()> activated_callback;
 
    public:
-      DtlsConnection(const std::string& r_addr, int r_port, int socket, bool is_server) : fd(socket) {
+      DtlsConnection([[maybe_unused]] const std::string& r_addr,
+                     [[maybe_unused]] int r_port,
```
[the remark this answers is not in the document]
reneme: My vote is for `BOTAN_UNUSED` and removing `[[maybe_unused]]` in those cases. Certainly not both. 🙂

src/examples/tls_ssl_key_log_file.cpp
```diff
@@ -165,7 +172,7 @@ class DtlsConnection : public Botan::TLS::Callbacks {
          }
       }
 
-      void tls_emit_data(std::span<const uint8_t> data) override {
+      void tls_emit_data([[maybe_unused]] std::span<const uint8_t> data) override {
```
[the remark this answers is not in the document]
reneme: Same as above. Let's not do both....

reneme (approved):
Nice! Thanks. A bit sad about the compiler pleasing in the examples, as it is potentially also incorporated into our handbook. But the alternative would be to lower the warning-levels for examples, which isn't desirable either, I guess? @randombit

reneme (approved):

randombit (approved):

PR #4309, opened 2024-08-15
Add Entropy Source and DRNG Manager (ESDM) RNG support

thillux:
ESDM is a Linux-based user-space PRNG daemon, with configurable entropy sources.

See: https://github.com/smuellerDD/esdm

It currently gets used, when a high level of control over entropy sources is desirable, e.g. for VPN appliance solutions.

In order to use this source, the ESDM server daemon has to be running and botan must be configured --with-esdm.

ESDM currently works only on Linux.

### DIFF
--- a/src/lib/rng/esdm_rng/esdm_rng.cpp
+++ b/src/lib/rng/esdm_rng/esdm_rng.cpp
@@ -1,0 +1,104 @@
+/*
+* ESDM RNG
+* (C) 2024, Markus Theil <theil.markus@gmail.com>
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/esdm_rng.h>
+
+#include <esdm/esdm_rpc_client.h>
+#include <mutex>
+
+namespace Botan {
+
+namespace {
+/**
+* This helper makes sure that the ESDM service is initialized and
+* finalized as needed in a threadsafe fashion. Finalization happens
+* as soon as all instances of ESDM_RNG are destructed. This may
+* happen multiple times in the lifetime of the process.
+*/
+class ESDM_Context {
+   public:
+      [[nodiscard]] static std::shared_ptr<void> instance() {
+         static ESDM_Context g_instance;
+         return g_instance.acquire();
+      }
+
+   private:
+      ESDM_Context() = default;
+
+      [[nodiscard]] std::shared_ptr<void> acquire() {
+         std::scoped_lock lk(m_mutex);
+         if(m_refs++ == 0) {
+            if(esdm_rpcc_init_unpriv_service(nullptr) != 0) {
+               throw Botan::System_Error("unable to initialize ESDM unprivileged service");
+            }
+         }
+         return std::shared_ptr<void>{nullptr, [this](void*) { this->release(); }};
+      }
+
+      void release() {
+         std::scoped_lock lk(m_mutex);
+         if(m_refs-- == 1) {
+            esdm_rpcc_fini_unpriv_service();
+         }
+      }
+
+   private:
+      std::mutex m_mutex;
+      size_t m_refs = 0;
+};
+}  // namespace
+
+ESDM_RNG::ESDM_RNG(bool prediction_resistance) :
+      m_prediction_resistance(prediction_resistance), m_ctx(ESDM_Context::instance()) {}
+
+void ESDM_RNG::fill_bytes_with_input(std::span<uint8_t> out, std::span<const uint8_t> in) {
+   /*
+   * This variable is implicitly set by the "esdm_invoke" macro that comes
+   * with the ESDM library. "esdm_invoke" implements a retry mechanism for
+   * the underlying RPC mechanism of ESDM.
+   */
+   ssize_t ret = 0;
+
+   if(!in.empty()) {
+      ret = 0;
+
+      /*
+      * take additional input, but do not account entropy for it,
+      * as this information is not included in the API
+      */
+      esdm_invoke(esdm_rpcc_write_data(in.data(), in.size()));
+      /*
+      * ret was set by esdm_invoke, as mentioned above
+      */
+      if(ret != 0) {
+         throw Botan::System_Error("Writing additional input to ESDM failed");
+      }
+   }
+
+   if(!out.empty()) {
+      ret = 0;
+
+      if(m_prediction_resistance) {
+         esdm_invoke(esdm_rpcc_get_random_bytes_pr(out.data(), out.size()));
+      } else {
+         esdm_invoke(esdm_rpcc_get_random_bytes_full(out.data(), out.size()));
+      }
+      /*
+      * ret was set by esdm_invoke, as mentioned above
+      */
+      if(ret != static_cast<ssize_t>(out.size())) {
+         throw Botan::System_Error("Fetching random bytes from ESDM failed");
+      }
+   }
+}
+
+RandomNumberGenerator& esdm_rng() {
+   static ESDM_RNG g_esdm_rng;
+   return g_esdm_rng;
+}
+
+}  // namespace Botan
--- a/src/lib/rng/esdm_rng/esdm_rng.h
+++ b/src/lib/rng/esdm_rng/esdm_rng.h
@@ -1,0 +1,111 @@
+/*
+* ESDM RNG
+* (C) 2024, Markus Theil <theil.markus@gmail.com>
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#ifndef BOTAN_ESDM_RNG_H_
+#define BOTAN_ESDM_RNG_H_
+
+#include <botan/rng.h>
+#include <memory>
+
+namespace Botan {
+
+/**
+* Return a shared reference to a global PRNG instance provided by ESDM
+*/
+BOTAN_PUBLIC_API(3, 7) RandomNumberGenerator& esdm_rng();
+
+/**
+* Entropy Source and DRNG Manager (ESDM) is a Linux/Unix based
+* PRNG manager, which can be seeded from NTG.1/SP800-90B sources.
+*
+* See:
+*  - Repository: https://github.com/smuellerDD/esdm
+*  - Further Docs: https://www.chronox.de/esdm/index.html
+*
+* Different entropy sources can be configured in respect of being
+* active and how much entropy is accounted for each of them.
+*
+* ESDM tracks its seed and reseed status and blocks, when not fully seeded.
+* For this functionality, the esdm_rpcc_get_random_bytes_pr (prediction resistant) or
+* esdm_rpcc_get_random_bytes_full (fully seeded) calls have to be used.
+*
+* Configurable modes:
+*   - fully seeded (-> fast): provide entropy from a DRBG/PRNG after beeing fully seeded,
+*          block until this point is reached, reseed from after a time
+*          and/or invocation limit, block again if reseeding is not possible
+*   - prediction resistance (-> slow): reseed ESDM with fresh entropy after each invocation
+*
+* You typically want to use the fast fully seeded mode, which is the default.
+*
+* Instances of this class communicate over RPC with ESDM. The esdm_rpc_client
+* library, provided by ESDM, is leveraged for this.
+*
+* Thread safety:
+*   It is fine to construct, destruct and use objects of this class concurrently.
+*   The communication with ESDM is thread-safe, as handled by esdm_rpc_client.
+*   The initialization of esdm_rpc_client is not thread safe, therefore this class
+*   takes care of it, with its embedded ESDM_Context.
+*/
+class BOTAN_PUBLIC_API(3, 7) ESDM_RNG final : public Botan::RandomNumberGenerator {
+   public:
+      /**
+      * Default constructor for ESDM, fully seeded mode
+      */
+      ESDM_RNG() : ESDM_RNG(false) {}
+
+      /**
+      * Construct ESDM instance with configurable mode
+      */
+      explicit ESDM_RNG(bool prediction_resistance);
+
+      std::string name() const override {
+         if(m_prediction_resistance) {
+            return "esdm-pr";
+         } else {
+            return "esdm-full";
+         }
+      }
+
+      /**
+      * ESDM blocks, if it is not seeded,
+      *
+      * @return true
+      */
+      bool is_seeded() const override { return true; }
+
+      /**
+      * ESDM can inject additional inputs
+      * but we do not account entropy for it
+      *
+      * @return true
+      */
+      bool accepts_input() const override { return true; }
+
+      /**
+      * the ESDM RNG does not hold any state outside ESDM, that should be cleared
+      * here
+      */
+      void clear() override {}
+
+   protected:
+      void fill_bytes_with_input(std::span<uint8_t> out, std::span<const uint8_t> in) override;
+
+   private:
+      /**
+      * tracks if predicition resistant or fully seeded interface should be queried
+      */
+      bool m_prediction_resistance;
+
+      /**
+      * takes care of thread-safe esdm_rpc_client initialization
+      */
+      std::shared_ptr<void> m_ctx;
+};
+
+}  // namespace Botan
+
+#endif /* BOTAN_ESDM_RNG_H_ */
--- a/src/scripts/ci/setup_gh_actions.sh
+++ b/src/scripts/ci/setup_gh_actions.sh
@@ -29,6 +29,19 @@ function build_and_install_jitterentropy() {
     rm -rf "${jel_dir}"
 }
 
+function build_and_install_esdm() {
+    # build dependencies
+    sudo apt-get -qq install libprotobuf-c-dev meson
+
+    # download, build and install ESDM
+    curl -L "https://github.com/smuellerDD/esdm/archive/refs/tags/v${ESDM_VERSION}.tar.gz" | tar -xz -C .
+    pushd "$(realpath esdm-*)"
+    meson setup build -Dselinux=disabled -Dais2031=false -Dlinux-devfiles=disabled -Des_jent=disabled --prefix=/usr --libdir=lib
+    meson compile -C build
+    sudo meson install -C build
+    popd
+}
+
 if [ -z "$REPO_CONFIG_LOADED" ]; then
     echo "Repository configuration not loaded" >&2
     exit 1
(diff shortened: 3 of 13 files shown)


thillux:
cc @smuellerDD

thillux:
I tested the FFI-Bindings against https://github.com/thillux/botan-rs/tree/esdm-rng.

thillux:
@reneme I reformatted the code with clang-format, after looking at the CI results.

reneme:
> I reformatted the code with clang-format, after looking at the CI results.

Thanks! FYI: There are some Emacs, VS Code and .editorconfig config files in [src/editors](https://github.com/randombit/botan/tree/master/src/editors). In case you use any of this, you might want to symlink them into the repo's root dir and benefit from things like "format on save".

src/lib/rng/esdm_rng/esdm_rng.h
```diff
@@ -0,0 +1,107 @@
...
+      * ESDM rpc client locks concurrent accesses, but initialization should be
+      * only done once
+      */
+      inline static std::mutex m_init_lock;
+
+      /**
+      * counts how many ESDM RNG instances are active in order to perform init and
+      * fini on first/last one
+      */
+      inline static size_t m_ref_cnt = 0;
```
reneme, line 102: If we need a singleton, we should probably implement a singleton. 😏 For instance: implement an internal `ESDM_Context` class that lives in `esdm_rng.cpp`, and provides a static factory method instead of a public constructor. Instances of `ESDM_RNG` could then hold a shared pointer of this context. The factory method could hold a weak pointer to that instance, so that the `ESDM_Context` is destructed as soon as the last `ESDM_RNG` dies. Like so:

https://godbolt.org/z/nr8o5zEeq

reneme: Note to self: look into the concurrency pitfalls of this....

src/lib/utils/esdm/info.txt
```diff
@@ -0,0 +1,13 @@
...
+
+<module_info>
+name -> "ESDM"
+</module_info>
+
+load_on vendor
+
+<libs>
+linux -> esdm_rpc_client
+</libs>
```
reneme, line 13: Do we need this? Couldn't we just `load_on vendor` the actual `esdm_rng` module?
If that's just a naming issues (for `--with-esdm`), I guess we could just rename the module in "rng" accordingly.

reneme (commented):
This seems interesting and the code looks good already. Thanks for the contribution! I've left a few suggestions down below.

Is there a way to test this in CI? I'd hope that we could just install and configure the daemon in `src/scripts/ci/setup_gh_actions.sh` and run a few basic smoke tests against in in the unit tests.

randombit:
So the change already looks pretty good, I think reneme already hit most of the points I would raise on it, but I am going to have to be a downer here and ask

Can you expand more on where and how ESDM is used? It does not seem to be widely distributed, eg per https://repology.org/project/esdm/versions it is only included in AUR and nix (the Debian experimental listing is a name collision with another package). In general our integrations with other systems have been for very widely deployed (eg TPM, PKCS11, OS provided interfaces, or widely deployed libs like zlib or sqlite3) and I'm hesitant to accept something that is so niche.

src/lib/rng/esdm_rng/esdm_rng.h
reneme: Re: concurrency: [cppreference.com](https://en.cppreference.com/w/cpp/memory/shared_ptr) states:

> All member functions (including copy constructor and copy assignment) can be called by multiple threads on different shared_ptr objects without additional synchronization even if these objects are copies and share ownership of the same object. If multiple threads of execution access the same shared_ptr object without synchronization and any of those accesses uses a non-const member function of shared_ptr then a data race will occur [...]

Additionally, [in std::weak_ptr<>::lock](https://en.cppreference.com/w/cpp/memory/weak_ptr/lock) it states:

> Effectively returns `expired() ? shared_ptr<T>() : shared_ptr<T>(*this)`, executed atomically.

Therefore, I am quite sure that [my proposed solution](https://godbolt.org/z/nr8o5zEeq) is correct, under the assumption that the "owned object" itself handles its thread synchronization as necessary (which your ESDM adapter is doing). The manipulation of the global weak ptr and in particular the lazy construction of the `ESDM_Context` object still need to be mutexed though. All shared_ptrs that come out of this factory method are copies of the weak_ptr's state though, so the condition stated in the first quote applies.

thillux:
> So the change already looks pretty good, I think reneme already hit most of the points I would raise on it, but I am going to have to be a downer here and ask
> 
> Can you expand more on where and how ESDM is used? It does not seem to be widely distributed, eg per https://repology.org/project/esdm/versions it is only included in AUR and nix (the Debian experimental listing is a name collision with another package). In general our integrations with other systems have been for very widely deployed (eg TPM, PKCS11, OS provided interfaces, or widely deployed libs like zlib or sqlite3) and I'm hesitant to accept something that is so niche.

You're right with ESDM being niche from a major Linux distribution point of view. I cannot weaken/invalidate this argument.

ESDM is currently mostly used under the public radar I would say. My company uses it to provide randomness in a regulated/certified market, on an installed base of >= 100k Linux-based notebooks and plans to use ESDM in more products in the future.

For such a usage in a regulated market or for governmental customers, it offers the following key benefits from my point of view:

- In combination with ESDM Botan can offer a NIST SP800-90B/C compatible RNG
- ESDM offers more control over configured entropy sources, when needed, without the need for a developer to change Botan
- ESDM can also use Botan for its internal cryptographic primitives, so less certification/code review work necessary for the developer
- ESDM is currently the only way to conform to the upcoming NTG.1 definition of the AIS 20/31 2024 from BSI in Germany
- It is easier to comply with changing governmental requirements in user-space in ESDM, than in kernel space (e.g. change /dev/random)

thillux:
> I'm pretty sure it doesn't work on Mac, for example. No idea if it runs on Windows.

That's correct, it runs on Unix/Linux.

thillux:
I started working on your suggestions and will squash them together again, when finished.

src/lib/utils/esdm/info.txt
thillux: Changed this to with-esdm_rng and dropped the esdm info file.

src/lib/rng/esdm_rng/esdm_rng.h
thillux: I adapted your concept, but had to move weak_ptr and mutex to be class variables in order to use them in the destructor again.

src/lib/rng/esdm_rng/esdm_rng.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+#include "esdm_rng.h"
+#include <esdm/esdm_rpc_client.h>
+
+namespace Botan {
+
+ESDM_RNG::ESDM_RNG() : ESDM_RNG(false) {}
+
+ESDM_RNG::ESDM_RNG(bool prediction_resistance) : m_prediction_resistance(prediction_resistance) {
+   std::lock_guard lg(m_init_lock);
```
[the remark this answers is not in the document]
thillux: Removed with introduction of ESDM_Context

thillux:
> This seems interesting and the code looks good already. Thanks for the contribution! I've left a few suggestions down below.
> 
> Is there a way to test this in CI? I'd hope that we could just install and configure the daemon in `src/scripts/ci/setup_gh_actions.sh` and run a few basic smoke tests against it in the unit tests.

How can I test this by myself? I think I'll need some iterations to get this running.

reneme:
> How can I test this by myself? I think I'll need some iterations to get this running.

The repository policy doesn't let first-time contributors run the CI pipeline. You could iterate in a branch on your private fork, though. For a full end-to-end test, unfortunately the process is commit->push->fail->fix->commit :cry:

You can, however, run the underlying CI script locally (that assumes that `setup_gh_actions.sh` -- or something else -- has already prepared the environment). Simply run: `src/scripts/ci_build.py shared` to run most things the CI would for the 'shared' (library) target.

thillux:
> > That's correct, it runs on Unix/Linux.
> 
> Does Botan aim to only run on Linux? And at that, on one (possibly major) distribution of Linux?

The ESDM_RNG class is added as optional dependency. On Windows and/or MacOS it just can't be used. This PR does not limit Botan to only Linux.

src/lib/rng/esdm_rng/esdm_rng.cpp
```diff
@@ -0,0 +1,81 @@
...
+      }
+
+      ~ESDM_Context() {
+         std::scoped_lock lk(g_mutex);
+
+         if(g_instance.use_count() == 1) {
+            esdm_rpcc_fini_unpriv_service();
+            g_instance.reset();
+         }
+      }
```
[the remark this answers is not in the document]
reneme: Sorry for the confusion.
Here's another suggestion: https://godbolt.org/z/cnd9r91so

src/lib/rng/esdm_rng/esdm_rng.h
reneme: Thanks! Though, I think, the whole `weak_ptr` idea was a blunder, unfortunately. See here: https://github.com/randombit/botan/pull/4309#discussion_r1719738277

src/lib/rng/esdm_rng/esdm_rng.cpp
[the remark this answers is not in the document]
thillux: Thanks for the suggestions. I'll have to think about this too. I'm not really happy with the new without free, which probably will confuse some sort of memory leak analyzer.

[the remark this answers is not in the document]
thillux: I think with my small change I'm happy with ESDM_Context now.

thillux:
I'll probably try to add the test this weekend or start of next week.

src/lib/rng/esdm_rng/esdm_rng.cpp
```diff
@@ -0,0 +1,83 @@
...
+class ESDM_Context {
+   private:
+      constinit inline static std::mutex g_mutex;
+
+   public:
+      [[nodiscard]] static std::shared_ptr<void> instance() {
+         std::scoped_lock lk(g_mutex);
+         static std::shared_ptr<ESDM_Context> g_instance{new ESDM_Context()};
+         return g_instance->acquire();
+      }
```
reneme, line 24: The `g_mutex` isn't necessary here. As of C++11 it is guaranteed that lazy initialization of local static variables is threadsafe: https://en.cppreference.com/w/cpp/language/storage_duration#Static_block_variables

> If multiple threads attempt to initialize the same static local variable concurrently, the initialization occurs exactly once (similar behavior can be obtained for arbitrary functions with [std::call_once](https://en.cppreference.com/w/cpp/thread/call_once)).

Also: its just fine to not use a smart pointer here but just directly initialize the `ESDM_Context`.

```suggestion
      [[nodiscard]] static std::shared_ptr<void> instance() {
         static ESDM_Context g_instance{};
         return g_instance->acquire();
      }
```

src/lib/rng/esdm_rng/esdm_rng.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+   private:
```
reneme, line 26: I forgot that in my godbolt example as well: Let's define the default constructor as private to make sure no-one creates another instance of this class:

```suggestion
   private:
      ESDM_Context() = default;
   
```

src/lib/rng/esdm_rng/esdm_rng.cpp
```diff
@@ -0,0 +1,83 @@
...
+*/
+
+#include <botan/esdm_rng.h>
+
+#include <esdm/esdm_rpc_client.h>
+
+namespace Botan {
+
+namespace {
+class ESDM_Context {
```
reneme, line 15:
```suggestion
/**
* This helper makes sure that the ESDM service is initialized and 
* finalized as needed in a threadsafe fashion. Finalization happens
* as soon as all instances of ESDM_RNG are destructed. This may
* happen multiple times in the lifetime of the process.
*/
class ESDM_Context {
```

... future-us will be grateful. 😄

reneme (commented):
One more suggestion for our little initialization-concurrency rabbit hole 🕳️, otherwise looks good to me now. Looking forward to a test.

Regarding the utility of this: I tend to agree that this is rather niche. Nevertheless, this has sparked some interest in our company as well, for similar reasons as pointed out by @thillux. FWIW.

thillux:
@reneme Smoke tests added and PR squashed again.

thillux:
I'm maybe to stupid to get esdm-server started for the coverage and sanitizer targets. I'll pushed a fix that hopefully works.

thillux:
All tests green in my repo now.

thillux:
I started rebasing, will push an updated version for 3.7 soon.

thillux:
@reneme @randombit CI was green on my testing branch yesterday.

thillux:
@reneme @randombit Is there something left in this PR which I have to adapt to and was not aware of?

src/scripts/test_python.py
```diff
@@ -179,6 +179,28 @@ def test_rng(self):
 
         user_rng.add_entropy('seed material...')
 
+        # execute ESDM tests only on Linux
+        if platform.system() != "Linux":
+            return
```
reneme, line 184:
```suggestion
        if platform.system() != "Linux":
            self.skipTest("This test requires Linux")
```

reneme (approved):
Thank you for your patience and effort that went into that! I left a few minor nits but generally this looks really good to me and also works fine on my local machine. I'll prepare a patch integrating the CI dependency as mentioned in one of the comments.

src/scripts/ci/setup_gh_actions.sh
```diff
@@ -185,6 +185,20 @@ if type -p "apt-get"; then
...
 
     fi
+
+    # ESDM for shared, coverage and sanitizer target
+    if [ "$TARGET" = "shared" ] || [ "$TARGET" = "coverage" ] || [ "$TARGET" = "sanitizer" ]; then
+        sudo apt-get -qq install libprotobuf-c-dev meson
+
+        # install ESDM 1.2.0
+        wget -O esdm.tar.gz https://github.com/smuellerDD/esdm/archive/refs/tags/v1.2.0.tar.gz
+        tar xvfz esdm.tar.gz
```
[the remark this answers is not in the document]
reneme: As promised: https://github.com/reneme/botan/pull/7/commits/a4f20d640cef54546852ce9df69032fa42678bc9
Could you please cherry-pick that into your branch? Please feel free to also squash it with your commit.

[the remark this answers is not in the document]
thillux: This was not by accident. I allowed maintainers of the upstream repo to edit my code, when I created the PR.

thillux:
I applied your suggestions and squashed them together with my last commit as suggested. Thanks for your review @reneme!

src/lib/rng/esdm_rng/esdm_rng.cpp
```diff
@@ -0,0 +1,97 @@
...
+   if(!in.empty()) {
+      ret = 0;
+
+      /*
+      * take additional input, but do not account entropy for it,
+      * as this information is not included in the API
+      */
+      esdm_invoke(esdm_rpcc_write_data(in.data(), in.size()));
+      if(ret != 0) {
+         throw Botan::System_Error("Writing additional input to ESDM failed");
```
randombit, line 74: This is pretty confusing on first read - is `esdm_invoke` a macro that sets `ret`? 😬 If so please add a block comment here calling out this.

src/scripts/config_for_oss_fuzz.py
```diff
@@ -29,7 +29,7 @@
     "--build-fuzzers=libfuzzer",
     "--build-targets=static",
     "--without-os-features=getrandom,getentropy",
-    "--disable-modules=system_rng,processor_rng",
+    "--disable-modules=system_rng,processor_rng,esdm_rng",
```
randombit, line 32: This shouldn’t be necessary - this is only built if explicitly requested. (Right??)

src/scripts/test_python.py
```diff
@@ -179,6 +179,27 @@ def test_rng(self):
 
         user_rng.add_entropy('seed material...')
 
+        if platform.system() != "Linux":
```
randombit, line 182: Don’t do this. This breaks anyone (such as distributions or end users) which want to run the Python tests but are not using ESDM. Instead try to call ESDM. If it succeeds, great. If not, verify that we throw something indicating not implemented / not available.

src/lib/rng/esdm_rng/esdm_rng.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+      void release() {
+         std::scoped_lock lk(m_mutex);
+         if(m_refs-- == 1) {
+            esdm_rpcc_fini_unpriv_service();
+         }
+      }
+
+   private:
+      std::mutex m_mutex;
```
randombit, line 49: Couple of things here

1) If using `std::mutex` then include `<mutex>` in the file
2) This then requires threads, which is ok BUT
2a) If we require threads here then add `os_features` block listing threads
2b) If it’s ok to use this on a machine without threads then use the wrappers in `utils/mutex.h`

I think the context ESDM is used in it probably doesn’t make sense to support systems without threads. This is typically just for people doing very niche baremetal/RTOS/etc kind of things so probably the correct fix is 2a.

randombit (changes requested):
@thillux Generally looks good and we can probably merge for 3.7.0, some comments to address first though.

src/scripts/test_python.py
thillux: Done, I moved this to a own unit test function, which checks for "not implemented" and gets skipped accordingly.

src/lib/rng/esdm_rng/esdm_rng.cpp
thillux: Done. I second to use 1 & 2a.

src/scripts/config_for_oss_fuzz.py
thillux: Done. I removed this, as you are right here.

src/lib/rng/esdm_rng/esdm_rng.cpp
thillux: The comment mentioning this, was already at the initial declaration of `ret` in line 64. I added a short note above every check against `ret` nevertheless, as this behavior is probably not expected.

thillux:
@randombit Thanks for your feedback! If you can live with the current state, I'll squash the WIP commit. I just left it for now, in order to make it easier to look at the changes related to your suggestions.

src/lib/rng/esdm_rng/esdm_rng.cpp
randombit: Sorry I missed this

src/scripts/test_python.py
randombit: Perfect thanks

randombit (approved):
Thanks @thillux LGTM pending a final squash

thillux:
> Thanks @thillux LGTM pending a final squash

Thanks for your review, squash is now pushed.

reneme (approved):

PR #4312, opened 2024-08-16
Provides auxiliary vector support for OpenBSD

brad0:
OpenBSD 7.6 provides elf_aux_info().


randombit:
I don't think we'd want to merge this until after 7.6 is released (October, IIUC?). Since OpenBSD only (AIUI) supports the most recent release, we're fine with merging changes that require the latest OpenBSD (per our general policy of only supporting operating systems that are supported by the original vendor) but since 7.6 isn't even out we'd wait until at least then.

Can you point me to OpenBSD's hwcap.h (or equivalent) headers? Particularly for armv7, aarch64, and ppc64. The main point for this is to perform CPU specific probing so if the feature bits don't match Linux it's not so helpful.

brad0:
https://github.com/openbsd/src/blob/master/sys/arch/arm/include/elf.h
https://github.com/openbsd/src/blob/master/sys/arch/arm64/include/elf.h
https://github.com/openbsd/src/blob/master/sys/arch/powerpc64/include/elf.h

randombit:
OK checked over all of the feature bits and everything (at least that we're currently using) seems to match. However I noticed a hangup - it appears `AT_HWCAP` is defined as 25 while on Linux it's 16. `AT_HWCAP2` is 26 on both systems. Currently this value is hardcoded (mostly IIRC because various glibc headers on some (old?) systems are missing the definitions). I'll make a change to fix this.

The comment in https://github.com/openbsd/src/commit/ef873df06dac50249b2dd380dc6100eee3b0d23d is that these values were taken from FreeBSD so I wonder what the value of `AT_HWCAP` is there. We may have had broken feature detection on BSD for a while and not noticed.

randombit:
On further consideration I don't think there's a need to wait to merge this - our next release is also scheduled for October, so in the worst case there is a week or two between Botan 3.6.0 and OpenBSD 7.6, where it is kind of mildly broken, but end users can always disable the feature at build time for OpenBSD 7.5 or prior.

randombit (approved):

brad0:
> The comment in [openbsd/src@ef873df](https://github.com/openbsd/src/commit/ef873df06dac50249b2dd380dc6100eee3b0d23d) is that these values were taken from FreeBSD so I wonder what the value of `AT_HWCAP` is there. We may have had broken feature detection on BSD for a while and not noticed.

The value is the same from FreeBSD.

https://github.com/freebsd/freebsd-src/blob/35399f68c8e8a9b3bb4905057ca59766751766f3/sys/sys/elf_common.h#L981

brad0:
> On further consideration I don't think there's a need to wait to merge this - our next release is also scheduled for October, so in the worst case there is a week or two between Botan 3.6.0 and OpenBSD 7.6, where it is kind of mildly broken, but end users can always disable the feature at build time for OpenBSD 7.5 or prior.

Ports / packages are only supported and updated for -current and this has been in -current for awhile. So that is fine. Thanks.

randombit:
> Ports / packages are only supported and updated for -current and this has been in -current for awhile.

OK perfect, good to merge as soon as you remove draft status then

> The value is the same from FreeBSD.

Thanks for tracking that down. This part should be sorted out for both systems with #4315.

PR #4314, opened 2024-08-16
Add basic asymmetric roundtrip tests

reneme:
### Pull Request Dependencies

* #4318 

### Description

This piggy-backs on the `PK_Key_Generation_Test` to perform basic roundtrip tests for Sign/Verify, Encrypt/Decrypt, Encaps/Decaps, and Key Agreement, depending on the capabilities of the keys.

The goal is explicitly not to be an exhaustive test but rather be a centralized sanity-check for the PK_*** operators and their meta-data methods. We added it mostly to serve as a place for testing whether the new `set_associated_data()` setting survives multiple signature/verify operations; and figured that it doesn't hurt to roundtrip-test all the other flavors as well.

This could probably also replace some of the roundtrip-tests in the new PQ implementations. We'll look into cleaning up their test harnesses after the FIPSes are implemented and merged.

### Caveat

We'll have to have a look at whether that adds too much time to the test harness. Edit: it almost doubles the runtime of the entire unit test harness on my laptop. That's probably not okay! Hence, in test runs that don't set `--run-long-tests` we only run those roundtrips for the first parameter set of each algorithm.

### DIFF
--- a/src/lib/pubkey/pk_ops.cpp
+++ b/src/lib/pubkey/pk_ops.cpp
@@ -82,6 +82,16 @@ secure_vector<uint8_t> PK_Ops::Key_Agreement_with_KDF::agree(size_t key_len,
    return z;
 }
 
+void PK_Ops::Signature::set_associated_data(std::span<const uint8_t> associated_data) {
+   BOTAN_UNUSED(associated_data);
+   throw Not_Implemented("This signature scheme does not support labels for signing");
+}
+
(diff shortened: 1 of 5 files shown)


reneme:
This will need to be revisited once #4318 has converged and landed. I converted it to a draft PR.

randombit:
Still worth having, waiting for #5489

PR #4315, opened 2024-08-16
Fix aarch64/armv7/ppc64 feature detection for systems with AT_HWCAP != 16

randombit:
This value was hard coded but it appears - while all known systems use the same bit flags for HWCAP/HWCAP2, they do not necessarily use the same value for AT_HWCAP itself. This was discovered during review of #4312

Previously this value was hardcoded, because some older glibc headers are missing the definition. Instead expose it via a pair of new functions in OS, which return the actual definition, if it's available, or otherwise default to the Linux/Android value.


brad0:
and just FYI this issue affects Botan 2.

Issue #4020, opened 2024-04-14
Botan doesn't support large OIDs


jdoubleu:
Botan cannot store OIDs with large _subidentifiers_. The standard (ITU-T X.660, ITU-T X.680) does not limit the number range on any subidentifier other than the root arc and some secondary identifiers.

For example, `2.299.62133450347103072435436460230495464564354` is a valid OID, but cannot be stored by Botan. That is, because Botan [stores the subidentifiers as `uint32_t`](https://github.com/randombit/botan/blob/master/src/lib/asn1/asn1_obj.h#L315).

While the above example may be unrealistic, there's a valid use-case for 128bit numbers: Arc `2.25.` allows for UUIDs to be embedded inside its OID (see [ITU-T X.667](https://itu.int/ITU-T/X.667) and http://oid-info.com/get/2.25).

See also: https://misc.daniel-marschall.de/asn.1/oid_facts.html#chap4

randombit:
TBH I had not heard of this `2.25` arc before for UUIDs. I'd be pretty inclined to go with "that's interesting but I doubt anyone uses it", especially as RFC 5280 only requires implementations support OID elements up to `2^28`, but I see for example Go has recently introduced a new OID type to handle such UUIDs in certificate policies. :/

Issue #3275, opened 2023-02-13
Type-based algorithm instantation 


falko-strenzke:
I would like to open up the discussion about a type-based interface for the algorithm instantiation as an enhancement to the existing string-based interface. I brought up the topic in the context of the integration of KMAC into Botan in issue #3262. 

My concrete proposal is a type based interface making use of C++ operator overloading to allow a concise and yet compile-time-checked specification of an algorithm.

The reason I see for such an enhancement are the shortcomings of the exising string-based interface. The string-based interface means instantiating for instance SHAKE128 with output size of 32 bytes as `Hash::create("SHAKE-128(256)")`
There are numerous drawbacks in the string-based interface ovious from this example:
* The exact spelling of the algorithm has to be known (e.g. with or without "-"); an error in the spelling will surface only at runtime.
* The meaning of the value in brackets as the output length has to be known, i.e. is not reflected in the caller's source code.
* It may not always be clear whether numbers refer to bit or byte lengths. In any case, the caller code has to take care of that conversion itself.
* If the output length is chosen at runtime, the calling code will have to encode a numeric value into a string, which should clearly be unnecessary.

One could attempt to mitigate some of the issues by at least providing defined string constants. But I think using a type-based system is the much more profound way. Regarding the KMAC interface in the issue mentioned above I propse for instance something in the flavour of

```
auto alg_spec = HMAC256 / Output_Length(Bits(128));
auto hash = Hash::create(alg_spec);
```
(assuming a specific namespace where those types live of course).

The advantages I see are:
* Good readability (and this is key in my opinion) of what is instantiated just as in the case of the string-based interface. This is the reason for proposing operator overloading. Lengthy function names should be avoided here, otherwise the poor readability would possibly outweigh the gains in my opinion.
* Compile time check. In the above example, the type system would allow instantiating HMAC without truncation as well as with it. But for other algorithms the type system would ensure that RSA can only be instantiated with a specific encoding, leading to **compile-time** errors when the encoding is missing.
* Bits(...) or Bytes(...) explicit type instantiation would ensure that there is never an ambiguity with respect to that and free the caller from any such conversions.

The idea here is to keep the string-based interface, but to implement it in terms of the the new type-based interface. This would work generically to a certain degree. The reasons to keep the string-based interface permanently are in view:
* ensure backwards compatibility
* the advantage it has for any application that simply wants to forward an external algorithm specification to Botan. An example where this is useful are for instance the Botan tests.

[...]

PR #4318, opened 2024-08-17
Add PK_Signature_Options

randombit:

~This is still missing some bits, mostly around each scheme has to validate that the requests are sensible. For example right now for most algorithm if you set a context, we just ignore it instead of rejecting it.~ Should be covered now.

There are also some future extension points to consider; for example for ECDSA/DSA we currently gate deterministic signatures just on the compile time existence of RFC6979. In the future this could be changed so that RFC6979 is only used if both available at compile time and requested.

Future work will add analogous options for encryption/decryption, and for KEM.

XMSS previously completely ignored it's parameters string. Eg a XMSS-SHA-512 key you could set params to "SHA-256" or "WHATEVER" or anything and it just worked. Now the parameters string, if set, must be exactly the hash function used by the key.

Todos
- [x] Also update verification
- [x] Fully check arguments (reject context if not supported, etc)
- [x] Move validation logic to an internal header to avoid exposing more than we need to (even if nominally `_` prefixed and "private")
- [x] RSA logic works for compat mode but is confusing using the new API. We'd want to support eg `PK_Signature_Options("SHA-256").with_padding("PSSR")` and right now I think that doesn't work.
- [x] Maybe add `PK_Signature_Options::with_hash` so the user can be extra explicit??
- [x] Finish gutting `EMSA::create`
- [x] Figure out what to do for DS2/DS3 implicit vs explicit
- [x] Update cli and examples
- [x] Compat shims for someone calling `create_signature_op`
- [ ] Make the RNG optional (currently it is required, even when creating a deterministic signature)
- [ ] Update documentation
- [ ] Modify tests (?) [maybe better to leave them as is to ensure coverage on `_parse`, IDK]
- [ ] Support providing a hash function object (more details in [this collapsed comment](https://github.com/randombit/botan/pull/4318#issuecomment-2297682058))
- [ ] Comprehensive tests for all algorithms that we reject parameters that are unexpected
- [ ] Figure out a story for FFI. In the short term (until Botan4) the old constructors do everything, but we might need to expose something like `PK_Signature_Options` to FFI in the long run.


src/lib/pubkey/pk_options.h
```diff
@@ -0,0 +1,143 @@
...
+      bool using_context() const { return context().has_value(); }
+
+      bool using_prehash() const { return m_use_prehash; }
+
+      bool using_padding() const { return padding().has_value(); }
+
+      bool using_provider() const { return provider().has_value() && provider().value() != "base"; }
+
+      // TODO
+      bool _valid_for_rsa() const;
```
randombit, line 113: These are just examining the data and don't need access to the internals; probably should be moved to an internal header.

reneme:
Nice! @FAlbertDev and I had discussed whether the "builder pattern" would be a good fit here. And we're convinced it is!

One alternative I think is worth considering besides Builder would be C++20's [aggregate initialization with designated initializers](https://en.cppreference.com/w/cpp/language/aggregate_initialization). As far as I'm aware, that's the closest we could get to Python's kwargs (for better and worse). From a user's perspective, it would allow instantiating a `PK_Signer` roughly so:

```C++
        auto key = std::make_unique<ML_DSA_PrivateKey>();

        Signer signer(*key, {
            .hash = "SHA-256",
            .context = as_byteview("API discussion"),
            .deterministic = false,
        });

        signer.sign_message(...);
```

Essentially, we would define an aggregate struct (no constructors, all members are public) that is taken by the c'tor of `Signer`. This struct is then aggregate-initialized in place with all configurations as needed. Full in-place construction isn't a must, it would also work like this:

```C++
Options opts{.hash = "SHA-256", .context = as_byteview("API discussion")};
if (/* some application-specific condition */) {
  opts.deterministic = true;
}
Signer signer(*key, opts);
```

There are a few caveats:
* The fields in the aggregate-initialization must be in the same order as declared in the struct. Essentially, this order becomes part of our semver guarantee (I guess), and we should never change it.
* Inter-dependent invariants between fields aren't checkable. Obviously, as the user may set whatever they want.

reneme:
Also, I'd love to throw in another requirement while we're adding flexibility to these classes:

Requirement: **Allow passing pre-constructed primitives (i.e. `Padding` and `HashFunction`)**

I acknowledge that most users will probably just pass the stringified algorithm spec, like "OAEP", or "SHA-256", or "PKCS1(SHA-1)". But, for special use cases, it would be great to circumvent the rigid algorithm factories. Either to drop in custom implementations (e.g. via some hardware) or to have more control over the configuration of the primitive.

As [discussed yesterday](https://github.com/randombit/botan/pull/4313#discussion_r1720613561), I ran into this when required to use custom OAEP labels. My [current workaround is to duplicate some parts of the RSA signing mechanism](https://github.com/Rohde-Schwarz/botan/blob/739435e86f3f612f02cd9a291330f307bed4c4ad/src/lib/prov/tpm2/tpm2_crypto_backend/tpm2_crypto_backend.cpp#L477-L513). But that's ugly, error prone and hard to maintain.

Again, from a user perspective, I'd love to be able to do this:

```C++
        auto key = std::make_unique<ML_DSA_PrivateKey>();

        Signer signer(*key, {
            .padding = std::unique_ptr<OAEP>(HashFunction::create_or_throw("SHA-256"), "some label"),
            .context = as_byteview("API discussion"),
            .deterministic = false,
        });

        signer.sign_message(...);
```

This could be achieved by making the `.padding` field a `std::variant<std::string, std::unique_ptr<EME>`. If the string is provided, we'd call `EME::create()` with it. Otherwise the prepared unique ptr could be used as is. The logic that implements this could be stowed away in the new `PK_Options` fairly easily.

reneme:
I've built [a comprehensive spike](https://godbolt.org/z/fv9Medo95) of the aggregate-initialization approach and also the proposed "pre-created primitives". There, I added a notion of an `OptionsReader` that somewhat alleviates the mentioned short-coming of having no central place to validate invariants in the aggregate-initializable struct.

The `OptionsReader` also contains logic to lazily instantiate crypto primitives from the algo-spec string as necessary. Additionally, it allows for somewhat centralized error handling by tracking which configs were pulled from the concrete scheme and failing when required configs weren't set or other configs weren't used.

Note that the `OptionsReader` is instantiated in the generic `PK_Signer` class and passed down to the concrete scheme implementations. It therefore acts as a gate keeper to prevent concrete schemes from circumventing the safeguards (and potential invariant checks) of `OptionsReader`. They simply cannot directly access the all-public `Options` aggregate struct. For the record: I'm not sure it is sustainable to do these kinds of invariant and usage checks so centrally and without knowledge of the concrete scheme. It might very well be that this doesn't stand up to the nitty-gritty details in the end.

_Each of these ideas should be easily adaptable to your "builder pattern" approach. I just explored them based on the aggregate-initialization to put that idea through its paces. Feel free to just adopt what you like or leave it all out._

... This has been quite a wall-of-text. Thanks for reading until here! I hope I was able to provide some additional inspiration to build a new and modern API on top of our asymmetric algorithms. Perhaps it makes sense to pick up this discussion in videochat?

reneme (commented):
Just a few minor comments from reading over the patch. I love the direction this is taking!

src/lib/pubkey/pk_keys.h
```diff
@@ -235,11 +253,11 @@ class BOTAN_PUBLIC_API(2, 0) Public_Key : public virtual Asymmetric_Key {
...
       * Return a verification operation for this key/params or throw
-      * @param params additional parameters
-      * @param provider the provider to use
+      *
+      * @param options which specify parameters of the signature beyond those
+      * implicit to the public key itself
       */
-      virtual std::unique_ptr<PK_Ops::Verification> create_verification_op(std::string_view params,
-                                                                           std::string_view provider) const;
+      virtual std::unique_ptr<PK_Ops::Verification> _create_verification_op(const PK_Signature_Options& options) const;
```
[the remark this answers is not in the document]
randombit: I always considered these functions internal and - had the convention existed earlier - they would have been `_` prefixed. That said, they *were not in fact _ prefixed*, and I guess a compatability shim here isn’t that onerous.

src/lib/pubkey/pk_options.h
```diff
@@ -0,0 +1,137 @@
...
+      // Returns padding plus hash formatted for RSA
+      std::string _padding_with_hash() const;
+
+      /// This is a compatability interface that parses padding in the context
+      /// of the key type, following internal logic used previously.
+      ///
+      /// This is an internal library function and should not be called by
+      /// applications. It will be removed in Botan4.
+      ///
+      static PK_Signature_Options _parse(const Public_Key& key, std::string_view padding, Signature_Format format);
```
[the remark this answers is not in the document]
randombit: There actually is no need for `_parse` to be a member function, I think it just accesses the public interface. [Since if _parse needed to do something that was not possible with the public interface, it would mean there is some configuration of hashes/etc that can be used in the library today that couldn’t be requested using the public API.] I’ll just move this to the internal `impl` header.

src/lib/pubkey/pk_options.h
```diff
@@ -0,0 +1,137 @@
...
+namespace Botan {
+
+/**
+* Signature generation/verification options
+*/
+class BOTAN_PUBLIC_API(3, 6) PK_Signature_Options {
+   public:
+      PK_Signature_Options(std::string_view hash_fn) : m_hash_fn(hash_fn) {}
+
+      PK_Signature_Options() : PK_Signature_Options("") {}
```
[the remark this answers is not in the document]
randombit: Yeah I was originally thinking “hash function is mandatory” (iterating on just ECDSA/RSA) and then hit XMSS/Dilithium/etc and things got tricky. Probably should be an optional.

src/lib/pubkey/pk_keys.h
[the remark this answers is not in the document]
reneme: Not 100% sure, but another alternative perhaps: couldn't we move those into a `protected:` region and declare the generic `PK_Signer`, ...classes as friends? That might need a sort-of "landing pad" non-virtual method in the base class and protected pure virtual methods as customization points to delegate to (think `update` vs. `add_data` in `Buffered_Computation`)

Certainly a lot of boilerplate just to omit underscored public methods. Just thinking out loud...

I'm in team "friends before private-ish public methods", if you couldn't tell. 😅

src/lib/pubkey/pk_options.h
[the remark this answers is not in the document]
randombit: No longer public/"not public"

randombit:
So on the aggregate initialize proposal, I think I'd pass on that. It does offer some marginal benefit in readability vs `PK_Signature_Options` proposal, but IMO the additional costs aren't worth it. We'll inevitably find we need to eg remove a field or change its type, and then be sad. The number of things over the years where something could not be easily fixed because we exposed more internals than necessary is large (for example #4020 is not easily fixable in Botan3, because we expose the OID components directly as uint32s)

[Actually, on that note I think I'm going to change most of the `PK_Signature_Operations` getters to not return a const reference...]

It does make sense to be able to provide the hash function as an object. For hardware, or for someone who wants to use some unusual/new hash, or whatever. I haven't done that here yet, but it seems very doable. I think instead of a variant I'd suggest

`with_hash(std::string_view hash)` -> calls `HashFunction::create_or_throw`
`with_hash(std::unique_ptr<HashFunction> hash)` -> moves directly into the struct

Ie downstream consumers of the option can expect that (if a hash function is being used) the object already exists, without having to fuss with variants. (Since they will inevitably always do the exact same thing - if it's an object, use it, otherwise create based on the string.)

For providing a padding object, note that since 3.0 these types have been completely hidden/internal. I would very much like it to keep them that way unless there is some truly critical blocker, especially because if you are crazy enough to yolo your own padding scheme it doesn't seem *that* bad to just use "Raw" padding and provide the padded bytes directly. It seems like your main driver here is being (understandably!) frustrated about OAEP labels situation. May I offer

```
PK_Encryptor enc(*key, PK_Encryption_Options().with_padding("OAEP").with_hash("SHA-256").with_context(the_oaep_label_bytes).with_salt_len(32));
```

doc/deprecated.rst
```diff
@@ -176,6 +176,13 @@ release, or where a backwards incompatible change is expected.
 - All support for loading, generating or using RSA keys with a public
   exponent larger than 2**64-1
 
+- Currently some public key padding mechanisms can be used with several
+  different names. This is deprecated.
+  "EMSA_PKCS1", "EMSA-PKCS1-v1_5", "EMSA3": Use "PKCS1v15"
+  "PSSR_Raw": Use "PSS_Raw"
+  "PSSR", "EMSA-PSS", "PSS-MGF1", "EMSA4": Use "PSS"
+  "EMSA_X931", "EMSA2": Use "X9.31"
```
reneme, line 184: ❤️

reneme:
> We'll inevitably find we need to eg remove a field or change its type, and then be sad.

Yep, I agree. The aggregate approach does lack flexibility because of its simplicity. Unfortunately the `OptionsReader` cannot fully compensate that. Let's go with the builder.

> It does make sense to be able to provide the hash function as an object. [...] I haven't done that here yet, but it seems very doable. I think instead of a variant I'd suggest [...]

By all means! The variant was necessary for the aggregate (because the `OptionsReader` had to decide on construction vs. forwarding after the effect). In the builder-pattern what you propose is the more sensible option!

> For providing a padding object, note that since 3.0 these types have been completely hidden/internal.

Yeah, that makes the pre-created padding object a non-starter, I didn't see they are internal -- which makes sense. Regarding the alternative suggestion: I'm not fully convinced about that yet. Don't we start mixing concerns in the builder then? Or do you mean to enable RSA encryption to consume an abstract "context" via the OAEP label that other schemes might support out of the box? This does sound reasonable to me at first sight.

Alternatively one could think of a dedicated builder for the paddings, and pass that onto the signature scheme's builder. Just putting that idea here for the record. Admittedly, that feels like overkill at first sight: "If you have a hammer, everything looks like a nail." 😏

```C++
PK_Encryptor enc(*key,
                 PK_Encryption_Options()
                    .with_padding(
                        PK_Padding_Options()
                            .with_hash("SHA-256")
                            .with_label("secret\0")
                            .with_mgf("MGF1"));
```

reneme:
Two more general points up for discussion:

1. We should probably reconsider Falko's suggestion of [Type-based algorithm instantiation](https://github.com/randombit/botan/issues/3275) while introducing this builder concept. Certainly not to implement it all at once, but in my opinion the string-based algorithm specifiers are an annoyance in our public API (see, eg. OAEP).

2. If we introduce a builder, we could (and perhaps should) actually take it all the way and let it spit out the actual `PK_Signer` instance:
   ```C++
   auto signer = PK_Signer_Builder()
     .with_key(my_rsa_private_key)
     .with_padding("PSS")
     .with_salt(...)
     .with_hash(...)
     .build();
   
   signer.sign_message(...)
   ```

doc/deprecated.rst
reneme: Probably its worth to also adapt the padding documentation in this regard: https://botan.randombit.net/handbook/api_ref/pubkey.html#available-encryption-padding-schemes. It lists those "alternative names" without giving guidance on what is the "preferred" name.

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
```diff
@@ -182,16 +182,41 @@ class Dilithium_PrivateKeyInternal {
...
+
+   if(options.using_salt_size()) {
+      BOTAN_ARG_CHECK(!options.using_deterministic_signature(),
+                      "Dilithium cannot support a salt while also being deterministic");
+      BOTAN_ARG_CHECK(options.salt_size().value() == 32, "Dilithium can only be used with a 32 byte salt");
+   }
+
+   // This is available in ML-DSA and might be supported in the future
+   BOTAN_ARG_CHECK(!options.using_prehash(), "Dilithium does not support prehashing");
+}
```
reneme, line 200: Not actually specific to dilithium: I find those explicit parameter checks are a substantial maintenance burden. We'd have to update essentially all algorithms when introducing a new option. And failing to do so, might break users' security assumptions when designing their protocols.

For instance: Let's say we introduce the "context" notion for asym encryptino, _but mess up the sanity check in RSA (or EME-PKCS1)_:

```C++
PK_Encryptor enc(rsa_public_key,
                 PK_Encryptor_Options()
                   .with_padding("PKCS1") // in contrast to OAEP, this won't actually consume the context
                   .with_context("The security of my protocol relies on this string "
                                 "being bound to the ciphertext"));
auto ct = enc.encrypt("my secret message");
```

... in [my aggregate-based example](https://github.com/randombit/botan/pull/4318#issuecomment-2295240036) I had proposed to let algorithms "take" the options they need, rather then "get" them. Read: reset the option after it being read by the consumer.

As a result, the `PK_Options` object can verify that all (or at least all "critical") options have been consumed. Note that we can't do that in `~PK_Options()` as this shouldn't throw. But the constructor of `PK_Signer` could explicitly invoke the verification mechanism after calling `_create_sig_op` on the private key.

```C++
PK_Signer::PK_Signer(Private_Key& key, PK_Options& opts) :
  m_op(key._create_sig_op(opts)) {
  opts.validate_options(key.algo_name()); // -> throws, e.g. "Dilithium did not consume critical option 'context'"
}
```

... or, [if we go full-on builder-pattern](https://github.com/randombit/botan/pull/4318#issuecomment-2297990007):

```C++
PK_Signer PK_Options::build() {
  auto key = this->take_key();
  PK_Signer signer(key, *this);
  this->validate_options(key.algo_name());
  return signer;
}
```

src/lib/pubkey/pk_options.h
```diff
@@ -0,0 +1,197 @@
...
+      ///
+      /// This can be further parameterized by calling with_xxx functions
+      PK_Signature_Options() = default;
+
+      PK_Signature_Options(PK_Signature_Options&& other) = default;
+      PK_Signature_Options& operator=(PK_Signature_Options&& other) = default;
+
+      PK_Signature_Options(const PK_Signature_Options&) = delete;
+      PK_Signature_Options& operator=(const PK_Signature_Options& other) = delete;
+      ~PK_Signature_Options();
```
reneme, line 41: The moment we act on [the "pre-constructed hash object" notion](https://github.com/randombit/botan/pull/4318#issuecomment-2297682058), we will have a `std::unique_ptr<>` member in this class and the rule-of-zero will apply: you could remove all of them.

src/lib/pubkey/pk_options.h
```diff
@@ -0,0 +1,197 @@
...
+
+      /// Format this PK_Signature_Options as a string
+      ///
+      /// This is primarily intended for debugging and error messages;
+      /// the format is not fixed
+      std::string to_string() const;
```
reneme, line 47: \**chef's kiss*\*

src/lib/pubkey/pk_options.h
```diff
@@ -0,0 +1,197 @@
...
+
+      /// Create a PK_Signature_Options specifying the hash to use
+      ///
+      /// Most but not all signture schemes require specifying the hash
```
reneme, line 51:
```suggestion
      /// Most but not all signature schemes require specifying the hash
```

src/lib/pubkey/pk_options.h
```diff
@@ -0,0 +1,197 @@
...
+      ///
+      /// This can be further parameterized by calling with_xxx functions
+      PK_Signature_Options(std::string_view hash_fn) : m_hash_fn(hash_fn) {}
+
+      /// Specify the hash function to use for signing/verification
+      ///
+      /// Most, but not all, schemes require specifying a hash function.
+      PK_Signature_Options with_hash(std::string_view hash) &&;
```
reneme, line 59: I'm a bit surprised by the constraints on the builder methods. I would have expected:

```C++
PK_Signature_Options& with_hash(std::string_view hash) {
  set_or_throw(m_hash, hash);
  return *this;
}
```

Some users would probably want to express things like this:

```C++
auto some_key = get_private_key();

auto opts = PK_Signature_Options("SHA-256");
if(key->algo_name() == "RSA) {
   opts.with_padding("PSS");
}
if(key->algo_name() == "ML-DSA") {
  opts.with_context("some context");
}

PK_Signer signer(some_key, opts);
```

I think people will grumble and then go for `std::move(opts).with_context(...)` as a workaround. See `pk_options_impl.cpp` 😜.

reneme (commented):
Some more observations.

reneme:
Here's another spike that sketches out most of the suggestions above:

* let the builder create the `PK_Signer` via [a final `create()`](https://github.com/randombit/botan/pull/4318#issuecomment-2297990007)
* allow [lvalues of the Options](https://github.com/randombit/botan/pull/4318#discussion_r1722861968)
* boolean options allow for [an scheme-specific default](https://github.com/randombit/botan/pull/4318#discussion_r1722745624)
* [centralize the error handling](https://github.com/randombit/botan/pull/4318#discussion_r1722826244) into the builder itself, with some template magic to generate decent error messages
   Examples:
   ```
   Scheme 'ML-DSA' requires the 'context' option
   Scheme 'ML-DSA' disdained some options: hash, padding
   Option 'padding' already contains a value
   Scheme 'Unknown Scheme' requires the 'private key' option
   ```

See here: https://godbolt.org/z/3n35TTTW6

doc/deprecated.rst
reneme: Oh, just noticed one more thing: When instantiating PSS, and asking the instance for its `name()`, it identifies as "EMSA4". This is somewhat inconsistent with this deprecation notice. Also note, that this assumption is hard-coded at places. At least I just stumbled over this in `RSA_Signature_Operation::algorithm_identifier()` in `rsa.cpp`.

randombit: Yeah currently it's a bit of a mess atm. Similarly the OID definitions use the (would be) deprecated names. Fixing this should happen but it's a significant bit of work all on its own.

src/lib/pubkey/pk_options.h
```diff
@@ -0,0 +1,197 @@
...
+      std::optional<std::string> m_hash_fn;
+      std::optional<std::string> m_prehash;
+      std::optional<std::string> m_padding;
+      std::optional<std::vector<uint8_t>> m_context;
+      std::optional<std::string> m_provider;
+      std::optional<size_t> m_salt_size;
+      bool m_using_prehash = false;
+      bool m_use_der = false;
+      bool m_deterministic_sig = false;
+      bool m_explicit_trailer_field = false;
```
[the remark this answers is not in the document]
randombit: I went back and forth on this issue but yeah this is probably the way to go.

src/lib/pubkey/pk_options.h
randombit: I went this way to avoid the problem that if there is a `unique_ptr<HashFunction>` in there we have to clone it for each and every `with_foo`. There is maybe a better way to handle this.

reneme: Returning a `PK_Options&` (`return *this`) avoids the copy, as long as the user doesn't assign it to another variable. That allows chaining `with_*` just fine.

reneme:
@randombit Any thoughts on my suggested improvements? I'm happy to adopt this PR and extend it as sketched in the godbolt link from two weeks ago. We do somewhat depend on this to finalize the standardized PQC algorithms before 3.6.0. :)

src/lib/pubkey/pk_options.h
reneme: I have to correct myself to a certain extent: When there's a `unique_ptr<HashFunction>` you are forced to first create an lvalue of the `PK_Options`. E.g. this wouldn't work:

```C++
auto opts = PK_Options().with_hash(Hash_Function::create_or_throw("SHA-256"));
// Above fails, because `with_hash()` returns a `PK_Options&` that cannot be copied into `auto opts`
// (because `PK_Options` contains a unique_ptr).

if (key->algo() == "ML-DSA") {
   opts.with_context("some context");
}
```

In C++20 we'd have to overload `with_hash()` for lvalue-ref and rvalue-ref. Providing both overloads, would compile the above example, as the rvalue-ref variant of `with_hash()` would be called on the temporary object and the temporary would just propagate through `with_hash()` and eventually end up initializing the `auto opts` lvalue.

```C++
class PK_Options {
   PK_Options& with_hash(std::unique_ptr<HashFunction> hash) & {
      m_hash = std::move(hash);
      return *this;
   }

   PK_Options with_hash(std::unique_ptr<HashFunction> hash) && {
      m_hash = std::move(hash);
      return std::move(*this);
   }
};
```

[C++23 introduces "deducing this"](https://devblogs.microsoft.com/cppblog/cpp23-deducing-this/) which would come in handy here. Essentially, `self` is deduced to either an lvalue-ref or an rvalue-ref. And the `decltype(auto)` would just pass the lvalue/rvalue-ref along. The below snippet would also compile the above example in C++23 😢

```C++
class PK_Options {
   decltype(auto) with_hash(this auto&& self, std::unique_ptr<HashFunction> hash) {
      self.m_hash = std::move(hash);
      return self;
   }
};
```

reneme:
> I'm happy to adopt this PR and extend it as sketched in the godbolt link from two weeks ago.

@randombit I started working on it with fair progress. Though, nothing that compiles yet. Just leaving that here to avoid duplicated work and merge conflicts.

reneme:
Here's another iteration ([GodBolt](https://godbolt.org/z/ovKq91Y8Y) that splits "building" and "consuming" into two classes. We hope that this results in a clearer interface that's harder to misuse: i.e. users cannot reach into the built-up options but have to push it into a consumer. This forces users to "commit" on a configuration and prevents consumers from retroactively adding options.

The prototype in Godbolt might look quite complex, but the actual usage surface is super neat, in my opinion. Roughly like this:

```C++
void end_user_view() {
   auto opts = SignerBuilder()
                  .with_padding("OAEP")
                  .with_hash("SHA-256")
                  .with_context(as_byteview("API discussion"));
   PK_Signer signer(key, opts.commit() /* transforms SignerBuilder into a SignerOptions */);
}

PK_Signer::PK_Signer(Private_Key& key, SignerOptions options) 
   m_op(key._create_signing_op(options)
{
   options.validate_option_consumption(); // -> throws Invalid_Argument when some option was not consumed
}

PK_Signing_Op RSA_PrivateKey::_create_signing_op(SignerOptions& opts) {
   const auto padding = opts.padding().required(); // throws Invalid_Argument if not provided
   const auto hash = opts.hash().optional(); // std::nullopt if not provided (algorithm can decide what to do)
   opts.context().not_implemented("comes in 3.8.0"); // explicitly consumes an option but throws Not_Implemented when provided
}
```

src/lib/pubkey/pk_options_impl.cpp
```diff
@@ -0,0 +1,210 @@
...
+      } else {
+         std::string userid;
+         std::string hash = "SM3";
+         auto comma = params.find(',');
+         if(comma == std::string::npos) {
+            userid = params;
+         } else {
+            userid = params.substr(0, comma);
+            hash = params.substr(comma + 1, std::string::npos);
+         }
```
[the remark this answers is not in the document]
reneme: Incorporated in my extensions.

reneme:
To my mind, this API is converging nicely (even though CI is probably not yet happy). As [suggested a few weeks ago](https://github.com/randombit/botan/pull/4318#issuecomment-2297990007), the options builder now incorporates _all degrees of freedom_ of signature creation and verification. That explicitly includes the key pair and the random number generator. The latter should be made optional in the future (for deterministic signatures).

As a result, the c'tors of `PK_Signer` and `PK_Verifier` now just take a `PK_Signature_Options` and bootstrap from there. Consequently, the legacy c'tors (taking key, RNG, params/padding (?), ...) now simply forward these parameters into a constructor of `PK_Signature_Options_Builder` that translates it into a `PK_Signature_Options` which is then passed into the new c'tors mentioned in the beginning.

To hide all of this complexity from the user (most notably: users shouldn't have to know about all those types and classes involved). Here's how an RSA signature rountrip would look like now:

```C++
// Create a key pair
auto pk = sk->public_key();

// Create a PK_Signer and sign a message
auto signer = sk->signer()
                  .with_rng(rng)
                  .with_padding("PSS")
                  .with_hash("SHA-256")
                  .with_salt_size(64)
                  .with_der_encoded_signature(false)
                  .create();
signer.update("Hello, world");
const auto signature = signer.signature(rng);

// Create a PK_Verifier and verify the signature
auto verifier = pk->signature_verifier()
                     .with_padding("PSS")
                     .with_hash("SHA-256")
                     .with_salt_size(64)
                     .with_der_encoded_signature(false)
                     .create();
verifier.update("Hello, world");
const bool ok = verifier.check_signature(signature);
```

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
reneme: With [the extensions I added in the last few days](https://github.com/randombit/botan/pull/4318#issuecomment-2344128175), this has improved much, now.

src/lib/pubkey/pubkey.cpp
```diff
@@ -244,21 +240,36 @@ void check_der_format_supported(Signature_Format format, size_t parts) {
...
+PK_Signer::PK_Signer(PK_Signature_Options options) {
+   // Appease GCC warning about a possibly dangling reference because when using
+   // std::reference_wrapper's implicit conversion to the contained reference
+   // the wrapper does not outlive the reference which seems to confuse GCC.
+   const auto key_wrapped = options.private_key().required();
+   const Private_Key& key = key_wrapped.get();
+
+   // TODO: The downstream algorithms should decide whether to require an RNG
+   //       (e.g. explicitly 'deterministic' signatures should not require it).
+   RandomNumberGenerator& rng = options.rng().required();
```
reneme, line 252: Currently this enforces that all algorithms specify an RNG (via `with_rng()`), even if the algorithm doesn't need an RNG for signing, like some ECC-based schemes or hardware-backed adapters, or when the algorithm is configured to work deterministically (via `with_deterministic_signature()`). In those cases, it shouldn't be required to pass an RNG, and perhaps passing one should even produce an error (because the RNG wasn't consumed).

src/lib/pubkey/pk_options.h
```diff
@@ -0,0 +1,459 @@
...
+      void validate_for_hash_based_signature(std::optional<std::string_view> acceptable_hash = std::nullopt);
+
+      /// This is a convenience helper for algorithms that do not support
+      /// specifying a provider.
+      /// @throws Provider_Not_Found if a provider is set
+      void exclude_provider() {
+         if(auto p = provider().optional()) {
+            throw Provider_Not_Found(product_name(), p.value());
+         };
+      }
```
reneme, line 106: Essentially, this is a convenience construction to replace the ubiquitous sanity check in every algorithm that used to read like:

```C++
if(provider.empty() || provider == "base") {
   // ... do the thing
}
throw Provider_Not_Found();
```

src/lib/pubkey/sphincsplus/sphincsplus_common/sphincsplus.cpp
```diff
@@ -346,21 +345,18 @@ class SphincsPlus_Signature_Operation final : public PK_Ops::Signature {
...
-                   "Unexpected parameters for signing with SPHINCS+");
-
-   const bool randomized = (params == "Randomized");
-   if(provider.empty() || provider == "base") {
-      return std::make_unique<SphincsPlus_Signature_Operation>(m_private, m_public, randomized);
-   }
-   throw Provider_Not_Found(algo_name(), provider);
+   options.exclude_provider();
+   options.context().not_implemented("will come in Botan 3.7.0");
+   options.prehash().not_implemented("will come in Botan 3.7.0");
```
reneme, line 356: Make sure this still aligns with our plans, before merging.

reneme (commented):
From a technical perspective (internals of the options builder) this is essentially done. I think, what's left to do is to double-check the semantics this new API provides. In particular that it enforces sane defaults and avoids footguns as much as reasonable.

@randombit The review comments below provide pointers into places that still need close attention. I'm sure I missed several things, but _I think the current state is a huge API improvement over what we have on master_ right now. For the time being I'll leave this alone, awaiting your feedback.

Regarding timeline: I think that this is a substantial change to the library's "look-and-feel" and _we should consider giving this more time on master than until 3.6.0_ which is just a month away. I suggest to finalize this and merge it soon after the 3.6.0 release. That will postpone support for the "new" FIPS signature features (prehash and context in ML/SLH-DSA) until 3.7.0 but we can probably live with that.

reneme:
Rebased to the 3.6.0 tag. Several conflicts in ML-DSA and SLH-DSA code paths, as was expectable.

reneme:
Rebased once more. The fix for SLH-DSA in 3.6.1 created a conflict again.

reneme:
> Figure out a story for FFI.

Some initial idea (also regarding #4411). Inspired by: https://stackoverflow.com/questions/17604811/builder-pattern-in-c

```C
botan_pk_sign_options_t options;
botan_pk_sign_config_create(&options, rsa_private_key);

botan_pk_sign_config_set_ptr(options, BOTAN_PK_SIGN_RNG, rng);
botan_pk_sign_config_set_str(options, BOTAN_PK_SIGN_PADDING, "PSS");
botan_pk_sign_config_set_str(options, BOTAN_PK_SIGN_HASH, "SHA-256");
// or (?):
botan_pk_sign_config_set_ptr(options, BOTAN_PK_SIGN_HASH, hash_instance);
botan_pk_sign_config_set_int(options, BOTAN_PK_SIGN_SALT_SIZE, 64);
botan_pk_sign_config_set_opt(options, BOTAN_PK_SIGN_DER_ENCODED);

botan_pk_op_sign_t signer;
botan_pk_op_sign_create_from_options(&signer, options);
botan_pk_sign_destroy(options); // this could also be implicit in the previous call, not sure (?)

// use 'signer' as usual...
```

dirkz:
Re the FFI story, looks good to me. Preferably with auto-destruction of options.

reneme:
@randombit I'd love to restart the discussion on this and hopefully converge on a design for the builder.

We'd like to develop the pre-hash feature for SLH-DSA and ML-DSA based on it. That is the final work item for the BSI project and it would be great to have it land in 3.7.0.

Let me know if you would like me to walk you through the idea of this rework.

reneme:
Rebased, resolved conflicts and adapted to the latest changes on master. Most notably: the TPM2 wrapper needed some significant love to be able to handle `PK_Signature_Options&`. Other conflicts were introduced by deprecating the `EC_Point` (#4518) and were easily solvable.

Note that I squashed the history fairly substantially to simplify the rebase. Nevertheless, we should rewrite this history once the API is converged. Intermediate commits don't necessarily build or pass tests currently.

reneme:
Rebased once more to solve conflicts mostly with #4600 and compilation issues introduced in #4642.

randombit:
Continued in #5489

PR #4322, opened 2024-08-21
FIX: X.509 key usage for KEMs

reneme:
As per [draft-ietf-lamps-kyber-certificates-latest, section 3](https://lamps-wg.github.io/kyber-certificates/draft-ietf-lamps-kyber-certificates.html#section-3):

> When any of the ML-KEM AlgorithmIdentifier appears in the SubjectPublicKeyInfo field of an X.509 certificate, the key usage certificate extension MUST only contain keyEncipherment [Section 4.2.1.3](https://rfc-editor.org/rfc/rfc5280#section-4.2.1.3) of [[RFC5280](https://lamps-wg.github.io/kyber-certificates/draft-ietf-lamps-kyber-certificates.html#RFC5280)].

... my intuition is that this should not only count for ML-KEM but any KEM. Until now, we also allowed the key usage "DataEncipherment" for KEMs.

This also repairs a segfault in the tests when enabling module `x509` but not `rsa`.

### DIFF
--- a/src/lib/x509/key_constraint.cpp
+++ b/src/lib/x509/key_constraint.cpp
@@ -74,9 +74,12 @@ bool Key_Constraints::compatible_with(const Public_Key& pub_key) const {
       permitted |= Key_Constraints::KeyAgreement | Key_Constraints::EncipherOnly | Key_Constraints::DecipherOnly;
    }
 
-   if(pub_key.supports_operation(PublicKeyOperation::Encryption) ||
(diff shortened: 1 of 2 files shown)


randombit:
TBH key encipherment vs data encipherment distinction never made sense to me, because what are you using the shared key for if not enciphering data `¯\_(ツ)_/¯` but whatever.

randombit (approved):

PR #4325, opened 2024-08-24
Add jitterentropy as RNG

ghost:
Add [jitterentropy-library](https://github.com/smuellerDD/jitterentropy-library) to Botan as a module. When enabled, the system RNG can use this additional entropy and depends on jitterentropy-library.

So far I have implemented this only for `BOTAN_TARGET_OS_HAS_CCRANDOM` for Botan 3, which got configured fairly automatically for my macOS machine.

If this is something that is useful for Botan. I can easily provide implementation for other systems that I have access to, (windows and ubuntu, plus macOS/Android with arc4random). Plus port it to the Botan 2 branch.

I did _not_ format `system_rng.cpp` yet because most changes would be in existing code (using vs code which should pick up `.clang-format`).

There is an [integration project](https://github.com/plancksecurity/botan-jitter-integration) (using submodules for Botan and the jitter library), in case this is of use for someone. It probably works on linux too, but I'm not sure it the jitter will actually be used there because of the usage of `BOTAN_TARGET_OS_HAS_CCRANDOM`. At least with tests for Botan 2, a WSL ubuntu picked `BOTAN_TARGET_OS_HAS_GETRANDOM` here for `System_RNG_Impl`.

### DIFF
--- a/doc/api_ref/rng.rst
+++ b/doc/api_ref/rng.rst
@@ -233,6 +233,15 @@ PKCS11_RNG
 
 This RNG type allows using the RNG exported from a hardware token accessed via PKCS11.
 
+Jitter_RNG
+^^^^^^^^^^^^^^^^^
+
+This is an RNG based on low-level CPU timing jitter, using the
+`jitterentropy library <https://github.com/smuellerDD/jitterentropy-library>`_.
+
+Can be enabled with ``configure.py`` via ``--enable-module="jitter_rng"``, provided
+you have the library installed and made available to the build, including headers.
+
 Entropy Sources
 ---------------------------------
 
--- a/src/lib/entropy/entropy_srcs.cpp
+++ b/src/lib/entropy/entropy_srcs.cpp
@@ -29,6 +29,10 @@
    #include <botan/internal/getentropy.h>
 #endif
 
+#if defined(BOTAN_HAS_JITTER_RNG)
+   #include <botan/jitter_rng.h>
+#endif
+
 namespace Botan {
 
 namespace {
@@ -83,6 +87,24 @@ class Processor_RNG_EntropySource final : public Entropy_Source {
 
 #endif
 
+#if defined(BOTAN_HAS_JITTER_RNG)
+
+class Jitter_RNG_EntropySource final : public Entropy_Source {
+   public:
+      size_t poll(RandomNumberGenerator& rng) override {
+         const size_t poll_bits = BOTAN_RNG_RESEED_POLL_BITS;
+         rng.reseed_from_rng(m_rng, poll_bits);
+         return poll_bits;
+      }
+
+      std::string name() const override { return m_rng.name(); }
+
+   private:
+      Jitter_RNG m_rng;
+};
+
+#endif
+
 }  // namespace
 
 std::unique_ptr<Entropy_Source> Entropy_Source::create(std::string_view name) {
@@ -115,6 +137,12 @@ std::unique_ptr<Entropy_Source> Entropy_Source::create(std::string_view name) {
 #if defined(BOTAN_HAS_ENTROPY_SRC_WIN32)
    if(name == "system_stats") {
       return std::make_unique<Win32_EntropySource>();
+   }
+#endif
+
+#if defined(BOTAN_HAS_JITTER_RNG)
+   if(name == "jitter_rng") {
+      return std::make_unique<Jitter_RNG_EntropySource>();
    }
 #endif
 
--- a/src/lib/rng/jitter_rng/jitter_rng.cpp
+++ b/src/lib/rng/jitter_rng/jitter_rng.cpp
@@ -1,0 +1,96 @@
+/*
+* CPU Jitter Random Number Generator
+* (C) 2024 Planck Security S.A.
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/jitter_rng.h>
+
+#include <jitterentropy.h>
+
+namespace Botan {
+
+struct Jitter_RNG_Internal {
+      Jitter_RNG_Internal();
+      ~Jitter_RNG_Internal();
+      void collect_into_buffer(std::span<uint8_t> buf);
+
+   private:
+      rand_data* m_rand_data;
+};
+
+Jitter_RNG_Internal::Jitter_RNG_Internal() {
+   static int result = jent_entropy_init();
+
+   // no further details documented regarding the return value
+   BOTAN_ASSERT(result == 0, "JitterRNG: initialization successful");
+
+   constexpr unsigned int oversampling_rate = 0;  // use default oversampling
+   constexpr unsigned int flags = 0;
+
+   m_rand_data = jent_entropy_collector_alloc(oversampling_rate, flags);
+   BOTAN_ASSERT_NONNULL(m_rand_data);
+}
+
+Jitter_RNG_Internal::~Jitter_RNG_Internal() {
+   if(m_rand_data) {
+      jent_entropy_collector_free(m_rand_data);
+      m_rand_data = nullptr;
+   }
+}
+
+void Jitter_RNG_Internal::collect_into_buffer(std::span<uint8_t> buf) {
+   if(buf.empty()) {
+      return;
+   }
+
+   BOTAN_STATE_CHECK(m_rand_data != nullptr);
+
+   ssize_t num_bytes = jent_read_entropy_safe(&m_rand_data, reinterpret_cast<char*>(buf.data()), buf.size());
+   if(num_bytes < 0) {
+      const auto error_msg = [&]() -> std::string_view {
+         switch(num_bytes) {
+            case -1:  // should never happen because of the check above
+               return "JitterRNG: Uninitilialized";
+            case -2:
+               return "JitterRNG: SP800-90B repetition count online health test failed";
+            case -3:
+               return "JitterRNG: SP800-90B adaptive proportion online health test failed";
+            case -4:
+               return "JitterRNG: Internal timer generator could not be initialized";
+            case -5:
+               return "JitterRNG: LAG predictor health test failed";
+            case -6:
+               return "JitterRNG: Repetitive count test (RCT) failed permanently";
+            case -7:
+               return "JitterRNG: Adaptive proportion test (APT) failed permanently";
+            case -8:
+               return "JitterRNG: LAG prediction test failed permanently";
+            default:
+               return "JitterRNG: Error reading entropy";
+         }
+      }();
+      throw Internal_Error(error_msg);
+   }
+
+   // According to the docs, `jent_read_entropy` itself runs its logic as often
+   // as necessary to gather the requested number of bytes,
+   // so this should actually never happen.
+   BOTAN_ASSERT(static_cast<size_t>(num_bytes) == buf.size(), "JitterRNG produced the expected number of bytes");
+}
+
+Jitter_RNG::Jitter_RNG() : m_jitter{std::make_unique<Jitter_RNG_Internal>()} {}
+
+Jitter_RNG::~Jitter_RNG() = default;
+
+void Jitter_RNG::clear() {
+   m_jitter = std::make_unique<Jitter_RNG_Internal>();
+}
+
+void Jitter_RNG::fill_bytes_with_input(std::span<uint8_t> out, std::span<const uint8_t> in) {
+   BOTAN_UNUSED(in);
+
+   m_jitter->collect_into_buffer(out);
+}
+};  // namespace Botan
--- a/src/lib/rng/jitter_rng/jitter_rng.h
+++ b/src/lib/rng/jitter_rng/jitter_rng.h
@@ -1,0 +1,40 @@
+/*
+* CPU Jitter Random Number Generator
+* (C) 2024 Planck Security S.A.
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#ifndef BOTAN_JITTER_RNG_H_
+#define BOTAN_JITTER_RNG_H_
+
+#include <botan/rng.h>
+
+namespace Botan {
+
+struct Jitter_RNG_Internal;
+
+/*
+* RNG using libjitterentropy (https://github.com/smuellerDD/jitterentropy-library).
+*/
+class BOTAN_PUBLIC_API(3, 6) Jitter_RNG final : public RandomNumberGenerator {
+   public:
+      Jitter_RNG();
+      ~Jitter_RNG();
+
+      std::string name() const override { return "JitterRNG"; }
+
+      bool is_seeded() const override { return true; }
+
+      bool accepts_input() const override { return false; }
+
+      void clear() override;
+
+   private:
+      void fill_bytes_with_input(std::span<uint8_t> out, std::span<const uint8_t> in) override;
+
+      std::unique_ptr<Jitter_RNG_Internal> m_jitter;
+};
+}  // namespace Botan
+
+#endif
--- a/src/tests/test_jitter_rng.cpp
+++ b/src/tests/test_jitter_rng.cpp
@@ -1,0 +1,64 @@
+/*
+* (C) 2024 Planck Security S.A.
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <vector>
+
+#include <botan/build.h>
+
+#ifdef BOTAN_HAS_JITTER_RNG
+
+   #include <botan/auto_rng.h>
+   #include <botan/entropy_src.h>
+   #include <botan/jitter_rng.h>
+   #include <botan/system_rng.h>
+
+   #include "tests.h"
+
+namespace Botan_Tests {
+
+namespace {
+
+std::vector<Test::Result> test_jitter_rng() {
+   return {
+      CHECK("Jitter_RNG basic usage",
+            [](Test::Result&) {
+               const std::vector<size_t> sample_counts{0, 1, 2, 4, 64, 128, 512};
+
+               Botan::Jitter_RNG rng;
+               for(auto sample_count : sample_counts) {
+                  [[maybe_unused]] auto buf = rng.random_vec(sample_count);
+               }
+            }),
+
+      CHECK("Jitter_RNG clear",
+            [](Test::Result&) {
+               const std::vector<size_t> sample_counts{64, 128};
+
+               Botan::Jitter_RNG rng;
+               for(auto sample_count : sample_counts) {
+                  [[maybe_unused]] auto buf = rng.random_vec(sample_count);
+                  rng.clear();
+               }
+            }),
+
+      CHECK("JitterRNG as entropy source",
+            [](Test::Result&) {
+               Botan::Entropy_Sources entropy_sources;
+               entropy_sources.add_source(Botan::Entropy_Source::create("jitter_rng"));
+               Botan::AutoSeeded_RNG rng{entropy_sources};
+
+               [[maybe_unused]] auto buf = rng.random_vec(512);
+            }),
+   };
+}
+
+}  // namespace
+
+BOTAN_REGISTER_TEST_FN("rng", "jitter_rng", test_jitter_rng);
+
+}  // namespace Botan_Tests
+
+#endif
--- a/src/lib/ffi/ffi_rng.cpp
+++ b/src/lib/ffi/ffi_rng.cpp
@@ -19,6 +19,10 @@
    #include <botan/processor_rng.h>
 #endif
 
+#if defined(BOTAN_HAS_JITTER_RNG)
+   #include <botan/jitter_rng.h>
+#endif
+
 extern "C" {
 
 using namespace Botan_FFI;
@@ -43,6 +47,11 @@ int botan_rng_init(botan_rng_t* rng_out, const char* rng_type) {
 #if defined(BOTAN_HAS_PROCESSOR_RNG)
       else if((rng_type_s == "rdrand" || rng_type_s == "hwrng") && Botan::Processor_RNG::available()) {
          rng = std::make_unique<Botan::Processor_RNG>();
+      }
+#endif
+#if defined(BOTAN_HAS_JITTER_RNG)
+      else if(rng_type_s == "jitter") {
+         rng = std::make_unique<Botan::Jitter_RNG>();
       }
 #endif
 
--- a/src/lib/rng/jitter_rng/info.txt
+++ b/src/lib/rng/jitter_rng/info.txt
@@ -1,0 +1,17 @@
+load_on dep
+
+<module_info>
+name -> "CPU Jitter Random Number Generator"
+</module_info>
+
+<defines>
+JITTER_RNG -> 20240901
+</defines>
+
+<header:public>
+jitter_rng.h
+</header:public>
+
+<libs>
+all -> jitterentropy
+</libs>
--- a/src/scripts/ci/setup_gh_actions.sh
+++ b/src/scripts/ci/setup_gh_actions.sh
@@ -16,6 +16,17 @@ ARCH="$2"
 
 SCRIPT_LOCATION=$(cd "$(dirname "$0")"; pwd)
 
+function build_and_install_jitterentropy() {
+    mkdir jitterentropy-library
+    curl -L https://github.com/smuellerDD/jitterentropy-library/archive/refs/tags/v3.6.0.tar.gz | tar -xz -C .
+    jel_dir="$(realpath jitterentropy-library-*)"
+    cmake -B "${jel_dir}/build" -S "${jel_dir}" -DCMAKE_BUILD_TYPE=Release
+    cmake --build "${jel_dir}/build"
+    sudo cmake --install "${jel_dir}/build"
+    echo "BOTAN_BUILD_WITH_JITTERENTROPY=1" >> "$GITHUB_ENV"
+    rm -rf "${jel_dir}"
+}
+
 if type -p "apt-get"; then
     # TPM2-TSS library (to build the library against)
     tpm2_specific_packages=("libtss2-dev")
@@ -56,6 +67,7 @@ if type -p "apt-get"; then
     if [ "$TARGET" = "valgrind" ] || [ "$TARGET" = "valgrind-full" ]; then
         # (l)ist mode (avoiding https://github.com/actions/runner-images/issues/9996)
         sudo NEEDRESTART_MODE=l apt-get -qq install valgrind
+        build_and_install_jitterentropy
 
     elif [ "$TARGET" = "static" ]; then
         sudo apt-get -qq install "${tpm2_specific_packages[@]}"
@@ -67,6 +79,7 @@ if type -p "apt-get"; then
 
     elif [ "$TARGET" = "examples" ] || [ "$TARGET" = "tlsanvil" ] || [ "$TARGET" = "clang-tidy" ] ; then
         sudo apt-get -qq install libboost-dev libtss2-dev
+        build_and_install_jitterentropy
 
     elif [ "$TARGET" = "clang" ]; then
         sudo apt-get -qq install clang
@@ -161,6 +174,8 @@ if type -p "apt-get"; then
 
         softhsm2-util --init-token --free --label test --pin 123456 --so-pin 12345678
         echo "PKCS11_LIB=/usr/lib/softhsm/libsofthsm2.so" >> "$GITHUB_ENV"
+
+        build_and_install_jitterentropy
 
     elif [ "$TARGET" = "docs" ]; then
         sudo apt-get -qq install doxygen python3-docutils python3-sphinx
--- a/src/scripts/ci_build.py
+++ b/src/scripts/ci_build.py
@@ -470,6 +470,9 @@ def determine_flags(target, target_os, target_cpu, target_cc, cc_bin, ccache,
                 #       only works for individual test names.
                 test_cmd += ["--tpm2-tcti-name=disabled"]
 
+        if is_running_in_github_actions() and 'BOTAN_BUILD_WITH_JITTERENTROPY' in os.environ:
+            flags += ['--enable-modules=jitter_rng']
+
         if target in ['coverage']:
             flags += ['--with-tpm']
             test_cmd += ['--run-online-tests']
--- a/src/tests/test_ffi.cpp
+++ b/src/tests/test_ffi.cpp
@@ -367,6 +367,15 @@ class FFI_RNG_Test final : public FFI_Test {
             result.test_eq("custom_destroy_cb called", cb_counter, 5);
          }
 
+   #ifdef BOTAN_HAS_JITTER_RNG
+         botan_rng_t jitter_rng;
+         if(TEST_FFI_OK(botan_rng_init, (&jitter_rng, "jitter"))) {
+            std::vector<uint8_t> buf(256);
+            TEST_FFI_OK(botan_rng_get, (jitter_rng, outbuf.data(), buf.size()));
+            TEST_FFI_OK(botan_rng_destroy, (jitter_rng));
+         }
+   #endif
+
          TEST_FFI_OK(botan_rng_destroy, (rng));
          TEST_FFI_OK(botan_rng_destroy, (null_rng));
          TEST_FFI_OK(botan_rng_destroy, (system_rng));


src/lib/jitter/info.txt
```diff
@@ -0,0 +1,19 @@
+# https://botan.randombit.net/handbook/dev_ref/configure.html#adding-a-new-module
+
+load_on never
```
randombit, line 3: This should probably be `load_on dep`  - `load_on never` is mostly just used for blacklisting a known broken module

src/lib/jitter/jitter.cpp
```diff
@@ -0,0 +1,67 @@
...
+* (C) 2024 Planck Security S.A.
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <span>
+
+#include "jitter.h"
+
+#ifdef BOTAN_HAS_JITTER
```
randombit, line 12: This guard is not necessary - this file is not compiled unless the module is included

src/lib/jitter/jitter.cpp
```diff
@@ -0,0 +1,67 @@
+/*
+* Jitter utility functions
+* (C) 2024 Planck Security S.A.
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <span>
+
+#include "jitter.h"
```
randombit, line 10: Should be `<botan/internal/...>`

src/lib/jitter/jitter.cpp
```diff
@@ -0,0 +1,67 @@
...
+void jitter_buffer(rand_data* collector, uint8_t buf[], size_t len) {
+   if(!collector) {
+      return;
+   }
+
+   if(len <= 0) {
+      return;
+   }
+
+   std::unique_ptr<char[]> jitter_data{new char[len]};
```
randombit, line 47: I don't understand why you would allocate this extra buffer; why not just write directly to buf?

src/lib/jitter/jitter.h
```diff
@@ -0,0 +1,24 @@
...
+* (C) 2024 Planck Security S.A.
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/build.h>
+
+#ifdef BOTAN_HAS_JITTER
+extern "C" {
+   #include <jitterentropy.h>
```
randombit, line 12: This external include should be included in at most one source file, not in the header

src/lib/jitter/jitter.h
```diff
@@ -0,0 +1,24 @@
+/*
+* Jitter utility functions interface
+* (C) 2024 Planck Security S.A.
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/build.h>
+
+#ifdef BOTAN_HAS_JITTER
```
randombit, line 10: No reason for the guard - the header will not be viewable/accessible if the module is not enabled at build time

src/lib/jitter/jitter.h
```diff
@@ -0,0 +1,24 @@
...
+extern "C" {
+   #include <jitterentropy.h>
+}
+
+namespace Botan {
+
+rand_data* jitter_collector_create();
+void jitter_collector_free(rand_data* collector);
+void jitter_buffer(rand_data* collector, std::span<uint8_t> buf);
+void jitter_buffer(rand_data* collector, uint8_t buf[], size_t len);
```
randombit, line 20: The interface exposed by this should map to one of the existing types within the library

`EntropySource` - used to feed a userspace PRNG such as HMAC-DRBG
`RandomNumberGenerator` - generates outputs directly

src/lib/rng/system_rng/system_rng.cpp
```diff
@@ -158,10 +165,16 @@ class System_RNG_Impl final : public RandomNumberGenerator {
...
    private:
+#ifdef BOTAN_HAS_JITTER
+      rand_data *m_jitter;
+#endif
       void fill_bytes_with_input(std::span<uint8_t> output, std::span<const uint8_t> /* ignored */) override {
          if(::CCRandomGenerateBytes(output.data(), output.size()) != kCCSuccess) {
             throw System_Error("System_RNG CCRandomGenerateBytes failed", errno);
          }
+#ifdef BOTAN_HAS_JITTER
+         jitter_buffer(m_jitter, output);
```
randombit, line 176: Categorically we do not do this without end-user (ie in application source code) consent. If an application developer types `System_RNG` that means they trust, specifically, the RNG provided by the OS vendor. They are not buying into "The System_RNG XORed with something else that I may be unaware of".

The developer is allowed to type `Jitter_RNG` and buy into "yeah lets use this thing", and then it is visible at the source code level what is happening.

src/lib/jitter/info.txt
```diff
@@ -0,0 +1,19 @@
+# https://botan.randombit.net/handbook/dev_ref/configure.html#adding-a-new-module
```
randombit, line 1: Comment should be removed

src/lib/jitter/info.txt
```diff
@@ -0,0 +1,19 @@
...
+
+load_on never
+
+<module_info>
+name -> "Add jitter entropy to the system RNG"
+</module_info>
+
+<defines>
+JITTER -> 20240819
```
randombit, line 10: This file should move to either `rng` or `entropy` (depending on how you want to present this) and the define updated accordingly

src/lib/jitter/jitter.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+   #include <memory>
+
+namespace Botan {
+
+rand_data* jitter_collector_create() {
+   static int result = jent_entropy_init();
+
+   if(result == 0) {
+      return jent_entropy_collector_alloc(1, 0);
```
randombit, line 22: The parameters here are unclear (and I say this even after examining the declarations and comments in `jitterentropy.h`) so I think named constants + comments are warranted.

src/lib/jitter/jitter.cpp
```diff
@@ -0,0 +1,67 @@
...
+   } else {
+      return nullptr;
```
randombit, line 24: No silent failures - if something is expected to work, it either works or throws an exception

randombit (changes requested):
So some significant skepticism here (ala #4309) but at least this is included in a few widely used Linux distros, so I'm relatively open to merging something that supports this source but I would be quite against doing so where the output of jitter is xored with the system RNG. This should either be its own RNG type, that is used if the application specifically wants to use it, or (I think much more appropriately) it should be an `EntropySource` that is used to feed userspace RNGs.

planckdz:
I rewrote this as an RNG and implemented all feedback points, plus a very basic test to make sure code gets executed. Please re-check and let me know.

src/lib/rng/jitter_rng/jitter.cpp
```diff
@@ -0,0 +1,64 @@
...
+void jitter_collect_into_buffer(rand_data* collector, std::span<uint8_t> buf) {
+   jitter_collect_into_buffer(collector, buf.data(), buf.size());
+}
+
+void jitter_collect_into_buffer(rand_data* collector, uint8_t buf[], size_t len) {
+   if(!collector) {
+      throw Invalid_State("no jitter entropy collector has been allocated");
+   }
+
+   if(len <= 0) {
```
randombit, line 54: An RNG should accept (as a no-op) an output length of zero; output 0 empty bytes always succeeds no matter what the RNG state!

src/lib/rng/jitter_rng/jitter.cpp
```diff
@@ -0,0 +1,64 @@
...
+      throw Invalid_Argument("number of jitter entropy bytes to return must be positive");
+   }
+
+   ssize_t num_bytes = jent_read_entropy(collector, reinterpret_cast<char*>(buf), len);
+   if(num_bytes < 0) {
+      throw Internal_Error("error reading jitter entropy");
+   }
```
randombit, line 61: This doesn’t detect short reads. We should fail (or retry) if the output is less than expected

src/tests/test_jitter_rng.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include <botan/jitter_rng.h>
+
+#include "tests.h"
+
+namespace Botan_Tests
+{
+class Jitter_RNG_Tests final : public Test {
+
+    std::vector<Test::Result> run() override {
+        int sample_count = 512;
```
randombit, line 22: As a (slightly) improved test, check that `Jitter_RNG` can output strings within any range 0..512 ie

```
for(size_t sample_count = 0; sample_count <= 512; ++sample_count) {
   …
}
```

randombit (changes requested):
OK looking better now, thanks for addressing the previous round. Some additional comments outside of the changed code so GH won’t let me comment

- It might be useful to (also) expose `Jitter_RNG` as an entropy source, in the similar way as we do for `System_RNG` and `Processor_RNG`
- Please also update the rng documentation
- Depending on your use case you may want to also expose this to the C wrapper in `botan_rng_init`

src/lib/rng/jitter_rng/jitter.h
```diff
@@ -0,0 +1,21 @@
+/*
+* jitterentropy utility functions
+* (C) 2024 Planck Security S.A.
+*
```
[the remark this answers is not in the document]
randombit: (Or for the collector function just inlined directly - we don’t even call the raw array variant right now IIUC)

planckdz:
I have addressed all issues except:

* Formatting (I'm using `clang-format` but it looks like the CI check is using different parameters)
* Entropy source
* FFI

Will tackle those next.

planckdz:
I have not figured out the formatting issue. If you have a script somewhere in the repo, please tell.
I responded to every feedback point, please let me know if I misunderstood. There's now a jitter RNG and jitter entropy source.

src/lib/rng/jitter_rng/jitter_rng.cpp
```diff
@@ -0,0 +1,126 @@
+/*
+* CPU Jitter Random Number Generator
+* (C) 2024 Planck Security S.A.
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/jitter_rng.h>
+
+extern "C" {
```
randombit, line 10: The version of the header on my system already includes `extern "C"` blocks if `__cplusplus` is defined; is this really necessary with any (recent) version?

src/lib/entropy/entropy_srcs.cpp
```diff
@@ -83,6 +87,24 @@ class Processor_RNG_EntropySource final : public Entropy_Source {
...
+
+class Jitter_RNG_EntropySource final : public Entropy_Source {
+   public:
+      size_t poll(RandomNumberGenerator& rng) override {
+         const size_t poll_bits = BOTAN_RNG_RESEED_POLL_BITS;
+         rng.reseed_from_rng(m_rng, poll_bits);
+         return poll_bits;
+      }
+
+      std::string name() const override { return "jitter_rng"; }
```
randombit, line 100: This should just forward to `m_rng.name()`

src/lib/rng/jitter_rng/jitter_rng.cpp
```diff
@@ -0,0 +1,126 @@
...
+void jitter_collect_into_buffer(rand_data* collector, uint8_t buf[], size_t len) {
+   if(len == 0) {
+      return;
+   }
+
+   if(!collector) {
+      throw Botan::Invalid_State("JitterRNG: No jitter entropy collector has been allocated");
+   }
+
+   if(len < 0) {
```
randombit, line 89: Dead code - `size_t` is unsigned

src/lib/rng/jitter_rng/jitter_rng.h
```diff
@@ -0,0 +1,48 @@
...
+#ifndef BOTAN_JITTER_RNG_H_
+#define BOTAN_JITTER_RNG_H_
+
+#include <botan/rng.h>
+
+struct rand_data;
+
+namespace Botan {
+
+class BOTAN_DLL Jitter_RNG_Data {
```
randombit, line 17: This should just be forward declared here, and defined in the source file

src/lib/rng/jitter_rng/jitter_rng.cpp
```diff
@@ -0,0 +1,126 @@
...
+void Jitter_RNG::clear() {}
+
+void Jitter_RNG::fill_bytes_with_input(std::span<uint8_t> out, std::span<const uint8_t> in) {
+   m_jitter.collect_into_buffer(out);
+}
+};  // namespace Botan
+
+namespace {
+
+rand_data* jitter_collector_create() {
```
randombit, line 55: At this point these functions can/should be inlined into their respective callers in `Jitter_RNG_Data`

src/lib/rng/jitter_rng/jitter_rng.cpp
planckdz: Done.

src/lib/entropy/entropy_srcs.cpp
planckdz: Done.

src/lib/rng/jitter_rng/jitter_rng.cpp
planckdz: Inlined.

src/lib/rng/jitter_rng/jitter_rng.h
planckdz: A naked pointer to `Jitter_RNG_Data` works.

planckdz:
TODO: Comply to doc/dev_ref/contributing.rst.

planckdz:
Please (re-)review.

src/lib/rng/jitter_rng/jitter_rng.h
```diff
@@ -0,0 +1,41 @@
...
+
+      bool accepts_input() const override;
+
+      void clear() override;
+
+   protected:
+      void fill_bytes_with_input(std::span<uint8_t> out, std::span<const uint8_t> in) override;
+
+   private:
+      Jitter_RNG_Data* m_jitter;
```
Delta-dev-99, line 37: There's a way to use `std::unique_ptr` here, like this:
`std::unique_ptr<Jitter_RNG_Data, void(Jitter_RNG_Data*)> m_jitter;`
Then just pass a function pointer (a lambda will do) where the `unique_ptr` is created in the `.cpp` file:
`Jitter_RNG::Jitter_RNG() : m_jitter{new Jitter_RNG_Data{}, [](Jitter_RNG_Data * p){ delete p; }} {}`

This approach is functionally equivalent to the c-like approach of defining a custom destruction function for a custom data structure; something like this:
```
// struct_a is not actually defined, but we can create a pointer to it
struct_a * my_a = create_struct_a();
// do something with the variable
destroy_struct_a(my_a);
```

planckdz: Yes, thank you, that helps the compiler. Pushed. I don't think I'm convinced that this simplifies things, plus there is technically still a naked `delete` in the code base, but it does avoid a destructor, and if it is preferred here then so be it, no problem.

src/lib/rng/jitter_rng/jitter_rng.h
```diff
@@ -0,0 +1,40 @@
...
+*/
+
+#ifndef BOTAN_JITTER_RNG_H_
+#define BOTAN_JITTER_RNG_H_
+
+#include <botan/rng.h>
+
+namespace Botan {
+
+class BOTAN_DLL Jitter_RNG_Data;
```
Delta-dev-99, line 15: Considering that `Jitter_RNG_Data` is only used internally and offers no API available to the user whatsoever, I believe its definition would be better placed inside `Jitter_RNG`. This way there will be one less element polluting the `Botan` namespace.

src/lib/rng/jitter_rng/jitter_rng.h
```diff
@@ -0,0 +1,40 @@
...
+
+      bool accepts_input() const override;
+
+      void clear() override;
+
+   protected:
+      void fill_bytes_with_input(std::span<uint8_t> out, std::span<const uint8_t> in) override;
+
+   private:
+      std::unique_ptr<Jitter_RNG_Data, void (*)(Jitter_RNG_Data*)> m_jitter;
```
Delta-dev-99, line 36: `Jitter_RNG_Data` could be defined here, maybe renamed to just `Data`:
```
private:
   struct Data;
   std::unique_ptr<Data, void (*)(Data*)> m_jitter;
```

src/lib/rng/jitter_rng/jitter_rng.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+#include <botan/jitter_rng.h>
+
+#include <jitterentropy.h>
+
+namespace Botan {
+
+class BOTAN_DLL Jitter_RNG_Data {
+   public:
+      Jitter_RNG_Data();
```
Delta-dev-99, line 16: This part would then change to:
```
class BOTAN_DLL Jitter_RNG::Data {
   public:
      Data();
```
Same for the rest of places where the name of the structure is written

Delta-dev-99 (changes requested):

src/lib/rng/jitter_rng/jitter_rng.cpp
Delta-dev-99: A side comment, just so there are no misunderstandings:
I'm not that familiar with this repo, so my suggestions may not be the usual or preferred thing to do here, it's just what I consider correct or helpful.
My words are not authoritative here. Best regards :)

src/lib/rng/jitter_rng/jitter_rng.h
dirkz: Makes perfect sense, updated.

src/lib/rng/jitter_rng/jitter_rng.h
```diff
@@ -0,0 +1,39 @@
...
+      bool accepts_input() const override;
+
+      void clear() override;
+
+   protected:
+      void fill_bytes_with_input(std::span<uint8_t> out, std::span<const uint8_t> in) override;
+
+   private:
+      struct Rand_Data;
+      std::unique_ptr<Rand_Data, void (*)(Rand_Data*)> m_jitter;
```
Delta-dev-99, line 35: Since you weren't liking the raw `delete p`, I got a better solution for this. I'm confident to say this should be the way to go.
### In short
You need to declare the destrutor of `Jitter_RNG` in the header and implement it in the source file. Then you can use the smart pointer as usual (no weird deleter or creation) without errors:
```
// in header, inside Jitter_RNG definition
public:
   ~Jitter_RNG(); // add this line
   Jitter_RNG();
...
private:
   struct Rand_Data;
   std::unique_ptr<Rand_Data> m_jitter; // normal smart pointer type with default deleter type
```
```
// in source file
Jitter_RNG::~Jitter_RNG() = default; // add this line
Jitter_RNG::Jitter_RNG() : m_jitter{std::make_unique<Rand_Data>()} {} // usual smart pointer creation
```
### Explanation
Turns out, the actual reason for the error you were seeing before, about `Rand_Data` being incomplete, is because `~unique_ptr<Rand_Data>` calls `delete` for it's internal `Rand_Data *` and that call requires `sizeof(Rand_Data)` to be valid. Calls to that `~unique_ptr<Rand_Data>` are implicitly generated inside both the constructor and destructor of `Jitter_RNG` (in the constructor, the destructors of the members are called if an exception is thrown, to avoid leaving members initialized which would lead to memory leaks).
Since the `~Jitter_RNG` was not explicitly defined, it was *implicitly defaulted as an inline function*. This means it's definition is available to all translation units that use it (this condition allows inlining). And in turn, this means that calls to `~unique_ptr<Rand_Data>` are potentially generated in translation units that do not have access to the definition of `Rand_Data`.
By preventing the destructor (and the constructor) from being inline functions, you effectively encapsulate all calls to `~unique_ptr<Rand_Data>` inside the translation unit that defines those functions and `Rand_Data`.

Best regards.

dirkz: Yes, that works, thank you very much again. Updated.

reneme (changes requested):
Here's another round of polishing. Along with a few nits and style/consistency suggestions, I have a left an idea for compacting/improving the test code.

We should also enable this module in at least a few CI targets to ensure sustainability. Since it introduces a new (optional) dependency it'll require a bit of CI script mangling. No worries, I'll send you a patch for it next week.

src/lib/rng/jitter_rng/jitter_rng.h
reneme: I like your train of thought in general (of not polluting the API further). However, if we declare it as a free-standing forward declare (without `BOTAN_DLL`), we can hide the implementation from public linkage within the cpp-file compilation unit. This is fairly common practice throughout the code base.

See also: https://github.com/randombit/botan/pull/4325#discussion_r1784634554

news.rst
```diff
@@ -4,6 +4,9 @@ Release Notes
 Version 3.6.0, Not Yet Released
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+* Add a CPU jitter RNG/entropy source based on the
+  `jitterentropy library <https://github.com/smuellerDD/jitterentropy-library>`_.
+
```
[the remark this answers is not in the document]
dirkz: It's mentioned in https://github.com/randombit/botan/blob/master/doc/dev_ref/contributing.rst. Will remove it.

src/lib/rng/jitter_rng/jitter_rng.h
```diff
@@ -0,0 +1,40 @@
...
+
+      bool accepts_input() const override;
+
+      void clear() override;
+
+   protected:
+      void fill_bytes_with_input(std::span<uint8_t> out, std::span<const uint8_t> in) override;
+
+   private:
+      struct Rand_Data;
```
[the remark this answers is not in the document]
Delta-dev-99: I have my doubts about this. I think the two `struct Jitter_RNG_Internal` declarations would be declaring different structures, since they are in different namespaces.
Actually, while writing this, I made a minimal example of this, the compiler complains about an ambiguous reference to the structure and mentions both declarations (in `library namespace` and in `library namespace::anonymous namespace`).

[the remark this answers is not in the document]
Delta-dev-99: But, structure declarations don't actually generate symbols, the symbols are generated instead for the functions and members of the structure (the parts that you could meaningfully refer to using pointers in code).
I presume Botan uses a compiler flag to make the default visibility hidden, because you have `BOTAN_DLL` that serves the purpose of exposing symbols that would otherwise be hidden. Perhaps then, there is no need to use an anonymous namespace.
Moving the declaration outside of the interface class (`class BOTAN_DLL Jitter_RNG`) might however make sense because visibility attributes propagate to nested definitions, as can be seen [in this SO question](https://stackoverflow.com/questions/1247745/default-visibility-of-c-class-struct-members). But maybe a better solution that also avoids polluting the `Botan` namespace would be to use an attribute to manually prevent the exposition of symbols within the internal structure declaration. It could look like this:
```
#define BOTAN_INTERNAL __attribute__((visibility("hidden")))
...........................................................
class BOTAN_DLL Jitter_RNG
{
   class BOTAN_INTERNAL Rand_Data;
};
```
I don't know if this approach is currently being used in the library or if it works at all (I will test it and post about it in a moment).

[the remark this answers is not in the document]
Delta-dev-99: furthermore, if in the `.cpp` file, all member functions of the implementation class (`Rand_Data`) are inline (either by defining them inside the class definition or explicitly declaring them to be `inline`), they won't generate any symbols.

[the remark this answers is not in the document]
Delta-dev-99: I tested the following approach:
```
#define BOTAN_INTERNAL [[gnu::visibility("hidden")]]
...........................................................
class BOTAN_DLL Jitter_RNG
{
   ...
   class BOTAN_INTERNAL Rand_Data;
};
```
It effectively hides symbols inside `Rand_Data`
The following also works:
```
// in header
class BOTAN_DLL Jitter_RNG
{
   ...
   class Rand_Data;
};

// in .cpp
class BOTAN_INTERNAL Jitter_RNG::Rand_Data
{
   ...
};
```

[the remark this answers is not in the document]
Delta-dev-99: ... I see the problem now. In windows there's no way to "use an attribute to manually prevent the exposition of symbols". There's no undoing the `__declspec(dllexport)`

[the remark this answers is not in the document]
dirkz: Looks like putting both definition and forward declaration into an anonymous namespace under Botan works, at least on macOS clang:

https://github.com/plancksecurity/botan-jitter/blob/jitter_review_2024-10-02/src/lib/rng/jitter_rng/jitter_rng.h#L13
https://github.com/plancksecurity/botan-jitter/blob/jitter_review_2024-10-02/src/lib/rng/jitter_rng/jitter_rng.cpp#L12

src/tests/test_jitter_rng.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+         Test::Result result{"JitterRNG basic usage"};
+         result.test_success();
+         return result;
+      }
+
+      Test::Result test_entropy_source() {
+         Botan::Entropy_Sources entropy_sources;
+         entropy_sources.add_source(Botan::Entropy_Source::create("jitter"));
+         Botan::AutoSeeded_RNG rng{Botan::system_rng(), entropy_sources};
```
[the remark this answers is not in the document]
dirkz: Which was _exactly_ what was happening, due to misnaming the entropy source. Fixed in my review branch, will ping once everything is integrated and ready for another round.

src/tests/test_jitter_rng.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+      std::vector<Test::Result> run() override {
+         std::vector<Test::Result> results;
+         results.push_back(test_basic_rng());
+         results.push_back(test_entropy_source());
+         return results;
+      }
+};
+
+BOTAN_REGISTER_TEST("rng", "jitter_rng_unit", Jitter_RNG_Tests);
```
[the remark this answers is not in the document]
dirkz: Worked flawlessly as-is (except for my misnamed entropy source of course).

src/tests/test_jitter_rng.cpp
```diff
@@ -0,0 +1,61 @@
...
+   #include <botan/jitter_rng.h>
+   #include <botan/system_rng.h>
+
+   #include "tests.h"
+
+namespace Botan_Tests {
+
+class Jitter_RNG_Tests final : public Test {
+      Test::Result test_basic_rng() {
+         const int max_sample_count = 512;
```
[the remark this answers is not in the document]
dirkz: Used `constexpr`, kept the literal `unsigned int` as it is currently declared in the library.

dirkz:
Thank you very much for the great feedback, I have pushed a new update. Especially the declaration/definition of `Jitter_RNG_Internal` may require another look. Please let me know.

Delta-dev-99:
About the declaration/definition of `Jitter_RNG_Internal`:

Using anonymous namespace in headers is strongly discouraged. It generates a warning on my computer (Ubuntu x86_64 using GCC 13.2). I don't think it's harmful in this case because there's only a forward declaration in there but I don't see any benefits either. The default visibility is already hidden.

Personally I think the best would be to just use an attribute to hide the nested internal class declaration, but this will probably not work on windows because there's no way to disable `__dclspec(dllexport)` of a nested definition of an exported class. I'm not sure if `__dclspec(dllexport)` propagates to nested type definitions though. I will test it.

Considering this, the best compromise will probably be to just make the internal class a standalone declaration. It won't generate symbols and we'll have to deal with a bit of namespace pollution.

src/lib/rng/jitter_rng/jitter_rng.cpp
```diff
@@ -0,0 +1,101 @@
...
+         }
+      }();
+      throw Internal_Error(error_msg);
+   }
+   if(num_bytes < buf.size()) {
+      // According to the docs, `jent_read_entropy` itself runs its logic as often
+      // as necessary to gather the requested number of bytes,
+      // so this should actually never happen.
+      throw Internal_Error("JitterRNG: Not enough bytes have been produced");
+   }
```
reneme, line 87: Alternatively:

```suggestion
   // According to the docs, `jent_read_entropy` itself runs its logic as often
   // as necessary to gather the requested number of bytes,
   // so this should actually never happen.
   BOTAN_ASSERT(num_bytes == buf.size(), "JitterRNG produced the expected number of bytes");
```


Explicitly note that such assertions won't be disabled in Botan's a production build (there is also `BOTAN_DEBUG_ASSERT` which can be disabled). Also, assertion failures won't terminate the program.

If the assertion fails, it will throw an `Internal_Error` stating the "asserted" message along with the code location of the assertion. This is a bit more idiomatic in the code base, but no hard feelings either way.

src/lib/rng/jitter_rng/jitter_rng.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include <botan/rng.h>
+
+namespace Botan {
+
+namespace {
+
+struct Jitter_RNG_Internal;
+
+}
```
reneme, line 19:
```suggestion
struct Jitter_RNG_Internal;
```

... as already concluded by @Delta-dev-99 [here](https://github.com/randombit/botan/pull/4325#issuecomment-2395046695).

src/lib/rng/jitter_rng/jitter_rng.h
```diff
@@ -0,0 +1,44 @@
...
+
+      std::string name() const override { return "JitterRNG"; }
+
+      bool is_seeded() const override { return true; }
+
+      bool accepts_input() const override { return false; }
+
+      void clear() override {}
+
+      void fill_bytes_with_input(std::span<uint8_t> out, std::span<const uint8_t> in) override;
```
reneme, line 37:
```suggestion
   private:
      void fill_bytes_with_input(std::span<uint8_t> out, std::span<const uint8_t> in) override;
```

... this is the "internal" customization point of random number generators and should not be available to the end-user. Instead, they would use `randomize_with_input()`. This delegation is done to preserve flexibility in the public API (of the `Botan::RandomNumberGenerator` base class) by decoupling it from the internal polymorphic invocation.

src/lib/rng/jitter_rng/jitter_rng.cpp
```diff
@@ -0,0 +1,101 @@
...
+      throw Internal_Error("JitterRNG: Can not be used");
+   }
+
+   constexpr unsigned int oversampling_rate = 0;  // use default oversampling
+   constexpr unsigned int flags = 0;
+
+   m_rand_data = jent_entropy_collector_alloc(oversampling_rate, flags);
+   if(!m_rand_data) {
+      throw Internal_Error("JitterRNG: Jitter entropy collector could not be allocated");
+   }
```
reneme, line 39:
```suggestion
   BOTAN_ASSERT_NONNULL(m_rand_data);
```

... at the expense of the more detailed error message, of course. But the assertion will state the code location in the error message which is just as helpful in debugging. Just a suggestion, please feel free to leave as is.

src/lib/rng/jitter_rng/jitter_rng.cpp
```diff
@@ -0,0 +1,101 @@
...
+      rand_data* m_rand_data;
+};
+
+Jitter_RNG_Internal::Jitter_RNG_Internal() {
+   static int result = jent_entropy_init();
+
+   if(result != 0) {
+      // no further details documented regarding the return value
+      throw Internal_Error("JitterRNG: Can not be used");
+   }
```
reneme, line 31:
```suggestion
   // no further details documented regarding the return value
   BOTAN_ASSERT(result == 0, "JitterRNG: initialization successful");
```

src/lib/rng/jitter_rng/jitter_rng.h
```diff
@@ -0,0 +1,44 @@
...
+      Jitter_RNG();
+      ~Jitter_RNG();
+
+      std::string name() const override { return "JitterRNG"; }
+
+      bool is_seeded() const override { return true; }
+
+      bool accepts_input() const override { return false; }
+
+      void clear() override {}
```
reneme, line 35: Perhaps it would make sense to re-set the `m_jitter` internal implementation on `clear()`?

Essentially (in the cpp file): `m_jitter = std::make_unique<Jitter_RNG_Internal>()`. Note that the `JitterRNG()` constructor could then simply call `clear()` internally without previously allocating `m_jitter` in its initializer list.

doc/api_ref/rng.rst
```diff
@@ -233,6 +233,15 @@ PKCS11_RNG
 
 This RNG type allows using the RNG exported from a hardware token accessed via PKCS11.
 
+Jitter_RNG
+^^^^^^^^^^^^^^^^^
+
+This is a hardware RNG based on CPU timing jitter, using the
+`jitterentropy library <https://github.com/smuellerDD/jitterentropy-library>`_.
```
reneme, line 240: Nomenclature nit: Botan has a base class `Hardware_RNG` (that simply implements `clear() final { /* NOOP */ }`). But this is used for "actual" hardware RNG interfaces like RDRAND or TPM-based RNGs. It shouldn't be used as a base class for `Jitter_RNG` in my opinion. Though, to prevent confusion, perhaps:

```suggestion
This is an RNG based on low-level CPU timing jitter, using the
`jitterentropy library <https://github.com/smuellerDD/jitterentropy-library>`_.
```

reneme (commented):
A few more minor nits. Most notably, the `BOTAN_ASSERT` suggestions are purely optional, in my opinion.

Also a suggestion for an implementation of `RandomNumberGenerator::clear()` that might (or might not) make sense for this adapter.

Generally, I think this converged nicely! Thanks for your patience. :)

src/lib/rng/jitter_rng/jitter_rng.h
reneme: continuation: https://github.com/randombit/botan/pull/4325#discussion_r1789859848

news.rst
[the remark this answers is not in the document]
reneme: Oh, okay. 😅

src/tests/test_jitter_rng.cpp
```diff
@@ -0,0 +1,49 @@
...
+               constexpr size_t max_sample_count = 512;
+
+               Botan::Jitter_RNG rng;
+               for(size_t sample_count = 0; sample_count <= max_sample_count; ++sample_count) {
+                  [[maybe_unused]] auto buf = rng.random_vec(sample_count);
+               }
+            }),
+
+      CHECK("JitterRNG as entropy source",
+            [](Test::Result& result) {
```
reneme, line 35:
```suggestion
            [](Test::Result&) {
```

... unused warning

src/tests/test_jitter_rng.cpp
```diff
@@ -0,0 +1,49 @@
...
+               Botan::Entropy_Sources entropy_sources;
+               entropy_sources.add_source(Botan::Entropy_Source::create("jitter_rng"));
+               Botan::AutoSeeded_RNG rng{entropy_sources};
+
+               [[maybe_unused]] auto buf = rng.random_vec(512);
+            }),
+   };
+}
+
+BOTAN_REGISTER_TEST_FN("rng", "jitter_rng", test_jitter_rng);
```
reneme, line 45: At least locally on my machine, this test ran for 23 seconds (against jitterentropy trunk). That feels excessive for the actual value added by this test. I think we should look into reducing this runtime drastically, or am I doing something wrong?

I would assume the main culprit is the loop in the first `CHECK`. Perhaps replace that by just trying a few explicit sample lengths.

src/lib/rng/jitter_rng/jitter_rng.cpp
reneme: Building locally, my clang complained about the comparison of integers with different signs. Given that we previously checked for `num_bytes < 0` we should probably getting away with a cast here:

```suggestion
   // According to the docs, `jent_read_entropy` itself runs its logic as often
   // as necessary to gather the requested number of bytes,
   // so this should actually never happen.
   BOTAN_ASSERT(static_cast<size_t>(num_bytes) == buf.size(), "JitterRNG produced the expected number of bytes");
```

src/tests/test_jitter_rng.cpp
dirkz: Yes, it's slow indeed. Replaced by a smaller sample size.

dirkz:
Again, thank you very much for your feedback. I have implemented all of it. Please re-check.

src/tests/test_jitter_rng.cpp
```diff
@@ -0,0 +1,60 @@
...
+      CHECK("Jitter_RNG clear",
+            [](Test::Result&) {
+               const std::vector<size_t> sample_counts{64, 128};
+
+               Botan::Jitter_RNG rng;
+               for(auto sample_count : sample_counts) {
+                  [[maybe_unused]] auto buf = rng.random_vec(sample_count);
+                  rng.clear();
+               }
+            }),
```
reneme, line 43: Thanks!

reneme (approved):
Thank you! This looks good now.

Still to do on our side:
* [x] CI integration (probably for the coverage, valgrind and sanitizer builds ?)

On the note of CI: `jitterentropy-library` (standalone, not the daemon) doesn't seem to exist as a package for ubuntu. Or am I missing something? For testing locally, I built it from source. This isn't a big deal as the library is small, but I'd like to omit this on CI if possible.

src/tests/test_jitter_rng.cpp
```diff
@@ -0,0 +1,60 @@
...
+   #include <botan/auto_rng.h>
+   #include <botan/entropy_src.h>
+   #include <botan/jitter_rng.h>
+   #include <botan/system_rng.h>
+
+   #include "tests.h"
+
+namespace Botan_Tests {
+
+std::vector<Test::Result> test_jitter_rng() {
```
reneme, line 22: While working on the CI: with raised warning levels this doesn't compile, because of a warning:

```
  /home/runner/work/botan/botan/source/src/tests/test_jitter_rng.cpp:22:27: error: no previous declaration for ‘std::vector<Botan_Tests::Test::Result> Botan_Tests::test_jitter_rng()’ [-Werror=missing-declarations]
     22 | std::vector<Test::Result> test_jitter_rng() {
  ccache g++ -fsanitize=address -fsanitize=undefined -fno-sanitize-recover=undefined -fstack-protector -m64 -pthread -std=c++20 -D_REENTRANT  -O1 -fno-optimize-sibling-calls -fno-omit-frame-pointer -Wall -Wextra -Wpedantic -Wstrict-aliasing -Wcast-align -Wmissing-declarations -Wpointer-arith -Wcast-qual -Wzero-as-null-pointer-constant -Wnon-virtual-dtor -Wold-style-cast -Wsuggest-override -Wshadow -Wextra-semi -Werror -Wno-error=strict-overflow -Wno-error=zero-as-null-pointer-constant -Wno-error=non-virtual-dtor -Wno-error=maybe-uninitialized -Wno-error=stringop-overread -Wno-error=stringop-overflow -Wno-error=free-nonheap-object -Wno-error=restrict -DBOTAN_IS_BEING_BUILT -msse2 -I /home/runner/work/botan/botan/build/build/include/public -I /home/runner/work/botan/botan/build/build/include/internal  -isystem /home/runner/work/botan/botan/build/build/include/external -c /home/runner/work/botan/botan/source/src/tests/test_simd.cpp -o /home/runner/work/botan/botan/build/build/obj/test/test_simd.o
        |                           ^~~~~~~~~~~~~~~
```

Essentially, you'll have to wrap this function into an anonymous namespace.

reneme:
Regarding CI: Here a patch that enables running your test in (some) CI configurations: https://github.com/reneme/botan/pull/6/commits/f22d6b894c821593eedf4641964b14d34c86be56
You might need to rebase to the latest `master` branch before this applies. Also note: https://github.com/randombit/botan/pull/4325#discussion_r1793084723

@randombit I had to build jitterentropy from source, because Ubuntu doesn't provide a package for it. 😞

src/tests/test_jitter_rng.cpp
dirkz: Done.

dirkz:
> Regarding CI: Here a patch that enables running your test in (some) CI configurations: [reneme@f22d6b8](https://github.com/reneme/botan/commit/f22d6b894c821593eedf4641964b14d34c86be56) You might need to rebase to the latest `master` branch before this applies. Also note: [#4325 (comment)](https://github.com/randombit/botan/pull/4325#discussion_r1793084723) Please cherry-pick my patch and update this pull request.

Rebased to current master and your CI change cherry-picked.

randombit (approved):
LGTM

randombit:
Seems low risk to me so fine for 3.6

PR #4326, opened 2024-08-26
Add a compile time version declaration in ffi.h

randombit:


reneme (commented):
So the idea is to increment this whenever new API is added/released?

Wouldn't it make sense to just have the release version accessible at compile time?

randombit:
> So the idea is to increment this whenever new API is added/released?

Exactly so.

> Wouldn't it make sense to just have the release version accessible at compile time?

FFI is effectively a parallel library with its own versioning strategy that happens to be compiled into the same binary as the C++ library; the libraries major/minor versions don't really matter to something that consumes the C interface, and we have different semantics, where if version `x` > version `y` then `x` supports all of the features of `y` which is quite unlike major.minor versioning where 4.x will intentionally drop features of 3.x

reneme (approved):

Issue #4328, opened 2024-08-26
Decoding_Error on Public_Key for `RSA/EMSA3(MD5)`


learn-more:
Botan cannot parse this certificate (openssl dump):
```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            06:37:6c:00:aa:00:64:8a:11:cf:b8:d4:aa:5c:35:f4
        Signature Algorithm: md5WithRSAEncryption
        Issuer: CN = Root Agency
        Validity
            Not Before: May 28 22:02:59 1996 GMT
            Not After : Dec 31 23:59:59 2039 GMT
        Subject: CN = Root Agency
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                Public-Key: (512 bit)
                Modulus:
                    00:81:55:22:b9:8a:a4:6f:ed:d6:e7:d9:66:0f:55:
                    bc:d7:cd:d5:bc:4e:40:02:21:a2:b1:f7:87:30:85:
                    5e:d2:f2:44:b9:dc:9b:75:b6:fb:46:5f:42:b6:9d:
                    23:36:0b:de:54:0f:cd:bd:1f:99:2a:10:58:11:cb:
                    40:cb:b5:a7:41
                Exponent: 65537 (0x10001)
        X509v3 extensions:
            commonName: 
                .GFor Testing Purposes Only Sample Software Publishing Credentials Agency
            2.5.29.1: 
                0>.....-...O..a!..dc..0.1.0...U....Root Agency...7l...d......\5.
    Signature Algorithm: md5WithRSAEncryption
    Signature Value:
        2d:2e:3e:7b:89:42:89:3f:a8:21:17:fa:f0:f5:c3:95:db:62:
        69:5b:c9:dc:c1:b3:fa:f0:c4:6f:6f:64:9a:bd:e7:1b:25:68:
        72:83:67:bd:56:b0:8d:01:bd:2a:f7:cc:4b:bd:87:a5:ba:87:
        20:4c:42:11:41:ad:10:17:3b:8c
```

here is the blob:
```
-----BEGIN CERTIFICATE-----
MIIByjCCAXSgAwIBAgIQBjdsAKoAZIoRz7jUqlw19DANBgkqhkiG9w0BAQQFADAW
MRQwEgYDVQQDEwtSb290IEFnZW5jeTAeFw05NjA1MjgyMjAyNTlaFw0zOTEyMzEy
[...]

PR #4330, opened 2024-08-29
Reorder/rephrase output from X509_Certificate::to_string

randombit:


src/lib/x509/x509cert.cpp
```diff
@@ -691,10 +691,20 @@ std::string X509_Certificate::to_string() const {
...
 
+   try {
+      auto pubkey = this->subject_public_key();
+      out << "Public Key [" << pubkey->algo_name() << "-" << pubkey->key_length() << "]\n\n";
+      out << X509::PEM_encode(*pubkey) << "\n";
+   } catch(Decoding_Error&) {
+      const AlgorithmIdentifier& alg_id = this->subject_public_key_algo();
+      out << "Public Key Invalid! OID=" << alg_id.oid().to_formatted_string()
+          << " Hex=" << hex_encode(this->subject_public_key_bitstring()) << "\n";
+   }
```
reneme, line 702: I think it would be helpful if this were to print the error message botan produced when decoding this key, no?

```suggestion
   } catch(const Decoding_Error& ex) {
      const AlgorithmIdentifier& alg_id = this->subject_public_key_algo();
      out << "Public Key Invalid! OID=" << alg_id.oid().to_formatted_string()
          << " Error=" << ex.what() << "\n";
          << " Hex=" << hex_encode(this->subject_public_key_bitstring()) << "\n";
   }
```

reneme (approved):

reneme:
Just for the record. I'm guessing this came as a reaction to: https://github.com/randombit/botan/issues/4328

PR #4335, opened 2024-09-06
Fix TLS::Context Visibility

atreiber94:
`TLS::Context` missed a `BOTAN_PUBLIC_API` flag which resulted in a linker error with a shared build in `examples/tls_stream_coroutine_client.cpp`.

This PR also builds the examples with `shared` to catch such issues.


reneme (approved):
For the record: This was introduced in #3799 which introduced a convenience constructor (`TLS::Context::Context(Server_Information)`), which is the first method in `TLS::Context` that is not implemented in the header itself.

It didn't get caught because this "default" example is currently used only in the examples and not the unit tests.

PR #4338, opened 2024-09-06
Remove code relying on Python distutils

randombit:
This was just used to check if Sphinx supports concurrency, but Sphinx is also barely faster with 1 core vs 32, so it hardly matters.

Debian bug 1080557


randombit:
Test failure probably related to Cloudflare Roughtime server shutdown

PR #4341, opened 2024-09-08
Add NEON implementation of Argon2's BLAMKA function

randombit:


src/lib/utils/simd/simd_2x64/simd_2x64.h
```diff
@@ -0,0 +1,182 @@
...
+
+      void store_le(uint64_t out[2]) const { this->store_le(reinterpret_cast<uint8_t*>(out)); }
+
+      void store_le(uint8_t out[]) const {
+#if defined(BOTAN_TARGET_SUPPORTS_SSSE3)
+         _mm_storeu_si128(reinterpret_cast<__m128i*>(out), m_simd);
+#elif defined(BOTAN_TARGET_SUPPORTS_NEON)
+         vst1q_u8(out, vreinterpretq_u8_u64(m_simd));
+#endif
+      }
```
reneme, line 65: Providing a `std::span<>` based overload (and additionally providing a `store_be`) will make this fulfill the `custom_storable` concept:

```suggestion
      void store_le(uint64_t out[2]) const { this->store_le({reinterpret_cast<uint8_t*>(out), sizeof(out)}); }

      void store_le(std::span<uint8_t, 16> out) const {
#if defined(BOTAN_TARGET_SUPPORTS_SSSE3)
         _mm_storeu_si128(reinterpret_cast<__m128i*>(out.data()), m_simd);
#elif defined(BOTAN_TARGET_SUPPORTS_NEON)
         vst1q_u8(out.data(), vreinterpretq_u8_u64(m_simd));
#endif
      }
```

randombit: This doesn't compile and I don't see an obvious fix, I'm backing out this change.

randombit:
Testing on several different Aarch64 machines show that the NEON code is slower than scalar. Sad!

PR #4344, opened 2024-09-20
Example fix: Use more secure KEM combination in example

FAlbertDev:
This is not a bug, but it may confuse Botan users.
Most modern KEM combiners (that aim for CCA security) also include the ciphertexts when combining the shared secrets. Also, I want to leave a warning to discourage users from tinkering with no well-known KEM combiner using this example.


reneme (approved):
Thanks!

PR #4347, opened 2024-09-23
Deprecate 0xFE30 X25519/Kyber512 code point

reneme:
This code point falls into the 'private code point' region ([RFC 8446 4.2.3](https://www.rfc-editor.org/rfc/rfc8446#section-4.2.3)) and was used by [pq.cloudflareresearch.com](https://pq.cloudflareresearch.com/) for hybrid key exchange using _X25519+KyberR3-512_.

By disabling this code point in `test_cli.py [...] pqc_hybrid_tests` it should fix the currently failing nightly build.

We plan to provide support for _X25519+ML-KEM-768_ as well as _secp256r1+ML-KEM-768_ ([draft-kwiatkowski-tls-ecdhe-mlkem-02](https://datatracker.ietf.org/doc/draft-kwiatkowski-tls-ecdhe-mlkem/)) soon after #3893 is merged.

### DIFF
--- a/src/lib/tls/tls_algos.h
+++ b/src/lib/tls/tls_algos.h
@@ -113,7 +113,7 @@ enum class Group_Params_Code : uint16_t {
 
    // Cloudflare code points for hybrid PQC
    // https://blog.cloudflare.com/post-quantum-for-all/
-   HYBRID_X25519_KYBER_512_R3_CLOUDFLARE = 0xFE30,
+   HYBRID_X25519_KYBER_512_R3_CLOUDFLARE BOTAN_DEPRECATED("removed without replacement") = 0xFE30,
 
    // libOQS defines those in:
    // https://github.com/open-quantum-safe/oqs-provider/blob/main/ALGORITHMS.md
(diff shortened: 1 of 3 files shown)


FAlbertDev (approved):

src/lib/tls/tls_algos.h
```diff
@@ -113,7 +113,7 @@ enum class Group_Params_Code : uint16_t {
 
    // Cloudflare code points for hybrid PQC
    // https://blog.cloudflare.com/post-quantum-for-all/
-   HYBRID_X25519_KYBER_512_R3_CLOUDFLARE = 0xFE30,
+   HYBRID_X25519_KYBER_512_R3_CLOUDFLARE BOTAN_DEPRECATED("removed without replacement") = 0xFE30,
```
randombit, line 116: This could maybe suggest one of the other currently available suites such as x25519/Kyber-768-r3

randombit (approved):

reneme:
Todo: CI failure is relevant, I believe.

reneme:
> Todo: CI failure is relevant, I believe.

In `is_pqc_hybrid()` we're using the deprecated enum value in a public header. The method is marked constexpr and I therefore don't want to move its implementation into the compilation unit. Hence, the 'ignore deprecation' macro. 😞

reneme:
image

🥳

Issue #3727, opened 2023-10-05
std::byte support


Trafo:
Hi,

from my perspective, the way to go for the future is to support std::byte as a native byte type. But Botan actually makes it hard to use std::byte. One of the most critical parts is actually this requirement of secure_alloacator:
```
template <typename T>
#if !defined(_ITERATOR_DEBUG_LEVEL) || _ITERATOR_DEBUG_LEVEL == 0
/*
  * Assert exists to prevent someone from doing something that will
  * probably crash anyway (like secure_vector<non_POD_t> where ~non_POD_t
  * deletes a member pointer which was zeroed before it ran).
  * MSVC in debug mode uses non-integral proxy types in container types
  * like std::vector, thus we disable the check there.
 */
   requires std::is_integral<T>::value
#endif
```
It is still Botan 2 because Botan 3 still does not have a conan recipe. But I would at least allow the std::byte type here. In the long run, it will be very beneficial as well, that the implementer can use the API with std::byte as well.

Best regards,
Trafo

reneme:
I agree that we should give `std::byte` more attention in the public API.

To begin with, there's certainly no reason for `Botan::secure_vector<>` to prohibit using `std::byte`. Though, on the other hand, I don't see much use to it either. At least until the library's public API allows consuming `secure_vector<std::byte>`. Or am I missing something?

That said, adding public overloads that consume and produce data as containers of `std::byte` shouldn't be hard to do -- given that it is designed to co-exist with integer-type API (read: `uint8_t`). Perhaps this can even go hand-in-hand with [the ongoing switch to `std::span<uint8_t>`-based public (and private) APIs](https://github.com/randombit/botan/issues/3318). Even without direct support of `std::byte`, this new interface paradigm might be beneficial to you with a bit of glue:

```C++
template <std::ranges::contiguous_range R>
std::span<const uint8_t> as_botan_span(R&& range)
    requires(sizeof(std::ranges::range_value_t<R>) == 1)
{
    return {reinterpret_cast<const uint8_t*>(std::ranges::data(range)), std::ranges::size(range)};
}
```

I don't think that Botan will internally switch away from `uint8_t` anytime soon, though.

> It is still Botan 2 because Botan 3 still does not have a conan recipe.

Mea culpa. 😞 We're working on that.

Trafo:
[...]

PR #4353, opened 2024-10-05
Allow using `secure_allocator` for enum types, like `std::byte`

Delta-dev-99:
This is a small change aimed to allow the ussage of `std::secure_vector<std::byte>`. Enum types are POD so this shouldn't cause problems.
I don't see the advantage of allowing enums other than `std::byte`, but including another header just for this purpose seems unwarranted.


randombit:
Seems fine to me, @reneme do you see an issue here?

reneme (approved):
This was actually requested exactly a year ago already: https://github.com/randombit/botan/issues/3727

PR #4357, opened 2024-10-08
Feature: ECC Support in TPM2

atreiber94:
This PR extends https://github.com/randombit/botan/pull/4337 to include ECC*, which checks another box of the ToDos in https://github.com/randombit/botan/issues/3877.

## Changes and Additions

- Adjust copyrights for funding attribution (that's why so many files are changed...).
- Analogously to RSA, an **EC Adapter module is added** which allows to manage, load and use* TPM2 EC Keys.
- Redundant logic of the RSA Adapter also used by the EC Adapter is generalized to `TPM2::Signature_Operation` and `TPM2::Verification_Operation` in `tpm2_pkops.h`.
- Augmentation of the **TPM2 Crypto Backend** to support ECDH key exchange in order to allow for **authenticated sessions with TPM2 EC keys**.
- Disentanglement of the Crypto Backend from the Adapters: Before, the Backend required the TPM2 RSA Adapter to establish a session via a TPM2 RSA key. Now, the backend RSA/ECDH functionalities require only the corresponding Botan software RSA/ECDH modules. (These are not  requirements of the backend module though - they are only required if one wants to establish a session with a key of the respective type).
- More defensive input checks in Crypto Backend.

##### * Fine Print: Limitations
- As PK Ops of EC keys this PR **only supports ECDSA** Signing/Verification. Other types such as ECSchnorr and SM2 are not supported.
- Particularly, **ECDH Key Agreement** in Botan with a TPM2 EC key is not yet supported. This is more involved since a TPM EC key may be used for ECDSA/ECDH/..., which is more of an _operation_ as opposed to a _key_ (ECDSA or ECDH ) in Botan. Note that this is not to be confused with ECDH for the Crypto Backend used for the sessions with any EC TPM2 key, which is supported (see above).
- **Only NIST curves are supported** since 25519/448 and Brainpool are not yet supported by `tpm2-tss`.

### DIFF
--- a/src/lib/prov/tpm2/tpm2_util.h
+++ b/src/lib/prov/tpm2/tpm2_util.h
@@ -1,7 +1,7 @@
 /*
 * TPM 2 internal utilities
 * (C) 2024 Jack Lloyd
-* (C) 2024 René Meusel, Amos Treiber - Rohde & Schwarz Cybersecurity GmbH
+* (C) 2024 René Meusel, Amos Treiber - Rohde & Schwarz Cybersecurity GmbH, financed by LANCOM Systems GmbH
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
--- a/src/lib/prov/tpm2/tpm2_util.h
+++ b/src/lib/prov/tpm2/tpm2_util.h
@@ -103,13 +103,19 @@ constexpr auto as_span(tpm2_buffer auto& data) {
    return std::span{data.buffer, data.size};
 }
 
+/// Set the size of @p data to @p length and construct a std::span
+/// as a view into @p data
+constexpr auto as_span(tpm2_buffer auto& data, size_t length) {
+   BOTAN_ASSERT_NOMSG(length <= sizeof(data.buffer));
+   data.size = static_cast<decltype(data.size)>(length);
+   return as_span(data);
+}
+
 /// Copy the @p data into the TPM2 buffer @p dest, assuming that the
 /// provided @p data is not larger than the capacity of the buffer.
 template <tpm2_buffer T>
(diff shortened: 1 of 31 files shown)


src/lib/prov/tpm2/tpm2_util.h
```diff
@@ -156,6 +162,15 @@ struct esys_liberator {
 template <typename T>
 using unique_esys_ptr = std::unique_ptr<T, esys_liberator>;
 
+template <typename T>
+constexpr unique_esys_ptr<T> make_unique_esys_ptr(T obj) {
+   auto ptr = unique_esys_ptr<T>(static_cast<T*>(std::calloc(1, sizeof(T))));
```
reneme, line 167: Note to self: This might actually need to be:
```suggestion
   auto ptr = unique_esys_ptr<T>(std::launder(reinterpret_cast<T*>(std::calloc(1, sizeof(T)))));
```

[[std::launder](https://en.cppreference.com/w/cpp/utility/launder)]

reneme (approved):
I was closely involved with the implementation, so this is really more of a self-review. The CI failures seem to be inherited from master.

randombit (commented):
Looks good. If possible I'd like to avoid introducing new public APIs that use `EC_Point` since ideally that type goes away in Botan4, and using `EC_AffinePoint` from the outset avoids API breakage.

atreiber94:
Thanks for the review!

I transitioned the public interface of `ecc_pubkey_from_tss2_public` from using `EC_Point` to `EC_AffinePoint`.

The Crypto Backend still uses `EC_AffinePoint::to_legacy_point()` because ECDH does not seem transitioned yet, but at least now it does not appear in the public API.

randombit (approved):
LGTM thanks

Issue #4356, opened 2024-10-07
Move GCC and MSVC sanitizer builds to nightly


randombit:
Both are quite slow to run, and 99% redundant with Clang sanitizer - no need to run on every single commit.

PR #4359, opened 2024-10-08
Move the GCC and MSVC sanitizer builds to nightly

randombit:
Clang sanitizer build/test cycle is significantly faster and should catch almost all issues. GH #4356

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -22,6 +22,44 @@ on:
     - cron:  '23 3 * * *'
 
 jobs:
+  sanitizer:
+    name: "Sanitizers"
+    strategy:
+      fail-fast: false
+
+      matrix:
+        include:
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yml
```diff
@@ -22,6 +22,44 @@ on:
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          path: ./source
+
+      - name: Fetch BoringSSL fork for BoGo tests
+        uses: actions/checkout@v4
+        with:
+          repository: randombit/boringssl
+          ref: rene/runner-20240524
```
reneme, line 51: Would be good if we could centralize this somehow. I'll have to update boring soon, anyway. For the new ML-KEM-hybrids. I can take care of this then. Let's try to collectively remember. 😅

reneme (approved):

PR #4361, opened 2024-10-09
TPM2: Basic bindings in FFI & Python

reneme:
Exposes basic TPM 2.0 functionality via the FFI and the Python bindings.

Currently, this allows creating a TPM context object, enabling the usage of the Botan-based crypto backend, setting up an unauthenticated session for parameter encryption and instantiating a TPM-based RNG object.

Additional functionality (particularly to interface with TPM-hosted key material) may be added later.

Here's an example how the above would look in Python:

```python
import botan3 as b

ctx = b.TPM2Context()
if ctx.supports_botan_crypto_backend():
    tpm2_ctx.enable_botan_crypto_backend(botan.RandomNumberGenerator("user"))
session = b.TPM2UnauthenticatedSession(ctx)

rng = b.RandomNumberGenerator("tpm2", tpm2_context=ctx, tpm2_sessions=session)

print("Randomness from TPM:", rng.get(32))
```

### DIFF
--- a/src/lib/ffi/ffi_tpm2.cpp
+++ b/src/lib/ffi/ffi_tpm2.cpp
@@ -1,0 +1,205 @@
+/*
+* (C) 2024 Jack Lloyd
+* (C) 2024 René Meusel, Amos Treiber - Rohde & Schwarz Cybersecurity GmbH
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/ffi.h>
+
+#include <botan/internal/ffi_pkey.h>
+#include <botan/internal/ffi_rng.h>
+#include <botan/internal/ffi_util.h>
+
+#if defined(BOTAN_HAS_TPM2)
+   #include <botan/tpm2_context.h>
+   #include <botan/tpm2_key.h>
+   #include <botan/tpm2_rng.h>
+   #include <botan/tpm2_session.h>
+#endif
+
+extern "C" {
+
+using namespace Botan_FFI;
+
+#if defined(BOTAN_HAS_TPM2)
+
+// These wrappers are required since BOTAN_FFI_DECLARE_STRUCT internally
+// produces a unique pointer, but the TPM types are meant to be used as
+// shared pointers.
+
+struct botan_tpm2_ctx_wrapper {
+      std::shared_ptr<Botan::TPM2::Context> ctx;
+};
+
+struct botan_tpm2_session_wrapper {
+      std::shared_ptr<Botan::TPM2::Session> session;
+};
+
+BOTAN_FFI_DECLARE_STRUCT(botan_tpm2_ctx_struct, botan_tpm2_ctx_wrapper, 0xD2B95E15);
+BOTAN_FFI_DECLARE_STRUCT(botan_tpm2_session_struct, botan_tpm2_session_wrapper, 0x9ACCAB52);
+
+}  // extern "C"
+
+namespace {
+
+Botan::TPM2::SessionBundle sessions(botan_tpm2_session_t s1, botan_tpm2_session_t s2, botan_tpm2_session_t s3) {
+   return Botan::TPM2::SessionBundle((s1 != nullptr) ? safe_get(s1).session : nullptr,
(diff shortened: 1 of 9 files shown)


src/lib/ffi/ffi_tpm2.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+int botan_tpm2_ctx_enable_crypto_backend(botan_tpm2_ctx_t ctx, botan_rng_t rng) {
+#if defined(BOTAN_HAS_TPM2)
+   return BOTAN_FFI_VISIT(ctx, [=](botan_tpm2_ctx_wrapper& ctx_wrapper) -> int {
+      Botan::RandomNumberGenerator& rng_ref = safe_get(rng);
+
+      // The lifetime of the RNG used for the crypto backend should be managed
+      // by the TPM2::Context. Here, we just need to trust the user that they
+      // keep the passed-in RNG instance intact for the lifetime of the context.
+      std::shared_ptr<Botan::RandomNumberGenerator> rng_ptr(&rng_ref, [](auto*) {});
```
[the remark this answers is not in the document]
randombit: :/ We could have this function steal (move) the RNG but that’s not otherwise a convention used in the FFI.

Another alternative would be to remove the choice of which RNG is used, and just wrap eg system_rng. However that goes against the library-wide convention that the application should in all contexts be able to specify what RNG they prefer to use.

src/python/botan3.py
```diff
@@ -20,11 +21,13 @@
 
 from ctypes import CDLL, CFUNCTYPE, POINTER, byref, create_string_buffer, \
     c_void_p, c_size_t, c_uint8, c_uint32, c_uint64, c_int, c_uint, c_char, c_char_p, addressof
+from typing import Callable
```
randombit, line 24: This introduces a hard dependency on at least Python 3.5. Which seems fine - that version was released in 2015 - but I realize we currently do not have any specific documentation regarding what our minimum supported Python version is.

src/python/botan3.py
```diff
@@ -631,14 +644,111 @@ def const_time_compare(x, y):
...
         self.__obj = c_void_p(0)
-        _DLL.botan_rng_init(byref(self.__obj), _ctype_str(rng_type))
+        if rng_type == 'tpm2':
+            ctx = kwargs.get("tpm2_context", None)
+            if not ctx or not isinstance(ctx, TPM2Context):
+                raise BotanException("Cannot instantiate a TPM2-based RNG without a TPM2 context, pass tpm2_context= argument?")
+            sessions = TPM2Session.session_bundle_(kwargs.get("tpm2_sessions", None))
+            _DLL.botan_tpm2_rng_init(byref(self.__obj), ctx.handle_(), *sessions)
+        else:
+            _DLL.botan_rng_init(byref(self.__obj), _ctype_str(rng_type))
```
randombit, line 751: Should check here that kwargs is otherwise empty

randombit (approved):
One tiny nit in the Python

atreiber94 (approved):
LGTM

src/lib/ffi/ffi_tpm2.cpp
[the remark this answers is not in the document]
reneme: I mean, this solution works and is acceptable in my opinion. Certainly better than hardcoding the RNG.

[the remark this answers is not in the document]
randombit: No objections to doing it this way, just wanted to offer any alternative approaches I could think of.

src/python/botan3.py
reneme: No strong need to introduce this. We could remove the type annotation for the sake of caution.

src/python/botan3.py
reneme: Used `kwargs.pop()` and validated that `kwargs` is empty after reading all relevant arguments. Also added a test for that.

reneme:
Thanks for the reviews. I addressed Jack's suggestion and rebased to master (after #4325 caused a merge conflict).

randombit (approved):
:+1:

PR #4367, opened 2024-10-10
Refactor public key related performance test code

randombit:


src/cli/perf_pk_kem.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+      std::vector<std::string> keygen_params(const PerfConfig& config) const override {
+         BOTAN_UNUSED(config);
+         return {
+            "Kyber-512-r3",
+            "Kyber-512-90s-r3",
+            "Kyber-768-r3",
+            "Kyber-768-90s-r3",
+            "Kyber-1024-r3",
+            "Kyber-1024-90s-r3",
```
reneme, line 99: Admittedly, a bit late to the party. Nevertheless: When building with `--disable-modules=kyber_90s`, this will error with:
```
Kyber-Kyber-512-r3 65559 KEM encrypt/sec; 0.02 ms/op 38070 cycles/op (26176 ops in 399 ms)
Kyber-Kyber-512-r3 52351 KEM decrypt/sec; 0.02 ms/op 47679 cycles/op (26176 ops in 500 ms)
Error: Not implemented requested Kyber mode is not enabled in this build
```

... instead of skipping the Kyber 90s modes. Similarly for other algorithms where specific groups of modes may be disabled at build time.

Could be fixed by (for instance):

```C++
#if defined(BOTAN_HAS_KYBER)
   "Kyber-512-r3",
   "Kyber-768-r3",
   "Kyber-1024-r3",
#endif
#if defined(BOTAN_AS_KYBER_90S)
   "Kyber-512-90s-r3",
   "Kyber-768-90s-r3",
   "Kyber-1024-90s-r3",
#endif
```

PR #4368, opened 2024-10-11
FFI/Python: Expose raw public/private key encoding functions

reneme:
This is mostly to replace the dedicated functions for kyber's raw encoding that were added to FFI and Python (#3546). The new functions are generically applicable for most other key types. We'll deprecate the replaced functions in the ML-KEM pull request.

See also: #4366


randombit (approved):
:+1: Thanks

PR #4374, opened 2024-10-15
ML-KEM: Provide type aliases

reneme:
This provides type aliases `ML_KEM_PublicKey`, `ML_KEM_PrivateKey` and `ML_KEM_Mode` allowing downstream users to refer to FIPS 203 by its actual name and avoid "Kyber" in the public API. It builds an API bridge into a future where we want to remove Kyber (Round 3).

Note that this approach might be confusing because the ABI-symbols retain "Kyber" everywhere. Though, we'd like to defer the full scale rename until we actually remove Kyber round 3 (and with it all the multi-module complexity). @randombit Do you agree with this approach? If yes, we'd do the same with ML-DSA and SLH-DSA.

### DIFF
--- a/src/examples/ml_kem.cpp
+++ b/src/examples/ml_kem.cpp
@@ -1,4 +1,4 @@
-#include <botan/kyber.h>
+#include <botan/ml_kem.h>
 #include <botan/pubkey.h>
 #include <botan/system_rng.h>
 #include <array>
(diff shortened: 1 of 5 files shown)


randombit:
> Though, we'd like to defer the full scale rename until we actually remove Kyber round 3 (and with it all the multi-module complexity).

Sounds good to me, I think in the near future, most people experimenting with ML-KEM already know what Kyber is and are unlikely to be confused overmuch. That of course will change as ML-KEM becomes more widely used and the historical context is less recent (see also Rijndael vs AES, Keccak vs SHA-3, ...)

randombit (approved):

PR #4380, opened 2024-10-16
test_pk_pad.cpp #ifdef fix

FAlbertDev:
The `#ifdef` used for the first test class in `src/tests/test_pk_pad.cpp` is too unspecific. If any module in `src/lib/pk_pad` is activated while `eme_pkcs1` is deactivated, the test runs and fails. This PR fixes the issue by correcting the respective `#ifdef`.


randombit (approved):
Thanks

PR #4381, opened 2024-10-16
cli speed: also measure keygen for specified duration

cr-marcstevens:
Running `botan speed --msec=10000` will run all individual benchmarks for at least 10secs, 
except key generation for public key operations.

This pull request measures (and records) also public key generation for the specified duration.

### DIFF
--- a/src/cli/perf_pk_enc.cpp
+++ b/src/cli/perf_pk_enc.cpp
@@ -48,6 +48,9 @@ class PerfTest_PKEnc : public PerfTest {
          auto keygen_timer = config.make_timer(nm, 1, "keygen");
 
          auto key = keygen_timer->run([&] { return Botan::create_private_key(algo, rng, params); });
+         while(keygen_timer->under(msec)) {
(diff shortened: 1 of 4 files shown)


randombit (commented):
Looks good, thank you. Can you rebase against master and squash the two commits?

randombit (approved):
LGTM!

cr-marcstevens:
Note I've only squashed the 2 commits at the moment. 

It's not rebased yet. I still need to figure out how to do manual rebasing between two repositories.
But usually GitHub also allows to merge a PR with rebasing it (instead of a merge commit)?

reneme:
> But usually GitHub also allows to merge a PR with rebasing it (instead of a merge commit)?

This is disabled for us on purpose, as it would invalidate GPG signatures on the rebased commits.

> I still need to figure out how to do manual rebasing between two repositories.

Since you simply pushed onto your own master, it should work roughly like that:

```
git remote add upstream https://github.com/randombit/botan
git pull --rebase upstream master
```

cr-marcstevens:
Thanks a lot for the commands. That has worked!

PR #4384, opened 2024-10-16
Refactor: Centralize X25519/X448 all-zero result rejection

reneme:
For certain public point inputs, both X25519 and X448 may produce all-zero shared secrets. This is by design and does virtually not appear in practice. Therefore, we should technically not even have to verify that in the key agreement operations. RFC 7748 does allow for it, though:

> both sides MAY check, without leaking extra information about the value of K, whether the resulting shared K is the all-zero value and abort if so.

In contrast, RFC 8422 _mandates_ that the computed premaster secret be checked for the all-zero value when these algorithms are used in TLS:

> With X25519 and X448, a receiving party MUST check whether the computed premaster secret is the all-zero value and abort the handshake if so, as described in Section 6 of [RFC7748].

This pull request centralizes said check into the key agreement operations of X25519 and X448. This reduces complexity by removing the need for special treatment when using them in TLS. Note that we currently fail to perform this check when X25519 comes paired with Kyber/ML-KEM in a hybrid key exchange. I found that while working on updating our adapter to the boringssl test suite, see here: https://github.com/google/boringssl/commit/ed95627d2f074b6a8636fc8eaaa595903dbfcc99.

Thanks @davidben.

### DIFF
--- a/src/lib/pubkey/curve448/x448/x448.cpp
+++ b/src/lib/pubkey/curve448/x448/x448.cpp
@@ -116,6 +116,20 @@ class X448_KA_Operation final : public PK_Ops::Key_Agreement_with_KDF {
          auto shared_secret = encode_point(x448(k, u));
          CT::unpoison(shared_secret);
 
+         // RFC 7748 Section 6.2
+         //    As with X25519, both sides MAY check, without leaking extra
+         //    information about the value of K, whether the resulting shared K
+         //    is the all-zero value and abort if so.
+         //
+         // TODO: once the generic Key Agreement operation creation is equipped
+         //       with a more flexible parameterization, this check could be
+         //       made optional.
+         //       For instance: `sk->agree().with_optional_sanity_checks(true)`.
+         //       See also:     https://github.com/randombit/botan/pull/4318
+         if(CT::all_zeros(shared_secret.data(), shared_secret.size()).as_bool()) {
+            throw Invalid_Argument("X448 public point appears to be of low order");
+         }
+
          return shared_secret;
       }
 
--- a/src/lib/pubkey/x25519/x25519.cpp
+++ b/src/lib/pubkey/x25519/x25519.cpp
@@ -10,6 +10,7 @@
 #include <botan/ber_dec.h>
 #include <botan/der_enc.h>
 #include <botan/rng.h>
+#include <botan/internal/ct_utils.h>
 #include <botan/internal/fmt.h>
 #include <botan/internal/pk_ops_impl.h>
 
@@ -120,7 +121,24 @@ class X25519_KA_Operation final : public PK_Ops::Key_Agreement_with_KDF {
 
       size_t agreed_value_size() const override { return 32; }
 
-      secure_vector<uint8_t> raw_agree(const uint8_t w[], size_t w_len) override { return m_key.agree(w, w_len); }
+      secure_vector<uint8_t> raw_agree(const uint8_t w[], size_t w_len) override {
+         auto shared_key = m_key.agree(w, w_len);
+
+         // RFC 7748 Section 6.1
+         //    Both [parties] MAY check, without leaking extra information about
+         //    the value of K, whether K is the all-zero value and abort if so.
+         //
+         // TODO: once the generic Key Agreement operation creation is equipped
+         //       with a more flexible parameterization, this check could be
+         //       made optional.
+         //       For instance: `sk->agree().with_optional_sanity_checks(true)`.
+         //       See also:     https://github.com/randombit/botan/pull/4318
+         if(CT::all_zeros(shared_key.data(), shared_key.size()).as_bool()) {
+            throw Invalid_Argument("X25519 public point appears to be of low order");
+         }
+
+         return shared_key;
+      }
 
    private:
       const X25519_PrivateKey& m_key;
--- a/src/lib/tls/tls12/msg_client_kex.cpp
+++ b/src/lib/tls/tls12/msg_client_kex.cpp
@@ -118,15 +118,6 @@ Client_Key_Exchange::Client_Key_Exchange(Handshake_IO& io,
          auto shared_secret =
             state.callbacks().tls_ephemeral_key_agreement(curve_id, *private_key, peer_public_value, rng, policy);
 
-         // RFC 8422 - 5.11.
-         //   With X25519 and X448, a receiving party MUST check whether the
-         //   computed premaster secret is the all-zero value and abort the
-         //   handshake if so, as described in Section 6 of [RFC7748].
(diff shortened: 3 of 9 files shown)


davidben:
Ah nice. Didn't know you all were also using our tests!

The all zeros check is to avoid non-contributory behavior, where one side can force the secret into some particular value (notably all zeros). Whether this matters depends on the protocol. It's been a while but if I recall correctly, TLS 1.3 is fine without it (but asks you do it anyway to be safe), but TLS 1.2 without EMS has some subtle cases where this matters. This came up with the 3SHAKE vulnerability.

src/lib/pubkey/curve448/x448/x448.cpp
```diff
@@ -116,6 +116,20 @@ class X448_KA_Operation final : public PK_Ops::Key_Agreement_with_KDF {
...
+         //    information about the value of K, whether the resulting shared K
+         //    is the all-zero value and abort if so.
+         //
+         // TODO: once the generic Key Agreement operation creation is equipped
+         //       with a more flexible parameterization, this check could be
+         //       made optional.
+         //       For instance: `sk->agree().with_optional_sanity_checks(true)`.
+         //       See also:     https://github.com/randombit/botan/pull/4318
+         if(CT::all_zeros(shared_secret.data(), shared_secret.size()).as_bool()) {
+            throw Invalid_Argument("X448 public point appears to be a of low order");
```
randombit, line 130: Using `Invalid_Argument` here seems somewhat strange in that the implication of `Invalid_Argument` is usually "there is a bug in your application and you called the library in the wrong way" rather than "the peer sent something invalid".

That said finite field DH uses `Invalid_Argument` for a similar purpose *and* it doesn't seem like there is any current exception type that would be more appropriate, so `¯\_(ツ)_/¯`

randombit: "to be a of" -> "to be of" ?

src/lib/pubkey/x25519/x25519.cpp
```diff
@@ -120,7 +121,24 @@ class X25519_KA_Operation final : public PK_Ops::Key_Agreement_with_KDF {
...
+         //    Both [parties] MAY check, without leaking extra information about
+         //    the value of K, whether K is the all-zero value and abort if so.
+         //
+         // TODO: once the generic Key Agreement operation creation is equipped
+         //       with a more flexible parameterization, this check could be
+         //       made optional.
+         //       For instance: `sk->agree().with_optional_sanity_checks(true)`.
+         //       See also:     https://github.com/randombit/botan/pull/4318
+         if(CT::all_zeros(shared_key.data(), shared_key.size()).as_bool()) {
+            throw Invalid_Argument("X25519 public point appears to be a of low order");
```
randombit, line 137: "to be a of" -> "to be of" ?

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -371,6 +371,9 @@ size_t Channel_Impl_12::from_peer(std::span<const uint8_t> data) {
    } catch(Decoding_Error&) {
       send_fatal_alert(Alert::DecodeError);
       throw;
+   } catch(Invalid_Argument&) {
```
randombit, line 374: One small concern I'd have here is that if we have an uncaught `Invalid_Argument` not due to the X25519/X448 all zeros check, we will send an incorrect and probably quite confusing/misleading alert...

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -371,6 +371,9 @@ size_t Channel_Impl_12::from_peer(std::span<const uint8_t> data) {
...
+      send_fatal_alert(Alert::IllegalParameter);
```
randombit, line 375: I don't see anything in RFC 8422 or 8446 that indicates what alert to send in this situation. If BoGo has a specific expectation I guess we should match that, but otherwise either of `decrypt_error` or `illegal_parameter` seem reasonable choices.

randombit (approved):
Change itself is fine - probably best to safeguard applications against non-contributory behavior, and I can see how it simplifies things wrt the TLS hybrid KEMs.

It does really feel like `Invalid_Argument` is not the right thing to indicate the error condition at hand. But I'm not sure what to name "the peer sent something such that the computed result is not valid" so ...

randombit:
Immediately after posting the review I realized what I'd rather see - 

instead of a blanket conversion from `Invalid_Argument` to anything (except `internal_error`, which for most cases is the correct alert to send if we're failing with `Invalid_Argument`) catch `Invalid_Argument` during the KEM operation and convert that to a `TLS_Exception`.

src/lib/pubkey/curve448/x448/x448.cpp
reneme: The selection of exception was inspired by the fact that TLS is supposed to produce an "illegal parameter" in this case.

src/lib/tls/tls12/tls_channel_impl_12.cpp
reneme: [RFC 8446 Section 6](https://datatracker.ietf.org/doc/html/rfc8446#section-6) is probably authorative here:

> Peers which receive a message which
   cannot be parsed according to the syntax (e.g., have a length
   extending beyond the message boundary or contain an out-of-range
   length) MUST terminate the connection with a "decode_error" alert.
   Peers which receive a message which is syntactically correct but
   semantically invalid (e.g., a DHE share of p - 1, or an invalid enum)
   MUST terminate the connection with an "illegal_parameter" alert.

BoGo switched to "illegal_parameter" recently: https://github.com/google/boringssl/commit/72a60506ded3407454d6ddc1d848c266020c0c82

reneme:
> instead of a blanket conversion from Invalid_Argument to anything (except internal_error, which for most cases is the correct alert to send if we're failing with Invalid_Argument) wrap the KEM operation in a try/catch, catch Invalid_Arguments and convert immediately to a TLS_Exception.

I'll remove the translation entirely in this pull request (mapping to "internal error" be default) and will implement the suggested error translation in the upcoming pull request that updates BoGo. There, we'll have to remap a few more errors in similar places and I'd rather have those adaptions in a single place. (See https://github.com/google/boringssl/commit/72a60506ded3407454d6ddc1d848c266020c0c82)

src/lib/tls/tls12/tls_channel_impl_12.cpp
reneme: Fair. Probably we should aim to not remove those catch-all error matches in general and do the remapping closer to the suspected culprits. For instance, `Invalid_Authentication_Tag` should probably be handled closer to the actual call to the AEAD.

PR #4385, opened 2024-10-16
Remove the BigInt based NIST reduction functions

randombit:
Going forward the assumption is that if the user cares about ECC performance then pcurves is enabled.


reneme (approved):

randombit:
@guidovranken FYI when merged this will break the cryptofuzz build

guidovranken:
> @guidovranken FYI when merged this will break the cryptofuzz build

Thanks, now removed https://github.com/guidovranken/cryptofuzz/commit/687d3064c5cef2b0fe1f30824065a2f2c9c0bbd8

PR #4375, opened 2024-10-15
PQC: Add hybrid groups x25519/ML-KEM-768 and secp256r1/ML-KEM-768

reneme:
### Pull Request Dependencies

* #4389 - which contains tests for x25519/ML-KEM
* #4391 - fixing an issue in ML-KEM when running BoGo with sanitizers

### Description

Those groups are described in [draft-kwiatkowski-tls-ecdhe-mlkem](https://datatracker.ietf.org/doc/draft-kwiatkowski-tls-ecdhe-mlkem/02/) and the code points are [registered by IANA](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8).

PR #4389, opened 2024-10-17
TLS: BoGo tests update

reneme:
This updates the BoGo test suilte to boringssl's trunk of yesterday. This includes two noteworthy changes:

1. Addition of test cases that use X25516/ML-KEM-768 for key exchange
  (its usage is disabled in this pull request and will be enabled [later](https://github.com/randombit/botan/pull/4375))
2. More specific tests regarding sanity checks and error handling of the key exchange

The latter required a minor refactoring in `TLS::Callbacks` where I disentangled public key deserialization and KEM-Encaps respectively KEX-Agree. `TLS::Callbacks::tls_deserialize_peer_public_key()` now handles the plain serialization while the pre-existing `tls_kem_encapsulate()` and `tls_ephemeral_key_agreement()` perform error re-mapping and the actual usage of the keys.
Certain users may benefit from the new callback when the want to introduce custom deserialization logic but want to rely on the standard implementation of key exchange. See the updated example.

Also, a few tweaks in TLS alert usage were needed to fulfill the new tests.

### DIFF
--- a/src/editors/vscode/scripts/bogo.py
+++ b/src/editors/vscode/scripts/bogo.py
@@ -6,7 +6,7 @@ from common import run_cmd, get_concurrency
 
 
 BORING_REPO = "https://github.com/randombit/boringssl.git"
-BORING_BRANCH = "rene/runner-20240524"
+BORING_BRANCH = "rene/runner-20241016"
 
 BORING_PATH = "build_deps/boringssl"
 BOGO_PATH = os.path.join(BORING_PATH, "ssl", "test", "runner")
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,7 +189,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: randombit/boringssl
-          ref: rene/runner-20240524
+          ref: rene/runner-20241016
           path: ./boringssl
         if: matrix.target == 'coverage' || matrix.target == 'sanitizer'
 
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -48,7 +48,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: randombit/boringssl
-          ref: rene/runner-20240524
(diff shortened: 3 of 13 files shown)


randombit:
CI failures look relevant

reneme:
> CI failures look relevant

Let's try again.

src/editors/vscode/scripts/bogo.py
```diff
@@ -6,7 +6,7 @@
 
 
 BORING_REPO = "https://github.com/randombit/boringssl.git"
-BORING_BRANCH = "rene/runner-20240524"
+BORING_BRANCH = "rene/runner-20241016"
```
reneme, line 9: This config is duplicated thrice in the repository: [I had said we should centralize this](https://github.com/randombit/botan/pull/4359#discussion_r1792331001).

Suggestion:

Lets add a file in `src/configs/repo_config.env` that can contain such configurations. Side note: Personally, I'd probably add it as a root-level file (eg. `.repo_config.env`), but I remember that you wanted to keep the repo root as tidy as possible.

This file could then be consumed by `.github/workflows/ci.yml` and `.../nightly.yml` as well as here and (for now) just contain:

```txt
# the fork of boringssl that should be used for BoGo tests
BORINGSSL_REPO=https://github.com/randombit/boringssl.git

# the branch in our fork of boringssl that should be used for BoGo tests
BORINGSSL_BRANCH=rene/runner-20241016
```

We might have more such configuration that is worth to centralize like that. In any case: I'd do this adaption in an independent PR (likely after 3.6.0).

What are your thoughts on this?

randombit: > Lets add a file in src/configs/repo_config.env that can contain such configurations. 

Sure sounds good.

> Side note: Personally, I'd probably add it as a root-level file (eg. .repo_config.env), but I remember that you wanted to keep the repo root as tidy as possible.

Indeed so. Thank you for accomodating my autistic needs on this topic.

randombit (approved):

PR #4393, opened 2024-10-18
PQC: Add TLS 1.3 groups for (pure) ML-KEM-512, -768 and -1024

reneme:
Those groups are described in [draft-connolly-tls-mlkem-key-agreement](https://datatracker.ietf.org/doc/draft-connolly-tls-mlkem-key-agreement/02/) and propose the registration of code points `0x0512`, `0x0768` and `0x1024` by IANA. Alas, IANA did not adopt them as of now. Also, [the email exchange on the IETF mailing list](https://mailarchive.ietf.org/arch/search/?q=%22draft-connolly-tls-mlkem-key-agreement%22) shows controversy whether or not it is too early to use standalone PQC for TLS key negotiation.

Nevertheless, think we should provide support for them to give users an easy way to use standalone PQC if they want or have to. Only, I'm wondering whether we should be bullish on the code points and call them `TLS::Group_Params::ML_KEM_512` or rather `TLS::Group_Params::ML_KEM_512_PRELIMINARY` (or similar).

### DIFF
--- a/src/lib/tls/tls_callbacks.cpp
+++ b/src/lib/tls/tls_callbacks.cpp
@@ -34,6 +34,10 @@
    #include <botan/kyber.h>
 #endif
 
+#if defined(BOTAN_HAS_ML_KEM)
+   #include <botan/ml_kem.h>
+#endif
+
 #if defined(BOTAN_HAS_FRODOKEM)
    #include <botan/frodokem.h>
 #endif
@@ -228,6 +232,12 @@ std::unique_ptr<Public_Key> TLS::Callbacks::tls_deserialize_peer_public_key(
    }
 #endif
 
+#if defined(BOTAN_HAS_ML_KEM)
+   if(group_params.is_pure_ml_kem()) {
+      return std::make_unique<ML_KEM_PublicKey>(key_bits, ML_KEM_Mode(group_params.to_string().value()));
+   }
+#endif
+
 #if defined(BOTAN_HAS_KYBER)
    if(group_params.is_pure_kyber()) {
       return std::make_unique<Kyber_PublicKey>(key_bits, KyberMode(group_params.to_string().value()));
@@ -244,6 +254,12 @@ std::unique_ptr<Public_Key> TLS::Callbacks::tls_deserialize_peer_public_key(
 }
 
 std::unique_ptr<Private_Key> TLS::Callbacks::tls_kem_generate_key(TLS::Group_Params group, RandomNumberGenerator& rng) {
+#if defined(BOTAN_HAS_ML_KEM)
+   if(group.is_pure_ml_kem()) {
+      return std::make_unique<ML_KEM_PrivateKey>(rng, ML_KEM_Mode(group.to_string().value()));
+   }
+#endif
+
 #if defined(BOTAN_HAS_KYBER)
    if(group.is_pure_kyber()) {
       return std::make_unique<Kyber_PrivateKey>(rng, KyberMode(group.to_string().value()));
--- a/src/lib/tls/tls_algos.cpp
+++ b/src/lib/tls/tls_algos.cpp
@@ -186,6 +186,16 @@ std::optional<Group_Params> Group_Params::from_string(std::string_view group_nam
       return Group_Params::KYBER_1024_R3_OQS;
    }
 
+   if(group_name == "ML-KEM-512") {
+      return Group_Params::ML_KEM_512;
+   }
+   if(group_name == "ML-KEM-768") {
+      return Group_Params::ML_KEM_768;
+   }
+   if(group_name == "ML-KEM-1024") {
+      return Group_Params::ML_KEM_1024;
+   }
+
    if(group_name == "eFrodoKEM-640-SHAKE") {
       return Group_Params::eFRODOKEM_640_SHAKE_OQS;
    }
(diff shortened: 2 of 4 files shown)


randombit (commented):
Seems ok to support pure PQ I guess (but can anyone _really_ not spare the extra 32 bytes for a X25519 or P-256 kex, idk…) but I’d lean towards deferring this to 3.7.0 just to avoid too much proliferation of draft TLS suites.

*Especially* what I’d be concerned with is, clearly this draft is still in flux. If something changes incompatibly in a future draft but using the same cutesy suite ids, we can either support the old draft (remaining compatible with ourselves, but incompatible with other impls) or support the new draft (and then break our ability to talk cross versions).

Already I’m trying to figure out if we can get away with dropping the Kyber r3 suites without too much disruption… I guess I did something similar with the Newhope suites back in the day and nobody complained.

reneme:
> (but can anyone really not spare the extra 32 bytes for a X25519 or P-256 kex, idk…)

The discussion revolves around compliance, and whether or not hybrid is advised, prohibited or tolerated by certain governmental bodies.

> I’d lean towards deferring this to 3.7.0 just to avoid too much proliferation of draft TLS suites.

Agreed. In the mean time, users have [the option to integrate custom groups](https://botan.randombit.net/handbook/api_ref/tls.html#code-example-tls-client-using-custom-curve). This even got a little easier with `TLS::Callbacks::tls_deserialize_peer_public_key()` (#4389).

> Already I’m trying to figure out if we can get away with dropping the Kyber r3 suites without too much disruption…

Perhaps its worth looking into generalizing the concept of key exchange groups in TLS to simplify the addition of "custom curves". Currently this it is fairly involved with understanding and customizing non-trivial methods in `TLS::Policy` and `TLS::Callbacks`. If that was simpler and less error prone, we might just get away with deprecating/removing the draft curves and pointing to a new (simple) interface.

I would argue that users who currently use any of the draft groups probably don't do so in production (hopefully) yet, and are thus able to adapt fairly quickly.

In the best case some (public) curve registration might be as simple as this (ignoring TLS 1.2's explicit DH groups):

```C++
register_group_for_code_point(
    0x1024, "ML-KEM-1024",
    [](RandomNumberGenerator& rng) {
        return std::make_unique<ML_KEM_PrivateKey>(rng, ML_KEM_Mode(group.to_string().value()));
    },
    [](std::span<const uint8_t> pubkey_bytes) {
        return std::make_unique<ML_KEM_PublicKey>(key_bits, ML_KEM_Mode(group_params.to_string().value()));
    });
```

randombit (approved):
Considering further - since we are dropping the pure Kyber r3 suites in 3.7.0 it would be nice if the ML-KEM variants were available at the same time.

I was concerned that these might change as the draft evolves but considering ML-KEM itself is fixed it is hard to see how this would occur, there isn't even a KEM share ordering to define in this case.

The suites very well may not see wide adoption, but that's also true for eg the FrodoKEM suites, so I don't think that in and of itself is a blocker.

src/lib/tls/tls_callbacks.cpp
```diff
@@ -244,6 +254,12 @@ std::unique_ptr<Public_Key> TLS::Callbacks::tls_deserialize_peer_public_key(
 }
 
 std::unique_ptr<Private_Key> TLS::Callbacks::tls_kem_generate_key(TLS::Group_Params group, RandomNumberGenerator& rng) {
+#if defined(BOTAN_HAS_ML_KEM)
+   if(group.is_pure_ml_kem()) {
+      return std::make_unique<Kyber_PrivateKey>(rng, KyberMode(group.to_string().value()));
```
randombit, line 259: `ML_KEM_Mode`?

reneme:
Oh no! [The 05-revision of the document](https://www.ietf.org/archive/id/draft-connolly-tls-mlkem-key-agreement-05.html) adapts the code points to 0x0200 (512), 0x0201, and 0x0202, also [listed as-such by IANA](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml). Those are clashing with the code points previously assigned to eFrodoKEM by OQS which we adopted previously.

How should we handle this? I'd probably suggest to just adapt the Frodo code points to [whatever OQS uses now](https://github.com/open-quantum-safe/oqs-provider/blob/main/ALGORITHMS.md) (and perhaps mark them as experimental?).

randombit:
Eep. I guess we will have to do something, since implementations that support these groups, even if we do not, will cause problems for us since we will disagree on what group is in use. I guess if OQS has changed the group IDs to something else already, we should do the same.

(Also I do not understand why IANA is assigning final numbers for drafts, but apparently that’s not a new thing since some other PQ drafts, even withdrawn, have assigned numbers…)

randombit:
For archeology reasons commit message should mention we had to change the Frodo IDs due to the conflict. Otherwise lgtm

reneme:
This could cause a merge conflict (or missing adaptions) in conjunction with #4305, because it introduces a new `is_pure_ml_kem()` in `Group_Params`. This new method must be considered in the refactorings done there. Also, we are planning to drop support for the round 3 version of Kyber, which will need modifications in similar locations. @randombit I'll keep my fingers off of the merge button on this PR. Please integrate as you see fit or leave a message if you need my help here.

PR #4134, opened 2024-06-17
CatKDF Example

FAlbertDev:
As mentioned in PR #4076, the [CatKDF KEM Combiner](https://www.etsi.org/deliver/etsi_ts/103700_103799/103744/01.01.01_60/ts_103744v010101p.pdf) doesn't quite fit into Botan's interface. So, this pull request provides an example of using CatKDF to meet one's specific needs. The example includes a standalone function implementation of CatKDF, a Known Answer Test (KAT) from the specification, and a sample key exchange that demonstrates using CatKDF with Kyber and ECDH (using the KEX-to-KEM adapter).

## Pull Request Dependencies
- #4067

PR #4394, opened 2024-10-18
Update the roadmap [ci skip]

randombit:


randombit:
@reneme wasn't sure if there is anything you wanted to add here

reneme:
Apart from Classic McEliece, we'll probably finish the work on the KEM combiners (in case we will see some consensus there), currently #4134, #4124, #4119.

We might [continue working on the TPM 2.0 integrations](https://github.com/randombit/botan/issues/3877#issuecomment-2418765070), but that's not in a state where I would want to commit to concrete features, yet.

[The API modernization we started regarding pubkey-operations](https://github.com/randombit/botan/pull/4318) might be worth noting for Botan3. Especially considering that this API style may-or-may-not proliferate into other parts of the library.

PR #4395, opened 2024-10-19
Add more value barriers to avoid compiler induced side channels

randombit:
The paper https://arxiv.org/pdf/2410.13489 claims that on specific architectures Clang and GCC may introduce jumps here. The donna128 issues only affect 32-bit processors, which explains why we would not see it in the x86-64 valgrind runs.

The GHASH leak would seem to be generic but the authors only observed it on RISC-V.

### DIFF
--- a/src/lib/pubkey/x25519/donna.cpp
+++ b/src/lib/pubkey/x25519/donna.cpp
@@ -344,14 +344,13 @@ void fmonty(uint64_t result_two_q_x[5],
 * This function performs the swap without leaking any side-channel
 * information.
 */
-inline void swap_conditional(uint64_t a[5], uint64_t b[5], uint64_t c[5], uint64_t d[5], uint64_t iswap) {
-   const uint64_t swap = 0 - iswap;
-
+inline void swap_conditional(uint64_t a[5], uint64_t b[5], uint64_t c[5], uint64_t d[5], CT::Mask<uint64_t> swap) {
    for(size_t i = 0; i < 5; ++i) {
-      const uint64_t x0 = swap & (a[i] ^ b[i]);
-      const uint64_t x1 = swap & (c[i] ^ d[i]);
+      const uint64_t x0 = swap.if_set_return(a[i] ^ b[i]);
       a[i] ^= x0;
       b[i] ^= x0;
+
+      const uint64_t x1 = swap.if_set_return(c[i] ^ d[i]);
       c[i] ^= x1;
       d[i] ^= x1;
    }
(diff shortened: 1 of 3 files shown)


randombit:
Despite the claims in the paper, the authors never contacted me about this :disappointed: I guess we're lucky I happened to see it on a Saturday morning just before the release.

src/lib/utils/ghash/ghash.cpp
```diff
@@ -60,16 +58,16 @@ void GHASH::ghash_multiply(secure_vector<uint8_t>& x, std::span<const uint8_t> i
       uint64_t Z[2] = {0, 0};
 
       for(size_t i = 0; i != 64; ++i) {
-         const uint64_t X0MASK = (ALL_BITS + (X[0] >> 63)) ^ ALL_BITS;
-         const uint64_t X1MASK = (ALL_BITS + (X[1] >> 63)) ^ ALL_BITS;
+         const auto X0MASK = CT::Mask<uint64_t>::expand_top_bit(X[0]);
+         const auto X1MASK = CT::Mask<uint64_t>::expand_top_bit(X[1]);
```
randombit, line 62: This area was not flagged in the paper but seems an obvious target for Clang's range analysis

reneme:
I wonder if it is possible to come up with (say) a custom clang-tidy rule that flags such locations for potential range analysis optimizations or the like.

randombit:
Something that flags the obvious cases would be nice to have.

One search, that would have caught the ghash issue already, is `x & 1` where `x` is secret, is almost certainly in need of using `CT::Mask` or `CT::Choice`, or at the very least a value barrier.

randombit:
@atreiber94  @reneme Possible issue here https://github.com/randombit/botan/blob/master/src/lib/pubkey/frodokem/frodokem_common/frodo_matrix.cpp#L84

reneme:
I feel we should build a nightly test matrix that runs the valgrind suite on a wider spectrum of compilers, targets and optimization levels, in the hope that this increases the signal of such vulnerabilities. Or at least see if we can build regression tests based on our existing tools.

randombit:
We're practically limited by what architectures are natively supported; running our tests inside qemu inside valgrind would be ... something. But adding even just x86-32 and aarch64 to the valgrind runs would improve coverage quite a bit.

src/lib/pubkey/x25519/donna.cpp
```diff
@@ -345,13 +345,14 @@ void fmonty(uint64_t result_two_q_x[5],
 * information.
 */
 inline void swap_conditional(uint64_t a[5], uint64_t b[5], uint64_t c[5], uint64_t d[5], uint64_t iswap) {
-   const uint64_t swap = 0 - iswap;
+   const auto swap = CT::Mask<uint64_t>::expand(iswap);
```
reneme, line 348: Nit-picky, but should't `swap_conditional` take a `CT::Choice` for `iswap` here?

src/lib/pubkey/x25519/donna.cpp
```diff
@@ -376,14 +377,14 @@ void cmult(uint64_t resultx[5], uint64_t resultz[5], const uint8_t n[32], const
...
-      const uint64_t bit6 = (n[31 - i] >> 1) & 1;
-      const uint64_t bit7 = (n[31 - i] >> 0) & 1;
+      const uint64_t bit0 = CT::value_barrier((n[31 - i] >> 7) & 1);
+      const uint64_t bit1 = CT::value_barrier((n[31 - i] >> 6) & 1);
+      const uint64_t bit2 = CT::value_barrier((n[31 - i] >> 5) & 1);
+      const uint64_t bit3 = CT::value_barrier((n[31 - i] >> 4) & 1);
+      const uint64_t bit4 = CT::value_barrier((n[31 - i] >> 3) & 1);
+      const uint64_t bit5 = CT::value_barrier((n[31 - i] >> 2) & 1);
+      const uint64_t bit6 = CT::value_barrier((n[31 - i] >> 1) & 1);
+      const uint64_t bit7 = CT::value_barrier((n[31 - i] >> 0) & 1);
```
reneme, line 387: Perhaps worth adding a higher-level `CT::is_even()` for this (if it doesn't exist; I didn't check), to make this:
`const CT::Choice bit0 = CT::is_odd(n[31 - i] >> 7)`. This would also play nicely with my other comment.

reneme (approved):

src/lib/pubkey/x25519/donna.cpp
randombit: A `Choice` would logically make sense but we'd then have to convert it to a `Mask<uint64_t>` to perform the swap so I changed `iswap` arg to a `Mask<uint64_t>`

src/lib/pubkey/x25519/donna.cpp
randombit: Already existing was `CT::Mask::expand_bit` which I used here.

reneme (approved):

PR #4397, opened 2024-10-19
Reduce risk of compiler value range propogation in FrodoKEM sampling

randombit:
No compiler has been found to perform optimizations that introduce a side channel here, but the construct as written was somewhat risky.

### DIFF
--- a/src/lib/pubkey/frodokem/frodokem_common/frodo_matrix.cpp
+++ b/src/lib/pubkey/frodokem/frodokem_common/frodo_matrix.cpp
@@ -78,10 +78,11 @@ FrodoMatrix FrodoMatrix::sample(const FrodoKEMConstants& constants,
 
    load_le<uint16_t>(elements.data(), r.data(), n);
 
-   for(size_t i = 0; i < n; ++i) {
-      uint32_t sample = 0;                         // Avoid integral promotion
-      const uint16_t prnd = elements.at(i) >> 1;   // Drop the least significant bit
-      const uint16_t sign = elements.at(i) & 0x1;  // Pick the least significant bit
+   for(auto& elem : elements) {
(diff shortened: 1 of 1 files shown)


src/lib/pubkey/frodokem/frodokem_common/frodo_matrix.cpp
```diff
@@ -78,18 +78,21 @@ FrodoMatrix FrodoMatrix::sample(const FrodoKEMConstants& constants,
...
       // No need to compare with the last value.
       for(size_t j = 0; j < constants.cdf_table_len() - 1; ++j) {
          // Constant time comparison: 1 if CDF_TABLE[j] < s, 0 otherwise.
          sample += CT::Mask<uint16_t>::is_lt(constants.cdf_table_at(j), prnd).if_set_return(1);
       }
       // Assuming that sign is either 0 or 1, flips sample iff sign = 1
-      elements.at(i) = static_cast<uint16_t>((-sign) ^ sample) + sign;
+      const uint16_t sample_u16 = static_cast<uint16_t>(sample);
+
+      elem = sign.select(~sample_u16, sample_u16) + (sign.value() & 1);
```
reneme, line 95: I'm assuming that `sign.value()` has a built-in value barrier, but does that matter in the second term? After all the compiler could still decide to branch on the last bit of whatever `sign.value()` returns. With the `select` statement we might as well pull the conditional addition of 1 into the select, no?

```suggestion
      elem = sign.select(~sample_u16 + 1, sample_u16);
```

reneme (approved):
One suggestion to simplify the calculation a bit further.

src/lib/pubkey/frodokem/frodokem_common/frodo_matrix.cpp
randombit: Good idea

Issue #4401, opened 2024-10-23
Illegal instruction crashes botan


SecTec:
### General Info

OS: Arch Linux
CPU: AMD FX(tm)-6300 Six-Core Processor
Affected version: 3.6.0-1
KeepassXC version: 2.7.9-4

### Issue

KeepassXC crashes after update of libbotan from v3.5.0-1 to v3.6.0-1.

**Debug output (GDB):**

```
> gdb --quiet --args keepassxc
…
(gdb) run
Starting program: /usr/bin/keepassxc
…
Thread 1 "keepassxc" received signal SIGILL, Illegal instruction.
0x00007ffff6fea869 in ?? () from /usr/lib/libbotan-3.so.6
…
```

### Conclusion

Maybe some code was added to libbotan in the last update that uses some CPU instructions / features that are not present in my CPU.

CPU Info (lscpu):

```
> LANG=C lscpu
Architecture:             x86_64
  CPU op-mode(s):         32-bit, 64-bit
  Address sizes:          48 bits physical, 48 bits virtual
  Byte Order:             Little Endian
CPU(s):                   6
  On-line CPU(s) list:    0-5
Vendor ID:                AuthenticAMD
  Model name:             AMD FX(tm)-6300 Six-Core Processor
    CPU family:           21
    Model:                2
    Thread(s) per core:   1
    Core(s) per socket:   6
    Socket(s):            1
    Stepping:             0
    Frequency boost:      enabled
    CPU(s) scaling MHz:   81%
    CPU max MHz:          3500.0000
    CPU min MHz:          1400.0000
    BogoMIPS:             6988.40
    Flags:                fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt pdpe1gb rdtscp lm co
                          nstant_tsc rep_good nopl nonstop_tsc cpuid extd_apicid aperfmperf pni pclmulqdq monitor ssse3 fma cx16 sse4_1 sse4_2 popcnt aes xsave avx f16c lahf_
                          lm cmp_legacy svm extapic cr8_legacy abm sse4a misalignsse 3dnowprefetch osvw ibs xop skinit wdt fma4 tce nodeid_msr tbm topoext perfctr_core perfct
[...]

PR #4404, opened 2024-10-24
Bump news and version to 3.6.1

randombit:
#4401 is bad enough to be worth a quick patch release

### DIFF
--- a/news.rst
+++ b/news.rst
@@ -1,6 +1,19 @@
 Release Notes
 ========================================
 
+Version 3.6.1, Not Yet Released
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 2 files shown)


news.rst
```diff
@@ -1,6 +1,19 @@
 Release Notes
 ========================================
 
+Version 3.6.1, Not Yet Released
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* Notice: Botan 3.7.0 will remove support for the currently supported
+  experimental Kyber r3 TLS ciphersuites, leaving only the standardized
+  ML-KEM suites. (GH #4403)
```
reneme, line 9: Yeah, February should be enough grace period for that. IMO: These things were added to enable easy-to-use experimentation with the new groups. Now that we have the standard, it should be fine to just remove those.

reneme (approved):

PR #4408, opened 2024-10-25
When scalar blinding use an odd blinding factor

randombit:
Since the group order is odd, this ensures that k and k + b*n have different parity. Analysis suggests that this reduces the incidence of self-additions which may occur during the multiplication algorithm.


FAlbertDev (approved):
Nice, thank you!

Issue #4340, opened 2024-09-07
Hard-coded usage of the system RNG in ffi_pk_op


planckdz:
There are functions in ffi_pk_op.cpp that directly use `Botan::system_rng()` without giving the caller the option to override that, e.g. here:

https://github.com/randombit/botan/blob/3.5.0/src/lib/ffi/ffi_pk_op.cpp#L40

Should the RNG not be a parameter?

randombit:
It should have been a paramater. I don’t recall anymore why this ended up hard coded in this way. We cannot change the current implementations, since many existing users depend on the current ABI, but we could certainly add new functions which allow specifying an RNG to use.

dirkz:
Implementation happening here: https://github.com/randombit/botan/pull/4411

PR #4411, opened 2024-10-27
Hard-coded usage of the system RNG in ffi_pk_op

dirkz:
Fixes https://github.com/randombit/botan/issues/4340

Open questions:

* How best to unify the `_with_rng` versions with their original ones?

### DIFF
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -1681,6 +1681,10 @@ typedef struct botan_pk_op_sign_struct* botan_pk_op_sign_t;
 BOTAN_FFI_EXPORT(2, 0)
 int botan_pk_op_sign_create(botan_pk_op_sign_t* op, botan_privkey_t key, const char* hash_and_padding, uint32_t flags);
 
+BOTAN_FFI_EXPORT(3, 7)
+int botan_pk_op_sign_create_with_rng(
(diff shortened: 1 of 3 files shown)


reneme:
Zooming out a bit: Perhaps this could be answered in the context of #4318. Essentially, for C++ we're experimenting with a new builder-style API to configure the public key operations. For instance:

```C++
auto rng = AutoSeeded_RNG{};

auto signer = sk->signer()
                  .with_rng(rng)
                  .with_padding("PSS")
                  .with_hash("SHA-256")
                  /* ... */
                  .create();

signer.sign(); 
```

In #4318 there's an open TODO on how to expose this paradigm for the FFI. I don't have a concrete proposal at this point. But perhaps its worth thinking in that direction and finding a usable (generic) design for this before creating many more FFI function with `_with_rng()`.

(Please don't take this a actual opposition to these new functions from my side. I just try to connect some loose ends here.)

dirkz:
> In #4318 there's an open TODO on how to expose this paradigm via the FFI. I don't have a concrete proposal at this point. But perhaps its worth thinking in that direction and finding a usable (generic) design for this before creating many more FFI function with `_with_rng()`.

Thank you very much for the heads up, this _is_ something to consider.

reneme:
Some initial idea: https://github.com/randombit/botan/pull/4318#issuecomment-2451487304
The linked API suggestion is very much a work in progress. Feel free to join the discussion.

PR #3840, opened 2023-12-14
CI: Enable Iterator Debugging

reneme:
### TODO

* [ ] temporarily revert 664a6f7df61563c9d724e906d124c51130b2e880 to ensure the bug would have been caught

Closes #3737

Issue #3737, opened 2023-10-07
CI: Look into iterator debugging


reneme:
As discussed in #3734.

thesamesam:
See https://github.com/randombit/botan/pull/3736#issuecomment-1751913027.

I don't see why the lightweight options (which don't break ABI) couldn't be added at the very least.

randombit:
GCC and MSVC already have configurations set up for iterator debugging - we just aren't using them in CI for whatever reason

randombit:
#4413

PR #4413, opened 2024-10-27
Add iterator debugging to CI

randombit:
See #3737 #3840

### DIFF
--- a/src/tests/test_bigint.cpp
+++ b/src/tests/test_bigint.cpp
@@ -544,11 +544,17 @@ BOTAN_REGISTER_TEST("math", "bn_rshift", BigInt_Rshift_Test);
 
 Test::Result test_const_time_left_shift() {
    Test::Result result("BigInt const time shift");
-   const size_t bits = Test::run_long_tests() ? 2 * 4096 : 2048;
+   const size_t bits = Test::run_long_tests() ? 4096 : 2048;
+
+   auto rng = Test::new_rng("const_time_left_shift");
+
+   result.start_timer();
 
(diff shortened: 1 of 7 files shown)


randombit:
Can't repro the CI failures locally, this may be something in the TPM2 code. Will investigate more later.

src/tests/test_bigint.cpp
```diff
@@ -544,11 +544,17 @@ BOTAN_REGISTER_TEST("math", "bn_rshift", BigInt_Rshift_Test);
...
+   auto rng = Test::new_rng("const_time_left_shift");
+
+   result.start_timer();
 
    Botan::BigInt a = Botan::BigInt::with_capacity(bits / sizeof(Botan::word));
    for(size_t i = 0; i < bits; ++i) {
-      a.set_bit(i);
+      if(rng->next_byte() & 1) {
+         a.set_bit(i);
+      }
```
reneme, line 557: Off topic: should we introduce something like `rng->next_bit() -> CT::Choice`? Not sure whether that would be useful in production code.

reneme (approved):
Nice! Thanks for picking this up. :)

src/tests/test_bigint.cpp
randombit: Maybe? I can't think of cases currently where we predicate on a single random bit (vs say the even-y bit of a EC point). Somewhere in the PQ algos perhaps?

reneme: Not really. They all just pull a seed and then sample from KDFs or XOFs. I didn't find something with an ad-hoc search either.

PR #4416, opened 2024-10-27
Fix random test failure introduced by #4385

### LINKED ISSUE #4415: Weird ecc_randomized failure
https://github.com/randombit/botan/actions/runs/11540997855/job/32122471011

Seed 1802538CC7A462A1

```

  ECC randomized secp160k1 ran 79 tests in 62.22 msec 1 FAILED
  Failure 1: ECC randomized secp160k1 secp160k1 with error EC_Scalar::from_bigint input out of range (at /home/runner/work/botan/botan/source/src/tests/test_ec_group.cpp:160)
```

randombit:
Fixes #4415

### DIFF
--- a/src/tests/test_ec_group.cpp
+++ b/src/tests/test_ec_group.cpp
@@ -59,6 +59,12 @@ Botan::BigInt test_integer(Botan::RandomNumberGenerator& rng, size_t bits, const
       }
    }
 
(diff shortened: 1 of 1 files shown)


randombit:
Another simple response would be if we find ourselves with x == 0 return a completely random scalar instead.

PR #4418, opened 2024-10-29
Refactor and fixes for URI type

randombit:
Continuation of #3509

### DIFF
--- a/src/lib/utils/socket/uri.cpp
+++ b/src/lib/utils/socket/uri.cpp
@@ -1,5 +1,6 @@
 /*
 * (C) 2019 Nuno Goncalves <nunojpg@gmail.com>
+*     2023,2024 Jack Lloyd
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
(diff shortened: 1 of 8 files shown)


src/lib/utils/socket/uri.cpp
```diff
@@ -20,146 +21,133 @@
...
    sockaddr_storage inaddr;
    return !!inet_pton(AF_INET, ip_str.c_str(), &inaddr);
 }
 
-bool isIPv6(std::string_view ip) {
+bool is_ipv6(std::string_view ip) {
    std::string ip_str(ip);
    sockaddr_storage in6addr;
    return !!inet_pton(AF_INET6, ip_str.c_str(), &in6addr);
 }
```
reneme, line 47: @randombit Sorry for digging up this old PR, I'm reviewing this in retrospect for the BSI documents. And I'm tripping over the `!!inet_pton()`. First, I thought the double exclaimation mark must be a typo. But then I thought it might be intentional as a shorthand for the conversion to bool. 😲

However, `inet_pton` [can return -1 if `AF_INET[6]` is not supported](https://www.man7.org/linux/man-pages/man3/inet_pton.3.html):

> inet_pton() returns 1 on success (network address was successfully converted). 0 is returned if src does not contain a character string representing a valid network address in the specified address family. If af does not contain a valid address family, -1 is returned and [errno](https://www.man7.org/linux/man-pages/man3/errno.3.html) is set to EAFNOSUPPORT.

Now, I'm not familiar enough with this API, but if we assume that it would return -1 when a system is explicitly configured to not support IPv4 or IPv6, the double-! would essentially consider _any_ input as `true`. I feel that this should be handled explicitly. Or am I missing something?

PR #4421, opened 2024-10-30
Nightly Valgrind Matrix

### LINKED ISSUE #4396: Set up a nightly build matrix for valgrind-based SCA
I feel we should build a nightly test matrix that runs the valgrind suite on a wider spectrum of compilers, targets and optimization levels, in the hope that this increases the signal of such vulnerabilities. Or at least see if we can build regression tests based on our existing tools.

_Originally posted by @reneme in https://github.com/randombit/botan/pull/4395#issuecomment-2423824723_

reneme:
This sets up a configuration matrix for the nightly Valgrind run. In the hope to generate a stronger signal from the `CT::poison()` helpers. For that, I introduced two new build targets 'valgrind-ct' and 'valgrind-ct-full'. Currently, as their only difference compared to the non-ct targets, they don't enable the `--leak-check=full`, `--show-reachables` and `--track-origins` Valgrind features.

Explicitly note that the `clang -Os` variant would have been able to detect #4107, as evident by the `ct_selftest.py` test "clang_vs_bare_metal_ct_mask... ok" which [reproduced this issue](https://github.com/randombit/botan/blob/2ae990ff00bfe3ab54bebed66dd33a57c5f9af73/src/ct_selftest/ct_selftest.cpp#L131-L157). This particular self-test is skipped for all other configurations.

Below are some runtimes. Note that the compiliation results were cached in ccache (from previous test runs). So these runtimes are mostly comprised of runner setup, cached compilation and valgrind-hosted test suite runtime.

| | GCC | clang |
| --- | --- | --- |
| -O1 | 52min | 63min |
| -O2 | 54min | 59min  |
| -O3 | 49min | 50min |
| -Os | 63min 🔸| 59min |

🔸, means that it ran a reduced set of tests (`valgrind-ct`)

The valgrind configuration matrix also contains the existing 'valgrind-full' run on clang with -O3. This is the only configuration that explicitly aims to find memory bugs and enables the `--leak-check=full`, `--show-reachables` and `--track-origins` Valgrind features.

---

Here are, now outdated, runtimes (before switching off `--leak-check`, `--show-reachables`, `--track-origins`) and compiling with a cold ccache:

| | GCC | clang |
| --- | --- | --- |
| -O0 | ❌  | ❌   |
| -O1 | 1h35m | 1h43m |
| -O2 | 1h34m | 1h33m |
| -O3 | 1h30m | 1h36m |
| -Os | 2h16m 🔸| 1h50m |

❌, means that they [failed in a timeout after 6 hours](https://github.com/Rohde-Schwarz/botan/actions/runs/11589691033).
🔸, means that it timed out with `valgrind-full` but succeeded with `valgrind`

Closes #4396

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -149,6 +149,35 @@ jobs:
 
   valgrind:
     name: "valgrind"
+    strategy:
+      fail-fast: false
+
+      # Targets:
+      #   - valgrind-full:    all tests on valgrind, aiming to catch memory bugs
+      #   - valgrind:         reduced set of tests on valgrind, aiming to catch memory bugs
+      #   - valgrind-ct-full: all tests on valgrind, aiming to catch secret-dependent execution issues
+      #   - valgrind-ct:      reduced set of tests on valgrind, aiming to catch secret-dependent execution issues
+
+      matrix:
+        # Run a matrix of compiler and optimization flag combinations to maximize
+        # the signal of secret-dependent execution issues introduced by compilers.
+        compiler: ["clang", "gcc"]
+        cxxflags: ["-O1", "-O2", "-O3"]
+        target:   ["valgrind-ct-full"]
+
+        include:
+          - compiler: clang
+            cxxflags: ""              # default compilation flags
+            target:   "valgrind-full" # memory bug detection
+          - compiler: clang
+            cxxflags: "-Os"
+            # Clang's -Os generated binary is fast enough to run the full test suite.
+            target:   "valgrind-ct-full"
+          - compiler: gcc
+            cxxflags: "-Os"
+            # GCC with -Os generates a much slower binary, that won't finish
+            # before timing out on GH Actions, so we run a reduced set of tests.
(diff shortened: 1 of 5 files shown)


reneme:
> I think the no optimizations case is not really that useful; if someone using that in production they are insane since performance will be terrible.

I wanted to agree, but the -O0 did actually find something that may or may not be worth fixing: https://github.com/randombit/botan/pull/4421/commits/67964a89f4aad2f5d5c89df55c23bbd376152f4e

reneme:
> `valgrind-ct `(whatever) runs only specifically designated tests, and without the leak check/origin tracking flags, across a range of targets.

Given that the runtime is bearable (for a nightly job and excluding -O0), I am wondering whether it is really worth the effort of splitting these out. Perhaps to tweak the valgrind flags for the CT-matrix. But I'd put a question mark on the hand-picked tests, as we'll inevitably miss adding relevant things to this list in the future.

randombit (approved):
Looks good thanks

PR #4429, opened 2024-11-07
Enhance Side Channel resistance of TPM2 RSA Decryption Wrapper

atreiber94:
The TPM2 Wrapper of RSA decryption had a bug with incompatible types while setting `valid_mask`.

During the fix up we noticed that the rest of the function did not run in constant time and may present an oracle on padding failures. Thanks to @phwork for finding this.

This PR enhances the wrapper's side channel resistance in that regard.

Unfortunately, the ESAPI call [already introduces some side channel leaks](https://github.com/tpm2-software/tpm2-tss/blob/25a5172f6fea905e3967c5ce96a67d148ed71c95/src/tss2-esys/api/Esys_RSA_Decrypt.c#L329-L343) by allocating the result buffer only if the decryption succeeds. This is by specification ([Section 14.3.1](https://trustedcomputinggroup.org/wp-content/uploads/TPM-2.0-1.83-Part-3-Commands.pdf)), as the call to the underlying RSA decryption function returns an error code if unpadding fails.

@randombit opinions are welcome, we are not sure how big of an issue this is and if our enhancement makes much sense.

### DIFF
--- a/src/lib/prov/tpm2/tpm2_rsa/tpm2_rsa.cpp
+++ b/src/lib/prov/tpm2/tpm2_rsa/tpm2_rsa.cpp
@@ -340,13 +340,25 @@ class RSA_Decryption_Operation final : public PK_Ops::Decryption {
                                                                         &label,
                                                                         out_ptr(plaintext)));
 
-         valid_mask = CT::Mask<uint8_t>::is_equal(rc, TPM2_RC_SUCCESS).value();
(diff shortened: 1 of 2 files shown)


randombit (approved):
How unfortunate!

I don’t see this as that big an issue but it’s probably best if we at least do not introduce any additional signal beyond what is intrinsic to the TPM2 API itself, so I do think the change makes sense.

PR #4430, opened 2024-11-08
Feature: Support external ESYS_CONTEXT in TPM2

atreiber94:
(Don't be scared, the large diff is due to file renaming...)

Certain applications may want the flexibilty to do more than currently possible within `Botan::TPM2` and therefore may rely on an `ESYS_CONTEXT` that is not created via Botan's interface.

This results in 2 major use cases:

- Create a `TPM2::Context` from an `ESYS_CONTEXT` managed by the application
  This is useful to selectively benefit from wrapped functionality in Botan while maintaining the flexibility to use all of ESAPI
- Use ESAPI for everything but register Botans TPM crypto backend

Unfortunately there is some added complexity because of the different kinds of memory management involved between the different APIs. Externally provided `ESYS_CONTEXT` objects are not retained by `TPM2::Context`, and are therefore not finalized in its destructor.


randombit (approved):
Looks fine to me but @reneme has a lot more context here so I’d like his review if possible

reneme (changes requested):
I was involved in the development of this patch, pair-programming with @atreiber94. Nevertheless, we forgot to test this with `--disabled-modules=tpm2_crypto_backend` 😨. Needs a little `#ifdef` fairy dust, therefore.

reneme (approved):

Issue #4417, opened 2024-10-28
Systematically time all test suites


randombit:
Some but not all of the tests report how long they took. Fix this.

Also the time reported should be normalized to always print seconds, instead of in ms if fast but in seconds if > 1 s.

KaganCanSit:
Hello,
I would like to take part in the issue here, but I think I need a little help. I have made a small change to the content under the test folder so that the duration is only in seconds. (https://github.com/KaganCanSit/botan/commit/0182be711278c067e3f459d1c35672e57bd946cf) 

But I have not yet been able to organize to get the time information of all test cases. Obviously there is more than one time measurement event. There is a way by assigning the time value in the Test::Result object to the TestSummary constructor in test_reporter.cpp. But on the other hand, elapsed_time and m_start_time class member are returned by assigning them via the next_test_run() function.

What I want to ask is that time information is available in test.h and test.cpp. But the time information printed during the execution of the tests is edited by the next_test_run() and render() functions called from test_runner.cpp in test_repoter.cpp. There is also a todo in tests.cpp that the result_string() function should be moved to the StdoutReporter class.

For the process going forward, should the time measurement operations performed inside the Botan_Tests object be removed, moved to test_reporter.cpp based on the Single Responsibility principle, and measured and used globally through test_runner?

Instead of measuring a range of values in all test suites, measure it all process? (construct/destruct)

PR #4435, opened 2024-11-14
Test duration values ​​are now presented in seconds with six digits of precision. Tests without time measurements have been edited.

KaganCanSit:
Pull Request: Systematically Time All Test Suites (#4417) - Test duration values ​​are now presented in seconds with six digits of precision. Tests without time measurements have been edited.

Summary
* Enhanced Time Precision:
Adjusted the printing of time values to display with 6 digits of precision in seconds. This improvement ensures more accurate and consistent timing information across all tests.

* Added Time Measurement 
Implemented timing calls for tests that previously did not measure execution time. This update affected several files to achieve full coverage of time measurements across all test 
While considering where to set these timing calls, I explored alternatives like placing the time measurement in the Result object's constructor and destructor. However, given that the Result objects are stored in a vector and altering the existing structure might have unforeseen side effects, I opted to integrate the timing calls in a way that preserves the current architecture.

* Additional Considerations
During testing, I observed that time values are not always printed due to the if (m_ns_taken > 0) condition in the std::string Test::Result::result_string() function located in "test.cpp." After careful consideration, I decided to retain this condition, as it appears to function as expected and aligns with the intended logic.

Best regards

Note: I am aware that this change affects a lot of files and their content. However, I have tried to make the impact minimal while remaining true to the existing architecture. If you have any content that you would like to see edited, I would be happy to help you out as much as I can.

### DIFF
--- a/src/tests/test_aead.cpp
+++ b/src/tests/test_aead.cpp
@@ -32,6 +32,7 @@ class AEAD_Tests final : public Text_Based_Test {
          const bool is_siv = algo.find("/SIV") != std::string::npos;
 
          Test::Result result(algo);
+         result.start_timer();
 
          auto enc = Botan::AEAD_Mode::create(algo, Botan::Cipher_Dir::Encryption);
 
@@ -177,6 +178,7 @@ class AEAD_Tests final : public Text_Based_Test {
                                [&]() { enc->set_associated_data(ad.data(), ad.size()); });
          }
 
+         result.end_timer();
          return result;
       }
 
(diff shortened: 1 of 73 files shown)


KaganCanSit:
Even though I checked, I think there are some points I missed. I will review why the checks failed and edit to fix it.

randombit:
@KaganCanSit thank you but this isn't really at all how I planned on addressing this; the problem is precisely that right now the timing start/stop has to have manual calls inserted, and that is error prone and noisy.

KaganCanSit:
@randombit I thought about this too. But I didn't want to change the current code structure. If we can exchange small tips/ideas for this, I would like to work on it.

PR #4437, opened 2024-11-18
Split loading and generating ECC private key c'tors

reneme:
Currently, the ECC private keys provide two constructors: 

1. taking an `AlgorithmIdentifier` and key bits
2. taking an RNG, an `EC_Group` and an optional secret value as `BigInt`

The second option either generates a new key or loads the optional secret value within the provided domain. If a user wants to load an ECC private key for which they don't have an `AlgorithmIdentifier` handy, they have to use this second constructor and "hope" that they'll use it correctly. E.g. by passing a `Null_RNG` to at least get some error if used incorrectly.

This PR deprecates the second constructor and replaces it by two new ones:

1. taking an RNG and a domain, to perform a key generation
2. taking a domain and a secret value, to perform a key loading

Users that never call the now-deprecated constructor with the optional secret value won't ever see the deprecation. Their code will now just bind to the new "generating" constructor. This is achieved by making the secret value mandatory in the now-deprecated constructor.

### DIFF
--- a/src/examples/ecc_raw_private_key.cpp
+++ b/src/examples/ecc_raw_private_key.cpp
@@ -13,13 +13,12 @@ int main() {
    const auto private_scalar_bytes =
       Botan::hex_decode("D2AC61C35CAEE918E47B0BD5E61DA9B3A5C2964AB317647DEF6DFC042A06C829");
 
-   Botan::Null_RNG null_rng;
    const auto domain = Botan::EC_Group::from_name(curve_name);
    const auto private_scalar = Botan::BigInt(private_scalar_bytes);
 
    // This loads the private scalar into an ECDH_PrivateKey. Creating an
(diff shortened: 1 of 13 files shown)


reneme:
To be considered: When using the now-deprecated constructor we have access to an RNG. This allows us to do a random blinding when perform the multiplication with the base point, to get the associated public key. The new "key loading" constructor doesn't have an RNG and hence won't be able to use random blinding.

However, the same is true for the existing (and probably much more useful) constructor taking an `AlgorithmIdentifier`.

randombit:
Thanks! I had this on the docket re #4027

Re the removed RNG being used for blinding, indeed I think historically that is the reason for this (kind of weird in retrospect) combined generator-or-load constructor that took an RNG instance.

randombit (approved):

PR #4439, opened 2024-11-20
Add a .git-blame-ignore-revs file

reneme:
This cleans up `git blame` (also `tig blame` or VSCode's GitLens annotations) by automatically skipping commits that just reformatted code using clang-format. The name of the ignore-file isn't standard, but it seems to be the convention across the community.

~~Leaving this as a "Draft" for now. I personally [just learnt about this feature](https://www.michaelheap.com/git-ignore-rev/), and find it super useful (worthy of a new top-level file 😏)~~

To use this, you'll have to adapt your git config like so:

```bash
git config --local blame.ignoreRevsFile src/configs/git-blame-ignore-revs
git config --local blame.markIgnoredLines true
```

The first config enables the usage of the ignore file by default, the second one marks blame-lines with a '?' if they skipped one or more commits.

### DIFF
--- a/doc/dev_ref/contributing.rst
+++ b/doc/dev_ref/contributing.rst
@@ -117,6 +117,15 @@ needed. If your diff is less than roughly 100 lines, it should probably be a
 single commit. Only split commits as needed to help with review/understanding of
 the change.
 
+Occasionally we apply and commit updated clang-format rules to the code base. To
+avoid cluttering the ``git blame`` output with these intrusive commits, we
+maintain a list of them in ``src/configs/git-blame-ignore-revs``. To use it,
+either manually add ``--ignore-revs-file=`` to your ``git blame`` command, or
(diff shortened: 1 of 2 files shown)


randombit (changes requested):
Useful!

Given that git doesn't understand this by default, I disagree with your assessment re top level directory. Fine for `src/configs`.

Maybe also worth a mention in the contributing guide.

reneme:
> Given that git doesn't understand this by default, I disagree with your assessment re top level directory. Fine for `src/configs`.

In fact, its probably better to not configure this globally after all. It seems that (at least) `tig blame` is tripped up if other repositories don't provide this file but it is configured globally in `~/.gitconfig`. Moved to `src/configs` and updated the documentation accordingly.

> Maybe also worth a mention in the contributing guide.

Done.

Issue #4440, opened 2024-11-21
Cannot create an HSS-LMS key through CLI


zugzwang:
I am trying to generate a private HSS-LMS key to subsequently obtain the public key, using the `keygen` and `pkcs8` commands.

According to what is working with XMSS, the following should do it:

```
botan keygen --algo=HSS-LMS --params="HSS-LMS(SHA-256,HW(5,1),HW(5,1))" > mypriv.key
botan pkcs8 --pub-out mypriv.key > mypub.key
```

However I am consistently getting `Error: Invalid number of arguments in HSS_LMS_Params:src/lib/pubkey/hss_lms/hss.cpp`. Note that the chosen parameter set (H = 2, SHA-256, Winternitz w = 1) is taken verbatim from section 8.10.15 of [the Handbook](https://botan.randombit.net/handbook/botan.pdf).

```
root@26ada406c902:/test# botan version
3.7.0
root@26ada406c902:/test# botan keygen --algo=HSS-LMS --params="HSS-LMS(SHA-256,HW(5,1),HW(5,1))"
Error: Invalid number of arguments in HSS_LMS_Params:src/lib/pubkey/hss_lms/hss.cpp
root@26ada406c902:/test# botan keygen --algo=HSS-LMS --params="HSS-LMS(SHA-256,HW(5,8))"
Error: Invalid number of arguments in HSS_LMS_Params:src/lib/pubkey/hss_lms/hss.cpp
root@26ada406c902:/test# botan keygen --algo=HSS-LMS --params="HSS-LMS(WRONG_HASH,HW(5,8))"
Error: Invalid number of arguments in HSS_LMS_Params:src/lib/pubkey/hss_lms/hss.cpp
root@26ada406c902:/test# botan keygen --algo=HSS-LMS --params="HSS-LMS(SHAKE-256(256),HW(5,8))"
Error: Invalid number of arguments in HSS_LMS_Params:src/lib/pubkey/hss_lms/hss.cpp
```

randombit:
This works

```
./botan keygen --algo=HSS-LMS --params='SHA-256,HW(5,1),HW(5,1)'
```

We should accept it with or without `HSS_LMS(...)` though

randombit:
Fixed now on master but the workaround mentioned in earlier comment works for current release.

PR #4441, opened 2024-11-21
Improve parsing of HSS-LMS parameters

randombit:
GH #4440


FAlbertDev (approved):
Thanks, I fine with that 👍

PR #4445, opened 2024-11-25
FIX? Internal Compiler Error on SH-4

### LINKED ISSUE #4444: Nightly fails on SH-4 cross compile
```
during RTL pass: ira
  src/lib/pubkey/classic_mceliece/cmce_keys_internal.cpp: In static member function ‘static std::shared_ptr<Botan::Classic_McEliece_PublicKeyInternal> Botan::Classic_McEliece_PublicKeyInternal::create_from_private_key(const Botan::Classic_McEliece_PrivateKeyInternal&)’:
  src/lib/pubkey/classic_mceliece/cmce_keys_internal.cpp:152:1: internal compiler error: in elimination_costs_in_insn, at reload1.cc:3536
    152 | }
        | ^
  0x16c6ddb internal_error(char const*, ...)
    ???:0
  0x64d6b8 fancy_abort(char const*, int, char const*)
    ???:0
  0xd10d61 calculate_elim_costs_all_insns()
    ???:0
  0xbb98de ira_costs()
  ccache sh4-linux-gnu-g++ -fPIC -fvisibility=hidden -fstack-protector -pthread -std=c++20 -D_REENTRANT  -O3 -Wall -Wextra -Wpedantic -Wstrict-aliasing -Wcast-align -Wmissing-declarations -Wpointer-arith -Wcast-qual -Wzero-as-null-pointer-constant -Wnon-virtual-dtor -Wold-style-cast -Wsuggest-override -Wshadow -Wextra-semi -Werror -Wno-error=strict-overflow -Wno-error=zero-as-null-pointer-constant -Wno-error=non-virtual-dtor -Wno-error=maybe-uninitialized -Wno-error=stringop-overread -Wno-error=stringop-overflow -Wno-error=free-nonheap-object -Wno-error=restrict -DBOTAN_IS_BEING_BUILT  -I build/include/public -I build/include/internal  -isystem build/include/external -c src/lib/pubkey/curve448/curve448_scalar.cpp -o build/obj/lib/pubkey_curve448_scalar.o
    ???:0
  0xbb294f ira_build()
    ???:0
  Please submit a full bug report, with preprocessed source (by using -freport-bug).
  Please include the complete backtrace with any bug report.
```

See: https://github.com/randombit/botan/actions/runs/12002657334/job/33454969797

reneme:
Closes #4444

### DIFF
--- a/src/lib/utils/ct_utils.h
+++ b/src/lib/utils/ct_utils.h
@@ -244,6 +244,20 @@ template <unpoisonable T>
 /// @}
 
 /**
+* This forces a volatile read on the input value and returns the same value and
+* type as passed in. Useful to prevent the compiler from reasoning about the
+* value's range and suppress optimizations harmful to constant-time code.
+*
+* Note that this imposes a runtime overhead, as @p x will be physically stored
+* and loaded from memory.
+*/
+template <typename T>
+constexpr inline T volatile_read(T x) {
+   volatile T vx = x;
+   return vx;
+}
+
+/**
 * This function returns its argument, but (if called in a non-constexpr context)
 * attempts to prevent the compiler from reasoning about the value or the possible
 * range of values. Such optimizations have a way of breaking constant time code.
@@ -254,8 +268,9 @@ constexpr inline T value_barrier(T x) {
       /*
       * For compilers without inline asm, is there something else we can do?
       *
-      * For instance we could potentially launder the value through a
-      * `volatile T` or `volatile T*`. This would require some experimentation.
+      * Currently, we launder the value through a `volatile T`, as is done in
+      * Rust's subtlecrypto for instance:
+      * https://github.com/dsprenkels/subtlecrypto/blob/b4b070/src/lib.rs#L213-L228
       *
       * GCC has an attribute noipa which disables interprocedural analysis, which
       * might be useful here. However Clang does not currently support this attribute.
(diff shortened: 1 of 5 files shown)


randombit:
Thanks for looking into this. The valgrind finding is interesting...

reneme:
This isn't a complete fix as of now. There are more ICEs within the test-code of the bitvector class that I didn't find workarounds for so far. @randombit In your experience, could this indicate some undefined behavior looming within bitvector or would you rather guess that the code generation for the 30-year old processor might be buggy?

After all the bitvector is pretty dense template code, so I wouldn't rule out either.

reneme:
Turns out that the value_barrier tripped up the compiler. If I understand the ICE correctly,  something goes wrong in the code generation when allocating registers somehow.

We added a quick-and-dirty fallback to a volatile-read for targets without inline assembly and explicitly disabled SuperH as well. This fixes the compilation, but this certainly needs discussion.

randombit:
IME a GCC ICE is just that, a compiler bug. I’ve seen what GCC calls ICE-on-invalid (ie, it crashes when fed code that is not conforming/normally compilable) but I’ve never seen ICE where it had any relation to undefined behavior in the code that triggered it. Especially on the niche arch (these  days that means anything except x86, POWER, ARM, and maybe RISCV) a lot of these compiler targets only have 1-2 maintainers so it’s not so surprising.

It would be nice if we could reduce this to something reasonable (~couple hundred lines) at which point it’s reportable. Usually given a testcase GCC devs will fix ICEs.

src/lib/utils/ct_utils.h
```diff
@@ -265,11 +280,16 @@ constexpr inline T value_barrier(T x) {
...
+      * Note that the asm-based approach is disabled on SuperH because it caused
+      * miscompilation on GCC 13.2. See GH #4444 for more details.
       */
-#if defined(BOTAN_USE_GCC_INLINE_ASM) && !defined(BOTAN_HAS_SANITIZER_MEMORY)
+#if defined(BOTAN_USE_GCC_INLINE_ASM) && !defined(BOTAN_HAS_SANITIZER_MEMORY) && !defined(BOTAN_TARGET_ARCH_IS_SUPERH)
       asm("" : "+r"(x) : /* no input */);
-#endif
       return x;
+#else
+      return volatile_read(x);
```
randombit, line 291: This introduces it for MSVC (and others) as well which maybe is the “right” thing to do but I could see this causing a massive performance regression and AFAIK MSVC is not known to ever perform Clang-style optimizations that undo constant time code.

Likewise it introduces the volatile reads for the sanitizer build which doesn’t seem necessary; in general MemSan isn’t safe for production builds anyway.

I think we need A Couple More Macros (tm) here

* `BOTAN_COMPILER_USE_CT_BARRIER` - possibly set from the compiler info file? For GCC, Clang, XCode, maybe others.
* `BOTAN_COMPILER_CT_BARRIER_USE_ASM_BLOCK`
* `BOTAN_COMPILER_CT_BARRIER_USE_VOLATILE`

Maybe even toggleable at configure time? (`none`, `asm`, `volatile`) OK at this point I’m convinced this is worthwhile but I can take it over the holiday weekend …

reneme: For the record: #4447

randombit:
ok to close?

PR #4446, opened 2024-11-25
EC_AffinePoint API improvements

randombit:
This adds point negation and point addition.

Point addition is not particularly efficient since it converts the result immediately to an affine coordinate, as we do not currently expose projective points at all. That said it is more than sufficient for simple protocols that just need to add a few points together.

This also corrects the return result of EC_AffinePoint::mul_px_qy, which would previously crash if the resulting point was the identity element. Instead it should return nullopt.

Discussion in #4027

### DIFF
--- a/src/lib/pubkey/ec_group/ec_apoint.cpp
+++ b/src/lib/pubkey/ec_group/ec_apoint.cpp
@@ -111,12 +111,26 @@ EC_AffinePoint EC_AffinePoint::mul(const EC_Scalar& scalar, RandomNumberGenerato
    return EC_AffinePoint(inner().mul(scalar._inner(), rng, ws));
 }
 
-EC_AffinePoint EC_AffinePoint::mul_px_qy(const EC_AffinePoint& p,
-                                         const EC_Scalar& x,
-                                         const EC_AffinePoint& q,
-                                         const EC_Scalar& y,
-                                         RandomNumberGenerator& rng) {
+std::optional<EC_AffinePoint> EC_AffinePoint::mul_px_qy(const EC_AffinePoint& p,
+                                                        const EC_Scalar& x,
+                                                        const EC_AffinePoint& q,
+                                                        const EC_Scalar& y,
+                                                        RandomNumberGenerator& rng) {
    auto pt = p._inner().group()->mul_px_qy(p._inner(), x._inner(), q._inner(), y._inner(), rng);
+   if(pt) {
+      return EC_AffinePoint(std::move(pt));
+   } else {
+      return {};
+   }
+}
+
+EC_AffinePoint EC_AffinePoint::add(const EC_AffinePoint& q) const {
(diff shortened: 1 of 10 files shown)


src/lib/pubkey/ec_group/ec_apoint.cpp
```diff
@@ -111,12 +111,26 @@ EC_AffinePoint EC_AffinePoint::mul(const EC_Scalar& scalar, RandomNumberGenerato
...
+                                                        const EC_Scalar& x,
+                                                        const EC_AffinePoint& q,
+                                                        const EC_Scalar& y,
+                                                        RandomNumberGenerator& rng) {
    auto pt = p._inner().group()->mul_px_qy(p._inner(), x._inner(), q._inner(), y._inner(), rng);
+   if(pt) {
+      return EC_AffinePoint(std::move(pt));
+   } else {
+      return {};
+   }
```
reneme, line 124: Perhaps its worth adding a utility for this pattern at one point. Something along those lines:

```C++
template <typename InnerT, typename T>
   requires std::constructible_from<InnerT, T> && requires(T t) {
      { static_cast<bool>(t) } -> std::same_as<bool>;
   }
auto maybe(T&& v) {
   std::optional<InnerT> result = std::nullopt;
   if(v) {
      result.emplace(std::forward<T>(v));
   }
   return result;
}
```

... the implementation of this function would then look like that:

```suggestion
   return maybe<EC_AffinePoint>(p._inner().group()->mul_px_qy(p._inner(), x._inner(), q._inner(), y._inner(), rng));
```

... and it would also be RVO-friendlier.

One downside: similarly to `std::make_unique` (and friends), this needs the constructor of `InnerT` (here: `EC_AffinePoint(std::unique_ptr<EC_AffinePoint_Data>)`) to be public. 😞 This would currently not be the case for this particular example.

src/lib/pubkey/ec_group/ec_apoint.h
```diff
@@ -79,11 +81,27 @@ class BOTAN_UNSTABLE_API EC_AffinePoint final {
       /// Compute 2-ary multiscalar multiplication - p*x + q*y
       ///
       /// This operation runs in constant time with respect to p, x, q, and y
-      static EC_AffinePoint mul_px_qy(const EC_AffinePoint& p,
-                                      const EC_Scalar& x,
-                                      const EC_AffinePoint& q,
-                                      const EC_Scalar& y,
-                                      RandomNumberGenerator& rng);
+      ///
+      /// Returns nullopt if the result was the point at infinity
```
reneme, line 85:
```suggestion
      /// @returns nullopt if the result was the point at infinity
```

... Doxygen. :)

reneme (approved):

src/lib/pubkey/ec_group/ec_apoint.cpp
randombit: #4450

PR #4447, opened 2024-11-26
Add configure logic for choosing how to perform CT value barriers

randombit:
Related to #4444 and #4445


src/lib/utils/ct_utils.h
```diff
@@ -250,28 +250,25 @@ template <unpoisonable T>
 */
 template <typename T>
 constexpr inline T value_barrier(T x) {
```
reneme, line 252: Unrelated, but we might want to restrict the input types for this. Likely `std::integral`, or perhaps any trivial type? I'm not sure about the restrictions of the asm context, tbh.

reneme (approved):
Nice! Thanks for setting this up. There's definitely value in allowing to configure this.

I feel that a bit more documentation would be good, though. Not necessarily user-facing; perhaps just briefly discussing the idea of each alternative in a doxygen for the `value_barrier`.

src/lib/utils/ct_utils.h
randombit: TBH I'm not sure if that would work or not. It is however probably not sensible to ever do so, even if it worked - if you need to `value_barrier` some complex `T` probably instead `T` should have had some constant time logic which internally used `value_barrier` on the integer types which `T` contains.

configure.py
```diff
@@ -3276,6 +3303,10 @@ def validate_options(options, info_os, info_cc, available_module_policies):
     if options.with_pdf and not options.with_sphinx:
         raise UserError('Option --with-pdf requires --with-sphinx')
 
+    if options.ct_value_barrier_type:
+        if options.ct_value_barrier_type not in ['asm', 'volatile', 'none']:
+            raise UserError('Unknown setting "%s" for --ct-value-barrier-type' % (options.ct_value_barrier_type))
```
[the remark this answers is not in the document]
randombit: There is (`choice` type options) but there is other validation that happens besides that (eg using `--ct-value-barrier-type=asm` with a compiler that doesn't support GCC-style inline asm) so it seems better to me to have explicit checks + throwing `UserError` in all cases.

src/lib/utils/ct_utils.h
randombit: I added a restriction that `CT::value_barrier` must be a non-boolean unsigned integral type.

PR #4448, opened 2024-11-27
Refactor: Some minor things in the CMCE code

reneme:
This is spun off of #4445 as simplifications added in the hope of avoiding an ICE on GCC 13.2 targeting SuperH. The ICE was eventually fixed by #4447, but I think the refactorings are still worthwhile.


FAlbertDev (approved):
LGTM, thanks!

randombit (approved):

Issue #4451, opened 2024-12-01
Build of Botan 3.6.1/3.7.0 still failing with LNK1146 error


ronchristie52:
Hi all,

Just cloned the latest version of Botan and tried to build an x86 static library using the following configure line:

python3 configure.py --cpu=x86_32 --build-tool=ninja --cc=msvc --disable-shared-library --msvc-runtime=MT --libdir=c:\botan\Botan-3.6.1.1\Release
ninja
ninja check
ninja install

The build fails with:

LINK : fatal error LNK1146: no argument specified with option '/MT'

when attempting to build botan-cli.exe and botan-test.exe.

When these builds fail, then ninja stops the process and it never gets to the install step.

randombit:
@ronchristie52 Can you try the patch in #4452?

randombit:
Fixed by #4452

PR #4200, opened 2024-07-11
Reflect CXXFLAGS also into the linker invocation

randombit:
GH #4196

PR #4452, opened 2024-12-01
Don't implicitly/always pass CXXFLAGS into LDFLAGS

randombit:
This was done to handle LTO (#4196 #4200) but causes problems especially for MSVC which in some (unclear) circumstances treats unknown flags to the linker as a hard error (#4451). Instead only pass CXXFLAGS into LDFLAGS when an extra option `--lto-cxxflags-to-ldflags` is provided to opt into this behavior.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -379,6 +379,9 @@ def process_command_line(args):
     target_group.add_option('--extra-cxxflags', metavar='FLAGS', default=[], action='append',
                             help='set extra compiler flags')
 
+    target_group.add_option('--lto-cxxflags-to-ldflags', default=False, action='store_true',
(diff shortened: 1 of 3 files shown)


randombit:
@ronchristie52 Is there something strange about your system trust store? Possibly some kind of enterprise wide configuration/policy that reduced the default set? This test looks for any of [5 publicly trusted CA certificates](https://github.com/randombit/botan/blob/master/src/tests/test_certstor_utils.cpp#L43-L79). It is the case that different systems sometimes (for reasons we're not entirely sure about) do miss 1 or 2 of them, which is why the test only fails if all 5 are not found.

Issue #4449, opened 2024-11-27
Need more Flexibility in KDFs SP.800-56C and SP.800-108


reneme:
Currently, there are some hard-coded assumptions. Let's take,

```C++
auto kdf = KDF::create("SP800-56C(SHA-256)");
```

... that will currently create a KDF that uses HMAC(SHA-256) for the extraction (which is fine) and hard-codes SP800-108-Feedback for the expansion. As far as I can see, there's no way to customize the expansion step (especially now that the concrete KDF-headers are internal). Or am I missing something?

Also, SP800-108 currently hard-codes the bitlengths for the counter and the encoding of the requested key lengths to 32 bits.

As it stands, customization of the mentioned hard-coded decisions should probably all be encodable using the algo-spec. Also, I believe, this can be done in a backward-compatilbe way using the existing bits as defaults. Albeit I foresee this to become somewhat cumbersome. Maybe #3275 would be useful here.

The fully-customized algo spec would probably look like this:

```
SP800-56C(SHA-256,SP800-108-Counter(SHA-256,8,16))
```

... with 8 and 16 being the bitlengths for the counter (aka `r`) and the key length `L` respectively.

randombit:
:sob: why could NIST not just standardize HKDF and be done with it

[Fine to extend this to handle these cases if you need them, I just greatly dislike how NIST keeps coming with new KDFs - we already implement ***8*** KDFs from various NIST SPs and they all have a zillion options]

reneme:
Yep. Try adding BSI to the mix. 😏

reneme:
As mentioned in #4551, it is a lost cause to try and reflect all degrees of freedom those specifications leave to the implementer in a generic enough way to be useful. Nevertheless, I think that allowing applications to configure the encoding widths of "counter" and "output length" is a nice improvement. FWIW, it makes Botan's SP.800-108 implementation usable for our application.

Apart from that, I'm currently not planning to implement the proposed extension of SP.800-56C "two-step" because Botan's current KDF API doesn't provide the required flexibility. Namely, this would need yet another domain separation parameter. One for the Extraction (`dK = HMAC(IKM, salt)`), and two more for the expansion (`SP.800-108-Counter(dK, label, context)`). The current `KDF::derive_key()` only takes two domain separators aside from the IKM.

Frankly, I feel, its simply a lost cause to hide this plethora of degrees of freedom behind an abstract API while keeping it usable. :(

PR #4455, opened 2024-12-02
Modernize Key Derivation Functions

reneme:
Introduces a private customization point in the `KDF` base class. Passes the parameters from the abstract (user-facing) interface as `std::span` wrappers. And, uses the latest convenience tools in the implementations of the KDFs.

Also, this introduces `KDF::derive_key<length>(secret, salt, label)` to conveniently obtain the key material as a std::array. And deprecates methods that take raw pointers in favor of the std::span/std::string_view based overloads.

This is a first step towards #4449.

### DIFF
--- a/src/lib/kdf/hkdf/hkdf.cpp
+++ b/src/lib/kdf/hkdf/hkdf.cpp
@@ -2,6 +2,7 @@
 * HKDF
 * (C) 2013,2015,2017 Jack Lloyd
 * (C) 2016 René Korthaus, Rohde & Schwarz Cybersecurity
+* (C) 2024 René Meusel, Rohde & Schwarz Cybersecurity
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
@@ -11,6 +12,7 @@
 #include <botan/exceptn.h>
 #include <botan/internal/fmt.h>
 #include <botan/internal/loadstor.h>
+#include <botan/internal/stl_util.h>
 
 namespace Botan {
 
@@ -22,20 +24,16 @@ std::string HKDF::name() const {
    return fmt("HKDF({})", m_prf->name());
 }
 
-void HKDF::kdf(uint8_t key[],
-               size_t key_len,
-               const uint8_t secret[],
-               size_t secret_len,
-               const uint8_t salt[],
-               size_t salt_len,
-               const uint8_t label[],
-               size_t label_len) const {
+void HKDF::perform_kdf(std::span<uint8_t> key,
+                       std::span<const uint8_t> secret,
+                       std::span<const uint8_t> salt,
+                       std::span<const uint8_t> label) const {
    HKDF_Extract extract(m_prf->new_object());
    HKDF_Expand expand(m_prf->new_object());
    secure_vector<uint8_t> prk(m_prf->output_length());
 
-   extract.kdf(prk.data(), prk.size(), secret, secret_len, salt, salt_len, nullptr, 0);
-   expand.kdf(key, key_len, prk.data(), prk.size(), nullptr, 0, label, label_len);
+   extract.derive_key(prk, secret, salt, {});
+   expand.derive_key(key, prk, {}, label);
 }
 
 std::unique_ptr<KDF> HKDF_Extract::new_object() const {
@@ -46,42 +44,31 @@ std::string HKDF_Extract::name() const {
    return fmt("HKDF-Extract({})", m_prf->name());
 }
 
-void HKDF_Extract::kdf(uint8_t key[],
-                       size_t key_len,
-                       const uint8_t secret[],
-                       size_t secret_len,
-                       const uint8_t salt[],
-                       size_t salt_len,
-                       const uint8_t /*label*/[],
-                       size_t label_len) const {
-   if(key_len == 0) {
+void HKDF_Extract::perform_kdf(std::span<uint8_t> key,
+                               std::span<const uint8_t> secret,
+                               std::span<const uint8_t> salt,
+                               std::span<const uint8_t> label) const {
+   const size_t prf_output_len = m_prf->output_length();
+   BOTAN_ARG_CHECK(key.size() <= prf_output_len, "HKDF-Extract maximum output length exceeeded");
+   BOTAN_ARG_CHECK(label.empty(), "HKDF-Extract does not support a label input");
+
+   if(key.empty()) {
       return;
    }
 
-   const size_t prf_output_len = m_prf->output_length();
-
-   if(key_len > prf_output_len) {
-      throw Invalid_Argument("HKDF-Extract maximum output length exceeeded");
-   }
-
-   if(label_len > 0) {
-      throw Invalid_Argument("HKDF-Extract does not support a label input");
-   }
-
-   if(salt_len == 0) {
+   if(salt.empty()) {
       m_prf->set_key(std::vector<uint8_t>(prf_output_len));
    } else {
-      m_prf->set_key(salt, salt_len);
+      m_prf->set_key(salt);
    }
 
-   m_prf->update(secret, secret_len);
+   m_prf->update(secret);
 
-   if(key_len == prf_output_len) {
+   if(key.size() == prf_output_len) {
       m_prf->final(key);
    } else {
-      secure_vector<uint8_t> prk;
-      m_prf->final(prk);
-      copy_mem(&key[0], prk.data(), key_len);
+      const auto prk = m_prf->final();
+      copy_mem(key, std::span{prk}.first(key.size()));
    }
 }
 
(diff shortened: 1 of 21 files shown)


src/lib/kdf/sp800_108/sp800_108.cpp
```diff
@@ -22,51 +23,37 @@ std::unique_ptr<KDF> SP800_108_Counter::new_object() const {
...
-   const uint32_t length = static_cast<uint32_t>(key_len * 8);
+   const auto length = static_cast<uint32_t>(key.size() * 8);
+   const auto prf_len = m_prf->output_length();
 
-   uint8_t* p = key;
-   uint32_t counter = 1;
-   uint8_t be_len[4] = {0};
-   secure_vector<uint8_t> tmp;
+   const uint64_t blocks_required = (key.size() + prf_len - 1) / prf_len;
+   BOTAN_ARG_CHECK(blocks_required <= 0xFFFFFFFF, "SP800_108_Counter output size too large");
```
reneme, line 38: Given that the counter starts at 1, this should probably be:

```suggestion
   BOTAN_ARG_CHECK(blocks_required < 0xFFFFFFFE, "SP800_108_Counter output size too large");
```

src/lib/kdf/sp800_108/sp800_108.cpp
```diff
@@ -78,54 +65,42 @@ std::unique_ptr<KDF> SP800_108_Feedback::new_object() const {
...
-   secure_vector<uint8_t> ctx(salt + iv_len, salt + salt_len);
+   const auto length = static_cast<uint32_t>(key.size() * 8);
+   const auto prf_len = m_prf->output_length();
+   const auto iv_len = (salt.size() >= prf_len ? prf_len : 0);
+   constexpr uint8_t delim = 0;
 
-   store_be(length, be_len);
-   m_prf->set_key(secret, secret_len);
+   const uint64_t blocks_required = (key.size() + prf_len - 1) / prf_len;
+   BOTAN_ARG_CHECK(blocks_required <= 0xFFFFFFFF, "SP800_108_Feedback output size too large");
```
reneme, line 82:
```suggestion
   BOTAN_ARG_CHECK(blocks_required < 0xFFFFFFFE, "SP800_108_Feedback output size too large");
```

src/lib/kdf/sp800_108/sp800_108.cpp
```diff
@@ -137,63 +112,46 @@ std::unique_ptr<KDF> SP800_108_Pipeline::new_object() const {
...
-   uint8_t be_len[4] = {0};
-   secure_vector<uint8_t> ai, ki;
+   const auto length = static_cast<uint32_t>(key.size() * 8);
+   const auto prf_len = m_prf->output_length();
+   constexpr uint8_t delim = 0;
 
-   store_be(length, be_len);
-   m_prf->set_key(secret, secret_len);
+   const uint64_t blocks_required = (key.size() + prf_len - 1) / prf_len;
+   BOTAN_ARG_CHECK(blocks_required <= 0xFFFFFFFF, "SP800_108_Feedback output size too large");
```
reneme, line 128:
```suggestion
   BOTAN_ARG_CHECK(blocks_required < 0xFFFFFFFE, "SP800_108_Feedback output size too large");
```

src/lib/kdf/prf_x942/prf_x942.cpp
```diff
@@ -20,48 +21,35 @@ namespace {
...
-
-   in.reserve(salt_len + label_len);
-   in += std::make_pair(label, label_len);
-   in += std::make_pair(salt, salt_len);
-
-   while(offset != key_len && counter) {
-      hash->update(secret, secret_len);
+   BufferStuffer k(key);
+   for(uint32_t counter = 1; !k.full() && counter; ++counter) {
+      BOTAN_ASSERT_NOMSG(counter != 0);  // overflow check
```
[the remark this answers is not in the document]
reneme: The `BOTAN_ARG_CHECK(blocks_required < 0xFFFFFFFF, "X942_PRF maximum output length exceeeded");` won't allow to access this loop anyway, if we request so much data. Let alone: A KDF producing 4GiB of output sounds like someone is pushing it pretty hard anyway... 😏

... I removed the `counter` from the loop condition. Good catch, though!

reneme:
@randombit Does that mean you're fine with this change? 😅 

image

src/lib/kdf/hkdf/hkdf.cpp
```diff
@@ -46,42 +44,31 @@ std::string HKDF_Extract::name() const {
...
-                       const uint8_t salt[],
-                       size_t salt_len,
-                       const uint8_t /*label*/[],
-                       size_t label_len) const {
-   if(key_len == 0) {
+void HKDF_Extract::perform_kdf(std::span<uint8_t> key,
+                               std::span<const uint8_t> secret,
+                               std::span<const uint8_t> salt,
+                               std::span<const uint8_t> label) const {
+   if(key.empty()) {
```
randombit, line 51: We should probably even still check that the label was empty

src/lib/kdf/hkdf/hkdf.cpp
```diff
@@ -93,75 +80,54 @@ std::string HKDF_Expand::name() const {
...
-   prefix[1] = get_byte<1>(length16);
-   prefix[2] = static_cast<uint8_t>(label.size());
-
-   copy_mem(prefix.data() + 3, cast_char_ptr_to_uint8(label.data()), label.size());
-
-   prefix[3 + label.size()] = static_cast<uint8_t>(hash_val_len);
+   const auto prefix = concat<std::vector<uint8_t>>(store_be(static_cast<uint16_t>(length)),
+                                                    store_be(static_cast<uint8_t>(label.size())),
+                                                    std::span{cast_char_ptr_to_uint8(label.data()), label.size()},
+                                                    store_be(static_cast<uint8_t>(hash_val.size())));
```
randombit, line 123: This is *probably* better overall than what's currently in master, but also I hate it :)

I think what I want for things like this is some kind of "binary format string" eg

> const auto prefix = format_binary("%u16be %u8bytes %u8", length, label, hash_val);

(not exactly this, those format string choices suck and are just for exposition; there is probably a lot of prior art here)

Anyway no change required here.

src/lib/kdf/kdf1_iso18033/kdf1_iso18033.cpp
```diff
@@ -9,41 +10,33 @@
...
+                             std::span<const uint8_t> secret,
+                             std::span<const uint8_t> salt,
+                             std::span<const uint8_t> label) const {
+   if(key.empty()) {
       return;
    }
 
-   const size_t blocks_required = key_len / m_hash->output_length();
+   const size_t blocks_required = key.size() / m_hash->output_length();
+   BOTAN_ARG_CHECK(blocks_required < 0xFFFFFFFF, "KDF1-18033 maximum output length exceeeded");
```
randombit, line 26: Shouldn't this be `blocks_required < 0xFFFFFFFE`?

src/lib/kdf/kdf2/kdf2.cpp
```diff
@@ -20,41 +22,31 @@ std::unique_ptr<KDF> KDF2::new_object() const {
...
+                       std::span<const uint8_t> secret,
+                       std::span<const uint8_t> salt,
+                       std::span<const uint8_t> label) const {
+   if(key.empty()) {
       return;
    }
 
-   const size_t blocks_required = key_len / m_hash->output_length();
+   const size_t blocks_required = key.size() / m_hash->output_length();
+   BOTAN_ARG_CHECK(blocks_required < 0xFFFFFFFF, "KDF2 maximum output length exceeeded");
```
randombit, line 34: `..FE`?

src/lib/kdf/prf_tls/prf_tls.cpp
```diff
@@ -58,21 +55,12 @@ std::unique_ptr<KDF> TLS_12_PRF::new_object() const {
...
-   msg.reserve(label_len + salt_len);
-   msg += std::make_pair(label, label_len);
-   msg += std::make_pair(salt, salt_len);
-
-   P_hash(key, key_len, *m_mac, secret, secret_len, msg.data(), msg.size());
+void TLS_12_PRF::perform_kdf(std::span<uint8_t> key,
+                             std::span<const uint8_t> secret,
+                             std::span<const uint8_t> salt,
+                             std::span<const uint8_t> label) const {
+   const auto msg = concat<secure_vector<uint8_t>>(label, salt);
```
randombit, line 62: No real reason for `secure_vector` here, label is a constant and the salts are in the public hello messages

randombit: Alternately could amend `P_hash` to take 2 arguments for these values

src/lib/kdf/prf_x942/prf_x942.cpp
```diff
@@ -20,48 +22,35 @@ namespace {
...
    }
 
-   const size_t blocks_required = key_len / 20;  // Fixed to use SHA-1
-
-   if(blocks_required >= 0xFFFFFFFE) {
-      throw Invalid_Argument("X942_PRF maximum output length exceeeded");
-   }
+   constexpr size_t sha1_output_bytes = 20;  // Fixed to use SHA-1
+   const size_t blocks_required = key.size() / sha1_output_bytes;
+   BOTAN_ARG_CHECK(blocks_required < 0xFFFFFFFF, "X942_PRF maximum output length exceeeded");
```
randombit, line 44: FE?

randombit (commented):
Looks really good. Had a few comments and also some definite confusion going on with the block count checks which occurred in several instances (..FF vs ..FE). Not sure if the original check was wrong, your change is wrong, or if they are equivalent and I'm mentally off by one.

src/lib/kdf/hkdf/hkdf.cpp
reneme: I do, in the `BOTAN_ARG_CHECK` below. That replaced the previous explicit `if(label_len > 0) throw Invalid_Argument`. Or am I missing something?

src/lib/kdf/hkdf/hkdf.cpp
reneme: Agree. That would be really helpful. Perhaps even in the public API (at some point). I'd love to have something like that on application level as well. :)

src/lib/kdf/kdf1_iso18033/kdf1_iso18033.cpp
reneme: Judging from just the implementation, I would argue that the check currently in master is too tight. Each block is indexed using a 32-bit counter starting at 0. So, we can at most produce 2^32 blocks before the output stream repeats.

On a 64-bit system (for `size_t` to be big enough to take `block_length * 2^32` bytes) this should even be fine with `blocks_required <= 0xFFFFFFFF`. The hashing loop does not use the counter as abort condition, hence it is technically okay if it wraps after the last iteration.

(Acknowledging that this is a completely theoretical discussion and that I didn't consult the spec whether they actually define the upper bound to be two blocks lower explicitly.)

src/lib/kdf/kdf2/kdf2.cpp
reneme: Same discussion as above. 32-bit counter initialized at 1 this time. So 2^32-1 iterations should be safe (theoretically).

src/lib/kdf/prf_x942/prf_x942.cpp
reneme: 32-bit counter, initialized at 1. We should (theoretically) be able to generate 2^32-1 unique blocks.

src/lib/kdf/prf_tls/prf_tls.cpp
reneme: I removed `P_hash` entirely and moved its implementation straight into `perform_kdf`.

reneme (commented):
As detailed [here](https://github.com/randombit/botan/pull/4455#discussion_r1908306940), you were right to doubt the checked theoretical limits of the KDF outputs. The variable name `blocks_required` that is used in several KDF implementations turned out to be misleading. It contains the result of an integer devision that may be short-by-one to the actual result. This is now cleaned up using bit_ops.h's `ceil_division()`.

Additionally, I opportunistically avoided an allocation if the desired output length is hash-block aligned. This produced a bit of duplicated code that could potentially move into `Hash_Function`; see the inline comment below.

src/lib/kdf/hkdf/hkdf.cpp
randombit: My point was that here if the key is empty we return immediately, prior to checking the label. So a request for HKDF of length zero with label "Foo" returns an empty string, even though HKDF isn't defined for non-empty labels.

src/lib/kdf/prf_tls/prf_tls.cpp
randombit: > I removed P_hash entirely 

Oh right that function only existed so that the TLS 1.0 and TLS 1.2 KDFs could share code, no need for it anymore now that TLS 1.0 is dead.

src/lib/kdf/kdf1_iso18033/kdf1_iso18033.cpp
```diff
@@ -1,49 +1,51 @@
...
+      m_hash->update(salt);
+
+      // Write straight into the output buffer, except if the hash output needs
+      // a truncation in the final iteration.
+      if(k.remaining_capacity() >= hash_output_length) {
+         m_hash->final(k.next(hash_output_length));
+      } else {
+         const auto h = m_hash->final();
+         k.append(std::span{h}.first(k.remaining_capacity()));
+      }
```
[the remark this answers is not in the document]
randombit: Yeah I'd be open to that - it comes up quite a bit internally, does come up at least occasionally for applications, and isn't onerous to maintain. And almost all hashes could implement this in a more efficient way.

We *could* even define is so `final(short_span)` just implicitly truncates, which would lead to some spiffy KDF code, but is probably a huge footgun.

src/lib/kdf/kdf.h
```diff
@@ -62,14 +63,17 @@ class BOTAN_PUBLIC_API(2, 0) KDF {
...
+      BOTAN_DEPRECATED("Use KDF::derive_key")
+      void kdf(uint8_t key[],
+               size_t key_len,
+               const uint8_t secret[],
+               size_t secret_len,
+               const uint8_t salt[],
+               size_t salt_len,
+               const uint8_t label[],
+               size_t label_len) const {
+         perform_kdf({key, key_len}, {secret, secret_len}, {salt, salt_len}, {label, label_len});
```
randombit, line 75: nit - when deprecating a function because it's supplanted by something else, I like to have the inlined implementation act as documentation of sorts of exactly how one should translate the call. Here we deprecate it saying use `derive_key`, but the implementation calls the internal `perform_kdf` instead, which isn't something a user can immediately copy and paste.

randombit (approved):
Thanks for adding the comments regarding the counter situation

src/lib/kdf/kdf1_iso18033/kdf1_iso18033.cpp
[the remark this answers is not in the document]
reneme: Yeah, I thought about suggesting that `final(short_span)` but would be concerned for the same reason. I'll open a ticket later.

Another cool thing could be `final<42>() -> std::array<uint8_t, 42>`.

src/lib/kdf/kdf.h
reneme: Sounds like a good convention to me, thanks. I'll keep that in mind.

src/lib/kdf/hkdf/hkdf.cpp
reneme: Fair. I moved the arg checks up. Also in HKDF expand.

PR #4457, opened 2024-12-03
Split os_utils into a submodule

randombit:


src/lib/rng/rng.cpp
```diff
@@ -8,36 +8,31 @@
...
-      store_le(OS::get_system_timestamp_ns(), s_additional_input.data());
-      s_additional_input = s_additional_input.subspan(s_sys_ts);
+      system_rng().randomize(additional_input);
+#endif
 
-      store_le(OS::get_process_id(), s_additional_input.data());
+#if defined(BOTAN_HAS_OS_UTILS)
+      store_le(&additional_input[0], OS::get_high_resolution_clock());
+      store_le(&additional_input[8], OS::get_system_timestamp_ns());
+      store_le(&additional_input[16], OS::get_process_id());
```
reneme, line 35: Perhaps avoid the pointers?

```suggestion
      store_le(std::span{additional_input}.subspan<0, 8>(), OS::get_high_resolution_clock());
      store_le(std::span{additional_input}.subspan<8, 8>(), OS::get_system_timestamp_ns());
      store_le(std::span{additional_input}.subspan<16, 4>(), OS::get_process_id());
```

Technically, also the following could work, but currently the multi-param `store_any` has the restriction that all params have to be of the same integer type. If we'd lift this, we could even do that:

```C++
      store_le(std::span{additional_input}.first<20>(),
               OS::get_high_resolution_clock(),
               OS::get_system_timestamp_ns(),
               OS::get_process_id());
```

src/lib/rng/rng.cpp
```diff
@@ -8,36 +8,31 @@
...
 #endif
 
       this->fill_bytes_with_input(output, additional_input);
```
reneme, line 38: For the record: This will typically* overwrite the majority of the random bits with the high-resolution timestamps below. Not sure its worth the extra complexity to handle that more efficiently, but perhaps like this:

```C++
      std::array<uint8_t, 32> additional_input = {0};

#if defined(BOTAN_HAS_OS_UTILS)
      store_le(std::span{additional_input}.subspan<0, 8>(), OS::get_high_resolution_clock());
      store_le(std::span{additional_input}.subspan<8, 8>(), OS::get_system_timestamp_ns());
      store_le(std::span{additional_input}.subspan<16, 4>(), OS::get_process_id());
      auto randomize = std::span{additional_input}.subspan<20>();
#else
      auto randomize = std::span{additional_input};
#endif

      system_rng().randomize(randomize);
#else
      BOTAN_UNUSED(randomize);
#endif

      this->fill_bytes_with_input(output, additional_input);
```

\* when both system rng and OS utils are available

reneme (approved):
Two nits, otherwise LGTM. Modulo the CI issues. :)

src/lib/rng/rng.cpp
randombit: To me it seemed easier to just accept this and always perform a longer read but it is a bit subtle and lacking at least a comment might strike someone as a bug/unintentional. Your approach is probably better here.

PR #4459, opened 2024-12-03
Remove workaround for GCC miscompilation

randombit:
It seems to not be required anymore.


reneme:
The GCC amalgamation build on linux seems to disagree. :(

randombit:
:sob:

PR #4460, opened 2024-12-03
Brush up various example programs

reneme:
This is in response to #4453.

Consistently use `Botan::secure_vector` for sensitive information. Use std::span based APIs where suitable. Replace std::string w/ string_view in some places. Showcase some newly added convenience methods where they make sense (e.g. `rng.random_array<32>()`).

Perhaps a bit controversial: I used `auto` for some `hex_decode[_locked]` results. This somewhat hides the distinction between Botan::secure_vector and std::vector, which may or may not be good for the educational goal of those examples. Though, it avoids overly long lines (or multi-line statements). Opinions welcome.

### DIFF
--- a/src/examples/aes.cpp
+++ b/src/examples/aes.cpp
@@ -4,19 +4,19 @@
 #include <iostream>
 
 int main() {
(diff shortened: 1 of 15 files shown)


randombit (approved):
Thanks!

`auto var = hex_decode_locked(...)` seems clear enough to me

PR #4464, opened 2024-12-04
Fix warnings of some MSVC versions

FAlbertDev:
For example, with MSVC 19.39, some warnings are triggered, like:
```log
warning C4244: '=': conversion from 'const _Ty' to '_Ty', possible loss of data
```
While some contexts for this warning seem pretty dumb, fixing them won't hurt, I guess.


randombit (approved):

FAlbertDev:
Wow, thanks for being faster than the CI 🚀 ❤️

PR #4469, opened 2024-12-06
Modernize GHASH: allowing GCM to accept input that is not block-aligned

reneme:
Most notably, this lifts the limitation that GHASH::update() requires block-aligned data input by transparently caching non-aligned data using the AlignmentBuffer helper. With that, the GCM mode of operation can stream-process data that is not block aligned.

The GMAC used to contain an AlignmentBuffer to handle the same limitation described above. This was removed, as GHASH now handles non-aligned data transparently.

Finally, this cleans up the state management of GHASH as much as possible and removes a few public interfaces.

### DIFF
--- a/src/lib/utils/ghash/ghash.cpp
+++ b/src/lib/utils/ghash/ghash.cpp
@@ -2,6 +2,7 @@
 * GCM GHASH
 * (C) 2013,2015,2017 Jack Lloyd
 * (C) 2016 Daniel Neus, Rohde & Schwarz Cybersecurity
+* (C) 2024 René Meusel, Rohde & Schwarz Cybersecurity
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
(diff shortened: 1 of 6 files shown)


src/lib/utils/ghash/ghash.cpp
```diff
@@ -33,7 +33,9 @@ std::string GHASH::provider() const {
    return "base";
 }
 
-void GHASH::ghash_multiply(secure_vector<uint8_t>& x, std::span<const uint8_t> input, size_t blocks) {
+void GHASH::ghash_multiply(std::span<uint8_t, GCM_BS> x, std::span<const uint8_t> input, size_t blocks) {
+   BOTAN_DEBUG_ASSERT(input.size() % GCM_BS == 0);
```
randombit, line 37: This maybe should even be a prod assert, it should be fast

randombit (approved):
Looks good, and a nice cleanup! Thanks

PR #4477, opened 2024-12-15
Remove usage of deprecated asio API

q66:
i'm currently working on updating boost to 1.87 in chimera linux, which involves rebuild of all boost revdeps; i found that botan was newly failing to build due to asio 1.87 dropping some stuff that was previously deprecated, and finally got it to build/pass tests so here's the pull request

PR #4484, opened 2024-12-18
Resolve Boost deprecations

randombit:
Followup of #4477

### DIFF
--- a/src/lib/utils/socket/socket.cpp
+++ b/src/lib/utils/socket/socket.cpp
@@ -46,20 +46,19 @@ class Asio_Socket final : public OS::Socket {
    public:
       Asio_Socket(std::string_view hostname, std::string_view service, std::chrono::milliseconds timeout) :
             m_timeout(timeout), m_timer(m_io), m_tcp(m_io) {
-         m_timer.expires_from_now(m_timeout);
+         m_timer.expires_after(m_timeout);
          check_timeout();
 
          boost::asio::ip::tcp::resolver resolver(m_io);
-         boost::asio::ip::tcp::resolver::query query(std::string{hostname}, std::string{service});
-         boost::asio::ip::tcp::resolver::iterator dns_iter = resolver.resolve(query);
+         boost::asio::ip::tcp::resolver::results_type dns_iter = resolver.resolve(std::string{hostname}, std::string{service});
 
          boost::system::error_code ec = boost::asio::error::would_block;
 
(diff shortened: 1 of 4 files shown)


src/lib/utils/socket/socket.cpp
```diff
@@ -46,20 +46,19 @@ class Asio_Socket final : public OS::Socket {
...
             m_timeout(timeout), m_timer(m_io), m_tcp(m_io) {
-         m_timer.expires_from_now(m_timeout);
+         m_timer.expires_after(m_timeout);
          check_timeout();
 
          boost::asio::ip::tcp::resolver resolver(m_io);
-         boost::asio::ip::tcp::resolver::query query(std::string{hostname}, std::string{service});
-         boost::asio::ip::tcp::resolver::iterator dns_iter = resolver.resolve(query);
+         boost::asio::ip::tcp::resolver::results_type dns_iter =
+            resolver.resolve(std::string{hostname}, std::string{service});
```
reneme, line 54: Nit: This could now be calld `dns_result`, `endpoints` or something similar, as it is not technically an iterator anymore.

src/lib/utils/socket/socket.cpp
```diff
@@ -46,20 +46,19 @@ class Asio_Socket final : public OS::Socket {
...
 
          boost::system::error_code ec = boost::asio::error::would_block;
 
-         auto connect_cb = [&ec](const boost::system::error_code& e, const boost::asio::ip::tcp::resolver::iterator&) {
-            ec = e;
-         };
+         auto connect_cb = [&ec](const boost::system::error_code& e,
+                                 const boost::asio::ip::tcp::resolver::results_type::iterator&) { ec = e; };
```
reneme, line 59: Another nit: If we don't care about the second parameter anyway, we might as well `auto` it:

```suggestion
         auto connect_cb = [&ec](const boost::system::error_code& e,
                                 const auto&) { ec = e; };
```

reneme (approved):
Thanks! Also, seems to be somewhat urgent now, because `brew` on macOS seems to have updated boost as well which breaks the build on master.

I'll look into backporting this to `release-2`.

src/lib/utils/socket/socket.cpp
randombit: A lot of this could *really* use some `auto` treatment, `boost::asio::ip::tcp::resolver::results_type` whyyyyyy

reneme:
@randombit I took the liberty to fix the formatting error and squash the "address review comments" commit.

PR #4002, opened 2024-04-11
Update roughtime.cloudflare.com public key and port

lukevalenta:
Fixes #3937. Note that the old Cloudflare public key is still present in test vectors in the codebase in two locations, but this isn't problematic as they don't require the server to be online:
* https://github.com/randombit/botan/blob/afd43c536ff96e85d45837caa6f99125483c160a/src/tests/data/roughtime/roughtime_response.vec#L50
* https://github.com/randombit/botan/blob/afd43c536ff96e85d45837caa6f99125483c160a/src/tests/test_roughtime.cpp#L144

PR #4485, opened 2024-12-19
Backport: Resolve Boost deprecations / Roughtime online test

reneme:
See also #4484, #4477
Also backports the change from #4002 to fix the roughtime online test against roughtime.cloudflare.com.


randombit (approved):
Thanks

PR #4490, opened 2024-12-21
Add option to quote performance in microseconds or nanoseconds

randombit:
These days quoting operations in milliseconds per operation is quite archaic; on a reasonable machine a fast operation like ECDSA is typically just a few tens of microseconds.


reneme (approved):
Implementation itself looks good to me.

I just feel that the timer helper could benefit from extended usage of `std::chrono`, which already has built-in support for these scales for instance. I didn't see an obvious way to do a small increment in the scope of this change, so this is merely a note to future-us.

randombit:
> from extended usage of std::chrono, which already has built-in support for these scales for instance

Ah good point. I can take a look at this later.

PR #4494, opened 2024-12-23
FIX: SM4 GFNI needs AVX2

reneme:
This module uses AVX2 along with GFNI and should note that in its `info.txt`. Not sure we want to do that in the long-term, though. Perhaps better to implement this as a special case in `configure.py`, to disable GFNI when AVX2 was explicitly disabled with `--disable-avx2`? Are there other examples of such dependencies that need some sort of generalization, perhaps?

### DIFF
--- a/src/lib/block/sm4/sm4_gfni/info.txt
+++ b/src/lib/block/sm4/sm4_gfni/info.txt
@@ -7,6 +7,10 @@ name -> "SM4 GFNI"
 brief -> "SM4 using Intel GFNI"
 </module_info>
 
+<requires>
+simd_avx2
+</requires>
+
 <isa>
 gfni
 </isa>


randombit:
> Are there other examples of such dependencies that need some sort of generalization, perhaps

Probably yeah. VAES for instance is defined as VAES+AVX2 bits.

We already handle disabling these at runtime in CPUID cleanly (if you use `BOTAN_CLEAR_CPUID=avx2` it clears GFNI, VAES, etc also) but do not have similar logic for build time.

reneme:
Makes sense. Otherwise it won't find the avx-related helper headers that are part of `simd_avx2`, I reckon. Updated.

randombit (approved):

PR #4501, opened 2024-12-27
Deprecate a few more obscure and/or obsolete algorithms [ci skip]

randombit:
MD5 held on for a long time due to needing it for TLS 1.0, but that's not a factor anymore.

X9.19 MAC was obsolete 20 years ago

The original Keccak proposal is only, as far as I know, used in Ethereum. We only implement it because the code was added during the SHA-3 competition, prior to the NIST changes.

### DIFF
--- a/src/lib/hash/keccak/info.txt
+++ b/src/lib/hash/keccak/info.txt
@@ -4,6 +4,8 @@ KECCAK -> 20131128
 
 <module_info>
 name -> "Keccak"
+brief -> "Pre-standard version of SHA-3 originally proposed during NIST competition"
+lifecycle -> "Deprecated"
 </module_info>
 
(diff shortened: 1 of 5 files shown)


src/lib/hash/keccak/info.txt
```diff
@@ -4,6 +4,7 @@ KECCAK -> 20131128
 
 <module_info>
 name -> "Keccak"
+lifecycle -> "Deprecated"
```
reneme, line 7: Perhaps retroactively add a 'brief' explaining that this is _not_ SHA-3. It did cause plenty confusion in the past, so repeating this a few times surely won't hurt... 🙂

reneme (approved):

PR #4503, opened 2024-12-29
Add support for x-only EC point multiplication

randombit:
This saves a couple of field multiplications, due to avoiding the need to convert the y coordinate out of projective form.


reneme (approved):
Nice! 👍

support81bradyjrwayne (commented):
What platform did you use to configure this

Issue #4403, opened 2024-10-24
Removing Kyber r3 TLS ciphersuites


randombit:
Chrome 131 (eta next month) is going to remove support for Kyber r3 suites and switch to ML-KEM only

https://security.googleblog.com/2024/09/a-new-path-for-kyber-on-web.html

IMO we should do the same for Botan 3.7.0

This is as fast moving area and I don't think it makes sense to continue to support Kyber r3 suites for the entire life of Botan3, which is already committed to another 3 years of support (through 2027) and may well live longer depending on circumstances.

reneme:
Given the imminent patch release: #4404, should we already mark those suites as "deprecated" in 3.6.1 with a clear statement that they won't stay until Botan4?

PR #4507, opened 2024-12-30
Remove support for Kyber r3 key exchange in TLS

randombit:
GH #4403


reneme (approved):
There's still documentation about the Kyber R3 suite in `tls.rst`. Oh, and this also needs an update re: the pure ML-KEM suites.

Otherwise looks good to me. :)

reneme (approved):

PR #4508, opened 2024-12-30
Consistently use ClassicMcEliece for CMCE parameter names

randombit:
Otherwise we have some places where we have to string match on "mceliece" from the OIDs while others we use "ClassicMcEliece"

Kind of a minor thing but it seems easy to fix (right now, before any of this is in a release)


reneme (commented):
Good point! CI failures seem to stem from missing updates in the KAT files.

FAlbertDev (approved):
I see your point. My main concern with this change is that the parameter names do not match with the specification (see [NIST Round 4 Spec](https://classic.mceliece.org/mceliece-spec-20221023.pdf) (chapter 7) and [ISO draft](https://classic.mceliece.org/iso-mceliece-20230419.pdf) (chapter 10)). But I agree that the inconsistency is quite confusing. I'm fine with both variants 👍

PR #4515, opened 2024-12-31
Fix ML-DSA verification bugs

randombit:


reneme:
Thanks for looking into this!

Good to know that the actual implementation is fine and this was just a discrepancy whether or not to accept a signature that reveals too much about the private key.

PR #4518, opened 2025-01-01
Move EC_Point and related code to deprecated submodule

randombit:
It's not so viable at the moment to actually disable `legacy_ec_point` from a build; while the library does build many tests fail because we no longer support various deprecated or application/test specific curves. I guess even with the test failures someone might find it useful to disable, since it drastically reduces the attack surface of EC operations. The big next step would be to handle [at least a reasonable subset of] the missing curves via a 'generic' pcurves instance (current WIP).

#4027


reneme:
> The big next step would be to handle [at least a reasonable subset of] the missing curves via a 'generic' pcurves instance (current WIP).

This would certainly be important for applications that require application specific curves. For that, it would be great to allow coining custom pcurves at application level. I.e. compile the inner curve parameters (and implementation/representations helpers) within the application and then instantiate Botan's polymorph pcurve wrapper class from it.

I'm wondering whether this is feasible without exposing most of the pcurve internals to the public API, though.

randombit:
> This would certainly be important for applications that require application specific curves. For that, it would be great to allow coining custom pcurves at application level. I.e. compile the inner curve parameters (and implementation/representations helpers) within the application and then instantiate Botan's polymorph pcurve wrapper class from it.

I think the furthest I'd want to go along these lines would be documenting how to add a new set of parameters to the library, and letting people do this as a local patch.

The 'generic pcurves' I mentioned is more of this shape - something that has the interface of `pcurves.h` but is generic at runtime (so no pre-computation of Montgomery, etc). You just provide `p`, `a`, `b`, etc as parameters and get something back that computes over that curve. This will likely be a less efficient than full strength current pcurves, but (hopefully) still better than `BigInt`-based `EC_Point`.

In the end we're going to end up with 3 elliptic curve implementations (at least for Botan3):

* Current pcurves approach of pushing `constexpr` to the limits for the set of common curves
* "Generic pcurves" which handles cases not covered by current pcurves and/or cases where the user doesn't want to include dedicated code to to space reasons (eg one disables `pcurves_secp192r1` and then P-192 support falls back to `pcurve_generic`)
* `EC_Point` for either people directly using `EC_Point` in their own code or for the truly weirdo cases (cofactors, 1024 bit field, etc)

The first two will hopefully be able to share a lot of code, and the last is removed in Botan4.

src/lib/pubkey/ec_group/ec_apoint.cpp
```diff
@@ -140,6 +149,19 @@ EC_AffinePoint EC_AffinePoint::negate() const {
...
 
+std::vector<uint8_t> EC_AffinePoint::serialize(EC_Point_Format format) const {
+   if(format == EC_Point_Format::Compressed) {
+      return this->serialize_compressed();
+   } else if(format == EC_Point_Format::Uncompressed) {
+      return this->serialize_uncompressed();
+   } else {
+      // TODO(Botan4) Remove this
+      auto enc = this->serialize_uncompressed();
+      enc[0] ^= enc[enc.size() - 1] & 0x01;
```
reneme, line 160:
```suggestion
      enc.front() ^= enc.back() & 0x01;
```

reneme, line 160: I'm guessing this implements the hybrid encoding? Perhaps a comment would be helpful, for clarity. :)

src/lib/pubkey/ecies/ecies.h
```diff
@@ -173,18 +180,29 @@ class BOTAN_PUBLIC_API(2, 0) ECIES_KA_Operation {
...
                                  const EC_Point& other_public_key_point) const;
+#endif
+
+      /**
+      * Performs a key agreement with the provided keys and derives the secret from the result
+      * @param eph_public_key_bin the encoded (ephemeral) public key which belongs to the used (ephemeral) private key
+      * @param other_public_key_point public key point of the other party
+      */
+      SymmetricKey derive_secret(const std::vector<uint8_t>& eph_public_key_bin,
+                                 const EC_AffinePoint& other_public_key_point) const;
```
reneme, line 205: The `eph_public_key_bin` should be a `std::span`.

src/lib/pubkey/ecies/ecies.cpp
```diff
@@ -160,6 +164,47 @@ SymmetricKey ECIES_KA_Operation::derive_secret(const std::vector<uint8_t>& eph_p
...
+#endif
+
+/**
+* ECIES secret derivation according to ISO 18033-2
+*/
+SymmetricKey ECIES_KA_Operation::derive_secret(const std::vector<uint8_t>& eph_public_key_bin,
+                                               const EC_AffinePoint& other_public_key_point) const {
+   if(other_public_key_point.is_identity()) {
+      throw Invalid_Argument("ECIES: peer public key point is the identity element");
+   }
```
reneme, line 176: Nit (not sure about the formatting):

```suggestion
   BOTAN_ARG_CHECK(!other_public_key_point.is_identity(),
                   "ECIES: peer public key point is the identity element");
```

reneme (approved):
So much code waiting to be deleted with Botan4 😄  Just some nits, otherwise LGTM.

src/lib/pubkey/ec_group/ec_apoint.cpp
randombit: Added a comment

randombit:
> assuming that the performance isn't terribly different. 

I'm generally aiming for at worst parity with `EC_Point` + `BigInt`. Given that the pcurves implementations are (with generally speaking the same multiplication algorithms) 2-3 times faster, I don't expect this to be an issue. Better inlining + everything on the stack makes a big difference.

> So much code waiting to be deleted with Botan4

Some of it I actually hope to remove sooner rather than later. Specifically, once `EC_Point` really is only used for the strangest curves, it seems to me fine to begin de-optimizing it in various ways to simplify the implementation. (For example by using something simple like a Montgomery ladder for all point multiplications.) We've already had a bit of that in #4385

PR #4520, opened 2025-01-02
Deprecate serializing compressed points in EC keys

randombit:
Note this is not deprecating compressed points in general, only the ability to choose for each EC key object (using set_point_encoding) how the point is serialized at any given time.

### DIFF
--- a/doc/deprecated.rst
+++ b/doc/deprecated.rst
@@ -55,6 +55,15 @@ example using ECDSA signatures or TLS, but only applications doing unusual thing
 such as custom elliptic curve parameters, or creating your own protocol using
 elliptic curve points.
 
+- The `EC_Point` class is used to perform point arithmetic. This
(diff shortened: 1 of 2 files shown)


reneme (commented):
FWIW: We actually use compressed points to encode public keys. This is done to protect the underlying (classified) curve parameters. Shoutout to Kerckhoff... 😏

Of course we could also trim the SEC1 encoding after the effect. However, when wrapping them into certificates (via the PKCS10 request) that becomes impractical. Currently, one can simply use `set_point_encoding()` on the underlying ECC private key and get a signed CSR with a compressed public point.

randombit:
> This is done to protect the underlying (classified) curve parameters.

I had a feeling this was the case.

Perhaps this can be handled in some other way; on a curve specific (vs key specific) encoding option? That would actually be a bit nicer anyway in this specific scenario. Lacking a specific plan here, I'll close this for now.

PR #4525, opened 2025-01-03
Let the entropy module be optional

randombit:
It was previously required by the rng module, but this isn't really necessary, especially for anyone who is just using the system RNG. (And/or NIST DRBGs seeded with the system RNG.)

### DIFF
--- a/src/lib/rng/auto_rng/auto_rng.cpp
+++ b/src/lib/rng/auto_rng/auto_rng.cpp
@@ -6,11 +6,12 @@
 
 #include <botan/auto_rng.h>
 
-#include <botan/entropy_src.h>
 #include <botan/hmac_drbg.h>
 #include <botan/internal/loadstor.h>
 
-#include <array>
+#if defined(BOTAN_HAS_ENTROPY_SOURCE)
+   #include <botan/entropy_src.h>
+#endif
 
 #if defined(BOTAN_HAS_SYSTEM_RNG)
    #include <botan/system_rng.h>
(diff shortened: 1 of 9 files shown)


src/lib/rng/auto_rng/auto_rng.cpp
```diff
@@ -6,19 +6,22 @@
...
+#endif
 
 #if defined(BOTAN_HAS_SYSTEM_RNG)
    #include <botan/system_rng.h>
 #endif
 
 namespace Botan {
 
+class Entropy_Sources;
+
```
reneme, line 24: This relates to lines 68 to 72 where `Entropy_Sources` is actually used of Botan doesn't have the system rng enabled. Probably, this needs another ifdef.

reneme (changes requested):

src/lib/rng/auto_rng/auto_rng.cpp
randombit: Wasn't even required at al - removed

reneme: Did you push all changes here? If I configure without `entropy` and `system_rng` botan fails to build with:

```
src/lib/rng/auto_rng/auto_rng.cpp:68:22: error: incomplete type 'Botan::Entropy_Sources' named in nested name specifier
   68 |       AutoSeeded_RNG(Entropy_Sources::global_sources(), reseed_interval)
      |                      ^~~~~~~~~~~~~~~~~
build/include/public/botan/rng.h:26:7: note: forward declaration of 'Botan::Entropy_Sources'
   26 | class Entropy_Sources;
      |       ^
```

reneme (changes requested):
See previous comment about auto_rng.cpp

src/lib/rng/auto_rng/auto_rng.cpp
randombit: Missing add, fixed now

reneme (approved):

PR #4038, opened 2024-05-02
Some EC_Group usage cleanups

randombit:
This hides all symbols of CurveGFp and CurveGFp_Repr (which were already marked as BOTAN_UNSTABLE_API and intentionally undocumented) and makes them accessible only to a few specific friends.

For EC_Point deprecate a number of functions only needed for internal use. Also add a couple of functions to simplify a future bridge between our current APIs and pcurves.

In EC_Group likewise begin some cleanup and deprecation.

PR #4533, opened 2025-01-05
Modify restrictions on application specific EC groups

randombit:
GH #4038 and #4089 introduced a new EC_Group constructor for application specific curves, which places various restrictions on the sizes and parameters.

This commit slightly ammends these restrictions.

* The lower bound on the size of the field is 192 bits, not 128. We continue to accept as low as 128 bits for now, but note it as deprecated. In Botan4, the smallest group we'll allow an application to create is 192 bits.

* As an extension on the size of p similar to accepting curves using P-521, accept the X9.62 239-bit prime. This will allow us to remove these X9.62 groups in Botan4, but give anyone using them in the wild the ability to add them back if they must.

* Check the discriminant is not zero, and that the generator is a valid point on the curve. This is arguably a SemVer violation but your curve was anyway not valid so too bad.

### DIFF
--- a/src/lib/pubkey/ec_group/ec_group.cpp
+++ b/src/lib/pubkey/ec_group/ec_group.cpp
@@ -370,11 +370,32 @@ EC_Group::EC_Group(const OID& oid,
                    const BigInt& base_y,
                    const BigInt& order) {
    BOTAN_ARG_CHECK(oid.has_value(), "An OID is required for creating an EC_Group");
-   BOTAN_ARG_CHECK(p.bits() >= 128, "EC_Group p too small");
+
+   // TODO(Botan4) remove this and require 192 bits minimum
+#if defined(BOTAN_DISABLE_DEPRECATED_FEATURES)
+   constexpr size_t p_bits_lower_bound = 192;
+#else
+   constexpr size_t p_bits_lower_bound = 128;
+#endif
+
+   BOTAN_ARG_CHECK(p.bits() >= p_bits_lower_bound, "EC_Group p too small");
    BOTAN_ARG_CHECK(p.bits() <= 521, "EC_Group p too large");
 
    if(p.bits() == 521) {
-      BOTAN_ARG_CHECK(p == BigInt::power_of_2(521) - 1, "EC_Group with p of 521 bits must be 2**521-1");
+      const auto p521 = BigInt::power_of_2(521) - 1;
+      BOTAN_ARG_CHECK(p == p521, "EC_Group with p of 521 bits must be 2**521-1");
+   } else if(p.bits() == 239) {
+      const auto x962_p239 = []() {
+         BigInt p239;
+         for(size_t i = 0; i != 239; ++i) {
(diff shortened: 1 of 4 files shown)


randombit:
With these changes, the curves currently in Botan that a) are currently deprecated and I expected to remove and b) could not be added back as application specific curves in Botan4 are

* brainpool160r1
* secp160r1
* secp160r2
* secp160k1
* secp224k1

The 160 bit curves are realistically already breakable by a well resourced attacker, and should absolutely go.

And secp224k1 is pretty weird - p mod 4 == 1, the order is one bit longer than the prime, etc. With a bit of effort we could support it in pcurves but it's really niche and doesn't seem worthwhile. I would guess most of the people using it are confusing it with secp224r1 or secp256k1.

src/lib/pubkey/ec_group/ec_group.cpp
```diff
@@ -370,11 +370,25 @@ EC_Group::EC_Group(const OID& oid,
                    const BigInt& base_y,
                    const BigInt& order) {
    BOTAN_ARG_CHECK(oid.has_value(), "An OID is required for creating an EC_Group");
+   // TODO(Botan4) change lower bound to 192 bits
    BOTAN_ARG_CHECK(p.bits() >= 128, "EC_Group p too small");
    BOTAN_ARG_CHECK(p.bits() <= 521, "EC_Group p too large");
```
reneme, line 375: Perhaps `#ifdef` on `BOTAN_DISABLE_DEPRECATED_FEATURES` and already raise the bound in Botan 3 (and disable all pcurves below this bound)?

src/lib/pubkey/ec_group/ec_group.cpp
```diff
@@ -370,11 +370,25 @@ EC_Group::EC_Group(const OID& oid,
...
+      BOTAN_ARG_CHECK(p == p521, "EC_Group with p of 521 bits must be 2**521-1");
+   } else if(p.bits() == 239) {
+      const auto x962_p239 = []() {
+         BigInt p239;
+         for(size_t i = 0; i != 239; ++i) {
+            if(i < 47 || ((i >= 94) && (i != 143))) {
+               p239.set_bit(i);
+            }
+         }
+         return p239;
```
reneme, line 388: This strikes me like a good place to use https://github.com/randombit/botan/pull/4475, no?

reneme (approved):

src/lib/pubkey/ec_group/ec_group.cpp
randombit: Maybe? This code only gets executed if you are actually trying to register an X9.62 239-bit prime, which I expect to happen ~~~ never. Like there is maybe one user of those curves, and I don't want to leave them completely high and dry, but don't care that much either.

[OTOH this perhaps argues to not including the 239-bit exception at all, and then if it really is an issue in Botan4 they can open an issue and then we can add the exception at the time.]

reneme: Check.

Issue #4519, opened 2025-01-02
Baremetal build fails if no RTC is available


TreshanAyesh:
I want to build the library for A53 cores on Zynq Ultrascale+ device to use on a standalone/baremetal domain.
I configured the files as below-
./configure.py --cc-bin=aarch64-none-elf-gcc --cc=gcc --ar-command=aarch64-none-elf-gcc-ar --cpu=armv8-a --os=none --prefix=/home/thakshana/Thakshana/SP-67/botan_a53_none/ --disable-modules=thread_utils --without-os-features=threads

but when the make command is executed, it fails giving the error below -

src/cli/timing_tests.cpp: In member function 'virtual void Botan_CLI::MARVIN_Test_Command::go()':
src/cli/timing_tests.cpp:562:51: error: no matching function for call to 'Botan::DataSource_Stream::DataSource_Stream(const std::string&)'
  562 |          Botan::DataSource_Stream key_src(key_file);
      |                                                   ^
In file included from build/include/public/botan/pkcs8.h:11,
                 from src/cli/timing_tests.cpp:49:
build/include/public/botan/data_src.h:156:7: note: candidate: 'Botan::DataSource_Stream::DataSource_Stream(std::istream&, std::string_view)'
  156 |       DataSource_Stream(std::istream&, std::string_view id = "<std::istream>");
      |       ^~~~~~~~~~~~~~~~~
build/include/public/botan/data_src.h:156:25: note:   no known conversion for argument 1 from 'const std::string' {aka 'const std::__cxx11::basic_string<char>'} to 'std::istream&' {aka 'std::basic_istream<char>&'}
  156 |       DataSource_Stream(std::istream&, std::string_view id = "<std::istream>");

reneme:
This is an issue in botan's CLI tool. Perhaps you don't need that anyway, on your target platform? You could try configuring with --build-targets=static (or shared, if you want a shared library). This will build just the library without tests or the CLI tool.

Regarding the actual build failure: My best guess is that the baremetal configuration doesn't have support for a file system and hence the required constructor is ifdef'ed out. This is something we should look into on our side, thanks for reporting.

randombit:
I'll check if we can bring back the baremetal CI build. IIRC there was some issue with the Ubunto arm-none gcc but I don't remember the details anymore.

randombit:
@TreshanAyesh the patch in #4521 should address the build problems, ptal.

[...]

PR #4535, opened 2025-01-07
Update docs and examples regarding custom RNG and entropy sources

randombit:
See #4519 for recent context


reneme (approved):
Nice, interesting read! (I can't really contribute to the actual content, as my understanding of embedded systems and platforms is lacking)

PR #4536, opened 2025-01-08
FIX: loading pinned certificates from OCSP responses

### LINKED ISSUE #4502: OCSP Response containing multiple certificates
### Discussed in https://github.com/randombit/botan/discussions/4497

<div type='discussions-op-text'>

<sup>Originally posted by **overocean** December 24, 2024</sup>
Hi guys, I found that when we construct OCSP Response, we call `decode_optional_list()` to construct the certificates. Inside the body of the while loop in this function, `certbits` will be the sequence of certificates in the response data, and the X509_Certificate constructor will construct an object with the first certificate data in `certbits`. Therefore, other certificates are discarded. This seems a bug to me.

```
   BER_Decoder list(obj);

   while(list.more_items()) {
      BER_Object certbits = list.get_next_object();
      X509_Certificate cert(certbits.bits(), certbits.length());
      output.push_back(std::move(cert));
   }
```
</div>

reneme:
Before, the OCSP response would be parsed but if it contained more than one pinned certificate, the list would be truncated to only the first one. This may prevent successful path-building for verifying the OCSP response, for instance.

Closes #4502

### DIFF
--- a/src/lib/x509/ocsp.cpp
+++ b/src/lib/x509/ocsp.cpp
@@ -35,12 +35,15 @@ void decode_optional_list(BER_Decoder& ber, ASN1_Type tag, std::vector<X509_Cert
    }
 
    BER_Decoder list(obj);
-
-   while(list.more_items()) {
-      BER_Object certbits = list.get_next_object();
-      X509_Certificate cert(certbits.bits(), certbits.length());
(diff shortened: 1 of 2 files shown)


randombit (approved):
Thanks for taking this!

Issue #4543, opened 2025-01-12
Fix ESDM build in CI


randombit:
It appears to not be using ccache so adds 2 minutes to the already slow coverage build

reneme:
Frankly, I believe the bigger issue is that (apparently) some TPM2 dependency installationis attempting to start some sort of service which fails and timeouts after a while. When watching the log, there is a fairly long wait while `apt install` is executed.

reneme:
Let's see if #4547 improves the situation nevertheless.

reneme:
But it seems that the ccache local cache is configured to be quite small. Too small to hold all cached object files, apparently:

```
ccache --show-stats
  Cacheable calls:   20678 / 20750 (99.65%)
    Hits:             68[57](https://github.com/randombit/botan/actions/runs/12846924045/job/35823853454#step:6:59) / 20678 (33.16%)
      Direct:         2161 /  6857 (31.52%)
      Preprocessed:   4696 /  6857 (68.48%)
    Misses:          13821 / 20678 (66.84%)
  Uncacheable calls:    72 / 20750 ( 0.35%)
  Local storage:
    Cache size (GB):   0.2 /   0.2 (99.13%)   <= apparently we're thrashing the cache
    Cleanups:         8731
    Hits:             6857 / 20678 (33.16%)
    Misses:          13821 / 20678 (66.84%)
```

reneme:
[Testing this with a larger cache](https://github.com/randombit/botan/pull/4574), I measured 36m 34s on a cold cache, vs. 29m 19s on a warm cache, for running the entire coverage CI job. The ESDM build benefitted from the warm ccache. So, I would say, the goal of this ticket was achieved by #4547.

PR #4547, opened 2025-01-14
Enable ccache for jitterentropy build

reneme:
See #4543.

This enables ccache for the jitterentropy build. The ESDM build picks up ccache automatically, via meson's default configuration.

However, we didn't benefit from caching because the dependency builds happened before the cache was restored. This is now fixed by splitting `setup_gh_actions.sh` similarly to `setup_gh_actions.ps`.


randombit (approved):
LGTM! Thanks for looking at this

PR #4548, opened 2025-01-14
Address various warnings new in clang-tidy 19

randombit:


src/.clang-tidy
```diff
@@ -27,11 +28,14 @@ Checks: >
     -cppcoreguidelines-slicing,
     -hicpp-explicit-conversions,
     -misc-const-correctness,
+    -misc-include-cleaner,
     -misc-redundant-expression,
     -misc-misplaced-const,
     -misc-confusable-identifiers,
     -modernize-avoid-bind,
     -modernize-pass-by-value,
+    -modernize-use-ranges,
```
reneme, line 37: We might want to enable and act on that as soon as we have the compilers for it. Perhaps leave a `TODO(Botan4)` or something? Edit: oh, right. I found the comment in the other list.

reneme (approved):

PR #4551, opened 2025-01-14
SP.800-108 may handle different counter and L encoding widths

reneme:
### Desclaimer

Turns out SP.800-108 leaves a lot of unspecified degrees of freedom. Even with this patch we don't come close to fully support all allowed variants of the KDF. Frankly, any application requiring more flexibility in the implementation of these KDFs should probably just implement them manually. For instance, in feedback-mode the "counter" value is optional, but we don't support that currently. Even the processing order of the "counter", "label" and "context" fields may be changed. Not even the encoding of the counter and length fields is specified and left to the implementation. 😨 

### This Patch

With this PR, we're extending the support for alternative field lengths for both the "counter" and the "output length". Supported are 8, 16, 24, and 32 bits (e.g. the byte-aligned values). To support this, we refactored the implementations of the "Counter", "Feedback" and "Pipeline" modes significantly.

Regarding test data: The [CAVP](https://csrc.nist.gov/Projects/Cryptographic-Algorithm-Validation-Program/Key-Derivation) does provide test vectors. But they are not based on the KDFs' public interface. Instead, they just provide random "fixed data inputs" that don't fit the internal data structure as specified in SP.800-108 and are thus not usable. >.<

We built simple python-implementations for the KDFs to generate test data. The used python scripts are added to the test vector files for future reference. The "32bit/32bit"-output of those python implementations matches the existing implementation that was tested against BouncyCastle, so we're fairly confident they are correct. Unfortunately, BouncyCastle doesn't support field encodings other than 32 bit wide.

See #4449


src/lib/kdf/sp800_108/sp800_108.h
```diff
@@ -24,8 +24,10 @@ class SP800_108_Counter final : public KDF {
 
       /**
       * @param mac MAC algorithm to use
+      * @param r  encoding bit-length of the internal counter {8, 16, or 32}
+      * @param L  encoding bit-length of the output length in bits {8, 16, or 32}
```
randombit, line 28: These can be 24 bits also yes?

src/lib/kdf/sp800_108/sp800_108.h
```diff
@@ -59,7 +63,12 @@ class SP800_108_Feedback final : public KDF {
 
       std::unique_ptr<KDF> new_object() const override;
 
-      explicit SP800_108_Feedback(std::unique_ptr<MessageAuthenticationCode> mac) : m_prf(std::move(mac)) {}
+      /**
+      * @param mac MAC algorithm to use
+      * @param r  encoding bit-length of the internal counter {8, 16, or 32}
+      * @param L  encoding bit-length of the output length in bits {8, 16, or 32}
```
randombit, line 69: Missing 24 bits?

randombit (commented):
Change looks fine modulo one minor issue in the comments regarding 24 bit fields.

Can you please though explain more about what you need this for? In particular it seems like, based on how the test vectors were generated, no other implementation of SP800-108 KDFs supports fields other than 32 bits?? So I'm a bit confused on this point, since the whole reason for our zoo of KDFs is compatibility with other implementations. We can discuss on Signal if you prefer.

reneme:
> Can you please though explain more about what you need this for?

Essentially, we want to replace an internal implementation of this KDF that uses shorter representations for both fields. Our initial hope was to be able to specify the full KDF (including extract-expand from SP.800-56c) in the algo-string parameter of `PK_Key_Agreement` (as sketched in #4449). Though, at this point, the degrees of freedom in the specifications of this construction just don't make it feasible without further complicating the Botan APIs involved. See [further details here](https://github.com/randombit/botan/issues/4449#issuecomment-2593428894). This patch lets us replace our internal SP.800-108 expansion step within SP.800-56c by a Botan-provided implementation at least. 🙂 

Perhaps one day, the API rework starting with #4318 (perhaps in combination with #3275) might provide enouogh API flexibility to allow that. But, admittedly, I'm not sure that this sort of flexibility is a goal worth aspiring to, given that SP.800-x is so vague in its specification.

> ..., no other implementation of SP800-108 KDFs supports fields other than 32 bits?

There's [a Rust implementation](https://docs.rs/rust-kbkdf/latest/rust_kbkdf/) that allows specifying the counter width at least. In fact, it even claims it allows disabling the counter altogether (which isn't completely insane for "feedback" and "pipeline" modes). I didn't find an implementation that allows configuring the output length encoding width, though.

reneme:
> Can you pull some tests from there to check for interop?

Done. Note though that we're not compatible to their currently-released version (1.1.0) because of [encoding bugs on their side](https://gitlab.com/michael.j.boquard/rust-kbkdf/-/merge_requests/1). There's even one more similar bug in their pipeline-mode implementation. I'll submit a merge request to them shortly.

randombit (approved):

PR #4555, opened 2025-01-15
Remove the workspace argument to various ECC interfaces

randombit:
This was a performance crutch for EC_Point and is no longer required.

This should not go into a release without #4554 as otherwise it would pessimize oddball curves

### DIFF
--- a/src/lib/pubkey/ec_group/ec_apoint.h
+++ b/src/lib/pubkey/ec_group/ec_apoint.h
@@ -52,9 +52,7 @@ class BOTAN_UNSTABLE_API EC_AffinePoint final {
       static std::optional<EC_AffinePoint> from_bigint_xy(const EC_Group& group, const BigInt& x, const BigInt& y);
 
       /// Multiply by the group generator returning a complete point
-      ///
-      /// Workspace argument is transitional
-      static EC_AffinePoint g_mul(const EC_Scalar& scalar, RandomNumberGenerator& rng, std::vector<BigInt>& ws);
+      static EC_AffinePoint g_mul(const EC_Scalar& scalar, RandomNumberGenerator& rng);
 
       /// Return the identity element
       static EC_AffinePoint identity(const EC_Group& group);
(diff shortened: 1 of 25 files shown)


src/lib/pubkey/ec_group/ec_apoint.h
```diff
@@ -78,16 +82,16 @@ class BOTAN_UNSTABLE_API EC_AffinePoint final {
...
+         return this->mul(scalar, rng);
+      }
 
       /// Multiply a point by a scalar, returning the byte encoding of the x coordinate only
-      ///
-      /// Workspace argument is transitional
-      secure_vector<uint8_t> mul_x_only(const EC_Scalar& scalar,
-                                        RandomNumberGenerator& rng,
-                                        std::vector<BigInt>& ws) const;
+      secure_vector<uint8_t> mul_x_only(const EC_Scalar& scalar, RandomNumberGenerator& rng) const;
```
reneme, line 94: For the record: This API [was added recently](https://github.com/randombit/botan/pull/4503) and hence doesn't require a deprecation, correct?

src/tests/test_ecc_pointmul.cpp
```diff
@@ -66,23 +65,22 @@ class ECC_Varpoint_Mul_Tests final : public Text_Based_Test {
 
          const auto group = Botan::EC_Group::from_name(group_id);
 
+   #if defined(BOTAN_HAS_LEGACY_EC_POINT)
          std::vector<Botan::BigInt> ws;
```
reneme, line 69: This one can go the way of the Dodo as well.

reneme (approved):

src/lib/pubkey/ec_group/ec_apoint.h
randombit: Correct.

If this ends up not making it for 3.7 I'll switch it to be like the others with a second deprecated function that ignores the arg

randombit: Changed so as to not break SemVer

PR #4558, opened 2025-01-16
Modify Montgomery exponentiation to return results in Montgomery form

randombit:
DSA and ElGamal just immediately convert the result to standard form; the performance of these doesn't matter much at all anymore.

For RSA, keep the values in Montgomery form using the same flow as described as "Smooth-CRT RSA" in https://eprint.iacr.org/2007/039.pdf

### DIFF
--- a/src/lib/pubkey/rsa/rsa.cpp
+++ b/src/lib/pubkey/rsa/rsa.cpp
@@ -40,7 +40,7 @@ class RSA_Public_Data final {
       BigInt public_op(const BigInt& m) const {
          const size_t powm_window = 1;
          auto powm_m_n = monty_precompute(m_monty_n, m, powm_window, false);
-         return monty_execute_vartime(*powm_m_n, m_e);
+         return monty_execute_vartime(*powm_m_n, m_e).value();
       }
 
       const BigInt& get_n() const { return m_n; }
@@ -50,6 +50,8 @@ class RSA_Public_Data final {
       size_t public_modulus_bits() const { return m_public_modulus_bits; }
 
       size_t public_modulus_bytes() const { return m_public_modulus_bytes; }
+
+      const std::shared_ptr<const Montgomery_Params>& monty_n() const { return m_monty_n; }
 
    private:
       BigInt m_n;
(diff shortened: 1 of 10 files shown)


src/lib/pubkey/rsa/rsa.cpp
```diff
@@ -473,10 +473,12 @@ class RSA_Private_Operation {
...
          /*
-         TODO
-         Consider using Montgomery reduction instead of Barrett, using
-         the "Smooth RSA-CRT" method. https://eprint.iacr.org/2007/039.pdf
+         All normal implementations generate p/q of the same bitlength,
+         so this should rarely occur in practice
          */
+         if(m_private->primes_imbalanced()) {
+            return monty_exp(m_public->monty_n(), m, m_private->get_d(), m_public->get_n().bits()).value();
+         }
```
FAlbertDev, line 481: @randombit Sorry for digging this up after the merge. We are currently preparing some documents for the BSI regarding Botan 3.7.1. 
I've skimmed the paper (https://eprint.iacr.org/2007/039.pdf). I admit that I do not understand why the attack described does not apply to balanced p and q. Can you help me understand? I want to mention this countermeasure in our crypto documentation.
 Also, doesn't this distinction induce a (very minor) side channel by leaking whether p and q are balanced or not?

randombit: None of the attacks in that paper apply to Botan (or at least not any recent version, I guess some variation of the BEEA attack maybe applied at some point). This PR was just adopting the SmoothCrt proposal because it's slightly faster and cleaner that converting into and out of Montgomery form several times. Also operations on `Montomery_Int` (unlike `BigInt`) always takes time proportional to the modulus rather than the values. So in the old code where we computed `j1-j2` we did it as `(p + j1) - j2` to avoid some of these problems, but even then if `j1` or `j2` was small (say a single word) this would be detectable via executed loop iterations. In contrast `Montgomery_Int` avoids this and always computes as if the integers were the same magnitude as the modulus.

Using a slow path without CRT for imbalanced primes is just to account for the fact that if the primes are really imbalanced (say p is 32 bits and q is 2016) then `Montgomery::from_wide_int` doesn't work. We probably could handle p and q being a bit or two apart, at least with some modification. However normal implementations don't generate keys like this, and making the common path slower or more complicated to account for this case doesn't seem worth it to me.

> Also, doesn't this distinction induce a (very minor) side channel by leaking whether p and q are balanced or not?

It absolutely does. However this leak seems to me inconsequential because it is a single static bit of information - are p and q balanced or not. This property is fixed for any key. If there was an attack made possible by this you could just assume they were imbalanced and proceed accordingly. FWIW BoringSSL uses a similar flow.

FAlbertDev: That makes sense. Thanks for your detailed elaboration!
> However normal implementations don't generate keys like this, and making the common path slower or more complicated to account for this case doesn't seem worth it to me.

I totally agree with that. 

> However this leak seems to me inconsequential because it is a single static bit of information - are p and q balanced or not.

I didn't want to be nitpicky. 😉 Also, since most implementations will create balanced primes, it's no problem anyway.

PR #4560, opened 2025-01-16
Further deprecations

randombit:
Finite field DSA may be with us for some time still; apparently even BoringSSL hasn't been able to excise it yet. But we should strongly warn away from it and deprecating is probably a good start.

Deprecate numthry.h - this is only really useful for someone using BigInt for a calculator, which isn't our intended use case.

Deprecate free standing Ed25519 functions - not sure how these ended up here.

Add a deprecation warning to the already deprecated gost_3410.h


reneme (approved):

randombit:
@reneme FYI the BSI policy requires DSA. I was surprised by this but checking the TR it's still recommended until 2029. :/ Currently the CI build for that also uses disable-deprecated, so the build fails. I'm not 100% sure of the intent behind disabling deprecated features for that config. For now I just removed the deprecated annotation in `dsa/info.txt`.

PR #4562, opened 2025-01-17
Add aarch64 native builds to CI

randombit:


randombit:
Valgrind builds are pointing to uninitialized memory in TLS in GCC, and a const-time issue in Classic McEliece. I can't replicate either issue with GCC 14/Clang 19 on a Neoverse N1 machine running Debian.

randombit:
cc @reneme @FAlbertDev regarding Classic McEliece

reneme:
@randombit caution! I'll push some commits into this branch to enable debug symbols for valgrind and work on the matrix.

reneme:
> Valgrind builds are pointing to uninitialized memory in TLS in GCC, and a const-time issue in Classic McEliece. I can't replicate either issue with GCC 14/Clang 19 on a Neoverse N1 machine running Debian.

I enabled `--with-debug-info` for the valgrind targets. Hopefully, this build will provide some extra info: https://github.com/randombit/botan/actions/runs/12831471225/job/35782005838?pr=4562

reneme:
The valgrind issue in CMCE was due to a missing `unpoison` before performing a sanity check. It remains unclear why this wasn't an issue on x86.

Our hypothesis: the sanity check ensures that some expected padding bits are all-zero. I.e. the values that are checked in the relevant `BOTAN_ARG_CHECK` are actually _not_ secret dependent (as they are only the 0-bits of the padding), the secret bits were shifted-out before.

Perhaps valgrind's bit-level initialization tracking behaves different on ARM. We tried to create a simple regression test for this in `ct_selftest`, though this came back clean. Nevertheless, the `unpoison` as added in the commit we pushed earlier seems sensible anyway.

randombit:
@reneme Thanks for digging into this. Glad it was that and not a situation where the compiler was unexpectedly inserting jumps into nominally constant time code, as that would cast some doubt on how reliable our general process is for writing/testing this kind of thing

randombit:
@reneme The nightly jobs are still failing. This needs further investigation before anything can merge - can you open up a new PR with the unpoison fix + ct_selftest change? Since that looks ready to go

Weird new error in https://github.com/randombit/botan/actions/runs/12952927152/job/36131406325

```
  kyber_helpers:
  compress<10> ran 1 tests in 0.32 msec all ok
  compress<10>(decompress())> ran 1 tests in 0.14 msec all ok
  compress<11> ran 1 tests in 0.28 msec all ok
  compress<11>(decompress())> ran 1 tests in 0.15 msec all ok
  compress<1> ran 1 tests in 0.36 msec all ok
  compress<1>(decompress()) ran 1 tests in 0.02 msec all ok
  compress<4> ran 1 tests in 0.36 msec all ok
  compress<4>(decompress()) ran 1 tests in 0.02 msec all ok
  compress<5> ran 1 tests in 0.31 msec all ok
  compress<5>(decompress()) ran 1 tests in 0.02 msec all ok
  decompress<10> ran 1 tests 1 FAILED
  Failure 1: decompress<10>(512) = 1665; expected 1664 (at src/tests/test_kyber.cpp:394)
  decompress<11> ran 1 tests 1 FAILED
  Failure 1: decompress<11>(1024) = 1665; expected 1664 (at src/tests/test_kyber.cpp:394)
  decompress<1> ran 1 tests in 0.02 msec all ok
  decompress<4> ran 1 tests in 0.02 msec all ok
  decompress<5> ran 1 tests in 0.02 msec all ok
```

reneme:
Currently trying [this commit](https://github.com/reneme/botan/commit/7435bae6c03d5035b6aa4ac890481fe71d2e675b) in my fork: https://github.com/reneme/botan/actions/runs/13053374814/job/36418434039

reneme:
Last remaining issue: clang finds some jump on uninitialized value in `TLS::Session_Manager_In_Memory`. I'd be surprised if that were a `ct_poison` related issue. I'm currently running it in my fork with "track origins" to be sure: https://github.com/reneme/botan/actions/runs/13068269432

reneme:
@randombit I just realized, that the CT selftest added in https://github.com/randombit/botan/commit/9545c7b81081c207c2c39b8116435700bbf767a3 actually fails for clang on ARM and `-O3`. This didn't occur to me before, but I might have missed it in all the valgrind barf. Do you see this in your nightly runs as well?

This is somewhat surprising, because it somewhat contradicts my understanding of how valgrind tracks "uninitialized bits" O.o

reneme:
Regarding the valgrind warnings in `TLS::Session_Manager_In_Memory`, I'm frankly starting to suspect a clang bug (or valgrind perhaps). But some undefined behavior bug can't fully be ruled out either.

The jump on uninitialized data happens within the `std::map` member of `Session_Manager_In_Memory`. When traversing the Rb-tree to insert/remove nodes. We're using a `Session_ID` (strong type on `std::vector<uint8_t>`) as the map's key (so internally `std::map` uses `std::lexicographical_compare`).

After some restructuring of the test code, I'm now seeing valgrind errors for both the `::store()` and the `::remove()` methods. Curiously, though, "track origins" _always_ points to a stack allocation in `::store()` for the origin of the undefined value. My current working hypothesis: `::store()` writes undefined data as the map's key under some circumstance, and `::remove()` trips over it while traversing the tree.

I'll push my current work-in-progress in here for reference. But unfortunately, I believe that's it for today.

PR #4566, opened 2025-01-17
Add a test that EC_Group with a custom OID looks up the correct pcurves

randombit:
See https://github.com/randombit/botan/pull/4554#discussion_r1919868906 for discussion


reneme (approved):

reneme:
Nice! That seems feasible. We didn't try to _explicitly_ register the OID. Thanks for looking into this. :)

PR #4569, opened 2025-01-18
When computing modular inverses distingush which case we are in

randombit:
There are several different scenarios where we need to compute a modular inverse

* A secret value modulo a secret prime (eg RSA-CRT setup)
* A (potentially) secret value modulo a public prime (eg inversion modulo an DL group order)
* A secret value modulo a public value that is not prime (RSA blinding setup)
* RSA secret exponent calculation, where e and phi(n) are relatively prime but phi(n) is not prime, and e is public
* The general case where we have no idea as to the structure of the modulus, and we don't know if it is public so must treat it as secret

Previously all of these went through `inverse_mod` which prevented any possible optimizations based on cases.

Add a new (internal) header which directly exposes the various cases and apply them within the codebase.

The only new algorithm implemented (so far) is Arazi's algorithm for inversion of a prime modulo a non-prime. (TIL) Specifically this is a special case for computing the RSA secret exponent when `e=65537`. (Would also work for `e=3`/`e=17`/etc but this is not implemented)

### DIFF
--- a/src/lib/math/numbertheory/mod_inv.cpp
+++ b/src/lib/math/numbertheory/mod_inv.cpp
@@ -1,11 +1,12 @@
 /*
-* (C) 1999-2011,2016,2018,2019,2020 Jack Lloyd
+* (C) 1999-2011,2016,2018,2019,2020,2025 Jack Lloyd
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
 
-#include <botan/numthry.h>
+#include <botan/internal/mod_inv.h>
 
+#include <botan/numthry.h>
 #include <botan/internal/ct_utils.h>
 #include <botan/internal/divide.h>
 #include <botan/internal/mp_core.h>
@@ -17,10 +18,10 @@ namespace {
 
 BigInt inverse_mod_odd_modulus(const BigInt& n, const BigInt& mod) {
    // Caller should assure these preconditions:
-   BOTAN_DEBUG_ASSERT(n.is_positive());
-   BOTAN_DEBUG_ASSERT(mod.is_positive());
-   BOTAN_DEBUG_ASSERT(n < mod);
-   BOTAN_DEBUG_ASSERT(mod >= 3 && mod.is_odd());
+   BOTAN_ASSERT_NOMSG(n.is_positive());
+   BOTAN_ASSERT_NOMSG(mod.is_positive());
+   BOTAN_ASSERT_NOMSG(n < mod);
+   BOTAN_ASSERT_NOMSG(mod >= 3 && mod.is_odd());
 
    /*
    This uses a modular inversion algorithm designed by Niels Möller
@@ -98,10 +99,7 @@ BigInt inverse_mod_odd_modulus(const BigInt& n, const BigInt& mod) {
       bigint_cnd_add(odd_u, u_w, mp1o2, mod_words);
    }
 
-   auto a_is_0 = CT::Mask<word>::set();
-   for(size_t i = 0; i != mod_words; ++i) {
-      a_is_0 &= CT::Mask<word>::is_zero(a_w[i]);
-   }
+   const auto a_is_0 = CT::all_zeros(a_w, mod_words);
 
    auto b_is_1 = CT::Mask<word>::is_equal(b_w[0], 1);
    for(size_t i = 1; i != mod_words; ++i) {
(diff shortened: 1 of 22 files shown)


src/lib/math/numbertheory/mod_inv.cpp
```diff
@@ -277,4 +279,100 @@ BigInt inverse_mod(const BigInt& n, const BigInt& mod) {
...
+      */
+
+      const word phi_mod_e = ct_mod_word(phi_n, 65537);
+
+      // Compute phi^-1 mod e using FLT
+
+      const word inv_phi_mod_e = [&]() {
+         // Need 64-bit here as accum*accum exceeds 32-bit if accum=0x10000
+         uint64_t accum = 1;
+         for(size_t i = 0; i != 16; ++i) {
```
randombit, line 332: TODO here - these reductions on x86 and ARM compile down to just multiplication by a constant, shifts/adds, etc. But on RISCV it branches to some runtime function which is likely not constant time.

randombit: Fixed to use a word-length Barrett reduction

src/lib/math/numbertheory/mod_inv.cpp
```diff
@@ -277,4 +277,117 @@ BigInt inverse_mod(const BigInt& n, const BigInt& mod) {
...
+   return inverse_mod_odd_modulus(x, p);
+}
+
+BigInt inverse_modulo_public_prime(const BigInt& x, const BigInt& p) {
+   return inverse_modulo_secret_prime(x, p);
+}
+
+BigInt inverse_modulo_rsa_public_modulus(const BigInt& x, const BigInt& n) {
+   BOTAN_ARG_CHECK(n.is_positive() && n.is_odd(), "RSA public modulus must be odd and positive");
+   BOTAN_ARG_CHECK(x.is_positive() and x < n, "Input must be positive and less than RSA modulus");
```
reneme, line 297:
```suggestion
   BOTAN_ARG_CHECK(x.is_positive() && x < n, "Input must be positive and less than RSA modulus");
```

Didn't even know this worked in the general case.

reneme (approved):
Actual change makes sense to us (/cc @FAlbertDev). A few minor style suggestions below.

src/lib/math/numbertheory/mod_inv.cpp
randombit: Just you wait I'm going to start defining move constructors as `MyClass(MyClass and other)` :imp: 

j/k better to be consistent, I didn't do this intentionally, just a mental autocomplete with Python on my mind I guess.

src/tests/test_bigint.cpp
```diff
@@ -666,19 +667,26 @@ class BigInt_InvMod_Test final : public Text_Based_Test {
...
-         else if((a % mod) > 0)
-            {
-            result.confirm("no inverse with gcd > 1", gcd(a, mod) > 1);
+            if(Botan::is_prime(mod, rng()) && mod != 2) {
+               BOTAN_ASSERT_NOMSG(expected > 0);
+               result.test_eq("inverse_modulo_secret_prime", Botan::inverse_modulo_secret_prime(a, mod), expected);
+
+               // Currently inverse_modulo_public_prime just forwards to inverse_modulo_secret_prime so we'll
+               // skip this test until a specific algorithm is implemented for this function
+               //result.test_eq("inverse_modulo_public_prime", Botan::inverse_modulo_public_prime(a, mod), expected);
```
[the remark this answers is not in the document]
randombit: Yeah fair

src/lib/math/numbertheory/mod_inv.cpp
reneme: # :fearful:

src/tests/test_bigint.cpp
```diff
@@ -744,6 +752,58 @@ class Lucas_Primality_Test final : public Test {
 
 BOTAN_REGISTER_TEST("math", "bn_lucas", Lucas_Primality_Test);
 
+class RSA_Compute_Exp_Test : public Test {
```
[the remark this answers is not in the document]
randombit: Likely yes. In fact we should do this for all of our (many) Monte Carlo tests. And likewise most of the fuzzers we have should also be tests that run as part of the regular test suite. Some of them do already exist in both systems, but only a few. I'll create a ticket describing the situation.

src/tests/tests.cpp
```diff
@@ -842,6 +842,17 @@ std::string Test::random_password(Botan::RandomNumberGenerator& rng) {
...
 
+size_t Test::random_index(Botan::RandomNumberGenerator& rng, size_t max) {
+   std::array<uint8_t, 8> v;
+   rng.randomize(v);
+
+   uint64_t x = 0;
+   for(size_t i = 0; i != v.size(); ++i) {
+      x = (x << 8) + v[i];
+   }
+   return static_cast<size_t>(x % max);
```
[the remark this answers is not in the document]
randombit: Yeah I just wanted to avoid the include of an internal header there I guess, but there is no real motivation for that

src/tests/test_bigint.cpp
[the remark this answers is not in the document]
randombit: #4576

erikloman:
The following line does not compile on VS2022 due to the "and" statement which I assume should be &&.
Note: I am compiling a debug build.

```
BigInt inverse_mod_secret_prime(const BigInt& x, const BigInt& p) {
   BOTAN_ARG_CHECK(x.is_positive() && p.is_positive(), "Parameters must be positive");
   BOTAN_ARG_CHECK(x < p, "x must be less than p");
   BOTAN_ARG_CHECK(p.is_odd() and p > 1, "Primes are odd integers greater than 1");  // <---- this does not compile due to "and"; I assume it must be &&

   // TODO possibly use FLT, or the algorithm presented for this case in
   // Handbook of Elliptic and Hyperelliptic Curve Cryptography

   return inverse_mod_odd_modulus(x, p);
}
```

PR #4572, opened 2025-01-19
Decouple support for explicit curves with legacy_ec_point

randombit:
Both are deprecated but for different reasons, and in the short term it's reasonable to want to disable legacy_ec_point while still supporting explicit curve encodings.

### DIFF
--- a/src/lib/pubkey/ec_group/ec_group.cpp
+++ b/src/lib/pubkey/ec_group/ec_group.cpp
@@ -13,6 +13,7 @@
 #include <botan/ber_dec.h>
 #include <botan/der_enc.h>
 #include <botan/mutex.h>
+#include <botan/numthry.h>
 #include <botan/pem.h>
 #include <botan/reducer.h>
 #include <botan/rng.h>
@@ -219,8 +220,6 @@ std::pair<std::shared_ptr<EC_Group_Data>, bool> EC_Group::BER_decode_EC_group(st
 
       return std::make_pair(data, false);
    } else if(next_obj_type == ASN1_Type::Sequence) {
-#if defined(BOTAN_HAS_LEGACY_EC_POINT)
-
       BigInt p, a, b, order, cofactor;
       std::vector<uint8_t> base_pt;
       std::vector<uint8_t> seed;
@@ -246,7 +245,8 @@ std::pair<std::shared_ptr<EC_Group_Data>, bool> EC_Group::BER_decode_EC_group(st
          throw Decoding_Error("ECC p parameter is invalid size");
       }
 
-      if(p.is_negative() || !is_bailie_psw_probable_prime(p)) {
+      Modular_Reducer mod_p(p);
+      if(p.is_negative() || !is_bailie_psw_probable_prime(p, mod_p)) {
          throw Decoding_Error("ECC p parameter is not a prime");
       }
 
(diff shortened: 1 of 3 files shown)


src/lib/pubkey/ec_group/ec_group.cpp
```diff
@@ -266,14 +266,46 @@ std::pair<std::shared_ptr<EC_Group_Data>, bool> EC_Group::BER_decode_EC_group(st
...
+
+      auto [g_x, g_y] = [&]() {
+         const uint8_t hdr = base_pt[0];
+
+         if(hdr == 0x04 && base_pt.size() == 1 + 2 * p_bytes) {
+            BigInt x = BigInt::decode(&base_pt[1], p_bytes);
+            BigInt y = BigInt::decode(&base_pt[p_bytes + 1], p_bytes);
+
+            if(x < p && y < p) {
+               return std::make_pair(x, y);
```
reneme, line 282: Performance nit: I'm quite sure, for optimal performance you'd have to `std::move()` x and y into the result pair (also in the else-if branch). Alternatively, you could pre-initialize a result pair in the beginning and decode the `BigInt`s straight into `.first` and `.second`. Then RVO should do the trick.

src/lib/pubkey/ec_group/legacy_ec_point/ec_point.h
```diff
@@ -20,9 +22,13 @@ namespace Botan {
...
 *
-* Use EC_AffinePoint in new code
+* Use EC_AffinePoint in new code; this type is no longer used internally at all
+* except to support very unfortunate (and deprecated) curve types, specifically
+* those with a cofactor, or with unreasonable sizes (above 521 bits), which
+* cannot be accomodated by the new faster EC library in math/pcurves. For
+* normal curves EC_AffinePoint will typically be 2 or 3 times faster.
 *
-* This type will be removed/hidden in Botan4
+* This type will be completely removed in Botan4
```
reneme, line 31: Perhaps use the `@deprecated` doxygen markup?

reneme (approved):
Just some nits.

src/lib/pubkey/ec_group/legacy_ec_point/ec_point.h
randombit: Clang's `-Wdocumentation-deprecated-sync` warns if a type is documented with `@deprecated` but not `[[deprecated]]` (or equiv attribute). Then if deprecated warnings are disabled (as they are when building the library itself), Clang warns about them being inconsistent. But if we don't disable the deprecated warnings, we'll receive deprecated warnings eg when testing the deprecated thing, or when that thing is used in a header to implement some other (usually also deprecated) function.

And XCode defaults to `-Wdocumentation` being on. AND apparently XCode emits the warning even for third party/system headers (why???) https://stackoverflow.com/questions/24453188/disable-documentation-comments-warning-for-selected-files https://stackoverflow.com/questions/42991455/silencing-documentation-issue-warnings-in-xcode

Notice `@deprecated` is not currently used anywhere in the code... sometimes the only winning move is not to play.

src/lib/pubkey/ec_group/ec_group.cpp
randombit: Fair enough but considering what else is happening here - at best just testing the curve equation and possibly a point decompression, a couple of extra copies seems negligible.

src/lib/pubkey/ec_group/legacy_ec_point/ec_point.h
randombit: I'd be open to some way of fixing this, obviously it would be better if the Doxygen comments reflected the state of things, but we seem to be trapped in a corner by Clang's behavior. Best I can come up with is using `@warning Deprecated`

reneme: Yep, concur. Thanks. Let's not do that.

Issue #1019, opened 2017-04-21
ECDSA performance


bellgrim:
Following up an issue reported here: https://github.com/opendnssec/SoftHSMv2/issues/213

Did some performance tests of ECDSA and compared it with OpenSSL. There are some significant differences. I compiled Botan with no additional flags. Should I add any to get better performance or is the issue related to https://github.com/randombit/botan/issues/664#issuecomment-253211974 ?

Number of signatures per second using a single thread:

Algo|OpenSSL|Botan
----|----------|------
secp256r1 | 17472.3 | 213
secp384r1 | 4067.4 | 140
secp521r1 | 1951.9 | 98

Full output from the performance tests:
```
$ openssl version
OpenSSL 1.1.0e  16 Feb 2017

$ botan version
2.1.0

$ openssl speed ecdsap256 ecdsap384 ecdsap521
Doing 256 bit sign ecdsa's for 10s: 174898 256 bit ECDSA signs in 10.01s 
Doing 256 bit verify ecdsa's for 10s: 74683 256 bit ECDSA verify in 10.00s
Doing 384 bit sign ecdsa's for 10s: 40715 384 bit ECDSA signs in 10.01s 
Doing 384 bit verify ecdsa's for 10s: 9782 384 bit ECDSA verify in 10.01s
Doing 521 bit sign ecdsa's for 10s: 19539 521 bit ECDSA signs in 10.01s 
Doing 521 bit verify ecdsa's for 10s: 4798 521 bit ECDSA verify in 10.01s
                              sign    verify    sign/s verify/s
 256 bit ecdsa (nistp256)   0.0001s   0.0001s  17472.3   7468.3
 384 bit ecdsa (nistp384)   0.0002s   0.0010s   4067.4    977.2
 521 bit ecdsa (nistp521)   0.0005s   0.0021s   1951.9    479.3

$ botan speed --msec=10000 ECDSA
ECDSA-secp256r1 305 keygen/sec; 3.28 ms/op 7848939 cycles/op (1 op in 3 ms)
ECDSA-secp256r1 213 EMSA1(SHA-256) sign/sec; 4.68 ms/op 11228207 cycles/op (2140 ops in 10005 ms)
ECDSA-secp256r1 418 EMSA1(SHA-256) verify/sec; 2.39 ms/op 5742684 cycles/op (4186 ops in 10002 ms)
[...]

PR #4578, opened 2025-01-20
Switch the TLS server fuzzer to use an ECDSA key

randombit:
This improves fuzzer iterations per second by roughly a factor of 10.


randombit:
I imagine the performance difference was a lot less (or perhaps even RSA was strictly faster than ECDSA) back when this fuzzer was written, some historical numbers in #1019

PR #4584, opened 2025-01-22
Fix compilation on GCC 15

lanurmi:
`<cstdint>` is needed for `uint8_t`.


randombit (approved):
Change is fine but be aware that Botan2 is end of life and no further releases are expected

lanurmi:
> Change is fine but be aware that Botan2 is end of life and no further releases are expected

I'm completely fine with that. Thanks!

PR #4586, opened 2025-01-23
Update pubkey.rst

randombit:
Along with a few doc comments in pubkey.h

### DIFF
--- a/doc/api_ref/pubkey.rst
+++ b/doc/api_ref/pubkey.rst
@@ -38,7 +38,6 @@ removed in a future major release.
 
      Check if this key could be used for the queried operation type.
 
-
 .. cpp:class:: Public_Key
 
    .. cpp:function:: size_t key_length() const = 0;
@@ -82,11 +81,47 @@ removed in a future major release.
 
    .. cpp:function:: std::vector<uint8_t> subject_public_key() const;
 
-      Return the X.509 SubjectPublicKeyInfo encoding of this key
+      Return the X.509 ``SubjectPublicKeyInfo`` encoding of this key. See
+      :rfc:`5280` for details.
 
    .. cpp:function:: std::string fingerprint_public(const std::string& alg = "SHA-256") const;
 
       Return a hashed fingerprint of this public key.
+
+.. cpp:class:: Private_Key
+
+   .. cpp:function:: std::unique_ptr<Public_Key> public_key() const
+
+      Return an object containing the public key corresponding to this private key.
+
+      Prefer this over the (deprecated) implicit conversion of a private key to
+      a public key currently possible due to an inheritence relation.
+
+   .. cpp:function:: secure_vector<uint8_t> private_key_info() const
+
+      Return the key encoded as a PKCS #8 `PrivateKeyInfo` structure. See
+      :rfc:`5208` for details.
+
+      Further functions relating to encoding and encrypting PKCS #8 private are
+      detailed in :ref:`serializing_private_keys`.
+
+   .. cpp:function:: secure_vector<uint8_t> private_key_bits() const
+
+      Return the serialization of the private key, cooresponding to the
+      `PrivateKey` field of a PKCS #8 `PrivateKeyInfo` structure. See
+      :rfc:`5208` for details.
+
+   .. cpp:function:: bool stateful_operation() const;
+
+      Returns true if this keys operation is stateful, that is if updating the
+      key is required after each private operation. Currently the only stateful
+      schemes included are XMSS and LMS.
+
+   .. cpp:function:: std::optional<uint64_t> remaining_operations() const
+
+      If this algorithm is stateful, returns the number of private operations
+      remaining before this key is exhausted. Returns `nullopt` if the key is
+      not stateful.
 
 Public Key Algorithms
 ------------------------
@@ -176,7 +211,8 @@ HSS-LMS
 
 A post-quantum secure hash-based signature scheme similar to XMSS. Contains
 support for multitrees. It is stateful, meaning the private key changes after
-each signature.
+each signature. If the same state is ever used to generate two signatures, then
+the whole scheme becomes insecure, and signatures can be forged.
 
 SLH-DSA (FIPS 205)
 ~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


doc/api_ref/pubkey.rst
```diff
@@ -82,12 +81,48 @@ removed in a future major release.
...
 
    .. cpp:function:: std::string fingerprint_public(const std::string& alg = "SHA-256") const;
 
       Return a hashed fingerprint of this public key.
 
+.. cpp:class:: Private_Key
+
+   .. cpp:function:: std::unique_ptr<Public_Key> public_key() const
+
+      Return an object containing the public key cooresponding to this private key.
```
reneme, line 95:
```suggestion
      Return an object containing the public key corresponding to this private key.
```

doc/api_ref/pubkey.rst
```diff
@@ -82,12 +81,48 @@ removed in a future major release.
...
+
+      Prefer this over the (deprecated) implicit conversion of a private key to
+      a public key currently possible due to an inheritence relation.
+
+   .. cpp:function secure_vector<uint8_t> private_key_info() const
```
reneme, line 100:
```suggestion
   .. cpp:function:: secure_vector<uint8_t> private_key_info() const
```

doc/api_ref/pubkey.rst
```diff
@@ -82,12 +81,48 @@ removed in a future major release.
...
+
+      Return the key encoded as a PKCS #8 `PrivateKeyInfo` structure. See
+      :rfc:`5208` for details.
+
+      Further functions relating to encoding and encrypting PKCS #8 private are
+      detailed in :ref:`serializing_private_keys`.
+
+   .. cpp:function secure_vector<uint8_t> private_key_bits() const
```
reneme, line 108:
```suggestion
   .. cpp:function:: secure_vector<uint8_t> private_key_bits() const
```

doc/api_ref/pubkey.rst
```diff
@@ -881,9 +1039,14 @@ signature is validated.
...
 
-Available signature padding schemes
+RSA signature padding schemes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+These signature padding mechanisms are specific to RSA; no other public
+key algorithms included in Botan make use of then. For historical reasons,
+many different padding schemes have been defined for RSA over the years.
+The most common are PSS and the (now obsolete) PKCS1v15.
+
```
reneme, line 1049: Thanks! I think this is really crucial information for crypto newcomers!

doc/api_ref/pubkey.rst
```diff
@@ -894,6 +1057,11 @@ PKCS #1 v1.5 Type 1 (signature)
 
 PKCS #1 v1.5 Type 1 (signature) padding, aka EMSA3 in IEEE 1363.
 
+.. note::
+
+   While note as actively unsafe as PKCS1v15 encryption padding is,
+   PKCS1 signature padding is considered quite obsolete.
```
reneme, line 1063:
```suggestion
.. note::

   While not as actively unsafe as PKCS1v15 encryption padding is,
   PKCS1 signature padding is considered quite obsolete.
```

doc/api_ref/pubkey.rst
```diff
@@ -938,6 +1106,9 @@ Don't use this unless you know what you are doing.
 ISO-9796-2
 """"""""""
 
+The ISO-9796-2 padding schemes are used for EMV signatures. There
+is likely no reason to use it in other contexts.
+
```
reneme, line 1111: Perhaps `-v` on EMV signatures? I've never come across this, FWIW.

doc/api_ref/pubkey.rst
```diff
@@ -979,8 +1154,13 @@ EMSA from X9.31 (EMSA2 in IEEE 1363).
...
 
-Sign inputs directly.
-Don't use this unless you know what you are doing.
+Sign inputs directly with no hashing or padding
+
+.. warning::
+
+   This exists as an escape hatch allowing an application to define
+   some protocol-specific padding scheme. Don't use this unless you
+   know what you are doing.
```
reneme, line 1163: Thinking out loud: once the `PK_Options` (#4318) are fully operational, we could potentially replace this escape hatch with the ability for a user to create their own subclass of `EMSA` to build application-specific paddings. And then pass it along as `.with_padding(ptr_to_an_object_of_my_emsa)`. (Assuming that `class EMSA` is even public API.)

That makes it more involved to YOLO your own padding, with more boilerplate. But perhaps that's actually a good thing.

doc/api_ref/pubkey.rst
```diff
@@ -1021,6 +1206,11 @@ Parameters specification:
 Ed25519 and Ed448 Variants
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. warning:: Ed25519 and Ed448 have different verification criteria, depending
+   on the implementation. This can be problematic in systems which rely on
+   consensus - see `It’s 255:19AM. Do you know what your validation criteria
+   are? <https://hdevalence.ca/blog/2020-10-04-its-25519am>`_ for details.
+
```
reneme, line 1213: For consistency 💪 

```suggestion
.. warning::

   Ed25519 and Ed448 have different verification criteria, depending
   on the implementation. This can be problematic in systems which rely on
   consensus - see `It’s 255:19AM. Do you know what your validation criteria
   are? <https://hdevalence.ca/blog/2020-10-04-its-25519am>`_ for details.

```

doc/api_ref/pubkey.rst
```diff
@@ -1089,34 +1279,34 @@ produce an output of the desired length.
...
 
   .. cpp:function:: SymmetricKey derive_key(size_t key_len, \
-                    const std::span<const uint8_t> in, \
-                    const std::string& params = "") const
+                    std::span<const uint8_t> peer_key, \
+                    const std::string& salt = "") const
 
-     Return a shared key. The *params* will be hashed along with the shared secret by the
+     Return a shared key. The *salt* will be hashed along with the shared secret by the
      KDF; this can be useful to bind the shared secret to a specific usage.
```
reneme, line 1300: ... if no KDF was defined, the salt and the key length are typically ignored. (Or a formulation along those lines.)

I actually find this to be an API pitfall, because users might have designed their protocol around the idea of injecting that context while it fact it is just silently ignored. Here, a to-be-built API along the lines of `PK_Options` (#4318) might come in handy. It would make it easy to throw an exception if `.with_salt()` (or `.with_output_length()`) was given but no KDF actually ended up consuming those values in `derive_key()`.

doc/api_ref/pubkey.rst
```diff
@@ -1089,34 +1279,34 @@ produce an output of the desired length.
...
   .. cpp:function:: SymmetricKey derive_key(size_t key_len, \
-                    std::span<const uint8_t> in, \
-                    const uint8_t params[], size_t params_len) const
+                    std::span<const uint8_t> peer_key, \
+                    const uint8_t salt[], size_t salt_len) const
 
   .. cpp:function:: SymmetricKey derive_key(size_t key_len, \
-                    const uint8_t in[], size_t in_len, \
-                    const std::string& params = "") const
+                    const uint8_t peer_key[], size_t peer_key_len_len, \
```
reneme, line 1292:
```suggestion
                    const uint8_t peer_key[], size_t peer_key_len, \
```

doc/api_ref/pubkey.rst
```diff
@@ -1213,10 +1403,10 @@ encapsulated key and returns the shared secret.
 Botan implements the following KEM schemes:
 
 1. RSA
-#. ML-KEM (Kyber)
+#. ML-KEM (aka Kyber)
```
reneme, line 1406:
```suggestion
#. ML-KEM (formerly known as Kyber)
```

... not sure how long until a typical user wouldn't know what Kyber stands for. Or can anyone type out "Rijndael" without googling "AES| to check for the correct spelling? 🤡

reneme (approved):
Lot's of important/useful information. Thanks a lot! Below are some suggestions and typo nits.

doc/api_ref/pubkey.rst
randombit: This [EMV](https://en.wikipedia.org/wiki/EMV)

I expanded this out to "EMV contactless payment card system" so it's unambigious

doc/api_ref/pubkey.rst
randombit: Actually we already do reject a salt if a KDF is not being used but I agree with your point in the broad sense.

doc/api_ref/pubkey.rst
randombit: The situation was a lot more confused in earlier releases

[Pre 2.0] Rabin-Williams and Nyberg-Rueppel could use the same padding schemes as RSA

"EMSA1" (aka the plain hash with possible truncation) was a real padding scheme that was used with *DSA and could be used with RSA as well.

Now things are pretty clearly bifurcated to "RSA specific" and "everyone else"

doc/api_ref/pubkey.rst
randombit: Updated to describe how keylen param is ignored without a kdf

PR #4587, opened 2025-01-23
Split blinding into a submodule

randombit:
Gets us a bit closer to being able to compile `pubkey` entirely without `BigInt`


src/lib/pubkey/blinding/info.txt
```diff
@@ -0,0 +1,17 @@
+<defines>
+PUBLIC_KEY_BLINDING -> 20250125
+</defines>
+
+<module_info>
+name -> "Public Key Blinding"
+brief -> "Helper for BigInt blinding"
+</module_info>
```
reneme, line 8:
```suggestion
<module_info>
name -> "Public Key Blinding"
brief -> "Helper for BigInt blinding"
type -> "Internal"
</module_info>
```

Mark this as internal?

reneme (approved):

PR #4588, opened 2025-01-23
Reduce overhead from computing modular reduction parameters

randombit:
This was inspired by an OpenSSL bug; apparently there is/was something wrong with their PKCS8 decoders in OpenSSL 3, and parsing private keys took a very long time. I wrote a test to check how Botan behaved and happily the PKCS8 decoding seems to be fine but it took a very long time to set up the Montgomery arithmetic. Specifically computing the Montgomery params requires setting up a Barrett reduction, which required a constant time division. The constant time division consumed well over 90% of the total runtime when parsing a 4096 bit RSA key in a loop.

First optimize the initial Barrett setup by adding a specialized implementation for `2^k / m`. I had hoped to find a specific algorithm - seems like something should be possible here - but Google is useless. But it's anyway a bit faster because we know only 1 bit is set and we can assume `k` is public.

Second commit follows up by distinguishing between public and secret moduli for Barrett, much as #4569 did for modular inverses. In the public modulus case we can use the much faster variable time division. This also puts some effort towards sharing Barrett constants once we've computed them, eg by making them available from `DL_Group` and caching them in the RSA key data.

Overall this reduces the cost of repeated parsing of a RSA 4096 bit private key to about 1/4 of current cost on master.

### DIFF
--- a/src/lib/math/bigint/divide.cpp
+++ b/src/lib/math/bigint/divide.cpp
@@ -63,7 +63,7 @@ void ct_divide(const BigInt& x, const BigInt& y, BigInt& q_out, BigInt& r_out) {
       const size_t b = x_bits - 1 - i;
       const bool x_b = x.get_bit(b);
 
-      r *= 2;
+      r <<= 1;
       r.conditionally_set_bit(0, x_b);
 
       const bool r_gte_y = bigint_sub3(t.mutable_data(), r._data(), r.size(), y._data(), y_words) == 0;
@@ -75,6 +75,46 @@ void ct_divide(const BigInt& x, const BigInt& y, BigInt& q_out, BigInt& r_out) {
    sign_fixup(x, y, q, r);
    r_out = r;
    q_out = q;
+}
+
+BigInt ct_divide_pow2k(size_t k, const BigInt& y) {
+   BOTAN_ARG_CHECK(!y.is_zero(), "Cannot divide by zero");
+   BOTAN_ARG_CHECK(!y.is_negative(), "Negative divisor not supported");
+   BOTAN_ARG_CHECK(k > 1, "Invalid k");
+
+   const size_t x_bits = k + 1;
+   const size_t y_bits = y.bits();
+
+   if(x_bits < y_bits) {
+      return BigInt::zero();
+   }
+
+   BOTAN_ASSERT_NOMSG(y_bits >= 1);
+   const size_t x_words = (x_bits + BOTAN_MP_WORD_BITS - 1) / BOTAN_MP_WORD_BITS;
+   const size_t y_words = y.sig_words();
+
+   BigInt q = BigInt::with_capacity(x_words);
+   BigInt r = BigInt::with_capacity(y_words + 1);
+   BigInt t = BigInt::with_capacity(y_words + 1);  // a temporary
+
+   r.set_bit(y_bits - 1);
+   for(size_t i = y_bits - 1; i != x_bits; ++i) {
+      const size_t b = x_bits - 1 - i;
+
+      if(i >= y_bits) {
+         bigint_shl1(r.mutable_data(), r.size(), r.size(), 1);
+      }
+
+      const bool r_gte_y = bigint_sub3(t.mutable_data(), r._data(), r.size(), y._data(), y_words) == 0;
+
+      q.conditionally_set_bit(b, r_gte_y);
+
+      bigint_cnd_swap(static_cast<word>(r_gte_y), r.mutable_data(), t.mutable_data(), y_words + 1);
+   }
+
+   // No need for sign fixup
+
+   return q;
 }
 
 void ct_divide_word(const BigInt& x, word y, BigInt& q_out, word& r_out) {
@@ -94,7 +134,7 @@ void ct_divide_word(const BigInt& x, word y, BigInt& q_out, word& r_out) {
 
       const auto r_carry = CT::Mask<word>::expand_top_bit(r);
 
-      r *= 2;
+      r <<= 1;
       r += x_b;
 
       const auto r_gte_y = CT::Mask<word>::is_gte(r, y) | r_carry;
@@ -128,7 +168,7 @@ word ct_mod_word(const BigInt& x, word y) {
 
       const auto r_carry = CT::Mask<word>::expand_top_bit(r);
 
-      r *= 2;
+      r <<= 1;
       r += x_b;
 
       const auto r_gte_y = CT::Mask<word>::is_gte(r, y) | r_carry;
@@ -154,7 +194,7 @@ BigInt ct_modulo(const BigInt& x, const BigInt& y) {
       const size_t b = x_bits - 1 - i;
       const bool x_b = x.get_bit(b);
 
-      r *= 2;
+      r <<= 1;
       r.conditionally_set_bit(0, x_b);
 
       const bool r_gte_y = bigint_sub3(t.mutable_data(), r._data(), r.size(), y._data(), y_words) == 0;
--- a/src/lib/math/numbertheory/reducer.cpp
+++ b/src/lib/math/numbertheory/reducer.cpp
@@ -29,8 +29,10 @@ Modular_Reducer::Modular_Reducer(const BigInt& mod) {
       m_mod_words = m_modulus.sig_words();
 
       // Compute mu = floor(2^{2k} / m)
-      m_mu.set_bit(2 * BOTAN_MP_WORD_BITS * m_mod_words);
-      m_mu = ct_divide(m_mu, m_modulus);
+      const size_t mu_bits = 2 * BOTAN_MP_WORD_BITS * m_mod_words;
+      m_mu.set_bit(mu_bits);
+      m_mu = ct_divide_pow2k(mu_bits, m_modulus);
+      //m_mu = BigInt::power_of_2(mu_bits) / m_modulus;
    }
 }
 
--- a/src/lib/math/numbertheory/reducer.cpp
+++ b/src/lib/math/numbertheory/reducer.cpp
@@ -25,15 +25,30 @@ Modular_Reducer::Modular_Reducer(const BigInt& mod) {
    m_mod_words = 0;
 
    if(mod > 0) {
-      m_modulus = mod;
-      m_mod_words = m_modulus.sig_words();
-
-      // Compute mu = floor(2^{2k} / m)
-      const size_t mu_bits = 2 * BOTAN_MP_WORD_BITS * m_mod_words;
-      m_mu.set_bit(mu_bits);
-      m_mu = ct_divide_pow2k(mu_bits, m_modulus);
-      //m_mu = BigInt::power_of_2(mu_bits) / m_modulus;
+      *this = Modular_Reducer::for_secret_modulus(mod);
    }
+}
+
+Modular_Reducer Modular_Reducer::for_secret_modulus(const BigInt& mod) {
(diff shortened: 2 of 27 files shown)


src/lib/math/numbertheory/reducer.h
```diff
@@ -64,11 +64,26 @@ class BOTAN_PUBLIC_API(2, 0) Modular_Reducer final {
...
+      */
+      static Modular_Reducer for_public_modulus(const BigInt& m);
+
+      /**
+      * Requires that m > 0
+      */
+      static Modular_Reducer for_secret_modulus(const BigInt& m);
 
    private:
+      Modular_Reducer(const BigInt& m, const BigInt& mu, size_t mw) : m_modulus(m), m_mu(mu), m_mod_words(mw) {}
```
reneme, line 85:
```suggestion
      Modular_Reducer(BigInt m, BigInt mu, size_t mw) : m_modulus(std::move(m)), m_mu(std::move(mu)), m_mod_words(mw) {}
```

src/lib/math/numbertheory/reducer.cpp
```diff
@@ -25,15 +25,36 @@ Modular_Reducer::Modular_Reducer(const BigInt& mod) {
...
+   BOTAN_ARG_CHECK(!mod.is_negative(), "Modulus cannot be negative");
+
+   // Left uninitialized if mod == 0
+   size_t mod_words = mod.sig_words();
+
+   // Compute mu = floor(2^{2k} / m)
+   const size_t mu_bits = 2 * BOTAN_MP_WORD_BITS * mod_words;
+   BigInt mu = BigInt::power_of_2(mu_bits) / mod;
+   return Modular_Reducer(mod, mu, mod_words);
+}
```
reneme, line 56: Let's avoid allocations and copies as much as we can. Similarly for the private variant.

```suggestion
Modular_Reducer Modular_Reducer::for_public_modulus(BigInt mod) {
   BOTAN_ARG_CHECK(!mod.is_zero(), "Modulus cannot be zero");
   BOTAN_ARG_CHECK(!mod.is_negative(), "Modulus cannot be negative");

   // Left uninitialized if mod == 0
   size_t mod_words = mod.sig_words();

   // Compute mu = floor(2^{2k} / m)
   const size_t mu_bits = 2 * BOTAN_MP_WORD_BITS * mod_words;
   BigInt mu = BigInt::power_of_2(mu_bits) / mod;
   return Modular_Reducer(std::move(mod), std::move(mu), mod_words);
}
```

src/lib/pubkey/blinding/blinding.cpp
```diff
@@ -31,10 +32,6 @@ BigInt Blinder::blinding_nonce() const {
 }
 
 BigInt Blinder::blind(const BigInt& i) const {
-   if(!m_reducer.initialized()) {
-      throw Invalid_State("Blinder not initialized, cannot blind");
-   }
```
reneme, line 36: Less state, more happiness! 👍

src/lib/pubkey/blinding/blinding.cpp
```diff
@@ -10,10 +10,11 @@
 namespace Botan {
 
 Blinder::Blinder(const BigInt& modulus,
+                 const Modular_Reducer& reducer,
```
reneme, line 13: pass-by-value and `std::move()` into the member, please. :)

src/lib/pubkey/dl_group/dl_group.cpp
```diff
@@ -39,6 +39,18 @@ class DL_Group_Data final {
             m_exponent_bits(dl_exponent_size(m_p_bits)),
             m_source(source) {}
 
+      DL_Group_Data(const BigInt& p, const BigInt& g, DL_Group_Source source) :
+            m_p(p),
+            m_g(g),
```
reneme, line 44:
```suggestion
      DL_Group_Data(BigInt p, BigInt g, DL_Group_Source source) :
            m_p(std::move(p)),
            m_g(std::move(g)),
```

... also in the pre-existing constructor above.

src/lib/pubkey/dl_group/dl_group.cpp
```diff
@@ -120,23 +129,24 @@ std::shared_ptr<DL_Group_Data> DL_Group::BER_decode_DL_group(const uint8_t data[
...
+      return std::make_shared<DL_Group_Data>(p, q, g, source);
    } else if(format == DL_Group_Format::ANSI_X9_42) {
+      BigInt p, g, q;
       ber.decode(p).decode(g).decode(q).discard_remaining();
+      return std::make_shared<DL_Group_Data>(p, q, g, source);
    } else if(format == DL_Group_Format::PKCS_3) {
-      // q is left as zero
+      BigInt p, g;
       ber.decode(p).decode(g).discard_remaining();
+      return std::make_shared<DL_Group_Data>(p, g, source);
```
reneme, line 146: Ugly, but it is parsing code, after all...

```C++
if(format == DL_Group_Format::ANSI_X9_57) {
   BigInt p, q, g;
   ber.decode(p).decode(q).decode(g).verify_end();
   return std::make_shared<DL_Group_Data>(std::move(p), std::move(q), std::move(g), source);
} else if(format == DL_Group_Format::ANSI_X9_42) {
   BigInt p, g, q;
   ber.decode(p).decode(g).decode(q).discard_remaining();
   return std::make_shared<DL_Group_Data>(std::move(p), std::move(q), std::move(g), source);
} else if(format == DL_Group_Format::PKCS_3) {
   BigInt p, g;
   ber.decode(p).decode(g).discard_remaining();
   return std::make_shared<DL_Group_Data>(std::move(p), std::move(g), source);
}
```

src/lib/pubkey/dl_group/dl_group.cpp
```diff
@@ -145,7 +155,11 @@ std::shared_ptr<DL_Group_Data> DL_Group::load_DL_group_info(const char* p_str, c
...
 
-   return std::make_shared<DL_Group_Data>(p, q, g, DL_Group_Source::Builtin);
+   if(q.is_zero()) {
+      return std::make_shared<DL_Group_Data>(p, g, DL_Group_Source::Builtin);
+   } else {
+      return std::make_shared<DL_Group_Data>(p, q, g, DL_Group_Source::Builtin);
+   }
 }
 
 //static
```
reneme, line 165: While we're here:

```C++
std::shared_ptr<DL_Group_Data> DL_Group::load_DL_group_info(const char* p_str, const char* g_str) {
   auto p = BigInt(p_str);
   auto q = (p - 1) / 2;
   auto g = BigInt(g_str);

   return std::make_shared<DL_Group_Data>(std::move(p), std::move(q), std::move(g), DL_Group_Source::Builtin);
}
```

src/lib/pubkey/dl_group/dl_group.cpp
```diff
@@ -324,14 +338,18 @@ DL_Group::DL_Group(RandomNumberGenerator& rng, const std::vector<uint8_t>& seed,
...
 * DL_Group Constructor
 */
 DL_Group::DL_Group(const BigInt& p, const BigInt& q, const BigInt& g) {
-   m_data = std::make_shared<DL_Group_Data>(p, q, g, DL_Group_Source::ExternalSource);
+   if(q.is_zero()) {
+      m_data = std::make_shared<DL_Group_Data>(p, g, DL_Group_Source::ExternalSource);
+   } else {
+      m_data = std::make_shared<DL_Group_Data>(p, q, g, DL_Group_Source::ExternalSource);
+   }
 }
```
reneme, line 353: https://www.youtube.com/shorts/NgsNi_mGe_A

reneme (approved):
Logic looks good to me. While we're on the topic of reducing overhead, I suggested plenty of places that could benefit from "pass-by-value" and `std::move()`. Perhaps this could be pushed into a follow-up PR, though. Your call... :)

src/lib/math/numbertheory/reducer.cpp
```diff
@@ -25,15 +25,36 @@ Modular_Reducer::Modular_Reducer(const BigInt& mod) {
...
+      *this = Modular_Reducer::for_secret_modulus(mod);
    }
 }
 
+Modular_Reducer Modular_Reducer::for_secret_modulus(const BigInt& mod) {
+   BOTAN_ARG_CHECK(!mod.is_zero(), "Modulus cannot be zero");
+   BOTAN_ARG_CHECK(!mod.is_negative(), "Modulus cannot be negative");
+
+   // Left uninitialized if mod == 0
+   size_t mod_words = mod.sig_words();
```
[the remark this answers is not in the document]
randombit: Copypasta from the previous constructor

src/lib/math/bigint/divide.cpp
```diff
@@ -77,6 +77,39 @@ void ct_divide(const BigInt& x, const BigInt& y, BigInt& q_out, BigInt& r_out) {
...
+   BOTAN_ASSERT_NOMSG(y_bits >= 1);
+   const size_t x_words = (x_bits + BOTAN_MP_WORD_BITS - 1) / BOTAN_MP_WORD_BITS;
+   const size_t y_words = y.sig_words();
+
+   BigInt q = BigInt::with_capacity(x_words);
+   BigInt r = BigInt::with_capacity(y_words + 1);
+   BigInt t = BigInt::with_capacity(y_words + 1);  // a temporary
+
+   r.set_bit(y_bits - 1);
+   for(size_t i = y_bits; i != x_bits; ++i) {
```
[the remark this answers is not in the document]
randombit: Yes it does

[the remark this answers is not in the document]
randombit: Actually we are allowed to (must) leak the size of the values to side channels so we can just check the bit lengths and correctly return zero if 2^k < y

src/lib/math/bigint/divide.cpp
```diff
@@ -94,7 +127,7 @@ void ct_divide_word(const BigInt& x, word y, BigInt& q_out, word& r_out) {
 
       const auto r_carry = CT::Mask<word>::expand_top_bit(r);
 
-      r *= 2;
+      r <<= 1;
```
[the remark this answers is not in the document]
randombit: I was just following the same pattern as elsewhere for consistency.

Also in this context - since we are accumulating into `r` bit-by-bit as the loop progresses - a shift is more logical than a multiplication.

src/lib/pubkey/blinding.cpp
```diff
@@ -10,10 +10,11 @@
 namespace Botan {
 
 Blinder::Blinder(const BigInt& modulus,
+                 const Modular_Reducer& reducer,
```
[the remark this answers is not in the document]
randombit: Here we already have a (somewhat undocumented) lifetime restriction on the `Blinder` - it captures a reference to the `RandomNumberGenerator&` and expects it to live as long as the `Blinder` itself does. We can do the same for the `Modular_Reducer&`

src/lib/math/numbertheory/reducer.cpp
randombit: Done for `mu` there is no point with the modulus imo

src/lib/pubkey/dl_group/dl_group.cpp
randombit: These and others seem accurate but IMO not going to move the needle much since we have more heavy setup of Montgomery parameters etc which will dominate. Fine to do it but a little afield from the goal in thie PR so I'll pass for now.

PR #4592, opened 2025-01-25
Some initial refactoring for splitting public and private keys

randombit:
That private keys derive from the public key types was an early design mistake that will be fixed in Botan4. We can't make this change right now since it is a significant SemVer break, but these changes set things in the right direction and can be made now.

### DIFF
--- a/src/lib/pubkey/pk_keys.h
+++ b/src/lib/pubkey/pk_keys.h
@@ -111,6 +111,32 @@ class BOTAN_PUBLIC_API(3, 0) Asymmetric_Key {
        * key pair. For other key types it might throw Not_Implemented.
        */
       virtual std::unique_ptr<Private_Key> generate_another(RandomNumberGenerator& rng) const = 0;
+
+      /*
+      * Test the key values for consistency.
+      * @param rng rng to use
+      * @param strong whether to perform strong and lengthy version of the test
+      * @return true if the test is passed
+      */
+      virtual bool check_key(RandomNumberGenerator& rng, bool strong) const = 0;
+
+      // Declarations for internal library functions not covered by SemVer follow
+
+      /**
+      * Certain signatures schemes such as ECDSA have more than
+      * one element, and certain unfortunate protocols decided the
+      * thing to do was not concatenate them as normally done, but
+      * instead DER encode each of the elements as independent values.
+      *
+      * If this returns a value x then the signature is checked to
+      * be exactly 2*x bytes and split in half for DER encoding.
+      */
+      virtual std::optional<size_t> _signature_element_size_for_DER_encoding() const { return {}; }
+
+      /*
+      * Return the format normally used by this algorithm for X.509 signatures
+      */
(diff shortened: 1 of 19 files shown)


src/lib/pubkey/pk_keys.h
```diff
@@ -111,6 +111,32 @@ class BOTAN_PUBLIC_API(3, 0) Asymmetric_Key {
...
+      /**
+      * Certain signatures schemes such as ECDSA have more than
+      * one element, and certain unfortunate protocols decided the
+      * thing to do was not concatenate them as normally done, but
+      * instead DER encode each of the elements as independent values.
+      *
+      * If this returns a value x then the signature is checked to
+      * be exactly 2*x bytes and split in half for DER encoding.
+      */
+      virtual std::optional<size_t> _signature_element_size() const { return {}; }
```
reneme, line 134:
```suggestion
      /**
      * Certain signatures schemes such as ECDSA have more than
      * one element, and certain unfortunate protocols decided the
      * thing to do was not concatenate them as normally done, but
      * instead DER encode each of the elements as independent values.
      *
      * If this returns a value x then the signature is checked to
      * be exactly 2*x bytes and split in half for DER encoding.
      */
      virtual std::optional<size_t> _signature_element_size_for_DER_encoding() const { return {}; }
```

... to make it clearer what the purpose of this function is. Especially when seeing it overridden in the relevant classes above, I wasn't sure what this was meant to do.

src/lib/pubkey/pubkey.cpp
```diff
@@ -315,7 +307,7 @@ size_t PK_Signer::signature_length() const {
    } else if(m_sig_format == Signature_Format::DerSequence) {
       // This is a large over-estimate but its easier than computing
       // the exact value
-      return m_op->signature_length() + (8 + 4 * m_parts);
+      return m_op->signature_length() + (8 + 4 * 2);
```
reneme, line 310: These are a lot of magic numbers. Perhaps clarify their meaning by defining them as `constexpr` variables just above, or at least a comment? My best guess: 8 bytes of general DER header, 4 bytes of length field times number of signature parts.

reneme (approved):
Nice. Some nits otherwise LGTM.

src/lib/pubkey/pubkey.cpp
randombit: Basically this, though I certainly do not remember exactly how they were derived. I think this could be tightened up quite a bit especially as here we restrict the number of elements of a signature to exactly 2 which is all we really need for any signature scheme.

src/lib/pubkey/pk_keys.h
randombit: Renamed

src/lib/pubkey/pubkey.cpp
randombit: Knowing there are at most 2 signature parts I just brute forced it for all group sizes, so we have an exact calculation for all EC* signatures and a smaller over-estimate for DSA groups

PR #4593, opened 2025-01-25
New interface for PKCS8 key encryption

randombit:
Quite a few interfaces exist for this but there really are only two things that change between them: if the PBKDF runtime is specified in "iterations" or in a duration, and if the result is PEM encoded or not.

Add a new single interface for PKCS8 encryption, PKCS8::encrypt_private_key, then define all of the previously existing functions in terms of that.


randombit:
Some things to work out

* Best name for `EncryptedPrivateKey`. Perhaps reflecting the PKCS8 structure name?
* Should we take the opportunity to move this to `Private_Key`

src/lib/pubkey/pkcs8.h
```diff
@@ -47,10 +49,105 @@ inline secure_vector<uint8_t> BER_encode(const Private_Key& key) {
...
+      //
+      // This may change over time
+      //
+      // TODO(Botan4) Consider changing this to Scrypt
+      static std::string default_pwhash();
+
+      // The default password hash duration
+      //
+      // This may change over time
+      static std::chrono::milliseconds default_pwhash_duration();
```
reneme, line 73: Do those need to be public?

src/lib/pubkey/pkcs8.h
```diff
@@ -15,8 +15,10 @@
 #include <chrono>
 #include <functional>
 #include <memory>
+#include <optional>
 #include <span>
 #include <string_view>
+#include <variant>
```
reneme, line 21: Seems orphaned and unused. I'm guessing you considered it to distinguish duration and iterations.

reneme (commented):
I'm very much in favor of simplifying this. It is actually quite hard to find how asymmetric keys are exported into containers in our API.

> Should we take the opportunity to move this to `Private_Key`

This would be more in line with what #4318 currently proposes for signing. If we were to fully embrace this API design for key export, we'd probably end up with something like:

```C++

auto pkcs8 = sk.export()
               .with_cipher("AES-128/CBC")
               .with_pbkdf("Scrypt")
               .with_rng(rng)
               .as_pem();
```

... admittedly, this would add quite a bit of (internal) boilerplate for the Options-builder plumbing. But, to my mind, such an API is quite ergonomic for the user, since they don't have to learn much library-specific type vocabulary. Just type `sk.` and your IDE will suggest `.export()`, just type `.` again and you'll be presented with the different options you could use.

Such an API would also conveniently cover "unencrypted export". Though, for safety reasons, I'd suggest to guard that like so:

```C++
auto unencrypted = sk.export()
                     .without_encryption() // to deliberately request "unsafe"
                     .as_pem();
```

randombit:
I toyed with a builder idea during development but I figured I would let you be the one to suggest it :P

reneme:
Perhaps we should pull the reusable portions of the `PK_Options` draft (which won't be exposed publicly anyway) into a separate PR/branch. It would certainly make it easier to prototype on API such as this one. Given that this builder paradigm would bring quite a big API shift I'd love to toy with it on different aspects of the API before fully committing on it.

reneme:
Note: #4694 acts on my last comment and proposes an alternative API based on the facilities proposed last year in #4318 for `PK_Signature_Options`.

PR #4600, opened 2025-01-26
Avoid using IEEE 1363 EMSA names in OID data or EMSA::name

randombit:
For now we still support them on the decoding side, but when formatting we use the conventional names.

Leave notes as to what remains to be updated.


reneme (approved):
Nice cleanup! Thanks.

PR #4604, opened 2025-01-27
Avoid requiring legacy_ec_group in order to run the tests

randombit:
Add `EC_Group::supports_application_specific_group` and `EC_Group::supports_named_group` so applications can query for support at runtime.

Modify `EC_Group::known_named_groups` to not return group names that we don't support in the current build configuration.

FFI fix is a driveby thing that I noticed when switching the Python test to use the ISRG root instead of the BSI one.


randombit:
LOL the ISRG root cert has an expiration of 2040 which breaks time conversions on systems still using 32-bit time. 

ISRG should have set the expiration to 03:14 Jan 19, 2038 to maximize the LOLs.

I guess I need to find another test certificate.

src/tests/test_pubkey.cpp
```diff
@@ -848,12 +848,18 @@ class PK_API_Sign_Test : public Text_Based_Test {
          }
          Test::Result result(test_name.str());
 
-         auto privkey = Botan::create_private_key(algorithm, this->rng(), algo_params, provider);
+         std::unique_ptr<Botan::Private_Key> privkey;
+
+         try {
+            privkey = Botan::create_private_key(algorithm, this->rng(), algo_params, provider);
+         } catch(Botan::Not_Implemented&) {}
```
reneme, line 855: Nit:

```suggestion
         auto privkey = [&]() -> std::unique_ptr<Botan::Private_Key> {
            try {
               return Botan::create_private_key(algorithm, this->rng(), algo_params, provider);
            } catch(Botan::Not_Implemented&) {
               return nullptr
            }
         }();
```

reneme (approved):
Good catch by @M4iKZ. Disabling deprecated features shouldn't disable significant supported parts of the library. 😨

randombit:
> I guess I need to find another test certificate.

I realized this isn't necessary at all. Internally we are ourselves correctly computing the seconds since epoch from the `notBefore` field. But the flow had involved `uint64_t -> time_t -> uint64_t` [*] so we were truncating. I just exposed directly a function to return the seconds since epoch without involving `std::chrono::`. This avoids any 2038-style problems with parsing certificates.

[*] `calendar.cpp days_since_epoch` / `calendar_point::to_std_timepoint` / `ASN1_Time::time_since_epoch`

src/tests/test_pubkey.cpp
randombit: Fixed in a different way - ordinarily `create_private_key` returns nullptr if the algorithm is not available/etc (which is maybe not great but that is what it does). So I modified it to first check if we support the named group and if not then just return nullptr same as if ECDSA itself was compiled out.

randombit:
TIL that `std::chrono` can't handle times after [2262-04-11 at 23:47:16](https://en.wikipedia.org/wiki/Time_formatting_and_storage_bugs#Year_2262) which I discovered when trying to add tests for dates beyond that. Spent quite a while thinking it was a bug in our epoch calculation logic.

PR #4605, opened 2025-01-28
Fix Classic McEliece unpoison issue

randombit:
Cherry picked from #4562 

@reneme feel free to force push this with signed commits


reneme (approved):
Thanks. Fine with merging under your signature.

PR #4610, opened 2025-01-29
Allow using SHA-3 with RSASSA-PSS

FAlbertDev:
This PR should discuss RSASSA-PSS's hash functions. Currently, [RFC 8017](https://datatracker.ietf.org/doc/html/rfc8017#appendix-A.2.3) defines it for only SHA-1 and SHA-2. But [RFC 8692](https://datatracker.ietf.org/doc/rfc8692/) introduces additional RSASSA-PSS instances with SHAKE. We also want to use it with SHA-3 (unfortunately not explicitly defined in any RFC). Do we want to remove the hash function restriction in Botan and transfer the responsibility to the user? What do you think?


randombit (changes requested):
I don't think it's a good idea to leave it completely open - RSA-PSS-MD5 is not really so great.

Also we *cannot* add support for SHAKE in this way, if you examine the RSA-PSS-SHAKE RFC in detail you'll see that it uses a different parameters encoding, uses SHAKE instead of MGF1, etc. Supporting this would be a significant patch.

I'd be ok with extending this to include also the SHA-3 variants. There is always the risk however that eventually an RFC is defined that does something different just because, as the PSS-SHAKE RFC did. Ideally there would be some set of tests that demonstrate interop with some other impl on this. It looks like OpenSSL's `RSA_verify_PKCS1_PSS` will accept any `EVP_MD` - this doesn't seem like an intentional decision, just OpenSSL doing OpenSSL things, but at least it would give a sanity check.

Unrelated but we should deprecate RSA-PSS-SHA1 support.

FAlbertDev:
I experimented with OpenSSL and bouncycastle. Both support SHA-3 flawlessly. For example, for OpenSSL (CLI), we can simply sign via:
```bash
openssl dgst -sha3-256 -sigopt rsa_padding_mode:pss -sigopt rsa_pss_saltlen:-1 -sign private_key.pem -out signature.bin message.txt
```
With bouncycastle, I created some test vectors. Luckily, bouncycastle also allows custom RNGs, so we can test with custom salts. Here is the code for KAT generation:
```java
        // Initialize Signature with RSASSA-PSS
        Signature signature = Signature.getInstance("RSASSA-PSS", "BC");
        
        PSSParameterSpec pssParameterSpec = new PSSParameterSpec(
            "SHA3-512", "MGF1", MGF1ParameterSpec.SHA3_512, salt.length, 1);
        signature.setParameter(pssParameterSpec);
        signature.initSign(privateKey, testRng);

        // Sign the message
        signature.update(message);
        byte[] signedMessage = signature.sign();
```

Both OpenSSL and bouncycastle seem to be compatible if we simply activate SHA-3.

randombit:
OK OpenSSL + BC both working the same way sounds like consensus to me.

Just so all bases are covered can you add a few OpenSSL generated PSS-SHA-3 signatures to `rsa_verify.vec`

randombit:
I checked and RustCrypto's RSA supports PSS-SHA3 and we accept signatures they generate

FAlbertDev:
> Just so all bases are covered can you add a few OpenSSL generated PSS-SHA-3 signatures to rsa_verify.vec

I added one for each SHA-3 size with OpenSSL :+1:

> I checked and RustCrypto's RSA supports PSS-SHA3 and we accept signatures they generate

Nice! It seems to be a silent consensus that RSASSA-PSS can be used with SHA-3.

randombit (approved):
LGTM. Seems low risk so fine to merge now for 3.7

PR #4613, opened 2025-01-30
Fix some errors caught by test_all_configs.py

randombit:


randombit:
still wip will reopen

PR #4617, opened 2025-01-31
Fix build/test errors caught by test_all_configs.py

randombit:


randombit:
There are still some residual test failures in various configurations. Based on commit history I haven't done this exercise since 2020 so quite a backlog of issues. I'll take another pass during the 3.8 cycle.

Could try running at least the build-only checks in a monthly CI run to prevent this from regressing so badly. It might take too long in the Actions images though, I'm currently running this on a 128 core machine...

PR #4619, opened 2025-01-31
Add support for constexpr field inversions

randombit:
This would be usable for making the SSWU constants compile time constants but MSVC, Clang 14, and XCode 15 don't like it.


randombit:
This works with Clang 19 and GCC 14 but I have a feeling the older compilers will not like this.

PR #4620, opened 2025-02-01
Add fast variable time modular inversion to pcurves

randombit:
This is significantly faster than the constant time inversion, even when we have a precomputed addition chain for FLT.

Improvement (in operations/second) of scalar inversions, for situations when it is safe to allow variable time computation.

| Curve          | FLT       | BEEA     | Ratio |
|----------------|-----------|----------|-------|
| secp256r1      | 120557    | 199896   |  1.65 |
| secp384r1      | 51752     | 120299   |  2.32 |
| secp521r1      | 19833     | 61521    |  3.10 |
| brainpool256r1 | 98457     | 200100   |  2.03 |
| brainpool384r1 | 37268     | 116450   |  3.12 |
| brainpool512r1 | 19664     | 67213    |  3.41 |

This improves overall ECDSA verification performance by 5 to 8%.

### DIFF
--- a/src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
+++ b/src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
@@ -459,6 +459,119 @@ class IntMod final {
       constexpr Self invert() const { return pow_vartime(Self::P_MINUS_2); }
 
       /**
+      * Helper for variable time BEEA
+      *
+      * Note this function assumes that its arguments are in the standard
+      * domain, not the Montgomery domain. invert_vartime converts its argument
+      * out of Montgomery, and then back to Montgomery when returning the result.
+      */
+      static constexpr void _invert_vartime_div2_helper(Self& a, Self& x) {
+         constexpr auto INV_2 = p_div_2_plus_1(Rep::P);
+
+         while((a.m_val[0] & 1) != 1) {
+            shift_right<1>(a.m_val);
+
+            W borrow = shift_right<1>(x.m_val);
+
+            if(borrow) {
+               bigint_add2_nc(x.m_val.data(), N, INV_2.data(), N);
+            }
+         }
+      }
+
+      /**
+      * Returns the modular inverse, or 0 if no modular inverse exists.
+      *
+      * This function assumes that the modulus is prime
+      *
+      * This function does something a bit nasty and converts from the normal
+      * representation (for scalars, Montgomery) into the "standard"
+      * representation. This relies on the fact that we aren't doing any
+      * multiplications within this function, just additions, subtractions,
+      * division by 2, and comparisons.
+      *
+      * The reason is there is no good way to compare integers in the Montgomery
+      * domain; we could convert out for each comparison but this is slower than
+      * just doing a constant-time inversion.
+      *
+      * This is loosely based on the algorithm BoringSSL uses in
+      * BN_mod_inverse_odd, which is a variant of the Binary Extended Euclidean
+      * algorithm. It is optimized somewhat by taking advantage of a couple of
+      * observations.
+      *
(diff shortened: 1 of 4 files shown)


src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
```diff
@@ -458,6 +458,131 @@ class IntMod final {
...
+            if((a.m_val[0] & 1) == 1) {
+               break;
+            }
+
+            shift_right<1>(a.m_val);
+
+            W borrow = shift_right<1>(x.m_val);
+
+            if(borrow) {
+               // Hack to work around GCC miscompilation
```
briansmith, line 481: Is there a GCC bug # that should be cited?

randombit: GCC was miscompiling but given it seemed implicated with inline asm, which is pretty finicky and sometimes just fails when called in a new context even if it has worked fine for years, I was not going to file a bug until I was quite certain it was an actual GCC bug.

It appears the issue is lacking "volatile" on a block of inline asm that (after some nasty macro expansion) is just an 8 word long addition-with-carry chain. Even though the extended asm clobbers memory, GCC seems to not realize that calling it twice might result in different effects.

I had added `volatile` to the other asm statements that write to memory in 9b1b666f81 but apparently missed this one and of course had forgotten about it entirely.

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
```diff
@@ -458,6 +458,123 @@ class IntMod final {
...
+      * domain, not the Montgomery domain. invert_vartime converts its argument
+      * out of Montgomery, and then back to Montgomery when returning the result.
+      */
+      static constexpr void _invert_vartime_div2_helper(Self& a, Self& x) {
+         constexpr auto INV_2 = p_div_2_plus_1(Rep::P);
+
+         for(;;) {
+            if((a.m_val[0] & 1) == 1) {
+               break;
+            }
```
reneme, line 474: Couldn't that be the loop bound anyway?

```suggestion
         while((a.m_val[0] & 1) != 1) {
```

reneme (approved):
Code looks okay, the math is beyond me, frankly.

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
randombit: Yes. I had planned to optimize this so that a is shifted just once, as many times as required, then the loop reducing x runs. But it doesn't seem worth bothering with since I think the average number of iterations is 1 or 2.

randombit:
> Code looks okay, the math is beyond me, frankly.

It's pretty simple in the end, there are just several derivation steps so this end result doesn't obviously derive from the basic BEEA. I'll probably do a blog post on this.

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
```diff
@@ -458,6 +458,120 @@ class IntMod final {
...
+      * Note this function assumes that its arguments are in the standard
+      * domain, not the Montgomery domain. invert_vartime converts its argument
+      * out of Montgomery, and then back to Montgomery when returning the result.
+      */
+      static constexpr void _invert_vartime_div2_helper(Self& a, Self& x) {
+         constexpr auto INV_2 = p_div_2_plus_1(Rep::P);
+
+         while((a.m_val[0] & 1) != 1) {
+
+            shift_right<1>(a.m_val);
```
reneme, line 473:
```suggestion
         while((a.m_val[0] & 1) != 1) {
            shift_right<1>(a.m_val);
```

... clang-format being pedantic.

PR #4627, opened 2025-02-02
Add Booth encoding based variable point multiplication

randombit:
Improves variable point mul (and ECDH) by around 4-5%


randombit:
This algorithm is based on OpenSSL's nistz_p256, which is at this point (afaict) the last remaining faster EC library. I had expected better speedups, considering, but it seems like the field arithmetic is really the critical thing here. Improving that is left for future work.

The comment [here](https://github.com/openssl/openssl/blob/master/crypto/ec/ecp_nistputil.c#L140) will likely be helpful wrt review.

randombit:
in theory https://eprint.iacr.org/2013/816.pdf describes the algorithm but it doesn't do a great job of it

Issue #4628, opened 2025-02-03
Allow applications to query for the ECC implementation in use


reneme:
This idea should be seen in the context of https://github.com/randombit/botan/pull/4566. Along the lines of the newly added predicates `EC_Group::supports_application_specific_group()` and `EC_Group::supports_named_group()`, it would be great if we could query which ECC implementation is in use for a given `EC_Group` object. This would allow our application to `assert(custom_group.engine_type() == EccEngine::pcurves)` (or similar).

I.e. could we have a new method `EC_Group::engine_type()` that returns an `enum { pcurves, pcurves_generic, legacy_bigint }`? Preferably in 3.7.0, of course. 🥺

Apologies if this already exists and I just missed it.

randombit:
This doesn't exist at the moment. Is the idea here that an application can check if it is in the fast path for the curves it uses, and if not log a warning/abort/etc?

reneme:
Yes, exactly. For instance, for a botan build that has "secp256r1" and "pcurves_generic" enabled, I'd like to ensure that the NIST curve doesn't accidentally fall back to generic. E.g. because I forgot an OID-remapping in an application that uses custom OIDs because "reasons".

PR #4630, opened 2025-02-03
Add EC_Group::engine

randombit:
#4628

### DIFF
--- a/src/lib/pubkey/ec_group/ec_group.h
+++ b/src/lib/pubkey/ec_group/ec_group.h
@@ -43,6 +43,22 @@ enum class EC_Group_Source {
    ExternalSource,
 };
 
+/**
+* Enum indicating the way the group in question is implemented
+*
+* This is returned by EC_Group::engine
+*/
+enum class EC_Group_Engine {
+   /// Using per curve implementation; fastest available
+   Optimized,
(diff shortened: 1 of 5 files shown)


src/lib/pubkey/ec_group/ec_group.h
```diff
@@ -43,6 +43,21 @@ enum class EC_Group_Source {
...
+*/
+enum class EC_Group_Engine {
+   /// Using per curve implementation; fastest available
+   Pcurve,
+   /// A generic implementation that handles many curves in one implementation
+   PcurveGeneric,
+   /// The old implementation, used as a fallback
+   /// TODO(Botan4) remove this
+   LegacyBigInt,
+};
```
reneme, line 59: One consideration from reading the docstrings: is "pcurve" actual lingo that we want to expose to "casual" users?

Perhaps:

```suggestion
/**
* Enum indicating the way the group in question is implemented
*
* This is returned by EC_Group::engine
*/
enum class EC_Group_Engine {
   /// Using per curve implementation; fastest available
   Optimized,
   /// A generic implementation that handles many curves in one implementation
   Generic,
   /// The old implementation, used as a fallback
   /// TODO(Botan4) remove this
   Legacy,
};
```

reneme (approved):
Thanks a lot!

PR #4632, opened 2025-02-04
Remove pcurves "user interface" API

randombit:
In the very initial stages of pcurves the idea was that pcurves would be the new internal interface used for EC arithmetic. For example that ECDSA would be implemented directly using pcurves. So some effort went towards providing a "nice"ish API directly from pcurves.h

At least in the short term we can't possible do this (due to needing to still handle application curves with cofactors or huge parameters), and in the long term, even as we drop support for such things, it seems better to take advantage to simplify EC_Scalar and company, which would have about the same effect on overhead.

### DIFF
--- a/src/lib/math/pcurves/pcurves.h
+++ b/src/lib/math/pcurves/pcurves.h
@@ -66,70 +66,6 @@ class PrimeOrderCurve {
             Scalar& operator=(Scalar&& other) = default;
             ~Scalar() = default;
 
-            /**
-            * Return the size of the byte encoding of Scalars
-            */
-            size_t bytes() const { return m_curve->scalar_bytes(); }
-
-            /**
-            * Return the fixed length serialization of this scalar
-            */
-            template <concepts::resizable_byte_buffer T = std::vector<uint8_t>>
-            T serialize() const {
-               T bytes(this->bytes());
-               m_curve->serialize_scalar(bytes, *this);
-               return bytes;
-            }
-
-            /**
-            * Perform integer multiplication modulo the group order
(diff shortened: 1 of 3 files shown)


reneme (approved):
Frankly, this makes the using code somewhat harder to read. But given that the call sites are merely wrappers and delegations, KISS seems totally justified.

randombit:
The calling code is eventually going away in any case; there are multiple levels of cleanups possible here, some of which can be done now (I plan to do these during 3.8 cycle) and some of which will probably have to wait for Botan4.

There are a lot of impedance between what `pcurves.h` provides and what we actually need in practice. This brings significant overhead to calling these functions, at least more than I'd like given how much effort I'm putting into making it fast. Consider `EC_Scalar::add`. It calls

* `EC_Scalar_Data::add` (virtual call)
* `EC_Scalar_Data_PC::add` calls `pcurve.scalar_add` (virtual) after a `checked_ref` (`dynamic_cast`) on one of the args.
* `PrimeOrderCurveImpl::scalar_add` copies its arguments from stack allocated `array` of words into shorter (exactly of appropriate length) `array`s.
* Then the addition happens :tada:
* Then `PrimeOrderCurveImpl::scalar_add` copies the result out to a larger array again
* `EC_Scalar_Data_PC::add` allocates a `unique_ptr` to hold that result

That you can do all this in a couple hundred cycles is kind of amazing tbh. Anyway the goal is to get this down to say one virtual call plus one malloc, and without bouncing the value into and out of `StorageUnit` array.

PR #4631, opened 2025-02-03
FIX: heap corruption in TPM2 RSA constructor on MSVC

reneme:
This isn't fully confirmed, yet. But it seems that MSVC has a problem with the way we're initializing the `TPM2::RSA_PublicKey` (and private key). The class hierarchy is a bit awkward (see below) and we see a heap corruption when trying to call the `Botan::RSA_PublicKey` copy c'tor from within the `Botan::TPM2::RSA_PublicKey` constructor.

```C++
Botan::RSA_PublicKey rsa_pubkey_from_tss2_public(const TPM2B_PUBLIC*);

RSA_PublicKey::RSA_PublicKey(Object handle, SessionBundle session_bundle, const TPM2B_PUBLIC* public_blob) :
      Botan::TPM2::PublicKey(std::move(handle), std::move(session_bundle)),
      Botan::RSA_PublicKey(rsa_pubkey_from_tss2_public(public_blob)) // <- up-calls Botan::RSA_PublicKey move (?) constructor
      {}
```

This patch contains an attempted workaround, avoiding to up-call the copy/move constructor. Instead it calls the explicit constructor that expects the modulus and exponent in `Botan::RSA_PublicKey`.

Our best guess is that something goes wrong when trying to copy/move into this virtual class hierarchy while it is still under construction. Here's also a minimal reproduction in Godbolt (MSVC crashes, clang is fine): https://godbolt.org/z/b5186McfG

```mermaid
classDiagram
    AsymmetricKey <|-- Public_Key
    Public_Key <|-- RSA_PublicKey
    Public_Key <|-- TPM2_Public_Key
    TPM2_Public_Key <|-- TPM2_RSA_PublicKey
    RSA_PublicKey <|-- TPM2_RSA_PublicKey
```

PR #4633, opened 2025-02-04
FIX: unreachable code if XMD is not available

reneme:
MSVC warned about both branches of the `if constexpr` being unreachable when the "xmd" module is disabled.

### DIFF
--- a/src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
+++ b/src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
@@ -2014,21 +2014,17 @@ inline auto map_to_curve_sswu(const typename C::FieldElement& u) -> typename C::
 * addition.
 */
 template <typename C, bool RO>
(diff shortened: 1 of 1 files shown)


randombit (approved):

reneme:
I'm waiting on feedback whether this actually fixed the problem from @weberph2. Just like with [the TPM2-related fix](https://github.com/randombit/botan/pull/4631#issuecomment-2633342310), please don't wait with 3.7.0 if it is any trouble for you. (I should have probably opened this as draft)

weberph2:
@reneme Thank you! The following warnings/errors are fixed by this PR:
```
include\internal\botan\internal\pcurves_impl.h(2041) : error C2220: the following warning is treated as an error
include\internal\botan\internal\pcurves_impl.h(2041) : warning C4702: unreachable code

include\internal\botan\internal\pcurves_impl.h(2034) : error C2220: the following warning is treated as an error
include\internal\botan\internal\pcurves_impl.h(2034) : warning C4702: unreachable code
```

PR #4635, opened 2025-02-05
Use span instead of vector as arguments to EMSA

randombit:


src/lib/pk_pad/emsa_pssr/pssr.cpp
```diff
@@ -88,7 +88,7 @@ bool pss_verify(HashFunction& hash,
       return false;
    }
 
-   std::vector<uint8_t> coded = pss_repr;
+   std::vector<uint8_t> coded(pss_repr.begin(), pss_repr.end());
    if(coded.size() < key_bytes) {
```
reneme, line 92: Perhaps add a TODO to rework that implementation as future work. From a quick skim, I feel like we could avoid this allocation/copy for input that is `>= key_bytes`.

src/lib/pk_pad/emsa_raw/emsa_raw.cpp
```diff
@@ -44,27 +44,27 @@ std::vector<uint8_t> EMSA_Raw::raw_data() {
...
-bool EMSA_Raw::verify(const std::vector<uint8_t>& coded, const std::vector<uint8_t>& raw, size_t /*key_bits*/) {
+bool EMSA_Raw::verify(std::span<const uint8_t> coded, std::span<const uint8_t> raw, size_t /*key_bits*/) {
    if(m_expected_size && raw.size() != m_expected_size) {
       return false;
    }
 
    if(coded.size() == raw.size()) {
-      return (coded == raw);
+      return constant_time_compare(coded, raw);
    }
```
reneme, line 68: Do we even need to perform the length check explicitly?

src/lib/pk_pad/iso9796/iso9796.cpp
```diff
@@ -46,7 +46,7 @@ std::vector<uint8_t> iso9796_encoding(const std::vector<uint8_t>& msg,
       msg1 = std::vector<uint8_t>(msg.begin(), msg.begin() + capacity);
       hash->update(std::span(msg).subspan(capacity));
    } else {
-      msg1 = msg;
+      msg1.assign(msg.begin(), msg.end());
    }
```
reneme, line 50: This could perhaps be re-phrased  now (GitHub doesn't let me comment the entire hunk):

```C++
const auto prefix = std::min(capacity, msg.size());
const auto msg1 = msg.first(prefix)
hash->update(msg.subspan(prefix)); // <= may be empty
const auto hmsg2 = hash->final_stdvec();
```

Disclaimer: I didn't test this.

reneme (approved):
Nice. 😃 Attached comments are not really relevant for the API change in this PR, but for future work. There's certainly more cleanup that could applied to the actual implementations as well.

src/lib/pk_pad/emsa_pssr/pssr.cpp
randombit: We have to have at least one copy since we're applying MGF1 to the data. But in master we have two copies which is certainly not necessary. I fixed this up a bit, likely there is more that can be done.

src/lib/pk_pad/emsa_raw/emsa_raw.cpp
randombit: No we do not, but if the lengths are not equal we perform other logic instead, relating to skipping leading zero bytes in the recovered sig.

src/lib/pk_pad/iso9796/iso9796.cpp
randombit: Did it in a slightly different way, but yeah there was no need for this to be copied to a vector

randombit:
> There's certainly more cleanup that could applied to the actual implementations as well.

Certainly concur there, these have not seen much attention in recent years.

With EMSA now a completely hidden/internal class *and* the only caller being RSA we can probably evolve the APIs more suitably as well.

src/lib/pk_pad/iso9796/iso9796.cpp
```diff
@@ -20,40 +21,38 @@ namespace Botan {
...
-      hash->update(std::span(msg).subspan(capacity));
-   } else {
-      msg1 = msg;
-   }
+   std::span<const uint8_t> msg1;
+   const size_t msg1_len = std::min(capacity, msg.size());
+   const size_t msg2_len = msg.size() - msg1_len;  // possibly zero
+   msg1 = msg.first(msg1_len);
+
+   hash->update(msg.subspan(msg1_len, msg2_len));  // possibly empty
```
reneme, line 50: Disclaimer: This is purely academic at this point, perhaps we should rather give some love to all EMSA implementations in a follow up anyway.

Though, I just realized we have the `BufferSlicer` included anyway, might as well use it:

```C++
   BufferSlicer msg_slicer(msg);
   const auto msg1 = msg_slicer.take(std::min(capacity, msg.size()));
   const auto hmsg2 = hash->process<std::vector<uint8_t>>(msg_slicer.take(msg_slicer.remaining()));

   //compute H(C||msg1 ||H(msg2)||S)
   const auto salt = rng.random_vec<std::vector<uint8_t>>(salt_len);
   hash->update_be(static_cast<uint64_t>(msg1.size()) * 8);
   hash->update(msg1);
   hash->update(hmsg2);
   hash->update(salt);
   const std::vector<uint8_t> H = hash->final_stdvec();

   std::vector<uint8_t> EM(output_length);

   BufferStuffer stuffer(EM);
   stuffer.append(0x00, stuffer.remaining_capacity() - (hash_len + salt_len + trailer_len + msg1.size() + 1));
   stuffer.append(0x01);
   stuffer.append(msg1);
   stuffer.append(salt);
```

reneme (approved):

src/lib/pk_pad/emsa_raw/emsa_raw.cpp
randombit: Ah but this can be cleaned up significantly

src/lib/pk_pad/iso9796/iso9796.cpp
randombit: I'll stop at this point but I agree, finding this many obvious cleanups and minor improvements after a brief inspection suggests there is more yet that could be done.

PR #4638, opened 2025-02-05
Reorganize code and reduce header dependencies

randombit:
Preprocessing each source file and counting the non-empty lines (see script in this PR), compiling libbotan involves the compiler looking at roughly 38 million lines of code (!). Changes here bring it down to a bit under 35 million.

Still not great tbh, and I think reducing compile times could use more focus.

This ignores the tests for now, I expect they could see similar improvement.


randombit:
Oops still wip closing for now

Issue #3861, opened 2023-12-31
Split up build.h


randombit:
A lot of `build.h` is completely for internal use. It contains:

* Version information (needs to be public)
* `BOTAN_MP_WORD_BITS`
* Installation paths
* Path to the system cert bundle
* Definition of `BOTAN_DLL` (only reason that `ffi.h` depends on `build.h` is this)
* Botan target OS info / flags
* If we can use GCC inline asm
* CPU info / flags
* Valgrind toggle
* Debug asserts option
* The `BOTAN_HAS_FOO` list (needs to be public)
* `%{local_config}` (we should probably just remove this)
* Some things that are nominally editable but really should just be `const size_t`s in some appropriate header.

From an API perspective the only things that the user needs to see is the version info and the HAS_FOO feature macros.

Some other things probably need to be visible since we use them in headers somehow (eg `BOTAN_MP_WORD_BITS` defines size of `word` and we use that in headers)

Shinleon:
This appears to be fixed, at least in that build.h doesn't exist, and buildh.in doesn't have BOTAN_HAS_FOO or %{local_config}.

randombit:
This has now been completed with the exception of `--with-local-config`, which is deprecated and will be removed in 3.9

PR #4639, opened 2025-02-06
Remove most toggles from build.h

randombit:
Users should not, in practice, be touching any of these.

Initial step regarding #3861


randombit:
@reneme Review req here mostly regarding any "wait we totally use toggle X in prod"

src/lib/rng/rng.h
```diff
@@ -30,6 +30,22 @@ class Entropy_Sources;
 */
 class BOTAN_PUBLIC_API(2, 0) RandomNumberGenerator {
    public:
+      /**
+      * Userspace RNGs like HMAC_DRBG will reseed after a specified number
+      * of outputs are generated. Set to zero to disable automatic reseeding.
+      */
+      static constexpr size_t DefaultReseedInterval = 1024;
```
reneme, line 37: Should this maybe live in the `Stateful_RNG` subclass? (no strong opinion, having it all in the same place is nice for sure)

reneme (approved):
I'm not aware that anyone around here would be tweaking those. Many things are overrideable at runtime anyway, and that should be enough. Frankly, I was always wondering why tweaking these was even possible. :)

src/lib/utils/mem_utils.cpp
```diff
@@ -20,28 +20,32 @@ void secure_scrub_memory(void* ptr, size_t n) {
 #if defined(BOTAN_TARGET_OS_HAS_RTLSECUREZEROMEMORY)
    ::RtlSecureZeroMemory(ptr, n);
 
-#elif defined(BOTAN_TARGET_OS_HAS_EXPLICIT_BZERO)
+#elif defined(BOTAN_TARGET_OS_HAS_EXPLICIT_BZERO) && 0
```
randombit, line 23: Oops revert this (was testing codegen for the memset function pointer)

randombit, line 23:
```suggestion
#elif defined(BOTAN_TARGET_OS_HAS_EXPLICIT_BZERO)
```

src/lib/rng/rng.h
randombit: Possibly the value is relevent for a stateful RNG that is not deriving from `Stateful_RNG`, for example because the state is not in our address space but lives in some external service process.

randombit:
Turns out to be quite easy with `std::conditional` and probably was easily doable even in C++03 TBH ... just never fixed due to inertia #4643

Issue #4636, opened 2025-02-05
Building against FFI broken with 3.7.0 if feature checks are used


tobiasbrunner:
The C++20 check that was added at the end of the generated `build.h` with https://github.com/randombit/botan/commit/27ac696adc23e4818bd9b50e22d792095f854b69 (released with 3.7.0) breaks building a C program/library that uses feature checks (i.e. has `#include <botan/build.h>` followed by `#ifdef BOTAN_HAS_*`).

Maybe there should be an `#ifdef __cplusplus` check, like `ffi.h` has, before enforcing a version? Or maybe the check could be moved into an internal header file (kinda seems related to #3861)?

randombit:
Ah foiled again.

We do want it in a public header because attempting to compile using the Botan headers without C++20 enabled doesn't work for various reasons.

But you're totally right that it makes sense for a C application using `ffi.h` to also want to look at `build.h`. I had failed to consider this, now that `ffi.h` does not itself included `build.h`.

I'll spin a 3.7.1 this morning, moving this somewhere else

tobiasbrunner:
Thanks for the quick fix (also successfully tested it in our testing environment).

randombit:
#4640 adds a test in CI that GCC in C89 mode accepts `build.h` and `ffi.h`, so regressions like this should not recur.

tobiasbrunner:
Nice, thanks!

PR #4640, opened 2025-02-06
Add a FFI example, acting as a test that C compiler can parse build.h and ffi.h

### LINKED ISSUE #2903: Todo: Add a CI test that ffi.h is parseable by a C89 compiler
It is too easy for inlines or such to end up in this file and nothing in our CI will detect this so it easily ends up being shipped in a release.

randombit:
With an aim of preventing any future regressions like #4636 

Closes #2903

### DIFF
--- a/src/scripts/ci_build.py
+++ b/src/scripts/ci_build.py
@@ -230,6 +230,8 @@ def determine_flags(target, target_os, target_cpu, target_cc, cc_bin, ccache,
 
     if target in ['docs']:
         flags += ['--with-doxygen', '--with-sphinx', '--with-rst2man']
+    else:
+        flags += ['--without-doc']
 
     if target in ['docs', 'codeql', 'hybrid-tls13-interop-test', 'limbo']:
         test_cmd = None
--- a/src/scripts/ci_build.py
+++ b/src/scripts/ci_build.py
@@ -837,6 +837,13 @@ def main(args=None):
 
             cmds.append(make_prefix + make_cmd + make_targets)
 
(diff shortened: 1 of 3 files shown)


src/scripts/ci_build.py
```diff
@@ -835,6 +837,13 @@ def main(args=None):
 
             cmds.append(make_prefix + make_cmd + make_targets)
 
+            if target in ['examples']:
+                cmds.append(['gcc', '-Wall', '-Wextra', '-std=c89',
+                             '-I%s' % (os.path.join(build_dir, 'build/include/public')),
+                             os.path.join(root_dir, 'src/examples/ffi.c'),
+                             '-L%s' % (build_dir), '-lbotan-3', '-o',
+                             os.path.join(build_dir, 'build/examples/ffi')])
```
reneme, line 845: Somewhat ugly, but it does get the job done, I guess. I fully agree that this test doesn't justify actually "supporting" C compilers in our build system.

Perhaps as a quick win in making this less brittle:

```suggestion
            if target in ['examples'] and options.cc in ['clang', 'gcc']:
                cmds.append([options.cc, '-Wall', '-Wextra', '-std=c89',
                             '-I%s' % (os.path.join(build_dir, 'build/include/public')),
                             os.path.join(root_dir, 'src/examples/ffi.c'),
                             '-L%s' % (build_dir), '-lbotan-3', '-o',
                             os.path.join(build_dir, 'build/examples/ffi')])
```

reneme (approved):

PR #4642, opened 2025-02-06
Add internal target_info.h header

randombit:
The end goal here is that if you build the same version of the library with the same module selection, but for different targets/compilers/whatever `build.h` is identical between them [*]. This assists users whose vendoring strategy involves having a single shared set of headers plus multiple different library binaries depending on the platform. (There are some tickets relating to this I should dig up.)

[*] Largely because the only things the end-goal `build.h` will contain is the version macros plus `BOTAN_HAS_FOO` feature flags

This PR does not achieve that, just a step in the direction. ~One specifically troubling problem is `BOTAN_MP_WORD_BITS` since that is used to select the definition of `word`, and we use `word` in public API in `BigInt`. I think it's doable, even within SemVer constraints, but that one is a big refactoring all on its own.~

#3861

### DIFF
--- a/src/lib/math/bigint/bigint.h
+++ b/src/lib/math/bigint/bigint.h
@@ -805,15 +805,6 @@ class BOTAN_PUBLIC_API(2, 0) BigInt final {
 
       BOTAN_DEPRECATED("replaced by internal API") void const_time_unpoison() const { _const_time_unpoison(); }
 
-#if defined(BOTAN_CT_POISON_ENABLED)
-      void _const_time_poison() const;
-      void _const_time_unpoison() const;
-#else
-      constexpr void _const_time_poison() const {}
-
-      constexpr void _const_time_unpoison() const {}
-#endif
-
       /**
        * @param rng a random number generator
        * @param min the minimum value (must be non-negative)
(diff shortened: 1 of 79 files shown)


reneme:
>  One specifically troubling problem is `BOTAN_MP_WORD_BITS` since that is used to select the definition of word, and we use word in public API in BigInt.

There is no portable way to find this at compile time? 😨

randombit:
> There is no portable way to find this at compile time?

Ha good point. Yes surely we can just use `sizeof(void*)`?

Oh hmm that fails for x32 which uses 32 bit pointers but supports a 64 bit word. OTOH nobody cares about x32.

randombit:
Turns out to be quite easy with std::conditional and probably was easily doable even in C++03 TBH ... just never fixed due to inertia https://github.com/randombit/botan/pull/4643

randombit:
There is a tiny bit more to do (endian, sanitizer flags) but I will take these out in future PRs

src/lib/math/bigint/bigint.h
```diff
@@ -20,6 +20,8 @@ namespace Botan {
 
 class RandomNumberGenerator;
 
+static constexpr size_t BOTAN_MP_WORD_BITS = sizeof(word) * 8;
```
reneme, line 23: SemVer nit: This is now much narrower in scope, i.e. only available when users include `bigint.h`. Is that something we want to live with?

Perhaps even deprecate it?

reneme (approved):

src/lib/math/bigint/bigint.h
randombit: I think it is best to just remove it right away. It is very unclear what exactly someone would be using it for, and we only documented it as a toggle to control the size of `word` (and probably roughly zero people ever did that.) I checked rnp, KeepassXC, and SoftHSM, none use it. No hits in Google search or Github code search either, except of course us.

Remove it in this PR gives this the same treatment as the other macros previously in build.h; they just vanish, as far as the user is concerned.

Maybe 3.8 comes out and someone complains. More likely nobody ever notices.

PR #4645, opened 2025-02-07
Use std::endian

randombit:


reneme:
TIL! Could we use this in `loadstor.h`?

PR #4646, opened 2025-02-07
MSVC Warning/Bug Fix

FAlbertDev:
This PR fixes an unreachable code warning and works around an older MSVC version's compiler bug.

### Unreachable Code Warning
Unreachable code warnings occur in lines `pcurves_wrap.h:256` and `pcurves_wrap.h:266` with the latest MSVC if the `xmd` module is disabled and pcurves are enabled. MSVC sees that, due to an ifdef, `hash_to_curve_sswu` always throws an exception. Therefore, it assumes that `stash()` is unreachable. I guess MSVC is simply too smart for us...
### Compiler Bug
`test_tpm2.cpp:1228` and `test_tpm2.cpp:1226` trigger an internal compiler error in MSVC 19.41. I'm not sure why, but adding `std::ignore` fixes it.

### DIFF
--- a/src/lib/math/pcurves/pcurves_impl/pcurves_wrap.h
+++ b/src/lib/math/pcurves/pcurves_impl/pcurves_wrap.h
@@ -253,7 +253,12 @@ class PrimeOrderCurveImpl final : public PrimeOrderCurve {
                                    std::span<const uint8_t> input,
                                    std::span<const uint8_t> domain_sep) const override {
          if constexpr(C::ValidForSswuHash) {
+#if defined(BOTAN_HAS_XMD)
             return stash(hash_to_curve_sswu<C, false>(hash, input, domain_sep));
+#else
+            BOTAN_UNUSED(hash, input, domain_sep);
+            throw Not_Implemented("Hash to curve not available due to missing XMD");
+#endif
          } else {
             throw Not_Implemented("Hash to curve is not implemented for this curve");
          }
@@ -263,7 +268,12 @@ class PrimeOrderCurveImpl final : public PrimeOrderCurve {
                                        std::span<const uint8_t> input,
                                        std::span<const uint8_t> domain_sep) const override {
          if constexpr(C::ValidForSswuHash) {
+#if defined(BOTAN_HAS_XMD)
             return stash(hash_to_curve_sswu<C, true>(hash, input, domain_sep));
+#else
+            BOTAN_UNUSED(hash, input, domain_sep);
+            throw Not_Implemented("Hash to curve not available due to missing XMD");
+#endif
          } else {
             throw Not_Implemented("Hash to curve is not implemented for this curve");
          }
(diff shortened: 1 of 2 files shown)


randombit:
I'd like to understand why we didn't see the warning in CI. Is there a newer version of MSVC than what is on GHA?

src/lib/math/pcurves/pcurves_impl/pcurves_wrap.h
```diff
@@ -253,7 +253,12 @@ class PrimeOrderCurveImpl final : public PrimeOrderCurve {
...
          if constexpr(C::ValidForSswuHash) {
+#if defined(BOTAN_HAS_XMD)
             return stash(hash_to_curve_sswu<C, false>(hash, input, domain_sep));
+#else
+            BOTAN_UNUSED(hash, input, domain_sep);
+            throw Not_Implemented("Hash to curve not available due to missing XMD");
+#endif
          } else {
             throw Not_Implemented("Hash to curve is not implemented for this curve");
          }
```
reneme, line 264: `if constexpr` and then additionally guarding with `#if defined()` makes me cringe. 😞
It would be amazing if `C::ValidForSswuHash` would actually _depend on_ `BOTAN_HAS_XMD`.

reneme (commented):
This is very likely a follow-up of https://github.com/randombit/botan/pull/4633 🤕 

Frankly, I'm concerned that you will eventually just push the warning one level up, as MSVC now basically sees that `hash_to_curve_nu` respectively `hash_to_curve_ro` always throw.

Let's use this problem and explore whether we can integrate the notion of "module availability" with C++20's `if constexpr` and concepts. Here's a basic idea:

If `configure.py` would generate consteval helpers, along with `#define BOTAN_HAS_...`, in a central header, such that:

```C++
consteval bool has_module_xmd() {
#if defined(BOTAN_HAS_XMD)
  return true;
#else
  return false;
#fi
}
```

Then we could use those helpers in `require` clauses, concept definitions or trait-like `constexpr bool something = ...` declarations.

We should then be able to get rid of the `#if defined(...)` in this patch and also remove the one altered in #4633.

reneme:
> I'd like to understand why we didn't see the warning in CI.

We probably don't have an MSVC build without XMD.

randombit:
Re `consteval bool has_module_xmd()`

I've definitely been contemplating something like this but perhaps even broader, using a generated function

```
enum class Feature {
   XMD,
   ...
};

consteval has_feature(Feature feat) {
  switch(feat) {
#if defined(BOTAN_HAS_XMD)
    case XMD:
        return true;
#endif 
   default:
     return false;
   }
}
```

FAlbertDev:
@reneme and I updated this PR with a simple `has_module_xmd()`. 

I like your enum approach. Seems very sensible and easy to use 👍

randombit:
So for such a small patch I really hate the XMD changes :)

It conflates supporting SSWU with XMD which seems wrong to me, SSWU could work with any KDF we just only support XMD right now because that's the RFC 9380 standardized on. [In actuality Argon2 would have made a *lot* of sense, hm I may look into this...]

My proposal: #4648

MSVC TPM2 error fix is fine, maybe split that out into a second PR and we can get that merged while we bikeshed this XMD yak.

randombit:
Addressed in #4662 and #4648

PR #4648, opened 2025-02-07
Refactor hash2curve expand_message

randombit:
Pass the implementation of expand_message as a callback to the hash2curve primitive. This will make it easy to support expand_message_xof or another alternative derivation scheme in the future should it be useful.

### DIFF
--- a/src/lib/math/pcurves/pcurves.h
+++ b/src/lib/math/pcurves/pcurves.h
@@ -13,11 +13,11 @@
 #include <botan/secmem.h>
 #include <botan/types.h>
 #include <array>
+#include <functional>
 #include <memory>
 #include <optional>
 #include <span>
 #include <string_view>
-#include <vector>
 
 namespace Botan {
 
@@ -310,19 +310,25 @@ class PrimeOrderCurve {
       * RFC 9380 hash to curve (NU variant)
       *
       * This is currently only supported for a few specific curves
+      *
+      * @param expand_message is a callback which must fill the provided output
+      * span with a sequence of uniform bytes, or if this is not possible due to
+      * length limitations or some other issue, throw an exception. It is
+      * invoked to produce the `uniform_bytes` value; see RFC 9380 section 5.2
       */
-      virtual AffinePoint hash_to_curve_nu(std::string_view hash,
-                                           std::span<const uint8_t> input,
-                                           std::span<const uint8_t> domain_sep) const = 0;
+      virtual AffinePoint hash_to_curve_nu(std::function<void(std::span<uint8_t>)> expand_message) const = 0;
 
       /**
       * RFC 9380 hash to curve (RO variant)
       *
       * This is currently only supported for a few specific curves
+      *
+      * @param expand_message is a callback which must fill the provided output
+      * span with a sequence of uniform bytes, or if this is not possible due to
+      * length limitations or some other issue, throw an exception. It is
+      * invoked to produce the `uniform_bytes` value; see RFC 9380 section 5.2
       */
-      virtual ProjectivePoint hash_to_curve_ro(std::string_view hash,
-                                               std::span<const uint8_t> input,
-                                               std::span<const uint8_t> domain_sep) const = 0;
+      virtual ProjectivePoint hash_to_curve_ro(std::function<void(std::span<uint8_t>)> expand_message) const = 0;
 };
 
 }  // namespace Botan::PCurve
--- a/src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
+++ b/src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
@@ -14,10 +14,6 @@
 #include <botan/internal/stl_util.h>
 #include <vector>
 
-#if defined(BOTAN_HAS_XMD)
-   #include <botan/internal/xmd.h>
-#endif
-
 namespace Botan {
 
 /*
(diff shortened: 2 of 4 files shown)


src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
```diff
@@ -2250,39 +2246,36 @@ inline auto map_to_curve_sswu(const typename C::FieldElement& u) -> typename C::
...
 *
-* Hashes the input using XMD and the specified hash function, producing either one or
-* two field elements `u`/(`u0`,`u1`) resp. These are then mapped to curve point(s)
-* using SSWU, and if a pair of points were generated these are combined using point
-* addition.
+* This is the Simplified Shallue-van de Woestijne-Ulas (SSWU) map.
+*
+* The expand function is called expand_message in RFC 9380 and could for the
+* curves implemented here should usually be XMD, but expand_message_xof is also
+* allowed by the RFC (though this is not implemented currently)
```
FAlbertDev, line 2253: Rephrase

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
```diff
@@ -2250,39 +2246,36 @@ inline auto map_to_curve_sswu(const typename C::FieldElement& u) -> typename C::
...
+*
+* For details see RFC 9380 sections 3, 5.2 and 6.6.2.
 */
-template <typename C, bool RO>
+template <typename C, bool RO, typename ExpandMsg>
    requires C::ValidForSswuHash
-inline auto hash_to_curve_sswu(std::string_view hash, std::span<const uint8_t> pw, std::span<const uint8_t> dst)
+inline auto hash_to_curve_sswu(ExpandMsg expand_message)
```
FAlbertDev, line 2259: Is there a downside to also using `std::function<void(std::span<uint8_t>)>` instead of `ExpandMsg`?

src/lib/pubkey/ec_group/ec_inner_data.cpp
```diff
@@ -322,7 +336,17 @@ std::unique_ptr<EC_AffinePoint_Data> EC_Group_Data::point_hash_to_curve_nu(std::
...
-      auto pt = m_pcurve->hash_to_curve_nu(hash_fn, input, domain_sep);
+      // This could be extended to support expand_message_xof or a MHF like Argon2
+      auto expand_message = [&](std::span<uint8_t> xmd) {
+#if defined(BOTAN_HAS_XMD)
+         expand_message_xmd(hash_fn, xmd, input, domain_sep);
+#else
+         BOTAN_UNUSED(hash_fn, xmd, input, domain_sep);
+         throw Not_Implemented("Hash to curve is not implemented due to XMD being disabled");
+#endif
+      };
```
FAlbertDev, line 347: You could outsource this into a separate function. Something like: 
```cpp
std::function<void(std::span<uint8_t>)>) expand_message_xmd_func(std::string_view hash_fn, std::span<const uint8_t> input), std::span<const uint8_t> domain_sep)){
    return [=](std::span<uint8_t> xmd) {
#if defined(BOTAN_HAS_XMD)
         expand_message_xmd(hash_fn, xmd, input, domain_sep);
#else
         BOTAN_UNUSED(hash_fn, xmd, input, domain_sep);
         throw Not_Implemented("Hash to curve is not implemented due to XMD being disabled");
#endif
      };
}
```
I'm not sure if it's worth it, though.

FAlbertDev (approved):
I like this modular approach 👍 This should also avoid the MSVC warning.

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
reneme: In general, `std::function<>` comes with an overhead. It implements a type erasure of the thing being called, which requires an internal allocation. When you (can) template for the callable type, this is not the case.

Regardless, I suggest using `std::invocable<std::span<uint8_t>>` as a concept for `ExpandMsg`.

src/lib/math/pcurves/pcurves_impl/pcurves_wrap.h
```diff
@@ -249,24 +253,20 @@ class PrimeOrderCurveImpl final : public PrimeOrderCurve {
...
-                                   std::span<const uint8_t> domain_sep) const override {
+      AffinePoint hash_to_curve_nu(std::function<void(std::span<uint8_t>)> expand_message) const override {
          if constexpr(C::ValidForSswuHash) {
-            return stash(hash_to_curve_sswu<C, false>(hash, input, domain_sep));
-         } else {
-            throw Not_Implemented("Hash to curve is not implemented for this curve");
+            return stash(hash_to_curve_sswu<C, false>(expand_message));
          }
+
+         throw Not_Implemented("Hash to curve is not implemented for this curve");
```
reneme, line 261: Out of extensive caution: Let's leave the `else` branch. I could imagine that MSVC will complain about unreachable code when the `if constexpr` branch is taken above. Same in the function below.

src/lib/math/pcurves/pcurves.h
```diff
@@ -310,18 +310,14 @@ class PrimeOrderCurve {
       *
       * This is currently only supported for a few specific curves
       */
-      virtual AffinePoint hash_to_curve_nu(std::string_view hash,
-                                           std::span<const uint8_t> input,
-                                           std::span<const uint8_t> domain_sep) const = 0;
+      virtual AffinePoint hash_to_curve_nu(std::function<void(std::span<uint8_t>)> expand_message) const = 0;
```
reneme, line 313: Even though this is an internal interface, I think the parameter requires a bit of documentation, e.g.:

`@param expand_message  a callback acting as a XOF generating a uniform stream of bytes`

reneme (approved):
Generally, I think this refactoring is sensible but I find it somewhat non-obvious what the callback should do semantically. All of this is internal-API, so its not a big deal to begin with.

Thinking out loud: Perhaps it would be sensible for such use cases to allow splitting the "conditioning" of KDFs from their expansion phase. E.g. for HKDF I would envision the extraction step to happen first and the expansion step to be invoked independently. In case of HKDF this would naturally model a XOF. But that might not be the case for all KDFs.

Then, instead of passing a callback into hash_to_curve, we could pass a conditioned instance of `Botan::KDF` and the implementation could then simply call `kdf.expand()` on it.

Some sketchy API:

```C++
// KDF_Output is an additional API that prevents attempts of re-conditioning
KDF_Output kdf = Botan::create_or_throw("XMD")->extract(secret, salt, label);
auto uniform_bytes = kdf.output<64>(); // returning std::array<>
auto uniform_vector = kdf.output(42); // returning secure_vector<>
```

... of course this only works with KDFs that don't require the desired output length already during conditioning.

src/lib/math/pcurves/pcurves_impl/pcurves_wrap.h
FAlbertDev: Good point!

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
FAlbertDev: Well, the calling functions `hash_to_curve_nu/ro` pass a `std::function` anyway.  I have no hard preference for both variants, though.

randombit:
> Then, instead of passing a callback into hash_to_curve, we could pass a conditioned instance of Botan::KDF and the implementation could then simply call kdf.expand() on it.

Interesting concept! It would be a bit of a slog though because we'd have to model both KDFs which include the output length in the input [as indeed XMD does] and also ones which have length limitations. And I'm not sure if it applies in enough situations to make it worthwhile. Lastly it precludes (at least without an additional adaptor) using Argon2 or similar, which is possibly the most interesting usage (in the context of PAKEs which rely on hash2curve)

src/lib/math/pcurves/pcurves.h
randombit: Done

src/lib/math/pcurves/pcurves_impl/pcurves_wrap.h
randombit: Done

src/lib/pubkey/ec_group/ec_inner_data.cpp
randombit: Probably worth it, it's a good bit of duplication otherwise. Also provides a single extension point in the future for XOF/MHF support.

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
randombit: Switched to `std::invocable`

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
randombit: Rewrote this

PR #4661, opened 2025-02-10
Update BigInt documentation

randombit:
Mostly by removing references to things which either no longer exist or which will be removed in the future, and warning users away from doing anything with BigInt that isn't just serializing or deserializing it.

### DIFF
--- a/doc/api_ref/bigint.rst
+++ b/doc/api_ref/bigint.rst
@@ -1,35 +1,74 @@
 BigInt
 ========================================
 
-``BigInt`` is Botan's implementation of a multiple-precision integer. Thanks to
-C++'s operator overloading features, using ``BigInt`` is often quite similar to
-using a native integer type. The number of functions related to ``BigInt`` is
-quite large, and not all of them are documented here. You can find the complete
-declarations in ``botan/bigint.h`` and ``botan/numthry.h``.
+``BigInt``, in ``bigint.h``, is an implementation of a signed magnitude
+multiple-precision integer, which is used to implement certain older public key
+algorithms such as RSA. It also appears in other contexts, for example X.509
+certificate serial numbers are technically integer values and can be quite
+large, and so are represented using a ``BigInt``.
+
+A ``BigInt`` is a sequence of smaller integers of type ``word``; this type is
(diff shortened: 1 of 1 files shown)


doc/api_ref/bigint.rst
```diff
@@ -1,35 +1,72 @@
...
 
       Create a BigInt from a string. By default decimal is expected. With an 0x
-      prefix instead it is treated as hexadecimal. A ``-`` prefix to indicate
+      prefix, instead it is treated as hexadecimal. A ``-`` prefix to indicate
       negative numbers is also accepted.
 
-   .. cpp:function:: BigInt(std::span<const uint8_t> buf)
+   .. cpp:function:: static BigInt::from_bytes(std::span<const uint8_t> buf)
 
       Create a BigInt from a binary array (big-endian encoding).
```
reneme, line 37: Is there a way to control the sign using this function?

doc/api_ref/bigint.rst
```diff
@@ -1,35 +1,72 @@
...
+
+      Encode this BigInt as a big-endian integer. The sign is ignored.
+
+      There must be sufficient space to encode the entire integer in ``buf``.
+      If ``buf`` is larger than required, sufficient zero bytes will be
+      prefixed.
+
+   .. cpp:function:: size_t bytes() const
+
+      Return number of bytes need to represent value of ``*this``
```
reneme, line 49:
```suggestion
      Return number of bytes needed to represent value of ``*this``
```

doc/api_ref/bigint.rst
```diff
@@ -1,35 +1,72 @@
...
+
+   .. cpp:function:: size_t bits() const
+
+      Return number of bits need to represent value of ``*this``
```
reneme, line 53:
```suggestion
      Return number of bits needed to represent value of ``*this``
```

reneme (approved):
Some minor nits and a request for clarification.

doc/api_ref/bigint.rst
randombit: Fixed

doc/api_ref/bigint.rst
randombit: No, it is always positive. I updated the doc

PR #4669, opened 2025-02-11
FIX: tpm2_ecc silently depends on legacy_ec_point

reneme:
For instance, this module configuration would fail to compile (adding `legacy_ec_point` would fix it):

```
--minimized-build --enable-modules=tpm2_ecc,tpm2_crypto_backend,ecdh --with-tpm2
```

The fix is trivial, my best guess is that it slipped through for #4518.

/cc @FAlbertDev


FAlbertDev (approved):
Thanks

Issue #4668, opened 2025-02-11
Feedback from Building Botan 3.7.1 in NixOS/nixpkgs


thillux:
I started to experiment with Botan 3.7.1 in https://github.com/NixOS/nixpkgs/pull/381091, below are my current findings and questions:

- If I build with "--with-tpm2", the tests inside the build jail fail, because no /dev/tpmrm0 or /dev/tmp0 is present. Make these tests optional and only run on device presence?
- When using "--module-policy=bsi", ffi is disabled, but needed for botan-rs or usage as pluging in strongswan IPsec daemon. Is this intentional, to forbid ffi, when in BSI mode?
- With BSI policy and enable module ffi, the ffi_ecdh test fails.

thillux:
I retested the current master and can confirm, that the botan tests are working in my nix build now with default settings.

If I enable the bsi policy policy together with TPM2 support the following error shows up:

```
 -Wsuggest-override -Wshadow -Wextra-semi -DBOTAN_IS_BEING_BUILT  -I build/include/public -I build/include/internal  -c src/lib/pubkey/classic_mceliece/cmce_keys_internal.cpp -o build/obj/lib/pubkey_classic_mceliece_cmce_keys_internal.o
       > g++ -fPIC -fvisibility=hidden -fstack-protector -m64 -pthread -std=c++20 -D_REENTRANT  -O3 -Wall -Wextra -Wpedantic -Wstrict-aliasing -Wcast-align -Wmissing-declarations -Wpointer-arith -Wcast-qual -Wzero-as-null-pointer-constant -Wnon-virtual-dtor -Wold-style-cast -Wsuggest-override -Wshadow -Wextra-semi -DBOTAN_IS_BEING_BUILT  -I build/include/public -I build/include/internal  -c src/lib/pubkey/classic_mceliece/cmce_matrix.cpp -o build/obj/lib/pubkey_classic_mceliece_cmce_matrix.o
       > src/lib/prov/tpm2/tpm2_key.cpp: In function 'Botan::RSA_PublicKey Botan::TPM2::rsa_pubkey_from_tss2_public(const TPM2B_PUBLIC*)':
       > src/lib/prov/tpm2/tpm2_key.cpp:34:22: error: 'rsa_pubkey_components_from_tss2_public' was not declared in this scope; did you mean 'rsa_pubkey_from_tss2_public'?
       >    34 |                      rsa_pubkey_components_from_tss2_public(public_area));
[...]

PR #4671, opened 2025-02-11
Fix test error if TPM2 is enabled but no hardware is available

randombit:
By default do not run the TPM2 tests; on end user devices they are unlikely to be using the simulator, and these tests are not necessarily safe on real hardware, with respect to the TPM2's internal state.

GH #4668

### DIFF
--- a/src/tests/main.cpp
+++ b/src/tests/main.cpp
@@ -61,7 +61,7 @@ int main(int argc, char* argv[]) {
    try {
       const std::string arg_spec =
          "botan-test --verbose --help --data-dir= --pkcs11-lib= --provider= "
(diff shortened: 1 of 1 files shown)


randombit:
Open to other approaches here. Probably some of the tests (like tpm2_hash) are safe enough on real hardware. But we in any case shouldn't fail if the hardware is not there. That would prevent most distros from shipping with TPM2 support since typically they require the tests pass for whatever they ship. Feel free to push onto this branch I'm afk after this.

atreiber94 (approved):
LGTM, and I agree that for local testing you'd alwaays want to specify with which tpm you want to talk to via the `tcti_*` test parameters (as we did in local development). 

Funnily enough, `test_python.py` did not run into this error for testing tpm2 because it also skips if `tpm2_tcti_name is None`.

reneme (approved):

PR #4687, opened 2025-02-14
Refactor EC_Group_Data_Map::lookup_or_create

randombit:
There were two very distinct codepaths here, one for the case where the group being created has an OID, and one where it does not. The logic for the two cases has very little overlap.

Split this into two functions, lookup_or_create and lookup_or_create_without_oid

Add signposts for Botan4 for what can be cleaned up


src/lib/pubkey/ec_group/ec_group.cpp
```diff
@@ -72,97 +65,106 @@ class EC_Group_Data_Map final {
...
+               *
+               * If the params don't match then the user attempted to create some
+               * other set of parameters under an already known OID
+               */
+               if(i->params_match(p, a, b, g_x, g_y, order, cofactor)) {
+                  return i;
+               } else {
+                  throw Invalid_Argument("Attempting to register a curve using OID " + oid.to_string() +
+                                         " but a distinct curve is already registered using that OID");
+               }
```
reneme, line 86: Proper nit: Control flow would be a little easier to understand if we sifted out the error case first and then return without the `else`, IMO.

```suggestion
               if(!i->params_match(p, a, b, g_x, g_y, order, cofactor)) {
                  throw Invalid_Argument("Attempting to register a curve using OID " + oid.to_string() +
                                         " but a distinct curve is already registered using that OID");
               }
               
               return i;
```

src/lib/pubkey/ec_group/ec_group.cpp
```diff
@@ -72,97 +65,106 @@ class EC_Group_Data_Map final {
...
          } else {
-            // Here try to use the order as a hint to look up the group id, to identify common groups
-            const OID oid_from_store = EC_Group::EC_group_identity_from_order(order);
-            if(oid_from_store.has_value()) {
-               std::shared_ptr<EC_Group_Data> data = EC_Group::EC_group_info(oid_from_store);
+            /*
+            * This path is taken for an application registering a new EC_Group with an OID specified
+            */
+            new_group = EC_Group_Data::create(p, a, b, g_x, g_y, order, cofactor, oid, source);
+         }
```
reneme, line 119: Another nit (trying to avoid the mutable variable):

```C++
         auto new_group = [&] {
            if(auto g = EC_Group::EC_group_info(oid); g != nullptr) {
               /*
               * This turned out to be the OID of one of the builtin groups. Verify
               * that all of the provided parameters match that builtin group.
               */
               if(!g->params_match(p, a, b, g_x, g_y, order, cofactor)) {
                  throw Invalid_Argument("Attempting to register an EC group under OID of hardcoded group");
               }

               return g;
            }
            
            return EC_Group_Data::create(p, a, b, g_x, g_y, order, cofactor, oid, source);
         }();
```

reneme (approved):
Nice! We were going through this logic just yesterday and got confused by the exact pieces you cleaned off here. Thanks!

src/lib/pubkey/ec_group/ec_group.cpp
randombit: Makes sense, applied

PR #4692, opened 2025-02-16
Don't bother with loop blocking for carry chains with constant lengths

randombit:
As long as the iteration bound is a compile time constant, and not too large [*], both GCC and Clang will completely unroll a loop calling word_add or word_sub.

Reformulate word_add so that it can use dword if that is helpful. Unfortunately, GCC generates better code with dword but Clang only recognizes our current pattern for a carry chain, so we choose at compile time based on `__clang__`.

Only notable regression is a 3% decrease for ECDSA verification with secp256r1 using GCC. I'm not sure of the cause. Everything else (ECDSA, ECDH, RSA, DH) is either unchanged, or improved. Improvements generally were between 1 and 7%. Biggest improvements seen with Clang were with ECDH secp256r1 (7%) and ECDSA verification secp256r1 (9%), and with GCC ECDH secp521r1 (7%) and ECDSA verification brainpool384r1 (9%)

[*] Measured 17 for GCC and 32 for Clang, but this probably varies based on a variety of factors.

### DIFF
--- a/src/lib/math/mp/mp_asmi.h
+++ b/src/lib/math/mp/mp_asmi.h
@@ -1,6 +1,6 @@
 /*
 * Lowest Level MPI Algorithms
-* (C) 1999-2010 Jack Lloyd
+* (C) 1999-2010,2025 Jack Lloyd
 *     2006 Luca Piccarreta
 *
 * Botan is released under the Simplified BSD License (see license.txt)
@@ -12,6 +12,7 @@
 #include <botan/compiler.h>
 #include <botan/types.h>
 #include <botan/internal/target_info.h>
+#include <concepts>
 
 #if !defined(BOTAN_TARGET_HAS_NATIVE_UINT128)
    #include <botan/internal/donna128.h>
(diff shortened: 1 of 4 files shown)


randombit:
I do need to double check this doesn't cause unexpected regressions on ARM or POWER

randombit:
Checked another x86-64 machine, AMD EPYC 7773X

Improvements with Clang 19:
ECDH P-256 7%
ECDH P-521 10%
ECDH BP256 9%
ECDSA verify P-256 11%
ECDSA verify P-521 8%
ECDSA verify BP256 13%
ECDSA verify BP512 6%
Rest unchanged or under 2% improvement

GCC 14
ECDSA P-256 regression not seen
ECDH P-256 verify up 6%
ECDSA P-384 verify up 5%
ECDSA BP256 verify up 6%
ECDSA BP384 verify up 10%
ECDH P-256 up 6%
ECDH P-521 up 8%
Rest unchanged or under 2% improvement

randombit:
Checked another x86-64 machine, no regression for GCC ECDSA P-256 verify, others saw improvements of 2-7%

randombit:
I can no longer replicate the secp256r1 ECDSA verification regression on the machine I originally saw it on, just noise I guess.

randombit:
Another one - really strange regression for secp256k1 on GCC. Without `word4_sub3` asm the `redc_crandall` function becomes much slower, almost a 20% regression overall. Looking at the generated code I don't see the problem.

PR #4694, opened 2025-02-17
Alternative: New interface for PKCS8 key encryption

reneme:
This is an alternative suggestion for #4593 using the generic `OptionsBuilder` facilities from #4318 (with a few minor extensions and improvements).

Usage examples:

```C++
auto sk = Botan::create_private_key("ECDSA", rng);

// simple usage (with default parameters)
auto plain_pem     = sk->serialize().as_unencrypted_pem();
auto encrypted_pem = sk->serialize().as_pem(rng, "supersecurepassword");

// customization of the key encryption parameters
auto encrypted_pem2 = sk->serialize()
                            .with_cipher("AES-128/GCM")
                            .with_pbkdf_iterations(100000)
                            .as_pem(rng, "nooneknowsme");

// usage error of the API (customize encryption but don't actually encrypt)
// (will throw: "Key Exporter for ECDSA failed to use the options password hash tuning duration")
using namespace std::chrono_literals;
auto plain_ber = sk->serialize()
                            .with_pbkdf_desired_runtime(1s)
                            .as_unencrypted_ber();
```

This is currently meant as a proof of concept for the `OptionsBuilder`. If we want to continue in that direction, we should also implement the existing `PKCS8::*_encode()` functions in terms of this (and/or deprecate them).

### DIFF
--- a/src/lib/pubkey/pk_keys.h
+++ b/src/lib/pubkey/pk_keys.h
@@ -10,6 +10,7 @@
 
 #include <botan/asn1_obj.h>
 #include <botan/pk_ops_fwd.h>
+#include <botan/pkcs8.h>
 #include <botan/secmem.h>
 
 #include <memory>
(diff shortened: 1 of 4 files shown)


reneme:
Closing as outdated.

PR #4698, opened 2025-02-18
FIX: setting 'DER' format explicitly on `PK_Signer` trips assertion

reneme:
`PK_Signer::m_sig_element_size` is set within the constructor iff the format specified in the constructor call is 'DER'. When a user uses the default format (by not passing the optional 'format' c'tor parameter) and then explicitly sets the format via `PK_Signer::set_output_format()`, an assertion in hit in PK_Signer::signature() during signature creation.

Apparently this format setter wasn't covered in our tests. I added a regression test that successfully failed (🤡) without the fix in this commit. We noticed that while updating one of our applications that use `PK_Signer` as described.

This was most likely introduced three weeks ago in GH #4592.


randombit (approved):
Thanks

PR #4699, opened 2025-02-19
Add IPAddrBlock and ASIdentifiers extensions (RFC 3779)

arckoor:
Adds support for RFC 3779.
Continuation of https://github.com/randombit/botan/pull/4443


src/lib/x509/x509_ext.h
```diff
@@ -560,6 +562,252 @@ class BOTAN_PUBLIC_API(3, 5) TNAuthList final : public Certificate_Extension {
...
+         IPv6 = 16,
+      };
+
+      template <Version V>
+      class BOTAN_PUBLIC_API(3, 7) IPAddress final {
+         public:
+            IPAddress() = default;
+            IPAddress(std::span<uint8_t> v);
+
+            std::array<uint8_t, static_cast<size_t>(V)> value() const { return m_value; }
```
randombit, line 584: Using the enum value as the implict length is fine but this function has a lot of casts. Add

```
static constexpr size_t Length = static_cast<size_t>(V);
```

and use that

src/lib/x509/x509_ext.h
```diff
@@ -560,6 +562,252 @@ class BOTAN_PUBLIC_API(3, 5) TNAuthList final : public Certificate_Extension {
...
+                     return std::strong_ordering::less;
+                  } else if(lhs.value()[i] > rhs.value()[i]) {
+                     return std::strong_ordering::greater;
+                  }
+               }
+               return std::strong_ordering::equal;
+            }
+
+            friend bool operator==(const IPAddress<V>& lhs, const IPAddress<V>& rhs) {
+               for(size_t i = 0; i < static_cast<size_t>(V); i++) {
```
randombit, line 610: This can just be `lhs.value() == rhs.value()`

src/lib/x509/x509_ext.h
```diff
@@ -560,6 +562,252 @@ class BOTAN_PUBLIC_API(3, 5) TNAuthList final : public Certificate_Extension {
...
+      };
+
+      class BOTAN_PUBLIC_API(3, 7) IPAddressFamily final : public ASN1_Object {
+         public:
+            void encode_into(DER_Encoder&) const override;
+            void decode_from(BER_Decoder& from) override;
+
+            IPAddressFamily() = default;
+
+            IPAddressFamily(const std::variant<IPAddressChoice<Version::IPv4>, IPAddressChoice<Version::IPv6>>& choice,
```
randombit, line 669: Add a typedef for the `std::variant` within `IPAddressFamily`, and use throughout

src/lib/x509/x509_ext.h
```diff
@@ -560,6 +562,252 @@ class BOTAN_PUBLIC_API(3, 5) TNAuthList final : public Certificate_Extension {
...
+               } else {
+                  m_afi = 2;
+               }
+            }
+
+            uint16_t afi() const { return m_afi; }
+
+            std::optional<uint8_t> safi() const { return m_safi; }
+
+            const std::variant<IPAddressChoice<Version::IPv4>, IPAddressChoice<Version::IPv6>>& addr_choice() const {
```
randombit, line 683: OK to just use `const auto&` as the return type here

src/build-data/oids.txt
```diff
@@ -379,6 +379,8 @@
 2.5.29.36 = X509v3.PolicyConstraints
 2.5.29.37 = X509v3.ExtendedKeyUsage
 1.3.6.1.5.5.7.1.1 = PKIX.AuthorityInformationAccess
+1.3.6.1.5.5.7.1.7 = PKIX.ipAddrBlocks
+1.3.6.1.5.5.7.1.8 = PKIX.autonomousSysIds
```
randombit, line 383: Convention within the codebase (not adhered to 100%, admittedly) is for the part after the dot to be capitialized, eg `PKIX.IpAddressBlocks` / `PKIX.AutonomousSysIds`

Also here please add a comment referencing the RFC to assist future readers

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,582 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+   std::vector<uint8_t> output;
+   DER_Encoder(output).start_sequence().encode_list(m_ip_addr_blocks).end_cons();
+   return output;
+}
+
+void IPAddressBlocks::decode_inner(const std::vector<uint8_t>& in) {
+   BER_Decoder(in).decode_list(m_ip_addr_blocks).verify_end();
+}
+
+void IPAddressBlocks::sort_and_merge() {
```
randombit, line 890: It would be helpful, both for me right now, and any future readers, if there was a comment explaining the intended behavior. Ideally with references to the RFC, if relevant

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,582 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+      if(block.safi().has_value()) {
+         afam = static_cast<uint32_t>(afam << 8) | block.safi().value();
+      }
+      std::vector<IPAddressFamily>& fams = afam_map[afam];
+      fams.push_back(block);
+   }
+
+   std::vector<IPAddressFamily> merged_blocks;
+   for(auto it = afam_map.begin(); it != afam_map.end(); it++) {
+      std::vector<IPAddressFamily>& fams = it->second;
```
randombit, line 904: At this point, can `fams` be empty? It seems to me it can. If it cannot, add a `BOTAN_ASSERT_NOMSG` so that's clear

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,582 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+   BER_Decoder(in).decode_list(m_ip_addr_blocks).verify_end();
+}
+
+void IPAddressBlocks::sort_and_merge() {
+   // std::map is ordered, so a uint32_t of [afi,safi] will be in the right order
+   std::map<uint32_t, std::vector<IPAddressFamily>> afam_map;
+   for(IPAddressFamily& block : m_ip_addr_blocks) {
+      uint32_t afam = block.afi();
+      if(block.safi().has_value()) {
+         afam = static_cast<uint32_t>(afam << 8) | block.safi().value();
```
randombit, line 896: Does this sort incorrectly if safi has no value? In that case `afam` isn't shifted upwards

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,582 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+         merged_blocks.push_back(merge<Version::IPv4>(fams));
+      } else {
+         merged_blocks.push_back(merge<Version::IPv6>(fams));
+      }
+   }
+   m_ip_addr_blocks = merged_blocks;
+}
+
+template <IPAddressBlocks::Version V>
+IPAddressBlocks::IPAddressFamily IPAddressBlocks::merge(std::vector<IPAddressFamily>& blocks) {
```
randombit, line 917: Again here some RFC references and/or commentary would be helpful.

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,582 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+   bool all_inherit = true;
+   bool none_inherit = true;
+   for(IPAddressFamily& block : blocks) {
+      IPAddressChoice<V> choice = std::get<IPAddressChoice<V>>(block.addr_choice());
+      all_inherit = !choice.ranges().has_value() && all_inherit;
+      none_inherit = choice.ranges().has_value() && none_inherit;
+   }
+
+   if(all_inherit) {
+      return IPAddressFamily(IPAddressChoice<V>(), blocks[0].safi());
```
randombit, line 931: If `blocks` is empty this crashes. This should be checked earlier and either handled correctly, if that's possible, or otherwise rejected with an exception.

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,582 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+   BER_Object obj = from.get_next_object();
+   ASN1_Type type_tag = obj.type_tag();
+   if(type_tag != ASN1_Type::Sequence) {
+      throw Decoding_Error(fmt("Unexpected type for IPAddressFamily {}", static_cast<uint32_t>(type_tag)));
+   }
+   BER_Decoder obj_ber = BER_Decoder(obj);
+
+   std::vector<uint8_t> addr_family;
+   obj_ber.decode(addr_family, ASN1_Type::OctetString);
+   size_t addr_family_length = addr_family.size();
```
randombit, line 981: Variable should be `const`

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,582 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+      throw Decoding_Error(fmt("Unexpected type for IPAddressChoice {}", static_cast<uint32_t>(next_tag)));
+   }
+}
+
+template <IPAddressBlocks::Version V>
+void IPAddressBlocks::IPAddressChoice<V>::sort_and_merge() {
+   if(!m_ip_addr_ranges.has_value()) {
+      return;
+   }
+   std::vector<IPAddressOrRange<V>> current = m_ip_addr_ranges.value();
```
randombit, line 1036: This could be captured by reference

`const auto& current = m_ip_addr_ranges.value()`

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,582 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+   if(current.empty()) {
+      return;
+   }
+   bool changed = true;
+   while(changed) {
```
randombit, line 1041: Commentary explaining what's happening with RFC references please

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,582 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+            changed = true;
+         }
+      }
+   }
+   m_ip_addr_ranges = current;
+}
+
+template <IPAddressBlocks::Version V>
+void IPAddressBlocks::IPAddressOrRange<V>::encode_into(Botan::DER_Encoder& into) const {
+   const uint8_t version_octets = static_cast<uint8_t>(V);
```
randombit, line 1066: In general use `size_t` for lengths unless there is a specific reason to use `uintX_t`

Also using `size_t` here avoids having to cast it later in the function

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,582 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+   std::array<uint8_t, version_octets> min = m_min.value();
+   std::array<uint8_t, version_octets> max = m_max.value();
+
+   uint8_t zeros = 0;
+   uint8_t ones = 0;
+
+   bool zeros_done = false;
+   bool ones_done = false;
+
+   // count contiguos 0s/1s from the right of the min/max addresses
```
randombit, line 1077: contiguous

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,582 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+         uint8_t local_ones = static_cast<uint8_t>(std::countr_one(max[i - 1]));
+         ones += local_ones;
+         ones_done = (local_ones != 8);
+      }
+
+      if(zeros_done && ones_done) {
+         break;
+      }
+   }
+
```
randombit, line 1095: Here again I feel some commentary outlining the general procedure would be helpful

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,582 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+      into.start_sequence()
+         .add_object(ASN1_Type::BitString, ASN1_Class::Universal, compressed_min)
+         .add_object(ASN1_Type::BitString, ASN1_Class::Universal, compressed_max)
+         .end_cons();
+   }
+}
+
+template <IPAddressBlocks::Version V>
+void IPAddressBlocks::IPAddressOrRange<V>::decode_from(Botan::BER_Decoder& from) {
+   const uint8_t version_octets = static_cast<uint8_t>(V);
```
randombit, line 1168: `size_t`

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,582 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+
+   ASN1_Type next_tag = from.peek_next_object().type_tag();
```
randombit, line 1170: Broadly speaking, if a variable can be `const`, make it `const`

For variables created in 2-3 line loop bodies it's ok to skip, but having to read over 20-50 lines to see if a variable changes or not is easily avoidable following this rule - if it is not marked `const` then it is being modified later. Here, it's not... but I had to look.

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,582 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+void ASBlocks::ASIdentifiers::decode_from(Botan::BER_Decoder& from) {
+   BER_Object obj = from.get_next_object();
+   ASN1_Type type_tag = obj.type_tag();
+   if(type_tag != ASN1_Type::Sequence) {
+      throw Decoding_Error(fmt("Unexpected type for ASIdentifiers {}", static_cast<uint32_t>(type_tag)));
+   }
+   BER_Decoder obj_ber = BER_Decoder(obj);
+
+   BER_Object elem_obj = obj_ber.get_next_object();
+   uint32_t elem_type_tag = static_cast<uint32_t>(elem_obj.type_tag());
```
randombit, line 1331: Can be `const`

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,582 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+void ASBlocks::ASIdentifierChoice::sort_and_merge() {
+   if(!m_as_ranges.has_value()) {
+      return;
+   }
+   std::vector<ASIdOrRange> current = m_as_ranges.value();
+   if(current.empty()) {
+      return;
+   }
+   bool changed = true;
+   while(changed) {
```
randombit, line 1407: Can this and the other `sort_and_merge` logic be pulled out to a new function (in an anonymous namespace in this file) that handles both? They seem nearly identical

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,582 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+   } else {
+      throw Decoding_Error(fmt("Unexpected type for ASIdentifierChoice {}", static_cast<uint32_t>(type_tag)));
+   }
+}
+
+void ASBlocks::ASIdentifierChoice::sort_and_merge() {
+   if(!m_as_ranges.has_value()) {
+      return;
+   }
+   std::vector<ASIdOrRange> current = m_as_ranges.value();
```
randombit, line 1402: Can be a `const` reference instead of creating a copy

src/tests/unit_x509.cpp
```diff
@@ -1592,8 +1593,885 @@ Test::Result test_x509_tn_auth_list_extension_decode() {
...
+
+   Botan::X509_Certificate cert(Test::data_file("x509/x509test/" + filename));
+
+   using Botan::Cert_Extension::IPAddressBlocks;
+
+   auto ip_addr_blocks = cert.v3_extensions().get_extension_object_as<IPAddressBlocks>();
+
+   const auto& addr_blocks = ip_addr_blocks->addr_blocks();
+   result.confirm("cert has IPAddrBlocks extension", ip_addr_blocks != nullptr, true);
+   result.confirm("cert has exactly one IpAddrBlock", addr_blocks.size() == 2, true);
```
randombit, line 1609: This should use `test_eq(addr_blocks.size(), 2)`

Also the "has exactly one" is IIUC incorrect

src/tests/unit_x509.cpp
```diff
@@ -1592,8 +1593,885 @@ Test::Result test_x509_tn_auth_list_extension_decode() {
...
+   const std::string sig_algo{"RSA"};
+   const std::string hash_fn{"SHA-256"};
+   const std::string padding_method{"EMSA3(SHA-256)"};
+
+   // trailing 0s, trailing 1s, and some arbitrary values
+   std::vector<uint8_t> edge_values = {2, 4, 8, 16, 32, 64, 128, 0, 1, 3, 7, 15, 31, 63, 127, 255, 12, 46, 123, 234};
+
+   for(size_t i = 0; i < edge_values.size(); i++) {
+      for(size_t j = 0; j < 18; j++) {
+         auto ca_key = make_a_private_key(sig_algo, *rng);
```
randombit, line 1972: Do you really need to create a new key for each case? I would think it sufficient to create one key for the entire test

randombit: (Likewise for the self signed CA cert and the `X509_CA` instance)

src/tests/unit_x509.cpp
```diff
@@ -1592,8 +1593,885 @@ Test::Result test_x509_tn_auth_list_extension_decode() {
...
+         const auto ca_cert = Botan::X509::create_self_signed_cert(ca_opts(), *ca_key, hash_fn, *rng);
+         Botan::X509_CA ca(ca_cert, *ca_key, hash_fn, padding_method, *rng);
+
+         auto key = make_a_private_key(sig_algo, *rng);
+
+         Botan::X509_Cert_Options opts1 = req_opts1(sig_algo);
+
+         std::vector<uint8_t> min_bytes = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+         std::vector<uint8_t> max_bytes = {
+            255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255};
```
randombit, line 1983:
```
std::vector<uint8_t> min_bytes(0, 0x00);
std::vector<uint8_t> max_bytes(16, 0xFF);
```

src/tests/unit_x509.cpp
```diff
@@ -1592,8 +1593,885 @@ Test::Result test_x509_tn_auth_list_extension_decode() {
    return result;
 }
 
+Test::Result test_x509_ip_addr_blocks_extension_decode() {
```
randombit, line 1596: Thank you for the careful tests. This file is already growing quite large; please move all of these tests into a new file eg `test_x509_rpki.cpp` and add them as a new `Test`

[I don't know if `rpki` is the best term to use here, if you think something else is better feel free to use that]

src/tests/unit_x509.cpp
```diff
@@ -1592,8 +1593,885 @@ Test::Result test_x509_tn_auth_list_extension_decode() {
...
+   result.confirm("cert has exactly one IpAddrBlock", addr_blocks.size() == 2, true);
+
+   const auto& ipv4block =
+      std::get<IPAddressBlocks::IPAddressChoice<IPAddressBlocks::Version::IPv4>>(addr_blocks[0].addr_choice());
+   const auto& ipv6block =
+      std::get<IPAddressBlocks::IPAddressChoice<IPAddressBlocks::Version::IPv6>>(addr_blocks[1].addr_choice());
+
+   auto& v4_blocks = ipv4block.ranges().value();
+
+   result.confirm("ipv4 block 0 min", v4_blocks[0].min().value() == std::array<uint8_t, 4>{192, 168, 0, 0}, true);
```
randombit, line 1618: For comparing octet strings in tests I generally prefer

`result.test_eq("foo equals bar", hex_encode(binary_data), "C0A80000")`

since then if it doesn't match you get a message like

"ABCDEF00 != C0A80000"

rather than, effectively

"it didn't match"

src/tests/unit_x509.cpp
```diff
@@ -1592,8 +1593,885 @@ Test::Result test_x509_tn_auth_list_extension_decode() {
...
+      bool inherit_ipv6 = i >> 3 & 1;
+      bool push_ipv4_family = i >> 4 & 1;
+      bool push_ipv6_family = i >> 5 & 1;
+
+      auto ca_key = make_a_private_key(sig_algo, *rng);
+      result.require("CA key", ca_key != nullptr);
+      const auto ca_cert = Botan::X509::create_self_signed_cert(ca_opts(), *ca_key, hash_fn, *rng);
+      Botan::X509_CA ca(ca_cert, *ca_key, hash_fn, padding_method, *rng);
+
+      auto key = make_a_private_key(sig_algo, *rng);
```
randombit, line 1790: There doesn't seem any need to create a new CA key, CA cert, or subject key for each iteration of the loop

src/tests/unit_x509.cpp
```diff
@@ -1592,8 +1593,885 @@ Test::Result test_x509_tn_auth_list_extension_decode() {
...
+
+   result.end_timer();
+   return result;
+}
+
    #endif
 
+Test::Result test_x509_ip_addr_blocks_extension_encode() {
+   Test::Result result("X509 IP Address Block encode");
+   #if defined(BOTAN_HAS_RSA)
```
randombit, line 1766: RSA is quite slow to generate and this test (IIUC) doesn't depend in any way on which algorithm is available. Since I anyway suggest splitting these tests out to a new source file dedicated to the RPKI extensions, I'd suggest adding a helper function like [untested but directionally correct]

```
std::unique_ptr<Botan::Private_Key> generate_key(RandomNumberGenerator& rng) {
#if defined(BOTAN_HAS_ECDSA)
   return Botan::make_private_key("ECDSA", rng, "secp256r1");
#endif

#if defined(BOTAN_HAS_ED25519)
   return Botan::make_private_key("Ed25519", rng, "");
#endif

#if defined(BOTAN_HAS_RSA)
   return Botan::make_private_key("RSA", rng, "1536");
#endif
}
```

and using it for the various tests. Then testing these functions does not depend on the existence of RSA, and by default we'll use ECDSA (or Ed25519) which are much faster to generate.

src/tests/unit_x509.cpp
```diff
@@ -1592,8 +1593,885 @@ Test::Result test_x509_tn_auth_list_extension_decode() {
...
+      ASBlocks::ASIdentifiers ident = ASBlocks::ASIdentifiers(include_asnum ? std::optional(asnum) : std::nullopt,
+                                                              include_rdi ? std::optional(rdi) : std::nullopt);
+
+      std::unique_ptr<ASBlocks> blocks = std::make_unique<ASBlocks>(ident);
+
+      auto ca_key = make_a_private_key(sig_algo, *rng);
+      result.require("CA key", ca_key != nullptr);
+      const auto ca_cert = Botan::X509::create_self_signed_cert(ca_opts(), *ca_key, hash_fn, *rng);
+      Botan::X509_CA ca(ca_cert, *ca_key, hash_fn, padding_method, *rng);
+      auto key = make_a_private_key(sig_algo, *rng);
```
randombit, line 2356: As before this key/cert generation step should be pulled out of the loop

randombit (changes requested):
There are some parts of this I don't understand, I still need to read the RFC, so this isn't quite a complete review.

src/lib/x509/x509_ext.cpp
arckoor: I'll try and get a more complete answer / documentation done somewhere around Tuesday, but the tl;dr for all the merge_* functions is that they are in response to https://datatracker.ietf.org/doc/html/rfc3779#section-2.2.3.3 and the last sentences of https://datatracker.ietf.org/doc/html/rfc3779#section-2.2.3.6 (https://datatracker.ietf.org/doc/html/rfc3779#section-3.2.3.4 for the AS nums)
Essentially, we want to compress the certificate to as few values as we can, merging ranges when possible and the like, both for size reasons and to make it "easier" for users of the cert to verify

For all the other areas I'll add it to the code when I have the time soon-ish

src/tests/unit_x509.cpp
randombit: Oh I forgot there is an existing helper for this in the tests

`result.test_eq("foo equals bar", binary_data, "00010203");`

src/lib/x509/x509_ext.cpp
butteronarchbtw: This should not be the case, since `fams` is an entry in the map with the key being an afi/safi combination. A block actually got pushed into fams if this afi/safi combination appeared in the extension object created by the user, so the length of `fams` has to be at least 1. A key should not be in the map without a block being pushed into the corresponding vector.

src/lib/x509/x509_ext.cpp
butteronarchbtw: Again, this does work afaics, since the calls to `sort` and `erase` cannot take a `const` reference.

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,607 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+}
+
+void IPAddressBlocks::decode_inner(const std::vector<uint8_t>& in) {
+   BER_Decoder(in).decode_list(m_ip_addr_blocks).verify_end();
+}
+
+void IPAddressBlocks::IPAddressFamily::encode_into(Botan::DER_Encoder& into) const {
+   into.start_sequence();
+
+   std::vector<uint8_t> afam = {static_cast<uint8_t>(m_afi >> 8), static_cast<uint8_t>(m_afi & 0xff)};
```
randombit, line 893: Use `{ get_byte<1>(m_afi), get_byte<0>(m_afi) }`

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,607 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+   }
+   into.end_cons();
+}
+
+void IPAddressBlocks::IPAddressFamily::decode_from(Botan::BER_Decoder& from) {
+   const BER_Object obj = from.get_next_object();
+   const ASN1_Type type_tag = obj.type_tag();
+   if(type_tag != ASN1_Type::Sequence) {
+      throw Decoding_Error(fmt("Unexpected type for IPAddressFamily {}", static_cast<uint32_t>(type_tag)));
+   }
```
randombit, line 914: You can use `peek_next_object` here and avoid having to create a second `BER_Decoder`

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,607 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+
+template <IPAddressBlocks::Version V>
+IPAddressBlocks::IPAddressFamily IPAddressBlocks::merge(std::vector<IPAddressFamily>& blocks) {
+   // Merge IPAddressFamilies that have the same afi/safi combination
+   //
+   // see: https://www.rfc-editor.org/rfc/rfc3779.html#section-2.2.3.3
+
+   if(blocks.empty()) {
+      throw Exception("at least 1 block is needed to merge");
+   }
```
randombit, line 991: If this should not occur absent a bug, this should be a `BOTAN_ASSERT`. If there is some circumstance where this is expected to occur, the error message should be clearer.

randombit (commented):
Thanks for addressing the comments. This is looking quite close. For my part I still need to review the RFC in order to understand the logic with merging/sorting.

src/lib/x509/x509_ext.h
```diff
@@ -560,6 +562,249 @@ class BOTAN_PUBLIC_API(3, 5) TNAuthList final : public Certificate_Extension {
...
+      static OID static_oid() { return OID("1.3.6.1.5.5.7.1.8"); }
+
+      OID oid_of() const override { return static_oid(); }
+
+      const ASIdentifiers& as_identifiers() const { return m_as_identifiers; }
+
+   private:
+      ASIdentifiers m_as_identifiers;
+
+      std::string oid_name() const override { return "PKIX.autonomousSysIds"; }
```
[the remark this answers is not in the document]
butteronarchbtw: At least for our tests, I can say they do not crash there, because we call `get_extension_object_as<IPAddressBlocks>()`, which itself only matches the `static_oid` and never the `oid_name` of the extension.

randombit:
More of a meta question - RFC 3779 defines (well, at least _alludes to_) how to verify this extension, in section 3.3. Is implementing this verification as part of the current path verification logic planned, perhaps in a later PR? Or were you planning on verifying yourselves out of band, after standard path verification? Either is fine; the first is preferable but if the second it should be documented that these extensions are exposed for informational purposes but are not examined as part of path verification.

arckoor:
Verification is planned as part of path verification. We can try and add it to this PR already, or add it to the FFI PR that should also come eventually (or do it entirely separate).

src/lib/x509/pkix_enums.h
```diff
@@ -67,6 +67,8 @@ enum class Certificate_Status_Code {
    CERT_CHAIN_TOO_LONG = 4002,
    CA_CERT_NOT_FOR_CERT_ISSUER = 4003,
    NAME_CONSTRAINT_ERROR = 4004,
+   IPADDR_BLOCK_ERROR = 4011,
+   ASBLOCKS_ERROR = 4012,
```
arckoor, line 71: We weren't really sure what to call the errors, nor what kind of error messages to use. We think it's definitely a validation error, but unfortunately there wasn't any space left within the enum, so now the numbers are kind of awkward.

arckoor:
@randombit We've added path validation for both extensions. The IP Address one is still missing a more thorough failure test, which we will add... eventually, hopefully by the end of the month. We've pushed our current changes for now, so you can hopefully already take a look.

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,802 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+
+      // address 1 (min)
+
+      // again accounting for the 'unused' octet at the beginning
+      if(addr_min.empty() || addr_min.size() > version_octets + 1) {
+         throw Decoding_Error(
+            fmt("IP address range entries must have a length between 1 and {} bytes.", static_cast<uint8_t>(V)));
+      }
+
+      addr_min.erase(addr_min.begin());  // remove 'unused' octet
```
arckoor, line 1355: Further down for the max address we check that if `max_addr.empty()` then `unused` must be 0. We need to check this because we need to set the unused bits to 1. For the min address however they are already 0, so until now we've just ignored the unused octet, as we don't need to examine it. Should we still check it? In the case that `unused != 1` and `addr_min.empty()` it would technically be a "bad" encoding - I'm just not sure if the user should be made aware of that.

arckoor:
@randombit Path validation is now tested more thoroughly. We've also left comments on some parts of the code where we're unsure, feedback there would also be appreciated c:

src/lib/x509/pkix_enums.h
randombit: On the naming, what you have seems fine. It's a bit obscure but anyone doing RPKI probably has at least some idea what these mean.

Re the numbering - oof. Absolutely my bad here. Can you add here // TODO(Botan4) annotations to renumber the enum? I do not know what I was thinking leaving so little (and/or zero) gaps here.

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,831 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+   }
+
+   // they are all 'inherit', short-circuit using default constructor for nullopt
+   if(all_inherit) {
+      return IPAddressFamily(IPAddressChoice<V>(), blocks[0].safi());
+   }
+
+   // some are inherit, and some have values - no sensible way to merge them
+   if(!all_inherit && !none_inherit) {
+      throw Exception("Only one of \"inherit\" / \"don't inherit\" is allowed per family");
```
randombit, line 1013: This should be a `Decoding_Error`

randombit: And probably with some prefix to give further hints to someone trying to debug this from a log eg "Invalid IpAddressBlocks: "

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,831 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+   return current;
+}
+
+template <typename T>
+bool validate_subject_in_issuer(const std::vector<T>& subject, const std::vector<T>& issuer) {
+   // ensures that the subject ranges are enclosed by the issuer ranges
+   // both vectors are already sorted, so we can do this in O(n+m)
+
+   // the issuer has 0 ranges to validate against, so this can only work if the subject also has none
+   if(issuer.size() == 0) {
```
randombit, line 1089: Should be `if(issuer.empty())` [I'm surprised that clang-tidy didn't warn about this]

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,831 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+      return subject.size() == 0;
```
randombit, line 1090: Likewise `subject.empty()`

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,831 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+         return false;
+      }
+      // the issuer range is smaller than the subject range, step to the next issuer range
+      if(subj->min() > issu->max()) {
+         issu++;
+         // we have run out of issuer ranges, but still have subject ranges left to validate
+         if(issu == issuer.end() && subj != subject.end()) {
+            return false;
+         }
+         continue;
```
randombit, line 1104: The `continue` here makes the logic harder to parse. (At least for me :sweat_smile:) Can this be done some other way? IIUC each of these are exclusive so could be done as `if else`

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,831 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+}
+
+template <IPAddressBlocks::Version V>
+void populate_validation_map(uint32_t afam,
+                             const IPAddressBlocks::IPAddressFamily::AddrChoice& choice,
+                             IPValidationMap<V>& map) {
+   const std::optional<IPRangeVec<V>>& ranges = std::get<IPAddressBlocks::IPAddressChoice<V>>(choice).ranges();
+   bool has_value = ranges.has_value();
+   const IPRangeVec<V>* value = has_value ? &ranges.value() : nullptr;
+   map.emplace(afam, std::make_pair(has_value, std::move(value)));
```
randombit, line 1122: It's not immediately clear what a `nullptr` means in an `IPValidationMap`. A comment near the type definition above would be helpful.

src/lib/x509/x509_ext.cpp
randombit: Thanks for calling this out. My preference would be that decoding be as strict as possible unless there was a significant mis-encoding in the ecosystem, such that not supporting that mis-encoding would impair interop (as for example in #4732)

src/lib/x509/x509_ext.cpp
```diff
@@ -868,6 +877,831 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+   size_t max;
+
+   if(next_tag == ASN1_Type::Integer) {
+      from.decode(min);
+      m_min = static_cast<uint32_t>(min);
+      m_max = m_min;
+   } else if(next_tag == ASN1_Type::Sequence) {
+      from.start_sequence().decode(min).decode(max).end_cons();
+      m_min = static_cast<uint32_t>(min);
+      m_max = static_cast<uint32_t>(max);
```
randombit, line 1635: Probably these should check for overflow. There are utils for this in `int_utils.h` (`checked_cast_to_or_throw`)

randombit (changes requested):
Not a complete review but some comments to keep things moving.

src/lib/x509/x509_ext.h
```diff
@@ -560,6 +562,273 @@ class BOTAN_PUBLIC_API(3, 5) TNAuthList final : public Certificate_Extension {
...
+
+         public:
+            IPAddress() = default;
+            IPAddress(std::span<uint8_t> v);
+
+            std::array<uint8_t, Length> value() const { return m_value; }
+
+            friend IPAddress<V> operator++(const IPAddress<V> v) {
+               std::array<uint8_t, Length> val = v.value();
+               for(size_t i = 0; i <= Length; i++) {
```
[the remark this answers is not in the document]
randombit: I think this is correct? @arckoor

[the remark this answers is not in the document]
butteronarchbtw: yes, it indeed is. we observed and fixed it locally already.

src/lib/x509/x509_ext.cpp
```diff
@@ -857,6 +863,841 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+         if(b.min() <= a.max() || b.min() == (a.max() + 1)) {
+            // erase old a and b
+            current.erase(current.begin() + i, current.begin() + i + 2);
+            current.insert(current.begin() + i, T(a.min(), std::max(a.max(), b.max())));
+            // look again at the new element we just created, might need to be merged with others still
+            i -= 1;
+            changed = true;
+         }
+      }
+   }
```
arckoor, line 1068: The while-loop is kind of redundant, at least with the `i -= 1` present. The latter in combination with the sort ensures that all possible merges happen, so the loop will always terminate at latest in the second iteration. The `i -= 1` method does lead to a nice size_t underflow if we merge on the first iteration, which is then directly incremented back to 0 again by the for-loop, but I think we should decide on one or the other, not both.

src/lib/x509/x509_ext.cpp
```diff
@@ -857,6 +863,841 @@ const std::string& TNAuthList::Entry::telephone_number() const {
...
+   return current;
+}
+
+template <typename T>
+bool validate_subject_in_issuer(const std::vector<T>& subject, const std::vector<T>& issuer) {
```
randombit, line 1073: Use `span` for input ranges

randombit (commented):
Did another full review. One tiny nit. Besides that and the issue you brought up regarding the merge loop this looks to me ready to merge.

src/lib/x509/x509_ext.cpp
randombit: To me the while loop seems easier to understand

Re the for loop, if retained please do leave a comment calling out the underflow when `i = 0` iiuc it's safe but it's a great example of something subtle that someone might "rediscover" years down the line and then spend time trying to figure out if it's a problem or not.

randombit (approved):

randombit:
Thanks for your patience working through the review process and addressing the many comments

Issue #4701, opened 2025-02-20
Surprising behavior of Ed25519_PrivateKey constructor


arekmula:
https://github.com/randombit/botan/blob/a2981b3c6c429b5047b8e97afc9d756387ea1e17/src/lib/pubkey/ed25519/ed25519_key.cpp#L94

Hey,
Is there any reason why `Ed25519_PrivateKey(const secure_vector<uint8_t>& secret_key)` uses the passed `secret_key` as a seed to the generator when its length is 32 bytes?
```
   if(secret_key.size() == 64) {
      m_private = secret_key;
      m_public.assign(m_private.begin() + 32, m_private.end());
   } else if(secret_key.size() == 32) {
      m_public.resize(32);
      m_private.resize(64);
      ed25519_gen_keypair(m_public.data(), m_private.data(), secret_key.data());
   } else {
      throw Decoding_Error("Invalid size for Ed25519 private key");
   }
}
```

Yesterday, I found it very surprising when debugging why our crypto flows are failing. 
One of our library's consumers passed us a byte array, which we didn't validate on input size before passing to this constructor. Our crypto flows started to fail, and after some debugging, I found out that the byte array that the consumer passed was 32-byte long. The first question that I asked myself was: "Why didn't we get an error indicating invalid size?". Then I looked into the `Ed25519_PrivateKey` constructor and spotted the surprising behavior that if the input size is 32-byte long, it's treated as a seed for keypair generation. 

I think that the design should be improved here, maybe by using factory functions?

randombit:
That's a good question. Checking `git` history it seems it's been done that way as long as `Ed25519_PrivateKey` has existed, and while I wrote that code I do not even recall now, 7 years on, if there was a specific requirement this be supported by the sponsor, or just something I decided to do.

I agree this is confusing, especially so as the behavior wasn't even documented in the header.

Does #4702 track with what you had in mind?

randombit:
#4702 deprecates the confusing (and previously poorly documented) constructor, and adds new functions that handle the key-from-seed and key-from-key situations.

PR #4702, opened 2025-02-20
Add named pseudo-constructors for Ed25519_PrivateKey

randombit:
The behavior of Ed25519_PrivateKey constructor taking a bytestring varies depending on the length. This behavior is long-standing but was not clearly documented.

Add new from_seed and from_bytes functions which verify that the input is of the expected length and then pass it to the pre-existing constructor.

GH #4701

### DIFF
--- a/src/lib/pubkey/ed25519/ed25519.h
+++ b/src/lib/pubkey/ed25519/ed25519.h
@@ -1,9 +1,7 @@
 /*
 * Ed25519
 * (C) 2017 Ribose Inc
-*
-* Based on the public domain code from SUPERCOP ref10 by
-* Peter Schwabe, Daniel J. Bernstein, Niels Duif, Tanja Lange, Bo-Yin Yang
+*     2025 Jack Lloyd
 *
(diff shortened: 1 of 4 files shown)


src/lib/pubkey/ed25519/ed25519.h
```diff
@@ -73,16 +74,41 @@ class BOTAN_PUBLIC_API(2, 2) Ed25519_PrivateKey final : public Ed25519_PublicKey
...
+      * If the input is 64 bytes long then it is treated as a pair of 32 byte
+      * values, first the private key and then the public key.
+      *
+      * If you are using only one of these paths, consider using instead
+      * from_seed or from_bytes.
+      *
       * @param secret_key the private key
       */
-      explicit Ed25519_PrivateKey(const secure_vector<uint8_t>& secret_key);
+      explicit Ed25519_PrivateKey(std::span<const uint8_t> secret_key);
```
arekmula, line 98: How about making this deprecated and redirect users to `from_seed` and `from_bytes`?

arekmula (approved):

PR #4704, opened 2025-02-20
FFI cleanups

randombit:


src/lib/ffi/ffi_pkey_algs.cpp
```diff
@@ -955,32 +1032,31 @@ int botan_pubkey_x448_get_pubkey(botan_pubkey_t key, uint8_t output[56]) {
...
+   }(key_len);
+
+   if(mode.has_value()) {
+      return ffi_guard_thunk(__func__, [=]() -> int {
+         auto kyber = std::make_unique<Botan::Kyber_PrivateKey>(std::span{privkey, key_len}, *mode);
+         *key = new botan_privkey_struct(std::move(kyber));
+         return BOTAN_FFI_SUCCESS;
+      });
+   } else {
+      return BOTAN_FFI_ERROR_BAD_PARAMETER;
```
reneme, line 1059:
```suggestion
   if(!mode.has_value()) {
      return BOTAN_FFI_ERROR_BAD_PARAMETER;
   }
   
   return ffi_guard_thunk(__func__, [=]() -> int {
      auto kyber = std::make_unique<Botan::Kyber_PrivateKey>(std::span{privkey, key_len}, *mode);
      *key = new botan_privkey_struct(std::move(kyber));
      return BOTAN_FFI_SUCCESS;
   });
```

... there's a parenthesis-fu that will probably break this patch if applied as is.

src/lib/ffi/ffi_pkey_algs.cpp
```diff
@@ -990,32 +1066,29 @@ int botan_privkey_load_kyber(botan_privkey_t* key, const uint8_t privkey[], size
...
+         return {};
+      }
+   }(key_len);
+
+   if(mode.has_value()) {
+      auto kyber = std::make_unique<Botan::Kyber_PublicKey>(std::span{pubkey, key_len}, *mode);
+      *key = new botan_pubkey_struct(std::move(kyber));
+      return BOTAN_FFI_SUCCESS;
+   } else {
+      return BOTAN_FFI_ERROR_BAD_PARAMETER;
```
reneme, line 1091: As above. Purely opinionated nitpick: Early return on error followed by the actual "happy path" at the very end of the function.

reneme (approved):

src/lib/ffi/ffi_pkey_algs.cpp
randombit: Current seems cleaner to me tbh

PR #4706, opened 2025-02-21
Remove PrimeOrderCurveId

randombit:
It wasn't doing much, but its existence implied quite a bit of per-curve boilerplate. Removing it simplifies matters slightly for users who wish to introduce a custom precompiled pcurve by patching the library.


randombit:
Next step would be extending `gen_ec_groups.py` so that `pcurves.cpp` and `pcurves_instance.h` are generated based on the contents of `ec_groups.txt`.

Step after that would be if a group is in `ec_groups.txt`, and `Impl` contains `pcurves`, and `pcurves_foo256v1` doesn't exist, `gen_ec_groups.py` creates it as a boilerplate curve.

At that point, an end user could insert their funky chicken curve into `ec_named.txt`, invoke `gen_ec_groups.py`, recompile the library and everything would just work.

reneme (approved):
Fully agree. I was anyway wondering why this indirection was necessary. Also, I like the fact that there are no dummy-definitions of the pcurve factory methods anymore. This struck me as odd and confusing.

❤️ Regarding the proposed path to auto-generating "chicken-curves" 🤣. As an alternative proposal to patching `ec_named.txt`: perhaps even make this a configure script feature, like `./configure.py --with-custom-curves=mycurves.txt`. One could potentially expose this through package managers (like Conan) and avoid having to fork recipes for a patch. Just an idea, perhaps we don't actually want to advertise this ability too much.

randombit:
TBH I would want to still require a manual patching step here as it hopefully raises the question to whoever is doing it of "Is this really a good idea? *Should* I be doing this?"

This is just about making that manual patching step not so inconvenient / prone to merge conflicts.

PR #4712, opened 2025-02-23
Add EC_Group::hash_to_curve_supported

randombit:
A bit nicer than having to invoke hash to curve and catching Not_Implemented


randombit:
Eh XMD situation makes this kind of tricky

PR #4714, opened 2025-02-25
Fix declaration in ffi.h

randombit:
In C (prior to C23) an empty parameter list means the function takes some undefined number/type of arguments, rather than no arguments.


reneme (approved):
Thanks.

PR #4723, opened 2025-02-27
Add support for ML-DSA in TLS

randombit:


randombit:
asio tests are failing which is somewhat unexpected, I'll look at this tomorrow

src/cli/tls_helpers.h
```diff
@@ -163,6 +161,12 @@ class Basic_Credentials_Manager : public Botan::Credentials_Manager {
                   continue;
                }
 
+               if(std::find(cert_signature_schemes.begin(),
+                            cert_signature_schemes.end(),
+                            i.certs[0].subject_public_key_algo()) == cert_signature_schemes.end()) {
+                  continue;
+               }
```
reneme, line 168: This is the reason that the ASIO tests fail to perform a successful handshake. The algorithm identifiers of the signature scheme involve the padding (e.g. "RSA/PKCS1v15(SHA-256)") but the subject public key's algorithm identifier is just that, e.g. "RSA", without the padding.

The ASIO integration tests happen to use an RSA certificate that doesn't pass this check. So the server fails to find a suitable certificate and raises a handshake failure.

KaganCanSit:
Hi @randombit,

I tried to develop a solution by addressing @reneme's comment explaining the reason for the ASIO test failures. I made the following changes.

**Changes:**
https://github.com/KaganCanSit/botan/commit/0f0afcfebbb8d9b8317a9eef305c3bf3064419ea

```cpp
#include <algorithm>

Botan::OID cert_oid = i.certs[0].subject_public_key_algo().oid();
bool compatible = 
	std::ranges::any_of(cert_signature_schemes, [&cert_oid](const Botan::AlgorithmIdentifier& scheme) {
		return scheme.oid() == cert_oid || (scheme.parameters_are_null_or_empty() &&
		  scheme.oid().to_formatted_string().starts_with(cert_oid.to_formatted_string()));
	});

if(!compatible) {
	continue;
}
```

Then I performed the following steps.

**Compilation:**
```bash
ninja clean && ./configure.py --without-documentation --with-boost --cc=clang --compiler-cache=ccache --build-targets=static,cli,tests --build-tool=ninja && ninja
```

**Test-1:**
```bash
./botan-test --test-threads=4 --run-long-tests
```

**Test-2:**
```bash
python3 src/scripts/test_cli.py ./botan cli_tls_socket_tests
```

Since I am not fully familiar with the functions here, I made improvements by checking the calls I could make. First, I compared the string components, but then I thought OID could be faster. I may have mistakes, but it can save you time and help you merge this PR content.

If you find it appropriate, I can create a PR or you can include it by cherry-pick.

PR #4725, opened 2025-02-28
Extended key usages are not set when creating PKCS#10 requests

neusdan:
Regression of #4032


randombit:
Looks good but please squash

neusdan:
Thanks for the review. Commits are squashed now.

randombit (approved):

PR #4731, opened 2025-02-28
Omit trailerField in RSA-PSS encoding

neusdan:
Issue found with [ZLint](https://github.com/zmap/zlint) --> https://github.com/zmap/zlint/blob/master/v3/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go

See https://datatracker.ietf.org/doc/html/rfc4055#section-3.1
See https://www.mozilla.org/en-US/about/governance/policies/security-group/certs/policy/#5-certificates

> Implementations that perform signature generation MUST omit the trailerField field, indicating that the default trailer field value was used.


randombit (approved):
TIL thanks

PR #4735, opened 2025-02-28
Use std::endian in loadstor.h

randombit:


randombit:
I suspect there may be some further simplifications possible here. In the old system we did have macros, but not all of the time. On certain platforms (like MIPS) by default we would not get a `IS_BIG_ENDIAN` *or* a `IS_LITTLE_ENDIAN` macro defined, because they can be either and it's a bit of a tossup. Whereas here we always know at compile time our endianness.

randombit:
I have no idea what to make of these CI failures. Can't replicate on my machine, everything work fine. It's as if `std::endian::native` was neither `big` nor `little`. Which ... wtf

src/lib/utils/loadstor.h
```diff
@@ -234,16 +213,16 @@ constexpr auto wrap_strong_type_or_enum(T t) {
  * This will be used only if the endianness of the target platform is unknown at
  * compile time.
```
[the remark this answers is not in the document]
randombit: If I change this to `consteval` both GCC and Clang reject and - assuming that  it's not that both of them are buggy - I don't understand why.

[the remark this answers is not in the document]
randombit: Oh I see it's because we invoke it from inside a lambda ... hm

randombit:
@reneme So both I'd like a review on this and also any insight you can provide why changing `fallback_store_any` to `consteval` does not work. `fallback_load_any` right now maybe can't be `consteval` just due to being invoked from within a lambda; but I don't see any reason at all `fallback_store_any` cannot be `constexpr` as it is only [outside of some tests atm] invoked from a `if(std::is_constant_evaluated())` block.

Also I have for the moment left the commits unsquashed, to see if you have any idea why my initial attempts failed. To me there is no logical difference between f9967fc and a5109c4 (and ftr both commits worked on my machine)

src/lib/utils/loadstor.h
```diff
@@ -55,6 +56,8 @@
 
 namespace Botan {
 
+static_assert(std::endian::native == std::endian::big || std::endian::native == std::endian::little);
```
reneme, line 59: Apparently, it is possible that `std::endian::native` is neither `big` nor `little`. At least that's what [the STL example in std::endian](https://en.cppreference.com/w/cpp/types/endian) implies. 😲 Perhaps we should add an error message to this static assert?

```suggestion
static_assert(std::endian::native == std::endian::big || std::endian::native == std::endian::little,
             "mixed-endian systems are not supported");
```

reneme (approved):
Changes look good to me. Certainly nice that we can rely on the STL for this! There are some unrelated clang-tidy nits (unnecessary `inline`, unnecessary forwarding reference, ...) that I would address independently.

##### Re: The failing `static_assert(native_endianness_is_unknown())`

Older compilers validate the `static_assert` even in `if constexpr` branches that are not taken. Starting from clang 17 and gcc 13 this isn't the case anymore. That behavior is utterly annoying! I'm not sure whether that was a bug or some errata that got fixed in the standard, frankly. See: https://godbolt.org/z/Majhn8xdh

##### Re: consteval fallback_store_any()

I don't think this has to do with any lambda, because lambdas are `constexpr` by default and shouldn't stand in the way of making this `consteval`. My first hypothesis was that this had to do with `if(std::is_constant_evaluated()) {}` vs. `if consteval {}`. But even when compiling in C++23 mode and changing to `if consteval` it still insists that all `store_be` are then immediate functions.

Probably we'd have to dig into the standard here. My guess is that all functions whose "side effects" (meaning in/out values; for the lack of better terminology) are dependent on an immediate function do inherit this immediate function property somehow. I guess that the compiler doesn't figure out that all in/out values are known at compile time in our case.

Perhaps that will be improved just like the `static_assert` vs. `if constexpr` quirk at one point. :(

src/lib/utils/loadstor.h
randombit: Yeah. C++'s continued insistence on supporting systems that haven't been commercially produced in decades is really obnoxious, especially when it comes at the cost of additional complexity for actual software running on actual computers.

Between 523b2a4ca48fa5 and 3dfa8f679bbc `CPUID` would throw upon initialization if it detected a mixed endian system at runtime.

randombit:
Thanks for looking into this

> Older compilers validate the static_assert even in if constexpr branches that are not taken.

What the fuck :see_no_evil: 

> Starting from clang 17 and gcc 13 this isn't the case anymore. 

OK well that decides what Botan4's minimum Clang version will be then!

randombit:
Thanks @lieser helpful to know the background on this.

The WG paper is gold

> As a result of static_assert(false) not working, people turn to workarounds. Which are, basically: how can I write false in a sufficiently complex way so as to confuse the compiler?

and

> Thanks to all the people over the years to whom I’ve had to explain why static_assert(false); doesn’t work and what they have to write instead for not immediately destroying their computers and switching to Rust.

src/lib/utils/loadstor.h
reneme: > ..., especially when it comes at the cost of additional complexity for actual software running on actual computers.

I guess that would have been okay, if at least the `std::endian` enum had explicitly listed `mixed` so that this is sort-of obvious. But then again: if we wouldn't have considered the mixed case out of ignorance, it would have probably "just worked" (™️). Until an actual mixed system would have come along, no?

src/lib/utils/loadstor.h
```diff
@@ -55,6 +56,9 @@
 
 namespace Botan {
 
+static_assert(std::endian::native == std::endian::big || std::endian::native == std::endian::little,
+              "Mixed endian systems are not supported");
```
reneme, line 60: Super nit: should that go into a more central header? `target_info.h` or one of its friends?

reneme (approved):
Thanks @lieser for digging up the `static_assert` background. Interesting how they bent to end-users' wishes there. And how the committee deemed it reasonable that `static_assert` behaved like this originally. 🤣 TIL!

src/lib/utils/loadstor.h
randombit: I am trying hard to keep header dependencies to a minimum for compilation time reasons, and I think it's likely this `static_assert` will never fire unless someone as a joke decides to cross compile for VAX. So leaving it here seems fine to me, especially since this is where the assumption re endian dualism are being made.

PR #4744, opened 2025-03-03
Avoid accidentally quadratic code during template processing in configure.py

randombit:
This especially affected generation of compile_commands.json and the Makefile

On my machine this cuts configure.py's runtime from 1.3s to .3s

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1860,6 +1860,7 @@ def process_template_string(template_text, variables, template_source):
             output = ""
             idx = 0
 
+            # pylint: disable=too-many-nested-blocks
             while idx < len(lines):
                 cond_match = self.cond_pattern.match(lines[idx])
                 for_match = self.for_pattern.match(lines[idx])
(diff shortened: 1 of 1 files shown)


configure.py
```diff
@@ -1911,7 +1911,9 @@ def insert_join(match):
                         else:
                             output += for_body.replace('%{i}', v).replace('%{i|upper}', v.upper())
 
-                        omitlast_match = self.omitlast_pattern.match(output)
+                        omitlast_match = None
+                        if output.find('%{omitlast') >= 0:
+                            omitlast_match = self.omitlast_pattern.match(output)
                         if omitlast_match:
                             output = omitlast_match.group(1)
```
reneme, line 1918: Nit: perhaps like that, to please the python gods and avoid pre-defining the `omitlast_match` variable?

```python
if '%{omitlast' in output:
    if omitlast_match := self.omitlast_pattern.match(output):
        output = omitlast_match.group(1)
        if i + 1 < len(var):
            output += omitlast_match.group(2)
        output += omitlast_match.group(3)
```

reneme: I tried with a ternary like: `match = ol_pattern.match() if "omitlast" in output else None`, but that does seem to execute the regex regardless of the `if` statement. :(

reneme (approved):
Oh my.

PR #4752, opened 2025-03-06
TLS 1.3 PQ hybrid NIST curve extensions and fixes

randombit:
Fix bugs that affected PQ suites using NIST curves in server mode. The expected public key lengths assumed a "fully compressed" format of just a single field element, which would have been the right thing to do in a historical sense but not what ended up getting standardized.

Add support for the secp384r1/ML-KEM-1024 TLS 1.3 suite

Add secp256r1/ML-KEM-768 and secp384r1/ML-KEM-1024 to the default offered suites.

I tested interop for  secp256r1/ML-KEM-768 and secp384r1/ML-KEM-1024 with OpenSSL master, which supports both suites. (`openssl s_client -groups SecP384r1MLKEM1024 ...`)

I was wondering how our existing tests did not notice this, we have a quite exhaustive set of tests that check we can talk to ourselves in a variety of scenarios .It turns out `unit_tls.cpp` was never updated for TLS 1.3 :( I will address this in another PR.


reneme:
Wow! But how come this worked against the liboqs test server?

reneme (approved):

randombit:
I think it was only the server that called the relevant functions

PR #4762, opened 2025-03-09
Run Valgrind in parallel

randombit:


randombit:
Nightly run here https://github.com/randombit/botan/actions/runs/13748966578 this change saves ~40-50 minutes per run

PR #4764, opened 2025-03-09
Run the cross build/qemu tests on Aarch64 runners

randombit:


randombit:
I figured for some architectures qemu on aarch64 might well be faster especially thanks to the extra 16 GPRs, but it was not to be. All times in seconds to execute just `botan-test`. Don't know if this is a CPU thing or qemu being better optimized for x86 or ... but in any case it's not a benefit to us right now. Probably worth trying again with qemu in 26.04 when that is out.

| Arch     | x86-64 | aarch64 |
|----------|--------|---------|
| alpha    | 401    |	701     |
| hppa64   | 1888   |	4509    |
| m68k     | 1434   |	3474    |
| mips     | 820    |	1040    |
| mips64   | 451    |	686     |
| ppc64    | 452    |	784     |
| riscv64  | 518    |	834     |
| s390x    | 1360   |	2120    |
| sh4		   | 1322   |	3385    |
| sparc64  | 1435   |	3128    |

PR #4777, opened 2025-03-19
Inline thunk calls in FFI

randombit:
This reduces the overall shared object size by over 2% on x86-64

| Compiler/Flags | Before (Bytes) | After (Bytes) | Savings (KB) |
| -------------- | -------------- | ------------- | ------------ |
| GCC 14 -Os     | 8167664        | 7943448       | 218          |
| Clang 19 -Os   | 8741000        | 8513688       | 221          |
| GCC 14 -O3     | 12055264       | 11792160      | 256          |
| Clang 19 -O3   | 10027248       | 9797856       | 224          |

### DIFF
--- a/src/lib/ffi/ffi_util.h
+++ b/src/lib/ffi/ffi_util.h
@@ -10,10 +10,10 @@
 #include <botan/exceptn.h>
 #include <botan/ffi.h>
 #include <botan/mem_ops.h>
+#include <concepts>
 #include <cstdint>
-#include <functional>
+#include <exception>
 #include <memory>
-#include <stdexcept>
 
 namespace Botan_FFI {
 
@@ -57,7 +57,11 @@ struct botan_struct {
    struct NAME final : public Botan_FFI::botan_struct<int, MAGIC> {}
 
 // Declared in ffi.cpp
-int ffi_error_exception_thrown(const char* func_name, const char* exn, int rc = BOTAN_FFI_ERROR_EXCEPTION_THROWN);
+void ffi_clear_last_exception();
+
+int ffi_error_exception_thrown(const char* func_name, const char* exn, int rc);
+
+int ffi_error_exception_thrown(const char* func_name, const char* exn, Botan::ErrorType err);
 
 template <typename T, uint32_t M>
 T& safe_get(botan_struct<T, M>* p) {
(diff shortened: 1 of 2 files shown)


src/lib/ffi/ffi_util.h
```diff
@@ -75,7 +79,24 @@ T& safe_get(botan_struct<T, M>* p) {
    throw FFI_Error("Invalid object pointer", BOTAN_FFI_ERROR_INVALID_OBJECT);
 }
 
-int ffi_guard_thunk(const char* func_name, const std::function<int()>& thunk);
+template <typename T>
```
reneme, line 82: Perhaps restrict this with a concept? `std::invocable<>` should do.

reneme:
I'm surprised that introducing an inlined template instead of a function call reduces the binary size like this.

Could you elaborate a bit on your method in general, please?

randombit:
I agree it seems quite silly that this provides savings considering it dumps the entire try/catch machinery into each function.

https://github.com/google/bloaty

On master

```
$ bloaty -d symbols -n 10 libbotan-3.so.8
    FILE SIZE        VM SIZE
 --------------  --------------
  89.3%  8.77Mi  91.0%  7.28Mi    [8161 Others]
   2.5%   251Ki   3.1%   251Ki    [section .gcc_except_table]
   1.6%   158Ki   1.2%  98.6Ki    std::_Function_handler<>::_M_invoke()
   1.4%   135Ki   0.0%       0    [section .symtab]
   1.0%   100Ki   0.4%  31.9Ki    std::_Function_handler<>::_M_manager()
   0.8%  81.1Ki   1.0%  79.7Ki    std::__shared_ptr<>::__shared_ptr<>()
   0.8%  79.1Ki   0.4%  29.9Ki    Botan_FFI::botan_ffi_visit<>()::{lambda()#1}
...
```

I have no idea what `_M_invoke` and `_M_manager` are doing but clearly it adds up over all N FFI functions.

In this branch almost all of that overhead is gone (we use `std::function` elsewhere so it's not removed entirely)

```
$ bloaty -n 100 -d symbols libbotan-3.so.8 | grep _Function_handler
   0.3%  29.0Ki   0.2%  14.9Ki    std::_Function_handler<>::_M_invoke()
   0.2%  20.7Ki   0.1%  5.31Ki    std::_Function_handler<>::_M_manager()
```

The sizes of individual FFI functions does often (not always) increase, but it's still a net win on the order of 200 KB. Also with the (size) cost of the function being contained entirely within that function, rather than chunks being delegated off to a template instantiation, makes further (size) profiling more effective.

It would be worth checking how things work out with libc++ and/or MSVC STL.

It's also possible the win comes from avoiding `std::function` and having `ffi_guard_thunk` take a plain C-style function pointer.

src/lib/ffi/ffi_util.h
randombit: Done

randombit:
Checked on a FreeBSD aarch64 machine, Clang 18

master: 10067464 bytes
here: 9182440 bytes

864 Kb ...

randombit:
OpenBSD x86-64 machine, Clang 16

master: 11069304
here: 10157152

890 KB

randombit:
Linux ppc64le, GCC 11

master: 12511200
here: 12274296

231 KB

randombit:
FreeBSD 15, x86-64, Clang 18

master: 10296232
here: 9377072

897 KB

randombit:
It seems like if anything libc++ has a much bigger space savings from this change. I wanted to also check a macOS machine but I don't have online/interactive access to a macOS machine with a recent enough XCode installed.

randombit:
Checked Linux x86-64 with Clang -stdlib=libc++ there again 850+ KB savings

reneme (approved):

reneme:
Thanks for the insights. Certainly interesting finding.

Number for macOS (14.7, Xcode 16):

here: 7565168
master: 8315824

733kiB

randombit:
Windows x86-64 DLL also ~200 KB saved from this change

master:   File './botan-3.dll' is 6317.00 kB
here: File './botan-3.dll' is 6128.50 kB

PR #4783, opened 2025-03-22
Update deprecated mem_ops.h function calls to use std::span-based versions and resolve some TODOs

KaganCanSit:
Hello,

This draft PR includes updates to modernize the codebase by replacing deprecated mem_ops.h function calls with their `std::span`-based counterparts in line with C++20 standards. Additionally, I also discussed some TODO comments that I thought would be quick fixes.

My reference points the [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines) and [CppReference](https://en.cppreference.com/w/). Since I am not actively developing in C++20 at the moment, I am marking this PR as a draft to review and feedback.

Changes made:
- Replaced xor_buf calls with std::span-based equivalents.
- Handled several smaller TODOs identified in the codebase.

Confirmed that builds and test suites pass with:
```
ninja clean && ./configure.py --without-documentation --cc=clang --compiler-cache=ccache --build-targets=static,cli,tests --build-tool=ninja && ninja && ./botan-test
```

Check code format with:
```
```

---
**Outstanding considerations:**
I intentionally did not modify the [following function](https://github.com/randombit/botan/blob/e1d1cb2c837f656c3b600ffa3eaa926c3f07a410/src/lib/utils/mem_ops.h#L424), as it touches multiple areas and could introduce bugs without extensive review:
```cpp
// TODO: deprecate and replace, use .subspan()
inline void xor_buf(std::span<uint8_t> out, std::span<const uint8_t> in, size_t n) {
   BOTAN_ARG_CHECK(out.size() >= n, "output span is too small");
   BOTAN_ARG_CHECK(in.size() >= n, "input span is too small");
   xor_buf(out.first(n), in.first(n));
}
```
Files affected by this function;
```
src/lib/modes/aead/eax/eax.cpp
src/lib/mac/cmac/cmac.cpp
src/lib/pubkey/dlies/dlies.cpp
```

---
**Additional notes:**
There is a change in TLS, so additional checking may be required.

Thanks in advance for your time and guidance.

### DIFF
--- a/src/lib/mac/cmac/cmac.cpp
+++ b/src/lib/mac/cmac/cmac.cpp
@@ -25,13 +25,13 @@ void CMAC::add_data(std::span<const uint8_t> input) {
    copy_mem(m_buffer.data() + m_position, input.data(), initial_fill);
 
    if(m_position + input.size() > bs) {
-      xor_buf(m_state, m_buffer, bs);
+      xor_buf(m_state, std::span{m_buffer}.first(bs));
       m_cipher->encrypt(m_state);
 
       BufferSlicer in(input);
       in.skip(bs - m_position);
       while(in.remaining() > bs) {
-         xor_buf(m_state, in.take(bs), bs);
+         xor_buf(m_state, in.take(bs));
          m_cipher->encrypt(m_state);
       }
 
@@ -47,13 +47,13 @@ void CMAC::add_data(std::span<const uint8_t> input) {
 * Finalize an CMAC Calculation
 */
 void CMAC::final_result(std::span<uint8_t> mac) {
-   xor_buf(m_state, m_buffer, m_position);
+   xor_buf(std::span{m_state}.first(m_position), std::span{m_buffer}.first(m_position));
 
    if(m_position == output_length()) {
-      xor_buf(m_state, m_B, output_length());
+      xor_buf(m_state, m_B);
    } else {
       m_state[m_position] ^= 0x80;
-      xor_buf(m_state, m_P, output_length());
+      xor_buf(m_state, m_P);
    }
 
    m_cipher->encrypt(m_state);
--- a/src/lib/mac/x919_mac/x919_mac.cpp
+++ b/src/lib/mac/x919_mac/x919_mac.cpp
@@ -21,7 +21,7 @@ void ANSI_X919_MAC::add_data(std::span<const uint8_t> input) {
    BufferSlicer in(input);
 
    const auto to_be_xored = in.take(std::min(8 - m_position, in.remaining()));
-   xor_buf(&m_state[m_position], to_be_xored.data(), to_be_xored.size());
+   xor_buf(std::span{m_state}.subspan(m_position, to_be_xored.size()), to_be_xored);
    m_position += to_be_xored.size();
 
    if(m_position < 8) {
@@ -30,12 +30,12 @@ void ANSI_X919_MAC::add_data(std::span<const uint8_t> input) {
 
    m_des1->encrypt(m_state);
    while(in.remaining() >= 8) {
-      xor_buf(m_state, in.take(8).data(), 8);
+      xor_buf(std::span{m_state}.first<8>(), in.take<8>());
       m_des1->encrypt(m_state);
    }
 
    const auto remaining = in.take(in.remaining());
-   xor_buf(m_state, remaining.data(), remaining.size());
+   xor_buf(std::span{m_state}.first(remaining.size()), remaining);
    m_position = remaining.size();
 }
 
--- a/src/lib/misc/roughtime/roughtime.cpp
+++ b/src/lib/misc/roughtime/roughtime.cpp
@@ -47,7 +47,8 @@ template <typename T>
 T copy(const uint8_t* t)
    requires(is_array<T>::value)
 {
-   return typecast_copy<T>(t);  //arrays are endianness independent, so we do a memcpy
+   //arrays are endianness independent, so we do a memcpy
+   return typecast_copy<T>(std::span<const uint8_t, sizeof(T)>(t, sizeof(T)));
 }
 
 template <typename T>
@@ -146,10 +147,7 @@ void hashNode(std::span<uint8_t, 64> hash, std::span<const uint8_t, 64> node, bo
 
 template <size_t N, typename T>
 std::array<uint8_t, N> vector_to_array(std::vector<uint8_t, T> vec) {
-   if(vec.size() != N) {
-      throw std::logic_error("Invalid vector size");
-   }
-   return typecast_copy<std::array<uint8_t, N>>(vec.data());
+   return typecast_copy<std::array<uint8_t, N>>(vec);
 }
 }  // namespace
 
@@ -159,7 +157,7 @@ Nonce::Nonce(const std::vector<uint8_t>& nonce) : m_nonce{} {
    if(nonce.size() != 64) {
       throw Invalid_Argument("Roughtime nonce must be 64 bytes long");
    }
-   m_nonce = typecast_copy<std::array<uint8_t, 64>>(nonce.data());
+   m_nonce = typecast_copy<std::array<uint8_t, 64>>(nonce);
 }
 
 Nonce::Nonce(RandomNumberGenerator& rng) : m_nonce(rng.random_array<64>()) {}
--- a/src/lib/modes/aead/eax/eax.cpp
+++ b/src/lib/modes/aead/eax/eax.cpp
@@ -144,13 +144,13 @@ void EAX_Encryption::finish_msg(secure_vector<uint8_t>& buffer, size_t offset) {
    update(buffer, offset);
 
    secure_vector<uint8_t> data_mac = m_cmac->final();
-   xor_buf(data_mac, m_nonce_mac, data_mac.size());
+   xor_buf(data_mac, m_nonce_mac);
 
    if(m_ad_mac.empty()) {
       m_ad_mac = eax_prf(1, block_size(), *m_cmac, nullptr, 0);
    }
 
-   xor_buf(data_mac, m_ad_mac, data_mac.size());
+   xor_buf(data_mac, m_ad_mac);
 
    buffer += std::make_pair(data_mac.data(), tag_size());
 
--- a/src/lib/modes/xts/xts.cpp
+++ b/src/lib/modes/xts/xts.cpp
@@ -193,9 +193,9 @@ void XTS_Encryption::finish_msg(secure_vector<uint8_t>& buffer, size_t offset) {
       buffer.resize(full_blocks + offset);
       update(buffer, offset);
 
-      xor_buf(last, tweak(), BS);
+      xor_buf(std::span(last).first(BS), std::span(tweak(), BS));
       cipher().encrypt(last);
-      xor_buf(last, tweak(), BS);
+      xor_buf(std::span(last).first(BS), std::span(tweak(), BS));
 
       for(size_t i = 0; i != final_bytes - BS; ++i) {
          last[i] ^= last[i + BS];
@@ -203,9 +203,9 @@ void XTS_Encryption::finish_msg(secure_vector<uint8_t>& buffer, size_t offset) {
          last[i] ^= last[i + BS];
       }
 
-      xor_buf(last, tweak() + BS, BS);
+      xor_buf(std::span(last).first(BS), std::span(tweak() + BS, BS));
       cipher().encrypt(last);
-      xor_buf(last, tweak() + BS, BS);
+      xor_buf(std::span(last).first(BS), std::span(tweak() + BS, BS));
 
       buffer += last;
    }
@@ -228,9 +228,9 @@ size_t XTS_Decryption::process_msg(uint8_t buf[], size_t sz) {
       const size_t to_proc = std::min(blocks, blocks_in_tweak);
       const size_t proc_bytes = to_proc * BS;
 
-      xor_buf(buf, tweak(), proc_bytes);
+      xor_buf(std::span(buf, proc_bytes), std::span(tweak(), proc_bytes));
       cipher().decrypt_n(buf, buf, to_proc);
-      xor_buf(buf, tweak(), proc_bytes);
+      xor_buf(std::span(buf, proc_bytes), std::span(tweak(), proc_bytes));
 
       buf += proc_bytes;
       blocks -= to_proc;
@@ -263,9 +263,9 @@ void XTS_Decryption::finish_msg(secure_vector<uint8_t>& buffer, size_t offset) {
       buffer.resize(full_blocks + offset);
       update(buffer, offset);
 
-      xor_buf(last, tweak() + BS, BS);
+      xor_buf(std::span(last).first(BS), std::span(tweak() + BS, BS));
       cipher().decrypt(last);
-      xor_buf(last, tweak() + BS, BS);
+      xor_buf(std::span(last).first(BS), std::span(tweak() + BS, BS));
 
       for(size_t i = 0; i != final_bytes - BS; ++i) {
          last[i] ^= last[i + BS];
@@ -273,9 +273,9 @@ void XTS_Decryption::finish_msg(secure_vector<uint8_t>& buffer, size_t offset) {
          last[i] ^= last[i + BS];
       }
 
-      xor_buf(last, tweak(), BS);
+      xor_buf(std::span(last).first(BS), std::span(tweak(), BS));
       cipher().decrypt(last);
-      xor_buf(last, tweak(), BS);
+      xor_buf(std::span(last).first(BS), std::span(tweak(), BS));
 
       buffer += last;
    }
--- a/src/lib/tls/tls12/tls_record.cpp
+++ b/src/lib/tls/tls12/tls_record.cpp
@@ -15,6 +15,7 @@
 #include <botan/rng.h>
 #include <botan/tls_ciphersuite.h>
 #include <botan/tls_exceptn.h>
+#include <botan/internal/concat_util.h>
 #include <botan/internal/fmt.h>
 #include <botan/internal/loadstor.h>
 #include <botan/internal/tls_seq_numbers.h>
@@ -108,16 +109,13 @@ std::vector<uint8_t> Connection_Cipher_State::aead_nonce(uint64_t seq, RandomNum
       case Nonce_Format::AEAD_XOR_12: {
          BOTAN_ASSERT_NOMSG(m_nonce.size() == 12);
          std::vector<uint8_t> nonce(12);
-         store_be(seq, nonce.data() + 4);
-         xor_buf(nonce, m_nonce.data(), m_nonce.size());
+         store_be(seq, std::span{nonce}.last<8>());
+         xor_buf(nonce, m_nonce);
          return nonce;
       }
       case Nonce_Format::AEAD_IMPLICIT_4: {
          BOTAN_ASSERT_NOMSG(m_nonce.size() == 4);
-         std::vector<uint8_t> nonce(12);
-         copy_mem(&nonce[0], m_nonce.data(), 4);  // NOLINT(*container-data-pointer)
-         store_be(seq, &nonce[nonce_bytes_from_handshake()]);
-         return nonce;
+         return concat(m_nonce, store_be(seq));
       }
    }
 
@@ -139,8 +137,8 @@ std::vector<uint8_t> Connection_Cipher_State::aead_nonce(const uint8_t record[],
       case Nonce_Format::AEAD_XOR_12: {
          BOTAN_ASSERT_NOMSG(m_nonce.size() == 12);
          std::vector<uint8_t> nonce(12);
-         store_be(seq, nonce.data() + 4);
-         xor_buf(nonce, m_nonce.data(), m_nonce.size());
+         store_be(seq, std::span{nonce}.last<8>());
+         xor_buf(nonce, m_nonce);
          return nonce;
       }
       case Nonce_Format::AEAD_IMPLICIT_4: {
@@ -148,10 +146,7 @@ std::vector<uint8_t> Connection_Cipher_State::aead_nonce(const uint8_t record[],
          if(record_len < nonce_bytes_from_record()) {
             throw Decoding_Error("Invalid AEAD packet too short to be valid");
          }
-         std::vector<uint8_t> nonce(12);
-         copy_mem(&nonce[0], m_nonce.data(), 4);  // NOLINT(*container-data-pointer)
-         copy_mem(&nonce[nonce_bytes_from_handshake()], record, nonce_bytes_from_record());
-         return nonce;
+         return concat(m_nonce, std::span{record, record_len}.first(nonce_bytes_from_record()));
       }
    }
 
--- a/src/lib/utils/mem_ops.h
+++ b/src/lib/utils/mem_ops.h
@@ -213,7 +213,7 @@ inline constexpr ToT typecast_copy(const FromR& src) {
    return dst;
 }
 
-// TODO: deprecate and replace
+BOTAN_DEPRECATED("This function is deprecated without a replacement")
 template <typename T>
 inline constexpr void typecast_copy(uint8_t out[], T in[], size_t N)
    requires std::is_trivially_copyable_v<T>
@@ -222,7 +222,7 @@ inline constexpr void typecast_copy(uint8_t out[], T in[], size_t N)
    typecast_copy(std::span<uint8_t>(out, sizeof(T) * N), std::span<const T>(in, N));
 }
 
-// TODO: deprecate and replace
+BOTAN_DEPRECATED("This function is deprecated without a replacement")
 template <typename T>
 inline constexpr void typecast_copy(T out[], const uint8_t in[], size_t N)
    requires std::is_trivial_v<T>
@@ -231,14 +231,14 @@ inline constexpr void typecast_copy(T out[], const uint8_t in[], size_t N)
    typecast_copy(std::span<T>(out, N), std::span<const uint8_t>(in, N * sizeof(T)));
 }
 
-// TODO: deprecate and replace
+BOTAN_DEPRECATED("This function is deprecated without a replacement")
 template <typename T>
 inline constexpr void typecast_copy(uint8_t out[], const T& in) {
    // asserts that *out points to the correct amount of memory
    typecast_copy(std::span<uint8_t, sizeof(T)>(out, sizeof(T)), in);
 }
 
-// TODO: deprecate and replace
+BOTAN_DEPRECATED("This function is deprecated without a replacement")
 template <typename T>
    requires std::is_trivial_v<std::decay_t<T>>
 inline constexpr void typecast_copy(T& out, const uint8_t in[]) {
@@ -246,7 +246,7 @@ inline constexpr void typecast_copy(T& out, const uint8_t in[]) {
    typecast_copy(out, std::span<const uint8_t, sizeof(T)>(in, sizeof(T)));
 }
 
-// TODO: deprecate and replace
+BOTAN_DEPRECATED("This function is deprecated without a replacement")
 template <typename To>
    requires std::is_trivial_v<To>
 inline constexpr To typecast_copy(const uint8_t src[]) noexcept {
@@ -417,14 +417,14 @@ inline void xor_buf(uint8_t out[], const uint8_t in[], const uint8_t in2[], size
    xor_buf(std::span{out, length}, std::span{in, length}, std::span{in2, length});
 }
 
-// TODO: deprecate and replace, use .subspan()
+BOTAN_DEPRECATED("This function is deprecated without a replacement")
 inline void xor_buf(std::span<uint8_t> out, std::span<const uint8_t> in, size_t n) {
    BOTAN_ARG_CHECK(out.size() >= n, "output span is too small");
    BOTAN_ARG_CHECK(in.size() >= n, "input span is too small");
    xor_buf(out.first(n), in.first(n));
 }
 
-// TODO: deprecate and replace, use .subspan()
+BOTAN_DEPRECATED("This function is deprecated without a replacement")
 template <typename Alloc>
 void xor_buf(std::vector<uint8_t, Alloc>& out, const uint8_t* in, size_t n) {
    BOTAN_ARG_CHECK(out.size() >= n, "output vector is too small");
@@ -432,7 +432,7 @@ void xor_buf(std::vector<uint8_t, Alloc>& out, const uint8_t* in, size_t n) {
    xor_buf(std::span{out}.first(n), std::span{in, n});
 }
 
-// TODO: deprecate and replace
+BOTAN_DEPRECATED("This function is deprecated without a replacement")
 template <typename Alloc, typename Alloc2>
 void xor_buf(std::vector<uint8_t, Alloc>& out, const uint8_t* in, const std::vector<uint8_t, Alloc2>& in2, size_t n) {
    BOTAN_ARG_CHECK(out.size() >= n, "output vector is too small");
--- a/src/tests/test_roughtime.cpp
+++ b/src/tests/test_roughtime.cpp
@@ -38,9 +38,9 @@ class Roughtime_Request_Tests final : public Text_Based_Test {
 
          // TODO should use byte comparison function
          if(type == "Valid") {
-            result.test_is_true("encode", request == Botan::typecast_copy<std::array<uint8_t, 1024>>(request_v.data()));
+            result.test_is_true("encode", request == Botan::typecast_copy<std::array<uint8_t, 1024>>(request_v));
          } else {
-            result.test_is_true("encode", request != Botan::typecast_copy<std::array<uint8_t, 1024>>(request_v.data()));
+            result.test_is_true("encode", request != Botan::typecast_copy<std::array<uint8_t, 1024>>(request_v));
          }
 
          return result;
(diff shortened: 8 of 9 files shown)


KaganCanSit:
[ci / Analysis (coverage, gcc, ubuntu-24.04) (pull_request)](https://github.com/randombit/botan/actions/runs/14004841261/job/39217312363?pr=4783)  Failing after 3m`

I see that GCC handles std::span quite differently than Clang. I'll do some research on that.

src/lib/pubkey/pqcrystals/pqcrystals.h
reneme, line 1: The changes in this file are independent of the rest of the pull request and need additional consideration. Please revert.

src/lib/tls/tls12/tls_record.cpp
```diff
@@ -89,13 +89,13 @@ std::vector<uint8_t> Connection_Cipher_State::aead_nonce(uint64_t seq, RandomNum
       case Nonce_Format::AEAD_XOR_12: {
          std::vector<uint8_t> nonce(12);
          store_be(seq, nonce.data() + 4);
-         xor_buf(nonce, m_nonce.data(), m_nonce.size());
+         xor_buf(std::span(nonce), std::span(m_nonce));
```
reneme, line 92:
```suggestion
         xor_buf(nonce, m_nonce);
```

... vectors are ranges by itself. No need for the spans.

src/lib/tls/tls12/tls_record.cpp
```diff
@@ -89,13 +89,13 @@ std::vector<uint8_t> Connection_Cipher_State::aead_nonce(uint64_t seq, RandomNum
...
          return nonce;
```
reneme, line 93: While we're on it:

```C++
std::vector<uint8_t> nonce(12);
store_be(seq, std::span{nonce}.last<8>());
xor_buf(nonce, m_nonce);
return nonce;
```

src/lib/tls/tls12/tls_record.cpp
```diff
@@ -89,13 +89,13 @@ std::vector<uint8_t> Connection_Cipher_State::aead_nonce(uint64_t seq, RandomNum
...
       }
       case Nonce_Format::AEAD_IMPLICIT_4: {
          BOTAN_ASSERT_NOMSG(m_nonce.size() == 4);
          std::vector<uint8_t> nonce(12);
-         copy_mem(&nonce[0], m_nonce.data(), 4);
+         copy_mem(std::span<uint8_t>(nonce.data(), 4), std::span<const uint8_t>(m_nonce.data(), 4));
          store_be(seq, &nonce[nonce_bytes_from_handshake()]);
          return nonce;
```
reneme, line 100: Maybe:

```C++
BOTAN_ASSERT_NOMSG(m_nonce.size() == 4);
std::vector<uint8_t> nonce(12);
copy_mem(std::span{nonce}.first<4>(), m_nonce);
store_be(seq, std::span{nonce}.last<8>());
return nonce;
```

... my understanding is that `nonce_bytes_from_handshake()` is always equal to `m_nonce.size()` and we assert that the latter is "4" in this case. So that should be safe.

src/lib/tls/tls12/tls_record.cpp
```diff
@@ -121,7 +121,7 @@ std::vector<uint8_t> Connection_Cipher_State::aead_nonce(const uint8_t record[],
       case Nonce_Format::AEAD_XOR_12: {
          std::vector<uint8_t> nonce(12);
          store_be(seq, nonce.data() + 4);
-         xor_buf(nonce, m_nonce.data(), m_nonce.size());
+         xor_buf(std::span(nonce), std::span(m_nonce));
          return nonce;
```
reneme, line 125:
```C++
std::vector<uint8_t> nonce(12);
store_be(seq, std::span{nonce}.last<8>());
xor_buf(nonce, m_nonce);
return nonce;
```

... as above.

src/lib/tls/tls12/tls_record.cpp
```diff
@@ -130,8 +130,9 @@ std::vector<uint8_t> Connection_Cipher_State::aead_nonce(const uint8_t record[],
             throw Decoding_Error("Invalid AEAD packet too short to be valid");
          }
          std::vector<uint8_t> nonce(12);
-         copy_mem(&nonce[0], m_nonce.data(), 4);
-         copy_mem(&nonce[nonce_bytes_from_handshake()], record, nonce_bytes_from_record());
+         copy_mem(std::span(nonce.data(), 4), std::span(m_nonce.data(), 4));
+         copy_mem(std::span(nonce.data() + nonce_bytes_from_handshake(), nonce_bytes_from_record()),
+                  std::span(record, nonce_bytes_from_record()));
          return nonce;
```
reneme, line 136:
```C++
std::vector<uint8_t> nonce(12);
copy_mem(std::span{nonce}.first<4>(), m_nonce);
copy_mem(std::span{nonce}.last(nonce_bytes_from_record()),
         std::span{record, record_len}.first(nonce_bytes_from_record()));
return nonce;
```

reneme (commented):
Some suggestions regarding the effective usage of `std::span`.

src/lib/tls/tls12/tls_record.cpp
reneme: Perhaps even (using tools from `stl_util.h`):

```C++
return concat(m_nonce, std::span{record, record_len}.first(nonce_bytes_from_record()));
```

src/lib/tls/tls12/tls_record.cpp
reneme: Or even like that (using `stl_util.h`):

```C++
BOTAN_ASSERT_NOMSG(m_nonce.size() == 4);
return concat(m_nonce, store_be(seq));
```

KaganCanSit:
Thank you for your input and guidance. I will incorporate the suggested changes, verify the behavior through testing, and submit a review request accordingly.

src/lib/utils/mem_ops.h
```diff
@@ -218,7 +218,7 @@ inline constexpr ToT typecast_copy(FromR&& src) noexcept {
    return dst;
 }
 
-// TODO: deprecate and replace
+BOTAN_DEPRECATED("This function is deprecated, use span version instead")
```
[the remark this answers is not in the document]
KaganCanSit: You are right. My wording is not correct according to the C++ standard and the structure of the functions. I have made the changes to correct it.

src/lib/x509/certstor_system_macos/certstor_macos.cpp
```diff
@@ -67,10 +67,8 @@ class scoped_CFType {
 X509_DN normalize(const X509_DN& dn) {
    X509_DN result;
 
-   for(const auto& rdn : dn.dn_info()) {
-      // TODO: C++14 - use std::get<ASN1_String>(), resp. std::get<OID>()
-      const auto oid = rdn.first;
-      auto str = rdn.second;
+   for(const auto& [oid, str_val] : dn.dn_info()) {
+      auto str = str_val;  // Create a modifiable copy
```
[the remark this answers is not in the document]
KaganCanSit: This change really looks more readable and elegant. I didn't want to edit it to avoid making too many changes.

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -204,8 +205,12 @@ Response Response::from_bits(const std::vector<uint8_t>& response, const Nonce&
...
+
+   std::span<const uint8_t> path_span{path};
+
    while(level < levels) {
-      hashNode(hash, typecast_copy<std::array<uint8_t, 64>>(path.data() + level * 64), index & 1);
+      auto node_data = path_span.subspan(level * 64, 64);
+      hashNode(hash, typecast_copy<std::array<uint8_t, 64>>(node_data), index & 1);
       ++level;
       index >>= 1;
    }
```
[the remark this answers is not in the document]
KaganCanSit: The approach here is really more efficient and readable. I will also look into your question. Thank you.

src/lib/tls/tls12/tls_record.cpp
KaganCanSit: Obviously I think it should be safe too, but I'll take the second way. There might be something I'm not seeing.

src/lib/tls/tls12/tls_record.cpp
KaganCanSit: Definitely cleaner and more readable. Thanks for the guidance.

src/lib/utils/mem_ops.h
```diff
@@ -427,15 +427,15 @@ inline void xor_buf(std::span<uint8_t> out, std::span<const uint8_t> in, size_t
    xor_buf(out.first(n), in.first(n));
 }
```
reneme, line 428: Did you leave this function out intentionally? (I'm commenting its closing bracket due to a review limitation of GitHub.)

Could also be deprecated, in my opinion:

```C++
BOTAN_DEPRECATED("This function is deprecated, use range-based version and std::span::first")
```

src/lib/mac/x919_mac/x919_mac.cpp
```diff
@@ -30,12 +30,12 @@ void ANSI_X919_MAC::add_data(std::span<const uint8_t> input) {
 
    m_des1->encrypt(m_state);
    while(in.remaining() >= 8) {
-      xor_buf(m_state, in.take(8).data(), 8);
+      xor_buf(std::span{m_state}.first<8>(), in.take<8>());
```
reneme, line 33: Just for the record: I just realized that `m_state` is actually 8 bytes wide by definition. Technically, this should work too:

```suggestion
      xor_buf(m_state, in.take<8>());
```

... BUT this change would incur a runtime length check for every iteration, because passing a `secure_vector` does not carry the compile-time length information.

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -149,7 +151,7 @@ std::array<uint8_t, N> vector_to_array(std::vector<uint8_t, T> vec) {
    if(vec.size() != N) {
       throw std::logic_error("Invalid vector size");
    }
-   return typecast_copy<std::array<uint8_t, N>>(vec.data());
+   return typecast_copy<std::array<uint8_t, N>>(vec);
```
reneme, line 154: Actually the range-based `typecast_copy` will do a length check to ensure that the input and output buffers match. So we could get rid of the error check here. Sorry, I missed that yesterday.

reneme (approved):
Looks good to me now. Thank you for taking this on!

For the record: The `mem_ops.h` header is marked as "future internal header". So, technically, we don't even need to mark the `typecast_copy` and `xor_buf` overloads as "deprecated", because for library users the entire header is going to go away in a future major release.

Personally, I still think your change in `mem_ops.h` makes sense as-is, for documentation purposes.

@randombit, over to you. :)

src/lib/utils/mem_ops.h
KaganCanSit: ​In the "Outstanding Considerations" section of the Pull Request description, I indicated my intention to deprecate a specific this function. Following various adjustments, I observed that this change adversely affected multiple tests, leading to failures. Despite exploring different approaches, I remained uncertain about certain aspects. Consequently, to mitigate the risk of errors, I deemed it prudent to exclude the impacted files from this merge request.​

However, since only this function remains, I can revisit it with additional support or create a separate merge request for it after this one. This approach would allow us to review it together as a Draft Pull Request. I am more motivated to pursue a separate request for this function.

src/lib/misc/roughtime/roughtime.cpp
KaganCanSit: It doesn't matter. What matters to me is the information I get from here. I'm going through it again and I'll edit it in a moment.

src/lib/mac/x919_mac/x919_mac.cpp
KaganCanSit: You are right, if we make this change, the length check will be repeated every iteration due to typecast_copy. We made some changes accordingly in the previous review.

Such unnecessary runtime checks can lead to performance loss and conflict with the goal of optimized readable code. Therefore, we could push this check to compile time by defining a length constant constexpr and using a fixed-size span like std::span<uint8_t, N>.

However, this solution may slightly reduce the readability of the code. There is an expression in Turkish. “The stone you throw is not worth the bird you scare.” (An idiom used when the resources used to do a job are more than the benefit you will get as a result of the job). Consider the following example:

constexpr size_t STATE_SIZE = 8;
xor_buf(std::span<uint8_t, STATE_SIZE>(m_state.data(), STATE_SIZE), in.take<STATE_SIZE>());

I am not very sure about readability; I think this point should be taken into account if changes are made.

reneme: Yeah, I agree. This can and should be tackled when we explicitly refactor this class. Fine by me to leave as-is (i.e. with the magic numbers) for now.

src/lib/utils/mem_ops.h
reneme: Alright. Understood. Let's not touch it for now. As stated before, this entire header will go away from the public API anyway. No need to over-complicate things now. Thanks.

KaganCanSit:
The conflict below was resolved by re-basing the branch with the master branch.

image

KaganCanSit:
Hi @reneme,

I've been reviewing this PR during my free time. While rebasing my recent PRs to the current master branch, I realized I could also address the function I had previously left out of scope. I'm not sure, but I don't have as many problems as before. Maybe some indirect changes made it easier to do. I mentioned this in the PR description and we [discussed it here](https://github.com/randombit/botan/pull/4783#discussion_r2018059308).

I worked on this after hours today and have now removed all TODO messages while implementing the deprecation notifications. The changes pass ./botan-test successfully in my local environment.
Since you've already reviewed this PR and it will require re-review, I wanted to consult with you before proceeding. Here's the [commit](https://github.com/KaganCanSit/botan/commit/804f83ab3d2c3e754b6cf8cb17754c8281f0feb6) containing my changes.

I'd like to request cherry-picking these changes to this branch, which would allow us to remove all TODO messages and publish the deprecation notifications together. This PR is closer to completion than my other requests, which motivates me to finalize it.

Since I don't actively use C++20 here, I may have mistakes. Therefore, a re-check will be needed.

What are your thoughts?

Best regards.

KaganCanSit:
Hello again Mr. @reneme,

I included the commit that I made improvements for the function I mentioned in the comment above, which I did not change, in the branch here. I also rebase to the master branch. I expect the tests to be successful, I will follow it.

I have not squashed the commits so that you can easily analyze the last parts I added here. Please let me know if there are any changes you want and I will try to fix them as soon as I get a chance. I think this branch is close to being merged.

After review, I will squash the commits and make them ready for merging.

Best regards.

src/lib/mac/cmac/cmac.cpp
```diff
@@ -25,13 +25,13 @@ void CMAC::add_data(std::span<const uint8_t> input) {
    copy_mem(m_buffer.data() + m_position, input.data(), initial_fill);
 
    if(m_position + input.size() > bs) {
-      xor_buf(m_state, m_buffer, bs);
+      xor_buf(std::span{m_state}, std::span{m_buffer}.first(bs));
```
reneme: No need to be so verbose here. `m_state` is a vector that automatically converts to a span anyway.

```suggestion
      xor_buf(m_state, std::span{m_buffer}.first(bs));
```

src/lib/mac/cmac/cmac.cpp
```diff
@@ -25,13 +25,13 @@ void CMAC::add_data(std::span<const uint8_t> input) {
...
       m_cipher->encrypt(m_state);
 
       BufferSlicer in(input);
       in.skip(bs - m_position);
       while(in.remaining() > bs) {
-         xor_buf(m_state, in.take(bs), bs);
+         xor_buf(std::span{m_state}, std::span{in.take(bs)});
```
reneme: No need to be so verbose here. `m_state` is a vector that automatically converts to a span anyway. and `BufferSlicer::take` already returns a span on its own:

```suggestion
         xor_buf(m_state, in.take(bs));
```

src/lib/mac/cmac/cmac.cpp
```diff
@@ -47,13 +47,13 @@ void CMAC::add_data(std::span<const uint8_t> input) {
 * Finalize an CMAC Calculation
 */
 void CMAC::final_result(std::span<uint8_t> mac) {
-   xor_buf(m_state, m_buffer, m_position);
+   xor_buf(std::span{m_state}.first(m_position), std::span{m_buffer}.first(m_position));
 
    if(m_position == output_length()) {
-      xor_buf(m_state, m_B, output_length());
+      xor_buf(std::span{m_state}, std::span{m_B});
```
reneme: As above:

```suggestion
      xor_buf(m_state, m_B);
```

src/lib/mac/cmac/cmac.cpp
```diff
@@ -47,13 +47,13 @@ void CMAC::add_data(std::span<const uint8_t> input) {
...
    } else {
       m_state[m_position] ^= 0x80;
-      xor_buf(m_state, m_P, output_length());
+      xor_buf(std::span{m_state}, std::span{m_P});
```
reneme: As above:

```suggestion
      xor_buf(m_state, m_P);
```

src/lib/misc/roughtime/roughtime.cpp
```diff
@@ -203,10 +202,10 @@ Response Response::from_bits(const std::vector<uint8_t>& response, const Nonce&
...
-   size_t level = 0;
-   while(level < levels) {
-      hashNode(hash, typecast_copy<std::array<uint8_t, 64>>(path.data() + level * 64), index & 1);
-      ++level;
+
+   BufferSlicer slicer(path);
+   for(size_t level = 0; level < levels; ++level) {
+      hashNode(hash, typecast_copy<std::array<uint8_t, 64>>(slicer.take<64>()), index & 1);
       index >>= 1;
    }
```
reneme: General remark (please don't tackle in this pull request): This roughtime implementation makes heavy use of `std::array<>` in places where (nowadays) it would be better off to use statically-sized `std::span`.

Here, `hashNode` is a good example: if it would be implemented using spans, you could avoid the data copies and simplify the code:

```C++
   BufferSlicer slicer(path);
   for(size_t level = 0; level < levels; ++level) {
      hashNode(hash, slicer.take<64>(), index & 1);
      index >>= 1;
   }
```

There are likely other places in this implementation that could avoid the `typecast_copy` (which is just a fancy name for an unnecessary data-copy).

src/lib/modes/aead/eax/eax.cpp
```diff
@@ -133,13 +133,13 @@ void EAX_Encryption::finish_msg(secure_vector<uint8_t>& buffer, size_t offset) {
    update(buffer, offset);
 
    secure_vector<uint8_t> data_mac = m_cmac->final();
-   xor_buf(data_mac, m_nonce_mac, data_mac.size());
+   xor_buf(std::span{data_mac}, std::span{m_nonce_mac});
```
reneme: See above:

```suggestion
   xor_buf(data_mac, m_nonce_mac);
```

src/lib/modes/aead/eax/eax.cpp
```diff
@@ -133,13 +133,13 @@ void EAX_Encryption::finish_msg(secure_vector<uint8_t>& buffer, size_t offset) {
...
 
    if(m_ad_mac.empty()) {
       m_ad_mac = eax_prf(1, block_size(), *m_cmac, nullptr, 0);
    }
 
-   xor_buf(data_mac, m_ad_mac, data_mac.size());
+   xor_buf(std::span{data_mac}, std::span{m_ad_mac});
```
reneme: See above:

```suggestion
   xor_buf(data_mac, m_ad_mac);
```

src/lib/modes/xts/xts.cpp
```diff
@@ -121,9 +121,9 @@ size_t XTS_Encryption::process_msg(uint8_t buf[], size_t sz) {
       const size_t to_proc = std::min(blocks, blocks_in_tweak);
       const size_t proc_bytes = to_proc * BS;
 
-      xor_buf(buf, tweak(), proc_bytes);
+      xor_buf(std::span(buf, proc_bytes), std::span(tweak(), proc_bytes));
       cipher().encrypt_n(buf, buf, to_proc);
-      xor_buf(buf, tweak(), proc_bytes);
+      xor_buf(std::span(buf, proc_bytes), std::span(tweak(), proc_bytes));
```
reneme: Here it would be better to rewrite the entire function to be span-aware. Currently, you mix pointer arithmetics with spans. In my opinion that makes it harder to understand the control flow.

A decent rewrite could look like this:

```C++
size_t XTS_Encryption::process_msg(uint8_t buf[], size_t sz) {
   BOTAN_STATE_CHECK(tweak_set());
   const size_t BS = cipher_block_size();

   BOTAN_ARG_CHECK(sz % BS == 0, "Input is not full blocks");
   const size_t bytes_in_tweak = tweak_blocks() * BS;

   auto buf_span = std::span(buf, sz);
   while(!buf_span.empty()) {
      const auto proc_bytes = buf_span.first(std::min(buf_span.size(), bytes_in_tweak));
      const auto proc_tweak = std::span(tweak(), proc_bytes.size());

      xor_buf(proc_bytes, proc_tweak);
      cipher().encrypt(proc_bytes);
      xor_buf(proc_bytes, proc_tweak);

      update_tweak(proc_bytes.size() / BS);
      buf_span = buf_span.subspan(proc_bytes.size());
   }

   return sz;
}
```

... eventually this will hopefully become easier to handle due to top-down moderizations like #4880 (so that `process_msg` receives a span already) and additional helper constructions like #4151.

The above applies to the remaining changes in this file as well (especially note that the changes in `finish_msg()` are going to clash hard with #4880). **I suggest we revert the changes in this file for now and tackle it once the mentioned top-down improvements are ready to clean this up properly.** That means you cannot deprecate the oldschool ptr-length-style `xor_buf` implementations for now, but I think the time is simply not right for that yet. :(

src/tests/test_roughtime.cpp
```diff
@@ -34,7 +34,9 @@ class Roughtime_Request_Tests final : public Text_Based_Test {
 
          const auto request = Botan::Roughtime::encode_request(nonce);
          result.test_eq(
-            "encode", type == "Valid", request == Botan::typecast_copy<std::array<uint8_t, 1024>>(request_v.data()));
+            "encode",
+            type == "Valid",
+            request == Botan::typecast_copy<std::array<uint8_t, 1024>>(std::span<const uint8_t>{request_v}));
```
reneme:
```suggestion
            request == Botan::typecast_copy<std::array<uint8_t, 1024>>(request_v));
```

src/tests/test_roughtime.cpp
```diff
@@ -113,11 +115,14 @@ class Roughtime final : public Test {
...
          auto rand64 = Botan::unlock(rng.random_vec(64));
          Botan::Roughtime::Nonce nonce_v(rand64);
-         result.confirm("nonce from vector",
-                        nonce_v.get_nonce() == Botan::typecast_copy<std::array<uint8_t, 64>>(rand64.data()));
-         Botan::Roughtime::Nonce nonce_a(Botan::typecast_copy<std::array<uint8_t, 64>>(rand64.data()));
-         result.confirm("nonce from array",
-                        nonce_v.get_nonce() == Botan::typecast_copy<std::array<uint8_t, 64>>(rand64.data()));
+         result.confirm(
+            "nonce from vector",
+            nonce_v.get_nonce() == Botan::typecast_copy<std::array<uint8_t, 64>>(std::span<const uint8_t>{rand64}));
```
reneme:
```suggestion
            nonce_v.get_nonce() == Botan::typecast_copy<std::array<uint8_t, 64>>(rand64));
```

src/tests/test_roughtime.cpp
```diff
@@ -113,11 +115,14 @@ class Roughtime final : public Test {
...
+         Botan::Roughtime::Nonce nonce_a(
+            Botan::typecast_copy<std::array<uint8_t, 64>>(std::span<const uint8_t>{rand64}));
```
reneme:
```suggestion
            Botan::typecast_copy<std::array<uint8_t, 64>>(rand64));
```

src/tests/test_roughtime.cpp
```diff
@@ -113,11 +115,14 @@ class Roughtime final : public Test {
...
+         result.confirm(
+            "nonce from array",
+            nonce_v.get_nonce() == Botan::typecast_copy<std::array<uint8_t, 64>>(std::span<const uint8_t>{rand64}));
```
reneme:
```suggestion
            nonce_v.get_nonce() == Botan::typecast_copy<std::array<uint8_t, 64>>(rand64));
```

src/tests/test_roughtime.cpp
```diff
@@ -135,9 +140,9 @@ class Roughtime final : public Test {
 
          auto rand64 = Botan::unlock(rng.random_vec(64));
          Botan::Roughtime::Nonce nonce_v(rand64);
-         result.confirm(
-            "empty chain nonce is blind",
-            c1.next_nonce(nonce_v).get_nonce() == Botan::typecast_copy<std::array<uint8_t, 64>>(rand64.data()));
+         result.confirm("empty chain nonce is blind",
+                        c1.next_nonce(nonce_v).get_nonce() ==
+                           Botan::typecast_copy<std::array<uint8_t, 64>>(std::span<const uint8_t>{rand64}));
```
reneme:
```suggestion
         result.confirm("empty chain nonce is blind",
                        c1.next_nonce(nonce_v).get_nonce() ==
                           Botan::typecast_copy<std::array<uint8_t, 64>>(rand64));
```

reneme (commented):
Thanks for your patience with this. I left a few suggestions for simplification (mostly unnecessary explicit construction of `std::span`).

Regarding the changes in `xts.cpp`: I think we should postpone those until some more top-down modernization has landed. Most notably #4880 and similar effort to spanify `process_msg` and `start_msg`. Once that is done, the abstract interface of all cipher mode implementations will be purely span-based and ready for such modernizations.

src/lib/misc/roughtime/roughtime.cpp
KaganCanSit: I agree that switching to std::span will eliminate unnecessary data duplication and improve code clarity.

As requested, I will not address this issue in the current PR. the use of typecast_copy was already present in the original implementation.

If you like, I can create a separate issue to track this modernization effort and work on it in a future PR when time permits. Alternatively, feel free to create an issue and assign it to me - I'd be happy to take a look at it.

src/lib/modes/xts/xts.cpp
KaganCanSit: I can see from a quick look at #4880 that there are comprehensive changes planned. Based on your suggestion, I'm reverting the changes in this file to avoid conflicts with the upcoming top-down modernizations.

I'll keep the deprecation warnings in the header file to prevent accidental additions of new ptr-length-style usage, especially since you mentioned at the beginning of this PR that `mem_ops.h` will eventually be removed.

This can be properly addressed once #4880 and #4151 are ready, allowing for a cleaner span-aware implementation.

KaganCanSit:
I have addressed all feedback from the review:

- Changes to `src/lib/modes/xts/xts.cpp` have been rolled back as suggested to avoid conflicts with upcoming modernizations (#4880, #4151)
- Retained deprecation warnings in headers to prevent new ptr-length style additions.
- The existing branch has been re-baselined to the parent branch.
- Implemented other requested fixes and improvements. (Open typing is a bit my style, sorry.)

roughtime std::span modernization proposal `BufferSlicer` is noted for future work - I may address this in a separate PR once it is merged. [Comment](https://github.com/randombit/botan/pull/4783#discussion_r2202686176)

I will follow the results of the CI tests and then we can merge if you see fit. I see you're quite busy with clang-tidy today, I'll rebase it from time to time if needed, just FYI.

reneme (approved):
Looks good to me. Thanks for your efforts.

KaganCanSit:
> Looks good to me. Thanks for your efforts.

Thank you for your time and for sharing your expertise. Your guidance has been very helpful throughout this process.

I’ll need to spend some additional time getting more familiar with the span construct. Since I'm currently use the C++11 standard, I've not using it in job.

@randombit, this development is ready to be merged at your discretion.

Best regards.

src/lib/misc/roughtime/roughtime.cpp
KaganCanSit: The improvement mentioned here was addressed as a separate PR in #4994 and included in the project.

randombit (commented):
Sorry for the wait here. Changes are generally fine but the issue is mem_ops.h is currently a public header *but* I don't think the library should be in the business of providing general utility functions like this, which is why the header is deprecated and will be removed in Botan4. Any new functionality along these lines should go into the (internal) mem_utils.h header. Ideally we can update all of the callers within the library and then guard the declarations here, which are no longer used internally but which we must continue to provide for now for compatability, with `!defined(BOTAN_IS_BEING_BUILT)` as is done with various other functions, so it's not possible to add new code to the library which uses them.

KaganCanSit:
> Sorry for the wait here. Changes are generally fine but the issue is mem_ops.h is currently a public header _but_ I don't think the library should be in the business of providing general utility functions like this, which is why the header is deprecated and will be removed in Botan4. Any new functionality along these lines should go into the (internal) mem_utils.h header. Ideally we can update all of the callers within the library and then guard the declarations here, which are no longer used internally but which we must continue to provide for now for compatability, with `!defined(BOTAN_IS_BEING_BUILT)` as is done with various other functions, so it's not possible to add new code to the library which uses them.

I recall we previously discussed removing the header file here. Actually, adding a new function for ‘mem_ops.h’ in this branch is not currently on the agenda. I was only focused on the TODO message; my goal was to deprecate the use of existing function calls and update them with std::span-based overloads.

Within the project, the use of functions marked with the declaration `BOTAN_DEPRECATED(‘This function is deprecated, use the range-based version instead’)` has now ended. These can be completely removed. However, since there may still be users who use them, I thought it would be prudent to declare them `BOTAN_DEPRECATED`. But additionally, to ensure they are not reused within the project, they can be wrapped with the `!defined(BOTAN_IS_BEING_BUILT)` macro, as you suggested.

I thought moving them to the (internal) mem_utils.h header would be a separate task. I'm not sure, maybe I misunderstood.

Some time has passed, so I need to review it again, but if you guide me step by step, I can work on it as soon as possible and apply the update you want.

reneme:
Given that this PR doesn't add any functionality to `mem_ops.h`, I would suggest the following steps :

1. Let's change the deprecation messages in `mem_ops.h` from _"This function is deprecated, use range-based version instead"_ to something along the lines of _"This function is deprecated without a replacement"_. That wording doesn't suggest that we're planning to provide support for this function post-Botan4.
2. Merge this PR (because the comment history is becoming unwieldy)
3. **In a follow-up PR:** Move the declarations and internal usages of the relevant functions over to the internal `mem_utils.h`, guard `mem_ops.h` with `!defined(BOTAN_IS_BEING_BUILT)` for backward compatibility, and eventually remove it in Botan4

KaganCanSit:
> Given that this PR doesn't add any functionality to `mem_ops.h`, I would suggest the following steps :
> 
> 1. Let's change the deprecation messages in `mem_ops.h` from _"This function is deprecated, use range-based version instead"_ to something along the lines of _"This function is deprecated without a replacement"_. That wording doesn't suggest that we're planning to provide support for this function post-Botan4.
> 2. Merge this PR (because the comment history is becoming unwieldy)
> 3. **In a follow-up PR:** Move the declarations and internal usages of the relevant functions over to the internal `mem_utils.h`, guard `mem_ops.h` with `!defined(BOTAN_IS_BEING_BUILT)` for backward compatibility, and eventually remove it in Botan4

Thank you for informing me and guiding me @reneme. I would like to take on this task in this thread and in the pull request that will be opened later. Is @randombit suitable for you?

---

**Updates:**
**_To avoid further clutter in the comments, I will update this comment._**
I am implementing the solution proposed by @reneme here. If deemed appropriate and these changes are merged, I am ready to take part in developing the remaining changes. For this, I will wait for this branch to be merged. (I don't want to repeatedly rebase a new branch based on this branch with every change.)

* 09.11.25: Rebased to main branch.
* 01.01.26: Rebased to main branch.

KaganCanSit:
Hello @randombit, @reneme,

First, I wish you a healthy, peaceful, and successful New Year. I hope to contribute more to the project in 2026.

I wanted to share a note regarding this long-standing development branch. Some of my branches have been open for a long time and for some I’ll even need to revisit the details to fully recall what I did (there are ones approaching a year). For this reason, we could follow the approach @reneme [outlined in this comment](https://github.com/randombit/botan/pull/4783#issuecomment-3346427126).

Another idea I have is: if avoiding changes to the header file (`mem_ops.h`) is important for you, then there aren’t actually any new functions added here. This PR only introduces deprecation annotations. If preferred, we could update the usages within the project to use std::span without making any changes to the header. That way, future modifications will be easier and will be future for Botan4 only change header file.

I would really like to contribute to this development. Unfortunately, I don’t have much practical experience with std::span yet, as I haven’t been using the latest standard extensively. However, I am always open to your guidance and feedback.

Kind regards.

Updates:
* 25/04/26: Rebased to main branch. Resolved `test_roughtime.cpp` merge conflict and updated `tls_record.cpp` to include `concat_util.h`, as `concat()` was moved out of `stl_util.h`.
* 24/05/26: Reabase to main branch.

PR #4785, opened 2025-03-24
Remove vestigial DHE_PSK branch

### LINKED ISSUE #4784: Always true condition in msg_client_kex
There's a likely copy/paste mistake here, with the nested `if(kex_algo == Kex_Algo::DH) {` always being true. I'd open a PR but I'm not entirely sure what the correct behaviour should be.

https://github.com/randombit/botan/blob/aa3c19853abe3bec508d66e70d66b2b6db4dc81a/src/lib/tls/tls12/msg_client_kex.cpp#L91C13-L91C21

Chaosvex:
Resolves #4784.

### DIFF
--- a/src/lib/tls/tls12/msg_client_kex.cpp
+++ b/src/lib/tls/tls12/msg_client_kex.cpp
@@ -85,10 +85,8 @@ Client_Key_Exchange::Client_Key_Exchange(Handshake_IO& io,
          }
 
          const auto private_key = state.callbacks().tls_generate_ephemeral_key(group, rng);
-         auto shared_secret = CT::strip_leading_zeros(
+         m_pre_master = CT::strip_leading_zeros(
             state.callbacks().tls_ephemeral_key_agreement(group, *private_key, peer_public_value, rng, policy));
-
-         m_pre_master = std::move(shared_secret);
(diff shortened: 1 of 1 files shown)


randombit (approved):
Thanks!

src/lib/tls/tls12/msg_client_kex.cpp
```diff
@@ -88,13 +88,7 @@ Client_Key_Exchange::Client_Key_Exchange(Handshake_IO& io,
...
             state.callbacks().tls_ephemeral_key_agreement(group, *private_key, peer_public_value, rng, policy));
 
-         if(kex_algo == Kex_Algo::DH) {
-            m_pre_master = std::move(shared_secret);
-         } else {
-            append_tls_length_value(m_pre_master, shared_secret, 2);
-            append_tls_length_value(m_pre_master, psk.bits_of(), 2);
-         }
-
+         m_pre_master = std::move(shared_secret);
```
reneme, line 91: Couldn't we just assign this onto `m_pre_master`?

Chaosvex: Good point. 😄

PR #4787, opened 2025-03-24
Parsing OIDs at compile-time via a user-defined literal

reneme:
The general idea is similar to the approach used in #4475. Basically, writing `"2.5.29.19"_oid` parses the string at compile-time and stores the OID elements into a `constexpr std::array<uint32_t>`. At runtime, the `OID` constructor simply allocates an appropriate `std::vector` and copies the elements.

Here's a godbolt to play with the core of the implementation: https://godbolt.org/z/57Pfd7GEb

This provides a significant performance boost to parsing X509 objects (around 20% faster for a CRL containing 500 entries on both my laptop and also our slowest target platform). The runtime-parsing of OIDs involves plenty of tiny allocations for the string handling and likewise many calls to `Botan::to_u32bit`. A quick benchmark using "callgrind" indicated that more than 11% of the time was spent in `to_u32bit` alone when parsing this CRL.

I'm quite sure there's more potential for optimization here. I just want to get some early feedback on the approach, especially because this might have to become part of the public API to be effective.

When parsing the CRL, `extension_from_oid()` is called repeatedly (for each `CRL_Entry`), which calls `Certificate_Extension::static_oid()` many many times. Each of those invocation re-parses these OIDs.

https://github.com/randombit/botan/blob/83d248aa8cb52bc1593384dbf830180b1b5150a7/src/lib/x509/x509_ext.cpp#L28-L90

### DIFF
--- a/src/lib/asn1/asn1_obj.h
+++ b/src/lib/asn1/asn1_obj.h
@@ -9,6 +9,7 @@
 
 #include <botan/exceptn.h>
 #include <botan/secmem.h>
+#include <array>
 #include <chrono>
 #include <iosfwd>
 #include <optional>
@@ -237,6 +238,13 @@ class BOTAN_PUBLIC_API(2, 0) OID final : public ASN1_Object {
       * Initialize an OID from a vector of integer values
       */
       explicit OID(std::vector<uint32_t>&& init);
+
+      /**
+      * Initialize an OID from an array of integer values
+      * TODO(Botan4): replace this and the c'tor above with std::span
+      */
+      template <size_t N>
+      explicit OID(const std::array<uint32_t, N>& init) : OID(std::vector<uint32_t>(init.begin(), init.end())) {}
 
       /**
       * Construct an OID from a string.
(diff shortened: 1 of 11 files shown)


randombit:
Initial reaction is this seems like a lot of complication to solve a simple problem. Why not just

```
-      static OID static_oid() { return OID("2.5.29.19"); }
+      static OID static_oid() { return OID({2,5,29,19}); }
```

reneme:
Yeah, absolutely. Just wanted to play with the user-defined literals, frankly. Though, it does make for a slightly more beautiful API, I find. ... if you squint your eyes. 😏 

Interestingly, clang seems to generate essentially the same code. GCC's initializer list code seems to be a bit more compact, though.

https://godbolt.org/z/EKv3KzMMr

randombit:
To me it seems more natural to "cheat" by taking advantage of the fact that the set we're looking up into is constant and has a lot of structure. (#4788)

And certainly (in addition to #4788) we should have `static_oid` avoid string parsing since there is no reason for it. To me using the existing `intializer_list` approach seems a lot simpler and is just as readable if not slightly more so. I'm also not too convinced that applications need such literals, so exposing it on the public API seems especially iffy.

reneme:
Closed and replaced by #4789.

PR #4788, opened 2025-03-24
Faster X.509 extension lookup

randombit:
Take advantage of the fact that the X.509 extensions we support have common OID prefixes to avoid most of the comparisons.

### DIFF
--- a/src/lib/x509/x509_ext.cpp
+++ b/src/lib/x509/x509_ext.cpp
@@ -25,65 +25,67 @@ namespace Botan {
 
 namespace {
 
+std::optional<uint32_t> is_sub_element_of(const OID& oid, std::initializer_list<uint32_t> prefix) {
+   const auto& c = oid.get_components();
+
+   if(c.size() != prefix.size() + 1) {
+      return {};
+   }
+
+   if(!std::equal(c.begin(), c.end() - 1, prefix.begin(), prefix.end())) {
+      return {};
+   }
+
+   return c[c.size() - 1];
+}
+
+template <std::derived_from<Certificate_Extension> T>
+auto make_extension([[maybe_unused]] const OID& oid) {
+   BOTAN_DEBUG_ASSERT(oid == T::static_oid());
+   return std::make_unique<T>();
+}
+
 std::unique_ptr<Certificate_Extension> extension_from_oid(const OID& oid) {
-   if(oid == Cert_Extension::Subject_Key_ID::static_oid()) {
-      return std::make_unique<Cert_Extension::Subject_Key_ID>();
+   if(auto iso_ext = is_sub_element_of(oid, {2, 5, 29})) {
+      // NOLINTNEXTLINE(*-switch-missing-default-case)
+      switch(*iso_ext) {
+         case 14:
+            return make_extension<Cert_Extension::Subject_Key_ID>(oid);
+         case 15:
+            return make_extension<Cert_Extension::Key_Usage>(oid);
+         case 17:
+            return make_extension<Cert_Extension::Subject_Alternative_Name>(oid);
+         case 18:
+            return make_extension<Cert_Extension::Issuer_Alternative_Name>(oid);
+         case 19:
+            return make_extension<Cert_Extension::Basic_Constraints>(oid);
+         case 20:
+            return make_extension<Cert_Extension::CRL_Number>(oid);
+         case 21:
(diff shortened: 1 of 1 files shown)


src/lib/x509/x509_ext.cpp
```diff
@@ -25,65 +25,98 @@ namespace Botan {
...
-      return std::make_unique<Cert_Extension::Subject_Alternative_Name>();
+   if(!std::equal(c.begin(), c.end() - 1, prefix.begin(), prefix.end())) {
+      return {};
    }
 
-   if(oid == Cert_Extension::Issuer_Alternative_Name::static_oid()) {
-      return std::make_unique<Cert_Extension::Issuer_Alternative_Name>();
-   }
+   return c[c.size() - 1];
+}
```
reneme, line 40: Perhaps it makes sense to have this as a member in `OID`?

src/lib/x509/x509_ext.cpp
```diff
@@ -25,65 +25,98 @@ namespace Botan {
...
 
-   if(oid == Cert_Extension::Basic_Constraints::static_oid()) {
-      return std::make_unique<Cert_Extension::Basic_Constraints>();
-   }
+std::unique_ptr<Certificate_Extension> extension_from_oid(const OID& oid) {
```
reneme, line 42: Perhaps a bit more compact with less potential for copy-paste issues in the future:

```C++

template <std::derived_from<Certificate_Extension> T>
auto make_extension([[maybe_unused]] const OID& oid) {
   BOTAN_DEBUG_ASSERT(oid == T::static_oid());
   return std::make_unique<T>();
}

std::unique_ptr<Certificate_Extension> extension_from_oid(const OID& oid) {
   if(auto iso_ext = is_sub_element_of(oid, {2, 5, 29})) {
      // NOLINTNEXTLINE(*-switch-missing-default-case)
      switch(*iso_ext) {
         case 14:
            return make_extension<Cert_Extension::Subject_Key_ID>(oid);
         case 15:
            return make_extension<Cert_Extension::Key_Usage>(oid);
         case 17:
            return make_extension<Cert_Extension::Subject_Alternative_Name>(oid);
         case 18:
            return make_extension<Cert_Extension::Issuer_Alternative_Name>(oid);
         case 19:
            return make_extension<Cert_Extension::Basic_Constraints>(oid);
         case 20:
            return make_extension<Cert_Extension::CRL_Number>(oid);
         case 21:
            return make_extension<Cert_Extension::CRL_ReasonCode>(oid);
         case 28:
            return make_extension<Cert_Extension::CRL_Issuing_Distribution_Point>(oid);
         case 30:
            return make_extension<Cert_Extension::Name_Constraints>(oid);
         case 31:
            return make_extension<Cert_Extension::CRL_Distribution_Points>(oid);
         case 32:
            return make_extension<Cert_Extension::Certificate_Policies>(oid);
         case 35:
            return make_extension<Cert_Extension::Authority_Key_ID>(oid);
         case 37:
            return make_extension<Cert_Extension::Extended_Key_Usage>(oid);
      }
   }

   if(auto pkix_ext = is_sub_element_of(oid, {1, 3, 6, 1, 5, 5, 7, 1})) {
      // NOLINTNEXTLINE(*-switch-missing-default-case)
      switch(*pkix_ext) {
         case 1:
            return make_extension<Cert_Extension::Authority_Information_Access>(oid);
         case 26:
            return make_extension<Cert_Extension::TNAuthList>(oid);
      }
   }

   return nullptr;  // unknown
}
```

reneme (approved):
Thanks for looking into this! Together with #4789, this makes for 35% speedup in the CRL parsing benchmark I suggested in #4790. 🚀

I see the optimization potential here. Though, the fact that we're now duplicating the knowledge of the OID assignments bugs me a little. Anyway, my guess is that that's exactly what's the `BOTAN_DEBUG_ASSERT` is meant to mitigate.

I'd suggest a bit of a cleanup using `switch` statements.

src/lib/x509/x509_ext.cpp
randombit: Possibly. I considered it, but for now this is the only place it is needed. If we end up doing a similar optimization elsewhere, that to me would be the time the promote this outside of an anon namespace [*]

[*] Though I'm not sure if I would put this as a member on `OID`, or just in an internal header somewhere. This seems like an odd thing for an application to need.

reneme (approved):

PR #4790, opened 2025-03-25
Add a benchmark for parsing X509 certs and CRLs

reneme:
This is the benchmark I used for the profiling mentioned in #4789. Perhaps it is useful for future performance tweaks in that direction.


randombit (approved):

randombit:
Actually one change req here - move to a new `perf_x509.cpp` so we have a place to put other similar benchmarks in the future

PR #4793, opened 2025-03-26
Add benchmark for public and private key parsing

randombit:
Plus various optimizations driven by this

### DIFF
--- a/src/lib/math/bigint/divide.cpp
+++ b/src/lib/math/bigint/divide.cpp
@@ -27,7 +27,7 @@ void sign_fixup(const BigInt& x, const BigInt& y, BigInt& q, BigInt& r) {
    }
 }
 
-inline bool division_check(word q, word y2, word y1, word x3, word x2, word x1) {
+inline bool division_check_vartime(word q, word y2, word y1, word x3, word x2, word x1) {
    /*
    Compute (y3,y2,y1) = (y2,y1) * q
    and return true if (y3,y2,y1) > (x3,x2,x1)
@@ -37,10 +37,13 @@ inline bool division_check(word q, word y2, word y1, word x3, word x2, word x1) 
    y1 = word_madd2(q, y1, &y3);
    y2 = word_madd2(q, y2, &y3);
 
-   const word x[3] = {x1, x2, x3};
-   const word y[3] = {y1, y2, y3};
-
-   return bigint_ct_is_lt(x, 3, y, 3).as_bool();
+   if(x3 != y3) {
+      return (y3 > x3);
+   }
+   if(x2 != y2) {
+      return (y2 > x2);
+   }
+   return (y1 > x1);
 }
 
 }  // namespace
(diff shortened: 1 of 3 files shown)


src/lib/math/bigint/divide.cpp
```diff
@@ -272,27 +277,30 @@ void vartime_divide(const BigInt& x, const BigInt& y_arg, BigInt& q_out, BigInt&
...
-      qjt = CT::Mask<word>::is_equal(x_j0, y_t0).select(WordInfo<word>::max, qjt);
+      word qjt = (x_j0 == y_t0) ? WordInfo<word>::max : bigint_divop_vartime(x_j0, x_j1, y_t0);
 
       // Per HAC 14.23, this operation is required at most twice
-      qjt -= division_check(qjt, y_t0, y_t1, x_j0, x_j1, x_j2);
-      qjt -= division_check(qjt, y_t0, y_t1, x_j0, x_j1, x_j2);
-      BOTAN_DEBUG_ASSERT(division_check(qjt, y_t0, y_t1, x_j0, x_j1, x_j2) == false);
+      while(division_check_vartime(qjt, y_t0, y_t1, x_j0, x_j1, x_j2)) {
+         qjt--;
+      }
```
reneme, line 285: Should we guard this loop with an upper bound then, if it is meant to run at most twice? 🤔

reneme (approved):

src/lib/math/bigint/divide.cpp
randombit: The code is a bit cleaner without the upper bound but I guess it might prevent a miscompilation elsewhere from turning the division into an ~infinite loop.

src/lib/math/bigint/divide.cpp
```diff
@@ -272,27 +277,34 @@ void vartime_divide(const BigInt& x, const BigInt& y_arg, BigInt& q_out, BigInt&
...
-      qjt -= division_check(qjt, y_t0, y_t1, x_j0, x_j1, x_j2);
-      qjt -= division_check(qjt, y_t0, y_t1, x_j0, x_j1, x_j2);
-      BOTAN_DEBUG_ASSERT(division_check(qjt, y_t0, y_t1, x_j0, x_j1, x_j2) == false);
+      for(size_t k = 0; k != 2; ++k) {
+         if(division_check_vartime(qjt, y_t0, y_t1, x_j0, x_j1, x_j2)) {
+            qjt--;
+         } else {
+            break;
+         }
+      }
```
reneme, line 289: > The code is a bit cleaner without the upper bound

# 😁

```suggestion
      for(size_t k = 0; k != 2 && ! division_check_vartime(qjt, y_t0, y_t1, x_j0, x_j1, x_j2); ++k, --gjt);
      BOTAN_DEBUG_ASSERT(!division_check(qjt, y_t0, y_t1, x_j0, x_j1, x_j2));
```

... I'm not sure. 😅

randombit: I am sure that is fucking hideous :)

PR #4794, opened 2025-03-26
Add missing compliance checks in EC(DH)

reneme:
As discussed out-of-band, this adds additional validation to the ECC key handling as well as ECDH key agreement. With those, it is not possible to create an EC private key using a zero-scalar, nor to create an EC public key with the point at infinity. Additionally, trying to perform an ECDH agreement with the point-at-infinity as the peer's public value is now also caught and rejected.

### DIFF
--- a/src/lib/pubkey/ecdh/ecdh.cpp
+++ b/src/lib/pubkey/ecdh/ecdh.cpp
@@ -34,20 +34,33 @@ class ECDH_KA_Operation final : public PK_Ops::Key_Agreement_with_KDF {
       size_t agreed_value_size() const override { return m_group.get_p_bytes(); }
 
       secure_vector<uint8_t> raw_agree(const uint8_t w[], size_t w_len) override {
-         if(m_group.has_cofactor()) {
+         const auto input_point = [&] {
+            if(m_group.has_cofactor()) {
 #if defined(BOTAN_HAS_LEGACY_EC_POINT)
-            EC_AffinePoint input_point(m_group, m_group.get_cofactor() * m_group.OS2ECP(w, w_len));
-            return input_point.mul_x_only(m_l_times_priv, m_rng);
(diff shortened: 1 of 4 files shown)


src/lib/pubkey/ecdh/ecdh.cpp
```diff
@@ -34,20 +34,33 @@ class ECDH_KA_Operation final : public PK_Ops::Key_Agreement_with_KDF {
...
+         }();
+
+         // Typical specs (such as BSI's TR-03111 Section 4.3.1) require that
+         // we check the resulting point of the multiplication to not be the
+         // point at infinity. However, since we ensure that our ECC private
+         // scalar can never be zero, checking the peer's input point is
+         // equivalent.
+         if(input_point.is_identity()) {
+            throw Decoding_Error("ECDH - Invalid elliptic curve point: identity");
          }
```
FAlbertDev, line 61: I don't want to be nitpicky, but there are peer's input points leading to zero, even if the scalar is non-zero. With Euler's theorem we should get $s_A * (s_B * P) = 0$ if $s_A * s_B = 0 \pmod{order(Curve)}$. I guess forging such $s_B$ should be cryptographically impossible if you do not know $s_A$. 
But why don't we simply check the resulting x-coordinate for being all zero bytes?

randombit: There are valid ECDH exchanges where the x coordinate of the resulting point is zero, I don't see why we should forbid them

FAlbertDev: Sorry, I messed up. Since the group order is prime, $s_A * s_B = 0 \pmod{order}$ iff either $s_A = 0 \pmod{order}$ or $s_B = 0 \pmod{order}$. Therefore, the current check should be okay.

FAlbertDev (approved):

randombit (approved):
LGTM thanks

PR #4796, opened 2025-03-27
Don't create throwaway strings in bcrypt_base64_encode

Chaosvex:
Just a small improvement to avoid allocating strings within the loop (even though it's not going to go through many iterations) when the compiler can reuse the existing string and produce much simpler code, sans allocations. `.back()` and `b64[b64.size() - 1]` will produce identical optimised code, so it doesn't really matter but it's slightly cleaner.

https://quick-bench.com/q/TN1XXPkL3uQ6Ozvd6RVoCLUw_rM


randombit (approved):
Thanks. I'm not sure why it was done that way.

PR #4800, opened 2025-03-28
Add support for RISCV64 CPU feature detection on Linux

randombit:


src/lib/utils/cpuid/cpuid_riscv64/cpuid_riscv64.cpp
```diff
@@ -0,0 +1,83 @@
...
+namespace {
+
+constexpr uint64_t bitflag(uint64_t b) {
+   return (static_cast<uint64_t>(1) << b);
+}
+
+template <typename... Bs>
+constexpr uint64_t bitflag(uint64_t b, Bs... rest) {
+   return bitflag(b) | bitflag(rest...);
+}
```
reneme, line 29:
```suggestion
template <std::convertible_to<uint64_t>... Bs>
constexpr uint64_t bitflag(Bs... bs) {
   return ((uint64_t(1) << bs) | ... | 0);
}
```

Fold expressions ftw. https://godbolt.org/z/T695nz7Wb

reneme: On a second thought, `bitflag()` without parameters seems silly:

```C++
template <std::convertible_to<uint64_t>... Bs>
    requires(sizeof...(Bs) > 0)
constexpr uint64_t bitflag(Bs... bs) {
   return ((uint64_t(1) << bs) | ...);
}
```

PR #4804, opened 2025-03-29
Run the ARMv7 build on Aarch64 rather than via qemu

randombit:


randombit:
It seems like the GHA Aarch64 machines are not set up to run 32-bit ARM binaries :(

PR #4805, opened 2025-03-29
Try re-enabling modules which MSVC in the past miscompiled

randombit:
Maybe they fixed the bugs!

(But probably not)


randombit:
`argon2_ssse3` is still broken on 32-bit MSVC

`serpent_avx2` seems to work now.

`serpent_avx512` is not testable but neither can I find any ticket or reason it was originally disabled beyond a general statement that it doesn't work (a76f7ea4d14470a0b651c3cd6e9f8158f2e7499d). yolo

PR #4807, opened 2025-03-30
Allow specifying AR command

### LINKED ISSUE #4806: Allow overriding ar command in ci_build.py
# Problem
Botan can not be cross compiled for Windows (x86-64) from Linux (x86-64) using MXE. This is because the ar command is hardcoded for that target.

# Solution
Allow users to specify the ar command.

# Implementation
I've implemented this using the AR variable, which is the same way Gnu Make handles this.

# Steps to reproduce

Tested on Debian 12 (bookworm).

1. Build MXE & update your PATH (see below)
2. Get the dependencies (see further below)
3. `AR=x86_64-w64-mingw32.static-ar CXX=x86_64-w64-mingw32.static-g++ python3 ./src/scripts/ci_build.py --os=windows --cc=gcc cross-win64`

Fails before the associated patch, succeeds after the patch.

## Building MXE

Takes about an hour on my computer.

```sh
sudo apt-get install -y autoconf automake autopoint bash bison bzip2 flex g++ g++-multilib gettext git gperf intltool libc6-dev-i386 libgdk-pixbuf2.0-dev libltdl-dev libgl-dev libpcre3-dev libssl-dev libtool-bin libxml-parser-perl lzip make openssl p7zip-full patch perl python3 python3-distutils python3-mako python3-packaging python3-pkg-resources python3-setuptools python-is-python3 ruby sed sqlite3 unzip wget xz-utils

git clone https://github.com/mxe/mxe.git
cd mxe
make MXE_TARGETS='x86_64-w64-mingw32.static i686-w64-mingw32.static' cc
export PATH="`pwd`/usr/bin:$PATH"
cd ..
```

Note 1: The list of apt-packages is not minimized, so there might be some packages in there that are not strictly required.

Note 2: There is an MXE apt repo, but it has an old version of gcc, which is why we have to build it from source. I plan on trying to coordinate with the MXE maintainers to see if I can get a new version built and put up there.

## Install dependencies

```sh
arch="x86-64"
sudo apt-get update
sudo apt-get install -y software-properties-common lsb-release gnupg
sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 86B72ED9
sudo add-apt-repository -y "deb [arch=amd64] https://pkg.mxe.cc/repos/apt buster main"
sudo apt-get update
sudo apt install -y mingw-w64-common mxe-$arch-w64-mingw32.static-qtwebsockets mxe-$arch-w64-mingw32.static-libwebsockets mxe-$arch-w64-mingw32.static-libgcrypt mxe-$arch-w64-mingw32.static-qtbase mxe-$arch-w64-mingw32.static-qt mxe-$arch-w64-mingw32.static-libusb1 libargon2-dev libbotan-2-dev libykpers-1-dev libpcsclite-dev cmake libqt5svg5-dev qttools5-dev libminizip-dev libkeyutils-dev libqt5x11extras5-dev libxtst-dev libqrencode-dev asciidoctor wine
```

Note: The list of apt-packages is not minimized, so there might be some packages in there that are not strictly required.

hax0rbana-adam:
Closes #4806

If we allow the ar command to be overridden for the cross-win64 target instead of being hard-coded, it will be possible to use MXE to do the compilation.

I chose the AR environment variable because that matches [what Gnu Make uses](https://www.gnu.org/software/make/manual/make.html#index-CFLAGS).

If AR is not set, the script falls back to the hard-coded value, making this change backward compatible.

### DIFF
--- a/src/scripts/ci_build.py
+++ b/src/scripts/ci_build.py
@@ -334,8 +334,11 @@ def determine_flags(target, target_os, target_cpu, target_cc, cc_bin, ccache,
         elif target == 'cross-win64':
             # MinGW in 16.04 is lacking std::mutex for unknown reason
             cc_bin = 'x86_64-w64-mingw32-g++'
-            flags += ['--cpu=x86_64', '--cc-abi-flags=-static',
-                      '--ar-command=x86_64-w64-mingw32-ar', '--without-os-feature=threads']
+            flags += ['--cpu=x86_64', '--cc-abi-flags=-static', '--without-os-feature=threads']
+            if os.getenv('AR') is None:
+                flags += [ '--ar-command=x86_64-w64-mingw32-ar' ]
+            else:
+                flags += [ '--ar-command=%s' % os.getenv('AR') ]
             test_cmd = [os.path.join(root_dir, 'botan-test.exe')] + test_cmd[1:]
             test_prefix = ['wine']
         else:


randombit:
We don't need this in CI currently and this script is not intended to be run by end users

PR #4811, opened 2025-03-31
Fixed the use of C++20 functions with updated Android NDK and XCode versions.

KaganCanSit:
Hello,

While reviewing the project code, I found improvements to the functions that could be made if C++20 support for Android NDK and XCode is completed.

**Android NDK Status:**
When I checked the notifications about Android NDK on GitHub, I noticed that there was an issue reported and it was fixed by adding updated support with a new release. [Related issue report](https://github.com/android/ndk/issues/1530) Accordingly, I noticed that the project is already using the r26 version.
> CI now uses Android NDK 26, and earlier NDKs are not supported
> ./src/configs/repo_config.env:ANDROID_NDK="android-ndk-r26”

**Xcode Status:**
I did some research on Xcode's support for C++20 and later standards. According to Apple's official documentation, the latest versions of Xcode support many features of the C++20 and C++23 standards. However, I couldn't find any clear information on whether is specifically functions supported or not. For more precise information, it would be useful to test the availability of this function in the Xcode version your project is targeting.

https://developer.apple.com/xcode/cpp/

Based on the above information, I thought there is no obstacle to perform the updates. The builds with Ninja are successful. However, I'm not sure if the tests performed under pull request will be sufficient since we don't have any precise information on the Xcode side. I would be interested if you can guide me for additional tests.

### DIFF
--- a/src/lib/tls/tls_session_manager.cpp
+++ b/src/lib/tls/tls_session_manager.cpp
@@ -107,54 +107,48 @@ std::vector<Session_with_Handle> Session_Manager::find_and_filter(const Server_I
          break;
       }
 
-      // TODO: C++20, use std::ranges::remove_if() once XCode and Android NDK caught up.
-      sessions_and_handles.erase(
-         std::remove_if(sessions_and_handles.begin(),
-                        sessions_and_handles.end(),
-                        [&](const auto& session) {
-                           const auto age =
-                              std::chrono::duration_cast<std::chrono::seconds>(now - session.session.start_time());
+      std::erase_if(sessions_and_handles, [&](const auto& session) {
+         const auto age = std::chrono::duration_cast<std::chrono::seconds>(now - session.session.start_time());
 
-                           // RFC 5077 3.3 -- "Old Session Tickets"
-                           //    The ticket_lifetime_hint field contains a hint from the
-                           //    server about how long the ticket should be stored. [...]
-                           //    A client SHOULD delete the ticket and associated state when
-                           //    the time expires. It MAY delete the ticket earlier based on
-                           //    local policy.
-                           //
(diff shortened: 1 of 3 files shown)


KaganCanSit:
For versions below macOS (XCode) 14 and for ARM Android, I see that the friend notifications fail. I will undo the changes in the 'pqcrystals.h' file for now and review it again.

Failed jobs;
> ci/Cross (cross-android-arm64, clang, ubuntu-24.04) (pull_request ) Failing after 1m
> ci/Cross (cross-ios-arm64, xcode, macos-13) (pull_request) Failing after 1m
> ci/macOS (amalgamation, xcode, macos-13) (pull_request) Failing after 1m
> ci / macOS (shared, xcode, macos-13) (pull_request) Failing after 2m

> build/include/internal/botan/internal/pqcrystals.h:162:11: error: calling a private constructor of class 'Botan::CRYSTALS::PolynomialVector<Botan::DilithiumPolyTraits, Botan::CRYSTALS::Domain::NTT>'
> return StructureT<Trait, To>(std::move(p));

I also need to make sure that other changes are processed correctly here. Testing is more important now that the current situation has been examined.

randombit:
Thanks! Regarding XCode versions, currently our support for XCode is limited to the latest published version. (Though in practice that's XCode 15, since that's the latest version supported on x86 versions of macOS). The situation is similar for the NDK.

 So if the PR compiles on the relevant CI builders, it is from a project standpoint good to go.

randombit:
This looks good to me but @reneme all three of these are TLS related and you've touched that much more than I in the last few years, so I'd like your input.

reneme (approved):
This looks good to me, too. I agree with you that the update to NDK r26 probably removed the need for this workaround. Thanks for applying it.

PR #4813, opened 2025-04-01
Update jitterentropy and work around a build issue in CI

reneme:
CMake [recently started refusing](https://github.com/randombit/botan/actions/runs/14186542044/job/39742803448) to configure this project as it claims support for old 2.x versions of CMake. Previously, this just produced a warning, but I guess GH Actions updated CMake on their runners.

See also: https://github.com/smuellerDD/jitterentropy-library/issues/147

### DIFF
--- a/src/configs/repo_config.env
+++ b/src/configs/repo_config.env
@@ -22,7 +22,7 @@ BORINGSSL_BRANCH="rene/runner-20241016"
 ANDROID_NDK="android-ndk-r26"
 
 # Jitterentropy library version to be used for testing the 'jitter_rng' module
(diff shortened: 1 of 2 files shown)


randombit (approved):
I cannot even begin to grasp the thought process behind this behavior from CMake

reneme:
> I cannot even begin to grasp the thought process behind this behavior from CMake

I was surprised as well. My best guess is that the semantics of the "minimal supported CMake version" is somehow more involved. Like: if some script claims compatibility with an ancient CMake version, they fall back to legacy behavior somehow. It certainly sounds messy.

PR #4814, opened 2025-04-01
Replace TLS1.2 raw pointers with unique ptr

KaganCanSit:
Hello,

While researching to contribute to the project, I noticed the following comment line for TLS 1.2. Accordingly, I have updated the raw pointer usage in the required files for TLS 1.2 by replacing it with unique_ptr. I think the getter methods should not be updated as they are for view access only. I don't have a good knowledge of TLS and the changes here may have side effects, even if they seem minor.

Initial testing with `./botan-test` and `python3 src/scripts/test_cli.py ./botan cli_tls_socket_tests` completed successfully, but I'd appreciate guidance on any additional testing that would be helpful.

> TODO: take unique_ptr instead of of raw pointers for all of these, as we're taking the ownership

**Important Note:** The existing null pointer control structure has been left untouched to limit the scope of the change and to facilitate review. However, I think it would be appropriate to add consistent null pointer checks in all functions in the future. Yes, such a usage is very unlikely, but I think a little more control would be good. **Currently some functions (like `client_hello`) do null checks, others do not. What is your opinion about this situation?**

### DIFF
--- a/src/lib/tls/tls12/tls_client_impl_12.cpp
+++ b/src/lib/tls/tls12/tls_client_impl_12.cpp
@@ -118,7 +118,7 @@ Client_Impl_12::Client_Impl_12(const Channel_Impl::Downgrade_Information& downgr
          downgrade_info.client_hello_message.begin() + 4 /* handshake header length */,
          downgrade_info.client_hello_message.end());
 
-      state.client_hello(new Client_Hello_12(client_hello_msg));
+      state.client_hello(std::make_unique<Client_Hello_12>(client_hello_msg));
       state.hash().update(downgrade_info.client_hello_message);
 
       secure_renegotiation_check(state.client_hello());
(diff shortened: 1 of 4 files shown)


KaganCanSit:
I came across the situation [here. - Update jitterentropy and work around a build issue in CI]( https://github.com/randombit/botan/pull/4813)

After the merge I will restart the failed `ci / Clang Tidy (pull_request)` job.

KaganCanSit:
Hello Mr. René Meusel (@reneme) and Mr. Jack Lloyd (@randombit),

I currently have four open merge requests and wanted to let you know that I have plenty of availability today and tomorrow.
If you're able to review any of them during that time, I'd be happy to assist or respond to any feedback promptly.

Just wanted to keep you informed—thanks in advance!

randombit (approved):

PR #4816, opened 2025-04-02
Add OID related methods to FFI

arckoor:
Adds various OID related methods, like parsing to and from strings, comparisons and the like.

### DIFF
--- a/src/lib/ffi/ffi_oid.cpp
+++ b/src/lib/ffi/ffi_oid.cpp
@@ -1,0 +1,82 @@
+/*
+* (C) 2025 Jack Lloyd
+* (C) 2025 Dominik Schricker
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/ffi.h>
+
+#include <botan/pk_keys.h>
+#include <botan/internal/ffi_oid.h>
+#include <botan/internal/ffi_pkey.h>
+#include <botan/internal/ffi_util.h>
+
+extern "C" {
+
+using namespace Botan_FFI;
+
+int botan_oid_destroy(botan_asn1_oid_t oid) {
+   return BOTAN_FFI_CHECKED_DELETE(oid);
+}
+
+int botan_oid_from_string(botan_asn1_oid_t* oid_obj, const char* oid_str) {
+   return ffi_guard_thunk(__func__, [=]() -> int {
+      if(oid_obj == nullptr || oid_str == nullptr) {
+         return BOTAN_FFI_ERROR_NULL_POINTER;
+      }
+      Botan::OID oid;
+      // This returns a Lookup_Error if an unknown name is passed,
+      // which would get turned into NOT_IMPLEMENTED
+      try {
+         oid = Botan::OID::from_string(oid_str);
+      } catch(Botan::Lookup_Error&) {
+         return BOTAN_FFI_ERROR_BAD_PARAMETER;
+      }
+      std::unique_ptr<Botan::OID> oid_ptr = std::make_unique<Botan::OID>(std::move(oid));
+      *oid_obj = new botan_asn1_oid_struct(std::move(oid_ptr));
+      return BOTAN_FFI_SUCCESS;
+   });
+}
+
+int botan_oid_register(botan_asn1_oid_t oid, const char* name) {
+   return BOTAN_FFI_VISIT(oid, [=](const auto& o) -> int {
+      if(name == nullptr) {
+         return BOTAN_FFI_ERROR_NULL_POINTER;
+      }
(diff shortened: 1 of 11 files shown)


src/lib/ffi/ffi_oid.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// todo string view thingy
+// also todo parse RSA and other such names
+
+int botan_oid_equal(botan_asn1_oid_t a_w, botan_asn1_oid_t b_w) {
+   return BOTAN_FFI_VISIT(a_w, [=](const auto& a) -> int { return a == safe_get(b_w); });
+}
+
+int botan_oid_cmp(int* result, botan_asn1_oid_t a_w, botan_asn1_oid_t b_w) {
```
arckoor, line 44: The implementation for this is basically stolen from `botan_mp_cmp`, but one thing I did notice is that `* result` isn't checked anywhere, not in the macro, nor within the function, i.e. this can segfault. Is this intended? There are a few other functions that also behave like this, e.g. `botan_mp_num_bytes`

src/lib/ffi/ffi_oid.h
```diff
@@ -0,0 +1,20 @@
...
+#ifndef BOTAN_FFI_OID_H_
+#define BOTAN_FFI_OID_H_
+
+#include <botan/asn1_obj.h>
+#include <botan/internal/ffi_util.h>
+
+extern "C" {
+
+// TODO get a proper magic number from somewhere:tm:
+BOTAN_FFI_DECLARE_STRUCT(botan_asn1_oid_struct, Botan::OID, 0x123456);
```
arckoor, line 17: How should I create / compute the magic number here?

randombit: Any random 32-bit integer (eg `botan rng --format=hex 4`). Main point is that it does not collide with any other FFI magic number, so someone accidentally using the wrong type gets a reasonable error rather than a memory corruption. Secondary point is tht it isn't likely otherwise to collide with memory in a random process stack (eg all 0s is valid 32-bit integer, doesn't collide with other uses, but can easily collide with something randomly on a heap)

src/lib/ffi/ffi_oid.cpp
randombit: `result` should be checked to not be null with the appropriate error code returned if it is. The general intention is, if a FFI function is called with a pointer value that is null when it should not be, we return the dedicated null pointer error code - any function that's missing such a check has a bug.

This is a lot stricter than the C++ API; there we might at times detect/reject a null pointer but it's not assured. FFI is in a weird place because it's commonly called from languages with effectively much weaker type safety than C++, so it goes to some efforts to make errors like this non-fatal.

arckoor: Alright, then I'll fix it for this PR's methods only for now, and in a separate PR I can maybe go through `ffi.h` and fix the rest.

src/python/botan3.py
```diff
@@ -2123,6 +2143,59 @@ def clear_bit(self, bit):
     def set_bit(self, bit):
         _DLL.botan_mp_set_bit(self.__obj, c_size_t(bit))
 
+
+class OID:
+    def __init__(self, obj=None):
+        if not obj:
+            obj = c_void_p(0)
+        self.__obj = obj
```
arckoor, line 2151: I intended to use the same `def __init__(self, obj=c_void_p(0)):` that `PublicKey` and `PrivateKey` use, but as it turns out python does _not_ create a new `c_void_p` everytime this is called, but instead creates it only once and then reuses it, leading to every instance of `OID` writing to the same memory. I would consider this unintended behavior, and thus solved it like so. `PrivateKey` and `PublicKey` get around this by meticulously creating new `c_void_p`s, and then passing that to their constructor. Should I adapt these to in the OID way, or leave them as is?

arckoor:
I have frankly no idea why `Clang Tidy` is failing :/

arckoor:
I snuck in the missing documentation for `stateful_operations` in ffi.rst and fixed some syntax errors there while I was at it

src/python/botan3.py
randombit: Yes I ran into exactly this issue at some point and was quite surprised! Using `=None` as here is probably the way to go.

randombit (approved):
LGTM @arckoor thanks! The changes to the Python wrappers re `c_void_p` are optional (but would be appreciated for sure)

randombit:
`clang-tidy` failure is unrelated, I will address in another PR

arckoor:
@randombit I've modified PubKey and PrivKey to correctly init the `c_void_p`, can be merged from my side

PR #4826, opened 2025-04-12
Avoid some heap usage in GHASH/GMAC/GCM

randombit:


reneme (approved):
I remember considering that back when I refactored GHASH the other day. I refrained because of the 'secure allocator' aspect. Should we perhaps somehow document the circumstances when stack allocation is acceptable?

randombit:
> Should we perhaps somehow document the circumstances when stack allocation is acceptable?

Yes. I'll write something up. But generally I'd say if the object is live for a short time (as here) it's fine to use the stack and not bother zeroing it because very likely there are other intermediate data pushed onto the stack in the course of computation that would leak key material, and we don't have any way of dealing with that.

PR #4827, opened 2025-04-12
Remove autogen of x509_dn_ub.cpp

randombit:
These constants should change effectively never since they are baked into RFC 5280, and there are likely to be few new standardized DN entries going forward; everything is in the subject alternative name now.

Take advantage of the same fast prefix check as used for X509 extensions.

This changes the accepted upper limits for generational qualifier, given name, and initials. Previously a comment was given that these types inherit the (uninteresting) bound for 32K chars from X520. But RFC 5280 Appendix A clearly gives much lower (and more sensible) bounds for all three of these elements.


randombit:
ping @reneme wasn't sure if you cared about this specially the limit reductions - but 5280 seems pretty clear on this

reneme:
@atreiber94 could you take this, please? (I'm traveling with limited time/access to internet)

atreiber94 (approved):
Looks fine for me and we do not really care about the reduction of the bounds for these three attributes. (It seems like OpenSSL does not really check for these bounds but since the implementation will be conform to the RFC I think it is fine)

PR #4832, opened 2025-04-18
Replace std::vector with std::array in codec_base to avoid allocations

Chaosvex:
Modernised the base32/64 code slightly so the vector buffers can be replaced by arrays, removing the need for performing any allocations for conversions.

### DIFF
--- a/src/lib/utils/codec_base.h
+++ b/src/lib/utils/codec_base.h
@@ -11,8 +11,8 @@
 
 #include <botan/exceptn.h>
 #include <botan/mem_ops.h>
+#include <array>
 #include <string>
-#include <vector>
 
 namespace Botan {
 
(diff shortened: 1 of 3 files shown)


Chaosvex:
The initial patch didn't work with older versions of Clang when compiling as C++20 thanks to a language defect and an arguable defect in Clang that haven't applied the fix (P2280R4) retroactively, so this seems to be the only way to get it to work across all supported compilers in C++20 without making more intrusive changes that would arguably make the code less flexible. It's of course slight less flexible now that it requires the interface to provide constexpr sizes but I don't think that's really a problem here.

src/lib/utils/codec_base.h
```diff
@@ -35,8 +36,8 @@ size_t base_encode(
    Base&& base, char output[], const uint8_t input[], size_t input_length, size_t& input_consumed, bool final_inputs) {
    input_consumed = 0;
 
-   const size_t encoding_bytes_in = base.encoding_bytes_in();
-   const size_t encoding_bytes_out = base.encoding_bytes_out();
+   constexpr size_t encoding_bytes_in = std::remove_reference_t<Base>::encoding_bytes_in();
```
randombit, line 39: Could you here (and also below) add a comment along the lines of 

```
// TODO(Botan4) Check if we can use just base. or Base:: here instead
```

randombit (commented):
Nice, thank you! Can you please squash? Beyond that and adding a reminder comment to hopefully remove that ugly workaround in the future, LGTM

src/lib/utils/codec_base.h
Chaosvex: Done.

randombit (approved):

PR #4833, opened 2025-04-21
Add EC_Group related methods to FFI

arckoor:
Adds the stable part of EC_Group (so not EC_AffinePoint) to the FFI to allow creating EC Keys from application specific curves.

### DIFF
--- a/doc/api_ref/ffi.rst
+++ b/doc/api_ref/ffi.rst
@@ -874,6 +874,100 @@ Object Identifiers
    Return 1 if ``a`` is equal to ``b``, 0 if ``a`` is not equal to ``b``
 
 
+EC Groups
+----------------------------------------
+
+.. versionadded:: 3.8.0
+
+.. cpp:type:: opaque* botan_ec_group_t
+
+   An opaque data type for an EC Group. Don't mess with it.
+
+.. cpp:function:: int botan_ec_group_destroy(botan_ec_group_t oid)
+
+   Destroy an object.
+
+.. cpp:function:: int botan_ec_group_supports_application_specific_group(int* out)
+
+   Checks if in this build configuration it is possible to register an application specific elliptic curve,
+   and sets ``out`` to 1 if so, 0 otherwise.
+
+.. cpp:function:: int botan_ec_group_supports_named_group(const char* name, int* out)
+
+   Checks if in this build configuration botan_ec_group_from_name(group_ptr, name) will succeed,
+   and sets ``out`` to 1 if so, 0 otherwise.
+
+.. cpp:function:: int botan_ec_group_from_params(botan_ec_group_t* ec_group, \
+                               botan_asn1_oid_t oid, \
+                               botan_mp_t p, \
+                               botan_mp_t a, \
+                               botan_mp_t b, \
+                               botan_mp_t base_x, \
+                               botan_mp_t base_y, \
+                               botan_mp_t order)
+
+   Create a new EC Group from the given parameters.
+
+   .. warning::
+      Use only elliptic curve parameters you trust.
+
+.. cpp:function:: int botan_ec_group_from_ber(botan_ec_group_t* ec_group, const uint8_t* ber, size_t ber_len)
+
+   Decode a BER encoded ECC domain parameter set
+
+.. cpp:function:: int botan_ec_group_from_pem(botan_ec_group_t* ec_group, const char* pem)
+
+   Initialize an EC Group from the PEM/ASN.1 encoding
+
+.. cpp:function:: int botan_ec_group_from_oid(botan_ec_group_t* ec_group, botan_asn1_oid_t oid)
+
+   Initialize an EC Group from a group named by an object identifier
+
+.. cpp:function:: int botan_ec_group_from_name(botan_ec_group_t* ec_group, const char* name)
+
+   Initialize an EC Group from a common group name (eg "secp256r1")
+
+.. cpp:function:: int botan_ec_group_view_der(botan_ec_group_t ec_group, botan_view_ctx ctx, botan_view_bin_fn view)
+
+   View an EC Group in DER encoding
+
+.. cpp:function:: int botan_ec_group_view_pem(botan_ec_group_t ec_group, botan_view_ctx ctx, botan_view_str_fn view)
+
+   View an EC Group in PEM encoding
+
+.. cpp:function:: int botan_ec_group_get_curve_oid(botan_asn1_oid_t* oid, botan_ec_group_t ec_group)
+
+   Get the curve OID of an EC Group
+
+.. cpp:function:: int botan_ec_group_get_p(botan_mp_t* p, botan_ec_group_t ec_group)
+
+   Get the prime modulus of the field
+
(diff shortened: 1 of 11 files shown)


src/tests/test_ffi.cpp
```diff
@@ -4275,6 +4275,233 @@ class FFI_OID_Test final : public FFI_Test {
...
+         TEST_FFI_RC(1, botan_ec_group_equal, (group_from_name, group_from_pem));
+
+         botan_privkey_t priv;
+         TEST_FFI_OK(botan_ec_privkey_create, (&priv, "ECDSA", secp384r1, rng));
+         char namebuf[32] = {0};
+         size_t name_len = sizeof(namebuf);
+
+         TEST_FFI_OK(botan_privkey_algo_name, (priv, &namebuf[0], &name_len));
+         result.test_eq(namebuf, namebuf, "ECDSA");
+
```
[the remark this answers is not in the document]
randombit: I think copilot's interpretation of the arguments isn't quite right but this does seem at least a bit confusing since the first arg to `test_eq` should be the test being formed eg "Name is expected value"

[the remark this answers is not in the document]
arckoor: Oh yes good catch, not quite sure how that happened :thinking:

randombit:
> Adds the stable part of EC_Group (so not EC_AffinePoint)

I had not realized `EC_AffinePoint` and `EC_Scalar` are still marked as unstable API. At this point it seems safe to say #4027 is a success and I can't imagine making major changes, even already efforts have been made to ensure backwards compatability where small API tweaks ended up being desirable. I'll mark those as full supported public APIs for 3.8

[No need to expose point operations in this PR, unless you happen to have a use for them; just the groups alone are useful for many cases, and anyway a necessary base for FFIing the point arithmetic in the future]

src/scripts/test_python.py
```diff
@@ -1083,6 +1083,44 @@ def test_oids(self):
         self.assertEqual(oid_rsa_pub.to_string(), oid_rsa.to_string())
         self.assertEqual(oid_rsa_pub.to_name(), oid_rsa_pub.to_name())
 
+    def test_ec_group(self):
+        self.assertTrue(botan.ECGroup.supports_application_specific_group())
+        self.assertTrue(botan.ECGroup.supports_named_group("secp256r1"))
```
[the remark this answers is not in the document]
arckoor: Should I just compare against the internal function, or disable the tests entirely if the build doesn't support it? On the python side `test_ecdh` fails either way with such a build, because it too tries to create an ecdh private key from a group name.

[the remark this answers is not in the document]
randombit: Ah right unlike in `test_ffi` there is no other value to compare against. I think here it's reasonable to check if the values are true, and if false exit early.

(Ideally the test still checks the secp256r1 case even when application specific groups aren't available)

arckoor:
I've tried to make it test as much as possible when only the named groups are supported. I've also altered some of the EC private key tests so they are skipped if their named group isn't supported.

doc/api_ref/ffi.rst
```diff
@@ -874,6 +874,100 @@ Object Identifiers
...
+.. cpp:function:: int botan_ec_group_from_params(botan_ec_group_t* ec_group, \
+                               botan_asn1_oid_t oid, \
+                               botan_mp_t p, \
+                               botan_mp_t a, \
+                               botan_mp_t b, \
+                               botan_mp_t base_x, \
+                               botan_mp_t base_y, \
+                               botan_mp_t order)
+
+   Create a new EC Group from the given parameters
```
randombit, line 909: Maybe worth referencing here the docs of the relevant `EC_Group` constructor which explains the restrictions on sizes/etc

doc/api_ref/ffi.rst
```diff
@@ -874,6 +874,100 @@ Object Identifiers
...
+
+   Get the y coordinate of the base point
+
+.. cpp:function:: int botan_ec_group_get_order(botan_mp_t* order, botan_ec_group_t ec_group)
+
+   Get the order of the base point
+
+.. cpp:function:: int botan_ec_group_equal(botan_ec_group_t a, botan_ec_group_t b)
+
+   Return 1 if ``a`` is equal to ``b``, 0 if ``a`` is not equal to ``b``
```
randombit, line 968: Maybe just to minimize confusion use `curve1` and `curve2` to avoid mixups re the curve `a`/`b` parameters

src/lib/ffi/ffi_ec.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+int botan_ec_group_from_ber(botan_ec_group_t* ec_group, const uint8_t* ber, size_t ber_len) {
+   return ffi_guard_thunk(__func__, [=]() -> int {
+      if(ec_group == nullptr || ber == nullptr) {
+         return BOTAN_FFI_ERROR_NULL_POINTER;
+      }
+
+      Botan::EC_Group group(ber, ber_len);
+
+      std::unique_ptr<Botan::EC_Group> group_ptr = std::make_unique<Botan::EC_Group>(std::move(group));
```
randombit, line 81: Could this just be

```
auto group_ptr = std::make_unique<Botan::EC_Group>(ber, ber_len);
```

?

[In general for `x = make_unique<Foo>(...)` it's assumed the type of `x` is clear and we use `auto` rather than the explicit type]

randombit: I guess OTOH what you have here better matches the pattern elsewhere, and the "static function named constructor" thing we do here interferes with `make_unique` in the general case...

src/lib/ffi/ffi.h
```diff
@@ -1159,6 +1159,141 @@ BOTAN_FFI_EXPORT(3, 8) int botan_oid_equal(botan_asn1_oid_t a, botan_asn1_oid_t
...
+* Get the order of the base point
+*/
+BOTAN_FFI_EXPORT(3, 8) int botan_ec_group_get_order(botan_mp_t* order, botan_ec_group_t ec_group);
+
+/**
+* @returns 0 if a != b
+* @returns 1 if a == b
+* @returns negative number on error
+*/
+BOTAN_FFI_EXPORT(3, 8) int botan_ec_group_equal(botan_ec_group_t a, botan_ec_group_t b);
```
randombit, line 1295: As in the comment in the docs lets use `curve1` `curve2` here

randombit (commented):
OK did a full review, a few very small nitpicks but otherwise ready to go

doc/api_ref/ffi.rst
arckoor: Can I somehow make a reference directly to that constructor? I tried adding `.. _ec_group_param_constructor:` directly before it, but it complained about duplicate references. I also tried using the function signature directly, that didn't work either, but I'm also not very knowledgeable about rst

randombit: Offhand if those don't work I'm not sure what Sphinx wants here. I'll mess with it later

randombit (approved):
The last doc changes need to be reverted to fix the build, otherwise ready to merge

PR #4834, opened 2025-04-22
Fix the test reporter to print the name of the testsuite being waited for

randombit:
Prior to #3027 the test runner would print the name of the test currently being run first, then it would run the test and print the results. This made it easy to see which test has a problem if for example something causes an infinite loop.

In the changes in #3027 this was lost, which at the time I thought was tolerable, but since I've missed it quite often.

And it turns out to be simple to add back, so let's do that.


reneme:
Thanks! This was certainly annoying at times. (sorry for introducing this. 😅)

PR #4835, opened 2025-04-22
New Barrett Reduction implementation

randombit:
The old (public) interface in `Modular_Reducer` was quite loose in that it allows negative inputs, out of range inputs, etc which we had to handle. This greatly complicated constant time implementation.

Now internally we use a new `Barrett_Reduction` class which only accepts positive and in-range inputs. The public (and already deprecated) `Modular_Reducer` interface now just uses `ct_modulo` for everything.


randombit:
Nightly run here https://github.com/randombit/botan/actions/runs/14616685403

Issue #4836, opened 2025-04-23
[Ubuntu x64] botan-config.cmake is failing on ubuntu multi architecture


Phatcat:
https://github.com/EmberEmu/Ember/actions/runs/14586233443/job/40912167037#step:3:2615

```
#16 1.202 -- Configuring done (1.1s)
#16 1.271 CMake Error in src/account/CMakeLists.txt:
#16 1.271   Imported target "Botan::Botan-static" includes non-existent path
#16 1.271 
#16 1.271     "/usr/lib/include/botan-3"
#16 1.271 
#16 1.271   in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:
#16 1.271 
#16 1.271   * The path was deleted, renamed, or moved to another location.
#16 1.271 
#16 1.271   * An install or uninstall procedure did not complete successfully.
#16 1.271 
#16 1.271   * The installation package was faulty and references files it does not
#16 1.271   provide.
```

It is looking in the wrong directory for the header files, they are actually located here:

`usr/include/botan-3`

The same goes the library, it expects it in

`usr/lib/`

But in reality it's located in

`usr/lib/x86_64-linux-gnu/`

This can be patched by running;

```
RUN sed -i 's|\${_Botan_PREFIX}/include|/usr/include|g' /usr/lib/x86_64-linux-gnu/cmake/Botan-3.7.1/botan-config.cmake
RUN sed -i 's|\${_Botan_PREFIX}/lib|/usr/lib/x86_64-linux-gnu|g' /usr/lib/x86_64-linux-gnu/cmake/Botan-3.7.1/botan-config.cmake
```

Tagging [securitykernel](https://github.com/securitykernel)

Issues from the following process:

```
# syntax=docker/dockerfile:experimental

FROM ubuntu:plucky AS builder
LABEL description="Development build environment"

# Update the distro and install our tools
RUN apt-get -y update && apt-get -y upgrade \
 && apt-get -y install software-properties-common \
 && apt-get -y install wget \
 # GCC stuff
 && apt-get -y install build-essential gcc-14 g++-14 \
 # Clang stuff
 #&& wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc >/dev/null \
 #&& add-apt-repository -y 'deb http://apt.llvm.org/noble/ llvm-toolchain-noble-18 main' \
 #&& apt update \
[...]

PR #4839, opened 2025-04-24
FIX: handle multi-arch directories in cmake-config package

reneme:
See #4836.

Essentially this adds mechanics to handle installations with an additional multi-arch directory. I.e. when the library is configured like `./configure.py --prefix=$(pwd)/install --libdir=lib/aarch64-linux-gnu`.

Currently, the cmake package tests in our CI only test the "default" installation layout (without an explicit `--libdir`) for the "static" and "shared" targets on macOS, Linux, and Windows. It would be good to also test with a multi-arch directory.

Perhaps when building "static" on Linux?


reneme:
Please don't merge this, yet. I'd like to give it another round of tests. Perhaps even build the Botan Ubuntu package and see if it works as expected.

I'm traveling over the weekend, so that might be a few days.

reneme:
... works on my machine.

PR #4840, opened 2025-04-25
Update EC_Group::PEM_encode to handle upcoming removal of explicit curves

randombit:
A very long time ago, an unfortunate decision was made that this function should always return the explicit encoding. This directly conflicts with the desire to remove such encodings entirely in Botan4, if at all practical.

Add an optional argument that allows specifying the encoding. And update the new-in-3.8 FFI function so that, there, the named curve encoding is always used.


randombit:
> hopefully that functionality (returning explicit when requested) will remain in the future.

Any reason why? There seems to be few upsides and many drawbacks to explicit curve encoding.

The likely outcome in the medium term is serialization will only use the OID encoding, while parsing will continue to accept the explicit encoding but only if it's encoding a group we otherwise know about.

PR #4841, opened 2025-04-26
Generate ec_named.cpp and relevant pcurves files from gen_ec_groups.py

randombit:
This is useful for end users who want to get curve-specific optimizations for some unusual or application specific curve.

@reneme I think we discussed this offline a while back


reneme:
Amazing, thanks! Do you think this should see some CI coverage? I could look into this.

src/scripts/dev_tools/gen_pcurves.py
reneme, line 1: I took the liberty to clean up this implementation a little bit. To my mind that is a fair bit more readable: https://github.com/reneme/botan/blob/jack/add-pcurves-gen-script/src/scripts/dev_tools/gen_pcurves.py

randombit: :+1: this was pretty quick and dirty, your changes look great can you merge here?

reneme: done.

randombit:
[This comment](https://github.com/randombit/botan/pull/4706#issuecomment-2673067229) describes a somewhat different approach that would be a bit more user-friendly, since end user would just need to re-run one script

reneme:
This branch gets close enough, no? When I tested it the other day, the only manual thing I had to do was to add the instantiation of my "funky chicken curve". :) Or am I missing something?
In any case: the easier, the better, obviously. ;)

randombit:
> This branch gets close enough, no?

I mean it's a good step, but you would still have to update `ec_named.cpp` which is a distinct step with another script, plus also hand modify `pcurves_instance.h` since that's currently not generated. Ideally the flow is

```
cat my_curve_info.txt >> src/build-data/ec_groups.txt
./src/scripts/gen_ec_groups.py
./configure.py
make
```

If we reached that level of automation it would also be easy to CI test that it actually works.

randombit:
Updated. Now everything is generated with `gen_ec_groups.py`

randombit:
@reneme just wanted your take on the new version. This seems low risk to me so ideally could still land in time for 3.8

reneme:
> just wanted your take on the new version.

Sorry, I was caught up in other things and only now see that you pinged me here. I won't have a chance to look at this until the last week of May unfortunately. No problem from our side that it didn't land in 3.8.x. Nevertheless, thanks a lot for the effort, this will come in handy for sure. 👍

src/scripts/dev_tools/gen_ec_groups.py
```diff
@@ -115,32 +107,100 @@ def format_names(names):
...
+        try:
+            os.makedirs(mod_dir)
+        except OSError as ex:
+            if ex.errno != errno.EEXIST:
+                raise
+
+        with open(info_path, 'w', encoding='utf8') as info_file:
+            info_file.write(dedent(f"""\
+                <defines>
+                PCURVES_{curve.upper()} -> {datestamp()}
```
reneme, line 179: If the provided EC curve name cannot be transformed into [a valid module preprocessor definition](https://github.com/randombit/botan/blob/b3c5e866bdbd428c239c631e7988a02f0bd61225/configure.py#L965), we should error out. Otherwise `./configure.py` will fail ungracefully. (I ran into this by chance while manually trying the documented flow.)

```suggestion
        module_define = f"PCURVES_{curve.upper()}"
        if not re.match('^[0-9A-Za-z_]{3,30}$', module_define):
            raise ValueError(f"Invalid preprocessor define name ({module_define}) for new pcurve module")

        with open(info_path, 'w', encoding='utf8') as info_file:
            info_file.write(dedent(f"""\
                <defines>
                {module_define} -> {datestamp()}
```

reneme (approved):
This works beautifully! Thanks a lot.

PR #4843, opened 2025-04-28
Prefer AES/GCM over ChaCha20Poly1305 in TLS

randombit:
For processors with hardware support, AES is much faster that ChaCha20Poly1305.


randombit:
@reneme There is something really odd happening in the cli TLS tests with this change. ` TestConfig("PSK TLS 1.3", "1.3", ...` fails with a somewhat nonsensical error that the server cannot find an appropriate cert. Adding `ciphers=ChaCha20Poly1305` to the test policy, it works again. I'm not really sure what's going on but it seems like something is broken here...

reneme:
> I'm not really sure what's going on but it seems like something is broken here...

My first observation: The "cannot find appropriate certificate" seems to be a red herring. Actually, the server produces a HelloRetryRequest and then rejects the second ClientHello attempt from the client because of a missing PSK extension. This results in an IllegalParameter alert and a failing handshake.

Two questions arise: 

* why does the added chacha cipher fix this, and
* why does the client's second attempt fail the way it does?

reneme:
The client offered a traditional X25519 key share but claims support for a PQ-hybrid without providing a key share for it. The server wanting to enforce PQ-safety issues a HelloRetryRequest. In the process of creating a second client hello, the client will call `PSK::filter()` on its PSK extension.

Current working hypothesis: Without the chacha cipher, the `PSK::filter()` removes the configured PSK, which causes the PSK extension to be `empty()` and is therefor dropped from the second client hello. This trips up the server with the IllegalParameter and failing the handshake.

reneme:
Turns out: the new default cipher suite which employs AES-256/GCM requires to use SHA-384 as the PRF. Externally provided PSKs have to also define the hash algorithm to be used ([RFC 8446 4.2.11](https://www.rfc-editor.org/rfc/rfc8446#section-4.2.11)), or default to "SHA-256".

> the Hash algorithm MUST be set when the PSK is established or default to SHA-256 if no such algorithm is defined.

As a result, the PSK defined in the test is not compatible with the AES-256/GCM cipher suite and the handshake fails. This is further obscured by the fact that we have a HelloRetryRequest roundtrip (to update the connection to use PQC).

I'll push a commit with a fix to the test onto this branch shortly.

reneme (approved):

PR #4844, opened 2025-04-28
In GHASH skip the slow key schedule if we have CPU instruction support

randombit:
We spent a significant amount of timing (and memory) setting up an expanded key schedule that will never be used.

### DIFF
--- a/src/lib/utils/ghash/ghash.cpp
+++ b/src/lib/utils/ghash/ghash.cpp
@@ -86,7 +86,7 @@ void GHASH::ghash_multiply(std::span<uint8_t, GCM_BS> x, std::span<const uint8_t
 }
 
 bool GHASH::has_keying_material() const {
-   return !m_HM.empty();
+   return !m_HM.empty() || !m_H_pow.empty();
 }
 
 void GHASH::key_schedule(std::span<const uint8_t> key) {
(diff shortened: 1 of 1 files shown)


src/lib/utils/ghash/ghash.cpp
```diff
@@ -97,6 +97,18 @@ void GHASH::key_schedule(std::span<const uint8_t> key) {
    BOTAN_ASSERT_NOMSG(key.size() == GCM_BS);
    auto H = load_be<std::array<uint64_t, 2>>(key.first<GCM_BS>());
 
+#if defined(BOTAN_HAS_GHASH_CLMUL_CPU)
+   if(CPUID::has_carryless_multiply()) {
+      zap(m_HM);
+      if(m_H_pow.size() != 8) {
+         m_H_pow.resize(8);
+      }
```
reneme, line 105: I'm fairly sure that [resize is a noop](https://en.cppreference.com/w/cpp/container/vector/resize) when the buffer is already of that size, no? Or am I missing something?

reneme (approved):

src/lib/utils/ghash/ghash.cpp
randombit: I think that is true, just a nervous tic on my part

PR #4873, opened 2025-05-16
Use std::span for block cipher padding schemes

reneme:
Modernize the padding scheme implementations and avoid imposing a byte buffer type (`secure_vector<uint8_t>`) on the downstream user. Instead, using code is now expected to pass a `std::span` with enough extra space to accommodate the padding bytes exactly. That way, the padding scheme implementations avoid handling any re-allocations or copying.

Additionally, the base class now has an internal customization point allowing to share some validation code across most implementations.

This is a prerequisite to refactor the internal customization point `Cipher_Mode::finish_msg()` to also take a sufficiently-sized `std::span` instead of a `secure_vector`. Eventually, I hope to have a public `Cipher_Mode::finish()` method that doesn't force users to pass their data via a `secure_vector` without providing a `std::span` alternative.

### DIFF
--- a/src/lib/modes/mode_pad/mode_pad.cpp
+++ b/src/lib/modes/mode_pad/mode_pad.cpp
@@ -2,6 +2,7 @@
 * CBC Padding Methods
 * (C) 1999-2007,2013,2018,2020 Jack Lloyd
 * (C) 2016 René Korthaus, Rohde & Schwarz Cybersecurity
+* (C) 2025 René Meusel, Rohde & Schwarz Cybersecurity
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
@@ -40,10 +41,29 @@ std::unique_ptr<BlockCipherModePaddingMethod> BlockCipherModePaddingMethod::crea
    return nullptr;
 }
 
+void BlockCipherModePaddingMethod::add_padding(std::span<uint8_t> buffer, size_t last_byte_pos, size_t BS) const {
+   BOTAN_ASSERT_NOMSG(valid_blocksize(BS));
+   BOTAN_ASSERT_NOMSG(last_byte_pos < BS);
+   BOTAN_ASSERT_NOMSG(buffer.size() % BS == 0);
+   BOTAN_ASSERT_NOMSG(buffer.size() >= BS);
+
+   auto poison = CT::scoped_poison(last_byte_pos, buffer);
+   apply_padding(buffer.last(BS), last_byte_pos);
+}
+
+size_t BlockCipherModePaddingMethod::unpad(std::span<const uint8_t> last_block) const {
+   if(!valid_blocksize(last_block.size())) {
+      return last_block.size();
+   }
+
+   auto poison = CT::scoped_poison(last_block);
+   return CT::driveby_unpoison(remove_padding(last_block));
+}
+
 /*
 * Pad with PKCS #7 Method
 */
-void PKCS7_Padding::add_padding(secure_vector<uint8_t>& buffer, size_t last_byte_pos, size_t BS) const {
+void PKCS7_Padding::apply_padding(std::span<uint8_t> last_block, size_t padding_start_pos) const {
    /*
    Padding format is
    01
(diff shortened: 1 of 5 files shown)


KaganCanSit:
Hello Mr. René Meusel (@reneme),

I wanted to help you while I had time. You helped me with spans in a merge request I opened last time. After I pulled the content of the branch that was receiving the error to my local environment and made a very small parameter adjustment, the compilation was successful with the following command. Then, when I ran the _**botan-test**_ application, I did not get any errors.

Compilation Command:
`./configure.py && make clean && make -j$(nproc)`

Changes I made:
https://github.com/KaganCanSit/botan/commit/84137204472f40bccf989c998c36a471838d91df

I quickly adding it as a review message, I hope it saves you time.

Best regards.

reneme:
> I hope it saves you time.

Thanks! I incorporated your changes and CI is green now. 👍

KaganCanSit:
> > I hope it saves you time.
> 
> Thanks! I incorporated your changes and CI is green now. 👍

You're welcome. Glad I could help. I'm always open to improvements that I can make and learn from.

src/lib/modes/mode_pad/mode_pad.h
```diff
@@ -63,16 +76,36 @@ class BOTAN_TEST_API BlockCipherModePaddingMethod {
...
+      /**
+      * Applies the concrete padding to the @p last_block assuming the padding
+      * bytes should start at @p padding_start_pos within the last block.
+      *
+      * Concrete implementations of this function must ensure not to leak
+      * @p padding_start_pos via side channels.
+      */
+      virtual void apply_padding(std::span<uint8_t> last_block, size_t padding_start_pos) const {
+         BOTAN_UNUSED(last_block, padding_start_pos);
+         BOTAN_ASSERT_UNREACHABLE();
```
randombit, line 90: I don't understand why this isn't a pure virtual

src/lib/modes/mode_pad/mode_pad.cpp
```diff
@@ -40,63 +41,61 @@ std::unique_ptr<BlockCipherModePaddingMethod> BlockCipherModePaddingMethod::crea
    return nullptr;
 }
 
+void BlockCipherModePaddingMethod::add_padding(std::span<uint8_t> buffer, size_t last_byte_pos, size_t BS) const {
+   BOTAN_DEBUG_ASSERT(valid_blocksize(BS));
+   BOTAN_DEBUG_ASSERT(last_byte_pos < BS);
+   BOTAN_DEBUG_ASSERT(buffer.size() % BS == 0);
+   BOTAN_DEBUG_ASSERT(buffer.size() >= BS);
```
randombit, line 48: These are all cheap to check might as well do it even in prod mode

src/lib/modes/mode_pad/mode_pad.h
```diff
@@ -63,16 +76,36 @@ class BOTAN_TEST_API BlockCipherModePaddingMethod {
...
       */
       virtual ~BlockCipherModePaddingMethod() = default;
+
+   protected:
+      /**
+      * Applies the concrete padding to the @p last_block assuming the padding
+      * bytes should start at @p padding_start_pos within the last block.
+      *
+      * Concrete implementations of this function must ensure not to leak
+      * @p padding_start_pos via side channels.
```
randombit, line 86: This should note that this function is always called with the input span in a poisoned state

src/lib/modes/mode_pad/mode_pad.h
```diff
@@ -63,16 +76,36 @@ class BOTAN_TEST_API BlockCipherModePaddingMethod {
...
+      */
+      virtual void apply_padding(std::span<uint8_t> last_block, size_t padding_start_pos) const {
+         BOTAN_UNUSED(last_block, padding_start_pos);
+         BOTAN_ASSERT_UNREACHABLE();
+      }
+
+      /**
+      * Removes the padding from @p last_block and returns the number of data
+      * bytes. If the padding is invalid, this returns the byte length of
+      * @p last_block.
```
randombit, line 96: Likewise this should note that the input is poisoned and the implementation must be constant time

randombit (commented):
Looks good

src/lib/modes/mode_pad/mode_pad.h
reneme: To support `Null_Padding` the `add_padding` method (that contains the sanity checks) must be `virtual` and is overridden by `Null_Padding` as a NOOP. Defining this `apply_padding` customization point with a "useless" implementation allowed me to avoid overriding `apply_padding` in `Null_Padding`.

I agree though, that this is fairly unusual. Changed to pure virtual here and moved the `BOTAN_ASSERT_UNREACHABLE` override definition down into `Null_Padding`.

randombit (approved):

PR #4877, opened 2025-05-20
Add FFI for X.509 creation

arckoor:
Adds bindings for creating X.509 certificates.

### DIFF
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -2194,6 +2194,49 @@ int botan_x509_cert_verify(int* validation_result,
 */
 BOTAN_FFI_EXPORT(2, 8) const char* botan_x509_cert_validation_status(int code);
 
+typedef struct botan_x509_cert_opts_struct* botan_x509_cert_opts_t;
+typedef struct botan_x509_ca_struct* botan_x509_ca_t;
+typedef struct botan_x509_pkcs10_req_struct* botan_x509_pkcs10_req_t;
+typedef struct botan_x509_time_struct* botan_x509_time_t;
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_create_cert_opts(botan_x509_cert_opts_t* opts_obj, const char* opts, uint32_t* expire_time);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_ca_key(botan_x509_cert_opts_t opts, size_t limit);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_create_self_signed_cert(botan_x509_cert_t* cert_obj,
+                                       botan_privkey_t key,
+                                       botan_x509_cert_opts_t opts,
+                                       const char* hash_fn,
+                                       const char* sig_padding,
+                                       botan_rng_t rng);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_create_ca(botan_x509_ca_t* ca_obj,
+                         botan_x509_cert_t ca_cert,
+                         botan_privkey_t key,
+                         const char* hash_fn,
+                         const char* sig_padding,
+                         botan_rng_t rng);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_create_pkcs10_req(botan_x509_pkcs10_req_t* req_obj,
+                                 botan_x509_cert_opts_t opts,
+                                 botan_privkey_t key,
+                                 const char* hash_fn,
+                                 botan_rng_t rng);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_sign_req(botan_x509_cert_t* cert_obj,
+                        botan_x509_ca_t ca,
+                        botan_x509_pkcs10_req_t req,
+                        botan_rng_t rng,
+                        botan_x509_time_t not_before,
+                        botan_x509_time_t not_after);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_create_time(botan_x509_time_t* time_obj, uint64_t time_since_epoch);
+
 /*
 * X.509 CRL
 **************************/
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -2126,6 +2126,12 @@ BOTAN_FFI_EXPORT(2, 0) int botan_x509_cert_get_serial_number(botan_x509_cert_t c
 BOTAN_FFI_EXPORT(2, 0) int botan_x509_cert_get_authority_key_id(botan_x509_cert_t cert, uint8_t out[], size_t* out_len);
 BOTAN_FFI_EXPORT(2, 0) int botan_x509_cert_get_subject_key_id(botan_x509_cert_t cert, uint8_t out[], size_t* out_len);
 
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_get_basic_constraints(botan_x509_cert_t cert, int* is_ca, size_t* limit);
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_get_key_constraints(botan_x509_cert_t cert, uint32_t* usage);
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_get_ocsp_responder(botan_x509_cert_t cert, botan_view_ctx ctx, botan_view_str_fn view);
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_is_self_signed(botan_x509_cert_t cert, int* out);
+
 BOTAN_FFI_EXPORT(2, 0) int botan_x509_cert_get_public_key_bits(botan_x509_cert_t cert, uint8_t out[], size_t* out_len);
 
 BOTAN_FFI_EXPORT(3, 0)
@@ -2145,6 +2151,8 @@ BOTAN_FFI_EXPORT(2, 0) int botan_x509_cert_to_string(botan_x509_cert_t cert, cha
 
 BOTAN_FFI_EXPORT(3, 0)
 int botan_x509_cert_view_as_string(botan_x509_cert_t cert, botan_view_ctx ctx, botan_view_str_fn view);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_view_pem(botan_x509_cert_t cert, botan_view_ctx ctx, botan_view_str_fn view);
 
 /* Must match values of Key_Constraints in key_constraints.h */
 enum botan_x509_cert_key_constraints {
@@ -2203,14 +2211,56 @@ BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_destroy(botan_x509_cert_opts_t o
 
 BOTAN_FFI_EXPORT(3, 9) int botan_x509_ca_destroy(botan_x509_ca_t ca);
 
-BOTAN_FFI_EXPORT(3, 9) int botan_x509_plcs10_req_destroy(botan_x509_pkcs10_req_t req);
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_pkcs10_req_destroy(botan_x509_pkcs10_req_t req);
 
 BOTAN_FFI_EXPORT(3, 9) int botan_x509_time_destroy(botan_x509_time_t time);
 
 BOTAN_FFI_EXPORT(3, 9)
 int botan_x509_create_cert_opts(botan_x509_cert_opts_t* opts_obj, const char* opts, uint32_t* expire_time);
 
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_common_name(botan_x509_cert_opts_t opts, const char* name);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_country(botan_x509_cert_opts_t opts, const char* country);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_organization(botan_x509_cert_opts_t opts, const char* organization);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_org_unit(botan_x509_cert_opts_t opts, const char* org_unit);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_locality(botan_x509_cert_opts_t opts, const char* locality);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_state(botan_x509_cert_opts_t opts, const char* state);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_serial_number(botan_x509_cert_opts_t opts, const char* serial_number);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_email(botan_x509_cert_opts_t opts, const char* email);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_uri(botan_x509_cert_opts_t opts, const char* uri);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_ip(botan_x509_cert_opts_t opts, const char* ip);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_dns(botan_x509_cert_opts_t opts, const char* dns);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_xmpp(botan_x509_cert_opts_t opts, const char* xmpp);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_challenge(botan_x509_cert_opts_t opts, const char* challenge);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_cert_opts_more_org_units(botan_x509_cert_opts_t opts, const char** more_org_units, size_t cnt);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_cert_opts_more_dns(botan_x509_cert_opts_t opts, const char** more_dns, size_t cnt);
+
 BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_ca_key(botan_x509_cert_opts_t opts, size_t limit);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_set_padding_scheme(botan_x509_cert_opts_t opts, const char* scheme);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_not_before(botan_x509_cert_opts_t opts, botan_x509_time_t not_before);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_not_after(botan_x509_cert_opts_t opts, botan_x509_time_t not_after);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_add_constraints(botan_x509_cert_opts_t opts, uint32_t usage);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_add_ex_constraint(botan_x509_cert_opts_t opts, botan_asn1_oid_t oid);
 
 BOTAN_FFI_EXPORT(3, 9)
 int botan_x509_create_self_signed_cert(botan_x509_cert_t* cert_obj,
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -2199,6 +2199,14 @@ typedef struct botan_x509_ca_struct* botan_x509_ca_t;
 typedef struct botan_x509_pkcs10_req_struct* botan_x509_pkcs10_req_t;
 typedef struct botan_x509_time_struct* botan_x509_time_t;
 
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_destroy(botan_x509_cert_opts_t opts);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_ca_destroy(botan_x509_ca_t ca);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_plcs10_req_destroy(botan_x509_pkcs10_req_t req);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_time_destroy(botan_x509_time_t time);
+
 BOTAN_FFI_EXPORT(3, 9)
 int botan_x509_create_cert_opts(botan_x509_cert_opts_t* opts_obj, const char* opts, uint32_t* expire_time);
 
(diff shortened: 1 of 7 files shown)


src/python/botan3.py
```diff
@@ -1771,14 +1813,144 @@ def _load_buf_or_file(filename, buf, file_fn, buf_fn):
...
+        usage_values = {"NO_CONSTRAINTS": 0,
+                        "DIGITAL_SIGNATURE": 32768,
+                        "NON_REPUDIATION": 16384,
+                        "KEY_ENCIPHERMENT": 8192,
+                        "DATA_ENCIPHERMENT": 4096,
+                        "KEY_AGREEMENT": 2048,
+                        "KEY_CERT_SIGN": 1024,
+                        "CRL_SIGN": 512,
+                        "ENCIPHER_ONLY": 256,
+                        "DECIPHER_ONLY": 128}
```
arckoor, line 1921: This exists in 3 different places now, I really wish it were an enum. But I suppose that would be a breaking change for `X509Cert().allowed_usage()`?

src/lib/ffi/ffi_cert.cpp
```diff
@@ -163,6 +169,79 @@ int botan_x509_cert_allowed_usage(botan_x509_cert_t cert, unsigned int key_usage
...
+   BOTAN_UNUSED(cert, out)
+   return BOTAN_FFI_ERROR_NOT_IMPLEMENTED;
+#endif
+}
+
+// TODO name constraints
+// TODO extended key usage
+// TODO cert policies
+// TODO ca issuers
+// TODO crl dist points
```
arckoor, line 243: These return `vec<T>`s, which I'm not really sure how to pass through the FFI. Strings and OIDs might be possible, but `name_constraints` consists of `vec<GeneralSubtree>`, which I don't see happening at all.
That said, I don't know if we actually *need* all of those in the FFI.

arckoor:
@randombit @reneme could you take a look and tell me what you think? Still WIP, almost no tests, but the python binding is usable-ish

arckoor:
The normal copilot comments I already mentioned myself, funnily enough the low-confidence suppressed ones are valid issues :thinking:

src/lib/ffi/ffi_cert.cpp
randombit: Fine to expose only a subset here, especially for the first pass

src/python/botan3.py
arckoor: Also apart from the above, this version is a setter, so it might be better to yell at the user when they pass something invalid rather than silently `pass`

src/python/botan3.py
```diff
@@ -483,17 +483,59 @@ def ffi_api(fn, args, allowed_errors=None):
...
+    ffi_api(dll.botan_x509_cert_opts_state, [c_void_p, c_char_p])
+    ffi_api(dll.botan_x509_cert_opts_serial_number, [c_void_p, c_char_p])
+    ffi_api(dll.botan_x509_cert_opts_email, [c_void_p, c_char_p])
+    ffi_api(dll.botan_x509_cert_opts_uri, [c_void_p, c_char_p])
+    ffi_api(dll.botan_x509_cert_opts_ip, [c_void_p, c_char_p])
+    ffi_api(dll.botan_x509_cert_opts_dns, [c_void_p, c_char_p])
+    ffi_api(dll.botan_x509_cert_opts_xmpp, [c_void_p, c_char_p])
+    ffi_api(dll.botan_x509_cert_opts_challenge, [c_void_p, c_char_p])
+    ffi_api(dll.botan_x509_cert_opts_more_org_units, [c_void_p, POINTER(c_char_p), c_size_t])
+    ffi_api(dll.botan_x509_cert_opts_more_dns, [c_void_p, POINTER(c_char_p), c_size_t])
```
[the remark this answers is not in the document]
randombit: `gen_ffi_decls.py` is definitely ... ah, works on a best effort basis I guess would be the nicest way to put it.

arckoor:
Won't compile, needs #4890

arckoor:
@randombit I think this is ready to take a look at. The python binding definitely still needs some attention, and I'm not sure how much more you want me to document. But all in all it works.

src/lib/ffi/ffi.h
```diff
@@ -2196,6 +2204,182 @@ int botan_x509_cert_verify(int* validation_result,
 */
 BOTAN_FFI_EXPORT(2, 8) const char* botan_x509_cert_validation_status(int code);
 
+typedef struct botan_x509_cert_opts_struct* botan_x509_cert_opts_t;
```
randombit: Both the name and general structure of `X509_Cert_Options` are not necessarily the best, it's just kind of been there. So I don't think it should necessarily be reflected in the FFI interface. At least to the extent it is avoidable.

Specific things resulting from this:

Instead of `opts` or "options" (which can mean a lot of things and generally speaking I don't think anyone refers to eg the commonName field of a cert as an "option", this was just bad terminology on my part) use ... uh IDK. Maybe `botan_x509_cert_params_builder_struct`?

Also let's pretend this is a builder even when it's sometimes not. (So later we can add a proper builder in the C++ side and just swap out the FFI, instead of embedding this limited struct into the FFI forever). So for example

> botan_x509_cert_opts_org_unit(botan_x509_cert_opts_t opts, const char* org_unit);

Call this

> botan_x509_cert_params_builder_add_org_unit(botan_x509_cert_params_builder_t params, const char* org_unit);

with the idea that an application could call it more than once. Now, with as `X509_Cert_Options` as defined, it can't actually be called more than once, because there is just a single `org_unit` member. So this function would check if `org_unit` was already set to some value, and if so return an error. (Which one? Good question... maybe `BOTAN_FFI_ERROR_INVALID_OBJECT_STATE`?? `)

src/lib/ffi/ffi.h
```diff
@@ -2196,6 +2204,182 @@ int botan_x509_cert_verify(int* validation_result,
...
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_time_destroy(botan_x509_time_t time);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_ext_ip_addr_blocks_destroy(botan_x509_ext_ip_addr_blocks_t ip_addr_blocks);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_ext_as_blocks_destroy(botan_x509_ext_as_blocks_t as_blocks);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_ca_destroy(botan_x509_ca_t ca);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_pkcs10_req_destroy(botan_x509_pkcs10_req_t req);
```
randombit: Keep the declarations sorted by type. Eg

```
typedef struct botan_x509_pkcs10_req_struct* botan_x509_pkcs10_req_t;
// all declarations relating to botan_x509_pkcs10_req_t 

... likewise for the others
```

src/lib/ffi/ffi.h
randombit: I happend in my example above to use `org_unit` forgetting there is also a `more_org_units` field. OK well in that case `botan_x509_cert_params_builder_add_org_unit` should overflow into the `more_org_units` if it is called more than once. (And the handling as I describe would apply to eg CN, organization, country field, etc)

src/lib/ffi/ffi.h
```diff
@@ -2196,6 +2204,182 @@ int botan_x509_cert_verify(int* validation_result,
...
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_ex_constraint(botan_x509_cert_opts_t opts, botan_asn1_oid_t oid);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_cert_opts_ext_ip_addr_blocks(botan_x509_cert_opts_t opts,
+                                            botan_x509_ext_ip_addr_blocks_t ip_addr_blocks);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_cert_opts_ext_as_blocks(botan_x509_cert_opts_t opts, botan_x509_ext_as_blocks_t as_blocks);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_create_time(botan_x509_time_t* time_obj, uint64_t time_since_epoch);
```
randombit: This is the only exposed function for the time type. Do we really need to expose it at all? It seems simpler to just have the not_before/not_after set as `uint64_t time_since_epoch` and we can convert as necessary internally.

src/lib/ffi/ffi.h
```diff
@@ -2196,6 +2204,182 @@ int botan_x509_cert_verify(int* validation_result,
...
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_create_self_signed_cert(botan_x509_cert_t* cert_obj,
+                                       botan_privkey_t key,
+                                       botan_x509_cert_opts_t opts,
+                                       const char* hash_fn,
+                                       const char* sig_padding,
+                                       botan_rng_t rng);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_create_ca(botan_x509_ca_t* ca_obj,
```
randombit: Likewise for the time type, the CA type is really only used for the one operation. (Similarly to `X509_Cert_Options`, `X509_CA` type is not necessarily the best thought out interface, and I don't think we should lock ourselves into it on the FFI side).

Let's just drop `botan_x509_ca_t` entirely and have a single function

```
int botan_x509_sign_cert_req(botan_x509_cert_t* subject_cert,
                        botan_x509_pkcs10_req_t subject_req,
                        botan_x509_cert_t issuing_cert,
                        botan_x509_privkey_t issuing_key,
                        botan_rng_t rng,
                        uint64_t not_before,
                        uint64_t not_after,
                        const char* hash_fn,
                        const char* padding);
```

arckoor: What do you think of ditching the `pkcs10_req` entirely too then?
I'm thinking exposing a `CertificateBuilder`, where you can call all the nice `add_*` functions, and then at the end either `self_sign` or `sign`.

randombit:
@arckoor At this point I've just looked over the declarations in ffi.h but this already should give you something to go on

src/lib/ffi/ffi.h
```diff
@@ -2196,6 +2204,182 @@ int botan_x509_cert_verify(int* validation_result,
...
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_cert_opts_more_org_units(botan_x509_cert_opts_t opts, const char** more_org_units, size_t cnt);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_cert_opts_more_dns(botan_x509_cert_opts_t opts, const char** more_dns, size_t cnt);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_ca_key(botan_x509_cert_opts_t opts, size_t limit);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_padding_scheme(botan_x509_cert_opts_t opts, const char* scheme);
```
randombit: Let's remove this one entirely. The padding scheme shouldn't have been part of that struct in the first place. (It also is *only* relevant for RSA and the desire to use RSA-PSS certs).

src/lib/ffi/ffi.h
```diff
@@ -2196,6 +2204,182 @@ int botan_x509_cert_verify(int* validation_result,
...
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_challenge(botan_x509_cert_opts_t opts, const char* challenge);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_cert_opts_more_org_units(botan_x509_cert_opts_t opts, const char** more_org_units, size_t cnt);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_cert_opts_more_dns(botan_x509_cert_opts_t opts, const char** more_dns, size_t cnt);
+
+BOTAN_FFI_EXPORT(3, 9) int botan_x509_cert_opts_ca_key(botan_x509_cert_opts_t opts, size_t limit);
```
randombit: Maybe `mark_as_ca_key`? Or `request_ca_key`?

src/lib/ffi/ffi.h
randombit: The PKCS10 struct makes sense as a struct since that's a standard thing that is serialized, sent over a network to another party, etc.

doc/api_ref/ffi.rst
```diff
@@ -1766,6 +1791,229 @@ X.509 Certificates
...
+
+   Retrieve information on a single asnum entry.
+
+.. cpp:function::int botan_x509_ext_as_blocks_get_rdi(botan_x509_ext_as_blocks_t as_blocks, int* present, size_t* count)
+
+.. cpp:function::int botan_x509_ext_as_blocks_get_rdi_at(botan_x509_ext_as_blocks_t as_blocks, size_t i, uint32_t* min, uint32_t* max)
+
+.. cpp:type:: opaque* botan_x509_cert_params_builder_t
+
+.. cpp:function::int botan_x509_cert_opts_destroy(botan_x509_cert_opts_t opts)
```
randombit: Struct reference here is out of date

doc/api_ref/ffi.rst
```diff
@@ -1766,6 +1791,229 @@ X.509 Certificates
...
+
+   Destroy the options object.
+
+.. cpp:function::int botan_x509_create_cert_params_builder(botan_x509_cert_params_builder_t* builder_obj, \
+                  const char* opts, \
```
randombit: I had missed this in the first look at the FFI decls, but lets not expose this parsing behavior - FFI should just call add_common_name, etc.

Likewise let's drop the optional expiration time relative to the current clock. This actually isn't used/useful! PKCS10 requests don't include any kind of validity period (that's up to the signer) and for creating self-signed certs the value is specified already and this value isn't used.

doc/api_ref/ffi.rst
```diff
@@ -1766,6 +1791,229 @@ X.509 Certificates
...
+
+.. cpp:function::int botan_x509_cert_params_builder_add_uri(botan_x509_cert_params_builder_t builder, const char* uri);
+
+.. cpp:function::int botan_x509_cert_params_builder_add_ip(botan_x509_cert_params_builder_t builder, const char* ip);
+
+.. cpp:function::int botan_x509_cert_params_builder_add_dns(botan_x509_cert_params_builder_t builder, const char* dns);
+
+.. cpp:function::int botan_x509_cert_params_builder_add_xmpp(botan_x509_cert_params_builder_t builder, const char* xmpp);
+
+.. cpp:function::int botan_x509_cert_params_builder_add_challenge(botan_x509_cert_params_builder_t builder, const char* challenge);
```
randombit: This one also should be dropped: a challenge password only makes sense/is used with respect to PKCS10 and I think it is quite obsolete/unused now.

src/lib/ffi/ffi.h
```diff
@@ -2196,6 +2204,198 @@ int botan_x509_cert_verify(int* validation_result,
...
+                                 botan_privkey_t key,
+                                 const char* hash_fn,
+                                 botan_rng_t rng);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_pkcs10_req_view_pem(botan_x509_pkcs10_req_t req, botan_view_ctx ctx, botan_view_str_fn view);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_sign_req(botan_x509_cert_t* subject_cert,
+                        botan_x509_pkcs10_req_t subject_req,
```
randombit: For PKCS10, what is here isn't really that useful. You can create a request. You can serialize it as PEM (but can't deserialize it). And you can sign one. But since you can't examine the request, you have no idea what you're signing!

This would either have to be expanded with getters on the various fields of the PKCS10 request, plus also deserialization. Or alternately can be dropped for this round, either is fine with me.

src/lib/ffi/ffi.h
```diff
@@ -2196,6 +2204,198 @@ int botan_x509_cert_verify(int* validation_result,
...
+int botan_x509_cert_params_builder_add_challenge(botan_x509_cert_params_builder_t builder, const char* challenge);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_cert_params_builder_mark_as_ca_key(botan_x509_cert_params_builder_t builder, size_t limit);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_cert_params_builder_add_not_before(botan_x509_cert_params_builder_t builder, uint64_t time_since_epoch);
+
+BOTAN_FFI_EXPORT(3, 9)
+int botan_x509_cert_params_builder_add_not_after(botan_x509_cert_params_builder_t builder, uint64_t time_since_epoch);
```
randombit: Remove these from the params builder - PKCS10 doesn't use them. Instead add not_before/not_after as args to `botan_x509_create_self_signed_cert` similar to `botan_x509_sign_req`

randombit (commented):
Some more comments on the interface. Haven't reviewed the implementation or tests yet.

Looking at this PR made me realize just how janky `X509_Cert_Options` is. I opened #4996 which adds a much cleaner interface which is quite similar to what you have now in the FFI with the idea that the FFI type is (at least eventually) backed by `CertificateParametersBuilder` instead of `X509_Cert_Options`

src/lib/ffi/ffi.h
arckoor: Yeah I will have a look what else I have to expose, turns out I need a lot more functionality in the ffi after all, like complete CSRs, CRL handling and the like. I'll try to get it done in a reasonable time so that there is at least the chance of getting it in this release cycle, but if that doesn't work, oh well

src/scripts/test_python.py
```diff
@@ -783,6 +784,142 @@ def test_certs(self):
...
+            _ = cert.ext_ip_addr_blocks()
+
+        with self.assertRaisesRegex(botan.BotanException, r".*No value available.*"):
+            _ = cert.ext_as_blocks()
+
+        self.assertEqual(cert.verify(None, [ca_cert]), 0)
+
+    def test_x509_rpki(self):
+        hash_fn = "SHA-256"
+        padding_method = "EMSA1(SHA-256)"
```
randombit: Shouldn't be necessary anymore; for ECDSA, padding should be able to be specified as `None`. The only case where this has to be set anymore is for RSA-PSS

doc/api_ref/python.rst
```diff
@@ -715,11 +715,145 @@ HOTP
...
+
+   .. py:method:: rdi()
+
+PKCS10Req
+-----------------------------------------
+.. versionadded:: 3.9.0
+
+.. py:class:: PKCS10Req()
+
+   .. py:method:: sign(issuing_cert, issuing_key, rng, not_before, not_after, hash_fn, padding)
```
randombit: padding should default to `None` most of the time it's not needed

doc/api_ref/python.rst
```diff
@@ -715,11 +715,145 @@ HOTP
...
+      ``not_before`` and ``not_after`` are expected to be the time since the UNIX epoch, in seconds.
+
+   .. py:method:: to_pem()
+
 X509Cert
 -----------------------------------------
 
 .. py:class:: X509Cert(filename=None, buf=None)
 
+   .. py:classmethod:: create_self_signed(key, opts, hash_fn, sig_padding, rng)
```
randombit: padding should be at the end and default to `None` since it's mostly not needed

randombit:
Can you try rebasing onto #4996? I think the implementation within FFI becomes a bit simpler since there are fewer error cases to worry about, and usage become more flexible (ie it becomes possible to set two commonName fields, which is not necessarily a common thing, but certainly allowed by the relevant specs). [Also if there are any issues that make using that new type in FFI difficult I'd like to deal with them prior to merging]

doc/api_ref/ffi.rst
arckoor: ~~Do I completely omit it or still expose it for pkcs10?~~ I'll keep it for now

doc/api_ref/python.rst
arckoor: ~~Do I default it to `None` or to `""`? Passing None means I have to do weird distinction between passing a nullptr and passing an actual value, i.e. I can't just `_ctype_str(value)`, I have do `None if value is None else _ctype_str(value)`. Similar for rust. I'd also have to accept `nullptr`s on the ffi side, and construct the `std::optional` there~~
It pays to look at what `_ctype_str` actually does, it does catch it. For rust the point still stands, but I'll try it like this for now.

arckoor:
That should be it, I hope I didn't miss anything. I'm not entirely convinced by the allowed `nullptr`s for `hash_fn`, `padding` and `challenge_password` yet though. But I like the new builder.
More functions for the PKCS#10 req I have noted down, I'll try to get that addressed as soon as I can.

randombit:
@arckoor Thanks. I'll review soon. I think it's likely this is not _quite_ going to make it in time for the 3.9 release, largely because exactly how the builder works is still somewhat up in the air. There is some chance reneme and I get that worked out very quickly (like in the next few days) which would unblock this PR from merging, but tbh I would not count on it. Lacking that, likely merging sometime ~~ late August.

src/lib/ffi/ffi.h
```diff
@@ -2408,6 +2428,42 @@ BOTAN_FFI_EXPORT(2, 13) int botan_x509_crl_load_file(botan_x509_crl_t* crl_obj,
...
+                          botan_rng_t rng,
+                          botan_x509_cert_t ca_cert,
+                          botan_privkey_t ca_key,
+                          uint64_t issue_time,
+                          uint32_t next_update,
+                          const botan_x509_cert_t* revoked,
+                          size_t revoked_len,
+                          uint8_t reason,
+                          const char* hash_fn,
+                          const char* padding);
```
arckoor: This is a little ridiculous imo, a CA type, whatever it's called would be appropriate here I think

src/lib/ffi/ffi_cert.cpp
```diff
@@ -802,37 +997,126 @@ int botan_x509_pkcs10_req_view_pem(botan_x509_pkcs10_req_t req, botan_view_ctx c
...
+         } else {
+            *limit = 32;
+         }
+      } else {
+         *is_ca = 0;
+         *limit = 0;
       }
-      if(padding != nullptr) {
-         padding_ = padding;
+      return BOTAN_FFI_SUCCESS;
```
arckoor: For the cert `is_ca` always returns 32 if it's not present, but for the request it's actually a `std::optional<uint8_t>`, so I'm not sure what the best behaviour is here

src/python/botan3.py
```diff
@@ -2348,22 +2433,92 @@ def is_revoked(self, crl):
...
+            ca_key.handle_(),
+            issue_time,
+            next_update,
+            arr_revoked,
+            revoked_len,
+            X509CRLReason.to_bits(reason),
+            _ctype_str(hash_fn),
+            _ctype_str(padding)
+        )
+        return crl
```
arckoor: I really don't love that the python binding has 0 type hints. You just have to know this method does not modify in place, while a `def revoke(...) -> X509CRL` would make that very clear.
If you're interested, I would be up for finally getting some type hints into the binding.

arckoor:
I'll wait until #5095 is merged, and then I'm probably going to pick this back up in a bunch of smaller PRs (CRLs, more getters for the Certs, and finally creation)
I think right now this thing is just way to big to reasonably review and merge

Issue #4697, opened 2025-02-18
finish, finish_msg functions and the secure_vector


IRONALEKS:
I was wondering, is there a particular reason why for example the CBC mode's finish function takes in a secure_vector, and not for example a span or pointer ?
As i understand, the finish function fills the vector with the final result(the encrypted data) so there is no reason for it to be a secure_array as far as understand. And CBC doesn't resize the result vector i think.
Wouldn't it be easier and more performant for user to be able to pass a span or a pointer to "finish" as not to copy for example from secure_vector to say a stream every time.
I guess my question is, is it secure_vector for security reasons or just as a convention and could be updated to use other options(span, pointer, etc.) ?

reneme:
CBC mode uses a padding if the input message isn't divisible by the underlying cipher's block size. So it does have to reallocate the vector in `finish`, potentially. There are other modes that don't, like CTR. But for consistency also those stick with the vector in `finish`.

However, if you want to avoid copies, you could use the `update()` function, which has a span overload, with block-aligned chunks of data and call `finish` with a vector of merely the remaining last few bytes.

PR #4880, opened 2025-05-21
Refactor: Cipher_Mode::finish_msg() using std::span

reneme:
_... this is work in progress and will currently not compile because I didn't refactor all subclasses of `Cipher_Mode`, yet. Early feedback would be very much appreciated._

I'm aiming at removing the `std::vector` dependency from `Cipher_Mode::finish_msg()` to avoid unneccessary copies or allocations. Another goal is to factor out the `offset` parameter handling from the concrete implementations into the `Cipher_Mode` base class. Note that this is not meant to be public-facing API just yet. I'm focussing on the private customization point in this PR.

The idea is that the new (probably eventually public) method `Cipher_Mode::bytes_needed_for_finalization()` provides an exact number of bytes that it will need to finalize a cipher mode. With that, the base class (or eventually the application?) allocates enough memory and passes it into `finish_msg()` as a span. `finish_msg()` then finalizes the cipher operation in-place and returns the number of bytes of the span that are actually relevant and part of the result. With this, the base class can strip authentication tags from decryption results for instance.

Personally, I find this API functional enough for internal use but fairly involved for a public interface. I'm open for suggestions here. 🙂 

... as stated, this is a follow up to #4873.


src/lib/modes/cipher_mode.h
```diff
@@ -204,6 +223,8 @@ class BOTAN_PUBLIC_API(2, 0) Cipher_Mode : public SymmetricAlgorithm {
       */
       virtual size_t output_length(size_t input_length) const = 0;
 
+      virtual size_t bytes_needed_for_finalization(size_t final_input_length) const = 0;
```
reneme, line 226: Documentation.

src/lib/modes/cipher_mode.cpp
```diff
@@ -5,6 +5,7 @@
 * Botan is released under the Simplified BSD License (see license.txt)
 */
 
+#include "botan/secmem.h"
```
reneme, line 8: Fix include

reneme:
This is quite a hefty change, because of significant adaptions in all `Cipher_Mode` finalization implementations. Though, I think, the modernization is worthwhile. It does save allocations and copies in several places. Also, I find, the use of span makes for easier-to-understand code. There's more room for improvement in some implementations once `process_msg` also learns to deal with span, for sure.

Unfortunately, I don't see an obvious way to split this up into smaller patches. I'll do a self-review after walking away from it for a few days, though I'd really appreciate more eyes on this.

randombit:
You may have to rebase for #4776

reneme:
> You may have to rebase for https://github.com/randombit/botan/pull/4776

Done.

src/lib/modes/aead/ccm/ccm.cpp
```diff
@@ -196,45 +201,56 @@ void CCM_Encryption::finish_msg(secure_vector<uint8_t>& buffer, size_t offset) {
...
+   const auto tag_length = tag_size();
+   const auto& buffered = msg_buf();
 
-   const size_t sz = buffer.size() - offset;
-   uint8_t* buf = buffer.data() + offset;
+   BOTAN_ASSERT_NOMSG(buffered.size() + input_bytes == buffer.size());
 
-   BOTAN_ARG_CHECK(sz >= tag_size(), "input did not include the tag");
+   const auto entire_payload = buffer.first(buffer.size() - tag_length);
+   const auto tag = buffer.last(tag_length);
```
FAlbertDev, line 246: Do we want an additional assert that buffer.size() >= tag_length?

src/lib/modes/aead/ccm/ccm.cpp
```diff
@@ -196,45 +201,56 @@ void CCM_Encryption::finish_msg(secure_vector<uint8_t>& buffer, size_t offset) {
...
    BOTAN_ARG_CHECK(ad.size() % CCM_BS == 0, "AD is block size multiple");
 
    const BlockCipher& E = cipher();
 
    secure_vector<uint8_t> T(CCM_BS);
-   E.encrypt(format_b0(sz - tag_size()), T);
+   E.encrypt(format_b0(entire_payload.size()), T);
 
    for(size_t i = 0; i != ad.size(); i += CCM_BS) {
       xor_buf(T.data(), &ad[i], CCM_BS);
```
FAlbertDev: Maybe also add a BufferSlicer for ad.

src/lib/modes/xts/xts.cpp
```diff
@@ -14,6 +14,22 @@
...
+constexpr void swap_bytes_via_xor(std::span<uint8_t> a, std::span<uint8_t> b) {
+   BOTAN_DEBUG_ASSERT(a.size() == b.size());
+
+   // TODO: use std::views::zip once we shed some older compilers, like:
+   //       for(auto& [front, back] : std::views::zip(a, b))
+   for(auto front = a.begin(), back = b.begin(); front != a.end() && back != b.end(); ++front, ++back) {
+      *front ^= *back;
+      *back ^= *front;
+      *front ^= *back;
+   }
```
FAlbertDev, line 28: Why not in one go? Or is this slower because of caching?
```suggestion
   xor_buf(a, b);
   xor_buf(b, a);
   xor_buf(a, b);
```

src/lib/modes/cipher_mode.h
```diff
@@ -76,10 +76,35 @@ class BOTAN_PUBLIC_API(2, 0) Cipher_Mode : public SymmetricAlgorithm {
...
+      *
+      * @param final_input_length  the number of payload bytes for finalization
+      * @returns  the exact byte size required as in/out buffer of finish_msg()
+      *           (including the @p final_input_length)
+      */
+      virtual size_t bytes_needed_for_finalization(size_t final_input_length) const = 0;
+
+      /**
       * Finishes a message
+      * @param final_block_and_more must be large enough to accomodate any
```
FAlbertDev:
```suggestion
      * @param final_block_and_more must be large enough to accommodate any
```

src/lib/modes/aead/ccm/ccm.cpp
reneme: Yes, thanks.

src/lib/modes/aead/ccm/ccm.cpp
reneme: Yeah, why not. :)

src/lib/modes/xts/xts.cpp
reneme: I frankly don't know. I just adapted the logic from what it was before. If I'm not mistaken we could also just `std::swap(a, b)`, no?

FAlbertDev: Does swap work on spans, without only swapping the spans themself but the data?

reneme: Nope, this is just swapping the spans, not the data within it. https://godbolt.org/z/xW5oE8Eq8 That's fairly logical, because it wouldn't be able to deal with spans of different sizes for instance. One would have to create a loop over the elements and swap them individually, I guess.

FAlbertDev: I have no strong opinion on std::swap vs. xor_buf vs. the original code. You may close this thread anytime :+1:

src/lib/tls/tls12/tls_null/tls_null.cpp
```diff
@@ -101,30 +101,34 @@ size_t TLS_NULL_HMAC_AEAD_Encryption::output_length(size_t input_length) const {
...
-                   "TLS_NULL_HMAC_AEAD_Decryption needs at least tag_size() bytes in final buffer");
+size_t TLS_NULL_HMAC_AEAD_Decryption::finish_msg(std::span<uint8_t> buffer, size_t input_bytes) {
+   const auto tag_length = tag_size();
+   BOTAN_ASSERT_NOMSG(buffer.size() == input_bytes);
+   BOTAN_ASSERT_NOMSG(buffer.size() >= tag_length);
 
-   const auto data_and_tag = std::span{buffer}.subspan(offset);
-   const auto data = data_and_tag.first(data_and_tag.size() - tag_size());
-   const auto tag = data_and_tag.subspan(data.size());
+   const auto remaining = buffer.first(buffer.size() - tag_size());
```
reneme:
```suggestion
   const auto remaining = buffer.first(buffer.size() - tag_length);
```

src/lib/tls/tls12/tls_cbc/tls_cbc.cpp
```diff
@@ -148,79 +148,83 @@ void TLS_CBC_HMAC_AEAD_Encryption::set_associated_data_n(size_t idx, std::span<c
...
-   buffer.resize(buffer.size() + padding_length);
+   BOTAN_ASSERT_NOMSG(buffer.size() >= padding_length);
+   BOTAN_DEBUG_ASSERT(buffer.size() % block_size() == 0);
 
    const uint8_t padding_val = static_cast<uint8_t>(padding_length - 1);
 
-   CT::poison(&padding_val, 1);
-   CT::poison(&padding_length, 1);
-   CT::poison(buffer.data(), buffer.size());
+   const auto last_block = buffer.last(block_size());
```
reneme: Perhaps cache the block size into a stack variable (for consistency with plenty other algorithms).

src/lib/tls/tls12/tls_cbc/tls_cbc.cpp
```diff
@@ -262,15 +266,17 @@ uint16_t check_tls_cbc_padding(const uint8_t record[], size_t record_len) {
...
 }
 
-void TLS_CBC_HMAC_AEAD_Decryption::cbc_decrypt_record(uint8_t record_contents[], size_t record_len) {
-   if(record_len == 0 || record_len % block_size() != 0) {
+void TLS_CBC_HMAC_AEAD_Decryption::cbc_decrypt_record(std::span<uint8_t> record) {
+   if(record.empty() || record.size() % block_size() != 0) {
       throw Decoding_Error("Received TLS CBC ciphertext with invalid length");
    }
 
+   const auto final_block = record.last(block_size());
```
reneme: Cache the block size into a stack variable?

src/lib/tls/tls12/tls_cbc/tls_cbc.cpp
```diff
@@ -355,62 +366,63 @@ void TLS_CBC_HMAC_AEAD_Decryption::perform_additional_compressions(size_t plen,
...
    // This early exit does not leak info because all the values compared are public
-   if(record_len < tag_size() || (record_len - (use_encrypt_then_mac() ? tag_size() : 0)) % block_size() != 0) {
+   if(record.size() < tag_size() || (record.size() - (use_encrypt_then_mac() ? tag_size() : 0)) % block_size() != 0) {
       throw TLS_Exception(Alert::BadRecordMac, "Message authentication failure");
    }
 
    if(use_encrypt_then_mac()) {
-      const size_t enc_size = record_len - tag_size();
-      const size_t enc_iv_size = enc_size + iv_size();
+      const auto payload_and_padding = record.first(record.size() - tag_size());
```
reneme: Cache `tag_length` as stack variable? (also block size maybe)

src/lib/modes/cbc/cbc.cpp
```diff
@@ -145,46 +153,40 @@ size_t CTS_Encryption::output_length(size_t input_length) const {
...
-
-      buffer += last;
+      const size_t full_blocks = ((buffer.size() / BS) - 1) * BS;
+      const auto tail = buffer.subspan(full_blocks);
+      BOTAN_ASSERT(tail.size() > BS && tail.size() < 2 * BS, "Left over size in expected range");
+
+      const auto final_ordinary_blocks = buffer.first(full_blocks);
+      const auto first_block_of_tail = tail.first(BS);
+      const auto final_bytes_of_tail = tail.subspan(BS);
+      const auto first_bytes_of_tail = tail.first(tail.size() - BS);
```
reneme:
```suggestion
      const auto first_bytes_of_tail = tail.first(final_bytes_of_tail.size());
```

src/lib/modes/cbc/cbc.cpp
```diff
@@ -254,48 +256,40 @@ size_t CTS_Decryption::minimum_final_size() const {
...
```
reneme:
```suggestion
      const auto first_bytes_of_tail = tail.first(final_bytes_of_tail.size());
```

src/lib/modes/aead/siv/siv.cpp
```diff
@@ -130,70 +130,74 @@ secure_vector<uint8_t> SIV_Mode::S2V(const uint8_t* text, size_t text_len) {
       V ^= m_nonce;
    }
 
-   if(text_len < block_size()) {
+   if(text.size() < block_size()) {
```
reneme: Cache `BS` as stack variable?

src/lib/modes/aead/siv/siv.cpp
```diff
@@ -115,10 +115,10 @@ size_t SIV_Mode::process_msg(uint8_t buf[], size_t sz) {
    return 0;
 }
 
-secure_vector<uint8_t> SIV_Mode::S2V(const uint8_t* text, size_t text_len) {
+std::array<uint8_t, SIV_Mode::tag_length> SIV_Mode::S2V(std::span<const uint8_t> text) {
    const std::vector<uint8_t> zeros(block_size());
 
-   secure_vector<uint8_t> V = m_mac->process(zeros.data(), zeros.size());
+   secure_vector<uint8_t> V = m_mac->process(zeros);
```
reneme, line 123:
```suggestion
   secure_vector<uint8_t> V = m_mac->process(zeros);
   BOTAN_DEBUG_ASSERT(V.size() == block_size());
```

src/lib/modes/aead/siv/siv.cpp
```diff
@@ -130,70 +130,74 @@ secure_vector<uint8_t> SIV_Mode::S2V(const uint8_t* text, size_t text_len) {
...
+   auto& buffered = msg_buf();
+   BOTAN_ASSERT_NOMSG(buffered.size() + input_bytes == buffer.size());
 
-   if(!msg_buf().empty()) {
-      buffer.insert(buffer.begin() + offset, msg_buf().begin(), msg_buf().end());
-      msg_buf().clear();
+   if(!buffered.empty()) {
+      copy_mem(buffer.last(input_bytes), buffer.first(input_bytes));
+      copy_mem(buffer.first(buffered.size()), buffered);
+      buffered.clear();
```
reneme, line 186: Double-check: Do we need to clear the `buffered` buffer in the encryption as well?

reneme (commented):
Self-review: A few notes to myself for some cleanups that I'd still like to address. Otherwise, I think that is ready for prime time. :)

src/lib/modes/aead/siv/siv.cpp
reneme: No, this happens during `start_msg()`

reneme:
Ready for review, from my side. @randombit

randombit:
I took a look already some time ago and this seemed fine in the outline but I need to do a full review. Next in my queue.

Initial comment, please rebase to pick up the recent clang-tidy CI changes

reneme:
> Initial comment, please rebase to pick up the recent clang-tidy CI changes

Done. One minor conflict in the SIV header. No clang-tidy hiccup. 😎

src/lib/modes/cipher_mode.h
```diff
@@ -176,24 +201,30 @@ class BOTAN_PUBLIC_API(2, 0) Cipher_Mode : public SymmetricAlgorithm {
...
-      * @param offset an offset into final_block to begin processing
-      */
       template <concepts::resizable_byte_buffer T>
       void finish(T& final_block, size_t offset = 0) {
-         Botan::secure_vector<uint8_t> tmp(final_block.begin(), final_block.end());
-         finish_msg(tmp, offset);
-         final_block.resize(tmp.size());
-         std::copy(tmp.begin(), tmp.end(), final_block.begin());
+         if(offset > final_block.size()) {
+            throw Invalid_Argument("invalid offset");
```
randombit: The change here from `BOTAN_ARG_CHECK` to `throw Invalid_Argument` causes quite a loss of context since the macro includes the function name, line number, etc.

These checks don't depend on `T` - maybe move them to a private helper in `cipher_mode.cpp` which then can formats the messages nicely, include `name()` in the message as a hint to the user [the replace the loss of the filename/function name], etc.

randombit:
> Personally, I find this API functional enough for internal use but fairly involved for a public interface.

Yeah the interface for cipher modes is really fiddly. Was designed trying to balance usability vs minimizing copies, I'm not sure it really ended up in the right place. But I also have not come up with anything better so far. :/

randombit:
In general the change looks ok. One thing I'm concerned about is that this seems to be adding quite a bit of overhead - though tbh right now I do not see where.

For example on master

CTR-BE(AES-128) encrypt buffer size 8 bytes: 685.081 MiB/sec 4.17 cycles/byte (342.56 MiB in 500.03 ms)

this branch

CTR-BE(AES-128) encrypt buffer size 8 bytes: 544.767 MiB/sec 5.24 cycles/byte (272.44 MiB in 500.10 ms)

In absolute terms the overhead here is probably quite small, but in cases where the message is small (so finish cost dominates) and the cipher is fast (as with AES-NI) it's quite noticeable. I see it also in other ciphers (eg ~7% regression for AES-128-GCM with 8 byte messages)

reneme:
> In absolute terms the overhead here is probably quite small, but in cases where the message is small (so finish cost dominates) and the cipher is fast (as with AES-NI) it's quite noticeable. 

Good point to measure it like that. That's surprising to me as well. I'll look into it, ~~though it may have to wait a bit, unfortunately~~. 

Edit: I couldn't wait to take a quick glance after all. Especially the CTR mode becoming slower is very unexpected. The finalization for stream ciphers is essentially a NOOP. Also, in the `Stream_Mode` base class I moved from calling `update()`, to `process()` which actually removes an indirection (albeit one that was likely inlined anyway), and with that a call to `std::vector::resize()` (that would always be a NOOP, as well).

The only thing I could immediately blame this on is the new virtual method call (`bytes_needed_for_finalization()`) in the generic implementation of `finish()`, that is useless for stream ciphers but can't be avoided in our class hierarchy.

A quick test on my M2 MacBook Air over breakfast coffee does not reproduce this finding on my machine, though:

```
➜  botan git:(master) ✗ ./botan speed --msec=2000 --buf-size=8 "CTR-BE(AES-128)"
CTR-BE(AES-128) encrypt buffer size 8 bytes: 551.325 MiB/sec (1102.688 MiB in 2000.069 ms)

➜  botan git:(refactor/span-for-cipher-mode-finalization) ✗ ./botan speed --msec=2000 --buf-size=8 "CTR-BE(AES-128)"
CTR-BE(AES-128) encrypt buffer size 8 bytes: 551.307 MiB/sec (1102.625 MiB in 2000.021 ms)
```

... multiple runs jitter between roughly 548MiB/s and 560MiB/s in both cases.

randombit:
It's very reliable for me on a Tiger Lake laptop. There is some noise to be sure (~5% back and forth depending on the run) but this is more like a 25%

master (Clang 20):
CTR-BE(AES-128) encrypt buffer size 8 bytes: 687.218 MiB/sec 4.16 cycles/byte (343.62 MiB in 500.02 ms)
CTR-BE(AES-128) decrypt buffer size 8 bytes: 595.302 MiB/sec 4.80 cycles/byte (297.69 MiB in 500.06 ms)
CTR-BE(AES-128) encrypt buffer size 16 bytes: 1149.485 MiB/sec 2.48 cycles/byte (574.75 MiB in 500.01 ms)
CTR-BE(AES-128) decrypt buffer size 16 bytes: 1050.926 MiB/sec 2.72 cycles/byte (525.50 MiB in 500.04 ms)

here (Clang 20):
CTR-BE(AES-128) encrypt buffer size 8 bytes: 553.520 MiB/sec 5.16 cycles/byte (276.81 MiB in 500.10 ms)
CTR-BE(AES-128) decrypt buffer size 8 bytes: 479.829 MiB/sec 5.95 cycles/byte (239.94 MiB in 500.05 ms)
CTR-BE(AES-128) encrypt buffer size 16 bytes: 980.258 MiB/sec 2.91 cycles/byte (490.19 MiB in 500.06 ms)
CTR-BE(AES-128) decrypt buffer size 16 bytes: 863.294 MiB/sec 3.31 cycles/byte (431.69 MiB in 500.05 ms)

master (GCC 15):
CTR-BE(AES-128) encrypt buffer size 8 bytes: 843.907 MiB/sec 3.38 cycles/byte (422.00 MiB in 500.06 ms)
CTR-BE(AES-128) decrypt buffer size 8 bytes: 695.764 MiB/sec 4.11 cycles/byte (347.94 MiB in 500.08 ms)
CTR-BE(AES-128) encrypt buffer size 16 bytes: 1557.379 MiB/sec 1.83 cycles/byte (778.75 MiB in 500.04 ms)
CTR-BE(AES-128) decrypt buffer size 16 bytes: 1306.543 MiB/sec 2.19 cycles/byte (653.31 MiB in 500.03 ms)

here (GCC 15):
CTR-BE(AES-128) encrypt buffer size 8 bytes: 710.878 MiB/sec 4.02 cycles/byte (355.50 MiB in 500.09 ms)
CTR-BE(AES-128) decrypt buffer size 8 bytes: 597.441 MiB/sec 4.78 cycles/byte (298.75 MiB in 500.05 ms)
CTR-BE(AES-128) encrypt buffer size 16 bytes: 1341.603 MiB/sec 2.13 cycles/byte (670.81 MiB in 500.01 ms)
CTR-BE(AES-128) decrypt buffer size 16 bytes: 1138.683 MiB/sec 2.51 cycles/byte (569.38 MiB in 500.03 ms)

I'll check on another machine

randombit:
Did some experimenting with disabling different parts of `finish` to see what the overhead is. The issue is definitely with the virtual calls. Basically I'm seeing a ~~ 25% perf drop (in this one specific small message/fast cipher case) and if I comment out

```
           if(final_input_bytes < minimum_final_size()) {
              throw Invalid_Argument("final input message is too small");
           }
```

I get ~12% back.

And if I replace 

```
           const auto final_buffer_bytes = bytes_needed_for_finalization(final_input_bytes);

           // Make room for additional overhead to be produced during finalization
           if(final_input_bytes < final_buffer_bytes) {
              final_block.resize(offset + final_buffer_bytes);
           }
```

with

```
const size_t final_buffer_bytes = final_input_bytes;
```

(only valid for stream modes obviously)

I get back the other ~12%

So it seems the virtual calls are more costly on this laptop that on your machine. Or possibly the M2's AES hardware is not fast enough that the overhead becomes visible? IDK

So one immediate suggestion that seems honestly an improvement *anyway*, regardless of the overhead issue:

Each implementation of `bytes_needed_for_finalization` should itself check if the input value is less than `minimum_final_size` and if so throw. Then 2 virtual calls becomes 1 (since `bytes_needed_for_finalization` occurs in `final` leaf classes and so the compiler can devirtualize the `minimum_final_size` call)

src/lib/modes/cipher_mode.h
reneme: Fair point. Done.

reneme:
> Each implementation of `bytes_needed_for_finalization` should itself check if the input value is less than `minimum_final_size` and if so throw. Then 2 virtual calls becomes 1 (since `bytes_needed_for_finalization` occurs in final leaf classes and so the compiler can devirtualize the `minimum_final_size` call)

Very much a fair point!

I did what you suggested (in this commit: https://github.com/randombit/botan/pull/4880/commits/79ae69b5e97f84d5fedac9c5a4a590f8b0e33fb4) and it turns out that most implementations of `bytes_needed_for_finalization` didn't even need to be adapted. Either because `minimum_final_size` was 'zero' anyway or because they already contained the check. The new function `calculate_final_input_bytes` (implemented in the *.cpp) now takes care of both the offset check (mentioned [here](https://github.com/randombit/botan/pull/4880#discussion_r2170118887)) and also a debug-assert as a safety net for the implementation specific check in `bytes_needed_for_finalization`.

Interestingly, the debug-assert forced me to adapt a few negative tests that were meant to trigger an `Invalid_State` exception but ran into the `Invalid_Argument` instead. 🙄 I pulled these test adaptions into [an independent pull request](https://github.com/randombit/botan/pull/4970) and [rebased this PR](https://github.com/randombit/botan/pull/4880#issuecomment-3065482944) after merging it.

As a general idea for consideration: The concrete implementations (for the modes of operation but also other primitives) are anyway hidden in internal headers. So, in theory, we should be able to refactor them using 'static polymorphy' behind the public base class interfaces (similar to pcurves). That way, many algorithm-specific traits (block size, tag length, intrinsic parallelism, ...) could become compile-time constants and thus cheap to access. One day, perhaps...

reneme:
Rebased after merging #4970. I'm happy to report that the stricter exception expectations did uncover another issue in commoncrypto that could have potentially caused actual harm!

Due to a missing "key is set" check in `CommonCrypto_Cipher_Mode::bytes_needed_for_finalization` a call to `finish()` on an unkeyed object called `CCCryptorGetOutputLength` with a `nullptr` handle. This resulted in an unreasonably large "expected output length" (probably some UB). The generic `Cipher_Mode::finish()` implementation then tried to allocate a huge vector causing some STL exception to be thrown.

Before #4970 this STL exception was 'good enough', now it was flagged as unexpected. 😎

randombit:
> As a general idea for consideration: The concrete implementations (for the modes of operation but also other primitives) are anyway hidden in internal headers. So, in theory, we should be able to refactor them using 'static polymorphy' behind the public base class interfaces (similar to pcurves). That way, many algorithm-specific traits (block size, tag length, intrinsic parallelism, ...) could become compile-time constants and thus cheap to access. 

Another (completely orthogonal) idea I had while thinking about this PR is that `Cipher_Mode` was ill-designed and it would have been much better to directly express encryption vs decryption modes in the API, rather than covering both operations with the same API. Since only encryption ever expands, while decryption contracts, likely we could offer dedicated APIs that are simpler for each respective usage.

This is still fixable, since happily all of the cipher implementations are themselves private. And at least unlike a refactoring of an existing API like here, it can be done incrementally one mode at a time. But it should probably happen as part of a larger rethink of how cipher operations are expressed (see mistakes.rst)

reneme:
> ...  it would have been much better to directly express encryption vs decryption modes in the API, rather than covering both operations with the same API. Since only encryption ever expands, while decryption contracts, likely we could offer dedicated APIs that are simpler for each respective usage.

Yes, full ack. I spent the morning trying to come up with a decent public API for a new `Cipher_Mode::finish` and got lost in the details. From what I saw, those are the complexities it would have to handle (essentially all at once):

* Encryption
  * Addition of the tag (usually at the end, but sometimes in the front, e.g. SIV)
  * Addition of a padding, or ciphertext-stealing-style finalizations
  * Sometimes both, e.g. CBC+MAC respectively MAC+CBC in TLS 1.2
* Decryption
  * Reading/Verifying the tag, then removing it
  * Reading/Verifying paddings or CTS's, then maybe removing it
* In general, usually mixed with the above
  * Buffered modes producing the entire ciphertext at the end, even when using `process()` or `update()`, along with padding/tag addition/removal

At this point, I would probably just expose an equivalent version of `finish_msg()` for users that want to avoid the allocations/copies (e.g. https://github.com/randombit/botan/issues/4697) that may come with the existing `std::vector`-based version. It's not very ergonomic but I doubt its going to get much better anyway. And for most people, the existing API is likely just fine for now.

reneme:
@randombit from my side this is ready to go in. Or am I missing something?

PR #4882, opened 2025-05-22
Enable GCC 14's stack scrubbing for SHA-224 and SHA-256

reneme:
This introduces support for [GCC 14's stack scrubbing](https://gcc.gnu.org/onlinedocs/gcc-14.2.0/gcc/Common-Type-Attributes.html#index-strub-type-attribute) in the SHA-224/256 implementation. At the moment, that is an opt-in feature even if your compiler is recent enough. Compile with `./configure.py --enable-stack-scrubbing` to enable it.

If enabled, the internal SHA-2 compression function implementations will be annotated with `__attribute__(strub("internal"))` for which GCC will emit code to zeroize the function's stack frame after it returned.

We are currently building with GCC 13 (the default in GHA's ubuntu 24.04 image), which doesn't support strubbing, yet. Hence, I extended the CI setup to selectively install and use GCC 14 when `compiler: gcc-14` is set in `ci.yml` (and its friends).

# Technical Details

Using a debugger, it is fairly easy to verify that the scrubbing of the annotated function actually happens. Below, `compress_digest` appears twice in the call stack. The lower one is generated by GCC and acts as a wrapper for the actual compression implementation. Once the upper one returns, the wrapped stack frame is cleared as seen in the assembly below.

image

After the implementation returned (the `call` instruction), there's a loop over the left-over stack frame (see `rax` and `rdx`) and `mov QWORD PTR [rax],0x0`.

image

### Previous (draft status) description

Coming back to this: https://github.com/randombit/botan/pull/4826#issuecomment-2799878635, I saw that GCC 14+ supports [stack scrubbing](https://gcc.gnu.org/onlinedocs/gcc-14.2.0/gcc/Common-Type-Attributes.html#index-strub-type-attribute) based on function annotations and I wanted to try it out for our (potential) use case. I'm using SHA-256 as a random illustration of the idea.

Here's my take of how it works: Adding `__attribute__ (strub("at-calls"))` to a function declaration tells every _caller_ to clean up the function's stack frame after it returned. For that, GCC alters the annotated function's signature with a hidden "watermark" variable that communicates the length of its stack frame back to the caller. I believe it does that to support and coalesce transitive cleanups of multiple stack frames. To me, it seems a bit cumbersome that the caller has to be aware of this. That makes it feel somewhat unsuitable to use across module boundaries or even public API. But I'm sure there's a good technical reason for this.

Anyway, Botan wraps an abstract interface around most fundamental algorithms. So handling this inconvenience within individual compilation units and applying this to most algorithms' core routine does seem feasible. Additionally inlining seems to help here, because it combines inlined helper functions into a single stack frame.

Naturally, it seems challenging to (automatically) verify that the scrubbing actually does what it should within a sensible scope. But that as well should be tangible within most algorithms' core routines, assuming the approach itself works reliably.

Alternatively (and perhaps more portable), [some explicit RAII helper](https://godbolt.org/z/jM834qTcz) around stack variables within the core routines could also do the trick. Though, I must say, I actually like the idea of just annotating (internal) functions as "sensitive" and let the compiler figure it out.

@randombit This is really just a test setup to have something to play with and discuss utility of it.

### DIFF
--- a/src/lib/hash/sha2_32/sha2_32.cpp
+++ b/src/lib/hash/sha2_32/sha2_32.cpp
@@ -12,6 +12,7 @@
 #include <botan/internal/loadstor.h>
 #include <botan/internal/rotate.h>
 #include <botan/internal/sha2_32_f.h>
+#include <botan/internal/stack_scrubbing.h>
 #include <botan/internal/stl_util.h>
 
 #if defined(BOTAN_HAS_CPUID)
@@ -55,7 +56,9 @@ std::string sha256_provider() {
 /*
 * SHA-224 / SHA-256 compression function
 */
-void SHA_256::compress_digest(digest_type& digest, std::span<const uint8_t> input, size_t blocks) {
+void BOTAN_SCRUB_STACK_AFTER_RETURN SHA_256::compress_digest(digest_type& digest,
+                                                             std::span<const uint8_t> input,
+                                                             size_t blocks) {
 #if defined(BOTAN_HAS_SHA2_32_X86)
    if(CPUID::has(CPUID::Feature::SHA)) {
       return SHA_256::compress_digest_x86(digest, input, blocks);
--- a/.github/actions/setup-build-agent/action.yml
+++ b/.github/actions/setup-build-agent/action.yml
@@ -11,6 +11,10 @@ inputs:
   target:
     description: The ci_build.py target going to be built on this agent
     required: true
+  compiler:
+    description: The compiler used to build this target
+    required: false
+    default: unknown
   cache-key:
     description: The actions/cache key to be used for this runs, caching will be disabled when no key is provided
     required: false
@@ -23,11 +27,11 @@ runs:
   using: composite
   steps:
     - name: Setup Build Agent (Windows)
-      run: ${{ github.action_path }}/../../../src/scripts/ci/setup_gh_actions.ps1 "${{ inputs.target }}" "${{ inputs.arch }}"
+      run: ${{ github.action_path }}/../../../src/scripts/ci/setup_gh_actions.ps1 "${{ inputs.target }}" "${{ inputs.compiler }}" "${{ inputs.arch }}"
       shell: pwsh
       if: runner.os == 'Windows'
     - name: Setup Build Agent (Unix-like)
-      run: ${{ github.action_path }}/../../../src/scripts/ci/setup_gh_actions.sh "${{ inputs.target }}" "${{ inputs.arch }}"
+      run: ${{ github.action_path }}/../../../src/scripts/ci/setup_gh_actions.sh "${{ inputs.target }}" "${{ inputs.compiler }}" "${{ inputs.arch }}"
       shell: bash
       if: runner.os != 'Windows'
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,7 @@ jobs:
         uses: ./.github/actions/setup-build-agent
         with:
           target: ${{ matrix.target }}
+          compiler: msvc
           cache-key: ${{ matrix.host_os }}-msvc-${{ matrix.arch }}-${{ matrix.target }}
           arch: ${{ matrix.arch }}
 
@@ -89,6 +90,7 @@ jobs:
         uses: ./.github/actions/setup-build-agent
         with:
           target: ${{ matrix.target }}
+          compiler: ${{ matrix.compiler }}
           cache-key: linux-${{ matrix.compiler }}-x86_64-${{ matrix.target }}
 
       - name: Build and Test Botan
@@ -126,6 +128,7 @@ jobs:
         uses: ./.github/actions/setup-build-agent
         with:
           target: ${{ matrix.target }}
+          compiler: ${{ matrix.compiler }}
           cache-key: macos-${{ matrix.compiler }}-${{ matrix.os }}-${{ matrix.target }}
 
       - name: Build and Test Botan
@@ -144,6 +147,7 @@ jobs:
         uses: ./.github/actions/setup-build-agent
         with:
           target: clang-tidy
+          compiler: clang
           cache-key: linux-x86_64-clang-tidy
 
       - name: Configure Build
@@ -212,6 +216,7 @@ jobs:
         uses: ./source/.github/actions/setup-build-agent
         with:
           target: ${{ matrix.target }}
+          compiler: ${{ matrix.compiler }}
           cache-key: ${{ matrix.host_os }}-${{ matrix.compiler }}-x86_64-${{ matrix.target }}
 
       - name: Build and Test Botan
@@ -260,6 +265,7 @@ jobs:
         uses: ./source/.github/actions/setup-build-agent
         with:
           target: ${{ matrix.target }}
+          compiler: ${{ matrix.compiler }}
           cache-key: ${{ matrix.host_os }}-${{ matrix.compiler }}-x86_64-${{ matrix.target }}
 
       - name: Build and Test Botan
@@ -319,6 +325,7 @@ jobs:
         uses: ./.github/actions/setup-build-agent
         with:
           target: ${{ matrix.target }}
+          compiler: ${{ matrix.compiler }}
           cache-key: ${{ matrix.host_os }}-${{ matrix.compiler }}-xcompile-${{ matrix.target }}
 
       - name: Build and Test Botan
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -58,6 +58,7 @@ jobs:
         uses: ./source/.github/actions/setup-build-agent
         with:
           target: ${{ matrix.target }}
+          compiler: ${{ matrix.compiler }}
           cache-key: ${{ matrix.host_os }}-${{ matrix.compiler }}-x86_64-${{ matrix.target }}
 
       - name: Build and Test Botan
@@ -119,6 +120,7 @@ jobs:
         uses: ./.github/actions/setup-build-agent
         with:
           target: ${{ matrix.target }}
+          compiler: ${{ matrix.compiler }}
           cache-key: ${{ matrix.host_os }}-${{ matrix.compiler }}-xcompile-${{ matrix.target }}
 
       - name: Build and Test Botan
@@ -139,6 +141,7 @@ jobs:
         uses: ./.github/actions/setup-build-agent
         with:
           target: clang-tidy
+          compiler: clang
           cache-key: linux-x86_64-clang-tidy
 
       - name: Install dependencies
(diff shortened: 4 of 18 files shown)


src/lib/hash/sha2_32/sha2_32.cpp
```diff
@@ -173,6 +146,39 @@ void SHA_256::compress_digest(digest_type& digest, std::span<const uint8_t> inpu
...
+   }
+#endif
+
+#if defined(BOTAN_HAS_SHA2_32_SIMD)
+   if(CPUID::has(CPUID::Feature::SIMD_4X32)) {
+      return SHA_256::compress_digest_x86_simd(digest, input, blocks);
+   }
+#endif
+
+   compress(digest, input, blocks);
```
reneme, line 179: `compress` is annotated for strubbing and GCC introduces a loop to `memset` the function's stack space to 0s.

Here's some disassembly:

```
0000000000002bc0 <Botan::SHA_256::compress_digest(std::vector<unsigned int, Botan::secure_allocator<unsigned int> >&, std::span<unsigned char const, 18446744073709551615ul>, unsigned long)>:
    2bc0:       f3 0f 1e fa             endbr64
    2bc4:       48 83 ec 18             sub    $0x18,%rsp
    2bc8:       4c 8d 44 24 08          lea    0x8(%rsp),%r8
    2bcd:       48 89 64 24 08          mov    %rsp,0x8(%rsp)
    2bd2:       e8 e9 d5 ff ff          call   1c0 compress(std::vector<unsigned int, Botan::secure_allocator<unsigned int> >&, std::span<unsigned char const, 18446744073709551615ul>, unsigned long)>
    2bd7:       48 8b 54 24 08          mov    0x8(%rsp),%rdx
    2bdc:       48 89 e0                mov    %rsp,%rax
    2bdf:       48 39 e2                cmp    %rsp,%rdx
    2be2:       73 10                   jae    2bf4 <Botan::SHA_256::compress_digest(std::vector<unsigned int, Botan::secure_allocator<unsigned int> >&, std::span<unsigned char const, 18446744073709551615ul>, unsigned long)+0x34>

    -- IIUC: this loops over compress()'s stack space and clears it
    2be4:       48 83 e8 08             sub    $0x8,%rax
    2be8:       48 c7 00 00 00 00 00    movq   $0x0,(%rax)
    2bef:       48 39 c2                cmp    %rax,%rdx
    2bf2:       72 f0                   jb     2be4 <Botan::SHA_256::compress_digest(std::vector<unsigned int, Botan::secure_allocator<unsigned int> >&, std::span<unsigned char const, 18446744073709551615ul>, unsigned long)+0x24>
    -- end loop

    2bf4:       48 83 c4 18             add    $0x18,%rsp
    2bf8:       c3                      ret
    2bf9:       f3 0f 1e fa             endbr64
    2bfd:       48 89 c7                mov    %rax,%rdi
    2c00:       e9 00 00 00 00          jmp    2c05 <Botan::SHA_256::compress_digest(std::vector<unsigned int, Botan::secure_allocator<unsigned int> >&, std::span<unsigned char const, 18446744073709551615ul>, unsigned long)+0x45>
    2c05:       90                      nop
    2c06:       66 2e 0f 1f 84 00 00    cs nopw 0x0(%rax,%rax,1)
    2c0d:       00 00 00 
```

randombit:
I was aware this existed but hadn't looked into it in much detail. I agree the `at-calls` approach seems awkward, especially in our context because using it practically seems to imply (as you did here) having to pull every bit of relevant logic out of the relevant class and into freestanding functions. It seems another mode `strub("internal")` is also available which *does not* modify the ABI of the function, and that might be more suitable for us. This has the limitation that (depending on circumstances) the arguments and return value might not be scrubbed. However I think practically this is not an issue since our typical usage is to pass pointers to bytes or words, plus a length field. There are plenty of functions that do take/return scalar arguments which might be sensitive (eg `rotr`) but these are almost all aggressively inlined.

> Alternatively (and perhaps more portable), some explicit RAII helper around stack variables

This approach seems to run up against an interesting footgun in the design (quotes from GCC manual)

> A function associated with ‘at-calls’ ‘strub’ mode (‘strub("at-calls")’, or just ‘strub’) undergoes interface changes.

> An automatically-allocated variable associated with the ‘strub’ attribute causes the (immediately) enclosing function to have ‘strub’ enabled.

ie if you declare a `strub` stack variable in your function, it changes the ABI :exploding_head: 

My current take

* Probably worth supporting
* But maybe not by default? At least not without an extensive look at the performance impact
* `strub("internal")` is probably more viable than `strub("at-calls")` or type annotations since these seem to practically require us to do a lot of refactoring. It seems to me much better to do something that works ~95% effectively that is easy to apply, than something that is 100% effective but requires extensive refactoring and thus only used in 5% of the relevant cases.

randombit:
>  At least not without an extensive look at the performance impact

Yikes. GCC 15.1 on a Tiger Lake core, compiling the entire tree with `-fstrub=internal`

Blowfish 447 MBs -> 259
SHA-256 (AXX2) 489 -> 96
Serpent (AVX512) 2169 -> 59 (!!!)
ECDSA P-256 sign 21214/s -> 7874/s
ECDSA P-256 verify 11229/s -> 4929/s

Possibly `-fstrub=internal` is massively interfering with inlining or something, especially the AVX512 regression is hard to explain otherwise.

As a spot check I tried adding `__attribute__((strub("internal"))` to Blowfish `encrypt_n` and `decrypt_n` and with that at least I don't see any notable change in performance.

reneme:
> It seems another mode `strub("internal")` is also available which _does not_ modify the ABI of the function, and that might be more suitable for us.

Agree. I didn't use that because it also somehow interferes with the "strict" mode and I didn't want to get into trouble with that. My understanding is that it essentially automagically creates an internal wrapper function like the free-standing `compress` I added for SHA-256. Let's try if we can just apply that to the definition of (say) `SHA_256::compress_digest` in the cpp file and leave the header unchanged. That would be really nice actually. (@randombit Is that what you did for your blowfish test?)

**Regarding performance:** I tried `./botan speed` with the patch in this PR and also didn't see a difference in throughput. Applying it to the entire code base certainly doesn't seem feasible at all. Probably this is essentially equivalent to `-O0` regarding inlining.

**Regarding annotating stack variables:** I think I would refrain from doing that anyway. It seems gimmicky and full of surprises to me. If I understood correctly, you can even screw up your ABI by just reading a global variable annotated like this. 😲 More so, [if you apply it to a stack variable that is declared as `auto`](https://godbolt.org/z/qfs4KsvPr) GCC ICE'es. (Perhaps worth a GCC bug ticket.)

**Regarding the RAII wrapper:** I meant to not use GCC's strubbing at all, but just rely on the RAII-destructor to call `scrub_memory` on whatever was passed in (similarly to `CT::scoped_poison()`). _I don't understand your point with the footgun, therefore._

randombit:
> Is that what you did for your blowfish test?

Yes I put the annotation directly on the member functions. It then did basically what is described in the manual, creating a wrapper function that zeros the stack from the inner function

```
Botan::Blowfish::encrypt_n(unsigned char const*, unsigned char*, unsigned long) const:
        subq    $24, %rsp       #,
# src/lib/block/blowfish/blowfish.cpp:156: __attribute__((strub("internal"))) void Blowfish::encrypt_n(const uint8_t in[], uint8_t out[], size_t blocks) const {
        movq    %fs:40, %rax    # MEM[(<address-space-1> long unsigned int *)40B], tmp104
        movq    %rax, 8(%rsp)   # tmp104, D.135729
        xorl    %eax, %eax      # tmp104
        movq    %rsp, %r8       #,
        movq    %rsp, (%rsp)    #, .strub.watermark.531
.LEHB4:
        call    Botan::Blowfish::encrypt_n(unsigned char const*, unsigned char*, unsigned long) const [clone .strub.0]  #
.LEHE4:
        movq    (%rsp), %rdx    # .strub.watermark.531, .strub.watermark.531
        movq    %rsp, %rax      #, tmp113
        cmpq    %rsp, %rdx      #, .strub.watermark.531
        jnb     .L139   #,
.L146:
        subq    $8, %rax        #, tmp116
        movq    $0, (%rax)      #,
        cmpq    %rax, %rdx      # tmp113, .strub.watermark.531
        jb      .L146   #,
...
```

> Regarding the RAII wrapper: I meant to not use GCC's strubbing at all, but just rely on the RAII-destructor to call scrub_memory on whatever was passed in (similarly to CT::scoped_poison()). I don't understand your point with the footgun, therefore.

Ah I misunderstood (your godbolt link doesn't go anywhere for me possibly you had a snippet there with an example?). I assumed you meant something like

```
template<typename T>
Strubbed {
   // getters and such
public:
   __attribute__((strub)) T m_val;
};

void foo() {
   Strubbed<int> z = 5;
}
```

where we would be somewhat implicitly using `strub` on variables without annotating the functions. Since it seems required *anyway* to annotate the functions in order to not change the ABI, to be honest I don't understand how or when it's possible to use `strub` annotated variables. What you were actually proposing of course would be much simpler to implement, and work for all compilers. However it would miss anything the compiler decides to push onto the stack.

> Perhaps worth a GCC bug ticket.

They would probably be interested, I can open a ticket I have an account on their bugzilla already

reneme:
> They would probably be interested, I can open a ticket I have an account on their bugzilla already

Yes please.

> your godbolt link doesn't go anywhere for me possibly you had a snippet there with an example?

Yeah, fixed now. Apparently I just copied the address line, not the short link. 🫤

> Yes I put the annotation directly on the member functions.

Fancy!

randombit:
Turns out that mixing `auto` and `__attribute__` has been broken in GCC since at least GCC 8.2 (WTF) https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89867

src/lib/hash/sha2_32/sha2_32.cpp
```diff
@@ -55,7 +56,9 @@ std::string sha256_provider() {
 /*
 * SHA-224 / SHA-256 compression function
 */
-void SHA_256::compress_digest(digest_type& digest, std::span<const uint8_t> input, size_t blocks) {
+void BOTAN_SCRUB_STACK_AFTER_RETURN SHA_256::compress_digest(digest_type& digest,
+                                                             std::span<const uint8_t> input,
+                                                             size_t blocks) {
```
reneme, line 61: This function contains fan-outs into platform dependent implementations as well as the pure software implementation of SHA-2. Because of the latter it should be annotated for stack scrubbing. Though, each of the platform-specific functions should also be annotated. This often results in the less-than-optimal situation where we end up scrubbing the platform-dependent stack frame and also the stack frame of this function, despite the software impl never being used.

I would argue that we should (after all) manually factor out the software implementation into an internal function that is annotated and remove the annotation from here. That way, we always just scrub one stack frame. If the annotation is disabled (e.g. because the compiler doesn't support it), inlining will revert back to the existing structure anyway.

What do you think?

reneme (commented):
This is now in a decent state to be considered a first step to making use of GCC's strubbing I think. One design question remains. See below.

At the moment I applied this to SHA-2, but enabling it for many other algorithms should be straight forward.

randombit (approved):

PR #4883, opened 2025-05-24
Add a Test utility for comparing optionals

randombit:
Merging #4699 caused the GCC ARM valgrind builds in nightly to fail. In all of the cases flagged, the line in question is comparing two optional values with ==. This seems like possibly a bug in how std::optional is implemented in this case, but it's hard to be sure.

Add an explicit helper for checking optionals that avoids directly using std::optional::operator== and use it in the RPKI tests.


randombit:
cc @arckoor

PR #4890, opened 2025-05-27
Improve RFC 3779 extension api

arckoor:
#4699 added x509 extensions for RFC 3779. While working on #4877 I realized that the api for actually using them is kind of horrible.


src/lib/x509/x509_ext.cpp
```diff
@@ -1510,6 +1510,10 @@ void ASBlocks::decode_inner(const std::vector<uint8_t>& in) {
 void ASBlocks::ASIdentifiers::encode_into(Botan::DER_Encoder& into) const {
    into.start_sequence();
 
+   if (!m_asnum.has_value() && !m_rdi.has_value()) {
+      throw Encoding_Error("One of asnum, rdi must be present");
+   }
+
```
arckoor, line 1516: When just doing `auto ext = std::make_unique<ASBlocks>()`, nothing is set initially, so the (std::nullopt, std::nullopt) case that was previously caught by the `ASIdentifiers` constructor now needs to be caught here too

src/lib/x509/x509_ext.h
```diff
@@ -750,8 +750,6 @@ class BOTAN_PUBLIC_API(3, 9) ASBlocks final : public Certificate_Extension {
 
             ASIdOrRange() = default;
 
-            ASIdOrRange(asnum_t id) : m_min(id), m_max(id) {}
-
             ASIdOrRange(asnum_t min, asnum_t max) : m_min(min), m_max(max) {
                if(max < min) {
                   throw Decoding_Error("AS range numbers must be sorted");
```
arckoor, line 755: I would love to make these constructors private too, but that clashes both with our helper methods in `x509_ext.cpp` as well as with `BER_Decoder`

arckoor:
@randombit I'm currently redoing all the tests, and I can test the encode by comparing the expected bits to `cert.v3_extensions().get_extension_bits(IPAddressBlocks::static_oid())`.
Can I somehow do something similar for decoding? I'd like to give it a DER hex string / vec<uint8_t> and have it decode that into an extension (or fail), but all the relevant functionality that I could find, that is the extensions `decode_inner()` function as well as all the `Extensions::` functions, are private. Is there a way I can get around that?
Of course I'd much prefer to just unit test the encoding / decoding functions of the individual classes, but that runs into the same private function problems.

src/lib/x509/x509_ext.cpp
```diff
@@ -1336,69 +1312,64 @@ void IPAddressBlocks::IPAddressOrRange<V>::decode_from(Botan::BER_Decoder& from)
...
+   }
 
-      // restore ones stored in 'unused' octet (which is the last octet before trailing ones)
-      for(size_t i = 0; i < unused_max; i++) {
-         addr_max[addr_len_max - 1] |= (1 << i);
-      }
+   // if they were 8, the entire octet should have been discarded
+   if(unused > 7) {
+      throw Decoding_Error("IP address range entry specified invalid number of unused bits.");
+   }
```
arckoor, line 1347: Fixes #4893

arckoor:
The API I'm mostly happy with now, though I'm still debating whether there should be helper methods for things like "inherited", or if the user should interpret the `std::nullopt` themselves. As for the tests, they're better now, they test encoding and decoding separately, but the decode side could be much better with proper access to the decoding methods.
I'll leave it as it is for now, maybe someone has better ideas.

randombit:
@arckoor Thanks I'll take a look asap. Can you please start by rebasing against latest master? There has been recently introduced some new checks in CI (most notably clang-tidy runs against headers now, while previously it only ran on `cpp` files)

arckoor:
:thinking: `clang-tidy` doesn't seem to check the `x509_ext.h` file. I also tried running it locally, same thing, but maybe I'm also just doing it wrong / missing something

src/tests/test_x509_rpki.cpp
```diff
@@ -152,53 +161,136 @@ constexpr auto IPv6 = Botan::Cert_Extension::IPAddressBlocks::Version::IPv6;
...
-   result.test_eq("ipv6 block 0 max", v6_blocks[0].max().value(), "FA800000000000007FFFFFFFFFFFFFFF");
-   result.test_eq("ipv6 block 1 min", v6_blocks[1].min().value(), "FE200000000000000000000000000000");
-   result.test_eq("ipv6 block 1 max", v6_blocks[1].max().value(), "FE20000007FFFFFFFFFFFFFFFFFFFFFF");
-   result.test_eq("ipv6 block 2 min", v6_blocks[2].min().value(), "2003000068293435042010C5000000C4");
-   result.test_eq("ipv6 block 2 max", v6_blocks[2].max().value(), "2003000068293435042010C5000000C4");
-   result.test_eq("ipv6 block 3 min", v6_blocks[3].min().value(), "AB010000000000000000000000000001");
-   result.test_eq("ipv6 block 3 max", v6_blocks[3].max().value(), "CD020000000000000000000000000002");
+      auto ext = cert.v3_extensions().get_extension_object_as<IPAddressBlocks>();
+      result.confirm("extension is not decoded", ext == nullptr);
+   }
```
arckoor, line 293: This I find kind of sketchy btw, as in why does it not throw, the ext should be critical. I will have to look into this again.

src/lib/x509/x509_ext.h
```diff
@@ -661,10 +661,11 @@ class BOTAN_PUBLIC_API(3, 9) IPAddressBlocks final : public Certificate_Extensio
...
 
-            // NOLINTNEXTLINE(*-explicit-conversions) FIXME
-            IPAddressChoice(std::optional<std::span<const IPAddressOrRange<V>>> ranges);
-
          private:
+            friend class IPAddressBlocks;
+
+            explicit IPAddressChoice(std::optional<std::span<const IPAddressOrRange<V>>> ranges);
+
             std::optional<std::vector<IPAddressOrRange<V>>> m_ip_addr_ranges;
```
arckoor, line 668: Was there a reason this was marked as a fixme instead of putting the `explicit` there right away?

arckoor:
@randombit can I please get a review of this? The next release is drawing closer, and we need to either merge this or revert #4699. We cannot leave it as is.

src/lib/x509/x509_ext.h
arckoor: Amazing Github changeset, I meant the removed `// NOLINTNEXTLINE(*-explicit-conversions) FIXME` above this

randombit:
@arckoor Thanks for the ping I'll review soon

src/lib/x509/x509_ext.h
randombit: IIRC adding an `explicit` there caused some compilation error that I could not immediately figure out. If you can get it working that would be nice since here (since so far these interfaces are not in a release) there is no compatibility issues with adding `explicit` right away.

src/lib/x509/x509_ext.h
```diff
@@ -580,11 +580,13 @@ class BOTAN_PUBLIC_API(3, 9) IPAddressBlocks final : public Certificate_Extensio
...
-            explicit IPAddress(std::span<uint8_t> v);
-
             std::array<uint8_t, Length> value() const { return m_value; }
 
+         private:
+            friend class IPAddressBlocks;
+            IPAddress() = default;
+            explicit IPAddress(std::span<const uint8_t> v);
+
             friend IPAddress<V> operator++(const IPAddress<V> v) {
```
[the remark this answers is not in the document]
randombit: Alternately here - drop the `operator++` entirely and use a named function (`next`?)

src/lib/x509/x509_ext.cpp
```diff
@@ -1283,48 +1282,25 @@ void IPAddressBlocks::IPAddressOrRange<V>::encode_into(Botan::DER_Encoder& into)
...
-
-      for(size_t i = 0; i < discarded_octets; i++) {
-         prefix.push_back(0);
-      }
+      // copy because we modify the address in `decode_single_address`, but we need it twice for min and max
+      std::vector<uint8_t> prefix_max(prefix_min);
 
-      m_min = IPAddress<V>(prefix);
+      // min address gets filled with 0's
+      decode_single_address(prefix_min, true);
```
randombit: Every call to `decode_single_address` immediately converts the vector to an `IPAddress<V>` - why not just return that directly?

randombit (changes requested):
@arckoor So the changes to x509_ext.{h,cpp} seem basically fine but tbh this PR is very hard to review because it both changes the API and modifies the tests both in terms of API usage but also what is being tested. The diff is basically unreadable.

I can see how for heavy usage the new API might be preferable; you are obviously more invested in this being ergonomic than I am. That said I generally prefer avoiding mutable interfaces (eg how `restrict_rdi` deletes previously entered state) *especially* since with X.509 extensions the only time the data is really (logically) mutable is during creation rather than after.

So I'd propose this
 - You removed or made private various constructors (eg `ASBlocks`) that had a fairly verbose but IMO usable interface. Bring these back.
 - Don't remove any of the existing tests (using the old, currently removed, interfaces).
 - Add new tests using your new interface

That admittedly leaves us with two slightly different ways of accomplishing creating an `ASBlocks` but TBH given the choice between the two I would prefer the verbose way of creating `ASIdentifier` list. And it should make then the diff comprehensible/reviewable.

arckoor:
@randombit fair points, I'll add the old constructors back in, but I'd really like to keep even the mutable methods like `restrict_*`, without them FFI would be very painful. In #4877 I already encountered the "extension should not be mutable when coming from a cert"-problem, and tried to work around it with a `self.__writable: bool` on the python side. Can I introduce a `m_writable` that gets set when decoding, and it just throws when you try to write to it anyway?

randombit:
> but I'd really like to keep even the mutable methods like restrict_*, without them FFI would be very painful. 

Yes that's fine.

> Can I introduce a m_writable that gets set when decoding, and it just throws when you try to write to it anyway?

That would work but perhaps that's not necessary in practice. In decoding , `Extensions` returns only a const pointer from its getters, so these mutating functions can't be called in any case.

src/tests/test_x509_rpki.cpp
```diff
@@ -115,6 +115,15 @@ std::tuple<std::string, std::string, std::string> get_sig_algo_padding() {
    return std::make_tuple(sig_algo, padding_method, hash_fn);
 }
 
+Botan::X509_Certificate make_self_signed(std::unique_ptr<Botan::RandomNumberGenerator>& rng,
+                                         const Botan::X509_Cert_Options& opts = std::move(ca_opts())) {
```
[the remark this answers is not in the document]
arckoor: If you do take it by value clang complains with `The parameter 'opts' is copied for each invocation but only used as a const reference; consider making it a const reference`

src/lib/x509/x509_ext.h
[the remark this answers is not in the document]
arckoor: The `operator+` has to stay either way, because I can't add a `next` to `uint32_t`, which is the base type for `AsIdOrRange::min()` (and max).
I need to be able to compute `max() + 1` for _both_ types, and the only way I know of to do this is to overload an existing operator for the IPAddress. I can change the `++` to a named next for IPAddress, but it's private & internal-only anyway.

src/lib/x509/x509_ext.h
```diff
@@ -716,6 +715,34 @@ class BOTAN_PUBLIC_API(3, 9) IPAddressBlocks final : public Certificate_Extensio
                     std::vector<std::set<Certificate_Status_Code>>& cert_status,
                     size_t pos) override;
 
+      template <Version V>
+      void add_address(const std::array<uint8_t, static_cast<size_t>(V)>& address,
```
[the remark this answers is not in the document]
arckoor: not sure where else I could call this without introducing footguns (like through a finish() method or so), the (old) constructor calls it right away too - I'd _like_ to call it right before encoding, but `encode_inner` is const

arckoor:
@randombit should be ready to take a look again. The only changed test I kept as it was is the decode test, since I find straight values more easily readable than hex. It also tests something different, since the logic was just wrong there.

~~Re the `m_writable`: yes in cpp it's not a problem, but the FFI can't know if the ext comes from being freshly constructed or from a cert, when a user gets an ext from a cert I have to copy the entire ext, because I can't guarantee that the user will destroy the ext before destroying the cert.
So I have an intended read-only ext, that is also writable. It of course won't change the cert it belongs to if you write to it, but is still kind of a footgun if the user expects to be able to modify the cert somehow. I can enforce the read-only-ness at the client, that is the python / rust binding, or I can enforce it in cpp, with an `m_writable` for `IPAddressBlocks` / `ASBlocks`. Whatever you prefer~~ Nevermind that, I'll just wrap it in a custom ffi struct, I forgot I could do that

src/lib/x509/x509_ext.h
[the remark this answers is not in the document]
randombit: Yeah I'm inclined to just accept the situation as it stands. I guess if someone is dumping thousands of ASes into their cert then things will get unpleasant but also this only affects construction which is (relative to parsing or cert validation) a rare event.

I guess an alternative would be to have some builder type

```
IPAddressBlocksBuilder builder;
builder.add_address(...);
IPAddressBlocks = builder.finish();
```

src/lib/x509/x509_ext.cpp
```diff
@@ -1336,69 +1309,64 @@ void IPAddressBlocks::IPAddressOrRange<V>::decode_from(Botan::BER_Decoder& from)
...
+   }
+}
 
-      // restore trailing zeros
-      const size_t addr_len_min = addr_min.size();
-      for(size_t i = 0; i < version_octets - addr_len_min; i++) {
-         addr_min.push_back(0);
-      }
+template <IPAddressBlocks::Version V>
+IPAddressBlocks::IPAddress<V> IPAddressBlocks::IPAddressOrRange<V>::decode_single_address(std::vector<uint8_t>& decoded,
```
randombit: If I'm reading it right, you are modifying `decoded` here to keep track of various state but (with now `IPAddress<V>` being returned) the actual value stored in `decoded` is not used after the function returns. That being the case I'd suggest taking the vector by value and (in callers) `std::move()`ing the vector.

randombit (commented):
This looks good to me. Can you clean up the commit history? Just squash to 1 if you like or try to split into API improvements vs new tests, however you prefer.

randombit:
There is also a merge conflict

PR #4892, opened 2025-05-28
Minor fixes for macOS and powerpc

barracuda156:
A few fixes to allow the build on older macOS versions.

1. The build hardcodes a few assumptions for macOS which may not hold. I cannot verify this on 10.7, so the threshold is based on documentation. I can say for sure that on 10.6 these are necessary.
2. Including `<sys/random.h>` may need <sys/types.h>, otherwise the build breaks on system headers, complaining that `u_int` has not been declared.
3. Unbreak powerpc module; this basically restores the code from Botan v2 with minor changes to fit it here. (I include OpenBSD simply because it is there in Botan code; I have no idea whether OpenBSD still needs a dedicated code, and can drop it from macros, if needed.)


barracuda156:
> CPUID changes look fine

Thank you, I will make changes according to your suggestions today and rebase the PR.

barracuda156:
@randombit I have dropped two commits, as advised, switched to `__has_include` and rebased.

randombit:
@barracuda156 Format checks are failing, you should be able to just apply the patch printed in the failing job

barracuda156:
@randombit I will check that, thanks.

KaganCanSit:
Hello @randombit,

I made a contribution here for the Clang format error. @barracuda156 responded very quickly. Thank you.

If you find it appropriate to revert the requested change request so that the CI processes can work, we can look into what happens. I don't have a MACOS device so I can't check locally.

If everything is OK, it can be merged as per your request.

Best regards.

randombit:
@barracuda156 LGTM please squash and then this can merge

randombit:
Closing. Any affected version of macOS is 15+ years old and cannot reasonably be considered secure, so complicating the code on its behalf is not in line with [the project goals](https://github.com/randombit/botan/blob/master/doc/goals.rst?plain=1#L98-L101)

PR #4895, opened 2025-06-01
FIX: create fresh garbage for each negative test

reneme:
With the current implementation of AEADs this doesn't matter. Though, given that the implementations of AEAD::update() and ::finish() can change the length of the in/out vector as a side-effect, this can result in unexpected test failures that aren't actually the implementation's fault.

Found when refactoring the OCB mode in #4880. When instantiated with a tag length of 12, a call to ::finish() had already made room for the tag before it failed (because of an expected state violation). Though, now the garbage vector's size has changed and doesn't fit the update granularity down the lane.

### DIFF
--- a/src/tests/test_aead.cpp
+++ b/src/tests/test_aead.cpp
@@ -40,13 +40,19 @@ class AEAD_Tests final : public Text_Based_Test {
          result.confirm("AEAD name is not empty", !enc->name().empty());
          result.confirm("AEAD default nonce size is accepted", enc->valid_nonce_length(enc->default_nonce_length()));
 
-         Botan::secure_vector<uint8_t> garbage = rng.random_vec(enc->update_granularity());
+         auto get_garbage = [&] { return rng.random_vec(enc->update_granularity()); };
 
          if(is_siv == false) {
-            result.test_throws("Unkeyed object throws for encrypt", [&]() { enc->update(garbage); });
+            result.test_throws("Unkeyed object throws for encrypt", [&]() {
+               auto garbage = get_garbage();
+               enc->update(garbage);
+            });
          }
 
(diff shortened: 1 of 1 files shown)


src/tests/test_aead.cpp
```diff
@@ -61,12 +67,20 @@ class AEAD_Tests final : public Text_Based_Test {
...
          if(is_siv == false) {
-            result.test_throws("Cannot process data until nonce is set (enc)", [&]() { enc->update(garbage); });
-            result.test_throws("Cannot process data until nonce is set (enc)", [&]() { enc->finish(garbage); });
+            result.test_throws("Cannot process data until nonce is set (enc)", [&]() {
+               auto garbage = get_garbage();
+               enc->update(garbage);
+            });
+            result.test_throws("Cannot process data until nonce is set (enc)", [&]() {
+               auto garbage = get_garbage();
+               enc->finish(garbage);
```
reneme, line 76: With #4880 this would re-alloc the garbage already in the base class (`Cipher_Mode`) to make room for the to-be-calculated tag. Then, the concrete implementation throws because of the (expected) invalid state.

src/tests/test_aead.cpp
```diff
@@ -61,12 +67,20 @@ class AEAD_Tests final : public Text_Based_Test {
...
+            });
          }
 
          enc->set_associated_data(mutate_vec(ad, rng));
          enc->start(mutate_vec(nonce, rng));
+
+         auto garbage = get_garbage();
          enc->update(garbage);
```
reneme, line 84: For OCB with a tag-length of 12 and an update_granularity of 16, the above-mentioned side effect (introduced in #4880) would result in a `garbage` vector that doesn't fit the update granularity and make this call to update fail unexpectedly.

randombit (approved):

Issue #4899, opened 2025-06-05
PQ nightly interop tests failing on nightly


randombit:
Just noticed this, should investigate. cc @reneme

randombit:
All failures seem to be eFrodoKEM related, IIRC there was some ciphersuite change here? I don't recall the details atm

reneme:
> All failures seem to be eFrodoKEM related, IIRC there was some ciphersuite change here? I don't recall the details atm

Probably this exchange is what you remember: https://github.com/randombit/botan/pull/4393#issuecomment-2536950763
I'll look into it.

Edit: OQS [changed the FrodoKEM code point assignments](https://github.com/open-quantum-safe/oqs-provider/commit/9447f68bd81aa6f3dc96d3187a83965384f5c424). This time into [the designated "private" area in the TLS code point](https://www.rfc-editor.org/rfc/rfc8446#page-42) space (0xFE.. - 0xFFFF).

PR #4900, opened 2025-06-06
FIX: re-assign eFrodoKEM code points to match OQS

reneme:
See: #4899

This harmonizes the code point assignments for eFrodoKEM and hybrid schemes with the ones [that OQS has assigned](https://github.com/open-quantum-safe/oqs-provider/commit/9447f68bd81aa6f3dc96d3187a83965384f5c424). All code points are defined in TLS 1.3's private code point section now.

Obviously, this is a breaking change for anyone that might use any of those groups already. I think we have a decision to make here:

1. Accept this and keep re-assigning eFrodoKEM code points to track whatever OQS happens to do. Also document that support for those code points is experimental and can change in minor releases.
2. Stop hard-coding non-standard code points and provide an API that allows re-assigning them at run time, or
3. Remove eFrodoKEM code points entirely and point people to documentation how to set up "custom" code points. Eg. [tls_custom_curves_client.cpp](https://github.com/randombit/botan/blob/master/src/examples/tls_custom_curves_client.cpp)


randombit:
We should in the short term do either (1) or just disable the OQS interop test for eFrodoKEM, to unbreak the nighly tests.

It seems to me like we might as well follow OQS here since that's (AFAIK) the only other stack that supports eFrodoKEM key exchange, and being to a) interop b) test that we can interop, are both useful.

Would be fine with 2 here also

3/4 seem like potentially a lot of complication

randombit (approved):
Fine with me to merge this as it is

PR #4907, opened 2025-06-08
Ed25519 cleanup

randombit:
Strikingly this PR is both much cleaner and also smaller and faster than ref10

With GCC 15 on Tigerlake, signing improves 13%, verification improves 8%, and the overall shared object is 136 Kb smaller.

With Clang 19 on Tigerlake, signing improves 4%, verification improves 2%, and the overall shared object is 45 Kb smaller.

### DIFF
--- a/src/lib/pubkey/ed25519/ed25519_fe.cpp
+++ b/src/lib/pubkey/ed25519/ed25519_fe.cpp
@@ -15,66 +15,57 @@
 namespace Botan {
 
 //static
-FE_25519 FE_25519::invert(const FE_25519& z) {
-   FE_25519 t0;
-   FE_25519 t1;
-   FE_25519 t2;
-   FE_25519 t3;
+Ed25519_FieldElement Ed25519_FieldElement::invert() const {
+   auto t0 = this->sqr();
+   auto t1 = t0.sqr_iter(2);
+   t1 = *this * t1;
+   t0 = t0 * t1;
+   auto t2 = t0.sqr();
+   t1 = t1 * t2;
+   t2 = t1.sqr_iter(5);
+   t1 = t2 * t1;
+   t2 = t1.sqr_iter(10);
+   t2 = t2 * t1;
+   auto t3 = t2.sqr_iter(20);
+   t2 = t3 * t2;
+   t2 = t2.sqr_iter(10);
+   t1 = t2 * t1;
+   t2 = t1.sqr_iter(50);
+   t2 = t2 * t1;
+   t3 = t2.sqr_iter(100);
+   t2 = t3 * t2;
+   t2 = t2.sqr_iter(50);
+   t1 = t2 * t1;
+   t1 = t1.sqr_iter(5);
 
-   fe_sq(t0, z);
-   fe_sq_iter(t1, t0, 2);
-   fe_mul(t1, z, t1);
-   fe_mul(t0, t0, t1);
-   fe_sq(t2, t0);
-   fe_mul(t1, t1, t2);
-   fe_sq_iter(t2, t1, 5);
-   fe_mul(t1, t2, t1);
-   fe_sq_iter(t2, t1, 10);
-   fe_mul(t2, t2, t1);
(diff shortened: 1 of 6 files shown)


src/lib/pubkey/ed25519/ed25519_fe.h
```diff
@@ -63,70 +65,69 @@ class FE_25519 {
...
 
-      FE_25519(const FE_25519& other) = default;
-      FE_25519& operator=(const FE_25519& other) = default;
+      constexpr Ed25519_FieldElement(const Ed25519_FieldElement& other) = default;
+      constexpr Ed25519_FieldElement& operator=(const Ed25519_FieldElement& other) = default;
 
-      FE_25519(FE_25519&& other) = default;
-      FE_25519& operator=(FE_25519&& other) = default;
+      constexpr Ed25519_FieldElement(Ed25519_FieldElement&& other) = default;
+      constexpr Ed25519_FieldElement& operator=(Ed25519_FieldElement&& other) = default;
```
reneme: Rule of zero? The destructor is gone as well, after all.

src/lib/pubkey/ed25519/ed25519_fe.h
```diff
@@ -63,70 +65,69 @@ class FE_25519 {
...
 
-      void from_bytes(const uint8_t b[32]);
-      void to_bytes(uint8_t b[32]) const;
+      static Ed25519_FieldElement deserialize(const uint8_t b[32]);
 
-      bool is_zero() const {
-         uint8_t s[32];
-         to_bytes(s);
+      void serialize(uint8_t b[32]) const;
```
reneme: In the nomenclature we established for pcurves, this should probably be called `serialize_to` and perhaps use a `std::span<uint8_t, 32>`.

src/lib/pubkey/ed25519/ed25519_fe.cpp
```diff
@@ -332,17 +332,17 @@ See fe_mul.c for discussion of implementation strategy.
...
+   int32_t f0 = m_fe[0];
+   int32_t f1 = m_fe[1];
+   int32_t f2 = m_fe[2];
+   int32_t f3 = m_fe[3];
+   int32_t f4 = m_fe[4];
+   int32_t f5 = m_fe[5];
+   int32_t f6 = m_fe[6];
+   int32_t f7 = m_fe[7];
+   int32_t f8 = m_fe[8];
+   int32_t f9 = m_fe[9];
```
reneme, line 336: Purely informational: Assuming `m_fe` were a `std::array`, such things could be rephrased to a structured binding. Also applies to other places, I believe.

```suggestion
   auto [f0, f1, f2, f3, f4, f5, f6, f7, f8, f9] = m_fe;
```

... obviously, that might be a matter of taste. ;)

randombit: Doesn't work because this implicitly declares each of the `f..` as `const int32_t` since the member function is const. :/

src/lib/pubkey/ed25519/ed25519_fe.h
randombit: Done

src/lib/pubkey/ed25519/ed25519_fe.h
randombit: :+1: rename and spanned

(Some of the other functions are still using raw array/ptr args at least some of the time, I will defer this)

src/lib/pubkey/ed25519/ed25519_fe.cpp
randombit: I realized the other instances of this we assign the values to `const` variables and could use this approach, but doing so caused a 5% overall performance regression for both sign and verify with Clang 20.

reneme: That's interesting. I guess the structured binding makes the compiler fetch the values straight from the array within the computation and that's bad for cache locality?

Issue #4909, opened 2025-06-08
Run clang-tidy on header files also


randombit:
I don't recall anymore why it doesn't currently work, but `clang-tidy` does not check/warn about issues in header files, only source files. Fix this.

(Would have prevented #4906)

randombit:
With #4919 now CI checks headers as well

PR #4910, opened 2025-06-08
Fix some clang-tidy warnings in header files

randombit:
~A very long road before this can be merged I think~ Realized I can just make checking the headers optional and currently off, so fixes can be merged in pieces.

#4909

### DIFF
--- a/src/lib/math/bigint/bigint.h
+++ b/src/lib/math/bigint/bigint.h
@@ -27,7 +27,7 @@ class BOTAN_PUBLIC_API(2, 0) BigInt final {
       /**
        * Base enumerator for encoding and decoding
        */
-      enum Base {
+      enum Base : uint16_t {
          Decimal BOTAN_DEPRECATED("All functions using this enum are deprecated") = 10,
          Hexadecimal BOTAN_DEPRECATED("All functions using this enum are deprecated") = 16,
          Binary BOTAN_DEPRECATED("All functions using this enum are deprecated") = 256
@@ -36,7 +36,7 @@ class BOTAN_PUBLIC_API(2, 0) BigInt final {
       /**
        * Sign symbol definitions for positive and negative numbers
        */
-      enum Sign { Negative = 0, Positive = 1 };
+      enum Sign : uint8_t { Negative = 0, Positive = 1 };
 
       /**
        * Create empty (zero) BigInt
@@ -164,14 +164,14 @@ class BOTAN_PUBLIC_API(2, 0) BigInt final {
       /**
        * Move constructor
        */
-      BigInt(BigInt&& other) { this->swap(other); }
+      BigInt(BigInt&& other) noexcept { this->swap(other); }
 
       ~BigInt() { _const_time_unpoison(); }
 
       /**
        * Move assignment
        */
-      BigInt& operator=(BigInt&& other) {
+      BigInt& operator=(BigInt&& other) noexcept {
          if(this != &other) {
             this->swap(other);
          }
@@ -188,12 +188,12 @@ class BOTAN_PUBLIC_API(2, 0) BigInt final {
        * Swap this value with another
        * @param other BigInt to swap values with
        */
-      void swap(BigInt& other) {
+      void swap(BigInt& other) noexcept {
          m_data.swap(other.m_data);
          std::swap(m_signedness, other.m_signedness);
       }
 
-      friend void swap(BigInt& x, BigInt& y) { x.swap(y); }
+      friend void swap(BigInt& x, BigInt& y) noexcept { x.swap(y); }
 
       BOTAN_DEPRECATED("Deprecated no replacement") void swap_reg(secure_vector<word>& reg) {
          m_data.swap(reg);
@@ -952,7 +952,7 @@ class BOTAN_PUBLIC_API(2, 0) BigInt final {
        * @warning this is an implementation detail which is not for
        * public use and not covered by SemVer.
        */
-      static BigInt _from_words(secure_vector<word>&& words) {
+      static BigInt _from_words(secure_vector<word>& words) {
          BigInt bn;
          bn.m_data.swap(words);
          return bn;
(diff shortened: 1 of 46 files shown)


randombit:
@reneme I could use your help on this one

```
build/include/internal/botan/internal/ct_utils.h:196:35: warning: forwarding reference parameter 'ts' is never forwarded inside the function body [
cppcoreguidelines-missing-std-forward]
  196 | constexpr void poison_all(Ts&&... ts) {
```

There are many instances of this in ct_utils.h and loadstor.h. I'm not sure if this is an issue or not. I'm fine with just silencing the warning locally _but_ I'm not sure if that's correct/optimal.

src/lib/math/bigint/bigint.h
```diff
@@ -1045,17 +1045,25 @@ class BOTAN_PUBLIC_API(2, 0) BigInt final {
...
-            void swap(secure_vector<word>& reg) {
+            // NOLINTBEGIN(*rvalue-reference-param-not-moved)
+            void swap(secure_vector<word>&& reg) noexcept {
                m_reg.swap(reg);
                invalidate_sig_words();
             }
 
-            void invalidate_sig_words() const { m_sig_words = sig_words_npos; }
+            // NOLINTEND(*rvalue-reference-param-not-moved)
+
```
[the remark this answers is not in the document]
randombit: @reneme this is another one I could use some help on if possible. Maybe we _just_ want `swap` taking a regular reference, and not rvalue references?

randombit:
There are still a ton more warnings but I'm going to call this PR good for the first tranche of fixes.

src/scripts/dev_tools/run_clang_tidy.py
```diff
@@ -114,6 +114,7 @@
     'readability-isolate-declaration',
     'readability-math-missing-parentheses',
     'readability-non-const-parameter',
+    'readability-redundant-inline-specifier', # Jack likes doing this
```
reneme, line 117: Please don't forget to `run_clang_tidy.py --regenerate-inline-config-file` :)

src/lib/math/bigint/bigint.h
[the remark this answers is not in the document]
reneme: (regarding the swap on bigint)

Yes. The rvalue ref doesn't seem right here. Except if you want to force the user of this method to always pass a temporary or use move, but that doesn't seem to make much sense for `swap`.

The regular lvalue reference should be fine.

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
```diff
@@ -40,7 +40,7 @@ inversions. See pcurves_secp256r1.cpp or pcurves_secp256k1.cpp for examples with
 the bells and whistles.
 */
 
-namespace {
+namespace {  // NOLINT(cert-dcl59-cpp)
```
reneme: Not exactly sure here, though in many other places we used `namespace detail` to "hide" such things from the public API. Or is this trying to do something else?

src/lib/modes/cipher_mode.h
```diff
@@ -23,9 +23,9 @@ namespace Botan {
 /**
 * The two possible directions a Cipher_Mode can operate in
 */
-enum class Cipher_Dir : int {
-   Encryption,
-   Decryption,
+enum class Cipher_Dir : uint8_t {
+   Encryption = 0,
+   Decryption = 1,
```
reneme, line 28: Isn't that a potential semver issue given that this enum (and potentially others in this PR) is/are public? Perhaps a somewhat strict interpretation of [Hyrum's Law](https://xkcd.com/1172/) from my side, I don't know.

src/lib/utils/ct_utils.h
```diff
@@ -356,6 +356,7 @@ class Choice final {
       constexpr Choice(Choice&& other) = default;
       constexpr Choice& operator=(const Choice& other) noexcept = default;
       constexpr Choice& operator=(Choice&& other) noexcept = default;
+      constexpr ~Choice() = default;
```
reneme, line 359: No default c'tor, all of the above are defaulted. rule-of-zero might also apply, no?

src/lib/utils/ct_utils.h
```diff
@@ -379,11 +380,14 @@ concept ct_conditional_assignable = requires(T lhs, const T& rhs, Choice c) { lh
...
    public:
-      static_assert(std::is_unsigned<T>::value && !std::is_same<bool, T>::value,
+      static_assert(std::is_unsigned_v<T> && !std::is_same_v<bool, T>,
                     "Only unsigned integer types are supported by CT::Mask");
 
       Mask(const Mask<T>& other) = default;
+      Mask(Mask<T>&& other) = default;
       Mask<T>& operator=(const Mask<T>& other) = default;
+      Mask<T>& operator=(Mask<T>&& other) = default;
+      ~Mask() = default;
```
reneme, line 390: rule-of-zero?

src/lib/utils/simd/simd_avx512/simd_avx512.h
```diff
@@ -22,6 +22,8 @@ class SIMD_16x32 final {
       SIMD_16x32& operator=(SIMD_16x32&& other) = default;
       SIMD_16x32(SIMD_16x32&& other) = default;
 
+      ~SIMD_16x32() = default;
```
reneme, line 25: rule-of-zero?

reneme:
> @reneme I could use your help on this one
> 
> ```
> build/include/internal/botan/internal/ct_utils.h:196:35: warning: forwarding reference parameter 'ts' is never forwarded inside the function body [
> cppcoreguidelines-missing-std-forward]
>   196 | constexpr void poison_all(Ts&&... ts) {
> ```
> 
> There are many instances of this in ct_utils.h and loadstor.h. I'm not sure if this is an issue or not. I'm fine with just silencing the warning locally _but_ I'm not sure if that's correct/optimal.

This is a reasonable warning from clang-tidy, IMO. They should be converted to `const&`. The forwarding reference seem to be somewhat muscle memory for me when I write templated wrappers. 😞 But it doesn't really make sense if the variable isn't actually forwarded. It doesn't have an obvious adverse effect either though, I think.

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
randombit: I think the anon namespace here is an artifact of an earlier iteration of pcurves, it is not needed anymore

src/lib/utils/ct_utils.h
randombit: Rule of zero applies for sure, thinking on it further though I decided I prefer doing all 5 even if more verbose because it documents clearly the intent. Whereas rule of zero results in code with two possible interpretations "the default behavior works correctly"_or_ "the default behavior is being used because it's the default and I forgot the change it, it might be correct here but maybe not"

src/lib/modes/cipher_mode.h
randombit: Maybe? I guess while certainly this is a visible change I'm not sure I consider

`static_assert(sizeof(Cipher_Dir) == sizeof(int))`

any more reasonable than

`static_assert(sizeof(BlockCipher) == 98)`

and I can't come up with something that isn't effectively the above as something that would be affected by this.

It does affect the ABI, but in general Botan does not / can not support building the application against one set of headers and then running against another version; even a single type changing size is enough to make this unsafe.

PR #4912, opened 2025-06-10
Further clang-tidy fixes in header files

randombit:
#4909

### DIFF
--- a/src/lib/hash/blake2s/blake2s.cpp
+++ b/src/lib/hash/blake2s/blake2s.cpp
@@ -148,9 +148,9 @@ void BLAKE2s::final_result(std::span<uint8_t> out) {
 
 std::unique_ptr<HashFunction> BLAKE2s::copy_state() const {
    std::unique_ptr<BLAKE2s> h = std::make_unique<BLAKE2s>(m_outlen << 3);
-   memcpy(h->m_b, m_b, sizeof(m_b));
-   memcpy(h->m_h, m_h, sizeof(m_h));
-   memcpy(h->m_t, m_t, sizeof(m_t));
+   std::memcpy(h->m_b, m_b, sizeof(m_b));
+   std::memcpy(h->m_h, m_h, sizeof(m_h));
+   std::memcpy(h->m_t, m_t, sizeof(m_t));
    h->m_c = m_c;
    return h;
 }
--- a/src/lib/pubkey/dilithium/dilithium_common/dilithium_constants.h
+++ b/src/lib/pubkey/dilithium/dilithium_common/dilithium_constants.h
@@ -70,19 +70,19 @@ class DilithiumConstants final {
       /// @}
 
    public:
-      enum DilithiumTau : uint32_t { _39 = 39, _49 = 49, _60 = 60 };
+      enum DilithiumTau : uint8_t { _39 = 39, _49 = 49, _60 = 60 };
 
-      enum DilithiumLambda : uint32_t { _128 = 128, _192 = 192, _256 = 256 };
+      enum DilithiumLambda : uint16_t { _128 = 128, _192 = 192, _256 = 256 };
 
       enum DilithiumGamma1 : uint32_t { ToThe17th = (1 << 17), ToThe19th = (1 << 19) };
 
       enum DilithiumGamma2 : uint32_t { Qminus1DevidedBy88 = (Q - 1) / 88, Qminus1DevidedBy32 = (Q - 1) / 32 };
 
-      enum DilithiumEta : uint32_t { _2 = 2, _4 = 4 };
+      enum DilithiumEta : uint8_t { _2 = 2, _4 = 4 };
 
-      enum DilithiumBeta : uint32_t { _78 = 78, _196 = 196, _120 = 120 };
+      enum DilithiumBeta : uint8_t { _78 = 78, _196 = 196, _120 = 120 };
 
-      enum DilithiumOmega : uint32_t { _80 = 80, _55 = 55, _75 = 75 };
+      enum DilithiumOmega : uint8_t { _80 = 80, _55 = 55, _75 = 75 };
 
       DilithiumConstants(DilithiumMode dimension);
       ~DilithiumConstants();
--- a/src/lib/pubkey/pqcrystals/pqcrystals.h
+++ b/src/lib/pubkey/pqcrystals/pqcrystals.h
@@ -25,7 +25,7 @@
 
 namespace Botan::CRYSTALS {
 
-enum class Domain { Normal, NTT };
+enum class Domain : uint8_t { Normal, NTT };
 
 template <typename T>
 concept crystals_constants =
@@ -75,6 +75,8 @@ class Trait_Base {
       /// @}
 
    protected:
+      Trait_Base() = default;  // NOLINT(*crtp-constructor-accessibility)
+
       /// @returns the number of polynomials in the polynomial vector @p polyvec.
       static constexpr size_t polys_in_polyvec(std::span<const T> polyvec) {
          BOTAN_DEBUG_ASSERT(polyvec.size() % N == 0);
@@ -232,7 +234,7 @@ class Polynomial {
        */
       template <Domain OtherD>
          requires(D != OtherD)
-      explicit Polynomial(Polynomial<Trait, OtherD>&& other) noexcept :
+      explicit Polynomial(Polynomial<Trait, OtherD>&& other) noexcept :  // NOLINT(*-rvalue-reference-param-not-moved)
             m_coeffs_storage(std::move(other.m_coeffs_storage)),
             m_coeffs(owns_storage() ? std::span<T, Trait::N>(m_coeffs_storage) : other.m_coeffs) {}
 
@@ -372,7 +374,8 @@ class PolynomialVector {
        */
       template <Domain OtherD>
          requires(D != OtherD)
-      explicit PolynomialVector(PolynomialVector<Trait, OtherD>&& other) noexcept :
+      /* NOLINTNEXTLINE(*rvalue-reference-param-not-moved) */ explicit PolynomialVector(
+         PolynomialVector<Trait, OtherD>&& other) noexcept :
             m_polys_storage(std::move(other.m_polys_storage)) {
          BOTAN_DEBUG_ASSERT(m_polys_storage.size() % Trait::N == 0);
          const size_t vecsize = m_polys_storage.size() / Trait::N;
--- a/src/lib/pubkey/sphincsplus/sphincsplus_common/sp_hash.h
+++ b/src/lib/pubkey/sphincsplus/sphincsplus_common/sp_hash.h
@@ -20,7 +20,7 @@ namespace Botan {
  * A collection of pseudorandom hash functions required for SLH-DSA
  * computations. See FIPS 205, Section 11.2.1 and 11.2.2.
  **/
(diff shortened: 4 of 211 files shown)


randombit:
~This now fixes (or silences) _almost_ all of the warnings except for `special-member-function` which I'll do in another PR~

Now with almost everything fixed except a few tricky or confusing issues

randombit:
The enum changes in Dilithium/Kyber are particularly worth taking a look at since sometimes arithmetic is done on the return values and now we may promote to `int`.

Possibly the `enum-size` warning should just be disabled, idk. Certainly the gains there are marginal in the best of cases.

src/lib/block/kuznyechik/kuznyechik.cpp
```diff
@@ -241,10 +236,12 @@ void Kuznyechik::key_schedule(std::span<const uint8_t> key) {
...
-   m_rke[0][0] = k0;
-   m_rke[0][1] = k1;
-   m_rke[1][0] = k2;
-   m_rke[1][1] = k3;
+   m_rke.resize(20);
+
+   m_rke[0] = k0;
+   m_rke[1] = k1;
+   m_rke[2] = k2;
+   m_rke[3] = k3;
```
reneme: Nit: I didn't look into the details here. But given this was a 2D-array before: perhaps leave a ` // TODO(C++23): use std::mdspan`?

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -148,9 +148,9 @@ void BLAKE2s::final_result(std::span<uint8_t> out) {
 
 std::unique_ptr<HashFunction> BLAKE2s::copy_state() const {
    std::unique_ptr<BLAKE2s> h = std::make_unique<BLAKE2s>(m_outlen << 3);
-   memcpy(h->m_b, m_b, sizeof(m_b));
-   memcpy(h->m_h, m_h, sizeof(m_h));
-   memcpy(h->m_t, m_t, sizeof(m_t));
+   std::memcpy(h->m_b, m_b, sizeof(m_b));
+   std::memcpy(h->m_h, m_h, sizeof(m_h));
+   std::memcpy(h->m_t, m_t, sizeof(m_t));
```
reneme, line 153: Perhaps use `copy_mem` once the migration to `secure_vector` is done?

src/lib/pubkey/dilithium/dilithium_common/dilithium_constants.h
```diff
@@ -70,19 +70,19 @@ class DilithiumConstants final {
...
       enum DilithiumGamma2 : uint32_t { Qminus1DevidedBy88 = (Q - 1) / 88, Qminus1DevidedBy32 = (Q - 1) / 32 };
 
-      enum DilithiumEta : uint32_t { _2 = 2, _4 = 4 };
+      enum DilithiumEta : uint8_t { _2 = 2, _4 = 4 };
 
-      enum DilithiumBeta : uint32_t { _78 = 78, _196 = 196, _120 = 120 };
+      enum DilithiumBeta : uint8_t { _78 = 78, _196 = 196, _120 = 120 };
 
-      enum DilithiumOmega : uint32_t { _80 = 80, _55 = 55, _75 = 75 };
+      enum DilithiumOmega : uint8_t { _80 = 80, _55 = 55, _75 = 75 };
```
reneme, line 85: > Possibly the `enum-size` warning should just be disabled, idk. Certainly the gains there are marginal in the best of cases.

I agree. Those enums were added to programatically restrict the value space of those variables to the standardized values.

src/lib/pubkey/pqcrystals/pqcrystals.h
```diff
@@ -232,7 +234,7 @@ class Polynomial {
        */
       template <Domain OtherD>
          requires(D != OtherD)
-      explicit Polynomial(Polynomial<Trait, OtherD>&& other) noexcept :
+      explicit Polynomial(Polynomial<Trait, OtherD>&& other) noexcept :  // NOLINT(*-rvalue-reference-param-not-moved)
             m_coeffs_storage(std::move(other.m_coeffs_storage)),
             m_coeffs(owns_storage() ? std::span<T, Trait::N>(m_coeffs_storage) : other.m_coeffs) {}
```
reneme, line 239: From [a quick-n-dirty experiment](https://godbolt.org/z/TTjPr8dWx) I would conclude that this might work too and silence the warning:

```suggestion
      explicit Polynomial(Polynomial<Trait, OtherD>&& other) noexcept : 
            m_coeffs_storage(std::move(other).m_coeffs_storage),
            m_coeffs(owns_storage() ? std::span<T, Trait::N>(m_coeffs_storage) : other.m_coeffs) {}
```

... it does feel slightly iffy, though, because we still use `other` to initialize `m_coeffs` in the next line.

src/lib/pubkey/pqcrystals/pqcrystals.h
```diff
@@ -372,7 +374,8 @@ class PolynomialVector {
        */
       template <Domain OtherD>
          requires(D != OtherD)
-      explicit PolynomialVector(PolynomialVector<Trait, OtherD>&& other) noexcept :
+      /* NOLINTNEXTLINE(*rvalue-reference-param-not-moved) */ explicit PolynomialVector(
+         PolynomialVector<Trait, OtherD>&& other) noexcept :
             m_polys_storage(std::move(other.m_polys_storage)) {
```
reneme, line 379:
```suggestion
      explicit PolynomialVector(
         PolynomialVector<Trait, OtherD>&& other) noexcept :
            m_polys_storage(std::move(other).m_polys_storage) {
```

... see above.

src/lib/pubkey/sphincsplus/sphincsplus_common/sp_hash.h
```diff
@@ -55,7 +55,7 @@ class BOTAN_TEST_API Sphincs_Hash_Functions {
 
       template <typename... BufferTs>
       void T(std::span<uint8_t> out, const Sphincs_Address& address, BufferTs&&... in) {
-         auto& hash = tweak_hash(address, (std::forward<BufferTs>(in).size() + ...));
+         auto& hash = tweak_hash(address, (in.size() + ...));
          (hash.update(std::forward<BufferTs>(in)), ...);
```
reneme: Perhaps even:

`const BufferTs&...` and no `std::forward` at all. Temporary lifetime extension should just be good enough here. (Similar to the pattern in the hash wrappers in Dilithium/Kyber).

src/lib/pubkey/xmss/xmss_tools.h
```diff
@@ -34,7 +36,7 @@ class XMSS_Tools final {
        *               integral value to.
        * @param src integral value to concatenate.
        **/
-      template <typename T, typename U = typename std::enable_if<std::is_integral<T>::value, void>::type>
+      template <typename T, typename U = std::enable_if_t<std::is_integral_v<T>, void>>
```
reneme:
```suggestion
      template <std::integral T>
```

... I think `std::enable_if` in itself should produce a clang-tidy warning in C++20.

src/lib/pubkey/xmss/xmss_tools.h
```diff
@@ -47,7 +49,7 @@ class XMSS_Tools final {
        * @param len number of bytes to concatenate. This value must be smaller
        *            or equal to the size of type T.
        **/
-      template <typename T, typename U = typename std::enable_if<std::is_integral<T>::value, void>::type>
+      template <typename T, typename U = std::enable_if_t<std::is_integral_v<T>, void>>
```
reneme:
```suggestion
      template <std::integral T>
```

... note that those templates are (for some reason) implemented out-of-band below this class. The definition template clauses need the same transformation for this to work. I.e. they also should read `template <std::integral T>`.

src/lib/tls/tls13/msg_certificate_13.cpp
```diff
@@ -264,8 +264,8 @@ Certificate_13::Certificate_13(const Client_Hello_13& client_hello,
 }
 
 Certificate_13::Certificate_Entry::Certificate_Entry(TLS_Data_Reader& reader,
-                                                     const Connection_Side side,
-                                                     const Certificate_Type cert_type) {
+                                                     Connection_Side side,
+                                                     Certificate_Type cert_type) {
```
reneme, line 268: Curious: was that a clang-tidy warning?

src/lib/utils/assert.h
```diff
@@ -112,10 +112,10 @@ namespace Botan {
 * BOTAN_UNUSED(a); or BOTAN_UNUSED(x, y, z);
 */
 template <typename T>
-constexpr void ignore_param(T&&) {}
+constexpr void ignore_param(const T&) {}
 
 template <typename... T>
-constexpr void ignore_params(T&&... args) {
+constexpr void ignore_params(const T&... args) {
```
reneme: Unrelated to clang-tidy, but [the non-variadic overload isn't necessary](https://godbolt.org/z/PEbeWf947) here. Like so:

```C+
template <typename... Ts>
constexpr void ignore_params(const Ts&...) {}
```

reneme (approved):
Some nits and suggestions. Thanks for taking this on. `clang-tidy` ftw.

src/lib/pubkey/pqcrystals/pqcrystals.h
randombit: Yeah that seems a) certain to set off a use-after-move b) be highly confusing to readers. I think we just have to accept clang-tidy doesn't model partial moves and silence where necessary.

src/lib/pubkey/xmss/xmss_tools.h
randombit: It seems like https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-constraints.html should be exactly this, it is not disabled in our configuration (afaict) but I have not seen the warning

randombit: Oh maybe this is why?

> Other uses, for example, in class templates for function parameters, are not currently supported by this tool.

src/lib/tls/tls13/msg_certificate_13.cpp
randombit: Yes

src/lib/utils/assert.h
randombit: I think I wrote that by analogy to the old pre-C++11 implementation which used a recursive macro

src/lib/pubkey/xmss/xmss_tools.h
reneme: Mhh. Well, at least there is a rule. 👍
I would have expected it to fire, though. It's a function template. No class template parameter involved. 🤷‍♂️

Oh well.

reneme: Mhh. Well, at least there is a rule. 👍
I would have expected it to fire, though. It's a function template. No class template parameter involved. 🤷‍♂️

Oh well.

randombit: Pulled these out of the pseudo-class and applied a concept instead of enable_if

src/lib/block/kuznyechik/kuznyechik.cpp
randombit: I think the 2d array notation is not really even helpful here.

PR #4913, opened 2025-06-10
FIX: clang-tidy warnings in loadstor.h

reneme:
work-in-progress

### DIFF
--- a/src/lib/utils/loadstor.h
+++ b/src/lib/utils/loadstor.h
@@ -211,7 +211,7 @@ constexpr auto wrap_strong_type_or_enum(T t) {
  * This is only used at compile time.
  */
 template <std::endian endianness, std::unsigned_integral OutT, ranges::contiguous_range<uint8_t> InR>
-inline constexpr OutT fallback_load_any(InR&& in_range) {
+inline constexpr OutT fallback_load_any(const InR& in_range) {
    std::span in{in_range};
    // clang-format off
    if constexpr(endianness == std::endian::big) {
@@ -233,7 +233,7 @@ inline constexpr OutT fallback_load_any(InR&& in_range) {
  * This will be used only at compile time.
  */
 template <std::endian endianness, std::unsigned_integral InT, ranges::contiguous_output_range<uint8_t> OutR>
-inline constexpr void fallback_store_any(InT in, OutR&& out_range) {
+inline constexpr void fallback_store_any(InT in, OutR&& out_range /* NOLINT(*-std-forward) */) {
    std::span out{out_range};
    // clang-format off
    if constexpr(endianness == std::endian::big) {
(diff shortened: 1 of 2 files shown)


src/lib/utils/ct_utils.h
```diff
@@ -93,32 +93,37 @@ inline bool poison_has_effect() {
...
  * Poison a contiguous buffer of trivial objects (e.g. integers and such)
  */
 template <ranges::spanable_range R>
-   requires std::is_trivially_copyable_v<std::ranges::range_value_t<R>>
-constexpr void poison(R&& r) {
+   requires std::is_trivially_copyable_v<std::ranges::range_value_t<R>> && (!custom_poisonable<R>)
+constexpr void poison(const R& r) {
    std::span s{r};
    poison(s.data(), s.size());
 }
```
reneme, line 122: Something like:

```C++
class Foo {
  auto begin();
  auto end();
  size_t size();

  void _const_time_poison();
  void _const_time_unpoison();
};
```

... is a spanable range and also provides custom poison hooks. For instance, `pqcrystals::Polynomial` is such a class. I didn't look into the details as to why converting from a universal reference to a `const&` uncovers this, but it does. Disambiguating like above fixes the problem.

src/lib/utils/loadstor.h
```diff
@@ -352,7 +352,7 @@ template <std::endian endianness,
           ranges::contiguous_range<uint8_t> InR>
    requires(unsigned_integralish<std::ranges::range_value_t<OutR>> &&
             (std::same_as<AutoDetect, OutT> || std::same_as<OutT, std::ranges::range_value_t<OutR>>))
-inline constexpr void load_any(OutR&& out, InR&& in) {
+inline constexpr void load_any(OutR&& out /* NOLINT(*-std-forward) */, const InR& in) {
```
reneme, line 355: Many of the occasions were similar to this, fairly tricky, issue. Let me unpack: `OutR` and `InR` are both contiguous ranges, i.e. "template types". Hence, `OutR&&` and `InR&&` are universal references. Clang-tidy is wondering why we use them if we don't forward anything. The "obvious" fix would be to replace them by `OutR&` and `const InR&`, where "out" is non-const, because it is an out-param.

Now, at the call site, `const&` can bind to rvalues (via lifetime extension). A non-const ref can only bind to lvalues (lifetime extension doesn't apply here, I believe because the compiler doesn't want to let use modify temporaries).

Why would we pass a temporary range as an out-param? Because `std::span`!

Consider this situation where the first `load_be` implements a generic `load_be` and we provide a wrapper that can be called like `load_be<uint32_t>(some_array)`, using the generic implementation:

```C++
namespace detail {
template <contiguous_range InR, contiguous_range OutR>
void load_be(OutR& out /* non-const ref !! */, const InR& in) {
  // some fancy, generic loading implementation...
}

template <std::unsigned_integral OutT, contiguous_range InR>
OutT load_be(const InR& in) {
  OutT out;
  load_be(std::span<OutT, 1>(&out, 1) /* rvalue - compile error! */, in);
  return out;
}
```

... the `std::span` in the wrapper is an rvalue, which cannot be passed to a non-const-ref parameter (see above). But it can be passed to a universal reference! The fact that `std::span` is used as a "temporary wrappper" to mimick a range in the second function is making things complicated here.

If the out-range (`OutR`) is a span we can (and also should) take it by-value. If it were an actual range (say a `std::vector<>`) we have to take it by-value, otherwise we'd copy the out-vector, write to the copy and leave blank our actual out-variable at the call-site. The universal reference `OutR&&` is giving us exactly this behaviour.

reneme (commented):
> I cannot easily determine the fix which makes me think there is too much magic happening here

I hope that my explanation is helpful. I do agree though that this is more magic than it should be. :( Especially the `OutR&&` case is worth another look. Perhaps there's a better and more convenient way to do this.

src/lib/utils/ct_utils.h
randombit: Yeah this was the one I hit and finally gave up because I wasn't sure the best way to proceed :sweat_smile:

randombit (approved):
Thanks for looking into this.

I'm going to merge this now and rebase the other fixes

PR #4919, opened 2025-06-14
Clang-tidy headers in CI

randombit:
Addresses the final warnings and modifies the `clang-tidy` runner script to always check headers


reneme:
> Header file build/include/internal/botan/internal/bitvector.h cannot be handled

😨 I'll look into this latest by the end of the week.

randombit:
Now seems to fail/report as expected. Interestingly there seem to be some warnings occuring here that I don't see locally with Clang 20 - the only one I had expected was the warning in bitvector.h re `~0`.

Ah, the other warning was removed in `clang-tidy 19` https://reviews.llvm.org/D147563

src/lib/tls/asio/asio_error.h
```diff
@@ -26,14 +26,29 @@
  */
 
 namespace Botan {
+
+/**
+* Generic base class wrapping boost::system::error_category and
+* adding a (bizarrely missing) virtual destructor.
```
reneme, line 32: #Apparently they do that deliberately, given that the respective warning is explicitly disabled (from boost's `error_category.hpp`):

```C++
#if ( defined( BOOST_GCC ) && BOOST_GCC >= 40600 ) || defined( BOOST_CLANG )
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wnon-virtual-dtor"
#endif
```

also:

```C++
    // We'd like to make the destructor protected, to make code that deletes
    // an error_category* not compile; unfortunately, doing the below makes
    // the destructor user-provided and hence breaks use after main, as the
    // categories may get destroyed before code that uses them

    // ~error_category() {}
```

reneme (approved):
🤞

reneme:
The full clang-tidy run is getting close to a runtime of 90 minutes, which is the GitHub Actions timeout, if I'm not mistaken. This is already parallelized, right? Maybe we have to split this job into "library" and "CLI,tests,..."?

randombit:
> The full clang-tidy run is getting close to a runtime of 90 minutes, which is the GitHub Actions timeout, if I'm not mistaken. This is already parallelized, right? Maybe we have to split this job into "library" and "CLI,tests,..."?

Limit is 6 hours per https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/troubleshooting-workflows/actions-limits

It's already running in parallel, yes.

PR adding caching for clang-tidy is coming shortly. With a hot cache re-scanning the entire tree takes under 30 seconds.

PR #4920, opened 2025-06-17
Add a cache for clang-tidy

randombit:
We only cache results where `clang-tidy` produced no results since that's by far the common/expected case, and not having to worry about replicating fixits etc simplifies the cache.

The cache index is a hash of the preprocessed source plus the clang-tidy configuration flags and the output of `clang-tidy --version`.

### DIFF
--- a/src/scripts/dev_tools/run_clang_tidy.py
+++ b/src/scripts/dev_tools/run_clang_tidy.py
@@ -1,21 +1,25 @@
 #!/usr/bin/env python3
 
 """
-(C) 2022,2023 Jack Lloyd
+(C) 2022,2023,2025 Jack Lloyd
 
 Botan is released under the Simplified BSD License (see license.txt)
 """
 
-import subprocess
-import sys
+import hashlib
 import json
+import multiprocessing
+from multiprocessing.pool import ThreadPool
 import optparse # pylint: disable=deprecated-module
 import os
+import random
 import re
-import multiprocessing
+import shlex
+import sqlite3
+import subprocess
+import sys
 import time
 import uuid
-from multiprocessing.pool import ThreadPool
 
 quick_checks = [
     '-clang-analyzer*', # has to be explicitly disabled
@@ -183,10 +187,6 @@ def run_clang_tidy(compile_commands_file,
 
     stdout = run_command(cmdline)
 
-    if options.verbose:
-        print("Checked", source_file)
-        sys.stdout.flush()
-
     if stdout != "":
         print("### Errors in", source_file)
         print(stdout)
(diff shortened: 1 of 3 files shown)


reneme:
image

... sorry, that wasn't on purpose. I meant to delete a comment I had pending.

## 🤦‍♂️

src/scripts/dev_tools/run_clang_tidy.py
```diff
@@ -204,13 +202,71 @@ def file_matches(file, args):
...
+
+    return outer_hash.hexdigest()
+
+class CacheDatabase:
+    CACHE_DEBUG = False
+
+    def __init__(self, db_path):
+        self.db = sqlite3.connect(db_path)
+        cur = self.db.cursor()
+        cur.execute("CREATE TABLE IF NOT EXISTS clang_tidy_clean(key UNIQUE)")
```
reneme, line 248: As far as I understand, the cache is ever-growing, right? Perhaps add a "last hit" timestamp right from the start, that gets updated when the cache line is hit. Otherwise, at least the cache on the master branch might grow quite large eventually.

The "last hit" timestamp would allow us to just remove things that weren't useful for more than -- i don't know -- 2 months? (this value could be configured in `repo_config.env` along with the ccache size)

Alternatively, one could of course just zap it once before each release, or so. But that might be forgotten leading to adverse effects once we hit some GH Actions limit.

src/scripts/dev_tools/run_clang_tidy.py
```diff
@@ -204,13 +203,78 @@ def file_matches(file, args):
...
+    outer_hash = hashlib.sha256()
+
+    for key in sorted(kwargs.keys()):
+        value = kwargs[key]
+        outer_hash.update(hashlib.sha256(key.encode('utf-8')).digest())
+        outer_hash.update(hashlib.sha256(value.encode('utf-8')).digest())
+
+    return outer_hash.hexdigest()
+
+class CacheDatabase:
```
reneme, line 237:
```suggestion
class CacheDatabase:
    """
    Caches SUCCESSFUL clang-tidy runs on source files to speed up whole-tree
    runs of clang-tidy. The cache key is calculated as the hash of the source file
    and a set of meta information such as 'clang-tidy --version' and the check configuration.
    """
```

src/scripts/dev_tools/run_clang_tidy.py
```diff
@@ -204,13 +203,78 @@ def file_matches(file, args):
...
+
+    def hit(self, **kwargs):
+        cache_key = hash_args(**kwargs)
+
+        cur = self.db.cursor()
+        cur.execute("SELECT key from clang_tidy_clean where key = ?", (cache_key, ))
+        res = cur.fetchall()
+
+        for r in res:
+            if r[0] == cache_key:
```
reneme: We just selected for the cache key. How can this be false? Am I missing something? Also, the cache key is declared to be unique. So the loop should also be redundant, no?

src/scripts/dev_tools/run_clang_tidy.py
```diff
@@ -204,13 +203,78 @@ def file_matches(file, args):
...
+
+    return outer_hash.hexdigest()
+
+class CacheDatabase:
+    CACHE_DEBUG = False
+
+    def __init__(self, db_path):
+        self.db = sqlite3.connect(db_path)
+        cur = self.db.cursor()
+        cur.execute("CREATE TABLE IF NOT EXISTS clang_tidy_clean(key UNIQUE)")
```
reneme, line 248: Do we need to explicitly close this? (I'm guessing the `self.db.commit()` in `record()` makes this redundant, just asking to be on the safe side)

src/scripts/dev_tools/run_clang_tidy.py
randombit: I wouldn't worry about it. GH Actions has a max 10 gigabytes on cache, after which caches are purged in LRU order. And their approach of only offering a write-once non-updateable cache means with ccache we blow past this limit all the time. Currently we are using 26 GB and the oldest available cache is from yesterday.

GH Actions caches are not long lived.

reneme: Maybe. But by pulling the cache db, extending it and then pushing it again our "top of cache" keeps growing forever. GitHub keeps a copy of each state (because the cache values itself are immutable), but still we keep increasing the "latest" incarnation of the database, no?

(I'm not saying that I'm expecting this to blow up quickly, just want to be a good citizen. 😏)

src/scripts/dev_tools/run_clang_tidy.py
randombit: I think this is kind of a leftover from debugging

src/scripts/dev_tools/run_clang_tidy.py
randombit: I still think it's quite unlikely to ever be a problem, but I added a random eviction with max 10K entries

src/scripts/dev_tools/run_clang_tidy.py
```diff
@@ -204,13 +203,89 @@ def file_matches(file, args):
...
+        cur = self.db.cursor()
+        if self.CACHE_DEBUG:
+            print("Cache save for", cache_key)
+        cur.execute("INSERT OR IGNORE INTO clang_tidy_clean values(?)", (cache_key, ))
+        self.db.commit()
+
+    def cleanup(self, limit):
+        cur = self.db.cursor()
+        cur.execute("DELETE FROM clang_tidy_clean ORDER BY RANDOM() LIMIT ((SELECT COUNT(*) FROM clang_tidy_clean) - ?)", (limit, ))
+        self.db.commit()
```
reneme: We seem to have a different mental model of how the GH cache works. Right now I'm fairly convinced that randomly removing entries will inevitably reduce the efficiency of this cache. To my mind, this cleanup would happen in every run (after some warm up time) and then start hurting its efficiency by evicting useful entries at random. For the record: That will very likely not result in huge problem, but I'm surprised by the approach nevertheless. Maybe I'm just too pedantic here, sorry for that. 😅 

Here's how I understand the cache operation: Let's say we are building 'master':

* *1st build:* the **cache is cold** and filled with ~600 "success" entries after the successful run
* *2nd build:* Assume we merged **just a small fix** to one *.cpp file: the cache is hot, only one cache miss resulting in a single new record. Now we have 601 "success" entries (of which 1 is likely never used again)
* *3rd build:* A **large new feature** got merged: 20 cache misses, now 621 "success" entries (21 will likely never be used again)
* _n-th build_: ... cache keeps being hot but **fills up with outdated entries** until we start evicting. By chance we evict 50 useful entries _(which is very unlikely)_
* _(n+1)-th build_: Let's say **just a tiny fix** again: the cache is now much colder than it cold be, because we got really unlucky in the random eviction before.

---

If we wouldn't evict at all, the database would keep growing over time, because the cache (for master) is never invalidated. The random eviction might (unluckily) remove useful entries while keeping outdated ones. But at least it won't grow forever -- _which is good enough for sure!_

Does this match your understanding of how this cache should work?

randombit: > Does this match your understanding of how this cache should work?

Yes your description does match my own understanding and intended behavior of this cache.

I checked and with 100 K entries using our current schema, the database is 15 MB. (Compare with the 300 MB we allocate for a ccache/sccache cache). At 100K entries we seem unlikely to evict more than ~5 "active" entries at any time given the current active set is under 1K.

> because the cache (for master) is never invalidated. The random eviction might (unluckily) remove useful entries while keeping outdated ones. 

Here your model is somewhat incorrect - the master cache will inevitably be deleted at random intervals, not really under our control in any direct way. Here's what happens:

* GH imposes a cap of 10 GB total cache usage
* If the cache is taking up more storage than this, caches are deleted LRU first
* If someone makes commits in a PR (say 5 or 10) then we'll create many caches, basically `jobs*commits` many.
* If all of the most recently used caches are in one PR, GH will happily delete all of the other caches including the `master` branch caches. [It will even do this if, prior to the deletion, the PR is merged and the branch deleted - it does keep track of which branch created the cache artifact but it does not take this into account when selecting for deletion]

Also if any cache object is not accessed for 7 days, it's deleted. So if things are quiet for a week, cache is cleared.

So honestly doing anything complicated or tricky trying to maximize cache efficiency doesn't seem particularly worthwhile.

randombit:
It works! https://github.com/randombit/botan/actions/runs/15719104875/job/44302771111

src/scripts/dev_tools/run_clang_tidy.py
reneme: Thank you for bearing with my paranoia. I didn't know that the caches are evicted after such a short idle time. That alone is probably "good enough" to deal with this.

PR #4925, opened 2025-06-18
Test: GCC's Stack Scrubbing

### LINKED ISSUE #4922: Automated Tests for GCC's Stack Scrubbing
After introducing basic support for GCC's stack scrubbing (#4882), I think it would be beneficial to have an automated test that verifies that the scrubbing actually works. Being a compiler feature, this doesn't produce any observable side effect within the programming language abstractions. At least as long as we don't venture into undefined-behavior territory.

I used the opportunity to play with GDB's python API and came up with the prototype below. Essentially, this script sets a breakpoint at `SHA_256::compress_digest` (which got annotated for stack scrubbing in #4882). Whenever this breakpoint is hit, the script takes note of the function's entire stack frame and lets the program continue until after `SHA_256::compress_digest` has returned. It then breaks again and checks that said stack frame is filled with zeros.

The good news: this works quite well (albeit making a few assumptions). The bad news: it is awfully slow when applied to the full unit test harness. Not just because it only works in debug mode, but also since it does this check for _every_ invocation of `SHA_256::compress_digest`. And that doesn't even include any other algorithm that is annotated like that.

Anyway I think its a good prototype towards ensuring that the stack scrubbing feature actually delivers what it promised.

Setup:

```bash
# First configure and build the library
./configure.py --build-targets=static,cli \
               --minimized-build          \
               --enable-modules=sha2_32   \
               --cc-bin=g++-14            \
               --enable-stack-scrubbing   \
               --debug-mode # debug mode is essential!
make cli

# Then run the CLI under gdb supplying the python script listed below
gdb -x strubtest.py --batch --args ./botan hash readme.rst
```

This should not produce any warnings. In contrast to a build that was configured without the `--enable-stack-scrubbing` switch. Then you should see:

```
[...]
Error: Botan::SHA_256::compress_digest didn't get its stack frame scrubbed after usage
Error: Botan::SHA_256::compress_digest didn't get its stack frame scrubbed after usage
Error: Botan::SHA_256::compress_digest didn't get its stack frame scrubbed after usage
[...]
```

Here's the python script:

```python
import gdb

def stack_pointer(frame):
    # Currently this will likely work on x86_64 only.
    # TODO: find a better way to do that!
    return frame.read_register('rsp')

def base_pointer(frame):
    # Currently this will likely work on x86_64 only.
    # TODO: find a better way to do that!
    return frame.read_register('rbp')

class PostStrubLocation(gdb.FinishBreakpoint):
    def __init__(self, caller_frame, target_frame, function_name):
        super(PostStrubLocation, self).__init__(caller_frame, True)
        self.function_name = function_name
        self.payload_stack_memory = (stack_pointer(target_frame),
                                     base_pointer(target_frame) - stack_pointer(target_frame))
        assert not self.is_stackframe_scrubbed()

    def stackframe_memory(self):
        return gdb.selected_inferior().read_memory(*self.payload_stack_memory)

    def is_stackframe_scrubbed(self):
        return all(b'\x00' == b for b in self.stackframe_memory())

    def stop(self):
        if not self.is_stackframe_scrubbed():
            print(f"Error: {self.function_name} didn't get its stack frame scrubbed after usage")

class StrubTarget(gdb.Breakpoint):
    def __init__(self, function_name):
        super(StrubTarget, self).__init__(function_name)
        self.function_name = function_name

    def stop(self):
        target_frame = gdb.newest_frame()
        caller_frame = gdb.newest_frame().older()

        assert target_frame.name() == self.function_name
        
        # __attribute__( strub("internal") ) creates a "virtual wrapper" around
        # the annotated function. This wrapper has the same symbol name as the
        # actual target function. To tell them apart, we simply _assume_ that
        # the wrapper is always at the lower address in the binary. This may
        # change in the future or differ across compiler versions!
        if len(self.locations) > 1:
            wrapper_address = min(loc.address for loc in self.locations)
            if wrapper_address == target_frame.pc():
                return False

        # Set a temporary breakpoint after the "virtual wrapper" (caller_frame)
        # has returned. This is just after the stack scrubbing was performed.
        PostStrubLocation(caller_frame, target_frame, self.function_name)

        # Don't stop for interactive inspection at this location
        return False

# Main function
def main():
    StrubTarget("Botan::SHA_256::compress_digest")
    gdb.execute("run")

# Register the main function to be called when the script is executed
gdb.execute("python main()")
```

reneme:
This introduces a new 'strubbing' CI job that builds the library with `--enable-stack-scrubbing --debug-mode` and then uses the gdb extension outlined in #4922 to exercise the currently annotated functions in SHA-256. Unfortunately these tests must be new CI jobs because the library has to be built in debug mode for this to work properly.

I think it makes sense to have this in the main CI while the annotation is being rolled out to more functions. After this is stable, I'd suggest to push the job into the nightly.

The gist of the new CI job is this:

```
python3 /home/runner/work/botan/botan/source/src/scripts/test_strubbed_symbols.py --botan-cli /home/runner/work/botan/botan/build/botan
  Checking Botan::version_string... fail (expected)
  Checking Botan::SHA_256::compress_digest... ok
  Checking Botan::SHA_256::compress_digest_x86... ok
  Checking Botan::SHA_256::compress_digest_x86_avx2... ok
  Checking Botan::SHA_256::compress_digest_x86_simd... ok
```

Currently, this works on x86_64 only. But it should be fairly simple to extend to other architectures, most notably for armv8 I guess. Essentially, the `stack_pointer()` and `base_pointer()` functions in the gdb extension would need to become aware of the architecture and the test script has to select its test cases accordingly.

Future work: it would be cool if we could somehow check that all strubbed symbols are covered. It seems that one can fetch those out of the binary, but at this point I'm not sure whether this is reliable:

```
nm libbotan-3.a | grep strub.0 | c++filt
```

### DIFF
--- a/src/scripts/ci_build.py
+++ b/src/scripts/ci_build.py
@@ -759,6 +759,7 @@ def main(args=None):
             'src/scripts/website.py',
             'src/scripts/bench.py',
             'src/scripts/test_python.py',
+            'src/scripts/test_strubbed_symbols.py',
             'src/scripts/test_fuzzers.py',
             'src/scripts/test_cli.py',
             'src/scripts/repo_config.py',
--- a/src/scripts/ci_build.py
+++ b/src/scripts/ci_build.py
@@ -72,6 +72,7 @@ def known_targets():
         'sde',
         'shared',
         'static',
+        'strubbing',
         'valgrind',
         'valgrind-full',
         'valgrind-ct',
@@ -111,7 +112,7 @@ class LoggingGroup:
 def build_targets(target, target_os):
     if target in ['shared', 'minimized', 'examples', 'limbo'] or target.startswith('policy-'):
         yield 'shared'
-    elif target in ['static', 'fuzzers', 'cross-arm32-baremetal', 'emscripten']:
+    elif target in ['static', 'fuzzers', 'cross-arm32-baremetal', 'emscripten', 'strubbing']:
         yield 'static'
     elif target_os in ['windows']:
         yield 'shared'
@@ -124,7 +125,7 @@ def build_targets(target, target_os):
     if target not in ['examples', 'limbo']:
         yield 'cli'
 
-    if target not in ['examples', 'limbo', 'hybrid-tls13-interop-test']:
+    if target not in ['examples', 'limbo', 'hybrid-tls13-interop-test', 'strubbing']:
         yield 'tests'
 
     if target in ['coverage']:
@@ -267,6 +268,12 @@ def determine_flags(target, target_os, target_cpu, target_cc, cc_bin, ccache,
     if target in ['coverage', 'valgrind', 'valgrind-full', 'valgrind-ct', 'valgrind-ct-full']:
         flags += ['--with-debug-info']
 
+    if target in ['strubbing']:
+        # Stack scrubbing tests are based on scripted gdb runs and won't work on
+        # an optimized build unfortunately.
+        flags += ['--debug-mode']
+        test_cmd = None
+
     if target in ['coverage', 'sanitizer', 'fuzzers']:
         flags += ['--unsafe-terminate-on-asserts', '--enable-modules=tls_null']
 
@@ -304,7 +311,7 @@ def determine_flags(target, target_os, target_cpu, target_cc, cc_bin, ccache,
         # need to find a way to run the wasm-compiled tests w/o a browser
         test_cmd = None
 
-    if target in ['sanitizer'] and target_cc in ['gcc']:
+    if target in ['sanitizer', 'strubbing'] and target_cc in ['gcc']:
         # Stack scrubbing is supported on GCC 14 and newer, only. This is newer
         # than the current default compiler on GHA's Linux image (ubuntu 24.04).
         # The CI setup has to ensure that we are configured to use a recent
@@ -908,6 +915,10 @@ def main(args=None):
                 test_data_arg = []
                 cmds.append([py_interp, os.path.join(root_dir, 'src/scripts', script)] +
                             args + test_data_arg + [botan_exe])
+
+        if target in ['strubbing']:
+            cmds.append([py_interp, os.path.join(root_dir, 'src/scripts/test_strubbed_symbols.py'),
+                         '--botan-cli', os.path.join(build_dir, 'botan')])
 
         if target in ['hybrid-tls13-interop-test']:
             cmds.append([py_interp, os.path.join(root_dir, 'src/scripts/test_cli.py'),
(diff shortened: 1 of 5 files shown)


src/scripts/ci_build.py
```diff
@@ -111,7 +112,7 @@ def __exit__(self, exc_type, exc_value, exc_tb):
 def build_targets(target, target_os):
```
reneme, line 112: Note to self: when #4926 is acceptable, we should disable the 'tests' build target for 'strubbing' (see line 128).

src/scripts/ci_build.py
```diff
@@ -257,6 +258,12 @@ def sanitize_kv(some_string):
     if target in ['coverage', 'valgrind', 'valgrind-full', 'valgrind-ct', 'valgrind-ct-full']:
         flags += ['--with-debug-info']
 
+    if target in ['strubbing']:
+        # Stack scrubbing tests are based on scripted gdb runs and won't work on
+        # an optimized build unfortunately.
+        flags += ['--debug-mode']
```
randombit, line 274: `--debug-mode` is just a shorthand for `--no-optimizations` + `--with-debug-info`, what's happening might be more clear using those directly

src/scripts/gdb/strubtest.py
```diff
@@ -0,0 +1,131 @@
...
+   Success: stackframe of Botan::SHA_256::compress_digest contains 10176 zero bytes after invocation
+   ECD8814EDC180601831FDAD7DEDCF24D57F9F002295346E7D558C941AD318F8E readme.rst
+"""
+
+import gdb
+
+def stack_pointer(frame):
+    # Currently this will likely work on x86_64 only.
+    # TODO: find a better way to do that!
+    return frame.read_register('rsp')
```
randombit: Maybe check (either here or at startup of the script) that we're on x86_64? Otherwise a user is likely to get an error that will be difficult to interpret.

randombit: A friendly LLM suggests `$sp` would work here across machine types

src/scripts/gdb/strubtest.py
```diff
@@ -0,0 +1,131 @@
...
+
+def base_pointer(frame):
+    # Currently this will likely work on x86_64 only.
+    # TODO: find a better way to do that!
+    return frame.read_register('rbp')
+
```
randombit, line 28: `$fp` maybe? (Haven't tested this beyond checking that gdb knows about `$fp` on x86)

randombit (approved):

src/scripts/ci_build.py
reneme: Done.

src/scripts/gdb/strubtest.py
reneme: Works fine, thanks. My regards to the LLM as well, please. 😏

src/scripts/gdb/strubtest.py
reneme: Works on my machine as well (which is also x86).

reneme:
If that passes (with `$sp` and `$fp`) in CI as well, I'll merge and look into an arm64 variant on my Mac, either later today or latest on the weekend.

Issue #4887, opened 2025-05-25
Latest master does not build on MacOS


crazydef:
I just updated my Botan repo to the latest version, but am now encountering a problem when building on MacOS:

```
   INFO: configure.py invoked with options "--library-suffix=.macOS.Debug --no-optimizations --debug-mode --link-method=copy --disable-armv8crypto --disable-neon --cc-abi-flags=-force_cpusubtype_ALL -arch x86_64 -arch arm64 --with-external-includedir=/Users/james/Work/3rdParty/botan/Build/../../zlib/Source --no-install-python-module --disable-shared-library --without-documentation --build-targets=static --includedir=/Users/james/Work/3rdParty/botan/Build/../include/macos --prefix=/Users/james/Work/3rdParty/botan/Build/Output/macos/Debug"
   INFO: Configuring to build Botan 3.9.0 (revision git:7964d1f52209a2f704471b1dc4462db65448cb23)
   INFO: Python version: "3.13.3 (main, Apr  8 2025, 13:54:08) [Clang 16.0.0 (clang-1600.0.26.6)]"
   INFO: Autodetected platform information: OS="Darwin" machine="arm64" proc="arm"
   INFO: Guessing target OS is darwin (use --os to set)
   INFO: Guessing to use compiler xcode (use --cc or CXX to set)
   INFO: Guessing target processor is a arm64 (use --cpu to set)
   INFO: Using /etc/ssl/cert.pem as system certificate store
   INFO: Auto-detected compiler version xcode 16.0
   INFO: Auto-detected compiler arch arm64
   INFO: Target is xcode:16.0-macos-arm64
   INFO: Skipping (dependency failure): asio certstor_sqlite3 chacha_simd32 noekeon_simd serpent_simd sessions_sqlite3 shacal2_simd tpm2_crypto_backend tpm2_ecc tpm2_rsa
[...]

PR #4927, opened 2025-06-18
Remove configure.py options that disable various intrinsics

randombit:
These options are not tested, they in many cases did not work as would be expected, and the use case for any of them is not clear.


reneme:
Note to myself: I think this must be reflected in the Conan package which forwards these switches.

doc/hardware_acceleration.rst
```diff
@@ -191,43 +191,9 @@ On loongarch64, the LSX extensions are used.
...
-   disable POWER Crypto intrinsics
-
-Additionally, ``--disable-modules=MODS`` can be used to remove a certain module,
-if desirable.
-
-Last but not least, the ``BOTAN_CLEAR_CPUID`` :doc:`environment variable <api_ref/env_vars>`
-can be set to a non-empty value *at runtime* to cause Botan to clear the CPUID bits for the CPU
-extensions it uses.
+If it is desirable to avoid using some form of acceleration, this can be accomplished
+at build time by using ``--disable-modules=``.
```
reneme:
```suggestion
If it is desirable to avoid using some form of acceleration, this can be accomplished
*at build time* by using ``--disable-modules=``. For instance, to remove support
of armv8 intrinsics for AES disable ``aes_armv8`` like that at build time.
```

... maybe helpful for someone who never used the module system before.

reneme:
Perhaps we should add basic support for wildcards in the module selection to simplify things [like this](https://github.com/randombit/botan/issues/4887#issuecomment-3014199460). Essentially make it possible to do `--disable-modules='*armv8'`. What do you think, @randombit?

randombit:
> Perhaps we should add basic support for wildcards in the module selection to simplify things https://github.com/randombit/botan/issues/4887#issuecomment-3014199460.

@reneme Yeah that seems fine

PR #4929, opened 2025-06-19
CI: install/simulate TPM2 simulation environment only when needed

reneme:
See https://github.com/randombit/botan/pull/4926#issuecomment-2985738948

This installs the tpm2 simulation packages only if they are actually needed. For clang-tidy we only need the actual libtss to compile against. Especially `tpm2-abrmd` is a culprit here, because it presumably takes forever (more than a minute) to launch background services. Perhaps this can be improved further, but that's for another day.


randombit (approved):
Thanks for looking into this

PR #4934, opened 2025-06-20
Add support for curve specific field element square roots

randombit:
Improves P-384 point decompression and hash2curve by approx 20%


randombit:
Copilot:

> The square-root chain is generated unconditionally. Consider verifying that pcurve['P'] % 4 == 3 or adding a guard comment so this exponentiation is only applied when (p+1)/4 is valid.

I'm honestly impressed it had enough context to pick that up

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
```diff
@@ -1709,17 +1667,17 @@ inline auto map_to_curve_sswu(const typename C::FieldElement& u) -> typename C::
...
+   const auto gx2 = C::x3_ax_b(x2);
 
    // Will be zero if gx1 is not a square
-   const auto [gx1_sqrt, gx1_is_square] = gx1.sqrt();
+   const auto [gx1_sqrt, gx1_is_square] = sqrt_field_element<C>(gx1);
 
    auto x = x2;
    // By design one of gx1 and gx2 must be a quadratic residue
-   auto y = gx2.sqrt().first;
+   auto y = sqrt_field_element<C>(gx2).first;
```
reneme: Nit: at first I was somewhat surprised that this doesn't use a structured binding to take the return value, because I missed the `.first` in the end. Could also be written as: `auto [y, _] = ...` (obviously `_` is just a normal variable name that is ignoreable by convention). I'd love this to use with `auto [y, std::ignore] = ...` but it doesn't. :(

src/lib/math/pcurves/pcurves_impl/pcurves_wrap.h
```diff
@@ -237,11 +237,44 @@ class PrimeOrderCurveImpl final : public PrimeOrderCurve {
...
+            return stash(C::AffinePoint::identity());
+         }
+
+         constexpr size_t FieldElementBytes = C::FieldElement::BYTES;
+         constexpr size_t CompressedBytes = C::FieldElement::BYTES + 1;
+         constexpr size_t UncompressedBytes = 2 * C::FieldElement::BYTES + 1;
+
+         if(bytes.size() == UncompressedBytes && bytes[0] == 0x04) {
+            auto x = C::FieldElement::deserialize(bytes.subspan(1, FieldElementBytes));
+            auto y = C::FieldElement::deserialize(bytes.subspan(1 + FieldElementBytes, FieldElementBytes));
```
reneme: Nit: maybe use `first()` and `last()`?

```suggestion
            const auto encoded_point = bytes.subspan(1);
            auto x = C::FieldElement::deserialize(encoded_point.first(FieldElementBytes));
            auto y = C::FieldElement::deserialize(encoded_point.last(FieldElementBytes));
```

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
randombit: Really this should return a CT::Option - I think that didn't exist when this was originally written

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
```diff
@@ -1709,23 +1684,18 @@ inline auto map_to_curve_sswu(const typename C::FieldElement& u) -> typename C::
...
-   const auto gx2 = C::AffinePoint::x3_ax_b(x2);
-
-   // Will be zero if gx1 is not a square
-   const auto [gx1_sqrt, gx1_is_square] = gx1.sqrt();
 
-   auto x = x2;
    // By design one of gx1 and gx2 must be a quadratic residue
-   auto y = gx2.sqrt().first;
+   const auto y1 = sqrt_field_element<C>(C::x3_ax_b(x1));
+   const auto y2 = sqrt_field_element<C>(C::x3_ax_b(x2));
```
reneme: That may be a bit pedantic; but partially I'd also like to know your opinion on that: Isn't that a place where we should qualify the expected type (i.e. a `CT::Option`) instead of using `auto`? Below, we're using `has_value()` and `value_or()` that both smell like potential side-channel oopsies if these variables would ever become `std::optional<>` for whatever reason.

reneme (approved):

src/lib/math/pcurves/pcurves_impl/pcurves_impl.h
randombit: I guess my default response would be if someone is making that kind of change to sqrt, the problem isn't in this function, it's in the change to sqrt. And if someone is making a kind of side-channel unaware change like that, I don't know that having to update a couple of call sites is going to slow them down. That said eh it's easy enough to spell out the type name this 2 times.

Issue #4928, opened 2025-06-18
Universal binary support on MacOS


ethindp:
Currently, ,building as a universal binary breaks everything. For example if I try with vcpkg and botan (3.7.1, so a bit dated), I get errors like:

```
build/include/internal/botan/internal/simd_32.h:310:19: error: unknown type name 'uint8x16_t'; did you mean 'uint16_t'?
  310 |             const uint8x16_t mask = vld1q_u8(maskb);
      |                   ^~~~~~~~~~
      |                   uint16_t
build/include/public/botan/types.h:99:12: note: 'uint16_t' declared here
   99 | using std::uint16_t;
      |            ^
In file included from /Users/m1/vcpkg/buildtrees/botan/src/3.7.1-3f4b712006.clean/src/lib/block/aes/aes_vperm/aes_vperm.cpp:16:
build/include/internal/botan/internal/simd_32.h:310:37: error: use of undeclared identifier 'vld1q_u8'
  310 |             const uint8x16_t mask = vld1q_u8(maskb);
      |                                     ^
build/include/internal/botan/internal/simd_32.h:311:62: error: use of undeclared identifier 'vreinterpretq_u8_u32'
  311 |             return SIMD_4x32(vreinterpretq_u32_u8(vqtbl1q_u8(vreinterpretq_u8_u32(m_simd), mask)));
      |                                                              ^
build/include/internal/botan/internal/simd_32.h:313:64: error: use of undeclared identifier 'vreinterpretq_u16_u32'
  313 |             return SIMD_4x32(vreinterpretq_u32_u16(vrev32q_u16(vreinterpretq_u16_u32(m_simd))));
      |                                                                ^
build/include/internal/botan/internal/simd_32.h:317:23: error: use of undeclared identifier 'vshlq_n_u32'
  317 |             vorrq_u32(vshlq_n_u32(m_simd, static_cast<int>(ROT)), vshrq_n_u32(m_simd, static_cast<int>(32 - ROT))));
      |                       ^
build/include/internal/botan/internal/simd_32.h:317:67: error: use of undeclared identifier 'vshrq_n_u32'
  317 |             vorrq_u32(vshlq_n_u32(m_simd, static_cast<int>(ROT)), vshrq_n_u32(m_simd, static_cast<int>(32 - ROT))));
      |                                                                   ^
fatal error: too many errors emitted, stopping now [-ferror-limit=]
20 errors generated.
make: *** [build/obj/lib/block_aes_vperm.o] Error 1
[...]

PR #4936, opened 2025-06-21
Improve docs regarding building universal binaries for macOS [ci skip]

randombit:
GH #4928

### DIFF
--- a/doc/building.rst
+++ b/doc/building.rst
@@ -197,19 +197,49 @@ that the Botan libraries were installed into.
 On macOS
 --------------
 
-A build on macOS works much like that on any other Unix-like system.
+A standard build on macOS works much like that on any other Unix-like system.
 
-To build a universal binary for macOS, for older macOs releases,
-you need to set some additional build flags.
-Do this with the `configure.py` flag `--cc-abi-flags`::
+One notable difference with macOS is the common usage of "universal binaries",
+which is effectively a multiarch binary. This was used first for the PowerPC to
+x86 transition, and more recently for the x86 to Aarch64 transition.
 
-  --cc-abi-flags="-force_cpusubtype_ALL -mmacosx-version-min=10.4 -arch i386 -arch ppc"
+Building a universal binary is a bit trickier for Botan compared with a standard
+application, as the library makes use of many architecture specific extensions,
+for example AES-NI and AVX2 on x86, and NEON and the ARMv8 crypto extensions on
+Aarch64. Botan's build system also assumes that it is knowable at setup time
+which files are to be compiled.
 
+Typically (for software with no architecture dependent code) a universal binary
+is built by adding additional compilation flags that look something like
+``-force_cpusubtype_ALL -arch x86_64 -arch arm64``. This effectively causes XCode
+to compile each file twice, once for x86_64 and again for Aarch64. For most source
+files this works fine, but for architecture-specific files it will result in errors
+when code specific to one architecture is encountered when compiling for a different
+architecture, resulting in errors like::
 
-for mac M1 on arm64, you can build the x86_64 arch version via Rosetta separately.
-Do this with with `arch -x86_64 configure.py --library-suffix=-x86_64`
-Then using lipo to create a fat binary.
-`lipo -create libbotan-arm64.dylib libbotan-x86_64.dylib -o libbotan.dylib`
+  $ make
+  ...
+  error: unknown target CPU 'armv8.2-a+sha3'
+  note: valid target CPU values are: ...
+
+There are currently two ways of proceeding.
+
+The first is to use ``--cpu=generic``. This disables all architecture specific
+code, which has performance implications, especially for algorithms with
+dedicated hardware support like AES. This can be alleviated somewhat by making
+sure the CommonCrypto provider (module ``commoncrypto``) is built, since then
+Botan offloads many of these specific operations to CommonCrypto, which will be
+able to use the CPU instructions.
+
+The second, and recommended, approach is to build twice and use ``lipo`` to
+combine the two binaries. This looks something like::
+
+  $ ./configure.py --build-dir=botan_x86_64 --cpu=x86_64 --extra-cxxflags='-arch x86_64' --library-suffix=-x86_64
+  $ make -j8 -f botan_x86_64/Makefile
+  $ ./configure.py --build-dir=botan_aarch64 --cpu=aarch64 --extra-cxxflags='-arch armv8' --library-suffix=-aarch64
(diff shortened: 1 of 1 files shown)


doc/building.rst
```diff
@@ -197,19 +197,49 @@ that the Botan libraries were installed into.
...
+
+The second, and recommended, approach is to build twice and use ``lipo`` to
+combine the two binaries. This looks something like::
+
+  $ ./configure.py --build-dir=botan_x86_64 --cpu=x86_64 --extra-cxxflags='-arch x86_64' --library-suffix=-x86_64
+  $ make -j8 -f botan_x86_64/Makefile
+  $ ./configure.py --build-dir=botan_aarch64 --cpu=aarch64 --extra-cxxflags='-arch armv8' --library-suffix=-aarch64
+  $ make -j8 -f botan_aarch64/Makefile
+  # Now combine the two binaries with lipo
+  $ lipo -create botan_aarch64/libbotan-aarch64.dylib botan_x86_64/libbotan-x86_64.dylib -o libbotan.dylib
```
reneme:
```suggestion
$ ./configure.py --with-build-dir=botan_x86_64 --disable-cc-tests --build-targets=shared  --cpu=x86_64 --extra-cxxflags='-arch x86_64' --ldflags='-arch x86_64' --library-suffix=-x86_64
$ make -j8 -f botan_x86_64/Makefile
$ ./configure.py --with-build-dir=botan_aarch64 --disable-cc-tests --build-targets=shared  --cpu=aarch64 --extra-cxxflags='-arch arm64' --ldflags='-arch arm64' --library-suffix=-aarch64
$ make -j8 -f botan_aarch64/Makefile
$ lipo -create botan_aarch64/libbotan-3-aarch64.dylib botan_x86_64/libbotan-3-x86_64.dylib -o libbotan-3.dylib
```

Key changes:

* `--disable-cc-tests`, otherwise the automatic target detection will complain that you explicitly ordered x86 but the compiler wants to emit arm64 (or vice versa). That is because the `--extra-cxxflags` aren't forwarded to `run_compiler_preproc` in _configure.py_
* `--ldflags='-arch ...'`, otherwise the linker will expect to create a shared object for the host platform and won't work for the cross-compile part
* `-arch arm64`, `armv8` isn't accepted by my machine at least
* The shared objects are called `libbotan-3.dylib` (the "-3" suffix was missing)

Regarding `--disable-cc-tests`: passing the `--extra-cxxflags` to the auto-detection is fairly easy, one just has to handle the command line separation by spaces properly. Below is a patch (I could also make a PR for that). I'm not sure that it is the best strategy to solve this, though.

```patch
diff --git a/configure.py b/configure.py
index 9062a0b08..23294bbb6 100755
--- a/configure.py
+++ b/configure.py
@@ -3409,7 +3409,8 @@ def check_compiler_arch(options, ccinfo, archinfo, source_paths):
     detect_version_source = os.path.join(source_paths.build_data_dir, 'detect_arch.cpp')
 
     abi_flags = ccinfo.mach_abi_link_flags(options).split(' ')
-    cc_output = run_compiler_preproc(options, ccinfo, detect_version_source, 'UNKNOWN', abi_flags).lower()
+    extra_flags = [flag for flags in options.extra_cxxflags for flag in flags.split(' ')]
+    cc_output = run_compiler_preproc(options, ccinfo, detect_version_source, 'UNKNOWN', abi_flags + extra_flags).lower()
 
     if cc_output == '':
         cc_output = run_compiler_preproc(options, ccinfo, detect_version_source, 'UNKNOWN').lower()
```

reneme (approved):
I refined the script template after trying it out on my MacBook. Frankly, I think our cross-compiling story isn't great unfortunately. One has to really study our _configure.py_ to figure this out, especially when using Clang.

For gcc I would simply provide the right cross-compiler binary and wouldn't need to care much about any cross-compile related flags. For clang, given that a single compiler/linker can build for multiple targets, one has to figure out how _configure.py_ passes those flags to the compiler/linker.

It would be really awesome if the build system could take over this burden. Essentially detect that the user wants to cross-compile and pass the right arguments to Clang automagically. Not sure how much extra complexity this would bring, though.

mouse07410:
I would appreciate if you make sure that it **does not** attempt to build Universal binaries on MacOS **by default**.

If there's an explicit option to do so - fine. But but automatically, please.

reneme:
> I would appreciate if you make sure that it does not attempt to build Universal binaries on MacOS by default.

Surely not. I'm not a big fan of universal binaries either. The approach documented in this pull request is generally sound: i.e. two individual builds, one of them cross-compiling to x86 or ARM (depending on your host platform).

I just think it would be beneficial to make the actual cross-compilation a bit easier. Other users could also benefit from this: for instance when cross-compiling for some embedded system.

randombit:
> Frankly, I think our cross-compiling story isn't great unfortunately.

I do find it interesting you say that as I find the cross compile process really seamless (I do aarch64 and armv7 cross builds all of the time) but maybe the difference is that I'm always doing it with GCC. I'm not sure I've ever even tried cross compiling with Clang.

randombit:
> I'm not a big fan of universal binaries either. 

I find the whole thing a little silly myself. Apple could have just implemented binary translation in each direction, x86 can run aarch64, aarch64 can run x86. Done. Instead they only did the translation in one direction (aarch64 running x86) which meant everyone has to do this universal binary thing to support x86 users. And then they drop running x86 binaries at the same time as dropping support for x86 hardware?? I don't get it.

Anyway in, what, 2 years the problem goes away and now that Apple controls their own destiny it's hard to imagine they will switch architectures again anytime soon. RISC-V doesn't offer them anything they can't get with aarch64 considering they are building the chips custom to their use case.

reneme:
> maybe the difference is that I'm always doing it with GCC. I'm not sure I've ever even tried cross compiling with Clang.

Same here. With GCC indeed it's fine: basically set the right binary and be done with it. For clang, getting the configure flags right was a bit of a pain, and I know configure.py inside out. For someone who doesn't, it's going to be a big hurdle.

In a best case scenario, as a user, I'd probably expect that `--cc clang --cpu armv8` should just do the trick. But it's far from true.

randombit:
> In a best case scenario, as a user, I'd probably expect that --cc clang --cpu armv8 should just do the trick. But it's far from true.

OK well that's worth looking into. I'll perhaps give Clang cross builds a try this weekend and see what it looks like. Could also perhaps add some Clang cross builds to CI  - I think the main constraint on the cross builds we have now are which GCC cross compilers are included in Ubuntu, while (AIUI) Clang ships out of the box supporting all architectures.

PR #4940, opened 2025-06-22
Range and span were used in line with TODO messages.

KaganCanSit:
Hello,

I was looking through the "Todo" messages to contribute to the project. I fulfilled the requirements of a few messages about C++20 under the test folder. However, I do not use C++20 in my work. Maybe I have a mistake. I saw that the tests were successful with the following command.

```bash
./botan-test --test-threads=4 --run-long-tests
```

You can contact me for the points you want to change or improve. I would like to help. 🤝

Have a good day.

### DIFF
--- a/src/tests/test_strong_type.cpp
+++ b/src/tests/test_strong_type.cpp
@@ -15,6 +15,7 @@
 
 #include <algorithm>
 #include <array>
+#include <ranges>
 #include <sstream>
 #include <string>
 #include <vector>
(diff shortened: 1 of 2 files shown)


src/tests/test_strong_type.cpp
```diff
@@ -157,8 +158,7 @@ std::vector<Test::Result> test_container_strong_type() {
 
                Test_Length_List hashes({Test_Length(3), Test_Length(1), Test_Length(4), Test_Length(2)});
 
-               // TODO: C++20 - std::ranges::sort
-               std::sort(hashes.begin(), hashes.end());
+               std::ranges::sort(hashes);
```
reneme, line 161: So far I didn't use the ranges algorithms because many aspects of the ranges library didn't work (well) with some older compiler we still have to support. Perhaps the situation is better now, or maybe `std::ranges::sort` doesn't fall in this category. If CI is green, I guess its fine. 😏

reneme (approved):

src/tests/test_strong_type.cpp
KaganCanSit: If @randombit deems it appropriate, it can stay. Otherwise, I can change it. No problem. Thanks for the quick response.

randombit: If it compiles in CI, that hits all of our min supported compilers so not problem on my side

randombit (approved):

PR #4942, opened 2025-06-22
Fix/reduce Cppcheck warnings in test code

KaganCanSit:
Hello,

I reviewed the warnings reported by cppcheck under the test folder, which I think are not false positives. I tried to make corrections to the appropriate ones. The messages I reviewed and intervened are listed below:

* [botan/src/tests/test_pad.cpp:33] (performance) Ineffective call of function 'substr' because a prefix of the string is assigned to itself. Use resize() or pop_back() instead. [uselessCallsSubstr]
* [botan/src/tests/test_ec_group.cpp:115] (style) The scope of the variable 'blind_ws' can be reduced. Warning: Be careful when fixing this message, especially when there are inner loops. When you see this message it is always safe to reduce the variable scope to 1 level. [variableScope]
* [botan/src/tests/test_hash.cpp:196] (style) Variable 'buf' is assigned a value that is never used. [unreadVariable] - !!! I think this change and function needs to be examined.
* [botan/src/tests/test_pkcs11_high_level.cpp:1067] (style) Variable 'curve' can be declared as reference to const [constVariable] 
[botan/src/tests/test_pkcs11_high_level.cpp:1092] (style) Variable 'curve' can be declared as reference to const [constVariable]
* [botan/src/tests/test_pkcs11_low_level.cpp:389] (style) Local variable 'label' shadows outer variable [shadowVariable]
* [botan/src/tests/test_tls_rfc8448.cpp:253] (style) Consider using std::find_if algorithm instead of a raw loop. [useStlAlgorithm]
* [botan/src/tests/test_uri.cpp:48] (style) Variable 'tests' can be declared as const array [constVariable]
* [botan/src/tests/test_utils.cpp:1012] (style) Consider using std::transform algorithm instead of a raw loop. [useStlAlgorithm]
* [botan/src/tests/test_utils.cpp:1329] (style) Consider using std::fill algorithm instead of a raw loop. [useStlAlgorithm]

I did not make any changes for the following warnings:
- 'explicit' and 'is not initialized in the constructor' messages. (I can try to create a separate PR request by reviewing this if desired.)
- Situations where I think there will be no significant benefit in using the algorithm in various raw loops.

Compile:
```bash
python3 src/scripts/dev_tools/run_clang_format.py --clang-format=clang-format-17 --src-dir=src && ninja clean && ./configure.py --without-documentation --with-boost --cc=clang --compiler-cache=ccache --build-targets=static,cli,tests --build-tool=ninja && ninja
```

Test:
```bash
./botan-test --test-threads=4 --run-long-tests && python3 src/scripts/test_cli.py ./botan cli_tls_socket_tests
```

You can contact me for the points you want to change or improve. I would like to help.

Have a nice day.

### DIFF
--- a/src/tests/test_pad.cpp
+++ b/src/tests/test_pad.cpp
@@ -30,7 +30,7 @@ class Cipher_Mode_Padding_Tests final : public Text_Based_Test {
             if(algo.substr(underscore + 1, std::string::npos) != "Invalid") {
                throw Test_Error("Unexpected padding header " + header);
             }
-            algo = algo.substr(0, underscore);
+            algo.resize(underscore);  // Use just the part before the underscore
          }
 
          Test::Result result(algo);
--- a/src/tests/test_uri.cpp
+++ b/src/tests/test_uri.cpp
@@ -40,17 +40,19 @@ class URI_Tests final : public Test {
       static Test::Result test_uri_parsing() {
          Test::Result result("URI parsing");
 
-         struct {
+         struct URITestCase {
                std::string uri;
                std::string host;
                Botan::URI::Type type;
                uint16_t port;
-         } tests[]{
-            {"localhost:80", "localhost", Botan::URI::Type::Domain, 80},
-            {"www.example.com", "www.example.com", Botan::URI::Type::Domain, 0},
-            {"192.168.1.1", "192.168.1.1", Botan::URI::Type::IPv4, 0},
-            {"192.168.1.1:34567", "192.168.1.1", Botan::URI::Type::IPv4, 34567},
-            {"[::1]:61234", "::1", Botan::URI::Type::IPv6, 61234},
+         };
+
+         const std::array tests{
+            URITestCase{"localhost:80", "localhost", Botan::URI::Type::Domain, 80},
+            URITestCase{"www.example.com", "www.example.com", Botan::URI::Type::Domain, 0},
+            URITestCase{"192.168.1.1", "192.168.1.1", Botan::URI::Type::IPv4, 0},
+            URITestCase{"192.168.1.1:34567", "192.168.1.1", Botan::URI::Type::IPv4, 34567},
+            URITestCase{"[::1]:61234", "::1", Botan::URI::Type::IPv6, 61234},
          };
 
          for(const auto& t : tests) {
--- a/src/tests/test_utils.cpp
+++ b/src/tests/test_utils.cpp
@@ -1327,9 +1327,7 @@ class UUID_Tests : public Test {
                explicit AllSame_RNG(uint8_t b) : m_val(b) {}
 
                void fill_bytes_with_input(std::span<uint8_t> output, std::span<const uint8_t> /* ignored */) override {
-                  for(auto& byte : output) {
-                     byte = m_val;
-                  }
+                  std::fill(output.begin(), output.end(), m_val);
                }
 
                std::string name() const override { return "zeros"; }
--- a/src/tests/test_ec_group.cpp
+++ b/src/tests/test_ec_group.cpp
@@ -112,9 +112,8 @@ std::vector<Test::Result> ECC_Randomized_Tests::run() {
 
       const Botan::EC_Point pt = create_random_point(this->rng(), group);
 
-      std::vector<Botan::BigInt> blind_ws;
-
       try {
+         std::vector<Botan::BigInt> blind_ws;
          const size_t trials = (Test::run_long_tests() ? 10 : 3);
          for(size_t i = 0; i < trials; ++i) {
             const Botan::BigInt a = test_integer(rng(), group.get_order_bits(), group.get_order());
--- a/src/tests/test_hash.cpp
+++ b/src/tests/test_hash.cpp
@@ -193,8 +193,6 @@ class Hash_NIST_MonteCarlo_Tests final : public Text_Based_Test {
             input.push_back(seed);
             input.push_back(seed);
 
-            std::vector<uint8_t> buf(hash->output_length());
-
             for(size_t j = 0; j <= count; ++j) {
                for(size_t i = 3; i != 1003; ++i) {
                   hash->update(input[0]);
--- a/src/tests/test_pkcs11_high_level.cpp
+++ b/src/tests/test_pkcs11_high_level.cpp
@@ -1064,7 +1064,7 @@ Test::Result test_ecdsa_generate_keypair() {
    curves.push_back("secp256r1");
    curves.push_back("brainpool512r1");
 
-   for(auto& curve : curves) {
+   for(const auto& curve : curves) {
       PKCS11_ECDSA_KeyPair keypair = generate_ecdsa_keypair(test_session, curve, EC_Group_Encoding::NamedCurve);
 
       keypair.first.destroy();
@@ -1089,7 +1089,7 @@ Test::Result test_ecdsa_sign_verify_core(EC_Group_Encoding enc, const std::strin
 
    auto rng = Test::new_rng(__func__);
 
-   for(auto& curve : curves) {
+   for(const auto& curve : curves) {
       // generate key pair
       PKCS11_ECDSA_KeyPair keypair = generate_ecdsa_keypair(test_session, curve, enc);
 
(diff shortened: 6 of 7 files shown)


src/tests/test_uri.cpp
```diff
@@ -45,7 +45,9 @@ class URI_Tests final : public Test {
                std::string host;
                Botan::URI::Type type;
                uint16_t port;
-         } tests[]{
+         }
+
+         const tests[]{
```
reneme: If we split this apart, I would suggest to give the struct a name, add a semicolon at the end of its declaration and use the type in the declaration of the std::array. To my mind, that boils down to personal preference, whether to use anonymous structs like that or not.

src/tests/test_utils.cpp
```diff
@@ -1006,11 +1006,9 @@ class Date_Format_Tests final : public Text_Based_Test {
...
-         std::vector<uint32_t> u32s;
-         u32s.reserve(parts.size());
-         for(const auto& sub : parts) {
-            u32s.push_back(Botan::to_u32bit(sub));
-         }
+         std::vector<uint32_t> u32s(parts.size());
+         std::transform(
+            parts.begin(), parts.end(), u32s.begin(), [](const std::string& sub) { return Botan::to_u32bit(sub); });
          return u32s;
       }
```
reneme: Such use of STL algorithms (especially pre-ranges) are really debatable in my opinion. Yes, using `std::transform`, `std::find_if`, `std::for_each`, and friends is more idiomatic. And there are plenty C++ connoisseurs that want to eliminate raw loops as much as possible. Though, I think we should ask ourselves: is that really easier to read than a range-based for loop?

For the record: I'm essentially indifferent about this. But I think its not worth putting in the effort to migrate to these pre-ranges STL algorithms. Once we shed support for some older compilers and can actually use the full potential of ranges, that's an entirely different discussion, in my opinion.

This would be C++23 ([godbolt](https://godbolt.org/z/a5xTEa7xj)):

```C++
namespace views = std::ranges::views;
namespace ranges = std::ranges;

auto u32s = parts | views::transform([](auto&& v) { return Botan::to_u32bit(v); })
                  | ranges::to<std::vector>();
```

... now, that would be real improvement! No more iterators, no more manual pre-allocation of the vector, no more raw loop, and not even explicit value types anymore. 🤩

src/tests/test_tls_rfc8448.cpp
```diff
@@ -249,10 +249,10 @@ class Test_TLS_13_Callbacks : public Botan::TLS::Callbacks {
...
 
-         for(const auto& mock : m_mock_signatures) {
-            if(mock.message_to_sign == msg) {
-               return mock.signature_to_produce;
-            }
+         const auto found =
+            std::ranges::find_if(m_mock_signatures, [&msg](const auto& mock) { return mock.message_to_sign == msg; });
+         if(found != m_mock_signatures.end()) {
+            return found->signature_to_produce;
          }
```
reneme: See my other comment about pre-ranges STL algorithms.

src/tests/test_utils.cpp
KaganCanSit: Frankly, there were similar warnings to this situation. Sometimes I think they made me think about readability. That's why I only took the ones that exactly match the call. Here, I leave the decision of whether to implement it or not to implement it to you.

The usage you gave as an example really excited me. I'll look into it. But as far as I understand, C++20 doesn't support this usage.

src/tests/test_uri.cpp
KaganCanSit: I understand, I wanted to implement it array based at first, but I thought it might not be welcomed because there would be too many changes.

I will give it a try after this message.

Thanks for the quick turnaround.

src/tests/test_tls_rfc8448.cpp
randombit: Nak on this one, I find these STL algo calls way less readable than a simple loop

src/tests/test_pad.cpp
```diff
@@ -30,7 +30,7 @@ class Cipher_Mode_Padding_Tests final : public Text_Based_Test {
             if(algo.substr(underscore + 1, std::string::npos) != "Invalid") {
                throw Test_Error("Unexpected padding header " + header);
             }
-            algo = algo.substr(0, underscore);
+            algo.resize(underscore);
```
randombit: These are semantically identical but the original is more clear about what it's doing. Fine I guess but add a comment eg // use just the part before the underscore

src/tests/test_uri.cpp
randombit: This struct should definitely have a name, a few tests use these anon structs and I don't like it much.

(We could also consider pulling it all the way out to a vec file, idk)

src/tests/test_utils.cpp
randombit: Yeah the example reneme posted is nice enough, similar to Rust style iterators. But this one should be reverted.

KaganCanSit:
@randombit @reneme Thank you very much for your quick feedback. I apologize for confusing the commit name while managing multiple branches. I corrected the PR and comment name.

* Rebased to master branch.
* STL algorithm calls have been rolled back as requested.
* Renamed the struct structure and used std::array.
* added a comment line for code readability at the point where resize is selected for performance instead of substr.

I hope there is no point I missed. I will follow CI.

src/tests/test_utils.cpp
reneme: > But as far as I understand, C++20 doesn't support this usage.

C++20 has ranges (just not all compilers we currently support). Chances are that Botan 4 will be able to use much of this. The only proper C++23-ism in my example above is `ranges::to<>`. But there's (afaik) no technical reason why it couldn't be custom built in C++20 until we adopt C++23 in the future.

KaganCanSit: > C++20 has ranges (just not all compilers we currently support). Chances are that Botan 4 will be able to use much of this. The only proper C++23-ism in my example above is `ranges::to<>`. But there's (afaik) no technical reason why it couldn't be custom built in C++20 until we adopt C++23 in the future.

Thanks for sharing your knowledge with me. I will try this approach in a small test code.

randombit (approved):
LGTM thanks

src/tests/test_uri.cpp
```diff
@@ -40,18 +40,20 @@ class URI_Tests final : public Test {
...
+         };
+
+         const std::array<URITestCase, 5> tests{{
             {"localhost:80", "localhost", Botan::URI::Type::Domain, 80},
             {"www.example.com", "www.example.com", Botan::URI::Type::Domain, 0},
             {"192.168.1.1", "192.168.1.1", Botan::URI::Type::IPv4, 0},
             {"192.168.1.1:34567", "192.168.1.1", Botan::URI::Type::IPv4, 34567},
             {"[::1]:61234", "::1", Botan::URI::Type::IPv6, 61234},
-         };
+         }};
```
reneme: Nit: I think it is worthwhile to avoid specifying the size of the array for such an initializer-list-style. C++ (I believe since 17, don't quote me on this), allows leaving the template parameters out if they can be deduced from the context. Unfortunately that is -- to the best of my knowledge -- an all-or-nothing approach, though.

I.e. you could rephrase this to:

```C++
         const std::array tests{
            URITestCase{"localhost:80", "localhost", Botan::URI::Type::Domain, 80},
            URITestCase{"www.example.com", "www.example.com", Botan::URI::Type::Domain, 0},
            URITestCase{"192.168.1.1", "192.168.1.1", Botan::URI::Type::IPv4, 0},
            URITestCase{"192.168.1.1:34567", "192.168.1.1", Botan::URI::Type::IPv4, 34567},
            URITestCase{"[::1]:61234", "::1", Botan::URI::Type::IPv6, 61234},
         };
```

... that forces you to repeat the inner type (i.e. `URITestCase`) for each element. Though, to my mind, that makes the whole thing actually easier to parse because now it is quite obviously a list of `URITestCase`-s. And, adding/removing things from this list makes for a clean patch, FWIW.

I'll let you be the judge of this.

KaganCanSit: Unfortunately this definition is invalid in C++11. I have used it here based on my own habits. 😢 I will address this in the direction you suggest.

I change this code.

PR #4946, opened 2025-06-24
Use concepts to constrain the allowed types of various integer-related templates

randombit:


src/lib/block/serpent/serpent_fn.h
```diff
@@ -10,9 +10,19 @@
...
 namespace Botan::Serpent_F {
 
+// Concept for types that support bitwise operations (unsigned integers or SIMD types)
+template <typename T>
+concept BitsliceT = std::unsigned_integral<T> || requires(T& a, const T& b) {
+   a ^= b;
+   a &= b;
+   a |= b;
+   ~a;
+};
```
reneme: Properly nitpicking here, but: `std::unsigned_integral` can do all the things you spelled out in the requires clause. You might as well leave it out as an explicit option. 🤷  Also: the `T`-suffix is somewhat against typical convention where concepts have more "natural" names. Maybe "bislicable" (what a mess), or "bitwise_operand", meh. Not sure.

reneme (approved):
Conceptualize all the things! 😄

src/lib/block/serpent/serpent_fn.h
randombit: Right but its an or here since the simd wrappers are not integral types

reneme: I meant to leave out `std::unsigned_integral` and rely on the operator requirements alone.

randombit: Ah got it sorry

PR #4947, opened 2025-06-27
FIX: Building a static library using Ninja on macOS

reneme:
For some reason the 'ar' command on macOS does not support @response_files. We have to revert back to providing the .o files via the command line there. Otherwise builds with `./configure.py --build-targets=static --build-tool=ninja` would fail (and [did since 3.6.0](https://github.com/randombit/botan/pull/4350)) on macOS. Fortunately no-one except me is trying to use Ninja on macOS apparently. ;)

We're not the only ones suffering from this by the way: https://gitlab.kitware.com/cmake/cmake/-/issues/16731

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -2241,6 +2241,10 @@ def create_template_vars(source_paths, build_paths, options, modules, disabled_m
         'makefile_path': os.path.join(build_paths.build_dir, '..', 'Makefile'),
         'ninja_build_path': os.path.join(build_paths.build_dir, '..', 'build.ninja'),
 
+        # Use response files for the archive command on windows
+        # Note: macOS (and perhaps other OSes) do not support this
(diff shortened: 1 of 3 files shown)


configure.py
```diff
@@ -2236,6 +2236,9 @@ def test_exe_extra_ldflags():
         'makefile_path': os.path.join(build_paths.build_dir, '..', 'Makefile'),
         'ninja_build_path': os.path.join(build_paths.build_dir, '..', 'build.ninja'),
 
+        # macOS does not support response files for the ar command.
```
randombit: Are we sure other `ar`s *do* support response files? OpenBSD, AIX, Solaris, ... ? IMO might be better to just say use it if on Linux otherwise don't.

reneme: Fair point. I changed it to use response files on Windows only. It was initially introduced because Windows had trouble with super long cmd lines.

PR #4955, opened 2025-07-03
Document the specific threat model the library operates under

randombit:


doc/threat_model.rst
```diff
@@ -0,0 +1,54 @@
...
+and precisely measure each conditional jump and memory access.
+
+This also covers the (weaker) threat model of an attacker on the same LAN who is
+performing attacks based purely on timing of operations.
+
+Wherever possible, code that manipulates secret data (for example when generating an ECDSA
+signature or decrypting an AES ciphertext) is written to be "constant time"; avoiding any
+conditional jumps or memory accesses where the predicate is (derived from) secret
+information. Botan uses extensive annotations (``CT::poison``) to indicate which values
+are secret, and uses automated analysis (currently using ``valgrind``, though support for
```
atreiber94:
```suggestion
are secret, and uses automated analysis (currently using ``valgrind`` similar to Adam Langley's ``ctgrind`` idea, though support for
```

doc/threat_model.rst
```diff
@@ -0,0 +1,54 @@
...
+other tools is welcome) to verify that the assembly created by the compiler in fact avoids
+all conditional jumps. This testing step is essential as some compilers (notably Clang)
```
atreiber94:
```suggestion
all secret-dependend conditional jumps or memory accesses. This testing step is essential as some compilers (notably Clang)
```

doc/threat_model.rst
```diff
@@ -0,0 +1,54 @@
...
+
+Some algorithms have a structure which allows for very practical blinding/re-randomization
+of the operations. This is used as an additional countermeasure in case some particular
+combination of compiler, compiler options, and target architecture results in a
+conditional jump being inserted in an unexpected place. For example during ECDSA signing,
+the inversion of ``k``, the scalar multiplication of ``g*k`` and the recombination of
+``x * r + m`` are all blinded, even though all of the relevant arithmetic operations are
+written and tested to avoid side channels.
+
+For more about specific side channel countermeasures, see :ref:`side_channels`.
```
atreiber94, line 43: I would also suggest to add a sentence making it very clear and explicit that it's kind of inherent that no full side channel resistance can be guaranteed. E.g. like this:   
```suggestion
For more about specific side channel countermeasures, see :ref:`side_channels`.

While these best-effort countermeasures and analysis tools are in place, the absence of any kind of side channel cannot be guaranteed by a software library like Botan.
```
This is just according to my taste though; the document is fine without it.

atreiber94 (approved):
Looks good, I think such a discussion of the threat model is very helpful for the documentation.

doc/threat_model.rst
randombit: Yeah this is absolutely the case and worth mentioning. I think I'll reword it somewhat but committing your suggested change as a start.

PR #4966, opened 2025-07-09
Support for wildcards in module selection

reneme:
As suggested in https://github.com/randombit/botan/pull/4927#issuecomment-3014202670

This allows using an asterisk as a wildcard in module selection. E.g. `./configure.py --disable-modules='*armv8'` would (by convention) disable all armv8-specific implementation modules. Currently, a wildcard always needs to match at least a single character. That's an arbitrary decision towards the more restrictive approach. If the wildcard doesn't match any modules, a warning is printed but the configuration _is not aborted_.


reneme:
> Maybe mention it in the docs?

... I thought about it, but the plane was about to land. 😏 Added now.

Also (hopefully) fixed the CI where `--enable-modules=tls*` enabled `tls_cbc` which is prohibited in the targeted build policies. Unfortunately, accompanying the wildcard with a `--disable-modules=tls_cbc` didn't do the trick. Maybe that's for the next flight. 🤞

PR #4970, opened 2025-07-12
Test: Improve AEAD and cipher mode error condition checks

reneme:
This is extracted from [the finding I came across yesterday](https://github.com/randombit/botan/pull/4880#issuecomment-3062846755) and **improves the negative tests in AEAD and cipher mode test cases**. Instead of just expecting *any* exception for the negative tests, the checks now explicitly specify which exception type they expect.

Turns out that some tests actually triggered on an `Invalid_Argument` (because `finish()` was called with insufficient input bytes) instead of the `Invalid_State` (for calling `finish()` before setting a key). I adapted these tests so that they pass the internal argument check and actually trigger the expected error condition.

Additionally, the more specific negative checks revealed that in `EAX_Decryption` a "key is set" state check was missing. Before, this cipher object claimed `Invalid_Authentication_Tag` on `finish()` even though a key was never specified on the cipher mode object. Similarly, an order of validation in `start()` had to be reversed in the CommonCrypto cipher mode adapter class.

For such negative check setups to be a little more flexible, I adapted the implementation of `Result::throws_as<ExpectedExceptionT>` slightly: Earlier the thrown exception type had to match exactly; now it is enough when it is derived from the expected exception. With that, negative tests that expect an `Invalid_State` exception still succeed if the code under test throws a `Key_Not_Set` which is derived from `Invalid_State`.

### DIFF
--- a/src/lib/modes/aead/eax/eax.cpp
+++ b/src/lib/modes/aead/eax/eax.cpp
@@ -154,6 +154,7 @@ size_t EAX_Decryption::process_msg(uint8_t buf[], size_t sz) {
 }
 
 void EAX_Decryption::finish_msg(secure_vector<uint8_t>& buffer, size_t offset) {
+   BOTAN_STATE_CHECK(!m_nonce_mac.empty());
    BOTAN_ARG_CHECK(buffer.size() >= offset, "Offset is out of range");
    const size_t sz = buffer.size() - offset;
    uint8_t* buf = buffer.data() + offset;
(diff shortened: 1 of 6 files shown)


randombit (approved):
lgtm thanks

reneme:
> lgtm thanks

Hang on, commoncrypto isn't up to the spec yet. Working on that. :)

reneme:
The [fix for commoncrypto](https://github.com/randombit/botan/pull/4970/commits/dd50bef879f8e7c76057fa6deffbda572cad0d17) was simple and I'd consider it an artifact of the tighter negative test expectation: Essentially the test expects `Invalid_Argument` when setting an invalid nonce on an unkeyed object, but the commoncrypto wrapper first performed a "key is set?" check. I simply reversed both checks.

... merging as soon as the CI comes back clean. Thanks for the ultra-swift approval!

PR #4971, opened 2025-07-12
Run: `run_clang_tidy.py --regenerate-inline-config-file`

reneme:
... this way I see the new clang-tidy stuff in my IDE. 😄 

Screenshot 2025-07-12 at 17 54 01


randombit:
Oops, thanks. There needs to be a CI check that the script and config are consistent

randombit (approved):

PR #4991, opened 2025-07-15
Scrub pcurves ECC private key values on destruction

reneme:
This adds a destructor to `EC_PrivateKey_Data` ensuring that the private scalar is cleaned up during destruction.

### DIFF
--- a/src/lib/pubkey/ec_group/ec_scalar.h
+++ b/src/lib/pubkey/ec_group/ec_scalar.h
@@ -204,6 +204,13 @@ class BOTAN_PUBLIC_API(3, 6) EC_Scalar final {
       void assign(const EC_Scalar& x);
 
       /**
+      * Equivalent to assigning a zero value, but also does so in a way that
+      * attempts to ensure the write always occurs even if a compiler can deduce
+      * the assignment is otherwise unnecessary.
+      */
+      void zeroize();
+
+      /**
       * Set *this to its own square modulo the group order
       */
       void square_self();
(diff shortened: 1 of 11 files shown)


src/lib/pubkey/ec_group/ec_scalar.h
```diff
@@ -203,6 +203,11 @@ class BOTAN_PUBLIC_API(3, 6) EC_Scalar final {
       */
       void assign(const EC_Scalar& x);
 
+      /**
+      * Destroy the value stored in this scalar
+      */
+      void zap();
```
randombit: Should this (and the associated other functions) be named instead `zeroize`? There is a `zap` in secmem.h but I'm not sure this is really standard/common usage.

Also I think the documentation here could be more clear. Does a destructor somewhere get invoked? Can the `EC_Scalar` be reused after calling this function? If it can be reused, is the value after the function returns deterministic?

Maybe

> Zeroize this EC_Scalar.
>
> Equivalent to assigning a zero value, but also does so in a way that attempts to ensure the write always occurs even if a compiler can deduce the assignment is otherwise unnecessary.

randombit (commented):
Basically fine. Could use a test that `zap`/`zeroize` actually sets the expected value.  `ECC_Scalar_Arithmetic_Tests` is probably the place to put it, there we already do a hack of `1 - 1` since there is no `EC_Scalar::zero`.

randombit:
Oh requesting also a change to commit message / PR title "Cleanup" makes it sound like this is related to a code cleanup, vs a zeroization issue.

reneme:
I added a test and renamed the function to `zeroize()`. Note that I changed both the implementations: For `BigInt` I used `std::swap` with a default-constructed `BigInt` to be sure that the destructor of the big int's `secure_vector` is called; for pcurves I added a `std::fill(..., 0)` after the `secure_scrub_memory()` because the latter does not guarantee "zeroization".

PR #4994, opened 2025-07-15
Replace std::array with std::span in roughtime hashNode function

KaganCanSit:
This optimization was spotted by @reneme in review PR #4783. While there are more instances of `typecast_copy' in this codebase, I wanted to improve it with a small change for this part only.

Referenced review comment: https://github.com/randombit/botan/pull/4783#discussion_r2202686176

Quote:
> General remark (please don't tackle in this pull request): This roughtime implementation makes heavy use of std::array<> in places where (nowadays) it would be better off to use statically-sized std::span.
> 
> Here, hashNode is a good example: if it would be implemented using spans, you could avoid the data copies and simplify the code:
> 
> ```cpp
>    BufferSlicer slicer(path);
>    for(size_t level = 0; level < levels; ++level) {
>       hashNode(hash, slicer.take<64>(), index & 1);
>       index >>= 1;
>    }
> ```
> There are likely other places in this implementation that could avoid the typecast_copy (which is just a fancy name for an unnecessary data-copy).

Following the merge of this branch, I will need to rebase #4783 against master. This should not present any issues. Please let me know if you require any modifications.

Best regards.

### DIFF
--- a/src/lib/misc/roughtime/roughtime.cpp
+++ b/src/lib/misc/roughtime/roughtime.cpp
@@ -13,6 +13,7 @@
 #include <botan/pubkey.h>
 #include <botan/rng.h>
 #include <botan/internal/socket_udp.h>
+#include <botan/internal/stl_util.h>
 
 #include <cmath>
 #include <map>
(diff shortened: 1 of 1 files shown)


KaganCanSit:
> LGTM thanks

Thanks for the quick response @randombit. I'm on Github today and will generally make time, but if you have time and see fit, I'd also love to complete and merge enhancement #4783 🤝.

Nice day.

PR #4995, opened 2025-07-15
CppCheck - Reduce/Fix warnings in utils directory

KaganCanSit:
This PR addresses multiple cppcheck warnings found in the utils directory to improve code quality and performance.

### Changes Made:

* [botan/src/lib/utils/read_kv.cpp:19] (style) The scope of the variable 'parts' can be reduced. Warning: Be careful when fixing this message, especially when there are inner loops. 

> Note: `split_on` seems to be called here for a check. The `parts` vector is already not used. Perhaps it might be healthier to go for one of the following solutions;
> 		- Remove the `parts` vector completely and call only `split_on`.
> 		- Evaluate the return as an explicit intent to emphasize that `split_on` is called for explicit control.
> ```cpp
> static_cast<void>(split_on(kv, ‘i’));
> ```

* [botan/src/lib/utils/read_kv.cpp:59] (style) The statement 'if (escaped) escaped=false' is logically equivalent to 'escaped=false'. [duplicateConditionalAssign]

* [botan/src/lib/utils/os_utils/os_utils.cpp:805] (performance) When an object of a class is created, the constructors of all member variables are called consecutively in the order the variables are declared, even if you don't explicitly write them to the initialization list. You could avoid assigning 'm_input_handle' a value by passing the value to the constructor in the initialization list. [useInitializationList] 
> Note: `GetStdHandle` does not throw an exception because it is a C API function.

* [botan/src/lib/utils/parsing.cpp:432] (style) Condition 'i>0' is always true [knownConditionTrueFalse] 
> Note: If i > 0, then name[i-1] is always accessible, so no bounds check needed

* [botan/src/lib/utils/http_util/http_util.cpp:146] (performance) Ineffective call of function 'substr' because a prefix of the string is assigned to itself. Use resize() or pop_back() instead. [uselessCallsSubstr]

### Last Status
No functional changes, only optimization and cleaning improvements.

Passes all available unit tests. Compiled and tested with the following command call.
```bash
ninja clean && ./configure.py --without-documentation --cc=clang --compiler-cache=ccache --build-targets=static,cli,tests --build-tool=ninja && ninja && ./botan-test --test-threads=4 --run-long-tests
```

If want me to make changes due to the notes I left or for the arrangements made, it will be enough to leave a comment. I will try to deal with it as soon as possible.

Best regards.

### DIFF
--- a/src/lib/utils/read_kv.cpp
+++ b/src/lib/utils/read_kv.cpp
@@ -16,10 +16,9 @@ std::map<std::string, std::string> read_kv(std::string_view kv) {
       return m;
    }
 
-   std::vector<std::string> parts;
-
    try {
-      parts = split_on(kv, ',');
+      // Validate input format before manual parsing
+      [[maybe_unused]] const auto _ = split_on(kv, ',');
    } catch(std::exception&) {
       throw Invalid_Argument("Bad KV spec");
    }
@@ -55,10 +54,7 @@ std::map<std::string, std::string> read_kv(std::string_view kv) {
          } else {
             cur_val += c;
          }
-
-         if(escaped) {
-            escaped = false;
-         }
+         escaped = false;
       }
    }
 
--- a/src/lib/utils/os_utils/os_utils.cpp
+++ b/src/lib/utils/os_utils/os_utils.cpp
@@ -816,8 +816,7 @@ std::unique_ptr<OS::Echo_Suppression> OS::suppress_echo_on_terminal() {
 
    class Win32_Echo_Suppression final : public Echo_Suppression {
       public:
-         Win32_Echo_Suppression() {
-            m_input_handle = ::GetStdHandle(STD_INPUT_HANDLE);
+         Win32_Echo_Suppression() : m_input_handle(::GetStdHandle(STD_INPUT_HANDLE)) {
             if(::GetConsoleMode(m_input_handle, &m_console_state) == 0)
                throw System_Error("Getting console mode failed", ::GetLastError());
 


KaganCanSit:
Hello Mr. Jack Lloyd (@randombit),

I noticed you're currently continuing work on static analysis (Clang-Tidy). This branch also includes a few cppcheck-related improvements. If you are available and think it is appropriate to make changes, we can quickly review and finalize them together. I'll be online for the next few hours and happy to assist.

Best regards

KaganCanSit:
Hello @randombit, @reneme,
 
The changes in this branch are relatively small, so I believe they can be reviewed quickly. When you have some time, I’d appreciate it if you could take a look. Also, congratulations on the 3.9 release.🚀
 
Best regards.

KaganCanSit:
* 01.11.25 - Rebase master

src/lib/utils/read_kv.cpp
```diff
@@ -16,10 +16,9 @@ std::map<std::string, std::string> read_kv(std::string_view kv) {
       return m;
    }
 
-   std::vector<std::string> parts;
-
    try {
-      parts = split_on(kv, ',');
+      // Validate input format before manuel parsing
+      static_cast<void>(split_on(kv, ','));
```
reneme: I believe nowadays we would write it like that:

```suggestion
      [[maybe_unused]] split_on(kv, ',');
```

reneme (approved):

src/lib/utils/read_kv.cpp
KaganCanSit: The master branch has been rebase and remove the `static_cast`.

KaganCanSit: I didn't realize it while editing quickly, but we need a variable here to make this qualification. Locally GCC just compiles the `[[maybe_unused]]' change and the tests pass. But CI is not good. I'm undecided, but the 3rd option still seems more attractive to me.

I'm reverting the last change for now.

```cpp
// Option 1
[[maybe_unused]] std::vector<std::string> parts = split_on(kv, ',');

// Option 2
[[maybe_unused]] auto x = split_on(kv, ',');

// Option 3
static_cast<void>(split_on(kv, ','));
```

reneme: No hard opinion either way. This is anyway a fairly inefficient parsing implementation that is exclusively used in tests. But for the sake of argument: my preferred option is number 2 with some small adjustment:

```C++
[[maybe_unused]] const auto _ = split_on(kv, ',');
```

... that is more text, but to my mind it is easier to read and understand the intent than the tool-pleasing static cast to void.

KaganCanSit: From a readability perspective, option 2 is actually a good choice, you're right.

Personally, I was undecided between 2 and 3 (I think this is partly due to habits).

As you mentioned, keeping the purpose clearer and improving readability most of the time is the best choice. I've updated it accordingly.

src/lib/utils/parsing.cpp
```diff
@@ -431,7 +431,7 @@ std::string check_and_canonicalize_dns_name(std::string_view name) {
       }
 
       if(mapped == '-') {
-         if(i == 0 || (i > 0 && name[i - 1] == '.')) {
+         if(i == 0 || name[i - 1] == '.') {
```
randombit: I'd say revert this - it *is* a redundant check but the point is that when reading you don't have to think for even a second "wait, are we sure i - 1 can't underflow here?"

src/lib/utils/read_kv.cpp
```diff
@@ -16,10 +16,9 @@ std::map<std::string, std::string> read_kv(std::string_view kv) {
       return m;
    }
 
-   std::vector<std::string> parts;
-
    try {
-      parts = split_on(kv, ',');
+      // Validate input format before manuel parsing
```
randombit: typo should be manual

randombit (changes requested):
Looks good two small comments

KaganCanSit:
Thanks for the review @randombit. I rebase to the master branch and made the necessary changes. It can be merged whenever you want.

KaganCanSit:
This PR needs to be updated based on PR #5567 (`loc`, `location`). I will rebase as soon as possible.

Updates:
06-05-2026 Rebased master.
11-05-2026 Rebased master and trigger CI for for socket permission error.
```cpp
PermissionError: [WinError 10013] An attempt was made to access a socket in a way forbidden by its access permissions
ERROR: Got output on stderr Socket read failed error code 10054 which did not match expected value ERROR: Public key does not match!
```

24-05-2026: @randombit Rebased master branch for conflict in `src/lib/utils/http_util/http_util.cpp`. This changes not require any more. This rebase has significantly reduced the scope of the changes here. Could you take a look when you have time? Rebasing might be more expensive. 😅

randombit (approved):
lgtm sorry about the wait on this one

PR #4996, opened 2025-07-16
Add CertificateParametersBuilder as a replacement for X509_Cert_Options

randombit:


reneme:
I really like the idea!

Perhaps an obvious question: did you consider modeling this after the builder abstractions I had cooked up in the drafts #4694 and #4318?

randombit:
> did you consider modeling this after the builder abstractions

TBH I don't think the extra complexity of it makes sense here, where there is a single consumer of the builder output. I'm still not convinced it's worth the extra complexity even with many different consumers in the form of many different signature schemes (which is why #4318 is still stalled). But here, where there is one consumer, will only ever be one consumer, and that consumer's implementation sits right next to the option implementation? It seems to me very hard to justify.

reneme:
> I don't think the extra complexity of it makes sense here, where there is a single consumer of the builder output.

Frankly, that's understandable. I'm not super happy about the complexity in there either, for sure. A lot of it comes from the "consumption error handling" which indeed might be overkill here?

I do have hope that the compile-time reflection in C++26 will eventually make this easier to handle, when the compiler can take care of the boilerplate. I'd hope that this would eventually collapse into something more elegant (if one considers reflection elegant). 

Independently, C++23's "deducing-this" would also be helpful for the rvalue-vs-lvalue builder handling -- which is a problem here too. You have to have an lvalue of the builder. The following doesn't work:

```C++
auto root_cert_params = Botan::CertificateParametersBuilder()
                                    .add_common_name("Benchmark Root")
                                    .add_country("DE")
                                    .add_organization("RS")
                                    .add_organizational_unit("CS")
                                    .add_dns("unobtainium.example.com")
                                    .add_email("idont@exist.com")
                                    .set_as_ca_certificate();
```

... admittedly, its not a big issue here, but it becomes annoying when trying to return a builder from a function (e.g. `privkey.signer().with_emsa(...).sign()`).

reneme:
That's just my general two cents. If we start establishing builders throughout the API (which really makes a lot of sense in many places, IMHO), I would love to converge on an agreement as to how flexible they should be. I'm fully aware that my previous drafts might have taken it way too far. 😏

randombit:
> You have to have an lvalue of the builder. The following doesn't work:

Yeah what you have is how I initially wrote that code and of course was disappointed. I guess the type could implement a copy constructor, but it'll be relatively expensive and a potential performance issue if for example someone accidentally writes code that copies the params struct for each name they are processing.

reneme:
> Yeah what you have is how I initially wrote that code and of course was disappointed.

Time for C++23 then? 😜

> I guess the type could implement a copy constructor, ...

It doesn't have to be expensive, if you use a `std::shared_ptr<State>`. But it might come with other gotchas because the shared state is highly non-obvious.

reneme:
What actually bugs me quite a bit is the final instantiation that takes a number of positional arguments in `into_self_signed_cert()`:

```C++
         Botan::CertificateParametersBuilder root_cert_params;
         auto cert = root_cert_params
            .add_common_name("Benchmark Root")
            .add_country("DE")
            .add_organization("RS")
            .add_organizational_unit("CS")
            .add_dns("unobtainium.example.com")
            .add_email("idont@exist.com")
            .set_as_ca_certificate()
            .into_self_signed_cert(not_before, not_after, *root_key, rng, get_hash_function());
```

... to my mind that ~somewhat~ spoils the advantage of the builder in that it isn't as extensible as it could be if all params were passed builder-style.

randombit:
> It doesn't have to be expensive, if you use a std::shared_ptr<State>. But it might come with other gotchas because the shared state is highly non-obvious.

Something nice to write fails to compile and you have to write it in a different way > potential for silent performance issue >>>>>>>>>>>>>>>>> hidden shared state with an API that's entirely based around mutating that state. Pimpl with `shared_ptr` works great for immutable objects (eg `X509_Certificate`) but I would not consider it for anything that can be modified.

> spoils the advantage of the builder in that it isn't as extensible as it could be if all params were passed builder-style.

Disagree I guess. There are common elements (eg the DN entries) but the two types (cert vs CSR) effectively have divergent inputs, and the final function call captures those. Also unlike the builder arguments, the final inputs are mostly mandatory; you can create a cert without a commonName, but you cannot create one without a key.

src/lib/x509/x509_builder.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+CertificateParametersBuilder::CertificateParametersBuilder() :
+      m_state(std::make_unique<CertificateParametersBuilder::State>()) {}
+
+CertificateParametersBuilder::CertificateParametersBuilder(CertificateParametersBuilder&& other) noexcept = default;
+
+CertificateParametersBuilder::~CertificateParametersBuilder() = default;
+
+X509_Certificate CertificateParametersBuilder::into_self_signed_cert(std::chrono::system_clock::time_point not_before,
+                                                                     std::chrono::system_clock::time_point not_after,
```
arckoor: This is slightly annoying, `X509_CA::sign_request` takes an `X509_Time`, this takes a time point. Would be nice to use only one

randombit: I'll add a second version of `sign_request` taking time points. It's `X509_Time` there mostly because that API predates adoption of C++11 and `std::chrono`.... but it's generally more useful for an application to specify time as a chrono type vs our ASN.1 type

arckoor: It's fine to leave it as is, it just means one 3 line conversion function for the ffi.
I already have one anyway for `uint64_t -> X509_Time`, so adding one for `uint64_t -> time_point -> X509_Time` isn't a big deal

reneme:
> Pimpl with `shared_ptr` works great for immutable objects (eg `X509_Certificate`) but I would not consider it for anything that can be modified.

Fully agree. That's the point I was trying to make.

> Unlike the builder arguments, the final inputs are mostly mandatory

I would totally agree if the final inputs actually were mandatory. Like in #4694, something like: `privkey->serialize().as_encrypted_pem(rng, "password")`.

But here you have to pass optional signing-related arguments (`hash_fn` and/or `padding`) and it depends on the specific key type which of them are needed:

```C++
      PKCS10_Request into_pkcs10_request(const Private_Key& key,
                                         RandomNumberGenerator& rng,
                                         std::optional<std::string_view> challenge_password = {},
                                         std::optional<std::string_view> hash_fn = {},
                                         std::optional<std::string_view> padding = {});
```

... don't get me wrong, compromises have to be made and it is what it is sometimes. Though, I think we could really do better here.

---

Throwing in a wild idea: In a perfect world these builders would be composable, I think. Assuming the API from #4318 would already be in master:

```C++
auto csr = CertificateParametersBuilder()
             .add_common_name("lol")
             // ...
             .with_signer(key.signer().with_padding(...))
             .into_pkcs10("challenge pw");
```

src/lib/x509/x509_builder.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+      void add_country(std::string_view country) { add_subject_dn("X520.Country", country); }
+
+      void add_state(std::string_view state) { add_subject_dn("X520.State", state); }
+
+      void add_locality(std::string_view locality) { add_subject_dn("X520.Locality", locality); }
+
+      void add_serial_number(std::string_view sn) { add_subject_dn("X520.SerialNumber", sn); }
+
+      void add_org(std::string_view org) { add_subject_dn("X520.Organization", org); }
```
arckoor: Everything but this has the same name, so I'd propose changing this to `add_organization` too

src/lib/x509/x509_builder.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+      void add_organizational_unit(std::string_view org_unit) { add_subject_dn("X520.OrganizationalUnit", org_unit); }
+
+      void add_extension(std::unique_ptr<Certificate_Extension> extn, bool is_critical) {
+         m_extensions.add_new(std::move(extn), is_critical);
+      }
```
arckoor, line 94: The intention with `add_new` here is clearly to add a *new* extension, but I get no feedback from this and thus can't yell at the user for adding an already existing one. It might be better to `replace` or report a failure.

src/lib/x509/x509_builder.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+   return X509_CA::make_cert(
+      signer, rng, sig_algo, pub_key, X509_Time(not_before), X509_Time(not_after), subject_dn, subject_dn, extensions);
+}
+
+PKCS10_Request CertificateParametersBuilder::into_pkcs10_request(const Private_Key& key,
+                                                                 RandomNumberGenerator& rng,
+                                                                 std::optional<std::string_view> challenge_password,
+                                                                 std::optional<std::string_view> hash_fn,
+                                                                 std::optional<std::string_view> padding) {
```
arckoor: I think `challenge_password` will ultimately be kept for PKCS#10, so a small nit would be that I think it should go at the end, just so that it's consistent with the python binding where I'll end up doing `(..., hash_fn, padding=None, challenge_password=None)`

arckoor: Though I suppose it would make sense for `hash_fn` to also default to None, like `into_self_signed_cert` already does essentially. I still think they should be ordered by descending important-ness.

randombit: hash, padding, challenge does seem like the correct ordering in terms of what a user is actually going to want to specify

randombit:
> Though, I think we could really do better here.

I think possibly we just have different models of the inputs to the certificate/CSR. This class is for building up the metadata which is eventually cryptographically bound to the key. The builder is for the metadata, the final arguments are for the binding. I wonder if some different name for this type would be helpful.

> Throwing in a wild idea: In a perfect world these builders would be composable, I think.

That would be pretty neat. But I think there (clearly) has not been consensus reached on what that should look like, which is why it's not on master

src/lib/x509/x509_builder.cpp
```diff
@@ -0,0 +1,288 @@
...
+};
+
+CertificateParametersBuilder::CertificateParametersBuilder() :
+      m_state(std::make_unique<CertificateParametersBuilder::State>()) {}
+
+CertificateParametersBuilder::CertificateParametersBuilder(CertificateParametersBuilder&& other) noexcept = default;
+
+CertificateParametersBuilder::~CertificateParametersBuilder() = default;
+
+X509_Certificate CertificateParametersBuilder::into_self_signed_cert(std::chrono::system_clock::time_point not_before,
```
arckoor: For the PKCS#10 req I can add a serial number when signing, here I can't. ~~Would it maybe be possible to add the serial no to the builder itself, so that I don't have to duplicate that functionality?~~ Probably impossible now that I think about it

src/lib/x509/x509_builder.cpp
```diff
@@ -0,0 +1,288 @@
...
+namespace Botan {
+
+class CertificateParametersBuilder::State final {
+   public:
+      const X509_DN& subject_dn() const { return m_subject_dn; }
+
+      Extensions finalize_extensions(const Public_Key& key) const {
+         auto extensions = m_extensions;
+
+         extensions.replace(setup_alt_name(extensions));
```
arckoor: Even if I manually construct `Cert_Extension::Subject_Alternative_Name` and add it as critical, doesn't this always make it non-critical?

arckoor:
Maybe as a bit of general feedback for the entire cert creation infrastructure, I'm repeatedly running into the issue of trying to figure out what extensions are already taken care of by the builder and just named differently, and what extensions I need to add myself - some documentation would be very helpful for the future

src/lib/x509/x509_builder.cpp
arckoor: The same should be true for `Extended_Key_Usage` below, though I don't know if that is ever "critical" in real-world use

src/lib/x509/x509_builder.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+      void add_email(std::string_view email) { m_email.emplace_back(email); }
+
+      void add_dns(std::string_view dns) { m_dns.emplace_back(dns); }
+
+      void add_uri(std::string_view uri) { m_uri.emplace_back(uri); }
+
+      void add_xmpp(std::string_view xmpp) { m_xmpp.emplace_back(xmpp); }
+
+      void add_ipv4(uint32_t ipv4) { m_ipv4.push_back(ipv4); }
```
arckoor: Not entirely sure if a big-endian encoded value is a good choice here. `subject_info` / `get_user_cert_info` explicitly return a decimal-dotted string when asked for the IP field, so I think accepting one would be the way to go here too.

arckoor:
So we tested a bit, and unfortunately the PKCS#10 flow won't work for our usecase.
Instead I would like to somehow expose `X509_CA::make_cert` to the FFI, but in a way that also uses the builder and does things similar to `CertificateParametersBuilder::into_pkcs10_request` (so that I don't have to expose extensions like SKID, Basic_Constraints, Key_Usage, ...)
Would that be possible?

src/lib/x509/x509_builder.cpp
falko-strenzke: I think there should be a type defined for an IPv4 address. It could have a static function `from_big_endian_u32(uint32_t)`.

arckoor:
@randombit @reneme is there anything I can do to move this forward? I'd like to finalise my X.509 FFI adventures with certificate creation :)

randombit:
Yeah this is worth picking up again. First step is rebasing, also we should now be using `URL`, `EmailAdddress`, etc types.

arckoor:
@randombit any updates on this?

randombit:
@arckoor Nothing so far but I'm going to try to come back to it later this week with an aim to merge for 3.13 or worst case right after the release for inclusion into 3.14

src/lib/x509/x509_builder.cpp
randombit: We have `IPv4Address` now

src/lib/x509/x509_builder.cpp
randombit: Fixed SAN so it copies the criticality of the extension if already existing rather than always setting it to non-critical

EKU is inserted here via `add_new` so if it was prior created via `add_extension` the new one created here is ignored. I did add a check that EKU is only set via either an extension or `add_extended_key_usage` but not both.

randombit:
@arckoor @reneme Rebased to latest master, fixed various issues and review comments and I think this is now ready for another look.

Only possible improvements from my pov

- Maybe name it `CertificateMetadataBuilder` to make it more obvious. (I'm not sure if this is more helpful in practice though)
- Right now the `into_foo` functions take a hash+padding optional strings, maybe should be `PK_Signature_Options` right from the start...

src/lib/x509/x509_builder.cpp
```diff
@@ -0,0 +1,345 @@
...
+CertificateParametersBuilder::CertificateParametersBuilder(CertificateParametersBuilder&& other) noexcept = default;
+
+CertificateParametersBuilder::~CertificateParametersBuilder() = default;
+
+X509_Certificate CertificateParametersBuilder::into_self_signed_cert(const ASN1_Time& not_before,
+                                                                     const ASN1_Time& not_after,
+                                                                     const Private_Key& key,
+                                                                     RandomNumberGenerator& rng,
+                                                                     std::optional<std::string_view> hash_fn,
+                                                                     std::optional<std::string_view> padding) const {
```
arckoor, line 198: I would like to be able to give them a custom serial here

arckoor (commented):
Looks good, but I still would like to see an `into_cert` like in the second commit of #5753.

`ASN1_Time` is very nice, any way this could be used for CRLs as well? Currently I can use `ASN1_Time::from_seconds_since_epoch` for certs, but still have the `timepoint_from_timestamp` around for CRLs

PR #5008, opened 2025-07-20
Remove 'EMSA' terminology where possible

randombit:
A long time ago IEEE 1363 was the future of public key cryptography standards, and using their naming schemes seemed helpful. But now 1363 is withdrawn and long since obscure, and nobody knows what an EMSA is.

Fortunately these days the whole concept is hidden and not visible to applications, so it's safe to rename the base type as well as all of the child types.

This retains the terminology in module names and exported feature macros since these are visible to applications. They can be changed in Botan4.


J08nY:
I was always surprised to see the EMSA terminology and overall use of P1363 in Botan (and not much anywhere else). Now I have the context for it :)

PR #5011, opened 2025-07-21
CI: build amalgamation on GCC 13 (Ubuntu 24.04)

reneme:
The amalgamation build is the default when using the library via Conan. Strange things seem to happen when compiling the amalgamation with GCC 13 on Ubuntu 24.04. Let's see what the CI thinks about this...

Edit: seems to work on GH Actions in first approximation. I'm seeing miscompilations in various places. RSA signatures that don't check out anymore, Serpent slithering out of its path, ...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,14 +71,21 @@ jobs:
         include:
           - compiler: gcc
             target: shared
+            os: ubuntu-22.04
           - compiler: gcc
             target: amalgamation
+            os: ubuntu-22.04
+          - compiler: gcc
+            target: amalgamation
+            os: ubuntu-24.04
           - compiler: gcc
             target: static
+            os: ubuntu-22.04
           - compiler: clang
             target: shared
+            os: ubuntu-22.04
 
-    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.os }}
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


reneme:
@randombit ... currently unclear what's going on here. On WSL2 with Ubuntu 24.04 (GCC 13.3) I'm seeing miscompilations when using the amalgamation. Both when linking statically and dynamically. Using clang 18 fixes it. Currently, I don't see an immediate lead why it would not happen on GHA. Shall I close this and open a ticket instead?

randombit:
I don't see an immediate reason that would be so. Maybe Conan is introducing new compiler flags and/or filtering out some of our default compiler flags? Some distros like to do this for whatever reason.

randombit:
I can replicate a Serpent AVX512 miscompilation with Arch GCC 13.3 plus amalgamation build. No RSA problems though.

Test suite passes with GCC 14.3 and GCC 15.1

a) Is there some specific reason the amalgamation is the default mode? It takes a lot more memory/time to compile, and has a tendency to aggravate compiler bugs.
b) Is there some specific reason for using GCC 13 instead of 14/15?

reneme:
> Is there some specific reason for using GCC 13 instead of 14/15?

Defaults. Otherwise, no.

> Is there some specific reason the amalgamation is the default mode?

No. Just historically me (or someone else, I frankly don't remember) thought that's a good idea. I'll look into changing it upstream in ConanCenter. Its really not cool that the library appears broken on first use when someone pulls it conveniently via Conan. Do you think it makes sense to try and hunt down this miscompilation?

randombit:
> Do you think it makes sense to try and hunt down this miscompilation?

I have so far not had much success in trying to reduce an amalgamation related miscompilation to something that might plausibly be looked at (under say 1000 lines). I mean we can certainly open a bug, attach a 60K line file and say "GCC miscompiles this" but I have a feeling nobody is going to look at that.

Another issue is that GCC 13 is going to get one more release, next year, and then that's it. So even in the best case scenario where it's reproduced, reduced, and fixed, it's still going to be miscompiled for another year. And by the time of 13.4 most people will have just moved on to 14/15/16 anyway...

If I was going to spend time on this (I'm not) I'd probably try fiddling with the code to avoid the miscompilation. If it's a "clean" patch like #3492, just merge it. If it's an ugly workaround like #2803 or #4717 `ifdef` it to the version with a note so it can get removed as quickly as possible.

reneme:
I created a pull request on ConanCenter to disable the amalgamation build by default https://github.com/conan-io/conan-center-index/pull/28010. Closing as "won't do".

PR #5027, opened 2025-07-30
improve botan-config.cmake

### LINKED ISSUE #5026: botan-config.cmake: incorrect paths on some systems
`botan-config.cmake` expects the compiled libraries to be located either in `${_Botan_PREFIX}/lib` or in `${_Botan_PREFIX}/lib/${CMAKE_LIBRARY_ARCHITECTURE}` however on some platforms the correct path fragment is `lib64` instead of `lib`.

On these platforms all attempts to compile a project which uses the Botan cmake targets will fail because the library path will be set incorrectly.

justusranvier:
Use a more robust search to detect the correct library path.

Fixes #5026


reneme (approved):
The change itself looks okay to me, but this cmake integration does really becomes unwieldy. It is also tested in our CI on Ubuntu only, so regressions for systems that use other directory structures will likely not be detected.

I'll open a ticket to track ideas to improve that situation. Edit: https://github.com/randombit/botan/issues/5030

PR #5036, opened 2025-08-03
Enable clang-tidy warning readability-implict-bool-conversion

randombit:


src/lib/ffi/ffi_mac.cpp
```diff
@@ -17,17 +17,19 @@ BOTAN_FFI_DECLARE_STRUCT(botan_mac_struct, Botan::MessageAuthenticationCode, 0xA
 
 int botan_mac_init(botan_mac_t* mac, const char* mac_name, uint32_t flags) {
    return ffi_guard_thunk(__func__, [=]() -> int {
-      if(!mac || !mac_name || flags != 0) {
```
falko-strenzke, line 20: What happened to the check for `flags != 0`?

randombit: Not intentional, not sure what happened there

randombit: Current `maser` seems fine though

```
  int botan_mac_init(botan_mac_t* mac, const char* mac_name, uint32_t flags) {
     return ffi_guard_thunk(func_name: __func__, thunk: [=]() -> int {
        if(any_null_pointers(mac, mac_name)) {
        }

        if(flags != 0) {
           return BOTAN_FFI_ERROR_BAD_FLAG;
        }
```

falko-strenzke: OK, that's good. I just spotted it by chance while performing the review (of 706 commits ;-) ) for the BSI.

PR #5041, opened 2025-08-05
Use span for MGF1

randombit:


src/lib/pk_pad/mgf1/mgf1.cpp
```diff
@@ -13,19 +12,18 @@
...
+void mgf1_mask(HashFunction& hash, std::span<const uint8_t> input, std::span<uint8_t> output) {
    uint32_t counter = 0;
 
    std::vector<uint8_t> buffer(hash.output_length());
-   while(out_len > 0) {
-      hash.update(in, in_len);
+   while(!output.empty()) {
+      hash.update(input);
       hash.update_be(counter);
       hash.final(buffer.data());
```
reneme:
```suggestion
      hash.final(buffer);
```

... while we're on it. :)

randombit:
> The using locations could be simplified with more spans

For sure, just trying to keep this one small and simple. Probably better to do other internal spanify work one algorithm/module at a time, where possible.

PR #5046, opened 2025-08-06
backport security fix for CVE-2024-50382

carlosrodfern:
@randombit ,

This small PR contains a backport patch for a security vulnerability backported into botan 2. It is very specific to Fedora Linux rawhide, so all the logic to support other environments isn't present. I was wondering if you could please review it when you have a chance to make sure I didn't miss anything, and then I'll just close it since it is only for this Linux distro.

I did check the assembly code generated for the `for`-loop body in the `ghash_multiply(..)` with GCC and Clang and it is very clean, no jumps.

The other change in the `configure.py` is related to [this change](https://fedoraproject.org/wiki/Changes/dropingOfCertPemFile) in Fedora Linux. I'll be sending another separate PR for that change for botan 3.

I'm planning to bring botan 3 to Fedora Linux to allow depending packages to migrate as well.

Thank you!

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -2939,6 +2939,7 @@ def set_defaults_for_unset_options(options, info_arch, info_cc, info_os): # pyli
         default_paths = [
             '/etc/ssl/certs/ca-certificates.crt', # Ubuntu, Debian, Arch, Gentoo
             '/etc/pki/tls/certs/ca-bundle.crt', # RHEL
+            '/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem', # Fedora
             '/etc/ssl/ca-bundle.pem', # SuSE
             '/etc/ssl/cert.pem', # OpenBSD, FreeBSD, Alpine
             '/etc/certs/ca-certificates.crt', # Solaris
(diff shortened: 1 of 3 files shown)


randombit:
@carlosrodfern Commit looks good in terms of addressing that specific side channel. I'd suggest one additional change here namely adding a `value_barrier` also to `CT::Mask<T>::value()` since this is likely helpful for addressing other similar issues.

randombit:
> I did check the assembly code generated for the for-loop body in the ghash_multiply(..) with GCC and Clang

Keep in mind this type of side channel is both compiler [*] and architecture dependent; in this case it was only observed on RISC-V.

[*] Actually also compiler version / compilation flags / ...: a few years ago I encountered a similar issue in some Rust code where using the standard Rust compiler builds there was no problem, but using Debian's packaging of `rustc` caused a side channel to be introduced.

carlosrodfern:
@randombit ,
Thank you for all your feedback - very helpful and insightful!  I'll go ahead and close this PR and will be putting this patch into the botan2 package.

Issue #5033, opened 2025-08-01
Allow Intermediate CAs as Trust Anchors


FAlbertDev:
We have a use case, where we need to permit intermediate CAs as trust anchors (ICA trust anchors). We are not concerned with who has signed the intermediate CA, as long as it is present in the relevant certificate store.

## Example
The full chain of trust is as follows:
```
  Root CA (self-signed)
  |
  +-- Intermediate CA (signed by Root CA)
      |
      +-- End Entity Certificate (signed by Intermediate CA)
```
We intend to keep only the intermediate CA in our trusted certificate store; we do not store the root CA, trusting the intermediate CA on its own. However, when attempting to validate the end entity certificate, it doesn't succeed with Botan unless the root CA is added to the trusted certificate store.

## Current Behavior

At present, ICA trust anchors are not compatible with Botan. The `x509_path_validate` function in Botan demands root CAs to be self-signed; otherwise, it returns a `CERT_ISSUER_NOT_FOUND` status due to the need for an existing issuer, even if the intermediate CA is included in the trusted certificate store.

What are your thoughts on whether we should permit ICA trust anchors?

### RFC 5280
I've gathered some sections of RFC 5280 to help determine whether ICA trust anchors might be suitable or not:

[RFC 5280, section 6.1](https://datatracker.ietf.org/doc/html/rfc5280#section-6.1)

>    When the trust anchor is provided in the form of a self-signed certificate, this self-signed certificate is not included as part of the prospective certification path.

[RFC 5280, section 6.1.1](https://datatracker.ietf.org/doc/html/rfc5280#section-6.1.1)

> (d)  trust anchor information, describing a CA that serves as a trust anchor for the certification path.  The trust anchor information includes:
>    (1)  the trusted issuer name,
>    (2)  the trusted public key algorithm,
>    (3)  the trusted public key, and
>    (4)  optionally, the trusted public key parameters associated with the public key.
> The trust anchor information **may be provided to the path processing procedure in the form of a self-signed certificate**. When the trust anchor information is provided in the form of a certificate, the name in the subject field is used as the trusted issuer name and the contents of the subjectPublicKeyInfo field is used as the source of the trusted public key algorithm and the trusted public key.  The trust anchor information is trusted  because it was delivered to the path processing procedure by some trustworthy out-of-band procedure.

[...]

PR #1363, opened 2017-12-19
Fix various x509 path validation bugs + path building with ambiguous DNs

fweissberg:
Various improvements for X509 path validation
* Some `Certificate_Status_Code`s between 500 and 1000 are considered "warnings"
* `x509_path_validate` uses the Certificate_Status_Code::SIG_ALGO_UNKNOWN instead of an exception if necessary
* Check for negative certificate serial numbers: `x509_path_validate` uses the warning Certificate_Status_Code::CERT_SERIAL_NEGATIVE
* Check for DistinguishedName attributes that are longer than permitted: `x509_path_validate` uses the warning Certificate_Status_Code::DN_TOO_LONG
* Check for unknown critical extensions in CRLs: `x509_path_validate` uses the Certificate_Status_Code::CERT_IS_REVOKED. rationale from NIST "Certificate Path Valiadation Testing" document: "When an implementation does not recognize a critical extension in the crlExtensions field, it shall assume that identified certificates have been revoked and are no longer valid"
* Check that x509v1 and x509v2 certificates do not contain extensions: `x509_path_validate` uses the Certificate_Status_Code::EXT_IN_V1_V2_CERT
* Implement a basic version of the issuing distribution point extension for CRLs: `x509_path_validate` uses the Certificate_Status_Code::NO_MATCHING_CRLDP
* Check that policy OIDs are unique within the Certificate Policies Extension: `x509_path_validate` uses the Certificate_Status_Code::DUPLICATE_CERT_POLICY

* CRL_Distribution_Points::decode_inner
    * only uris were considered, now all contents of the AlternativeName are concatenated
* AlternativeName::decode_from
    * AlternativeNames can be DistinguishedNames

* Cache some extensions in X509_CRL objects, as indicated by already existing comments (crl number, authority key id, crl_issuing_distribution_point)
    * fix: cached m_crl_number is now set correctly
    * fix: cached m_auth_key_id is now set correctly
    * add: cache m_issuing_distribution_point

* Implement `build_all_certificate_paths` that generates a list of possibly valid paths. A certificate's issuing certificate not be unambiguous in a set of certificates so there might be more than one path to check. 

[#1283](https://github.com/randombit/botan/issues/1283)
[#1359](https://github.com/randombit/botan/issues/1359)

PR #5047, opened 2025-08-06
X.509 Path: Option for Non-Self-Signed Trust Anchors

FAlbertDev:
This PR introduces a `Path_Validation_Restrictions` option for X.509 path validation, named ~allow_non_self_signed_trust_anchors~ `require_self_signed_trust_anchors`. This option forbids or permits trust anchors that are not self-signed during path validation. This feature is beneficial if intermediate or leaf certificates need to be trusted directly, without the need for a complete path to a self-signed root. For more details, refer to #5033.

## Implementation
To implement the new feature, the following modifications were made:

- The `Path_Validation_Restrictions` class was expanded to include the new `require_self_signed_trust_anchors` option.

- Adjustments to the `PKIX::build_all_certificate_paths` logic are as follows:
  - Previously, this method was not linked due to the missing `BOTAN_PUBLIC_API` annotation. I added `BOTAN_PUBLIC_API(3, 10)`, making it available for applications and testing.
  - The `end_entity` parameter was changed from const `std::optional<X509_Certificate>&` to `const X509_Certificate&`, as handling of `std::nullopt` was incorrect anyway.
  - The function now returns any path from the end entity to any trust anchor in the provided certificate stores, including paths not ending in a self-signed certificate. For instance, with a chain LEAF ← ICA1 ← ICA2 ← ROOT and LEAF, ICA1, and ROOT in the certificate store, the function will return these paths:
    - LEAF
    - LEAF ← ICA1
    - LEAF ← ICA1 ← ICA2 ← ROOT
    
  - A chain of length 1 with only a self-signed certificate, which is contained in a store, is now possible. (see #5023)
    
- The `PKIX::build_certificate_path` logic now relies on `PKIX::build_all_certificate_paths` and returns the first path ending in a self-signed certificate. If no such path exists, it returns the first path not ending in a self-signed certificate.

- The `PKIX::check_chain` logic now incorporates the `require_self_signed_trust_anchors` restriction. When this flag is set to false, non-self-signed trust anchors are permitted. Consequently, signature verification and other checks requiring issuer information are skipped for non-self-signed trust anchors. Note that [RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280#section-6.1) requires checks only for certificates beneath the trust anchor:

  > RFC 5280, Section 6.1:
      "When the trust anchor is provided in the form of a self-signed
      certificate, this self-signed certificate is not included as part of
      the prospective certification path."

- The `PKIX::x509_path_validate` logic also utilizes the new `require_self_signed_trust_anchors` restriction. By default, it  ignores all paths not ending in a self-signed certificate. If the option is set to false, it accepts paths ending in trusted leaf or intermediate certificates.

## Tests
Extensions to the tests in test_x509_path.cpp:

- Tests based on x509test, NIST's, and BSI's path validation tests are executed both with and without the new `require_self_signed_trust_anchors` option. I could not detect any significant performance impact. Importantly, no tests failed with the option enabled, indicating these test suites do not explicitly prohibit non-self-signed trust anchors.

- A new test class, `Non_Self_Signed_Trust_Anchors_Test`, has been added. This class encompasses:

  - Testing `x509_path_validate` with non-self-signed trust anchors.
  - Testing `build_all_certificate_paths` and `build_certificate_path` with and without non-self-signed trust anchors.
  - Testing that the default `require_self_signed_trust_anchors=true` still prohibits non-self-signed trust anchors in `x509_path_validate` and `check_chain`.
  - Testing a chain of length 1 that only contains a self-signed certificate (regression test for #5023).
  
 ## PR Dependencies
 - #5081


src/tests/test_x509_path.cpp
```diff
@@ -215,27 +255,24 @@ std::vector<Test::Result> NIST_Path_Validation_Tests::run() {
...
          store.add_certificate(root_cert);
          store.add_crl(root_crl);
 
+         std::vector<Botan::X509_Certificate> end_certs = {
+            Botan::X509_Certificate(Test::data_file("x509/nist/" + test_name + "/end.crt"))};
+
          for(const auto& file : all_files) {
             if(file.ends_with(".crt") && file != "end.crt") {
-               store.add_certificate(Botan::X509_Certificate(file));
+               end_certs.emplace_back(file);
```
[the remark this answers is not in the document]
FAlbertDev: Yeah, the old code was unprecise. It did not matter until now (because non-self-signed certs in stores were handles as untrusted chain certs). Now all non-root certificates are part of the chain (end_certs) and not in the store.

src/tests/test_x509_path.cpp
```diff
@@ -158,30 +201,27 @@ class X509test_Path_Validation_Tests final : public Test {
...
+         return results;
+      }
+
+   private:
+      std::vector<Test::Result> run_with_restrictions(const Botan::Path_Validation_Restrictions& restrictions);
 };
 
-std::vector<Test::Result> NIST_Path_Validation_Tests::run() {
+std::vector<Test::Result> NIST_Path_Validation_Tests::run_with_restrictions(
+   const Botan::Path_Validation_Restrictions& restrictions) {
```
jdoubleu, line 226: nit: Why is `NIST_Path_Validation_Tests::run()` now part of the Class definition, but `NIST_Path_Validation_Tests::run_with_restrictions` is defined outside the class?

src/tests/test_x509_path.cpp
```diff
@@ -693,14 +730,203 @@ class Root_Cert_Time_Check_Test final : public Test {
...
+            const Botan::Certificate_Store_In_Memory trusted(certs[trusted_cert_idx]);
+
+            std::vector<Botan::X509_Certificate> cert_chain(certs.begin(), certs.begin() + trusted_cert_idx);
+            // Always add at least the leaf
+            if(cert_chain.empty()) {
+               cert_chain.push_back(certs.at(0));
+            }
+
+            Botan::Path_Validation_Result path_result = Botan::x509_path_validate(
+               cert_chain, restrictions, trusted, "", Botan::Usage_Type::UNSPECIFIED, validation_time);
```
jdoubleu: Maybe add a test, where the chain is not ordered (i.e. `ICA2, ROOT, Leaf, ICA1`)?

src/tests/test_x509_path.cpp
```diff
@@ -693,14 +730,203 @@ class Root_Cert_Time_Check_Test final : public Test {
...
+         std::vector<Test::Result> results;
+
+         const auto restrictions = get_allow_non_self_signed_anchors_restrictions(false, false);
+         const auto certs = test_cert_chain_with_validation_time().first;
+
+         using Cert_Path = std::vector<Botan::X509_Certificate>;
+
+         // Helper to create a cert path {certs[0],...,certs[last_cert]}
+         const auto path_to = [&](size_t last_cert) -> Cert_Path {
+            return {certs.begin(), certs.begin() + last_cert + 1};
```
jdoubleu, line 809: Maybe add an `assert( last_cert <= certs.size() )`?

src/tests/test_x509_path.cpp
```diff
@@ -693,14 +730,203 @@ class Root_Cert_Time_Check_Test final : public Test {
...
+
+         const auto path_result = Botan::x509_path_validate(
+            self_signed_root, restrictions, {&cert_store}, "", Botan::Usage_Type::UNSPECIFIED, validation_time);
+
+         result.test_eq("unexpected x509_path_validate result",
+                        path_result.result_string(),
+                        to_string(Botan::Certificate_Status_Code::OK));
+
+         return {result};
+      }
```
jdoubleu: Maybe add some more explicit variations: positive and negative (restriction, wise) test with non-self-signed cert?

src/lib/x509/x509path.h
```diff
@@ -330,13 +347,35 @@ Path_Validation_Result BOTAN_PUBLIC_API(2, 0)
...
-                                                    const std::optional<X509_Certificate>& end_entity,
-                                                    const std::vector<X509_Certificate>& end_entity_extra);
+/**
+* Create all certificate paths by identifying all possible routes from the
+* end-entity certificate to any certificate in the certificate store list. Paths
+* can also end in intermediate or leaf certificates found in the certificate
+* stores.
+*
+* Note that the validity (e.g. signatures or constraints) of these paths IS NOT
+* checked.
```
jdoubleu:
```suggestion
* may also end in intermediate or leaf certificates found in the certificate
* stores.
*
* WARNING: The validity (e.g. signatures or constraints) of these paths IS NOT checked.
```

src/lib/x509/x509path.h
```diff
@@ -330,13 +347,35 @@ Path_Validation_Result BOTAN_PUBLIC_API(2, 0)
...
+                               const std::vector<X509_Certificate>& end_entity_extra);
 
 /**
-* Build certificate path
+* Same as build_all_certificate_paths but only outputs a single path. If there are
+* paths ending in self-signed certificates, these are prioritized  over paths ending
+* in intermediate or leaf certificates of the certificate store.
+*
+* Note that the validity (e.g. signatures or constraints) of the output path IS
+* NOT checked.
```
jdoubleu: see above
```suggestion
* WARNING: The validity (e.g. signatures or constraints) of the output path IS NOT checked.
```

src/lib/x509/x509path.cpp
```diff
@@ -52,10 +52,18 @@ CertificatePathStatusCodes PKIX::check_chain(const std::vector<X509_Certificate>
...
+      const bool at_trust_anchor = (i == cert_path.size() - 1);
 
       const X509_Certificate& subject = cert_path[i];
-      const X509_Certificate& issuer = cert_path[at_self_signed_root ? (i) : (i + 1)];
+
+      // If using intermediate CAs as trust anchors, the signature of the trust
+      // anchor cannot be verified since the issuer is not part of the
+      // certificate chain
+      if(restrictions.allow_non_self_signed_trust_anchors() && at_trust_anchor && !subject.is_self_signed()) {
+         continue;
```
jdoubleu, line 59: This would also accept completely bogus signatures, right? However, if it's self-signed you'd get distinct error types for when the signature is invalid.

src/lib/x509/x509path.cpp
```diff
@@ -136,17 +144,25 @@ CertificatePathStatusCodes PKIX::check_chain(const std::vector<X509_Certificate>
...
+            return cert_path[i + 1];
+         }
+         // Non self-signed trust anchors have no checkable issuers.
+         return {};
+      }();
 
-      if(at_self_signed_root && (issuer.is_self_signed() == false)) {
+      if(!restrictions.allow_non_self_signed_trust_anchors() && !issuer.has_value()) {
          status.insert(Certificate_Status_Code::CHAIN_LACKS_TRUST_ROOT);
       }
```
jdoubleu: what happended to the `issuer.is_self_signed() == false` check?

src/lib/x509/x509path.h
jdoubleu: This is also a breaking change, right? Because previously, they _always_ ended with self-signed certs?

src/tests/test_x509_path.cpp
FAlbertDev: I tried to minimize the Diff for better reviewability by not touching the exisiting code as much as possible. For consistency, I could also move `run` out of the class definition, though.

src/tests/test_x509_path.cpp
FAlbertDev: Good point. This should also work, but an additional test doesn't hurt :+1:

src/lib/x509/x509path.h
FAlbertDev: Yeah, I hope this is okay, since `build_all_certificate_paths` was not available anyway (due to the missing `BOTAN_PUBLIC_API)`

src/lib/x509/x509path.cpp
FAlbertDev: Correct! In fact, the RFC does not require to check the signatures of Trust Anchors at all. So, IMO validating the self-signature is unnecessary anyway... Nevertheless, since I cannot verify non-self signed Trust Anchor signatures and also do not want to break the current behavior with self-signed ones, I have no other choice to keep it this way, I guess.

src/lib/x509/x509path.cpp
FAlbertDev: The old condition `at_self_signed_root && (issuer.is_self_signed() == false` is exactly the one leading to a certificate that has no issuer (neither itself nor a parent), i.e. a non-self-signed trust anchor.

src/tests/test_x509_path.cpp
FAlbertDev: Which scenarios do you have in mind? I can think of testing this with different restrictions and without the `isCA` Flag. Do you know of any additional negative testing scenarios?

FAlbertDev:
>What exactly was the previous behavior when validating self-signed EE certs? These worked, even if it didn't had the isCA flag set?

Unfortunately, this didn't work at all, either for self-signed EEs or CAs (which seems to be a bug IMO, see #5023). Adding a (regression) test for this case is sensible for sure!

src/lib/x509/x509path.cpp
jdoubleu: Why can't you at least check, whether the signature algorithm of the ICA would be supported by Botan (as is done for self-signed roots)?

src/tests/test_x509_path.cpp
jdoubleu: Yeah, mainly only the ones you suggested:
| restriction | root CA type | expected outcome |
| :-: | :-: | :-: |
| default | self-signed | ✅ |
| allow intermediates | self-signed | ✅ |
| default | ICA | ❌ |
| allow intermediates | ICA  | ✅ |
| default | self-signed EE (w/o `isCA`) | ✅? |
| allow intermediates | self-signed EE (w/o `isCA`) | ✅? |

src/lib/x509/x509path.cpp
FAlbertDev: Even if it isn't supported, I see no reason to reject it. In fact, [RFC 5914](https://datatracker.ietf.org/doc/html/rfc5914#section-3) even allows trust anchors in the format of a to-be-signed certificate, i.e. certificates without a signature. 
@randombit What do you think?

src/tests/test_x509_path.cpp
FAlbertDev: Sounds great! I'll add the missing cases 👍

src/tests/test_x509_path.cpp
FAlbertDev: Done: I moved the run method outside the class definition as well.

src/tests/test_x509_path.cpp
FAlbertDev: Done: I added a test.
However, the LEAF must always be the first one:
```cpp
/* PKIX Path Validation
 * @param end_certs certificate chain to validate (with end entity certificate in end_certs[0])
```

src/tests/test_x509_path.cpp
FAlbertDev: Done.

src/tests/test_x509_path.cpp
FAlbertDev: I added all cases. self-signed EEs also work (which is expected IMO).

src/lib/x509/x509path.h
FAlbertDev: Done: I applied your suggestion above.

src/lib/x509/x509path.h
FAlbertDev: Done.

FAlbertDev:
Internal review is done.

src/lib/x509/x509path.cpp
```diff
@@ -136,17 +144,25 @@ CertificatePathStatusCodes PKIX::check_chain(const std::vector<X509_Certificate>
...
-      const X509_Certificate& issuer = cert_path[at_self_signed_root ? (i) : (i + 1)];
+      const auto issuer = [&]() -> std::optional<X509_Certificate> {
+         if(at_trust_anchor && subject.is_self_signed()) {
+            return cert_path[i];
+         } else if(!at_trust_anchor) {
+            return cert_path[i + 1];
+         }
+         // Non self-signed trust anchors have no checkable issuers.
+         return {};
+      }();
```
reneme, line 154: Perhaps easier to follow the control flow if you switch it around like so:


```suggestion
      const auto issuer = [&]() -> std::optional<X509_Certificate> {
         if(!at_trust_anchor) {
            return cert_path[i + 1];
         } else if(subject.is_self_signed()) {
            return cert_path[i];
         } else {
            return {}; // Non self-signed trust anchors have no checkable issuers.
         }
      }();
```

src/lib/x509/x509path.cpp
```diff
@@ -907,6 +882,18 @@ Path_Validation_Result x509_path_validate(const std::vector<X509_Certificate>& e
...
    }
 
+   // If we don't allow non-self-signed trust anchors we need to filter all paths
+   // not ending in a self-signed certificate.
+   std::erase_if(cert_paths, [&](const auto& cert_path) {
+      if(cert_path.empty()) {
+         return true;
+      }
+      return !cert_path.back().is_self_signed() && !restrictions.allow_non_self_signed_trust_anchors();
+   });
```
reneme:
```suggestion
   if(!restrictions.allow_non_self_signed_trust_anchors()) {
      std::erase_if(cert_paths, [&](const auto& cert_path) {
         return cert_path.empty() || !cert_path.back().is_self_signed();
      });
   }
```

... this would make the control flow better reflect your comment above.

src/lib/x509/x509path.h
```diff
@@ -141,6 +148,15 @@ class BOTAN_PUBLIC_API(2, 0) Path_Validation_Restrictions final {
...
       bool ignore_trusted_root_time_range() const { return m_ignore_trusted_root_time_range; }
 
+      /**
+       * Botan typically requires certificates to be self-signed to serve as
+       * trust anchors. Consequently, intermediate CA certificates and leaf
+       * certificates cannot be used as trust anchors, even if they are included
+       * in the Certificate Store. However, this restriction can be removed by
+       * setting allow_non_self_signed_trust_anchors=true in the constructor.
+       */
+      bool allow_non_self_signed_trust_anchors() const { return m_allow_non_self_signed_trust_anchors; }
```
reneme: That might be an annoying change in this patch, but wouldn't it be better to describe this requirement positively? I.e.


```suggestion
      /**
       * By default Botan requires trust anchors to be self-signed.
       * This prevents using intermediate CA certificates and leaf certificates
       * as trust anchors, even if they are included in the Certificate Store.
       * This restriction can be removed by setting
       * require_self_signed_trust_anchors=false in the constructor.
       */
      bool require_self_signed_trust_anchors() const { return m_require_self_signed_trust_anchors; }
```

src/lib/x509/x509path.cpp
FAlbertDev: I like your refactoring idea 👍  Done.

src/lib/x509/x509path.cpp
FAlbertDev: I like it. Done.

src/lib/x509/x509path.h
FAlbertDev: I agree, the positive phrasing feels cleaner. I've applied the change and updated all usages.

jdoubleu (approved):

src/lib/x509/x509path.cpp
```diff
@@ -203,8 +218,10 @@ CertificatePathStatusCodes PKIX::check_chain(const std::vector<X509_Certificate>
       if(subject.x509_version() < 3 && !extensions_vec.empty()) {
          status.insert(Certificate_Status_Code::EXT_IN_V1_V2_CERT);
       }
-      for(const auto& extension : extensions_vec) {
-         extension.first->validate(subject, issuer, cert_path, cert_status, i);
+      if(issuer.has_value()) {
+         for(const auto& extension : extensions_vec) {
+            extension.first->validate(subject, *issuer, cert_path, cert_status, i);
+         }
```
reneme: Most X.509 extensions actually don't require the issuer certificate for validation. So, technically, we could change the `Certificate_Extension::validate()` method to allow a `std::nullopt` for the issuer and still be able to validate many (or most?) extensions.

~~Note that this is public API, but as far as I understand, our semver guarantees don't include user-derived classes. So applications that define their own X.509 extensions would have to accommodate this change.~~

Especially when when using intermediate certificates as trust anchors this is quite relevant (as otherwise the trust anchor's extensions won't be validated). Of course, one could argue that the trust anchor's extensions are "okay by default", because we -- well -- trust them. But I think this needs more consideration.

src/lib/x509/x509path.cpp
```diff
@@ -695,15 +668,18 @@ using cert_maybe_trusted = std::pair<std::optional<X509_Certificate>, bool>;
...
    if(!cert_paths_out.empty()) {
       throw Invalid_Argument("PKIX::build_all_certificate_paths: cert_paths_out must be empty");
    }
 
-   if(end_entity->is_self_signed()) {
-      return Certificate_Status_Code::CANNOT_ESTABLISH_TRUST;
-   }
+   auto cert_in_any_trusted_store = [&](const X509_Certificate& cert) {
+      return std::any_of(trusted_certstores.begin(), trusted_certstores.end(), [&](const Certificate_Store* store) {
+         BOTAN_ARG_CHECK(store != nullptr, "certificate store must not be nullptr");
```
reneme: Nit: perhaps you could also use `BOTAN_ASSERT_NONNULL`.

src/lib/x509/x509path.cpp
```diff
@@ -752,20 +729,18 @@ Certificate_Status_Code PKIX::build_all_certificate_paths(std::vector<std::vecto
...
-               continue;
-            }
-            // found an untrustworthy root
-            else {
-               stats.push_back(Certificate_Status_Code::CANNOT_ESTABLISH_TRUST);
-               continue;
-            }
+         if(last->is_self_signed() && !trusted) {
+            stats.push_back(Certificate_Status_Code::CANNOT_ESTABLISH_TRUST);
+            continue;
```
reneme, line 739:
```C++
         if(trusted) {
            cert_paths_out.push_back(path_so_far);
            cert_paths_out.back().push_back(*last);
         } else if(last->is_self_signed()) {
            stats.push_back(Certificate_Status_Code::CANNOT_ESTABLISH_TRUST);
            continue;
```

... `&& !trusted` is implied by taking the `else` branch anyway.

src/lib/x509/x509path.cpp
```diff
@@ -907,6 +882,18 @@ Path_Validation_Result x509_path_validate(const std::vector<X509_Certificate>& e
       return Path_Validation_Result(path_building_result);
    }
 
+   // If we require trust anchors to be self-signed we need to filter all paths
+   // not ending in a self-signed certificate.
+   if(restrictions.require_self_signed_trust_anchors()) {
+      std::erase_if(cert_paths,
+                    [&](const auto& cert_path) {  //
+                       return cert_path.empty() || !cert_path.back().is_self_signed();
+                    });
```
reneme:
```suggestion
                    [&](const auto& cert_path) {
                       return cert_path.empty() || !cert_path.back().is_self_signed();
                    });
```

src/lib/x509/x509path.h
```diff
@@ -330,13 +349,35 @@ Path_Validation_Result BOTAN_PUBLIC_API(2, 0)
...
 
 /**
-* Build certificate path
+* Same as build_all_certificate_paths but only outputs a single path. If there are
+* paths ending in self-signed certificates, these are prioritized  over paths ending
+* in intermediate or leaf certificates of the certificate store.
+*
+* WARNING: The validity (e.g. signatures or constraints) of the output path IS
+* NOT checked.
+*
```
reneme, line 380:
```suggestion
* Same as build_all_certificate_paths but only outputs a single path. If there are
* paths ending in self-signed certificates, these are prioritized over paths ending
* in intermediate or leaf certificates of the certificate store.
*
* WARNING: The validity (e.g. signatures or constraints) of the output path IS
* NOT checked.
*
```

... sometimes I'm a kerning nerd. ;)

src/lib/x509/x509path.cpp
FAlbertDev: PKIX::build_all_certificate_paths may be called by users. Therefore, I would rather throw the invalid argument exception in this case. However, I noticed that putting this input validation in `std::any_of` is not elegant, since it does not run on all vector elements. I'll move it into a separate check at the top.

src/lib/x509/x509path.cpp
FAlbertDev: Fair 👍

src/lib/x509/x509path.cpp
FAlbertDev: I put the `//` to avoid a newline by clang-format. Otherwise, it would like the following, which is a bit stuffed IMO:
```cpp
std::erase_if(cert_paths,
                    [&](const auto& cert_path) { return cert_path.empty() || !cert_path.back().is_self_signed(); });
```

src/lib/x509/x509path.cpp
reneme: To extend this API backward-compatibly, I see three options:

* add a `validate()` overload that doesn't take an issuer at all
* pass a default-constructed `X509_Certificate` as issuer and handle that correctly in each overload (🤢)
* just change the public API, using `std::optional<X509_Certificate>` for the issuer and rely on implicit conversion to keep it backward compatible. (❤️)

For the first two options I would propose adding a `protected` customization point and make the public `validate()` methods non-const. The third option doesn't even need that.

All suggestions would break downstream code that derived from `Certificate_Extension`, but that [doesn't break our semver guarantees](https://botan.randombit.net/handbook/sem_ver.html#exception-1-deriving-from-library-classes).

@randombit What do you think?

FAlbertDev: I agree! If we decide to apply any of these suggestions I'll create a separate pull request for this Extension API change.

src/lib/x509/x509path.cpp
reneme: 🤣 fair enough.

src/lib/x509/x509path.cpp
randombit: I'd be fine with changing this in a non-backward compatible way. Deriving a new certificate extension is already going to be rare and a certification extension that also has validation logic more so. Of the three options I'd prefer the last.

src/lib/x509/x509path.cpp
```diff
@@ -602,70 +619,26 @@ Certificate_Status_Code PKIX::build_certificate_path(std::vector<X509_Certificat
...
                                                      const X509_Certificate& end_entity,
                                                      const std::vector<X509_Certificate>& end_entity_extra) {
-   if(end_entity.is_self_signed()) {
-      return Certificate_Status_Code::CANNOT_ESTABLISH_TRUST;
-   }
-
-   /*
-   * This is an inelegant but functional way of preventing path loops
-   * (where C1 -> C2 -> C3 -> C1). We store a set of all the certificate
-   * fingerprints in the path. If there is a duplicate, we error out.
```
randombit, line 608: Why was this removed? I don't see the logic being moved elsewhere

randombit: (I *think* we have a check that we detect and reject cycles but I'm not positive...)

FAlbertDev: I based the entire logic of `build_certificate_path` on a call to `build_all_certificate_paths`. Otherwise, this feels like duplication of quite complex code. Additionally, the `build_certificate_path` function has not been tested yet. 
I didn't measure if there is a notable performance decrease, though.

src/lib/x509/x509path.cpp
FAlbertDev: #5081

FAlbertDev: I rebased to master and updated the check so that extensions are checked even for non-self signed trust anchors.

reneme: LGTM. Thanks.

src/lib/x509/x509path.cpp
randombit: I feel like something has gone awry here. Since #1363 `build_certificate_path` has been both entirely unused and entirely untested. Yet still exposed as a public interface for all this time. :/ We should have deprecated it for removal at that point. Let's fix that here.

src/lib/x509/x509path.cpp
```diff
@@ -165,28 +181,27 @@ CertificatePathStatusCodes PKIX::check_chain(const std::vector<X509_Certificate>
          }
       }
 
-      // Only warn, if trusted root is not in time range if configured this way
-      const bool is_trusted_root_and_time_ignored =
-         restrictions.ignore_trusted_root_time_range() && at_self_signed_root;
+      // Only warn, if trust anchor is not in time range if configured this way
```
randombit: I know the comment was pre-existing but this is awkwardly worded. Maybe

> If so configured, allow trust anchor outside validity period with a warning rather than a hard error

src/lib/x509/x509path.cpp
```diff
@@ -165,28 +181,27 @@ CertificatePathStatusCodes PKIX::check_chain(const std::vector<X509_Certificate>
...
+      const bool is_trust_anchor_and_time_ignored = restrictions.ignore_trusted_root_time_range() && at_trust_anchor;
```
randombit: Likewise to previous comment we already had `_time_ignored` suffix here but really we are not ignoring the time, we're just not strictly enforcing the validity period.

Maybe

const bool enforce_validity_period = !at_trust_anchor || (at_trust_anchor && !restrictions.ignore_trusted_root_time_range());

Then inverting the logic of the warn vs error cases below to

```
if(enforce_validity_period) {
   // insert error code
} else {
  // insert warning code
}```

src/lib/x509/x509path.cpp
randombit: I'd suggest naming the lambda instead

src/lib/x509/x509path.cpp
FAlbertDev: Alright! I deprecated `build_certificate_path` 👍

src/lib/x509/x509path.cpp
FAlbertDev: I like your idea. I refactored the respective code and improved the comment above the variable👍

src/lib/x509/x509path.cpp
FAlbertDev: I use a separate (i.e. named) lambda now.

src/lib/x509/x509path.cpp
FAlbertDev: see thread below

FAlbertDev:
@randombit Big thanks for your review! I applied your suggestions; the PR should be ready to go now. After approval, I'll rebase (and cleanup the commit history) and merge the PR 👍

FAlbertDev:
@randombit If I remember correctly, you already reviewed everything but the latest commit of this PR. If you find the time for reviews, this PR may be a low-hanging fruit :)

randombit (approved):
LGTM besides needing a history cleanup, thanks

FAlbertDev:
Big thanks! I rebased and raised the new BOTAN_PUBLIC_API to 3.11 for this MR 👍

PR #5048, opened 2025-08-07
fix: use fedora new pem cert location

carlosrodfern:
Fedora 43 currently has the ca bundle for TLS in this location, and the classic location /etc/pki/tls/certs/ca-bundle.crt is targeted to be deleted in Fedora Linux 43.
See https://fedoraproject.org/wiki/Changes/dropingOfCertPemFile

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -3176,6 +3176,7 @@ def set_defaults_for_unset_options(options, info_arch, info_cc, info_os):
         default_paths = [
             '/etc/ssl/certs/ca-certificates.crt', # Ubuntu, Debian, Arch, Gentoo
             '/etc/pki/tls/certs/ca-bundle.crt', # RHEL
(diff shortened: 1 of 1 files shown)


randombit (approved):
:+1: Thanks

randombit:
@carlosrodfern Just realized this targeted `release-3` which is not correct, all commits should go into `master` and then `release-3` is synced at the point that a release occurs. I just force-pushed back `release-3`, but I can't reopen the PR because GH still thinks it has merged. Can you create a new PR targeting `master` branch? Thanks

carlosrodfern:
> @carlosrodfern Just realized this targeted `release-3` which is not correct, all commits should go into `master` and then `release-3` is synced at the point that a release occurs. I just force-pushed back `release-3`, but I can't reopen the PR because GH still thinks it has merged. Can you create a new PR targeting `master` branch? Thanks

Yes, here is the PR: https://github.com/randombit/botan/pull/5052
Thank you!

PR #5056, opened 2025-08-08
Custom ECDH adapter in TLS 1.2 must no-longer derive from `ECDH_PublicKey`

reneme:
[tbd]


src/lib/tls/tls12/msg_client_kex.cpp
```diff
@@ -114,19 +121,10 @@ Client_Key_Exchange::Client_Key_Exchange(Handshake_IO& io,
             append_tls_length_value(m_pre_master, psk.bits_of(), 2);
          }
 
-         if(curve_id.is_ecdh_named_curve()) {
-            auto* ecdh_key = dynamic_cast<ECDH_PublicKey*>(private_key.get());
```
reneme, line 118: This line caused the concrete issue: `private_key` is passed in by the user and must derive from `PK_Key_Agreement_Key`. But if it handles a known ECDH group we just try to downcast to `ECDH_PublicKey` to handle the `EC_Point_Format` selection (compressed vs. uncompressed) in the serialization of the public value.

src/lib/tls/tls12/msg_server_kex.cpp
```diff
@@ -97,33 +97,23 @@ Server_Key_Exchange::Server_Key_Exchange(Handshake_IO& io,
...
       const uint16_t named_curve_id = m_shared_group.value().wire_code();
       m_params.push_back(3);  // named curve
       m_params.push_back(get_byte<0>(named_curve_id));
       m_params.push_back(get_byte<1>(named_curve_id));
 
-      append_tls_length_value(m_params, ecdh_public_val, 1);
+      // Note: In contrast to public_value(), raw_public_key_bits() takes the
+      // point format (compressed vs. uncompressed) into account that was set
+      // in its construction within tls_generate_ephemeral_key().
+      append_tls_length_value(m_params, m_kex_key->raw_public_key_bits(), 1);
```
jdoubleu, line 123: Is there any other failsafe, when the `tls_generate_ephemeral_ecdh_key()`  ignores the point format? TLS may fail, right?

src/lib/tls/tls_callbacks.h
```diff
@@ -463,6 +464,33 @@ class BOTAN_PUBLIC_API(2, 0) Callbacks /* NOLINT(*-special-member-functions) */
...
+       *
+       * @param group the group identifier to generate an ephemeral keypair for
+       *              TLS 1.2 allows for specifying custom discrete logarithm
+       *              parameters as part of the protocol. Hence the variant<>.
+       * @param rng a random number generator
+       * @param tls12_ecc_pubkey_encoding_format the key's serialization format
+       *
+       * @return an ECDH private key of an algorithm usable for key agreement
+       */
+      virtual std::unique_ptr<PK_Key_Agreement_Key> tls_generate_ephemeral_ecdh_key(
```
jdoubleu: When it's only meant for TLS1.2, we could clarify that in the function name as well (e.g. `tls12_generate_...`)?

src/tests/unit_tls.cpp
```diff
@@ -863,6 +991,65 @@ class TLS_Unit_Tests final : public Test {
...
+      void test_custom_ecdh_provider(std::vector<Test::Result>& results,
+                                     const std::shared_ptr<Credentials_Manager_Test>& creds,
+                                     const std::shared_ptr<Botan::RandomNumberGenerator>& rng) {
+         auto noop_session_manager = std::make_shared<Botan::TLS::Session_Manager_Noop>();
+
+         const auto groups = {
+            Botan::TLS::Group_Params::SECP256R1,
+            Botan::TLS::Group_Params::BRAINPOOL256R1,
+            Botan::TLS::Group_Params::BRAINPOOL512R1,
+         };
```
jdoubleu, line 1003: What about X25519 and X448 keys?

src/lib/tls/tls12/msg_server_kex.cpp
reneme: True, if the callback is overridden and the user fails to configure the key correctly, the handshake might (probably will) fail. But there are plenty of ways the user could wreak havoc when doing something wrong in those callbacks. Especially the optional ones that are meant to be overridden only for very specialized use cases.

On the TLS protocol level: if we were to start misbehaving due to some error in the callback overrides, typically it would be the peer's job to detect that and abort the handshake.

src/lib/tls/tls_callbacks.h
reneme: Fair point.

src/tests/unit_tls.cpp
reneme: Those need special handling in the callbacks, because they are not used via their `ECDH_PrivateKey` but have their own implementations. Albeit, you're right. We should test those as well.

reneme: Taking the "special handling of X25519/X448" into account: let's just scope this new callback to "ordinary" ECDH and let X25519/X448 be handled by the existing callback. The new callback is needed exclusively because of the adjustable ECC point encoding in TLS 1.2 and X25519/X448 ` are not affected by that, because it always uses "compressed" encoding by definition.

src/lib/tls/tls12/msg_client_kex.cpp
```diff
@@ -114,19 +128,10 @@ Client_Key_Exchange::Client_Key_Exchange(Handshake_IO& io,
...
```
reneme, line 118: This caused the concrete issue: `private_key` is passed in by the user and must derive from `PK_Key_Agreement_Key`. But if it handles a known named ECDH group we just try to downcast to `ECDH_PublicKey` to handle the `EC_Point_Format` selection (compressed vs. uncompressed) in the serialization of the public value.

doc/api_ref/tls.rst
```diff
@@ -1185,6 +1185,20 @@ Code Example: TLS Client using Custom Curve
...
+Users that wish to implement a custom ECDH provider for TLS 1.2 (e.g. to offload the implementation of
+standard curves to some crypto hardware), must take the negotiated ECC point encoding (compressed vs.
+uncompressed) into account. They should use the special callback  ``tls12_generate_ephemeral_ecdh_key``
+which provides the desired ECC point encoding as an input parameter.
+
+This special callback is called *only for TLS 1.2* and *only for ECDH using standardized curves* that
+Botan is aware of (for instance ``secp256r1``, ``brainpool256r1`` and such). Explicitly, that *does not
+include X25519 and X448* as those algorithms have a well-defined point format. TLS 1.3 does not allow
+negotiating the ECC point encoding (see `RFC 8446 Section 4.2.8.2 <https://www.rfc-editor.org/rfc/rfc8446#section-4.2.8.2>`_)
+and thus does not call this callback either.
```
jdoubleu:
```suggestion
and thus does not call this callback neither.
```

reneme: "does not ... neither" would be a double-negation, no? Cambridge dictionary is on my side too: https://dictionary.cambridge.org/grammar/british-grammar/neither-neither-nor-and-not-either_2#neither-neither-nor-and-not-either_2__45

randombit: Maybe worth adding a note here that compressed point support in TLS 1.2 is deprecated and that this callback will be removed when such support is dropped.

src/lib/tls/tls_callbacks.h
```diff
@@ -463,6 +464,33 @@ class BOTAN_PUBLIC_API(2, 0) Callbacks /* NOLINT(*-special-member-functions) */
...
+       * security risks are associated with customizing TLS's key exchange
+       * mechanism.
+       *
+       * @throws TLS_Exception(Alert::DecodeError) if the @p group is not known.
+       *
+       * @param group ECDH group identifier to generate an ephemeral keypair for
+       * @param rng a random number generator
+       * @param tls12_ecc_pubkey_encoding_format the key's serialization format
+       *
+       * @return an ECDH private key of an algorithm usable for key agreement
```
randombit, line 492: Worth adding a similar note here re this callback being removed when support for compressed points in TLS is dropped

randombit (approved):

reneme:
Thanks!

PR #5058, opened 2025-08-14
Run an additional clang-tidy job on macOS

randombit:
This covers both Aarch64 and macOS specific code which is otherwise missed by the usual clang-tidy run.


randombit:
> pyyaml has been disabled because it does not meet homebrew/core's requirements for Python library formulae

OK never mind I guess

PR #5060, opened 2025-08-14
Fix various clang-tidy warnings in aarch64 specific code

randombit:


src/build-data/cc/clang.txt
```diff
@@ -88,6 +88,8 @@ ppc64:power9 -> "-mcpu=power9"
 
 arm64:armv8crypto -> "-march=armv8+crypto"
 arm64:armv8sha512 -> "-march=armv8.2-a+sha3"
+arm64:armv8sm3 -> "-march=armv8.2-a+sm4"
+arm64:armv8sm4 -> "-march=armv8.2-a+sm4"
```
reneme, line 92: Is that a typo, or do you really want to enable sm4 for sm3 in the first line?

randombit: GCC and Clang for whatever reason bundle both extensions together, and neither recognizes `-march=armv8.2-a+sm3`

Issue #4943, opened 2025-06-23
Feature Request: ASCON


Dath-Ohne-R:
Hello dear Botan developers,

First of all, thank you for your fantastic work on Botan — I truly appreciate the library and use it with great enthusiasm!

There’s one feature I would love to see added: support for ASCON https://ascon.isec.tugraz.at/
ASCON is a modern and efficient cipher that has won both the CAESAR and the NIST Lightweight Cryptography competitions. Designed specifically for IoT and constrained environments, it offers a comprehensive suite including AEAD, hashing, tagging, extended output function and stream cipher capabilities — essentially everything one might need for secure and lightweight encryption.

It would be fantastic to see ASCON integrated into Botan’s portfolio. I believe it would be a valuable addition for many developers working in resource-constrained contexts or communicating with their embedded devices.

Thank you for considering this request, and keep up the great work!

If this is not the right place to voice a feature request, I sincerely apologise, but I found no hint on your website where to put it.

randombit:
Thanks for the kind words. And feature requests via issues is fine.

Ascon is certainly in scope for the library, especially considering the standardization status. And actually I find the design quite nice.

Is there some specific aspect of Ascon that  you would be particular interested in (AEAD, XOF, etc)?

randombit:
Looked at this a bit today. I would want to wait until the final NIST SP is published since NIST seems to like making changes between the ipd and final spec, and having to possibly support two variants for an extended period adds to ongoing maintenance costs, which I'd like to avoid if possible.

Dath-Ohne-R:
Thank you for considering my feature request.

I am particularly interested in the AEAD and Hash, but since the basic structure of the cipher is used in all variations, maintenance of multiple aspects should be in reach.

I agree that the final NIST publication should be awaited first.

reneme:
> I would want to wait until the final NIST SP is published

It is out now. :) https://www.nist.gov/news-events/news/2025/08/nist-finalizes-lightweight-cryptography-standard-protect-small-devices

reneme:
The Hash and XOF are merged now. AEAD and cXOF are still to-do.

The AEAD needs [some extra functionality](https://github.com/randombit/botan/pull/5061#pullrequestreview-3136436743) in the permutation. Also, the spec allows for some flexibility (truncated tag, masked nonce) in its usage of which we may or may not implement everything. @Dath-Ohne-R do you have any specific needs here?

reneme:
[...]

PR #5061, opened 2025-08-16
NIST SP.800-232: Ascon-Hash256

reneme:
As requested in https://github.com/randombit/botan/issues/4943

This is work in progress... The permutation itself evidently works as expected (because after `Ascon_Hash256::clear()` the state is equivalent to what SP.800-232 Table 12 is defining). But the hash output itself doesn't match with the [test vectors from ACVP](https://github.com/usnistgov/ACVP-Server/blob/master/gen-val/json-files/Ascon-Hash256-SP800-232/internalProjection.json), yet. Probably the absorb or squeeze functions (which I currently just copy-pasted from Keccak) still have some bug.


src/lib/hash/ascon_hash256/ascon_hash256.cpp
```diff
@@ -0,0 +1,46 @@
...
+}
+
+std::unique_ptr<HashFunction> Ascon_Hash256::copy_state() const {
+   return std::make_unique<Ascon_Hash256>(*this);
+}
+
+void Ascon_Hash256::init() {
+   // See NIST SP.800-232, Section 5.1.1 (Initialization)
+   m_ascon_p.clear();
+   m_ascon_p.permute<12>();
```
reneme: Note to self: I'm not convinced anymore that passing the IV as a template parameter was that smart after all. Probably better to just pass it here and statically make sure that the right amount of IV words is passed.

Or: actually make all the parameters that build up the IV explicit and calculate it constexpr as explained in SP.800-232 Apx. B. That might be a liiiiittle overkill, admittedly. 😏

src/tests/test_hash.cpp
```diff
@@ -146,7 +146,7 @@ class Hash_Function_Tests final : public Text_Based_Test {
                result.test_eq(provider, "hashing split", fork->final(), expected);
             }
 
-            if(hash->hash_block_size() > 0) {
+            if(hash->hash_block_size() > 0 && hash->name() != "Ascon-Hash256" /* which has an 8 bytes block size */) {
```
randombit: The internal block size is only exposed for HMAC and Ascon is not approved for use with HMAC, so probably this should be set to zero for Ascon

reneme: Alright. Thanks.

src/lib/hash/ascon_hash256/ascon_hash256.cpp
reneme: Introduced an `AsconParameters` struct that holds these values as specified in Appendix B.

doc/api_ref/hash.rst
```diff
@@ -97,6 +97,20 @@ Available Hash Functions
...
 
+Ascon-Hash256
+^^^^^^^^^^^^^
+
+Available if ``BOTAN_HAS_ASCON_HASH256`` is defined.
+
+A hash function based on the Ascon permutation, specifically designed to allow
+small footprint implementations. Its main use case is in constrained
+environments, such as IoT devices where traditional cryptographic functions
+may be too resource intensive.
```
randombit, line 108: This hints at it but I think this should be spelled out more concretely

> Unless you are interoperating with an existing device which due to resource contraints can only use Ascon, prefer more typical hashes such as SHA-256, SHA-512, or SHA-3.

src/lib/hash/ascon_hash256/ascon_hash256.cpp
```diff
@@ -0,0 +1,36 @@
+/*
+* Ascon-Hash256 (NIST SP.800-232)
+* (C) 2025 Jack Lloyd
+*.    2025 René Meusel
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include <botan/assert.h>
+#include <botan/internal/ascon_hash256.h>
```
randombit, line 9: Wrong ordering of includes, header defining the type always goes first, then an empty line (to bypass `clang-format` header reordering) then all other headers.

src/lib/permutations/ascon_perm/ascon_perm.h
```diff
@@ -0,0 +1,193 @@
...
+         static_cast<uint64_t>(rate_bytes) << 40;
+};
+
+/**
+ * Ascon_p as specified in NIST SP.800-232, Section 3
+ */
+template <typename Params>
+class Ascon_p final {
+   protected:
+      using reg_t = uint64_t;
```
randombit: I don't think this typedef improves understanding, Ascon is defined over 64-bit words

src/lib/hash/ascon_hash256/ascon_hash256.h
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+* Ascon-Hash256 (NIST SP.800-232 Section 5.1)
+*/
+class Ascon_Hash256 : public HashFunction {
+   private:
+      using Params = AsconParameters<2, 12, 12, 256, 8>;
+      using Permutation = Ascon_p<Params>;
+
+      constexpr static Permutation initial_state = Permutation{}.permute<Params::init_and_final_rounds>();
```
randombit: So while being able to precompute the initial state is a nice trick I really don't like how the definition of the permutation leaks out into all callers of the Ascon types. Which at least since it's no longer a public type is still constrainted to internal, but why should `hash.cpp` or `aead.cpp` care about the permutation?

It just seems a lot more straightforward to copy the constants in Table 12.

An analogy I would make is that SHA-256's round constants are the cube roots of primes, and we *could* constexpr compute these - but would doing so make the code easier or harder to understand?

src/lib/permutations/ascon_perm/ascon_perm.h
```diff
@@ -0,0 +1,193 @@
...
+      using state_t = std::array<reg_t, 5>;
+
+   public:
```
randombit, line 35: I would prefer `Ascon_p` just be initialized at construction with the number of rounds to use for init + processing plus the byte rate and the entire (prehashed) 5 word initial state. Then `absorb`, `squeeze`, and `finish` can all be defined in a cpp file instead of being inlined, and `permute` is no longer parameterized at compile time.

Unless the compiler is *very* smart, I'm pretty sure what would happen with this is that as various Ascon types (AEAD, XOF, ...) are added, we'll end up shipping 2*N different implementations of the Ascon permutation logic, since each depends on a `Params` which is going to be unique for that usage, and then for each type we get two  `permute` that different on both `Params` and the round count.

src/lib/permutations/ascon_perm/ascon_perm.h
```diff
@@ -0,0 +1,193 @@
...
+         static_cast<uint64_t>(processing_rounds) << 20 | static_cast<uint64_t>(algo_bit_number) << 24 |
+         static_cast<uint64_t>(rate_bytes) << 40;
+};
+
+/**
+ * Ascon_p as specified in NIST SP.800-232, Section 3
+ */
+template <typename Params>
+class Ascon_p final {
+   protected:
```
randombit: `protected` is weird here

src/lib/hash/ascon_hash256/ascon_hash256.h
```diff
@@ -0,0 +1,51 @@
...
+#include <botan/secmem.h>
+#include <botan/internal/ascon_perm.h>
+#include <string>
+
+namespace Botan {
+
+/**
+* Ascon-Hash256 (NIST SP.800-232 Section 5.1)
+*/
+class Ascon_Hash256 : public HashFunction {
```
randombit: should be `final`

randombit (changes requested):

reneme:
In the mean time I added an implementation of Ascon-XOF128 (which of course is trivial having the Hash variant). This is currently based on the templated permutation implementation.

Thanks for the review on this. I agree with reverting the templated permutation and making it a more traditional compilation unit. I was hoping that the "lightweigt crypto" is lightweight enough that the compiler would actually unroll the permutation loop or do other clever things. It didn't. Neither for 12 rounds nor for 8. And thus there's really no benefit in templating this.

I'll address your comments during the week.

src/lib/hash/ascon_hash256/ascon_hash256.h
reneme: Done.

src/lib/permutations/ascon_perm/ascon_perm.h
reneme: Done.

reneme:
As agreed upon, I replaced the templated implementation of the permutation with a traditional compilation unit that can be parameterized at runtime. For readability within the Hash, XOF, (and future AEAD) implementations, I introduced a `struct AsconConfig` that wraps the permutation parameters so that I can structure-initialize them with their names.

The AEAD will require further extensions to the permutation implementation. Namely:

* incorporating the Nonce and Key into the initial state
* xor-ing the Key and AAD domain separator into `m_S[2]`, `m_S[3]` and `m_S[4]` in various places
* extract `m_S[0]` and `m_S[1]` after absorbing the plaintext/ciphertext to obtain the ciphertext/plaintext
* obtain `m_S[3]` and `m_S[4]` to calculate the AEAD tag

This might require some refactoring in the permutation. For the sake of a clean commit history we could do this here. But I think it makes sense to merge the hash/XOF implementations now and continue in a new PR.

Regarding the cXOF: for SHAKE we currently implement the customizable variant but we don't expose it via the public API. Not sure whether or not we should stick to that line and just not implement it until explicitly requested.

src/lib/permutations/ascon_perm/ascon_perm.h
```diff
@@ -0,0 +1,65 @@
...
+      constexpr size_t capacity() const { return (sizeof(AsconState) * 8) - m_bit_rate; }
+
+      constexpr size_t bit_rate() const { return m_bit_rate; }
+
+      constexpr size_t byte_rate() const { return m_bit_rate / 8; }
+
+      void absorb(std::span<const uint8_t> input, std::optional<uint8_t> permutation_rounds = std::nullopt);
+      void squeeze(std::span<uint8_t> output);
+      void finish();
+      void permute(uint8_t rounds);
```
randombit, line 67: Does permute need to be public? Neither hash nor xof use it

doc/api_ref/hash.rst
```diff
@@ -97,6 +97,42 @@ Available Hash Functions
...
+
+Unless you are interoperating with an existing device which due to resource
+constraints can only use Ascon, prefer more typical hashes such as SHA-256,
+SHA-512, or SHA-3.
+
+This hash function is standardized by NIST in SP.800-232.
```
randombit: Maybe to head off confusion

> This hash function is standardized by NIST in SP.800-232. It is not compatible with earlier versions of the Ascon specification.

src/lib/permutations/ascon_perm/ascon_perm.h
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Ascon_p as specified in NIST SP.800-232, Section 3
+ */
+class Ascon_p final {
+   public:
+      constexpr explicit Ascon_p(AsconConfig config) :
+            m_S(config.initial_state),
+            m_bit_rate(config.bit_rate),
+            m_init_final_rounds(config.init_and_final_rounds),
+            m_processing_rounds(config.processing_rounds) {}
```
randombit: Should have some basic checks of the input like rounds > 0 and <= 16, bitrate is possible, ...

src/lib/permutations/ascon_perm/ascon_perm.cpp
```diff
@@ -0,0 +1,141 @@
...
+   // We must always add a padded final input block, if the last verbatim
+   // input block aligned with the byte rate, the final block may be just
+   // padding bytes, otherwise the final block is padded as needed.
+
+   absorb(std::span{padding}.first(byte_rate() - m_S_inpos), m_init_final_rounds);
+   BOTAN_ASSERT_NOMSG(m_S_inpos == 0);
+}
+
+void Ascon_p::permute(uint8_t rounds) {
+   // NIST SP.800-232, Table 5
```
randombit, line 103: Maybe an assert here that rounds is not out of range

src/lib/permutations/ascon_perm/ascon_perm.cpp
```diff
@@ -0,0 +1,141 @@
...
+#include <botan/internal/stl_util.h>
+
+namespace Botan {
+
+void Ascon_p::absorb(std::span<const uint8_t> input, std::optional<uint8_t> permutation_rounds) {
+   BufferSlicer input_slicer(input);
+
+   // Block-wise incorporation of the input data into the sponge state until
+   // all input bytes are processed
+   while(!input_slicer.empty()) {
```
randombit, line 22: This could be optimized for the case where there is no data currently accumulated and at least a full buffer available. Maybe not worth bothering with at least initially though

src/lib/permutations/ascon_perm/ascon_perm.cpp
```diff
@@ -0,0 +1,141 @@
...
+      for(; !input_this_round.empty() && m_S_inpos % 8 > 0; ++m_S_inpos) {
+         m_S[m_S_inpos / 8] ^= static_cast<uint64_t>(input_this_round.take_byte()) << (8 * (m_S_inpos % 8));
+      }
+
+      // Process as many aligned 64-bit integer values as possible
+      for(; input_this_round.remaining() >= 8; m_S_inpos += 8) {
+         m_S[m_S_inpos / 8] ^= load_le(input_this_round.take<8>());
+      }
+
+      // Read remaining output data, causing misalignment, if necessary
```
randombit: input data?

src/lib/permutations/ascon_perm/info.txt
```diff
@@ -0,0 +1,8 @@
+<internal_defines>
+ASCON_PERM -> 20250816
```
randombit: No need for this I think - all users just hard depend on it rather than checking for the macro

randombit (commented):
Looking good thanks for taking this on

src/lib/permutations/ascon_perm/ascon_perm.h
reneme: The AEAD might need it. Not sure yet. For now I made it private.

src/lib/permutations/ascon_perm/ascon_perm.cpp
reneme: I think that should be tackled later. As mentioned, absorb and squeeze are essentially borrowed from Keccak. I guess both could be optimized and shared. Let's leave that as future work.

src/lib/permutations/ascon_perm/info.txt
reneme: Check.

src/lib/permutations/ascon_perm/ascon_perm.cpp
```diff
@@ -0,0 +1,141 @@
...
+         output_this_round.next_byte() = static_cast<uint8_t>(m_S[m_S_outpos / 8] >> (8 * (m_S_outpos % 8)));
+      }
+
+      // We reached the end of a sponge state block... permute() and start over
+      // TODO: Currently this might perform an unnecessary permutation if the
+      //       to-be-generated output length is divisible by the byte rate,
+      //       for instance with Ascon-Hash256 this is always the case.
+      if(m_S_outpos == byte_rate()) {
+         permute(m_processing_rounds);
+         m_S_outpos = 0;
```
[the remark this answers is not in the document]
reneme: That's a bit too easy, Copilot. :) We can look at that once we tackle [this](https://github.com/randombit/botan/pull/5061#discussion_r2287988682).

src/lib/permutations/ascon_perm/ascon_perm.h
```diff
@@ -0,0 +1,81 @@
...
+         }
+
+         if(m_init_final_rounds > 16) {
+            throw Botan::Invalid_Argument("Invalid Ascon initialization/finalization rounds");
+         }
+
+         if(m_processing_rounds > 16) {
+            throw Botan::Invalid_Argument("Invalid Ascon processing rounds");
+         }
+      }
```
reneme, line 52: I think, this is a noteworthy setup which might be useful in other places.

Essentially, this is the only explicitly defined constructor of this class. It therefore only provides the implicit copy/move constructors, but no default constructor. So users are **forced** to use this constructor to set up new permutation objects. Given that it is also `consteval` they are **forced** to provide configuration constants that are known at compile time. As a result, the argument checks are actually guaranteed to be performed at compile time and a misconfiguration would result in a compiler error.

In a way, this curiously combines the benefits of passing the configuration as template parameters (the static sanity checks) with the simplicity of a traditional runtime-parameterized implementation.

randombit: That's quite nice, thanks for pointing this out

src/lib/permutations/ascon_perm/ascon_perm.cpp
randombit: :+1: Yes fine with me

PR #5062, opened 2025-08-17
BLAKE2s modernization: Resolve TODO and Clang-Tidy warning suppression

KaganCanSit:
Hello Botan Developer Team,

This PR addresses TODO comments and suppressed Clang-Tidy warnings in `blake2s.cpp` while maintaining minimal changes to preserve existing functionality and performance characteristics. I avoided making major changes overall. The list of items I addressed is as follows:
- Update C-style arrays with `std::array`
- Edits for Clang Tidy warnings
     *  `modernize-loop-convert`: Replaced index-based loops with range-based for loops where appropriate
     * `*-special-member-functions`: Added explicit deletion declarations for copy/move operations (I'm still not entirely sure about this!)
     * NOLINTNEXTLINE suppressions removed

When I compiled and tested using the Clang compiler (`Debian clang version 19.1.7`) with the following command, I did not encounter any errors.

```bash
ninja clean && ./configure.py --without-documentation --with-boost --cc=clang --compiler-cache=ccache --build-targets=static,cli,tests --build-tool=ninja && ninja && ./botan-test --test-threads=4 --run-long-tests
```

**Notes**
- Regarding `secure_vector` TODO: The remaining TODO in `blake2s.h` regarding `secure_vector` usage was intentionally left unchanged. Based on @reneme's previous feedback about the need for careful evaluation of `secure_vector` usage, I prefer to address this in a separate discussion/PR to ensure proper implementation.

- Future improvements: Additional modernizations (span usage, [[nodiscard]], etc.) are possible but were excluded to keep this review focused. I'm happy to address these in follow-up PRs if deemed appropriate.

This PR is marked as draft to gather feedback on the modernization approach before final review.
Best regards.

### DIFF
--- a/src/lib/hash/blake2s/blake2s.cpp
+++ b/src/lib/hash/blake2s/blake2s.cpp
@@ -13,6 +13,7 @@
 #include <botan/internal/fmt.h>
 #include <botan/internal/loadstor.h>
 #include <botan/internal/rotate.h>
+#include <botan/internal/stl_util.h>
 
 namespace Botan {
 
@@ -20,12 +21,12 @@ namespace {
 
 // Initialization Vector.
 
-const uint32_t blake2s_iv[8] = {
+constexpr std::array<uint32_t, 8> blake2s_iv{
    0x6A09E667, 0xBB67AE85, 0x3C6EF372, 0xA54FF53A, 0x510E527F, 0x9B05688C, 0x1F83D9AB, 0x5BE0CD19};
 
 // Mixing function G.
 
-inline void B2S_G(uint8_t a, uint8_t b, uint8_t c, uint8_t d, uint32_t x, uint32_t y, uint32_t* v) {
+inline void B2S_G(uint8_t a, uint8_t b, uint8_t c, uint8_t d, uint32_t x, uint32_t y, std::span<uint32_t, 16> v) {
    v[a] = v[a] + v[b] + x;
    v[d] = rotr<16>(v[d] ^ v[a]);
    v[c] = v[c] + v[d];
@@ -46,19 +47,15 @@ std::string BLAKE2s::name() const {
 //      (keylen=0: no key)
 
 void BLAKE2s::state_init(size_t outlen, const uint8_t* key, size_t keylen) {
-   for(size_t i = 0; i < 8; i++) {  // state, "param block"
-      m_h[i] = blake2s_iv[i];
-   }
+   m_h = blake2s_iv;  // state, "param block"
+
    m_h[0] ^= 0x01010000 ^ (keylen << 8) ^ outlen;
 
-   m_t[0] = 0;  // input count low word
-   m_t[1] = 0;  // input count high word
-   m_c = 0;     // pointer within buffer
+   m_t = {0};  // input count (both words zeroed)
+   m_b = {0};  // zero input block
+   m_c = 0;    // pointer within buffer
    m_outlen = outlen;
 
-   for(size_t i = keylen; i < 64; i++) {  // zero input block
-      m_b[i] = 0;
-   }
    if(keylen > 0) {
       add_data(std::span<const uint8_t>(key, keylen));
       m_c = 64;  // at the end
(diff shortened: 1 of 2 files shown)


src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -68,41 +67,40 @@ void BLAKE2s::state_init(size_t outlen, const uint8_t* key, size_t keylen) {
...
+                                                            {2, 12, 6, 10, 0, 11, 8, 3, 4, 13, 7, 5, 15, 14, 1, 9},
+                                                            {12, 5, 1, 15, 14, 13, 4, 10, 0, 7, 6, 3, 9, 2, 8, 11},
+                                                            {13, 11, 7, 14, 12, 1, 3, 9, 5, 0, 15, 4, 8, 6, 2, 10},
+                                                            {6, 15, 14, 9, 11, 3, 0, 8, 12, 2, 13, 7, 1, 4, 10, 5},
+                                                            {10, 2, 8, 4, 7, 6, 1, 5, 15, 11, 9, 14, 3, 12, 13, 0}}};
+
+   std::array<uint32_t, 16> v{};
+   // init work variables
+   std::copy_n(m_h.data(), 8, v.data());
+   std::copy_n(blake2s_iv.data(), 8, v.data() + 8);
```
reneme: (you need to add `botan/internal/stl_util.h` as an include for this)


```suggestion
   // init work variables
   std::array<uint32_t, 16> v = concat(m_h, blake2s_iv);
```

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -68,41 +67,40 @@ void BLAKE2s::state_init(size_t outlen, const uint8_t* key, size_t keylen) {
...
-      B2S_G(1, 5, 9, 13, m[sigma[i][2]], m[sigma[i][3]], v);
-      B2S_G(2, 6, 10, 14, m[sigma[i][4]], m[sigma[i][5]], v);
-      B2S_G(3, 7, 11, 15, m[sigma[i][6]], m[sigma[i][7]], v);
-      B2S_G(0, 5, 10, 15, m[sigma[i][8]], m[sigma[i][9]], v);
-      B2S_G(1, 6, 11, 12, m[sigma[i][10]], m[sigma[i][11]], v);
-      B2S_G(2, 7, 8, 13, m[sigma[i][12]], m[sigma[i][13]], v);
-      B2S_G(3, 4, 9, 14, m[sigma[i][14]], m[sigma[i][15]], v);
+
+   std::array<uint32_t, 16> m{};
+   load_le<uint32_t>(m.data(), m_b.data(), 16);  // get little-endian words
```
reneme: The `load_le` helper is actually more powerful than that which allows us to define `m` as const:


```suggestion
   const auto m = load_le<std::array<uint32_t, 16>>(m_b);  // get little-endian words
```

src/lib/hash/blake2s/blake2s.h
```diff
@@ -39,11 +46,11 @@ class BLAKE2s final : public HashFunction /* NOLINT(*-special-member-functions)
       void compress(bool last);
 
       // TODO use secure_vector here
```
reneme: That can probably just be removed. We did start using `std::array<>` for plenty such states and the class scrubs this memory in its destructor anyway.

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -46,9 +46,8 @@ std::string BLAKE2s::name() const {
 //      (keylen=0: no key)
 
 void BLAKE2s::state_init(size_t outlen, const uint8_t* key, size_t keylen) {
-   for(size_t i = 0; i < 8; i++) {  // state, "param block"
-      m_h[i] = blake2s_iv[i];
-   }
+   std::copy_n(blake2s_iv.data(), 8, m_h.data());  // state, "param block"
+
```
reneme, line 51: For some reason GitHub doesn't let me comment on lines that are unchanged. This is meant to refer to three lines below, the initialization of `m_t`:

Now that `m_t` is a `std::array` we might as well call its constructor:

```C++
m_t = {0};
```

src/lib/hash/blake2s/blake2s.h
KaganCanSit: Okay, I remember your warning about using the secure_vector structure in previous reviews, so it was the right decision to add a note. Thank you.

src/lib/hash/blake2s/blake2s.cpp
KaganCanSit: Well-designed overload: provides clean usage. I will use this as an example for my own projects. Thanks.

KaganCanSit:
Thanks for your review and feedback(@reneme), I think it's ready now. I'm taking it out of draft. 🚀

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -46,14 +47,12 @@ std::string BLAKE2s::name() const {
 //      (keylen=0: no key)
 
 void BLAKE2s::state_init(size_t outlen, const uint8_t* key, size_t keylen) {
-   for(size_t i = 0; i < 8; i++) {  // state, "param block"
-      m_h[i] = blake2s_iv[i];
-   }
+   std::copy_n(blake2s_iv.data(), 8, m_h.data());  // state, "param block"
```
reneme: Sorry, I missed that before. Both types are an array of 32-bit integers with 8 items. The assignment operator should be able to handle this:


```suggestion
   m_h = blake2s_iv;  // state, "param block"
```

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -46,14 +47,12 @@ std::string BLAKE2s::name() const {
...
+
    m_h[0] ^= 0x01010000 ^ (keylen << 8) ^ outlen;
 
-   m_t[0] = 0;  // input count low word
-   m_t[1] = 0;  // input count high word
-   m_c = 0;     // pointer within buffer
+   m_t = {0};  // input count (both words zeroed)
+   m_c = 0;    // pointer within buffer
```
reneme, line 56: As before, GitHub doesn't let me comment on unchanged code. :( This is meant to replace the for-loop setting `m_b` to zero:

```C++
m_b = {0};
```

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -151,9 +145,9 @@ void BLAKE2s::final_result(std::span<uint8_t> out) {
 
 std::unique_ptr<HashFunction> BLAKE2s::copy_state() const {
    std::unique_ptr<BLAKE2s> h = std::make_unique<BLAKE2s>(m_outlen << 3);
-   std::memcpy(h->m_b, m_b, sizeof(m_b));
-   std::memcpy(h->m_h, m_h, sizeof(m_h));
-   std::memcpy(h->m_t, m_t, sizeof(m_t));
+   h->m_b = m_b;
+   h->m_h = m_h;
+   h->m_t = m_t;
    h->m_c = m_c;
```
reneme: Debatable: If you were to define the copy constructor of this type, you could just:


```suggestion
   return std::make_unique<BLAKE2s>(*this);
```

KaganCanSit:
> Sorry, I had missed a few more opportunities to simplify this further.

No problem, I'm happy to help as much as I can. I've made these changes. 

Additionally:
- I used `std::fill` instead of a loop and increment operation in the `final_result` function.
- I used span instead of raw pointer for `B2S_G`. I think this improves both type safety and readability.

Just leave a comment if you want any further changes. 🤝

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -138,10 +130,8 @@ void BLAKE2s::final_result(std::span<uint8_t> out) {
       m_t[1]++;        // high word
    }
 
-   while(m_c < 64) {  // fill up with zeros
-      m_b[m_c++] = 0;
-   }
-   compress(true);  // final block flag = 1
+   std::fill(m_b.begin() + m_c, m_b.end(), 0);  // fill up with zeros
```
KaganCanSit, line 133: Now, after some sleep, I checked here, the `m_c`value wasn't increased. The tests passed successfully, but would it be more appropriate to set the value here?

```cpp
m_c = 64;
// Or
m_c = m_b.size();
```

PR #5066, opened 2025-08-18
Avoid problematic construction in some bit_ops.h functions

randombit:
Some of these functions attempt to be constant time, but were using a problematic construct `s * (mask & 1)` which is quite easy for a compiler to convert into a conditional jump, though no compiler has been observed doing so yet.

Instead generate the mask such that it is as big as a size_t and then use `s & mask`

### DIFF
--- a/src/lib/utils/bit_ops.h
+++ b/src/lib/utils/bit_ops.h
@@ -21,7 +21,7 @@
 namespace Botan {
 
 /**
-* If top bit of arg is set, return ~0. Otherwise return 0.
+* If top bit of arg is set, return |1| (all bits set). Otherwise return |0| (all bits unset)
 */
 template <std::unsigned_integral T>
 BOTAN_FORCE_INLINE constexpr T expand_top_bit(T a) {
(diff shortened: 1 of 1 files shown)


src/lib/utils/bit_ops.h
```diff
@@ -54,10 +54,13 @@ BOTAN_FORCE_INLINE constexpr bool is_power_of_2(T arg) {
 */
 template <std::unsigned_integral T>
 BOTAN_FORCE_INLINE constexpr size_t high_bit(T n) {
+   static_assert(sizeof(T) <= sizeof(size_t));
```
reneme: I recently saw a compelling argument for preferring C++20's `requires` clauses over checking using `static_assert` in the template function's body: The latter breaks SFINAE! This may or may not apply here, but it is definitely worth to be aware of.

Say, we wanted to write code that uses `if constexpr` to guard the usage of `high_bit()` based on whether or not it is compatible with a certain type. To figure that out we usually use some SFINAE-construction. Like so ([full example on Godbolt](https://godbolt.org/z/46zrx9hTr)):

```C++
template <typename T>
auto get_high_bit(T x) {
   if constexpr(is_high_bit_helper_invocable<T>::value) {
      return high_bit(x);
   } else {
      return /* ... use some alternative method to get the high bit of T. */;
   }
}
```

However, if `static_assert` is used this is not possible. The compiler would interpret the assertion and error out instead of just taking the alternative path of the `if constexpr`.

As a general rule-of-thumb, I would therefore suggest to prefer `requires` clauses over such `static_assert` to check type constraints of templates.

reneme (approved):
Changes look good to me. One suggestion/argument regarding the usage of `requires` over `static_assert`.

src/lib/utils/bit_ops.h
randombit: Interesting point. I guess for me the preference for `static_assert` comes from if you violate that you get a clear message that an assertion failed, where if you fail to satisfy a `requires` clause, IIRC the error message is much less clear (the compiler in theory could tell you, "I would have used this implementation but you failed to satisfy the required clause X because Y" but I'm not sure that any compiler does)

In this specific case, to correctly handle 32-bit systems I had to rewrite the code in such a way that there is no `static_assert` or `requires` necessary.

PR #5068, opened 2025-08-18
Add explicit algorithm for variable-time division of 2^k / x

randombit:
So far this is just a small simplification on the Knuth division but hopefully something better can be devised in the future. Speedup of about 15% for smaller integers, basically no improvement at 4096 bits though.

### DIFF
--- a/src/lib/math/bigint/divide.cpp
+++ b/src/lib/math/bigint/divide.cpp
@@ -222,12 +222,109 @@ BigInt ct_modulo(const BigInt& x, const BigInt& y) {
    return r;
 }
 
+BigInt vartime_divide_pow2k(size_t k, const BigInt& y_arg) {
+   constexpr size_t WB = WordInfo<word>::bits;
+
+   BOTAN_ARG_CHECK(!y_arg.is_zero(), "Cannot divide by zero");
+   BOTAN_ARG_CHECK(!y_arg.is_negative(), "Negative divisor not supported");
+   BOTAN_ARG_CHECK(k > 1, "Invalid k");
+
+   BigInt y = y_arg;
+
+   const size_t y_words = y.sig_words();
+
+   BOTAN_ASSERT_NOMSG(y_words > 0);
+
+   // Calculate shifts needed to normalize y with high bit set
+   const size_t shifts = y.top_bits_free();
+
+   if(shifts > 0) {
+      y <<= shifts;
+   }
+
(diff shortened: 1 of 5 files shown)


src/lib/math/bigint/divide.cpp
```diff
@@ -222,10 +222,97 @@ BigInt ct_modulo(const BigInt& x, const BigInt& y) {
...
+   if(shifts > 0) {
+      y <<= shifts;
+   }
+
+   BigInt r;
+   r.set_bit(k + shifts);  // (2^k) << shifts
+
+   // we know y has not changed size, since we only shifted up to set high bit
+   const size_t t = y_words - 1;
+   const size_t n = std::max(y_words, r.sig_words()) - 1;  // r may have changed size however
```
randombit: This value can be computed directly from `k + shifts`

src/lib/math/bigint/divide.cpp
```diff
@@ -222,10 +222,97 @@ BigInt ct_modulo(const BigInt& x, const BigInt& y) {
...
+      * the generic divide it is hit at a tenth that rate on average.
+      */
+      if(qit == 1) {
+         r -= shifted_y;
+      } else {
+         r -= qit * shifted_y;
+      }
+
+      if(r.is_negative()) {
+         qit--;
```
randombit: Should have an assert here that qit > 0

src/lib/math/bigint/divide.cpp
```diff
@@ -222,10 +222,97 @@ BigInt ct_modulo(const BigInt& x, const BigInt& y) {
...
+
+      /*
+      * Special case qit == 1 which occurs relatively often here due to a
+      * combination of the fixed 2^k and in many cases the typical structure of
+      * public moduli (as this function is called by Barrett_Reduction::for_public_modulus).
+      *
+      * This case occurs as often as 5% of the loop iterations; in constrast in
+      * the generic divide it is hit at a tenth that rate on average.
+      */
+      if(qit == 1) {
```
randombit: Does qit ever equal 0?

randombit: Yes, quite common in this case and an easy shortcut

src/lib/math/bigint/divide.cpp
randombit: Not worth it

PR #5072, opened 2025-08-22
find x509 cert by issuer dn and serial number

Brassinolide:
In CMS or PKCS#7, it is often necessary to use the issuer's DN and certificate serial number to find the signer certificate.

```ASN.1
IssuerAndSerialNumber ::= SEQUENCE {
    issuer Name,
    serialNumber CertificateSerialNumber
}
```

### DIFF
--- a/src/lib/x509/certstor.h
+++ b/src/lib/x509/certstor.h
@@ -57,6 +57,16 @@ class BOTAN_PUBLIC_API(2, 0) Certificate_Store /* NOLINT(*-special-member-functi
          const std::vector<uint8_t>& subject_hash) const = 0;
 
       /**
+      * Find a certificate by searching for one with a matching issuer DN and
+      * serial number. Used for CMS or PKCS#7.
+      * @param issuer_dn the distinguished name of the issuer
+      * @param serial_number the certificate's serial number
+      * @return a matching certificate or nullopt otherwise
+      */
+      virtual std::optional<X509_Certificate> find_cert_by_issuer_dn_and_serial_number(
+         const X509_DN& issuer_dn, std::span<const uint8_t> serial_number) const = 0;
+
+      /**
       * Finds a CRL for the given certificate
       * @param subject the subject certificate
       * @return the CRL for subject or nullopt otherwise
@@ -133,6 +143,9 @@ class BOTAN_PUBLIC_API(2, 0) Certificate_Store_In_Memory final : public Certific
 
       std::optional<X509_Certificate> find_cert_by_raw_subject_dn_sha256(
          const std::vector<uint8_t>& subject_hash) const override;
+
+      std::optional<X509_Certificate> find_cert_by_issuer_dn_and_serial_number(
+         const X509_DN& issuer_dn, std::span<const uint8_t> serial_number) const override;
 
       /**
       * Finds a CRL for the given certificate
--- a/src/lib/x509/certstor_flatfile/certstor_flatfile.cpp
+++ b/src/lib/x509/certstor_flatfile/certstor_flatfile.cpp
@@ -56,6 +56,7 @@ Flatfile_Certificate_Store::Flatfile_Certificate_Store(std::string_view file, bo
          m_dn_to_cert[cert.subject_dn()].push_back(cert);
          m_pubkey_sha1_to_cert.emplace(cert.subject_public_key_bitstring_sha1(), cert);
          m_subject_dn_sha256_to_cert.emplace(cert.raw_subject_dn_sha256(), cert);
+         m_issuer_dn_to_cert[cert.issuer_dn()].push_back(cert);
       } else if(!ignore_non_ca) {
          throw Invalid_Argument("Flatfile_Certificate_Store received non CA cert " + cert.subject_dn().to_string());
       }
@@ -72,17 +73,17 @@ std::vector<X509_DN> Flatfile_Certificate_Store::all_subjects() const {
 
 std::vector<X509_Certificate> Flatfile_Certificate_Store::find_all_certs(const X509_DN& subject_dn,
                                                                          const std::vector<uint8_t>& key_id) const {
-   std::vector<X509_Certificate> found_certs;
-   try {
-      const auto certs = m_dn_to_cert.at(subject_dn);
-
-      for(const auto& cert : certs) {
-         if(key_id.empty() || key_id == cert.subject_key_id()) {
-            found_certs.push_back(cert);
-         }
-      }
-   } catch(const std::out_of_range&) {
+   if(!m_dn_to_cert.contains(subject_dn)) {
       return {};
+   }
+
+   const auto& certs = m_dn_to_cert.at(subject_dn);
+
+   std::vector<X509_Certificate> found_certs;
+   for(const auto& cert : certs) {
+      if(key_id.empty() || key_id == cert.subject_key_id()) {
+         found_certs.push_back(cert);
+      }
    }
 
    return found_certs;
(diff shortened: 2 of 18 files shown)


src/lib/x509/certstor.h
```diff
@@ -56,6 +56,16 @@ class BOTAN_PUBLIC_API(2, 0) Certificate_Store /* NOLINT(*-special-member-functi
...
 
+      /**
+      * Find a certificate by searching for one with a matching issuer DN and
+      * serial number. Used for CMS or PKCS#7.
+      * @param issuer_dn the distinguished name of the issuer
+      * @param serial_number the certificate's serial number
+      * @return a matching certificate or nullopt otherwise
+      */
+      virtual std::optional<X509_Certificate> find_cert_by_issuer_dn_and_serial_number(
+         const X509_DN& issuer_dn, const std::vector<uint8_t>& serial_number) const = 0;
```
randombit: Use `span` for input bytestrings (we don't do this consistently only because we were able to use span only relatively recently)

src/lib/x509/certstor_system_macos/certstor_macos.cpp
```diff
@@ -392,6 +392,13 @@ std::optional<X509_Certificate> Certificate_Store_MacOS::find_cert_by_raw_subjec
    throw Not_Implemented("Certificate_Store_MacOS::find_cert_by_raw_subject_dn_sha256");
 }
 
+std::optional<X509_Certificate> Certificate_Store_MacOS::find_cert_by_issuer_dn_and_serial_number(
+   const X509_DN& issuer_dn, const std::vector<uint8_t>& serial_number) const {
+   BOTAN_UNUSED(issuer_dn);
+   BOTAN_UNUSED(serial_number);
+   throw Not_Implemented("Certificate_Store_MacOS::find_cert_by_issuer_dn_and_serial_number");
```
randombit: @reneme any idea if this can be implemented using the Mac cert api?

src/lib/x509/certstor_flatfile/certstor_flatfile.cpp
```diff
@@ -118,6 +119,23 @@ std::optional<X509_Certificate> Flatfile_Certificate_Store::find_cert_by_raw_sub
    return std::nullopt;
 }
 
+std::optional<X509_Certificate> Flatfile_Certificate_Store::find_cert_by_issuer_dn_and_serial_number(
+   const X509_DN& issuer_dn, const std::vector<uint8_t>& serial_number) const {
+   try {
```
randombit: Use contains instead of a try/catch to handle the not found case

randombit:
Looks generally ok. Needs tests

Brassinolide:
I use `std::ranges::equal` instead of `std::equal` for comparison because `std::ranges::equal` is supported by gcc 11.2 and clang 14 and vs2022 (listed in https://github.com/randombit/botan/blob/master/doc/support.rst#supported-platforms )

https://github.com/randombit/botan/blob/0852a2a74a6823a103368db6a835d4a5f98d0c1a/src/lib/asn1/asn1_oid.cpp#L156

src/lib/x509/certstor_system_macos/certstor_macos.cpp
Brassinolide:
```C++
std::optional<X509_Certificate> Certificate_Store_MacOS::find_cert_by_issuer_dn_and_serial_number(
   const X509_DN& issuer_dn, std::span<const uint8_t> serial_number) const {
   Certificate_Store_MacOS_Impl::Query query;
   query.addParameter(kSecAttrIssuer, normalizeAndSerialize(issuer_dn));
   query.addParameter(kSecAttrSerialNumber, serial_number);

   return m_impl->findOne(std::move(query));
}
```

I added the code for macOS, but the test fails because it can't find the certificate.

I don't think it's an issue with the X509_DN, as other tests using the DN pass. Could it be an issue with the serial_number? However, the test passes on other systems. Does macOS have some special encoding rules for the serial_number?

Brassinolide:
@randombit The main work has been completed, but the certificate search under macOS is still a problem. I have tried many methods but none of them worked, and there is no relevant reference in the official documentation. I might be able to find the certificate by `issuer_dn` only, then use the `findAll` method and use a loop to filter the `serial_number`, but this may affect the efficiency. Should I do this or wait for others to solve it?

Brassinolide:
Finally, I used this method to find the certificate, and it worked. Why can't I find the certificate directly using `kSecAttrSerialNumber`?

Does macOS require any special encoding for SerialNumber? I can't find it in the development documentation.

```C++
std::optional<X509_Certificate> Certificate_Store_MacOS::find_cert_by_issuer_dn_and_serial_number(
   const X509_DN& issuer_dn, std::span<const uint8_t> serial_number) const {
   Certificate_Store_MacOS_Impl::Query query;
   query.addParameter(kSecAttrIssuer, normalizeAndSerialize(issuer_dn));

   for(const auto& cert : m_impl->findAll(std::move(query))) {
      if(std::ranges::equal(cert.serial_number(), serial_number)) {
         return cert;
      }
   }

   return std::nullopt;
}
```

src/lib/x509/certstor_system_macos/certstor_macos.h
```diff
@@ -63,6 +63,12 @@ class BOTAN_PUBLIC_API(2, 10) Certificate_Store_MacOS final : public Certificate
       std::optional<X509_Certificate> find_cert_by_raw_subject_dn_sha256(
          const std::vector<uint8_t>& subject_hash) const override;
 
+      /**
+       * @throws Not_Implemented
+       */
```
randombit: Now incorrect right?

src/lib/x509/certstor_system_macos/certstor_macos.cpp
```diff
@@ -392,6 +392,20 @@ std::optional<X509_Certificate> Certificate_Store_MacOS::find_cert_by_raw_subjec
    throw Not_Implemented("Certificate_Store_MacOS::find_cert_by_raw_subject_dn_sha256");
 }
 
+std::optional<X509_Certificate> Certificate_Store_MacOS::find_cert_by_issuer_dn_and_serial_number(
+   const X509_DN& issuer_dn, std::span<const uint8_t> serial_number) const {
+   Certificate_Store_MacOS_Impl::Query query;
+   query.addParameter(kSecAttrIssuer, normalizeAndSerialize(issuer_dn));
+
```
randombit, line 405: Add a comment/todo here that the API for searching by serial did not seem to work and is left for future investigation.

src/tests/test_certstor_flatfile.cpp
```diff
@@ -228,6 +228,27 @@ Test::Result certstore_contains_user_certificate() {
...
+   try {
+      result.start_timer();
+      Botan::Flatfile_Certificate_Store certstore(get_valid_ca_bundle_path());
+      auto cert = certstore.find_cert_by_issuer_dn_and_serial_number(get_dn(), get_serial_number());
+      result.end_timer();
+
+      if(result.test_not_nullopt("found certificate", cert)) {
+         auto cns = cert->subject_dn().get_attribute("CN");
+         result.test_int_eq("exactly one CN", cns.size(), 1);
+         result.test_eq("CN", cns.front(), get_subject_cn());
```
randombit, line 243: Should check here the serial matches too!

src/tests/test_certstor_system.cpp
```diff
@@ -227,6 +227,26 @@ Test::Result find_all_subjects(Botan::Certificate_Store& certstore) {
...
+
+   try {
+      result.start_timer();
+      auto cert = certstore.find_cert_by_issuer_dn_and_serial_number(get_dn(), get_serial_number());
+      result.end_timer();
+
+      if(result.test_not_nullopt("found certificate", cert)) {
+         auto cns = cert->subject_dn().get_attribute("CN");
+         result.test_is_eq("exactly one CN", cns.size(), size_t(1));
+         result.test_eq("CN", cns.front(), get_subject_cn());
```
randombit, line 241: Should check for the serial here too

Brassinolide:
force-pushed to rebase + squash

randombit (approved):

PR #5076, opened 2025-08-27
Feature: Ascon-AEAD128

### LINKED ISSUE #4943: Feature Request: ASCON

reneme:
This needs serious cleanup, but it does pass the KATs of the reference implementation.

Closes #4943


randombit:
Should get a mention in the relevant docs

src/lib/modes/aead/ascon_aead128/ascon_aead128.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+void Ascon_AEAD128_Mode::key_schedule(std::span<const uint8_t> key) {
+   clear();
+   m_key = as_array_of_uint64<2>(key);
+}
+
+void Ascon_AEAD128_Mode::set_associated_data_n(size_t idx, std::span<const uint8_t> ad) {
+   BOTAN_ARG_CHECK(idx == 0, "Ascon-AEAD128: cannot handle non-zero index in set_associated_data_n");
+
+   BOTAN_STATE_CHECK(has_keying_material());
```
randombit: This (and `associated_data_requires_key` returning true) doesn't make sense to me - we are just assigning to `m_ad` here without needing keying material afaict

reneme: I'm guessing this is a left-over from a previous iteration where I wanted to process the AD right away without storing it in the internal member. This would have required the key, of course.

reneme:
Something went wrong here. The current state of this branch does not contain the integration in `aead.cpp`, so `AEAD::create("Ascon-AEAD128")` returns a nullptr. Hence, in this current state the Ascon-AEAD KATs don't run (aka: "succeed"). That must have gotten lost in a rebase back when I was finishing the work on this.

Now, after adding the integration back, the tests fail. :( I'll have to spend some time debugging what has happened here. That won't happen tonight anymore I'm afraid.

reneme:
Two things were wrong and are fixed now:

1. after setting up the initial sponge state (containing the static IV, the key and the nonce), a 12-round permutation must be performed (before it was an 8-round permutation). I guess I actually lost that in rebasing, but can't be sure anymore.
2. there was a bug in the `PartialWordBounds::masked_assignment()` (which is used only in the decryption of Ascon-AEAD). If this method was used with `length=0`, the mask creation would read like `(uint64_t(0) - 1) >> (8 - 0) * 8` (i.e. an rshift of 64 of a 64-bit integer -> undefined behaviour). This can happen if one calls `AEAD_Mode::update()` with an empty input buffer and it was caught by the AEAD unittests.

src/lib/modes/aead/ascon_aead128/ascon_aead128.h
```diff
@@ -0,0 +1,99 @@
...
+
+namespace Botan {
+
+class Ascon_AEAD128_Mode : public AEAD_Mode {
+   public:
+      void set_associated_data_n(size_t idx, std::span<const uint8_t> ad) final;
+
+      bool associated_data_requires_key() const final { return false; }
+
+      std::string name() const final { return "Ascon_AEAD128"; }
```
randombit: This doesn't match the value that is matched on in `AEAD::create`.

I *thought* there was a test for this (that names round trip) but maybe it's only for some types.

src/lib/modes/aead/ascon_aead128/ascon_aead128.h
```diff
@@ -0,0 +1,99 @@
...
+
+      size_t update_granularity() const final { return 1; }
+
+      size_t ideal_granularity() const final { return 32; }
```
randombit, line 31: Seems kind of low tbh

doc/api_ref/cipher_modes.rst
```diff
@@ -350,6 +350,25 @@ Both ChaCha20Poly1305 and AES with GCM are widely implemented. SIV is somewhat
...
+small footprint implementations. Its main use case is in constrained
+environments, such as IoT devices where traditional cryptographic functions
+may be too resource intensive.
+
+Unless you are interoperating with an existing device which due to resource
+constraints can only use Ascon, prefer more typical AEADs such as AES-256/GCM,
+AES-256/SIV, or ChaCha20Poly1305.
+
+This AEAD scheme is standardized by NIST in SP.800-232. It is not compatible
+with earlier versions of the Ascon specification.
```
randombit: Maybe mention also the options like nonce masking that are defined in the SP but not supported.

src/lib/modes/aead/ascon_aead128/ascon_aead128.h
reneme: Good catch!

src/lib/modes/aead/ascon_aead128/ascon_aead128.h
reneme: What would be a more reasonable value? E.g. for "AES/GCM" this is `BLOCK_SIZE*std::max(2, AES_parallelism)`, so also just 32 in the worst case.

randombit: Hmm fair enough. Both seem low then :) but that calls for a more systematic evaluation

randombit (approved):

PR #5078, opened 2025-08-27
Update homebrew db to pick up fixed Emscripten package

randombit:


randombit:
Still broken ... guess we need to wait for the next Emscripten release ...

reneme:
Meh! 😑 Perhaps there's a better way to install Emscripten than homebrew on macOS. I recall that we switched it to macOS because it wasn't readily available on Linux. Maybe that changed in the meantime and we could rely on a more stable package?

Homebrew always forces the bleeding edge on us, and it's not the first time this breaks things...

randombit:
Yeah unfortunately we are stuck between Homebrew with bleeding edge just released stuff, and Ubuntu where most software is first aged in Debian, then given LTS status by Ubuntu, ie thoroughly obsolete by the time it gets to us. Specifically Ubuntu 24.04's Emscripten is 3.1.6 which was released almost 4 years ago (!) and I guess [does not fully support C++20 concepts](https://github.com/randombit/botan/actions/runs/17293149842/job/49084993641?pr=5079#step:5:62)

PR #5079, opened 2025-08-28
Move Emscripten build to Linux

randombit:
Since Homebrew seems content shipping a broken Emscripten for weeks on end ... https://github.com/Homebrew/homebrew-core/issues/234170


randombit:
Ubuntu solves this problem by shipping a 4 year old thoroughly obsolete version of Emscripten instead.

PR #5081, opened 2025-08-29
Allow Empty Issuer in Certificate_Extension::validate

FAlbertDev:
See https://github.com/randombit/botan/pull/5047#discussion_r2306929703. Required for #5047.

Changes the interface of `Certificate_Extension::validate` to use an optional issuer.
image

Currently, all `validate` overloads ignore the issuer parameter, so there are no other relevant logic changes. 

While the new interface breaks the current API, existing calls to the `validate` function will be correctly deduced in most cases (see [godbolt](https://godbolt.org/z/d9sY9P7cd)).

### DIFF
--- a/src/lib/x509/pkix_types.h
+++ b/src/lib/x509/pkix_types.h
@@ -446,13 +446,14 @@ class BOTAN_PUBLIC_API(2, 0) Certificate_Extension /* NOLINT(*-special-member-fu
       * an appropriate status code shall be added to cert_status.
       *
       * @param subject Subject certificate that contains this extension
-      * @param issuer Issuer certificate
+      * @param issuer Issuer certificate. nullopt for certificates with no
+      *        available issuer (e.g. non self-signed trust anchors).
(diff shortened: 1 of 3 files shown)


src/lib/x509/pkix_types.h
```diff
@@ -446,13 +446,14 @@ class BOTAN_PUBLIC_API(2, 0) Certificate_Extension /* NOLINT(*-special-member-fu
       * an appropriate status code shall be added to cert_status.
       *
       * @param subject Subject certificate that contains this extension
-      * @param issuer Issuer certificate
+      * @param issuer Issuer certificate. nullopt for certificates with no
+      *        issuer (e.g. non self-signed trust anchors).
```
reneme:
```suggestion
      * @param issuer Issuer certificate. nullopt for certificates with no
      *        available issuer (e.g. non self-signed trust anchors).
```

reneme (approved):

PR #5082, opened 2025-08-29
FIX: passing raw bytes into PKSign in botan3.py

reneme:
Previously, the `_ctype_str()` wrapper caused an error if the passed variable wasn't of type `str`. This spoils it for many use cases of `PKSign`, because raw data cannot be reliably transformed into UTF-8 and back. I'm guessing that this was caused by copy-pasta, because all other `PK...` wrappers (including `PKVerify`) use `_ctype_bits()`. I'll add a (regression) test for it to this PR shortly.

Note that [downstream user even seem to monkey patch this](https://gitlab.com/archive-rip/slh_dsa_botan3/-/blob/27b48c5c20a760af4271f69d69a09548baa14ff2/slh_dsa_botan3/__init__.py#L130-140). For what it's worth, here's the workaround we're currently using for this (directly calling through the FFI API):

```python
s = botan3.c_void_p(0)
mode = "Raw".encode()  # directly sign the hash
flags = botan3.c_uint32(1)  # DER encoding of signature
botan3._DLL.botan_pk_op_sign_create(botan3.byref(s), self._sk.handle_(), mode, flags)

try:
    botan3._DLL.botan_pk_op_sign_update(s, hash, len(hash))
    outbuf_sz = botan3.c_size_t(0)
    botan3._DLL.botan_pk_op_sign_output_length(s, botan3.byref(outbuf_sz))
    outbuf = botan3.create_string_buffer(outbuf_sz.value)
    botan3._DLL.botan_pk_op_sign_finish(s, rng.handle_(), outbuf, botan3.byref(outbuf_sz))
    return _to_bytes(outbuf.raw[0 : int(outbuf_sz.value)])
finally:
    botan3._DLL.botan_pk_op_sign_destroy(s)
```

### DIFF
--- a/src/scripts/test_python.py
+++ b/src/scripts/test_python.py
@@ -462,6 +462,53 @@ ofvkP1EDmpx50fHLawIDAQAB
         except botan.BotanException as e:
             self.assertEqual(str(e), "botan_pubkey_load_rsa failed: -1 (Invalid input): Invalid RSA public key parameters")
 
+    def _pksign_roundtrips(self, sk, pk, param_str):
+        def verify_positive_and_negative(verifier, sig):
+            self.assertTrue(verifier.check_signature(sig))
+            invalid_sig = bytes(sig[0] ^ 0xFF) + sig[1:]
+            self.assertFalse(verifier.check_signature(invalid_sig))
+
+        rng = botan.RandomNumberGenerator()
+        msg = "test message"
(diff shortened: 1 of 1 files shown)


randombit (approved):
Thanks

reneme:
> I'll add a (regression) test for it to this PR shortly.

Done.

PR #5083, opened 2025-08-29
Allow loading (compressed) SEC.1-encoded ECC keys via FFI and Python

reneme:
## Desired Use Case Description

Currently, loading a SEC.1 encoded ECC key doesn't seem to be possible conveniently via the FFI. It provides two suitable functions: `botan_pubkey_load()` which expects a PEM container and the low-level `botan_pubkey_load_ecdh()` which requires both coordinates of the public point. Unfortunately in my use case I only have a compressed SEC.1 encoding of an ECDH public key, so `*_load_ecdh()` is not really feasible. As a work-around I am wrapping this encoding into a PEM container by hard-coding the ASN.1 prologue for my curve. Meh!

## Changes in this PR

This adds three new FFI functions: `botan_pubkey_load_ecdsa_sec1()`, `botan_pubkey_load_ecdh_sec1()`, and `botan_pubkey_load_sm2_sec1()` to facilitate the above-described use case. Also, I exposed them via the Python wrapper like `botan3.PublicKey.load_ecdh_sec1()`, ...

Also, this PR bumps the FFI API requirement. Which I think is necessary because of the new APIs, right? I refrained from documenting those new functions, given that the existing low-level functions aren't explicitly documented either.

I'm certainly open for better ideas for naming or approach. Or perhaps I missed some way to do that on master?


src/tests/test_ffi.cpp
```diff
@@ -2747,6 +2747,72 @@ class FFI_DSA_Test final : public FFI_Test {
...
+ * Helper class for testing "view"-style API functions that take a callback
+ * that gets passed a variable-length buffer of bytes.
+ *
+ * Example:
+ *   botan_privkey_t priv;
+ *   ViewBytesSink sink;
+ *   botan_privkey_view_raw(priv, sink.delegate(), sink.callback());
+ *   std::cout << hex_encode(sink.get()) << std::endl;
+ */
+class ViewBytesSink final {
```
arckoor: I would propose just moving this to the top of the file, I had to move it for #4877 too, and the Certificate tests are pretty far up.

src/scripts/test_python.py
```diff
@@ -609,6 +616,11 @@ def test_ecdsa(self):
         priv2 = botan.PrivateKey.load_ecdsa(group, priv.get_field('x'))
         self._pksign_roundtrips(priv2, pub, hash_fn)
 
+        # Load public key from SEC.1 encoding
+        compressed_sec1 = BotanPythonTests._ecc_sec1_convert_to_compressed(pub.to_raw())
+        pub3 = botan.PublicKey.load_ecdsa_sec1(group, compressed_sec1)
```
randombit: Should we be testing both compressed and uncompressed here?

src/lib/ffi/ffi.cpp
```diff
@@ -219,6 +219,11 @@ uint32_t botan_ffi_api_version() {
 }
 
 int botan_ffi_supports_api(uint32_t api_version) {
+   // This is the API introduced in 3.10
+   if(api_version == 20250829) {
```
randombit, line 223: This should be listed in the table in ffi.rst

src/tests/test_ffi.cpp
reneme: Fair point. Done.

src/scripts/test_python.py
reneme: Now all three (ECDH, ECDSA, SM2) test both compressed and uncompressed.

PR #5085, opened 2025-09-01
Replace `std::lexicographical_compare` with `three_way` variant in Session_ID `operator<` - TODO

KaganCanSit:
Hello Botan Development Team,

This development was carried out to resolve the TODO comment in `tls_session.h`. The `lexicographical_compare` call has been updated to `lexicographical_compare_three_way` as specified.

Here, the function's use of the `<` operator can be extended to the `<=>` operator. However, I don't think this is necessary at the moment. Of course, I can make this change if you wish.

I did not encounter any errors in “Compiler Explorer” using GCC11.4 and Clang 14.0. I hope it compiles. I compiled it in a local environment, but my versions are high. (Debian 13 - Clang 19.1.7 - GCC 14.2.0)

I'm not entirely sure about testing the change here. I ran general tests and additionally used the following command.

```bash
python3 src/scripts/test_cli.py ./botan cli_tls
```

I can make any changes you request. I'm open to your feedback.

Best regards.

### DIFF
--- a/src/lib/tls/tls_session.h
+++ b/src/lib/tls/tls_session.h
@@ -38,10 +38,7 @@ using Session_Ticket = Strong<std::vector<uint8_t>, struct Session_Ticket_>;
 using Opaque_Session_Handle = Strong<std::vector<uint8_t>, struct Opaque_Session_Handle_>;
 
 inline auto operator<(const Session_ID& id1, const Session_ID& id2) {
-   // TODO: C++20 better use std::lexicographical_compare_three_way
-   //       that was not available on all target platforms at the time
-   //       of this writing.
-   return std::lexicographical_compare(id1.begin(), id1.end(), id2.begin(), id2.end());
+   return std::is_lt(std::lexicographical_compare_three_way(id1.begin(), id1.end(), id2.begin(), id2.end()));
 }
 
 /**


KaganCanSit:
Even though I checked via “godbolt,” macOS wasn't on my mind. I guess it needs time. Frankly, I was hopeful because it's been two years since TODO was written. I'll close the branch after all the compilations are complete and I see the status. It might be good for you to know. (@reneme)

```bash
NOTE
INFO: Python version: "3.13.7 (v3.13.7:bcee1c32211, Aug 14 2025, 19:10:51) [Clang 16.0.0 (clang-1600.0.26.6)]"
Godbolt -> https://godbolt.org/z/qx5fKT6Y6 | -std=c++20 | OK
Godbolt -> https://godbolt.org/z/j7G5ePbf1 | -std=c++20 -stdlib=libc++ | NOT OK
```

I have multiple branches that are blocked by macOS compatibility requirements. Would it be easier for you to track if I reported these improvements under one issue? Or are they already checked when updated? (This branch and #5063) 

I can also scan for TODO messages in the code. However, since I don't have a macOS device, it might require me to test it in CI. I'm not sure.

randombit:
I think for the short term we're stuck with whatever XCode 15 supports since that's the latest version supported on the `macos-13` runners, which are the final version that GH Actions supports that targets macOS x86. The good (?) news is that Github is deprecating the `macos-13` runners entirely, and they will be disabled in November. At that point we will switch entirely to `macos-15` and whatever the latest XCode is supported on those images. This will likely improve the situation, though we're still stuck with various GCC 11 and Clang 14 limitations (especially source_location). Given the timing of the shutdown I guess that will occur shortly after the 3.10 release in November.

KaganCanSit:
> I think for the short term we're stuck with whatever XCode 15 supports since that's the latest version supported on the `macos-13` runners, which are the final version that GH Actions supports that targets macOS x86. The good (?) news is that Github is deprecating the `macos-13` runners entirely, and they will be disabled in November. At that point we will switch entirely to `macos-15` and whatever the latest XCode is supported on those images. This will likely improve the situation, though we're still stuck with various GCC 11 and Clang 14 limitations (especially source_location). Given the timing of the shutdown I guess that will occur shortly after the 3.10 release in November.


Good to know that an update for macOS is coming soon, thank you very much for the information. I can send the evaluation of the two branches mentioned here after MacOS update. (If I not catch, please ping)

If there are additional contributions I can make for you, feel free to direct me. I am open to do my best to learn and experiment.

See you soon. 🫡

PR #5086, opened 2025-09-03
Type hints for botan3.py

arckoor:
When using `botan3.py` one often has to guess the required types when calling functions (e.g. `def foo(self, key, ...)` -> is `key` a `str`? `bytes`? `PublicKey`? `PrivateKey`? who knows?)
This adds type hints for such cases.

Additionally, as hinted at in #5040, I would like to introduce a formatter for the python side.
[black](https://github.com/psf/black) remains my favorite, and I've applied it on defaults to the entire codebase, so let me know what you think. If this is acceptable, I would also add a CI job and maybe tidy a few places that lead to weird formatting (like the `# # Section #` comments in `botan3.py`)


reneme:
Nice! When everything is agreed upon and ready for merge. Please add the reformatting commit sha to https://github.com/randombit/botan/blob/9857b91cba6214ef43cbe1d58ca78eaeaebdae1d/src/configs/git-blame-ignore-revs with a comment. That way, `git blame` skips that commit and stays useful (if set up correctly).

randombit:
So definitely on board with type hints for Python - started on this for configure.py a while back and got bogged down, but it certainly makes the code easier to understand and use. Do we need to do anything special so the type hints are checked by Python? I can't remember anymore if there is a special tool or Python runtime mode we have to use to check the hints.

Also on board with formatting. I will introduce a small bikeshed here and say I had planned on doing this using `ruff` rather than black, largely because ruff is very fast and also can probably replace the (quite slow) pylint for linting checks. For formatting the two tools are apparently mostly compatible outside of some corner cases. I guess one downside of ruff is it is not quite as widely distributed yet (eg not in Debian/Ubuntu) so maybe a bit harder for people to get started [*]

[*] That said for C++ formatting we're stuck on one particular version of `clang-format` which is certainly more of a pain in the ass to get then installing ruff via pip.

I'd prefer also that a formatting-only change happens not just in its own commit but it's own PR, so please split this into type hints PR + formatting PR which can be merged seperately.

randombit (approved):
lgtm

PR #5089, opened 2025-09-03
Fix remaining warnings from the ruff linter

randombit:


randombit:
All in the default ruff configuration that is, there are ton with `--select=ALL`

PR #5090, opened 2025-09-03
Add running ruff linter to CI

randombit:


randombit:
ruff is working ok but for whatever reason github isn't picking up on the annotations to inline into the file view, I cannot figure out why :/

randombit:
Possibly related? https://github.com/astral-sh/ruff/issues/17433

reneme:
Perhaps it works better using the [dedicated ruff action](https://docs.astral.sh/ruff/integrations/#github-actions)?

randombit:
Maybe so but given they have an example in that doc doing exactly what's being done here it's surprising it doesn't seem to work.

PR #5092, opened 2025-09-03
Fix some typing errors in botan3.py

randombit:
pyright reports others that are not so easily fixable

cc @arckoor


arckoor:
Oops.
By the way, do we need to update `python.rst` too?

randombit:
> By the way, do we need to update python.rst too?

I guess we don't *need* to, since the type annotations are optional. The additional context would probably be appreciated by users though!

Issue #5087, opened 2025-09-03
Do we need a devcontainer.json?


reneme:
Botan has very minimal 3rd-party dependencies, making it fairly simple to build from source for new contributors. Therefore, just the (implicit) toolchain requirements don't really justify the introduction of a [devcontainer.json](https://containers.dev/) (or similar setups).

However, we do have plenty of tool configuration files hidden away in `src/editors`, `src/configs` and `src/scripts` to keep the repository's root directory clutter-free. As a result, any first-time interaction with the code base isn't as frictionless as it could be. At least until one reads [contributing.rst](https://botan.randombit.net/handbook/dev_ref/contributing.html) carefully.

My suggestion: let's introduce a top-level `devcontainer.json` that then automagically sets up all the configurations and goodies we have stashed away at the moment. I'm thinking of things like `.vscode/`, `pyproject.toml`, `.clang-format`, `.editorconfig` and perhaps more that could be lifted into the repo root within the container without maintaining it in the repo root on GitHub.

**In the long-term** (read: _just my wild idea for now_) more configurations and development conveniences could be added like that. For instance: something like [mise-en-place](https://mise.jdx.dev/) could improve the accessibility of certain manual development steps that are currently hidden behind scripts and/or documented processes. I'm thinking of things like "running bogo tests", "running TLS anvil", "rolling a release", "running clang-tidy", and such.

randombit:
I would say my minimal dependencies philosophy here also applies to the tooling required to do development [*]. But AIUI this proposal, this would just be a convenience and not required, yeah?

[*] This is not *just* an ascetic concern on my part; I pretty regularly do builds/tests/development on machines that I do not have root on, eg the GCC cfarm machines or machines otherwise made available to me for testing. Often the available toolset is basically just GCC/make/vim plus whatever I bother to compile and install into $HOME.

reneme:
> But AIUI this proposal, this would just be a convenience and not required, yeah?

Absolutely! This should never become mandatory at all.

reneme:
I just merged a first iteration of a dev container. Personally I'll start using this as my development environment, so some tweaks will certainly come up over time. I'm closing this nevertheless.

PR #5094, opened 2025-09-05
Set up a Dev Container

reneme:
See #5087 

This is fairly VSCode-focussed currently, and is meant to eventually replace the existing `src/editors/vscode` configuration. It should also work with just the `devcontainer` CLI for other editors (currently not tested). The underlying container should provide all the things one needs to get productive quickly.

I moved the `.clang-format` file from the repo root into `src/configs` and set up the devcontainer to find it there. Also the `run_clang_format.py` script defaults to that config location now. We'll probably not get around adding `.devcontainer` to the root, so that way the number of top-level entries stays the same at least. 😏

Assuming one uses VSCode those things should work out-of-the-box from within the devcontainer:

* Pre-configured editor settings (e.g. 3-space tab width in *.h and *.cpp)
* Auto-formatting via *clang-format* on save
* *clangd* for code completion and navigation
* compiling w/ gcc-13 and clang 18
* debugging of unittests in the VSCode GUI
* pylint and ruff warnings on save in Python files
* git blame ignores commits from `src/configs/git-blame-ignore-revs`
* default-configured OhMyZsh

Currently known limitations:

* OhMyZsh `fzf` key bindings don't work
* my GPG key on my YubiKey isn't passed through the Windows -> WSL2 -> DevContainer chain

Other things that could be added in the future:

* auto-formatting on save for Python (/cc @arckoor)
* type checking on save for Python (perhaps selectively for certain "production" scripts)
* auto-discovery and running of tests from within the editor (@FAlbertDev plz. help)
* some "task palette" to setup and run certain test suites (eg. CLI-tests, python binding tests, BoGo, ...)
* ... many more things


reneme:
I think this is a decent first step. I'm sure there will be further tweaks and fixes, but it would be nice to have this on master to be able to conveniently use it for development on other branches.

arckoor:
I think this is a good change, but I hope it will be possible to continue developing without a devcontainer. I was able to try it out, and it does work rather well (python formatting just worked for me too, I think vscode autodetects ruff being installed), however I am very much not fond of needing to give myself the `docker` group (because root and such), and getting vscode to correctly invoke the commands with sudo is cumbersome at best, so I hope all the config files, scripts and such will remain agnostic of the environment they're invoked in.

reneme:
Yes, absolutely! [Jack had the same request](https://github.com/randombit/botan/issues/5087#issuecomment-3249124942), and this is a totally reasonable requirement.

doc/dev_ref/contributing.rst
```diff
@@ -165,24 +165,16 @@ run the tests, test the fuzzers against a corpus, and produce an HTML report
 of total coverage. This coverage build requires the development headers for
 zlib, bzip2, liblzma, TrouSerS (libtspi), and Sqlite3.
 
-Editor Integrations
```
randombit, line 168: There is still the emacs config and `editorconfig` in `src/editors` so I think this section should not  be removed entirely. Instead maybe bump it below the dev container section

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+        wget \
+        zsh
+
+RUN echo "installing Botan-specific tools and dependencies" \
+    && DEBIAN_FRONTEND=noninteractive apt-get -y --no-install-recommends install \
+        clang \
+        clang-format-17 \
+        clangd-20 \
+        doxygen \
+        gdb \
```
randombit: It doesn't really matter one way or another but gdb and clang being in this section vs above seems odd

randombit (approved):

.devcontainer/Dockerfile
reneme: Indeed. No idea how that came to be. Fixed.

PR #5095, opened 2025-09-05
Add type hints to python.rst

arckoor:
Adds type hints to `python.rst`, and fixes some remaining errors in the type hinting.
Also documents _some_ things that weren't documented at all until now.
I also reordered `python.rst` and `botan3.py` so that they are ordered the same way. (I pinky promise the first commit only did reordering and some formatting fixes in the rst)

To be completely honest, the more I work with `python.rst`, the less convinced I am. It's a pain to keep it consistent with the actual code (it's still out of date), and more significantly it can't be used to generate inline documentation for IDEs. What about just ditching the handwritten version, moving all the documentation inside `botan3.py`, and autogenerating the rst file?


randombit:
> What about just ditching the handwritten version, moving all the documentation inside botan3.py, and autogenerating the rst file?

That could probably work tbh especially because

- There is effectively zero "implementation" in the Python wrapper, it's all public interfaces outside of the ffi decls and code for finding the library to dlopen
- Documentation tool of choice is Sphinx which is anyway a native Python doc generator (aiui)

Only real downside is that the doc would no longer be visible/readable to someone just perusing a git checkout, or an installed version that was installed without Sphinx (in that case we just install the rst files since they are still pretty readable and better than no docs at all)

arckoor:
Why would it not be visible to someone viewing a git checkout? Sure it would be interleaved with code, but I think in an IDE docstrings should be easily distinguishable from normal code due to coloring.
But anyway, if you're on board with this I'll give it a look sometime, but I would go ahead with this PR in the meantime anyway

randombit:
> Why would it not be visible to someone viewing a git checkout? Sure it would be interleaved with code, but I think in an IDE docstrings should be easily distinguishable from normal code due to coloring.

Visible in the sense of someone looking at `docs` in a git checkout (or more likely, via Github file browser) to get a general sense of what was available would not see anything about the Python binding, they instead have to find `src/python/botan3.py` and read that. Which, balanced against the docs being out of date or incomplete or harder to use (eg with IDE integrations) is *fine*.

arckoor:
@randombit I couldn't resist after all. ~~Note that with sphinx autogenerating it, it requires that it imports it, which in turn requires finding a `libbotan-3`. I never quite figured out what the recommended procedure is for this in development, so I've just been replacing the `dll = CDLL(dll_name)` line with `dll = CDLL(f"./{dll_name}")`, but I obviously can't push that. I suspect CI needs either installation for the docs job to work or some other hotfix I'm not aware of.~~

Turns out you can just overwrite `ctypes.CDLL()` with a dummy impl that does nothing, python just lets you do that :melting_face:

arckoor:
@randombit Any chance of getting this and #5096 merged before the next release?

randombit:
Just a general comment here that this PR has stalled because it's *very* hard to review, there are just too many unrelated changes/moves and it's hard to keep track of it all.

I'll see what I can do but tbh a much faster merge strategy would be

- One PR just adding the doc comments to `botan3.py` *WITHOUT* any reordering of definitions and without modifying the docs
- One PR reordering struct definitions, if this really is required (I don't see why that is though)
- Final PR removing the hand written docs and using the autodoc feature

arckoor:
@randombit The reorder was done because even on master `botan3.py` and `python.rst` have different orders, and it was a bit annoying while copy pasting. I also tried to order them in a way that "made sense", e.g.:
- putting `BlockCipher` near `SymmetricCipher`, instead of having `HashFunction` and `MsgAuthCode` between them
- `MPI` has things like `is_negative`, `is_positive`, ... which are all together, `is_prime` is buried somewhere in between the mod / mul functions
- `__del__` is usually right below `__init__`, but sometimes it's somewhere in the middle or at the end

Many more such examples, but it's mostly developer preference.
Ideally I'd still like `botan3.py` and `python.rst` to at least have the same class order, but future me problem.

Anyway, what I've done:
- only copied the doc to `botan3.py`
- addressed the typos that copilot found, but only in the py file
- turned `List[T]` annotation into `list[T]` annotations, because it's [deprecated](https://docs.python.org/3/library/typing.html#typing.List) (and we're already using `-> tuple` instead of `-> Tuple`)
- added a handful of missing type annotations

PR #5096, opened 2025-09-06
Format all Python code using ruff

randombit:
Currently this is compatible with both the default output of `ruff ` and also `black`.

CI checking is omitted here, will add in a later PR.

### DIFF
--- a/src/configs/git-blame-ignore-revs
+++ b/src/configs/git-blame-ignore-revs
@@ -19,3 +19,6 @@ fe7a2a1d850ae02934ea04d25fa214c825442535
 
 # [clang-format] AllowShortBlocksOnASingleLine
 de48a2eb6d2d77e104044da84c2fb7df5c08d65a
+
+# [ruff] initial application
+a05a80f3a993a726c53efe513c11480c594572be


src/python/botan3.py
```diff
@@ -174,7 +195,10 @@ def ffi_api(fn, args, allowed_errors=None):
     ffi_api(dll.botan_mac_final, [c_void_p, c_char_p])
     ffi_api(dll.botan_mac_clear, [c_void_p])
     ffi_api(dll.botan_mac_name, [c_void_p, c_char_p, POINTER(c_size_t)])
-    ffi_api(dll.botan_mac_get_keyspec, [c_void_p, POINTER(c_size_t), POINTER(c_size_t), POINTER(c_size_t)])
+    ffi_api(
+        dll.botan_mac_get_keyspec,
+        [c_void_p, POINTER(c_size_t), POINTER(c_size_t), POINTER(c_size_t)],
+    )
```
arckoor, line 201: It might make sense to encapsulate the entire `ffi_api` block with
```
# fmt: off
...
# fmt: on
```
but up to you

arckoor:
Also, before you merge this, could you take a look at #5095? This is going to cause quite the conflict :p

randombit:
@arckoor Thanks for the ping I didn't realize that PR was also modifying the Python lib but yeah let's get that landed first then the reformat.

PR #5100, opened 2025-09-12
Add CI builder for macOS 26 / XCode 26

randombit:


reneme:
Seems like there's some issue in the python test handling. Unfortunately, I didn't dare to update my MacBook to 26, yet. So I can't reproduce locally. Perhaps disable the culprit for macOS 26 until then?

randombit:
Continued in #5159

PR #5101, opened 2025-09-13
Refactor: Sponge<> base class for Ascon-p and Keccak-p

reneme:
The generic base class `Sponge<>` allows sharing some code between the Ascon-p and Keccak-p implementations. The full potential of this will be used in #5076 to implement additional sponge processing functions for data encryption and decryption.

Instead of implementing the actual "sponge processing" logic straight into the `Sponge<>` base class, I extracted it into an additional header that is included in *keccak_p.cpp* and *ascon_p.cpp*, only. That way, the definition of this template doesn't leak into users of the permutation classes.

### Pull Request Dependencies

* #5097


reneme:
Rebased to master after #5097 was merged.

src/lib/permutations/sponge/sponge.h
```diff
@@ -0,0 +1,75 @@
...
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#ifndef BOTAN_SPONGE_CONSTRUCTION_H_
+#define BOTAN_SPONGE_CONSTRUCTION_H_
+
+#include <array>
+#include <span>
+
+#include <botan/exceptn.h>
```
randombit, line 12: Wrong ordering of includes / no need for the empty line here unless intentionally creating two blocks that won't be sorted together.

src/lib/permutations/sponge/sponge_processing.h
```diff
@@ -0,0 +1,256 @@
...
+
+#ifndef BOTAN_SPONGE_PROCESSING_H_
+#define BOTAN_SPONGE_PROCESSING_H_
+
+#include <array>
+#include <span>
+
+#include <botan/exceptn.h>
+#include <botan/internal/loadstor.h>
+#include <botan/internal/stl_util.h>
```
randombit, line 14: Similarly I think the headers should all be one block and then sorted using the usual clang-format rule

randombit:
GCC 15 also has a slight SHAKE regression here but it's small barely .1 cpb

Also interestingly GCC 15 is much faster than Clang 21 for Keccak (eg SHAKE-128 output is 4.42 cpb with GCC vs 5.67 cpb with Clang)

randombit:
With #5133 the difference becomes even more pronounced (all numbers for SHAKE-128 output since that's the critical one for PQ)

* Current `master` 504 MB/s
* #5133 alone: 702 MB/s
* This PR alone: 467 MB/s
* This PR + 5133: 612 MB/s

Merging both SHAKE is still faster overall, but we are losing 15% somewhere which is unfortunate. It's definitely visible in the ML-KEM benchmarks.

Adding `BOTAN_FORCE_INLINE` to the main `process_bytes_in_sponge` and `inline` to all other functions in `sponge_processing.h` brings it from 467 to 480 MB which certainly seems worth it.

The main thing is that somehow even with `BOTAN_FORCE_INLINE`, `process_bytes_in_sponge` is not being fully inlined; it is still showing up as an independent function in profiles (using valgrind) taking up about 10% of the overall runtime. The profile with GCC doesn't show this at all; the permutation takes up 95% of the total runtime and the buffer operations are no more than 2%. I haven't looked at the generated asm yet.

src/lib/permutations/sponge/sponge.h
reneme: .... and also `<span>` isn't actually used anymore.

reneme:
> I am seeing a somewhat noticable .45 cpb regression for SHAKE-128 on a Tiger Lake laptop with Clang 21. Ascon and SHA-3 mostly unaffected. SHAKE regression would be somewhat unfortunate since the various PQ schemes rely on it.

Somewhat surprising to me that this really affects SHAKE only. Perhaps there's some difference on the usage level somehow? I'll see if I can look into it in the coming days, but I don't seem to be able to reproduce that on my Mac easily. Anyway, I agree, the regression is really unfortunate for sure. 

> GCC 15 also has a slight SHAKE regression here but it's small barely .1 cpb

In the measurements I did in September (the table at the top), there is a slight regression in throughput of much less than 1%. Frankly, I considered that just noise in the measurements. But it is consistent across SHAKE instances, so maybe not.

randombit:
For whatever reason it seems to affect output worse than input, so less noticable for the Keccak functions that have a fixed small output length. Comparing SHAKE-128 master (pre force-inlining the permutation) with this PR, using 16K buffers, input performance goes from 507 MB/s to 489 (~3% reduction), while output goes from 512 to 475 (~7%)

I would need to look over the change again but I wonder if there is some (in the squeeze case) no-op lambda that's not being inlined/removed by Clang.

randombit:
I think given this PR unblocks Ascon AEAD, the performance loss only seems to affect LLVM Clang, and that the force-inline of the permutation at least prevents it from being an outright regression, I think this is still ok to merge. Please do add the inlining guidance though that seems to help a bit.

reneme:
> Please do add the inlining guidance though that seems to help a bit.

Done in https://github.com/randombit/botan/pull/5101/commits/d3d5b730a68225f031a110531cde3d33132d28f7 to the best of my understanding. Please double-check.

> I think this is still ok to merge.

Let's leave a ticket for a follow-up. I'd still like to see if I can reproduce this.

src/lib/permutations/sponge/sponge_processing.h
```diff
@@ -0,0 +1,257 @@
...
+   using word_t = typename SpongeT::word_t;
+
+   BufferSlicer input_slicer(input);
+   process_bytes_in_sponge(sponge, input.size(), permutation_fn, [&](word_t state_word, auto bounds) {
+      return state_word ^ bounds.read_from(input_slicer);
+   });
+   BOTAN_ASSERT_NOMSG(input_slicer.empty());
+}
+
+void absorb_into_sponge(detail::SpongeLikeWithTrivialPermute auto& sponge, std::span<const uint8_t> input) {
```
randombit: Missed inline here

reneme:
I'll take care of it tonight.

PR #5102, opened 2025-09-13
Refactor: Keccak permutation has to be configured statically

reneme:
This applies the `consteval` constructor idiom [we talked about before](https://github.com/randombit/botan/pull/5061#discussion_r2288451042) to the `Keccak_Permutation` class. As a result, it is not possible to instantiate the Keccak-p at runtime anymore, but all configurations have to be statically declared and simply copied at runtime.

At this point, I'm not sure that this is really worth it (due to the fairly verbose and duplicated declarations in the Keccak-based algorithm implementations). I'd suggest to integrate the other refactorings first and come back to this in isolation. Hence the "draft" status.

# Pull Request Dependencies

* #5097 
* #5101


reneme:
Apparently clang had a oopsie on macos-13 building the amalgamation. This seems to be caused by the changes in this PR, given that the base PR (#5101) seems to pass this just fine. 😨 

```
  clang: error: unable to execute command: Segmentation fault: 11
  clang: error: clang frontend command failed due to signal (use -v to see invocation)
  Apple clang version 15.0.0 (clang-1500.1.0.2.5)
  Target: x86_64-apple-darwin22.6.0
  Thread model: posix
  InstalledDir: /Applications/Xcode_15.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
  clang: note: diagnostic msg: 
  ********************
  
  PLEASE ATTACH THE FOLLOWING FILES TO THE BUG REPORT:
  Preprocessed source(s) and associated run script(s) are located at:
  clang: note: diagnostic msg: /var/folders/vk/nx37ffx50hv5djclhltc26vw0000gn/T/botan_all-1371a4.cpp
  clang: note: diagnostic msg: /var/folders/vk/nx37ffx50hv5djclhltc26vw0000gn/T/botan_all-1371a4.sh
  clang: note: diagnostic msg: Crash backtrace is located in
  clang: note: diagnostic msg: /Users/runner/Library/Logs/DiagnosticReports/clang_<YYYY-MM-DD-HHMMSS>_<hostname>.crash
  clang: note: diagnostic msg: (choose the .crash file that corresponds to your crash)
  clang: note: diagnostic msg: 
  
  ********************
```

randombit:
@reneme Does this seem still worth pursuing to you? Seems kind of verbose to me but otherwise no strong opinions on it.

reneme:
Given that it guts the older clang, I'd say let's table that for now.

PR #5107, opened 2025-09-26
Add optional callback for user-defined TLS 1.2 key derivation function

lars-du:
# Use Case
We use Botan for TLS-PSK on embedded devices equipped with a hardware-protected environment (HPE), such as a Hardware Security Module (HSM) or Trusted Execution Environment (TEE). The pre-shared key (PSK) is securely provisioned to the devices and stored within the HPE. For security reasons, the PSK must not leave the HPE at runtime. This presents a challenge: Botan runs in the normal user environment (Rich Execution Environment), while the PSK is only accessible inside the HPE, meaning Botan cannot access the PSK directly.
To address this issue, this PR introduces an optional callback that allows the TLS key derivation function (KDF) to be delegated to the application. In our scenario, this callback is used to offload the computation of the TLS master secret to the HPE during the TLS 1.2 handshake. This enables Botan to use the master secret derived from the PSK without ever requiring direct access to the PSK itself.
 
# Usage example
## Callback implementation in user code
```c++
class tls_callbacks_example : public Botan::TLS::Callbacks
{
    // mandatory botan callbacks:
    // - void tls_emit_data(std::span<const uint8_t> data) final ...
    // - void tls_record_received(uint64_t seq_no, std::span<const uint8_t> data) final ...
    // - void tls_alert(Botan::TLS::Alert alert) final ...
 
    // new additional optional callback:
    std::unique_ptr<Botan::KDF>
    tls12_protocol_specific_kdf(std::string_view prf_algo) const final
    {
        if(prf_algo == "MD5" || prf_algo == "SHA-1") {     
            return std::make_unique<tls_12_prf_hpe>("SHA-256");
        }
 
        return std::make_unique<tls_12_prf_hpe>(prf_algo);
    }
};
```
 
## Implementation of user-defined TLS 1.2 pseudo random function
```c++
class tls_12_prf_hpe final : public Botan::KDF
{
public:
    explicit tls_12_prf_hpe(std::string_view hash_function)
      : m_hash_function{hash_function}
      , m_original_prf{Botan::KDF::create_or_throw(name())}
    {
    }
 
    std::string name() const final;
 
    std::unique_ptr<Botan::KDF> new_object() const final;
 
    void perform_kdf(std::span<uint8_t> key,
                     std::span<const uint8_t> secret,
                     std::span<const uint8_t> salt,
                     std::span<const uint8_t> label) const override final;
 
private:
    std::string const           m_hash_function;
    std::unique_ptr<Botan::KDF> m_original_prf;
 
    bool kdf_internal(std::span<std::uint8_t>       out_derived_key,
                      std::span<std::uint8_t const> secret,
                      std::span<std::uint8_t const> salt,
                      std::span<std::uint8_t const> label) const;
};
 
std::string
tls_12_prf_hpe::name() const
{
    return "TLS-12-PRF(" + m_hash_function + ")";
}
 
std::unique_ptr<Botan::KDF>
tls_12_prf_hpe::new_object() const
{
    return std::make_unique<tls_12_prf_hpe>(m_hash_function);
}
 
bool
tls_12_prf_hpe::kdf_internal(std::span<std::uint8_t>       out_derived_key,
                             std::span<std::uint8_t const> secret,
                             std::span<std::uint8_t const> salt,
                             std::span<std::uint8_t const> label) const
{
    // The TLS 1.2 PRF is called for multiple purposes (with different labels), but we are only interested in the case
    // where the label ends with "master secret" and the hash function is "SHA-256".
    // In this case, we calculate the master secret inside the HPE.
    //
    // The following labels are expected:
    // - "master secret"          -> calculate master secret in HPE
    // - "extended master secret" -> calculate master secret in HPE
    // - "key expansion"          -> derive using original KDF from botan
    // - "client finished"        -> derive using original KDF from botan
    // - "server finished"        -> derive using original KDF from botan
    // - ...                      -> derive using original KDF from botan
 
    const std::string label_str = {label.begin(), label.end()};
    if(not label_str.ends_with("master secret") || (m_hash_function != "SHA-256")) {
        m_original_prf->derive_key(out_derived_key, secret, salt, label);
        return true;
    }

    // ignore secret since it contains a dummy pre-master secret only
    auto const master_secret = hpe_derive_tls_master_secret(salt, label);  // CALL TO HPE
    if (master_secret.size() != out_derived_key.size()) {
        return false;
    }
 
    std::copy(master_secret.begin(), master_secret.end(), out_derived_key.begin());
    return true;
}
 
void tls_12_prf_hpe::perform_kdf(std::span<uint8_t> key,
                                 std::span<const uint8_t> secret,
                                 std::span<const uint8_t> salt,
                                 std::span<const uint8_t> label) const {
{
    if(not kdf_internal(key, secret, salt, label)) {
        // In case the key could not be derived, return a random key that will provoke a TLS alert due to an invalid
        // record MAC. This ensures that no information about invalid key identifiers is revealed to the remote peer.
        fill_with_random_bytes(key);
    }
}
```

### DIFF
--- a/src/lib/tls/tls12/tls_handshake_state.cpp
+++ b/src/lib/tls/tls12/tls_handshake_state.cpp
@@ -187,13 +187,7 @@ Session_Ticket Handshake_State::session_ticket() const {
 }
 
 std::unique_ptr<KDF> Handshake_State::protocol_specific_prf() const {
-   const std::string prf_algo = ciphersuite().prf_algo();
-
-   if(prf_algo == "MD5" || prf_algo == "SHA-1") {
-      return KDF::create_or_throw("TLS-12-PRF(SHA-256)");
-   }
-
-   return KDF::create_or_throw("TLS-12-PRF(" + prf_algo + ")");
+   return m_callbacks.tls12_protocol_specific_kdf(ciphersuite().prf_algo());
 }
 
 std::pair<std::string, Signature_Format> Handshake_State::choose_sig_format(const Private_Key& key,
--- a/src/lib/tls/tls_callbacks.cpp
+++ b/src/lib/tls/tls_callbacks.cpp
@@ -21,6 +21,7 @@
 #include <botan/tls_policy.h>
 #include <botan/tls_session.h>
 #include <botan/x509path.h>
+#include <botan/internal/fmt.h>
 #include <botan/internal/stl_util.h>
 
 #if defined(BOTAN_HAS_X25519)
@@ -434,6 +435,14 @@ void TLS::Callbacks::tls_ssl_key_log_data(std::string_view label,
                                           std::span<const uint8_t> client_random,
                                           std::span<const uint8_t> secret) const {
    BOTAN_UNUSED(label, client_random, secret);
+}
+
+std::unique_ptr<KDF> TLS::Callbacks::tls12_protocol_specific_kdf(std::string_view prf_algo) const {
+   if(prf_algo == "MD5" || prf_algo == "SHA-1") {
+      return KDF::create_or_throw("TLS-12-PRF(SHA-256)");
+   }
+
+   return KDF::create_or_throw(Botan::fmt("TLS-12-PRF({})", prf_algo));
 }
 
 }  // namespace Botan
--- a/src/lib/tls/tls_callbacks.h
+++ b/src/lib/tls/tls_callbacks.h
@@ -4,6 +4,7 @@
 *     2016 Jack Lloyd
 *     2017 Harry Reimann, Rohde & Schwarz Cybersecurity
 *     2022 René Meusel, Rohde & Schwarz Cybersecurity
+*     2025 Frederik Dornemann, CARIAD SE
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
@@ -12,6 +13,7 @@
 #define BOTAN_TLS_CALLBACKS_H_
 
 #include <botan/ec_point_format.h>
+#include <botan/kdf.h>
 #include <botan/pubkey.h>
 #include <botan/tls_alert.h>
 #include <botan/tls_algos.h>
@@ -705,6 +707,20 @@ class BOTAN_PUBLIC_API(2, 0) Callbacks /* NOLINT(*-special-member-functions) */ 
       virtual void tls_ssl_key_log_data(std::string_view label,
                                         std::span<const uint8_t> client_random,
                                         std::span<const uint8_t> secret) const;
+
+      /**
+       * Returns the key derivation function to be used for TLS 1.2
+       *
+       * The default implementation can be overridden to provide a user-defined
+       * key derivation function, for example to delegate key derivation to a
+       * hardware-protected environment when a pre-shared key must remain
+       * inaccessible to the non-secure world.
+       *
+       * @param prf_algo  name of the hash function (e.g. "SHA-256")
+       *
+       * @return  TLS 1.2 KDF implementation
+       */
+      virtual std::unique_ptr<KDF> tls12_protocol_specific_kdf(std::string_view prf_algo) const;
 };
 
 }  // namespace TLS
--- a/doc/authors.txt
+++ b/doc/authors.txt
@@ -37,6 +37,7 @@ FlexSecure GmbH
 Florent Le Coz
 Francis Dupont
 Frank Schoenmann
+Frederik Dornemann (CARIAD SE)
 Google Inc
 Gustavo Serra Scalet
 guywithcrookedface
--- a/news.rst
+++ b/news.rst
@@ -85,6 +85,8 @@ Version 3.11.0, Not Yet Released
 
 * Fix various clang-tidy and cppcheck warnings (GH #5172 #5207 #5204 #5205)
 
+* Add optional callback to provide a user-defined TLS 1.2 key derivation function (GH #5107)
+
 Version 3.10.0, 2025-11-06
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/src/tests/unit_tls.cpp
+++ b/src/tests/unit_tls.cpp
@@ -4,6 +4,7 @@
 *     2017 René Korthaus, Rohde & Schwarz Cybersecurity
 *     2017 Harry Reimann, Rohde & Schwarz Cybersecurity
 *     2023 René Meusel, Rohde & Schwarz Cybersecurity
+*     2025 Lars Dürkop, CARIAD SE
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
@@ -254,6 +255,7 @@ class TLS_Handshake_Test final {
                                                      const std::vector<uint8_t>&,
                                                      Botan::RandomNumberGenerator&,
                                                      const Botan::TLS::Policy&)>;
+      using custom_kdf_clbk = std::function<std::unique_ptr<Botan::KDF>(std::string_view)>;
 
    public:
       TLS_Handshake_Test(const std::string& test_descr,
@@ -317,6 +319,16 @@ class TLS_Handshake_Test final {
       void set_server_expected_handshake_alert(Botan::TLS::Alert alert) {
          BOTAN_ASSERT_NONNULL(m_server_cb);
          m_server_cb->set_expected_handshake_alert(alert);
+      }
+
+      void set_client_custom_kdf_callback(custom_kdf_clbk clbk) {
+         BOTAN_ASSERT_NONNULL(m_client_cb);
+         m_client_cb->set_custom_kdf_callback(std::move(clbk));
+      }
+
+      void set_server_custom_kdf_callback(custom_kdf_clbk clbk) {
+         BOTAN_ASSERT_NONNULL(m_server_cb);
+         m_server_cb->set_custom_kdf_callback(std::move(clbk));
       }
 
    private:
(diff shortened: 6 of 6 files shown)


src/lib/tls/tls_callbacks.h
```diff
@@ -675,6 +677,20 @@ class BOTAN_PUBLIC_API(2, 0) Callbacks /* NOLINT(*-special-member-functions) */
...
+       * Useful to implement a key derivation inside a protected hardware environment
+       * if a PSK is used that shall not be exposed to the user land.
+       *
+       * @param name           name of the key derivation function ("TLS-12-PRF" for TLS 1.2)
+       * @param hash_function  name of the hash function (e.g. "SHA-256")
+       *
+       * @return  the user-defined KDF or nullptr if the default shall be used
+       */
+      virtual std::unique_ptr<KDF> tls_protocol_specific_kdf(std::string_view name,
+                                                             std::string_view hash_function) const;
```
reneme: Could you please make it clearer that this is exclusively a TLS 1.2 callback? E.g. like it is done with `tls12_generate_ephemeral_ecdh_key()`. (Assuming that we don't find a way to generalize this to TLS 1.3)


```suggestion
      /**
       * Optional callback: Allows provision of a user-defined key derivation function for TLS 1.2
       *
       * Useful to implement a key derivation inside a protected hardware environment
       * if a PSK is used that shall not be exposed to the user land.
       *
       * @param name           name of the key derivation function ("TLS-12-PRF" for TLS 1.2)
       * @param hash_function  name of the hash function (e.g. "SHA-256")
       *
       * @return  the user-defined KDF or nullptr if the default shall be used
       */
      virtual std::unique_ptr<KDF> tls12_protocol_specific_kdf(std::string_view name,
                                                               std::string_view hash_function) const;
```

src/lib/tls/tls_callbacks.cpp
```diff
@@ -435,4 +435,11 @@ void TLS::Callbacks::tls_ssl_key_log_data(std::string_view label,
    BOTAN_UNUSED(label, client_random, secret);
 }
 
+std::unique_ptr<KDF> TLS::Callbacks::tls_protocol_specific_kdf(std::string_view name,
+                                                               std::string_view hash_function) const {
+   BOTAN_UNUSED(name);
+   BOTAN_UNUSED(hash_function);
+   return nullptr;
+}
```
reneme: Typically the implementation strategy for such optional callbacks is to place the "default behavior" into the default implementation of the callback. That way, applications can override this and flexibly up-call into the default implementation (or not).

I'm essentially suggesting to move most of the implementation in `Handshake_State::protocol_specific_prf()` in here. Perhaps except the getter call into the ciphersuite member.

src/lib/tls/tls_callbacks.h
lars-du: Changed in https://github.com/randombit/botan/pull/5107/commits/53fad43c79e60e3e83aecc467e97b3c38789e185

src/lib/tls/tls_callbacks.cpp
lars-du: Changed accordingly in https://github.com/randombit/botan/pull/5107/commits/53fad43c79e60e3e83aecc467e97b3c38789e185

frederik-do:
> I think it would be valuable to generalize this to TLS 1.3 though. And that shouldn't be too hard either, given that TLS 1.3's key schedule is also just based on KDFs. See the constructor of `Cipher_State` in `tls_cipher_state.cpp` (roughly line 442). We should be able to replace those hard-coded constructions of `HKDF_Extract` and `HKDF_Expand` with calls into your new callback.
> 
> The only caveat I see immediately: in your example you're using the "label" to decide when to delegate to your HPE to obtain the PSK. In TLS 1.3 the invocation to HKDF-Extract doesn't have use-case specific labels. So that might be a show stopper.

Yes, we also thought about extending this to TLS 1.3 and we did some analysis a while ago.
Unfortunately, it turned out not to be so easy as in TLS 1.2.

The issue with TLS 1.3 is, that it [derives a lot of keys](https://datatracker.ietf.org/doc/html/rfc8446#section-7.1) and some of them already quite early before any random/connection specific data is involved.

TLS 1.3 derives the `Early Secret` from the PSK via `HKDF-Extract(0*HASHLEN, PSK)`.
Means, the `Early Secret` remains the same for all TLS connections the use the same PSK.

If the `Early Secret` would be derived from the PSK inside a hardware protected environment (HPE) and then passed to the 
userland, an attacker who gains access to the `Early Secret` could use it to establish new TLS connections.
Therefore, we must keep the `Early Secret` in the HPE as well.

From the `Early Secret` TLS derives the `binder_key`:
`binder_key = HKDF-Expand(Early Secret, [Len]["tls13 ext binder"][HASH("")], HASH-LEN)`

The `binder_key` is also the same for all new connections but at least bound to the purpose label `"tls13 ext binder"`.
So we could keep the PSK and the `Early Secret` inside the HPE and pass all following keys to the userland.
That would include these keys:

```
client_early_traffic_secret  = HKDF-Expand(EarlySecret, [Len]["tls13 c e traffic"][HASH(ClientHello)], HASH-LEN)
early_exporter_master_secret = HKDF-Expand(EarlySecret, [Len]["tls13 e exp master"][HASH(ClientHello)], HASH-LEN)
Derived Early Secret         = HKDF-Expand(EarlySecret, [Len]["tls13 derived"][HASH("")], HASH-LEN)
```

The `Derived Early Secret` is also the same for all connections using the same PSK.
The `client_early_traffic_secret` and `early_exporter_master_secret` are connection specific since they are based on
the hash of the ClientHello which includes a random value.
But the last two keys are not relevant for all types of connections as far as I know.

So if an attacker gets the `binder_key` and the `Derived Early Secret` it might be enough to establish new connections if 
they do not employ early traffic or make use of the `early_exporter_master_secret`.

This might still be some gain in security compared to having the PSK in the userland, but not much.

Since the `binder_key` is used to calculate the HMAC for the psk identities inside the ClientHello we would also need to
delegate the corresponding HMAC operation to the HPE.

This would then maybe provide sufficient security as the `binder_key` is intended to bind the PSK to one particular handshake only.

So all in all, we think it would be possible to achieve something similar with TLS 1.3 but it is definitely way more complex.
We suggest to keep this out of this PR and maybe address this at a later point of time.

lars-du:
Hi @reneme @randombit, do you need more input from our side for this PR?

src/lib/tls/tls12/tls_handshake_state.cpp
```diff
@@ -254,11 +254,7 @@ Session_Ticket Handshake_State::session_ticket() const {
 std::unique_ptr<KDF> Handshake_State::protocol_specific_prf() const {
    const std::string prf_algo = ciphersuite().prf_algo();
 
-   if(prf_algo == "MD5" || prf_algo == "SHA-1") {
-      return KDF::create_or_throw("TLS-12-PRF(SHA-256)");
-   }
-
-   return KDF::create_or_throw("TLS-12-PRF(" + prf_algo + ")");
+   return m_callbacks.tls12_protocol_specific_kdf(prf_algo);
```
reneme: Nit:

```suggestion
   return m_callbacks.tls12_protocol_specific_kdf(ciphersuite().prf_algo());
```

reneme (approved):
Thanks for your patience on this. This looks good to me, but I'm leaving a second review and the merge to @randombit.

reneme:
> So all in all, we think it would be possible to achieve something similar with TLS 1.3 but it is definitely way more complex.
We suggest to keep this out of this PR and maybe address this at a later point of time.

I agree with your observations. Essentially, in Botan we would need to make some of the TLS 1.3 key schedule implementation (currently in [tls_cipher_state.h](https://github.com/randombit/botan/blob/master/src/lib/tls/tls13/tls_cipher_state.h)) customizable and optionally expose it to the application. Currently, objects of this class derive and hold all the relevant secrets of a TLS 1.3 connection and implement their usages.

For your use case of keeping the PSK in a protected environment, most of the key derivation would therefore need to be implemented inside the HPE (at least until the ServerHello is received and the ephemeral key exchange is done). Therefore, in Botan, we would need to at least make the state transition methods customizable by the application. See [tls_cipher_state.cpp](https://github.com/randombit/botan/blob/master/src/lib/tls/tls13/tls_cipher_state.cpp) for some more documentation.

Al least those methods would need to become customizable:

* For PSK-based setups and session resumptions
  * `init_with_psk` - to calculate the `binder_key` using the PSK
  * `advance_with_client_hello` - to allow sending early data (which is currently not supported by Botan, though)
  * `advance_with_server_hello` - incorporating the ephemeral shared secret after which all derived secrets are truly session specific

* For handshakes that don't use a PSK
  * `init_with_server_hello` - key derivation starts only once the ephemeral shared secret is available and all derived keys are always session specific

But I think, if we ever extend the TLS 1.3 implementation in that way, it would probably make sense to make the entire cipher state class customizable. That would give an application maximal flexibility when and even if key material leaves the HPE. In the extreme case no such key material would ever have to leave the HPE including for the bulk traffic encryption if this is feasible or desired.

src/lib/tls/tls12/tls_handshake_state.cpp
lars-du: Has been fixed in https://github.com/randombit/botan/compare/b0a2ccfdcf584c4e4ff808592c52f11825bfd836..19c70c57e3acdc6362933f96768c69338701665c

randombit:
@lars-du Sorry about the delay here, LGTM but needs a rebase

PR #5108, opened 2025-09-27
Allow disabling LASX/LSX on LoongArch (loongarch64)

MingcongBai:
Some embedded-oriented Loongson processors based on the LoongArch ISA, such as 2K0300 and 2K0500 do not have SIMD support (LSX or LASX), and embedded SoCs such as 2K1000LA/2K1500/2K2000/2K3000/3B6000M do not come with support for LASX.

Add two switches for disabling SIMD extensions. Even when runtime detection is implemented, this saves code sizes.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -441,7 +441,7 @@ def process_command_line(args):
 
     add_with_without_pair(target_group, 'compilation-database', True, 'disable compile_commands.json')
 
-    isa_extensions_that_can_be_disabled = [('NEON', 'arm32')]
+    isa_extensions_that_can_be_disabled = [('NEON', 'arm32'), ('LASX', 'loongarch64'), ('LSX', 'loongarch64')]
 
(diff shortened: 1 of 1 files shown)


randombit:
The main reason this flag exists for NEON is for distros (basically just Debian and derivatives) which for whatever reason require their ARMv7 builds to not include NEON at all, even as optional codepaths with feature detection. Does some similar situation apply here?

In general for embedded builds where code size is a concern the recommended approach is to use the option `--minimized-build` and enable only those modules specifically desirable to the use case.

MingcongBai:
> The main reason this flag exists for NEON is for distros (basically just Debian and derivatives) which for whatever reason require their ARMv7 builds to not include NEON at all, even as optional codepaths with feature detection. Does some similar situation apply here?

It is also the case here and the platforms diverts - LoongArch desktop/server targets assumes 128-bit vector support, whereas embedded targets have no vector support at all.

PR #5111, opened 2025-10-07
Streebog FIXME fix

huven:
This PR addresses one FIXME in the Streebog hash. While this is not a recommended hash, it is important for me for VeraCrypt compatibility.

This specific FIXME is important as it is inside the function consuming the most CPU time. The way this PR fixes it improves the performance with approx. 30%.

Really simple test by running the tests:

Before:
Streebog-256 ran 1571 tests in 1.73 msec all ok
Streebog-512 ran 1583 tests in 1.73 msec all ok

After:
Streebog-256 ran 1571 tests in 1.25 msec all ok
Streebog-512 ran 1583 tests in 1.24 msec all ok

Profiling the code revealed this in the first place (screenshot taken from my application, not the tests):
Screenshot 2025-10-07 at 15 12 04


src/lib/hash/streebog/streebog.cpp
```diff
@@ -106,9 +106,8 @@ inline uint64_t force_le(uint64_t x) {
 }
 
 inline void lps(uint64_t block[8]) {
-   uint8_t r[64];
-   // FIXME
-   std::memcpy(r, block, 64);
+   const uint64_t block2[8] = {block[0], block[1], block[2], block[3], block[4], block[5], block[6], block[7]};
+   std::span<const uint8_t> r{reinterpret_cast<const uint8_t*>(block2), 64};
```
reneme, line 110: If I'm understanding this algorithm correctly,

reneme:
For better measurements, you could use the `./botan speed` utility like listed below. Could you please re-measure with that?

```
./botan speed --msec=5000 Streebog-256 Streebog-512
```

huven:
Sure, here are the results.

Before PR:
Streebog-256 hash buffer size 1024 bytes: 141.119 MiB/sec (705.625 MiB in 5000.224 ms)
Streebog-512 hash buffer size 1024 bytes: 136.399 MiB/sec (682.000 MiB in 5000.051 ms)

After PR:
Streebog-256 hash buffer size 1024 bytes: 210.037 MiB/sec (1050.188 MiB in 5000.012 ms)
Streebog-512 hash buffer size 1024 bytes: 215.212 MiB/sec (1076.062 MiB in 5000.022 ms)

randombit:
I see a similar improvement on Linux with GCC. Kind of surprising since I'd effectively consider these two expressions equivalent - afaict there is no reason `memcpy` has to be slower here.

Change seems fine but please amend the commit message to more clearly state what was done eg

> Avoid a memory copy in a core loop of Streebog improving performance by 30% on x86

To be clear this doesn't really address the FIXME comment that was there - that was related to using `std::memcpy` instead of one of our wrappers. I would say instead it adds another FIXME since reinterpreting types like this is gross and occasionally UB, and at least I like to contain such things to easily grepable names like is done for `typecast_copy`.

huven:
Changed the commit message as suggested. Added arm64 too as my tests were done on macOS with M2.

PR #5115, opened 2025-10-09
Modernize logging implementation with C++20 and solve FIXME message

KaganCanSit:
Hello Botan Development Team,

This PR modernizes the log application by addressing all FIXME comments within `shim_log()`.

>       /*
>       FIXMEs:
>        - Rewrite this to use a std::ostream instead
>        - Allow using the env variable to point to where the log is written
>        - Avoid rechecking the env variable with each call (!)
>       */

Please leave a comment for any changes or edits. The local `botan-test` passed successfully, but I would like to perform comprehensive tests if directed.

Best regards.

_**NOTE**: `std::ofstream` and `std::format` may throw exceptions. Additionally, performing a one-time file path validation may be more prudent.  I'll consider it for now. I'm opening it as a draft to see the test results._

### DIFF
--- a/src/bogo_shim/bogo_shim.cpp
+++ b/src/bogo_shim/bogo_shim.cpp
@@ -30,7 +30,10 @@
 #include <botan/internal/stl_util.h>
 #include <botan/internal/target_info.h>
 
+#include <cstring>
 #include <ctime>
+#include <fstream>
+#include <iomanip>
 #include <iostream>
 #include <map>
 #include <memory>
@@ -57,29 +60,27 @@ int shim_output(const std::string& s, int rc = 0) {
    return rc;
 }
 
-void shim_log(const std::string& s) {
-   if(::getenv("BOTAN_BOGO_SHIM_LOG") != nullptr) {
-      /*
-      FIXMEs:
-       - Rewrite this to use a std::ostream instead
-       - Allow using the env variable to point to where the log is written
-       - Avoid rechecking the env variable with each call (!)
-      */
+void shim_log(std::string_view s) {
+   static const auto log_path = []() -> std::string {
+      const char* env = ::getenv("BOTAN_BOGO_SHIM_LOG");
+      if(env == nullptr) {
+         return {};
+      }
 
-      // NOLINTNEXTLINE(*-avoid-non-const-global-variables,*-owning-memory)
-      static FILE* g_log = std::fopen("/tmp/bogo_shim.log", "w");
+      auto log_file_path = std::string(env);
+      if(log_file_path.empty() || log_file_path == "1") {
+         return "/tmp/bogo_shim.log";
+      }
+      return env;
+   }();
 
-      if(g_log != nullptr) {
-         timeval tv{};
+   if(!log_path.empty()) {
+      static std::ofstream g_log(log_path, std::ios::out | std::ios::trunc);
+      if(g_log.is_open() && g_log.good()) {
+         const auto duration = std::chrono::system_clock::now().time_since_epoch();
+         const auto seconds = std::chrono::duration_cast<std::chrono::duration<double>>(duration);
 
-         ::gettimeofday(&tv, nullptr);
-         // NOLINTNEXTLINE(*-vararg)
-         static_cast<void>(std::fprintf(g_log,
-                                        "%lld.%lu: %s\n",
-                                        static_cast<unsigned long long>(tv.tv_sec),
-                                        static_cast<unsigned long>(tv.tv_usec),
-                                        s.c_str()));
-         static_cast<void>(std::fflush(g_log));
+         g_log << std::fixed << std::setprecision(6) << seconds.count() << ": " << s << std::endl;
       }
    }
 }


src/bogo_shim/bogo_shim.cpp
```diff
@@ -57,29 +58,18 @@ int shim_output(const std::string& s, int rc = 0) {
...
+   static const char* env_value = ::getenv("BOTAN_BOGO_SHIM_LOG");
+   if(env_value != nullptr) {
+      static std::ofstream g_log((*env_value != '\0') ? env_value : "/tmp/bogo_shim.log",
+                                 std::ios::out | std::ios::trunc);
+      if(g_log.is_open() && g_log.good()) {
+         const auto duration = std::chrono::system_clock::now().time_since_epoch();
+         const auto seconds = std::chrono::duration_cast<std::chrono::seconds>(duration);
+         const auto microseconds = std::chrono::duration_cast<std::chrono::microseconds>(duration - seconds);
+
+         g_log << std::format("{}.{:06d}: {}\n", seconds.count(), microseconds.count(), s);
```
reneme: According to [this table](https://en.cppreference.com/w/cpp/compiler_support.html#C.2B.2B20_library_features), `std::format()` (called "Text formatting" in the linked table) is fully supported on GCC 13 and Clang 17. Both of which are newer than [the minimal compiler versions we currently support](https://botan.randombit.net/handbook/support.html#supported-platforms). `bogo_shim.cpp` isn't compiled for all the platforms on GitHub Actions, and therefore the CI probably doesn't fail, but nevertheless we should avoid using `std::format` for now. Unfortunately.

src/bogo_shim/bogo_shim.cpp
```diff
@@ -57,29 +58,18 @@ int shim_output(const std::string& s, int rc = 0) {
...
-         ::gettimeofday(&tv, nullptr);
-         // NOLINTNEXTLINE(*-vararg)
-         static_cast<void>(std::fprintf(g_log,
-                                        "%lld.%lu: %s\n",
-                                        static_cast<unsigned long long>(tv.tv_sec),
-                                        static_cast<unsigned long>(tv.tv_usec),
-                                        s.c_str()));
-         static_cast<void>(std::fflush(g_log));
+void shim_log(std::string_view s) {
+   static const char* env_value = ::getenv("BOTAN_BOGO_SHIM_LOG");
```
reneme: Please use a `std::string` here and check for `env_value.empty()` in the `if` statements below. Who knows how long the pointer returned by `::getenv()` remains valid. Storing and reusing it as global state might backfire in unexpected ways.

src/bogo_shim/bogo_shim.cpp
```diff
@@ -57,29 +58,18 @@ int shim_output(const std::string& s, int rc = 0) {
...
+   if(env_value != nullptr) {
+      static std::ofstream g_log((*env_value != '\0') ? env_value : "/tmp/bogo_shim.log",
+                                 std::ios::out | std::ios::trunc);
+      if(g_log.is_open() && g_log.good()) {
+         const auto duration = std::chrono::system_clock::now().time_since_epoch();
+         const auto seconds = std::chrono::duration_cast<std::chrono::seconds>(duration);
+         const auto microseconds = std::chrono::duration_cast<std::chrono::microseconds>(duration - seconds);
```
reneme: Maybe use the fact that `std::chrono` can inherently deal with floats:


```C++
         const auto duration = std::chrono::system_clock::now().time_since_epoch();
         const auto secs = std::chrono::duration_cast<std::chrono::duration<double>>(duration);
         std::cout << std::format("{:.6f}: {}\n", secs.count(), s);
```

KaganCanSit:
Hello again @reneme,

I tried to implement the changes you suggested on this branch. Thanks again for your guidance. 

* The use of `format` has been removed.
* I applied the simplification you suggested for `chrono`.
* `const char *` usage change `std::string`

I have a few questions, I want to be sure;

I'm unsure about a few things regarding the `env` variable. I couldn't find complete information about the message left under FIXME: `Allow using the env variable to point to where the log is written`. I found a reference in the code tree under `doc/dev_ref`, but I couldn't find clear information about what values to provide during configuration. (Maybe I am missing.)

Based on this, I assumed that if the `env` variable is empty or has a value of `1`, logging is enabled by default; if there is no value, logging is disabled; but if the value is `1` and not empty, it is interpreted as a full path. (I also got some help from AI at this point.)

Could you provide a document or reference when you have time?

reneme:
> Based on this, I assumed that if the env variable is empty or has a value of 1, logging is enabled by default; if there is no value, logging is disabled; but if the value is 1 and not empty, it is interpreted as a full path. 

That's what I would intuitively assume. Note that this `bogo_shim.cpp` is just for testing. It is not part of the "production" library at all. Unlikely that there would be any more documentation on this environment variable. I think, what you implemented is just fine.

src/bogo_shim/bogo_shim.cpp
```diff
@@ -57,29 +60,27 @@ int shim_output(const std::string& s, int rc = 0) {
...
-         ::gettimeofday(&tv, nullptr);
-         // NOLINTNEXTLINE(*-vararg)
-         static_cast<void>(std::fprintf(g_log,
-                                        "%lld.%lu: %s\n",
-                                        static_cast<unsigned long long>(tv.tv_sec),
-                                        static_cast<unsigned long>(tv.tv_usec),
-                                        s.c_str()));
-         static_cast<void>(std::fflush(g_log));
+void shim_log(std::string_view s) {
+   static const std::string log_path = []() -> std::string {
```
reneme: Nit: if you specify the return value of the lambda, you might as well use `auto` for the variable definition.


```suggestion
   static const auto log_path = []() -> std::string {
```

src/bogo_shim/bogo_shim.cpp
```diff
@@ -57,29 +60,27 @@ int shim_output(const std::string& s, int rc = 0) {
...
+      const char* env = ::getenv("BOTAN_BOGO_SHIM_LOG");
+      if(env == nullptr) {
+         return {};
+      }
+
+      if(*env == '\0' || std::strcmp(env, "1") == 0) {
+         return "/tmp/bogo_shim.log";
+      }
+      return env;
```
reneme: I think we should avoid manually handling of C-style strings as much as we can. Maybe like that:


```suggestion
      auto logfile_path = std::string(env);
      if(logfile_path.empty() || logfile_path == "1") {
         logfile_path = "/tmp/bogo_shim.log";
      }
      return logfile_path;
```

src/bogo_shim/bogo_shim.cpp
```diff
@@ -57,29 +60,27 @@ int shim_output(const std::string& s, int rc = 0) {
...
+   }();
+
+   if(!log_path.empty()) {
+      static std::ofstream g_log(log_path, std::ios::out | std::ios::trunc);
+      if(g_log.is_open() && g_log.good()) {
+         const auto duration = std::chrono::system_clock::now().time_since_epoch();
+         const auto seconds = std::chrono::duration_cast<std::chrono::duration<double>>(duration);
+
+         g_log << std::fixed << std::setprecision(6) << seconds.count() << ": " << s << '\n';
+         g_log.flush();
```
reneme:
```suggestion
         g_log << std::fixed << std::setprecision(6) << seconds.count() << ": " << s << std::endl;
```

... `std::endl` implicitly flushes the stream. Usually this is considered a disadvantage, but here we could make use of it.

KaganCanSit: In general use, a more appropriate approach is to use 'std::endl' or 'flush' when data needs to be intentionally flushed when various operations are performed in the buffer or when content accumulates. However, as you said, the buffer is already emptied during the write process. So `std::endl` is a more readable solution.

src/bogo_shim/bogo_shim.cpp
KaganCanSit: I wanted to avoid using an additional variable here. However, considering that more reliable structures are used and this part will only run once, an additional variable can be defined.

src/bogo_shim/bogo_shim.cpp
KaganCanSit: I generally prefer open type writing, out of habit. I edited it as per your request.

KaganCanSit:
> That's what I would intuitively assume. Note that this `bogo_shim.cpp` is just for testing. It is not part of the "production" library at all. Unlikely that there would be any more documentation on this environment variable. I think, what you implemented is just fine.

If you agree, this PR can be included in the project in this way. Thank you for your feedback. If the tests are successful, I will finalize the draft.

src/bogo_shim/bogo_shim.cpp
reneme: I agree, the extra variable adds complexity. But the snake pit that are C strings is certainly worse. Also consider that compilers will very likely see through this and avoid unnecessary object creations and/or copies.

PR #5073, opened 2025-08-23
BLAKE2s modernization: Resolve TODO and Clang-Tidy warning suppression

KaganCanSit:
Hello Botan Developer Team,

This PR addresses TODO comments and suppressed Clang-Tidy warnings in `blake2s.cpp` while maintaining minimal changes to preserve existing functionality and performance characteristics. I avoided making major changes overall. The list of items I addressed is as follows:
- Update C-style arrays with `std::array`
- Edits for Clang Tidy warnings
     *  `modernize-loop-convert`: Replaced index-based loops with range-based for loops where appropriate
     * `*-special-member-functions`: Added explicit deletion declarations for copy/move operations (I'm still not entirely sure about this!)
     * NOLINTNEXTLINE suppressions removed

When I compiled and tested using the Clang compiler (`Debian clang version 19.1.7`) with the following command, I did not encounter any errors.

```bash
ninja clean && ./configure.py --without-documentation --with-boost --cc=clang --compiler-cache=ccache --build-targets=static,cli,tests --build-tool=ninja && ninja && ./botan-test --test-threads=4 --run-long-tests
```

**Notes**
- Regarding `secure_vector` TODO: The remaining TODO in `blake2s.h` regarding `secure_vector` usage was intentionally left unchanged. Based on @reneme's previous feedback about the need for careful evaluation of `secure_vector` usage, I prefer to address this in a separate discussion/PR to ensure proper implementation.

- Future improvements: Additional modernizations (span usage, [[nodiscard]], etc.) are possible but were excluded to keep this review focused. I'm happy to address these in follow-up PRs if deemed appropriate.

This PR is marked as draft to gather feedback on the modernization approach before final review.
Best regards.

PR #5117, opened 2025-10-10
Modernize BLAKE2s and improve performance by removing per-byte loops

KaganCanSit:
Hello Botan Developer Team,

This PR combines the changes from:
- #5062 - BLAKE2s modernization: Resolve TODO and Clang-Tidy warning suppression (#5073 — BLAKE2s modernization (resolving TODOs and Clang-Tidy suppressions))
- #5116 — Performance optimization by eliminating per-byte loops

### Summary of Changes
- Improved performance by eliminating byte-by-byte loops and using block-sized operations (@huven)
- Addressed remaining TODOs and Clang-Tidy warnings in blake2s.cpp/h
- Replaced C-style arrays with std::array
- Added explicit deletion of copy/move operations where appropriate
- Removed redundant NOLINTNEXTLINE suppressions

### Notes
- This PR is **not intended for immediate merge**; it serves testing and review purposes only.
- Additional modernizations (e.g., std::span, [[nodiscard]]) are intentionally deferred.

Optimization for performance was performed by;
Co-authored-by: Richard Huveneers <https://github.com/huven>

Use 64-bit counter and AlignmentBuffer in BLAKE2s. Combined two refactors improving alignment and counter consistenct;
Co-authored-by: René Meusel <github@renemeusel.de>

### DIFF
--- a/src/lib/hash/blake2s/blake2s.cpp
+++ b/src/lib/hash/blake2s/blake2s.cpp
@@ -1,6 +1,8 @@
 /*
  * BLAKE2s
- * (C) 2023           Richard Huveneers
+ * (C) 2023, 2025       Richard Huveneers
+ * (C) 2025             Kagan Can Sit
+ * (C) 2025             René Meusel, Rohde & Schwarz Cybersecurity
  *
  * Based on the RFC7693 reference implementation
  *
@@ -13,6 +15,7 @@
 #include <botan/internal/fmt.h>
 #include <botan/internal/loadstor.h>
 #include <botan/internal/rotate.h>
+#include <botan/internal/stl_util.h>
 
 namespace Botan {
 
@@ -20,12 +23,14 @@ namespace {
 
 // Initialization Vector.
 
-const uint32_t blake2s_iv[8] = {
+constexpr std::array<uint32_t, 8> blake2s_iv{
    0x6A09E667, 0xBB67AE85, 0x3C6EF372, 0xA54FF53A, 0x510E527F, 0x9B05688C, 0x1F83D9AB, 0x5BE0CD19};
 
 // Mixing function G.
 
-inline void B2S_G(uint8_t a, uint8_t b, uint8_t c, uint8_t d, uint32_t x, uint32_t y, uint32_t* v) {
+template <uint8_t a, uint8_t b, uint8_t c, uint8_t d>
+   requires(a < 16 && b < 16 && c < 16 && d < 16)
+constexpr void B2S_G(uint32_t x, uint32_t y, std::span<uint32_t, 16> v) {
    v[a] = v[a] + v[b] + x;
    v[d] = rotr<16>(v[d] ^ v[a]);
    v[c] = v[c] + v[d];
@@ -42,67 +47,52 @@ std::string BLAKE2s::name() const {
    return fmt("BLAKE2s({})", m_outlen << 3);
 }
 
-//      Secret key (also <= 32 bytes) is optional (keylen = 0).
-//      (keylen=0: no key)
+// BLAKE2s is specified as a message authentication code. For that, the
+// key would need to be zero-padded and incorporated into the initial hash
+// state. See RFC 7693 Section 3.3 and Appendix D.2 `blake2s_init()`.
+void BLAKE2s::state_init(size_t outlen) {
+   m_h = blake2s_iv;  // state, "param block"
+   m_h[0] ^= 0x01010000 ^ outlen;
 
-void BLAKE2s::state_init(size_t outlen, const uint8_t* key, size_t keylen) {
-   for(size_t i = 0; i < 8; i++) {  // state, "param block"
-      m_h[i] = blake2s_iv[i];
-   }
-   m_h[0] ^= 0x01010000 ^ (keylen << 8) ^ outlen;
-
-   m_t[0] = 0;  // input count low word
-   m_t[1] = 0;  // input count high word
-   m_c = 0;     // pointer within buffer
+   m_bytes_processed = 0;
    m_outlen = outlen;
-
-   for(size_t i = keylen; i < 64; i++) {  // zero input block
-      m_b[i] = 0;
-   }
-   if(keylen > 0) {
-      add_data(std::span<const uint8_t>(key, keylen));
-      m_c = 64;  // at the end
-   }
+   m_buffer.clear();
 }
 
 // Compression function. "last" flag indicates last block.
+void BLAKE2s::compress(bool last, std::span<const uint8_t> buf) {
+   BOTAN_ASSERT_NOMSG(buf.size() == block_size);
+   constexpr std::array<std::array<uint8_t, 16>, 10> sigma{{{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15},
+                                                            {14, 10, 4, 8, 9, 15, 13, 6, 1, 12, 0, 2, 11, 7, 5, 3},
+                                                            {11, 8, 12, 0, 5, 2, 15, 13, 10, 14, 3, 6, 7, 1, 9, 4},
+                                                            {7, 9, 3, 1, 13, 12, 11, 14, 2, 6, 5, 10, 4, 0, 15, 8},
+                                                            {9, 0, 5, 7, 2, 4, 10, 15, 14, 1, 11, 12, 6, 8, 3, 13},
+                                                            {2, 12, 6, 10, 0, 11, 8, 3, 4, 13, 7, 5, 15, 14, 1, 9},
+                                                            {12, 5, 1, 15, 14, 13, 4, 10, 0, 7, 6, 3, 9, 2, 8, 11},
+                                                            {13, 11, 7, 14, 12, 1, 3, 9, 5, 0, 15, 4, 8, 6, 2, 10},
+                                                            {6, 15, 14, 9, 11, 3, 0, 8, 12, 2, 13, 7, 1, 4, 10, 5},
+                                                            {10, 2, 8, 4, 7, 6, 1, 5, 15, 11, 9, 14, 3, 12, 13, 0}}};
 
-void BLAKE2s::compress(bool last) {
-   const uint8_t sigma[10][16] = {{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15},
-                                  {14, 10, 4, 8, 9, 15, 13, 6, 1, 12, 0, 2, 11, 7, 5, 3},
-                                  {11, 8, 12, 0, 5, 2, 15, 13, 10, 14, 3, 6, 7, 1, 9, 4},
-                                  {7, 9, 3, 1, 13, 12, 11, 14, 2, 6, 5, 10, 4, 0, 15, 8},
-                                  {9, 0, 5, 7, 2, 4, 10, 15, 14, 1, 11, 12, 6, 8, 3, 13},
-                                  {2, 12, 6, 10, 0, 11, 8, 3, 4, 13, 7, 5, 15, 14, 1, 9},
-                                  {12, 5, 1, 15, 14, 13, 4, 10, 0, 7, 6, 3, 9, 2, 8, 11},
-                                  {13, 11, 7, 14, 12, 1, 3, 9, 5, 0, 15, 4, 8, 6, 2, 10},
-                                  {6, 15, 14, 9, 11, 3, 0, 8, 12, 2, 13, 7, 1, 4, 10, 5},
-                                  {10, 2, 8, 4, 7, 6, 1, 5, 15, 11, 9, 14, 3, 12, 13, 0}};
-   uint32_t v[16];
-   uint32_t m[16];
+   // init work variables
+   std::array<uint32_t, 16> v = concat(m_h, blake2s_iv);
 
-   for(size_t i = 0; i < 8; i++) {  // init work variables
-      v[i] = m_h[i];
-      v[i + 8] = blake2s_iv[i];
-   }
-
-   v[12] ^= m_t[0];  // low 32 bits of offset
-   v[13] ^= m_t[1];  // high 32 bits
-   if(last) {        // last block flag set ?
+   v[12] ^= static_cast<uint32_t>(m_bytes_processed);
+   v[13] ^= static_cast<uint32_t>(m_bytes_processed >> 32);
+   if(last) {  // last block flag set ?
       v[14] = ~v[14];
    }
-   load_le<uint32_t>(m, m_b, 16);  // get little-endian words
 
-   // NOLINTNEXTLINE(modernize-loop-convert) TODO clean this up
-   for(size_t i = 0; i < 10; i++) {  // ten rounds
-      B2S_G(0, 4, 8, 12, m[sigma[i][0]], m[sigma[i][1]], v);
-      B2S_G(1, 5, 9, 13, m[sigma[i][2]], m[sigma[i][3]], v);
-      B2S_G(2, 6, 10, 14, m[sigma[i][4]], m[sigma[i][5]], v);
-      B2S_G(3, 7, 11, 15, m[sigma[i][6]], m[sigma[i][7]], v);
-      B2S_G(0, 5, 10, 15, m[sigma[i][8]], m[sigma[i][9]], v);
-      B2S_G(1, 6, 11, 12, m[sigma[i][10]], m[sigma[i][11]], v);
-      B2S_G(2, 7, 8, 13, m[sigma[i][12]], m[sigma[i][13]], v);
-      B2S_G(3, 4, 9, 14, m[sigma[i][14]], m[sigma[i][15]], v);
+   const auto m = load_le<std::array<uint32_t, 16>>(buf);  // get little-endian words
+
+   for(const auto& perm : sigma) {  // ten rounds
+      B2S_G<0, 4, 8, 12>(m[perm[0]], m[perm[1]], v);
+      B2S_G<1, 5, 9, 13>(m[perm[2]], m[perm[3]], v);
+      B2S_G<2, 6, 10, 14>(m[perm[4]], m[perm[5]], v);
+      B2S_G<3, 7, 11, 15>(m[perm[6]], m[perm[7]], v);
+      B2S_G<0, 5, 10, 15>(m[perm[8]], m[perm[9]], v);
+      B2S_G<1, 6, 11, 12>(m[perm[10]], m[perm[11]], v);
+      B2S_G<2, 7, 8, 13>(m[perm[12]], m[perm[13]], v);
+      B2S_G<3, 4, 9, 14>(m[perm[14]], m[perm[15]], v);
    }
 
    for(size_t i = 0; i < 8; ++i) {
@@ -114,34 +104,32 @@ void BLAKE2s::compress(bool last) {
  * Clear memory of sensitive data
  */
 void BLAKE2s::clear() {
-   state_init(m_outlen, nullptr, 0);
+   state_init(m_outlen);
 }
 
-void BLAKE2s::add_data(std::span<const uint8_t> in) {
-   // NOLINTNEXTLINE(modernize-loop-convert) TODO optimize this
-   for(size_t i = 0; i < in.size(); i++) {
-      if(m_c == 64) {        // buffer full ?
-         m_t[0] += m_c;      // add counters
-         if(m_t[0] < m_c) {  // carry overflow ?
-            m_t[1]++;        // high word
-         }
-         compress(false);  // compress (not last)
-         m_c = 0;          // counter to zero
+void BLAKE2s::add_data(std::span<const uint8_t> input) {
+   BufferSlicer in(input);
+
+   while(!in.empty()) {
+      if(const auto one_block = m_buffer.handle_unaligned_data(in)) {
+         m_bytes_processed += block_size;
+         compress(false, *one_block);
       }
-      m_b[m_c++] = in[i];
+
+      if(m_buffer.in_alignment()) {
+         while(const auto aligned_block = m_buffer.next_aligned_block_to_process(in)) {
+            m_bytes_processed += block_size;
+            compress(false, *aligned_block);
+         }
+      }
    }
 }
 
 void BLAKE2s::final_result(std::span<uint8_t> out) {
-   m_t[0] += m_c;      // mark last block offset
-   if(m_t[0] < m_c) {  // carry overflow
-      m_t[1]++;        // high word
-   }
+   m_bytes_processed += m_buffer.elements_in_buffer();
 
-   while(m_c < 64) {  // fill up with zeros
-      m_b[m_c++] = 0;
-   }
-   compress(true);  // final block flag = 1
+   m_buffer.fill_up_with_zeros();
+   compress(true, m_buffer.consume());
 
    // little endian convert and store
    copy_out_le(out.first(output_length()), m_h);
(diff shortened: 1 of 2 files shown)


src/lib/hash/blake2s/blake2s.h
```diff
@@ -35,15 +42,14 @@ class BLAKE2s final : public HashFunction /* NOLINT(*-special-member-functions)
...
-
-      // TODO use secure_vector here
-      uint8_t m_b[64]{};    // input buffer
-      uint32_t m_h[8]{};    // chained state
-      uint32_t m_t[2]{};    // total number of bytes
-      uint8_t m_c = 0;      // pointer for b[]
-      size_t m_outlen = 0;  // digest size
+      void state_init(std::size_t outlen, const uint8_t* key, std::size_t keylen);
+      void compress(bool last, std::span<const uint8_t> buf);
+
```
huven, line 51: I would keep the "// TODO use secure_vector here" comment here.

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -118,18 +112,35 @@ void BLAKE2s::clear() {
 }
 
 void BLAKE2s::add_data(std::span<const uint8_t> in) {
-   // NOLINTNEXTLINE(modernize-loop-convert) TODO optimize this
-   for(size_t i = 0; i < in.size(); i++) {
-      if(m_c == 64) {        // buffer full ?
-         m_t[0] += m_c;      // add counters
-         if(m_t[0] < m_c) {  // carry overflow ?
-            m_t[1]++;        // high word
+   constexpr std::size_t block_size = 64;
```
huven: Nice to add the block_size constant, perhaps use "block_size = hash_block_size()" here instead of redefining the value?

src/lib/hash/blake2s/blake2s.h
KaganCanSit: In my first PR request, @reneme and I discussed that a array could be used here and that the destructor had already cleaned it.  (https://github.com/randombit/botan/pull/5062#discussion_r2284279129)

Note: I accidentally closed this PR. I later made the necessary edits and reopened. (https://github.com/randombit/botan/pull/5073#issuecomment-3217269399)

src/lib/hash/blake2s/blake2s.cpp
KaganCanSit: I missed that. I'm fixing it now.

Update: 
@huven `hash_block_size()` is a virtual function called at runtime, so it cannot be used in a constexpr context. Using the constant `64` here is consistent with the value returned by `hash_block_size()` and allows for compile-time optimization. This function may be called externally. I need to review it.

However, perhaps we could make the following definition header private and have the function return it. I'm not sure. Something like this:

```cpp
public:
std::size_t hash_block_size() const override { return block_size; }

private:
constexpr std::size_t block_size = 64;
```

Also, could this constant be used in a few more places?
```cpp
m_c = 64;  // at the end
...
BOTAN_ASSERT_NOMSG(buf.size() == 64);
```
Update:
```cpp
m_c = block_size;  // at the end
...
BOTAN_ASSERT_NOMSG(buf.size() == block_size);
```

huven: I think that's a great idea.

huven:
Thanks, looks perfect to me.

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -13,19 +14,20 @@
...
 // Initialization Vector.
 
-const uint32_t blake2s_iv[8] = {
+constexpr std::array<uint32_t, 8> blake2s_iv{
    0x6A09E667, 0xBB67AE85, 0x3C6EF372, 0xA54FF53A, 0x510E527F, 0x9B05688C, 0x1F83D9AB, 0x5BE0CD19};
 
 // Mixing function G.
 
-inline void B2S_G(uint8_t a, uint8_t b, uint8_t c, uint8_t d, uint32_t x, uint32_t y, uint32_t* v) {
+inline void B2S_G(uint8_t a, uint8_t b, uint8_t c, uint8_t d, uint32_t x, uint32_t y, std::span<uint32_t, 16> v) {
```
reneme: I think this warrants some assertions about the parameters and the function could be marked `constexpr` (which implies `inline`).


```suggestion
constexpr void B2S_G(uint8_t a, uint8_t b, uint8_t c, uint8_t d, uint32_t x, uint32_t y, std::span<uint32_t, 16> v) {
   BOTAN_DEBUG_ASSERT(a <= v.size());
   BOTAN_DEBUG_ASSERT(b <= v.size());
   BOTAN_DEBUG_ASSERT(c <= v.size());
   BOTAN_DEBUG_ASSERT(d <= v.size());
```

src/lib/hash/blake2s/blake2s.h
```diff
@@ -35,15 +43,16 @@ class BLAKE2s final : public HashFunction /* NOLINT(*-special-member-functions)
...
+      std::array<uint8_t, 64> m_b{};  // input buffer
+      std::array<uint32_t, 8> m_h{};  // chained state
+      std::array<uint32_t, 2> m_t{};  // total number of bytes
+      uint8_t m_c = 0;                // pointer for b[]
+      std::size_t m_outlen = 0;       // digest size
+
+      static inline constexpr std::size_t block_size = 64;
```
reneme: Style nit: please move this constant definition into a separate `private:` section at the top of the class declaration. Like:

```C++
class Blake2s : ... {
   private:
      static constexpr std::size_t block_size = 64;
```

(I believe the `inline` isn't really needed for constexprs.)

src/lib/hash/blake2s/blake2s.h
```diff
@@ -35,15 +43,16 @@ class BLAKE2s final : public HashFunction /* NOLINT(*-special-member-functions)
...
-      uint8_t m_c = 0;      // pointer for b[]
-      size_t m_outlen = 0;  // digest size
+      void state_init(std::size_t outlen, const uint8_t* key, std::size_t keylen);
+      void compress(bool last, std::span<const uint8_t> buf);
+
+      std::array<uint8_t, 64> m_b{};  // input buffer
+      std::array<uint32_t, 8> m_h{};  // chained state
+      std::array<uint32_t, 2> m_t{};  // total number of bytes
+      uint8_t m_c = 0;                // pointer for b[]
+      std::size_t m_outlen = 0;       // digest size
```
reneme: Hot take: these variables are all initialized in `state_init` which is called in the class' constructor. So technically we could skip initialization here and save a few cycles.

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -45,67 +47,59 @@ std::string BLAKE2s::name() const {
...
+                                                            {12, 5, 1, 15, 14, 13, 4, 10, 0, 7, 6, 3, 9, 2, 8, 11},
+                                                            {13, 11, 7, 14, 12, 1, 3, 9, 5, 0, 15, 4, 8, 6, 2, 10},
+                                                            {6, 15, 14, 9, 11, 3, 0, 8, 12, 2, 13, 7, 1, 4, 10, 5},
+                                                            {10, 2, 8, 4, 7, 6, 1, 5, 15, 11, 9, 14, 3, 12, 13, 0}}};
+
+   // init work variables
+   std::array<uint32_t, 16> v = concat(m_h, blake2s_iv);
 
    v[12] ^= m_t[0];  // low 32 bits of offset
    v[13] ^= m_t[1];  // high 32 bits
```
reneme: Instead of manually handling this 64-bit "processed bytes" counter as an array of two 32-bit values and handling the overflow manually, let's just use an ordinary `uint64_t` and split it up into lo and hi bits here:


```suggestion
v[12] ^= static_cast<uint32_t>(m_bytes_processed)
v[13] ^= static_cast<uint32_t>(m_bytes_processed >> 32)
```

... as a side effect, the omitted branches in the hot path (for the overflow handling) will probably provide a bit of a performance improvement.

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -118,18 +112,34 @@ void BLAKE2s::clear() {
...
+         compress(false, in.first(block_size));
+         in = in.subspan(block_size);
       }
-      m_b[m_c++] = in[i];
    }
+
+   BOTAN_ASSERT_NOMSG(in.size() <= (block_size - m_c));
+   std::copy(in.begin(), in.end(), m_b.begin() + m_c);
+   m_c += static_cast<uint8_t>(in.size());
 }
```
reneme: This implementation could greatly profit from the `AlignmentBuffer<>` helper, like so:

```C++
void BLAKE2s::add_data(std::span<const uint8_t> input) {
   BufferSlicer in(input);

   while(!in.empty()) {
      if(const auto one_block = m_buffer.handle_unaligned_data(in)) {
         m_bytes_processed += block_size;
         compress(false, *one_block);
      }

      if(m_buffer.in_alignment()) {
         while(const auto aligned_block = m_buffer.next_aligned_block_to_process(in)) {
            m_bytes_processed += block_size;
            compress(false, *aligned_block);
         }
      }
   }
}
```

Note that this assumes the `m_bytes_processed` refactoring suggested elsewhere. Also, the finalization and clearing needs to be adapted accordingly.

huven: Nice helper. Will need careful handling of edge cases. Like the current code tests deliberately on '> block_size' so if the tail data is exactly one block it will not yet be compressed (as it might be the final block). Similarly for the unaligned data.

KaganCanSit:
These are truly valuable changes that will help me learn new things. Thank you @reneme. I will try to look into it today between 8:00 p.m. and 11:00 p.m. (UTC +3).

After making all the changes, I will send you a notification with a comment @huven.

src/lib/hash/blake2s/blake2s.cpp
reneme: Thanks for confirming that. Given that that's the case for other hash functions as well, the helper can (optionally) account for that by setting an extra template parameter when configuring it (in the header, please see the linked adaptations).

src/lib/hash/blake2s/blake2s.h
huven: The initializers were added recently by commit e1c4b3b (Enable and fix clang-tidy warning hicpp-member-init)

src/lib/hash/blake2s/blake2s.cpp
KaganCanSit: Constexpr makes sense here, but shouldn't the checks be `<` instead of `<=`? The content of `v.size()` will return the value 16. However, index 16 is not reachable. (Out of bounds)

src/lib/hash/blake2s/blake2s.h
KaganCanSit: In C++11/14, non-inline cases could result in ODR (One Definition Rule) violations. With C++17 and later, the game changed with "A static constexpr data member is impicitly inline." This is a bit of a habit. I agree that it's unnecessary.

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -45,67 +53,58 @@ std::string BLAKE2s::name() const {
...
+   m_buffer.clear();
 
-   for(size_t i = keylen; i < 64; i++) {  // zero input block
-      m_b[i] = 0;
-   }
    if(keylen > 0) {
       add_data(std::span<const uint8_t>(key, keylen));
-      m_c = 64;  // at the end
+      m_bytes_processed = block_size;  // at the end
    }
```
huven: It looks like an error was introduced here. Note: currently keylen is always 0, so the tests won't pick this up. The keylen provides support for blake2s MAC, I left that code in from the RFC to allow possible extension in the future.

I think m_bytes_processed should not be updated here, as that should only be done on compress.
What should be done is filling the buffer with zero's after adding the key, then it will behave like the previous code (setting m_c to 64 just added zeros).

huven: Oh... and I **love** the other changes you made!

KaganCanSit: I'll take a quick look at this. But I'm quite tired. @reneme make major changes. I didn't. I'm basically helping you and @reneme. 😄 

Not sure, but:
```cpp
m_buffer.fill_up_with_zeros();
```

And a test to check for this error condition would be nice.

huven: Yup, that's it 👍

huven:
To test the 64bit counter, we need to feed more than 4GB (2^32) to the hash. I checked it manually.

This build:

```
% dd if=/dev/zero bs=1048576 count=4097 | ./botan hash --algo='BLAKE2s(256)' --format=hex  --buf-size=1024     
08A966F3ADF402721335271ED9B42857B60AE0137EBD19DA05255607B0EE9E69 -
```

and another tool

```
% dd if=/dev/zero bs=1048576 count=4097 | openssl dgst -blake2s256 
BLAKE2S-256(stdin)= 08a966f3adf402721335271ed9b42857b60ae0137ebd19da05255607b0ee9e69
```

All good 😄

src/lib/hash/blake2s/blake2s.h
KaganCanSit: When I check, Clang-Tidy throws an error because the class components haven't been initialized. The `state_init` function isn't constructor initialized directly,  (constructor(param) : m_param(param) )after all. Instead of suppressing the Clang warning, I'll accept a few extra loops.

```bash
Checked src/lib/hash/blake2/blake2b.cpp
### Errors in src/lib/hash/blake2s/blake2s.cpp
src/lib/hash/blake2s/blake2s.cpp:156:1: warning: constructor does not initialize these fields: m_bytes_processed, m_h, m_outlen [cppcoreguidelines-pro-type-member-init,hicpp-member-init]
  156 | BLAKE2s::BLAKE2s(std::size_t output_bits) {
      |
```

src/lib/hash/blake2s/blake2s.cpp
reneme: > shouldn't the checks be < instead of <=?

Yes, sorry.

KaganCanSit: Ok, I already fix this.

KaganCanSit:
The reviews here (especially https://github.com/randombit/botan/pull/5117#discussion_r2425894564 and https://github.com/randombit/botan/pull/5117#discussion_r2425431472) gave me a good perspective as a developer. Thank you @reneme for taking the time to share your knowledge. You've already made a lot of changes, and I've fixed a few remaining minor issues.

It was a lot of fun collaborating on an open-source development branch. Thanks @huven.

I think this branch is ready now.

src/lib/hash/blake2s/blake2s.cpp
reneme: **First and foremost:** this is dead code that we can't test through the `HashFunction` interface. I appreciate that it could become useful when implementing the MAC. But it doesn't justify maintaining untested and/or dead code. Let's remove this and leave a comment like:

```C++
// BLAKE2s is specified as a message authentication code. For that, the
// key would need to be zero-padded and incorporated into the initial hash
// state here. See RFC 7693 Section 3.3 and Appendix D.2 `blake2s_init()`.
```

---

While the current suggestion of "fill up with zeros" after `add_data()` might actually work, it isn't clear control flow to my mind. What I would do instead, would probably be along those lines (untested):

```C++
if (keylen > 0) {
   BOTAN_ASSERT_NOMSG(m_buffer.elements_in_buffer() == 0);
   BOTAN_ASSERT_NOMSG(m_buffer.size() <= keylen);
   m_buffer.append({key, keylen});
   m_buffer.fill_up_with_zeros();
   compress(false, m_buffer.consume());
   m_bytes_processed = block_size;
}
```

... main advantage: this doesn't mess with any state of the hash function object but straight-out implements what the RFC demands: "zero-pad the key and compress it into the initial hash state". But again: _let's just remove dead code._

src/lib/hash/blake2s/blake2s.h
reneme: Yeah, I agree. Thanks for trying.

src/lib/hash/blake2s/blake2s.cpp
```diff
@@ -13,19 +15,25 @@
...
    0x6A09E667, 0xBB67AE85, 0x3C6EF372, 0xA54FF53A, 0x510E527F, 0x9B05688C, 0x1F83D9AB, 0x5BE0CD19};
 
 // Mixing function G.
 
-inline void B2S_G(uint8_t a, uint8_t b, uint8_t c, uint8_t d, uint32_t x, uint32_t y, uint32_t* v) {
+constexpr void B2S_G(uint8_t a, uint8_t b, uint8_t c, uint8_t d, uint32_t x, uint32_t y, std::span<uint32_t, 16> v) {
+   BOTAN_DEBUG_ASSERT(a < v.size());
+   BOTAN_DEBUG_ASSERT(b < v.size());
+   BOTAN_DEBUG_ASSERT(c < v.size());
+   BOTAN_DEBUG_ASSERT(d < v.size());
```
reneme: Being the template-fanboy that I am, alternatively this could be:


```suggestion
template <uint8_t a, uint8_t b, uint8_t c, uint8_t d>
   requires(a < 16 && b < 16 && c < 16 && d < 16)
constexpr void B2S_G(uint32_t x, uint32_t y, std::span<uint32_t, 16> v) {
```

... then the constraints on the offset constants would always be checked at compile time, but the call site would arguably be somewhat awkward as:

```C++
   for(const auto& perm : sigma) {  // ten rounds
      B2S_G<0, 4, 8, 12>(m[perm[0]], m[perm[1]], v);
      B2S_G<1, 5, 9, 13>(m[perm[2]], m[perm[3]], v);
      B2S_G<2, 6, 10, 14>(m[perm[4]], m[perm[5]], v);
      B2S_G<3, 7, 11, 15>(m[perm[6]], m[perm[7]], v);
      B2S_G<0, 5, 10, 15>(m[perm[8]], m[perm[9]], v);
      B2S_G<1, 6, 11, 12>(m[perm[10]], m[perm[11]], v);
      B2S_G<2, 7, 8, 13>(m[perm[12]], m[perm[13]], v);
      B2S_G<3, 4, 9, 14>(m[perm[14]], m[perm[15]], v);
   }
```

_... really not necessary to change that, though._ The compiler would likely generate the exact same machine code either way.

reneme (changes requested):
We should work on this: https://github.com/randombit/botan/pull/5117#discussion_r2428066967

Otherwise the changes look good to me.

@KaganCanSit One nit-picky remark regarding commit messages and author attribution: I saw that you wanted to use `Co-Authored-By ...` to mention @huven and me. Thanks for that, very thoughtful of you. However, GitHub doesn't seem to pick that up if you write those annotations inline. Perhaps like so:

```
Modernize blake2s performance (eliminating per-byte loops) and solve clang-tidy suppress (todo message etc.)

Optimization for performance was performed by @huven
Use 64-bit byte counter and AlignmentBuffer in BLAKE2s by @reneme

Co-Authored-By: Richard Huveneers <richard@huveneers.com>
Co-Authored-By: René Meusel <github@renemeusel.de>
```

src/lib/hash/blake2s/blake2s.cpp
huven: Removing is fine with me. That will reduce the state_init signature to

```
void BLAKE2s::state_init(std::size_t outlen)
```

and m_h[0] initialization to

```
m_h[0] ^= 0x01010000 ^ outlen;
```

src/lib/hash/blake2s/blake2s.cpp
KaganCanSit: I'm not very familiar with templates, as they aren't as powerful as they were in C++17/20 under the old standards. However, I think we can avoid the verification cost for debug builds. For me, it's a great opportunity to experiment and tinker around. 😄 Thanks.

KaganCanSit:
* The current branch rebased master.
* The commit message was edited to properly handle `Co-Authored-By` statements.
* Template usage was implemented.
* ~https://github.com/randombit/botan/pull/5117#discussion_r2428066967 is still being reviewed.~ Update:
   * The `key` and `keylen` parameters in the `state_init` function have been removed.
   * The specified comment (@reneme) lines were added to the beginning of the `state_init` function.
   * The initialization of the `n_h[0]` parameter has been updated according to @huven's suggestion.
   * The comment line that lost its meaning as a result of the edit has been removed.
        ```cpp
        //      Secret key (also <= 32 bytes) is optional (keylen = 0).
        //      (keylen=0: no key)
        ```
   * The constructor `BLAKE2s::BLAKE2s` and `clear` functions has been updated because the `state_init` signature has changed.
   * Compile, test and speed:
      * `ninja clean && ./configure.py --without-documentation --cc=clang --compiler-cache=ccache --build-targets=static,cli,tests --build-tool=ninja && ninja`
      * `./botan-test hash blake2s` -> `Tests complete ran 85816 tests in 269.96 msec all tests ok`
      * `./botan speed BLAKE2s` -> `BLAKE2s(256) hash buffer size 1024 bytes: 600.969 MiB/sec 4.27 cycles/byte (300.50 MiB in 500.03 ms)`
      * Check good 😃  
      ```bash
           % dd if=/dev/zero bs=1048576 count=4097 | ./botan hash --algo='BLAKE2s(256)' --format=hex  --buf-size=1024
           % dd if=/dev/zero bs=1048576 count=4097 | openssl dgst -blake2s256 
        ```
      * Speed Test Detail: 
         ```bash
               Virtual Machine
               Debian GNU/Linux 13 (trixie)
               CPU(s): 12
               Model name: 12th Gen Intel(R) Core(TM) i7-12700H
               RAM: 16GB
         ```

src/lib/hash/blake2s/blake2s.cpp
KaganCanSit: All arrangements have been made. Check https://github.com/randombit/botan/pull/5117#issuecomment-3403218067

reneme (approved):
LGTM. Thanks. It was fun to collaborate on this.
@randombit do you want to take another look?

src/lib/hash/blake2s/blake2s.h
```diff
@@ -9,22 +11,31 @@
...
 
+      BLAKE2s(const BLAKE2s&) = default;
+      BLAKE2s& operator=(const BLAKE2s&) = delete;
+      BLAKE2s(BLAKE2s&&) = delete;
+      BLAKE2s& operator=(BLAKE2s&&) = delete;
+
       std::string name() const override;
 
-      size_t output_length() const override { return m_outlen; }
+      std::size_t output_length() const override { return m_outlen; }
```
randombit: In general we don't apply `std::` prefix on `size_t` or the `uintN_t` integer types, instead relying on the `using` declarations in `types.h`

KaganCanSit: Thank you for your feedback. It's just a habit; I used it in a few places in this file. I'll fix it right away.

KaganCanSit:
You are welcome @randombit, it was truly an enjoyable and wonderful experience. 

I wanted to contact you directly because I didn't think it would be appropriate to bring this up in GitHub discussions. I've been following the Botan project for a long time and I really like it. It's a very valuable project for me in terms of my learning process and portfolio. Although my contributions have been small so far, I plan to continue making meaningful contributions to the project and contribute more as I progress. Of course, I will make mistakes and have shortcomings, but I don't want to stay away from the project and the code. 

If it's okay with you, I'd like to add my name and basic information to the doc/credits.txt file in another merge request. I believe this would be a great step towards meeting knowledgeable developers and further improving myself. 

Thank you for your time. Best regards.

PR #5118, opened 2025-10-11
Update TPM2 session FIXME comments to document intentional implicit conversions

KaganCanSit:
Hello Botan Developer Team,

~This pull request handling Clang-Tidy suppression in the `tmp2_session.h` file.~

~- Mark SessionBundle constructor as explicit~
~- Mark SessionHandle::operator ESYS_TR() as explicit~

~Source: [C++ Core Guidlines C.46 By default, declare single-argument constructors explicit](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c46-by-default-declare-single-argument-constructors-explicit)~

UPDATE:
This PR updates FIXME comments in `tpm2_session.h` to properly document the intentional use of implicit conversions in TPM2 session wrapper classes.

- Updated `SessionHandle::operator ESYS_TR()` NOLINTNEXTLINE comment
- Updated `SessionBundle` constructor NOLINTNEXTLINE comment
- No functional changes to code behavior

If need any change or edits, please write comment to me. I fix.
Regards.

### DIFF
--- a/src/lib/prov/tpm2/tpm2_session.h
+++ b/src/lib/prov/tpm2/tpm2_session.h
@@ -70,7 +70,7 @@ class BOTAN_UNSTABLE_API SessionHandle final {
 
       ~SessionHandle();
 
-      // NOLINTNEXTLINE(*-explicit-conversions) FIXME
+      // NOLINTNEXTLINE(*-explicit-conversions) Intentional: SessionHandle is a wrapper around ESYS_TR for C API interop
       [[nodiscard]] operator ESYS_TR() && noexcept;
 
    private:
@@ -165,7 +165,7 @@ class BOTAN_PUBLIC_API(3, 6) Session {
  */
 class SessionBundle {
    public:
-      // NOLINTNEXTLINE(*-explicit-conversions) FIXME
+      // NOLINTNEXTLINE(*-explicit-conversions) Intentional: Allows convenient single-session usage without explicit SessionBundle construction
       SessionBundle(std::shared_ptr<Session> s1 = nullptr,
                     std::shared_ptr<Session> s2 = nullptr,
                     std::shared_ptr<Session> s3 = nullptr) :


src/lib/prov/tpm2/tpm2_session.h
```diff
@@ -70,8 +70,7 @@ class BOTAN_UNSTABLE_API SessionHandle final {
 
       ~SessionHandle();
 
-      // NOLINTNEXTLINE(*-explicit-conversions) FIXME
-      [[nodiscard]] operator ESYS_TR() && noexcept;
+      [[nodiscard]] explicit operator ESYS_TR() && noexcept;
```
reneme: I would argue that this is fine and the `FIXME` should be replaced by:

```C++
// NOLINTNEXTLINE(*-explicit-conversions) SessionHandle is meant to be a wrapper
```

The `SessionHandle` is meant to be a thin wrapper around the `ESYS_TR` type. The implicit conversion is even scoped to just rvalue objects. It allows to conveniently pass wrapped TPM2 session handles into the ESYS API. After all we're wrapping a C API here, some linting excuses should be okay, in my opinion.

Apart from that: This is public API and must not be changed outside of major releases.

src/lib/prov/tpm2/tpm2_session.h
```diff
@@ -165,10 +164,11 @@ class BOTAN_PUBLIC_API(3, 6) Session {
...
    public:
-      // NOLINTNEXTLINE(*-explicit-conversions) FIXME
-      SessionBundle(std::shared_ptr<Session> s1 = nullptr,
-                    std::shared_ptr<Session> s2 = nullptr,
-                    std::shared_ptr<Session> s3 = nullptr) :
+      SessionBundle() = default;
+
+      explicit SessionBundle(std::shared_ptr<Session> s1,
+                             std::shared_ptr<Session> s2 = nullptr,
+                             std::shared_ptr<Session> s3 = nullptr) :
```
reneme: This implicit conversion was intentional, providing API ergonomics for the wrapped TPM2 functionality. TPM2 functions can take up to three "session" objects to control access to TPM resources in different ways. In my experience, one tends to use only a single session handle most of the time, though. So being able to pass a single session object to a wrapped TPM function (without worrying about the `SessionBundle` wrapper) is actually helping readability of downstream code without obscuring the intent.

Also: This is public API and must not be changed outside a major release.

src/lib/prov/tpm2/tpm2_session.h
reneme: This is intentional. `SessionHandle` is meant to be an oblivious wrapper around `ESYS_TR`. Especially given that this implicit conversion is scoped to rvalue objects I think we can live with the linter exception here. We're wrapping a C API and trying to add some C++'y convenience to it. Some linting liberty should be granted. ;)

I suggest to remove the `FIXME`. Also: This is public API that must not be changed outside of major releases.

KaganCanSit:
> Unfortunately I have to oppose this PR. The implicit conversion operators and constructors were intentional for API ergonomics. The required adaptions in the implementation (adding and explicit constructor calls in many places) are exactly what they were meant to avoid. 🙂`static_cast`

Alright. Then I'm retrieving the header contents here in a similar way to #5119.

* Remove explicit qualifiers.
* Update FIXME messages.
* Review the remaining conversions. Remove or edit “static_cast” contents.

Here, along with “SessionBundle,” most likely this entire branch will be reverted and only the FIXME messages will be updated, but I'm not sure. I'll look into it.

Thanks for review.

KaganCanSit:
> > Unfortunately I have to oppose this PR. The implicit conversion operators and constructors were intentional for API ergonomics. The required adaptions in the implementation (adding and explicit constructor calls in many places) are exactly what they were meant to avoid. 🙂`static_cast`
> 
> Alright. Then I'm retrieving the header contents here in a similar way to #5119.
> 
> * Remove explicit qualifiers.
> * Update FIXME messages.
> * Review the remaining conversions. Remove or edit “static_cast” contents.
> 
> Here, along with “SessionBundle,” most likely this entire branch will be reverted and only the FIXME messages will be updated, but I'm not sure. I'll look into it.
> 
> Thanks for review.

Thank you for the feedback again. I apologize for the confusion.

I misunderstood the scope - I focused on the FIXME comments but incorrectly assumed the implicit conversions should be removed. I now understand they are  intentional design choices for API ergonomics.

I updated this PR to only replace the FIXME comments in `tpm2_session.h` with proper explanations, and revert all implementation changes. And I change commit, PR text contents.

PR #5119, opened 2025-10-12
Resolve FIXME: Enforce explicit conversion for Context to ESYS_CONTEXT*

KaganCanSit:
Hello Botan Developer Team,

This pull request handling Clang-Tidy suppression in the `tpm2_context.h` file. Resolve FIXME annotation in 'tpm2_context.h' by enforcing explicit conversion from 'Context' to 'ESYS_CONTEXT*', improving type safety.

Changes:
* Added 'explicit' keyword to 'operator ESYS_CONTEXT*()'
* Removed NOLINT suppression directive

NOTE: There is a conflict with branch #5118 here. Therefore, it may be beneficial to track and manage the merges sequentially.

Source: [C++ Core Guidlines C.46 By default, declare single-argument constructors explicit](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c46-by-default-declare-single-argument-constructors-explicit)

If need any change or edits, please write comment to me. I fix.
Regards.

### DIFF
--- a/src/lib/prov/tpm2/tpm2_context.h
+++ b/src/lib/prov/tpm2/tpm2_context.h
@@ -101,7 +101,7 @@ class BOTAN_PUBLIC_API(3, 6) Context final : public std::enable_shared_from_this
       /// @return an ESYS_CONTEXT* for use in other TPM2 functions.
       ESYS_CONTEXT* esys_context() noexcept;
 
-      // NOLINTNEXTLINE(*-explicit-conversions) FIXME
+      // NOLINTNEXTLINE(*-explicit-conversions) Intentional: enables transparent ESYS_CONTEXT* wrapper usage
       operator ESYS_CONTEXT*() noexcept { return esys_context(); }
 
       /// @return the Vendor of the TPM2
--- a/src/lib/prov/tpm2/tpm2_hash.cpp
+++ b/src/lib/prov/tpm2/tpm2_hash.cpp
@@ -71,7 +71,7 @@ void HashFunction::lazy_setup() {
 
    const auto auth = init_empty<TPM2B_AUTH>();
    const auto rc = check_rc_expecting<TPM2_RC_HASH>("Esys_HashSequenceStart",
-                                                    Esys_HashSequenceStart(*m_handle.context(),
+                                                    Esys_HashSequenceStart(m_handle.context()->esys_context(),
                                                                            m_sessions[0],
                                                                            m_sessions[1],
                                                                            m_sessions[2],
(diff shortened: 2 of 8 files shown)


src/lib/prov/tpm2/tpm2_context.h
```diff
@@ -101,8 +101,7 @@ class BOTAN_PUBLIC_API(3, 6) Context final : public std::enable_shared_from_this
       /// @return an ESYS_CONTEXT* for use in other TPM2 functions.
       ESYS_CONTEXT* esys_context() noexcept;
 
-      // NOLINTNEXTLINE(*-explicit-conversions) FIXME
-      operator ESYS_CONTEXT*() noexcept { return esys_context(); }
+      [[nodiscard]] explicit operator ESYS_CONTEXT*() noexcept { return esys_context(); }
```
reneme: This is public API and shouldn't be changed in incompatible ways outside of major releases. Please revert this and add a `TODO(Botan4):` in front of the `FIXME` so it won't be forgotten.

KaganCanSit: Sory, I hadn't considered it from that perspective. I will look into it and fix it today between 8:00 PM and 11:00 PM (UTC +3). Thanks for fast review.

reneme: Actually, after having looked at #5118, I would argue this was intentional and should remain allowed with the `FIXME` removed. This class is meant as a wrapper for `ESYS_CONTEXT` and the implicit conversion is there to simplify its usage.

Not relying on the implicit conversion internally (and using the explicit `esys_context()` method instead) is certainly fine though. So the rest of this PR may change as it is from my point of view.

KaganCanSit: I understand. In this case;

* I'm removing the explicit qualifier.
* I'm keeping the remaining esys_context()

I'll keep the Clang-Tidy suppression. However, an explanation might be better than a FIXME message. Other developers might take the same approach as me. Any idea?

reneme: > an explanation might be better than a FIXME message. Other developers might take the same approach as me. Any idea?

Full ack. Perhaps: "intentional, for oblivious wrapper"

KaganCanSit: I revert change and update comment.

```cpp
// NOLINTNEXTLINE(*-explicit-conversions) Intentional: enables transparent ESYS_CONTEXT* wrapper usage
```

PR #5121, opened 2025-10-15
Rework ChaCha_RNG

thillux:
When reading the source code of the ChaCha_RNG I saw some room for improvement:
- style: use initialization lists
- seeding: use HMAC(SHA-512) instead of HMAC(SHA-256) for less effects of the injectivity of hash functions
- internal state: add optional chacha key overwrite after each operation in order to provide backtracking resistance
- internal state: also use IV/nonce to widen effective key material

I'll adapted KAT test results via copy paste from CLI output and can also generate them externally if needed.


thillux:
CC @reneme

src/lib/rng/chacha_rng/chacha_rng.h
```diff
@@ -116,6 +163,11 @@ class BOTAN_PUBLIC_API(2, 3) ChaCha_RNG final : public Stateful_RNG {
 
       std::unique_ptr<MessageAuthenticationCode> m_hmac;
       std::unique_ptr<StreamCipher> m_chacha;
+      const bool m_fast_key_erasure;
+      const size_t m_chacha_keylen;
+
+      const static inline char* m_stream_cipher_algo = "ChaCha(20)";
+      const static inline char* m_hmac_algo = "HMAC(SHA-512)";
```
reneme: Could be made constexpr.

```suggestion
      constexpr static std::string_view stream_cipher_algo = "ChaCha(20)";
      constexpr static std::string_view hmac_algo = "HMAC(SHA-512)";
```

Also, some style nits: Personally, I made it a habit to place such "public" constants (as in: defined in the header, not necessarily with public visibility) at the top of class definitions so that they don't get missed so easily when reading the code base. Additionally, to the best of my knowledge, constants aren't treated as member variables (i.e. they don't bear the `m_` prefix).

src/lib/rng/chacha_rng/chacha_rng.cpp
```diff
@@ -11,45 +11,58 @@
...
+      m_fast_key_erasure(fast_key_erasure),
+      m_chacha_keylen(m_chacha->key_spec().maximum_keylength()) {
    clear();
 }
 
 void ChaCha_RNG::clear_state() {
    m_hmac->set_key(std::vector<uint8_t>(m_hmac->output_length(), 0x00));
-   m_chacha->set_key(m_hmac->final());
+   const auto chacha_key = m_hmac->final();
+   m_chacha->set_key(chacha_key.data(), m_chacha_keylen);
```
reneme: I'm guessing we can be sure that the HMAC always produces enough bytes for ChaCha's key length, but an assertion would probably not hurt here. Also, let's try to avoid fiddling with raw pointers.


```suggestion
   const auto chacha_key = m_hmac->final();
   BOTAN_ASSERT_NOMSG(chacha_key.size() >= m_chacha_keylen);
   m_chacha->set_key(std::span{chacha_key}.first(m_chacha_keylen));
```

src/lib/rng/chacha_rng/chacha_rng.cpp
```diff
@@ -60,11 +73,18 @@ void ChaCha_RNG::generate_output(std::span<uint8_t> output, std::span<const uint
    }
 
    m_chacha->write_keystream(output);
+
+   // optionally overwrite key after each output operation for backtracking resistance
+   if(m_fast_key_erasure) {
+      const auto next_key = m_chacha->keystream_bytes(m_chacha_keylen);
+      m_chacha->set_key(next_key);
```
reneme: This is purely a question of taste. No real need for the `next_key` variable in my opinion, though.

```suggestion
      m_chacha->set_key(m_chacha->keystream_bytes(m_chacha_keylen));
```

src/lib/rng/chacha_rng/chacha_rng.cpp
```diff
@@ -60,11 +73,18 @@ void ChaCha_RNG::generate_output(std::span<uint8_t> output, std::span<const uint
...
+   }
 }
 
 void ChaCha_RNG::update(std::span<const uint8_t> input) {
    m_hmac->update(input);
-   m_chacha->set_key(m_hmac->final());
+   const auto chacha_key = m_hmac->final();
+   m_chacha->set_key(chacha_key.data(), m_chacha_keylen);
```
reneme: As mentioned above: an assertion would be nice here. Maybe extract into a helper function and reuse it in both locations.

```C++
namespace {

std::span<const uint8_t> truncate(std::span<const uint8_t> in, size_t bytes) {
   BOTAN_ASSERT_NOMSG(in.size() >= bytes);
   return in.first(bytes);
}

}
```

```C++
void ChaCha_RNG::update(std::span<const uint8_t> input) {
   m_hmac->update(input);
   m_chacha->set_key(truncate(m_hmac->final(), m_chacha_keylen));
   // [...]
}
```

src/lib/rng/chacha_rng/chacha_rng.h
```diff
@@ -116,6 +163,11 @@ class BOTAN_PUBLIC_API(2, 3) ChaCha_RNG final : public Stateful_RNG {
 
       std::unique_ptr<MessageAuthenticationCode> m_hmac;
       std::unique_ptr<StreamCipher> m_chacha;
+      const bool m_fast_key_erasure;
+      const size_t m_chacha_keylen;
```
reneme: Please don't mark member variables as `const` because this implicitly disables the move assignment operator unnecessarily. 


```suggestion
      bool m_fast_key_erasure;
      size_t m_chacha_keylen;
```

src/lib/rng/chacha_rng/chacha_rng.h
```diff
@@ -47,7 +76,7 @@ class BOTAN_PUBLIC_API(2, 3) ChaCha_RNG final : public Stateful_RNG {
       * in response. In this case, an exception will be thrown rather
       * than generating duplicated output.
       */
-      ChaCha_RNG();
+      explicit ChaCha_RNG(bool fast_key_erasure = false);
```
reneme: Please add the documentation for `fast_key_erasure` here as well.

src/lib/rng/chacha_rng/chacha_rng.h
```diff
@@ -116,6 +164,14 @@ class BOTAN_PUBLIC_API(2, 3) ChaCha_RNG final : public Stateful_RNG {
...
       std::unique_ptr<MessageAuthenticationCode> m_hmac;
       std::unique_ptr<StreamCipher> m_chacha;
+      const bool m_fast_key_erasure;
+      const size_t m_chacha_keylen;
+
+      const static inline char* m_stream_cipher_algo = "ChaCha(20)";
+      const static inline char* m_hmac_algo = "HMAC(SHA-512)";
+
+      // use "classic" 8 byte nonce as extended key material
+      const static inline size_t m_chacha_iv_len = 8;
```
reneme: Similarly as above: Please move to the top of the class and make it `constexpr size_t`

reneme (commented):
Just some general remarks so far.

thillux:
@reneme Thanks for your review, I probably addressed everything mentioned. I left your suggestions open for you to compare and close, as I did not apply them 1:1. My code also uses the IV/Nonce of ChaCha20 now to extend the key material. I therefore centrally check the length assertion in the suggested truncate function (now a split into two sub-key chunks). I needed some additional variables typically named `key_material`, otherwise e.g. `m_hmac->final()` would be vanished before the spans taken from it are used. Tests did not deliver reproducible known answers with your suggestion applied 1:1.

thillux:
Just FYI speed comparison on `AMD Ryzen 9 5950X` with different buffer sizes (motivating default disable of fast key erasure):

## 4 Byte
```
ChaCha_RNG generate buffer size 4 bytes: 12.029 MiB/sec 269.56 cycles/byte (6.06 MiB in 503.99 ms)
ChaCha_RNG (with key erasure) generate buffer size 4 bytes: 5.607 MiB/sec 578.24 cycles/byte (2.81 MiB in 501.56 ms)
```

## 8 Byte
```
ChaCha_RNG generate buffer size 8 bytes: 24.119 MiB/sec 134.44 cycles/byte (12.06 MiB in 500.13 ms)
ChaCha_RNG (with key erasure) generate buffer size 8 bytes: 11.219 MiB/sec 289.01 cycles/byte (5.62 MiB in 501.37 ms)
```

## 16 Byte
```
ChaCha_RNG generate buffer size 16 bytes: 47.715 MiB/sec 67.96 cycles/byte (23.88 MiB in 500.37 ms)
ChaCha_RNG (with key erasure) generate buffer size 16 bytes: 22.178 MiB/sec 146.20 cycles/byte (11.12 MiB in 501.62 ms)
```

## 32 Byte
```
ChaCha_RNG generate buffer size 32 bytes: 93.908 MiB/sec 34.53 cycles/byte (47.00 MiB in 500.49 ms)
ChaCha_RNG (with key erasure) generate buffer size 32 bytes: 44.689 MiB/sec 72.56 cycles/byte (22.38 MiB in 500.68 ms)
```

## 64 Byte
```
ChaCha_RNG generate buffer size 64 bytes: 181.763 MiB/sec 17.84 cycles/byte (90.94 MiB in 500.31 ms)
ChaCha_RNG (with key erasure) generate buffer size 64 bytes: 89.538 MiB/sec 36.21 cycles/byte (44.81 MiB in 500.49 ms)
```

## 512 Byte
```
ChaCha_RNG generate buffer size 512 bytes: 1035.250 MiB/sec 3.13 cycles/byte (517.62 MiB in 500.00 ms)
ChaCha_RNG (with key erasure) generate buffer size 512 bytes: 588.339 MiB/sec 5.51 cycles/byte (294.19 MiB in 500.03 ms)
```

## 1024 Byte
```
ChaCha_RNG generate buffer size 1024 bytes: 1590.335 MiB/sec 2.04 cycles/byte (795.19 MiB in 500.01 ms)
ChaCha_RNG (with key erasure) generate buffer size 1024 bytes: 1001.902 MiB/sec 3.24 cycles/byte (501.00 MiB in 500.05 ms)
```

All taken from `./botan speed RNG --buf-size=XXX`

thillux:
Ready from my side, will squash, when review is done.

src/lib/rng/chacha_rng/chacha_rng.cpp
thillux: Done, but helper functions splits key material now, as I also use IV/Nonce.

src/lib/rng/chacha_rng/chacha_rng.cpp
thillux: Left these additional variables in the code now, as I take multiple spans and need intermediate values to stay on the stack.

src/lib/rng/chacha_rng/chacha_rng.cpp
thillux: Assertion is centrally checked in split function.

src/lib/rng/chacha_rng/chacha_rng.h
thillux: Done and moved.

src/lib/rng/chacha_rng/chacha_rng.cpp
```diff
@@ -60,16 +75,33 @@ void ChaCha_RNG::generate_output(std::span<uint8_t> output, std::span<const uint
    }
 
    m_chacha->write_keystream(output);
+
+   // optionally overwrite key after each output operation for backtracking resistance
+   if(m_fast_key_erasure) {
+      // additional variable for key material is needed to stay on the stack
+      auto key_material = m_chacha->keystream_bytes(chacha_key_len + chacha_iv_len);
```
reneme: Nit:


```suggestion
      const auto key_material = m_chacha->keystream_bytes(chacha_key_len + chacha_iv_len);
```

src/lib/rng/chacha_rng/chacha_rng.cpp
reneme: Taking the helper even further (and making it a private member function) we could avoid returning a std::span that might become dangling. (What a blunder on my side, yikes!)

The `BufferSlicer` checks for appropriate buffer lengths internally, no need for the assertion anymore.

```C++

void ChaCha_RNG::update_chacha_state(std::span<const uint8_t> key_material) {
   BufferSlicer bs(key_material);
   m_chacha->set_key(bs.take<chacha_key_len>());
   m_chacha->set_iv(bs.take<chacha_iv_len>());
}
```

reneme: Here's a patch that you could just cherry-pick: https://github.com/reneme/botan/commit/5f7ea6ae11f3dbcc2af0634ef38dee8ed10b4031

reneme (commented):
@thillux I didn't realize I am authorized to push into your fork 😨 I meant to push into my own branch, but a sole `git push -f` was a bit too vigorous. Please make sure I didn't accidentally crush something. Sorry about that.

---

Changes itself look good to me and I think the improvements make sense. Thanks!

To the best of my understanding, the ChaCha RNG isn't standardized in any way, so we should have the liberty to improve the implementation as suggested. However, depending on how pedantically we want to follow semantic versioning, these changes might be seen as a violation. E.g. if someone based their unit tests on a hard-coded seed for this RNG their tests would break (much like Botan's KATs had to be updated).

As far as I can see, "breaking" changes are:

* "HMAC(SHA-256)" replaced by "HMAC(SHA-512)"
* derivation and usage of an IV for the internal ChaCha updates

@randombit What's your take on that? Do we want to guarantee stable DRNG outputs across minor versions?

thillux:
Your push did not break things for me :). I'll squash together now.

thillux:
@reneme  Squashed everything into two commits (change DRNG + tests).

src/lib/rng/chacha_rng/chacha_rng.h
```diff
@@ -20,24 +20,64 @@ class Entropy_Sources;
...
+*   regarding its seed due to being a random permutation for each key.
+*   Furthermore ChaCha has a 512 bit block width, which shifts block
+*   collisions in a very unlikely range (regarding output block count).
+* - Using ChaCha(20) instead of ChaCha(8) or ChaCha(12) has the
+*   advantage of being a conservative choice also taken by the
+*   Linux kernel, where it is already accepted as a secure CSPRNG
+*   implementation by many people and organizations.
+* - Providing optional fast key erasure is necessary to reach
+*   backtracking resistance of the internal state. This costs
+*   performance depending on the requests sizes of the user.
```
reneme:
```suggestion
* - Providing optional fast key erasure is necessary to reach
*   backtracking resistance of the internal state. This costs
*   performance depending on the request sizes of the user.
```

reneme (approved):
Changes look good to me, apart from the semver concerns and a typo.

Oh, and please note and update [the related documentation](https://botan.randombit.net/handbook/api_ref/rng.html#chacha-rng) in `rng.rst`

thillux:
I updated the documentation.

thillux:
While there, I saw, that ESDM_RNG is not documented there. Can I do it in this PR or open another one?

reneme:
> Can I do it in this PR or open another one?

Please open a separate one.

randombit:
I am a bit hesitant to change the output bytes here since (in a Hyrum's Law sense) the `ChaCha_RNG` output has been stable since introduction and hypothetically someone might be depending on this. That said
* The docs never guaranteed that it was stable over time
* Applications relying on this kind of RNG stability are *relatively* uncommon
* The library doesn't generally do a good job of respecting RNG stability anyway (outside of a few specific cases like RSA and ECDSA where we test that it works)

So probably fine to do.

I would request that the doc update note that a) the output changed in 3.10 b) the output might change again at any future time, and if RNG stability is required to use `HMAC_DRBG` instead.

[I still do want to review this PR fully before merging]

thillux:
@randombit I added some more notes on output stability as requested.

thillux:
Shall I rebase this branch?

thillux:
@randombit I rebased and fixed compilation against recent master.

thillux:
Formatting is still broken. Will work on this later this day.

thillux:
Your CI seems to be happy again now. :-)

doc/api_ref/rng.rst
```diff
@@ -190,21 +194,29 @@ for this use by NIST.
...
 
-This is a very fast userspace PRNG based on ChaCha20 and HMAC(SHA-256). The key
-for ChaCha is derived by hashing entropy inputs with HMAC. Then the ChaCha
+This is a very fast userspace PRNG based on ChaCha20 and HMAC(SHA-512). The key
+and IV for ChaCha are derived by hashing entropy inputs with HMAC. Then the ChaCha
 keystream generator is run, first to generate the new HMAC key (used for any
 future entropy additions), then the desired RNG outputs.
 
+Please note, that you should not rely on the stability of this RNG's outputs
+regarding identical seeds across versions. The output changed in version 3.10.0
```
randombit: Should be 3.13.0 here

src/lib/rng/chacha_rng/chacha_rng.cpp
```diff
@@ -59,17 +69,28 @@ void ChaCha_RNG::generate_output(std::span<uint8_t> output, std::span<const uint
       update(input);
    }
 
+   //  Always generate next key material from the first blocks generated by a key
+   //  as it removes a positional argument from a possible analysis.
```
randombit: I'm having a difficult time parsing this comment

src/lib/rng/chacha_rng/chacha_rng.cpp
```diff
@@ -8,48 +8,58 @@
...
 }
 
 void ChaCha_RNG::clear_state() {
    m_hmac->set_key(std::vector<uint8_t>(m_hmac->output_length(), 0x00));
-   m_chacha->set_key(m_hmac->final());
+   update_chacha_state(m_hmac->final());
+}
+
+void ChaCha_RNG::update_chacha_state(std::span<const uint8_t> key_material) {
+   BufferSlicer bs(key_material);
```
randombit: Just for sanity purposes this should assert that key_material.size() > chacha_key_len + chacha_iv_len, BufferSlicer will catch it at runtime but an assert is cleaner

src/tests/test_rng_behavior.cpp
```diff
@@ -673,20 +673,100 @@ class ChaCha_RNG_Unit_Tests final : public Stateful_RNG_Tests {
...
+      }
+
+      Test::Result test_security_level() override {
+         Test::Result result("ChaCha_RNG(FKE) Security Level");
+         const Botan::ChaCha_RNG rng;
+         result.test_sz_eq("Expected security level", rng.security_level(), size_t(256));
+         return result;
+      }
+
+      Test::Result test_max_number_of_bytes_per_request() override {
```
randombit: This and `test_reseed_interval_limits` below seem pretty pointless as tests

src/lib/rng/chacha_rng/chacha_rng.h
```diff
@@ -60,8 +104,11 @@ class BOTAN_PUBLIC_API(2, 3) ChaCha_RNG final : public Stateful_RNG {
       * than generating duplicated output.
       *
       * @param seed the seed material, should be at least 256 bits
+      * @param fast_key_erasure overwrite state after each operation
+      * for backtracking resistance, costs performance depending
+      * on request size mix, deactivated by default
       */
-      BOTAN_FUTURE_EXPLICIT ChaCha_RNG(std::span<const uint8_t> seed);
+      BOTAN_FUTURE_EXPLICIT ChaCha_RNG(std::span<const uint8_t> seed, bool fast_key_erasure = false);
```
randombit: So recently we've started using named enums instead of bools and I think it leads to quite a bit more readable code vs raw `true`/`false`, eg a class-local `enum class KeyErasure : uint8_t { NoErasure, WithKeyErasure }`

randombit (changes requested):
So change seems fine we can go ahead. A few minor comments. Also please rebase for CI purposes since at this point the base commit is pretty old and there may be some additional eg `clang-tidy` warnings that pop up

thillux:
On this now. Sorry for the delay.

src/lib/rng/chacha_rng/chacha_rng.h
thillux: Done.

src/tests/test_rng_behavior.cpp
thillux: They were there to satisfy the virtual base class. At least not duplicated now, by deriving from the ChaCha_RNG_Unit_Tests.

src/lib/rng/chacha_rng/chacha_rng.cpp
thillux: Added.

doc/api_ref/rng.rst
thillux: Fixed.

src/lib/rng/chacha_rng/chacha_rng.cpp
thillux: I changed this comment. Please check if its clearer now. I can also drop this comment, if it adds nothing worth knowing.

thillux:
These CI failures seem to be unrelated to my changes:

```
[0](https://github.com/randombit/botan/actions/runs/31366759069/job/93386862253?pr=5121#step:6:5626)
  PASSED (ClientAuth-NoCertificate-Server-TLS-Sync)
  FAILED (TLS13-1RTT-Client-TLS-Sync)
  unexpected failure
  got:
  	stderr:
  		
  	local: "didResume is false, but we expected the opposite"
  	child: "none"
  	stdout: 
  
  extra stderr:
  
  PASSED (TLS13-HelloRetryRequest-Client-TLS-Sync)
  PASSED (TLS13-HelloRetryRequest-Server-TLS-Sync)
  PASSED (ClientAuth-ECDSA-Client-TLS-Sync)
  PASSED (ClientAuth-NoCertificate-Server-TLS13-TLS-Sync)
  ```
  
  Please let me know if I broke anything here.

**Comment by thillux** [comment]
Updated comment for 3.14 and rebased to master.

**Comment by randombit** [comment]
BoGo still failing, definitely needs some investigation as I have not seen BoGo to be flaky at any point. OTOH I definitely do *not* see how this change could affect BoGo or TLS at all....

**Comment by randombit** [comment]
Oh interesting ... we are using `ChaCha_RNG` with a fixed seed for BoGo, possibly the change in behavior is introducing a flake, will investigate...

File: src/lib/rng/chacha_rng/chacha_rng.cpp [hunk 15]
```diff
@@ -59,17 +71,32 @@ void ChaCha_RNG::generate_output(std::span<uint8_t> output, std::span<const uint
       update(input);
    }
 
+   const bool erase_key = (m_key_erasure == KeyErasure::WithKeyErasure);
+
+   // Use a fixed structure for fast key erasure. Next key is always the first output of
+   // the keystream (without updates in between). Therefore the next key becomes independent
+   // of the request pattern. This should not harm security and may ease mathematical analysis,
+   // as no variable position of the next key in the output stream has to be modeled.
+   const auto key_material = m_chacha->keystream_bytes(erase_key ? chacha_key_len + chacha_iv_len : 0);
```
randombit: This is going to require a heap allocation + deallocation plus a pretty unnecessary zeroization. I'd suggest storing instead to a local array (fixed at 32+8 = 40 bytes)

Also it calls into a virtual even if we're not using key erasure.

Something like this instead

```
std::array<uint8_t, chacha_key_len + chacha_iv_len> key_material{};
if(erase_key) {
  m_chacha->write_keystream(key_material);
}
```

src/lib/rng/chacha_rng/chacha_rng.cpp
```diff
@@ -8,48 +8,60 @@
...
 }
 
 void ChaCha_RNG::clear_state() {
    m_hmac->set_key(std::vector<uint8_t>(m_hmac->output_length(), 0x00));
-   m_chacha->set_key(m_hmac->final());
+   update_chacha_state(m_hmac->final());
+}
+
+void ChaCha_RNG::update_chacha_state(std::span<const uint8_t> key_material) {
+   BOTAN_ASSERT_NOMSG(key_material.size() >= chacha_key_len + chacha_iv_len);
```
randombit, line 66: Should probably be exact equality here, getting pass more implies something went wrong somewhere...

src/lib/rng/chacha_rng/info.txt
randombit: This date should be updated to indicate the changed design

src/lib/rng/chacha_rng/chacha_rng.cpp
```diff
@@ -8,48 +8,60 @@
...
+
+   BufferSlicer bs(key_material);
+   m_chacha->set_key(bs.take<chacha_key_len>());
+   m_chacha->set_iv(bs.take<chacha_iv_len>());
```
randombit: If the span is always checked to be exactly the correct length this could skip the BufferSlicer entirely and use `key_material.first(chacha_key_len)` and `key_material.last(chacha_iv_len)`

randombit (changes requested):
Very close, a few more nits. Also needs a rebase to pick up the BoGo fix

src/lib/rng/chacha_rng/chacha_rng.cpp
thillux: This was not directly possible, as the sum gives 320 bit whereas SHA-512 returns 512 bit. I nevertheless tried to shave off the usage of the BufferSlicer. Please have a look again.

src/lib/rng/chacha_rng/chacha_rng.cpp
thillux: Done.

src/lib/rng/chacha_rng/info.txt
thillux: Done.

thillux:
@randombit Should be done now. Please have a look again at the assert + BufferSlice removal. As this was not directly possible as you suggested.

thillux:
Rebased, to fix the conflict in news.rst.

PR #5124, opened 2025-10-20
Cooperative Cancellation in PasswordHash

### LINKED ISSUE #5123: Cooperative Cancellation in PasswordHash
This issue to check if a PR would be accepted that adds support to abort a key derivation operation. This is useful e.g. when the user aborts a request that needs a computationally expensive key.

This could be implemented by adding a set_stop_token([std::stop_token](https://en.cppreference.com/w/cpp/thread/stop_token.html)) to PasswordHash. A sub-class can check the token at well-chosen intervals. If stop requested, the derivation can throw a cancellation exception to signal that the operation was aborted.

Would something along these lines be a valuable addition, or too specific to my use-case?

As C++20 is already required this should not introduce a new dependency.

huven:
Fixes #5123

Todo:

- [X] Check std::stop_token cross-platform support
- [x] Unittest
- [x] Make all PasswordHash subclasses check stop_token
    - [x] Argon2
    - [x] PBKDF2
    - [x] Scrypt
    - [x] OpenPGP-S2K
    - [x] Bcrypt-PBKDF
    - [x] PKCS12
- [x] Drop added supports_cooperative_cancellation() when all done
- [x] Update documentation

Note:

- The deprecated API (non-PasswordHash based) has not been touched
- I've added the default std::nullopt only to the base class. This is not inherited by virtual overrides, which I think is ok as the default will be taken from the static class. Similar to other methods like from_params.
- macOS runners have been updated from 13/14 to 26
- Added -march=armv6 flag to bare metal build (to ensure availability of atomic operations)
- OpenPGP-S2K has been excluded from unit-testing, as it cannot be tuned above a few seconds on modern hardware, which is impossible to test reliably on GitHub runners.

### DIFF
--- a/src/lib/pbkdf/argon2/argon2.cpp
+++ b/src/lib/pbkdf/argon2/argon2.cpp
@@ -303,7 +303,8 @@ void process_block(secure_vector<uint64_t>& B,
                    size_t threads,
                    uint8_t mode,
                    size_t memory,
-                   size_t time) {
+                   size_t time,
+                   const std::optional<std::stop_token>& stop_token) {
    uint64_t T[128];
    size_t index = 0;
    if(n == 0 && slice == 0) {
@@ -320,6 +321,10 @@ void process_block(secure_vector<uint64_t>& B,
    }
 
    while(index < segments) {
+      if((index & 63) == 0 && stop_token.has_value() && stop_token->stop_requested()) {
+         throw Botan::Operation_Canceled("argon2");
+      }
+
       const size_t offset = lane * lanes + slice * segments + index;
 
       size_t prev = offset - 1;
@@ -350,7 +355,12 @@ void process_block(secure_vector<uint64_t>& B,
    }
 }
 
-void process_blocks(secure_vector<uint64_t>& B, size_t t, size_t memory, size_t threads, uint8_t mode) {
+void process_blocks(secure_vector<uint64_t>& B,
+                    size_t t,
+                    size_t memory,
+                    size_t threads,
+                    uint8_t mode,
+                    const std::optional<std::stop_token>& stop_token) {
    const size_t lanes = memory / threads;
    const size_t segments = lanes / SYNC_POINTS;
 
@@ -365,7 +375,7 @@ void process_blocks(secure_vector<uint64_t>& B, size_t t, size_t memory, size_t 
 
             for(size_t lane = 0; lane != threads; ++lane) {
                fut_results.push_back(thread_pool.run(
-                  process_block, std::ref(B), n, slice, lane, lanes, segments, threads, mode, memory, t));
+                  process_block, std::ref(B), n, slice, lane, lanes, segments, threads, mode, memory, t, stop_token));
             }
 
             for(auto& fut : fut_results) {
@@ -381,7 +391,7 @@ void process_blocks(secure_vector<uint64_t>& B, size_t t, size_t memory, size_t 
    for(size_t n = 0; n != t; ++n) {
       for(size_t slice = 0; slice != SYNC_POINTS; ++slice) {
          for(size_t lane = 0; lane != threads; ++lane) {
-            process_block(B, n, slice, lane, lanes, segments, threads, mode, memory, t);
+            process_block(B, n, slice, lane, lanes, segments, threads, mode, memory, t, stop_token);
          }
       }
    }
@@ -398,7 +408,8 @@ void Argon2::argon2(uint8_t output[],
                     const uint8_t key[],
                     size_t key_len,
                     const uint8_t ad[],
-                    size_t ad_len) const {
+                    size_t ad_len,
+                    const std::optional<std::stop_token>& stop_token) const {
    BOTAN_ARG_CHECK(output_len >= 4 && output_len <= std::numeric_limits<uint32_t>::max(),
                    "Invalid Argon2 output length");
    BOTAN_ARG_CHECK(password_len <= std::numeric_limits<uint32_t>::max(), "Invalid Argon2 password length");
@@ -430,7 +441,7 @@ void Argon2::argon2(uint8_t output[],
    secure_vector<uint64_t> B(memory * 1024 / 8);
 
    init_blocks(B, *blake2, H0, memory, m_p);
-   process_blocks(B, m_t, memory, m_p, m_family);
+   process_blocks(B, m_t, memory, m_p, m_family, stop_token);
 
    clear_mem(output, output_len);
    extract_key(output, output_len, B, memory, m_p);
--- a/src/lib/pbkdf/argon2/argon2.cpp
+++ b/src/lib/pbkdf/argon2/argon2.cpp
@@ -303,7 +303,8 @@ void process_block(secure_vector<uint64_t>& B,
                    size_t threads,
                    uint8_t mode,
                    size_t memory,
-                   size_t time) {
+                   size_t time,
+                   const std::optional<std::stop_token>& stop_token) {
    uint64_t T[128];
    size_t index = 0;
    if(n == 0 && slice == 0) {
@@ -320,6 +321,10 @@ void process_block(secure_vector<uint64_t>& B,
    }
 
    while(index < segments) {
+      if((index & 63) == 0 && stop_token.has_value() && stop_token->stop_requested()) {
+         throw Botan::Operation_Canceled("argon2");
+      }
+
       const size_t offset = lane * lanes + slice * segments + index;
 
       size_t prev = offset - 1;
@@ -350,7 +355,12 @@ void process_block(secure_vector<uint64_t>& B,
    }
 }
 
-void process_blocks(secure_vector<uint64_t>& B, size_t t, size_t memory, size_t threads, uint8_t mode) {
+void process_blocks(secure_vector<uint64_t>& B,
+                    size_t t,
+                    size_t memory,
+                    size_t threads,
+                    uint8_t mode,
+                    const std::optional<std::stop_token>& stop_token) {
    const size_t lanes = memory / threads;
    const size_t segments = lanes / SYNC_POINTS;
 
@@ -365,7 +375,7 @@ void process_blocks(secure_vector<uint64_t>& B, size_t t, size_t memory, size_t 
 
             for(size_t lane = 0; lane != threads; ++lane) {
                fut_results.push_back(thread_pool.run(
-                  process_block, std::ref(B), n, slice, lane, lanes, segments, threads, mode, memory, t));
+                  process_block, std::ref(B), n, slice, lane, lanes, segments, threads, mode, memory, t, stop_token));
             }
 
             for(auto& fut : fut_results) {
@@ -381,7 +391,7 @@ void process_blocks(secure_vector<uint64_t>& B, size_t t, size_t memory, size_t 
    for(size_t n = 0; n != t; ++n) {
       for(size_t slice = 0; slice != SYNC_POINTS; ++slice) {
          for(size_t lane = 0; lane != threads; ++lane) {
-            process_block(B, n, slice, lane, lanes, segments, threads, mode, memory, t);
+            process_block(B, n, slice, lane, lanes, segments, threads, mode, memory, t, stop_token);
          }
       }
    }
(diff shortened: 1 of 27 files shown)


reneme:
Looking at the CI outputs, we seem to get in trouble with macOS (Xcode 13 and 14) and Android NDK. For both of which [we claim to only support the "latest" version](https://botan.randombit.net/handbook/support.html). Both could be updated on our CI: Xcode (to 26 that comes with Clang 17) and the NDK (from 28 to 29) [[1](https://github.com/randombit/botan/pull/5100), [2](https://github.com/randombit/botan/pull/5125)] which may or may not fix this issue.

Additionally the build configuration for arm32-baremetal (which links to `-lnosys`) fails because the symbol for `__atomic_fetch_sub_4` is missing. I'm guessing that the stop token needs this for synchronization. Not sure whether that would become a showstopper, frankly.

huven:
For macOS/iOS I know for sure it works with recent Xcode, as I already included a basic implementation in my app, works perfectly 😄.  
It looks like the current images use clang 14 and 15 while stop_token was introduced at 16.

Will have some time later today to dive into this.

huven:
Merged both PRs you linked in here, let's see what happens..

reneme:
> Merged both PRs you linked in here, let's see what happens..

Thanks. I was about to propose exactly that.

The macOS 26 PR currently doesn't work -- some Python-based test fails. But the build succeeds, despite the stop token usage. Similarly, NDK r29 seems to ship the stop token as well. So both are probably a green light (if we're willing to ditch support for older NDKs and Xcodes). 🙂

That leaves the linking issue on the "arm32-baremetal" CI job which doesn't seem to provide the required atomic functionality. 🙁

huven:
One python test fails indeed on macos-26. I see the test is disabled on Windows.

I have one Mac here at 26, interestingly enough all tests run ok on it (though an immediate second run fails the cli_tls_socket_tests test (`Error: server bind failed`), but that looks like another issue).

Will debug this a bit further, would be nice if the cli_tls_proxy_tests would fail on my device too.

huven:
Didn't succeed in reproducing the tls_proxy test failure. Installed both boost and python with the exact same version as in GitHub runner, copied the configure.py command verbatim, all tests still pass 😢 

For now, I'm tempted to add macOS to the Windows exception for that test and leave fixing the test to another issue.

```
[...]
   INFO: Ran cli_tls_proxy_tests in 0.41 sec
[...]
Ran 238 tests with 0 failures in 7.82 seconds
```

huven:
The arm32 cross-compile might need a -latomic added to the linker phase, will test that quick&dirty first. If that works it needs to be added somewhere in the configure system.

huven:
After some more experimenting (the -latomic didn't help, will revert that), it seems that the atomic primitives are only available if the cpu supports them, so e.g.

```
arm-none-eabi-c++ -std=c++20 -march=armv6 -O2 -specs=nosys.specs  stop_token_test.cpp
```

works like a charm, as ARMv6 introduced LDREX/STREX. Lowering that

```
arm-none-eabi-c++ -std=c++20 -march=armv5t -O2 -specs=nosys.specs  stop_token_test.cpp
/usr/lib/gcc/arm-none-eabi/12.2.1/../../../arm-none-eabi/bin/ld: /tmp/cc1A7gy2.o: in function `std::stop_token::stop_requested() const':
stop_token_test.cpp:(.text._ZNKSt10stop_token14stop_requestedEv[_ZNKSt10stop_token14stop_requestedEv]+0x14): undefined reference to `__sync_synchronize'
/usr/lib/gcc/arm-none-eabi/12.2.1/../../../arm-none-eabi/bin/ld: /tmp/cc1A7gy2.o: in function `main':
stop_token_test.cpp:(.text.startup+0x30): undefined reference to `__atomic_fetch_add_4'
/usr/lib/gcc/arm-none-eabi/12.2.1/../../../arm-none-eabi/bin/ld: stop_token_test.cpp:(.text.startup+0x70): undefined reference to `__atomic_fetch_sub_4'
/usr/lib/gcc/arm-none-eabi/12.2.1/../../../arm-none-eabi/bin/ld: stop_token_test.cpp:(.text.startup+0x88): undefined reference to `__atomic_fetch_sub_4'
[...]
```

Assuming Botan is used on pre-ARMv6 architectures, that will require some additional work, like providing a custom implementation of some atomic operations, using some kind of locking.  
Or throwing in a lot of #ifdef's in the code, which I prefer to avoid..

huven:
After giving this some thought: I think it's a good idea to decide first if it is safe to assume that ARMv6 is available, or if prior architectures need to be supported too.

- If assuming ARMv6 is ok, the -march=armv6 can be added to CI fixing the build error.
- If ARMv5 and older need to be supported, this feature will need a flag and use #ifdef's in the code.

huven:
To followup my own question: I will assume ARMv6 or higher, so added that to the CI build. Should succeed now (let's see).

And I've added a unit test to test the stop_token with Argon2 (can be extended to other pwdhashes once they support the stop_token).
The test only succeeds if an Invalid_State exception is thrown (i.e. if the derivation was aborted).

src/lib/pbkdf/argon2/argon2.cpp
```diff
@@ -296,6 +297,10 @@ void process_block(secure_vector<uint64_t>& B,
    }
 
    while(index < segments) {
+      if((index & 63) == 0 && stop_token.has_value() && stop_token->stop_requested()) {
+         throw Botan::Invalid_State("Cancelled");
```
reneme: I agree that `Invalid_State` is probably the closest exception type for this. But I would argue that it is worthwhile to introduce a new exception type `Operation_Canceled` that inherits from `Invalid_State` so that applications can explicitly distinguish cancellations.

src/tests/test_pbkdf.cpp
```diff
@@ -245,6 +252,68 @@ class Argon2_KAT_Tests final : public Text_Based_Test {
...
+            return results;
+         }
+
+         result.start_timer();
+
+         // Test will be cancelled after 10ms below. If cancellation fails, test runs for 1s.
+         const auto run_time = std::chrono::milliseconds(1000);
+         const auto tune_time = std::chrono::milliseconds(10);
+         const size_t max_mem = 32;
+         auto pwdhash = fam->tune(32, run_time, max_mem, tune_time);
```
reneme: Given that this is inherently a race condition and judging from previous experience with GitHub Actions, I would suggest to be way more conservative here, unfortunately.

Maybe `run_time = 10s` and `cancellation_time = 500ms`? It's okay if a test failure runs for a long time, because it should practically never happen. But we _really_ want to avoid flaky tests.

src/tests/test_pbkdf.cpp
```diff
@@ -245,6 +252,68 @@ class Argon2_KAT_Tests final : public Text_Based_Test {
...
+            src.request_stop();  // fire the cancellation
+         });
+
+         // Run the derivation on the main thread and evaluate the result
+         try {
+            pwdhash->derive_key(
+               out.data(), out.size(), password.c_str(), password.size(), salt.data(), salt.size(), src.get_token());
+            // If we reach this line, the stop token was ignored
+            result.test_failure("Derivation completed without observing stop token");
+         } catch(const Botan::Invalid_State& e) {
```
reneme: See my other comment, this should then expect `Operation_Canceled`.

reneme (commented):
> To followup my own question: I will assume ARMv6 or higher, so added that to the CI build.

For the sake of this PR, assuming that is certainly fine. I have no way of knowing whether that holds in general though.

Given that we already wrap `std::mutex` and `std::lock_guard`, we could consider similarly wrapping the stop token. If a platform doesn't support it (e.g. because atomics are disabled for bare metal), it could simply fall back to a dummy implementation and ignore the cancellation request.

I'll look into that independently.

src/lib/pbkdf/argon2/argon2.cpp
huven: My reasoning for using Invalid_State was that the caller can easily distinguish the cancelled state, because the caller owns the stop_source.

In my application, I currently call it like this:

```
try {
            passwordHash->derive_key(..., stop_source.get_token());
        } catch (Botan::Exception& e) {
            if (stop_source.stop_requested()) {
                 // handle cancellation
            } else {
                 // handle other error
            }
        }
```

Adding Operation_Cancelled is nice, that simplifies the catch to simply catching Botan:: Operation_Cancelled.

I'll add it in the next commit.

huven: I see I didn't add that logic to the unit test, my bad.. Anyway, will be fixed after adding the new exception.

huven:
> > To followup my own question: I will assume ARMv6 or higher, so added that to the CI build.
> 
> For the sake of this PR, assuming that is certainly fine. I have no way of knowing whether that holds in general though.
> 
> Given that we already wrap `std::mutex` and `std::lock_guard`, we could consider similarly wrapping the stop token. If a platform doesn't support it (e.g. because atomics are disabled for bare metal), it could simply fall back to a dummy implementation and ignore the cancellation request.

Wrapping crossed my mind too. I tried to avoid that as derive_key is public API, having the std:: type there makes the API easier to use, as the caller will know how the argument behaves.

I also considered #ifdef's inside the prototype, wrapping the prototype with ifdef/else, and overloading with ifdef. All of them have some arguments in favor and against (readability, duplication of code/comments).

I ended up assuming ARMv6 is ok🤞

src/lib/pbkdf/argon2/argon2.cpp
huven: Operation_Canceled has been added in commit 0e63412

src/tests/test_pbkdf.cpp
huven: Changed the timing as suggested. Ideally you'd want to test timely cancellation (say within 1 second after requesting cancellation) as that might indicate the stop_token is not checked frequently enough.  
But I fully agree with the approach taken here.

src/lib/pbkdf/argon2/argon2.cpp
huven: In ffi.cpp I mapped OperationCanceled to BOTAN_FFI_ERROR_INVALID_OBJECT_STATE, as the stop_token is not exposed to FFI, so this error cannot occur inside ffi (currently).

If preferred I can add a BOTAN_FFI_ERROR_OPERATION_CANCELED there but that might confuse consumers of the ffi interface.

reneme: I think mapping this in FFI as described is fine. Please just pair up the `case Botan::ErrorType::OperationCanceled:` with the existing `case Botan::ErrorType::InvalidObjectState:` as this seems to be what was done in this switch mapping for other ambiguous error codes.

reneme:
> I ended up assuming ARMv6 is ok🤞

I'm guessing it is. Existing users still have the escape hatch of disabling the modules that use password-based key derivation if they don't need that for their use case. @randombit gets the final say on that, though. :)

src/lib/pbkdf/argon2/argon2.cpp
huven: done

src/tests/test_pbkdf.cpp
huven: Looks like your suggestion to use 10s was no overkill, here is a recent runner output:

```
  Argon2d stop_token cancellation ran 1 tests in 4.00 sec all ok
  Argon2i stop_token cancellation ran 1 tests in 3.63 sec all ok
  Argon2id stop_token cancellation ran 1 tests in 3.46 sec all ok
  Bcrypt-PBKDF stop_token cancellation ran 1 tests in 9.14 sec 1 FAILED
  Failure 1: Derivation completed without observing stop token (at /home/runner/work/botan/botan/source/src/tests/test_pbkdf.cpp:208)
  PBKDF2(SHA-256) stop_token cancellation ran 1 tests in 736.86 msec all ok
  Scrypt stop_token cancellation ran 1 tests in 827.48 msec all ok
```

The first three are already much delayed (the cancellation is supposed to happen after 0.5s), so multi-tasking is definitely not working as it should during testing. The fourth test even managed to complete (this time).  

Will think how/if I can improve that, I already added a private thread_pool, but that didn't help apparently. I'll increase the 10s further if nothing comes up.

huven:
@randombit @reneme I think this PR is ready for review, looking forward to your comments if any further improvements are necessary. The  opening comment contains a Note section with some thoughts and important changes to review.

huven:
Rebased to current master

huven:
Rebased again to current master. Anything needed to get this merged?

PR #5127, opened 2025-10-27
Resolve TODO comments in socket utilities

KaganCanSit:
Hello Botan Developer Team,

This pull request contains some TODO message resolves for socket utilities;
* Replaced `inet_pton` with `string_to_ipv4` for IPv4 validation
   - Since there is a need for an additional parsing functions for IPv6 here, I have not updated ipv_6 for now.
* Implemented `strerror_s` per TODO comment - Windows
   - Uses 94-byte buffer as specified in Microsoft documentation
   - Thread-safe error string conversion
   
I don't except there to be any behavioral difference here. I believe I tested it by compiling with the following command on the Linux but need more additional testing. For Windows I will investigate the situation in CI.

```bash
ninja clean && ninja && python3 src/scripts/test_cli.py ./botan cli_tls
```

Regards.

### DIFF
--- a/src/cli/socket_utils.h
+++ b/src/cli/socket_utils.h
@@ -52,8 +52,16 @@ inline void stop_sockets() {
 }
 
 inline std::string err_to_string(int e) {
-   // TODO use strerror_s here
-   return "Error code " + std::to_string(e);
+   /*
+    * MS documentation specifies 94 character max for user messages.
+    * strerror_s truncates to buffer size - 1 and guarantees null termination.
+    * Using 100 bytes yo ensure sufficient space with safety margin.
+    * https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/strerror-s-strerror-s-wcserror-s-wcserror-s
+    */
+   std::array<char, 100> buf{};
+   const auto res = strerror_s(buf.data(), buf.size() - 1, e);
+   const std::string_view msg = (res == 0) ? buf.data() : "failed to map error with strerror_s()";
+   return Botan::fmt("Error: {} - {}", e, msg);
 }
 
 inline int close(int fd) {
--- a/src/lib/utils/socket/uri.cpp
+++ b/src/lib/utils/socket/uri.cpp
@@ -37,12 +37,7 @@ bool is_domain_name(std::string_view domain) {
 }
 
 bool is_ipv4(std::string_view ip) {
-   std::string ip_str(ip);
-   sockaddr_storage inaddr{};
-
-   // TODO use string_to_ipv4 here
-   // NOLINTNEXTLINE(*-implicit-bool-conversion)
-   return !!inet_pton(AF_INET, ip_str.c_str(), &inaddr);
+   return string_to_ipv4(ip).has_value();
 }
 
 bool is_ipv6(std::string_view ip) {


src/cli/socket_utils.h
```diff
@@ -52,7 +52,14 @@ inline void stop_sockets() {
...
-   // TODO use strerror_s here
+   /*
+    * Buffer size per MS documentation for strerror_s. "Your string message can be, at most, 94 characters long."
+    * https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/strerror-s-strerror-s-wcserror-s-wcserror-s
+    */
+   std::array<char, 94> buf{};
+   if(strerror_s(buf.data(), buf.size(), e) == 0) {
+      return std::string(buf.data());
+   }
    return "Error code " + std::to_string(e);
```
reneme: There are a two (potential) issues here:

* The documentation states that the error message can be 94 characters long. I understand this as "94 characters of error message WITHOUT the trailing `\0` byte". The documentation also states that the error message is truncated and a `\0` byte is always added. So, for the longest error message of 94 characters, we would _loose_ the last character. Also, let's not _depend_ on `strerror_s` actually holding up its promise of adding a `\0` byte. Instead, lets invest into a few extra bytes for `buf`, and ensure an "emergency `\0` byte".
* Let's try to keep the error message as consistent as possible regardless whether `strerror_s` fails or not.


```suggestion
   /*
    * Buffer size per MS documentation for strerror_s. "Your string message can be, at most, 94 characters long."
    * https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/strerror-s-strerror-s-wcserror-s-wcserror-s
    */
   const auto msg = [&]() -> std::string {
      std::array<char, 100> buf = {0};
      if(strerror_s(buf.data(), buf.size() - 1 /* pedantic \0 byte */, e) == 0) {
         return std::string(buf.data());
      } else {
         return "failed to map error with strerror_s()";
      }
   }();
   return "Error code " + std::to_string(e) + " - " + msg;
```

reneme (changes requested):
Some things regarding C-string handling, otherwise LGTM. Thank you!

src/cli/socket_utils.h
KaganCanSit: I thought a null terminator had been added. I overlooked it. I'll fix it later today.

We can also ensure full security by initializing the string constructor with the size parameter. 🤔

reneme: > We can also ensure full security by initializing the string constructor with the size parameter.

Probably not, because if you use the `std::string` constructor that takes a char pointer and a length will always copy the full buffer length. And you might have bytes of garbage at the end.

> I thought a null terminator had been added. I overlooked it. I'll fix it later today.

Yes, Microsoft promises that it will always add a null terminator. So, technically, your code is correct. I'm only suggesting to not trust the documentation and add our own null terminator additionally.

KaganCanSit: I've implemented the change here. Thank you for your guidance.

I'm curious about something. Why did you decide to use lambda instead of the `?` operator or an additional `std::string` variable?

reneme: Good question. I wanted to make sure that `msg` can be declared as `const`. These days, the IIFE pattern is basically my go-to thing to initialize a `const` variable if the initialization is non-trivial.

But you are right, a ternary operator is easier to read here and might even save an allocation when paired with `Botan::fmt`. Like so:

```C++
inline std::string err_to_string(int e) {
   /*
    * MS documentation specifies 94 character max for user messages.
    * strerror_s truncates to buffer size - 1 and guarantees null termination.
    * Using 100 bytes to ensure sufficient space with safety margin.
    * https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/strerror-s-strerror-s-wcserror-s-wcserror-s
    */
    std::array<char, 100> buf = {0};
    const auto res = strerror_s(buf.data(), buf.size() - 1, e);
    const std::string_view msg = (res == 0) ? buf.data() : "failed to map error with strerror_s()";
    return Botan::fmt("Error: {} - {}", e, msg);
}
```

(I'm hoping `Botan::fmt` works with `std::string_view`. [`std::format` certainly does](https://godbolt.org/z/W5P4nzqfW).)

KaganCanSit:
Errors are appearing in Windows tests, but I don't think this is related to changes in this branch.

~I occasionally see the message “Failure 1: Serpent producer 'avx512' unexpected result for encrypt” in the ‘Nightly’ build and other branches. I can retrigger CI with the “amend” comment without making any changes, or merge if appropriate.~
Update: For this case, commit number 2cb60851ce7fb3010dacbbf3c2a91bfd6d7c4d1a was included. I rebased branch.

PR #5133, opened 2025-10-29
Force the inlining of the Keccak permutation round function

randombit:
Clang really doesn't want to inline this, but it should. Improves performance of SHA-3 and SHAKE-128 by about 40% on a Tiger Lake system with Clang 21.


randombit:
@reneme I'd be interested if this makes any difference for XCode on Apple hw

reneme (approved):
> I'd be interested if this makes any difference for XCode on Apple hw

On my M2 MacBook Air with XCode 16 I see gains of about 20%.

PR #5140, opened 2025-11-02
Preserve file metadata when installing

bkmgit:
See:
- https://docs.python.org/3/library/shutil.html#shutil.copyfile
- https://docs.python.org/3/library/shutil.html#shutil.copy2 for explantions


randombit:
Could use some context here, why would this be useful?

bkmgit:
Useful for Fedora GNU/Linux packaging:
https://docs.fedoraproject.org/en-US/packaging-guidelines/#_timestamps
allows one to determine when file was last changed.

PR #5135, opened 2025-10-30
FIX: SRP6 step1 fails after FFI reported insufficient buffer space

reneme:
In short: The first invocation performed the state transition, then couldn't hand out the value due to insufficient space. The second invocation saw the non-pristine state and reported "invalid object".

Closes #5112.

PR #5141, opened 2025-11-02
Change FFI SRP6 to use fixed length outputs for integer values

randombit:
The test added in #5135 (quite reasonably) assumed that the SRP6 outputs are always exactly the same size as the group element, which would occasionally (~ 1/256 of the time) cause a failure if the integer had a leading zero byte.

Change the FFI layer to zero pad as required so that these short outputs cannot occur. This is anyway a more consistent and less surprising behavior.

### DIFF
--- a/src/lib/ffi/ffi_srp6.cpp
+++ b/src/lib/ffi/ffi_srp6.cpp
@@ -49,7 +49,7 @@ int botan_srp6_group_size(const char* group_id, size_t* group_p_bytes) {
    }
 
    return ffi_guard_thunk(__func__, [=]() -> int {
-      auto group = Botan::DL_Group::from_name(group_id);
+      const auto group = Botan::DL_Group::from_name(group_id);
       *group_p_bytes = group.p_bytes();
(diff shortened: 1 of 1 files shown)


src/lib/ffi/ffi_srp6.cpp
```diff
@@ -131,8 +131,10 @@ int botan_srp6_generate_verifier(const char* username,
       }
       try {
          std::vector<uint8_t> salt_vec(salt, salt + salt_len);
-         auto verifier_bn = Botan::srp6_generate_verifier(username, password, salt_vec, group_id, hash_id);
-         return write_vec_output(verifier, verifier_len, verifier_bn.serialize());
+         auto group = Botan::DL_Group::from_name(group_id);
```
reneme: Super-nit:


```suggestion
         const auto group = Botan::DL_Group::from_name(group_id);
```

reneme (approved):
Thanks! I just encountered that as well: https://github.com/randombit/botan/actions/runs/19062404545/job/54444887338?pr=5143 🫨

Issue #5136, opened 2025-10-31
Certificate_Store_In_Memory, additional constructor?


bazineta:
`Certificate_Store_in_Memory` has a convenience constructor:
```
      /**
      * Adds given certificate to the store.
      */
      explicit Certificate_Store_In_Memory(const X509_Certificate& cert);
```
I think this has the effect of signaling to the consumer that a CRL is optional, but if one's taking the policy defaults, it's not, and someone porting from ASIO's TLS can spend a lot of time in `error 42` purgatory before realizing that with the default policies, it's not optional -- it can be empty, but it's not optional.

It might be helpful to have a second convenience constructor:
```
      /**
      * Adds given certificate and certificate revocation list (CRL) to the store.
      */
      explicit Certificate_Store_In_Memory(const X509_Certificate& cert, const X509_CRL& crl);
```
And modifying the comment in the first constructor to note that with default policies, the second constructor is probably the one you want. I think that as things stand now, the presence of the constructor that only takes a certificate, when combined with other documentation that speaks to the optionality of CRLs, is a bit misleading given how the policies work by default.

bazineta:
So, here's the thing. I was looking through the TODOs at the bottom of the `Certificate_Store_In_Memory` class:

https://github.com/randombit/botan/blob/f7975f0dbbf71fd670141f5e02e2818566ec0789/src/lib/x509/certstor.h#L156

And I was thinking, well, the library does conditionally compile with boost, and there's multi-index containers in boost, it'd be simple to handle that with a multi-index container, could just check for boost being present, have an alternate version that'd solve those TODOs....

And then the light dawned that:

1. my use case is exactly one certificate and exactly one CRL
2. the TODOs were indicative that the class is more a demo of how things could be done, not necessarily how they had to be done
3. since I'm never storing more than one certificate and one CRL, I'd be better off just deriving my own class from `Certificate_Store`

That did turn out to be the short path, so probably best that I close this, but figured documenting the thought process was perhaps worthwhile.

bazineta:
@randombit Jack, my single certificate / single CRL implementation has turned out well. I suspect this might be a reasonably common use case; we use it for private client / server TLS communications, where we generate a private CA and then issue certificates and keys from it, allowing our own clients and servers to communicate securely with one another.

Is this something you might be interested in a PR for? I can provide a gist if that'd be helpful to determine if it'd be more broadly useful, or too narrow in scope to want to consider for inclusion.

reneme:
[...]

PR #5143, opened 2025-11-04
Add another `Certificate_Store_In_Memory` convenience constructor

reneme:
This aims at easing new users' first steps with the ASIO TLS wrapper. Typically, the first issue that new users bump into when trying the example code is that Botan requires certificate revocation info by default. I.e. if you just compile our `tls_stream_client.cpp` example and run it against any well-known webserver (say "google.com"), you'll end up with the fairly generic error "bad_certificate". I added [a warning to the documentation about this already a year ago](https://github.com/randombit/botan/pull/3799), but having it presented in runnable code is probably even more important.

Also, as suggested in #5136, this adds another constructor to `Certificate_Store_In_Memory` that takes a trusted certificate along with a trusted CRL.

### DIFF
--- a/src/lib/x509/certstor.cpp
+++ b/src/lib/x509/certstor.cpp
@@ -100,8 +100,7 @@ std::optional<X509_Certificate> Certificate_Store_In_Memory::find_cert_by_pubkey
       throw Invalid_Argument("Certificate_Store_In_Memory::find_cert_by_pubkey_sha1 invalid hash");
    }
 
-   auto hash = HashFunction::create("SHA-1");
-   BOTAN_ASSERT_NONNULL(hash);
+   auto hash = HashFunction::create_or_throw("SHA-1");
 
    for(const auto& cert : m_certs) {
       hash->update(cert.subject_public_key_bitstring());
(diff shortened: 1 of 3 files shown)


src/lib/x509/certstor.cpp
```diff
@@ -100,6 +101,7 @@ std::optional<X509_Certificate> Certificate_Store_In_Memory::find_cert_by_pubkey
    }
 
    auto hash = HashFunction::create("SHA-1");
+   BOTAN_ASSERT_NONNULL(hash);
```
randombit: IMO either this should be  `create_and_throw`, or else add a comment here to the effect of
> The sha1 module is a hard dependency of x509 so the hash should always be available
[Fine with either]

src/lib/x509/certstor.cpp
```diff
@@ -118,6 +120,7 @@ std::optional<X509_Certificate> Certificate_Store_In_Memory::find_cert_by_raw_su
    }
 
    auto hash = HashFunction::create("SHA-256");
+   BOTAN_ASSERT_NONNULL(hash);
```
randombit: Same as prior comment

src/lib/x509/certstor.cpp
reneme: I had the same train of thought, given that the hashes are hard dependencies. Anyway, changed to `create_or_throw` as this is what is done usually across the code base.

PR #5144, opened 2025-11-04
Added missing dependency in cpuid_aarch64

huven:
Title says it all..

### DIFF
--- a/src/lib/utils/cpuid/cpuid_aarch64/info.txt
+++ b/src/lib/utils/cpuid/cpuid_aarch64/info.txt
@@ -9,3 +9,7 @@ name -> "CPUID for Aarch64"
 <arch>
 arm64
 </arch>
+
+<requires>
+os_utils
+</requires>


randombit:
I think this is not quite what we want since for the macos case we don't need anything from `os_utils`. #5145 

OTOH while #5145 fixes the issue in the more idiomatic way (at least in terms of how module dependencies are usually handled) it is going to be potentially surprising that not enabling `os_utils` causes CPUID detection to fail on the many architectures that require invoking an OS function to detect CPU features (Aarch64, ARM, LoongArch and RISCV among others).

Not really sure how best to approach this, but definitely I'd want the various architectures involved to all behave in roughly the same way. So I think for now #5145 is the way to go since it maintains that consistency, but open to changing the overall approach to such things.

huven:
> I think this is not quite what we want since for the macos case we don't need anything from `os_utils`. #5145
> 
> OTOH while #5145 fixes the issue in the more idiomatic way (at least in terms of how module dependencies are usually handled) it is going to be potentially surprising that not enabling `os_utils` causes CPUID detection to fail on the many architectures that require invoking an OS function to detect CPU features (Aarch64, ARM, LoongArch and RISCV among others).
> 
> Not really sure how best to approach this, but definitely I'd want the various architectures involved to all behave in roughly the same way. So I think for now #5145 is the way to go since it maintains that consistency, but open to changing the overall approach to such things.

In that case it might be better to leave it as it is, the build error is very clear, much easier to debug than a CPU test failing silently.

In my case, adding shacal2_armv8 module triggered the issue (which implicitly added cpuid). Without the CPU feature test, the non-asm version will be used, which would have required considerably more effort to debug.

huven:
or... move run_cpu_instruction_probe out of os_utils into the cpuid module, as that is the only consumer, and that is likely to stay that way as cpuid should be used to get to the supported features.

randombit:
BTW if you are running on Linux you definitely want `os_utils` to get `getauxval` support, since that is much safer than the SIGILL based probing.

> move run_cpu_instruction_probe out of os_utils into the cpuid module, as that is the only consumer, and that is likely to stay that way as cpuid should be used to get to the supported features.

Maybe so will have to think about it

huven:
> BTW if you are running on Linux you definitely want `os_utils` to get `getauxval` support, since that is much safer than the SIGILL based probing.

I'm on macOS/iOS.

huven:
I see #5145 has already been merged, so I'll close this one..

PR #5150, opened 2025-11-05
Unroll loops within the generic Montgomery reduction

randombit:
For GCC 15 this improves large DH and RSA operations by about 20%. For Clang improvement is minor, under 5%.

I need to confirm the improvement across a sample of machines before merging, opening now as draft for CI


randombit:
Summary so far

* Faster on x86-64
* Faster on Aarch64
* Slower on POWER

randombit:
Win on common platforms vs regression on rare platform seems worth it, if needed we can conditionally choose which implementation to use.

PR #5153, opened 2025-11-09
Add std::array support to BigInt::serialize function

KaganCanSit:
This PR resolves the TODO comment in `BigInt::serialize()` by adding support for `std::array<uint8_t, N>` alongside existing `std::vector` and `secure_vector` support.

TODO Message:
```cpp
// TODO this supports std::vector and secure_vector
// it would be nice if this also could work with std::array as in
// bn.serialize_to<std::array<uint8_t, 32>>(32);
```

### DIFF
--- a/src/lib/math/bigint/bigint.h
+++ b/src/lib/math/bigint/bigint.h
@@ -9,6 +9,7 @@
 #ifndef BOTAN_BIGINT_H_
 #define BOTAN_BIGINT_H_
 
+#include <botan/concepts.h>
 #include <botan/secmem.h>
 #include <botan/types.h>
 #include <iosfwd>
@@ -743,11 +744,8 @@ class BOTAN_PUBLIC_API(2, 0) BigInt final {
        * Throws if the BigInt is too large to encode in the length
        * specified.
        */
-      template <typename T = std::vector<uint8_t>>
+      template <concepts::resizable_byte_buffer T = std::vector<uint8_t>>
       T serialize(size_t len) const {
-         // TODO this supports std::vector and secure_vector
-         // it would be nice if this also could work with std::array as in
-         //   bn.serialize_to<std::array<uint8_t, 32>>(32);
          T out(len);
          this->serialize_to(out);
          return out;
@@ -756,9 +754,25 @@ class BOTAN_PUBLIC_API(2, 0) BigInt final {
       /**
        * Serialize the value of this BigInt as a big endian encoding.
        */
-      template <typename T = std::vector<uint8_t>>
+      template <concepts::resizable_byte_buffer T = std::vector<uint8_t>>
       T serialize() const {
          return serialize<T>(this->bytes());
+      }
+
+      /**
+       * Serialize the BigInt into a fixed-size std::array.
+       * 
+       * This is a convenience method for serialization into compile-time fixed
+       * size arrays.
+       * 
+       * @tparam len the size of the output array in bytes
+       * @return std::array containing the serialized BigInt (big-endian)
+       */
+      template <size_t len>
+      std::array<uint8_t, len> serialize_to_array() const {
+         std::array<uint8_t, len> out{};
+         serialize_to(out);
+         return out;
       }
 
       /**
--- a/src/tests/test_bigint.cpp
+++ b/src/tests/test_bigint.cpp
@@ -991,6 +991,48 @@ std::vector<Test::Result> test_bigint_serialization() {
                b.binary_encode(enc5.data(), enc5.size());
                res.test_bin_eq("BigInt::binary_encode", enc5, "000000000000000000000000FEDCBA9876543210BAADC0FFEE");
             }),
+
+      CHECK("BigInt std::array serialization",
+            [](Test::Result& res) {
+               const Botan::BigInt testData(0x1234567890ABCDEF);
+
+               const auto arr32 = testData.serialize_to_array<32>();
+               res.test_sz_eq("BigInt::serialize std::array size", arr32.size(), size_t{32});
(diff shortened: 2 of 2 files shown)


reneme (changes requested):
Nice, thanks!

Interesting way to detect an array! However, IIUC this would work for other tuple-types as well, no? Like `std::pair<int, std::string>`. That would produce very strange compiler errors for sure... But then again, the API does also allow passing a `std::vector<uint32_t>`, which is just as awful I guess. You could look into using the `statically_spanable_range<>` concept to detect the array instead.

Anyway: Could you please add a few tests that serialize a `BigInt` into an array? `src/tests/test_bigint.cpp` has a function `test_bigint_serialization`, please add another `CHECK` that uses `std::array`. Also check that the exception is thrown as expected if the array is too short or too long.

KaganCanSit:
> Nice, thanks!
> 
> Interesting way to detect an array! However, IIUC this would work for other tuple-types as well, no? Like `std::pair<int, std::string>`. That would produce very strange compiler errors for sure... But then again, the API does also allow passing a `std::vector<uint32_t>`, which is just as awful I guess. You could look into using the `statically_spanable_range<>` concept to detect the array instead.
> 
> Anyway: Could you please add a few tests that serialize a `BigInt` into an array? `src/tests/test_bigint.cpp` has a function `test_bigint_serialization`, please add another `CHECK` that uses `std::array`. Also check that the exception is thrown as expected if the array is too short or too long.

You're right that it will provide general support for tuple types. I've made the changes as you suggested.

Additionally, for the situation you mentioned, `the API does also allow passing a std::vector<uint32_t>, which is just as awful I guess.` I think we could use `requires std::same_as<typename T::value_type, uint8_t`. There were no issues during compilation and testing. 

However, since this is a general change to the API, I haven't added it to the branch. What do you think?


Update: 20.01.2026 - Rebase master branch and remove `static_cast` usage in `big_int` test code.

src/lib/math/bigint/bigint.h
reneme: Looking into this again, I think what we actually want here, is an overload of this method specifically for arrays. See for example `RandomNumberGenerator::random_array`. Because (1) this wouldn't require a repeat of the desired length (once for the static array size and once for the `size_t len` parameter above), and (2) it constrains the output type to `std::array`, no other user-defined types allowed (hence, no room for surprises or undefined behavior).

My proposal: create a new method like so:

```C++
template <size_t bytes>
std::array<uint8_t, bytes> serialize_to_array() const {
   std::array<uint8_t, bytes> out;
   serialize_to(out);
   return out;
}
```

... by the way: there's also no need to check if the passed array size is large enough, because `serialize_to` does that anyway.

src/lib/math/bigint/bigint.h
```diff
@@ -709,10 +710,15 @@ class BOTAN_PUBLIC_API(2, 0) BigInt final {
...
+               BOTAN_ARG_CHECK(len == std::tuple_size_v<T>, "BigInt::serialize array size must match requested length");
+               return T{};  // Fixed-size containers (std::array)
+            } else {
+               return T(len);  // Dynamic containers (std::vector, secure_vector)
+            }
+         }();
+
          this->serialize_to(out);
          return out;
       }
```
reneme: Instead of making this more generic, we should restrict it to the `resizable_byte_buffer` concept. I.e.

```C++
      template <concepts::resizable_byte_buffer T = std::vector<uint8_t>>
      T serialize(size_t len) const {
         T out(len);
         this->serialize_to(out);
         return out;
      }
```

The `resizable_byte_buffer` concept enforces that: (1) the type is a contiguous container, (2) it can be initialized with a given length at runtime, and (3) it holds `uint8_t`. For instance `std::vector<uint8_t>` and `Botan::secure_vector<uint8_t>` fulfill these requirements. But `std::array<uint8_t, ...>` does not because of rule 2.

src/tests/test_bigint.cpp
```diff
@@ -983,6 +984,52 @@ std::vector<Test::Result> test_bigint_serialization() {
                            enc5,
                            Botan::hex_decode("000000000000000000000000FEDCBA9876543210BAADC0FFEE"));
             }),
+
+      CHECK("BigInt std::array serialization",
+            [](Test::Result& res) {
+               const Botan::BigInt testData(0x1234567890ABCDEF);
+
+               auto arr32 = testData.serialize<std::array<uint8_t, 32>>(32);
```
reneme: Obviously the occurrences below will need to be adapted as well with the suggestions above. But just for instance, this should become:

```suggestion
               const auto arr32 = testData.serialize_to_array<32>();
```

... also note the added `const` which is being enforced by clang-tidy in the meantime, I believe.

reneme (changes requested):
Sorry for changing my mind on this, but I think there's a better strategy to implement this TODO.

KaganCanSit:
> Sorry for changing my mind on this, but I think there's a better strategy to implement this TODO.

Thanks for the detailed feedback and this learning opportunity.

I've pushed the changes but since I worked on this after hours, I'd like to revisit and review them more thoroughly when I'm fresher (I'm not yet comfortable using concepts actively). I want to make sure I fully understand the reasoning behind each change. I will write experimental code blocks for exploration and possibly add more tests later.

But absolutily like this way, more readable and effective.

reneme (approved):
LGTM. @randombit I'm guessing you agree with adding a `BigInt::serialize_to_array<32>()` given that you left the TODO?

KaganCanSit:
Updates:
24.02.2026: Hit #5376, #5328, #5334, #5341... Rebased master branch.
24.05.2026 - Rebase master branch.

PR #5156, opened 2025-11-14
Increased max HMAC key length to 8192

huven:
This change is needed to be able to implement ProxyCrypt keyfile support, where a max of 4096 bytes are concatenated to a password (max 500 bytes) and then used as hmac key.

### DIFF
--- a/src/scripts/ci/setup_gh_actions.sh
+++ b/src/scripts/ci/setup_gh_actions.sh
@@ -31,7 +31,7 @@ fi
 
 if type -p "apt-get"; then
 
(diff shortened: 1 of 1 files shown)


huven:
The failed checks look like some issue in the Linux GitHub runner setup, not related to this PR.

Added -f flag to suppress the error

> rm -f /var/lib/man-db/auto-update

randombit:
4096+512 is kind of a non-obvious derivation, if larger is required lets just do 8192 for this

huven:
8192 it is 😄

randombit (approved):

PR #5157, opened 2025-11-17
fix: fix bsi policy to allow use of argon2

xemwebe:
Hash method blake2 is prohibited by BSI policy, but required by argon2, which is recommended by BSI to be used for key derivation. This is a backport of fix in botan3

### DIFF
--- a/src/build-data/policy/bsi.txt
+++ b/src/build-data/policy/bsi.txt
@@ -155,7 +155,7 @@ emsa_raw
 emsa_x931
 
 # hash
-blake2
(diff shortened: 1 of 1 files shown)


randombit:
CI failure is unrelated.

Fine to merge but keep in mind that no further releases of Botan2 are expected - absent a truly critical bug (eg memory error) being discovered in Botan2, the last release will be the existing 2.19.5. And Botan3 has many optimizations and security improvements that may be relevant to your use case.

PR #5168, opened 2025-11-22
Add clang-format 17 formatting rules for attributes and inline assembly

KaganCanSit:
Hi,

This PR adds two new clang-format 17 options to the `.clang-format` configuration, as specified in #4049.

I initially prepared all the requested changes from #4049 in this [branch](https://github.com/KaganCanSit/botan/commits/update-clang-format-rules/). However, considering the broad impact across the codebase, I believe it would be better for you to handle those changes directly, as you have a better understanding of the project's current state and appropriate timing.

Therefore, this PR focuses solely on adding the two new clang-format 17 flags and applying the necessary formatting adjustments.

**Changes:**
- Added `BreakAfterAttributes: Always`
- Added `BreakBeforeInlineAsmColon: OnlyMultiLine`

**Command used:**
```bash
```

**Clang-format version:**
```bash
Debian clang-format version 17.0.6 (22+b2)
```

**Note:** Currently, `BreakBeforeInlineAsmColon = OnlyMultiLine` has no effect on any files in the codebase.

I believe this approach minimizes the review scope while still making incremental progress on the formatting improvements outlined in #4049.

Best regards.

### DIFF
--- a/src/bogo_shim/bogo_shim.cpp
+++ b/src/bogo_shim/bogo_shim.cpp
@@ -85,7 +85,8 @@ void shim_log(std::string_view s) {
    }
 }
 
-[[noreturn]] void shim_exit_with_error(const std::string& s, int rc = 1) noexcept {
+[[noreturn]]
+void shim_exit_with_error(const std::string& s, int rc = 1) noexcept {
    shim_log("Exiting with " + s);
    std::cerr << s << "\n";
    std::exit(rc);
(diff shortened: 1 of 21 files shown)


KaganCanSit:
Hmm, the static analysis failed because I shifted the suppression lines for clang-tidy. I can make adjustments to those lines. But I'm not sure what you'd prefer revert this or continue. So I'm looking forward to hearing from you.


Why same functions have a third line? Interesting. I expected it to be more like two lines.
```cpp
[[nodiscard]]
detail::SessionHandle operator[](size_t i) const noexcept {
```

Problem is `tpm2_util.h` and `stl_util.h` files. 
```cpp
```

KaganCanSit:
[As I mentioned in the comment on the issue](https://github.com/randombit/botan/issues/4049#issuecomment-3025217364), it seems healthier for this change not to be made by me.

I closed now.

PR #5172, opened 2025-12-10
Fix and enable clang-tidy warning misc-const-correctness

randombit:


src/lib/pubkey/hss_lms/hss.cpp
```diff
@@ -147,13 +147,14 @@ std::shared_ptr<HSS_LMS_PrivateKeyInternal> HSS_LMS_PrivateKeyInternal::from_byt
...
       params.push_back({LMS_Params::create_or_throw(lms_type), LMOTS_Params::create_or_throw(lmots_type)});
    }
-   std::string hash_name = params.at(0).lms_params().hash_name();
-   if(std::any_of(params.begin(), params.end(), [&hash_name](HSS_LMS_Params::LMS_LMOTS_Params_Pair& lms_lmots_params) {
-         bool invalid_lmots_hash = lms_lmots_params.lmots_params().hash_name() != hash_name;
-         bool invalid_lms_hash = lms_lmots_params.lms_params().hash_name() != hash_name;
-         return invalid_lmots_hash || invalid_lms_hash;
-      })) {
-      throw Decoding_Error("Inconsistent hash functions are not allowed.");
+   const std::string hash_name = params.at(0).lms_params().hash_name();
```
reneme: `.hash_name()` returns a `const&`, so perhaps:


```suggestion
   const auto& hash_name = params.at(0).lms_params().hash_name();
```

randombit:
> And I wish const was the default in C++.

I also. At the very least, and more practical I think, I wish you could mark variables as `mutable` to explicitly call out the possible modification, plus a `clang-tidy` rule to enforce that each variable is annotated as either `const` or `mutable`. AFAICT that could be done without breaking existing code as long as the default remains mutable.

reneme (approved):

PR #5180, opened 2025-12-16
Avoid integer modulo operations when sieving for primes

randombit:
Add a CT value barrier in the function computing the Montgomery inverse; since the value of bi is just a single bit Clang sometimes decides to use jumps to skip the various computations.

Add poison annotations to the RSA key generation function.

### DIFF
--- a/src/lib/math/numbertheory/make_prm.cpp
+++ b/src/lib/math/numbertheory/make_prm.cpp
@@ -12,6 +12,7 @@
 #include <botan/internal/barrett.h>
 #include <botan/internal/bit_ops.h>
 #include <botan/internal/ct_utils.h>
+#include <botan/internal/divide.h>
 #include <botan/internal/loadstor.h>
 #include <algorithm>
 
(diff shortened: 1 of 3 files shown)


src/lib/math/numbertheory/make_prm.cpp
```diff
@@ -233,6 +248,8 @@ BigInt generate_rsa_prime(RandomNumberGenerator& keygen_rng,
    while(true) {
       BigInt p(keygen_rng, bits);
 
+      CT::poison(p);
```
reneme: Maybe `scoped_poison()`? The only caveat I see is that this would unpoison `p` after each loop iteration but that's probably a good idea anyways, no?

src/lib/utils/bit_ops.h
```diff
@@ -41,9 +41,7 @@ BOTAN_FORCE_INLINE constexpr T ct_is_zero(T x) {
 */
 template <std::unsigned_integral T>
 BOTAN_FORCE_INLINE constexpr size_t ct_if_is_zero_ret(T x, size_t s) {
-   const T a = ~x & (x - 1);
-   const size_t mask = static_cast<size_t>(0) - static_cast<size_t>(a >> (sizeof(T) * 8 - 1));
-   return mask & s;
+   return ct_is_zero(x) & s;
 }
```
reneme, line 54: Doesn't that cause trouble if `T` is shorter than `size_t`? Then `ct_is_zero()` returns a 1-mask that is shorter than `size_t` which would truncate the MSBs of `s`.

randombit: Yeah this is wrong I was just trying to see if I could get clang to not emit a jump here, but [alas no](https://github.com/randombit/botan/actions/runs/20283982870/job/58253097583#step:5:1005) so I think I need to hoist the value barrier into a new header so it can be used here as well as in ct_utils.h

randombit: #5181

PR #5184, opened 2025-12-20
Use gcc-14 for valgrind CT checks

randombit:
For some unclear reason the CT checks started failing after #5180 but only with GCC. I cannot replicate locally with GCC 14 or 15. The test output has every appearance that the scoped_poison is not unpoisoning the prime before returning, causing various derived values, including the public modulus, to be considered poisoned as well.

Switch the nightly checks to using GCC 14


randombit:
@reneme Any ideas why GCC 13 is ignoring the unpoison when scoped_poison goes out of scope?

PR #5187, opened 2025-12-22
Add AVX-512 implementation of the Keccak permutation

randombit:
This is not yet particularly well optimized but is already 20-25% faster than the BMI2 version on a Tiger Lake processor.


randombit:
#5191

PR #5188, opened 2025-12-22
FFI: Generic getter(s) for X.509 objects

reneme:
We're currently working on extending [the strongswan botan plugin](https://github.com/strongswan/strongswan/tree/f79504994ae210904f5517abe195cccfa44843ba/src/libstrongswan/plugins/botan) to parse and verify X.509 certificates and CRLs using Botan. That use case requires adding a number of new getters to fetch detailed information from certificates and CRLs, many of which are duplicated for both types like the raw signature bits, the serial number or the (optional) authority key ID.

Instead of extending the FFI with several new functions (including the duplication for `botan_x509_cert_t` and `botan_x509_crl_t`) we would like to propose a more generic API that can work with both types and provide access to a number of (shared) data fields, like so:

```C++

int copy_to_buffer(void*, uint8_t* bin, size_t); // some binary view function
int copy_to_string(void*, uint8_t* bin, size_t); // some string view function

botan_x509_cert_t cert = /* ... */;
botan_x509_crl_t crl   = /* ... */;

uint8_t cert_akid[32];
uint8_t crl_akid[32];
size_t outlen = 32;
botan_x509_cert_get_binary_values(cert, BOTAN_X509_AUTHORITY_KEY_ID, 0, cert_akid, copy_to_buffer);
botan_x509_crl_get_binary_values(crl , BOTAN_X509_AUTHORITY_KEY_ID, 0, crl_akid, copy_to_buffer);

char uri[128];
int rc = BOTAN_FFI_SUCCESS;
for (size_t i = 0; rc != BOTAN_FFI_ERROR_OUT_OF_RANGE; ++i) {
   rc = botan_x509_cert_get_string_values(cert, BOTAN_X509_OCSP_RESPONDER_URLS, i, uri, copy_to_string);
}
/* ... use the AKIDs somehow */
```

Our hope is that this significantly reduces the boilerplate and API surface while also being easily extensible by amending the "type enum" in an ABI-compatible way. Note that the proposed API allows to access multi-value fields using a simple index parameter.

Currently, the following data fields can be retrieved via this API:

| Data field | Data Type | Certificate | CRL |
| --- |:---:|:---:|:---:|
| TBS data | binary | x | x |
| Signature Bits | binary | x | x |
| Encoded Signature Scheme | binary | x | x |
| Encoded Subject DN | binary | x | | 
| Encoded Issuer DN | binary | x | x |
| Subject Key ID | binary | (x) | |
| Authority Key ID | binary | (x) | (x) |
| Serial Number | binary | x | x |
| Public Key w/ Algorithm Identifier | binary | x | |
| DER Encoding | binary | x | x |
| PEM Encoding | string | x | x |
| CRL distribution points | multi-value string | (x) | |
| OCSP responders | multi-value string | (x) | |
| CA issuers URLs | multi-value string | (x) | |

### DIFF
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -2284,6 +2284,8 @@ BOTAN_FFI_EXPORT(2, 8) int botan_x509_cert_dup(botan_x509_cert_t* new_cert, bota
 BOTAN_FFI_EXPORT(3, 11)
 int botan_x509_cert_view_binary_values(
    botan_x509_cert_t cert, botan_x509_value_type value_type, size_t index, botan_view_ctx ctx, botan_view_bin_fn view);
+BOTAN_FFI_EXPORT(3, 11)
+int botan_x509_cert_view_binary_values_count(botan_x509_cert_t cert, botan_x509_value_type value_type, size_t* count);
 
 /**
  * Retrieve a specific string value from an X.509 certificate.
@@ -2298,6 +2300,8 @@ int botan_x509_cert_view_binary_values(
 BOTAN_FFI_EXPORT(3, 11)
 int botan_x509_cert_view_string_values(
    botan_x509_cert_t cert, botan_x509_value_type value_type, size_t index, botan_view_ctx ctx, botan_view_str_fn view);
+BOTAN_FFI_EXPORT(3, 11)
+int botan_x509_cert_view_string_values_count(botan_x509_cert_t cert, botan_x509_value_type value_type, size_t* count);
 
 /* Prefer botan_x509_cert_not_before and botan_x509_cert_not_after */
 BOTAN_FFI_EXPORT(2, 0) int botan_x509_cert_get_time_starts(botan_x509_cert_t cert, char out[], size_t* out_len);
@@ -2564,6 +2568,8 @@ int botan_x509_crl_view_binary_values(botan_x509_crl_t crl_obj,
                                       size_t index,
                                       botan_view_ctx ctx,
                                       botan_view_bin_fn view);
+BOTAN_FFI_EXPORT(3, 11)
+int botan_x509_crl_view_binary_values_count(botan_x509_crl_t crl_obj, botan_x509_value_type value_type, size_t* count);
 
 /**
  * Retrieve a specific string value from an X.509 certificate revocation list.
@@ -2581,6 +2587,9 @@ int botan_x509_crl_view_string_values(botan_x509_crl_t crl_obj,
                                       size_t index,
                                       botan_view_ctx ctx,
                                       botan_view_str_fn view);
+BOTAN_FFI_EXPORT(3, 11)
+int botan_x509_crl_view_string_values_count(botan_x509_crl_t crl_obj, botan_x509_value_type value_type, size_t* count);
+
 /**
  * Given a CRL and a certificate,
  * check if the certificate is revoked on that particular CRL
--- a/src/lib/ffi/ffi_cert.cpp
+++ b/src/lib/ffi/ffi_cert.cpp
@@ -110,6 +110,38 @@ std::optional<botan_x509_general_name_types> to_botan_x509_general_name_types(Bo
    BOTAN_ASSERT_UNREACHABLE();
 }
 
+/**
+ * Given some enumerator-style function @p fn, count how many values it can
+ * produce before returning BOTAN_FFI_ERROR_OUT_OF_RANGE. If the first call to
+ * @p fn returns BOTAN_FFI_ERROR_NO_VALUE, zero is written to @p count.
+ *
+ * If this function returns BOTAN_FFI_SUCCESS, @p count contains the number of
+ * values that can be enumerated. Otherwise, the value of @p count is undefined.
+ */
+template <std::invocable<size_t> EnumeratorT>
+int enumerator_count_values(size_t* count, EnumeratorT fn) {
+   if(Botan::any_null_pointers(count)) {
+      return BOTAN_FFI_ERROR_NULL_POINTER;
+   }
+
+   *count = 0;
+   for(;; ++(*count)) {
+      const auto rc = fn(*count);
+      switch(rc) {
+         case BOTAN_FFI_ERROR_NO_VALUE:
+         case BOTAN_FFI_ERROR_OUT_OF_RANGE:
+            // hit the end of the enumeration
+            return BOTAN_FFI_SUCCESS;
+         case BOTAN_FFI_SUCCESS:
+            // got a value, continue counting
+            break;
+         default:
+            // unexpected error from enumerator function
+            return rc;
+      }
+   }
+}
+
 }  // namespace
 
 }  // namespace Botan_FFI
@@ -186,8 +218,14 @@ namespace {
 
 int botan_x509_object_view_value(const Botan::X509_Object& object,
                                  botan_x509_value_type value_type,
+                                 size_t index,
                                  botan_view_ctx ctx,
                                  botan_view_str_fn view_fn) {
+   if(index != 0) {
+      // As of now there are no multi-value generic string entries.
+      return BOTAN_FFI_ERROR_OUT_OF_RANGE;
+   }
+
    auto view = [=](const std::string& value) { return invoke_view_callback(view_fn, ctx, value); };
 
    switch(value_type) {
@@ -200,8 +238,14 @@ int botan_x509_object_view_value(const Botan::X509_Object& object,
 
 int botan_x509_object_view_value(const Botan::X509_Object& object,
                                  botan_x509_value_type value_type,
+                                 size_t index,
                                  botan_view_ctx ctx,
                                  botan_view_bin_fn view_fn) {
+   if(index != 0) {
+      // As of now there are no multi-value generic binary entries.
+      return BOTAN_FFI_ERROR_OUT_OF_RANGE;
+   }
+
    auto view = [=](std::span<const uint8_t> value) { return invoke_view_callback(view_fn, ctx, value); };
 
    switch(value_type) {
@@ -262,7 +306,7 @@ int botan_x509_cert_view_binary_values(botan_x509_cert_t cert,
          case BOTAN_X509_SIGNATURE_SCHEME_BITS:
          case BOTAN_X509_SIGNATURE_BITS:
          case BOTAN_X509_DER_ENCODING:
-            return botan_x509_object_view_value(c, value_type, ctx, view_fn);
+            return botan_x509_object_view_value(c, value_type, index, ctx, view_fn);
 
          case BOTAN_X509_PEM_ENCODING:
          case BOTAN_X509_CRL_DISTRIBUTION_URLS:
@@ -275,6 +319,18 @@ int botan_x509_cert_view_binary_values(botan_x509_cert_t cert,
    });
 #else
    BOTAN_UNUSED(cert, value_type, index, ctx, view_fn);
+   return BOTAN_FFI_ERROR_NOT_IMPLEMENTED;
+#endif
+}
+
+int botan_x509_cert_view_binary_values_count(botan_x509_cert_t cert, botan_x509_value_type value_type, size_t* count) {
+#if defined(BOTAN_HAS_X509_CERTIFICATES)
+   return enumerator_count_values(count, [=](size_t index) {
+      return botan_x509_cert_view_binary_values(
+         cert, value_type, index, nullptr, [](auto, auto, auto) -> int { return BOTAN_FFI_SUCCESS; });
+   });
+#else
+   BOTAN_UNUSED(cert, value_type, count);
    return BOTAN_FFI_ERROR_NOT_IMPLEMENTED;
 #endif
 }
@@ -333,7 +389,7 @@ int botan_x509_cert_view_string_values(botan_x509_cert_t cert,
          case BOTAN_X509_CA_ISSUERS_URLS:
             return enumerate(c.ca_issuers(), index);
          case BOTAN_X509_PEM_ENCODING:
-            return botan_x509_object_view_value(c, value_type, ctx, view_fn);
+            return botan_x509_object_view_value(c, value_type, index, ctx, view_fn);
 
          case BOTAN_X509_SERIAL_NUMBER:
          case BOTAN_X509_SUBJECT_DN_BITS:
@@ -352,6 +408,18 @@ int botan_x509_cert_view_string_values(botan_x509_cert_t cert,
    });
 #else
    BOTAN_UNUSED(cert, value_type, index, ctx, view_fn);
+   return BOTAN_FFI_ERROR_NOT_IMPLEMENTED;
+#endif
+}
+
+int botan_x509_cert_view_string_values_count(botan_x509_cert_t cert, botan_x509_value_type value_type, size_t* count) {
+#if defined(BOTAN_HAS_X509_CERTIFICATES)
+   return enumerator_count_values(count, [=](size_t index) {
+      return botan_x509_cert_view_string_values(
+         cert, value_type, index, nullptr, [](auto, auto, auto) -> int { return BOTAN_FFI_SUCCESS; });
+   });
+#else
+   BOTAN_UNUSED(cert, value_type, count);
    return BOTAN_FFI_ERROR_NOT_IMPLEMENTED;
 #endif
 }
@@ -1079,7 +1147,7 @@ int botan_x509_crl_view_binary_values(botan_x509_crl_t crl_obj,
          case BOTAN_X509_SIGNATURE_SCHEME_BITS:
          case BOTAN_X509_SIGNATURE_BITS:
          case BOTAN_X509_DER_ENCODING:
-            return botan_x509_object_view_value(crl, value_type, ctx, view_fn);
+            return botan_x509_object_view_value(crl, value_type, index, ctx, view_fn);
 
          case BOTAN_X509_SUBJECT_DN_BITS:
          case BOTAN_X509_SUBJECT_KEY_IDENTIFIER:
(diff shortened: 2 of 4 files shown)


randombit:
I'm not firmly against it. I would definitely suggest view-style functions since, with cases like this where the lengths can be somewhat open-ended (especially so given the set of fields might be expanded over time), retry handling is kind of a pain.

OTOH if we have a single kind of getter dispatch function, is there really that much duplication to have 2 instead, one for certs and one for CRLs? One thing I did try for in the FFI interface is making it a type safe as possible when used from C. (This stems from the fact that a major driver of the FFI interface as it exists today is due to the requirements of RNP, which at the time was written entirely in C). This function which will require a cast for either input type bypasses that a bit, and tbh I don't really see how 2 type specific getter-dispatchers would be a problem, admittedly I'm not at all familiar with strongswan internals.

reneme:
> I would definitely suggest view-style functions since, with cases like this where the lengths can be somewhat open-ended [...].

:+1:

> OTOH if we have a single kind of getter dispatch function, is there really that much duplication to have 2 instead, one for certs and one for CRLs?

No, we could split it up. Perhaps the boilerplate might grow a bit faster once we introduce additional getters for strings, lists and such. Mostly I was interested in exploring the ergonomics of such a unified API. That said, I think I'll leave it in until the work on the strongswan plugin converges. Reverting this won't be a lot of work anyway.

src/lib/ffi/ffi.h
```diff
@@ -2249,6 +2249,40 @@ BOTAN_FFI_EXPORT(2, 13) int botan_x509_crl_destroy(botan_x509_crl_t crl);
...
+* X.509 Cert/CRL generic getters
+**************************/
+
+/**
+* This could either be a botan_x509_cert_t or botan_x509_crl_t.
+*/
+typedef void* botan_x509_object_t;
+
+typedef enum /* NOLINT(*-enum-size) */ {
+   BOTAN_X509_SERIAL_NUMBER,            /* binary big-endian encoding */
```
arckoor: I really wish we could make a breaking change and make this an MPI. For cert creation botan requires an MPI, so the setters I wrote (for that now closed PR) reflected that. For #5166 I am also returning MPIs. I think for the general usability of the API it'd be really nice if it were consistent everywhere, but I do understand if it remains at BE encoded arrays. Perhaps that could then be addressed in botan 4?

reneme: I understand the desire for consistency here. The current state of this PR adds functions that either view strings or binary values. Hence, the serial number is returned as a big-endian encoded binary value.

If we would somewhat relax the type safety constraints of the current FFI, we could add another getter function with a `void*` out-param. _In my opinion this could also be useful to generically access integer or boolean values_ such as validity timestamps, constraint flags or path length specifications.

Maybe like that:

```C++
int botan_x509_cert_get_values(botan_x509_cert_t cert,
                               botan_x509_value_type value_type,
                               size_t index,
                               void* out);

// read the certificate's serial as MP integer
botan_mp_t serial;
botan_x509_cert_get_values(cert, BOTAN_X509_SERIAL_NUMBER, 0, &serial);
botan_mp_destroy(serial);

// get the beginning of the certificate's validity interval
time_t not_before;
botan_x509_cert_get_values(cert, BOTAN_X509_NOT_BEFORE_TIMESTAMP, 0, &not_before);

// get the CA flag of the certificate
int is_ca_cert;
botan_x509_cert_get_values(cert, BOTAN_X509_IS_CERTIFICATE_AUTHORITY, 0, &is_ca_cert);
```

... obviously, this comes with all the undefined behavior pitfalls one would expect here. Perhaps @randombit has an opinion on that?

reneme:
In my opinion this is now ready-for-review. Documentation as well as PR description is updated accordingly.
We're heavily relying on this API within the upcoming strongSwan extension and it allows quite ergonomic usage.

> OTOH if we have a single kind of getter dispatch function, is there really that much duplication to have 2 instead, one for certs and one for CRLs?

I split the functions into CRL/cert as proposed. The required overhead indeed does not justify the added complexity in the API contract as well as the underlying implementation of the polymorphy.

The [potential extension discussed in an inline thread](https://github.com/randombit/botan/pull/5188#discussion_r2649158506) would make a lot of sense, in my opinion. Some details may need further discussion though, so I would suggest to split this off into another PR.

src/lib/ffi/ffi.h
```diff
@@ -2136,6 +2137,35 @@ int botan_mceies_decrypt(botan_privkey_t mce_key,
 
 typedef struct botan_x509_cert_struct* botan_x509_cert_t;
 
+/**
+ * Generic values that may be retrieved from X.509 certificates or CRLs via
+ * the generic getter functions.
+ *
+ * When extending this list the existing entries must stay backward-compatible
+ * to remain ABI compatible across versions. Therefore, new values must be added
+ * to the end of this list.
```
reneme, line 2237: To the best of my understanding C89 does not allow defining the bit-width of the underlying enum type. Currently, I naively listed all available value types but my gut tells me that we might need additional precautions to really keep this ABI and backward compatible.

Maybe assign values to all entries up-front? Perhaps grouping them by some arbitrary scheme while leaving value gaps to provide some flexibility when extending this enum in the future? Perhaps even define a `BOTAN_X509_RESERVED_VALUE = 0x0FFFFFFF` to enforce at least a 4-byte underlying integer.

Perhaps someone has some experience with such a thing?

arckoor:
Could you increase the FFI version? Would be nice if it could be the intended release date like outlined [here](https://github.com/randombit/botan/pull/5166#discussion_r2550432483). Also any plans for an impl for the python binding, or should that happen at a later date by someone(tm)? :p

reneme:
> Also any plans for an impl for the python binding, or should that happen at a later date by someone(tm)?

Shouldn't be too hard to do. I would leave it to a future step though, once the API in this is finalized.

src/lib/ffi/ffi.h
randombit: Strategy I've always gone with is grouping plus gaps and explicit assignment, it's so easy otherwise to accidentally insert a new enumeration and change the implicit integer values without noticing.

In C89 an enum is guaranteed to be an `int` anyway, which seems totally sufficient in terms of range.

src/lib/ffi/ffi_cert.cpp
```diff
@@ -78,6 +81,156 @@ int botan_x509_cert_load(botan_x509_cert_t* cert_obj, const uint8_t cert_bits[],
...
+#if defined(BOTAN_HAS_X509_CERTIFICATES)
+   auto enumerate = [view_fn, ctx](auto values, size_t idx) -> int {
+      if(idx >= values.size()) {
+         return BOTAN_FFI_ERROR_OUT_OF_RANGE;
+      } else {
+         return invoke_view_callback(view_fn, ctx, values[idx]);
+      }
+   };
+
+   auto enumerate_crl_distribution_points = [view_fn, ctx](const Botan::X509_Certificate& c, size_t idx) -> int {
```
atreiber94: When looking at this method, initially I was confused by the names `enumerate` and `enumerate_crl_distribution_points`, imo something like `view_at_idx` sounds better but no strong opinions here.

src/lib/ffi/ffi_cert.cpp
```diff
@@ -78,6 +81,156 @@ int botan_x509_cert_load(botan_x509_cert_t* cert_obj, const uint8_t cert_bits[],
...
+         case BOTAN_X509_SERIAL_NUMBER:
+            return view(c.serial_number());
+         case BOTAN_X509_SUBJECT_DN_BITS:
+            return view(c.raw_subject_dn());
+         case BOTAN_X509_ISSUER_DN_BITS:
+            return view(c.raw_issuer_dn());
+         case BOTAN_X509_SUBJECT_KEY_IDENTIFIER:
+            return c.subject_key_id().empty() ? BOTAN_FFI_ERROR_NO_VALUE : view(c.subject_key_id());
+         case BOTAN_X509_AUTHORITY_KEY_IDENTIFIER:
+            return c.authority_key_id().empty() ? BOTAN_FFI_ERROR_NO_VALUE : view(c.authority_key_id());
```
randombit: IMO - have `view` check if its argument is empty, if so immediately return `BOTAN_FFI_ERROR_NO_VALUE`, or otherwise invoke the view callback. The current approach misses that for example the DN *can* be empty [*]

[*] Also I just looked and while I haven't tested this it appears that if the serial number is zero we present that, in the byte encoding of the serial number, as an empty string. Which seems wrong probably? But anyway making the point that any of these values *might* be empty.

src/lib/ffi/ffi_cert.cpp
```diff
@@ -404,6 +557,51 @@ int botan_x509_crl_destroy(botan_x509_crl_t crl) {
...
+                                      size_t index,
+                                      botan_view_ctx ctx,
+                                      botan_view_bin_fn view_fn) {
+#if defined(BOTAN_HAS_X509_CERTIFICATES)
+   if(index != 0) {
+      // As of now there are no multi-value binary entries.
+      return BOTAN_FFI_ERROR_OUT_OF_RANGE;
+   }
+
+   auto view = [=](std::span<const uint8_t> value) { return invoke_view_callback(view_fn, ctx, value); };
```
randombit: Ditto earlier comment about checking for values in the `view` lambda so that the check is systematic

randombit (commented):
Thanks looking good

src/lib/ffi/ffi.h
reneme: @randombit This isn't super relevant for both this PR and the 3.11.0 release IMO. But perhaps for the (not-so-far) future direction of the X.509-related FFI. I'd like to have your general opinion on this idea before putting time into a draft for this.

TL;DR: Additionally to `..._get_binary_values()` and `..._get_string_values()` we could add a generic `_get_other_values()` (or whatever name) that takes a `void*` out-pointer and returns "any other data type" in the same fashion. E.g. integers such as validity timestamps, booleans such as basic constraints or even FFI object handles such as `botan_mp_t*` or `botan_public_key_t*`, depending on the requested value.

... I envision that this would allow deprecating a number of existing single-purpose getters, such as `botan_x509_cert_get_serial_number` or its proposed overload in #5230.

src/lib/ffi/ffi.h
reneme: Done like that. Thanks.

src/lib/ffi/ffi_cert.cpp
```diff
@@ -750,6 +920,90 @@ int botan_x509_crl_destroy(botan_x509_crl_t crl) {
...
+         case BOTAN_X509_SUBJECT_KEY_IDENTIFIER:
+         case BOTAN_X509_AUTHORITY_KEY_IDENTIFIER:
+         case BOTAN_X509_PUBLIC_KEY_PKCS8_BITS:
+         case BOTAN_X509_TBS_DATA_BITS:
+         case BOTAN_X509_SIGNATURE_SCHEME_BITS:
+         case BOTAN_X509_SIGNATURE_BITS:
+         case BOTAN_X509_DER_ENCODING:
+         case BOTAN_X509_CRL_DISTRIBUTION_URLS:
+         case BOTAN_X509_OCSP_RESPONDER_URLS:
+         case BOTAN_X509_CA_ISSUERS_URLS:
```
randombit, line 1205: GCC (correctly) warns that this function doesn't return a value if the enum is out of range. Probably should have a `default: return BOTAN_FFI_ERROR_BAD_PARAMETER` here

src/lib/ffi/ffi_cert.cpp
```diff
@@ -80,6 +83,173 @@ int botan_x509_cert_load(botan_x509_cert_t* cert_obj, const uint8_t cert_bits[],
...
+      case BOTAN_X509_TBS_DATA_BITS:
+         return view(object.tbs_data());
+      case BOTAN_X509_SIGNATURE_SCHEME_BITS:
+         return view(object.signature_algorithm().BER_encode());
+      case BOTAN_X509_SIGNATURE_BITS:
+         return view(object.signature());
+      case BOTAN_X509_DER_ENCODING:
+         return view(object.BER_encode());
+      default:
+         return BOTAN_FFI_ERROR_INTERNAL_ERROR; /* called with unexpected (non-generic) value_type */
```
randombit: Is this really an internal error? Seems more like a bad parameter to me...

src/lib/ffi/ffi_cert.cpp
randombit: Ah I see the patern in the others is to switch over the known values then have a final return outside of the switch, which is probably better overall. I left some mistaken comments suggesting adding `default:` cases, now deleted. This just needs a final return after the switch.

src/lib/ffi/ffi_cert.cpp
```diff
@@ -80,6 +83,173 @@ int botan_x509_cert_load(botan_x509_cert_t* cert_obj, const uint8_t cert_bits[],
...
+                                 botan_x509_value_type value_type,
+                                 botan_view_ctx ctx,
+                                 botan_view_str_fn view_fn) {
+   auto view = [=](const std::string& value) { return invoke_view_callback(view_fn, ctx, value); };
+
+   switch(value_type) {
+      case BOTAN_X509_PEM_ENCODING:
+         return view(object.PEM_encode());
+      default:
+         return BOTAN_FFI_ERROR_INTERNAL_ERROR; /* called with unexpected (non-generic) value_type */
```
randombit: Internal error here seems strange

src/lib/ffi/ffi_cert.cpp
reneme: > GCC (correctly) warns that this function doesn't return a value if the enum is out of range.

Yeah, I think that was an oversight yesterday.

src/lib/ffi/ffi_cert.cpp
reneme: This function is only called internally (to handle generic `X509_Object&` things). So if you call it internally with an unexpected type it is an internal error in my opinion.

reneme:
@randombit FYI: I'm currently looking into [@arckoor's suggestion to have additional API to get the expected "count" of entries](https://github.com/randombit/botan/issues/5219#issuecomment-3735165225). I'm hoping to do that in a way that doesn't multiply the `switch(type)` instances even more.

src/lib/ffi/ffi_cert.cpp
reneme: > Also I just looked and while I haven't tested this it appears that if the serial number is zero we present that, in the byte encoding of the serial number, as an empty string. Which seems wrong probably?

True. We had to add a special case in the strongSwan adapter. We could move this special-case into Botan which would handle this case more gracefully maybe. Also note that [RFC 5820 Section 4.1.2.2](https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.2) demands that certificates have a non-zero serial.

> The serial number MUST be a positive integer assigned by the CA to
   each certificate.  [...]  CAs MUST force the serialNumber to be a non-negative
   integer.
   [...]
   Note: Non-conforming CAs may issue certificates with serial numbers
   that are negative or zero.  Certificate users SHOULD be prepared to
   gracefully handle such certificates.

src/lib/ffi/ffi_cert.cpp
reneme: As stated above: This is meant to be used internally only (and only for the listed value types). Hence, if this default case is ever hit, this is an internal error.

Alternatively, we could add `BOTAN_ASSERT_UNREACHABLE()` here, perhaps.

src/lib/ffi/ffi_cert.cpp
reneme: > IMO - have view check if its argument is empty, if so immediately return BOTAN_FFI_ERROR_NO_VALUE, or otherwise invoke the view callback. The current approach misses that for example the DN can be empty.

I'm not convinced that this would yield the expected behavior. To stay with the example of DNs: even if wouldn't contain any RDNs, the encoding used here would still contain the bytes of an empty ASN.1 sequence , thus the to-be-viewed buffer wouldn't be empty and your suggestion wouldn't yield `NO_VALUE` either.

Frankly, I think for the empty DN case _we actually want_ to return the "empty" ASN.1 sequence. For instance, in the strongSwan code, this value is passed into a strongSwan-specific DN structure that expects to find said ASN.1 sequence albeit empty in this example.

reneme: That said, I would argue that `c.subject_key_id().empty()` is a special case. It would technically be better to check if the certificate contains the SKI extension and if not, yield "no value". Though, our `X509_Certificate` class defines the SKI buffer to be empty in this case, and that was easier to check here.

reneme:
I've added `*_count()` functions to the generic getters as suggested and addressed the latest review comments.

Re: `*_count()` - their implementations are maximally trivial for now, basically just actually "counting" how many entries are produced by the enumerator functions where `BOTAN_FFI_ERROR_NO_VALUE` counts as "0". Given that (1) the enumerators mostly just invoke the viewer functions straight on the C++ objects' memory and (2) long lists are not to-be-expected for X.509 certificates, I believe this is just good enough for now. It definitely has the nice side effect that the `*_count()` functions aren't at any risk to become inconsistent with the enumerator implementations.

src/lib/ffi/ffi_cert.cpp
```diff
@@ -80,6 +119,187 @@ int botan_x509_cert_load(botan_x509_cert_t* cert_obj, const uint8_t cert_bits[],
...
+
+   return BOTAN_FFI_VISIT(cert, [=](const Botan::X509_Certificate& c) -> int {
+      switch(value_type) {
+         case BOTAN_X509_CRL_DISTRIBUTION_URLS:
+            return enumerate_crl_distribution_points(c, index);
+         case BOTAN_X509_OCSP_RESPONDER_URLS:
+            return enumerate(wrap_or_empty(c.ocsp_responder()), index);
+         case BOTAN_X509_CA_ISSUERS_URLS:
+            return enumerate(c.ca_issuers(), index);
+         case BOTAN_X509_PEM_ENCODING:
```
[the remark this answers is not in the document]
reneme: Good catch! This would have been an endless loop in the trivial counter implementation! 😨

src/lib/ffi/ffi.h
randombit: Not a fan tbh, there is some implicit knowledge required to use this interface (if I ask for FOO it's an int, for a QUUX it's a mp_t, ..) that is not visible to the compiler and so uncheckable, which I think is worth avoiding. Also in any binding to a language that has typing of any kind, you'd just have to split out the interface again to handle the different return types. (Eg the Rust binding at most could have a "get_int_field" that handles getting values that are integer types, being careful to not call it with the wrong enumerator).

The only upside I see is it *could* replace many API endpoints with one. But then the implementation of that one API endpoint is huge because it has to handle 50 different whatevers, and personally I find 50 small single-purpose functions easier to reason about than one big multiplexer.

src/lib/ffi/ffi_cert.cpp
randombit: Yeah those functions should have returned an `optional`, but it didn't exist back in C++98... and I think `std::optional` doesn't work right with references *anyway* so it's not clear that adding a new interface is worth it either...

randombit: My point is that - if `X509_Certificate` returns an empty vector for one of these values (or for some future type added next month), what is the more reasonable behavior

- Call the view with an empty string
- Return `BOTAN_FFI_ERROR_NO_VALUE`

Either seems fine! I don't understand "empty string for this" / "error for that"

reneme: I got hung up on the DN counter-example. Given that this buffer won't ever be empty (and thus it won't ever return `NO_VALUE` for DNs), I'm convinced that this makes sense.

src/lib/ffi/ffi_cert.cpp
randombit: Using `BOTAN_ASSERT_UNREACHABLE()` would have the advantage that someone looking at the exception message (via `botan_error_last_exception_message`) would get some clue what happened.

randombit:
@reneme Shit sorry I thought this had already landed on master. Looks goods to me, needs a rebase though.

src/lib/ffi/ffi_cert.cpp
reneme: Fair. And it is an internal function that should essentially never be called with a wrong parameter anyway, so "assert unreachable" actually makes more sense.

reneme:
> I thought this had already landed on master.

... actually, me too. We're still waiting on another team to test run this in their system and it fully disappeared from my short term memory. Thanks for noticing and moving it forward.

I applied your suggestion re:"assert unreachable" and rebased to master (testing locally that it compiles/unittests with `--disable-modules=x509` as well).

reneme:
@randombit This is now rebased and ready to go. I can haz ✅ plz?

randombit (approved):
🍏

PR #5191, opened 2025-12-24
Add AVX-512 implementation of the Keccak permutation

randombit:


randombit:
This is only a bit faster than the earlier version in #5187 which at first seems strange since it literally executes 1/5 as many instructions in the main loop. However the current set of Intel cores can issue just 1 512-bit vector instruction per clock, but 2 - or rarely even 3 - 128/256-bit vector ops. This is confirmed by `perf` which reports this version issuing 1.7 instructions per cycle vs the original which was north of 4 instructions per cycle. The original approach also didn't need any explicit transposition step, since it just reordered the registers, and the 5x5 transposition takes up about 30% of the entire runtime. Still in the end this is ~5% faster than the original on Tiger Lake, and hopefully will improve more than #5187 would on future cores with higher issue rates (since it seems more likely to me that a future core will be able to execute 2 512-bit vectors ops per cycle, than it will be to execute 3 or 4 128 bit ops per cycle). Based on Agner Fog's instruction tables Zen5 *can* issue 2 512-bit vector ops per cycle, but also can still only issue 2 128/256 bit vector ops, which seems to validate this. If any readers have access to a Zen5 system I'd be very curious at the performance difference between this PR and #5187

randombit:
One thing that was a little surprising is that this is *much* faster than the AVX512 optimized Keccak in OpenSSL, at least on this particular core

```
type             16 bytes     64 bytes    256 bytes   1024 bytes   8192 bytes  16384 bytes
sha3-512         43868.54k   177694.59k   228740.10k   259500.03k   279207.94k   280281.09k
```

```
algo           operation       16 bytes        64 bytes       256 bytes      1024 bytes      8192 bytes     16384 bytes
SHA-3(512)     hash            82917.85       323070.52       346828.16       375555.55       396498.46       395550.60
```

PR #5163, opened 2025-11-20
Port `simd_2x64` to SIMD128 and make use of it in `ghash_vperm`

polarnis:
This is a follow-up to https://github.com/randombit/botan/pull/5155, extending the library's support for Wasm SIMD128. This time the `simd_2x64` and `ghash_vperm` modules were ported. I had some difficulty settling on good names for a few `SIMD_2x64` functions (like `reverse_all_bytes`), trying to be as unambiguous as possible - feedback appreciated. The `simd_2x64` tests are essentially a one-to-one adaptation of the existing `simd_4x32` tests (with `simd_2x64`-specific additions).

And some numbers (AMD Ryzen 7 4800H, V8 in Chrome 142, buffer size 4096):
|           | Before #5155 (2d01ad0cedd1d5ae6c00ce372942df5c5a029e8a) | After #5155 (884e1bac17b587ad12475a32745229ba80a00435) | This PR (ee6e281a02103b12e11767f21587fc3cfd91f455) |
|-----------------|-------------------------------------------|-------------------------------------------|-------------------------------------------|
| AES-128/GCM(16) encrypt (MiB/s) | 52.912                                    | 100.558                                   | 145.069                                   |
| AES-128/GCM(16) decrypt (MiB/s) | 53.658                                    | 100.948                                   | 144.288                                   |
| AES-256/GCM(16) encrypt (MiB/s) | 44.044                                    | 92.388                                   | 125.686                                   |
| AES-256/GCM(16) decrypt (MiB/s) | 44.071                                    | 92.815                                   | 125.869                                   |

PR #5201, opened 2025-12-31
Use `SIMD_2x64` in Argon2 on all available architectures

polarnis:
https://github.com/randombit/botan/pull/5163 ported `SIMD_2x64` to SIMD128, but the Argon2 module could not benefit from it, even though it already uses `SIMD_2x64`, because the `argon2_ssse3` module hard-codes SSSE3 as the only supported ISA. This patch removes that constraint and renames `argon2_ssse3` to `argon2_simd64` to better reflect its implementation.

This is technically a breaking change, as users might explicitly enable or disable the `argon2_ssse3` module, which is now removed. I'm not sure whether this is acceptable.

And some numbers (node 22.16.0, Ryzen 7 4800H):

| Argon2id params | Before (ms/op) | After (ms/op) |
|-----------------|----------------|---------------|
| (8192, 1, 1)    | 8.67           | 6.74          |
| (8192, 1, 4)    | 8.80           | 6.87          |
| (8192, 4, 1)    | 28.55          | 20.95         |
| (8192, 4, 4)    | 29.07          | 21.64         |
| (65536, 1, 1)   | 74.39          | 59.32         |
| (65536, 1, 4)   | 73.57          | 58.72         |
| (65536, 4, 1)   | 247.26         | 191.28        |
| (65536, 4, 4)   | 247.94         | 189.46        |
| (262144, 1, 1)  | 313.41         | 256.57        |
| (262144, 1, 4)  | 305.64         | 247.23        |
| (262144, 4, 1)  | 1036.42        | 808.23        |
| (262144, 4, 4)  | 1030.25        | 803.99        |


randombit:
> This is technically a breaking change, as users might explicitly enable or disable the argon2_ssse3 module, which is now removed. I'm not sure whether this is acceptable.

Yes that's fine - our SemVer guarantees that code written against 3.N will build and run with 3.{N+1} but there are no similar guarantees with regards to *building* the library.

randombit:
CI failure is unrelated

randombit (approved):
LGTM thanks!

PR #5202, opened 2026-01-02
Migrate to native WebAssembly exceptions

arekmula:
This patch brings some improvements. It fixes the following build error:
```
em++: error: DISABLE_EXCEPTION_CATCHING=0 is not compatible with -fwasm-exceptions
```
Emscripten explicitly forbids combining the legacy JS exception system with the modern native WASM exception system. 

It also improves the performance as the native WASM exceptions are significantly more performant than the old method.

### DIFF
--- a/src/build-data/cc/emcc.txt
+++ b/src/build-data/cc/emcc.txt
@@ -2,8 +2,8 @@ macro_name CLANG
 
 binary_name em++
 
-lang_flags "-s DISABLE_EXCEPTION_CATCHING=0 -std=c++20 -D_REENTRANT"
-lang_binary_linker_flags "-s ALLOW_MEMORY_GROWTH=1 -s WASM=1 -s NO_DISABLE_EXCEPTION_CATCHING"
+lang_flags "-s -fwasm-exceptions -std=c++20 -D_REENTRANT"
+lang_binary_linker_flags "-s ALLOW_MEMORY_GROWTH=1 -s WASM=1 -s -fwasm-exceptions"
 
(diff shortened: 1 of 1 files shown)


randombit:
cc @polarnis

randombit:
Am I understanding correctly that this new exception handling mechanism is the same as the one covered in https://caniuse.com/wf-wasm-exception-handling

polarnis:
@randombit Yes, that matches my understanding as well (it might be a bit confusing, I admit). Emscripten effectively supports three exception-handling modes:

1. JavaScript-based EH: `-fexceptions` (slower, widely supported)
2. Wasm-based EH (legacy): `-fwasm-exceptions` + `WASM_LEGACY_EXCEPTIONS=1` (fast, widely supported)
3. Wasm-based EH with exnref: `-fwasm-exceptions` + `WASM_LEGACY_EXCEPTIONS=0` (fast, but not yet broadly supported)

Since `WASM_LEGACY_EXCEPTIONS` defaults to 1, using `-fwasm-exceptions` alone selects the second mode. That seems like a solid default choice.

You can read more about it [here](https://github.com/emscripten-core/emscripten/blob/420db67b7ffc4702c76128e7998e828dd2f4a64d/src/settings_internal.js#L177-L182) and [here](https://github.com/emscripten-core/emscripten/blob/420db67b7ffc4702c76128e7998e828dd2f4a64d/src/settings.js#L785-L790).

randombit (approved):

PR #5209, opened 2026-01-04
Improve support for running tests on Emscripten

polarnis:
While it would be possible to run tests in a headless browser on CI (with some struggle), it would be even harder to collect JUnit test reports. An easier option is to target Node.js with [`NODERAWFS`](https://emscripten.org/docs/api_reference/Filesystem-API.html#noderawfs), which forwards to the actual filesystem making test reports collection very simple (and additionally makes cert store tests pass).

An example run (in my fork) can be found [here](https://github.com/polarnis/botan/actions/runs/20699908607/job/59420614694).
See individual commits for more details.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -2197,6 +2197,11 @@ def create_template_vars(source_paths, build_paths, options, modules, disabled_m
 
     def test_exe_extra_ldflags():
         if osinfo.matches_name("emscripten"):
+            # It doesn't make much sense (and it's not even possible) to preload files when FS is not virtualized.
+            virtualized_fs = '-sNODERAWFS=1' not in cc.ldflags(options)
+            if not virtualized_fs:
+                return ''
+
             return '--preload-file=%s@src/tests/data' % source_paths.test_data_dir
 
         return ''
--- a/src/scripts/ci_build.py
+++ b/src/scripts/ci_build.py
@@ -308,9 +308,10 @@ def determine_flags(target, target_os, target_cpu, target_cc, cc_bin, ccache,
         flags += ['--disable-modules=locking_allocator']
 
     if target == 'emscripten':
-        flags += ['--cpu=wasm']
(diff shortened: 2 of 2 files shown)


src/tests/test_os_utils.cpp
```diff
@@ -87,21 +87,29 @@ class OS_Utils_Tests final : public Test {
...
+            if(hr_ts1 < hr_ts2) {
+               break;
+            }
+
+            ++trials;
+            hr_ts2 = Botan::OS::get_high_resolution_clock();
          }
 
-         result.test_lt("high resolution clock eventually changes value", counts, max_repeats);
+         result.confirm("high resolution clock eventually changes value", hr_ts1 < hr_ts2);
```
reneme: Perhaps like that?

```C++
for (size_t trials = 0; trials < max_trials; ++trials) {
   if (hr_ts1 < Botan::OS::get_high_resolution_clock()) {
      result.test_success("high resolution clock made progress");
      return result;
   }
}

result.test_failure("high resolution clock didn't make progress, even after many samples");
```

src/tests/test_filters.cpp
```diff
@@ -676,7 +677,7 @@ class Filter_Tests final : public Test {
       static Test::Result test_pipe_fd_io() {
          Test::Result result("Pipe file descriptor IO");
 
-   #if defined(BOTAN_HAS_PIPE_UNIXFD_IO) && defined(BOTAN_HAS_CODEC_FILTERS)
+   #if defined(BOTAN_HAS_PIPE_UNIXFD_IO) && defined(BOTAN_HAS_CODEC_FILTERS) && !defined(BOTAN_TARGET_OS_IS_EMSCRIPTEN)
```
randombit, line 680: The really correct thing here is likely to not build `fd_unix` on Emscripten at all, but this requires some additional help from the build system. :/ I will take a look at this separately, this is fine for now

PR #5215, opened 2026-01-08
In configure.py guard against invalid invalid OS/CPU/compiler names

randombit:
This would cause a generic KeyError exception which is pretty unfriendly.


randombit:
Robot was right but also wrong, it was true that `validate_options` already was checking these but that was called only *after* the maps were accessed, I think this was changed when stack scrubbing was introduced since that wanted to pass the compiler version to `validate_options`. I just moved that check to later within the function (and also changed it to be a warning instead of a hard error)

PR #5217, opened 2026-01-08
FFI: Query Certificate Subject/Issuer Alternative Names and Name Constraints

reneme:
This introduces FFI query functions for X.509 certificate subject/issuer alternative names and name constraint entries. Both are enumerating getter functions that create new `botan_x509_general_name_t` object handles. It is worth noting that the newly introduced `botan_x509_general_name_t` handle wraps the `GeneralName` class which is slated for removal from the public API with Botan4.

Regardless, the design of the proposed FFI extensions closely follows the definitions in [RFC 5280 Section 4.2.1.6 and 4.2.1.7](https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.6) where GeneralName is defined as a union type of several name type options. Both subject and issuer alternative names are defined as collections of this GeneralName type.

```
GeneralName ::= CHOICE {
     otherName                       [0],
     rfc822Name                      [1],
     dNSName                         [2],
     x400Address                     [3],
     directoryName                   [4],
     ediPartyName                    [5],
     uniformResourceIdentifier       [6],
     iPAddress                       [7],
     registeredID                    [8]}

GeneralNames ::= SEQUENCE SIZE (1..MAX) OF GeneralName

SubjectAltName ::= GeneralNames
IssuerAltName ::= GeneralNames
```

Note that Botan currently supports a subset of those name types only. Namely "otherName", "x400Address", "ediPartyName" and "registeredID" are not supported.

Similarly [RFC 5280 Section 4.2.1.10](https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.10) essentially defines each NameConstraint as a list of GeneralNames with some extra annotation. These annotations are not exposed in the FFI extension proposed here.

```
NameConstraints ::= SEQUENCE {
           permittedSubtrees       [0]     GeneralSubtrees OPTIONAL,
           excludedSubtrees        [1]     GeneralSubtrees OPTIONAL }

GeneralSubtrees ::= SEQUENCE SIZE (1..MAX) OF GeneralSubtree
GeneralSubtree ::= SEQUENCE {
      base                    GeneralName,
      minimum         [0]     BaseDistance DEFAULT 0,
      maximum         [1]     BaseDistance OPTIONAL }
```

Hence, reusing the new `botan_x509_general_name_t` for name constraint querying leads to a sleek API in my opinion.

Regarding the deprecation of `GeneralName`: I don't know why this class is deprecated, but I believe that the design of the C API extensions is future-proof in the sense that it can easily be implemented with other underlying C++ types in Botan4. Hence, I opted for a minor extension of `GeneralName` here, despite its deprecation.

### DIFF
--- a/src/lib/ffi/ffi_cert.cpp
+++ b/src/lib/ffi/ffi_cert.cpp
@@ -469,6 +469,107 @@ int botan_x509_cert_excluded_name_constraints(botan_x509_cert_t cert,
    return BOTAN_FFI_ERROR_NOT_IMPLEMENTED;
 #endif
 }
+}
+
+namespace {
+
+/**
+ * As specified in RFC 5280 Section 4.2.1.6. alternative names essentially are a
+ * collection of GeneralNames. This allows mapping a single entry of @p altnames
+ * to a GeneralName by its @p index. If the index is out of range, std::nullopt
+ * is returned.
+ */
+std::optional<Botan::GeneralName> extract_general_name_at(const Botan::AlternativeName& altnames, size_t index) {
+   if(index < altnames.email().size()) {
+      auto itr = altnames.email().begin();
+      std::advance(itr, index);
+      return Botan::GeneralName::email(*itr);
+   }
+   index -= altnames.email().size();
+
+   if(index < altnames.dns().size()) {
+      auto itr = altnames.dns().begin();
+      std::advance(itr, index);
+      return Botan::GeneralName::dns(*itr);
+   }
+   index -= altnames.dns().size();
+
+   if(index < altnames.directory_names().size()) {
+      auto itr = altnames.directory_names().begin();
+      std::advance(itr, index);
+      return Botan::GeneralName::directory_name(*itr);
+   }
+   index -= altnames.directory_names().size();
+
+   if(index < altnames.uris().size()) {
+      auto itr = altnames.uris().begin();
+      std::advance(itr, index);
+      return Botan::GeneralName::uri(*itr);
+   }
+   index -= altnames.uris().size();
+
+   if(index < altnames.ipv4_address().size()) {
+      auto itr = altnames.ipv4_address().begin();
+      std::advance(itr, index);
+      return Botan::GeneralName::ipv4_address(*itr);
+   }
+
+   return std::nullopt;
+}
+
+}  // namespace
+
+extern "C" {
+
+int botan_x509_cert_subject_alternative_names(botan_x509_cert_t cert,
+                                              size_t index,
+                                              botan_x509_general_name_t* alt_name) {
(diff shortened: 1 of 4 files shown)


randombit:
> Regarding the deprecation of GeneralName: I don't know why this class is deprecated

The issue with current `GeneralName` is that its API was completely dedicated to name constraint processing but also that API was useless for correct and fast name constraint processing. It may make sense to keep it (or at least the name of the struct) as a name type without the bad matching logic. Anyway no problem with adding a generalname type to FFI

src/configs/typos.toml
```diff
@@ -42,6 +42,7 @@ ser = "ser"
 stuf = "stuf"
 typ = "typ"
 indx = "indx"
+ACI = "ACI"
```
randombit, line 45: I'm not seeing this appear elsewhere in the diff?

src/lib/ffi/ffi_cert.cpp
```diff
@@ -261,6 +262,256 @@ int botan_x509_cert_view_public_key_bits(botan_x509_cert_t cert, botan_view_ctx
...
+#if defined(BOTAN_HAS_X509_CERTIFICATES)
+   return BOTAN_FFI_VISIT(name, [=](const Botan::GeneralName& n) {
+      if(Botan::any_null_pointers(type)) {
+         return BOTAN_FFI_ERROR_NULL_POINTER;
+      }
+
+      if(auto mapped_type = to_botan_x509_general_name_types(n.type_code())) {
+         *type = mapped_type.value();
+         if(*type != BOTAN_X509_OTHER_NAME /* ... viewing of other-names not supported */) {
+            return BOTAN_FFI_SUCCESS;
```
randombit: The logic here is inverted and so hard to parse vs

```
if(*type == BOTAN_X509_OTHER_NAME) {
  // not currently supported
  return BOTAN_FFI_ERROR_...
}
```

src/configs/typos.toml
reneme: Its in the test case.

```C++
TEST_FFI_OK(botan_x509_general_name_view_binary_value, (dn, bin.delegate(), bin.callback()));
const auto organization = read_distinguished_name(bin.get()).get_first_attribute("O");
result.test_eq("dn as expected", organization, "ACI Informatica S.p.A.");
```

... a quick glance at the documentation of typos didn't teach me how to silence such warnings with inline comments as for clang-tidy or pylint, like `// NOLINT`. But perhaps I missed that? I'd much prefer that over the central config for such one-off things.

randombit: It doesn't offer anything directly, but apparently it can be hacked using the builtin regex support. I'll take a look.

PR #5220, opened 2026-01-09
FFI: Enumeration of entries in an X.509 CRL

reneme:
This adds `botan_x509_crl_get_entries` that allows listing individual CRL entries along with the affected certificate serial, the revocation time and reason.

The serial number is retrieved as a big-endian binary encoding into a user-provided buffer. Using the view-callback style doesn't strike me as feasible or ergonomic for this. One would either have to introduce an additional function just for the serial or mixing out-params with view invocation in a single function call.

### DIFF
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -2239,6 +2239,7 @@ BOTAN_FFI_EXPORT(2, 8) const char* botan_x509_cert_validation_status(int code);
 **************************/
 
 typedef struct botan_x509_crl_struct* botan_x509_crl_t;
+typedef struct botan_x509_crl_entry_struct* botan_x509_crl_entry_t;
 
 BOTAN_FFI_EXPORT(2, 13) int botan_x509_crl_load_file(botan_x509_crl_t* crl_obj, const char* crl_path);
 BOTAN_FFI_EXPORT(2, 13)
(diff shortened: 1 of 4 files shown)


src/lib/ffi/ffi.h
```diff
@@ -2249,6 +2250,28 @@ BOTAN_FFI_EXPORT(2, 13) int botan_x509_crl_destroy(botan_x509_crl_t crl);
...
+* @param expire_time_seconds_since_epoch  the revocation timestamp of this entry
+* @param reason                           the reason code (see RFC 5280 - 5.3.1)
+* @param serial_bits                      the revoked cert's serial number
+* @param serial_bits_len                  the length of the cert's serial number
+*
+* @returns BOTAN_FFI_ERROR_OUT_OF_RANGE if the given @p index is out of range of
+*          the CRL entry list.
+*/
+BOTAN_FFI_EXPORT(3, 11)
+int botan_x509_crl_get_entries(botan_x509_crl_t crl,
```
randombit: It kind of feels like a view-style callback would be nice here, but we'd have to set up dedicated infrastructure for just this one tuple which would be unlikely to be reusable for anything else...

randombit: An alternative would be adding a dedicated CRL entry type with it's own getters but that seems like a lot of overhead on our end and not, practically speaking, that much easier to use.

randombit (approved):
lgtm

I don't particularly like the API but the alternatives I can come up with are more work on our end for little to no benefit for end users so (shrug)

src/lib/ffi/ffi.h
reneme: Mhm. I also wondered about mixing the view style with the integers as out-params in the same function signature. I guess that would work but it's also quite confusing.

Having an extra type is a good point. But then you're forced to keep re-allocating this. Whereas for the serial, one might get away with a fixed size array given that most CAs probably have a consistent length for the serial.

reneme: That said, individual CRL entries can have extensions (e.g. to specify the issuer of revoked certificates in an "indirect CRL"). Also the "reason" is encoded in such an extension. I didn't think of this before, but that really speaks for the dedicated type as it provides extensibility for such extensions in the future. I'll try this.

reneme: > I'll try this.

FWIW: the additional work needed wasn't that bad and the call site is actually much nicer with a dedicated type.

src/lib/ffi/ffi.h
```diff
@@ -2249,6 +2251,40 @@ BOTAN_FFI_EXPORT(2, 13) int botan_x509_crl_destroy(botan_x509_crl_t crl);
...
+* in seconds.
+*/
+BOTAN_FFI_EXPORT(3, 11)
+int botan_x509_crl_entry_revocation_date(botan_x509_crl_entry_t entry, uint64_t* time_since_epoch);
+
+/**
+* View the serial number associated with the given CRL @p entry.
+*/
+BOTAN_FFI_EXPORT(3, 11)
+int botan_x509_crl_entry_view_serial_number(botan_x509_crl_entry_t entry, botan_view_ctx ctx, botan_view_bin_fn view);
```
reneme, line 2289: Using the same argument as in #5221 regarding OIDs and also [this explicit remark](https://github.com/randombit/botan/pull/5188#discussion_r2649158506) by @arckoor, we should consider adding an "overload" to export the CRL entry serial as a `botan_mp_t` object. I.e. provide `botan_x509_crl_entry_serial_number(..., botan_mp_t* serial)` along with this byte-oriented view function.

arckoor: There may be an argument for also adding an MPI view function for the existing X509 cert serial

reneme: I'll combine those in a follow-up PR after merging the current stack of changes.

src/lib/ffi/ffi.h
```diff
@@ -2254,6 +2256,40 @@ BOTAN_FFI_EXPORT(2, 13) int botan_x509_crl_destroy(botan_x509_crl_t crl);
...
+*
+* @param crl     the CRL whose entries should be listed
+* @param index   the index of the CRL entry to return
+* @param entry   an object handle containing the CRL entry data
+*
+* @returns BOTAN_FFI_ERROR_OUT_OF_RANGE if the given @p index is out of range of
+*          the CRL entry list.
+*/
+BOTAN_FFI_EXPORT(3, 11)
+int botan_x509_crl_entries(botan_x509_crl_t crl, size_t index, botan_x509_crl_entry_t* entry);
```
arckoor, line 2270: Small nit: botan-rs has [this](https://github.com/randombit/botan-rs/blob/ac29f2d6357ad58d7e7753a88329b294ffa95277/botan/src/lib.rs#L26) neat macro, to simplify calling methods that write to a new object. It expects that new object to be the first parameter though. (used e.g. like [so](https://github.com/randombit/botan-rs/blob/ac29f2d6357ad58d7e7753a88329b294ffa95277/botan/src/pubkey.rs#L38))

reneme: Discussion in https://github.com/randombit/botan/pull/5230#discussion_r2680356734

PR #5221, opened 2026-01-09
FFI: Allow querying more X.509 Certificate Constraints

atreiber94:
This adds FFI wrappers for querying additional constraints from X.509 certificates. Namely, whether or not a certificate has the CA-bit set, its optional path limit, and whether or not certain extended key usages are enabled.

### DIFF
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -2194,7 +2194,7 @@ BOTAN_FFI_EXPORT(2, 0) int botan_x509_cert_to_string(botan_x509_cert_t cert, cha
 BOTAN_FFI_EXPORT(3, 0)
 int botan_x509_cert_view_as_string(botan_x509_cert_t cert, botan_view_ctx ctx, botan_view_str_fn view);
 
-/* Must match values of Key_Constraints in key_constraints.h */
+/* Must match values of Key_Constraints in pkix_enums.h */
 enum botan_x509_cert_key_constraints /* NOLINT(*-enum-size,*-use-enum-class) */ {
    NO_CONSTRAINTS = 0,
    DIGITAL_SIGNATURE = 32768,


src/lib/ffi/ffi.h
```diff
@@ -2199,6 +2210,22 @@ enum botan_x509_cert_key_constraints /* NOLINT(*-enum-size,*-use-enum-class) */
...
+* Typical OIDs to check for:
+*   * "PKIX.ServerAuth"
+*   * "PKIX.ClientAuth"
+*   * "PKIX.CodeSigning"
+*   * "PKIX.OCSPSigning"
+*
+* The @p oid parameter can be either a canonical OID string or identifiers as
+* indicated in the examples above.
+*/
+BOTAN_FFI_EXPORT(3, 11) int botan_x509_cert_allowed_extended_usage(botan_x509_cert_t cert, const char* oid);
```
randombit: There is `botan_oid_t` maybe that should be used here instead of the string arg?

[Don't feel super strongly about this, I guess it does add a bit of verbosity at the call site but feels a little cleaner to use a proper type here given it already exists]

reneme: I saw that new OID type and also concluded that this would be more consistent. But then I hesitated because of the overhead this would potentially bring. Let's say you wanted to implement an `is_ocsp_signer()` helper.

```C++
// before
bool is_ocsp_signer(botan_x509_cert_t cert) {
    return botan_x509_cert_allowed_extended_usage(cert, "PKIX.OCSPSigner") == BOTAN_FFI_SUCCESS;
}

// after
bool is_ocsp_signer(botan_x509_cert_t cert) {
    botan_asn1_oid_t oid;
    if(botan_asn1_oid_from_string(&oid, "PKIX.OCSPSigner") != BOTAN_FFI_SUCCESS) {
        return false; // Failed to create OID
    }

    bool allowed = botan_x509_cert_allowed_extended_usage_oid(cert, oid) == BOTAN_FFI_SUCCESS;
    botan_asn1_oid_destroy(oid);

    return allowed;
}
```

Perhaps provide an "overload" so that both work?

randombit: Yeah something like `botan_x509_cert_allowed_extended_usage_str` vs `botan_x509_cert_allowed_extended_usage_oid` sound good, helpful if the application is already using the OID type but not getting in the way if they just want to check a single somewhat static value as in your OCSP example

reneme: Done.

PR #5224, opened 2026-01-09
Bump the FFI version for 3.11

randombit:


randombit:
I updated the table. I also added a checklist into ffi.txt

PR #5230, opened 2026-01-11
FFI: Return serial numbers as `botan_mp_t` for X.509 certificates and CRL entries

reneme:
See here: https://github.com/randombit/botan/pull/5220#discussion_r2679478411. Personally, I don't need this. But I do see the benefit for consistency.

/cc @arckoor

### DIFF
--- a/src/lib/ffi/ffi_cert.cpp
+++ b/src/lib/ffi/ffi_cert.cpp
@@ -15,6 +15,7 @@
    #include <botan/x509_crl.h>
    #include <botan/x509cert.h>
    #include <botan/x509path.h>
+   #include <botan/internal/ffi_mp.h>
    #include <botan/internal/ffi_oid.h>
 #endif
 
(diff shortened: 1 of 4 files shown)


randombit (approved):
CI failure is relevant

src/lib/ffi/ffi_cert.cpp
```diff
@@ -272,6 +273,22 @@ int botan_x509_cert_get_serial_number(botan_x509_cert_t cert, uint8_t out[], siz
 #endif
 }
 
+int botan_x509_cert_serial_number(botan_x509_cert_t cert, botan_mp_t* serial_number) {
```
arckoor, line 277: Same nit as in #5220, just in case

reneme: Let's discuss here, given that this PR is currently still open:

Original nit from #5220:

> botan-rs has [this](https://github.com/randombit/botan-rs/blob/ac29f2d6357ad58d7e7753a88329b294ffa95277/botan/src/lib.rs#L26) neat macro, to simplify calling methods that write to a new object. It expects that new object to be the first parameter though. (used e.g. like [so](https://github.com/randombit/botan-rs/blob/ac29f2d6357ad58d7e7753a88329b294ffa95277/botan/src/pubkey.rs#L38))

Frankly, I'm hesitant to break API consistency for this. Namely, the convention that the "this-pointer" of the object being worked on is always the first parameter (AFAICS).

For instance, another such "initializer function":

```C++
int botan_x509_cert_get_public_key(botan_x509_cert_t cert, botan_pubkey_t* key)
```

... but anyway, this is not the hill I'm willing to die on.

arckoor: If there is such a convention, I have readily [ignored](https://github.com/randombit/botan/blob/ab88122f8a55d78dd7ac13060e081a732053de96/src/lib/ffi/ffi_ec.cpp#L162) it in the past 😅

randombit: > If there is such a convention

That is the convention, I missed this completely in the review of that, guess we're stuck with it now, alas.

randombit: (Re the macro, it would be easy to extend that macro to support arbitrary positions of the output object, I'd much rather do that and keep the FFI interface as consistent in approach as possible)

Issue #5219, opened 2026-01-08
strongSwan: parse/validate X.509 objects in strongSwan's Botan plugin


reneme:
StrongSwan contains [a plugin for using Botan as a crypto provider](https://github.com/strongswan/strongswan/tree/master/src/libstrongswan/plugins/botan) for quite some time already. We're currently working on extending this plugin with functionality for parsing/validating X.509 certificates and CRLs. To access the relevant bits in Botan, a number of FFI functions have to be added. This issue is meant to provide an overview of the work in progress.

In the best-case scenario we'd like to have most of these additions available in the upcoming Botan 3.11.0 release so that we can start integrating this into strongSwan upstream. Locally, we already have a working prototype.

### Required new Functionality in the FFI

* querying various data fields from certificates and CRLs (as applicable)
  _(made available via #5188)_
  * serial number
  * subject/issuer DN as DER encoding
  * subject/issuer key identifiers
  * subject public key as PKCS#8 container
  * "to be signed" data for signature validation
  * the object's signature as raw bits
  * the signature algorithm identifier as DER encoding
  * CRL distribution point URLs
  * OCSP responder URLs
* basic/extended constraints
  * is_CA bit
  * path length constraint
  * extended key usage bits
* subject/issuer alternative names
  _(see #5217)_
* name constraint definitions
  _(see #5217)_
* IP address range information (from [the IPAddrBlock extension](https://github.com/randombit/botan/pull/4699))
* CRL validity interval (this update, next update)
* enumeration of individual CRL entries

### Pull Requests related to this

* https://github.com/randombit/botan/pull/5188
* https://github.com/randombit/botan/pull/5189
* https://github.com/randombit/botan/pull/5217

### Currently Known Limitations

* **IPv6 addresses**
  Botan currently doesn't support handling IPv6 addresses in subject/issuer alternative names as well as name constraints. Any such IPv6 address entries would currently not be visible to strongSwan when using the Botan plugin
* **OCSP**
  We don't require any OCSP-based certificate revocation checks at the moment. Integration with Botan's OCSP functionality is left for future work

/cc @tobiasbrunner

randombit:
Re IPv6 names and Delta CRLs you are correct these are not currently supported, I'd be open to adding support if needed.

Re the currently open PRs I owe you some reviews but I don't think there are any showstoppers that would prevent these being merged in time for 3.11

reneme:
[...]

PR #5231, opened 2026-01-12
X509: Multiple OCSP Responders

atreiber94:
This addresses one of the limitations hindering integration of Botan's X509 parsing to strongSwan (see #5219), namely multiple OCSP responder URIs.

Before, `ocsp_responder()` of the AuthorityInformationAccess extension or of `X509_Certificate` would return a single URI, even if the certificate parsed contains multiple URIs as OCSP responders.

The single responder interfaces have been deprecated (e.g. it is not clear what "a OCSP responder" means in the now-deprecated methods - in my tests, it would return the last URI). The legacy interface is kept by returning the first URI (or the empty string if there are no URIs).

### PR Dependencies

This is based on #5188 so I could test the FFI integration (also directly against the strongSwan X509 test suite).

### OCSP `check_online`
The changes here only expose multiple OCSP responder URIs to certificate parsing as well as the constructors/accessors of the AuthorityInformationAccess extension and `X509_Certificate`. I did not include this into the OCSP features like `OCSP::check_online` since this code does not appear to be well tested (we should rather discuss what to do with this code going forward).

### DIFF
--- a/src/lib/x509/x509_ext.h
+++ b/src/lib/x509/x509_ext.h
@@ -322,16 +322,28 @@ class BOTAN_PUBLIC_API(2, 0) Certificate_Policies final : public Certificate_Ext
 class BOTAN_PUBLIC_API(2, 0) Authority_Information_Access final : public Certificate_Extension {
    public:
       std::unique_ptr<Certificate_Extension> copy() const override {
-         return std::make_unique<Authority_Information_Access>(m_ocsp_responder, m_ca_issuers);
+         return std::make_unique<Authority_Information_Access>(m_ocsp_responders, m_ca_issuers);
       }
 
       Authority_Information_Access() = default;
 
+      BOTAN_DEPRECATED("Use constructor with list of OCSP responders")
       explicit Authority_Information_Access(std::string_view ocsp,
                                             const std::vector<std::string>& ca_issuers = std::vector<std::string>()) :
-            m_ocsp_responder(ocsp), m_ca_issuers(ca_issuers) {}
+            m_ocsp_responders{std::string(ocsp)}, m_ca_issuers(ca_issuers) {}
 
-      std::string ocsp_responder() const { return m_ocsp_responder; }
+      explicit Authority_Information_Access(std::vector<std::string> ocsp_responders,
+                                            std::vector<std::string> ca_issuers = std::vector<std::string>()) :
+            m_ocsp_responders(std::move(ocsp_responders)), m_ca_issuers(std::move(ca_issuers)) {}
+
+      BOTAN_DEPRECATED("Use ocsp_responders") std::string ocsp_responder() const {
+         if(m_ocsp_responders.empty()) {
+            return {};
+         }
+         return m_ocsp_responders[0];
+      }
+
+      const std::vector<std::string>& ocsp_responders() const { return m_ocsp_responders; }
 
       static OID static_oid() { return OID({1, 3, 6, 1, 5, 5, 7, 1, 1}); }
 
@@ -342,12 +354,12 @@ class BOTAN_PUBLIC_API(2, 0) Authority_Information_Access final : public Certifi
    private:
       std::string oid_name() const override { return "PKIX.AuthorityInformationAccess"; }
 
-      bool should_encode() const override { return (!m_ocsp_responder.empty() || !m_ca_issuers.empty()); }
+      bool should_encode() const override { return (!m_ocsp_responders.empty() || !m_ca_issuers.empty()); }
 
       std::vector<uint8_t> encode_inner() const override;
       void decode_inner(const std::vector<uint8_t>& in) override;
 
-      std::string m_ocsp_responder;
+      std::vector<std::string> m_ocsp_responders;
       std::vector<std::string> m_ca_issuers;
 };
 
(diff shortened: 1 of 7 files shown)


src/lib/x509/x509_ext.h
```diff
@@ -322,16 +322,27 @@ class BOTAN_PUBLIC_API(2, 0) Certificate_Policies final : public Certificate_Ext
...
 
       Authority_Information_Access() = default;
 
-      explicit Authority_Information_Access(std::string_view ocsp,
+      BOTAN_DEPRECATED("Use constructor with list of OCSP responders") explicit Authority_Information_Access(
+         std::string_view ocsp, const std::vector<std::string>& ca_issuers = std::vector<std::string>()) :
+            m_ocsp_responders{std::string(ocsp)}, m_ca_issuers(ca_issuers) {}
+
+      explicit Authority_Information_Access(const std::vector<std::string>& ocsp_responders,
                                             const std::vector<std::string>& ca_issuers = std::vector<std::string>()) :
```
reneme: Given that this is a new constructor, let's use this opportunity and make it more idiomatic by letting it take lvalues and move them into the object members:

```C++
explicit Authority_Information_Access(std::vector<std::string> ocsp_responders,
                                      std::vector<std::string> ca_issuers = std::vector<std::string>()) :
            m_ocsp_responders(std::move(ocsp_responders)), m_ca_issuers(std::move(ca_issuers)) {}
```

src/lib/x509/x509cert.cpp
```diff
@@ -797,8 +804,12 @@ std::string X509_Certificate::to_string() const {
       }
    }
 
-   if(!ocsp_responder().empty()) {
-      out << "OCSP responder " << ocsp_responder() << "\n";
+   const std::vector<std::string> ocsp_responders = this->ocsp_responders();
```
reneme: `const auto&` ... no need to copy this for just printing the values.

src/lib/x509/x509cert.h
```diff
@@ -339,7 +339,12 @@ class BOTAN_PUBLIC_API(2, 0) X509_Certificate : public X509_Object {
       /**
       * Return the listed address of an OCSP responder, or empty if not set
       */
-      std::string ocsp_responder() const;
+      BOTAN_DEPRECATED("Use ocsp_responders") std::string ocsp_responder() const;
+
+      /**
+      * Return the listed addresses of OCSP responders, or empty if not set
+      */
+      std::vector<std::string> ocsp_responders() const;
```
reneme: Return a `const&` here, given that we hold it as a vector of this type anyway?

src/lib/x509/x509_ext.h
```diff
@@ -322,16 +322,27 @@ class BOTAN_PUBLIC_API(2, 0) Certificate_Policies final : public Certificate_Ext
...
-      std::string ocsp_responder() const { return m_ocsp_responder; }
+      explicit Authority_Information_Access(std::vector<std::string> ocsp_responders,
+                                            std::vector<std::string> ca_issuers = std::vector<std::string>()) :
+            m_ocsp_responders(std::move(ocsp_responders)), m_ca_issuers(std::move(ca_issuers)) {}
+
+      BOTAN_DEPRECATED("Use ocsp_responders") std::string ocsp_responder() const {
+         if(m_ocsp_responders.empty()) {
+            return {};
+         }
+         return m_ocsp_responders[0];
```
reneme, line 343: If we were extremely nitpicky regarding semver this should be `m_ocsp_responders.back()`, judging from the now-replaced parsing code:

```C++
      if(oid == OID::from_string("PKIX.OCSP")) {
         const BER_Object name = info.get_next_object();

         if(name.is_a(6, ASN1_Class::ContextSpecific)) {
            m_ocsp_responder = ASN1::to_string(name);
         }
      }
```

... that did repeatedly overwrite the `std::string m_ocsp_responder` until settling on the last entry.

src/lib/x509/x509cert.cpp
```diff
@@ -537,7 +537,14 @@ bool X509_Certificate::is_critical(std::string_view ex_name) const {
 }
 
 std::string X509_Certificate::ocsp_responder() const {
-   return data().m_ocsp_responder;
+   if(data().m_ocsp_responders.empty()) {
+      return {};
+   }
+   return data().m_ocsp_responders[0];
```
reneme, line 544: `data().m_ocsp_responders.back()` might be slightly better. See comment above.

src/lib/x509/x509_ext.h
randombit: I'd be inclined to say it returned an arbitrary value that until now happened to always be the last one, this was *not* an intentional thing and also not afaik documented anywhere.

atreiber94: Intuitively I'd expect the first value to be returned. Since the previous behavior was never documented I'd agree returning the first value would not be a notable semver issue. But I don't have a strong opinion on this, I can also understand returning the last value for consistent behavior.

atreiber94:
Re-based onto the latest changes of #5188 which should hopefully turn the CI green.

randombit:
Needs another rebase

reneme:
Rebased again. We'll rebase once more once #5188 is in, because this still stacks onto it.

reneme:
#5188 has landed. Could you rebase?

PR #5232, opened 2026-01-12
FFI: add `*_count()` functions for some FFI enumerators

reneme:
This adds functions to query the number of entries in index-based enumerator functions that were introduced recently in #5220 and #5217 [as suggested](https://github.com/randombit/botan/issues/5219#issuecomment-3735165225).

### DIFF
--- a/doc/api_ref/ffi.rst
+++ b/doc/api_ref/ffi.rst
@@ -1825,6 +1825,11 @@ X.509 Certificates
    objects. If the given index is not available,
    :cpp:enumerator:`BOTAN_FFI_ERROR_OUT_OF_RANGE` is returned.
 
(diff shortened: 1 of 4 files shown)


arckoor:
While you're at it, could you add `botan_x509_cert_get_issuer_dn_count` and `botan_x509_cert_get_subject_dn_count`? These two were introduced in 2.0, and also have that behaviour, but aren't exposed in botan-rs yet so I still have the chance to use the nicer api there :p

reneme:
> While you're at it, could you add botan_x509_cert_get_issuer_dn_count and botan_x509_cert_get_subject_dn_count?

I'll do a follow-up PR. Edit: #5234.

PR #5233, opened 2026-01-13
Generate python documentation with sphinx

arckoor:
Changes the python documentation to be generated by sphinx.

I added mentions in the doc for things that were previously not there, e.g. TOTP, key wrapping, ...

I also moved a few doc comments from `__init__` to the class itself, I had incorrectly assumed that sphinx would generate an extra constructor for those, it does not (so the doc would not be visible at all). At least my intellisense also doesn't pick up on any comments in the init, so I think this is better.

### DIFF
--- a/doc/api_ref/python.rst
+++ b/doc/api_ref/python.rst
@@ -17,836 +17,231 @@ for Botan 3 it is ``botan3``.
 
 Versioning
 ----------------------------------------
-.. py:function:: version_major()
+.. autofunction:: version_major
 
-   Returns the major number of the library version.
+.. autofunction:: version_minor
 
-.. py:function:: version_minor()
+.. autofunction:: version_patch
 
-   Returns the minor number of the library version.
+.. autofunction:: ffi_api_version
 
-.. py:function:: version_patch()
+.. autofunction:: version_string
 
-   Returns the patch number of the library version.
 
-.. py:function:: version_string()
+Utilities
+----------------------------------------
 
-   Returns a free form version string for the library
+.. autofunction:: const_time_compare
+
+.. autodata:: MPILike
+
+.. autoclass:: BotanException
+   :members:
 
 Random Number Generators
 ----------------------------------------
 
-.. py:class:: RandomNumberGenerator(rng_type = 'system')
-
-     Previously ``rng``
-
-     Type 'user' also allowed (userspace HMAC_DRBG seeded from system
-     rng). The system RNG is very cheap to create, as just a single file
-     handle or CSP handle is kept open, from first use until shutdown,
-     no matter how many 'system' rng instances are created. Thus it is
-     easy to use the RNG in a one-off way, with `botan.RandomNumberGenerator().get(32)`.
-
-     For some use cases it can be useful to provide a custom RNG implementation.
-     Use 'custom' as the rng_type and provide the ``get_callback=`` and
-     ``add_entropy_callback=`` arguments. The latter is optional.
-     ``get_callback`` takes an integer and is expected to return a bytes object
-     with the requested number of random bytes.
-     ``add_entropy_callback`` takes a bytes object containing entropy bytes and
-     is expected to add the given entropy to the RNG.
-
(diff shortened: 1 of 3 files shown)


arckoor:
My proposed reordering would be the following:
- Move the TPM section below `RandomNumberGenerator`
- Move `BlockCipher` below `SymmetricCipher` (and make a # Ciphers section)
- Move `_load_buf_or_file` to # Internal utilities

The resulting class / section order should be reflected in `python.rst`.

Reorder within the classes:
- function order: `__init__` -> `__del__` (and other `__functions__`) -> `handle_` (if applicable) (handle could also be before the other `__functions__`)
- `@classmethod`s afterwards
- then: `algo_name`, followed by "info-methods", e.g. `key_length`, `estimated_strength`, ...
- then: functions that operate on their class, `clear`, `reset`, `set_key`, ...

Here's a [branch](https://github.com/arckoor/botan/tree/python-reorder) where I played that out once (for reference, not for merging). The [first](https://github.com/randombit/botan/commit/cc33691931384016412e3c745e57744d69e9030a) commit reorders the classes, the [second](https://github.com/randombit/botan/commit/ee067a22155935e0c617d65f46467309acf057a4) the functions.

I don't think this is *the* way to reorder, just one way to do it. I think at least some ideas have merit, so feel free to choose whatever you feel is best and rely on intuition otherwise c:

arckoor:
@randombit just fyi, without this the python `XOF` class will not be documented in the manual

arckoor:
@randombit I've rebased against master to include the padded nist key wrap, could you take a look sometime?

randombit:
@arckoor Sorry finally taking the time to really think about this and our doc situation in general. Yes change makes sense and mostly looks good. There does seem to be a problem building the docs if PYTHONPATH is not set to include `src/python` though - after a `configure.py` and `make docs` I get 

```
WARNING: autodoc: failed to import 'version_major' from module 'botan3'; the following exception was raised:
  File "/usr/lib/python3.14/site-packages/sphinx/ext/autodoc/_dynamic/_importer.py", line 131, in _import_from_module_and_path
    module = _import_module(module_name, try_reload=True)
  File "/usr/lib/python3.14/site-packages/sphinx/ext/autodoc/_dynamic/_importer.py", line 218, in _import_module
    raise ModuleNotFoundError(msg, name=modname)  # NoQA: TRY301
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'botan3'
 [autodoc.import_object]
WARNING: autodoc: failed to import 'version_minor' from module 'botan3'; the following exception was raised:
  File "/usr/lib/python3.14/site-packages/sphinx/ext/autodoc/_dynamic/_importer.py", line 131, in _import_from_module_and_path
    module = _import_module(module_name, try_reload=True)
  File "/usr/lib/python3.14/site-packages/sphinx/ext/autodoc/_dynamic/_importer.py", line 218, in _import_module
    raise ModuleNotFoundError(msg, name=modname)  # NoQA: TRY301
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'botan3'
.......
```

I think in `conf.py` where you are doing the various patching for the mock DLL you also need to add a `sys.path` dir (relative to the `conf.py`) pointing into `src/python`.

randombit:
[CI passes because CI runs everything with `PYTHONPATH=src/python` set]

arckoor:
I was for some reason testing with `ci_build.py` instead of `make docs` when first opening this, sorry for not catching it earlier. Rebased to master, everything that was added should be documented in the rst file.

randombit (approved):

arckoor:
The complaint about turning single backticks into double backticks was warranted, even if copilot did miss a lot of instances of it :p
At least on VS Code double backticks render exactly the same as single ones, so for IDEs it shouldn't break anything, and for sphinx it does result in nicer output

PR #5235, opened 2026-01-13
Add XOF Wrapper to botan3.py

reneme:
... what it said on the box.

### DIFF
--- a/src/python/botan3.py
+++ b/src/python/botan3.py
@@ -28,8 +28,9 @@ from collections.abc import Iterable
 
 # This Python module requires the FFI API version introduced in Botan 3.10.0
 #
+# 3.11.0 - XOF API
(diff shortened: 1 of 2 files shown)


src/python/botan3.py
```diff
@@ -983,6 +999,59 @@ def final(self) -> bytes:
...
         return _ctype_bufout(out)
 
+#
+# eXtensible Output Functions
+#
+class XOF:
+    """eXtensible Output Function (XOF)"""
+
+    def __init__(self, algo: str | c_void_p):
+        """The ``algo`` param is a string (e.g 'SHAKE-256', 'Ascon-XOF128')"""
```
arckoor: Please move this to the docstring of the class itself, see https://github.com/randombit/botan/pull/5233#issue-3806555838

PR #5238, opened 2026-01-15
Apply a workaround for GCC miscompilation of Serpent AVX512 in amalgamation mode

randombit:
Workaround is similar to the one already in place for dealing with a GCC miscompilation of the AVX2 Serpent code. This has been seen in CI with GCC 11 and GCC 13. I can't replicate the problem using GCC 14 or GCC 15.


randombit:
Indeed this hack seems to have done it - I restarted the x86-64 amalgamation build until it ran on an AVX-512 enabled machine, and now the tests pass.

PR #5240, opened 2026-01-15
Some changes to reduce time taken by the coverage build

randombit:
Running the CLI tests takes approximately 25% of the entire runtime of the coverage CI build, and that is dominated by the xmss_sign_test and speed_pk_tests

Change CI to only run the 'slow' CLI tests in the shared build which doesn't have the overhead of coverage collection.

Split up the speed pk tests into fast and slow variants; only a few of the public key algorithms (most notably RSA and the hash based signatures) are really that costly.

Move a few tests that were categorized as slow, but actually are not, into the regular test sequence.

Remove generating safe primes from the random_prime benchmark. This was very slow, there aren't enough samples for it to be a useful performance guide, and at this point the performance of generating a safe prime isn't even interesting since finite field DH is kil.

### DIFF
--- a/src/scripts/test_cli.py
+++ b/src/scripts/test_cli.py
@@ -262,6 +262,8 @@ def test_cli(cmd, cmd_options,
 
     if "\r\n" in stdout:
         stdout = stdout.replace("\r\n", "\n")
+    if "\r\n" in stderr:
+        stderr = stderr.replace("\r\n", "\n")
 
     if stderr:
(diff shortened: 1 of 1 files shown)


randombit:
Ah this has the side effect of running a few more of the CLI tests on Windows and mac and there we seem to have some problems. For Windows I expect it's a line-ending thing. For macOs it appears having the additional `commoncrypto` provider surprises a test.

PR #5252, opened 2026-01-20
FIX: `botan_x509_cert_allowed_extended_usage_*` should return "1" on true

reneme:
Similar to #5236.


reneme:
> Looks good, but no doc or doc comment updates??

Neither specified the return value for true/false, and given #5249 I also didn't see a reason to add them.

reneme:
@randombit 🔔 ... anything left to do here?

PR #5254, opened 2026-01-21
Avoid hard dependency on `pcurves_secp*` in the TLS module

reneme:
This adds some accommodations to the test suite and the base TLS module to allow building and successfully running the unit tests with a configuration where TLS 1.2 is enabled but secp* curves are disabled.

### DIFF
--- a/src/build-data/policy/modern.txt
+++ b/src/build-data/policy/modern.txt
@@ -51,6 +51,13 @@ ffi
 tls
 prf_tls
 
+# pcurves
+pcurves_secp256r1
+pcurves_secp384r1
+pcurves_secp521r1
+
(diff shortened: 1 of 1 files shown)


src/lib/pubkey/pk_algs.cpp
```diff
@@ -7,7 +7,9 @@
 
 #include <botan/pk_algs.h>
 
+#include <botan/asn1_obj.h>
 #include <botan/assert.h>
+#include <botan/pk_keys.h>
```
reneme, line 12: When building with `--minimized-build --enable-modules=x509` this fails to build without those includes.

randombit (approved):

PR #5258, opened 2026-01-23
Simplify value_exists implementation

KaganCanSit:
Hello,

This PR request was created to replace the manual loop in `value_exists` with `std::find`. There are no functional changes.

I also considered the following, but I don't think they are necessary;
* `std::ranges::find()` - (I try compiler explorer GCC11.0 and Clang 14.0.0 OK)
* `requires std::equality_comparable_with<T, V>`

Regards.

### DIFF
--- a/src/lib/utils/stl_util.h
+++ b/src/lib/utils/stl_util.h
@@ -15,6 +15,7 @@
 #include <botan/secmem.h>
 #include <botan/strong_type.h>
 
+#include <algorithm>
 #include <cstring>
 #include <functional>
 #include <optional>
@@ -49,12 +50,7 @@ RetT reduce(const std::vector<KeyT>& keys, RetT acc, ReducerT reducer)
 */
 template <typename T, typename V>
 bool value_exists(const std::vector<T>& vec, const V& val) {
-   for(size_t i = 0; i != vec.size(); ++i) {
-      if(vec[i] == val) {
-         return true;
-      }
-   }
-   return false;
+   return std::find(vec.begin(), vec.end(), val) != vec.end();
 }
 
 template <typename T, typename Pred>


src/lib/utils/stl_util.h
```diff
@@ -49,12 +50,7 @@ RetT reduce(const std::vector<KeyT>& keys, RetT acc, ReducerT reducer)
 */
 template <typename T, typename V>
 bool value_exists(const std::vector<T>& vec, const V& val) {
```
reneme, line 52: While you're on it: this could probably even use `std::span<const T>` to abstract from the container type.

src/lib/utils/stl_util.h
```diff
@@ -15,6 +15,7 @@
 #include <botan/secmem.h>
 #include <botan/strong_type.h>
 
+#include <algorithm>
```
randombit, line 18: Thank you but - no

`algorithm` is an enormous header and seems to have a relatively large impact on compilation times

src/lib/utils/stl_util.h
randombit: Sure but then every caller has to explicitly `span{v}` its arguments and why?

KaganCanSit:
Thanks for the review @reneme @randombit. I understand that while it may improve readability, it also comes at a higher cost.

I made another change in a different branch. I'd like to ask about that here as well. There are a total of 4 ‘FIXME’ messages in the `stl_util.h` file. (`out_ptr` and `out_opt`) We can explicitly mark the constructors here, but I believe the others should not be marked due to design considerations. (They are hidden by design for C API interoperability.) I can make this adjustment; would that be helpful?

```cpp
constexpr out_ptr_t(T& outptr) noexcept : m_ptr(outptr), m_rawptr(nullptr) {}

[[nodiscard]] constexpr operator typename T::element_type **() && noexcept { return &m_rawptr; }



constexpr out_opt_t(std::optional<T>& outopt) noexcept : m_opt(outopt) {}

[[nodiscard]] constexpr operator T*() && noexcept { return &m_raw; }
```

```cpp
constexpr explicit out_ptr_t(T& outptr) noexcept : m_ptr(outptr), m_rawptr(nullptr) {}

// NOLINTNEXTLINE(*-explicit-conversions)  implicit by design for C API interop
[[nodiscard]] constexpr operator typename T::element_type **() && noexcept { return &m_rawptr; }



constexpr explicit out_opt_t(std::optional<T>& outopt) noexcept : m_opt(outopt) {}

// NOLINTNEXTLINE(*-explicit-conversions)  implicit by design for C API interop
[[nodiscard]] constexpr operator T*() && noexcept { return &m_raw; }
```

PR #5261, opened 2026-01-25
Replace index-based loop with range-based for in value_exists

KaganCanSit:
Hello,

This PR replaces the index-based loop with a range-based for loop in the `value_exists` function. While modern compilers optimize both approaches equivalently, the range-based for loop is more readable and idiomatic in modern C++ (explicit const and loop variable).

This addresses the concerns from PR #5258 regarding the compilation overhead of including `<algorithm>`.

Best regards.

### DIFF
--- a/src/lib/utils/stl_util.h
+++ b/src/lib/utils/stl_util.h
@@ -49,8 +49,8 @@ RetT reduce(const std::vector<KeyT>& keys, RetT acc, ReducerT reducer)
 */
 template <typename T, typename V>
 bool value_exists(const std::vector<T>& vec, const V& val) {
-   for(size_t i = 0; i != vec.size(); ++i) {
-      if(vec[i] == val) {
+   for(const auto& elem : vec) {
+      if(elem == val) {
          return true;
       }
    }


devnexen:
> Hello,
> 
> This PR aims to improve readability by using range base instead of index base in the `value_exists` function and to eliminate the overhead of `size()` calls and index bounds checks.

I am curious about the overhead of size().

KaganCanSit:
> > Hello,
> > This PR aims to improve readability by using range base instead of index base in the `value_exists` function and to eliminate the overhead of `size()` calls and index bounds checks.
> 
> I am curious about the overhead of size().

Actually, there's no cost involved for modern compilers, you're right. The main point here is that range-based for loops are more readable and more convenient in modern C++.

I will edit the PR statement accordingly.

devnexen:
I myself prefer, 9 times out of 10, the modern form as well, the change itself looks good but I ll let @randombit decide about it :)

KaganCanSit:
> I myself prefer, 9 times out of 10, the modern form as well, the change itself looks good but I ll let @randombit decide about it :)

Thank you for your quick response.

The range-based loop can prevent very simple errors (especially shadowing and index errors within nested loops, etc.), and I also prefer it as much as possible. With the current PR description, it can be reviewed more accurately and clearly by the maintainers.

Additional Note: I don't think it's necessary, but type safety checks could be added for this function using concepts. (for implicit comparisons)

devnexen:
Yes, I do not think the type check via concept is a terrible idea.

PR #5265, opened 2026-01-25
modes: starting to replace raw pointer arithmetic with std::span view.

devnexen:


src/lib/modes/aead/ccm/ccm.cpp
```diff
@@ -187,7 +187,7 @@ void CCM_Encryption::finish_msg(secure_vector<uint8_t>& buffer, size_t offset) {
    E.encrypt(format_b0(sz), T);
 
    for(size_t i = 0; i != ad.size(); i += CCM_BS) {
-      xor_buf(T.data(), &ad[i], CCM_BS);
+      xor_buf(std::span{T}.first(CCM_BS), std::span(&ad[i], CCM_BS));
       E.encrypt(T);
    }
```
reneme, line 192: For the record: This could be simplified even further, along those lines:


```suggestion
   BufferSlicer ad(ad_buf());
   BOTAN_ARG_CHECK(ad.remaining() % CCM_BS == 0, "AD is block size multiple");

   const BlockCipher& E = cipher();

   secure_vector<uint8_t> T(CCM_BS);
   E.encrypt(format_b0(sz), T);

   while(!ad.empty()) {
      xor_buf(T, ad.take<CCM_BS>());
      E.encrypt(T);
   }
   BOTAN_DEBUG_ASSERT(ad.empty());
```

... no need for the `.first` for T because T is allocated with exactly `CCM_BS` bytes anyway. Also, you might have to include `stl_util.h` for the `BufferSlicer`.

src/lib/base/symkey.cpp
```diff
@@ -109,8 +110,8 @@ OctetString operator+(const OctetString& k1, const OctetString& k2) {
 OctetString operator^(const OctetString& k1, const OctetString& k2) {
    secure_vector<uint8_t> out(std::max(k1.length(), k2.length()));
 
-   copy_mem(out.data(), k1.begin(), k1.length());
-   xor_buf(out.data(), k2.begin(), k2.length());
+   copy_mem(std::span{out}.first(k1.length()), k1.bits_of());
+   xor_buf(std::span{out}.first(k2.length()), k2.bits_of());
    return OctetString(out);
```
reneme, line 115: While you're on it:


```suggestion
   copy_mem(std::span{out}.first(k1.length()), k1.bits_of());
   xor_buf(std::span{out}.first(k2.length()), k2.bits_of());
return OctetString(std::move(out));
```

src/lib/modes/aead/ccm/ccm.cpp
```diff
@@ -239,7 +240,7 @@ void CCM_Decryption::finish_msg(secure_vector<uint8_t>& buffer, size_t offset) {
    E.encrypt(format_b0(sz - tag_size()), T);
 
    for(size_t i = 0; i != ad.size(); i += CCM_BS) {
-      xor_buf(T.data(), &ad[i], CCM_BS);
+      xor_buf(std::span{T}.first(CCM_BS), std::span(&ad[i], CCM_BS));
       E.encrypt(T);
    }
```
reneme, line 245: similarly as above.

PR #5274, opened 2026-01-31
Improve code readability in parsing utilities

KaganCanSit:
Hello,

This PR contains several changes to improve the readability of the code in `src/lib/utils/parsing.cpp` and the header file. They were primarily discovered using clang-tidy. There are no API or function changes.

Additional Note: I updated the parameter name `s` to `str` based on the naming conventions in the header file. But of course, this name is to change.

If you leave a comment for further review, I'd be happy to look into it when I have time.

Best regards.

### DIFF
--- a/src/lib/utils/parsing.cpp
+++ b/src/lib/utils/parsing.cpp
@@ -54,12 +54,12 @@ uint32_t to_u32bit(std::string_view str_view) {
 /*
 * Parse a SCAN-style algorithm name
 */
-std::vector<std::string> parse_algorithm_name(std::string_view namex) {
-   if(namex.find('(') == std::string::npos && namex.find(')') == std::string::npos) {
-      return {std::string(namex)};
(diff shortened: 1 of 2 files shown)


src/lib/utils/parsing.cpp
```diff
@@ -359,11 +359,7 @@ bool host_wildcard_match(std::string_view issued_, std::string_view host_) {
    }
 
    // Wildcard issued name must have at least 3 components
-   if(dots_seen < 2) {
-      return false;
-   }
```
randombit, line 364: Please revert this bit - this function is quite security critical and was already involved in *two* security issues (CVE-2015-7826, CVE-2018-9127), which is basically a record for us, it is written in a very particular way.

Rest of change looks fine.

KaganCanSit: I'm taking it back immediately.

randombit (approved):

Issue #5164, opened 2025-11-20
Investigate slow build times


randombit:
Relating to #5162 

I modified the makefile to record how long it took to compile each file, and did a build using Clang 21 on a Linux machine. Overall runtime was about 4500 seconds. Median compilation time was 6.24 seconds, 10% percentile was 2.76, 90% percentile 10.75, 98% percentile was 16.57 seconds, worst was 29.7 seconds (`test_tls_rfc8448.cpp`)

The tests seem disproportionately slow to compile. The *fastest* test source compilation was 5.8 seconds, and 42% of the files taking over 10 seconds were test files (despite tests, as a percentage of number of files or lines of code, being only about 25% of the codebase). Something is definitely wrong here - especially considering most of the test data is intentionally kept out of the compilation path entirely and instead read from files at runtime.

KaganCanSit:
A few measurement methods and values ​​to provide some assistance;

## Compiler Time Analysis

### Compile: 
```bash
ninja clean && ./configure.py --without-documentation --cc=clang --build-targets=static,cli,tests --build-tool=ninja --extra-cxxflags="-ftime-trace" && ninja
```

### General:
```bash
awk '{d = $2 - $1; sum += d; count++} END {printf "Total %.2f s\nAverage: %.2f ms\nFile: %d\n", sum/1000, sum/count, count}' .ninja_log
```

```bash
Total: 2633.51 s
Average: 3735.47 ms
File: 705
```

### General Slowest: 
```bash
awk '{print ($2 - $1)/1000, $4}' .ninja_log | sort -rn | head -20
```

```bash
17.702 build/obj/test/test_ffi.o
17.141 build/obj/test/test_tls_session_manager.o
17.127 build/obj/test/test_tls_rfc8448.o
14.383 build/obj/lib/x509_ext.o
13.973 build/obj/test/test_utils_bitvector.o
13.143 build/obj/test/test_x509_rpki.o
12.384 build/obj/test/test_utils.o
12.219 build/obj/test/test_x509_path.o
11.976 build/obj/test/test_tls_cipher_state.o
11.01 build/obj/test/test_bigint.o
10.796 build/obj/lib/math_pcurves_numsp512d1.o
10.737 build/obj/test/test_tls_record_layer_13.o
10.177 build/obj/test/test_kyber.o
10.06 build/obj/lib/math_pcurves_sm2p256v1.o
10.003 build/obj/test/unit_x509.o
[...]

PR #5279, opened 2026-02-02
Move ASN1_Time to its own header to avoid <chrono> inclusions

randombit:
GH #5164

### DIFF
--- a/src/lib/asn1/asn1_obj.h
+++ b/src/lib/asn1/asn1_obj.h
@@ -9,7 +9,6 @@
 
 #include <botan/exceptn.h>
 #include <botan/secmem.h>
-#include <chrono>
 #include <iosfwd>
 #include <optional>
 #include <span>
@@ -21,6 +20,8 @@ namespace Botan {
 
 class BER_Decoder;
 class DER_Encoder;
+class ASN1_Time;  // in asn1_time.h
+typedef ASN1_Time X509_Time;
 
 /**
 * ASN.1 Class Tags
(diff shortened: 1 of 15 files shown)


configure.py
```diff
@@ -2910,7 +2910,17 @@ def header_contents(self, name):
...
-        for line in self.file_contents[name]:
+        # TODO(Botan4) remove this garbage
+        def nasty_hack(name, contents):
+            if name == 'asn1_obj.h':
+                idx = contents.index('   #include <botan/asn1_time.h>\n')
+                assert idx > 0
+                # remove guard and #endif
+                contents[idx-1:idx+2] = contents[idx][:3]
+
+            return contents
```
reneme: Alternatively: maybe it would make sense to `#define BOTAN_AMALGAMATION_BUILD` early-on in the amalgamation header creation and additionally guard the include with `!defined(BOTAN_AMALGAMATION_BUILD)`. This wouldn't be a nasty hack and might prove useful for other things in the future.

src/lib/asn1/asn1_obj.h
```diff
@@ -510,4 +441,9 @@ class std::hash<Botan::OID> {
       size_t operator()(const Botan::OID& oid) const noexcept { return static_cast<size_t>(oid.hash_code()); }
 };
 
+// TODO(Botan4) remove this
+#if !defined(BOTAN_IS_BEING_BUILT) && !defined(BOTAN_DISABLE_DEPRECATED_FEATURES)
+   #include <botan/asn1_time.h>
+#endif
```
reneme: Maybe leave a note to future-us?


```suggestion
// Backward-compatibility for downstream applications that use ASN1_Time while
// including asn1_obj.h but not (yet) asn1_time.h introduced in Botan 3.11.0
//
// TODO(Botan4) remove this
#if !defined(BOTAN_IS_BEING_BUILT) && !defined(BOTAN_DISABLE_DEPRECATED_FEATURES)
   #include <botan/asn1_time.h>
#endif
```

src/lib/asn1/asn1_time.h
```diff
@@ -0,0 +1,86 @@
...
+* Comparison Operations
+*/
+BOTAN_PUBLIC_API(2, 0) bool operator==(const X509_Time& x, const X509_Time& y);
+BOTAN_PUBLIC_API(2, 0) bool operator!=(const X509_Time& x, const X509_Time& y);
+BOTAN_PUBLIC_API(2, 0) bool operator<=(const X509_Time& x, const X509_Time& y);
+BOTAN_PUBLIC_API(2, 0) bool operator>=(const X509_Time& x, const X509_Time& y);
+BOTAN_PUBLIC_API(2, 0) bool operator<(const X509_Time& x, const X509_Time& y);
+BOTAN_PUBLIC_API(2, 0) bool operator>(const X509_Time& x, const X509_Time& y);
+
+typedef X509_Time ASN1_Time;
```
reneme: If the new authoritative name is now `X509_Time` (which does make sense in my opinion), should we consider deprecating this typedef and therefore the old `ASN1_Time` name?

src/lib/asn1/asn1_time.h
```diff
@@ -0,0 +1,86 @@
...
+      std::string to_string() const;
+
+      /// Returns a human friendly string representation of no particular formatting
+      std::string readable_string() const;
+
+      /// Return if the time has been set somehow
+      bool time_is_set() const;
+
+      ///  Compare this time against another
+      int32_t cmp(const X509_Time& other) const;
```
reneme: While we're working on this header (perhaps as a follow-up): this could be conveniently replaced by a spaceship operator. I would suggest deprecating this method in favor of ordinary comparison operators. See also other comments for the mechanics.


```suggestion
      ///  Compare this time against another
      BOTAN_DEPRECATED("use ordinary comparison operators instead") int32_t cmp(const X509_Time& other) const;
```

src/lib/asn1/asn1_time.h
```diff
@@ -0,0 +1,86 @@
...
+
+/*
+* Comparison Operations
+*/
+BOTAN_PUBLIC_API(2, 0) bool operator==(const X509_Time& x, const X509_Time& y);
+BOTAN_PUBLIC_API(2, 0) bool operator!=(const X509_Time& x, const X509_Time& y);
+BOTAN_PUBLIC_API(2, 0) bool operator<=(const X509_Time& x, const X509_Time& y);
+BOTAN_PUBLIC_API(2, 0) bool operator>=(const X509_Time& x, const X509_Time& y);
+BOTAN_PUBLIC_API(2, 0) bool operator<(const X509_Time& x, const X509_Time& y);
+BOTAN_PUBLIC_API(2, 0) bool operator>(const X509_Time& x, const X509_Time& y);
```
reneme: Let's use C++20's comparison features here:

```suggestion
/*
* Comparison Operations
*/
BOTAN_PUBLIC_API(3, 11) std::strong_ordering operator<=>(const X509_Time& x, const X509_Time& y);
BOTAN_PUBLIC_API(3, 11) inline bool operator==(const X509_Time& x, const X509_Time& y) {
   return (x <=> y) == std::strong_ordering::equal;
}
```

... note that for this free-standing implementation to work, we need to befriend the spaceship within the class:

```C++
friend std::strong_ordering operator<=>(const X509_Time& x, const X509_Time& y);
```

The implementation in the cpp file would look like that:

```C++
std::strong_ordering operator<=>(const X509_Time& t1, const X509_Time& t2) {
   if(!t1.time_is_set() || !t2.time_is_set()) {
      throw Invalid_State("X509_Time: Cannot compare empty times");
   }

   return std::tie(t1.m_year, t1.m_month, t1.m_day, t1.m_hour, t1.m_minute, t1.m_second) <=>
          std::tie(t2.m_year, t2.m_month, t2.m_day, t2.m_hour, t2.m_minute, t2.m_second);
}
```

... by the way, something that wasn't clear to me before: if the spaceship has **a custom implementation** (i.e. not `= default`), the compiler **does not map equality** onto the spaceship. Hence the explicit definition of `operator==`. In-equality is then implicitly mapped to the explicitly defined `operator==`.

src/lib/asn1/asn1_time.cpp
```diff
@@ -162,9 +162,9 @@ int32_t ASN1_Time::cmp(const ASN1_Time& other) const {
    return SAME_TIME;
 }
```
reneme, line 163: If we go the spaceship-route, we could replace this with:

```C++
int32_t X509_Time::cmp(const X509_Time& other) const {
   const auto result = *this <=> other;

   constexpr int32_t EARLIER = -1;
   constexpr int32_t LATER = 1;
   constexpr int32_t SAME_TIME = 0;

   if(result == std::strong_ordering::less) {
      return EARLIER;
   }
   if(result == std::strong_ordering::greater) {
      return LATER;
   }

   return SAME_TIME;
}
```

reneme (commented):
Some nits, and a suggestion to modernize using `operator<=>` which might as well move into a follow-up. I tried this locally and could just open a PR with it after this is integrated.

configure.py
randombit: Ah indeed that works - I just used the preexisting header guard for this

src/lib/asn1/asn1_time.h
randombit: TBH the reason for the rename here was because the class is for some historical? reason referred to as `X509_Time` everywhere already *and* I for some reason was thinking that it's not possible to typedef a forward-declared type, while of course that actually works fine. But the date types are ASN.1 types just like any other so imo `ASN1_Time` is somewhat a clearer name. I reverted the rename and just added a typedef in `asn1_obj.h`

src/lib/asn1/asn1_time.h
randombit: Spaceship is fine but should be a distinct PR

reneme (approved):
I'll take care of the spaceship once this merges. 
# 👾

PR #5284, opened 2026-02-04
Avoid <chrono> in rng.h

randombit:
It was only used for setting a timeout for entropy polling, but this functionality was already slated for removal since - as of 3.0 - any potentially slow entropy sources have been removed anyway. The use of the chrono types was added in C++11 at which point <chrono> was very cheap to compile. Unfortunately C++20 adds sufficient amount of (whatever) to this header that it takes over a second to compile, each time it is used.

This change leaves reseed() available for applications to call, but intances of RandomNumberGenerator must instead implement reseed_from_sources which omits the timeout argument. This change will break (some) applications which derive their own RandomNumberGenerator instance; this derivation is not covered by our current SemVer policy.

### DIFF
--- a/src/lib/rng/rng.cpp
+++ b/src/lib/rng/rng.cpp
@@ -48,12 +48,12 @@ void RandomNumberGenerator::randomize_with_ts_input(std::span<uint8_t> output) {
    }
 }
 
-size_t RandomNumberGenerator::reseed(Entropy_Sources& srcs, size_t poll_bits, std::chrono::milliseconds poll_timeout) {
+size_t RandomNumberGenerator::reseed_from_sources(Entropy_Sources& srcs, size_t poll_bits) {
    if(this->accepts_input()) {
 #if defined(BOTAN_HAS_ENTROPY_SOURCE)
-      return srcs.poll(*this, poll_bits, poll_timeout);
+      return srcs.poll(*this, poll_bits);
 #else
-      BOTAN_UNUSED(srcs, poll_bits, poll_timeout);
+      BOTAN_UNUSED(srcs, poll_bits);
 #endif
    }
 
--- a/src/lib/rng/rng.h
+++ b/src/lib/rng/rng.h
@@ -14,11 +14,21 @@
 #include <botan/secmem.h>
 
 #include <array>
-#include <chrono>
 #include <concepts>
 #include <span>
 #include <string>
 #include <type_traits>
+
+/*
+* We only include <chrono> in downstream applications to avoid
+* breaking semver wrt RandomNumberGenerator::reseed. Within the
+* library we avoid it because it slows down compilation significantly.
+*
+* TODO(Botan4): remove this entirely
+*/
+#if !defined(BOTAN_IS_BEING_BUILT)
+   #include <chrono>
+#endif
 
 namespace Botan {
 
@@ -39,11 +49,6 @@ class BOTAN_PUBLIC_API(2, 0) RandomNumberGenerator {
       * Number of entropy bits polled for reseeding userspace RNGs like HMAC_DRBG
       */
       static constexpr size_t DefaultPollBits = 256;
-
-      /**
-      * Default poll timeout
-      */
-      static constexpr auto DefaultPollTimeout = std::chrono::milliseconds(50);
 
       virtual ~RandomNumberGenerator() = default;
 
@@ -166,18 +171,15 @@ class BOTAN_PUBLIC_API(2, 0) RandomNumberGenerator {
       virtual bool is_seeded() const = 0;
 
       /**
-      * Poll provided sources for up to poll_bits bits of entropy
-      * or until the timeout expires. Returns estimate of the number
-      * of bits collected.
-      *
+      * Poll provided sources for up to poll_bits bits of entropy.
+      * Returns estimate of the number of bits collected.
       * Sets the seeded state to true if enough entropy was added.
       *
-      * TODO(Botan4) this doesn't need to be virtual anymore
-      * TODO(Botan4) remove poll timeout argument entirely (polling is always fast now)
+      * @throws Exception if RNG accepts input but reseeding failed.
       */
-      virtual size_t reseed(Entropy_Sources& srcs,
-                            size_t poll_bits = RandomNumberGenerator::DefaultPollBits,
-                            std::chrono::milliseconds poll_timeout = RandomNumberGenerator::DefaultPollTimeout);
+      size_t reseed_from(Entropy_Sources& srcs, size_t poll_bits = RandomNumberGenerator::DefaultPollBits) {
+         return reseed_from_sources(srcs, poll_bits);
+      }
 
       /**
       * Reseed by reading specified bits from the RNG
@@ -185,11 +187,10 @@ class BOTAN_PUBLIC_API(2, 0) RandomNumberGenerator {
       * Sets the seeded state to true if enough entropy was added.
       *
       * @throws Exception if RNG accepts input but reseeding failed.
-      *
-      * TODO(Botan4) this does not need to be virtual
       */
-      virtual void reseed_from_rng(RandomNumberGenerator& rng,
-                                   size_t poll_bits = RandomNumberGenerator::DefaultPollBits);
+      void reseed_from(RandomNumberGenerator& rng, size_t poll_bits = RandomNumberGenerator::DefaultPollBits) {
+         return reseed_from_rng(rng, poll_bits);
+      }
 
       // Some utility functions built on the interface above:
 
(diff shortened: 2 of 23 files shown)


src/lib/rng/rng.h
```diff
@@ -165,28 +163,42 @@ class BOTAN_PUBLIC_API(2, 0) RandomNumberGenerator {
...
+      */
+      static constexpr auto DefaultPollTimeout = std::chrono::milliseconds(50);
+
+      /**
+       * Poll provided sources for up to poll_bits bits of entropy.
+       * Returns estimate of the number of bits collected.
+       *
+       * Sets the seeded state to true if enough entropy was added.
+       */
+      BOTAN_DEPRECATED("Use reseed_from_sources")
```
reneme: `TODO(Botan4): remove this`?

src/lib/rng/rng.h
```diff
@@ -14,12 +14,15 @@
...
 #include <array>
-#include <chrono>
 #include <concepts>
 #include <span>
 #include <string>
 #include <type_traits>
 
+#if !defined(BOTAN_IS_BEING_BUILT)
+   #include <chrono>
+#endif
```
reneme, line 31:
```suggestion
// We only include <chrono> in downstream applications to avoid
// breaking semver. Within the library we avoid it because it slows down
// compilation significantly.
//
// TODO(Botan4): remove this entirely
#if !defined(BOTAN_IS_BEING_BUILT)
   #include <chrono>
#endif
```

src/lib/rng/rng.h
```diff
@@ -165,28 +163,42 @@ class BOTAN_PUBLIC_API(2, 0) RandomNumberGenerator {
...
+       * Returns estimate of the number of bits collected.
+       *
+       * Sets the seeded state to true if enough entropy was added.
+       */
+      BOTAN_DEPRECATED("Use reseed_from_sources")
+      size_t reseed(Entropy_Sources& srcs,
+                    size_t poll_bits = RandomNumberGenerator::DefaultPollBits,
+                    std::chrono::milliseconds /*unused_timeout*/ = DefaultPollTimeout) {
+         return reseed_from_sources(srcs, poll_bits);
+      }
```
reneme: I understand that this method should only be available to downstream applications (and won't ever be part of a shared object for instance). I feel a bit uneasy about ABI incompatibility issues with this. But given that its only defined in the header, I concur that it should actually never become an issue. Perhaps define it as `inline` to be doubly sure?


```suggestion
      inline size_t reseed(Entropy_Sources& srcs,
                    size_t poll_bits = RandomNumberGenerator::DefaultPollBits,
                    std::chrono::milliseconds /*unused_timeout*/ = DefaultPollTimeout) {
         return reseed_from_sources(srcs, poll_bits);
      }
```

src/tests/test_cmce.cpp
```diff
@@ -21,6 +21,7 @@
    #include <botan/internal/cmce_gf.h>
    #include <botan/internal/cmce_parameters.h>
    #include <botan/internal/cmce_poly.h>
+   #include <algorithm>
```
reneme, line 24: Oh, so `<chrono>` pulls in `<algorithm>`? Maybe that would explain why it makes the compiler sweat so much.

src/lib/rng/rng.cpp
```diff
@@ -48,12 +48,12 @@ void RandomNumberGenerator::randomize_with_ts_input(std::span<uint8_t> output) {
    }
 }
 
-size_t RandomNumberGenerator::reseed(Entropy_Sources& srcs, size_t poll_bits, std::chrono::milliseconds poll_timeout) {
+size_t RandomNumberGenerator::reseed_from_sources(Entropy_Sources& srcs, size_t poll_bits) {
    if(this->accepts_input()) {
 #if defined(BOTAN_HAS_ENTROPY_SOURCE)
-      return srcs.poll(*this, poll_bits, poll_timeout);
+      return srcs.poll(*this, poll_bits, std::chrono::milliseconds(50));
```
reneme:
```suggestion
      // All potentially-slow sources were removed, this hard-coded timeout
      // is just keeping API promises until Botan4.
      return srcs.poll(*this, poll_bits, std::chrono::milliseconds(50));
```

src/lib/rng/rng.h
```diff
@@ -165,28 +163,42 @@ class BOTAN_PUBLIC_API(2, 0) RandomNumberGenerator {
...
       *
-      * TODO(Botan4) this doesn't need to be virtual anymore
-      * TODO(Botan4) remove poll timeout argument entirely (polling is always fast now)
+      * @throws Exception if RNG accepts input but reseeding failed.
       */
-      virtual size_t reseed(Entropy_Sources& srcs,
-                            size_t poll_bits = RandomNumberGenerator::DefaultPollBits,
-                            std::chrono::milliseconds poll_timeout = RandomNumberGenerator::DefaultPollTimeout);
+      virtual size_t reseed_from_sources(Entropy_Sources& srcs,
+                                         size_t poll_bits = RandomNumberGenerator::DefaultPollBits);
```
reneme: Super-nitty: perhaps `reseed_from()` only?

Usage would then look like this: `rng->reseed_from(my_fancy_sources)` which usually reads better and makes the method name shorter.

reneme (approved):

src/tests/test_cmce.cpp
randombit: That's part of it but `chrono` is apparently huge/complicated in C++20

```
$ echo "#include <chrono>\n" | time clang++ -std=c++20 -O3 -x c++ -c - -o /dev/null
0.79s user 0.03s system 99% cpu 0.821 total
$ echo "#include <algorithm>\n" | time clang++ -std=c++20 -O3 -x c++ -c - -o /dev/null
0.11s user 0.02s system 99% cpu 0.131 total
$ echo "#include <iostream>\n" | time clang++ -std=c++20 -O3 -x c++ -c - -o /dev/null
0.23s user 0.02s system 99% cpu 0.254 total
```

whereas in C++17 it was cheap

```
echo "#include <chrono>\n" | time clang++ -std=c++17 -O3 -x c++ -c - -o /dev/null
0.04s user 0.01s system 98% cpu 0.056 total
```

reneme: Quite unfortunate. Good to know, though. ... waiting for "modules" to finally fix all of this. ⏳

src/lib/rng/rng.h
randombit: It's definitely a bit nasty and I think strictly speaking an ODR violation. I think in practice it's safe enough as a non-virtual inline.

If it ends up causing problems I'll declare a force majeure and remove the timeout argument entirely, SemVer or not.

src/lib/rng/rng.h
randombit: One can't have virtuals with the same name, I think it would be better to add two (non-virtual) overloads `reseed_from` handling both entropy sources and rngs.

(done)

src/lib/rng/rng.cpp
randombit: Did one better and added another `Entropy_Source::poll` without a timeout argument, deprecated the one with a timeout, and called that variant here.

src/lib/rng/rng.h
reneme: Fair point. But given that this is just being introduced, we could also add it (Edit: the virtual `reseed_from_sources`) as a private member and keep it more flexible for future changes.

src/lib/rng/rng.h
```diff
@@ -166,33 +171,49 @@ class BOTAN_PUBLIC_API(2, 0) RandomNumberGenerator {
...
       * Reseed by reading specified bits from the RNG
       *
       * Sets the seeded state to true if enough entropy was added.
       *
       * @throws Exception if RNG accepts input but reseeding failed.
-      *
-      * TODO(Botan4) this does not need to be virtual
       */
       virtual void reseed_from_rng(RandomNumberGenerator& rng,
                                    size_t poll_bits = RandomNumberGenerator::DefaultPollBits);
```
reneme: [Similarly to the `reseed_from_sources()`](https://github.com/randombit/botan/pull/5284#discussion_r2769874703), this should probably be deprecated with a reference to the new `reseed_from` that is not virtual. Edit: In Botan4 it could then be moved out of the public interface of the RNG.

src/lib/rng/rng.h
randombit: Yeah good point

randombit: Now protected

src/lib/rng/rng.h
randombit: Makes sense I will leave to a followup tho

PR #5294, opened 2026-02-07
Split part of concepts.h into range_concepts.h

randombit:
Due to <ranges> this header is quite expensive to compile, and the ranges parts aren't needed by many users of concepts.h - and in fact isn't needed at all in the public API outside of the (deprecated for public use) mem_ops.h

Move the parts of concepts.h related to strong types to that header.

Remove the opportunistic exposure of checked accessors in strong type; it wasn't used outside of tests and the concept is somewhat complicated.

### DIFF
--- a/src/lib/utils/info.txt
+++ b/src/lib/utils/info.txt
@@ -6,9 +6,9 @@ brief -> "Various utility functions and types"
 load_on always
 
 <header:public>
+allocator.h
 api.h
 assert.h
-allocator.h
 compiler.h
 concepts.h
 data_src.h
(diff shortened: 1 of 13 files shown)


src/lib/utils/info.txt
```diff
@@ -6,18 +6,20 @@ brief -> "Various utility functions and types"
...
 compiler.h
 concepts.h
 data_src.h
 database.h
 exceptn.h
 mem_ops.h
 mutex.h
-types.h
+# TODO(Botan4) move this to internal
+range_concepts.h
```
reneme, line 20: Does it make sense to add the "future internal header" macro to this header then?

reneme (approved):

src/lib/utils/info.txt
randombit: Yes it does

PR #5296, opened 2026-02-07
Various changes to reduce header dependencies in TLS

randombit:
This cuts about 50 CPU-seconds off the time my machine needs to build TLS

### DIFF
--- a/src/lib/tls/tls12/msg_session_ticket_12.cpp
+++ b/src/lib/tls/tls12/msg_session_ticket_12.cpp
@@ -16,7 +16,7 @@ namespace Botan::TLS {
 New_Session_Ticket_12::New_Session_Ticket_12(Handshake_IO& io,
                                              Handshake_Hash& hash,
                                              Session_Ticket ticket,
-                                             std::chrono::seconds lifetime) :
+                                             uint32_t lifetime) :
       m_ticket_lifetime_hint(lifetime), m_ticket(std::move(ticket)) {
    hash.update(io.send(*this));
 }
@@ -32,14 +32,13 @@ New_Session_Ticket_12::New_Session_Ticket_12(const std::vector<uint8_t>& buf) {
 
    TLS_Data_Reader reader("SessionTicket", buf);
 
-   m_ticket_lifetime_hint = std::chrono::seconds(reader.get_uint32_t());
+   m_ticket_lifetime_hint = reader.get_uint32_t();
    m_ticket = Session_Ticket(reader.get_range<uint8_t>(2, 0, 65535));
    reader.assert_done();
 }
 
 std::vector<uint8_t> New_Session_Ticket_12::serialize() const {
-   std::vector<uint8_t> buf(4);
-   store_be(static_cast<uint32_t>(m_ticket_lifetime_hint.count()), buf.data());
+   auto buf = store_be<std::vector<uint8_t>>(m_ticket_lifetime_hint);
    append_tls_length_value(buf, m_ticket.get(), 2);
    return buf;
 }
--- a/src/lib/tls/tls12/tls_messages_12.h
+++ b/src/lib/tls/tls12/tls_messages_12.h
@@ -11,6 +11,12 @@
 
 #include <botan/tls_messages.h>
 
+namespace Botan {
+
+class PK_Key_Agreement_Key;
+
+}
+
 namespace Botan::TLS {
 
 /**
@@ -53,19 +59,21 @@ class BOTAN_UNSTABLE_API Client_Key_Exchange final : public Handshake_Message {
 /**
 * Certificate Message of TLS 1.2
 */
-class BOTAN_UNSTABLE_API Certificate_12 final : public Handshake_Message {
+class BOTAN_UNSTABLE_API Certificate_12 final : public Handshake_Message /* NOLINT(*-special-member-functions) */ {
    public:
       Handshake_Type type() const override { return Handshake_Type::Certificate; }
 
       const std::vector<X509_Certificate>& cert_chain() const { return m_certs; }
 
-      size_t count() const { return m_certs.size(); }
+      size_t count() const;
 
       bool empty() const { return m_certs.empty(); }
 
       Certificate_12(Handshake_IO& io, Handshake_Hash& hash, const std::vector<X509_Certificate>& certs);
 
       Certificate_12(const std::vector<uint8_t>& buf, const Policy& policy);
+
+      ~Certificate_12() override;
 
       std::vector<uint8_t> serialize() const override;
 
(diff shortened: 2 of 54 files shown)


src/lib/x509/x509cert.h
```diff
@@ -19,15 +19,6 @@ class NameConstraints;
 class Public_Key;
 class X509_DN;
```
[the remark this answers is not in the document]
randombit: Included indirectly already

src/lib/tls/tls12/msg_session_ticket_12.cpp
```diff
@@ -32,14 +32,14 @@ New_Session_Ticket_12::New_Session_Ticket_12(const std::vector<uint8_t>& buf) {
...
-   m_ticket_lifetime_hint = std::chrono::seconds(reader.get_uint32_t());
+   m_ticket_lifetime_hint = reader.get_uint32_t();
    m_ticket = Session_Ticket(reader.get_range<uint8_t>(2, 0, 65535));
    reader.assert_done();
 }
 
 std::vector<uint8_t> New_Session_Ticket_12::serialize() const {
    std::vector<uint8_t> buf(4);
-   store_be(static_cast<uint32_t>(m_ticket_lifetime_hint.count()), buf.data());
+   store_be(m_ticket_lifetime_hint, buf.data());
```
reneme: FWIW: This could be slightly simplified now:

```C++
auto buf = store_be<std::vector<uint8_t>>(m_ticket_lifetime_hint);
```

src/lib/tls/tls12/tls_messages_12.h
```diff
@@ -237,7 +259,7 @@ class BOTAN_UNSTABLE_API New_Session_Ticket_12 final : public Handshake_Message
       std::vector<uint8_t> serialize() const override;
 
    private:
-      std::chrono::seconds m_ticket_lifetime_hint{};
+      uint32_t m_ticket_lifetime_hint = 0;
```
reneme, line 262: It really stinks that `<chrono>` is so heavy. Ditching `::seconds` (and friends) as strong types does reduce the expressiveness of the code. :( ... it's what it is, I guess.

src/lib/tls/tls13/tls_messages_13.h
```diff
@@ -290,7 +293,7 @@ class BOTAN_UNSTABLE_API Certificate_Request_13 final : public Handshake_Message
       const std::vector<uint8_t>& context() const { return m_context; }
 
    private:
-      Certificate_Request_13(std::vector<X509_DN> acceptable_CAs, const Policy& policy, Callbacks& callbacks);
+      Certificate_Request_13(const std::vector<X509_DN>& acceptable_CAs, const Policy& policy, Callbacks& callbacks);
```
reneme, line 296: Does it really hurt so much to pull in `X509_DN` ("pkix_types.h") here? In `::maybe_create` this actually uses this lvalue to avoid a copy. Not that it is a particularly hot path in the code, but I'm wondering about the general rule-of-thumb for the trade-off between compile time and run time efficiency.

Similarly in `Certificate_Authorities`.

src/lib/tls/tls12/tls_messages_12.h
randombit: I know, it's incredibly frustrating when the basic time unit types were *so* useful as interface types and then the committee decided to throw in so much stuff in to the header that they are no longer viable for that purpose.

IMO C++ is failing badly at both "You don't pay for what you don't use" and "Leave no room for a lower-level language" and the worst part is from all visible signs the committee thinks it's going great.

src/lib/tls/tls13/tls_messages_13.h
randombit: I guess it's a matter of slow cuts. Adding `pkix_types.h` slows down compilation of `tls_messages.h` by about 8% (.71s->.76s for GCC, 1.0s->1.07s for Clang) which is not so bad. But `tls_messages.h` is then included into 40 or 50 TUs. So making it even slightly more expensive balloons out.

Where *exactly* the line is drawn is certainly open for debate but broadly

* If something is included into many TUs it should be as light as possible. It doesn't matter if any single particular source file is slow. For example kuznyechik.o is one of the very slowest object files in the entire library to compile, due to fancy `consteval` tricks, and do we really need to be doing those, no - but it's just one file and not hurting anything else so who cares. Whereas eg `concepts.h` or `loadstor.h` being even 1ms faster to parse cuts 10s of seconds off the build. Also being common headers they are an ongoing tax; when another source file is added which includes them, the overall compilation cost of these headers grows further.

* If the performance doesn't matter much (for example, if something could cost twice as much and nobody would ever notice) then might as well prefer saving compilation time, since compilation time is a scarce resource of uncertain availability [1] and that resource is shared across the entire codebase.

[1]: In that there is no fixed hard budget but slower compilation time hurts developer velocity and likely as not hurts adoption if it's bad enough to notice.

In this specific case, I'd say: many TUs are affected, it's not really performance critical (only happens during the handshake and the handshake is anyway expensive), so prefer saving compilation time. I guess if we had profiles for the TLS handshake performance [which would be a useful benchmark to have!!] and the extra copy was showing up somehow, that would be worth working on.

As an example going the other way, headers like `ct_utils.h` and `simd_4x32.h` are somewhat expensive, about .25-.35 seconds to compile, and used everywhere. But these are run in hot loops that are written expecting values like `CT::Mask` and `SIMD_4x32` will be perfectly inlined, it's difficult to reduce the compilation time meaningfully without drastically impacting performance.

src/lib/tls/tls12/tls_messages_12.h
reneme: #JustUseModules is the answer, I guess...

The way I see it, You don't pay for what you don't use" only accounts for the final binary. And there it delivers in most cases. But compile times are a significant issue if you don't happen to have a 64 core beast in your basement.

src/lib/tls/tls13/tls_messages_13.h
reneme: Thanks for the guidelines. I concur in general terms. Though, still I feel sad to introduce runtime efficiencies like that. But from a gut feeling I also don't think this particular copy would produce measurable runtime slowdowns. On that note, we're starting to use the TLS stack in a system that might actually experience some significant load. Therefore, benchmarks for TLS might come up on my TODO list eventually...

Anyway: with all the header patrol work you've put in, I think it would be really cool to have an automated compile benchmark in CI to catch regressions early and weigh the pros and cons of new API designs early on. (For instance the new `loadstor.h` is convenient but I do question whether it was really necessary to push the limits that far. 😶)

src/lib/tls/tls12/tls_messages_12.h
randombit: > The way I see it, You don't pay for what you don't use" only accounts for the final binary.

That's clearly their interpretation as well. Unfortunately that interpretation assumes developer wait time is free, which is not necessarily accurate.

src/lib/tls/tls13/tls_messages_13.h
randombit: > I think it would be really cool to have an automated compile benchmark in CI to catch regressions early and weigh the pros and cons of new API designs early on.

Yes I have something specifically like this in mind to include in the nightly run (and can be invoked on demand if a PR seems risky in this way) - Clang's `-ftime-trace` dumps a lot of useful information.

PR #5298, opened 2026-02-08
Add support for precompiled headers

randombit:
#5164 

Stacked on #5297 

Tried at MSVC support but it's a quite different model and I got tired of going back and forth with CI. Left to an interested party, hopefully one with a local Windows machine for faster iterations.

On my laptop, building with Clang 21:

* 3.10.0 release:  `3973.09s user 70.02s system 770% cpu 8:44.74 total`
* `master` plus #5294 + #5296: `3238.27s user 95.83s system 770% cpu 7:12.99 total`
* above plus PCH: `2070.82s user 58.87s system 772% cpu 4:35.66 total`

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -250,6 +250,7 @@ class BuildPaths:
         self.libobj_dir = os.path.join(self.build_dir, 'obj', 'lib')
         self.cliobj_dir = os.path.join(self.build_dir, 'obj', 'cli')
         self.testobj_dir = os.path.join(self.build_dir, 'obj', 'test')
+        self.pch_dir = os.path.join(self.build_dir, 'obj', 'pch') if options.enable_pch else None
 
         self.doc_output_dir = os.path.join(self.build_dir, 'docs')
         self.handbook_output_dir = os.path.join(self.doc_output_dir, 'handbook')
@@ -327,6 +328,8 @@ class BuildPaths:
             out += [self.fuzzobj_dir, self.fuzzer_output_dir]
         if self.example_output_dir:
             out += [self.example_obj_dir, self.example_output_dir]
+        if self.pch_dir:
+            out += [self.pch_dir]
         return out
 
     def format_public_include_flags(self, cc):
@@ -563,6 +566,8 @@ def process_command_line(args):
     build_group.add_option('--disable-cc-tests', dest='enable_cc_tests',
                            default=True, action='store_false',
                            help=optparse.SUPPRESS_HELP)
+
+    add_enable_disable_pair(build_group, 'pch', True, 'disable precompiled headers')
 
     add_with_without_pair(build_group, 'valgrind', False, 'use valgrind API')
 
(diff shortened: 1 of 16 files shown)


randombit:
I experimented a bit with including our own headers also into the PCH and the improvement was at best minor and often regressed vs just using standard library headers. I also ran into some situations where a PCH had an old version of something from Botan itself and the resulting error messages were very hard to interpret, so I'm inclined to avoid unless it really helps.

randombit:
>    error: default visibility for functions and variables [-fvisibility] differs in PCH file vs. current file

Seems like older Clang doesn't like this? Clang 21 is fine.

Unfortunate, probably suggests there needs to be 2 PCHs one for the library and the other for tests/cli

randombit:
Mac build issue

```
   In file included from <built-in>:1:
  In file included from ./build/obj/pch/pch.h:41:
  /Applications/Xcode_16.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/16/include/immintrin.h:14:2: error: "This header is only meant to be used on x86 and x64 architecture"
     14 | #error "This header is only meant to be used on x86 and x64 architecture"
        |  ^
```

Then why ship it you doofs

randombit:
Wow GCC also being stupid about making `immintrin.h` available on non-x86 builds

```
  In file included from /usr/lib/gcc-cross/powerpc64le-linux-gnu/13/include/immintrin.h:29,
                   from ././build/obj/pch/pch.h:41,
                   from <command-line>:
  /usr/lib/gcc-cross/powerpc64le-linux-gnu/13/include/mmintrin.h:52:2: error: #error "Please read comment above.  Use -DNO_WARN_X86_INTRINSICS to disable this error."
     52 | #error "Please read comment above.  Use -DNO_WARN_X86_INTRINSICS to disable this error."
        |  ^~~~~
```

randombit:
Examples can't use the existing PCH either:

>  cc1plus: error: ././build/obj/pch/pch.h.gch: not used because `BOTAN_IS_BEING_BUILT' not defined [-Werror=invalid-pch]

Possibly need 3 then

* Library flags
* Test/cli flags
* Examples

Or more likely just skip on PCH for the examples, they are small and the link time dominates anyway.

randombit:
`immintrin.h` might not be worth it, only difference between `pch_lib.h` and `pch_exe.h` is inclusion of that header and the size difference is crazy:

Clang 21: exe 12 Mb, lib 17 Mb
GCC 15: exe 69 Mb, lib 99 Mb (yikes)

randombit:
@copilot Try adding PCH support to the ninja build

randombit:
Weird error right in the middle of a build

```
  fatal error: file '/home/runner/work/botan/botan/build/obj/pch/pch_exe.h' has been modified since the precompiled header './build/obj/pch/pch_exe.h.pch' was built: mtime changed (was 1770567362, now 1770569364)
  note: please rebuild precompiled header './build/obj/pch/pch_exe.h.pch'
  1 error generated.
  make: *** [Makefile:1864: build/obj/cli/utils.o] Error 1
```

Not sure what to make of this

Possibly an interaction with ccache. The ccache docs say re PCH:  If you use Clang, you must compile with -fno-pch-timestamp.

randombit:
@copilot OK now try to extend to support Microsoft Visual C++ PCH

IIUC you'll have to modify configure to write some little stub cpp files for MSVC to compile to create the PCH

IIUC the relevant options are `/Yc`, `/Yu`, and `/FI`.

randombit:
Needs a squash but should otherwise be good to go

PR #5300, opened 2026-02-08
IWYU - Remove unnecessary header includes to improve build performance

KaganCanSit:
Hello,

This PR addresses the removal of unnecessary header includes detected by the IWYU (Include What You Use) tool. False positives were manually filtered out from the report generated by the following command, and only the definitely unnecessary includes that I could identify were removed:

```bash
iwyu_tool -p build -j $(nproc) -- -Xiwyu --no_comments >> include_report.txt
```

It's great to see that the steps you've taken recently to reduce build time are making a difference. I should mention that even without using `ccache` at the moment, I'm able to get builds in a shorter time. Hopefully, these changes will help.

Notes:
* On Linux, I can successfully compile and pass tests with GCC and Clang compilers without using `ccache`. However, I will review this in CI.
* Since the IWYU report contained many false positives, only confirmed unnecessary includes were included in this PR. I may create additional PRs as I discover more detailed findings.
* I am aware that there are points affecting the API, but I am not sure what to do. I can revert changes in specific files if you wish.

Please write your suggestions for changes or corrections in the comments, and I will be happy to help when I have time.

Best regards.

### DIFF
--- a/src/lib/tls/tls_session_manager_memory.h
+++ b/src/lib/tls/tls_session_manager_memory.h
@@ -9,7 +9,6 @@
 #ifndef BOTAN_TLS_SESSION_MANAGER_IN_MEMORY_H_
 #define BOTAN_TLS_SESSION_MANAGER_IN_MEMORY_H_
 
-#include <botan/mutex.h>
 #include <botan/tls_session.h>
 #include <botan/tls_session_manager.h>
 
--- a/src/lib/utils/os_utils/os_utils.cpp
+++ b/src/lib/utils/os_utils/os_utils.cpp
@@ -27,7 +27,6 @@
    #include <pthread.h>
    #include <setjmp.h>
    #include <signal.h>
-   #include <stdlib.h>
    #include <sys/mman.h>
    #include <sys/resource.h>
    #include <sys/types.h>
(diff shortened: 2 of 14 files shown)


src/lib/tls/tls_session_manager_memory.h
```diff
@@ -9,7 +9,6 @@
 #ifndef BOTAN_TLS_SESSION_MANAGER_IN_MEMORY_H_
 #define BOTAN_TLS_SESSION_MANAGER_IN_MEMORY_H_
 
-#include <botan/mutex.h>
```
KaganCanSit, line 12: I'm not sure, maybe it would be better to move the `.cpp` file?

src/lib/utils/os_utils/os_utils.cpp
```diff
@@ -27,7 +27,6 @@
    #include <pthread.h>
    #include <setjmp.h>
    #include <signal.h>
-   #include <stdlib.h>
```
randombit, line 30: This one I wasn't so sure about since we are relying on not just what is included per the standard but also various functions defined by various operating systems. But checking the history it seems this include was added in #1798 for `posix_memalign`, which was only used for a brief period. So I think it can also safely be removed now.

randombit (approved):
Looks great and CI seems happy as well, thanks!

KaganCanSit:
> Looks great and CI seems happy as well, thanks!

You're welcome. I'm glad it was helpful. I'll look into it more thoroughly when I find the time. 🫡

PR #5306, opened 2026-02-10
Allow unregistering EC Groups

### LINKED ISSUE #5283: Unregistering EC Goups
We're trying to migrate a rather low level key exchange from openssl to botan, and are currently doing this in our rust code:
```rs
let mut group = openssl::ec::EcGroup::from_components(p, a, b, &mut bnctx).unwrap();
group.set_generator(g, q, h).unwrap();
```
All the group parameters are specified externally, so this is potentially a different group every time.
botan-rs [provides](https://docs.rs/botan/latest/botan/struct.EcGroup.html#method.from_params) something like this as well (nevermind that you can't create a key with it currently, can be added easily), but it requires an OID, constructors without exist, but they are deprecated.
This leaves us needing to create a random OID every time, which is a little unergonomic.
More importantly though, after we're done we cannot remove the group (and thus free the OID), which will become a problem for long running processes. There is `EC_Group::clear_registered_curve_data`, but it is a) internal and b) it deletes everything, not just a single curve.

What is the recommended way forward here?

arckoor:
Closes #5283.

### DIFF
--- a/doc/api_ref/ffi.rst
+++ b/doc/api_ref/ffi.rst
@@ -967,6 +967,13 @@ EC Groups
 
    Initialize an EC Group from a common group name (eg "secp256r1")
 
+.. cpp:function:: int botan_ec_group_unregister(botan_asn1_oid_t oid)
+
+   Unregister a previously registered group. Returns 1 if the group was found and unregistered, else 0.
+
+   Using this is discouraged for normal use. This is only useful or necessary if
+   you are registering a very large number of distinct groups, and need to worry about memory constraints.
+
 .. cpp:function:: int botan_ec_group_view_der(botan_ec_group_t ec_group, botan_view_ctx ctx, botan_view_bin_fn view)
 
    View an EC Group in DER encoding
--- a/src/lib/pubkey/ec_group/ec_group.cpp
+++ b/src/lib/pubkey/ec_group/ec_group.cpp
@@ -35,6 +35,22 @@ class EC_Group_Data_Map final {
          return count;
       }
 
+      bool unregister(const OID& oid) {
+         // TODO(Botan4)
+         if(oid.empty()) {
+            throw Invalid_Argument("OID must not be empty");
+         }
+
+         const lock_guard_type<mutex_type> lock(m_mutex);
+         for(size_t i = 0; i < m_registered_curves.size(); i++) {
+            if(m_registered_curves[i]->oid() == oid) {
+               m_registered_curves.erase(m_registered_curves.begin() + i);
+               return true;
+            }
+         }
+         return false;
+      }
+
       std::shared_ptr<EC_Group_Data> lookup(const OID& oid) {
          const lock_guard_type<mutex_type> lock(m_mutex);
 
@@ -560,6 +576,11 @@ EC_Group::EC_Group(std::span<const uint8_t> ber) {
    auto data = BER_decode_EC_group(ber, EC_Group_Source::ExternalSource);
    m_data = data.first;
    m_explicit_encoding = data.second;
+}
+
+// static
+bool EC_Group::unregister(const OID& oid) {
+   return ec_group_data().unregister(oid);
 }
 
 const EC_Group_Data& EC_Group::data() const {
--- a/src/lib/pubkey/ec_group/ec_group.h
+++ b/src/lib/pubkey/ec_group/ec_group.h
@@ -205,6 +205,16 @@ class BOTAN_PUBLIC_API(2, 0) EC_Group final {
       */
       BOTAN_DEPRECATED("Deprecated no replacement") EC_Group();
 
+      /**
+      * Unregister a previously registered group.
+      *
+      * Using this is discouraged for normal use. This is only useful or necessary if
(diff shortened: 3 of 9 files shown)


doc/api_ref/ffi.rst
```diff
@@ -967,6 +967,10 @@ EC Groups
 
    Initialize an EC Group from a common group name (eg "secp256r1")
 
+.. cpp:function:: int botan_ec_group_unregister(botan_asn1_oid_t oid)
+
+   Unregister a previously registered group. Returns 1 if the group was found and unregistered, else 0.
```
randombit, line 972: There should be some text here indicating that this isn't something that one normally must do, the scenarios where it does make sense, ...

src/tests/test_ec_group.cpp
```diff
@@ -597,6 +598,34 @@ class EC_Group_Registration_Tests final : public Test {
...
+                                         secp256r1.get_g_y(),
+                                         secp256r1.get_order());
+
+         result.confirm("Internal group is unregistered", Botan::EC_Group::unregister(secp256r1_oid) == true);
+         result.confirm("Unregistering internal group again does nothing",
+                        Botan::EC_Group::unregister(secp256r1_oid) == false);
+         result.confirm("User defined group is unregistered", Botan::EC_Group::unregister(custom_oid) == true);
+         result.confirm("Unregistering user defined group again does nothing",
+                        Botan::EC_Group::unregister(custom_oid) == false);
+
```
randombit, line 663: I'd want to see more tests here. In particular I'd want there to be a test that a live `EC_Group` that references the custom parameters continues to be live/correct after a `unregister` call.

src/lib/pubkey/ec_group/ec_group.h
```diff
@@ -205,6 +205,12 @@ class BOTAN_PUBLIC_API(2, 0) EC_Group final {
       */
       BOTAN_DEPRECATED("Deprecated no replacement") EC_Group();
 
+      /**
+      * Unregister a previously registered group.
+      * Returns true if the group was found and unregistered.
```
randombit, line 214: Again some brief context here is good, mostly with the goal of steering away people from using this just because. Eg "This is only useful or necessary if .."

src/lib/pubkey/ec_group/ec_group.cpp
```diff
@@ -35,6 +35,17 @@ class EC_Group_Data_Map final {
...
+      bool unregister(const OID& oid) {
+         const lock_guard_type<mutex_type> lock(m_mutex);
+         for(size_t i = 0; i < m_registered_curves.size(); i++) {
+            if(m_registered_curves[i]->oid() == oid) {
+               m_registered_curves.erase(m_registered_curves.begin() + i);
+               return true;
+            }
+         }
+         return false;
+      }
```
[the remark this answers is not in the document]
randombit: Copilot is right! OID being internally nullable is an artifact of pre-`optional` days and kind of a nasty trap, maybe should be fixed for Botan4...

arckoor:
@randombit I've added some more checks, though I don't know if it's enough / what you hoped for. With my somewhat limited api knowledge I'm not really sure what else I could reasonably test :/

src/tests/test_ec_group.cpp
```diff
@@ -598,6 +602,57 @@ class EC_Group_Registration_Tests final : public Test {
...
+         result.test_is_true("Group has correct order parameter", reg_group.get_order() == secp256r1.get_order());
+
+   #if defined(BOTAN_HAS_ECDSA)
+         std::unique_ptr<Botan::RandomNumberGenerator> rng = Test::new_rng("test_ec_group_unregistration");
+         std::unique_ptr<Botan::Private_Key> key = Botan::create_ec_private_key("ECDSA", reg_group, *rng);
+         result.test_success("Can still create use group");
+         result.test_is_true("Key was created correctly", key->check_key(*rng, true));
+   #endif
+
+         try {
```
randombit, line 656: Add a `// TODO(Botan4) remove this when OIDs lose internal nullability` just so we don't forget it

src/tests/test_ec_group.cpp
```diff
@@ -598,6 +602,57 @@ class EC_Group_Registration_Tests final : public Test {
...
+                                         secp256r1.get_g_y(),
+                                         secp256r1.get_order());
+
+         result.test_is_true("Internal group is unregistered", Botan::EC_Group::unregister(secp256r1_oid));
+         result.test_is_false("Unregistering internal group again does nothing",
+                              Botan::EC_Group::unregister(secp256r1_oid));
+         result.test_is_true("User defined group is unregistered", Botan::EC_Group::unregister(custom_oid));
+         result.test_is_false("Unregistering user defined group again does nothing",
+                              Botan::EC_Group::unregister(custom_oid));
+
```
randombit, line 632: To expand a bit more on these tests I'd go with

* Before unregistering, get another copy of the group via OID
* Verify that after unregistering the group, `from_OID` fails
* Use the copy of the group retrieved via OID for the ECDSA test

randombit (approved):
LGTM thanks @arckoor

Issue #4666, opened 2025-02-10
Botan4


randombit:
Holding ticket

crazydef:
Completely removing modules will make people reconsider Botan as a reliable, stable library moving forward.  Myself included.

While I appreciate algorithms become less secure over time, many applications need to maintain compatibility with other systems that still use them - for whatever reason.  Being able to validate a CRC32 or MD5 hash, for example, that has been received from another application/system is still a valid use case.  (Or even using insecure algorithms in systems that do not require absolute security - such as file hashing for data storage in games, for example.)

May I humbly recommend that they're disabled from the builds by default, but can still be re-enabled for those who really need them?

randombit:
> Completely removing modules will make people reconsider Botan as a reliable, stable library moving forward.

By clearly documenting "X is deprecated and may be removed in a future major release" and then, years later, removing X in a future major release?

You make an argument for CRC32 and MD5. Which are pretty plausible I guess. Can you make a similar statement about Lion, Noekeon, or HyMES McEliece? How about other algorithms which similarly have been added and later removed like Blue Midnight Wish, DESX, Square, Skipjack, FORK-256, CS-Cipher, SHARK, ThreeWay, Rabin-Williams, EMAC, MISTY1, Kasumi, or CAST-256?

Already the cost of removing a feature is high; once something is added we have to maintain it for years and years, at least until a new major release. If the cost of removing a feature is literally infinite - once added we are stuck with it forever - then the bar to *adding* an algorithm becomes equally high. For example the preliminary versions of Kyber and Dilithium would not have been merged, if it was impossible to remove them later. Instead we would have waited (years!) for the final NIST competition results. And only *then* would we have been able to start work on post quantum secure TLS.

If there is a specific algorithm (say CRC32 or MD5) that you wish to argue against removal please do open a subticket and we can discuss. I am _very_ open to arguments of the form "I am currently using, or might in the future want to use, (this specific thing) and removing it will be an inconvenience for me." (see for example #4678) I am however not interested in an argument of the form "You cannot remove anything ever for any reason." If you have been operating under the impression that - in a major release - we would never remove some number of {experimental,insecure,obsolete,obscure} algorithms, after a clearly documented deprecation phase, then I regret to inform you that this may not be the library for you.

> that they're disabled from the builds by default, but can still be re-enabled for those who really need them?

[...]

PR #5307, opened 2026-02-10
Fix ML-DSA private key encoding

### LINKED ISSUE #5002: ML-DSA private key encoding format has changed
Starting with [**`draft-ietf-lamps-dilithium-certificates-09`**](https://datatracker.ietf.org/doc/html/draft-ietf-lamps-dilithium-certificates-09#name-private-key-format) dated May 7 2025, the ML-DSA private key format changed to
```
ML-DSA-44-PrivateKey ::= CHOICE {
  seed [0] OCTET STRING (SIZE (32)),
  expandedKey OCTET STRING (SIZE (2560)),
  both SEQUENCE {
      seed OCTET STRING (SIZE (32)),
      expandedKey OCTET STRING (SIZE (2560))
      }
  }

ML-DSA-65-PrivateKey ::= CHOICE {
  seed [0] OCTET STRING (SIZE (32)),
  expandedKey OCTET STRING (SIZE (4032)),
  both SEQUENCE {
      seed OCTET STRING (SIZE (32)),
      expandedKey OCTET STRING (SIZE (4032))
      }
  }

ML-DSA-87-PrivateKey ::= CHOICE {
  seed [0] OCTET STRING (SIZE (32)),
  expandedKey OCTET STRING (SIZE (4896)),
  both SEQUENCE {
      seed OCTET STRING (SIZE (32)),
      expandedKey OCTET STRING (SIZE (4896))
      }
  }
```
Thus since Botan currently supports the keyseed format only, an **`ASN.1 CONTEXT 0`**  tag has to be inserted after the **`ASN.1 OCTET STRING`** tag of the PKCS#8 private key encoding. When will this change of the encoding be supported by Botan library?

falko-strenzke:
This updates the ML-DSA [private key format to RFC 9881](https://www.rfc-editor.org/rfc/rfc9881.html#name-example-private-keys). 

The implementation now by default writes the "seed-only" format. If the generative seed is not contained in the private key, then it encodes in the "expanded-only" format. It can read all three formats. In case of the "both" format, the key expanded from the seed and the encoded expanded key are compared. If they differ, a Decoding_Error is thrown.

Some points to discuss:

One real problem of this fix is that it destroys backwards compatibility for those users who relied on the private key being the raw seed. Even in the "seed-only" format, now the encoded private key is wrapped into an ASN.1/DER TLV structure. Unfortunately, Botan provides no possibility to distinguish the context in which a key is encoded. It knows only a single format, which happended to be the tangible raw seed format previously and thus was tempting for downstream users to rely on it to provide the seed. The clean solution would be to have the encoding layer independent from the cryptographic keys and always force the user to explicitly choose an encoding context. In any case, a fundamental solution for the problem is out of reach of this PR. I see no other way than breaking backwards compatibility now.

For compatibility with the DilithiumRound3 and previous ML-DSA draft, the private key decoding in this implementation also still accepts a raw seed, which is not specified by RFC 9881. We could remove this for the decoding of ML-DSA keys but leave it for the DilithiumRound3 keys. Note that OpenSSL also still supports the raw seed for ML-DSA.

~~A minor point with the change of the format, the existing ML-DSA KAT tests show errors because the hash of the private key is not as expected. I circumvented this by checking the private key hash only for the Dilithium KAT tests and skipping it for the ML-DSA tests. This should be fine since the key expansion is effectively checked via the Dilithium KAT tests.~~ -- Update Falko 2026-07-30: test works now as before.

Fixes #5002.


falko-strenzke:
@randombit @reneme 
The crucial point to decide is whether this PR brings breaking change in the ML-DSA private key encoding. Downstream users who relied on public_key_bits() to return the raw seed will face errors. However, in my view, it is not a breaking change, since the function documentation says "Returns: BER encoded private key bits". Thus downstream users were already required to query the seed.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -6,22 +6,118 @@
...
+#include "botan/exceptn.h"
+#include "botan/internal/dilithium_types.h"
 #include <botan/internal/ml_dsa_impl.h>
 
 #include <botan/internal/dilithium_algos.h>
 
+#include <botan/ber_dec.h>
+#include <botan/der_enc.h>
+#include <string>
+#include <utility>
```
reneme:
```suggestion
#include <botan/internal/ml_dsa_impl.h>

#include <botan/asn1_obj.h>
#include <botan/ber_dec.h>
#include <botan/der_enc.h>
#include <botan/exceptn.h>
#include <botan/internal/dilithium_algos.h>
#include <botan/internal/dilithium_types.h>
#include <string>
#include <utility>
```

The general rule-of-thumb regarding includes:

1. The principal header (what this is meant to directly implement) comes first, then a newline is added
2. All other includes follow without any newlines
3. Including library headers is always done via angle brackets, essentially never using quotes

The first two points ensure that `clang-format` does _the right thing_ ™.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -6,22 +6,118 @@
...
                                                                         DilithiumConstants mode) const {
-   return Dilithium_Algos::expand_keypair(DilithiumSeedRandomness(private_key_seed), std::move(mode));
+   if(private_key_bits.size() == 32) {
+      // backwards compatibility (not RFC 9881 conforming) to raw seed format.
+      return Botan::Dilithium_Algos::expand_keypair(Botan::DilithiumSeedRandomness(private_key_bits), std::move(mode));
+   }
+   /* we have to check 3 different format: "seed-only", "expanded-only", and "both"
+     */
+   decoding_func fn_arr[3] = {decode_seed_plus_expanded_or_throw, decode_seed_only_or_throw, decode_expanded_only_or_throw};
+   for(auto fn : fn_arr) {
```
reneme: No need to bother with explicitly specifying the function pointer type (i.e. just remove the `decoding_func` typedef above). This should work just fine:


```suggestion
   for(auto fn : {decode_seed_plus_expanded_or_throw, decode_seed_only_or_throw, decode_expanded_only_or_throw}) {
```

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -6,22 +6,118 @@
...
+      try {
+         return fn(private_key_bits, mode);
+      } catch(const Botan::Decoding_Error& e) {
+         // pass
+      }
```
reneme: While I do find this loop to be a clever approach, it is generally questionable to use exception handling for non-error control flow in my opinion. Both because this is quite inefficient on many platforms and because it can make debugging confusing when you want to set "Break on all C++ exceptions".

Especially because it would be easily avoidable here, like so:

1. Check if `private_key_bits` is 32bytes long and go with the "backward compatibility" case
   _This is already implemented like that above._
2. Create a `BER_Decoder(private_key_bits)` and pull the first octet from it
   1. if this octet is 32bytes long, consider it the seed
   2. if it is longer, it is the expanded key, and we're done decoding (i.e. `.verify_end()`)
3. If the first octet was the 32bytes, try pulling another octet (`.decode_optional()` see [ber_dec.h:260](https://github.com/randombit/botan/blob/e7443105fe58e165b9a0a147c923ea60b2a1fc02/src/lib/asn1/ber_dec.h#L259-L260))
   1. the second octet must be the expanded key (i.e. check for >32bytes)
   2. Ensure that the decoding is done (i.e. `.verify_end()`)

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -6,22 +6,118 @@
...
+      Botan::Dilithium_Algos::decode_keypair(Botan::DilithiumSerializedPrivateKey(expanded), mode);
+   Botan::DilithiumInternalKeypair key_pair_from_seed =
+      Botan::Dilithium_Algos::expand_keypair(Botan::DilithiumSeedRandomness(seed), std::move(mode));
+
+   DilithiumSerializedPrivateKey expanded_from_seed = Dilithium_Algos::encode_keypair(key_pair_from_seed);
+
+   if(expanded_from_seed.get() != expanded) {
+      throw Botan::Decoding_Error("seed and expanded key in ML-DSA serialized key do not match");
+   }
+   return key_pair;
```
reneme: Please return the `key_pair_from_seed` instead, because `::expand_keypair()` will ensure that the `seed` value is retained in the internal private key structure. That way, it will be available if the user decides to encode the private key again later.

Also, it would be great to have a test that ensures this guarantee. I.e. "if I load a key from an ASN.1 blob containing both the expanded key and the seed, then it is possible to encode the key in seed-only or 'both' modes."

src/lib/pubkey/dilithium/dilithium_common/dilithium.h
```diff
@@ -142,6 +142,8 @@ class BOTAN_PUBLIC_API(3, 0) Dilithium_PrivateKey final : public virtual Dilithi
                                                              std::string_view params,
                                                              std::string_view provider) const override;
 
+      bool is_mldsa() const;
```
reneme: Do we need this? Support for the pre-standard Dilithium implementation will likely disappear at some point (maybe as early as Botan4 in 2027-ish). And then, this API will be obsolete. No real objection to providing this from my side, though. But perhaps at least provide an equivalent `is_dilithium_round3()` as the counterpart.

src/tests/test_dilithium.cpp
```diff
@@ -9,7 +9,11 @@
  * Botan is released under the Simplified BSD License (see license.txt)
  */
 
+#include "botan/pk_keys.h"
+#include "test_rng.h"
 #include "tests.h"
+#include <memory>
+#include <vector>
```
reneme:
```suggestion
#include "test_rng.h"
#include "tests.h"

#include <botan/pk_keys.h>
#include <memory>
#include <vector>
```

reneme (changes requested):
Thanks for taking that on. Apart from some initial comments regarding your implementation, below are some ideas regarding your discussion items.

Also note that CI is currently unhappy with relevant issues.

---

> One real problem of this fix is that it destroys backwards compatibility for those users who relied on the private key being the raw seed. Even in the "seed-only" format, now the encoded private key is wrapped into an ASN.1/DER TLV structure. Unfortunately, Botan provides no possibility to distinguish the context in which a key is encoded. [...] I see no other way than breaking backwards compatibility now.

Yeah, we have a similar API-ergonomics issue for encoding ECC public keys that may be stored 'compressed' or 'uncompressed'. The solution there essentially boils down to a state setter in the key object (see below). This isn't optimal but is probably the best bet for the time being, instead of breaking backward compatibility

https://github.com/randombit/botan/blob/963f57521e26f6b37be6a395d530744de2a40839/src/lib/pubkey/ecc_key/ecc_key.h#L80-L84

> The clean solution would be to have the encoding layer independent from the cryptographic keys and always force the user to explicitly choose an encoding context. In any case, a fundamental solution for the problem is out of reach of this PR.

I agree. In the past @randombit and I had sketched out a fairly generic builder-style API for such "ad-hoc configuration needs" (see #5021, #4996, #4593, #4694). Something like that would likely be helpful here as well. Probably it is time to resurrect this effort after the release of 3.11.0

> For compatibility with the DilithiumRound3 and previous ML-DSA draft, the private key decoding in this implementation also still accepts a raw seed, which is not specified by RFC 9881.

We should definitely keep the possibility for importing raw keys. Not all applications rely on ASN.1 containers for storing their keys, and we want to be compatible with those. For instance, one can also import ECC private keys just via their private scalar (and their group parameters).

> A minor point with the change of the format, the existing ML-DSA KAT tests show errors because the hash of the private key is not as expected.

This should be addressed by allowing to export the old "raw" format.

falko-strenzke:
> I agree. In the past @randombit and I had sketched out a fairly generic builder-style API for such "ad-hoc configuration needs" (see #5021, #4996, #4593, #4694). Something like that would likely be helpful here as well. Probably it is time to resurrect this effort after the release of 3.11.0

I don't think that the builder pattern is really addressing the problem. Encoding always happens in a context like X.509/PKCS#8 etc. or OpenPGP. Without a specific context, encoding of keys doesn't really make much sense. Agreeable, that has changed a bit since NIST defined explicit encodings for public and private keys for the PQC schemes. In my view that would mean that Botan should support the NIST and PKCS#8 encoding contexts.

The problem with the current implementation in Botan is that private keys offer `private_key_bits()` as a general encoding function that is agnostic to the context. So in one way or another, each invocation of an encoding operation should specifiy the encoding context. That would solve our problem.

falko-strenzke:
> This should be addressed by allowing to export the old "raw" format.

In my above comment I explained the problem with that. Allowing to export the raw format means allowing to create invalid PKCS#8 containers. This is a no-go for a cryptographic library. Invalid artifacts that are not interoperable and will cause problems for users that unintentionally create them.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: The code was refactored according to your comments below, this is now N/A.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: Makes a lot of sense. I refactored the code in this spirit, with some differences in the structure though.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: Fixed.

src/tests/test_dilithium.cpp
falko-strenzke: Adopted.

src/lib/pubkey/dilithium/dilithium_common/dilithium.h
falko-strenzke: The function is currently needed in the tests. (We could solve that by recalculating the private key hashes if that is deemed necessary). The disappearance of Dilithium-Round-3 is something that I am not yet convinced of. I added the function `is_dilithium_round3()` as you suggested.

reneme:
> The problem with the current implementation in Botan is that private keys offer private_key_bits() as a general encoding function that is agnostic to the context. So in one way or another, each invocation of an encoding operation should specifiy the encoding context. That would solve our problem.

One way of adding such an API would be the builder pattern in my opinion, though. Along the lines [outlined here](https://github.com/randombit/botan/pull/4593#pullrequestreview-2575929039):

```C++
auto nist_encoding = private_key.export()     // -> returns some generic builder API
                                .as_raw()     // -> raw (NIST specified)
                                .as_seed();   // -> use the seed

auto pkcs8 = private_key.export()   // -> same generic builder API
                        .as_pkcs8()
                        .with_seed()           // use seed only
                        .without_encryption()  // no container encryption
                        .as_der();  // no ASCII armor
```

... for sure the devil is in the details here. E.g. most keys won't understand what `with_seed()` means, but the API has to be generic because it must be defined on the polymorphic `Private_Key` base class.

Or am I missing your point somehow?

src/lib/pubkey/dilithium/dilithium_common/dilithium.h
falko-strenzke: @reneme 
We should still decide whether to remove the ML-DSA private key hashes in the test data or whether to update them to the new encoding. The latter is somewhat fragile, since the choice of the concrete encoding format (currently "seed-only") is arbitrary and might change in the future.

falko-strenzke:
> One way of adding such an API would be the builder pattern in my opinion, though. Along the lines [outlined here](https://github.com/randombit/botan/pull/4593#pullrequestreview-2575929039):
> 
> ```c++
> auto nist_encoding = private_key.export()     // -> returns some generic builder API
>                                 .as_raw()     // -> raw (NIST specified)
>                                 .as_seed();   // -> use the seed
> 
> auto pkcs8 = private_key.export()   // -> same generic builder API
>                         .as_pkcs8()
>                         .with_seed()           // use seed only
>                         .without_encryption()  // no container encryption
>                         .as_der();  // no ASCII armor
> ```
> 
> ... for sure the devil is in the details here. E.g. most keys won't understand what `with_seed()` means, but the API has to be generic because it must be defined on the polymorphic `Private_Key` base class.
> 
> Or am I missing your point somehow?

I would prefer a solution where the key doesn't have a public function to encode itself, but there is an API 
``` 
secure_vector<byte> PKCS8::Encode_Private_Key(Private_Key const& key);
bool PKCS8::Private_Key_supports_PKCS8_encoding(Private_Key const& key);
``` 
That keeps the private key interface clean from the encoding layer. Possibly it could also be enhanced to control the encoding format variant where such exists, but I don't think it is really necessary. Not many users will need that.

reneme:
I'd be fine with freestanding encoding functions as well. Though, such an interface still needs to be configurable, though. I.e. to allow the user to choose the inclusion of the seed and/or the expanded key.

falko-strenzke:
> I'd be fine with freestanding encoding functions as well. Though, such an interface still needs to be configurable, though. I.e. to allow the user to choose the inclusion of the seed and/or the expanded key.

OK, sure, we can also offer the option of the concrete encoding variant in the new API. But I think it might take us some time before we have worked out and agreed on the details of a new API. For now I suggest that we get this bugfix upstream. Currently Botan produces invalid PKCS#8 for ML-DSA. This PR fixes the bug and uses fixed seed only encoding for ML-DSA and can still decode the previous "raw seed" format. DilithiumRound3 is left with raw seed decoding.

falko-strenzke:
I now implemented `raw_private_key_bits()` to return the raw seed if available or otherwise throw. Accessing the raw seed is necessary for the MLDSA-composite certificates.

falko-strenzke:
For the record: I can't figure out the problem with the remaining CI failure. Do I possibly have to exclude the test "mldsa_private_key" from the script call 

```
src/scripts/run_tests_under_valgrind.py --test-binary=/home/runner/work/botan/botan/build/botan-test --verbose --bunch --track-origins --with-leak-check --skip-tests=argon2,bcrypt,bcrypt_pbkdf,compression_tests,cryptobox,dh_invalid,dh_kat,dh_keygen,dl_group_gen,dlies,dsa_kat_verify,dsa_param,ecc_basemul,ecdsa_verify_wycheproof,ed25519_sign,elgamal_decrypt,elgamal_encrypt,elgamal_keygen,ffi_dh,ffi_dsa,ffi_elgamal,frodo_kat_tests,hash_nist_mc,hss_lms_keygen,hss_lms_sign,mce_keygen,passhash9,pbkdf,pcurves_arith,pwdhash,rsa_encrypt,rsa_pss,rsa_pss_raw,scrypt,sphincsplus,sphincsplus_fors,slh_dsa_keygen,slh_dsa,srp6_kat,srp6_rt,unit_tls,x509_path_bsi,x509_path_rsa_pss,xmss_keygen,xmss_keygen_reference,xmss_sign,xmss_unit_tests,xmss_verify,xmss_verify_invalid,dilithium_kat_6x5_Deterministic,dilithium_kat_6x5_Randomized,dilithium_kat_8x7_Deterministic,dilithium_kat_8x7_Randomized,dilithium_kat_6x5_AES_Deterministic,dilithium_kat_6x5_AES_Randomized,dilithium_kat_8x7_AES_Deterministic,dilithium_kat_8x7_AES_Randomized,ml_dsa_kat_6x5_Deterministic,ml_dsa_kat_6x5_Randomized,ml_dsa_kat_8x7_Deterministic,ml_dsa_kat_8x7_Randomized
```
which supposedly failed? Is there possibly any performance problem associated with this test that breaks the execution on the CI machine?

Since the above call runs fine locally on my system (including the new test "mldsa_private_key") and the CI output doesn't indicated the error, I am now re-running the job with debug logging enabled.

randombit:
@falko-strenzke Can you rebase and see if the ACVP and/or Wycheproof tests can be extended here - IIRC both have various tests relating to seed vs expanded-key encodings which are currently being skipped.

falko-strenzke:
> > A minor point with the change of the format, the existing ML-DSA KAT tests show errors because the hash of the private key is not as expected.
> 
> This should be addressed by allowing to export the old "raw" format.

@reneme 

Done.

src/tests/data/mldsa_privkey.vec
```diff
@@ -0,0 +1,29 @@
...
...132ffb7a0709aed5a9c0ff05fb0f2bbf28409eed7b5f5801be964ced019e1cb7851d3851f10290674e19ffb008b301c4acf641a2bb14216e1d69cabf52b5ef227496b0f30799a855d117fad3744a6fa33503ea798b52ddd7ee5426609dbfcd3f0c13b164d6c051f7ed4a119719a712e388d328402081ff1354b554d2c237afed3b151c4ba8e9f4bdeb8499a3066e26bbc69e8af089dec71731d1dc529eab17ef7374734c0fe475494c83836bdd34a03b9bc89914716061bfb98ec6e61c3ed4438edcaf25243c647086b9ea7018b0d9a8a0b00cecb00abde2498d69c2336101a772cbe4f571523f51bd05882cdf358b849cc140aa1faf22423a12851ce0e33fd48975a4959fa5c5fe418c93908191ab6e741b77bfe02cbd698ee795c466d615619e6441382c6eac01834ee9ab73cea80bbe235c78da91bd79b6f82f899785d68700d393e675c2224d6b7a1ad21320495679adaed70167b50866713a53109db7b6f7d81304ecdfd83b319b1ef248306b45ad29e7ddcc863dac56048b5d69ea175011f7614c00a86a863cde1872a8932878b9ac7e1ac5bda4997b72064f0cd75f4c814e034de11acb9013cf7ea926b4e7eaace070c7ba2188efad2e431e1223d45dd05c4d8403c2e45cee6413ecbe7527e873e455c4e610a61839aacc0bd56d2483e78f298b66a478eb2f558cbafca86be847baeb02c5b216c8cd88fea4df249b09e670a20703abac24b0a91abc4a5646601442ba10becfd30993880051d07f56a05a9379e7a8e6befee3f22faa106398f7706006e42e9be1ef89d25c272f11a95095c587d713732284de9dbd3c7217b0689e21d8eb0ff69668
```
randombit: This doesn't seem to match the value given in RFC 9881.

Considering those test vectors are published as PEM it might be better to include each as individual named files (eg rfc9881_mldsa44_seed.pem, rfc9881_mldsa44_expand.pem, ...)

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -8,20 +8,113 @@
...
+Botan::DilithiumInternalKeypair decode_seed_plus_expanded(std::span<const uint8_t> key_bits,
+                                                          Botan::DilithiumConstants mode) {
+   Botan::secure_vector<uint8_t> expanded;
+   Botan::secure_vector<uint8_t> seed;
+   Botan::BER_Decoder(key_bits)
+      .start_sequence()
+      .decode(seed, Botan::ASN1_Type::OctetString)
+      .decode(expanded, Botan::ASN1_Type::OctetString)
+      .end_cons()
+      .verify_end();
```
randombit, line 52: This should verify the decoded seed is the correct length

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -8,20 +8,113 @@
...
+      return Botan::Dilithium_Algos::expand_keypair(Botan::DilithiumSeedRandomness(private_key_bits), std::move(mode));
+   }
+   // "seed-only" format from RFC 9881
+   BER_Decoder ber_dec(private_key_bits);
+   auto obj = ber_dec.peek_next_object();
+   if(obj.type() == ASN1_Type(0)) {
+      return decode_seed_only(private_key_bits, mode);
+   }
+   // now it could still be "expanded-only" or "both"
+   if(obj.type() == ASN1_Type::OctetString) {
```
randombit: These both should be checking the class as well as the type flag

src/tests/test_dilithium.cpp
```diff
@@ -288,4 +294,56 @@ BOTAN_REGISTER_TEST("pubkey", "dilithium_keygen", Dilithium_Keygen_Tests);
...
+         if(name.ends_with("-invalid")) {
+            expect_decoding_failure = true;
+         }
+         std::unique_ptr<Botan::Dilithium_PrivateKey> priv_key;
+         try {
+            priv_key = std::make_unique<Botan::Dilithium_PrivateKey>(key_bits, mode);
+         } catch(const Botan::Decoding_Error&) {
+            result.test_is_true("invalid ML-DSA key rejected", expect_decoding_failure);
+            return result;
+         }
```
randombit, line 391: At this point the key has decoded, so it should test that `expect_decoding_failure` is false

src/tests/test_dilithium.cpp
```diff
@@ -288,4 +294,56 @@ BOTAN_REGISTER_TEST("pubkey", "dilithium_keygen", Dilithium_Keygen_Tests);
 
 }  // namespace
 
+#if defined(BOTAN_HAS_DILITHIUM_COMMON) && defined(BOTAN_HAS_SHA3)
```
randombit: Build guard here seems wrong this requires specifically `BOTAN_HAS_ML_DSA` and doesn't seem to directly rely on SHA-3 (I guess implicitly of course, but that's handled via the ML-DSA dependency) but does depend directly on AES.

src/tests/test_dilithium.cpp
```diff
@@ -50,8 +54,10 @@ class Dilithium_KAT_Tests : public Text_Based_Test {
 
          const Botan::Dilithium_PrivateKey priv_key(*dilithium_test_rng, DerivedT::mode);
 
-         result.test_bin_eq(
-            "generated expected private key hash", sha3_256->process(priv_key.private_key_bits()), ref_sk_hash);
+         if(!priv_key.is_mldsa()) {
+            result.test_bin_eq(
+               "generated expected private key hash", sha3_256->process(priv_key.private_key_bits()), ref_sk_hash);
+         }
```
randombit: For ML-DSA this could continue to check using `raw_private_key_bits`

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -8,20 +8,113 @@
...
+#include <botan/exceptn.h>
 #include <botan/internal/dilithium_algos.h>
+#include <botan/internal/dilithium_types.h>
+#include <utility>
 
 namespace Botan {
 
+namespace {
+typedef Botan::DilithiumInternalKeypair (*decoding_func)(std::span<const uint8_t> key_bits,
+                                                         Botan::DilithiumConstants mode);
```
randombit: Unused?

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -8,20 +8,113 @@
...
+   }
+
+   return result;
 }
 
-DilithiumInternalKeypair ML_DSA_Expanding_Keypair_Codec::decode_keypair(std::span<const uint8_t> private_key_seed,
+DilithiumInternalKeypair ML_DSA_Expanding_Keypair_Codec::decode_keypair(std::span<const uint8_t> private_key_bits,
                                                                         DilithiumConstants mode) const {
-   return Dilithium_Algos::expand_keypair(DilithiumSeedRandomness(private_key_seed), std::move(mode));
+   if(private_key_bits.size() == 32) {
```
randombit: Should be `SEED_RANDOMNESS_BYTES`

src/tests/test_dilithium.cpp
```diff
@@ -288,4 +294,56 @@ BOTAN_REGISTER_TEST("pubkey", "dilithium_keygen", Dilithium_Keygen_Tests);
...
+         auto signature = signer.sign_message(ref_msg.data(), ref_msg.size(), rng);
+
+         const Botan::Dilithium_PublicKey pub_key(priv_key->public_key_bits(), mode);
+         auto verifier = Botan::PK_Verifier(pub_key, "");
+         verifier.update(ref_msg.data(), ref_msg.size());
+         result.test_is_true("signature verifies", verifier.check_signature(signature.data(), signature.size()));
+
+         auto reencoded_priv_key = priv_key->private_key_bits();
+         auto redecoded_priv_key = Botan::Dilithium_PrivateKey(reencoded_priv_key, mode);
+         result.test_is_true("re-encoding and subsequent re-decoding of private ML-DSA key without error", true);
```
randombit: This test is vacuous. Could instead check that eg the raw public key matches the expected value.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
randombit: And also that expanded is the correct size for the mode

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -8,20 +8,113 @@
...
+      .decode(seed, Botan::ASN1_Type::OctetString, Botan::ASN1_Type(0), Botan::ASN1_Class::ContextSpecific)
+      .verify_end();
+   return Botan::Dilithium_Algos::expand_keypair(Botan::DilithiumSeedRandomness(seed), std::move(mode));
+}
+
+Botan::DilithiumInternalKeypair decode_expanded_only(std::span<const uint8_t> key_bits,
+                                                     Botan::DilithiumConstants mode) {
+   Botan::secure_vector<uint8_t> expanded;
+   Botan::BER_Decoder(key_bits).decode(expanded, Botan::ASN1_Type::OctetString).verify_end();
+   Botan::DilithiumInternalKeypair key_pair =
```
randombit, line 38: Should check that the size of expanded is consistent here

randombit (changes requested):
Some review comments. This also really needs a rebase to latest master to pick up for instance the new ACVP and Wycheproof test runs in CI.

We had to deal with the same issue for ML-KEM in #4817 and it probably makes sense to align those interfaces. [At a minimum adding an enum identifying the key serialization in use and allowing multiple different serializations]

src/tests/test_dilithium.cpp
falko-strenzke: Done

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: Right, done.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: Done.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: Done.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: I added the size check in the invoked decoding function, I think this is where it belongs.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: Also here, both size checks are (now) done in the decoding functions.

src/tests/test_dilithium.cpp
falko-strenzke: Done

src/tests/test_dilithium.cpp
falko-strenzke: Done.

src/tests/test_dilithium.cpp
falko-strenzke: I fixed it properly, all tests can be used as before.

falko-strenzke:
Done.

falko-strenzke:
> @falko-strenzke Can you rebase and see if the ACVP and/or Wycheproof tests can be extended here - IIRC both have various tests relating to seed vs expanded-key encodings which are currently being skipped.

Yes, done. expanded format is supported in both now.

src/tests/data/mldsa_privkey.vec
falko-strenzke: I changed / fixed different things:
- Previous test vectors were the "internal" private keys from the RFC 9881 vectors. Now we use the PEM as you suggest files directly.
- For the invalid test vectors, the format was erroneously still PKCS#8, which made them useless (now fixed implicitly)
- ML-DSA private key decoding has now an additional consistency check for t_0, which is checked by the RFC 9881 test vectors.

falko-strenzke:
> We had to deal with the same issue for ML-KEM in #4817 and it probably makes sense to align those interfaces. [At a minimum adding an enum identifying the key serialization in use and allowing multiple different serializations]


I see, but the additions in [#4817](https://github.com/randombit/botan/pull/4817/ ) are not complete. ML-KEM private key encoding is also still broken, i.e., is not compliant with RFC 9935. We now need to cover the distinction between raw, i.e. mere private key material as in FIPS 204, and the properly ASN.1/DER wrapped structure required for RFC-conforming PKCS#8 encoding. So the whole picture looks to me like this:

- We create a new module `ml_common`
  - which defines an interface class that both ML-DSA and ML-KEM inherit from. The interface defines:
    - `enum class MlPrivateKeyFormat`
      - which is extended by *both* format
    - `MlPrivateKeyFormat private_key_format() const;`
    - `secure_vector<uint8_t> private_key_bits_with_format(MlPrivateKeyFormat format) const;`
    - `secure_vector<uint8_t> raw_private_key_bits_with_format(MlPrivateKeyFormat format) const;`
  - General behaviour: 
    - A key decoded from a specific format (seed, expanded, *both*) is encoded in the same format again unless the format is explicitly requested (as it is currently realized by ML-KEM). I think this point is clear.
    - A newly generated key encodes itself in a predefined format, which we have to decide on: either the seed format or *both* format. OpenSSL outputs in *both* by default, from my point of view that is the most reasonable and conservative choice.

The current implementation status is:

| Feature                                                               | ML-KEM       | ML-DSA |
| ---                                                                   | ---          | ---    |
| `private_key_format()`                                                | ✓            | □      |
| `private_key_bits_with_format(MlPrivateKeyFormat format)`             | ✓            | □      |
| `raw_private_key_bits_with_format(MlPrivateKeyFormat format)`         | □            | □      |


So the decisions that need to be taken before I can move on with this PR are in my view:
- Introduce the interface class in a new module `ml_common` or just mirror the functions and logic in both ML-... Pro (new module and interface): the common interface underlines the commitment to the same functions and logic. Con: If the key encoding mechanism gets reworked fundamentally, this module will become obsolete again.
- Encode newly generated keys as seed-only or *both*.

reneme:
One side note regarding "a new common module": If we go that way, you could use the existing `pqcrystals` module which currently implements helpers and base classes for both ML-DSA and ML-KEM. This is currently an "Internal" module, we might have to change that to let it expose public headers if needed.

reneme:
I agree with your suggestions regarding the API additions. Also, moving and extending the `MlPrivateKeyFormat` enum as you suggested makes sense to me. As mentioned before, the common `pqcrystals` module may be a good place for that.

Allow me to extend on your table to make sure we're all on the same page:

| API                 | Behavior                                                   | ML-KEM       | ML-DSA | Behavior Change |
| ---                   | ---                                                | ---          | ---    | --- |
| `private_key_format()` | Returns whatever `MlPrivateKeyFormat` is currently set  | ✓           | □      | none |
| `raw_private_key_bits_with_format(MlPrivateKeyFormat format)`  | Encodes the private key according to `format`        | □            | □      | none |
| `private_key_bits_with_format(MlPrivateKeyFormat format)` | Wraps the output of `raw_private_key_bits_with_format()` into ASN.1/DER             | ✓            | □      | none |
| `raw_private_key_bits()` | Encodes the private key according to `private_key_format()` |  ✓            | ✓      |  now depends on `private_key_format()` |
| `private_key_bits()` | Wraps the output of `raw_private_key_bits()` into ASN.1/DER  | ?             | ?      |  now produces "wrapped" keys and depends on `private_key_format()` |

In my opinion the changes in the existing API are okay and/or outright bugfixes: Especially `private_key_bits()` just being an alias for `raw_private_key_bits()` was always incorrect. The documentation in the base class in `pk_keys.h` literally states: "Return the PKCS8 private key encoding". Hence, I think it is safe to assume that existing applications stayed away from this method and used `raw_private_key_bits()` anyway. 🤞

Additionally, I think we should offer `set_private_key_format(MlPrivateKeyFormat format)` for both ML-KEM/ML-DSA that sanity checks whether or not the desired encoding is possible (i.e. throw if the private seed would be needed but isn't available). That setter would make the API consistent with `EC_Public_Key::set_point_encoding()`.

reneme:
> Encode newly generated keys as seed-only or both.

My vote is on "both" as well. Applications can always override that choice anyway.

falko-strenzke:
@reneme Just to be clear: for `private_key_bits_with_format(MlPrivateKeyFormat format)` that would mean a breaking change: previously, it returned the raw key material, and now we plan to let it return the properly ASN.1-wrapped format as needed for the inner PKCS#8.  The table thus should look like this in my understanding:

| API                                                           | Behavior                                                                | ML-KEM | ML-DSA | Behavior Change                                                                          |
| ---                                                           | ---                                                                     | ---    | ---    | ---                                                                                      |
| `private_key_format()`                                        | Returns whatever `MlPrivateKeyFormat` is currently set                  | ✓      | □      | none                                                                                     |
| `raw_private_key_bits_with_format(MlPrivateKeyFormat format)` | Encodes the private key according to `format`                           | □      | □      | N/A                                                                                      |
| `private_key_bits_with_format(MlPrivateKeyFormat format)`     | Wraps the output of `raw_private_key_bits_with_format()` into ASN.1/DER | ✓      | □      | ~~none~~ **Previously returned the raw key material, now returns ASN.1/DER-wrapped key** |
| `raw_private_key_bits()`                                      | Encodes the private key according to `private_key_format()`             | ✓      | ✓      | now depends on `private_key_format()`, **no change: yields raw key material**            |
| `private_key_bits()`                                          | Wraps the output of `raw_private_key_bits()` into ASN.1/DER             | ✓    | ✓      | now produces "wrapped" keys and depends on `private_key_format()`                        |

reneme:
You are right. In #4817 we should have called the new ML-KEM-specific function `raw_private_key_bits_with_format()`. That's too much of a SemVer violation for my taste, to be honest. Because this is a very specific API that some application might actually depend on legitimately. 😒

**1) Alternative suggestion:** Let's deprecate that API and instead introduce new methods:

* `BOTAN_DEPRECATED("Use formatted_raw_private_key_bits() instead") private_key_bits_with_format(MlPrivateKeyFormat)`
* `formatted_raw_private_key_bits(MlPrivateKeyFormat format)` ... returning the raw key in the passed format
* `formatted_private_key_bits(MlPrivateKeyFormat format)` ... returning the ASN.1/DER wrapped key in the passed format

**2) Another alternative:** just deprecate `private_key_bits_with_format(MlPrivateKeyFormat)` and don't introduce any explicit format-aware API. Instead just introduce a `set_private_key_format(MlPrivateKeyFormat)` and use the existing polymorphic functions for serialization based on the object's internal state. That's what we've been doing for the ECC public point encoding as well after all.

Sucks a little, but I guess it's what it is...

@randombit what do you think about this?

falko-strenzke:
> **1) Alternative suggestion:** Let's deprecate that API and instead introduce new methods:
> 
>     * `BOTAN_DEPRECATED("Use formatted_raw_private_key_bits() instead") private_key_bits_with_format(MlPrivateKeyFormat)`
> 
>     * `formatted_raw_private_key_bits(MlPrivateKeyFormat format)` ... returning the raw key in the passed format
> 
>     * `formatted_private_key_bits(MlPrivateKeyFormat format)` ... returning the ASN.1/DER wrapped key in the passed format
> 
> 
> **2) Another alternative:** just deprecate `private_key_bits_with_format(MlPrivateKeyFormat)` and don't introduce any explicit format-aware API. Instead just introduce a `set_private_key_format(MlPrivateKeyFormat)` and use the existing polymorphic functions for serialization based on the object's internal state. That's what we've been doing for the ECC public point encoding as well after all.

Letting the key's state decide the encoding format is not the proper way in my view. This is for the same reason that we don't tell an integer that it is a "hex" or "decimal" integer. The functionality for choosing alternative encodings should live outside the object being encoded as much as possible. (The limitation here is the availability of the seed.)

Another suggestion would be the introduction of a framework for key encoding, for now maybe only applied to ML-... This framework could also address decoding. I don't think accepting just any 32-byte strings as secret keys, which allow for backwards compatibility, is what is suitable in every scenario. Where one expects a properly encoded key, one should be able to specify that in the API.

reneme:
I 100% agree with you, that the internal serialization state in the key is crappy. I really do. I also support the introduction of a versatile encoding/decoding framework to disentangle live key objects and their serialization concerns.

However, for the sake of progress on this particular PR, I would suggest to go with one of the suggested solutions. I'm happy to throw around ideas and prototype APIs independent of this. Perhaps to be introduced in Botan4.

falko-strenzke:
> However, for the sake of progress on this particular PR, I would suggest to go with one of the suggested solutions.

You are right, makes no sense to make this a blocker now. Then I vote for your propsal 1.:

**1) Alternative suggestion:** Let's deprecate that API and instead introduce new methods:

* `BOTAN_DEPRECATED("Use formatted_raw_private_key_bits() instead") private_key_bits_with_format(MlPrivateKeyFormat)`
* `formatted_raw_private_key_bits(MlPrivateKeyFormat format)` ... returning the raw key in the passed format
* `formatted_private_key_bits(MlPrivateKeyFormat format)` ... returning the ASN.1/DER wrapped key in the passed format

falko-strenzke:
> However, for the sake of progress on this particular PR, I would suggest to go with one of the suggested solutions.

You are right, makes no sense to make this a blocker now. Then I vote for your propsal 1.:

**1) Alternative suggestion:** Let's deprecate that API and instead introduce new methods:

* `BOTAN_DEPRECATED("Use formatted_raw_private_key_bits() instead") private_key_bits_with_format(MlPrivateKeyFormat)`
* `formatted_raw_private_key_bits(MlPrivateKeyFormat format)` ... returning the raw key in the passed format
* `formatted_private_key_bits(MlPrivateKeyFormat format)` ... returning the ASN.1/DER wrapped key in the passed format

@randombit Would you agree? This is at least a mid-term interface decision it seems.

falko-strenzke:
@reneme The new interface is in place. The "pqcrystals" module, since it was not a public one, was renamed to "module_lattice" (`src/lib/pubkey/module_lattice/`).

doc/api_ref/pubkey.rst
```diff
@@ -194,6 +207,78 @@ Currently two flavors of Kyber are implemented in separate Botan modules:
...
+      no raw (non-ASN.1) encoding of this format.
+
+.. cpp:class:: Module_Lattice_PrivateKey : public virtual Private_Key
+
+   .. cpp:function:: MlPrivateKeyFormat private_key_format() const
+
+      The format this key was loaded from or, for a freshly generated key, the
+      default format of the scheme (``Seed`` for ML-KEM, ``Both`` for ML-DSA).
+      This is the format used by ``private_key_bits()``, ``private_key_info()``
+      and ``raw_private_key_bits()``.
```
reneme, line 250: How come that the default encoding for ML-KEM and ML-DSA differ? Are that the specified defaults?

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
```diff
@@ -476,15 +477,47 @@ Dilithium_PrivateKey::Dilithium_PrivateKey(const AlgorithmIdentifier& alg_id, st
...
+   switch(format) {
+      case MlPrivateKeyFormat::Seed: {
+         if(is_dilithium_round3()) {
+            throw Encoding_Error("Dilithium round 3 private keys only support the expanded format");
+         }
+         const auto& seed = m_private->seed();
+         if(!seed.has_value()) {
+            throw Encoding_Error("ML-DSA private key does not contain the seed, cannot encode it in the seed format");
+         }
+         return seed.value().get();
```
reneme: Ultra-nit:


```suggestion
         return seed->get();
```

src/lib/pubkey/dilithium/dilithium_common/dilithium_algos.cpp
```diff
@@ -444,7 +436,23 @@ DilithiumInternalKeypair decode_keypair(StrongSpan<const DilithiumSerializedPriv
...
+   // rho and t1 form the public key. An inconsistent t0 should be detected
+   // explicitly (see RFC 9881, Appendix C.4 for such malformed keys).
+   auto t0_consistent = CT::Mask<uint32_t>::set();
+   for(size_t i = 0; i < t0.size(); ++i) {
+      const auto stored = t0[i].coefficients();
+      const auto derived = t0_derived[i].coefficients();
+      for(size_t j = 0; j < stored.size(); ++j) {
+         t0_consistent &=
+            CT::Mask<uint32_t>::is_equal(static_cast<uint32_t>(stored[j]), static_cast<uint32_t>(derived[j]));
+      }
```
reneme, line 452: Can we be absolutely sure that `derived` holds the same number of coefficients as `stored`? I think, a `BOTAN_ASSERT_NOMSG()` checking this wouldn't hurt.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -8,20 +8,134 @@
...
+
+Botan::DilithiumInternalKeypair decode_seed_only(std::span<const uint8_t> key_bits, Botan::DilithiumConstants mode) {
+   Botan::secure_vector<uint8_t> seed;
+   Botan::BER_Decoder(key_bits)
+      .decode(seed, Botan::ASN1_Type::OctetString, Botan::ASN1_Type(0), Botan::ASN1_Class::ContextSpecific)
+      .verify_end();
+   if(seed.size() != Botan::DilithiumConstants::SEED_RANDOMNESS_BYTES) {
+      throw Botan::Decoding_Error("invalid length of ML-DSA private key seed");
+   }
+   return Botan::Dilithium_Algos::expand_keypair(Botan::DilithiumSeedRandomness(seed), std::move(mode));
```
reneme: Ugh. It would be good if the `BER_Decoder` could handle the wrapped byte array strong types. Not in this PR, though.

For now, you might want to at least move the `seed` into the strong type, though:


```suggestion
   Botan::secure_vector<uint8_t> seed;
   Botan::BER_Decoder(key_bits)
      .decode(seed, Botan::ASN1_Type::OctetString, Botan::ASN1_Type(0), Botan::ASN1_Class::ContextSpecific)
      .verify_end();
   if(seed.size() != Botan::DilithiumConstants::SEED_RANDOMNESS_BYTES) {
      throw Botan::Decoding_Error("invalid length of ML-DSA private key seed");
   }
   return Botan::Dilithium_Algos::expand_keypair(Botan::DilithiumSeedRandomness(std::move(seed)), std::move(mode));
```

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -8,20 +8,134 @@
...
+Botan::DilithiumInternalKeypair decode_expanded_only(std::span<const uint8_t> key_bits,
+                                                     Botan::DilithiumConstants mode) {
+   Botan::secure_vector<uint8_t> expanded;
+   Botan::BER_Decoder(key_bits).decode(expanded, Botan::ASN1_Type::OctetString).verify_end();
+   if(expanded.size() != mode.private_key_bytes()) {
+      throw Botan::Decoding_Error("invalid length of ML-DSA (or Dilithium) expanded private key byte string");
+   }
+   Botan::DilithiumInternalKeypair key_pair =
+      Botan::Dilithium_Algos::decode_keypair(Botan::DilithiumSerializedPrivateKey(expanded), std::move(mode));
+   return key_pair;
```
reneme:
```suggestion
   return Botan::Dilithium_Algos::decode_keypair(Botan::DilithiumSerializedPrivateKey(std::move(expanded)), std::move(mode));
```

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -8,20 +8,134 @@
...
+      .decode(seed, Botan::ASN1_Type::OctetString)
+      .decode(expanded, Botan::ASN1_Type::OctetString)
+      .end_cons()
+      .verify_end();
+   const Botan::DilithiumInternalKeypair key_pair =
+      Botan::Dilithium_Algos::decode_keypair(Botan::DilithiumSerializedPrivateKey(expanded), mode);
+   const Botan::DilithiumInternalKeypair key_pair_from_seed =
+      Botan::Dilithium_Algos::expand_keypair(Botan::DilithiumSeedRandomness(seed), std::move(mode));
+
+   DilithiumSerializedPrivateKey expanded_from_seed = Dilithium_Algos::encode_keypair(key_pair_from_seed);
```
reneme, line 61:
```suggestion
   const Botan::DilithiumInternalKeypair key_pair =
      Botan::Dilithium_Algos::decode_keypair(Botan::DilithiumSerializedPrivateKey(std::mov(expanded)), mode);
   Botan::DilithiumInternalKeypair key_pair_from_seed =
      Botan::Dilithium_Algos::expand_keypair(Botan::DilithiumSeedRandomness(std::move(seed)), std::move(mode));

   DilithiumSerializedPrivateKey expanded_from_seed = Dilithium_Algos::encode_keypair(key_pair_from_seed);
```

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -8,20 +8,134 @@
...
+          *
+          * <80 20>
+          * 0  32: [0]
+          *      :   00 01 02 03 04 05 06 07 08 09 0A 0B 0C 0D 0E 0F
+          *      :   10 11 12 13 14 15 16 17 18 19 1A 1B 1C 1D 1E 1F
+          *
+          * Note: The previous, non-standard format, which only contained the
+          * seed as a raw byte string, is still accepted by decode_keypair().
+          */
+         der_enc.encode(seed.value().get(), ASN1_Type::OctetString, ASN1_Type(0), ASN1_Class::ContextSpecific);
```
reneme: Super nit (also a few times below):


```suggestion
         der_enc.encode(seed->get(), ASN1_Type::OctetString, ASN1_Type(0), ASN1_Class::ContextSpecific);
```

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -8,20 +8,134 @@
...
-DilithiumInternalKeypair ML_DSA_Expanding_Keypair_Codec::decode_keypair(std::span<const uint8_t> private_key_seed,
-                                                                        DilithiumConstants mode) const {
-   return Dilithium_Algos::expand_keypair(DilithiumSeedRandomness(private_key_seed), std::move(mode));
+DilithiumDecodedKeypair ML_DSA_Expanding_Keypair_Codec::decode_keypair(std::span<const uint8_t> private_key_bits,
+                                                                       DilithiumConstants mode) const {
+   if(private_key_bits.size() == DilithiumConstants::SEED_RANDOMNESS_BYTES) {
+      // backwards compatibility (not RFC 9881 conforming) to the raw seed format
+      return {Botan::Dilithium_Algos::expand_keypair(Botan::DilithiumSeedRandomness(private_key_bits), std::move(mode)),
+              MlPrivateKeyFormat::Seed};
+   }
```
reneme: For readability, could you please use structured initialization here (and also below)?


```suggestion
      return {
            .keypair = Botan::Dilithium_Algos::expand_keypair(Botan::DilithiumSeedRandomness(private_key_bits), std::move(mode)),
            .format = MlPrivateKeyFormat::Seed,
         };
   }
```

news.rst
```diff
@@ -4,6 +4,19 @@ Release Notes
...
+  supported; a key loaded from one of them is re-encoded in the same format,
+  newly generated keys are encoded in the "both" format. Previously the raw
+  32-byte seed was written, which is still accepted when loading a key.
+  Add the interface ``Module_Lattice_PrivateKey`` (``botan/module_lattice_keys.h``)
+  implemented by ``ML_DSA_PrivateKey`` and ``ML_KEM_PrivateKey`` to query the
+  private key format and to encode a key in a specific format; the enum
+  ``MlPrivateKeyFormat`` moved to that header and gained the value ``Both``.
+  ``Kyber_PrivateKey::private_key_bits_with_format`` is deprecated in favor of
+  ``formatted_raw_private_key_bits``. The internal module ``pqcrystals`` was
+  renamed to ``module_lattice``. (GH #5002 #5307)
```
reneme: I think this part is too much detail for the news.rst (interested users will go to the documentation).

src/lib/pubkey/module_lattice/module_lattice_keys.h
```diff
@@ -0,0 +1,122 @@
...
+       */
+      secure_vector<uint8_t> private_key_bits() const override;
+
+      /**
+       * Equivalent to formatted_raw_private_key_bits(private_key_format()).
+       *
+       * @note Keys in the format MlPrivateKeyFormat::Both, which has no raw
+       *       encoding, return the seed instead.
+       */
+      secure_vector<uint8_t> raw_private_key_bits() const override;
```
reneme: Those are now generic implementations for the public and key encoding methods inherited from the `Private_Key` base class. To the best of my understanding, they ~~shouldn't~~ mustn't be overridden neither by ML-DSA nor by ML-KEM. Instead, the subclasses should implement the new `formatted_***` customization points. Hence:


```suggestion
      /**
       * Equivalent to formatted_private_key_bits(private_key_format())
       */
      secure_vector<uint8_t> private_key_bits() const final;

      /**
       * Equivalent to formatted_raw_private_key_bits(private_key_format()).
       *
       * @note Keys in the format MlPrivateKeyFormat::Both, which has no raw
       *       encoding, return the seed instead.
       */
      secure_vector<uint8_t> raw_private_key_bits() const final;
```

reneme (commented):
This is coming together nicely in my opinion. It badly needs a git history cleanup, though.

In my opinion the added encoding API will serve its purpose just fine until we (hopefully) come up with something more ergonomic for Botan4 (see https://github.com/randombit/botan/issues/4666#issuecomment-5680362827).

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
reneme: Once #5935 landed you will be able to read straight into the `DilithiumSeedRandomness` strong type, like so:

```suggestion
   Botan::DilithiumSeedRandomness seed;
   Botan::BER_Decoder(key_bits)
      .decode(seed, Botan::ASN1_Type::OctetString, Botan::ASN1_Type(0), Botan::ASN1_Class::ContextSpecific)
      .verify_end();
   if(seed.size() != Botan::DilithiumConstants::SEED_RANDOMNESS_BYTES) {
      throw Botan::Decoding_Error("invalid length of ML-DSA private key seed");
   }
   return Botan::Dilithium_Algos::expand_keypair(std::move(seed), std::move(mode));
```

... this is applicable in other places as well, please double-check.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
reneme: Note #5935.

doc/api_ref/pubkey.rst
falko-strenzke: This is only a temporary inconsistency, since this PR does not bring the fix of the ML-KEM private key format. This happens in #5934. However, I will make this clear in the text. Hopefully both fixes land in the same Botan version so the users will never see the inconsistency.

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
falko-strenzke: Fix comes in the next commit.

src/lib/pubkey/dilithium/dilithium_common/dilithium_algos.cpp
falko-strenzke: Fix comes in the next commit.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: Fix with `std::move` comes in the next commit. The strong type is only for later...

doc/api_ref/pubkey.rst
reneme: Awesome. Thanks for clarifying. I think it's safe to assume that both will land in the same release.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: Fix comes in next commit. (But without the "const" deletion, this is not required and will conflict with clang-tidy rules I fear).

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: Applied the std::move()

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: `seed` is moved now. `expanded` cannot be moved at this point: a few lines below it is
compared against the re-encoding of the seed-derived key pair (the consistency check of
RFC 9881, Section 8), so moving it empties the comparison operand and every "both" key
gets rejected. I kept the copy there and added a comment saying why.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: Done

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: Done, for all five returns in decode_keypair() and, for consistency, in the round 3 codec.

news.rst
falko-strenzke: Shortened to the user-visible essentials

src/lib/pubkey/module_lattice/module_lattice_keys.h
falko-strenzke: Done, both are final now.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
reneme: #5935 is in master now. I'll point out the locations where it might make sense to use this new helper independently.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -8,20 +8,147 @@
...
+Botan::DilithiumInternalKeypair decode_seed_only(std::span<const uint8_t> key_bits, Botan::DilithiumConstants mode) {
+   Botan::secure_vector<uint8_t> seed;
+   Botan::BER_Decoder(key_bits)
+      .decode(seed, Botan::ASN1_Type::OctetString, Botan::ASN1_Type(0), Botan::ASN1_Class::ContextSpecific)
+      .verify_end();
+   if(seed.size() != Botan::DilithiumConstants::SEED_RANDOMNESS_BYTES) {
+      throw Botan::Decoding_Error("invalid length of ML-DSA private key seed");
+   }
+   return Botan::Dilithium_Algos::expand_keypair(Botan::DilithiumSeedRandomness(std::move(seed)), std::move(mode));
+}
```
reneme, line 32: Please rebase to latest master (to pull in #5946) and then:


```suggestion
Botan::DilithiumInternalKeypair decode_seed_only(std::span<const uint8_t> key_bits, Botan::DilithiumConstants mode) {
   Botan::DilithiumSeedRandomness seed;
   Botan::BER_Decoder(key_bits)
      .decode(seed, Botan::ASN1_Type::OctetString, Botan::ASN1_Type(0), Botan::ASN1_Class::ContextSpecific)
      .verify_end();
   if(seed.size() != Botan::DilithiumConstants::SEED_RANDOMNESS_BYTES) {
      throw Botan::Decoding_Error("invalid length of ML-DSA private key seed");
   }
   return Botan::Dilithium_Algos::expand_keypair(std::move(seed), std::move(mode));
}
```

.. technically, you should be able to drop a lot of `Botan::` namspace qualifications in this file, but that's really just a nit.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -8,20 +8,147 @@
...
+Botan::DilithiumInternalKeypair decode_expanded_only(std::span<const uint8_t> key_bits,
+                                                     Botan::DilithiumConstants mode) {
+   Botan::secure_vector<uint8_t> expanded;
+   Botan::BER_Decoder(key_bits).decode(expanded, Botan::ASN1_Type::OctetString).verify_end();
+   if(expanded.size() != mode.private_key_bytes()) {
+      throw Botan::Decoding_Error("invalid length of ML-DSA (or Dilithium) expanded private key byte string");
+   }
+   return Botan::Dilithium_Algos::decode_keypair(Botan::DilithiumSerializedPrivateKey(std::move(expanded)),
+                                                 std::move(mode));
+}
```
reneme, line 43: Please rebase to latest master (to pull in #5946) and then:


```suggestion
Botan::DilithiumInternalKeypair decode_expanded_only(std::span<const uint8_t> key_bits,
                                                     Botan::DilithiumConstants mode) {
   Botan::DilithiumSerializedPrivateKey expanded;
   Botan::BER_Decoder(key_bits).decode(expanded, Botan::ASN1_Type::OctetString).verify_end();
   if(expanded.size() != mode.private_key_bytes()) {
      throw Botan::Decoding_Error("invalid length of ML-DSA (or Dilithium) expanded private key byte string");
   }
   return Botan::Dilithium_Algos::decode_keypair(expanded, std::move(mode));
}
```

... note that `expanded` isn't moved on purpose. `decode_keypair` takes this by strong-span.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -8,20 +8,147 @@
...
+   const Botan::DilithiumInternalKeypair key_pair_from_seed =
+      Botan::Dilithium_Algos::expand_keypair(Botan::DilithiumSeedRandomness(std::move(seed)), std::move(mode));
+
+   DilithiumSerializedPrivateKey expanded_from_seed = Dilithium_Algos::encode_keypair(key_pair_from_seed);
+
+   if(expanded_from_seed.get() != expanded) {
+      throw Botan::Decoding_Error("seed and expanded key in ML-DSA serialized key do not match");
+   }
+   return key_pair_from_seed;
+}
```
reneme, line 67: Please rebase to latest master (to pull in #5946) and then:


```suggestion
Botan::DilithiumInternalKeypair decode_seed_plus_expanded(std::span<const uint8_t> key_bits,
                                                          Botan::DilithiumConstants mode) {
   Botan::DilithiumSerializedPrivateKey expanded;
   Botan::DilithiumSeedRandomness seed;
   Botan::BER_Decoder(key_bits)
      .start_sequence()
      .decode(seed, Botan::ASN1_Type::OctetString)
      .decode(expanded, Botan::ASN1_Type::OctetString)
      .end_cons()
      .verify_end();
   // expanded is still needed for the consistency check below, hence no std::move
   const Botan::DilithiumInternalKeypair key_pair =
      Botan::Dilithium_Algos::decode_keypair(expanded, mode);
   const Botan::DilithiumInternalKeypair key_pair_from_seed =
      Botan::Dilithium_Algos::expand_keypair(std::move(seed), std::move(mode));

   const auto expanded_from_seed = Dilithium_Algos::encode_keypair(key_pair_from_seed);

   if(expanded_from_seed != expanded) {
      throw Botan::Decoding_Error("seed and expanded key in ML-DSA serialized key do not match");
   }
   return key_pair_from_seed;
}
```

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -8,20 +8,147 @@
...
+   Botan::secure_vector<uint8_t> seed;
+   Botan::BER_Decoder(key_bits)
+      .start_sequence()
+      .decode(seed, Botan::ASN1_Type::OctetString)
+      .decode(expanded, Botan::ASN1_Type::OctetString)
+      .end_cons()
+      .verify_end();
+   // expanded is still needed for the consistency check below, hence no std::move
+   const Botan::DilithiumInternalKeypair key_pair =
+      Botan::Dilithium_Algos::decode_keypair(Botan::DilithiumSerializedPrivateKey(expanded), mode);
```
reneme, line 57: Do we really need to parse the expanded key? If the expansion is faulty, the sanity-check below will not pass anyway. And we're returning the result of the seed-expansion in the end.

If we'd remove this, we would likely miss out on some more on-point error message that might (or might not) be produced by the parser. I'm quite indifferent, because this won't likely happen on any critical path.

I would say: either remove or add a comment why it should stay.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
```diff
@@ -8,20 +8,147 @@
...
+   const Botan::DilithiumInternalKeypair key_pair_from_seed =
+      Botan::Dilithium_Algos::expand_keypair(Botan::DilithiumSeedRandomness(std::move(seed)), std::move(mode));
+
+   DilithiumSerializedPrivateKey expanded_from_seed = Dilithium_Algos::encode_keypair(key_pair_from_seed);
+
+   if(expanded_from_seed.get() != expanded) {
+      throw Botan::Decoding_Error("seed and expanded key in ML-DSA serialized key do not match");
+   }
```
reneme, line 65: This comparison could be done using a constant-time comparison for good measure, but it would really be cosmetic, I guess. Attacker-controlled private key seems like trouble anyway.

src/lib/pubkey/dilithium/dilithium_common/dilithium.h
```diff
@@ -109,38 +110,80 @@ BOTAN_DIAGNOSTIC_PUSH
...
+       *
+       * @throws Encoding_Error if the key cannot be encoded in @p format
+       *         (see formatted_raw_private_key_bits())
+       */
+      secure_vector<uint8_t> formatted_private_key_bits(MlPrivateKeyFormat format) const override;
 
       std::unique_ptr<PK_Ops::Signature> _create_signature_op(RandomNumberGenerator& rng,
                                                               const PK_Signature_Options& options) const override;
 
+      bool is_mldsa() const;
```
reneme, line 178: Please rename to `is_ml_dsa()`. The public-facing `Dilithium_Mode` uses this naming scheme already, and I'd like to stay consistent. As discussed earlier in the year: this function is likely to disappear in Botan4.

reneme (approved):
This is in good shape now. Thanks for the patience on this. Please address the few nits and adaptions mentioned below and clean up the history. Then, that's good to go from my side.

src/lib/pubkey/dilithium/dilithium_common/dilithium.h
falko-strenzke: OK, done, comes in next commit.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: You are right, it's unneeded, removed the parsing of expanded, length check on the expanded part remains.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: Done.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: Done. Also removed the `Botan::`.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: Done.

src/lib/pubkey/dilithium/ml_dsa/ml_dsa_impl.cpp
falko-strenzke: Done.

PR #5309, opened 2026-02-10
CI/Test: Build and test with `--disable-modules=tls13`

reneme:
### See [here](https://github.com/randombit/botan/issues/2990#issuecomment-3854594155) for an overview of the related pull requests to disentangle TLS 1.2 and 1.3.

In the context of [allowing to disable TLS 1.2 at build time](https://github.com/randombit/botan/issues/2990#issuecomment-3854594155) I think it makes sense to test the library with TLS 1.2 or TLS 1.3 disabled respectively. This adds a boringSSL test suite configuration that doesn't require TLS 1.3 and runs it in CI with a build of Botan that is configured with `--disable-modules=tls13`. Once TLS 1.2 can be disabled independently at build time I'll add this as a test configuration as well.

### DIFF
--- a/src/configs/repo_config.env
+++ b/src/configs/repo_config.env
@@ -16,7 +16,7 @@
 BORINGSSL_REPO="randombit/boringssl"
 
 # The branch in our fork of boringssl that should be used for BoGo tests
-BORINGSSL_BRANCH="rene/runner-20241016"
+BORINGSSL_BRANCH="rene/runner-20241016-1"
 
 # The Android NDK to for the cross platform builds to Android
 ANDROID_NDK="android-ndk-r29"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -269,6 +269,9 @@ jobs:
           - target: minimized
             compiler: gcc
             host_os: ubuntu-24.04
+          - target: no_tls13
+            compiler: gcc
+            host_os: ubuntu-24.04
           - target: policy-bsi
             compiler: gcc
             host_os: ubuntu-24.04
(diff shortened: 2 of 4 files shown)


reneme:
@randombit This is independent of the other TLS related PRs. I'd suggest, let's merge this now and already benefit from it while merging the stack of open PRs. I'll open another one with a TLS 1.3-only test setup at the end.

reneme:
Hang on... disabling TLS 1.2 tests would be a _huge_ list of tests in the config.json, globbing test names only takes us so far. I'm going to add a patch to the `runner.go` to selectively disable tests based on protocol version.

src/configs/repo_config.env
```diff
@@ -16,7 +16,7 @@
 BORINGSSL_REPO="randombit/boringssl"
 
 # The branch in our fork of boringssl that should be used for BoGo tests
-BORINGSSL_BRANCH="rene/runner-20241016"
+BORINGSSL_BRANCH="rene/runner-20241016-1"
```
reneme, line 19: This version of the boringssl test suite now contains [a patch adding `-skip-dtls`, `-skip-tls12` and `-skip-tls13`](https://github.com/randombit/boringssl/commit/9bb38ad585cc6abf131cd9110884114fb3ea3569). That way, our BoGo integration can fairly easily disable most of these tests. Unfortunately, this isn't perfect. Some tests are actually meant to test TLS 1.2 but assume that the shim offers/expects TLS 1.3 or vice versa. These tests still have to be manually disabled via the config.json file, but they are very few and this is hopefully much less brittle.

reneme:
@randombit Ready to go now. Sorry for the back-and-forth.

PR #5312, opened 2026-02-11
FIXME - Make bitref_base::operator bool explicit

KaganCanSit:
Hello,

This PR contains changes made to `FIXME` annotation on the implicit `operator bool()` in `bitref_base` within `bitvector.h`.

```cpp
constexpr operator bool() const noexcept { return is_set(); }
```

> Note: I reviewed the contextual conversion problem triggered by the `!` operator and the `.back()`, `.front()` calls, but I may have missed some points. (I thought there should be a single standard for the writing style here. But I can change that.)

I just want to see how CI is doing and marking it as a draft. I'll review it when I have time, and if you want any corrections, please leave a comment and I'd be happy to help.

Best regards.

### DIFF
--- a/src/lib/pubkey/classic_mceliece/cmce_decaps.cpp
+++ b/src/lib/pubkey/classic_mceliece/cmce_decaps.cpp
@@ -30,7 +30,7 @@ Classic_McEliece_Polynomial Classic_McEliece_Decryptor::compute_goppa_syndrome(
       auto g_alpha = goppa_poly(alphas[i]);
       auto r = (g_alpha * g_alpha).inv();
 
-      auto c_mask = GF_Mask::expand(static_cast<bool>(code_word.at(i)));
+      auto c_mask = GF_Mask::expand(code_word.at(i).is_set());
 
       for(size_t j = 0; j < 2 * params.t(); ++j) {
          syndrome[j] += c_mask.if_set_return(r);
--- a/src/lib/pubkey/classic_mceliece/cmce_field_ordering.cpp
+++ b/src/lib/pubkey/classic_mceliece/cmce_field_ordering.cpp
@@ -305,7 +305,7 @@ Classic_McEliece_Field_Ordering Classic_McEliece_Field_Ordering::create_from_con
       const size_t gap = size_t(1) << std::min(i, 2 * params.m() - 2 - i);
       for(size_t j = 0; j < size_t(n) / 2; ++j) {
          const size_t pos = (j % gap) + 2 * gap * (j / gap);
-         auto mask = CT::Mask<uint16_t>::expand_bool(control_bits[i * n / 2 + j]);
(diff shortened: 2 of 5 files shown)


KaganCanSit:
Hello @reneme, I'm unsure about the right approach here and would appreciate your feedback. (Note: Rebased master)

The `FIXME` comment suggested marking `operator bool` as `explicit`. However, since `bitref_base` has three distinct use cases — bit state querying, CT operations, and bit manipulation — I wanted to verify whether `explicit` alone is sufficient or if a stronger constraint is warranted.

I first applied `explicit` and resolved the resulting compilation errors by migrating call sites to `.is_set()`. Then I went further and deleted `operator bool` entirely to catch any remaining implicit conversion paths. This brought up a few more that obviously couldn't be caught on their own:

```cpp
// cmce_matrix.cpp:210 - implicit conversion via operator!
const bool diag_bit_zero = !mat[diag_pos].at(diag_pos);

// cmce_decaps.cpp:33 - explicit cast bypassed by explicit operator bool
auto c_mask = GF_Mask::expand(static_cast<bool>(code_word.at(i)));
```

And three test sites using the same pattern:
```cpp
bv[0] &= !bv[1];   // test_utils_bitvector.cpp:203
bv[1] |= !bv[2];   // test_utils_bitvector.cpp:214
bv[2] ^= !bv[3];   // test_utils_bitvector.cpp:222
```

All of these are straightforward to fix with `.is_set()`. Given that `is_set()` already provides a clear and explicit alternative, deleting `operator bool` entirely seems like the safer choice — particularly in a CT context where an accidental reduction to `bool` could have subtle consequences. I have not applied this yet and wanted to check your preference before proceeding.

Additionally, I think I can read more easily that there is an assignment in the section below;

```cpp
bv[0] &= !bv[1].is_set();
bv[1] |= !bv[2].is_set();
bv[2] ^= !bv[3].is_set();
```

KaganCanSit:
Hit #5328 and #5334, rebased master.

reneme:
Frankly, just as in #5335, I would argue that the `FIXME` was misplaced there. The whole thing is meant to be a "vector of bits", hence I find it reasonable to allow accessing individual bits with the subscript operator. And that's what the implicit conversion to bool was meant to facilitate.

It just makes for a cleaner API and improves readability at the usage sites by removing visual clutter, in my opinion. Sure, one pays for this convenience with allowing a little "magic" to happen. You're making a valid point regarding constant time mitigations. But in practise the explicit `is_set()` likely isn't solving this either, unfortunately.

KaganCanSit:
Following discussions numbered #5335 and #5312 (this PR), it has been decided to remove only the FIXME explanatory note, while continuing to revert the changes and suppress NOLINT.

As you can see by following the messages in the PR content, both implicit conversions were intentionally designed as discussed with @reneme.

As a result, since it will only be a two-line change, the request here has been closed in favor of #5335 for management and convenience. ✋🏻

PR #5313, opened 2026-02-11
x509: allow AnyExtendedKeyUsage for CA

### LINKED ISSUE #5308: x509cert: allow anyExtendedKeyUsage (2.5.29.37.0) for CA
currently, if certificate has `Extended Key Usage` extension with oid anyExtendedKeyUsage (2.5.29.37.0)
botan ignores it and keeps `m_is_ca_certificate: false` so `X509_Certificate::is_CA_cert()` returns `false`.
https://github.com/randombit/botan/blob/master/src/lib/x509/x509cert.cpp#L234

it would be nice to add some policy to allow certificate to be considered as CA when it has anyExtendedKeyUsage.

related place in RFC (https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.12):
```
   If a CA includes extended key usages to satisfy such applications,
   but does not wish to restrict usages of the key, the CA can include
   the special KeyPurposeId anyExtendedKeyUsage in addition to the
   particular key purposes required by the applications.  Conforming CAs
   SHOULD NOT mark this extension as critical if the anyExtendedKeyUsage
   KeyPurposeId is present.  Applications that require the presence of a
   particular purpose MAY reject certificates that include the
   anyExtendedKeyUsage OID but not the particular OID expected for the
   application.
```
----
example of the root certificate (Neustar Canada uses it to issue intermediate certs for SHAKEN identities signing):
```
-----BEGIN CERTIFICATE-----
MIICmzCCAkKgAwIBAgIUQVJ25K/BQU4Tby6TTgieMvfdS6owCgYIKoZIzj0EAwIw
gZIxCzAJBgNVBAYTAkNBMSkwJwYDVQQKDCBOZXVzdGFyIEluZm9ybWF0aW9uIFNl
cnZpY2VzIEluYzEcMBoGA1UECwwTY21zLWNhLmNjaWQubmV1c3RhcjE6MDgGA1UE
AwwxTmV1c3RhciBDYW5hZGEgQ2VydGlmaWVkIENhbGxlciBJRCBTSEFLRU4gUm9v
dCBDQTAeFw0yMDEwMDgxNDQ5NDlaFw00MDEwMDgxNDQ5NDlaMIGSMQswCQYDVQQG
EwJDQTEpMCcGA1UECgwgTmV1c3RhciBJbmZvcm1hdGlvbiBTZXJ2aWNlcyBJbmMx
HDAaBgNVBAsME2Ntcy1jYS5jY2lkLm5ldXN0YXIxOjA4BgNVBAMMMU5ldXN0YXIg
Q2FuYWRhIENlcnRpZmllZCBDYWxsZXIgSUQgU0hBS0VOIFJvb3QgQ0EwWTATBgcq
hkjOPQIBBggqhkjOPQMBBwNCAASjwCO8A6fzPwY1F5MuaXLpJ3idemAo1q8CdAyQ
XS6Ln4yVkNhghSM+EhKjc7RlBX340zfExlIaIzpyHTZVNvpUo3QwcjAPBgNVHRMB
Af8EBTADAQH/MB8GA1UdIwQYMBaAFEgDf+8c7YwuP4vEBHIWEAP62tkuMA8GA1Ud
JQQIMAYGBFUdJQAwHQYDVR0OBBYEFEgDf+8c7YwuP4vEBHIWEAP62tkuMA4GA1Ud
DwEB/wQEAwIBhjAKBggqhkjOPQQDAgNHADBEAiARILOdPc23+PI4QwMZXudEjoNv
usfK+5RMKwDnQu5LMAIgf5PAGmqtyAdFlRVuJ00qWw7G2+Hw5Oq3NuHYjFZfzKE=
-----END CERTIFICATE-----
```

and just for context the ugly patch that fixes trust chain validation with this cert:
```
diff --git a/src/lib/x509/x509cert.cpp b/src/lib/x509/x509cert.cpp
index 97969c531..07c6cebbc 100644
--- a/src/lib/x509/x509cert.cpp
+++ b/src/lib/x509/x509cert.cpp
@@ -227,9 +227,10 @@ std::unique_ptr<X509_Certificate_Data> parse_x509_cert_body(const X509_Object& o
             const auto server_auth = OID::from_name("PKIX.ServerAuth");
             const auto client_auth = OID::from_name("PKIX.ClientAuth");
             const auto ocsp_sign = OID::from_name("PKIX.OCSPSigning");
+            const auto anyExtendedKeyUsage = OID{2,5,29,37,0};
 
             for(const auto& oid : ext_ku) {
-               if(oid == server_auth || oid == client_auth || oid == ocsp_sign) {
+               if(oid == server_auth || oid == client_auth || oid == ocsp_sign || oid == anyExtendedKeyUsage) {
                   return true;
                }
             }
```
----
not sure how exactly it should be done correctly so creating this issue instead of a merge request.

need your opinion whether anyExtendedKeyUsage support sounds reasonable at all and where to configure such behavior.

furmur:
Resolves #5308

* add new OID 2.5.29.37.0 = X509v3.AnyExtendedKeyUsage
* consider X509v3.AnyExtendedKeyUsage as certificate signing compatible purpose


furmur:
@reneme changed `X509_Certificate::allowed_extended_usage()` as well.

as for RFC vagueness, that's why i was asking in the issue #5308 whether this behavior should be covered by policy but as _randombit_ said it already tolerates omitted extension so no big changes here.

reneme (commented):
Changes look fine to me.

Do you happen to have a certificate handy that we could place in the test suite to exercise those new checks? Perhaps [along those lines](https://github.com/randombit/botan/blob/e7443105fe58e165b9a0a147c923ea60b2a1fc02/src/tests/unit_x509.cpp#L572-L601). I.e. a CA certificate that doesn't have any specific EKUs but just "any". Just place it in "src/tests/data/x509/misc" and add a test ensuring that it recognized as 1) "a CA" and 2) capable of signing CRLs.

furmur:
>  2) capable of signing CRLs, OCSPs or similar.

added test. but cert is considered as not capable for OCSP because `X509_Certificate::allowed_usage()` uses 
`has_ex_constraint()` instead of modified `allowed_extended_usage()`.

maybe it's ok to leave it this way (otherwise should to modify `has_ex_constraint()` as well)

testing cert  ku/eku:
```
X509v3 extensions:
    X509v3 Extended Key Usage: 
        Any Extended Key Usage
    X509v3 Key Usage: critical
        Digital Signature, Certificate Sign, CRL Sign
```

reneme:
> added test. but cert is considered as not capable for OCSP because X509_Certificate::allowed_usage() uses
has_ex_constraint() instead of modified allowed_extended_usage().

I found it surprising that "PKIX.ServerAuth" and "PKIX.ClientAuth" is checked with `allowed_extended_usage` but "PKIX.OCSPSigning" uses `has_ex_constraint`. I did some digging, and that was changed in #3067 as a fix for [CVE-2022-43705](https://github.com/randombit/botan/security/advisories/GHSA-4v9w-qvcq-6q7w). I'm not 100% sure anymore but this was probably meant to enforce the existence of the extended key usage and fail when no EKU-extension is available instead of just letting it pass.

> maybe it's ok to leave it this way (otherwise should to modify has_ex_constraint() as well)

I think, `has_ex_constraint()` should also say "yes" if the new "any" EKU is present.

furmur:
changed:
* `has_ex_constraint()` to return `true` on `X509v3.AnyExtendedKeyUsage` in extended key usage oids
* `allowed_extended_usage()` to use `has_ex_constraint()`

`allowed_usage(Usage_Type::OCSP_RESPONDER)` for certs with `X509v3.AnyExtendedKeyUsage` and without `PKIX.OCSPSigning` now returns `true`.

PR #5317, opened 2026-02-12
add ML-DSA certificate tests with certificates from RFC 9881

falko-strenzke:


src/tests/test_dilithium.cpp
```diff
@@ -9,16 +9,18 @@
  * Botan is released under the Simplified BSD License (see license.txt)
  */
 
+#include "botan/pkix_enums.h"
```
reneme: Please move this into the `#if defined(BOTAN_HAS_DILITHIUM_COMMON)` for consistency and use angle brackets instead of quotes.

reneme (approved):
Nice! Please rebase to latest master to fix the merge conflict and consider the minor comment. Otherwise this looks good to me.

`clang-tidy` might have some opinions regarding const declarations, though.

randombit:
CI failures are relevant

falko-strenzke:
@randombit 
All green now, can we merge this one?

src/tests/test_dilithium.cpp
```diff
@@ -9,16 +9,19 @@
  * Botan is released under the Simplified BSD License (see license.txt)
  */
 
+#include "test_rng.h"
```
randombit: "tests.h" should always be the first include, "test_rng.h" should be in this case under the HAS_DILITHIUM block since the include will not be used if that feature is not available anyway.

src/tests/test_dilithium.cpp
```diff
@@ -9,16 +9,19 @@
...
+   #include <botan/data_src.h>
    #include <botan/dilithium.h>
    #include <botan/hash.h>
    #include <botan/pk_algs.h>
+   #include <botan/pkix_enums.h>
    #include <botan/pubkey.h>
+   #include <botan/x509cert.h>
 
    #include "test_pubkey.h"
-   #include "test_rng.h"
```
randombit, line 21: ^ yeah it should be included right here

randombit (changes requested):
One minor thing on the headers, also needs a rebase+squash, this should probably be a single commit

src/tests/test_dilithium.cpp
```diff
@@ -9,16 +9,19 @@
...
 #include "tests.h"
 
 #if defined(BOTAN_HAS_DILITHIUM_COMMON)
+   #include <botan/data_src.h>
    #include <botan/dilithium.h>
    #include <botan/hash.h>
    #include <botan/pk_algs.h>
+   #include <botan/pkix_enums.h>
    #include <botan/pubkey.h>
+   #include <botan/x509cert.h>
```
randombit: Actually this will fail to compile in a build with ML-DSA but not X509 certs. Both of these new includes should be later and guarded additionally by `BOTAN_HAS_X509`, likewise the test

src/tests/test_dilithium.cpp
falko-strenzke: Done

src/tests/test_dilithium.cpp
falko-strenzke: Done

randombit:
Needs a rebase and possibly CI fixes

PR #5303, opened 2026-02-09
Refactor: Organize TLS Extensions into TLS 1.2 and 1.3 Modules

reneme:
### See [here](https://github.com/randombit/botan/issues/2990#issuecomment-3854594155) for an overview of the related pull requests to disentangle TLS 1.2 and 1.3.

Many extensions are understood and semanticaly relevant by both revisions but some are strongly version specific. This moves the lattter ones into their respective tls12 and tls13 modules. The selection roughly follows RFC 8446 Section 4.2: extensions that aren't listed there (anymore) are classified as specific to TLS 1.2, whereas extensions that are introduced by RFC 8446 and that clearly aren't retrofittable, are classified as specific to TLS 1.3.

If either TLS 1.2 or TLS 1.3 is disabled at build time, such extensions are not parsed by Botan but their content is captured by the Unknown_Extension class instead.

@copilot Again this is mostly moving code around. Please review only the the last commit and put strong empathizes on validating that no relevant implementation changes occurred. Generate a list of locations that didn't just move but where semantic changes were made to the code. This is meant to help a human reviewer with this fairly large patch set.

### Pull Request Dependencies

* #5292
* #5293

PR #5318, opened 2026-02-12
Allow disabling TLS 1.2 at Build Time

reneme:
### See [here](https://github.com/randombit/botan/issues/2990#issuecomment-3854594155) for an overview of the related pull requests to disentangle TLS 1.2 and 1.3.

Since the integration of TLS 1.3 it was possible to disable it at build time while still being able to use TLS 1.2. However, the opposite wasn't possible. Now applications can build with `--disable-modules=tls12` without also loosing support for TLS 1.3. This adds a CI build configuration to exercise this possibility (equivalent to #5309 for TLS 1.3).

See https://github.com/randombit/botan/issues/2990#issuecomment-3854594155

### Pull Request Dependencies

* #5293
* #5303

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -269,6 +269,9 @@ jobs:
           - target: minimized
             compiler: gcc
             host_os: ubuntu-24.04
+          - target: no_tls12
+            compiler: gcc
+            host_os: ubuntu-24.04
           - target: no_tls13
             compiler: gcc
             host_os: ubuntu-24.04
@@ -306,7 +309,7 @@ jobs:
           repository: ${{ env.BORINGSSL_REPO }}
           ref: ${{ env.BORINGSSL_BRANCH }}
           path: ./boringssl
-        if: matrix.target == 'no_tls13'
+        if: matrix.target == 'no_tls12' || matrix.target == 'no_tls13'
 
       - name: Setup Build Agent
         uses: ./source/.github/actions/setup-build-agent
--- a/src/bogo_shim/config_no_tls12.json
+++ b/src/bogo_shim/config_no_tls12.json
@@ -1,0 +1,286 @@
+{
+    "LooseErrorTests": {
+        "AppDataBeforeHandshake": "BoGo expects different error before vs after CCS",
+        "AppDataBeforeHandshake-Empty": "Invalid record message",
+        "ServerHelloBogusCipher": "Unexpected error",
+        "Garbage": "Decoding error",
+        "Resume-Client-CipherMismatch": "Unexpected error",
+        "InvalidECDHPoint-Server": "Unexpected error",
+        "NoSharedCipher": "Unexpected error",
+        "NoSharedCipher-TLS13": "Unexpected error",
+
+        "PartialFinishedWithServerHelloDone": "Unexpected record vs excess handshake data",
+        "HelloRetryRequest-DuplicateCurve-TLS13": "expects 'illegal parameter' but we want to stick with 'decode error'",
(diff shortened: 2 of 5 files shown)


src/tests/test_tls_stream_integration.cpp
```diff
@@ -910,16 +910,16 @@ class SystemConfiguration {
 
 std::vector<SystemConfiguration> get_configurations() {
    return {
+   #if defined(BOTAN_HAS_TLS_12)
       SystemConfiguration("TLS 1.2 only", "allow_tls12=true\nallow_tls13=false", "allow_tls12=true\nallow_tls13=false"),
+      SystemConfiguration(
+         "TLS 1.x server, TLS 1.2 client", "allow_tls12=true\nallow_tls13=false", "allow_tls12=true\nallow_tls13=true"),
+      SystemConfiguration(
+         "TLS 1.2 server, TLS 1.x client", "allow_tls12=true\nallow_tls13=true", "allow_tls12=true\nallow_tls13=false"),
+   #endif
```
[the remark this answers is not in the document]
reneme: Yeah, fair point.

src/lib/tls/info.txt
```diff
@@ -45,14 +45,17 @@ tls_reader.h
...
 ecdh
 ecdsa
 gcm
 hmac
 rng
 sha2_32
 sha2_64
-tls12
 x509
 </requires>
```
reneme, line 60: Side note (not to be addressed here): it would be actually nice to not hard-depend on specific algorithms here. Specialised applications might have a fairly narrow (and predefined) zoo of algorithms that they actually need for TLS within their ecosystem. Nevertheless, this pulls in much more than actually needed.

For instance, TLS will really _need_ just AEAD, RNG, and perhaps things like HKDF and maybe even SHA. But it could certainly work without Finite-Field Diffie-Hellman, RSA (currently required in TLS 1.2) and GCM in some contexts.

On the other hand, this makes the module selection much harder and frustrating for novice users. Often times, missing modules show up only at runtime and are fairly hard to debug. Or catching all relevant platform acceleration modules is tricky and also a maintenance burden when updating.

Perhaps an additional `<recommended>` block of optional module dependencies could help with that? So that one could build along the lines of `--minimized-build --enable-modules=tls13 --enable-recommended-dependencies`.

randombit: > But it could certainly work without Finite-Field Diffie-Hellman, RSA (currently required in TLS 1.2) and GCM in some contexts.

Yeah these are a very long standing todos

```
fc3f52703f doc/todo.rst         (Jack Lloyd  2017-02-07 09:46:02 -0500  66) * Make RSA optional at build time
e29024608f doc/todo.rst         (Jack Lloyd  2016-08-13 11:13:49 -0400  67) * Make finite field DH optional at build time
```

TBH for DH at least I was content to leave it be, considering FFDHE is on the docket to be removed entirely in Botan4.

> Or catching all relevant platform acceleration modules is tricky and also a maintenance burden when updating.

This should just be fixed where the machine specific implementations are enabled unless specifically disabled by the user.

> Perhaps an additional <recommended> block of optional module dependencies could help with that? 

Yeah that would likely be helpful, kind of similar to the `if_available` block in the module policies. If it's in `<recommended>` then it's included unless some other setting (eg explicit `--disable-modules=` or forbidden by module policy) would have prohibited it.

I guess with that the above case of the machine specific implementations could then be resolved via just adding `<recommended>` blocks to each of the base module `info.txt` files.

reneme: I've created a ticket for that: #5353.

PR #5320, opened 2026-02-12
Remove buggy operator!= in bitvector, rely on C++20 rewrite rules

KaganCanSit:
Hello,

While making edits for the `fixme` message in PR #5312, I was trying out some test code. At that point, I noticed that `operator!=` returned the result of `equals_vartime()` without negating it. This caused the expression `a != c` to incorrectly return `true` for equal bitvectors.

Since it has been in the code tree for a long time, I don't think it is actively used anywhere. However, as a precaution, I made the correction and added a test to verify the scope.

Best regards.

Note: Based on @reneme's guidance and information (thanks), the `!=` operator function was removed in accordance with the C++20 rewrite rule. The added test suite was left in place to prevent recurrence.

### DIFF
--- a/src/lib/utils/bitvector/bitvector.h
+++ b/src/lib/utils/bitvector/bitvector.h
@@ -1350,11 +1350,6 @@ bool operator==(const T1& lhs, const T2& rhs) {
    return lhs.equals_vartime(rhs);
 }
 
-template <bitvectorish T1, bitvectorish T2>
-bool operator!=(const T1& lhs, const T2& rhs) {
-   return lhs.equals_vartime(rhs);
-}
-
 namespace detail {
 
 /**
--- a/src/tests/test_utils_bitvector.cpp
+++ b/src/tests/test_utils_bitvector.cpp
@@ -331,6 +331,37 @@ std::vector<Test::Result> test_bitvector_capacity(Botan::RandomNumberGenerator& 
                result.confirm("0 is false", !bv[0]);
                result.confirm("5 is false", !bv[5]);
             }),
+
+      CHECK("binary bitwise and comparison operators",
+            [](Test::Result& result) {
+               Botan::bitvector a(8);
+               a.set(0).set(3);
+               Botan::bitvector b(8);
+               b.set(1).set(3);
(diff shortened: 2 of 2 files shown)


src/lib/utils/bitvector/bitvector.h
```diff
@@ -1352,7 +1352,7 @@ bool operator==(const T1& lhs, const T2& rhs) {
 
 template <bitvectorish T1, bitvectorish T2>
 bool operator!=(const T1& lhs, const T2& rhs) {
-   return lhs.equals_vartime(rhs);
+   return !lhs.equals_vartime(rhs);
 }
```
reneme: Actually, given that C++20 can implicitly create comparison operators you should be fine to just remove this entire `operator!=` implementation and your test should still pass....

After all, if there's no duplicated logic there's also no potential for copy-pasta. 😶

KaganCanSit: Actually, implementing the approach you mentioned is more correct.

Let's proceed as follows:
First, I will remove the comparison operators in the local, compile it and check the tests. If the compiler can handle this, I will remove the addition to the test code and completely delete the operator definitions.

This way, we will achieve a cleaner, background-managed structure.

reneme: What I meant is that you can remove `operator!=` because C++20 compilers can just deduce it from the `operator==` canonically. You won't be able to remove both.

Let's leave the regression test in, after all it would have caught this bug...

KaganCanSit: I made the changes and tested them. Everything is fine. I also edited the PR description and commit message.

Thanks guidance. 🐧🤝

reneme (approved):
Nice! Thanks again. Fixing bugs by removing code. The best fix ever!

KaganCanSit:
> Nice! Thanks again. Fixing bugs by removing code. The best fix ever!

Absolutely, and it comes with learning new things for me. Thanks for the feedback. Being one of the first 14 right now contributors to the project with this PR is quite motivating. 🐧

PR #5323, opened 2026-02-13
Avoid including hash implementation headers into pubkey code

randombit:
SLH-DSA and FrodoKEM do this as well but there fixing it seems to introduce a performance regression and also the impact on compile times is relatively limited since for those, the offending headers are only included into one file.


src/lib/pubkey/dilithium/dilithium_common/dilithium_shake/dilithium_shake_xof.h
```diff
@@ -11,32 +11,43 @@
...
-      static Botan::XOF& XOF(Botan::XOF& xof, std::span<const uint8_t> seed, uint16_t nonce) {
+      static void setupXOF(Botan::XOF& xof, std::span<const uint8_t> seed, uint16_t nonce) {
+         const uint8_t nonce8[2] = {static_cast<uint8_t>(nonce), static_cast<uint8_t>(nonce >> 8)};
+
          xof.clear();
          xof.update(seed);
-         xof.update(store_le(nonce));
-         return xof;
+         xof.update(nonce8);
       }
```
reneme, line 46: I had deliberately left `update_le` and `update_be` out of the XOF API (the Hash API offers them via `Buffered_Computation`) because I thought the composition of `xof.update(store_be(...))` is nicer anyway. But perhaps we should introduce it to avoid ad-hoc endian encoding when `loadstor.h` is too heavy?

randombit: I guess another option here would just be move `setupXOF` into the source file, I can't imagine that it would make any difference perf wise, then `loadstor.h` can be included without it flowing into all of the files which (transitively) include this header

reneme: Your call. :)

Issue #5322, opened 2026-02-13
numthry's deprecation


Chaosvex:
Concerning this pull request: https://github.com/randombit/botan/pull/4560

I've been on older versions of Botan for a while and discovered that `numthry` has been deprecated 'recently'. Although the linked PR implies that there isn't, is there any replacement functionality for things such as `power_mod` without having to go overkill just to get the functionality back? I've been leaning on some of these functions for over a decade across projects (yes, using BigInt as a calculator) and it's unfortunate to see that these primitives are on the way out of the public API.

I haven't found any real discussion on `numthry`, so I suppose I'll be the one to say that it'd be *nice* for this functionality to be kept, even if it's shuffled around elsewhere. At least, unless there's an alternative that I've overlooked.

Thanks!

randombit:
The main reason for deprecating that header is that most of the interfaces are not really suitable for cryptography (at least not efficient cryptography) because they don't have enough context on their arguments. Eg `power_mod` for historical reasons accepts things certainly not useful for cryptography (negative bases, moduli that are a power of 2, etc) but which it nontheless has to handle, and also since we don't know if the inputs are secret or not we have to assume they are. Now we could start making all of that context available (see for example `numbertheory/mod_inv.h` in the source for the different cases of modular inverse which are handled in their own specific ways), but that then commits us further to a myriad of public interfaces that we won't necessarily want to keep forever.

Anyway, I guess it's possible some small subset of the header could remain for calculator-like purposes. Concretely what would your ask here be. `power_mod`, maybe `gcm`, `is_prime`?

Chaosvex:
`power_mod` is the main one that seems like it'd be a pain to reinvent for BigInt if the library were to drop it (and is the one that prompted this issue). `is_prime` is nice but I can live without it.

randombit:
#4666

randombit:
I guess in FFI we already have `botan_mp_powmod`, `botan_mp_gcd`, `botan_mp_isprime` and `botan_mp_mod_inverse` so we're anyway stuck with providing that functionality.

PR #5324, opened 2026-02-13
Don't fully deprecate numthry.h

randombit:
Due to FFI we are anyway committed to supporting generic modular inverse, modular exponentiation, GCD, etc.

Remove future-internal marker from numthry.h but explicitly deprecate a number of specific functions declared within it.

GH #5322


randombit:
cc @Chaosvex

Chaosvex:
Thanks very much! :)

PR #5325, opened 2026-02-13
Include botan/assert.h in auto_rng

Chaosvex:
Fixes a compilation error caused by BOTAN_UNUSED potentially not being transitively included by other modules


Chaosvex:
More info: was using --minimized-build and defining the subset of modules that I use to speed building Botan up (trying to solve another problem with it).

randombit (approved):
Thanks

PR #5328, opened 2026-02-14
Change Test::Result integer and bool predicates to be specifically named

randombit:
Already this existed in a partial way (as test_eq_sz, test_int_eq, and test_is_eq<integer_type>) precisely because the type is not necessarily deducible from literal arguments.

[Part of a series, trying to break things into pieces as best as possible]

### DIFF
--- a/src/tests/test_pkcs11_low_level.cpp
+++ b/src/tests/test_pkcs11_low_level.cpp
@@ -160,7 +160,7 @@ Test::Result test_function(const std::string& name,
 
    // test throw variant
    if(expect_failure) {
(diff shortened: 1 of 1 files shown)


randombit:
> src/tests/test_pkcs11_low_level.cpp:163:7: warning: Potential memory leak [clang-analyzer-cplusplus.NewDeleteLeaks]

Don't see this from with Clang 21 and seems almost certainly a false positive, also I don't really get why this change in particular sets it off.

Update: changing the lambda capture of the function to be capture by reference seems to fix it and anyway is a bit more efficient.

PR #5331, opened 2026-02-14
Remove unused test vector parameters in bitvector test

KaganCanSit:
Hello,

This PR contains the removal of unnecessary test parameters. I noticed this today.

`check_set()` iterates over `bits[0..size()-1]`. The expected set of 20 values was never visited because both the lhs and rhs are 20-bit vectors (valid indices: 0..19), and this expectation was silently skipped without being tested. The correct OR result of the two operands is `{0,1,4,15,16,17,18}`.

I don't think this will have any negative effects. It's just more correct for it not to exist at all than for it to be silently skipped.

Regards.


reneme (approved):
Eagle eyes! Yeah, absolutely. Thanks.

PR #5335, opened 2026-02-14
Remove misplaced FIXME annotations in bitvector.h

KaganCanSit:
Hello,

~This PR resolves the FIXME annotation on the implicitly converting constructor of `bitvector_base` in `bitvector.h`:~

```cpp
// bitvector_base(std::span<const uint8_t> bytes, /* NOLINT(*-explicit-conversions) FIXME */
//                        std::optional<size_type> bits = std::nullopt) : m_bits() {
```

~A call point in the `cmce_keys_internal.cpp` file required updating to use an explicit construct. I have modified it to be consistent with the existing pattern already used for initializing `field_ordering` in the same function.~

**UPDATE: Following the discussion in #5335 and #5312, this PR removes the FIXME annotations from the NOLINT comments. Both implicit conversions are intentional by design, as discussed with @reneme. The NOLINT suppression remains in place; only the FIXME marker is removed.**

If you leave a message for any additional changes or review, I'd be look into it when I have time.
Best regards.

### DIFF
--- a/src/lib/utils/bitvector/bitvector.h
+++ b/src/lib/utils/bitvector/bitvector.h
@@ -279,7 +279,7 @@ class bitvector_base final {
             ~bitref_base() = default;
 
          public:
-            // NOLINTNEXTLINE(*-explicit-conversions) FIXME
+            // NOLINTNEXTLINE(*-explicit-conversions)
             constexpr operator bool() const noexcept { return is_set(); }
 
             constexpr bool is_set() const noexcept { return (m_block & m_mask) > 0; }
@@ -389,7 +389,7 @@ class bitvector_base final {
        * @param bits  The number of bits to be loaded. This must not be more
        *              than the number of bytes in @p bytes.
        */
-      bitvector_base(std::span<const uint8_t> bytes, /* NOLINT(*-explicit-conversions) FIXME */
+      bitvector_base(std::span<const uint8_t> bytes, /* NOLINT(*-explicit-conversions) */
                      std::optional<size_type> bits = std::nullopt) :
             m_bits() {
          from_bytes(bytes, bits);


KaganCanSit:
> Frankly, I would argue that this FIXME was misplaced. Implicit conversion from "array of bytes" to "array of bits" doesn't sound overly surprising to me. The additional verbosity in the construction of the strong type is bearing witness to that fact, in my opinion.

Hello,

My primary focus while making this change was to resolve the FIXME annotation. After reviewing it in more detail following your comment, I'm honestly undecided.  

You're right that the concept of bytes → bits conversion is natural.  Additionally, I see that the increased verbosity in strong types is a real cost. However, I'll just share what comes to mind, and we can decide together:

* This constructor calls `resize()` → `vector::resize()` via `from_bytes`, meaning it performs heap allocation and data copying. Here, unlike general implicit calls, I think the background cost could be overlooked.

* Since there are two `using/typedef` declarations, I expect this will reduce unwanted conversions between them.

```cpp
using secure_bitvector = bitvector_base<secure_allocator>;
using bitvector = bitvector_base<std::allocator>;
```

```cpp
std::span<const uint8_t> raw = get_bytes();
process_sensitive(secure_bitvector(raw));
process_normal(bitvector(raw));
```

* Regarding the strong type verbosity — looking at the actual call site, the `Strong<>` wrapper itself requires explicit construction, so even if the bitvector constructor were implicit the line wouldn't shorten. (`sk_slicer.take()`)

I'll leave the final decision to you. I can either keep the PR as-is (explicit, FIXME resolved), or revert to implicit and just update NOLINT comment.

reneme:
> This constructor calls `resize()` → `vector::resize()` via `from_bytes`, meaning it performs heap allocation and data copying. Here, unlike general implicit calls, I think the background cost could be overlooked.

Yes, one could indeed overlook the allocation. But that's true also when implicitly converting a `const char*` (aka a "string literal") into a `std::string`, for instance.

> Since there are two using/typedef declarations, I expect this will reduce unwanted conversions between them.

I don't really see how the implicit conversion is helping here. Sure, you're forced to make the conversion explicit at the call site. But otherwise? If you were to actually overload the two functions like so:

```C++
void process(bitvector bv);
void process(secure_bitvector sbv);

std::span<const uint8_t> my_bytes;
process(my_bytes);
```

... it wouldn't work with implicit conversions anyway, because [the compiler would complain about ambiguity](https://godbolt.org/z/vzsvxWf1s).

> Regarding the strong type verbosity — looking at the actual call site, the `Strong<>` wrapper itself requires explicit construction, so even if the bitvector constructor were implicit the line wouldn't shorten.

Instead of `StrongSecureBitvector(secure_bitvector(my_bytes))` you just have to write `StrongSecureBitvector(my_bytes)`. I don't understand your argument here.

> I can either keep the PR as-is (explicit, FIXME resolved), or revert to implicit and just update NOLINT comment.

I would vote for removing the "FIXME" in the linter-annotation, but I'm ready to discuss this further if you have more questions or points.

KaganCanSit:
> I would vote for removing the "FIXME" in the linter-annotation, but I'm ready to discuss this further if you have more questions or points.

I'll update it like this when I have time, thanks for the guidance.

Update: I wasn't available during the day to review it calmly. Now that I've looked at it, your perspective seems more accurate. I'll make the necessary changes and merge them into this branch. I'll also write an explanation for the other PR request. I don't think two separate PRs are necessary.

KaganCanSit:
I made the necessary changes on this development branch and rebased master.

Since it's only a two-line change, I closed branch #5312 in favor of this development branch for management and convenience.

I updated the PR title and commit message. I don't think the branch name will be an issue after merging.

Thank you again @reneme for sharing your perspective and knowledge. At least someone else won't have to spend time on the incorrect FIXME annotation here.

reneme (approved):
Thanks! And sorry for the time that went into the first solution.

KaganCanSit:
> Thanks! And sorry for the time that went into the first solution.

It doesn't matter, I thank you for sharing your knowledge. As long as you share your knowledge, perspective and experience, I can try out new content/code/language. This is how we move the project forward.

PR #5336, opened 2026-02-14
Refactor(XMSS): Remove atomic header file, replace Atomic<T> wrapper with std::atomic<T>

KaganCanSit:
Hello,

This PR contains code changes for `XMSS` by completely removing the `atomic.h` header file.

The initial goal was to address the `FIXME` message by marking `operator T()` as `explicit`. However, during the code changes, I noticed that there were no copying operations in the `atomic.h` class description. I also realized that the wrapper functions I wrote for the calls directly overlapped with the contents of the `atomic` object.

Then I reviewed some past commits. As far as I understand _**(it's now 11:45 PM and I've been on GitHub all day**_), the `Atomic<T>` wrapper in `atomic.h` was originally introduced to make `std::atomic<T>` copyable for use in `std::vector` containers. Over time, usage changed to here, `shared_ptr` already provides the indirection needed for container compatibility, making the wrapper's copy/move semantics unnecessary.

Accordingly, I completely removed the `atomic.h` file and replaced all references to `Atomic<size_t>` with `std::atomic<size_t>`.

I don't expect any changes here. However, I will review the CI and marked this PR draft. Feel free to leave comments for feedback or suggestions. I'll review them when I have time.

Best regards.


randombit:
Thanks, your assessment here is completely correct. Due to having to handle some concurrency bugs the index registry needs to be modified anyway, rather than end up with conflicts on either side I've made a similar change in #5382 and so am closing this PR.

PR #5347, opened 2026-02-16
Changes to reduce dependencies in TLS sources/headers

randombit:


src/lib/tls/tls_session.h
```diff
@@ -41,19 +47,17 @@ class BOTAN_PUBLIC_API(3, 0) Session_Base {
...
-            m_extended_master_secret(extended_master_secret),
-            m_encrypt_then_mac(encrypt_then_mac),
-            m_peer_certs(std::move(peer_certs)),
-            m_peer_raw_public_key(std::move(peer_raw_public_key)),
-            m_server_info(std::move(server_info)) {}
+                   Server_Information server_info);
+
+      Session_Base(const Session_Base& other);
+      Session_Base& operator=(const Session_Base& other);
+
```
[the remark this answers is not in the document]
randombit: We are not ABI stable. If someone wants us to be ABI stable they can start paying me a million dollars a year to care.

src/lib/tls/tls13/tls_messages_13.h
```diff
@@ -173,8 +179,19 @@ class BOTAN_UNSTABLE_API Certificate_13 final : public Handshake_Message {
...
+            Certificate_Entry& operator=(Certificate_Entry&& other) noexcept;
+
+            ~Certificate_Entry();
+
          private:
-            std::optional<X509_Certificate> m_certificate;
+            /* This is always either empty or not; would have been std::optional but that requires
+               that the declaration be visible, which we are trying to avoid here
+            */
+            std::vector<X509_Certificate> m_certificate;
```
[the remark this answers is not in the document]
reneme: I'm with Copilot here.

Isn't that taking it a little far? Maybe use a `unique_ptr` for such things so that the extra allocation becomes obvious at least. Also, this won't give the false impression of a 0..n relation.

src/lib/tls/tls_callbacks.h
```diff
@@ -24,6 +23,8 @@
 
 namespace Botan {
 
+enum class Signature_Format : uint8_t;
+enum class Usage_Type : uint8_t;
```
reneme, line 27: The `pkix_enums.h` is merely a bunch of enum declarations (+ two transitive headers with some `#define ...` soup). How expensive can it be? (Or am I going to be surprised? 😶)

src/lib/tls/tls_callbacks.h
```diff
@@ -414,6 +415,9 @@ class BOTAN_PUBLIC_API(2, 0) Callbacks /* NOLINT(*-special-member-functions) */
        *
        * @returns the shared secret both in plaintext and encapsulated with
        *          @p encoded_public_key.
+       *
+       * TODO(Botan4) change this return type to something else so the pubkey.h
+       * dependency is removed
```
reneme, line 420: Also take the `encoded_public_key` as a `std::span` (here and in many other places in this header). Back when we designed most of those we hadn't decided on C++17/20, yet.

src/lib/tls/tls13/tls_messages_13.h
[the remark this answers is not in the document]
randombit: Yeah I guess that works too

src/lib/tls/tls_callbacks.h
randombit: It's quite cheap (at least it is if you are otherwise including `<string>` anyway, which is the case here) but `tls_callbacks.h` is included into 31 TUs, and the extra include is easily avoided so ...

src/lib/tls/tls13/tls_messages_13.h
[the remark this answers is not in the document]
randombit: Switched to `unique_ptr`

src/lib/tls/tls_callbacks.h
randombit: Done, I also checked over the other declarations here and added TODOs for Botan4

src/lib/tls/tls12/tls_record.h
KaganCanSit: I think this forward declaration `class Callbacks;` is not necessary. More cleanup.

src/lib/tls/tls_client.cpp
```diff
@@ -11,6 +11,7 @@
 #include <botan/tls_client.h>
```
KaganCanSit, line 11: Forward declaration of `class Handshake_IO` in the `tls_client.h` file is unnecessary.

src/tests/test_tls_rfc8448.cpp
```diff
@@ -38,6 +38,7 @@
    #include <botan/tls_server.h>
    #include <botan/tls_session_manager.h>
```
KaganCanSit, line 40: I think the forward declaration of class `Session_Handle` in the `tls_session_manager.h` file is unnecessary. Beacuse define comes `tls_session_id.h`.

src/tests/test_tls.cpp
```diff
@@ -17,10 +17,13 @@
    #include <botan/tls_session.h>
    #include <botan/tls_signature_scheme.h>
```
KaganCanSit, line 18: In the `tls_signature_scheme.cpp`, `#include <algorithm>` is unnecessary.

KaganCanSit (commented):
I noticed a few things while checking. Although I'm not entirely sure, they can be checked and added to this branch.

I hope this helps.

randombit:
Thanks @KaganCanSit

PR #5356, opened 2026-02-18
Refactor: Generalize the SEC.1 decoder

reneme:
As [promised a few weeks ago](https://github.com/randombit/botan/pull/5268#discussion_r2735309242). This decomposes SEC1-parsing from the context-specific interpretation of the values (point decompression, point-on-curve check, ...).

There's now an individually testable `sec1_decode()` function that passes an instance of `SEC1_Identity`, `SEC1_Compressed`, or `SEC1_Full` to a user-provided handler function. Call sites can conveniently use the `Botan::overloaded{}` helper to implement this handler.

This also centralizes the handling of the deprecated "hybrid" format. Before this was supported in some places but not everywhere. I would argue that keeping support for it might now not even be a real maintenance burden anymore.


randombit:
Closing in favor of 5725

PR #5358, opened 2026-02-18
Implementation of XMSS^MT

### LINKED ISSUE #5256: Implementation of XMSS^MT
MTG will implement XMSS^MT in Botan as part of the work for the BSI in project P663.
The implementation will follow RFC 8391 and NIST SP800-208.

We plan to implement XMSS^MT as a new module in `src/lib/pubkey/xmssmt/`.
On an abstract level, we will add the three algorithms `XMSSMT_keyGen` (Algorithm 10' from SP800-208), and `XMSSMT_sign`, and `XMSSMT_verify` (Algorithms 16 and 17 from RFC 8391) to the existing XMSS functionality.

Since XMSS is essentially a special case of XMSS^MT with one layer, we expect to be able to use much of the existing XMSS code. However, we expect that some code changes and refactoring will be required s.t. the existing XMSS functionality accomodates the more general multi-tree case.

For the state management we intend to directly use the `XMSS_Index_Registry` class (for XMSS^MT the prf/seed equally identify a private key).
For testing, we plan to follow the structure and approach of the existing XMSS tests.

Any feedback is welcome.

TJ-91:
This implements XMSS^MT as a separate module which re-uses internal XMSS functionality.

I refactored and extended XMSS as far as is needed to not re-implement any XMSS functionality. I moved some core XMSS algorithms like tree_hash to `xmss_core_ops.cpp` and made it independent of XMSS keys and instead pass the required data.

Some noteworthy changes:

* ~~I changed the XMSS Index Registry to use `uint64_t` as leaf index since XMSS^MT has larger trees (up to a total height of 60). XMSS^MT simply uses the same registry and that should be no problem: The seed and the prf are equally used in XMSS^MT and the keys can be uniquely identified by it.~~ (outdated)
* in `test_pubkey.cpp` I changed 
  ```diff
  -        result.test_sz_lt("Key has reasonable estimated strength (upper)", key.estimated_strength(), 512);
  +        result.test_sz_lte("Key has reasonable estimated strength (upper)", key.estimated_strength(), 512);
  ```
  It seems like I added the first test where an algorithm of strength 512 is actually used and using "lt" seems to be a typo.

Further notes:
* ~~The X.509-related stuff is not complete. PQC certificate formats will be implemented in future PRs (as part of the project P663 for the German BSI).~~ X.509 changes have been cherry-picked from #5690
* Are there any opinions on the speed and number of XMSS^MT tests? I added the tests from `test_xmssmt.cpp` to slow tests (ci) and generally left out the parameter sets that have large XMSS subtrees (height 20).
* For the original XMSS code, `size_t` is used for the leaf index which can be more than 16 bits. Is there any particular reason why one wouldn't simply use `uint32_t` or similar?

closes #5256


TJ-91:
@randombit @reneme I removed the draft status and welcome any feedback

randombit:
> For the original XMSS code, size_t is used for the leaf index which can be more than 16 bits. Is there any particular reason why one wouldn't simply use uint32_t or similar?

No I think this should be a specific length, probably `uint64_t` since that's what is used in the new state index (#5382) currently XMSS just casts those values down to `size_t`.

randombit:
Sorry about all the merge conflicts btw, this recent work was unplanned, the issue of errors on concurrent usage was only recently brought to my attention and that's what is driving the XMSS patches.

src/lib/pubkey/xmssmt/xmssmt_privatekey.cpp
```diff
@@ -0,0 +1,257 @@
...
+namespace {
+
+// fall back to raw decoding for previous versions, which did not encode an OCTET STRING
+secure_vector<uint8_t> extract_raw_private_key(std::span<const uint8_t> key_bits,
+                                               const XMSSMT_Parameters& xmssmt_params) {
+   secure_vector<uint8_t> raw_key;
+
+   // The public part of the input key bits was already parsed, so we can
+   // decide depending on the buffer length whether this must be BER decoded.
+   if(key_bits.size() == xmssmt_params.raw_private_key_size()) {
```
falko-strenzke: The fallback decoding is not needed, since there are no previous XMSS^MT private keys.

TJ-91: true, it's the same for the public keys but I will then need to encode the test vectors accordingly.

src/lib/pubkey/xmssmt/xmssmt_publickey.cpp
```diff
@@ -0,0 +1,118 @@
...
+   uint32_t raw_id = 0;
+   for(size_t i = 0; i < 4; i++) {
+      raw_id = ((raw_id << 8) | raw_key[i]);
+   }
+
+   return static_cast<XMSSMT_Parameters::xmssmt_algorithm_t>(raw_id);
+}
+
+// fall back to raw decoding
+// TODO: in contrast to XMSS there are no old versions to support, however, the test vectors are decoded as raw keys (no octet string)
```
falko-strenzke: Yes, there should be no fallback decoding for XMSS^MT in my view.

src/lib/pubkey/xmssmt/xmssmt_signature_operation.cpp
```diff
@@ -0,0 +1,143 @@
...
+   XMSS_Address adrs;
+   const XMSSMT_Parameters& params = xmssmt_priv_key.xmssmt_parameters();
+   std::vector<XMSS_TreeSignature> tree_sigs;
+
+   uint64_t idx_tree = m_leaf_idx;
+
+   secure_vector<uint8_t> node = msg_hash;
+   adrs.set_type(XMSS_Address::Type::OTS_Hash_Address);
+   for(size_t i = 0; i < params.tree_layers(); i++) {
+      const uint32_t idx_leaf = (idx_tree & ((1 << params.xmss_tree_height()) - 1));
```
falko-strenzke: Here and everywhere else where we the assumption is made that `params.xmss_tree_height()` is larger than 0 I suggest to make a corresponding `BOTAN_ASSERT()`

src/lib/pubkey/xmssmt/xmssmt_signature_operation.cpp
```diff
@@ -0,0 +1,143 @@
...
+void XMSSMT_Signature_Operation::initialize() {
+   // return if we already initialized and reserved a leaf index for signing.
+   if(m_is_initialized) {
+      return;
+   }
+
+   secure_vector<uint8_t> index_bytes;
+   // reserve leaf index so it can not be reused by another signature
+   // operation using the same private key.
+   m_leaf_idx = static_cast<uint64_t>(m_priv_key.reserve_unused_leaf_index());
```
falko-strenzke: Cast is redundant (clang-tidy reports that).

src/lib/pubkey/xmssmt/xmssmt_verification_operation.cpp
```diff
@@ -0,0 +1,85 @@
...
+ * (C) 2026 Johannes Roth
+ *
+ * Botan is released under the Simplified BSD License (see license.txt)
+ **/
+
+#include <botan/internal/xmssmt_verification_operation.h>
+
+#include <botan/internal/xmss_core.h>
+#include <botan/internal/xmss_tools.h>
+#include <array>
```
falko-strenzke: Not needed acc. to clang-tidy.

src/lib/pubkey/xmssmt/xmssmt_parameters.h
```diff
@@ -0,0 +1,197 @@
...
+ * [1] XMSS: Extended Hash-Based Signatures,
+ *     Request for Comments: 8391
+ *     Release: May 2018.
+ *     https://datatracker.ietf.org/doc/rfc8391/
+ * [2] Recommendation for Stateful Hash-Based Signature Schemes
+ *     NIST Special Publication 800-208
+ *     Release: October 2020.
+ *     https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-208.pdf
+ **/
+class BOTAN_PUBLIC_API(2, 0) XMSSMT_Parameters {
```
falko-strenzke:
```suggestion
class BOTAN_PUBLIC_API(3, 12) XMSSMT_Parameters {
```

And everywhere else also.

src/lib/pubkey/xmssmt/xmssmt_verification_operation.cpp
TJ-91: Does this refer to `#include <array>`? IIRC without it some Windows CI fails

TJ-91:
I squashed before rebasing to make my life (a lot) easier and now switched to only allowing the ASN.1 OCTET STRING encoding

src/lib/pubkey/xmssmt/xmssmt_verification_operation.cpp
falko-strenzke: I see, forget about my comment and let the CI do the job ;-)

src/lib/pubkey/xmssmt/xmssmt.h
```diff
@@ -0,0 +1,236 @@
...
+      /**
+       * Creates a new XMSS^MT public key for the chosen XMSS^MT signature method.
+       * New public and prf seeds are generated using rng. The appropriate WOTS
+       * signature method will be automatically set based on the chosen XMSS
+       * signature method.
+       *
+       * @param xmssmt_oid Identifier for the selected XMSS^MT signature method.
+       * @param rng A random number generator to use for key generation.
+       **/
+      XMSSMT_PublicKey(XMSSMT_Parameters::xmssmt_algorithm_t xmssmt_oid, RandomNumberGenerator& rng);
```
falko-strenzke: This constructor is obviously intended for the generation of a public key from a seed that at the same time can be used to generate an XMSS^MT private key. This needs to be documented here, specifically the reference to the corresponding constructor for the private key.

TJ-91: You are right. I now documented it and made the constructor `protected`. That should make the intent more clear. I also changed it for the XMSS code where I copied that pattern from.

TJ-91:
I now added `CT::poison` to the private seed/prf values for XMSS and XMSS^MT.
I unpoison when the values become public: WOTS+ Signature and WOTS+ Public Key components, as well as signature randomness.

In my understanding this should cover all cases but if a more sophisticated approach is required please let me know.

As expected, no leaks are found (I tested the example programs since they only execute one relatively fast parameter set)

randombit:
@TJ-91 sorry have not and will not have time to review this in time for 3.12 (being released next Tuesday), bumped to 3.13

TJ-91:
With https://github.com/randombit/botan/pull/5358/commits/0906d721208bd61f34b9a32a36861f7adb161bd1 there is a pipeline failure that I cannot make sense of. Probably it has nothing to do with my change and is just a hiccup. I do not know how to restart the check (or whether I have permission for that)

falko-strenzke:
> With [0906d72](https://github.com/randombit/botan/commit/0906d721208bd61f34b9a32a36861f7adb161bd1) there is a pipeline failure that I cannot make sense of. Probably it has nothing to do with my change and is just a hiccup. I do not know how to restart the check (or whether I have permission for that)

I restarted it with debug-logging enabled.

src/lib/pubkey/xmssmt/xmssmt.h
```diff
@@ -0,0 +1,242 @@
...
+
+template <typename>
+class Atomic;
+
+/**
+ * Determines how WOTS+ private keys are derived from the XMSS^MT private key
+ */
+
+/**
+ * An XMSS: Extended Hash-Based Signature private key.
```
sehlen-bsi: XMMS^MT

src/lib/pubkey/xmssmt/xmssmt_privatekey.cpp
```diff
@@ -0,0 +1,268 @@
...
+      void set_unused_leaf_index(uint64_t idx) {
+         if(idx >= (1ULL << m_xmssmt_params.tree_height())) {
+            throw Decoding_Error("XMSS^MT private key leaf index out of bounds");
+         } else {
+            Stateful_Key_Index_Registry::global().set_index_lower_bound(m_keyid, idx);
+         }
+      }
+
+      uint64_t reserve_unused_leaf_index() {
+         const uint64_t idx = Stateful_Key_Index_Registry::global().reserve_next_index(m_keyid);
```
sehlen-bsi, line 131: reserve_next_index() is called first. Afterwards it is checked whether the returned index is beyond total_number_of_signatures(). If the key is exhausted, this seems to advance the registry past the valid range before throwing. Please check.

TJ-91: You are right, the index is advanced. But the implementation will not use it to create a signature. The registry doesn't know the limitations so you can advance it further. The actual check is done in the XMSS^MT code afterwards.

There is also a test in `src/tests/test_xmssmt.cpp` in the `xmssmt_statefulness()` method that checks that the state can be exhausted and no more signature is generated.

I think you are onto something, though, and the index registry is not well designed and has a fatal flaw: You advance the index. The index is a uint64_t value, i.e., an unsigned integer. If you repeatedly try to sign with an exhausted state, you will repeatedly increase the counter and repeatedly refuse to sign. Until you overflow the integer. Admittedly this will take a large amount of tries. But it will happen at some point and you will again sign with the first leaf index, i.e., reuse a OTS key.

A cleaner approach is to give the limit to the index registry and only advance if the limit is not reached.

I will make an issue for this as it's not something that should be fixed within this PR and is not primarily part of the XMSS^MT implementation but the index registry implementation. Thanks for questioning this!

TJ-91: addressed in #5662

falko-strenzke:
The unit test `xmssmt_sign` currently takes almost 5 minutes (without specifying long tests). This has to be reduced.

```
Tests complete ran 3150 tests in 275.75 sec all tests ok
```

src/tests/test_xmssmt.cpp
```diff
@@ -0,0 +1,219 @@
...
+   #include <botan/internal/loadstor.h>
+#endif
+
+namespace Botan_Tests {
+
+namespace {
+
+#if defined(BOTAN_HAS_XMSSMT_RFC8391)
+
+class XMSSMT_Signature_Tests final : public PK_Signature_Generation_Test {
```
falko-strenzke, line 29: Even without "long tests", this test takes almost 5 minutes on my machine, this must be brought down to about 10s.

> Tests complete ran 3150 tests in 275.75 sec all tests ok

TJ-91: I now simply remove most parameter sets from the normal tests in c86ee57

src/lib/pubkey/xmssmt/xmssmt_publickey.cpp
```diff
@@ -0,0 +1,100 @@
...
+                   "XMSS^MT: unexpected byte length of public seed");
+}
+
+std::vector<uint8_t> XMSSMT_PublicKey::raw_public_key_bits() const {
+   return concat<std::vector<uint8_t>>(store_be(static_cast<uint32_t>(m_xmssmt_params.oid())), m_root, m_public_seed);
+}
+
+std::vector<uint8_t> XMSSMT_PublicKey::public_key_bits() const {
+   // Contrary to the private key, the public key is not wrapped in an ASN.1
+   // OCTET STRING (see RFC 9802 Section 2.3)
```
weitkaemper-bsi:
```suggestion
   // OCTET STRING (see RFC 9802 Section 5.3)
```

randombit:
@TJ-91 needs a rebase to address some merge conflicts

It may be worth splitting this up into 2-3 PRs eg

- One adding the MT OIDs (easy merge)
- One performing the necessary XMSS refactorings without adding MT (relatively easy merge and once merged likely eliminates most future merge conflicts)
- Final one (this PR basically) actually adding MT support

TJ-91:
@randombit I rebased it and split it into two commits: One for XMSS, one for XMSS^MT. This PR contains the full changes (both commits). I created PR #5931 for the standalone XMSS changes (only the XMSS commit).

PR #5359, opened 2026-02-19
Add a test covering concurrent PK operations

randombit:


src/tests/test_concurrent_pk.cpp
```diff
@@ -0,0 +1,440 @@
...
+
+namespace Botan_Tests {
+
+#if defined(BOTAN_HAS_PUBLIC_KEY_CRYPTO) && defined(BOTAN_TARGET_OS_HAS_THREADS)
+
+/*
+* Test that public key operations (signing, verification, encryption,
+* decryption, KEM) with a shared key from multiple threads produce
+* correct results without data races.
+*/
```
reneme, line 27: Nit: The listing in brackets is missing "key agreement".

Perhaps test concurrent keygen as well? Less likely that we'd accidentally use some shared state there, but while we're here...

src/tests/test_concurrent_pk.cpp
```diff
@@ -0,0 +1,429 @@
...
+            auto privkey = tc.try_create_key(*rng);
+            if(!privkey) {
+               auto result = tc.result("signing");
+               result.test_note("Skipping due to missing algorithm", tc.algo_name());
+               results.push_back(std::move(result));
+               continue;
+            }
+            auto pubkey = privkey->public_key();
+            results.push_back(test_concurrent_signing(tc, *privkey, *pubkey));
+            results.push_back(test_concurrent_verification(tc, *privkey, *pubkey));
```
reneme: Maybe add a `tc.skip()` that encapsulates the result creation and "skip note"? Would avoid re-doing that for every test group and `tc.result()` could probably become private. Also, perhaps an `} else {` branch instead of continue.

src/tests/test_concurrent_pk.cpp
```diff
@@ -0,0 +1,429 @@
...
+         }
+      }
+
+      void concurrent_encryption_tests(std::vector<Test::Result>& results) {
+         const std::vector<ConcurrentPkTestCase> test_cases = {
+            ConcurrentPkTestCase("RSA", "1536", "OAEP(SHA-256)"),
+            ConcurrentPkTestCase("ElGamal", "modp/ietf/1536", "PKCS1v15"),
```
reneme, line 411: ECIES perhaps? (Isn't that modeled as an asymmetric algorithm that allows encryption?)

randombit: ECIES would be modeled as that in a theoretical sense but in practice our interface for ECIES is pretty weird and doesn't really fit into what's happening here, it would have to be a completely separate test flow.

src/tests/test_concurrent_pk.cpp
randombit: Added keygen

src/tests/test_concurrent_pk.cpp
randombit: Yeah this was awkward I changed to `if(auto privkey = ...) { ... } else { ... }`

randombit: Added a helper for the missing algorithm case as well

src/tests/test_concurrent_pk.cpp
randombit: It would be worthwhile to have the ECIES code does some relatively funky stuff, but I'm not going to do it right now. I left a todo comment

PR #5360, opened 2026-02-19
Add the math perf tests to the coverage build run

randombit:
These should not be too slow even with coverage enabled.


randombit:
Yeah even with coverage tracking on it's fine

>   INFO: Ran cli_speed_math_tests in 3.20 sec

PR #5361, opened 2026-02-19
Fix FrodoKEM concurrency issue

randombit:
The FrodoKEMConstants class had an unique_ptr<XOF> which was shared, which meant if multiple threads tried to encap/decap with the same key without some form of external locking, they would produce bad results.

In general Botan's position is that using an object in multiple threads without locking is not supported. However this case is a bit subtle since the key is being passed as a const reference and there is no obvious reason it shouldn't work. And indeed it does (and likely has for many years) been ok to do this with ECDSA, RSA, and most other algorithms - for FrodoKEM to be a special case here is certainly violating the principle of least astonishment.

### DIFF
--- a/src/lib/pubkey/frodokem/frodokem_common/frodo_constants.cpp
+++ b/src/lib/pubkey/frodokem/frodokem_common/frodo_constants.cpp
@@ -97,9 +97,8 @@ FrodoKEMConstants::FrodoKEMConstants(FrodoKEMMode mode) : m_mode(mode), m_len_a(
 
 FrodoKEMConstants::~FrodoKEMConstants() = default;
 
-XOF& FrodoKEMConstants::SHAKE_XOF() const {
-   m_shake_xof->clear();
-   return *m_shake_xof;
+std::unique_ptr<XOF> FrodoKEMConstants::create_xof() const {
+   return m_shake_xof->new_object();
 }
 
 }  // namespace Botan
--- a/src/lib/pubkey/frodokem/frodokem_common/frodo_constants.h
+++ b/src/lib/pubkey/frodokem/frodokem_common/frodo_constants.h
@@ -76,9 +76,7 @@ class BOTAN_TEST_API FrodoKEMConstants final {
 
       FrodoDomainSeparator keygen_domain_separator() const { return FrodoDomainSeparator({0x5F}); }
 
-      // TODO: those aren't actually const. We worked around some constness
-      //       issues when playing with the XOFs that are residing in this class.
-      XOF& SHAKE_XOF() const;
+      std::unique_ptr<XOF> create_xof() const;
 
    private:
       FrodoKEMMode m_mode;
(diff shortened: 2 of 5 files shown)


src/lib/pubkey/frodokem/frodokem_common/frodo_constants.cpp
```diff
@@ -91,15 +91,12 @@ FrodoKEMConstants::FrodoKEMConstants(FrodoKEMMode mode) : m_mode(mode), m_len_a(
...
-   m_shake_xof = XOF::create_or_throw(m_shake);
 }
 
 FrodoKEMConstants::~FrodoKEMConstants() = default;
 
-XOF& FrodoKEMConstants::SHAKE_XOF() const {
-   m_shake_xof->clear();
-   return *m_shake_xof;
+std::unique_ptr<XOF> FrodoKEMConstants::create_xof() const {
+   return XOF::create_or_throw(m_shake);
```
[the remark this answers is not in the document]
randombit: Ah interesting point. We might want to retain a prototype object here and then have `create_xof` just return `m_shake->new_object()`....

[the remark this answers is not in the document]
randombit: Done, didn't seem to make much difference on perf

src/lib/pubkey/frodokem/frodokem_common/frodo_constants.cpp
```diff
@@ -97,9 +97,8 @@ FrodoKEMConstants::FrodoKEMConstants(FrodoKEMMode mode) : m_mode(mode), m_len_a(
 
 FrodoKEMConstants::~FrodoKEMConstants() = default;
 
-XOF& FrodoKEMConstants::SHAKE_XOF() const {
-   m_shake_xof->clear();
-   return *m_shake_xof;
+std::unique_ptr<XOF> FrodoKEMConstants::create_xof() const {
+   return m_shake_xof->new_object();
 }
```
reneme, line 102: We don't actually need the `m_shake_xof` instance anymore, it can be removed entirely. Instead, just create a fresh XOF instance here.

```suggestion
std::unique_ptr<XOF> FrodoKEMConstants::create_xof() const {
   return XOF::create_or_throw(m_shake);
}
```

randombit: See copilot's earlier comment.

Also it seems at least some of the time cloning a prototype can be quite a bit faster that going through the string based lookup (no parsing, etc) though in  a quick test it seemed about equivalent here.

randombit: [Possibly because FrodoKEM is comparatively much slower than ML-KEM so the excess cost of `create_or_throw` is swamped by the lattice work]

reneme: Oh. Wow. The machines are getting ahead of us. Fair.

PR #5362, opened 2026-02-19
store pqcrystals polynomial in secure memory

falko-strenzke:


reneme:
Note that the `PolynomialVector` class is tightly coupled with the `Polynomial` class in a sense that it can borrow its byte storage (that's what the redundant `std::span` is for). I.e. also the `std::vector` in the `PolynomialVector` should be replaced by `secure_vector` then.

randombit:
How much does this affect performance?

randombit:
Actually @reneme isn't this always of known length namely `Trait::N`? Why not a `std::array`?

reneme:
> Actually @reneme isn't this always of known length namely Trait::N? Why not a std::array?

This was done to allow the `Polynomial` class to borrow its storage from the `PolynomialVector`, in which case the array would just be wasted stack space. Essentially, instances of polynomials can be free-standing (owning their own storage in the `std::vector`) or part of a vector of polynomials where the storage is coalesced into a single (appropriately larger) `std::vector` in the outer `PolynomialVector` instance. I'm not 100% sure anymore why we added this complexity, frankly. [I believe it was faster](https://github.com/randombit/botan/pull/4024).

Anyway, the computation is always implemented over the `std::span<T, Trait::N>` to let the compiler leverage the statically known bounds.

falko-strenzke:
> How much does this affect performance?

I tried to measure performance with the speed test. The natural variation of the timings is so large that observing any performance impact is impossible. If the performance aspect is really so important here, I would have to ask for the preferred exact method to measure it.

PR #5367, opened 2026-02-20
Fix Kyber/ML-KEM concurrency issues

randombit:
The key objects had shared mutable state which would cause invalid outputs if the same key was used across multiple threads.

In general sharing Botan objects across threads without external locking is not supported, but this case is a bit surprising, and easy to support.

### DIFF
--- a/src/lib/pubkey/kyber/kyber_common/kyber_algos.cpp
+++ b/src/lib/pubkey/kyber/kyber_common/kyber_algos.cpp
@@ -384,10 +384,14 @@ KyberPolyMat sample_matrix(StrongSpan<const KyberSeedRho> seed, bool transposed,
 
    KyberPolyMat mat(mode.k(), mode.k());
 
+   const auto& sym = mode.symmetric_primitives();
+   std::unique_ptr<Botan::XOF> xof;
+
    for(uint8_t i = 0; i < mode.k(); ++i) {
       for(uint8_t j = 0; j < mode.k(); ++j) {
          const auto pos = (transposed) ? std::tuple(i, j) : std::tuple(j, i);
-         sample_ntt_uniform(mat[i][j], mode.symmetric_primitives().XOF(seed, pos));
+         sym.setup_XOF(xof, seed, pos);
+         sample_ntt_uniform(mat[i][j], *xof);
       }
    }
 
--- a/src/lib/pubkey/kyber/kyber_common/kyber_algos.h
+++ b/src/lib/pubkey/kyber/kyber_common/kyber_algos.h
@@ -91,7 +91,12 @@ class PolynomialSampler {
       }
 
    private:
-      KyberSamplingRandomness prf(size_t bytes) { return m_mode.symmetric_primitives().PRF(m_seed, m_nonce++, bytes); }
+      KyberSamplingRandomness prf(size_t bytes) {
+         const auto& sym = m_mode.symmetric_primitives();
+         auto seed_span = m_seed.get();
+         sym.setup_PRF(m_prf_xof, seed_span, m_nonce++);
+         return m_prf_xof->output<KyberSamplingRandomness>(bytes);
+      }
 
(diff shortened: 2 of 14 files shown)


src/lib/pubkey/kyber/kyber_common/kyber_algos.cpp
```diff
@@ -384,10 +384,18 @@ KyberPolyMat sample_matrix(StrongSpan<const KyberSeedRho> seed, bool transposed,
...
    for(uint8_t i = 0; i < mode.k(); ++i) {
       for(uint8_t j = 0; j < mode.k(); ++j) {
          const auto pos = (transposed) ? std::tuple(i, j) : std::tuple(j, i);
-         sample_ntt_uniform(mat[i][j], mode.symmetric_primitives().XOF(seed, pos));
+         if(!xof) {
+            xof = sym.XOF(seed, pos);
+         } else {
+            sym.reseed_XOF(*xof, seed, pos);
+         }
+         sample_ntt_uniform(mat[i][j], *xof);
```
reneme, line 394: Given that this is used only here: Maybe move the state management into the symmetric primitives wrapper? Like so:

```C++
std::unique_ptr<Botan::XOF> xof;
for {
  for {
    sym.reseed_XOF(xof, seed, pos);
    sample_ntt_uniform(mat[i][j], *xof);
  }
}
```

... with `reseed_XOF` roughly following these lines:

```C++
void reseed_XOF(std::unique_ptr<Botan::XOF>& xof,
                StrongSpan<const KyberSeedRho> seed,
                std::tuple<uint8_t, uint8_t> matrix_position) const {
   if(!xof) {
      xof = create_XOF(seed, matrix_position);
   } else {
      init_XOF(xof, seed, matrix_position);
   }
}
```

src/lib/pubkey/kyber/kyber_common/kyber_algos.h
```diff
@@ -91,7 +91,16 @@ class PolynomialSampler {
...
    private:
-      KyberSamplingRandomness prf(size_t bytes) { return m_mode.symmetric_primitives().PRF(m_seed, m_nonce++, bytes); }
+      KyberSamplingRandomness prf(size_t bytes) {
+         const auto& sym = m_mode.symmetric_primitives();
+         auto seed_span = m_seed.get();
+         if(!m_prf_xof) {
+            m_prf_xof = sym.create_seeded_PRF(seed_span, m_nonce++);
+         } else {
+            sym.reseed_PRF(*m_prf_xof, seed_span, m_nonce++);
+         }
```
reneme: Same as above. This state handling could probably live within the symmetric primitives.

src/lib/pubkey/kyber/kyber_common/kyber_symmetric_primitives.h
```diff
@@ -57,41 +57,60 @@ class Kyber_Symmetric_Primitives /* NOLINT(*-special-member-functions) */ {
...
       template <concepts::contiguous_strong_type T1,
                 concepts::contiguous_strong_type T2,
                 ranges::contiguous_range... InputTs>
       std::pair<T1, T2> G_split(const InputTs&... inputs) const {
-         auto& g = get_G();
-         (g.update(inputs), ...);
-         auto s = g.final();
+         auto g = create_G();
+         (g->update(inputs), ...);
+         auto s = g->final();
```
reneme: Nit: could be `const`, I think.

src/lib/pubkey/kyber/kyber_common/kyber_algos.h
randombit: Done

src/lib/pubkey/kyber/kyber_common/kyber_algos.cpp
randombit: Done

PR #5372, opened 2026-02-21
Debug: try std::async instead of Thread_Pool for XMSS signing

randombit:


randombit:
Does not help re the crash

```
Command '.\botan-test --data-dir=.\src\tests\data --run-memory-intensive-tests --test-results-dir=junit_results --report-properties=ci_target:shared,os:windows' failed with error code 3221225477
```

PR #5376, opened 2026-02-21
Fix race in XMSS_Index_Registry

randombit:


randombit:
5 CI runs plus 2 nightly runs with no crashes, so :crossed_fingers: I think we're good

randombit:
Weirdly this race never showed up with helgrind, GCC or Clang's ThreadSanitizer, etc. I'm not sure why.

src/lib/pubkey/xmss/xmss_index_registry.cpp
```diff
@@ -11,61 +12,57 @@
...
    hash->update(prf);
-   secure_vector<uint8_t> result = hash->final();
-   uint64_t key_id = 0;
-   for(size_t i = 0; i < sizeof(key_id); i++) {
-      key_id = ((key_id << 8) | result[i]);
-   }
 
+   std::array<uint8_t, 32> key_id{};
+   hash->final(key_id);
    return key_id;
```
reneme, line 32: Side note: It would be nice if this were possible:

```C++
return hash->final_array<32>();
```

... we have similar API in the `RandomNumberGenerator` and also proposed for `BigInt`: https://github.com/randombit/botan/pull/5153

src/lib/pubkey/xmss/xmss_index_registry.cpp
```diff
@@ -11,61 +12,57 @@
...
-         return i;
-      }
-   }
+XMSS_Index_Registry::XMSS_Index_Registry() = default;
 
-   return std::numeric_limits<size_t>::max();
-}
+XMSS_Index_Registry::~XMSS_Index_Registry() = default;
+
+std::shared_ptr<Atomic<size_t>> XMSS_Index_Registry::get(uint32_t params,
```
reneme, line 47: The `uint32_t` is actually a membre of the `xmss_algorithm_t` enum. I think is's worth using that type explicitly here.

src/lib/pubkey/xmss/xmss_index_registry.cpp
```diff
@@ -11,61 +12,57 @@
...
 
-//static
-uint64_t XMSS_Index_Registry::make_key_id(const secure_vector<uint8_t>& private_seed,
-                                          const secure_vector<uint8_t>& prf) {
+namespace {
+
+std::array<uint8_t, 32> make_xmss_index_key_id(uint32_t params,
+                                               std::span<const uint8_t> private_seed,
+                                               std::span<const uint8_t> prf) {
    const std::string_view index_hash_function = "SHA-256";
```
reneme, line 23: Nit: For the moment XMSS explicitly depends on `sha2_32` in its _info.txt_ so this hard-coded hash algorithm is just fine. But if that were to ever change, we'd need to opportunistically select a supported hash algorithm from "SHA-256", "SHA-512", or "SHAKE-256". One of them has to be available, because otherwise XMSS doesn't make sense in the first place.

src/lib/pubkey/xmss/xmss_index_registry.cpp
```diff
@@ -11,61 +12,57 @@
...
+   const auto id = make_xmss_index_key_id(params, private_seed, prf);
 
-size_t XMSS_Index_Registry::add(uint64_t id, size_t last_unused) {
    const lock_guard_type<mutex_type> lock(m_mutex);
-   const size_t pos = get(id);
-   if(pos < m_key_ids.size()) {
-      if(last_unused > *(m_leaf_indices[pos])) {
-         m_leaf_indices[pos] = std::make_shared<Atomic<size_t>>(last_unused);
+
+   BOTAN_ASSERT_NOMSG(m_key_ids.size() == m_leaf_indices.size());
```
reneme: While we're at it: perhaps refactor the vectors `m_key_ids` and `m_leaf_indices` into a single vector `m_registry` that contains a

```C++
struct { 
  key_id_t id;
  std::shared_ptr<Atomic<size_t>> leaf_index;
};
```

<details>
<summary>Here's a patch for that</summary>

```patch
diff --git a/src/lib/pubkey/xmss/xmss_index_registry.cpp b/src/lib/pubkey/xmss/xmss_index_registry.cpp
index 619e35350..e09da431c 100644
--- a/src/lib/pubkey/xmss/xmss_index_registry.cpp
+++ b/src/lib/pubkey/xmss/xmss_index_registry.cpp
@@ -52,17 +52,14 @@ std::shared_ptr<Atomic<size_t>> XMSS_Index_Registry::get(uint32_t params,
 
    const lock_guard_type<mutex_type> lock(m_mutex);
 
-   BOTAN_ASSERT_NOMSG(m_key_ids.size() == m_leaf_indices.size());
-
-   for(size_t i = 0; i < m_key_ids.size(); i++) {
-      if(m_key_ids[i] == id) {
-         return m_leaf_indices[i];
+   for(size_t i = 0; i < m_registry.size(); i++) {
+      if(m_registry[i].key_id == id) {
+         return m_registry[i].leaf_index;
       }
    }
 
-   m_key_ids.push_back(id);
-   m_leaf_indices.push_back(std::make_shared<Atomic<size_t>>(0));
-   return m_leaf_indices.back();
+   m_registry.push_back({id, std::make_shared<Atomic<size_t>>(0)});
+   return m_registry.back().leaf_index;
 }
 
 }  // namespace Botan
diff --git a/src/lib/pubkey/xmss/xmss_index_registry.h b/src/lib/pubkey/xmss/xmss_index_registry.h
index 1e74915e5..e91391e47 100644
--- a/src/lib/pubkey/xmss/xmss_index_registry.h
+++ b/src/lib/pubkey/xmss/xmss_index_registry.h
@@ -22,6 +22,14 @@ namespace Botan {
  * independent copies of the same key.
  **/
 class XMSS_Index_Registry final {
+   private:
+      using key_id_t = std::array<uint8_t, 32>;
+
+      struct Entry {
+            key_id_t key_id;
+            std::shared_ptr<Atomic<size_t>> leaf_index;
+      };
+
    public:
       XMSS_Index_Registry(const XMSS_Index_Registry&) = delete;
       XMSS_Index_Registry(XMSS_Index_Registry&&) = delete;
@@ -56,10 +64,7 @@ class XMSS_Index_Registry final {
    private:
       XMSS_Index_Registry();
 
-      typedef std::array<uint8_t, 32> key_id_t;
-
-      std::vector<key_id_t> m_key_ids;
-      std::vector<std::shared_ptr<Atomic<size_t>>> m_leaf_indices;
+      std::vector<Entry> m_registry;
       mutex_type m_mutex;
 };
 </details>

File: src/lib/pubkey/xmss/xmss_privatekey.cpp [hunk 5]
```diff
@@ -66,7 +66,10 @@ class XMSS_PrivateKey_Internal {
             m_wots_derivation_method(wots_derivation_method),
             m_prf(rng.random_vec(xmss_params.element_size())),
             m_private_seed(rng.random_vec(xmss_params.element_size())),
-            m_index_reg(XMSS_Index_Registry::get_instance()) {}
+            m_index_reg(XMSS_Index_Registry::get_instance()) {
+         // Eagerly register so the entry exists before any concurrent signing
+         recover_global_leaf_index();
```
reneme: From a concurrency correctness perspective this isn't really necessary. Concurrently calling `m_index_reg.get()` is perfectly legal with the mutex within it. Or am I missing something?

src/lib/pubkey/xmss/xmss_index_registry.cpp
reneme: ... ah, nevermind. In #5382 this won't work anyway.

src/lib/pubkey/xmss/xmss_index_registry.cpp
randombit: Alternately we could require that the caller pass a hash reference. This requires that the caller itself is careful to never hash the same key with two different hash functions since otherwise it would result in two distinct key IDs for the same key, breaking the invariants. But in the XMSS/LMS case there is typically already a hash function at hand so that is maybe ok.

IIRC the `sha_32` dependency in XMSS is precisely because of this usage in the index.

src/lib/pubkey/xmss/xmss_privatekey.cpp
randombit: It's definitely not necessary for correctness and maybe confusingly implies that it is. I wanted it for debugging initially. Removed.

src/lib/pubkey/xmss/xmss_index_registry.cpp
randombit: Done

src/lib/pubkey/xmss/xmss_index_registry.cpp
reneme: With #5382 this dependency can move into the generic registry module making it more obvious that the hash is only needed for the registry. And then we might as well just leave it in for future work. I'm not really a fan of giving the responsibility of selecting the hash function to the XMSS/LMS modules for the brittleness it introduces.

PR #5380, opened 2026-02-22
Fix ML-DSA/Dilithium concurrency issues

randombit:
They used shared hashes/XOFs which cause problem if multiple threads try to use the same key object concurrently without external locking.


src/lib/pubkey/dilithium/dilithium_round3/dilithium_aes/dilithium_aes.cpp
```diff
@@ -41,13 +42,10 @@ class AES_XOF final : public DilithiumXOF {
          const std::array<uint8_t, 12> iv{get_byte<1>(nonce), get_byte<0>(nonce), 0};
          const auto key = seed.first(32);
 
-         xof.clear();
-         xof.start(iv, key);
+         auto xof = std::make_unique<AES_256_CTR_XOF>();
```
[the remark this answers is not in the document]
randombit: How would this reduce allocations robot

src/lib/pubkey/dilithium/dilithium_common/dilithium_shake/dilithium_shake_xof.cpp
```diff
@@ -8,9 +8,17 @@
...
 
+//static
+std::unique_ptr<Botan::XOF> DilithiumShakeXOF::createXOF(std::string_view name,
+                                                         std::span<const uint8_t> seed,
+                                                         uint16_t nonce) {
+   auto xof = Botan::XOF::create_or_throw(name);
+   const uint8_t nonce8[2] = {static_cast<uint8_t>(nonce), static_cast<uint8_t>(nonce >> 8)};
+   xof->update(seed);
+   xof->update(nonce8);
+   return xof;
```
reneme, line 22: Now that it's a *.cpp file: maybe re-introduce the `update(store_be(nonce))` composition by including `loadstor.h`? 😏

src/lib/pubkey/dilithium/dilithium_common/dilithium_algos.cpp
```diff
@@ -733,8 +733,8 @@ DilithiumPolyVec expand_mask(StrongSpan<const DilithiumSeedRhoPrime> rhoprime,
                              const DilithiumConstants& mode) {
    DilithiumPolyVec s(mode.l());
    for(auto& p : s) {
-      auto& xof = mode.symmetric_primitives().H(rhoprime, nonce++);
-      poly_unpack_gamma1(p, xof, mode);
+      auto xof = mode.symmetric_primitives().H(rhoprime, nonce++);
+      poly_unpack_gamma1(p, *xof, mode);
    }
    return s;
```
reneme, line 739: There might eventually be an opportunity for optimization here if we pull out the handling of the nonce from `.H()`. Roughly like so:

```C++
DilithiumPolyVec s(mode.l());
auto xof = mode.symmetric_primitives().H(rhoprime);
for(auto& p : s) {
   auto round_xof = xof.copy_state();
   round_xof->update(store_be(nonce++));
   poly_unpack_gamma1(p, *round_xof, mode);
}
return s;
```

... not sure this is really worth it, though. Also: currently this doesn't work because of the AES-variant that incorporates the nonce as IV in `XOF::start()` not with an independent `update()` after the seed.

randombit: Could also do a SIMD Keccak to compute all of the permutations in parallel.

Feel like a lot of these optimizations will be easier when 90s modes are dropped though

PR #5382, opened 2026-02-22
Generalize XMSS_Index_Registry and use it for LMS as well

randombit:


src/lib/pubkey/stateful_key_index/info.txt
```diff
@@ -0,0 +1,12 @@
+<module_info>
+name -> "Stateful Key Index"
+brief -> "Tracks updates for stateful signing algorithms"
+</module_info>
```
reneme, line 5: Should be an internal module (`type -> "Internal"`).

src/lib/pubkey/stateful_key_index/stateful_key_index_registry.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+std::array<uint8_t, 32> make_key_id(std::string_view algo_name,
+                                    uint32_t algo_params,
+                                    std::span<const uint8_t> key_material_1,
+                                    std::span<const uint8_t> key_material_2) {
+   auto hash = HashFunction::create_or_throw("SHA-256");
+   hash->update("Botan Stateful_Key_Index_Registry KeyID");
+   hash->update(algo_name);
+   hash->update_be(algo_params);
+   hash->update_be(static_cast<uint64_t>(key_material_1.size()));
```
reneme, line 35: Why the static cast? This doesn't need to be consistent across builds, so shouldn't `size_t` be just fine?

src/lib/pubkey/stateful_key_index/stateful_key_index_registry.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+namespace {
+
+std::array<uint8_t, 32> make_key_id(std::string_view algo_name,
+                                    uint32_t algo_params,
+                                    std::span<const uint8_t> key_material_1,
+                                    std::span<const uint8_t> key_material_2) {
+   auto hash = HashFunction::create_or_throw("SHA-256");
+   hash->update("Botan Stateful_Key_Index_Registry KeyID");
+   hash->update(algo_name);
```
reneme, line 33: For good measure, also incorporate the length here? For instance "XMSS" and "XMSS^MT" have the same prefix. Not sure this could ever cause issues, but alas.

src/lib/pubkey/xmss/info.txt
```diff
@@ -29,9 +27,6 @@ asn1
 rng
 hash
 sha2_32
```
reneme: I guess this can go now, it is required by `stateful_key_index` and fits better there.

src/lib/pubkey/stateful_key_index/stateful_key_index_registry.cpp
randombit: It would have been and that's what I did at first, but on macOS `size_t` is neither `uint32_t` nor `uint64_t` so the overload is considered ambiguous.

src/lib/pubkey/xmss/info.txt
randombit: Removed

randombit:
> One general suggestion, though: I think we could encapsulate the entire std::atomic handling within Stateful_Key_Index_Registry and don't let the XMSS and LMS implementations fiddle with fetch_add and friends at all.

Yeah that does make sense. Certainly there are some shared usage patterns. I'll look at this.

> Perhaps, we could even just base everything off of the m_mutex and ditch the dependency on atomics entirely. Its probably a little less efficient, but we're anyway about to sign with a hash-based signature when fetching a new free index...

Probably so, at least as long as your prior suggestion works out.

src/lib/pubkey/stateful_key_index/stateful_key_index_registry.cpp
reneme: Yeah, indeed. `size_t` expands to something like `_darwin_size_t` which is defined as the macro `_SIZE_TYPE_` which is incompatible on the type system level with the typedefs of `uint32_t` and `uint64_t`. ... w.t.f.

src/lib/pubkey/stateful_key_index/stateful_key_index_registry.cpp
randombit: :+1: Not sure if it could either but easily avoided

randombit:
> I think we could encapsulate the entire std::atomic handling within Stateful_Key_Index_Registry and don't let the XMSS and LMS implementations fiddle with fetch_add and friends at all.

Done - this worked out pretty well I think.

> Its probably a little less efficient, but we're anyway about to sign with a hash-based signature when fetching a new free index...

If these schemes are ever optimized enough that the lock is noticeable, that's a good day

Speaking of performance btw, the other day I asked Claude to try AVX2 8-wide parallel SHA-256 evaluation and plug it into XMSS. The code is kind of a mess and I'm not sure if I want to take any of it, at the very least it would need a lot of reworking (plus probably has a *lot* of conflicts now), but it does seem to work and is about 3x faster signing.

src/lib/pubkey/stateful_key_index/stateful_key_index_registry.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+Stateful_Key_Index_Registry::Stateful_Key_Index_Registry() = default;
+
+Stateful_Key_Index_Registry::~Stateful_Key_Index_Registry() = default;
+
+Stateful_Key_Index_Registry::KeyId::KeyId(std::string_view algo_name,
+                                          uint32_t algo_params,
+                                          std::span<const uint8_t> key_material_1,
+                                          std::span<const uint8_t> key_material_2) :
+      m_val() {
```
reneme, line 28: Was the compiler asking to default-init the array? Otherwise that probably won't be necessary.

src/lib/pubkey/stateful_key_index/stateful_key_index_registry.cpp
```diff
@@ -0,0 +1,79 @@
...
+   hash->update("Botan Stateful_Key_Index_Registry KeyID");
+   hash->update_be(static_cast<uint64_t>(algo_name.size()));
+   hash->update(algo_name);
+   hash->update_be(algo_params);
+   hash->update_be(static_cast<uint64_t>(key_material_1.size()));
+   hash->update(key_material_1);
+   hash->update_be(static_cast<uint64_t>(key_material_2.size()));
+   hash->update(key_material_2);
+
+   hash->final(m_val);
```
reneme, line 42: Perhaps:


```suggestion
   BOTAN_DEBUG_ASSERT(m_val.size() == hash->output_length());
   hash->final(m_val);
```

src/lib/pubkey/stateful_key_index/stateful_key_index_registry.cpp
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// Lock must be held while this function is called
+Stateful_Key_Index_Registry::RegistryMap::iterator Stateful_Key_Index_Registry::lookup(const KeyId& key_id) {
+   auto [i, _inserted] = m_registry.insert(std::make_pair(key_id, 0));
```
reneme:
```suggestion
   auto [i, _inserted] = m_registry.emplace(key_id, 0);
```

src/lib/pubkey/hss_lms/hss.cpp
```diff
@@ -193,17 +213,20 @@ secure_vector<uint8_t> HSS_LMS_PrivateKeyInternal::to_bytes() const {
...
 HSS_Sig_Idx HSS_LMS_PrivateKeyInternal::reserve_next_idx() {
-   HSS_Sig_Idx next_idx = m_current_idx;
-   if(next_idx >= m_hss_params.max_sig_count()) {
+   const uint64_t idx = Stateful_Key_Index_Registry::global().reserve_next_index(m_keyid);
+   if(idx >= m_hss_params.max_sig_count().get()) {
       throw Decoding_Error("HSS private key is exhausted");
    }
-   set_idx(m_current_idx + 1);
-   return next_idx;
+   return HSS_Sig_Idx(idx);
```
reneme: If you already wrap the output of `reserve_next_index` you don't need to unwrap the strong type (I'm assuming that's what the `.get()` is for):


```suggestion
   const auto idx = HSS_Sig_Idx(Stateful_Key_Index_Registry::global().reserve_next_index(m_keyid));
   if(idx >= m_hss_params.max_sig_count()) {
throw Decoding_Error("HSS private key is exhausted");
}
   return idx;
```

src/tests/data/pubkey/hss_lms_sig.vec
```diff
@@ -1,12 +1,9 @@
...
+#
+# NOTE: Some of test vectors use the same seed and identifier. They must be
+# ordered by index, as Stateful_Key_Index_Registry enforces that the index for
+# any key only moves forward.
```
reneme, line 6: Oh, nice.

src/lib/pubkey/stateful_key_index/stateful_key_index_registry.cpp
randombit: I think I was getting a warning which is why that's there, but I don't recall now, I'll check

randombit: Ah that's what it was

`clang-tidy` `cppcoreguidelines-pro-type-member-init: Constructor does not initialize these fields: m_val`

src/lib/pubkey/hss_lms/hss.cpp
randombit: Nice, done

src/lib/pubkey/stateful_key_index/stateful_key_index_registry.cpp
randombit: Good idea. Did it as a runtime assert since why not

randombit:
Nightly https://github.com/randombit/botan/actions/runs/22306982887

TJ-91:
> Speaking of performance btw, the other day I asked Claude to try AVX2 8-wide parallel SHA-256 evaluation and plug it into XMSS. The code is kind of a mess and I'm not sure if I want to take any of it, at the very least it would need a lot of reworking (plus probably has a _lot_ of conflicts now), but it does seem to work and is about 3x faster signing.

Just to add to that: For XMSS, almost all computational effort is spent in hash computations. Therefore, any speedup for computing hashes will be directly noticable.

Aside from that, you can get **much** faster signing speed by avoiding the recomputation of trees alltogether for signatures. There are several algorithms to implement such a time-memory tradeoff, the most well-known being BDS, which I think is also implemented in the reference implementation (the "fast" variant).

Bottom line:
* Key Generation will benefit from faster hash implementations
* Signaure Generation will benefit from faster hash implementations but even more from strategically storing internal nodes to prevent costly leaf recomputations.
* Verification Operation will benefit from faster hash implementations but is already reasonably fast since you only need to "complete the chains" of one leaf and then "hash up the the tree with the help of the auth path"

Edit: To clarify, the additional state (storing some internal tree nodes) is stored in the private key

TJ-91:
I just remembered that the SPHINCS+/SLH-DSA reference code also includes AVX2 code: https://github.com/sphincs/sphincsplus/blob/master/sha2-avx2/

For example, here https://github.com/sphincs/sphincsplus/blob/master/sha2-avx2/utilsx8.c treehash with sha2-256 using 8 inputs at the same time is implemented.

I suppose the most speedup comes from https://github.com/sphincs/sphincsplus/blob/master/sha2-avx2/wots.c where the leaf computation is sped up.

Implementing this stuff means changing the algorithms fundamentally to enable the data-level-parallelization. 

There might be some low-effort approaches. For example, simply speeding up the chain method by parallelizing the computation of diferent chains should be fairly simple to implement and on its own mean a significant speedup. I don't have numbers on what methods are the most time consuming, but I would rank that pretty high and assume it's by far the most time consuming method. For example signing/keygen with len=67 and winternitz param=16, each leaf will require 67*15 (around 1000) hash computations to compute chains. And that is for each leaf.

So if you ask me, improving the chain function would be the logical first step for the following reasons:
* most impact (presumably)
* very localized code changes
* presumably easy to implement since different chains are completely independent of each other

Note that the code will need to have variants for 192-, 256, and 512-bit hash output lengths.

IMO using data-level-parallization for other functionality may be worthwhile but it also makes the code a lot more complex since computing the internal tree nodes is not as independent from the other nodes computing the chains.

PR #5384, opened 2026-02-24
For FFDH exponents follow the guidance of NIST SP 800-56A and 800-56B

randombit:
For many common group sizes this results in using a slightly smaller exponent and thus faster computation

| Group Size | Old Exponent Size | NIST SP 800-56A Exponent |
| ---------- | ----------------- | ------------------------ |
|       2048 |              256  |                      224 |
|       3072 |              384  |                      256 |
|       4096 |              384  |                      304 |
|       6144 |              512  |                      352 |
|       8192 |              512  |                      400 |

cc @reneme since possibly BSI has opinions on this, though I would hope it more or less matches up with NIST


reneme:
> cc @reneme since possibly BSI has opinions on this, though I would hope it more or less matches up with NIST

I take the liberty to pass this question on to @falko-strenzke :)

falko-strenzke:
@randombit 
Comparing it with the requirements in [BSI TR 02102](https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Publikationen/TechnischeRichtlinien/TR02102/BSI-TR-02102.pdf?__blob=publicationFile#page=39), the parameters look OK, since BSI requires `prime ≥ 3000` and `exponent ≥ 250`. This is fulfilled for by the table. So I think it should be fine.

falko-strenzke (approved):

PR #5396, opened 2026-03-01
FIX: botan_509_is_revoked return 1 on revoked, 0 otherwise

KaganCanSit:
Similar to #5395. Read the notes there.

### DIFF
--- a/doc/api_ref/ffi.rst
+++ b/doc/api_ref/ffi.rst
@@ -2084,7 +2084,7 @@ X.509 Certificate Revocation Lists
 .. cpp:function:: int botan_x509_is_revoked(botan_x509_crl_t crl, botan_x509_cert_t cert)
 
    Check whether a given ``crl`` contains a given ``cert``.
-   Return ``0`` when the certificate is revoked, ``-1`` otherwise.
+   Return ``1`` when the certificate is revoked, ``0`` otherwise.
 
 .. cpp:function:: int botan_x509_crl_entries(botan_x509_crl_t crl, \
                                              size_t index, \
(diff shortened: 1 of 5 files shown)


KaganCanSit:
[ci / Windows (shared, x86_64, windows-2022, msvc) (pull_request)](https://github.com/randombit/botan/actions/runs/22537060926/job/65286397675?pr=5396) failed;

> PermissionError: [WinError 10013] An attempt was made to access a socket in a way forbidden by its access permissions
    ERROR: Got output on stderr Socket read failed error code 0 which did not match expected value ERROR: Public key does not match!

I will amend the last commit to trigger the tests again (`--amend --no-edit`).

reneme:
Thanks for wanting to help, but please consider that we cannot just flip the return values of already existing functions. This would break existing applications that use them from previous versions of the library.

I did that for a few functions recently, but the difference is that those functions were introduced so recently that they were not part of any release yet.

Please look into the discussion here: https://github.com/randombit/botan/pull/5236#issuecomment-3747933998

KaganCanSit:
> Thanks for wanting to help, but please consider that we cannot just flip the return values of already existing functions. This would break existing applications that use them from previous versions of the library.
> 
> I did that for a few functions recently, but the difference is that those functions were introduced so recently that they were not part of any release yet.
> 
> Please look into the discussion here: [#5236 (comment)](https://github.com/randombit/botan/pull/5236#issuecomment-3747933998)

Ohh, I actually anticipated this and added a note. I missed the beginning of the first paragraph while reading the early discussion. (`... I think the most important thing is that we all agree on one option, codify it in the "rules of engagement" and then harmonize it for Botan4.`) I'm closing these so they don't create noise.

Sorry.

PR #5399, opened 2026-03-01
Use RAII for addrinfo lifetime management

KaganCanSit:
Hello,
It has been over a year since my PR #4660, and I honestly don't remember many of the details. While revisiting that branch, I identified changes that could independently contribute to the project and decided to submit them as separate PRs.

This PR applies RAII-based lifetime management to addrinfo resources returned by `getaddrinfo()`. In the current codebase, `::freeaddrinfo()` is called manually, which can leak if an exception is thrown before the cleanup call is reached.

The changes are based on the approach discussed with @reneme in [this review comment](https://github.com/randombit/botan/pull/4660#discussion_r1977732281) (Since it is marked as resolved, you cannot directly move to the comment.): a `std::unique_ptr` with a stateless lambda deleter combined with `Botan::out_ptr()` from `stl_util.h`.

Quotes:
@KaganCanSit
> The freeaddrinfo() function is defined to free one or more addrinfo structures returned by getaddrinfo() and the additional memory associated with them. It usually frees the resulting linked list (addrinfo chain) when getaddrinfo() succeeds. The POSIX standard assumes that the freeaddrinfo() call will only apply to a valid addrinfo list returned by getaddrinfo(). The text of the standard does not explicitly define the case where an invalid pointer (e.g. NULL) is passed to freeaddrinfo() – this case is left undefined.
>
> So I'm a bit hesitant here. I thought it might be UB if the addrinfo retrieval fails and is freed. However, I liked the approach you suggested in other comments. I learned, thanks. I'll reconsider in the last case.

@reneme 
> Mhm, you might have a point here. It certainly doesn't hurt to check the pointer for nullptr before calling freeaddrinfo(). When it comes to C-APIs I've seen both: some are fine with receiving a nullptr others aren't.
>
> Instead of writing this out multiple times, you could simply define a typedef in the new socket_platform.h and use that in tls_client.cpp and the socket*.cpps.
>
>Like so:
>
> using unique_addrinfo_ptr = std::unique_ptr<addrinfo, decltype([](addrinfo* p) {
>                                               if(p != nullptr) {
>                                                  ::freeaddrinfo(p);
>                                               }
>                                            })>;
>// in the *.cpp files
>Botan::OS::Socket_Platform::unique_addrinfo_ptr res = nullptr;

The using declarations are defined as private members in each class since there is currently no shared internal header across socket files. (`socket_utils.h` exists but is external and not included, I think there is a reason for this.)

I edited the `bogo_shim.cpp` file to ensure full compatibility with the others and to allow direct intervention when `Botan::out_ptr` is removed. However, if I am mistaken, please let me know.

Compile and Test
```
./configure.py --without-documentation --compiler-cache=ccache --build-targets=static,cli,tests && 
make clean && make -j$(nproc) && 
python3 src/scripts/test_cli.py ./botan cli_tls
```

I am happy to make further adjustments based on your feedback.
Best regards.

### DIFF
--- a/src/bogo_shim/bogo_shim.cpp
+++ b/src/bogo_shim/bogo_shim.cpp
@@ -381,24 +381,23 @@ class Shim_Socket final {
 
       static std::string get_last_socket_error() { return ::strerror(errno); }
 
-      using unique_addrinfo_t = std::unique_ptr<addrinfo, decltype(&::freeaddrinfo)>;
+      using unique_addr_info_ptr = std::unique_ptr<addrinfo, decltype([](addrinfo* p) {
+                                                      if(p != nullptr) {
+                                                         ::freeaddrinfo(p);
+                                                      }
+                                                   })>;
 
    public:
       Shim_Socket(const std::string& hostname, int port, const bool ipv6) : m_socket(-1) {
          addrinfo hints{};
-         std::memset(&hints, 0, sizeof(hints));
          hints.ai_family = AF_UNSPEC;
          hints.ai_socktype = SOCK_STREAM;
          hints.ai_flags = AI_NUMERICSERV;
 
          const std::string service = std::to_string(port);
 
-         // TODO: C++23 will introduce std::out_ptr() that should replace the
-         //       temporary variable for the call to ::getaddrinfo() and
-         //       std::unique_ptr<>::reset().
-         unique_addrinfo_t::pointer res_tmp = nullptr;
-         const int rc = ::getaddrinfo(hostname.c_str(), service.c_str(), &hints, &res_tmp);
-         const unique_addrinfo_t res(res_tmp, &::freeaddrinfo);
+         unique_addr_info_ptr res = nullptr;
+         const int rc = ::getaddrinfo(hostname.c_str(), service.c_str(), &hints, Botan::out_ptr(res));
 
          shim_log("Connecting " + hostname + ":" + service);
 
--- a/src/cli/tls_client.cpp
+++ b/src/cli/tls_client.cpp
@@ -10,6 +10,7 @@
 
 #include "cli.h"
 
+#include <botan/internal/stl_util.h>
 #include <botan/internal/target_info.h>
 
 #if defined(BOTAN_HAS_TLS) && defined(BOTAN_TARGET_OS_HAS_FILESYSTEM) && defined(BOTAN_TARGET_OS_HAS_SOCKETS)
@@ -387,19 +388,19 @@ class TLS_Client final : public Command {
    private:
       static socket_type connect_to_host(const std::string& host, uint16_t port, bool tcp) {
          addrinfo hints{};
-         std::memset(&hints, 0, sizeof(hints));
          hints.ai_family = AF_UNSPEC;
          hints.ai_socktype = tcp ? SOCK_STREAM : SOCK_DGRAM;
-         addrinfo* res = nullptr;
 
-         if(::getaddrinfo(host.c_str(), std::to_string(port).c_str(), &hints, &res) != 0) {
+         unique_addr_info_ptr res = nullptr;
+
+         if(::getaddrinfo(host.c_str(), std::to_string(port).c_str(), &hints, Botan::out_ptr(res)) != 0) {
             throw CLI_Error("getaddrinfo failed for " + host);
          }
 
          socket_type fd = 0;
          bool success = false;
 
-         for(const addrinfo* rp = res; rp != nullptr; rp = rp->ai_next) {
+         for(const addrinfo* rp = res.get(); rp != nullptr; rp = rp->ai_next) {
             fd = ::socket(rp->ai_family, rp->ai_socktype, rp->ai_protocol);
 
             if(fd == invalid_socket()) {
@@ -414,8 +415,6 @@ class TLS_Client final : public Command {
             success = true;
             break;
          }
-
-         ::freeaddrinfo(res);
 
          if(!success) {
             // no address succeeded
@@ -434,6 +433,12 @@ class TLS_Client final : public Command {
       }
 
       socket_type m_sockfd = invalid_socket();
+
+      using unique_addr_info_ptr = std::unique_ptr<addrinfo, decltype([](addrinfo* p) {
+                                                      if(p != nullptr) {
+                                                         ::freeaddrinfo(p);
+                                                      }
+                                                   })>;
 };
 
 namespace {
--- a/src/lib/utils/socket/socket.cpp
+++ b/src/lib/utils/socket/socket.cpp
@@ -10,6 +10,7 @@
 #include <botan/exceptn.h>
 #include <botan/mem_ops.h>
 #include <botan/internal/fmt.h>
+#include <botan/internal/stl_util.h>
 #include <botan/internal/target_info.h>
 #include <chrono>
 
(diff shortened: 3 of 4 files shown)


KaganCanSit:
If this development is deemed appropriate, the following simplification can through a separate branch.

```diff
diff --git a/src/cli/tls_client.cpp b/src/cli/tls_client.cpp
index 56cc4a76a..f18f106ef 100644
--- a/src/cli/tls_client.cpp
+++ b/src/cli/tls_client.cpp
@@ -397,11 +397,8 @@ class TLS_Client final : public Command {
             throw CLI_Error("getaddrinfo failed for " + host);
          }
 
-         socket_type fd = 0;
-         bool success = false;
-
          for(const addrinfo* rp = res.get(); rp != nullptr; rp = rp->ai_next) {
-            fd = ::socket(rp->ai_family, rp->ai_socktype, rp->ai_protocol);
+            socket_type fd = ::socket(rp->ai_family, rp->ai_socktype, rp->ai_protocol);
 
             if(fd == invalid_socket()) {
                continue;
@@ -412,16 +409,10 @@ class TLS_Client final : public Command {
                continue;
             }
 
-            success = true;
-            break;
-         }
-
-         if(!success) {
-            // no address succeeded
-            throw CLI_Error("Connecting to host failed");
+            return fd;
          }
-
-         return fd;
+         // no address succeeded
+         throw CLI_Error("Connecting to host failed");
       }
```

src/bogo_shim/bogo_shim.cpp
```diff
@@ -381,24 +381,23 @@ class Shim_Socket final {
...
+                                                      }
+                                                   })>;
 
    public:
       Shim_Socket(const std::string& hostname, int port, const bool ipv6) : m_socket(-1) {
          addrinfo hints{};
-         std::memset(&hints, 0, sizeof(hints));
          hints.ai_family = AF_UNSPEC;
          hints.ai_socktype = SOCK_STREAM;
          hints.ai_flags = AI_NUMERICSERV;
```
reneme: Didn't check if the compiler eats it, but in C++20 you might even:

```C++
auto hints = addrinfo{
         .ai_family = AF_UNSPEC,
         .ai_socktype = SOCK_STREAM,
         .ai_flags = AI_NUMERICSERV,
};
```

KaganCanSit: You're welcome. I'll try it with GCC and Clang in two hours and check it. If possible, I'll change and try it in CI.

Thanks for review. 👋🏻

---
Update 1: Compiler Explorer and local compile looks good. I want to see CI, if all test green. Later mark this conversation solved.

[GCC11.2](https://godbolt.org/z/G3GnxqzK5)
[Clang 14](https://godbolt.org/z/sfnYY71jK)

Update 2: `-Wmissing-field-initializers`'s flag definitely didn't like this change. I'm taking it back.

```bash
2026-03-02T17:03:46.0812608Z src/lib/utils/socket/socket.cpp:213:10: error: missing initializer for member ‘addrinfo::ai_flags’ [-Werror=missing-field-initializers]
2026-03-02T17:03:46.0817853Z ccache g++ -fstack-protector -m32 -pthread -std=c++20 -D_REENTRANT  -O3 -Wall -Wextra -Wpedantic -Wstrict-aliasing -Wcast-align -Wmissing-declarations -Wpointer-arith -Wcast-qual -Wzero-as-null-pointer-constant -Wnon-virtual-dtor -Wold-style-cast -Wsuggest-override -Wshadow -Wextra-semi -Werror -Wno-error=strict-overflow -Wno-error=zero-as-null-pointer-constant -Wno-error=non-virtual-dtor -Wno-error=maybe-uninitialized -Wno-error=stringop-overread -Wno-error=stringop-overflow -Wno-error=free-nonheap-object -Wno-error=restrict -DBOTAN_IS_BEING_BUILT -I build/include/public -I build/include/internal  -isystem build/include/external -c src/tests/test_ecies.cpp -o build/obj/test/test_ecies.o
2026-03-02T17:03:46.0823569Z   213 |          };
2026-03-02T17:03:46.0824164Z       |          ^
2026-03-02T17:03:46.0825643Z src/lib/utils/socket/socket.cpp:213:10: error: missing initializer for member ‘addrinfo::ai_protocol’ [-Werror=missing-field-initializers]
2026-03-02T17:03:46.0828236Z src/lib/utils/socket/socket.cpp:213:10: error: missing initializer for member ‘addrinfo::ai_addrlen’ [-Werror=missing-field-initializers]
2026-03-02T17:03:46.0830474Z src/lib/utils/socket/socket.cpp:213:10: error: missing initializer for member ‘addrinfo::ai_addr’ [-Werror=missing-field-initializers]
2026-03-02T17:03:46.0832979Z src/lib/utils/socket/socket.cpp:213:10: error: missing initializer for member ‘addrinfo::ai_canonname’ [-Werror=missing-field-initializers]
2026-03-02T17:03:46.0835061Z src/lib/utils/socket/socket.cpp:213:10: error: missing initializer for member ‘addrinfo::ai_next’ [-Werror=missing-field-initializers]
```

KaganCanSit:
> Patches that complicate the code in order to support obsolete operating systems will likely be rejected.

I completely agree with this principle. In business life, I see that maintaining dependency on old platforms significantly increases maintenance costs in the long run.

Actually, my assessment here is largely in line with yours. Since `addrinfo` structures are stored as a **linked list**, a structure with a length of zero is needed to determine the end of the list (`NULL` / `nullptr`). A typical `freeaddrinfo` implementation also frees the chain as follows:

```cpp
while (ai) {
   next = ai->ai_next;
   free(ai);
   ai = next;
}
```

During development and look again today, what particularly caught my attention was that some examples did not allow early termination with a call to `freeaddrinfo(NULL)`, and the standard did not explicitly state this behavior as it does for `free(NULL)`.

https://linux.die.net/man/3/freeaddrinfo

```cpp
struct addrinfo hints;

	// ...

   s = getaddrinfo(NULL, argv[1], &hints, &result);
    if (s != 0) {
        fprintf(stderr, "getaddrinfo: %s\n", gai_strerror(s));
        exit(EXIT_FAILURE);
    }

   /* getaddrinfo() returns a list of address structures.
       Try each address until we successfully bind(2).
       If socket(2) (or bind(2)) fails, we (close the socket
       and) try the next address. */

   for (rp = result; rp != NULL; rp = rp->ai_next) {
        sfd = socket(rp->ai_family, rp->ai_socktype,
                rp->ai_protocol);
        if (sfd == -1)
            continue;

       if (bind(sfd, rp->ai_addr, rp->ai_addrlen) == 0)
            break;                  /* Success */

       close(sfd);
    }

   if (rp == NULL) {               /* No address succeeded */
        fprintf(stderr, "Could not bind\n");
        exit(EXIT_FAILURE);
    }

   freeaddrinfo(result);           /* No longer needed */
```

I also saw this issue specifically addressed in the following projects (2018, 2019, 2023):
* https://patchwork.openvpn.net/project/openvpn2/patch/20230209153658.24001-1-ynezz@true.cz/#5791
* https://github.com/OpenSMTPD/OpenSMTPD/issues/929
* https://git.openembedded.org/openembedded-core-contrib/tree/meta/recipes-connectivity/nfs-utils/nfs-utils/0001-Do-not-pass-null-pointer-to-freeaddrinfo.patch?h=paule/distrodata-selftest-fix

As you mentioned, it has a very small cost (a single if). However, I am aware that this could be an overly protective approach.

I am considering addressing the above comment as a separate PR. If you prefer, we can remove this change or leave it as is.

Thank you for your time, review, and sharing your knowledge.

Best regards.

PR #5404, opened 2026-03-03
Add EC scalars and points to FFI

arckoor:
Exposes most of the EC interfaces to allow implementing low level protocols that want to work with points themselves.

### DIFF
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -1441,6 +1441,117 @@ BOTAN_FFI_EXPORT(3, 8) int botan_ec_group_get_order(botan_mp_t* order, botan_ec_
 BOTAN_FFI_EXPORT(3, 8) int botan_ec_group_equal(botan_ec_group_t curve1, botan_ec_group_t curve2);
 
 /*
+* EC Points and Scalars
+*/
+typedef struct botan_ec_scalar_struct* botan_ec_scalar_t;
+typedef struct botan_ec_point_struct* botan_ec_point_t;
+
+BOTAN_FFI_EXPORT(3, 12) int botan_ec_scalar_destroy(botan_ec_scalar_t ec_scalar);
+
+/**
+* Create a new random scalar value
+*/
+BOTAN_FFI_EXPORT(3, 12)
+int botan_ec_scalar_random(botan_ec_scalar_t* ec_scalar, botan_ec_group_t ec_group, botan_rng_t rng);
+
+/**
+* Convert from an MPI to a scalar
+* @returns a negative number if the provided MPI is negative or too large, 0 on success
+*/
+BOTAN_FFI_EXPORT(3, 12)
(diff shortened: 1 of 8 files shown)


src/lib/ffi/ffi.h
```diff
@@ -1402,13 +1404,223 @@ BOTAN_FFI_EXPORT(3, 8) int botan_ec_group_get_g_y(botan_mp_t* g_y, botan_ec_grou
...
+*/
+BOTAN_FFI_EXPORT(3, 12)
+int botan_ec_group_get_generator(botan_ec_group_t ec_group, botan_ec_point_t* generator);
+
+/**
+* Get the identity element of the group
+*/
+BOTAN_FFI_EXPORT(3, 12)
+int botan_ec_group_get_identity(botan_ec_group_t ec_group, botan_ec_point_t* identity);
+
```
arckoor: Currently both group -> identity / generator and point(group) -> identity / generator work, it might be better to only offer one. Might also be a nice to have for api end users, I'm open to both

arckoor (commented):
Still needs tests, rst doc and the python binding. If after a quick glance you think there is something that just shouldn't be exposed, let me know though

randombit:
In general fine to export FFI for this, I'm quite strongly convinced that dedicated types for these objects (eg `EC_Scalar` rather than `BigInt` for elements of the integers modulo the group order) is the right choice, and I think it's all pretty stable now. And considering Botan's ECC is these days faster than nearly every other library, might be of interest to a variety of users.

I'd question a bit adding some of the operations, especially viewing just `x`/`y` bytes of a point, and the full set of scalar arithmetic. I guess it depends on the usage one envisions.... they don't seem like they add to much extra work I guess, in terms of maintenance, but feels kind of noisy tbh.

src/lib/ffi/ffi.h
arckoor: Those convenience functions can be provided by the wrapping bindings, not needed in the ffi

arckoor:
@randombit I ended up removing the scalar arithmetic for now, though I've stashed what I had already written. If I do end up needing them I'll do a separate PR. As for all the view functions, I know I need at least one of them, and not including the others felt wrong too.
I'm not sure exactly what you're envisioning for tests here since testing that the arithmetic itself works is rather cumbersome via the FFI (and I hope that is already covered very well in CPP tests), so I settled for some easy to test invariants.

randombit:
If you don't need `EC_Scalar::one` I'd drop that as well. That's really only need for key setup in SM2 and can be easily performed using the from_mp anyway. What you have for tests seems fine.

randombit:
Relevant CI failures

arckoor:
Whoops, thanks. I thought I got all the uses of `_ec_scalar_one` but evidently I missed one :sweat_smile:

randombit (approved):

PR #5416, opened 2026-03-06
Include service field in TLS session cache lookups

oviano:
## Summary

`Session_Manager_SQL::find_some()` queries sessions by `hostname` and `hostport` only, ignoring the `service` field of `Server_Information`. This means sessions for different services (e.g. TLS vs DTLS) on the same host:port are returned interchangeably, which can cause a DTLS session ticket to be resumed on a TLS connection.

Note that `Server_Information::operator==` and `operator<` both compare the `service` field, so the in-memory comparison is correct — only the SQL-backed session manager has this gap.

## Changes

- Add `service` column to the `tls_sessions` schema
- Store `Server_Information::service()` in `store()`
- Filter by `service` in `find_some()`
- Bump schema revision to `BOTAN_3_1` so existing databases are recreated

## Impact

Existing session databases are dropped and recreated on first open (same migration strategy as `PRE_BOTAN_3_0` → `BOTAN_3_0`). Session resumption continues to work; cached sessions are simply lost once on upgrade.

### DIFF
--- a/src/lib/tls/sessions_sql/tls_session_manager_sql.cpp
+++ b/src/lib/tls/sessions_sql/tls_session_manager_sql.cpp
@@ -29,13 +29,14 @@ void Session_Manager_SQL::create_or_migrate_and_open(std::string_view passphrase
       case CORRUPTED:
       case PRE_BOTAN_3_0:
       case EMPTY:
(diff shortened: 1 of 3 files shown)


randombit:
@oviano Change looks good to me but CI is failing

Issue #5435, opened 2026-03-12
'make check' fails if pwdhash is tested, but BOTAN_HAS_OS_UTILS is not defined


ffplexim:
When compiling and testing a small subset of Botan, I encountered a failure during `make check`:
```
pwdhash:
pwdhash ran 1 tests 1 FAILED
Failure 1: Not implemented No system clock available
```
(I'm using 3.10.0.)

One way to reproduce:
1. `./configure.py --minimized-build --enable-modules=argon2`
2. `make -j8`
3. `make check -j8`

KaganCanSit:
Hello,

I took a quick look at this issue and wanted to share a couple of thoughts that I think might help you for now.

The problem stems from the `tune_params()` call in the `Pwdhash_Tests::run()` function in `src/tests/test_pbkdf.cpp`. This function uses `measure_cost()` in `src/lib/pbkdf/argon2/argon2pwhash.cpp`. `measure_cost()` is conditionally linked to the `BOTAN_HAS_OS_UTILS` macro in `src/lib/utils/time_utils.h`; if this macro is not defined, the `Not_Implemented` branch is selected at compile time and thrown at runtime.

Temporary workaround until this issue is resolved:
```
./configure.py --minimized-build --enable-modules=argon2,os_utils
make -j8
make check -j8
```

ffplexim:
@KaganCanSit thanks a lot for looking into this! This made me realize that my set of modules should have `os_utils` in it since I plan to use `tune_params()`.

KaganCanSit:
This has been resolved via the development in branch #5438. Thank you for looking into this. It can now be closed @randombit.

PR #5436, opened 2026-03-12
fix: add missing os_utils dependency to argon2 module

KaganCanSit:
Hello,

This change was made to correct the error in notification #5435.

`argon2pwhash.cpp` includes `time_utils.h` which depends on `os_utils` for `measure_cost()`. Without this dependency declared, minimized builds (`--minimized-build --enable-modules=argon2`) fail at runtime with `Not_Implemented("No system clock available")` when `tune_params()` is called.

```bash
make check -j8
# ------------------------------------------------------------------------------------
Tests complete ran 61340 tests in 213.52 msec 1 tests failed (in pwdhash)
ERROR:root:Error running ./botan-test --data-dir=./src/tests/data
make: *** [Makefile:65: check] Error 1

./botan-test --data-dir=./src/tests/data
# ------------------------------------------------------------------------------------
Tests complete ran 61340 tests in 211.70 msec 1 tests failed (in pwdhash)
```

Best regards.

### DIFF
--- a/src/lib/pbkdf/argon2/info.txt
+++ b/src/lib/pbkdf/argon2/info.txt
@@ -8,6 +8,7 @@ name -> "Argon2"
 
 <requires>
 blake2
+os_utils
 </requires>
 
 <header:public>


randombit:
Thanks for the patch but this isn't really what we'd want to do here - if `os_utils` is disabled that presumably is an intentional choice by the user, and the only thing it's actually required for is to access the clock for tuning. The test should just handle `Not_Implemented` as a valid outcome for tuning.

PR #5438, opened 2026-03-13
Handle Not_Implemented in pwdhash tune_params test

KaganCanSit:
While investigating this issue, I initially added `os_utils` as a dependency in `argon2/info.txt`, assuming `argon2` should be fully functional in all build configurations. However, as @randombit pointed out in [the PR comment](https://github.com/randombit/botan/pull/5436#issuecomment-4050449271).

> "if os_utils is disabled that presumably is an intentional choice by the user, and the only thing it's actually required for is to access the clock for tuning. The test should just handle Not_Implemented as a valid outcome for tuning."

Following that guidance, this patch wraps the `tune_params()` call in a `try-catch` block and treats `Not_Implemented` as a valid outcome rather than a test failure. This is consistent with how other tests in the codebase handle optional functionality (e.g. `test_xof.cpp`, `test_certstor.cpp`).

Verified with:
```
--minimized-build --enable-modules=argon2
```

### DIFF
--- a/src/tests/test_pbkdf.cpp
+++ b/src/tests/test_pbkdf.cpp
@@ -8,6 +8,7 @@
 #include "tests.h"
 
 #if defined(BOTAN_HAS_PBKDF)
+   #include <botan/exceptn.h>
    #include <botan/pbkdf.h>
    #include <botan/pwdhash.h>
 #endif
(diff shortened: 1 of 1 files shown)


src/tests/test_pbkdf.cpp
```diff
@@ -88,28 +89,33 @@ class Pwdhash_Tests : public Test {
...
 
-               auto default_pwhash = pwdhash_fam->default_params();
-               std::vector<uint8_t> output3(32);
-               default_pwhash->hash(output3, password, salt);
+                  auto default_pwhash = pwdhash_fam->default_params();
+                  std::vector<uint8_t> output3(32);
+                  default_pwhash->hash(output3, password, salt);
+               } catch(const Botan::Not_Implemented&) {
+                  // tune_params requires os_utils for clock access; not available in minimized builds
+                  result.test_note("tune_params not available in current build config");
```
ffplexim: Would it make sense to fail the test here if `os_utils` is available?

KaganCanSit: I don’t see any specific distinctions, particularly using `#if define`, within the `tests` directory, and there doesn’t appear to be an additional `require` module defined in `src/lib/pbkdf/argon2/info.txt`.

I will be waiting for a review from the maintainers on this. Maybe:

```diff
diff --git a/src/tests/test_pbkdf.cpp b/src/tests/test_pbkdf.cpp
index 0a1c42877..cb2ab454e 100644
--- a/src/tests/test_pbkdf.cpp
+++ b/src/tests/test_pbkdf.cpp
@@ -113,8 +113,12 @@ class Pwdhash_Tests : public Test {
                   std::vector<uint8_t> output3(32);
                   default_pwhash->hash(output3, password, salt);
                } catch(const Botan::Not_Implemented&) {
+   #if defined(BOTAN_HAS_OS_UTILS)
                   // tune_params requires os_utils for clock access; not available in minimized builds
                   result.test_failure("tune_params threw Not_Implemented but os_utils is available");
+   #else
+                  result.test_note("tune params not available in current build config");
+   #endif
                }
 
                result.end_timer();
```

However, it seems more appropriate to keep a general message rather than adding `#if defined`; this way, the comment line can simply be removed or edited.

src/tests/test_pbkdf.cpp
```diff
@@ -88,28 +89,33 @@ class Pwdhash_Tests : public Test {
...
-               result.test_bin_eq("PasswordHash produced same output when run with same params", output1, output2);
+                  result.test_bin_eq("PasswordHash produced same output when run with same params", output1, output2);
 
-               auto default_pwhash = pwdhash_fam->default_params();
-               std::vector<uint8_t> output3(32);
-               default_pwhash->hash(output3, password, salt);
+                  auto default_pwhash = pwdhash_fam->default_params();
+                  std::vector<uint8_t> output3(32);
+                  default_pwhash->hash(output3, password, salt);
+               } catch(const Botan::Not_Implemented&) {
```
randombit: This puts way to much logic under this `catch` - if something else like say `default_params` threw `Not_Implemented` we'd probably want to know about it.

It's a bit awkward but I'd suggest something like

```
std::unique_ptr<Botan::PasswordHash> tuned_pwdhash;
try {
    tuned_pwdhash = pwdhash_fam->tune_params(...);
} catch(Not_Implemented&) {
   result.test_note(...);
}

if(tuned_pwdhash != nullptr) {

}
```

ffplexim:
Thanks @KaganCanSit and @randombit!

PR #5445, opened 2026-03-15
Remove redundant typename keywords

randombit:
`clang-tidy 22` complains about these

Ironically I think we needed these because of old Clang having bugs that required them in contexts where every other compiler was OK without them, guess we'll see what CI thinks.


randombit:
Yeah

```

  build/include/public/botan/strong_type.h:109:26: error: missing 'typename' prior to dependent type name 'T::value_type'
        using value_type = T::value_type;
```

PR #5451, opened 2026-03-16
MLDSA-composite

falko-strenzke:
This PR is based on https://github.com/randombit/botan/pull/5531 (context parameter encoding for ML-DSA).

Implementation of MLDSA composite defined in draft-ietf-lamps-pq-composite-sigs.

The various composite algorithms are modelled as parameter of the ML-DSA-composite algorithm.

Our internal review is still pending, this is why I mark it as draft. But early feedback on the overall architecture is welcome. Once we have agreed on the principal design, I can start with the MLKEM composite implementation. The code will be equivalent to a great degree.

(Also note that creating the branch in the upstream repo was not intended ...)

Open Tasks:

- [x] decide on parameterization model. either
  - ~~each algorithm combination (parameter set) is one algorithm in Botan (currently implemented)~~
  - or introduce generic "MLDSA-composite" algorithm with algorithm combination as parameter (DONE)
- [x] all values MLDSA_Composite_Param] should be available/visible regardless of the compilation support and result in Not_Implemented if something doesn't work
- [ ] ~~make the implementation experimental until the RFC is finalized~~ (Draft is far advanced in publication, so wire format changes are most unlikely now)
- [x] factor out the ECC private key OID encoding patch as distinct PR: #5532
- [x] factor out the ML-DSA context string argument into distinct PR: #5531


src/cli/perf_pk_sig.cpp
```diff
@@ -405,4 +410,57 @@ BOTAN_REGISTER_PERF_TEST("ML-DSA", PerfTest_ML_DSA);
...
+   #define BOTAN_REGISTER_PERF_TEST_TMPL(name, Perf_Class)                                   \
+      /* NOLINTNEXTLINE(cert-err58-cpp) */                                                   \
+      const Botan_CLI::PerfTest::Registration reg_perf_##Perf_Class##_##name(                \
+         #name, []() -> std::unique_ptr<Botan_CLI::PerfTest> {                               \
+            /* NOLINTNEXTLINE(bugprone-macro-parentheses) */                                 \
+            return std::make_unique<Perf_Class<Botan::MLDSA_Composite_Param::id_t::name>>(); \
+         })
+
+   #if defined(BOTAN_HAS_RSA)
+BOTAN_REGISTER_PERF_TEST_TMPL(MLDSA44_RSA2048_PKCS15_SHA256, PerfTest_MLDSA_Composite);
```
randombit: This approach is very unlike how the rest of the perf tests are presented where there is a single entry point which then iterates over a list of possible parameters (eg `botan speed ECDH`) IMO it should follow that style rather than presenting one perflet per parameter set.

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
```diff
@@ -60,6 +64,85 @@ DilithiumMode::Mode dilithium_mode_from_string(std::string_view str) {
...
+            const T& value() const { return m_value; }
+
+         private:
+            bool m_is_set;
+            T m_value;
+      };
+
+      explicit MLDSA_Signing_Parameters(std::string_view param_str) : m_rndmzd(true), m_ctx({}) {
+         const char* error_tmpl = "Parameter string '{}' is not a valid ML-DSA or Dilithium parameter specification";
+         if(param_str.empty()) {
```
randombit: This function seems like an excellent argument for #5021 ...

[Note added by @falko-strenzke: The active and current PR for the signature options mentioned above is #5489]

src/lib/pubkey/mldsa-composite/mldsa_comp_parameters.cpp
```diff
@@ -0,0 +1,365 @@
...
+const MLDSA_Composite_Param MLDSA_Composite_Param::mldsa_composite_registry[] = {
+
+#if defined(BOTAN_HAS_RSA)
+   MLDSA_Composite_Param(MLDSA_Composite_Param::id_t::MLDSA44_RSA2048_PKCS15_SHA256,
+                         "MLDSA44-RSA2048-PKCS15-SHA256",
+                         "COMPSIG-MLDSA44-RSA2048-PKCS15-SHA256",
+                         "SHA-256",
+                         DilithiumMode::ML_DSA_4x4,
+                         "RSA",
+                         "PKCS1v15(SHA-256)",
```
randombit, line 33: Why on earth are we standardizing new PKCSv1.5 schemes in 2026.... :/

src/lib/pubkey/mldsa-composite/mldsa_comp_parameters.h
```diff
@@ -0,0 +1,130 @@
...
+#include <botan/exceptn.h>
+#include <botan/types.h>
+#include <botan/internal/oid_map.h>
+
+namespace Botan {
+
+class BOTAN_PUBLIC_API(3, 0) MLDSA_Composite_Param {
+   public:
+      enum id_t : uint32_t /* NOLINT(*-enum-size,*-use-enum-class) */ {
+#if defined(BOTAN_HAS_RSA)
```
randombit: This class is public so having enumeration values disappear (and indeed, change their encoding values) depending on if the associated algorithm is available or not seems not ideal. These should be available/visible regardless and result in `Not_Implemented` if something doesn't work (indeed you would have to handle this case anyway, since for example it's entirely possible now for `BOTAN_HAS_ECDSA` to be true, but ECDSA over say brainpool384r1 is not available)

randombit (commented):
Not a full review at all, these comments are just what immediately jumped out at me from a brief scan.

A more broad observation is that this PR is way too large. At the very least the ECC private key OID encoding patch and the ML-DSA context string argument should both be split out into distinct PRs, with tests of their own independent of ML-DSA-composite.

Also looking how some now very complex policies are being dumped into a string here, I think something like #5021 is basically required.

[Note added by @falko-strenzke: The active and current PR for the signature options mentioned above is #5489]

src/cli/perf_pk_sig.cpp
falko-strenzke: `MLDSA44_RSA2048_PKCS15_SHA256` is not a parameter, it's an algorithm.
Currently, all the different algorithms defined in draft-ietf-lamps-pq-composite-sigs-15 are
modelled as such, that is, algorithms, in Botan.
I don't really see an alternative to this.
If it is desired to turn them into parameters of an abstract `MLDSA-composite` algorithm, that seems
also possible.
It doesn't reflect the to-be-RFCs modelling, though.
Furthermore, the availability of the individual parameter sets is determined by compiled support for
the traditional algorithm components.
The case that none are available could be caught at compile time, but I am not sure if the
conditional availability of parameter sets might break assumptions in some places.
I guess this is an unprecedented case.
Apart from these considerations, the approach to model X.509 algorithms as parameters is already
taken for ML-DSA and SLH-DSA.
What I would refrain from in any case is assigning a default parameter when none is given.
The choice seems too deliberate for the composites in my view, and would all the more have to vary
based on compiled support for the traditional components.

src/lib/pubkey/dilithium/dilithium_common/dilithium.cpp
falko-strenzke: Extending the public key API regarding the parameterization is a great idea. If you remember, I made a proposal some time ago. In my view, the API extension should not continue the usage of string function parameters. Instead, we should have a typed framework, that ideally already verifies parameter choices at compile time. At the same time, the legacy string-based API will most likely need continued support for a longer time. If we can agree on a reasonable approach here, I would try to see what we can do in the scope of the current BSI project, which runs until October of this year.

src/lib/pubkey/mldsa-composite/mldsa_comp_parameters.cpp
falko-strenzke: Because it was requested by some parties I guess...

src/lib/pubkey/mldsa-composite/mldsa_comp_parameters.h
falko-strenzke: Yeah, makes absolutely sense. I will refactor this.

falko-strenzke:
> At the very least the ECC private key OID encoding patch and the ML-DSA context string argument should both be split out into distinct PRs, with tests of their own independent of ML-DSA-composite.

For ML-DSA context this is straightforward. I put it on my task list.

@randombit 
For the ECC encoding PR I have one question: We have the [ECDSA key encoding implemented in the composite code](https://github.com/randombit/botan/pull/5451/changes#diff-1b50aa9347cf562c6604419af91b4b1d395b82820dc638c821d89f47c509f134R319). This will also be needed in the ML-KEM-composite. Should we place this also into the ECC Key class in some way or copy it verbatim in both types of composites?

falko-strenzke:
@randombit 
I would like to make the ML-DSA composite an experimental feature until the RFC is published. How do I mark this module as experimental, i.e., not compiled by default?

src/lib/pubkey/mldsa-composite/mldsa_comp_parameters.h
falko-strenzke: Done. The enum value are now defined unconditionally.

src/cli/perf_pk_sig.cpp
falko-strenzke: @reneme  @randombit How do we proceed with the issue of parameterization that I outline above? Should I keep the current model "every MLDSA-composite parameter set is an algorithm" or should I refactor it to the model "MLDSA-composite is an algorithm with the various parameter sets"?

I would favor a quick decision here because I will soon start with the implementation of MLKEM-composite and ideally would create it in the right spirit to begin with.

src/lib/pubkey/mldsa-composite/mldsa_comp_parameters.h
```diff
@@ -0,0 +1,136 @@
...
+      std::string prehash_func() const { return std::string(this->m_prehash_func); }
+
+      std::string traditional_algorithm() const { return std::string(this->m_traditional_algorithm); }
+
+      std::string traditional_padding() const { return std::string(this->m_traditional_padding); }
+
+      std::string curve() const { return std::string(this->m_curve); }
+
+      std::string get_traditional_algo_param_str() const;
+      size_t traditional_pubkey_encoded_size() const;
```
falko-strenzke: Function not defined, delete this declaration.

src/lib/pubkey/mldsa-composite/mldsa_comp.cpp
```diff
@@ -0,0 +1,435 @@
...
+#include <memory>
+#include <string_view>
+#include <vector>
+
+namespace Botan {
+
+namespace {
+std::span<const uint8_t> mldsa_pubkey_subspan(const MLDSA_Composite_Param& param, std::span<const uint8_t> key_bits) {
+   const OID oid(param.mldsa_oid_str());
+   const AlgorithmIdentifier aid(oid, AlgorithmIdentifier::Encoding_Option::USE_EMPTY_PARAM);
```
falko-strenzke: not needed

src/lib/pubkey/mldsa-composite/mldsa_comp.cpp
```diff
@@ -0,0 +1,435 @@
...
+   if(key_bits.size() <= param.mldsa_pubkey_size()) {
+      throw Invalid_Argument(fmt("encoded MLDSA component public key is too short (len = {})", key_bits.size()));
+   }
+   return std::span<const uint8_t>(key_bits.begin(), param.mldsa_pubkey_size());
+}
+
+std::span<const uint8_t> mldsa_privkey_subspan(const MLDSA_Composite_Param& param, std::span<const uint8_t> key_bits) {
+   const OID oid(param.mldsa_oid_str());
+   const AlgorithmIdentifier aid(oid, AlgorithmIdentifier::Encoding_Option::USE_EMPTY_PARAM);
```
falko-strenzke: oid and aid not needed

doc/api_ref/pubkey.rst
```diff
@@ -168,6 +168,15 @@ Currently two flavors of Dilithium are implemented in separate Botan modules:
  * ``dilithium_aes``, that uses AES instead of Keccak-based primitives.
    This mode is deprecated and will be removed in a future release.
 
+ML-DSA (draft-ietf-lamps-pq-composite-sigs-15)
```
falko-strenzke:
```suggestion
ML-DSA-composite (draft-ietf-lamps-pq-composite-sigs-15)
```

src/lib/pubkey/mldsa-composite/mldsa_comp_parameters.h
```diff
@@ -0,0 +1,136 @@
...
+         MLDSA65_ECDSA_P384_SHA512,
+         MLDSA65_ECDSA_brainpoolP256r1_SHA512,
+         MLDSA87_ECDSA_P384_SHA512,
+         MLDSA87_ECDSA_brainpoolP384r1_SHA512,
+         MLDSA87_ECDSA_P521_SHA512,
+         MLDSA44_Ed25519_SHA512,
+         MLDSA65_Ed25519_SHA512,
+         MLDSA87_Ed448_SHAKE256,
+      };
+
```
falko-strenzke, line 39:
```suggestion
static inline const char* generic_algo_name = "MLDSA-Composite";¬
```

src/lib/ffi/ffi_pkey_algs.cpp
```diff
@@ -7,6 +7,8 @@
 */
 
 #include <botan/ffi.h>
+#include <botan/mldsa_comp.h>
```
falko-strenzke: Need to guard this

src/lib/pubkey/mldsa-composite/mldsa_comp.h
```diff
@@ -0,0 +1,157 @@
...
+
+      /**
+       * Return the pessimistic estimated key strength, i.e., the smaller strength of the component keys.
+       * This is justified by the assumption that composite algorithms are used in the assumption that
+       * one of the component algorithms might be broken.
+       *
+       * @return the mimium of the components' estimated strengths.
+       */
+      size_t estimated_strength() const override {
+         return std::max(this->m_mldsa_pubkey->estimated_strength(), this->m_traditional_pubkey->estimated_strength());
```
TJ-91, line 51: >        * @return the mimium of the components' estimated strengths.

wouldn't `std::min` be correct?

src/lib/pubkey/mldsa-composite/mldsa_comp.cpp
```diff
@@ -0,0 +1,432 @@
...
+   if(key_bits.size() <= param.mldsa_privkey_size()) {
+      throw Invalid_Argument("encoded MLDSA component private key is too short");
+   }
+   return std::span<const uint8_t>(key_bits.begin(), param.mldsa_privkey_size());
+}
+
+std::span<const uint8_t> traditional_pubkey_subspan(const MLDSA_Composite_Param& param,
+                                                    std::span<const uint8_t> key_bits) {
+   const size_t offset = param.mldsa_pubkey_size();
+   if(key_bits.size() <= 1 + offset) {
```
TJ-91: Just a nit, I would write:

```
   if(key_bits.size() <= offset) {
```

This checks for "at least 1-byte trad key" vs "at least 2-byte trad key". Probably the first is what you intended.
(same for `traditional_privkey_subspan()`)

src/lib/pubkey/mldsa-composite/mldsa_comp.cpp
```diff
@@ -0,0 +1,432 @@
...
+      throw Invalid_Argument(fmt("encoded traditional component public key is too short (len = {})", key_bits.size()));
```
TJ-91: A bit misleading to print the total length here instead of size - offset for the trad component

src/lib/pubkey/mldsa-composite/mldsa_comp.cpp
```diff
@@ -0,0 +1,432 @@
...
+            BigInt ri;
+            BigInt si;
+            dec.start_sequence().decode(ri).decode(si).end_cons();
+            const auto group = Botan::EC_Group::from_name(m_parameters.curve());
+            const EC_Scalar r = EC_Scalar::from_bigint(group, ri);
+            const EC_Scalar s = EC_Scalar::from_bigint(group, si);
+
+            trad_sig_buf = EC_Scalar::serialize_pair(r, s);
+            trad_sig = trad_sig_buf;
+         }
```
TJ-91, line 139: Probably an else case to directly return false (instead of trying to verify the ASN.1 string) makes sense here if the trad algorithm is not supported/compiled.
(if the code can be reached in that state anyway)

src/lib/pubkey/mldsa-composite/mldsa_comp.cpp
```diff
@@ -0,0 +1,432 @@
...
+            std::vector<uint8_t> enc_sig;
+            const std::span<uint8_t> rs(trad_sig.begin(), trad_sig.begin() + trad_sig.size() / 2);
+            const std::span<uint8_t> ss(trad_sig.begin() + trad_sig.size() / 2, trad_sig.end());
+            const BigInt r(rs);
+            const BigInt s(ss);
+            DER_Encoder enc(enc_sig);
+            enc.start_sequence().encode(r).encode(s).end_cons();
+            std::swap(trad_sig, enc_sig);
+         }
+#endif
```
TJ-91, line 195: might be unreachable but I would throw if no case matches

src/lib/pubkey/mldsa-composite/mldsa_comp.cpp
```diff
@@ -0,0 +1,432 @@
...
+         msg.push_back(0);  // ctx = empty
+         msg.insert(msg.end(), ph.begin(), ph.end());
+
+         const std::span<const uint8_t> mldsa_sig(sig.begin(), sig.begin() + mldsa_sig_size);
+         std::span<const uint8_t> trad_sig(sig.begin() + mldsa_sig_size, sig.end());
+         std::vector<uint8_t> trad_sig_buf;
+
+#if defined(BOTAN_HAS_ECDSA)
+         if(m_parameters.traditional_algorithm() == "ECDSA") {
+            BER_Decoder dec(trad_sig);
```
TJ-91, line 125: For a verify operation I think it should catch decoding errors and similar and return false in such cases, insted of throwing.

src/lib/pubkey/mldsa-composite/mldsa_comp.h
falko-strenzke: We had discussed this in the project and agreed that strength should be considered "optimistically", i.e. assuming that both schemes have to be broken to forge a signature. Thus `max` is correct.

TJ-91: Then the documentation (comment above) should be changed to reflect that

src/lib/pubkey/mldsa-composite/mldsa_comp.cpp
falko-strenzke: Not sure what exactly you mean: an else in the preprocessor logic to handle the "ECDSA" case when it is not supported? That won't be necessary. But I don't see what else you could mean.

src/lib/pubkey/mldsa-composite/mldsa_comp.cpp
falko-strenzke: Similar like above, I don't understand which case you want to prevent. If ECDSA is not compiled in, then corresponding keys cannot be instantiated in the first place.

src/lib/pubkey/mldsa-composite/mldsa_comp.cpp
TJ-91: I simply meant to have a runtime-`else` like

```
if(m_parameters.traditional_algorithm() == "ECDSA") {
  ...
}
else {
  // throw invalid state
}
```

In the spirt of defensive programming

falko-strenzke: But this is not an error case, the else branch would be entered on RSA, X25519, and X448 keys.

TJ-91: Ok, then I misunderstood that, sorry!

src/lib/pubkey/mldsa-composite/mldsa_comp.cpp
```diff
@@ -0,0 +1,437 @@
...
+// static
+std::shared_ptr<Public_Key> MLDSA_Composite_PublicKey::load_traditional_public_key(const MLDSA_Composite_Param& param,
+                                                                                   std::span<const uint8_t> key_bits) {
+#if defined(BOTAN_HAS_ECDSA)
+   if(param.traditional_algorithm() == "ECDSA") {
+      const auto group = Botan::EC_Group::from_name(param.curve());
+      return std::make_shared<Botan::ECDSA_PublicKey>(group, EC_AffinePoint(group, key_bits));
+   }
+#endif
+   return load_public_key(param.get_traditional_algorithm_id(), key_bits);
```
sehlen-bsi: Is it possible that this loads keys of the wrong size (for instance an RSA key with 1024 bits)?

src/lib/pubkey/mldsa-composite/mldsa_comp.cpp
```diff
@@ -0,0 +1,437 @@
...
+            BER_Decoder dec(trad_sig);
+            BigInt ri;
+            BigInt si;
+            try {
+               dec.start_sequence().decode(ri).decode(si).end_cons();
+            } catch(Exception&) {
+               return false;
+            }
+            const auto group = Botan::EC_Group::from_name(m_parameters.curve());
+            const EC_Scalar r = EC_Scalar::from_bigint(group, ri);
```
sehlen-bsi, line 134: The Botan ECDSA verificaiton would usually obtian scalars through EC_Scalar::deserialize_pair() and this would perform validaiton. Here, it seems we don't perform the same validation. An AI-aided review warns me that this might maybe lead to malleability of encoded signatures. Please check.

src/lib/pubkey/mldsa-composite/mldsa_comp.h
```diff
@@ -0,0 +1,157 @@
...
+      }
+
+      bool check_key(RandomNumberGenerator& rng, bool strong) const override {
+         return this->m_mldsa_pubkey->check_key(rng, strong) && this->m_traditional_pubkey->check_key(rng, strong);
+      }
+
+      OID object_identifier() const override;
+
+      /**
+       * Return the pessimistic estimated key strength, i.e., the smaller strength of the component keys.
```
sehlen-bsi: Contradiction in the comment between minimum and maximum being returned.

src/tests/test_mldsa_composite.cpp
```diff
@@ -0,0 +1,537 @@
...
+         result.reserve(all_params.size());
+         for(const auto& param : all_params) {
+            result.push_back(run_detail_test(param.id()));
+         }
+         return result;
+      }
+
+   private:
+      static void check_encoded_ecdsa_private_key(std::span<uint8_t> composite_private_key, Test::Result& result) {
+         Botan::OID key_parameters;
```
sehlen-bsi, line 463: The key_parameters are not checked here which might be fine. The question is here (similar to the RSA case) if it is checked anywhere/when loading the key.

sehlen-bsi (commented):
I left a few smaller comments.

src/lib/pubkey/mldsa-composite/mldsa_comp.cpp
falko-strenzke: No, that's nonsense from AI. The function you name is not doing ASN.1 decoding.

src/lib/pubkey/mldsa-composite/mldsa_comp.cpp
falko-strenzke: Correct. I fixed that by adding a check during loading of public and private keys.

src/lib/pubkey/mldsa-composite/mldsa_comp.h
falko-strenzke: Right, fixed now.

src/tests/test_mldsa_composite.cpp
falko-strenzke: For the private key the check was indeed missing. I added this now together with a new test.

falko-strenzke:
> Implementation of MLDSA composite defined in draft-ietf-lamps-pq-composite-sigs-15.
> 
> The various composite algorithms are modelled as algorithms in Botan.
> 
> This PR also is dependend on #5531 (context parameter encoding for ML-DSA).
> 
> Our internal review is still pending, this is why I mark it as draft. But early feedback on the overall architecture is welcome. Once we have agreed on the principal design, I can start with the MLKEM composite implementation. The code will be equivalent to a great degree.
> 
> (Also note that creating the branch in the upstream repo was not intended ...)
> 
> Open Tasks:
> 
>     * [x]  decide on parameterization model. either
>       
>       * ~each algorithm combination (parameter set) is one algorithm in Botan (currently implemented)~
>       * or introduce generic "MLDSA-composite" algorithm with algorithm combination as parameter (DONE)
> 
>     * [x]  all values MLDSA_Composite_Param] should be available/visible regardless of the compilation support and result in Not_Implemented if something doesn't work
> 
>     * [ ]  ~~make the implementation experimental until the RFC is finalized~~
> 
>     * [x]  factor out the ECC private key OID encoding patch as distinct PR: [add parsing of ECC group from ECC parameters #5532](https://github.com/randombit/botan/pull/5532)
> 
>     * [x]  factor out the ML-DSA context string argument into distinct PR: [add ML-DSA signature context #5531](https://github.com/randombit/botan/pull/5531)

falko-strenzke:
Rebased and now stacked on https://github.com/randombit/botan/pull/5531 (context parameter encoding for ML-DSA).

PR #5458, opened 2026-03-17
Enable testing TLS::Strict_Policy with TLS 1.3

randombit:


randombit:
@reneme Just a ping if you had any context here - just a leftover that got forgotten? I was surprised to find this

reneme (approved):
Yeah, very likely just a leftover. Thanks.

randombit:
This has been refactored away now

PR #5466, opened 2026-03-18
Increase macOS XCode version

randombit:
As the current default version seems to be unable to link binaries (extraordinary)


randombit:
Didn't help

PR #5478, opened 2026-03-23
Add PKCS#12 KDF

dmazzella:
Added PKCS#12 KDF corresponding test vectors added


dmazzella:
@randombit any suggestion on this PR?

dmazzella:
@randombit do you have time to review this PR? I'd like to move forward with any improvements and submit the PR to PBE and PKCS#12.

src/tests/test_passhash.cpp
```diff
@@ -186,6 +190,63 @@ BOTAN_REGISTER_TEST("pbkdf", "passhash9", Passhash9_Tests);
...
+      Test::Result run_one_test(const std::string& algo_spec, const VarMap& vars) override {
+         const std::string passphrase = vars.get_req_str("Passphrase");
+         const std::vector<uint8_t> salt = vars.get_req_bin("Salt");
+         const size_t iterations = vars.get_req_sz("Iterations");
+         const std::vector<uint8_t> expected = vars.get_req_bin("Output");
+
+         Test::Result result(algo_spec);
+
+         // Parse "PKCS12-KDF(Hash,id)" to extract the hash name and purpose id
+         const auto lparen = algo_spec.find('(');
```
randombit: Use `SCAN_Name` for this

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.h
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+* PKCS#12 KDF (RFC 7292 Appendix B)
+*
+* The id parameter (purpose byte) is fixed at family construction time:
+*   1 = encryption key, 2 = IV, 3 = MAC key
+*
+* The family name is "PKCS12-KDF(hash,id)", e.g. "PKCS12-KDF(SHA-1,1)".
+*/
+class BOTAN_PUBLIC_API(3, 11) PKCS12_KDF final : public PasswordHash {
```
randombit: This header shouldn't be public API - that eg `argon2.h` is public was an accident in Botan3, and the plan is to make them internal in the next major version. Users should be going entirely through the top level class interface.

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+   size_t out_offset = 0;
+   while(out_offset < out_len) {
+      // Compute A = H^iterations(D || I)
+      hash->update(D);
+      hash->update(I);
+      hash->final(A.data());
+
+      for(size_t iter = 1; iter < iterations; ++iter) {
+         hash->update(A.data(), A.size());
```
randombit: Should be able to use `span` based interfaces here and most other places eg `hash->update(A)`

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.cpp
```diff
@@ -0,0 +1,206 @@
...
+      // then I_j = (I_j + B + 1) mod 2^(v*8)
+      copy_repeat(B.data(), v, A.data(), hash_len);
+
+      for(size_t j = 0; j < I_len; j += v) {
+         uint16_t carry = 1;
+         for(size_t k = v; k > 0; --k) {
+            carry += static_cast<uint16_t>(I[j + k - 1]) + static_cast<uint16_t>(B[k - 1]);
+            I[j + k - 1] = static_cast<uint8_t>(carry & 0xFF);
+            carry >>= 8;
+         }
```
randombit: This increment logic is replicated in the measurement loop maybe extract out to a function

randombit (changes requested):
@dmazzella Sorry have not a lot of time for reviews this week I did a brief pass and had a few comments generally looks pretty good though.

dmazzella:
> @dmazzella Sorry have not a lot of time for reviews this week I did a brief pass and had a few comments generally looks pretty good though.

No problem, I've made all the changes you suggested

dmazzella:
@randombit Do you have any available slots for a PR check?

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.cpp
```diff
@@ -0,0 +1,224 @@
...
+}  // namespace
+
+secure_vector<uint8_t> pkcs12_encode_password(std::string_view password) {
+   if(password.empty()) {
+      return secure_vector<uint8_t>({0, 0});
+   }
+   const std::vector<uint8_t> ucs2 = utf8_to_ucs2(std::string(password));
+   secure_vector<uint8_t> result(ucs2.begin(), ucs2.end());
+   result.push_back(0);
+   result.push_back(0);
```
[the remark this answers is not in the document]
dmazzella: @randombit what do you think about this?

dmazzella:
@randombit Do you have any available slots for a PR check?

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.cpp
```diff
@@ -0,0 +1,227 @@
...
+void PKCS12_KDF::derive_key(uint8_t out[],
+                            size_t out_len,
+                            const char* password,
+                            size_t password_len,
+                            const uint8_t salt[],
+                            size_t salt_len) const {
+   const std::string_view pwd =
+      (password != nullptr && password_len > 0) ? std::string_view(password, password_len) : std::string_view{};
+   pkcs12_kdf(out, out_len, pkcs12_encode_password(pwd), salt, salt_len, m_iterations, m_id, m_hash_name);
+}
```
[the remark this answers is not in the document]
dmazzella: @randombit what do you think about this? only docs?

dmazzella:
@randombit  @reneme  any available slots for a PR check?

src/tests/test_passhash.cpp
```diff
@@ -186,6 +190,46 @@ BOTAN_REGISTER_TEST("pbkdf", "passhash9", Passhash9_Tests);
...
+            return result;
+         }
+
+         auto pwdhash = pwdhash_fam->from_iterations(iterations);
+
+         std::vector<uint8_t> derived(expected.size());
+         pwdhash->hash(derived, passphrase, salt);
+         result.test_bin_eq("derived key", derived, expected);
+
+         // Verify from_params(iterations) produces the same result
```
randombit: This seems redundant (IIUC); maybe copied from another test?

src/tests/data/pbkdf/pkcs12_kdf.vec
```diff
@@ -0,0 +1,72 @@
+# PKCS#12 KDF Test Vectors
+#
+# SHA-1 vectors are taken from RFC 7292 Appendix B.4 (the only digest
```
randombit: I can find no test vectors in 7292 B.4 (or anywhere else in the RFC for that matter)

src/lib/pbkdf/pwdhash.cpp
```diff
@@ -100,6 +104,23 @@ std::unique_ptr<PasswordHashFamily> PasswordHashFamily::create(std::string_view
...
+      // The PKCS#12 KDF needs the hash's block size (RFC 7292 'v'); reject hashes
+      // that don't expose one rather than letting derive_key throw later.
+      if(auto hash = HashFunction::create(req.arg(0))) {
+         if(hash->hash_block_size() == 0) {
+            return nullptr;
+         }
+         const auto id_param = req.arg_as_integer(1);
+         if(id_param < 1 || id_param > 3) {
+            return nullptr;
+         }
```
randombit: This pre-validation step generally isn't something we do, and leads to a quite confusing result situation where for example `PKCS12-KDF(SHA-256,6)` returns `nullptr` (or throws `Not_Implemented`) rather than throwing `Invalid_Argument` with a string that describes the problem. Likewise for the hash block size.

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.cpp
```diff
@@ -0,0 +1,241 @@
...
+   // Create D (diversifier): v bytes of id
+   secure_vector<uint8_t> D(v, id);
+
+   // Create I = S || P
+   secure_vector<uint8_t> I(I_len);
+   if(S_len > 0) {
+      copy_repeat(std::span{I}.first(S_len), {salt, salt_len});
+   }
+   if(P_len > 0) {
+      copy_repeat(std::span{I}.subspan(S_len, P_len), pwd_bytes);
```
randombit: IIUC, `std::span{I}.last(P_len)` works here and makes the `S || P` split a bit more clear

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.cpp
```diff
@@ -0,0 +1,241 @@
...
+   return result;
+}
+
+namespace {
+
+void pkcs12_kdf_with_hash(uint8_t out[],
+                          size_t out_len,
+                          std::span<const uint8_t> pwd_bytes,
+                          const uint8_t salt[],
+                          size_t salt_len,
```
randombit: I think `salt` could just be a span also?

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.cpp
```diff
@@ -0,0 +1,241 @@
...
+      return;
+   }
+
+   const size_t hash_len = hash.output_length();
+
+   // Block size depends on hash algorithm (RFC 7292 uses the hash's block size as v)
+   const size_t v = hash.hash_block_size();
+   if(v == 0) {
+      throw Invalid_Argument(fmt("PKCS12-KDF does not support hash '{}': undefined block size", hash.name()));
+   }
```
randombit, line 86: Super nit thing but: move this check before the `out_len == 0` check. Otherwise eg `PKCS12-KDF(SomeHashWithNoBlockSize,1)` works for output length zero but not for other output lengths, when really the KDF is just not defined for such a hash.

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.cpp
```diff
@@ -0,0 +1,241 @@
...
+      for(size_t j = 0; j < I_len; j += v) {
+         bigendian_add_one(std::span{I}.subspan(j, v), B);
+      }
+   }
+}
+
+}  // namespace
+
+void pkcs12_kdf(uint8_t out[],
+                size_t out_len,
```
randombit: `out` and `salt` can both be `span` here

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.cpp
```diff
@@ -0,0 +1,241 @@
...
+}
+
+std::unique_ptr<PasswordHash> PKCS12_KDF_Family::tune_params(size_t output_length,
+                                                             uint64_t desired_msec,
+                                                             std::optional<size_t> /*max_memory*/,
+                                                             uint64_t tuning_msec) const {
+   // Benchmark the real KDF at a fixed iteration count and target output length,
+   // so the measurement captures hash(D || I), the full I-update carry loop, the
+   // inner A-rehash loop, and the per-output-block cost.
+   const size_t tuning_iterations = 64;
```
randombit: This tuning iterations might be a little small - have you tried `./botan pbkdf_tune --check` to see if this gives reasonably plausible results? For example on my machine

```
$ ./botan pbkdf_tune --check --algo=PBKDF2\(SHA-256\) 10 100 500 1000
For 10 ms selected PBKDF2(HMAC(SHA-256),50000) took 9.13415 msec to compute
For 100 ms selected PBKDF2(HMAC(SHA-256),524000) took 95.8112 msec to compute
For 500 ms selected PBKDF2(HMAC(SHA-256),2680000) took 490.484 msec to compute
For 1000 ms selected PBKDF2(HMAC(SHA-256),5316000) took 974.6 msec to compute
```

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.h
```diff
@@ -0,0 +1,129 @@
...
+ * @param salt the salt value
+ * @param salt_len length of salt in bytes
+ * @param iterations number of hash iterations
+ * @param id purpose ID: 1 = key, 2 = IV, 3 = MAC key
+ * @param hash_algo hash algorithm to use (default: SHA-1). Any registered hash
+ *                  with a defined block size is accepted (RFC 7292 needs 'v',
+ *                  the hash's block size); hashes that report block size 0
+ *                  are rejected at derivation time.
+ */
+void pkcs12_kdf(uint8_t out[],
```
randombit: Does this function need to be exposed, even to library-internal callers? There are analogous function decls in `pbkdf2.h` and `argon2.h` but these are basically historical artifacts that are kept only for compat, no need to replicate here unless required.

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.h
```diff
@@ -0,0 +1,129 @@
...
+                size_t out_len,
+                std::span<const uint8_t> password_bytes,
+                const uint8_t salt[],
+                size_t salt_len,
+                size_t iterations,
+                uint8_t id,
+                std::string_view hash_algo = "SHA-1");
```
randombit: If this function is kept, the hash should be explicit

randombit (commented):
Thanks, looks quite good at this point. Some mostly nit comments and also a concern re the test vector source.

dmazzella:
@randombit thanks for review. I have applied all your suggestions

src/tests/data/pbkdf/pkcs12_kdf.vec
dmazzella: @randombit  Sorry for the incorrect comment; the vectors were generated and validated using PKCS12_key_gen_uni.

[gen_pkcs12_vectors.c](https://github.com/user-attachments/files/28253190/gen_pkcs12_vectors.c)

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.cpp
```diff
@@ -0,0 +1,223 @@
...
+   clear_mem(ucs2);
+   result.push_back(0);
+   result.push_back(0);
+   return result;
+}
+
+namespace {
+
+void pkcs12_kdf_with_hash(uint8_t out[],
+                          size_t out_len,
```
randombit: Missed conversion to span here

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.h
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Encode a UTF-8 password for use with the PKCS#12 KDF (RFC 7292 Appendix B).
+ * Converts to PKCS#12 BMPString encoding (UCS-2BE / BMP-only, not full UTF-16)
+ * and appends a 2-byte null terminator ({0x00, 0x00}).
+ * An empty password produces {0x00, 0x00}, matching the RFC 7292 convention.
+ *
+ * Non-BMP characters are not representable in UCS-2/BMPString and are rejected
+ * by the underlying UTF-8-to-UCS-2 conversion.
+ */
+secure_vector<uint8_t> pkcs12_encode_password(std::string_view password);
```
randombit: This is exposed in the header but unused, could just be another anon namespace helper

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.cpp
```diff
@@ -0,0 +1,223 @@
...
+      copy_mem(buf.subspan(pos, to_copy), data.first(to_copy));
+      pos += to_copy;
+   }
+}
+
+/*
+* Big-endian addition: block = (block + addend + 1) mod 2^(len*8)
+*/
+void bigendian_add_one(std::span<uint8_t> block, std::span<const uint8_t> addend) {
+   uint16_t carry = 1;
```
randombit, line 48: This implicitly assumes block.size() == addend.size(), make this explicit with a `BOTAN_DEBUG_ASSERT`

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.h
```diff
@@ -0,0 +1,91 @@
...
+namespace Botan {
+
+/**
+* PKCS#12 KDF (RFC 7292 Appendix B)
+*
+* The id parameter (purpose byte) is fixed at family construction time:
+*   1 = encryption key, 2 = IV, 3 = MAC key
+*
+* The family spec used with PasswordHashFamily::create() is
+* "PKCS12-KDF(hash,id)", e.g. "PKCS12-KDF(SHA-1,1)" — two arguments.
```
randombit: No UTF-8 in comments plz

src/lib/pbkdf/pkcs12_kdf/pkcs12_kdf.cpp
```diff
@@ -0,0 +1,223 @@
...
+}  // namespace
+
+secure_vector<uint8_t> pkcs12_encode_password(std::string_view password) {
+   if(password.empty()) {
+      return secure_vector<uint8_t>({0, 0});
+   }
+
+   std::vector<uint8_t> ucs2 = utf8_to_ucs2(password);
+   secure_vector<uint8_t> result(ucs2.begin(), ucs2.end());
+   clear_mem(ucs2);
```
randombit: Should use `secure_scrub_memory` here

src/lib/pbkdf/pwdhash.cpp
```diff
@@ -100,6 +104,15 @@ std::unique_ptr<PasswordHashFamily> PasswordHashFamily::create(std::string_view
    }
 #endif
 
+#if defined(BOTAN_HAS_PKCS12_KDF)
+   if(req.algo_name() == "PKCS12-KDF" && req.arg_count() == 2) {
+      if(auto hash = HashFunction::create(req.arg(0))) {
+         const auto id_param = req.arg_as_integer(1);
+         return std::make_unique<PKCS12_KDF_Family>(std::move(hash), static_cast<uint8_t>(id_param));
```
randombit: The cast to uint8_t can narrow so that `PKCS12-KDF(SHA-256,257)` is accepted. I'd suggest just passing the id as a `size_t` through to `PKCS12_KDF_Family` and then validating the rang there. That's more in keeping with the general style where `size_t` is used for integers of no specific size; in our usage `uint8_t` is suggestive that any 8-bit integer is acceptable, which is not the case here.

randombit (commented):
Few more comments

dmazzella:
> Few more comments

done

src/tests/data/pbkdf/pkcs12_kdf.vec
```diff
@@ -0,0 +1,71 @@
...
+Salt = 0A58CF64530D823F
+Iterations = 2048
+Output = 7D7B4032062808484775EA7FC2AC3552D670B7D62B22280A7C296A4642ABE5F5
+
+[PKCS12-KDF(SHA-256,3)]
+
+Passphrase = password
+Salt = 0A58CF64530D823F
+Iterations = 2048
+Output = A790374DDEE08C419FFF12A3983E9CA9451BCC6B44B7973A19F5BB72CA61D7A7
```
randombit, line 86: None of the tests seem to have more than 1 hash's worth of output, which means the chaining/addition logic isn't being tested. I'd suggest adding several tests with 2 or 3 hashes worth of output (eg 64+ bytes for SHA-256)

randombit (commented):
One testing issue.

Also please add a bit of doc to `api_ref/pbkdf.rst` can be as simple as a header, a version added marker, a description of the string formats used to request it, and a ref to the RFC.

dmazzella:
> One testing issue.
> 
> Also please add a bit of doc to `api_ref/pbkdf.rst` can be as simple as a header, a version added marker, a description of the string formats used to request it, and a ref to the RFC.

done

randombit (approved):
LGTM thanks for your patience

dmazzella:
@randombit

`PKCS12_KDF` cannot express OpenSSL's "no password bytes" case

RFC 7292 Appendix B.1 defines password encoding for the KDF:
- Non-empty password -> UTF-16BE + NULL terminator
- Empty password -> `{0x00, 0x00}` (a single NULL terminator in UTF-16BE)

These encoded bytes form vector  `P`, which is repeated to a multiple of `v` (hash block size) and concatenated with `S` (repeated salt) to form `I = S || P`.

**OpenSSL's behavior** for empty password: it sets `passlen=0`, meaning `P_len = 0` — no `P` contribution at all, so `I = S` (salt only). This produces a different derived key than the RFC encoding.

Relevant code

`pkcs12_encode_password()` always returns `{0x00, 0x00}` for empty input — never an empty vector:

```cpp
secure_vector<uint8_t> pkcs12_encode_password(std::string_view password) {
   if(password.empty()) {
      return secure_vector<uint8_t>({0, 0});  // <- always 2 bytes, never empty
   }
   // ...
}
```

`PKCS12_KDF::derive_key()` unconditionally encodes the password before calling the internal KDF:

```cpp
void PKCS12_KDF::derive_key(...) const {
   const std::string_view pwd =
      (password != nullptr && password_len > 0)
         ? std::string_view(password, password_len)
         : std::string_view{};
   pkcs12_kdf_with_hash({out, out_len}, pkcs12_encode_password(pwd), ...);
}
```

`pkcs12_kdf_with_hash()` correctly computes P_len = pwd_len > 0 ? v * ((pwd_len + v - 1) / v) : 0 — but since it always receives at least {0x00, 0x00} (pwd_len=2), P_len is never 0.

Practical consequence
It's impossible to verify the MAC of .p12 files generated by OpenSSL with an empty password (openssl pkcs12 -export -passout pass:) using only the `PKCS12_KDF`  class API.

Possible solutions:
- Add a `derive_key_raw()` overload that accepts pre-encoded bytes (bypasses `pkcs12_encode_password`
- Expose `pkcs12_kdf_with_hash()` as an internal (non-anonymous-namespace) function
- Accept that Botan strictly follows RFC 7292 and does not support OpenSSL's empty-password behavior

Which option do you prefer?

randombit:
I'd be inclined to stick with just supporting the RFC behavior since IIUC all other implementations (eg Windows, Java, ...) handle this in the correct way, and this is a well known compatibility issue with OpenSSL PFX files. If there is in the end somehow enormous interest in this, we can add it later via an internal-only function that the PFX parser can try as a fallback for the empty password case, leaving the outer PasswordHash RFC-only.

PR #5489, opened 2026-03-26
Add PK_Signature_Options

randombit:

The PR previously known as #5021 #4318

### DIFF
--- a/src/cli/pubkey.cpp
+++ b/src/cli/pubkey.cpp
@@ -16,12 +16,12 @@
    #include <botan/hex.h>
    #include <botan/pk_algs.h>
    #include <botan/pk_keys.h>
+   #include <botan/pk_options.h>
    #include <botan/pkcs8.h>
    #include <botan/pubkey.h>
    #include <botan/x509_key.h>
    #include <botan/internal/workfactor.h>
    #include <fstream>
-   #include <sstream>
 
    #if defined(BOTAN_HAS_DL_GROUP)
       #include <botan/dl_group.h>
@@ -96,26 +96,17 @@ BOTAN_REGISTER_COMMAND("keygen", PK_Keygen);
 
 namespace {
 
-std::string choose_sig_padding(const std::string& key, const std::string& padding, const std::string& hash) {
-   if(key == "RSA") {
-      std::ostringstream oss;
-      if(padding.empty()) {
-         oss << "PSS";
-      } else {
-         oss << padding;
-      }
-
-      oss << "(" << hash << ")";
-      return oss.str();
-   } else if(padding.empty()) {
-      return hash;
-   } else if(hash.empty()) {
-      return padding;
-   } else {
-      std::ostringstream oss;
-      oss << padding << "(" << hash << ")";
-      return oss.str();
+Botan::PK_Signature_Options sig_options(
+   std::string_view key, std::string_view padding, std::string_view hash, bool use_der, std::string_view provider) {
+   if(key == "RSA" && padding.empty()) {
+      return sig_options(key, "PSS", hash, use_der, provider);
    }
+
+   return Botan::PK_Signature_Options()
+      .with_hash(hash)
+      .with_padding(padding)
+      .with_der_encoded_signature(use_der)
+      .with_provider(provider);
 }
 
 }  // namespace
@@ -198,21 +189,14 @@ class PK_Sign final : public Command {
             throw CLI_Error_Unsupported("hashing", hash_fn);
          }
 
-         const std::string sig_padding = choose_sig_padding(key->algo_name(), get_arg("padding"), hash_fn);
-
-         auto format = Botan::Signature_Format::Standard;
-
-         if(flag_set("der-format")) {
-            if(!key->_signature_element_size_for_DER_encoding()) {
-               throw CLI_Usage_Error("Key type " + key->algo_name() +
-                                     " does not support DER formatting for signatures");
-            }
-            format = Botan::Signature_Format::DerSequence;
+         if(flag_set("der-format") && !key->_signature_element_size_for_DER_encoding()) {
+            throw CLI_Usage_Error("Key type " + key->algo_name() + " does not support DER formatting for signatures");
          }
 
-         const std::string provider = get_arg("provider");
+         const auto options =
+            sig_options(key->algo_name(), get_arg("padding"), hash_fn, flag_set("der-format"), get_arg("provider"));
 
-         Botan::PK_Signer signer(*key, rng(), sig_padding, format, provider);
+         Botan::PK_Signer signer(*key, rng(), options);
 
          auto onData = [&signer](const uint8_t b[], size_t l) { signer.update(b, l); };
          Command::read_file(get_arg("file"), onData);
@@ -256,18 +240,9 @@ class PK_Verify final : public Command {
             throw CLI_Error_Unsupported("hashing", hash_fn);
          }
 
-         const std::string sig_padding = choose_sig_padding(key->algo_name(), get_arg("padding"), hash_fn);
+         const auto options = sig_options(key->algo_name(), get_arg("padding"), hash_fn, flag_set("der-format"), "");
 
-         auto format = Botan::Signature_Format::Standard;
-         if(flag_set("der-format")) {
-            if(key->message_parts() == 1) {
-               throw CLI_Usage_Error("Key type " + key->algo_name() +
-                                     " does not support DER formatting for signatures");
-            }
-            format = Botan::Signature_Format::DerSequence;
-         }
-
-         Botan::PK_Verifier verifier(*key, sig_padding, format);
+         Botan::PK_Verifier verifier(*key, options);
          auto onData = [&verifier](const uint8_t b[], size_t l) { verifier.update(b, l); };
          Command::read_file(get_arg("file"), onData);
 
--- a/src/cli/pubkey.cpp
+++ b/src/cli/pubkey.cpp
@@ -16,12 +16,12 @@
    #include <botan/hex.h>
    #include <botan/pk_algs.h>
    #include <botan/pk_keys.h>
+   #include <botan/pk_options.h>
    #include <botan/pkcs8.h>
    #include <botan/pubkey.h>
    #include <botan/x509_key.h>
    #include <botan/internal/workfactor.h>
    #include <fstream>
-   #include <sstream>
 
    #if defined(BOTAN_HAS_DL_GROUP)
       #include <botan/dl_group.h>
@@ -96,29 +96,83 @@ BOTAN_REGISTER_COMMAND("keygen", PK_Keygen);
 
 namespace {
 
-std::string choose_sig_padding(const std::string& key, const std::string& padding, const std::string& hash) {
-   if(key == "RSA") {
-      std::ostringstream oss;
-      if(padding.empty()) {
-         oss << "PSS";
-      } else {
-         oss << padding;
-      }
-
-      oss << "(" << hash << ")";
-      return oss.str();
-   } else if(padding.empty()) {
-      return hash;
-   } else if(hash.empty()) {
-      return padding;
-   } else {
-      std::ostringstream oss;
-      oss << padding << "(" << hash << ")";
-      return oss.str();
-   }
+/*
+* Signature schemes for which the caller must select a hash function; the
+* other schemes either fix the hash by the key type or do not use one.
+*/
+bool signature_scheme_requires_hash(std::string_view algo_name) {
+   return algo_name == "RSA" || algo_name == "DSA" || algo_name == "ECDSA" || algo_name == "ECGDSA" ||
+          algo_name == "ECKCDSA" || algo_name.starts_with("GOST-34.10");
 }
 
 }  // namespace
+
+/*
+* Shared handling of the signature related options of the sign and verify commands
+*/
+class PK_Signature_Command : public Command {
+   protected:
+      using Command::Command;
+
+      Botan::PK_Signature_Options signature_options(const Botan::Public_Key& key) const {
+         const std::string algo_name = key.algo_name();
+
+         auto options = Botan::PK_Signature_Options();
+
+         const bool prehashed = flag_set("prehashed");
+
+         std::string hash_fn = get_arg("hash");
+         // With a prehashed input an unnamed hash means the digest is signed as is
+         if(hash_fn.empty() && !prehashed && signature_scheme_requires_hash(algo_name)) {
+            hash_fn = "SHA-256";
+         }
+         if(!hash_fn.empty() && !Botan::HashFunction::create(hash_fn)) {
+            throw CLI_Error_Unsupported("hashing", hash_fn);
+         }
+         options = options.with_hash(hash_fn);
+
+         if(prehashed) {
+            options = options.with_externally_computed_prehash();
+         }
+
+         std::string padding = get_arg("padding");
+         if(padding.empty() && algo_name == "RSA") {
+            padding = "PSS";
+         }
+         options = options.with_padding(padding);
+
+         if(const auto prehash = get_arg_maybe("prehash")) {
+            if(*prehash == "default") {
+               options = options.with_prehash();
+            } else {
+               options = options.with_prehash(prehash);
+            }
+         }
+
+         if(const auto context = get_arg_maybe("context")) {
+            options = options.with_context(*context);
+         }
+
+         if(get_arg_maybe("salt-size")) {
+            options = options.with_salt_size(get_arg_sz("salt-size"));
+         }
+
+         if(flag_set("deterministic")) {
+            options = options.with_deterministic_signature();
+         }
+
+         if(flag_set("der-format")) {
+            if(!key._signature_element_size_for_DER_encoding()) {
+               throw CLI_Usage_Error("Key type " + algo_name + " does not support DER formatting for signatures");
+            }
+            options = options.with_der_encoded_signature();
+         }
+
+         options = options.with_provider(get_arg_or("provider", ""));
+
+         return options;
+      }
+};
 
 class PK_Fingerprint final : public Command {
    public:
@@ -175,12 +229,12 @@ std::unique_ptr<Botan::Private_Key> load_private_key(const std::string& key_file
 
 }  // namespace
 
-class PK_Sign final : public Command {
+class PK_Sign final : public PK_Signature_Command {
    public:
       PK_Sign() :
-            Command(
-               "sign --der-format --passphrase= --hash=SHA-256 --padding= --provider= --rng-type= --drbg-seed= key file") {
-      }
+            PK_Signature_Command(
+               "sign --der-format --passphrase= --hash= --padding= --prehash= --prehashed --context= --salt-size= "
+               "--deterministic --provider= --rng-type= --drbg-seed= key file") {}
 
       std::string group() const override { return "pubkey"; }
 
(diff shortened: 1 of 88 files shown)


src/lib/pubkey/pk_keys.h
```diff
@@ -107,6 +108,23 @@ class BOTAN_PUBLIC_API(3, 0) Asymmetric_Key /* NOLINT(*special-member-functions)
...
+      * This is only supported by a few algorithm types, so default
+      * implementation returns false.
+      *
+      * Note that support for contextual data may depend on both the algorithm
+      * and the library version. For example Ed25519 can support contextual
+      * data, using RFC 8032's Ed25519ctx construction, but this is not
+      * currently supported. In a future release, if Ed25519ctx was supported,
+      * then this function would start returning true for Ed25519 keys.
+      */
+      virtual bool supports_context_data() const;
```
falko-strenzke: Is it - from a principle point of view - really the key that supports the context parameter or is it rather the signature scheme? In the case of the currently affected signature schemes (EdDSA, ML-DSA, SLH-DSA, and ML-DSA-composite), that doesn't seem to make a difference, since they tie the key to specific signature scheme. However, you never know, if it might make a difference in the future...

src/cli/pubkey.cpp
```diff
@@ -96,26 +96,17 @@ BOTAN_REGISTER_COMMAND("keygen", PK_Keygen);
...
-   } else if(padding.empty()) {
-      return hash;
-   } else if(hash.empty()) {
-      return padding;
-   } else {
-      std::ostringstream oss;
-      oss << padding << "(" << hash << ")";
-      return oss.str();
+Botan::PK_Signature_Options sig_options(
+   std::string_view key, std::string_view padding, std::string_view hash, bool use_der, std::string_view provider) {
```
falko-strenzke: Using string arguments is not a good approach in my view. But maybe for now it is the simplest option. However, in the future, we can add a type framework for the parameters and start to define further constructors for the PK_Signature_Options.

falko-strenzke:
I did not perform a full review yet. But regarding the general approach, I think it is a great improvement. The new signature options type is a good thing in my view because it facilitates signature options to some degree already and lays a foundation for future improvemets. Further improvements could be typed signature parameters and possibly even compile-time validation of the parameter specification in the future.

src/lib/pubkey/pk_options.h
```diff
@@ -0,0 +1,197 @@
...
+* Signature generation/verification options
+*
+* The normal usage of this is in a builder style, eg
+*
+* PK_Signature_Options()
+*   .with_hash("SHA-256")
+*   .with_der_encoded_signature()
+*   .with_context("Foo")
+*/
+class BOTAN_PUBLIC_API(3, 6) PK_Signature_Options {
```
reneme, line 29: Probably 3.12 and maybe `final`.

src/lib/pubkey/pk_options.h
```diff
@@ -0,0 +1,197 @@
...
+      ///
+      /// This can be further parameterized by calling with_xxx functions
+      PK_Signature_Options() = default;
+
+      PK_Signature_Options(PK_Signature_Options&& other) = default;
+      PK_Signature_Options& operator=(PK_Signature_Options&& other) = default;
+
+      PK_Signature_Options(const PK_Signature_Options&) = default;
+      PK_Signature_Options& operator=(const PK_Signature_Options& other) = default;
+      ~PK_Signature_Options();
```
reneme, line 41: Rule-of-zero? The d'tor is currently defaulted in the cpp file, but there's no real reason for that, as none of the members would actually require that AFAICS.

randombit: I prefer to be explicit since rule of zero is ambiguous [does it mean that copy is definitely intentionally supported? or just that someone forgot to write the copy constructor...]

src/lib/pubkey/pk_options.h
randombit: 3.14 now, and `final`

randombit:
It being the 2 year anniversary of the original PR I think it's time to get this over the line :sweat_smile: 

@reneme I know you had concerns about options being able to be set and then silently ignored. I went with a somewhat different approach to addressing this than prior proposals; basically a `mutable` bitmask tracking if an option was looked at or not. If any options were unviewed in the course of the computation, fail.

src/lib/pubkey/pk_keys.h
randombit: Removed now - this was mostly there to support argument checking but that's now handled in a different way.

src/cli/pubkey.cpp
randombit: This is now replaced with a builder-style interface, there is a single string constructor taking just the hash since that's a relatively common convenience. [But might even be better to remove that too idk]

randombit:
@reneme @falko-strenzke I would like to merge this by ~ end of this month so as to unblock early the various PRs blocked by this as well as the (not yet implemented) options type for public key encryption/KEM, so that all of these changes (+ probably FFI) can be merged for 3.14

src/lib/pubkey/pk_options.h
```diff
@@ -0,0 +1,274 @@
...
+      /// schemes also sometimes offer a prehashing variant where the
+      /// message is hashed on its own, then the hash is signed.
+      ///
+      /// If given this specifies what hash function to use for prehashing.
+      /// If prehash is nullopt, this requests prehashing using an algorithm
+      /// specific default function
+      ///
+      /// If the scheme does not support prehashing, it will throw an
+      /// exception when presented with such an option.
+      PK_Signature_Options with_prehash(std::optional<std::string> prehash = std::nullopt);
```
falko-strenzke, line 87: Maybe not right now in this PR, but later, it can be considered to add the option "with_externally_computed_prehash()" or simply "with_external_prehash()", which causes the signature operation to expect the externally computed pre-hash as it's input.

To avoid confusion of both options, I would recommend the following change:


```suggestion
      PK_Signature_Options with_internal_prehash(std::optional<std::string> prehash = std::nullopt);
```

falko-strenzke (approved):
I didn't do a full review of all change sites, but I think I understood the concept and agree with it. I have only one suggested change regarding the naming of a function.

falko-strenzke:
> @reneme @falko-strenzke I would like to merge this by ~ end of this month so as to unblock early the various PRs blocked by this as well as the (not yet implemented) options type for public key encryption/KEM, so that all of these changes (+ probably FFI) can be merged for 3.14

@randombit 

Does that mean you will propose a different solution for the private key encoding variants from what @reneme and me agreed on in https://github.com/randombit/botan/pull/5307? Or is that out of scope from this workstream and I can take it on as planned in #5307?

src/lib/pubkey/pk_options.h
```diff
@@ -0,0 +1,274 @@
...
+      /// This is primarily intended for debugging and error messages;
+      /// the format is not fixed
+      std::string to_string() const;
+
+      /// Create a PK_Signature_Options specifying the hash to use
+      ///
+      /// Most but not all signature schemes require specifying the hash
+      ///
+      /// This can be further parameterized by calling with_xxx functions
+      explicit PK_Signature_Options(std::string_view hash_fn) : m_hash_fn(hash_fn) {}
```
reneme: It's a convenience thing, but it's also magic in a way. Not sure we actually want to provide this additional path to configure a hash function.

src/lib/pubkey/pk_options.h
```diff
@@ -0,0 +1,274 @@
...
+
+      /// Specify the hash function to use for signing/verification
+      ///
+      /// Most, but not all, schemes require specifying a hash function.
+      PK_Signature_Options with_hash(std::string_view hash);
```
reneme, line 60: Perhaps future work but worth considering up-front: it would be really cool if there was an `with_hash(std::unique_ptr<HashFunction>)` overload. E.g. to provide a custom hash function implementation (hardware-backed or whatever). For hash functions that might not be sooo important. But it might become more useful in other library aspects where we might introduce a similar builder pattern in the future.

Anyway: for that, we would need to be able to hold a `std::unique_ptr` inside the `PK_Signature_Options` object which would kill copy-ability. :( ... Hence, the current immutable-copy design actively prevents this use case (or at least complicates it by forcing us to implement explicit deep-copies for such params).

One could work around it with something like `with_hash_provider(std::function<std::unique_ptr<HashFunction>()>)`, but that sucks, IMO.

src/lib/pubkey/pk_options.h
```diff
@@ -0,0 +1,274 @@
...
+      std::optional<std::string> m_hash_fn;
+      std::optional<std::string> m_prehash;
+      std::optional<std::string> m_padding;
+      std::optional<std::vector<uint8_t>> m_context;
+      std::optional<std::string> m_provider;
+      std::optional<size_t> m_salt_size;
+      bool m_using_prehash = false;
+      bool m_use_der = false;
+      bool m_deterministic_sig = false;
+      bool m_explicit_trailer_field = false;
```
reneme, line 302: Nit: The boolean flags allow double-assignment like: `PK_Signature_Options().use_der().use_der().use_der()` while the other parameters would throw in that case. Really just a minor thing, I'm indifferent but I find it worth pointing out.

src/lib/pubkey/pk_options_impl.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+   /*
+   * Stopgap until PK_Signature_Options is exposed through the FFI: a
+   * trailing ",Deterministic" (eg "SHA-256,Deterministic") requests a
+   * deterministic signature, so that RFC 6979 ECDSA remains reachable
+   * through the string based interfaces.
+   */
+   const std::string_view det_suffix = ",Deterministic";
+   const bool deterministic = params.ends_with(det_suffix);
+   const std::string_view hash_params = deterministic ? params.substr(0, params.size() - det_suffix.size()) : params;
```
reneme, line 216: This changes the default from deterministic to randomized signatures. I guess that was intentional and it does make a lot of sense, IMO. Just wanted to point it out explicitly, as it might break some people's assumptions in a minor versions (not our's, but who knows).

reneme (approved):
I'm very happy with the outcome of this loooong design process and I hope it can serve as a foundation for a consistent and ergonomic API philosophy for many aspects in Botan4. Some nits, and discussions follow.

Also, I let Kimi K3 go through the changes and make sure that the API translation is sound and doesn't introduce surprises. Apart from the randomized/determinstic signature thing pointed out as a comment below, I think none of its findings are really problematic. Nevertheless, below is what it came up with:

<details>
<summary>Kimi K3 Review Results</summary>
The mechanical translation is largely faithful: the legacy string parser (`parse_legacy_sig_options`) reproduces the old per-scheme string formats accurately, including deprecated aliases. However, the commit **does change existing behavior in several places** — one of them major. Most appear intentional (the tests were updated to match), but the commit message ("This allows controlling all details...") doesn't disclose any of them.

## Behavior changes affecting existing call sites

### 1. ECDSA and DSA are now randomized by default — the big one

Old code unconditionally used RFC 6979 deterministic nonces when `BOTAN_HAS_RFC6979_GENERATOR` was enabled:

```137:145:src/lib/pubkey/ecdsa/ecdsa.cpp
#if defined(BOTAN_HAS_RFC6979_GENERATOR)
         if(options.using_deterministic_signature()) {
            m_rfc6979 = std::make_unique<RFC6979_Nonce_Generator>(
               this->rfc6979_hash_function(), m_group.get_order_bits(), ecdsa._private_key());
         }
#else
```

vs. the old unconditional `m_rfc6979 = std::make_unique<...>()`. Same in `dsa.cpp` (old: `BOTAN_UNUSED(rng)` + `generate_rfc6979_nonce`; new: random nonce unless `with_deterministic_signature()`).

**Consequence:** every existing caller — including the deprecated string constructors and the FFI, which funnels through them — now gets non-reproducible ECDSA/DSA signatures. Any application that (perhaps unwisely) depended on deterministic signatures breaks silently. It is clearly *intentional*: the RFC 6979 KATs were moved to explicit `with_deterministic_signature()` tests, `acvp_tests.py` adds `",Deterministic"` with an explanatory comment, and `test_cli.py` now needs `--deterministic` for its fixed expected signature. But it's buried in a commit whose message says nothing about it, and there's no `news.rst` entry in this commit. This deserves prominent changelog treatment.

The `",Deterministic"` string suffix (`pk_options_impl.cpp:200-202`) is a reasonable stopgap to keep RFC 6979 reachable via FFI/legacy strings.

### 2. SM2: the two constructors now disagree on the default

- Legacy `PK_Signer(key, rng, "")` → parser preserves the historical bug: empty params → `with_context("")` → **empty user id**.
- New `PK_Signer(key, rng)` (default options) → no context → `sm2_default_userid()` → **"1234567812345678"** (GM/T 0009).

Both are defensible in isolation (bug-compat vs. spec-correct), and the new `sm2_default_hash` test confirms the new default matches legacy `"1234567812345678,SM3"`. But a user migrating from `PK_Signer(key, rng, "")` to `PK_Signer(key, rng)` gets different, non-interoperable signatures. Worth a doc note on the new constructor.

### 3. XMSS / SPHINCS+ / SLH-DSA: parameters no longer silently ignored

Old XMSS ignored `params` entirely (`std::string_view /*params*/`); new code calls `validate_for_hash_based_signature`, which throws unless the hash option names the key's own hash. Proof this bites real call sites: the commit had to fix its own test vector —

```
- SigParams = SHA2_10_256
+ SigParams = SHA-256
```

in `src/tests/data/pubkey/api_sign.vec`. Any existing code passing a nominal-but-wrong params string to XMSS/SLH-DSA now throws. Stricter is better here, but it's a compat break.

### 4. Dilithium/SPHINCS+: "Pure" and verification params newly accepted

- Old Dilithium signing rejected `"Pure"` (`BOTAN_ARG_CHECK` allowed only `""`/`"Deterministic"`/`"Randomized"`); the new parser maps `"Pure"` → deterministic (`pk_options_impl.cpp:34`). More permissive.
- Old Dilithium *verification* rejected `"Deterministic"`/`"Randomized"`; now the parser accepts them and `PK_Verifier` centrally strips the deterministic flag (`pubkey.cpp:400-402`). More permissive.

### 5. Providers now reject a mismatched explicit provider

PKCS#11, TPMv1, TPM2 previously *ignored* the provider argument entirely (`/*provider*/`); now an explicit provider that doesn't match (e.g. `"tpm2"` on a PKCS#11 key) throws `Provider_Not_Found`. Only affects callers who passed a wrong provider string — previously silently honored. Again stricter, again undisclosed.

### 6. CLI default-hash semantics changed per scheme

Old CLI defaulted `--hash=SHA-256` for *everything*, which had two buggy consequences now fixed:
- `botan sign` with an ML-DSA key **failed** (Dilithium rejected the hash param) — now works.
- Ed25519 default signing used an ad-hoc non-standard prehash (`params="SHA-256"` → hashed op without the RFC 8032 domain separator); now defaults to pure Ed25519 (the new `cli_pk_sign_tests` verifies against the RFC 8032 KAT). SM2 similarly no longer uses the literal string "SHA-256" as the user id.

These are bug fixes, but scripts that verified old CLI-produced Ed25519 signatures with `--hash=SHA-256` must now pass `--prehash=SHA-256`.

## Verified as faithfully preserved

- **RSA legacy strings**: all aliases (`EMSA_PKCS1`, `EMSA3`, `PSSR`, `EMSA4`, `EMSA2`, `X9.31`, …) map correctly; PSS salt default (hash length) and `m_required_salt_len` semantics match the old two-constructor behavior; `ISO_9796_DS2` salt default and `imp`/`exp` trailer parsing match; `PKCS1v15(Raw)` and `PKCS1v15(Raw,hash)` map to the same `PKCS1v15_Raw` states; `Raw`/`Raw(hash)` validate the hash at construction as before.
- **X.509 paths**: RSA `parse_rsa_signature_algorithm` produces equivalent options (OID names were already `RSA/PKCS1v15(...)` in `static_oids.cpp`, so the new `scan.algo_name() == "PKCS1v15"` check is safe; no `RSA/X9.31` OIDs exist that the new code would reject). GOST OID→hash mapping unchanged. `x509_obj.cpp` widening `catch(Algorithm_Not_Found&)` to `catch(Lookup_Error&)` is correct since the new factory throws `Lookup_Error` directly.
- **Ed25519/Ed448**: `""`/`"Identity"`/`"Pure"` → pure; `"Ed25519ph"`/`"Ed448ph"` → default prehash with domain separator; anything else → named prehash without it. Exact match.
- **PKCS#11 mechanism maps**: canonical names (`PSS(SHA-256)`, `PKCS1v15(SHA-256)`, …) are all present in the map, and `PKCS1v15(Raw)` was added to cover the canonicalized alias. TPM2's mapping accepts the canonical `PSS`/`PKCS1v15` names, and its old rejection of PSS-with-salt is preserved as an explicit `Not_Implemented`.
- **`SCAN_Name::arg_as_integer`**: old defaults (`arg_as_integer(2, 0)` / hash-length) were only reachable when the arg was absent; the new code only parses when present and applies the same defaults in the constructors. Equivalent.

## Minor nits

- `"EMSA1(SHA-256,extra)"` for ECDSA & co. now silently drops `extra` (parser takes `arg(0)` unconditionally); old code required exactly one arg and threw. Slightly more permissive of malformed input.
- Ed25519 with params `"Raw"` or `"EMSA1(...)"` was previously accepted via `create_signature_hash` quirks; now throws. Arguably a fix.
- HSS-LMS rejects a hash option even when it names the key's hash, while XMSS/SLH-DSA accept it via `validate_for_hash_based_signature` — inconsistent strictness across sibling schemes.
- A few exception-type changes on error paths (`Algorithm_Not_Found` → `Invalid_Argument`/`Lookup_Error`; TPM2 `Invalid_Argument` → `Not_Implemented`). Only matters to callers catching specific types.
- `sig_padding` now `requires pubkey` in `info.txt` — a new module dependency edge; fine, but worth noting for minimal builds.

## Bottom line

No accidental behavior regressions found in the translation itself — the parser and factory faithfully reproduce the old string semantics, including the deliberate preservation of the SM2 empty-userid quirk. But the commit bundles at least six behavioral changes, of which **randomized-by-default ECDSA/DSA (#1)** is a significant, silent change for every existing caller and really should be called out in the commit message and `news.rst`. Items #2–#5 are defensible hardening, but they're compat-relevant and currently undocumented.

If you want, I can look at the new test file `test_pk_sig_options.cpp` next to check whether the intended semantics are actually pinned down, or draft the missing `news.rst` entry (in Agent mode).
</details>

src/lib/pubkey/pk_options.h
randombit: Yeah probably better to be explicit, I'll remove this, we can always re-add later if `with_hash` seems too onerous.

src/lib/pubkey/pk_options.h
randombit: Thanks, fixed

src/lib/pubkey/pk_options.h
randombit: I split these two notions out - now `with_prehash` is only used for internal prehashes (like Ed25519) and the RSA style external prehash (where we consume just a fixed length value with the assumption the caller hashed it for us) is `with_externally_computed_prehash`

randombit:
@falko-strenzke I think 5307 is basically unrelated and not touched either way (though I still do want to look at that before it merges)

src/lib/pubkey/pk_options.h
randombit: I'm open to this but also not my usecase so don't care too much. I'd be inclined to say just deep-copy the object - I'd imagine in almost all cases where this is actually useful, the `HashFunction` initially basically has no state, on first update call you initialize a handle on the hardware. So it's an extra malloc of a small struct, but that's true for copying the strings too...

src/lib/pubkey/pk_options_impl.cpp
randombit: Definitely intentional. I modified the commit message to mention it and will call it out specially in the release notes.

PR #5491, opened 2026-03-28
Add getters for RFC 3779 extensions to FFI

arckoor:
Adds some getters for the RFC 3779 extensions to the FFI. The FFI structs are not declared with `BOTAN_FFI_DECLARE_STRUCT` to allow them to contain an extra `m_writable` bool. I plan to add methods to build these extensions from scratch once e.g. #4996 is finalized and the FFI can create certificates from scratch, and I want to disallow writing to extension objects that have been obtained from an existing certificate (and can thus not be changed without creating a new certificate). This does prohibit copying the extension from an old cert to a new cert while also adjusting it, but I think this is better solved with a `_duplicate` method instead of allowing writes to all extension objects.

### DIFF
--- a/doc/api_ref/ffi.rst
+++ b/doc/api_ref/ffi.rst
@@ -2125,6 +2125,79 @@ X.509 Certificates
    Return a (statically allocated) string associated with the verification
    result, or NULL if the code is not known.
 
+.. cpp:function:: int botan_x509_ext_ip_addr_blocks_get_counts(botan_x509_cert_t cert, \
+                  size_t* v4_count, \
+                  size_t* v6_count)
+
+   Get info about the IP Address Blocks extension from `RFC 3779 <https://www.rfc-editor.org/rfc/rfc3779>`_.
+   ``v4_count`` is set to the number of v4 families contained in the extension,
+   ``v6_count`` to the number of v6 families. If the extension is not present, :cpp:enumerator:`BOTAN_FFI_ERROR_NO_VALUE` is returned.
+
+   If the extension is not present or an error occurs, ``v4_count`` and ``v6_count`` are not modified.
+
+.. cpp:function:: int botan_x509_ext_ip_addr_blocks_get_family(botan_x509_cert_t cert, \
+                  int ipv6, \
+                  size_t i, \
+                  int* has_safi, \
+                  uint8_t* safi, \
+                  int* present, \
+                  size_t* count)
+
+   Get info about a specific family in the extension.
+   ``ipv6`` should be set to 0 for v4 families, 1 for v6 families.
+   ``i`` is the local index for each family type, the first v4 family is at (``i = 0``, ``ipv6 = 0``), the first v6 family is at (``i = 0``, ``ipv6 = 1``).
+   The number of v4 / v6 families corresponds to the ``v4_count`` / ``v6_count`` values obtained from :cpp:func:`botan_x509_ext_ip_addr_blocks_get_counts`.
+   ``has_safi`` is set to 1 if the family has an associated SAFI, else 0.
+   ``safi`` contains the SAFI if the family has one, otherwise its value is not modified.
+   ``present`` is set to 1 if the family has range values, 0 if it is marked as "inherit".
+   ``count`` is set to the number of ranges contained if any, otherwise its value is not modified.
+
+   The output parameters ``has_safi``, ``safi``, ``present`` and ``count`` may be modified even if the extension is not present or some other error occurs.
+   In this event, the value of each output parameter after the call returns is undefined.
+
+.. cpp:function:: int botan_x509_ext_ip_addr_blocks_get_address(botan_x509_cert_t cert, \
+                  int ipv6, \
+                  size_t i, \
+                  size_t entry, \
+                  uint8_t min_out[], \
+                  uint8_t max_out[], \
+                  size_t* out_len)
+
+   Get info about a specific range in the extension.
+   ``ipv6`` and ``i`` behave as in :cpp:func:`botan_x509_ext_ip_addr_blocks_get_family`.
+   ``entry`` is the index to the range in the family, between 0 and (not including) ``count``.
+   ``min_out`` and ``max_out`` are set to the min and max addresses of the range respectively.
+   ``out_len`` should be set to 4 for v4 families, 16 for v6 families, the two arrays must also be that size.
+
+   The output parameters ``min_out``, ``max_out`` and ``out_len`` may be modified even if the extension is not present or some other error occurs.
+   In this event, the value of each output parameter after the call returns is undefined.
+
+.. cpp:function:: int botan_x509_ext_as_blocks_get_info(botan_x509_cert_t cert, \
+                  int asnum, \
+                  int* present, \
+                  size_t* count)
+
+   Get info about the AS Blocks extension from `RFC 3779 <https://www.rfc-editor.org/rfc/rfc3779>`_.
+   ``asnum`` should be set to 1 to get info about the ASNUM part of the extension, 0 for RDI.
+   ``present`` is set to 1 if a value is contained, 0 if that part of the extension is marked as "inherit".
+   If the part is not present at all, :cpp:enumerator:`BOTAN_FFI_ERROR_NO_VALUE` will be returned.
+   ``count`` is set to the number of entries for that part if any, otherwise its value is not modified.
+
+   If the extension is not present or an error occurs, ``present`` and ``count`` are not modified.
+
+.. cpp:function:: int botan_x509_ext_as_blocks_get_entry_at(botan_x509_cert_t cert, \
+                  int asnum, \
+                  size_t i, \
+                  uint32_t* min, \
+                  uint32_t* max)
+
+   Get info about a specific entry from the extension.
+   ``asnum`` behaves as in :cpp:func:`botan_x509_ext_as_blocks_get_info`, ``i`` is the index for that part,
+   between 0 and (not including) ``count``.
+   ``min`` and ``max`` will be set to the minimum and maximum AS numbers of the range respectively.
+
+   If the extension is not present or an error occurs, ``min`` and ``max`` are not modified.
+
 X.509 Certificate Revocation Lists
 ----------------------------------------
 
--- a/src/lib/ffi/ffi.h
+++ b/src/lib/ffi/ffi.h
@@ -2705,6 +2705,80 @@ int botan_x509_cert_verify(int* validation_result,
 BOTAN_FFI_EXPORT(2, 8) const char* botan_x509_cert_validation_status(int code);
 
 /*
+* X.509 Extensions
+*/
+
+/**
+* Get info about the IP Address Blocks extension from RFC 3779
+* @param v4_count is set to the number of v4 families contained in the extension
+* @param v6_count is set to the number of v6 families
+* @returns 0 on success, negative number on error
+*
+* If the extension is not present or an error occurs, `v4_count` and `v6_count` are not modified
+*/
+BOTAN_FFI_EXPORT(3, 13)
+int botan_x509_ext_ip_addr_blocks_get_counts(botan_x509_cert_t cert, size_t* v4_count, size_t* v6_count);
+
+/**
+* Get info about a specific family in the IP Address Blocks extension from RFC 3779
+* @param ipv6 must be set to 1 if the family is an IPv6 family, 0 for IPv4 families
+* @param i is the (local) index for this family kind (the first v4 family is at i = 0, ipv6 = 0; the first v6 family is at i = 0, ipv6 = 1)
+* @param has_safi will be set to 1 if the family has an associated SAFI
+* @param safi will be set to the families' SAFI, if it has one, otherwise `safi` is not modified
+* @param present is set to 1 if the family contains values (ranges), 0 if it is marked as "inherit"
+* @param count is set to the number of values (ranges), if they were present, otherwise `count` is not modified
+* @returns 0 on success, negative number on error
+*
+* The output parameters `has_safi`, `safi`, `present` and `count` may be modified even if the extension is not present or some other error occurs.
+* In this event, the value of each output parameter after the call returns is undefined.
+*/
+BOTAN_FFI_EXPORT(3, 13)
+int botan_x509_ext_ip_addr_blocks_get_family(
+   botan_x509_cert_t cert, int ipv6, size_t i, int* has_safi, uint8_t* safi, int* present, size_t* count);
+
+/**
+* Get info about a specific range in the IP Address Blocks extension from RFC 3779
+* @param ipv6 must be set to 1 if the family is an IPv6 family, 0 for IPv4 families
+* @param i is the (local) index of the family, see `botan_x509_ext_ip_addr_blocks_get_family`
+* @param entry is the index of the range
+* @param min_out is set to the lower address of the range
+* @param max_out is set to the upper address of the range
+* @param out_len is set to the length of the addresses (4 for IPv4, 16 for IPv6)
+* @returns 0 on success, negative number on error
+*
+* The output parameters `min_out`, `max_out` and `out_len` may be modified even if the extension is not present or some other error occurs.
+* In this event, the value of each output parameter after the call returns is undefined.
+*/
+BOTAN_FFI_EXPORT(3, 13)
+int botan_x509_ext_ip_addr_blocks_get_address(
+   botan_x509_cert_t cert, int ipv6, size_t i, size_t entry, uint8_t min_out[], uint8_t max_out[], size_t* out_len);
+
+/**
+* Get basic info about the AS Blocks extension from RFC 3779
+* @param asnum must be set to 1 to get info about AS numbers, 0 for RDIs (the type)
+* @param present is set to 1 if the extension contains entries for the type, 0 if it is marked as "inherit"
+* @param count is set to number of entries for this type, if it was present, otherwise `count` is not modified
+* @returns 0 on success, negative number on error
+*
+* If the extension is not present or an error occurs, `present` and `count` are not modified
+*/
+BOTAN_FFI_EXPORT(3, 13)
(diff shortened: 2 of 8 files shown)


src/lib/ffi/ffi_cert_ext.cpp
```diff
@@ -0,0 +1,309 @@
...
+#endif
+}
+
+int botan_x509_ext_ip_addr_blocks_get_address(botan_x509_ext_ip_addr_blocks_t ip_addr_blocks,
+                                              int ipv6,
+                                              size_t i,
+                                              size_t entry,
+                                              uint8_t min_out[],
+                                              uint8_t max_out[],
+                                              size_t* out_len) {
```
arckoor: I considered for a while to use view functions here, but allowing functions to receive two view callbacks seemed like a major headache (from both the FFI side and the consuming side), so I left it at writing to application supplied arrays. As its always just v4 / v6 addresses the application doesn't even need to guess the length, it's always exactly known.

arckoor:
@reneme if possible I'd like to redeem your offer to assist with something like this :p
I'll write the rst doc with some more general explanations of what the expected call flow is when the code itself is finalized

src/lib/ffi/ffi_cert_ext.cpp
```diff
@@ -0,0 +1,307 @@
...
+#if defined(BOTAN_HAS_X509_CERTIFICATES)
+   return ffi_guard_thunk(__func__, [=]() -> int {
+      const auto* const ext = safe_get(cert).v3_extensions().get_extension_object_as<Botan::Cert_Extension::ASBlocks>();
+
+      if(Botan::any_null_pointers(ext)) {
+         return BOTAN_FFI_ERROR_NO_VALUE;
+      }
+      return ffi_new_object(as_blocks,
+                            std::unique_ptr<Botan::Cert_Extension::ASBlocks>(
+                               dynamic_cast<Botan::Cert_Extension::ASBlocks*>(ext->copy().release())),
```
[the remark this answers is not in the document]
arckoor: Clang Tidy does not like this

src/lib/ffi/ffi_cert_ext.cpp
```diff
@@ -0,0 +1,307 @@
...
+      return BOTAN_FFI_ERROR_BAD_PARAMETER;
+   }
+   auto choice = std::get<Botan::Cert_Extension::IPAddressBlocks::IPAddressChoice<V>>(family.addr_choice());
+
+   if(!choice.ranges().has_value()) {
+      *present = 0;
+   } else {
+      *present = 1;
+      *count = choice.ranges().value().size();
+   }
```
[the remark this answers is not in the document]
arckoor: If `present` is not 1, don't read `count`. A count of 0 is very much different than the entry not being present.

src/python/botan3.py
```diff
@@ -2359,6 +2381,128 @@ def is_revoked(self, crl: X509CRL) -> bool:
...
+                    for entry in range(count.value):
+                        min_, max_ = _call_fn_returning_vec_pair(
+                            size, size, lambda mi, _, ma, out_len, ipv6=ipv6, i=i, entry=entry: _DLL.botan_x509_ext_ip_addr_blocks_get_address(
+                                self.__obj,
+                                c_int(ipv6),
+                                c_size_t(i),
+                                c_size_t(entry),
+                                mi,
+                                ma,
+                                out_len))
```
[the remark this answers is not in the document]
arckoor: I'm not set on this. I guess it is kind of valid criticism, but written as it currently is it'll just never fail. `_call_fn_returning_vec_pair` is just convenient to use since it already exists, but if you really want me to I can manually create the buffers.

arckoor:
@reneme could you take a look?

src/python/botan3.py
```diff
@@ -2448,6 +2470,128 @@ def is_revoked(self, crl: X509CRL) -> bool:
...
+                        ranges.append((tuple(min_), tuple(max_)))
+
+                safi = safi.value if has_safi.value == 1 else None
+                ranges = ranges if ranges is not None else None
```
randombit: This seems like a no-op?

If `ranges` is `None`, then assign `None` to `ranges`
If `ranges` is not `None`, then assign `ranges` to `ranges`

src/lib/ffi/ffi.h
```diff
@@ -2687,6 +2687,103 @@ int botan_x509_cert_verify(int* validation_result,
...
+*/
+BOTAN_FFI_EXPORT(3, 12)
+int botan_x509_ext_ip_addr_blocks_get_counts(botan_x509_ext_ip_addr_blocks_t ip_addr_blocks,
+                                             size_t* v4_count,
+                                             size_t* v6_count);
+
+/**
+* Get info about a specific family in the extension
+* @param ipv6 must be set to 1 if the family is an IPv6 family, 0 for IPv4 families
+* @param i is the global index of the family (i.e. the first v4 family is at index 0, the first v6 family is at `v4_count`)
```
randombit: This seems kind of awkward to me - why not just have this be the index within the family ie the first IPV4 is at `ipv6 = 0` `i = 0` and the first IPv6 is at `ipv6 = 1` `i = 0`

randombit (changes requested):
Sorry I did not get to the review here in time.

I'm not so confident of the overall shape of this. Having the getter being in term of the extension rather than the certificate is quite unlike the rest of the interface. It would make sense to have the extension as a standalone type _when_ we reach the point of actually being able to create certificates. But even then - there isn't that much value (that I can see) in being able to read the values out of an extension just created, you would just want some setters for that. I think it's better if these were all just functions on `botan_x509_cert_t`

src/lib/ffi/ffi.h
arckoor: The extension itself doesn't know when the V6's start, internally it's only a vector: `[V4, V4, ..., V6, ...]`, hence the global index and the `ipv6` bool. I can make it a local index, at the cost of an extra parameter that tells the function what the `v4_count` is, am fine with either. (I could also just recompute it in the function every time, but that just seems wasteful)

src/lib/ffi/ffi_cert_ext.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+      const auto& ext = safe_get(cert).v3_extensions().get_extension_object_as<Botan::Cert_Extension::ASBlocks>();
+      if(Botan::any_null_pointers(ext)) {
+         return BOTAN_FFI_ERROR_NO_VALUE;
+      }
+
+      const auto& asnum_or_rdi = asnum == 1 ? ext->as_identifiers().asnum() : ext->as_identifiers().rdi();
+      if(!asnum_or_rdi.has_value() || !asnum_or_rdi.value().ranges().has_value()) {
+         return BOTAN_FFI_ERROR_NO_VALUE;
+      }
```
arckoor, line 239: @randombit I've rewritten it to use certs directly, but for the AS Blocks I can have the situation that a) the extension is not present at all or b) the requested value (i.e. either ASNums or RDIs) is not present. Currently both return a `NO_VALUE` here, but I think it would be useful to differentiate between them. What could I return instead? `INVALID_OBJECT_STATE`? I've considered misusing `NULL_POINTER` for the non-present extension, since on the C++ side it is actually a nullptr if it's not present, but might be very unintuitive

src/lib/ffi/ffi.h
```diff
@@ -2694,6 +2694,82 @@ int botan_x509_cert_verify(int* validation_result,
 */
 BOTAN_FFI_EXPORT(2, 8) const char* botan_x509_cert_validation_status(int code);
 
+/*
+* X.509 Extensions
+*/
+
+/**
+* Get info about the IP Address Blocks extension
```
randombit: Just for doc purposes this (and following doc comments) should probably reference RFC 3779 and RPKI, so it's more clear this is unrelated to IP address stored in the SAN or elsewhere

src/lib/ffi/ffi.h
```diff
@@ -2694,6 +2694,82 @@ int botan_x509_cert_verify(int* validation_result,
...
+                                              size_t entry,
+                                              uint8_t min_out[],
+                                              uint8_t max_out[],
+                                              size_t* out_len);
+
+/**
+* Get basic info about the AS Blocks extension
+* @param asnum must be set to 1 to get info about AS numbers, 0 for RDIs (the type)
+* @param present is set to 1 if the extension contains entries for the type, 0 if it is marked as "inherit"
+* @param count is set to number of entries for this type, if it was present
```
randombit: > if it was present, otherwise `count` is not modified

src/lib/ffi/ffi.h
```diff
@@ -2694,6 +2694,82 @@ int botan_x509_cert_verify(int* validation_result,
...
+                                              size_t i,
+                                              size_t entry,
+                                              uint8_t min_out[],
+                                              uint8_t max_out[],
+                                              size_t* out_len);
+
+/**
+* Get basic info about the AS Blocks extension
+* @param asnum must be set to 1 to get info about AS numbers, 0 for RDIs (the type)
+* @param present is set to 1 if the extension contains entries for the type, 0 if it is marked as "inherit"
```
randombit, line 2759: > If an error occurs (eg due to an invalid parameter), `present` is not modified

randombit (commented):
Thanks looks good - I think my main concern at this point is the implementation doesn't write to the output parameters in certain error cases, which is fine, but the doc comments should call this out specifically that the values are not written at all if an error occurs. I left notes on some but there are probably others eg `min`/`max`

arckoor:
I added a disclaimer for each function that in case of an error "\<everything that is usually modified\> is not modified". For `botan_x509_ext_ip_addr_blocks_get_family` that is technically a lie, since if you pass the wrong family, e.g. `ipv6 = 1` when accessing a V4 family, `has_safi` and `safi` get set, after which it errors when trying to access the family. Not sure if that is a huge issue though.

doc/api_ref/ffi.rst
```diff
@@ -2125,6 +2125,79 @@ X.509 Certificates
...
+   Get info about a specific family in the extension.
+   ``v4_count`` is obtained from :cpp:func:`botan_x509_ext_ip_addr_blocks_get_counts`, ``ipv6`` should be set to 0 for v4 families,
+   1 for v6 families. ``i`` is the local index for each family type, the first v4 family is at ``i = 0``,
+   ``ipv6 = 0``, the first v6 family is at ``i = 0``, ``ipv6 = 1``.
+   ``has_safi`` is set to 1 if the family has an associated SAFI, else 0.
+   ``safi`` contains the SAFI if the family has one, otherwise its value is not modified.
+   ``present`` is set to 1 if the family has range values, 0 if it is marked as "inherit".
+   ``count`` is set to the number of ranges contained if any, otherwise its value is not modified.
+
+   If the extension is not present or an error occurs, ``has_safi``, ``safi``, ``present`` and ``count`` are modified.
```
randombit: "are not modified"?

doc/api_ref/ffi.rst
```diff
@@ -2125,6 +2125,79 @@ X.509 Certificates
...
+                  uint8_t max_out[], \
+                  size_t* out_len)
+
+   Get info about a specific range in the extension.
+   ``v4_count``, ``ipv6`` and ``i`` behave as in `botan_x509_ext_ip_addr_blocks_get_family`.
+   ``entry`` is the index to the range in the family, between 0 and (not including) ``count``.
+   ``min_out`` and ``max_out`` are set to the min and max addresses of the range respectively.
+   ``out_len`` should be set to 4 for v4 families, 16 for v6 families, the two arrays must also be that size.
+
+   If the extension is not present or an error occurs, ``min_out``, ``max_out``, and ``out_len`` are modified.
```
randombit: "are not modified"?

src/lib/ffi/ffi.h
```diff
@@ -2704,6 +2704,92 @@ int botan_x509_cert_verify(int* validation_result,
...
+* @param safi will be set to the families' SAFI, if it has one, othwise `safi` is not modified
+* @param present is set to 1 if the family contains values (ranges), 0 if it is marked as "inherit"
+* @param count is set to the number of values (ranges), if they were present, otherwise `count` is not modified
+* @returns 0 on success, negative number on error
+*
+* If the extension is not present or an error occurs, `has_safi`, `safi`, `present` and `count` are modified
+*/
+BOTAN_FFI_EXPORT(3, 13)
+int botan_x509_ext_ip_addr_blocks_get_family(botan_x509_cert_t cert,
+                                             size_t v4_count,
```
randombit: Let's remove this parameter and just have the function count the number of v4 addresses on the spot. This should anyway be fast, and the extension object could precompute and save the counts if the overhead of counting the addresses seems high.

src/lib/ffi/ffi.h
```diff
@@ -2704,6 +2704,92 @@ int botan_x509_cert_verify(int* validation_result,
...
+BOTAN_FFI_EXPORT(3, 13)
+int botan_x509_ext_ip_addr_blocks_get_counts(botan_x509_cert_t cert, size_t* v4_count, size_t* v6_count);
+
+/**
+* Get info about a specific family in the IP Address Blocks extension from RFC 3779
+* @param v4_count must be set to the number of v4 families contained in the cert, obtained from `botan_x509_ext_ip_addr_blocks_get_counts`
+* @param ipv6 must be set to 1 if the family is an IPv6 family, 0 for IPv4 families
+* @param i is the (local) index for this family kind (the first v4 family is at i = 0, ipv6 = 0; the first v6 family is at i = 0, ipv6 = 1)
+* @param has_safi will be set to 1 if the family has an associated SAFI
+* @param safi will be set to the families' SAFI, if it has one, othwise `safi` is not modified
```
randombit: Typo: otherwise

src/lib/ffi/ffi.h
```diff
@@ -2704,6 +2704,92 @@ int botan_x509_cert_verify(int* validation_result,
...
+* @param present is set to 1 if the family contains values (ranges), 0 if it is marked as "inherit"
+* @param count is set to the number of values (ranges), if they were present, otherwise `count` is not modified
+* @returns 0 on success, negative number on error
+*
+* If the extension is not present or an error occurs, `has_safi`, `safi`, `present` and `count` are modified
```
randombit: "are not modified"?

src/lib/ffi/ffi.h
```diff
@@ -2704,6 +2704,92 @@ int botan_x509_cert_verify(int* validation_result,
...
+* @param v4_count must be set to the number of v4 families present
+* @param ipv6 must be set to 1 if the family is an IPv6 family, 0 for IPv4 families
+* @param i is the (local) index of the family, see `botan_x509_ext_ip_addr_blocks_get_family`
+* @param entry is the index of the range
+* @param min_out is set to the lower address of the range
+* @param max_out is set to the upper address of the range
+* @param out_len is set to the length of the addresses (4 for IPv4, 16 for IPv6)
+* @returns 0 on success, negative number on error
+*
+* If the extension is not present or an error occurs, `min_out`, `max_out`, and `out_len` are modified
```
randombit: "are not modified"?

randombit (changes requested):
Some more comments. Mostly just quesions on the docs, plus one API change request which I think should be pretty easy to handle.

doc/api_ref/ffi.rst
randombit: If this "are modified" is intentional the sentences with that should be rewritten in way that makes it clear it's not an accidental omission of the `not` eg

> The output parameters ``has_safi``, ``safi``, ``present`` and ``count`` may be modified even if the extension is not present or some other error occurs. In this event, the value of each output parameter after the call returns is undefined.

arckoor: :grimacing:

arckoor: That second reply didn't even load for me at first, github moment
But the `not` is indeed missing, except for the `botan_x509_ext_ip_addr_blocks_get_family` case mentioned above where it's sort of a lie. I can more exhaustively rewrite it for that case

src/lib/ffi/ffi_cert_ext.cpp
```diff
@@ -0,0 +1,252 @@
...
+         return BOTAN_FFI_ERROR_BAD_PARAMETER;
+      }
+
+      const auto& ext =
+         safe_get(cert).v3_extensions().get_extension_object_as<Botan::Cert_Extension::IPAddressBlocks>();
+      if(Botan::any_null_pointers(ext)) {
+         return BOTAN_FFI_ERROR_NO_VALUE;
+      }
+
+      const size_t index = ipv6 == 0 ? i : ext->v4_count() + i;
```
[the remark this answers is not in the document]
arckoor: I suppose instead of checking for `index >= addr_blocks.size()` this could check like so:
```cpp
size_t limit = ipv6 == 0 ? ext->v4_count() : ext->v6_count();
if (i >= limit) {
    return BOTAN_FFI_ERROR_OUT_OF_RANGE;
}
```
IMO checking for `>= addr_blocks.size()` is more easily read as what it actually does, making sure no out of bounds index happens. Checking for the counts directly seems kind of arbitrary, but I don't mind either

arckoor:
Second commit only for readability, will squash before merge.
I ended up tracking the counts in the extension itself, might be interesting for regular C++ consumers as well, and this way it doesn't duplicate the logic as much (also avoids counting based on `afi` alone, and instead is type-backed).
The one unresolved copilot comment is worth of consideration, if you prefer that way of doing it :)

The CI analysis failure is unrelated I think, ruff seems very unhappy all of a sudden, probably got a version bump

src/lib/ffi/ffi_cert_ext.cpp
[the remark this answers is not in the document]
randombit: I'd be inclined for checking the index against the v4/v6 count directly (your suggested version is better than copilots), esp as it prevents a (unlikely, definitely application bug) footgun where `v4_count + i`  overflows. The `at(index)` makes it clear that worst case it's still a checked index.

randombit (approved):
LGTM! Just needs a squash and then can merge

PR #5495, opened 2026-03-29
refactor(rng): Deprecate random_vec(span) in favor of randomize()

KaganCanSit:
The `random_vec(std::span<uint8_t>)` overload marked with the `todo` message is functionally identical to `randomize(std::span<uint8_t>)`. It does not perform any resizing or allocation operations.

In this regard:
* `random_vec` function marked as deprecated.
* `todo` message updated for its removal in Botan4.
* Internal call sites have been migrated from `random_vec(std::span<uint8_t>)` to `randomize()`.

**Notes**
`random_vec(T& v, size_t bytes)` and `T random_vec(size_t bytes)` overloads are retained — they provide resize/allocate semantics that have no equivalent in `randomize()`.

**Notes 2**
The `@deprecated` Doxygen tag was considered but omitted — it is not used elsewhere in the project and triggers `-Wdocumentation-deprecated-sync` warnings across all translation units that include this header.

Updates:
06.05.2026 - Rebase master for `src/tests/test_strong_type.cpp` conflict.
24.05.2026 - Rebase master branch.
23.08.2026 - Rebased onto master. The deprecation rationale was moved out of the function's Doxygen block and into the `Other Deprecated Functionality` section of `deprecated.rst`, keeping the header comment limited to `@param/@throws/TODO(Botan4)`, consistent with `reseed()` elsewhere in this file.

### DIFF
--- a/doc/deprecated.rst
+++ b/doc/deprecated.rst
@@ -253,6 +253,14 @@ release, or where a backwards incompatible change is expected.
   "PSSR", "EMSA-PSS", "PSS-MGF1", "EMSA4": Use "PSS"
   "EMSA_X931", "EMSA2": Use "X9.31"
 
+- ``RandomNumberGenerator::random_vec(std::span<uint8_t>)`` is deprecated
+  in favor of ``RandomNumberGenerator::randomize(std::span<uint8_t>)``.
+  Unlike the other ``random_vec`` overloads it performs no resizing or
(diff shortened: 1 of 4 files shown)


KaganCanSit:
One question: should this also get a `news.rst` entry, given it is a user-visible deprecation?

@randombit Would you mind taking a look at this when you have time?

PR #5517, opened 2026-04-03
Discard TLS handshake state after handshake has completed

randombit:
Various bits of data relating to the handshake remained attached to the Channel even after the handshake completed, but in practice we only need a few pieces of information. Collect those into a new class, Active_Connection_State_{12,13}, and discard the rest.

### DIFF
--- a/src/lib/tls/tls12/tls_channel_impl_12.cpp
+++ b/src/lib/tls/tls12/tls_channel_impl_12.cpp
@@ -13,6 +13,7 @@
 #include <botan/tls_messages_12.h>
 #include <botan/tls_policy.h>
 #include <botan/x509cert.h>
+#include <botan/internal/concat_util.h>
 #include <botan/internal/ct_utils.h>
 #include <botan/internal/loadstor.h>
 #include <botan/internal/mem_utils.h>
@@ -97,19 +98,20 @@ std::shared_ptr<Connection_Cipher_State> Channel_Impl_12::write_cipher_state_epo
 }
 
 std::vector<X509_Certificate> Channel_Impl_12::peer_cert_chain() const {
-   if(const auto* active = active_state()) {
-      return get_peer_cert_chain(*active);
+   if(m_active_state.has_value()) {
+      return m_active_state->peer_certs();
    }
    return std::vector<X509_Certificate>();
 }
 
 std::optional<std::string> Channel_Impl_12::external_psk_identity() const {
-   const auto* state = (active_state() != nullptr) ? active_state() : pending_state();
-   if(state != nullptr) {
-      return state->psk_identity();
-   } else {
-      return std::nullopt;
+   if(m_active_state.has_value()) {
+      return m_active_state->psk_identity();
    }
+   if(const auto* state = pending_state()) {
+      return state->psk_identity();
+   }
+   return std::nullopt;
 }
 
 Handshake_State& Channel_Impl_12::create_handshake_state(Protocol_Version version) {
@@ -117,8 +119,8 @@ Handshake_State& Channel_Impl_12::create_handshake_state(Protocol_Version versio
       throw Internal_Error("create_handshake_state called during handshake");
    }
 
-   if(const auto* active = active_state()) {
-      const Protocol_Version active_version = active->version();
+   if(m_active_state.has_value()) {
+      const Protocol_Version active_version = m_active_state->version();
 
       if(active_version.is_datagram_protocol() != version.is_datagram_protocol()) {
          throw TLS_Exception(Alert::ProtocolVersion,
@@ -161,8 +163,8 @@ Handshake_State& Channel_Impl_12::create_handshake_state(Protocol_Version versio
 
    m_pending_state = new_handshake_state(std::move(io));
 
-   if(const auto* active = active_state()) {
-      m_pending_state->set_version(active->version());
+   if(m_active_state.has_value()) {
+      m_pending_state->set_version(m_active_state->version());
    }
 
    return *m_pending_state;
@@ -182,12 +184,12 @@ void Channel_Impl_12::renegotiate(bool force_full_renegotiation) {
       return;
    }
 
-   if(const auto* active = active_state()) {
+   if(m_active_state.has_value()) {
       if(!force_full_renegotiation) {
          force_full_renegotiation = !policy().allow_resumption_for_renegotiation();
       }
 
-      initiate_handshake(create_handshake_state(active->version()), force_full_renegotiation);
+      initiate_handshake(create_handshake_state(m_active_state->version()), force_full_renegotiation);
    } else {
       throw Invalid_State("Cannot renegotiate on inactive connection");
    }
@@ -250,7 +252,7 @@ void Channel_Impl_12::change_cipher_spec_writer(Connection_Side side) {
 }
 
 bool Channel_Impl_12::is_handshake_complete() const {
-   return (active_state() != nullptr);
+   return m_active_state.has_value();
 }
 
 bool Channel_Impl_12::is_active() const {
@@ -262,10 +264,11 @@ bool Channel_Impl_12::is_closed() const {
 }
 
 void Channel_Impl_12::activate_session() {
-   std::swap(m_active_state, m_pending_state);
-   m_pending_state.reset();
+   BOTAN_ASSERT_NONNULL(m_pending_state);
 
-   if(!m_active_state->version().is_datagram_protocol()) {
+   const auto& state = *m_pending_state;
+
+   if(!state.version().is_datagram_protocol()) {
       // TLS is easy just remove all but the current state
       const uint16_t current_epoch = sequence_numbers().current_write_epoch();
 
@@ -274,6 +277,15 @@ void Channel_Impl_12::activate_session() {
       map_remove_if(not_current_epoch, m_write_cipher_states);
       map_remove_if(not_current_epoch, m_read_cipher_states);
    }
+
+   // For DTLS, keep the handshake IO for last-flight retransmission.
+   if(m_is_datagram) {
+      m_active_state = Active_Connection_State_12(state, application_protocol(), m_pending_state->take_handshake_io());
+   } else {
+      m_active_state = Active_Connection_State_12(state, application_protocol());
+   }
+
+   m_pending_state.reset();
 
    callbacks().tls_session_activated();
 }
@@ -324,10 +336,10 @@ size_t Channel_Impl_12::from_peer(std::span<const uint8_t> data) {
             throw TLS_Exception(Alert::RecordOverflow, "TLS plaintext record is larger than allowed maximum");
          }
 
-         const bool epoch0_restart = m_is_datagram && record.epoch() == 0 && active_state() != nullptr;
+         const bool epoch0_restart = m_is_datagram && record.epoch() == 0 && m_active_state.has_value();
          BOTAN_ASSERT_IMPLICATION(epoch0_restart, allow_epoch0_restart, "Allowed state");
 
-         const bool initial_record = epoch0_restart || (pending_state() == nullptr && active_state() == nullptr);
+         const bool initial_record = epoch0_restart || (pending_state() == nullptr && !m_active_state.has_value());
          bool initial_handshake_message = false;
          if(record.type() == Record_Type::Handshake && !m_record_buf.empty()) {
             const Handshake_Type type = static_cast<Handshake_Type>(m_record_buf[0]);
@@ -345,8 +357,8 @@ size_t Channel_Impl_12::from_peer(std::span<const uint8_t> data) {
                   record.version() != pending->version()) {
                   throw TLS_Exception(Alert::ProtocolVersion, "Received unexpected record version");
                }
-            } else if(const auto* active = active_state()) {
-               if(record.version() != active->version() && !initial_handshake_message) {
+            } else if(m_active_state.has_value()) {
+               if(record.version() != m_active_state->version() && !initial_handshake_message) {
                   throw TLS_Exception(Alert::ProtocolVersion, "Received unexpected record version");
                }
             }
@@ -410,8 +422,10 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
             if(epoch == current_epoch) {
                create_handshake_state(record_version);
             } else if(current_epoch > 0 && epoch == current_epoch - 1) {
-               BOTAN_ASSERT(m_active_state, "Have active state here");
-               m_active_state->handshake_io().add_record(record.data(), record.size(), record_type, record_sequence);
+               BOTAN_ASSERT(m_active_state.has_value() && m_active_state->dtls_handshake_io(),
+                            "Have DTLS handshake IO for retransmission");
+               m_active_state->dtls_handshake_io()->add_record(
+                  record.data(), record.size(), record_type, record_sequence);
             }
          } else {
             create_handshake_state(record_version);
@@ -432,7 +446,7 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
             break;
          }
 
-         process_handshake_msg(active_state(), *pending, msg.first, msg.second, epoch0_restart);
+         process_handshake_msg(*pending, msg.first, msg.second, epoch0_restart);
 
          if(!m_pending_state) {
             break;
@@ -442,7 +456,7 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
 }
 
 void Channel_Impl_12::process_application_data(uint64_t seq_no, const secure_vector<uint8_t>& record) {
-   if(active_state() == nullptr) {
+   if(!m_active_state.has_value()) {
       throw Unexpected_Message("Application data before handshake done");
    }
 
@@ -463,19 +477,17 @@ void Channel_Impl_12::process_alert(const secure_vector<uint8_t>& record) {
    //    no_renegotiation
    //       Sent by the client in response to a hello request or by the
    //       server in response to a client hello after initial handshaking.
-   if(alert_msg.type() == Alert::NoRenegotiation && active_state() != nullptr) {
+   if(alert_msg.type() == Alert::NoRenegotiation && m_active_state.has_value()) {
       m_pending_state.reset();
    }
 
    callbacks().tls_alert(alert_msg);
 
-   if(alert_msg.is_fatal()) {
-      if(const auto* active = active_state()) {
-         BOTAN_ASSERT_NONNULL(active->server_hello());
-         const auto& session_id = active->server_hello()->session_id();
-         if(!session_id.empty()) {
-            session_manager().remove(Session_Handle(session_id));
-         }
+   // If the alert is fatal on an active session, prevent later resumptions
+   if(alert_msg.is_fatal() && m_active_state.has_value()) {
+      const auto& sid = m_active_state->session_id();
+      if(!sid.empty()) {
+         session_manager().remove(Session_Handle(sid));
       }
    }
 
@@ -496,10 +508,9 @@ void Channel_Impl_12::write_record(Connection_Cipher_State* cipher_state,
                                    Record_Type record_type,
                                    const uint8_t input[],
                                    size_t length) {
-   BOTAN_ASSERT(m_pending_state || m_active_state, "Some connection state exists");
+   BOTAN_ASSERT(m_pending_state || m_active_state.has_value(), "Some connection state exists");
 
-   const Protocol_Version record_version =
-      (m_pending_state) ? (m_pending_state->version()) : (m_active_state->version());
+   const Protocol_Version record_version = (m_pending_state) ? (m_pending_state->version()) : m_active_state->version();
 
    const uint64_t next_seq = sequence_numbers().next_write_sequence(epoch);
 
@@ -555,16 +566,15 @@ void Channel_Impl_12::send_alert(const Alert& alert) {
       }
    }
 
-   if(alert.type() == Alert::NoRenegotiation && active_state() != nullptr) {
+   if(alert.type() == Alert::NoRenegotiation && m_active_state.has_value()) {
       m_pending_state.reset();
    }
 
    if(alert.is_fatal()) {
-      if(const auto* active = active_state()) {
-         BOTAN_ASSERT_NONNULL(active->server_hello());
-         const auto& session_id = active->server_hello()->session_id();
-         if(!session_id.empty()) {
-            session_manager().remove(Session_Handle(Session_ID(session_id)));
+      if(m_active_state.has_value()) {
+         const auto& sid = m_active_state->session_id();
+         if(!sid.empty()) {
+            session_manager().remove(Session_Handle(sid));
          }
       }
       reset_state();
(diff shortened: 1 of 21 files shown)


src/lib/tls/tls12/tls_connection_state_12.h
```diff
@@ -0,0 +1,115 @@
...
+                                 std::optional<std::string> psk_identity,
+                                 std::span<const uint8_t> server_random,
+                                 Session_ID session_id,
+                                 secure_vector<uint8_t> master_secret,
+                                 std::string prf_algo,
+                                 bool client_supports_secure_renegotiation,
+                                 bool server_supports_secure_renegotiation,
+                                 std::vector<uint8_t> client_finished_verify_data,
+                                 std::vector<uint8_t> server_finished_verify_data,
+                                 bool supports_extended_master_secret);
```
reneme: Instead of passing in all those individual data points, what about passing in the full handshake state object as a whole? The constructor of `Active_Connection_State_12` could then scavenge all the things it needs by itself. That avoids this humongous parameter list and centralizes the knowledge about what's needed after the handshake in this class.

The call site in `tls_channel_12.cpp` would collapse to something like that:

```C++
state = Active_Connection_State_12(state, application_protocol());
```

src/lib/tls/tls12/tls_connection_state_12.h
```diff
@@ -0,0 +1,115 @@
...
+      bool supports_extended_master_secret() const { return m_supports_extended_master_secret; }
+
+      /**
+       * For DTLS: the handshake IO from the completed handshake, needed
+       * to retransmit the last flight when records arrive under the
+       * previous epoch. Null for stream TLS.
+       */
+      Handshake_IO* dtls_handshake_io() { return m_dtls_handshake_io.get(); }
+
+      void set_dtls_handshake_io(std::unique_ptr<Handshake_IO> io);
```
reneme: Somewhat related to the suggestion in the c'tor: This setter could be omitted entirely, if the constructor implementation would just decide internally that it needs this bit selectively for DTLS.

No setters, means immutable object, means less room for state issues in the future.

src/lib/tls/tls12/tls_channel_impl_12.h
```diff
@@ -235,8 +235,7 @@ class Channel_Impl_12 : public Channel_Impl {
       /* sequence number state */
       std::unique_ptr<Connection_Sequence_Numbers> m_sequence_numbers;
 
-      /* pending and active connection states */
-      std::unique_ptr<Handshake_State> m_active_state;
+      /* pending handshake state (null when no handshake is in progress) */
       std::unique_ptr<Handshake_State> m_pending_state;
```
reneme, line 237: Didn't think that through yet, putting it out anyway: let's make this a `std::variant<std::unique_ptr<Handshake_State>, std::unique_ptr<Active_Connection_State>>`. That would make it impossible to have both states set and available at the same time. In conjunction with my suggestion of the "smart c'tor" in `Active_Connection_State` the state transition code could look like that:

```C++
m_state = std::make_unique<Active_Connection_State>(std::get<std::unique_ptr<Handshake_State>(std::move(m_state)));
```

... admittedly, the bare variant is a bit awkward to use (hence the "didn't think that through, yet" disclaimer), but it buys us a clear cut contract:

1. Active state and pending state can never co-exist.
2. Active state must consume pending state to be created
3. There's essentially no way back (except re-creating a plain pending state)

... I'm hopeful that most of the variant's awkwardness could be hidden behind some private helper methods. I would look into this if you agree with the general idea.

src/lib/tls/tls12/tls_connection_state_12.h
```diff
@@ -0,0 +1,115 @@
...
+      std::vector<uint8_t> m_server_random;
+      Session_ID m_session_id;
+      secure_vector<uint8_t> m_master_secret;
+      std::string m_prf_algo;
+      bool m_client_supports_secure_renegotiation = false;
+      bool m_server_supports_secure_renegotiation = false;
+      std::vector<uint8_t> m_client_finished_verify_data;
+      std::vector<uint8_t> m_server_finished_verify_data;
+      bool m_supports_extended_master_secret = false;
+      std::unique_ptr<Handshake_IO> m_dtls_handshake_io;
```
reneme: Perhaps a hot-take, but: if this entire object would be _immutable_ and we explicitly removed the need to be able to `std::move` it around (e.g. by always using it within a unique ptr). Then, all those members could essentially be defined as `const`. And (IMHO) that woulld remove the need for accessors altogether and this state object could become a `struct {}` with publicly accessible members and a constructor to set it up.

... not the hill I'd want to die on, though.

reneme (commented):
Some high-level suggestions up-front. I didn't go through all the nitty-gritty access locations so far. The suggestions I made for TLS 1.2 would also apply for the TLS 1.3 part (perhaps apart from the `std::variant` idea in first approximation).

src/lib/tls/tls12/tls_connection_state_12.h
randombit: Yeah that's much better!

src/lib/tls/tls12/tls_channel_impl_12.h
randombit: I'm not sure this works - during renegotiation we really can have an active state + a pending state right?

src/lib/tls/tls12/tls_connection_state_12.h
randombit: Not my preference tbh, I get a bit squicked out by this even when it's an internal class (like `X509_Object`s `Signed_Data` struct) definitely don't like it for anything with a wider scope than that. No technical objection of course it works fine just don't like it.

randombit:
Going to wait for #5522  to land on master then rebase this PR onto it. The reverse order would be a pain to rebase due to wanting to keep the many different commits of #5522 distinct.

src/lib/tls/tls12/tls_channel_impl_12.h
reneme: Meh. Fair enough.

randombit:
Likewise going to wait for #5523 to land then rebase on top of that

src/lib/tls/tls12/tls_connection_state_12.h
randombit: Done!

randombit:
Rebased again after #5550

src/lib/tls/tls13/tls_server_impl_13.cpp
```diff
@@ -128,32 +149,49 @@ size_t Server_Impl_13::send_new_session_tickets(const size_t tickets) {
 }
 
 void Server_Impl_13::process_handshake_msg(Handshake_Message_13 message) {
+   BOTAN_STATE_CHECK(m_handshake != nullptr);
```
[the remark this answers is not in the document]
randombit: This isn't supposed to ever fire

src/lib/tls/tls12/tls_connection_state_12.h
randombit: done

randombit:
@reneme fyi probably will merge soon to unblock other work, lmk if you can review in next ~day or two.

src/lib/tls/tls13/tls_server_impl_13.cpp
```diff
@@ -100,17 +116,22 @@ size_t Server_Impl_13::send_new_session_tickets(const size_t tickets) {
    auto flight = aggregate_post_handshake_messages();
    size_t tickets_created = 0;
 
+   BOTAN_STATE_CHECK(m_active_state.has_value());
+
    for(size_t i = 0; i < tickets; ++i) {
       auto nonce = m_cipher_state->next_ticket_nonce();
+      const auto ticket_age_add = load_be<uint32_t>(rng().random_vec(4).data(), 0);
```
reneme:
```suggestion
      const uint32_t ticket_age_add = load_be(rng().random_array<4>());
```

src/lib/tls/tls13/tls_client_impl_13.cpp
```diff
@@ -70,39 +70,49 @@ Client_Impl_13::Client_Impl_13(const std::shared_ptr<Callbacks>& callbacks,
...
-   std::visit([&](auto msg) { handle(msg); }, m_handshake_state.received(std::move(message)));
+   std::visit(
+      [&](auto msg) {
+         if constexpr(std::is_constructible_v<Server_Post_Handshake_13_Message, decltype(msg)>) {
+            handle(msg);
+         } else {
+            throw TLS_Exception(Alert::UnexpectedMessage, "received an unexpected post-handshake message");
+         }
+      },
+      std::move(message));
```
reneme, line 107: Not necessarily in this PR: but this could become an `stl_util`, perhaps. Essentially a counterpart to `generalize_to`. Like so (roughly):

```C++
template <typename SpecificVariantT, GeneralVariantT>
std::optional<SpecificVariantT> specialize_to(GeneralVariantT&& v)
{
   return std::visit([]<typename AlternativeT>(AlternativeT&& obj) -> std::optional<SpecificVariantT> {
      if constexpr(std::is_constructible_v<SpecificVariantT, AlternativeT>) {
         return std::forward<AlternativeT>(obj);
      } else {
         return std::nullopt;
      },
      std::forward<GeneralVariantT>(v));
} 
```

... also useful in the server (and perhaps other places) and probably also works around the GCC 11 ICE mentioned in the server impl in a sustainable way.

src/lib/tls/tls13/tls_client_impl_13.cpp
```diff
@@ -388,18 +398,19 @@ void Client_Impl_13::handle(const Hello_Retry_Request& hrr) {
    // RFC 8446 4.1.4
    //    If a client receives a second HelloRetryRequest in the same connection [...],
    //    it MUST abort the handshake with an "unexpected_message" alert.
-   m_transitions.set_expected_next(Handshake_Type::ServerHello);
+   m_handshake->transitions.set_expected_next(Handshake_Type::ServerHello);
 }
 
 void Client_Impl_13::handle(const Encrypted_Extensions& encrypted_extensions_msg) {
    const auto& exts = encrypted_extensions_msg.extensions();
+   BOTAN_ASSERT_NONNULL(m_handshake);
```
reneme, line 409: This assertion isn't applied consistently throughout the `handle()` methods. Might be worthwhile. Perhaps applies also in the server impl (didn't check now).

src/lib/tls/tls13/tls_client_impl_13.cpp
```diff
@@ -630,12 +641,50 @@ void Client_Impl_13::handle(const Finished_13& finished_msg) {
...
+         }
+         if(m_handshake->resumed_session.has_value()) {
+            return m_handshake->resumed_session->session.peer_raw_public_key();
+         }
+         return nullptr;
+      }();
+
+      m_active_state = Active_Connection_State_13(m_handshake->state,
+                                                  std::move(certs),
+                                                  std::move(raw_pk),
```
reneme: Nit: if you'd call the extraction lambdas here instead of immediately, you could avoid the `std::move()`s. Perhaps rename them into `extract_certs` and `extract_raw_pk`, let them take the handshake wrapper type by reference (instead of capturing it) and make it so:

```suggestion
                                                  extract_certs(m_handshake),
                                                  extract_raw_pk(m_handshake),
```

... certainly a matter of taste, but communicates the intent more cleanly without explicit state, IMO.

reneme (commented):
Not a full review yet. I'll continue as I have time. Should be finished latest by Sunday, but feel free to merge earlier to unblock things. I'll go through it after the effect if needed. 

Mostly looked at the TLS 1.3 changes so far. A few nits attached.

src/lib/tls/tls12/tls_connection_state_12.h
```diff
@@ -0,0 +1,104 @@
...
+
+      Active_Connection_State_12(const Active_Connection_State_12&) = delete;
+      Active_Connection_State_12& operator=(const Active_Connection_State_12&) = delete;
+
+      Active_Connection_State_12(const Handshake_State& state, std::string application_protocol);
+
+      // DTLS variant: takes the handshake IO for replay of final flight
+      Active_Connection_State_12(const Handshake_State& state,
+                                 std::string application_protocol,
+                                 std::unique_ptr<Handshake_IO> io);
```
reneme, line 45: I'm guessing it would make sense to allow destroying the handshake I/O in DTLS as well. Once the peer has sent the first protected application payload data retransmissions of the handshake messages shouldn't occur anymore, right?

Perhaps that's an optimization for a follow-up PR, to not burry the necessary state machine changes in this fairly large patch, though.

Maybe leave a TODO?

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -463,18 +476,17 @@ void Channel_Impl_12::process_alert(const secure_vector<uint8_t>& record) {
...
 
    callbacks().tls_alert(alert_msg);
 
    if(alert_msg.is_fatal()) {
-      if(const auto* active = active_state()) {
-         BOTAN_ASSERT_NONNULL(active->server_hello());
-         const auto& session_id = active->server_hello()->session_id();
-         if(!session_id.empty()) {
-            session_manager().remove(Session_Handle(session_id));
+      if(m_active_state.has_value()) {
```
reneme: Nit: could be merged into a single if-statement now, given that the nested condition doesn't involve an assignment anymore.

src/lib/tls/tls13/tls_client_impl_13.cpp
randombit: It was already being done in `process_handshake_method` so was safe without it - but assert is cheap, added it to all `handle` methods in client and server.

src/lib/tls/tls13/tls_client_impl_13.cpp
randombit: Yeah that's nice, done

src/lib/tls/tls13/tls_client_impl_13.cpp
randombit: Yeah probably reasonable, I think this pattern happens elsewhere too. Not going to do this here

src/lib/tls/tls13/tls_server_impl_13.cpp
randombit: done

randombit:
Realistically the rework I'm doing that that is stacked on this PR is not sufficient cooked to ship in time for 3.12.0 so it's slightly less urgent. But only slightly; I'd still want this PR in 3.12.0 and at least for the moment that's scheduled for Tuesday.

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -345,8 +356,8 @@ size_t Channel_Impl_12::from_peer(std::span<const uint8_t> data) {
                   record.version() != pending->version()) {
                   throw TLS_Exception(Alert::ProtocolVersion, "Received unexpected record version");
                }
-            } else if(const auto* active = active_state()) {
-               if(record.version() != active->version() && !initial_handshake_message) {
+            } else if(m_active_state.has_value()) {
+               if(record.version() != m_active_state->version() && !initial_handshake_message) {
```
reneme, line 361: Nit: no need to nest these if statements anymore.

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -432,7 +445,7 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
             break;
          }
 
-         process_handshake_msg(active_state(), *pending, msg.first, msg.second, epoch0_restart);
+         process_handshake_msg(*pending, msg.first, msg.second, epoch0_restart);
```
reneme, line 449: This used to use `active_state()`. Strikes me as out of the ordinary that it now processes based on `*pending`. I don't really see why this should change. Or was it simply wrong before?

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -579,11 +589,8 @@ void Channel_Impl_12::secure_renegotiation_check(const Client_Hello_12* client_h
    BOTAN_ASSERT_NONNULL(client_hello);
    const bool secure_renegotiation = client_hello->secure_renegotiation();
 
-   if(const auto* active = active_state()) {
-      BOTAN_ASSERT_NONNULL(active->client_hello());
-      const bool active_sr = active->client_hello()->secure_renegotiation();
-
-      if(active_sr != secure_renegotiation) {
+   if(m_active_state.has_value()) {
+      if(m_active_state->client_supports_secure_renegotiation() != secure_renegotiation) {
```
reneme: Nit: collapse into single if?

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -602,11 +609,8 @@ void Channel_Impl_12::secure_renegotiation_check(const Server_Hello_12* server_h
    BOTAN_ASSERT_NONNULL(server_hello);
    const bool secure_renegotiation = server_hello->secure_renegotiation();
 
-   if(const auto* active = active_state()) {
-      BOTAN_ASSERT_NONNULL(active->server_hello());
-      const bool active_sr = active->server_hello()->secure_renegotiation();
```
reneme, line 607: Nit: collapse into single if?

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -622,28 +626,25 @@ void Channel_Impl_12::secure_renegotiation_check(const Server_Hello_12* server_h
...
-   if(const auto* active = active_state()) {
-      BOTAN_ASSERT_NONNULL(active->client_finished());
-      BOTAN_ASSERT_NONNULL(active->server_finished());
-      std::vector<uint8_t> buf = active->client_finished()->verify_data();
-      buf += active->server_finished()->verify_data();
+   if(m_active_state.has_value()) {
+      std::vector<uint8_t> buf = m_active_state->client_finished_verify_data();
+      buf += m_active_state->server_finished_verify_data();
       return buf;
    }
```
reneme, line 636:
```suggestion
   if(m_active_state.has_value()) {
      return concat(m_active_state->client_finished_verify_data(),
                    m_active_state->server_finished_verify_data())
}
```

... easier to understand and saves an allocation. (Needs an include of `concat_util.h`)

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -658,35 +659,31 @@ bool Channel_Impl_12::secure_renegotiation_supported() const {
...
+         throw Invalid_Argument("key_material_export context is too long");
       }
-
-      return SymmetricKey(prf->derive_key(length, master_secret, salt, as_span_of_bytes(label)));
-   } else {
-      throw Invalid_State("Channel_Impl_12::key_material_export connection not active");
+      salt.push_back(get_byte<0>(static_cast<uint16_t>(context_size)));
+      salt.push_back(get_byte<1>(static_cast<uint16_t>(context_size)));
+      salt += as_span_of_bytes(context);
    }
```
reneme:
```suggestion
   const auto salt = [&] {
      if(context.empty()) {
         return concat(m_active_state->client_random(), m_active_state->server_random());
      } else {
         return concat(m_active_state->client_random(),
                       m_active_state->server_random(),
                       store_be(static_cast<uint16_t>(context.size())),
                       as_span_of_bytes(context));
      }
   }();
```

... easier to understand and more efficient.

reneme (commented):
Done with the TLS 1.2 part. Some nits and a request for double-check and/or clarification. Otherwise LGTM. I'm doing a final pass over TLS 1.3 and should be done with that PR in latest an hour.

src/lib/tls/tls13/tls_client_impl_13.h
```diff
@@ -104,14 +104,15 @@ class Client_Impl_13 : public Channel_Impl_13 {
...
-
+      struct Pending_Handshake {
+            Client_Handshake_State_13 state;
+            Handshake_Transitions transitions;
+            bool should_send_ccs = false;
+            std::optional<Session_with_Handle> resumed_session;
+            std::optional<std::string> psk_identity;
+      };
+
+      std::unique_ptr<Pending_Handshake> m_handshake;
```
reneme, line 115: Given that `Pending_Handshake` is anyway fully defined in the header, we might consider using `std::optional` instead of `unique_ptr`. That would make "existence checks" slightly more explicit (via `has_value()` in lieu of `!= nullptr`) and an accidental dereference of a `std::nullopt` is an exception instead of a segfault.

Of course at the expense of wasted memory after the handshake is done, which is probably tipping the scale towards the `unique_ptr`. Anyway, leaving it here for consideration.

src/lib/tls/tls13/tls_server_impl_13.cpp
```diff
@@ -128,32 +149,49 @@ size_t Server_Impl_13::send_new_session_tickets(const size_t tickets) {
...
+            return true;
+         } else {
+            return false;
+         }
+      },
+      std::move(message));
+
+   if(!handled) {
+      throw TLS_Exception(Alert::UnexpectedMessage, "Received an unexpected post-handshake message");
+   }
```
reneme, line 186: Note to self: #5563

src/lib/tls/tls13/tls_server_impl_13.cpp
```diff
@@ -585,25 +640,64 @@ void Server_Impl_13::handle(const Finished_13& finished_msg) {
...
+      }();
+
+      const bool supports_psk_dhe =
+         m_handshake->state.client_hello().extensions().has<PSK_Key_Exchange_Modes>() &&
+         value_exists(m_handshake->state.client_hello().extensions().get<PSK_Key_Exchange_Modes>()->modes(),
+                      PSK_Key_Exchange_Mode::PSK_DHE_KE);
+
+      m_active_state = Active_Connection_State_13(m_handshake->state,
+                                                  std::move(certs),
+                                                  std::move(raw_pk),
```
reneme: Same as [previously pointed out in the client](https://github.com/randombit/botan/pull/5517#discussion_r3172454177), it would be nicer to call those lambdas here and avoid the move from the local variables.

As suggested before, passing `m_handshake` as a parameter into the extraction lambdas would make this even more explicit and free the lambdas from any captures.

reneme (approved):

src/lib/tls/tls13/tls_client_impl_13.h
randombit: > Of course at the expense of wasted memory after the handshake is done, which is probably tipping the scale towards the unique_ptr

The big driver here is specifically reducing the amount of memory that is consumed by a live post-handshake connection, so I don't think this works. It is a shame that at no point the C++ WG has bothered to decide to convert accessing a nullptr `unique_ptr` from UB to idk "just throw an exception". I am starting to understand why all the really big C++ projects just ignore the C++ standard library entirely and go their own way.

src/lib/tls/tls12/tls_connection_state_12.h
randombit: Not doing this here, but the big followup stacked on this PR I alluded to earlier is a big rework of DTLS, I'll look at this for that PR. And you are right, after application data is flowing it is safe to discard old handshake state.

src/lib/tls/tls12/tls_channel_impl_12.cpp
randombit: Here I find the if pending ... if active ... split easier to read than if both conditionals were placed together

src/lib/tls/tls12/tls_channel_impl_12.cpp
randombit: `*pending` was already being passed too , the change here is just that active state is referred to via the member variable rather than being passed as an argument.

reneme (approved):

Issue #5000, opened 2025-07-17
DTLS v1.3


randombit:
Holding ticket

reneme:
_Originally posted [here](https://github.com/randombit/botan/issues/1961), but this ticket is probably more appropriate._

It turns out we have a use case for DTLS 1.3 and will likely start working on an implementation based on the existing TLS 1.3 code. I did a first read-through of [RFC 9147](https://www.rfc-editor.org/rfc/rfc9147.html) and below are some scratchy notes on what probably has to be done. At this point, we're not ready to provide an exact time frame for this, but it should be measured in months not in years.

**TL;DR:** We'll probably need alternative implementations or significant extensions of `TLS::Record_Layer`, `TLS::Handshake_Layer`, and `TLS::Cipher_State`. All the higher layers of the protocol, particularly the implementation of the actual handshake messages and the TLS state machines in `TLS::[Client/Server]_Impl_13` can probably stay mostly untouched.

* **Record Layer**
  * Support demux of DTLS records ✅ 
  * Handling of Epoch and Sequence Number ✅
  * Deprotection of Epoch/SeqNumber (see [here](https://www.rfc-editor.org/rfc/rfc9147.html#name-record-number-encryption)) ✅
  * Replay protection via a sliding window per epoch ✅

* **Handshake Layer**
  * Fragmentation (additional header values + handling of them) ✅
    Note: Transcript hash is calculated from the TLS marshalling (w/o the additional headers)
  * Allow re-fragmenting handshake messages if necessary (MTU discovery)
  * Timeout and Retransmission ✅
    * Handle the retransmission state machine (see [RFC 9147 Section 5.8.1](https://www.rfc-editor.org/rfc/rfc9147.html#section-5.8.1))
    * Example: https://www.rfc-editor.org/rfc/rfc9147.html#section-6
    * Probably `Handshake_Layer_13` will need to be able to actively create messages (i.e. ACKs)

* **Key Management in `Cipher_State`**
  * Explicit cipher state handling (Epochs) ✅
    Out-of-order messages at epoch boundaries requires keeping multiple sets of keys for some period ✅
  * Adopt DTLS-specific HKDF labels ✅

* **Handshake Messages and Extensions**
  * ClientHello/HelloRetryRequest
[...]

PR #5523, opened 2026-04-08
Update BoGo shim, various TLS updates

randombit:
Rebases the BoGo patches onto the latest commit to `main` of BoringSSL

Fixes and additions to the prod library include

* RFC 6066 has that we should not echo SNI in a resumption; BoGo applies this to TLS 1.3 as well which seems reasonable. This however does imply that the published RFC 8448 test vectors are incorrect :cry: 
* The client should verify that the server selected an ALPN protocol that it offered
* Add support for RFC 9528 PSK import
* ~More KeyUpdate throttling; BoGo assumes a counter based approach which is nonsensical and easily bypassed, but easy to implement and it allows testing throttling behavior in BoGo so fine.~
* ~Send the certificate type extension in the pure TLS 1.2 client hello as well as the 1.3 hello~ Avoid sending `certificate_type` extension if we might possibly negotiation 1.2


src/lib/tls/tls_algos.cpp
```diff
@@ -133,43 +136,122 @@ Auth_Method auth_method_from_string(std::string_view str) {
...
+   codes[n++] = static_cast<uint16_t>(Group_Params_Code::HYBRID_X25519_eFRODOKEM_640_AES_OQS);
+   #endif
+
+   #if defined(BOTAN_HAS_X448)
+   codes[n++] = static_cast<uint16_t>(Group_Params_Code::HYBRID_X448_eFRODOKEM_976_SHAKE_OQS);
+   codes[n++] = static_cast<uint16_t>(Group_Params_Code::HYBRID_X448_eFRODOKEM_976_AES_OQS);
+   #endif
+#endif
+
+   std::sort(codes.begin(), codes.begin() + n);
```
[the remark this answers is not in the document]
randombit: > calls std::sort, which is not constexpr in C++20 

Or is it

src/bogo_shim/bogo_shim.cpp
```diff
@@ -441,8 +448,18 @@ class Shim_Socket final {
...
-         m_socket = -1;
+         if(m_socket >= 0) {
+            // Signal that we are done writing so pending alert records
+            // are delivered with a FIN rather than lost to a RST.
+            ::shutdown(m_socket, SHUT_WR);
+            // Drain unread incoming data; if the receive buffer is
+            // non-empty when we close(), the kernel sends RST which
+            // discards our outgoing data (including any alert we sent).
+            char buf[256];
+            while(::read(m_socket, buf, sizeof(buf)) > 0) {}
```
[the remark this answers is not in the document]
randombit: It works, which is sufficient, not an adversarial situation here

src/bogo_shim/bogo_shim.cpp
```diff
@@ -1024,6 +1114,39 @@ class Shim_Policy final : public Botan::TLS::Policy {
...
+               }
+            }
+            if(have_classical && have_pq) {
+               break;
+            }
+         }
+
+         if(to_offer.empty() && !groups.empty()) {
+            to_offer.push_back(groups.front());
+         }
```
reneme: This cannot ever happen, no? We first iterated all groups and will have either inserted it for being pq or not. But the "or not" is a simple else branch. Hence, I feel this could be an invariant check via assertion and abort. Or am I missing something?

src/lib/tls/tls_policy.h
```diff
@@ -547,6 +548,14 @@ class BOTAN_PUBLIC_API(2, 0) Policy /* NOLINT(*-special-member-functions) */ {
       */
       virtual uint64_t minimum_key_update_interval_ms() const;
 
+      /**
+      * @return the maximum number of consecutive KeyUpdate messages that
+      * can be received without intervening application data before the
+      * connection is terminated. Return 0 to disable the limit.
+      * @note Only applies to TLS 1.3 connections.
+      */
+      virtual size_t maximum_consecutive_key_updates() const;
```
reneme: Similar to the comments in #5522 (for instance: https://github.com/randombit/botan/pull/5522#discussion_r3050464175), please consider using a `std::optional<>` here for consistency and API clarity.

src/lib/tls/tls12/msg_client_hello_12.cpp
```diff
@@ -173,6 +173,13 @@ Client_Hello_12::Client_Hello_12(Handshake_IO& io,
       m_data->extensions().add(new SRTP_Protection_Profiles(policy.srtp_profiles()));
    }
 
+   // RFC 7250 4.1
+   //    In order to indicate the support of raw public keys, clients include
+   //    the client_certificate_type and/or the server_certificate_type
+   //    extensions in an extended client hello message.
+   m_data->extensions().add(new Client_Certificate_Type(policy.accepted_client_certificate_types()));
+   m_data->extensions().add(new Server_Certificate_Type(policy.accepted_server_certificate_types()));
```
reneme, line 181: But if the policy actually allows raw public keys, TLS 1.2 wouldn't be able to handle that... At least I didn't find a TLS 1.2 code path that would actually call `Callbacks::tls_verify_raw_public_key()`.

src/lib/tls/tls13/tls_channel_impl_13.cpp
```diff
@@ -169,6 +169,7 @@ size_t Channel_Impl_13::from_peer(std::span<const uint8_t> data) {
             if(!m_cipher_state->can_decrypt_application_traffic()) {
                throw Unexpected_Message("Application data received before handshake completion");
             }
+            m_key_update_count = 0;
```
reneme: I'm not exactly sure about the definition of this rate limit, but technically we might want to consider resetting this counter also on post-handshake messages other than KeyUpdate, i.e. solely on NewSessionTicket at the moment.

Likewise: once implemented, it might also make sense to enforce a similar rate limit on post-handshake authentication. Perhaps leave a TODO-comment about that? For instance in `Client_Impl_13::handle(const Certificate_Request_13&)`.

src/lib/tls/tls13/tls_cipher_state.h
```diff
@@ -62,7 +62,8 @@ class BOTAN_TEST_API Cipher_State {
    public:
       enum class PSK_Type : uint8_t {
          Resumption,
-         External,  // currently not implemented
+         External,
+         Imported,  // RFC 9258 PSK importer - uses "imp binder" label
       };
```
reneme, line 67: Nit: if you start adding the origins of those types, perhaps mention that the other two are from (I guess) RFC 8446.

src/lib/tls/tls13/tls_messages_13.h
```diff
@@ -146,7 +146,10 @@ class BOTAN_UNSTABLE_API Hello_Retry_Request final : public Server_Hello_13 {
 class BOTAN_UNSTABLE_API Encrypted_Extensions final : public Handshake_Message {
    public:
       explicit Encrypted_Extensions(const std::vector<uint8_t>& buf);
-      Encrypted_Extensions(const Client_Hello_13& client_hello, const Policy& policy, Callbacks& cb);
+      Encrypted_Extensions(const Client_Hello_13& client_hello,
+                           const Policy& policy,
+                           Callbacks& cb,
+                           bool is_resumption = false);
```
reneme: Nit: no reason to default the `is_resumption`. The only place we instantiate via this constructor is passing the value (in `tls_server_impl_13.cpp`). The API is public, though marked as unstable. If I were actually using this constructor in my application code, I'd probably want to know that this parameter was added, though.

src/lib/tls/tls_external_psk.cpp
```diff
@@ -18,4 +23,81 @@ secure_vector<uint8_t> ExternalPSK::extract_master_secret() {
...
+   const auto ctx_len = static_cast<uint16_t>(m_context.size());
+   std::vector<uint8_t> imported_identity(2 + m_identity.size() + 2 + m_context.size() + 2 + 2);
+   BufferStuffer id_stuffer(imported_identity);
+   store_be(id_stuffer.next<2>(), id_len);
+   id_stuffer.append(m_identity);
+   store_be(id_stuffer.next<2>(), ctx_len);
+   id_stuffer.append(m_context);
+   store_be(id_stuffer.next<2>(), target_protocol);
+   store_be(id_stuffer.next<2>(), target_kdf);
+   BOTAN_ASSERT_NOMSG(id_stuffer.full());
```
reneme: `concat` would be a better fit here (needs to include `botan/internal/concat_util.h` instead of `buffer_stuffer.h`).

```suggestion
   const auto imported_identity = concat<std::vector<uint8_t>>(
      store_be(id_len), m_identity, store_be(ctx_len), m_context, store_be(target_protocol), store_be(target_kdf));
```

src/lib/tls/tls_external_psk.cpp
```diff
@@ -18,4 +23,81 @@ secure_vector<uint8_t> ExternalPSK::extract_master_secret() {
...
+PSKImporter::PSKImporter(std::span<const uint8_t> key,
+                         std::span<const uint8_t> identity,
+                         std::span<const uint8_t> context,
+                         std::string_view hash) :
+      m_key(key.begin(), key.end()),
+      m_identity(identity.begin(), identity.end()),
+      m_context(context.begin(), context.end()),
+      m_hash(hash) {
+   BOTAN_ARG_CHECK(m_hash == "SHA-256" || m_hash == "SHA-384", "PSK importer hash must be SHA-256 or SHA-384");
+}
```
reneme, line 39:
```suggestion
PSKImporter::PSKImporter(std::span<const uint8_t> key,
                         std::span<const uint8_t> identity,
                         std::span<const uint8_t> context,
                         std::string_view hash) :
      m_key(key.begin(), key.end()),
      m_identity(identity.begin(), identity.end()),
      m_context(context.begin(), context.end()),
      m_hash(hash) {
   BOTAN_ARG_CHECK(m_hash == "SHA-256" || m_hash == "SHA-384", "PSK importer hash must be SHA-256 or SHA-384");
   BOTAN_ARG_CHECK(m_identity.size() <= std::numeric_limits<uint16_t>::max(), "PSK importer identity too long");
   BOTAN_ARG_CHECK(m_context.size() <= std::numeric_limits<uint16_t>::max(), "PSK importer context too long");
}
```

.... these values will be encoded as "length (16bit) + value" in the derivation function and we should check that the passed strings don't overflow those lengths. Note that this needs `#include <limits>`

src/lib/tls/tls_external_psk.cpp
```diff
@@ -18,4 +23,81 @@ secure_vector<uint8_t> ExternalPSK::extract_master_secret() {
...
+
+   // TLS 1.3 HkdfLabel: length (2) || label length (1) || label || context length (1) || context
+   std::vector<uint8_t> hkdf_label(2 + 1 + prefixed_label.size() + 1 + identity_hash.size());
+   BufferStuffer label_stuffer(hkdf_label);
+   store_be(label_stuffer.next<2>(), expand_out_len);
+   label_stuffer.next_byte() = static_cast<uint8_t>(prefixed_label.size());
+   label_stuffer.append(prefixed_label);
+   label_stuffer.next_byte() = static_cast<uint8_t>(identity_hash.size());
+   label_stuffer.append(identity_hash);
+   BOTAN_ASSERT_NOMSG(label_stuffer.full());
```
reneme: With `concat` you can avoid manually pre-allocating the input buffer because it does that behind the scenes.

```suggestion
   const auto prefixed_label_len = static_cast<uint8_t>(prefixed_label.size());
   const auto identity_hash_len = static_cast<uint8_t>(identity_hash.size());
   const auto hkdf_label = concat<std::vector<uint8_t>>(store_be(expand_out_len),
                                                        store_be(prefixed_label_len),
                                                        prefixed_label,
                                                        store_be(identity_hash_len),
                                                        identity_hash);
```

src/lib/tls/tls_external_psk.cpp
```diff
@@ -18,4 +23,81 @@ secure_vector<uint8_t> ExternalPSK::extract_master_secret() {
...
+   hkdf_extract->derive_key(epskx, m_key, salt, {});
+
+   // HKDF-Expand-Label(epskx, "derived psk", Hash(ImportedIdentity), L)
+   // The EPSK's hash drives Extract and Expand; target_hash sets output length only.
+   const std::string target_hash_str(target_hash);
+   auto target_hash_fn = HashFunction::create_or_throw(target_hash_str);
+   const size_t target_hash_len = target_hash_fn->output_length();
+   const auto expand_out_len = static_cast<uint16_t>(target_hash_len);
+
+   auto hkdf_expand = KDF::create_or_throw("HKDF-Expand(" + m_hash + ")");
```
reneme, line 81: Nit: `Cipher_State::hkdf_expand_label` provides an implementation of this very extraction which is currently private to the `Cipher_State` class but we could move it to some common place and re-use it here.

On the one hand, that would make it easier to understand the implementation (as it would read similar to the RFC). On the other hand, it could come in handy especially once we want to support DTLS which uses 'dtls13' instead of 'tls13 ' (trailing space is intentional) as the label prefix. See: https://www.rfc-editor.org/rfc/rfc9147.html#name-cryptographic-label-prefix

_... maybe at least leave a TODO comment so that I can do the refactoring eventually as part of #5000._

src/lib/tls/tls13/tls_cipher_state.cpp
```diff
@@ -480,7 +480,18 @@ void Cipher_State::advance_with_psk(PSK_Type type, secure_vector<uint8_t>&& psk)
...
-   const char* binder_label = (type == PSK_Type::Resumption) ? "res binder" : "ext binder";
+   // RFC 8446 and RFC 9258 specify these strings
+   const char* binder_label = [type]() -> const char* {
+      switch(type) {
+         case PSK_Type::Resumption:
+            return "res binder";
+         case PSK_Type::External:
+            return "ext binder";
+         case PSK_Type::Imported:
+            return "imp binder";
```
reneme, line 491: Nit: See the state machine on the top of this file: Might be worth to extend it in line 17, noting that  "imp binder" is an option.

src/lib/tls/tls_external_psk.h
```diff
@@ -51,10 +55,55 @@ class BOTAN_PUBLIC_API(3, 2) ExternalPSK {
...
+};
+
+/**
+ * RFC 9258 PSK Importer.
+ *
+ * Holds the base key material and identity for a pre-shared key and
+ * derives imported PSKs for specific TLS protocol versions and cipher
+ * suite hash algorithms using the PSK importer mechanism
+ */
+class BOTAN_PUBLIC_API(3, 12) PSKImporter {
```
reneme, line 80: RFC 9258 explicitly defines this importer for use in TLS 1.3 only. Technically, it should be implemented in the TLS 1.3 sub-module for that reason. Somewhat nitpicky, maybe still worth considering to create a dedicated header/cpp for it in the tls13 module.

src/lib/tls/tls_external_psk.cpp
```diff
@@ -18,4 +23,81 @@ secure_vector<uint8_t> ExternalPSK::extract_master_secret() {
...
+   auto hash_fn = HashFunction::create_or_throw(m_hash);
+   hash_fn->update(imported_identity);
+   const auto identity_hash = hash_fn->final_stdvec();
+
+   // HKDF-Extract(0, epsk) -- using the EPSK's hash per above
+   const size_t psk_hash_len = hash_fn->output_length();
+   auto hkdf_extract = KDF::create_or_throw("HKDF-Extract(" + m_hash + ")");
+   secure_vector<uint8_t> epskx(psk_hash_len);
+   std::vector<uint8_t> salt(psk_hash_len, 0);
+   hkdf_extract->derive_key(epskx, m_key, salt, {});
```
reneme:
```suggestion
   const std::vector<uint8_t> salt(psk_hash_len, 0);
   const auto epskx = hkdf_extract->derive_key(psk_hash_len, m_key, salt, {});
```

... easier to reason about, because the "output value" is clearly visible (and `const`).

src/lib/tls/tls_external_psk.h
```diff
@@ -51,10 +55,55 @@ class BOTAN_PUBLIC_API(3, 2) ExternalPSK {
...
+      /**
+       * @param key the base pre-shared key
+       * @param identity the external PSK identity
+       * @param context optional importer context
+       * @param hash the hash algorithm provisioned with this PSK ("SHA-256" or "SHA-384")
+       */
+      PSKImporter(std::span<const uint8_t> key,
+                  std::span<const uint8_t> identity,
+                  std::span<const uint8_t> context,
+                  std::string_view hash);
```
reneme: RFC 9258 Section 5.1 says: "If the EPSK does not have [...] an associated hash function, SHA-256 [[SHA2](https://www.rfc-editor.org/rfc/rfc9258.html#SHA2)] SHOULD be used."

Nit: 


```suggestion
                  std::string_view hash = "SHA-256");
```

src/lib/tls/tls_external_psk.h
```diff
@@ -51,10 +55,55 @@ class BOTAN_PUBLIC_API(3, 2) ExternalPSK {
        */
       const std::string& prf_algo() const { return m_prf_algo; }
 
+      /**
+       * Returns true if this PSK was derived using the PSK importer
+       * mechanism from RFC 9258. Imported PSKs use the "imp binder"
+       * label for binder computation instead of "ext binder".
+       */
+      bool is_imported() const { return m_is_imported; }
```
reneme, line 64: Nit: If this class was implemented within the TLS 1.3 submodule, it might be worthwhile to let this return `Cipher_State::PSK_Type` instead. In the production code we really just use that as input to a ternary to distinguish between `PSK_Type::Imported` and `PSK_Type::External` anyway.

src/lib/tls/info.txt
```diff
@@ -52,6 +52,7 @@ dh
 ecdh
 ecdsa
 gcm
+hkdf
```
reneme, line 55: Moving the new Importer class to the TLS 1.3 submodule would remove the need for this.

src/lib/tls/tls_algos.cpp
```diff
@@ -133,43 +136,122 @@ Auth_Method auth_method_from_string(std::string_view str) {
...
+   // For group codes we recognize, check the build-time availability table.
+   // Unknown codes may be user-supplied custom groups handled via callbacks.
+   if(to_string().has_value()) {
+      static constexpr auto groups = available_group_params();
+      const auto* begin = groups.first.data();
+      const auto* end = begin + groups.second;
+      return std::binary_search(begin, end, wire_code());
+   }
    return true;
 }
```
reneme, line 242: The compiler shall sweat even harder. Also: no way we f*ck up the lengths and offsets like that.

```suggestion
namespace {

consteval auto available_group_params() {
   auto codes = std::array{
#if defined(BOTAN_HAS_PCURVES_SECP256R1) || defined(BOTAN_HAS_PCURVES_GENERIC)
      Group_Params_Code::SECP256R1,
#endif

#if defined(BOTAN_HAS_PCURVES_SECP384R1) || defined(BOTAN_HAS_PCURVES_GENERIC)
      Group_Params_Code::SECP384R1,
#endif

#if defined(BOTAN_HAS_PCURVES_SECP521R1) || defined(BOTAN_HAS_PCURVES_GENERIC)
      Group_Params_Code::SECP521R1,
#endif

#if defined(BOTAN_HAS_PCURVES_BRAINPOOL256R1) || defined(BOTAN_HAS_PCURVES_GENERIC)
      Group_Params_Code::BRAINPOOL256R1,
#endif

#if defined(BOTAN_HAS_PCURVES_BRAINPOOL384R1) || defined(BOTAN_HAS_PCURVES_GENERIC)
      Group_Params_Code::BRAINPOOL384R1,
#endif

#if defined(BOTAN_HAS_PCURVES_BRAINPOOL512R1) || defined(BOTAN_HAS_PCURVES_GENERIC)
      Group_Params_Code::BRAINPOOL512R1,
#endif

#if defined(BOTAN_HAS_X25519)
      Group_Params_Code::X25519,
#endif

#if defined(BOTAN_HAS_X448)
      Group_Params_Code::X448,
#endif

#if defined(BOTAN_HAS_DIFFIE_HELLMAN)
      Group_Params_Code::FFDHE_2048,
      Group_Params_Code::FFDHE_3072,
      Group_Params_Code::FFDHE_4096,
      Group_Params_Code::FFDHE_6144,
      Group_Params_Code::FFDHE_8192,
#endif

#if defined(BOTAN_HAS_ML_KEM)
      Group_Params_Code::ML_KEM_512,
      Group_Params_Code::ML_KEM_768,
      Group_Params_Code::ML_KEM_1024,

   #if defined(BOTAN_HAS_PCURVES_SECP256R1) || defined(BOTAN_HAS_PCURVES_GENERIC)
      Group_Params_Code::HYBRID_SECP256R1_ML_KEM_768,
   #endif

   #if defined(BOTAN_HAS_PCURVES_SECP384R1) || defined(BOTAN_HAS_PCURVES_GENERIC)
      Group_Params_Code::HYBRID_SECP384R1_ML_KEM_1024,
   #endif

   #if defined(BOTAN_HAS_X25519)
      Group_Params_Code::HYBRID_X25519_ML_KEM_768,
   #endif
#endif

#if defined(BOTAN_HAS_FRODOKEM)
      Group_Params_Code::eFRODOKEM_640_SHAKE_OQS,
      Group_Params_Code::eFRODOKEM_976_SHAKE_OQS,
      Group_Params_Code::eFRODOKEM_1344_SHAKE_OQS,
      Group_Params_Code::eFRODOKEM_640_AES_OQS,
      Group_Params_Code::eFRODOKEM_976_AES_OQS,
      Group_Params_Code::eFRODOKEM_1344_AES_OQS,

   #if defined(BOTAN_HAS_PCURVES_SECP256R1) || defined(BOTAN_HAS_PCURVES_GENERIC)
      Group_Params_Code::HYBRID_SECP256R1_eFRODOKEM_640_SHAKE_OQS,
      Group_Params_Code::HYBRID_SECP256R1_eFRODOKEM_640_AES_OQS,
   #endif

   #if defined(BOTAN_HAS_PCURVES_SECP384R1) || defined(BOTAN_HAS_PCURVES_GENERIC)
      Group_Params_Code::HYBRID_SECP384R1_eFRODOKEM_976_SHAKE_OQS,
      Group_Params_Code::HYBRID_SECP384R1_eFRODOKEM_976_AES_OQS,
   #endif

   #if defined(BOTAN_HAS_PCURVES_SECP521R1) || defined(BOTAN_HAS_PCURVES_GENERIC)
      Group_Params_Code::HYBRID_SECP521R1_eFRODOKEM_1344_SHAKE_OQS,
      Group_Params_Code::HYBRID_SECP521R1_eFRODOKEM_1344_AES_OQS,
   #endif

   #if defined(BOTAN_HAS_X25519)
      Group_Params_Code::HYBRID_X25519_eFRODOKEM_640_SHAKE_OQS,
      Group_Params_Code::HYBRID_X25519_eFRODOKEM_640_AES_OQS,
   #endif

   #if defined(BOTAN_HAS_X448)
      Group_Params_Code::HYBRID_X448_eFRODOKEM_976_SHAKE_OQS,
      Group_Params_Code::HYBRID_X448_eFRODOKEM_976_AES_OQS,
   #endif
#endif
   };

   std::sort(codes.begin(), codes.end());

   return codes;
}

}  // namespace

bool Group_Params::is_available() const {
   // For group codes we recognize, check the build-time availability table.
   // Unknown codes may be user-supplied custom groups handled via callbacks.
   if(to_string().has_value()) {
      static constexpr auto groups = available_group_params();
      return std::binary_search(groups.begin(), groups.end(), code());
   }
   return true;
}
```

reneme (commented):
Implementations look semantically correct to me, except for the fact that our TLS 1.2 doesn't actually support using raw public keys but the Hello-extensions now might negotiate it when the TLS policy is set up accordingly. This would then result in a decoding error (I guess) later down in the handshake.

Also left some suggestions for code improvements and a few nits.

src/lib/tls/tls_algos.cpp
randombit: Nice

src/lib/tls/tls12/msg_client_hello_12.cpp
randombit: Hm so we have a bit of a problem here already - since we can always send a 1.3 hello with raw public key support indicated, and the then server replies with a 1.2 server hello with a raw public key.

src/lib/tls/tls13/tls_channel_impl_13.cpp
randombit: TBH I'm thinking about reverting this counter entirely - as a DoS countermeasure it's completely ineffective and the only reason I did it was to enable KeyUpdate throttling that was testable by BoGo.

src/bogo_shim/bogo_shim.cpp
randombit: Yeah can't happen, removed

src/lib/tls/tls_policy.h
randombit: I think I'm going to revert this anyway but to me size_t with an implicit 0 disables makes sense anyway considering that if you actually enforced 0 consecutive key updates, you could not complete a handshake, so 0 is already your distinguished value. Kind of the same way that in Rust `Option<NonZero<usize>>` is the same as a `usize`. Otherwise you'd have explicitly handle the non-sensensical `Some(0)`

randombit: Reverted

src/lib/tls/tls13/tls_channel_impl_13.cpp
randombit: Reverted

src/lib/tls/tls13/tls_messages_13.h
randombit: Removed

src/lib/tls/tls13/tls_cipher_state.cpp
randombit: done

src/lib/tls/tls_external_psk.h
randombit: TBH I wanted to avoid it because already have a lot of public TLS headers

```
tls_alert.h
tls_algos.h
tls_callbacks.h
tls_channel.h
tls_ciphersuite.h
tls_client.h
tls_exceptn.h
tls_extensions_12.h
tls_extensions_13.h
tls_extensions.h
tls_external_psk.h
tls.h
tls_handshake_msg.h
tls_magic.h
tls_messages_12.h
tls_messages_13.h
tls_messages.h
tls_policy.h
tls_psk_identity_13.h
tls_server.h
tls_server_info.h
tls_session.h
tls_session_id.h
tls_session_manager.h
tls_session_manager_hybrid.h
tls_session_manager_memory.h
tls_session_manager_noop.h
tls_session_manager_sql.h
tls_session_manager_sqlite.h
tls_session_manager_stateless.h
tls_signature_scheme.h
tls_version.h
```

And on the one hand what's one more at this point ... but also damn.

Also there is no reason in principal that PSK import could not be defined for TLS 1.2 either by IETF (unlikely I guess considering they are not even willing to add standardize PQ support to 1.2) or as a local extension, the same protocol as is would even work considering it already takes the protocol version as a derivation input.

src/lib/tls/tls_external_psk.cpp
randombit: done

src/lib/tls/tls_external_psk.cpp
randombit: This PR is already pretty big with many commits somewhat carefully structured to not conflict with each other too much so I am not going to try any refactorings here. I left a `TODO(DTLS1.3)` comment

src/lib/tls/tls_external_psk.cpp
randombit: done

src/bogo_shim/bogo_shim.cpp
```diff
@@ -653,9 +672,67 @@ class Shim_Arguments final {
...
+   bool building = false;
+
+   auto flush = [&]() {
+      if(building) {
+         creds.emplace_back(std::move(key), std::move(identity), std::move(context), std::move(hash));
+         key.clear();
+         identity.clear();
+         context.clear();
+         hash.clear();
+         building = false;
```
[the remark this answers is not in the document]
randombit: fixed

src/lib/tls/tls_algos.cpp
```diff
@@ -133,43 +136,108 @@ Auth_Method auth_method_from_string(std::string_view str) {
...
+         Group_Params_Code::HYBRID_X25519_eFRODOKEM_640_AES_OQS,
+   #endif
+
+   #if defined(BOTAN_HAS_X448)
+         Group_Params_Code::HYBRID_X448_eFRODOKEM_976_SHAKE_OQS, Group_Params_Code::HYBRID_X448_eFRODOKEM_976_AES_OQS,
+   #endif
+#endif
+   };
+
+   std::sort(codes.begin(), codes.end());
```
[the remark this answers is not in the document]
randombit: > where std::sort is not constexpr

Why do you insist on believing this

src/lib/tls/tls12/msg_client_hello_12.cpp
reneme: True that. 😨

reneme: Anyway, a credible short-term fix would be to send the extension in TLS 1.3 only if TLS 1.2 is explicitly disabled, like we do it for the record size limit extension that is also available in TLS 1.3 only:

https://github.com/randombit/botan/blob/ea96b69d31d9420ad9761e232a07a52427722b01/src/lib/tls/tls13/msg_client_hello_13.cpp#L223-L228

src/lib/tls/tls_external_psk.h
reneme: > Also there is no reason in principal that PSK import could not be defined for TLS 1.2 either by IETF

Well, yeah, in theory... But this construction uses the KDFs defined for TLS 1.3 (e.g. HKDF-Expand-Label, which literally uses "tls13 " as a domain separator) and it also hooks fairly deeply into the TLS 1.3 key schedule (with the "imp binder" label). I honestly doubt its ever going to be useful for TLS 1.2.

Suggestion: What about introducing a new `tls_psk_13.h` header for it and moving the content of `tls_psk_identity_13.h` in there as well. The latter could then be just a deprecated delegate header until Botan4 comes around.

The reason I'm not letting go of this is because the mentioned refactoring opportunity for DTLS 1.3 (the reuse of the KDF implementation) would make much more sense if this would live in TLS 1.3. Because then we could also make the TLS KDFs live solely within the tls13 submodule.

src/lib/tls/tls_policy.h
```diff
@@ -535,6 +535,7 @@ class BOTAN_PUBLIC_API(2, 0) Policy /* NOLINT(*-special-member-functions) */ {
       /**
       * @return the maximum size of the certificate chain, in bytes.
       * Return 0 to disable this and accept any size.
+      * TODO(Botan4) rename this maximum_certificate_chain_bytes
       */
       virtual size_t maximum_certificate_chain_size() const;
```
reneme, line 540: As a consumer I would probably be grateful for a deprecation warning pre-Botan4. Perhaps introduce the new function now and make this deprecated and delegate to the new one?

Admittedly, that is a nit. Especially given that usage of this functions is probably pretty nieche.... More of a general question about our MO.

src/tests/data/tls_13_rfc8448/transcripts.vec
```diff
@@ -34,8 +34,8 @@ Client_EarlyAppData = 414243444546
...
 Record_ServerHello = 16030300600200005c03033ccfd2dec890222763472ae8136777c9d7358777bb66e91ea5122495f559ea2d00130100003400290002000000330024001d0020121761ee42c333e1b9e77b60dd57c2053cd94512ab47f115e86eff50942cea31002b00020304
 Message_ServerHello = 0200005c03033ccfd2dec890222763472ae8136777c9d7358777bb66e91ea5122495f559ea2d00130100003400290002000000330024001d0020121761ee42c333e1b9e77b60dd57c2053cd94512ab47f115e86eff50942cea31002b00020304
-Record_ServerHandshakeMessages = 1703030061dc48237b4b879f50d0d4d262ea8b4716eb40ddc1eb957e11126e8a7149c2d012d37a7115957e64ce30008b9e0323f2c05a9c1c77b4f37849a695ab255060a33fee770ca95cb8486bfd0843b87024865ca35cc41c4e515c64dcb1369f98635bc7a5
-Message_EncryptedExtensions = 080000280026000a00140012001d00170018001901000101010201030104001c0002400100000000002a0000
+Record_ServerHandshakeMessages = 170303005ddc4823774b839f50d0d4d262ea8b4716eb40ddc1eb957e11126e8a7149c2d012d37a7115955464ce242a8bbec05967ce9127b806c07218590d06b66ff0aa95b56562f404c410b702dda9a4a4bb3b9cc44fe20e2675e63020d66b936120
+Message_EncryptedExtensions = 080000240022000a00140012001d00170018001901000101010201030104001c00024001002a0000
```
reneme, line 38: Please leave a comment that this deviates from RFC8448 for "reasons". Otherwise one of future-us will undoubtedly waste a day of work on this at one point. :)

src/lib/tls/tls12/msg_client_hello_12.cpp
randombit: Yeah that works for now

src/lib/tls/tls_external_psk.h
randombit: All right that works

src/lib/tls/tls_policy.h
randombit: In general I'm a big fan of making it easy to support the same code running on both versions across a major version bump (eg same code can work on latest 3.x and 4.0), and IMO it's generally a mistake to skip this step. For example Nettle 4.0 changed how all hash function APIs work with no backwards compat and also not supporting the new interfaces in the old Nettle 3, which means that now all applications must upgrade to Nettle 4 in lockstep across an entire distro, which is probably why even Arch is still shipping Nettle 3. It might take years before all relevant applications add a Nettle 4 path - especially when nobody will notice their code is broken because their distro, as with everyone else's, still uses Nettle 3.

Here (and for TLS policies in general) it's kind of complicated because we explicitly allow applications to override, so we have to handle 1) application implemented old policy hook 2) application implemented new policy hook 3) application implemented both 4) application implemented both and the values returned are inconsistent. Might be ok here though I'll try it.

src/lib/tls/tls12/msg_client_hello_12.cpp
randombit: BTW I was curious what exactly BoGo was testing here (it has a bunch of RPK tests that passed even in TLS 1.2): turns out it just indicates support in the hello but doesn't actually test the full RPK handshake.

randombit: Now restricted to 1.3-only negitiations

src/lib/tls/info.txt
randombit: Moved to `tls13`, dep removed

src/lib/tls/tls_external_psk.h
randombit: Would require pulling `PSK_Type` out of `Cipher_State` since that's an internal header, so passing on this here.

src/lib/tls/tls_policy.h
randombit: I reverted the TODO for now, not going to do this in this PR

src/tests/data/tls_13_rfc8448/transcripts.vec
randombit: done

src/bogo_shim/bogo_shim.cpp
```diff
@@ -37,6 +37,10 @@
 #include <botan/internal/stl_util.h>
 #include <botan/internal/target_info.h>
 
+#if defined(BOTAN_HAS_TLS_13)
+   #include <botan/tls_psk_13.h>
+#endif
```
reneme, line 42: The usage of the `PSKImporter` also has to be disabled when TLS 1.3 isn't available. That's the failing CI job.

reneme (approved):
LGTM, apart from the relevant CI failure.

PR #5528, opened 2026-04-14
Feature: More configuration options for installation directory structure

reneme:
This provides `./configure.py --without-include-namespace` to let botan install headers straight into "\<prefix>/include" instead of using "\<prefix>/include/botan-3". That's required to fix the vcpkg build, see here: https://github.com/microsoft/vcpkg/issues/50177

Note that this also adds additional sanity checks:

* libdir, bindir and includedir are now all opportunistically prefixed with 'prefix', if they weren't provided as absolute paths to begin with
* libdir, bindir and includedir are validated to be subpaths of prefix

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -2236,7 +2236,7 @@ def create_template_vars(source_paths, build_paths, options, modules, disabled_m
         'bindir': absolute_install_dir(options.bindir or osinfo.bin_dir),
         'libdir': absolute_install_dir(options.libdir or osinfo.lib_dir),
         'mandir': options.mandir or osinfo.man_dir,
-        'includedir': options.includedir or osinfo.header_dir,
+        'includedir': absolute_install_dir(options.includedir or osinfo.header_dir),
         'docdir': options.docdir or osinfo.doc_dir,
 
         'with_documentation': options.with_documentation,
@@ -2376,18 +2376,32 @@ def create_template_vars(source_paths, build_paths, options, modules, disabled_m
         'disabled_mod_list': sorted([m.basename for m in disabled_modules]),
     }
 
+    if not os.path.isabs(variables['prefix']):
+        raise UserError("The installation root must be an absolute path")
+
+    if not is_subpath(variables['libdir'], variables['prefix']):
+        raise UserError("The libdir must be a subdirectory of the prefix")
+
+    if not is_subpath(variables['includedir'], variables['prefix']):
+        raise UserError("The includedir must be a subdirectory of the prefix")
+
     variables['namespaced_includedir'] = os.path.join(
         variables['includedir'],
         ('botan-%d' % Version.major()) if options.with_include_namespace else '')
     variables['installed_include_dir'] = os.path.join(
-        variables['prefix'],
         variables['namespaced_includedir'],
         'botan')
 
+    # A long time ago some packages required a bindir that was outside the installation
+    # prefix. In the CMake config we need the bindir to find DLLs on Windows. If the
+    # bindir is configured to be outside the prefix, CMake will fall back to a hard-coded
+    # path instead of a relative path for relocatability.
+    if is_subpath(variables['bindir'], variables['prefix']):
+        variables['bindir_rel'] = normalize_source_path(os.path.relpath(variables['bindir'], variables['prefix']))
+
     variables['libdir_rel'] = normalize_source_path(os.path.relpath(variables['libdir'], variables['prefix']))
-    variables['bindir_rel'] = normalize_source_path(os.path.relpath(variables['bindir'], variables['prefix']))
-    variables['includedir_rel'] = normalize_source_path(os.path.relpath(absolute_install_dir(variables['includedir']), variables['prefix']))
-    variables['namespaced_includedir_rel'] = normalize_source_path(os.path.relpath(absolute_install_dir(variables['namespaced_includedir']), variables['prefix']))
+    variables['includedir_rel'] = normalize_source_path(os.path.relpath(variables['includedir'], variables['prefix']))
+    variables['namespaced_includedir_rel'] = normalize_source_path(os.path.relpath(variables['namespaced_includedir'], variables['prefix']))
 
     # On MSVC, the "ABI flags" should be passed to the compiler only, on other platforms, the
     # ABI flags are passed to both the compiler and the linker and the compiler flags are also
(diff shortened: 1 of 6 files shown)


randombit:
Change seems fine but tbh I'm not understanding the issue here and I would have thought the problem would lay more with either the CMake output we produce, or a bug in vcpkg itself. I guess if it's a vgpkg issue maybe they just aren't going to fix it so we need to do this as a workaround in which case fine. But feel like I'm missing some context here. I read the linked ticket and everything about it seems like "vcpkg is holding it wrong"

reneme:
> I read the linked ticket and everything about it seems like "vcpkg is holding it wrong"

Yeah, pretty much. _Our CMake config generator assumes that distributors don't change the structure of the install directory_. Most notably the relative location of the include and lib files as seen from the botan-config.cmake are hard-coded into the generated CMake config file.

Right now the vcpkg recipe moves around both the installed headers and the botan-cmake.config file within the installed directory structure. That obviously breaks things.

For customizing the installation location of the CMake config files, we already provide `./configure.py --cmakeconfigdir=`.

I don't know why they move the headers out of the "botan-3" namespace folder, but I simply assumed that this is done for a valid reason. (Perhaps @justusranvier or @Chaosvex can provide insights here?) Anyway, removing the "botan-3" namespace folder without breaking the CMake config file generation would be facilitated by the `--without-include-namespace` switch introduced here.

Layering the effects of these configure switches within the vcpkg recipe resulted in a functional package in my local tests.

justusranvier:
I do not know why vcpkg did their packaging that way but I know that in order to use Botan in various scenarios (built by vcpkg, manually compiled, installed by linux distribution as a system package) I have do to this:

```
#if __has_include(<botan/hash.h>)
#include <botan/hash.h>  // IWYU pragma: export
#elif __has_include(<botan-3/hash.h>)
#include <botan-3/hash.h>  // IWYU pragma: export
#else
#error "missing <botan/hash.h>"
#endif
#if __has_include(<botan/mac.h>)
#include <botan/mac.h>  // IWYU pragma: export
#elif __has_include(<botan-3/mac.h>)
#include <botan-3/mac.h>  // IWYU pragma: export
#else
#error "missing <botan/mac.h>"
#endif

...etc
```

My best guess about how that came to pass is that some packagers have a policy of not including the version number in the include path.

Chaosvex:
I thought vcpkg added that to work around an older issue with Botan3's CMake but I could be mistaken.

reneme:
> I thought vcpkg added that to work around an older issue with Botan3's CMake but I could be mistaken.

I started digging in the recipe's commit history and the rename from "include/botan-2/botan/\*.h" to "include/botan/\*.h" was always there all the way back when the recipe was originally introduced to vcpkg: https://github.com/microsoft/vcpkg/pull/773.

In fact, a (presumable) maintainer added it during code review stating:

> I made a few changes while merging:
>
> [...]
>
> * The headers were installed to include\botan-2\botan. All of the Botan examples online use #include <botan/XYZ.h>, so I reorganized them to work out of the box.
>
> [...]

To me it seems that vcpkg just isn't flexible when it comes to these "namespaced sub-directories". In fact, [their documentation about the packaged installation tree layout](https://learn.microsoft.com/en-us/vcpkg/reference/installation-tree-layout#layout-include) states (regarding the "include" sub-directory):

> Contains header files (.h, .hpp, .hxx). The layout under this directory should reflect the intended usage of the package's header files. For example, a contoso library that intends to use #include <contoso/contoso.h> should provide the header file include/contoso/contoso.h.

---

FWIW: In Conan I can just define the package's include path to contain the "botan-3" sub-directory [in the `def package_info()`](https://github.com/conan-io/conan-center-index/blob/93c1228cd7cfd98825e04db9be04dbb3fcea815f/recipes/botan/all/conanfile.py#L278) and be done with it.

---

@randombit TL;DR: We need the new `--without-include-namespace` because vcpkg, in contrast to Conan for instance, is inflexible about their packages' include directory structure. It has to reflect the intended include pattern exactly (e.g. `#include <botan/hex.h>`). And since re-arranging the  final installed headers is iffy, we have to have a way to let the package recipe customize that.

doc/building.rst
```diff
@@ -1105,6 +1105,13 @@ Set the man page installation dir.
 
 Set the include file installation dir.
 
+``--without-include-namespace``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+By default, the header files (e.g. ``botan/hex.h``) are installed into an
+additional subdirectory named ``botan-<version>``. This option causes them to be
+installed directly into ``<includedir>`` instead.
```
randombit, line 1113: I'd want to discourage this with eg

> This option is not needed for normal usage and is only required in order to work around limitations in certain package managers.

configure.py
```diff
@@ -2375,10 +2376,29 @@ def test_exe_extra_ldflags():
...
-    variables['installed_include_dir'] = os.path.join(
-        variables['prefix'],
+    if not os.path.isabs(variables['prefix']):
+        raise UserError("The installation root must be an absolute path")
+
+    if not is_subpath(variables['libdir'], variables['prefix']):
+        raise UserError("The libdir must be a subdirectory of the prefix")
+
+    if not is_subpath(variables['bindir'], variables['prefix']):
+        raise UserError("The bindir must be a subdirectory of the prefix")
```
randombit: Is this strictly required by this change or just an extra sanity check? I ask because ISTR (a long time ago) some packagers used a bindir that was not in prefix. Maybe not an issue anymore in any event, but we may have to loosen this.

randombit (approved):
One question and one add to the docs otherwise ok

configure.py
reneme: Just an extra sanity check that seemed to make sense when introducing the explicit restrictions on `libdir` and `includedir`. I'll remove if for `bindir`.

reneme: Wait. `bindir_rel` was introduced for the CMake config file and it assumes that `bindir` lives within `prefix`. The CMake config needs the bindir on Windows, because that's where the DLL goes, AFAICT.

Let me look into this over the weekend.

reneme: @randombit I removed the check that bindir must be a subdirectory of prefix. But if it happens to be a subdirectory, the CMake config is aware of it and doesn't hard-code the entire path. I hope that's flexible enough.

No more work to be done here, please merge if you're okay with this change.

reneme:
The additions are now proposed for integration in vcpkg's recipe, for reference:

* https://github.com/microsoft/vcpkg/pull/51600

PR #5529, opened 2026-04-14
Enable KEM certificate issuance via CLI

### LINKED ISSUE #4516: Question: how to create a certificate for ML-KEM pubkey?
### Background
Using OpenSSL-3.4.0 with OQS provider (from Open Quantum Safe), I 

1. generated ML-DSA and ML-KEM key-pairs, 
2. created from ML-DSA key-pair an ML-DSA self-signed certificate to act as CA, and 
3. created an ML-KEM certificate from ML-KEM pubkey, signed by the above CA.  

Botan-3.7.0 parsed the above certs and validated the ML-KEM cert.

### Goal
I want to replicate the above (1)-(3) with Botan, aka generate keypairs and create corresponding certificates.

### Current progress
1. Generated key-pairs:
```text
$ botan3 keygen --algo="ML-KEM" --params="ML-KEM-1024" > prkey-kem.pem
$ botan3 keygen --algo="ML-DSA" --params="ML-DSA-87" > prkey-dsa.pem
$ botan3 keygen --algo="ML-DSA" --params="ML-DSA-8x7" > prkey-dsa.pem
$ botan3 pkcs8 --pub-out prkey-kem.pem > pubkey-kem.pem
$ botan3 pkcs8 --pub-out prkey-dsa.pem > pubkey-dsa.pem
$
```
2. Created CA (self-signed cert) from ML-DSA key-pair:
```text
$ botan3 gen_self_signed prkey-dsa.pem "/CN=ML-DSA-CA" --ca > dsa-ca.pem
$
```

### Problem
Cannot figure out how to create a certificate signed by the above CA for the ML-KEM public key. Would appreciate help.

falko-strenzke:
Fixes #4516.

#  TODO
 - [x] src/lib/x509/x509self.cpp: create_cert_req()
   - [x] adds default extensions
 - [x] src/lib/x509/x509_ca.cpp: X509_CA::choose_extensions():
   - [x] add or update basic constraints based on req.is_CA()
   - [x] add / replace further extensions
 - [x] src/lib/x509/x509self.cpp: create_alt_name_ext()
   - [x] set SAN value from opts

### DIFF
--- a/src/cli/x509.cpp
+++ b/src/cli/x509.cpp
@@ -5,7 +5,9 @@
 * Botan is released under the Simplified BSD License (see license.txt)
 */
 
+#include "botan/x509_key.h"
 #include "cli.h"
+#include <iostream>
 
 #if defined(BOTAN_HAS_X509_CERTIFICATES) && defined(BOTAN_TARGET_OS_HAS_FILESYSTEM)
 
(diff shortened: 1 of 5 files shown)


src/cli/x509.cpp
```diff
@@ -381,6 +382,100 @@ class Generate_PKCS10 final : public Command {
...
+         opts.organization = get_arg("organization");
+         opts.email = get_arg("email");
+         opts.more_dns = Command::split_on(get_arg("dns"), ',');
+
+         for(const std::string& ext_ku : Command::split_on(get_arg("ext-ku"), ',')) {
+            opts.add_ex_constraint(ext_ku);
+         }
+
+         // TODO(Botan4) remove --emsa option and this logic
+         const std::string padding = [&]() {
```
randombit, line 417: Seems unnecessary to have any compatibility considerations here

src/cli/x509.cpp
```diff
@@ -381,6 +382,100 @@ class Generate_PKCS10 final : public Command {
...
+         return "Generate a certificate by a CA for a given public key. This is for instance needed to issue ML-KEM certificates, for which PKCS#10 requests cannot be generated.";
+      }
+
+      void go() override {
+         const std::string ca_key_file = get_arg("ca_key");
+         const std::string passphrase = get_passphrase_arg("Passphrase for CA key " + ca_key_file, "ca-key-pass");
+
+         auto key = load_private_key(ca_key_file, passphrase);
+
+         Botan::X509_Cert_Options opts;
```
randombit, line 404: This shouldn't be used here, just write the subject DN directly

randombit (changes requested):
Basic idea is fine, could be a lot simpler in implementation see prev comments

mouse07410:
@falko-strenzke as some of us eagerly await on this PR getting merged, might I ask you to address @randombit 's comments, or discuss with him why not...?

src/cli/x509.cpp
falko-strenzke: What does it mean "write subject DN directly"? Is there an example for this somewhere in the code?

The code I propose is following the PKCS#10 request generation code. Naively, I would assume that should be right.

src/cli/x509.cpp
falko-strenzke: That means we drop "emsa"?

randombit: Exactly so, this is only handled in the other cmdlets because that was the only way to specify it, doesn't seem necessary to add it for new cmdlets

src/cli/x509.cpp
randombit: I mean you pretty much are doing exactly that below - just with an intermediate step of placing the values into a `X509_Cert_Options` struct, which seems unnecessary.

falko-strenzke:
Now I realize there are quite a few things still missing, I noted the following:

         // - [ ] src/lib/x509/x509self.cpp: create_cert_req()
         //   - [ ] adds default extensions
         // - [ ] src/lib/x509/x509_ca.cpp: X509_CA::choose_extensions():
         //   - [ ] add or update basic constraints based on req.is_CA()
         //   - [ ] add / replace further extensions
         // - [ ] src/lib/x509/x509self.cpp: create_alt_name_ext()
         //   - [ ] set SAN value from opts

This might require some refactoring in order not to duplicate too much code. And the list might not be complete. I will have to postpone this for now.

src/cli/x509.cpp
falko-strenzke: Done

src/cli/x509.cpp
falko-strenzke: Done

falko-strenzke:
@randombit 
From my side the PR is complete now. Let's see if the CI runs through.

mouse07410:
@falko-strenzke it would be great if you added examples to this PR, and maybe edits to the Botan manual, to make it easier for the user of this capability to figure out how exactly to generate certs for ML-KEM pubkeys.

mouse07410:
@randombit I think both change requests have been done? (Unless I'm missing something?)

PR #5532, opened 2026-04-15
add parsing of ECC group from ECC parameters

falko-strenzke:
Tasks:

- [x] integrate new test vectors into new test in the `test_ecdh.cpp`


randombit:
Change looks fine, needs some test cases though

src/tests/test_pubkey.cpp
```diff
@@ -5,6 +5,7 @@
 * Botan is released under the Simplified BSD License (see license.txt)
 */
 
+#include "botan/asn1_obj.h"
```
randombit: Includes of library headers should be `#include <botan/foo.h>`

src/tests/test_pubkey.cpp
```diff
@@ -1037,6 +1041,73 @@ BOTAN_REGISTER_TEST("pubkey", "pk_key_decoding", PK_Key_Decoding_Test);
 
 }  // namespace
 
+/**
+ * @brief Test ECC private key decoding
+ */
+class EC_Private_Key_Decoding_Test : public Test {
+   public:
+      //EC_Private_Key_Decoding_Test() {}
```
randombit: Dead?

src/tests/test_pubkey.cpp
```diff
@@ -21,6 +22,9 @@
    #include <botan/x509_key.h>
    #include <botan/internal/fmt.h>
    #include <botan/internal/stl_util.h>
+   #if defined(BOTAN_HAS_ECC_PUBLIC_KEY_CRYPTO)
```
randombit, line 25: This is not an appropriate file for this test - `test_pubkey.cpp` is test infrastructure for public key algorithms

For example it would be appropriate in this file to add `PK_Private_Key_Decoding_Test` base class and then deriving it somewhere for your actual test.

Considering the test is already hard coded to ECDH keys might as well be in `test_ecdh.cpp` I suppose.

src/tests/test_pubkey.cpp
```diff
@@ -1037,6 +1041,73 @@ BOTAN_REGISTER_TEST("pubkey", "pk_key_decoding", PK_Key_Decoding_Test);
 
 }  // namespace
 
+/**
+ * @brief Test ECC private key decoding
```
randombit, line 1045: Kind of misleading comment tbh since this is testing just one very specific aspect of this either remove it entirely or be more specific.

src/tests/test_pubkey.cpp
```diff
@@ -1037,6 +1041,73 @@ BOTAN_REGISTER_TEST("pubkey", "pk_key_decoding", PK_Key_Decoding_Test);
...
+ */
+class EC_Private_Key_Decoding_Test : public Test {
+   public:
+      //EC_Private_Key_Decoding_Test() {}
+
+   protected:
```
randombit, line 1050: `run` being protected is fine since it's only ever called from the base class, but this doesn't match the general convention used in the suite.

src/tests/test_pubkey.cpp
randombit: Actually we already have exactly such a generic test - `PK_Key_Decoding_Test` which doesn't even need to be derived from it just reads from `src/tests/data/pubkey/key_encoding.vec`

The only issue is that right now this test only handles valid keys and not invalid. This should be easy to extend by adding a `[Valid]` header for the current test data plus `[Invalid]` which inverts the success/fail logic.

randombit (changes requested):

src/tests/test_pubkey.cpp
falko-strenzke: OK, but won't be relvant any more since we are going to integrate the new tests into `PK_Key_Decoding_Test`.

src/tests/test_pubkey.cpp
falko-strenzke: The new test will disappear ...

src/tests/test_pubkey.cpp
falko-strenzke: Done.

src/tests/test_pubkey.cpp
falko-strenzke: removed.

falko-strenzke:
@randombit 
I am further wondering about the way that 
```
EC_PrivateKey::EC_PrivateKey(const AlgorithmIdentifier& alg_id,
                             std::span<const uint8_t> key_bits,
                             bool with_modular_inverse)
``` 
handles the alg_id: it never looks at the OID, only checks for the presence of parameters. A caller that calls the function with a specific algorithm identifier might expect the function to check that `alg_id` actually denotes a valid ECC key. Should we add a check for `alg_id` indicating either an ECDSA or an ECDH key?

I also think that to make the interface clearer, passing `alg_id` should be optional. Getting the group from the encoded key might be what the caller intends.

randombit:
> Should we add a check for alg_id indicating either an ECDSA or an ECDH key?

Not helpful since this is generic across algorithms this could also be SM2, ECKCDSA, GOST, ...

src/tests/test_pubkey.cpp
falko-strenzke: > Actually we already have exactly such a generic test - `PK_Key_Decoding_Test` which doesn't even need to be derived from it just reads from `src/tests/data/pubkey/key_encoding.vec`

The existing test is different. It decodes a complete private key info structure, that is not working for the intended test. I will thus create a new one in the `test_ecdh.cpp`.

falko-strenzke: @randombit 
I think I addressed all your review comments now. Could you check it again?

randombit (approved):

PR #5538, opened 2026-04-20
In timing_test shuffle the inputs before measurement

randombit:
This would otherwise systematically bias the outputs by ~20 ns or so.

### DIFF
--- a/src/cli/timing_tests.cpp
+++ b/src/cli/timing_tests.cpp
@@ -30,6 +30,7 @@
 #include <chrono>
 #include <fstream>
 #include <iostream>
+#include <numeric>
 #include <sstream>
 
 #if defined(BOTAN_HAS_BIGINT)
@@ -83,7 +84,18 @@ namespace Botan_CLI {
 
 namespace {
 
-class TimingTestTimer {
+void shuffle_idx(std::vector<size_t>& vec, Botan::RandomNumberGenerator& rng) {
+   const size_t n = vec.size();
+   for(size_t i = 0; i != n; ++i) {
+      uint8_t jb[sizeof(uint64_t)];
+      rng.randomize(jb, sizeof(jb));
+      const uint64_t j8 = Botan::load_le<uint64_t>(jb, 0);
+      const size_t j = i + static_cast<size_t>(j8) % (n - i);
+      std::swap(vec[i], vec[j]);
+   }
+}
+
+class TimingTestTimer final {
    public:
       TimingTestTimer() : m_start(get_high_resolution_clock()) {}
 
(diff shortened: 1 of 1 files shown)


src/cli/timing_tests.cpp
```diff
@@ -402,9 +417,16 @@ std::vector<std::vector<uint64_t>> Timing_Test::execute_evaluation(const std::ve
    size_t total_runs = 0;
    std::vector<uint64_t> results(inputs.size());
 
+   std::vector<size_t> indexes;
+   for(size_t i = 0; i != inputs.size(); ++i) {
```
KaganCanSit: By including the `#include <numeric>` header file, we can delegate the index assignment function to `std::iota`. The same pattern can be applied on line 664.

```c++
std::vector<size_t> indexes(inputs.size());
std::iota(indexes.begin(), indexes.end(), size_t{0});
```

src/cli/timing_tests.cpp
```diff
@@ -106,8 +117,12 @@ class Timing_Test {
          A constant seed is ok here since the timing test rng just needs to be
          "random" but not cryptographically secure - even std::rand() would be ok.
          */
-         const std::string drbg_seed(64, 'A');
-         m_rng = cli_make_rng("", drbg_seed);  // throws if it can't find anything to use
+
+#if defined(BOTAN_HAS_CHACHA_RNG)
+         m_rng = std::make_shared<Botan::ChaCha_RNG>(std::vector<uint8_t>(64, 0));
```
KaganCanSit: I think `m_rng` is used only within the `Timing_Test` class, it is not shared or moved elsewhere. Maybe we use `std::unique_ptr`.
```c++
std::unique_ptr<Botan::RandomNumberGenerator> m_rng; // LINE 148

#if defined(BOTAN_HAS_CHACHA_RNG)
         m_rng = std::make_unique<Botan::ChaCha_RNG>(std::vector<uint8_t>(64, 0));
#else
         m_rng = std::make_unique<Botan::System_RNG>();
#endif
```

KaganCanSit (commented):
Since it is a test file, not sure about this changes necessity but I've included a few minor suggestions anyway. Feel free to implement them if they seem appropriate.

src/cli/timing_tests.cpp
randombit: done

src/cli/timing_tests.cpp
randombit: Yeah I'm not sure why this was `shared_ptr` in the first place but `unique_ptr` is the logical choice in the context

PR #5539, opened 2026-04-20
Refactor and optimize Windows certificate store

randombit:
@reneme I don't know if you have any special interest re the Windows store but wanted to bring to your attention the cert cache, and the new possibility of directly implementing `contains`, both of which likely will also be beneficial for the macOS system store. I was able to do this one using MinGW+Wine but don't think there is anything similar for Mac APIs on Linux.

### DIFF
--- a/src/lib/x509/certstor_system_windows/certstor_windows.cpp
+++ b/src/lib/x509/certstor_system_windows/certstor_windows.cpp
@@ -1,6 +1,6 @@
 /*
 * Certificate Store
-* (C) 1999-2021 Jack Lloyd
+* (C) 1999-2021,2026 Jack Lloyd
 * (C) 2018-2019 Patrik Fiedler, Tim Oesterreich
 * (C) 2021      René Meusel
 *
@@ -12,10 +12,15 @@
 #include <botan/assert.h>
 #include <botan/ber_dec.h>
 #include <botan/hash.h>
+#include <botan/mutex.h>
 #include <botan/pkix_types.h>
+#include <botan/internal/fmt.h>
+#include <botan/internal/x509_cert_cache.h>
 #include <algorithm>
 #include <array>
 #include <functional>
+#include <unordered_map>
+#include <unordered_set>
 #include <vector>
 
 #define NOMINMAX 1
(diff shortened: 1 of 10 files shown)


src/lib/x509/x509cert.h
```diff
@@ -446,14 +452,14 @@ class BOTAN_PUBLIC_API(2, 0) X509_Certificate : public X509_Object {
...
-      explicit X509_Certificate(const std::vector<uint8_t>& in);
+      explicit X509_Certificate(std::span<const uint8_t> in);
 
       /**
       * Create a certificate from a buffer
       * @param data the buffer containing the DER-encoded certificate
       * @param length length of data in bytes
       */
-      X509_Certificate(const uint8_t data[], size_t length);
+      X509_Certificate(const uint8_t data[], size_t length) : X509_Certificate(std::span{data, length}) {}
```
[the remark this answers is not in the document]
randombit: jesse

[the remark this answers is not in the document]
randombit: > this is a source-breaking API change and also breaks existing in-tree call sites

Just extraordinary

src/lib/x509/certstor_system_windows/certstor_windows.cpp
reneme: Nit or "a matter of taste": I think this could become easier to read using `if constexpr`. Perhaps even inlined into the destructor avoiding the private `close()` method altogether. Along those lines:

```C++
if(m_context) {
   if constexpr(std::same_as<T, PCCERT_CONTEXT>) {
      CertFreeCertificateContext(m_context);
   } else if constexpr(std::same_as<T, HCERTSTORE>) {
      // second parameter is a flag that tells the store how to deallocate memory
      // using the default "0", this function works like decreasing the reference counter
      // in a shared_ptr
      CertCloseStore(m_context, 0);
   } else {
      BOTAN_ASSERT_UNREACHABLE();
   }
}
```

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -40,17 +45,21 @@ const std::array<const char*, 2> cert_store_names{"Root", "CA"};
  * in case of an exception and at the end of the iterative process.
  */
 template <class T>
-class Handle_Guard {
+class Handle_Guard final {
```
reneme: Perhaps restrict `T` to only the handle types that we can actually support?

```C++
template <class T>
   requires std::same_as<T, HCERTSTORE> || std::same_as<T, PCCERT_CONTEXT>
class Handle_Guard final
```

src/lib/x509/certstor_system_windows/certstor_windows.h
```diff
@@ -11,24 +11,26 @@
...
       Certificate_Store_Windows();
 
-      Certificate_Store_Windows(const Certificate_Store_Windows&) = delete;
-      Certificate_Store_Windows(Certificate_Store_Windows&&) = delete;
-      Certificate_Store_Windows& operator=(const Certificate_Store_Windows&) = delete;
-      Certificate_Store_Windows& operator=(Certificate_Store_Windows&&) = delete;
+      Certificate_Store_Windows(const Certificate_Store_Windows&) = default;
+      Certificate_Store_Windows(Certificate_Store_Windows&&) = default;
+      Certificate_Store_Windows& operator=(const Certificate_Store_Windows&) = default;
+      Certificate_Store_Windows& operator=(Certificate_Store_Windows&&) = default;
```
reneme, line 33: No destructor --> rule of zero --> just remove all those.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -84,109 +95,186 @@ class Handle_Guard {
...
-                                                           find_type,
-                                                           &blob,
-                                                           cert_context.get()))) {
-         X509_Certificate cert(cert_context->pbCertEncoded, cert_context->cbCertEncoded);
-         if(filter(certs, cert)) {
-            if(return_on_first_found) {
-               return {cert};
+      X509_Certificate materialize(const BYTE* der, DWORD len) {
+         return m_cert_cache.find_or_insert(std::span(der, len));
+      }
```
reneme: Nit:


```suggestion
      X509_Certificate materialize(const BYTE* der, DWORD len) {
         return m_cert_cache.find_or_insert({der, len});
      }
```

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -84,109 +95,186 @@ class Handle_Guard {
...
+                                                                 &blob,
+                                                                 cert_context.get()))) {
+               auto cert = materialize(cert_context->pbCertEncoded, cert_context->cbCertEncoded);
+               if(!seen.insert(cert.tag()).second) {
+                  continue;
+               }
+               if(filter(cert)) {
+                  if(return_on_first_found) {
+                     return {cert};
+                  }
```
reneme: This early-return might break RVO. Breaking out of the nested loops is cumbersome, though. On a broader note: the mechanic of going through all cert stores and then iterating all certificates in each store repeats plenty of times across this file. Perhaps its worth implementing a generic generator for it. In C++23 this could even be a coroutine using `co_yield` and `std::generator`, but in C++20 we would need to implement the coroutine promise mechanics ourselves.

Here's a (fairly simple) suggestion for a plain old enumerator:

```C++
      // TODO(C++23): consider using std::generator instead
      class CertEnumerator {
      public:
         using GetNextFn = std::function<PCCERT_CONTEXT(HCERTSTORE, PCCERT_CONTEXT)>;

         explicit CertEnumerator(GetNextFn fn)
            : m_get_next(std::move(fn))
            , m_store(open_cert_store(cert_store_names[m_store_idx])) {}
         
         PCCERT_CONTEXT next() {
            while(true) {
                  if(m_ctx.assign(m_get_next(m_store.get(), m_ctx.get())))
                  {
                     return m_ctx.get();
                  }

                  if(++m_store_idx >= cert_store_names.size())
                  {
                     return nullptr;
                  }

                  m_store.assign(open_cert_store(cert_store_names[m_store_idx]));
                  m_ctx.assign(nullptr);
            }

            BOTAN_ASSERT_UNREACHABLE();
         }

      private:
         GetNextFn                    m_get_next;
         size_t                       m_store_idx = 0;
         Handle_Guard<HCERTSTORE>     m_store;
         Handle_Guard<PCCERT_CONTEXT> m_ctx;
      };
```

With the enumerator this `search_cert_stores` method would become quite a bit simpler because it avoids the nested loops altogether. That would allow us to just `break` out of the loop and certainly benefit from RVO as a side effect.

```C++
      std::vector<X509_Certificate> search_cert_stores(const _CRYPTOAPI_BLOB& blob,
                                                       DWORD find_type,
                                                       const std::function<bool(const X509_Certificate&)>& filter,
                                                       bool return_on_first_found) {
         std::vector<X509_Certificate> certs;
         std::unordered_set<X509_Certificate::Tag, X509_Certificate::TagHash> seen;

         CertEnumerator enumerator([&blob, find_type](HCERTSTORE store, PCCERT_CONTEXT prev) {
            return CertFindCertificateInStore(
               store, PKCS_7_ASN_ENCODING | X509_ASN_ENCODING,
               WINCRYPT_UNUSED_PARAM, find_type, &blob, prev);
         });

         while(auto* ctx = enumerator.next()) {
            auto cert = materialize(ctx->pbCertEncoded, ctx->cbCertEncoded);
            if(!seen.insert(cert.tag()).second) {
               continue;
            }

            if(filter(cert)) {
               certs.push_back(std::move(cert));

               if(return_on_first_found) {
                  break;
               }
            }
         }

         return certs;
      }
```

... also, the enumerator should easily be reusable in all the other methods that follow this nested loop pattern.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -84,109 +95,186 @@ class Handle_Guard {
...
-
-         X509_DN dn;
-         dn.decode_from(dec);
-         subject_dns.emplace_back(std::move(dn));
+      bool contains(const X509_Certificate& cert) {
+         // Query the Windows stores directly by SHA-256 of the certificate DER
+
+#if !defined(CERT_FIND_SHA256_HASH)
+         constexpr DWORD CERT_FIND_SHA256_HASH = 0x00220000;
+#endif
```
reneme: 😨 ... is that to support older versions of `windows.h` that didn't define this? This really looks scary to me.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -84,109 +95,186 @@ class Handle_Guard {
...
+
+         const auto cert_sha256 = cert.certificate_data_sha256();
+
+         CRYPT_HASH_BLOB blob{};
+         blob.cbData = static_cast<DWORD>(cert_sha256.size());
+         blob.pbData = const_cast<BYTE*>(cert_sha256.data());
```
reneme: Nit:


```suggestion
         const CRYPT_HASH_BLOB blob{
            .cbData = static_cast<DWORD>(cert_sha256.size()),
            .pbData = const_cast<BYTE*>(cert_sha256.data()),
         };
```

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -199,18 +287,19 @@ std::optional<X509_Certificate> Certificate_Store_Windows::find_cert_by_pubkey_s
    blob.cbData = static_cast<DWORD>(key_hash.size());
    blob.pbData = const_cast<BYTE*>(key_hash.data());
```
reneme: Nit:


```C++
const CRYPT_HASH_BLOB blob{
   .cbData = static_cast<DWORD>(key_hash.size()),
   .pbData = const_cast<BYTE*>(key_hash.data()),
};
```

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -223,15 +312,13 @@ std::optional<X509_Certificate> Certificate_Store_Windows::find_cert_by_issuer_d
    const X509_DN& issuer_dn, std::span<const uint8_t> serial_number) const {
    std::vector<uint8_t> dn_data = issuer_dn.BER_encode();
 
-   _CRYPTOAPI_BLOB blob;
+   _CRYPTOAPI_BLOB blob{};
    blob.cbData = static_cast<DWORD>(dn_data.size());
    blob.pbData = reinterpret_cast<BYTE*>(dn_data.data());
```
reneme: Nit:


```suggestion
   const _CRYPTOAPI_BLOB blob{
      .cbData = static_cast<DWORD>(dn_data.size()),
      .pbData = reinterpret_cast<BYTE*>(dn_data.data()),
   };
```

src/lib/x509/certstor_system_windows/certstor_windows.cpp
```diff
@@ -84,109 +95,186 @@ class Handle_Guard {
...
+         if(store == nullptr) {
+            throw Internal_Error("failed to open windows certificate store '" + std::string(cert_store_name) +
+                                 "' (Error Code: " + std::to_string(::GetLastError()) + ")");
+         }
+         return store;
+      }
+
+      X509_Certificate_Cache m_cert_cache;
+      mutex_type m_mutex;
+      std::map<std::vector<uint8_t>, std::optional<X509_Certificate>> m_sha1_pubkey_to_cert;
```
reneme: Knee-jerk reaction: could we re-use the new `X509_Certificate_Cache` here? On first approximation it would probably need to be customizable over the used hash function and support "negative entries". Not sure its worth it, though.

reneme (approved):
Semantically, the changes look good to me. Also the certificate cache that would probably come in handy for a similar optimization on macOS. Perhaps it could be made a bit more generic so that it could also be reused for the SHA-1 cache in one of the windows cert store methods. But I might have missed something there.

I left a few suggestions regarding code reuse and readability. None of it is critical, I would say. But certainly nice-to-have.

Please feel free to open an issue re:macOS and assign it to me.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
randombit: Yeah that's fine too. This change was just addressing a `clang-tidy` warning about the use of `enable_if` so I didn't really consider alternatives.

src/lib/x509/certstor_system_windows/certstor_windows.h
randombit: TBH I prefer an explicit statement of the intent rather than rule of zero implicit, since zero it leaves open the question both for users and future-selves "is it intended that this type is copyable/assignable? or just an accident because the relevant decls were omitted."

src/lib/x509/certstor_system_windows/certstor_windows.cpp
randombit: Maybe? I'll leave a todo. Main goal of the cache is to prevent materializing many copies of a common root cert (for example right now if you ask for the ISRG root cert 20 times, you get 20 X509_Certificates each freshly deserialized, while with the cache you instead get 20 X509_Certificate that all point to the same `shared_ptr`). But certainly we can consider expanding the scope of what the cache can do over time.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
randombit: So this is definitely missing from MinGW `wincrypt.h`

So I figured why not just define it, and then Wine will do its thing.

But then I wondered, when exactly was this field added to Windows anyway, the MSDN docs say nothing about this. Then I thought to check when Wine added support - at the very least that would have been after Windows added it, and the commit might have some details I was otherwise unable to find.

And it turns out that afaict Wine doesn't support this flag at all. (?!?) At least I can't find any reference to it in the Wine's git history.

Which leaves open the question how `contains` is returning true in the certstore tests! The tests were passing for me just fine with this hack + MinGW + Wine. So this needs further investigation.

And I think unfortunately we will have to use `CERT_FIND_SHA1_HASH` instead :/

randombit: Though if we are forced to do that it might actually be faster and simpler to just do an explicit iteration plus a `memcmp`.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
randombit: Thanks. One other change that I want to make relates to this:

I had not realized this at first but `CertOpenSystemStoreA` is actually re-parsing *all* of the store certificates each time it's called! Or at least this is certainly true for Windows 2003 Server and I don't find any indications this was changed later. So we definitely want to cache the handles here.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
randombit: I think I'm just going to split this into two non-template types

src/lib/x509/certstor_system_windows/certstor_windows.cpp
randombit: The issue here is that the tests are implemented in terms of `Certificate_Store_System` which forwards calls, but it was not forwarding `contains` so the logic was broken but the test missed this.

Fixed the forward and changed it to search using `CERT_FIND_SHA1_HASH` then reconfirm by comparing the SHA-256.

src/lib/x509/certstor_system_windows/certstor_windows.cpp
randombit: done

src/lib/x509/certstor_system_windows/certstor_windows.cpp
randombit: done

src/lib/x509/certstor_system_windows/certstor_windows.cpp
randombit: Did something like above but changed slightly so that stores aren't constantly being opened+closed

src/lib/x509/certstor_system_windows/certstor_windows.cpp
randombit: Thinking further I'd be very hesitant to add negative entries because this cache is really intended to go in one direction - "I have some bytes that represent a certificate, please give me an `X509_Certificate` now" while a negative result is authoritatively "this cert does not exist" which doesn't make sense because the cache is not the authoritative source of anything.

[This is already an issue with the Windows cert store as it is exists right now in fact - if there is some cert `X` with a SHA-1(pubkey) hash of `Y`, if the application asks for the cert with pubkey hash `Y` and `X` is not currently in the store, it returns null. But then if `X` is added to the store, the cached negative result continues to be returned!]

PR #5544, opened 2026-04-23
chore(modernize-loop-convert): Replace index-based loops with range-for

KaganCanSit:
Fixes for clang-tidy “modernize loop conversion” warnings (not include Botan4 TODOs). NOLINT suppressions were removed.


randombit:
Thank you, but declined. If there is `NOLINT` that doesn't have a `TODO` or equivalent after it, generally the NOLINT is intentional and basically permanent in the sense that, in that particular case, it seems better to have the code in the way it currently is rather than whatever clang-tidy wants to see.

PR #5547, opened 2026-04-25
chore(IWYU): Remove redundant <vector> includes in favor of secmem.h

KaganCanSit:
Based on the IWYU pragma directive in the `secmem.h` file;

```cpp
#ifndef BOTAN_SECURE_MEMORY_BUFFERS_H_
#define BOTAN_SECURE_MEMORY_BUFFERS_H_

#include <botan/allocator.h>
#include <botan/types.h>  // IWYU pragma: export
#include <span>
#include <type_traits>
#include <vector>  // IWYU pragma: export
```

The `vector` includes found in the same file as `secmem.h` have been removed. **No changes were made to those included indirectly from other header files.**

Updates:
02.05.2026: Rebased the main.
24.05.2026: Rebased the main.

### DIFF
--- a/src/lib/asn1/der_enc.h
+++ b/src/lib/asn1/der_enc.h
@@ -13,7 +13,6 @@
 #include <functional>
 #include <optional>
 #include <span>
-#include <vector>
 
 namespace Botan {
 
--- a/src/lib/filters/filter.h
+++ b/src/lib/filters/filter.h
@@ -12,7 +12,6 @@
 #include <botan/secmem.h>
 #include <span>
 #include <string>
-#include <vector>
 
 namespace Botan {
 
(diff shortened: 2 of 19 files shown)


randombit:
Sorry for delay in response here. We don't currently use IWYU and it's not clear if these annotations are the correct choice in general, they were added a long time ago at a point of initial experimentation with it. Closing.

KaganCanSit:
> Sorry for delay in response here. We don't currently use IWYU and it's not clear if these annotations are the correct choice in general, they were added a long time ago at a point of initial experimentation with it. Closing.

@randombit Actually, I was aware of this when I made the change. For this reason I wrote in the description, I edited the `vector` header in parallel with the `secmem.h` header file. I did not make any adjustments to the other warnings.

The obvious reason why I arranged it this way was `secure_vector`. But of course, we can proceed however you wish.

PR #5549, opened 2026-04-26
refactor(aes_ni): Replace raw __m128i intrinsics with SIMD_4x32 in key schedules

KaganCanSit:
Resolves the TODO comment in `aes_ni.cpp` by converting AES-128, AES-192, and AES-256 key schedule functions to use `SIMD_4x32` instead of raw `__m128i` intrinsics.

```cpp
// NOLINTBEGIN(portability-simd-intrinsics) TODO convert to using SIMD_4x32
```

There was no performance difference compared to the `master` branch.  To be honest, I wasn’t expecting one anyway due to force inline. When I run the test multiple times, I get fairly close values. But here are the one examples;

_This Branch:_

> AES-128 [aesni] encrypt buffer size 1024 bytes: 13540.578 MiB/sec 0.19 cycles/byte (40621.75 MiB in 3000.00 ms)
> AES-128 [aesni] decrypt buffer size 1024 bytes: 13523.766 MiB/sec 0.19 cycles/byte (40571.31 MiB in 3000.00 ms)
> AES-192 [aesni] encrypt buffer size 1024 bytes: 11268.424 MiB/sec 0.23 cycles/byte (33805.31 MiB in 3000.00 ms)
> AES-192 [aesni] decrypt buffer size 1024 bytes: 11270.326 MiB/sec 0.23 cycles/byte (33811.00 MiB in 3000.00 ms)
> AES-256 [aesni] encrypt buffer size 1024 bytes: 9531.211 MiB/sec 0.27 cycles/byte (28593.69 MiB in 3000.01 ms)
> AES-256 [aesni] decrypt buffer size 1024 bytes: 9453.548 MiB/sec 0.27 cycles/byte (28360.69 MiB in 3000.00 ms)

_Master:_

> AES-128 [aesni] encrypt buffer size 1024 bytes: 13547.448 MiB/sec 0.19 cycles/byte (40642.38 MiB in 3000.00 ms)
> AES-128 [aesni] decrypt buffer size 1024 bytes: 13518.252 MiB/sec 0.19 cycles/byte (40554.81 MiB in 3000.00 ms)
> AES-192 [aesni] encrypt buffer size 1024 bytes: 11316.532 MiB/sec 0.23 cycles/byte (33949.62 MiB in 3000.00 ms)
> AES-192 [aesni] decrypt buffer size 1024 bytes: 11272.753 MiB/sec 0.23 cycles/byte (33818.31 MiB in 3000.00 ms)
> AES-256 [aesni] encrypt buffer size 1024 bytes: 9581.535 MiB/sec 0.27 cycles/byte (28744.62 MiB in 3000.00 ms)
> AES-256 [aesni] decrypt buffer size 1024 bytes: 9502.485 MiB/sec 0.27 cycles/byte (28507.50 MiB in 3000.00 ms)

---
Additional;
`SIMD` is not a I use frequently. For this reason, I wanted to treat the `TODO` messages here as an opportunity. During development, I fallowed older PRs that had been edited previously. But I am not sure full of it. I think it would be better to mark this branch draft.

With C++26, it seems like `std::simd` could be used to partially resolve future compilation complexities. https://en.cppreference.com/cpp/numeric/simd

### DIFF
--- a/src/lib/block/aes/aes_ni/aes_ni.cpp
+++ b/src/lib/block/aes/aes_ni/aes_ni.cpp
@@ -19,53 +19,50 @@ namespace {
 // NOLINTBEGIN(portability-simd-intrinsics)
 
 template <uint8_t RC>
-BOTAN_FN_ISA_AESNI inline __m128i aes_128_key_expansion(__m128i key, __m128i key_getting_rcon) {
-   __m128i key_with_rcon = _mm_aeskeygenassist_si128(key_getting_rcon, RC);
+BOTAN_FN_ISA_AESNI inline SIMD_4x32 aes_128_key_expansion(SIMD_4x32 key, SIMD_4x32 key_getting_rcon) {
+   __m128i key_with_rcon = _mm_aeskeygenassist_si128(key_getting_rcon.raw(), RC);
    key_with_rcon = _mm_shuffle_epi32(key_with_rcon, _MM_SHUFFLE(3, 3, 3, 3));
-   key = _mm_xor_si128(key, _mm_slli_si128(key, 4));
-   key = _mm_xor_si128(key, _mm_slli_si128(key, 4));
-   key = _mm_xor_si128(key, _mm_slli_si128(key, 4));
-   return _mm_xor_si128(key, key_with_rcon);
+   key ^= key.shift_elems_left<1>();
+   key ^= key.shift_elems_left<1>();
+   key ^= key.shift_elems_left<1>();
+   key ^= SIMD_4x32(key_with_rcon);
+   return key;
 }
 
 BOTAN_FN_ISA_AESNI
 void aes_192_key_expansion(
-   __m128i* K1, __m128i* K2, __m128i key2_with_rcon, secure_vector<uint32_t>& out, size_t offset) {
-   __m128i key1 = *K1;
-   __m128i key2 = *K2;
+   SIMD_4x32& K1, SIMD_4x32& K2, SIMD_4x32 key2_with_rcon, secure_vector<uint32_t>& out, size_t offset) {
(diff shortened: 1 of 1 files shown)


randombit:
Change here looks good

> With C++26, it seems like std::simd could be used to partially resolve future compilation complexities. 

Not really. Like most such proposals it is very much oriented to SAXPY type problems, ala I have two arrays of 1000000 floats and want to multiply them together asap. Almost all of our SIMD code is quite specific to the SIMD width, and often makes use of instructions unlikely to be supported by `std::simd`, Highway, or similar.

src/lib/block/aes/aes_ni/aes_ni.cpp
randombit: It's optional (what you have here is already an improvement) but ideally this function (and the others) would take as input `SIMD_4x32` values, and return `SIMD_4x32`, similar to `aesenc` and others.

KaganCanSit: > Almost all of our SIMD code is quite specific to the SIMD width.

Ohh, I see. You are wants express this specific 128-bit crypto instruction in the most direct way. 

I wanted to touch on the parts you suggested, but at first I thought there were too many changes. However, looking at it again now, I think a better revision could have been made. I need to examine the situation specific to AES-192. I’ll make the changes and see what happens.

Thanks for the quick response and information.

KaganCanSit: @randombit Converted `aes_128_key_expansion` and `aes_256_key_expansion` to take and return `SIMD_4x32` as you suggested, which cleaned up the remaining `.raw()` wrapping at the call sites. Also replaced `_mm_slli_si128` with `shift_elems_left<1>()` to keep the three helpers consistent.

~I have created a separate commit to make it easier for you to review. I can squash later.~

Updates:
02.05.2026: Rebased the master branch and squash commits.

randombit (approved):
LGTM thanks!

PR #5552, opened 2026-04-29
add test for point not on curve

falko-strenzke:


src/tests/test_ecdh.cpp
```diff
@@ -94,6 +94,22 @@ class ECDH_AllGroups_Tests : public Test {
                   const auto a_ss = ka.derive_key(0, sec1_infinity);
                });
 
+               // Regression test: prohibit loading a point not on the curve
+               result.test_throws<Botan::Decoding_Error>("point is not on curve", [&] {
+                  const auto& base_point = group.get_base_point();
```
randombit: This should use `EC_AffinePoint::generator` since the test should continue to run even after `legacy_ec_point` is removed

randombit (approved):

PR #5562, opened 2026-05-03
Various OCSP hardenings

randombit:


src/lib/x509/ocsp.h
```diff
@@ -176,6 +177,21 @@ class BOTAN_PUBLIC_API(2, 0) Response final {
...
+      *       certificate is trustworthy and authorized to do so.
+      *
+      * @param signing_certificate the certificate that signed this response
+      *                            (@sa Response::find_signing_certificate)
+      * @param restrictions on the signature validation
+      *
+      * @return status code indicating the validity of the signature
+      */
+      Certificate_Status_Code verify_signature(const X509_Certificate& signing_certificate,
+                                               const Path_Validation_Restrictions& restrictions) const;
```
reneme, line 193: Perhaps just add a default-constructed `Path_Validation_Restrictions` object and avoid the method overload that way?

reneme (approved):
Nice improvements! Thanks.

src/lib/x509/ocsp.h
randombit: Yeah sounds good

randombit: Actually I would rather not do that since it requires the definition of `Path_Validation_Restrictions` to be available which brings in quite a bit of header.

PR #5582, opened 2026-05-08
Add twisted brainpool curves

arckoor:
Adds the twisted brainpool curves 256t1 - 512t1, the parameters for which I got [here](https://std.neuromancer.sk/brainpool).
As this is my first foray into the curve world I'm not quite sure how I should proceed, especially how I should generate test cases.


src/lib/math/pcurves/pcurves_brainpool256t1/pcurves_brainpool256t1.cpp
```diff
@@ -0,0 +1,342 @@
...
+
+namespace Botan::PCurve {
+
+namespace {
+
+namespace brainpool256t1 {
+
+
+// clang-format off
+class Params final : public EllipticCurveParameters<
```
randombit, line 17: What are your performance requirements here? This ("`pcurves`") implementation approach is the fastest but also has relatively significant compile time and binary size overheads. For more niche curves typically we just fall back to the generic implementation, which is still pretty good.

randombit (commented):
Can you outline more the reasoning here especially why it's not feasible or desirable do do these as application registered curves? These t1 curves are pretty niche, as evidenced IMO by the fact that BSI has been funding work on the library for a decade and as far as I can recall it's never once come up, so I'm quite hesitant to accept.

Re testing the main thing is testing curve specific scalar multiplications and other operations (big ones being `ecc_base_point_mul.vec`, `ecc_var_point_mul.vec`, `ecc_var_point_mul2.vec`) which should be generated by some other library that supports either the curves directly or supports application specific curves (OpenSSL supports arbitrary GF(p) curves so that's my usual goto for this), or taken from any relevant reference documents. There are already some tests that run over all groups that test general mathematical invariants (`ec_point_arith`, `ecc_scalar_arith`, `ecdh_all_groups`, etc). Plus since presumably this is driven by a peer actually using it, some higher level test data (a certificate that we test validates correctly or something like that) would be helpful to make sure it's all wired in correctly.

randombit:
Hm, there is also another specific issue if we did go ahead with these curves - the t1 and r1 curves have the same group order, but we use the group order as the "key" to determine a curve uniquely from parameters. That logic would have to be modified in some way.

arckoor:
I noticed, in the end I am not the one that'll use these curves, so am currently checking with the people that will.
I'll update the PR when I know more. Just defining the curves in the app seems like a plausible route though.

randombit:
@reneme are you using these t1 curves at all?

reneme:
> are you using these t1 curves at all?

To the best of my knowledge, we don't use those.

randombit:
@arckoor any update here

One kind of intermediate make-life-easier option that would be an easy merge would be to add the OIDs and curve parameters but leave the `Impl = ` block empty [this might need some extension to the script so it knows not to do anything] then if you need these curves can just modify `ec_groups.txt` and rerun the script.

arckoor:
No, haven't gotten anything concrete back yet
I didn't close it yet because I didn't want to open the same PR again in the case I do need it, but if that's fine for you I can also just close it for the time being

src/lib/pubkey/ec_group/ec_named.cpp
```diff
@@ -386,6 +422,10 @@ OID EC_Group::EC_group_identity_from_order(const BigInt& order)
       return OID{1, 3, 36, 3, 3, 2, 8, 1, 1, 7};
    }
 
+   if(low_bits == 0x974856A7 && order == BigInt("0xA9FB57DBA1EEA9BC3E660A909D838D718C397AA3B561A6F7901E0E82974856A7")) {
+      return OID{1, 3, 36, 3, 3, 2, 8, 1, 1, 8};
+   }
```
randombit, line 427: This isnt going to work (look at earlier conditional here) - this uses group order to detect the group but is ambigious for r1 vs t1 as they have identical orders. Would need adjustment to handle this

PR #5584, opened 2026-05-08
Fix a KMAC padding edge case

randombit:
With certain very large keys KMAC would compute an incorrect result.


falko-strenzke (approved):
Fixes the bug and looks fine to me.

PR #5595, opened 2026-05-11
Add support for RFC 9608 No Revocation Available extension

randombit:


src/lib/x509/x509_ext.cpp
```diff
@@ -1872,6 +1874,66 @@ void OCSP_NoCheck::decode_inner(const std::vector<uint8_t>& buf) {
...
+   //   party MUST consider the certificate invalid.
+
+   const Extensions& exts = subject.v3_extensions();
+
+   if(const auto* bc = exts.get_extension_object_as<Basic_Constraints>(); bc != nullptr && bc->is_ca()) {
+      // RFC 9608 Section 3:
+      //    The certificate MUST NOT also include the basic constraints
+      //    certificate extension with the cA BOOLEAN set to TRUE
+      cert_status.at(pos).insert(Certificate_Status_Code::NO_REV_AVAIL_INVALID_USE);
+   }
```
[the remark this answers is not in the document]
randombit: Unlike the other requirements which we enforce here, non-criticality is a requirement only on issuers not on relaying parties.

src/lib/x509/x509path.cpp
```diff
@@ -709,7 +746,10 @@ CertificatePathStatusCodes PKIX::check_ocsp_online(const std::vector<X509_Certif
       const auto& subject = cert_path.at(i);
       const auto& issuer = cert_path.at(i + 1);
 
-      if(subject.ocsp_responder().empty()) {
+      if(skip_revocation_check(subject)) {
+         ocsp_response_futures.emplace_back(
+            std::async(std::launch::deferred, []() -> std::optional<OCSP::Response> { return std::nullopt; }));
```
reneme, line 751: Without trying, I think this should work and avoid launching a thread for nothing:

```suggestion
         auto p = std::promise<std::optional<OCSP::Response>>();
         ocsp_response_futures.emplace_back(p.get_future());
         p.set(std::nullopt);
```

src/lib/x509/x509path.cpp
```diff
@@ -765,6 +805,12 @@ CertificatePathStatusCodes PKIX::check_crl_online(const std::vector<X509_Certifi
 
    for(size_t i = 0; i != cert_path.size(); ++i) {
       const auto& cert = cert_path.at(i);
+
+      if(skip_revocation_check(cert)) {
+         future_crls.emplace_back(std::future<std::optional<X509_CRL>>());
```
reneme: This won't produce a valid `std::future` that can be queried, to the best of my knowledge. I believe you'll have to do the promise-to-future dance sketched above.

reneme (approved):

src/lib/x509/x509path.cpp
randombit: `std::launch::deferred` doesn't launch a thread - it just returns a future that evaluates the first time `get` is called

src/lib/x509/x509_ext.cpp
```diff
@@ -1872,6 +1874,66 @@ void OCSP_NoCheck::decode_inner(const std::vector<uint8_t>& buf) {
...
+   BER_Decoder(buf, BER_Decoder::Limits::DER()).decode_null().verify_end();
+}
+
+void NoRevocationAvailable::validate(const X509_Certificate& subject,
+                                     const std::optional<X509_Certificate>& /*issuer*/,
+                                     const std::vector<X509_Certificate>& /*cert_path*/,
+                                     std::vector<std::set<Certificate_Status_Code>>& cert_status,
+                                     size_t pos) {
+   // RFC 9608 Section 2:
+   //    This extension MUST NOT be present in CA public key certificates.
```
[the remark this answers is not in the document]
randombit: Doesn't seem necessary here, none of these checks are actually critical in any way the RFC just decided to dump issuer requirements on us.

src/lib/x509/x509path.cpp
randombit: Indeed so, that's just an invalid/uninitialized future. Fixed

PR #5598, opened 2026-05-13
Add URI and email name constraint processing

randombit:
Also first class `DNSName`, `URI`, and `EmailAddress` types


randombit:
@reneme fyi ping. Certainly reviews would be welcome, don't let the +7000 throw you off too much, +4000 of that is test data...

src/lib/utils/socket/socket_udp.cpp
```diff
@@ -361,11 +361,11 @@ std::unique_ptr<OS::SocketUDP> OS::open_socket_udp(std::string_view hostname,
 }
 
 std::unique_ptr<OS::SocketUDP> OS::open_socket_udp(std::string_view uri_string, std::chrono::microseconds timeout) {
-   const auto uri = URI::from_any(uri_string);
-   if(uri.port() == 0) {
+   const auto authority = URI::Authority::parse(uri_string);
+   if(!authority.has_value() || !authority->port().has_value()) {
       throw Invalid_Argument("UDP port not specified");
    }
```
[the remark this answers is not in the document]
randombit: This was actually only used as host:port, for Roughtime.

src/lib/utils/socket/socket_udp.cpp
```diff
@@ -361,11 +361,11 @@ std::unique_ptr<OS::SocketUDP> OS::open_socket_udp(std::string_view hostname,
 }
 
 std::unique_ptr<OS::SocketUDP> OS::open_socket_udp(std::string_view uri_string, std::chrono::microseconds timeout) {
-   const auto uri = URI::from_any(uri_string);
-   if(uri.port() == 0) {
+   const auto authority = URI::Authority::parse(uri_string);
+   if(!authority.has_value() || !authority->port().has_value()) {
```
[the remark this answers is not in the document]
randombit: Good catch actually, added a check in the URI parser

randombit:
@reneme FYI this last force push added many more test cases and also consolidated all of the name constraint related test chains into a single directory. No logic changes.

randombit:
Last force push had one minor logic change: in checking email name constraints on RFC822 names we applied the conservative check (`mailbox_form_constraint_covers_domain`), which caused us to exclude very reasonable chains that other implementations accept (due to the fact that apparently nobody implements RFC 9598). Now that's only apply to SmtpUtf8Mailbox names.

src/lib/utils/dns_name/dns_name.cpp
```diff
@@ -0,0 +1,146 @@
...
+   return canon;
+}
+
+}  // namespace
+
+//static
+std::optional<DNSName> DNSName::from_string(std::string_view name) {
+   try {
+      auto canon = check_and_canonicalize_dns_name(name);
+      if(canon.find('*') != std::string::npos) {
```
reneme: Nit:


```suggestion
      // TODO(C++23): use std::string::contains`
      if(canon.find('*') != std::string::npos) {
```

... about time that this is added. >.<

src/lib/utils/dns_name/dns_name.cpp
```diff
@@ -0,0 +1,146 @@
...
+      if(cu >= 128) {
+         throw Decoding_Error("DNS name must not contain any extended ASCII code points");
+      }
+      const uint8_t mapped = DNS_CHAR_MAPPING[cu];
+      if(mapped == 0) {
+         throw Decoding_Error("DNS name includes invalid character");
+      }
+
+      if(mapped == '-') {
+         if(i == 0 || (i > 0 && name[i - 1] == '.')) {
```
reneme, line 96: Side note: Technically the `i > 0` is implied by the first part of the condition, but the compiler will probably be smart enough to optimize this out.

src/lib/utils/dns_name/dns_name.cpp
```diff
@@ -0,0 +1,146 @@
...
+         } else if(i == name.size() - 1 || (i < name.size() - 1 && name[i + 1] == '.')) {
+            throw Decoding_Error("DNS name has label with trailing hyphen");
+         }
+      }
+      canon.push_back(static_cast<char>(mapped));
+   }
+
+   if(current_label_length == 0) {
+      throw Decoding_Error("DNS name contains empty label");
+   }
```
reneme, line 110: IIUC this can only happen if the input DNS name candidate ends with a `.` or is empty altogether both of which is checked explicitly above, so this seems redundant.

src/lib/utils/email/email.cpp
```diff
@@ -0,0 +1,121 @@
...
+   * RFC 5322 atext: ALPHA / DIGIT plus "!#$%&'*+-/=?^_`{|}~" (plus ".")
+   *
+   * Anything outside this set in the local part requires quoting,
+   * which we deliberately don't accept.
+   */
+   constexpr auto is_atext_or_dot = CharacterValidityTable::alpha_numeric_plus(".!#$%&'*+-/=?^_`{|}~");
+
+   if(addr.empty() || !is_valid_utf8(std::string(addr))) {
+      return {};
+   }
```
reneme: Sad that this sanity check needs a short-lived `std::string` allocation. Worth a TODO perhaps?

src/lib/utils/email/email.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+   if(addr.empty()) {
+      return {};
+   }
```
reneme: Redundant. Checked before.

src/lib/utils/email/email.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+   // RFC 3696 section 3:
+   //
+   //  period (".") may also appear [in an email address local-part],
+   //  but may not be used to start or end the local part, nor may two
+   //  or more consecutive periods appear.
+
+   if(local.front() == '.' || local.back() == '.' || local.find("..") != std::string_view::npos) {
```
reneme: Super-Nit:


```suggestion
   // TODO(C++23): use std::string::contains
   if(local.starts_with('.') || local.ends_with('.') || local.find("..") != std::string_view::npos) {
```

src/lib/utils/email/email.cpp
```diff
@@ -0,0 +1,121 @@
...
+         return {};
+      }
+   }
+
+   auto parsed_domain = DNSName::from_string(domain);
+   if(!parsed_domain.has_value()) {
+      return {};
+   }
+
+   return std::make_pair(std::string(local), parsed_domain.value());
```
reneme, line 68: If you would define `local` as non-const in line 41 you could avoid the additional allocation here and simply `std::move` it.

src/lib/utils/email/email.cpp
```diff
@@ -0,0 +1,121 @@
...
+   auto parsed = parse_email_address(addr);
+
+   if(parsed) {
+      // Verify the local-part is all ASCII
+      for(const char c : parsed->first) {
+         if(static_cast<uint8_t>(c) >= 0x80) {
+            return {};
+         }
+      }
+      return EmailAddress(parsed->first, parsed->second);
```
reneme: Probably you want to move those:


```suggestion
      return EmailAddress(std::move(parsed->first), std::move(parsed->second));
```

src/lib/utils/email/email.cpp
```diff
@@ -0,0 +1,121 @@
...
+      /*
+      * RFC 9598 Section 3
+      *   SmtpUTF8Mailbox subjectAltName MUST NOT be used unless the Local-part
+      *   of the email address contains non-ASCII characters.  When the Local-
+      *   part is ASCII, rfc822Name subjectAltName MUST be used instead of
+      *   SmtpUTF8Mailbox.
+      *
+      * We do not currently enforce this on the decoding side.
+      */
+      return SmtpUtf8Mailbox(parsed->first, parsed->second);
```
reneme:
```suggestion
      return SmtpUtf8Mailbox(std::move(parsed->first), std::move(parsed->second));
```

src/lib/utils/dns_name/dns_name.cpp
```diff
@@ -0,0 +1,146 @@
...
+std::optional<DNSName> DNSName::from_string(std::string_view name) {
+   try {
+      auto canon = check_and_canonicalize_dns_name(name);
+      if(canon.find('*') != std::string::npos) {
+         return std::nullopt;
+      }
+      return DNSName(std::move(canon));
+   } catch(Decoding_Error&) {
+      return std::nullopt;
+   }
```
reneme: In the internal parsing function of the email addresses you're using `std::optional` for the "invalid" case instead of throwing. Given that this anyway swallows the error message and translates it into a `std::nullopt` I would prefer that here as well. I.e. make `check_and_canonicalize_dns_name()` return an optional on parsing error and don't throw unneccesarily.

src/lib/utils/email/email.h
```diff
@@ -0,0 +1,73 @@
...
+      std::string m_local_part;
+      DNSName m_domain;
+};
+
+/**
+* A parsed internationalized mailbox (`SmtpUTF8Mailbox`) as defined by
+* RFC 9598. The mailbox is `local-part "@" domain`, where local-part
+* is allowed to be UTF-8
+*/
+class BOTAN_PUBLIC_API(3, 13) SmtpUtf8Mailbox final {
```
reneme, line 66: Technically, all instances of `EmailAddress` could also be valid instances of `SmtpUtf8Mailbox`, right? Does it make sense to make it a base class and allow for polymorphy? (I didn't look into the usage locations at this point, so this might by utter bs.)

src/lib/utils/uri/uri.cpp
```diff
@@ -0,0 +1,328 @@
...
+   }
+   uint32_t port = 0;
+   for(const char c : s) {
+      if(!is_ascii_digit(c)) {
+         return {};
+      }
+      port = port * 10 + static_cast<uint32_t>(c - '0');
+      if(port > 65535) {
+         return {};
+      }
```
reneme: Super-nit: technically this can only happen after all five digits were parsed. So it is enough to check that after the loop, along with `port == 0`.

(I wonder whether compilers figure that out and hoist the check automatically... Edit: no, [doesn't look like they do](https://godbolt.org/z/M3v3b7MMd).)

src/lib/utils/uri/uri.cpp
```diff
@@ -0,0 +1,328 @@
...
+      host = *ipv6;
+   } else {
+      // host[:port] with no brackets. Only one ':' is allowed (port).
+      const auto colon = raw.find(':');
+      if(colon == std::string_view::npos) {
+         host_view = raw;
+      } else {
+         host_view = raw.substr(0, colon);
+         port_str = raw.substr(colon + 1);
+      }
```
reneme, line 293: The "only one colon is allowed" isn't checked explicitly. It will not make it through the parsing of the port number, so all good technically. But perhaps its worth considering checking that here explicitly, especially if its already mentioned in a comment like that.

src/lib/utils/uri/uri.cpp
```diff
@@ -0,0 +1,328 @@
...
+
+      if(host_view.empty()) {
+         return {};
+      }
+
+      // Technically valid per RFC 3986 but likely not something we want to support
+      if(host_view.ends_with('.')) {
+         return {};
+      }
```
reneme, line 302: No objection to verifying that explicitly here, but it will be rejected by the `DNSName::from_string()` anyway. So perhaps not worth to check it here again?

src/lib/utils/uri/uri.cpp
```diff
@@ -0,0 +1,328 @@
...
+   switch(m_host.index()) {
+      case 0:
+         return HostKind::DNS;
+      case 1:
+         return HostKind::IPv4;
+      case 2:
+         return HostKind::IPv6;
+      default:
+         BOTAN_ASSERT_UNREACHABLE();
+   }
```
reneme: Personally, I would prefer this:


```suggestion
   if (std::holds_alternative<DNSName>(m_host)) {
      return HostKind::DNS;
   } else if (std::holds_alternative<IPv4Address>(m_host)) {
      return HostKind::IPv4;
   } else if (std::holds_alternative<IPv6Address>(m_host)) {
      return HostKind::IPv6;
   }
   
   BOTAN_ASSERT_UNREACHABLE();
```

... because it won't break if the underlying variant definition changes. And if it breaks, it will fail to compile. Otherwise, you might as well YOLO it like so:


```suggestion
   return static_cast<HostKind>(m_host.index());
```

src/lib/x509/name_constraint.cpp
```diff
@@ -25,8 +26,264 @@ class DER_Encoder;
...
+* seems in practice implementations accept subtrees for DNS
+* constraints as well.
+*
+* If @p require_fqdn is true, then single-label host names are rejected.
+*/
+std::optional<std::string> validate_subtree_constraint_host(std::string_view input, bool require_fqdn) {
+   if(input.empty()) {
+      return std::nullopt;
+   }
+   const bool subtree = (input.front() == '.');
```
reneme: Nit:

```suggestion
   const bool subtree = input.starts_with('.');
```

src/lib/x509/name_constraint.cpp
```diff
@@ -25,8 +26,264 @@ class DER_Encoder;
...
+   }
+}
+
+std::optional<GeneralName::URIConstraint> GeneralName::URIConstraint::from_san_value(std::string_view full_uri) {
+   if(URI::parse(full_uri).has_value()) {
+      return URIConstraint(std::string(full_uri));
+   } else {
+      return std::nullopt;
+   }
+}
```
reneme, line 99: Technically it would be better to take a `std::string` by value here, and `std::move` it into the `URIConstraint` constructor if the parsing succeeded.

src/lib/x509/name_constraint.cpp
```diff
@@ -25,8 +26,264 @@ class DER_Encoder;
...
+   if(input.empty()) {
+      return std::nullopt;
+   }
+   if(input.find('@') != std::string_view::npos) {
+      // Mailbox form:
+      auto email = EmailAddress::from_string(input);
+      if(!email.has_value()) {
+         return std::nullopt;
+      }
+      return EmailConstraint(email->local_part() + "@" + email->domain().to_string());
```
reneme: `email->to_string()` should do the trick here, no?


```suggestion
      return EmailConstraint(email->to_string());
```

src/lib/x509/name_constraint.cpp
```diff
@@ -25,8 +26,264 @@ class DER_Encoder;
...
+      return false;
+   }
+
+   if(constraint.empty()) {
+      return true;
+   }
+
+   BOTAN_ASSERT_NOMSG(name.size() > constraint.size());
+
+   const std::string_view substr = name.substr(name.size() - constraint.size(), constraint.size());
```
reneme: Super-nit:


```suggestion
   const std::string_view substr = name.substr(name.size() - constraint.size());
```

... but I find by not repeating the length, this actually becomes easier to reason about.

reneme (commented):
Not a full review so far (still need to go through the name constraint checking itself), plenty of opinionated nits but no smoking guns anywhere.

(The GitHub webapp is struggleing considerably with this change set. ^^)

src/lib/utils/email/email.cpp
randombit: There is a todo for this already in charset.h

src/lib/utils/dns_name/dns_name.cpp
randombit: Main goal for code like this is to be not just correct but obviously correct/defensive, so reader doesn't have to reason out "is this safe/checked, or not?"

src/lib/utils/email/email.h
randombit: > Technically, all instances of EmailAddress could also be valid instances of SmtpUtf8Mailbox, right? 

Well yes, sorta. But also definitely no.

`EmailAddress` is RFC 5280's

```
   GeneralName ::= CHOICE {
        ...
        rfc822Name                      [1]     IA5String,
        ...
}
```

Thus it cannot contain UTF-8 characters.

`SmtpUtf8Mailbox` is from RFC 9598:

```
   SmtpUTF8Mailbox ::= UTF8String (SIZE (1..MAX))
```

This type is literally *required* to contain UTF-8 high characters (RFC 9598: "SmtpUTF8Mailbox subjectAltName MUST NOT be used unless the Local-part of the email address contains non-ASCII characters.") this is not enforced currently but we may have to. [

But more critically, the rules for name constraint processing between the two names are subtly different (RFC 9598 Section 6). Because of course they are. If we treated rfc822Name names as we do SmtpUTF8Mailbox, we would have to reject reasonable and historically valid chains. Likewise if we treated SmtpUtf8Mailbox like rfc822Names, we would accept chains which are forbidden by the RFC.

Thus IMO two distinct types unrelated by any kind of inheritance is warranted. The header comments should be expanded a bit to explain more about the connections and differences between the two types, though.

src/lib/utils/dns_name/dns_name.cpp
randombit: `check_and_canonicalize_dns_name` was just moved here wholesale, which is why it's like that, but yeah makes sense to modify it to ruturn an optional and then change the exception messages to comments prior to each conditional.

src/lib/utils/dns_name/dns_name.cpp
reneme: Totally reasonable. The compiler will minimize this anyway.

src/lib/utils/uri/uri.cpp
randombit: Would want to leave it since technically it's valid for either and we may need to remove the restriction on DNS without necessarily wanting to at the same time unrestrict URLs, tiny footgun to trust DNSName will always restrict this.

src/lib/utils/dns_name/dns_name.cpp
randombit: I'm going to leave it for defensiveness against future logic changes, but left a comment that it should never be hit if things up until now had gone as expected.

src/lib/utils/email/email.cpp
randombit: On further look - up until this point `local` is just a `string_view` there is no second allocation. I changed the types from `auto` to `string_view` so this is more obvious.

src/lib/utils/uri/uri.cpp
randombit: I had it written that way due to worrying about possible integer overflows, but the input is already checked to be at most 5 chars so that can't happen regardless. Moved.

src/lib/utils/uri/uri.cpp
randombit: done

src/lib/x509/name_constraint.cpp
randombit: I'd rather leave this as is since the argument at this point is anyway a `string_view`, and in a very minor way that would pre-commit us to `URIConstraint`s underlying type being a `std::string` which is not necessarily actually the right choice in the long run. It was just the expedient option for getting this PR over with. :sweat_smile: I've been working on this branch for 4 weeks now and at this point I'm tired of it.

randombit:
Review changes pushed as separate commits for easier reading, I'll squash those before merge.

src/lib/utils/email/email.cpp
randombit: Fixed in #5603

randombit: Unresolving as a reminder to fix the allocation here when rebasing

randombit:
@reneme I know the NC changes are complicated. If you're happy with the `URI`/`DNSName`/`EmailAddress` alternately I can split those changes out to their own distinct PR. I have several branches/PRs that could use these types but are don't depend on the name constraints changes so they could start landing.

reneme:
> I have several branches/PRs that could use these types but are don't depend on the name constraints changes so they could start landing.

Sure, by all means. The strong-type changes and their respective parsing/validation logic LGTM now.

src/lib/x509/name_constraint.cpp
```diff
@@ -25,8 +26,264 @@ class DER_Encoder;
...
+         return false;
+      }
+      return wildcard_label_matches(p_left, c_base);
+   }
+
+   // p_tail starts with ".". If it ends (label-aligned) with "." + c_base,
+   // then every wildcard expansion produces a name ending with that
+   // suffix, which is inside c_base's subtree (both bare-host and
+   // leading-dot forms accept proper-subdomain entries).
+   if(p_tail.size() >= c_base.size() + 1) {
```
reneme: Perhaps overly cautious, but if `c_base` has a length of `SIZE_MAX` this wraps the integer. I don't think that can be reached or exploited via supported public interfaces, but this function is (technically) exposed to the user via a `BOTAN_TEST_API` export and takes raw string views. Maybe a case to use `checked_add` here?

src/lib/x509/name_constraint.cpp
```diff
@@ -25,8 +26,264 @@ class DER_Encoder;
...
+   }
+   const auto prefix = pattern_label.substr(0, star);
+   const auto suffix = pattern_label.substr(star + 1);
+   if(candidate.size() < prefix.size() + suffix.size()) {
+      return false;
+   }
+   if(!candidate.starts_with(prefix)) {
+      return false;
+   }
+   return candidate.substr(candidate.size() - suffix.size()) == suffix;
```
reneme: Consistently using the higher level abstractions provided by `string_view` would simplify that:


```suggestion
   return candidate.starts_with(prefix) && candidate.ends_with(suffix)
```

... the length-checks are hidden behind string_view's methods.

src/lib/x509/name_constraint.cpp
```diff
@@ -193,34 +489,135 @@ void GeneralName::decode_from(BER_Decoder& ber) {
...
+      * matching domain using the additional check in mailbox_form_constraint_covers_domain.
+      */
+      return false;
+   }
+   if(!constraint.empty() && constraint.front() == '.') {
+      // Leading-dot subtree form: suffix match including the dot.
+      if(candidate_domain.size() < constraint.size()) {
+         return false;
+      }
+      return candidate_domain.substr(candidate_domain.size() - constraint.size()) == constraint;
```
reneme: Similarly to the previous comment:


```suggestion
      return candidate_domain.ends_with(constraint);
```

src/lib/x509/name_constraint.cpp
```diff
@@ -25,8 +26,264 @@ class DER_Encoder;
...
+}
+
+}  // namespace
+
+std::optional<GeneralName::DNSConstraint> GeneralName::DNSConstraint::from_string(std::string_view input) {
+   if(auto canonical = validate_subtree_constraint_host(input, /*require_fqdn=*/false)) {
+      return DNSConstraint(std::move(*canonical));
+   } else {
+      return std::nullopt;
+   }
```
reneme: Nit for C++23: 


```suggestion
   // TODO(C++23): return validate_subtree_constraint_host(input, /*require_fqdn=*/false)
   //                        .transform([](std::string s) { return DNSConstraint(std::move(s)); });
   if(auto canonical = validate_subtree_constraint_host(input, /*require_fqdn=*/false)) {
      return DNSConstraint(std::move(*canonical));
   } else {
      return std::nullopt;
   }
```

This applies in other places too, I'm not going to repeat it though.

src/lib/x509/name_constraint.cpp
```diff
@@ -25,8 +26,264 @@ class DER_Encoder;
...
+/*
+* Validate a host constraint - either a DNS name or a subtree of the
+* form of "." followed by a DNS name. RFC 5280 4.2.1.10 defines this
+* style for URI and email constraints. For DNS it is silent, but it
+* seems in practice implementations accept subtrees for DNS
+* constraints as well.
+*
+* If @p require_fqdn is true, then single-label host names are rejected.
+*/
+std::optional<std::string> validate_subtree_constraint_host(std::string_view input, bool require_fqdn) {
```
reneme: Certainly a matter of taste, but replacing boolean parameters by purpose-built little enums results in better readabiltiy than documenting the meaning of a boolean parameter as `/*inline=*/` comments at the call site. 

Like so:

```C++
enum class RequireFQDN : bool { yes = true, no = false };

auto validate_subtree_constraint_host(..., RequireFQDN fqdn);

// call site
if (auto s = validate_subtree_constraint_host(..., RequireFQDN::yes)) {}
```

... and it doesn't even introduce an awful lot of boiler plate, IMO.

src/lib/x509/name_constraint.cpp
```diff
@@ -25,8 +26,264 @@ class DER_Encoder;
...
+   const bool subtree_form = (constraint.front() == '.');
+   const std::string_view c_base = subtree_form ? constraint.substr(1) : constraint;
+   if(c_base.empty()) {
+      return false;
+   }
+
+   const auto first_dot = pattern.find('.');
+   const std::string_view p_left = (first_dot == std::string_view::npos) ? pattern : pattern.substr(0, first_dot);
+   const std::string_view p_tail =
+      (first_dot == std::string_view::npos) ? std::string_view{} : pattern.substr(first_dot);
```
reneme, line 169: Not sure this idiom actually repeats, but for readability this would be helpful:


```suggestion
   auto [p_left, p_tail] = split_before(pattern, '.');
```

... with an implementation along those lines:

```C++
std::pair<std::string_view, std::string_view> split_before(std::string_view s, char delim) {
    const auto cut = s.find(delim);
    return (cut != std::string_view::npos)
       ? std::make_pair(s.substr(0, cut), s.substr(cut))
       : std::make_pair(s, std::string_view{});
}
```

src/lib/x509/pkix_types.h
```diff
@@ -393,38 +432,80 @@ class BOTAN_PUBLIC_API(2, 0) GeneralName final : public ASN1_Object {
...
+
+            auto operator<=>(const URIConstraint&) const = default;
+
+         private:
+            explicit URIConstraint(std::string value) : m_value(std::move(value)) {}
+
+            std::string m_value;
+      };
+
+      using NameVariant = std::variant<EmailConstraint, DNSConstraint, URIConstraint, X509_DN, IPv4Subnet, IPv6Subnet>;
```
reneme, line 506: I didn't go through all usages in a structured way, but judging from `.decode_from()` this variant (in `m_name` ) could easily become the single source of state for that class (entirely ditching the `m_type` member). I might have missed something, though.

Simply add a `std::monostate` as the first variant alternative (to represent `NameType::Unknown`). _This would anyway be a good idea to have a "well defined" default-constructed state in the variant._ And then create a trivial `struct OtherName` to represent `NameType::Other`.

The `.type_code()` getter could just map into the public enum, and internally we could just stick to the variant.

... obviously this doesn't have to (perhaps shouldn't) happen in this PR.

src/lib/x509/name_constraint.cpp
```diff
@@ -193,34 +489,135 @@ void GeneralName::decode_from(BER_Decoder& ber) {
...
+      "example.com".  When the constraint does not begin with a period,
+      it specifies a host.
+
+   So a bare-host URI constraint is exact-match only; subdomains don't
+   satisfy it. dns_subtree_match handles the leading-dot form correctly.
+   */
+   if(!constraint.empty() && constraint.front() == '.') {
+      return dns_subtree_match(host, constraint);
+   }
+   return host == constraint;
```
reneme, line 552: Is an empty constraint even allowed (or sensible in any way)? Perhaps just:


```suggestion
   return (constraint.starts_with('.')
      ? dns_subtree_match(host, constraint)
      : host == constraint;
```

src/lib/x509/name_constraint.cpp
```diff
@@ -256,31 +653,35 @@ GeneralName::MatchResult GeneralName::matches(const X509_Certificate& cert) cons
...
+            if(cn.find('.') == std::string::npos) {
+               continue;
+            }
+            if(string_to_ipv4(cn).has_value()) {
+               continue;
+            }
+            if(auto dns_form = DNSName::from_san_string(cn)) {
+               score.add(dns_subtree_match(dns_form->to_string(), constraint));
             }
          }
```
reneme, line 673: This is what is being deprecated for removal in Botan4, right? Maybe worth adding a `TODO(Botan4)`, if I didn't miss it somehow.

src/lib/x509/name_constraint.cpp
reneme, line 697: Similar to before: perhaps a `TODO(Botan4)` re:deprecation?

src/lib/x509/name_constraint.cpp
```diff
@@ -586,20 +1020,57 @@ bool NameConstraints::is_permitted(const X509_Certificate& cert, bool reject_unk
...
+            if(!is_permitted_ipv4(ipv4.value())) {
+               return false;
+            }
+         } else if(cn.find('.') != std::string::npos) {
+            if(auto dns_form = DNSName::from_san_string(cn)) {
+               if(!is_permitted_dns_name(*dns_form)) {
                   return false;
                }
             }
          }
```
reneme, line 1054: Again: Deprecation, right?

src/lib/x509/name_constraint.cpp
reneme, line 673: Deprecation TODO?

src/lib/x509/name_constraint.cpp
```diff
@@ -754,20 +1305,49 @@ bool NameConstraints::is_excluded(const X509_Certificate& cert, bool reject_unkn
...
+            if(is_excluded_ipv4(ipv4.value())) {
+               return true;
+            }
+         } else if(cn.find('.') != std::string::npos) {
+            if(auto dns_form = DNSName::from_san_string(cn)) {
+               if(is_excluded_dns_name(*dns_form)) {
                   return true;
                }
             }
          }
```
reneme, line 1340: Deprecation note?

reneme (approved):
Neither me nor Claude found anything really worth commenting about regarding the correctness of the implementation. Admittedly, I didn't study the relevant RFC thoroughly but reviewed for "sensible" implementation. The extensive test harness will very likely do the rest. 👍 

A few nits and suggestions along the way, otherwise looks good to me. Thanks a lot!

src/lib/x509/name_constraint.cpp
randombit: > Is an empty constraint even allowed

This is, as far as I can tell, completely unspecified by the RFCs. Surveying implementations, opinions differ. OpenSSL accepts an empty constraint and interprets it to mean "anything goes". Go and some other libraries use the same interpretation - but only for DNS names, not URI/email. It seems to me like a minefield so I went with "it's an invalid constraint". In the worst case it's a rejects-valid issue, someone can open an issue with a real world chain that exhibits the problem.

src/lib/x509/name_constraint.cpp
randombit: I added the comment once, when it comes up seems like it would be obvious to apply elsewhere within the same file.

src/lib/x509/name_constraint.cpp
randombit: Seems implausible in most scenarios, but easier to check than try to conclusively prove it could never be a problem.

src/lib/x509/pkix_types.h
randombit: It's a good idea. I left a TODO comment here

PR #5608, opened 2026-05-18
Add URI, DNS, and EmailAddress types

randombit:
Forked out of #5598


randombit:
I am not sure this really can be pulled in without significant other name constraint changes, which will just cause some fun merge conflicts later. Closing, this lands with #5598

PR #5610, opened 2026-05-19
Fix various cipher mode edge case bugs

randombit:
Add tests that catch any such edge cases across all modes


reneme (approved):

randombit:
Fixed some additional GCM bugs pointed out by @haruki0409 then found and fixed further state machine errors in XTS, CFB, and OCB.

PR #5624, opened 2026-05-26
chore(test): Resolve TODO in Roughtime_Request test

KaganCanSit:
Following the `// TODO should use byte comparison function` message, I updated the `test_is_true/false` calls to `test_bin_eq/ne`. 

However, after reviewing the commit where this message was added, I’m not entirely sure this is exactly what was intended. This change could have been implemented in commit 505d93a. 

```cpp
// Before
result.test_eq("encode", type == "Valid", request == Botan::typecast_copy<std::array<uint8_t, 1024>>(request_v.data()));


// After
// TODO should use byte comparison function
if(type == "Valid") {
    result.test_is_true("encode", request == Botan::typecast_copy<std::array<uint8_t, 1024>>(request_v.data()));
} else {
    result.test_is_true("encode", request != Botan::typecast_copy<std::array<uint8_t, 1024>>(request_v.data()));
}
```

Is `test_bin_eq` / `test_bin_ne` the intended direction here, or should the behavior be preserved differently? For example, should the `request_v` span be taken as a fixed size of `1024`? The test vectors in the `src/tests/data/roughtime/` directory are always 1024 bytes long, so I believe they can be compared directly. Additionally, if this were not the case, `typecast_copy` would result in an undefined behavior in cases where the length is less than 1024.

### DIFF
--- a/src/tests/test_roughtime.cpp
+++ b/src/tests/test_roughtime.cpp
@@ -36,11 +36,10 @@ class Roughtime_Request_Tests final : public Text_Based_Test {
 
          const auto request = Botan::Roughtime::encode_request(nonce);
 
-         // TODO should use byte comparison function
          if(type == "Valid") {
-            result.test_is_true("encode", request == Botan::typecast_copy<std::array<uint8_t, 1024>>(request_v.data()));
+            result.test_bin_eq("encode", request, request_v);
          } else {
-            result.test_is_true("encode", request != Botan::typecast_copy<std::array<uint8_t, 1024>>(request_v.data()));
+            result.test_bin_ne("encode", request, request_v);
          }
 
          return result;


KaganCanSit:
I'm seeing that most CI tests are failing with the message `Your account is suspended. Please visit https://support.github.com for more information.` However, since the runners aren't running through my profile, this might be a technical issue on GitHub's end. I'll retrigger them with `--amend --no-edit` in a little while.

randombit:
> Is test_bin_eq / test_bin_ne the intended direction here,

Yes I think that was the idea. I'm not sure why it was not done at the time. Possibly `test_bin_eq` at that point did not work with std::array/std::span, and was vector-specific?

CI failures are due to an ongoing Github incident, they seem to have somehow suspended the system account that github actions jobs run as.

KaganCanSit:
> > Is test_bin_eq / test_bin_ne the intended direction here,
> 
> Yes I think that was the idea. I'm not sure why it was not done at the time. Possibly `test_bin_eq` at that point did not work with std::array/std::span, and was vector-specific?
> 
> CI failures are due to an ongoing Github incident, they seem to have somehow suspended the system account that github actions jobs run as.

Thanks for the information; I think I’m on the right track then. I’ll trigger the CI again.

Github Status Right Now
```
Update - Actions is experiencing degraded availability. We are continuing to investigate.
May 26, 2026 - 11:19 UTC
Investigating - We are investigating reports of degraded performance for Actions and Pages
May 26, 2026 - 10:57 UTC
```

Also, if you have time, I’m available this week. I can set aside some time to my PR requests here and if require make the requested changes. (Some PR is very small (line < 10))


Update:
When I reviewed the 505d93a commit, I found that the necessary function calls for binary content were not available at that time. And size it is const as (1024), I don’t think any further changes will be needed if the CI passes.

randombit (approved):
LGTM thanks

Issue #1449, opened 2018-02-09
[FEATURE] PKCS12/PFX support


DevSidious:
Support **PKCS12**. That is to say, add functionality to support loading of certificates and private keys from **PKCS12** formatted data and dumping of certificates and private keys to **PKCS12** formatted data.

**I am willing to financially support the development effort.**

What say ye? Anyone interested?

randombit:
I have heard @securitykernel and company have some interest in supporting PKCS12 but I don't know what their timeline is. I emailed him pointing to here, but he likely will not see my mail before Monday. If a patch is not coming from them in the near future, I do take on funded dev projects so that's an option.

My opinions on PKCS12 are mostly unprintable, but it is commonly used so probably worth supporting given botan's general goals.

DevSidious:
Thanks @randombit . Works for me. How would you define **the near future** ? :)

randombit:
OK so turns out R&S has deferred working on PKCS12 support. So if this is something you really want, the options would be to fund the development or of course do it yourself. If you are interested in having me do it, drop me an email (`jack@randombit.net`) and we can work out the specifics.

DevSidious:
Thanks. I'll get back at you ASAP.

PR #5625, opened 2026-05-27
Add PKCS#12

dmazzella:
### Summary

This PR adds comprehensive PKCS#12 (PFX) support to Botan, enabling parsing and generation of the widely-used certificate/key.

### Motivation

Currently, Botan users must rely on external tools or libraries to handle PKCS#12 files. This PR provides native support and solve https://github.com/randombit/botan/issues/1449


src/lib/pkcs12/pkcs12.h
```diff
@@ -0,0 +1,255 @@
...
+#include <vector>
+
+namespace Botan {
+
+class RandomNumberGenerator;
+
+/**
+ * Options for PKCS#12 generation
+ */
+struct BOTAN_PUBLIC_API(3, 13) PKCS12_Options {
```
randombit: We've been trying to move away from this style of struct-of-options interface. For existing interfaces like `X509_Cert_Options` (#4996) this inevitably results in bikeshedding over exactly how to replace it, etc but we can just skip it for completely new interfaces like here.

Also the legacy flag having this kind of override behavior is a bit confusing.

Specific suggestions:

* Completely encapsulated `class` with getters
* Use `optional<T>` to distinguish empty vs unset
* Use static pseudo-constructors as shorthand for common settings

Sketch of possible callers:

```
const auto legacy = PKCS12_Options::legacy_compat(password, friendly_name);
const auto modern = PKCS12_Options::modern(password); // friendly_name defaults to nullopt...
// for when you want to do something really specific:
const auto custom = PKCS12_Options(password, friendly_name, pbe_name, ...);
```

src/lib/pkcs12/pkcs12.h
```diff
@@ -0,0 +1,255 @@
...
+
+      /**
+       * @return the private key, or nullptr if not present
+       */
+      const std::shared_ptr<Private_Key>& private_key() const { return m_private_key; }
+
+      /**
+       * @return the end-entity certificate, or nullptr if not present
+       */
+      const std::shared_ptr<X509_Certificate>& certificate() const { return m_certificate; }
```
randombit: `optional<X509_Certificate>` would be more idiomatic - X509_Certificate is internally a shared_ptr to make them cheap to copy.

src/lib/pkcs12/pkcs12.h
```diff
@@ -0,0 +1,255 @@
...
+
+      /**
+       * @return intermediate/CA certificates in the chain
+       */
+      const std::vector<std::shared_ptr<X509_Certificate>>& ca_certificates() const { return m_ca_certs; }
```
randombit: I'm not quite sure of the usage here but maybe `optional<vector<X509_Certificate>>`?

src/lib/pkcs12/pkcs12.h
```diff
@@ -0,0 +1,255 @@
...
+
+      /**
+       * @return all certificates (end-entity + chain)
+       */
+      std::vector<std::shared_ptr<X509_Certificate>> all_certificates() const;
```
randombit: Here again should just be X509_Certificate

This should specify the ordering, if any. (If none it should explicitly say that the ordering is arbitrary)

src/lib/pkcs12/pkcs12.h
```diff
@@ -0,0 +1,255 @@
...
+
+      /**
+       * @return the friendly name attribute, if present
+       */
+      const std::string& friendly_name() const { return m_friendly_name; }
```
randombit: Should be a `const std::optional<std::string>&`

src/lib/pkcs12/pkcs12.h
```diff
@@ -0,0 +1,255 @@
...
+
+      /**
+       * @return true if this contains any CA certificates
+       */
+      bool has_ca_certificates() const { return !m_ca_certs.empty(); }
+
+      /**
+       * @return true if this contains a friendly name
+       */
+      bool has_friendly_name() const { return !m_friendly_name.empty(); }
```
randombit: This (and maybe most of the others predicates) could be removed - `friendly_name().has_value()` seems idiomatic enough.

src/lib/pkcs12/pkcs12.h
```diff
@@ -0,0 +1,255 @@
...
+
+      /**
+       * @return the localKeyId bag attribute of the private key, or empty if absent
+       */
+      const std::vector<uint8_t>& local_key_id() const { return m_local_key_id; }
```
randombit: Again optional

src/lib/pkcs12/pkcs12.h
```diff
@@ -0,0 +1,255 @@
...
+       * legacy-compatible values ("PBE-SHA1-3DES", "SHA-1", 2048) regardless of
+       * what was set above. Use when interoperability with old software is required.
+       */
+      bool legacy_compat = false;
+};
+
+/**
+ * Result of parsing a PKCS#12 file
+ */
+class BOTAN_PUBLIC_API(3, 13) PKCS12_Data {
```
randombit: should probably be `final`

src/lib/pkcs12/pkcs12.h
randombit: this should also be `final`

src/lib/pkcs12/pkcs12.h
```diff
@@ -0,0 +1,255 @@
...
+       *         (e.g. SecretBag). Empty in normal files.
+       */
+      const std::vector<OID>& unknown_bag_types() const { return m_unknown_bag_types; }
+
+   private:
+      friend class PKCS12;
+
+      std::shared_ptr<Private_Key> m_private_key;
+      std::shared_ptr<X509_Certificate> m_certificate;
+      std::vector<std::shared_ptr<X509_Certificate>> m_ca_certs;
```
randombit: The shared_ptr on the certificates aren't doing anything.

src/lib/pkcs12/pkcs12.h
randombit: Looking more at usage, this should maybe be renamed `PKCS12_Export_Options`? Or something along those lines.

src/lib/pkcs12/pkcs12.h
```diff
@@ -0,0 +1,255 @@
...
+      /**
+       * Parse a PKCS#12/PFX file
+       *
+       * @param data the PFX file contents
+       * @param password the password to decrypt the file
+       * @return parsed contents
+       * @throws Decoding_Error if parsing fails
+       * @throws Invalid_Authentication_Tag if MAC verification fails
+       */
+      static PKCS12_Data parse(std::span<const uint8_t> data, std::string_view password);
```
randombit: This PKCS12/PKCS12_Data split seems at the last non-idiomatic for this codebase. I'd suggest `PKCS12_Data` becomes `PKCS12` and the parse function moves to it

src/lib/pkcs12/pkcs12.h
```diff
@@ -0,0 +1,255 @@
...
+
+      /**
+       * Parse a PKCS#12/PFX file from a DataSource
+       *
+       * @param source data source containing the PFX file
+       * @param password the password to decrypt the file
+       * @return parsed contents
+       */
+      static PKCS12_Data parse(DataSource& source, std::string_view password);
```
randombit: `DataSource` has been kind of a problematic design - if it is not strictly needed I'd suggest dropping this  at least for now.

src/lib/pkcs12/pkcs12.h
```diff
@@ -0,0 +1,255 @@
...
+       * Create a PKCS#12/PFX file
+       *
+       * @param key the private key to include (may be nullptr)
+       * @param cert the end-entity certificate (may be nullptr)
+       * @param ca_certs additional CA/intermediate certificates
+       * @param options generation options
+       * @param rng random number generator
+       * @return DER-encoded PFX file
+       */
+      static std::vector<uint8_t> create(const Private_Key* key,
```
randombit: This parse + create split works well for values that never change after creation, but seems a bit unfortunate with respect to PKCS12 where it's quite reasonable to, for example, add a certificate to the bundle later.

I'd suggest this. What is currently `PKCS12_Data` gains setters plus an `export()` function which returns the new encrypted blob. Then creation looks something like

```
PKCS12_Data pkcs12;
pkcs12.set_friendly_name("My Bundle");
pkcs12.add_key(...);
const auto blob = pkcs12.export(PKCS12_Export_Options::modern(password));
```

src/lib/pkcs12/pkcs12.h
```diff
@@ -0,0 +1,255 @@
...
+ * Result of parsing a PKCS#12 file
+ */
+class BOTAN_PUBLIC_API(3, 13) PKCS12_Data {
+   public:
+      PKCS12_Data() = default;
+
+      /**
+       * @return the private key, or nullptr if not present
+       */
+      const std::shared_ptr<Private_Key>& private_key() const { return m_private_key; }
```
randombit: This seems to support just a single private key, IIUC pkcs12 can support arbitrary number of keys, even if not supported during parsing it would be nice if the API didn't require being extended to support this case later.

randombit (changes requested):
Thanks. This was just a review of `pkcs12.h` on the general shape of the public API, haven't looked at the implementation yet at all.

dmazzella:
> Thanks. This was just a review of `pkcs12.h` on the general shape of the public API, haven't looked at the implementation yet at all.

All changes done

doc/api_ref/pkcs12.rst
```diff
@@ -0,0 +1,343 @@
...
+      Throws ``Invalid_Argument`` if the options are inconsistent (e.g.
+      unsupported algorithm, MAC enabled with empty password) or if a
+      stored private key does not match any stored certificate.
+
+      Iteration counts above ``PKCS12_MAX_ITERATIONS`` (1 000 000) and a
+      SafeContentsBag nesting depth above ``PKCS12_MAX_NESTING`` (10) are
+      rejected.
+
+Example
+-----------------------------------------
```
reneme: Please change those examples into compilable examples in `src/examples`. Then just pull the examples into the documentation. Please see other parts of the documentation for an idea of how this works. 

Files in `src/examples` are regularly built in our CI and make sure that our documented examples actually compile and are properly updated when the library's API evolves.

dmazzella: done

doc/api_ref/pkcs12.rst
```diff
@@ -0,0 +1,290 @@
...
+     - Legacy; widest compatibility
+   * - ``with_mac_digest``
+     - ``"SHA-384"``, ``"SHA-512"``
+     - ``sha2_64``
+     - Uncommon; supported for parsing and generation
+
+Command Line Interface
+-----------------------------------------
+
+Two CLI commands are available when Botan is built with filesystem support.
```
randombit: The cli docs should go to `doc/cli.rst`

src/tests/test_pkcs12.cpp
```diff
@@ -0,0 +1,1679 @@
...
+         return result;
+      }
+
+      Test::Result test_parse_file_openssl_empty_password(const std::string& filename) {
+         Test::Result result("PKCS12 parse file openssl empty password: " + filename);
+
+         try {
+            const std::vector<uint8_t> pfx_data = Test::read_binary_data_file("pkcs12/" + filename);
+            result.test_throws<Botan::Invalid_Authentication_Tag>("openssl empty password throws",
+                                                                  [&]() { Botan::PKCS12(pfx_data, ""); });
```
randombit: So IIUC the issue here is we are testing that a default OpenSSL p12 file *doesn't parse*, and the real only issue is because OpenSSL defaults to encoding an empty password as the (incorrect, non-standard) empty string, rather than `{0x00, 0x00}`. Can we just, when decrypting and the password string is empty, try first the correct encoding and then fall back to the OpenSSL encoding? This causes a 2x overhead when decrypting but that's probably a somewhat nicer UX than the p12 file simply failing to decrypt.

src/cli/pkcs12.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+         const auto pfx_data = bundle.export_to(options, rng());
+
+         write_output(pfx_data);
+      }
+};
+
+BOTAN_REGISTER_COMMAND("pkcs12_export", PKCS12_Export);
+
+class PKCS12_Import final : public Command {
```
randombit: This doesn't really import *into* anything, IMO should be named `pkcs12_info` matching eg `cert_info`

src/examples/pkcs12_export_chain.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <memory>
+#include <vector>
+
+int main() {
+   Botan::AutoSeeded_RNG rng;
+   const auto group = Botan::EC_Group::from_name("secp256r1");
+
+   // Issuing CA.
+   const Botan::ECDSA_PrivateKey ca_key(rng, group);
+   Botan::X509_Cert_Options ca_opts("CN=Example CA");
```
randombit: This constructor doesn't parse the argument as a standard DN string, so this creates a cert with a literal CN of "CN=Example CA"

src/examples/pkcs12_export_chain.cpp
```diff
@@ -0,0 +1,43 @@
...
+   ca_opts.CA_key();
+   const auto ca_cert = Botan::X509::create_self_signed_cert(ca_opts, ca_key, "SHA-256", rng);
+
+   // End-entity, signed by the CA.
+   auto ee_key = std::make_shared<Botan::ECDSA_PrivateKey>(rng, group);
+   Botan::X509_Cert_Options ee_opts("CN=example.com");
```
randombit: Similar CN= issue here

src/examples/pkcs12_export.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include <iostream>
+#include <memory>
+
+int main() {
+   Botan::AutoSeeded_RNG rng;
+
+   // Generate an ECDSA private key + self-signed certificate to bundle.
+   auto key = std::make_shared<Botan::ECDSA_PrivateKey>(rng, Botan::EC_Group::from_name("secp256r1"));
+
+   const Botan::X509_Cert_Options cert_opts("CN=example.com");
```
randombit: CN= is wrong (or at least highly confusing) here also

src/lib/pkcs12/pkcs12_pbe/pkcs12_pbe.cpp
```diff
@@ -0,0 +1,160 @@
...
+   const OID sha1_3des = OID::from_string("PBE-SHA1-3DES");
+   if(oid == sha1_3des) {
+      return {sha1_3des, "TripleDES/CBC", 24};
+   }
+   const OID sha1_2des = OID::from_string("PBE-SHA1-2DES");
+   if(oid == sha1_2des) {
+      return {sha1_2des, "TripleDES/CBC", 16};
+   }
+   throw Decoding_Error(fmt("Unsupported PKCS#12 PBE algorithm: {}", oid.to_string()));
+}
```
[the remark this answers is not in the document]
randombit: Please revert this change (also for functions below): the map from OID string to OID value is very fast now (since #4896) so I'd rather just see `if(oid == OID::from_string("..."))`

src/lib/pkcs12/pkcs12_pbe/pkcs12_pbe.h
```diff
@@ -0,0 +1,57 @@
...
+
+namespace Botan {
+
+class RandomNumberGenerator;
+
+/// Maximum allowed iteration count for PKCS#12 KDF/PBE/MAC operations
+inline constexpr size_t PKCS12_MAX_ITERATIONS = 1'000'000;
+
+/// Maximum allowed SafeContentsBag nesting depth
+inline constexpr size_t PKCS12_MAX_NESTING = 10;
```
randombit: This constant seems unrelated to the PBE and probably should be defined elsewhere

src/lib/pkcs12/pkcs12.cpp
```diff
@@ -0,0 +1,915 @@
...
+/*
+* Resolve a MAC digest OID to a hash name, throwing on unsupported algorithms.
+*/
+std::string resolve_mac_hash(const OID& oid) {
+   static const OID sha1_oid = OID::from_string("SHA-1");
+   static const OID sha224_oid = OID::from_string("SHA-224");
+   static const OID sha256_oid = OID::from_string("SHA-256");
+   static const OID sha384_oid = OID::from_string("SHA-384");
+   static const OID sha512_oid = OID::from_string("SHA-512");
+   static const OID sha512_256_oid = OID::from_string("SHA-512-256");
```
randombit: These should just do `if(oid == OID::from_string("SHA-1"))` avoiding the statics

src/lib/pkcs12/pkcs12.cpp
```diff
@@ -0,0 +1,915 @@
...
+* Parse attributes from a SafeBag. Only FriendlyName and LocalKeyId are
+* handled; other attributes are silently skipped (RFC 7292 §4.2).
+*/
+void parse_bag_attributes(BER_Decoder& decoder, std::string& friendly_name, std::vector<uint8_t>& local_key_id) {
+   if(!decoder.more_items()) {
+      return;
+   }
+
+   static const OID friendly_name_oid = OID::from_string("PKCS9.FriendlyName");
+   static const OID local_key_id_oid = OID::from_string("PKCS9.LocalKeyId");
```
randombit: Remove `static` for these (but leave as named variables since we are comparing them repeatedly in the loop below)

src/lib/pkcs12/pkcs12.cpp
```diff
@@ -0,0 +1,915 @@
...
+                         std::vector<OID>& unknown_bag_types,
+                         size_t depth = 0) {
+   if(depth >= PKCS12_MAX_NESTING) {
+      throw Decoding_Error("PKCS#12: SafeContentsBag nesting too deep");
+   }
+   static const OID cert_bag_oid = OID::from_string("PKCS12.CertBag");
+   static const OID shrouded_key_bag_oid = OID::from_string("PKCS12.PKCS8ShroudedKeyBag");
+   static const OID key_bag_oid = OID::from_string("PKCS12.KeyBag");
+   static const OID safe_contents_bag_oid = OID::from_string("PKCS12.SafeContentsBag");
+   static const OID x509_cert_oid = OID::from_string("PKCS9.X509Certificate");
```
randombit: Remove `static` here

src/lib/pkcs12/pkcs12.cpp
```diff
@@ -0,0 +1,915 @@
...
+// PKCS12
+//
+
+PKCS12::PKCS12(std::span<const uint8_t> data, std::string_view password) {
+   std::vector<ParsedCert> cert_entries;
+   std::vector<ParsedKey> key_entries;
+
+   DataSource_Memory src(data.data(), data.size());
+
+   BER_Decoder pfx(src);
```
randombit: You can use the `span` constructor of `BER_Decoder` here; right now the behavior between `span` and `DataSource_Memory` constructors is basically the same but we'll be able to take advantage of the `span` constructor later eg due to knowing the precise length of the data.

src/lib/pkcs12/pkcs12.cpp
```diff
@@ -0,0 +1,915 @@
...
+      }
+   }
+
+   static const OID cert_bag_oid = OID::from_string("PKCS12.CertBag");
+   static const OID shrouded_key_oid = OID::from_string("PKCS12.PKCS8ShroudedKeyBag");
+   static const OID x509_cert_oid = OID::from_string("PKCS9.X509Certificate");
+   static const OID friendly_name_oid = OID::from_string("PKCS9.FriendlyName");
+   static const OID local_key_id_oid = OID::from_string("PKCS9.LocalKeyId");
+   static const OID pkcs7_data_oid = OID::from_string("PKCS7.Data");
+   static const OID pkcs7_enc_data_oid = OID::from_string("PKCS7.EncryptedData");
```
randombit: Should just be local `const` without the static

src/tests/test_pkcs12.cpp
```diff
@@ -0,0 +1,1679 @@
+/*
+* PKCS#12 Tests
+* (C) 2026 Damiano Mazzella
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include "tests.h"
+
+#if defined(BOTAN_HAS_PKCS12) && defined(BOTAN_HAS_ECDSA)
```
randombit: ECDSA guard seems odd here. It maybe should just be restricted to some subset of the tests?

src/tests/test_pkcs12.cpp
```diff
@@ -0,0 +1,1679 @@
...
+      }
+
+      // Build a PFX (no MAC) with a KeyBag containing an unencrypted private key
+      Test::Result test_key_bag() {
+         Test::Result result("PKCS12 KeyBag (unencrypted key)");
+
+         auto rng = Test::new_rng("PKCS12_key_bag");
+         auto creds = generate_credentials(*rng);
+
+         const Botan::OID pkcs7_data_oid = Botan::OID::from_string("PKCS7.Data");
```
randombit: For this kind of test just generate the data locally once, save to `src/tests/data`, and parse

src/tests/test_pkcs12.cpp
```diff
@@ -0,0 +1,1679 @@
...
+
+      Test::Result test_nesting_depth_exceeded() {
+         Test::Result result("PKCS12 nesting depth exceeded rejected");
+
+         const Botan::OID pkcs7_data_oid = Botan::OID::from_string("PKCS7.Data");
+         const Botan::OID safe_contents_bag_oid = Botan::OID::from_string("PKCS12.SafeContentsBag");
+
+         // Build SafeContents with nesting depth > PKCS12_MAX_NESTING (10)
+         // Each level: SafeBag { SafeContentsBag, [0] { SafeContents { ... } } }
+         std::vector<uint8_t> safe_contents;
```
randombit: Likewise this one should just be pre-generated and saved to `src/tests/data/pkcs12`

randombit (changes requested):
Thanks overall looking really good. My big criticism would be in the tests where there are several taking ~20-50 lines to construct some kind of invalid p12 file, and then testing how parsing reacts to it. Tests of invalid/degenerate inputs are great! But the test input should just be pre-generated and stored in `src/tests/data/pkcs12` with the rest, and then the test code is just a read+parse.

Also various minor nits. I see Copilot suggested caching OID values as static, not unreasonable but we've made lookups sufficiently fast now that I think it's not worth the cognitive complexity of doing anything further. I left comments on some but not all of the spots, grepping `static const OID` should find them all.

src/lib/pkcs12/pkcs12.cpp
```diff
@@ -0,0 +1,938 @@
...
+            fmt("PKCS#12: unsupported cert encryption algorithm '{}'", opts.cert_encryption_algo()));
+      }
+      if(opts.password().empty()) {
+         throw Invalid_Argument("PKCS#12: cert_encryption_algo requires a non-empty password");
+      }
+   }
+   if(opts.include_mac()) {
+      if(opts.password().empty()) {
+         throw Invalid_Argument("PKCS#12: include_mac requires a non-empty password");
+      }
```
randombit: Certainly encrypting/MAC with an empty password is theater, but is there any reason to actually *prohibit* it? IIUC it's fairly common for Java keystores to end up in this state

src/lib/pkcs12/pkcs12.h
```diff
@@ -0,0 +1,254 @@
...
+#include <string>
+#include <string_view>
+#include <vector>
+
+namespace Botan {
+
+class RandomNumberGenerator;
+
+/// Maximum allowed SafeContentsBag nesting depth during parsing (anti-DoS).
+inline constexpr size_t PKCS12_MAX_NESTING = 10;
```
randombit: I don't see any reason for this to be in the public header

randombit:
@dmazzella Almost there I think. One tiny nit re the nesting constant, and one more general policy question - I think either the restriction on empty pw should be removed, or the reasoning for rejecting it be made more clear in the docs. Possibly this is just a holdover from earlier where decryption could not handle the buggy OpenSSL encoding of an empty password and so trying to avoid that situation?

dmazzella:
> @dmazzella Almost there I think. One tiny nit re the nesting constant, and one more general policy question - I think either the restriction on empty pw should be removed, or the reasoning for rejecting it be made more clear in the docs. Possibly this is just a holdover from earlier where decryption could not handle the buggy OpenSSL encoding of an empty password and so trying to avoid that situation?

Thanks for the review! You're right on both counts. The empty-password rejection was really just leftover caution from the earlier OpenSSL empty-encoding handling; there's no real technical reason to block it, especially since those files turn up all the time in practice. I've dropped the restriction: encryption and the MAC both round-trip fine with an empty password via the RFC 7292 encoding, and I updated the docs and tests to match. I also moved PKCS12_MAX_NESTING into the .cpp, since nothing outside it needs the constant. Pushed.

randombit (approved):
LGTM thanks for working through this extended review process @dmazzella

PR #5626, opened 2026-05-28
Add missing FIPS 186-5 GCD check in Lucas probable prime test

Shounak611:
Implements the additional FIPS 186-5 B.3.3 check:

gcd(C, (1-D)/4) == 1

inside is_lucas_probable_prime().

This aligns the implementation with the updated FIPS 186-5 specification.

### DIFF
--- a/src/lib/math/numbertheory/primality.cpp
+++ b/src/lib/math/numbertheory/primality.cpp
@@ -52,6 +52,24 @@ bool is_lucas_probable_prime(const BigInt& C, const Barrett_Reduction& mod_C) {
       }
    }
 
+   // FIPS 186-5 B.3.3 additional check: GCD(C, (1-D)/4) = 1
+   BigInt Q_abs;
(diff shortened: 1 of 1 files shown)


randombit:
I would at the very least want some testcases that actually hits this. `(1-D)/4)` is going to just be 1,2,3,... afaict this is basically just testing (in a somewhat expensive way) that `C`doesn't have any very small factors. And I'm pretty sure our existing check that `jacobi(C,D) != 0` catches this anyway.

If needed, it would be much more effective to precompute `2*3*5*...` up to some reasonable size then gcd C with that. That GCD would eliminate strictly more values of `C` than this check would.

Shounak611:
I’ll verify whether the additional GCD check actually rejects any cases not already handled by the existing Jacobi condition and try to produce concrete examples/benchmarks. If it turns out to be redundant or not materially beneficial, I’ll remove the change from the PR.

PR #5642, opened 2026-06-02
Improve prefetching for table based implementations

randombit:


KaganCanSit:
Whenever I have time, I read the project's PRs to learn from the experience of you and the other maintainers and contributors. It was nice to learn that Camellia was developed by Mitsubishi Electric and NTT as an independent block cipher with security comparable to AES — I had heard the name before but never looked into it (https://www.rfc-editor.org/rfc/rfc3713.html). Thank you.

This PR targets to performance; however, as a small suggestion to enhance readability/DRY. I thought the eight M constant values defined in the same way within the `F` and `F_1tab` could be moved to a shared MDS namespace along with the SBOX tables. This would prevent the need to copy and move them in the event of any similar changes in the future (though this is unlikely) and make updates easier.

```cpp
namespace Camellia_F {

// ... SBOX definitions ...

// P-function diffusion matrix columns (RFC 3713)
namespace MDS {
   constexpr uint64_t M1 = 0x0101010001000001;
   constexpr uint64_t M2 = 0x0001010101010000;
   constexpr uint64_t M3 = 0x0100010100010100;
   constexpr uint64_t M4 = 0x0101000100000101;
   constexpr uint64_t M5 = 0x0001010100010101;
   constexpr uint64_t M6 = 0x0100010101000101;
   constexpr uint64_t M7 = 0x0101000101010001;
   constexpr uint64_t M8 = 0x0101010001010100;
}  // namespace MDS

uint64_t F(uint64_t v, uint64_t K) {
   const uint64_t x = v ^ K;
   const uint64_t Z1 = MDS::M1 * SBOX1[get_byte<0>(x)];
   const uint64_t Z2 = MDS::M2 * SBOX2[get_byte<1>(x)];
   const uint64_t Z3 = MDS::M3 * SBOX3[get_byte<2>(x)];
   // ...
   // and F_1tab would use the same MDS:: prefix
```

randombit:
@KaganCanSit not a bad idea thanks. I went with another approach for deduplicating (extracting out the entire P-function linear layer into its own function)

KaganCanSit:
> @KaganCanSit not a bad idea thanks. I went with another approach for deduplicating (extracting out the entire P-function linear layer into its own function)

Extracting the entire linear layer is a more complete solution, it deduplicates both the constants and the logic in one step. An elegant solution. 🏹 Thanks for the ping.

PR #5654, opened 2026-06-05
check output span length in hex_decode

sahvx655-wq:
The std::span overload of hex_decode forwards straight to the raw pointer version without ever consulting output.size(), so the decoder writes input.length()/2 bytes (the leading clear_mem alone zeroes that many) into whatever span it was handed. Handing it a span shorter than the decoded output walks off the end; under AddressSanitizer this shows up as a heap-buffer-overflow WRITE inside hex_decode, reached purely through the public span API. While checking the codec helpers I noticed base64_decode(std::span,...) already rejects this exact case, so the two decoders disagreed on a basic safety property.

The fix rejects an undersized buffer up front, mirroring base64_decode. The bound is (input.size() + 1) / 2 rather than input.size() / 2 because an odd-length input still touches output[input.size()/2] for the trailing half nibble before that byte is reported as incomplete. Keeping the check inside the span overload means the buffer length travels with the buffer, so callers using the span form get the same guarantee the base64 form already provides.

### DIFF
--- a/src/lib/codec/hex/hex.cpp
+++ b/src/lib/codec/hex/hex.cpp
@@ -130,6 +130,9 @@ size_t hex_decode(uint8_t output[], std::string_view input, bool ignore_ws) {
 }
 
 size_t hex_decode(std::span<uint8_t> output, std::string_view input, bool ignore_ws) {
+   if(output.size() < (input.size() + 1) / 2) {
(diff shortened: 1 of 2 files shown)


randombit:
Thank you for the patch. By chance this was already addressed (in a slightly different way) in #5649

Issue #5661, opened 2026-06-11
Possible integer overflow in Stateful_Key_Index_Registry


TJ-91:
In https://github.com/randombit/botan/pull/5358#discussion_r3381241853 Stephan Ehlen points out that the XMSS^MT implementation first advances the index and then refuses to sign if the state is exhausted, advancing the index past the valid range.

This does not seem like an issue at first, since we check whether the index exceeds the total number of signatures afterwards. However, when repeatedly trying to sign with an exhausted key, at some point, you will overflow the `uint64_t` counter and then end up reserving the first leaf nodes again, breaking the security of the system. This affects all stateful hash signature algorithms. 

I propose to re-design the registry s.t. it does not advance the index if the max is reached. Either the `reserve_next_index` must receive the maximum number of signatures as a parameter, or, when registering a key, it must already provide the maximum number of signatures and store it alongside the index. Then, after locking, we can check if the index can be advanced or not.

randombit:
> However, when repeatedly trying to sign with an exhausted key, at some point, you will overflow the uint64_t counter

This isn't practically possible - even if each failed attempt took just 1 nanosecond an overflow would take 500 years to occur. Storing the max as part of the registry is fine though and would simplify `remaining_operations`

TJ-91:
I didn't do the math before - you're right, it doesn't seem possible to run into the overflow. I still have arguments for changing the implementation:

It's non-sensical to increase the index beyond the max and it's just cleaner to not do it. Even without an overflow, this already gives unexpected results when querying the current index. Someone checking for `current_index() == max` (instead of using `>= max`, `< max`, or using `remaining_operations`) would run into this. I didn't check, but if the private key gets serialized with an invalid index value, I can imagine loading the private key again will fail due to a sanity check on the index range. Instead, the key should load and run into the much clearer and expected failure mode and indicate an exhausted state when signing.

Further, it's a popular approach to split a SHBS key into distinct subtrees (distinct set of leaf indices) in order to allow for simultanious and independent use of a key on multiple devices. If someone implemented this a little carelessly, they might run into the overflow issue. Sorry for the somewhat contrived argument but the point is, I think it's a bug and it might cause subtle problems in some way or another.

randombit:
Fix in #5662 ptal

PR #5662, opened 2026-06-11
Fix some stateful signature edge cases

randombit:
If you invoked an exhausted XMSS/LMS key ~ 2**64 times, eventually the counter would overflow, resulting in a leaf reuse. This practically speaking cannot occur since it requires the actions of the legitimate signer and they are probably not going to spin in a failing signature generation loop for years waiting for an overflow.

Fix other related bugs. In HSS-LMS, it was possible to decode a corrupted private key with its index pre-set to 2**64-1. The loader would miss that the index exceeded the maximum for the key. The first signature attempt would fail, and a subsequent attempt would reuse leaf 0. An attacker who can arbitrarily set the index can more easily attack the key by rolling back to a prior state (eg setting index = 0 directly) with the same effect, so this isn't really an attack, but this particular instance is easily avoidable.

In XMSS correctly handle exhausted keys. An exhausted key could previously be serialied, but not deserialized again. Now it can be deserialized and just reports itself as exhausted.

#5661

### DIFF
--- a/src/lib/pubkey/hss_lms/hss.cpp
+++ b/src/lib/pubkey/hss_lms/hss.cpp
@@ -74,6 +74,26 @@ std::vector<LMS_Tree_Node_Idx> derive_lms_leaf_indices_from_hss_index(HSS_Sig_Id
    return q;
 }
 
+/**
+ * Encode the HSS-LMS parameter set, for use as the algorithm parameters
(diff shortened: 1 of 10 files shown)


TJ-91:
thanks, looks good to me and should address all of my points :+1:

PR #5680, opened 2026-06-19
Fix bugs in ISO 9796 signature verification, and deprecate it

randombit:
Verification would ignore explicit vs implicit settings, and accept either format. It would also reject roughly 1/128 valid signatures due to mistakes in handling leading zero bytes in the recovered message representative.

The choice of implicit vs explicit hash identifier is done using "exp" vs "imp" tags. However if the string was neither value this would be ignored, defaulting to explicit.

Considering the fact that any production usage would quickly hit the rejects-valid bug, and that we have received no reports of this problem, it seems likely that there are no production users of this padding scheme. Given this, deprecate it.

Following the logic in #4293, remove iso9796 from the bsi module policy.


randombit:
@FAlbertDev review ping here is re the BSI policy changes. Alternate approach would be to change configure.py so it allows a module policy to request a deprecated module, I'd be fine with that also.

FAlbertDev:
>@FAlbertDev review ping here is re the BSI policy changes. Alternate approach would be to change configure.py so it allows a module policy to request a deprecated module, I'd be fine with that also.

I think your approach is sensible. IMO, it's best to keep the policies free from any deprecated modules. If users want to enable deprecated modules it has to be a concious decision 👍

FAlbertDev (approved):

PR #5686, opened 2026-06-22
Add ML-KEM composite KEM

falko-strenzke:
Integration of ML-KEM composites according to draft-ietf-lamps-pq-composite-kem

# Already solved issues
~~The following problem arises in the context of the MLKEM-composite implementation.
It is based on the inconsistency between these two functions defined in `pk_keys.h`:~~

```c++
      virtual std::unique_ptr<PK_Ops::Encryption> create_encryption_op(RandomNumberGenerator& rng,
                                                                       std::string_view params,
                                                                       std::string_view provider) const;

      virtual std::unique_ptr<PK_Ops::KEM_Encryption> create_kem_encryption_op(std::string_view params,
                                                                               std::string_view provider) const;
```

~~The problem is that the implementation of  `create_kem_encryption_op` in `MLKEM_Composite_PublicKey` requires an RNG to internally call `create_encryption_op()` for RSA.
Thus inside the constructor of the `MLKEM_Composite_Encapsulation_Operation`, an RNG is needed to instantiate the traditional encryption operation.~~

~~It is not apparent to me why `create_encryption_op()` should require an RNG in the first place – and I didn't investigate. But generally, contrary to a decryption OP, it should not require randomization in setting up precomputations due to a side channel countermeasure.~~

~~The only reasonable solution that I came to is to add an optional RNG argument to the interface:~~

```
      virtual std::unique_ptr<PK_Ops::KEM_Encryption> create_kem_encryption_op(
         std::string_view params, std::string_view provider, RandomNumberGenerator* rng_may_be_null = nullptr) const;
```

~~This also requires a new function in the FFI interface. I am quite certain that this exact approach to the problem will not find acceptance. It's purpose is to make things work for now and provide a basis for the discussion of the proper solution.~~

~~The only alternative I could think of would be to let the constructor of `MLKEM_Composite_Encapsulation_Operation`  to store the encoded public key in the instance and create the RSA encryption operation during every KEM encapsulation. This would be a both a strain on memory and performance, thus I discarded this idea.~~


randombit:
> It is not apparent to me why create_encryption_op() should require an RNG in the first place – and I didn't investigate. But generally, contrary to a decryption OP, it should not require randomization in setting up precomputations due to a side channel countermeasure.

In principle blinding during encryption might be useful, and possibly that's why the argument is there. TBH I don't recall. It does seem to be the case that currently no implementation uses it, and since the encryption operation itself is provided an RNG any necessary blinding setup could just be deferred to the first invocation. Likely this argument could be removed entirely.

I would suggest rather than modifying the KEM interface in this way, take advantage of the fact that RSA's create_encryption_op ignores the RNG object anyway, pass a `Null_RNG`, catch `PRNG_Unseeded` and convert it into an `Internal_Error`.

falko-strenzke:
> I would suggest rather than modifying the KEM interface in this way, take advantage of the fact that RSA's create_encryption_op ignores the RNG object anyway, pass a `Null_RNG`, catch `PRNG_Unseeded` and convert it into an `Internal_Error`.

Absolutely makes sense, I applied that idea.

falko-strenzke:
There are still CI failures apparently  stemming from current master branch.

src/lib/pubkey/pk_keys.h
```diff
@@ -12,6 +12,7 @@
 #include <botan/pk_ops_fwd.h>
 #include <botan/secmem.h>
 
+#include <algorithm>
```
reneme: What's the reason for this new include without changing the rest of the file?

doc/api_ref/pubkey.rst
```diff
@@ -194,6 +194,21 @@ Currently two flavors of Kyber are implemented in separate Botan modules:
  * ``kyber_90s``, that uses AES/SHA-2 instead of Keccak-based primitives.
    This mode is deprecated and will be removed in a future release.
 
+ML-KEM-composite (draft-ietf-lamps-pq-composite-kem-16)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+ML-KEM-composite algorithms are accessible in Botan via the generic algorithm string
+"MLKEM-composite", which is defined in the constant `Botan::MLKEM_Composite_Param::generic_algo_name` and the respective OIDs registered
```
reneme: I'm not sure it is relevant for the user that the algorithm base name is defined as a constant somewhere internally. The string "MLKEM-composite" effectively becomes part of the public API contract anyway. Note: I'm certainly not opposing that it is defined as a constant internally, just that this bit of information isn't really useful to a user.

src/lib/pubkey/mlkem-composite/mlkem_comp_parameters.cpp
```diff
@@ -0,0 +1,369 @@
...
+      return 1184;
+   }
+   return 1568;  // must be ML-KEM-1024
+}
+
+// std::string MLDSA_Composite_Param::mlkem_param_str() const {
+//    std::string label_str(m_label);
+//    std::vector<uint8_t> label_vec(label_str.begin(), label_str.end());
+//    return std::string("Pure,Randomized,ctx_hex=") + hex_encode(label_vec);
+// }
```
reneme: Please remove commented-out code.

src/lib/pubkey/pk_algs.cpp
```diff
@@ -183,6 +187,21 @@ std::unique_ptr<Public_Key> load_public_key(const AlgorithmIdentifier& alg_id,
    }
 #endif
 
+#if defined(BOTAN_HAS_MLKEM_COMPOSITE)
+   {
+      auto mlkem_comp_param_opt = MLKEM_Composite_Param::from_algo_id(alg_id);
+      if(mlkem_comp_param_opt.has_value()) {
+         auto& comp_parm = mlkem_comp_param_opt.value();
```
reneme, line 194: Nit: typo: `comp_param`.

src/lib/pubkey/mlkem-composite/mlkem_comp_parameters.h
```diff
@@ -0,0 +1,133 @@
...
+         MLKEM768_ECDH_P384_SHA3_256,
+         MLKEM1024_RSA3072_SHA3_256,
+         MLKEM768_ECDH_brainpoolP256r1_SHA3_256,
+         MLKEM1024_ECDH_P384_SHA3_256,
+         MLKEM1024_ECDH_brainpoolP384r1_SHA3_256,
+         MLKEM1024_X448_SHA3_256,
+         MLKEM1024_ECDH_P521_SHA3_256
+      };
+
+      static inline const char* generic_algo_name = "MLKEM-Composite";
```
reneme: As this algorithm identifier string becomes part of the public API, I'd like to suggest to make it "ML-KEM-Composite" for consistency with the stand-alone name for "ML-KEM".

src/lib/pubkey/mlkem-composite/mlkem_comp_parameters.cpp
```diff
@@ -0,0 +1,369 @@
...
+#if defined BOTAN_HAS_X448
+         return true;
+#else
+         return false;
+#endif
+
+      default:
+         return false;
+   }
+}
```
reneme, line 299: I'd like to propose an alternative implementation for this, see below. This reduces the `#ifdef` jungle a little and also allows using the 'generic' pcurve for ECDH groups. Also note, that you want to check for `BOTAN_HAS_ECDH` instead of `..._ECDSA`.

```suggestion
bool MLKEM_Composite_Param::is_supported() const {
   constexpr auto supported = std::array{
#if defined(BOTAN_HAS_RSA) && defined(BOTAN_HAS_OAEP)
      MLKEM768_RSA2048_SHA3_256,
      MLKEM768_RSA3072_SHA3_256,
      MLKEM768_RSA4096_SHA3_256,
      MLKEM1024_RSA3072_SHA3_256,
#endif

#if defined(BOTAN_HAS_ECDH)
   #if defined(BOTAN_HAS_PCURVES_GENERIC) || defined(BOTAN_HAS_PCURVES_SECP256R1)
      MLKEM768_ECDH_P256_SHA3_256,
   #endif

   #if defined(BOTAN_HAS_PCURVES_GENERIC) || defined(BOTAN_HAS_PCURVES_SECP384R1)
      MLKEM768_ECDH_P384_SHA3_256,
   #endif

   #if defined(BOTAN_HAS_PCURVES_GENERIC) || defined(BOTAN_HAS_PCURVES_BRAINPOOL256R1)
      MLKEM768_ECDH_brainpoolP256r1_SHA3_256,
   #endif

   #if defined(BOTAN_HAS_PCURVES_GENERIC) || defined(BOTAN_HAS_PCURVES_SECP384R1)
      MLKEM1024_ECDH_P384_SHA3_256,
   #endif

   #if defined(BOTAN_HAS_PCURVES_GENERIC) || defined(BOTAN_HAS_PCURVES_BRAINPOOL384R1)
      MLKEM1024_ECDH_brainpoolP384r1_SHA3_256,
   #endif

   #if defined(BOTAN_HAS_PCURVES_GENERIC) || defined(BOTAN_HAS_PCURVES_SECP521R1)
      MLKEM1024_ECDH_P521_SHA3_256,
   #endif
#endif

#if defined(BOTAN_HAS_X25519)
      MLKEM768_X25519_SHA3_256,
#endif

#if defined(BOTAN_HAS_X448)
      MLKEM1024_X448_SHA3_256,
#endif
   };

   return std::find(supported.begin(), supported.end(), m_id) != supported.end();
}
```

reneme (commented):
Not a full review. Just some up-front comments.

src/lib/pubkey/pk_keys.h
falko-strenzke: Must have been my IDE, slipped my attention. Fixed in next commit.

doc/api_ref/pubkey.rst
falko-strenzke: Agreed, I removed that clause.

src/lib/pubkey/mlkem-composite/mlkem_comp_parameters.h
falko-strenzke: Right, but on the other hand the composite drafts consistently use the string "MLDSA" and "MLKEM" which I think here is also relevant for the context. But I tend to agree with you, and will change it accordingly.

src/lib/ffi/ffi_pkey_algs.cpp
```diff
@@ -1348,6 +1352,63 @@ int botan_pubkey_load_ml_kem(botan_pubkey_t* key, const uint8_t pubkey[], size_t
...
+*/
+
+int botan_privkey_load_mlkem_composite(botan_privkey_t* key,
+                                       const uint8_t privkey[],
+                                       size_t key_len,
+                                       const char* mlkem_composite_algo) {
+#if defined(BOTAN_HAS_MLKEM_COMPOSITE)
+   if(key == nullptr || privkey == nullptr || mlkem_composite_algo == nullptr) {
+      return BOTAN_FFI_ERROR_NULL_POINTER;
+   }
```
reneme: Nit: There's a helper that could be used here:


```suggestion
   if(Botan::any_null_pointers(key, privkey, mlkem_composite_algo)) {
      return BOTAN_FFI_ERROR_NULL_POINTER;
   }
```

src/lib/ffi/ffi_pkey_algs.cpp
```diff
@@ -1348,6 +1352,63 @@ int botan_pubkey_load_ml_kem(botan_pubkey_t* key, const uint8_t pubkey[], size_t
...
+#endif
+}
+
+int botan_pubkey_load_mlkem_composite(botan_pubkey_t* key,
+                                      const uint8_t pubkey[],
+                                      size_t key_len,
+                                      const char* mlkem_composite_algo) {
+#if defined(BOTAN_HAS_MLKEM_COMPOSITE)
+   if(key == nullptr || pubkey == nullptr || mlkem_composite_algo == nullptr) {
+      return BOTAN_FFI_ERROR_NULL_POINTER;
```
reneme: Same as above:


```suggestion
   if(Botan::any_null_pointers(key, pubkey, mlkem_composite_algo)) {
      return BOTAN_FFI_ERROR_NULL_POINTER;
```

src/lib/pubkey/mlkem-composite/info.txt
```diff
@@ -0,0 +1,22 @@
...
+<header:public>
+mlkem_comp.h
+mlkem_comp_parameters.h
+</header:public>
+
+<header:internal>
+</header:internal>
+
+<requires>
+kyber
```
reneme: Please pull in `ml_kem` instead.

src/lib/pubkey/mlkem-composite/mlkem_comp.h
```diff
@@ -0,0 +1,154 @@
...
+#include <botan/kyber.h>
+#include <botan/ml_dsa.h>
+#include <botan/mlkem_comp_parameters.h>
+#include <botan/pk_keys.h>
+#include <botan/pk_ops_fwd.h>
+#include <memory>
+#include <span>
+
+namespace Botan {
+class BOTAN_PUBLIC_API(3, 0) MLKEM_Composite_PublicKey : public virtual Public_Key {
```
reneme:
```suggestion
class BOTAN_PUBLIC_API(3, 13) MLKEM_Composite_PublicKey : public virtual Public_Key {
```

... I guess.

src/lib/pubkey/mlkem-composite/mlkem_comp.h
```diff
@@ -0,0 +1,154 @@
...
+   public:
+      /**
+       * Loads a public key.
+       *
+       * @param key_bits DER encoded public key bits
+       */
+      BOTAN_FUTURE_EXPLICIT MLKEM_Composite_PublicKey(MLKEM_Composite_Param::id_t id,
```
reneme, line 34: No need to mark this as `FUTURE_EXPLICIT` given that it is a newly introduced API. That macro was only introduced to avoid semantic versioning issues of existing API.

Also, if the constructor takes more than one parameter, `explicit` isn't necessary at all, IMO.

... also applies to the constructor below.


```suggestion
      MLKEM_Composite_PublicKey(MLKEM_Composite_Param::id_t id,
```

src/lib/pubkey/mlkem-composite/mlkem_comp.h
```diff
@@ -0,0 +1,154 @@
...
+                                                      std::span<const uint8_t> key_bits);
+
+      BOTAN_FUTURE_EXPLICIT MLKEM_Composite_PublicKey(const AlgorithmIdentifier& algo_id,
+                                                      std::span<const uint8_t> key_bits);
+
+      std::string algo_name() const override { return "MLKEM_Composite"; }
```
reneme: This algorithm name should be equal to `MLKEM_Composite_Param::generic_algo_name`.

src/lib/pubkey/mlkem-composite/mlkem_comp_parameters.h
```diff
@@ -0,0 +1,132 @@
...
+#ifndef BOTAN_MLKEM_COMP_PARAMETERS_H_
+#define BOTAN_MLKEM_COMP_PARAMETERS_H_
+
+#include <botan/exceptn.h>
+#include <botan/ml_kem.h>
+#include <botan/types.h>
+
+namespace Botan {
+
+class BOTAN_PUBLIC_API(3, 0) MLKEM_Composite_Param {
```
reneme:
```suggestion
class BOTAN_PUBLIC_API(3, 13) MLKEM_Composite_Param {
```

src/lib/pubkey/mlkem-composite/mlkem_comp_parameters.h
```diff
@@ -0,0 +1,132 @@
...
+                            const char* traditional_algorithm,
+                            const char* traditional_padding,
+                            const char* curve,
+                            uint32_t traditional_key_size) noexcept;
+
+      const char* m_id_str;
+      const char* m_label;
+      const char* m_traditional_algorithm;
+      const char* m_traditional_padding;
+      const char* m_curve;
```
reneme: Without having looked deeper into the implementation details: Wouldn't `std::string_view` be a better fit for those raw char pointers?

reneme (commented):
Sorry for the fairly shallow high-level comments again. I hope I'll have more time to do a proper review in the next few days.

src/lib/pubkey/mlkem-composite/mlkem_comp.cpp
```diff
@@ -0,0 +1,597 @@
...
+#include <optional>
+#include <string_view>
+#include <variant>
+#include <vector>
+
+namespace Botan {
+
+namespace {
+
+// with C++23, to be replaced by std::ranges::contains()
```
reneme: I appreciate the guidance for future improvement! Conventionally we formulate it like so (for easy grep-ability):


```suggestion
// TODO(C++23): to be replaced by std::ranges::contains()
```

src/lib/pubkey/mlkem-composite/info.txt
reneme, line 1: Commenting on this file for the lack of a better place: Traditionally, we named our modules with underscores not dashes. For consistency: could you rename this module to `mlkem_composite`, please? I.e. the parent directory of this file.

reneme:
@falko-strenzke On a general note: We should look into consolidating support for this IETF draft with the existing facilities in the `pubkey/hybrid_kem` and `pubkey/kex_to_kem_adapter` modules which are currently used in the TLS 1.3 PQC key exchange. Those provide base classes for hybrid KEMs already, and we shouldn't duplicate this logic here.

Following this idea, your MLKEM-Composite implementation should derive from `Hybrid_PublicKey` and `Hybrid_PrivateKey`. Translation from the ECDH/X25519/X448 algorithms to a KEM interface can be done conveniently using the `KEX_to_KEM_Adapter`. That way you should be able to avoid all dynamic casts to concrete ECDH public key classes such as `ECDH_PublicKey`, or `X25519_PublicKey`. This doesn't cover RSA-OAEP, but a similar adapter construction could be built for this.

The usage of generic adapters (for RSA and for ECDH) allow the composition to be implemented exclusively based on KEMs, without explicit handling of specific traditional algorithms. _Also, such an implementation could be easily extensible to other KEMs in the future (e.g. FrodoKEM or HQC)_.

The existing `hybrid_kem` module is an internal interface and can thus be adapted as needed to create a decent foundation for both TLS's needs as well as this draft's implementation. Most notably, at the moment it is designed to be super generic by allowing arbitrary combinations of one-to-many PQ and traditional KEMs (not just specifically 1xPQ and 1xTraditional). In retrospect, this was probably a design mistake that we should fix.

I would suggest a design like so:

Screenshot 2026-06-30 at 13 28 47

[excalidraw.com](https://excalidraw.com/#json=Uqp4TN5nbjt3gJvG6K9_C,bPNpnRPVp7fpKyoWyW5pBA)

I could take over the relevant work in the mentioned base classes and the TLS hybrid implementation.

FAlbertDev:
@falko-strenzke 
One feature that would be very interesting for our use case is the ability to use a custom implementation of the underlying algorithms. For example, we plan to use Composite ML-KEM in combination with smart cards where the ML-KEM decapsulation and the ECDH operation are performed by the smart card. For that, we would use underlying keys of type PKCS11_EC_PrivateKey and (not yet implemented, but something like) PKCS11_ML_KEM_PrivateKey.

Therefore, it would be really nice if the interface of the composite ML-KEM private key would allow the processing of such "custom" private keys. I could imagine adding an additional constructor like:
```c++
MLKEM_Composite_PrivateKey(MLKEM_Composite_Param::id_t id, std::unique_ptr<Botan::PrivateKey> mlkem_privkey, std::unique_ptr<Botan::PrivateKey> traditional_privkey);
```
The information, currently obtained via dynamic casting the private key pointers, must be derived from the MLKEM_Composite_Param in this case. What do you think?

reneme:
@falko-strenzke See #5715 please. I'm currently working on a quick'n'dirty retrofit of your work on this branch onto the changes in #5715 to see if it provides the required flexibility to accommodate this implementation of the LAMPS draft. I'll share it with you once I have something working.

src/lib/pubkey/mlkem-composite/mlkem_comp.h
```diff
@@ -0,0 +1,154 @@
...
+                                                                       std::string_view provider) const override;
+
+      MLKEM_Composite_PublicKey(const MLKEM_Composite_PublicKey& other);
+
+      MLKEM_Composite_PublicKey& operator=(const MLKEM_Composite_PublicKey& rhs);
+
+      ~MLKEM_Composite_PublicKey() override = default;
+
+      MLKEM_Composite_PublicKey(const MLKEM_Composite_PublicKey&& other) = delete;
+      MLKEM_Composite_PublicKey& operator=(const MLKEM_Composite_PublicKey&& rhs) = delete;
```
reneme:
```suggestion
      MLKEM_Composite_PublicKey(MLKEM_Composite_PublicKey&& other) = delete;
      MLKEM_Composite_PublicKey& operator=(MLKEM_Composite_PublicKey&& rhs) = delete;
```

... otherwise those don't qualify as move constructors.

src/lib/pubkey/mlkem-composite/mlkem_comp.h
```diff
@@ -0,0 +1,154 @@
...
+                                                                       std::string_view provider) const override;
+
+      MLKEM_Composite_PrivateKey(const MLKEM_Composite_PrivateKey& other);
+
+      MLKEM_Composite_PrivateKey& operator=(const MLKEM_Composite_PrivateKey& rhs);
+
+      ~MLKEM_Composite_PrivateKey() override = default;
+
+      MLKEM_Composite_PrivateKey(const MLKEM_Composite_PrivateKey&& other) = delete;
+      MLKEM_Composite_PrivateKey& operator=(const MLKEM_Composite_PrivateKey&& rhs) = delete;
```
reneme:
```suggestion
      MLKEM_Composite_PrivateKey(MLKEM_Composite_PrivateKey&& other) = delete;
      MLKEM_Composite_PrivateKey& operator=(MLKEM_Composite_PrivateKey&& rhs) = delete;
```

... otherwise those don't qualify as move constructors.

src/lib/pubkey/mlkem-composite/mlkem_comp.h
```diff
@@ -0,0 +1,154 @@
...
+
+      MLKEM_Composite_PublicKey(const MLKEM_Composite_PublicKey&& other) = delete;
+      MLKEM_Composite_PublicKey& operator=(const MLKEM_Composite_PublicKey&& rhs) = delete;
+
+   protected:
+      static std::shared_ptr<Public_Key> load_traditional_public_key(const MLKEM_Composite_Param& param,
+                                                                     std::span<const uint8_t> key_bits);
+      MLKEM_Composite_PublicKey() = default;
+
+      std::shared_ptr<MLKEM_Composite_Param> m_parameters;  // NOLINT(*non-private-member-variable*)
```
reneme: Is there a concrete reason that the parameters are kept as a shared pointer? The parameters objects doesn't strike me as expensive to copy. Let's just hold it as an ordinary member without any smart pointer.

src/lib/pubkey/mlkem-composite/mlkem_comp.h
```diff
@@ -0,0 +1,154 @@
...
+
+   private:
+      static std::shared_ptr<Private_Key> load_traditional_private_key(const MLKEM_Composite_Param& param,
+                                                                       std::span<const uint8_t> key_bits);
+      static std::unique_ptr<Private_Key> create_traditional_private_key(RandomNumberGenerator& rng,
+                                                                         MLKEM_Composite_Param param);
+      void init_pubkey_members();
+
+      secure_vector<uint8_t> encode_traditional_private_key() const;
+      friend class MLKEM_Composite_Signature_Operation;
```
reneme: Copy-pasta? 🙂 


```suggestion
```

src/lib/pubkey/mlkem-composite/mlkem_comp.h
reneme: Generally, a class that can be copied but not moved seems to be rather unconventional. Any particular reason for that? Same is true for the associated public key class.

src/lib/pubkey/mlkem-composite/mlkem_comp.h
```diff
@@ -0,0 +1,154 @@
...
+      MLKEM_Composite_PublicKey& operator=(const MLKEM_Composite_PublicKey& rhs);
+
+      ~MLKEM_Composite_PublicKey() override = default;
+
+      MLKEM_Composite_PublicKey(const MLKEM_Composite_PublicKey&& other) = delete;
+      MLKEM_Composite_PublicKey& operator=(const MLKEM_Composite_PublicKey&& rhs) = delete;
+
+   protected:
+      static std::shared_ptr<Public_Key> load_traditional_public_key(const MLKEM_Composite_Param& param,
+                                                                     std::span<const uint8_t> key_bits);
```
reneme: I don't see any need why this function should be declared in a public header. Let's just make it an implementation detail in the cpp file, along those lines:

```C++
namespace {

std::shared_ptr<Public_Key> load_tradtional_public_key(/*...*/) {/* ... */}

}
```

src/lib/pubkey/mlkem-composite/mlkem_comp.h
```diff
@@ -0,0 +1,154 @@
...
+      ~MLKEM_Composite_PrivateKey() override = default;
+
+      MLKEM_Composite_PrivateKey(const MLKEM_Composite_PrivateKey&& other) = delete;
+      MLKEM_Composite_PrivateKey& operator=(const MLKEM_Composite_PrivateKey&& rhs) = delete;
+
+   private:
+      static std::shared_ptr<Private_Key> load_traditional_private_key(const MLKEM_Composite_Param& param,
+                                                                       std::span<const uint8_t> key_bits);
+      static std::unique_ptr<Private_Key> create_traditional_private_key(RandomNumberGenerator& rng,
+                                                                         MLKEM_Composite_Param param);
```
reneme: As in the public key: there's no need for those functions to be declared publicly. Just declare them in an anonymous namespace in the cpp file please.

src/lib/pubkey/mlkem-composite/mlkem_comp.cpp
```diff
@@ -0,0 +1,606 @@
...
+      return std::shared_ptr<Private_Key>(std::make_shared<X25519_PrivateKey>((X25519_PrivateKey(trad_key_bits))));
+   }
+#endif
+#if defined(BOTAN_HAS_X448)
+   if(param.traditional_algorithm() == "X448") {
+      return std::shared_ptr<Private_Key>(std::make_shared<X448_PrivateKey>((X448_PrivateKey(trad_key_bits))));
+   }
+#endif
+
+   return load_private_key(param.get_traditional_algorithm_id(), trad_key_bits);
```
reneme: I'm somewhat surprised that the `load_private_key()` approach of tail-case doesn't work for the X25519/X448 cases as well. To my mind there's no real reason that those need to be handled in a special way.

Perhaps I'm missing something, but it might be worth looking into if there's some issue in `load_private_key` that prevents you from using it vanilla.

src/lib/pubkey/mlkem-composite/mlkem_comp.h
```diff
@@ -0,0 +1,154 @@
...
+
+      ~MLKEM_Composite_PublicKey() override = default;
+
+      MLKEM_Composite_PublicKey(const MLKEM_Composite_PublicKey&& other) = delete;
+      MLKEM_Composite_PublicKey& operator=(const MLKEM_Composite_PublicKey&& rhs) = delete;
+
+   protected:
+      static std::shared_ptr<Public_Key> load_traditional_public_key(const MLKEM_Composite_Param& param,
+                                                                     std::span<const uint8_t> key_bits);
+      MLKEM_Composite_PublicKey() = default;
```
reneme: A default constructed object of this class is inherently in a broken state. The constructor is protected, so there's no strong chance that a user could accidentally call this constructor, but nevertheless it is a serious code smell in my opinion.

Instead, please set up a proper constructor that takes the required values to set up this class in a valid state from the get-go. That way you will also be able to avoid the private `init_pubkey_members()` function in the private key class. Instead you can just up-call the proper public key constructor in the private key constructors.

src/lib/pubkey/mlkem-composite/mlkem_comp.h
```diff
@@ -0,0 +1,154 @@
...
+
+      MLKEM_Composite_PrivateKey(const MLKEM_Composite_PrivateKey&& other) = delete;
+      MLKEM_Composite_PrivateKey& operator=(const MLKEM_Composite_PrivateKey&& rhs) = delete;
+
+   private:
+      static std::shared_ptr<Private_Key> load_traditional_private_key(const MLKEM_Composite_Param& param,
+                                                                       std::span<const uint8_t> key_bits);
+      static std::unique_ptr<Private_Key> create_traditional_private_key(RandomNumberGenerator& rng,
+                                                                         MLKEM_Composite_Param param);
+      void init_pubkey_members();
```
reneme: Objects should always be "valid on construction" (see [C++ Core Guidelines C.41](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rc-complete)). Any init members are a serious code smell if there's no real technical reason behind it, in my opinion. With a proper constructor in the public key class (as mentioned above) you could fully avoid this init function.

Along those lines:

```C++
MLKEM_Composite_PrivateKey::MLKEM_Composite_PrivateKey(...)
  : MLKEM_Composite_PublicKey(params, t_pubkey, pq_pubkey),
  , m_parameters(....)
  , m_mlkem_privkey(...)
  , m_traditional_privkey(....) {}
```

... admittedly, Botan's crooked class hierarchy where private keys inherit from public keys makes this harder. A good way to work around most complications is creating a single low-level constructor, that takes exactly the required values and distributes them both into the base class(es) and the object's members, and then building higher-level constructors that delegate onto this low-level constructor. Like so (pretty sketchy):

```C++
class MLKEM_Composite_PrivateKey {
   public:
      // The "low-level" constructor that sets everything up based on
      // exactly the values we need for initialization.
      MLKEM_Composite_PrivateKey(const MLKEM_Composite_Param& param, 
                                 std::unique_ptr<ML_KEM_PrivateKey> mlkem_sk,
                                 std::unique_ptr<Private_Key> trad_sk)
         : MLKEM_Composite_PublicKey(param, mlkem_sk->public_key(), trad_sk->public_key())
         , m_paramters(std::make_shared<MLKEM_Composite_Param>(param))
         , m_mlkem_sk(std::move(mlkem_sk))
         , m_trad_sk(std::move(trad_sk)) {}

      MLKEM_Composite_PrivateKey(RandomNumberGenerator& rng, MLKEM_Composite_Param param)
         : MLKEM_Composite_PrivateKey(param, 
                                      create_mlkem_private_key(param, rng),
                                      create_traditional_private_key(param, rng)) {}
      
      MLKEM_Composite_PrivateKey(const AlgorithmIdentifier& alg_id, std::span<const uint8_t> sk)
         : MLKEM_Composite_PrivateKey(MLKEM_Composite_Param::from_algo_id_or_throw(alg_id),
                                      load_mlkem_private_key(MLKEM_Composite_Param::from_algo_id_or_throw(alg_id), sk),
                                      load_traditional_private_key(MLKEM_Composite_Param::from_algo_id_or_throw(alg_id), sk)) {}
};
```

... coincidentally this low-level constructor is also roughly what @FAlbertDev [was asking for](https://github.com/randombit/botan/pull/5686#issuecomment-4892536386) as an escape hatch for customizable (e.g. hardware-backed) component keys.

src/lib/pubkey/mlkem-composite/mlkem_comp.cpp
```diff
@@ -0,0 +1,606 @@
...
+OID MLKEM_Composite_PublicKey::object_identifier() const {
+   return m_parameters->object_identifier();
+}
+
+std::vector<uint8_t> MLKEM_Composite_PublicKey::public_key_bits() const {
+   std::vector<uint8_t> result(this->m_mlkem_pubkey->public_key_bits());
+   std::vector<uint8_t> trad_bytes = this->m_traditional_pubkey->public_key_bits();
+   result.insert(result.end(), trad_bytes.begin(), trad_bytes.end());
+   return result;
+}
```
reneme: Along those lines:


```suggestion
std::vector<uint8_t> MLKEM_Composite_PublicKey::public_key_bits() const {
   return concat(this->m_mlkem_pubkey->public_key_bits(),
                 this->m_traditional_pubkey->public_key_bits());
}
```

src/lib/pubkey/mlkem-composite/mlkem_comp.cpp
```diff
@@ -0,0 +1,606 @@
...
+
+secure_vector<uint8_t> MLKEM_Composite_PrivateKey::private_key_bits() const {
+   secure_vector<uint8_t> result =
+      m_mlkem_privkey
+         ->raw_private_key_bits();  // "raw_...()" should still return the raw seed even after fixing the PKCS#8 encoding format for ML-KEM
+   secure_vector<uint8_t> trad_bytes = encode_traditional_private_key();
+   result.insert(result.end(), trad_bytes.begin(), trad_bytes.end());
+   return result;
+}
```
reneme: Along those lines:


```suggestion
secure_vector<uint8_t> MLKEM_Composite_PrivateKey::private_key_bits() const {
   return concat(m_mlkem_privkey->raw_private_key_bits(),
                 encode_traditional_private_key());
}
```

src/lib/pubkey/mlkem-composite/info.txt
```diff
@@ -0,0 +1,22 @@
...
+rng
+hash
+asn1
+</requires>
```
reneme, line 23: Instead of depending on `hash` this should actually depend on sha3 specifically, because that's what it uses directly. The generic `hash` module will be pulled in transitively.


```suggestion
<requires>
kyber
rng
sha3
asn1
</requires>
```

src/lib/pubkey/mlkem-composite/mlkem_comp.cpp
```diff
@@ -0,0 +1,606 @@
...
+                                                       const Private_Key* trad_privkey,
+                                                       std::string_view provider) :
+            m_parameters(param),
+            m_mlkem_dec_op(mlkem_privkey.create_kem_decryption_op(rng, "Raw", "")),
+            m_traditional_pubkey_encoded(trad_privkey->public_key()->public_key_bits()) {
+         if(trad_privkey->algo_name() == "RSA") {
+            m_traditional_dec_op.emplace<std::unique_ptr<PK_Ops::Decryption>>(
+               trad_privkey->create_decryption_op(rng, param.traditional_padding(), provider));
+            // For some reason, the RSA decryption OP can't tell us it's ciphertext size. Thus we have to delve into the details...
+            m_rsa_modulus_bytes = dynamic_cast<RSA_PublicKey*>(trad_privkey->public_key().get())->get_n().bytes();
```
reneme: See #5717.

reneme (changes requested):

reneme:
@falko-strenzke I took the liberty and refactored your patch (mostly `mlkem_comp.cpp`) based on #5715 and #5717. See here: https://github.com/reneme/botan/tree/falko/mlkem-composite (squashed to one commit and based on #5715).

This also addresses plenty of the suggestions I added above. But most notably, it gets rid of almost all low-level algorithm-specific handling by reusing existing concepts.

I'd appreciate if you could have a read-through my refactorings and pull them into this PR.

falko-strenzke:
> @falko-strenzke I took the liberty and refactored your patch (mostly `mlkem_comp.cpp`) based on #5715 and #5717. See here: https://github.com/reneme/botan/tree/falko/mlkem-composite (squashed to one commit and based on #5715).
> 
> This also addresses plenty of the suggestions I added above. But most notably, it gets rid of almost all low-level algorithm-specific handling by reusing existing concepts.
> 
> I'd appreciate if you could have a read-through my refactorings and pull them into this PR.

@reneme As I wrote in #5715, I am not entirely convinced of the multi-algorithm framework that you rebased the implementation onto. As a plus, it reduces the code size in `mlkem_comp.cpp` by one third, on the negative side, the trivial combination of two KEMs becomes somewhat harder to read due to a concept that doesn't actually bring much functionaility. In any case, I am fine with your proposal, but I would like to hear from @randombit first before pulling it in. The decision to use the multi-algorithm framework throughout the library in long term seems to be fundamental one.

falko-strenzke:
> @falko-strenzke One feature that would be very interesting for our use case is the ability to use a custom implementation of the underlying algorithms. For example, we plan to use Composite ML-KEM in combination with smart cards where the ML-KEM decapsulation and the ECDH operation are performed by the smart card. For that, we would use underlying keys of type PKCS11_EC_PrivateKey and (not yet implemented, but something like) PKCS11_ML_KEM_PrivateKey.
> 
> Therefore, it would be really nice if the interface of the composite ML-KEM private key would allow the processing of such "custom" private keys. I could imagine adding an additional constructor like:
> 
> ```c++
> MLKEM_Composite_PrivateKey(MLKEM_Composite_Param::id_t id, std::unique_ptr<Botan::PrivateKey> mlkem_privkey, std::unique_ptr<Botan::PrivateKey> traditional_privkey);
> ```
> 
> The information, currently obtained via dynamic casting the private key pointers, must be derived from the MLKEM_Composite_Param in this case. What do you think?

Well, with @reneme 's [new proposal](https://github.com/randombit/botan/pull/5686#issuecomment-4915869830), I think that would or should be taken care of by the hybridization framework. @reneme Is that already the case?

reneme:
> Is that already the case?

Pretty much. There are two central things to that:

1. Have a low-level constructor that takes any keypair objects from the user,
2. Refrain from any assumptions of the concrete key types (i.e. don't use `dynamic_cast<RSA_PublicKey>` for instance), because for a custom implementation that won't work.

The second point might need some work. The first point should be taken care of already.

sehlen-bsi:
For the RSA composites, there seems to be the same issue as mentioned for #5451 that the RSA key sizes are not bound to the modulus size specified by the composite OID. I might be wrong here, please check!

falko-strenzke:
> For the RSA composites, there seems to be the same issue as mentioned for #5451 that the RSA key sizes are not bound to the modulus size specified by the composite OID. I might be wrong here, please check!

You are correct as far as I can see, but I first want to adopt @reneme 's refactoring before addressing it.

falko-strenzke:
> > @falko-strenzke I took the liberty and refactored your patch (mostly `mlkem_comp.cpp`) based on #5715 and #5717. See here: https://github.com/reneme/botan/tree/falko/mlkem-composite (squashed to one commit and based on #5715).
> > This also addresses plenty of the suggestions I added above. But most notably, it gets rid of almost all low-level algorithm-specific handling by reusing existing concepts.
> > I'd appreciate if you could have a read-through my refactorings and pull them into this PR.
> 
> @reneme As I wrote in #5715, I am not entirely convinced of the multi-algorithm framework that you rebased the implementation onto. As a plus, it reduces the code size in `mlkem_comp.cpp` by one third, on the negative side, the trivial combination of two KEMs becomes somewhat harder to read due to a concept that doesn't actually bring much functionaility. In any case, I am fine with your proposal, but I would like to hear from @randombit first before pulling it in. The decision to use the multi-algorithm framework throughout the library in long term seems to be fundamental one.


I just made a diff of our branches. After seeing many unrelated differences, I rebased on master (where I rather should have rebased onto #5715, as I saw afterwards), and still have many unrelated differences. Coud you @reneme rebase your branch on current master (4695b4a661af36714c6a323d0c345b16ac2facc8 is what this PR is based on now), too , so that we have a clean diff?

reneme:
> Coud you rebase your branch on current master?

I rebased #5715 onto a8f6c6a59e850016f77dd2a1d3d10991d2610eba

src/lib/ffi/ffi_pkey_algs.cpp
falko-strenzke: Done

src/lib/ffi/ffi_pkey_algs.cpp
falko-strenzke: Done

src/lib/pubkey/mlkem-composite/info.txt
falko-strenzke: Done

src/lib/pubkey/mlkem-composite/info.txt
falko-strenzke: Done

src/lib/pubkey/mlkem-composite/mlkem_comp.h
falko-strenzke: Was in you branch that I adopted.

src/lib/pubkey/mlkem-composite/mlkem_comp_parameters.h
falko-strenzke: Done

src/lib/pubkey/mlkem-composite/mlkem_comp.h
falko-strenzke: Done

src/lib/pubkey/mlkem-composite/mlkem_comp_parameters.h
falko-strenzke: Done by adopting your branch with the hybrid KEM refactoring.

src/lib/pubkey/mlkem-composite/mlkem_comp.cpp
falko-strenzke: I think done via your changes I adopted.

src/lib/pubkey/mlkem-composite/mlkem_comp.h
falko-strenzke: I think you removed those functions anyway.

src/lib/pubkey/mlkem-composite/mlkem_comp.h
falko-strenzke: Resolved through the adoptio nof the hybrid kem framework.

src/lib/pubkey/mlkem-composite/mlkem_comp.cpp
falko-strenzke: Resolved via said refactoring.

src/lib/pubkey/mlkem-composite/mlkem_comp.cpp
falko-strenzke: Resolved via said refactoring to use hybrid KEM framework.

src/lib/pubkey/mlkem-composite/mlkem_comp.cpp
falko-strenzke: Resolved via said refactoring to use hybrid KEM framework.

src/lib/pubkey/mlkem-composite/mlkem_comp.h
falko-strenzke: Done.

src/lib/pubkey/mlkem-composite/mlkem_comp.h
falko-strenzke: Resolved via said refactoring to use hybrid KEM framework which removes them.

src/lib/pubkey/mlkem-composite/mlkem_comp.h
falko-strenzke: Right, done.

falko-strenzke:
> > Coud you rebase your branch on current master?
> 
> I rebased #5715 onto [a8f6c6a](https://github.com/randombit/botan/commit/a8f6c6a59e850016f77dd2a1d3d10991d2610eba)

@reneme 

Yes, but  not https://github.com/reneme/botan/tree/falko/mlkem-composite as far as I could see..

But no problem, I now adopted the changes from that branch and should now have addressed all your review comments. Maybe you want to look over it again.

falko-strenzke:
> > For the RSA composites, there seems to be the same issue as mentioned for #5451 that the RSA key sizes are not bound to the modulus size specified by the composite OID. I might be wrong here, please check!
> 
> You are correct as far as I can see, but I first want to adopt @reneme 's refactoring before addressing it.

Now done. We have the same key size and EC group checks as for ML-DSA composites.

PR #5690, opened 2026-06-23
Add and fix XMSS(^MT) Certificate Algorithm OIDs

falko-strenzke:
This is a pull request on top of xmssmt-dev, which introduces XMSS^MT. It adds the missing certificate validation for XMSS^MT and adds the official XMSS OID from RFC 9802.

@TJ-91 I got the test certificates from RFC 9802 to run with
```
./botan-test x509_path_xmss
```
~~However, I had to modify the public key decoding. I assume it was previously simply arbitrary to have the OCTET STRING encoding around it. However, the existing XMSS^MT tests are now failing, probably due to~~
- ~~ncompatible public key *encoding* ((naturally, since I fixed only the decoding)~~
- ~~existing test data.~~

Please review this ~~and fix the tests.~~

(Public key encoding and tests are fixed now.)

I also added the XMSS test certificate from RFC 9802 for which I had to add the official XMSS OID. Probably that should become the default one, i.e., be registered as "XMSS". But I don't know who or what is dependend on the inofficial XMSS OID that is currently registered as "XMSS", so I left it as it is.


TJ-91:
I have cherry-picked your commits to #5358

falko-strenzke:
> I have cherry-picked your commits to #5358

So we can close this one.

PR #5691, opened 2026-06-24
[TLS 1.3] Support for Brainpool ECDH code-points as defined in RFC8734

reneme:
This is a revamp of https://github.com/randombit/botan/pull/3810, it essentially follows [my suggestion for simplification](https://github.com/randombit/botan/pull/3810#issuecomment-1847116717):

>  Let's introduce the new code points and don't do any additional handling or validation. That will result in Botan allowing to mix and match protocol version and code point revisions. And this might result in interop issues with other implementations. But it gives the library user full flexibility to work around interop issues for their specific use case.

The old brainpool code points (26-28) were deprecated in the TLS 1.3 RFC (for non-widespread usage). So, technically, they should not be used with TLS 1.3. Instead, [RFC8734](https://www.rfc-editor.org/info/rfc8734/) re-introduced new code points (31-33) that are meant for TLS 1.3 explicitly. Cryptographically and usage-wise those are identical. The new code points are identified using a "tls13" suffix after "brainpool256r1" for instance. This approach is used in other implementations, too.

This patch does not introduce any special validation for either of the code point variants: i.e. they both may be used for both TLS versions, which is arguably too liberal, but provides maximum configuration flexibility. Given my assumption that brainpool is typically used only in very specific use-cases, that should probably suffice functionality-wise.

Note that neither the old nor the new code points in TLS 1.3 are currently covered by any unit or integration tests, as far as I know. However, we successfully tested both against `openssl s_server`. `unit_tls.cpp` would be a good candidate to end-to-end test such things, but its currently only fit for TLS 1.2. We should consider extending this independently.


randombit (approved):
Probably about the best we can reasonably do here. Implementation seems fine.

Issue #4782, opened 2025-03-21
DTLS session resumption; frequent "Can't interleave application and handshake data" errors


oviano:
When my DTLS client connects to the server over a certain distance/latency (around 100-200ms latency) I have found the above error occurs a lot, but only when session resumption is enabled and not always. It can go through periods were repeated handshakes fail in this manner, and other times when it works just fine.

With session resumption disabled (by overriding tls_should_persist_resumption_information to return false, the issue apparently disappears completely, and the full handshake succeeds.

At first I thought I was simply seeing this issue https://github.com/randombit/botan/issues/2498 but it turns out on closer inspection that neither the client or server are retransmitting anything during these failed handshakes.

oviano:
Ok here is some more detailed log, basically dumping the contents of the DTLS headers on both client and server in the order they are sent and received:

First the client:

```
--------------------------------------------------
SENDING DTLS DATA
--------------------------------------------------
DTLS Record Header
16fefd00000000000000000161
Message Type    : Handshake
Protocol Version: 65022
epoch           : 0
Sequence Number : 0
Length          : 353
--------------------------------------------------
--------------------------------------------------
RECEIVING TLS DATA
--------------------------------------------------
DTLS Record Header
16fefd00000000000000000063
Message Type    : Handshake
Protocol Version: 65022
epoch           : 0
Sequence Number : 0
Length          : 99
--------------------------------------------------
--------------------------------------------------
RECEIVING TLS DATA
--------------------------------------------------
DTLS Record Header
14fefd00000000000000010001
Message Type    : ChangeCipherSpec
Protocol Version: 65022
epoch           : 0
Sequence Number : 1
Length          : 1
--------------------------------------------------
--------------------------------------------------
RECEIVING TLS DATA
--------------------------------------------------
DTLS Record Header
16fefd00010000000000000030
Message Type    : Handshake
Protocol Version: 65022
epoch           : 1
Sequence Number : 0
Length          : 48
--------------------------------------------------
TLS_CALLBACKS: Started TLS session (version DTLS v1.2, ciphersuite ECDHE_ECDSA_WITH_AES_256_GCM_SHA384, was_resumption true)
TLS_CALLBACKS: Updated TLS resumption information
TLS_CALLBACKS: Activated TLS session
--------------------------------------------------
SENDING DTLS DATA
--------------------------------------------------
DTLS Record Header
14fefd00000000000000010001
Message Type    : ChangeCipherSpec
Protocol Version: 65022
[...]

PR #5696, opened 2026-06-25
Fix DTLS handshake retransmission edge cases

### LINKED ISSUE #2310: DTLS last write update?
When calling timeout_check() on DTLS channel, there is this piece of code in Datagram_Handshake_IO
```
   const uint64_t ms_since_write = steady_clock_ms() - m_last_write;

   if(ms_since_write < m_next_timeout)
      return false;

   retransmit_last_flight();
```

retransmit_last_flight later calls send_message. However m_last_write gets updated only in send_under_epoch (which also uses send_message). 

In the consequence the message is retransmitted (almost) every time you call timeout_check.

### LINKED ISSUE #2498: The dTLS server (1.2) is not handling properly the re-transmissions of flight 5 from the client

### LINKED ISSUE #4022: Facing retransmission issue while adding delay in the server side in the DTLS implementation
Botan version used: 2.19.3
We are trying to implement client server communication through DTLS protocol on top of botan library. In Botan::TLS::Callbacks tls_emit_data() we are inducing a delay of 5 seconds when the engine is server to test our implementation for the retransmission behavior. When we do this, the client is sending the ClientKeyExchange Flight before it receives ServerHelloDone flight from the server. We could also see partial read in the client side as server has sent ServerHello alone but not ServerHelloDone.

```
void
CallbacksPSK::tls_emit_data(uint8_t const buf[], size_t const length)
{
    if (nullptr == mEngine) {
        throw std::runtime_error("Null engine pointer!");
    }
   if(mEngine->mIsServer){
       sleep(5);  
   }
}
```

When we do this the handshake is not getting completed and the server is unable to process the request from client ands it throws an error:
**terminate called after throwing an instance of 'Botan::Internal_Error'
  what():  Internal error: TLS::Channel No write cipherstate for epoch 65535
Aborted (core dumped)**

The same delay when we add it for client side the handshake is getting completed.

1. Is this the right way to add delay?
2. Is adding delay affects the handshake process ?

### LINKED ISSUE #4036: The dTLS server (1.2) is not handling properly the re-transmissions in case of delay in flight we see clientKeyExchange before serverKeyExchange and serverHelloDone
We tried to test a scenario that the server has a delay before sending message, so we expect the client to do retransmission, but it seems like the client sends the next flight before getting all the flight for the server (We can see clientKeyExchange before serverKeyExchange and serverHelloDone).

```
void
CallbacksPSK::tls_emit_data(uint8_t const buf[], size_t const length)
{
    if (nullptr == mEngine) {
        throw std::runtime_error("Null engine pointer!");
    }
   if(mEngine->mIsServer){
       sleep(5);  //delay of 5 sec at server side                             
   }
}
```
This is not the right behavior according to the RFC, because in case of partial read, we shouldn't move to Sending state.
The same delay when we add it for client side the handshake is getting completed-
```
if(!mEngine->mIsServer){
       sleep(5);  //delay of 5 sec at client side                             
   }
```

Is adding delay at server side affecting retransmission behaviour?

Thiesius:
Hello,

This PR fixes several DTLS 1.2 handshake retransmission edge cases around lost, duplicated, delayed/partial, and retransmitted handshake flights. Some of these issues were already tracked.

Fixes #2310 (DTLS last write update)
Fixes #2498 (The DTLS server 1.2 is not handling properly the retransmissions of flight 5 from the client)
Fixes #4022 (Facing retransmission issue while adding delay in the server side in the DTLS implementation)
Fixes #4036 (The DTLS server 1.2 is not handling properly the retransmissions in case of delay in flight at server side)

Related to #4782 (DTLS session resumption; frequent "Can't interleave application and handshake data" errors)

Some fixes were ported from our company repository, and I also stress-tested the implementation to discover and validate related edge cases.

I will not pretend that the fixes and especially the unit tests were not AI-assisted. However, I have thoroughly self-reviewed the changes and, more importantly, stress-tested them against lossy/reordered/delayed DTLS traffic.

## So what is the context of these fixes

I think this might come up as obvious, but let's recap anyway. DTLS runs over unreliable datagrams. A peer may retransmit an earlier handshake flight after the local endpoint has already progressed, or even after it has locally activated the session. In those cases, we must be able to reproduce the correct previous response flight instead of treating the retransmission as a fresh or invalid handshake event.

Without these fixes, handshakes can stall or fail in cases such as:

- client retransmits `ClientHello` because `HelloVerifyRequest` was lost
- client proceeds after receiving only a partial server flight
- client finishes but loses the server final flight
- server activates but still needs to retransmit its final CCS/Finished
- resumed handshakes lose the client final flight after client activation
- peer application data arrives together with the final resumed-handshake flight

## Actual Fix List

- Retain `HelloVerifyRequest` as retransmittable DTLS flight data while still excluding it from handshake transcript hashes.
- Reconstruct retransmitted flights starting from the previous epoch when needed, so CCS is emitted before `Finished`.
- Detect retransmitted already-consumed handshake records and respond with the previous flight.
- Retain DTLS handshake IO after activation for final-flight retransmission.
- Continue timeout polling briefly after local DTLS activation when the final flight may still need retransmission.
- Stop final-flight retransmission once peer application data proves the peer progressed.
- Update timeout pacing state after retransmission.

## Unit Test Coverage

I grouped the new coverage under DTLS core regression tests. The tests use Botan’s real DTLS client/server machinery, but simulate UDP behavior by moving emitted datagram bytes between peers manually. The test cpp has grown quite big, maybe worth being split from TLS tests in follow up PR.

Covered cases:

- `timeout_check()` retransmission pacing
- lost `HelloVerifyRequest`
- duplicate `HelloVerifyRequest`
- lost server handshake flight
- partial server flight delivery where the client must wait for `ServerHelloDone`
- lost server final flight
- retransmitted epoch-gap flight including CCS
- final-flight retransmission before application data
- resumed handshake accepting app data
- resumed client final-flight retransmission after activation
- resumed final flight and immediate application data arriving in one receive

## Local Stress Testing

The stress tools are not included in this PR. I used them only for local validation.

Originally I used `clumsy`, a Windows traffic-control tool, but those tests were heavily dependent on randomness. I then used a simple UDP fault proxy and standalone DTLS stress app to make the scenarios easier to repeat and vary.

Stress-tested scenarios included:

- clean DTLS handshake and echo
- 50-seed handshake drop sweep
- directional client-to-server drops
- directional server-to-client drops
- lost first client handshake datagram
- lost first server handshake datagram / `HelloVerifyRequest`
- lost server final flight
- out-of-order handshake flights
- duplicated handshake datagrams
- delayed handshake datagrams
- mixed delay/drop/duplicate/reorder handshake conditions
- app-data echo under clean proxy
- app-data echo under delay/reorder/duplicate when application datagrams are not intentionally dropped

Dropped DTLS application data was not treated as a bug. DTLS preserves datagram boundaries and protects records, but it does not make UDP application data reliable.

### DIFF
--- a/doc/api_ref/tls.rst
+++ b/doc/api_ref/tls.rst
@@ -264,9 +264,6 @@ available:
       retransmission is due. If ``std::nullopt`` is returned, no timeout check
       is currently needed.
 
-      This may return a duration briefly after ``is_active`` becomes true,
-      because a DTLS peer may not have received the final handshake flight yet.
-
    .. cpp:function:: bool is_closed()
 
       Returns true if and only if either a close notification or a
@@ -287,10 +284,8 @@ available:
    .. cpp:function:: bool timeout_check()
 
       This function does nothing unless the channel represents a DTLS
-      connection and a handshake still needs retransmission handling. This
-      includes handshakes in progress and locally active sessions whose final
-      handshake flight may not have reached the peer yet. Returns true if a
-      timeout condition occurred and handshake packets were retransmitted.
+      connection with a handshake in progress. Returns true if a timeout
+      condition occurred and handshake packets were retransmitted.
 
    .. cpp:function:: void renegotiate(bool force_full_renegotiation = false)
 
--- a/doc/api_ref/tls.rst
+++ b/doc/api_ref/tls.rst
@@ -257,12 +257,15 @@ available:
      this connection and the connection has not been subsequently
      closed.
 
-   .. cpp:function:: bool requires_timeout_check()
+   .. cpp:function:: std::optional<std::chrono::milliseconds> next_retransmission_timeout()
 
-      Returns true if ``timeout_check`` should be polled to drive DTLS
-      handshake retransmissions. This may remain true briefly after
-      ``is_active`` becomes true, because a DTLS peer may not have
-      received the final handshake flight yet.
+      Returns the remaining time until a call to ``timeout_check`` might
+      retransmit DTLS handshake data. A zero duration means that a
+      retransmission is due. If ``std::nullopt`` is returned, no timeout check
+      is currently needed.
+
+      This may return a duration briefly after ``is_active`` becomes true,
+      because a DTLS peer may not have received the final handshake flight yet.
 
    .. cpp:function:: bool is_closed()
 
--- a/doc/api_ref/tls.rst
+++ b/doc/api_ref/tls.rst
@@ -257,6 +257,13 @@ available:
      this connection and the connection has not been subsequently
      closed.
 
+   .. cpp:function:: bool requires_timeout_check()
+
+      Returns true if ``timeout_check`` should be polled to drive DTLS
+      handshake retransmissions. This may remain true briefly after
+      ``is_active`` becomes true, because a DTLS peer may not have
+      received the final handshake flight yet.
+
    .. cpp:function:: bool is_closed()
 
       Returns true if and only if either a close notification or a
@@ -277,10 +284,10 @@ available:
    .. cpp:function:: bool timeout_check()
 
       This function does nothing unless the channel represents a DTLS
-      connection and a handshake is actively in progress. In this case
-      it will check the current timeout state and potentially initiate
-      retransmission of handshake packets. Returns true if a timeout
-      condition occurred.
+      connection and a handshake still needs retransmission handling. This
+      includes handshakes in progress and locally active sessions whose final
+      handshake flight may not have reached the peer yet. Returns true if a
+      timeout condition occurred and handshake packets were retransmitted.
 
    .. cpp:function:: void renegotiate(bool force_full_renegotiation = false)
 
--- a/src/bogo_shim/config.json
+++ b/src/bogo_shim/config.json
@@ -265,7 +265,6 @@
         "KeyChangeWithBufferedMessages-DTLS": "No DTLS 1.3",
         "Renegotiate-DTLS-Server-Forbidden": "Botan tolerates DTLS renegotiation",
         "Renegotiate-DTLS-Client-Forbidden": "Botan tolerates DTLS renegotiation",
-        "DTLS12-SendExtraFinished-*": "Botan tolerates extra Finished messages in DTLS",
         "MixCompleteMessageWithFragments-DTLS-TLS12": "DTLS fragment handling difference",
         "RejectPSSKeyType-*": "Botan does not reject RSA-PSS key type",
         "CertificateCipherMismatch-PSS": "Botan does not reject PSS cipher mismatch",
--- a/src/lib/tls/tls12/tls_channel_impl_12.cpp
+++ b/src/lib/tls/tls12/tls_channel_impl_12.cpp
@@ -186,31 +186,9 @@ Handshake_State& Channel_Impl_12::create_handshake_state(Protocol_Version versio
    return *m_pending_state;
 }
 
-Handshake_IO* Channel_Impl_12::retransmission_io() {
-   return const_cast<Handshake_IO*>(std::as_const(*this).retransmission_io());
-}
-
-const Handshake_IO* Channel_Impl_12::retransmission_io() const {
-   if(!m_is_datagram || m_has_been_closed) {
-      return nullptr;
-   }
-
-   if(m_pending_state) {
-      return &m_pending_state->handshake_io();
-   }
-
-   // Until protected application data confirms that the peer processed our
-   // final flight, keep it eligible for timeout-driven retransmission.
-   if(m_active_state.has_value() && !m_active_state->peer_sent_protected_application_data()) {
-      return m_active_state->dtls_handshake_io();
-   }
-
-   return nullptr;
-}
-
 bool Channel_Impl_12::timeout_check() {
-   if(auto* io = retransmission_io()) {
-      return io->timeout_check();
+   if(m_is_datagram && !m_has_been_closed && m_pending_state) {
+      return m_pending_state->handshake_io().timeout_check();
    }
 
    //FIXME: scan cipher suites and remove epochs older than 2*MSL
@@ -298,8 +276,8 @@ bool Channel_Impl_12::is_active() const {
 }
 
 std::optional<std::chrono::milliseconds> Channel_Impl_12::next_retransmission_timeout() const {
-   if(const auto* io = retransmission_io()) {
-      return io->next_retransmission_timeout();
+   if(m_is_datagram && !m_has_been_closed && m_pending_state) {
+      return m_pending_state->handshake_io().next_retransmission_timeout();
    }
 
    return std::nullopt;
@@ -324,13 +302,18 @@ void Channel_Impl_12::activate_session() {
       map_remove_if(not_current_epoch, m_read_cipher_states);
    }
 
-   // For DTLS, keep the handshake IO for last-flight retransmission.
+   // In a full handshake the server sends the terminal flight; in an
+   // abbreviated handshake the client does. Both endpoints retain handshake
+   // sequence state, but only the terminal sender replays its outgoing flight.
+   const bool sent_terminal_dtls_flight = m_is_datagram && (m_is_server == (state.server_hello_done() != nullptr));
+
    if(m_is_datagram) {
       m_active_state = Active_Connection_State_12(state, application_protocol(), m_pending_state->take_handshake_io());
       if(auto* dtls_io = m_active_state->dtls_handshake_io()) {
-         // Explicitly mark the just-sent final flight complete so timeout_check()
-         // may retransmit it if the peer never receives it.
-         dtls_io->finalize_handshake();
+         // Retain receive sequence state on both endpoints to distinguish a
+         // retransmission from an unexpected new handshake message. Only the
+         // terminal-flight sender responds by replaying its final flight.
+         dtls_io->finalize_handshake(sent_terminal_dtls_flight);
       }
    } else {
       m_active_state = Active_Connection_State_12(state, application_protocol());
@@ -387,7 +370,7 @@ size_t Channel_Impl_12::from_peer(std::span<const uint8_t> data) {
             throw TLS_Exception(Alert::RecordOverflow, "TLS plaintext record is larger than allowed maximum");
          }
 
-         const bool epoch0_restart = m_is_datagram && record.epoch() == 0 && m_active_state.has_value();
+         const bool epoch0_restart = allow_epoch0_restart && record.epoch() == 0 && m_active_state.has_value();
          BOTAN_ASSERT_IMPLICATION(epoch0_restart, allow_epoch0_restart, "Allowed state");
 
          const bool initial_record = epoch0_restart || (pending_state() == nullptr && !m_active_state.has_value());
@@ -457,6 +440,13 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
                                             Record_Type record_type,
                                             Protocol_Version record_version,
                                             bool epoch0_restart) {
+   const auto process_retransmitted_record = [&] {
+      BOTAN_ASSERT(m_active_state.has_value(), "Have active DTLS association for retransmission");
+      BOTAN_ASSERT_NONNULL(m_active_state->dtls_handshake_io());
+      m_active_state->dtls_handshake_io()->add_retransmitted_record(
+         record.data(), record.size(), record_type, record_sequence);
+   };
+
    if(!m_pending_state) {
       // With no pending handshake this is either a new handshake attempt or a
       // DTLS retransmission from the previous handshake. The latter must not
@@ -465,9 +455,7 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
          const bool starts_new_handshake = is_new_dtls_association_client_hello(record, record_type);
 
          if(!starts_new_handshake) {
-            BOTAN_ASSERT_NONNULL(m_active_state->dtls_handshake_io());
-            m_active_state->dtls_handshake_io()->add_retransmitted_record(
-               record.data(), record.size(), record_type, record_sequence);
+            process_retransmitted_record();
             return;
          }
       }
@@ -488,17 +476,12 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
                     static_cast<Handshake_Type>(record[0]) == Handshake_Type::HelloRequest));
 
                if(m_active_state.has_value() && !starts_new_handshake) {
-                  BOTAN_ASSERT_NONNULL(m_active_state->dtls_handshake_io());
-                  m_active_state->dtls_handshake_io()->add_retransmitted_record(
-                     record.data(), record.size(), record_type, record_sequence);
+                  process_retransmitted_record();
                } else {
                   create_handshake_state(record_version);
                }
             } else if(current_epoch > 0 && epoch == current_epoch - 1) {
-               BOTAN_ASSERT(m_active_state.has_value(), "Have active DTLS association for retransmission");
-               BOTAN_ASSERT_NONNULL(m_active_state->dtls_handshake_io());
-               m_active_state->dtls_handshake_io()->add_retransmitted_record(
-                  record.data(), record.size(), record_type, record_sequence);
+               process_retransmitted_record();
             }
          } else {
             create_handshake_state(record_version);
@@ -539,8 +522,6 @@ void Channel_Impl_12::process_application_data(uint64_t seq_no, const secure_vec
    if(read_epoch == 0) {
       throw Unexpected_Message("Application data received in unexpected read epoch");
    }
-
-   m_active_state->mark_peer_as_having_sent_protected_application_data();
 
    callbacks().tls_record_received(seq_no, record);
 }
--- a/src/lib/tls/tls12/tls_channel_impl_12.cpp
+++ b/src/lib/tls/tls12/tls_channel_impl_12.cpp
@@ -366,7 +366,22 @@ size_t Channel_Impl_12::from_peer(std::span<const uint8_t> data) {
             return 0;
          }
 
+         const bool old_unprotected_record = m_is_datagram && record.epoch() == 0 && m_active_state.has_value() &&
+                                             sequence_numbers().current_read_epoch() > 0;
+
+         // Once encrypted traffic is expected, epoch-zero records are
+         // unauthenticated. Only handshake records can be useful as part of a
+         // retransmitted flight or an explicitly allowed association restart.
+         if(old_unprotected_record && record.type() != Record_Type::Handshake &&
+            record.type() != Record_Type::ChangeCipherSpec) {
+            continue;
+         }
+
          if(m_record_buf.size() > MAX_PLAINTEXT_SIZE) {
+            if(old_unprotected_record) {
+               continue;
+            }
+
             throw TLS_Exception(Alert::RecordOverflow, "TLS plaintext record is larger than allowed maximum");
          }
 
@@ -380,7 +395,7 @@ size_t Channel_Impl_12::from_peer(std::span<const uint8_t> data) {
             initial_handshake_message = (type == Handshake_Type::ClientHello);
          }
 
-         if(record.type() != Record_Type::Alert) {
+         if(record.type() != Record_Type::Alert && !old_unprotected_record) {
             if(initial_record) {
                // For initial records just check for basic sanity
                if(record.version().major_version() != 3 && record.version().major_version() != 0xFE) {
@@ -443,8 +458,16 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
    const auto process_retransmitted_record = [&] {
       BOTAN_ASSERT(m_active_state.has_value(), "Have active DTLS association for retransmission");
       BOTAN_ASSERT_NONNULL(m_active_state->dtls_handshake_io());
-      m_active_state->dtls_handshake_io()->add_retransmitted_record(
-         record.data(), record.size(), record_type, record_sequence);
+      try {
+         m_active_state->dtls_handshake_io()->add_retransmitted_record(
+            record.data(), record.size(), record_type, record_sequence);
+      } catch(...) {
+         // Epoch-zero records are unauthenticated and may be spoofed. Invalid
+         // retransmissions must not tear down an established association.
+         if((record_sequence >> 48) > 0) {
+            throw;
+         }
+      }
    };
 
    if(!m_pending_state) {
@@ -462,8 +485,6 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
 
       if(m_is_datagram && !epoch0_restart) {
          if(m_sequence_numbers) {
-            sequence_numbers().read_accept(record_sequence);
-
             const uint16_t epoch = record_sequence >> 48;
 
             const uint16_t current_epoch = sequence_numbers().current_read_epoch();
--- a/src/lib/tls/tls12/tls_channel_impl_12.cpp
+++ b/src/lib/tls/tls12/tls_channel_impl_12.cpp
@@ -21,20 +21,21 @@
 #include <botan/internal/tls_handshake_state.h>
 #include <botan/internal/tls_record.h>
 #include <botan/internal/tls_seq_numbers.h>
+#include <utility>
 
 namespace Botan::TLS {
 
 namespace {
 
-bool is_new_dtls_association_client_hello(std::span<const uint8_t> record, Record_Type record_type) {
+bool is_new_dtls_association_client_hello(std::span<const uint8_t> msg_and_header, Record_Type record_type) {
    constexpr size_t DTLS_HANDSHAKE_HEADER_SIZE = 12;
 
    // Every new DTLS handshake starts at message_seq 0. A cookie-bearing
    // ClientHello uses message_seq 1, so one arriving late belongs to the
    // previous handshake and must not replace the active association's state.
-   return record_type == Record_Type::Handshake && record.size() >= DTLS_HANDSHAKE_HEADER_SIZE &&
-          static_cast<Handshake_Type>(record[0]) == Handshake_Type::ClientHello &&
-          load_be<uint16_t>(&record[4], 0) == 0;
+   return record_type == Record_Type::Handshake && msg_and_header.size() >= DTLS_HANDSHAKE_HEADER_SIZE &&
+          static_cast<Handshake_Type>(msg_and_header[0]) == Handshake_Type::ClientHello &&
+          load_be(msg_and_header.subspan<4, 2>()) == 0;
 }
 
 }  // namespace
@@ -69,7 +70,6 @@ Channel_Impl_12::Channel_Impl_12(const std::shared_ptr<Callbacks>& callbacks,
 void Channel_Impl_12::reset_state() {
    m_active_state.reset();
    m_pending_state.reset();
-   m_dtls_peer_traffic_seen = true;
    m_readbuf.clear();
    m_write_cipher_states.clear();
    m_read_cipher_states.clear();
@@ -79,7 +79,6 @@ void Channel_Impl_12::reset_active_association_state() {
    // This operation only makes sense for DTLS
    BOTAN_ASSERT_NOMSG(m_is_datagram);
    m_active_state.reset();
-   m_dtls_peer_traffic_seen = true;
    m_read_cipher_states.clear();
    m_write_cipher_states.clear();
 
@@ -187,15 +186,31 @@ Handshake_State& Channel_Impl_12::create_handshake_state(Protocol_Version versio
    return *m_pending_state;
 }
 
-bool Channel_Impl_12::timeout_check() {
-   if(m_pending_state) {
-      return m_pending_state->handshake_io().timeout_check();
+Handshake_IO* Channel_Impl_12::retransmission_io() {
+   return const_cast<Handshake_IO*>(std::as_const(*this).retransmission_io());
+}
+
+const Handshake_IO* Channel_Impl_12::retransmission_io() const {
+   if(!m_is_datagram || m_has_been_closed) {
+      return nullptr;
    }
 
-   if(m_is_datagram && !m_has_been_closed && m_active_state.has_value() && !m_dtls_peer_traffic_seen) {
-      if(auto* dtls_io = m_active_state->dtls_handshake_io()) {
-         return dtls_io->timeout_check();
-      }
+   if(m_pending_state) {
+      return &m_pending_state->handshake_io();
+   }
+
+   // Until protected application data confirms that the peer processed our
+   // final flight, keep it eligible for timeout-driven retransmission.
+   if(m_active_state.has_value() && !m_active_state->peer_sent_protected_application_data()) {
+      return m_active_state->dtls_handshake_io();
+   }
+
+   return nullptr;
+}
+
+bool Channel_Impl_12::timeout_check() {
+   if(auto* io = retransmission_io()) {
+      return io->timeout_check();
    }
 
    //FIXME: scan cipher suites and remove epochs older than 2*MSL
@@ -282,16 +297,12 @@ bool Channel_Impl_12::is_active() const {
    return !is_closed() && is_handshake_complete();
 }
 
-bool Channel_Impl_12::requires_timeout_check() const {
-   if(m_pending_state) {
-      return true;
+std::optional<std::chrono::milliseconds> Channel_Impl_12::next_retransmission_timeout() const {
+   if(const auto* io = retransmission_io()) {
+      return io->next_retransmission_timeout();
    }
 
-   // A DTLS peer might not have received our final handshake flight yet, even
-   // though this side is already active. Keep timer ticks enabled until peer
-   // traffic confirms that it has moved on to the negotiated epoch.
-   return m_is_datagram && !m_has_been_closed && m_active_state.has_value() && !m_dtls_peer_traffic_seen &&
-          m_active_state->dtls_handshake_io() != nullptr;
+   return std::nullopt;
 }
 
 bool Channel_Impl_12::is_closed() const {
@@ -317,14 +328,12 @@ void Channel_Impl_12::activate_session() {
    if(m_is_datagram) {
       m_active_state = Active_Connection_State_12(state, application_protocol(), m_pending_state->take_handshake_io());
       if(auto* dtls_io = m_active_state->dtls_handshake_io()) {
-         // Mark the just-sent flight complete so timeout_check() may
-         // retransmit it if the peer never receives it.
-         dtls_io->conclude_flight();
+         // Explicitly mark the just-sent final flight complete so timeout_check()
+         // may retransmit it if the peer never receives it.
+         dtls_io->finalize_handshake();
       }
-      m_dtls_peer_traffic_seen = false;
    } else {
       m_active_state = Active_Connection_State_12(state, application_protocol());
-      m_dtls_peer_traffic_seen = true;
    }
 
    m_pending_state.reset();
@@ -452,18 +461,18 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
       // With no pending handshake this is either a new handshake attempt or a
       // DTLS retransmission from the previous handshake. The latter must not
       // create fresh pending state; it only asks us to replay our last flight.
-      if(record_version.is_datagram_protocol() && epoch0_restart && m_sequence_numbers && m_active_state.has_value()) {
+      if(epoch0_restart && m_sequence_numbers && m_active_state.has_value()) {
          const bool starts_new_handshake = is_new_dtls_association_client_hello(record, record_type);
 
          if(!starts_new_handshake) {
-            BOTAN_ASSERT(m_active_state->dtls_handshake_io(), "Have DTLS handshake IO for retransmission");
+            BOTAN_ASSERT_NONNULL(m_active_state->dtls_handshake_io());
             m_active_state->dtls_handshake_io()->add_retransmitted_record(
                record.data(), record.size(), record_type, record_sequence);
             return;
          }
       }
 
-      if(record_version.is_datagram_protocol() && !epoch0_restart) {
+      if(m_is_datagram && !epoch0_restart) {
          if(m_sequence_numbers) {
             sequence_numbers().read_accept(record_sequence);
 
@@ -479,19 +488,17 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
                     static_cast<Handshake_Type>(record[0]) == Handshake_Type::HelloRequest));
 
                if(m_active_state.has_value() && !starts_new_handshake) {
-                  BOTAN_ASSERT(m_active_state->dtls_handshake_io(), "Have DTLS handshake IO for retransmission");
+                  BOTAN_ASSERT_NONNULL(m_active_state->dtls_handshake_io());
                   m_active_state->dtls_handshake_io()->add_retransmitted_record(
                      record.data(), record.size(), record_type, record_sequence);
-                  return;
                } else {
                   create_handshake_state(record_version);
                }
             } else if(current_epoch > 0 && epoch == current_epoch - 1) {
-               BOTAN_ASSERT(m_active_state.has_value() && m_active_state->dtls_handshake_io(),
-                            "Have DTLS handshake IO for retransmission");
+               BOTAN_ASSERT(m_active_state.has_value(), "Have active DTLS association for retransmission");
+               BOTAN_ASSERT_NONNULL(m_active_state->dtls_handshake_io());
                m_active_state->dtls_handshake_io()->add_retransmitted_record(
                   record.data(), record.size(), record_type, record_sequence);
-               return;
             }
          } else {
             create_handshake_state(record_version);
@@ -533,9 +540,7 @@ void Channel_Impl_12::process_application_data(uint64_t seq_no, const secure_vec
       throw Unexpected_Message("Application data received in unexpected read epoch");
    }
 
-   if(m_is_datagram) {
-      m_dtls_peer_traffic_seen = true;
-   }
+   m_active_state->mark_peer_as_having_sent_protected_application_data();
 
    callbacks().tls_record_received(seq_no, record);
 }
--- a/src/lib/tls/tls12/tls_channel_impl_12.cpp
+++ b/src/lib/tls/tls12/tls_channel_impl_12.cpp
@@ -24,6 +24,21 @@
 
 namespace Botan::TLS {
 
+namespace {
+
+bool is_new_dtls_association_client_hello(std::span<const uint8_t> record, Record_Type record_type) {
+   constexpr size_t DTLS_HANDSHAKE_HEADER_SIZE = 12;
+
+   // Every new DTLS handshake starts at message_seq 0. A cookie-bearing
+   // ClientHello uses message_seq 1, so one arriving late belongs to the
+   // previous handshake and must not replace the active association's state.
+   return record_type == Record_Type::Handshake && record.size() >= DTLS_HANDSHAKE_HEADER_SIZE &&
+          static_cast<Handshake_Type>(record[0]) == Handshake_Type::ClientHello &&
+          load_be<uint16_t>(&record[4], 0) == 0;
+}
+
+}  // namespace
+
 Channel_Impl_12::Channel_Impl_12(const std::shared_ptr<Callbacks>& callbacks,
                                  const std::shared_ptr<Session_Manager>& session_manager,
                                  const std::shared_ptr<RandomNumberGenerator>& rng,
@@ -54,6 +69,7 @@ Channel_Impl_12::Channel_Impl_12(const std::shared_ptr<Callbacks>& callbacks,
 void Channel_Impl_12::reset_state() {
    m_active_state.reset();
    m_pending_state.reset();
+   m_dtls_peer_traffic_seen = true;
    m_readbuf.clear();
    m_write_cipher_states.clear();
    m_read_cipher_states.clear();
@@ -63,6 +79,7 @@ void Channel_Impl_12::reset_active_association_state() {
    // This operation only makes sense for DTLS
    BOTAN_ASSERT_NOMSG(m_is_datagram);
    m_active_state.reset();
+   m_dtls_peer_traffic_seen = true;
    m_read_cipher_states.clear();
    m_write_cipher_states.clear();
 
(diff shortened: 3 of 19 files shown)


randombit:
@Thiesius thanks for digging into this and resolving these longstanding issues. This PR is big so it may take a while to review completely but I'll try a first pass asap.

Thiesius:
> @Thiesius thanks for digging into this and resolving these longstanding issues. This PR is big so it may take a while to review completely but I'll try a first pass asap.

No rush at all. 

BoGo tests seems to have caught one more regression that I have introduced in this PR. I have added a fix + unit test around it so that it's closer to the source. I didn't test it in the WSL, but if the CI/CD runners are costly, then I can maybe do that next time before sending a commit.

I also feel like the `add_record` now carries significant cognitive load, and it could probably use a little break-down. I would maybe postpone that for a follow up PR, so that the review is not completely hellish.

Thiesius:
I have enabled the build in WSL so that I could investigate the failures around BoGo. I think I have them all covered, but I have been checking the RFCs and seems like there is one more thing we don't have covered. I'm going to look into that in following day(s).

@randombit Should I mark the PR as a draft until the pipeline is green?

randombit:
No particular need, locally draft is more for signaling "don't even bother looking at this yet it's still going to probably change dramatically before being all the way done" whereas this sounds like it's just (hopefully) a few smaller changes.

Thiesius:
> No particular need, locally draft is more for signaling "don't even bother looking at this yet it's still going to probably change dramatically before being all the way done" whereas this sounds like it's just (hopefully) a few smaller changes.

Alrighty.

**As for the follow-up**: We should now be able to distinguish retransmissions from new handshake messages, and wait for complete terminal-flight "evidence" including CCS/Finished + correctly recognize `HelloRequest` as server-initiated renegotiation. 

Slightly stepping out of my comfortable domain here, since my stress testing was mainly focused on "vanilla" client-server handshakes and their reliability on unstable networks (5G, Radio, WiFi) 😅 

Anyway I have also removed the obsolete `DTLS12-SendExtraFinished-*` exclusion from bogo `config.json`. The Botan-configured BoGo suite tests passed. 

I know that I originally mentioned that I would leave the `add_record()` refactor for later, but it was becoming painful to watch it grow, so at least the previous-flight handling now lives in a separate helper method.

The stress-test maybe discovered one more thingy that is not covered by Bogo or our unit tests. I will investigate it tomorrow.

Thiesius:
So, while stress-testing through the UDP fault proxy, one deterministic reorder seed still stalled after the server had activated. 

Packet tracing showed a delayed cookie-bearing `ClientHello` arriving with epoch 0 and message_seq = 1. We treated every epoch-0 `ClientHello` as a new association, creating pending handshake state and preventing the retained active handshake state from responding when the client retransmitted its final flight.

src/lib/tls/tls_channel_impl.h
```diff
@@ -85,6 +85,12 @@ class Channel_Impl {
       */
       virtual bool is_active() const = 0;
 
+      /**
+      * @return true iff timeout_check() should be polled to drive handshake
+      *         retransmissions.
+      */
+      virtual bool requires_timeout_check() const { return !is_handshake_complete(); }
```
reneme: This should probably just return `false`. For DTLS 1.2 the callback is overridden anyway and DTLS 1.3 isn't implemented, yet. So this base implementation isn't hit in reasonable application code anyway.

Alternatively, I suggest to additionally check if the channel represents a DTLS communication and only then return `true` when a handshake is in progress. TCP-based TLS doesn't need this timeout check at all and should thus always return `false`.

doc/api_ref/tls.rst
```diff
@@ -255,6 +255,13 @@ available:
      this connection and the connection has not been subsequently
      closed.
 
+   .. cpp:function:: bool requires_timeout_check()
+
+      Returns true if ``timeout_check`` should be polled to drive DTLS
+      handshake retransmissions. This may remain true briefly after
+      ``is_active`` becomes true, because a DTLS peer may not have
+      received the final handshake flight yet.
```
reneme: Since we're introducing new API: I think it would be nice to have an indication on the current (internal) retransmission timeout. That way, applications could avoid calling the `timeout_check()` unnecessarily.

Perhaps like so:

```C++
/**
 * Returns a hint when a call to `timeout_check()` would potentially
 * retransmit a previous message. If std::nullopt is returned, no
 * invocation to `timeout_check()` is necessary at all.
 */
std::optional<std::chrono::milliseconds> next_retransmission_timeout();
```

src/lib/tls/tls_channel.h
```diff
@@ -121,6 +121,12 @@ class BOTAN_PUBLIC_API(2, 0) Channel {
       */
       virtual bool is_active() const = 0;
 
+      /**
+      * @return true iff timeout_check() should be polled to drive handshake
+      *         retransmissions.
+      */
+      virtual bool requires_timeout_check() const { return !is_handshake_complete(); }
```
reneme: Virtual methods of this class must be abstract. Otherwise, you could get away with not overriding it in `tls_client.h` or `tls_server.h` and you would never delegate the call to a concrete underlying TLS implementation. Currently, this patch is doing the overrides correctly but for defensive programming this must be:


```suggestion
      virtual bool requires_timeout_check() const = 0;
```

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -24,6 +24,21 @@
...
+
+bool is_new_dtls_association_client_hello(std::span<const uint8_t> record, Record_Type record_type) {
+   constexpr size_t DTLS_HANDSHAKE_HEADER_SIZE = 12;
+
+   // Every new DTLS handshake starts at message_seq 0. A cookie-bearing
+   // ClientHello uses message_seq 1, so one arriving late belongs to the
+   // previous handshake and must not replace the active association's state.
+   return record_type == Record_Type::Handshake && record.size() >= DTLS_HANDSHAKE_HEADER_SIZE &&
+          static_cast<Handshake_Type>(record[0]) == Handshake_Type::ClientHello &&
+          load_be<uint16_t>(&record[4], 0) == 0;
```
reneme: Nit:


```suggestion
          load_be(record.subspan(4, 2)) == 0;
```

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -175,6 +192,12 @@ bool Channel_Impl_12::timeout_check() {
       return m_pending_state->handshake_io().timeout_check();
    }
 
+   if(m_is_datagram && !m_has_been_closed && m_active_state.has_value() && !m_dtls_peer_traffic_seen) {
+      if(auto* dtls_io = m_active_state->dtls_handshake_io()) {
+         return dtls_io->timeout_check();
+      }
+   }
```
reneme: Perhaps add a comment explaining the rationale here. Like so? (assuming I understood correctly)


```suggestion
   // If our last handshake flight wasn't yet acknowledged by application data
   // sent from the peer, we might need to retransmit out last flight.
   if(m_is_datagram && !m_has_been_closed && m_active_state.has_value() && !m_dtls_peer_traffic_seen) {
      if(auto* dtls_io = m_active_state->dtls_handshake_io()) {
         return dtls_io->timeout_check();
      }
   }
```

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -175,6 +192,12 @@ bool Channel_Impl_12::timeout_check() {
...
+
    //FIXME: scan cipher suites and remove epochs older than 2*MSL
    return false;
 }
```
reneme, line 196: I'm worried that this (fairly complex) logic is duplicated in `requires_timeout_check()`. We should just re-use it here by somewhat inverting the flow of the method. Like so:


```C++
bool Channel_Impl_12::timeout_check() {
   if(!requires_timeout_check()) {
      //FIXME: scan cipher suites and remove epochs older than 2*MSL
      return false;
   }

   if(m_pending_state) {
      return m_pending_state->handshake_io().timeout_check();
   }

   if(auto* dtls_io = m_active_state->dtls_handshake_io()) {
      return dtls_io->timeout_check();
   }
}
```

... if the generic check sees no reason for the timeout check we just leave it be. Otherwise, we delegate to the pending state. Or, if no more pending state is there, an opportunistic delegation to the active state should be tried.

src/lib/tls/tls12/tls_channel_impl_12.h
```diff
@@ -246,6 +246,7 @@ class Channel_Impl_12 : public Channel_Impl {
       secure_vector<uint8_t> m_record_buf;
 
       bool m_has_been_closed;
+      bool m_dtls_peer_traffic_seen = true;
```
reneme: I understand that this flag should indicate that we successfully received protected application data from the peer, right?

Instead of placing it so centrally in the TLS channel implementation, couldn't we move it into the "active state" object. From what I've seen its only ever relevant once we reached the "active" state, anyway. To my mind, that would be a more suitable place to hold it. Also, then the flag is reset more naturally when the active state is destroyed (e.g. for a new incoming association).

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -63,6 +79,7 @@ void Channel_Impl_12::reset_active_association_state() {
    // This operation only makes sense for DTLS
    BOTAN_ASSERT_NOMSG(m_is_datagram);
    m_active_state.reset();
+   m_dtls_peer_traffic_seen = true;
```
reneme: Disclaimer: I'm not deep enough into the DTLS implementation to make the following statement with full confidence. Hence, take it with a grain of salt.

That said: this is called whenever an active connection sees a new epoch=0 ClientHello. I.e. when a new handshake is meant to be performed to re-associate with a restarted client. Therefore, shouldn't we treat this new incoming connection with `m_dtls_peer_traffic_seen = false` to allow for final-flight retransmission in the handshake if necessary?

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -407,25 +449,49 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
...
    if(!m_pending_state) {
-      // No pending handshake, possibly new:
+      // With no pending handshake this is either a new handshake attempt or a
+      // DTLS retransmission from the previous handshake. The latter must not
+      // create fresh pending state; it only asks us to replay our last flight.
+      if(record_version.is_datagram_protocol() && epoch0_restart && m_sequence_numbers && m_active_state.has_value()) {
+         const bool starts_new_handshake = is_new_dtls_association_client_hello(record, record_type);
+
+         if(!starts_new_handshake) {
+            BOTAN_ASSERT(m_active_state->dtls_handshake_io(), "Have DTLS handshake IO for retransmission");
```
reneme:
```suggestion
            BOTAN_ASSERT_NONNULL
            (m_active_state->dtls_handshake_io());
```

... maybe?

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -407,25 +449,49 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
...
-               create_handshake_state(record_version);
+               // Either endpoint can initiate renegotiation from FINISHED:
+               // clients send ClientHello, servers send HelloRequest.
+               const bool starts_new_handshake =
+                  (record_type == Record_Type::Handshake && !record.empty() &&
+                   (static_cast<Handshake_Type>(record[0]) == Handshake_Type::ClientHello ||
+                    static_cast<Handshake_Type>(record[0]) == Handshake_Type::HelloRequest));
+
+               if(m_active_state.has_value() && !starts_new_handshake) {
+                  BOTAN_ASSERT(m_active_state->dtls_handshake_io(), "Have DTLS handshake IO for retransmission");
```
reneme: As above. Perhaps `BOTAN_ASSERT_NONNULL` might be better suited here:


```suggestion
                  BOTAN_ASSERT_NONNULL(m_active_state->dtls_handshake_io());
```

reneme (commented):
Some comments below. Generally this looks like a nice improvement. I still have to dig into the handshake I/O extensions.

src/lib/tls/tls12/tls_handshake_io.h
```diff
@@ -144,10 +145,42 @@ class Datagram_Handshake_IO final : public Handshake_IO {
...
+      // an active DTLS association for final-flight recovery.
+      void add_retransmitted_record(const uint8_t record[],
+                                    size_t record_len,
+                                    Record_Type type,
+                                    uint64_t sequence_number);
+
       std::pair<Handshake_Type, std::vector<uint8_t>> get_next_record(bool expecting_ccs,
                                                                       size_t max_message_size) override;
 
+      void conclude_flight();
```
reneme: I find the name of this misleading because the only flight that is actively "concluded" is the last flight of each peer's handshake messages. Other flight conclusions are implicitly signalled by calling `get_next_record()`. Simply calling this `conclude_flight()` makes me expect that this needs to be called whenever a flight was prepared fully.

Perhaps rather: `conclude_outgoing_handshake_traffic()`? Or simply `conclude_handshake()`?
Alternatively, we could change the implicit flight conclusion to an explicit invocation wherever required and adapt the state machine implementations respectively.

Also: a little documentation as to what the side effects of this are would be nice.

reneme (changes requested):
From what I can tell, the changes look sound. Also a Claude Sonnet review didn't bring up anything substantial at this point.

I'm setting to "Request Changes" regarding my mentioned points, but generally this looks solid to me. 👍

doc/api_ref/tls.rst
Thiesius: Agreed, a remaining time hint is more useful than a boolean and avoids unnecessary polling. Done.

src/lib/tls/tls12/tls_channel_impl_12.cpp
Thiesius: Done (with constexpr extent though)

src/lib/tls/tls12/tls_channel_impl_12.cpp
Thiesius: The new association should indeed begin as not progressed, but this reset happens before it becomes active. 

The pending handshake owns timeout handling until activation. Moving the flag into the active state makes that transition more readable (as per another comment of your below), I have also renamed it to `peer_has_progressed`.

src/lib/tls/tls12/tls_channel_impl_12.cpp
Thiesius: I have re-shaped the structure a bit around the timers (with the prior PR comment). But the comment is in place.

src/lib/tls/tls12/tls_channel_impl_12.cpp
Thiesius: Sure. I addressed this slightly differently alongside the timeout-hint API change. 

Both `timeout_check()` and `next_retransmission_timeout()` now use a shared `retransmission_io()` helper, which selects the pending handshake first and otherwise the eligible active DTLS handshake. Now the state-selection logic should be deduplicated.

Its const/non-const overload pair uses a localized `const_cast` to preserve pointer mutability without duplicating state selection (therefor no UB here). `Deducing this` would make this cleaner, but I belive Botan currently targets C++20 and not C++23.

src/lib/tls/tls12/tls_channel_impl_12.cpp
Thiesius: Done. No hard opinion on this.

src/lib/tls/tls12/tls_channel_impl_12.cpp
Thiesius: Done. No hard opinion on this.

src/lib/tls/tls12/tls_channel_impl_12.h
Thiesius: Yes, moved to the active state. See the prior comment.

src/lib/tls/tls12/tls_handshake_io.h
Thiesius: Yes, I agree that `conclude_flight()` is a bit vague. I have renamed it to `finalize_handshake()` and documented that it closes only the terminal outgoing flight so timeout replay can identify it

src/lib/tls/tls_channel.h
Thiesius: Done. Slipped around me.

src/lib/tls/tls_channel_impl.h
Thiesius: True. 

With the new timeout hint API, the public method is now abstract and the internal base implementation returns std::nullopt. DTLS 1.2 overrides it, while stream TLS and the current TLS 1.3 implementation therefore never request timeout polling.

Thiesius (commented):
Thanks for the review!

doc/api_ref/tls.rst
```diff
@@ -255,6 +255,16 @@ available:
...
 
+   .. cpp:function:: std::optional<std::chrono::milliseconds> next_retransmission_timeout()
+
+      Returns the remaining time until a call to ``timeout_check`` might
+      retransmit DTLS handshake data. A zero duration means that a
+      retransmission is due. If ``std::nullopt`` is returned, no timeout check
+      is currently needed.
+
+      This may return a duration briefly after ``is_active`` becomes true,
+      because a DTLS peer may not have received the final handshake flight yet.
```
reneme: @atreiber94 Could you please verify that this new API would work in our scenario.

src/lib/tls/tls12/tls_channel_impl_12.h
```diff
@@ -200,6 +200,10 @@ class Channel_Impl_12 : public Channel_Impl {
 
       void reset_state();
 
+      class Handshake_IO* retransmission_io();
+
+      const class Handshake_IO* retransmission_io() const;
```
reneme: Frankly, I've never seen this syntax before. I'm guessing this is essentially an inline forward-declaration. For consistency with the rest of the code base: could you please do an explicit forward declaration at the beginning of this file and remove the `class` keyword here? There's already a bunch of forward declares up there.

src/lib/tls/tls12/tls_connection_state_12.h
```diff
@@ -83,6 +83,12 @@ class Active_Connection_State_12 final {
 
       const Datagram_Handshake_IO* dtls_handshake_io() const { return m_dtls_handshake_io.get(); }
 
+      // Protected application data proves that the peer processed our final
+      // handshake flight, so timeout-driven replay is no longer necessary.
+      bool peer_has_progressed() const { return m_peer_has_progressed; }
+
+      void mark_peer_as_progressed() { m_peer_has_progressed = true; }
```
reneme: Naming nit: perhaps `peer_sent_protected_application_data()`, its long but specific.

src/lib/tls/tls_channel.h
```diff
@@ -122,10 +123,10 @@ class BOTAN_PUBLIC_API(2, 0) Channel {
       virtual bool is_active() const = 0;
 
       /**
-      * @return true iff timeout_check() should be polled to drive handshake
-      *         retransmissions.
+      * @return the remaining time until timeout_check() might retransmit
+      *         handshake data, or std::nullopt if no timeout check is needed.
       */
-      virtual bool requires_timeout_check() const { return !is_handshake_complete(); }
+      virtual std::optional<std::chrono::milliseconds> next_retransmission_timeout() const = 0;
```
reneme, line 129: @randombit This introduces `<chrono>` in the public API. I did just remember that you weren't a fan of heavy STL-headers in the public headers. I'm in the "readability over compile-time camp" for this, frankly, but definitely your call. (I asked for it in my review yesterday, so my bad.)

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -21,9 +21,24 @@
...
+
+bool is_new_dtls_association_client_hello(std::span<const uint8_t> record, Record_Type record_type) {
+   constexpr size_t DTLS_HANDSHAKE_HEADER_SIZE = 12;
+
+   // Every new DTLS handshake starts at message_seq 0. A cookie-bearing
+   // ClientHello uses message_seq 1, so one arriving late belongs to the
+   // previous handshake and must not replace the active association's state.
+   return record_type == Record_Type::Handshake && record.size() >= DTLS_HANDSHAKE_HEADER_SIZE &&
+          static_cast<Handshake_Type>(record[0]) == Handshake_Type::ClientHello && load_be(record.subspan<4, 2>()) == 0;
+}
```
reneme: Another naming nit: `record` isn't correct here. The buffer contains the marshalled handshake message without the record header. I find that an important detail, because the function relies on the specified marshalling of RFC 6347 (for the message type, sequence number and record length).


```suggestion
bool is_new_dtls_association_client_hello(std::span<const uint8_t> msg_and_header, Record_Type record_type) {
   constexpr size_t DTLS_HANDSHAKE_HEADER_SIZE = 12;

   // Every new DTLS handshake starts at message_seq 0. A cookie-bearing
   // ClientHello uses message_seq 1, so one arriving late belongs to the
   // previous handshake and must not replace the active association's state.
   return record_type == Record_Type::Handshake && msg_and_header.size() >= DTLS_HANDSHAKE_HEADER_SIZE &&
          static_cast<Handshake_Type>(msg_and_header[0]) == Handshake_Type::ClientHello && load_be(msg_and_header.subspan<4, 2>()) == 0;
}
```

... code formatting may be different than this ad-hoc patch.

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -281,6 +326,11 @@ void Channel_Impl_12::activate_session() {
    // For DTLS, keep the handshake IO for last-flight retransmission.
    if(m_is_datagram) {
       m_active_state = Active_Connection_State_12(state, application_protocol(), m_pending_state->take_handshake_io());
+      if(auto* dtls_io = m_active_state->dtls_handshake_io()) {
+         // Mark the just-sent flight complete so timeout_check() may
+         // retransmit it if the peer never receives it.
+         dtls_io->finalize_handshake();
```
reneme:
```suggestion
         // Explicitly mark the just-sent final flight complete so timeout_check()
         // may retransmit it if the peer never receives it.
         dtls_io->finalize_handshake();
```

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -407,25 +457,49 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
                                             Protocol_Version record_version,
                                             bool epoch0_restart) {
    if(!m_pending_state) {
-      // No pending handshake, possibly new:
+      // With no pending handshake this is either a new handshake attempt or a
+      // DTLS retransmission from the previous handshake. The latter must not
+      // create fresh pending state; it only asks us to replay our last flight.
+      if(record_version.is_datagram_protocol() && epoch0_restart && m_sequence_numbers && m_active_state.has_value()) {
```
reneme: This is merely a hunch (also because the existing code does the same), but: shouldn't we rather rely on the `m_is_datagram` flag to decide whether or not we're doing DTLS? Using the incoming record version (which may be attacker-controlled in the worst case) seems like an unnecessary risk.

In this particular conditional, one could argue that the `is_datagram_protocol()` is redundant anyway, because `epoch0_restart` will only ever be true for a DTLS channel (see `::from_peer()`).

@randombit perhaps a follow-up?

doc/api_ref/tls.rst
atreiber94: Yeah, this API would work nicely.

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -407,25 +457,49 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
...
+               const bool starts_new_handshake =
+                  (record_type == Record_Type::Handshake && !record.empty() &&
+                   (static_cast<Handshake_Type>(record[0]) == Handshake_Type::ClientHello ||
+                    static_cast<Handshake_Type>(record[0]) == Handshake_Type::HelloRequest));
+
+               if(m_active_state.has_value() && !starts_new_handshake) {
+                  BOTAN_ASSERT_NONNULL(m_active_state->dtls_handshake_io());
+                  m_active_state->dtls_handshake_io()->add_retransmitted_record(
+                     record.data(), record.size(), record_type, record_sequence);
+                  return;
```
reneme: In my opinion the early-returns in this branch of the function are adding to the mental load of this -- already fairly complicated -- function. And in the full picture they're also quite useless. If this branch doesn't call `create_handshake_state()` the `if(m_pending_state)` in line 513 won't be taken ever and the function is just exitted.

Unfortunately, that doesn't necessarily hold true for the first early-return in the hunk above within the branch that opens in line 463. I'm having a hard time reasoning about all the potential states that could produce side-effects in (this) following branch if the early-return in line 470 is removed. So I'd probably keep that, but the ones in lines 493 and 502 could savely be removed. 

Given a second thought: factoring out the entire `if(!m_pending_state)` conditional into its own function would probably simplify the control flow significantly. @Thiesius that really doesn't have to happen in this PR, though.

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -467,6 +541,10 @@ void Channel_Impl_12::process_application_data(uint64_t seq_no, const secure_vec
       throw Unexpected_Message("Application data received in unexpected read epoch");
    }
 
+   if(m_is_datagram) {
+      m_active_state->mark_peer_as_progressed();
+   }
```
reneme: Nit: dropping the `if(m_is_datagram)` won't do any harm. For ordinary TLS it would just be an essential no-op. No hard opinion either way, though.

reneme (commented):
Apart from the commented-on nits below:

I think there's one situation that we don't handle gracefully, yet: An `epoch0_restart` (ClientHello with epoch=0 and `m_active_state.has_value()`) while currently performing another handshake already. In other words: if the client sends a new epoch0 ClientHello before a previous handshake has been finished successfully.

Such a new ClientHello message will be passed into `process_handshake_ccs()` where it'll skip the first conditional because `m_pending_state` already exists. Therefore it'll just be forwarded into the lower handshake logic. So far I didn't check in detail what'll happen there, but very likely it won't be "the right thing".

To the best of my understanding RFC 6347 also doesn't really specify what should happen in such a case. Additionally, I think, a potential avenue for DoS has to be considered for this situation: i.e. a malicious client forcing the server to perpetually perform expensive handshake crypto by simply sending new ClientHellos all the time.

Perhaps that's something we should just gracefully catch for now and (probably in `process_handshake_ccs()`) address it in a folllow up.

Thiesius:
I'm not very familiar with GitHub's review UI, and the inline threads became hidden after the latest commit, so I'll answer them together here to not bomb the mails with separate comments.

**What happens when another epoch-zero ClientHello arrives while a restart is already pending?**

Once the first `message_seq=0` ClientHello is consumed, `m_in_message_seq` advances to 1. Another sequence-zero ClientHello is therefore handled as previous-flight traffic and replays the pending response instead of invoking ClientHello processing again.

I added a regression test which verifies that HelloVerifyRequest is replayed, cookie processing is not repeated, the existing association remains active, and the pending replacement association can still complete.

**Can `Handshake_IO` use a normal forward declaration?**

Yes fully agree. Just AI pulling my leg. Not sure how I missed it. Good catch.

**Can the peer-progress state have a more specific name?**

Yes. It is now named around the exact thing being tracked: protected application data received from the peer.

**Should `<chrono>` remain in the public API?**

I left this unchanged pending @randombit's preference. But by a quick swoop through the repo we can see, that Botan already exposes chrono types in public TLS headers such as `tls_callbacks.h` and `tls_policy.h`. I would personally lean towards "safer" `std::chrono::milliseconds`, but I don't see the whole picture here (yet).

**Is `record` the wrong name for the marshalled handshake message?**

Done. It is now named `msg_and_header`, since it contains the DTLS handshake header and message rather than a complete record.

**Should the `finalize_handshake()` comment explicitly mention the final flight?**

Yes. The comment now explains that this explicitly completes the final outgoing flight so it remains available for timeout retransmission.

**Should channel mode come from `m_is_datagram` rather than the incoming record version?**

100% . The code should be the authoritative source. The redundant protocol check was also removed where `epoch0_restart` already implies DTLS.

**Can the redundant early returns be removed?**

Yes. The two unnecessary returns were removed. Control now falls through to the existing pending-state check. I left the larger function extraction for a possible follow-up to avoid complicating this PR further.

**Can the `m_is_datagram` guard around recording protected application data be removed?**

Yes. Only the DTLS retransmission logic currently consumes it.

reneme (approved):
Looks good to me. Tests are comprehensive albeit a little chatty. I don't think that should stand in the way of integrating this patch, though.

@randombit I think you should do another pass, perhaps focussed on the handshake I/O changes.

src/lib/tls/tls12/tls_handshake_io.cpp
```diff
@@ -428,9 +591,16 @@ std::vector<uint8_t> Datagram_Handshake_IO::send_under_epoch(const Handshake_Mes
       m_send_hs(epoch, Record_Type::ChangeCipherSpec, msg_bits);
       return std::vector<uint8_t>();  // not included in handshake hashes
    } else if(msg_type == Handshake_Type::HelloVerifyRequest) {
-      // This message is not included in the handshake hashes
-      send_message(m_out_message_seq, epoch, msg_type, msg_bits);
+      // This message is not included in the handshake hashes, but it is a
+      // DTLS flight and must be available for retransmission if it is lost.
+      m_flights.rbegin()->push_back(m_out_message_seq);
```
randombit: I think this is not correct. HelloVerifyRequest doesn't need to be replayed directly but instead recreated as a byproduct of the peer sending us a new ClientHello. RFC 6347 section 3.2.1 also explicitly excludes it "Note that timeout and retransmission do not apply to the HelloVerifyRequest, ..."

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -170,9 +186,31 @@ Handshake_State& Channel_Impl_12::create_handshake_state(Protocol_Version versio
...
+
    if(m_pending_state) {
-      return m_pending_state->handshake_io().timeout_check();
+      return &m_pending_state->handshake_io();
+   }
+
+   // Until protected application data confirms that the peer processed our
+   // final flight, keep it eligible for timeout-driven retransmission.
+   if(m_active_state.has_value() && !m_active_state->peer_sent_protected_application_data()) {
+      return m_active_state->dtls_handshake_io();
```
randombit: This logic seems overly loose. A post-handshake retransmit is only required if we are the protocol side that would send the last flight. Also it seems it would, for a protocol that is completely unidirectional, end up sending periodic handshake retransmissions for the lifetime of the connection.

src/lib/tls/tls12/tls_handshake_io.cpp
```diff
@@ -160,13 +162,16 @@ void Datagram_Handshake_IO::retransmit_flight(size_t flight_idx) {
 
    BOTAN_ASSERT(!flight.empty(), "Nonempty flight to retransmit");
 
-   uint16_t epoch = m_flight_data[flight[0]].epoch;
+   const uint16_t first_epoch = m_flight_data[flight[0]].epoch;
+   uint16_t epoch = (first_epoch == 0) ? 0 : first_epoch - 1;
```
randombit: IIUC this logic change will cause extra CCSs to be inserted in inappropriate circumstances. Consider renegotiation. The renegotiation ClientHello is sent under epoch 1, since it's already encrypted. This logic will then see that we're sending a message in epoch 1, decide an epoch gap has ensued, and insert a epoch 0 CCS. Depending on the peer this may cause the connection to fail immediately.

randombit (changes requested):
In general looks good and makes sense. Three specific issues where I think the behavior is distinctly wrong, though admittedly I haven't tried writing a testcase to confirm the behavior.

Thiesius:
Thanks for the review, @randombit. I fixed the comments in two follow-up commits and then did one more pass over the resulting `FINISHED` handling.

**HelloVerifyRequest**
Yep, I treated `HelloVerifyRequest` too much like a normal flight. It is no longer cached and does not arm a timer. If it gets lost, the client retransmits its initial `ClientHello` and the server recreates the HVR with the same `message_seq`.

The regression checks both the missing timer and the recovery path. Good catch.

**Post-handshake retransmission**
Also fixed. An activated endpoint has no proactive retransmission timer. The waiting peer retransmits its preceding flight, which makes the terminal sender replay its final flight: server for a full handshake, client for a resumed one.

Both sides keep the handshake IO for `message_seq` bookkeeping, but only the terminal sender can replay. A silent connection therefore stays silent.

**CCS inference and renegotiation**
Correct. A non-zero epoch tells us nothing about whether a CCS preceded the message and the renegotiation proves that nicely.

CCS is now recorded when actually emitted, including its position and epoch. Retransmission replays only what was really there. The renegotiation tests cover both initiating sides, while the final-flight tests still verify real CCS replay.

**DTLS record replay windows**
One extra bug fell out of the recovery tests. The anti-replay code used the packed epoch and record sequence as one global sliding window. After accepting an epoch-1 record, valid epoch-0 retransmissions looked ancient and were dropped -potentially including the CCS needed for final-flight recovery.

There is now one replay window per epoch, using only that epoch’s 48-bit record sequence. Write-side sequencing is unchanged.

I also "tightened": epoch-zero restart detection so only the configured server path treats it as a restart. An active client can process old epoch-zero retransmissions normally.

----------------------------------

I kept the changes in two commits to make them easier to review. The first addresses the three comments. The second is the additional FINISHED and replay-window pass.

I ran the usual tests + stress-tests again.

randombit:
>  The anti-replay code used the packed epoch and record sequence as one global sliding window. After accepting an epoch-1 record, valid epoch-0 retransmissions looked ancient and were dropped -potentially including the CCS needed for final-flight recovery.

> There is now one replay window per epoch, using only that epoch’s 48-bit record sequence. Write-side sequencing is unchanged.

Nice catch there!

src/lib/tls/tls12/tls_connection_state_12.h
```diff
@@ -75,12 +76,20 @@ class Active_Connection_State_12 final {
...
        */
       Datagram_Handshake_IO* dtls_handshake_io() { return m_dtls_handshake_io.get(); }
 
+      const Datagram_Handshake_IO* dtls_handshake_io() const { return m_dtls_handshake_io.get(); }
+
+      // Protected application data proves that the peer processed our final
+      // handshake flight, so timeout-driven replay is no longer necessary.
+      bool peer_sent_protected_application_data() const { return m_peer_sent_protected_application_data; }
+
+      void mark_peer_as_having_sent_protected_application_data() { m_peer_sent_protected_application_data = true; }
```
randombit: IIUC, now that final-flight replay is reactive (to a replay from the peer) this flag is no longer needed anymore

src/lib/tls/tls12/tls_handshake_io.cpp
```diff
@@ -151,60 +151,216 @@ Protocol_Version Datagram_Handshake_IO::initial_record_version() const {
...
+bool Datagram_Handshake_IO::timeout_check() {
+   const auto timeout = next_retransmission_timeout();
+   if(!timeout || timeout->count() > 0) {
       return false;
    }
 
    retransmit_last_flight();
 
+   // Retransmission restarts the backoff window just like a normal write.
+   m_last_write = steady_clock_ms();
```
randombit: The other calls to `retransmit_last_flight` in this file (which are reactive based on peer traffic) do not set `m_last_write` so IIUC we can end up in a scenario where we replay a final flight based on a peer retransmit, and then have a timer expire and replay again. Not the end of the world of course but seems easily avoidable by updating the clock [possibly the clock write should be moved into `retransmit_last_flight`?]

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -170,9 +186,31 @@ Handshake_State& Channel_Impl_12::create_handshake_state(Protocol_Version versio
...
+   }
+
    if(m_pending_state) {
-      return m_pending_state->handshake_io().timeout_check();
+      return &m_pending_state->handshake_io();
+   }
+
+   // Once protected application data confirms peer progress, the completed
+   // handshake IO is no longer relevant to timeout scheduling.
+   if(m_active_state.has_value() && !m_active_state->peer_sent_protected_application_data()) {
```
randombit: This branch, as I understand it, is basically dead. In `Datagram_Handshake_IO` both `next_retransmission_timeout` and `timeout_check` return nullopt if m_finished is true. But if we have an active state, then `m_finished` must be true (right?). So the functions which actually use this return value in `Channel_Impl_12` will just return nullopt anyway. Might as well skip it and just return nullptr if we do not have a pending state.

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -170,9 +186,31 @@ Handshake_State& Channel_Impl_12::create_handshake_state(Protocol_Version versio
    return *m_pending_state;
 }
 
-bool Channel_Impl_12::timeout_check() {
+Handshake_IO* Channel_Impl_12::retransmission_io() {
+   return const_cast<Handshake_IO*>(std::as_const(*this).retransmission_io());
```
randombit: Not a big fan of `const_cast`. And especially with the simplification of not having to look at the active state anymore (see other comment on the main `retransmission_io` body) , it seems more straightforward to just duplicate the logic of `retransmission_io` into `timeout_check` and `next_retransmission_timeout` directly.

src/bogo_shim/config.json
```diff
@@ -265,7 +265,6 @@
         "KeyChangeWithBufferedMessages-DTLS": "No DTLS 1.3",
         "Renegotiate-DTLS-Server-Forbidden": "Botan tolerates DTLS renegotiation",
         "Renegotiate-DTLS-Client-Forbidden": "Botan tolerates DTLS renegotiation",
-        "DTLS12-SendExtraFinished-*": "Botan tolerates extra Finished messages in DTLS",
```
randombit, line 268: This should also be fixed in `config_no_tls13.json`

randombit (changes requested):
Another pass, a few more comments - mostly fallout from earlier review comments

Thiesius:
**Application-data progress flag**

Yep. Removed the flag and the application-data update. Final-flight recovery is reactive now, so the flag no longer serves a purpose.

**Reactive replay not updating `m_last_write`**

Good catch. I moved the clock update into `retransmit_last_flight()`, so both timer-driven and reactive retransmissions restart the clock.

**Dead active-state timeout branch**

Yep, that branch was dead once the handshake IO had entered `FINISHED`. Removed it. Timeout handling now only considers a pending handshake.

**`const_cast` helper**

Agreed. With the active-state case gone, the helper was more code than the duplicated check.

**`config_no_tls13.json`**

Fixed there as well.
- `DTLS12-SendExtraFinished-Reordered`
- `DTLS12-SendExtraFinished-Packed`

-----------------------------

I also added a separate "cleanup" commit (`91bc2fab6`): removed stale timeout documentation and dead test code, tightened const correctness and map access, and simplified a few container operations. No behavioral changes intended.

src/lib/tls/tls_channel.h
randombit: I'll live

randombit:
@Thiesius thanks looking very good at this point. Does need a rebase to address a merge conflict, also could use some history cleanup [worst case, squash everything to a single commit; in any case commits just addressing old reviews don't need to be in the history]

src/lib/tls/tls12/tls_seq_numbers.h
```diff
@@ -120,56 +123,71 @@ class Datagram_Sequence_Numbers final : public Connection_Sequence_Numbers {
       uint64_t next_read_sequence() override { throw Invalid_State("DTLS uses explicit sequence numbers"); }
 
       bool already_seen(uint64_t sequence) const override {
-         const size_t window_size = sizeof(m_window_bits) * 8;
+         const uint16_t epoch = static_cast<uint16_t>(sequence >> 48);
```
randombit, line 126: This change has an unintended side effect that if, after we've moved to epoch 1, we'll still accept data from epoch 0. So it becomes possible to send a offpath fatal alert packet with epoch 0 and a sufficiently high sequence number, and tear down an existing connection.

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -406,26 +440,48 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
                                             Record_Type record_type,
                                             Protocol_Version record_version,
                                             bool epoch0_restart) {
+   const auto process_retransmitted_record = [&] {
+      BOTAN_ASSERT(m_active_state.has_value(), "Have active DTLS association for retransmission");
+      BOTAN_ASSERT_NONNULL(m_active_state->dtls_handshake_io());
+      m_active_state->dtls_handshake_io()->add_retransmitted_record(
+         record.data(), record.size(), record_type, record_sequence);
```
randombit: I think we can at least (partially?) address the issue of fake epoch 0 alerts [or invalid messages, etc] causing a connection teardown by here by running `add_retransmitted_record` in a try/catch and only rethrowing if epoch > 0

randombit (changes requested):
One more issue that needs to be addressed, should be a fairly simple fix

Thiesius:
Synced and rebased.

1) **Replay window.** 
Yeah, good catch. With the latest change we now silently discard epoch 0 alerts, application data, invalid record types and oversized records once encryption is active. Only handshake and CCS records reach the retained handshake state.

I also stopped advancing the epoch 0 replay window after moving to a later epoch. Otherwise a spoofed high sequence number could poison the window and suppress the real final-flight retransmission.

Added a regression that sends a spoofed fatal alert with a high sequence number and then verifies that the genuine final-flight retransmission still completes.

2) **Epoch 0 alerts**

Done. Exceptions from `add_retransmitted_record()` are now ignored for epoch 0. Exceptions from encrypted epochs are still rethrown. The regression also sends a malformed epoch 0 handshake record and verifies that it neither aborts the active association nor prevents the genuine retransmission afterward.

----------------------

After we are done, I will try to unleash the AI to group up the commits logically. Hopefully I won't get the "You are working on cybersecurity related tasks, verify with your ID" again 😸

randombit:
Last fix commit looks good to me thanks, ready for a history rewrite imo

Thiesius:
> Last fix commit looks good to me thanks, ready for a history rewrite imo

I have "rewritten" the branch (13 commits) into four logical commits covering the handshake-flight fixes, retransmission timing/restarts, final-flight handling, and epoch-zero filtering. The final tree looks identical.

randombit (approved):

PR #5697, opened 2026-06-25
During CRL parsing, reject if a CRL entry reason is RemoveFromCrl

randombit:
This reason code only makes sense in the context of delta CRLs, which we do not support. If seen, reject the CRL entirely rather than trying to interpret it.

### DIFF
--- a/src/lib/x509/crl_ent.cpp
+++ b/src/lib/x509/crl_ent.cpp
@@ -100,7 +100,19 @@ void CRL_Entry::decode_from(BER_Decoder& source) {
       // Extensions  ::=  SEQUENCE SIZE (1..MAX) OF Extension
 
       if(const auto* ext = data->m_extensions.get_extension_object_as<Cert_Extension::CRL_ReasonCode>()) {
-         data->m_reason = ext->get_reason();
+         const auto reason = ext->get_reason();
+
+         /*
+         * This reason code only makes sense in the context of delta CRL processing, which
+         * we do not currently support. Seeing it in a regular CRL suggests that something
+         * has gone very wrong (eg we are somehow processing a delta CRL, though that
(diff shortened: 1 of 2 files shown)


arckoor:
"Don't support" as in don't support parsing? Or don't support in general?
Because through the FFI (and by extension, the CPP interface), they can be [created](https://github.com/randombit/botan/blob/626928ce6c14a3ef45e02161dc006705369397e7/src/lib/ffi/ffi.h#L2741) I think?
I haven't actually tested creating a CRL entry with this reason, if it works my point stands, if it doesn't that should probably be documented somewhere (or that enum value removed for Botan4, if Botan rejects CRLs with it anyway there should be no way to actually load one that has it)

randombit:
> "Don't support" as in don't support parsing? Or don't support in general?

This is completely unsupported, a delta CRL must be created and interpreted with respect to a base CRL, and we don't have that interpretation step at all. Additionally a delta CRL must be marked with a critical Delta CRL Indicator extension (RFC 5280 5.2.4), which is not something we currently support at all, either to parse or generate. [The extension is mandated to be critical precisely to prevent an implementation that doesn't support delta CRLs from treating a delta CRL as a base CRL and misinterpreting it.] So a "real" delta CRL would be rejected anyway, the only way we would ever see this reason code is that the CRL is malformed.

> If it doesn't that should probably be documented somewhere (or that enum value removed for Botan4, if Botan rejects CRLs with it anyway there should be no way to actually load one that has it)

Probably so, at least unless/until real support for delta CRLs is added. At best using this you can create a CRL which is  invalid with respect to RFC 5280 and whose interpretation will depend on the implementation (eg OpenSSL treats remove-from-CRL in a base CRL as not revoked, while BoringSSL ignores the reason code and considers it revoked). This PR changes Botan's interpretation from "it's revoked" to "syntactically invalid".

PR #5711, opened 2026-07-03
Add SPAKE2+ password authenticated key exchange (RFC 9383)

randombit:


src/python/botan3.py
```diff
@@ -3486,6 +3406,184 @@ def srp6_client_agree(username: str, password: str, group: str, hsh: str, salt:
...
+            _DLL.botan_spake2p_params_init(byref(self.__obj), _ctype_str(ciphersuite))
+
+    @classmethod
+    def custom(cls, group: ECGroup, seed: bytes, hash_fn: str) -> Spake2pParams:
+        """
+        Create custom system parameters for an arbitrary group, deriving the
+        M/N group elements from the seed using hash to curve (which not all
+        groups support). Both peers must use the same group, seed, and hash.
+        """
+        params = cls()
```
arckoor: Nit: I think this is the only instance of it being done like so, the usual pattern would be `params = Spake2pParams()`

src/python/botan3.py
```diff
@@ -3486,6 +3406,184 @@ def srp6_client_agree(username: str, password: str, group: str, hsh: str, salt:
...
+        Return the size in bytes of a key confirmation message (confirmP or confirmV)
+        """
+        return _call_fn_returning_sz(lambda sz: _DLL.botan_spake2p_params_confirmation_size(self.__obj, sz))
+
+def spake2p_derive_secret(params: Spake2pParams, password: str, prover_id: bytes = b'', verifier_id: bytes = b'', salt: bytes = b'') -> bytes:
+    """
+    Derive a SPAKE2+ (RFC 9383) prover secret from a password, using Argon2id.
+
+    The returned secret is password equivalent, and is used with
+    spake2p_registration_record and Spake2pProver.
```
arckoor: Maybe enclose it in backticks `spake2p_registration_record`? More easily identified as a function when mousing over in IDEs

src/python/botan3.py
```diff
@@ -3486,6 +3406,184 @@ def srp6_client_agree(username: str, password: str, group: str, hsh: str, salt:
...
+        Consume the prover's key share and return the verifier's response
+        (its own key share followed by a key confirmation), which is sent
+        to the prover. Can be called only once.
+        """
+        return _call_fn_viewing_vec(lambda vc, vf:
+                                    _DLL.botan_spake2p_verifier_process_message(self.__obj, rng.handle_(),
+                                                                                peer_message, len(peer_message),
+                                                                                vc, vf))
+
+    def verify_confirmation(self, confirmation: bytes) -> None:
```
arckoor, line 3894: The `None` should be inferable, but I do see why it's there

src/python/botan3.py
```diff
@@ -3486,6 +3406,184 @@ def srp6_client_agree(username: str, password: str, group: str, hsh: str, salt:
...
+    def custom(cls, group: ECGroup, seed: bytes, hash_fn: str) -> Spake2pParams:
+        """
+        Create custom system parameters for an arbitrary group, deriving the
+        M/N group elements from the seed using hash to curve (which not all
+        groups support). Both peers must use the same group, seed, and hash.
+        """
+        params = cls()
+        _DLL.botan_spake2p_params_init_custom(byref(params.handle_()),
+                                              group.handle_(),
+                                              seed, len(seed),
```
arckoor, line 3753: ruff will almost certainly reformat stuff like this into one item per line

doc/api_ref/python.rst
```diff
@@ -723,6 +723,99 @@ HOTP
...
+An implementation of the SPAKE2+ password authenticated key exchange
+(RFC 9383). The prover knows the password itself, while the verifier
+stores only a registration record derived from the password.
+
+.. py:class:: Spake2pParams(ciphersuite)
+
+   The system parameters, selecting the elliptic curve group, the SPAKE2+
+   M/N group elements, and the hash function. The ciphersuite is one of
+   "P256-SHA256", "P256-SHA512", "P384-SHA256", "P384-SHA512", or
+   "P521-SHA512".
```
arckoor: There are slight differences to the `botan3.py` doc here that will disappear when / if #5233 is merged, just fyi

src/lib/ffi/ffi_spake2p.cpp
```diff
@@ -0,0 +1,362 @@
...
+BOTAN_FFI_DECLARE_DUMMY_STRUCT(botan_spake2p_params_struct, 0x2E1B4A96);
+BOTAN_FFI_DECLARE_DUMMY_STRUCT(botan_spake2p_prover_struct, 0x9F337C29);
+BOTAN_FFI_DECLARE_DUMMY_STRUCT(botan_spake2p_verifier_struct, 0xD70A9E13);
+#endif
+
+int botan_spake2p_params_init(botan_spake2p_params_t* params, const char* ciphersuite) {
+#if defined(BOTAN_HAS_PAKE_SPAKE2PLUS)
+   if(any_null_pointers(params, ciphersuite)) {
+      return BOTAN_FFI_ERROR_NULL_POINTER;
+   }
```
arckoor, line 71: This could be checked before the #ifdef check, that way behaviour for nullptrs doesn't change based on feature availability

randombit: Fixed

src/python/botan3.py
randombit: Seems better to be explicit to me (tbh I had no idea that Python could infer this)

src/python/botan3.py
randombit: Done

src/python/botan3.py
randombit: Fixed

src/lib/ffi/ffi_spake2p.cpp
```diff
@@ -0,0 +1,362 @@
...
+int botan_spake2p_params_destroy(botan_spake2p_params_t params) {
+   return BOTAN_FFI_CHECKED_DELETE(params);
+}
+
+int botan_spake2p_params_share_size(botan_spake2p_params_t params, size_t* share_size) {
+#if defined(BOTAN_HAS_PAKE_SPAKE2PLUS)
+   return BOTAN_FFI_VISIT(params, [=](const auto& p) -> int {
+      if(share_size == nullptr) {
+         return BOTAN_FFI_ERROR_NULL_POINTER;
+      }
```
arckoor, line 111: ^

src/lib/ffi/ffi_spake2p.cpp
```diff
@@ -0,0 +1,362 @@
...
+int botan_spake2p_registration_record(botan_spake2p_params_t params,
+                                      botan_rng_t rng,
+                                      const uint8_t secret[],
+                                      size_t secret_len,
+                                      botan_view_ctx ctx,
+                                      botan_view_bin_fn view) {
+#if defined(BOTAN_HAS_PAKE_SPAKE2PLUS)
+   if(any_null_pointers(params, rng, secret)) {
+      return BOTAN_FFI_ERROR_NULL_POINTER;
+   }
```
arckoor, line 176: `secret` sure, but `_t` structs like `params` and `rng` are not explicitly checked anywhere else I think, because `safe_get` takes care of that 🤔

src/lib/ffi/ffi_spake2p.cpp
```diff
@@ -0,0 +1,362 @@
...
+                              const uint8_t prover_id[],
+                              size_t prover_id_len,
+                              const uint8_t verifier_id[],
+                              size_t verifier_id_len,
+                              const uint8_t context[],
+                              size_t context_len) {
+#if defined(BOTAN_HAS_PAKE_SPAKE2PLUS)
+   if(any_null_pointers(prover, params, secret)) {
+      return BOTAN_FFI_ERROR_NULL_POINTER;
+   }
```
arckoor, line 201: ^

src/lib/ffi/ffi_spake2p.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+int botan_spake2p_prover_generate_message(botan_spake2p_prover_t prover,
+                                          botan_rng_t rng,
+                                          botan_view_ctx ctx,
+                                          botan_view_bin_fn view) {
+#if defined(BOTAN_HAS_PAKE_SPAKE2PLUS)
+   return BOTAN_FFI_VISIT(prover, [=](auto& p) -> int {
+      if(rng == nullptr) {
+         return BOTAN_FFI_ERROR_NULL_POINTER;
+      }
```
arckoor: unneeded?

src/lib/ffi/ffi_spake2p.cpp
```diff
@@ -0,0 +1,362 @@
...
+                                         botan_rng_t rng,
+                                         const uint8_t peer_message[],
+                                         size_t peer_message_len,
+                                         botan_view_ctx ctx,
+                                         botan_view_bin_fn view) {
+#if defined(BOTAN_HAS_PAKE_SPAKE2PLUS)
+   return BOTAN_FFI_VISIT(prover, [=](auto& p) -> int {
+      if(any_null_pointers(rng, peer_message)) {
+         return BOTAN_FFI_ERROR_NULL_POINTER;
+      }
```
arckoor: ^

src/lib/ffi/ffi_spake2p.cpp
```diff
@@ -0,0 +1,362 @@
...
+                                const uint8_t prover_id[],
+                                size_t prover_id_len,
+                                const uint8_t verifier_id[],
+                                size_t verifier_id_len,
+                                const uint8_t context[],
+                                size_t context_len) {
+#if defined(BOTAN_HAS_PAKE_SPAKE2PLUS)
+   if(any_null_pointers(verifier, params, record)) {
+      return BOTAN_FFI_ERROR_NULL_POINTER;
+   }
```
arckoor: ^

src/lib/ffi/ffi_spake2p.cpp
```diff
@@ -0,0 +1,362 @@
...
+                                           botan_rng_t rng,
+                                           const uint8_t peer_message[],
+                                           size_t peer_message_len,
+                                           botan_view_ctx ctx,
+                                           botan_view_bin_fn view) {
+#if defined(BOTAN_HAS_PAKE_SPAKE2PLUS)
+   return BOTAN_FFI_VISIT(verifier, [=](auto& v) -> int {
+      if(any_null_pointers(rng, peer_message)) {
+         return BOTAN_FFI_ERROR_NULL_POINTER;
+      }
```
arckoor: ^

src/lib/ffi/ffi_spake2p.cpp
```diff
@@ -0,0 +1,362 @@
...
+}
+
+int botan_spake2p_verifier_verify_confirmation(botan_spake2p_verifier_t verifier,
+                                               const uint8_t confirmation[],
+                                               size_t confirmation_len) {
+#if defined(BOTAN_HAS_PAKE_SPAKE2PLUS)
+   return BOTAN_FFI_VISIT(verifier, [=](auto& v) -> int {
+      if(confirmation == nullptr) {
+         return BOTAN_FFI_ERROR_NULL_POINTER;
+      }
```
arckoor: ^

arckoor (commented):
A couple are missing 😓
The extra checks on `_t` structs kind of throw my whole motivation of "don't change behaviour based on features" out of the window, since everywhere else they are not checked beforehand (and thus do change), but for consistency might still be better to leave those checks to `safe_get`
IDK, your call, the extra checks certainly don't hurt anyone

src/lib/ffi/ffi_spake2p.cpp
randombit: True. I changed this PR to match the existing convention, though I think checking the struct params explicitly is a bit easier to understand and makes the invariants clearer. But changing things in that direction should be a mass PR changing all of FFI rather than in some incremental way.

doc/api_ref/python.rst
randombit: Switched to using autodoc for SPAKE2+ and merged the rst and doc comments

PR #5712, opened 2026-07-04
Improve CDP, AIA and IDP extension decoding and handling

randombit:
Previously the CRLDistributionPoints, AuthorityInformationAccess, and IssuingDistributionPoint extensions were only partially decoded, with many fields being dropped.

Adds new public types DistributionPointName and ReasonFlags which model the corresponding ASN.1 structures used within these extensions.

Various aspects of CRL matching are also improved. A CRL whose IDP scope excludes the certificate or whose name does not match any of the cert's distribution points is now skipped, as if no CRL had been provided, instead of being consulted for revocation status. A reason-partitioned CRL (DP reasons or IDP onlySomeReasons) is still accepted as indicating a revocation, but no longer treated as evidence of non-revocation, since reason accumulation is still not yet implemented.


randombit:
@reneme fyi. Don't know if you have interest in this topic

PR #5717, opened 2026-07-08
Introduce PK_Decryptor(_EME)::ciphertext_length()

reneme:
This allows querying an instance of `PK_Decryptor` for an estimate of the expected ciphertext size given an expected plaintext length. In the general use case that might not be overly useful, assuming that the plaintext size is not known before decryption.

However, for many algorithms (most notably RSA or ElGamal) the expected ciphertext size is actually fixed and only dependent on the key length. For those, the new method facilitates an algorithm-agnostic querying of the expected ciphertext size, e.g. for input validation, pre-allocation and such.

Concretely, it comes in handy in the context of #5686 where we need an RSA-to-KEM Adapter and `PK_KEM_Decryptor` has an `encapsulated_key_length()` accessor that can be implemented conveniently with this new method.

Additionally this fixes two confusing typos in the abstract public interface classes.

### DIFF
--- a/src/lib/prov/pkcs11/p11_rsa.cpp
+++ b/src/lib/prov/pkcs11/p11_rsa.cpp
@@ -141,6 +141,8 @@ class PKCS11_RSA_Decryption_Operation final : public PK_Ops::Decryption {
 
       size_t plaintext_length(size_t /*ctext_len*/) const override { return m_key.get_n().bytes(); }
 
+      size_t ciphertext_length(size_t /*ptext_len*/) const override { return m_key.get_n().bytes(); }
+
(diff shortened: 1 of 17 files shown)


randombit (approved):
Interface looks fine. Unfortunate about having to create new cipher objects in DLIES/ECIES but given we allow CBC or indeed effectively arbitrary transforms here, would be hard to do anything else without some kind of significant refactoring, the shape of which I'm not even sure of....

reneme:
> Unfortunate about having to create new cipher objects in DLIES/ECIES

Yeah, but frankly: I doubt that there's a substantial user base for those. Even less so users that would start actually using this API. 🤷

PR #5721, opened 2026-07-14
FIX: Minor quirks in TLS 1.3

reneme:
While trying to extend the `unit_tls` test harness for TLS 1.3 I found a few minor things that may be worth fixing independently. Unfortunately, extending `unit_tls` proved to be quite cumbersome and will require much more restructuring within the test case than expected. I might get back to that later.

Fixes:

1. When calling `TLS::Channel::received_data()` with a buffer that is aligned at a full record, TLS 1.3 would read all contained records and then report that it needs "5 more bytes" (which are the hypothetical header bytes of the "next" record). That's inconsistent with the behaviour of TLS 1.2 where it would report "0 more bytes" in that case. Now TLS 1.3 also reports "0 more bytes" in that case.
2. TLS 1.3 potentially called `Credentials_Manager::find_cert_chain()` with a vector of desired `cert_key_types` with duplicated entries (for instance `{"RSA", "RSA", "ECDSA"}`). It now deduplicates this wish list.
3. Within the `unit_tls` test case: when calling `Credentials_Manager_Test::find_preshared_keys()` with an empty `identities` vector, an endless recursion would ensue. For TLS 1.2 that code was apparently never reached. It showed up only in my TLS 1.3 experiments. This was meant to be an up-call, I guess.

### DIFF
--- a/src/lib/tls/tls13/msg_certificate_13.cpp
+++ b/src/lib/tls/tls13/msg_certificate_13.cpp
@@ -17,6 +17,7 @@
 #include <botan/tls_extensions.h>
 #include <botan/tls_policy.h>
 #include <botan/x509_key.h>
+#include <botan/internal/stl_util.h>
 #include <botan/internal/tls_reader.h>
 #include <algorithm>
 #include <iterator>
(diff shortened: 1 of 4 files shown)


randombit:
> Unfortunately, extending unit_tls proved to be quite cumbersome

`unit_tls` is kind of a wreck and at this point I hate having to touch it. I am working slowly on an alternative approach (vaguely: think like BoGo/TLS-Anvil but entirely driven by a DSL describing the handshake flows) but it won't be ready for at least a couple more months.

randombit (approved):
lgtm

reneme:
Re:unit_tls -- okay, then I'm going to slaughter it and make it work for what we need for the DTLS 1.3 implementation.

The biggest change I'll probably need is the session resumption that currently spans across the individual test cases. Making that "test case local" takes away quite a bit of brittleness I hope.

Re:new TLS integration test -- 🥳

PR #5725, opened 2026-07-15
Properly split EC_AffinePoint deserialization

randombit:
The current deserialization function accepts compressed, uncompressed, identity element, and the deprecated hybrid encoding. In many contexts only 1 or 2 of these is actually allowed. Add explicit deserializer functions for compressed and uncompressed points.

### DIFF
--- a/src/examples/tls_custom_curves_client.cpp
+++ b/src/examples/tls_custom_curves_client.cpp
@@ -3,6 +3,7 @@
 #include <botan/dl_group.h>
 #include <botan/ec_group.h>
 #include <botan/ecdh.h>
+#include <botan/exceptn.h>
 #include <botan/tls.h>
 
 /**
@@ -49,7 +50,16 @@ class Callbacks : public Botan::TLS::Callbacks {
             std::get<Botan::TLS::Group_Params>(group) == Botan::TLS::Group_Params(0xFE00)) {
             // load the peer's public key of my custom curve
             const auto ec_group = Botan::EC_Group::from_name("numsp256d1");
-            return std::make_unique<Botan::ECDH_PublicKey>(ec_group, Botan::EC_AffinePoint(ec_group, public_value));
+            auto point = Botan::EC_AffinePoint::deserialize_uncompressed(ec_group, public_value);
+
+            if(!point) {
+               // TLS 1.2 allows negotiating compressed points
+               point = Botan::EC_AffinePoint::deserialize_compressed(ec_group, public_value);
+            }
+            if(!point) {
+               throw Botan::Decoding_Error("Invalid ECDH public key encoding");
+            }
+            return std::make_unique<Botan::ECDH_PublicKey>(ec_group, std::move(*point));
          } else {
             // no custom curve used: up-call the default implementation
             return Botan::TLS::Callbacks::tls_deserialize_peer_public_key(group, public_value);
--- a/src/lib/prov/tpm2/tpm2_key.cpp
+++ b/src/lib/prov/tpm2/tpm2_key.cpp
@@ -50,10 +50,13 @@ std::pair<EC_Group, EC_AffinePoint> ecc_pubkey_from_tss2_public(const TPM2B_PUBL
    auto curve = Botan::EC_Group::from_name(curve_name.value());
    // Create an EC_AffinePoint from the x and y coordinates as SEC1 uncompressed point.
    // According to the TPM2.0 specification Part 1 C.8, each coordinate is already padded to the curve's size.
-   auto point = EC_AffinePoint::deserialize(curve,
-                                            concat(std::vector<uint8_t>({0x04}),
-                                                   as_span(public_blob->publicArea.unique.ecc.x),
-                                                   as_span(public_blob->publicArea.unique.ecc.y)));
+
+   auto point = EC_AffinePoint::deserialize_uncompressed(
+      curve,
+      concat<std::vector<uint8_t>>(std::array<uint8_t, 1>{0x04},
+                                   as_span(public_blob->publicArea.unique.ecc.x),
+                                   as_span(public_blob->publicArea.unique.ecc.y)));
+
    if(!point) {
       throw Invalid_Argument("Invalid ECC Point");
    }
--- a/src/lib/pubkey/ecc_key/ec_key_data.cpp
+++ b/src/lib/pubkey/ecc_key/ec_key_data.cpp
@@ -6,10 +6,36 @@
 
 #include <botan/internal/ec_key_data.h>
 
+#include <botan/exceptn.h>
 #include <botan/mem_ops.h>
 #include <botan/rng.h>
 
 namespace Botan {
+
+namespace {
+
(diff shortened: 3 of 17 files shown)


src/examples/tls_custom_curves_client.cpp
```diff
@@ -48,8 +49,13 @@ class Callbacks : public Botan::TLS::Callbacks {
...
             std::get<Botan::TLS::Group_Params>(group) == Botan::TLS::Group_Params(0xFE00)) {
             // load the peer's public key of my custom curve
+            // (TLS uses the uncompressed point encoding)
             const auto ec_group = Botan::EC_Group::from_name("numsp256d1");
-            return std::make_unique<Botan::ECDH_PublicKey>(ec_group, Botan::EC_AffinePoint(ec_group, public_value));
+            auto point = Botan::EC_AffinePoint::deserialize_uncompressed(ec_group, public_value);
+            if(!point) {
+               throw Botan::Decoding_Error("Invalid ECDH public key encoding");
+            }
+            return std::make_unique<Botan::ECDH_PublicKey>(ec_group, std::move(*point));
```
reneme, line 62: This is risky and might become a gotcha for users that base their implementation on this example! TLS 1.2 can negotiate compressed point encoding. In such a case the deserialization would fail and abort the handshake.

That could happen even when we explicitly set the uncompressed point encoding in `tls_generate_ephemeral_key()` above, because the default implementation of `tls12_generate_ephemeral_ecdh_key()` calls `tls_generate_ephemeral_key()` and re-sets the point encoding to whatever it negotiated with the peer.

src/lib/prov/tpm2/tpm2_key.cpp
```diff
@@ -50,10 +50,10 @@ std::pair<EC_Group, EC_AffinePoint> ecc_pubkey_from_tss2_public(const TPM2B_PUBL
...
    // Create an EC_AffinePoint from the x and y coordinates as SEC1 uncompressed point.
    // According to the TPM2.0 specification Part 1 C.8, each coordinate is already padded to the curve's size.
-   auto point = EC_AffinePoint::deserialize(curve,
-                                            concat(std::vector<uint8_t>({0x04}),
-                                                   as_span(public_blob->publicArea.unique.ecc.x),
-                                                   as_span(public_blob->publicArea.unique.ecc.y)));
+   auto point = EC_AffinePoint::deserialize_uncompressed(curve,
+                                                         concat(std::vector<uint8_t>({0x04}),
+                                                                as_span(public_blob->publicArea.unique.ecc.x),
+                                                                as_span(public_blob->publicArea.unique.ecc.y)));
```
reneme: Nit: while you're here anyway. This would avoid an unnecessary allocation:


```suggestion
   auto point =
      EC_AffinePoint::deserialize_uncompressed(
         curve,
         concat<std::vector<uint8_t>>(std::array<uint8_t, 1>{0x04},
                                      as_span(public_blob->publicArea.unique.ecc.x),
                                      as_span(public_blob->publicArea.unique.ecc.y)));
```

(not sure about the formatting)

src/lib/pubkey/ecc_key/ec_key_data.cpp
```diff
@@ -6,11 +6,38 @@
...
+   *    any other value is included in the first octet.
+   */
+   auto pt = EC_AffinePoint::deserialize_uncompressed(group, bytes);
+   if(!pt) {
+      pt = EC_AffinePoint::deserialize_compressed(group, bytes);
+   }
+   if(!pt) {
+      throw Decoding_Error("Failed to deserialize elliptic curve point");
+   }
+   return *pt;
```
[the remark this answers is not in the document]
reneme: The robot has a point. Because you need to unwrap the optional, RVO won't work here.

Perhaps like so, instead (also avoiding the state change in the `pt` out variable):

```C++
EC_AffinePoint decode_ec_public_key_point(const EC_Group& group, std::span<const uint8_t> bytes) {
   /*
   * RFC 5480 section 2.2:
   *    The first octet of the OCTET STRING indicates whether the key is
   *    compressed or uncompressed.  The uncompressed form is indicated
   *    by 0x04 and the compressed form is indicated by either 0x02 or
   *    0x03 (see 2.3.3 in [SEC1]).  The public key MUST be rejected if
   *    any other value is included in the first octet.
   */
   if(auto pt_uncompressed = EC_AffinePoint::deserialize_uncompressed(group, bytes)) {
      return std::move(pt_uncompressed).value();
   } else if(auto pt_compressed = EC_AffinePoint::deserialize_compressed(group, bytes)) {
      return std::move(pt_compressed).value();
   }

   throw Decoding_Error("Failed to deserialize elliptic curve point");
}
```

... strictly speaking that's still not RVO, but at least the return value is move-constructed.

src/lib/tls/tls_callbacks.cpp
```diff
@@ -217,7 +217,17 @@ std::unique_ptr<Public_Key> TLS::Callbacks::tls_deserialize_peer_public_key(
...
+      // hybrid encoding and the identity element are never accepted.
+      auto point = EC_AffinePoint::deserialize_uncompressed(ec_group, key_bits);
+      if(!point) {
+         point = EC_AffinePoint::deserialize_compressed(ec_group, key_bits);
+      }
+      if(!point) {
+         throw Decoding_Error("Invalid ECDH public key encoding");
+      }
+      return std::make_unique<ECDH_PublicKey>(ec_group, std::move(*point));
    }
```
reneme, line 234: Similarly to the proposed change in `ec_key_data.cpp`. IIFE could make this a little nicer, IMHO:

```C++
   if(group_params.is_ecdh_named_curve()) {
      const auto ec_group = EC_Group::from_name(group_params.to_algorithm_spec().value());

      // TLS 1.3 requires uncompressed points (checked when parsing the key
      // share); TLS 1.2 may negotiate the compressed format. The deprecated
      // hybrid encoding and the identity element are never accepted.
      auto point = [&]() -> EC_AffinePoint {
         if(auto pt_uncompressed = EC_AffinePoint::deserialize_uncompressed(ec_group, key_bits);) {
            return std::move(pt_uncompressed).value();
         } else if(auto pt_compressed = EC_AffinePoint::deserialize_compressed(ec_group, key_bits);) {
            return std::move(pt_compressed).value();
         }

         throw Decoding_Error("Invalid ECDH public key encoding");
      }();

      return std::make_unique<ECDH_PublicKey>(ec_group, std::move(point));
   }
```

reneme (commented):
I guess this obsoletes https://github.com/randombit/botan/pull/5356 (which admittedly came out more complex than I had hoped anyway). Don't hesitate to close that if you don't like it. :)

src/examples/tls_custom_curves_client.cpp
randombit: Oops, indeed. Changed to fallback to trying compressed

src/lib/tls/tls_callbacks.cpp
randombit: Better, thanks

PR #5729, opened 2026-07-15
[TLS 1.3] Extend `unit_tls` with support for TLS 1.3

reneme:
As stated [yesterday](https://github.com/randombit/botan/pull/5721#issue-4884217737), this now extends the `unit_tls` integration test with support for TLS 1.3 and hopefully should make it easily extendable for DTLS 1.3 as well.

The biggest change is that we have to abstain from adding custom extensions to the TLS 1.3 ServerHello and HelloRetryRequest because those are designed to be as minimal as possible in TLS 1.3 and our implementation actually checks that there are no unexpected extensions in those messages.

### Pull Request Dependencies

~~Currently, this is stacked on top of #5728. Only the very last commit is actually adding support for TLS 1.3.~~

* #5728 
* #5727 
* #5726 
* #5721

### DIFF
--- a/src/tests/unit_tls.cpp
+++ b/src/tests/unit_tls.cpp
@@ -28,6 +28,7 @@
    #include <botan/tls_server.h>
    #include <botan/tls_session_manager_memory.h>
    #include <botan/tls_session_manager_noop.h>
+   #include <botan/internal/stl_util.h>
    #include <botan/internal/tls_reader.h>
 
    #include <botan/ec_group.h>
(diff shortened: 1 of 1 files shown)


reneme:
Rebased after #5728 got merged. Ready for review/merge now.

src/tests/unit_tls.cpp
```diff
@@ -405,29 +409,62 @@ class TLS_Handshake_Test final {
 
             void tls_modify_extensions(Botan::TLS::Extensions& extn,
                                        Botan::TLS::Connection_Side which_side,
-                                       Botan::TLS::Handshake_Type /* unused */) override {
-               extn.add(new Test_Extension(which_side));  // NOLINT(*-owning-memory)
+                                       Botan::TLS::Handshake_Type msg_type) override {
+               // We don't alter any TLS 1.3 ServerHellos (including the special
+               // HelloRetryRequest), as Botan explicitly rejects any custom
+               // extensions in those messages. Even though RFC 9846 doesn't
+               // explicitly forbid them.
```
randombit, line 416: An interesting question if we should allow them, but I guess that can wait until a specific use case arises

randombit (approved):
lgtm

src/tests/unit_tls.cpp
reneme: Yeah, I did think about that as well and then decided to leave it as-is. For most use cases such custom extensions should go into the EncryptedExtensions message anyway. Except if they somehow influence the key exchange itself. If that ever arises, we can think about loosening this restriction (e.g. by adding a policy configuration for it).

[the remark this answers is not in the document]
reneme: No, RFC 9846 is the new TLS 1.3 RFC revision.

PR #5739, opened 2026-07-17
Updated fuzzer documentation and Coonfiguration

Raghav09445:
Updated fuzzer documentation and Configuration

### DIFF
--- a/doc/dev_ref/fuzzing.rst
+++ b/doc/dev_ref/fuzzing.rst
@@ -9,16 +9,19 @@ the library.
 Fuzzing with libFuzzer
 ------------------------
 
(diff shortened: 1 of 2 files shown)


randombit:
CI is failing - format is relevant to this change, the ACVP+macOS failures are due to flakes that have been addressed on master. Rebase against master to pick up the fixes + run clang-format (the failing job prints the diff so you can just apply it directly, otherwise you need exactly version 17 of `clang-format`). Otherwise lgtm

randombit:
The fuzzer job is also failing, definitely relevant - I think you just need to add an include of `botan/rng.h` so the base class definition is visible

randombit:
Feel free to reopen if you are going to come back to this

PR #5746, opened 2026-07-19
Add HashFunction::security_level

randombit:


src/lib/prov/commoncrypto/commoncrypto_hash.cpp
```diff
@@ -47,6 +47,14 @@ class CommonCrypto_HashFunction final : public HashFunction {
 
       size_t output_length() const override { return m_info.digestLength; }
 
+      size_t security_level() const override {
+         if(m_info.name == "SHA-1") {
+            // Collision attacks with cost ~2^61 are known (Leurent and Peyrin, 2020)
+            return 61;
+         }
+         return 4 * output_length();
```
reneme: Nit: If we have a default implementation, we might as well up-call into it?
(also applies to other places)


```suggestion
         return HashFunction::output_length();
```

reneme (approved):

src/lib/prov/commoncrypto/commoncrypto_hash.cpp
randombit: Assume you meant `HashFunction::security_level` here, but yes

PR #5749, opened 2026-07-21
Output filename and line number with deprecation messages on MSVC

lanurmi:
Unlike other compilers, MSVC does not by default output the filename with pragma messages. So the build log would literally show deprecation warnings as just "this header is deprecated", with no hint about what header, or whose header, i.e. not even a hint that this message is coming from Botan.


randombit (approved):
Thanks

PR #5750, opened 2026-07-21
Rename URI::parse to ::from_string

arckoor:
All the other types, i.e. `EmailAddress`, `DNSName`, `IPv4Address` and `IPv6Address` name their string creation methods `::from_string`, except for `URI`.
This makes it annoying to write generic code that works for all of them.

In tests I changed direct mentions of `::parse` to `::from_string`, but otherwise kept the parse terminology

### DIFF
--- a/src/tests/unit_x509.cpp
+++ b/src/tests/unit_x509.cpp
@@ -498,14 +498,14 @@ Test::Result test_x509_encode_authority_info_access_extension() {
 
    // CA Issuer information
    const std::vector<Botan::URI> ca_issuers = {
-      Botan::URI::parse("http://www.d-trust.net/cgi-bin/Bdrive_Test_CA_1-2_2017.crt").value(),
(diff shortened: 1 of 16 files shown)


src/tests/unit_x509.cpp
```diff
@@ -2143,7 +2143,7 @@ std::vector<std::string> get_sig_paddings(const std::string& sig_algo, const std
          "PKCS1v15(" + hash + ")",
    #endif
    #if defined(BOTAN_HAS_EMSA_PSS)
-            "PSS(" + hash + ")",
```
arckoor, line 2146: I won't pretend to understand why the formatter thinks this needs to be indented twice when the above one is only indented once, at least my local formatter doesn't agree, but it's also clang-format-19 instead of -17 used by CI

randombit (approved):
Makes sense, thanks

Issue #2492, opened 2020-11-13
TLS FFI API


winterland1989:
Just to make sure, is there any plan to add TLS APIs to ffi.h ? I'm trying add botan binding to Haskell and I'm not sure if I sure add wrappers around current C++ APIs.

randombit:
This is definitely the plan in the long term, so it can also be used from Rust/Python/etc, but it's not likely to happen in say the next 6 months.

bitberlin:
TLS support in FFI API would be very fine! This is currently the only reason why i can't use botan for my project.

benrushcc:
Hi, I'm checking in on this old issue about exposing TLS APIs via the FFI.

I noticed that there is a refs/heads/jack/ffi-tls branch in the repository, which seems to suggest some related work might be in progress. Is that branch part of the plan to address this?

I'm currently evaluating Botan for a project that needs TLS support through C bindings (for Java). Knowing the current status and roadmap would be very helpful for my planning.

Thanks!

randombit:
@benrushcc That particular branch is not really complete or ready to go in any way. There is an effort currently underway to get funding to support the completion of this significant project. We'll probably know more in the next few months. But even if funding comes through it likely will not be available until - at the absolute most optimistic - sometime in 2027.

moritzschmitt:
@randombit We are interested in this issue because we would like to build a Go library that uses Botan for TLS through the C API. We have some experience with the FFI already (most recently #5903).

Before we invest time, I would like to ask whether a community contribution is wanted here, or whether it would conflict with the funding effort you mentioned above. We do not want to get in the way of a funded project.

If a contribution is welcome, we would start with a proposal rather than code. This would be a draft of the header and the reasoning behind the main decisions (I/O model, credentials, session handling, error reporting), meant as the starting point for a design discussion. So far we have only a rough idea of how this could work. We can post it here or as a GitHub Discussion, whichever you prefer.

randombit:
@moritzschmitt Just got word that the funding we had hoped for which would have covered FFI-TLS along with some other work will not occur (alas), so at this point a contribution here would be completely welcome. Let's use this issue for design discussions.

moritzschmitt:
@randombit Sorry to hear about the funding, but we will surely be able to do without.

Here is a first sketch of how we could approach it. The idea is a thin C shim over the existing C++ TLS API, without any new protocol code. Please read it critically. Our understanding of the TLS code is new and still brittle, so we may well have got details wrong.

[...]

PR #5753, opened 2026-07-21
Add certificate creation to FFI

arckoor:
Adds a "minimal" certificate builder to the FFI (extensions I'll do in a separate PR)

- First commit is #4996 squashed into one commit
- Second commit are the changes needed to make it build on master, along with extra fixes. My hope is that this can be cherry picked and used in #4996, either in whole or at least in part
- Last commit is what this PR would actually like to add

I would've liked to test more in python, but `X509Cert` doesn't provide most of the things added in #5188, and I don't currently have the time to add them myself (cc @reneme maybe, but I'll do more extensive testing on the cpp side anyway, so not "critical")

### DIFF
--- a/src/lib/ffi/ffi_cert.cpp
+++ b/src/lib/ffi/ffi_cert.cpp
@@ -1,5 +1,6 @@
 /*
 * (C) 2015,2017,2018 Jack Lloyd
+* (C) 2026 Dominik Schricker
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
@@ -7,6 +8,11 @@
 #include <botan/ffi.h>
 
 #include <botan/assert.h>
+#include <botan/dns_name.h>
+#include <botan/email.h>
+#include <botan/ipv4_address.h>
+#include <botan/ipv6_address.h>
+#include <botan/uri.h>
 #include <botan/internal/ffi_cert.h>
 #include <botan/internal/ffi_pkey.h>
 #include <botan/internal/ffi_rng.h>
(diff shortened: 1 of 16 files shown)


moritzschmitt:
All the new declarations in `ffi.h` are tagged `BOTAN_FFI_EXPORT(3, 13)`, but 3.13.0 was tagged on 2026-08-13 and this branch is based on a commit after that tag. As tagged, the header claims these functions exist in the released 3.13.0. They need to be `(3, 14)`.

Since this would be the first 3.14 FFI addition, the FFI API date also needs bumping in the four places listed at the top of `src/lib/ffi/info.txt` (`info.txt`, `BOTAN_FFI_API_VERSION` in `ffi.h`, `botan_ffi_supports_api` in `ffi.cpp`, `BOTAN_FFI_VERSION` in `botan3.py`). That is how the 3.13 date was introduced in af3887268.

Same applies to the RFC 3779 additions in #5853.

src/lib/ffi/ffi_cert.cpp
```diff
@@ -1032,6 +1067,339 @@ const char* botan_x509_cert_validation_status(int code) {
...
+   });
+#else
+   BOTAN_UNUSED(subject_req, issuing_cert, issuing_key, rng, not_before, not_after, serial_number, hash_fn, padding);
+   return BOTAN_FFI_ERROR_NOT_IMPLEMENTED;
+#endif
+}
+
 int botan_x509_crl_load_file(botan_x509_crl_t* crl_obj, const char* crl_path) {
    if(Botan::any_null_pointers(crl_obj, crl_path)) {
       return BOTAN_FFI_ERROR_NULL_POINTER;
```
moritzschmitt, line 1405: The `#else` branches reference parameter names that don't exist, so a build with the FFI but without the x509 module fails. `ffi` lists `x509` as optional in its `info.txt`, so that is a supported configuration. CI doesn't cover it. Reproducer on this branch:

    ./configure.py --disable-modules=x509 --without-documentation
    make build/obj/lib/ffi_cert.o

    src/lib/ffi/ffi_cert.cpp:1251:43: error: use of undeclared identifier 'key'
    src/lib/ffi/ffi_cert.cpp:1398:30: error: use of undeclared identifier 'issuing_cert'
    src/lib/ffi/ffi_cert.cpp:1398:44: error: use of undeclared identifier 'issuing_key'

Line 1251 should name `pubkey`, line 1398 `ca_cert` and `ca_key`. The `#else` branches of `into_self_signed_cert` (line 1212) and `into_cert` (line 1251) also omit `serial_number`, which produces `-Wunused-parameter` warnings in the same build.

arckoor:
Yeah, all still a little rough around the edges, only really tested the happy path for now to be able to provide feedback for the cert builder PR (hence the draft status). The `FFI_EXPORT` macros I will update when it seems reasonable that this PR will make it into a release, it might not even be 3.14.

src/python/botan3.py
```diff
@@ -2491,8 +2571,10 @@ class X509Cert: # pylint: disable=invalid-name
 
     def __init__(self, filename: str | None = None, buf: bytes | None = None):
         """Load a certificate from either a file or a bytestring, but not both"""
-        self.__obj = c_void_p(0)
-        self.__obj = _load_buf_or_file(filename, buf, _DLL.botan_x509_cert_load_file, _DLL.botan_x509_cert_load)
+        if not filename and not buf:
```
moritzschmitt, line 2574: I ran into the same problem in my TLS FFI work for #2492: the Python binding needs to wrap an `X509Cert` around a handle the library just produced, and `X509Cert.__init__` only knows how to load from a file or a buffer.

I'd suggest not solving it by making `X509Cert()` (and `PKCS10Request()`) legal. With this change a caller who forgets the argument gets an object with a null handle, and every later call on it fails with `BOTAN_FFI_ERROR_NULL_POINTER` instead of the constructor raising. It also changes `X509Cert(buf=b"")`: `_load_buf_or_file` checks `is None`, so an empty buffer used to reach `botan_x509_cert_load` and raise a decoding error, whereas `not buf` now yields the null-handle object silently.

I know `X509CRL` already does this since 7ab352a20, so this is consistent with that. But an alternative that keeps the public constructor strict is a private classmethod that adopts a handle, which is what I ended up with on my branch:

    @classmethod
    def _from_handle(cls, obj: c_void_p) -> X509Cert:
        """Wrap an FFI certificate handle, taking ownership of it"""
        cert = cls.__new__(cls)
        cert._X509Cert__obj = obj
        return cert

`into_self_signed_cert` and friends would then do `obj = c_void_p(0)`, pass `byref(obj)` to the FFI call, and `return X509Cert._from_handle(obj)`. `PublicKey`/`PrivateKey` take the other route and accept an optional `obj` in `__init__`, so either is defensible; I'd just rather not extend the null-handle constructor to two more classes.

PR #5762, opened 2026-07-27
Add missing doc comments to types and functions in utils

randombit:


src/lib/utils/data_src.h
```diff
@@ -180,13 +261,24 @@ class BOTAN_PUBLIC_API(2, 0) DataSource_Stream final : public DataSource {
...
       DataSource_Stream(const DataSource_Stream&) = delete;
+
+      /// Stream data sources are not movable
       DataSource_Stream(DataSource_Stream&&) = delete;
+
+      /// Stream data sources are not copy assignable
       DataSource_Stream& operator=(const DataSource_Stream&) = delete;
+
+      /// Stream data sources are not move assignable
       DataSource_Stream& operator=(DataSource_Stream&&) = delete;
```
reneme: Not sure we want such comments.

randombit: I guess I don't see the problem? That said they will never be rendered so I guess redundant.

randombit: Removed

PR #5765, opened 2026-07-27
Add doc comments to Python bindings that had none

randombit:
Also rename `handle_` to `_handle`, and insert `_` prefixes on utility code (like the view callback handlers) which is not intended for use by applications, matching our general underscore-prefix convention for internal interfaces.

### DIFF
--- a/src/python/botan3.py
+++ b/src/python/botan3.py
@@ -35,13 +35,16 @@ from enum import IntEnum
 # 3.12.0 - EcScalar/EcPoint, DRBG
 # 3.11.0 - XOF API
 # 3.10.0 - introduced botan_pubkey_load_ec*_sec1()
-BOTAN_FFI_VERSION = 20260811
+BOTAN_FFI_VERSION = 20260811  #: The minimum FFI API version this module requires of the library.
 
 
 class BotanException(Exception):
     """Base exception for all exceptions raised from this module"""
 
     def __init__(self, message, rc=0):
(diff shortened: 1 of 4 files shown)


src/python/botan3.py
```diff
@@ -35,13 +35,16 @@
...
+BOTAN_FFI_VERSION = 20260811  #: The minimum FFI API version this module requires of the library.
 
 
 class BotanException(Exception):
     """Base exception for all exceptions raised from this module"""
 
     def __init__(self, message, rc=0):
+        """Create an exception with ``message``. If ``rc`` is a nonzero library error
+        code, the description of that error plus the library's most recent exception
+        message are appended."""
```
arckoor, line 47: That won't  show up anywhere, has to be in the class description itself, see https://github.com/randombit/botan/pull/5233#issue-3806555838

randombit: When I thought about this more, the behavior seemed wrong to me. For example `help` in the Python REPL prints the docs for `__init__` just fine. It seems Sphinx autodoc by default extracts just the class doc, but you can configure it to take both the class and `__init__` docs using `autoclass_content = "both"`, so I have configured Sphinx thus.

Not sure about your IDE though. Maybe it has a similar class doc vs `__init__` doc toggle?

arckoor: The sphinx one is good to know, that definitely makes things clearer then. Perhaps the things I moved in the autogen PR could / should be reverted then, but maybe also not a huge deal.

PR #5767, opened 2026-07-27
Refactor: Handle TLS 1.3 Compat Mode via the TLS state machine

reneme:
This refactors [the TLS 1.3 middlebox compatibility mode](https://www.rfc-editor.org/rfc/rfc9846.html#name-middlebox-compatibility-mod) handling to fully rely on the TLS state machine. Essentially the state machine now calls `maybe_handle_compatibility_mode()` in all locations where a dummy ChangeCipherSpec might need to be emitted.

Before, this handling was partially implemented within the fairly generic `send_record()` method in the `TLS::Channel` base class. That didn't fit the level of abstraction within `send_record()` and resulted in hard-to-understand control flow. Also, while adding DTLS (#5000) this code smell became even more involved.

So here's a refactoring to clean this up.

### DIFF
--- a/src/lib/tls/tls13/tls_client_impl_13.cpp
+++ b/src/lib/tls/tls13/tls_client_impl_13.cpp
@@ -63,15 +63,7 @@ Client_Impl_13::Client_Impl_13(const std::shared_ptr<Callbacks>& callbacks,
       preserve_client_hello(msg);
    }
 
-   // RFC 8446 Appendix D.4
-   //    If not offering early data, the client sends a dummy change_cipher_spec
-   //    record [...] immediately before its second flight. This may either be before
-   //    its second ClientHello or before its encrypted handshake flight.
-   //
-   // TODO: don't schedule ccs here when early data is used
-   if(policy->tls_13_middlebox_compatibility_mode()) {
-      m_handshake->should_send_ccs = true;
-   }
+   maybe_handle_compatibility_mode(Compat_Mode_Situation::AfterSendingFirstClientHello);
 
    m_handshake->transitions.set_expected_next({Handshake_Type::ServerHello, Handshake_Type::HelloRetryRequest});
 }
@@ -395,6 +387,7 @@ void Client_Impl_13::handle(const Hello_Retry_Request& hrr) {
 
    callbacks().tls_examine_extensions(hrr.extensions(), Connection_Side::Server, Handshake_Type::HelloRetryRequest);
 
+   maybe_handle_compatibility_mode(Compat_Mode_Situation::BeforeSendingSecondClientHello);
    send_handshake_message(std::reference_wrapper(ch));
 
    // RFC 8446 4.1.4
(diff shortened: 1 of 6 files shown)


src/lib/tls/tls13/tls_client_impl_13.cpp
```diff
@@ -774,8 +768,64 @@ std::optional<std::string> Client_Impl_13::external_psk_identity() const {
...
+void Client_Impl_13::maybe_handle_compatibility_mode(Compat_Mode_Situation situation) {
+   // RFC 9846 E.4
+   //    This "compatibility mode" is partially negotiated: the client can opt
+   //    [in] or not [...].
+   if(!policy().tls_13_middlebox_compatibility_mode()) {
+      return;
+   }
+
+   // RFC 9846 E.4
+   //    Either side can send change_cipher_spec at any time during the handshake.
```
randombit: This quote seems out of context for the logic here. The situation in the code is this: we don't have a handshake state, which implies that we have already sent a Finished message, and sending a CCS after Finished is not valid. The RFC quote kind of points to one side of this (when a CCS can be sent), without saying when it cannot.

Maybe more relevant?

RFC 9846 Section 5

> An implementation [...] which receives a protected change_cipher_spec record MUST abort the handshake with an "unexpected_message" alert.

src/lib/tls/tls13/tls_server_impl_13.cpp
```diff
@@ -240,17 +243,34 @@ void Server_Impl_13::maybe_handle_compatibility_mode() {
...
-      (policy().tls_13_middlebox_compatibility_mode() || client_requested_compatibility_mode)) {
-      send_dummy_change_cipher_spec();
+   switch(situation) {
+      case Compat_Mode_Situation::AfterSendingFirstServerHello:
+      case Compat_Mode_Situation::AfterSendingHelloRetryRequest:
+         // RFC 9846 E.4
+         //    The server sends a dummy change_cipher_spec record immediately after
+         //    its first handshake message. This may either be after a ServerHello or
+         //    a HelloRetryRequest.
+         send_dummy_change_cipher_spec();
```
randombit, line 258: So I can't comment on `send_dummy_change_cipher_spec` directly, but I'd like it to have an check that we haven't sent our finished yet, since otherwise sending the CCS would violate (from Section 5 of the RFC)

> If an implementation detects a change_cipher_spec record received before the first ClientHello message or after the peer's Finished message, it MUST be treated as an unexpected record type

I'd be open to either just ignoring it (silently not sending the CCS) or throwing `Internal_Error` and failing the connection.

randombit (approved):
Overall LGTM; one comment nit and one missing paranoid check.

src/lib/tls/tls13/tls_server_impl_13.cpp
reneme: I did use `BOTAN_STATE_CHECK(!is_handshake_complete())` for the check. That is a little less tight than checking for the client's or server's Finished message but `send_dummy_change_cipher_spec` resides in the `TLS::Channel` where I don't have easy access to the handshake states (because they live in the Client/Server).

We could introduce a new protected virtual method (i.e. `is_ccs_valid_now()`) that covers this check exactly, but I did consider that overkill for a state machine sanity check.

randombit:
@reneme TLS-Anvil broke overnight after this merged can you take a look?

PR #5768, opened 2026-07-27
Use Ubuntu 26.04 image where possible in GHA CI

randombit:


randombit:
Skips MIPS builds (no MIPS cross compiler in 26.04 it seems), format (will require a dedicated PR due to clang-format change), and the old GCC/Clang builds

randombit:
@thillux maybe of interest to you, it seems `edsm-pr` does not work correctly when run in the Ubuntu 26.04 runner. Unclear to me what's happening here. For now I've just removed some of the esdm-pr related testing...

thillux:
> @thillux maybe of interest to you, it seems `edsm-pr` does not work correctly when run in the Ubuntu 26.04 runner. Unclear to me what's happening here. For now I've just removed some of the esdm-pr related testing...

Thanks for the heads up. I'm currently working on ESDM 1.2.4 and will check with this version for the esdm-pr tests.

randombit:
@thillux it seems to be affecting ESDM on the 24.04 images as well, so likely related to the very recent upgrade from esdm 1.2.0 to 1.2.3.

PR #5771, opened 2026-07-28
Support ECDSA with brainpool in TLS 1.3

### LINKED ISSUE #3811: TLS `SignatureScheme` lacks support for Brainpool
Currently, the enum `TLS::Signature_Scheme::Code` does not contain code points for brainpool signatures [RFC 8734](https://www.rfc-editor.org/rfc/rfc8734.html) (0x081A-0x081C). Instead, the implementation accepts the usage of Brainpool keys with the code points for `ECDSA_SHAxxx` (0x0403, 0x0503, 0x0603).

As defined in [RFC 8446 Section 4.2.3](https://www.rfc-editor.org/rfc/rfc8446#page-42), the code points 0x0*03 are meant for the NIST curves exclusively. Instead, we should introduce the brainpool-specific code points from RFC 8734 and fix the validation code that ensures proper association.

See also:

* `TLS::Signature_Scheme::is_suitable_for(Private_Key)`
* `Handshake_State::parse_sig_format()` (TLS 1.2)
* `Certificate_Verify_13::choose_signature_scheme()`

reneme:
This is described in [RFC 8734](https://www.rfc-editor.org/rfc/rfc8734.html) and is essentially straight-forward: Adding explicit code points for ECDSA with Brainpool, similar to the additions in #5691. As before I wouldn't restrict those code points for usage in TLS 1.3 explicitly. Perhaps we shouldn't enable them by default, though.

_Currently, that's really just a prove of concept (which succeeds in performing a handshake against a Bouncycastle server that is configured with a brainpool-based certificate)._ I need to go through this RFC and RFC 8446 once more to see if anything else has to be done (such as restricting the standard ECDSA code points to the NIST curves).

It would be nice to land that before Botan 3.13, if possible.

Closes #3811

### DIFF
--- a/src/lib/tls/tls_signature_scheme.cpp
+++ b/src/lib/tls/tls_signature_scheme.cpp
@@ -39,6 +39,10 @@ const std::vector<Signature_Scheme>& Signature_Scheme::all_available_schemes() {
       RSA_PSS_SHA256,
       RSA_PSS_SHA512,
 
+      ECDSA_BRAINPOOL384R1_TLS13_SHA384,
+      ECDSA_BRAINPOOL256R1_TLS13_SHA256,
+      ECDSA_BRAINPOOL512R1_TLS13_SHA512,
+
       RSA_PKCS1_SHA384,
       RSA_PKCS1_SHA512,
       RSA_PKCS1_SHA256,
@@ -84,6 +88,16 @@ Signature_Scheme Signature_Scheme::from_string(std::string_view str) {
       return RSA_PSS_SHA512;
    }
 
+   if(str == "ECDSA_BRAINPOOL256R1_TLS13_SHA256") {
+      return ECDSA_BRAINPOOL256R1_TLS13_SHA256;
+   }
+   if(str == "ECDSA_BRAINPOOL384R1_TLS13_SHA384") {
+      return ECDSA_BRAINPOOL384R1_TLS13_SHA384;
+   }
+   if(str == "ECDSA_BRAINPOOL512R1_TLS13_SHA512") {
+      return ECDSA_BRAINPOOL512R1_TLS13_SHA512;
+   }
+
    // Parse signature schemes passed as hexadecimal code points (e.g. "0x081A")
    if(str.size() == 6 && str.starts_with("0x")) {
       try {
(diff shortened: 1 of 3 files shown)


randombit:
CI is failing, but yeah seems fine. BTW 3.13.0 would normally be Aug 4 but I pushed it a week so that any findings from Patch The Planet can get addressed prior to the release, rather than right after a release. So should be plenty of time to get this in.

reneme:
> CI is failing, but yeah seems fine.

Looks better now. It failed because `Signature_Scheme::is_suitable_for()` sanity-checked key lengths against the hash algorithm for RSA as well and thus rejected all schemes.

reneme:
I'm happy with this patch as-is for now. A few lessons learned:

1. There's currently **no functional escape hatch to let users provide TLS signature schemes** that aren't implemented or supported by Botan. If one overrides `Policy::allowed_signature_schemes()` and returns a "custom" code point, the `TLS::Signature_Scheme::is_available()` (and potentially others) will treat it as "unavailable" and the TLS implementation hard-rejects it therefore.
2. **`Signature_Scheme::ECDSA_SHA256` (and friends) should be called `::ECDSA_SECP256R1_SHA256` (and respectively)** because (at least in TLS 1.3) these code points are explicitly meant for the NIST curves. In TLS 1.2 the 16-bit code points [are considered as two 8-bit code points](https://www.rfc-editor.org/info/rfc5246/#section-7.4.1.4.1) for "SHA-xxx" and "ECDSA" respectively.
#5772
3. The **`TLS::Text_Policy` does not override `allowed_signature_schemes()` or `accepted_signature_schemes()`**. Instead, the scheme lists are determined from `allowed_signature_method()` (i.e. "RSA", "ECDSA", ...) and `allowed_signature_hash()`. That was fine for the fairly liberal TLS 1.2 interpretation of the signature algorithms. But for TLS 1.3, currently the text policy cannot restrict the ECDSA schemes to just use NIST curves for instance.
#5773
4. Currently there's **no proper test for a handshake with a Brainpool-based certificate**.

All of the above are probably worth further investigation.

reneme:
Rebased after #5773 got merged (and adapted `Signature_Scheme::from_string()` accordingly).

randombit (approved):

PR #5772, opened 2026-07-29
Explicitly state Signature_Scheme::ECDSA_SECPxxxR1_TLS13_SHAxxx

reneme:
For TLS 1.3 the implemented signature scheme codepoints for ECDSA imply "on the NIST curves" explicitly. Other concrete code points exist for ECDSA on the Brainpool curves (GH #5771) paired with the respective hash functions.

In contrast, for TLS 1.2 these code points are interpreted as ECDSA "on any curve" paired with the specified hash function. I think it makes sense to reflect this distinction in the `Signature_Scheme` enum definition. However, since we are continuing to support TLS 1.2 we can't really deprecate the existing enum entries, in my opinion. Instead I just introduced aliases and some documentation explaining this mess...

I mean this as a suggestion, addressing one of the points [here](https://github.com/randombit/botan/pull/5771#issuecomment-5115865269), hence the "draft" status. I'm definitely open for better suggestions.


reneme:
Rebased after #5773 was merged. Also saw another opportunity to simplify the RFC8448 test case.

reneme:
Rebased after #5771 was merged.

randombit (approved):

PR #5773, opened 2026-07-29
`TLS::Text_Policy::allowed_signature_schemes()`

reneme:
This adds overrides for `allowed_signature_schemes` and `acceptable_signature_schemes` in `TLS::Text_Policy` (addressing an API short-coming pointed out [here](https://github.com/randombit/botan/pull/5771#issuecomment-5115865269)). That way, users can specify exactly what signature schemes they wish to use. Before, users could only specify the allowed hash function and allowed signature algorithm. That made it impossible to restrict ECDSA to just the brainpool curves in TLS 1.3, for instance.

As a by-product this allows simplifying the RFC 8448 test somewhat by using this new policy feature.


randombit (approved):
lgtm

PR #5782, opened 2026-08-02
Disable finite field DH key exchange in the default TLS policy

randombit:
Support for FFDHE in TLS has been deprecated in Botan since 2.10.0, all the way back in 2019. Realistically, it should have been removed in 3.0. But alas.

More recently, RFC 10015 deprecates support at the specification level.

Given this, plus the fact that FFDHE is certainly going away in Botan4, disable it from the default TLS policies.


FAlbertDev (approved):

reneme (approved):
Techically unrelated: The `allowed_key_exchange_methods()` policy is a little confusing I find. From what I understand it really just has an effect on TLS 1.2, right? But from a user's perspective I would expect that "KEM" (and perhaps "HYBRID") should also show up there for the PQ-safe variants in TLS 1.3.

Probably something for another day.

randombit:
> Techically unrelated: The allowed_key_exchange_methods() policy is a little confusing I find.

I don't *think* that's true - `Policy::ciphersuite_list` calls `allowed_key_exchange_methods` and `ciphersuite_list` is used in both 1.2 and 1.3. That said the Policy surface is huge and with some weird overlaps and intersections. I'd like to try cleaning this up for Botan4, though I don't have any specific proposal at the moment.

PR #5783, opened 2026-08-02
DTLS bug fixes and hardenings

randombit:
Hardens DTLS against misbehaving peers and networks. The individual commits are fairly well contained. The main themes of the overall branch are

* Offpath attackers injecting packets should not be able to successfully disrupt an active connection or in-progress handshake.
* Avoid denial of service from a misbehaving peer. Prevent fragmented handshake messages from consuming unbounded memory. Bound retransmissions and eventually give up rather than transmitting forever. Discard state that is no longer needed.
* Avoid allowing unsafe configurations. It is now required for DTLS servers to implement `tls_peer_network_identity` and to have a non-empty DTLS cookie secret. CBC mode and CCM-8 are both outright prohibited in DTLS; CBC due to the residual Lucky13 timing channel and CCM-8 due to the short MAC which allows packet forgery given sufficient attempts.

This also extends the BoGo shim to implement the clock manipulation logic and enables a number of DTLS retransmission tests.

@reneme would appreciate a review, I know timeline is short but I think this is a must-merge for 3.13, I would rather delay the release by another week than not ship.
@Thiesius fyi

### DIFF
--- a/src/lib/tls/tls12/tls_channel_impl_12.cpp
+++ b/src/lib/tls/tls12/tls_channel_impl_12.cpp
@@ -63,6 +63,27 @@ void prune_old_cipher_states(std::map<uint16_t, T>& states) {
    }
 }
 
+// Run fn, absorbing the exceptions that report malformed input when `absorb`
+// is set. Used where the input is unauthenticated epoch-zero data that must
+// not be able to tear down an established association. Only TLS_Exception and
+// Decoding_Error are absorbed: an Internal_Error from one of the reassembly
+// accounting assertions, or a failed allocation, still propagates rather than
+// leaving the association running on state those assertions exist to protect.
+template <typename F>
+void absorb_malformed_input_errors(bool absorb, F fn) {
+   try {
+      fn();
+   } catch(const TLS_Exception&) {
+      if(!absorb) {
+         throw;
+      }
+   } catch(const Decoding_Error&) {
+      if(!absorb) {
+         throw;
+      }
+   }
+}
+
 }  // namespace
 
 Channel_Impl_12::Channel_Impl_12(const std::shared_ptr<Callbacks>& callbacks,
@@ -636,16 +657,14 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
    const auto process_retransmitted_record = [&] {
       BOTAN_ASSERT(m_active_state.has_value(), "Have active DTLS association for retransmission");
       BOTAN_ASSERT_NONNULL(m_active_state->dtls_handshake_io());
-      try {
+      // Epoch-zero records are unauthenticated and may be spoofed, so a
+      // malformed one must not tear down an established association.
+      const bool unauthenticated = (record_sequence >> 48) == 0;
+
+      absorb_malformed_input_errors(unauthenticated, [&] {
          m_active_state->dtls_handshake_io()->add_retransmitted_record(
             record.data(), record.size(), record_type, record_sequence);
-      } catch(...) {
-         // Epoch-zero records are unauthenticated and may be spoofed. Invalid
-         // retransmissions must not tear down an established association.
-         if((record_sequence >> 48) > 0) {
-            throw;
-         }
-      }
+      });
    };
 
    if(!m_pending_state) {
@@ -664,7 +683,6 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
       if(m_is_datagram && !epoch0_restart) {
          if(m_sequence_numbers) {
             const uint16_t epoch = record_sequence >> 48;
-
             const uint16_t current_epoch = sequence_numbers().current_read_epoch();
             if(epoch == current_epoch) {
                // Either endpoint can initiate renegotiation from FINISHED:
@@ -692,21 +710,36 @@ void Channel_Impl_12::process_handshake_ccs(const secure_vector<uint8_t>& record
 
    // May have been created in above conditional
    if(m_pending_state) {
-      m_pending_state->handshake_io().add_record(record.data(), record.size(), record_type, record_sequence);
+      // An epoch-zero record is unauthenticated. Once an association is
+      // established, one arriving during a pending renegotiation must not be
+      // able to destroy it, exactly as for the no-pending-handshake path above.
+      // Without this a single forged CCS or handshake fragment tore down the
+      // active association and the renegotiation along with it.
+      //
+      // Delivery is inside the guard as well as reassembly. A bare 12-byte
+      // header declaring a zero-length message reassembles cleanly and only
+      // fails when the message itself is parsed or dispatched, which reaches
+      // the same teardown by a later route.
+      const bool unauthenticated_against_active_association =
+         m_is_datagram && (record_sequence >> 48) == 0 && m_active_state.has_value();
 
-      while(auto* pending = m_pending_state.get()) {
-         auto msg = pending->get_next_handshake_msg(policy().maximum_handshake_message_size());
+      absorb_malformed_input_errors(unauthenticated_against_active_association, [&] {
+         m_pending_state->handshake_io().add_record(record.data(), record.size(), record_type, record_sequence);
 
-         if(msg.first == Handshake_Type::None) {  // no full handshake yet
-            break;
+         while(auto* pending = m_pending_state.get()) {
+            auto msg = pending->get_next_handshake_msg(policy().maximum_handshake_message_size());
+
+            if(msg.first == Handshake_Type::None) {  // no full handshake yet
+               break;
+            }
+
+            process_handshake_msg(*pending, msg.first, msg.second, epoch0_restart);
+
+            if(!m_pending_state) {
+               break;
+            }
          }
-
-         process_handshake_msg(*pending, msg.first, msg.second, epoch0_restart);
-
-         if(!m_pending_state) {
-            break;
-         }
-      }
+      });
    }
 }
 
--- a/src/lib/tls/tls12/tls_channel_impl_12.cpp
+++ b/src/lib/tls/tls12/tls_channel_impl_12.cpp
@@ -448,9 +448,11 @@ void Channel_Impl_12::activate_session() {
       map_remove_if(not_current_epoch, m_read_cipher_states);
    }
 
-   // In a full handshake the server sends the terminal flight; in an
-   // abbreviated handshake the client does. Both endpoints retain handshake
-   // sequence state, but only the terminal sender replays its outgoing flight.
+   // RFC 6347 4.2.4: "the node that transmits the last flight (the server in an
+   // ordinary handshake or the client in a resumed handshake) MUST respond to a
+   // retransmit of the peer's last flight with a retransmit of the last
+   // flight." Both endpoints retain handshake sequence state, but only that
+   // node replays its outgoing flight.
    const bool sent_terminal_dtls_flight = m_is_datagram && (m_is_server == (state.server_hello_done() != nullptr));
 
    if(m_is_datagram) {
--- a/src/lib/tls/tls12/tls_channel_impl_12.cpp
+++ b/src/lib/tls/tls12/tls_channel_impl_12.cpp
@@ -259,6 +259,10 @@ Handshake_State& Channel_Impl_12::create_handshake_state(Protocol_Version versio
       }
    }
 
+   // Read epochs at or below this one belong to the association already in place,
+   // so application data under them stays deliverable while this handshake runs.
+   // Anything above it is this handshake's own, unauthenticated until its
+   // Finished. See the application-data gate in from_peer.
    m_epochs_before_latest_renegotiation = Epochs_Before_Latest_Renegotiation{sequence_numbers().current_read_epoch(),
                                                                              sequence_numbers().current_write_epoch()};
 
@@ -639,7 +643,34 @@ size_t Channel_Impl_12::from_peer(std::span<const uint8_t> data) {
                throw TLS_Exception(Alert::UnexpectedMessage, "Received application data after connection closure");
             }
             if(pending_state() != nullptr) {
-               throw TLS_Exception(Alert::UnexpectedMessage, "Can't interleave application and handshake data");
+               /*
+               What matters is which epoch the record belongs to, not which role we
+               are playing.
+
+               RFC 6347 4.2.4: "Implementations MUST either discard or buffer all
+               application data packets for the new epoch until they have received
+               the Finished message for that epoch." Data under the epoch this
+               handshake installed is not authenticated until its Finished, so it
+               must not reach the application; equally it is not an error, because
+               ordinary reordering produces it whenever a peer writes immediately
+               after activating.
+
+               Data under an epoch the established association owns stays valid
+               while a renegotiation is in flight, per 4.1.
+
+               Epoch zero is neither: application data there is plaintext, so it is
+               never legitimate and no association is at stake.
+               */
+               if(m_is_datagram && record.epoch() > 0) {
+                  const uint16_t active_epoch =
+                     m_epochs_before_latest_renegotiation ? m_epochs_before_latest_renegotiation->read_epoch : 0;
+
+                  if(!m_active_state.has_value() || record.epoch() > active_epoch) {
+                     continue;  // this handshake's epoch, still unauthenticated
+                  }
+               } else {
+                  throw TLS_Exception(Alert::UnexpectedMessage, "Can't interleave application and handshake data");
+               }
             }
             process_application_data(record.sequence(), m_record_buf);
          } else if(record.type() == Record_Type::Alert) {
--- a/src/lib/tls/tls12/tls_channel_impl_12.cpp
+++ b/src/lib/tls/tls12/tls_channel_impl_12.cpp
@@ -117,9 +117,52 @@ void Channel_Impl_12::reset_state() {
    m_active_state.reset();
    m_pending_state.reset();
    m_epochs_before_latest_renegotiation.reset();
+   m_resumption_handle.reset();
    m_readbuf.clear();
    m_write_cipher_states.clear();
    m_read_cipher_states.clear();
+}
+
+void Channel_Impl_12::note_resumption_handle(std::optional<Session_Handle> handle) {
+   m_resumption_handle = std::move(handle);
+}
+
+std::vector<Session_Handle> Channel_Impl_12::take_sessions_to_invalidate() {
+   // A ticket-backed session is not cached under the ServerHello session ID, so
+   // both handles have to be collected.
+   std::vector<Session_Handle> handles;
+
+   if(m_resumption_handle.has_value()) {
+      handles.push_back(m_resumption_handle.value());
+      m_resumption_handle.reset();
+   }
+
+   if(m_active_state.has_value()) {
+      const auto& sid = m_active_state->session_id();
+      if(!sid.empty()) {
+         handles.emplace_back(sid);
+      }
+   }
+
+   return handles;
+}
+
+void Channel_Impl_12::invalidate_sessions(const std::vector<Session_Handle>& handles) {
+   // RFC 5246 7.2.2: "Servers and clients MUST forget any session-identifiers,
+   // keys, and secrets associated with a failed connection.  Thus, any
+   // connection terminated with a fatal alert MUST NOT be resumed."
+   //
+   // Best effort, and deliberately last: remove() reaches application-supplied
+   // storage and can throw, and by then the keys are already gone. Letting a
+   // failed cache eviction abort the teardown would leave the channel usable
+   // after a security-fatal event, which is the worse of the two outcomes. A
+   // stateless ticket issuer has nothing to remove and cannot revoke what it
+   // already handed out.
+   for(const auto& handle : handles) {
+      try {
+         session_manager().remove(handle);
+      } catch(...) {}
+   }
 }
 
 void Channel_Impl_12::reset_active_association_state() {
@@ -480,6 +523,14 @@ size_t Channel_Impl_12::from_peer(std::span<const uint8_t> data) {
 
    try {
       while(input_size > 0) {
+         // A fatal alert destroys the cipher states, so nothing further can even
+         // be decrypted. Closure by close_notify is different: the responding
+         // close_notify still has to be read, so those records keep flowing
+         // through the loop and are filtered per record type below.
+         if(m_had_fatal_alert) {
+            return 0;
+         }
+
          size_t consumed = 0;
 
          auto get_epoch = [this](uint16_t epoch) { return read_cipher_state_epoch(epoch); };
@@ -559,6 +610,14 @@ size_t Channel_Impl_12::from_peer(std::span<const uint8_t> data) {
                   throw TLS_Exception(Alert::ProtocolVersion, "Received unexpected record version");
                }
             }
+         }
+
+         // RFC 5246 7.2.1: "Any data received after a closure alert is ignored."
+         // This is about a closure alert the peer sent us. A peer that keeps
+         // talking after *our* close_notify is a different case, kept as an
+         // error below; BoGo's Shutdown-Shim-ApplicationData requires it.
+         if(m_peer_closed_connection && record.type() != Record_Type::Alert) {
+            continue;
          }
 
          if(record.type() == Record_Type::Handshake || record.type() == Record_Type::ChangeCipherSpec) {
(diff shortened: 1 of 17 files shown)


src/lib/tls/tls12/tls_handshake_io.cpp
```diff
@@ -267,20 +378,74 @@ bool Datagram_Handshake_IO::reassemble_retransmitted_fragment(const uint8_t frag
...
+                                                 size_t ceiling,
+                                                 const uint8_t fragment[],
+                                                 size_t fragment_length,
+                                                 size_t fragment_offset,
+                                                 uint16_t epoch,
+                                                 Handshake_Type msg_type,
+                                                 size_t msg_length) {
+   if(m_pending_reassembly_bytes + fragment_length + Handshake_Reassembly::SEGMENT_OVERHEAD > ceiling) {
+      return false;
+   }
```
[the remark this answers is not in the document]
randombit: done

src/lib/tls/tls12/tls_handshake_io.h
```diff
@@ -117,24 +117,20 @@ class Stream_Handshake_IO final : public Handshake_IO {
 /**
 * Handshake IO for datagram-based handshakes
 */
-class Datagram_Handshake_IO final : public Handshake_IO {
+class BOTAN_TEST_API Datagram_Handshake_IO final : public Handshake_IO {
    public:
       typedef std::function<void(uint16_t, Record_Type, const std::vector<uint8_t>&)> writer_fn;
 
+      // Returns a monotonic ms timestamp; for tests, can be a virtual clock.
+      using clock_fn = std::function<uint64_t()>;
```
reneme: Consistency nit: my vote is for `using xxx = ...` but at least it should be locally consistent. :)

src/lib/tls/tls_callbacks.h
```diff
@@ -662,6 +662,16 @@ class BOTAN_PUBLIC_API(2, 0) Callbacks /* NOLINT(*-special-member-functions) */
...
 
+      /**
+       * Optional callback: provide the current monotonic timer in milliseconds.
+       *
+       * Used internally by DTLS for retransmit timer scheduling. The default
+       * implementation reads std::chrono::steady_clock. Override only to
+       * substitute a deterministic / virtual clock for testing; production
+       * code should not need to.
+       */
+      virtual uint64_t tls_current_timer_ms();
```
reneme: Especially with the planned switch away from the chrono types in Botan4: shouldn't this be merged? At least leave a TODO for it in the new `tls_current_timer_ms()`, because right now we unfortunately can't use the existing `tls_current_timestamp()` for this use case given that it returns a `system_clock` timestamp that might not be steady.

src/lib/tls/tls_policy.h
```diff
@@ -558,6 +558,19 @@ class BOTAN_PUBLIC_API(2, 0) Policy /* NOLINT(*-special-member-functions) */ {
...
+      * retransmitted on timeout before the handshake is abandoned. After this
+      * many timer-driven retransmissions without progress, timeout_check()
+      * throws to signal the handshake has failed. Return 0 to retransmit
+      * indefinitely (the historical behavior).
+      *
+      * RFC 6347 4.2.4.1 gives the retransmission timer schedule but states no
+      * condition for giving up, so this bound is local policy rather than a
+      * protocol requirement.
+      */
+      virtual size_t dtls_maximum_retransmissions() const;
```
reneme: Perhaps a matter of taste: But a `std::optional<size_t>` with `std::nullopt` indicating that indefinite retransmissions shall be allowed seems more intuitive than the special "zero" value. Also, I believe we used this idiom in several places already.

src/lib/tls/tls_ciphersuite.h
```diff
@@ -86,6 +86,12 @@ class BOTAN_PUBLIC_API(2, 0) Ciphersuite final {
        */
       bool aead_ciphersuite() const;
 
+      /**
+       * @return true if this suite uses AES-CCM with an 8 octet authentication
+       * tag, ie one of the _CCM_8 suites
+       */
+      bool ccm_8_ciphersuite() const;
```
reneme: In DTLS 1.3 there's another good reason to prohibit CCM(8) (See [RFC 9147 Section 4.2.3](https://www.rfc-editor.org/rfc/rfc9147.html#section-4.2.3)) altogether. The sequence number encryption assumes a record ciphertext that's at least 16 bytes long. Given the short authentication tag of CCM(8) that's not necessarily true, therefore, CCM(8) explicitly requires a padding to extend very short plaintexts.

Anyway: I suggest naming this predicate a little more generically:

```C++
      /**
       * @return true if this suite uses a symmetric encryption with less than
       *         16 octets of authentication tag (e.g. one of the _CCM_8 suites)
       */
      bool uses_short_authentication_tag() const;
```

... then we could hard-reject all cipher suites that would require the explained workaround in DTLS 1.3 and just refrain from implementing support for it altogether.

src/lib/tls/tls12/tls_handshake_io.cpp
```diff
@@ -690,19 +690,25 @@ std::vector<uint8_t> Datagram_Handshake_IO::send_message(uint16_t msg_seq,
...
 
-   if(no_fragment.size() + DTLS_HEADER_SIZE <= m_mtu) {
+   /**
+   * Largest possible overhead is for SHA-384 CBC ciphers, with 16 byte IV,
+   * 16+ for padding and 48 bytes for MAC. 128 is probably a strict
+   * over-estimate here. When CBC ciphers are removed this can be reduced
+   * since AEAD modes have no padding, at most 16 byte mac, and smaller
+   * per-record nonce.
+   */
+   const size_t ciphersuite_overhead = (epoch > 0) ? 128 : 0;
```
reneme: Side note: from a user's perspective it would be helpful to know the overhead that any negotiated DTLS association poses on the plaintext datagrams. In the best case, that would be an exact number based on the negotiated ciphersuite but a reasonably accurate upper bound would also help.

It would be great to expose this to the user somehow. Perhaps within the `Session_Summary` in `Callbacks::tls_session_established` or (even better) as a method on `TLS:Channel`.

Certainly that's future work, just pointing it out here.

src/lib/tls/tls12/tls_handshake_io.h
```diff
@@ -124,6 +124,10 @@ class BOTAN_TEST_API Datagram_Handshake_IO final : public Handshake_IO {
       // Returns a monotonic ms timestamp; for tests, can be a virtual clock.
       using clock_fn = std::function<uint64_t()>;
 
+      // msg_type(1) + length(3) + message_seq(2) + fragment_offset(3) +
+      // fragment_length(3)
+      static constexpr size_t DTLS_HANDSHAKE_HEADER_SIZE = 12;
```
reneme, line 135: Maybe promote this into `tls_magic.h`? The same is true for DTLS 1.3 anyway.

src/tests/test_tls_handshake_io.cpp
```diff
@@ -22,12 +26,43 @@ namespace {
...
+   r[6] = static_cast<uint8_t>((fragment_offset >> 16) & 0xFF);
+   r[7] = static_cast<uint8_t>((fragment_offset >> 8) & 0xFF);
+   r[8] = static_cast<uint8_t>(fragment_offset & 0xFF);
+   r[9] = static_cast<uint8_t>((fragment_length >> 16) & 0xFF);
+   r[10] = static_cast<uint8_t>((fragment_length >> 8) & 0xFF);
+   r[11] = static_cast<uint8_t>(fragment_length & 0xFF);
+   if(!fragment_bytes.empty()) {
+      std::memcpy(r.data() + 12, fragment_bytes.data(), fragment_bytes.size());
+   }
+   return r;
```
reneme, line 55: Untested code suggestion, but we can really do better than that:

```C++
const auto msg_len_bytes = Botan::store_be(msg_len);
const auto fragment_offset_bytes = Botan::store_be(fragment_offset)
const auto fragment_length_bytes = Botan::store_be(fragment_bytes.size())

return Botan::concat<std::vector<uint8_t>>(
   Botan::store_be(msg_type),
   std::span{msg_len_bytes}.last<3>(),
   Botan::store_be(message_seq),
   std::span{fragment_offset_bytes}.last<3>(),
   std::span{fragment_length_bytes}.last<3>(),
   fragment_bytes);
```

src/lib/tls/tls12/tls_handshake_io.h
```diff
@@ -212,6 +231,14 @@ class BOTAN_TEST_API Datagram_Handshake_IO final : public Handshake_IO {
 
       class Handshake_Reassembly final {
          public:
+            // Approximate real cost of one segment: red-black tree node holding
+            // the offset key and an inline std::vector, plus that vector's own
+            // heap allocation. Rounded up, since under-charging is what charging
+            // for segments at all is meant to prevent.
+            static constexpr size_t SEGMENT_OVERHEAD = 96;
```
reneme, line 269: I can't seem to comment on a specific commit, so I used this as an anchor. Frankly: the commit "Bound DTLS handshake reassembly memory" feels pretty over-engineered. Admittedly, I didn't go down all the gory details of it. But I think it deserves another pass.

src/lib/tls/tls12/tls_server_impl_12.cpp
```diff
@@ -259,6 +259,20 @@ std::map<std::string, std::vector<X509_Certificate>> get_server_certs(
...
+   try {
+      cookie_secret = creds.dtls_cookie_secret();
+   } catch(...) {}
+
+   if(cookie_secret.empty()) {
+      throw Invalid_State("DTLS server requires setting a non-empty cookie secret");
+   }
+
+   return SymmetricKey(cookie_secret);
+}
```
reneme, line 282: Perhaps use this opportunity and remove the `SymmetricKey` here (given that the type seems to be a bit of a relic)?

I.e.

```C++
secure_vector<uint8_t> load_dtls_cookie_secret(Credentials_Manager& creds) {
   auto cookie_secret = []() -> secure_vector<uint8_t> {
      try {
         return creds.dtls_cookie_secret();
      } catch(...) {
         return {};
      }
   }();

   if(cookie_secret.empty()) {
      throw Invalid_State("DTLS server requires setting a non-empty cookie secret");
   }

   return cookie_secret;
}
```

... the `Hello_Verify_Request` takes a `SymmetricKey` in its constructor and then passes it on into a `set_key()` which takes a `std::span`, so the rest of the change should be trivial as well.

src/lib/tls/tls12/tls_server_impl_12.cpp
```diff
@@ -270,6 +284,12 @@ Server_Impl_12::Server_Impl_12(const std::shared_ptr<Callbacks>& callbacks,
                                size_t io_buf_sz) :
       Channel_Impl_12(callbacks, session_manager, rng, policy, true, is_datagram, io_buf_sz), m_creds(creds) {
    BOTAN_ASSERT_NONNULL(m_creds);
+
+   // Fail at construction rather than on the first ClientHello. The secret is
+   // re-fetched per cookie verification so it can be rotated on a live listener.
+   if(is_datagram) {
+      load_dtls_cookie_secret(*m_creds);
+   }
```
reneme: At least DTLS 1.3 doesn't _require_ the Cookie exchange, I'm not sure about DTLS 1.2. The changes in "Require address-bound cookies for DTLS servers" seem to ossify Botan's DTLS implementation on enforcing it, though.

The cookie exchange introduces an extra round trip that might be worth avoiding for certain use cases. I think it would make sense to introduce a `TLS::Policy::dtls_require_initial_roundtrip_check()` (or similar) that enables the cookie mechanism by default (as suggested in RFC 9147) but ultimately lets the user decide on whether or not to use this feature.

As I understand the usage of cookies is currently hard-wired in DTLS 1.2, hence this is future work that we would take on in the context of #5000.

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -470,7 +470,16 @@ size_t Channel_Impl_12::from_peer(std::span<const uint8_t> data) {
...
-            process_handshake_ccs(m_record_buf, record.sequence(), record.type(), record.version(), epoch0_restart);
+            try {
+               process_handshake_ccs(m_record_buf, record.sequence(), record.type(), record.version(), epoch0_restart);
+            } catch(Exception&) {
+               // The restart ClientHello is unauthenticated until its cookie
+               // validates; if it fails, roll back the state.
+               if(!epoch0_restart) {
+                  throw;
+               }
+               clear_pending_handshake_state();
```
reneme, line 639: Strictly speaking, even after the cookie validates, there's no authentication. We know that the peer is reachable on the source address it claims but that's no authentication.

Also: if we're not in the DTLS case here, that `catch()` might actually suppress errors that should have resulted in  an abort of the TLS (without the D) handshake. Or can we rule that out somehow?

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -503,11 +512,65 @@ size_t Channel_Impl_12::from_peer(std::span<const uint8_t> data) {
...
+   Discard the state instead, so a new cookie exchange begins. Restricted to a
+   pending state that has consumed a message but holds no accepted ClientHello,
+   ie one that has answered with a HelloVerifyRequest and is still waiting for
+   the cookie to come back. A handshake that got past the cookie check keeps its
+   state, so this cannot be used to reset one in progress.
+   */
+   if(dtls_io->in_message_seq() > 0 && m_pending_state->client_hello() == nullptr) {
+      clear_pending_handshake_state();
+   }
+}
```
reneme, line 750: This is trying to detect a ClientHello in an incoming handshake protocol fragment and performs a number of sanity checks on it. To me, that feels like major band-aid and out-of-place here, to be honest.

It's duplicating parsing logic that is implemented in `Datagram_Handshake_IO::add_record()`. Also, if we handle ClientHellos, why do we do so in the generic Channel implementation? Shouldn't it at least be done in the Server impl only?

reneme (commented):
So far I went through the first third of commits, concretely until "[Ignore unauthenticated epoch-zero input in active DTLS associations](https://github.com/randombit/botan/pull/5783/changes/dd1ab5f1bc205b41181cc741f4ca1349d8a6a583)".

This seems to increase the complexity of the implementation quite significantly, I fear. I'm sure it's a net win, though at this point I can't assess how much this costs in terms of maintainability.

Below are a few things that I found worth pointing out so far, but that's no thorough review by a long shot.

randombit:
> This seems to increase the complexity of the implementation quite significantly, I fear. I'm sure it's a net win, though at this point I can't assess how much this costs in terms of maintainability.

I can hardly disagree on this point. But I think a major factor is DTLS just *is* complicated and difficult to implement correctly, much more so than TLS and more than I had appreciated, and the implementation was failing to correctly handle many situations that can arise in the datagram setting.

BTW some concrete lessons learned here that imo can be applied for DTLS 1.3:

* DTLS should be in its own module that can be turned on/off
* DTLS should require using specific interfaces; the combined TLS 1.2+DTLS 1.2 Channel was a bad idea.
* DTLS server listening/cookie exchange step should be stateless - a great part of the complexity here is trying to minimize the damage from the fact that anyone can send a packet from anywhere, with a forged source address, and cause us to allocate memory.

As a followup to this PR (for 3.14) I'm going to split the DTLS 1.2 logic out to `dtls12` submodule

reneme:
> BTW some concrete lessons learned here that imo can be applied for DTLS 1.3:
> * DTLS should be in its own module that can be turned on/off

Yeah, we've set it up like that already. The major differences and complexity (mostly record handling, handshake message parsing and management, and cipher state handling) are implemented as polymorphic classes that can live in its own module. Some state machine shenanigans are handled with conditionals, but we try to keep this minimal.

> * DTLS should require using specific interfaces; the combined TLS 1.2+DTLS 1.2 Channel was a bad idea.

I'm not sure about that. The only major difference we found so far is that `TLS::Channel::from_peer()` and `TLS::Callbacks::tls_emit_data()` must handle data at datagram granularity: i.e. the user always drops a single full datagram into the implementation and always get a single full datagram out of it. Otherwise the API surface seems to work just fine (apart from a number of specific callbacks and policy methods). What am I missing?

> * DTLS server listening/cookie exchange step should be stateless

Fair, we yet have to look into that for DTLS 1.3

randombit:
> Otherwise the API surface seems to work just fine (apart from a number of specific callbacks and policy methods). What am I missing?

I guess honestly mostly callbacks and policies, plus that stateless listening isn't really tractable. The situation is not horrible but there has been to me a consistent trend where attempts to unify two fundamentally different things usually results in a not ideal interface for either. The worst was probably #475 - cipher modes and compression algorithms used to share a common base class and forcing them into a common API caused a lot of complications which we still suffer with today. Other cases are `Buffered_Computation`, the entire `Filter` hierarchy, and the fact that cipher mode encryption and decryption have the same interface, even though splitting them would allow significantly simpler interfaces for both cases.

So now I'm wary of that situation recurring with new code. For example a few years back I started working OPAQUE and CPace support and of course the first thing to do was define a generic framework for PAKEs which could handle those plus SRP. I ended up abandoning the entire thing, and you can see for SPAKE2+ I just went with directly representing the SPAKE2+ RFC types. If someone needs to make their application generic over multiple PAKEs (and why would anyone, really?) they can define an abstraction on top and live with it themselves.

src/lib/tls/tls_policy.h
randombit: Yeah that's fine too.

But, should 0 mean "never retransmit"? Or just an error?

I'd be inclined for that to mean never retransmit, since that setting is actually useful if you are running DTLS over a reliable datagram channel.

src/lib/tls/tls12/tls_handshake_io.cpp
randombit: Cosigned. Opened #5786

src/lib/tls/tls12/tls_handshake_io.h
randombit: This is the end result of several iterations (over many months - this branch originally started in April) of trying one fix, some model or another pointing out that the fix didn't completely address the issue, repeat. I'll look at it again though.

src/lib/tls/tls12/tls_handshake_io.cpp
```diff
@@ -445,53 +526,89 @@ void Datagram_Handshake_IO::add_record(const uint8_t record[],
...
+   // Charge the reassembly budget by bytes actually committed, not by
+   // the claimed msg_length. Otherwise an attacker can lock the full
+   // budget by sending small (e.g. fragment_length=1) fragments with
+   // a large claimed msg_length, displacing legitimate handshake
+   // messages.
+   // Reserve headroom for the message actually being waited on. Otherwise
+   // fragments for the fifteen slots beyond it can consume the whole
+   // budget, after which every fragment of the expected message is
+   // silently dropped and the handshake cannot proceed.
+   const size_t ceiling = (message_seq == m_in_message_seq) ? m_max_pending_reassembly : m_max_pending_reassembly / 2;
```
atreiber94, line 733: Can't an attacker also lock the full budget in the case `message_seq < m_in_message_seq`? (The other `charged_add_fragment` calls use the full `m_max_pending_reassembly`.)

src/lib/tls/tls12/tls_handshake_io.h
atreiber94: For context: We're actually currently implementing this for DTLS1.3 and it turns out to be more complicated than anticipated if we truly want to avoid memory-exhaustion attacks. It'd be great to get this logic shared between the implementations, since the reassembly is quite similar between the versions.

src/lib/tls/tls12/tls_server_impl_12.cpp
randombit: As far as specification goes, DTLS 1.2 doesn't require it either. And there are certainly environments where it isn't necessary. The issue I was trying to address here is that it's pretty easy for an application to accidentally not set a secret which would completely disable cookie exchanges, which if the server is internet facing will lead to a bad time.

A policy toggle defaulting to true (in which case the secret must be set) but if false turns off the exchange is fine.

One minor question on policy: if `dtls_require_initial_roundtrip_check` is false but a DTLS cookie secret exists, should the HelloVerifyRequest be sent? I'd lean to yes.

src/lib/tls/tls12/tls_channel_impl_12.cpp
randombit: This is certainly supposed to be DTLS only but I'll double check this.

(Another case for splitting out the DTLS logic to its own module)

src/lib/tls/tls12/tls_handshake_io.cpp
```diff
@@ -142,6 +140,20 @@ std::vector<uint8_t> Stream_Handshake_IO::send(const Handshake_Message& msg) {
    return buf;
 }
 
+namespace {
+
+size_t max_pending_reassembly(size_t policy_hs_max) {
+   constexpr size_t overall_cap = 1024 * 1024;  // arbitrary
```
atreiber94, line 158: Not sure if some specific users may have issues with this overall_cap (looking at, e.g., Classic McEliece sizes). It even overrides the `policy_hs_max == 0` cap, which originally meant to say that any size is acceptable. This would then lead to silent, unexpected drops of messages with a size users deemed admissable.

randombit: Classic McEliece over DTLS would be something. But yeah reasonable to let the application override this if they want rather than imposing a fixed cap.

src/lib/tls/tls12/tls_handshake_io.h
randombit: > It'd be great to get this logic shared between the implementations, since the reassembly is quite similar between the versions.

Sounds good to me. Probably this ends up as `dtls12` and `dtls13`  modules plus a `dtls_utils` internal module that handles common elements.

src/lib/tls/tls_callbacks.h
randombit: Re-reading your comment I am not sure I understand what you're asking for. Are you suggesting a single entry point that returns either monotonic or system time depending eg on some parameter? I split out this commit into #5788, not all of this PR splits out so easily but I think #5787 and #5788 will help quite a bit [and with those merged more can be split out, I don't want to stack too deeply here...] but maybe lets continue the discussion there.

src/lib/tls/tls12/tls_handshake_io.h
randombit: Fixed (changed the upper typedef to `using`) in #5788

src/lib/tls/tls_callbacks.h
reneme: I wanted to suggest that in Botan4 there should be only one callback:

`uint64_t tls_current_time_ms()` (or similar)

that returns a UTC timestamp which can be used for both "wall clock time" as well as "steady time measurements" for timeouts and such.

Probably the hard guarantees that `std::chrono::steady_clock` provides are stricter than that. But likely we can live with the slim chance of a delayed timeout over a simpler API.

Alternatively, if we really want to keep both, I'd suggest to stick to the `<chrono>` types and live with the heavy include for API clarity here.

src/lib/tls/tls_policy.h
reneme: Yeah, 0 should mean 0 retransmissions, IMO.

src/lib/tls/tls12/tls_server_impl_12.cpp
reneme: > One minor question on policy: if dtls_require_initial_roundtrip_check is false but a DTLS cookie secret exists, should the HelloVerifyRequest be sent? I'd lean to yes.

It would be somewhat surprising, I guess. But it's the safer default. Perhaps mention the behavior in the docstrings.

src/lib/tls/tls_callbacks.h
randombit: Feel like in the face of NTP time jumps combining the two clocks is inevitable to end in tragedy for somebody.

src/lib/tls/tls_ciphersuite.h
randombit: Renamed. Extracted this change as #5791

src/lib/tls/tls12/tls_server_impl_12.cpp
randombit: Extracted this part into #5792. I skipped on the policy returns false but secret exists case I mentioned earlier - if the policy returns true the server must send a cookie or will die trying, if the policy is false then the cookie step is skipped and neither the secret nor the peer identity are looked at.

randombit:
So I've split out some of the simpler to review pieces into individual PRs. After those merge I'll rebase this PR, then probably again try to extract out another 3-5 smaller PRs. There is is a hard/tangled core around the epoch0 handling that probably will remain.

randombit:
Rebased to remove changes made on master through distinct PRs. I'll split out another round shortly

src/lib/tls/tls12/tls_handshake_io.cpp
randombit: I kept this but increased it to 16 Mb since that covers any conceivable usage (imo) and prevents a truly unbounded runaway allocation setup which is otherwise possible.

randombit:
Most of this has landed, what's left is still sufficiently half-baked that I'm deferring to 3.14

PR #5788, opened 2026-08-03
Make DTLS retransmission timing deterministic

randombit:
This is commit 3 from #5783 ~stacked on the test refactor commits in #5787~

### DIFF
--- a/src/lib/tls/tls12/tls_handshake_io.h
+++ b/src/lib/tls/tls12/tls_handshake_io.h
@@ -117,24 +117,20 @@ class Stream_Handshake_IO final : public Handshake_IO {
 /**
 * Handshake IO for datagram-based handshakes
 */
-class Datagram_Handshake_IO final : public Handshake_IO {
+class BOTAN_TEST_API Datagram_Handshake_IO final : public Handshake_IO {
    public:
-      typedef std::function<void(uint16_t, Record_Type, const std::vector<uint8_t>&)> writer_fn;
+      using writer_fn = std::function<void(uint16_t, Record_Type, const std::vector<uint8_t>&)>;
+
+      // Lambda pointing to clock function (normally TLS::Callbacks::tls_current_monotonic_clock_ms)
+      using steady_clock_fn = std::function<uint64_t()>;
 
       Datagram_Handshake_IO(writer_fn writer,
+                            steady_clock_fn clock_ms,
                             class Connection_Sequence_Numbers& seq,
                             uint16_t mtu,
                             uint64_t initial_timeout_ms,
                             uint64_t max_timeout_ms,
-                            size_t max_handshake_msg_size) :
-            m_seqs(seq),
-            m_flights(1),
-            m_flight_ccs(1),
-            m_initial_timeout(initial_timeout_ms),
-            m_max_timeout(max_timeout_ms),
-            m_send_hs(std::move(writer)),
-            m_mtu(mtu),
-            m_max_handshake_msg_size(max_handshake_msg_size) {}
+                            size_t max_handshake_msg_size);
 
       Protocol_Version initial_record_version() const override;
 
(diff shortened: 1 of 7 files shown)


src/lib/tls/tls_callbacks.h
```diff
@@ -662,6 +662,16 @@ class BOTAN_PUBLIC_API(2, 0) Callbacks /* NOLINT(*-special-member-functions) */
...
 
+      /**
+       * Optional callback: provide the current monotonic timer in milliseconds.
+       *
+       * Used internally by DTLS for retransmit timer scheduling. The default
+       * implementation reads std::chrono::steady_clock. Override only to
+       * substitute a deterministic / virtual clock for testing; production
+       * code should not need to.
+       */
+      virtual uint64_t tls_current_monotonic_clock_ms();
```
reneme, line 673: Thanks, the name change is making the reason for the two time-related callbacks reasonably clear, I would say. Regarding the discussion here: https://github.com/randombit/botan/pull/5783#discussion_r3703638318

src/tests/test_dtls12.cpp
```diff
@@ -81,6 +81,21 @@ class DTLS_Test_Callbacks final : public Botan::TLS::Callbacks {
...
       }
 
+      // Returns the normal clock unless advance_clock_ms has been called, in
+      // which case the timer becomes controlled entirely by the test
+      uint64_t tls_current_monotonic_clock_ms() override {
+         if(m_virtual_now_ms.has_value()) {
+            return m_virtual_now_ms.value();
+         } else {
+            return Botan::TLS::Callbacks::tls_current_monotonic_clock_ms();
+         }
```
reneme, line 92: Nit:

```suggestion
         return m_virtual_now_ms.value_or(Botan::TLS::Callbacks::tls_current_monotonic_clock_ms());
```

... that's slightly less efficient, because it will always read the clock, even if the optional is set. But it's a test implementation...

src/tests/test_tls_handshake_io.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+               fix.advance_clock_ms(initial_timeout);
+               result.test_is_true("retransmit fires once the timeout elapses", fix.io().timeout_check());
+               result.test_sz_eq("the flight was resent", fix.handshake_records_sent(), size_t(2));
+            }),
+
+   };
+}
+
+BOTAN_REGISTER_TEST_FN("tls", "dtls_handshake_io", dtls_handshake_io_tests);
```
reneme: Perhaps point out in the test logs that this relates to TLS 1.2 only:


```suggestion
BOTAN_REGISTER_TEST_FN("tls", "dtls12_handshake_io", dtls_handshake_io_tests);
```

... in TLS 1.3 the "handshake_layer" tests are related.

src/lib/tls/tls12/tls_handshake_io.h
```diff
@@ -282,13 +278,18 @@ class Datagram_Handshake_IO final : public Handshake_IO {
...
+      // cannot be spelled as a reserved timestamp.
+      std::optional<uint64_t> m_last_write;
+
       uint64_t m_next_timeout = 0;
 
       uint16_t m_in_message_seq = 0;
       uint16_t m_out_message_seq = 0;
 
       writer_fn m_send_hs;
+      clock_fn m_clock_ms;
```
reneme: Nit: make it clear that this is a steady clock (as was done before in the local free-standing function):


```suggestion
      clock_fn m_steady_clock_ms;
```

reneme (approved):
Just nits, LGTM.

src/tests/test_tls_handshake_io.cpp
randombit: fixed

src/lib/tls/tls12/tls_handshake_io.h
randombit: done (also renamed the constructor argument name)

src/tests/test_dtls12.cpp
randombit: I feel like that read a little more awkwardly tbh. I left a note here to use `std::optional::or_else` when that becomes an option.

reneme (approved):

PR #5791, opened 2026-08-03
Refuse unsafe ciphersuites in DTLS

randombit:
Our Lucky13 countermeasure is decent but not completely constant time, and DTLS offers a repeated oracle. So reject CBC modes in DTLS.

Similarly, CCM-8 has only a 64-bit MAC so the relevant standards suggest not supporting CCM-8 in DTLS.

Extracted from #5783

### DIFF
--- a/src/tests/unit_tls.cpp
+++ b/src/tests/unit_tls.cpp
@@ -924,6 +924,27 @@ class TLS_Unit_Tests final : public Test {
          };
       }
 
+      // Policy::ciphersuite_list refuses both CBC+HMAC and _CCM_8 suites for
+      // DTLS, so those policies leave no DTLS ciphersuite to negotiate.
+      static std::vector<Botan::TLS::Protocol_Version> versions_for_policy(
+         std::vector<Test::Result>& results,
+         std::vector<Botan::TLS::Protocol_Version> versions,
+         const std::string& cipher_policy,
+         const std::string& mac_policy) {
+         const bool dtls_has_no_suite =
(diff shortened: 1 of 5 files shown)


src/tests/unit_tls.cpp
```diff
@@ -924,6 +924,21 @@ class TLS_Unit_Tests final : public Test {
...
+         const std::string& cipher_policy,
+         const std::string& mac_policy) {
+         const bool dtls_has_no_suite =
+            (mac_policy != "AEAD") || (cipher_policy != "NULL") || cipher_policy.find("CCM(8)") != std::string::npos;
+
+         if(dtls_has_no_suite) {
+            std::erase_if(versions, [](const auto& version) { return version.is_datagram_protocol(); });
+         }
+         return versions;
+      }
```
reneme, line 946: If this ever accidentally filters all versions (because of a future misconfiguration or whatever), tests would all be green because `test_with_policy` does something along those lines:

```C++
         try {
            for(const auto& version : versions) {
                // ... ... ...
            }
         } catch(std::exception& e) {
            results.push_back(Test::Result::Failure(test_descr, e.what()));
         }
```

I think it would be worthwhile to pass the `results` in here and write some logs about what versions were filtered and fail the test if no versions remain.

reneme (approved):
One defensive suggestion in the test implementation, otherwise LGTM.

src/tests/unit_tls.cpp
randombit: done

reneme (approved):

PR #5792, opened 2026-08-03
Require address-bound cookies for DTLS servers

randombit:
Previously an application could skip implementing tls_peer_network_identity, which would have the effect of making the cookie exchange trivially bypassed since the cookies would not be bound to the source address. Now, if DTLS cookies are being used, tls_peer_network_identity must return a non-empty string.

Additionally, add a new Policy toggle dtls_server_require_cookie_exchange. If true (the default) then the cookie exchange is required. Previously an application indicated a desire to use the cookie exchange by returning a DTLS cookie secret. But a misconfigured application could easily just not configure such a secret, in which case the cookie exchange is silently disabled. The new policy switch makes such a misconfiguration a hard error - an application must now do something *explicit* (namely overriding dtls_server_require_cookie_exchange to return false) in order to disable the cookie exchange.

Split out from #5783

### DIFF
--- a/src/bogo_shim/bogo_shim.cpp
+++ b/src/bogo_shim/bogo_shim.cpp
@@ -2149,6 +2149,8 @@ class Shim_Callbacks final : public Botan::TLS::Callbacks {
          return g_now + m_clock_skew;
       }
 
+      std::string tls_peer_network_identity() override { return "bogo-shim-peer"; }
+
       void tls_inspect_handshake_msg(const Botan::TLS::Handshake_Message& msg) override {
          if(msg.type() == Botan::TLS::Handshake_Type::HelloRetryRequest) {
             m_hello_retry_request = true;
--- a/src/lib/tls/tls12/tls_server_impl_12.cpp
+++ b/src/lib/tls/tls12/tls_server_impl_12.cpp
@@ -259,6 +259,28 @@ std::map<std::string, std::vector<X509_Certificate>> get_server_certs(
    return cert_chains;
 }
 
+secure_vector<uint8_t> load_dtls_cookie_secret(Credentials_Manager& creds) {
+   auto cookie_secret = [&]() -> secure_vector<uint8_t> {
+      try {
+         return creds.psk("tls-server", "dtls-cookie-secret", "").bits_of();
+      } catch(...) {
+         return {};
+      }
+   }();
+
+   if(cookie_secret.empty()) {
+      // TODO(Botan4): Simplify this error message that was meant to ease the
+      //               burden on users running into a deliberate semver violation.
+      throw Invalid_State(
+         "Since Botan 3.13 DTLS server requires setting a non-empty cookie secret. "
+         "Either override Credentials_Manager::dtls_cookie_secret() or disable the "
+         "cookie exchange using TLS::Policy::dtls_server_require_cookie_exchange(), "
+         "if you understand the security implications of doing so.");
+   }
+
+   return cookie_secret;
+}
+
 }  // namespace
 
 Server_Impl_12::Server_Impl_12(const std::shared_ptr<Callbacks>& callbacks,
@@ -270,6 +292,12 @@ Server_Impl_12::Server_Impl_12(const std::shared_ptr<Callbacks>& callbacks,
                                size_t io_buf_sz) :
       Channel_Impl_12(callbacks, session_manager, rng, policy, true, is_datagram, io_buf_sz), m_creds(creds) {
    BOTAN_ASSERT_NONNULL(m_creds);
+
+   // Try to load the cookie secret on initialization, rather than waiting to fail
+   // until the first client connects.
+   if(is_datagram && policy->dtls_server_require_cookie_exchange()) {
+      load_dtls_cookie_secret(*m_creds);
+   }
 }
 
 Server_Impl_12::Server_Impl_12(const Channel_Impl::Downgrade_Information& downgrade_info) :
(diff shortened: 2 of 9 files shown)


src/bogo_shim/bogo_shim.cpp
```diff
@@ -2149,6 +2149,8 @@ class Shim_Callbacks final : public Botan::TLS::Callbacks {
          return g_now + m_clock_skew;
       }
 
+      std::string tls_peer_network_identity() override { return "bogo-shim-peer"; }
```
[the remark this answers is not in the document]
randombit: Not relevant

reneme:
> The new policy switch makes such a misconfiguration a hard error - an application must now do something explicit [...] in order to disable the cookie exchange.

For the record: Given the default of the new policy toggle, this is a semver violation. Users that (accidentally) didn't configure a cookie secret will get a runtime error after updating.

I believe it's justified here, just pointing it out.

src/lib/tls/tls12/tls_server_impl_12.cpp
```diff
@@ -259,6 +259,22 @@ std::map<std::string, std::vector<X509_Certificate>> get_server_certs(
...
+   auto cookie_secret = [&]() -> secure_vector<uint8_t> {
+      try {
+         return creds.psk("tls-server", "dtls-cookie-secret", "").bits_of();
+      } catch(...) {
+         return {};
+      }
+   }();
+
+   if(cookie_secret.empty()) {
+      throw Invalid_State("DTLS server requires setting a non-empty cookie secret");
```
reneme: Related to my [previous comment about violating semver](https://github.com/randombit/botan/pull/5792#issuecomment-5174770619): maybe its worthwhile to make this error message more verbose to help users that run into this:


```suggestion
      // TODO(Botan4): Simplify this error message that was meant to ease the
      //               burden on users running into a deliberate semver violation.
      throw Invalid_State("Since Botan 3.13 DTLS server requires setting a non-empty cookie secret. "
                          "Either override Credentials_Manager::dtls_cookie_secret() or disable the "
                          "cookie exchange using TLS::Policy::dtls_server_require_cookie_exchange(), "
                          "if you understand the security implications of doing so.");
```

src/lib/tls/tls12/tls_server_impl_12.cpp
```diff
@@ -420,31 +442,31 @@ void Server_Impl_12::process_client_hello_msg(Server_Handshake_State& pending_st
...
-            pending_state.set_expected_next(Handshake_Type::ClientHello);
-            return;
+      const std::string client_identity = callbacks().tls_peer_network_identity();
+      if(client_identity.empty()) {
+         // RFC 9147 Section 11: the cookie MUST depend on the client's address
+         //
+         // Without an application-supplied identity the cookie is reusable from
+         // any source address, defeating the whole point of the cookie exchange.
+         throw Invalid_State("DTLS server requires tls_peer_network_identity() return a non-empty value");
+      }
```
reneme, line 465: Similarly, that's a reasonable semver violation IMO, but we should make it as easy to troubleshoot for the user as we can:

```suggestion
      const std::string client_identity = callbacks().tls_peer_network_identity();
      if(client_identity.empty()) {
         // RFC 9147 Section 11: the cookie MUST depend on the client's address
         //
         // Without an application-supplied identity the cookie is reusable from
         // any source address, defeating the whole point of the cookie exchange.
         //
         // TODO(Botan4): Remove the version hint about the breaking change.
         throw Invalid_State("Since Botan 3.13 DTLS server requires tls_peer_network_identity() return a non-empty value");
      }
```

... I think, explicitly stating the version where the violation happened should make it easier for users to figure out what's going on when they suddenly see errors that weren't there before.

src/lib/tls/tls_policy.cpp
```diff
@@ -471,6 +471,10 @@ bool Policy::allow_dtls_epoch0_restart() const {
    return false;
 }
 
+bool Policy::dtls_server_require_cookie_exchange() const {
+   return true;
```
reneme, line 487: I think it doesn't hurt to provide some reference for users to read up on the implications of that:

```suggestion
   // RFC 9147 11. "Security Considerations"
   //    The primary additional security consideration raised by DTLS is that of
   //    denial of service by excessive resource consumption. DTLS includes a
   //    cookie exchange designed to protect against denial of service. [...]
   //    DTLS servers that do not use the cookie exchange may be used as attack
   //    amplifiers even if they themselves are not experiencing DoS. Therefore,
   //    DTLS servers SHOULD use the cookie exchange unless there is good reason
   //    to believe that amplification is not a threat in their environment.
   return true;
```

src/lib/tls/tls_policy.h
```diff
@@ -534,6 +534,22 @@ class BOTAN_PUBLIC_API(2, 0) Policy /* NOLINT(*-special-member-functions) */ {
       */
       virtual bool allow_dtls_epoch0_restart() const;
 
+      /**
+      * DTLS defines a cookie exchange protocol where an inbound client hello
+      * must first perform a cookie exchange. The server handles this by hashing
+      * the original client hello with a secret key plus the peer's address.
+      * This implies one extra round trip during the handshake.
```
reneme: Let's also answer the "Why?" here to educate the decision of the user.

```suggestion
      * DTLS defines a cookie exchange protocol where an inbound client hello
      * must first perform a cookie exchange. This is done to ensure return-
      * routability between server and client and to prevent certain DoS
      * amplification attacks.
      *
      * The server handles this by hashing the original client hello with a
      * secret key plus the peer's address. This implies one extra round trip
      * during the handshake.
```

reneme (approved):
Just some suggestions regarding user-centric documentation, educating them on what the cookie exchange is done for and why we decided to change the behavior in a minor version. LGTM.

src/lib/tls/tls12/tls_server_impl_12.cpp
randombit: done

src/lib/tls/tls12/tls_server_impl_12.cpp
randombit: done

src/lib/tls/tls_policy.h
randombit: done (also generally expanded upon and rewrote this block, some of the original phrasing was bad)

reneme (approved):

src/lib/tls/tls_policy.cpp
randombit: done

PR #5793, opened 2026-08-03
Bound DTLS handshake retry attempts and properly abandon timed out states

randombit:
Extracted from #5783, currently stacked on #5788


randombit:
Rebased after #5788 no changes otherwise

src/bogo_shim/bogo_shim.cpp
```diff
@@ -2272,18 +2293,39 @@ int main(int /*argc*/, char* argv[]) {
...
+                     // pending timer to the application; consume the bytes and
+                     // skip. AdvanceClock failures (unexpected packets / no
+                     // packets) still catch incorrect timing.
+                     uint8_t bytes[8];
+                     socket.read_exactly(bytes, sizeof(bytes));
+                  } else if(opcode == 'M') {
+                     // SetPeerMTU: Botan's MTU is fixed at handshake
+                     // construction; consume and ignore.
+                     uint8_t bytes[4];
+                     socket.read_exactly(bytes, sizeof(bytes));
```
reneme, line 2320: Perhaps at least a `shim_log("New MTU: " + std::to_string(Botan::load_be(bytes)) + ", currently ignored!")` for future-us?

src/bogo_shim/bogo_shim.cpp
```diff
@@ -2272,18 +2293,39 @@ int main(int /*argc*/, char* argv[]) {
...
+                  } else if(opcode == 'U') {
+                     // SetPeerTimeout: not yet plumbed through Botan's IO
+                     // (no API to override the running m_next_timeout).
+                     // Consume and ignore; SetTimeout-* tests will fail.
+                     uint8_t bytes[4];
+                     socket.read_exactly(bytes, sizeof(bytes));
```
reneme, line 2327: Similarly: maybe at least emit a `shim_log()` to hint that we ignored this?

src/lib/tls/tls_policy.cpp
```diff
@@ -496,6 +496,12 @@ size_t Policy::dtls_maximum_timeout() const {
    return 60 * 1000;
 }
 
+std::optional<size_t> Policy::dtls_maximum_retransmissions() const {
+   // Matches BoringSSL's DTLS1_MAX_TIMEOUTS. With the default 1s initial / 60s max
+   // schedule this gives up after roughly 7 minutes (1+2+4+8+16+32+60*6 s).
+   return 12;
+}
```
reneme, line 523: Strictly speaking that is a semver break, no? Then again, 7 minutes is a long time for any existing application to care, but nevertheless: this is now throwing an exception that wasn't there before. So for sure it's worth pointing out. Or, alternatively, just return `std::nullopt` to keep the historical behavior.

src/lib/tls/tls12/tls_channel_impl_12.h
```diff
@@ -237,6 +237,22 @@ class Channel_Impl_12 : public Channel_Impl {
...
 
+      // Epochs in force when the pending handshake began. Whether either has
+      // moved decides if an abandoned handshake can be discarded or has to
+      // take the association with it.
+      struct Epochs_Before_Pending final {
+            uint16_t read_epoch;
+            uint16_t write_epoch;
+      };
+
+      std::optional<Epochs_Before_Pending> m_epochs_before_pending;
```
reneme: Nit: members are typically declared at the very end of a class. Here, a few method definitions follow.

src/lib/tls/tls12/tls_handshake_io.h
```diff
@@ -278,6 +279,13 @@ class BOTAN_TEST_API Datagram_Handshake_IO final : public Handshake_IO {
       uint64_t m_initial_timeout = 0;
       uint64_t m_max_timeout = 0;
 
+      // Maximum timer-driven retransmissions of the current flight before the
+      // handshake is abandoned (nullopt unlimited). m_retransmit_count tracks the
+      // number fired for the in-flight wait; it resets to 0 whenever a new
+      // flight is sent (forward progress) and is incremented on each timeout.
+      std::optional<size_t> m_max_retransmissions = 0;
```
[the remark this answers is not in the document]
reneme: I tend to agree with the robot on this. Probably would just leave out the explicit initialization here, as it is the default behavior anyway.

src/lib/tls/tls12/tls_handshake_io.cpp
```diff
@@ -229,6 +231,19 @@ bool Datagram_Handshake_IO::timeout_check() {
...
    }
 
+   // The retransmit timer has expired. Count this attempt and, once the
+   // configured cap is reached, abandon the handshake rather than retransmit
+   // forever. RFC 6347 4.2.4.1 gives the backoff schedule but states no
+   // condition for giving up, so the cap is local policy, not a requirement.
+   // No alert is sent - the peer is by definition unresponsive.
+   m_retransmit_count += 1;
+   if(m_max_retransmissions.has_value() && m_retransmit_count > m_max_retransmissions.value()) {
+      throw TLS_Exception(Alert::None, "DTLS handshake timed out: maximum retransmissions exceeded");
```
reneme, line 241: The fact that `TLS::Channel::timeout_check()` now regularly throws after a bounded number of attempts should be documented in the base class API.

src/lib/tls/tls12/tls_channel_impl_12.h
reneme: Super-Nit: Maybe `Epochs_Before_Latest_Renegotiation`? The term "Pending" feels fairly generic to me. Also in the name of the member variable.

src/lib/tls/tls12/tls_handshake_io.h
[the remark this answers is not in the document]
randombit: Unintentional, but actually this is better - an uninitialized constructor variable causes retransmits to not occur at all, and so fail in a noticeable way. In contrast nullopt would cause an uninitialized variable to silently enable infinite retransmits.

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -188,9 +194,54 @@ Handshake_State& Channel_Impl_12::create_handshake_state(Protocol_Version versio
...
+      clear_pending_handshake_state();
+   } else {
+      // Either there is no established association to fall back to, or a
+      // ChangeCipherSpec has already moved an epoch and there is no rollback
+      // that would leave keys, identity and sequence numbers describing the
+      // same handshake. Close.
+      m_has_been_closed = true;
+      reset_state();
+   }
+}
```
reneme, line 235: The implemented strategy makes sense, and I don't object integrating it like that. However, from the perspective of an application that uses `TLS::Channel::timeout_check()`, the behavior might be surprising and I don't see the benefit of distinguishing the two cases just yet.

Let's say we have an existing association on which we call `renegotiate()` followed by the occasional `timeout_check()`. Eventually we catch the timeout exception from `timeout_check()`. If the peer became unresponsive before the renegotiation started, the association remains intact and the application could try again. Otherwise it breaks entirely.

In both cases we essentially know that the peer is not reachable, so why do we even bother trying to save the association?

I would expect that a renegotiation handshake failing due to timeout would behave the same as if it failed due to some protocol error. What am I missing?

reneme (approved):

reneme:
Side note: Our current working hypothesis for DTLS 1.3 (#5000) is that we won't support renegotiation at all. Obviously, because TLS 1.3 doesn't support renegotiation anyway. That should remove the entire complexity with handling pending handshake state in existing associations.

Nevertheless, a server may receive a new ClientHello from a peer that it holds an existing association with (See RFC [9147 5.11](https://www.rfc-editor.org/rfc/rfc9147.html#name-establishing-new-associatio)), e.g. after a reboot of the client. Being aware of the complexity this adds, I would like to delegate that handling to the application. Frankly, I didn't think of a suitable API to let the application know that a new server association should be instantiated, yet. Perhaps just throw a special exception from `TLS::Channel::received_data())`?

@randombit What do you think about that?

src/lib/tls/tls_policy.cpp
randombit: I mean strictly speaking anything is a SemVer break depending on ones perspective. My guess is the median user of DTLS, upon being told that Botan retried failing handshakes literally forever without ever giving up, would be surprised by this. Certainly will be worth calling out in the release notes. But between this current push and the previous retransmit work, I think 3.13 will be a worthwhile upgrade for any DTLS user.

randombit:
> Nevertheless, a server may receive a new ClientHello from a peer that it holds an existing association with (See RFC [9147 5.11](https://www.rfc-editor.org/rfc/rfc9147.html#name-establishing-new-associatio)), e.g. after a reboot of the client. Being aware of the complexity this adds, I would like to delegate that handling to the application. Frankly, I didn't think of a suitable API to let the application know that a new server association should be instantiated, yet. Perhaps just throw a special exception from TLS::Channel::received_data())?

I would approve of not supporting this in DTLS 1.3. The epoch0 restart stuff is really tricky to get right and greatly complicates (already complicated) handshake code. I do know it (was) being made use of, because it was added due to a specific request [car manufacturer using DTLS between ECUs over the CAN bus; occasionally an ECU reboots for whatever reason, and there is no port number or other thing to disambiguate the associations]. In retrospect I should have honestly rejected the request and told them to synthesize an association ID (eg by prepending each packet with a 128 bit random string generated at device startup). In fact I think I will deprecate this feature; not sure if it makes sense to remove for Botan4 since that's already only a year away, but at the least it indicates this isn't something to be relied upon for the long term, especially if it's not going to be supported in 1.3

No good ideas offhand about how to indicate this state. Magic exception seems maybe best. The main tricky thing with an exception that it can't tear down the existing association, or you create an instant 1 packet DoS. Maybe easier: when it happens, discard the packet as with any other invalid junk, but invoke an `Callback` that by default does nothing? Application can override and handle it. Would need to provide enough context to the call that the application could actually handle it correctly though. [Addition: Callback approach has the nice benefit that we could probably support it in both 1.2 and 1.3 without a bunch of duplicated work]

reneme:
Good that we agree on the epoch0-restart implementation strategy. I like the callback approach to signal that a new ClientHello arrived and will keep that in mind. There's still quite some way to go in the DTLS 1.3 work before that anyways.

src/lib/tls/tls12/tls_channel_impl_12.cpp
randombit: I think that makes assumptions about the application that are not necessarily true. It chose to use a datagram protocol over a lossy channel; by definition it must be willing to accept packet loss and replays in its traffic. If it wishes to have a channel that does something better than 100% packet loss, it must add acks, retries, timeouts, etc with respect to its own protocols, whatever they may be. In contrast when a peer calls `renegotiate`, they are invoking into the DTLS protocol, which has already defined set of retries and timeouts (some parts set out by the RFCs, some parts set by local policy). The DTLS rehandshake protocol can fail, but the association is still valid and the application can continue to use it with its own timeout logic. The application decides if the association is timed out, we are only authoritative on the state of the handshake.

src/bogo_shim/bogo_shim.cpp
randombit: done

src/bogo_shim/bogo_shim.cpp
randombit: done

src/lib/tls/tls12/tls_channel_impl_12.h
randombit: done for both

randombit: fixed

src/lib/tls/tls12/tls_handshake_io.cpp
randombit: Done, I also documented what policy tweaks are necessary to disable the timeouts

src/lib/tls/tls12/tls_channel_impl_12.cpp
reneme: ... however, if the epoch rolls over we abort the association entirely. In the TLS 1.2 handshake that epoch roll-over happens before the Finished messages, correct? So within the server's first flight. Therefore, if the server responded at least once, we assume that communication was possible (albeit perhaps only very briefly). Any timeout after that is treated as fatal for the association. That does make some sense to me, because it puts the DTLS protocol in a position to have an "opinion" on the healthiness of the communication.

_No objection to leave it like that._ I just wonder how a downstream application could actually benefit from this behavior in a reasonable way. Especially given that there's no dedicated API feedback when abandoning a handshake apart from the exception coming from `TLS::Channel::timeout_check()` (I believe). And even that may or may not mean that the association was killed.

reneme:
> In retrospect I should have honestly rejected the request and told them to synthesize an association ID (eg by prepending each packet with a 128 bit random string generated at device startup).

That's more or less what the ConnectionID in DTLS 1.3 should achieve, no? [RFC 9146](https://www.rfc-editor.org/rfc/rfc9146.html) specifies that for DTLS 1.2 as well. Maybe it is worth implementing this and then deprecating the epoch0 restart with decent documentation on how to handle it on application level.

src/lib/tls/tls12/tls_channel_impl_12.cpp
randombit: Valid point.

I think in the interests of unblocking I'm going to merge this as it stands, and look forward to next year when renegotiation support can just be completely dropped, which will enormously simplify this code.

PR #5794, opened 2026-08-04
Note in SemVer docs that we will break as needed for security

randombit:
This has always been the unofficial position, and it has been used a few times over the years. Make it clear that this is something that happens.

### DIFF
--- a/doc/sem_ver.rst
+++ b/doc/sem_ver.rst
@@ -13,7 +13,16 @@ way that causes trouble, it is probably a bug; please report it on Github.
 There are important exceptions to the SemVer guarantees that you
 should be aware of, described in the following list.
 
-Exception #1: Deriving from Library Classes
+Exception #1: Security Trumps SemVer
+--------------------------------------------
+
+If fixing a significant security issue requires breaking compatibility for some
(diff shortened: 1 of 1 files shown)


doc/sem_ver.rst
```diff
@@ -13,7 +13,16 @@ way that causes trouble, it is probably a bug; please report it on Github.
...
 
-Exception #1: Deriving from Library Classes
+Exception #1: Security Trumps SemVer
+--------------------------------------------
+
+If fixing a significant security issue requires breaking compatibility for some
+specific class of applications, this breakage will occur. In any case where this
+is necessary, the developers try to minimize the overall impact and scope of the
+breakage. The nature and reasoning for the change will clearly called out in the
+release notes for that version.
```
reneme:
```suggestion
breakage. The nature and reasoning for the change will be clearly called out in
the release notes for that version.
```

..., I believe.

reneme (approved):
Good idea. One language nit.

doc/sem_ver.rst
randombit: you are correct, thanks

PR #5796, opened 2026-08-04
Fix native ownership of Python SRP server sessions

### LINKED ISSUE #5795: Python SRP server sessions share native state between instances
`botan3.Srp6ServerSession` stores its owning `ctypes.c_void_p` handle in a class attribute. `__init__()` passes that shared container by reference to `botan_srp6_server_session_init`, so constructing a second Python session overwrites the handle used by every existing instance.

Relevant code: [`Srp6ServerSession` defines `__obj` on the class and reuses it during construction, operations, and destruction](https://github.com/randombit/botan/blob/32556ae61220d4efc2314d43c8d4b3f7b8ff7aac/src/python/botan3.py#L3641-L3672).

As a result, more than one live server session in the same Python process does not remain independent:

- Operations on an older object are routed to the most recently constructed native SRP session.
- An older login can return the key derived for a different overlapping login instead of its own state.
- Destruction of either wrapper can invalidate the handle still referenced by its peer, producing native heap use-after-free reads on later use or destruction.

Constructing two sessions sequentially, calling `step1()` on each, and then calling `step2()` on the first object reproduces the state substitution. The existing single-session flow continues to work, which is why the current test does not detect the problem.

The expected behavior is that each constructed `Srp6ServerSession` owns a distinct native handle for its entire lifetime. The public Python API does not document a one-live-session-per-process restriction. The C++ SRP API, C FFI API, and stateless Python SRP client helper are unaffected.

The issue was introduced with the Python SRP wrapper and is present from Botan 3.0.0 through 3.12.0 and current master.

fcasal:
## Summary

`Srp6ServerSession` currently stores its native pointer in a mutable class attribute. Constructing a second session therefore replaces the pointer used by existing objects, causing overlapping sessions to route operations and destruction to the newest native session.

This change gives each Python session independent native ownership.

Fixes #5795.

## Changes

- Initialize the native handle as an instance attribute before calling the FFI constructor.
- Detach and null the instance handle before destruction so cleanup is partial-initialization-safe and exactly-once.
- Reject shallow and deep copies, which would otherwise duplicate ownership of the native handle.
- Add a regression that constructs overlapping sessions for different accounts and verifies that the first object does not return the second session's key.
- Add regressions requiring shallow and deep copies of a live session to fail.

## Test Results

At the test-only commit, both new regressions fail normally: the overlapping-session test observes the second session's key through the first object, and the copy-ownership test observes that copying succeeds.

With the fix applied, both regressions and the existing single-session SRP test pass:

```text
Ran 3 tests in 0.005s

OK
```

The regression verification used one Python process and sequential execution. The vulnerable-path cleanup in the tests detaches an aliased handle before assertion unwinding, so the fail-before run does not trigger the separate duplicate-destruction fault.

### DIFF
--- a/src/python/botan3.py
+++ b/src/python/botan3.py
@@ -3667,17 +3667,25 @@ def nist_key_unwrap_padded(kek: bytes, wrapped: bytes, cipher: str | None = None
 class Srp6ServerSession:
     """The server side of the SRP-6a password authenticated key exchange"""
 
-    __obj = c_void_p(0)
-
     def __init__(self, group: str):
         """Create a session using the named group (eg "modp/srp/2048")"""
+        self.__obj = c_void_p(0)
         _DLL.botan_srp6_server_session_init(byref(self.__obj))
         self.__group = group
         self.__group_size = _call_fn_returning_sz(
             lambda len: _DLL.botan_srp6_group_size(_ctype_str(group), len))
 
     def __del__(self):
-        _DLL.botan_srp6_server_session_destroy(self.__obj)
+        obj = getattr(self, '_Srp6ServerSession__obj', None)
+        self.__obj = c_void_p(0)
+        if obj:
+            _DLL.botan_srp6_server_session_destroy(obj)
+
+    def __copy__(self):
+        raise TypeError('Srp6ServerSession objects cannot be copied')
+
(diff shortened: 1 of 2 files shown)


randombit (approved):
LGTM thanks @fcasal!

PR #5797, opened 2026-08-04
Ensure TLS roles are respected during session resumption

randombit:
An application which acted as both a TLS server and a TLS client, and which shared a session manager between those two roles, could end up accepting a resumed session that resumed in the wrong role.

### DIFF
--- a/src/lib/tls/tls12/tls_server_impl_12.cpp
+++ b/src/lib/tls/tls12/tls_server_impl_12.cpp
@@ -82,6 +82,13 @@ std::optional<Session> check_for_resume(const Session_Handle& handle_to_resume,
       return std::nullopt;
    }
 
+   // A session manager shared with our client role must not resume a session
+   // here that we stored as a client: its peer certificates are a remote
+   // server's, and this connection would adopt them as the client's.
+   if(session->side() != Connection_Side::Server) {
+      return std::nullopt;
+   }
+
    // wrong version
    if(client_hello->legacy_version() != session->version()) {
       return std::nullopt;
--- a/src/lib/tls/tls13/tls_extensions_psk.cpp
+++ b/src/lib/tls/tls13/tls_extensions_psk.cpp
@@ -303,11 +303,16 @@ std::unique_ptr<PSK> PSK::select_offered_psk(std::string_view host,
          session_mgr.choose_from_offered_tickets(psk_identities, cipher.prf_algo(), callbacks, policy)) {
       auto& [session, psk_index] = selected_session.value();
 
+      // A session manager shared with our client role must not resume a session
+      // here that we stored as a client: its peer certificates are a remote
+      // server's, and this connection would adopt them as the client's.
+      const bool established_as_server = session.side() == Connection_Side::Server;
+
       // Refuse to resume a ticket across SNI: a session minted for one
       // virtual host must not be presentable against another. Treat as a
       // cache miss and fall through to the external PSK path rather than
       // failing the connection.
-      if(session.server_info().hostname() == host) {
+      if(established_as_server && session.server_info().hostname() == host) {
          // RFC 8446 4.6.1
          //    Any ticket MUST only be resumed with a cipher suite that has the
          //    same KDF hash algorithm as that used to establish the original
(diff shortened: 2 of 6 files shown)


src/lib/tls/tls_session_manager.cpp
```diff
@@ -112,6 +120,16 @@ std::vector<Session_with_Handle> Session_Manager::find_and_filter(const Server_I
...
 
+      // The counterpart to the role check in retrieve(): a session established
+      // while acting as a TLS server must not be offered for resumption as a
+      // client. These are left in storage as they remain valid for the server
+      // role, so a retry would just yield them again.
+      std::erase_if(sessions_and_handles,
+                    [](const auto& session) { return session.session.side() != Connection_Side::Client; });
+      if(sessions_and_handles.empty()) {
+         break;
+      }
```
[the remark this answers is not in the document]
randombit: Failed resumption is not the end of the world and applications acting in both roles are rare. Could consider modifying the session manager API for Botan4 to handle role split directly though.

src/lib/tls/tls_session_manager.cpp
```diff
@@ -112,6 +120,16 @@ std::vector<Session_with_Handle> Session_Manager::find_and_filter(const Server_I
          break;
       }
 
+      // The counterpart to the role check in retrieve(): a session established
+      // while acting as a TLS server must not be offered for resumption as a
+      // client. These are left in storage as they remain valid for the server
+      // role, so a retry would just yield them again.
+      std::erase_if(sessions_and_handles,
+                    [](const auto& session) { return session.session.side() != Connection_Side::Client; });
```
reneme, line 128: Why not just move this filter before the early-return before. Then there would be no need to for the additional early-return below.

src/lib/tls/tls13/tls_extensions_psk.cpp
```diff
@@ -303,11 +303,16 @@ std::unique_ptr<PSK> PSK::select_offered_psk(std::string_view host,
          session_mgr.choose_from_offered_tickets(psk_identities, cipher.prf_algo(), callbacks, policy)) {
       auto& [session, psk_index] = selected_session.value();
 
+      // A session manager shared with our client role must not resume a session
+      // here that we stored as a client: its peer certificates are a remote
+      // server's, and this connection would adopt them as the client's.
+      const bool established_as_server = session.side() == Connection_Side::Server;
```
reneme, line 309: `Session_Manager::choose_from_offered_tickets()` should also filter for `.side() == ::Server` because clients never get offered any sessions for resumptions. Then, this could become a `BOTAN_ARG_CHECK`, or better a conditional `Alert::Internal_Error`.

src/lib/tls/tls12/tls_server_impl_12.cpp
```diff
@@ -82,6 +82,13 @@ std::optional<Session> check_for_resume(const Session_Handle& handle_to_resume,
       return std::nullopt;
    }
 
+   // A session manager shared with our client role must not resume a session
+   // here that we stored as a client: its peer certificates are a remote
+   // server's, and this connection would adopt them as the client's.
+   if(session->side() != Connection_Side::Server) {
+      return std::nullopt;
+   }
```
reneme, line 90: `Session_Manager::retrieve()` is now filtering for `.side() == ::Server`, so this could technically be a hard error (i.e. `BOTAN_ARG_CHECK` or `Alert::Internal_Error`). That would produce a tangible error message for users to act on, rather than just a silent resumption-failure if the assertion should ever break.

reneme (commented):
In fact, I would expand on the proposed changes (perhaps in a follow-up), like so:

1. **Extend the documentation of the virtual storage interface methods**: It already states that `retrieve_one` is only called on the server side and likewise `find_some` only on the client side. Those docstrings should also mention that the methods _must_ filter their outputs accordingly (making the filters introduced in this PR a belt-and-suspenders idiom ensuring that custom session managers don't get funky).
2. **Adapt our `TLS::Session_Manager` implementations**: we provide a bunch of them, and they should filter their outputs accordingly.

As a result, applications that hold both client and server state in the same place should have a much higher chance of actually benefitting from resumption as expected.

PR #5799, opened 2026-08-04
Make legacy McEliece GF2m caches thread-safe

fcasal:
## Summary

Legacy McEliece's exponent and logarithm tables used unsynchronized check-then-assignment into process-global vectors. Concurrent first construction of separate private keys could race vector publication and deallocation, causing use-after-free and double-free.

Botan documents that separate objects may be used concurrently. The regression uses a separate independently seeded RNG and key construction in every thread, so the caller does not share a Botan object.

## Changes

- Construct the complete table sets through thread-safe function-local initialization.
- Publish the completed tables as immutable vectors.
- Preserve field arithmetic and supported extension degrees 2 through 15.
- Add a cold, synchronized concurrency regression using independent field objects.

## Testing

The regression was checked in a minimized ASan/UBSan build:

```text
./configure.py --minimized-build --enable-modules=mce --build-targets=static,tests --build-tool=ninja --without-documentation --enable-sanitizers=address,undefined --with-debug-info
ninja -j4 botan-test
ASAN_OPTIONS=detect_leaks=0:halt_on_error=1 ./botan-test gf2m
```

On the test-only commit, the first run aborts with an ASan heap-use-after-free in `Botan::GF2m_Field::GF2m_Field()`. The freed allocation and read are reached through separate workers in the new regression.

With the fix applied, the same command completes without sanitizer findings:

```text
GF2m ran 131064 tests all ok
GF2m concurrent initialization ran 1 tests all ok
Tests complete ran 131065 tests in 4.92 msec all tests ok
```

The separate-object behavior, valid and invalid degree boundaries, and field-operation compatibility were also checked independently.

### DIFF
--- a/src/lib/pubkey/mce/gf2m_small_m.cpp
+++ b/src/lib/pubkey/mce/gf2m_small_m.cpp
@@ -12,6 +12,7 @@
 
 #include <botan/exceptn.h>
 #include <botan/internal/ct_utils.h>
+#include <array>
 #include <string>
 
 namespace Botan {
@@ -19,6 +20,8 @@ namespace Botan {
 const size_t MAX_EXT_DEG = 15;
 
 namespace {
+
+using GF2m_Tables = std::array<std::vector<gf2m>, MAX_EXT_DEG + 1>;
 
 const gf2m prim_poly[MAX_EXT_DEG + 1] = {
    01,      /* extension degree 0 (!) never used */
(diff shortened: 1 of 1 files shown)


randombit:
Thanks @fcasal change looks good there is one tiny formatting issue (header inclusion order) and can you drop the test commit d0cdf83 entirely - we legitimately support systems without thread support (arm32 baremetal CI failing for this reason) and while the poc is good to prove it was a real issue I don't think it's worth the extra complexity of shipping the test.

randombit (approved):

PR #5801, opened 2026-08-05
In (D)TLS discard old cipher state, and prevent DTLS epoch wraparound

randombit:
Extracted from #5783

### DIFF
--- a/src/lib/tls/tls12/tls_channel_impl_12.cpp
+++ b/src/lib/tls/tls12/tls_channel_impl_12.cpp
@@ -38,6 +38,32 @@ bool is_new_dtls_association_client_hello(std::span<const uint8_t> msg_and_heade
           load_be(msg_and_header.subspan<4, 2>()) == 0;
 }
 
+// Cap on the number of non-zero-epoch cipher states retained per direction.
+// The current epoch is in active use; one earlier epoch covers any in-flight
+// DTLS records sent immediately before a rekey. Epoch 0 is the pre-handshake
+// plaintext placeholder and is always retained.
+//
+// Note the logic in prune_epochs in tls_seq_numbers.h assumes this is exactly 2
+// so adjust that code if changing this.
+constexpr size_t TLS_RETAINED_CIPHERSTATES = 2;
+
+// The "default MSL specified for TCP" of RFC 6347 4.1; RFC 793 sets it to two
+// minutes. Bounds how long a retired DTLS read epoch stays usable.
+constexpr uint64_t TCP_MSL_MS = 2 * 60 * 1000;
+
+template <typename T>
+void prune_old_cipher_states(std::map<uint16_t, T>& states) {
+   // std::map iterates in ascending key order. Drop the lowest non-zero
+   // entries until at most TLS_RETAINED_CIPHERSTATES remain. The newly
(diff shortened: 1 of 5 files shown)


src/lib/tls/tls12/tls_seq_numbers.h
```diff
@@ -184,6 +190,38 @@ class Datagram_Sequence_Numbers final : public Connection_Sequence_Numbers {
...
+      template <typename T>
+      static void prune_epochs(std::map<uint16_t, T>& epochs, uint16_t current) {
+         for(auto i = epochs.begin(); i != epochs.end();) {
+            if(i->first == 0 || i->first + 1 >= current) {
+               ++i;
+            } else {
+               i = epochs.erase(i);
+            }
+         }
+      }
```
reneme, line 226: This is implicitly assuming the value of `TLS_RETAINED_CIPHERSTATES` in the tls_channel_impl_12.cpp, right? Perhaps define that in this header and use it in both places?

reneme (approved):
One double-check. Otherwise looks good to me.

src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -103,7 +126,20 @@ std::shared_ptr<Connection_Cipher_State> Channel_Impl_12::read_cipher_state_epoc
...
+   // earlier epochs, but if packet loss causes noticeable problems they MAY
+   // choose to retain keying material from previous epochs for up to the
+   // default MSL specified for TCP [TCP] to allow for packet reordering."
+   // read_dtls_record drops the record when this throws.
+   if(const auto& retired_at = i->second.retired_at; retired_at.has_value()) {
+      if(callbacks().tls_current_monotonic_clock_ms() - retired_at.value() > TCP_MSL_MS) {
+         throw Invalid_State("TLS::Channel_Impl_12 Read cipherstate for epoch " + std::to_string(epoch) +
+                             " is past its retention window");
+      }
+   }
```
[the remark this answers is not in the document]
randombit: If the monotonic clock is not monotonic, all bets are off

src/lib/tls/tls12/tls_seq_numbers.h
randombit: It is. I'll see if I can clean this up a bit more. For TLS really we just need to keep 1 epoch around since there is never any kind of traffic replay, but that would be more complicated to handle I think.

[As with everything this all gets simpler once renegotation support is removed...]

src/lib/tls/tls12/tls_channel_impl_12.cpp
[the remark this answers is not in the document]
randombit: I guess we can add an assert here though, internal error is entirely appropriate for misbehaving clock...

src/lib/tls/tls12/tls_seq_numbers.h
reneme: Not the hill I'd want to die on. A prominent comment might be enough.

randombit: Added some comments at each site pointing in the other direction. Enough I think, given this code lives for ~1 year before being removed.

PR #5805, opened 2026-08-06
Fix an integer overflow in botan_block_cipher_block_size

randombit:


randombit:
@reneme don't particularly need a review here (obv feel free) but wanted to draw your attention to C++20's `std::in_range` https://en.cppreference.com/cpp/utility/in_range which I definitely did not know existed until today...

reneme:
Thanks, I didn't know about `std::in_range` either. Sometimes, it really are the little things!

PR #5807, opened 2026-08-06
Use the encoded byte length in Python PKSign.update

### LINKED ISSUE #5806: Python PKSign truncates non-ASCII string messages before signing
## Summary

The Python wrapper's `PKSign.update()` method UTF-8 encodes a `str` but
passes the original Python character count to an FFI function whose length is
measured in bytes. Non-ASCII strings are therefore truncated before signing.

The affected code is
[`PKSign.update()`](https://github.com/randombit/botan/blob/38bb4bff9873f69e285fad083e00a8f8699f407f/src/python/botan3.py#L2224-L2227):

```python
_DLL.botan_pk_op_sign_update(self.__obj, _ctype_bits(msg), len(msg))
```

`_ctype_bits(msg)` returns the complete UTF-8 byte string, while `len(msg)`
counts Unicode code points. The native
[`botan_pk_op_sign_update` contract](https://github.com/randombit/botan/blob/38bb4bff9873f69e285fad083e00a8f8699f407f/src/lib/ffi/ffi.h#L3155-L3163)
interprets that length as the number of bytes to read.

This appears to be an accidental unit mismatch introduced when
[raw-byte support changed the conversion helper](https://github.com/randombit/botan/commit/9fec5a85538ef3b285c2a8d82e179ce32e096fb5)
without changing the accompanying length expression.

## Reproducer

```python
import botan3

rng = botan3.RandomNumberGenerator()
private_key = botan3.PrivateKey.create("RSA", "1024", rng)
public_key = private_key.get_public_key()

message = "éX"  # 2 Python characters, but 3 UTF-8 bytes

signer = botan3.PKSign(private_key, "PKCS1v15(SHA-256)")
signer.update(message)
signature = signer.finish(rng)

original = botan3.PKVerify(public_key, "PKCS1v15(SHA-256)")
original.update(message)
assert not original.check_signature(signature)

truncated = botan3.PKVerify(public_key, "PKCS1v15(SHA-256)")
truncated.update("é")
assert truncated.check_signature(signature)
```

Although `"éX"` encodes to three bytes, `PKSign.update()` passes a length of
two. The signer therefore receives only the two-byte encoding of `"é"`: the
signature does not verify for the requested message, but does verify for the
truncated one.

The behavior reproduces for one-shot and segmented updates. ASCII and `bytes`
inputs are unaffected because their supplied length already equals their byte
length.

## Suggested fix

Encode once and pass the encoded byte length:

```python
bits = _ctype_bits(msg)
_DLL.botan_pk_op_sign_update(self.__obj, bits, len(bits))
```

Add a regression covering both one-shot and segmented non-ASCII input.

fcasal:
## Summary

`PKSign.update(str)` converts its input to UTF-8 but currently passes the
Unicode code-point count to `botan_pk_op_sign_update()`, whose length argument
is measured in bytes. As a result, non-ASCII strings are signed only up to the
shorter character count.

Encode the input once and pass `len(bits)`, matching `PKVerify.update()` and
the FFI contract.

Fixes #5806.

## Tests

Add `BotanPythonTests.test_pksign_utf8_input` covering:

- a one-shot message containing two-, three-, and four-byte UTF-8 characters;
- the same message divided across multiple `update()` calls.

The regression-only commit fails both cases against current master. The fix
commit passes both cases.

```text
Regression-only 76e7f33a8: FAILED (failures=2)
Fixed           7df88179b: OK
```

Focused command:

```sh
python3 src/scripts/test_python.py BotanPythonTests.test_pksign_utf8_input
```

### DIFF
--- a/src/python/botan3.py
+++ b/src/python/botan3.py
@@ -2224,7 +2224,8 @@ class PKSign: # pylint: disable=invalid-name
 
     def update(self, msg: str | bytes):
         """Add more data to be signed"""
-        _DLL.botan_pk_op_sign_update(self.__obj, _ctype_bits(msg), len(msg))
+        bits = _ctype_bits(msg)
+        _DLL.botan_pk_op_sign_update(self.__obj, bits, len(bits))
 
     def finish(self, rng_obj: RandomNumberGenerator) -> bytes:
         """Returns the signature of the message provided so far, and resets for another message"""
--- a/src/scripts/test_python.py
+++ b/src/scripts/test_python.py
@@ -706,6 +706,26 @@ ofvkP1EDmpx50fHLawIDAQAB
         verify = botan.PKVerify(pk, param_str)
         verify_positive_and_negative(verify, sig)
 
+    def test_pksign_utf8_input(self):
+        # A Python str's character count can be shorter than the length of its
+        # UTF-8 encoding. PKSign must pass the encoded byte length to the FFI,
+        # both for a single update and when the message is split across calls.
+        rng = botan.RandomNumberGenerator()
+        private_key = botan.PrivateKey.create('RSA', '1024', rng)
+        public_key = private_key.get_public_key()
+        unicode_msg = "sign this: café € 😀"
+
+        for updates in [[unicode_msg], ["sign this: ", "café ", "€ ", "😀"]]:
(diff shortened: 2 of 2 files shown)


randombit (approved):
Nice find, thanks

PR #5809, opened 2026-08-06
Propagate Threaded_Fork branch exceptions

### LINKED ISSUE #5808: Threaded_Fork converts branch exceptions into process termination
`Threaded_Fork` lets exceptions from branch filters escape its worker threads, causing `std::terminate`. Malformed input that plain `Fork` reports as a catchable Botan exception can therefore abort the process.

The affected call is [`filter->write()` in `Threaded_Fork::thread_entry()`](https://github.com/randombit/botan/blob/32556ae61220d4efc2314d43c8d4b3f7b8ff7aac/src/lib/filters/threaded_fork.cpp#L134-L147). The [public documentation](https://github.com/randombit/botan/blob/32556ae61220d4efc2314d43c8d4b3f7b8ff7aac/src/lib/filters/filters.h#L870-L899) calls this class a threaded, drop-in replacement for `Fork`.

A reproducer uses only public APIs and a built-in
`Base64_Decoder(FULL_CHECK)`:

```cpp
#include <botan/exceptn.h>
#include <botan/filters.h>
#include <botan/pipe.h>

#include <iostream>
#include <string_view>

int main() {
   try {
      Botan::Pipe pipe(new Botan::Threaded_Fork(
         new Botan::Base64_Decoder(Botan::FULL_CHECK),
         new Botan::Hex_Encoder));
      pipe.process_msg(std::string_view("@@@@"));
   } catch(const Botan::Exception& e) {
      std::cout << "caught: " << e.what() << '\n';
   }
}
```

The `catch` block is never reached. Instead, the exception escapes the worker
thread and terminates the process:

```text
libc++abi: terminating due to uncaught exception of type Botan::Invalid_Argument:
base64_decode: invalid character '@'
```

The process exits through `SIGABRT` (exit status 134 on the tested system).
Replacing `Threaded_Fork` with plain `Fork` makes the same exception reach the
caller's `catch` block.

fcasal:
## Summary

- Catch exceptions thrown by `Threaded_Fork` branch filters.
- Preserve the first worker exception and rethrow it on the calling thread after all workers reach the completion barrier.
- Clear saved exception state before each dispatch.
- Add a regression test using strict Base64 rejection through the public Pipe API.

Without this boundary, a branch exception escapes `std::thread` and invokes `std::terminate`, so input that plain `Fork` reports as a catchable Botan exception aborts the process.

Fixes #5808.

## Testing

At the regression-only commit, `botan-test filter` aborts with exit 134 on the new test. With the fix:

```text
Threaded_Fork ran 12 tests all ok
Tests complete ran 68 tests in 21.12 msec all tests ok
```

`run_clang_format.py --check --skip-version-check -j 1 src/lib/filters/threaded_fork.cpp src/tests/test_filters.cpp` passes, as does `git diff --check`.

### DIFF
--- a/src/lib/filters/threaded_fork.cpp
+++ b/src/lib/filters/threaded_fork.cpp
@@ -13,6 +13,9 @@
    #include <botan/internal/barrier.h>
    #include <botan/internal/semaphore.h>
 
+   #include <exception>
+   #include <mutex>
+
 namespace Botan {
 
 struct Threaded_Fork_Data {
@@ -38,6 +41,27 @@ struct Threaded_Fork_Data {
       * The length of the work that needs to be done.
       */
       size_t m_input_length = 0;
+
+      void clear_exception() {
+         std::lock_guard lock(m_exception_mutex);
+         m_exception = nullptr;
+      }
+
+      void capture_exception() {
+         std::lock_guard lock(m_exception_mutex);
+         if(!m_exception) {
+            m_exception = std::current_exception();
+         }
+      }
+
+      std::exception_ptr exception() {
+         std::lock_guard lock(m_exception_mutex);
+         return m_exception;
+      }
+
+   private:
+      std::mutex m_exception_mutex;
+      std::exception_ptr m_exception;
 };
 
 /*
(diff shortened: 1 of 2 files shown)


src/lib/filters/threaded_fork.cpp
randombit, line 23: The addition of private member variables later in the struct causes `clang-tidy` to start warning about these public member variables. OK to just wrap in a NOLINTBEGIN...NOLINTEND block

randombit (approved):
LGTM - one clang-tidy warning that can just be disabled by comment. Other CI failures are GH network flakes.

PR #5811, opened 2026-08-06
Ignore unauthenticated epoch-zero input in active DTLS associations

randombit:
Extracted from #5783


randombit:
Force push to poke gha

PR #5813, opened 2026-08-07
Refactor: TLS downgrade does not re-parse Client Hello

reneme:
Before, the ClientHello message was passed from the 1.3-implementation to the 1.2-implementation in marshalled form (with its header). This is fine for TLS, where the handshake message protocol header is equivalent across both versions.

However, DTLS is handling this protocol header differently. DTLS 1.2's transcript hash is calculated over a fictious unfragmented version of the marshalled message, whereas in DTLS 1.3 the transcript hash omits the fragmentation information in the message protocol headers.

Also, this prepares some new API in the TLS 1.2 `Handshake_IO` for DTLS where the `Datagram_Handshake_IO` needs some state set-up after downgrading.

### DIFF
--- a/src/lib/tls/tls_channel_impl.h
+++ b/src/lib/tls/tls_channel_impl.h
@@ -21,6 +21,10 @@
 #include <utility>
 #include <vector>
 
+#if defined(BOTAN_HAS_TLS_DOWNGRADE_SUPPORT)
+   #include <botan/tls_messages_13.h>
+#endif
+
 namespace Botan {
 
 class Credentials_Manager;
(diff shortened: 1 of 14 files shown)


src/lib/tls/tls_channel_impl.h
```diff
@@ -198,7 +202,7 @@ class Channel_Impl {
        */
       struct Downgrade_Information {
             /// The client hello message including the handshake header bytes as transferred to the peer.
-            std::vector<uint8_t> client_hello_message;
+            std::optional<Client_Hello_13> client_hello;
```
reneme, line 205: This is now explicitly referencing the TLS 1.3 version of the Client Hello. Hence, this type is available only in builds that have TLS 1.3 enabled. cf. [my previous comment](https://github.com/randombit/botan/pull/5812#issuecomment-5216965252).

randombit (approved):

PR #5814, opened 2026-08-07
Reject embedded NULs at the Python C-string boundary

fcasal:
# Reject embedded NULs at the Python C-string boundary

Python strings can contain embedded NUL characters, while Botan's FFI string parameters are NUL-terminated. Reject these values centrally in `_ctype_str()` so native code cannot observe a different prefix from the value validated by Python application code.

Add regression coverage for bcrypt passwords, X.509 hostname matching, hostname verification, and trust-directory selection.

## Test plan

- Confirm the bcrypt regression fails against the original wrapper because no `ValueError` is raised.
- Confirm the bcrypt regression passes after `_ctype_str()` rejects the value.
- Confirm the X.509 hostname and trust-path regression tests reject embedded NULs.

### DIFF
--- a/src/python/botan3.py
+++ b/src/python/botan3.py
@@ -739,6 +739,8 @@ def _ctype_str(s: str | None) -> bytes | None:
     if s is None:
         return None
     assert isinstance(s, str)
+    if '\0' in s:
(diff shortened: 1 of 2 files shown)


randombit (approved):
LGTM

PR #5816, opened 2026-08-07
Authenticate OCSP responses in the ocsp_check command

fcasal:
## Summary

`ocsp_check` currently calls `Response::status_for()` directly after fetching an OCSP response. That validates the CertID, status, and time, but does not verify the BasicOCSPResponse signature or signer authorization.

Route the response through Botan's existing PKIX OCSP verification path before reporting success, and return a nonzero exit status for validation failures.

## Changes

- Authenticate downloaded OCSP responses through `PKIX::check_ocsp()`.
- Require an overall successful PKIX result and an explicit `OCSP_RESPONSE_GOOD` status.
- Return exit status 1 for validation failures.
- Add a deterministic CLI regression test with an unsigned `GOOD` response and a valid issuer-signed control.
- Use stable fixtures valid through 2037.

## Test plan

- Test-only commit `20cb5e0fd`: 2 tests, 1 failure; vulnerable CLI accepts the unsigned `GOOD` response with exit status 0.
- Fixed commit `a6831331b`: 2 tests, 0 failures.
- Extended real-CLI matrix passes for unsigned, invalid-signature, unauthorized-responder, revoked, unknown, mismatched-CertID, valid issuer, valid delegated responder, malformed DER, and multiple-AIA retry cases.
- `git diff --check` passes.
- `run_clang_format.py --check` passes for `src/cli/x509.cpp`.


randombit:
Thanks, change looks good, had some fiddly review comments that was easier to just address directly, continued in #5825

PR #5819, opened 2026-08-07
Enforce live client-auth policy during TLS 1.3 resumption

### LINKED ISSUE #5818: TLS 1.3 resumption bypasses required client-certificate authentication
# TLS 1.3 resumption bypasses required client-certificate authentication

Botan's TLS 1.3 server accepts a previously issued resumption ticket without checking it against the live `require_client_certificate_authentication()` policy. Because selecting a ticket suppresses `CertificateRequest`, a client with a ticket minted before mTLS became mandatory can establish an active resumed connection with no authenticated client certificate. This crosses the server's client-authentication boundary.

## Vulnerable code

[`Server_Impl_13::handle_reply_to_client_hello`](https://github.com/randombit/botan/blob/32556ae61220d4efc2314d43c8d4b3f7b8ff7aac/src/lib/tls/tls13/tls_server_impl_13.cpp#L292) imports the stored peer identity without testing the current policy. Later, [PSK selection suppresses `CertificateRequest`](https://github.com/randombit/botan/blob/32556ae61220d4efc2314d43c8d4b3f7b8ff7aac/src/lib/tls/tls13/tls_server_impl_13.cpp#L367).

## Reproducer

The runnable test is included beside this report as `confirming_test.cpp` (SHA-256 `c7807fab9ef19c9a01dfed697e8f3cb4251c52ad23b039bba1b6e08fde18e06c`). Run the following from this finding folder:

The adjacent confirming test contains the reproducible build and run procedure.

The run exits 0 and reports that a fresh mTLS connection is inactive, ticket minting before the policy change is active, and replay after the change is active and resumed with zero peer certificates.

`confirming_test_policy_transition.cpp` independently preserves the same live-policy transition, fresh required-mTLS rejection, and active certificate-free resumption controls.

### Regression trace

The in-tree regression on the test-only commit reproduces the bypass as two failed security invariants:

```text
$ botan-test tls13_client_auth_resumption
exit status: 1
5 assertions run, 2 failed

FAILED: unauthenticated session cannot resume after client auth becomes mandatory
FAILED: rejected session is not reported as resumed
```

With the proposed fix applied, the focused test and the related TLS 1.3/session-manager suites pass:

```text
$ botan-test tls13_client_auth_resumption
exit status: 0
5 assertions run, all passed

$ botan-test tls13_client_auth_resumption tls_rfc8448_server tls_session_manager
exit status: 0
643 assertions run, all passed
```

## Recommended fix

Reject a stored session if its recorded client identity does not satisfy the live authentication policy. Make that decision before committing the resumed session or sending `ServerHello`. Add a regression test that changes an existing session's policy from optional to mandatory client authentication.

fcasal:
# Enforce live client-auth policy during TLS 1.3 resumption

TLS 1.3 ticket acceptance currently suppresses `CertificateRequest` even when the current server policy requires client authentication and the stored session has no authenticated peer. Reject such stale tickets before establishing PSK state. This preserves authentication policy changes across session lifetime and makes the fresh and resumed paths enforce the same invariant.

Fixes #5818.

## Testing

- The regression-only commit fails two of five assertions because the unauthenticated ticket resumes after client authentication becomes mandatory.
- With the fix, `tls13_client_auth_resumption` passes all five assertions.
- `tls13_client_auth_resumption tls_rfc8448_server tls_session_manager` passes all 643 assertions.

### DIFF
--- a/src/lib/tls/tls13/tls_server_impl_13.cpp
+++ b/src/lib/tls/tls13/tls_server_impl_13.cpp
@@ -291,6 +291,12 @@ void Server_Impl_13::handle_reply_to_client_hello(Server_Hello_13 server_hello) 
 
       psk_cipher_state = std::visit(
          overloaded{[&, this](Session session) {
+                       if(policy().require_client_certificate_authentication() && session.peer_certs().empty() &&
+                          !session.peer_raw_public_key()) {
+                          throw TLS_Exception(Alert::AccessDenied,
+                                              "Resumed session does not satisfy current client authentication policy");
+                       }
+
                        m_handshake->resumed_session = std::move(session);
                        return Cipher_State::init_with_psk(Connection_Side::Server,
                                                           Cipher_State::PSK_Type::Resumption,
--- a/src/tests/test_tls13_client_auth_resumption.cpp
+++ b/src/tests/test_tls13_client_auth_resumption.cpp
@@ -1,0 +1,211 @@
+/*
+* TLS 1.3 Client Authentication Resumption Tests
+* (C) 2026 The Botan Authors
+*
+* Botan is released under the Simplified BSD License (see license.txt)
+*/
+
+#include "tests.h"
+
+#if defined(BOTAN_HAS_TLS_13) && defined(BOTAN_HAS_RSA) && defined(BOTAN_HAS_X509_CERTIFICATES)
+   #include <botan/auto_rng.h>
+   #include <botan/credentials_manager.h>
+   #include <botan/rsa.h>
+   #include <botan/tls_callbacks.h>
+   #include <botan/tls_client.h>
+   #include <botan/tls_policy.h>
+   #include <botan/tls_server.h>
+   #include <botan/tls_session_manager_memory.h>
+   #include <botan/x509self.h>
+
+   #include <algorithm>
+   #include <memory>
+   #include <span>
+   #include <utility>
+   #include <vector>
+#endif
+
+namespace Botan_Tests {
+
+namespace {
+
+#if defined(BOTAN_HAS_TLS_13) && defined(BOTAN_HAS_RSA) && defined(BOTAN_HAS_X509_CERTIFICATES)
+
+class Pipe_Callbacks final : public Botan::TLS::Callbacks {
+   public:
+      void tls_emit_data(std::span<const uint8_t> data) override {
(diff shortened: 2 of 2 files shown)


src/lib/tls/tls13/tls_server_impl_13.cpp
```diff
@@ -295,6 +291,12 @@ void Server_Impl_13::handle_reply_to_client_hello(Server_Hello_13 server_hello)
 
       psk_cipher_state = std::visit(
          overloaded{[&, this](Session session) {
+                       if(policy().require_client_certificate_authentication() && session.peer_certs().empty() &&
+                          !session.peer_raw_public_key()) {
+                          throw TLS_Exception(Alert::AccessDenied,
+                                              "Resumed session does not satisfy current client authentication policy");
+                       }
```
reneme, line 298: The robot has made reasonable points here. It should be easy to do those checks (and the one suggested by Copilot) after the binder is validated. To simplify this, I'd like to suggest extending the `Session_Base` type a little. Along those lines (uncompiled, untested):

In `tls_session.h`, add a new method like so:

```C++
/**
 * Return the type of credential used by the peer, nullopt if no credential
 * was used.
 */
std::optional<Certificate_Type> Session_Base::peer_credential_type() const {
   if(!m_peer_certs.empty()) {
      return Certificate_Type::X509;
   }

   if(m_peer_raw_public_key) {
      return Certificate_Type::RawPublicKey;
   }

   return std::nullopt;
}
```

... and then, below the check with `validate_binder()`:

```C++
if(m_handshake->resumed_session.has_value() && policy().require_client_certificate_authentication()) {
   if(!m_handshake->resumed_session->peer_credential_type().has_value()) {
      throw TLS_Exception(Alert::AccessDenied,
                           "Resumed session does not contain mandatory authentication credential");
   }

   if(!value_exists(policy().accepted_client_certificate_types(),
                     m_handshake->resumed_session->peer_credential_type().value())) {
      throw TLS_Exception(Alert::AccessDenied,
                           "Resumed session does not contain an acceptable authentication credential type");
   }
}
```

reneme (changes requested):
The additional check looks reasonable to me. I'd like to extend on it a little, following the suggestions by Copilot.

randombit:
Continued in #5837 fixing some similar issues in TLS 1.2 as well thanks @fcasal

PR #5826, opened 2026-08-08
In the bitsliced AES use the native word size instead of always 32-bit

randombit:
On all tested 64-bit processors (Tiger Lake, POWER10, Cortex X925) this regresses performance somewhat (10-30%) for 1 and 2 block messages, and improves for 3 blocks and up. The improvement for 1024 byte messages is 50 to 70%, compared to the 32-bit implementation.

The short input performance could be clawed back by continuing the use the 32-bit pathway for 1-2 block inputs, but the extra complexity does not seem worth it at this time.

### DIFF
--- a/src/lib/block/aes/aes.cpp
+++ b/src/lib/block/aes/aes.cpp
@@ -1,5 +1,5 @@
 /*
-* (C) 1999-2010,2015,2017,2018,2020 Jack Lloyd
+* (C) 1999-2010,2015,2017,2018,2020,2026 Jack Lloyd
 *
 * Botan is released under the Simplified BSD License (see license.txt)
 */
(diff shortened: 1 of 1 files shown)


randombit:
@reneme ping re the perf - this does regress eg AES/CBC encryption and AES-CMAC on 64-bit processors without hwaes or vperm support. Don't know if this impacts RSCS usage.

reneme:
@FAlbertDev or @neusdan might be in a better position to judge the performance impact. But OTOH, I don't expect the regression for such short messages to matter.

FAlbertDev:
Nice, this sounds great! We only use AES-XTS (with many blocks at once) in performance critical applications 👍

PR #5828, opened 2026-08-08
refactor(utils): Replace custom barrier with std::barrier (C++20)

KaganCanSit:
This PR makes the necessary adjustments to replace the `barrier` class implementation located under `utils/thread_utils` with `std::barrier`, as provided by the C++20 standard. It aims to migrate the existing functionality to the C++ standard library and reduce the code base. However, it does involve some logical changes. 

---

My initial goal here was to resolve the `TODO` messages in the `test_thread_utils.cpp` file. However, I later questioned why there is a separate `barrier` class. I expect this change will not cause any issues in current versions of GCC, Clang, and MSVC.

To be honest, I understand that the filter functions (pipe design) are synchronized here, although I'm still not fully convinced about some aspects of the approach. Passing `n` to the standard does reduce some flexibility, but that flexibility wasn't being used in practice anyway.

I would be happy to have more information and assist with testing if needed. I'll mark the branch as a draft and I spen more time here for now.


src/lib/filters/threaded_fork.cpp
```diff
@@ -111,6 +118,10 @@ void Threaded_Fork::set_next(Filter* f[], size_t n) {
          m_threads.push_back(std::make_shared<std::thread>([this, next = m_next[i]] { thread_entry(next); }));
       }
    }
+
+   if(port_count_changed || !m_thread_data->m_input_complete_barrier.has_value()) {
```
KaganCanSit, line 122: When the number of ports remains unchanged, the protection mechanism prevents the barrier from being destroyed or recreated. This is important because when the `emplace()` function is called on a busy object, the existing `std::barrier` is destroyed; and if this barrier is destroyed while a worker is still inside `arrive_and_wait()`, it leads to undefined behavior (UB).

Can the `set_next()` function be called while a message is being processed after creation? Although the only here seems to indicate that it can only be called from constructors, I don’t want to overlook this possibility.

---

Update:
`set_next()` is a private member function and cannot be accessed from any external interface. The `Fork::Fanout_Filter`, which is inherited, has a `set_next()` function. Therefore, name hiding also occurs. I don’t see a scenario here where the `if(n < m_threads.size())` condition could be triggered. Maybe it’s dead code?

Could it be something like this?

```cpp
void Threaded_Fork::set_next(Filter* f[], size_t n) {
   // Private and non-virtual: only reachable from the constructors, so this runs exactly once per object.
   BOTAN_ASSERT_NOMSG(m_threads.empty() && !m_thread_data->m_input_complete_barrier.has_value());

   Fork::set_next(f, n);
   n = m_next.size();

   m_thread_data->m_input_complete_barrier.emplace(static_cast<std::ptrdiff_t>(n) + 1);

   m_threads.reserve(n);
   for(size_t i = 0; i != n; ++i) {
      m_threads.push_back(std::make_shared<std::thread>([this, next = m_next[i]] { thread_entry(next); }));
   }
}
```

---

Update 2: Actually, the same applies to `semaphore`. The external class can be removed by using `std::counting_semaphore`.

Considering the comment in `filters.h`: `TODO(Botan4) remove this and all associated helpers (like Barrier and Semaphore` - https://github.com/randombit/botan/commit/d717074e418ab6f02144cba1ad9bb2f6c59b4c45) I must have overlooked this yesterday—I’m sorry, should we remove it now, or should we leave it alone until Botan4?

reneme:
Frankly, I think it's not worth switching `Threaded_Fork` over to `std::barrier` anymore. The whole class is deprecated and slated for removal in Botan4. Until then, we can live with the homebrew barrier.

https://github.com/randombit/botan/blob/2445ace0b8f6b0902f86d646c9206d32b9adeba5/src/lib/filters/filters.h#L878-L880

Thanks for looking into this regardless. I learnt something from it: I didn't know the STL provided these abstractions.

KaganCanSit:
> Frankly, I think it's not worth switching `Threaded_Fork` over to `std::barrier` anymore. The whole class is deprecated and slated for removal in Botan4. Until then, we can live with the homebrew barrier.
> 
> https://github.com/randombit/botan/blob/2445ace0b8f6b0902f86d646c9206d32b9adeba5/src/lib/filters/filters.h#L878-L880
> 
> Thanks for looking into this regardless. I learnt something from it: I didn't know the STL provided these abstractions.

You're welcome; I learned that too. It was a bit disappointing that I didn't see the comment about it being removed from Botan4 until later. In that case, I think closing the thread is the best course of action. All the content may be removed from Botan4 at some point. Thank you for the time.

PR #5832, opened 2026-08-10
jitter_rng: make FIPS mode optional, allow NTG.1 and custom OSR

thillux:
I did some small maintenance of the jitter_rng in order to reflect jitterentropy >= 3.7.0 in the code. jitterentropy often lead to failures in our CI runs because of failing health tests in the always enabled FIPS mode. I relaxed this setting, by defaulting to no health tests and letting the user enable them. Anyone actually using this in a FIPS setting nevertheless has to think about the used options, so this change should make no harm here.

This also allows to enable jitterentropy in the NixOS botan package again.

### DIFF
--- a/news.rst
+++ b/news.rst
@@ -7,6 +7,12 @@ Version 3.14.0, Not Yet Released
 * Modify the bitsliced AES implementation to use the native word size of the
   processor, instead of always 32 bits. (GH #5826)
 
+* ``Jitter_RNG`` now accepts the compliance mode and the oversampling rate which
+  should be used by the jitterentropy library. AIS 20/31 NTG.1 compliance can be
+  requested with ``Jitter_RNG::Mode::NTG1`` if jitterentropy 3.7.0 or later is
+  used. Note that ``Jitter_RNG`` no longer forces the FIPS mode of jitterentropy
+  by default; use ``Jitter_RNG::Mode::FIPS`` to retain the previous behavior.
+
 Version 3.13.0, 2026-08-13
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/src/lib/rng/jitter_rng/jitter_rng.cpp
+++ b/src/lib/rng/jitter_rng/jitter_rng.cpp
@@ -10,14 +10,26 @@
 #include <botan/assert.h>
 #include <botan/exceptn.h>
 #include <botan/mutex.h>
+#include <botan/internal/int_utils.h>
 
 #include <jitterentropy.h>
+
+/*
+* BSI AIS 20/31 NTG.1 support (JENT_NTG1) was added in jitterentropy 3.7.0.
+* The version encoding used by JENT_VERSION and jent_version() is
+* major * 1000000 + minor * 10000 + patchlevel * 100.
+*/
+#define BOTAN_JITTER_RNG_NTG1_MIN_VERSION 3070000
+
+#if defined(JENT_NTG1) && JENT_VERSION >= BOTAN_JITTER_RNG_NTG1_MIN_VERSION
+   #define BOTAN_JITTER_RNG_HAS_NTG1
+#endif
 
 namespace Botan {
 
 class Jitter_RNG_Internal final {
    public:
-      Jitter_RNG_Internal();
+      Jitter_RNG_Internal(Jitter_RNG::Mode mode, size_t osr);
       ~Jitter_RNG_Internal();
       void collect_into_buffer(std::span<uint8_t> buf);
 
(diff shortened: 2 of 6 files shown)


src/lib/rng/jitter_rng/jitter_rng.h
```diff
@@ -17,14 +17,59 @@ namespace Botan {
...
+      * Create a Jitter_RNG, throwing if jitterentropy cannot be initialized
+      *
+      * @param mode the compliance mode to request from jitterentropy
+      * @param osr the oversampling rate to use. Higher values collect more
+      * timing samples per output bit, which is slower but more conservative.
+      * The default lets the library apply its own (minimal) rate. Values the
+      * library considers out of range cause the constructor to throw.
+      */
+      explicit Jitter_RNG(Mode mode = Mode::Default, size_t osr = default_osr);
+
```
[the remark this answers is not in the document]
randombit: Irrelevant

src/lib/rng/jitter_rng/jitter_rng.cpp
```diff
@@ -80,25 +134,30 @@ void Jitter_RNG_Internal::collect_into_buffer(std::span<uint8_t> buf) {
...
+   // According to the docs, `jent_read_entropy_safe` itself runs its logic as often
    // as necessary to gather the requested number of bytes,
    // so this should actually never happen.
    BOTAN_ASSERT(static_cast<size_t>(num_bytes) == buf.size(), "JitterRNG produced the expected number of bytes");
 }
 
-Jitter_RNG::Jitter_RNG() : m_jitter{std::make_unique<Jitter_RNG_Internal>()} {}
+Jitter_RNG::Jitter_RNG(Mode mode, size_t osr) : m_jitter{std::make_unique<Jitter_RNG_Internal>(mode, osr)} {}
 
 Jitter_RNG::~Jitter_RNG() = default;
```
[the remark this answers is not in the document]
randombit: Irrelevant

src/lib/rng/jitter_rng/jitter_rng.cpp
```diff
@@ -27,16 +39,58 @@ class Jitter_RNG_Internal final {
...
+}
+
+Jitter_RNG_Internal::Jitter_RNG_Internal(Jitter_RNG::Mode mode, size_t osr) {
+   const auto oversampling_rate =
+      checked_cast_to_or_throw<unsigned int, Invalid_Argument>(osr, "Jitter_RNG oversampling rate is too large");
+   const unsigned int flags = flags_for_mode(mode);
 
    // if flags and osr are used, use the same values for init and alloc
-   static const int result = jent_entropy_init_ex(oversampling_rate, flags);
+   const int result = jent_entropy_init_ex(oversampling_rate, flags);
```
randombit, line 99: This will repeatedly re-initialize - is this safe? What happens when you initialize twice with different flags?

src/tests/test_jitter_rng.cpp
```diff
@@ -40,6 +41,37 @@ std::vector<Test::Result> test_jitter_rng() {
...
+
+      CHECK("Jitter_RNG NTG.1",
+            [](Test::Result& result) {
+               const auto ntg1 = Botan::Jitter_RNG::Mode::NTG1;
+
+               if(Botan::Jitter_RNG::ntg1_supported()) {
+                  try {
+                     Botan::Jitter_RNG rng(ntg1);
+                     result.test_sz_eq("output produced", rng.random_vec(64).size(), size_t(64));
+                  } catch(const Botan::Internal_Error&) {
```
randombit, line 66: What's throwing `Internal_Error` in this context? Seems like it should be `Not_Implemented` if possible.

randombit (commented):
Seems fine overall, some questions/concerns. Right now CI is using 3.6.2, should this be changed?

randombit (changes requested):
Re previous concerns

thillux:
> Re previous concerns

Sry, will fix this soon. Probably this weekend.

src/lib/rng/jitter_rng/jitter_rng.cpp
thillux: The OSR and flags alter the health tests done. While it makes not much sense to do this e.g. with and without FIPS mode or different OSRs in the same process, it should do a fresh health test every time.

src/tests/test_jitter_rng.cpp
thillux: I took another look at the tests and entropy source code. The failure strings should now make the errors clearer. Internal_Error comes from failing health tests, Not_Implemented from missing NTG.1 mode on jitterentropy < 3.7.0.

thillux:
@randombit  Besides CI may arguing with me, this should be ready for review again.

src/lib/rng/jitter_rng/jitter_rng.cpp
randombit: Could you leave a comment here to this effect ie that initialization in this context just performs new health checks but is otherwise safe, given the lack of any kind of output parameter/state object, repeated init like this looks really sketchy.

news.rst
```diff
@@ -124,6 +124,12 @@ Version 3.13.0, 2026-08-13
 
 * Add ``HashFunction::security_level`` (GH #5746)
 
+* ``Jitter_RNG`` now accepts the compliance mode and the oversampling rate which
+  should be used by the jitterentropy library. AIS 20/31 NTG.1 compliance can be
+  requested with ``Jitter_RNG::Mode::NTG1`` if jitterentropy 3.7.0 or later is
+  used. Note that ``Jitter_RNG`` no longer forces the FIPS mode of jitterentropy
+  by default; use ``Jitter_RNG::Mode::FIPS`` to retain the previous behavior.
```
randombit: Should go into the 3.14.0 block (just added on master, rebase again to pick it up)

src/lib/rng/jitter_rng/jitter_rng.cpp
thillux: Done.

news.rst
thillux: Done.

randombit (approved):
lgtm thanks

PR #5836, opened 2026-08-10
In DTLS distinguish a delayed or replayed handshake message from a current one

randombit:
A new DTLS handshake restarts the message_seq field, but the epoch remains the same (and increments once the CCS is processed). During a DTLS renegotiation, a packet retransmit from the first handshake would take up a reassembly slot, which would block the real message from completing and stall the handshake.

A fragment whose epoch is below the most recently delivered one is now recognized as belonging to an earlier handshake and dropped. A delayed Finished needs a further check. A genuine Finished is sent under the epoch its own ChangeCipherSpec installed, always above the epoch the handshake began in, so one at or below the current epoch cannot be genuine and is now discarded. Messages left buffered from a superseded epoch are also dropped once a later epoch is delivered.

Extracted from #5783

### DIFF
--- a/src/lib/tls/tls12/tls_channel_impl_12.cpp
+++ b/src/lib/tls/tls12/tls_channel_impl_12.cpp
@@ -237,7 +237,7 @@ std::optional<std::string> Channel_Impl_12::external_psk_identity() const {
    return std::nullopt;
 }
 
-Handshake_State& Channel_Impl_12::create_handshake_state(Protocol_Version version) {
+Handshake_State& Channel_Impl_12::create_handshake_state(Protocol_Version version, bool epoch0_restart) {
    if(pending_state() != nullptr) {
       throw Internal_Error("create_handshake_state called during handshake");
    }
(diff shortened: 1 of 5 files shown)


src/lib/tls/tls12/tls_channel_impl_12.cpp
```diff
@@ -267,6 +267,15 @@ Handshake_State& Channel_Impl_12::create_handshake_state(Protocol_Version versio
    m_epochs_before_latest_renegotiation = Epochs_Before_Latest_Renegotiation{sequence_numbers().current_read_epoch(),
                                                                              sequence_numbers().current_write_epoch()};
 
+   // Floor for the pending handshake's reassembly: a delayed record from the
+   // handshake before it arrives under a lower epoch and would otherwise take
+   // the sequence slot the real message needs.
```
reneme: I feel that something went wrong in the wording here. Like so?


```suggestion
   // Floor for the pending handshake's reassembly: a delayed record from the
   // handshake before arrives under a lower epoch and must be rejected. It
   // would otherwise take the sequence slot the real message needs.
```

reneme (approved):

src/lib/tls/tls12/tls_channel_impl_12.cpp
randombit: fixed

PR #5839, opened 2026-08-11
In randomize_with_ts_input prefer system RNG if available

randombit:
Clarify the doc comment so it's clear the goal here is not necessarily exactly that it contains a timestamp, but rather that it contains *some* type of input which should diverge across eg a VM rollback. For that purpose, bytes from the system RNG should be strictly superior to any other option.

### DIFF
--- a/src/lib/rng/rng.cpp
+++ b/src/lib/rng/rng.cpp
@@ -15,9 +15,7 @@
 
 #if defined(BOTAN_HAS_SYSTEM_RNG)
    #include <botan/system_rng.h>
-#endif
-
-#if defined(BOTAN_HAS_OS_UTILS)
+#elif defined(BOTAN_HAS_OS_UTILS)
    #include <botan/internal/os_utils.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


src/lib/rng/rng.cpp
```diff
@@ -27,23 +25,26 @@ namespace Botan {
...
+      store_le(std::span{additional_input}.subspan<8, 4>(), pid);
+      const size_t written = 8 + (pid != 0) ? 4 : 0;
 #else
-      BOTAN_UNUSED(offset);
+      // Nothing to use in this case
+      constexpr size_t written = 0;
 #endif
 
-      this->fill_bytes_with_input(output, additional_input);
+      this->fill_bytes_with_input(output, std::span{additional_input}.first(written));
```
reneme, line 47: In case you feel extra fancy (needs an include of `concat_util.h` to compile successfully for the OS_UTILS case):


```suggestion
      const auto additional_input = [] {
#if defined(BOTAN_HAS_SYSTEM_RNG)
         // If we have a system RNG just read 128 bits from that
         return system_rng().random_array<16>();
#elif defined(BOTAN_HAS_OS_UTILS)
         // Otherwise take clock + pid
         const uint64_t clock = OS::get_high_resolution_clock();
         const uint32_t pid = OS::get_process_id();  // 0 if no PIDs on this system
         return concat(store_le(clock), store_le(pid));
#else
         // Nothing to use in this case
         return std::array<uint8_t, 0>{};
#endif
      }();

      this->fill_bytes_with_input(output, additional_input);
```

... that way `additional_data` naturally translates to a `std::array<>` with a length of 16, 12, or 0.

reneme (approved):

src/lib/rng/rng.cpp
randombit: Fancy indeed. Definitely not bad but feels just a touch more magical than I'd want in this context.

PR #5841, opened 2026-08-12
Refactor: TLS 1.3 Cipher_State (de)protects full Record objects

reneme:
This is preparatory work for DTLS 1.3 (#5000) where the record protection is more involved and more meta-data of the record must be available for (de)protection. It makes sense to handle this similarly for TLS as well.

Concretely, the DTLS record header contains an encrypted sequence number hint. Also, in DTLS records may be out-of-order and some sequence number and epoch guess work might be necessary. All of this requires more logic and information to move into the `Cipher_State`.

**What changed**
- **`Cipher_State` API redesign**: `encrypt_record_fragment()`/`decrypt_record_fragment()` → `protect_record()`/`deprotect_record()`. Protection now assembles the full marshalled record (header + `TLSInnerPlaintext` + AEAD tag) including padding; deprotection takes a complete `Record_TLS`, validates sizes, decrypts in place, strips padding (constant-time), and returns a `Record_Content` (type, sequence number, payload).
- **New `Record_TLS` type** (`tls_record_13.h/cpp`): incrementally re-assembles a record from arbitrary byte streams and owns all wire-format validation (record type, ossified major version, zero-length fragments, `MAX_PLAINTEXT_SIZE` / `MAX_CIPHERTEXT_SIZE_TLS13` limits). Payload buffer is reserved to the exact announced length once the header is complete and can be moved out copy-free via `take_payload()`.
- **`Record_Layer` read path**: the flat read buffer with offset-based compaction was replaced by a `std::deque<Record_TLS>`, eliminating the per-`copy_data` memmove and one payload copy per record.
- Tests: new `Record_TLS` unit tests (fragmented/byte-wise assembly, multi-record buffers), padding round-trip tests, RFC 8448 vectors adapted to the new API.


src/lib/tls/tls13/tls_record_layer_13.h
```diff
@@ -106,6 +92,8 @@ class BOTAN_TEST_API Record_Layer {
       void disable_receiving_compat_mode() { m_receiving_compat_mode = false; }
 
    private:
+      std::deque<Record_TLS> m_incoming_records;
+
       std::vector<uint8_t> m_read_buffer;
       size_t m_read_offset = 0;
```
reneme: This can probably go.

src/lib/tls/tls13/tls_cipher_state.cpp
```diff
@@ -254,30 +261,165 @@ uint64_t Cipher_State::encrypt_record_fragment(const std::vector<uint8_t>& heade
...
+   //    pad. [...] Implementations MUST set the padding octets to all zeros
+   //    before encrypting.
+   result.get().insert(result.end(), plaintext.begin(), plaintext.end());  // content
+   result.get().push_back(to_underlying(type));                            // type
+   result.get().insert(result.end(), padding_bytes, 0x00);                 // zeros (padding)
+
+   BOTAN_ASSERT_NOMSG(result.size() == unprotected_record_length);
+   m_encrypt->set_associated_data(std::span{result}.first<TLS_HEADER_SIZE>());
+   m_encrypt->start(current_nonce(m_write_seq_no++, m_write_iv));
+   m_encrypt->finish(result, TLS_HEADER_SIZE /* skip header when protecting the payload */);
```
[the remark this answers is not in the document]
reneme: That's a very good catch! For the record: I think with an improved API that would be facilitated by #4880 (or a similar effort) we'd be able to better avoid such gotchas.

src/lib/tls/tls13/tls_record_layer_13.cpp
```diff
@@ -312,132 +207,63 @@ std::vector<uint8_t> Record_Layer::prepare_records(const Record_Type type,
...
+   m_incoming_records.pop_front();
 
-   // The first received record(s) are likely a client or server hello. To be able to
-   // perform protocol downgrades we must be less vigorous with the record's
-   // legacy version. Hence, `check_tls13_version` is `false` for the first record(s).
-   const TLSPlaintext_Header plaintext_header({header_begin, header_end}, !m_receiving_compat_mode);
+   // RFC 9846 5.1
+   //    legacy_record_version:  MUST be set to 0x0303 for all records
+   //                            generated by a TLS 1.3 implementation
+   if(!m_receiving_compat_mode && record.legacy_version() != Protocol_Version::TLS_V12) {
```
randombit, line 237: Is this flag needed for DTLS or ? It looks very strange when paired with a comment indicating this is a MUST, with no other explanation why we want to accept invalid

reneme: No, it is also needed for TLS 1.3. Essentially, all records but the very first must feature the ossified legacy version. The very first can also say 0x0301 (I believe), for compatibility reasons.

This check used to live in the now-removed `TLS_Plaintext_Header` helper class, where the flag was called `check_tls13_version` which made it significantly clearer.

```C++
         // RFC 8446 5.1
         //    legacy_record_version:  MUST be set to 0x0303 for all records
         //                            generated by a TLS 1.3 implementation
         if(check_tls13_version && m_legacy_version.version_code() != 0x0303) {
            throw TLS_Exception(Alert::IllegalParameter, "Received unexpected record version");
         }
```

I'll extend the RFC reference comment accordingly.

reneme: Done.

src/lib/tls/tls13/tls_cipher_state.cpp
```diff
@@ -254,30 +261,165 @@ uint64_t Cipher_State::encrypt_record_fragment(const std::vector<uint8_t>& heade
...
+   // If the protected record contains less bytes than the expected AEAD tag we
+   // can already fail early because the decryption will fail anyway.
+   if(record.payload().size() < m_decrypt->minimum_final_size()) {
+      throw TLS_Exception(Alert::BadRecordMac, "incomplete record mac received");
+   }
+
+   // RFC 9846 6.2
+   //    record_overflow: A TLSCiphertext record was received that had a length
+   //    more than 2^14 + 256 bytes, or a record decrypted to a TLSPlaintext
+   //    record with more than 214 bytes (or some other negotiated limit).
```
randombit, line 333: 2^14

src/lib/tls/tls13/tls_cipher_state.cpp
```diff
@@ -96,11 +96,15 @@
 #include <botan/hash.h>
 #include <botan/secmem.h>
 #include <botan/tls_ciphersuite.h>
+#include <botan/tls_exceptn.h>
 #include <botan/tls_magic.h>
 
+#include <botan/internal/concat_util.h>
```
randombit, line 102: Unused included?

src/lib/tls/tls13/tls_record_layer_13.cpp
```diff
@@ -14,6 +14,7 @@
 #include <botan/tls_policy.h>
 #include <botan/tls_version.h>
 #include <botan/internal/ct_utils.h>
```
randombit, line 16: Unused now I think

randombit (approved):
A few nits, otherwise lgtm

reneme:
Gnah. I made the requested edits but screwed up the commit amendment somehow. I'll incorporate them in some related PR in the future -- don't think its worth a dedicated follow-up PR. Sorry.

Issue #5777, opened 2026-07-31
[TLS 1.3] Modify how record padding is requested


randombit:
@reneme 

I was just thinking about this and I realized there is something we could do (ideally before 3.13 when the new APIs land in a release):

Right now the interface is

```virtual std::optional<uint16_t> minimum_record_size() const;```

What I'd suggest as an alternative/replacement is

```virtual size_t record_padding_bytes(size_t plaintext_bytes) const```

which strictly subsumes `minimum_record_size`.

* If `minimum_record_size` would return `nullopt` (or 0), disabling padding, equivalent `record_padding_bytes` is ignoring the plaintext bytes and returning 0.
* If `minimum_record_size` returns any value N, requesting small records be padded up to N bytes, then equivalent `record_padding_bytes` is `if(plaintext_bytes < N) { return N - plaintext_bytes; } else { return 0; }`

The difference is this can express padding policies that `minimum_record_size` can't handle, like:

* Pad all records to the next 32-byte boundary.
* Pad records shorter than 512 bytes to exactly 512 bytes, and all larger records to the next 32-byte boundary.
* Randomized padding lengths (I'm not sure if this is necessarily a good idea, but you could do it.)

wdyt

reneme:
Yeah, fine with me. In fact I had a similar design in mind and then settled on the simplest (fairly static) approach that's currently upstream.
Shall I adapt the implementation like that?

@atreiber94 FYI.

randombit:
Sounds good thank you. I should note that this does make it more difficult to express via the text policy - I'm not sure how much that matters to RSCS (or anyone else).

reneme:
We could make it a callback, which it essentially is with the new approach anyway. Text policy isn't an issue for us.

randombit:
Done in #5843

PR #5843, opened 2026-08-12
Make TLS 1.3 record padding more configurable

randombit:
The initial implementation only allowed setting a minimum record size.  Change this so instead the policy indicates the number of padding bytes that should be applied, relative to the size of the plaintext. This allows not just setting a minimum record size but also for example rounding the record to the next multiple of 32 bytes.

#5777

Rather more churn than I expected tbh but overall worth it.

### DIFF
--- a/src/lib/tls/tls_policy.h
+++ b/src/lib/tls/tls_policy.h
@@ -473,20 +473,23 @@ class BOTAN_PUBLIC_API(2, 0) Policy /* NOLINT(*-special-member-functions) */ {
       virtual std::optional<uint16_t> record_size_limit() const;
 
       /**
-       * Defines the minimum plaintext size of any protected TLS 1.3 record.
-       * If an outgoing record contains less plaintext payload than this value,
-       * it will be transparently padded to reach the specified minimum size.
+       * Defines the number of padding octets added to a protected TLS 1.3
+       * record that contains @p plaintext_bytes of plaintext. The plaintext
+       * size is counted like the record size limit, i.e. per RFC 8449 4.:
+       * "The value includes the content type and padding added in TLS 1.3
+       * (that is, the complete length of TLSInnerPlaintext)."
        *
        * This may be used to reduce the amount of information leaked by the
        * length of TLS records.
        *
-       * The value must be compatible with the negotiated record size limit.
+       * Padding that would grow a record beyond the negotiated record size
+       * limit is truncated to reach exactly that limit.
        *
        * @note This feature is available in TLS 1.3 only (see RFC 9846 5.4).
        *
-       * Default: std::nullopt (no minimum record size is enforced)
+       * Default: 0 (records are not padded)
        */
-      virtual std::optional<uint16_t> minimum_record_size() const;
+      virtual size_t record_padding_bytes(size_t plaintext_bytes) const;
 
       /**
       * Indicates whether certificate status messages should be supported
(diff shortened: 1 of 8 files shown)


src/lib/tls/tls_text_policy.cpp
```diff
@@ -118,10 +118,11 @@ std::optional<uint16_t> Text_Policy::record_size_limit() const {
...
-std::optional<uint16_t> Text_Policy::minimum_record_size() const {
-   const auto limit = get_len("minimum_record_size", 0);
-   BOTAN_ARG_CHECK(limit <= record_size_limit().value_or(16385), "TLS 1.3 minimum record size too large");
-   return (limit > 0) ? std::make_optional(static_cast<uint16_t>(limit)) : std::nullopt;
+size_t Text_Policy::record_padding_bytes(size_t plaintext_bytes) const {
+   // Text policies can express the simplest padding scheme only: pad
+   // records to a fixed minimum size.
+   const auto minimum_record_size = get_len("minimum_record_size", 0);
+   return (plaintext_bytes < minimum_record_size) ? minimum_record_size - plaintext_bytes : 0;
 }
```
reneme, line 126: We could probably come up with some 'policy DSL' for the text policy, like: `pad_to_blocks_of(32)`, `pad_to_at_least(1024)`, or whatever.

Certainly only if requested by someone and not in this PR, of course.

src/lib/tls/tls_policy.h
```diff
@@ -473,20 +473,23 @@ class BOTAN_PUBLIC_API(2, 0) Policy /* NOLINT(*-special-member-functions) */ {
...
+       * Padding that would grow a record beyond the negotiated record size
+       * limit is truncated to reach exactly that limit.
        *
        * @note This feature is available in TLS 1.3 only (see RFC 9846 5.4).
        *
-       * Default: std::nullopt (no minimum record size is enforced)
+       * Default: 0 (records are not padded)
        */
-      virtual std::optional<uint16_t> minimum_record_size() const;
+      virtual size_t record_padding_bytes(size_t plaintext_bytes) const;
```
reneme, line 492: Given that this is a function rather than a static policy now, it could make sense to [make it a `TLS::Callback`](https://github.com/randombit/botan/issues/5777#issuecomment-5165672275) instead. On the other hand there's precedent for such functions also in the policy. I'm indifferent honestly. Just wanted to make sure it was a deliberate decision.

reneme (approved):
> Rather more churn than I expected tbh but overall worth it.

The implementation of `Record_Layer::prepare_records` is a proper mess and pretty hard to reason about. We're looking into simplifying this soon (probably as part of #5841).

src/lib/tls/tls_policy.h
randombit: There is not a super clear line here but generally to me

- `Policy` is specific to the application but usually the same for all connections the application makes
- `Callback` is specific to or relative to the one connection

The line is blurred by the various crypto operations (eg `tls_ephemeral_key_agreement`) which are on Callbacks but typically the same for any one application - though I'm planning on moving all of those to a dedicated struct for Botan4 in any case.

Anyway I think Policy is the correct place for this.

src/lib/tls/tls_text_policy.cpp
randombit: Yeah I'd be fine with that. Not really clear to me how commonly used the text policy support is in any event.

PR #5845, opened 2026-08-14
Add a value barrier in Kyber decompress helper

randombit:
With d = 1, mask becomes 1, and Clang 22 on x86 converts it into a jump.


randombit:
@reneme fyi

reneme:
@falko-strenzke FYI.

PR #5851, opened 2026-08-16
Modify botan3.py so it can adapt to any release of Botan3

randombit:
This turns out to be quite simple since ctypes already lets us query if the function symbol exists or not. If it doesn't, then just synthesize a -40 (not implemented) error which is handled in the normal way.

cc @arckoor 

Inspired by (and much easier than) the equivalent change to the Rust binding in https://github.com/randombit/botan-rs/pull/193

### DIFF
--- a/src/scripts/test_python.py
+++ b/src/scripts/test_python.py
@@ -9,6 +9,8 @@ Botan is released under the Simplified BSD License (see license.txt)
 import argparse
 import binascii
 import copy
+import functools
+import gc
 import hashlib
 import os
 import platform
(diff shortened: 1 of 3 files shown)


src/scripts/test_python.py
```diff
@@ -1067,6 +1142,9 @@ def test_x509_rejects_embedded_nul_strings(self):
             cert.verify(trusted_path="/trusted/roots\0/attacker")
 
     def test_x509_extensions(self):
+        if not library_at_least(3, 13):
+            self.skipTest("The RFC 3779 extension accessors require Botan 3.13")
+
```
arckoor: Why at all and why only for this test? If included it would make sense for tests that also completely rely on a certain version, e.g. `test_crls`, `test_ec_group`, ...

arckoor (approved):
Along with the versioning functions already provided by `botan3`, it might make sense to add something like `unsupported_symbols()` that returns `_FFISymbolResolver.unavailable`, but then a consumer can't really *do* a lot with that, so perhaps also better to omit

randombit:
> Along with the versioning functions already provided by botan3, it might make sense to add something like unsupported_symbols() that returns _FFISymbolResolver.unavailable, but then a consumer can't really do a lot with that

It would always be a subset of the actually non-supported symbols, in that a library might have been configured with a missing feature, so the symbol exists in FFI but just returns not implemented when called. So lacking a clear usage I'd be inclined to skip it.

src/scripts/test_python.py
randombit: Got lazy, will fix

randombit: Changed to run in both versions, checks the result in 3.13+ and otherwise checks that we throw the right exception

PR #5853, opened 2026-08-17
Add RFC 3779 extensions to FFI certificate builder

arckoor:
Stacked on top of #5753. Could go into that PR as well, but together they're kind of big.
The first 3 commits are from #5753, the last commit is the new stuff.


moritzschmitt:
The `BOTAN_FFI_EXPORT(3, 13)` tags need to be `(3, 14)` here too, see my comment on #5753.

PR #5858, opened 2026-08-19
Optimize base58 encoding/decoding, and make constant time

randombit:
Previously while the base58 character classification routines used constant time code, for the radix conversions it used BigInt which leaks.

Additionally optimize various points of the conversion, especially using full radix conversions (58^5 or 58^10) where possible, and using an 8-wide SWAR character classifier. Improves encoding by 5x-12x, and decoding by 2-3x.


randombit:
#5858, nice

Issue #4065, opened 2024-05-20
Compile-time output length


Chaosvex:
Just wanted to open a discussion about HashFunction's `output_length()`. 

One feature that would be nice to have would be to a way to get the output length at compile-time rather than relying on runtime asserts. In cases where you don't want to Botan allocating a vector for the result, it's cumbersome to have to do something like:

```cpp
std::array<std::uint8_t, 20> result;
...
assert(hash.size() == hasher->output_length();
hasher->update(...);
hasher->final(result.data());
```
It'd be nice to be able to just write something akin to:
```cpp
std::array<std::uint8_t, Botan::SHA_160::output_length()> result;
static_assert(Botan::SHA_160::output_length() == result.size()); // *or* this
...
hasher->update(...);
hasher->final(result.data());
```

This would have been trivial to add under the original v1 design but the newer string-based interface rules it out. I saw that there's an [open discussion on providing a better interface
](https://github.com/randombit/botan/issues/3275) but that has a much wider scope and as such, it seems unlikely to go anywhere.

I do have some thoughts on possible solutions but the changes required would feel a little disjointed, but at least the scope would be narrow and wouldn't introduce any breaking changes. However, if a new interface was likely to be on the horizon, it'd be better to hold off.

Thanks!

randombit:
I can absolutely understand why you'd want this. But it may not be really possible for us to provide it. Really we couldn't provide it even in the old interface with a named-class-per-algorithm; it works in your example but not for SHA-3, BLAKE, SHAKE, etc since there there is a variable output length which is parameterized at runtime.

The best approach I can see that might get you some of the way to what you're looking for would be a hypothetical API that I just made up.

```
enum class HashFunctionParams {
  SHA_1,
  SHA_256,
  SHA_3(size_t), // in my imagination C++ also has sum types; we could fake this in various ways
  ...
};

constexpr size_t HashFunctionParams::output_length() {
   switch(*this) {
     case HashFunctionParams::SHA_1: return 20;
     case HashFunctionParams::SHA_256: return 32;
     //...
    }
}

constexpr auto hash_id = HashFunctionParams::SHA_1;
auto hash = HashFunction::create_from_params(hash_id);
hash->update(whatever);
std::array<uint8_t, hash_id.output_length()> digest;
hash->final(digest.data());
```

This would actually be pretty nice and avoid a lot of string handling that we're otherwise stuck with during algorithm creation.

But, this isn't something that's likely to happen very soon.

[...]

PR #5864, opened 2026-08-23
Add PK_Encryption_Options, PK_KEM_Options, and PK_Key_Agreement_Options

randombit:
The design is based on the signature options class and these are all happily much simpler due to a somewhat less profusion of options and modes, so hopefully not too bad to review despite the size.

@reneme also this finally gets you binary strings as OAEP labels

### DIFF
--- a/doc/api_ref/pubkey.rst
+++ b/doc/api_ref/pubkey.rst
@@ -668,6 +668,110 @@ the underlying mathematical properties of the algorithm.  Additionally, they
 will add randomness, so encrypting the same plaintext twice produces two
 different ciphertexts.
 
+The padding scheme and its parameters are specified using a
+:cpp:class:`PK_Encryption_Options` object.
+
+.. _pk_encryption_options:
+
+Encryption Options
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 3.14.0
+
+.. cpp:class:: PK_Encryption_Options
+
+   Describes how a message is to be encrypted or decrypted. As with
+   :cpp:class:`PK_Signature_Options`, the object is built up in a builder
+   style; each ``with_xxx`` function returns a new options object with that
+   option added::
+
+      auto opts = Botan::PK_Encryption_Options()
+                     .with_padding("OAEP")
+                     .with_hash("SHA-256");
+
+      Botan::PK_Encryptor_EME enc(pub_key, rng, opts);
+      Botan::PK_Decryptor_EME dec(priv_key, rng, opts);
+
+   Every option that is set must be understood and acted upon by the scheme in
+   use; if it is not (for example a context with ``PKCS1v15`` padding, or any
+   padding scheme with SM2) then constructing the :cpp:class:`PK_Encryptor_EME`
+   or :cpp:class:`PK_Decryptor_EME` throws an exception naming the offending
+   option(s). An option is never silently ignored. Each option can be set at
+   most once; setting one a second time throws ``Invalid_State``.
+
+   The encryptor and decryptor must be configured identically, since every
+   option affects the ciphertext format.
+
+   .. cpp:function:: PK_Encryption_Options()
+
+      Creates an empty set of options. This is sufficient only for SM2; RSA and
+      ElGamal require a padding scheme.
+
+   .. cpp:function:: PK_Encryption_Options with_padding(std::string_view padding)
+
+      Specify the padding scheme. RSA and ElGamal require this; the available
+      schemes are ``OAEP``, ``PKCS1v15`` and ``Raw``, described in
+      :ref:`eme`. Encrypting without padding is dangerous, so there is no
+      default. SM2 does not use padding and rejects this option.
+
+      An empty string is ignored.
+
+   .. cpp:function:: PK_Encryption_Options with_hash(std::string_view hash)
+
+      Specify the hash function. ``OAEP`` requires this; the hash is used for
+      the label and, unless :cpp:func:`PK_Encryption_Options::with_mgf1_hash`
+      is also given, for the mask generation function. For SM2 it selects the
+      hash used within the scheme, and defaults to SM3 if not given.
+      ``PKCS1v15`` and ``Raw`` padding do not use a hash and reject this
+      option.
+
+      An empty string is ignored.
+
+   .. cpp:function:: PK_Encryption_Options with_mgf1_hash(std::string_view hash)
+
+      Use a different hash function for OAEP's MGF1 than for the label. This
+      is rarely needed, but some protocols (and some implementations, by
+      default) use for example SHA-256 for OAEP with SHA-1 for MGF1. Only
+      ``OAEP`` supports this option.
+
+      An empty string is ignored.
+
+   .. cpp:function:: PK_Encryption_Options with_context(std::span<const uint8_t> context)
+   .. cpp:function:: PK_Encryption_Options with_context(std::string_view context)
+
+      Specify a context which is bound into the ciphertext. Currently this is
+      only supported by ``OAEP``, where it is the label (``L`` in RFC 8017).
+      The label defaults to the empty string. A ciphertext created with one
+      label will not decrypt with another.
+
+   .. cpp:function:: PK_Encryption_Options with_provider(std::string_view provider)
+
+      Request a specific implementation. This is rarely needed; the main use is
+      with keys held in hardware (providers ``pkcs11``, ``tpm2``), which accept
+      only their own provider name. The default (an empty string) selects
+      whatever implementation the key has; ``base`` names the software
+      implementation explicitly, and is therefore rejected for hardware keys.
+
+      The hardware providers support ``OAEP`` and ``PKCS1v15`` (PKCS#11 also
+      ``Raw``). The TPM2 provider does not support an OAEP label, nor an MGF1
+      hash differing from the OAEP hash.
+
+   .. cpp:function:: std::string to_string() const
+
+      Formats the options as a string, for debugging and error messages. The
+      format is not fixed.
+
+Encryption Options Accepted by Each Scheme
+""""""""""""""""""""""""""""""""""""""""""""""""
+
+- **RSA** and **ElGamal**: require a padding scheme. With ``OAEP`` a hash is
+  also required, and an MGF1 hash and a context (the label) are accepted.
+  ``PKCS1v15`` and ``Raw`` take no further options.
+- **SM2**: accepts a hash (default SM3) and nothing else.
+
+Encrypting and Decrypting
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
 The primary interface for encryption is
 
 .. cpp:class:: PK_Encryptor
@@ -706,18 +810,32 @@ support it directly, such as RSA or ElGamal; these use the EME class:
 
    .. cpp:function:: PK_Encryptor_EME(const Public_Key& key, \
          RandomNumberGenerator& rng, \
+         const PK_Encryption_Options& options)
+
+     With *key* being the key you want to encrypt messages to, configured as
+     described by *options* (see :ref:`pk_encryption_options`).
+
+     If you are not sure what padding to use, use
+     ``PK_Encryption_Options().with_padding("OAEP").with_hash("SHA-256")``.
+     If you need compatibility with protocols using the PKCS #1 v1.5 standard,
+     you can also use ``with_padding("PKCS1v15")``.
+
+     For SM2 encryption, only a hash may be specified; normally it is left at
+     the default of SM3.
+
+   .. cpp:function:: PK_Encryptor_EME(const Public_Key& key, \
+         RandomNumberGenerator& rng, \
          std::string_view padding, \
          std::string_view provider = "")
 
-     With *key* being the key you want to encrypt messages to. The padding
-     method to use is specified in *padding*.
+     The padding method to use is specified by the string *padding*, which is
+     the interface available in previous versions of Botan. The string is
+     translated into the equivalent :cpp:class:`PK_Encryption_Options`; new
+     code should prefer to construct the options directly.
 
-     If you are not sure what padding to use, use "OAEP(SHA-256)". If you need
-     compatibility with protocols using the PKCS #1 v1.5 standard, you can also
-     use "PKCS1v15".
-
-     For SM2 encryption, the padding string specifies which hash function to
-     use; normally this would be "SM3".
+     For RSA and ElGamal the string names a :ref:`padding scheme <eme>`, for
+     example "OAEP(SHA-256)" or "PKCS1v15". For SM2 encryption, the padding
+     string specifies which hash function to use; normally this would be "SM3".
 
 .. cpp:class:: DLIES_Encryptor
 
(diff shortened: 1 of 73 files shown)


doc/api_ref/pubkey.rst
```diff
@@ -668,6 +668,110 @@ the underlying mathematical properties of the algorithm.  Additionally, they
...
+.. cpp:class:: PK_Encryption_Options
+
+   Describes how a message is to be encrypted or decrypted. As with
+   :cpp:class:`PK_Signature_Options`, the object is built up in a builder
+   style; each ``with_xxx`` function returns a new options object with that
+   option added::
+
+      auto opts = Botan::PK_Encryption_Options()
+                     .with_padding("OAEP")
+                     .with_hash("SHA-256");
```
reneme, line 690: Given that this API philosophy is only starting to unfold in the library, I'll take the liberty to throw around more crazy ideas. 

A while ago @falko-strenzke had suggested (#3275) to move away from configuration strings and the the options builders are definitely a big step in that direction. But at this point the individual builder parameters are still "just strings" and they start to have hard-to-understand inter-dependencies (e.g. `with_mgf1_hash` which only makes sense if `with_padding` is "OAEP").

I did [some prototyping](https://godbolt.org/z/aEc5cj99r), and a type-safe API like the following seems entirely achievable with GCC 11:

```C++
using u8 = uint8_t;

using namespace Botan::Hashes;
using namespace Botan::Paddings;

auto opts_default =
    Botan::Enc_Opts()
        .with_padding(OAEP); // default params (Note: no braces here!)

auto opts_with_sha1 =
    Botan::Enc_Opts()
        .with_padding(OAEP(SHA_256, SHA_1)); // OAEP with custom hash and MGF hash

auto oaep_label = {u8{42}, u8{0}, u8{1}};
auto oaep_with_label = OAEP(SHA_256, SHA_1, oaep_label);

auto opts_with_label =
    Botan::Enc_Opts()
        .with_padding(oaep_with_label)
        .with_hash(SHAKE_128(32)); // Forgetting the output length would be a compile-time error
```

The types behind those algorithm specifiers would also be useful to expose compile-time information such as hash output lengths, cipher block sizes or asymmetric key lengths (#3706 #4065).

No hard feeling if this leads to a strict "no" at this point, @randombit. 😶 Something like this ~~could~~ should be introduced independently of the builders. But the API decisions have implications onto each other, hence this suggestion.

src/lib/pubkey/pubkey.cpp
```diff
@@ -94,11 +94,22 @@ secure_vector<uint8_t> PK_Decryptor::decrypt_or_random(const uint8_t in[],
...
-   m_op = key.create_encryption_op(rng, padding, provider);
+                                   std::string_view provider) :
+      PK_Encryptor_EME(key, rng, parse_legacy_enc_options(key, padding).with_provider(provider)) {}
+
+PK_Encryptor_EME::PK_Encryptor_EME(const Public_Key& key,
+                                   RandomNumberGenerator& rng,
+                                   const PK_Encryption_Options& user_options) {
+   // Track option usage on a private copy, so the caller's object is never modified
+   const PK_Encryption_Options options(user_options);  // NOLINT(*-unnecessary-copy-initialization) clang-tidy bug
+   options.reset_examined();
```
reneme, line 105: Given that the options builders are copyable anyway, why don't we just accept them by value? That would clearly communicate to the caller that they don't need to worry about side effects on their copy of the options objects. Also it avoids an additional copy when the options are set up in-place.

That will also avoid the clang-tidy warning as we're calling non-const members (`reset_examined()`) on the by-value parameter, I think.

```suggestion
PK_Encryptor_EME::PK_Encryptor_EME(const Public_Key& key,
                                   RandomNumberGenerator& rng,
                                   PK_Encryption_Options options) {
   options.reset_examined();
```

(Likely applies to other PK_Operations as well)

src/lib/pk_pad/enc_padding/eme_oaep/oaep.cpp
```diff
@@ -146,6 +147,24 @@ size_t OAEP::maximum_input_size(size_t keybits) const {
...
+   auto hash = HashFunction::create_or_throw(options.hash_function_name());
+
+   if(options.using_mgf1_hash()) {
+      m_mgf1_hash = HashFunction::create_or_throw(options.mgf1_hash_function().value());
+   } else {
+      m_mgf1_hash = hash->new_object();
+   }
+
+   // The label is optional and defaults to empty
+   const std::vector<uint8_t> label = options.context().value_or(std::vector<uint8_t>());
```
reneme, line 164: Nit: 
```suggestion
   const auto label = options.context().value_or({});
```

... who cares about types these days. 😶

src/lib/pk_pad/enc_padding/enc_padding.cpp
```diff
@@ -24,45 +23,36 @@
...
 
-std::unique_ptr<EncryptionPaddingScheme> EncryptionPaddingScheme::create(std::string_view algo_spec) {
+std::unique_ptr<EncryptionPaddingScheme> EncryptionPaddingScheme::create_or_throw(
+   const PK_Encryption_Options& options) {
+   // Encrypting without padding is dangerous, so it must be requested explicitly
+   if(!options.using_padding()) {
+      throw Lookup_Error("Public key encryption requires specifying a padding scheme");
+   }
+
+   const std::string padding = options.padding().value();
```
reneme, line 33: No need to copy this once more:


```suggestion
   const std::string& padding = options.padding().value();
```

src/lib/prov/pkcs11/p11_mechanism.cpp
```diff
@@ -88,73 +88,88 @@ class RSA_SignMechanism final : public MechanismData {
...
-      size_t padding_size() const { return m_padding_size; }
+MechanismWrapper MechanismWrapper::create_rsa_crypt_mechanism(const PK_Encryption_Options& options) {
+   if(!options.using_padding()) {
+      throw Lookup_Error("PKCS#11 RSA encrypt/decrypt requires specifying a padding scheme");
+   }
 
-   private:
-      // mechanism ID of the message digest algorithm used to calculate the digest of the encoding parameter
-      MechanismType m_hash;
+   const std::string padding = options.padding().value();
```
reneme, line 101:
```suggestion
   const std::string& padding = options.padding().value();
```

doc/api_ref/pubkey.rst
randombit: The approach just doesn't feel like a good fit to me. The final validity is anyway not determined until runtime (for example ML-DSA + PSS is nonsensical no matter what params you use), we have to expose lots of / all of the details of the various padding schemes/hashes/etc in the type system [something I have tried hard to get away with over time, for example hiding `aes.h` and co so all accesses are via a vtable - I would tbh like to do the same for RSA and co though I am unsure if that's viable in Botan4 timeframe], and this design seems to me Not Fun to expose in FFI, while the current builder style maps quite naturally to a C interface. It would be a bit of a different story with eg a Crypto++ style interface where the scheme/padding/hash are all effectively fixed as template parameters at compile time - but I was aware of and disliked Crypto++'s API philosophy sufficiently to motivate me to start this project in the first place so...

reneme: > The final validity is anyway not determined until runtime [...], we have to expose lots of / all of the details of the various padding schemes/hashes/etc in the type system [...]

Yeah, I agree. Without moving substantial information and validation into the public type system this doesn't provide big benefits apart from API cleverness. And for that it's really not worth it. FFI is certainly another strike against it.

Thanks for considering, anyway. :+1:

src/lib/prov/tpm2/tpm2_rsa/tpm2_rsa.cpp
```diff
@@ -262,7 +263,31 @@ class RSA_Verification_Operation final : public Verification_Operation {
...
+      throw Lookup_Error("TPM2 RSA encryption requires specifying a padding scheme");
+   }
+
+   std::string padding = options.padding().value();
+
+   if(padding == "OAEP") {
+      if(!options.using_hash()) {
+         throw Lookup_Error("TPM2 RSA OAEP requires specifying a hash function");
+      }
+      const std::string hash_name = options.hash_function_name();
```
reneme, line 277:
```suggestion
      const std::string& hash_name = options.hash_function_name();
```

src/lib/pubkey/classic_mceliece/cmce.cpp
```diff
@@ -136,13 +136,13 @@ bool Classic_McEliece_PrivateKey::check_key(RandomNumberGenerator& /*rng*/, bool
...
+std::unique_ptr<PK_Ops::KEM_Decryption> Classic_McEliece_PrivateKey::_create_kem_decryption_op(
+   RandomNumberGenerator& rng, const PK_KEM_Options& options) const {
    BOTAN_UNUSED(rng);
-   if(provider.empty() || provider == "base") {
-      return std::make_unique<Classic_McEliece_Decryptor>(this->m_private, params);
+   if(!options.using_provider()) {
+      return std::make_unique<Classic_McEliece_Decryptor>(this->m_private, options);
    }
-   throw Provider_Not_Found(algo_name(), provider);
+   throw Provider_Not_Found(algo_name(), options.provider().value());
```
reneme, line 145: This validation seems to repeat across essentially all software-based algorithms. Would it be feasible to introduce a `require_software_provider()` similar to the hardware validation?

src/lib/pubkey/pk_ops_impl.h
```diff
@@ -22,8 +22,24 @@ class EncryptionPaddingScheme;
 
 namespace Botan::PK_Ops {
 
+/**
+* Describes the raw shared key produced by a KEM, ie the value returned when
+* no KDF is applied to it.
+*/
+enum class RawKemSharedKey : uint8_t {
```
reneme, line 29: Naming nit: maybe `KemSharedKeyQuality`?

src/lib/pubkey/pk_options_impl.cpp
```diff
@@ -243,6 +244,81 @@ PK_Signature_Options parse_legacy_sig_options(const Public_Key& key, std::string
...
+   // TODO(Botan4) Remove all but "OAEP"
+   if(req.algo_name() == "OAEP" || req.algo_name() == "EME-OAEP" || req.algo_name() == "EME1") {
+      if(req.arg_count_between(1, 3)) {
+         auto options = PK_Encryption_Options().with_padding("OAEP").with_hash(req.arg(0));
+
+         if(req.arg_count() >= 2 && req.arg(1) != "MGF1") {
+            const auto mgf_params = parse_algorithm_name(req.arg(1));
+            if(mgf_params.size() != 2 || mgf_params[0] != "MGF1") {
+               throw Lookup_Error(fmt("Unknown OAEP mask generation function {}", req.arg(1)));
+            }
```
reneme, line 284:
```suggestion
            if(mgf_params.size() != 2 || mgf_params[0] != "MGF1" || mgf_params[1].empty()) {
               throw Lookup_Error(fmt("Unknown OAEP mask generation function {}", req.arg(1)));
            }
```

... that catches `MGF1()` which otherwise would be accepted. I believe nothing bad would happen but perhaps still worth catching. Courtesy Kimi K3.

src/lib/pubkey/pk_options_impl.cpp
```diff
@@ -243,6 +244,81 @@ PK_Signature_Options parse_legacy_sig_options(const Public_Key& key, std::string
...
+   * OAEP(hash), OAEP(hash,MGF1), OAEP(hash,MGF1,label), OAEP(hash,MGF1(mgf_hash)),
+   * OAEP(hash,MGF1(mgf_hash),label) (or the aliases EME-OAEP and EME1)
+   */
+
+   // TODO(Botan4) Remove all but "PKCS1v15"
+   if(params == "PKCS1v15" || params == "EME-PKCS1-v1_5") {
+      return PK_Encryption_Options().with_padding("PKCS1v15");
+   }
+
+   const SCAN_Name req(params);
```
reneme, line 273: An empty `params` currently goes into `SCAN_Name` here, resulting in an `Invalid_Argument` exception that isn't a `Lookup_Error`. Not critical, I guess. Still worth mentioning. Kimi K3 found this.

Maybe:


```suggestion
   if(params.empty()) {
      return PK_Encryption_Options(); // algorithms will reject this, if they require a padding
   }

   const SCAN_Name req(params);
```

src/lib/pubkey/pk_options_impl.cpp
```diff
@@ -243,6 +244,81 @@ PK_Signature_Options parse_legacy_sig_options(const Public_Key& key, std::string
...
+PK_Key_Agreement_Options parse_legacy_ka_options(std::string_view params) {
+   // The string interface has always required naming a KDF (or "Raw")
+   if(params.empty()) {
+      throw Invalid_Argument("Key agreement requires specifying a KDF, or Raw to use the agreed value directly");
+   } else if(params == "Raw") {
+      return PK_Key_Agreement_Options().with_raw_shared_key();
+   } else {
+      return PK_Key_Agreement_Options().with_kdf(params);
+   }
+}
```
reneme, line 320: Kimi K3 claims that this prevents users from specifying ECDH PKCS#11 "Raw,Cofactor" as the KDF which used to work in the old API.

`params="Raw,Cofactor"` would currently be passed into `with_kdf()` and that would be rejected by the consumer, because it rejects all KDFs starting with "Raw". Perhaps that's intentional, I'm not sure about this API surface.

src/lib/pubkey/pk_options.h
```diff
@@ -303,6 +308,348 @@ class BOTAN_PUBLIC_API(3, 14) PK_Signature_Options final {
...
+      ///
+      /// This will throw an exception if no hash function was configured
+      std::string hash_function_name() const;
+
+      /*
+      * Getters; these are mostly for internal use
+      *
+      * Calling any of these records that the respective option was examined
+      * by the scheme, so a scheme should only read an option it will act on.
+      */
```
reneme, line 388: I think we should find a way to hide these accessors from the public API. This comment will easily go unnoticed by users and if they start relying on those getters, the usage check in `note_examined` breaks. Also, we wouldn't be able to fix this after the release, because it is then essentially ossified API.

Two ideas:

1. Just use the "public but stay away" underscore that we introduced into the API contract some time ago and rename these getters to `_padding()`, `_hash_function()` and so on.
2. Set up internal `PK_xyz_Options_Reader` classes and befriend them in the public API. Users can't access them and harmful API misuse is essentially impossible.

My preference would be idea (2), despite the added boilerplate. It has the added benefit that the entire `note_examined()` machinery could move out of the public API and would be easier to modify that way, e.g. with C++26 and compile-time reflection we might be able to get rid of most boilerplate.

src/lib/pubkey/pk_options.h
```diff
@@ -303,6 +308,348 @@ class BOTAN_PUBLIC_API(3, 14) PK_Signature_Options final {
...
+      ///
+      /// This is primarily intended for debugging and error messages;
+      /// the format is not fixed
+      std::string to_string() const;
+
+      /// Specify the padding scheme
+      ///
+      /// For RSA (and ElGamal) this is required, and is one of "OAEP", "PKCS1v15"
+      /// or "Raw". Other schemes do not use padding.
+      PK_Encryption_Options with_padding(std::string_view padding);
```
reneme, line 347: Kimi K3 came up with an interesting suggestion regarding general API critique:

> Raw RSA and missing KDFs got extra friction. PKCS#1 v1.5 encryption is the padding with a live oracle history, and with_padding("PKCS1v15") is as easy as OAEP. A name like with_pkcs1v15_encryption() plus a warning in the type’s doc (you already warn in the EME section) would match the “make the dangerous thing obvious” bar you set for raw secrets. Not a blocker; it is a missed chance.

... I think it has a point, but it's also (by design) inconsistent API. I'm undecided, but I found it worth discussing.

reneme (approved):
> @reneme also this finally gets you binary strings as OAEP labels

### 🥳 

Thanks for this effort! Really nice to see that this makes its way through the API.


Below are some nits, some suggestions, and a few findings (partially by a ~~Kimi K3~~ Cursor Grok 4.6 review).

src/lib/pubkey/pubkey.cpp
randombit: Yeah makes sense considering the mandatory copying happening here

src/lib/pubkey/pk_options_impl.cpp
randombit: Makes sense

src/lib/pubkey/pk_options.h
randombit: I'll try an options reader type, if it's tolerable that's cleanest in terms of API. If not at least underscore indicates intent a bit better.

src/lib/pubkey/pk_options.h
randombit: I think to me the critical difference is doing something sketchy/possibly insecure by _omission_ vs _an explicit request for that thing_. Eg

ECDH requires an explicit `with_kdf` or `with_raw_shared_key`. If the default (no `with_kdf`) was the raw point, we silently end up in the less secure state, so instead we require an explicit opt-in for it.

If RSA silently defaulted (with no other `with_padding` specified) to Raw _or_ PKCS1v1.5, that would of course be awful.

The point [re raw points, etc] is not extra friction (imo) it's just that you have to opt into it explicitly and visibly.

src/lib/pubkey/pk_options.h
randombit: Realized we can't make the options reader types internal because we do support externally defined key types (ex `src/examples/hybrid_key_encapsulation.cpp`). But I put the reader types in a distinct (public) header and marked them as unstable.

src/lib/pubkey/pk_options_impl.cpp
randombit: Fixed

src/lib/pubkey/pk_options.h
randombit: A lot of churn but not too much additional boilerplate, probably a win given it minimizes the visible API surface

src/lib/pubkey/pubkey.cpp
randombit: Obviated by the addition of the reader type

PR #5875, opened 2026-08-25
fix implib output directory for gcc/mingw

dbear496:
The `--out-implib` linker option needs to include the output path for the import library; otherwise, the import library always goes to the current directory. This is important for the installer, which expects the import library to be in the build directory.


randombit:
Has this always behaved that way? I'm looking at #2804 which originated this and it doesn't have anything related; I did consider that perhaps at that time our generated paths were absolute and this had regressed at some point, but that doesn't seem to be the case. I'm just wondering if this requirement re the path has *always* been broken or perhaps some change in MinGW that we should adjust for or ???

dbear496:
It's possible that the `--with-build-dir` option wasn't used (or set to `.`) when testing the installer under mingw.

randombit (approved):

Issue #5719, opened 2026-07-13
[TLS 1.3] Check the Implementation against RFC 9846


reneme:
[RFC 9846](https://www.rfc-editor.org/info/rfc9846) is an updated version of RFC 8446 that tightens some requirements and clarifies some details. The new revision is fully backward-compatible with the old one, though. Check if we're compliant and fix things as needed. Brownie-points for updating all RFC references in the code (as an independent PR).

PR #5877, opened 2026-08-26
In TLS 1.3 send a KeyUpdate when traffic approaches the AEAD limit

randombit:
This was a SHOULD in RFC 8446 and is a MUST in RFC 9846

#5719

### DIFF
--- a/src/lib/tls/tls13/tls_channel_impl_13.cpp
+++ b/src/lib/tls/tls13/tls_channel_impl_13.cpp
@@ -16,9 +16,6 @@
 #include <botan/internal/concat_util.h>
 #include <botan/internal/tls_cipher_state.h>
 
-#include <array>
-#include <chrono>
-
 namespace {
 bool is_user_canceled_alert(const Botan::TLS::Alert& alert) {
    return alert.type() == Botan::TLS::Alert::UserCanceled;
@@ -54,6 +51,7 @@ Channel_Impl_13::Channel_Impl_13(const std::shared_ptr<Callbacks>& callbacks,
       m_can_read(true),
       m_can_write(true),
       m_opportunistic_key_update(false),
+      m_key_update_requested(false),
       m_first_message_sent(false),
       m_first_message_received(false) {
    BOTAN_ASSERT_NONNULL(m_callbacks);
@@ -212,10 +210,15 @@ void Channel_Impl_13::handle(const Key_Update& key_update) {
       throw Unexpected_Message("Unexpected additional post-handshake message data found in record");
    }
 
-   if(const uint64_t min_interval = policy().minimum_key_update_interval_ms(); min_interval > 0) {
-      const uint64_t now =
-         std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now().time_since_epoch())
-            .count();
+   // A non-requesting KeyUpdate received while our own request is outstanding
+   // is the reciprocation we solicited. It is exempt from rate limiting (and
+   // invisible to it), so that a peer whose own key update crossed ours in
+   // flight is not penalized for the resulting back to back KeyUpdates.
+   const bool solicited_reciprocation = m_key_update_requested && !key_update.expects_reciprocation();
+
+   if(const uint64_t min_interval = policy().minimum_key_update_interval_ms();
+      min_interval > 0 && !solicited_reciprocation) {
+      const uint64_t now = callbacks().tls_current_monotonic_clock_ms();
 
       if(m_last_key_update_ms != 0 && (now - m_last_key_update_ms) < min_interval) {
          throw TLS_Exception(Alert::UnexpectedMessage, "Peer is requesting KeyUpdates too frequently");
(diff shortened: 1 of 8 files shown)


src/lib/tls/tls13/tls_channel_impl_13.cpp
```diff
@@ -298,6 +315,9 @@ void Channel_Impl_13::to_peer(std::span<const uint8_t> data) {
    if(m_opportunistic_key_update) {
       update_traffic_keys(false /* update_requested */);
       m_opportunistic_key_update = false;
+   } else if(needs_traffic_based_key_update()) {
+      // If approaching traffic limits request the peer update their own keys as well
+      update_traffic_keys(true /* update_requested */);
```
reneme: It is a rare event anyway, given the amount of traffic until this typically gets triggered, so I'm not opposing the requested reciprocation request at all. However, considering connections that mostly send but rarely receive, I think, it would be a sound design when this opportunistic key update was triggered independently. Or does the RFC demand it?

reneme (approved):

src/lib/tls/tls13/tls_channel_impl_13.cpp
randombit: I find [Section 5.5](https://datatracker.ietf.org/doc/html/rfc9846#section-5.5) quite inprecise but my interpretation of it is that writers should update their keys prior to sending close to the limits, but should never enforce any limit when reading ("Receiving implementations SHOULD NOT enforce these limits"). So a strict reading aiui would have this be `update_request` false, but also *not* have any limit on the reader side.

The main reason to request the peer update as well is that (prior to this patch) Botan would not on its own accord initiate a key refresh. Per [this report](https://cybernuke.bensmyth.com/findings/FOREVER-tls13-no-keyupdate-forward-secrect) many other implementations are in a similar situation, where they will reply to a KeyUpdate but not initiate one without an application call. Since KeyUpdates are in the end pretty cheap (certainly so relative to encrypting 2**23 records) seemed to me better to just roll both directions even if not strictly required. That does leave a hole where we are on connection where we are the read-heavy side and the peer doesn't want to rollover, but that is status quo behavior and per the RFC at least not something we should be checking.

src/lib/tls/tls13/tls_channel_impl_13.cpp
```diff
@@ -287,6 +287,23 @@ void Channel_Impl_13::to_peer(std::span<const uint8_t> data) {
       throw Invalid_State("Data cannot be sent on inactive TLS connection");
    }
 
+   // RFC 9846 5.5
+   //    Implementations MUST either close the connection or do a key update as
+   //    described in Section 4.7.3 prior to reaching these limits.
+   //
+   // [This is a SHOULD in RFC 8446]
```
reneme: Technically, ChaCha-based connection wouldn't need to roll-over. Perhaps it's worth mentioning that we're wrapping the key anyway.

src/lib/tls/tls13/tls_channel_impl_13.cpp
reneme: Agreed, in general.

> That does leave a hole where we are on connection where we are the read-heavy side and the peer doesn't want to rollover [...r

I'd read `"Receiving implementations SHOULD NOT enforce these limits"` as: don't hard-close the connection. Therefore, I think sending a reciprocal KeyUpdate once we read 2**23 records would be acceptable, no? That would force an non-compliant peer into updating the keys when we're on the read-heavy end.

reneme (approved):

src/lib/tls/tls13/tls_channel_impl_13.cpp
randombit: Yeah true. Added it so that we also will send a KeyUpdate if we have read 1.5* the limit. That mult is just to avoid a case where two peers with the same limit will cross their KeyUpdates on the wire, which is mostly harmless (except that it would trigger the KeyUpdate rate limiter by default - also changed this so the rate limiter does not trigger on KeyUpdates which are in response to KeyUpdates we sent). This all is still only triggered on a write, so a protocol that actually literally *just* reads with literally zero writes will still not see a rotation unless the peer itself knows to rotate (or if the application requests one explicitly).

src/lib/tls/tls13/tls_channel_impl_13.cpp
randombit: Added a note here

src/lib/tls/tls13/tls_channel_impl_13.cpp
reneme: > That mult is just to avoid a case where two peers with the same limit will cross their KeyUpdates on the wire, which is mostly harmless.

FYI: in DTLS 1.3 the KeyUpdate must be ACKed and it is forbidden to send another KeyUpdate until that ACK arrived from the peer. We will need to handle this regardless, but avoiding a potential race condition bug might not be a bad idea additionally. So: I think this design makes a lot of sense.

reneme (approved):
Looks good to me now, thanks. The addition of DTLS will probably require some work on the KeyUpdate state machine, given that it already maintains two dedicated bool flags and DTLS introduces mandatory ACKs to the mix.... one day at a time.

## 😵‍💫

PR #5878, opened 2026-08-26
Use Hash_Engine batching in ML-KEM

randombit:
Improvement for one-shot ML-KEM (ie where the KEM encryptor object is created once per operation) which matches the common usage in practice of using ML-KEM as an ephemeral key

```
+ 96.0% improvement in ML-KEM-1024 KEM encrypt
+ 73.5% improvement in ML-KEM-1024 KEM decrypt
+ 72.6% improvement in ML-KEM-1024 keygen
+ 56.7% improvement in ML-KEM-768 KEM encrypt
+ 55.0% improvement in ML-KEM-768 keygen
+ 43.3% improvement in ML-KEM-768 KEM decrypt
+ 30.3% improvement in ML-KEM-512 keygen
+ 30.1% improvement in ML-KEM-512 KEM encrypt
+ 21.6% improvement in ML-KEM-512 KEM decrypt
```

### DIFF
--- a/src/lib/pubkey/kyber/kyber_common/kyber_algos.cpp
+++ b/src/lib/pubkey/kyber/kyber_common/kyber_algos.cpp
@@ -114,6 +114,52 @@ void sample_ntt_uniform(KyberPolyNTT& p, XOF& xof) {
 }
 
 /**
+ * As above, but sampling from a pre-squeezed XOF output buffer. Returns
+ * false if the buffer did not contain enough accepted candidates.
+ */
+bool sample_ntt_uniform_from_buffer(KyberPolyNTT& p, std::span<const uint8_t> buf) {
+   size_t pos = 0;
+   std::optional<uint16_t> stashed_coeff;
+
+   auto lowerthan_q = [](uint32_t d) -> std::optional<uint16_t> {
+      if(d < KyberConstants::Q) {
+         return static_cast<uint16_t>(d);
+      } else {
+         return std::nullopt;
+      }
+   };
+
+   for(auto& coeff : p) {
+      if(auto stashed = std::exchange(stashed_coeff, std::nullopt)) {
+         coeff = *stashed;
+         continue;
+      }
+
+      while(true) {
+         if(pos + 3 > buf.size()) {
+            return false;
+         }
+         const auto x = load_le3(buf.subspan(pos).first<3>());
+         pos += 3;
+         const auto d1 = lowerthan_q(x & 0x0FFF);
+         const auto d2 = lowerthan_q(x >> 12);
+
+         if(d1.has_value()) {
+            stashed_coeff = d2;  // keep candidate d2 for the next coefficient
+            coeff = *d1;
+            break;
+         } else if(d2.has_value()) {
+            // d1 was invalid, d2 is valid, nothing to stash
+            coeff = *d2;
+            break;
+         }
+      }
+   }
+
+   return true;
+}
+
+/**
  * NIST FIPS 203, Algorithm 8 (SamplePolyCBD)
  *
  * Implementations for eta = 2 and eta = 3 are provided separately as template
@@ -334,8 +380,9 @@ KyberInternalKeypair expand_keypair(KyberPrivateKeySeed seed, KyberConstants mod
 
    // The nonce N is handled internally by the PolynomialSampler
    Kyber_Algos::PolynomialSampler ps(sigma, mode);
-   auto s = ntt(ps.sample_polynomial_vector_cbd_eta1());
-   const auto e = ntt(ps.sample_polynomial_vector_cbd_eta1());
+   auto [s_std, e_std] = ps.sample_polynomial_vector_pair_cbd_eta1();
+   auto s = ntt(std::move(s_std));
+   const auto e = ntt(std::move(e_std));
 
    auto t = montgomery(A * s);
    t += e;
@@ -390,13 +437,35 @@ KyberPolyMat sample_matrix(StrongSpan<const KyberSeedRho> seed, bool transposed,
    KyberPolyMat mat(mode.k(), mode.k());
 
    const auto& sym = mode.symmetric_primitives();
-   std::unique_ptr<Botan::XOF> xof;
+
+   // Pre-squeeze the XOFs of all matrix entries at once, sized so that
+   // the rejection sampling loop rarely (under 1% of the time) needs more.
+   constexpr size_t buffered_bytes = 3 * 168;  // 3 SHAKE-128 blocks
+
+   const size_t lanes = static_cast<size_t>(mode.k()) * mode.k();
+   std::vector<uint8_t> buffer(lanes * buffered_bytes);
+   std::vector<std::tuple<uint8_t, uint8_t>> positions;
+   std::vector<std::span<uint8_t>> outputs;
+   positions.reserve(lanes);
+   outputs.reserve(lanes);
 
    for(uint8_t i = 0; i < mode.k(); ++i) {
       for(uint8_t j = 0; j < mode.k(); ++j) {
-         const auto pos = (transposed) ? std::tuple(i, j) : std::tuple(j, i);
-         sym.setup_XOF(xof, seed, pos);
-         sample_ntt_uniform(mat[i][j], *xof);
+         positions.push_back((transposed) ? std::tuple(i, j) : std::tuple(j, i));
+         outputs.push_back(std::span(buffer).subspan(outputs.size() * buffered_bytes, buffered_bytes));
+      }
+   }
+
+   sym.XOF_batch(outputs, seed, positions);
+
+   std::unique_ptr<Botan::XOF> xof;
+   size_t lane = 0;
+   for(uint8_t i = 0; i < mode.k(); ++i) {
+      for(uint8_t j = 0; j < mode.k(); ++j, ++lane) {
+         if(!sample_ntt_uniform_from_buffer(mat[i][j], outputs[lane])) {
+            sym.setup_XOF(xof, seed, positions[lane]);
+            sample_ntt_uniform(mat[i][j], *xof);
+         }
       }
    }
 
@@ -411,7 +480,7 @@ KyberPolyMat sample_matrix(StrongSpan<const KyberSeedRho> seed, bool transposed,
  */
 void sample_polynomial_from_cbd(KyberPoly& poly,
                                 KyberConstants::KyberEta eta,
-                                const KyberSamplingRandomness& randomness) {
+                                StrongSpan<const KyberSamplingRandomness> randomness) {
    switch(eta) {
       case KyberConstants::KyberEta::_2:
          return sample_poly_cbd<KyberConstants::KyberEta::_2>(poly, randomness);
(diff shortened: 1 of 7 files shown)


src/lib/pubkey/kyber/kyber_common/kyber_algos.cpp
```diff
@@ -113,6 +113,52 @@ void sample_ntt_uniform(KyberPolyNTT& p, XOF& xof) {
...
+         } else if(d2.has_value()) {
+            // d1 was invalid, d2 is valid, nothing to stash
+            coeff = *d2;
+            break;
+         }
+      }
+   }
+
+   return true;
+}
```
reneme, line 160: Perhaps somewhat a matter of taste, but by setting this up more similarly to the XOF-based implementation, you could avoid essentially all state modifications (except within the `sample()` generator).

And additionally: 

* the sampling `while` loop is bounded now
* the `pos` variable is gone, replaced by re-aligning the input span
* early-return on "not enough bytes" has more obvious control flow

```suggestion
/**
 * As above, but sampling from a pre-squeezed XOF output buffer. Returns
 * false if the buffer did not contain enough accepted candidates.
 */
bool sample_ntt_uniform_from_buffer(KyberPolyNTT& p, std::span<const uint8_t> buf) {
   auto sample = [in = buf, stashed_coeff = std::optional<uint16_t>{}]() mutable -> std::optional<uint16_t> {
      auto lowerthan_q = [](uint32_t d) -> std::optional<uint16_t> {
         if(d < KyberConstants::Q) {
            return static_cast<uint16_t>(d);
         } else {
            return std::nullopt;
         }
      };

      if(auto stashed = std::exchange(stashed_coeff, std::nullopt)) {
         return stashed;
      }

      while(in.size() >= 3) {
         const auto x = load_le3(in.first<3>());
         in = in.subspan(3);

         const auto d1 = lowerthan_q(x & 0x0FFF);
         const auto d2 = lowerthan_q(x >> 12);

         if(d1.has_value()) {
            stashed_coeff = d2;  // keep candidate d2 for the next invocation
            return d1;
         } else if(d2.has_value()) {
            // d1 was invalid, d2 is valid, nothing to stash
            return d2;
         }
      }

      return std::nullopt;  // not enough candidates left in the buffer
   };

   for(auto& coeff : p) {
      if(auto c = sample()) {
         coeff = *c;
      } else {
         return false;  // buffer exhausted before all coefficients could be sampled
      }
   }

   return true;
}
```

src/lib/pubkey/kyber/kyber_common/kyber_algos.h
```diff
@@ -66,58 +68,81 @@ class PolynomialSampler {
...
+      void sample_polys_cbd(std::span<KyberPoly* const> polys, KyberConstants::KyberEta eta) {
+         BOTAN_ASSERT_NOMSG(!polys.empty());
+         const size_t poly_bytes = cbd_randomness_bytes(*polys[0], eta);
 
-            BOTAN_ASSERT_UNREACHABLE();
-         }();
+         KyberSamplingRandomness randomness(polys.size() * poly_bytes);
+         std::vector<std::span<uint8_t>> outputs(polys.size());
+         for(size_t i = 0; i < polys.size(); ++i) {
+            outputs[i] = std::span(randomness.get()).subspan(i * poly_bytes, poly_bytes);
```
reneme, line 130: Nit: No need to `.get()` the inner type of the strong type. The span works just fine with the strong type wrapper as well.


```suggestion
            outputs[i] = std::span(randomness).subspan(i * poly_bytes, poly_bytes);
```

src/lib/pubkey/kyber/kyber_common/kyber_algos.h
```diff
@@ -66,58 +68,81 @@ class PolynomialSampler {
...
+         }
 
-         sample_polynomial_from_cbd(poly, eta, randomness);
+         const auto& sym = m_mode.symmetric_primitives();
+         sym.PRF_batch(outputs, m_seed.get(), m_nonce);
```
reneme, line 134: Nit: as above:


```suggestion
         sym.PRF_batch(outputs, m_seed, m_nonce);
```

src/lib/pubkey/kyber/kyber_common/kyber_algos.h
```diff
@@ -66,58 +68,81 @@ class PolynomialSampler {
...
+         m_nonce += static_cast<uint8_t>(polys.size());
+
+         for(size_t i = 0; i < polys.size(); ++i) {
+            sample_polynomial_from_cbd(*polys[i], eta, StrongSpan<const KyberSamplingRandomness>(outputs[i]));
+         }
```
reneme, line 139: A type system connoisseur might suggest this (defining the `outputs` vector using the strong span), rather than forcing the type on the parameter of `sample_polynomial_from_cbd`:


```suggestion
         KyberSamplingRandomness randomness(polys.size() * poly_bytes);
         std::vector<StrongSpan<const KyberSamplingRandomness>> outputs(polys.size());
         for(size_t i = 0; i < polys.size(); ++i) {
            outputs[i] = StrongSpan(randomness).subspan(i * poly_bytes, poly_bytes);
         }

         const auto& sym = m_mode.symmetric_primitives();
         sym.PRF_batch(outputs, m_seed, m_nonce);
         m_nonce += static_cast<uint8_t>(polys.size());

         for(size_t i = 0; i < polys.size(); ++i) {
            sample_polynomial_from_cbd(*polys[i], eta, outputs[i]);
         }
```

... this does require to extend the StrongSpan template a little, though:

```C++
      /// Create a typed subspan of this span
      /// @param offset the index of the first element of the subspan
      /// @param count the number of elements in the subspan
      /// @return a new StrongSpan<T> covering the subspan
      decltype(auto) subspan(typename underlying_span::size_type offset,
                             typename underlying_span::size_type count = std::dynamic_extent) const noexcept {
         return StrongSpan<T>(m_span.subspan(offset, count));
      }
```

src/lib/pubkey/kyber/kyber_common/kyber_algos.h
```diff
@@ -66,58 +68,81 @@ class PolynomialSampler {
...
-         return poly;
+      /// Samples two polynomial vectors in a single PRF batch
+      std::pair<KyberPolyVec, KyberPolyVec> sample_polynomial_vector_pair_cbd_eta1() {
+         auto result = std::make_pair(KyberPolyVec(m_mode.k()), KyberPolyVec(m_mode.k()));
+         std::vector<KyberPoly*> polys;
+         collect_polys(polys, result.first);
+         collect_polys(polys, result.second);
+         sample_polys_cbd(polys, m_mode.eta1());
+         return result;
       }
```
reneme, line 83: Mhh, I'm a bit uneasy with the low-level `KyberPoly*` but I see why it's needed and convenient here. Though, I have a suggestion how to make this a little less scary:

We could provide a `PolynomialVector::polynomials()` method like so:

```C++
      auto polynomials() {
         std::vector<Polynomial<Trait, D>*> result;
         result.reserve(m_vec.size());
         for(auto& p : m_vec) {
            result.push_back(&p);
         }
         return result;
      }
```

... and then, this  function would become:

```suggestion
      /// Samples two polynomial vectors in a single PRF batch
      std::pair<KyberPolyVec, KyberPolyVec> sample_polynomial_vector_pair_cbd_eta1() {
         auto result = std::make_pair(KyberPolyVec(m_mode.k()), KyberPolyVec(m_mode.k()));
         sample_polys_cbd(concat(result.first.polynomials(),  //
                                 result.second.polynomials()),
                          m_mode.eta1());
         return result;
      }
```

... at least its a little more declarative like that. The same pattern works for the other usages of `collect_polys` as well and makes it obsolete. I'll comment suggestions there as well, for illustration.

src/lib/pubkey/kyber/kyber_common/kyber_algos.h
```diff
@@ -66,58 +68,81 @@ class PolynomialSampler {
...
          requires std::same_as<KyberEncryptionRandomness, SeedT>
       {
-         KyberPolyVec vec(m_mode.k());
+         std::pair<KyberPolyVec, KyberPoly> result{KyberPolyVec(m_mode.k()), KyberPoly()};
+         std::vector<KyberPoly*> polys;
+         collect_polys(polys, result.first);
+         polys.push_back(&result.second);
+         sample_polys_cbd(polys, m_mode.eta2());
+         return result;
+      }
```
reneme, line 95: As mentioned above:

```suggestion
      /// Samples a polynomial vector and a single polynomial in one PRF batch
      std::pair<KyberPolyVec, KyberPoly> sample_polynomial_vector_and_poly_cbd_eta2()
         requires std::same_as<KyberEncryptionRandomness, SeedT>
      {
         std::pair<KyberPolyVec, KyberPoly> result{KyberPolyVec(m_mode.k()), KyberPoly()};
         sample_polys_cbd(concat(result.first.polynomials(), std::array{&result.second}), m_mode.eta2());
         return result;
      }
```

... the `std::array{}` hack makes this slightly awkward, because `concat` cannot deal with scalar values of the concatenated type. I've been planning to extend this for a long time and never actually got around to doing it.

src/lib/pubkey/kyber/kyber_common/kyber_algos.h
```diff
@@ -66,58 +68,81 @@ class PolynomialSampler {
...
-      KyberSamplingRandomness prf(size_t bytes) {
-         const auto& sym = m_mode.symmetric_primitives();
-         auto seed_span = m_seed.get();
-         sym.setup_PRF(m_prf_xof, seed_span, m_nonce++);
-         return m_prf_xof->output<KyberSamplingRandomness>(bytes);
+      void sample_poly_vec_cbd(KyberPolyVec& vec, KyberConstants::KyberEta eta) {
+         std::vector<KyberPoly*> polys;
+         collect_polys(polys, vec);
+         sample_polys_cbd(polys, eta);
       }
```
reneme, line 119: As mentioned above. We might as well inline this one-liner into `sample_polynomial_vector_cbd_eta1` and get rid of the entire function that way.

```suggestion
      void sample_poly_vec_cbd(KyberPolyVec& vec, KyberConstants::KyberEta eta) {
         sample_polys_cbd(vec.polynomials(), eta);
      }
```

FAlbertDev (approved):
Since ML-KEM is complex I did not comprehend every detail of these changes, but I trust in our tests and valgrind for side-channel analysis 😜

reneme (approved):
Some structuring nits and suggestions.

src/lib/pubkey/kyber/kyber_common/kyber_algos.cpp
randombit: done

src/lib/pubkey/kyber/kyber_common/kyber_algos.h
randombit: done

src/lib/pubkey/kyber/kyber_common/kyber_algos.h
randombit: done

PR #5882, opened 2026-08-28
Fix outdated description of SubjectAlternativeName parsing in the handbook

moritzschmitt:
The handbook's X.509v3 extensions listing states that when parsing
SubjectAlternativeName only "rfc822Name", "dNSName",
"uniformResourceIdentifier" and raw IPv4 are stored and "all others
are ignored".

This no longer matches the decoder. `AlternativeName::decode_from`
(`src/lib/x509/alt_name.cpp`) stores rfc822Name, dNSName, URIs,
directoryName, registeredID, and both IPv4 and IPv6 addresses — as
well as otherName entries, whose type-id and raw value are captured
verbatim and exposed via `AlternativeName::other_name_values()`.
otherName values whose inner value is a string type of the Universal
class are additionally available in decoded form via the deprecated
`AlternativeName::other_names()` accessor (limited to the common
string types — e.g. the UTF8String used by the Microsoft UPN; wide
types such as BMPString are not retained). SmtpUTF8Mailbox otherNames
are parsed specifically. The only GeneralName choices that are
actually ignored are x400Address and ediPartyName (both are validated
to be non-empty, then discarded).

The outdated note is actively misleading: it suggests that, for
example, the Microsoft UPN (an otherName) cannot be retrieved — which
misled us during an evaluation of Botan for certificate-based
authentication. This PR updates the bullet accordingly; the "Issuer
Alternative Names" bullet refers to it and stays consistent
automatically. It is a documentation-only change, touching only
`doc/api_ref/x509.rst`.


randombit (approved):
Thank you

PR #5892, opened 2026-08-31
Optimize Whirlpool base implementation on ARM64 and x86-64

huven:
## Summary

This improves the performance of Whirlpool's base compression function on ARM64 and x86-64.

The 256-entry transformation table is expanded from 2 KiB to 4 KiB by storing two consecutive copies of each 64-bit entry. This allows the byte rotations required by `whirl()` to be performed using loads at offsets 0 through 7, avoiding explicit rotate instructions.

The key-schedule and message-round loops are also fused. Instead of materializing all 11 round keys in an 88-word array, the implementation keeps the current eight-word key state and updates it alongside the message state during each round.

`whirl()` is force-inlined because GCC on ARM64 otherwise failed to inline it, resulting in a performance regression.

## Platform scope

The optimized implementation is limited to ARM64 and x86-64. Both platforms support efficient unaligned 64-bit loads and were benchmarked with this implementation.

Although the loads are alignment-safe at the C++ level, other architectures may lower them to slower byte-wise operations. The larger table could also regress performance on processors with smaller caches. Those platforms therefore retain the existing 2 KiB table, explicit rotations, and separate key schedule.

The x86-64 SIMD implementations are unchanged and bypass this base implementation. x86-64 base benchmarks were run with assembly disabled to prevent SIMD dispatch.

## Benchmarks

Representative 1024-byte Whirlpool results, rounded:

- Apple M3 ARM64, Apple Clang 21: **+22%**
- Debian ARM64 VM, GCC 14.2: **+14%**
- x86-64 Mac Pro, Apple Clang 21, assembly disabled: **+45%**

All 136 Whirlpool test vectors pass.

### DIFF
--- a/src/lib/hash/whirlpool/whirlpool.cpp
+++ b/src/lib/hash/whirlpool/whirlpool.cpp
@@ -7,10 +7,12 @@
 
 #include <botan/internal/whirlpool.h>
 
+#include <botan/compiler.h>
 #include <botan/internal/bit_ops.h>
 #include <botan/internal/buffer_slicer.h>
 #include <botan/internal/loadstor.h>
 #include <botan/internal/rotate.h>
+#include <botan/internal/target_info.h>
 #include <array>
 
 #if defined(BOTAN_HAS_CPUID)
@@ -42,6 +44,12 @@ consteval std::array<uint8_t, 256> whirlpool_sbox() noexcept {
    return S;
 }
 
+#if defined(BOTAN_TARGET_ARCH_IS_ARM64) || defined(BOTAN_TARGET_ARCH_IS_X86_64) || defined(BOTAN_TARGET_ARCH_IS_PPC64)
+constexpr bool WhirlpoolUnalignedTables = true;
+#else
+constexpr bool WhirlpoolUnalignedTables = false;
+#endif
+
 // Combined S-box + MDS diffusion table
 consteval std::array<uint64_t, 256> whirlpool_T_table(const std::array<uint8_t, 256>& S) noexcept {
    // MDS circulant matrix first row: [1, 1, 4, 1, 8, 5, 2, 9] over GF(2^8)
@@ -54,6 +62,22 @@ consteval std::array<uint64_t, 256> whirlpool_T_table(const std::array<uint8_t, 
    return T;
 }
 
+constexpr size_t WHIRL_U_STRIDE = 2 * sizeof(uint64_t);
+
+[[maybe_unused]] consteval std::array<uint8_t, 256 * WHIRL_U_STRIDE> whirlpool_U_table(
+   const std::array<uint64_t, 256>& T) noexcept {
+   // Duplicating each entry permits unaligned loads to perform the rotations
+   std::array<uint8_t, 256 * WHIRL_U_STRIDE> U = {};
+
+   for(size_t i = 0; i != 256; ++i) {
+      for(size_t j = 0; j != sizeof(uint64_t); ++j) {
+         U[WHIRL_U_STRIDE * i + j] = static_cast<uint8_t>(T[i] >> (8 * j));
+         U[WHIRL_U_STRIDE * i + sizeof(uint64_t) + j] = static_cast<uint8_t>(T[i] >> (8 * j));
+      }
+   }
+   return U;
+}
+
 // Round constants are from the first 64 elements of the sbox
 consteval std::array<uint64_t, 10> whirlpool_rc(const std::array<uint8_t, 256>& S) noexcept {
    std::array<uint64_t, 10> RC = {};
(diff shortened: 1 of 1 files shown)


src/lib/hash/whirlpool/whirlpool.cpp
```diff
@@ -43,6 +45,24 @@ consteval std::array<uint8_t, 256> whirlpool_sbox() noexcept {
 }
 
 // Combined S-box + MDS diffusion table
+#if defined(BOTAN_TARGET_ARCH_IS_ARM64) || defined(BOTAN_TARGET_ARCH_IS_X86_64)
```
randombit: OK to sniff architecture once, but then move to `constexpr` land wherever possible

```
#if defined(BOTAN_TARGET_ARCH_IS_ARM64) || defined(BOTAN_TARGET_ARCH_IS_X86_64)
   static constexpr bool WhirlpoolUnalignedTables = true;
   // define a dedicated macro here (eg `BOTAN_WHIRLPOOL_USE_UNALIGNED_TABLES`) only if really necessary
#else 
   static constexpr bool WhirlpoolUnalignedTables = false;
#endif

... no further `BOTAN_TARGET_ARCH_` checks follow
```

src/lib/hash/whirlpool/whirlpool.cpp
```diff
@@ -67,6 +88,21 @@ constexpr auto WHIRL_S = whirlpool_sbox();
 alignas(256) constexpr auto WHIRL_T = whirlpool_T_table(WHIRL_S);
 constexpr auto WHIRL_RC = whirlpool_rc(WHIRL_S);
 
+#if defined(BOTAN_TARGET_ARCH_IS_ARM64) || defined(BOTAN_TARGET_ARCH_IS_X86_64)
+BOTAN_FORCE_INLINE uint64_t
+whirl(uint64_t x0, uint64_t x1, uint64_t x2, uint64_t x3, uint64_t x4, uint64_t x5, uint64_t x6, uint64_t x7) {
+   const uint64_t s0 = load_le<uint64_t>(WHIRL_T.data() + WHIRL_T_STRIDE * get_byte<0>(x0), 0);
```
randombit: Have just one definition of `whirl` (ok to force inline in all cases) then use `if constexpr(WhirlpoolUnalignedTables)` to decide which variant to execute.

src/lib/hash/whirlpool/whirlpool.cpp
```diff
@@ -164,6 +207,45 @@ void Whirlpool::compress_n(digest_type& digest, std::span<const uint8_t> input,
       uint64_t B6 = M[6] ^ K[6];
       uint64_t B7 = M[7] ^ K[7];
 
+#if defined(BOTAN_TARGET_ARCH_IS_ARM64) || defined(BOTAN_TARGET_ARCH_IS_X86_64)
```
randombit: This is somewhat independent of the unaligned tables. It might well be a win (or neutral) on most processors, since it reduces the stack usage quite a bit and may offer a bit more ILP. If so we could just always use this approach, regardless of if unaligned tables are being used or not. [It apparently winning on both x86-64 and aarch64 is probably sufficient justification to always use this method, those are certainly the most relevant processors these days]

src/lib/hash/whirlpool/whirlpool.cpp
```diff
@@ -164,6 +207,45 @@ void Whirlpool::compress_n(digest_type& digest, std::span<const uint8_t> input,
...
+         const uint64_t K0 = whirl(K[0], K[7], K[6], K[5], K[4], K[3], K[2], K[1]) ^ WHIRL_RC[r];
+         const uint64_t K1 = whirl(K[1], K[0], K[7], K[6], K[5], K[4], K[3], K[2]);
+         const uint64_t K2 = whirl(K[2], K[1], K[0], K[7], K[6], K[5], K[4], K[3]);
+         const uint64_t K3 = whirl(K[3], K[2], K[1], K[0], K[7], K[6], K[5], K[4]);
+         const uint64_t K4 = whirl(K[4], K[3], K[2], K[1], K[0], K[7], K[6], K[5]);
+         const uint64_t K5 = whirl(K[5], K[4], K[3], K[2], K[1], K[0], K[7], K[6]);
+         const uint64_t K6 = whirl(K[6], K[5], K[4], K[3], K[2], K[1], K[0], K[7]);
+         const uint64_t K7 = whirl(K[7], K[6], K[5], K[4], K[3], K[2], K[1], K[0]);
+
+         const uint64_t T0 = whirl(B0, B7, B6, B5, B4, B3, B2, B1) ^ K0;
```
randombit, line 197: Just out of curiosity did you try using Kn immediately after it's computed, eg

```
         const uint64_t K0 = whirl(K[0], K[7], K[6], K[5], K[4], K[3], K[2], K[1]) ^ WHIRL_RC[r];
         const uint64_t T0 = whirl(B0, B7, B6, B5, B4, B3, B2, B1) ^ K0;
         const uint64_t K1 = whirl(K[1], K[0], K[7], K[6], K[5], K[4], K[3], K[2]);
...
```

One of those things where maybe it helps, maybe it hurts, maybe the compiler does the same thing no matter how you express it. Worth a check anyway.

src/lib/hash/whirlpool/whirlpool.cpp
```diff
@@ -43,6 +45,24 @@ consteval std::array<uint8_t, 256> whirlpool_sbox() noexcept {
 }
 
 // Combined S-box + MDS diffusion table
+#if defined(BOTAN_TARGET_ARCH_IS_ARM64) || defined(BOTAN_TARGET_ARCH_IS_X86_64)
+constexpr size_t WHIRL_T_STRIDE = 2 * sizeof(uint64_t);
+
+consteval std::array<uint8_t, 256 * WHIRL_T_STRIDE> whirlpool_T_table(const std::array<uint8_t, 256>& S) noexcept {
```
randombit: I'd prefer if this function, instead of replacing `T`, it took in `T` as an input, eg

```
alignas(256) constexpr auto WHIRL_U = whirlpool_U_table(whirlpool_T_table(WHIRL_S));
```

where `whirlpool_U_table` just sprays out the bytes of T without repeating the MDS mul code.

src/lib/hash/whirlpool/whirlpool.cpp
randombit: Also likely the simplest thing is to define `WHIRL_T`/`WHIRL_U` as `static constexpr` inside the definition of `whirl`, since this function is the only one that needs the table at all.

randombit (commented):
So general approach seems fine and a nice win, I'd rather see less preprocessor though, see comments.

src/lib/hash/whirlpool/whirlpool.cpp
huven: There is actually something unexpected here: on x64, when disabling the unaligned table, and keeping the fused loops, performance *degrades* with about 5%. Just the aligned table (no fused loop) gives 28% increase, and both together even more.

Just to illustrate that the fused loop by itself is not guarantee for improvement. I didn't expected that, but that's why I enabled them both or none of the two.

randombit: On POWER10 just this fused loop without the new tables improves by 32% (!) [using the unaligned tables adds another 20% on top of that]

On RISCV64, fused loops has no change. Unaligned tables causes a large regression.

Leaning towards use the fused loops everywhere, and conditionally setting the unaligned tables based on architecture. Might cause a regression on some rare machine, or compiler version we don't test, etc but that's true for anything. Significant clear win on x86-64, aarch64, and POWER and flat on riscv64 seems like a decent outcome to me, no reason to complicate the code more than necessary.

huven: That's very good input, ok, I was already refactoring the code with the constexpr instead of preprocessing, I'll drop the non-fused loops then.

src/lib/hash/whirlpool/whirlpool.cpp
huven: I just gave it a try: it regressed about 3% on x64 and about 6% on ARM64

src/lib/hash/whirlpool/whirlpool.cpp
huven: This change is done.

src/lib/hash/whirlpool/whirlpool.cpp
huven: Done too.

src/lib/hash/whirlpool/whirlpool.cpp
huven: Non-fused loop is dropped.

src/lib/hash/whirlpool/whirlpool.cpp
huven: That's better indeed, this is done too.

randombit:
@huven Can you squash to a single commit? Besides that looks good to me.

On Tiger Lake GCC 16, improvement of almost 40%

huven:
> On Tiger Lake GCC 16, improvement of almost 40%

@randombit Nice!

I've squashed the commits, no other changes.

randombit (approved):
Very nice improvement thanks!

PR #5903, opened 2026-09-02
Expose SAN/IAN otherName entries via the FFI GeneralName API

### LINKED ISSUE #5885: FFI: SAN otherName entries (e.g. Microsoft UPN) are not accessible
### Motivation

In certificate-based authentication against Active Directory / AD CS-issued
certificates, the identity frequently lives in a SubjectAlternativeName
`otherName` entry: the Microsoft UPN (szOID_NT_PRINCIPAL_NAME,
OID `1.3.6.1.4.1.311.20.2.3`, value encoded as a UTF8String). Reading it is a
recurring requirement in enterprise and government environments.

With OpenSSL's C API this is straightforward (`GEN_OTHERNAME` +
`NID_ms_upn`). With Botan, the C++ API can do it
(`AlternativeName::other_name_values()` exposes the type-id OID and the raw
inner value), but the C89 FFI cannot surface otherName entries in any
structured form (they only appear as `OtherName <oid>: <hex>` lines inside
the human-readable `botan_x509_cert_view_as_string()` dump) — which also
affects every language binding built on top of the FFI. For C integrators
migrating from OpenSSL to Botan this is currently a hard gap.

Context: we integrate Botan (the BSI-recommended line) in security products
for sectors with elevated security requirements, where UPN extraction from
client certificates is a routine need.

### Current behavior (master)

* `ffi.h` states above `enum botan_x509_general_name_types`: *"there is
  currently no way to access OTHER_NAME values via the FFI."* The handbook
  (`doc/api_ref/ffi.rst`, `botan_x509_general_name_get_type`) says: *"Support
  for `BOTAN_X509_OTHER_NAME` is deprecated and cannot be viewed using these
  functions."* Both sentences were introduced together with the API in
  #5217.
* The SAN iteration helpers in `src/lib/ffi/ffi_cert.cpp`
  (`extract_general_name_at()` / `count_general_names_in()`) cover email, DNS,
  directory names, URIs, IPv4 and IPv6 — otherNames are omitted, so
  `botan_x509_cert_subject_alternative_names()` never yields them.
* `botan_x509_general_name_get_type()` deliberately returns
  `BOTAN_FFI_ERROR_INVALID_OBJECT_STATE` when the type maps to
  `BOTAN_X509_OTHER_NAME` (inline comment: *"viewing of other-names not
  supported"*), and both view functions reject that type as well — today
  reachable only via the name-constraint accessors.
* Underlying limitation: `Botan::GeneralName` declares `NameType::Other` but
  has no factory and no payload storage for an otherName (type-id + value), so
  a `GeneralName` cannot carry one; `GeneralName::decode_from()` records only
  the type tag. A `TODO` in `pkix_types.h` suggests adding an
  `OtherConstraint` alternative to the `NameVariant`. Note that `GeneralName`
  is itself documented as deprecated ("will be removed in a future major
  release"), which #5217 already accepted when wrapping it in
  `botan_x509_general_name_t`.

### Proposal

Two possible shapes; we are happy to implement either and submit a PR once
there is agreement on the direction.

#### Option A (our suggestion): include otherNames in the existing GeneralName enumeration

* Extend `Botan::GeneralName` with an otherName payload (the `OtherConstraint`
  `NameVariant` alternative the `TODO` considers, holding type-id OID + raw
  BER value): a factory along the lines of
  `GeneralName::other_name(OID type_id, std::span<const uint8_t> ber_value)`,
  plus storage and accessors, sourced from
  `AlternativeName::other_name_values()` on the SAN/IAN path;
  `GeneralName::decode_from()` would retain type-id + value likewise, so that
  otherName *name constraints* (today the only source of an otherName-typed
  handle) carry a payload too. Since `GeneralName` is deprecated, the factory
  could be an underscore-prefixed helper like `_uri_san_value()` (declared in
  the public header but marked internal) rather than supported public API.
* Include otherNames in `count_general_names_in()` /
  `extract_general_name_at()` (the NOTE comments there already anticipate
  extending the handled set).
* FFI additions:

  ```c
  /**
  * Views the otherName type-id as a dotted-decimal OID string, or returns
  * BOTAN_FFI_ERROR_INVALID_OBJECT_STATE if the name is not an otherName.
  */
  BOTAN_FFI_EXPORT(3, 14)
  int botan_x509_general_name_view_other_name_type_id(botan_x509_general_name_t name,
                                                      botan_view_ctx ctx,
                                                      botan_view_str_fn view);
  ```

* `botan_x509_general_name_view_binary_value()`: for otherNames, view the raw
  BER of the inner value (the complete TLV inside the `EXPLICIT [0]` wrapper,
  i.e. exactly what `OtherNameValue::value()` holds).
* Optionally, `botan_x509_general_name_view_string_value()`: succeed for
  otherNames whose inner value is a UTF8String or a UTF-8-subset string type
  (IA5String, PrintableString, VisibleString, NumericString) — this directly
  covers the UPN case and mirrors what `AlternativeName::decode_from()`
  already accepts into the (deprecated) `other_names()` set;
  `BOTAN_FFI_ERROR_INVALID_OBJECT_STATE` otherwise.
* Remove the OTHER_NAME carve-out in `get_type()`, the *"no way to access"*
  comment in `ffi.h`, and the *"deprecated"* sentence in `ffi.rst`.

Compatibility note: existing callers will observe additional SAN/IAN entries,
and for otherName name constraints `get_type()` will succeed where it
previously errored. Since the API is young (3.11) and callers are expected to
switch on `get_type()`, this seems acceptable. Note that `SmtpUTF8Mailbox`
entries (RFC 9598) are part of `other_name_values()` and would therefore also
show up as otherNames.

#### Option B (minimal alternative): dedicated other-name accessors

Leave `GeneralName` and the current enumeration untouched and add a parallel
API over `AlternativeName::other_name_values()`:

```c
BOTAN_FFI_EXPORT(3, 14)
int botan_x509_cert_subject_other_names_count(botan_x509_cert_t cert, size_t* count);

BOTAN_FFI_EXPORT(3, 14)
int botan_x509_cert_subject_other_name_view_type_id(botan_x509_cert_t cert,
                                                    size_t index,
                                                    botan_view_ctx ctx,
                                                    botan_view_str_fn view);

BOTAN_FFI_EXPORT(3, 14)
int botan_x509_cert_subject_other_name_view_value(botan_x509_cert_t cert,
                                                  size_t index,
                                                  botan_view_ctx ctx,
                                                  botan_view_bin_fn view);
```

Pro: zero behavior change and no public C++ API change. Con: SAN contents are
split across two enumerations (and, given the existing subject/issuer
symmetry, a matching `botan_x509_cert_issuer_other_name*` set would be
needed).

We prefer Option A because `BOTAN_X509_OTHER_NAME` is already reserved in the
enum and the `get_type()` carve-out reads like a placeholder for exactly this.
That said, the handbook calls OTHER_NAME support "deprecated"; since #4678
explicitly kept otherName support in the C++ API, we read that as referring
to the deprecated string-only `other_names()` accessor — if the enumerator is
instead meant to be phased out, Option B would be the natural fit, so a short
confirmation of the direction would be appreciated.

### Testing

We would include FFI tests using a certificate carrying an msUPN otherName.
The existing X.509 test data contains otherName SANs (SmtpUTF8Mailbox,
SRVName, permanentIdentifier, … under `src/tests/data/x509/name_constraints/`)
and an msUPN otherName *name constraint*
(`othername-msupn-constraint-no-upn-san-valid.pem`), but no certificate with
an msUPN otherName SAN, so we would add a generated one (and reuse the
existing otherName certificates for the non-string case). Per
`doc/dev_ref/contributing.rst`, the PR would also add the `botan3.py`
declarations plus a Python test — note that `botan3.py` currently has no
bindings for the `botan_x509_general_name_t` API at all, so those would come
along.

moritzschmitt:
Fixes #5885, implementing the direction agreed on the issue: otherName entries are surfaced through the existing `botan_x509_general_name_t` enumeration (Option A), with the type-id returned as the FFI's first-class OID object rather than as a string.

## C++ library

`Botan::GeneralName` now retains the otherName payload (type-id OID plus the raw inner value) in its `NameVariant`, reusing `AlternativeName::OtherNameValue`. A factory `GeneralName::other_name()` and an accessor `GeneralName::other_name_value()` are added, `GeneralName::decode_from()` retains type-id and value for otherName name constraints (using the same structure checks as `AlternativeName::decode_from()`), and `GeneralName::encode_into()` can now encode otherNames, so the corpus re-encode test covers them.

One consequence: a malformed otherName in a NameConstraints extension (value not wrapped in the `EXPLICIT [0]` tag, missing type-id or value, trailing content) now fails to decode, where previously it was retained as an opaque entry without a payload. The extension then becomes an `Unknown_Extension` with `failed_to_decode` set, and path validation reports `EXTENSION_ENCODING_ERROR` (matching how a malformed otherName in a SubjectAlternativeName already behaves).

## FFI

`botan_x509_cert_subject_alternative_names()`, `botan_x509_cert_issuer_alternative_names()` and the name constraint enumerators now yield otherName entries (appended after the previously supported types, so existing indices are unchanged), and `botan_x509_general_name_get_type()` returns `BOTAN_X509_OTHER_NAME` instead of erroring. For an otherName:

* the new `botan_x509_general_name_other_name_type_id()` returns the type-id as a new OID object (to be freed with `botan_oid_destroy`), following `botan_pubkey_oid()`; callers can compare it via `botan_oid_equal()` against e.g. `botan_oid_from_string("Microsoft UPN")`.
* `botan_x509_general_name_view_binary_value()` views the raw BER of the value, i.e. the complete TLV inside the `EXPLICIT [0]` tag (exactly what `AlternativeName::OtherNameValue::value()` holds).
* `botan_x509_general_name_view_string_value()` succeeds for values of the string types that `AlternativeName::decode_from()` also decodes into the deprecated `other_names()` set — UTF8String, IA5String, PrintableString, VisibleString and NumericString with content valid for the type — which directly covers the Microsoft UPN. Since the string view hands out a NUL-terminated C string, values with an embedded NUL are additionally rejected; both view functions document this. Anything else returns `BOTAN_FFI_ERROR_INVALID_OBJECT_STATE`.

The FFI changes are additive (new API version 20260901, exports tagged 3.14). As discussed in the issue, existing callers will observe additional SAN/IAN entries, `get_type()` succeeds for otherName name constraints where it previously errored, and SmtpUTF8Mailbox entries (RFC 9598) appear as otherNames. The stale "no way to access OTHER_NAME" comment in `ffi.h` and the "deprecated" sentence in `ffi.rst` are removed.

## Python binding

`botan3.py` previously had no bindings for the GeneralName API at all; this adds `X509GeneralName`/`X509GeneralNameType` plus the `X509Cert` accessors for subject/issuer alternative names and permitted/excluded name constraints, along with documentation and tests.

## Test data

`src/tests/data/x509/othername/` contains generated certificates covering the msUPN case, every accepted and rejected string type, non-string values, multiple otherNames beside other name types, an IssuerAlternativeName otherName, otherName name constraints, and an empty as well as a 3000-byte value. The generator script is committed (`src/scripts/dev_tools/gen_othername_testdata.sh`); the README documents each certificate and the expected type-ids and inner values, which were read off `openssl asn1parse` output independently of the implementation. String values that are invalid for their tag cannot be produced with OpenSSL, so one certificate is built in the test itself via the C++ API. The existing otherName certificates under `name_constraints/` are reused to pin the behavior for SmtpUTF8Mailbox, SRVName, permanentIdentifier and SIM entries.


doc/api_ref/ffi.rst
```diff
@@ -259,6 +259,7 @@ supported it.
 ============== ===================
 FFI Version    Supported Starting
 ============== ===================
+20260901       3.14.0
```
randombit: This should be 20261103 (expected 3.14 release date)

arckoor, line 262: 3.14 would be 20261103, usually it's the date of the actual release. May be better to leave out completely and let it be handled in a separate PR, but as long as no one else opens a PR also looking to do this at the same time should probaly be fine :p

src/lib/ffi/ffi_cert.cpp
```diff
@@ -775,6 +826,25 @@ int botan_x509_general_name_view_binary_value(botan_x509_general_name_t name,
 #endif
 }
 
+int botan_x509_general_name_other_name_type_id(botan_asn1_oid_t* oid, botan_x509_general_name_t name) {
+#if defined(BOTAN_HAS_X509_CERTIFICATES)
+   return BOTAN_FFI_VISIT(name, [=](const Botan::GeneralName& n) -> int {
+      if(Botan::any_null_pointers(oid)) {
+         return BOTAN_FFI_ERROR_NULL_POINTER;
+      }
```
arckoor, line 834: This before the #ifdef please, so behaviour re `nullptr`s doesn't change based on how the lib is compiled

arckoor (commented):
Rough review only, and only really of the parts I'm comfortable with, i.e. FFI and Python.

I'm not sure exposing `X509GeneralName` in the python binding directly is a good call, I've been wanting to add something like this for a while, see this [branch](https://github.com/arckoor/botan/tree/more-python-cert-getters). I'm not convinced what's done there is the way to go either (since highly inefficient if you do want *everything*), but perhaps parts are still of interest

doc/api_ref/ffi.rst
moritzschmitt: Fair point, I wasn't sure about this either, but 3.14 was too tempting a version to leave unclaimed. If you'd rather keep the bump out of this PR, I'll drop it. Otherwise it can be adjusted to the actual release date the way the 3.11 stamp was.

src/lib/ffi/ffi_cert.cpp
moritzschmitt: I went back and forth on this one and ended up following the immediate neighbours: the GeneralName functions already in the file (`botan_x509_general_name_get_type`, the `*_alternative_names` / `*_name_constraints` enumerators and their `_count` variants) all do their argument checks inside the `#if`, and so does `botan_x509_cert_serial_number`, which has the same shape (out-pointer for a new object plus a handle). The file has both styles: `not_before`, `get_fingerprint`, `get_public_key` and the `ext_*` getters check before the `#if`, the newer X.509 additions check inside.

The other thing that held me back is that moving the `oid` check alone doesn't quite give config-independent behaviour: the handle is validated by `BOTAN_FFI_VISIT`, which sits inside the `#if` in every function of this file, so `(&oid, nullptr)` still returns `NOT_IMPLEMENTED` in a build without X.509 either way. Since no `botan_x509_general_name_t` can exist in such a build at all, `NOT_IMPLEMENTED` for any argument combination seemed a reasonable answer to me.

That said, it's a two-line change and I don't feel strongly about it. If you'd rather have the out-pointer checked first, I'll move it, and would suggest doing the same for the rest of the GeneralName block in a small follow-up so the API stays uniform. Happy to do either.

moritzschmitt:
> I'm not sure exposing `X509GeneralName` in the python binding directly is a good call, I've been wanting to add something like this for a while, see this [branch](https://github.com/arckoor/botan/tree/more-python-cert-getters). I'm not convinced what's done there is the way to go either (since highly inefficient if you do want _everything_), but perhaps parts are still of interest

Thanks for the pointer, I had a look and tried the branch against the test certs. I think the two approaches want different things: `subject_alt_name(type) -> list[str]` is handy for the "give me the DNS names" case, but it can't carry a directoryName (the string view raises there, and the handle isn't freed on that path), an IP constraint with its mask, or an otherName, which needs the type-id next to the value. That's why I suggest to go with a small wrapper object. The value type depends on the name type, and the object lets the caller pick the right accessor, the same way `X509CRL.revoked()` returns `X509CRLEntry` objects rather than tuples. It's also the cheaper option for "everything": one pass over the SAN instead of one per type (11 vs 50 enumerations on the 11-entry test cert).

If a strings-only convenience for the common case would be useful, I'm happy to add one on top of the object list, in a single pass.

arckoor:
The thought about the handle potentially not being freed is a good point, I hadn't considered (yet anyway). What might also work is wrapping the `list[str]` in an entirely new class, and having that class contain a `.filter(type: X509GeneralNameType) -> list[str]`. Not saying to do that right now, just food for thought

src/lib/ffi/ffi_cert.cpp
arckoor: Yeah I agree, it's kind of a mess. The ones that use the `FFI_VISIT` macro will always perform differently, but for FFI functions I've written in the past that require an explicit `nullptr` check, I've usually followed the (mostly self-imposed) "check before ifdef" rule. It's mostly a matter of taste, not a big deal either way :)

moritzschmitt:
Agreed, a filter helper on top of the object list would be a natural follow-up. The one thing to decide there is what filter(DIRECTORY_NAME) returns, since that type has no string form.

